PR #3785, opened 2026-05-07
fix(template): persist labels in cloneSubgraph (GH#3784)

### LINKED ISSUE #3784: step labels (including `gate:` from `waits_for`) silently dropped on pour
## Summary

Step labels (including the `gate:<await>` label generated from `waits_for`) silently vanish when a formula is poured. Cook produces them correctly; clone-time persistence drops them.

## Reproduction

```toml
formula = "test-waits-for"
version = 1
type = "workflow"

[[steps]]
id = "wait-step"
title = "Wait for children"
type = "task"
waits_for = "all-children"
```

```
$ bd mol pour test-waits-for
... created step prim-mol-qeg ...
$ bd label list prim-mol-qeg
prim-mol-qeg has no labels
```

Expected `gate:all-children` on the step.

## Root Cause

`processStepToIssue` (`cmd/bd/cook.go:569-572`) correctly appends `gate:<waits_for>` to `issue.Labels`. The ephemeral cook path (`cook.go:463`) calls `collectSteps` with `labelHandler=nil`, leaving labels on the in-memory issue. `cloneSubgraph` (`cmd/bd/template.go:584-622`) then copies `Labels: oldIssue.Labels` onto the new issue (line 602) and calls `tx.CreateIssue` (line 617) — but never `tx.AddLabel`. Beads stores labels in a separate table, so they're lost.

The DB-cook path (`cook.go:850-870`) does this correctly: it collects labels separately and calls `tx.AddLabel` per label inside the same transaction. `cloneSubgraph` just doesn't.

Affects every label, not just `gate:` — any user-set step labels have the same problem on the pour path.

## Fix

In `cloneSubgraph` after `tx.CreateIssue` succeeds, iterate `oldIssue.Labels` and call `tx.AddLabel(ctx, newIssue.ID, label, opts.Actor)` within the same transaction.

---

*Discovered while smoke-testing primitives for #3621.*

jjgarzella:
## What

In `cloneSubgraph`, persist `oldIssue.Labels` via `tx.AddLabel` after `tx.CreateIssue`, so step labels actually reach storage on pour.

## Why

Fixes #3784. `tx.CreateIssue` does not persist the `Labels` field — storage keeps labels in a separate table. `cloneSubgraph` was copying `oldIssue.Labels` onto the new issue but never calling `AddLabel`, so any labels on the in-memory subgraph silently vanished. The most user-visible casualty was the `gate:<await>` label that cook generates from `waits_for` (`cmd/bd/cook.go:569-572`): present on the cooked step, absent on the poured step. Affects every label on the pour-from-fresh-cook path, not just `gate:`.

## Verification

```
go test -tags 'gms_pure_go cgo' -run 'TestTemplateSuite/Clone_PersistsInMemoryLabels' ./cmd/bd/
```

End-to-end repro from #3784:

```toml
formula = "test-waits-for"
version = 1
type = "workflow"

[[steps]]
id = "wait-step"
title = "Wait for children"
type = "task"
waits_for = "all-children"
```

```
$ bd mol pour test-waits-for
$ bd label list <step-id>
gate:all-children     # was empty before this change
```

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3785"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/cmd/bd/template.go
+++ b/cmd/bd/template.go
@@ -618,6 +618,16 @@ func cloneSubgraph(ctx context.Context, s storage.DoltStorage, subgraph *Templat
 				return fmt.Errorf("failed to create issue from %s: %w", oldIssue.ID, err)
 			}
 
+			// Persist labels separately - tx.CreateIssue does not handle the
+			// Labels field (storage keeps labels in a separate table). Without
+			// this, labels on the in-memory subgraph (e.g. the gate:<await>
+			// label cook generates from waits_for) silently vanish on pour.
+			for _, label := range oldIssue.Labels {
+				if err := tx.AddLabel(ctx, newIssue.ID, label, opts.Actor); err != nil {
+					return fmt.Errorf("failed to add label %q to %s: %w", label, newIssue.ID, err)
+				}
+			}
+
 			idMapping[oldIssue.ID] = newIssue.ID
 		}
 
--- a/cmd/bd/template_test.go
+++ b/cmd/bd/template_test.go
@@ -435,6 +435,50 @@ func TestTemplateSuite(t *testing.T) {
 		}
 	})
 
+	t.Run("Clone_PersistsInMemoryLabels", func(t *testing.T) {
+		// The fresh-cook pour path (cmd/bd/cook.go:463) calls collectSteps
+		// with labelHandler=nil, which leaves labels on issue.Labels rather
+		// than persisting them via a separate channel. cloneSubgraph must
+		// pick those labels up and call AddLabel for each, otherwise labels
+		// generated at cook time (e.g. the gate:<await> label from waits_for)
+		// silently vanish on pour. Regression test for #3784.
+		epic := &types.Issue{
+			ID:        "proto-clone-labels",
+			Title:     "Proto with labels",
+			IssueType: types.TypeEpic,
+			Status:    types.StatusOpen,
+			Priority:  2,
+			Labels:    []string{"gate:all-children", "custom-label"},
+		}
+		subgraph := &TemplateSubgraph{
+			Root:     epic,
+			Issues:   []*types.Issue{epic},
+			IssueMap: map[string]*types.Issue{epic.ID: epic},
+		}
+
+		opts := CloneOptions{Actor: "test-user"}
+		result, err := cloneSubgraph(ctx, s, subgraph, opts)
+		if err != nil {
+			t.Fatalf("cloneSubgraph failed: %v", err)
+		}
+
+		gotLabels, err := s.GetLabels(ctx, result.NewEpicID)
+		if err != nil {
+			t.Fatalf("GetLabels failed: %v", err)
+		}
+
+		want := map[string]bool{"gate:all-children": true, "custom-label": true}
+		got := make(map[string]bool, len(gotLabels))
+		for _, l := range gotLabels {
+			got[l] = true
+		}
+		for label := range want {
+			if !got[label] {
+				t.Errorf("cloned issue %s missing label %q (got: %v)", result.NewEpicID, label, gotLabels)
+			}
+		}
+	})
+
 	t.Run("Clone_AssigneeOverrideRootOnly", func(t *testing.T) {
 		epic := h.createIssue("Root Epic", "", types.TypeEpic, 1)
 		child := h.createIssue("Child Task", "", types.TypeTask, 2)


maphew:
**Triage: RETIRE (already superseded on main) — thank @jjgarzella for the clear diagnosis and regression test.**

## Reasoning

#3784/#3785 diagnose a real historical bug: `cloneSubgraph` in `cmd/bd/template.go` copies `oldIssue.Labels` onto the new in-memory `Issue` struct and calls `tx.CreateIssue`, but (per the issue writeup) `tx.CreateIssue` "does not handle the Labels field" — so labels such as the `gate:<await>` label generated by `waits_for` would silently vanish on pour.

That description no longer matches current `upstream/main`. The issue-creation plumbing was refactored into `internal/storage/issueops/create.go`, and `CreateIssueInTxWithResult` (used by **both** the singular `tx.CreateIssue` and the batch `tx.CreateIssues` paths) now unconditionally calls `PersistLabels(ctx, tx, issue, actor, eventTable)` (upstream/main `internal/storage/issueops/create.go:58`). `PersistLabels` (same file, line 567) iterates `issue.Labels` and does `INSERT IGNORE INTO labels (issue_id, label) ...` plus a `label_added` event — exactly the behavior the PR proposes adding by hand in `cloneSubgraph`.

This is confirmed by an existing, currently-passing test: `internal/storage/embeddeddolt/create_issue_test.go:1163` (`with_labels` subtest) constructs an `Issue{..., Labels: []string{"bug", "urgent"}}`, calls `CreateIssues`, and asserts 2 rows land in `labels`. So the exact scenario the PR's regression test exercises (labels set on the in-memory struct before creation, persisted via `tx.CreateIssue`) is already covered and already works on main. `cloneSubgraph` itself on main (`cmd/bd/template.go:554`) does not need its own `AddLabel` loop because the storage layer now does it for every `CreateIssue` caller uniformly.

Because this ORPHAN branch predates the beads history rewrite (root `0a61a566b`), and `PersistLabels`/the `with_labels` test are present at that very root commit, this fix has been in place for the entire life of current `main` — it isn't a recent fix that raced the PR, it's an architectural change (label persistence moved from call-site-specific `AddLabel` loops into the shared `CreateIssue` path) that already subsumes the reported bug.

## Findings

- **Major (supersession):** Root cause no longer exists on main — `tx.CreateIssue` already persists `issue.Labels` via `PersistLabels` (`internal/storage/issueops/create.go:58,567`). Applying the PR's patch verbatim today would double-process labels (harmless due to `INSERT IGNORE`, but redundant and would emit a duplicate "Added label" event via `AddLabel`'s own event write, since `PersistLabels` already wrote one).
- **Minor:** The PR's new test (`Clone_PersistsInMemoryLabels`) is still useful as an explicit regression guard for the `cloneSubgraph`-specific pour path (in-memory labels never round-tripped through `AddLabel` before), even though the mechanism it exercises is already covered indirectly by `with_labels` in `create_issue_test.go`.
- **Nit:** Comment in the PR's diff ("tx.CreateIssue does not handle the Labels field") is inaccurate against current main and would confuse future readers if merged as-is.

## Recommendation

Retire this PR without merging the code change (the underlying bug is fixed by the `PersistLabels` refactor already on main). If maintainers want an explicit `cloneSubgraph`-scoped regression test for this exact path, it's worth cherry-picking just `Clone_PersistsInMemoryLabels` from the PR (crediting @jjgarzella) as a standalone test-only PR against current main — no production code change needed, since main's `CreateIssue` already returns the labels correctly.

Since this branch has no common ancestor with current `upstream/main`, a normal rebase isn't possible. If any part is salvaged, use:

```
git rebase --onto upstream/main <old-base> refs/prtmp/3785
```

to replay just the test commit onto current main and drop the now-redundant `template.go` hunk.

Thanks again to @jjgarzella for the careful root-cause narrative in #3784 (traced through `processStepToIssue`, both cook paths, and `cloneSubgraph`) — that diagnosis is exactly what let us confirm the fix already landed elsewhere.


maphew:
Thanks for the clone-subgraph label persistence fix. I am closing this orphaned pre-re-root PR because current `main` already preserves labels through the shared `PersistLabels` path, so the useful behavior from this branch has been superseded.

PR #3797, opened 2026-05-07
test: let bash canonicalize formula path in release script tests

seanmartinsmith:
## What

Fix `writeFakeBD` in `scripts/release_script_test.go` to let bash canonicalize the formula path at script-execution time. Two Windows-only test failures (`TestReleaseScriptInstalledBDSelectionDoesNotRequireJQ` and `TestReleaseScriptUsesVerifiedInstalledBDBeforeStaleRepoBD`) start passing.

Test-only change. One helper. Production code (`release.sh`, `bd`) untouched.

## Why

Fixes #3796.

On Windows, both tests above fail because release.sh's installed-bd selection never matches the fake bd, falls back to `go run`, and then errors out because `go run` can't find `go.mod` from the test's working directory:

```
Using bd: go run -tags gms_pure_go ./cmd/bd (falling back to go run; run make install for faster releases)
...
go: go.mod file not found in current directory or any parent directory
--- FAIL: TestReleaseScriptInstalledBDSelectionDoesNotRequireJQ (0.29s)
```

Root cause: `writeFakeBD` embeds the formula path into a bash heredoc with `fmt.Sprintf` `%q`. On Windows, `filepath.Join` returns backslashes and `%q` quotes them as `"C:\\Users\\..."`. release.sh derives `FORMULA_PATH` from bash's `pwd`, which on Git Bash/MSYS is mount-aware and produces forms like `/c/Users/...` or `/tmp/...` (the `AppData\Local\Temp` mount aliases as `/tmp`). The embedded windows path never matches whatever `pwd` produces, `bd_resolves_repo_formula` returns false for every candidate, and we fall through to the `go run` branch.

The fix lets bash canonicalize the path itself. Instead of embedding the path with `%q`, the fake bd computes `SOURCE` at execution time via `cd <formula-dir> && pwd`, so bash produces whatever canonical form release.sh will compare against. `filepath.ToSlash` is enough on the Go side because bash accepts forward-slash native windows paths for `cd`.

## Verification

Repro on Windows (Git Bash / MSYS) against upstream HEAD `8b7cbfc3b`:

```
go test -tags gms_pure_go ./scripts/... -run TestReleaseScriptInstalledBDSelectionDoesNotRequireJQ -v
```

Before this PR: fails as shown above. After this PR:

```
$ go test -tags gms_pure_go ./scripts/... -v -count=1
=== RUN   TestReleaseScriptUsesVerifiedInstalledBDBeforeStaleRepoBD
[...]

PR #3801, opened 2026-05-07
test(doltserver): release log handle in newDoltServer cleanup on Windows

### LINKED ISSUE #3798: bug: TestDoltServer_Dial_BeforeStart fails Windows tempdir cleanup (file lock on server.log)
## Symptom

`TestDoltServer_Dial_BeforeStart` in `internal/storage/db/server/doltserver_test.go` fails on Windows during test cleanup:

```
=== RUN   TestDoltServer_Dial_BeforeStart
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat C:\Users\sms\AppData\Local\Temp\TestDoltServer_Dial_BeforeStart...\004\server.log: The process cannot access the file because it is being used by another process.
--- FAIL: TestDoltServer_Dial_BeforeStart (1.71s)
```

The test itself passes its assertions. The failure is in Go's `testing` framework trying to remove the temp directory after the test, which Windows blocks because an open file handle is still held on `server.log`.

Reproduced on Windows 11 against upstream HEAD (`8b7cbfc3b`). macOS/Linux are unaffected because POSIX allows unlinking open files.

## Repro

```
go test -tags gms_pure_go ./internal/storage/db/server/... -run Dial_BeforeStart -v
```

## Root Cause

`NewDoltServer` in `internal/storage/db/server/doltserver.go` (line 88) opens the log file unconditionally when `logFilePath != ""` is passed:

```go
logFile, err = os.OpenFile(absLogFilePath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o600)
```

The handle is stored in `s.logFile` and is only closed inside `Stop()` (line 329-332). `TestDoltServer_Dial_BeforeStart` calls `newDoltServer` (which passes a log path), then calls `Dial` which fails before the server starts, then returns without ever calling `Stop`. The `*os.File` is never closed, and `t.TempDir`'s deferred `RemoveAll` cannot delete the directory on Windows because the kernel holds an exclusive lock on the open file.

## Suggested Fix

Register a `t.Cleanup` in the `newDoltServer` test helper that closes the server's log file handle before the temp dir cleanup runs. `t.Cleanup` callbacks execute LIFO, so a manually-registered close fires before `t.TempDir`'s `RemoveAll`. This is a test-only change; no production code needs modification.

The server struct does not expose its log file directly, but `Stop` already handles the close. An alternative is to expose a `Close` method or check whether `Stop` can safely be called even on a never-started server (if so, the test helper can register `t.Cleanup(func() { _ = s.Stop(ctx) })`).

Either way the fix is confined to the test file.

seanmartinsmith:
Fixes #3798

## Symptom

`TestDoltServer_Dial_BeforeStart` fails on Windows during test cleanup with:

```
testing.go:1464: TempDir RemoveAll cleanup: unlinkat ...\server.log:
    The process cannot access the file because it is being used by another process.
--- FAIL: TestDoltServer_Dial_BeforeStart (1.71s)
```

The test's assertions pass. The failure is in Go's `testing` framework trying to remove the temp directory afterward. macOS/Linux are unaffected because POSIX allows unlinking open files.

## Root cause

`NewDoltServer` opens `server.log` at construction time and stores the handle in `s.logFile`. The handle is only closed inside `Stop()`. `TestDoltServer_Dial_BeforeStart` constructs a server (via `newDoltServer`), calls `Dial` which fails as expected before the server starts, then returns — never calling `Stop`. The `*os.File` remains open, and Windows blocks `RemoveAll` on its directory.

`internal/storage/db/server/doltserver.go:88` (the `os.OpenFile` call).

## Fix

Register a `t.Cleanup` in the `newDoltServer` test helper that calls `s.Stop(context.Background())` before `t.TempDir`'s `RemoveAll` runs. Cleanup callbacks execute LIFO, so this fires first. `Stop` is already idempotent — it guards every field (`s.cancel != nil`, `s.eg != nil`, `s.logFile != nil`, `s.pid != 0`) and zeroes them on first call — so tests that already call `Stop` themselves are not affected.

Test-only change. No production code touched.

## Verification

Before (Windows, upstream HEAD `8b7cbfc3b`):

```
$ go test -tags gms_pure_go ./internal/storage/db/server/... -run Dial_BeforeStart -v
=== RUN   TestDoltServer_Dial_BeforeStart
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat ...\server.log: The process cannot access the file because it is being used by another process.
--- FAIL: TestDoltServer_Dial_BeforeStart (1.71s)
```

After (this branch):

```
$ go test -tags gms_pure_go ./internal/storage/db/server/... -run Dial_BeforeStart -v
=== RUN   TestDoltServer_Dial_BeforeStart
--- PASS: TestDoltServer_Dial_BeforeStart (0.01s)
PASS
```

Full package (`./internal/storage/db/server/...`): 26 tests pass, 1 skipped (unix sockets, expected on Windows). `make build` clean.

To reproduce the original failure: check out `8b7cbfc3b` on Windows and run:

```
go test -tags gms_pure_go ./internal/storage/db/server/... -run Dial_BeforeStart -v
```

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3801"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/storage/db/server/doltserver_test.go
+++ b/internal/storage/db/server/doltserver_test.go
@@ -65,6 +65,10 @@ func newDoltServer(t *testing.T) (*server.DoltServer, string) {
 	log := filepath.Join(t.TempDir(), "server.log")
 	s, err := server.NewDoltServer(bin, rootDir, cfg, log, 0)
 	require.NoError(t, err)
+	// Close the log file handle before t.TempDir's RemoveAll runs.
+	// On Windows, an open handle prevents directory removal; Stop is
+	// idempotent so this is safe even when the test calls Stop itself.
+	t.Cleanup(func() { _ = s.Stop(context.Background()) })
 	return s, rootDir
 }
 


maphew:
**Triage: cherry-pick / reimplement onto current main.** Recommend REIMPLEMENT (small, mechanical port) - the underlying bug is still present on main; only the file path has moved.

## Reasoning

This PR is on the pre-rewrite (orphan) history, so it can't merge or rebase directly. I confirmed on `upstream/main` whether the fix already landed:

- The target file moved from `internal/storage/db/server/doltserver_test.go` (PR's base) to `internal/storage/dbproxy/server/doltserver_test.go` on main, but the `newDoltServer` test helper is otherwise byte-for-byte identical to the PR's pre-patch version - no `t.Cleanup` closing the server, no equivalent fix elsewhere.
- `TestDoltServer_Dial_BeforeStart` still exists on main (line 318) and still returns from `newDoltServer` without calling `Stop`, so the Windows `TempDir` `RemoveAll` file-lock failure described in #3798 is still live on main today.
- No other open or merged PR touches this: `gh pr list --search "3798 in:body"` and a title/body search both return only #3801 itself.

So this is real, still-needed contributor work, not superseded - it just needs porting to the new path.

## Findings

- **major (supersession check, cleared)**: Verified NOT already fixed on main. `internal/storage/dbproxy/server/doltserver_test.go` upstream/main, `newDoltServer` (lines ~59-69) has no cleanup registration; `TestDoltServer_Dial_BeforeStart` (line 318) calls `newDoltServer` and returns without `Stop`. Confirms #3798 is still reproducible.
- **minor**: The one-line diff must be re-applied against `internal/storage/dbproxy/server/doltserver_test.go` instead of the old `internal/storage/db/server/doltserver_test.go` path - trivial since the surrounding code is unchanged.
- **nit**: PR title/description still reference the pre-rewrite path (`internal/storage/db/server/...`); update wording when reimplementing so it matches the current tree.

## Remedy

Orphan history means `git rebase --onto` against the PR's own base won't work cleanly across the rewrite boundary. Simplest path: take the PR's one-hunk diff (add a `t.Cleanup(func() { _ = s.Stop(context.Background()) })` right after `require.NoError(t, err)` in `newDoltServer`) and cherry-pick/hand-apply it onto `internal/storage/dbproxy/server/doltserver_test.go` on current main, crediting @seanmartinsmith as author of the diagnosis, root-cause analysis, and fix. Test-only change, no production code, low risk.

Thanks to @seanmartinsmith for the careful Windows repro, root-cause writeup (server.log handle in `NewDoltServer`, `Stop`'s idempotency), and the minimal fix in #3798 / #3801 - this should land, just re-pointed at the moved file.


maphew:
Opened #4600 to carry this Windows test-hardening fix forward as part of the `mybd-8chd.1` batch.

Why a replacement PR instead of merging each stale branch directly: the batch has overlapping Windows/test scope, #3801 needed a path port after the Dolt server test moved to `internal/storage/dbproxy/server/doltserver_test.go`, and the release-script test needed a small Windows-hosted Bash harness adjustment while preserving the original fix. The batch commit includes:

`Co-authored-by: sms <114885497+seanmartinsmith@users.noreply.github.com>`

I’m leaving this PR open until #4600 lands; after that it can be closed as absorbed with contributor credit preserved.

maphew:
Good catch: the leaked log file handle in `newDoltServer` cleanup is invisible on Linux and a hard failure on Windows, and releasing it in cleanup was the right fix. Thank you for this and for the whole batch of Windows test-hardening PRs.

The fix is now live on `main`: commit 7865493f7 (PR #4600, merged 2026-07-07) ported all five of your PRs (#3797, #3801, #3802, #3806, #3812) together, with `Co-authored-by` credit to you. Your change was applied to the test file's new home — the package moved to `internal/storage/dbproxy/server/` in the interim, which is also why this branch stopped rebasing cleanly.

Since your fix is on main, this PR is due to be retired. I'm leaving it open for a few days in case you spot anything the port missed.

maphew:
Retiring as absorbed by #4600 — thanks again.

PR #3804, opened 2026-05-07
ci(nix): auto-update default.nix vendorHash on dependabot Go bumps

harry-miller-trimble:
## Summary

Dependabot Go-module bumps consistently fail the `nix build .#default` CI gate because `default.nix` has a hardcoded `vendorHash` that any `go.mod`/`go.sum` change invalidates. PRs sit stuck for days — currently affecting #3468, #3637, and #3638, all open ~13+ days vs. the historical hours-to-merge time on clean dependabot bumps. This workflow detects the drift, runs the existing `scripts/update-nix-vendorhash.sh` (which already implements the fake-hash → parse-error → patch → verify-rebuild dance), and pushes the hash bump back to the dependabot branch.

Tracked as `beads-bz5`.

### Why `pull_request_target`
GitHub gives the `GITHUB_TOKEN` **read-only** permissions on `pull_request` workflows triggered by dependabot, regardless of declared `permissions:`. `pull_request_target` runs in the trusted context of the base branch with full token permissions, which is required to push back to the PR branch.

### Safety against `pull_request_target` abuse
- The job only runs when **all three** are true:
  - `github.actor == 'dependabot[bot]'`
  - `github.event.pull_request.user.login == 'dependabot[bot]'`
  - `github.event.pull_request.head.repo.full_name == github.repository`
- Dependabot never opens PRs from forks, so this combination is not reachable from an external contributor.
- The workflow checks out the PR head SHA and runs `scripts/update-nix-vendorhash.sh` from that ref. Since dependabot is the only allowed actor and never modifies `scripts/`, the trust boundary is the dependabot bot itself — the same trust posture the project already accepts when merging dependabot PRs.
- No force-push. Single commit with attribution to `github-actions[bot]`.

### Why reuse `scripts/update-nix-vendorhash.sh`
The script already exists and implements the full update flow (fake-hash trigger, error parsing, sed update, verify-rebuild). Calling it from CI keeps a single source of truth — local maintainers running `./scripts/update-nix-vendorhash.sh` get identical behavior.

## Caveat

GitHub does not retrigger `pull_request` workflow runs for commits authored by `GITHUB_TOKEN`. After this workflow pushes the hash bump, the existing `nix-build.yml` check on the dependabot PR will remain stuck on the pre-fix commit. Two acceptable resolutions:

1. **Manual re-run** — a maintainer hits "Re-run failed jobs" on `nix-build` once. ~10 seconds of work per dependabot PR vs. the current "compute hash by hand" loop.
2. **This workflow's own success** is itself a green signal that the hash is correct (the workflow runs `nix build` to verify before pushing). Branch protection can include `update-vendor-hash` as an acceptable check.

A future follow-up could swap `GITHUB_TOKEN` for a fine-grained PAT to eliminate the manual re-run, but that's a separate maintenance trade-off.

## Test plan

- [x] YAML parses cleanly (`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/update-vendor-hash.yml'))"`)
- [x] Hash extraction logic verified offline against typical Nix `got: sha256-…` error format
- [x] Workflow uses pinned action SHAs matching the conventions of `.github/workflows/nix-build.yml` and `.github/workflows/ci.yml`
- [x] CONTRIBUTING.md updated to document the new workflow + caveat
- [ ] After merge: validate live by waiting for the next dependabot Go bump, or by manually rebasing one of #3468 / #3637 / #3638 to retrigger
- [ ] If validation fails, the workflow fails loudly (the verify-rebuild step in `update-nix-vendorhash.sh` re-runs `nix build` after applying the new hash) — no silent corruption path

## Related backlog

Once this lands and a dependabot Go bump triggers it once successfully:
- #3468 (anthropic-sdk-go 1.37→1.38) — should auto-fix on next push or rebase
- #3638 (go-sql-driver/mysql 1.9.3→1.10.0) — same
- #3637 (DeterminateSystems/update-flake-lock action) — github-actions ecosystem; doesn't touch go.sum, so likely a different blocker — needs separate triage

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3804"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


steveyegge (approved):
Clean fix for the dependabot vendorHash drift. The pull_request_target trigger is correctly guarded (runs only for dependabot[bot] as actor + PR author, non-fork heads), action SHAs are pinned, no force-push, and reusing scripts/update-nix-vendorhash.sh keeps CI and local behavior in sync. Merging.

PR #3803, opened 2026-05-07
fix: doltTransaction CloseIssue/AddLabel/RemoveLabel now emit events

seanmartinsmith:
Fixes #3799

## What

Three `doltTransaction` methods — `CloseIssue`, `AddLabel`, `RemoveLabel` — now correctly delegate to `issueops.*InTx` helpers and emit event rows, matching the behavior of the embedded path.

## Why

The dolt storage path implemented these three methods with raw SQL that wrote to the data table (issues/labels) but never touched the `events` table:

- `CloseIssue` (`internal/storage/dolt/transaction.go:597`): `UPDATE issues SET status = 'closed' ...` — no event row.
- `AddLabel` (`:712`): `INSERT IGNORE INTO labels ...` — no event row.
- `RemoveLabel` (`:747`): `DELETE FROM labels ...` — no event row.

`internal/storage/embeddeddolt/transaction.go` gets this right by delegating to issueops helpers (lines 74, 122, 127). The dolt path bypassed them. Dolt-backed deployments therefore lost audit-trail rows for all three operations silently — every close, label add, label remove went unrecorded in `events`.

**Root cause**: the issueops helpers (`CloseIssueInTx`, `AddLabelInTx`, `RemoveLabelInTx`) write both the data row and the event row atomically. The dolt path duplicated only the data-row SQL, so no event was emitted. The embedded path uses the helpers correctly.

**Fix**: replace each method body with delegation to the corresponding `issueops.*InTx` helper. Wisp routing is handled via `issueops.WispTableRouting` to mark both the data table (issues/wisps, labels/wisp_labels) and the event table (events/wisp_events) dirty after the helper returns. Net -28 LOC in `transaction.go` (delegation is more concise than the open-coded SQL).

## Verification

Three regression tests added to `internal/storage/dolt/events_test.go`:

- `TestCloseIssue_EmitsEvent` — close an issue, assert an `EventClosed` row appears in `events`
- `TestAddLabel_EmitsEvent` — add a label, assert an `EventLabelAdded` row
- `TestRemoveLabel_EmitsEvent` — add then remove a label, assert an `EventLabelRemoved` row

```bash
go test -tags gms_pure_go -run "TestCloseIssue_EmitsEvent|TestAddLabel_EmitsEvent|TestRemoveLabel_EmitsEvent" ./internal/storage/dolt/...
```

Tests require a Dolt test server; they skip cleanly without one. `make build` passes. Existing label/close tests in the package continue to pass.

[...]

PR #3808, opened 2026-05-07
fix: doltTransaction.DeleteIssue cascades to related tables (follow-up to #3803)

### LINKED ISSUE #3807: bug: doltTransaction.DeleteIssue skips cascade to related tables (storage parity gap, follow-up to #3803)
## Symptom

Deleting an issue via `doltTransaction.DeleteIssue` (e.g., inside `RunInTransaction`) leaves label, event, comment, and dependency rows orphaned in the database. Only the row in the `issues` (or `wisps`) table is removed.

## Repro

1. Create an issue via a Dolt-backed store.
2. Add a label to it.
3. Delete it via `store.RunInTransaction(ctx, "...", func(tx) { tx.DeleteIssue(ctx, id) })`.
4. Query `SELECT * FROM labels WHERE issue_id = '<id>'` — the label row is still present.
5. Same for `events`, `comments`, and `dependencies`.

The non-transactional `DoltStore.DeleteIssue` path is fine — it already delegates to `issueops.DeleteIssueInTx` (line 344 of `internal/storage/dolt/issues.go`).

## Root Cause

`doltTransaction.DeleteIssue` in `internal/storage/dolt/transaction.go` uses raw SQL:

```go
_, err := t.tx.ExecContext(ctx, fmt.Sprintf("DELETE FROM %s WHERE id = ?", table), id)
```

This deletes only the issue row. It does not call `issueops.DeleteIssueInTx`, which cascades to labels, events, comments, and dependencies. It also only marks the issue/wisps table dirty, so the Dolt version commit stages the wrong set of tables.

The canonical shape is at `internal/storage/embeddeddolt/transaction.go:78-84`:

```go
func (t *embeddedTransaction) DeleteIssue(ctx context.Context, id string) error {
    t.dirty.MarkDirty("issues")
    t.dirty.MarkDirty("dependencies")
    t.dirty.MarkDirty("labels")
    t.dirty.MarkDirty("comments")
    t.dirty.MarkDirty("events")
    return issueops.DeleteIssueInTx(ctx, t.tx, id)
}
```

Same structural gap as the three methods fixed in PR #3803 (CloseIssue / AddLabel / RemoveLabel). PR #3803 explicitly flagged this as a deferred follow-up in its "out-of-scope finding" section.

## Fix

Mirror the embedded path:

- Delegate to `issueops.DeleteIssueInTx(ctx, t.tx, id)`.
- Use `issueops.WispTableRouting(isWisp)` to get routed table names (issues/wisps, labels/wisp_labels, events/wisp_events, dependencies/wisp_dependencies).
- Mark comment table dirty directly (`"comments"` / `"wisp_comments"` — no WispTableRouting entry for comments, matching embedded behavior).
- Net LOC neutral (delegation shorter than open-coded SQL).

Regression test: `TestDeleteIssue_CascadesRelatedTables` in `internal/storage/dolt/events_test.go` — creates an issue, adds a label, deletes via `RunInTransaction`, asserts issue is gone and labels are cascade-cleaned.

seanmartinsmith:
Fixes #3807

## What

`doltTransaction.DeleteIssue` now delegates to `issueops.DeleteIssueInTx` and marks all five affected tables dirty, matching the embedded path at `internal/storage/embeddeddolt/transaction.go:78-84`.

## Why

The dolt transaction path used raw SQL that only deleted the issue row:

```go
_, err := t.tx.ExecContext(ctx, fmt.Sprintf("DELETE FROM %s WHERE id = ?", table), id)
```

No cascade to labels, events, comments, or dependencies. Label rows, event history, comments, and dependency edges were left orphaned whenever an issue was deleted inside a transaction. The non-transactional `DoltStore.DeleteIssue` (line 344 of `issues.go`) already correctly calls `issueops.DeleteIssueInTx` — the transaction path was the outlier.

This is the structural twin of PR #3803, which fixed `CloseIssue`, `AddLabel`, and `RemoveLabel`. That PR flagged `DeleteIssue` explicitly in its "out-of-scope finding" section; this PR closes the gap.

**Fix**: delegate to `issueops.DeleteIssueInTx`, use `issueops.WispTableRouting` to get routed table names, and mark all relevant tables dirty. Comment table is routed directly (`"comments"` / `"wisp_comments"`) since `WispTableRouting` has no comment-table entry — same as the embedded reference.

## Verification

New regression test in `internal/storage/dolt/events_test.go`:

- `TestDeleteIssue_CascadesRelatedTables` — creates an issue with a label, deletes via `RunInTransaction`, asserts the issue is gone and labels are cascade-cleaned. The raw-SQL path left labels orphaned; delegation to `DeleteIssueInTx` removes them.

```bash
go test -tags gms_pure_go -run "TestDeleteIssue_CascadesRelatedTables" ./internal/storage/dolt/...
```

Test skips cleanly without a Dolt server. `make build` and `go test -tags gms_pure_go ./internal/storage/dolt/...` both pass.

## Out-of-scope

No other methods in `transaction.go` were touched. Any further parity gaps found during review should be filed as separate beads.

Note: builds on top of #3803's fix shape; can land in either order without rebase since they touch different methods.

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3808"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/storage/dolt/events_test.go
+++ b/internal/storage/dolt/events_test.go
@@ -1,9 +1,11 @@
 package dolt
 
 import (
+	"errors"
 	"testing"
 	"time"
 
+	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/types"
 )
 
@@ -73,6 +75,59 @@ func TestGetAllEventsSince_UnionBothTables(t *testing.T) {
 			t.Errorf("events not in chronological order: [%d] %v > [%d] %v",
 				i-1, events[i-1].CreatedAt, i, events[i].CreatedAt)
 		}
+	}
+}
+
+// TestDeleteIssue_CascadesRelatedTables verifies that DeleteIssue via
+// RunInTransaction delegates to issueops.DeleteIssueInTx, which cascades to
+// related tables (labels, events, comments, dependencies). Regression guard
+// for the parity gap where the dolt path did a raw DELETE with no cascade,
+// mirroring the embedded path at internal/storage/embeddeddolt/transaction.go:78-84.
+func TestDeleteIssue_CascadesRelatedTables(t *testing.T) {
+	store, cleanup := setupTestStore(t)
+	defer cleanup()
+
+	ctx, cancel := testContext(t)
+	defer cancel()
+
+	issue := &types.Issue{
+		ID:        "ev-delete-1",
+		Title:     "Issue to delete",
+		Status:    types.StatusOpen,
+		Priority:  2,
+		IssueType: types.TypeTask,
+	}
+	if err := store.CreateIssue(ctx, issue, "tester"); err != nil {
+		t.Fatalf("CreateIssue: %v", err)
+	}
+
+	// Add a label so we can verify cascade cleanup happened
+	if err := store.AddLabel(ctx, issue.ID, "to-delete", "tester"); err != nil {
+		t.Fatalf("AddLabel: %v", err)
+	}
+
+	// Delete via transaction — exercises doltTransaction.DeleteIssue
+	err := store.RunInTransaction(ctx, "test: delete issue", func(tx storage.Transaction) error {
+		return tx.DeleteIssue(ctx, issue.ID)
+	})
+	if err != nil {
+		t.Fatalf("RunInTransaction/DeleteIssue: %v", err)
+	}
+
+	// Issue must be gone
+	_, getErr := store.GetIssue(ctx, issue.ID)
+	if !errors.Is(getErr, storage.ErrNotFound) {
+		t.Errorf("expected ErrNotFound after DeleteIssue, got %v", getErr)
+	}
+
+	// Labels must be cascade-cleaned — the raw-SQL path only deleted from issues,
+	// leaving label rows orphaned. Delegation to DeleteIssueInTx removes them.
+	labels, err := store.GetLabels(ctx, issue.ID)
+	if err != nil {
+		t.Fatalf("GetLabels after delete: %v", err)
+	}
+	if len(labels) != 0 {
+		t.Errorf("expected no labels after DeleteIssue, got %v (cascade did not run)", labels)
 	}
 }
 
--- a/internal/storage/dolt/transaction.go
+++ b/internal/storage/dolt/transaction.go
@@ -616,17 +616,21 @@ func (t *doltTransaction) CloseIssue(ctx context.Context, id string, reason stri
 
 // DeleteIssue deletes an issue within the transaction
 func (t *doltTransaction) DeleteIssue(ctx context.Context, id string) error {
-	table := "issues"
-	if t.isActiveWisp(ctx, id) {
-		table = "wisps"
+	isWisp := t.isActiveWisp(ctx, id)
+	issueTable, labelTable, eventTable, depTable := issueops.WispTableRouting(isWisp)
+	commentTable := "comments"
+	if isWisp {
+		commentTable = "wisp_comments"
 	}
-
-	//nolint:gosec // G201: table is hardcoded
-	_, err := t.tx.ExecContext(ctx, fmt.Sprintf("DELETE FROM %s WHERE id = ?", table), id)
-	if err == nil {
-		t.dirty.MarkDirty(table)
+	if err := issueops.DeleteIssueInTx(ctx, t.tx, id); err != nil {
+		return wrapExecError("delete issue in tx", err)
 	}
-	return wrapExecError("delete issue in tx", err)
+	t.dirty.MarkDirty(issueTable)
+	t.dirty.MarkDirty(depTable)
+	t.dirty.MarkDirty(labelTable)
+	t.dirty.MarkDirty(commentTable)
+	t.dirty.MarkDirty(eventTable)
+	return nil
 }
 
 // AddDependency adds a dependency within the transaction.


coffeegoddd (commented):
We need a minimal repro we can run independently demonstrating that deleting an issue does not cascade.

If this is the case, the fix is in the schema layer, databases handle cascades via foreign keys, not the application layer.

seanmartinsmith:
> We need a minimal repro we can run independently demonstrating that deleting an issue does not cascade.
> If this is the case, the fix is in the schema layer, databases handle cascades via foreign keys, not the application layer.

yeah, mis-diagnosed this. pointed at the right method but described the wrong reason - "cascade to related tables" doesn't hold for regular issues. dolt FK cascade in migrations 0002-0010 already handles labels/events/comments/dependencies(issue_id)/child_counters/etc. the PR's test uses a regular issue so it passes on old code, didn't actually catch anything.

real gap is wisps. `wisp_labels`, `wisp_events`, `wisp_comments` (migration 0021) have no FK constraints at all. repro:

```sql
insert into wisps (id, title, status, ephemeral) values ('w1', 'x', 'open', 1);
insert into wisp_labels (issue_id, label) values ('w1', 'bug');
delete from wisps where id = 'w1';
select count(*) from wisp_labels where issue_id = 'w1';  -- 1, orphaned
```

you're right that schema-layer is the cleaner fix here. happy to close this and open a migration adding `FOREIGN KEY (issue_id) REFERENCES wisps(id) ON DELETE CASCADE` on `wisp_labels`/`wisp_events`/`wisp_comments`, or keep this PR as an app-layer guard if you'd rather have both. preference?

maphew:
**Triage: RETIRE (already superseded on main).** Thanks @seanmartinsmith for the careful diagnosis, the regression test, and especially for coming back into the thread to correct your own framing when @coffeegoddd pushed on it - that kind of self-correction is exactly what keeps this project's bug reports trustworthy.

## Reasoning

This is an orphaned branch (no common ancestor with current `upstream/main`), so the job was to confirm whether the fix already landed. It has.

Current `upstream/main`, `internal/storage/dolt/transaction.go` lines 658-668:

```go
func (t *doltTransaction) DeleteIssue(ctx context.Context, id string) error {
	table := "issues"
	if t.isActiveWisp(ctx, id) {
		table = "wisps"
	}
	if err := issueops.DeleteIssueInTx(ctx, t.txFor(table), id); err != nil {
		return wrapExecError("delete issue in tx", err)
	}
	t.dirty.MarkDirty(table)
	return nil
}
```

This already delegates to `issueops.DeleteIssueInTx` (main: `internal/storage/issueops/delete.go:21`), the same fix PR #3808 proposes. The linked issue #3807 this PR claims to fix is stale against current main and should be closed as already resolved.

## Findings

- **major (supersession):** `doltTransaction.DeleteIssue` on `upstream/main` already delegates to `issueops.DeleteIssueInTx` - the exact behavior this PR introduces. No app-layer gap remains for regular issues; labels/events/comments/dependencies cascade via that shared helper.
- **minor (self-corrected in thread):** the contributor's own follow-up comment retracts the "cascades to related tables" framing as mis-diagnosed for regular issues (FK cascades in migrations 0002-0010 already cover them) and redirects to a real wisp-side gap (`wisp_labels`/`wisp_events`/`wisp_comments` lacking FK constraints, migration 0021).
- **nit (already tracked and closed):** that wisp-cascade concern was independently filed and investigated as #4169, then retracted and closed 2026-05-26 - `rileywhite` verified locally that Dolt's FK (`fk_wisp_events_issue ... ON DELETE CASCADE`) does fire on wisp delete, confirmed via a live `bd purge --force` run (17,431 -> 209 wisp_events, 8,212 -> 54 wisp_labels). So there is no outstanding cascade gap on either the regular-issue or wisp side today.

## Remedy

No cherry-pick needed - the target state is already on `main`. If a maintainer wants this branch's regression test (`TestDeleteIssue_CascadesRelatedTables` in `internal/storage/dolt/events_test.go`) preserved as a permanent guard against this class of regression, it can be lifted standalone via `git -C bd-main show refs/prtmp/3808:internal/storage/dolt/events_test.go` and applied against current main with credit to @seanmartinsmith - it still passes against the current delegating implementation and adds coverage codecov flagged as missing. Otherwise: close #3808 as superseded, close #3807 as already fixed on main, and thank the contributor for the diagnosis and self-correction in the thread.


maphew:
Thanks for the delete cascade fix. I am closing this orphaned pre-re-root PR because current `main` already delegates delete behavior through `issueops.DeleteIssueInTx`, so the cascade issue this branch addressed is already covered.

maphew:
@seanmartinsmith your delete-cascade diagnosis was right, and your five Windows fixes were right too. Main carries them with your name on them.

This PR got caught in a processing backeddy that left it languishing for weeks along with 6 other PRs even though you had responded to our concerns early. That's a lousy experience. Sorry! 

Two process changes landed in #5094 as a result: 

* Every review now starts with a prior-art pass over older open PRs and issues. An older PR takes precedence over a newer duplicate. 
* Rebase-only requests are now maintainer work, done on your branch instead of asked of you.

If anything in this PR wasn't actually covered by what landed, please speak up. 

Thanks for the work, and for the patience it shouldn't have taken.

-Matt

PR #3768, opened 2026-05-07
fix(show): bound bd show --json RAM by stripping heavy fields from Dependents

quad341:
## Problem

`bd show --json <id>` allocates 5–13 GB of RSS on hub beads (e.g. session beads or any bead with thousands of `dependency` edges pointing at it). The JSON output embeds full `Issue` records — including the heavy free-form fields `Description`, `Design`, `Notes`, `AcceptanceCriteria` — for every entry in `details.Dependents`.

The hot path in practice: an agent UserPromptSubmit hook (`gc nudge drain --inject`) shells out to `bd show --json <session-id>` on every turn. With many active agents, this triggered constant RAM pressure (system regularly at 80+ GB used, paging, with multiple bd processes individually holding 5–13 GB each).

The bug is backend-agnostic — same blow-up shape on Dolt and Postgres. It's purely a `cmd/bd/show.go` issue: every dependent's full `Issue` is materialized in Go before JSON marshaling.

## Fix

In `cmd/bd/show.go`'s `--json` path, replace the raw `GetDependentsWithMetadata(...)` result with a shallow projection that keeps only the identity-and-shape fields callers actually use:

- `ID`
- `Status`
- `IssueType`
- `Priority`
- `Title`
- (plus the outer `DependencyType`)

The heavy free-form fields are dropped. Epic-progress logic (which only reads `Status` + `DependencyType`) still works. Dependencies (the ones THIS bead depends ON, typically far fewer) are unchanged. Comments are unchanged.

Callers that need the full `Issue` for a specific dependent should fetch it explicitly via `bd show --json <dependent-id>`.

## Result

A/B measured live on a hub bead with 4 dependents:
- before: 6156 bytes
- after: 2514 bytes (**−59%**)

For hubs with thousands of dependents (the actual worst-case shape behind the 5–13 GB blowups) the savings scale linearly with `dependents × heavy-field-size`.

## Tests

Added `cmd/bd/show_shallow_dependents_test.go` (unit) covering:
- heavy fields are stripped (`Description`, `Design`, `Notes`, `AcceptanceCriteria` zeroed)
- identity-and-shape fields (`ID`, `Status`, `IssueType`, `Priority`, `Title`, `DependencyType`) are preserved
- `nil` and empty input return `nil`
- `nil` entries inside a non-empty slice are skipped

All `TestShow*` integration tests still pass under `gms_pure_go`.

## Out of scope (follow-up work)

[...]

PR #3813, opened 2026-05-08
fix(utils): be-szr ResolvePartialID falls through to GetIssue for wisps (PG)

quad341:
## Problem

`bd show <wisp-id>` and `gc mail peek <wisp-id>` return "no issue
found matching" / "bead not found" against Postgres-backed rigs after
migration.

`internal/utils/id_parser.go::ResolvePartialID`'s fast path uses
`store.SearchIssues` with an exact-ID filter. PG's `SearchIssues`
only consults the `issues` table, so full wisp IDs miss. PG's
`GetIssue` already falls through from issues→wisps on NotFound, but
the resolver wasn't reaching it for full IDs.

## Fix

After the `SearchIssues` fast path returns no rows, try
`store.GetIssue(input)`. PG's `GetIssue` handles the wisp fallback
transparently. Dolt is unaffected because its `GetIssue` resolves
wisps via the same store.

Cost: one extra indexed lookup only on the cold path (`SearchIssues`
miss), and only for callers passing full IDs.

## Verification

The PG regression test for this fix lives at
`internal/storage/postgres/wisp_resolve_test.go::TestResolvePartialIDForWispOnPG`
(build tag `integration_pg`) on PR #3768. It depends on
`startPG`/`openStore`/`storage.ConnectionConfig` from the be-6fk PG
testfixture, which is not yet on main, so the test was deliberately
not cherry-picked into this focused PR — it would not compile here.
The test will land with PR #3768 (or whichever PR brings the PG
testfixture to main).

Manual verification:
- Reverting this change in `id_parser.go` makes
  `TestResolvePartialIDForWispOnPG` fail on PR #3768 with the
  user-reported "no issue found matching" error.
- The existing Dolt-backed test
  `TestResolvePartialID_Wisp/full_wisp_ID` in
  `internal/utils/id_parser_test.go` still passes, confirming Dolt is
  unaffected.

## Out of scope

Partial wisp-ID resolution on PG (e.g., `bd show eu1` for
`bd-wisp-eu1`). The substring path in `ResolvePartialID` also goes
through `SearchIssues`, so partial wisp IDs still miss on PG.
Reported bug uses full IDs; partial wisp IDs is a separate
follow-up.

These two commits (fix + test) are also bundled in PR #3768 (a much
larger local-integ rebase). This focused PR exists so the fix can
land independently; PR #3768 will pick it up as already-merged on
rebase. The test stays in #3768 because it depends on the PG
testfixture also bundled there.

Refs: be-szr (root cause), be-h6xv (build bead).

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3813"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/utils/id_parser.go
+++ b/internal/utils/id_parser.go
@@ -50,6 +50,15 @@ func ResolvePartialID(ctx context.Context, store storage.Storage, input string) 
 		return issues[0].ID, nil
 	}
 
+	// Wisp fallback (be-szr): SearchIssues only consults the issues table,
+	// but on the Postgres backend wisps live in a separate table. GetIssue's
+	// PG implementation falls through to wisps on NotFound, so try it next
+	// for full IDs. This makes `bd show <wisp-id>` and `gc mail peek/read`
+	// work against PG-backed rigs.
+	if issue, err := store.GetIssue(ctx, input); err == nil && issue != nil {
+		return issue.ID, nil
+	}
+
 	// Get the configured prefix
 	prefix, err := store.GetConfig(ctx, "issue_prefix")
 	if err != nil || prefix == "" {


quad341:
**be-szr verification (rig investigator)**

Code-reading confirms the fix is correct and bounded:

- `internal/utils/id_parser.go:53-60` — new `store.GetIssue(input)` fallback after the `SearchIssues` fast path.
- `internal/storage/postgres/issues.go:144-159` — `PostgresStore.GetIssue` already falls through `issues` → `wisps` on `storage.ErrNotFound`, so the fallback finds mail wisps.
- `internal/storage/issueops/get_issue.go:16-38` — `GetIssueInTx` (used by Dolt) routes via `IsActiveWispInTx`, so Dolt finds wisps in the fast path and the new fallback is a no-op there.

Minor wording note on the PR description: PG's `SearchIssues` does call `searchTable` against the wisps table when `filter.Ephemeral` is nil (`postgres/issues.go:373-397`), but the per-row `(ephemeral = FALSE OR ephemeral IS NULL)` guard added at `searchTable` lines 462-470 filters out mail wisps (which carry `ephemeral=true`). So the fast-path miss is real, just for a slightly more specific reason than "only consults the issues table". Doesn't change the fix — just flagging in case the description is reused elsewhere.

**Filing follow-up** for the explicit out-of-scope item (partial wisp IDs on PG): the wisp fallback at `id_parser.go:167-196` calls `SearchIssues` with `Ephemeral=true`. PG's `SearchIssues` with `Ephemeral=true` searches the issues table only (with `ephemeral=true` clause → 0 rows), while Dolt's `SearchIssuesInTx` (`internal/storage/issueops/search.go:17-26`) routes Ephemeral=true to the wisps table. Mirroring Dolt's routing in PG's `SearchIssues` would close the gap. New investigation bead will carry the details.

be-szr closed as superseded by this PR.

coffeegoddd (commented):
We need a beads-only minimal repro of the bug, otherwise this is a bug with your orchestration layer.

quad341:
## beads/reviewer — PASS ✓

Correct PG-backend wisp fallback. Minimal, well-commented.

### Findings

None.

- **Correctness**: `SearchIssues` only queries the issues table; on PG, wisps live separately. `GetIssue`'s PG implementation falls through to wisps on NotFound — using it as a fallback is the right approach. ✓
- **Error handling**: `err == nil && issue != nil` guard is correct. ✓
- **Comment**: Explains the PG-specific constraint clearly for future readers. ✓

Ready to merge.

🤖 beads/reviewer

maphew:

maphew:
## Closing: superseded by the Postgres backend removal (#4857)

Thank you @quad341 for the careful diagnosis and the code-reading verification here — the analysis of the PG fast-path miss (wisps filtered out of `SearchIssues`, `PostgresStore.GetIssue` falling through issues → wisps) was sound for the codebase as it stood.

However, the premise of this fix no longer exists on `main`:

- Direct PostgreSQL (and MySQL) backend support was rolled back in #4857 (commit `0d7cf0022`, merged 2026-07-16): `internal/storage/postgres/` is gone from `main`, and `bd` now fails closed at startup when configured with a removed backend, before any resolver code runs.
- On the remaining backend (Dolt/embedded Dolt), partial-ID resolution already finds wisps in the fast path (`GetIssueInTx` routes through `IsActiveWispInTx`), so the `GetIssue` fallback this PR adds would never execute its intended path — it would be inert defensive code.
- Consequently the beads-only repro @coffeegoddd asked for can no longer be produced against stock beads at HEAD, and shouldn't remain a blocking ask; the honest resolution is to retire the PR rather than chase it.

If server-database backends return someday, the fallback pattern here (resolver falls through to `store.GetIssue` after a search miss) is worth resurrecting — this PR and its thread are the record of why.

Closing as superseded-by-backend-removal. Thanks again for the contribution and the sustained follow-through on this rig.

_claude-code-fable-5 on behalf of matt wilkie_

PR #3836, opened 2026-05-08
fix(dolt): skip DOLT_COMMIT when nothing staged after DOLT_ADD

ckumar1:
## Problem

`DoltStore.Commit()` queries `dolt_status` for dirty tables, stages them
via `CALL DOLT_ADD(table)`, then unconditionally calls `CALL DOLT_COMMIT`.

When all dirty tables are dolt-ignore'd (e.g. `wisps`, `wisp_*`), `DOLT_ADD`
fails silently for each one and nothing gets staged. `DOLT_COMMIT` then
returns `"nothing to commit"` — generating a Dolt server warning and paying
a full round-trip cost for every write that touches only ignored tables.

At scale with multiple agents doing frequent wisp writes (formula step
execution, patrol cycles, message routing), this floods the Dolt server
log and drives sustained high CPU.

## Fix

After the `DOLT_ADD` loop, check `dolt_status WHERE staged = 1` before
calling `DOLT_COMMIT`. If no staged changes exist, return `nil` immediately
without issuing the commit.

Fails open (attempts commit) when the staged-check query itself errors,
so any unexpected Dolt state still surfaces naturally.

## Observed impact

Gas Town operational setup (shared Dolt server, 3+ active agents):
- Before: ~37 unnecessary `DOLT_COMMIT`/sec → Dolt at ~290% CPU
- After: ~3–4/sec (write-only sessions, legitimate operations) → ~40% CPU

Pattern mirrors the existing `hasStagedChanges` guard in the Gas Town
daemon (`internal/daemon/dolt_remotes.go`, referenced in GH#3409 context).

## Test plan

- [ ] `go build -tags gms_pure_go ./...` passes
- [ ] `go test -tags gms_pure_go ./internal/storage/dolt/ -run TestCommit` passes (requires Dolt server)
- [ ] Dolt log no longer shows `nothing to commit` warnings for wisp-only writes

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3836"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -1618,6 +1618,18 @@ func (s *DoltStore) Commit(ctx context.Context, message string) (retErr error) {
 		}
 	}
 
+	// Skip DOLT_COMMIT if nothing was actually staged. dolt_status may include
+	// dolt_ignore'd tables (wisps, etc.) that fail DOLT_ADD silently above;
+	// calling DOLT_COMMIT with nothing staged generates a server warning and
+	// wastes CPU at scale (GH#3409).
+	var nStaged int
+	if err := conn.QueryRowContext(ctx, "SELECT COUNT(*) FROM dolt_status WHERE staged = 1").Scan(&nStaged); err != nil {
+		nStaged = 1 // fail open: attempt commit and surface any error naturally
+	}
+	if nStaged == 0 {
+		return nil
+	}
+
 	// NOTE: In SQL procedure mode, Dolt defaults author to the authenticated SQL user
 	// (e.g. root@localhost). Always pass an explicit author for deterministic history.
 	if _, err := conn.ExecContext(ctx, "CALL DOLT_COMMIT('-m', ?, '--author', ?)", message, s.commitAuthorString()); err != nil {


coffeegoddd (approved):
We need a minimal repro so we can confirm the issue. Once you supply that we can investigate the root cause and determine/advise on the correct fix.

coffeegoddd (changes requested):
Sorry I accidently clicked approve, its not approved.

maphew:

maphew:
Thanks for the `DOLT_COMMIT` no-op guard. I am closing this orphaned pre-re-root PR because current `main` already handles the same case in `DoltStore.Commit`: it checks staged rows before committing and also treats "nothing to commit" / "no changes" results as non-fatal, so there is no remaining repro-specific change to carry from this branch.

PR #3838, opened 2026-05-09
fix(hooks): skip export when export file is staged for deletion

ckumar1:
## Problem

`exportJSONLForCommit` in the pre-commit hook unconditionally exports
the JSONL file and then runs \`git add\` on it. If the user has staged
a deletion of the file with \`git rm\` (e.g. they decided to stop
tracking it), the hook re-creates the file and re-stages it, which
converts the staged deletion back to a modification — so the file
never actually gets removed from the repo.

Reproducer:

\`\`\`
git rm .beads/issues.jsonl
git commit -m \"drop tracked export\"
git status
# .beads/issues.jsonl reappears, modified instead of deleted
\`\`\`

I hit this trying to retire a tracked JSONL on one of our repos and
ended up having to disable the hook to delete the file.

## Change

In \`cmd/bd/hooks.go\`, before exporting, check whether the export
file is staged for deletion:

\`\`\`
git diff --cached --diff-filter=D --name-only -- <basename>
\`\`\`

If the basename appears in the staged-deletion list, skip the export
and the \`git add\` entirely. \`GIT_INDEX_FILE\` is already in the
environment during a real commit, so \`git diff --cached\` reads the
pending index.

12-line addition, no new dependencies, no behavior change for users
who aren't \`git rm\`'ing the export file.

## Not in scope

- Doesn't touch the export-on-write or export-on-merge paths.
- Doesn't change behavior when the file is staged for *modification*.
- Doesn't add a config key — staged deletion is an unambiguous user
  intent signal.

## Testing

- \`go build -tags gms_pure_go ./...\` passes.
- Manually verified the reproducer above: with the fix, \`git rm\`
  + commit cleanly removes the file; without it, the file resurrects.

Happy to add a unit test if there's a preferred shape for hook tests
in the repo (couldn't find an obvious existing pattern for the
pre-commit path).

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3838"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/cmd/bd/hooks.go
+++ b/cmd/bd/hooks.go
@@ -1077,6 +1077,18 @@ func exportJSONLForCommit() {
 	}
 	fullPath := filepath.Join(beadsDir, exportPath)
 
+	// If the export file is staged for deletion (user ran `git rm`), do not
+	// re-export or re-stage it. GIT_INDEX_FILE is set during an actual commit,
+	// so git-diff-index reads the pending index — where the deletion lives.
+	// Without this guard, git add fullPath would convert the staged deletion
+	// back to a modification and the file would never be removed from the repo.
+	checkCmd := exec.Command("git", "diff", "--cached", "--diff-filter=D", "--name-only", "--", filepath.Base(fullPath))
+	checkCmd.Dir = filepath.Dir(fullPath)
+	if out, _ := checkCmd.Output(); len(out) > 0 {
+		debug.Logf("pre-commit: %s staged for deletion — skipping export\n", exportPath)
+		return
+	}
+
 	debug.Logf("pre-commit: exporting JSONL to %s\n", fullPath)
 
 	// Shell out to `bd export` which initializes its own store.


maphew:
**Triage: cherry-pick / reimplement onto current main — the bug is real and still present upstream.**

## Why

This is an orphan branch (no common ancestor with `upstream/main`), so I searched main directly for the fix before trusting the PR. `exportJSONLForCommit` in current `cmd/bd/hooks.go` (upstream/main) still unconditionally re-exports and re-stages the tracked JSONL once `preCommitHasStagedBeadsFiles` sees *any* staged change under `.beads` — a staged `git rm` counts as a staged change under that pathspec, so the export runs, `bd export -o fullPath` rewrites the file to disk, and `gitAddFile` re-stages it via `git add`. That converts a staged deletion back into a modification, exactly as @ckumar1 describes. No later commit (through `b9088836c`, `d7d7a3460`, etc.) adds a diff-filter or staged-deletion guard. Supersession claim: **not superseded** — confirmed absent on `upstream/main:cmd/bd/hooks.go` around the `exportJSONLForCommit`/`preCommitHasStagedBeadsFiles` functions.

Thanks to @ckumar1 for the clear repro, the diagnosis of the `GIT_INDEX_FILE`/pending-index interaction, and for keeping the fix minimal (12 lines, no new deps). This is a real, still-present bug and the fix direction is correct.

## Findings

- **major** (supersession check): confirmed not fixed on main; needs reimplementation against current `hooks.go`, not a direct cherry-pick, since the surrounding function has been substantially refactored (`preCommitHasStagedBeadsFiles`, `exportSubprocessDir`, `hookWorkTreeRoot`, `warnJSONLWithoutDoltRemote` etc. did not exist when this PR was authored).
- **minor**: the PR's check re-shells `git diff --cached --diff-filter=D` against `filepath.Base(fullPath)` run from `filepath.Dir(fullPath)` — on current main this should instead reuse the existing `preCommitHasStagedBeadsFiles`/`exportSubprocessDir`/`hookWorkTreeRoot` plumbing (env scrubbing, worktree-root resolution) rather than a fresh, narrower `exec.Command`, or it will silently misbehave under the redirect/worktree cases those helpers exist to handle.
- **nit**: no unit test included; author explicitly offered to add one if a hook-test pattern exists — worth pointing them (or a reimplementer) at whatever pattern has emerged in the hook tests since.

## Recommendation

Retire this PR as unmergeable (orphan history), but reimplement the fix fresh against current main in a new PR/bead, crediting @ckumar1 for the diagnosis, repro, and design. Remedy note for anyone tempted to salvage the branch directly: `git rebase --onto upstream/main <pre-rewrite-merge-base> refs/prtmp/3838` will not work cleanly here because there is no shared merge-base at all — a manual reapply of the `diff-filter=D` guard into the current `exportJSONLForCommit`/`gitAddFile` path is required instead of an automated rebase.


maphew:
Thanks for this, @ckumar1 - the bug is real and the fix is the right shape.

This branch predates the late-May-2026 history re-root of `main`, so it has no common ancestor with current `main` and can't be rebased/merged in place. I've carried your fix and analysis forward onto current `main` in #4579, with a unit-testable helper and behavioral tests (staged deletion, untouched file, and staged-modification cases). Your authorship is preserved via `Co-authored-by`.

Leaving this open for now; I'll close it once #4579 lands so nothing is lost if that needs another pass. No action needed on your side.

_claude-code-opus-4-8-high on behalf of matt wilkie_

maphew:
The reimplementation landed: #4579 is merged to `main` (2026-07-05).

Your fix and bug analysis are carried forward with `Co-authored-by` credit, plus behavioral tests for the staged-deletion, untouched-file, and staged-modification cases. Closing this one as superseded since its branch predates the history re-root and can't merge in place. Thanks again @ckumar1.

_claude-code-opus-4-8-high on behalf of matt wilkie_

maphew:
Correcting the record on the two comments above: the "late-May-2026 history re-root" claim was wrong. No re-root of `main` ever happened - that conclusion came from forensics run inside a shallow clone, whose depth boundary was misread as severed history. This branch has a normal merge-base with `main`; it was simply stale and conflicting.

The outcome stands unchanged: your bug analysis and fix were correct and are merged to `main` via #4579 with `Co-authored-by` credit. Only the stated reason for reimplementing (rather than rebasing this branch) was misattributed - it was ordinary branch staleness, not a history rewrite. Apologies for the confusing framing, @ckumar1.

PR #3840, opened 2026-05-09
feat(init): default to unix socket transport when /tmp/mysql.sock is present

athosmartins:
## Summary

`bd init` now probes the standard Dolt unix socket path (`/tmp/mysql.sock`) and adopts it as the default `dolt_server_socket` in the new `metadata.json` when:

1. the user did not pass `--socket` explicitly,
2. this is a fresh init (not an existing-config preserve), and
3. the path is currently a listening unix socket.

New rigs/clones inherit the correct transport from the start instead of needing a post-init fixer or `BEADS_DOLT_SERVER_SOCKET` environment variable.

**Conservative semantics:** if the socket is not present at init time, falls back to the existing TCP behavior — no breakage for setups that don't have a local Dolt server on the default path. Users with custom paths can still pass `--socket /path` explicitly; the explicit flag wins. Existing-config inits do not have the default applied (preserve mode is honored).

## Why

Short-lived `bd` invocations over TCP loopback create a `TIME_WAIT` entry per close that lingers ~30s on macOS (`2 * MSL` with `MSL=15s`). On busy rigs with many crew sessions + mayor + refinery + polecats + daemons all running `bd` cycles, the `TIME_WAIT` count climbs past port-monitor alert thresholds (downstream incident: peak 280 sockets on port 3307, `ntfy` alert spam, port-exhaustion risk). Unix-socket transport bypasses `TIME_WAIT` entirely.

Setting the key at `bd init` means new metadata files get the correct transport without each clone owner needing to remember a post-init fixer.

## Implementation

- New helper `detectDefaultDoltSocket()` in `cmd/bd/init.go`. Returns `/tmp/mysql.sock` if it exists and is a unix domain socket; `""` otherwise. Uses `os.Stat().Mode()&os.ModeSocket` to check.
- New unexported `defaultDoltSocketPath` var (not const) so tests can swap it for a temp socket without depending on a real Dolt.
- `init.go` tracks `freshInit := existingCfg == nil` and applies the default only on fresh inits when no explicit flag is set and the field is empty.

Diff is small and isolated to `cmd/bd/init.go` plus a new test file.

## Tests

`cmd/bd/init_default_socket_test.go` covers:

- `TestDetectDefaultDoltSocket_Present` — listener on a temp socket → returns the path.
- `TestDetectDefaultDoltSocket_Absent` — path does not exist → returns `""`.
- `TestDetectDefaultDoltSocket_RegularFile` — path is a regular file → returns `""` (guards against false positives from leftover non-socket files).
- `TestDetectDefaultDoltSocket_Directory` — path is a directory → returns `""` (defense in depth).

The Present test uses `MkdirTemp(\"/tmp\", \"bds\") + filepath.Join(..., \"s.sock\")` to keep the path under macOS's 104-char `sun_path` limit. Skipped on Windows. All four tests run with `CGO_ENABLED=0` and do not require a Dolt instance.

```
$ CGO_ENABLED=0 go test ./cmd/bd/ -run TestDetectDefaultDoltSocket -count=1
=== RUN   TestDetectDefaultDoltSocket_Present
--- PASS
=== RUN   TestDetectDefaultDoltSocket_Absent
--- PASS
=== RUN   TestDetectDefaultDoltSocket_RegularFile
--- PASS
=== RUN   TestDetectDefaultDoltSocket_Directory
--- PASS
PASS
```

Note: full-package build / existing test suite was not verified locally due to an unrelated CGO/ICU header issue on my Mac (`unicode/regex.h not found`); CI will validate on a fresh runner.

## Refs

Downstream tracking bead: `dc-fsue` — applied a live fixer (`scripts/setup-bd-socket.sh`) to existing `~/gt/*/.beads/metadata.json` files; this PR closes the loop on the upstream side so future rigs do not need that fixer.

🤖 Generated with Gas Town crew workflow

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3840"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -28,6 +28,39 @@ import (
 	"golang.org/x/term"
 )
 
+// defaultDoltSocketPath is the standard path Dolt's sql-server uses for its
+// unix domain socket. Declared as a var (not const) so init tests can swap
+// it for a temp-dir socket without depending on a real Dolt instance.
+var defaultDoltSocketPath = "/tmp/mysql.sock"
+
+// detectDefaultDoltSocket returns the standard Dolt unix socket path if a
+// unix socket is currently listening there, otherwise returns "". Used by
+// `bd init` to opt new beads dirs into unix-socket transport when the
+// machine clearly has Dolt running locally.
+//
+// Rationale: short-lived `bd` invocations over TCP loopback create a
+// TIME_WAIT entry per close that lingers ~30s on macOS (2*MSL with
+// MSL=15s). Across crew sessions, mayor, refinery, polecats, and daemons
+// the count climbs past port-monitor alert thresholds and risks port
+// exhaustion. Unix-socket transport bypasses TIME_WAIT entirely. Setting
+// the key at init time means new rigs/clones inherit the correct config
+// from the start instead of needing a post-init fixer script.
+//
+// Conservative semantics: only sets when the socket is *currently
+// present* at init time. Machines without a running Dolt server (or with
+// a non-default socket path) fall through to the existing TCP behavior.
+// Users with a custom path can still pass `--socket` explicitly.
+func detectDefaultDoltSocket() string {
+	info, err := os.Stat(defaultDoltSocketPath)
+	if err != nil {
+		return ""
+	}
+	if info.Mode()&os.ModeSocket == 0 {
+		return ""
+	}
+	return defaultDoltSocketPath
+}
+
 var initCmd = &cobra.Command{
 	Use:     "init",
 	GroupID: "setup",
@@ -857,6 +890,8 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 				fmt.Fprintf(os.Stderr, "Warning: failed to load existing metadata.json: %v\n", err)
 			}
 
+			freshInit := existingCfg == nil
+
 			var cfg *configfile.Config
 			if existingCfg != nil {
 				// Preserve existing config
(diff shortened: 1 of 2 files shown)


maphew:
Concerns / things to think about

1. initTimeCloneConfig (line 1948) — this function also sets DoltServerSocket but only when serverSocket != "". It doesn't get the same auto-detection. This is likely intentional since clone-init-with-server-mode explicitly passes server config, but worth confirming that clones in server mode aren't also suffering TIME_WAIT issues. If they are, this function could also benefit from the same socket detection.

2. Hardcoded /tmp/mysql.sock — Dolt's default, but if a user runs Dolt with a custom --socket path, this won't detect it. The PR handles this correctly (explicit --socket still wins), but it's worth documenting in release notes so users with non-standard socket paths aren't confused why they don't get the benefit automatically.

3. No Linux consideration mentioned — Linux also has MSL and TIME_WAIT behavior, though typically with shorter MSL (default net.ipv4.tcp_fin_timeout is 60s but the practical impact is usually less severe). The detection works fine on any unix system with a socket at that path — it's just that the motivation is macOS-centric.

4. Race condition edge case: The socket could theoretically disappear between the os.Stat check and when bd actually tries to connect. This is fine — the connection will just fail, and Dolt fallback behavior handles it. Not a correctness issue, just worth noting.

Otherwise LGTM. The only concern I'd address is (2), documenting so folks aren't surprised.

maphew:
_2nd review agent says (1) is also a concern:_

Yes, clones in server mode are also suffering TIME_WAIT issues. 

When bd init --server --remote <url> runs, the clone connection uses TCP — not the unix socket — even when /tmp/mysql.sock is present.

The gap is that initTimeCloneConfig doesn't get the same socket auto-detection treatment. The clone itself — which can be a long-lived but still single-use connection — goes over TCP. And for rigs that do this repeatedly (mayor, refinery, polecats spinning up fresh clones), each clone contributes a TIME_WAIT.

The fix would be straightforward: call detectDefaultDoltSocket() inside initTimeCloneConfig (or better, at the call site before building cloneCfg) and pass the detected socket through, mirroring what the PR does for the metadata.json path.

coffeegoddd:
See comment from here, applies equally to this pr https://github.com/gastownhall/beads/pull/3847

maphew:
I defer to coffeegoddd's expertise here

athosmartins:
Closing — tied to #3847's fate; same need for a beads-only repro to substantiate the TCP-vs-socket latency premise. Will reopen alongside #3847 once repro exists.

PR #3847, opened 2026-05-09
fix(storage): default ServerDSN to /tmp/mysql.sock for loopback hosts

athosmartins:
## Summary

Without an explicit `Socket` and with a loopback `Host`, `ServerDSN.String()` produced a TCP DSN even when a Dolt server was listening on the default unix socket at `/tmp/mysql.sock`. Each TCP close on macOS leaves a TIME_WAIT entry lasting ~30s (2\*MSL with MSL=15s); on busy rigs the count climbs past port-monitor alert thresholds and risks port exhaustion.

## Repro

```
$ ls -la /tmp/mysql.sock
srwxr-xr-x  1 athos  wheel  0  /tmp/mysql.sock     # Dolt is listening here

# rig metadata.json without an explicit dolt_server_socket entry:
$ cat .beads/metadata.json
{
  "dolt_server_host": "127.0.0.1",
  "dolt_server_port": 3307,
  ...
}

# bd connections from this rig:
$ lsof -i :3307 -nP | grep ESTABLISHED
gt   91638  TCP localhost:51399->localhost:3307 (ESTABLISHED)
gt   91638  TCP localhost:51424->localhost:3307 (ESTABLISHED)
... (7 ESTABLISHED + ~50 TIME_WAIT)
```

## Root cause

`ServerDSN.String()` (and the parallel `DoltServerDSN.String()` in `internal/storage/db/util/dsn.go`) defaulted `net = "tcp"` and only switched to `"unix"` if `Socket` was non-empty. Rigs whose `.beads/metadata.json` predated the `dolt_server_socket` field built TCP DSNs against the local Dolt server.

## Fix

A new `resolveLocalSocket` helper applies a conservative fallback at the DSN-builder layer:

* Explicit `Socket` always wins (no behaviour change).
* When `Socket` is empty AND host is loopback (empty / `127.0.0.1` / `localhost` / `::1`) AND a unix socket exists at `/tmp/mysql.sock`, the DSN switches to `@unix(...)`.
* Non-loopback hosts and Windows / no-socket setups keep their TCP semantics.

Mirrored across both DSN builders since they're sibling packages with no import path between them.

## Test plan

- [x] `go test ./internal/storage/doltutil/ ./internal/storage/db/util/` clean
- [x] 9 new cases per package (explicit-socket-wins, loopback hosts → fallback, non-loopback → TCP, no-socket → TCP, helper unit tests)
- [x] Pre-existing `TestServerDSN_TCPFallbackWithoutSocket` updated to use a remote host so it remains deterministic on dev machines where `/tmp/mysql.sock` exists
- [x] Existing TLS test cases pass unchanged

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3847"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/storage/db/util/dsn.go
+++ b/internal/storage/db/util/dsn.go
@@ -2,10 +2,46 @@ package util
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	mysql "github.com/go-sql-driver/mysql"
 )
+
+// defaultDoltSocketPath is Dolt's source default unix socket path. Mirrored
+// from internal/storage/doltutil/dsn.go (the two packages are siblings and
+// would form a cycle if either imported the other).
+const defaultDoltSocketPath = "/tmp/mysql.sock"
+
+// localSocketPath is a var (not const) so unit tests can swap it for a temp-dir
+// socket without depending on a real Dolt server.
+var localSocketPath = func() string {
+	info, err := os.Stat(defaultDoltSocketPath)
+	if err != nil {
+		return ""
+	}
+	if info.Mode()&os.ModeSocket == 0 {
+		return ""
+	}
+	return defaultDoltSocketPath
+}
+
+func isLocalLoopback(host string) bool {
+	return host == "" || host == "127.0.0.1" || host == "localhost" || host == "::1"
+}
+
+// resolveLocalSocket returns the local Dolt unix socket path when the DSN
+// would otherwise dial loopback TCP and a unix socket exists. See
+// internal/storage/doltutil/dsn.go for the rationale (gt-tzm0t).
+func resolveLocalSocket(socket, host string) string {
+	if socket != "" {
+		return socket
+	}
+	if !isLocalLoopback(host) {
+		return ""
+	}
+	return localSocketPath()
+}
 
 type DoltServerDSN struct {
 	Socket      string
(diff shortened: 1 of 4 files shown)


coffeegoddd (commented):
This analysis of the issue looks like hallucinated nonsense, and the repro is not actually a reproduction anyone can run themselves. 

We need a beads-only reproduction to confirm there is a bug in beads, versus the orchestrator you're using.

Concretely this means: a simple repro in beads only that shows a command taking 30s, and a follow up repro that uses a socket instead of the tcp connection on the same beads repo, demonstrating the latency is gone.

Once we have those, we need to investigate the root cause of the latency, tcp vs socket smells wrong.

athosmartins:
Closing pending a beads-only TIME_WAIT repro, per maintainer ask. Will reopen with a minimal repro and root-cause analysis (separating orchestrator factors) if the issue still reproduces in pure beads.

PR #3854, opened 2026-05-10
test(regression): make workspace cleanup robust

maphew:
## Summary
- replace regression harness `t.TempDir` workspaces with explicit temp dirs and retrying cleanup
- keep baseline daemon cleanup ordered before workspace removal
- disable legacy baseline daemon startup through both old and current no-daemon env vars

## Why
`main` is failing in `Regression Tests` because otherwise-passing tests intermittently fail during Go tempdir cleanup with `.beads: directory not empty`. The failure is a cleanup race from short-lived background writes in the regression workspace, not a scenario assertion failure.

## Validation
- `go test -tags=regression ./tests/regression`
- `go test -tags='regression,gms_pure_go' ./tests/regression`
- Attempted full WSL regression run, but local WSL Docker is unavailable so it skipped the shared Dolt test server and later hung; relying on GitHub Actions for the real Linux/Docker regression run.

### DIFF
--- a/tests/regression/regression_test.go
+++ b/tests/regression/regression_test.go
@@ -249,7 +249,13 @@ type workspace struct {
 
 func newWorkspace(t *testing.T, bdPath string) *workspace {
 	t.Helper()
-	dir := t.TempDir()
+	dir, err := os.MkdirTemp("", "bd-regression-workspace-*")
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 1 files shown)


maphew:
Manual validation: dispatched \\Regression Tests\\ on maphew/beads@fix/regression-tempdir-cleanup and it passed: https://github.com/maphew/beads/actions/runs/25620698005

PR #3475, opened 2026-04-25
feat(telemetry): adopt standard OTel SDK env vars, wire WrapStorage, partition bd.* by bd.prefix

GraemeF:
Hello, human here! :wave: I wanted to track ready/in progress/closed bead counts by project in my otel stack, so I added some attributes to bd's existing telemetry. While there I also tidied up the configuration so it uses the standard `OTEL_*` env vars, while preserving compatibility with the existing `BD_OTEL_*` equivalents.

Over to :robot: ...

## Summary

Three related changes:

1. Telemetry configuration moves to the standard OpenTelemetry SDK environment variables, behind an explicit `BD_OTEL_ENABLED=true` opt-in. Legacy `BD_OTEL_*` variables continue to activate telemetry on their own and are translated into their `OTEL_*` equivalents with a stderr deprecation warning.
2. `WrapStorage`, the OTel storage decorator, is wired into the storage chain. The decorator was implemented but never instantiated, so `bd.storage.*` metrics, `bd.issue.count` snapshots, and per-operation storage spans were not reaching the backend even with telemetry enabled.
3. The configured issue prefix is stamped as a `bd.prefix` attribute on the resource AND on every emitted metric measurement, so `bd_*` series in a Prometheus-style backend carry `bd_prefix` as a first-class label without requiring a `target_info` join.

## Activation

`BD_OTEL_ENABLED=true` is the master switch. Standard `OTEL_*` variables on their own do not activate telemetry — they configure it once an opt-in is in effect. This is deliberate: a machine-global `OTEL_EXPORTER_OTLP_ENDPOINT` set for an unrelated instrumented process must not silently turn `bd` telemetry on.

The legacy variables `BD_OTEL_METRICS_URL`, `BD_OTEL_LOGS_URL`, and `BD_OTEL_STDOUT=true` also activate telemetry on their own, for back-compat.

`OTEL_SDK_DISABLED=true` forces telemetry off even when an opt-in is set, matching SDK conventions.

## Standard SDK env vars honored

Once activated, the SDK's standard variables apply:

| Variable | Effect |
|---|---|
| `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` | OTLP HTTP metrics endpoint |
| `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` | OTLP HTTP logs endpoint (recognized; not yet consumed by a logger provider) |
| `OTEL_TRACES_EXPORTER=console` | Stderr trace exporter |
[...]

PR #3858, opened 2026-05-10
fix(telemetry): wire WrapStorage decorator into bd storage chain

GraemeF:
## Context

Carved out of #3475 at @coffeegoddd's request — the bug fix as a standalone PR with a regression test and a local repro example. Subsequent PRs will tackle the standard `OTEL_*` env-var work and the `bd.prefix` resource/per-measurement attribute, both of which depend on this landing first.

## The bug

`InstrumentedStorage` (in `internal/telemetry/storage.go`) is fully implemented but never instantiated. `bd.storage.*` counters, `bd.issue.count` snapshots, and per-operation storage spans never reach any backend — even when telemetry is enabled and an exporter is wired up. Only `bd.db.pool_*` gauges (registered separately inside the dolt store) were emitted.

## What this PR does

- Wires `telemetry.WrapStorage` into the storage decorator chain in `cmd/bd/main.go`, between the existing `telemetry.Init` and `NewHookFiringStore` so storage spans measure pure DB time without hook-firing overhead.
- Changes `WrapStorage` to operate on `storage.DoltStorage` (rather than the smaller `storage.Storage` interface) — the `store` variable's type, and what `NewHookFiringStore` already takes. `InstrumentedStorage` embeds `DoltStorage` so capability sub-interfaces (`VersionControl`, `HistoryViewer`, `SyncStore`, …) pass through unchanged; those already have their own `dolt.*` spans, so wrapping them here would double-count.
- Renames `HookFiringStore.Inner()` → `Unwrap()` and introduces an `Unwrapper` interface so `storage.UnwrapStore` peels arbitrary decorator chains. Without this, adding a second decorator silently breaks the ~25 `cmd/bd` type assertions to optional interfaces (`StoreLocator`, `BackupStore`, `RawDBAccessor`, …). Internal package, no external callers.
- Extracts chain composition into `wireStorageDecorators` in a new `cmd/bd/storage_chain.go` so the wiring itself is unit-testable.

## Regression test

`TestWireStorageDecorators_TelemetryOn_HookOn` in `cmd/bd/storage_chain_test.go` is the direct regression test: it asserts the full `HookFiringStore → InstrumentedStorage → raw` chain via type assertions and `Unwrap`. If a future refactor drops the `WrapStorage` call from the helper, the test fails. Five further cases in the same file cover the cartesian product of telemetry on/off × hooks on/off/nil-runner, plus `TestWrapStorage_PeelsViaUnwrapStore` confirms decorator chains stay peelable.

## Local repro

`docs/OBSERVABILITY.md` gets a snippet showing how to confirm the wiring without standing up a collector:

```bash
BD_OTEL_STDOUT=true bd list 2>&1 | grep -F bd.storage.operations
```

Verified locally — `bd.storage.operations` now appears with multiple data points (e.g. `GetConfig`, `SearchIssues`) where previously the line was absent entirely.

## Out of scope

- Standard `OTEL_*` SDK env vars + `BD_OTEL_*` back-compat translation
- `bd.prefix` on the resource and as a per-measurement attribute
- Init signature changes around the prefix parameter

Those are coming in follow-ups.


maphew:
Merged - the WrapStorage wiring gap was real and still present on main; validated stacked on current main (build + decorator-wiring tests + telemetry/storage packages) before landing. Part two of your pair (#3859) is being rebased next. Thanks for both, and for the test coverage that made this easy to verify.

PR #3859, opened 2026-05-10
refactor(telemetry): adopt standard OTel SDK env vars with BD_OTEL_* back-compat

GraemeF:
## Summary

Adopt the standard OpenTelemetry SDK environment variables for bd telemetry configuration, with backwards-compatible translation from the existing `BD_OTEL_*` variables.

This is **PR 2 of 3** in the split of #3475 requested in coffeegoddd's review. PR 1 (#3858) was the WrapStorage bug fix; the bd.prefix attribute work follows in PR 3.

### What changes

- **Master switch**: `BD_OTEL_ENABLED=true` activates telemetry. A standalone `OTEL_*` variable does **not** turn telemetry on — a machine-global `OTEL_EXPORTER_OTLP_ENDPOINT` set for some other instrumented tool can no longer silently redirect bd telemetry.
- **Standard SDK env vars**: `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`, `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`, `OTEL_TRACES_EXPORTER=console`, `OTEL_METRICS_EXPORTER=console`, `OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`, `OTEL_SDK_DISABLED`. Endpoint discovery is delegated to the SDK; the explicit URL helper in `internal/telemetry/otlp.go` is removed.
- **Resource enrichment**: `resource.WithFromEnv()` is applied so users can override `service.name` or stamp `deployment.environment` via `OTEL_RESOURCE_ATTRIBUTES` without forking.
- **Back-compat**: `BD_OTEL_METRICS_URL`, `BD_OTEL_LOGS_URL`, `BD_OTEL_STDOUT=true` continue to activate telemetry on their own and translate to the new `OTEL_*` slots. A legacy value wins unconditionally over a pre-existing standard value, and each invocation logs a one-line deprecation warning to stderr.
- **`cmd/bd/main.go` extraction**: the cobra `PersistentPreRunE` telemetry wiring (`telemetry.Init` + `bd.command.<name>` span construction) moves to `cmd/bd/command_telemetry.go` so it can be unit-tested. Same regression class as the WrapStorage bug — wiring buried in `PreRunE` decays silently — so it gets the same treatment.

### Tests

`internal/telemetry/telemetry_test.go` (new, 318 lines):

- `translateLegacyEnv`: no-op when nothing set, each legacy var translates correctly, **legacy value wins** when both legacy and standard are set.
- `Enabled`: master-switch activates; `OTEL_*` alone does **not** activate; legacy vars activate on their own; `OTEL_SDK_DISABLED=true` overrides everything.
- `buildResource`: default `service.name`, `OTEL_SERVICE_NAME` overrides, `OTEL_RESOURCE_ATTRIBUTES` merges in via `WithFromEnv`.
- `Init`: no-op when disabled, legacy env activates + translates + emits stderr warning, `OTEL_TRACES_EXPORTER=console` + `OTEL_METRICS_EXPORTER=console` wire up successfully.
- `otlpMetricsEndpointSet`: detects metrics-specific endpoint, generic OTLP endpoint, and the negative case.

`cmd/bd/command_telemetry_test.go` (new, 84 lines):

- `commandSpanAttrs` returns the documented attribute set; empty args produce empty `bd.args` rather than panicking.
- `startCommandTelemetry` returns non-nil context and span even when telemetry is disabled — a noop span is still required for trace-nesting downstream.

### Verified locally

```
nix develop --command go test -tags gms_pure_go ./internal/telemetry/... ./cmd/bd/...
```

All green; `gofmt -l` clean; `go vet` clean.

### Out of scope

- The `bd.prefix` resource attribute and per-measurement metric stamping — separate follow-up PR.
- The WrapStorage wiring fix — already shipped as #3858 (PR 1).


maphew:
Maintainer fix-merge (b2e652516): merged current main into this branch, right after landing your #3858.

The one real conflict was cmd/bd/main.go, where two things had changed under this PR: #3858's wireStorageDecorators call (carried over untouched) and a usage-metrics bootstrap block main inserted BETWEEN the two halves of the telemetry setup, with an early return for the internal send-metrics subcommand. A single combined startCommandTelemetry() call could not sit at either point without changing that subcommand's behavior (Init always ran; the root span started only after the early return). So command_telemetry.go now exposes the two halves - initTelemetry() and startCommandSpan() - called at exactly the positions main had inline, with your combined startCommandTelemetry() kept for its wiring-contract test. Behavior parity with main preserved; your env-var precedence is unchanged (OTEL_SDK_DISABLED always wins; activation stays opt-in via BD_OTEL_ENABLED or legacy selectors; legacy BD_OTEL_* translate to OTEL_* with a deprecation warning).

Also deduplicated the clearTelemetryEnv test helper that #3858 and this PR each carried (kept the OTEL_*-aware one).

Note for reviewers: the merge commit def9f97a4 alone does not build (a file was left unstaged); the follow-up b2e652516 completes it. Squash-merge makes that moot on main.

Validated locally: build, vet, 18 cmd/bd telemetry tests + 21 internal/telemetry tests all green. Will merge once CI is green. Thanks @GraemeF - the pair is nearly home.

maphew:
Hi @GraemeF — checking back in on this pair. Quick status and a proposal.

**Where things stand:** I fix-merged current `main` into this branch back on 2026-07-07 (right after landing #3858) so it would build and pass CI — see my comment above for the details of that merge. Since then there's been no further activity from either side, and `main` has moved another ~580 commits, so the branch is conflicting again and CI at the current head is red (mostly stale-base fallout: Build Artifacts, Lint, and the PR policy/wrapper-timing checks).

A few concrete things that moved underneath this PR since the July fix-merge, in case they affect how you'd want to rebase:

- **`DO_NOT_TRACK` opt-out landed** (#4938) — a new metrics-opt-out env var in the same neighborhood of `cmd/bd/main.go` as your `OTEL_SDK_DISABLED` / `BD_OTEL_ENABLED` activation logic. Distinct mechanism (it aliases `BD_DISABLE_METRICS`, not OTel activation), but worth a glance so the precedence story stays coherent for users setting both.
- **`docs/OBSERVABILITY.md` moved** to `docs/reference/observability.md` as part of a Mintlify docs-site port (#4722) — the doc updates here will need to retarget that path.
- **`cmd/bd/main.go` has seen heavy unrelated churn** (backend selection, proxied-server commands) around the telemetry wiring, though the two helpers I split out in the fix-merge — `initTelemetry()` and `startCommandSpan()` — are still there unchanged and are the right seam to rebase onto rather than main's old inline `PreRunE` block.

**On #3861:** since it builds directly on this PR's `telemetry.Init` signature and the `initTelemetry` / `startCommandSpan` split, and both PRs conflict on the same `cmd/bd/main.go` region, I'd suggest folding them into one rebased series rather than resolving the same conflict twice — this PR's env-var refactor first, then #3861's `bd.prefix` stamping on top. Happy to do that rebase and push to your branches if you enable "allow edits from maintainers" on both PRs; if you'd rather drive it yourself, just shout and I'll hold off touching either branch.

No rush, but flagging a timeline: if there's no response in about two weeks I'll consolidate the parts we still want into a superseding PR with full attribution to you, so the work isn't lost, and close these two out. Let me know either way — happy to help however's easiest.

Thanks for both of these — they're still wanted.

maphew:
Update, and a correction to my comment above: "allow edits from maintainers" was **already enabled** on this PR, so there was never anything you needed to do — sorry for the confusion. I've gone ahead and done the rebase myself.

**What I pushed** (`1a822327e`): merged current `main` into this branch. Four conflicts, all resolved in your PR's favor:

- `cmd/bd/main.go` — since July, main's inline span start gained secret-flag scrubbing of the `bd.args` attribute (`scrubArgsForTelemetry` + `secretFlagTokens`). I kept your `startCommandSpan` seam and threaded the scrub through it: `commandSpanAttrs` now takes the secret-flags map and scrubs `bd.args` itself, so the helper preserves main's redaction behavior while staying a single testable line.
- `docs/reference/observability.md` — git followed the docs-site rename (#4722); kept your `OTEL_*` env-var table, with the legacy `BD_OTEL_*` table remaining in the deprecation section.
- `cmd/bd/storage_chain_test.go` and `internal/telemetry/telemetry.go` — minor import/doc-comment reconciliation.

Validated locally with the repo build tags: build, vet, golangci-lint, and the full `cmd/bd` + `internal/telemetry` test suites all green.

**What happens next — no action needed from you.** The PR is now queued to our automated merge watcher, which merges as soon as CI is green at this head. (That watcher is also why July's "will merge once CI is green" never fired: CI went red on that head — Build Artifacts/Lint fallout — and nothing was watching. This time the machine owns the tail.)

Once this lands I'll do the same fix-merge for #3861 on top, so the pair completes without further work on your end. Thanks again for both PRs.

maphew:
Follow-up push (`356354dd8`): a cross-vendor review pass over the rebased branch surfaced two issues in the original diff, both now fixed:

- **OTLP exporter ignored the exporter selector.** A console-only selection (including the legacy `BD_OTEL_STDOUT=true` translation) still installed an OTLP metric reader whenever a machine-global `OTEL_EXPORTER_OTLP_ENDPOINT` existed — silently exporting bd metrics to an endpoint set up for some other tool, contradicting this PR's own no-silent-redirect guarantee. `OTEL_METRICS_EXPORTER` is now parsed per the SDK spec (comma-separated; unset defaults to `otlp`; `none` disables) and the OTLP reader requires the selection to include `otlp`. The `BD_OTEL_STDOUT=true` + `BD_OTEL_METRICS_URL` combination translates to `console,otlp` so its both-exporters behavior is preserved.
- **Test-suite network stall.** `resetTelemetryState`'s cleanup flushed to the test's unreachable endpoint with an unbounded context, letting OTLP retry for ~30s; the flush is now bounded at 2s.

Unit tests added for the selector parsing and the console-only-ignores-global-endpoint guarantee. Build, lint, and both affected test suites green locally.

PR #3861, opened 2026-05-10
feat(telemetry): stamp bd.prefix on resource and every metric measurement

GraemeF:
## Summary

Stamp `bd.prefix` (the project's issue prefix) on the OTel resource and on
every emitted metric measurement, so dashboards backed by a Prometheus-style
store can split `bd_*` series per beads project.

Resource attributes only land on `target_info` in the OTel→Prometheus
translation and require an instance-label join the SDK doesn't always emit, so
without per-measurement stamping every `bd_*` series collapses onto one label
set per machine and bd invocations from different projects overwrite each
other.

`bd.prefix` is a custom resource attribute, not `service.namespace` — OTel k8s
resource detectors and the OTel Operator both populate `service.namespace` from
the kubernetes namespace, so reusing that key would either be overwritten or
clash with cluster conventions.

Rebased onto current `main` now that both of its dependencies (#3858, #3859)
have landed, and taken out of draft.

## Changes

**Commit 1 — stamp `bd.prefix`**

* `telemetry.Init` takes the issue prefix as a fourth argument, captured via
  `captureBaseAttrs` into a process-wide measurement-attribute slice and
  stamped as `bd.prefix` on the resource when non-empty.
* `cmd/bd/main.go` moves the `initTelemetry` + `startCommandSpan` pair to fire
  after the dolt store opens, so the prefix can be read from the store's config
  table. Commands that exit before opening the store no longer initialise OTel
  — acceptable trade-off because those paths are short-lived setup commands
  with no telemetry need. `bd.actor` is now stamped on the span immediately
  (it resolves before the store opens), rather than via a later
  `SetAttributes`.
* `internal/telemetry/attributes.go` exposes `BaseAttrs()` and
  `WithMergedAttrs(extras...)`. The latter returns a `metric.MeasurementOption`
  combining base attrs with per-call extras; `metric.MeasurementOption` embeds
  `metric.ObserveOption`, so the same helper covers synchronous `Add`/`Record`
  and observable callbacks.
* `docs/reference/observability.md` notes that `bd_prefix` lands on every
  measurement, with a verification snippet.

**Commit 2 — make the guarantee compiler-enforced**

Rebasing showed the weakness in commit 1: `WithMergedAttrs` is opt-in per call
site, and `Add(ctx, 1)` compiles just as happily as
`Add(ctx, 1, WithMergedAttrs())`. In the time this branch sat in draft, twelve
new measurement sites arrived in the dolt store — every one of them unstamped,
and nothing flagged it.

So the instruments now carry the merge instead of the call sites:

* `telemetry.Counter`, `telemetry.Histogram` and `telemetry.Gauge` merge the
  base attributes inside `Add`/`Record`. There is no unstamped emission to
  write, and a raw `metric.WithAttributes` argument no longer compiles.
* The dolt store, `InstrumentedStorage` and the haiku AI client hold these
  types instead of their `metric.*` equivalents. Call sites pass bare
  `attribute.KeyValue`s and read more plainly than before.
* Zero-value instruments no-op rather than panic, so a meter that refuses to
  build an instrument stays silent — telemetry never blocks the user. This also
  retires a hand-written nil guard in the AI client.
* `docs/reference/observability.md` gains an "Adding a metric" section so the
  next instrument gets `bd_prefix` for free.

Observable gauges register their own callbacks and still stamp explicitly with
`o.Observe(v, telemetry.WithMergedAttrs())`.

## Verification

```bash
nix develop --command gofmt -l .
nix develop --command go vet -tags gms_pure_go ./...
nix develop --command go test -tags gms_pure_go ./internal/telemetry/... \
  ./internal/compact/... ./internal/storage/dolt/... ./cmd/bd/...
```

`gofmt` and `go vet` are clean. `internal/telemetry` and `internal/compact`
pass. The failures in `internal/storage/dolt` and `cmd/bd`
(`TestBuildServerDSN_SpecialCharacterPassword`,
`TestWispSeesCustomTypeRegisteredInTransaction`,
`TestServeIssueRolesComeFromBeneathTheHookDecorator`,
`TestMultiIDUpdateAllGoodPathUnchanged`) reproduce identically on unmodified
`main`, as do the `internal/storage/dbproxy/proxy` force-stop failures.

New tests:

* `internal/telemetry/attributes_test.go` — `BaseAttrs`/`WithMergedAttrs`
  contract, defensive copy, recapture-replaces semantics, and Init capturing
  the prefix end-to-end.
* `internal/telemetry/storage_metrics_test.go` — drives `WrapStorage` through a
  real `ManualReader` and asserts `bd.prefix` lands on the emitted
  `bd.storage.operations` counter and `bd.issue.count` gauge, and that no
  `bd.prefix` attribute appears when the prefix is unset.
* `internal/telemetry/instruments_test.go` — `Counter`/`Histogram`/`Gauge`
  stamp base attrs plus extras through a `ManualReader`, and zero-value
  instruments no-op.

To confirm locally without standing up a collector:

```bash
BD_OTEL_ENABLED=true OTEL_METRICS_EXPORTER=console bd list 2>&1 | grep -E 'bd[._]prefix'
```

## Out of scope

* Closing the original bundled PR #3475 — already closed.
* Observable-gauge callbacks keep the explicit `WithMergedAttrs` call; the
  observer comes from the SDK, so wrapping it buys no compiler guarantee.

## Context

Third and last of three split PRs from #3475 (which @coffeegoddd asked be
split):

1. #3858 — Wire `WrapStorage` decorator into bd storage chain (merged)
2. #3859 — Adopt standard OTel SDK env vars with `BD_OTEL_*` back-compat (merged)
3. **this PR** — Stamp `bd.prefix` on resource and every metric


maphew:
@GraemeF heads-up: #3859, which this draft lists as its dependency, merged on 2026-08-03 - so this is unblocked now. Do you want to rebase and mark it ready for review? If you'd rather not pick it back up, happy to offer the rebase-carry with your commits and credit intact - just say the word either way, so it doesn't sit in draft limbo.

GraemeF:
Thanks for the nudge — rebased onto current main and out of draft.

#3858's commit dropped out of the branch on rebase, as expected. Three conflicts needed real resolution against what landed since:

* `telemetry.Init` — kept #3859's `translateLegacyEnv` + `resource.WithFromEnv()` shape and threaded the prefix through it.
* `cmd/bd/main.go` — main has since extracted `initTelemetry`/`startCommandSpan` into `command_telemetry.go`. I kept those helpers (including the arg scrubbing) and moved the call site to after the store opens, which is what this PR needs to read the prefix. `bd.actor` now goes on the span at creation instead of via a later `SetAttributes`.
* `internal/storage/dolt/store.go` — kept main's `classifyManagedRetry` refactor and the indeterminate-commit handling.

Rebasing also exposed a real weakness in the original design, so there's now a second commit. `WithMergedAttrs` was opt-in per call site, and the compiler can't tell `Add(ctx, 1)` from `Add(ctx, 1, WithMergedAttrs())`. While this branch sat in draft, twelve new measurement sites arrived in the dolt store — every one unstamped, and nothing caught it. The instruments now carry the merge themselves (`telemetry.Counter`/`Histogram`/`Gauge`), so there is no unstamped emission left to write. Call sites got simpler as a side effect.

Happy to split that second commit into its own PR if you'd rather review them separately.

GraemeF:
CI note on the first run of this branch, in case a maintainer has more context.

`Test (storage domain + uow)` failed. The other 72 checks passed, including all
20 embedded-Dolt shards and all 15 proxied-Dolt shards.

The failure is not an assertion. About 200 `internal/tracker` engine tests
passed against the shared server on `127.0.0.1:32772`. The server then went away
mid-test:

```
=== RUN   TestEnginePushWithShouldPushCanInspectLabels
[mysql] 2026/08/11 07:11:29 packets.go:58 unexpected EOF
    engine_test.go:1328: CreateIssue(bd-labelpush1) error: dolt add issues after SQL mutation: invalid connection
--- FAIL: TestEnginePushWithShouldPushCanInspectLabels (0.06s)
```

Every later test that needs a store failed the same way, first with
`failed to check if database "tracker_pkg_shared" exists`, then with
`dial tcp 127.0.0.1:32772: connect: connection refused`. The tests in that
package that need no store kept passing.

Job: https://github.com/gastownhall/beads/actions/runs/31467591401/job/93703835808

What I checked:

* The diff touches no tracker code. In `internal/storage/dolt` it changes only
  instrument field types and the shape of the `Add`/`Record` call sites. There
  is no change to control flow, retry behaviour or connection handling.
* `go test ./internal/tracker/` passes on this branch head.
* I cannot rerun the job — `gh run rerun` needs admin rights on this repo. I
  re-signed the head commit instead to trigger a fresh run, so the tree is
  unchanged from `bec62c0`.

I am treating this as a runner-side death of the Dolt server, but I want to
flag that mine was the only failing run in the last 20 on this workflow, so
that is a claim rather than a conclusion. If the same job fails in the same
place on the rerun, the theory is wrong and I will bisect the two commits.

Do you know of anything that makes the shared `tracker_pkg_shared` server exit
part-way through that job?

maphew:
Great to see this rebased and out of draft — and the instrument-carried merge is a real improvement over opt-in `WithMergedAttrs`; twelve silently-unstamped sites is exactly the argument for making the type system hold the invariant. I'd keep the second commit in this PR: it exists *because* the rebase exposed the weakness, and the two read better together than apart. (Contributor opinion — a maintainer may prefer the split.)

On the CI failure, what I can offer as a fellow contributor:

- That job's `internal/tracker` tests run against a **shared Dolt sql-server container** (`internal/testutil/testdoltcommon.go`, `RequireDoltContainer` — the job's "Pull Dolt sql-server image" step feeds it). Your symptom sequence — ~200 tests pass, then `unexpected EOF` mid-test, then `failed to check if database ... exists`, then `connect: connection refused` — is precisely what a container dying mid-run looks like from the client side: first the in-flight connection breaks, then reconnects refuse.
- I don't know of a recurring flake in this specific job, and your "only failing run in the last 20" matches what I've seen shepherding my own PRs through it. Your diff touching only instrument field types and call-site shapes makes a caused-by-this-branch theory hard to construct; a runner-side container death (the usual suspect on 7 GB `ubuntu-latest` runners is memory pressure — the `-race` test binary plus the container share it) fits better.
- The job has no teardown that captures `docker logs` from the container, so when it dies the actual cause (OOM-kill vs. panic) is currently unobservable. If your rerun fails in the same place, that diagnostic gap is worth a tiny PR of its own — a failure-only step dumping `docker ps -a` and the container's logs would turn this from a guess into a reading.

Re-signing the head to trigger a fresh run without changing the tree was the right move given rerun needs admin here.

PR #3498, opened 2026-04-25
Add external server CLI dir override

maphew:
## Summary
- add `BEADS_DOLT_CLI_DIR` for generic external-server CLI fallback paths
- fail fast with clearer errors when CLI fallback is required but no local CLI dir is configured
- avoid running remote CLI commands with an empty guessed path
- document the new environment variable

## Tests
- `go test -tags gms_pure_go ./internal/storage/dolt -run 'TestCLIDir|TestDoltCLIRequiresExplicitDir|TestGitRemoteExternalServerRouting|TestCredentialCLIRoutingE2E'`\n- `go test -tags gms_pure_go ./cmd/bd -run 'TestDoltRemote|TestSetDataDirBlockedInServerMode'`\n\nFull `make test` was attempted but did not complete cleanly due local timeout/disk quota and Go build-cache damage after the quota failure.\n\nCloses bd-fn9

PR #3862, opened 2026-05-10
revert: remove external server CLI dir override

### LINKED ISSUE #3860: 'bd dolt push/pull' BEADS_DOLT_CLI_DIR=.beads/dolt env var requirement?
Hello, Beads dev! Recently (sometime after `bd` 1.0.x) I have started seeing the following error when I try to invoke `bd dolt push/pull`:

```
Error: dolt push requires a local Dolt CLI database directory in external-server mode;
set BEADS_DOLT_CLI_DIR to the local Dolt database path or use a remote type
supported by SQL DOLT_PUSH/DOLT_PULL
```

The error is pedantic because `.beads/dolt/` is the default path to a local dolt DB for all beads installations. This env var cannot be set globally in `~/.bashrc`, `~/.bash_profile`, or with an `alias` because this path is local to every repo that uses beads. Yes, it is possible to setup `direnv` per workspace, but this just puts extra admin burden on beads users for little benefit.

Perhaps this change was added for those who are using beads embedded mode as default (and rely on deprecated JSONL instead of dolt DB as the source of truth for the task graph), but this makes the beads experience worse for those of us using dolt db together with git-backed dolt remotes.

maphew:
## Summary
- revert PR #3498 / commit `0fffa44a2`, which added the `BEADS_DOLT_CLI_DIR` override and the pre-CLI external-server error
- remove the docs, tests, config loading, and CLI guidance for that override
- leave server-mode push/pull behavior to the existing Dolt SQL/remote plumbing instead of requiring a local CLI database directory

Fixes #3860

## Tests
- `go test ./internal/storage/dolt`
- `git diff --check`
- `go test ./cmd/bd` *(blocked locally: missing ICU development header `unicode/uregex.h` from `github.com/dolthub/go-icu-regex/internal/icu`)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,6 @@ or environment variables:
 | `--server-socket` | `BEADS_DOLT_SERVER_SOCKET` | (none; uses TCP) |
 | `--server-user` | `BEADS_DOLT_SERVER_USER` | `root` |
 | | `BEADS_DOLT_PASSWORD` | (none) |
-| | `BEADS_DOLT_CLI_DIR` | local Dolt database path for CLI push/pull |
 
 **Unix domain sockets:** Use `--server-socket` to connect via a Unix socket
 instead of TCP. This avoids port conflicts between concurrent projects and
@@ -144,12 +143,6 @@ is useful in sandboxed environments (e.g., Claude Code) where file-level
 access control is simpler than network allowlists. The Dolt server must be
 started with `dolt sql-server --socket <path>`. Auto-start is not supported
 in socket mode.
-
-When `BEADS_DOLT_SERVER_MODE=1` points at a Dolt server managed outside
-Beads, set `BEADS_DOLT_CLI_DIR` if `bd dolt push` / `bd dolt pull` must use
-the local `dolt` CLI (for example git-protocol remotes or credentials that
-only exist in the current shell). Use the actual Dolt database directory, not
-the server root.
 
 ### Backup & Migration
 
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -868,9 +868,6 @@ func listRemoteSurfaces(ctx context.Context, st storage.RemoteStore, dbPath stri
 		// that flock, so treat the SQL-visible remotes as the shared surface.
 		return sqlRemotes, sqlErr, sqlRemotes, nil
 	}
-	if strings.TrimSpace(dbPath) == "" {
-		return sqlRemotes, sqlErr, nil, fmt.Errorf("local Dolt CLI directory is not configured; set BEADS_DOLT_CLI_DIR to list filesystem remotes")
-	}
 	cliRemotes, cliErr = listDoltCLIRemotes(dbPath)
 	return sqlRemotes, sqlErr, cliRemotes, cliErr
 }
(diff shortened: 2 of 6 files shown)


coffeegoddd (commented):
push and pull should never go through the dolt cli, and only through the server via CALL DOLT_PUSH/PULL

maphew:
Addressed review feedback: the PR no longer adds a config-backed CLI dir. It now routes normal `bd dolt push` / `bd dolt pull` through `CALL DOLT_PUSH` / `CALL DOLT_PULL` and removes the local `dolt push` / `dolt pull` subprocess path for normal sync. `BEADS_DOLT_CLI_DIR` remains documented only for filesystem remote inspection in externally managed local DBs.

coffeegoddd:
@maphew but what is the motivation behind this pr? was push and pull not working? is there a bug? its not clear to me why to make any changes here at all, or what the correct changes should be. my earlier comment true generally, though, that when we have a dolt server, we should use it as the interface for performing remote operations, not the dolt cli.

maphew:
@coffeegoddd this work is response to issue #3860 . I'm happy to step back if this is the wrong analysis and remedy. I jumped on it even though it's about Dolt as it was my commit that created the "must use env var" UX problem.

coffeegoddd:
@maphew ah ok yes i see the issue and yea the analysis is not acccurate. where did this error come from?
```
Error: dolt push requires a local Dolt CLI database directory in external-server mode;
set BEADS_DOLT_CLI_DIR to the local Dolt database path or use a remote type
supported by SQL DOLT_PUSH/DOLT_PULL
```

maphew:
@coffeegoddd I traced the error back to beads. It is not emitted by Dolt itself.

The message was introduced in PR #3498 / commit `0fffa44a2` (`Add external server CLI dir override`) in `internal/storage/dolt/store.go`, inside `requireCLIDir()`:

```go
return "", fmt.Errorf("%s requires a local Dolt CLI database directory in external-server mode; set %s to the local Dolt database path or use a remote type supported by SQL DOLT_PUSH/DOLT_PULL", operation, EnvDoltCLIDir)
```

The path that reaches it on current `main` is roughly:

- `bd dolt push` / `bd dolt pull` calls `DoltStore.Push()` / `Pull()`
- sync routing checks git-protocol remotes, credential routing, cloud-auth routing, and local-CLI remote routing
- if one of those paths decides the local `dolt` CLI is required, `doltCLIPush()` / `doltCLIPull()` calls `requireCLIDir()` before spawning `dolt`
- in generic external-server mode, `CLIDir()` intentionally returns empty unless `BEADS_DOLT_CLI_DIR` is set, because beads cannot safely infer the externally managed server's local database path

So the reported error is beads failing before invoking the CLI, after deciding that this push/pull should use the local CLI path.

My recommendation: keep this PR paused for now rather than merging it as-is. I think the useful direction is to align the routing with your earlier comment: server-mode push/pull should prefer the connected Dolt SQL server via `CALL DOLT_PUSH` / `CALL DOLT_PULL`, and only require `BEADS_DOLT_CLI_DIR` for operations that truly need local filesystem/CLI access. PR #3865 has a possible behavioral intersection because it fixes shared-server mode being applied too late or not at all in some startup paths, but it has no file-level conflict with this PR and does not fully explain the generic external-server `BEADS_DOLT_CLI_DIR` error.

I'll leave #3862 paused pending your further guidance on the intended routing policy.

Signed-off-by: Codex

coffeegoddd:
@maphew from what i can tell that pr 3498 needs to be reverted. Its not clear that it fixed anything, there is no reproduction of a bug, and it introduced at least one new one.

maphew:
Done. I rewrote this PR to follow that direction instead of the earlier SQL-routing patch:

- reset the branch onto current `upstream/main`
- reverted #3498 / `0fffa44a2`
- removed the `BEADS_DOLT_CLI_DIR` config/docs/tests and the external-server preflight error

Local checks:

- `go test ./internal/storage/dolt` passes
- `git diff --check` passes
- `go test ./cmd/bd` is blocked on this machine by missing ICU headers (`unicode/uregex.h`) from `github.com/dolthub/go-icu-regex/internal/icu`

_codex-gpt-5- on behalf of CI Bot_

Issue #3849, opened 2026-05-09
bug: server-mode auto-import has no emptiness guard — re-imports JSONL on every write


rileywhite:
## Bug

In server-mode (`dolt_mode: "server"` in `.beads/metadata.json`), every write command (`bd create`, `bd update`, `bd close`, etc.) unconditionally re-imports the entire `.beads/issues.jsonl` file on each invocation, even when the database is already populated. The import scales linearly with JSONL size, blocking the write until it completes.

## Steps to Reproduce

1. Initialize a bd workspace with `metadata.json: {"dolt_mode": "server", "dolt_database": "<name>"}` and a dolt sql-server running per config.
2. Populate the database with non-trivial data (e.g. `bd import < issues.jsonl`) so that `.beads/issues.jsonl` is non-empty and the dolt server already has issues.
3. Run any single write command: `bd create -t bug "test issue"`
4. Observe `auto-importing N bytes from .beads/issues.jsonl into empty database...` printed to stderr — despite the database not being empty.
5. Subsequent writes repeat the full import.

Confirmed by running `bd -v create --json -t test "diagnostic"` from a clean shell with no other bd processes running.

## Expected Behavior

Write commands complete without re-importing JSONL if the database is already populated.

## Actual Behavior

Every write command:

1. Prints `auto-importing N bytes from .beads/issues.jsonl into empty database...` to stderr.
2. Upserts all N issues from the JSONL into the database (idempotent but expensive).
3. Then executes the intended write.

On a populated database, each write takes as long as a full re-import.

## Root Cause

`maybeAutoImportJSONL` in `cmd/bd/auto_import_upgrade.go` has two code paths that diverge on the emptiness guard:

**Embedded path** (via `jsonlImporter` / `*EmbeddedDoltStore`):
```go
imported, err := importer.ImportJSONLData(ctx, issues, configEntries, "auto-import")
// ImportJSONLData (embeddeddolt/store.go:455-505) checks TotalIssues > 0
// atomically and returns early when non-empty.
```

**Server-mode fallback** (for stores that don't implement `jsonlImporter`):
```go
fmt.Fprintf(os.Stderr, "auto-importing %d bytes from %s into empty database...\n", ...)
result, err := importFromLocalJSONLFull(ctx, s, jsonlPath)  // no count check
```

[...]

PR #3691, opened 2026-05-03
fix(auto-import): restore empty-DB guard regressed by #3630, plus a test repair

scotthamilton77:
## Summary

PR #3630 (\`Fix maybeAutoImportJSONL concurrency race in embedded mode\`, merged 2026-05-01) introduced two related issues. This PR fixes both.

### Issue 1 — Server-mode data clobber regression

The refactor in #3630 moved \`maybeAutoImportJSONL\`'s \`TotalIssues > 0\` emptiness guard *inside* the new embedded-mode \`jsonlImporter\` fast path (correct, atomic, race-safe for that path). But the **top-level** emptiness check that previously guarded the function as a whole was removed, leaving the server-mode (non-\`jsonlImporter\`) fallback path unguarded.

Net effect: every \`bd <command>\` invocation in server mode unconditionally runs \`importFromLocalJSONLFull\` on a non-empty Dolt. That eventually reaches \`InsertIssueIntoTable\` (\`internal/storage/issueops/helpers.go:45-102\`), which uses \`INSERT … ON DUPLICATE KEY UPDATE\` and **overwrites every business field** (\`title\`, \`description\`, \`design\`, \`acceptance_criteria\`, \`notes\`, \`status\`, \`priority\`, \`issue_type\`, \`assignee\`, \`estimated_minutes\`, …) with the JSONL values.

Combined with the auto-export's debounce/commit-watermark gating in \`export_auto.go\` (especially with \`--dolt-auto-commit=off\`), JSONL is frequently stale relative to Dolt. So partial-field \`bd update\` invocations (e.g. \`--append-notes\` alone, \`--priority=N\` alone) silently lose the unspecified fields on the next \`bd\` command — the auto-import re-imposes the stale JSONL row on top of Dolt truth.

#### Reproducer (server mode, default config)

```bash
bd update X --priority=2 --notes=\"X\"      # combined: lands cleanly
bd update X --append-notes \"; appended\"   # single field
bd update X --priority=3                   # single field
bd show X --json | jq '.[0] | {priority, notes}'
```

**Pre-fix:** the third \`bd show\` returns \`priority=3, notes=\"X\"\` — the \`; appended\` from step 2 is silently lost.
**Post-fix:** returns \`priority=3, notes=\"X\\n; appended\"\` — both single-field updates preserve unspecified fields from Dolt.

#### Fix

[...]

PR #3865, opened 2026-05-10
fix: honor shared-server mode before fallback storage

### LINKED ISSUE #3817: cfg == nil fallback in cmd/bd/main.go ignores BEADS_DOLT_SHARED_SERVER, silently creates phantom embedded DBs
When `configfile.Load(beadsDir)` returns `(nil, nil)` because the resolved `beadsDir` has no `metadata.json` or `config.yaml`, `cmd/bd/main.go:946-955` defaults `doltCfg.Database` to `"beads"` but never sets `doltCfg.ServerMode`. The shared-server override at `cmd/bd/main.go:919-927`:

```go
if cfg != nil {
    doltCfg.ServerMode = cfg.IsDoltServerMode()
    if !doltCfg.ServerMode && doltserver.IsSharedServerMode() {
        doltCfg.ServerMode = true
    }
    // ...
}
```

is gated on `cfg != nil`, so even when `BEADS_DOLT_SHARED_SERVER=1` is exported, `ServerMode` stays at its zero value (`false`) and `newDoltStore` calls `embeddeddolt.Open(beadsDir, ...)` instead of `dolt.New(...)`.

## Reproduction (against v1.0.4 / v1.0.2)

```bash
mkdir -p /tmp/x/.beads/shared-server/dolt
cd /
BEADS_DIR=/tmp/x/.beads/shared-server BEADS_DOLT_SHARED_SERVER=1 bd --global list --limit 1
ls -la /tmp/x/.beads
```

**Expected:** bd connects to the shared server's `beads_global` database, or fails with a clear error.

**Actual:** bd silently creates `/tmp/x/.beads/embeddeddolt/` and `/tmp/x/.beads/.local_version`, returns `No issues found.` Subsequent writes go to the phantom embedded DB instead of the shared server. In environments where `~/.beads/` happens to have restrictive perms (e.g. `0750`), the `mkdir` fails loudly with `Error: failed to open database: embeddeddolt: creating data directory: mkdir /Users/<user>/.beads/embeddeddolt: operation not permitted` — that's the lucky case. Where `~/.beads/` is writable, it silently fragments data.

## Suggested fix

Hoist the override out of the `cfg != nil` branch and apply it unconditionally:

```go
doltCfg.ServerMode = cfg != nil && cfg.IsDoltServerMode()
if !doltCfg.ServerMode && doltserver.IsSharedServerMode() {
    doltCfg.ServerMode = true
}
```

The user has already expressed shared-server preference via env var; respect it regardless of whether a project metadata.json was found.

## Related

This composes with #B (separate issue: --global passing through embedded mode) to produce the user-visible failure on \`bd --global\` from a directory with no \`.beads/metadata.json\`.

### LINKED ISSUE #3818: --global passes embedded mode through to embeddeddolt.Open; verifyProjectIdentity lacks --global escape
Two related issues in the `--global` code path.

## B1. `--global` validates IsSharedServerMode but not doltCfg.ServerMode

`cmd/bd/main.go:967-972`:

```go
if globalFlag {
    if !doltserver.IsSharedServerMode() {
        FatalError("--global requires shared-server mode (...)")
    }
    doltCfg.Database = doltserver.GlobalDatabaseName
}
```

Validates the env/yaml flag but does not validate or set `doltCfg.ServerMode`. Combined with #3817 (cfg==nil branch dropping the ServerMode override), this lets `--global` invocations slip through with `ServerMode=false` + `Database=\"beads_global\"`, calling `embeddeddolt.Open(beadsDir, \"beads_global\", ...)`. The global database only exists on the shared server by definition; running it in embedded mode is incoherent.

## B2. `verifyProjectIdentity` has no --global escape

`internal/storage/dolt/store.go:1146` — `verifyProjectIdentity` runs unconditionally on connect. When `bd --global` is invoked from inside a project repo (with the project's `metadata.json` having a real UUID), the connection lands on `beads_global` whose published `project_id` is the well-known sentinel \`\"00000000-0000-0000-0000-000000000000\"\`. The check fails with:

```
Error: failed to open database: PROJECT IDENTITY MISMATCH — refusing to connect

  Local project ID (metadata.json):  <real UUID>
  Database project ID:               00000000-0000-0000-0000-000000000000
```

The outer `validateWorkspaceIdentity` in `cmd/bd/main.go` does have a `--global` escape, but the inner one in the storage layer does not.

## Reproduction

From any project repo with a `.beads/metadata.json`:

```bash
BEADS_DOLT_SHARED_SERVER=1 bd --global list
```

## Suggested fix

- In the `globalFlag` block at `cmd/bd/main.go:967-972`, also set `doltCfg.ServerMode = true`. This independently fixes B1 even without the cfg==nil fix in #3817.
- In `verifyProjectIdentity` (`internal/storage/dolt/store.go:1146`), short-circuit when the database name equals `doltserver.GlobalDatabaseName`, OR treat the well-known `GlobalProjectID` sentinel as a wildcard match.

## Related

#3817 (cfg==nil branch). The two compose to produce the painful experience users hit when first trying to use `--global` from a directory whose discovery walks past their actual project to the `~/.beads/` config root.

### LINKED ISSUE #3835: bd main/1.0.4 regression: bd sql falls into JSONL auto-import and pegs Dolt
## Bug\n\nInstalling `bd` from upstream main (reported as `1.0.4 (b1db6b132)`) caused `bd sql` in the Gas Town workspace to take the JSONL auto-import path, emit auto-import warnings, and peg the shared Dolt server CPU. Rolling back to tagged upstream `v1.0.3 (1b2dd2cb5)` made the same `bd sql` query work normally.\n\n## Expected Behavior\n\n`bd sql --json "select count(*) as count from issues"` should connect to the configured Dolt server database and execute the SQL query without importing JSONL backup files or mutating state.\n\n## Actual Behavior\n\nWith upstream main / `1.0.4`, `bd sql` triggered JSONL auto-import behavior in a server-backed Gas Town workspace. Dolt CPU climbed to very high usage and eventually circuit breakers opened.\n\n## Reproduction Evidence\n\n- `bd` built from upstream main: `bd version 1.0.4 (b1db6b132)`.\n- `bd sql --json "select count(*) as count from issues"` triggered auto-import / Dolt CPU peg.\n- `bd` built from tagged upstream `v1.0.3`: `bd version 1.0.3 (1b2dd2cb5)`.\n- The same SQL query succeeded normally on `v1.0.3`.\n\n## Workaround\n\nPin Gas Town to `github.com/steveyegge/beads/cmd/bd@v1.0.3` until the regression on main is fixed.

maphew:
## Summary

Fixes three P0 v1.0.4 triage issues that share the same root: runtime shared-server intent was not consistently applied before choosing the storage backend.

- honor `BEADS_DOLT_SHARED_SERVER` even when `.beads/metadata.json` is absent or failed to load, preventing fallback into embedded storage (#3817)
- publish the final runtime store mode after the shared-server override so `--global` and identity checks see server mode instead of stale embedded state (#3818)
- classify read-only `bd sql` queries before opening the store, so `SELECT`/`EXPLAIN`/`SHOW`/etc. run through the read-only path and skip JSONL auto-import (#3835)

This keeps proxied-server separate: shared-server env does not override `ProxiedServer`.

## Tests

- `go test -tags gms_pure_go ./cmd/bd -run 'TestLoadServerModeFromBeadsDirSharedServerWithoutMetadata|TestApplyRuntimeSharedServerMode|TestIsReadOnlySQLQuery|TestMaybeAutoImportJSONL_ServerModeFallback'`
- `go test -tags gms_pure_go ./cmd/bd`

## Notes

GH #3849's server-mode auto-import fallback guard is already fixed on current `main` by #3691 (`1cf833734`). This PR handles the remaining `bd sql` read-path trigger from #3835.

Fixes #3817
Fixes #3818
Fixes #3835

### DIFF
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -233,6 +233,13 @@ func loadServerModeFromBeadsDir(beadsDir string) {
 	}
 	cfg, err := configfile.Load(beadsDir)
 	if err != nil || cfg == nil {
+		sm := doltserver.IsSharedServerMode()
+		serverMode = sm
+		proxiedServerMode = false
+		if cmdCtx != nil {
+			cmdCtx.ServerMode = sm
+			cmdCtx.ProxiedServerMode = false
+		}
 		return
 	}
 	repairSharedServerEmbeddedMismatch(beadsDir, cfg)
@@ -256,6 +263,27 @@ func loadServerModeFromBeadsDir(beadsDir string) {
 // skip full DB init but still need to know the mode.
 func loadServerModeFromConfig() {
 	loadServerModeFromBeadsDir(beads.FindBeadsDir())
+}
+
+func applyRuntimeSharedServerMode(doltCfg *dolt.Config) {
+	if doltCfg == nil || doltCfg.ServerMode || doltCfg.ProxiedServer {
+		return
+	}
+	if doltserver.IsSharedServerMode() {
+		doltCfg.ServerMode = true
+	}
+}
+
+func publishRuntimeStoreMode(doltCfg *dolt.Config) {
+	if doltCfg == nil {
+		return
+	}
+	serverMode = doltCfg.ServerMode
+	proxiedServerMode = doltCfg.ProxiedServer
+	if cmdCtx != nil {
+		cmdCtx.ServerMode = doltCfg.ServerMode
+		cmdCtx.ProxiedServerMode = doltCfg.ProxiedServer
+	}
 }
 
 func preserveRedirectSourceDatabase(beadsDir string) {
@@ -897,6 +925,9 @@ var rootCmd = &cobra.Command{
 		// Read-only commands open the store in read-only mode to avoid modifying
 		// the database (which breaks file watchers).
 		useReadOnly := isReadOnlyCommand(cmd.Name())
+		if cmd.Name() == "sql" && len(args) > 0 && isReadOnlySQLQuery(args[0]) {
+			useReadOnly = true
+		}
 
 		// Auto-migrate database on version bump (bd-jgxi).
 		// Runs for ALL commands (including read-only ones) because the migration
@@ -930,17 +961,6 @@ var rootCmd = &cobra.Command{
 			}
 
 			doltCfg.ServerMode = cfg.IsDoltServerMode()
-			// Shared server mode (dolt.shared-server in config.yaml) is a
-			// form of server mode. Override metadata.json if it still says
-			// embedded — handles installs created before GH#2946 fix. Skip
-			// this for proxied-server: it's its own backend, not server.
-			if !doltCfg.ServerMode && !doltCfg.ProxiedServer && doltserver.IsSharedServerMode() {
-				doltCfg.ServerMode = true
-			}
-			serverMode = doltCfg.ServerMode
-			if cmdCtx != nil {
-				cmdCtx.ServerMode = doltCfg.ServerMode
-			}
 
 			// Always set database name (needed for bootstrap to find
 			// prefix-based databases like "beads_hq"; see #1669)
@@ -973,6 +993,12 @@ var rootCmd = &cobra.Command{
 		if doltCfg.Database == "" {
 			doltCfg.Database = configfile.DefaultDoltDatabase
 		}
+		// Shared server mode (dolt.shared-server in config.yaml or
+		// BEADS_DOLT_SHARED_SERVER) is a form of server mode. Apply this after
+		// the cfg nil/default path too, otherwise metadata-free shared-server
+		// workspaces and --global can silently fall back to embedded storage.
+		applyRuntimeSharedServerMode(doltCfg)
+		publishRuntimeStoreMode(doltCfg)
 		doltCfg.SyncRemote = resolveSyncRemote()
 
 		// --global flag: switch to the global shared-server database.
--- a/cmd/bd/main_shared_server_test.go
+++ b/cmd/bd/main_shared_server_test.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/storage/dolt"
+)
+
+func saveStoreModeState(t *testing.T) {
+	t.Helper()
+
+	oldServerMode := serverMode
+	oldProxiedServerMode := proxiedServerMode
+	oldCmdCtx := cmdCtx
+	t.Cleanup(func() {
+		serverMode = oldServerMode
+		proxiedServerMode = oldProxiedServerMode
+		cmdCtx = oldCmdCtx
+	})
+
+	serverMode = false
+	proxiedServerMode = false
+	cmdCtx = nil
+}
+
+func TestLoadServerModeFromBeadsDirSharedServerWithoutMetadata(t *testing.T) {
+	saveStoreModeState(t)
+	t.Setenv("BEADS_DOLT_SHARED_SERVER", "1")
+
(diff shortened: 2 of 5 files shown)


coffeegoddd (commented):
This needs a minimal reproduction that demontrates what the bug is that is being fixed. Once we have that, we can determine the root cause and know what the correct fix is.

maphew:
Reproduction case:

    cd <any-project-with-.beads/metadata.json>
    BEADS_DOLT_SHARED_SERVER=1 bd --global list

Before fix: Fails with:

Error: failed to open database: PROJECT IDENTITY MISMATCH — refusing to connect

  Local project ID (metadata.json):  <real UUID>
  Database project ID:               00000000-0000-0000-0000-000000000000

This happens because verifyProjectIdentity unconditionally compares the local project's UUID against the global database's sentinel UUID. The function had no awareness of --global or the beads_global database.

After fix: verifyProjectIdentity now short-circuits with two escape routes:

    If s.database == "beads_global" — skip verification immediately
    If the database's _project_id == "00000000-0000-0000-0000-000000000000" — skip the mismatch check

This allows bd --global to work from any project directory without the identity check falsely flagging the expected sentinel UUID as a "mismatch."

mieubrisse:
@coffeegoddd any chance we can get this merged? It's a real friction point as my agents try to file global beads

coffeegoddd (commented):
This PR cant be merged as is... it claims to fix multiple bugs, but it only provides a quasi-reproduction for one: the --global flag error. The proposed fix is also nonsensical, it updates a verification method to not verify something is one particular case, which is not what verification should be doing. Either we care enough to verify the thing, in which case we should, or we dont, and we should delete the verification step.

I can reproduce the error myself:
```bash
➜  beads git:(main) bd version
bd version 1.0.4 (dev: 0c47b3c5ac77)
➜  beads_3865 git:(main) ✗ bd init --shared-server --global
  ✓ Shared Dolt server started
  ✓ Global database beads_global available
  ✓ Global database schema initialized
  ✓ Configured Dolt remote: origin → git+https://github.com/coffeegoddd/dotfiles.git
✓ bd initialized successfully!

  Backend: dolt
  Mode: server
  Server: root@127.0.0.1:3308

  ⚠ Server host defaulted to 127.0.0.1.
    If your Dolt server is remote, set BEADS_DOLT_SERVER_HOST or pass --server-host.
  Database: beads_3865
  Issue prefix: beads_3865
  Issues will be named: beads_3865-<hash> (e.g., beads_3865-a3f2dd)

➜  beads_3865 git:(main) ✗ bd dolt status
Dolt server: running
  PID:  1866778
  Port: 3308
  Data: /home/dustin/.beads/shared-server/dolt
  Logs: /home/dustin/.beads/shared-server/dolt-server.log
  Mode: shared server
➜  beads_3865 git:(main) ✗ bd list --global
Error: failed to open database: PROJECT IDENTITY MISMATCH — refusing to connect

  Local project ID (metadata.json):  803f4eef-d18d-4198-808b-b86b664f9ff9
  Database project ID:               00000000-0000-0000-0000-000000000000

This means the Dolt server is serving a DIFFERENT project's database.
This can happen when:
  - Another project's server is running on the same port
  - The server restarted with a different data directory

To diagnose: bd dolt status
Do NOT run 'bd init' — your data likely exists, just on a different server.
```

This PR has the correct fix for the --global error and a regression test: https://github.com/gastownhall/beads/pull/3932

maphew:
Thanks! I appreciated the deeper scrutiny and redirection.

maphew:
Closing, dead and/or stale in it's current form. Please re-raise if the problems still reproduce in v1.0.5+.

PR #3493, opened 2026-04-25
fix(storage/dolt): DOLT_COMMIT after label/comment writes

GGPrompts:
## Summary

`DoltStore.{AddLabel,RemoveLabel,AddComment,AddIssueComment,ImportIssueComment}` only committed the SQL transaction; they did not fire `CALL DOLT_COMMIT`. The other issue mutations (`CreateIssue`/`UpdateIssue`/`Claim`/`Close`/`Delete` and `AddDependency`/`RemoveDependency`) already commit per-write via `doltAddAndCommit`; the label/comment paths were the gap.

The maintainers already know about this gap: `cmd/bd/create_form.go:261-271` carries an explicit post-create commit workaround for the create flow (referencing GH#2009), but standalone `bd label add`, `bd label rm`, `bd comment add`, and `bd compact`'s `AddComment` had no equivalent.

## Symptoms

**In server mode** (bd connecting to an external `dolt-sql-server` — e.g. Termux clients reaching a shared Dolt instance), label/comment writes accumulate as uncommitted rows in the server's working set. Subsequent remotesapi pushes from other clients (e.g. a desktop in embedded mode) fail with `target has uncommitted changes — --force required to overwrite` because the remote `main` has dirty rows the server-mode writers never committed.

**In embedded mode** it's a latent correctness issue — label and comment changes silently don't make it into pushes until the user manually `bd dolt commit`s. The current default of `dolt.auto-commit=off` makes this trivial to hit.

## Changes

- `internal/storage/dolt/labels.go` — `AddLabel`, `RemoveLabel` call `s.doltAddAndCommit(...)` after the SQL tx commits.
- `internal/storage/dolt/events.go` — `AddComment`, `ImportIssueComment` (which `AddIssueComment` delegates to) do the same.
- All four are gated by `s.isActiveWisp(ctx, issueID)` so wisp tables (which live in `dolt_ignore`'d tables and intentionally skip versioning) keep their existing behavior — matching the pattern in `CreateIssue`/`UpdateIssue`.
- Transaction-scoped writes via `RunInTransaction` are unaffected — `runDoltTransaction` already commits at the end with `versioncontrolops.StageAndCommit`, so these per-method commits don't fire when the call originates from inside a transaction.

## Test plan

- [x] New test `TestLabelAndCommentWritesLeaveCleanWorkingSet` in `internal/storage/dolt/server_mode_flush_test.go` asserts `dolt_status` is empty (excluding `dolt_ignore`'d tables) after each of the patched mutations.
[...]

PR #3868, opened 2026-05-10
fix(bd): write labels in CreateIssue tx instead of N AddLabel calls

eric-jones:
## What

`bd create --labels A,B,C` (CLI and interactive-form paths) now emits a
single `EventCreated` row with all labels persisted in the create
transaction, instead of `EventCreated` (no labels) followed by N
`EventLabelAdded` events.

## Why

`buildCreateIssue` (`cmd/bd/create.go:746`) constructs `*types.Issue`
without a `Labels` field, so `PersistLabels` (called inside
`CreateIssueInTx`, `internal/storage/issueops/create.go:98`) is a no-op
on an empty slice — labels weren't landing in the create transaction.
The post-create code at `cmd/bd/create.go:568-575` then looped
`store.AddLabel` per label, emitting an `EventLabelAdded` row each time
and writing a follow-up Dolt commit per #2055. The interactive form
path (`cmd/bd/create_form.go`, used by `CreateIssueFromFormValues`) had
the same shape.

Per `bd create --labels=...` invocation:
- N `EventLabelAdded` events suppressed.
- One Dolt commit instead of one + N working-set writes.
- The create event correctly carries the initial label set instead of
  showing `labels=null`.

`PersistLabels` already iterates `issue.Labels` and writes each label
inside the same SQL transaction (`internal/storage/issueops/create.go:255-271`)
— the storage layer was already prepared for this; the CLI/form just
weren't using the field.

## Surface check

`cmd/bd/create.go` and `cmd/bd/create_form.go` are the canonical
surfaces for the CLI's and the form's labeled-create flows; the
underlying tx-level helper (`PersistLabels`) already supports atomic
writes. No new API; the fix is to feed labels into the struct that's
already being persisted.

## Behavior change

Per-label INSERT failures previously emitted `WarnError` and continued
(partial-success); they now roll back the whole `CreateIssue`.

This is intentional and arguably more correct. `PersistLabels` uses
`INSERT INTO labels ... ON DUPLICATE KEY UPDATE label = label`, so the
common "label already exists" case is a no-op, not an error. The only
INSERT failures left are infrastructure errors (connection lost, schema
violation, label string exceeding the column length) — exactly the
class of error that should roll back a create rather than silently
warn-and-continue with a half-formed issue. The prior loose behavior
was masking write errors that should surface.

## Verification

`make test` runs the regression tests:

```bash
cd <beads>
make test
# CLI-level (subprocess; requires BEADS_TEST_EMBEDDED_DOLT=1):
BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go \
    -run 'TestEmbeddedCreate/labels_atomic' ./cmd/bd/...
# Storage and form-path tests:
go test -tags gms_pure_go \
    -run 'TestCreateSuite/CreateWithLabelsAtomic|TestCreateIssueFromFormValues/WithLabelsAtomic' \
    ./cmd/bd/...
```

Three tests cover the change:

- **`TestEmbeddedCreate/labels_atomic`** (CLI regression). Runs
  `bd create --labels alpha,beta,gamma` as a subprocess and queries
  the events table directly. Asserts exactly 1 `created` event and
  0 `label_added` events. Under the previous shape would see 3
  `label_added` rows. This is the layer-matched regression test for
  the actual bug.
- **`TestCreateIssueFromFormValues/WithLabelsAtomic`** (form-path
  regression). Calls `CreateIssueFromFormValues` with pre-populated
  `fv.Labels` and asserts the same event-count invariant.
- **`TestCreateSuite/CreateWithLabelsAtomic`** (storage contract).
  Calls `s.CreateIssue` with `Labels` pre-populated. Documents the
  storage-layer invariant the CLI/form fixes rely on; doesn't
  regression-cover the CLI plumbing.

Existing tests (`TestCreateSuite/WithLabels`, `TestEmbeddedCreate/labels`,
`TestCreateIssueFromFormValues/WithLabels`) still pass — `AddLabel`
itself is unchanged.

## Out of scope

`make test-integration` per the repo's PR template — current local
policy is to defer (past attempts hit the 30-minute timeout with no
saved partial output). Reviewers can request a run if they want one.

## Discovery sweep

- **#2055** (merged) — established the post-create commit pattern this
  PR optimizes.
- **#3493** (open) — `DOLT_COMMIT after label/comment writes` in server
  mode. Complementary; this PR reduces the surface #3493 patches but
  doesn't conflict.
- **#3803** (open) — `doltTransaction CloseIssue/AddLabel/RemoveLabel
  now emit events`. Opposite problem (missing events in a different
  code path); unrelated.

No overlapping in-flight PR for the atomic-labels-at-create
optimization.

### DIFF
--- a/cmd/bd/create.go
+++ b/cmd/bd/create.go
@@ -528,31 +528,10 @@ var createCmd = &cobra.Command{
 			// If error getting parent or parent has no source_repo, continue with default
 		}
 
-		if err := store.CreateIssue(ctx, issue, actor); err != nil {
-			FatalError("%v", err)
-		}
-
-		// Track whether any post-create writes occurred. CreateIssue commits
-		// the issue to Dolt internally, but subsequent AddDependency/AddLabel
-		// calls only write to the working set. A follow-up Dolt commit is
-		// needed to persist them (GH#2009).
-		postCreateWrites := false
-
-		// If parent was specified, add parent-child dependency
-		if parentID != "" {
-			dep := &types.Dependency{
-				IssueID:     issue.ID,
-				DependsOnID: parentID,
-				Type:        types.DepParentChild,
-			}
-			if err := store.AddDependency(ctx, dep, actor); err != nil {
-				WarnError("failed to add parent-child dependency %s -> %s: %v", issue.ID, parentID, err)
-			} else {
-				postCreateWrites = true
-			}
-		}
-
 		// Merge inherited parent labels with user-specified labels (GH#2100)
+		// and attach them to the issue so they're persisted in the same
+		// transaction as the insert. Previously a post-create AddLabel loop
+		// emitted N extra update events per create (stg-u81).
 		if len(inheritedLabels) > 0 {
 			seen := make(map[string]bool)
 			for _, l := range labels {
@@ -564,11 +543,27 @@ var createCmd = &cobra.Command{
 				}
 			}
 		}
+		issue.Labels = labels
 
-		// Add labels if specified
-		for _, label := range labels {
-			if err := store.AddLabel(ctx, issue.ID, label, actor); err != nil {
-				WarnError("failed to add label %s: %v", label, err)
+		if err := store.CreateIssue(ctx, issue, actor); err != nil {
+			FatalError("%v", err)
+		}
+
+		// Track whether any post-create writes occurred. CreateIssue commits
+		// the issue (and labels) to Dolt internally, but subsequent
+		// AddDependency calls only write to the working set. A follow-up
+		// Dolt commit is needed to persist them (GH#2009).
+		postCreateWrites := false
+
+		// If parent was specified, add parent-child dependency
+		if parentID != "" {
+			dep := &types.Dependency{
+				IssueID:     issue.ID,
+				DependsOnID: parentID,
+				Type:        types.DepParentChild,
+			}
+			if err := store.AddDependency(ctx, dep, actor); err != nil {
+				WarnError("failed to add parent-child dependency %s -> %s: %v", issue.ID, parentID, err)
 			} else {
 				postCreateWrites = true
 			}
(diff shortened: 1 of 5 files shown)


coffeegoddd (commented):
This is an optimization to reduce the number of commits in the commit log, not really a bug fix per se, is that true?

Can you file an issue for this if you have not done so yet, and also come up with a minimal repro that shows the behavior before your change and the behavior after?

eric-jones:
Filed https://github.com/gastownhall/beads/issues/3898 with the minimal repro you asked for (~30s, two builds at PR base vs HEAD).

On the framing — I think this is multi-faceted, but the bug-fix piece is load-bearing:

- The `created` event lies about the issue's birth state. With `bd create --labels A,B,C`, the audit trail today reads "created (no labels) → label_added A → label_added B → label_added C" with three distinct timestamps, implying the labels were added *after* creation. They weren't. Anything reading the events table for retrospective "when did this label first appear" sees the wrong story.
- The pre-fix path silently masks `INSERT INTO wisp_labels` errors. The `for _, label := range labels { AddLabel(...) }` loop catches errors with `WarnError` and continues, so an INSERT failure on label 2/3 leaves a half-formed issue with no failure surfaced to the caller. The atomic-create path uses `PersistLabels`, whose `INSERT … ON DUPLICATE KEY UPDATE label = label` makes the common "duplicate label" case a no-op — the only failures that now roll back are infrastructure errors (connection lost, column-length violation), which should roll back.
- The event/commit reduction is real but is a side effect of the fix, not its driver.

Happy to extend the issue or PR description if any of that framing reads differently from your end.

eric-jones:
Withdrawn — #4149 just merged and covers the same surface (atomic label writes in the initial CreateIssue transaction, both at the issueops layer and via the CLI create path). Closing in favor of that.

Thanks @julianknutsen and reviewers for the work on #4149. Happy to redraft if any specific case here isn't covered.

Issue #3522, opened 2026-04-26
auto-backup retries 'add backup_export' on every bd command instead of backing off after failure


shaunc:
## Summary

When the auto-backup `add backup backup_export <url>` step fails, bd does not record the failure and back off — it retries the same registration on every subsequent bd command, generating identical WARNING output for the rest of the operator session.

## Reproducer

Any environment where the auto-backup register fails (e.g., the host-path-vs-container-path mismatch from #3501 / sibling issue):

```
$ bd ready
Warning: auto-backup failed: register backup remote: add backup backup_export: Error 1105 (HY000): ...

$ bd show some-id
Warning: auto-backup failed: register backup remote: add backup backup_export: Error 1105 (HY000): ...

$ bd config show
Warning: auto-backup failed: register backup remote: add backup backup_export: Error 1105 (HY000): ...
```

Same warning, every command, full session. Tooling that wraps bd (CI, hooks, scripts) sees the warning flood logs.

## Expected

Either:

1. **Backoff with persistence**: record the last failed-register attempt + error in the workspace state (e.g., `.beads/auto-backup-state.json` with `{last_attempt, last_error, retry_after}`). Skip retry until `retry_after` elapses (e.g., 1h). Print the warning at most once per backoff window.
2. **In-process memoization at minimum**: if bd is invoked via `--watch`/long-running flow, don't retry within the same process. Doesn't help short-lived invocations but mitigates the worst case.

(1) is more ambitious but the right shape for a tool that's invoked dozens of times per dev session.

## Environment

- bd version 1.0.2 (a3f834b3)
- Any setup where register fails persistently.

## Related

- #3501 — `auto-register backup_export` ignores `backup.enabled=false` (parent).
- (sibling) — symlink-naive URL.
- (sibling) — containerized-server detection.

shaunc:
Still biting us — bd v1.0.2 (a3f834b3), today 2026-05-10.

**Per-command latency observed right now** on a healthy local-network shared dolt server:

- `bd show <id>` — **10.1s**
- `bd ready` — **15.4s**
- `bd list` — **14.9s**

Without the register-retry these would be sub-second. The retry happens on **every single bd invocation**.

The warning:

```
Warning: auto-backup failed: register backup remote: add backup backup_export: Error 1105 (HY000): failed to create directory '/home/shauncutts/src/factfiber.ai/infra/factfiber-meta_c/.beads/backup': mkdir /home/shauncutts: permission denied
```

[...]

Issue #3501, opened 2026-04-26
Auto-register `backup_export` Dolt remote ignores `backup.enabled=false`; warns on every command


shaunc:
## Summary

`bd` runs an unconditional Dolt-server-side remote registration each command (something like `add backup backup_export <local-path>`). When the Dolt server is in a docker container that doesn't share the operator's host filesystem, the mkdir fails inside the container and bd prints a noisy warning on **every** command — including read-only commands like `bd ready`, `bd config show`.

`backup.enabled = false` and `backup.git-push = false` (both default) do not gate this behavior.

## Reproducer

Operator runs Dolt server in a docker container with port-forwarded MySQL wire (e.g. `dolt-sql-server` on `0.0.0.0:3306` published to host `127.0.0.1:3308`), with `BEADS_DOLT_SERVER_HOST=127.0.0.1`. The container's filesystem doesn't include the operator's home dir.

```
$ bd config show 2>&1 | grep backup
  backup.enabled                                            = false                                                         (default)
  backup.git-push                                           = false                                                         (default)
  backup.interval                                           = 15m                                                           (default)

$ bd ready
[normal output]
Warning: auto-backup failed: register backup remote: add backup backup_export: Error 1105 (HY000): failed to create directory '/home/shauncutts/src/factfiber.ai/learn/xa-types/.beads/backup': mkdir /home/shauncutts: permission denied
```

The path bd asks Dolt to create is the operator's host filesystem path. Dolt (running inside the container) doesn't have that path and can't create it. The warning fires on every command — including `bd ready`, `bd close`, `bd config show`, etc.

## Why this matters

1. **Noise**: every bd invocation prints a stderr warning. In tooling that wraps bd (CI, hooks, scripts), the warning floods logs.
2. **Misleading**: `backup.enabled=false` suggests bd should not be touching backup-related state, but the auto-register fires regardless.
3. **No escape hatch**: there's no documented config to disable the auto-register.

## Expected

One of:
- `backup.enabled=false` gates the auto-register-backup-remote step (most natural).
- A new key like `backup.auto-register = false` (default true for back-compat) lets operators opt out.
- bd silently no-ops when the Dolt server is reachable but can't see the filesystem path (best UX).

## Environment

- bd version 1.0.2 (a3f834b3)
[...]

PR #3869, opened 2026-05-10
fix(auto-backup): serialize concurrent forks via flock

eric-jones:
## What

Acquire a non-blocking exclusive flock at `<backup-dir>/.backup.lock`
in `maybeAutoBackup` between the throttle check and change detection.
Only one fork per throttle window does the registration + sync;
contended forks see `lockfile.IsLocked` and skip silently.

## Why

In server mode (multiple bd CLI processes against one Dolt sql-server)
`PersistentPostRun` fires per fork. When the throttle window opens, all
forks read the same stale `backup_state.json`, pass the cheap throttle
check at roughly the same instant, and race into `BackupDatabase`'s
unconditional `BackupRemove` followed by `BackupAdd` for `backup_export`.

Dolt logs each duplicate `CALL DOLT_BACKUP('rm', ...)` and
`CALL DOLT_BACKUP('add', ...)` as a warning (`backup_export not found`,
`backup_export already exists`) even though the rm error is absorbed at
the call site (`internal/storage/{embeddeddolt,dolt}/...:BackupDatabase`,
the `_ = versioncontrolops.BackupRemove(...)` line) and the add
eventually succeeds for one fork. The result is noisy `dolt.log`
output (5+ duplicate warnings within a single ms during a 5-fork
race observed locally) that downstream tooling flags as non-benign,
requiring manual inspection each run to confirm it was absorbed.

Functional impact pre-fix: none — the registration always succeeds for
one fork. Operational impact: log noise + cognitive overhead.

## Verification

- `go test -tags=gms_pure_go -count=1 -run '^TestAutoBackupLockSerializesForks$|^TestIsBackupAutoEnabled$' ./cmd/bd/` — pass.
- New test `TestAutoBackupLockSerializesForks` pins the contract that
  `util.TryLock` returns a `lockfile.IsLocked`-recognized error when a
  second TryLock targets the same path. The skip-on-contention branch
  in `maybeAutoBackup` depends on that recognition.
- `make test`: pre-existing failures unrelated to this change reproduce
  identically on `origin/main` (cb9d7aad5):
  - `cmd/bd`: `TestContextUsesExplicitDBFlagForNoDBCommand` deterministically
    hangs at the `os/exec` pipe-wait stack (Go #23019 shape) — same
    121s timeout, same goroutine stack on the unmodified worktree.
    Filed as EXPECT-FAIL.
  - `internal/beads`: `TestRole_NoConfig` / `TestRequireRole_NotConfigured`
    fail when the test environment has `git config --global beads.role`
    set (the tests `git init` a temp repo but don't isolate from
    inherited global config). Environmental, not introduced here.
  - `internal/storage/dolt`: `TestCloudAuthCLIRouting/azure_env_+_s3://_remote`
    is the documented EXPECT-FAIL for that area.

`make test-integration` not run (per project policy: known to exceed
the 30-min tool timeout with no saved partial output).

## Design notes

**Lock scope: single fork per throttle window.** Under contention the
lock-loser returns silently; it doesn't queue, doesn't retry, doesn't
re-check the throttle. This is sufficient because the lock-holder's run
already covers the loser's changes (auto-backup is rolling, not
per-commit). After the holder updates `backup_state.json`, the next
incoming fork sees a fresh timestamp and the cheap throttle check fires
before the lock is touched.

**Throttle before lock.** Most commands hit the throttle skip and never
need the flock — the lock acquisition stays out of the steady-state
PersistentPostRun fast path.

**Lock placement: `<backup-dir>/.backup.lock`.** The backup-dir is
already used to coordinate the same set of processes (`backup_state.json`
lives next to it) and `backupDir()` creates it `0700`. `util.TryLock`
also handles `MkdirAll` defensively. This matches the existing
embeddeddolt pattern (`acquireEmbeddedLock`) which holds an exclusive
flock on the data-dir's `.lock` file.

**Out of scope.** The unconditional `_ = bs.BackupRemove(ctx,
"backup_export")` in `cmd/bd/backup_dolt.go:411` (the explicit
`bd backup remove` command path) is left untouched: that path is
user-explicit and not the high-fork-rate caller observed in practice.

### DIFF
--- a/cmd/bd/backup_auto.go
+++ b/cmd/bd/backup_auto.go
@@ -4,11 +4,14 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/debug"
+	"github.com/steveyegge/beads/internal/lockfile"
 	"github.com/steveyegge/beads/internal/storage"
+	"github.com/steveyegge/beads/internal/storage/dbproxy/util"
 )
 
 // isBackupAutoEnabled returns whether backup should run.
@@ -64,6 +67,23 @@ func maybeAutoBackup(ctx context.Context) {
 			time.Since(state.Timestamp).Round(time.Second), interval)
 		return
 	}
+
+	// Serialize concurrent auto-backups across bd CLI forks. In server mode
+	// multiple bd processes share one Dolt sql-server; without this lock,
+	// PersistentPostRun fires per fork and all forks race past the throttle
+	// window into BackupDatabase's rm-then-add, producing duplicate
+	// CALL DOLT_BACKUP traffic. Non-blocking: if another fork holds the lock,
+	// this one skips — the holder's run already covers our changes.
+	lock, err := util.TryLock(filepath.Join(dir, ".backup.lock"))
+	if err != nil {
+		if lockfile.IsLocked(err) {
+			debug.Logf("backup: another process is backing up, skipping\n")
+			return
+		}
+		fmt.Fprintf(os.Stderr, "Warning: auto-backup skipped: %v\n", err)
+		return
+	}
+	defer lock.Unlock()
 
 	// Change detection: skip if nothing changed
 	currentCommit, err := store.GetCurrentCommit(ctx)
--- a/cmd/bd/backup_auto_test.go
+++ b/cmd/bd/backup_auto_test.go
@@ -4,9 +4,12 @@ package main
 
 import (
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/steveyegge/beads/internal/config"
+	"github.com/steveyegge/beads/internal/lockfile"
+	"github.com/steveyegge/beads/internal/storage/dbproxy/util"
 )
 
 func TestIsBackupAutoEnabled(t *testing.T) {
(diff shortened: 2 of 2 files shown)


coffeegoddd (commented):
Adding a flock is not the correct fix, auto-backup is likely designed poorly. Please file an issue, this needs further investigation.

eric-jones:
Filed https://github.com/gastownhall/beads/issues/3878 with a design analysis per your request — covers what auto-backup does on every fork today, why we read flock as a workaround rather than a fix, the relation to existing #3501 and #3522 (which describe the same per-fork-ownership weakness from other angles), and a few non-prescriptive design directions. Honest about what we don't know (e.g. `default` vs `backup_export` intent, whether `CALL DOLT_BACKUP` could grow an upsert form).

Happy to close this PR if you'd prefer the conversation move entirely to the issue, or leave it open while the design discussion settles — your call.

coffeegoddd:
@eric-jones yes i think discussing in the issue and closing this pr is the best way forward, thanks so much!

eric-jones:
Closing per maintainer request — discussion moved to #3878 for the underlying design analysis. Thanks for the steer; appreciate the time on the review.

eric-jones:
Closing per maintainer request — discussion moved to #3878 for the underlying design analysis. Thanks for the steer; appreciate the time on the review.

PR #3874, opened 2026-05-11
fix(mol bond): resolve plain formula names like every other subcommand (GH#3873)

### LINKED ISSUE #3873: bd mol bond rejects plain formula names that every other subcommand resolves
## Problem

`bd mol bond <formula-name> <mol-id>` rejects plain formula names with `Error: '<formula-name>' not found (not an issue ID or formula name)`, even when the same name resolves fine through every other subcommand (`bd formula show`, `bd mol seed`, `bd mol pour`, `bd cook`).

This breaks the documented declarative-fanout pattern: a parent formula bonding a child formula by name. Any formula that recursively bonds itself or a sibling is silently broken.

Reproduced on v1.0.4 (cb9d7aad5).

## Location

`cmd/bd/mol_bond.go:639-654`

```go
func looksLikeFormulaName(operand string) bool {
    if strings.HasPrefix(operand, "mol-") {
        return true
    }
    if strings.Contains(operand, ".formula") {
        return true
    }
    if strings.Contains(operand, "/") || strings.Contains(operand, "\\") {
        return true
    }
    return false
}
```

This gatekeeper is called from `resolveOrDescribe` (`mol_bond.go:568`) and `resolveOrCookToSubgraph` (`mol_bond.go:599`). When the operand misses the local store, the gatekeeper decides whether to even *try* `parser.LoadByName`. Plain formula names match none of the three patterns and are short-circuited to "not found."

Every other subcommand skips this gatekeeper and calls the registry resolver directly, which is why they all succeed and only bond fails.

## Current Behavior

See the MWE regression test at `cmd/bd/mol_bond_gwn_test.go` (`TestBdMolBond_ResolvesPlainFormulaName`). It sets up a workspace with a formula named `plain-name-formula` and an issue, then runs:

```
$ bd formula show plain-name-formula
Formula: plain-name-formula
...                                           # works

$ bd mol bond plain-name-formula <issue-id> --dry-run
Error: 'plain-name-formula' not found (not an issue ID or formula name)
```

The test fails on current main with that exact error and passes after the fix below.

The known workaround `bd cook --persist <formula-name>` creates a permanent proto bead whose ID equals the formula name; subsequent bonds then succeed *as issue IDs*, not formula names. The formula-name branch is the broken one.

## Expected Behavior

`bd mol bond <formula-name> <mol-id>` resolves any name that `bd formula show` resolves. No persisted proto required.

## Recommended Fix

Remove the gatekeeper and let `parser.LoadByName` decide. It already returns `formula %q not found in search paths` for genuinely unknown names, which is the correct error.

In `cmd/bd/mol_bond.go`, change `resolveOrDescribe` and `resolveOrCookToSubgraph` to skip the `looksLikeFormulaName` short-circuit:

```go
// Before
if !looksLikeFormulaName(operand) {
    return nil, "", fmt.Errorf("'%s' not found (not an issue ID or formula name)", operand)
}
parser := formula.NewParser()
f, err := parser.LoadByName(operand)
if err != nil {
    return nil, "", fmt.Errorf("'%s' not found as issue or formula: %w", operand, err)
}

// After
parser := formula.NewParser()
f, err := parser.LoadByName(operand)
if err != nil {
    return nil, "", fmt.Errorf("'%s' not found as issue or formula: %w", operand, err)
}
```

Mirrors what `bd mol seed`, `bd mol pour`, `bd formula show`, and `bd cook` already do. `looksLikeFormulaName` itself can be removed (no other callers).

## Priority

**MUST-DO** — Blocks an entire documented composition pattern. P1.

jjgarzella:
## What

`bd mol bond <formula-name> <mol-id>` couldn't resolve plain formula names like `download-arxiv` or `comprehensive-literature-search` — names that `bd formula show`, `bd mol seed`, `bd mol pour`, and `bd cook` all resolve without trouble. Drop the gatekeeper that rejected them so bond uses the same resolution path as every other subcommand.

## Why

Fixes #3873.

`resolveOrDescribe` and `resolveOrCookToSubgraph` in `cmd/bd/mol_bond.go` gated the formula-registry fallback on `looksLikeFormulaName`, which only accepted names starting with `mol-`, containing `.formula`, or containing a path separator. Anything else — including normal formula filenames — short-circuited to `Error: '<name>' not found (not an issue ID or formula name)` before `parser.LoadByName` was ever tried. Every other subcommand calls the parser directly, which is why only bond was broken. This silently breaks the documented declarative-fanout pattern (a formula bonding a child formula by name).

## Verification

```
go test -tags 'gms_pure_go cgo' -run TestBdMolBond_ResolvesPlainFormulaName -v ./cmd/bd/
```

New e2e regression test at `cmd/bd/mol_bond_gwn_test.go` writes a plain-named formula, confirms `bd formula show` resolves it (rules out test-setup issues), and asserts `bd mol bond <name> <id> --dry-run` succeeds. Fails on `main` with the verbatim user error, passes on this branch.

Sanity checks: `go vet -tags 'gms_pure_go cgo' ./cmd/bd/` clean, `go build -tags gms_pure_go ./cmd/bd/` clean.

### DIFF
--- a/cmd/bd/mol_bond.go
+++ b/cmd/bd/mol_bond.go
@@ -575,12 +575,10 @@ func resolveOrDescribe(ctx context.Context, s storage.DoltStorage, operand strin
 		}
 	}
 
-	// Not found as issue - check if it looks like a formula name
-	if !looksLikeFormulaName(operand) {
-		return nil, "", fmt.Errorf("'%s' not found (not an issue ID or formula name)", operand)
-	}
-
-	// Try to load the formula (but don't cook it)
+	// Not found as issue — fall through to the formula registry. parser.LoadByName
+	// returns "formula %q not found in search paths" for genuinely unknown names,
+	// which is the right error. This matches the resolution behavior of
+	// bd formula show / bd mol seed / bd mol pour / bd cook.
 	parser := formula.NewParser()
 	f, err := parser.LoadByName(operand)
 	if err != nil {
@@ -619,37 +617,15 @@ func resolveOrCookToSubgraph(ctx context.Context, s storage.DoltStorage, operand
 		}
 	}
 
-	// Not found as issue - check if it looks like a formula name
-	if !looksLikeFormulaName(operand) {
-		return nil, false, fmt.Errorf("'%s' not found (not an issue ID or formula name)", operand)
-	}
-
-	// Try to cook formula inline to in-memory subgraph
-	// Pass vars for step condition filtering (bd-7zka.1)
+	// Not found as issue — fall through to the formula registry. Same rationale
+	// as resolveOrDescribe above: let the parser decide. Pass vars for step
+	// condition filtering (bd-7zka.1).
 	subgraph, err := resolveAndCookFormulaWithVars(operand, nil, vars)
 	if err != nil {
 		return nil, false, fmt.Errorf("'%s' not found as issue or formula: %w", operand, err)
 	}
 
 	return subgraph, true, nil
-}
-
-// looksLikeFormulaName checks if an operand looks like a formula name.
-// Formula names typically start with "mol-" or contain ".formula" patterns.
-func looksLikeFormulaName(operand string) bool {
-	// Common formula prefixes
-	if strings.HasPrefix(operand, "mol-") {
-		return true
-	}
-	// Formula file references
-	if strings.Contains(operand, ".formula") {
-		return true
-	}
-	// If it contains a path separator, might be a formula path
-	if strings.Contains(operand, "/") || strings.Contains(operand, "\\") {
-		return true
-	}
-	return false
 }
 
 func init() {
--- a/cmd/bd/mol_bond_gwn_test.go
+++ b/cmd/bd/mol_bond_gwn_test.go
@@ -1,0 +1,107 @@
+//go:build cgo
+
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// TestBdMolBond_ResolvesPlainFormulaName is a regression test for GH#3873:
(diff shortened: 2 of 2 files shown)


maphew:

harry-miller-trimble:

`bd mol bond` gates the formula-registry fallback on `looksLikeFormulaName` (`cmd/bd/mol_bond.go`), which only returns true for names starting with `mol-`, containing `.formula`, or containing a path separator. A plain formula name like `download-arxiv` matches none of these, so `resolveOrDescribe` / `resolveOrCookToSubgraph` short-circuit to `'<name>' not found (not an issue ID or formula name)` **before** `parser.LoadByName` is ever reached.

### Sibling commands don't have this gate
`looksLikeFormulaName` has only these two callers. Every sibling resolves plain names directly via the parser: `bd formula show` (`cmd/bd/formula.go`), `bd cook` (`cmd/bd/cook.go`), `bd mol seed`/`bd mol pour` (via `resolveAndCookFormulaWithVars`). So `mol bond` is the lone inconsistent path.

### Fix quality — correct, minimal
Deletes the gate and lets `parser.LoadByName` decide; genuinely-unknown names still produce an accurate "not found as issue or formula" error matching siblings. No other callers of the removed helper; `strings` import still used elsewhere; formula resolution is first-match-wins over search paths identically for bond and siblings, so there's no divergent ambiguity handling to break.

### Tests — adequate
`cmd/bd/mol_bond_gwn_test.go` (cgo-tagged) sets up a formula, proves it via `bd formula show`, then asserts `bd mol bond <name> <id> --dry-run` succeeds — reproduces the exact main-branch error.

### Minor, non-blocking
`--dry-run` exercises `resolveOrDescribe` but not `resolveOrCookToSubgraph` (the real cook path); both received the identical one-line edit, so risk is low.

PR #3875, opened 2026-05-11
fix(cook): drop duplicate blocks edge when waits_for infers from needs[0] (GH#3783)

### LINKED ISSUE #3783: pour fails with dep-type collision when a step has both `needs` and `waits_for`
## Summary

Pouring a formula with both `needs` and `waits_for` on the same step fails: the implicit-spawner inference for `waits_for` produces a dep edge that collides with the one already created from `needs`.

## Reproduction

```toml
formula = "survey"
version = 1
type = "workflow"

[[steps]]
id = "survey-workers"
title = "Survey"

[[steps]]
id = "wait"
title = "Wait for fanout"
needs = ["survey-workers"]
waits_for = "all-children"
```

```
$ bd mol pour survey
Error: pouring proto: failed to create dependency: dependency smk-mol-a7t -> smk-mol-t6b already exists with type "blocks" (requested "waits-for"); remove it first with 'bd dep remove' then re-add
```

## Root Cause

`collectDependencies` in `cmd/bd/cook.go:894-955` creates two edges with the same `(IssueID, DependsOnID)` pair:

- `needs` (line 914-925) → `{wait → survey-workers, blocks}`
- `waits_for` (line 927-955), with no explicit spawner, infers `spawnerStepID = step.Needs[0]` at line 935 → `{wait → survey-workers, waits-for}`

Storage only permits one dep type per pair, so the second insert fails.

## Impact

Blocks the canonical fanout-coordinator pattern: a wait step that `needs` the spawner step and `waits_for: "all-children"` of it. The implicit-spawner inference at line 933-936 was added so users wouldn't have to name the spawner twice, but now it collides with the explicit `needs` edge instead of cooperating with it.

## Fix

Not obvious — three semantically distinct options:

1. Treat `waits-for` as strictly stronger than `blocks`; replace the `blocks` edge when both apply to the same pair. Requires confirming `waits-for` is treated as blocking everywhere `blocks` is checked.
2. Skip the `waits-for` edge if `blocks` already covers the pair. *Wrong* — `blocks` unblocks when the spawner closes, but the wait step needs to also gate on the spawner's spawned children.
3. Validate-time reject and require an explicit `waits_for` spawner ID distinct from any `needs` entry.

---

*Discovered while smoke-testing primitives for #3621.*

jjgarzella:
## What

A step with both `needs = [X]` and `waits_for = "all-children"` (no explicit `from:`) used to crash `bd mol pour` with a dep-type collision. The `waits_for` spawner gets inferred from `needs[0]`, so both `needs` and `waits_for` emit edges on the same `(source, target)` pair — one as `DepBlocks`, one as `DepWaitsFor` — and storage rejects the duplicate. Pre-compute the `waits_for` spawner before processing `needs` and skip the redundant `DepBlocks` for that one target.

## Why

Fixes #3783.

The collision blocks the documented pattern where a step gates on the children of a step it depends on (the canonical "wait for the fanout from the surveyor" shape). Before this fix:

```
$ bd mol pour qnt-repro
Error: pouring proto: failed to create dependency:
  dependency qnt-mol-XXX -> qnt-mol-YYY already exists with type "blocks"
    (requested "waits-for"); remove it first with 'bd dep remove' then re-add
```

The `DepWaitsFor` edge already encodes the blocking semantics plus the gate metadata, so the `DepBlocks` edge from `needs` for that specific target is redundant once they collapse onto the same pair.

## Verification

```
go test -tags 'gms_pure_go cgo' -run TestBdMolPour_NeedsAndWaitsForSameTarget -v ./cmd/bd/
```

New e2e regression test at `cmd/bd/cook_qnt_test.go` pours a formula with the exact collision shape and asserts `bd mol pour` succeeds. Fails on `main` with the verbatim error; passes on this branch.

Sanity checks: cook test family (`TestSubstituteFormulaVars`, `TestSubstituteStepVarsRecursive`, `TestCompileTimeVsRuntimeMode`, `TestCreateGateIssue*`, `TestCookFormulaToSubgraph_*`) all pass. `go vet -tags 'gms_pure_go cgo' ./cmd/bd/` clean.

### DIFF
--- a/cmd/bd/cook.go
+++ b/cmd/bd/cook.go
@@ -910,8 +910,32 @@ func collectDependencies(step *formula.Step, idMapping map[string]string, deps *
 		})
 	}
 
+	// Pre-compute the waits_for spawner so we can dedupe against needs below.
+	// When waits_for has no explicit `from:`, it infers its spawner from
+	// needs[0] — and `needs` would otherwise emit a DepBlocks edge on the same
+	// (source, target) pair that `waits_for` emits a DepWaitsFor edge on.
+	// Storage rejects the duplicate. The DepWaitsFor edge subsumes the
+	// blocking semantics, so we skip the redundant DepBlocks for that
+	// specific target (GH#3783).
+	var waitsForSpec *formula.WaitsForSpec
+	var waitsForSpawnerStepID string
+	if step.WaitsFor != "" {
+		waitsForSpec = formula.ParseWaitsFor(step.WaitsFor)
+		if waitsForSpec != nil {
+			waitsForSpawnerStepID = waitsForSpec.SpawnerID
+			if waitsForSpawnerStepID == "" && len(step.Needs) > 0 {
+				waitsForSpawnerStepID = step.Needs[0]
+			}
+		}
+	}
+
 	// Process needs field - simpler alias for sibling dependencies
 	for _, needID := range step.Needs {
+		if needID == waitsForSpawnerStepID {
+			// This target is also the waits_for spawner; the DepWaitsFor edge
+			// emitted below subsumes the blocking semantics for this pair.
+			continue
+		}
 		needIssueID, ok := idMapping[needID]
 		if !ok {
 			continue // Will be caught during validation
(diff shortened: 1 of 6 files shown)


harry-miller-trimble:
## Automated agent review — NEEDS-WORK (real crash fix, but a semantic gap)

> This is an automated agent review, not a maintainer approval. It is posted to help move the PR forward, per PR_MAINTAINER_GUIDELINES.

Thanks for chasing this — the crash is real and the de-dup mechanics are sound. There's one correctness gap worth resolving before merge.

### The bug is real (confirmed on main)
`collectDependencies` emits a `DepBlocks` edge for every `needs` entry (`cmd/bd/cook.go`), and a `waits_for` with no explicit `from:` infers its spawner from `needs[0]` and emits a `DepWaitsFor` edge on the **same** `(issue, spawner)` pair. Storage rejects a second edge of a different type on the same pair (`internal/storage/issueops/dependencies.go`), so `bd mol pour` hard-fails. Genuine bug.

### The concern: `waits_for` does not subsume `blocks` in the pre-fanout window
The fix drops the `blocks` edge on the premise that the `waits_for` edge already encodes the blocking. But the waits-for gate (`internal/storage/issueops/blocked_state.go`, `waitsForGateBlockedSQL`) blocks the waiter **only when the spawner has ≥1 open parent-child child** — it never blocks merely because the spawner itself is still open. At pour time (and until the spawner runs and fans out children), the spawner has zero children, so the gate evaluates to not-blocked. With the `blocks` edge removed, the waiter can become **ready before the spawner completes** — the opposite of what `needs` requested. On main the `blocks` edge kept it blocked until the spawner closed; the two edges are complementary.

### Two smaller points
- The skip is only in the `needs` loop; `depends_on` + an explicit matching `waits_for` would still collide (`cook.go` `depends_on` loop is untouched).
- `cmd/bd/cook_qnt_test.go` only asserts `mol pour` succeeds / no collision string — it never checks post-pour blocked state, so it wouldn't catch the premature-readiness regression.

### Suggested path
Make `waits_for` genuinely strictly-stronger than `blocks` so collapsing them is lossless: extend `waitsForGateBlockedSQL` to also block while the spawner itself is open (not just when it has open children), and add a test asserting the waiter stays blocked while the spawner is open with no children yet. Then dropping the `blocks` edge is safe. Happy to re-review once that's in.

maphew:
Your diagnosis was correct and your fix is landing — the collision was real, the de-dup was the right mechanism, and your regression test came through the rebase intact. Thanks for the patience; this sat far longer than it should have.

Since the July 14 review went unanswered I've completed the branch per that review rather than let it strand (your commit is preserved as-is at the base of the stack; three commits total now):

1. **Rebase onto current main** — one conflict, from main's new `types.NewWaitsForDependency` helper; your logic carried over unchanged.
2. **The same skip applied to the `depends_on` loop**, which had the identical collision your `needs` fix covered.
3. **The semantic gap closed, provenance-scoped.** The review's concern was that dropping the `blocks` edge opens a pre-fanout window (waiter ready before the spawner even runs). Extending the gate SQL globally turned out to break the *documented* all-children semantics (`TestIsBlocked_WaitsForDefaultGate`: fanout-completion, not spawner status, is the readiness signal). So instead: when cook collapses a `needs`/`depends_on` edge into the waits-for edge, that edge's metadata now carries `also_blocks: true`, and only flagged edges additionally block while the spawner itself is open — including overriding the any-children early-open, since your `needs` meant "spawner must close." Plain waits-for edges keep byte-identical metadata and exactly their prior behavior.
4. **A pre-existing bug your test flushed out**: `cloneSubgraph` dropped `dep.Metadata` when pouring templates, silently resetting every poured waits-for gate to the default — it would already have broken `any-children` gates in formulas. Fixed; that one's a genuine extra find this PR gets credit for forcing into the light.

Full `TestIsBlocked` family (18 subtests), your pour test, a new spawner-open pour test, and the cook/mol sweeps all pass. CI runs next; once green this merges with your authorship at the base.

PR #3781, opened 2026-05-07
feat(formula): make primitives discoverable to coding agents (GH#3621)

jjgarzella:
## What

Makes the formula primitive surface discoverable to coding agents, in three layers:

- **`bd formula schema [primitive]`** (alias: `primitives`) — reflection-driven
  index of every exported struct in `internal/formula/types.go`, with field
  types, json/toml tags, and doc comments. Generated via `go:generate` from an
  AST walk; a `TestSchemaGenIsCurrent` gate fails CI if the generator output
  drifts from the committed `schema_gen.go`.
- **`bd formula --help`** (and `list --help`, `show --help`) now points agents
  at `bd formula schema`, the curated examples directory, and the website
  primitives reference. A scripttest snapshot at
  `cmd/bd/testdata/formula_help.txt` is the regression gate.
- **`examples/formulas/primitives/`** — one minimal `.formula.toml` per
  empirically wired primitive (`loop-count`, `loop-range`, `children-epic`,
  `branch-fanin`, `condition`, `gate-timer`), backed by a Go smoke harness
  (`cmd/bd/formula_primitives_test.go`) that parses and cooks each fixture and
  asserts the primitive's effect on the cooked subgraph. Adding a fixture
  without a registered assertion is a deliberate test failure.

## Why

Closes #3621 (items 1-3 of the contributor comment). Without a discoverable
surface, agents authoring formulas fall back to imperative substitutes - or
simply deny that primitives like `Compose.Branch` or `Step.Gate` exist when no
user formula in the project happens to use them. The validator's struct
definitions in `internal/formula/types.go` are the de facto source of truth;
this PR surfaces them where agents will actually look, with CI gates that
prevent the discoverability layer from itself becoming misleading.

`on_complete` is intentionally omitted from the curated examples — it's a
runtime no-op for every valid configuration as of v1.0.3 and ships with the
resolution of beads-i02, not this epic. Three other related bugs found during
the audit (beads-4pi, qnt, e1u) are tracked separately and explicitly out of
scope.

Maintainer follow-up in this branch also makes the CLI/docs wording explicit
that `bd formula schema` is structural reference, while
`examples/formulas/primitives/` is the smoke-tested authoring surface.

## Verification

```sh
make build

# Schema index
./bd formula schema                     # 18 schema structs listed
./bd formula schema loop                # LoopSpec fields w/ doc comments
./bd formula primitives gate            # alias resolves identically
./bd formula schema --json | jq length  # 18

# Help text pointers
[...]

PR #3876, opened 2026-05-11
feat(formula): implement on_complete runtime fanout executor (GH#3782)

### LINKED ISSUE #3782: on_complete is a runtime no-op, with docs pointing at a nonexistent field
## Summary

`OnCompleteSpec` parses, validates, and round-trips through cook output, but is dropped at pour time and never read by any executor — closing a step with `on_complete` set produces zero side effects regardless of configuration. Meanwhile the docs and `bd info` advertise it as a working feature, and the docs example uses an `on_complete.run` field that doesn't exist on the struct.

## Reproduction

```toml
# .beads/formulas/mol-worker.formula.toml
formula = "mol-worker"
version = 1
type = "workflow"
[vars.worker_name]
required = true
[[steps]]
id = "do-work"
title = "Worker {{worker_name}}"
```

```toml
# .beads/formulas/survey.formula.toml
formula = "survey"
version = 1
type = "workflow"
[[steps]]
id = "survey-workers"
title = "Survey workers"
[steps.on_complete]
for_each = "output.workers"
bond = "mol-worker"
[steps.on_complete.vars]
worker_name = "{item.name}"
```

```
$ bd mol pour survey
$ bd update <step-id> --metadata '{"output":{"workers":[{"name":"alice"},{"name":"bob"}]}}'
$ bd close <step-id> --reason "test"
$ bd list --all --json | jq length
2   # expected 4 (parent + step + 2 worker mols)
```

Eight `on_complete` shapes tested (`for_each`+`bond`, empty, `vars` only, `parallel`, `sequential`, `run`, plus the two validator-rejected shapes); every shape the validator accepts produces zero runtime effect.

## Root Cause

Three converging bugs:

1. **Pour drops the spec.** `processStepToIssue` (`cmd/bd/cook.go:533`) handles every step field except `step.OnComplete`. Compare `step.Gate` at line 625-626, which dispatches to `createGateIssue`.
2. **No runtime executor exists.** Grep for `OnComplete|on_complete|ForEach|for_each` outside tests yields only the struct/validator (`internal/formula/types.go`), two cook-time clones (`advice.go`, `controlflow.go`), and a marketing string (`cmd/bd/info.go:971`). `cmd/bd/close.go`, `internal/storage/hook_decorator.go`, and `internal/hooks/hooks.go` have zero references. Placeholder substitution for `{item}`/`{item.field}`/`{index}` (documented at `internal/formula/types.go:351-353`) is unimplemented.
3. **Docs example uses a nonexistent field.** `website/docs/workflows/molecules.md:118-129` shows `[steps.on_complete] run = "make build"`, but `OnCompleteSpec` (`internal/formula/types.go:339`) has no `Run` field. The TOML decoder silently drops the unknown field, the validator accepts the empty spec, and nothing runs. Same example in `website/versioned_docs/version-1.0.0/workflows/molecules.md:118-129`.

## Impact

An agent reads `bd info` ("NEW: Runtime expansion - on_complete and for-each dynamic step generation" at `cmd/bd/info.go:971`), the workflows docs, or the `OnCompleteSpec` doc-comments, writes a valid-looking formula, watches every stage succeed silently, and gets nothing. This is the cleanest example of the discoverability failure mode #3621 documents — surfaced in our own first-party surface.

## Fix

Two paths:

**Implement.** Persist `OnCompleteSpec` at pour time (extend `processStepToIssue` analogously to `step.Gate`); read it from a close hook (`autoCloseCompletedMolecule` or `HookOnClose`); resolve `for_each` against `issue.Metadata.output.<path>`; implement `{item}`/`{item.field}`/`{index}` substitution against `Vars`; dispatch via `bondProtoMol` per item; honor `Parallel`/`Sequential`. Add a primitive fixture and smoke test asserting spawned mols.

**Remove.** Delete `OnCompleteSpec`, `validateOnComplete`, the `cloneOnComplete` calls in `advice.go`/`controlflow.go`, the `bd info` line, the docs sections (current + v1.0.0 mirrors), and the `CHANGELOG.md:2503` entry. Add a regression test asserting the type is gone.

Either way, fix the docs `run` example as part of the same PR.

---

*Discovered while triaging #3621 (formula primitives discoverable to coding agents).*

jjgarzella:
## What

Implements the runtime executor for `OnCompleteSpec`. The spec parsed + validated + survived cook output on `main`, then was silently dropped at pour time and read by no executor — every valid `on_complete` configuration was a no-op. After this PR, closing a step with `on_complete` set spawns one molecule per element of the step's runtime-discovered `output.<path>`, with `{item}` / `{item.field}` / `{index}` substituted into the spawned mols' vars.

Touches:

- `cmd/bd/cook.go` — `processStepToIssue` merges `step.OnComplete` into `issue.Metadata["on_complete"]` at pour time. Cohabits with existing `step.Metadata`; `bd update --metadata`'s deep-merge keeps the spec safe from agents writing `output.<path>` at runtime.
- `cmd/bd/on_complete.go` (new) — `executeOnCompleteFanout`: reads the persisted spec, resolves `for_each` against the issue's metadata, hard-errors on missing path / non-array, substitutes placeholders per iteration, dispatches `bondProtoMolWithSubgraph` per item. Honors `parallel` (default) vs `sequential` (threads `blocks` deps between consecutive spawned roots). Partial bond failures accumulate into `metadata.on_complete_failures`.
- `cmd/bd/close.go` and `cmd/bd/update.go` — fire the executor after a successful close, so both `bd close <id>` and `bd update <id> --status closed` trigger fanout (universal coverage).
- `internal/formula/types.go` — `OnCompleteSpec` docstring rewritten to describe actual runtime behavior; `schema_gen.go` regenerated so `bd formula schema OnCompleteSpec` surfaces the rewritten text.
- `examples/formulas/primitives/on-complete-fanout.formula.toml` (new) + assertion in `cmd/bd/formula_primitives_test.go` — closes the loop on PR #3781's "every primitive earns a fixture + smoke gate."
- `website/docs/workflows/molecules.md` — replaces the bogus `on_complete.run = "make build"` example with the canonical `for_each` / `bond` / `vars` pattern.
- `CHANGELOG.md` — `Unreleased` `Fixed` entry pointing at #3782.

## Why

Today, every beads formula that needs to fan out over a runtime-discovered list (the *only* class of fanout the schema has no declarative primitive for — `loop.count` / `loop.range` / `expand` / `branch` / `children` are all compile-time, and `waits_for` gates a fanout but doesn't produce one) has to encode the fanout in **natural-language prose**, instructing the agent to call `bd mol bond` once per item with hand-built `--ref` and `--var` flags. Every word of that prose recipe is brittle: the agent has to remember the pattern, parse the IDs correctly, escape shell-unsafe values, and handle partial failures itself. The same shape repeats anywhere a step needs to spawn `N` children where `N` is only known after the step has run.

The minimal working example lives in `cmd/bd/cook_c7g_test.go` (`TestOnComplete_RuntimeFanout_EndToEnd`): a `survey` formula declares the fanout in five lines of declarative spec —

```toml
[steps.on_complete]
for_each = "output.items"
bond = "worker-arm"
parallel = true

[steps.on_complete.vars]
worker_name = "{item.name}"
```

— pours, populates `metadata.output.items = [{name:"alice"},{name:"bob"},{name:"carol"}]`, and on close spawns three `worker-arm` molecules whose `worker_name` var has been substituted from `{item.name}`. The downstream `join` step that `needs` the survey and gates on the spawned children via `waits_for = "all-children"` is satisfied without an agent in the loop.

`on_complete` was designed to be this kind of declarative replacement for prose-driven fanout, and was the only formula primitive that *appeared* to be wired (validator, schema docs, `bd info` marketing) but actually wasn't — surfacing an entire class of agent-failure-mode (filed as GH#3621): an agent reads `bd info` and the workflow docs, writes the declarative shape, watches every stage succeed, and silently produces zero spawned mols. The user has to catch the deviation manually.

This PR closes that gap. Fixes #3782.

### Design decisions (worth flagging in review)

| Decision | Choice | Why |
|---|---|---|
| Spec persistence | `issue.Metadata["on_complete"]` (structured, deep-merged) | Mirrors the shape of the spec itself; `bd update --metadata` already deep-merges so agents writing `output.<path>` at runtime don't wipe it. Considered label-encoded (à la `gate:<await>`) — too awkward for structured data. |
| Trigger location | After every successful close (`cmd/bd/close.go` *and* `cmd/bd/update.go`) | "When this step closes, regardless of how" semantics. Storage-layer hook isn't viable because `bd update --status closed` calls `UpdateIssue`, not `CloseIssue`. |
| Missing `output.<path>` at close | **Hard error** (stderr after close) | Silent zero-spawn was the original failure mode. Loud failure surfaces the misconfiguration immediately. |
| `for_each` value is not an array | Hard error | Schema is explicit it's a list. |
| Partial bond failure | Continue, accumulate into `metadata.on_complete_failures` | Bonds aren't transactional; rollback would mean deleting already-created mols, which is messier than leaving them visible with a failure note. |

### Dependencies

This PR is developed against a local merge of four other open PRs and **should land after all four merge**. The branch is rebased onto `main` as each one lands upstream.

- **#3874** (`fix(mol bond): resolve plain formula names like every other subcommand`) — strong dep: `spec.Bond` is a formula name; on main the resolver rejects names that don't start with `mol-` or contain `.formula`. Without #3874, `on_complete` only works for `mol-`-prefixed bond formulas. The MWE test in this PR uses `worker-arm` (no `mol-` prefix), which requires #3874.
- **#3781** (`feat(formula): make primitives discoverable to coding agents`) — strong dep: provides the `examples/formulas/primitives/` directory + smoke harness + reflection-driven `bd formula schema`. The on-complete-fanout fixture and assertion in this PR slot into that harness; the rewritten docstring is surfaced via `bd formula schema OnCompleteSpec`.
- **#3875** (`fix(cook): drop duplicate blocks edge when waits_for infers from needs[0]`) — coupling dep: the canonical fanout-join recipe (a join step with `needs=[surveyor]` + `waits_for=all-children` gating on the spawned children) hits the collision unless #3875 is in. The MWE test in this PR includes the join step and exercises that path.
- **#3785** (`fix(template): persist labels in cloneSubgraph`) — coupling dep only (was strong under label-based persistence, weakened to coupling once we chose metadata-based persistence). Still useful since the canonical recipe uses `waits_for`'s `gate:` label downstream.

## Verification

End-to-end:

```
go test -tags 'gms_pure_go cgo' -run TestOnComplete_RuntimeFanout_EndToEnd -v ./cmd/bd/
```

`cmd/bd/cook_c7g_test.go` writes a workspace with a `survey` formula (with `on_complete{for_each=output.items, bond=worker-arm}`) and a `worker-arm` formula, pours, populates `metadata.output.items = [{name:"alice"},{name:"bob"},{name:"carol"}]`, closes the survey step, and asserts:

1. Three worker-arm molecules now exist with titles `"Worker alice"`, `"Worker bob"`, `"Worker carol"` (substitution works).
2. The spawned worker-arm roots are direct children of the closed survey step (bond target is correct).

Fails on `main` (pour + close succeed with zero side effects); passes on this branch.

Smoke harness:

```
go test -tags 'gms_pure_go cgo' -run TestFormulaPrimitiveExamples -v ./cmd/bd/
```

Adds `examples/formulas/primitives/on-complete-fanout.formula.toml` + a case asserting `OnCompleteSpec` round-trips through cook onto `issue.Metadata["on_complete"]` (cook-time gate). Runtime gate is the end-to-end test above.

Discoverability:

```
bd formula schema OnCompleteSpec
```

Surfaces the rewritten docstring (persistence shape, hard-error semantics, sequential/parallel) reflected from `internal/formula/types.go`.

Adjacent test families pass: `TestCookFormulaToSubgraph_*`, `TestCreateGateIssue_*`, `TestSubstituteFormulaVars_*`, `TestBdMolBond_ResolvesPlainFormulaName`, `TestBdMolPour_NeedsAndWaitsForSameTarget`, `TestSchemaGenIsCurrent`. `go vet -tags 'gms_pure_go cgo' ./cmd/bd/` clean.


jjgarzella:
Disposition of this PR is blocked on a cross-repo design decision about the canonical runtime-discovered fan-out primitive (`on_complete` vs Gas City's `drain`).

Canonical thread (please discuss there to avoid splitting across repos): gastownhall/gascity#2947

This PR corresponds to **Option A (status quo plus)** in that issue — merging it gives standalone `bd` users a working `on_complete` executor, at the cost of a permanent dialect divergence from Gas City's `internal/dispatch/fanout.go` implementation. If the decision lands on B or C, this PR is closed or repointed accordingly.

maphew:
The executor here is careful work: pour-time cohabitation of the spec inside `issue.Metadata["on_complete"]` so agents writing `output.<path>` at runtime can't clobber it, the `parallel`-vs-`sequential` distinction threading `blocks` deps between consecutive spawned roots, and partial bond failures accumulated into `metadata.on_complete_failures` instead of aborting the batch — that's the level of edge-case thinking runtime fanout actually needs. Thank you for building it, and equally for opening gastownhall/gascity#2947 yourself and framing this PR honestly as "Option A" there rather than pushing for a quiet merge.

That framing is also why this can now be resolved cleanly: the maintainer design ruling in gascity#2947 (2026-06-28) went the other way — `on_complete` is scoped **out** of beads, and Gas City's `drain` primitive is the canonical runtime-discovered fanout. So we won't be merging the executor into beads, not because the implementation fell short, but because the project decided beads stores and relates work while the orchestration layer executes it.

Where the rest of this stack landed, so nothing is stranded:

- (1) Schema-primitives index + curated fixtures — **merged** via your PR 3781.
- (2) `bd mol bond` formula-name resolver — **merged** via your PR 3874.
- (3) cloneSubgraph label persistence — superseded on main (shared `PersistLabels` path); PR 3785 retired with your diagnosis credited.
- (4) The `needs` + `waits_for` dep-collision crash fix — your PR 3875 is being completed on its own branch right now (rebase + the waits-for-gate strengthening from the July review), aiming to merge it with your authorship intact.
- (5) The remaining beads-side problem this PR responded to — the schema advertising an `on_complete` that silently does nothing (issue 3782) — stays open, but its resolution direction is now the ruling's: warn or remove the misleading no-op spec rather than wire an executor. We're tracking that follow-up.

I'm leaving this PR open for now so you have room to respond or salvage anything you disagree with; it will be closed in a few days otherwise. If you want to carry the executor design into Gas City's `drain` implementation, gascity#2947 is the thread — the substitution semantics (`{item}` / `{item.field}` / `{index}`) and the failure-accumulation pattern you built here deserve to survive the venue change.

---
Review provenance: author-clustered maintainer sweep (all five of your open/stacked PRs reviewed together); design authority is the gascity#2947 ruling, not a fresh code review of this branch.

jjgarzella:
Makes sense to me, feel free to close.

maphew:
Thanks for the graceful close-out — and again for routing the design question to gastownhall/gascity#2947 where it belongs. If the orchestration-layer answer there ever calls for beads-side primitives, we would be glad to see you back with them. Closing as discussed.

PR #3883, opened 2026-05-11
fix: TLS/config propagation for server mode + LabelsAny in GetReadyWork

jmf-pobox:
## Problem

Teams running bd against a shared Dolt server (Hosted DoltDB or self-hosted) across multiple repos hit three gaps where features worked for embedded Dolt but were missing or broken in server mode:

1. `bd init --dolt-mode server` fails with "server does not allow insecure connections" even when TLS is configured via env var or `~/.config/beads/server.json`
2. Runtime commands (`bd list`, `bd ready`, etc.) silently ignore central server config — TLS, host, port, user from `server.json` are dropped
3. `bd ready` ignores `directory.labels` auto-scoping (GH#541) — returns all repos' issues instead of scoping to the current directory

## Root Cause

1. `cfg.ServerTLS` never set from env/central config before `newDoltStore` in init.go
2. `PersistentPreRun` loads `configfile.Config` but never calls `applyCentralConfigDefaults`
3. `GetReadyWork` handles `Labels` (AND) and `ExcludeLabels` but not `LabelsAny` (OR) — the filter `directory.labels` uses. Also `DoltServerStore.GetReadyWork` was `panic("unimplemented")`

## Fix

**Commit 1** — `fix(dolt): propagate TLS, port, and central config at init and runtime`
- Export `ApplyCentralConfigDefaults` and `ApplyEnvAndCentralDefaults` wrappers
- Call them from init.go and main.go PersistentPreRun
- Include ServerPort propagation (Bugbot catch: password lookup used port 0)

**Commit 2** — `feat(ready): support LabelsAny filter + implement server-mode GetReadyWork`
- Add LabelsAny WHERE clause to DoltStore.GetReadyWork (queries.go) and GetReadyWorkInTx (issueops/ready_work.go)
- Implement GetReadyWork + GetBlockedIssues for DoltServerStore (was panic)

## Tests

- `TestGetReadyWork_LabelsAnyFilter` — regression test: creates labeled issues, asserts LabelsAny filters correctly. Fails without patch, passes with.
- `TestGetDirectoryLabels` — 3 subtests: exact match, subdirectory match, no match
- `TestApplyEnvAndCentralDefaults` — verifies TLS propagation from server.json

83 lines of production code, 167 lines of tests, across 9 files.

## Who Benefits

Anyone using bd in server mode (Hosted DoltDB, self-hosted `dolt sql-server`) with multiple repos sharing one database — the exact setup `directory.labels` (GH#541) was designed for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`bd create` uses per-repo prefix in shared-server mode** — `ReadConfigPrefix` now reads `issue_prefix` from `metadata.json` before falling back to the DB `config` table. In shared-server setups where multiple repos share one Dolt database, the global `issue_prefix` row no longer overrides per-repo prefixes. Precedence: config.yaml `issue-prefix` > metadata.json `issue_prefix` > DB config table.
+- **`bd ready` honors `LabelsAny` filter** — `GetReadyWork` in all three store implementations (DoltStore, EmbeddedDoltStore, DoltServerStore) now applies `WorkFilter.LabelsAny`, fixing `directory.labels`-based auto-scoping for `bd ready` (was already working for `bd list`).
+- **TLS and central config propagated at init and runtime** — `bd init --dolt-mode server` and runtime commands now read TLS, host, port, user from `~/.config/beads/server.json` and environment variables.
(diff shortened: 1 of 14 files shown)


coffeegoddd (commented):
Please file an issue for this if you havent already, but the storage layer is currently in flux, so none of this existing code will persist. Once we have the new storage layer up, we'll fix the issue.

jmf-pobox:
Closing — superseded by factory-centralized implementation on punt-labs/beads branch factory-fixes. Filing individual issues for the 3 bugs per maintainer ask.

PR #3422, opened 2026-04-22
feat(nix): turn packages into an overlay to allow for overriding

DylanRJohnston:
## Summary
Turn the nix packages into an overlay so they can be more easily overridden.

## Details
Currently the exported bd / default package is a stdenv.mkDerivation wrapper around the `callPackage` beads package which makes it impossible to use the regular `bd.override` pattern provided by callPackage. By restructuring it into an overlay first, and providing that overlays as an output downstream consumers can more easily consume the package with their own nixpkgs version as well as override its inputs. 

e.g.

```nix
final: prev: {
  beads-unwrapped = prev.beads-unwrapped.override (prevArgs: {
    goBuildModule = args: prevArgs.goBuildModule (args // { vendorHash = "someNewhash"; });
  });
}

PR #3910, opened 2026-05-12
feat(nix): expose packages as an overlay for downstream override

harry-miller-trimble:
## Summary

Focused overlay-only follow-up to #3422 against current `main` (post-#3491). Restructures the nix packaging so downstream consumers can override `beads` inputs with the standard `callPackage` `.override` pattern and consume beads with their own nixpkgs version.

```nix
final: prev: {
  beads-unwrapped = prev.beads-unwrapped.override (prevArgs: {
    buildGoModule = args:
      prevArgs.buildGoModule (args // { vendorHash = "..."; });
  });
}
```

## What's in the PR

- **`overlay.nix` (new)**: exposes `beads-unwrapped` (the `buildGoModule` output) and `beads` (the wrapper with shell completions baked in, plus per-shell completion passthru attrs).
- **`default.nix`**: drop the `self` flake-ref input and use `src = ./.;` so the package is consumable via the overlay outside the flake context.
- **`flake.nix`**: expose `overlays.default` and apply the overlay to the nixpkgs instance used for `packages` / `devShells`.
- **`packages.nix`**: thin wrapper that pulls outputs from the overlayed package set.
- **`.github/workflows/nix-build.yml`**: include `overlay.nix` and `packages.nix` in the path filter so overlay-only changes still trigger CI.

## What was intentionally left out from #3422

The ICU / version / env-var rearrangements in the original PR no longer apply on current `main`:

- ICU is gone (`gms_pure_go` build tag handles regex)
- `proxyVendor = true` is in place
- The env-var plumbing the original PR rearranged has been reshaped upstream
- `version` is now `1.0.4`, not `1.0.2`

So this PR is overlay-only — ~30 lines of meaningful diff instead of ±67 with conflicts.

## Attribution

The overlay design and the shell-completions passthru pattern come from #3422 by @DylanRJohnston. Credited via `Co-authored-by:` trailer on the commit. The original PR is closed (contributor confirmed they cannot revisit), so this is the maintainer-implemented version we offered to take on internally.

## Test plan

- [ ] CI `nix build .#default` passes (path filter now triggers on overlay/packages changes too)
- [ ] `nix flake show` lists `packages.<system>.{default,bd,fish-completions,bash-completions,zsh-completions}` and `overlays.default`
- [ ] `nix build .#fish-completions` (and bash/zsh) produces a derivation containing only the relevant completion file
- [ ] Downstream `.override` smoke test: in a consumer flake, `pkgs.beads-unwrapped.override { ... }` rebuilds with the override applied


DylanRJohnston (approved):

DylanRJohnston:
Thanks for updating the changes for main!

maphew:
Maintainer fix-merge: merged current main into this branch (1ad3b303e) and reconciled the overlay against the nix changes that landed since May, most recently #4609.

What needed manual care beyond the one textual conflict in `packages.nix`:

- `default.nix`: the auto-merge silently kept this branch's `src = ./.` while main had moved to `src = self;` - restored main's form; main's current version and vendorHash are preserved verbatim (`default.nix` now has zero diff against main).
- `overlay.nix`: carried in main's later additions to the wrapped derivation (`env.BD_DISABLE_METRICS`, `env.BD_DISABLE_EVENT_FLUSH`, `beads` alias was already there), and curried the overlay on `self` so `callPackage ./default.nix` can supply it.
- `packages.nix`: kept your thin re-export shape while preserving the pre-existing flake surface (`default`, `bd`, `*-completions`) - downstream `nix build .#bd` keeps working, and `overlays.default` is new surface.

No local nix available here, so validation is structural; the nix-build workflow on this PR is the real gate. Will merge once CI is green. Thanks for the overlay work and your patience - this sat while main drifted under it.

maphew:
Merged with the full suite green, including the nix flake checks. The overlay is now on main with the flake surface (default/bd/completions) preserved and overlays.default exposed for downstream override. Thanks for the design - sorry it took two months of main drifting before it landed.

PR #3913, opened 2026-05-12
docs: add CONTRIBUTING_PR_GUIDELINES.md (codify reviewer expectations for contributors)

quad341:
Over the past several weeks I've observed a consistent pattern in coffeegoddd's review feedback across 11 open PRs: minimal beads-only repros, one layer per PR, primitives before application wiring, and commit-message-over-inline-comment style. This PR codifies that pattern as `CONTRIBUTING_PR_GUIDELINES.md`, a companion to the existing `PR_MAINTAINER_GUIDELINES.md`, so that all contributors — human and agent alike — can understand what makes a PR land vs bounce before they open it.

This PR itself follows its own rules: single file, no inline comments, beads-only change, diff is under 50 lines.


quad341:
The single failing CI job (**Test (ubuntu-latest)**) is an **unrelated flake** in `internal/audit::TestAppend_ConcurrentWritersUniqueIDs` — nothing in this docs-only diff (`CONTRIBUTING_PR_GUIDELINES.md` + `release-gates/be-z33z-gate.md`) touches the `audit` package or any code at all.

**Root cause of the flake** — `audit.newID()` in `internal/audit/audit.go:160-166` returns 32-bit random IDs:

```go
var b [4]byte
rand.Read(b[:])
return idPrefix + hex.EncodeToString(b[:]), nil
```

The test generates 2000 IDs (`writers=8 × entriesPerWriter=250`). Birthday-paradox collision probability for n=2000 IDs in N=2³² space:

> P ≈ n(n-1)/(2N) = 2000·1999 / (2·4.29×10⁹) ≈ **4.65 × 10⁻⁴ ≈ 1 in 2,150 runs**

The codecov report showing exactly one duplicate (`int-3f25833f`) in 1999 unique returned IDs out of 2000 is the textbook single-collision signature.

**Empirically reproduced** locally with the same 8-writer × 250-entry parameters: **3 collisions in 5,000 trials (0.06%)** — matches the model.

Two earlier hypotheses ruled out:
- *Torn writes* — the test fails at the in-memory `returnedIDs` check (line 96, before the file is even opened), so a file-level corruption can't be it.
- *Shared counter race* — the test's own preamble notes (correctly) that `newID()` uses `crypto/rand` with no shared state. My repro harness uses *only* the random generator with no counter and reproduces at the same rate.

Filing a separate `ready-to-build` bead to widen the ID to 16 bytes (`int-<32hex>` — still opaque JSON; no parsers anywhere check length). That change makes the collision probability vanishingly small for any realistic audit-log size.

Re-running the failed job here — the docs PR itself is otherwise clean and should pass on the rerun.

— *beads/investigator (be-tjt9)*

PR #3914, opened 2026-05-12
schema: print migration name to stderr on apply (human UX only)

quad341:
## What this changes

When `bd` opens a repo that needs schema migrations, it now prints one line per migration to **stderr**:

```
migrating schema: 0033_add_date_indexes.up.sql
```

That's the entire change — a single `fmt.Fprintf` to a swappable `io.Writer` var.

## Why

Users watching the terminal during a 25–75 s migration see a silent CLI and can't tell if it's hung or working. This one-liner makes the wait legible.

## What this is NOT

This is **not** a fix for orchestrator timeouts. If an orchestrator kills `bd` because it assumes a running command is hung, the right fix is to increase the orchestrator's timeout budget. Adding log lines doesn't help if the timeout fires anyway — coffeegoddd's review of #3780 was correct on that point.

## Scope

- `internal/storage/schema/schema.go`: +6 lines (overridable `stderr` var + one `fmt.Fprintf`)  
- `internal/storage/schema/schema_test.go`: new, 47 lines — verifies the line is emitted to stderr via the overridable var

Total: 53 lines added, 0 deleted.

Replaces #3780 (previously closed).

## Test plan

- [x] `go test ./internal/storage/schema/...` passes
- [x] `go build ./...` clean

PR #3780, opened 2026-05-07
schema: migration progress + bd list filter regression tests (be-ldr + be-acnquj)

quad341:
## What this changes

This PR bundles two independent be-ldr release items.

### 1. Schema migration progress + large-rig warning (be-8ja)

`bd init`, `bd migrate`, and any first run that opens a rig now emit a
per-migration progress line on **stderr** as schema migrations apply:

```
Applying migration 0033: add_date_indexes…
  done (1.4s)
```

On rigs whose `issues` table holds more than 10 000 rows, a one-shot
notice fires once before the migration loop:

```
Large rig detected (49187 issues). This migration may take up to 60 seconds; do not interrupt.
```

The migration loop itself is unchanged — same `splitStatements`
per-statement execution (GH#3363), same `INSERT IGNORE INTO
schema_migrations` accounting. Only the operator-visible stderr output
is new.

### 2. `bd list` filter regression tests + `--json` hint suppression (be-acnquj)

PM filed be-acnquj reporting that `bd list -l X`, `--label-any`,
`--label-pattern`, and `--title-contains` were silently ignored. Triage
showed the filter wiring (`BuildIssueFilterClauses`) was already correct
at HEAD — the original repro was a stale-binary artifact (against an
older installed `bd`). Two changes:

- **`cmd/bd/list_output.go`** — `printTruncationHint` now early-returns
  when `jsonOutput` is true, so the trailing
  `Showing N issues; more results matched...` pager line no longer
  appears under `--json`. (Stderr-only, but tools that inspect both
  streams now see clean output.) This is AC#7 of the bug.
- **`cmd/bd/list_test.go`** — `TestListLabelFiltersAcnquj` exercises
  AC#1-6 against `s.SearchIssues` (single label, AND composition,
  `--label-any` OR, `--label-pattern` glob, `--title-contains`
  case-insensitive, nonexistent-label empty). `TestPrintTruncationHintJSONSuppression`
  covers the new AC#7 guard with a 6-case truth table.

## Why one PR

The two pieces ship together because they're both be-ldr release work
already on the same branch. The first commit was reviewed and gated;
the be-acnquj fix was rebased on top and re-gated separately
(`release-gates/be-acnquj-gate.md`). Splitting them now would just
churn the existing CI.

## Why stderr (be-8ja)

Progress text goes to stderr deliberately. JSON consumers like
`bd list --json | jq` read stdout, and a stray `Applying migration…`
line on stdout would corrupt those pipelines. The default writer
[...]

PR #3919, opened 2026-05-12
schema: migration progress output + large-rig warning (be-ldr)

quad341:
## Maintainer edit (2026-07-23): slimmed to the pieces #3914 doesn't cover

`bd` now writes per-migration progress lines to stderr as it runs `MigrateUp`. Each line names the migration and reports how long it took:

```
Applying migration 0033: add_date_indexes…
  done (0.1s)
```

On databases with more than 10 000 issues, a one-shot warning fires before the first migration: "Large rig detected (N issues). This migration may take up to 90 seconds; do not interrupt."

Output goes to the same stderr writer #3914 already uses (TTY-gated — piped/non-terminal stderr stays silent, so `bd list --json | jq` etc. are unaffected). Fresh installs where the issues table does not yet exist suppress the warning — `COUNT(*)` errors are treated as "no issues, no warning needed."

**Why this PR was rebased and slimmed rather than merged as-is:** this PR's core progress-visibility feature (a package-level `progressOut` writer, printing a migration name to stderr) was superseded on `main` by #3914 — same author, same intent, landed independently with a different, TTY-gated mechanism (`stderr` var + `defaultStderr()`), plus a `preMigrationRepair`/`commitEachStep` evolution of `runMigrations` this branch predates. Rebasing this PR as-is would have reintroduced a second, parallel progress writer and reverted #3914's TTY gating and the newer `runMigrations` machinery.

What survives here is only what #3914 does **not** cover, verified absent from current `main` before doing this work:

- (a) the large-rig warning before slow migrations (`emitLargeRigNotice` / `issueRowCounter` / `largeRigThreshold`)
- (b) per-migration timing output (`done (N.Ns)`)
- (c) human-readable migration names (`humanMigrationName`, e.g. `add_date_indexes` instead of the raw filename)

These are layered onto #3914's existing `stderr` writer — no second writer, no change to `preMigrationRepair` threading or `commitEachStep`. Net diff: ~171 lines, mostly a new test file (`progress_test.go`), touching only `internal/storage/schema/schema.go` and its tests.

Quad341's original acceptance-criteria commits (`ce9d6e04`, `602ea7c4`) are preserved verbatim as the first commit; a second, separately-authored commit does the #3914-compatible wiring and updates the two existing `runMigrations` mock-DB tests that would otherwise panic on the new `issueRowCounter` call.

---

## Original PR description (quad341)

`bd` now writes per-migration progress lines to stderr as it runs `MigrateUp`. Each line names the migration and reports how long it took:

```
Applying migration 0033: add_date_indexes…
  done (0.1s)
```

On databases with more than 10 000 issues, a one-shot warning fires before the first migration: "Large rig detected (N issues). This migration may take up to 90 seconds; do not interrupt."

Output goes to stderr so it does not pollute stdout pipelines (`bd list --json | jq` etc.). Fresh installs where the issues table does not yet exist suppress the warning — `COUNT(*)` errors are treated as "no issues, no warning needed."

### Review notes

- `largeRigThreshold = 10000` — triggers at **strictly greater than** 10 000 (not at 10 000 exactly; test `at_threshold_no_warning` covers this)
- `issueRowCounter` var — stubbed in tests so schema tests don't need a live DB
- The 90-second estimate in the warning text is a conservative ceiling for typical large-rig schema runs; not enforced by code

### Test plan

- [x] `go build -tags gms_pure_go ./...` — clean
- [x] `go vet -tags gms_pure_go ./...` — clean
- [x] `go test -tags gms_pure_go -count=1 -run 'Migrat' ./internal/storage/schema/...` — PASS (7.0s)
- [x] `gofmt -l internal/storage/schema/` — clean

This is a rebase of PR #3780 (closed) onto current main after PR #3871 removed the `tolerateExisting` bool from `runMigrations`, then slimmed on 2026-07-23 after #3914 landed the progress-line subset independently.

### DIFF
--- a/internal/storage/schema/progress_test.go
+++ b/internal/storage/schema/progress_test.go
@@ -1,0 +1,107 @@
+package schema
+
+import (
+	"bytes"
+	"errors"
+	"strings"
+	"testing"
+)
+
+// TestEmitLargeRigNotice covers the be-8ja large-rig warning branch. The
+// fresh-install case (count reported via error, typically "table doesn't
+// exist") must stay silent — that's the UX contract: on a first-ever bd run,
+// there is no rig to warn about.
+func TestEmitLargeRigNotice(t *testing.T) {
+	cases := []struct {
+		name      string
+		count     int64
+		countErr  error
+		wantEmpty bool
+		wantSub   string
+	}{
+		{
+			name:      "fresh_install_table_missing",
+			count:     0,
+			countErr:  errors.New("Table 'issues' doesn't exist"),
+			wantEmpty: true,
+		},
+		{
+			name:      "small_rig_below_threshold",
+			count:     9_999,
+			countErr:  nil,
+			wantEmpty: true,
+		},
+		{
+			name:      "at_threshold_no_warning",
+			count:     10_000,
+			countErr:  nil,
+			wantEmpty: true,
+		},
+		{
+			name:    "one_past_threshold_warns",
+			count:   10_001,
+			wantSub: "Large rig detected (10001 issues)",
+		},
+		{
+			name:    "typical_large_rig",
+			count:   49_187,
+			wantSub: "Large rig detected (49187 issues)",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			var buf bytes.Buffer
+			emitLargeRigNotice(&buf, tc.count, tc.countErr)
+
+			got := buf.String()
+			if tc.wantEmpty {
+				if got != "" {
+					t.Errorf("want no output, got %q", got)
+				}
+				return
+			}
+			if !strings.Contains(got, tc.wantSub) {
+				t.Errorf("want substring %q; got %q", tc.wantSub, got)
+			}
+			if !strings.Contains(got, "do not interrupt") {
+				t.Errorf("warning missing operator guidance; got %q", got)
+			}
+			if !strings.HasSuffix(got, "\n") {
+				t.Errorf("warning must end with newline; got %q", got)
+			}
+		})
+	}
+}
+
+// TestHumanMigrationName pins the filename → display-name mapping for the
+// per-migration progress line. The be-8ja progress output MUST be stable and
+// human-readable; regression here would change operator-visible text.
+func TestHumanMigrationName(t *testing.T) {
+	cases := map[string]string{
+		"0033_add_date_indexes.up.sql":          "add_date_indexes",
+		"0001_initial.up.sql":                   "initial",
+		"0027_add_started_at.up.sql":            "add_started_at",
+		"noversion.up.sql":                      "noversion",
+		"0099_multi_word_migration_name.up.sql": "multi_word_migration_name",
+	}
+	for in, want := range cases {
+		if got := humanMigrationName(in); got != want {
+			t.Errorf("humanMigrationName(%q) = %q; want %q", in, got, want)
+		}
+	}
+}
+
+// TestProgressOutDefaultsToStderr guards the be-8ja invariant that the
+// package-level writer starts pointed at os.Stderr. A regression here would
+// silently leak migration progress into stdout and break bd <anything> --json
+// pipelines — exactly the failure mode the bead called out.
+func TestProgressOutDefaultsToStderr(t *testing.T) {
+	if progressOut == nil {
+		t.Fatal("progressOut must not be nil")
+	}
+	// Package-level default must be os.Stderr so stdout-sensitive consumers
+	// (bd list --json | jq, etc.) stay unpolluted. We don't pin the pointer
+	// identity here — tests swap progressOut via a helper — but a nil or
+	// stdout-bound default would be a regression worth catching.
+}
--- a/internal/storage/schema/progress_test.go
+++ b/internal/storage/schema/progress_test.go
@@ -92,16 +92,8 @@ func TestHumanMigrationName(t *testing.T) {
 	}
 }
 
-// TestProgressOutDefaultsToStderr guards the be-8ja invariant that the
-// package-level writer starts pointed at os.Stderr. A regression here would
-// silently leak migration progress into stdout and break bd <anything> --json
-// pipelines — exactly the failure mode the bead called out.
-func TestProgressOutDefaultsToStderr(t *testing.T) {
-	if progressOut == nil {
-		t.Fatal("progressOut must not be nil")
-	}
-	// Package-level default must be os.Stderr so stdout-sensitive consumers
-	// (bd list --json | jq, etc.) stay unpolluted. We don't pin the pointer
-	// identity here — tests swap progressOut via a helper — but a nil or
-	// stdout-bound default would be a regression worth catching.
-}
+// Note: be-8ja's original TestProgressOutDefaultsToStderr (pinning a
+// package-level progressOut default) doesn't apply after the #3919 slim —
+// this branch reuses main's existing `stderr` writer (see defaultStderr,
+// TTY-gated), landed independently by #3914. That invariant is exercised by
+// #3914's own tests, not duplicated here.
--- a/internal/storage/schema/schema.go
+++ b/internal/storage/schema/schema.go
@@ -35,6 +35,47 @@ func defaultStderr() io.Writer {
 	return io.Discard
 }
 
+// progressOut is where migration progress lines are written. Defaults to
+// os.Stderr so that JSON pipelines on stdout (e.g. bd list --json | jq) are
+// not polluted. Unexported so tests in this package can swap it without
+// leaking a setter into production API.
+var progressOut io.Writer = os.Stderr
+
+const largeRigThreshold = 10000
+
+// issueRowCounter returns the current issues-table row count, or an error if
+// the table is unreachable (fresh install → table doesn't exist yet). The
+// caller uses the error as the "no warning" signal. Variable so tests in this
+// package that exercise runMigrations against a non-DB mock can stub out the
+// query without panicking on QueryRowContext.
+var issueRowCounter = func(ctx context.Context, db DBConn) (int64, error) {
+	var n int64
+	err := db.QueryRowContext(ctx, "SELECT COUNT(*) FROM issues").Scan(&n)
+	return n, err
+}
+
+// emitLargeRigNotice writes the one-line large-rig warning to out when the
+// issues count exceeds largeRigThreshold. An error from the counter is
+// treated as "fresh install / table missing" and suppresses the warning —
+// see be-8ja for the UX rationale.
+func emitLargeRigNotice(out io.Writer, count int64, err error) {
+	if err != nil || count <= largeRigThreshold {
+		return
+	}
+	fmt.Fprintf(out, "Large rig detected (%d issues). This migration may take up to 90 seconds; do not interrupt.\n", count)
+}
+
+// humanMigrationName turns "0033_add_date_indexes.up.sql" into
+// "add_date_indexes" for the progress line.
+func humanMigrationName(filename string) string {
+	s := strings.TrimSuffix(filename, ".up.sql")
+	parts := strings.SplitN(s, "_", 2)
+	if len(parts) < 2 {
+		return s
+	}
+	return parts[1]
+}
+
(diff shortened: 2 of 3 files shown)


quad341:
Triaged the broad CI failures and pushed `de2881e57` to fix the most damaging one.

## Root cause: `runMigrations` ignores its caller's `migrationSource`

This branch refactored `(m migrationSource).migrate` to delegate to a new free function `runMigrations`, but `runMigrations` hardcodes `mainSource` for `list()`, `files`, `dir`, and `cursorTable` instead of using the passed-in source. The two production callers of `migrate` are `mainSource.migrate` (line 137) **and `ignoredSource.migrate` (line 151)** — so the ignored migrations under `internal/storage/schema/migrations/ignored/` never ran. `ignored_schema_migrations` stayed empty, and any table the ignored migrations were supposed to create (e.g. `wisp_dependencies.depends_on_wisp_id` from `0003_split_wisp_dependencies_target.up.sql`) was missing.

Fix: thread `src migrationSource` through `runMigrations` and use it instead of `mainSource`. `TestRunMigrationsProgressOutput` now passes `mainSource` explicitly (same behavior as before). Verified the schema unit test still passes locally with `go test -tags gms_pure_go ./internal/storage/schema/`.

## What this fix should clear on next CI run

Direct cascades from the missing ignored migrations:
- `Test (Embedded Dolt Storage)` → `TestSchemaAfterInit` (no rows in `ignored_schema_migrations`)
- `Test (Embedded Dolt Cmd 18/20)` → `TestEmbeddedSquashWispMoleculeClearsRootEphemeral` (`Unknown column 'depends_on_wisp_id'`)
- Possibly `Test (Embedded Dolt Cmd 15/20)` → "column `depends_on_issue_id` could not be found" (similar cascade)
- Possibly `Test (Embedded Dolt Cmd 16/20)` → migration 0041 timeout (state inconsistency from missing ignored tables)

## What this fix does NOT address — pre-existing test-helper issue

The remaining failures show JSON parse errors like `invalid character 'A' looking for beginning of value` (TestEmbeddedQuery, TestEmbeddedVC, TestEmbeddedReady, TestEmbeddedBranch, TestEmbeddedExport, TestEmbeddedSearch, TestEmbeddedOrphans). The `'A'` is the first character of `Applying migration…` — this PR's new stderr progress output bleeding into stdout. Same shape as the known `bd-embedded-test-json-stderr-leak` flake.

The PR's design is correct here: progress writes go to `os.Stderr` so `bd … --json | jq` pipelines on stdout stay clean. The fix is in the test helpers (`cmd.CombinedOutput()` → split `Stdout`/`Stderr`), which **PR #3996** is doing across the remaining `*_embedded_test.go` files. Once #3996 merges and this branch rebases on it, those failures should drop.

Suggest holding this PR for #3996 before re-running CI on the remaining (non-cascade) failures; once #3996 is in, a rebase + re-run should be the final check.

quad341:
## beads/reviewer — PASS ✓

Migration progress tests are thorough and cover the right edge cases.

### Findings

None.

- **Test design**: 5 cases covering fresh-install silence (error path), below-threshold silence, at-threshold boundary, and large-rig warning. The fresh-install case (table doesn't exist → error → silent) is exactly the right UX contract to test. ✓
- **Threshold boundary**: `> 10_000` (not `>=`) means 10001 triggers the warning. Test case `at_threshold_no_warning` verifies the exclusive boundary explicitly. ✓

Ready to merge.

🤖 beads/reviewer

maphew:
**Triage: request-changes (needs rebase + overlap resolution with #3914; conflict carries a real correctness trap).**

Thanks @quad341 for this - visible migration progress plus a large-rig warning is genuinely useful UX for anyone sitting through a multi-minute `MigrateUp` on a big rig, and the test coverage here (fresh-install silence, threshold boundary, timing format) is solid.

## Findings

- **[blocker] Refactor drops `preMigrationRepair` if conflict-resolved naively.** Current `upstream/main`'s `migrationSource.migrate` calls `m.preMigrationRepair(ctx, db, mf.version)` inside the per-migration loop (added after this PR's merge-base, alongside `migration_repairs.go`). This PR's `runMigrations` extraction rewrites that same loop body without that call. GitHub reports `CONFLICTING`, so a naive "take theirs" resolution on `schema.go` would silently drop pre-migration repair - a correctness regression, not just a merge nit. Whoever rebases this needs to explicitly re-thread `preMigrationRepair` through `runMigrations`.
- **[major] Overlaps PR #3914, same author, same area, opened the same day.** #3914 is the earlier, minimal version (single progress line, TTY-aware `stderr` var using `golang.org/x/term`, discards to `io.Discard` when not a terminal). #3919 is a superset feature-wise (large-rig warning, per-migration timing, human-readable names) but drops the TTY-detection idea in favor of always-stderr. Both are `CONFLICTING`/`DIRTY` and stale since 2026-06-15. One of these should be closed in favor of the other - recommend closing #3914 and, if the TTY-aware `io.Discard` fallback still has merit (e.g. avoiding noise in log-redirected/non-interactive runs), folding that idea into #3919 with credit to that draft.
- **[major] Real divergence, not a mechanical rebase.** Since the PR's merge-base (`8e18581d`, 2026-06-15), `internal/storage/schema/schema.go` alone gained ~150 lines across `migration_repairs.go`, `remote_migrate_gate.go`/`smart_remote_migrate_gate.go` expansions, a content-hash-probe replacement (`SHOW COLUMNS`, #4479), and `schema_skew_test.go`. `schema_test.go` diverged by 665 lines. This is a substantive rebase requiring understanding of the intervening repair/gate logic, not a whitespace conflict.
- **[minor]** Codecov flags 2 uncovered lines (92.3% patch coverage) in `schema.go` - likely an error-path branch; worth a quick look during rebase.
- **[nit]** `humanMigrationName` silently returns the full trimmed filename when there's no `_` separator (e.g. a hypothetical single-word migration name) - fine as a fallback, just note it's untested.

## Design fit
No storage-boundary concern: this only adds a read-only `COUNT(*)` for the warning and stderr writes, no new locks/retries against storage failure modes. Aligned with current architecture.

## Rebase-conflict scope
`mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`. Merge-base `8e18581d` (2026-06-15) vs current `upstream/main` tip (2026-07-04, ~3 weeks / many commits apart). Conflicting files: `internal/storage/schema/schema.go` (core - the `migrate`/`runMigrations` loop, see blocker above) and `internal/storage/schema/schema_test.go` (main added 665 lines of new tests since merge-base). New files added by the PR (`progress_test.go`, `release-gates/be-ldr-gate.md`) apply cleanly. Not mechanical - needs a human/agent to re-verify `preMigrationRepair` and any other intervening logic is preserved.

## Recommendation
Ask @quad341 (or whoever picks this up) to: (1) decide #3914 vs #3919 and close the redundant one, (2) rebase #3919 onto current `upstream/main`, explicitly re-threading `preMigrationRepair` through the new `runMigrations` function, and (3) re-run the schema test suite post-rebase before requesting re-review.

maphew:
Rebased onto current `upstream/main` and slimmed the diff.

Verified before touching anything: #3914 (same author) landed independently on `main` with a different mechanism — a TTY-gated package-level `stderr` var (`defaultStderr()`, `golang.org/x/term`) — and `runMigrations` on `main` now also threads `preMigrationRepair` and a `commitEachStep` atomic-commit path this branch predates (`main` added `preMigrationRepair` for migration 0053 after this branch forked). Rebasing this PR's original diff as-is would have reintroduced a second, parallel `progressOut` writer, reverted the TTY gating, and dropped the newer `runMigrations` machinery.

Grepped `main` for the three pieces this PR claims as unique and confirmed all three are genuinely absent:

- `largeRig` / `LargeRig` / `humanMigrationName` / `time.Since` around the migration loop: no hits on `main`'s `internal/storage/schema/schema.go`.
- The only progress line on `main` is a single `fmt.Fprintf` of `"migrating schema: %s"` plus filename — filename as-is, no timing, no threshold warning.

So this is not a close-candidate; it's a genuine slim-and-rebase. Dropped everything #3914 already covers (the progress-line-exists mechanism, the `preMigrationRepair`/source-threading fix, TTY gating) and kept only:

- (a) the large-rig warning (`emitLargeRigNotice`, `issueRowCounter`, `largeRigThreshold`)
- (b) per-migration timing (`done (N.Ns)`)
- (c) human-readable migration names (`humanMigrationName`)

layered onto #3914's existing `stderr` writer instead of introducing a second one. Net diff vs. `upstream/main`: 3 files, ~171 lines, mostly a new test file. Two commits: the first carries quad341's original `ce9d6e04`/`602ea7c4` helper + test code verbatim (unwired); the second (mine) adapts the wiring onto `main`'s `stderr` mechanism and fixes the two existing mock-DB `runMigrations` tests that would otherwise panic on the new `issueRowCounter` call.

Preflight, all green on the slimmed branch:
- `go build -tags gms_pure_go ./...` — clean
- `go vet -tags gms_pure_go ./...` — clean
- `go test -tags gms_pure_go -count=1 -run 'Migrat' ./internal/storage/schema/...` — PASS (7.0s)
- `gofmt -l internal/storage/schema/` — clean

(The wider `-run 'Migrat' ./internal/storage/...` sweep also touches `internal/storage/dolt`, which has 7 pre-existing failures in this sandbox unrelated to this change — reproduced identically on a clean `upstream/main` checkout with no changes applied, so not a regression from this PR.)

PR body updated to record this maintainer edit and scope change, with quad341's original description preserved below it.

maphew:
Round 2, addressing cross-vendor review (opus approve-with-nits, codex two P2s):

1. **Duplicate large-rig warning.** `emitLargeRigNotice` fired unconditionally at the top of `runMigrations`, which `MigrateUp` calls once for `mainSource` and once for `ignoredSource` — a large rig with pending migrations in both sources got the "one-shot" warning (and the `COUNT(*)` query) twice. Fixed by gating on `src.cursorTable == mainSource.cursorTable`, the same pattern `migrate()` already uses for `commitEachStep`. Added `TestRunMigrationsLargeRigNoticeOnlyOnMainSource` to pin single emission across both source calls.

2. **`done (N.Ns)` printed before the per-step Dolt commit.** On the production embedded path, `commitMigrationStep` (`DOLT_ADD`/`DOLT_COMMIT`) is the expensive, fallible part of a step, but it ran after the done line — an operator could see "done" then hit a commit error, and the printed timing excluded the commit's cost. Moved the done print and its `time.Since` to after the `commitEachStep` block, so it lands last on both paths, and a failed step commit returns before either print statement runs.

`TestRunMigrationsStderrOutput`/`TestRunMigrationsUsesProvidedSource` both use `commitEachStep=false`, so the reordering is a no-op for them — confirmed by the full suite.

Preflight on the updated head: `go build`/`go vet` clean, full `internal/storage/schema` suite PASS (9.5s, includes the new test plus the existing `TestEmitLargeRigNotice`/`TestHumanMigrationName`), `gofmt` clean.

maphew (approved):
## Cross-vendor paired review, adjudicated: approved

Round 1 (claude-opus + codex-gpt-5.6-sol on the slimmed PR): the safety-critical properties held from the start — every new code path is advisory-only (no error can reach the migration error flow, verified against all early returns), the row counter runs before the loop and outside any sensitive span, output is TTY-gated to stderr with nothing on stdout, `humanMigrationName` is panic-safe, and the mockDB test changes tighten rather than weaken assertions. Two convergent fixes were requested: the large-rig notice double-fired (once per migration source), and "done (N.Ns)" printed before the per-step Dolt commit — a misleading success line ahead of the expensive/fallible part.

Round 2 (48e05f2c8, claude-opus delta): CLEAN-PASS. The notice is now gated to the main-source pass using the same idiom as the existing `commitEachStep` gate (stable against source reordering; the COUNT query is also skipped on the ignored pass), the done-print sits after the step commit with the failure path returning before it, the `commitEachStep=false` path's output order is unchanged, and the new `TestRunMigrationsLargeRigNoticeOnlyOnMainSource` is gate-load-bearing (dropping the gate fails it).

Nice slim-down overall: what remains is exactly the three pieces main didn't get from the merged sibling work — the large-rig warning, per-migration timing, and human-readable names — layered on main's own TTY-gated writer instead of a parallel one. Merging once CI completes green. Thanks @quad341.

_claude-code-fable-5 adjudicating claude-opus-4-8 + codex-gpt-5.6-sol-high, on behalf of matt wilkie_

PR #3920, opened 2026-05-12
cmd: add --mem-profile flag and BEADS_MEM_PROFILE env var (be-tg2o)

quad341:
## Summary

- Adds `--mem-profile FILE` flag (mirrors existing `--profile` CPU flag)
- `BEADS_MEM_PROFILE=FILE` env var fallback for scripts/CI
- `BEADS_MEM_PROFILE_NOGC=1` skips pre-write GC to capture peak allocation
- `BEADS_MEM_STATS=FILE` writes a one-line HeapAlloc/HeapSys/HeapInuse/HeapObjects summary
- Documents all four knobs in `docs/CONFIG.md §Profiling`

## Context

The investigator root-caused be-81n using a hand-patched heap profile (uncommitted). This promotes that pattern to a proper flag so future RAM debugging doesn't require re-instrumenting the binary.

## Implementation

Follows the existing `--profile`/`pprof.StartCPUProfile` pattern in `cmd/bd/main.go`:
- `memProfilePath string` global holds the flag value
- `BEADS_MEM_PROFILE` env var checked as fallback in PersistentPostRun
- `#nosec G304` annotations (user-supplied debug paths, same category as batch file paths)

## Gates

- `go build ./...` — clean
- `go vet ./cmd/bd/...` — clean
- `gofmt -l cmd/bd/main.go` — clean
- Smoke test: `BEADS_MEM_PROFILE=/tmp/h.prof bd version && go tool pprof -top /tmp/h.prof` — valid pprof heap output

Closes be-tg2o


quad341:
Fixed the one failing check ("Check doc flags freshness"):

- The new `--mem-profile` global flag wasn't reflected in `docs/CLI_REFERENCE.md`; CI's freshness check (`scripts/check-doc-flags.sh` → `scripts/generate-cli-docs.sh --check`) flagged the diff.
- Regenerated via `./scripts/generate-cli-docs.sh ./bd` — produces exactly the one-line addition CI expected (commit 6d0616048). Website CLI reference pages were already in sync.
- Verified locally: full `./scripts/check-doc-flags.sh ./bd` now exits PASS. The Check 3 "legacy storage references" output remains a WARN (not a FAIL) and is unrelated/pre-existing across the docs tree.

Ready for another pass.

quad341:
## beads/reviewer — PASS ✓

Clean developer profiling flag addition.

### Findings

None.

- **`os.Create(heapDest)` / `os.Create(statsDest)`**: `#nosec G304` comments are accurate — these are user-supplied developer profiling paths, not an injection risk. ✓
- **Env var fallback**: `BEADS_MEM_PROFILE` → `--mem-profile` precedence is correct. ✓
- **GC semantics**: Pre-GC before heap dump is the right default; `BEADS_MEM_PROFILE_NOGC=1` to capture peak is a correct expert escape hatch. ✓

Ready to merge.

🤖 beads/reviewer

maphew:

quad341:
Thanks for sweeping this! I think #3920 got picked up as a bug report — it's
actually a small **developer-tooling** PR rather than a fix, so there isn't a
repro in the usual sense. It *adds* a heap-profiling capability (the sibling of
the existing `--profile` CPU flag); the "expected output" is the profile
artifact the new flag writes, not a corrected behavior.

**What it adds** (all opt-in — no default behavior change):

- `--mem-profile FILE` / `BEADS_MEM_PROFILE=FILE` — write a heap profile on exit
- `BEADS_MEM_PROFILE_NOGC=1` — skip the pre-write `runtime.GC()` to capture peak allocation
- `BEADS_MEM_STATS=FILE` — one-line `HeapAlloc/HeapSys/HeapInuse/HeapObjects` summary

**Why:** an earlier RAM investigation had to hand-patch a heap dump into the
binary (uncommitted); this promotes that into a first-class flag so the next
memory investigation doesn't require re-instrumenting `bd`.

**Smallest sequence to exercise it** — storage-agnostic (the dump runs in
`PersistentPostRun`, so any subcommand works; `version` needs no DB). Captured
just now from a fresh build of the PR head:

```console
$ bd version
bd version 1.0.4

$ bd version --mem-profile /tmp/h.prof
$ go tool pprof -top -nodecount=6 /tmp/h.prof
Type: inuse_space
Showing nodes accounting for 7774.80kB, 88.36% of 8798.82kB total
      flat  flat%   sum%        cum   cum%
 4108.76kB 46.70% 46.70%  4108.76kB 46.70%  runtime.mallocgc
 1577.69kB 17.93% 64.63%  1577.69kB 17.93%  regexp/syntax.(*compiler).inst
  532.26kB  6.05% 70.68%   532.26kB  6.05%  github.com/yuin/goldmark/util.init.0
  ...

# env-var form, for scripts/CI
$ BEADS_MEM_STATS=/tmp/stats.txt bd version && cat /tmp/stats.txt
HeapAlloc=4256848 HeapSys=15269888 HeapInuse=6897664 HeapObjects=16604
```

Environment for the capture above:

- `bd version`: **1.0.4**
- Storage mode: **backend-agnostic** (the profiler is independent of storage; captured here with embedded Dolt)
- OS/arch: **linux/amd64** (built `CGO_ENABLED=0 -tags gms_pure_go`)

**On the merge conflict:** the code itself (`cmd/bd/main.go`) still rebases onto
`main` cleanly — I verified it auto-merges with no conflict markers, rebuilds,
and the flag still produces valid profiles. The only real drift is docs: the
Profiling section was written against `docs/CONFIG.md`, which was removed by the
docs→site migration (5c2fe0d49), so that section needs re-homing into the new
site (`docs/reference/configuration.md`) with `docs/CLI_REFERENCE.md`
regenerated.

Happy to rebase and re-home the docs if you'd like to take it — just say the
word and I'll push.

maphew:
Rebased onto current upstream/main (was 44 commits behind; now on top of `4f0df4e6a`).

The only conflict was in `docs/CONFIG.md`, which upstream deleted as part of a docs
reorg (content merged into `docs/reference/configuration.md`, see 5c2fe0d49). That
file no longer carries a narrative Profiling section for either `--profile` or the
new `--mem-profile` flag, so I dropped this PR's `docs/CONFIG.md` hunk rather than
reintroduce a file upstream removed. `docs/CLI_REFERENCE.md` (auto-generated) still
documents `--mem-profile` correctly after the regen commit.

`cmd/bd/main.go` itself merged cleanly - the flag-registration and
`PersistentPostRun` seam this PR extends is unchanged in shape.

Verified in a clean worktree:
- `go build -tags gms_pure_go ./cmd/bd/...`
- `go vet -tags gms_pure_go ./cmd/bd/` (CGO_ENABLED=1; without the build tag this
  environment fails on a missing system ICU header, unrelated to this PR)
- `go test -tags gms_pure_go -count=1 -run 'Profile' ./cmd/bd/.` - passes, no test
  currently matches that name pattern

Force-pushed the rebase to this PR's branch after confirming the remote tip
hadn't moved.

Separately: the unanswered 2026-06-16 comment on this thread asking for a repro
recipe appears to have been intended for a different PR - this PR only adds a
profiling flag, there's no bug to repro. The flag's behavior is demonstrated by
`bd --mem-profile <file> <cmd>` writing a heap profile to `<file>`.

maphew (approved):
## Review (claude-opus, adjudicated): approve with non-blocking notes

Verified after the rebase onto current `main`:

- Heap profile writes at the right lifecycle point (`PersistentPostRunE`, after store close/telemetry/CPU-profile stop), calls `runtime.GC()` before `pprof.WriteHeapProfile` (matching `go test -memprofile` convention) with `BEADS_MEM_PROFILE_NOGC=1` to skip for peak capture; file handle closed.
- No conflict with `--profile` (distinct files, both can be set); early `os.Exit` paths skip both profiles equally — pre-existing parity, not introduced here.
- Flag wins over `BEADS_MEM_PROFILE` env fallback, as advertised.
- `docs/CLI_REFERENCE.md:306` addition is byte-identical to the built binary's rendered help, correctly placed.
- Dropping the `docs/CONFIG.md` hunk in the rebase was right (file removed upstream in the docs reorg).

Non-blocking notes for a possible follow-up (not worth another CI round here):
1. `BEADS_MEM_STATS` and `BEADS_MEM_PROFILE_NOGC` are new behavioral surface documented nowhere — worth a line in an env-var reference.
2. `--mem-profile /bad/path` fails silently (`os.Create` error swallowed) — same parity as `--profile`, but an explicit user-supplied path deserves a stderr warning.

Will merge once CI completes green.

_claude-code-fable-5 adjudicating claude-opus-4-8, on behalf of matt wilkie_

maphew:
Follow-up on CI: pushed `0e1728d84`, removing the `--mem-profile` line from `docs/CLI_REFERENCE.md`. "Check doc flags freshness" builds `bd` from the release pin in `docs/cli-docs.pin` (currently `v1.1.0`) rather than this branch's HEAD, and `v1.1.0` predates `--mem-profile`, so the pinned generator's own regen omits that line (its uploaded fix patch was exactly `docs/CLI_REFERENCE.md | 1 deletion(-)`, which is what I applied) - the line will reappear automatically once the pin advances past a release containing this flag. "Build Artifacts" and "PR Policy (wrapper timing)" both failed for the identical reason: both run `make ci-pr-policy`, whose "Run policy checks"/"Run PR policy wrapper" step invokes the same doc-flags script, so neither is an independent artifact-build or policy-wrapper bug in this PR's diff - confirmed by job logs (`Run policy checks` and `Run PR policy wrapper` both fail with the same `FAIL: docs/CLI_REFERENCE.md is out of sync` output before `make: *** [Makefile:124: ci-pr-policy] Error 1`). "Test Nix Flake" remains separately-failing infra noise unrelated to this PR: it fails on a Nix flake hash mismatch for `beads-1.1.0-go-modules` before any repo code runs.

PR #3929, opened 2026-05-13
fix(github) bd github sync (GitHub issues integration)

sealocal:
Fixed: `bd github sync now` reads `github.token` from `.bead/config.yaml` correctly, by borrowing the pattern already set in the Linear integration.

## What

internal/github/tracker.go:Init called t.getConfig(...), which only checked t.store.GetConfig (Dolt DB) and the env var

github.token is a yaml-only secret key (per internal/config/yaml_config.go:73), so it's never in the DB. The yaml lookup was missing.

## How

Mirror the pattern already used by internal/linear/tracker.go and cmd/bd/github.go:getGitHubConfigValue(): check config.IsYamlOnlyKey first and read from config.yaml for those keys.

Also adds TestGetConfig_YamlOnlyKeyBypassesStore: constructs a Tracker with a nil store and verifies getConfig still resolves github.token via env-var fallback — proof that yaml-only keys no longer deref the store. Without the fix this test would panic on the nil store.

## Why

```sh
bd github sync
Error: initializing GitHub tracker: GitHub token not configured (set github.token or GITHUB_TOKEN)
```

The github tracker's getConfig() only checked store.GetConfig (Dolt DB) and the env var, but github.token is a yaml-only secret key and is never stored in the DB. As a result, `bd github sync` errored with "GitHub token not configured" even when the token was correctly set in .beads/config.yaml via `bd config set github.token <token>`.

## Verification

```
bd config set github.user your-username
bd config set github.repo your-repo
bd config set github.token "github_pat_1234" # your GH PAT
```

## TODO

* check if the same bug exists in the jira, gitlab, ado, and notion integration/trackers

### DIFF
--- a/internal/github/tracker.go
+++ b/internal/github/tracker.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/tracker"
 	"github.com/steveyegge/beads/internal/types"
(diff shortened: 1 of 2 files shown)


harry-miller-trimble (approved):
Clean fix, solid tests — nice work @sealocal.

The pattern mirrors what Linear's tracker already does (`internal/linear/tracker.go:614`), the two tests cover both the yaml-read and env-var-fallback paths well, and the nil-store trick is a smart way to prove the store bypass without needing a full integration harness.

Good catch on the TODO about jira/gitlab/ado/notion having the same bug — confirmed all four have the identical missing `IsYamlOnlyKey` check. We'll file follow-ups for those.

Thanks for the contribution!

PR #758, opened 2025-12-27
Fix: Migration 028 column mismatch with migration 029

maphew:
Fixes database migration error when running on a fresh database.

## Problem
Migration 029 adds the `created_by` column, but migration 028 uses `SELECT *` to copy data. When 028 runs, it copies all columns including created_by (from the existing schema), then tries to insert them into the new table which only has 34 columns.

Error: `sqlite3: SQL logic error: table issues_new has 34 columns but 35 values were supplied`

## Solution
Explicitly list the 34 columns in migration 028 to avoid the mismatch. This ensures we only copy columns that should exist at that point in the migration sequence.

PR #3939, opened 2026-05-14
fix(schema): use dynamic column intersection in migration 0035

### LINKED ISSUE #3938: Migration 0035 fails with Error 1105 on upgraded DBs (suspected regression of #2168)
hey, i upgraded Dolt to v2.0.1 - bd worked fine, then today i pulled latest beads upstream (`b4f5d2aa8`) and reinstalled (`make install`) - reads still work (`bd list` is fine), but every write fails on first contact with `failed to initialize schema: ... Error 1105`. so reads are OK, writes are blocked.

filing now so it's tracked; repro and what i've found so far are below. flagging that this is a "suspected" framing - i haven't branched and tried a fix yet, so something could shift once i actually dig in. happy to put up a PR if the diagnosis holds.

## What I see

```
Error: failed to open database: failed to initialize schema: schema migration: migration 0035_migrate_infra_to_wisps.up.sql: Error 1105 (HY000): number of values does not match number of columns provided
```

Reproduces on any `bd` write (e.g. `bd create`) after upgrading. The schema cursor on my DB stops at version `34` and won't advance. `bd list` and other read-only paths still work.

## Suspected root cause

Line 4 of `internal/storage/schema/migrations/0035_migrate_infra_to_wisps.up.sql`:

```sql
INSERT IGNORE INTO wisps SELECT * FROM issues WHERE issue_type IN (...)
```

`SELECT *` doesn't account for column drift between `issues` and `wisps`. On my DB:

| table | columns |
|---|---|
| `issues` | 49 |
| `wisps` | 55 |

(direction can flip on different upgrade paths - see #2168 for an earlier instance with the inverse drift.) Dolt v2's stricter MySQL semantics reject the column-count mismatch with Error 1105. The down migration appears to have the symmetric pattern.

This looks like the **same bug class** PR #2168 ("fix: make migration 007 resilient to schema evolution differences") fixed in the predecessor Go migration `007_infra_to_wisps.go` (commit `4891d870f`, Feb 27 2026). When 007 was rewritten as SQL migration 0035, the `SELECT *` antipattern looks like it came back along for the ride.

PR #3918 (merged 2026-05-13) added `IF table exists` guards to 0035 and introduced `migrations/ignored/0001_create_local_state_tables.up.sql` to recreate working-set tables - but `ignored/0001` creates `__temp__*` working tables, not the real `wisps`. Upgrade DBs are never reconciled and remain stuck at 0034.

## Suspected fix shape

Same as #2168: replace `SELECT *` in 0035 up/down with `INFORMATION_SCHEMA.COLUMNS`-driven dynamic column intersection + `PREPARE`/`EXECUTE`. Two precedents already in tree:

1. PR #2168 fix for the Go predecessor (`4891d870f`)
2. Migrations `0023`, `0027`, `0033`, `0034`, `0037`, `0038` already use `INFORMATION_SCHEMA.COLUMNS` + `PREPARE` for state-aware migrations

Schema reconciliation (ALTERing `wisps` to add the missing columns) is probably the wrong shape - migrations 0033/0034 added those columns to `issues` only, which seems intentional. Dynamic intersection preserves that intent.

That said, i haven't actually branched and tried it yet - there may be wrinkles around the new `dolt_nonlocal_tables` machinery from #3918 that this assumption doesn't survive. Will share whatever i find when i dig in.

## Possible test gap

There doesn't appear to be a regression test running the full `0001 -> head` migration chain against a DB seeded with a representative upgraded-DB starting state. PRs #758 (migration 028/029 column mismatch), #2168 (this same bug for migration 007), and this case make three of the same class - a chain-replay test would catch the next one.

## Repro

Any DB created before migrations 0033 and 0034 landed, then upgraded to a `bd` binary containing 0035. Fresh `bd init` is unaffected.

## Environment

- `bd version 1.0.4 (b4f5d2aa8)` - upstream/main as of 2026-05-14
- Dolt v2.0.1
- Windows 11
- shared-server mode

## Sources

- Migration source: `internal/storage/schema/migrations/0035_migrate_infra_to_wisps.up.sql` (line 4)
- Prior-art fix: PR #2168, commit `4891d870f`
- Recent overhaul that didn't catch the upgrade path: PR #3918 (`Db/ignore migration`), commit `b4f5d2aa8`
- Earlier related: PR #758 (migration 028/029 column mismatch), PR #1372 (added `spec_id` to `issues`)
- Asymmetric column adds: migrations `0033_add_wisp_type_column.up.sql`, `0034_add_spec_id_column.up.sql`
- Existing INFORMATION_SCHEMA precedent: migrations `0023`, `0027`, `0033`, `0034`, `0037`, `0038`

seanmartinsmith:
## What

rewrites the two SELECT * lines in migration 0035 (up and down) to use INFORMATION_SCHEMA-driven dynamic column intersection via GROUP_CONCAT + PREPARE/EXECUTE.

## Why

Fixes #3938.

migration 0035 (move infra-type rows from issues to wisps) fails with Error 1105 on databases upgraded across the asymmetric column additions in 0033 (wisp_type) and 0034 (spec_id) - SELECT * pulls a different column count than the target accepts. Dolt v2's stricter MySQL semantics reject the mismatch.

same bug class was fixed for the Go predecessor migration 007 in commit 4891d870f (PR #2168) and returned when 007 was rewritten as SQL. down has the symmetric pattern, fixed the same way.

approach mirrors the existing INFORMATION_SCHEMA + PREPARE pattern already used in migrations 0023, 0027, 0033, 0034, 0037, 0038 - so this isn't introducing a new shape, just applying the existing convention where it was missing.

## Verification

reproduced and fixed against an upgraded DB locally:

- before: migration cursor stuck at 34, every `bd` write fails with `failed to initialize schema: ... Error 1105 (HY000): number of values does not match number of columns provided`
- after: chain advances cleanly through 0040, `bd create` succeeds, `bd list` continues to work

column counts on my DB confirmed asymmetric (issues=49, wisps=55). fix moves rows via the intersection of common columns; doesn't touch the table schemas themselves.

scope is intentionally narrow - just the two broken statements. flagged in #3938 that a chain-replay regression test would catch the next instance of this class (third occurrence after #758, #2168), but that's separate work with its own design decisions about test infra. happy to take that on as a follow-up if it'd be useful.

### DIFF
--- a/internal/storage/schema/migrations/0035_migrate_infra_to_wisps.down.sql
+++ b/internal/storage/schema/migrations/0035_migrate_infra_to_wisps.down.sql
@@ -1,5 +1,26 @@
-INSERT IGNORE INTO issues SELECT * FROM wisps
-WHERE issue_type IN ('agent', 'rig', 'role', 'message');
+-- Move infra-type rows from wisps back into issues using the intersection
+-- of columns common to both tables. See up migration for rationale.
+SET SESSION group_concat_max_len = 32768;
+SET @cols = (
+    SELECT GROUP_CONCAT(COLUMN_NAME ORDER BY ORDINAL_POSITION SEPARATOR ',')
+    FROM INFORMATION_SCHEMA.COLUMNS
+    WHERE TABLE_SCHEMA = DATABASE()
+      AND TABLE_NAME = 'issues'
+      AND COLUMN_NAME IN (
+          SELECT COLUMN_NAME FROM INFORMATION_SCHEMA.COLUMNS
+          WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'wisps'
+      )
+);
+SET @sql = IF(
+    @cols IS NOT NULL AND @cols <> '',
+    CONCAT(
+        'INSERT IGNORE INTO issues (', @cols, ') ',
+        'SELECT ', @cols, ' FROM wisps ',
+        'WHERE issue_type IN (''agent'', ''rig'', ''role'', ''message'')'
+    ),
+    'SELECT 1'
+);
+PREPARE stmt FROM @sql; EXECUTE stmt; DEALLOCATE PREPARE stmt;
 
 UPDATE issues SET ephemeral = 0
 WHERE issue_type IN ('agent', 'rig', 'role', 'message');
--- a/internal/storage/schema/migrations/0035_migrate_infra_to_wisps.up.sql
+++ b/internal/storage/schema/migrations/0035_migrate_infra_to_wisps.up.sql
@@ -1,7 +1,26 @@
+-- Move infra-type rows from issues into wisps using the intersection of
+-- columns common to both tables. Using SELECT * crashes on upgraded DBs
+-- where the two tables have drifted (e.g. 0033 added wisp_type and 0034
+-- added spec_id to issues only). Mirrors the fix in #2168 / commit
+-- 4891d870f for the Go predecessor migration 007.
+SET SESSION group_concat_max_len = 32768;
+SET @cols = (
+    SELECT GROUP_CONCAT(COLUMN_NAME ORDER BY ORDINAL_POSITION SEPARATOR ',')
+    FROM INFORMATION_SCHEMA.COLUMNS
+    WHERE TABLE_SCHEMA = DATABASE()
+      AND TABLE_NAME = 'wisps'
+      AND COLUMN_NAME IN (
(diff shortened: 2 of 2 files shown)


coffeegoddd:
we are not gonna merge this, we are actively working on the schema and storage layer

seanmartinsmith:
deferring to #3942 - same fix shape from coffeegoddd; appreciate it landing.

Issue #3356, opened 2026-04-19
bd init registers git origin as Dolt remote even when remote has no Dolt data


kevglynn:
## Summary

`bd init` unconditionally registers the parent git repo's `origin` URL as a Dolt remote, even when the remote has no `refs/dolt/data`. This causes every subsequent Dolt fetch operation to fail (the remote is a plain Git source repo, not a Dolt database), generating orphaned `tmp_pack_*` files that accumulate unboundedly (see #3354).

## Root Cause

In `cmd/bd/init.go` (~L493-639), the logic is:

1. If `sync.remote` config is empty AND we're in a git repo with an `origin` remote, set `syncURL` to the normalized origin URL
2. Set `syncFromRemote = true` ONLY if `git ls-remote origin refs/dolt/data` succeeds
3. Clone from remote ONLY if `syncFromRemote` is true (correct)
4. **Add the Dolt remote** whenever `syncURL != ""` — **regardless of whether `syncFromRemote` is true** (bug)
5. **Persist `sync.remote`** to `config.yaml` whenever `syncURL != ""` — **regardless of whether the remote has Dolt data** (bug)

The `AddRemote` and `sync.remote` persistence at steps 4-5 should be gated on `syncFromRemote` (or on explicit user config), not just on `syncURL != ""`.

## Impact

When `bd init` is run in a repo like `gastownhall/beads` itself (a Go source repo, not a Dolt database):

1. `origin` = `https://github.com/gastownhall/beads.git` gets registered as a Dolt remote (`git+https://...`)
2. Every Dolt operation that touches the remote triggers `git fetch` against the bare cache repo
3. Every fetch fails (no Dolt refs exist on the remote)
4. Each failed fetch leaves a 100-900 MB `tmp_pack_*` file in `.dolt/git-remote-cache/`
5. Files accumulate at ~15 GB/day, filling the disk (#3354)

## Contrast with `bd bootstrap`

`detectBootstrapAction` in `cmd/bd/bootstrap.go` (~L286-297) correctly gates on `refs/dolt/data`:

```go
if gitLsRemoteHasRef("origin", "refs/dolt/data") {
    plan.SyncRemote = normalizeRemoteURL(originURL)
    // ...
}
```

`bd init` should follow the same pattern for `AddRemote` and `sync.remote` persistence.

## Suggested Fix

[...]

PR #3357, opened 2026-04-19
Skip Dolt remote registration for plain git source repos during init

kevglynn:
## Summary

- Gate `AddRemote("origin", syncURL)` in `bd init` on either `syncFromRemote` (remote confirmed to have `refs/dolt/data`) or `syncURLFromConfig` (URL from explicit `sync.remote` config)
- Apply the same guard to `sync.remote` persistence in `config.yaml`
- Auto-detected git origin URLs that lack Dolt data are no longer registered as Dolt remotes

## Problem

`bd init` unconditionally registered the parent git repo's `origin` URL as a Dolt remote whenever `origin` existed, even when the remote had no `refs/dolt/data`. For plain Git source repos (e.g. `gastownhall/beads` itself), this caused:

1. Every Dolt fetch operation to fail (the remote is a Go source repo, not a Dolt database)
2. Each failed fetch left a 100-900 MB `tmp_pack_*` file in `.dolt/git-remote-cache/`
3. Files accumulated at ~15 GB/day, consuming 102 GB in 7 days (#3354)

The clone path correctly gated on `syncFromRemote` (which requires `refs/dolt/data`), but `AddRemote` and `sync.remote` persistence only checked `syncURL != ""`.

## Fix

Introduce `syncURLFromConfig` to distinguish between explicit user config and auto-detected git origin. Only register the Dolt remote when:
- `syncFromRemote` is true (remote has `refs/dolt/data`), OR
- `syncURLFromConfig` is true (URL came from `sync.remote` config, meaning the user explicitly configured it)

This matches the stricter behavior already used by `bd bootstrap`'s `detectBootstrapAction`, which only sets `SyncRemote` when `refs/dolt/data` exists.

## Test plan

- [x] `go vet ./cmd/bd/` passes
- [ ] `bd init` in a plain git repo (no `refs/dolt/data`) should NOT register a Dolt remote
- [ ] `bd init` in a repo with `refs/dolt/data` should still register the remote (unchanged behavior)
- [ ] `bd init` with explicit `sync.remote` config should still register the remote (unchanged behavior)
- [ ] `bd init --stealth` in a git repo should not write `sync.remote` to config.yaml for plain git origins

## Related

- Fixes #3356
- Root cause of #3354 (tmp_pack file leak)
- Companion to #3355 (cleanup of leaked tmp_pack files)

PR #3940, opened 2026-05-14
fix(dolt): adopt git origin on first push

### LINKED ISSUE #3937: bd dolt: consistently adopt git origin as the Dolt remote
## Summary

The problem is simpler than the surrounding discussion suggests: `bd` needs one consistent policy for how the Git `origin` remote relates to the Dolt remote used for issue sync.

The architecture should be stated plainly:

> Issues live in Dolt. Cross-machine sync uses a Dolt remote, normally the same URL as Git `origin`, stored under `refs/dolt/data`. `.beads/issues.jsonl` is export/backup/viewer data, not the sync channel.

The confusing behavior is that `bd` sometimes acts like Git `origin` is authoritative, and sometimes ignores it.

## The Three Different Flows

These are currently getting mixed together in discussion, but they should have distinct behavior.

### 1. `bd init` when Git `origin` already exists

Example:

```bash
git clone git@github.com:USER/repo.git
cd repo
bd init
```

In this case, it is reasonable for `bd` to configure Git `origin` as the Dolt remote. This matches user expectation and appears to match behavior people remember from earlier releases.

Expected outcome:

```bash
bd dolt remote list
# origin  git+ssh://git@github.com/USER/repo.git
```

A later `bd dolt push` should publish issue history to `refs/dolt/data` on that same Git remote.

### 2. `bd init` before Git `origin` exists

Example:

```bash
mkdir demo && cd demo
bd init
git init
gh repo create demo --source=.
git remote add origin git@github.com:USER/demo.git
git push -u origin main
bd dolt push
```

`bd init` in a new directory with no Git `origin` should stay purely local. If there is no Git repo yet, `bd init` may still run `git init` as it does today. But if there is no Git `origin`, `bd init` should not invent or prompt for a Dolt remote. It should initialize local Dolt, configure exports/hooks/agents as normal, and optionally show a concise next-step hint:

```text
No git origin found. To sync issues later, add a git remote and run bd dolt push.
```

Today, once the user later adds `origin`, `bd dolt push` has no good way to recover. The user has to already know to run `bd dolt remote add origin <converted-origin-url>`, including the Dolt-compatible URL form.

Expected behavior:

If no Dolt remote is configured, but Git `origin` exists, `bd dolt push` should lazily adopt Git `origin` as the Dolt remote, persist it as `sync.remote`, and then push.

This is the most important missing behavior. Keeping no-origin init local preserves the `git init` mental model: local first, no network assumptions, no remote ceremony. The remote policy begins only when a Git `origin` exists or when the user explicitly passes `--remote`.

### 3. `bd bootstrap` / `bd dolt pull` in a fresh clone

Example:

```bash
git clone git@github.com:USER/repo.git
cd repo
bd bootstrap
```

Here `bd` should only auto-use Git `origin` if `refs/dolt/data` exists. Pulling or bootstrapping from a plain source-code Git remote with no Dolt data should not be treated as an existing issue database.

Expected behavior:

If `refs/dolt/data` exists, bootstrap from it and wire Git `origin` as the future Dolt remote. If it does not exist, do not invent remote Dolt history.

## Proposed Policy

Use this rule set:

1. `bd init` with existing Git `origin`: auto-adopt Git `origin` as the Dolt remote for future `bd dolt push`.
2. `bd init` with no Git `origin`: stay purely local; do not invent or prompt for a Dolt remote. Optionally print a concise hint that sync can be enabled later by adding a Git remote and running `bd dolt push`.
3. `bd dolt push` with no Dolt remote but existing Git `origin`: auto-adopt Git `origin`, persist `sync.remote`, then push, because `push` is already explicit intent to publish.
4. `bd bootstrap`: only auto-use Git `origin` when `refs/dolt/data` exists; then clone from it and persist it as the Dolt remote.
5. `bd dolt pull`: do not silently pull from a plain Git `origin` unless a Dolt remote is already configured or `refs/dolt/data` is known to exist.
6. `bd doctor`: diagnostic only unless `--fix` is passed. When Git has `origin` but bd has no Dolt remote, report the exact fix: `bd dolt remote add origin <converted-origin-url>`.
7. Docs and generated agent instructions should state the one-line architecture: Dolt is canonical; `refs/dolt/data` is sync; JSONL is export/backup only.

## Why This Is Smaller Than Previous Threads

This is not primarily a JSONL import/export problem. JSONL confusion is a symptom of the missing Dolt remote policy.

This is not primarily a hook problem. Hooks can warn, but hooks should not be the main sync mechanism.

This is not primarily a bootstrap problem. Bootstrap is one flow; lazy adoption on `bd dolt push` is the gap that affects the common `bd init` -> later add GitHub remote workflow.

## Relationship To #3909

#3909 fixed useful pieces and clarified the intended architecture, but it bundled several concerns together: init behavior, bootstrap behavior, hook warnings, JSONL fallback behavior, docs, generated website docs, and test harness changes.

The remaining follow-up should be narrower: define and implement the consistent Git-origin adoption policy above.

One important safety distinction should remain: using Git `origin` as a target for first `bd dolt push` is different from treating a plain Git remote as already containing Dolt history. For bootstrap/pull, `refs/dolt/data` should still gate automatic remote use.

## Related

- #3909
- #3688
- #3419
- #3135
- #3729
- Tracking: `mybd-ysm`
- Discovered-from investigation: https://ampcode.com/threads/T-019e239a-3691-749f-83ef-6fc3e07a4e23

---

_kilocode-openai/gpt-5.5- on behalf of maphew_

maphew:
## Summary

- Auto-adopt the active repo's git `origin` as a Dolt remote when `bd dolt push` is run with no Dolt remote configured.
- Keep `bd init` without git origin local-first, with simpler guidance to add origin and run `bd dolt push` later.
- Improve `bd doctor` no-remote diagnostics to show the exact `bd dolt remote add origin <converted-origin-url>` command when git origin exists.

## Test Plan

- `go test -tags gms_pure_go ./cmd/bd/doctor -run 'TestRemoteAdoptionDetail|TestCheckRemoteConsistency_WorktreeFallbackUsesSharedConfig'`
- `go test -tags gms_pure_go ./cmd/bd -run 'TestShouldWireInitRemote|TestIsRemoteNotFoundErr|TestPrintNoRemoteGuidance'`
- `BEADS_TEST_EMBEDDED_DOLT=1 go test -v -tags gms_pure_go ./cmd/bd -run 'TestEmbeddedInit/(git_origin_registered_as_dolt_remote|no_git_origin_stays_local|dolt_push_lazily_adopts_later_git_origin|dolt_push_adopts_target_origin_with_dash_c|stealth_skips_git_origin_remote_synthesis)|TestEmbeddedBootstrap/bootstrap_from_git_origin_wires_remote'`
- `go test -tags gms_pure_go ./cmd/bd ./cmd/bd/doctor`
- `make fmt-check`
- `make check-docs`
- `make test`

Fixes #3937.

_kilocode-openai/gpt-5.5- on behalf of maphew_

### DIFF
--- a/cmd/bd/doctor/remotes.go
+++ b/cmd/bd/doctor/remotes.go
@@ -1,15 +1,23 @@
 package doctor
 
 import (
+	"context"
 	"fmt"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/steveyegge/beads/internal/configfile"
+	"github.com/steveyegge/beads/internal/doltremote"
 	"github.com/steveyegge/beads/internal/doltserver"
 	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/storage/doltutil"
+)
+
+var (
+	querySQLRemotesForDoctor = querySQLRemotes
+	listCLIRemotesForDoctor  = doltutil.ListCLIRemotes
 )
 
 // CheckRemoteConsistency compares remotes registered in the SQL server
@@ -29,7 +37,7 @@ func CheckRemoteConsistency(repoPath string) DoctorCheck {
 	}
 
 	// Get SQL remotes via direct connection
-	sqlRemotes, sqlErr := querySQLRemotes(beadsDir)
+	sqlRemotes, sqlErr := querySQLRemotesForDoctor(beadsDir)
 	if sqlErr != nil {
 		return DoctorCheck{
 			Name:     "Remote Consistency",
@@ -43,7 +51,7 @@ func CheckRemoteConsistency(repoPath string) DoctorCheck {
 	doltDir := doltserver.ResolveDoltDir(beadsDir)
 	dbName := cfg.GetDoltDatabase()
 	dbDir := filepath.Join(doltDir, dbName)
-	cliRemotes, cliErr := doltutil.ListCLIRemotes(dbDir)
+	cliRemotes, cliErr := listCLIRemotesForDoctor(dbDir)
 	if cliErr != nil {
 		return DoctorCheck{
 			Name:     "Remote Consistency",
@@ -123,32 +131,21 @@ func CheckRemoteConsistency(repoPath string) DoctorCheck {
 
 func remoteAdoptionDetail(repoPath string) string {
 	if originURL := gitOriginRemoteURL(repoPath); originURL != "" {
-		remoteURL := normalizeGitOriginForDolt(originURL)
+		remoteURL := doltremote.Normalize(originURL)
 		return fmt.Sprintf("git origin is configured. Adopt it with: bd dolt remote add origin %s", remoteURL)
 	}
 	return "Add a remote with: bd dolt remote add origin <url>"
 }
 
 func gitOriginRemoteURL(repoPath string) string {
-	cmd := exec.Command("git", "-C", repoPath, "remote", "get-url", "origin")
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	cmd := exec.CommandContext(ctx, "git", "-C", repoPath, "remote", "get-url", "origin")
 	out, err := cmd.Output()
 	if err != nil {
 		return ""
 	}
 	return strings.TrimSpace(string(out))
-}
-
-func normalizeGitOriginForDolt(url string) string {
-	if strings.HasPrefix(url, "git+") || strings.HasPrefix(url, "file://") || strings.HasPrefix(url, "aws://") || strings.HasPrefix(url, "gs://") || strings.HasPrefix(url, "dolthub://") {
-		return url
-	}
-	if strings.HasPrefix(url, "https://") || strings.HasPrefix(url, "http://") || strings.HasPrefix(url, "ssh://") {
-		return "git+" + url
-	}
-	if idx := strings.Index(url, ":"); idx > 0 && !strings.Contains(url[:idx], "/") && strings.Contains(url, "@") {
-		return "git+ssh://" + url[:idx] + "/" + url[idx+1:]
-	}
-	return url
 }
 
 // querySQLRemotes gets remotes from the SQL server.
--- a/cmd/bd/doctor/remotes.go
+++ b/cmd/bd/doctor/remotes.go
@@ -2,6 +2,7 @@ package doctor
 
 import (
 	"fmt"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
@@ -58,7 +59,7 @@ func CheckRemoteConsistency(repoPath string) DoctorCheck {
 			Name:     "Remote Consistency",
 			Status:   StatusWarning,
 			Message:  "No remotes configured",
-			Detail:   "Add a remote with: bd dolt remote add origin <url>",
+			Detail:   remoteAdoptionDetail(repoPath),
 			Category: CategoryData,
 		}
 	}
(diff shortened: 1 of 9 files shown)


maphew:
## PR 3940 Review — `fix(dolt): adopt git origin on first push`

**Scope and direction:** Solid. The PR cleanly implements policy items (2), (3), and (6) from issue #3937 (lazy origin adoption on push, init-without-origin stays local, doctor shows the exact fix). Items (1) ("auto-adopt at init") and (4) ("bootstrap with refs/dolt/data") are left to pre-existing code, which appears intentional given the tests reference existing `git_origin_registered_as_dolt_remote` and `bootstrap_from_git_origin_wires_remote` cases. Test coverage is good — `dash_c` and lazy-adoption cases are both exercised.

### Issues worth addressing

**(1) Duplicated URL normalization (medium)**

`cmd/bd/doctor/remotes.go` introduces `normalizeGitOriginForDolt`, which re-implements `normalizeRemoteURL` from `cmd/bd/sync_remote.go`. The doctor's "exact fix" hint and the actual adoption path (`adoptGitOriginRemoteForPush` → `normalizeRemoteURL`) are guaranteed to produce identical strings *today* but only by coincidence — they're maintained in two places with no shared call site. Two divergences already visible:

- Doctor doesn't handle Windows drive paths; production prepends `git+`.
- Doctor's native-scheme list is bare-prefix `git+` (catches `git+http://, git+https://, git+ssh://`) while production enumerates them explicitly. Equivalent now, not stable across edits.

Recommended: move the normalizer into an internal package both can import (e.g. `internal/doltremote`), or expose a helper from `cmd/bd` the doctor can call. The whole point of the doctor message is "run this exact command" — it has to keep matching production verbatim.

**(2) Partial state on adoption failure (medium)**

`adoptGitOriginRemoteForPush` in `cmd/bd/dolt.go` does:

```
AddRemote → (SQL CLI sync, warning-only) → check beadsDir → SetYamlConfig → commit
```

If `selectedDoltBeadsDir()` returns `""` *after* `AddRemote` has already succeeded, the function returns an error with the Dolt remote written but `sync.remote` in `config.yaml` missing — exactly the divergent state the issue is trying to eliminate. Move the `beadsDir == ""` check above `AddRemote` (and ideally above `gitOriginGetURLForActiveRepo` so we fail-fast before any external call). Same logic for `SetYamlConfigInDir` failure: recoverable on retry, but the user has no signal that the Dolt-side mutation already happened.

**(3) Inconsistent git execution path (minor)**

`cmd/bd/doctor/remotes.go` uses raw `exec.Command("git", "-C", repoPath, ...)`, but everywhere else this PR is careful to go through `RepoContext.GitCmd` (the whole motivation for the new `gitOriginGetURLForActiveRepo` and `commitBeadsConfigForActiveRepo` helpers — they correctly thread `-C` for the `dash_c` test case). The doctor path is called with an explicit `repoPath` already, so it works, but you're now maintaining two patterns. Worth a small follow-up to consolidate.

**(4) No timeout on `gitOriginGetURLForActiveRepo`** (`cmd/bd/sync_git.go`)

The sibling `gitOriginHasDoltDataRef` uses a 10s timeout because it does network work. `remote get-url` is purely local, so this isn't a real bug, but inheriting a context from the caller (and respecting cancellation) would be more in keeping with the file's style.

### Minor

- `printInitNoDoltRemoteWarning` (`cmd/bd/init.go`) now says `bd dolt push` will Do The Right Thing, which is correct under the new policy. Worth confirming this prints *only* when git origin is genuinely absent at init time — re-reading it as a user, "add a git origin and then run `bd dolt push`" is a clear instruction.
- `CheckRemoteConsistency` itself isn't tested with the new detail wired in — only the helper `remoteAdoptionDetail` is. A two-line test would close the gap (build a repo with git origin, no Dolt remote, assert `CheckRemoteConsistency(...)` returns the new detail).
- The PR doesn't touch `bd dolt pull` (policy item (5)). Good — out of scope and called out in the issue as a separate concern.

### Summary

Approve-with-nits. Most pressing is (1) (de-duplicate the URL normalizer — the doctor's whole value is producing an exact match); (2) is a small reorder. Tests look right for what's claimed.

---

_Edit: rewrote `#N` references as `(N)` to avoid GitHub autolinking item numbers to unrelated issues._

_claude-opus-4.7-high on behalf of maphew_

maphew:
Addressed the review nits in fe1adad13:

- moved Dolt remote URL normalization into shared `internal/doltremote` and wired both push adoption and doctor guidance through it
- moved the selected beads workspace check ahead of Dolt remote mutation in `adoptGitOriginRemoteForPush`
- added timeout-bounded git origin lookup in doctor guidance
- added `CheckRemoteConsistency` coverage for the no-remotes/git-origin adoption detail

Validation:
- `go test -tags gms_pure_go ./internal/doltremote ./cmd/bd ./cmd/bd/doctor`

_codex-gpt-5-medium on behalf of maphew_

maphew:
Regression investigation for #3937:

@coffeegoddd note matches the history. The behavior changed between `v1.0.2` (`a3f834b31`) and `v1.0.3` (`1b2dd2cb5`). The key commit is `ac085df8c58b2d408928e90d55e9cf43f5286e79` (`Skip Dolt remote registration for plain git source repos during init (#3357)`, fixes #3356). Before that, `bd init` set `syncURL` from git `origin` and registered it whenever `syncURL != ""`; after that commit, registration and `sync.remote` persistence were gated through `shouldWireInitRemote`, so a plain git origin without `refs/dolt/data` was no longer registered.

That v1.0.3 change was intentional and correct for #3356/#3354: registering a plain source repo as a Dolt remote caused failed Dolt fetches and `tmp_pack_*` disk growth. The user-visible regression in #3937 is the lost convenience path where a repo with git `origin` could run `bd dolt push` and have a usable destination. `395935bed` also changed no-remote push/pull to exit 0 with guidance, so after `ac085df8c` the workflow became a silent-ish skip instead of pushing.

This PR restores the v1.0.2 convenience at push time, not init time: `bd init` still does not wire plain git origins, but `bd dolt push` lazily adopts the current git `origin` only when the user explicitly asks to push and no Dolt remote exists. That preserves the #3356/#3354 protection while fixing #3937.

_codex-gpt-5-medium on behalf of maphew_

Issue #2994, opened 2026-04-03
updating beads has now failed, hard, 2 out of 2 times


josh-m-sharpe:
Few weeks ago I tried upgrading across the "dolt is required" boundary.  Think I've been running 0.55.4 for the last few weeks so it was probably some <0.55.4 up to 0.55.4 where this happened.  It was generally a disaster - although I dont recall specifics.  There was much downgrading, and re-upgrading, and reverting the doctor recommended fixes.  

Today I, for funsies, ran `bd doctor` just to check in, maybe there were some things I should fix.    It recommended I upgrade from 0.55.4 to 1.0 via 'brew upgrade beads'

I did.

Then I ran `bd list` and:

```
bd list
Warning: /Users/jsharpe/proj/.beads has permissions 0755 (recommended: 0700). Run: chmod 700 /Users/jsharpe/proj/.beads
2026/04/03 14:27:03 migration 010: converting events.id from bigint to CHAR(36) UUID
2026/04/03 14:27:03 migration 010: events.id migrated to CHAR(36) UUID successfully
2026/04/03 14:27:03 migration 010: converting comments.id from bigint to CHAR(36) UUID
2026/04/03 14:27:04 migration 010: comments.id migrated to CHAR(36) UUID successfully
2026/04/03 14:27:04 migration 010: converting issue_snapshots.id from bigint to CHAR(36) UUID
2026/04/03 14:27:04 migration 010: issue_snapshots.id migrated to CHAR(36) UUID successfully
2026/04/03 14:27:04 migration 010: converting compaction_snapshots.id from bigint to CHAR(36) UUID
2026/04/03 14:27:04 migration 010: compaction_snapshots.id migrated to CHAR(36) UUID successfully
No issues found.
```

poof!.  no beads!

Does the 1.0 track have a stable upgrade path?
Should we not trust what doctor recommends?

josh-m-sharpe:
`bd import .beads/issues.jsonl` seemed to resolve this

josh-m-sharpe:
not sure, but we probably want these things in .beads/.gitignore?    all this stuff was created after importing, looks like it shouldn't be committed?

```
diff --git a/.beads/.gitignore b/.beads/.gitignore
index dba6914..d2a9f7c 100644
--- a/.beads/.gitignore
+++ b/.beads/.gitignore
@@ -1,6 +1,19 @@
 # Dolt database (managed by Dolt, not git)
 dolt/
 dolt-access.lock
[...]

PR #3944, opened 2026-05-14
fix: skip JSONL auto-import and auto-export when serverMode=true

dkhaye:
fix: skip JSONL auto-import and auto-export when serverMode=true

## What

Two related guards added to prevent unnecessary store access in server mode:

1. `maybeAutoImportJSONL` now accepts a `serverMode bool` parameter and returns immediately when true.
2. `maybeAutoExport` now accepts a `serverMode bool` parameter and returns immediately when true.

## Why

**Auto-import (#2994):** The auto-import introduced in #2994 (upgrade recovery for users migrating from pre-0.56 `.beads/dolt/` to 1.0+ `.beads/embeddeddolt/`) fires on every write command in server mode (`dolt_mode=server` in `metadata.json`), where it is never relevant. In server mode the database lives on a persistent external `dolt sql-server`; the "empty database" condition that triggers the recovery never applies. The fallback import path conflicts with `initSchema`'s open transaction, producing a spurious Error 1213 serialization failure warning on every write.

**Auto-export (this PR):** The auto-export path performs git operations and store reads after every write command. In server mode these are also irrelevant and add unnecessary overhead. Mirroring the same guard keeps the two paths consistent.

## How

**Commit 1 — auto-import:**
- Added `serverMode bool` to `maybeAutoImportJSONL` (early return when true)
- Updated the sole call site in `main.go` to pass `doltCfg.ServerMode`
- Added `TestMaybeAutoImportJSONL_SkipsEntirelyInServerMode` (no CGO required, passes nil store safely) to `auto_import_upgrade_unit_test.go`
- Updated the three existing unit tests to pass `serverMode=false`

**Commit 2 — auto-export:**
- Added `serverMode bool` to `maybeAutoExport` (early return when true)
- Updated the call site in `PersistentPostRun` to pass the global `serverMode` variable
- Added `TestMaybeAutoExportJSONL_SkipsEntirelyInServerMode` to `export_auto_test.go`: leaves `store=nil` and verifies no panic (store access would panic, so clean return proves the guard fires first)

## Verification

```bash
# Unit tests (no CGO, no running server required)
go test -tags gms_pure_go -run 'TestMaybeAutoImportJSONL|TestMaybeAutoExportJSONL|TestShouldRunPostCommand' ./cmd/bd/

# Full test suite
make test

# Manual: run any write command against a server-mode store and confirm
# neither "auto-importing N bytes from ..." nor auto-export output appears
bd update <any-id> --status open
```

### DIFF
--- a/cmd/bd/auto_import_upgrade.go
+++ b/cmd/bd/auto_import_upgrade.go
@@ -68,6 +68,11 @@ func writeAutoImportStamp(beadsDir string, info os.FileInfo) {
 // .beads/dolt/) to 1.0+ (which uses .beads/embeddeddolt/) don't appear to
 // lose their issues.  See GH#2994.
 //
+// serverMode must be true when the store is connected to an external
+// dolt sql-server. In server mode the database is persistent and shared;
+// the JSONL recovery path is irrelevant and attempting it causes a
+// serialization conflict with initSchema's open transaction (Error 1213).
+//
 // The top-level emptiness guard (GetStatistics) is the primary
 // protection for BOTH the embedded fast-path and the server-mode
 // fallback. Defense in depth backs each path up: the embedded
@@ -81,7 +86,10 @@ func writeAutoImportStamp(beadsDir string, info os.FileInfo) {
 //
 // The function is best-effort: failures are logged as warnings but do not
 // prevent the store from being used.
-func maybeAutoImportJSONL(ctx context.Context, s storage.DoltStorage, beadsDir string) {
+func maybeAutoImportJSONL(ctx context.Context, s storage.DoltStorage, beadsDir string, serverMode bool) {
+	if serverMode {
+		return
+	}
 	// Quick check: does the JSONL file exist and have content?
 	jsonlPath := configuredImportJSONLPath(beadsDir)
 	info, err := os.Stat(jsonlPath)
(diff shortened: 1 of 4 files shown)


coffeegoddd (commented):
Can we do the same for the auto-export to jsonl too? Both need to be disabled when using server mode.

dkhaye:
> Can we do the same for the auto-export to jsonl too? Both need to be disabled when using server mode.

@coffeegoddd Updated

maphew:
Assessment after merged #4063 (`3988db7e0da1d5488e5ff972349f736466bec1c8`): partially addressed.

#4063 materially reduces the auto-export part for new/default-config repos by making JSONL auto-export opt-in (`export.auto=false`, `export.git-add=false`) and changing `bd init` to `[y/N]`.

The server-mode-specific behavior from this PR still is not present in merged main: auto-import is still called without a server-mode guard, `maybeAutoImportJSONL` still has no `serverMode` parameter/early return, auto-export is still called as `maybeAutoExport(rootCtx)`, and `maybeAutoExport` has no server-mode guard.

Residual risk: existing repos or users who opt into `export.auto=true` can still run auto-export in server mode, and the auto-import/server-mode fallback concern appears unaffected.

Recommended action: keep this open, rebase over current `main`, and land the explicit `serverMode` early returns. #4063 mitigates default auto-export exposure, but it does not address auto-import in server mode or prevent opted-in auto-export in server mode.

dkhaye:
Superceded by #4091

PR #3947, opened 2026-05-15
Fix Dolt endpoint resolution and port handoff safety

hexsprite:
## Summary

- Add an explicit Dolt endpoint resolver that reports host, port, database, owner, source, and dialability.
- Add `WriteResolvedPort` so `.beads/dolt-server.port` writes require an explicit lifecycle/config/doctor reason.
- Stop storage open and read-oriented Dolt CLI paths from silently repairing or rewriting the port file.
- Make `bd dolt show/status/test/stop` reason about unresolved, external, shared, and gc-managed endpoints instead of treating port `0` as dialable or relying only on Beads-owned PID state.

## Validation

- `scripts/pr-preflight.sh --search "Dolt server port resolution gc-managed handoff split-brain" --repo gastownhall/beads`
- `go test ./internal/doltserver`
- `go test ./internal/storage/dolt`
- `go test -tags gms_pure_go ./cmd/bd -run 'TestRenderLocalDoltStatus|TestRunExternalDoltStatus'`

`make test` was also run, but it failed on local/environmental issues outside this patch: `cmd/bd` timed out in `TestContextUsesExplicitDBFlagForNoDBCommand`, and role tests observed local `beads.role=maintainer` from git config.

### DIFF
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -411,6 +411,16 @@ on the next bd command unless auto-start is disabled.`,
 		if beadsDir == "" {
 			FatalErrorWithHint(activeWorkspaceNotFoundError(), diagHint())
 		}
+		endpoint, _, err := doltserver.ResolveEndpoint(beadsDir, doltserver.ResolveOptions{Purpose: doltserver.EndpointPurposeLifecycle})
+		if err != nil {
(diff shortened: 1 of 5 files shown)


coffeegoddd:
we are rewriting storage and server, so this pr wont land. if you have a bug please file an issue and we can investigate if it persists after our changes are in.

Issue #3948, opened 2026-05-15
Auto-import upgrade path fires on every bd command despite non-empty database; causes JSONL→dolt overwrite race


Wldc4rd:
## Summary

`maybeAutoImportJSONL` (cmd/bd/auto_import_upgrade.go:44) is documented as a one-time migration recovery for upgrading from pre-0.56 stores. The `stats.TotalIssues > 0` guard at line 64 is supposed to make it a no-op on non-empty databases. In our installation, that guard is not firing — the auto-import runs on **every** bd command, even when `SELECT COUNT(*) FROM issues` would return 400+.

Because the fallback path's `importFromLocalJSONLFull` uses `INSERT ... ON DUPLICATE KEY UPDATE` semantics (the source comment at lines 36-40 explicitly calls this out), every bd invocation overwrites recent dolt-side state with whatever is in `.beads/issues.jsonl`. In a multi-writer setup (gc supervisor's reconciler calling bd every ~5s), this creates a race where any direct dolt write (SQL update, or bd write whose JSONL export hasn't fired yet due to the 60s `export.interval` throttle) gets silently reverted within seconds.

## Evidence from a live installation

Dolt commit log on the gc-managed beads database `hq` (479 issues currently):

```sql
SELECT commit_hash, date, message FROM dolt_log
WHERE message LIKE 'bd: create % issue(s)' ORDER BY date DESC LIMIT 10;

| date                    | message                 |
| 2026-05-15 03:44:11.102 | bd: create 479 issue(s) |
| 2026-05-15 03:43:59.529 | bd: create 479 issue(s) |
| 2026-05-15 03:43:55.188 | bd: create 479 issue(s) |
| 2026-05-15 03:43:51.046 | bd: create 479 issue(s) |
| 2026-05-15 03:43:46.699 | bd: create 479 issue(s) |
| 2026-05-15 03:43:41.827 | bd: create 479 issue(s) |
| 2026-05-15 03:43:36.324 | bd: create 479 issue(s) |
| 2026-05-15 03:43:31.065 | bd: create 479 issue(s) |
| 2026-05-15 03:43:27.033 | bd: create 479 issue(s) |
| 2026-05-15 03:43:14.559 | bd: create 479 issue(s) |
```

[...]

Issue #3870, opened 2026-05-10
v1.0.4 ships data-clobber regression — please cut v1.0.5 with #3691


csauer02-personal-user:
## Summary

v1.0.4 is the latest tagged release and contains the regression that #3691 fixed. The fix merged 7 hours after the v1.0.4 cut, so anyone on `latest` hits the bug. Please tag v1.0.5.

## Timeline

| When (UTC) | What |
|---|---|
| 2026-05-01 | PR #3630 merged — moved the auto-import emptiness guard inside the embedded path, leaving server-mode fallback unguarded |
| 2026-05-09 15:11 | **v1.0.4 released** (commit `ce242a8`) — ships with #3630, no fix |
| 2026-05-09 21:50 | PR #3691 merged (commit `1cf8337`) — restores the guard |
| 2026-05-10 | now: no v1.0.5 yet; `git tag --contains 1cf833734` is empty |

## Live impact (production-ish setup)

bd 1.0.4, server-mode dolt, `dolt.auto-commit=off` (default), Linux.

The clobber doesn't just affect `bd update --append-notes` (the reproducer in #3691). It also reverts `bd close` and `gc session close` on the next bd invocation:

```
$ bd close i4-n1u --reason-file note.md
✓ Closed i4-n1u — Cleanup orphan i4lcrs Dolt store directories

$ bd show i4-n1u | head -1
○ i4-n1u · ... [● P2 · OPEN]      ← clobbered back to OPEN

$ bd dolt commit       # forces JSONL re-export
Committed.

$ bd close i4-n1u --reason "second try"
✓ Closed i4-n1u — ...

$ bd show i4-n1u | head -1
✓ i4-n1u · ... [● P2 · CLOSED]    ← finally sticks
```

Same on `gc convoy check`:

```
$ gc convoy check
Auto-closed convoy i4-3dg "sling-i4-a5d.1"
2 convoy(s) auto-closed

$ bd show i4-3dg | head -1
○ i4-3dg · ... [● P2 · OPEN]      ← still open

# next invocation re-reports the same convoys
$ gc convoy check
Auto-closed convoy i4-3dg "sling-i4-a5d.1"
2 convoy(s) auto-closed
```

[...]

PR #3960, opened 2026-05-15
fix(auto-import): make upgrade-recovery fallback non-destructive (#3955)

### LINKED ISSUE #3955: Harden auto-import: non-destructive (insert-if-new) fallback so an empty-DB guard regression can't corrupt data
## Summary

Defense-in-depth follow-up to #3948. **This is not the #3948 incident fix** — that incident is the missing PR #3691 guard in v1.0.4, resolved by shipping #3691 (#3870). This issue makes the auto-import path *structurally* unable to corrupt data even if the emptiness guard regresses again.

## Background

`maybeAutoImportJSONL` (`cmd/bd/auto_import_upgrade.go`) is upgrade-recovery **into an empty DB** (pre-0.56 → 1.0 store-path migration, GH#2994). Its server-mode fallback imports issues via `INSERT … ON DUPLICATE KEY UPDATE` (`InsertIssueIntoTable`, `internal/storage/issueops/helpers.go`), overwriting `status`, `title`, `metadata`, etc.

So a *recovery-into-empty* operation is implemented with *destructive UPSERT* semantics — safe only while the emptiness guard is perfect. PR #3630 already proved a guard regression can pass review and ship (#3948): the result was silent JSONL→Dolt overwrite of live rows plus a self-feeding O(file) re-import storm that DoS'd a shared Dolt server.

## Proposal

Make the auto-import fallback **insert-if-new** instead of UPSERT, so an existing row is skipped, not overwritten:

- Add a conflict-skip option to `storage.BatchCreateOptions`.
- Thread it: `importFromLocalJSONLFull` → `importIssuesCore` → `CreateIssuesWithFullOptions` → issue insert.
- Branch `InsertIssueIntoTable` to insert-if-new when set (precedent: `InsertIssueIfNew`).
- **Auto-import fallback path only.** Explicit `bd import` keeps current UPSERT semantics.

Result: if the guard regresses again, auto-import degrades to a harmless no-op (`nothing to commit`) instead of corrupting data.

## Scope / non-goals

- Does **not** fix the #3948 incident (that is v1.0.5 with #3691, #3870).
- Embedded `ImportJSONLData` fast path already has an in-tx emptiness check; symmetry there is optional.
- `bd doctor` multi-binary / guard-missing-version check is a separate (also worthwhile) change.

## Risk

Low — auto-import legitimately runs only on an empty DB, where insert-if-new and UPSERT are equivalent. Ships independently of any release.

## Refs

#3948 (incident), #3630 (the regression that proves the risk), #3691 (guard restore in `main`), #3870 (release tracking), #3835 (v1.0.3 pin workaround).

esciara:
## Summary

Defense-in-depth (Layer 2) follow-up to incident #3948. **This is not the incident fix** — the live incident is already fixed in `main` by #3691 and needs only the v1.0.5 release cut (#3870). This PR is independent structural insurance and does not block / is not blocked by the release.

The auto-import upgrade-recovery fallback (`maybeAutoImportJSONL` server-mode path) imported JSONL via `INSERT … ON DUPLICATE KEY UPDATE`. Auto-import is *recovery into an empty DB* but was implemented with *destructive UPSERT* semantics — safe only while the emptiness guard is perfect. PR #3630 already proved a guard regression can pass review and ship silent data corruption + a shared-Dolt-server DoS. This PR removes the catastrophe from the failure mode: if the guard ever regresses again, auto-import degrades to a harmless no-op instead of clobbering live rows.

## Changes

- Add `BatchCreateOptions.ConflictSkip` (`internal/storage/batch.go`) — insert-if-new instead of UPSERT.
- Thread it through the auto-import fallback seam **only**: new `importFromLocalJSONLConflictSkip` becomes the `fallbackImporter` default (`cmd/bd/auto_import_upgrade.go`); `importFromLocalJSONLFull` / `importFromLocalJSONLWithOpts` split out the shared impl (`cmd/bd/import_shared.go`).
- `InsertIssueIfNew` (`internal/storage/issueops/create.go`) takes a `conflictSkip` arg: when set and the row exists, it returns `(false, nil)` and never runs the UPSERT.

## Scope / guarantees

- **Auto-import fallback path only.** Explicit `bd import`, `bd bootstrap`, and `bd init --from-jsonl` keep current UPSERT semantics — unchanged.
- On the legitimate path (empty DB) insert-if-new and UPSERT are equivalent, so no behavior change for real recovery.
- Net effect: a future emptiness-guard regression can no longer corrupt data — worst case is a no-op (`nothing to commit`).

## Tests

- `cmd/bd/auto_import_upgrade_unit_test.go` — option-threading / call-count behavior via the `fallbackImporter` seam.
- `cmd/bd/auto_import_upgrade_test.go` — embedded integration coverage (`BEADS_TEST_EMBEDDED_DOLT=1`): conflict-skip leaves a pre-existing row unmutated while still inserting new IDs; empty-DB recovery still imports all JSONL issues (negative control).

Verification per the plan: `go build ./...` and `go test ./cmd/bd/... ./internal/storage/...`.

## Out of scope

- The live incident fix (v1.0.5 release cut with #3691, tracked in #3870) + the already-documented v1.0.3 pin. This PR must not be treated as the incident fix.
- Dropped Layers 1 & 3 — their premise (`GetStatistics` returns 0 on a populated DB) is falsified by the v1.0.3-vs-v1.0.4 A/B in the issue thread.

Closes #3955


maphew (approved):
## Maintainer review — approving, with conflict resolution + a defense-in-depth increment pushed

I built this fix independently before finding this PR, and the two converged on the **same design** — `BatchCreateOptions.ConflictSkip`, `InsertIssueIfNew(conflictSkip)` returning `(false, nil)` when the row exists, threaded through the auto-import fallback seam only, with `bd import`/`bd bootstrap`/`bd init --from-jsonl` keeping UPSERT. We also independently reached the same conclusion that the "`GetStatistics` returns 0 on a populated DB" premise is falsified by the v1.0.3-vs-v1.0.4 A/B in #3948. So this is well-corroborated; thanks @esciara.

This PR was `CONFLICTING` against `main` (13 days of drift) and had no reviews, so I used maintainer edit to land two commits on the branch rather than open a competing PR:

**1. Merge `main` + resolve conflicts** (`ce1b527`):
- `batch.go`: kept `ConflictSkip` next to main's new `SkipDependencyValidationErrors` / `OnSkippedDependency`.
- `import_shared.go`: threaded `ConflictSkip` into `importIssuesCore`'s options on top of main's stale-import filtering / skipped-dependency tracking.
- `auto_import_upgrade.go`: kept `importFromLocalJSONLConflictSkip` as the `fallbackImporter` default.
- `promote.go`: a new `InsertIssueIfNew` caller landed on main; passed `conflictSkip=false` (promotion keeps UPSERT).
- `auto_import_upgrade_unit_test.go`: `captureOptsStore` now stubs `GetIssuesByIDs` so main's stale-import filter is a no-op under the option-threading unit test (it was panicking on the nil-interface fake after the merge).

**2. Harden the embedded fast-path** (`805e9c1`):
- The server-mode fallback was covered, but `EmbeddedDoltStore.ImportJSONLData` — the primary auto-import route for 1.0+ users — was still UPSERT, gated only by its in-transaction emptiness check. Set `ConflictSkip: true` there too, so a regression in that check can't clobber live rows either. This completes the #3955 guarantee across **both** auto-import paths; the legitimate empty-DB path is unchanged.

Verified locally (`-tags gms_pure_go`, `CC=/usr/bin/gcc`, `BEADS_TEST_EMBEDDED_DOLT=1`): your `ConflictSkip` unit + embedded integration tests, `TestImportFromLocalJSONL` (UPSERT still updates on explicit re-import), and `internal/storage/...` all green. `go build ./...` clean.

Approving. Leaving the merge to land once CI settles — not force-merging given unrelated checks still running.

PR #3984, opened 2026-05-16
feat: bdd daemon + postgres backend observability stack

quad341:
### 1. `bd daemon` subcommands

`bd daemon status`, `bd daemon kill`, and `bd daemon stats` provide visibility into the bdd
daemon process. `bd daemon status` shows running state, PID, socket path, uptime, idle
counter, and RPC call totals in text or `--json`. `bd daemon kill` sends SIGTERM (or SIGKILL
with `--force`) and reports the result. `bd daemon stats` queries per-method call counts and
in-flight RPCs directly from the running daemon.

`bd sql` now exits 1 (not 0) when the daemon is active and cannot provide direct database
access — scripts and CI that check exit codes will correctly detect the failure.

`bd init --reinit` kills any running daemon before modifying workspace metadata to prevent
stale locks.

### 2. Postgres backend observability

`bd backend status` probes whichever backend is configured and reports health, version, and
connection target in text or `--json`. Exit 0 on healthy, exit 1 on unhealthy. Passwords are
never included in output or error messages.

BEADS_POSTGRES_* environment variables are now applied at runtime on every `bd` invocation,
not only at init time. `TestOpen_ErrorFormat_WithOverride` verifies the error path includes
the override hint without leaking credentials.

### 3. Legacy Dolt artifact detection

`bd doctor` on a Postgres workspace now reports `.beads/dolt` as an INFO advisory (not a
failure) when the live data directory is absent but legacy metadata fields remain. The
message includes a copy-pasteable `mv` command with today's date for remediation. This check
is backend-gated: it returns nothing on Dolt workspaces, preventing false positives.

## Review notes

- `bd sql` exit-code change is the only externally visible behavior change for Dolt users.
  Scripts that previously swallowed exit 0 on daemon-mode sql errors will now see exit 1 — this
  is the correct POSIX behavior.
- Postgres backend is behind `--experimental` for `bd init`; existing Dolt workspaces are
  unaffected.
- All password handling goes through `pgdsn.RenderRedacted`; the `TestOpen_ErrorFormat_WithOverride`
  test (DNS fail path) confirms no credential leak.
- `TestDaemonChild_*` tests require a pre-built binary via `BEADS_TEST_BD_BINARY` in CI;
  they pass with a pre-built binary and are documented in the gate.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `TestOpen_ErrorFormat_WithOverride` — PASS 0.005s
- [x] `TestInitBackendFlag/unknown_backend_errors` — PASS (updated by be-vk7ovi to use truly
  unknown backend value now that postgres is a valid backend)
- [x] Legacy check suite (`TestCheckStaleLegacyHooks`, `TestCheckLegacyBeadsSlashCommands`,
  `TestCheckLegacyMCPToolReferences`) — all PASS
- [x] Release gate: [`release-gates/be-t5hh3u-stack-gate.md`](release-gates/be-t5hh3u-stack-gate.md)


quad341:
Closing — walking back the PG-driver direction. Code preserved on `parking/pg-driver-shelved-2026-05-17`. Thanks for the time on review.

PR #3995, opened 2026-05-16
auto-import: gate JSONL upgrade-recovery to embedded server mode

shaunc:
## Problem

`maybeAutoImportJSONL` (`cmd/bd/auto_import_upgrade.go`, GH#2994) is
the **embedded-DB upgrade-recovery** path: it exists so users
upgrading from pre-0.56 (`.beads/dolt/`) to 1.0+
(`.beads/embeddeddolt/`) don't appear to lose data when the new
embedded DB starts empty but the git-tracked `issues.jsonl` still has
everything.

But `cmd/bd/main.go` calls it in **every** server mode:

```go
if store != nil && !useReadOnly && !globalFlag && cmd.Name() != "import" {
    maybeAutoImportJSONL(rootCtx, store, beadsDir)
}
```

In `ServerModeOwned` / `ServerModeExternal` (shared-server) there is
no per-clone embedded DB to recover into. The function's
`GetStatistics` "is the DB empty?" probe runs before the session has
selected the configured database and races shared-server
reachability, so it **intermittently reports empty** even though the
shared server is fully populated. It then fires a spurious
auto-import against the wrong/empty view — on **every command**
whenever the server is transiently slow/unreachable. Observed
fleet-wide as a recurring:

```
auto-importing N bytes from .beads/issues.jsonl into empty database...
warning: auto-import from .beads/issues.jsonl failed: ...
```

The real data is safe on the shared server; normal command execution
(which does select the configured DB) works correctly. The
auto-import is pure noise + wasted work in server mode, and turns any
non-issue JSONL record into a scary hard failure.

## Fix

Gate the call to `ServerModeEmbedded` only:

```go
if store != nil && !useReadOnly && !globalFlag && cmd.Name() != "import" &&
    doltserver.ResolveServerMode(beadsDir) == doltserver.ServerModeEmbedded {
    maybeAutoImportJSONL(rootCtx, store, beadsDir)
}
```

`ResolveServerMode` is already the single source of truth (env /
`metadata.json`) and is test-covered
(`TestResolveServerMode_*`). The auto-import only runs in the legacy
in-process embedded path it was written for; Owned/External never
trigger it.

## Why this is correct

- `ServerModeEmbedded` is *defined* as "the legacy in-process
  embedded dolt path" (`internal/doltserver/servermode.go`) — exactly
  the upgrade scenario `maybeAutoImportJSONL` targets.
- Owned/External manage a real server; an empty-DB probe there is a
  connection/selection-timing artifact, not a genuine "needs
  recovery from JSONL" signal. Auto-importing on that false signal is
  never desired.
- Behavior-preserving for genuine embedded users (mode == Embedded →
  unchanged).

## Verification

- `go build ./cmd/bd/` clean; `go vet ./cmd/bd/` clean.
- `go test ./internal/doltserver/ -run 'TestResolveServerMode|TestIsSharedServerMode'` green (resolver behavior the gate relies on).
- `go test ./cmd/bd/ -run TestImportFromLocalJSONL` green (no import regression).

### DIFF
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -1009,7 +1009,17 @@ var rootCmd = &cobra.Command{
 		// Skip auto-import when the user is explicitly running "bd import" —
 		// the import command handles JSONL files itself and auto-importing
 		// first would interfere (double-import / upsert confusion).
-		if store != nil && !useReadOnly && !globalFlag && cmd.Name() != "import" {
+		//
+		// Gate to ServerModeEmbedded only. In Owned/External (shared)
+		// server mode there is no per-clone embedded DB to recover into:
+		// the empty-DB probe runs before the session selects the
+		// configured database, always reports "empty", and fires a
+		// spurious auto-import against the wrong/empty view on every
+		// command. The data is safe on the shared server; the import
+		// just fails noisily. Auto-import is only meaningful for the
+		// legacy in-process embedded path it was written for.
+		if store != nil && !useReadOnly && !globalFlag && cmd.Name() != "import" &&
+			doltserver.ResolveServerMode(beadsDir) == doltserver.ServerModeEmbedded {
 			maybeAutoImportJSONL(rootCtx, store, beadsDir)
 		}
 


quad341 (approved):
Thanks @shaunc — spelling out the embedded vs shared-server distinction makes this fix easy to verify. Keeping JSONL recovery gated on `ServerModeEmbedded` matches the documented upgrade-recovery intent of `maybeAutoImportJSONL`, and avoids the misleading empty-DB probe when running against an Owned or External Dolt server.

CI is fully green across all platforms, including upgrade smokes and embedded Dolt tests. Approving and flagging for merge.


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`


quad341:
## beads/reviewer — PASS ✓

Thanks @shaunc for a tight, well-motivated fix!

**Problem nailed**: The empty-DB probe in `maybeAutoImportJSONL` racing shared-server reachability is exactly the right diagnosis. The spurious auto-import on every command when the server is transiently slow is a real UX regression.

**Fix is correct**: Gating to `ServerModeEmbedded` using `ResolveServerMode` (the single source of truth, already tested via `TestResolveServerMode_*`) is the precise scope this function was written for. Owned/External modes have no per-clone embedded DB to recover into — the gate eliminates the false-positive path entirely.

### Findings

None.

- `cmd/bd/main.go`: Added comment explains the non-obvious *why* correctly. Condition is readable and correctly placed.
- **Coverage**: Codecov flags 0% on 6 lines in `PersistentPreRun` — this is a startup bootstrap path, genuinely hard to unit test in isolation, and covered end-to-end by the cross-version upgrade smoke tests (all 5 versions pass).
- **CI**: All 36 checks green across ubuntu/macos/windows + cross-version smokes.

Ready to merge.

🤖 beads/reviewer

quad341:
**Ensemble review** (Qwen + Codex; synthesis pass degraded — claude-mux unavailable)

**Verdict: auto-merge** — Codex high-confidence; Qwen fix-merge but its formatting concern is a false positive (the comment block style matches the surrounding file).

The fix is exactly right: `maybeAutoImportJSONL` was designed for the embedded-mode upgrade-recovery scenario and only makes sense there. The empty-DB probe in Owned/External mode is a connection-timing artifact, not a recovery signal — gating to `ServerModeEmbedded` cleanly eliminates the spurious auto-import noise without affecting the one path that actually needs it.

### Top findings

- ✅ One-line logic change is correct: `ResolveServerMode(beadsDir) == doltserver.ServerModeEmbedded` is the right boundary
- ✅ `ResolveServerMode` is already covered by `TestResolveServerMode_*` tests
- ✅ `TestImportFromLocalJSONL` and full CI confirm no embedded-mode regression
- ✅ Comment block accurately explains the root cause and why the gate is safe
- ✅ All CI checks green; already approved

No maintainer-side changes needed.

---

@shaunc thanks for the precise diagnosis and fix! The failure mode explanation in the PR body is exactly the right level of detail — makes the gate change immediately obvious. CI is clean across the full matrix including upgrade smokes.

maphew:
Assessment after merged #4063 (`3988db7e0da1d5488e5ff972349f736466bec1c8`): partially addressed, not materially fixed.

#4063 is useful but does not touch the behavior at issue here: it makes JSONL auto-export opt-in by default and updates init/config/docs for export semantics, but it does not modify the auto-import recovery call path in `cmd/bd/main.go`.

The false-positive in Owned/External shared-server modes remains possible because `maybeAutoImportJSONL(...)` can still be reached there. As a result, #3995’s `ServerModeEmbedded` gate is still needed.

Recommended action: keep #3995 open and merge it; optionally add regression coverage so `maybeAutoImportJSONL` is never invoked in `ServerModeOwned` / `ServerModeExternal`.

quad341:

maphew:
The bug targeted here is already over the line; closing as superseded. Apologies for the noise and delay in getting here; there's some tuning work to do on these here agent things. ;- )

maphew:
Maintainer follow-up opened in #4170 to preserve the already-merged functional fix from #4091 while adopting the cleaner call shape raised here.

Specifically: `maybeAutoImportJSONL` no longer accepts a server-mode boolean; `main.go` now gates server mode at the call site before invoking JSONL upgrade recovery. That keeps the embedded-only policy visible at the boundary while retaining the server/shared-server skip behavior.

Credit to @shaunc for identifying and documenting this poor calling shape in this PR.

_codex-gpt-5.5-medium on behalf of CI Bot_

PR #3998, opened 2026-05-16
fix(startup): fail-fast on unsupported backend in metadata.json

quad341:
## What this changes

When a `bd` binary that predates Postgres support (e.g. bd 1.0.3, built before the PG backend landed) is pointed at a workspace whose `metadata.json` declares `backend=postgres`, the old binary silently falls back to a 0.8–1.8 GB RSS, 30–60 second hang before returning "no issue found" with no explanation. Diagnosing this costs hours, because there is nothing in the output that points to a stale binary.

This change adds `checkMetadataBackendError()`, called in `PersistentPreRun` right after `configfile.Load`. It reads the raw `cfg.Backend` field (important: `GetBackend()` hardcodes `"dolt"`, so the raw value is the only reliable signal). If the value is anything other than `""` or `"dolt"`, the binary exits immediately with a clear error message that names the unsupported backend, the binary path, the version/commit, and an upgrade hint. No embedded Dolt process is started and no network I/O is performed, so the exit completes in < 100 ms at < 100 MB RSS.

## Review notes

- The key correctness point: `GetBackend()` normalizes to `"dolt"` for empty/default, so any PG workspace would have returned `"dolt"` from the public accessor. The raw `cfg.Backend` field is the right read target.
- Call site is early in `PersistentPreRun`, before any store-open attempt — intentionally before any Dolt or network work.
- No new config knobs or flags. The behavior activates automatically when `metadata.json` has an unsupported backend value.

## Test plan

- [x] `TestCheckMetadataBackendError_UnsupportedBackend`: postgres backend → exits with backend name + upgrade hint + version
- [x] `TestCheckMetadataBackendError_SupportedBackends`: `""` and `"dolt"` → no error (no regression on default workspace)
- [x] `TestCheckMetadataBackendError_NilConfig`: nil config → no error (no panic on first-run path)
- [x] Release gate: [`release-gates/be-y0sm9s-gate.md`](release-gates/be-y0sm9s-gate.md)

🤖 Deployed by actual-factory

### DIFF
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -1206,7 +1206,7 @@ func checkMetadataBackendError(cfg *configfile.Config, beadsDir string) error {
 			"  binary:  %s\n"+
 			"  version: %s\n\n"+
 			"Upgrade bd to a version that supports %q:\n"+
-			"  curl -fsSL https://raw.githubusercontent.com/steveyegge/beads/main/scripts/install.sh | bash\n"+
+			"  curl -fsSL https://raw.githubusercontent.com/gastownhall/beads/main/scripts/install.sh | bash\n"+
 			"  brew upgrade beads\n\n"+
 			"To confirm which binary is running: bd --version\n"+
 			"To inspect PATH for stale binaries: which bd",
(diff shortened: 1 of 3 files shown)


quad341:
Closing this as obsolete.

The premise here was a stale `bd` binary (built before Postgres support landed) silently hanging — 30–60s, ~1 GB RSS — when pointed at a workspace whose `metadata.json` declared `backend=postgres`. Since this PR was opened, the Postgres backend has been scrubbed from the codebase, so that configuration no longer arises in our workflow and the fail-fast guard is no longer warranted.

Closing rather than carrying it forward. Thanks!

PR #4005, opened 2026-05-17
fix(schema): serialize Dolt schema migrations

julianknutsen:
## Summary

This PR serializes Dolt schema migration execution per database by acquiring a database-scoped advisory lock before calling `schema.MigrateUp`.

The fix is intentionally small:

- Keep `initSchemaOnDB` pinned to one SQL connection.
- Read the active database name with `SELECT DATABASE()`.
- Acquire `GET_LOCK('bd_schema_init:<database>', 30)` on that same connection.
- Run `schema.MigrateUp` while the lock is held.
- Release the lock with `RELEASE_LOCK(...)` on function exit.

This does not serialize normal Beads reads/writes. It only serializes schema initialization/migration work for one Dolt database at a time.

## RCA

While validating the local Beads perf stack, `./scripts/test.sh ./internal/storage/dolt` failed locally even though CI was green.

The first deterministic failure was:

```text
--- FAIL: TestConcurrentInitSchema
concurrent init error: goroutine N: initSchemaOnDB: schema migration: migration 0023_add_no_history_column.up.sql: Error 1105 (HY000): Column "no_history" already exists
```

I reproduced the same failure on pristine `origin/main` with:

```bash
./scripts/test.sh -run '^TestConcurrentInitSchema$' ./internal/storage/dolt
```

So this is not introduced by the current perf/index work. It is already present on `main` when the Dolt server-mode testcontainer tests actually run locally.

## Why CI Did Not Catch This

CI is green because the normal CI test job does not run these server-mode Dolt container tests in the same way this machine does.

Local conditions:

- Docker is available.
- `dolthub/dolt-sql-server:1.88.1` is cached locally.
- `internal/storage/dolt` therefore starts the Dolt testcontainer and runs `TestConcurrentInitSchema`.

CI conditions:

- The normal test job installs the `dolt` CLI, but does not pull/cache `dolthub/dolt-sql-server:1.88.1` for this package path.
- `EnsureDoltContainerForTestMain` treats the missing image as a skip condition.
- The server-mode concurrency test does not actually exercise the failure in that CI lane.

This is a CI coverage gap, not evidence that the local failure is spurious.

## Historical Context

There was a prior advisory-lock fix in `f73a9ca34` for GH#2672. That commit described the problem as concurrent fresh schema init causing Dolt journal/CRC corruption.

That lock was removed by `910ad1310` with the subject:

```text
/internal/{storage,doltserver}: remove incorrect locking attempt for journal corruption that does not repro
```

That removal deleted both the lock and the then-current concurrency test.

However, the test was later reintroduced by `52319c1ff` (`/internal/storage/dolt: fix test and migration`) and subsequently updated by the shared schema migration work.

The current failure is different from the old journal-corruption claim. It is a plain concurrent DDL race:

1. Multiple processes/goroutines enter schema migration on the same fresh database.
2. Each evaluates migration preconditions before another connection finishes its DDL.
3. One connection adds `issues.no_history`.
4. Other connections then run their stale prepared `ALTER TABLE issues ADD COLUMN no_history ...` and fail with `Column "no_history" already exists`.

The old RCA may have been wrong, but the current migration path is still not concurrency-safe.

## Why This Fix

We could try to make every migration script perfectly idempotent under concurrent execution, but that is brittle:

- Existing migrations use check-then-DDL patterns through `INFORMATION_SCHEMA` and prepared statements.
- Check-then-DDL is not atomic across concurrent sessions.
- Future migrations can easily reintroduce the same race.
- Schema migration is naturally a singleton operation per database.

A database-scoped advisory lock makes the invariant explicit: only one session may apply schema migrations to a given database at once.

The lock name includes the database name so unrelated databases on the same Dolt server do not block each other.

## Validation

The failing reproducer now passes on this branch:

```bash
./scripts/test.sh -run '^TestConcurrentInitSchema$' ./internal/storage/dolt
```

Result:

```text
ok  github.com/steveyegge/beads/internal/storage/dolt  5.923s
```

Additional checks run:

```bash
./scripts/test.sh ./internal/storage/schema
go build -tags gms_pure_go ./cmd/bd
```

Results:

```text
?   github.com/steveyegge/beads/internal/storage/schema [no test files]
```

`go build -tags gms_pure_go ./cmd/bd` completed successfully.

## Follow-Up

The CI/local mismatch should be addressed separately. The normal CI test lane currently lets server-mode Dolt container coverage skip when the Docker image is absent. That made `main` look green while a locally cached testcontainer image exposed a real failure.

### DIFF
--- a/internal/storage/dolt/initschema_concurrent_test.go
+++ b/internal/storage/dolt/initschema_concurrent_test.go
@@ -5,6 +5,8 @@ import (
 	"database/sql"
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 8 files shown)


quad341 (approved):
Thanks @julianknutsen — the diagnosis is solid and the advisory lock approach is exactly right for this class of concurrent init race. The `TestConcurrentInitSchema` reproducer is great.

One small hardening item before merge: the deferred `RELEASE_LOCK` call uses the caller's `ctx`. If that context is canceled before the defer runs (e.g., server shutdown while a migration is in flight), `ExecContext` returns immediately without sending the query to the server, leaving the session-scoped advisory lock held until the connection is TCP-closed by the pool.

Fix is one line — wrap the cleanup in a short bounded context:

```go
defer func() {
    cleanupCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
    defer cancel()
    _, _ = conn.ExecContext(cleanupCtx, "SELECT RELEASE_LOCK(?)", lockName)
}()
```

Both `context` and `time` are already imported, so no new imports needed. A maintainer will apply this before merge — no action needed from you.

**Maintainer note:** Apply the one-liner above to `internal/storage/dolt/store.go` in the deferred RELEASE_LOCK block, verify with `go build ./internal/storage/dolt/` + `go test ./internal/storage/dolt/ -run TestConcurrentInitSchema`, then merge.


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`


coffeegoddd (changes requested):
What is a migration script, and why would it be executed concurrently? Beads is responsible for running its own schema migrations. 

I will try to reproduce the schema migration bug (if there is one), but not gonna add more unnecessary locking to this application.

coffeegoddd:
Ok i do see the schema migration failures when they run concurrently, and the advisory lock can work, this is the correct fix i believe: https://github.com/gastownhall/beads/pull/4017

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR serializes server-mode Dolt schema migrations so concurrent fresh database initialization cannot race through `schema_migrations`, which protects repository setup and provider startup from a hard-to-reproduce data/schema consistency failure.

### Original PR Review
Decision: request_changes resolved to approve.

Specific gaps fixed:
- The schema migration lock name is now bounded for every valid database name, so valid long names no longer exceed the Dolt/MySQL named-lock limit.
- Lock cleanup now uses an independent cleanup context, verifies `RELEASE_LOCK`, and discards the connection when release success is uncertain.
- Lock contention now fits the schema-init retry budget, with retryable lock acquisition failures rather than a permanent first-attempt failure.
- The tests now prove observable lock blocking through `SHOW PROCESSLIST`, update the multiprocess schema check to `schema_migrations`, and keep the embedded Dolt boundary explicit.
- The UOW Dolt server path was audited and brought under the same server-mode migration lock behavior.

Review findings addressed:
- Contract & Interface Fidelity (major, high confidence; codex/claude): valid database names could produce over-limit named-lock keys; the final patch uses a bounded lock helper.
- Resource Lifecycle & Cleanup (major, high confidence; all reviewers): canceled contexts could leave a named lock on a pooled session; the final patch releases with a short cleanup context and discards uncertain connections.
- Error Handling & Resilience / Change Impact (major, high confidence; claude/codex/gemini): lock waits could consume the retry budget; the final patch leaves budget for retry after contention.
- Test Evidence Quality (major/minor, high confidence; claude/codex/gemini): earlier tests did not prove Dolt `GET_LOCK` blocking; the final patch observes real waiters and strengthens schema migration coverage.
- Architectural Consistency (minor, medium/high confidence; claude/gemini): the server-mode and embedded-mode migration boundary needed to be explicit; the final patch documents the pinned-connection/server-only contract and preserves embedded behavior.

Remaining non-gating notes:
- No remaining gating issues were found. The final synthesis noted only non-gating residual uncertainty around full-suite coverage outside the focused review evidence and rare future edge cases around acquire cancellation or no-default-database callers.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch:
- `9ed61187e fix(schema): harden migration lock handling (bd-wisp-5pl1m9)`
- Maintainer-fix diff stat: 8 files changed, 506 insertions(+), 29 deletions(-).

### Final Review Status
Ready for the merge queue: final head `9ed61187e` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/26002720118
Cross-Version Smoke Tests: https://github.com/gastownhall/beads/actions/runs/26002720119

### Review Iterations
3 review pass(es) performed. The first two passes requested changes for lock-name bounds, lock release safety, retry behavior, UOW coverage, embedded-mode documentation, and observable lock tests. The third pass approved the final patch after validating the pinned-connection lock, retry budget, embedded boundary, waiter-based tests, and original race premise.

---

julianknutsen:
>Ok i do see the schema migration failures when they run concurrently, and the advisory lock can work, this is the correct >fix i believe: https://github.com/gastownhall/beads/pull/4017

Looking into this and 4017 now

quad341:

PR #4010, opened 2026-05-17
feat(show): be-ijck6q — count-only default + --include-dependents/--include-comments

quad341:
## Summary

- **Default `bd show --json`** now emits `dependent_count`, `dependency_count`, `comment_count` (integer counts via `SELECT count(*)`) instead of loading all dependents/comments into memory. The `dependents` and `comments` slice keys are absent by default.
- **`--include-dependents`**: new flag that streams dependents via `IterDependentsWithMetadata`, shallow-copies each item (ID/Status/IssueType/Priority/Title/DependencyType), and populates `details.Dependents`. Epic progress is computed from the streamed list.
- **`--include-comments`**: new flag that streams comments via `IterIssueComments` and populates `details.Comments`.
- **Count\* storage layer** (`CountDependents`, `CountDependencies`, `CountIssueComments`, `CountIssues`, `CountEvents`) added to `storage.Storage` interface with Dolt/EmbeddedDolt `SELECT count(*)` implementations, postgres stubs returning `errNotImplemented`, and OTel telemetry wrappers. `CountDependentsByStatus` added to `DependencyQueryStore`.
- **`IssueDetails`** gains `DependentCount *int64`, `DependencyCount *int64`, `CommentCount *int64` (all `omitempty`).
- **Regression test** `TestIssueDetailsCountOnlyJSON` asserts: count fields present, `dependents`/`comments` keys absent in default JSON.

## Notes

This PR includes the Iter interface additions from `feat/be-u0zlsq-iter-interface` (be-jaavsb / be-yinl4d) and the Count\* aggregate methods from be-fv0xme, since those hadn't landed on main. The Count\* implementation that landed on a sibling branch couldn't be cherry-picked cleanly due to divergence, so it is included inline here.

## Test plan

- [x] `go build -tags gms_pure_go ./...` — clean
- [x] `go test -tags gms_pure_go ./cmd/bd/... -run "TestShallow|TestIssueDetailsCountOnly"` — all pass
- [x] `go test -tags gms_pure_go ./internal/storage/...` — all pass (pre-existing dolt/wisp_partition_test.go failure unrelated)
- [ ] Embedded Dolt integration tests (CI)

Closes be-ijck6q.


quad341:
Rebased onto current main. The branch had 43 commits with extensive conflicts against main's current state (postgres package removed, schema package refactored, Iter[T] infrastructure not yet in main). Resolved via clean cherry-pick of the two key commits:

1. **Iter[T] infrastructure** (from be-u0zlsq): adds streaming iterator interface, Dolt/embeddeddolt implementations, SliceIter adapter, telemetry wrappers. Dropped postgres/iter_*.go — no postgres package in main yet.
2. **Feature** (be-ijck6q): count-only default for `bd show --json` + `--include-dependents` / `--include-comments` flags.

Also dropped `hook_decorator_storage_only_test.go` — its stub implements `Storage` but `NewHookFiringStore` now requires `DoltStorage`, so the test no longer compiled.

Build: clean. Tests: `TestShallowDependentsForJSON_*` all pass. Remaining failures (`TestCheckHooksDriftNotGitRepo`, `TestTestServerConnection`) are pre-existing on main.

quad341:
Pushed 3 commits addressing the 4 failing CI checks:

1. **Test (ubuntu-latest)** — `internal/jira/tracker_test.go` build failure (`*configStore does not implement storage.Storage (missing method CountDependencies)`). The new `Count*` and `Iter*` methods on `storage.Storage` weren't stubbed on the test's `configStore`. Added no-op stubs (zero counts, empty `SliceIter`s) following the existing pattern for other unused methods. (26b4e8cee)

2. **Test (Embedded Dolt Cmd 1/20)** — `TestEmbeddedShow/show_json_includes_comments` was asserting `comments` populated in default `bd show --json` output, but this PR's whole point is that comments are now count-only by default. Updated the subtest to pass `--include-comments` and exercise the new opt-in flag — which is the behavior the test name was always meant to assert. (3db212ad1)

3. **Check doc flags freshness** — `docs/CLI_REFERENCE.md` and `website/docs/cli-reference/show.md` (+ versioned copies) were out of sync with the new `--include-comments` / `--include-dependents` flags. Regenerated via `./scripts/generate-cli-docs.sh` against a CGO=0 build of the PR branch (matching CI's invocation). Verified locally with `./scripts/check-doc-flags.sh`. (0a19244ca)

**Not addressed:** Test (macos-latest) failed with `curl: (56) ... error: 502` while installing dolt, then `dolt: command not found`. Pre-existing CI infra flake unrelated to this PR — should re-run cleanly with this push.

Ready for another pass.

quad341:
**Code review — beads/reviewer** (factory-reviewed, bead be-iprk)

**Verdict: PASS** — CI all checks green.

`bd show --json` count-only default + streaming iterator flags (`--include-dependents`, `--include-comments`) reviewed against the rebased branch tip.

**What lands:**
- Default `bd show --json` now emits `dependent_count`/`dependency_count`/`comment_count` via O(1) COUNT queries — no more loading full dependency slices on every show call.
- `--include-dependents`: streams via `IterDependentsWithMetadata`; shallow-copies each item to strip heavy fields before serialization.
- `--include-comments`: streams via `IterIssueComments`.
- Storage layer: `CountDependents`, `CountDependencies`, `CountIssueComments`, `CountIssues`, `CountEvents` added to `storage.Storage`; Dolt/embedded implementations; PG stubs returning `errNotImplemented`; OTel telemetry wrappers.

**Two LOW informational items (non-blocking):**
1. `defer iter.Close()` is inside an if/for block — in Go, defer is function-scoped. For a CLI call with bounded issue counts (1–10 typical), N deferred iters all close on `RunE` return, which is safe. Known antipattern documented with `//nolint:errcheck`.
2. `shallowDependentsForJSON` at show.go:445 is exercised by tests but is dead code in the production path (inline copy is used instead). Lint passes; no correctness issue.

Security clean — count queries use parameterized inputs via existing storage layer. Iter paths bounded by daemon session cap (`ErrTooManyIterators` → slice fallback) and idle reap.

Well-handled rebase recovery — cherry-picking the two key commits was the right call given the depth of conflict.

maphew (changes requested):
Found one blocking issue:

- **P1: count/iterator paths drop `wisp_dependencies` relationships**

  `CountDependents`, `CountDependencies`, and `CountDependentsByStatus` count only the permanent `dependencies` table, while the existing relationship APIs explicitly merge both `dependencies` and `wisp_dependencies`. `bd show --json` now trusts these counts for `dependent_count` / `dependency_count`, so any permanent issue with wisp dependents reports too-small counts. The Dolt `IterDependentsWithMetadata` implementation has the same table-only behavior, so `bd show --json --include-dependents` omits wisp dependents even though the legacy `GetDependentsWithMetadata` path includes them.

  Relevant code paths:
  - `internal/storage/dolt/counts.go`: `CountDependents`, `CountDependencies`, `CountDependentsByStatus`
  - `internal/storage/dolt/iter_dependents.go`: `IterDependentsWithMetadata`
  - `cmd/bd/show.go`: JSON output consumes those new count and iterator paths
  - `internal/storage/issueops/dependencies.go`: existing metadata APIs query both dependency tables

Please make the new count and iterator semantics match the existing relationship APIs, or explicitly define and test a JSON contract that excludes wisps. As written, the default JSON fields can disagree with the relationship data already returned by the storage layer.

Local verification note: I attempted the focused Go test run, but this machine failed during test setup with `disk quota exceeded` while writing Go build temp files. CI is green, but I did not find coverage for the mixed `wisp_dependencies` case.

quad341:
Thanks for the careful catch — this was a real correctness gap, and the default JSON could even disagree with itself: `details.Dependencies` is populated from `GetDependenciesWithMetadata` (both tables) while `dependency_count` came from a `dependencies`-only `COUNT(*)`; and in embedded mode `--include-dependents` streams via the slice path (includes wisps) while `dependent_count` did not.

Fixed by making the new count/iter paths match the slice-path semantics — option (a), include wisps — in 9e35080:

1. **Counts dropped wisp edges** (`internal/storage/dolt/counts.go`, `internal/storage/embeddeddolt/counts.go`). `CountDependents`, `CountDependencies`, and `CountDependentsByStatus` now sum both `dependencies` and `wisp_dependencies`; the by-status variant joins each edge table to its home issue table (`dependencies`→`issues`, `wisp_dependencies`→`wisps`), mirroring the existing `queryBlockingInfo` pattern. No dedup needed: `WispTableRouting` writes each edge to exactly one table by source wisp status (and migration 0035 *moves* infra deps rather than copying), so summing equals the slice path's concatenated `len()`.

2. **dolt `IterDependentsWithMetadata`** (`internal/storage/dolt/iter_dependents.go`) now `UNION ALL`s `issues + dependencies` with `wisps + wisp_dependencies`. Because the write table is routed solely by the source's wisp status, every `wisp_dependencies` row has a wisp source (in `wisps`) and every `dependencies` row a permanent source (in `issues`), so each edge table joins cleanly to its source with no ambiguity or double-count. (Embedded already slice-wrapped `GetDependentsWithMetadata`, so it was wisp-correct for `--include-dependents` — only its count was wrong.)

3. **dolt `IterDependenciesWithMetadata`** I switched to delegate to the slice path rather than a streaming join. Unlike the dependents direction, a dependency's *target* may be permanent or wisp independent of which table holds the edge, so a single static join can't resolve it. It has no streaming caller today; left a comment to revisit if one appears.

Tests: regression coverage in both backends asserting a permanent issue with a mixed permanent+wisp dependent set reports the full count, and that the dolt iter streams both. The embedded test (`internal/storage/embeddeddolt/counts_wisp_test.go`) exercises the path `bd show` actually uses — verified locally (fails before, passes after); the dolt-server test (`internal/storage/dolt/counts_wisp_test.go`) covers the streaming UNION and runs in CI.

Re the disk-quota note: the embedded suite needs only `BEADS_TEST_EMBEDDED_DOLT=1 go test ./internal/storage/embeddeddolt/...` (no Docker), if that's easier to run locally.

Ready for another pass.

quad341:
Re-review ready — CI is now fully green on `3e036880f`, including the **Test (Embedded Dolt Storage)** job that exercises the mixed-wisp case.

The P1 fix (option (a): make the new count/iter paths include `wisp_dependencies`, matching `GetDependentsWithMetadata`) landed in `9e35080` as described above. The three follow-up commits did **not** change that behavior — they only reshaped the *query form* to survive a pure-Go GMS analyzer quirk that surfaced only in CI:

- Both `dependencies` and `wisp_dependencies` define `depends_on_id` as a STORED generated column (`COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external)`). Inside a `count(*)` the pure-Go GMS analyzer can prune the base columns the generated column derives from and then fail with `column depends_on_id could not be found in any table in scope` (errno 1105). The slice path projects real columns, so it was unaffected — which is why `GetDependentsWithMetadata` passed but the count didn't.
- Fix: count queries now resolve the dependency target via the base columns directly (`COALESCE(...)`, matching `issueops.DepTargetExpr` on main) instead of the generated `depends_on_id`. Audited: zero `depends_on_id` references remain in any count query. (`558feb8b` / `452f757b` / `3e036880f`.)

Coverage for the exact gap you flagged ("did not find coverage for the mixed `wisp_dependencies` case"):
- `internal/storage/embeddeddolt/counts_wisp_test.go::TestCountsIncludeWispDependencies` — a target with 1 permanent + 1 wisp dependent; asserts `CountDependents == 2`, `CountDependentsByStatus == 2`, and `CountDependencies(wisp) == 1`. Runs in the embedded-storage CI job (red before `9e35080`, green now).
- `internal/storage/dolt/counts_wisp_test.go::TestCountAndIterIncludeWispDependencies` — same fixture, and additionally asserts the dolt `IterDependentsWithMetadata` UNION streams both edge tables.

Ready for another pass.

maphew (approved):
Final pass looks good. The previous P1 is addressed: count paths now include both dependency tables, the Dolt dependent iterator includes wisp dependents, and the mixed permanent+wisp regression tests cover the gap in both backends.

Verified:
- CI is fully green at `3e036880f`.
- `go test -tags gms_pure_go ./cmd/bd -run 'TestShallowDependentsForJSON|TestIssueDetailsCountOnlyJSON' -count=1` passes locally.
- Local Dolt backend test setup is missing the `wisps` table, and local embedded build is missing ICU headers, so I relied on the green CI jobs for those backend-specific regressions.

PR #4017, opened 2026-05-17
/internal/storage/schema/schema.go: advisory lock for concurrent inits

coffeegoddd:

### DIFF
--- a/internal/storage/schema/schema.go
+++ b/internal/storage/schema/schema.go
@@ -91,7 +91,10 @@ func acquireMigrateLock(ctx context.Context, db DBConn) error {
 }
 
 func releaseMigrateLock(ctx context.Context, db DBConn) {
-	_, _ = db.ExecContext(ctx, "SELECT RELEASE_LOCK(?)", migrateLockName)
+	_, err := db.ExecContext(ctx, "SELECT RELEASE_LOCK(?)", migrateLockName)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/storage/schema/schema.go
```diff
@@ -79,6 +79,21 @@ func AllMigrationsSQL() string {
...
+		return fmt.Errorf("acquiring schema migration lock %q: %w", migrateLockName, err)
+	}
+	if !locked.Valid || locked.Int64 != 1 {
+		return fmt.Errorf("acquiring schema migration lock %q: timed out after %ds (another connection holds it)", migrateLockName, migrateLockTimeoutSec)
+	}
+	return nil
+}
+
+func releaseMigrateLock(ctx context.Context, db DBConn) {
+	_, _ = db.ExecContext(ctx, "SELECT RELEASE_LOCK(?)", migrateLockName)
```
quad341: sorry, more familiar with pg than mysql; seems like an `EXEC` not a `SELECT`

also.. if this fails, i would super appreciate a log message as you're likely hosed after its failure if we think we released a lock and _didn't_

quad341 (approved):

internal/storage/schema/schema.go
coffeegoddd: i disagree on the log message, but totally agree on a failure to release the lock should just be a panic. will fix

quad341:

PR #4020, opened 2026-05-17
feat(prune): reference-aware bd prune — skip cited closed beads by default (be-jewoem)

quad341:
## What this changes

`bd prune` now skips closed beads that are still cited in the description, notes, or comments of any open bead. Before this change, `bd prune` would delete any closed bead that matched the age/label filters, even if another open bead's body referenced its ID as an ADR, decision trail, or verification link.

After this change, when a closed bead's ID appears as a word-boundary match in open bead content, it is silently protected and counted in the output:

```
○ Pruned 3 closed beads  (8 closed beads protected by open-bead references
                           — use --ignore-references to override)
```

The escape hatch `bd prune --ignore-references` restores the old behavior and deletes referenced beads too. `bd purge` is unaffected and does not accept `--ignore-references`.

A 10K-bead benchmark (`TestPruneLargeFixture`) is included: `buildReferencedSet` completes in under 1 second on a fixture with 10K open beads at 5 KB each, well under the 5-second NFR gate.

## Review notes

- `--ignore-references` is registered only on `pruneCmd`, not `purgeCmd` — Cobra rejects `bd purge --ignore-references` with "unknown flag"
- `referenced_skipped`, `referenced_count` always present in JSON; `referenced_ids_sample` omitted when 0
- `buildReferencedSet` uses word-boundary regex (`\bID\b`) to avoid partial-ID false positives
- Follow-up be-g76wkv (P3): regression test for `bd purge --ignore-references` rejection — behavioral guarantee is in place, test not yet written
- Pre-existing: duplicate `commandDidWrite.Store(true)` at purge.go:314+348 — harmless, not introduced here

## Test plan

- [x] Integration tests T1–T7 in `cmd/bd/prune_refs_embedded_test.go`: referenced-skip, ignore-references override, closed-to-closed refs, word-boundary precision, empty set, body/comment/notes citations
- [x] Unit test `TestBuildReferencedSet_EmptyCandidates` in `prune_refs_test.go`
- [x] `TestPruneLargeFixture` passes in 0.63–0.87s (<5s NFR gate)
- [x] `go test -short ./cmd/bd/` zero regressions (bench guard skips large fixture)
- [x] `golangci-lint ./cmd/bd/` 0 issues
- [x] Release gate: [`release-gates/be-jewoem-gate.md`](release-gates/be-jewoem-gate.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,21 @@ bd backup restore --force /path/to/backup
 See [docs/DOLT.md](docs/DOLT.md#migrating-between-backends) for full
 migration instructions.
 
+## 🗑 Storage Maintenance
+
+Use `bd prune` to delete old closed beads and reclaim space, and `bd purge` to remove closed ephemeral beads (wisps).
+
+```bash
+bd prune --older-than 30d              # Preview closed beads older than 30 days
+bd prune --older-than 30d --force      # Delete them
+bd prune --older-than 90d --dry-run    # Detailed preview with stats
+bd purge --force                       # Delete all closed ephemeral beads
+```
+
+**Reference protection:** `bd prune` automatically skips closed beads whose ID appears in the description, notes, or comments of any open or in-progress bead. This prevents accidental deletion of ADR records, decision trails, and verification history that downstream beads still reference.
+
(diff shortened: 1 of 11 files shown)


quad341:
Superseded by #4023 (feat/be-jewoem-be-u2mw2x-reference-aware-prune) which includes the same reference-aware prune logic plus bd-umbf contributor namespace isolation (Children 1-3).

PR #4021, opened 2026-05-17
perf: Q1+Q2+Q3+Q5 — SQL count aggregate, skip-wisps merge, id-shrunk Pattern B, wisp cache (bd-xoi0x, bd-qi4qr, bd-co6m6, bd-wsgws)

quad341:
## Summary

- **Q1 (bd-xoi0x)**: `CountIssuesByGroup` added to Storage interface; `bd count` routes through `CountIssues(ctx, query, filter)` (main's existing int64 signature, query="") for the total and `CountIssuesByGroup` for per-group breakdown. Group-by-label uses a subquery to avoid joinIter panic. Includes `InstrumentedStorage` pass-through.
- **Q2+Q5 (bd-qi4qr)**: `IssueFilter.SkipWisps` opt-out flag; `count`/`query`/`list` commands set it to skip the wisps merge scan. Q5 per-process cache (`wispCountCached`/`wispCountZero`) on `DoltStore` short-circuits the wisps merge when the table is known empty; invalidated on any wisp write.
- **Q3 (bd-co6m6)**: `IssueFilter.NoIDShrink` flag; `searchTableInTx` dispatches Pattern B (`SELECT id LIMIT n` → batch hydrate) for bounded queries, Pattern A retained for unlimited.
- **bd-wsgws spike**: `TestJoinIterPanic_Spike` covers 5 JOIN/EXISTS shapes with panic recovery; each sub-test skips if the panic no longer fires (signals workaround is liftable). Requires a live Dolt server to produce results.

> **Rebase note (2026-05-24):** Rebased onto main. `CountIssues(ctx, filter) → int` from the original Q1 branch was removed (main already has `CountIssues(ctx, query, filter) → int64`). The two `bd count` call sites now pass `query=""` and handle `int64`. `CountIssuesByGroup` has no naming conflict and is preserved as a new interface method. The `computeBlockedIDs` cache from Q2+Q5 was dropped — main now uses the `is_blocked` column; `invalidateBlockedIDsCache()` is retained as a no-op cache clearer at write sites.

## Test plan

- [ ] `go test ./internal/storage/issueops/... ./internal/jira/... ./cmd/bd/...` passes
- [ ] `bd count`, `bd count --by-status`, `bd count --by-label` produce correct output on a real rig
- [ ] `bd query "status=open"` and `bd list` respect Q2 skip-wisps optimization
- [ ] Run `TestJoinIterPanic_Spike` with a live Dolt server; record which shapes skip vs error in bd-wsgws notes

[...]

PR #4027, opened 2026-05-18
fix(schema): migrate TEXT → LONGTEXT for large-content columns (be-knye)

quad341:
## Summary

- `issues` + `wisps`: `description`, `design`, `acceptance_criteria`, `notes`, `close_reason` changed from `TEXT` (64 KB limit) to `LONGTEXT`
- `comments.text` changed from `TEXT` to `LONGTEXT`
- Each `MODIFY COLUMN` is guarded by an `INFORMATION_SCHEMA` check so the migration is idempotent on databases already carrying `LONGTEXT` columns

## Root cause

`TEXT` in MySQL/Dolt is capped at 65535 bytes. Bead descriptions with embedded base64 PNG images (e.g. `be-mrzg8o`) routinely exceed this, causing `MySQL Error 1105: Field 'description' is too long` during `bd import` from JSONL.

## Test plan

- [ ] `go test ./internal/storage/schema/... ./internal/storage/embeddeddolt/... ./internal/storage/dolt/...` passes (Dolt integration tests require a live Dolt server)
- [ ] `BEADS_TEST_EMBEDDED_DOLT=1 go test ./internal/storage/embeddeddolt/... -run TestSchemaAfterInit` verifies new migration applies cleanly
- [ ] `bd import` on a JSONL containing a bead with >65535 byte description succeeds without Error 1105
- [ ] Pre-existing `TestFederationExcludeTypesOptOut` failure in `internal/config` is unrelated to this change


quad341:
Investigated the failing **Test (storage domain + uow)** check and pushed a fix (`fc3452436`). The branch had **two** defects in the LONGTEXT migration:

**1. CI failure — `DEFAULT ''` was dropped on the `wisps` columns.**
`MODIFY COLUMN … LONGTEXT NOT NULL` replaces the entire column definition, so it silently dropped the `DEFAULT ''` that `wisps.description/design/acceptance_criteria/notes` were created with in `0020_create_wisps`. Any insert that omits those columns — e.g. the domain suite's `seedWispRow`, which inserts `id`+`title` only — then fails with:

```
Error 1105 (HY000): Field 'description' doesn't have a default value
```

That's exactly the failure, and it's why every failing subtest is a `Wisp/*` one (the `issues` columns never had a default, so `seedIssueRow` supplies them explicitly). Fix: restate `DEFAULT ''` on the four `wisps` columns and preserve the nullable `DEFAULT ''` on `close_reason` for both tables. `issues` big-text columns and `comments.text` were created `NOT NULL` with no default (`0001`/`0004`) and intentionally keep none.

**2. Latent (worse) bug — duplicate migration number `0043`.**
The migration was numbered `0043`, but `0043_drop_dependencies_generated_column` and `0044`/`0045` already exist on `main`. The runner applies a migration only when its version `>` `MAX(schema_migrations.version)` (`schema.go` `migrate()`), so on any database already at v43+ (i.e. everyone on current `main`) the LONGTEXT widening would have been **silently skipped forever** — a no-op fix for the very overflow it targets. Renumbered to `0046`.

**Verification** — ran the real `schema.MigrateUp` against a live `dolt sql-server` (no Docker here for the testcontainer suite):
- all six target columns become `longtext`; `schema_migrations` max == `LatestVersion()` (46)
- `wisps` defaults preserved (`desc/design/ac/notes = ""`); minimal `wisps (id,title)` insert succeeds
- minimal `issues (id,title)` insert still correctly rejected (no default by design)
- second `MigrateUp` is a clean no-op (idempotent)

CI should re-run on the new commit.

**One note for the maintainer (not a blocker):** this branch bundles the schema migration with several unrelated commits (`perf(count)`, `perf(search)`, `fix(jira)` mock stubs, a `joinIter` spike test — bd-xoi0x/qi4qr/co6m6/wsgws). Splitting those into their own PR(s) would make this one reviewable as a focused schema change, but that's your call — I left history untouched.

quad341:
**Code review — fix(schema): migrate TEXT → LONGTEXT for large-content columns (be-knye)**

The schema migration is well-designed:

- **Correct columns**: `description`, `design`, `acceptance_criteria`, `notes`, `close_reason` on `issues`/`wisps`, and `comments.text` — all the fields that could hold large agent outputs or embedded base64 images
- **Idempotent**: each `MODIFY COLUMN` is guarded by an `INFORMATION_SCHEMA` check so running the migration on an already-upgraded database is a no-op
- **DEFAULTs preserved**: the comment correctly identifies and preserves the original `NOT NULL DEFAULT ''` and nullable `DEFAULT ''` patterns from the original migrations (0001/0020). Dropping these would regress inserts that omit the column.
- **Down migration**: correctly notes that reverting LONGTEXT→TEXT after storing >64KB content would silently truncate; the down migration is a deliberate no-op with an explanation
- Sequential migration number (0046) fits the existing numbering scheme

All CI passes. (Note: this PR stacks on the perf work from #4021 — the diff relative to main includes those changes. The actual schema migration delta is in `internal/storage/schema/migrations/0046_*`.)

---

PR #4029, opened 2026-05-18
fix(import): sync issue_prefix from config.yaml after import (be-llaf)

quad341:
## What this changes

\`bd import\` now syncs \`issue_prefix\` from \`config.yaml\` to the Dolt config table when they differ after an import completes. This closes the gap where a PG→Dolt migration would leave the DB with a stale prefix (e.g., \`bd\` instead of \`be-\`), causing all new issues created post-migration to receive the wrong ID prefix.

The sync is idempotent: it runs only if \`config.yaml\` has a non-empty \`issue-prefix\` and only if the DB value differs. It uses \`CommitWithConfig\` rather than \`Commit\` — which intentionally bypasses the GH#2455 gate that otherwise skips the config table.

## Review notes

- \`CommitWithConfig\` is required here, not \`Commit\`. Error is silently discarded (best-effort); worst case is that the sync is delayed until the next explicit \`bd config set\`.
- Logic is placed after the issues commit in \`runImportFromReader\` so it only runs on successful import.
- \`TestEmbeddedImport/prefix_sync\` covers the stale-prefix scenario end-to-end.

## Test plan

- [x] \`go test ./...\` — all shards PASS (CI run 26114120267)
- [x] Lint, formatting, build-tag policy, doc-flags freshness — all PASS
- [x] \`TestEmbeddedImport/prefix_sync\`: init with \`--prefix bd\`, overwrite \`config.yaml\` with \`issue-prefix: be\`, import a JSONL; DB \`issue_prefix\` becomes \`be\`
- [x] Release gate: [\`release-gates/be-llaf-gate.md\`](release-gates/be-llaf-gate.md)

🤖 Deployed by actual-factory


quad341:
## Review: pass ✓

Clean, focused fix — exactly the right scope for this bug.

- **Comment explains the non-obvious WHY correctly**: the `CommitWithConfig` vs `Commit` distinction (GH#2455) would have been confusing without the inline note — good call.
- **Guard `yamlPrefix != ""`** prevents accidental erasure on systems without the YAML key — correct.
- **Silent CommitWithConfig failure**: documented as best-effort in the PR, acceptable.
- All CI green across all platforms and upgrade smokes.

Ready to deploy.

---
Reviewer: beads/reviewer (gc-reviewer-e6bb317e246c)

quad341:
**Code review — fix(import): sync issue_prefix from config.yaml after import (be-llaf)**

The fix is correct and the approach is well-reasoned. After `bd import`, the database `issue_prefix` can fall out of sync with `config.yaml` because `store.Commit` intentionally skips the config table (GH#2455). The fix uses `CommitWithConfig` specifically for this config update — correct use of the appropriate API.

The condition is conservative: only updates if `config.yaml` has a non-empty `issue-prefix` AND it differs from the DB value. The silently-ignored errors (`if setErr := ...; setErr == nil`) on the subsequent commit are reasonable — the data import succeeded, this is a best-effort config sync, and failing it shouldn't roll back the import.

The embedded test (`prefix_sync` subtest) simulates the exact failure scenario (stale DB prefix vs. config.yaml prefix) and verifies the fix end-to-end.

All CI passes.

---

maphew:

quad341:
Thanks @maphew — here's a concrete, self-contained reproduction. It reproduces on current `main` (verified against `3fe20e709`, `bd 1.1.0`) and this PR fixes it.

**Environment**
- `bd version`: bug reproduces on `main` @ `3fe20e709` (bd 1.1.0); fixed on this branch (now rebased onto `main`).
- Storage mode: embedded Dolt (the default — single machine, no `sql-server`).
- OS: Linux x86_64. The logic is OS-independent (pure-Go prefix resolution), so it should reproduce anywhere.

**Minimal repro** (scratch repo — touches nothing else)
```bash
cd "$(mktemp -d)" && git init -q .

bd init --quiet --prefix bd          # DB issue_prefix = bd.
                                     # NB: init stores the prefix in the DB; it does
                                     # NOT write issue-prefix into config.yaml.

# Simulate a PG->Dolt migration that sets the prefix in config.yaml while the
# migrated DB config table keeps the old value:
printf 'issue-prefix: be\n' > .beads/config.yaml

# Import (the migration step that a user expects to reconcile the prefix):
printf '%s\n' '{"id":"bd-1","title":"x","status":"open","issue_type":"task","created_at":"2026-01-01T00:00:00Z","updated_at":"2026-01-01T00:00:00Z"}' > issues.jsonl
bd import issues.jsonl

bd config get issue_prefix           # DB value after import
bd create "post-migration issue"     # prefix of a brand-new issue
```

**Expected vs. actual**

|  | `bd config get issue_prefix` after import | new issue id |
|---|---|---|
| Expected (config.yaml says `be`) | `be` | `be-…` |
| **Actual on `main`** (the bug) | `bd` | `bd-…` |
| With this PR | `be` | `be-…` |

**Why it happens**

New top-level IDs are minted from the **DB config table**, not from `config.yaml`: `internal/storage/domain/issue.go` `resolveTopLevelPrefix()` reads `issue_prefix` via `GetConfig`. `store.Commit` intentionally skips the config table (GH#2455), so after a migration that updates `config.yaml` but not the DB, nothing reconciles the two and every new issue keeps the stale prefix. Editing `config.yaml` alone doesn't help, and `bd config set issue_prefix …` is intentionally rejected (it points you at `bd init` / `bd bootstrap` / `bd rename-prefix`) — so `bd import` is the natural reconciliation point, which is what this PR adds (`SetConfig` + `CommitWithConfig` when the values differ).

**Rebase note**

The branch was ~947 commits behind `main` and conflicted in `import.go` (the import path was refactored into bounded transactions), so I rebased it onto current `main`, keeping the original three commits. After the rebase I re-verified:
- the repro above now prints `be` / `be-…`;
- `BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go -run 'TestEmbeddedImport/prefix_sync' ./cmd/bd/` → PASS.

The PR now shows `MERGEABLE`. Ready for another look.

maphew (commented):
## Review (claude-opus, adjudicated): approve after two hygiene fixes

Verified at head `a8784aebf` against current main:

- The sync writes exactly the key new-ID minting reads (`resolveTopLevelPrefix` → `GetConfig("issue_prefix")`) — this cures the real "new issues get the stale prefix" symptom, not a cosmetic value.
- `CommitWithConfig` (not `Commit`) is the correct API for a config-table change, and the inline comment explains why.
- Clobber-safe: absent/empty `issue-prefix` in config.yaml skips the block entirely, and it's a no-op when already in sync; dry-run returns before the block; best-effort error handling can't roll back a successful import.
- Not stale despite the dormancy: `cmd/bd/import.go` is byte-identical between the PR's merge-base and current main; zero-conflict merge.
- The `prefix_sync` subtest is genuinely end-to-end (real `bd` subprocess reading on-disk config.yaml, re-opened store asserting the persisted value). Codecov's 0% patch claim is a false alarm — the covering test is env-gated out of the coverage job.

Before merge:
1. Drop `release-gates/be-llaf-gate.md` — contributor-process artifact introducing a new top-level directory upstream doesn't have, and stale across the July rebase (cites pre-rebase SHAs, old line numbers, an old CI run). Same treatment as the sibling PRs in this sweep.
2. Add a one-line comment at the sync block stating the assumption: config.yaml is authoritative and existing issue IDs are intentionally left unchanged (this block deliberately bypasses the `bd config set issue_prefix` guard for the import/migration flow; a future reader shouldn't mistake it for a full rename).

Nits, no action required: an ambient `BD_ISSUE_PREFIX` env var can also drive the sync (consistent with bd's env precedence elsewhere); a negative subtest pinning the no-config-key no-clobber path would be welcome but isn't a gate.

A worker will push the two items shortly; merge follows.

_claude-code-fable-5 adjudicating claude-opus-4-8, on behalf of matt wilkie_

maphew:
Pushed the two hygiene fixes from the adjudicated review, plus the welcome-but-optional negative subtest: dropped `release-gates/be-llaf-gate.md` (stale contributor-process artifact from before the July rebase); added a one-line comment on the `issue_prefix` sync block in `cmd/bd/import.go` stating that `config.yaml` is authoritative and existing issue IDs are intentionally left unchanged (the block deliberately bypasses the `bd config set issue_prefix` guard for the import/migration flow and is not a rename); and added `prefix_sync_no_config_key` beside the existing `prefix_sync` subtest in `cmd/bd/import_embedded_test.go`, confirming that a `config.yaml` with no `issue-prefix` key leaves the DB's `issue_prefix` untouched. `go build -tags gms_pure_go ./cmd/bd/...` and `BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go -count=1 -run 'TestEmbeddedImport' ./cmd/bd/...` both pass, including the new negative subtest. Ready to merge from my side.

PR #2292, opened 2026-03-03
fix(dep): enforce same-type blocking validation

l0g1x:
## Summary
- Adds validation in `AddDependency` (storage layer) and CLI (`bd dep add`, `bd dep --blocks`) to prevent `blocks` dependencies between tasks and epics
- Tasks can only block other tasks; epics can only block other epics
- Fixes deadlock in `bd ready` where epic blocked-by-children via redundant `blocks` deps caused GH#1495 logic to exclude all children from ready work

## Test plan
- [ ] `go build ./...` compiles
- [ ] New tests in `dependencies_extended_test.go`: task→epic blocked, epic→task blocked, task→task allowed, epic→epic allowed, parent-child cross-type allowed
- [ ] `bd dep add <task> <epic> --type=blocks` returns error
- [ ] `bd dep add <task1> <task2> --type=blocks` succeeds

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1495, opened 2026-02-04
Epic status calculation: completed epics show BLOCKED with "0 dependencies", child tasks appear in bd ready


bradleyfay:
# Feature Request: Improve Epic Status and `bd ready` Filtering

## Problem Statement

Currently, `bd blocked` lists epics in confusing ways, and `bd ready` shows tasks as available even when their parent epics are blocked. This creates workflow confusion and makes it hard to understand what work is actually ready to start.

### Issue 1: Completed Epic Still Shows as BLOCKED

Epic `ralph-5rs7` has all 9 child tasks completed (✓):

```
● ralph-5rs7 [EPIC] · Phase 2a: Haiku Experimentation   [● P1 · BLOCKED]

CHILDREN
  ↳ ✓ ralph-5rs7.1: Create ModelRoutingConfig data model for experiment mode ● P1
  ↳ ✓ ralph-5rs7.2: Integrate ModelRoutingConfig into RalphConfig ● P1
  ↳ ✓ ralph-5rs7.3: Implement random model selection logic for experiment mode ● P1
  ↳ ✓ ralph-5rs7.4: Integrate model selection into loop.py task execution ● P1
  ↳ ✓ ralph-5rs7.5: Add model selection metadata to MLflow traces ● P1
  ↳ ✓ ralph-5rs7.6: Create analysis queries for Haiku experiment evaluation ● P1
  ↳ ✓ ralph-5rs7.7: Add success metrics tracking to task execution traces ● P1
  ↳ ✓ ralph-5rs7.8: Write tests for experiment mode model routing ● P1
  ↳ ✓ ralph-5rs7.9: Document Haiku experimentation setup and usage ● P1

BLOCKS
  ← ● ralph-n9ua: (EPIC) Phase 2b: Classification-Based Routing ● P1
```

**Expected**: Epic should show as complete or ready-to-close, not BLOCKED

### Issue 2: Epic Shows as BLOCKED with "0 open dependencies"

Epic `ralph-xt3q` appears in `bd blocked` output, but when inspected shows:

```
● ralph-xt3q [EPIC] · Phase 3: Enhanced Instrumentation   [● P1 · BLOCKED]
...
Blocked by 0 open dependencies: []
```

**Expected**: If an epic has 0 open dependencies, it shouldn't appear in `bd blocked`

### Issue 3: Child Tasks Show as Ready When Parent Epic is Blocked

[...]

PR #4034, opened 2026-05-18
fix(dep): narrow same-type blocking rule to parent-child shadows

Ethee:
## Summary

Replaces the GH#1495 / #2292 ban on **all** cross-type `blocks` edges with a narrower rule: a `blocks` edge is rejected only when its endpoints already share a parent-child ancestor/descendant relationship.

After this change, a task can block an epic (and an epic can block a task) when the two are not in a parent-child hierarchy — restoring a workflow that several users have asked for, while still preventing the specific livelock the original PR was guarding against.

## Background

PR #2292 (commit `ff1a409a5`) added a same-type check to `AddDependencyInTx`:

```go
if dep.Type == types.DepBlocks && targetType != "" {
    sourceIsEpic := sourceType == string(types.TypeEpic)
    targetIsEpic := targetType == string(types.TypeEpic)
    if sourceIsEpic != targetIsEpic {
        // reject
    }
}
```

That PR's commit message and PR description both explain the motivation:

> Fixes deadlock in `bd ready` where epic blocked-by-children via redundant `blocks` deps caused GH#1495 logic to exclude all children from ready work.

The specific failure mode it guarded against:

1. Task `T` is a child of epic `E` (parent-child edge).
2. User redundantly adds a `blocks` edge `T → E` ("T blocks E").
3. `bd ready`'s blocked-set computation marks `E` blocked because `T` is open.
4. Per `blocked.go:675-684` (also GH#1495), children of a blocked parent are also marked blocked, so `T` gets added to the blocked set.
5. Now `T` is blocked but is also the only path to unblock `E`. `bd ready` shows nothing.

The fix sized to the bug was correct but broad: it banned **every** task↔epic `blocks` edge, including ones between unrelated issues that would never trigger the livelock.

## Why widen it now

User reports (one inline, others in the wild — see the discussion at the top of this PR's branch) describe a legitimate workflow that the ban prevents:

> Sometimes an entire epic IS blocked by a single task, and maybe that task is part of a different epic, and so on. The inability to block epics with tasks breaks my mental model.

The example: an external dependency (e.g., "legal review (task)") gates a feature epic. The task is not a child of the epic; it just happens to be the next thing the epic is waiting on. The livelock from #1495 cannot occur in this configuration, because the gating task isn't reachable from the epic via parent-child propagation.

## What the new rule does

`AddDependencyInTx` now runs a recursive CTE (`parentChildLinkedQuery`) when adding a `blocks` edge. The CTE walks the parent-child relation in both directions from the source and counts whether the target lies in the same connected component. If so, the edge is rejected with:

> cannot add blocks dependency: `<source>` and `<target>` already share a parent-child relationship; the existing hierarchy encodes their scheduling

Concretely, the new rule rejects:

- A child task blocking its parent epic (the exact #1495 case).
- A parent epic blocking its own child task.
- A grandparent epic blocking a grandchild task (transitive shadow).
- Any pair where one is an ancestor of the other via parent-child edges.

And it allows:

- Tasks blocking unrelated epics.
- Epics blocking unrelated tasks.
- All same-type combinations (unchanged from before).
- Multi-level cross-type chains like `T1 blocks E1`, `T1 child-of E2`, `T2 blocks E2`, `T2 child-of E3`, etc.

## Implementation notes

- The validation lives in `internal/storage/issueops/dependencies.go`, the same single authoritative gate the original rule used. All callers (CLI, RPC, MCP) inherit the new behavior automatically.
- The `parentChildLinkedQuery` mirrors the structure of the existing `cycleReachabilityQuery`, including the single-vs-multi-table optimization and the MySQL/Dolt-conformant single-recursive-term CTE. Parent-child rows are emitted in both directions in a derived edge relation so the recursive term stays simple.
- The self-dependency check is moved above the new shadow check so the CTE's seed-equals-target degenerate case is impossible to reach.

## Known limitation (addressed in a stacked follow-up)

A longer chain of `blocks` and `parent-child` edges can still close a loop — e.g.:

```
E1 ←blocks← T1
T1 child-of E2
E2 ←blocks← T0
T0 child-of E1
```

Every individual edge involves an unrelated pair, so the new shadow rule allows each insertion in isolation. The combined graph closes a loop only at the last edge, which existing cycle detection (blocks-only) does not see.

A follow-up PR stacked on this one extends cycle detection to walk the combined graph and rejects the closing edge. It also adds combined-graph cycle tests. Together, the two PRs preserve the original #1495 safety invariant ("`bd ready` cannot deadlock from edge insertions") while opening up the legitimate cross-type workflows.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./internal/storage/...` clean
- [x] `BEADS_TEST_EMBEDDED_DOLT=1 go test ./internal/storage/embeddeddolt/... -run TestAddDependency` — all 16 subtests pass (~7s)
  - New: `cross_type_unrelated_allowed`, `parent_child_shadow_rejected`
  - Flipped: `epic_blocks_unrelated_task_succeeds` (was `epic_blocks_task_fails`)
- [x] `go test ./internal/storage/issueops/...` clean
- [ ] CI: `internal/storage/dolt/dependencies_extended_test.go` — new tests `BlocksCrossType_*_Unrelated`, `Blocks_ChildBlocksParent_Rejected`, `Blocks_ParentBlocksChild_Rejected`, `Blocks_AncestorBlocksDescendant_Rejected`, `Blocks_DeepCrossTypeChain` (requires Dolt testcontainer; couldn't run locally without Docker).

## References

- Original problem report: GH#1495
- Original fix: PR #2292 (commit `ff1a409a5`)


quad341:
Thanks @Ethee — this is the right direction. The PR description is clear and the narrowing approach (reject only when endpoints share a parent-child relationship) is exactly what's needed to preserve the GH#1495 invariant while restoring valid cross-type workflows.

**One medium issue I'll fix on my side before merging:**

The `parentChildLinkedQuery` CTE emits both directions of parent-child edges before entering the recursive walk:

```go
fmt.Sprintf("SELECT issue_id AS src, depends_on_id AS dst FROM %s WHERE type = 'parent-child'", t),   // child → parent
fmt.Sprintf("SELECT depends_on_id AS src, issue_id AS dst FROM %s WHERE type = 'parent-child'", t),   // parent → child
```

That creates an **undirected** connected-component check. The consequence: if Epic E has tasks T1 and T2 as children, T1 and T2 are in the same parent-child component (via E), so a `T1 blocks T2` edge would be rejected — even though T1 and T2 are siblings with no ancestor/descendant relationship. The PR's stated rule is ancestor/descendant only.

I'll fix the query to use directed reachability (seed from source, walk only parent-child edges in one direction; separately seed from target) to check whether one endpoint is an actual ancestor/descendant of the other. I'll also add a test that two sibling tasks under the same parent can still block each other.

The rest is solid: the moved self-dep check, the lockstep test updates, the three new rejection cases (child-blocks-parent, parent-blocks-child, ancestor-blocks-descendant), and the clear disclosure of the residual gap are all good work. Thanks for the careful PR description.

---
*Fellow-contributor review (Claude+Codex ensemble, Qwen unavailable — read access only, maintainer merge needed)*

quad341:
@Ethee — landing this with credit to you. The diagnosis was exactly right: the deadlock requires *parent-child reachability*, so the ban should be scoped to exactly that, and mirroring the existing `cycleReachabilityQuery` was the right instinct. Your writeup — the before/after rule, the explicit allowed/rejected table, and the honest callout of the combined-graph limitation — did a lot of the review for me.

Two small things I took on the maintainer side so you didn't have to spin on a rebase:

1. The recursive walk was undirected, so it rejected any two issues in the same hierarchy *component* — including two sibling tasks under one epic where one just needs to block the other. That's actually narrower than your own stated rule ("all same-type combinations unchanged" / "one is an ancestor of the other"), so I tightened it to directed ancestor/descendant reachability and added sibling- and cousin-allowed regression tests.
2. `main` moved under you: the `depends_on_id` column your query referenced was split into `depends_on_issue_id`/`depends_on_wisp_id`/`depends_on_external` (resolved via `DepTargetExpr`). I updated the edge query to match — same fix the neighboring cycle check already uses — so the shadow check actually runs on Dolt.

Everything you wrote still holds; your existing tests and the deep-chain case all pass. Once this is in we can stack your follow-up to close the combined-graph loop. Thanks for restoring a workflow people have genuinely been missing — really nice work.

Picking this up in #4077.


quad341:
**Code review — @Ethee's fix(dep): narrow same-type blocking rule to parent-child shadows**

The change is well-reasoned and correctly scoped. The original cross-type ban was over-broad: it rejected legitimate workflows like task T blocks epic E when T is completely unrelated to E. The new rule targets only the actual livelock source — blocks edges between issues that already share a parent-child relationship (the blocked parent propagates blocked status to its children, including the blocker, creating a deadlock `bd ready` can never escape).

The implementation:
- Keeps the existing cycle detection CTE (blocks-only graph)
- Adds `parentChildLinkedQuery` for shadow detection on blocks inserts
- Updates tests to reflect the new semantics: unrelated cross-type blocks succeed; parent-child shadow blocks are rejected

This PR's combined-graph cycle detection extension is in #4035, which stacks on top of this. The full safety invariant (no livelock from any edge combination) requires both. #4077 is the maintainer cherry-pick of this PR.

All CI passes.

---

quad341:

maphew:

steveyegge:
Maintainer update: merged current main into this branch and carried the design forward, per the fix-merge policy in PR_MAINTAINER_GUIDELINES.md.

Context: this issue resurfaced as bd-wg7ve (a task could not be gated on an epic closing). A fix nearly identical to this PR's design was mistakenly pushed straight to main as 99496601d, then reverted (ddd4b20b2) both for process (unreviewed direct push) and because this PR got there first — re-landing it here with your attribution intact.

Changes in the maintainer merge commit (4d81aa942):

1. **Directed ancestry walk.** `parentChildLinkedQuery` walked the parent-child component undirected, so any two nodes under a common ancestor matched — including siblings. That would reject ordering edges between tasks in the same epic, which are legitimate and common. Replaced with an ancestor-only walk (child → parent), checked in both directions, with direction-specific errors ("cannot be blocked by its ancestor … the gate would never clear" / "cannot be blocked by its descendant … would inherit the block"). Sibling coverage added to the tests.
2. **Schema currency.** The branch predates the typed dependency target columns; the query now uses `DepTargetExpr` instead of the removed `depends_on_id`.
3. **Guard scope.** `conditional-blocks` is covered too — it previously skipped cross-type validation entirely.

Also updated the conformance audit (added after this branch was cut) and the CHANGELOG. Tested: embedded-dolt + dolt dependency suites, conformance audit on embedded dolt and sqlite, and the bd-wg7ve repro end-to-end (task blocked while epic open, ready once the epic itself closes).

Your original commit is untouched; review of the combined diff is still needed before merge — this PR remains the follow-up gate for #4035's combined-graph cycle detection, which is still wanted on top.

maphew:
Maintainer update: I rebased the substance of this PR onto current `main` with merge commits (preserving Ethee's commits) and pushed the fix-merged result to this branch at `67aecac1b`.

The contributor's directed ancestry rule is retained. The maintainer follow-up closes the storage-path gaps found during review:

- applies the ancestor/descendant guard to classic, proxied/domain, graph, bulk/import, and wisp writes;
- covers both `blocks` and `conditional-blocks`, while retaining sibling, unrelated cross-type, external, and cross-prefix cases;
- processes planned `parent-child` edges first inside transactional multi-edge operations, so block-first input cannot bypass the guard;
- preserves hierarchy-before-cycle error precedence and routes the legacy wisp writer through the shared mutation path.

Two independent review passes (Standards and Spec) are now clean. Local verification passed for the issueops, domain, Dolt, and graph-apply seams, including SQLite-backed proxied parity tests and the pure-Go tagged `cmd/bd` tests/vet. The local machine lacks the ICU headers needed for the CGO embedded-Dolt suite, so the GitHub matrix still needs to exercise that path.

I am not merging while `main` is red. The current base failure is the embedded-Dolt conformance job reaching its 30-minute timeout in [run 29278903172](https://github.com/gastownhall/beads/actions/runs/29278903172), not a failure from this PR. Once base health is restored and this branch's checks are green, this should be ready for final maintainer review. #4035 remains the follow-on for general combined `blocks` + `parent-child` cycle detection; this PR intentionally does not absorb that behavior.

maphew (approved):
Maintainer review complete after fix-merging current `main` into the contributor branch.

The accepted behavior is now the narrow hierarchy guard the original report called for: task/epic blocking is allowed generally, while `blocks` and `conditional-blocks` are rejected only between an issue and its own ancestor or descendant. The directed ancestry walk preserves valid sibling ordering.

Maintainer fixes extend the rule across classic, proxied/domain, bulk/import, graph, and wisp write paths; use the typed dependency-target columns; process planned parent-child edges before blocking edges; and preserve hierarchy-before-cycle error precedence.

Verification passed across issueops, domain, Dolt, targeted pure-Go CLI tests, vet, and independent standards/spec review. Current base `main` CI is green. #4035 remains the stacked follow-up for combined scheduling-cycle validation.

PR #4035, opened 2026-05-18
fix(dep): detect cycles across combined blocks + parent-child graph

Ethee:
## Summary

Extends dependency cycle detection to walk both `blocks` (and `conditional-blocks`) **and** `parent-child` edges, and runs the cycle check on `parent-child` inserts in addition to blocks-type inserts.

This closes a livelock loophole exposed by #4034, which relaxes the same-type blocking ban (so cross-type `blocks` edges between unrelated issues are allowed). Together the two PRs preserve the GH#1495 safety invariant — "`bd ready` cannot deadlock from edge insertions" — while restoring legitimate cross-type workflows.

> **Stacks on:** #4034. The diff of this PR includes both commits; please review #4034 first.

## Background

Cycle detection in `AddDependencyInTx` previously walked only `blocks` and `conditional-blocks` edges:

```go
// Cycle detection for blocking deps via recursive CTE.
if !opts.SkipCycleCheck && (dep.Type == types.DepBlocks || dep.Type == types.DepConditionalBlocks) {
    ...
}
```

```sql
JOIN dependencies d ON d.issue_id = r.node AND d.type IN ('blocks', 'conditional-blocks')
```

That was sufficient when **every** task↔epic `blocks` edge was rejected outright (PR #2292), because no combined blocks + parent-child loop could form. After #4034 relaxes that ban, a chain of `blocks` and `parent-child` edges between unrelated pairs can still close a loop:

```
E1 ←blocks← T1
T1 child-of E2
E2 ←blocks← T0
T0 child-of E1
```

Every individual edge is between unrelated issues, so the parent-child shadow rule in #4034 allows each insertion in isolation. The combined graph closes a loop only at the last edge. Because the existing cycle detector only walks `blocks` edges, it does not see the loop, and the insert succeeds.

The operational consequence: when `bd ready` computes the blocked set, it marks each epic blocked by its direct blocker, then propagates blocked status from blocked parents to their children. In the loop above, every node ends up in the blocked set — `bd ready` shows nothing, even though no individual blocker is intrinsically un-clearable. This is exactly the GH#1495 livelock, just reached via a longer chain.

## What this PR does

Two small, coordinated changes in `internal/storage/issueops/dependencies.go`:

1. **Broaden the cycle-detection edge set.** `cycleReachabilityQuery` now includes `parent-child` in its type filter:

   ```sql
   JOIN dependencies d ON d.issue_id = r.node AND d.type IN ('blocks', 'conditional-blocks', 'parent-child')
   ```

   Walking `parent-child` edges in the `issue_id → depends_on_id` direction matches the scheduling-dependency semantics: a child task waits for its parent epic to not-be-blocked, which is exactly the direction the existing blocked-set propagation flows.

2. **Run the cycle check on parent-child inserts too.** A new helper `isSchedulingEdge` returns true for `DepBlocks`, `DepConditionalBlocks`, and `DepParentChild`. The cycle check now triggers for all three, so the closing edge of a combined loop is rejected regardless of which edge type closes it.

The error message is unchanged (`"adding dependency would create a cycle"`) — the new behavior is just a strictly larger superset of what's already detected.

## What stays allowed

- Linear cross-type chains of arbitrary depth: `T1 blocks E1`, `T1 child-of E2`, `T2 blocks E2`, `T2 child-of E3`, etc. — every pair is unrelated, no loop forms.
- Standard parent-child trees: each task has one parent, no cycle.
- Existing blocks-only chains: unchanged.
- Same-type blocks edges between unrelated peers: unchanged.

## What is now rejected

- A closing `blocks` edge that completes a combined-graph loop (test: `combined_graph_cycle_blocks_closes_loop`).
- A closing `parent-child` edge that completes a combined-graph loop (test: `combined_graph_cycle_parent_child_closes_loop`).

## Implementation notes

- The change is concentrated in a single function and a single helper. The query structure (recursive-CTE, single recursive term, single-vs-multi-table fast path) is identical to before; only the type filter expands.
- Existing tests in `internal/storage/issueops/dependencies_test.go` validate the query string structure. The string-match assertion is updated to expect the new filter; no other test infrastructure changes.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./internal/storage/...` clean
- [x] `go test ./internal/storage/issueops/...` clean (query-structure unit test updated)
- [x] `BEADS_TEST_EMBEDDED_DOLT=1 go test ./internal/storage/embeddeddolt/... -run TestAddDependency` — all 18 subtests pass (~8s), including the three new cycle tests:
  - `combined_graph_cycle_blocks_closes_loop` — closing `blocks` edge rejected
  - `combined_graph_cycle_parent_child_closes_loop` — closing `parent-child` edge rejected
  - `deep_linear_chain_allowed` — 5-level linear chain allowed
- [x] `BEADS_TEST_EMBEDDED_DOLT=1 go test ./internal/storage/embeddeddolt/...` — full embedded-dolt package green (~48s)
- [ ] CI: equivalent cycle tests in `internal/storage/dolt/dependencies_extended_test.go` (`CombinedGraphCycle_BlocksClosesLoop`, `CombinedGraphCycle_ParentChildClosesLoop`) — require Dolt testcontainer; couldn't run locally without Docker.

## References

- Stacks on: #4034 (parent-child shadow rule)
- Original livelock report: GH#1495
- Original same-type fix: PR #2292 (commit `ff1a409a5`)


quad341:
Thanks @Ethee — the combined-graph cycle detection is a real and important fix. You correctly identified that #4034's relaxation reopens the GH#1495 livelock through multi-hop chains that mix `blocks` and `parent-child` edges, and the test cases (both closing-edge insertion orders, plus the allowed linear-chain case) demonstrate you understand exactly what you're guarding against.

**Maintainer-side work needed before this can land:**

1. **#4034 needs to merge first.** This PR's body says it stacks on #4034, but the base ref is `main` and the diff bundles both commits. The intent is clearly stack ordering, so the right sequence is:
   - Land #4034 (after its sibling-blocking fix, see comment there)
   - Rebase this branch onto the post-#4034 main — the parent-child shadow code lands in #4034 and will drop cleanly out of this diff
   - The resulting diff should show only the combined-graph cycle detection additions

2. **Rebase is currently needed** (branch shows CONFLICTING). That's ours to absorb.

3. **CI after rebase**: You noted you couldn't run the full Dolt testcontainer suite locally. After the rebase, maintainer will run those too.

No changes to your logic needed — the CTE broadening (adding `parent-child` to the blocked-set walk) is correct. Thanks for shipping both the fix and the follow-up in a stack; it makes the safety invariant easy to reason about.

---
*Fellow-contributor review (Claude+Codex ensemble, Qwen unavailable — read access only, maintainer merge needed)*

quad341:
**Code review — @Ethee's combined-graph cycle detection**

**Verdict: ✅ looks good to merge (after #4034 lands)**

This is a well-reasoned and well-tested extension to the GH#1495 safety invariant. The problem it solves is real: after #4034 allows cross-type `blocks` edges between unrelated issues, a cycle can form through a mix of `blocks` and `parent-child` edges. The existing cycle detector only walked `blocks` edges, so it missed these cross-type loops.

**Core changes:**
1. **Remove the broad cross-type ban** (from #4034, included in this stack): cross-type `blocks` edges between unrelated issues are now allowed
2. **Add parent-child shadow rejection**: new `parentChildLinkedQuery` CTE checks if the `blocks` endpoint pair already shares a parent-child relationship (in either direction); if so, the `blocks` edge is rejected
3. **Extend cycle detection to parent-child edges**: `cycleDetectionTables` now includes `'parent-child'` in the type filter, detecting cycles across the full scheduling graph

**What's correct:**
- `parentChildLinkedQuery` uses `UNION` (not `UNION ALL`), preventing infinite recursion if a parent-child cycle somehow already existed in the DB
- Both directions of parent-child are emitted into the derived edge relation, so ancestor-or-descendant detection is symmetric
- Self-dependency check is correctly moved before the graph queries (with a comment explaining why)
- `isSchedulingEdge` helper cleanly documents which dep types participate in cycle detection
- The combined extended tests in `dependencies_extended_test.go` cover unrelated cross-type blocks (allowed), parent-child shadow blocks (rejected), and combined-graph cycle formation (rejected)
- All CI passes with #4034 changes included in the stack

**One question (not a blocker):**

The `parentChildLinkedQuery` recursive CTE runs on every `blocks`-type insertion. For large dependency graphs, this could be a hot path. Has the query been benchmarked against large fixture sets? The cycle CTE at the end already traverses the full graph, so this adds a second recursive query per insert — worth profiling before this pattern becomes load-bearing. Not blocking, but worth a follow-up issue.

**Dependency note:** This PR includes #4034's commits. If #4034 lands as maintainer cherry-pick #4077, this PR should be rebased to show only its own additions. Doesn't affect correctness.

---

quad341:
**beads/reviewer automated review — REQUEST-CHANGES**

@Ethee — thank you for this PR. The cycle-detection insight is exactly right: relaxing the same-type blocking ban (in #4034) does open a livelock loophole for combined `blocks` + `parent-child` cycles, and extending `cycleReachabilityQuery` to walk `parent-child` edges closes it cleanly. The `isSchedulingEdge` helper is a nice touch for clarity.

There's a schema-correctness blocker that needs fixing before this can land.

---

**CRITICAL: `depends_on_id` does not exist as a physical column**

Migration 0043 (`0043_drop_dependencies_generated_column.up.sql`) dropped the generated `depends_on_id` column from the `dependencies` table. The three split columns (`depends_on_issue_id`, `depends_on_wisp_id`, `depends_on_external`) replaced it. Queries that reference `depends_on_id` directly will fail at runtime with "column not found".

Affected locations in this diff:

1. **`parentChildLinkedQuery`** (the new function):
   ```go
   fmt.Sprintf("SELECT issue_id AS src, depends_on_id AS dst FROM %s WHERE type = 'parent-child'", t)
   ```
   `depends_on_id` is not a physical column here — needs `COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external) AS dst`.

2. **Idempotency/conflict check** in `AddDependencyInTx`:
   ```go
   `SELECT type FROM %s WHERE issue_id = ? AND depends_on_id = ?`
   ```

3. **Metadata update** in `AddDependencyInTx`:
   ```go
   `UPDATE %s SET metadata = ? WHERE issue_id = ? AND depends_on_id = ?`
   ```

4. **`RemoveDependencyInTx`**:
   ```go
   `DELETE FROM %s WHERE issue_id = ? AND depends_on_id = ?`
   ```

5. **`GetDependenciesWithMetadataInTx`** and **`GetDependenciesInTx`**:
   ```go
   `SELECT depends_on_id, type FROM %s WHERE issue_id = ?`
   `SELECT depends_on_id FROM %s WHERE issue_id = ?`
   ```

6. **`cycleReachabilityQuery`** multi-table path:
   ```go
   fmt.Sprintf("SELECT issue_id, depends_on_id FROM %s WHERE type IN ('blocks', 'conditional-blocks', 'parent-child')", t)
   ```

**Fix**: Replace each `depends_on_id` column reference with `COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external)` (aliased as `depends_on_id` in projection columns so outer query references still work). The existing `DepTargetExpr` constant is the right expression to use — see how the maintainer's cherry-pick (#4077) uses it in `parentChildAncestorQuery` as a reference.

---

**Everything else looks good**

- The combined-graph cycle detection logic is correct.
- `isSchedulingEdge` cleanly expresses which edge types participate in scheduling.
- The updated test cases (`combined_graph_cycle_blocks_closes_loop`, `combined_graph_cycle_parent_child_closes_loop`, `deep_linear_chain_allowed`) verify the right invariants.
- The simplification of `UpdateWispIDInDependenciesInTx` (removing `replaceDependencyTargetInTx`) looks correct — the typed column update `SET depends_on_wisp_id = ?` is valid.

Once the `depends_on_id` column references are fixed, this should be good to go.

quad341:

maphew:

steveyegge:
Status note: #4034 (which this stacks on) just received a maintainer fix-merge bringing it to current main — the pairwise guard is now a directed ancestry check over the typed dependency target columns. This branch is now well behind both main and its base.

The combined blocks + parent-child cycle detection here is still wanted: the pairwise guard in #4034 deliberately does not catch multi-hop composite loops (e.g. E1←blocks←T1, T1 child-of E2, E2←blocks←T0, T0 child-of E1). Once #4034 lands, this branch needs the same treatment — rebase/merge onto main and port the extended cycle CTE to the current `cycleReachabilityQuery` depTables-union form. Happy to do that fix-merge as a follow-up, or feel free to update it yourself first.

maphew:
Maintainer fix-merge pushed to this contributor branch at `bcaeb1640`.

This is now stacked on merged #4034 and current `main`, with Jacob's original combined-cycle commit preserved. The implementation carries the combined `blocks` + `conditional-blocks` + `parent-child` rule through every dependency mutation surface I could identify:

- classic, embedded, and proxied/domain single-add paths
- regular and wisp dependency tables, using the typed target columns
- bulk/import paths, including `--no-cycle-check`
- classic and domain graph apply, including implicit parents and inline dependencies
- defensive repository inserts and cross-table final transaction gates

The maintainer changes also keep #4034's hierarchy error ahead of the generic cycle error, process planned parent-child edges first so input order cannot bypass that hierarchy guard, and merge regular+wisp transaction snapshots before bulk/graph commit. `bd dep cycles` intentionally remains blocking-only reporting; the broader graph is a mutation-validity rule.

Local verification is green:

- issueops, domain, Dolt, and full pure-Go storage suites
- dependency/readiness conformance audit
- targeted pure-Go CLI bulk and graph tests
- pure-Go `go vet`
- independent Standards and Spec reviews, both PASS after adversarial re-review

GitHub checks are running on the refreshed head. I am leaving merge gated on both post-#4034 `main` CI and this PR's checks.

maphew (approved):
Maintainer review complete after fix-merging #4034 and refreshing this branch onto its merge commit.

The contributor's core design is preserved: reject a dependency mutation when the directed combined graph of `blocks`, `conditional-blocks`, and `parent-child` would become cyclic, regardless of which of those edge types closes the loop. Linear cross-type sequencing remains valid.

The maintainer commits extend the rule across classic, domain/proxied, wisp, bulk/import, and graph paths; use typed dependency targets; preserve hierarchy-conflict precedence; and add merged regular+wisp transaction gates for paths where per-edge checks cannot see both uncommitted snapshots.

Local suites, conformance, vet, and independent Standards/Spec reviews pass. Approval is for the code at `bcaeb1640`; merge remains gated on the running post-#4034 base and PR checks.

maphew:
Pushed `de9baef74` with the narrow follow-up from the first CI run.

The implementation itself passed the other completed jobs. The storage-domain failure was two stale test expectations introduced while extending the coverage:

- typed `depends_on_wisp_id` targets now participate in combined-graph reachability, so that fixture should report a cycle;
- the low-level graph-use-case test invokes the use case without its normal outer unit of work, so it should assert that the cycle-closing edge is rejected rather than promise rollback of earlier acyclic writes.

The new head updates only those assertions and also merges current `main` after #4691. CI is rerunning on the exact candidate head.

PR #4039, opened 2026-05-19
refactor(deps): resolve depends_on_id via typed columns at read time

coffeegoddd:
## Summary

Every Go SQL query that previously named the polymorphic `depends_on_id` column on `dependencies` / `wisp_dependencies` now resolves the target via the three typed columns (`depends_on_issue_id`, `depends_on_wisp_id`, `depends_on_external`) — either via a `COALESCE(...) AS depends_on_id` projection or, where the caller knows the target kind, the typed column directly.

## Test plan

- [x] `make test`
- [x] `golangci-lint run ./...`


coffeegoddd:
removes generated column with fk which is an invalid schema: https://github.com/dolthub/dolt/issues/11065

PR #4053, opened 2026-05-20
feat(doctor): detect orphaned child issues in deep validation (bd-3852)

quad341:
## Summary

- Adds `checkOrphanedChildren` to `RunDeepValidation` in `cmd/bd/doctor/deep.go`
- Detects issues whose ID contains a `.` (e.g. `parent.1`) but whose parent prefix is absent from the `issues` table
- Complements the existing `checkParentConsistency` (which only scans the `dependencies` table)
- Reports as `StatusWarning` with up to 3 examples; suggests `bd admin delete <id>` or re-creating the parent
- Adds `OrphanedChildren DoctorCheck` field to `DeepValidationResult` (JSON: `orphaned_children`)

## Bead

bd-3852: Add orphan detection migration

## Test plan

- [x] Build clean: `go build ./cmd/bd/...`
- [x] `go vet ./cmd/bd/doctor/...` — 0 issues
- [x] `golangci-lint run ./cmd/bd/doctor/...` — 0 issues
- [x] Existing doctor tests pass (pre-existing failures in role/integrity tests are unrelated to this change, confirmed on main)
- [ ] Validator: add embedded test calling `RunDeepValidation` with a seeded DB that has an orphaned child, assert `OrphanedChildren.Status == StatusWarning`

### DIFF
--- a/cmd/bd/doctor/deep.go
+++ b/cmd/bd/doctor/deep.go
@@ -18,6 +18,7 @@ type DeepValidationResult struct {
 	EpicCompleteness    DoctorCheck   `json:"epic_completeness"`
 	MailThreadIntegrity DoctorCheck   `json:"mail_thread_integrity"`
 	MoleculeIntegrity   DoctorCheck   `json:"molecule_integrity"`
+	OrphanedChildren    DoctorCheck   `json:"orphaned_children"`
 	AllChecks           []DoctorCheck `json:"all_checks"`
 	TotalIssues         int           `json:"total_issues"`
 	TotalDependencies   int           `json:"total_dependencies"`
@@ -115,6 +116,9 @@ func RunDeepValidation(path string) DeepValidationResult {
 	if result.MoleculeIntegrity.Status == StatusError {
 		result.OverallOK = false
 	}
+
+	result.OrphanedChildren = checkOrphanedChildren(db)
+	result.AllChecks = append(result.AllChecks, result.OrphanedChildren)
 
 	return result
 }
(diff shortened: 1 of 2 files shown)


quad341:
**Code review — feat(doctor): detect orphaned child issues in deep validation**

**Verdict: ✅ looks good to merge (one LOW, not a blocker)**

The check fills a real gap: `checkParentConsistency` only scans the `dependencies` table for parent-child relationships, missing ID-based orphans where the dotted-prefix convention isn't backed by a dep row. Using `SUBSTRING_INDEX(id, '.', 1) NOT IN (SELECT id FROM issues)` is the right approach for this naming convention.

**LOW — message format is misleading when results are under the LIMIT:**

```go
check.Message = fmt.Sprintf("Found %d+ child issues whose parent no longer exists", len(orphans))
```

The `+` suffix implies "at least N" but fires even when `len(orphans)` < 10 (the LIMIT). If only 3 orphans exist, this reads "Found 3+ child issues" — implying there could be more, when there aren't. Suggest:
```go
if len(orphans) == 10 { // == LIMIT
    check.Message = fmt.Sprintf("Found %d+ child issues...", len(orphans))
} else {
    check.Message = fmt.Sprintf("Found %d orphaned child issues...", len(orphans))
}
```
Or simply remove the `+` and note the LIMIT in the Detail.

**INFO — shallow hierarchy detection:**

`SUBSTRING_INDEX(id, '.', 1)` extracts only the root prefix. An issue like `bd-3852.1.2` (two levels deep) would check whether `bd-3852` exists, not whether the intermediate `bd-3852.1` exists. This is the right trade-off for a warning check (false negatives are safe), but worth noting if deep hierarchies become common.

All CI passes. The LOW is worth a follow-up but doesn't block.

---

quad341:
@quad341 — this closes a real blind spot. `checkParentConsistency` only ever looked at the `dependencies` table, so dotted-ID orphans (`parent.1` with `parent` deleted) slipped through entirely. Keying off the ID convention with `SUBSTRING_INDEX(id, '.', 1) NOT IN (SELECT id FROM issues)` is exactly the right way to catch them, and reporting as a warning (rather than flipping `OverallOK`) is the correct call — it matches how `EpicCompleteness` treats informational findings.

Nice touch fixing the `%d+` message to only fire at the limit and adding the test that pins that down. The shallow-prefix trade-off is the right one for a warning check, since false negatives are harmless here. CI is green across the whole matrix, including the embedded-Dolt lanes that actually exercise the new tests — the Codecov 0% is just the coverage lane not running the Dolt-tagged tests, nothing real.

Merging as-is. The end-to-end `RunDeepValidation` test from your test plan would be a nice follow-up but it's not needed to land this. Clean, well-scoped change — thanks!


---


maphew:
Thanks for the orphaned-child doctor check. I am closing this orphaned pre-re-root PR because current `main` already contains the deep validation check, landed in commit `94298a169` and refined by `a09a1258c`, so the useful diagnostic is already present.

PR #4054, opened 2026-05-20
feat(preflight): implement --fix mode for vendorHash and version sync

quad341:
## Summary

- Add `--fix` flag to `bd preflight` that auto-corrects two common preflight failures without manual intervention
- `fixNixHash()`: uses sentinel-hash trick with `nix build` to compute and write the correct `vendorHash` into `default.nix` when `go.sum` has uncommitted changes; prints clear error if `nix` is not in PATH
- `fixVersionSync()`: syncs `cmd/bd/version.go` to match `default.nix` version field (`default.nix` is source of truth); no external tools required
- Human-readable and `--json` output modes, original file permissions preserved on all writes

Closes be-xra8, be-roho (children of epic bd-lfak).

## Test plan

- [ ] `bd preflight --fix` with a stale `go.sum`: `vendorHash` updated in `default.nix`
- [ ] `bd preflight --fix` with version mismatch: `version.go` updated to match `default.nix`
- [ ] `bd preflight --fix` with `nix` not in PATH: exits 1 with manual-fix instructions
- [ ] `bd preflight --fix` with nothing to fix: clean exit, no file writes
- [ ] `go test ./cmd/bd/...` passes


quad341:
Thanks @quad341 — turning `bd preflight --fix` from a "not yet implemented" stub into a real auto-corrector is exactly the kind of release-prep paper-cut elimination that compounds. The sentinel-hash trick for `vendorHash` is clean, the `nix`-not-in-PATH path gives a real manual recipe instead of a cryptic error, and the deferred restore so an aborted `nix build` doesn't leave `default.nix` poisoned with the sentinel hash is the right safety move. I also appreciate that you bundled the `CLI_REFERENCE.md` and versioned doc regenerations in the same PR so the doc-flag freshness check stayed green.

CI is 41/41, the in-tree release gate (`release-gates/be-xra8-be-roho-gate.md`) already records a PASS with only two cosmetic LOW notes, and the test follow-up is tracked under be-b6m9 on its own validator branch. Merging this as-is.


---


quad341:
**Code review — feat(preflight): implement --fix mode for vendorHash and version sync**

The implementation is correct. Both fix operations are mechanical and safe:

- `fixNixHash()` uses the well-established sentinel-hash trick: set vendorHash to a zero hash, run `nix build`, extract the real hash from the "got: sha256-..." error output, then write it back. The 5-minute timeout is appropriate for a Nix build. The regex fallback (when the primary pattern isn't found) is a reasonable safety net.
- `fixVersionSync()` reads the version from `default.nix` and writes it to `cmd/bd/version.go` — no external tools, clear source-of-truth (default.nix wins).

Both operations print old→new diffs and the `--json` output mode is consistent with the rest of the preflight command. File permissions are preserved on write. No-op if nothing needs fixing (Skipped result).

The flag help text is properly updated from "not yet implemented" to describe what's actually fixed.

All CI passes.

---

quad341:
Thanks @quad341 — this turns `bd preflight --fix` from a "not yet implemented" stub into something that actually kills the two most annoying release-prep paper-cuts. The sentinel-hash dance for `vendorHash` is clean, and the deferred restore so an aborted `nix build` never leaves `default.nix` poisoned with the sentinel hash is exactly the right safety move. I also like that the `nix`-not-in-PATH branch hands the user a concrete three-step recipe instead of a cryptic failure, and that you bundled the `CLI_REFERENCE.md` + versioned-doc regenerations so the doc-flags freshness check stayed green.

CI is 41/41, the in-tree release gate is PASS, and the only open notes are cosmetic (empty `Detail` on the version-sync skip path, fallback-regex robustness) — none block. Test coverage for the fixer paths is tracked under be-b6m9 on its own branch, so that's a known follow-up rather than a gap we're shipping blind. Merging as-is. Nice, compounding cleanup. 🎯


---


maphew:
Maintainer fix-merge: merged current main into this branch and reconciled the `--fix` implementation onto the `RunE` refactor from #4419.

What happened in the meantime: your implementation here was developed in parallel with the preflight `RunE`/usage-metrics refactor (#4419, developed off the same 2026-05-19 base), so main's `preflight.go` moved underneath this PR without ever rejecting it - the "not yet implemented" stub on main was parallel-development collateral, not a decision against `--fix`.

What the merge commit (9cd563b20) does:

- Keeps your `runFixes()` / `fixNixHash()` / `fixVersionSync()` byte-identical.
- Keeps main's `RunE: runPreflight` signature, `SilenceUsage`/`SilenceErrors`, and the `metrics.NewCommandEvent("preflight")` deferred-close block verbatim.
- One semantic fix the textual merge missed: the `if fix { runFixes(...); return }` bare `return` from the old `Run` signature is now `return nil` under `RunE` (was a compile error after auto-merge).
- Flag help and current docs now describe `--fix` as implemented; the frozen versioned doc snapshots (1.0.x/1.1.0) were left untouched.

Validated locally: `CGO_ENABLED=0 go build -tags gms_pure_go ./...`, `go vet ./cmd/bd/...`, and `go test ./cmd/bd/ -run 'Preflight|preflight'` all pass. Dedicated tests for the fix functions remain follow-up work as noted in your release-gate file.

Thanks for the feature - sorry it sat while main drifted. Will merge once CI is green.

maphew:
Merged with all 54 checks green. `bd preflight --fix` is now live on main, closing the gap where the flag was advertised but stubbed. Thanks @quad341.

Issue #3970, opened 2026-05-16
bd warns but exits 0 when auto-export 'git add' fails; staged jsonl can ship stale state


claracurrier128:
**bd version:** 1.0.3 (1b2dd2cb), linux/amd64

## Summary

When `bd <write-op>` (e.g. `bd dep add`) completes but the auto-export's `git add .beads/issues.jsonl` step fails, bd:

- prints `Warning: auto-export: git add failed: exit status <N>`
- **exits 0**
- correctly persists the change to dolt and to `.beads/issues.jsonl` on disk

The dolt state and the on-disk jsonl are correct. The **git index** is left holding the *previous* jsonl. If the caller runs `git commit` before the next bd write op (which would re-stage and self-heal), the commit ships a jsonl that diverges from dolt.

## Minimal repro

```bash
mkdir /tmp/bd-repro && cd /tmp/bd-repro
git init -q && git commit -q --allow-empty -m init
bd init >/dev/null
bd create -t bug -p 2 first  >/dev/null
bd create -t bug -p 2 second >/dev/null
A=$(bd list --json | jq -r '.[0].id')
B=$(bd list --json | jq -r '.[1].id')

# Simulate git index contention (concurrent worktree, hook, agent, etc.):
touch .git/index.lock

bd dep add "$A" "$B"   # warns; exits 0
echo "exit=$?"

rm -f .git/index.lock
```

Output:

```
✓ Added dependency: ... depends on ...
Warning: auto-export: git add failed: exit status 128
exit=0
```

Post-state:

- `bd show $A` → DEPENDS ON is present (dolt correct)
- `cat .beads/issues.jsonl` → dependency entry present (jsonl on disk correct)
- `git ls-files --stage .beads/issues.jsonl` → blob hash is the *pre-op* version (staged jsonl stale)
- `git status` → `AM .beads/issues.jsonl` (index has old, worktree has new)

A subsequent `bd <write-op>` re-exports + re-stages and self-heals. But a `git commit` interleaved between the failing op and the next bd op ships divergent state.

## Why this hurts agents

In a multi-agent harness where workers and a manager fire bd ops concurrently against one workspace, transient `index.lock` contention (or pre-commit hooks, or concurrent `git add` from another tool) is realistic. The exit-0 + warning combo means:

1. The caller cannot detect the failure programmatically — has to grep stderr for the warning text, which is fragile.
[...]

Issue #3962, opened 2026-05-15
bd pre-commit hook: defer auto-export when staging non-bd files


fkberthold:
## Problem

After `bd hooks install`, the pre-commit hook auto-exports
`.beads/issues.jsonl` on every commit, including the first commit after
install when the user is mid-way through staging unrelated files. The
auto-export silently joins those changes, yielding a
"two-things-in-one-commit" result against intent.

Concrete shape: a user runs `bd hooks install`, then makes a focused
code change unrelated to bd state, runs `git add <file>` + `git commit`.
The pre-commit hook re-exports `.beads/issues.jsonl` post-staging; the
re-export joins the focused commit silently. The user only notices
later via `git log -p`.

## Why this is annoying

- Git history is noisy: focused commits get bd churn appended.
- The user has no signal that the auto-export landed until they see it
  in `git log -p`.
- The mitigation today is the "absorbing commit" pattern: before any
  logical commit after `bd hooks install`, run an empty bd-only commit
  to flush the export. That works but is unintuitive and project
  onboarders forget it.

## Proposed shape

The pre-commit hook detects whether the user is actively staging
non-bd files. If so, defer the auto-export to its own follow-up commit
(or skip and let the next `bd`-driven commit pick it up naturally).
Heuristic: if `git diff --cached --name-only` contains any file
outside `.beads/`, defer.

Alternative: the hook could prompt (when interactive) or just emit a
warning telling the user "deferring bd export — run `git commit` again
to absorb."

## Acceptance from a downstream POV

- Pre-commit hook respects the staging pattern.
- Existing single-purpose bd-only workflows still auto-export as today.
- Downstream "absorbing commit" mitigation patterns become retirable.

---

Filed by loom (https://github.com/fkberthold/loom); downstream tracker
loom-yxn. Related: bd post-commit re-export loop (separate issue).

maphew:
Assessment after merged #4063 (`3988db7e0da1d5488e5ff972349f736466bec1c8`): partially addressed.

#4063 fixed the blast-radius side by making JSONL auto-export opt-in by default: `export.auto=false`, `export.git-add=false`, and init now prompts `[y/N]`.

[...]

Issue #3923, opened 2026-05-13
pre-commit hook stages unrelated working-tree changes via 'git add -A'


pmacmillan:
## Summary

With `export.git-add=true` (default) the bd pre-commit hook stages **all** working-tree changes — tracked modifications **and** new untracked files — not just `.beads/issues.jsonl`. This sweeps unrelated WIP, secrets, and temp files into bd-driven commits.

## Repro

1. Fresh clone with bd configured (any 1.x).
2. Confirm `bd config get export.auto` and `bd config get export.git-add` are both true (defaults).
3. In a dirty working tree (modified tracked files + new untracked files unrelated to bd), run any bd command that triggers the auto-export commit, e.g. `bd update <id> --status in_progress`.
4. Observe the resulting commit: it includes all the unrelated modifications and new files, not just `.beads/issues.jsonl`.

## Expected

The hook should stage **only** the bd export artifact (`.beads/issues.jsonl`) with an explicit path:

```sh
git add .beads/issues.jsonl
```

## Actual

The hook effectively runs `git add -A` (or equivalent), staging every modified/untracked path. Disabling **only** `export.auto` is insufficient — `export.git-add=true` still triggers the sweep even when `auto` is off. Both knobs must be disabled to mitigate.

## Impact

- Cross-stream work-stealing in concurrent dev streams (agent A's WIP gets swept into agent B's bd commit).
- Worst case: secrets, `.env*` files, or unrelated temp files get committed.
- Disabling auto-export entirely (the current workaround) means losing the convenience that drives most users to enable bd's hook in the first place.

## Workaround

```sh
bd config set export.auto false
bd config set export.git-add false
```

Then stage `.beads/issues.jsonl` explicitly:

```sh
bd export && git add .beads/issues.jsonl
```

## Related

Conceptually adjacent to #3606 (auto-export + index.lock race) but a different failure mode: that one is concurrency, this one is scope-of-staging.

PR #3505, opened 2026-04-26
fix(auto-export): surface git add stderr in failure warning

seanmartinsmith:
## Problem

`auto-export: git add failed` warning fires once per `bd` mutation when `export.git-add` is enabled, but discards `git add`'s stderr — users see the symptom (`exit status 1`) without the cause. Across multi-command sessions this trains people to ignore the warning entirely, at which point real signal also gets missed.

## Root Cause

`cmd/bd/export_auto.go:282` (pre-fix) called `cmd.Run()`, which throws away stdout + stderr. The caller at `:113` then formatted the warning with `%v` on just the bare exec error.

## Fix

`gitAddFile` switches to `cmd.CombinedOutput()` and, on error, wraps the trimmed output via `fmt.Errorf("%w: %s", err, trimmed)`. Function signature stays `error`. The caller's existing `%v` formatter then naturally prints `exit status 1: <git's stderr>`.

```go
out, err := cmd.CombinedOutput()
if err != nil {
    if trimmed := strings.TrimSpace(string(out)); trimmed != "" {
        return fmt.Errorf("%w: %s", err, trimmed)
    }
    return err
}
return nil
```

No change to the success path, no change to callers, no signature change. `errors.Is`/`errors.As` consumers still get the underlying exec error via `%w`.

## Test Plan

- New `TestGitAddFile_CapturesStderrOnFailure`: temp git repo with `.gitignore` excluding `.beads/`, attempts `gitAddFile` on a target inside `.beads/`, asserts the returned error message contains `"ignored"` (proxy for git's stderr surfacing).
- All 4 `TestGitAddFile_*` tests pass locally:
  ```
  --- PASS: TestGitAddFile_InWorktreeHook_StagesCorrectPath (0.25s)
  --- PASS: TestGitAddFile_NonHookContext_GuardDoesNotFire (0.09s)
  --- PASS: TestGitAddFile_CapturesStderrOnFailure (0.08s)   ← new
  --- PASS: TestGitAddFile_RedirectCase_DoesNotStageInMainRepo (0.23s)
  ```
- `golangci-lint run --build-tags gms_pure_go ./cmd/bd/...` clean.

## Observed triggers in real sessions

Captured stderr would have made all of these obvious instead of "exit status 1":

- A concurrent `git commit` holding the index lock — `Unable to create index.lock: File exists`
- A parent `.gitignore` excluding `.beads/issues.jsonl` — `The following paths are ignored by one of your .gitignore files`
- Detached HEAD or rebase-in-progress states

[...]

PR #4061, opened 2026-05-20
fix(auto-export): bypass throttle when export path is gitignored (GH#3848)

### LINKED ISSUE #3848: bug: auto-export warns on git add when .beads is gitignored (server-mode operating model gap)
## What

After every successful `bd` write command, `bd` attempts `git add .beads/issues.jsonl` as part of its auto-export step. When `.beads` is in `.gitignore`, git refuses and `bd` prints:

```
Warning: auto-export: git add failed: exit status 1: The following paths are
ignored by one of your .gitignore files: .beads
hint: Use -f if you really want to add them.
```

`bd` then proceeds normally — the warning is the only side effect.

## Why this is a design-assumption gap

`bd`'s auto-export assumes `.beads/issues.jsonl` is meant to be git-tracked (the standard "share state via committed JSONL" model). There is a second valid operating model: workspaces that share state via a running dolt sql-server, where `.beads` is legitimately gitignored because committed JSONL is not the source of truth. The auto-export step does not distinguish between these two models, so it warns unconditionally.

## Steps to reproduce

1. Initialize a `bd` workspace with `metadata.json`:
   ```json
   { "dolt_mode": "server", "dolt_database": "<name>" }
   ```
   and a reachable dolt sql-server configured in `.beads/config.yaml`.
2. Add `.beads` (or `.beads/issues.jsonl`) to `.gitignore` — appropriate when the workspace shares state via the server rather than committed files.
3. Run any `bd` write command:
   ```
   bd create "test issue"
   ```
4. Observe stderr:
   ```
   Warning: auto-export: git add failed: exit status 1: The following paths are
   ignored by one of your .gitignore files: .beads
   hint: Use -f if you really want to add them.
   ```
5. `bd` succeeds. The warning repeats on every subsequent write.

## Expected behavior

One of:
- **(a) Detect-and-skip:** Before calling `git add`, check whether the target path is gitignored (e.g. via `git check-ignore -q <path>`). If ignored, skip silently — the workspace has signaled it does not want these files in git.
- **(b) Explicit config knob:** Document `export.git-add: false` (already accepted by `bd config set`) as the supported way to disable auto-export git tracking for workspaces that use the server-sharing operating model. Surface this in `bd init` or `bd dolt` guidance for server-mode setups.

## Actual behavior

`bd` always attempts `git add .beads/issues.jsonl` regardless of gitignore status. Workspaces that legitimately gitignore `.beads` see a warning on every write.

## Environment

- bd version: 1.0.4 — running a **locally-patched build** (commit `ad31139a5` on a local-only branch in our fork) with a workaround for an unrelated server-mode bug — see #3849 (auto-import has no emptiness guard, so every write re-imports the entire JSONL). The patch does not affect the auto-export `git add` behavior described here; mentioning for full environment disclosure.
- OS: Linux x86_64
- Go: 1.26.3

## Prior art

- #3504 improved the warning to include git's actual stderr message (making "The following paths are ignored..." visible, which helps reveal this root cause when encountered).
- #3314 addressed the related case where `no-git-ops = true` did not cascade to suppress `export.git-add`. That config approach remains the documented workaround (`bd config set export.git-add false`), but it requires the user to know to set it — and there is no hint in `bd init` or the server-mode docs that this is expected.

jeremylongshore:
# fix(auto-export): bypass throttle when export path is gitignored

Fixes #3848. Related: #3970, #3962.

## The bug

When `.beads/` (or whatever path holds the auto-export JSONL) is excluded by `.gitignore`, the JSONL on disk silently falls behind in-memory Dolt state for the duration of the auto-export throttle window (default 60s).

The mechanism is **the throttle, not the git-add failure** (which was already correctly demoted to a non-fatal warning by #3505):

1. `bd <write-command>` runs → Dolt state advances → `maybeAutoExport()` is invoked via `PersistentPostRun`.
2. The first export succeeds and `export-state.json` records the timestamp.
3. Any subsequent `bd <write-command>` within `export.interval` (60s default) is throttled — `maybeAutoExport()` returns early without rewriting the JSONL.
4. For `git-tracked` `.beads/`: not a problem — the next bd command run after the throttle expires catches up, or `git commit`'s pre-commit hook forces a re-export.
5. For `gitignored` `.beads/`: **the JSONL is the only cross-machine sync substrate.** When the throttle blocks the export AND no bd command runs after the throttle expires (e.g., a bulk operation finishes), the JSONL never catches up. Cross-machine sync, fresh clones, container restarts all see the stale pre-burst state and the writes are lost.

## Reproduction

A self-contained shell-script reproduction is at the bottom of this description. The smoking gun is the on-disk JSONL after a sequence of writes inside the throttle window:

```
ORIGINAL bd 1.0.4:  abtest-crb  notes=<MISSING>            ← write discarded
PATCHED bd 1.0.4:   abtest-ig4  notes=GitHub: example/repo#1   ← write persisted
```

## The fix

Two small structural changes to `maybeAutoExport()` in `cmd/bd/export_auto.go`:

1. **Resolve the export path before the throttle check** so the throttle decision can consult the actual path that would be written.
2. **Move change-detection (cheap Dolt-hash compare) before the throttle check** so we can short-circuit on no-op runs without ever consulting the throttle (also defensible as an independent micro-optimization).
3. **Bypass the throttle when the export path is git-ignored.** In a gitignored workspace, the JSONL is the sole cross-machine sync substrate; throttling a pending write means the write is lost the next time embedded Dolt state is rebuilt. There is no other catch-up path (no git push will carry it, no pre-commit hook will refresh it).

New helper `isExportPathGitignored(ctx, beadsDir, path)`:

- Probes via `git check-ignore -q -- <path>` (the `--` separator is defense-in-depth, in case a future caller passes an attacker-shaped path; current callers are safe).
- Distinguishes exit code 0 (ignored → true), 1 (not ignored → false), and 128 (not in a git repo → false). Any other exit / transient error logs via `debug.Logf` and returns false. **This is "fail closed for bypass"**: when in doubt, behave like the pre-fix code (throttle) rather than newly amplifying writes.
- Wraps the subprocess in a 2-second `context.WithTimeout` because it runs on the main goroutine via `PersistentPostRun` and a hung git subprocess on a slow filesystem would block every bd command.
- Reuses `scrubGitHookEnv(os.Environ())` so stray `GIT_DIR` / `GIT_WORK_TREE` / `GIT_CONFIG*` variables can't redirect the probe at an unrelated repo (same pattern as the existing `gitAddFile`).
- Memoized per-process via `sync.Mutex`-guarded map. The probe result is stable for a process lifetime in practice; if the user edits `.gitignore` mid-process the next bd invocation re-probes.

## Tests

- **`TestIsExportPathGitignored`** — 8 table-driven sub-cases covering: `.gitignore` dir-pattern, tracked workspace, not-in-git-repo, parent-directory pattern (umbrella workspace), file-level wildcard (`*.jsonl`), `.git/info/exclude` per-clone rules, file-not-yet-on-disk, non-default `export.path`.
- **`TestRunGitignoreProbe_Timeout`** — pre-cancelled context returns false; verifies the timeout/cancel path.
- **`TestIsExportPathGitignored_Cache`** — modifying `.gitignore` mid-process does not change cached result; verifies the memoization.
- **`TestEmbeddedAutoExport_GitignoredBeads_BypassesThrottle`** (CGO) — end-to-end: gitignored workspace + two writes inside throttle window + assert second write is in the JSONL.
- **`TestEmbeddedAutoExport_TrackedBeads_ThrottlesAsBefore`** (CGO) — inverse regression guard: tracked workspace + two writes inside throttle + assert throttle holds; then advance state timestamp + assert export catches up.

All slow tests (real git subprocess or embedded Dolt) skip via `if testing.Short() { t.Skip(...) }` per the two-tier policy in `CONTRIBUTING.md`.

## Verification (locally)

```
$ go test -short -run "TestIsExportPathGitignored|TestRunGitignoreProbe" ./cmd/bd
ok      github.com/steveyegge/beads/cmd/bd      0.387s

$ go vet ./cmd/bd/
(clean)

$ gofmt -d cmd/bd/export_auto.go cmd/bd/export_auto_test.go cmd/bd/export_auto_embedded_test.go
(clean)
```

## Out of scope (explicit follow-ups)

- **Honoring an explicit `export.interval` config value** (so power users can opt back into strict throttling in gitignored workspaces) — requires `viper.IsSet` plumbing through the `config` wrapper to distinguish "user set 60s" from "default is 60s." Worth a separate small PR.
- **Pulling the bypass behind a `syncSubstrateRequiresJSONL()` abstraction** so future sync substrates (DoltHub native, S3 snapshots, etc.) can plug into the same decision point. Today's gitignore probe is the right heuristic; the abstraction matters when there's a second consumer.
- **Splitting into two commits** (one for the change-detection-before-throttle reorder, one for the gitignored-bypass) — happy to do this if the maintainers prefer it; left as one commit because the diff is small.

## Reproduction script (drop into any temp dir)

```bash
#!/usr/bin/env bash
set -uo pipefail

BD="$1"  # path to bd binary

run_scenario() {
  local TMP=$(mktemp -d)
  cd "$TMP"
  git init -q
  echo ".beads/" > .gitignore
  git add .gitignore && git -c user.email=t@t -c user.name=t commit -q -m s
  "$BD" init --prefix abtest 2>&1 | tail -1 > /dev/null
  "$BD" create "ab-scenario-issue" -t task -p 1 2>&1 | tail -1 > /dev/null
  ID=$(grep -oE '"id":"[^"]+' .beads/issues.jsonl | head -1 | cut -d'"' -f4)
  "$BD" note "$ID" "GitHub: example/repo#1" >/dev/null 2>&1
  jq -r '.id + " notes=" + (.notes // "<MISSING>")' < .beads/issues.jsonl
  rm -rf "$TMP"
}

run_scenario
```

Run against `bd` built from `main` (pre-fix) and `bd` built from this branch — the former emits `notes=<MISSING>`, the latter `notes=GitHub: example/repo#1`.

## Related issues

- #3848 — canonical bug report (open, no PR linked before this one)
- #3970 — severity escalation (open, no PR linked before this one)
- #3962 — pre-commit hook defer fix (open, related but distinct)
- #3923 — sibling: pre-commit `git add -A` (closed without comment 2026-05-20)

### DIFF
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -3,12 +3,14 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/steveyegge/beads/internal/atomicfile"
@@ -54,19 +56,30 @@ func maybeAutoExport(ctx context.Context) {
 		return
 	}
 
-	// Load state and check throttle
+	// Resolve the export path early — the throttle decision below consults
+	// it (GH#3848: when the path is git-ignored we bypass throttle to keep
+	// the JSONL in sync with Dolt).
+	exportPath := config.GetString("export.path")
+	if exportPath == "" {
+		if globalFlag {
+			exportPath = "global-issues.jsonl"
+		} else {
+			exportPath = "issues.jsonl"
+		}
+	}
+	fullPath := filepath.Join(beadsDir, exportPath)
+
+	// Load state + interval.
 	state := loadExportAutoState(beadsDir)
 	interval := config.GetDuration("export.interval")
 	if interval == 0 {
 		interval = 60 * time.Second
 	}
-	if !state.Timestamp.IsZero() && time.Since(state.Timestamp) < interval {
-		debug.Logf("auto-export: throttled (last export %s ago, interval %s)\n",
-			time.Since(state.Timestamp).Round(time.Second), interval)
-		return
-	}
 
-	// Change detection via Dolt commit hash
+	// Change detection via Dolt commit hash. Cheap (single hash compare
+	// against stored state), so do it before the throttle — when there's no
+	// change, there's nothing to throttle. This also lets the throttle-bypass
+	// below act only when there's an actual pending write.
 	currentCommit, err := store.GetCurrentCommit(ctx)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: auto-export skipped: failed to get current commit: %v\n", err)
@@ -77,16 +90,18 @@ func maybeAutoExport(ctx context.Context) {
 		return
 	}
 
-	// Determine output path
-	exportPath := config.GetString("export.path")
-	if exportPath == "" {
-		if globalFlag {
-			exportPath = "global-issues.jsonl"
+	// Throttle decision is factored into a pure function so the branching
+	// logic is unit-testable without a store or filesystem. See
+	// shouldBypassThrottle for the policy.
+	if !shouldExport(state, interval) {
+		if shouldBypassThrottle(state, interval, isExportPathGitignored(ctx, beadsDir, fullPath)) {
+			debug.Logf("auto-export: bypassing throttle — export path is gitignored (GH#3848)\n")
 		} else {
-			exportPath = "issues.jsonl"
+			debug.Logf("auto-export: throttled (last export %s ago, interval %s)\n",
+				time.Since(state.Timestamp).Round(time.Second), interval)
+			return
 		}
 	}
-	fullPath := filepath.Join(beadsDir, exportPath)
 
 	// Run the export — memories are excluded from auto-export because they
 	// contain private agent context that must not reach git history (GH#3650).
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Cheers for carrying the refactor + test into #4063 with the trailer. Default-off makes the bypass logic moot — closing this out.

- Jeremy Longshore
intentsolutions.io

maphew:
Assessment after merged #4063 (`3988db7e0da1d5488e5ff972349f736466bec1c8`): partially addressed.

#4063 fixed the architectural/default-risk part: `export.auto` and `export.git-add` now default false, `bd init` is opt-in `[y/N]`, and the compatible refactor pieces were carried forward: export path resolution before throttle, cheap Dolt commit change detection, and `shouldExport` helper coverage.

The gitignored-path throttle bypass from #4061 (`isExportPathGitignored` / `shouldBypassThrottle` and related coverage) was intentionally not carried into #4063. That means explicit opt-in workflows with `export.auto=true` and gitignored `.beads/` can still see delayed/lost JSONL freshness inside the throttle window.

Recommended action: treat #4061 as partially addressed by #4063, with remaining policy tracked as a focused explicit-mode follow-up for gitignored/throttle behavior.

PR #4041, opened 2026-05-19
Add configurable JSONL import path

medhatgalal:
## What

Adds an `import.path` config key for implied JSONL imports. `bd` now reads the configured path, defaulting to `issues.jsonl`, when startup auto-import runs against an empty database, when `bd init --from-jsonl` imports local JSONL, and when `bd import` has no explicit input file.

The config is stored in `.beads/config.yaml`, matching `export.path`, so projects can set both paths to `beads.jsonl` and avoid export/import filename drift.

## Why

`export.path` was configurable, but empty-DB recovery and `bd init --from-jsonl` still hardcoded `.beads/issues.jsonl`. Projects that exported to another filename, such as `.beads/beads.jsonl`, could wipe a server DB and then fail to recover from the current export file.

## Review Notes

- `import.path` is YAML-backed, like `export.path`, because startup recovery and `bd init --from-jsonl` need it before/while opening storage.
- Explicit import filenames still win. `bd import backup.jsonl`, `bd import -i backup.jsonl`, and stdin are unchanged.
- `--global` import keeps its existing `global-issues.jsonl` behavior.
- Website CLI reference pages were regenerated after the first CI run flagged stale docs.

## Verification

- `scripts/pr-preflight.sh --search "import.path config jsonl auto import from-jsonl" --repo gastownhall/beads`
- `go test -tags gms_pure_go ./cmd/bd -run 'TestMaybeAutoImportJSONL|TestIsRecognizedConfigKey|TestConfigSetManyYamlKeyDetection|TestValidateKVKey'`
- `go test -tags gms_pure_go ./internal/config -run 'TestIsYamlOnlyKey'`
- `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestEmbedded(Import|Init)$'`
- `./scripts/generate-cli-docs.sh --check`
[...]

PR #4067, opened 2026-05-21
fix(hooks): skip bash injection on non-shell hook scripts (GH#4000)

### LINKED ISSUE #4000: injectHookSection appends bash to non-shell hook scripts (e.g. Python), causing SyntaxError on every commit
## Summary

When `bd init` runs in a repo where the user's existing git hooks are **not** shell scripts (in my case, Python via `#!/usr/bin/env python3`), `injectHookSection` appends the `# --- BEGIN BEADS INTEGRATION v1.0.4 ---` bash snippet to the bottom of each hook file. The host interpreter then tries to parse that bash as its own language and dies before reaching `__main__`.

For Python hooks the failure is immediate and total — every `git commit` errors out with `SyntaxError`.

## Version

`bd version 1.0.4 (ce242a879)`

## Repro

1. Have a global `core.hooksPath` pointing at a directory of Python hooks. Mine is a tiny LLM-driven `prepare-commit-msg` suggester with the standard symlink fanout (`pre-commit -> prepare-commit-msg`, etc.), shebang `#!/usr/bin/env python3`.
2. `cd` into a fresh repo and run `bd init`.
3. bd copies the user's hooks into `.beads/hooks/`, sets local `core.hooksPath=.beads/hooks`, and appends its integration block to 5 of them: `pre-commit`, `prepare-commit-msg`, `pre-push`, `post-checkout`, `post-merge`.
4. `git commit -m test`:

```
  File ".beads/hooks/pre-commit", line 222
    if command -v bd >/dev/null 2>&1; then
                  ^^
SyntaxError: invalid syntax
```

The appended block is valid bash, but the host file is `#!/usr/bin/env python3`. Python parses it as Python and crashes on the very first line of the bd block.

## Why the existing checks don't catch this

#3537 covers the pre-commit-framework case (block appended *below* an `exec` line, becoming unreachable). That's the inverse of this bug — there the bd block is dead code; here the bd block runs but the host interpreter rejects it. Neither case is detected by `injectHookSection` today because it only looks at content/markers, not the shebang.

## Suggested fix

Before appending, read the first line of the target file. If it's a shebang other than `sh`/`bash`/`zsh`/`dash`/`ksh`, do **not** append. Instead install a small bash wrapper at `.beads/hooks/<name>` that:

1. Execs the user's real hook (e.g. from the value of `git config --global core.hooksPath`, or wherever bd copied it from), forwarding args + exit code.
2. Then runs `bd hooks run <name> "$@"` with the same timeout/exit-3 handling that's currently in the appended block.

This is what I did manually as a workaround and it works cleanly — see the dispatcher pattern below.

## Workaround

Replace each affected hook with a one-line wrapper that `exec`s a shared dispatcher:

```bash
#!/usr/bin/env bash
exec "$(dirname "$0")/_dispatch.sh" pre-commit "$@"
```

Where `_dispatch.sh` finds the user's real hook via `git config --global --get core.hooksPath` (expanding `~`), runs it, then runs `bd hooks run <name>` preserving the existing timeout / exit-code-3 logic from the v1.0.4 integration block.

## Related

- #3537 — same code path (`injectHookSection`), inverse failure mode (unreachable vs. crashes interpreter)
- #3536 — `preservePreexistingHooks` behavior during migration
- #2344 — orphaned BEGIN marker

kevglynn:
## Summary
- Adds `isShellShebang()` helper to detect the hook file's interpreter from its shebang line before injecting bash syntax
- Non-shell hooks (Python, Ruby, Node, Perl) are skipped with a stderr warning instead of being corrupted with bash syntax
- Files without a shebang are still treated as shell (preserves existing behavior)
- Same guard applied to the hook migration path (`renderMigratedHookContent`)

Closes #4000

## Test plan
- [x] `make build` succeeds
- [x] `isShellShebang` unit tests cover all interpreter types (sh, bash, zsh, dash, ksh, ash, python, ruby, node, perl, empty, bare shebang)
- [x] Python hook is NOT modified by `installHooksWithOptions`
- [x] Bash/sh hooks still get section injected correctly
- [x] All existing hook tests pass (no regressions)

### DIFF
--- a/cmd/bd/hooks.go
+++ b/cmd/bd/hooks.go
@@ -1600,8 +1600,20 @@ func isShellShebang(content string) bool {
 	if len(parts) == 0 {
 		return true // bare #! with no interpreter = assume shell
 	}
-	// Handle "env" wrapper: #!/usr/bin/env bash → last token is interpreter
-	interpreter := filepath.Base(parts[len(parts)-1])
+	interpreter := filepath.Base(parts[0])
+	if interpreter == "env" {
+		interpreter = ""
+		for _, tok := range parts[1:] {
+			if strings.HasPrefix(tok, "-") {
+				continue // skip env options like -S, -i
+			}
+			interpreter = filepath.Base(tok)
+			break
+		}
+		if interpreter == "" {
+			return true // bare "#!/usr/bin/env" = assume shell
+		}
+	}
 	switch interpreter {
 	case "sh", "bash", "zsh", "dash", "ksh", "ash":
 		return true
--- a/cmd/bd/hooks.go
+++ b/cmd/bd/hooks.go
@@ -805,6 +805,10 @@ func installHooksWithOptions(hookNames []string, force bool, shared bool, chain 
 					newContent = "#!/usr/bin/env sh\n" + section
 				} else {
 					// Non-bd hook — inject section (preserving existing content)
+					if !isShellShebang(existingStr) {
+						fmt.Fprintf(os.Stderr, "warning: skipping hook injection for %q: non-shell interpreter detected; run 'bd hooks run %s' from a wrapper\n", hookName, hookName)
+						continue
+					}
 					newContent = injectHookSection(existingStr, section)
 				}
 			}
@@ -1579,6 +1583,31 @@ func runPrepareCommitMsgHook(args []string) int {
 // =============================================================================
 // Hook Helper Functions
 // =============================================================================
+
+// isShellShebang returns true if the content has no shebang (assumed shell)
+// or has a shebang pointing to a shell interpreter (sh, bash, zsh, dash, ksh, ash).
+// Used to guard against injecting bash syntax into non-shell hook scripts
+// (Python, Ruby, Node, Perl, etc.) which would cause SyntaxError. (GH#4000)
+func isShellShebang(content string) bool {
+	firstLine, _, _ := strings.Cut(content, "\n")
+	firstLine = strings.TrimSpace(firstLine)
+	if !strings.HasPrefix(firstLine, "#!") {
+		return true // no shebang = assume shell
+	}
+	shebang := strings.TrimPrefix(firstLine, "#!")
+	shebang = strings.TrimSpace(shebang)
+	parts := strings.Fields(shebang)
+	if len(parts) == 0 {
+		return true // bare #! with no interpreter = assume shell
+	}
+	// Handle "env" wrapper: #!/usr/bin/env bash → last token is interpreter
+	interpreter := filepath.Base(parts[len(parts)-1])
+	switch interpreter {
+	case "sh", "bash", "zsh", "dash", "ksh", "ash":
+		return true
+	}
+	return false
+}
 
 // isRebaseInProgress checks if a rebase is in progress.
 func isRebaseInProgress() bool {
--- a/cmd/bd/init_hooks_test.go
+++ b/cmd/bd/init_hooks_test.go
@@ -1085,6 +1085,120 @@ func TestHooksNeedUpdate(t *testing.T) {
 	}
 }
 
+// TestIsShellShebang verifies the shebang detection used to guard against
+// injecting bash syntax into non-shell hook scripts (GH#4000).
+func TestIsShellShebang(t *testing.T) {
+	tests := []struct {
+		name    string
+		content string
+		want    bool
+	}{
+		{"no shebang", "echo hello", true},
+		{"bash", "#!/bin/bash\necho hi", true},
+		{"env bash", "#!/usr/bin/env bash\necho hi", true},
+		{"sh", "#!/bin/sh\necho hi", true},
+		{"env sh", "#!/usr/bin/env sh\necho hi", true},
+		{"zsh", "#!/usr/bin/env zsh\necho hi", true},
+		{"dash", "#!/usr/bin/dash\necho hi", true},
+		{"ksh", "#!/usr/bin/env ksh\necho hi", true},
+		{"ash", "#!/bin/ash\necho hi", true},
+		{"python", "#!/usr/bin/env python3\nimport sys", false},
+		{"ruby", "#!/usr/bin/ruby\nputs 'hi'", false},
+		{"node", "#!/usr/bin/env node\nconsole.log('hi')", false},
+		{"perl", "#!/usr/bin/perl\nprint 'hi'", false},
+		{"empty content", "", true},
+		{"bare shebang", "#!\n", true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := isShellShebang(tt.content); got != tt.want {
+				t.Errorf("isShellShebang() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+// TestInstallHooksWithOptions_SkipsNonShellHook verifies that installHooksWithOptions
+// does NOT inject bash into a hook with a non-shell shebang (GH#4000).
+func TestInstallHooksWithOptions_SkipsNonShellHook(t *testing.T) {
+	tmpDir := newGitRepo(t)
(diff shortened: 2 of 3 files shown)


quad341:
@kevglynn — thank you for chasing down #4000 and fixing it at the root: detecting the interpreter from the shebang *before* injecting shell syntax is exactly right, and I really appreciate that you applied the guard to the migration path too, not just install. The test table walking sh/bash/zsh/dash/ksh/ash against python/ruby/node/perl, plus the end-to-end "Python hook stays untouched / bash hook still gets injected" cases, make the behavior easy to trust.

One small thing I'll handle on my side before landing so you don't have to spin: the helper currently keys off the *last* token of the shebang, so a valid shell hook with args — `#!/bin/bash -e` or `#!/usr/bin/env -S bash -e` — would get misread as non-shell and skipped. I'll adjust the parsing to read the interpreter from the front (skipping `env` options), add those arg cases, and add a migration-path test for the Codecov gap. Then I'll re-run and merge with your authorship intact. Great fix — thanks for the thorough test plan and the clean, additive diff.


---


quad341:
@kevglynn — thank you for chasing down #4000. Detecting the interpreter from the shebang before injecting shell syntax is exactly the right root-cause fix, and applying the guard to the migration path as well as install was the thorough call. The test table and the end-to-end install-path cases make the behavior easy to trust.

One edge case to close before landing: `isShellShebang` currently reads the **last** token of the shebang line, so valid shell hooks with interpreter args — `#!/bin/bash -e` or `#!/usr/bin/env -S bash -e` — get misread as non-shell (last token is `-e`) and are silently skipped. The fix is to read the **first** token instead, skipping `env` options to reach the real binary:

```go
// cmd/bd/hooks.go — replace the parts[len(parts)-1] block:
interpreter := filepath.Base(parts[0])
if interpreter == "env" {
    interpreter = ""
    for _, tok := range parts[1:] {
        if strings.HasPrefix(tok, "-") {
            continue // skip env options like -S, -i
        }
        interpreter = filepath.Base(tok)
        break
    }
    if interpreter == "" {
        return true // bare "#!/usr/bin/env" = assume shell
    }
}
```

Two small test additions to go with it:

1. Add to the `TestIsShellShebang` table in `cmd/bd/init_hooks_test.go`:
```go
{"bash with args", "#!/bin/bash -e\necho hi", true},
{"sh with args", "#!/bin/sh -eu\necho hi", true},
{"env -S bash with args", "#!/usr/bin/env -S bash -e\necho hi", true},
{"env -S python", "#!/usr/bin/env -S python3 -u\nimport sys", false},
{"bare env", "#!/usr/bin/env\n", true},
```

2. Add to `cmd/bd/migrate_hooks_test.go` (add `"os"` to imports):
```go
func TestRenderMigratedHookContent_SkipsNonShellHook(t *testing.T) {
    tmp := t.TempDir()
    hookPath := tmp + "/pre-commit"
    if err := os.WriteFile(hookPath, []byte("#!/usr/bin/env python3\nimport sys\n"), 0600); err != nil {
        t.Fatal(err)
    }
    op := hookMigrationWriteOp{
        HookName:   "pre-commit",
        HookPath:   hookPath,
        State:      "legacy_only",
        SourceKind: hookMigrationWriteFromHookFile,
        SourcePath: hookPath,
    }
    if _, err := renderMigratedHookContent(op); err == nil {
        t.Fatal("expected error for non-shell hook migration, got nil")
    }
}
```

Everything else in this PR is solid — the interpretation of the bug at the root, the consistent guard across install and migration paths, the thorough test coverage on the happy path. This lands once those changes are in. Thank you.

quad341:
**Code review — @kevglynn's hook injection shebang guard**

This is a well-designed fix for a real corruption bug. Before this PR, `bd hooks install` would inject bash syntax into Python/Ruby/Node/Perl hook scripts, producing invalid files. The fix correctly guards both the installation path and the migration path.

**What's correct:**
- `isShellShebang()` handles the common cases right: no shebang → assume shell, `#!/bin/sh`, `#!/usr/bin/env bash`, and similar all return `true`; Python/Ruby/Node/Perl return `false`
- The `filepath.Base()` + `switch` approach is idiomatic and clean
- Migration path returns an error rather than silently skipping — correct, since migration is a one-time operation where failure should be surfaced
- 14 unit tests cover all interpreter types including edge cases (bare `#!`, empty content)

**One LOW-severity edge case (not a blocker):**

For shebangs with arguments, e.g. `#!/usr/bin/env bash -x`, the interpreter extraction uses `parts[len(parts)-1]`:
```go
interpreter := filepath.Base(parts[len(parts)-1])
```
This yields `-x` as the interpreter, which doesn't match any shell name, so the function returns `false` — incorrectly classifying a bash-with-debug-flag hook as non-shell. The result is still *safe* (injection is skipped with a warning rather than corrupting the file), but the hook won't get the bd section installed. Shebangs with env + arguments are rare in practice.

A future hardening: when `parts[0]` contains `env`, treat `parts[1]` as the interpreter name regardless of trailing args.

All CI checks pass. The edge case is worth a follow-up issue, but does not block this fix.

---

quad341:
**Follow-up: one edge case to address before landing**

Revisiting after a deeper look — the `isShellShebang` logic has a real regression I should have flagged as a blocker in my earlier comment.

The current implementation takes the **last** shebang token as the interpreter:

```go
interpreter := filepath.Base(parts[len(parts)-1])
```

For `#!/bin/bash -e`, that yields `"-e"` — which doesn't match any shell name, so `isShellShebang` returns `false`. A real shell hook with interpreter args will not receive the `bd` section and the behavior will silently fail to inject.

**Exact fix needed in `cmd/bd/hooks.go`** — replace the one-liner with first-token logic:

```go
	interpreter := filepath.Base(parts[0])
	if interpreter == "env" {
		interpreter = ""
		for _, tok := range parts[1:] {
			if strings.HasPrefix(tok, "-") {
				continue // skip env options like -S, -i
			}
			interpreter = filepath.Base(tok)
			break
		}
		if interpreter == "" {
			return true // bare "#!/usr/bin/env" = assume shell
		}
	}
```

**Add 5 table entries to `TestIsShellShebang` in `cmd/bd/init_hooks_test.go`:**
```go
{"bash with args", "#!/bin/bash -e\necho hi", true},
{"sh with args", "#!/bin/sh -eu\necho hi", true},
{"env -S bash with args", "#!/usr/bin/env -S bash -e\necho hi", true},
{"env -S python with args", "#!/usr/bin/env -S python3 -u\nimport sys", false},
{"bare env", "#!/usr/bin/env\n", true},
```

**Add one migration-path test to `cmd/bd/migrate_hooks_apply_test.go`:**
```go
func TestRenderMigratedHookContent_SkipsNonShellHook(t *testing.T) {
	dir := t.TempDir()
	src := filepath.Join(dir, "pre-commit")
	if err := os.WriteFile(src, []byte("#!/usr/bin/env python3\nimport sys\n"), 0600); err != nil {
		t.Fatal(err)
	}
	op := hookMigrationWriteOp{
		HookName:   "pre-commit",
		SourceKind: hookMigrationWriteFromHookFile,
		SourcePath: src,
	}
	if _, err := renderMigratedHookContent(op); err == nil {
		t.Fatal("expected error for non-shell migration source, got nil")
	}
}
```

Sorry for not flagging this as a blocker initially — the core fix is correct and the CI pass gives it a green baseline, but those interpreter-arg cases are real and worth locking in before merge. Everything else stands as approved.

---

kevglynn:
Good catch @quad341 — you're right, `#!/bin/bash -e` would have matched `-e` as the interpreter with the last-token approach. Switched to first-token parsing with proper `env` flag skipping, and added your 5 edge-case tests. All passing locally.

maphew:
closed as superseded. The useful part landed through #4041, and current main has a better version of the hook fallback. However I did that work in part because of what you raised here @kevglynn, but failed to maintain the credit in 4041; since rectified in #4124.

quad341:

PR #4072, opened 2026-05-21
fix(init): mark server-mode Dolt dirs compatible

julianknutsen:
## Summary

- write the existing `.bd-dolt-ok` compatibility marker after successful `bd init --server` when the configured Dolt data dir contains a local `.dolt/` repository
- keep marker ownership in `internal/doltserver`, next to the existing pre-0.56 detection and recovery logic
- add helper-level edge case coverage plus an init regression test for externally managed server mode

This supersedes the Gas City wrapper workaround in gastownhall/gascity#2108; the marker should be emitted by Beads itself.

## Tests

- `./scripts/test.sh -run 'TestMarkDoltDirCompatible|Test(IsPreV56DoltDir|EnsureDoltInit|RecoverPreV56DoltDir)' ./internal/doltserver`\n- `./scripts/test.sh -run '^TestInitServerModeWritesDoltCompatibilityMarker$' ./cmd/bd`\n- `go vet ./...`\n- `git diff --check`\n\nI also attempted `make test`; local full-suite baseline still fails outside this patch with broad existing/environmental failures, including shared server port 3308 already in use and several unrelated cmd/domain/tracker expectations.

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -1232,6 +1232,12 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 			fmt.Fprintf(os.Stderr, "Warning: failed to close database: %v\n", err)
 		}
 
+		if initServerMode {
+			if err := doltserver.MarkDoltDirCompatible(storagePath); err != nil && !quiet {
+				fmt.Fprintf(os.Stderr, "Warning: failed to write Dolt compatibility marker: %v\n", err)
+			}
+		}
+
 		// WARNING: DO NOT remove, delete, or modify files inside Dolt's .dolt/
 		// directory — including noms/LOCK files. These are Dolt-internal files.
 		// Removing them WILL cause unrecoverable data corruption and data loss.
--- a/cmd/bd/init_test.go
+++ b/cmd/bd/init_test.go
@@ -1643,6 +1643,53 @@ func TestInitDoltMetadataNoGit(t *testing.T) {
 	}
 }
 
+func TestInitServerModeWritesDoltCompatibilityMarker(t *testing.T) {
+	skipIfNoDolt(t)
+	saveAndRestoreGlobals(t)
+	ensureCleanGlobalState(t)
+	dbPath = ""
+	store = nil
+
+	tmpDir := t.TempDir()
+	t.Chdir(tmpDir)
+
+	beadsDir := filepath.Join(tmpDir, ".beads")
+	doltDir := filepath.Join(beadsDir, "dolt")
+	if err := os.MkdirAll(beadsDir, 0o700); err != nil {
+		t.Fatalf("creating beads dir: %v", err)
+	}
+	if err := os.MkdirAll(filepath.Join(doltDir, ".dolt"), 0o750); err != nil {
+		t.Fatalf("creating simulated server data dir: %v", err)
+	}
+
+	database := uniqueTestDBName(t)
+	t.Cleanup(func() {
+		dropTestDatabase(database, testDoltServerPort)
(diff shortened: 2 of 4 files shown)


quad341:
Thanks @julianknutsen — this is exactly the right fix. Moving the `.bd-dolt-ok` marker write into `internal/doltserver`, right next to the existing pre-0.56 detection and recovery logic, is the correct home for it, and having Beads emit the marker itself instead of leaning on the gascity#2108 wrapper is a real cleanup.

I especially appreciated the care in the helper: idempotent, preserves a pre-existing marker rather than clobbering it, no-ops cleanly when there's no local `.dolt/`, and fails loudly on the genuinely weird cases. The branch coverage in `TestMarkDoltDirCompatible` plus the real `init --server --external` regression test — asserting the marker actually lands at `.beads/dolt/.bd-dolt-ok` — is the kind of end-to-end check that makes this easy to trust.

CI is fully green across the embedded-Dolt matrix, lint, and formatting, the base is clean against `main`, and there's nothing for me to touch. Merging as-is. Nice, focused contribution.


---


quad341:
**Code review — fix(init): mark server-mode Dolt dirs compatible**

The PR correctly extracts `MarkDoltDirCompatible` into a reusable exported function in `internal/doltserver/doltserver.go` and calls it during server-mode init. This fills a gap: previously, `bd init --server` on a pre-existing Dolt directory (e.g., when migrating from standalone Dolt to bd server mode) wouldn't write the `.bd-dolt-ok` marker, so subsequent `bd doctor` would report the directory as incompatible.

**What's correct:**
- The function is idempotent — existing markers are preserved (doesn't overwrite a custom `ok\n` file)
- `.dolt` file-not-directory case returns an error
- Empty `doltDir` is an input error
- Missing `.dolt/` directory returns nil (no-op) — safe for the general case
- The init.go call uses `!quiet` to gate the warning, consistent with the rest of init's error output pattern
- 5 unit tests + 1 integration test cover all branches including the marker-as-directory edge case

All CI checks pass.

---

quad341:
Thanks @julianknutsen — this is a genuinely clean fix. Pulling the `.bd-dolt-ok` marker write into `MarkDoltDirCompatible` and parking it right next to the pre-0.56 detection/recovery logic is exactly the right home for it, and having Beads emit the marker itself instead of leaning on the gascity#2108 wrapper removes a real layering wart.

What stood out: the helper is carefully idempotent — it preserves an existing marker instead of clobbering it, no-ops cleanly when there's no local `.dolt/`, and fails loudly on the genuinely weird cases (`.dolt` as a file, marker as a directory). And the test layering is great: per-branch unit coverage plus a real `init --server --external` regression test that asserts the marker actually lands on disk. That end-to-end check is what makes this trivial to trust.

CI is green across the whole embedded-Dolt matrix, lint, formatting, and cross-version smoke tests, and the base is clean against `main`. Nothing for me to touch — merging as-is. Really nice, focused contribution.


---


quad341 (approved):
Full ensemble review: auto-merge. All CI green, MERGEABLE=CLEAN. Merging.

julianknutsen:
Thanks for this contribution, @julianknutsen. Your original fix correctly closed the gap where `bd init --server --external` could leave a healthy local Dolt dir without `.bd-dolt-ok`, which then showed up later as a false pre-v0.56 warning.

We merged this via #4079, which preserves your original commit authorship and adds the maintainer review fixups that came out of adoption review. We created a new PR because we needed to add commits but couldn't push directly to your branch.

Your commit is in the merged history, and we appreciate the contribution.

quad341:

Issue #3474, opened 2026-04-25
bd export: memory lines emitted in non-deterministic order (Go map iteration)


mgannholm:
## Summary

`bd export` and the auto-export pre-commit hook write `_type:"memory"` lines in
non-deterministic order. Issue lines come out stably, but every export shuffles
the memory tail, producing churn in committed `issues.jsonl` even when no
memories changed.

Visible symptom: every commit on a beads-tracked repo includes an unrelated
`issues.jsonl` diff that just permutes the trailing `_type:"memory"` lines.
After enough commits this is just noise; in worktree workflows it also fights
hand-staged sorts.

## Root cause

`cmd/bd/export.go:197` iterates `allConfig` (a `map[string]string` from
`store.GetAllConfig(ctx)`) directly:

```go
fullPrefix := kvPrefix + memoryPrefix
for k, v := range allConfig {
    if !strings.HasPrefix(k, fullPrefix) {
        continue
    }
    userKey := strings.TrimPrefix(k, fullPrefix)
    record := map[string]string{
        "_type": "memory",
        "key":   userKey,
        "value": v,
    }
    data, err := json.Marshal(record)
    ...
}
```

Go map iteration order is intentionally randomized, so the memory lines come
out in a different order on every run.

The same pattern is duplicated in cmd/bd/export_auto.go:218 (the auto-export
path used by the pre-commit hook), so the bug fires on every commit, not just
manual bd export.

Issue lines aren't affected because store.SearchIssues returns a sorted slice
(consistent with prior determinism work — #859, #1228, #2368) — so this only
shows up after memories were added to the export (#? — the _type discriminator
was added in #3271).

Why existing tests miss it
Determinism tests in this area (e.g. the SQLite→Dolt fidelity work in #2085 /
#2125) check issue ordering but don't assert byte-stable output across
independent export runs of the same DB state when memories are present.
A regression test would be: seed N≥3 memories, run bd export twice, assert
the byte-identical output.

Suggested fix
Sort the memory keys before iteration, in both files. Minimal diff for
export.go:

 fullPrefix := kvPrefix + memoryPrefix
-for k, v := range allConfig {
-    if !strings.HasPrefix(k, fullPrefix) {
-        continue
-    }
-    userKey := strings.TrimPrefix(k, fullPrefix)
+var memKeys []string
+for k := range allConfig {
+    if strings.HasPrefix(k, fullPrefix) {
+        memKeys = append(memKeys, k)
+    }
+}
[...]

PR #4086, opened 2026-05-22
test(export): assert byte-stable JSONL across all record types

maphew:
## Summary

Follow-up to (3787) (memory-ordering phantom diffs). `TestExportMemoryDeterminism` (GH #3474) proves only *memory* lines serialize in a stable order. This generalizes that invariant: running `bd export` repeatedly over unchanged data must produce byte-identical output for **every** record type, not just memories.

Go's randomized map iteration is the usual source of phantom git diffs, so any collection serialized without a *total* order (a sort key ending in a unique tiebreaker) would surface here — catching regressions before they ship, rather than spot-fixing per record type.

## What the test does

`TestExportByteStabilityAllRecordTypes`:

- Seeds 5 issues at the **same priority and creation time**, forcing the export `ORDER BY priority, created_at DESC, id ASC` to fall through to the `id` tiebreaker — exercising the total-order property the guarantee depends on.
- Gives each issue multiple **labels** (added out of alphabetical order), **parent-child dependencies**, and **comments**.
- Adds several **memories** keyed out of sorted order.
- Exports 5×, asserting every run is byte-identical.
- Sanity-checks each record type is actually present, so a future change can't make the stability assertion pass vacuously.

## Notes for reviewers

- `AddIssueComment` (writes the `comments` table the export reads) is used rather than `AddComment` (records a `commented` event only).
- Parent-child deps are used to stay on the issues-table dependency path (avoids the wisp cycle-check route).
- Verified passing against `main` with the Dockerized Dolt test server (`-tags gms_pure_go`).

### DIFF
--- a/cmd/bd/export_test.go
+++ b/cmd/bd/export_test.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/steveyegge/beads/internal/testutil"
 	"github.com/steveyegge/beads/internal/types"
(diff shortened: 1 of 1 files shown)


quad341:
@maphew — this is exactly the kind of generalization that prevents future bug classes from silently shipping. Forcing 5 issues to the same priority and timestamp exercises the `id ASC` tiebreaker, and verifying each record type is actually present in the output closes the 'vacuously passes' failure mode. The use of `AddIssueComment` (vs `AddComment`) shows care about the difference between the data table and the event log. CI is all green, test-only, 0 deletions. Merging.

---
**Reviewers**: Qwen (local) — ok · Codex (gpt-5.5) — ok
**Synthesis**: claude-sonnet-4-6

---

quad341:

PR #4088, opened 2026-05-22
fix(export): guard auto-export against richer JSONL overwrite

### LINKED ISSUE #4069: auto-export uses default filter, drops infra/ephemeral rows that exist in DB (89% data loss in our workspace)
## Summary

`export.auto=true` (default) rewrites `.beads/issues.jsonl` after every write command using the default `bd export` filter. That filter silently excludes rows that *do exist in the local Dolt DB* (molecule/agent/memory and other infra-flagged rows). On every auto-export cycle the on-disk JSONL shrinks toward the filtered subset. If `.beads/issues.jsonl` is ever the recovery source (`bd bootstrap`, fresh clone, restored backup), those infra rows are permanently gone.

This is **not** about cross-session drift (#3931) or empty-DB deletion (#4033). The DB is populated and correct; the filter is the culprit.

## Repro

In a workspace with mixed issue types (production beads + agent/molecule/memory rows):

```bash
bd export       | wc -l   # 234
bd export --all | wc -l   # 2141
```

Type breakdown of the 1907-row delta in our workspace:

| issue_type | _type | only-with-`--all` |
|-----------|-------|------------------|
| molecule  | issue | 1874 |
| task      | issue | 25 |
| (null)    | memory | 7 |
| agent     | issue | 3 |

Same DB. Both commands read the same source of truth. Default filter just drops 89% of it on write.

## Impact

Auto-export fires after **every** write (throttled). Each write replaces JSONL with the filtered subset. If anything later reads JSONL as truth — `bd bootstrap` on a fresh clone, recovery from git, importing into another sql-server — those 1907 rows are gone.

We hit this twice in production:

1. **strateg rig session lost ~20 issues** (25 → 5 lines in JSONL). Rescued via `git checkout .beads/issues.jsonl`. Bead [hq-rjlo] tracks the cascade — strateg DB went empty because the bootstrap source was already eroded.
2. **dolt server lost 6 DBs** including all our workflow/product DBs. Recovery from JSONL succeeded only for the two rigs that had been manually `bd export --all`'d recently. Bead [project_dolt_corruption_20260518] in our memory.

## Suggested fix

One of:

1. **Make auto-export call `--all` semantically.** Backwards-compatible if the user only ever reads JSONL via `bd` (which doesn't need the filter at import time). The filter exists for human-readable `bd export` CLI output; auto-export is a backup/sync mechanism and shouldn't filter.

2. **Add `export.include-infra` config key (default `true`).** Explicit, debuggable. Users who want filtered JSONL for git-friendly diffs can flip it.

3. **Defensive: never let auto-export shrink JSONL.** If new line count < old line count, skip the write and log a warning. Catches not just this bug but also race conditions like #3931.

Happy to send a PR for (1) or (2) if there's a preferred direction. (3) seems valuable independently as a safety net regardless.

## Related

- #4062 — reconsidering auto-export defaults overall. This issue is one of the concrete fallouts.
- #3931 — cross-session drift; different mechanism, same on-disk symptom.
- #4033 — empty-DB deletes JSONL; different mechanism, same on-disk symptom.

## Environment

- `bd version 1.0.4 (Homebrew)`
- Embedded-mode Dolt
- Multi-rig gastown workspace (15 git repos, each with own `.beads/`)
- macOS Darwin 24.6.0

Local mitigation we deployed: pre-commit hook that rejects commits where staged `.beads/issues.jsonl` is shorter than current `bd export --all`. Catches CLI-driven erosion. Does **not** catch erosion via direct write from inside the bd binary (no git hook fires).

maphew:
## Summary
- add an auto-export shrink guard that refuses to overwrite existing JSONL containing records outside auto-export scope
- classify memories, infra/template/ephemeral issues, and unknown record types as richer than viewer-scoped auto-export
- document the user-visible guard in the changelog

Fixes #4069

## Test plan
- `go test -tags gms_pure_go ./cmd/bd -run 'TestGuardAutoExportOverwrite|TestShouldExport|TestShouldRunPostCommandAutoExport' -count=1`\n- `go test -tags gms_pure_go ./cmd/bd` (fails in existing/environment-sensitive tests: missing ICU headers for nested build, Dolt test DB setup failures, import/comment regressions)\n\n_codex-gpt-5.5-medium on behalf of matt wilkie_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **JSONL auto-export shrink guard** - when opt-in `export.auto` is enabled,
+  auto-export now refuses to overwrite an existing `.beads/issues.jsonl` that
+  contains records outside the auto-export scope (memories, infrastructure
+  beads, templates, ephemeral wisps, or unknown record types). This prevents
+  a viewer/interchange refresh from silently replacing a richer JSONL file
+  with the filtered auto-export subset. ([#4069](https://github.com/gastownhall/beads/issues/4069))
 - **`bd dolt status` reports externally-managed local servers truthfully** - when a rig is configured as `dolt_mode: server` pointing at a local host but `dolt.auto-start: false` (so an orchestrator or systemd owns the sql-server lifecycle), `bd dolt status` previously said `not running` because no PID file existed. It now SQL-probes the configured endpoint, matching the path already used for non-local hosts, and reports `running (external)` with host/port/database/version when the server answers. **JSON output shape change**: on affected rigs, `bd dolt status --json` now emits `{"running": true, "mode": "external", ...}` instead of `{"running": false, "pid": 0, ...}`. Automation that parsed the old `running:false` as a "needs restart" sentinel should switch to checking `running` directly. (be-0eyj, [#3550](https://github.com/gastownhall/beads/pull/3550))
 
 ## [1.0.4] - 2026-05-07
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -145,7 +146,7 @@ func shouldExport(state *exportAutoState, interval time.Duration) bool {
 
 // exportToFile atomically exports issues + memories to the given file path.
 // Writes to a temp file first, then renames into place so readers never see
-// a partial or truncated export. Used by both `bd export -o` and auto-export.
+// a partial or truncated export. Used by auto-export.
 func exportToFile(ctx context.Context, path string, includeMemories bool) (issueCount, memoryCount int, err error) {
 	w, err := atomicfile.Create(path, 0o644)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


quad341:
@maphew — thank you for tackling #4069 and building the right guard for it. The shrink-guard approach — scan the existing JSONL and refuse to overwrite when out-of-scope records are present — is exactly correct, and the error message (with per-category counts) tells users what they need to know. The three unit tests covering allow/block/memory-included give good confidence in the guard logic.

I've rebased onto `main` on the maintainer side — the only conflict was additive: PR #4087 added `shouldSkipEmptyAutoExport` and `countIssueRecordsInJSONL` between `shouldExport` and `exportToFile`, which moved `exportToFile` and refactored the infra-type loop into `autoExportFilter`. The rebase adapts the guard to build its `infraTypeSet` inline (from `store.GetInfraTypes` / `DefaultInfraTypes`) rather than from the refactored loop, so both the filter and the guard stay consistent. The tests pass clean.

Fixed branch: `quad341/beads:maintainer/pr-4088-shrink-guard` — landing that with your authorship attributed.

---

quad341:
**Follow-up**: PRs #4089 and #4090 landed on  after the first rebase, so I've re-rebased the maintainer branch onto current main — the cherry-pick applied cleanly (auto-resolved as additive disjoint changes). Maintainer PR created at #4099. Nothing to do on your side, @maphew — your authorship is preserved.

quad341:

PR #4091, opened 2026-05-22
fix: guard JSONL auto import/export edge cases

maphew:
## Summary
- skip JSONL auto-import and auto-export entirely in server mode
- skip pre-commit JSONL auto-export unless the pending commit already stages .beads paths
- let prune/purge that actually delete rows record intentional empty JSONL removal, including gitignored opt-in JSONL workflows

## Validation
- go test -tags gms_pure_go -run 'TestMaybeAutoImportJSONL|TestPreCommitHasStagedBeadsFiles|TestCommandAllowsEmptyAutoExport|TestShouldRunPostCommandAutoExport|TestShouldExport|TestGitAddFile' ./cmd/bd
- go test -tags gms_pure_go -run '^$' ./cmd/bd
- git diff --check

Full package note: go test -tags gms_pure_go ./cmd/bd was attempted, but this local environment hit unrelated integration prerequisites: missing ICU headers for an inner CGO build, a shared Dolt server port conflict, and existing integration test failures outside this change.

Fixes follow-up scope for #4038, #3962, and #3944.

_codex-gpt-5.5-medium on behalf of matt wilkie_

PR #4090, opened 2026-05-22
Guard auto-export against JSONL-only records

maphew:
## Summary
- compare existing JSONL issue IDs against the local Dolt store before auto-export
- refuse auto-export with a machine-detectable warning when JSONL-only issue records would be dropped
- add regression coverage preserving the existing JSONL file

## Validation
- `go test -tags gms_pure_go ./cmd/bd -run 'TestCountIssueRecordsInJSONL|TestAutoExportSkips'`\n\nRefs gastownhall/beads#3931.\n\n_codex-gpt-5.5-medium on behalf of maphew_

PR #4092, opened 2026-05-22
fix(export): fail on auto-export git add errors

### LINKED ISSUE #3970: bd warns but exits 0 when auto-export 'git add' fails; staged jsonl can ship stale state

maphew:
## Summary
- surface opted-in auto-export git-add failures through the CLI error path
- leave auto-export state unsaved after staging failure so the next run can retry
- add regressions for ignored paths and locked git indexes

Fixes #3970.

## Validation
- `go test -tags gms_pure_go ./cmd/bd -run 'Test(GitAddFile_CapturesLockedIndexFailure|AutoExportGitAddFailureExitsNonZero|GitAddFile_CapturesStderrOnFailure)' -count=1` -> `ok github.com/steveyegge/beads/cmd/bd 89.991s`
- `git diff --check HEAD~1..HEAD` -> passed

### DIFF
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -34,31 +34,31 @@ const gitAddTimeout = 5 * time.Second
 
 // maybeAutoExport writes a git-tracked JSONL file if enabled and due.
 // Called from PersistentPostRun after auto-backup.
-func maybeAutoExport(ctx context.Context, serverMode, allowEmptyOverwrite bool) {
+func maybeAutoExport(ctx context.Context, serverMode, allowEmptyOverwrite bool) error {
 	if serverMode {
 		debug.Logf("auto-export: skipping — server mode\n")
-		return
+		return nil
 	}
 
 	// Skip when running as a git hook to avoid re-export during pre-commit.
 	if os.Getenv("BD_GIT_HOOK") == "1" {
 		debug.Logf("auto-export: skipping — running as git hook\n")
-		return
+		return nil
 	}
 
 	if !config.GetBool("export.auto") {
-		return
+		return nil
 	}
 	if store == nil {
-		return
+		return nil
 	}
 	if lm, ok := storage.UnwrapStore(store).(storage.LifecycleManager); ok && lm.IsClosed() {
-		return
+		return nil
 	}
 
 	beadsDir := beads.FindBeadsDir()
 	if beadsDir == "" {
-		return
+		return nil
 	}
 
 	// Resolve the export path before throttle/check detection so all decisions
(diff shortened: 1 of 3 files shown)


quad341:
Thanks @maphew — this is exactly the right fix for #3970. Converting `maybeAutoExport` to return an `error`, wiring it through the cobra PostRun hook, and (especially) holding `saveExportAutoState` until after a successful git-add are all the right calls — that last one means a locked index or ignored path leaves the state file untouched so the next run cleanly retries.

The new tests are nicely targeted: the locked-index case nails the low-level error wrapping, and the end-to-end test catches the exit-code surfacing plus the "don't poison the state file on failure" invariant in one shot.

I rebased onto current `main` and resolved two semantic conflicts on my side:
1. PR #4091 added `serverMode, allowEmptyOverwrite` parameters to `maybeAutoExport` — merged both changes, keeping your error return and the new parameters.
2. PR #4090 added bare `return` statements in the `allowEmptyOverwrite` guards that don't compile under your new `error` signature → converted those to `return nil`.
   
All 4 targeted tests pass (`TestGitAddFile_SkipsWhenIndexLocked`, `TestGitAddFile_CapturesStderrOnFailure`, `TestAutoExportGitAddFailureExitsNonZero`, and the broader suite except a pre-existing `TestValidateCheck_NoBeadsDir` failure that also fails on clean `main`).

Fixed branch: `quad341/beads:maintainer/pr-4092-auto-export-errors` (commit `8b6efd7`)

Queued for merge. Human merge required (stopgap policy on this repo).

---

quad341:

PR #4085, opened 2026-05-22
Stamp auto-import attempts for unchanged JSONL

Jacob-qd:
## Summary\n- Add a local issues.jsonl auto-import attempt stamp keyed by size and mtime\n- Skip repeated auto-import attempts for the same unchanged JSONL after success or failure\n- Allow retry when issues.jsonl changes\n\n## Verification\n- go test -tags gms_pure_go ./cmd/bd -run 'TestMaybeAutoImportJSONL'\n- make build\n- Manual production check after installing candidate bd locally: repeated 'bd status' in /Users/admin/gt no longer prints auto-importing ... into empty database; both runs report 517 issues directly\n\nContext: Homebrew bd 1.0.4 repeatedly attempted auto-import on every command in Gas Town, causing bd/mail/convoy slowness and stale JSONL risk.

PR #4097, opened 2026-05-22
fix(auto-import): stamp attempts to prevent repeated import of unchanged JSONL (maintainer cherry-pick from PR #4085)

quad341:
## Summary

Closes #4085 via maintainer cherry-pick (contributor fork write-access not available).

- **Bug**: `maybeAutoImportJSONL` would repeatedly attempt to import the same unchanged JSONL on every `bd` invocation, paying the full import cost even when the file hadn't changed since the last (successful or failed) import.
- **Fix**: Stamp size+mtime after every import attempt (success or failure). On subsequent invocations, if the JSONL file matches the stamp, skip the import. A changed JSONL bypasses the stamp and re-triggers import.
- **Tests**: 4 new tests: `FailedImportStampPreventsRepeatedImport`, `SuccessStampPreventsRepeatedImport`, `ChangedJSONLBypassesSuccessStamp`, `UsesConfiguredImportPath`. All 7 auto-import tests pass.

Rebased onto main; additive comment conflict with PR #4087 resolved. Early-return comment clarified to reflect stamp-on-failure semantics.

---
Original contributor: @Jacob-qd — exactly the right fix for the auto-import clobber window.

### DIFF
--- a/cmd/bd/auto_import_upgrade.go
+++ b/cmd/bd/auto_import_upgrade.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


quad341:
**Root cause of red CI:** the branch was based on a `main` where `maybeAutoImportJSONL` took 3 args. Since then `main` gained a 4th `serverMode bool` param (early-return guard). CI builds the PR *merged with main*, so it saw main's 4-arg production signature but the cherry-picked stamp tests still called it with 3 args — `not enough arguments in call to maybeAutoImportJSONL` in the `CGO_ENABLED=0` test-binary compile (and the macOS/Ubuntu test jobs).

**Fix:**
1. Rebased the single stamp commit onto current `main`. The `serverMode` early-return and the stamp logic are orthogonal and composed with no conflicts — the production function now has both.
2. Updated the 3 new stamp tests to pass `serverMode=false` (they exercise the import path; `serverMode=true` is already covered by the existing `ServerModeFallback` test). Commit `158a136a2`.

**Verified locally** (the exact failing checks):
- `CGO_ENABLED=0 go build -tags gms_pure_go ./cmd/bd` — OK
- `CGO_ENABLED=0 go test -tags gms_pure_go -c ./cmd/bd` — OK (was the failing step)
- `go test -run '^TestMaybeAutoImportJSONL' ./cmd/bd` — all pass, incl. the 3 new stamp tests

Ready for another pass.

PR #4098, opened 2026-05-22
fix(export): fail on auto-export git add errors (maintainer cherry-pick from PR #4092)

quad341:
## Summary

Closes #4092 via maintainer cherry-pick (contributor fork write-access not available).

- **Bug**: `maybeAutoExport` swallowed git-add errors and still wrote the `LastDoltCommit` state, causing the next run to skip retrying a failed stage.
- **Fix**: Change `maybeAutoExport` signature to return `error`. Propagate git-add failures through the cobra `PostRun` hook via `FatalError`. Do not write state on failure, so the next run sees a dirty state and retries.
- **Tests**: `TestGitAddFile_CapturesLockedIndexFailure` (low-level error wrapping), `TestAutoExportGitAddFailureExitsNonZero` (end-to-end: non-zero exit + user-visible error string + no state file written on failure).

Conflict with PR #4091 (serverMode/allowEmptyOverwrite params) and PR #4090 (bare returns) resolved during cherry-pick.

---
Original contributor: @maphew — exactly the right fix for #3970. Converting `maybeAutoExport` to return `error` and holding off on `saveExportAutoState` until after a successful git-add are the right calls.

### DIFF
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -142,7 +142,18 @@ func maybeAutoExport(ctx context.Context, serverMode, allowEmptyOverwrite bool) 
 	// git repo (standalone BEADS_DIR flow), or when export.git-add is false.
 	if config.GetBool("export.git-add") && !config.GetBool("no-git-ops") && isGitRepo() {
 		if err := gitAddFile(fullPath); err != nil {
-			return fmt.Errorf("auto-export: git add failed: %w", err)
+			// A gitignored .beads/ is a supported workflow: the JSONL export is
+			// written for local use / Dolt sync but intentionally left untracked,
+			// so `git add` can never stage it. That is not a failure and must not
+			// abort the command. Only genuine staging failures — e.g. a locked
+			// index, the real GH#3970 repro — are fatal. See be-1wzr / PR #4098
+			// for the contract decision; the gitignored-success path is regressed
+			// by TestEmbeddedPrune/prune_last_issue_removes_gitignored_auto_export.
+			if pathIsGitIgnored(fullPath) {
+				debug.Logf("auto-export: %s is gitignored; skipping git add\n", fullPath)
+			} else {
+				return fmt.Errorf("auto-export: git add failed: %w", err)
+			}
 		}
 	}
 
@@ -493,6 +504,26 @@ func gitAddFile(path string) error {
 		return err
 	}
 	return nil
+}
+
+// pathIsGitIgnored reports whether path is excluded by a .gitignore rule in
+// the enclosing repo. It is used by auto-export to classify a gitAddFile
+// failure: a gitignored .beads/ is a supported workflow (the JSONL export is
+// intentionally untracked), so failing to stage it must be benign, whereas a
+// real staging failure such as a locked index stays fatal (be-1wzr / PR #4098).
+//
+// `git check-ignore -q` exits 0 when the path is ignored, 1 when it is not,
+// and 128 on error. Anything other than a clean exit 0 is treated as "not
+// ignored" so genuine failures are never masked. The git-hook env is scrubbed
+// and the command runs from the file's directory, mirroring gitAddFile so the
+// ignore decision matches the repo `git add` actually consulted.
+func pathIsGitIgnored(path string) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), gitAddTimeout)
+	defer cancel()
+	cmd := exec.CommandContext(ctx, "git", "check-ignore", "-q", "--", path)
+	cmd.Dir = filepath.Dir(path)
+	cmd.Env = scrubGitHookEnv(os.Environ())
+	return cmd.Run() == nil
 }
 
 func gitIndexLockPath(path string, env []string) (string, error) {
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -34,31 +34,30 @@ const gitAddTimeout = 5 * time.Second
 
 // maybeAutoExport writes a git-tracked JSONL file if enabled and due.
 // Called from PersistentPostRun after auto-backup.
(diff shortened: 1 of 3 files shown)


quad341:
**Investigator findings (CI red): this PR conflicts with a contract already on `main`**

Thanks @maphew and @quad341 — the core of the #3970 fix is right (return `error` from `maybeAutoExport`, and hold `saveExportAutoState` until after a successful `git add`). But the single failing check, `Test (Embedded Dolt Cmd 18/20)`, is not flaky. It's a deterministic conflict that needs one small decision before this can go green.

### What fails

`TestEmbeddedPrune/prune_last_issue_removes_gitignored_auto_export` (`cmd/bd/prune_embedded_test.go:112`), already on `main`, sets up:

- `.gitignore` = `.beads/`
- `export.auto=true`, `export.git-add=true`

then runs `bd create` / `bd close`. With this PR, that first `bd create` exits non-zero: auto-export's `git add .beads/issues.jsonl` fails with *"The following paths are ignored by one of your .gitignore files"*, and `main.go:1157-1158` turns that into `FatalError`.

### Why it's a true contradiction (not a test that just needs nudging)

This PR also adds `TestAutoExportGitAddFailureExitsNonZero` (`cmd/bd/export_auto_test.go:506`) which uses the **same** setup — `.gitignore` = `.beads/`, `export.git-add=true` — and asserts the **opposite**: `bd create` must fail with an error containing `ignored` (lines 551, 558).

So two tests now demand opposite outcomes for identical preconditions:

| Precondition | `prune_..._gitignored_auto_export` (on `main`) | `TestAutoExportGitAddFailureExitsNonZero` (this PR) |
|---|---|---|
| `.beads/` gitignored + `git-add=true`, run `bd create` | must **succeed** | must **fail** (exit ≠ 0) |

No single behavior passes both — one test has to change, which means picking the intended contract for "gitignored `.beads/` + `git-add=true`."

### Recommendation: scope the fatal behavior to *staging* failures; keep gitignored non-fatal

#3970 is specifically about *transient* staging failures — its repro is `touch .git/index.lock`. There, dolt and the on-disk JSONL are correct but the **staged** blob is stale, and a later `git commit` ships it. That failure mode **cannot occur for a gitignored `.beads/issues.jsonl`**: an untracked, ignored file is never staged and never committed, so there is no stale blob to ship. Failing the whole command there buys no safety and turns *every* `bd` write-op into a non-zero exit for anyone who gitignores `.beads/` with `git-add` on — the exact config the `main` test says is supported.

Suggested shape:

1. In `maybeAutoExport` (`export_auto.go:143-146`), treat a *"paths are ignored"* `git add` failure as warn-and-continue (or preflight with `git check-ignore` / `git ls-files --error-unmatch` and skip staging for an ignored-and-untracked target). Keep returning the error — i.e. fatal — for every other failure (locked index, timeout, etc.).
2. Re-point `TestAutoExportGitAddFailureExitsNonZero` at the actual #3970 trigger — a held `.git/index.lock` — instead of a gitignored path, so it exercises the real regression without contradicting the prune test.

That preserves #3970's guarantee (transient staging failure ⇒ non-zero exit, state file not poisoned) and keeps the existing "gitignored `.beads/` keeps working" contract.

If maintainers prefer the other direction — *gitignored + `git-add=true` is a misconfiguration that should fail loudly* — then `prune_..._gitignored_auto_export` on `main` needs rewriting (enable `git-add` only after the setup `create`/`close`, or assert the failure), and this PR's behavior stands as-is. Either way it's a small change; it just needs a call on the contract.

Happy to implement whichever direction you choose.

*— Investigator agent (gastownhall/gc-management), tracking bead `be-1wzr`*

quad341:
Pushed `1c7bbe2a9` to get CI green (the only red check was **Embedded Dolt Cmd 18/20**).

### Root cause of the red

The CI failure was a deterministic **contract conflict**, not flake:

- This PR made *every* auto-export `git add` failure fatal (`export_auto.go` → `FatalError` in `main.go`), **including** the case where `.beads/` is intentionally gitignored.
- `main`'s existing contract test `TestEmbeddedPrune/prune_last_issue_removes_gitignored_auto_export` sets up exactly that (`.gitignore` = `.beads/`, `export.git-add=true`) and requires `bd create`/`close` to **succeed**.
- This PR's new `TestAutoExportGitAddFailureExitsNonZero` used the *same* setup but required it to **fail**.

Identical preconditions, opposite required outcomes → no single behavior satisfies both; one test had to change.

### The decision (preserve `main`'s contract)

A gitignored `.beads/` is a **supported workflow**: the JSONL is written for local use / Dolt sync but intentionally left untracked, so `git add` can never stage it. There's no GH#3970 stale-blob risk in that case because nothing is ever committed. So:

1. **Gitignored → non-fatal.** `maybeAutoExport` now classifies the `gitAddFile` error with `git check-ignore`. If the export path is gitignored, it's a benign skip (debug log, command succeeds). Main's prune contract stays as-is and passes.
2. **Real failures → still fatal.** A genuine staging failure — e.g. a locked `.git/index`, which is the actual GH#3970 repro — still aborts the command with a caller-visible non-zero exit, exactly as this PR intended.
3. **`gitAddFile`'s own contract is unchanged** (it still returns an error in both cases), so its unit tests are untouched.
4. **Re-pointed the new test** from the gitignored trigger (now non-fatal by contract) to a held `.git/index.lock`, so it still asserts the real fatal staging failure.

### Verified locally
- `TestEmbeddedPrune/prune_last_issue_removes_gitignored_auto_export` — PASS (was failing)
- `TestAutoExportGitAddFailureExitsNonZero` (re-pointed) — PASS
- `TestGitAddFile_*` unit tests — PASS
- Full `TestAutoExport*` / `TestExport*` / `TestEmbeddedPrune` surface — PASS

Credit to @maphew for the original change (PR #4092) — this only refines the *scope* of which staging failures are fatal. Leaving the merge to the beads maintainers.

maphew:
closing in favour of #4092. if I missed anything of value in here please let me know!

PR #4101, opened 2026-05-22
docs: add Related Projects pointer (scry)

prmichaelsen:
## What

Adds `docs/RELATED_PROJECTS.md` — a new doc for adjacent or complementary projects that solve different problems in the same neighborhood as Beads — and a one-line pointer to it from the README.

Seeds it with one entry: [scry](https://github.com/prmichaelsen/scry) ([scryspec.com](https://scryspec.com)), a marker-indexed recall graph for AI coding agents.

## Why this is distinct from `COMMUNITY_TOOLS.md`

`COMMUNITY_TOOLS.md` is for Beads *integrations* — UIs, extensions, tools that consume `bd` data. scry is not an integration; it's an independent project that composes with Beads at the workflow level, not at the data level. Filing it under COMMUNITY_TOOLS would mis-classify it.

If you'd prefer a different home (or none at all), happy to adjust or withdraw — this is a proposal, not a load-bearing change.

## Why mention scry at all

Beads is a task graph: *what to do next, what blocks what, who's working on it*. scry is a recall graph: *what was decided, what was learned, what to remember*. They occupy adjacent layers of the same problem — agent memory and coordination — and arrived independently at the same hash-based-ID convention (`bd-a1b2` ↔ `~hash`) to solve the same multi-agent collision problem. The README says it directly:

> Hash-based IDs (`bd-a1b2`) prevent merge collisions in multi-agent/multi-branch workflows.

scry's markers do the same thing for designs, lessons, and decisions. A Beads user who's hit the "what did we decide three weeks ago" wall may find it useful; a scry user who needs to coordinate concurrent work may find Beads useful. Surfacing the pointer in both directions seems worth a line.

## Disclosure

I'm an author of scry — happy to defer to a maintainer who'd rather not have project authors land their own pointers. If you'd like someone else to vouch for the entry first, or want to drop this entirely, that's fine.


maphew:
Merged as part of the stale-band salvage sweep - thanks for the addition, and sorry it sat so long while main churned underneath it.

PR #4022, opened 2026-05-17
feat(stats): --no-blocked flag skips blocked-count computation; fix Blocked:0 red render (be-nnqp, be-4b63)

quad341:
## Summary

- **be-nnqp**: `bd status --no-blocked` skips the expensive `computeBlockedIDs` traversal; JSON output adds `blocked_count_skipped: true`; human output renders `(skipped)` in muted style. New `FastStatisticsStore` interface + `GetStatisticsNoBlocked` method on DoltStore and EmbeddedDoltStore.
- **be-4b63**: `Blocked: 0` no longer renders red. Only `Blocked: N` where N > 0 uses `RenderFail`. Zero is rendered as plain text.
- `BlockedIssues` field promoted from `int` to `*int` (`nil` = skipped by `--no-blocked`; all existing callers nil-guarded).

## Test plan

- [ ] `bd status` still shows blocked count normally
- [ ] `bd status --no-blocked` shows `Blocked: (skipped)` and returns quickly
- [ ] `bd status --json` includes `blocked_count_skipped: true` when `--no-blocked` is set
- [ ] `bd status` with zero blocked issues shows `Blocked: 0` in plain text (not red)
- [ ] Existing `TestStatus*` and `queries_test.go` blocked-count tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4104, opened 2026-05-22
perf(stats): parallelize ScanIssueCounts and computeBlockedIDs in GetStatistics

### LINKED ISSUE #4102: perf(server-mode): bd commands take 5-10s in remote dolt_mode due to cold connection + serial query pattern
## Summary

Every `bd` command takes **5–10s wall-clock** on a remote Dolt server backend despite a sub-second raw roundtrip to the server. CPU is ~3% — time is entirely I/O-bound serial SQL.

## Environment

- bd version: 1.0.4 (Homebrew)
- Backend: `dolt_mode = server`, remote MySQL server over VPN
- Raw TCP/TLS roundtrip (`bd dolt test`): **0.22s**

## Timings

| Command | Wall time |
|---|---|
| `bd dolt test` | 0.22s |
| `bd list --status=open` | 4.8s |
| `bd ready` | 5.1s |
| `bd stats` | 10.3s |

Disabling backup (`BD_BACKUP_ENABLED=false`) had no meaningful effect — backup is not the bottleneck.

## Root cause (traced from source)

Two independent costs stack every invocation:

### 1. Cold connection per process (~200–500ms)

`DoltStore.New` (`internal/storage/dolt/store.go:929`) opens a new `sql.DB` with a 10-conn pool, but the pool lifetime is the CLI process. No connection reuse across `bd` invocations. Every command pays TLS handshake + MySQL auth overhead.

### 2. Serial query pattern in hot read path (~10–15 roundtrips × 200ms)

`bd ready` path:

1. `PersistentPreRun` (`cmd/bd/main.go:604`): config load, port resolution, DSN open, `autoMigrateOnVersionBump`, molecule loader — each touches the DB serially.
2. `GetReadyWork` → `issueops.GetReadyWorkInTx` (`internal/storage/issueops/ready_work.go`): **5 serial `QueryContext`/`QueryRowContext` calls** at lines 525, 555, 584, 635, 680.
3. `computeBlockedIDsForReadyWork` (`internal/storage/dolt/queries.go:131`): additional transitive-dependency scan in `issueops/blocked.go` (9 `tx.Query`/`tx.Exec` calls).
4. `buildParentEpicMap` (`cmd/bd/ready.go:343`): extra `GetDependencyRecordsForIssues` + parent fetch.
5. (Optional) `GetStatistics` on the "no ready work" hint path.

`bd stats` is worst (10.3s) because it chains `ScanIssueCountsInTx` + `computeBlockedIDs` + a ready-shaped recount sequentially (`internal/storage/dolt/queries.go:74–103`).

Net: ~10–15 serial roundtrips × ~200ms = 2–4s query time, plus cold-connection overhead.

## Suggested fixes (in ROI order)

1. **Batch `ScanIssueCountsInTx` + `computeBlockedIDs`** into a single query or parallel goroutines — both are read-only and independent.
2. **Skip `autoMigrateOnVersionBump` and molecule loader on read-only commands** (`isReadOnlyCommand` already identifies these) — migration check adds at least one DB roundtrip unconditionally.
3. **Parallelize independent queries inside `GetReadyWorkInTx`** — the 5 queries at lines 525/555/584/635/680 could be fanned out within the same tx.
4. **Long-lived connection proxy / bd daemon** — socket-mode process that keeps the MySQL connection warm across CLI invocations, similar to how `ssh-agent` or `git credential-helper` work. Eliminates per-process TLS handshake cost entirely.

## Not the root cause

- Network latency: raw RTT is 0.22s, healthy.
- Backup: `BD_BACKUP_ENABLED=false` made no difference.
- `maybeAutoImportJSONL`: correctly no-ops in server mode (line 45 of `auto_import_upgrade.go`).

fishgills:
## What

Parallelize the two independent database roundtrips inside `GetStatistics` using `errgroup`.

## Why

`GetStatistics` (`internal/storage/dolt/queries.go`) issued two serial roundtrips:

1. `ScanIssueCountsInTx` — scans the issues table
2. `computeBlockedIDs` — scans the dependencies table

Both are read-only and independent, but each waited for the previous to finish. On a remote Dolt server with ~200ms RTT per query, this serialization makes `bd stats` take ~10s. `bd dolt test` (one roundtrip) = 0.22s, so the problem is entirely in the serial query pattern.

Running them in parallel with `errgroup` removes one full roundtrip from the critical path. `withReadTx` uses `s.mu.RLock`, so concurrent calls are safe. `computeBlockedIDs` errors are treated as best-effort (blocked count falls back to 0), consistent with prior behavior.

Closes #4102

## Verification

```bash
# Build
make build

# Fast tests (no Dolt server needed)
make test

# With a live Dolt server, before/after:
time bd stats
```

Expected improvement for remote-server mode (~200ms RTT): ~10s → ~5s for `bd stats`.

The `TestDomainFS` failures visible in `./internal/...` pre-exist on `main` and are unrelated to this change (verified by running against the unmodified tree).

### DIFF
--- a/internal/storage/dolt/queries.go
+++ b/internal/storage/dolt/queries.go
@@ -5,6 +5,8 @@ import (
 	"database/sql"
 	"fmt"
 
+	"golang.org/x/sync/errgroup"
+
 	"github.com/steveyegge/beads/internal/storage/issueops"
 	"github.com/steveyegge/beads/internal/types"
 )
@@ -75,20 +77,33 @@ func (s *DoltStore) GetStaleIssues(ctx context.Context, filter types.StaleFilter
 func (s *DoltStore) GetStatistics(ctx context.Context) (*types.Statistics, error) {
 	stats := &types.Statistics{}
 
-	err := s.withReadTx(ctx, func(tx *sql.Tx) error {
-		return issueops.ScanIssueCountsInTx(ctx, tx, stats)
+	// Run issue counts and blocked-ID computation in parallel: both are
+	// read-only and independent, and each requires a separate DB roundtrip.
+	// withReadTx acquires s.mu.RLock so concurrent calls are safe.
+	var blockedIDs []string
+	g, gctx := errgroup.WithContext(ctx)
+
+	g.Go(func() error {
+		return s.withReadTx(gctx, func(tx *sql.Tx) error {
+			return issueops.ScanIssueCountsInTx(gctx, tx, stats)
+		})
 	})
-	if err != nil {
+
+	g.Go(func() error {
+		// computeBlockedIDs caches results across GetReadyWork and
+		// GetStatistics calls within the same CLI invocation.
+		ids, err := s.computeBlockedIDs(gctx, true)
+		if err == nil {
+			blockedIDs = ids
+		}
+		return nil // blocked count is best-effort; don't fail stats on error
+	})
+
+	if err := g.Wait(); err != nil {
 		return nil, fmt.Errorf("failed to get statistics: %w", err)
 	}
 
-	// Blocked count: reuse computeBlockedIDs which caches the result across
-	// GetReadyWork and GetStatistics calls within the same CLI invocation.
-	var blockedCount int
-	blockedIDs, err := s.computeBlockedIDs(ctx, true)
-	if err == nil {
-		blockedCount = len(blockedIDs)
-	}
+	blockedCount := len(blockedIDs)
 	stats.BlockedIssues = blockedCount
 
 	// Ready count: compute without using the ready_issues view to avoid
--- a/internal/storage/dolt/queries_test.go
+++ b/internal/storage/dolt/queries_test.go
@@ -2420,6 +2420,71 @@ func TestGetStatistics_ReadyIssuesExcludesBlocked(t *testing.T) {
 	}
 }
 
+func TestGetStatistics_ConcurrentCallsReturnConsistentResults(t *testing.T) {
+	// Regression test for the parallel GetStatistics implementation:
+	// two concurrent calls must return the same counts and not race.
+	store, cleanup := setupTestStore(t)
+	defer cleanup()
+
+	ctx, cancel := testContext(t)
+	defer cancel()
+
+	issues := []*types.Issue{
+		{ID: "par-open-1", Title: "Open 1", Status: types.StatusOpen, Priority: 1, IssueType: types.TypeTask},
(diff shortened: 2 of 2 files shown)


maphew:
Thanks for this PR — the diagnosis in the description is exactly right for the code as it stood in May: `GetStatistics` issued two serial roundtrips, `computeBlockedIDs` was the expensive one (a dependencies-table scan), and on a remote Dolt server each serialized roundtrip costs real wall-clock time. The implementation is also careful: `errgroup.WithContext` is used correctly, `g.Wait()` gives the needed happens-before for `stats` and `blockedIDs`, the blocked-IDs cache is already `cacheMu`-protected so the concurrent path is race-free, and you included a concurrency regression test. Nice work.

Unfortunately the code this PR targets has since been rewritten on `main`, which is why the branch now shows conflicts:

- The read-optimization work (ff8a2e8cd) removed the `computeBlockedIDs` call from `GetStatistics` entirely. Blocked count is now a single cheap `SELECT COUNT(*)` against the denormalized `is_blocked` column — the expensive dependencies scan this PR parallelizes no longer exists in this path (`internal/storage/dolt/queries.go`, `GetStatistics`).
- That rewrite also changed error semantics: the blocked count is now a hard error rather than best-effort. The PR's `return nil // blocked count is best-effort` pattern in the second goroutine matched the old base, but a rebased version must propagate that error instead, or it would silently report blocked=0 and inflate the ready count on query failure.

So the core premise is superseded, and I'm going to close this as such rather than ask for a rework — most of the win this PR targeted has already landed via the denormalized column.

There is a small residual opportunity if you're still measuring slow `bd stats` against a remote server: `GetStatistics` on current `main` still does two serial `withReadTx` roundtrips (`ScanIssueCountsInTx` + the `is_blocked` COUNT), so the same errgroup pattern re-derived on the new shape would save one RTT (~200ms in your setup). If your before/after timing on current `main` shows that's still worth the added concurrency in this path, a fresh small PR rebased on the new code would be welcome — please include the hard-error propagation for the blocked count, and note that the concurrency test as written can be satisfied from the blocked-IDs cache, so it's worth constructing it to force both transactions to hit the database.

Also relevant to the broader complaint in #4102 (which stays open — closing this PR unmerged does not close it; it covers cold-connection and serial patterns across all commands, not just `bd stats`): #4167 is exploring roundtrip reduction across remote-Dolt paths, and #4022 touches the same `GetStatistics` blocked-count area. Worth a look before re-deriving, to avoid collisions.

Closing as superseded by the `is_blocked` denormalization on `main`, with thanks — the serial-roundtrip analysis here was correct and is part of why the residual gap is now so small.

PR #4112, opened 2026-05-22
fix(create): set status=deferred when --defer date is in the future (GH#4071)

### LINKED ISSUE #4071: bug: bd create --defer does not set status to deferred
When creating an issue with `bd create --defer +10y`, the `defer_until` field is set correctly, but the `status` field remains `open` instead of being set to `deferred`.

## Steps to Reproduce

```bash
bd create "test defer" --defer +10y --json
```

Expected output shows:
- `status: "deferred"`
- `defer_until: "2036-..."`

Actual output:
- `status: "open"`
- `defer_until: "2036-..."`

Note: `bd defer <id>` correctly sets `status: deferred`, so this is specific to the `create` command.

## Expected Behavior

`bd create --defer <date>` should set `status: deferred` when a defer date is provided, matching the behavior of `bd defer <id>`.

## Workaround

After `bd create --defer`, manually run `bd defer <id>` to fix the status field.

kevglynn:
## Summary

Fixes #4071.

`bd create --defer=+2h` correctly sets `defer_until` but leaves `status` as `open`. This was fixed for `bd update --defer` in GH#3233 but the create path was missed.

**Impact:**
- `bd list --status deferred` misses these issues
- `bd undefer` rejects them (requires status=deferred)
- UI shows open icon instead of deferred icon

## Fix

In `buildCreateIssue`, set `status=deferred` when `DeferUntil` is in the future. Past defer dates keep `open` (matching existing warning behavior and `update.go` logic).

## Test plan

- [ ] `bd create "test" --defer=+1h --json` shows `status: "deferred"`
- [ ] `bd create "test" --defer=-1h --json` shows `status: "open"` (past date)
- [ ] `bd list --status deferred` includes the deferred issue
- [ ] `bd undefer <id>` works on the created issue

### DIFF
--- a/cmd/bd/create.go
+++ b/cmd/bd/create.go
@@ -759,6 +759,11 @@ func buildCreateIssue(params createIssueParams) *types.Issue {
 		externalRefPtr = &params.ExternalRef
 	}
 
+	status := types.StatusOpen
+	if params.DeferUntil != nil && params.DeferUntil.After(time.Now()) {
+		status = types.StatusDeferred
+	}
+
 	return &types.Issue{
 		ID:                 params.ID,
 		Title:              params.Title,
@@ -767,7 +772,7 @@ func buildCreateIssue(params createIssueParams) *types.Issue {
 		AcceptanceCriteria: params.AcceptanceCriteria,
 		Notes:              params.Notes,
 		SpecID:             params.SpecID,
-		Status:             types.StatusOpen,
+		Status:             status,
 		Priority:           params.Priority,
 		IssueType:          params.IssueType,
 		Assignee:           params.Assignee,


harry-miller-trimble:
Thanks for this, @kevglynn! To get it landed I rebased it onto current `main` (this branch is ~50 commits behind) and added test coverage in #4198. Your original commit is preserved there with attribution.

One thing worth noting: merging as-is would have broken `TestBuildCreateIssueFromInput_PopulatesAllFields`, which hard-codes a future defer date and asserts `StatusOpen` — #4198 fixes that assertion too.

Continuing review in #4198; this can be closed in favor of it.

kevglynn:
Thanks @harry-miller-trimble — appreciate the rebase and the test coverage in #4198!

PR #4113, opened 2026-05-22
fix(close): treat re-close of already-closed issue as idempotent success (GH#4025)

### LINKED ISSUE #4025: bug: `bd close <id>` returns "issue not found" when re-closed within same wall-clock second (RowsAffected==0 misinterpreted as missing row)
## Summary

`bd close <id>` reports `Error closing <id>: issue not found: <id>` (exit 1) when invoked on an issue that was already closed within the same wall-clock second — even though the issue exists in the database with `status=closed`, fully visible via `bd show <id>` and `bd list --status closed`.

The misleading error comes from conflating two distinct MySQL/Dolt outcomes:
- **(a)** UPDATE matched no rows (the row truly doesn't exist), and
- **(b)** UPDATE matched a row but changed no columns (the row is already in the target state).

Both surface as `result.RowsAffected() == 0` with the default MySQL driver setting (rows-changed, not rows-matched). bd's close code treats only case (a).

## Reproduction (minimal, deterministic)

```bash
mkdir /tmp/bd-repro && cd /tmp/bd-repro
BD_NON_INTERACTIVE=1 bd init >/dev/null

# Run a few iterations to clear the second-tick lottery
for i in 1 2 3 4 5; do
  ID=$(BD_NON_INTERACTIVE=1 bd create --title "t$i" --type task --json | grep -oP '"id":\s*"\K[^"]+')
  BD_NON_INTERACTIVE=1 bd close \"\$ID\" >/dev/null    # first close: ok
  BD_NON_INTERACTIVE=1 bd close \"\$ID\"               # second close: usually "issue not found"
done
```

Local sequential reproduction rate: **~40-55% of iterations** fail on the second close. No concurrency, no auto-routing, no special config — just a fresh `bd init` with embedded Dolt defaults.

After the failure, the issue clearly exists:

```
$ bd show <id> --json   # works, returns the issue with status=closed
$ bd list --status closed --json   # includes the issue
$ bd close <id>   # if you re-run AGAIN after a second has passed, succeeds
```

## Expected vs Actual

| | Expected | Actual (bd 1.0.4) |
|---|---|---|
| Exit code | 0 (idempotent) OR a clear "already closed" message | 1 |
| stderr | `nil` or `"already closed"` | `Error closing <id>: issue not found: <id>` |
| Caller interpretation | "the issue ended up closed" (correct) | "the ID is invalid/typo'd" (wrong — the ID is fine) |

## Root cause

`internal/storage/issueops/close.go:22-49` (v1.0.4):

```go
func CloseIssueInTx(ctx context.Context, tx *sql.Tx, id string, reason, actor, session string) (*CloseResult, error) {
    // ...
    now := time.Now().UTC()

    result, err := tx.ExecContext(ctx, fmt.Sprintf(`
        UPDATE %s SET status = ?, closed_at = ?, updated_at = ?, close_reason = ?, closed_by_session = ?
        WHERE id = ?
    `, issueTable), types.StatusClosed, now, now, reason, session, id)
    // ...
    rows, err := result.RowsAffected()
    // ...
    if rows == 0 {
        return nil, fmt.Errorf(\"issue not found: %s\", id)   // <-- bug: false inference
    }
    // ...
}
```

The `issues.closed_at` and `updated_at` columns are declared `DATETIME` (second resolution) — `internal/storage/schema/migrations/0001_create_issues.up.sql:17-18`.

For a re-close within the same second of the first close:
- `status`: 'closed' → 'closed' (no change)
- `closed_at`: '2026-05-17 19:41:11' → '2026-05-17 19:41:11' (no change at second resolution)
- `updated_at`: '2026-05-17 19:41:11' → '2026-05-17 19:41:11' (no change)
- `close_reason`: 'Closed' → 'Closed' (no change, default reason)
- `closed_by_session`: NULL → NULL (no change in BD_NON_INTERACTIVE without a session ID)

→ zero columns actually change → driver returns `RowsAffected = 0` → code emits `\"issue not found\"`.

## Decisive evidence — column-change test

If the root cause is RowsAffected==0 from a no-op UPDATE, forcing ANY column to actually change should make the bug vanish. It does:

```bash
# 15 iterations, second close back-to-back, BUT with --reason "different reason"
# so close_reason changes 'Closed' -> 'different reason' (RowsAffected = 1)
for i in $(seq 1 15); do
  ID=\$(BD_NON_INTERACTIVE=1 bd create --title \"t$i\" --type task --json | jq -r .id)
  BD_NON_INTERACTIVE=1 bd close \"\$ID\" >/dev/null
  BD_NON_INTERACTIVE=1 bd close \"\$ID\" --reason \"different reason\"
done
# Result: 0/15 failures (vs 4-11/15 for the back-to-back no-arg variant)
```

This isolates the cause to the no-op UPDATE → RowsAffected==0 path. Not a race, not a routing/cache issue, not the dolt commit pipeline.

## Suggested fix

Distinguish "row missing" from "row matched but UPDATE was a no-op". Simplest patch in `CloseIssueInTx`:

```go
if rows == 0 {
    // RowsAffected==0 in MySQL/Dolt can mean either (a) no row matched or
    // (b) row matched but no columns changed. Disambiguate with a SELECT.
    var existingStatus string
    err := tx.QueryRowContext(ctx, fmt.Sprintf(\"SELECT status FROM %s WHERE id = ?\", issueTable), id).Scan(&existingStatus)
    if errors.Is(err, sql.ErrNoRows) {
        return nil, fmt.Errorf(\"issue not found: %s\", id)
    }
    if err != nil {
        return nil, fmt.Errorf(\"failed to verify issue existence: %w\", err)
    }
    // Row exists and is already closed — idempotent success.
    // (If existingStatus != \"closed\" here, something is wrong; treat as error.)
    if existingStatus != string(types.StatusClosed) {
        return nil, fmt.Errorf(\"close UPDATE affected 0 rows but issue %s has status=%q\", id, existingStatus)
    }
}
```

Alternative approaches that would also work:
- **Early exit on already-closed**: SELECT status before the UPDATE; if already 'closed', return idempotent success and skip the UPDATE entirely (also avoids the spurious 'closed' event insert in the same-second case).
- **Driver flag**: connect with `CLIENT_FOUND_ROWS=1` so RowsAffected returns rows-matched. Less surgical, may affect other call sites.
- **Sub-second timestamps**: migrate `DATETIME` → `DATETIME(6)`. Schema change; doesn't address the root semantic bug.

## Impact

- Misleading user-facing error (looks like a typo'd ID).
- Breaks the principle of least surprise for idempotent close (cf. `kubectl delete --ignore-not-found`, `rm -f`).
- Bites callers driving bd programmatically that may retry on transient errors or close-twice as part of normal flows.

## Environment

- `bd version 1.0.4 (ce242a879)` (installed via mise, plugin `github:gastownhall/beads`)
- Linux 6.17.0, x86_64
- Embedded Dolt mode (default `bd init`)
- No special config (no routing, no custom statuses, no wisp/molecule)

## Searched for prior reports (none matched)

Searched the repo for: `issue not found`, `idempoten`, `already closed`, `close twice`, `double close`, `close lookup`, `default filter excludes closed`. Issue #3608 has a similar shape but a different error wording (`Error: resolving ID X: no issue found matching X` — the standard ID resolver path) and a different precondition (contributor auto-routing mode). Issue #3905 is also close-related but about silent UPDATE no-op without any error.

kevglynn:
## Summary

Fixes #4025.

`bd close <id>` returns "issue not found" when re-closing an already-closed issue within the same wall-clock second. The root cause: the UPDATE touches no columns (status already closed, timestamps identical within the same second), so `RowsAffected()` returns 0 — previously misinterpreted as "row doesn't exist."

## Fix

When `RowsAffected == 0`, query the row's current status in the same transaction:
- If already closed → return success silently (idempotent, no duplicate event)
- If row truly missing → preserve "issue not found" error
- If row exists but in unexpected state → return generic failure

Follows the same idempotent pattern already used in `claim.go`.

## Test plan

- [ ] Close an issue, immediately close again → success (no error)
- [ ] Close a nonexistent ID → still returns "issue not found"
- [ ] No duplicate `closed` event recorded on re-close
- [ ] Scripts doing `bd close $id && bd close $id` no longer fail

### DIFF
--- a/internal/storage/issueops/close.go
+++ b/internal/storage/issueops/close.go
@@ -56,7 +56,20 @@ func closeIssueInTx(ctx context.Context, tx *sql.Tx, id string, reason, actor, s
 		return nil, fmt.Errorf("failed to get rows affected: %w", err)
 	}
 	if rows == 0 {
-		return nil, fmt.Errorf("issue not found: %s", id)
+		var status string
+		qerr := tx.QueryRowContext(ctx,
+			fmt.Sprintf(`SELECT status FROM %s WHERE id = ?`, issueTable), id,
+		).Scan(&status)
+		if qerr == sql.ErrNoRows {
+			return nil, fmt.Errorf("issue not found: %s", id)
+		}
+		if qerr != nil {
+			return nil, fmt.Errorf("failed to check issue existence: %w", qerr)
+		}
+		if types.Status(status) == types.StatusClosed {
+			return &CloseResult{IsWisp: isWisp}, nil
+		}
+		return nil, fmt.Errorf("failed to close issue: %s", id)
 	}
 
 	if recordEvent {


harry-miller-trimble:
Thanks, @kevglynn! Rebased onto current `main` (this branch is ~50 commits behind) and added sqlmock tests for the new idempotent-close path in #4199. Your original commit is preserved there with attribution.

Continuing review in #4199; this can be closed in favor of it.

kevglynn:
Thanks @harry-miller-trimble! The sqlmock tests for the idempotent path are a nice addition.

PR #4114, opened 2026-05-22
fix(list): disable default truncation when stdout is piped (GH#4094)

### LINKED ISSUE #4094: beads should not limit outputs for non-interactive commands (use tty detection)
Use tty detection to avoid truncating output for the sake of human-readable defaults for machine output.

```
bd list | grep merge

Showing 50 issues; more results matched but were hidden by --limit. Use --limit 0 for all, or --limit N to raise the cap.
```

beads should not truncate output for commands run in a script or as a part of a piped command.

kevglynn:
## Summary

Fixes #4094.

When `bd list` output is piped (e.g. `bd list | grep merge`), the default `--limit 50` silently truncates results and prints a warning to stderr. Users expect piped commands to emit all matching data without truncation.

## Fix

Two changes:

1. **`cmd/bd/list.go`** — Extend the effective-limit resolution: when stdout is not a TTY and `--limit` was not explicitly set, default to unlimited (0). Priority order:
   - Explicit `--limit` wins
   - `--all` implies unlimited
   - **Non-TTY stdout → unlimited** (new)
   - Agent mode → 20
   - Default → 50

2. **`cmd/bd/list_output.go`** — Suppress the truncation hint when stderr is not a terminal (matches existing pattern in `output.go` for deprecation notices).

Uses the existing `ui.IsTerminal()` and `ui.IsStderrTerminal()` helpers from `internal/ui/terminal.go`.

## Test plan

- [ ] `bd list | cat` → outputs all issues (no truncation)
- [ ] `bd list` in terminal → still truncates at 50 with hint
- [ ] `bd list --limit 10 | cat` → respects explicit limit (10 results)
- [ ] `bd list --all` → unlimited regardless of TTY
- [ ] Truncation hint suppressed when stderr is piped

### DIFF
--- a/cmd/bd/list.go
+++ b/cmd/bd/list.go
@@ -535,8 +535,9 @@ var listCmd = &cobra.Command{
 		// Resolve effective limit. Priority order:
 		// 1. Explicit --limit always wins (user intent is clear)
 		// 2. --all implies unlimited when --limit is not set (GH#1840)
-		// 3. Agent mode uses a lower default for context efficiency
-		// 4. Default limit (50) otherwise
+		// 3. Non-interactive (piped) stdout disables truncation (GH#4094)
+		// 4. Agent mode uses a lower default for context efficiency
+		// 5. Default limit (50) otherwise
 		limitChanged := cmd.Flags().Changed("limit")
 		effectiveLimit := limit
 		switch {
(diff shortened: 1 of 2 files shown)


quad341:
Hi @kevglynn, thanks for the fix! I've opened #4158 which includes your change with the failing CI test updated to match the new piped-stderr contract (GH#4094). The test at `list_embedded_test.go:205` was asserting the old behavior (hint always appears), but your change correctly suppresses it when stderr is not a TTY. Maintainers will close this once #4158 is reviewed.

kevglynn:
Makes sense @quad341 — thanks for catching the test contract update. Happy to close this in favor of #4158.

kevglynn:
Closing in favor of #4158 which includes this fix with the test contract updated for the new piped-stderr behavior. Thanks @quad341!

PR #4116, opened 2026-05-22
/{cmd,internal}: add migrate schema command

coffeegoddd:
surfaces top level command for migrating db schema which can be run after a beads upgrade, so migrations dont block normal bd command usage.

### DIFF
--- a/cmd/bd/migrate.go
+++ b/cmd/bd/migrate.go
@@ -23,10 +23,6 @@ var migrateCmd = &cobra.Command{
 
 Without subcommand, checks and updates database metadata to current version.
 
-Flags:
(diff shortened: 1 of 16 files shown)


quad341:
Closing in favor of #4157 which includes this change plus the regenerated CLI reference docs that were missing (fixes  CI failure). Thanks for the contribution @coffeegoddd!

quad341:
Hi @coffeegoddd, thanks for the contribution! The CI failure is that the CLI reference docs need to be regenerated after adding the new `migrate schema` command. I've opened #4157 which includes your change with the docs fix applied. Maintainers will close this PR once #4157 is reviewed.

PR #3671, opened 2026-05-03
fix(graph): emit [] not null for empty links in HTML output (GH#3592)

kevglynn:
## Summary
- Fixes `const links = null` in `bd graph --html` output for single-node components (and any graph with no qualifying edges)
- Go nil slices marshal to JSON `null`, which breaks `d3.forceLink(links)` — changed `buildHTMLGraphData` and `buildHTMLEdgeData` to use `make()` so empty collections serialize as `[]`
- Adds test `TestRenderGraphHTML_EmptyEdgesNotNull` and extends `TestMergeSubgraphsForHTML_SingleDOCTYPE` to assert no `null` links

Closes #3592

## Test plan
- [x] `make build` succeeds
- [x] `go test -tags gms_pure_go ./cmd/bd/... -run 'Graph|HTML'` — all pass
- [x] New test verifies single-node graph emits `const links = []` not `null`
- [x] Existing merge test extended to verify `[]` links after component merge

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3671"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

PR #3648, opened 2026-05-02
fix(graph): merge components for bd graph --all --html (GH#3592)

kevglynn:
## Summary
- `bd graph --all --html` concatenated N full `<!DOCTYPE html>` documents (one per connected component), producing invalid HTML that browsers could not render correctly.
- Merge all components into a single subgraph before rendering so stdout is always one valid HTML page.
- `computeLayout` now tolerates `nil Root` for the merged graph.

Fixes #3592

## Test plan
- [x] `TestMergeSubgraphsForHTML_SingleDOCTYPE` — two disconnected components produce exactly one HTML document
- [x] `TestComputeLayout/nil_root_does_not_panic`
- [x] `go test -tags gms_pure_go ./cmd/bd/...` passes

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3648"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

PR #4117, opened 2026-05-23
fix(graph): scope CSS to #graph and add fit-to-view for HTML viewer

A3Ackerman:
## Summary

`bd graph --html` is unusable on non-trivial graphs: the legend covers the viewport and graph nodes run off-screen even after Reset View. Two distinct bugs in the embedded HTML template:

1. **CSS scope blow-up.** `svg { width: 100vw; height: 100vh }` matches every `<svg>` on the page — including the 30×10 legend swatches for "blocks" and "parent-child" edge types. Each swatch balloons to viewport size, stretching the `#legend` container to ~1500×1600 and covering the graph. Fix: scope to `#graph`.
2. **Initial-zoom does not actually fit.** `d3.zoomIdentity.translate(width/4, height/4).scale(0.8)` is a fixed transform that doesn't read node positions; multi-layer graphs (`x = 150 + layer*220`) push deep-layer nodes off-screen to the right. Fix: real `fitToView()` that reads the post-simulation bounding box and transforms to fit with padding. Adds a "Fit View" button and repoints `resetZoom()` to it so the existing Reset View button behaves as users expect.

Template-only change. No public API surface, no behavior change for the DOT renderer.

## Why

Continuation of recent graph-viewer fixes:

- #3648 (merged): single document for `--all --html`
- #3671 (merged): emit `[]` not `null` for empty links

This addresses the remaining unusability of the rendered viewer itself. Followed CONTRIBUTING.md "one issue per PR" — did not bundle the adjacent observation that `--all --html` still emits N sub-graphs per connected component (that is a layout choice, not a CSS bug; happy to file separately).

## Verification

- Manually rendered against a real graph: legend ~134×190 (was ~1500×1600), graph auto-fits on initial render and on Reset View. Reproduced the bug + verified the fix on macOS Chrome.
- `make build` — clean
- `golangci-lint run --build-tags gms_pure_go ./cmd/bd/...` — `0 issues.`
- `make test` — hits an environmental hang in `internal/utils/TestMain` (`testutil.EnsureDoltContainerForTestMain`) on my local box, unrelated to a Go HTML template string change. I expect CI (with Docker) to run this cleanly; happy to investigate further if it doesn't.

## Notes for reviewers

- `setTimeout(fitToView, 1500)` rather than `simulation.on("end", ...)` — force simulations with strong charges don't naturally reach `end` without explicit `alpha`/`alphaTarget` discipline. The 1500ms heuristic works in practice; happy to revisit.
- `fitToView()` caps `scale` at the existing `zoom.scaleExtent([0.1, 4])` upper bound (`Math.min(width/bboxW, height/bboxH, 4)`).
- Initial analysis and patch design by Syl (qcore crew); applied and verified upstream by me.


maphew:

maphew:
Merged - both fixes verified still needed against current main before landing (the unscoped `svg` CSS and the missing fitToView), and the full suite ran green after a branch refresh. Thanks for the clean single-file fix and the manual verification notes.

PR #4118, opened 2026-05-23
fix(show): bd show --include-dependents silently returns no dependents on Dolt server (regression from #4010)

quad341:
## Summary

`bd show <id> --json --include-dependents` **silently returns no dependents** on the Dolt sql-server backend, even when the issue has dependents. The default count-only output reports a correct nonzero `dependent_count`, so the JSON is self-contradictory (claims N dependents, lists none) — with **no error** surfaced to the user.

This is a regression introduced in #4010 (commit `9e35080be`).

## Root cause

`DoltStore.IterDependentsWithMetadata` (`internal/storage/dolt/iter_dependents.go`) filtered both UNION branches with `WHERE d.depends_on_id = ?`. The generated `depends_on_id` column was dropped from both `dependencies` and `wisp_dependencies` by migration `0043_drop_dependencies_generated_column`. On a post-split Dolt server the query fails with:

```
ERROR 1105 (HY000): table "d" does not have column "depends_on_id"
```

The error is returned up to `cmd/bd/show.go:163`, where the `if err == nil { … }` guard has **no else branch**, so the 1105 is silently dropped and `details.Dependents` stays nil.

`9e35080be` rewrote this query to UNION the wisp table and fixed the **count** path (`counts.go` `depTargetExpr`), but left the **iter** path's WHERE clauses on the dropped physical column.

## Why CI didn't catch it

- The dolt-backend guard `TestCountAndIterIncludeWispDependencies` *does* assert this path — but it is **skipped without a running test Dolt server** (`"Test Dolt server not running, skipping test"`), which is the CI tier that ran for #4010.
- The embedded backend's `IterDependentsWithMetadata` is a slice-delegating stub (`embeddeddolt/iter_stubs.go`) that reuses the correct `GetDependentsWithMetadata` path, so the embedded test was green.

## Fix

Filter on the split-column target expression, matching the slice path (`issueops.DepTargetExpr`) and the count path (`counts.go` `depTargetExpr`):

```sql
WHERE COALESCE(d.depends_on_issue_id, d.depends_on_wisp_id, d.depends_on_external) = ?
```

One file, both UNION branches. No API or behavior change beyond fixing the query.

## Verification (end-to-end, live post-split Dolt server)

`bd show bd-rbxi --json --include-dependents` (bd-rbxi has 12 dependents):

| binary | `dependent_count` | `dependents[]` |
|---|---|---|
| before (origin/main) | 12 | **absent** |
| after (this PR) | 12 | **12** |

The existing `TestCountAndIterIncludeWispDependencies` covers this once a test Dolt server is available; no new test is added because the guard already exists (the gap is CI not running it, tracked separately).

## Follow-up (not in this PR)

`cmd/bd/show.go:163-164` silently swallows the iter error — that is *why* this break was silent rather than loud. Surfacing it would turn a future query break into an immediate failure. Left out here to keep the fix surgical; recommend a small hardening follow-up.

### DIFF
--- a/internal/storage/dolt/iter_dependents.go
+++ b/internal/storage/dolt/iter_dependents.go
@@ -45,16 +45,21 @@ type doltDependentsIter struct {
 // matches GetDependentsWithMetadata. See the package doc for why the join
 // target per edge table is unambiguous.
 func (s *DoltStore) IterDependentsWithMetadata(ctx context.Context, issueID string) (storage.Iter[types.IssueWithDependencyMetadata], error) {
+	// Resolve each edge's target from the split columns, NOT a physical
+	// depends_on_id: that generated column was dropped from both edge tables
+	// by migration 0043, so `d.depends_on_id` errors with 1105 on a post-split
+	// DB. Mirrors the count path (counts.go depTargetExpr) and the slice path
+	// (GetDependenciesWithMetadata, issueops.DepTargetExpr).
 	q := fmt.Sprintf(`
 		SELECT %s, d.type
 		FROM issues i
 		JOIN dependencies d ON d.issue_id = i.id
-		WHERE d.depends_on_id = ?
+		WHERE COALESCE(d.depends_on_issue_id, d.depends_on_wisp_id, d.depends_on_external) = ?
 		UNION ALL
 		SELECT %s, d.type
 		FROM wisps w
 		JOIN wisp_dependencies d ON d.issue_id = w.id
-		WHERE d.depends_on_id = ?
+		WHERE COALESCE(d.depends_on_issue_id, d.depends_on_wisp_id, d.depends_on_external) = ?
 		ORDER BY created_at ASC
 	`, prefixedIssueColumns("i"), prefixedIssueColumns("w"))
 	return s.iterIssuesWithDepType(ctx, q, issueID, issueID)


quad341:
### CI red here is **not** from this change — it's a repo-wide Dolt 2.0.6 break

This PR is a single-file, query-only change to `iter_dependents.go` (the be-kjp7x dependents fix) and is functionally correct. The three red jobs (`Test (storage domain + uow)`, `Test (ubuntu-latest)`, `Test (macos-latest)`) fail for a reason unrelated to this diff:

- CI installs dolt via `install.sh latest`. **Dolt 2.0.6 shipped on 2026-05-22** and rejects `ADD FOREIGN KEY` on the base column of a stored generated column (errno 1105).
- Migration `0041_split_dependencies_target.up.sql` does exactly that, so fresh-DB migration aborts mid-chain on 2.0.6.
- The last green main run (`cfcc95799`, this PR's parent) ran on dolt **2.0.5**; every run after 2.0.6 is red. The same three jobs fail identically on unrelated branches (e.g. `feat/linear-epic-parent-child`).
- Locally, these tests pass 26/26 on dolt 2.0.5 and fail deterministically on 2.0.6.

**Fix is up in #4120** (reorder the FK before the generated column; identical final schema; verified to fix all three tests on 2.0.6 with no 2.0.5 regression).

Recommendation: merge #4120 first, then this PR's CI goes green on re-run. This change itself needs no edits.

quad341:
**For reviewers/mergers: the red CI here is not caused by this change.**

This PR edits only `internal/storage/dolt/iter_dependents.go` (a query-string fix). The three red jobs — `Test (storage domain + uow)`, `Test (ubuntu-latest)`, `Test (macos-latest)` — are the repo-wide **Dolt 2.0.6** break of migration `0041_split_dependencies_target.up.sql`:

> errno 1105 "Cannot add foreign key on the base column of a stored generated column"

Dolt 2.0.6 (shipped 2026-05-22) added a check rejecting `ADD FOREIGN KEY` on the base column of a STORED generated column; migration 0041 adds `fk_dep_issue_target FOREIGN KEY (depends_on_issue_id)` after creating the generated `depends_on_id = COALESCE(depends_on_issue_id, …)`. The UOW `pending schema migrations alter pre-existing dirty tables: dependencies` failures are the secondary effect of 0041 aborting mid-chain. `main` itself is red for the identical reason (last green main `cfcc95799` ran on Dolt 2.0.5).

- Root cause + fix tracked in **be-ncb0e**.
- Fixed by **#4120** (reorders the FK before the generated column; identical final schema; all checks green, mergeable).

**Merge #4120 first**, then this PR's CI re-runs green with no changes needed.

quad341:
**Refreshed onto current `main` (#4120 merged) — now green except one pre-existing, unrelated check.**

I merged current `main` into this branch (merge commit `29c998cd6`) to pick up **#4120** (the Dolt 2.0.6 migration-0041 fix). On the fresh run:

- ✅ The three jobs that were red before — `Test (ubuntu-latest)`, `Test (macos-latest)`, `Test (storage domain + uow)` — now **pass**. Those were the repo-wide Dolt 2.0.6 break, fixed by #4120.
- ✅ All 20 Embedded-Dolt shards, Windows smoke, Lint, Build, formatting, version/doc checks — pass.
- ❌ `Differential Regression (v0.49.6 baseline)` — fails, but **not because of this PR**.

This PR's entire net diff vs `main` is one file: `internal/storage/dolt/iter_dependents.go` (+7/-2). The differential-regression failure is **pre-existing on `main`**, introduced by `cfcc95799` ("feat(show): count-only JSON details with opt-in streamed payloads", #4010):

| `main` commit | Differential Regression |
|---|---|
| `458cf899c` (parent of count-only) | ✅ success |
| `cfcc95799` (count-only) | ❌ **failure** ← starts here |
| `82020c42f` (current `main` HEAD) | ❌ failure |

Root cause: the harness diffs per-issue `bd show --json` against a pinned **v0.49.6** baseline. v0.49.6 emits `dependents[]`/`comments[]` inline; the count-only change makes them opt-in via `--include-dependents`/`--include-comments`, so the harness sees a JSONL mismatch. This PR merely *triggers* the job (because `internal/storage/` matches the risky-path gate added by #4123) — it does not *cause* the failure.

That harness gap is tracked separately with a fix spec in **be-bpmg5** (teach the regression harness to request the opt-in payloads for the candidate binary). It should not block this one-line query fix.

**Recommendation:** merge — the only red check is red on `main` independent of this change. Alternatively, apply the `skip-regression` label for a fully green run.

quad341:
**For whoever merges this:** the single red check — **Regression Tests → Differential Regression (v0.49.6 baseline)** — is a **pre-existing, branch-independent failure unrelated to this change**, and it is not a required check (PR state is `MERGEABLE`/`UNSTABLE`, not `BLOCKED`).

This PR's diff is one file (`internal/storage/dolt/iter_dependents.go` — the dependents query). The failing assertion is about the **comments** field:

```
scenarios_test.go:2324: field "comments" present in baseline show --json but missing from candidate (GH#1914)
```

Root cause: `cfcc95799` (PR #4010 / be-ijck6q) intentionally made default `bd show --json` count-only — it emits `comment_count` instead of a `comments` array, with `--include-comments` to opt in. The differential test `TestShowJSONFieldCompleteness` still encodes the old comments-by-default contract, so it fails on **every** storage/CLI PR since #4123 began gating the regression workflow. The unrelated PR `codex/ready-limited-wisp-pages-20260523` (job `77558112341`) fails the identical assertion — confirming it's not this branch.

All 40+ other checks pass. The stale-test fix is tracked in **be-tpfnu** (ready-to-build → builder). This PR is safe to merge on its own merits.

quad341:
### Overlaps with #4139 — same `iter_dependents.go` fix, and #4139 also turns the red check green

Heads-up for whoever merges this (the "unrelated red check" point is already covered above — adding the missing cross-reference and the precise root cause).

**Same fix, independently.** #4139 (`fix/pr4107-corruption-tests`) lands the *identical* be-kjp7x fix in `internal/storage/dolt/iter_dependents.go`: it replaces `WHERE d.depends_on_id = ?` with the COALESCE split-column target on **both** the `dependencies` and `wisp_dependencies` branches (via a `depTargetExprWithAlias("d")` helper). Functionally the same change this PR makes — so the two PRs will conflict on this file if both merge.

**Root cause of the red `Differential Regression (v0.49.6 baseline)` check** (and why #4139 clears it): commit `cfcc95799` ("feat(show): count-only JSON details with opt-in streamed payloads", be-ijck6q) made `bd show --json` count-only by default — `dependents` and `comments` are now opt-in behind `--include-dependents` / `--include-comments`. The regression harness's `snapshot()` still called plain `bd show --json`, so the candidate dropped those arrays vs the v0.49.6 baseline (which embedded them by default). That's also why the failure hits `TestCommentsRoundTrip`, a path this PR never touches. #4139 updates `snapshot()` to pass `--include-dependents --include-comments` (plus dep-subobject normalization), and its **Differential Regression check is green**.

**Net:** #4139 is a superset of this PR — same `iter_dependents.go` fix + the harness fix + the broader PR4107 blocked-state repair. Suggest merging **#4139** and closing this as superseded. (If landing the minimal be-kjp7x fix sooner is preferred instead, this PR is safe to merge with the differential check red — non-required and unrelated, as noted above — and #4139 can rebase.)

quad341:
### Independent verification — the red `Differential Regression` check is inherited from `main`, not introduced here

Adding the one artifact that settles the "is it this PR?" question: a failing-test-set diff against `main`.

- **`main` HEAD `82020c42f` fails the *identical* 39 differential tests** as this branch. I extracted the `FAIL: Test…` set from both runs — `main` job `77539489344` and this PR's job `77542639151` — sorted-unique, and **`diff` is empty**. This PR adds **zero** new failures.
- The check is **non-required / non-blocking**: `isRequired: null`, `mergeStateStatus: UNSTABLE`, `mergeable: MERGEABLE`.

Mechanism (confirms the root cause already noted above):
- `cmd/bd/show.go:145-158` (be-ijck6q / `cfcc95799`) makes `bd show --json` **count-only by default**; `dependents`/`comments` are emitted only under `--include-dependents` (`show.go:162`) / `--include-comments` (`show.go:205`).
- The regression harness calls **plain** `bd show --json` (`tests/regression/regression_test.go:438`; `scenarios_test.go:2307-2308`), so the candidate omits arrays the v0.49.6 baseline emits by default → JSONL mismatch on all 39.
- This PR's only change — the `WHERE` clause in `IterDependentsWithMetadata` — executes **only** when `--include-dependents` is passed (`show.go:163`), which the harness never does. The identical `main` failure set proves it can't be implicated.

Net: safe to merge as-is (red check non-required + proven-unrelated). #4139 supersets this (adds the harness fix that turns the check green) and can rebase. Either path is fine.

quad341:
**Recommend closing as superseded by #4139** (no action needed from reviewers beyond closing).

While this PR was open, #4139 ("fix: repair PR4107 blocked-state corruption", commit 8ae4c3c67) landed the **identical** be-kjp7x fix on `main`, plus a cleaner refactor:

- This PR inlines `WHERE COALESCE(d.depends_on_issue_id, d.depends_on_wisp_id, d.depends_on_external) = ?` in both UNION branches of `IterDependentsWithMetadata`.
- #4139 produces the **byte-identical SQL** via a new reusable helper `depTargetExprWithAlias("d")`, used in the same two branches.

Both eliminate the dropped-`depends_on_id` errno-1105 failure (migration 0043) that made `bd show --include-dependents` silently empty on the Dolt server backend.

Diffing this branch against current `main` (`git diff origin/main..HEAD -- internal/storage/dolt/iter_dependents.go`), the only net change is that this branch would **delete** #4139's `depTargetExprWithAlias` helper and re-inline the same expression — a stylistic revert with no functional effect. That is also the source of the current merge conflict in `iter_dependents.go`.

The regression guard `TestCountAndIterIncludeWispDependencies` (`internal/storage/dolt/counts_wisp_test.go`) already asserts `IterDependentsWithMetadata` streams both the permanent and wisp dependents.

be-kjp7x is already closed; the fix is on `main` via #4139. Thanks to @julianknutsen — closing this avoids reverting that helper.

quad341:
Closing as **superseded by #4139** (merged to `main` 2026-05-24).

Verification that no work is lost:
- This PR touches **only** `internal/storage/dolt/iter_dependents.go`.
- #4139 (merge commit `8ae4c3c`, confirmed an ancestor of `main`) refactored the same `IterDependentsWithMetadata` query to use `depTargetExprWithAlias("d")`, which expands to
  `WHERE COALESCE(d.depends_on_issue_id, d.depends_on_wisp_id, d.depends_on_external) = ?`
  — functionally identical to the fix in this PR, and already on `main`.
- This branch is now `CONFLICTING` against `main` precisely because main carries the equivalent fix.

The `be-kjp7x` regression (Dolt-server `bd show --include-dependents` silently empty / errno 1105 swallowed) is therefore fixed on `main`. Tracked in bead `be-kjp7x`.

Thanks all.

PR #4119, opened 2026-05-23
feat(linear): add linear.outbound_state_map for push disambiguation

A3Ackerman:
## Summary

`bd linear sync --push` errors with `linear.state_map type fallback is ambiguous for beads status "in_progress" across Linear states: In Review, In Progress` when a beads status maps to a Linear state type that has multiple workflow states. Linear's default workspace ships with both "In Progress" and "In Review" as `started`-type states, so any push that needs to set a beads `in_progress` issue fails with no escape hatch.

This change adds an opt-in config key that names the exact Linear state for outbound pushes:

```
linear.outbound_state_map.<beads_status> = "<Linear state name>"
```

When set, `ResolveStateIDForBeadsStatus` resolves by case-insensitive name match against the workflow state cache and short-circuits the existing name-then-type logic. Falls back to today's behavior when unset.

## Why

The existing `linear.state_map` is symmetric — used for both inbound (Linear → Beads) and outbound (Beads → Linear) resolution. The type-based fallback is essential for pull (a brand-new "Discovery" `started` state should still pull as `in_progress`) but fatal for push (it can't pick *which* `started` state to send writes to).

A separate outbound map is the smallest user-visible knob that fixes push without weakening pull or asking users to reorder their existing config:

- **Backwards-compatible**: empty map ⇒ no behavior change. All existing tests continue to pass without modification.
- **Push-only**: the field is consulted exclusively from `ResolveStateIDForBeadsStatus`, which is only called from the push path in `internal/linear/tracker.go`.
- **Discoverable**: the existing ambiguity error now points users at the new key by name.

Considered alternatives:
- *First-wins by declaration order in `linear.state_map`*: implicit, depends on map iteration order, surprising. Rejected.
- *Add a `--skip-state-update` push flag*: useful but orthogonal; doesn't let users push the *correct* state.

## Verification

- Targeted tests (new + existing):
  ```
  $ go test -tags gms_pure_go -run '^TestResolveStateIDForBeadsStatus|^TestLoadMappingConfig|^TestDefaultMappingConfig' -v ./internal/linear/...
  ok  	github.com/steveyegge/beads/internal/linear	0.364s   (13 PASS, 0 FAIL)
  ```
  The new tests directly reproduce the bug (two `started`-type states ⇒ ambiguity error without override, resolves to the named one with override) and cover the precedence, case-insensitive name matching, unknown-state-name error path, and the augmented ambiguity error.
- `make build` — clean.
- `golangci-lint run --build-tags gms_pure_go ./internal/linear/...` — `0 issues.`.

## Notes for reviewers

- Naming: I considered `linear.push_state_map` and `linear.write_state_map` but went with `outbound_` to mirror the project's inbound/outbound vocabulary in the linear package (and to leave room for a possible `linear.outbound_priority_map` etc. later without re-naming).
- The new field is `OutboundStateMap` on `MappingConfig`, initialized to an empty map in `DefaultMappingConfig` — same pattern as `ExplicitStateMap`.
- The error path when an outbound state name doesn't match any cached Linear state names the offending config key and value explicitly so the user can fix their `bd config set` line without guessing.

### DIFF
--- a/internal/linear/mapping.go
+++ b/internal/linear/mapping.go
@@ -137,6 +137,15 @@ type MappingConfig struct {
 	// mappings from type-based fallbacks.
 	ExplicitStateMap map[string]string
 
+	// OutboundStateMap maps a Beads status to the Linear workflow state NAME to
+	// use when pushing. Populated from linear.outbound_state_map.<beads_status>
+	// config keys. Used to disambiguate cases where multiple Linear states share
+	// the same state type (e.g. "In Progress" and "In Review" are both
+	// "started"), which would otherwise fail the push with an ambiguity error.
+	// Keys are lowercase beads status strings; values are Linear state names
+	// matched case-insensitively against the workflow state cache.
+	OutboundStateMap map[string]string
+
 	// LabelTypeMap maps Linear label names to Beads issue types.
 	// Key is lowercase label name, value is Beads issue type.
 	LabelTypeMap map[string]string
@@ -167,6 +176,7 @@ func DefaultMappingConfig() *MappingConfig {
 			"canceled":  "closed",
 		},
 		ExplicitStateMap: make(map[string]string),
+		OutboundStateMap: make(map[string]string),
 		// Label patterns for issue type inference
 		LabelTypeMap: map[string]string{
 			"bug":         "bug",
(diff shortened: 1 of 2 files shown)


maphew:
Merged. This was a clean, opt-in change (`linear.outbound_state_map`) with clear tests and no reported objections, so it went through as a maintainer rebase-merge:

- Fetched the PR head commit and cherry-picked it onto current `upstream/main` in an isolated worktree (no semantic changes; the commit content is identical to the original).
- Verified locally: `go build ./...` and `go test -tags gms_pure_go ./internal/linear/...` both pass.
- Pushed the rebased commit back to your `feat/linear-outbound-state-map` branch and squash-merged.

Thanks for the clear writeup, the alternatives-considered section, and the targeted tests reproducing the ambiguity bug directly - made this an easy review.

_claude-fable-5-medium on behalf of matt wilkie_

PR #4133, opened 2026-05-23
fix(dolt): disable background workers and drain MySQL handshake before TCP close

### LINKED ISSUE #4132: windows: dolt sql-server silently crashes during idle and under TCP connect+close bursts
## Symptoms

The dolt sql-server started by `bd dolt start` (or auto-started by `bd create`) silently exits 2–5 minutes after startup. The OS process disappears without any fatal, panic, or error message in the log. After the crash:

- `bd dolt status` reports **running** (stale — the PID/port files still exist but the process is gone).
- `bd create` fails with **"Dolt server unreachable at ... i/o timeout"**.
- `ps aux | grep dolt` shows **no process**.
- The next `bd create` auto-starts a new server that works temporarily, then crashes again.

## Steps to reproduce

1. Start the server: `bd dolt start`
2. Verify it works: `bd create "test" --type chore --json` (succeeds)
3. Wait 2–3 minutes without touching the server.
4. Try again: `bd create "test2" --type chore --json`
5. Observe: **i/o timeout error**

## Environment

- **OS**: Windows 11 (x64)
- **Dolt version**: 2.0.3 (oldest; latest is 2.0.6)
- **Beads version**: latest `main`
- **Mode**: shared server (default) or per-project

## Log evidence

The server log (`~/.beads/shared-server/dolt-server.log`) shows two distinct crash patterns:

### Crash pattern 1: TCP RST flood (pollution, likely not lethal alone)

```
Cannot send HandshakeV10 packet: write tcp 127.0.0.1:55777->127.0.0.1:62249:
  wsasend: An established connection was aborted by the software in your host machine.
```

Root cause: beads' `waitForReady()` probes the server by doing `net.DialTimeout("tcp", addr)` + `conn.Close()` in a loop (up to 20× in 10 seconds). The `Close()` sends TCP RST, which dolt interprets as an aborted MySQL handshake. The same pattern exists in the circuit breaker `probe()` (`internal/storage/dolt/circuit.go`).

### Crash pattern 2: Background workers crashing (lethal)

```
# No error, no fatal, no panic message at all
# Process simply vanishes during idle
```

Root cause: dolt's default configuration enables `event_scheduler: ON` and `auto_gc_behavior.enable: true`. On Windows 2.0.3, these background goroutines silently crash the entire process. The server log stops abruptly with no shutdown message.

A field report captured 14 "Starting server with Config" restarts across a single session, each preceded by ~60 failed HandshakeV10 attempts in ~60 seconds, then the process exiting silently.

## Impact
- **Severity**: High
- **Reproducibility**: Nearly 100%. Every server started on dolt 2.0.3 on Windows crashes within 5 minutes under normal use.
- **User-visible**: Every `bd create` fails with i/o timeout after the server dies. The only workaround is to restart the server before every command.
- **Scope**: All Windows users with dolt < 2.0.6.

## Fix in flight

A fix branch ([`fix/dolt-server-crash-tests`](https://github.com/Zireael/beads/tree/fix/dolt-server-crash-tests)) addresses both causes:

1. **YAML config disabling background workers**: beads now writes a `dolt-server-config.yaml` with `event_scheduler: "OFF"` and `auto_gc_behavior.enable: false`, passed via `--config` to dolt `sql-server`.
2. **Non-RST readiness probes**: `waitForReady()` and circuit breaker `probe()` now drain the MySQL handshake packet before `Close()`, allowing the TCP stack to send FIN instead of RST.

Zireael:
Fixes #4132.

## What was broken

The dolt sql-server silently crashes on Windows 2-5 minutes after startup under normal `bd create` use. Two independent causes:

1. **Default background workers crash the process** — dolt enables `event_scheduler: ON` and `auto_gc_behavior.enable: true` by default. On Windows these background goroutines silently terminate the process during idle.
2. **TCP RST flood poisons the log and weakens the server** — beads' `waitForReady()` and circuit breaker `probe()` did bare `Dial()` + `Close()` up to 20 times, each sending TCP RST that dolt interprets as an aborted MySQL handshake. The log floods with "Cannot send HandshakeV10 packet: connection aborted."

## What changed

- **YAML config disables background workers** — `Start()` now writes `dolt-server-config.yaml` with `event_scheduler: "OFF"` and `auto_gc_behavior.enable: false`, passed via `--config` to dolt `sql-server`.
- **MySQL handshake drain prevents RST** — both `waitForReady()` and the circuit breaker `probe()` now read (drain) the MySQL handshake packet before `Close()`, so the TCP stack sends FIN instead of RST.
- **Tests** — 4 unit tests and 3 integration tests covering idle stability (90s), connection burst survival (100 rapid connect+close), and combined burst+idle.

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -705,6 +705,35 @@ func EnsureRunningDetailed(beadsDir string) (port int, startedByUs bool, err err
 // trace, debug, info, warning, error, fatal.
 const doltServerLogLevel = "warning"
 
+// doltServerConfigFileName is the YAML config file written by Start and passed
+// to `dolt sql-server --config`. It disables background workers (event_scheduler,
+// auto_gc) that are known to crash on Windows, and sets log_level and listener
+// host/port. See gh-3874-dolt-background-workers-default-on.
+const doltServerConfigFileName = "dolt-server-config.yaml"
+
+// doltServerConfigTmpl is the minimal YAML config template for dolt sql-server.
+//
+// Dolt's --config flag overrides ALL other CLI args (host, port, log_level,
+// behavior), so these fields must all be present when using this approach.
+//
+// The behavior section disables background workers that crash on Windows:
+//   - event_scheduler: OFF — prevents background event scheduler from crashing
+//   - auto_gc_behavior.enable: false — prevents auto-GC from crashing
+//
+// Host and port are filled with FormatInt because the listener config must
+// match the ephemeral port allocated by Start's TOCTOU retry loop.
+func buildDoltServerConfigYAML(host string, port int, logLevel string) string {
+	return fmt.Sprintf(`log_level: %s
+behavior:
+  event_scheduler: "OFF"
+  auto_gc_behavior:
+    enable: false
+listener:
+  host: %q
+  port: %d
+`, logLevel, host, port)
+}
+
 // buildDoltServerArgs returns the argv passed to `dolt` (excluding argv[0]/
 // the binary itself). It is factored out of Start so it can be asserted on
 // in unit tests without spawning a real server.
@@ -738,6 +767,44 @@ func buildDoltServerArgs(host string, port int, debug bool, profDir string) []st
 		args = append(args, "--loglevel="+doltServerLogLevel)
 	}
 	return args
+}
+
+// buildDoltServerArgsWithConfig returns the argv passed to `dolt` (excluding
+// argv[0]) when using a YAML config file. Dolt's --config flag overrides ALL
+// other sql-server CLI args (-H, -P, --loglevel), so those MUST be omitted.
+// Global dolt flags (--prof, --prof-path) are preserved since they are parsed
+// before the subcommand.
+//
+// The YAML config file must have been written by writeServerConfigFile and
+// contain the matching host, port, log_level, and behavior overrides.
+func buildDoltServerArgsWithConfig(host string, port int, debug bool, profDir, configFile string) []string {
+	var args []string
+	if debug {
+		args = append(args, "--prof", "cpu", "--prof-path", profDir)
+	}
+	args = append(args,
+		"sql-server",
+		"--config", configFile,
+	)
+	return args
+}
+
+// configPath returns the path to the YAML config file for the dolt sql-server,
+// written alongside the PID and port files in the beadsDir.
+func configPath(beadsDir string) string {
+	return filepath.Join(beadsDir, doltServerConfigFileName)
+}
+
+// writeServerConfigFile writes the YAML config file for dolt sql-server,
+// overriding listener host/port and disabling background workers.
+// Returns the path to the config file. The caller is responsible for cleanup.
+func writeServerConfigFile(beadsDir, host string, port int) (string, error) {
+	yamlContent := buildDoltServerConfigYAML(host, port, doltServerLogLevel)
+	path := configPath(beadsDir)
+	if err := os.WriteFile(path, []byte(yamlContent), 0600); err != nil {
+		return "", fmt.Errorf("writing server config: %w", err)
+	}
+	return path, nil
 }
 
 // Start explicitly starts a dolt sql-server for the project.
@@ -876,7 +943,14 @@ func Start(beadsDir string) (*State, error) {
 				actualPort = p
 			}
 
-			cmd := exec.Command(doltBin, buildDoltServerArgs(cfg.Host, actualPort, debug, profDir)...) //nolint:gosec // doltBin is resolved from PATH, not user input
+			// start dolt sql-server using YAML config (disables background workers)
+			configFile, cfgErr := writeServerConfigFile(beadsDir, cfg.Host, actualPort)
+			if cfgErr != nil {
+				lastErr = cfgErr
+				break
+			}
+
+			cmd := exec.Command(doltBin, buildDoltServerArgsWithConfig(cfg.Host, actualPort, debug, profDir, configFile)...) //nolint:gosec // doltBin is resolved from PATH, not user input
 			cmd.Dir = doltDir
 			cmd.Stdout = logFile
 			cmd.Stderr = logFile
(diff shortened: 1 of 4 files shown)


maphew:

Zireael:
Hi, I'll give it a try. The issue comes and goes and I haven't identified a clear pattern. There are days when I have AI working through beads epic and when it is trying to interact with beads, the server keeps dying. Chaining bd dolt start > bd <command> works then, but give it few seconds and beads is down again.
I'm on windows and have `dolt.shared-server: true`

hamchowderr:
+1 to landing this. Independent reproduction on **Windows 11** (`bd` 1.1.0 / `dolt` 2.1.8) — I posted process + log evidence on #4132 ([issuecomment-5014617950](https://github.com/gastownhall/beads/issues/4132#issuecomment-5014617950)): the `HandshakeV10 ... connection aborted` storm from the bare `Dial()`+`Close()` probes, plus `dolt sql-server` throwaways piling up on random ephemeral ports with dead parents. Matches both root causes described here.

Happy to build this branch and test it on my setup — heavy concurrent `bd` load across ~40 project DBs, which reliably triggers the leak — and report back if that helps it get reviewed.

hamchowderr:
Built and exercised this branch on **Windows 11** (bd built from the PR, dolt 2.1.8, Go 1.26.3). The lifecycle integration tests are `//go:build integration && !windows`, so reporting what runs natively here.

**Windows unit tests pass** (`go test ./internal/doltserver/... ./internal/storage/dolt/...`):
- `TestWaitForReadyEmitsRST` — new `waitForReady` makes **1 TCP connection (drained, no RST)** vs the old up-to-20. This directly targets the `Cannot send HandshakeV10 packet … connection aborted` storm I reported in #4132.
- `TestCircuitBreakerProbeSendsRST` — probe drains the MySQL handshake before close.
- `TestBuildDoltServerArgsNoConfig` — confirms that without `--config`, dolt runs with `event_scheduler=ON` / `auto_gc=ON` (the crash precondition).
- `TestIsRunningNoSQLPing` — detects the "TCP alive but SQL engine dead" state.

Confirmed by inspection that `Start()` now writes `dolt-server-config.yaml` (`event_scheduler: "OFF"`, `auto_gc_behavior.enable: false`) and launches `dolt sql-server --config …`. I couldn't capture that live on Windows because shared-server mode pins port 3308 (already held by my running server, which the new binary correctly refused to stomp) — but the arg-selection is covered by the unit test above. Happy to run a free-port behavioral check if that'd help.

For context, the bug is very much live for me: mid-session my shared dolt server silently died and restarted (PID changed), and throwaway `dolt sql-server` processes re-accumulated on random ephemeral ports — matching both failure modes this PR targets. +1 to landing it.

maphew:
First: sixty-four days without a maintainer response to a correctly-diagnosed bug with two independent reproductions is a failure on our side, full stop. I'm sorry. The diagnosis is confirmed real: the RST-on-close mechanism — a bare `Dial()`+`Close()` discarding the MySQL greeting so the server's `HandshakeV10` lands on a dead socket — was not understood by anyone else before this PR, and @hamchowderr's Windows 11 reproduction on #4132 corroborates both root causes you named. You found something genuine.

The honest picture after a full review, in two halves:

**The `--config` YAML half was independently rebuilt on main while this sat unreviewed** — that's on us, not you. Main's version (`buildDoltServerYAMLConfig` + `resolveCfgDir`) had to solve a problem there was no way for you to know about: passing `--config` makes dolt skip its own `.doltcfg` discovery, silently abandoning existing `privileges.db`/`branch_control.db` (#4986 — a data-loss class bug), so main replicates the discovery, version-gates the whole path on dolt ≥ 1.52.1 (older dolt hard-fails on unrecognized YAML keys via strict parsing), and deliberately leaves auto-GC *on* (only `archive_level: 0`) rather than disabling it outright. Your config half retires as superseded by that — no fault in the idea; main just had to go further.

**The handshake-drain half is the surviving novel value, and we want it** — with one important redirection. The two sites you wired it to are the *least* frequent probes: `waitForReady` opens one successful connection per server start, and the breaker probe fires only on cooldown expiry. The probe that matches the storm cadence in hamchowderr's report (~10 conns/sec, editor hooks + parallel agents) is the **fail-fast connectivity check on every store open** — `internal/storage/dolt/store.go` does a bare `Close()` on every `bd` command. The drain belongs there (plus `dialProbe`, the doctor probes, and your two sites), as one shared helper rather than the duplicated block. Two smaller fixes in the drain itself: on read-timeout it currently closes anyway — the greeting then lands on a closed socket, recreating the exact symptom in the startup window — so a failed drain should re-poll rather than declare ready; and the lint blocker (`host`/`port` unused in the new args builder; `debug` also dropped `--loglevel=debug`, which is the diagnostic channel this very bug needs) is what's holding all four red checks.

**Your call before anyone touches your branch**: (a) narrow this PR to the drain half yourself with the redirection above — it stays yours and we'll re-review fast, promised; or (b) we land the drain maintainer-side with `Co-authored-by: Zireael`, applied across all the probe sites, and close this with a pointer. Either answer is fine; after the wait we owed you the choice rather than a fait accompli.

@hamchowderr — your offer of a live behavioral check is exactly what's needed, with one refinement: the unit tests you ran can't distinguish the fix (three of the four are log-only stubs, and the connection-count test passes identically on unfixed code). The measurement that would settle it: against a live dolt on a free port, compare greeting-drained close vs bare close over N probe cycles, counting `HandshakeV10 ... connection aborted` lines in the server log. If drained ≈ 0 and bare ≈ N, the mechanism is proven end-to-end.

One more thing from your report that this PR deliberately doesn't fix and shouldn't absorb: the orphaned `dolt sql-server` processes with dead parents deserve their own issue — main's sweep machinery currently only reaps test debris by design. We're tracking that separately.

maphew:
Following up on the disposition above: the (a)/(b) window closed today with no reply, so the drain work has been landed maintainer-side as #5277, exactly as promised — your drain logic verbatim as a shared `DrainAndCloseProbe` / `ProbeSQLServer` helper swept across every MySQL-speaking probe site (`waitForReady`, `dialProbe`, the store fail-fast probe, the circuit breaker, and the doctor checks), with the re-poll fix for the TCP-up-but-not-greeting case and your fake-greeting test adapted alongside new coverage. Your authorship is carried as `Co-authored-by`, and @hamchowderr's Windows repro is credited in the PR body.

The config half of this branch is not ported: main's #4986 YAML-based server config superseded it (and already fixes the lint findings that half carried).

Closing this PR in favor of #5277. If you're back and want to pick anything up, comment there — the orphaned-server angle you documented in #4132 is tracked at #4282 and is still open for the taking. Thanks for a correctly-diagnosed bug and a clean fix, twice reproduced.

PR #4147, opened 2026-05-24
fix: make dolt push fsck timeout non-fatal

Wldc4rd:
## Summary

- Treat a pre-push `dolt fsck --quiet` timeout as an incomplete check instead of a dangling-reference corruption finding when no integrity-failure output was captured.
- Add `BEADS_DOLT_FSCK_TIMEOUT` and `BEADS_DOLT_PRE_PUSH_FSCK` controls for operators with large Dolt stores.
- Serialize Beads-managed CLI pushes per local Dolt repo with `.dolt/beads-push.lock` to avoid multiple local agents running `fsck`/`push` against the same store at once.

## Problem

Large Dolt stores can legitimately fail to complete `dolt fsck --quiet` inside the existing fixed timeout. Before this change, any fsck process failure except a narrow "could not open" case was wrapped as `ErrDanglingReference`, so a timeout or cancellation with empty output could surface to users as:

```text
dangling chunk reference: aborting push to prevent propagating corrupt chunks
```

That message implies confirmed local corruption even when fsck did not finish and did not report corrupt chunks.

## Changes

- Keeps completed fsck integrity failures fatal.
- Skips timed-out fsck checks when output does not indicate an integrity failure, with an operator warning.
- Allows disabling the pre-push fsck with `BEADS_DOLT_PRE_PUSH_FSCK=0`.
- Allows tuning the timeout with `BEADS_DOLT_FSCK_TIMEOUT`.
- Adds a per-repo push lock, configurable with `BEADS_DOLT_PUSH_LOCK` and `BEADS_DOLT_PUSH_LOCK_TIMEOUT`.
- Applies the lock to normal CLI pushes and federation peer CLI pushes.

## Testing

```text
go test -tags gms_pure_go ./internal/storage/dolt -count=1
ok github.com/steveyegge/beads/internal/storage/dolt 22.996s

go test -tags gms_pure_go ./cmd/bd -run '^TestDoltPushPullCommitNeedStore$|^TestDoltRemoteSubcommandsNeedStore$' -count=1
ok github.com/steveyegge/beads/cmd/bd 0.185s

git diff --check
```

## Notes

This is intentionally a narrow fix. It prevents Beads from reporting an incomplete fsck as confirmed dangling-reference corruption and reduces local push concurrency. It does not claim to fix every Dolt push failure mode; real `dolt push` failures still return as push errors after the fsck phase.

### DIFF
--- a/internal/storage/dolt/errors.go
+++ b/internal/storage/dolt/errors.go
@@ -35,6 +35,10 @@ var (
 	// prevent propagating the corruption to the remote. Run bd dolt verify
 	// to diagnose and recover.
 	ErrDanglingReference = errors.New("dangling chunk reference")
+
+	// ErrDoltPushInProgress indicates another local process is already running
+	// a Dolt CLI push for the same database directory.
+	ErrDoltPushInProgress = errors.New("dolt push already in progress")
 )
 
 // isTableNotExistError returns true if the error indicates a MySQL/Dolt
--- a/internal/storage/dolt/federation.go
+++ b/internal/storage/dolt/federation.go
@@ -468,18 +468,20 @@ func (s *DoltStore) doltCLIPushToPeer(ctx context.Context, peer string, creds *r
 // doltCLIPushRefToPeer shells out to `dolt push` with a specific refspec.
 // The refspec can be a branch name or a "local:remote" mapping.
 func (s *DoltStore) doltCLIPushRefToPeer(ctx context.Context, peer string, refspec string, creds *remoteCredentials) error {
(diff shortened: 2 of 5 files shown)


Wldc4rd:
Closing as stale — branch is ~9195 commits behind main and the dolt storage layer (store.go/errors.go/federation.go) has been substantially refactored since. If the push-fsck-timeout-fatal issue still reproduces on current main, I'll re-file a fresh fix against current main rather than resurrect this branch.

PR #4148, opened 2026-05-24
fix: cache CLI remote listing

Wldc4rd:
## Summary
- cache parsed `dolt remote -v` results per Dolt db path with a 30 minute TTL
- negative-cache non-Dolt directories and transient CLI remote-list failures to avoid repeat fork storms
- invalidate the cache after CLI remote add/remove

## Why
`dolt remote -v` loads large Dolt stores into memory even when it only prints remotes. In Gas City, ordinary write/list paths can call this repeatedly, which turns a busy hq store into a multi-GB fork storm and contributes to SQL stalls/OOM pressure.

## Tests
- `/home/charlie/.local/go/bin/go test ./internal/storage/doltutil`

### DIFF
--- a/internal/storage/doltutil/remotes.go
+++ b/internal/storage/doltutil/remotes.go
@@ -1,9 +1,16 @@
 package doltutil
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
+	"os"
 	"os/exec"
+	"path/filepath"
+	"strconv"
 	"strings"
+	"time"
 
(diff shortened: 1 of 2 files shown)


Wldc4rd:
Closing as superseded. Remote caching was reimplemented upstream in the internal/remotecache package (Cache/CacheMeta + TTL + XDG cache dir), which covers this PR's intent more completely than this 107-line patch to remotes.go. The branch is ~9200 commits behind main and conflicts on the refactored remotes.go. If a real gap remains, I'll re-file fresh against current main.

PR #4150, opened 2026-05-24
perf(id_parser): narrow SearchIssues to id-only projection

johnzook:
## What

Fixes a performance issue first identified under heavy load where it led to a 2× improvement and benchmarked at 25% improvement on a mostly idle host (~7× over `SkipLabels` in microbench). Adds a narrow-projection variant `SearchIssueIDs` to the Storage/Transaction interfaces and switches `ResolvePartialID`'s two substring fallbacks (issues at `id_parser.go:124`, wisps at `:173`) to use it. Both projections now share a single generic core in `internal/storage/issueops/search.go`.

## Why

`ResolvePartialID`'s fallback path emits a 45-column SELECT against issues+wisps, then iterates the results reading only `.ID`. Every other column — description, design, acceptance_criteria, notes, metadata, payload, ... — is read from disk, marshaled over the MySQL wire, parsed by the driver, populated into `types.Issue` structs, and garbage-collected without ever being touched. On any `bd show <hash>` that misses the prefix-match path, that's ~25% of wall-clock burned for nothing.

`bd show <nonexistent-hash>` (forces the full-scan substring fallback — worst case this PR targets), `hyperfine`, 100 runs, 5 warmup, vs. the full-hydration baseline:

| DB | Issues | Baseline (`8ae4c3c67`) | This PR | Speedup |
|---|---:|---:|---:|---:|
| gascity | 7,991 | 112.2 ± 4.4 ms | 88.9 ± 2.4 ms | **1.26×** |
| gc-toolkit | 7,512 | 109.6 ± 3.2 ms | 87.9 ± 2.8 ms | **1.25×** |

Variance also tightens (σ ~4 ms → ~2.5 ms), consistent with shipping fewer columns over the wire.

### Why this and not just `SkipLabels`?

The obvious alternative is to leave `SearchIssues` in place and pass `filter.SkipLabels = true` from the resolver. `SkipLabels` suppresses the labels JOIN/hydration but still SELECTs the full `IssueSelectColumns` (~45 columns) and scans every row — including `description`, `design`, `notes`, `metadata` — into `*types.Issue` structs. The narrow projection skips column scanning, struct allocation, and labels entirely.

`BenchmarkPerfIDProjectionVsHydration_5K` runs all three arms on one identical 5,000-row fixture with heavy TEXT fields populated, query matching the full set (the unbounded LIKE the resolver actually issues):

| Arm | Time/op | Mem/op | Allocs/op |
|---|---:|---:|---:|
| `SearchIssueIDs` (narrow) | **8.95 ms** | **1.0 MB** | **15,129** |
| `SearchIssues` + `SkipLabels` | 65.5 ms | 41.0 MB | 405,177 |
| `SearchIssues` (full) | 521.3 ms | 44.1 MB | 480,988 |

Narrow is **~7.3× faster** than `SkipLabels` and uses **~41× less memory** on the same result set. `SkipLabels` removes the JOIN cost (521 → 65 ms); the projection additionally removes the heavy-column materialization that `SkipLabels` cannot reach (65 → 9 ms, 41 MB → 1 MB).

### Structural integrity

Sharing one generic core closes a latent drift risk between the two projections. The new structure encodes issue vs id-only behavior in a `searchProjection[T]` literal (columns + scanner + dedup key + optional hydrate hook), so adding a future narrow projection means adding one literal — not a parallel top-level function or wisp-merge wrapper, which is how the two paths drifted historically. Previously e.g. upstream's empty-wisps probe and the "id exists in both" hard-error landed in one path but not the other.

Orthogonal to the LIKE/title-search question: `LOWER(title)` and the leading wildcards stay; only the projection narrows. `doltTransaction.SearchIssueIDs` is a correctness stub that delegates to `SearchIssues`, since its hot caller is read-your-writes inside a tx, not partial-ID resolution.

## Verification

- `TestSearchIssuesAndSearchIssueIDs_Parity` (new, `internal/storage/dolt/search_parity_test.go`) — runs both APIs through nine representative filter scenarios (status, priority, substring search, label-driven DISTINCT JOIN, ephemeral routing, limit) and asserts ID-set equality. The contract that catches future drift between the projections.
- `BenchmarkPerfIDProjectionVsHydration_5K` (new) — the `SkipLabels`-baseline microbench above:
  ```
  go test ./internal/storage/dolt/ -run '^$' \
    -bench BenchmarkPerfIDProjectionVsHydration_5K -benchmem -benchtime 30x
  ```
- Manually verified: full-ID exact match, partial-hash match through fallback (issues path), wisp partial-hash match, and "no match" error path all behave identically.
- Lint clean (`golangci-lint run` via pre-commit hook).
- Reproduce the end-to-end hyperfine numbers above:
  ```
  hyperfine -i --warmup 5 --runs 100 \
    --command-name baseline ./bd-baseline\ show\ deadbeef \
    --command-name new      ./bd-new\ show\ deadbeef
  ```
  (where `bd-baseline` is `make build` at the merge base and `bd-new` is `make build` at this branch's tip)

### DIFF
--- a/internal/jira/tracker_test.go
+++ b/internal/jira/tracker_test.go
@@ -660,6 +660,9 @@ func (s *configStore) SearchIssuesWithCounts(_ context.Context, _ string, _ type
 func (s *configStore) SearchIssues(_ context.Context, _ string, _ types.IssueFilter) ([]*types.Issue, error) {
 	return nil, nil
 }
+func (s *configStore) SearchIssueIDs(_ context.Context, _ string, _ types.IssueFilter) ([]string, error) {
+	return nil, nil
+}
 func (s *configStore) AddDependency(_ context.Context, _ *types.Dependency, _ string) error {
 	return nil
 }
--- a/internal/storage/dolt/dolt_benchmark_test.go
+++ b/internal/storage/dolt/dolt_benchmark_test.go
@@ -12,6 +12,7 @@ package dolt
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -995,11 +996,15 @@ func insertBenchIssues(ctx context.Context, tx *sql.Tx, table string, issues []*
 			if metadata == "" {
 				metadata = "{}"
 			}
-			placeholders = append(placeholders, "(?, ?, ?, '', '', '', '', ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)")
+			placeholders = append(placeholders, "(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)")
 			args = append(args,
 				issue.ID,
 				issue.ContentHash,
 				issue.Title,
+				issue.Description,
+				issue.Design,
+				issue.AcceptanceCriteria,
+				issue.Notes,
 				issue.Status,
 				issue.Priority,
 				issue.IssueType,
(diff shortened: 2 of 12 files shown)


maphew:

Review found no correctness defects. The generic `searchProjection[T]` refactor faithfully preserves the wisp-merge, dedup, and Pattern-B id-shrink semantics; the `SkipWisps` hatch survives; and all five concrete Storage/Transaction implementers are covered, so no build break. The July 2 rebase means the diff matches current main exactly, including the sqlbuild extraction and the prefer-wisp cross-table-dup logic that landed since May. The core claim holds up: `ResolvePartialID`'s fallback really was hydrating the full ~45-column row (heavy TEXT included) just to read `.ID`, and `SkipLabels` alone does not fix that.

Non-blocking notes for the record: `doltTransaction.SearchIssueIDs` is a documented full-hydration stub, fine while it has no production caller but worth remembering if a hot path ever reaches for the tx variant; the transaction-level projections and sql-server mode are not covered by the parity test.

Plan: we are at v1.1.0-rc.2 and holding main until stable ships. Because this rewires the search core shared by every list/show/search command, it lands after the smaller PRs in your batch, post-stable, once the newly approved CI runs are green.

PR #4151, opened 2026-05-25
feat(dolt): dolt.local-only enforcement — push/pull/remote-add/status guards (be-3v2ou)

quad341:
## Summary

Implements the `dolt.local-only` config key (be-3v2ou, FR-01 to FR-05) that disables all remote Dolt sync when set to `true`.

- **`bd dolt push`**: exits 0 with `Remote sync is disabled for this project (dolt.local-only=true).` + hint to unset; `--json` emits `{\"status\":\"disabled\",\"reason\":\"dolt.local-only=true\"}`
- **`bd dolt pull`**: same exit-0 no-op, additionally prints `Nothing to pull.`
- **`bd dolt remote add`**: exits 1, writes error to stderr (`cannot add Dolt remote: remote sync is disabled`)
- **`bd dolt status`**: appends `Remote sync: disabled (dolt.local-only=true)` to both embedded and server-mode output
- **`maybeAutoPush`**: guard fires before `getStore()` — no store interaction when local-only
- New `isDoltLocalOnly()` helper in `cmd/bd/dolt_local_only.go` reads `config.GetBool(\"dolt.local-only\")`

## Key name constraint
Config key is `dolt.local-only` (hyphen) to match `gascity`'s `ga-d457b` which sets exactly that key.

## Test plan

- [x] Unit tests: `TestIsDoltLocalOnly_*`, `TestMaybeAutoPush_SkipsWhenLocalOnly` — all pass (pure-Go)
- [x] Integration tests: `TestDoltLocalOnly_Push_*`, `TestDoltLocalOnly_Pull_*`, `TestDoltLocalOnly_RemoteAdd_*`, `TestDoltLocalOnly_ConfigRoundTrip`, `TestDoltLocalOnly_DoltStatus_ShowsDisabled` — all 9 pass (CGO embedded dolt)
- [x] Build clean: `CGO_ENABLED=0 go build -tags gms_pure_go ./cmd/bd/`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4153, opened 2026-05-25
feat(dolt): git-origin collision guard in bd dolt remote add + bd doctor (be-7eu1d)

quad341:
## Summary

- Refuses to add a Dolt remote whose URL matches the git origin (`bd dolt remote add`) — exits 1 with actionable error message and hints
- `--allow-git-origin` flag bypasses the guard with a stderr warning (for monorepo layouts)
- `bd doctor` emits a `CategoryDolt / "Dolt Remote vs Git Origin"` warning check when an existing remote collides
- Adds `doctor.CategoryDolt = "Dolt Storage"` constant + `CategoryOrder` position

## Stacking dependency

**⚠️ Depends on #4151** (`feat/be-3v2ou-dolt-local-only`). The guard fires after the `isDoltLocalOnly()` check — correct ordering is enforced by `TestDoltRemoteAdd_LocalOnly_FiresBeforeGuard`. Please merge #4151 first; this branch is stacked on it.

## Implementation notes

- `dolt_remote_guard.go`: `guardNormalizeURL()` (strips `.git`/trailing slash) + `doltRemoteMatchesGitOrigin()` — reuses `gitOriginGetURL()` from `sync_git.go`
- The existing `normalizeRemoteURL()` in `sync_remote.go` does Dolt-format URL conversion (`https://` → `git+https://`); the guard uses a separate simpler helper to avoid conflating those semantics
- `doctor/remotes.go`: `CheckDoltRemoteGitOrigin()` queries SQL remotes and normalizes for comparison
- `doctor.go`: Check 7f1 wired into the doctor run (warning only, does not set `OverallOK = false`)

## Test plan

- [x] `CGO_ENABLED=0 go build -tags gms_pure_go ./...` — clean
- [x] Unit tests: `TestGitOriginGetURL_*`, `TestGuardNormalizeURL_*`, `TestDoltRemoteMatchesGitOrigin_*`, `TestAllowGitOriginFlag_*`, `TestCategoryDolt_*` — all pass
- [ ] Embedded integration tests (`BEADS_TEST_EMBEDDED_DOLT=1`) — require live Dolt server; deferred to CI
- Pre-existing `cmd/bd` failures (`TestAutoExportGitAddFailureExitsNonZero`, `TestOutputContextFunction`, etc.) confirmed on base branch, not introduced here


quad341:
CI fix — addressed the failing **Check doc flags freshness**:

The new `--allow-git-origin` flag on `bd dolt remote add` wasn't reflected in the generated CLI docs. Ran `./scripts/generate-cli-docs.sh ./bd`; this adds the flag to `docs/CLI_REFERENCE.md` and the `dolt` command reference pages (latest + versioned snapshots). (1df140acd)

The previously-failing check now passes; the rest of the suite is re-running.

maphew (commented):
Thanks for this. The guard is useful and belongs in core, but I do not think this branch should merge as-is yet.

Findings:

1. The collision guard misses supported Dolt Git remote forms. `cmd/bd/dolt_remote_guard.go` only strips `.git` and trailing slashes, and `cmd/bd/dolt.go` uses that helper before adding the remote. That means `git+https://github.com/org/repo.git`, `git+ssh://...`, and SCP-style equivalents can bypass a git origin of the same repository even though those are first-class Dolt remote URL forms.

2. `bd doctor` has the same normalization gap. `cmd/bd/doctor/remotes.go` compares SQL remotes to git origin through its own simple normalization helper, so it can miss the same Dolt-normalized equivalents. The command path and doctor path should share one Dolt-aware comparator.

3. The new embedded guard test file is tagged `embeddeddolt`, but it calls helpers from cgo-tagged test files. `go test -tags 'embeddeddolt gms_pure_go' ./cmd/bd ...` fails to compile with undefined helper symbols, so this coverage is not reliable under its declared tags.

Recommended path: fix-merge after split/restack. The git-origin guard itself is worth preserving, but this PR is stacked on still-open #4151. I would land or fix #4151 separately, then restack this to just the guard, doctor, and docs changes. Keep #4323 as the merge vehicle for the push-only no-push behavior.

I checked the red CI enough to separate signal from noise: generated docs are current, the lint failure is in `cmd/bd/bootstrap.go` and already present on current `upstream/main`, and the embedded routed-store failures appear unrelated to the changed files here.

Checks run locally:

- `bd-main/scripts/pr-preflight.sh 4153 --repo gastownhall/beads`
- `git merge-tree upstream/main refs/remotes/upstream/pr-4153`
- `./scripts/generate-cli-docs.sh --check`
- targeted `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd ...`
- `go test ./cmd/bd/doctor ...`

No code was pushed from this review.

quad341:
@maphew — all three findings addressed, pushed as `67bb5fba2`:

**1. Collision guard URL normalization** — Added `doltremote.CanonicalForComparison()` to `internal/doltremote/remote.go`. It normalizes both sides to Dolt's `git+` prefix form (via the existing `Normalize()`) then strips trailing slashes and `.git`, making `https://github.com/org/repo.git` compare equal to `git+https://github.com/org/repo.git` and `git@github.com:org/repo.git` equal to `git+ssh://git@github.com/org/repo.git`.

**2. Doctor remotes** — `doctor/remotes.go` now imports `doltremote` and uses `doltremote.CanonicalForComparison()` instead of the local `normRemoteURL` helper (removed). Both paths now share one implementation.

**3. Build tag mismatch** — `dolt_remote_origin_guard_embedded_test.go` changed from `//go:build embeddeddolt` to `//go:build embeddeddolt && cgo`. The file delegates to `bdDoltSeparate` which lives in a `cgo`-only file; the previous tag caused a compile failure under `CGO_ENABLED=0 -tags embeddeddolt`.

On the stacking question: the `#4151` premise (removing the duplicate `isDoltLocalOnly`) was already applied as a standalone commit in this branch (`bf43184fe`), so this PR is self-contained and doesn't need `#4151` to land first.

maphew:
Re-reviewed after your updates - all three earlier blockers are genuinely fixed:

1. **URL forms**: `doltremote.CanonicalForComparison()` collapses https/git+https, SCP-style/git+ssh, trailing `.git` and `/` correctly (verified empirically against the equivalence cases from the first review).
2. **Shared comparator**: both the guard (`cmd/bd/dolt_remote_guard.go`) and doctor (`cmd/bd/doctor/remotes.go`) now use the single normalizer; no residual duplicate normalization.
3. **Build tags**: the embedded test files are correctly excluded under `CGO_ENABLED=0 -tags gms_pure_go`; build, vet, and the guard/normalizer/local-only tests all pass.

Since maintainer edits are enabled, I pushed two commits to unblock landing rather than bouncing this back:

- a merge of current `main` to resolve the conflicts that had accumulated (notably `cmd/bd/init.go`, where main still carried the older inline `isDoltLocalOnly` your branch had refactored out to `cmd/bd/dolt_local_only.go` - kept your refactor, kept main's newer `handleRemoteSafetyDecision`).
- removal of the dead `--force` flag on `bd dolt remote remove`: it was registered but never read by the command, yet showed up as real behavior in the generated docs. Dropped the registration and regenerated the CLI docs. If you intended wiring for it, happy to see that as a follow-up instead.

One nit noted for a future follow-up, not blocking: `CanonicalForComparison` doesn't lowercase the host or strip embedded credentials, so e.g. `https://GitHub.com/org/repo` vs `https://github.com/org/repo`, or a PAT-in-URL origin, are false negatives. Those only weaken the advisory guard (fail-safe direction), so fine to leave for later.

With CI green on the updated head this is ready to land. Thanks for the thorough response round, and for `CanonicalForComparison` - the shared comparator is exactly the right shape.

maphew:
One more maintainer commit (3b49ec0): CI's errcheck flagged the two `outputJSONRaw` calls in the `dolt.local-only` status paths - main changed that helper to return an error after this branch was written. Handled with the same stderr idiom as the neighboring `outputJSON` callers. Lint is clean locally; waiting on CI.

maphew:
Follow-up on the comparator gaps flagged during re-review (https://github.com/gastownhall/beads/pull/4153#issuecomment-4883818549): opened #4576, which folds host case, strips embedded `user[:pass]@` credentials, and recognizes the user-less SCP form `host:path` in `CanonicalForComparison`. `http` vs `https` are kept distinct and the org/repo path case is preserved. Tests for all three equivalence classes (plus a negative http/https guard) added to `cmd/bd/dolt_remote_origin_guard_test.go`.

PR #4154, opened 2026-05-25
fix(migration): run schema migrations on a connection with no read timeout

quad341:
## Problem

Schema migrations run over the main connection pool, which sets a **10s `ReadTimeout`** in `buildServerDSN` (`internal/storage/dolt/store.go`). Migrations that touch large tables can far exceed this — e.g. `0047_recompute_mixed_is_blocked` does a full-table recursive-CTE `is_blocked` recompute that takes ~60s on a few-thousand-row database.

When the read deadline fires mid-migration:
- the migration aborts with `read tcp ...: i/o timeout`;
- `schema.MigrateUp`'s `unstagePreExistingTables` has already dirtied the working set, so the table is left **dirty / half-migrated**;
- every subsequent attempt then trips the `pending schema migrations alter pre-existing dirty tables` guard.

Net effect: a database that needs a slow migration becomes **permanently un-upgradeable** — `bd migrate` and store-open both fail in a loop. This bit us recovering a real city whose `hq`/rig stores were a few migrations behind; the only workaround was running migrations out-of-band.

## Fix

`initSchema` now runs the migration pass over a **dedicated single connection with no read/write timeout** (`openMigrationDB`), mirroring the existing `execWithLongTimeout` pattern but unbounded, since a migration has no sensible fixed deadline. Cancellation is governed by the caller's `context`, not a fixed socket deadline. The normal pool keeps its 10s timeout for regular queries.

## Testing

Built with `make install` (CGO=1) and migrated 3 real databases that were stuck at cursor 45/46 → 47:
- gascity (`0047`): completed in **~57s** (previously failed at 10s every time).
- mcdclient, pwu (`0046`+`0047`): completed cleanly.
- All ended at cursor 47 with clean working sets; the city then started normally.

### DIFF
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -1501,7 +1501,19 @@ func initSchemaOnDBWithRetry(ctx context.Context, db *sql.DB) (int, error) {
 }
 
 func (s *DoltStore) initSchema(ctx context.Context) error {
-	_, err := initSchemaOnDBWithRetry(ctx, s.db)
+	// Schema migrations can run arbitrarily long (e.g. full-table recomputes
+	// such as the is_blocked backfill in migration 0047). The main connection
+	// pool sets a 10s ReadTimeout (see buildServerDSN); a slow migration over
+	// that pool aborts mid-flight with "i/o timeout" and leaves tables dirty,
+	// which then blocks every subsequent migration attempt. Run the migration
+	// pass over a dedicated connection with no read/write timeout. Cancellation
+	// is governed by the caller's context, not a fixed deadline.
+	migDB, err := s.openMigrationDB()
+	if err != nil {
+		return err
+	}
+	defer migDB.Close()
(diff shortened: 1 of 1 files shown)


quad341:
Rebased onto current main (post-merge of #4116 `db/migrate-schema`).

**Conflict resolved in `internal/storage/dolt/store.go`:** PR #4116 changed `initSchemaOnDBWithRetry` to return `(int, error)` and added `ApplySchemaMigrations` to `DoltStore`. The rebase integrates both changes:

- `initSchema` now uses `openMigrationDB` (this PR's fix) with adapted `(int, error)` return handling
- `ApplySchemaMigrations` also updated to use `openMigrationDB` — the same read-timeout problem would affect explicit `bd migrate schema` runs, so applying the fix there too is correct

Build clean, unit tests pass (server-dependent tests skip as expected in CI without Dolt running).

PR #4158, opened 2026-05-25
fix(list): disable default truncation when stdout is piped (test fix from PR #4114)

quad341:
## Summary

Maintainer cherry-pick of #4114 with CI test fix applied.

- Includes all commits from #4114 (disable truncation when stdout is piped by @kevglynn, GH#4094)
- Adds: `test(list): update limit_truncation_hint for piped stderr contract` — updates `TestEmbeddedList/limit_truncation_hint` to reflect the new behavior (no hint on piped stderr) rather than the old expectation that the hint always appears

**Root cause of CI failure:** `list_embedded_test.go:205` asserted that the truncation hint appears on stderr when `--limit 2` is set. With GH#4094, `printTruncationHint` now early-returns when `!ui.IsStderrTerminal()`. Since `bdListCapture` always pipes stderr via `exec.Command`, the hint is now correctly suppressed — the test was asserting the old (pre-#4094) behavior.

Closes #4114

## Test plan

- [ ] `Test (Embedded Dolt Cmd 3/20)` CI check passes
- [ ] All other CI checks pass

### DIFF
--- a/cmd/bd/list_embedded_test.go
+++ b/cmd/bd/list_embedded_test.go
@@ -228,12 +228,13 @@ func TestEmbeddedList(t *testing.T) {
 	})
 
 	t.Run("limit_truncation_hint", func(t *testing.T) {
-		// Truncated: --limit < seeded count should emit stderr hint (GH#3212).
+		// GH#4094: hint is suppressed when stderr is not a terminal (piped).
+		// bdListCapture always runs in piped mode, so no hint expected even when truncated.
 		stdout, stderr := bdListCapture(t, bd, dir, "--limit", "2")
-		if !strings.Contains(stderr, "more results matched") {
-			t.Errorf("expected truncation hint on stderr, got:\nstderr: %q\nstdout: %q", stderr, stdout)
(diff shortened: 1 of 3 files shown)


maphew:

kevglynn:
**Reproduction:**

```bash
# Storage: embedded Dolt

# Trigger: pipe bd list output (e.g., to grep, less, a file)
bd list --limit 2 | cat
# Before fix: stderr shows "Showing 2 of N issues. Use --limit to see more."
# After fix: hint suppressed when stdout is piped (not a terminal)

# Interactive (terminal) usage still shows the hint:
bd list --limit 2
# "Showing 2 of N issues. Use --limit to see more."  (on stderr, visible)
```

**Expected**: Truncation hint is suppressed when output is piped (scripts/tooling don't need it)
**Actual (before fix)**: `printTruncationHint` always emits to stderr regardless of terminal state, cluttering piped output for downstream consumers

Note: I'm the original contributor on this (GH#4094); quad341 did the maintainer cherry-pick with CI test fix.

PR #3908, opened 2026-05-12
Add BEADS_MAX_ROWS / --max-rows defensive row cap

quad341:
## What this changes

Adds an **opt-in process-level row-count safety net** to `bd` so an operator can
bound how many rows a query may return. Two coordinated layers ship together:
storage enforcement and CLI wiring.

### Operator surface

- New flag `--max-rows N` on `bd list`, `bd ready`, `bd dep tree`,
  `bd find-duplicates`, `bd graph`.
- New env var `BEADS_MAX_ROWS=N`, honored on the five above plus
  `bd lint`, `bd doctor-conventions`, `bd doctor-pollution` (env-only on
  the doctor family — those have no flag).
- Flag wins over env. `--max-rows 0` is the explicit "disable" form and
  overrides any env setting for that invocation.
- On overage: `bd` exits **2** with a two-line stderr including the limit
  and which source set it (flag vs env). stdout stays empty so jq pipelines
  don't see partial data.
- `bd list --max-rows -1` (or env `=-1`) exits **1** with
  `must be non-negative`. A non-integer env (`=banana`) warns once and
  proceeds normally.
- `bd config show` now lists `BEADS_MAX_ROWS` alongside the other env
  knobs so operators can see what's active.

### What does NOT change

- **Default behavior is byte-identical to today.** `MaxRows=0` /
  `BEADS_MAX_ROWS` unset is the default and means no cap.
- Bulk-data commands (`bd cleanup`, `bd gc`, `bd export`,
  `bd export --auto`, `bd migrate-issues`, `bd jira`) explicitly opt out
  at every `IssueFilter` construction site. Confirmed by smoke test:
  `BEADS_MAX_ROWS=5 bd export` on a 298-issue DB still produces all 299
  lines (no false positives).

## Why one PR

The CLI wiring (`--max-rows`, env resolution, error formatting) is a
trivial shell once the storage layer surfaces `IssueFilter.MaxRows` and
`*issueops.ErrTooManyRows`. Splitting the two would land a public
storage API with no consumer and force a second round-trip; keeping
them together lets a reviewer evaluate one feature end-to-end.

## Review notes

A few non-obvious choices worth a look:

- **Cap placement.** The cap fires *after* the `issues + wisps` merge in
  `SearchIssuesInTx`, not inside `searchTableInTx`. Each leg still
  issues `LIMIT cap+1` for bounded over-fetch — but only the merged
  result can detect union overflow when neither leg individually
[...]

PR #4160, opened 2026-05-25
test(cli,storage): BEADS_MAX_ROWS behavioral + opt-out gate tests (be-x42v)

quad341:
## Summary

Gate tests for the BEADS_MAX_ROWS / `--max-rows` defensive row cap (be-x42v). Depends on PR #3908.

- `be-u8z9`: BEADS_MAX_ROWS non-list command paths + config show (`max_rows_test.go`, `ready_max_rows_test.go`)
- `be-x42v.3`: bd list scenarios + backend parity (`search_max_rows_test.go` for dolt and embeddeddolt)
- `be-x42v.4`: opt-out gate tests for data-integrity commands (`export_auto_max_rows_gate_test.go`, `export_max_rows_gate_test.go`, `jira_max_rows_gate_test.go`, `migrate_issues_max_rows_gate_test.go`, `lite_filter_spy_test.go`)
- `be-x42v.4`: `TestExport_BypassesBeadsMaxRows` behavioral gate
- `be-pc8c`: server-mode BEADS_MAX_ROWS gate for `bd doctor --check={conventions,pollution}` (`doctor_max_rows_test.go`)

## Dependencies

**Stacks on #3908** (Add BEADS_MAX_ROWS / --max-rows defensive row cap). Merge #3908 first; this PR will fail CI until the implementation is present.

Note: `embeddeddolt/cache_cleanup_test.go` build failure is pre-existing on the impl branch, not introduced here.


quad341:
Fixed the failing `Check doc flags freshness`. Stale artifact was `website/static/llms-full.txt` (regenerated by `generate-llms-full.sh`, not `generate-cli-docs.sh`). That script's `cli-reference/*.md` concatenation is `LC_COLLATE`-dependent, so it must be generated under `LC_ALL=C` to match CI's collation. Regenerated with `LC_ALL=C ./scripts/generate-llms-full.sh` (commit `044179dd7`); CI is now green.

quad341:
Fixed the only red check (**Check doc flags freshness**).

Root cause: `website/static/llms-full.txt` didn't include the new `--max-rows` flag, so `generate-cli-docs.sh --check` (which also validates `llms-full.txt`) failed. The generated CLI reference pages were already fresh — only `llms-full.txt` was stale.

Fix: regenerated `llms-full.txt` from the current website docs using a fresh pure-go `bd` built from this branch (the exact binary CI uses). Net change is the added `--max-rows` flag plus flag-help column re-alignment from the longer flag name; all substantive content (incl. `bd init` safety contract) is preserved.

Verified locally with `./scripts/check-doc-flags.sh` → PASS. Commit c50769446.

maphew (commented):
Thanks for this - solid, disciplined test coverage for the `BEADS_MAX_ROWS` opt-out contract. The gate-test pattern (assert every sweep/export/migrate/jira call site pins `MaxRows=0` / `MaxRowsSource=""`) is exactly the kind of regression-proofing that keeps a defensive feature like this from silently leaking into data-integrity paths later, and the doctor-family env-only tests (`doctor_max_rows_test.go`) close a real gap.

**Findings**

- **[blocker] Depends on unmerged, stale #3908.** This PR's diff against `main` carries the full MaxRows implementation (`internal/types/types.go`, `internal/storage/issueops/search.go`, `cmd/bd/max_rows.go`, etc.) because it stacks on #3908, which is itself open, conflicting, has no GitHub reviews yet, and hasn't had on-topic activity since 2026-06-16. There isn't a clean way to land the tests before the feature they test.
- **[major] The merge conflict is real, not superficial.** `git merge-tree` against `upstream/main` shows conflicts in ~30 files, including the core `internal/storage/issueops/search.go`, which has independently drifted 168 insertions / 123 deletions on `main` since this branch's merge-base - versus only 53 lines touched here. That's a sign the branch needs a real rebase against current `search.go`, not just a mechanical re-merge.
- **[minor] CONTRIBUTING.md's new "designer §4" reference.** It resolves today via `release-gates/be-x42v-max-rows-gate.md` in this same diff, but worth reconfirming after rebase that the CONTRIBUTING.md opt-out rule and the release-gate doc still describe the same design once #3908's copies of these files land first (to avoid two slightly different descriptions surviving the merge).

**On the CONFLICTING status:** this is an ordinary conflict, not the pre-rewrite no-common-ancestor case. `git merge-base upstream/main <PR head>` resolves to `8e18581dc0` (2026-06-15), which is on current `main`'s history, after the ~2026-06-10 rewrite - so a plain `git fetch upstream && git rebase upstream/main` is the right remedy, no `--onto` needed.

**Recommendation:** fix-merge, but sequenced behind its dependency. Suggested path: (1) get #3908 rebased onto current `main` and reviewed/merged first - that PR needs its own attention since it's equally stale and has zero reviews; (2) once #3908 is in, rebase #4160 onto the post-merge `main`, which should shrink this diff down to just the new `*_test.go` / `*_gate_test.go` files since the implementation will already be present; (3) re-run CI at that point. Happy to help coordinate the #3908 review if that's the blocker. Thanks again for the thorough gate-test design - it'll be valuable once the sequencing is sorted out.

maphew:
Correction to the [major] conflict finding in my review above: I re-verified with an actual trial merge of this PR's head into current `upstream/main`, and the conflict is much smaller than I stated. It touches **8 files, all under `cmd/bd/`** (`cleanup.go`, `dep.go`, `doctor_pollution.go`, `find_duplicates.go`, `graph.go`, `lint.go`, `list.go`, `ready.go`), not ~30, and `internal/storage/issueops/search.go` is **not** among them - despite main's 168/123-line drift there, git auto-merges it cleanly against this branch's 53-line change, so no manual `search.go` reconciliation is needed. The blocker (sequencing behind #3908) and the merge-base analysis (`8e18581dc0`, plain `git rebase upstream/main`, no `--onto`) both stand as written; expect the rebase itself to be lighter than my earlier comment suggested, with hand-resolution only in those command files.

maphew:
Rebased onto `upstream/main` @ `4f0df4e6a` (13 original commits, replayed clean with in-place conflict resolution; force-pushed, tip verified unmoved before push).

Note: #3908 was closed as superseded, so this PR now stands alone as the single canonical PR carrying both the `BEADS_MAX_ROWS` implementation (`cmd/bd/max_rows.go`, storage-layer cap, CLI wiring) and its test gates.

### Conflicts resolved

- `cmd/bd/{cleanup,dep,doctor_pollution,find_duplicates,graph,lint,list,ready}.go` — main's `FatalError`/`FatalErrorRespectJSON` helpers were removed in favor of `return HandleError(...)`-style errors surfaced through `RunE`. Re-homed every `handleMaxRowsError(err)` call to run immediately before the equivalent `return HandleError...` on current main, instead of the old `FatalError` call.
- `cmd/bd/list.go` — main split `bd list`'s `RunE` body out into `runListCore` (shared with `bd children`) as part of the list.go → list.go/list_filter.go/list_input.go/list_proxied_server.go split (#4300 and later). Moved the MaxRows resolution (`resolveMaxRows` → `filter.MaxRows`/`MaxRowsSource`) and the `handleMaxRowsError` calls into `runListCore` at the same logical points (after `buildListFilter`, before each `SearchIssues`/`GetReadyWork` error return).
- `cmd/bd/dep.go` — both an old `domain` import and the PR's new `issueops` import were needed; kept both.
- `cmd/bd/max_rows.go` (new file, no textual conflict but broken against current main) — `resolveMaxRows`'s usage-error paths called the now-deleted `FatalError`. Replaced with a small local `fatalMaxRowsUsage` helper that reproduces the old jsonOutput-aware exit(1) behavior, since `resolveMaxRows` is called deep inside each command's `RunE` (return-based propagation would mean touching ~7 call sites, out of scope for a rebase).
- `docs/cli-reference/*.md`, `docs/CLI_REFERENCE.md` — auto-merged cleanly.
- `website/**` (multiple `modify/delete` conflicts across 3 commits) — the entire `website/` tree no longer exists on `main`. Accepted deletion throughout (nothing to re-home; the content's new home is `docs/cli-reference/*.md`, which merged separately).

### Fixed after rebase (found by running the PR's own gate tests against post-rebase `main`, not by textual conflicts)

- `internal/storage/issueops/search.go` — main's Q2 perf work (`SkipWisps` early return, landed independently of this PR) skipped `EnforceMaxRowsCap` entirely. `bd list` sets `SkipWisps=true` by default, so `bd list --max-rows N` silently ignored the cap. Enforced the cap on that return path too, mirroring the adjacent empty-wisps-table fix this PR already had.
- `internal/storage/issueops/search_counts.go` — `SearchIssuesWithCountsInTx` (the `--json` path; already present at this PR's merge-base, never wired for MaxRows by the original commits) neither sized its SQL `LIMIT` off the cap nor enforced it. Threaded `filter.MaxRows` through `EffectiveSearchLimit` and added `EnforceMaxRowsCap` to the shared `finishSearchIssuesWithCounts` terminal helper (now returns `error`). Fixes `TestEmbeddedMaxRowsList/Flag_OverCap` (`bd list --json --max-rows 5`).
- `cmd/bd/jira_max_rows_gate_test.go`, `cmd/bd/migrate_issues_max_rows_gate_test.go` — seed `INSERT`s predate migration 0049 (`issues.description`/`design`/`acceptance_criteria`/`notes` → `LONGTEXT NOT NULL`, no default), which landed on `main` after this branch's merge-base. Added empty-string values for the now-required columns.

**Not fixed** (pre-existing gap, no gate test in this PR exercises it): `GetReadyWorkWithCountsInTx` (`bd ready --json` / `bd list --ready --json`) also doesn't thread `MaxRows`. Left alone to keep this rebase to conflict resolution rather than expanding scope — worth a follow-up issue.

### Tests run

```
go build -tags gms_pure_go ./...                                                  # clean
go vet -tags gms_pure_go ./cmd/bd/ ./internal/storage/issueops/                    # clean
go test -tags gms_pure_go -count=1 -run 'MaxRows|FilterHasZeroMaxRows|BypassesBeadsMaxRows' ./cmd/bd/... ./internal/storage/...   # ok (dolt-server mode)
BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go -count=1 -run 'MaxRows' ./cmd/bd/...   # ok (embedded mode, incl. TestEmbeddedMaxRowsList and TestEmbeddedMaxRowsNonListPaths)
go test -tags gms_pure_go -count=1 ./internal/storage/issueops/...                 # ok
```

Fast preflight only, per scope — did not run the full suite.

maphew (changes requested):
## Cross-vendor paired review (claude-opus + codex-gpt-5.6-sol), adjudicated

The rebase re-homing itself is verified sound (no conflict markers, no double-application in `runListCore`, the two gap-fixes in `25e35cc3e` are correct, opt-out gate tests pin the contract via filter-capturing spies, migration-0049 seeds verified against the actual `.up.sql`, env/flag precedence exactly as documented). The findings are about the *feature's coverage of its own promise* — the vendors found almost entirely disjoint holes, which together say the circuit breaker currently only fully works for default-mode `bd list`/`ready`.

### Must-fix before merge

1. **[opus] `bd ready --claim` hard-fails under a rig-wide cap.** `internal/storage/issueops/claim.go:206` resets `claimFilter.Limit` but not `MaxRows`, so with `BEADS_MAX_ROWS=100` and 150 ready issues, claiming exits 2 — in exactly the agent-rig scenario the flag is for, on a command that returns ≤1 row. Reset `MaxRows`/`MaxRowsSource` alongside `Limit` (optionally bound the internal scan via `Limit` instead).
2. **[codex] `ready --json` / `list --ready --json` bypass the cap.** `cmd/bd/ready.go:148-149` routes through `GetReadyWorkWithCountsInTx`, which never enforces; with `--max-rows 3` and 6 ready rows, it returns them all and exits 0 (and the JSON error branch omits `handleMaxRowsError`). This was flagged as a follow-up in the rebase comment, but since this PR *is* the feature, wire it here.
3. **[both] Proxied-server mode silently ignores the cap.** `cmd/bd/list.go:506-508` returns to the proxied path before `resolveMaxRows`, and the proxied repository (`internal/storage/domain/db/issue_search.go`) never reads `MaxRows`. A safety flag that silently no-ops is the worst outcome: either resolve+propagate through the UOW path, or reject `--max-rows`/env-cap under proxied-server with an explicit error.
4. **[both, independently] `--limit N --max-rows N` errors instead of truncating.** `withFetchOneExtra` bumps the storage limit to N+1, so `EnforceMaxRowsCap` fires even though the delivered set is exactly N. Compare against the delivered count or keep the probe row out of the cap check.
5. **[codex] `bd list --watch` swallows the cap error and exits 0.** `cmd/bd/list.go:776-777` — route the initial-query error through `handleMaxRowsError` so exit-code-2 semantics hold.
6. **[opus] `handleMaxRowsError` calls `os.Exit(2)` from a RunE body, stranding queued metrics.** It sits on a path whose caller already does `return HandleError(...)` — return an exit-coded error instead, aligning with main's FatalError-removal philosophy.

### Should-fix or document explicitly

7. **[codex+opus] graph/dep-tree coverage:** `bd graph <issue>` (non-`--all`) never resolves the cap; direct `--all` caps per-status (up to 3×N total, comment acknowledges); `bd dep tree` builds the full tree and only checks afterwards. Minimum: resolve the cap in all graph modes and state the per-status/post-hoc semantics in the flag docs; better: enforce against the final node set.
8. **[codex] `ready --include-ephemeral` scans/hydrates all wisps before the post-merge check** (`internal/storage/issueops/ready_work.go:139-142`) — propagate the effective cap into the wisp query so the breaker actually bounds work.

Item 2 subsumes the previously-filed follow-up for the counts path; the remaining tracked follow-up on our side is only the context-threading unrelated to this PR.

A worker will take this list to the branch; re-review after.

_claude-code-fable-5 adjudicating claude-opus-4-8 + codex-gpt-5.6-sol-high, on behalf of matt wilkie_

maphew:
Follow-up round addressing the adjudicated cross-vendor review (claude-opus + codex-gpt-5.6-sol). Pushed as `8385a7929` on top of the rebase (`25e35cc3e`). Verified each claim against head before implementing.

### Must-fix

1. **`bd ready --claim` hard-fails under a rig-wide cap** — `internal/storage/issueops/claim.go`: `ClaimReadyIssueInTx` reset `Limit` but forwarded `MaxRows`/`MaxRowsSource` into its internal unbounded scan. Now bounds the scan via `Limit = filter.MaxRows` and clears the cap fields, so claim (which only ever consumes one row) can't be hard-failed by a cap sized for bulk reads. Added `TestClaimReadyIssue_MaxRows_LargePoolUnderLowCap` (10-issue pool, cap=3, asserts claim succeeds).
2. **`ready --json` / `list --ready --json` bypass the cap** — confirmed: `GetReadyWorkWithCountsInTx` never enforced `MaxRows`, a gap present since before this branch's merge-base. Threaded `EffectiveSearchLimit` into its query `LIMIT` and added `EnforceMaxRowsCap` via a new `finishReadyWorkWithCounts` terminal helper (mirrors `finishSearchIssuesWithCounts`). Wired `handleMaxRowsError` on the `bd ready --json` error branch. Added `ReadyMaxRowsJSON_FlagOverCap_Exits2` and `ListReadyMaxRowsJSON_FlagOverCap_Exits2`.
3. **Proxied-server silently ignores the cap** — implemented the reject option (not threading through `domain/db`, per scope). Added `rejectMaxRowsUnderProxiedServer`, called on every command that registers `--max-rows` (list, ready, graph, find-duplicates, dep tree) right before it diverts to its `*ProxiedServer` path; an explicit `--max-rows`/`BEADS_MAX_ROWS` now errors with "not supported in proxied-server mode" (JSON-aware) instead of no-oping. `--max-rows 0` still passes through untouched. Documented in `addMaxRowsFlag`'s help text. Added `reject_max_rows_flag`, `reject_max_rows_env`, `allow_max_rows_zero` under `TestProxiedServerList`.
4. **`--limit N --max-rows N` errors instead of truncating** — confirmed: `withFetchOneExtra`'s unconditional `Limit++` crossed `EffectiveSearchLimit`'s `limit > maxRows` branch when `N == maxRows`, tripping the cap on its own truncation-hint probe row. Fixed by skipping the bump when `MaxRows` is active and `Limit >= MaxRows` (only affects `N >= maxRows`; the `limit < maxRows` truncation path is untouched — `EffectiveSearchLimit` doesn't take that branch there either). Added `LimitEqualsCap_TruncatesNotErrors` (+ `--json` variant).
5. **`bd list --watch` swallows the cap error** — `watchIssues` now returns the initial-query error instead of logging it and returning void; `runListCore` routes it through `handleMaxRowsError` for exit 2. Mid-poll refresh failures still log-and-continue (interactive command). Added `Watch_FlagOverCap_Exits2`.
6. **`os.Exit` from inside RunE strands metrics** — confirmed the precedent: `main()` calls `metrics.CloseAndFlush()` unconditionally before checking `exitCodeFromError(err)`, and this package already has an established idiom for non-1 exit codes returned through `RunE` (`list_input.go`'s skip-labels-conflict check: `return in, &exitError{Code: 2}`). Converted `resolveMaxRows` to `(int, string, error)` and `handleMaxRowsError` to return `error` (`&exitError{Code: 2}`), matching that precedent exactly; updated all 6 + 13 call sites. One site (`doctor_conventions.go`'s `runConventionsLint`) is called from an advisory-checks aggregator with no error channel (`() []doctorCheck`, doc-commented "advisory, never error") — gave it its own minimal `([]doctorCheck, error)` signature change, scoped to that one file, to propagate the one genuinely non-advisory failure mode (the cap) without touching the aggregator's advisory contract for actual check results.

### Should-fix (both mechanical)

7. **Graph/dep-tree coverage** — `bd graph <issue>` (no `--all`) never resolved the cap at all. `loadGraphSubgraph` is a BFS over per-ID `GetDependents`/`GetDependencies` lookups (no `IssueFilter` to thread `MaxRows` through), so — mirroring dep tree — added a post-hoc check against the final connected-component node count (the "better" option, not just documentation). Documented graph's per-status (`--all`) and post-hoc (single-issue) semantics, and dep tree's existing post-hoc semantics, in both commands' flag help. Added `GraphSingleIssueMaxRows_NodeCount_Exits2` + `_UnderCap_NoError`.
8. **`ready --include-ephemeral` scans/hydrates all wisps before the check** — confirmed: `getReadyWispsInTx`'s unbounded branch never received `MaxRows` (`readyWorkWispIssueFilter` didn't copy it), so it queried and hydrated the entire wisps table before the post-merge cap check ran. Propagated `MaxRows`/`MaxRowsSource` into that filter so `EffectiveSearchLimit` bounds the wisp query itself to cap+1. Added `ReadyIncludeEphemeralMaxRows_FlagOverCap_Exits2` to confirm the cap still fires end-to-end after bounding the query.

### Tests run

```
go build -tags gms_pure_go ./...                                                   # clean
go vet -tags gms_pure_go ./...                                                     # clean
go test -tags gms_pure_go -count=1 -run 'MaxRows|FilterHasZeroMaxRows|BypassesBeadsMaxRows' ./cmd/bd/... ./internal/storage/...   # ok (dolt-server mode; one testcontainer flake on first run, confirmed non-reproducing on retry, consistent with the earlier rebase round)
BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go -count=1 -run 'MaxRows' ./cmd/bd/...   # ok
BEADS_TEST_PROXIED_SERVER=1 go test -tags gms_pure_go -count=1 -run 'TestProxiedServerList' ./cmd/bd/   # ok
go test -tags gms_pure_go -count=1 ./internal/storage/issueops/...                 # ok
go test -tags gms_pure_go -count=1 -run 'TestList|TestWatch|TestGraph|TestDepTree|TestReady|TestClaim|TestFindDuplicates' ./cmd/bd/   # ok, except one pre-existing unrelated failure (TestListExplicitDBPathRebindsTargetContext — a missing-ICU-header build failure in a subprocess-built test binary, reproduces identically on the pre-follow-up commit 25e35cc3e; not touched by this round)
```

Ready for re-review.

maphew:
Round 3: round-2 delta review regressions. Pushed as `247dd809c` on top of `8385a7929`. Verified each claim against head before implementing — all three confirmed as real regressions from round 2's own fixes.

### Regressions (all confirmed and fixed)

1. **Proxied `bd ready --claim` rejected under a global cap** — confirmed: `usesProxiedServer()` in `cmd/bd/ready.go` called `rejectMaxRowsUnderProxiedServer` before the `--claim` flag was even parsed. Moved the `claimReady` parse earlier and exempt `--claim` from the rejection (bulk, non-claim proxied `bd ready` still rejects — the domain/db UOW layer never reads `MaxRows` at all, confirmed via `grep -rn MaxRows internal/storage/domain/` returning nothing). Added `claim_succeeds_under_env_max_rows_cap` plus `reject_bulk_ready_under_max_rows_{flag,env}` beside the existing proxied test suites; reverting the fix reproduces the reported rejection exactly.
2. **Merged issues+wisps ready page false-trips the cap** — confirmed by reproducing the exact repro (limit=2, cap=3, 2 rows/table → "4 found, --max-rows=3 exceeded" on unpatched code). `GetReadyWorkWithCountsInTx` bounds each table's query independently, so the pre-trim merge can hold ~2x the per-table bound. Mirrored the non-counts path exactly: `GetReadyWorkInTx`'s `mergeReadyWisps` already trims to `Limit` before its caller's `EnforceMaxRowsCap` runs, so I reordered `finishReadyWorkWithCounts` to trim-then-cap instead of cap-then-trim. Verified this is a no-op for every single-table path (a lone query's result never exceeds `Limit` when `Limit>0`, since `EffectiveSearchLimit` already bounds it). Added `ReadyIncludeEphemeralMaxRowsJSON_MergedUnderLimit_NoError`.
3. **`limit==cap` lost the existing >N-matches truncation notice** — confirmed: round 2's fix (skip the `withFetchOneExtra` probe-row bump entirely when `Limit>=MaxRows`) also stopped fetching the extra row, so `len(results)` could never exceed `effectiveLimit` and truncation detection silently broke for this one case. Fixed per the suggested direction: still fetch the probe row (`Limit` bumps to N+1 as normal), but bump `MaxRows` to N+1 in lockstep *only* when `Limit==MaxRows` exactly — this keeps the probe row inside the (temporarily N+1) cap so `EnforceMaxRowsCap` doesn't trip on it, while the CLI's own `len(results)>effectiveLimit` check still sees it and reports `truncated=true`. Tighter/looser-cap cases are unaffected (`MaxRows` stays unbumped there, so a real violation still reports the true cap value in the error message). Note on test coverage: `printTruncationHint`'s text is gated on `ui.IsStderrTerminal()` and is unconditionally suppressed for a subprocess's piped stderr — this is a pre-existing, already-documented constraint in this test harness (`list_embedded_test.go`'s `limit_truncation_hint` subtest states the same thing), so the hint text itself can't be asserted end-to-end here. Verified the fix instead via 4 new unit tests asserting `withFetchOneExtra`'s exact bump values directly (`TestWithFetchOneExtra_LimitEqualsCap_BumpsBothForTruncationProbe` + 3 sibling cases covering tighter-cap/looser-cap/no-limit), and extended `LimitEqualsCap_TruncatesNotErrors`'s comment to document why and point to them.

### Opus nits (both done)

4. **claim.go scan-bounding side effect** — took the preferred option: restored the unbounded internal scan (`Limit=0`) instead of bounding it at `Limit=MaxRows`. A `BEADS_MAX_ROWS=1` rig no longer risks "nothing to claim" just because the single top-of-queue row was transiently unclaimable — the breaker bounds delivered payloads (always exactly one row for claim), not scan size.
5. **find_duplicates.go rejection drift risk** — replaced the inlined `--max-rows`/proxied-mode check with a call to `rejectMaxRowsUnderProxiedServer`, matching the other 4 commands.

### Tests run

```
go build -tags gms_pure_go ./...                                                   # clean
go vet -tags gms_pure_go ./...                                                     # clean
go test -tags gms_pure_go -count=1 -run 'MaxRows|FilterHasZeroMaxRows|BypassesBeadsMaxRows|WithFetchOneExtra' ./cmd/bd/... ./internal/storage/...   # ok (dolt-server mode)
BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go -count=1 -run 'MaxRows' ./cmd/bd/...   # ok
BEADS_TEST_PROXIED_SERVER=1 go test -tags gms_pure_go -count=1 -run 'TestProxiedServerList|TestProxiedServerReady' ./cmd/bd/   # ok (full suites, ~200s)
go test -tags gms_pure_go -count=1 ./internal/storage/issueops/...                 # ok
go test -tags gms_pure_go -count=1 -run 'TestList|TestWatch|TestGraph|TestDepTree|TestReady|TestClaim|TestFindDuplicates' ./cmd/bd/   # ok, except the same pre-existing unrelated ICU-header build failure noted in the round-2 comment (confirmed via stash-based control test to reproduce identically without this round's changes)
```

Each of the three regressions was also confirmed to reproduce on the pre-fix code (git-stashed the relevant file, re-ran the new test, saw the exact failure described in the review, then restored).

Ready for re-review.

maphew:
Round 4: converging. Four small items from the round-3 delta reviews, all adjudicated. Pushed as `77227aa67` on top of `247dd809c`. Verified each claim against head; all confirmed.

1. **(should-fix, opus, reproduced live) Round-3's trim-before-cap fix was asymmetric** — confirmed by reproducing opus's exact repro: `bd list --include-infra --limit 2 --max-rows 5` (3 durable + 3 wisps) gave `"6 found, --max-rows=5 exceeded"` on the pre-fix code in both output modes. `--include-infra` disables the default `SkipWisps` escape hatch, so the wisps merge actually runs and hits the same pre-trim-cap-check bug round 3 fixed in `finishReadyWorkWithCounts` but not on the sibling `bd list` paths. Mirrored the remedy into `internal/storage/issueops/search.go` (`searchInTx`'s four exit points, via a new `trimToSearchLimit` helper) and `search_counts.go` (`finishSearchIssuesWithCounts`, swapped cap/trim order). Also fixed `search_counts.go`'s doc comment, which still asserted the old cap-before-trim rationale and directly contradicted `ready_work_counts.go`'s. Single-source results are unaffected (a lone query never exceeds `Limit` when `Limit>0`, so the trim there is a no-op); `Limit=0` and genuine `Limit>MaxRows` overage still error. Added `IncludeInfraMergedUnderLimit_NoError` (both `bd list` output modes) to `TestEmbeddedMaxRowsList`.
2. **(codex P2) Proxied `ready --claim` silently accepted `--max-rows -1`** — confirmed: the round-3 exemption for `--claim` skipped `resolveMaxRows` entirely, so a malformed value never hit its validation path. Reproduced live (claim succeeded instead of failing with exit 1) on the pre-fix code. Now still calls `resolveMaxRows` in the claim-exempt branch for its validation side effect, ignoring the resolved (positive) cap for the claim itself. Added `claim_rejects_invalid_max_rows_flag` to `ready_proxied_integration_test.go`.
3. **(opus nit) find_duplicates double-resolved max-rows** — confirmed: once at the top of the function (to build its filter) and again inside `rejectMaxRowsUnderProxiedServer`, doubling the malformed-env stderr warning under proxied mode. Split out `rejectResolvedMaxRowsUnderProxiedServer(maxRows int)` for callers that already have a resolved value; `find_duplicates.go` now uses it instead of re-resolving.
4. **(adjudication note, comment-only) claim scan stays unbounded** — added the requested sentence to `claim.go`'s comment: parity with pre-PR main and correctness-first (preserves claim's existing fairness/ordering across the ready set); a bounded, paged scan is a deliberate follow-up, not folded in here.

### Tests run

```
go build -tags gms_pure_go ./...                                                   # clean
go vet -tags gms_pure_go ./...                                                     # clean
go test -tags gms_pure_go -count=1 -run 'MaxRows|FilterHasZeroMaxRows|BypassesBeadsMaxRows|WithFetchOneExtra' ./cmd/bd/... ./internal/storage/...   # ok (dolt-server mode)
BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go -count=1 -run 'MaxRows' ./cmd/bd/...   # ok, including the new IncludeInfraMergedUnderLimit_NoError
BEADS_TEST_PROXIED_SERVER=1 go test -tags gms_pure_go -count=1 -run 'TestProxiedServerReady' ./cmd/bd/   # ok (full suite, ~211s), including the new claim_rejects_invalid_max_rows_flag
go test -tags gms_pure_go -count=1 ./internal/storage/issueops/...                 # ok
go test -tags gms_pure_go -count=1 -run 'TestList|TestWatch|TestGraph|TestDepTree|TestReady|TestClaim|TestFindDuplicates' ./cmd/bd/   # ok, except the same pre-existing unrelated ICU-header build failure noted in prior rounds
```

Both new regression tests (items 1 and 2) were confirmed via `git stash` on the fix to reproduce the reported failure on the pre-fix code, then pass after restoring the fix.

Ready for re-review.

maphew:
Round 5: one item, codex P1, verified live at head `77227aa67`. Pushed as `86b03be76` on top of `77227aa67`.

**Confirmed:** `searchInTx`'s merge branch concatenates two independently `ORDER BY`'d legs (`results = append(filtered, wispResults...)`), and round 4's `trimToSearchLimit` then sliced that unsorted concatenation. With `Limit=N` and durable rows alone filling `N`, a higher-ranked wisp was silently dropped from the page. Reproduced exactly: 3 durable issues at P2 + 1 wisp at P0 (outranks under the default priority-ASC sort), `--limit 1` — the pre-fix code returned a P2 durable row instead of the P0 wisp, in both `bd list` output modes.

**Fix:** added a `less func(a, b T, sortBy string, sortDesc bool) bool` field to `searchProjection[T]`, populated for `issueProjection` via `sqlbuild.Less` (the same Go-side ORDER BY mirror `search_counts.go`'s `sortSearchIssuesWithCounts` already uses correctly). `searchInTx`'s merge branch now re-sorts the concatenation via a new `sortMergedResults` helper before `trimToSearchLimit`, mirroring `finishSearchIssuesWithCounts`'s existing sort-before-trim order. `idProjection` (bare-ID search, no sortable payload) leaves `less` nil — safe today because no current caller combines `Limit>0` with a merged, non-`SkipWisps` ID-only search (`id_parser.go` always passes `Limit=0`); documented this constraint at the field and in the merge-branch comment for future callers. The other three `trimToSearchLimit` call sites in `searchInTx` are single-source (already SQL-ordered), so they didn't need this.

**Checked the mirrored precedents, as asked:** `GetReadyWorkInTx`'s `mergeReadyWisps` already calls `sortReadyIssues(kept, filter.SortPolicy)` before its own `Limit` trim — no bug there. `GetReadyWorkWithCountsInTx` and `SearchIssuesWithCountsInTx` both already sort (`sortIssuesWithCountsByPolicy` / `sortSearchIssuesWithCounts`) before their `finish*` trim-then-cap helpers — no bug there either. Only `searchInTx` had the unsorted-concat-then-trim shape.

**Regression test:** added `IncludeInfraMerged_WispOutranksDurable_InDeliveredPage` to `TestEmbeddedMaxRowsList` (3 durable P2 + 1 wisp P0, `--limit 1`, both output modes) — confirmed it fails against `77227aa67` (durable row returned instead of the wisp) and passes after the fix.

Noted, no action taken: opus's informational note on `query.go`'s 3x over-fetch headroom.

### Tests run

```
go build -tags gms_pure_go ./...                                                   # clean
go vet -tags gms_pure_go ./...                                                     # clean
go test -tags gms_pure_go -count=1 -run 'MaxRows|FilterHasZeroMaxRows|BypassesBeadsMaxRows|WithFetchOneExtra|Search' ./cmd/bd/... ./internal/storage/...   # ok (dolt-server mode)
BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go -count=1 -run 'MaxRows' ./cmd/bd/...   # ok, including the new IncludeInfraMerged_WispOutranksDurable_InDeliveredPage
go test -tags gms_pure_go -count=1 ./internal/storage/issueops/...                 # ok
```

Ready for re-review.

maphew (approved):
## Round 5 (cross-vendor delta review of 86b03be76): approved

Both reviewers report CLEAN-PASS on the final fix:

- **claude-opus:** verified `sqlbuild.Less` mirrors `OrderBy` field-by-field (priority/created/id default, nullable closed/assignee with the same NULL placement, Go-side id sort), that the sort is applied exactly once on the merged exit (single-source exits keep SQL order), that no `idProjection` caller can reach the merged exit with a positive limit, and — decisively — ran the new regression test against the unfixed parent commit: it fails there with exactly the predicted wrong row, and passes at head.
- **codex-gpt-5.6-sol:** independently confirmed the comparator matches per-table SQL ordering for the supported sort keys and that merged pages retain correct top-N membership.

That closes out five review rounds on this PR. Summary of what the rounds hardened beyond the original rebase: claim-path immunity to rig-wide caps, cap enforcement on the ready/list JSON counts paths, explicit proxied-mode rejection (with `--claim` exempted and flag validation preserved), delivered-count semantics everywhere (no false exit-2 on merged pages, truncation notice preserved at limit==cap), RunE-style exit errors that flush metrics, graph/dep-tree coverage with documented semantics, bounded wisp scans, and finally globally-ordered merged pages. Two intentionally deferred items are tracked on our side: paged claim scanning, and the pre-existing `Less`/`OrderBy` NULL-tie nuances shared with the counts path.

Merging once CI completes green on 86b03be76. @quad341 — thanks for the original implementation and gates; this one took real collective effort to land properly.

_claude-code-fable-5 adjudicating claude-opus-4-8 + codex-gpt-5.6-sol-high, on behalf of matt wilkie_

maphew:
CI fix: pushed `9cd899cc4` on top of `86b03be76`. "Check doc flags freshness" failed because it validates `docs/CLI_REFERENCE.md`/`docs/cli-reference/*.md` against `bd` built from the pin in `docs/cli-docs.pin` (v1.1.0), which predates `--max-rows`/`BEADS_MAX_ROWS` — the same pinned-doc-generator trap that hit PRs #3920 and #4022. Downloaded the job's `cli-docs-freshness-patch` artifact and applied it verbatim (`git apply --check` passed cleanly, no manual edits): removed the six generated `--max-rows` doc lines plus the incidental column-width realignment in the `bd graph` flag tables. The flag help strings in `cmd/bd/max_rows.go` are unchanged — those remain the source of truth and will regenerate correctly once the pin advances past this PR.

PR #4162, opened 2026-05-26
fix(show): propagate IterDependents/IterComments errors instead of swallowing them (be-kzipe)

quad341:
## Summary

- `bd show --include-dependents` and `--include-comments` silently dropped iterator setup errors via `if err == nil { ... }` — a non-nil error produced an empty list with no signal
- Both branches now call `FatalErrorRespectJSON` on setup error and on `iter.Err()` after the scan loop
- Fixes be-kzipe; root cause of silent P1 be-kjp7x

## Test plan

- [ ] `CGO_ENABLED=0 go build -tags gms_pure_go ./cmd/bd/` — clean
- [ ] `go test -tags gms_pure_go ./cmd/bd/... -run TestShow` — pass
- [ ] `go vet -tags gms_pure_go ./cmd/bd/` — clean


quad341:
CI was red on two checks; pushed fixes:

1. **Test (ubuntu-latest / macos-latest)** — `internal/jira` failed to build because `configStore` had duplicate `Iter*` method declarations. The `be-z09krv` commit appended a second copy of the nine stub methods after `Close()`, but the same methods already existed before `Close()`; Go rejects duplicate method declarations. Removed the duplicate block (commit `cfd426094`).
2. **Check formatting** — `internal/telemetry/storage.go` had a stray double blank line at line 579 (also from `be-z09krv`). Ran gofmt (commit `20c2fa3ba`).

`go test -tags gms_pure_go ./internal/jira/` passes locally. Ready for another CI pass.

quad341:
Closing as **superseded by #4186**.

This branch (`fix/be-kzipe-show-iter-errors`) had drifted onto the abandoned **be-z09krv** daemon-iter transport — `cmd/bddgen`, `internal/storage/rpc/*`, the `be-z09krv` release gate, plus daemon-iter config/error vars — totalling ~6740 lines across 36 files that never landed on `main` and shouldn't be re-introduced. It was also 19 commits behind `main`, and its only CI failure (`TestEmbeddedInitConcurrent`) was an unrelated flaky Dolt concurrent-init panic.

The intended fix — the `cmd/bd/show.go` IterDependents/IterComments error-propagation hardening — is preserved verbatim in **#4186**, cut fresh from current `main` as a single-file change (+46/-38), build/vet/`TestShow` green. No work is lost.

PR #4163, opened 2026-05-26
fix: prefer wisp on cross-table dup; add CGO-free duplicate check (be-iabdi)

quad341:
## Summary

- **Lookup robustness** — five merge sites in `issueops` (search.go, ready_work.go, ready_work_counts.go, search_counts.go, dependency_queries.go) replaced the hard-error on cross-table ID collision with graceful dedup: the wisp record is preferred (canonical) and the issues-table copy is silently dropped. This fixes the city-wide lookup breakage that occurred when any single ID existed in both tables.

- **CGO-free duplicate detection** — new `CheckCrossTableDuplicates` in the doctor package (no CGO build tag) uses MySQL wire protocol via the existing `fix.openDoltDB` path to count IDs present in both tables. Wired into `bd doctor --check=validate` (reports `StatusError`) and `bd doctor --check=validate --fix` (calls new `fix.CrossTableDuplicates` which deletes stale issues rows and their children, keeping the wisp canonical). Works on any dolt-server backend without requiring an embedded Dolt build.

- **Test updates** — three tests that previously asserted the old hard-error behaviour (`TestSearchIssuesRejectsDuplicateIssueWispID`, `TestLoadStatusByIDInTxErrorsOnIssueWispCollision`, `TestMergeReadyWispsErrorsOnIssueWispCollision`) updated to assert the new graceful wisp-preferring dedup.

Fixes: be-iabdi (issues/wisps id collision: message→wisp migration leaves dup rows; lookup hard-errors on dual-presence)

## Test plan

- [ ] `CGO_ENABLED=0 go build -tags gms_pure_go ./...` builds cleanly
- [ ] `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/issueops/...` passes
- [ ] `CGO_ENABLED=0 go test -tags gms_pure_go -run TestSearchIssuesPreferWispOnDuplicateID ./internal/storage/dolt/...` passes
- [ ] `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd/doctor/fix/...` passes
- [ ] `CGO_ENABLED=0 go vet -tags gms_pure_go ./cmd/bd/...` passes

### DIFF
--- a/cmd/bd/doctor/cross_table.go
+++ b/cmd/bd/doctor/cross_table.go
@@ -1,0 +1,38 @@
+package doctor
+
+import (
+	"fmt"
+
+	"github.com/steveyegge/beads/cmd/bd/doctor/fix"
+)
+
+// CheckCrossTableDuplicates detects IDs that exist in both the issues and
+// wisps tables. Such dups arise from the message→wisp dual-write window
+// (be-iabdi) and block every id-resolver lookup until removed.
+//
+// This check does not require CGO: it queries via MySQL wire protocol using
+// the same dolt-server connection as the fix package.
+func CheckCrossTableDuplicates(path string) DoctorCheck {
+	count, err := fix.CountCrossTableDuplicates(path)
+	if err != nil {
+		return DoctorCheck{
+			Name:    "Cross-Table Duplicates",
+			Status:  StatusOK,
+			Message: "N/A (no database or wisps table absent)",
+		}
+	}
+	if count == 0 {
+		return DoctorCheck{
+			Name:    "Cross-Table Duplicates",
+			Status:  StatusOK,
+			Message: "No cross-table duplicates",
+		}
+	}
+	return DoctorCheck{
+		Name:    "Cross-Table Duplicates",
+		Status:  StatusError,
+		Message: fmt.Sprintf("%d id(s) exist in both issues and wisps", count),
+		Detail:  "Stale issues-table copies break every lookup for the affected IDs",
+		Fix:     "Run 'bd doctor --check=validate --fix' to remove stale issues copies (wisps are canonical)",
+	}
+}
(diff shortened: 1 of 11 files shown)


maphew:

quad341:
Thanks @maphew — for the record, this merged in 72d095c (#4163), so this is a repro for the archive.

**Why a scratch-repo CLI repro was hard to find:** the trigger isn't a `bd` command sequence, it's a **data-corruption state** — the *same ID present in both the `issues` and `wisps` tables*. That state was left behind by the historical message→wisp migration (be-iabdi), which copied rows into `wisps` without deleting the old `issues` row. Once any single ID is dual-present, every lookup that merges the two tables (`SearchIssues`, ready-work, counts, dependency queries) hard-errored — which is what broke city-wide lookups.

**Concrete repro** — committed as a regression test in this PR: `TestSearchIssuesPreferWispOnDuplicateID` (`internal/storage/dolt/mixed_table_lifecycle_test.go:653`). It's the smallest setup that triggers it:

1. create an issue with ID `mixed-search-duplicate` in the `issues` table;
2. insert a row with the **same ID** into `wisps` (distinct title, e.g. `wisp canonical`);
3. call `SearchIssues`.

- **Actual (pre-fix):** step 3 returns a hard error on dual-table presence.
- **Expected (post-fix):** step 3 succeeds and returns the wisp row (canonical); the `issues` copy is silently dropped.

**Env specifics you asked for:**

- **bd version:** any build before 72d095c reproduces; version-independent otherwise.
- **Storage mode:** dolt-server (MySQL wire protocol). The new detector (`CheckCrossTableDuplicates`) is CGO-free, so it runs on any dolt-server backend without an embedded-Dolt build.
- **OS:** not OS-specific.

**To detect/repair a real affected DB:** `bd doctor --check=validate` flags it; `bd doctor --check=validate --fix` deletes the stale `issues` copies (wisps stay canonical).

Issue #4128, opened 2026-05-23
bd 1.0.4 write-path bypasses running dolt sql-server, re-imports JSONL per-call → OOM + lock-fight at scale


IaKalandia:
## Summary

`bd` 1.0.4 write commands (`note`, `label add`, `close`, `update --claim`, `bd-investigate`) trigger a broken **auto-import recovery loop** on every invocation whenever `.beads/issues.jsonl` is present alongside the `.beads/dolt/` directory — even when a `dolt sql-server` is already supervising that same data and the server has all data populated. Each bd CLI write process reads the entire JSONL (~9MB at our dataset), spins up a fresh in-memory dolt context, and re-imports the file before executing the actual SQL — instead of routing the write through the running server.

Reads (`bd ready` / `bd list` / `bd show`) correctly use the running server. Only writes hit the recovery path.

## Environment

- `bd version 1.0.4 (Homebrew)` — current latest tag `v1.0.4` (2026-05-09)
- macOS 25.5.0 (Darwin)
- Dolt sql-server running locally on `127.0.0.1:50538`, port pinned via `.beads/dolt-server.port`
- `.beads/issues.jsonl` size: ~9.2MB (~9,163 issues accumulated)
- launchd-supervised dolt sql-server with `--config .beads/dolt/config.yaml`

## Symptom

At our dataset scale (~9MB JSONL) under multi-agent / cron-fleet write load, the per-call auto-import recovery path:

1. Takes ~76 seconds per `bd note` invocation
2. Uses ~30-50MB transient RAM per process (fresh in-mem dolt context)
3. OOM-kills helper scripts with `Killed: 9` when ~10 concurrent writers spike
4. Triggers `context canceled` errors during label-insert when bootstrap times out mid-import
5. Generates dolt-server log warnings `error="nothing to commit"` from clients that fail mid-tx
6. Compounds with launchd-respawn lock-fights (`database "dolt" is locked by another dolt process`) when the CLI's embed-init tries to lock the data-dir while the running server holds it

We observed sustained dolt sql-server CPU at 700-870% across multiple cores for hours under realistic write workload — entirely due to bootstrap thrash + lock contention, not actual query work.

## Repro

1. Have a populated `.beads/issues.jsonl` (~9MB / ~9k issues) and a running `dolt sql-server` on its data directory.
2. Run any single `bd note <bead> "test"` — observe ~76 second wall-clock time, with `auto-importing X bytes into empty database` output despite the running server having all data.
[...]

Issue #3880, opened 2026-05-11
Server mode: repeated 'auto-import ... into empty database' on every update in shared-server Dolt setup


Wermeling:
## Summary
In a real project using `bd` in Dolt **server mode** with shared server enabled, mutating commands repeatedly log:

- `auto-importing .../.beads/issues.jsonl into empty database...`
- `auto-imported ... issues ...`

This happens on ordinary updates, even after re-initializing the repo and restoring from backup.

## Environment
- `bd version 1.0.4 (ce242a879: HEAD@ce242a879678)`
- OS: Linux
- Repo-local metadata:
  - `.beads/metadata.json`: `backend=dolt`, `dolt_mode=server`, `dolt_database=PortfolioWizard`
  - `.beads/config.yaml`: `dolt.shared-server: true`
- `bd dolt status`:
  - running shared server
  - port `3308`
  - data dir `/home/chris/.beads/shared-server/dolt`

## Repro
1. In repo configured for server mode + shared server, run:
   - `bd update <issue-id> --status open`
2. Observe output includes auto-import from `.beads/issues.jsonl`.
3. Repeat same command; auto-import happens again.

### Concrete command/log capture
Command:
```bash
BD_DEBUG=1 BD_DEBUG_SYNC=1 BD_DEBUG_RPC=1 bd -v update PortfolioWizard-d70.1 --status open 2> /tmp/opencode/bd-debug.log
```

Captured stderr (`/tmp/opencode/bd-debug.log`):
```text
Debug: loaded config from /home/chris/code/PortfolioWizard/.beads/config.yaml
auto-importing 3364083 bytes from /home/chris/code/PortfolioWizard/.beads/issues.jsonl into empty database...
auto-imported 1709 issues from /home/chris/code/PortfolioWizard/.beads/issues.jsonl
backup: throttled (last backup 6m43s ago, interval 15m0s)
auto-export: wrote 1711 issues and 0 memories to /home/chris/code/PortfolioWizard/.beads/issues.jsonl
Warning: auto-export: git add failed: exit status 1: The following paths are ignored by one of your .gitignore files:
.beads/issues.jsonl
```

## What I already tried
- Exported full issues backup (`bd export`)
- Archived entire `.beads/`
- Re-initialized repo in server mode (`bd init --server`)
- Re-imported all issues (`bd import <backup.jsonl>`)

Result: behavior persists (auto-import still triggers on mutating commands).

## Expected
[...]

Issue #3760, opened 2026-05-06
Proposal: bd serve daemon to amortize dolt connection setup


DarranShepherd:
# [Proposal] Opt-in `bd serve` daemon to amortize per-invocation dolt connection setup

**Target repo**: `gastownhall/beads` (bd-side daemon code) — with cross-reference to `gastownhall/gascity` (gc-side wiring)
**Type**: Feature request / design discussion. Asking if the approach is upstream-acceptable BEFORE drafting a PR.

---

## Summary

In a small Gas City topology with only 4 always-on agents (mayor, deacon, boot, one rig's witness) and no active polecat work, the managed Dolt SQL server pegs at **400-550% CPU** at steady state. The dominant cost is **per-connection setup work in the SQL server** (view resolution and similar) paid for **every** `bd` subprocess invocation, multiplied by the rate at which gc tooling and patrol agents shell out.

This issue proposes adding an opt-in `bd serve` daemon mode: a long-running `bd` process listening on a Unix socket, with `bd` invocations forwarding via `BEADS_DAEMON_SOCKET` env var when set, falling through to direct connection on stale-socket. The daemon maintains a per-workspace store registry so it can serve cross-rig requests from one process.

A working prototype exists locally; it reduced connection rate from ~50 conn/sec → ~0.4 conn/sec and Dolt CPU from 400-600% → ~50% on the same topology. Posting before drafting the PR to confirm the upstream maintainers want the daemon shaped this way (or at all).

---

### Hardware / versions
- Apple M5, 10 cores (4 perf + 6 eff)
- `dolt version`: 1.86.6
- `bd version`: 1.0.3 (bbd4e4d3c125)
- `gc version`: 1.0.1

### Topology
Minimal Gas City with 4 always-running agents:
- `gastown.boot`
- `gastown.deacon`
- `gastown.mayor`
- `ampernomics/gastown.witness`

Polecats / refinery / dog pool scaled to 0. **No active development work.** Just the supervisor, patrols, and idle-but-running scaffolding.

Bead workload:
- HQ database: ~600 beads
- Active rig database: ~13,700 beads
- Two suspended forks: 0 beads each

### Measurements

```
$ # Connection rate over 15s
$ START=$(... SELECT MAX(id) FROM information_schema.processlist ...)
$ sleep 15
$ END=$(... SELECT MAX(id) FROM information_schema.processlist ...)
$ echo "$(($END - $START)) connections in 15s"
[...]

PR #4165, opened 2026-05-26
perf(dolt): reduce remote-Dolt round trips on read commands (showcase — do not merge)

Shockwave2k:
## What

Cuts wall-clock cost of bd against an external Dolt SQL server. On a connection with ~200 ms RTT to a remote host, the changes drop:

- `bd list` over 200 issues: **12.3 s → 2.6 s** (4.7×)
- `bd create` (warm): **>161 s → ~9 s** (>17×; stock v1.0.4 routinely times out)
- Other read commands: **2× to 5×** faster, no regressions

Pre-existing v1.0.4 bugs surfaced and fixed in passing:

- `maybeAutoImportJSONL` re-imported the entire JSONL on every server-mode write because the fallback path has no atomic emptiness check — GH#3880, GH#4128
- `schema.MigrateUp` + the 17 idempotent `RunCompatMigrations` + `EnsureIgnoredTables` ran on every write, paying ~40 SQL probes (~8 s on this link) just to confirm "nothing to do"
- `GetCommentCountsInTx` partitioned its input IDs by wisp status with a per-ID `IsActiveWispInTx` loop — 200 round trips for a 200-issue corpus, same family as GH#3239

## Why

Running `bd 1.0.4` with `dolt_mode=server` against a remote Dolt server (not localhost) made every command take 5–30 s wall-clock despite the server itself answering a single `SELECT 1` in ~200 ms. CPU on the client was ~1 %; the rest was serial network waits. Writes were worse — `bd create` regularly hit the timeout. Same symptom as #4102, adjacent to #3239 and #3760. For teams using a shared Dolt server, this is the blocker for actually adopting beads in dolt_mode=server.

## How (nine focused commits on top of v1.0.4)

A per-call SQL trace (`BD_DOLT_TRACE_SQL=1`, see below) drove the work commit by commit:

1. **DSN + connection layer.** Pin `MaxAllowedPacket` (the Go MySQL driver otherwise issues a `SELECT @@max_allowed_packet` discovery query after every fresh connection) and enable `InterpolateParams` (skip the `COM_STMT_PREPARE` round trip on parameterized reads). Skip the dual-pool `SHOW DATABASES` probe when `CreateIfMissing=false`. Drop the redundant post-open `PingContext` and the no-op `net.DialTimeout` preflight for non-local TCP.
2. **`withReadConn` helper.** Pure-read methods (`GetConfig`, `GetAllConfig`, `GetMetadata`, `GetLocalMetadata`, and `SearchIssues` when the wisps table is known empty) borrow a pooled `*sql.Conn` instead of opening a `BEGIN/ROLLBACK` transaction. Saves 2 round trips per simple read.
3. **`preloadSessionCaches`.** At store-open, one multi-statement query — `SELECT key,value FROM config; SELECT 1 FROM wisps LIMIT 1` — warms five per-store caches (config map, custom statuses / types / infra types, wisps-existence). Subsequent `GetCustomStatusesDetailed`, `GetCustomTypes`, `GetInfraTypes`, `GetAllConfig`, plus the wisps-merge step in `SearchIssues`, all become cache hits.
4. **`SearchIssues` on a pooled conn.** When the wisps cache reports empty, `SearchIssues` runs through `withReadConn` instead of `withReadTx`, dropping `BEGIN/ROLLBACK` for the main query. Four `issueops` helpers were widened from `*sql.Tx` to a small `SQLQuerier` interface; `*sql.Tx` still satisfies it, so existing callers compile unchanged.
5. **Read-only PostRun gating.** `maybeAutoExport` / `maybeAutoBackup` / `maybeAutoPush` now skip on read-only commands. `export.auto` defaults to `true`, so before this every `bd list` was running a second `SearchIssues` plus a `GetCurrentCommit` in PostRun.
6. **Routing batching.** The routing module's six sequential `GetConfig` lookups collapse into one `GetAllConfig`. Result is cached on the `DoltStore` for the rest of the command.
7. **Server-mode write fixes** (round 5):
   - `maybeAutoImportJSONL` now probes `SELECT 1 FROM issues LIMIT 1` before the server-mode fallback import and skips when the database is clearly not empty. Mirrors the spirit of the embedded path's in-transaction emptiness check. Resolves GH#3880, GH#4128.
   - `initSchemaOnDB` stamps `bd_schema_init_done_at_version` in the dolt-ignored `local_metadata` table after a successful run. Subsequent opens check the stamp (one SELECT) and short-circuit the migration chain when it's already ≥ `schema.LatestVersion()`. An older bd binary stamps a lower value, so a newer binary still re-runs its newer migrations correctly.
8. **`GetCommentCountsInTx` batched partition.** Replaces the per-ID `IsActiveWispInTx` loop with the existing `PartitionWispIDsInTx` helper (which the sibling `GetCommentsForIssuesInTx` already uses). One batched `SELECT id FROM wisps WHERE id IN (...)` instead of N round trips.
9. **Parallelize `maybeAutoExport` hydration.** The five independent bulk-load reads inside `exportToFile` (labels, deps, comments, commentCounts, depCounts) run via `sync.WaitGroup` on the existing connection pool instead of back-to-back.
10. **`EnsureIgnoredTables` fast path.** Gate behind the same schema-init stamp. With the stamp set, skip the 7 sequential `SHOW TABLES LIKE` probes (~1.4 s on remote).
11. **`GetConfig(key)` honours `allConfigCache`.** Per-key reads after preload hit the in-memory cache instead of round-tripping.
12. **Auto-export throttle: 5 min in server mode.** Embedded backend stays on the 60-second default (rebuild is ~150 ms there). Server mode rebuilds in 5-10 s, so a longer default avoids paying that cost on every burst write. Users can override via `export.interval` in `.beads/config.yaml`.

A new optional `BD_DOLT_TRACE_SQL=1` env var emits per-call timings to stderr so anyone can verify these numbers themselves; `BD_DOLT_TRACE_STACK=1` adds full call stacks.

## ⚠️ Showcase only — please do not merge

A few pieces here are deliberately additive (new code next to old) so each change is reviewable in isolation against `v1.0.4`:

- `SearchIssuesNonWispInTx` is a near-duplicate of `SearchIssuesInTx`. Proper version: one function with a `skipWispsMerge` hint.
- `withReadConn` is a sibling of `withReadTx`. Proper version: one helper that picks tx vs. conn from caller intent.
- `internal/storage/dolt/sqltrace.go` parallels the existing OTel `doltTracer`. Proper version: fold the few extra spans in and drop `sqltrace.go`.

The intent: land *this* PR to validate the numbers and approach with maintainers, then open a follow-up PR that collapses the duplicates and removes `sqltrace.go`. Marked as **draft**.

### Output parity (200-issue seed)

`bd list`, `bd list --all`, `bd list --json`, `bd list --type bug`, `bd list --status in_progress`, `bd list --label …`, `bd list --flat`, `bd ready`, `bd count`, `bd show <id>`, `bd search …` all produce byte-for-byte identical output between stock `bd v1.0.4` and this branch on both backends (use `bd export | bd import` to reproduce the seed on both).

### Wall-clock measurements

A 200-issue corpus (10 epics, 40 features, 110 tasks, 25 bugs, 15 chores; mixed statuses; parent-child hierarchy across 3 levels; `blocks` dependencies on every 5th task; ~10% closed) seeded identically into:

- embedded Dolt (local disk)
- external Dolt server (~200 ms RTT TCP)

3 timed iterations after a warmup, timed with `time.monotonic()` from a Python wrapper.

#### Embedded backend — no regressions, modest wins from no-op gating

| command | stock (ms) | this PR (ms) | speedup |
|---|---|---|---|
| `bd list` | 655 | 434 | 1.51× |
| `bd list --all` | 614 | 447 | 1.37× |
| `bd list --json` | 742 | 540 | 1.37× |
| `bd list --type bug` | 696 | 478 | 1.46× |
| `bd list --status in_progress` | 657 | 459 | 1.43× |
| `bd list --label …` | 655 | 452 | 1.45× |
| `bd list --flat` | 725 | 519 | 1.40× |
| `bd ready` | 861 | 621 | 1.39× |
| `bd stats` | 371 | 374 | 0.99× (within noise) |
| `bd count` | 314 | 242 | 1.30× |
| `bd show <id>` | 491 | 420 | 1.17× |
| `bd search` | 344 | 290 | 1.19× |
| `bd create` (warm) | 740 | 730 | 1.01× (no regression) |
| `bd update` (warm) | 528 | 574 | 0.92× (within noise) |
| `bd close` (warm) | 594 | 589 | 1.01× (no regression) |

#### Remote-Dolt backend — major wins on every command tested

Read commands (3 timed iterations, drop the first as warmup, against the seeded server repo):

| command | stock (ms) | this PR (ms) | speedup |
|---|---|---|---|
| `bd list` | 12,281 | **2,593** | **4.74×** |
| `bd list --all` | 11,945 | **2,731** | **4.37×** |
| `bd list --json` | 36,858 | **14,935** | **2.47×** |
| `bd list --type bug` | 13,755 | **2,527** | **5.44×** |
| `bd list --status in_progress` | 12,523 | **2,610** | **4.80×** |
| `bd list --label …` | 12,664 | **2,486** | **5.09×** |
| `bd list --flat` | 10,470 | **4,020** | **2.60×** |
| `bd ready` | 30,425 | **15,630** | **1.95×** |
| `bd count` | 5,531 | **2,396** | **2.31×** |
| `bd show <id>` | 12,236 | **5,726** | **2.14×** |
| `bd search` | 6,989 | **2,768** | **2.52×** |

Write commands (single-iteration sampling; stock was killed at the 3-minute mark because it was clearly going to run for many minutes):

| command | stock (ms) | this PR (ms) | speedup |
|---|---|---|---|
| `bd create` cold (first call, schema-init stamp absent) | **>161,565** (killed at 2:42, still running) | **~18,800** | **>8×** |
| `bd create` warm (avg over 4 subsequent calls) | n/a — never completed | **~9,100** | **>17× lower bound** |

Per-iteration values for the warm sample: 11,348 / 11,722 / 11,880 / 11,951 ms.

`bd stats` is **not in the table** because it hangs against this server on both stock bd and this branch — the SQL path that backs it executes one query per metric serially and the server's response to one of them never completes within 30 s. Reproducible on stock v1.0.4 and not introduced here.

### Unit tests

Existing test suites in the touched packages all pass (`-short`):

- `internal/storage/doltutil` — DSN tests cover the new `MaxAllowedPacket` / `InterpolateParams` defaults indirectly via `mysql.ParseDSN` round-trip checks.
- `internal/storage/issueops` — including `TestGetCommentCountsInTx*`, exercises the new batched-partition path.
- `internal/storage/dolt` — schema-init, connection-pool, store-unit tests. Note: the pre-existing `TestPullWithAutoResolve_BranchTrackingFallbackSuccess` requires the `dolt` CLI on PATH and was failing on `main` before this PR — unchanged by this work.
- `internal/storage/dolt/migrations` — compat migrations still pass; the new `bd_schema_init_done_at_version` stamp is best-effort and doesn't change migration semantics.

### Reproducing the numbers

The build artifacts include an opt-in SQL trace gated by env vars (off by default), so any reviewer can run `BD_DOLT_TRACE_SQL=1 bd list` and see every round trip annotated with timing and caller info.

### How to try it locally

Build the patched binary alongside your installed `bd`:

```bash
git clone https://github.com/gastownhall/beads
cd beads
git fetch origin pull/<this PR number>/head:perf/remote-dolt-rtt-104-pr
git checkout perf/remote-dolt-rtt-104-pr

# Embedded mode (full CGO build, lets you run against .beads/embeddeddolt/)
make build
mv bd ~/.local/bin/bd-perf

# Or server-mode-only (no CGO, faster build, no embedded backend)
CGO_ENABLED=0 go build -tags gms_pure_go -o ~/.local/bin/bd-perf ./cmd/bd
```

Then point an existing repo at it without disturbing your installed `bd`:

```bash
# Read commands (always safe — no mutations)
bd-perf -C /path/to/.beads/repo list
bd-perf -C /path/to/.beads/repo list --json
bd-perf -C /path/to/.beads/repo show <issue-id>

# Side-by-side compare against stock 1.0.4
time bd list           # baseline
time bd-perf list      # this branch

# See the trace yourself
BD_DOLT_TRACE_SQL=1 bd-perf list 2>&1 | head -40
BD_DOLT_TRACE_STACK=1 bd-perf list   # adds full call stacks per SQL op
```

For server mode, `bd-perf` reads the same `.beads/metadata.json`, `.beads/dolt-server.port`, and `~/.config/beads/credentials` as stock `bd` — no extra configuration required. Output is byte-for-byte identical to stock for the commands in the parity table above, so it's safe to drop into existing workflows for read commands. **Writes work and are dramatically faster, but the patches are framed as a showcase — please don't merge or run them in production without a follow-up that absorbs the duplicates listed in the showcase note above.**

To revert to stock `bd`, just delete `~/.local/bin/bd-perf` — your `~/.local/bin/bd` is untouched.

## Related issues

- **#4102** — `perf(server-mode): bd commands take 5-10s in remote dolt_mode due to cold connection + serial query pattern` — direct response.
- **#3239** — `Batch wisp routing checks to reduce WAN latency amplification in Dolt server mode` — addressed via the wisps-existence cache + batched routing config + the `GetCommentCountsInTx` fix.
- **#3760** — `Proposal: bd serve daemon to amortize dolt connection setup` — complementary; this PR demonstrates what's achievable *without* a daemon.
- **#3501** + **#3522** — `auto-backup` runs on every command — partially mitigated by the read-only gating in this PR.
- **#3880** + **#4128** — server-mode write-path re-imports JSONL / re-runs schema init on every write — fixed in round 5.

## Authorship

Mixed human + AI effort. Investigation, code, tests, and benchmarks were a back-and-forth between **Hans Boese** (human, account opening this PR) and **Claude** (Anthropic, via Claude Code, working as a paired contributor). Every commit has `Co-Authored-By: Claude <noreply@anthropic.com>`. The roles split roughly as: Hans set the goal and reviewed each step; Claude wrote the trace harness, ran the measurements, drafted the patches and the benchmark report, iterating in tight loops with Hans deciding what to keep at each round.

## Follow-up (not in this PR)

- Collapse `SearchIssuesNonWispInTx` into `SearchIssuesInTx` via a hint param.
- Replace `withReadTx` + `withReadConn` pair with a single helper that picks tx vs. conn from caller intent.
- Remove `internal/storage/dolt/sqltrace.go`; fold its handful of spans into the existing OTel `doltTracer`.
- Combine `CreateIssueInTx` + `doltAddAndCommit` into one round trip via a stored procedure or multi-statement (currently two transactions per write = ~4-5 s of the remaining ~9 s create cost).
- Audit other `IsActiveWispInTx`-in-loop call sites for the same N+1 pattern (`dependency_queries.go:406` is a known candidate, only triggered on `bd close`).
- Consider `bd serve` (per #3760) for sub-1 s reads via amortized connection setup.

Each of those is a separate PR-sized change.

### DIFF
--- a/cmd/bd/auto_import_upgrade.go
+++ b/cmd/bd/auto_import_upgrade.go
@@ -10,6 +10,23 @@ import (
 	"github.com/steveyegge/beads/internal/types"
 )
 
+// storeAlreadyHasIssues reports whether the given store's `issues` table
+// already contains at least one row. Used by maybeAutoImportJSONL to skip
+// the slow server-mode fallback import when the database is clearly not
+// empty (a re-import on every write is a v1.0.4 server-mode bug — see
+// GH#3880 and GH#4128).
+//
+// Best-effort: any error (table missing, transient network glitch, etc.)
+// is treated as "unknown emptiness", which preserves the pre-existing
+// behaviour of attempting the import.
+func storeAlreadyHasIssues(ctx context.Context, s storage.DoltStorage) bool {
+	issues, err := s.SearchIssues(ctx, "", types.IssueFilter{Limit: 1})
+	if err != nil {
+		return false
+	}
+	return len(issues) > 0
+}
+
 // jsonlImporter is implemented by stores that support single-transaction
 // JSONL import (currently EmbeddedDoltStore). Stores that don't implement
 // this fall back to the multi-call path.
@@ -71,6 +88,17 @@ func maybeAutoImportJSONL(ctx context.Context, s storage.DoltStorage, beadsDir s
 	}
 
 	// Fallback for non-embedded stores: multi-call path (original behavior).
+	// In server mode this path has no atomic emptiness check, so on v1.0.4
+	// every write triggered a full re-import of issues.jsonl (~150 s for a
+	// 200-issue corpus over a ~200 ms RTT link). Guard with a cheap probe
+	// of the issues table — if the database already has data, we have
+	// nothing to import. Matches the spirit of the embedded path's
+	// in-transaction emptiness check (see ImportJSONLData). See GH#3880
+	// and GH#4128.
+	if storeAlreadyHasIssues(ctx, s) {
+		return
+	}
+
 	fmt.Fprintf(os.Stderr, "auto-importing %d bytes from %s into empty database...\n", info.Size(), jsonlPath)
 
 	result, err := importFromLocalJSONLFull(ctx, s, jsonlPath)
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -59,7 +59,19 @@ func maybeAutoExport(ctx context.Context) {
 	state := loadExportAutoState(beadsDir)
 	interval := config.GetDuration("export.interval")
 	if interval == 0 {
-		interval = 60 * time.Second
+		// Default throttle: 60 s for the embedded backend, 5 min for the
+		// remote-Dolt backend. The remote case rebuilds the JSONL with a
+		// full SearchIssues + 5 hydration queries every fire, which costs
+		// 5-10 s on a 200 ms RTT link. With dolt_mode=server the database
+		// is already the source of truth and the JSONL is a convenience
+		// for git-tracking and offline replay — a 5-minute lag is fine
+		// for that role and skips most per-write rebuild churn.
+		// Users can override by setting export.interval explicitly.
+		if serverMode {
+			interval = 5 * time.Minute
+		} else {
+			interval = 60 * time.Second
+		}
 	}
 	if !state.Timestamp.IsZero() && time.Since(state.Timestamp) < interval {
 		debug.Logf("auto-export: throttled (last export %s ago, interval %s)\n",
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/steveyegge/beads/internal/atomicfile"
@@ -177,11 +178,27 @@ func exportToFile(ctx context.Context, path string, includeMemories bool) (issue
 		for i, issue := range issues {
 			issueIDs[i] = issue.ID
 		}
-		labelsMap, _ := store.GetLabelsForIssues(ctx, issueIDs)
-		allDeps, _ := store.GetDependencyRecordsForIssues(ctx, issueIDs)
-		commentsMap, _ := store.GetCommentsForIssues(ctx, issueIDs)
-		commentCounts, _ := store.GetCommentCounts(ctx, issueIDs)
-		depCounts, _ := store.GetDependencyCounts(ctx, issueIDs)
+		// Parallelize the five independent bulk-load calls. On a remote
+		// Dolt connection each is a full BEGIN/SELECT/ROLLBACK round trip
+		// (~1.2 s on a ~200 ms RTT link); running them sequentially was
+		// ~6 s of pure wall-clock waste inside maybeAutoExport. The pool's
+		// MaxOpenConns is 10 by default so five concurrent reads share
+		// the warm pool without forcing extra handshakes.
+		var (
+			labelsMap     map[string][]string
+			allDeps       map[string][]*types.Dependency
+			commentsMap   map[string][]*types.Comment
+			commentCounts map[string]int
+			depCounts     map[string]*types.DependencyCounts
+			wg            sync.WaitGroup
+		)
+		wg.Add(5)
+		go func() { defer wg.Done(); labelsMap, _ = store.GetLabelsForIssues(ctx, issueIDs) }()
+		go func() { defer wg.Done(); allDeps, _ = store.GetDependencyRecordsForIssues(ctx, issueIDs) }()
+		go func() { defer wg.Done(); commentsMap, _ = store.GetCommentsForIssues(ctx, issueIDs) }()
+		go func() { defer wg.Done(); commentCounts, _ = store.GetCommentCounts(ctx, issueIDs) }()
+		go func() { defer wg.Done(); depCounts, _ = store.GetDependencyCounts(ctx, issueIDs) }()
+		wg.Wait()
 
 		for _, issue := range issues {
 			issue.Labels = labelsMap[issue.ID]
--- a/cmd/bd/list.go
+++ b/cmd/bd/list.go
@@ -947,9 +947,14 @@ var listCmd = &cobra.Command{
 			}
 
 			// Regular tree display (no parent filter)
-			// Load dependencies for tree structure
-			// Best effort: display gracefully degrades with empty data
-			allDeps, _ := activeStore.GetAllDependencyRecords(ctx)
+			// Load dependencies for tree structure (skipped when there are no
+			// issues — the dependency query would otherwise burn a full read
+			// transaction against the remote server for an empty result set).
+			// Best effort: display gracefully degrades with empty data.
+			var allDeps map[string][]*types.Dependency
+			if len(issues) > 0 {
+				allDeps, _ = activeStore.GetAllDependencyRecords(ctx)
+			}
 			displayPrettyListWithDeps(issues, false, allDeps)
 			printTruncationHint(truncated, effectiveLimit)
 			return
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -1109,16 +1109,19 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
-		// Auto-backup: export JSONL to .beads/backup/ if enabled and due
-		maybeAutoBackup(rootCtx)
-
-		// Auto-export: write git-tracked JSONL for portability if enabled and due
-		maybeAutoExport(rootCtx)
-
-		// Auto-push: push to Dolt remote if enabled and due.
-		// Skip for read-only commands to avoid unnecessary network operations
-		// and metadata writes on commands like bd list/show/ready (GH#2191).
+		// Auto-backup, auto-export, auto-push: skip for read-only commands.
+		// These post-run maintenance tasks issue extra round trips (GetCurrentCommit,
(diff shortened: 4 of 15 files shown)


Shockwave2k:
Closing this in favor of a fresh PR rebased onto current `main`.

When I opened this, I built against the v1.0.4 release tag because that's what users on Homebrew actually run. After opening I noticed `main` has independently landed two of the ideas in this PR with different mechanisms:

- Read-only PostRun gating on `maybeAutoExport` / `maybeAutoBackup` / `maybeAutoPush` — landed on `main` via `shouldRunPostCommandAutoExport(cmd)`. Same intent as our round-2 gating; **closing-comment ours is a duplicate** of yours.
- `maybeAutoImportJSONL` emptiness check — landed on `main` via a top-level `GetStatistics(ctx)` guard. Note from my benchmarking: against a remote Dolt server with ~200 ms RTT, `bd stats` (and therefore `GetStatistics`) hangs on a single metric query that never returns within 30 s — so the guard may not actually short-circuit in `dolt_mode=server`. Worth checking. A cheaper probe like `SELECT 1 FROM issues LIMIT 1` (what this PR used) doesn't hit that hang.

The 17 conflicts vs current `main` are mostly the two-fixes overlap above. Rather than force-push a rewritten branch here, I'm opening a fresh PR against `main` that drops the duplicated work and keeps only the changes `main` doesn't already have:

- `withReadConn` for simple metadata reads
- DSN tweaks (`MaxAllowedPacket`, `InterpolateParams`) + single-pool open
- `preloadSessionCaches` + wisps-merge cache + `SearchIssues` on a pooled conn
- routing config batching
- `local_metadata` schema-init stamp (skip `EnsureIgnoredTables` + `RunCompatMigrations` on warm opens)
- `GetCommentCountsInTx` batched-partition (N+1 fix, GH#3239 family)
- `maybeAutoExport` parallel hydration + 5-min default throttle for server mode

Benchmark numbers from this PR remain valid as evidence of the underlying RTT problem on a v1.0.4 install — see the Notion writeup linked in the body for the full data.

Shockwave2k:
Superseded by #4167 — same fixes rebased onto current `main` (most importantly: drops the parts `main` had already absorbed since v1.0.4, keeps only the still-novel work in four cleanly separable commits).

Issue #3239, opened 2026-04-13
Batch wisp routing checks to reduce WAN latency amplification in Dolt server mode


fengning-starsend:
## Summary

Remote Dolt server mode appears to hit WAN latency amplification for issue hydration commands such as `bd show --json` and `bd list --json --limit N`. A narrow, code-grounded contributor is that the `issueops` layer performs per-ID active-wisp routing checks inside transactions.

This is not a general complaint about Dolt server mode. Simple connectivity commands are fast. The slowdown appears when Beads hydrates issues, labels, dependencies, comments, and related records.

## Benchmarks

Environment:

- `bd version 1.0.0 (dev: 9fa62ee3979e)` on both hosts
- Dolt SQL server on Linux host `epyc12`
- Client on macOS M-series laptop
- Tailscale network, roughly 100-200ms RTT during testing
- Remote server mode via `BEADS_DOLT_SERVER_HOST=100.107.173.83`, `BEADS_DOLT_SERVER_PORT=3307`, `BEADS_DOLT_SERVER_MODE=1`
- Runtime path: `BEADS_DIR=~/.beads-runtime/.beads`

Fresh timings from the macOS client:

| Command | Remote Dolt server mode | SSH to Dolt host and run `bd` co-located |
| --- | ---: | ---: |
| `bd dolt test --json` | 0.27s | 1.36s including SSH overhead |
| `bd show bd-q6o1p --json` | 7.62s | 1.36s including SSH overhead |
| `bd list --json --limit 5` | 15.25s | 1.29s including SSH overhead |

The important signal is that `dolt test` is fast over the remote SQL path, while hydration-heavy commands are much slower. The SSH path is not inherently faster transport; it is faster here because `bd` runs on the same host as the Dolt server and avoids repeated WAN roundtrips.

## Code evidence

The transaction-scoped wisp routing helper checks one ID at a time:

```go
// internal/storage/issueops/wisp_routing.go
func IsActiveWispInTx(ctx context.Context, tx *sql.Tx, id string) bool {
    var exists int
    err := tx.QueryRowContext(ctx, "SELECT 1 FROM wisps WHERE id = ? LIMIT 1", id).Scan(&exists)
    return err == nil
}
```

Several bulk paths call this repeatedly inside a transaction, for example:

- `internal/storage/issueops/labels.go` in label hydration for issue IDs
- `internal/storage/issueops/comments.go` in comment count hydration
[...]

Issue #3523, opened 2026-04-26
auto-backup should skip file:// auto-register when Dolt server's filesystem view differs from client's (container/remote)


shaunc:
## Summary

bd's auto-backup hardcodes a `file://` URL pointing at a path on the **client's** filesystem. When the Dolt server runs in a container or on a different host (tailnet/SSH-tunneled), the server can't see the client's filesystem at all. The `add backup backup_export file://<client-path>` always fails server-side; auto-backup is structurally broken in this topology.

`file://` Dolt remotes are useful when client and server share the filesystem (binary local mode). For containerized/remote servers, an S3 / azure / gcs remote is required, or auto-backup should be skipped entirely.

## Reproducer

Operator setup:
- `BEADS_DOLT_SERVER_HOST=100.64.0.1` (or any non-`127.0.0.1` host)
- bd workspace has `dolt_mode=server`
- Dolt server is in a container with bind-mounts that don't include the client's pwd

Every bd command fires the failing register attempt (compounded by the no-backoff issue in the sibling filing).

## Expected

Detect that the server's filesystem doesn't include the client's path and skip the `file://` auto-register. Detection options:

1. **Heuristic**: if `BEADS_DOLT_SERVER_HOST` (or the resolved server host) is non-loopback (`!= 127.0.0.1` AND `!= localhost` AND `!= ::1`), the server is presumed remote. Skip file:// auto-register; emit a one-time INFO ("server is remote — skipping auto file:// backup; configure an S3/gcs/azure remote manually").
2. **Probe**: at first connect, query Dolt's `dolt_status` or similar for hostname/working-directory; compare to client's. If different, skip.

(1) is cheaper and covers the common case. (2) is more rigorous.

Either way, an explicit config opt-in for non-`file://` remotes (S3 etc.) lets operators with cross-filesystem setups still get backup coverage if they want it.

## Environment

- bd version 1.0.2 (a3f834b3)
- Setup: containerized Dolt server (e.g., `dolthub/dolt-sql-server:1.82.4` in docker)
- `BEADS_DOLT_SERVER_HOST` set to a non-loopback address
- Client's pwd not bind-mounted into the container

## Related

- #3501 — `auto-register backup_export` ignores `backup.enabled=false` (parent).
- (sibling) — symlink-naive URL.
- (sibling) — no backoff on retry.

PR #4167, opened 2026-05-26
perf(dolt): reduce remote-Dolt round trips (showcase — do not merge)

Shockwave2k:
## What

Cuts wall-clock cost of `bd` against an external Dolt SQL server (`dolt_mode=server`). On a connection with ~200 ms RTT to a remote host, the changes drop:

- `bd list` over 200 issues: **~12 s → ~3 s**
- `bd create` warm: **>2 min on stock v1.0.4 → ~8 s** on this branch

Four focused commits on top of current `main`. Each is independently reviewable.

## Why

I (Hans) was running `bd 1.0.4` in `dolt_mode=server` against a remote Dolt server (not localhost). Every command took 5–30 s wall-clock despite the server itself answering a single `SELECT 1` in ~200 ms. CPU on the client was ~1 %; the rest was serial network waits. Writes were worse — `bd create` regularly timed out. Same symptom as #4102, adjacent to #3239 and #3760. For teams using a shared Dolt server, this gap is the blocker for actually adopting beads in `dolt_mode=server`.

A separate, parallel investigation against **v1.0.4** built up a 9-commit showcase (closed-but-linked PR #4165, with a full Notion writeup). After tracing the same paths against current `main`, two ideas from the v1.0.4 work had already landed here with different mechanisms:

- Read-only PostRun gating on `maybeAutoExport` / `maybeAutoPush` (already on `main` via `shouldRunPostCommandAutoExport(cmd)`).
- `maybeAutoImportJSONL` emptiness check (already on `main` via a top-level `GetStatistics(ctx)` guard). **Heads-up from the v1.0.4 investigation**: against a remote Dolt server `bd stats` (and therefore `GetStatistics`) hangs on a single metric query that never returns within 30 s. The emptiness guard may not actually short-circuit in `dolt_mode=server`. A cheaper probe (`SELECT 1 FROM issues LIMIT 1`) doesn't hit that hang. Worth a separate look.
- `GetCommentCountsInTx` batched wisp partition (already on `main`).
- `routing_read.go` batched `GetAllConfig` (already on `main`).
- `RunCompatMigrations` + `EnsureIgnoredTables` chain (already removed on `main` — schema is consolidated under the embedded migration system).

This PR keeps only the changes that `main` does **not** already have.

## How (four commits)

A per-call SQL trace (`BD_DOLT_TRACE_SQL=1`, see below) drove the work:

1. **`perf(dolt): cut remote-Dolt round trips on metadata reads + connection open`**
   - New `withReadConn` helper: `GetConfig`, `GetAllConfig`, `GetMetadata`, `GetLocalMetadata` borrow a pooled `*sql.Conn` instead of opening a `BEGIN/ROLLBACK` transaction. Saves 2 round trips per simple read.
   - Widen four `issueops` `Get*InTx` helpers from `*sql.Tx` to a small `SQLQuerier` interface so callers can pass either tx or conn. `*sql.Tx` still satisfies it, so existing callers compile unchanged.
   - DSN: pin `MaxAllowedPacket = 64 MiB` so the driver skips its `SELECT @@max_allowed_packet` discovery query (–1 RTT per cold conn).
   - DSN: `InterpolateParams = true` so the driver substitutes parameters client-side instead of `COM_STMT_PREPARE` + `COM_STMT_EXECUTE` (halves parameterized-read cost: 2 RTT → 1 RTT).
   - `openServerConnection` fast path: when `CreateIfMissing=false` (every read-only command), skip the second init-only pool and the `SHOW DATABASES` probe. Just ping the main pool and translate MySQL error 1049 (Unknown database) to `databaseNotFoundError`. Saves a handshake + a probe per command.

2. **`perf(dolt): trace harness + gate maybeAutoBackup + skip empty-list deps + skip preflight dial`**
   - `cmd/bd/main.go`: gate `maybeAutoBackup` behind `!isReadOnlyCommand`. `main` already gates `maybeAutoExport` and `maybeAutoPush`; this extends the same intent to backup, which still runs on every command and pays a `GetCurrentCommit` round trip.
   - `cmd/bd/list.go`: skip `GetAllDependencyRecords` when the issue list is empty. Default tree-mode display would otherwise pay 3 RTT for an empty result set on a remote link.
   - `internal/storage/dolt/sqltrace.go`: new env-gated per-call SQL trace (`BD_DOLT_TRACE_SQL=1`, `BD_DOLT_TRACE_STACK=1`). Off by default; lets reviewers verify per-RTT timing themselves. This is the harness that drove every other change.
   - `internal/storage/dolt/store.go`: skip the `net.DialTimeout` preflight for non-local TCP servers. The MySQL driver's own `Timeout` already fast-fails; preflight only matters when `AutoStart` can kick in on a local server.

3. **`perf(dolt): store-open prefetch + wisps-merge skip`**
   - `preloadSessionCaches` fires a 2-statement multi-query (`SELECT key,value FROM config; SELECT 1 FROM wisps LIMIT 1`) at store-open and populates five per-store caches (config map, custom statuses / types / infra types, wisps existence). One round trip up front in exchange for eliminating 3–5 round trips spread across the rest of the command.
   - `GetConfig(key)` and `GetAllConfig` honor `allConfigCache` so the routing module and other repeat callers do not pay a round trip per key once the cache is warm. `SetConfig` and `DeleteConfig` invalidate the cache.
   - Adds `SearchIssuesNonWispInTx`: sibling of `SearchIssuesInTx` that runs only the issues-table search and skips the wisps merge entirely. `DoltStore.SearchIssues` routes to it when `canSkipWispsMerge` reports the cache knows wisps is empty AND the caller is not explicitly asking for ephemeral rows. Saves the wisps-empty probe round trip that `main`'s `SearchIssuesInTx` pays on every call.

4. **`perf(auto-export): parallelize bulk-load reads, default to 5-min throttle in server mode`**
   - Inside `exportToFile`, the five independent bulk-load reads (labels, deps, comments, commentCounts, depCounts) run via `sync.WaitGroup` on the existing connection pool instead of back-to-back. Each was ~1.2 s on a remote link; ~6 s sequential → ~1.5 s parallel. Pool default `MaxOpenConns=10` covers the five concurrent reads.
   - `export.interval` default is now backend-aware: 60 s for embedded (rebuild costs ~150 ms), 5 min for `dolt_mode=server` (rebuild costs 5–10 s on a remote link). Users can override by setting `export.interval` explicitly.

A new optional `BD_DOLT_TRACE_SQL=1` env var emits per-call timings to stderr so anyone can verify these numbers themselves; `BD_DOLT_TRACE_STACK=1` adds full call stacks.

## ⚠️ Showcase only — please do not merge

A few pieces here are deliberately additive (new code next to old) so each change is reviewable in isolation:

- `SearchIssuesNonWispInTx` is a near-duplicate of `SearchIssuesInTx`. Proper version: one function with a `skipWispsMerge` hint.
- `withReadConn` is a sibling of `withReadTx`. Proper version: one helper that picks tx vs. conn from caller intent.
- `internal/storage/dolt/sqltrace.go` parallels the existing OTel `doltTracer`. Proper version: fold the few extra spans in and drop `sqltrace.go`.

The intent: land *this* PR to validate the numbers and approach with maintainers, then open a follow-up PR that collapses the duplicates and removes `sqltrace.go`. Marked as **draft**.

### Output parity

`bd list`, `bd list --all`, `bd list --json`, `bd list --type bug`, `bd list --status in_progress`, `bd list --label …`, `bd list --flat`, `bd ready`, `bd count`, `bd show <id>`, `bd search …` all produce byte-for-byte identical output between stock `bd` and this branch on both backends. Use `bd export | bd import` to reproduce a seed on both.

### Wall-clock measurements

A 200-issue corpus (10 epics, 40 features, 110 tasks, 25 bugs, 15 chores; mixed statuses; parent-child hierarchy across 3 levels; `blocks` dependencies on every 5th task; ~10% closed) was seeded identically into:

- embedded Dolt (local disk)
- external Dolt server (~200 ms RTT TCP)

3 timed iterations after a warmup, timed with `time.monotonic()` from a Python wrapper.

These numbers come from the parallel v1.0.4 investigation (closed PR #4165 + the linked Notion writeup). The rebased branch in this PR was spot-checked against the same server and produced the same shape of result — `bd list` ~3 s, warm `bd create` ~8 s.

#### Verified against current `main` (`f8b940016`)

Spot-check on the same server and corpus, both binaries built with `CGO_ENABLED=0 -tags gms_pure_go` so the toolchain is identical (rules out driver / CGO compile-variance noise). 3 timed iterations after one warmup run per binary:

| binary | min | avg | max |
|---|---|---|---|
| stock `bd` from `origin/main` (`f8b940016`) | 8228 ms | **8257 ms** | 8288 ms |
| this branch (4 commits on top of `f8b940016`) | 3355 ms | **3424 ms** | 3481 ms |

**Speedup vs current `main`: 2.4×** on `bd list`. (For reference: stock v1.0.4 was ~12.3 s on the same corpus, so `main` has already absorbed ~32% of the speedup independently — most of that via its `GetCommentCountsInTx` batching, the routing-config batch, and the dropped `RunCompatMigrations` chain — and this PR adds another ~2.4× on top.)

#### Embedded backend — no regressions

Embedded mode is unchanged in shape; the no-op gating and cache changes save a small amount per command (~30–50 % on read paths, 0–10 % on writes). No command is slower outside measurement noise.

#### Remote-Dolt backend — major wins on every read command

| command | stock v1.0.4 (ms) | this branch (ms) | speedup |
|---|---|---|---|
| `bd list` | 12,281 | **~3,000** | **~4×** |
| `bd list --type bug` | 13,755 | **~3,000** | **~4.5×** |
| `bd list --status in_progress` | 12,523 | **~3,000** | **~4×** |
| `bd list --label …` | 12,664 | **~3,000** | **~4×** |
| `bd count` | 5,531 | **~2,400** | **~2.3×** |
| `bd ready` | 30,425 | **~15,000** | **~2×** |
| `bd show <id>` | 12,236 | **~5,500** | **~2.2×** |
| `bd search` | 6,989 | **~2,800** | **~2.5×** |

Write commands:

| command | stock v1.0.4 (ms) | this branch (ms) | speedup |
|---|---|---|---|
| `bd create` cold first | **>161,565** (timed out at 2:42, still running) | ~16,000 | ~10× |
| `bd create` warm avg | n/a — never completed | **~8,000** | **>20×** |

Full data + breakdown by phase: see Notion writeup linked in closed PR #4165.

### Unit tests

Existing test suites in the touched packages pass (`-short`):

- `internal/storage/doltutil`
- `internal/storage/issueops`
- `internal/storage/dolt/migrations`

Note: `internal/storage/dolt/TestPullWithAutoResolve_BranchTrackingFallbackSuccess` requires the `dolt` CLI on PATH and was failing on `main` before this PR — unchanged.

### Reproducing the numbers

`BD_DOLT_TRACE_SQL=1 bd-perf list` dumps a per-call trace to stderr — every round trip annotated with timing and caller info. `BD_DOLT_TRACE_STACK=1` adds full call stacks.

### How to try it locally

```bash
git clone https://github.com/gastownhall/beads
cd beads
git fetch origin pull/<this PR number>/head:perf-rebased
git checkout perf-rebased

# Full CGO build (lets you run against .beads/embeddeddolt/ too)
make build
mv bd ~/.local/bin/bd-perf

# Or server-mode-only (no CGO, faster build, no embedded backend)
CGO_ENABLED=0 go build -tags gms_pure_go -o ~/.local/bin/bd-perf ./cmd/bd
```

Then point an existing repo at it without disturbing your installed `bd`:

```bash
bd-perf -C /path/to/.beads/repo list
bd-perf -C /path/to/.beads/repo show <issue-id>

# Side-by-side compare against your stock bd
time bd list
time bd-perf list

# See the trace
BD_DOLT_TRACE_SQL=1 bd-perf list 2>&1 | head -40
```

`bd-perf` reads the same `.beads/metadata.json`, `.beads/dolt-server.port`, and `~/.config/beads/credentials` as stock `bd` — no extra configuration. Output is byte-for-byte identical to stock for the commands in the parity table, so it's safe to drop into existing workflows for read commands. **Writes work and are dramatically faster, but the patches are framed as a showcase — please don't merge or run them in production without a follow-up that absorbs the duplicates noted above.**

To revert, just delete `~/.local/bin/bd-perf` — your `~/.local/bin/bd` is untouched.

## Related issues

- **#4102** — `perf(server-mode): bd commands take 5-10s in remote dolt_mode due to cold connection + serial query pattern` — direct response.
- **#3239** — `Batch wisp routing checks to reduce WAN latency amplification in Dolt server mode` — addressed via the wisps-existence cache + the existing `PartitionWispIDsInTx` already landed on main.
- **#3760** — `Proposal: bd serve daemon to amortize dolt connection setup` — complementary; this PR demonstrates what's achievable *without* a daemon.
- **#3880** + **#4128** — server-mode write-path re-imports JSONL on every write — `main` has already fixed the emptiness check via `GetStatistics`. See "Why" section above for a heads-up about that guard not actually short-circuiting in `dolt_mode=server` (a `bd stats` hang issue we measured).
- Closed companion PR: **#4165** (against v1.0.4, contains the full investigation against the released version).

## Authorship

Mixed human + AI effort. Investigation, code, tests, and benchmarks were a back-and-forth between **Hans Boese** (human, account opening this PR) and **Claude** (Anthropic, via Claude Code, working as a paired contributor). Every commit has `Co-Authored-By: Claude <noreply@anthropic.com>`. The roles split roughly as: Hans set the goal and reviewed each step; Claude wrote the trace harness, ran the measurements, drafted the patches and the benchmark report, iterating in tight loops with Hans deciding what to keep at each round.

## Follow-up (not in this PR)

- Collapse `SearchIssuesNonWispInTx` into `SearchIssuesInTx` via a hint param.
- Replace `withReadTx` + `withReadConn` pair with a single helper that picks tx vs. conn from caller intent.
- Remove `internal/storage/dolt/sqltrace.go`; fold its handful of spans into the existing OTel `doltTracer`.
- Investigate why `bd stats` (and therefore the auto-import emptiness `GetStatistics` guard) hangs against a remote Dolt server.
- Combine `CreateIssueInTx` + `doltAddAndCommit` into one round trip via a stored procedure or multi-statement (currently two transactions per write).

### DIFF
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/steveyegge/beads/internal/atomicfile"
@@ -77,7 +78,19 @@ func maybeAutoExport(ctx context.Context, serverMode, allowEmptyOverwrite bool) 
 	state := loadExportAutoState(beadsDir)
 	interval := config.GetDuration("export.interval")
 	if interval == 0 {
-		interval = 60 * time.Second
+		// Default throttle: 60 s for the embedded backend, 5 min for the
+		// remote-Dolt backend. The remote case rebuilds the JSONL with a
+		// full SearchIssues + 5 hydration queries every fire, which costs
+		// 5-10 s on a 200 ms RTT link. With dolt_mode=server the database
+		// is already the source of truth and the JSONL is a convenience
+		// for git-tracking and offline replay — a 5-minute lag is fine
+		// for that role and skips most per-write rebuild churn.
+		// Users can override by setting export.interval explicitly.
+		if serverMode {
+			interval = 5 * time.Minute
+		} else {
+			interval = 60 * time.Second
+		}
 	}
 
 	// Change detection via Dolt commit hash. This is cheap, so do it before
@@ -368,11 +381,27 @@ func exportToFile(ctx context.Context, path string, includeMemories bool) (issue
 		for i, issue := range issues {
 			issueIDs[i] = issue.ID
 		}
-		labelsMap, _ := store.GetLabelsForIssues(ctx, issueIDs)
-		allDeps, _ := store.GetDependencyRecordsForIssues(ctx, issueIDs)
-		commentsMap, _ := store.GetCommentsForIssues(ctx, issueIDs)
-		commentCounts, _ := store.GetCommentCounts(ctx, issueIDs)
-		depCounts, _ := store.GetDependencyCounts(ctx, issueIDs)
+		// Parallelize the five independent bulk-load calls. On a remote Dolt
+		// connection each is a full BEGIN/SELECT/ROLLBACK round trip (~1.2 s
+		// on a 200 ms RTT link); running them sequentially was ~6 s of pure
+		// wall-clock waste inside maybeAutoExport. The pool's MaxOpenConns
+		// is 10 by default so five concurrent reads share the warm pool
+		// without forcing extra handshakes.
+		var (
+			labelsMap     map[string][]string
+			allDeps       map[string][]*types.Dependency
+			commentsMap   map[string][]*types.Comment
+			commentCounts map[string]int
+			depCounts     map[string]*types.DependencyCounts
+			wg            sync.WaitGroup
+		)
+		wg.Add(5)
+		go func() { defer wg.Done(); labelsMap, _ = store.GetLabelsForIssues(ctx, issueIDs) }()
+		go func() { defer wg.Done(); allDeps, _ = store.GetDependencyRecordsForIssues(ctx, issueIDs) }()
+		go func() { defer wg.Done(); commentsMap, _ = store.GetCommentsForIssues(ctx, issueIDs) }()
+		go func() { defer wg.Done(); commentCounts, _ = store.GetCommentCounts(ctx, issueIDs) }()
+		go func() { defer wg.Done(); depCounts, _ = store.GetDependencyCounts(ctx, issueIDs) }()
+		wg.Wait()
 
 		for _, issue := range issues {
 			issue.Labels = labelsMap[issue.ID]
--- a/cmd/bd/list.go
+++ b/cmd/bd/list.go
@@ -1021,9 +1021,14 @@ var listCmd = &cobra.Command{
 			}
 
 			// Regular tree display (no parent filter)
-			// Load dependencies for tree structure
-			// Best effort: display gracefully degrades with empty data
-			allDeps, _ := activeStore.GetAllDependencyRecords(ctx)
+			// Load dependencies for tree structure (skipped when there are no
+			// issues — the dependency query would otherwise burn a full read
+			// transaction against the remote server for an empty result set).
+			// Best effort: display gracefully degrades with empty data.
+			var allDeps map[string][]*types.Dependency
+			if len(issues) > 0 {
+				allDeps, _ = activeStore.GetAllDependencyRecords(ctx)
+			}
 			displayPrettyListWithDeps(issues, false, allDeps)
 			printTruncationHint(truncated, effectiveLimit)
 			return
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -1147,8 +1147,14 @@ var rootCmd = &cobra.Command{
 				}
 			}
 
-			// Auto-backup: sync a Dolt-native backup if enabled and due
-			maybeAutoBackup(rootCtx)
+			// Auto-backup: sync a Dolt-native backup if enabled and due.
+			// Gate behind !isReadOnlyCommand so that bd list / show / ready
+			// don't pay an extra GetCurrentCommit round trip on every call —
+			// a read can't have produced data worth backing up. The auto-export
+			// gate just below covers the same intent for its JSONL pass.
+			if !isReadOnlyCommand(cmd.Name()) {
+				maybeAutoBackup(rootCtx)
+			}
 
 			// Auto-export: write git-tracked JSONL for portability if enabled and due.
 			// Read-only commands must not perform post-run maintenance writes or emit
@@ -1174,6 +1180,18 @@ var rootCmd = &cobra.Command{
 			if store != nil {
 				_ = store.Close() // Best effort cleanup
 			}
+
+			// Auto-push: push to Dolt remote if enabled and due.
+			maybeAutoPush(rootCtx)
+		}
+
+		// Signal that store is closing (prevents background flush from accessing closed store)
+		storeMutex.Lock()
+		storeActive = false
+		storeMutex.Unlock()
+
+		if store != nil {
+			_ = store.Close() // Best effort cleanup
 		}
 
 		// End the command span and flush OTel data before process exit.
--- a/internal/storage/dolt/config.go
+++ b/internal/storage/dolt/config.go
@@ -56,9 +56,9 @@ func (s *DoltStore) SetConfig(ctx context.Context, key, value string) error {
 // GetConfig retrieves a configuration value
 func (s *DoltStore) GetConfig(ctx context.Context, key string) (string, error) {
 	var value string
-	err := s.withReadTx(ctx, func(tx *sql.Tx) error {
+	err := s.withReadConn(ctx, func(q issueops.SQLQuerier) error {
 		var err error
-		value, err = issueops.GetConfigInTx(ctx, tx, key)
+		value, err = issueops.GetConfigInTx(ctx, q, key)
 		return err
 	})
 	return value, err
@@ -67,9 +67,9 @@ func (s *DoltStore) GetConfig(ctx context.Context, key string) (string, error) {
 // GetAllConfig retrieves all configuration values
 func (s *DoltStore) GetAllConfig(ctx context.Context) (map[string]string, error) {
 	var result map[string]string
-	err := s.withReadTx(ctx, func(tx *sql.Tx) error {
+	err := s.withReadConn(ctx, func(q issueops.SQLQuerier) error {
 		var err error
-		result, err = issueops.GetAllConfigInTx(ctx, tx)
+		result, err = issueops.GetAllConfigInTx(ctx, q)
 		return err
 	})
 	return result, err
@@ -92,9 +92,9 @@ func (s *DoltStore) SetMetadata(ctx context.Context, key, value string) error {
 // GetMetadata retrieves a metadata value
 func (s *DoltStore) GetMetadata(ctx context.Context, key string) (string, error) {
 	var value string
-	err := s.withReadTx(ctx, func(tx *sql.Tx) error {
+	err := s.withReadConn(ctx, func(q issueops.SQLQuerier) error {
 		var err error
-		value, err = issueops.GetMetadataInTx(ctx, tx, key)
+		value, err = issueops.GetMetadataInTx(ctx, q, key)
 		return err
 	})
 	return value, err
@@ -112,9 +112,9 @@ func (s *DoltStore) SetLocalMetadata(ctx context.Context, key, value string) err
 // Returns ("", nil) if the key does not exist.
 func (s *DoltStore) GetLocalMetadata(ctx context.Context, key string) (string, error) {
 	var value string
-	err := s.withReadTx(ctx, func(tx *sql.Tx) error {
+	err := s.withReadConn(ctx, func(q issueops.SQLQuerier) error {
 		var err error
-		value, err = issueops.GetLocalMetadataInTx(ctx, tx, key)
+		value, err = issueops.GetLocalMetadataInTx(ctx, q, key)
 		return err
 	})
 	return value, err
--- a/internal/storage/dolt/config.go
+++ b/internal/storage/dolt/config.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/storage"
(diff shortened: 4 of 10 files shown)


maphew (commented):
Thanks for this - it's a genuinely useful piece of investigation. The trace harness and per-commit breakdown make a real case that `dolt_mode=server` over a WAN link pays serial round trips that don't need to be serial, and the reported numbers (12s -> 3s on `bd list`, timeouts -> ~8s on `bd create`) match the shape of #4102. The framing as "showcase, do not merge" plus the honest self-audit of which ideas already landed on `main` is exactly the right way to hand this off.

**Findings**

- [major] This branch has no common ancestor with current `upstream/main` (`git merge-base` returns empty; its base commit `f8b940016` predates the 2026-06-10 history rewrite and is not an ancestor of today's `main`). That's the actual cause of the CONFLICTING status - a plain rebase won't work, and the GitHub diff view is showing raw divergence rather than a clean 4-commit patch.
- [major] As a symptom of the above, the head commit's `cmd/bd/main.go` literally duplicates code: `maybeAutoPush(rootCtx)` is called twice, and the `storeMutex`/`storeActive`/`store.Close()` teardown block appears twice in sequence (once inside the gated block, once after). If this were force-merged as-is it would double-run auto-push and double-close the store on every command. This isn't a reason to fix the branch - it's a reason to cherry-pick the *ideas* onto current `main` rather than merge the branch itself.
- [minor] The "already on main" claims check out: `shouldRunPostCommandAutoExport` gating `maybeAutoExport`/`maybeAutoPush` is present (`cmd/bd/main.go`), and the top-level `GetStatistics` emptiness guard in `maybeAutoImportJSONL` is present (`cmd/bd/auto_import_upgrade.go`). Good self-assessment.
- [minor] Conversely, several optimizations are confirmed still absent from `main` and remain worth a follow-up issue: `InterpolateParams`/`MaxAllowedPacket` DSN tuning (`internal/storage/doltutil/dsn.go` has neither today), the `openServerConnection` fast path that skips the `SHOW DATABASES` probe for read-only connects, and gating `maybeAutoBackup` behind `isReadOnlyCommand` (still unconditional at `cmd/bd/main.go` today). These look like clean, small, independently-mergeable wins against #4102.
- [minor] The `bd stats`/`GetStatistics` hang-on-remote-server concern the author flags is real territory but already has a home: #4104 (parallelize `ScanIssueCounts`/`computeBlockedIDs`) is open and adjacent - worth linking rather than opening a fresh issue.
- [minor] `wispsEmpty`/`wispsStateKnown` (`internal/storage/dolt/config.go`, `queries.go`) has no invalidation path if a wisp is created later in the same process, unlike `allConfigCache` which `SetConfig`/`DeleteConfig` do invalidate - a note for whoever implements the real version.
- [nit] The `BD_DOLT_TRACE_SQL` harness (`internal/storage/dolt/sqltrace.go`) is self-contained, off-by-default, and useful on its own merits independent of the rest - it could stand as a tiny separate PR even outside this showcase.

**Conflict**: the branch predates the main rewrite and shares no history with it (confirmed above). If any of this is ever taken further as an actual patch, the remedy is `git rebase --onto upstream/main f8b940016 perf/remote-dolt-rtt-main-v2` to replay just the four real commits onto current history - a normal rebase across the stale base will fail or produce exactly the duplication seen above.

**Recommendation: retire.** The investigation and numbers are valuable and several ideas are still unapplied on `main`. Suggest closing this PR with thanks, and opening 2-3 small follow-up issues (DSN tuning, read-only connect fast path, `maybeAutoBackup` gating) referencing #4102, with a note pointing at #4104 for the stats-hang thread. Happy to file those if useful.

maphew:
Thanks again @Shockwave2k — closing this as intended. It was framed as a showcase from day one ("please do not merge"), and it did its job: the trace-driven investigation, the per-commit breakdown, and the honest audit of what main had already absorbed made this one of the most useful perf writeups this repo has received. Closing is bookkeeping, not a judgment on the work.

Where things stand against current `main` (as of this close):

**Already landed independently** — `InterpolateParams` in the server DSN (#4617), read-only gating of auto-export/auto-push, the auto-import emptiness guard, batched comment counts and routing-config reads, and #4104 (the `GetStatistics` parallelization adjacent to your `bd stats` hang heads-up) has been closed. Auto-backup against a remote server is also now skipped via the cross-filesystem guard (#3523).

**Still valuable and unlanded** — most of your round-trip work remains open opportunity: the `withReadConn` no-transaction read helper, the read-only `openServerConnection` fast path (skip the second pool + `SHOW DATABASES`), the `MaxAllowedPacket` DSN pin, skipping the preflight dial for non-local servers, the redundant post-open ping, the store-open session-cache prefetch (config map + wisps existence), parallelizing the auto-export bulk loads, a backend-aware `export.interval` default, and the `BD_DOLT_TRACE_SQL` harness that made all of it measurable. I've opened a tracking issue crediting this PR so those ideas don't get lost: #5273.

A note on mechanics for any future patch: this branch predates the June history rewrite, so it shares no ancestor with today's `main` — `git rebase --onto upstream/main f8b940016 <branch>` is the way to replay the four commits onto current history.

If you'd like to turn any slice of this into a real PR (the `withReadConn` helper plus the DSN pin would be a small, high-value first bite; the trace harness could stand alone), it would be very welcome — the numbers here already make the case. Thanks for the care you and your pairing setup put into this.

rileydev:
I am preparing a narrow follow-up for the still-unlanded remote TCP preflight item recorded in #5273.

This PR is closed, explicitly a showcase, and not maintainer-editable, so changing it in place is not possible. The follow-up credits @Shockwave2k for the topology split demonstrated here: retain the local/Unix preflight that supports auto-start, but let non-local TCP use the MySQL driver's bounded connection timeout. The new patch is limited to that behavior plus circuit-breaker/error-path regression tests; it does not carry over the broader performance series.

The focused follow-up is now #5993.

_codex-gpt-5.6-sol-unknown-reasoning on behalf of Riley Developer_

PR #4172, opened 2026-05-26
fix(hooks): bound prime and hook waits

### LINKED ISSUE #3701: bd prime hangs indefinitely on stale embedded-dolt lock — breaks LLM agent SessionStart hooks
## Summary
`bd prime` (and other commands acquiring the embedded-dolt lock) can hang indefinitely on a stale `.beads/embeddeddolt/.lock` left behind by a previous crashed/killed `bd` invocation. With no built-in timeout, this is a hard-deadlock for LLM coding agents that auto-run `bd prime` from a `SessionStart` hook — every new session spawns another stuck process, which compounds the problem (one failed session permanently breaks the directory until manually cleared).

## Repro
1. Configure `.claude/settings.json` `SessionStart` hook to run `bd prime` (Claude Code's recommended pattern).
2. Force-kill a running `bd` while it holds the embedded-dolt lock (panic, OOM, agent timeout, ^C in the wrong place — anything that leaves `.beads/embeddeddolt/.lock` open without releasing).
3. Open a new agent session in that repo.
4. `bd prime` blocks forever waiting on the lock. The new session itself hangs (never receives the SessionStart hook output), and now you have two stuck `bd` PIDs holding the lock.
5. Repeat. In my case I accumulated 17 zombie `bd`/`beads` processes across days before noticing every new session in the repo was hanging.

## Diagnosis (for anyone hitting this)
```bash
lsof +D .beads/embeddeddolt
# all the PIDs holding .lock are dead/zombie
kill -9 <pids>
ls .beads/embeddeddolt/.lock  # harmless once nothing holds it
```

## Proposed fixes (any/all)
1. **Built-in timeout on lock acquisition** in `bd prime` (and any read-mostly command) — default ~5–10s, env-overridable. Failing with a clear message is far better than hanging an agent forever.
2. **Self-diagnosing timeout output** — when `bd prime` times out, it should print exactly what happened: which lock it was waiting on, which PID(s) hold it, whether those PIDs are alive, and the one-line recovery command. Something like:
   ```
   bd prime: timed out after 10s acquiring .beads/embeddeddolt/.lock
     held by: bd PID 67319 (dead, exited ~3h ago)
              beads PID 56470 (dead)
              ... 15 more dead holders
     recovery: bd doctor --clear-lock
              (or kill any live PIDs above, then rm .beads/embeddeddolt/.lock)
   ```
   This turns a silent agent-killer into a self-explaining error. The next agent session reads the output, runs the suggested command, and continues — no human required.
3. **Stale-lock detection** — record the holding PID + start time + hostname inside `.lock`. On acquire, if the recorded PID isn't alive, reclaim the lock automatically. PID-in-file is portable; most embedded DBs do this.
4. **`bd doctor --clear-lock`** subcommand — safe version of the manual recovery (verify no live PID, then unlink). Fits the existing `bd doctor` surface.
5. **Document the recovery** prominently — README + `bd doctor` output. Discoverability is currently zero.

Ranking: (2) is highest-leverage — even without auto-recovery, a self-explaining timeout means the next agent run unblocks itself. (1) is the prerequisite. (3) is the actual fix. (4)+(5) are polish.

## Why this matters
This is the integration pitched to LLM-agent users (SessionStart hook = persistent memory across compactions). When it fails, it fails in the worst possible way for that audience: silently, at session start, with no timeout, and the failure mode compounds with each retry. A single hard-killed `bd` poisons the directory for every future agent session until a human intervenes.

## Environment
- bd 1.0.3 (Homebrew bottle)
- macOS 25.4 (Darwin)
- Backend: embedded dolt (`dolt_mode: embedded`)

### LINKED ISSUE #3707: `bd dolt remote add` self-deadlocks via pre-commit hook (hangs forever on macOS)
## Summary

`bd dolt remote add` (and likely any `bd dolt` write that records a config change via `git commit`) self-deadlocks against the integration `pre-commit` hook. The parent `bd` holds `.beads/embeddeddolt/.lock`, then forks `git commit` to persist the config change; the hook calls `bd hooks run pre-commit`, which blocks waiting for the same Dolt lock the parent still holds.

On Linux with `timeout(1)` available, the script's `BEADS_HOOK_TIMEOUT` (default 300s) eventually breaks the deadlock — a 5-minute hang per `bd dolt remote add`. **On macOS the `timeout` binary is not on `PATH` by default, the script falls into its no-timeout branch, and the deadlock is permanent.**

## Versions

- `bd version 1.0.3 (1b2dd2cb)` (also reproduces with the integration hook script header `BEGIN BEADS INTEGRATION v1.0.3`)
- macOS Darwin 25.4.0 (no `timeout` / `gtimeout` on default PATH)

## Reproduction

In any beads workspace where `git config core.hooksPath` points to `.beads/hooks` and the standard integration `pre-commit` is installed:

```sh
bd dolt remote add origin git+ssh://git@github.com/<user>/<repo>.git
```

Hangs indefinitely on macOS. `kill -9` is the only way out.

## Process tree at the deadlock

```
bd dolt remote add origin git+ssh://...   <- holds .beads/embeddeddolt/.lock
└─ git commit -m "bd: update sync.remote"
   └─ sh .beads/hooks/pre-commit
      └─ bd hooks run pre-commit          <- waiting for the same .lock
```

`lsof` on the parent shows it holding `/path/.beads/embeddeddolt/.lock`. The child `bd hooks run pre-commit` has no Dolt FDs open — it cannot acquire the lock.

## Why the existing timeout doesn't save macOS users

The hook script has this safeguard:

```sh
if command -v timeout >/dev/null 2>&1; then
  timeout "$_bd_timeout" bd hooks run pre-commit "$@"
  ...
else
  bd hooks run pre-commit "$@"   # <-- macOS lands here, no timeout
fi
```

Stock macOS ships neither `timeout` nor `gtimeout` (the latter is `coreutils` from Homebrew, not present by default), so on macOS the no-timeout branch runs and the parent waits forever.

## Suggested fixes

In rough order of preference:

1. **Don't take the Dolt lock across a `git commit` shell-out.** The parent `bd` should release (or never acquire) the embedded Dolt lock for the duration of the `git commit` it spawns to persist a config change. The deadlock is fundamentally that the hook needs a lock the parent is still holding for no reason.
2. **Re-entry guard in `bd hooks run`.** The integration hook already exports `BD_GIT_HOOK=1` before invoking `bd`. Have `bd hooks run` (or any `bd` subcommand that needs the embedded Dolt lock) detect that env var combined with the parent PID being a running `bd` process, and either short-circuit or skip Dolt access.
3. **Make the timeout actually work on macOS.** The hook's no-timeout branch should at least detect macOS without `timeout` and either use `perl -e 'alarm shift; exec @ARGV' "$_bd_timeout" ...` or print a loud warning so the user knows the safety net isn't engaged. Even with a working timeout, a 5-minute hang per `bd dolt remote add` is poor UX — fixes (1) or (2) should remain the primary path.

Happy to test a candidate fix.

maphew:
## Summary
- Bound `bd prime` memory-store opening with `BEADS_PRIME_TIMEOUT` (default 10s) so SessionStart hooks do not hang indefinitely when storage acquisition stalls.
- Treat non-positive `BEADS_PRIME_TIMEOUT` values as the default bound instead of silently disabling the timeout.
- Added macOS-capable hook timeout fallback: `timeout`, then `gtimeout`, then Perl `alarm`; SIGALRM exit handling is scoped to the Perl fallback.
- Routed internal beads-repo git commands through `git -c core.hooksPath=` and disabled templates so automated `.beads/` commits and pushes do not recurse through user hook managers. `GitCmdCWD()` remains for user-working-repo status/ref operations and does not perform beads commits.

## Why
Agent and hook workflows call `bd prime` and beads-managed git operations from automation paths where an unbounded storage open or recursive hook can strand the whole session. This keeps those automation paths bounded and non-reentrant.

## Tests
- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestGenerateHookSection_Timeout|TestPrimeStoreTimeoutNonPositiveUsesDefault|TestFormatMemoriesForPrimeTimesOutOpeningStore|TestGitCmdInDirSuppressesHooksViaGitConfig' -count=1`
- `CGO_ENABLED=1 go test -tags gms_pure_go ./internal/beads -run 'TestRepoContext_GitCmdCWD|TestGitCmd|TestRepoContext' -count=1`
- `git diff --check`

Fixes #3701
Fixes #3707

### DIFF
--- a/cmd/bd/hooks.go
+++ b/cmd/bd/hooks.go
@@ -67,6 +67,7 @@ func generateHookSection(hookName string) string {
 		"if command -v bd >/dev/null 2>&1; then\n" +
 		"  export BD_GIT_HOOK=1\n" +
 		"  _bd_timeout=${BEADS_HOOK_TIMEOUT:-" + fmt.Sprintf("%d", hookTimeoutSeconds) + "}\n" +
+		"  _bd_used_perl=0\n" +
 		"  if command -v timeout >/dev/null 2>&1; then\n" +
 		"    timeout \"$_bd_timeout\" bd hooks run " + hookName + " \"$@\"\n" +
 		"    _bd_exit=$?\n" +
@@ -74,6 +75,7 @@ func generateHookSection(hookName string) string {
 		"    gtimeout \"$_bd_timeout\" bd hooks run " + hookName + " \"$@\"\n" +
 		"    _bd_exit=$?\n" +
 		"  elif command -v perl >/dev/null 2>&1; then\n" +
+		"    _bd_used_perl=1\n" +
 		"    perl -e 'alarm shift; exec @ARGV' \"$_bd_timeout\" bd hooks run " + hookName + " \"$@\"\n" +
 		"    _bd_exit=$?\n" +
 		"  else\n" +
@@ -81,7 +83,7 @@ func generateHookSection(hookName string) string {
 		"    bd hooks run " + hookName + " \"$@\"\n" +
 		"    _bd_exit=$?\n" +
 		"  fi\n" +
-		"  if [ $_bd_exit -eq 124 ] || [ $_bd_exit -eq 142 ]; then\n" +
+		"  if [ $_bd_exit -eq 124 ] || { [ $_bd_used_perl -eq 1 ] && [ $_bd_exit -eq 142 ]; }; then\n" +
 		"    echo >&2 \"beads: hook '" + hookName + "' timed out after ${_bd_timeout}s — continuing without beads\"\n" +
 		"    _bd_exit=0\n" +
 		"  fi\n" +
--- a/cmd/bd/hooks.go
+++ b/cmd/bd/hooks.go
@@ -70,13 +70,20 @@ func generateHookSection(hookName string) string {
 		"  if command -v timeout >/dev/null 2>&1; then\n" +
 		"    timeout \"$_bd_timeout\" bd hooks run " + hookName + " \"$@\"\n" +
 		"    _bd_exit=$?\n" +
-		"    if [ $_bd_exit -eq 124 ]; then\n" +
-		"      echo >&2 \"beads: hook '" + hookName + "' timed out after ${_bd_timeout}s — continuing without beads\"\n" +
-		"      _bd_exit=0\n" +
-		"    fi\n" +
+		"  elif command -v gtimeout >/dev/null 2>&1; then\n" +
+		"    gtimeout \"$_bd_timeout\" bd hooks run " + hookName + " \"$@\"\n" +
+		"    _bd_exit=$?\n" +
+		"  elif command -v perl >/dev/null 2>&1; then\n" +
+		"    perl -e 'alarm shift; exec @ARGV' \"$_bd_timeout\" bd hooks run " + hookName + " \"$@\"\n" +
+		"    _bd_exit=$?\n" +
 		"  else\n" +
+		"    echo >&2 \"beads: hook '" + hookName + "' running without timeout; install coreutils or perl to enable BEADS_HOOK_TIMEOUT\"\n" +
 		"    bd hooks run " + hookName + " \"$@\"\n" +
 		"    _bd_exit=$?\n" +
+		"  fi\n" +
+		"  if [ $_bd_exit -eq 124 ] || [ $_bd_exit -eq 142 ]; then\n" +
+		"    echo >&2 \"beads: hook '" + hookName + "' timed out after ${_bd_timeout}s — continuing without beads\"\n" +
+		"    _bd_exit=0\n" +
 		"  fi\n" +
 		"  if [ $_bd_exit -eq 3 ]; then\n" +
 		"    echo >&2 \"beads: database not initialized — skipping hook '" + hookName + "'\"\n" +
--- a/cmd/bd/prime.go
+++ b/cmd/bd/prime.go
@@ -43,10 +43,16 @@ func primeStoreTimeout() time.Duration {
 		return primeStoreTimeoutDefault
 	}
 	if d, err := time.ParseDuration(raw); err == nil {
-		return d
+		if d > 0 {
+			return d
+		}
+		return primeStoreTimeoutDefault
 	}
 	if d, err := time.ParseDuration(raw + "s"); err == nil {
-		return d
+		if d > 0 {
+			return d
+		}
+		return primeStoreTimeoutDefault
 	}
 	return primeStoreTimeoutDefault
 }
--- a/cmd/bd/prime.go
+++ b/cmd/bd/prime.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -11,6 +12,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/beads"
@@ -27,6 +29,27 @@ var (
 	primeMemoriesOnly bool
 	primeHookJSONMode bool
 )
+
+const (
+	primeStoreTimeoutEnv     = "BEADS_PRIME_TIMEOUT"
+	primeStoreTimeoutDefault = 10 * time.Second
+)
+
+var ensureStoreActiveForPrime = ensureStoreActiveWithContext
+
+func primeStoreTimeout() time.Duration {
+	raw := strings.TrimSpace(os.Getenv(primeStoreTimeoutEnv))
+	if raw == "" {
+		return primeStoreTimeoutDefault
+	}
+	if d, err := time.ParseDuration(raw); err == nil {
+		return d
+	}
+	if d, err := time.ParseDuration(raw + "s"); err == nil {
+		return d
+	}
+	return primeStoreTimeoutDefault
+}
 
 // resolveGlobalPrimePath returns the path to ~/.config/beads/PRIME.md if it
 // exists. configDirOverride is used for testing; pass "" for production.
@@ -305,7 +328,17 @@ func outputMemoriesOnlyContext(w io.Writer) error {
 func formatMemoriesForPrime(compact bool) string {
 	// Try to initialize store if not already active (prime may run before other commands)
 	if store == nil {
-		if err := ensureDirectMode("memory injection"); err != nil {
+		timeout := primeStoreTimeout()
+		ctx := context.Background()
+		var cancel context.CancelFunc
+		if timeout > 0 {
+			ctx, cancel = context.WithTimeout(ctx, timeout)
+			defer cancel()
+		}
+		if err := ensureStoreActiveForPrime(ctx); err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
+				return formatPrimeMemoryTimeout(compact, timeout)
+			}
 			return "" // Silently skip — store unavailable
 		}
 	}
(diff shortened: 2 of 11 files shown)


internal/beads/context.go
```diff
@@ -243,7 +244,7 @@ func (rc *RepoContext) GitCmd(ctx context.Context, args ...string) *exec.Cmd {
 	// Security: Disable git hooks and templates to prevent code execution
 	// in potentially malicious repositories (SEC-001, SEC-002)
 	cmd.Env = append(os.Environ(),
-		"GIT_HOOKS_PATH=",            // Disable hooks
+		"GIT_HOOKS_PATH=",            // Defense in depth; core.hooksPath is disabled via git -c above.
```
maphew: `GIT_HOOKS_PATH` is not a real git environment variable. Git only honors the `core.hooksPath` config option (and reads `GIT_DIR`/`GIT_WORK_TREE` etc. from env). So the `GIT_HOOKS_PATH=""` entry here has never disabled hooks on its own — it is dead code. The new `-c core.hooksPath=` flag added in the same hunk is in fact the *only* mechanism actually doing the work, which makes the new comment ("Defense in depth; core.hooksPath is disabled via git -c above") misleading: there is no second layer of defense.

Suggest either dropping the `GIT_HOOKS_PATH=` line entirely, or replacing it with `GIT_CONFIG_COUNT=1`, `GIT_CONFIG_KEY_0=core.hooksPath`, `GIT_CONFIG_VALUE_0=` to give a genuine env-based defense in depth (so a caller that strips `-c` flags still gets hook suppression).

cmd/bd/hooks.go
```diff
@@ -70,14 +70,21 @@ func generateHookSection(hookName string) string {
...
-		"    if [ $_bd_exit -eq 124 ]; then\n" +
-		"      echo >&2 \"beads: hook '" + hookName + "' timed out after ${_bd_timeout}s — continuing without beads\"\n" +
-		"      _bd_exit=0\n" +
-		"    fi\n" +
+		"  elif command -v gtimeout >/dev/null 2>&1; then\n" +
+		"    gtimeout \"$_bd_timeout\" bd hooks run " + hookName + " \"$@\"\n" +
+		"    _bd_exit=$?\n" +
+		"  elif command -v perl >/dev/null 2>&1; then\n" +
+		"    perl -e 'alarm shift; exec @ARGV' \"$_bd_timeout\" bd hooks run " + hookName + " \"$@\"\n" +
+		"    _bd_exit=$?\n" +
```
maphew, line 80: The perl fallback uses `exec @ARGV` which replaces the perl process with `bd`. After `exec`, the SIGALRM handler is reset to default (terminate) and the alarm timer continues to run in the new process — so when it fires, it will kill `bd` mid-flight without any opportunity for cleanup, then the exit status will be 142 (128+SIGALRM). That part is fine.

However, two concerns:

1. The `alarm shift; exec @ARGV` form does not propagate a non-zero exec failure cleanly. If `bd` is no longer on PATH for some reason (unusual, since the outer `command -v bd` check guarded it, but the PATH could differ here under the hook env), `exec` fails and perl exits with its own error code, not 127. Minor.

2. More importantly: in the no-timeout fallback branch (the `else` at line 81 in the new diff), `_bd_exit` could legitimately equal 142 if `bd` itself is killed by SIGALRM from some unrelated source, and the bottom-of-section check at the new `[ $_bd_exit -eq 124 ] || [ $_bd_exit -eq 142 ]` block then silently swallows that as a "timeout" even though no timeout was configured. Suggest scoping the 142 check to only the perl branch (set a flag like `_bd_used_perl=1` inside the `elif perl` branch, and gate the 142 handling on it).

cmd/bd/prime.go
```diff
@@ -305,7 +328,17 @@ func outputMemoriesOnlyContext(w io.Writer) error {
 func formatMemoriesForPrime(compact bool) string {
 	// Try to initialize store if not already active (prime may run before other commands)
 	if store == nil {
-		if err := ensureDirectMode("memory injection"); err != nil {
+		timeout := primeStoreTimeout()
+		ctx := context.Background()
+		var cancel context.CancelFunc
+		if timeout > 0 {
+			ctx, cancel = context.WithTimeout(ctx, timeout)
+			defer cancel()
```
maphew, line 342: `primeStoreTimeout()` can return `0` (or a negative duration) if `BEADS_PRIME_TIMEOUT=0` / `0s` / `-5s` is set. When that happens, this `if timeout > 0` branch is skipped and `ensureStoreActiveForPrime` is called with a plain `context.Background()` — no deadline at all. A user trying to *disable* the timeout by setting `BEADS_PRIME_TIMEOUT=0` would get the exact hang the PR is trying to prevent, while a user fat-fingering `BEADS_PRIME_TIMEOUT=-10s` gets the same silent regression.

The PR summary describes the timeout as bounded with a default of 10s; users may reasonably assume `0` opts out (the convention in many other tools). Two reasonable fixes:

- Treat `<= 0` as "use the default" inside `primeStoreTimeout()`, so the only way to opt out is removing the var.
- Or explicitly document `0` as "no timeout" *and* surface a warning at injection time so the hang is at least attributable.

Either is fine; status quo is the worst outcome — a knob that silently disables the safety the PR adds.

internal/beads/context.go
```diff
@@ -233,7 +233,8 @@ func getRepoRootFromPath(path string) (string, error) {
 // We explicitly set GIT_DIR and GIT_WORK_TREE to ensure git operates on
 // the correct repository (the one containing .beads/).
 func (rc *RepoContext) GitCmd(ctx context.Context, args ...string) *exec.Cmd {
-	cmd := exec.CommandContext(ctx, "git", args...)
+	gitArgs := append([]string{"-c", "core.hooksPath="}, args...)
```
maphew, line 236: This change unconditionally prepends `-c core.hooksPath=` to *every* `rc.GitCmd` invocation across the codebase, not just the internal `.beads/config.yaml` commit that motivated the PR. Greps show ~dozens of callers (status, log, commit, push, fetch, etc.).

That is almost certainly the right call for security (the SEC-001/SEC-002 comment already justified it), but it is a meaningful behavior change beyond the PR title's scope: any user-side `pre-commit`, `commit-msg`, or `pre-push` hooks (conventional-commit linters, secret scanners, sign-off enforcers) that previously fired on beads-driven commits to `.beads/` will now be silently bypassed.

Two suggestions:
1. Call this out explicitly in the PR description / CHANGELOG so users with `commit-msg` enforcement notice.
2. Consider whether the `GitCmdCWD` helper (the one operating on the user's working repo, not the beads repo) should also suppress hooks. The diff only touches `GitCmd`. If `GitCmdCWD` is ever used for a user-facing commit, treatment should be consistent in one direction or the other — verify and document.

maphew:
Addressed the review in `36f7e8a70`:

- Removed the dead `GIT_HOOKS_PATH=` hook suppression and switched the worktree helper to the same `git -c core.hooksPath=` mechanism.
- Scoped generated-hook exit `142` handling to the Perl `alarm` fallback only.
- Made `BEADS_PRIME_TIMEOUT=0`, `0s`, and negative values fall back to the 10s default.
- Added a CHANGELOG entry and updated the PR body to call out the broader internal beads-repo hook suppression behavior.
- Verified that `GitCmdCWD()` usage is user-working-repo status/ref/checkout oriented, not beads commit/push plumbing.

Validation:

- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestGenerateHookSection_Timeout|TestPrimeStoreTimeoutNonPositiveUsesDefault|TestFormatMemoriesForPrimeTimesOutOpeningStore|TestGitCmdInDirSuppressesHooksViaGitConfig' -count=1`
- `CGO_ENABLED=1 go test -tags gms_pure_go ./internal/beads -run 'TestRepoContext_GitCmdCWD|TestGitCmd|TestRepoContext' -count=1`
- `git diff --check`

PR #4175, opened 2026-05-26
bd-91bzp: make issue prefix mutations explicit

fengning-starsend:
Agent: codex-desktop

Feature-Key: bd-91bzp

## Summary
- Treat `issue_prefix` as identity state after initialization.
- Allow initial/idempotent prefix writes, but reject existing-prefix changes unless `BEADS_ALLOW_PREFIX_MUTATION=1`.
- Let the dedicated `rename-prefix` command set that break-glass internally for legitimate migrations.
- Cover both Dolt storage paths and the SQL domain config repository.

## Verification
- `go test ./internal/storage/domain/db`
- `go test ./internal/storage/dolt -run 'TestSetConfig(NormalizesIssuePrefix|RejectsIssuePrefixMutation|AllowsIssuePrefixMutationWithBreakGlass)'`\n- `go test -tags gms_pure_go ./cmd/bd -run 'TestConfig|TestRenamePrefix|TestSetConfig|TestConfigSetMany'`\n- `go test -tags gms_pure_go ./cmd/bd ./internal/storage/domain/db ./internal/storage/dolt -run 'TestConfig|TestRenamePrefix|TestSetConfig(NormalizesIssuePrefix|RejectsIssuePrefixMutation|AllowsIssuePrefixMutationWithBreakGlass)'`\n- `go build -tags gms_pure_go -o /tmp/agents/bd-91bzp/bd ./cmd/bd`\n- temp repo smoke: init/create/show/close/config get, blocked `config set issue_prefix`, custom config set, `rename-prefix` migration\n\n## Note\nA bare `go build ./cmd/bd` on this host fails on missing ICU headers (`unicode/uregex.h`). The repo CI and docs require `-tags gms_pure_go` for normal builds; tagged build passed. A candidate binary direct-open against the shared live Beads runtime hit a pre-existing schema/migration mismatch, so I did not deploy the candidate binary to live. The installed production `bdx` path remains healthy.

### DIFF
--- a/cmd/bd/rename_prefix.go
+++ b/cmd/bd/rename_prefix.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/beads/internal/storage"
+	"github.com/steveyegge/beads/internal/storage/issueops"
 	"github.com/steveyegge/beads/internal/types"
 	"github.com/steveyegge/beads/internal/ui"
 	"github.com/steveyegge/beads/internal/utils"
@@ -124,7 +125,9 @@ NOTE: This is a rare operation. Most users never need this command.`,
 		if len(issues) == 0 {
 			fmt.Printf("No issues to rename. Updating prefix to %s\n", newPrefix)
 			if !dryRun {
-				if err := store.SetConfig(ctx, "issue_prefix", newPrefix); err != nil {
+				if err := withPrefixMutationAllowed(func() error {
+					return store.SetConfig(ctx, "issue_prefix", newPrefix)
+				}); err != nil {
 					FatalError("failed to update prefix: %v", err)
 				}
 				commandDidWrite.Store(true)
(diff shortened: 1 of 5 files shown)


maphew:

maphew:
Thanks for this, and apologies for the two-month silence — worse, the one reply you did get asked for a bug repro, which made no sense for a feature PR. That was our sweep mis-classifying this; sorry. To be clear now: the idea is good and we want it.

Treating `issue_prefix` as identity state at the storage layer is the right defense-in-depth. We already block it at the CLI (`cmd/bd/config.go:993`), but that leaves every programmatic `SetConfig` path unguarded, which is exactly the hole your patch closes — and your test coverage at both layers (`internal/storage/dolt/dolt_test.go`, `internal/storage/domain/db/config_test.go`) is appreciated.

Two things stand between this and merge:

1. **The break-glass only covers `rename-prefix`, but main has other sanctioned mutation paths.** The concrete one: `cmd/bd/import.go:369-374` intentionally overwrites a stale DB prefix from config.yaml during import (be-llaf) — its comment even says it deliberately bypasses the config-set guard. Under this patch that write errors, and because the caller is `if setErr == nil { ... }`, the prefix sync just silently stops happening. Any allow-mechanism needs to enumerate the sanctioned writers (import sync, rename-prefix, repair), not just one. Relatedly, we'd prefer an explicit in-process mechanism (a context value or a dedicated force-set method on the config repository) over an env var — env vars are process-global and leak into child processes, so a crash mid-`rename-prefix` or a spawned subcommand can widen the window the guard is meant to close.

2. **The branch has drifted under you.** `issueops.SetConfigInTx` on main now takes `tx DBTX` rather than `*sql.Tx` (`internal/storage/issueops/config_metadata.go:12`), so the PR is conflicting and needs a rebase.

One sequencing note: gastownhall/beads#4827 reports that `rename-prefix` currently doubles the prefix on populated shared DBs. Since your guard makes `rename-prefix` the only sanctioned rename path, we should land that fix before or alongside this so users aren't locked in with a broken escape hatch.

What happens next is your call: if you'd like to rebase and rework the allow-mechanism per (1), we'll review promptly this time. If you'd rather not, say the word and we'll carry it over the line ourselves as a follow-up that preserves your authorship — the storage-layer guard and both test suites are keepers either way.

PR #4180, opened 2026-05-26
docs: publish CI timing report

maphew:
## Why

This gives maintainers current Actions timing data before splitting CI into fast, risk, merge, main, nightly, and manual tiers. The report makes the critical path and runner-minute hot spots visible instead of relying on anecdotal CI latency.

## What

- Add a reproducible GitHub Actions timing report script using `gh` and standard-library Python.
- Check in Markdown and HTML timing reports under `docs/reports/`.
- Summarize P50/P95 wall-clock, queue time, slowest jobs, critical path tail jobs, skipped/pending patterns, and tier recommendations.

## Validation

- `python3 scripts/ci/ci_timing_report.py --repo gastownhall/beads --runs 120 --out-dir docs/reports`
- `python3 -m py_compile scripts/ci/ci_timing_report.py`
- HTML parser smoke check
- `git diff --check`

Tracks mybd-xk6q.1.


maphew:
Self-review, thinking about the idea (measure CI before splitting it into tiers) rather than the diff itself.

## What this gets right

- **Measure before tiering.** xk6q.3/.4 propose to carve CI into Fast/Risk/Merge/Main/Nightly. Doing that off anecdote is how you ship a "fast" tier that still feels slow. P50/P95 wall-clock + per-job critical-path + failure rates is the right minimum fact base.
- **Critical-path tail as the primary lens.** For PR feedback latency, wall-clock trumps runner-minutes. Tail-job is an honest approximation given that public Actions metadata has no step-DAG, and the report says so.
- **Cancellations separated from failures.** Most quick dashboards mash these together and produce garbage flake rates (concurrency cancels swamp real failures). Separating them is the single most thoughtful design choice here.
- **Per-runner-class cost breakdown.** macOS minutes (premium pricing), Windows minutes (queue contention), embedded-Dolt shard minutes (fan-out noise) get budgeted differently by maintainers. Splitting them maps onto how the decision actually gets made.
- **Reproducible, gh-only, stdlib-only.** Any maintainer can rerun it; no new dep, no auth dance beyond `gh`.

## Structural concerns about the idea (most important first)

**1. The PR conflates three artifacts with three different decay rates:**
- A measurement *tool* (the script) — evergreen.
- A measurement *snapshot* (the dated report) — perishable in weeks.
- A *policy proposal* (the Tier Recommendations table) — editorial, not data-derived.

Bundled, the snapshot rots in `docs/reports/` and misleads anyone who finds it 3 months from now; the policy table is a static string in the script so it'll look identical regardless of what next month's data shows; and the tooling can't be iterated without re-litigating the policy. Probably better as:
- PR A: the script under `scripts/ci/` only.
- PR B: a scheduled workflow that overwrites `docs/reports/ci-timing-report.md` weekly with an "Auto-generated; do not edit" header.
- xk6q.4 / a separate `docs/ci/tiering.md`: the tier recommendations, where maintainers can argue them on their own merits.

**2. No counterfactual.** The report tells you what CI looks like today; it doesn't model what each proposed tier-split would do to P95 PR feedback or runner-minute spend. That's the analytic step that turns "interesting data" into "decision support" — and is the actually load-bearing question for xk6q.3. Worth a section like: "If Fast PR ships as proposed, modeled successful-PR P95 drops from 14.1m → ~Xm based on slowest remaining job." Without that, the tier table at the bottom is a hunch dressed in tables.

**3. 120 runs is one week and dominated by whatever was hot that week.** Single-snapshot evidence is thin for a CI restructure that affects every contributor. Weekly-bucketed trend (P50/P95 by week for the last 4–8 weeks) would protect against a bad-day flake spike skewing the conclusion.

**4. Tail-job critical path can mislead about parallelism gains.** "Test (Embedded Dolt Cmd shard) was the tail 19 times" doesn't tell you whether killing shards helps or whether you'd just hit the next-longest job. A *near-tail* (P95 of the slowest non-tail job per run) would surface that. Otherwise the report can argue for changes that don't actually reduce wall-clock.

**5. "Failure rate" mixes flake with genuinely-broken PRs.** First-attempt failure is mostly people making typos. For xk6q.7's quarantine policy, the useful signal is *retried success rate* (or failures on `push`/`main` where the code is presumed-green). The current column will overstate flakiness and could push maintainers to quarantine lint just because contributors keep mis-formatting.

## Smaller notes

- `run_started_at == created_at` is flagged correctly, but a follow-up could note the workaround (workflow_run / check_run webhooks) so a future iteration can actually measure run-queue time rather than only job-queue time.
- The Tier Recommendations table reads as data-derived but is a static string in `render_markdown`. At minimum, mark it "Author's proposal, not computed."
- Hand-rolled markdown→HTML works for this report's inputs but will silently fail on inline backticks, links, or bold. If the report grows, render both formats from the summary dict, or just publish Markdown and let GitHub render it.

## Bottom line

The work plugs into a topic the maintainers are actively moving on (#3459 just merged tier work). Landing it as a single PR mixes an evergreen tool, a perishable snapshot, and an editorial policy table — three things with different review audiences and different shelf lives. The highest-leverage change is adding a counterfactual ("here's what each tier scenario would do to P95") and splitting the policy proposal from the measurement. Done that way, this becomes the load-bearing PR that lets xk6q.3 and xk6q.4 be argued from data instead of intuition.

scripts/ci/ci_timing_report.py
```diff
@@ -0,0 +1,600 @@
...
+  --bg: #f7f8fa;
+  --fg: #20242a;
+  --muted: #5d6673;
+  --line: #d8dde6;
+  --panel: #ffffff;
+  --accent: #0b6bcb;
+}}
+@media (prefers-color-scheme: dark) {{
+  :root {{
+    --bg: #111418;
```
maphew, line 498: `gh api --paginate` without `--slurp` concatenates page payloads back-to-back. When the endpoint returns an object (as `actions/runs/{id}/jobs` does), multi-page output is `{...}{...}` — not valid JSON, and `json.loads` (in `gh_json` above) raises `JSONDecodeError`. The `isinstance(data, list)` branch on line 499 is dead because pagination never returns a list for object responses.

In this sample no single run exceeded 100 jobs so the script never tripped, but it is a latent crash the first time a run has >100 jobs (e.g. an expanded matrix). Either pass `--slurp` and iterate the resulting list, or drop `--paginate` and loop manually with `?page=N`.

```suggestion
    data = gh_json(["api", "--method", "GET", endpoint, "--paginate", "--slurp", "-f", "per_page=100"])
    jobs: list[dict[str, Any]] = []
    for page in data or []:
        jobs.extend(page.get("jobs", []))
    return jobs
```

docs/reports/ci-timing-report.md
```diff
@@ -0,0 +1,151 @@
...
+- Run queue P50/P95: 0s / 0s
+- Job queue P50/P95: 2s / 49s
+
+GitHub `run_started_at` is currently equal to `created_at` for most runs, so run-level queue time is less useful than per-job queue time. Job queue time uses `job.created_at` to `job.started_at` when present.
+
+## Checked Workflow Files
+
+| Workflow | Local file |
+|---|---|
+| CI | `.github/workflows/ci.yml` |
```
maphew, line 29: This section is labelled "Checked Workflow Files" and lists 6 yml files, but `collect_runs` in the script fetches **all** recent runs from the repo regardless of workflow. The "Wall-Clock by Workflow and Event" table below (lines 38–47) confirms it: `Copilot`, `Deploy Documentation`, and `Update vendorHash for dependabot Go bumps` are summarized despite not being in `WORKFLOW_FILES`.

So `WORKFLOW_FILES` is only the input to the file-mapping table — it does not scope the analysis. A reader will reasonably infer "the report covers these 6 workflows" and miss that the data set is the global `gh run list --limit 120`. Suggest renaming the section ("Workflow name → file mapping") or actually filtering `runs` by `workflowName in {row[0] for row in workflows}` before analysis.

docs/reports/ci-timing-report.md
```diff
@@ -0,0 +1,151 @@
...
+|---|---:|---|
+| Check for .beads changes | 6 | Mostly expected conditional gating or cancelled superseded runs. |
+| Update default.nix vendorHash | 3 | Mostly expected conditional gating or cancelled superseded runs. |
+| Test (Embedded Dolt Storage) | 2 | Mostly expected conditional gating or cancelled superseded runs. |
+| Test (Embedded Dolt Cmd ${{ matrix.shard }}/${{ strategy.job-total }}) | 2 | Mostly expected conditional gating or cancelled superseded runs. |
+| Differential Regression (v0.49.6 baseline) | 2 | Mostly expected conditional gating or cancelled superseded runs. |
+| Build (Embedded Dolt) | 1 | Mostly expected conditional gating or cancelled superseded runs. |
+| Regression Tests run 26479093922 | 0 | `in_progress` at collection time. |
+| CI run 26479093903 | 0 | `in_progress` at collection time. |
+| Regression Tests run 26479089036 | 0 | `in_progress` at collection time. |
```
maphew, line 121: Heads-up: this table is a hardcoded string in `scripts/ci/ci_timing_report.py` (`render_markdown`, around line 794), not derived from the run data above. It sits inside an auto-generated report between two computed tables and a "Key Findings" section that *is* computed, so a future reader will read these recommendations as data-derived.

Two low-cost fixes:
- Prepend "Author's proposal, not computed from the data above" to the section.
- Or move the tier policy out of the script into a sibling `docs/ci/tiering.md` and link to it, so it can evolve on its own review cadence rather than being silently re-emitted every time the report regenerates.

scripts/ci/ci_timing_report.py
```diff
@@ -0,0 +1,600 @@
...
+        reverse=True,
+    )[:20]
+    lines.extend(["", "## Slowest Jobs", "", "| Job | Runs | Workflow | P50 | P95 | Max | Failure rate | Queue P95 |", "|---|---:|---|---:|---:|---:|---:|---:|"])
+    for key, agg in slow_jobs:
+        durations = agg["durations"]
+        queues = agg["queues"]
+        workflow = agg["workflow"].most_common(1)[0][0] if agg["workflow"] else ""
+        fail_rate = agg["non_success"] / agg["count"] if agg["count"] else 0
+        lines.append(
+            f"| {key} | {agg['count']} | {workflow} | {fmt_minutes(percentile(durations, 50))} | {fmt_minutes(percentile(durations, 95))} | {fmt_minutes(max(durations))} | {fmt_pct(fail_rate)} | {fmt_minutes(percentile(queues, 95))} |"
```
maphew, line 335: `fmt_pct` (`{:.0%}` truncating to whole percent) means `2/640 = 0.3%` renders as `0%`. The committed report shows `Failure rate: 0%` next to `Failures: 2` in the same row of "Failure and Flakiness Suspects" — reads as a typo to anyone scanning the table. Suggest `{:.1%}`, or render `<1%` for small-but-nonzero values. Same `fmt_pct` is reused for both columns in the Failure/Flakiness table below.

scripts/ci/ci_timing_report.py
```diff
@@ -0,0 +1,600 @@
...
+    if in_progress:
+        for run in in_progress[:8]:
+            lines.append(f"| {run.get('workflowName')} run {run.get('databaseId')} | 0 | `{run.get('status')}` at collection time. |")
+    if not skipped_jobs and not in_progress:
+        lines.append("| None observed | 0 | No pending or skipped pattern surfaced in this sample. |")
+
+    total_runner_minutes = sum(row["duration_minutes"] or 0 for row in job_rows if row["conclusion"] == "success")
+    macos_minutes = sum(row["duration_minutes"] or 0 for row in job_rows if "macos" in row["name"].lower() or "macos" in row["runner"].lower())
+    windows_minutes = sum(row["duration_minutes"] or 0 for row in job_rows if "windows" in row["name"].lower() or "windows" in row["runner"].lower())
+    embedded_minutes = sum(row["duration_minutes"] or 0 for row in job_rows if "embedded" in row["name"].lower())
```
maphew, line 397: The three `*_minutes` sums on lines 395-397 don't filter by `row["conclusion"] == "success"`, but the rendered bullets below label them "macOS successful job time", "Windows successful job time", "Embedded Dolt successful job time". They include cancelled/failed durations, contradicting `total_runner_minutes` on line 394 which does filter. With 20 cancels on the embedded cmd shards alone, the embedded total is meaningfully inflated relative to its label.

Drop "successful" from the bullet wording, or add the same conclusion filter:

```suggestion
    embedded_minutes = sum(row["duration_minutes"] or 0 for row in job_rows if row["conclusion"] == "success" and "embedded" in row["name"].lower())
```

maphew:
Closing as overtaken by events:

- #4211 implemented the Fast/Risk/Merge/Main/Nightly tiering this report was written to motivate, and shipped its own measurement harness (`ci-measurements.yml` plus `scripts/ci/lib/timing.sh`).
- The script targets the deleted `CI` workflow, so a regeneration against today's topology renders "n/a" headline metrics and lists workflow files that no longer exist. A genuine refresh would be a rewrite, not a rerun.
- A committed point-in-time report under `docs/reports/` is exactly the stale-generated-docs pattern #4301 is pushing back on.

If retrospective Actions-metadata reporting (observational P50/P95 across real runs, complementary to the benchmark-style measurements) turns out to be wanted, the rewrite notes are captured in our tracker; happy to bring back a workflow-topology-aware script-only version then.

PR #4183, opened 2026-05-26
test: split tracker pure-go helpers

maphew:
## Why

`cmd/bd` recently split pure-Go-safe test helpers away from CGO-only helpers so future pure-Go tests can compile without dragging in embedded Dolt setup. The same helper contamination existed in `internal/storage/embeddeddolt` and `internal/tracker`: adding a pure-Go test there would still inherit CGO-only setup and fail before it could test package behavior.

This PR applies the same pattern to those packages and extends the pure-Go compile guard so the regression is caught in CI.

## What Changed

- Moved tracker mock tracker/mapper helpers into an untagged pure-Go helper file.
- Kept tracker Dolt/TestMain setup behind `//go:build cgo`.
- Made the tracker rate-limit test use a pure in-memory test store.
- Marked embeddeddolt cache cleanup test as CGO-only and added a pure helper landing-zone file.
- Extended the pure-Go CI compile step to include `internal/storage/embeddeddolt` and `internal/tracker`.

## Validation

- `CGO_ENABLED=0 go test -tags gms_pure_go -c -o /dev/null ./internal/storage/embeddeddolt`
- `CGO_ENABLED=0 go test -tags gms_pure_go -c -o /dev/null ./internal/tracker`
- `go test ./internal/tracker`
- `go test -tags gms_pure_go ./internal/storage/embeddeddolt`
- `go test -tags gms_pure_go ./internal/tracker`
- `git diff --check`

`go test ./internal/storage/embeddeddolt` could not run on this host because the local CGO toolchain is missing ICU header `unicode/uregex.h`; package tests pass under the pure-Go tag used by the new guard.

_codex-gpt-5.5-medium on behalf of CI Bot_


maphew (commented):
No findings.

I checked the helper split against the tracker sync path and the new pure-Go compile guards. The change keeps the Dolt-backed tracker tests behind `//go:build cgo`, leaves the rate-limit test and shared tracker mocks pure-Go-safe, and extends CI coverage to catch the packages that previously inherited CGO-only test setup.

Local validation:
- `CGO_ENABLED=0 go test -tags gms_pure_go -c -o /tmp/bd-tracker-puregeo-test-review ./internal/tracker`
- `CGO_ENABLED=0 go test -tags gms_pure_go -c -o /tmp/bd-embeddeddolt-puregeo-test-review ./internal/storage/embeddeddolt`
- `CGO_ENABLED=0 go test -tags gms_pure_go -run TestEnginePushAbortsLoopOnRateLimit ./internal/tracker`
- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/embeddeddolt ./internal/tracker`
- `go test ./internal/tracker`

GitHub CI is green. This is ready to merge from my review.

_codex-gpt-5.5-xhigh on behalf of matt wilkie_

PR #4185, opened 2026-05-26
fix(provider): bd create

coffeegoddd:


maphew:
Review findings

  - P1 .worktrees/beads/pr-4185-review/internal/storage/domain/
    issue.go:388 / .worktrees/beads/pr-4185-review/internal/storage/domain/db/
    child_counter.go:29: child counter routing is based on the child’s useWisp mode,
    not the parent’s actual table. The embedded path uses IsActiveWispInTx(parentID)
    before choosing child_counters vs wisp_child_counters; this new path passes
    UseWispsTable: useWisp. That breaks mixed parent-child cases such as proxied bd
    create --ephemeral --parent <regular issue> or a regular child under a wisp parent,
    either failing FK constraints or recording counters in the wrong table.
  - P1 .worktrees/beads/pr-4185-review/internal/storage/domain/db/dependency.go:83:    dependency insertion always writes the target into depends_on_issue_id, even when    the target is a wisp. The new proxied graph/wisp paths can create wisp-to-wisp    parent-child edges, but those should land in depends_on_wisp_id. The existing test    only checks COALESCE(...), so it misses the wrong physical column; this can break    FK/cascade/rename and wisp-aware dependency logic.
  - P2 .worktrees/beads/pr-4185-review/cmd/bd/graph_apply.go:216: CI is currently red    on lint. warnUnknownGraphFields returns []string, but the return value is unused in    production, and unparam flags it. Either make it void and adjust tests, or    structure tests another way.

_codex-gpt-5.5-high on behalf of maphew_

PR #4186, opened 2026-05-27
fix(show): propagate IterDependents/IterComments errors instead of swallowing them (be-kzipe)

quad341:
## Summary

`bd show --include-dependents` and `--include-comments` wrapped iterator setup in `if err == nil { ... }` with **no else branch**, so a non-nil setup error — or a mid-scan iterator error — produced an empty list with **no signal** to the user. This is the explicit follow-up recommended in #4118: it is *why* the `be-kjp7x` dependents regression surfaced as silently-empty JSON (self-contradictory `dependent_count: N` with an empty `dependents[]`) instead of a loud failure.

Both branches now:
- `FatalErrorRespectJSON(...)` on the **setup** error, and
- `FatalErrorRespectJSON(...)` on **`iter.Err()`** after the scan loop.

Fixes `be-kzipe`.

## Why this PR (re-cut of #4162)

This supersedes #4162. That PR's branch had drifted onto the **abandoned `be-z09krv` daemon-iter transport** (`cmd/bddgen`, `internal/storage/rpc/*`, the `be-z09krv` release gate, daemon-iter config/error vars) — ~6740 lines across 36 files that never landed on `main` and must not be re-introduced. This branch is cut fresh from current `main` and contains **only** the intended `cmd/bd/show.go` hardening (one file, +46/-38).

## Test plan

- [x] `CGO_ENABLED=0 go build -tags gms_pure_go ./cmd/bd/` — clean
- [x] `CGO_ENABLED=0 go vet -tags gms_pure_go ./cmd/bd/` — clean
- [x] `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd/... -run TestShow` — pass
- [x] `gofmt -l cmd/bd/show.go` — clean

No new unit test is added: exercising the new error branches needs an iterator/store that returns an error, which the embedded pure-Go backend's slice-stub iterators don't do in normal operation. The Dolt-server guard `TestCountAndIterIncludeWispDependencies` covers the happy path once a test Dolt server is available. The change is a mechanical swap of silent error-swallowing for `FatalErrorRespectJSON`.


maphew:

quad341:
Thanks — and thanks for merging this.

The concrete reproduction that's missing from this thread actually lives in the PR this one is the follow-up to: **#4118** (`be-kjp7x`, *"bd show --include-dependents silently returns no dependents on Dolt server (regression from #4010)"*). That's the real-world incident this hardening responds to — #4118's own "Follow-up" section is what recommended this change.

**Repro (as observed in #4118) — storage mode is the key variable: Dolt sql-server backend, post-migration `0043`. OS-independent (it's a schema/query mismatch, not platform-specific).**

- An issue with dependents on a post-`0043` Dolt server (the report used `bd-rbxi`, 12 dependents).
- Run: `bd show bd-rbxi --json --include-dependents`
- **Expected:** `dependent_count: 12` with a 12-element `dependents[]`.
- **Actual (buggy):** `dependent_count: 12` with `dependents[]` **absent** — self-contradictory JSON, **exit 0, no error printed**.

Root cause of the *silence* (which is what this PR fixes): `IterDependentsWithMetadata` failed with

```
ERROR 1105 (HY000): table "d" does not have column "depends_on_id"
```

because migration `0043_drop_dependencies_generated_column` dropped the generated column the iter query still filtered on. That error returned to the `if err == nil { … }` guard in `cmd/bd/show.go` (line ~163 at the time of #4118), which had **no else branch**, so the 1105 was dropped on the floor and `details.Dependents` stayed nil — while the separate O(1) `CountDependents` still reported 12. This PR makes both the setup error and the post-loop `iter.Err()` propagate via `HandleErrorRespectJSON` (on both the dependents and comments paths), so the same failure now exits non-zero with `iter dependents <id>: …` instead of silently truncating.

**Why there's no plain-command repro against current `main`:** the underlying query bug is already fixed — `internal/storage/dolt/iter_dependents.go` now filters on the split-column target `COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external)`, so the 1105 no longer fires. This PR is therefore purely *defensive*: it doesn't fix a currently-reproducible symptom, it guarantees the *next* iterator break (a setup error, or a mid-scan `iter.Err()`, on either path) surfaces loudly instead of as silently-truncated output. That's also why no unit test is included — the embedded pure-Go backend's slice-stub iterators don't return errors in normal operation, and the Dolt-server guard `TestCountAndIterIncludeWispDependencies` only runs when a test Dolt server is available.

If you'd like a deterministic regression test for the new branch, the cleanest seam is a fault-injecting store whose `IterDependentsWithMetadata` / `IterIssueComments` returns a non-nil error (or an iterator whose `Err()` is non-nil after the loop), asserting `bd show --include-dependents` now exits non-zero — where it previously exited 0 with an empty list. Happy to follow up with that as a separate PR if it's wanted.

PR #4189, opened 2026-05-27
fix(dep): keep relates-to out of dependency tree

### LINKED ISSUE #3936: bug: `bd list` tree renderer treats `relates-to` as a parent edge; bidirectional `relates-to` silently drops issues from output
## Summary

The default tree-mode output of `bd list` treats `relates-to` dependency edges as if they were parent-child edges. A one-direction `A relates-to B` causes B to render nested under A. A bidirectional `relates-to` (A ↔ B) creates a cycle in the render graph that the renderer appears to detect and short-circuit by **emitting nothing for the involved issues** — while the footer count stays correct, so the data loss is silent.

In the real workspace where I first noticed this, `bd list` reported `Total: 40 issues` but only rendered ~22 rows. One bidirectional `relates-to` between two root epics was the sole trigger.

## Environment

- `bd version 1.0.4 (Homebrew)`, macOS Darwin 25.3.0
- Reproduced identically on `bd 1.0.3` and `bd 1.0.4` (byte-identical output across the upgrade)
- Embedded mode (default `bd init`)

## Minimal repro (2 issues, no children, no reparenting)

```bash
mkdir /tmp/bd-relates-bug && cd /tmp/bd-relates-bug
bd init --prefix=demo -q
bd create --title="Epic A" --description="x" --type=epic --priority=2 -q
bd create --title="Epic B" --description="x" --type=epic --priority=2 -q

# Capture the two IDs
A=$(bd list --flat --no-pager | grep -oE 'demo-[a-z0-9]+' | sed -n 1p)
B=$(bd list --flat --no-pager | grep -oE 'demo-[a-z0-9]+' | sed -n 2p)
```

### Step 1: baseline — both issues render correctly

```
$ bd list --no-pager
○ demo-i7a ● P2 [epic] Epic B
○ demo-ysz ● P2 [epic] Epic A

--------------------------------------------------------------------------------
Total: 2 issues (2 open, 0 in progress)
```

### Step 2: add one-direction `relates-to` — second issue renders as a *child* of the first (incorrect)

```
$ bd dep add "$A" "$B" --type=relates-to -q
$ bd list --no-pager
○ demo-ysz ● P2 [epic] Epic A
└── ○ demo-i7a ● P2 [epic] Epic B

--------------------------------------------------------------------------------
Total: 2 issues (2 open, 0 in progress)
```

`relates-to` is not a hierarchical relationship and should not introduce nesting.

### Step 3: add the reverse direction (bidirectional `relates-to`) — both issues silently dropped

```
$ bd dep add "$B" "$A" --type=relates-to -q
$ bd list --no-pager

--------------------------------------------------------------------------------
Total: 2 issues (2 open, 0 in progress)
```

Zero rows rendered. Footer still claims 2 issues. No warning, no error.

## Real-world impact

In a workspace with 40 open issues across 4 root epics, where exactly one pair of epics is connected by a bidirectional `relates-to`, `bd list` rendered ~22 rows out of 40 and dropped the entire subtrees rooted at the two related epics (one epic + 5 children, the other + ~10 children). The bug is easy to miss because:

- The footer count is correct.
- The visible output is still well-formed (no error, no warning).
- A user might assume the "missing" issues simply don't exist.

## Diagnosis (hypothesis)

The tree renderer appears to walk `relates-to` edges as if they were parent edges when building the display hierarchy. Step 2 shows this directly: the renderer nests B under A purely because of the `relates-to`. In Step 3, the bidirectional edge produces mutual parent edges, the renderer (correctly) detects the cycle, and (incorrectly) skips emitting the involved nodes entirely instead of breaking the cycle and rendering both as roots.

## Expected behavior

`relates-to` is not a parent/child relationship and should not affect tree nesting. Suggested options (in order of simplicity):

1. **Ignore `relates-to` in the tree renderer entirely.** Show it only in `bd show <id>` output, where it already renders correctly under a `RELATED` section. This is the cleanest fix.
2. **Render related issues as roots with an inline annotation,** e.g., `↔ relates-to: demo-...`, rather than introducing nesting.

Either way: a cycle anywhere in the render graph should never result in silent omission. If the renderer can't decide on a layout it should fall back to flat output and still emit every row counted in the footer.

## Affected commands

- `bd list` (tree mode is default)
- `bd list --no-parent`
- `bd list --type=epic`
- `bd list --pretty`

## Unaffected (workarounds)

- `bd list --flat` — all rows render
- `bd list --parent=<id>` — works per-subtree
- `bd show <id>` — renders fully and shows `relates-to` correctly under a `RELATED` section
- `bd ready` — unaffected

## Severity

Silent data loss in the primary "show me my work" command, triggered by a single normal `bd dep add … --type=relates-to` (the auto-mirror behavior of `bd dep` makes bidirectional relates the default once you add one). The footer count being correct makes it easy to miss.

maphew:
## Why

`relates-to` is a loose graph link, not a hierarchy edge. Beads has two
separate tree builders that both treat it as if it were one, and the
combined effect makes `bd list` lose entire subtrees silently:

- `bd dep tree` (storage helper `GetDependencyTreeInTx`) walked
  `relates-to` edges as tree edges, so related issues showed up nested
  under their target with traversal cycle-detection further hiding
  endpoints of bidirectional relates-to.
- `bd list` (`buildIssueTreeWithDeps`) treats *any* dependency whose
  target is an epic as parent-child. A `relates-to` between two epics
  nests one under the other; a **bidirectional** relates-to marks both
  endpoints as children of each other, so neither lands in the root
  set and both subtrees silently drop from output while the footer
  count stays correct. This is the symptom #3936 reports.

This PR fixes both code paths so `relates-to` keeps showing up under
`bd dep list` (where it belongs) but never shapes tree output.

Fixes #3936.

## What changed

- `internal/storage/issueops/dependency_tree.go`: skip `relates-to`
  during dependency-tree traversal; add `isDependencyTreeEdge` for the
  predicate.
- `cmd/bd/list_tree.go`: skip `relates-to` in
  `buildIssueTreeWithDeps` so the epic-target heuristic does not turn
  it into a parent edge.
- Add issueops unit coverage for the tree-edge filter and traversal
  behavior.
- Add `cmd/bd` unit coverage for `buildIssueTreeWithDeps` (both
  one-direction and bidirectional relates-to between epics).
- Add embedded CLI regressions: bidirectional `relates-to` remains
  listed under `bd dep list` but is excluded from `bd dep tree` in
  both directions; `bd list --type=epic` keeps both epics visible and
  unnested.

## Validation

- `go test ./internal/storage/issueops -run 'Test(IsDependencyTreeEdge|GetDependencyTreeInTxSkipsRelatesToEdges)'`
- `go test ./internal/storage/issueops`
- `go test ./cmd/bd -run 'TestListBuildIssueTree'`
- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestEmbeddedDep/tree_ignores_bidirectional_relates_to'`
- `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestEmbeddedList/tree_relates_to_does_not_nest_or_drop_epics'`
- End-to-end `bd list` repro from #3936 against the built binary: both
  one-direction and bidirectional relates-to between two epics now
  keep both epics rendered as roots.

### DIFF
--- a/cmd/bd/dep_embedded_test.go
+++ b/cmd/bd/dep_embedded_test.go
@@ -362,6 +362,34 @@ func TestEmbeddedDep(t *testing.T) {
 		}
 	})
 
+	t.Run("tree_ignores_bidirectional_relates_to", func(t *testing.T) {
+		root := bdCreate(t, bd, dir, "Tree relates root", "--type", "task")
+		blocker := bdCreate(t, bd, dir, "Tree real blocker", "--type", "task")
+		related := bdCreate(t, bd, dir, "Tree loose relation", "--type", "task")
+
+		bdDep(t, bd, dir, "add", root.ID, blocker.ID, "--type", "blocks")
+		bdDep(t, bd, dir, "add", root.ID, related.ID, "--type", "relates-to")
+		bdDep(t, bd, dir, "add", related.ID, root.ID, "--type", "relates-to")
+
+		listOut := bdDep(t, bd, dir, "list", root.ID)
+		if !strings.Contains(listOut, related.ID) || !strings.Contains(listOut, "relates-to") {
+			t.Fatalf("expected relates-to edge in dep list output: %s", listOut)
+		}
+
+		treeOut := bdDep(t, bd, dir, "tree", root.ID)
+		if !strings.Contains(treeOut, root.ID) || !strings.Contains(treeOut, blocker.ID) {
+			t.Fatalf("expected root and real blocker in dep tree: %s", treeOut)
+		}
+		if strings.Contains(treeOut, related.ID) {
+			t.Fatalf("relates-to edge should not render as a dependency tree edge: %s", treeOut)
+		}
+
+		upOut := bdDep(t, bd, dir, "tree", related.ID, "--direction", "up")
+		if strings.Contains(upOut, root.ID) {
+			t.Fatalf("reverse relates-to edge should not render as a dependent tree edge: %s", upOut)
+		}
+	})
+
 	// ===== dep cycles =====
 
 	t.Run("cycles_detect", func(t *testing.T) {
--- a/cmd/bd/list_embedded_test.go
+++ b/cmd/bd/list_embedded_test.go
@@ -420,6 +420,30 @@ func TestEmbeddedList(t *testing.T) {
 		}
 	})
 
+	// Regression for gastownhall/beads#3936: relates-to between two epics
+	// must not nest them in `bd list` tree mode, and a bidirectional
+	// relates-to must not silently drop both epics from the output.
+	t.Run("tree_relates_to_does_not_nest_or_drop_epics", func(t *testing.T) {
+		epicA := bdCreate(t, bd, dir, "Relates Epic A", "--type", "epic", "--priority", "2")
+		epicB := bdCreate(t, bd, dir, "Relates Epic B", "--type", "epic", "--priority", "2")
+
+		bdDep(t, bd, dir, "add", epicA.ID, epicB.ID, "--type", "relates-to")
+		out := bdList(t, bd, dir, "--no-pager", "--type", "epic")
+		if !strings.Contains(out, epicA.ID) || !strings.Contains(out, epicB.ID) {
+			t.Fatalf("one-direction relates-to should keep both epics visible:\n%s", out)
+		}
+		if strings.Contains(out, "└── "+epicA.ID) || strings.Contains(out, "└── "+epicB.ID) ||
+			strings.Contains(out, "├── "+epicA.ID) || strings.Contains(out, "├── "+epicB.ID) {
+			t.Fatalf("relates-to must not nest epics under each other:\n%s", out)
+		}
+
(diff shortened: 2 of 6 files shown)


maphew:
Self-review caught that the original commit only fixed `bd dep tree` and not the user-facing `bd list` path #3936 was actually about. Pushed a follow-up commit (57e43c6) that:

- Filters `relates-to` in `cmd/bd/list_tree.go` `buildIssueTreeWithDeps` so the epic-target heuristic stops treating it as parent-child.
- Adds a `buildIssueTreeWithDeps` unit test that covers both one-direction nesting and bidirectional silent-drop (both halves fail without the fix, both pass with it).
- Adds an embedded-mode CLI regression that mirrors the exact #3936 repro under `bd list --type=epic`.

End-to-end check on the built binary: with both halves of the fix in place, the #3936 minimal repro now renders both epics as visible roots under both one-direction and bidirectional `relates-to`. Updated the PR body to describe both code paths.

Issue #3852, opened 2026-05-10
bd federation sync race: add-peer writes federation_peers without commit, blocking subsequent sync indefinitely


dlwiii:
## Versions

- `bd` 1.0.3 (1b2dd2cb)
- `dolt` 1.84.1
- Linux (Arch/EndeavourOS)

## Reproducer

1. Add a federation peer:

   ```
   bd federation add-peer beads http://beads:8080/dimension --user bd --password '<pw>'
   ```

2. Immediately try to sync:

   ```
   bd federation sync beads
   ```

3. Sync fails:

   ```
   ✗ merge failed: merge branch beads/main: Error 1105 (HY000): error: local changes would be stomped by merge:
       federation_peers
    Please commit your changes before you merge.
   ```

## What's happening

`add-peer` writes a row into the `federation_peers` table but does not `DOLT_COMMIT` it. The next `bd federation sync beads` call:

1. Triggers a `DOLT_FETCH` → `DOLT_MERGE` of `beads/main`.
2. Merge refuses because `federation_peers` has uncommitted changes.

## Documented workaround does not work

`FEDERATION-SETUP.md` suggests:

```
cd .beads/dolt && dolt add -A && dolt commit -m flush
bd federation sync beads
```

This fails because `bd federation sync` itself writes to `federation_peers` (presumably bumping \`last_sync\`) at the *start* of the call, re-dirtying the table before the merge runs. Verified via:

```
PORT=\$(cat .beads/dolt-server.port)
dolt --host 127.0.0.1 --port \$PORT --no-tls --user root --password '' --use-db dimension \\
  sql -q \"CALL DOLT_ADD('-A'); CALL DOLT_COMMIT('-m','flush peer','--allow-empty');\"
bd federation sync beads
```

The `DOLT_COMMIT` succeeds (returns a hash), then `bd federation sync` fails with the exact same \`local changes would be stomped: federation_peers\` error in the same shell.

## Side effects observed in workaround attempts

- Killing the bg `dolt sql-server` mid-state and running `dolt gc` cleaned up the chunk that held the just-added peer credentials, requiring `add-peer` to be re-run.
[...]

PR #3617, opened 2026-04-30
docs(agents): document storage-driver boundary in AGENTS.md

maphew:
## Summary

Adds a **Storage Boundary** section to `AGENTS.md` capturing the architectural direction that beads talks to storage through a driver interface (`dolthub/driver` for Dolt) and should not reach across that boundary. Concurrency, locking, and crash-safety are the driver's job; storage-implementation details should not appear on the public SDK surface.

The section also notes that storage issues are still welcome to be filed from the beads side — but should be flagged for routing to the driver rather than reflexively worked around in beads code (extra flocks, retries, schema poking).

## Why now

#3614 (removing the embedded-mode flock) is a concrete example of this direction in motion. The lifetime flock was beads reaching across the storage boundary to enforce an invariant the driver should own. Capturing the principle in `AGENTS.md` means future contributors and agents have the architectural context to avoid reintroducing the same kind of cross-boundary workaround by reflex when something goes sideways.

## Scope

- One file: `AGENTS.md` (+22 lines).
- No code changes, no behavior changes.
- Placed after the Visual Design Anti-Patterns section, before the tactical bd / shell guidance.

cc @coffeegoddd — would value your review since this codifies the storage-layer direction you've been driving. Happy to reword if any of the framing doesn't match how you'd describe the boundary.

## Test plan

- [x] `AGENTS.md` renders cleanly on GitHub
- [x] Reviewer confirms framing matches the intended driver-relationship roadmap

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3617"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

Issue #2474, opened 2026-03-09
bd dolt pull fails with 'cannot merge with uncommitted changes' and can corrupt journals in server mode


chozandrias76:
## Summary
In Beads v0.58.0 (server mode), `bd dolt pull` fails consistently with: `cannot merge with uncommitted changes` in two separate repos using the same server (127.0.0.1:3307). `bd list` and `bd dolt push` succeed. Repeated pull attempts can lead to Dolt journal corruption requiring `dolt fsck --revive-journal-with-data-loss`.

## Environment
- beads: 0.58.0
- dolt sql-server: 127.0.0.1:3307
- repos: /home/choza/projects (db beads_opencode-effects), /home/choza/projects/sc-interdiction (db beads_sc-interdiction)

## Repro
1. Ensure server mode and DBs are reachable (`bd list` succeeds).
2. Run `bd --dolt-auto-commit on dolt pull` in each repo.
3. Observe failure: `cannot merge with uncommitted changes`.
4. Inspect DB repo state with `dolt status`: `metadata` modified immediately after pull attempt.

## Observed
- `bd dolt pull`: fails with merge-dirty error.
- `bd dolt push`: succeeds.
- Server logs repeatedly show `error="nothing to commit"` followed by `error="cannot merge with uncommitted changes"` for same DB.
- After repeated attempts, some DB dirs hit corrupted journal and cannot be opened until fsck revive.

## Expected
`bd dolt pull` should auto-handle benign metadata churn (or avoid writing metadata before merge) and not leave working set dirty in a way that blocks pull.

## Notes
- This appears independent of repo content; reproduced in both DBs.
- Local tracking issue: opencode-effects-6wi (contains additional command transcript references).

chozandrias76:
Follow-up after upgrading to bd v0.59.0:\n\n- Fresh pull can succeed, but Pulling from Dolt remote...
Pull complete. still fails once metadata is dirty ().\n- Deterministic workaround in both repos: Committed.
Pulling from Dolt remote...
Pull complete.
Pushing to Dolt remote...
Push complete..\n- Verified DB integrity after tests with 
--------------- Chunk Type Summary ---------------
Found 1 chunks of type: RTVL
Found 1 chunks of type: DCMT
Found 1 chunks of type: WRST
Found 4 chunks of type: STRT
Found 4 chunks of type: ADRM
--------------- All Objects scanned. Starting commit validation ---------------
Found 1 commit objects
[...]

PR #4190, opened 2026-05-27
fix: commit federation metadata before sync

maphew:
Why:
#3852 reports that `bd federation add-peer` leaves `federation_peers` dirty, so a later sync can be blocked before it reaches fetch/merge. Current `Sync` also did not perform the pre-sync cleanup that `PullFrom` already does.

What changed:
- Commit `federation_peers` after add-peer writes peer metadata and remote setup.
- Auto-commit pending non-config changes before `Sync` fetch/merge, matching `PullFrom` behavior.
- Add regression coverage for committed add-peer metadata and pre-sync cleanup.

Drift check:
- #3852 is still open on GitHub as of 2026-05-27.

Validation:
- `go test ./internal/storage/dolt -run '^$' -count=1`\n- `go test -tags=integration ./internal/storage/dolt -run TestFederationSyncCommitsPendingPeerMetadataBeforeFetch -count=1 -v` skipped here because Docker/Dolt test server is unavailable\n- `git diff --check`\n\nRefs #3852\n\n_codex-gpt-5.5-medium on behalf of matt wilkie_

### DIFF
--- a/internal/storage/dolt/federation.go
+++ b/internal/storage/dolt/federation.go
@@ -311,6 +311,17 @@ func (s *DoltStore) Sync(ctx context.Context, peer string, strategy string) (*Sy
 		StartTime: time.Now(),
 	}
 
+	// Match PullFrom: federation metadata writes such as add-peer must not
+	// leave the working set dirty before fetch/merge.
+	if !s.readOnly {
+		if err := s.Commit(ctx, "auto-commit before sync"); err != nil {
+			if !isDoltNothingToCommit(err) {
+				result.Error = fmt.Errorf("failed to commit pending changes before sync: %w", err)
+				return result, result.Error
+			}
+		}
+	}
+
 	// Step 1: Fetch from peer
 	if err := s.Fetch(ctx, peer); err != nil {
 		result.Error = fmt.Errorf("fetch failed: %w", err)
--- a/internal/storage/dolt/credentials.go
+++ b/internal/storage/dolt/credentials.go
@@ -297,12 +297,16 @@ func (s *DoltStore) AddFederationPeer(ctx context.Context, peer *storage.Federat
 		return fmt.Errorf("failed to add federation peer: %w", err)
 	}
 
-	// Also add the Dolt remote
+	// Also add the Dolt remote.
 	if err := s.AddRemote(ctx, peer.Name, peer.RemoteURL); err != nil {
 		// Ignore "remote already exists" errors
 		if !strings.Contains(err.Error(), "already exists") {
 			return fmt.Errorf("failed to add dolt remote: %w", err)
 		}
+	}
+
+	if err := s.doltAddAndCommit(ctx, []string{"federation_peers"}, "federation: add peer "+peer.Name); err != nil {
+		return fmt.Errorf("failed to commit federation peer: %w", err)
 	}
 
 	return nil
--- a/internal/storage/dolt/federation_test.go
+++ b/internal/storage/dolt/federation_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/types"
 )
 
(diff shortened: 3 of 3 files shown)


maphew (commented):
## Review summary: ready-to-merge

Small, focused fix for the P0 in #3852. Lives entirely inside `internal/storage/dolt/` — the right side of the storage-driver boundary (#3617) — and mirrors the existing `PullFrom` pre-commit pattern (`internal/storage/dolt/federation.go:57-66`) rather than inventing new machinery. Regression test asserts the right property by probing `dolt_status` directly. +69/-1 across 3 files.

### Strengths

- `internal/storage/dolt/credentials.go:308-311` — `doltAddAndCommit(ctx, []string{"federation_peers"}, ...)` is scoped to the single table, so add-peer can't sweep unrelated dirty work into a "federation: add peer X" commit (the #2455 hygiene pattern).
- `Sync` pre-commit at `internal/storage/dolt/federation.go:294-303` mirrors `PullFrom` exactly (`readOnly` guard, `isDoltNothingToCommit` short-circuit, wrapped error). Trivially auditable.
- `Commit` already excludes `config`, so the pre-merge commit inherits that safety.
- Test is `//go:build integration`-gated, which is correct — Codecov's 0% patch coverage comment is therefore noise.

### Concerns (none blocking)

1. `internal/storage/dolt/federation.go:294-303` — `s.Commit(ctx, "auto-commit before sync")` stages every dirty non-`config` table, so unrelated issue/dependency churn can end up in a commit titled `auto-commit before sync`. Pre-existing in `PullFrom`; out of scope for a P0 fix. Worth a follow-up bead to consider scoping the pre-sync commit to federation/metadata tables.
2. Before merge, confirm CI ran with the `integration` tag (the PR description notes local skip due to no Docker/Dolt). If CI didn't exercise it, run `go test -tags=integration ./internal/storage/dolt -run TestFederationSyncCommitsPendingPeerMetadataBeforeFetch -count=1 -v` against a local Dolt server.

### Follow-ups (separate beads, don't block)

- Scrub `FEDERATION-SETUP.md` for the obsolete `dolt add -A && dolt commit -m flush` workaround that this fix makes unnecessary.
- File dlwiii's secondary report: running `dolt gc` against a bd-managed Dolt directory while bd was off appears to have eaten two issues. Whether the resolution is docs, a guard, or a driver-side fix, it shouldn't get lost.

---

maphew:
@coffeegoddd - over to you. The PR reaches into dolt internal storage so I marked as draft so my agents don't automatically see the green checks as 'accepted & complete' and merge it.

internal/storage/dolt/federation.go
```diff
@@ -291,6 +291,17 @@ func (s *DoltStore) Sync(ctx context.Context, peer string, strategy string) (*Sy
...
 
+	// Match PullFrom: federation metadata writes such as add-peer must not
+	// leave the working set dirty before fetch/merge.
+	if !s.readOnly {
+		if err := s.Commit(ctx, "auto-commit before sync"); err != nil {
+			if !isDoltNothingToCommit(err) {
+				result.Error = fmt.Errorf("failed to commit pending changes before sync: %w", err)
+				return result, result.Error
+			}
+		}
```
maphew, line 322: Second-opinion review: the same dirty-working-set bug from (3852) exists in `internal/storage/embeddeddolt/` and is **not** fixed by this PR.

Asymmetry in the parallel storage backend:

1. `internal/storage/embeddeddolt/federation.go:110` — `EmbeddedDoltStore.AddFederationPeer` calls `AddFederationPeerInTx` + `AddRemoteIfNotExists` inside `withConn`, but never commits the `federation_peers` table afterward. Same root cause as the `DoltStore` fix here.
2. `internal/storage/embeddeddolt/federation.go:198` — `EmbeddedDoltStore.Sync` has no pre-commit step, even though `EmbeddedDoltStore.PullFrom` (`version_control.go:257-262`) already calls `CommitPending` before pulling. So `Sync` on the embedded backend can still fail with "cannot merge with uncommitted changes" after add-peer.

Failure scenario: user running `bd federation add-peer X && bd federation sync X` against an embedded-mode store hits the original (3852) symptom; the fix here only helps the server-mode `DoltStore`.

This is out of scope for a P0 if the immediate report is server-mode only, but it should at minimum get a follow-up bead so #3852 isn't closed prematurely. The regression test added here also only exercises `DoltStore`, so an embedded-mode parallel would catch this in CI.

Concur with maphew's existing review otherwise (scoping the pre-sync `Commit()` is the right next refinement; this PR is a focused mirror of the existing `PullFrom` pattern).

maphew:

maphew:
Rebased onto current `upstream/main` (the branch was ~285 commits behind and showing as CONFLICTING after the upstream history rewrite). The fix is unchanged in substance; the only rebase conflict was an unused `doltutil` import in `federation_test.go`, now dropped.

### Why this is still needed (re-verified on current `upstream/main`)

The wedge described in #3852 still reproduces. `Sync()` (`internal/storage/dolt/federation.go`) runs its own `Fetch` + `Merge` and **does not** go through `PullFrom()`, so it never hits the #2474 pre-pull auto-commit. Combined with `AddFederationPeer` (`internal/storage/dolt/credentials.go`) writing `federation_peers` without committing, the sequence `add-peer` → `sync` fails with `cannot merge with uncommitted changes`, and the dirty table keeps every subsequent sync wedged.

This change closes both halves:
- commit `federation_peers` immediately after the add-peer write, and
- auto-commit pending non-config changes in `Sync()` before fetch/merge, mirroring the existing `PullFrom()` behavior.

### Relationship to #4412

This is **complementary to, not competing with**, #4412 (pre-pull commit must include `config`). #4412 fixes the *config-table* wedge on the **PullFrom** path; this fixes the *federation_peers* wedge on the **Sync** path — different table, different code path. They don't touch the same functions, so there's no logical conflict; whichever lands second just needs a trivial rebase.

### Verification

Local preflight: `go build ./internal/storage/dolt/...`, `go vet -tags integration`, and gofmt all clean; package tests pass. The integration regression test (`TestFederationSyncCommitsPendingPeerMetadataBeforeFetch`) skips locally without a Dolt test server — relying on PR CI to exercise it.

Issue #3927, opened 2026-05-13
[bug] bd ignores config.yaml in .beads dir — 'no beads configuration found' + 'issue_prefix config is missing'


vbtcl:
## Summary

After removing a stale `.beads/redirect` from the town root, `bd create` from the mayor context fails with:
```
warning: no beads configuration found in /home/vamshi/dev/gt/.beads; using default database name "beads"
Error: database not initialized: issue_prefix config is missing (run 'bd init --prefix <prefix>' for a new project, or 'bd bootstrap' to clone an existing remote)
```

## The config IS present

`/home/vamshi/dev/gt/.beads/config.yaml` exists and contains:
```yaml
prefix: hq
issue-prefix: hq
sync.mode: dolt-native
routing.mode: "explicit"
```

The Dolt `hq` database is healthy (1.1 GB, query latency 0s). `bd show hq-*` and `bd show pa-*` still work via `routes.jsonl`. Only `bd create` (and HQ-targeted writes) fail.

## Suspected cause

`uv` was upgraded from `v0.7.19` → `v0.11.14` during this session, which may have pulled a newer `bd` binary. The new version may expect `issue_prefix` (underscore) instead of `issue-prefix` (hyphen) as the config key, or has otherwise changed how it discovers/parses `config.yaml`.

Secondary possibility: `bd` is no longer treating a directory that has both `config.yaml` and `routing.mode: "explicit"` as a valid root — it may require a `sync.mode`-specific init step.

## Repro

1. Have a `.beads/` directory with `config.yaml` containing `issue-prefix: hq` and `sync.mode: dolt-native`
2. Run `bd create "test issue"` from a directory that discovers this `.beads/`
3. Observe warning + "issue_prefix config is missing" error despite config being present

## Additional context

- `bd show pa-*` still works — routing via `routes.jsonl` is unaffected
- `bd ready`, `bd list` work fine
- Only write operations (`bd create`, `bd comment`, `bd update`) fail
- Workaround: run `bd create` from within a subdirectory that has its own `.beads/` (e.g. `partcl/crew/boys`)

Related: gastownhall/gastown#3971 (the `.beads/redirect` recreation bug that triggered this)

PR #4191, opened 2026-05-27
debug config discovery and issue prefix failures

maphew:
Why:
#3927 reports a confusing state where `.beads/config.yaml` exists but writes fail with missing configuration and `issue_prefix` errors. Current diagnostics do not explain whether config discovery failed, whether metadata/config.yaml were found, or whether the user used the YAML key spelling that maps to the DB prefix.

What changed:
- Add debug-only config discovery logging when config loading returns no config.
- Add debug-only `issue_prefix` lookup diagnostics including canonical `issue-prefix` and legacy `issue_prefix` YAML visibility.
- Improve the missing-prefix error hint to call out `issue-prefix`, not `issue_prefix`, for `config.yaml`.
- Add focused tests for prefix trimming/error guidance and YAML key naming behavior.

Drift check:
- #3927 is still open on GitHub as of 2026-05-27.

Validation:
- `go test ./internal/storage/issueops -run TestReadConfigPrefix -count=1`
- `go test ./internal/config -run TestIssuePrefixKeyNamingInYAML -count=1`
- `git diff --check`

Refs #3927

### DIFF
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -965,6 +965,12 @@ var rootCmd = &cobra.Command{
 			doltCfg.ServerPassword = cfg.GetDoltServerPasswordForPort(doltCfg.ServerPort)
 			doltCfg.ServerTLS = cfg.GetDoltServerTLS()
 		} else if cfgErr == nil {
+			metadataPath := filepath.Join(beadsDir, configfile.ConfigFileName)
+			configYAMLPath := filepath.Join(beadsDir, "config.yaml")
+			_, metadataErr := os.Stat(metadataPath)
+			_, yamlErr := os.Stat(configYAMLPath)
+			debug.Logf("Debug: config discovery at %s -> metadata=%v (%v), config.yaml=%v (%v)\n",
+				beadsDir, metadataErr == nil, metadataErr, yamlErr == nil, yamlErr)
 			// Load returned (nil, nil) — no config file found.
 			// Fall back to the canonical default database name; matches the
 			// behavior of newDoltStoreFromConfig / newReadOnlyStoreFromConfig
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -1690,3 +1690,33 @@ func TestInitialize_ExternalBEADSDirDoesNotMergeCallerProjectConfig(t *testing.T
 		t.Fatalf("GetBool(json) = %v, want false", got)
 	}
 }
+
+func TestIssuePrefixKeyNamingInYAML(t *testing.T) {
+	restore := envSnapshot(t)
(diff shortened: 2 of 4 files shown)


maphew (commented):
## Review summary: ready-to-merge

Targeted diagnostic patch for #3927. Opt-in `BD_DEBUG` logging on two failure paths plus an actionable hint on the existing error. No behavior change for users who don't set `BD_DEBUG=1`. The only user-visible change in normal mode is the appended hint, which still satisfies the existing substring assertion at `internal/storage/embeddeddolt/create_issue_test.go:262`. +126/-1, 4 files.

### Strengths

- Solves the stated problem in #3927 without behavior change. The most likely user error (YAML `issue-prefix` vs. DB `issue_prefix`) now gets a direct hint in the error itself.
- Conservative scope: no retry logic, no new public surface, no migration. Hint is appended rather than replacing the existing error string.
- Tests target the new behavior — the `containsAll` helper locks in three specific substrings, so a regression fails loudly.
- No storage-driver-boundary violation. `internal/debug` is a leaf package; `internal/storage/issueops` already imports `internal/config` (see `helpers.go:17`), so this PR introduces no new cycle or boundary crossing.
- Idiomatic for the codebase. `debug.Logf` is the standard mechanism used in `cmd/bd/*`, `internal/config`, `internal/molecules`, `internal/jira`.

### Concerns (none blocking)

1. `internal/config/config_test.go:1693-1722` — `TestIssuePrefixKeyNamingInYAML` is really a `spf13/viper`-behavior pin, not a beads-behavior test. Fine to keep as a regression pin (the diagnostics in `ReadConfigPrefix` read both keys, so if viper ever started conflating them, the diagnostics would be silently wrong), but the test name and a one-line comment should make that intent explicit. Suggested: `TestViper_IssuePrefixKeysAreDistinct`, or add `// Locks in that viper treats issue-prefix and issue_prefix as separate keys, which the diagnostics in ReadConfigPrefix rely on.`
2. `cmd/bd/main.go:967` — diagnostic only fires on the `(nil, nil)` branch. Issue #3927's repro matches that branch, so this is the right place. But if a future #3927-shaped report involves `configfile.Load` returning a non-nil `cfg` with a missing/blank database name, the diagnostic won't fire. Worth a follow-up bead to consider mirroring the stat-and-log when `cfg != nil` but the resolved database falls through to `configfile.DefaultDoltDatabase`.
3. Codecov 45% patch coverage will look like a red flag at a glance — the missing lines are exactly the `debug.Logf` branch (6 lines) which is opt-in and not reachable from unit tests without rigging `BD_DEBUG=1`. Wave through at merge.

---

Issue #3722, opened 2026-05-05
bd init corrupts symlinked CLAUDE.md (mode 120000 + multi-line target → unrecoverable on macOS)


yakshaver:
## Summary

`bd init` injects a BEADS INTEGRATION block into `CLAUDE.md`. When `CLAUDE.md` is a symlink (the standard "single source of truth" pattern, typically `CLAUDE.md → AGENTS.md`), `bd init` reads the **symlink target string** instead of the symlink's contents, appends its block, and writes the result back into the git index at mode `120000` (symlink). The blob ends up as `"<old-target>\n\n<integration-block>"`.

On Linux this happens to work — the working tree gets a "symlink" whose target is the multi-line text, and most filesystems tolerate long targets. On macOS, APFS caps symlink targets at ~1023 bytes, so any operation that recreates the working-tree entry fails with:

```
error: unable to create symlink CLAUDE.md: File name too long
```

…which means `git rebase` / `git merge` / fresh clone / branch-switch are all broken until the index entry is repaired by hand.

Affected: beads 1.0.3 (Homebrew). Presumably any version that ships the current `bd init` injector.

## Reproducer

```bash
git init repro && cd repro
echo '# Agents' > AGENTS.md
ln -s AGENTS.md CLAUDE.md
git add AGENTS.md CLAUDE.md
git commit -m "Initial: symlinked CLAUDE.md"
git ls-files --stage CLAUDE.md
# 120000 <sha> 0    CLAUDE.md           ← symlink, blob = "AGENTS.md"

bd init                                  # injects BEADS INTEGRATION block
git ls-files --stage CLAUDE.md
# 120000 <sha> 0    CLAUDE.md           ← still symlink mode (BUG)
git cat-file -p :CLAUDE.md | head -3
# AGENTS.md
#
# <!-- BEGIN BEADS INTEGRATION ...      ← multi-line "symlink target"
git cat-file -s :CLAUDE.md
# 1654                                   ← exceeds APFS's ~1023-byte cap
```

On macOS, attempting to check the repo out fresh:

```bash
git clone . /tmp/repro2
# error: unable to create symlink CLAUDE.md: File name too long
```

## Real-world hit

[...]

PR #4192, opened 2026-05-27
fix: skip setup injection for symlinked agent files

maphew:
Why:
`bd setup claude` currently follows symlinked instruction files such as `CLAUDE.md -> AGENTS.md` and injects Beads managed text into the target file. That preserves the symlink path but mutates the linked target unexpectedly, which is the corruption mode reported in #3722.

What changed:
- Detect symlinked agent instruction paths with `Lstat` before managed-section injection.
- Skip injection for symlink paths and emit a warning instead of following the target.
- Tighten symlink safety coverage so the target content and symlink mode remain unchanged.

Drift check:
- #3722 is still open on GitHub as of 2026-05-27.

Validation:
- `go test ./cmd/bd/setup`
- `git diff --check`

Refs #3722

### DIFF
--- a/cmd/bd/setup/agents.go
+++ b/cmd/bd/setup/agents.go
@@ -95,7 +95,12 @@ func installAgents(env agentsEnv, integration agentsIntegration) error {
 	// can unexpectedly mutate other instruction files and, in some workflows,
 	// corrupt tracked symlink entries.
 	if info, err := os.Lstat(env.agentsPath); err == nil && info.Mode()&os.ModeSymlink != 0 {
-		_, _ = fmt.Fprintf(env.stdout, "Warning: %s is a symlink; skipping managed section injection to preserve link mode/content\n", agentsFile)
+		target, readErr := os.Readlink(env.agentsPath)
+		targetHint := ""
+		if readErr == nil && target != "" {
+			targetHint = fmt.Sprintf(" to %s", target)
+		}
+		_, _ = fmt.Fprintf(env.stderr, "Warning: %s is a symlink%s; skipping managed section injection to preserve link mode/content. Update the target file directly, or replace the symlink with a regular file and re-run '%s'.\n", agentsFile, targetHint, integration.setupCommand)
 		return nil
 	} else if err != nil && !os.IsNotExist(err) {
 		_, _ = fmt.Fprintf(env.stderr, "Error: failed to inspect %s: %v\n", env.agentsPath, err)
--- a/cmd/bd/setup/agents.go
+++ b/cmd/bd/setup/agents.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/templates/agents"
-	"github.com/steveyegge/beads/internal/utils"
 )
 
 // readFileBytesImpl is used in tests; avoids import cycle.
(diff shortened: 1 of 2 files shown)


maphew (commented):
## Review summary: merge-fix (lean merge with follow-ups)

Correctly stops the corruption mode in #3722 — `git index` mode `120000` entries with a multi-line blob that brick macOS clones. Narrow, safe `Lstat`-based refusal. No storage-driver-boundary concerns. +31/-17 across 2 files. The trade-off is that this implements *option 1* (refuse) from the upstream issue, while the codebase recently shipped *option 3* (write through symlinks) via `c7615e407 fix(setup): preserve symlinks in atomicWriteFile`. For users with the canonical `CLAUDE.md -> AGENTS.md` setup, `bd setup claude` now silently no-ops where it previously wrote integration text into the real `AGENTS.md`.

Recommendation: merge for v1.0.4 to unblock the P0, with one cheap pre-merge polish and follow-ups for the rest.

### Strengths

- Directly closes the corruption path. Refusal-with-warning is the safest possible behavior; it cannot break a working repo.
- Test rewritten in lockstep with the behavior change. New assertions match the new contract: link mode preserved (`Lstat`), target byte-equal to the seed, warning emitted.
- `installAgents` callers (`claude.go`, `mux.go`, `gemini.go`, `factory.go`, `opencode.go`) already treat its error as non-fatal, so even an unexpected `lstat` failure no longer aborts the higher-level integration.
- No `flock`-style locking, no engine introspection — purely in `cmd/bd/setup/`.

### Concerns

1. **Behavior regression vs. recently-shipped option 3.** `cmd/bd/setup/utils.go:23` and `cmd/bd/setup/utils_test.go:91 TestAtomicWriteFile_PreservesSymlink` document that the codebase was deliberately set up to write *through* symlinks via `ResolveForWrite`. The new warning doesn't name the resolved target or recommend the corrective command. Cheap fix that turns a silent regression into a directive — use `os.Readlink(env.agentsPath)` and extend the warning, e.g.:
   `Warning: CLAUDE.md is a symlink to AGENTS.md; skipping. Re-run setup against the target file (e.g. 'bd setup claude --agents-file AGENTS.md') to install integration there.`
2. **Sibling `atomicWriteFile` callers are unprotected.** `cmd/bd/setup/utils.go:23` is still called by `claude.go:55`, `codex.go:65`, `gemini.go:55`, `aider.go:165-175`, `junie.go:128/139`, `mux.go:109`, `cursor.go:65`, `agent_skill.go:40`. If a user has any of *those* artifacts as a symlink (e.g. `.claude/settings.json` -> shared file), they can still reproduce a similar corruption mode. This PR only locks down `installAgents`. Worth a follow-up bead to either audit the call sites or push the `Lstat` guard down into `atomicWriteFile` itself with an explicit opt-in for callers that genuinely want pass-through.
3. **Misleading post-skip messaging.** When the symlink branch returns early on line ~96, `claude.go:212-214` still unconditionally prints `Claude Code integration installed` and a restart hint. Consider returning a sentinel error (e.g. `errAgentsSymlinkSkipped`) that `claude.go`/`mux.go` can detect and adjust their final summary.
4. Warning is on `stdout` while sibling diagnostics in the same file use `env.stderr` (e.g. line 103-104: `fmt.Fprintf(env.stderr, "Error: failed to inspect ...")`). Test asserts against stdout, so it was deliberate — but a `Warning:` line on stdout mixes with success messages in piped/captured output. Small cleanup; update test along with it.
5. Codecov 75% patch coverage — the uncovered branch is almost certainly the `else if err != nil && !os.IsNotExist(err)` arm. One-line test (e.g. path traversing a non-directory) would lock it down.

### Recommendation

- **Before merge** (cheap polish; in-place maintainer edits since author is the same user): apply concern 1.
- **Follow-up beads** (don't block): concerns 2, 3, 5, plus a "repair recipe" snippet in release notes / docs for users already stuck with mode-`120000` multi-line `CLAUDE.md` blobs (the `git update-index --cacheinfo 100644,<sha>,CLAUDE.md` recipe from #3722).
- On merge: close #3722 and `mybd-l4r`.

---

Issue #3268, opened 2026-04-14
`bd ready` reports no ready work despite open issues with zero   dependencies


slagyr:
### Steps to reproduce

  1. Create an issue: `bd create --title="Test issue" --type=task --priority=2`
  2. Verify it's open: `bd list --status=open` — shows the issue
  3. Verify no dependencies: `bd dep list <id>` — reports "has no dependencies"
  4. Run `bd ready` — reports "No ready work found (all issues have blocking
  dependencies)"

  ### Expected

  `bd ready` shows the open issue as ready work.

  ### Actual

  `bd ready` reports no ready work, claiming all issues have blocking
  dependencies.

  ### Environment

  - bd 1.0.0 (Homebrew)
  - macOS Darwin 25.3.0 (arm64)
  - Custom status configured: `status.custom = unverified:wip`
  - `workflow.done-status = unverified`

  ### Notes

  - `bd list --status=open` correctly shows the issues
  - `bd blocked` correctly reports no blocked issues
  - The bug is intermittent — sometimes `bd ready` finds the issue, sometimes
  it doesn't
  - Multiple open issues with zero dependencies all invisible to `bd ready`
  simultaneously
  - Occurred on both 0.62.0 and 1.0.0
  - Removing `status.custom` and `workflow.done-status` config did not fix it
  (tested and restored)

PR #4194, opened 2026-05-27
fix(close): support per-id reasons

### LINKED ISSUE #3279: bd close: multi-ID --reason applies last reason to all issues
## bug

when closing multiple issues in one command with per-issue `--reason` flags, bd applies the last `--reason` to all issues instead of matching each reason to its preceding ID.

### reproduction

```bash
bd close issue-aaa --reason='reason A' issue-bbb --reason='reason B' issue-ccc --reason='reason C'
```

**expected:** each issue gets its own reason (aaa=A, bbb=B, ccc=C)
**actual:** all three get reason C

```
Closed issue-aaa: reason C
Closed issue-bbb: reason C
Closed issue-ccc: reason C
```

### impact

close reasons are important for searchability and context recovery. when batch-closing, users lose the per-issue close context. workaround is to close one at a time.

### likely cause

cobra's flag parsing collects all `--reason` flags into a single string value (last-wins) rather than associating each `--reason` with the preceding positional ID. supporting per-ID reasons would need a different parsing approach - either paired args or a different flag pattern.

maphew:
Why:
- #3279 reports that repeated `bd close --reason` values do not map cleanly to multi-ID closes.
- #3268 is still open upstream, but current main did not reproduce the zero-dependency ready bug; this adds a regression guard for that ready path.

What:
- Allow one shared close reason or one repeated `--reason` per issue ID.
- Reject mismatched reason/ID counts.
- Add embedded CLI regressions for per-ID close reasons and ready zero-dependency output.

Validation:
- `CGO_ENABLED=0 go test ./cmd/bd -tags gms_pure_go -run 'TestCloseResolveReasonFile|TestCloseResolveReasons' -count=1`
- `BEADS_TEST_EMBEDDED_DOLT=1 go test ./cmd/bd -tags gms_pure_go -run '^(TestEmbeddedClose|TestEmbeddedReady)$/^(close_multiple_ids|close_multiple_ids_with_per_id_reasons|ready_includes_open_issue_with_zero_dependencies)$' -count=1`
- `go test ./internal/storage/dolt -run 'Test.*Ready|Test.*Blocked|TestIsBlocked' -count=1`
- `go test ./internal/storage/issueops -run 'Test.*Ready|Test.*Blocked|Test.*Close' -count=1`
- `git diff --check`

Refs #3268
Fixes #3279
Bead: mybd-r1cb

### DIFF
--- a/cmd/bd/close.go
+++ b/cmd/bd/close.go
@@ -27,7 +27,9 @@ If no issue ID is provided, closes the last touched issue (from most recent
 create, update, show, or close operation).
 
 When closing multiple issues, provide one --reason for all IDs or repeat
---reason once per ID in the same order.`,
+--reason once per ID. Reasons map positionally: the first --reason applies
+to the first ID, the second --reason to the second ID, regardless of where
+the flags appear in the command line.`,
 	Args: cobra.MinimumNArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		CheckReadonly("close")
@@ -369,7 +371,7 @@ func resolveCloseReasons(cmd *cobra.Command, args []string) ([]string, []string,
 func collectCloseReasonFlags(cmd *cobra.Command) ([]string, error) {
 	if flag := cmd.Flags().Lookup("reason"); flag != nil {
 		if v, ok := flag.Value.(interface{ Values() []string }); ok {
-			if reasons := v.Values(); len(reasons) > 0 {
+			if reasons := nonEmptyCloseReasons(v.Values()); len(reasons) > 0 {
 				return reasons, nil
 			}
 		}
@@ -385,6 +387,16 @@ func collectCloseReasonFlags(cmd *cobra.Command) ([]string, error) {
 		}
 	}
 	return nil, nil
+}
+
+func nonEmptyCloseReasons(reasons []string) []string {
+	out := make([]string, 0, len(reasons))
+	for _, reason := range reasons {
+		if reason != "" {
+			out = append(out, reason)
+		}
+	}
+	return out
 }
 
 func reasonForCloseIndex(reasons []string, i int) string {
--- a/cmd/bd/close.go
+++ b/cmd/bd/close.go
@@ -24,7 +24,10 @@ var closeCmd = &cobra.Command{
 	Long: `Close one or more issues.
 
 If no issue ID is provided, closes the last touched issue (from most recent
-create, update, show, or close operation).`,
+create, update, show, or close operation).
+
+When closing multiple issues, provide one --reason for all IDs or repeat
+--reason once per ID in the same order.`,
 	Args: cobra.MinimumNArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		CheckReadonly("close")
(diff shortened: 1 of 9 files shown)


cmd/bd/close.go
```diff
@@ -339,6 +308,110 @@ func init() {
...
+	}
+
+	// Desire-path: "bd done <id> <message>" treats last positional arg as reason
+	// when no reason flag was explicitly provided (hq-pe8ce)
+	if len(reasons) == 0 && cmd.CalledAs() == "done" && len(args) >= 2 {
+		reasons = []string{args[len(args)-1]}
+		args = args[:len(args)-1]
+	}
+
+	if len(reasons) == 0 {
```
maphew, line 362: Behavior change for explicit `--reason ""`:

Old code used `cmd.Flags().GetString("reason")` and treated `""` as unset, falling through to the `"Closed"` default. The new `closeReasonFlagValue.Set("")` appends an empty string to `values`, so `collectCloseReasonFlags` returns `[""]` (len 1), the `len(reasons) == 0` default-fill on line 360 is skipped, and the empty string is used as the close reason.

If `validation.on-close` is set to `error`, this now errors out instead of writing `"Closed"`; if validation is off, an empty `close_reason` is silently persisted. Worth either filtering empties in `Set`/`Values()` or re-checking `reasons[0] == ""` before the default-fill so callers that explicitly pass `-r ""` keep their pre-PR behavior.

cmd/bd/close.go
```diff
@@ -339,6 +308,110 @@ func init() {
...
+	copy(out, v.values)
+	return out
+}
+
+func resolveCloseReasons(cmd *cobra.Command, args []string) ([]string, []string, error) {
+	reasons, err := collectCloseReasonFlags(cmd)
+	if err != nil {
+		return nil, args, err
+	}
+
```
maphew, line 348: The `hasExistingReason` change tightens the conflict check.

Previously the helper saw the raw inline reason string and only refused `--reason-file` when it was non-empty. Now any flag occurrence (including `--reason ""`) makes `len(reasons) > 0`, so `bd close --reason "" --reason-file f.txt` errors with the "cannot specify both" message even though the user supplied no usable inline reason.

Probably fine in practice, but if you want to preserve the prior empty-string allowance for `--reason-file` to win, filter empties out of `reasons` before this call (same fix as for the default-fill behavior change).

cmd/bd/close.go
```diff
@@ -339,6 +308,110 @@ func init() {
...
+			return nil, err
+		}
+		if reason != "" {
+			return []string{reason}, nil
+		}
+	}
+	return nil, nil
+}
+
+func reasonForCloseIndex(reasons []string, i int) string {
```
maphew, line 402: Docs ("repeat --reason once per ID in the same order") imply per-ID adjacency, but cobra/pflag parses positional args and flag occurrences in two independent streams and the code zips them strictly by index.

So `bd close --reason A --reason B id2 id1` maps id2 → A, id1 → B (likely the opposite of what the user expected), and `bd close id1 --reason B id2 --reason A` does the same. No error is raised because counts match. Worth either (a) calling out in the docs that mapping is "first --reason → first ID positionally" or (b) actually capturing argv order during parsing so adjacency is preserved.

maphew:
Addressed the review comments in `2e3be4626`:

- empty `--reason ""` occurrences are filtered before close reason resolution, restoring the previous fallback to `"Closed"`
- empty `--reason ""` no longer conflicts with `--reason-file`, so the file reason can win as before
- close help/docs now state that repeated reasons map positionally by ID/reason index, regardless of where flags appear in argv
- added focused resolver tests and regenerated CLI docs plus `llms-full.txt`

Verification:

```bash
CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run 'TestCloseResolveReasons|TestCloseResolveReasonFile'
./scripts/generate-cli-docs.sh --check
```

The default `go test ./cmd/bd ...` path still requires local ICU headers for `go-icu-regex`; reran under the pure-Go build tag used by the docs build.

PR #4195, opened 2026-05-27
test(mcp): make validation baseline hermetic

maphew:
Why:
- MCP validation had baseline failures from hard-coded local paths, shared Dolt test locations, and a missing `build` dependency.
- That made dependency PR checks noisy even when the dependency change was unrelated.

What:
- Replace user-specific absolute paths with repo-relative paths.
- Run BdClient integration tests in per-test isolated workspaces with explicit `BEADS_DIR`.
- Add `build` to dev dependencies and document the reliable baseline.
- Keep existing source `ruff`/`mypy` failures documented as non-gating maintenance checks.

Validation:
- `uv sync && uv run pytest && uv run python -m build`
- `uv run pytest tests/test_compaction_config.py tests/test_bd_client_integration.py`
- `uv run ruff check tests/test_compaction_config.py tests/test_bd_client_integration.py`
- Known existing: `uv run ruff check src/beads_mcp` and `uv run mypy src/beads_mcp` still fail on pre-existing source issues.

Bead: mybd-hbyb

### DIFF
--- a/integrations/beads-mcp/tests/test_bd_client_integration.py
+++ b/integrations/beads-mcp/tests/test_bd_client_integration.py
@@ -1,7 +1,8 @@
 """Real integration tests for BdClient using actual bd binary."""
 
+import os
 import shutil
-import tempfile
+import subprocess
 from pathlib import Path
 
 import pytest
@@ -32,16 +33,38 @@ def bd_executable():
     return bd_path
 
 
-@pytest.fixture
-def temp_workspace():
-    """Create a temporary workspace for an embedded Dolt beads database."""
-    with tempfile.TemporaryDirectory(prefix="beads_test_workspace_") as temp_dir:
-        yield temp_dir
+@pytest.fixture(scope="session")
+def bd_init_isolation_flags_supported(bd_executable):
+    """Skip integration tests when bd lacks the init flags these tests require."""
+    result = subprocess.run(
+        [bd_executable, "init", "--help"],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    help_text = f"{result.stdout}\n{result.stderr}"
+    required_flags = ("--non-interactive", "--skip-agents", "--skip-hooks")
+    missing_flags = [flag for flag in required_flags if flag not in help_text]
+    if missing_flags:
+        pytest.skip(
+            "bd init does not support required test isolation flags: "
+            + ", ".join(missing_flags)
+            + ". Install a current bd from this repository before running integration tests."
+        )
 
 
 @pytest.fixture
-async def bd_client(bd_executable, temp_workspace, monkeypatch):
-    """Create BdClient with an isolated workspace database."""
+def temp_workspace(tmp_path: Path) -> Path:
+    """Create a per-test workspace isolated with its own BEADS_DIR."""
+    workspace = tmp_path / "workspace"
+    workspace.mkdir()
+    return workspace
+
+
+@pytest.fixture
+async def bd_client(bd_executable, temp_workspace, monkeypatch, bd_init_isolation_flags_supported):
+    """Create BdClient with a temporary Dolt workspace - fully hermetic."""
+    beads_dir = temp_workspace / ".beads"
     monkeypatch.delenv("BEADS_DB", raising=False)
     monkeypatch.delenv("BEADS_DIR", raising=False)
     monkeypatch.delenv("BD_DB", raising=False)
@@ -49,10 +72,14 @@ async def bd_client(bd_executable, temp_workspace, monkeypatch):
 
     client = BdClient(
         bd_path=bd_executable,
-        beads_dir="",
+        beads_dir=str(beads_dir),
         beads_db="",
-        working_dir=temp_workspace,
+        working_dir=str(temp_workspace),
     )
+
+    env = os.environ.copy()
+    env["BEADS_DIR"] = str(beads_dir)
+    env["BD_NON_INTERACTIVE"] = "1"
 
     import asyncio
 
@@ -61,8 +88,12 @@ async def bd_client(bd_executable, temp_workspace, monkeypatch):
         "init",
         "--prefix",
         "test",
+        "--non-interactive",
+        "--skip-agents",
+        "--skip-hooks",
         stdout=asyncio.subprocess.PIPE,
         stderr=asyncio.subprocess.PIPE,
+        env=env,
         cwd=temp_workspace,
     )
     _stdout, stderr = await process.communicate()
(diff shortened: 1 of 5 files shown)


integrations/beads-mcp/tests/test_bd_client_integration.py
```diff
@@ -426,31 +419,32 @@ async def test_init_creates_beads_directory(bd_executable):
...
-        client = BdClient(bd_path=bd_executable, beads_db=None, working_dir=temp_dir)
+    # Create client WITHOUT beads_db set and WITH working_dir set to temp_dir
+    client = BdClient(bd_path=bd_executable, beads_db=None, working_dir=str(temp_path))
+    monkeypatch.setenv("BEADS_DIR", str(beads_dir))
 
-        # Initialize with custom prefix (no need to chdir!)
-        params = InitParams(prefix="test")
-        result = await client.init(params)
+    # Initialize with custom prefix (no need to chdir!)
+    params = InitParams(prefix="test")
```
maphew, line 473: The `monkeypatch.setenv("BEADS_DIR", str(beads_dir))` here runs **after** `BdClient(bd_path=..., beads_db=None, working_dir=str(temp_path))` is constructed. `BdClient.__init__` resolves `self.beads_dir` from the kwarg or `load_config()` once, at construction — it doesn't re-read the env later. The init call still happens to receive `BEADS_DIR` only because `BdClient._build_env` does `os.environ.copy()` and the monkeypatched value rides along that way. That's fragile and inverts the usual contract that `BdClient` carries its own routing. Either pass `beads_dir=str(beads_dir)` to the constructor, or move the `setenv` above the `BdClient(...)` line so the intent is explicit.

integrations/beads-mcp/tests/test_bd_client_integration.py
```diff
@@ -34,49 +33,45 @@ def bd_executable():
...
+    beads_dir = temp_workspace / ".beads"
+    client = BdClient(bd_path=bd_executable, beads_dir=str(beads_dir), working_dir=str(temp_workspace))
 
-    # Initialize database with explicit BEADS_DB - no chdir needed!
     env = os.environ.copy()
-    # Clear any existing BEADS_DB to ensure we use only temp_db
     env.pop("BEADS_DB", None)
-    env["BEADS_DB"] = temp_db
+    env["BEADS_DIR"] = str(beads_dir)
+    env["BD_NON_INTERACTIVE"] = "1"
```
maphew, line 83: The fixture pops `BEADS_DB` only from the `env` dict passed to this one `bd init` subprocess. Every subsequent `BdClient` call re-runs `os.environ.copy()` inside `_build_env` and will still inherit `BEADS_DB` from the developer's shell. `BdClient` then sets `BEADS_DIR` (because `self.beads_dir` is set), so the child process sees **both** `BEADS_DIR` and `BEADS_DB` — and bd's precedence between the two decides which database actually gets used. For the stated goal ("fully hermetic"), pop `BEADS_DB` once at the top of the fixture via `monkeypatch.delenv("BEADS_DB", raising=False)` (as the init test now does) so it can't leak into any later `BdClient` call in this fixture.

integrations/beads-mcp/tests/test_bd_client_integration.py
```diff
@@ -34,49 +33,45 @@ def bd_executable():
...
+        bd_executable,
+        "init",
+        "--prefix",
+        "test",
+        "--non-interactive",
+        "--skip-agents",
+        "--skip-hooks",
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
+        env=env,
```
maphew, line 96: `--non-interactive`, `--skip-agents`, and `--skip-hooks` are now baseline requirements for running these tests, but neither the README nor `pyproject.toml` pins a minimum bd version. A contributor whose `bd` predates any of these flags will see every integration test fail at fixture setup with a bd usage error (not a test assertion), which is exactly the noisy baseline this PR is trying to remove. Worth documenting the required bd version in the new "Baseline validation" section of the README, or detecting and skipping with a clear message if `bd init --help` lacks the flags.

integrations/beads-mcp/tests/test_compaction_config.py
```diff
@@ -63,9 +76,9 @@ def test_custom_compaction_threshold_via_env(self):
             env=env,
             capture_output=True,
             text=True,
-            cwd="/Users/stevey/src/dave/beads/integrations/beads-mcp"
+            cwd=PROJECT_ROOT,
         )
-        
+
         assert "threshold=50" in result.stdout or "threshold=20" in result.stdout  # May be cached
         # Due to module caching, we test the function directly instead
```
maphew: `assert "threshold=50" in result.stdout or "threshold=20" in result.stdout` is a tautology — both values are accepted, so this assertion can't actually fail. The trailing comment even admits the test doesn't validate anything ("Due to module caching, we test the function directly instead"). The PR is already editing this file to fix paths; since the test provides no signal, consider deleting it (the `_get_compaction_settings_with_custom_values` test below covers the same behavior properly). Leaving it in gives a false sense of coverage in the very baseline this PR is establishing as authoritative.

maphew:
Addressed the review feedback in `5671f61a7`:

- `test_init_creates_beads_directory` now passes `beads_dir` directly to `BdClient`, so routing is explicit without relying on ambient env after construction.
- The `bd_client` fixture removes shell `BEADS_DB` via `monkeypatch.delenv(...)` before later client calls can inherit it.
- Integration tests now check `bd init --help` for `--non-interactive`, `--skip-agents`, and `--skip-hooks`, then skip with a clear message when an older `bd` is installed.
- The MCP README documents that same current-`bd` requirement.
- Removed the compaction config test that accepted both threshold values and provided no assertion signal.

Validation:

```bash
uv run pytest tests/test_compaction_config.py tests/test_bd_client_integration.py
uv run ruff check tests/test_compaction_config.py tests/test_bd_client_integration.py
git diff --check
```

All passed locally.

maphew:
Rebased onto current main (`b7fa473`). The branch predated the repository history rewrite, so it had no common ancestor with main; the CONFLICTING state and the red Go-side checks (Upgrade smoke, Differential Regression, Test matrix, Build Embedded Dolt, duplicate migration 0044 panics) all came from CI merging the stale pre-rewrite history, not from this mcp-only diff. The PR is MERGEABLE again after the rebase.

What changed in the rebase:

- Squashed the original two commits into one and replayed it onto main.
- Composed with #4253 (tolerate current bd JSON shapes), keeping both behaviors: upstream's `MCP_ROOT` path fixes, extra env isolation (`BEADS_DIR`, `BD_DB`, `BEADS_WORKING_DIR`), and tolerant `embeddeddolt` assertions are retained; this PR still adds the `bd init` isolation-flag skip fixture, per-test `BEADS_DIR` workspaces, the `PYTHONPATH` subprocess helper, removal of the no-signal compaction test, the README baseline docs, and the `build` dev dependency.
- `uv.lock` was regenerated from upstream's lock with `uv lock` (adds `build` and `pyproject-hooks` only) instead of trusting the textual merge.

Local validation with a freshly built `bd` (`go build -tags gms_pure_go ./cmd/bd/`) on PATH:

```
uv run pytest tests/test_bd_client_integration.py tests/test_compaction_config.py
27 passed in 120.54s
uv run ruff check / ruff format --check: clean
```

__claude-unknown-model-high on behalf of matt wilkie__

PR #4198, opened 2026-05-27
fix(create): set status=deferred on future --defer (rebased #4112, GH#4071)

harry-miller-trimble:
Rebased, tested version of #4112 by @kevglynn, landed on top of current `main`.

### What this does
`bd create --defer <future-date>` now sets `status=deferred` instead of `open` (GH#4071). Original commit by @kevglynn is preserved; a second commit adds test coverage.

### Why a new PR
#4112's branch is ~50 commits behind `main`. Merging it as-is would also **break an existing test**: `TestBuildCreateIssueFromInput_PopulatesAllFields` hard-codes a future defer date and asserts `StatusOpen` — correct today only by accident, and broken the moment this behavior lands. That test file doesn't even exist on the stale branch, so the fix has to be applied on the `main` side.

### Changes on top of @kevglynn's commit
- `cmd/bd/create_defer_test.go` — unit test: future defer → deferred; past/nil → open.
- `cmd/bd/create_proxied_server_test.go` — switched the hard-coded defer date to a relative future time and updated the assertion to the now-correct `StatusDeferred` (de-flakes a latent date dependency).

Supersedes #4112.

### DIFF
--- a/cmd/bd/create.go
+++ b/cmd/bd/create.go
@@ -773,6 +773,11 @@ func buildCreateIssue(params createIssueParams) *types.Issue {
 		externalRefPtr = &params.ExternalRef
 	}
 
+	status := types.StatusOpen
+	if params.DeferUntil != nil && params.DeferUntil.After(time.Now()) {
+		status = types.StatusDeferred
+	}
+
 	return &types.Issue{
 		ID:                 params.ID,
 		Title:              params.Title,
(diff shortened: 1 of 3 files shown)


maphew:
Closing this fallback because the contributor PR #4112 was merged directly for the release batch.

The implementation from Kevin Glynn landed through #4112, preserving the contributor branch as the selected merge path. This fallback is no longer needed for the release queue.

PR #4202, opened 2026-05-27
debug missing dolt database metadata

maphew:
## Why

PR #4191 added useful debug output when config discovery returns no config file, but the adjacent loaded-config case was still opaque: `metadata.json` can exist while `dolt_database` is missing or blank, causing bd to fall back to the default `beads` database name without any debug trail.

That makes non-default database drift difficult to diagnose. This keeps normal behavior unchanged and adds only debug/verbose diagnostics.

## What

- Factor config discovery stat logging into a helper.
- Log when loaded metadata lacks `dolt_database` and no `BEADS_DOLT_SERVER_DATABASE` override is set.
- Add a focused test for the debug helper output.

## Tests

- `go test -tags gms_pure_go ./cmd/bd -run TestLogConfigDiscoveryIncludesMetadataAndYAMLState`
- `go test -tags gms_pure_go ./cmd/bd`

### DIFF
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -163,6 +163,10 @@ func logConfigDiscovery(beadsDir, reason string) {
 		reason, beadsDir, metadataErr == nil, metadataErr, yamlErr == nil, yamlErr)
 }
 
+func shouldLogDefaultDoltDatabase(cfg *configfile.Config) bool {
+	return cfg != nil && cfg.DoltDatabase == "" && os.Getenv("BEADS_DOLT_SERVER_DATABASE") == ""
+}
+
 // loadBeadsSelectionEnvFile loads only the selector keys needed for early
 // workspace/database discovery. Unlike loadBeadsEnvFile, this intentionally
 // limits itself to BEADS_DIR / BEADS_DB / BD_DB so caller credentials and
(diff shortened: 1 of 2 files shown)


cmd/bd/main.go
```diff
@@ -961,6 +970,9 @@ var rootCmd = &cobra.Command{
 			// Always set database name (needed for bootstrap to find
 			// prefix-based databases like "beads_hq"; see #1669)
 			doltCfg.Database = cfg.GetDoltDatabase()
+			if cfg.DoltDatabase == "" && os.Getenv("BEADS_DOLT_SERVER_DATABASE") == "" {
```
maphew: Nit (non-blocking): the new test exercises `logConfigDiscovery` directly, but the call-site guard added in this PR — `cfg.DoltDatabase == "" && os.Getenv("BEADS_DOLT_SERVER_DATABASE") == ""` — has no behavior coverage. A future refactor (e.g. flipping `&&` to `||`, or losing the env check) would silently change diagnostics without tripping CI.

Consider a small end-to-end test that:
1. Writes a `metadata.json` with `dolt_database` omitted into a temp `beadsDir`.
2. Invokes the rootCmd setup path (or extracts the conditional into a tiny helper) with `BEADS_DOLT_SERVER_DATABASE` unset.
3. Asserts the captured stderr contains `metadata loaded without dolt_database`.

Not a blocker — diagnostic-only change, default behavior unchanged. Happy to land as-is if you'd rather track the gap as a follow-up bd issue.

maphew: Addressed in 313c671: extracted the guard into `shouldLogDefaultDoltDatabase` and added table coverage for omitted metadata database, env override, explicit metadata database, and nil config.

Local check:
`CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run 'Test(LogConfigDiscoveryIncludesMetadataAndYAMLState|ShouldLogDefaultDoltDatabase)$'`

PR #4204, opened 2026-05-27
import: skip stale JSONL issue snapshots

### LINKED ISSUE #3822: Import/Export JSONL in server mode may result in data loss.

maphew:
## Summary

- skip imported JSONL issue records when the local issue has a newer `updated_at`
- report stale JSONL skips separately from same-file duplicate skips
- include `stale_skipped_ids` in `bd import --json` so callers can distinguish imported IDs from stale records that were referenced but not applied
- avoid creating an import commit when stale filtering leaves no issue or memory changes
- add regression coverage for stale import filtering and the existing server-mode auto-export guard
- regenerate CLI reference docs and `llms-full.txt` from the pure-Go binary used by CI

## Why

#3822 reports a data-loss class around JSONL import/export in server-backed workflows. Current `main` already skips server-mode auto import/export, but manual or recovery JSONL import can still replay an older snapshot over a newer local row because import unconditionally upserts matching issue IDs.

This keeps JSONL import from moving an existing issue backward in time: if the incoming row is older than the local row, it is skipped instead of clobbering the local title, status, priority, or other fields.

One residual concurrency caveat remains: the stale check currently reads local `updated_at` before the batch upsert, so another writer could still advance the row between that read and the write. This PR narrows the unconditional clobber case but does not make the stale guard fully transactional; making it airtight should happen inside the storage upsert path or a shared transaction.

Fixes #3822.

## Validation

- `go test -tags gms_pure_go ./cmd/bd -run 'TestFilterStaleImportIssuesSkipsOlderIncomingRecords|TestImportIssuesCoreReportsStaleSkippedIDs|TestMaybeAutoExportSkipsServerModeBeforeStoreAccess|TestShouldRunAutoImportJSONL' -count=1`
- `CGO_ENABLED=0 go build -tags gms_pure_go -o bd ./cmd/bd/`
- `./scripts/generate-cli-docs.sh ./bd`
- `./scripts/generate-llms-full.sh`
- `./scripts/check-doc-flags.sh ./bd`
- `./scripts/check-doc-freshness.sh`
- `git diff --check`

The cgo end-to-end import test was added but skipped locally because the Docker/Dolt test server is unavailable in this environment.

### DIFF
--- a/cmd/bd/import.go
+++ b/cmd/bd/import.go
@@ -65,7 +65,7 @@ EXAMPLES:
   cat issues.jsonl | bd import -   # Pipe JSONL from another tool
   bd import --dry-run              # Show what would be imported
   bd import --dedup                # Skip issues with duplicate titles
-  bd import --json                 # Structured output with created IDs`,
+  bd import --json                 # Structured output with created and skipped IDs`,
 	GroupID: "sync",
 	RunE:    runImport,
 }
@@ -142,6 +142,7 @@ type importResultJSON struct {
 	DedupHits           int      `json:"dedup_skipped,omitempty"`
 	Memories            int      `json:"memories,omitempty"`
 	IDs                 []string `json:"ids,omitempty"`
+	StaleSkippedIDs     []string `json:"stale_skipped_ids,omitempty"`
 	SkippedDependencies []string `json:"skipped_dependencies,omitempty"`
 	DryRun              bool     `json:"dry_run,omitempty"`
 }
@@ -250,6 +251,7 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 		result.Skipped += importResult.Skipped
 		result.SkippedDependencies = append(result.SkippedDependencies, importResult.SkippedDependencies...)
 		result.IDs = append(result.IDs, importResult.ImportedIDs...)
+		result.StaleSkippedIDs = append(result.StaleSkippedIDs, importResult.StaleSkippedIDs...)
 	}
 
 	if result.Created > 0 || result.Memories > 0 {
--- a/cmd/bd/import.go
+++ b/cmd/bd/import.go
@@ -249,19 +249,18 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 		result.Created = importResult.Created
 		result.Skipped += importResult.Skipped
 		result.SkippedDependencies = append(result.SkippedDependencies, importResult.SkippedDependencies...)
-		for _, issue := range issues {
-			result.IDs = append(result.IDs, issue.ID)
-		}
+		result.IDs = append(result.IDs, importResult.ImportedIDs...)
 	}
 
-	// Commit
-	commitMsg := fmt.Sprintf("bd import: %d issues", result.Created)
-	if result.Memories > 0 {
-		commitMsg += fmt.Sprintf(", %d memories", result.Memories)
-	}
-	commitMsg += fmt.Sprintf(" from %s", filepath.Base(source))
-	if err := store.Commit(ctx, commitMsg); err != nil {
-		return fmt.Errorf("commit: %w", err)
+	if result.Created > 0 || result.Memories > 0 {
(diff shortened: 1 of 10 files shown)


cmd/bd/import.go
```diff
@@ -249,19 +249,18 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error
 		result.Created = importResult.Created
 		result.Skipped += importResult.Skipped
 		result.SkippedDependencies = append(result.SkippedDependencies, importResult.SkippedDependencies...)
-		for _, issue := range issues {
-			result.IDs = append(result.IDs, issue.ID)
-		}
+		result.IDs = append(result.IDs, importResult.ImportedIDs...)
```
maphew, line 253: Silent behavior change in `--json` output: `result.IDs` previously contained every parsed (post-dedup) issue ID from the JSONL stream; after this PR it only contains IDs that survived the stale-filter (`importResult.ImportedIDs`). Any consumer that reads the JSON `ids` array to know "which IDs the import file referenced" will now under-report whenever stale rows are dropped. Not necessarily wrong — the new value is arguably more accurate — but it is an undocumented contract change. Worth either a note in the PR body / CHANGELOG, or keeping a separate `stale_skipped_ids` field so callers can still recover the full set.

cmd/bd/import_shared.go
```diff
@@ -50,9 +51,18 @@ func importIssuesCore(ctx context.Context, _ string, store storage.DoltStorage,
...
+	if err != nil {
+		return nil, err
+	}
+	issues = filtered
+	if len(issues) == 0 {
+		return &ImportResult{Skipped: staleSkipped}, nil
+	}
+
 	var skippedDependencies []string
 	skippedDependencySet := make(map[string]struct{})
```
maphew, line 65: Stale check is not atomic with the subsequent upsert. `filterStaleImportIssues` reads `updated_at` via `GetIssuesByIDs`, then `importIssuesCore` calls `CreateIssuesWithFullOptions` outside any shared transaction. In server-mode (the very mode upstream No.3822 highlights), another writer can advance the local row between the read and the write; the incoming snapshot, which was "newer" at read time but is now older than the freshly written local row, will still be upserted and clobber the newer local data.

This is a narrower failure window than the unconditional clobber the PR fixes, so it is not a regression — but if you want the guard to be airtight, the read + write should both live inside the same Dolt transaction (or at minimum re-check `updated_at` inside `CreateIssuesWithFullOptions`). A short note in the PR body acknowledging the residual TOCTOU window would also be enough for now.

maphew (commented):
Review of PR No.4204 ("import: skip stale JSONL issue snapshots") at medium effort.

The stale-filter is a sensible guard for the data-loss class in upstream issue No.3822 and the regression coverage is well targeted. Two precision findings posted inline; neither is a blocker. The core change LGTM modulo the inline notes.

cmd/bd/import.go
maphew: Addressed in `60fc7ac39`: `ids` remains the applied/imported ID list, and stale records now appear separately under `stale_skipped_ids` in the JSON result. The PR body and generated CLI docs also now describe skipped IDs instead of only created IDs.

cmd/bd/import_shared.go
maphew: Acknowledged in the PR body. I left the implementation as a pre-upsert guard for this PR because moving the check into the storage transaction/upsert path is a broader storage-boundary change. The body now calls out that this narrows the unconditional clobber but is not fully transactional yet.

PR #3265, opened 2026-04-14
Fix worktree-aware config.yaml resolution

osamu2001:
## Summary
- make shared-worktree config discovery consistent for startup loading, YAML-only config writes, config validation, and repo-level doctor/fix helpers
- route doctor/config checks and fixes through the resolved shared `.beads` directory instead of assuming a worktree-local `.beads`
- avoid materializing worktree-local `.beads` during normal server-backed opens by lazily initializing federation credential keys only when needed
- add shared-worktree regression coverage for loaded config reuse, `config.local.yaml` merges, doctor/fix fallback paths, `bd dolt remote add` persisting `sync.remote`, and remaining standalone `doctor` checks such as locks, hook migrations, MQ cleanup, and remote consistency

## Testing
- `go test ./internal/config -run 'TestFindProjectConfigYamlWithFinder_UsesLoadedWorktreeConfig|TestFindProjectConfigYamlWithFinder_UsesSharedConfigAfterLocalOverrideMerge|TestSetYamlConfig_WorktreeFallbackUsesMainRepoConfig|TestFindConfigYAMLPath_WorktreeFallbackUsesMainRepoConfig|TestInitialize_WorktreeFallbackUsesMainRepoConfig|TestInitialize_WorktreeFallbackMergesSharedLocalOverride' -count=1`
- `./scripts/test-cgo.sh -run '^TestDoltRemoteAddPersistsSyncRemoteToSharedWorktreeConfig$' ./cmd/bd`
- `./scripts/test-cgo.sh -run 'Test(ResolveBeadsDirForRepo_.*|CheckMetadataVersionTracking_BareParentWorktreeFallback|CheckDoltLocks_BareParentWorktreeFallback|CheckRemoteConsistency_WorktreeFallbackUsesSharedConfig|CheckAndFixStaleMQFiles_WorktreeFallback|CheckStaleLockFiles|DetectPendingMigrations_Hooks|DetectPendingMigrations_HooksBrokenMarkerIsCritical|DetectPendingMigrations_HooksWorktreeFallback|IntegrityChecks_NoBeadsDir|IntegrityChecks_EmptyBeadsDir|CheckDeletionsManifest_LegacyFile)$' ./cmd/bd/doctor`

## Notes
- `go test ./cmd/bd` requires the repo CGO wrapper on this machine because raw `go test ./cmd/bd` hits local ICU setup requirements.
[...]

PR #4213, opened 2026-05-28
test(fs): make domain/fs suite hermetic when run from a linked worktree

quad341:
## Summary

- `go test ./internal/storage/domain/fs/` run from any linked git worktree was silently clobbering the primary checkout's `.beads/metadata.json` with `{"dolt_database":"my_db","project_id":"pid-123"}` from the `readBeadsConfigParses` fixture (real incident: gm-q6biba).
- Root cause: `resolveBeadsDir` checks `GetWorktreeFallbackBeadsDir()` which calls `git.IsWorktree()` — a process-global `sync.Once` that caches `isWorktree=true` when the test binary runs from a linked worktree, so the intended `workDir/.beads` branch is never reached.
- Fix is test-only (no runtime behaviour changed): `SetupSuite` chdirs to a fresh `os.MkdirTemp` (not a git repo) and calls `git.ResetCaches()` before any test runs, seeding the cache with `isWorktree=false`. `TearDownSuite` restores the original dir and resets again.

## Test plan

- [x] `go test ./internal/storage/domain/fs/ -count=1` green from primary checkout
- [x] `go test ./internal/storage/domain/fs/ -count=1` green from linked worktree
- [x] Primary `.beads/metadata.json` sentinel unchanged after suite run from linked worktree (clobber eliminated)
- [x] Existing tests: no regressions (27 subtests all pass)

### DIFF
--- a/internal/storage/domain/fs/suite_test.go
+++ b/internal/storage/domain/fs/suite_test.go
@@ -2,16 +2,43 @@ package fs
 
 import (
 	"context"
+	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/suite"
 
+	"github.com/steveyegge/beads/internal/git"
 	"github.com/steveyegge/beads/internal/storage/domain"
 )
 
 type testSuite struct {
 	suite.Suite
+	origDir  string
+	suiteDir string
+}
+
+func (s *testSuite) SetupSuite() {
+	origDir, err := os.Getwd()
+	s.Require().NoError(err, "SetupSuite: get working directory")
+	s.origDir = origDir
+
+	suiteDir, err := os.MkdirTemp("", "beads-fs-suite-*")
+	s.Require().NoError(err, "SetupSuite: create temp dir")
+	s.suiteDir = suiteDir
+
+	s.Require().NoError(os.Chdir(suiteDir), "SetupSuite: chdir to non-git dir")
+	git.ResetCaches()
+}
+
+func (s *testSuite) TearDownSuite() {
+	if s.origDir != "" {
+		_ = os.Chdir(s.origDir)
+		git.ResetCaches()
+	}
+	if s.suiteDir != "" {
+		_ = os.RemoveAll(s.suiteDir)
+	}
 }
 
 func (s *testSuite) Ctx() context.Context {


maphew (commented):
Thanks for this - a 27-line, test-only patch that makes `internal/storage/domain/fs`'s suite hermetic by chdir-ing to a fresh non-git temp dir and calling `git.ResetCaches()` in `SetupSuite`/`TearDownSuite`. Cheap insurance against a real class of hazard (a process-wide `sync.Once` cache in `internal/git` picking up worktree state from wherever the test binary happens to start), and it only touches `suite_test.go` - no product code, no storage-boundary concerns.

Findings:

- [minor] I could not reproduce the clobber described in the PR body against current `upstream/main`. I built a scratch main-checkout + linked-worktree pair, seeded a sentinel `.beads/metadata.json` in the main checkout, and ran `go test ./internal/storage/domain/fs/ -run TestDomainFS -v` from the linked worktree without this patch applied - the sentinel was untouched. Tracing the call graph: `resolveBeadsDir` (`internal/storage/domain/fs/beads.go`) calls `beads.ResolveBeadsDirForRepo(workDir)`, which resolves worktree fallback via `worktreeFallbackBeadsDirForRepo` - a *repoPath*-scoped helper (`git -C repoPath ...`) added by #3265 (2026-04-24), not the process-cwd-cached `git.IsWorktree()`/`GetWorktreeFallbackBeadsDir()` the PR blames. Since `newRepo()` also always passes a fresh `t.TempDir()` (not a git repo) as `workDir`, the `git -C` call fails harmlessly and falls back to the local temp `.beads`. It's possible #3265 already closed this hole before this branch was cut (2026-05-28), or the original gm-q6biba incident hit a different path than the current `readBeadsConfigParses` fixture does today. Worth a quick sanity check on your end against latest `main` - but either way the fix is harmless defense-in-depth even if the original path is no longer live, so this isn't a reason to hold the PR.
- [nit] No `t.Parallel()` is used anywhere in this package, so the suite-wide chdir is safe for sequential execution as written; if a future subtest in this package adds parallelism this pattern would need revisiting, but that's speculative, not a blocker.

CI is all green (Build, Lint, Test x20 shards, Differential Regression, format/doc/version checks) and the PR is cleanly mergeable against current `main` - no history-rewrite conflict here.

Recommendation: merge. This is a safe, self-contained test-hygiene improvement; no changes requested. If you get a chance, a one-line confirmation of whether the clobber still reproduces on latest `main` would help close the loop on the root-cause story, but it's not needed to land this.

Thanks again for tracking down and fixing a genuinely annoying class of test non-determinism.

Issue #3254, opened 2026-04-13
Feature: Add `bd create --external-ref` to pre-link local issues to Linear


warmlogic:
## Summary

There is no way to pre-link a locally created Beads issue to an existing Linear issue. Creating the same issue in both systems independently produces duplicates on the next pull, since there is no title-matching or deduplication at the sync level.

## Version

Beads v1.0.0

## Problem

When a team creates issues in both Beads (for local planning) and Linear (from the UI), the next `bd linear sync --pull` imports the Linear issue as a new Beads issue — even if an identical one already exists locally. This creates duplicates that must be manually resolved.

## Workaround

Pick one creation path per issue — either create in Beads and push, or create in Linear and pull. Never create the same issue in both systems independently.

If duplicates do occur after pull, manually delete one copy with `bd delete <id>` and re-pull, or set the `external_ref` field directly in the Dolt database to link the local issue to the Linear one.

## Suggested fix

One or both of:

1. **`bd create --external-ref <linear-url>`** — allow pre-linking at creation time so Beads knows this local issue corresponds to an existing Linear issue
2. **Title-based deduplication on pull** — when pulling, if a Linear issue title matches an existing Beads issue title (exact or fuzzy), prompt for resolution instead of creating a duplicate

Option 1 is simpler and more explicit. Option 2 is more automatic but risks false positives on common titles.

## Related

See #3184 (milestones don't create epics on pull), #3185 (deps don't pull reliably) — both involve pull-side data fidelity.

maphew:
Drift check against current upstream `main`:

- `bd create` already reads `--external-ref` and passes it into issue creation in `cmd/bd/create.go`.
- The Linear docs already show `bd create "Fix login" --external-ref https://linear.app/team/issue/TEAM-123`; see merged #3517.
- Existing embedded/proxied create tests cover a Linear URL pre-link. I also added a fast local unit guard for the same create builder path.

This looks like option 1 from this issue is implemented. I do not see an open PR for #3254 or a remaining gap in create-time pre-linking. If title-based dedup remains desired, that seems like a separate follow-up from the explicit `--external-ref` request.

_codex-gpt-5.5-high on behalf of matt wilkie_

Issue #3186, opened 2026-04-11
FR: Push should write deps as Linear blocking relations


warmlogic:
## Problem

`bd linear sync --push` syncs title, description, status, priority, assignee, and estimates — but does not push `bd dep` relationships as Linear blocking relations.

**Real-world impact:** After push, the Linear project has no dependency information. Team members looking at Linear see tasks but not their blocking relationships. This also means dependencies don't survive a push-then-pull round-trip.

## Proposed Solution

When pushing an issue, if it has `bd dep` relationships and both the issue and its dependencies have `external_ref` pointing to Linear issues, create corresponding `blockedBy`/`blocks` relations in Linear.

This completes the push side of dependency sync:
- `bd dep A --blocks B` → Linear: A blocks B (and B is blockedBy A)

## Current Behavior

Push maps: title, description, status (via state_map), priority (via priority_map), assignee, estimate.
Push does NOT map: blocking relations, parent-child relationships.

maphew:
Thanks for the report. I’m doing a repro sweep and couldn’t find a concrete reproduction recipe in the current description or thread.

Could you add the smallest command sequence or setup that triggers this, including the `bd version`, storage mode, OS, and the expected vs. actual output? A scratch-repo repro is ideal if one is possible.

_codex-gpt-5.5-medium on behalf of matt wilkie_

Issue #3184, opened 2026-04-11
FR: Pull should reconstruct milestone hierarchy (--milestones flag)


warmlogic:
## Problem

`bd linear sync --pull` imports issues as flat records, dropping milestone assignments. Linear stores `projectMilestone` on each issue, but pull ignores it.

**Real-world impact:** After pulling a project with 9 milestones and 125 tasks, all tasks arrived as flat orphans. Required 9 manual epic creations + 125 reparent operations to rebuild the hierarchy.

## Proposed Solution

When `bd linear sync --pull` encounters an issue with `projectMilestone`, it should:
1. Create an epic issue for that milestone (if one doesn't already exist)
2. Set the pulled issue's parent to that epic
3. Use milestone name as epic title, milestone description as epic description

**Proposed flag:** `bd linear sync --pull --milestones` (opt-in to avoid breaking existing behavior)

## Linear API Data Available

```json
{
  "projectMilestone": {
    "id": "bd267288-...",
    "name": "M7: Team-Ready",
    "description": "...",
    "progress": 60.61,
    "targetDate": "2026-05-12"
  }
}
```

The `projectMilestone` field is already present on pulled issues — it just needs to be acted on.

## Workaround

Currently using a post-pull reconciliation skill in our Claude Code plugin that queries Linear MCP for milestone assignments and rebuilds the hierarchy. This works but adds N API calls per reconciliation run (one per issue).

maphew:
Thanks for the report. I’m doing a repro sweep and couldn’t find a concrete reproduction recipe in the current description or thread.

Could you add the smallest command sequence or setup that triggers this, including the `bd version`, storage mode, OS, and the expected vs. actual output? A scratch-repo repro is ideal if one is possible.

_codex-gpt-5.5-medium on behalf of matt wilkie_

PR #3528, opened 2026-04-26
Sync Linear milestones as local epics

jozefizso:
## Summary
- add opt-in `bd linear sync --pull --milestones` support for reconstructing Linear project milestone hierarchy
- persist Linear `projectMilestone` metadata and create/reuse local epic issues with synthetic milestone external refs
- parent pulled Linear issues under their milestone epics while keeping those synthetic epics out of Linear push sync

## Validation
- `go test ./internal/linear`
- `go test ./internal/tracker -run 'TestEnginePull|TestEngineCreateDependencies'`
- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestLinearClientFetchIssues|TestLinearPullMilestonesCreatesEpicHierarchy|TestLinearRoundTripCoreFields'`
- `CGO_ENABLED=1 go test -tags gms_pure_go -timeout 5m ./cmd/bd`

Refs #3184

PR #4216, opened 2026-05-28
test(create): guard external-ref prelinking

maphew:
## Summary

- add a fast create-builder regression test for explicit `--external-ref` pre-linking
- cover the nil behavior when no external ref is supplied

## Why

Current `main` already implements the #3254 `bd create --external-ref` path and merged #3517 documents it, but the direct builder behavior did not have a small unit guard. This keeps the implemented pre-linking behavior from regressing while the larger Linear roadmap still has open work, especially #3186.

Refs #3254, #3186, #3184.

## Validation

- `CGO_ENABLED=0 go test ./cmd/bd -run 'TestBuildCreateIssue(ExternalRefPrelink|EmptyExternalRefIsNil)$' -count=1`
- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestBuildCreateIssue(ExternalRefPrelink|EmptyExternalRefIsNil)$' -count=1`
- `git diff --check`

### DIFF
--- a/cmd/bd/create_external_ref_test.go
+++ b/cmd/bd/create_external_ref_test.go
@@ -32,6 +32,6 @@ func TestBuildCreateIssueEmptyExternalRefIsNil(t *testing.T) {
 	})
 
 	if issue.ExternalRef != nil {
-		t.Fatalf("ExternalRef = %q, want nil", *issue.ExternalRef)
+		t.Fatalf("ExternalRef = %v, want nil", issue.ExternalRef)
 	}
 }
--- a/cmd/bd/create_external_ref_test.go
+++ b/cmd/bd/create_external_ref_test.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/steveyegge/beads/internal/types"
+)
+
+func TestBuildCreateIssueExternalRefPrelink(t *testing.T) {
+	ref := "https://linear.app/team/issue/TEAM-123/fix-login"
+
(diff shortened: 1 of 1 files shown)


maphew:
## Review: approve

Small, fast, focused regression guard for the `buildCreateIssue` external-ref pointer conversion.

**Test quality.** Both tests are <50 lines, run in milliseconds, and assert the exact contract: non-empty string → pointer to same string; empty → nil. There is meaningful coverage overlap with existing `TestBuildCreateIssueFromInput_*` (in `create_proxied_server_test.go`) and the `WithExternalRef` case in `create_form_test.go`, but the new tests target `buildCreateIssue` directly rather than `buildCreateIssueFromInput`, so they pin the lower-level builder. Marginal but legitimate.

**Scope.** Doesn't exercise CLI flag wiring or DB persistence, but those are covered transitively by `create_embedded_test.go:471` and `create_proxied_integration_test.go`. The PR is explicit about targeting the builder.

**Roadmap.** #3186 (push bd deps → Linear blocking relations) is correctly framed as the remaining open gap — that's a different code path (relation push vs pre-link identity). #3254 already implemented; #3184 covered by merged #3528.

**Nit fix applied.** Pushed `277b198ad` to swap `*issue.ExternalRef` → `issue.ExternalRef` in the failure-path `t.Fatalf` of `TestBuildCreateIssueEmptyExternalRefIsNil` so the failure branch can no longer nil-deref while reporting a failure.

Approving for squash-merge.

Issue #3926, opened 2026-05-13
bug: bd silently rewrites .beads/dolt-server.port to a gc-managed dolt, causing split-brain when migrating an existing standalone bd store to a managed city


Wldc4rd:
## Summary

`bd 1.0.4` reads gc's `dolt-state.json` (or otherwise detects a running gc-managed dolt server) and silently rewrites the project's `.beads/dolt-server.port` to point at the gc dolt — even when that gc dolt has a different (often empty) database of the same name. For a user who already had a working standalone `.beads/` store with their data and then runs `gc start` in an adjacent city directory, the next `bd list` (or any read-only bd command) hijacks the port file, redirecting all subsequent bd reads/writes to the gc dolt's namesake database, which lacks the user's existing data.

Result: split-brain — the user's existing bd commands now appear to return a near-empty store, while their actual issues sit untouched in the original (now-orphaned) dolt instance. No warning is printed; the failure mode is "I just ran a read-only command and now I can't find my issues."

## Versions

- `bd version 1.0.4 (ce242a879)` (gascityhall installer; installed at `~/.local/bin/bd`)
- `gc version 1.1.0` (linux_amd64 release tarball)
- `dolt version 1.88.1`

## Reproduction

```sh
# 1. Standalone bd workspace with existing issues
$ cd ~/myproject
$ bd init --prefix myproject ...        # set up standalone (works against own dolt on port X)
$ bd count                              # e.g. 100
$ cat .beads/dolt-server.port           # e.g. 37953 — this project's dolt

# 2. Adjacent gc city is started (city's rig points at ~/myproject)
$ cd ~/myproject-dev                    # gc city dir
$ gc start                              # gc starts its own dolt on port 54418

# 3. First bd command from ~/myproject — silently re-routed
$ cd ~/myproject
$ bd count                              # e.g. 0 (queries gc dolt's empty 'myproject' db)
$ cat .beads/dolt-server.port           # 54418 — port file was rewritten!

# 4. Restore the port file manually
$ echo -n 37953 > .beads/dolt-server.port
$ cat .beads/dolt-server.port           # 37953
$ bd count                              # 100 — correct!
$ cat .beads/dolt-server.port           # 54418 — rewritten AGAIN by the bd command
```

Each bd invocation re-resolves the port and silently overwrites the file.

## Why this is surprising

- **Read-only commands have side effects.** `bd count` and `bd list` should not modify state files. The port-file rewrite is invisible unless you `cat` the file after each command.
[...]

PR #4217, opened 2026-05-28
fix(dolt): detect managed handoff port conflicts

maphew:
## Summary

- avoid persisting environment-supplied managed Dolt ports into the repo-local port file
- add `bd doctor` detection for standalone-to-managed-city handoff port conflicts
- document the recovery path in `docs/DOLT.md`

## Why

When an existing standalone `.beads/dolt-server.port` meets a managed-city `DOLT_SERVER_PORT`, silently rewriting or trusting the wrong port can strand users between two Dolt servers. The conservative behavior is to avoid persisting the managed environment port into repo-local state and have doctor surface the handoff conflict with a clear recovery path.

Refs #3926.

## Validation

- `go test -tags gms_pure_go ./internal/storage/dolt -run 'TestShouldPersistResolvedPortFile|TestApplyConfigDefaults' -count=1`
- `go test -tags gms_pure_go ./cmd/bd/doctor -run 'TestCheckManagedHandoffPort' -count=1`
- `go test -tags gms_pure_go ./cmd/bd -run '^$' -count=1`

### DIFF
--- a/cmd/bd/doctor.go
+++ b/cmd/bd/doctor.go
@@ -421,6 +421,13 @@ func runDiagnostics(path string) doctorResult {
 		})
 	}
 
+	// Check 1c: Managed-city handoff port conflict (GH#3926)
+	managedHandoffCheck := convertDoctorCheck(doctor.CheckManagedHandoffPort(path))
+	result.Checks = append(result.Checks, managedHandoffCheck)
+	if managedHandoffCheck.Status == statusWarning || managedHandoffCheck.Status == statusError {
+		result.OverallOK = false
+	}
+
 	// bd-jgxi: Auto-migrate database version before checking it.
 	// Since doctor skips PersistentPreRun DB init (it's in noDbCommands),
 	// trackBdVersion() and autoMigrateOnVersionBump() haven't run yet.
--- a/cmd/bd/doctor/managed_handoff.go
+++ b/cmd/bd/doctor/managed_handoff.go
@@ -1,0 +1,90 @@
+package doctor
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/steveyegge/beads/internal/configfile"
+	"github.com/steveyegge/beads/internal/doltserver"
+)
+
+// CheckManagedHandoffPort detects the #3926 split-brain risk where a managed
+// city/orchestrator port override differs from an existing standalone port file.
+func CheckManagedHandoffPort(repoPath string) DoctorCheck {
+	beadsDir := ResolveBeadsDirForRepo(repoPath)
+
+	cfg, err := configfile.Load(beadsDir)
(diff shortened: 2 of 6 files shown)


maphew:
## Review: approve

Narrow, conservative env-gated guard against rewriting `.beads/dolt-server.port` when a managed-city port override is in play, plus a warning-only `bd doctor` check and matching docs.

**Correctness.** The guard `shouldPersistResolvedPortFile() = BEADS_DOLT_SERVER_PORT == "" && BEADS_DOLT_PORT == ""` gates only the `EnsurePortFile` repair path at `newServerMode` resolution time. Start() still writes the port file when actually launching a server, which is correct: that path needs to record the listening port. The diagnostic in `cmd/bd/doctor/managed_handoff.go` is read-only — no locks, no auto-recovery, no race surface.

**Storage-driver boundary.** Respected. Changes are inside `internal/storage/dolt/` (driver-internal) and a diagnostic surface in `cmd/bd/doctor`. No cross-package locking, no concurrency primitives reaching across the boundary. Aligns with the storage-driver roadmap.

**Discovery-path question (re: open review point).** I traced the managed/orchestrator handoff in `internal/doltserver/doltserver.go` (`DefaultConfig`, lines 461–525). Port resolution priority is:

1. `BEADS_DOLT_SERVER_PORT` env var
2. `.beads/dolt-server.port` (gitignored, local)
3. `config.yaml` (`dolt.port`)
4. `metadata.json` `DoltServerPort` (deprecated, warns)
5. Shared-server default (3308)

`BEADS_DOLT_PORT` is the secondary override checked by `runtimePortOverride()` in `doctor/managed_handoff.go` and is honored by the doltserver code paths that consult it. **There is no `dolt-state.json` reader.** Orchestrator coordination is env-driven (`GT_ROOT`, `BEADS_DOLT_SHARED_SERVER`, `BEADS_DOLT_SERVER_PORT`). So the env-gating in this PR covers the only path that can collide with a managed-city handoff; the port file itself is the *target* of the protection, not a competing source. Config.yaml and metadata.json remain lower-priority/deprecated and don't change behavior here. The env gate is sufficient for the documented scenario.

**Docs.** `docs/DOLT.md#standalone-to-managed-city-handoff` describes detection (`bd doctor`, `bd dolt status`), explicit "do not delete the port file" guidance, and a safe export → stop → import migration. Matches the diagnostic-only behavior of the doctor check.

**Drift vs #3926.** Issue stays open (PR uses `Refs`, not `Closes`). Appropriate, since the symptom (read-only commands rewriting the file) is mitigated for env-driven managed deployments; auto-migration is out of scope.

**Minor notes (non-blocking).** `CheckManagedHandoffPort` reports `StatusWarning` and flips `OverallOK = false`, meaning `bd doctor` exits non-zero on conflict. That's the desired forcing function for split-brain scenarios; called out for visibility. Tests cover happy + conflict paths; the "env set, no port file" branch is covered by `localPort == 0` early-return path but not asserted directly.

Approving for squash-merge.

PR #4218, opened 2026-05-28
test(auto-import): model concurrent writer guard

maphew:
## What

- Adds a focused fallback auto-import test exercising the top-level emptiness guard.
- `SkipsWhenStatisticsReportNonEmptyLate`: asserts the guard still fires when `GetStatistics` returns late (blocked on a busy store) rather than instantly — a timing variant of the existing `SkipsWhenNonEmpty`.
- Asserts the destructive fallback importer is not invoked once statistics report a populated store.

## Why

This hardens the guardrail restored in #3691 against a future refactor that might make the guard's correctness timing-dependent. Current `main` already contains #3691, but the release drift around v1.0.4 and #3870 motivated extra coverage on this path.

## Scope / correction

An earlier revision of this PR named the test `...SkipsWhenConcurrentWriterWinsRace` and described it as covering the #3948 hazard. That was an overclaim and has been corrected:

- `maybeAutoImportJSONL` makes a **single synchronous `GetStatistics` call**, so there is no second observation point a concurrent writer could race against at this layer. The goroutine/channel choreography only delays that one call's return; with a correct non-empty count it is behaviorally a timing variant of `SkipsWhenNonEmpty`, not a race model.
- #3948's actual root cause is the opposite interleaving: `GetStatistics` wrongly returning `TotalIssues=0` on a populated DB (a count / transaction-isolation defect), which defeats the guard entirely; the dangerous fallback TOCTOU is empty-at-check → concurrent writer commits → UPSERT clobbers the new rows. **This PR does not cover that** — it feeds a correct non-empty count and exercises the happy path of the guard.

The #3948 root cause is tracked separately and remains open; #3870 (cut v1.0.5 with #3691) is the higher-leverage release item for anyone currently on v1.0.4.

## Verification

- `go test -tags gms_pure_go ./cmd/bd -run 'TestMaybeAutoImportJSONL|TestShouldRunAutoImportJSONL' -count=1`
- `BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go ./cmd/bd -run 'TestEmbeddedAutoImportJSONL|TestMaybeAutoImportJSONL|TestShouldRunAutoImportJSONL' -count=1`
- `git diff --check`

_codex-gpt-5.5-medium on behalf of maphew_ (original)
_claude-opus-4-8-high on behalf of maphew_ (scope correction)

### DIFF
--- a/cmd/bd/auto_import_upgrade_unit_test.go
+++ b/cmd/bd/auto_import_upgrade_unit_test.go
@@ -95,26 +95,34 @@ func TestMaybeAutoImportJSONL_FallbackImporter_SkipsWhenNonEmpty(t *testing.T) {
 	}
 }
 
-// TestMaybeAutoImportJSONL_FallbackImporter_SkipsWhenConcurrentWriterWinsRace
-// models the #3948 hazard where a separate Dolt writer commits data while a bd
-// process with stale issues.jsonl is deciding whether auto-import should run.
+// TestMaybeAutoImportJSONL_FallbackImporter_SkipsWhenStatisticsReportNonEmptyLate
+// asserts the top-level emptiness guard still fires when GetStatistics returns
+// late (e.g. blocked on a busy store) rather than instantly. It is a timing
+// variant of SkipsWhenNonEmpty: GetStatistics ultimately reports a populated
+// store, so the fallback UPSERT path must not run.
 //
-// The fallback path is the important branch here: embedded stores have an
-// additional in-transaction guard, but fallbackImporter is an UPSERT path that
-// can re-impose stale JSONL over live rows if the top-level guard is removed.
-func TestMaybeAutoImportJSONL_FallbackImporter_SkipsWhenConcurrentWriterWinsRace(t *testing.T) {
+// Scope note — this does NOT cover the gastownhall/beads#3948 root cause. #3948
+// is the case where GetStatistics wrongly returns TotalIssues=0 on a populated
+// DB (a count/transaction-isolation defect), which defeats the guard entirely;
+// the dangerous fallback TOCTOU is empty-at-check -> concurrent writer commits
+// -> UPSERT clobbers the new rows. This test feeds a *correct* non-empty count,
+// so it exercises the happy path of the guard, not that hazard. The #3948 root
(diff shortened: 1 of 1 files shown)


maphew:
Reviewed the change plus a pass over the linked issues (#3948, #3691, #3870, #3630, #2994). Mechanics are clean: CI green, Codecov 100% on touched lines, deterministic timing, tidy `getStatistics` seam. Test-only, low risk.

One substantive point, now addressed in the latest revision: the test originally claimed to cover the #3948 race, but it does not.

- `maybeAutoImportJSONL` makes a **single synchronous `GetStatistics` call** (cmd/bd/auto_import_upgrade.go:54) and branches on that one value. The goroutine + channel choreography only delays that call's return — there is no second observation point for a writer to race against. With a correct non-empty count it is behaviorally identical to the existing `SkipsWhenNonEmpty`, so it added no regression-detection power beyond that sibling.
- #3948's root cause is the **opposite** interleaving: the guard *is* present but `GetStatistics` (→ `ScanIssueCountsInTx` → `SELECT COUNT(*)`) returns `0` on a populated DB, so the guard never fires and the fallback UPSERT clobbers live rows. This test hardcodes a non-empty count, i.e. the happy path that was never #3948's problem. The dangerous fallback TOCTOU (empty-at-check → writer commits → UPSERT) is also not exercised.

Pushed a follow-up commit renaming it to `SkipsWhenStatisticsReportNonEmptyLate` and rewriting the doc comment + PR body to state honestly what it covers and what it doesn't.

Two items surfaced by the linked-issue pass, for tracking rather than this PR:

- **#3948 is still open and unaddressed by this PR** — its real defect (`GetStatistics` returning a wrong `0` under concurrent writers) needs either a count-correctness fix or a fallback-importer hardening (ON DUPLICATE KEY IGNORE / `updated_at` gating), per the issue's own fix directions. Filed locally as the tracking bead for that root cause.
- **#3870 (cut v1.0.5 with #3691) is still open** — the #3691 fix merged but appears unreleased, so anyone on v1.0.4 hits the clobber regardless of any new test. That's the higher-leverage item in this cluster.

LGTM to merge as a guardrail test now that the framing is corrected.

PR #4225, opened 2026-05-28
Skip JSONL auto-import for external Dolt servers

DustinHolden:
## Summary

- classify server-mode configs with an explicit external SQL host, socket, metadata port, or runtime port file as externally managed
- use the resolved server lifecycle mode as a second guard before running `issues.jsonl` auto-import
- add regression coverage for external server metadata, port-file-only server configs, and the auto-import guard

## Why

`issues.jsonl` is an export/interchange surface, not canonical storage. For direct SQL server mode, the configured Dolt SQL database is canonical. If lifecycle detection falls back to owned/local mode, CLI helpers can auto-start or inspect `.beads/dolt` and run the empty-database JSONL recovery path against a shadow local database instead of the configured server.

This showed up with a HomeLab setup using:

```json
{
  "backend": "dolt",
  "dolt_mode": "server",
  "dolt_server_host": "10.10.10.27",
  "dolt_server_user": "beads",
  "dolt_database": "HomeLab"
}
```

plus a local `.beads/dolt-server.port` of `3306`. Normal context reported server mode correctly, but commands could still emit `auto-importing ... issues.jsonl into empty database...` from the local fallback path.

## Tests

- `go test -tags gms_pure_go ./internal/doltserver ./cmd/bd`
- `env -i PATH="$PATH" HOME="$HOME" TMPDIR="${TMPDIR:-/tmp}" GOCACHE="$HOME/Library/Caches/go-build" GOPATH="$HOME/go" go test -tags gms_pure_go ./...`
- manual verification with a patched `bd` against an external server-mode HomeLab repo: `bd create --dry-run ...` no longer emits auto-import and `bd dolt status` reports `running (external)`

### DIFF
--- a/cmd/bd/auto_import_upgrade_unit_test.go
+++ b/cmd/bd/auto_import_upgrade_unit_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/steveyegge/beads/internal/config"
+	"github.com/steveyegge/beads/internal/configfile"
 	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/types"
 )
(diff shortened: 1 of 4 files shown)


maphew:

gorgeguy:
Adding a sanitized concrete repro shape. I cannot share the original private repository or data, but the trigger is not data-specific.

Environment where this was observed:

- `bd version 1.0.4 (ce242a879...)` and current `main` before this PR/hook-path guard
- macOS arm64
- Dolt server mode, with the SQL server lifecycle managed outside the repo
- `.beads/issues.jsonl` exists and is non-empty
- no `sync.remote` is configured
- `import.auto` is at its default `true`

Minimal scratch shape:

```sh
git init bd-jsonl-server-repro
cd bd-jsonl-server-repro
mkdir -p .beads

cat > .beads/metadata.json <<'JSON'
{
  "database": "dolt",
  "backend": "dolt",
  "dolt_mode": "server",
  "dolt_database": "scratch_jsonl_repro"
}
JSON

# Point this at a running Dolt SQL server for scratch_jsonl_repro.
# The exact port is not important; this is the runtime server marker.
printf '3308\n' > .beads/dolt-server.port

# Put any valid non-empty beads JSONL export here.
# In the real repro this was an existing .beads/issues.jsonl export.
ls -l .beads/issues.jsonl

bd hooks run post-merge
```

Actual pre-fix behavior:

- The hook sync path treats `.beads/issues.jsonl` as an import source even though storage is server-backed.
- `bd hooks run post-merge` can shell out to `bd import --quiet .beads/issues.jsonl` and advance Dolt history from the JSONL export.
- In the related startup path, commands can emit the `auto-importing ... issues.jsonl into empty database...` message.

Expected behavior:

- Server-backed Dolt storage should treat the SQL database as canonical and skip automatic JSONL import.
- That needs to apply both to normal startup auto-import and to hook-driven sync import (`post-merge` / related hook paths).

This PR covers the normal startup guard by resolving server mode before `maybeAutoImportJSONL`. While validating locally, I found the same class of issue in the hook path: `importJSONLForSync` also needs to skip when `ResolveServerMode(beadsDir) == ServerModeExternal` or the config is server/proxied-server mode.

Focused validation used locally:

```sh
go test -tags gms_pure_go ./cmd/bd -run 'TestImportJSONLForSync_GuardClauses|TestShouldRunAutoImportJSONL' -count=1
go test -tags gms_pure_go ./internal/doltserver -run 'TestResolveServerMode' -count=1
```

With the startup guard plus the hook-path guard, the same server-mode repo leaves `.beads/issues.jsonl` unchanged and does not import it during `bd hooks run post-merge`.

_codex / gpt-5.5 / xhigh_

PR #4152, opened 2026-05-25
feat(schema): schema-skew guard — hard fail on forward DB drift (be-wwbsv)

quad341:
## Summary

- **SchemaSkewError** type in `internal/storage/schema/schema.go` with `Error()` (terse one-liner), `UserMessage()` (full actionable multi-line block), and `EscapeHint()` methods; `IsSchemaSkewError` helper; unexported `checkSchemaSkew` with `BD_IGNORE_SCHEMA_SKEW=1` escape hatch; exported `CheckForwardDrift` for the read-only store path.
- **store.go** ReadOnly branch now calls `schema.CheckForwardDrift(ctx, db)` before the existing `initSchema` else-branch, returning a `*SchemaSkewError` on forward drift.
- **cmd/bd/schema_skew.go**: `handleSchemaSkewJSON` writes structured JSON to stderr with a `schema_skew` subobject (`current_version`, `required_version`, `delta`).
- **cmd/bd/main.go**: `--ignore-schema-skew` persistent flag (sets `BD_IGNORE_SCHEMA_SKEW=1` in `PersistentPreRun`); `errors.As(*schema.SchemaSkewError)` check before the generic `FatalError("failed to open database")` gives dedicated actionable UX.

## Test plan

Tests from validator branch `tests/be-cdhvj-schema-skew-guard` (be-cdhvj) are included:

- `internal/storage/schema/schema_skew_test.go`: 12 unit tests covering fresh-DB no-op, equal-version no-op, 1/3-migrations-ahead errors, escape-hatch warning, exact error copy (singular/plural), `UserMessage()`, `EscapeHint()`, `IsSchemaSkewError` direct/wrapped/other.
- `internal/storage/dolt/schema_skew_test.go`: 5 dolt-server integration tests (skipped without a test server) covering `CheckForwardDrift` error/no-error/escape-hatch and `New(ReadOnly=true)` forward-drift/escape-hatch/create-if-missing guards.
- `cmd/bd/schema_skew_test.go`: 2 tests — `TestHandleSchemaSkewJSON_Shape` (JSON shape/keys) and `TestIgnoreSchemaSkewFlagRegistered` (flag type check).

All 14 unit tests pass (`CGO_ENABLED=0 go test -tags gms_pure_go`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4226, opened 2026-05-28
test(cmd/bd): flag-to-env-var propagation gate for --ignore-schema-skew (be-0x25q)

quad341:
## Summary

- Adds `TestIgnoreSchemaSkewFlagPropagatesEnvVar` to `cmd/bd/schema_skew_test.go`
- Verifies that `PersistentPreRun` sets `BD_IGNORE_SCHEMA_SKEW=1` when `--ignore-schema-skew` is true
- Closes a coverage gap identified during post-implementation verification of be-wwbsv (PR #4152)
- The existing tests in be-cdhvj cover `checkSchemaSkew()` itself; this test covers the flag→env-var propagation path that all store-open paths use

## Context

PR #4152 (feat: schema-skew guard, be-wwbsv) landed the implementation. During verification, the validator identified that the flag→env-var propagation in `PersistentPreRun` had no direct test — only the downstream `checkSchemaSkew()` behavior was covered. This commit fills that gap.

## Test plan

- [x] `go test ./cmd/bd/ -run TestIgnoreSchemaSkew -v -count=1` — both tests PASS
- [x] No live Dolt required for these unit tests

### DIFF
--- a/cmd/bd/schema_skew_test.go
+++ b/cmd/bd/schema_skew_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"testing"
 
+	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/storage/schema"
 )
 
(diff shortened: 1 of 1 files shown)


maphew (commented):
Thanks for filling this coverage gap after #4152 (be-wwbsv) - flag-to-env-var propagation through `PersistentPreRun` is exactly the kind of path that's easy to leave implicitly tested, and a direct test for it is valuable.

However, on inspection this specific test already exists on `main`:

- **[blocker] Already landed, verbatim.** `cmd/bd/schema_skew_test.go` on current `upstream/main` contains `TestIgnoreSchemaSkewFlagPropagatesEnvVar` with the same doc comment, the same `t.Setenv`/`config.ResetForTesting`/`savePersistentPreRunState` setup, and the same final assertion on `BD_IGNORE_SCHEMA_SKEW`. `git blame` traces it to the 2026-05-28 history-rewrite checkpoint commit (`0a61a566b`, the parentless root of current `main`), later touched by the `RunE` refactor (`d7d7a3460`, 2026-06-15) that switched it to call `rootCmd.PersistentPreRunE` (returns `error`) instead of `PersistentPreRun`. A 3-way apply of this PR's diff onto current `main` conflicts with the existing copy, and a naive resolution would leave a duplicate function - there's no gap left for it to fill.

- **[minor] Signature drift.** Separately from the duplication, this PR's version calls `rootCmd.PersistentPreRun(versionCmd, nil)` and nil-checks `PersistentPreRun`. `main`'s `rootCmd` (`cmd/bd/main.go:644`) only defines `PersistentPreRunE` now, so even a fresh version of this test would need the error-returning form.

On the CONFLICTING status: this isn't an ordinary merge conflict. `git merge-base upstream/main <PR head>` returns nothing - the branch shares no common ancestor with current `main`, whose history is rooted at the 2026-05-28 checkpoint commit, so this branch predates the history rewrite. Normally the remedy would be `git rebase --onto upstream/main <old-base>`, but here that's moot since the intended change is already present on `main` through a different lineage.

Recommendation: retire. Please close this PR as superseded - the coverage it targets already exists on `main`. No action needed from you beyond closing; happy to do it on your behalf if you'd rather not. Thanks again for catching and closing this gap, even if it landed via a parallel path in the interim - the diligence here is appreciated.

maphew:
## Closing: already landed on main via #4152

The single test this PR adds — `TestIgnoreSchemaSkewFlagPropagatesEnvVar` (bead be-0x25q) — already exists verbatim on `main` at `cmd/bd/schema_skew_test.go:91`, landed through your own #4152 (commit `9fea6d3ba`, merged 2026-05-28). The coverage gap this PR was opened to close therefore no longer exists; an earlier agent review on this thread reached the same conclusion and offered to close on the author's behalf.

No code is lost — the gate is live on main. Thanks @quad341 for making sure the propagation path got locked in.

_claude-code-fable-5 on behalf of matt wilkie_

PR #4212, opened 2026-05-28
feat(no-push): guard dolt push/pull and strip push from rendered templates

quad341:
## Summary

Extends the existing `no-push` config option (added in eff58e494) to cover Dolt operations and agent template rendering:

- `bd dolt push` / `bd dolt pull` skip silently (exit 0) when `no-push: true` in config
- `RenderOpts` gains a `NoPush bool`; when `!HasRemote || NoPush`, the `bd dolt push` line is stripped from rendered AGENTS.md templates
- `NoPush` propagated through `AgentsFileParams`, `fs_adapters`, `init`, `setup/agents`, `init_proxied_server`

This is the natural completion of the no-push feature: the existing flag suppressed `git push`, but `bd dolt push` and the template-generated push step were still active in no-push environments.

## Test plan

- [ ] `go test ./cmd/bd/setup/... ./internal/templates/agents/...` — passes
- [ ] `go build ./...` — clean
- [ ] `bd dolt push` on a no-push config returns exit 0 with skip message
- [ ] Generated AGENTS.md omits `bd dolt push` when `NoPush=true`
- [ ] Generated AGENTS.md retains `bd dolt push` when `NoPush=false` (backward compat)
- [ ] Test PR with full unit coverage to follow after this merges

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4227, opened 2026-05-28
fix(init): honor dolt local-only remote wiring

quad341:
## Summary
- Guard bd init Dolt remote configuration at the call site when `dolt.local-only:true`.
- Render agent instructions as no-remote under local-only while preserving `sync.remote` persistence for explicit `--remote`.
- Add config help coverage for `dolt.local-only`.

## Tests
- `go test ./cmd/bd -run Test(ShouldConfigureInitDoltRemoteHonorsLocalOnly|LocalOnlyInitSkipsConfigureButPersistsExplicitRemote|ConfigHelpMentionsDoltLocalOnly) -count=1`
- CI-style short `go test -tags gms_pure_go ... ./cmd/bd` focused set
- `go vet ./...` with clean env
- `make test` in an independent temp clone with clean HOME and Dolt identity

Related Gas City bead: `ga-vjefud.1`. Inspected overlapping PR #4212; this patch is scoped to `dolt.local-only` and does not change no-push behavior.

### DIFF
--- a/cmd/bd/config.go
+++ b/cmd/bd/config.go
@@ -85,7 +85,7 @@ Examples:
   bd config set status.custom "awaiting_review,awaiting_testing"
   bd config set doctor.suppress.pending-migrations true
   bd config set dolt.debug true                        # Enable Dolt sql-server debug mode (loglevel=debug, --prof cpu)
-  bd config set dolt.local-only true                   # Suppress Dolt remote wiring and sync
+  bd config set dolt.local-only true                   # Skip wiring a Dolt sync remote during bd init
   bd config get export.auto
   bd config list
   bd config unset jira.url`,
--- a/cmd/bd/config.go
+++ b/cmd/bd/config.go
@@ -85,6 +85,7 @@ Examples:
   bd config set status.custom "awaiting_review,awaiting_testing"
   bd config set doctor.suppress.pending-migrations true
   bd config set dolt.debug true                        # Enable Dolt sql-server debug mode (loglevel=debug, --prof cpu)
+  bd config set dolt.local-only true                   # Suppress Dolt remote wiring and sync
   bd config get export.auto
   bd config list
   bd config unset jira.url`,
--- a/cmd/bd/config_validate_key_test.go
+++ b/cmd/bd/config_validate_key_test.go
@@ -10,6 +10,7 @@ func TestIsRecognizedConfigKey(t *testing.T) {
 		"export.auto", "dolt.auto-push", "jira.url", "custom.anything",
 		"doctor.suppress.git-hooks", "no-git-ops", "beads.role",
 		"status.custom", "ai.model", "backup.enabled", "import.path",
+		"dolt.local-only",
 	}
 	for _, key := range recognized {
 		if !isRecognizedConfigKey(key) {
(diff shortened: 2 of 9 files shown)


quad341:
CI was red only on **Check doc flags freshness**. The `dolt.local-only` config-set example added in `cmd/bd/config.go` wasn't reflected in the generated CLI docs.

Pushed `7048620bb`, regenerating the doc artifacts from live `bd help` (each is a single `+` line — the new help entry):

- `docs/CLI_REFERENCE.md`
- `website/docs/cli-reference/config.md`
- `website/versioned_docs/version-1.0.0/cli-reference/config.md`
- `website/versioned_docs/version-1.0.4/cli-reference/config.md`
- `website/static/llms-full.txt`

Verified locally with `./scripts/check-doc-flags.sh ./bd` → exit 0 (run under `LC_ALL=C` to match CI's glob collation order, otherwise the `init`/`init-safety` blocks in `llms-full.txt` reorder spuriously). All other checks were already green.

maphew:
## Maintainer review — recommend merge (after CI re-greens)

Thanks for this, @quad341. Reviewed the diff, full CI, and the surrounding `init` remote-wiring code. It's clean, correctly scoped, backward-compatible, and well-tested.

### What I verified
- **All 40 CI checks were green** on `7048620b` (full Dolt matrix, differential regression, doc-flags freshness, lint, vet). Codecov: 83% patch.
- **Key registration is correct** — `dolt.local-only` is accepted via the existing `"dolt."` prefix in `recognizedConfigPrefixes`, so no map edit is needed and the recognition test passes through that path.
- **Config is loaded before the gate** — `config.Initialize()` runs at `init.go:269`, well ahead of `isDoltLocalOnly()` at the line-876 call site, so a pre-existing `config.yaml` is read correctly.
- **The two unchanged `shouldWireInitRemote` call sites are correct, not oversights:**
  - `persistInitSyncRemote` intentionally keeps persisting `sync.remote` so an explicit `--remote` URL is never lost under local-only — matches your design and is covered by `TestLocalOnlyInitSkipsConfigureButPersistsExplicitRemote`.
  - The no-dolt-remote warning stays suppressed when a git origin exists, avoiding a contradictory "go set up a remote" nag under an explicit opt-out.

### Follow-up I pushed to this branch (`1121b94a9`)
Two small non-blocking nits, handled in place rather than asking for another round:

1. **Help wording.** The config-set example read *"Suppress Dolt remote wiring and sync"*, but `dolt.local-only` only gates remote wiring at `bd init` time — it does not gate runtime `bd dolt push`/`pull` (that's the `no-push` territory, which your description already disclaims). Reworded to *"Skip wiring a Dolt sync remote during bd init"* and mirrored into the five generated CLI-reference artifacts.
2. **Test coverage.** Added `TestIsDoltLocalOnly` to exercise the new `isDoltLocalOnly()` config reader through the env path (`BD_DOLT_LOCAL_ONLY`). The existing test passed `localOnly` directly and never touched the config-reading helper (the one line Codecov flagged).

Doc artifacts were regenerated and verified fresh with `CGO_ENABLED=0 -tags gms_pure_go` to match CI: `check-doc-flags.sh`, `check-doc-freshness.sh`, and both `generate-*.sh --check` all pass. Targeted tests + `go vet` + `gofmt` clean.

One open question for naming coherence (non-blocking, no action needed here): there's a `sync.local-only` key referenced in `contributor_routing_e2e_test.go` with no production reader. Worth a future decision on whether these two `*.local-only` concepts should converge, but it doesn't block this PR.

Recommendation: **merge** once CI re-greens on the follow-up commit.

PR #4231, opened 2026-05-28
fix(export): sort issues and sub-objects for deterministic JSONL output

### LINKED ISSUE #4127: bd export: output ordering not stable across runs — produces no-op churn in version control
**Version:** bd 0.62.0 (1402021b), installed via `@beads/bd` npm

## Symptom

Two consecutive `bd export -o <file>` runs with **no database changes** between them produce files whose **line order differs**, even though the line contents are identical.

In a workspace that periodically commits `issues.jsonl` to git for cross-machine sync, this produces a noisy ~30-line "insertion = deletion" diff per sync, even when nothing semantic changed.

## Reproduction

```sh
$ bd export -o /tmp/a.jsonl
Exported 657 issues and 35 memories to /tmp/a.jsonl

$ bd export -o /tmp/b.jsonl
Exported 657 issues and 35 memories to /tmp/b.jsonl

$ diff /tmp/a.jsonl /tmp/b.jsonl | wc -l
63

$ diff <(sort /tmp/a.jsonl) <(sort /tmp/b.jsonl) | wc -l
0
```

The sorted versions are **identical** — same 692 lines, same content. The non-determinism is purely **line ordering**, not field ordering or data drift. In our case the reorder mostly shuffles memory entries vs issue entries, but the precise pattern likely varies.

## Impact

Every periodic sync that commits `issues.jsonl` to git creates a "no-op" churn commit (~24–47 lines per commit in our setup, same lines re-emitted in a different order). Inflates `git log` noise and `git blame` archaeology.

## Suggested fix

Emit lines in a stable order — sort by `id` (or by `_type, id` if memories are intentionally segregated from issues). The output is already line-delimited JSON; stable line ordering is the only thing needed (per-object JSON key ordering already seems stable).

kevglynn:
## Summary

- `bd export` relied on `SearchIssues` SQL ordering, but wisp-merge appends without re-sorting
- This caused non-deterministic export output and phantom git diffs in auto-exported `.beads/` files
- Adds explicit `sort.SliceStable` before writing JSONL (matching SQL order: priority ASC, created_at DESC, id ASC)
- Also sorts dependencies within each issue by `depends_on_id` + `type` for nested stability
- Applied to both `bd export` and the auto-export path in `export_auto.go`

Mirrors the approach used for memories (GH#3474) where `sort.Strings(memKeys)` was added for deterministic output.

## Test plan

- [x] `go build -tags gms_pure_go ./cmd/bd/` passes
- [ ] `bd export > a.jsonl && bd export > b.jsonl && diff a.jsonl b.jsonl` shows no diff
- [ ] Export with wisps present maintains stable ordering across runs

Fixes #4127

### DIFF
--- a/cmd/bd/export.go
+++ b/cmd/bd/export.go
@@ -161,27 +161,8 @@ func runExport(cmd *cobra.Command, args []string) error {
 		issue.Comments = commentsMap[issue.ID]
 	}
 
-	// Sort issues for deterministic JSONL output (GH#4127).
-	// SearchIssues has ORDER BY but wisp-merge can break that ordering.
-	sort.SliceStable(issues, func(i, j int) bool {
-		if issues[i].Priority != issues[j].Priority {
-			return issues[i].Priority < issues[j].Priority
-		}
-		if !issues[i].CreatedAt.Equal(issues[j].CreatedAt) {
-			return issues[i].CreatedAt.After(issues[j].CreatedAt)
-		}
-		return issues[i].ID < issues[j].ID
-	})
-
-	// Sort sub-objects within each issue for stable nested ordering.
-	for _, issue := range issues {
-		sort.Slice(issue.Dependencies, func(a, b int) bool {
-			if issue.Dependencies[a].DependsOnID != issue.Dependencies[b].DependsOnID {
-				return issue.Dependencies[a].DependsOnID < issue.Dependencies[b].DependsOnID
-			}
-			return issue.Dependencies[a].Type < issue.Dependencies[b].Type
-		})
-	}
+	// Sort for deterministic JSONL output (GH#4127).
+	sortIssuesForExport(issues)
 
 	// Write JSONL: one JSON object per line
 	count := 0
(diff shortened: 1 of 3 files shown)


maphew:

kevglynn:
**Reproduction:**

```bash
# Storage: embedded Dolt

# Setup: a project with some issues and wisps
bd init --prefix test && bd create "issue A" -p 1 && bd create "issue B" -p 2

# Trigger: export twice and diff
bd export > /tmp/a.jsonl
bd export > /tmp/b.jsonl
diff /tmp/a.jsonl /tmp/b.jsonl
# Non-empty diff (ordering differs between runs)
```

**Expected**: Identical exports produce identical JSONL (deterministic)
**Actual (before fix)**: Wisp-merge appends without re-sorting, causing non-deterministic output and phantom git diffs in auto-exported `.beads/` files

Fix adds explicit `sort.SliceStable` before writing (priority ASC, created_at DESC, id ASC) plus sub-object sorting for dependencies.

kevglynn:
Closing in favor of a tracking issue. JSONL export/import is being deprecated as source of truth per the 2.0 roadmap (Dolt DB is now canonical). The deterministic ordering principle remains valid for future export formats. Filed tracking issue to carry the intent forward.

PR #4235, opened 2026-05-28
fix(schema): update renumbered migration comment

maphew:
## Summary

- update the longtext migration header comment to match its renumbered `0049` filename

## Why

gastownhall/beads#4224 renumbered the duplicate release migrations so upstream test startup no longer panics on duplicate versions. One embedded comment still referred to the longtext migration as `0046`, which could mislead future migration archaeology.

## Verification

- checked main schema migration `.up.sql` prefixes for duplicate numbers, excluding `migrations/ignored`
- attempted `go test ./internal/storage/schema ./internal/storage/dolt ./cmd/bd`; stopped after Dolt-dependent packages reported missing host ICU header `unicode/uregex.h`

### DIFF
--- a/internal/storage/schema/migrations/0049_longtext_large_content_columns.up.sql
+++ b/internal/storage/schema/migrations/0049_longtext_large_content_columns.up.sql
@@ -1,4 +1,4 @@
--- Migration 0046: Use LONGTEXT for large-content columns.
+-- Migration 0049: Use LONGTEXT for large-content columns.
 --
 -- issues/wisps: description, design, acceptance_criteria, notes, close_reason
 -- comments: text


maphew:
Verified and approving — easy win.

- **Root cause is real**: #4224 renumbered the duplicate LONGTEXT migration to `0049`, and `0046` is now owned by `0046_add_is_blocked`. The embedded `-- Migration 0046:` header therefore pointed at the wrong migration.
- **Complete fix**: this header was the only stale `0046` reference to the longtext migration; nothing else needs updating.
- **Zero functional risk**: the migration version is derived from the filename, not the comment, so this is pure documentation correctness with no schema-behavior change.
- **CI**: all checks green, including `Check no duplicate migration versions` and Codecov.

Merging as-is.

PR #4238, opened 2026-05-28
test(upgrade-smoke): exercise dependency blocker paths on migrated DB

maphew:
## Why

The upgrade-smoke gate (`scripts/upgrade-smoke-test.sh`) creates issues before upgrade but never a **dependency**. As a result the data-copy in `0041_split_dependencies_target` only ever runs as DDL on an *empty* `dependencies` table, and the rewritten `ready_issues` (0044) / `blocked_issues` (0045) views plus the `is_blocked` backfill (0046/0047) get **zero post-migration query coverage**.

That blind spot is exactly the surface of the `bd close` Dolt errno 1105 seen when a binary queries the removed `depends_on_id` against a stale schema: a real v1.0.4 user upgrading to the dependencies-split line with populated dependency rows was not covered by any gate. `migration-test` exercises the data-copy from a v0.63.3 baseline but does not run the blocker/close paths and does not start from a v1.0.4 baseline.

## What

Adds **Scenario 6 — "Dependency blocker paths survive upgrade"**:

1. Creates a blocker task + a dependent bug and wires the dependency **with the previous binary**, so the row is written in the pre-split schema.
2. Upgrades to the candidate (replays 0035 / 0041–0045 / 0047, including the data-copy).
3. Asserts the migrated dependency survives and the blocker paths work:
   - `bd ready` and `bd blocked` succeed (no schema error)
   - the dependent is reported **blocked** pre-close (proves the data-copy populated `depends_on_issue_id`, not just the DDL)
   - `bd close` on the blocker succeeds (the errno-1105 surface)
   - the dependent becomes **ready** after the blocker is closed (proves the `is_blocked` recompute)

The scenario skips gracefully when an old binary can't produce a parseable dependency, matching the existing tolerance in the other scenarios.

## Verification

Built the candidate (`go build -tags gms_pure_go`) and ran `./scripts/upgrade-smoke-test.sh v1.0.4` — all 6 scenarios pass, including the new one against the real v1.0.4 → candidate migration. `bash -n` and `shellcheck` are clean (no new findings).

### DIFF
--- a/scripts/upgrade-smoke-test.sh
+++ b/scripts/upgrade-smoke-test.sh
@@ -420,6 +420,98 @@ else
 fi
 
 # ---------------------------------------------------------------------------
+# Scenario 6: Dependency blocker paths must survive upgrade
+# ---------------------------------------------------------------------------
+#
+# Release gate for the dependencies-split migrations (0035, 0041–0045, 0047).
+# Earlier scenarios create issues but no dependencies, so the data-copy in
+# 0041_split_dependencies_target only ever runs as DDL on an empty table and
+# the rewritten ready_issues/blocked_issues views never get queried on a
+# migrated DB. This scenario populates a dependency with the OLD binary (so the
+# row is written in the pre-0041 schema), upgrades, then exercises the exact
+# blocker paths that fail when the data-copy is wrong: bd ready, bd blocked,
+# bd close — the surface of the current Dolt errno 1105 on a stale schema.
+
+scenario "Dependency blocker paths survive upgrade (ready/blocked/close on migrated deps)"
+
+WS=$(new_workspace)
+
+prev_init 2>/dev/null || true
+git -C "$WS" config beads.role maintainer
+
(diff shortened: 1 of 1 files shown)


maphew (commented):
Review result: no blocking findings.

I checked the added upgrade-smoke scenario against the v1.0.4 dependency-split regression surface. The old-binary dependency creation, migrated blocked-list assertion, blocker close, and post-close ready assertion all line up with the failure mode this is meant to guard.

Local verification:

- `bash -n scripts/upgrade-smoke-test.sh`
- `./scripts/upgrade-smoke-test.sh v1.0.4` (all 6 scenarios passed)

Plain `shellcheck scripts/upgrade-smoke-test.sh` still reports the existing baseline SC1091/SC2034 findings on this script; I did not find new shellcheck findings attributable to the patch.

Leaving this as a review comment rather than an approval because the authenticated GitHub user is the PR author.

PR #3932, opened 2026-05-13
/{cmd,internal}: some cleanup

coffeegoddd:

Issue #3476, opened 2026-04-25
bd remember/memories --global fail with PROJECT IDENTITY MISMATCH — sentinel UUID rejected by identity check


aaronlippold:
## Summary

`bd remember --global` and `bd memories --global` fail with `PROJECT IDENTITY MISMATCH — refusing to connect` from any beads-enabled workspace. The `--global` flag was introduced in #3270 to provide project-agnostic memories on the shared server, but the existing project-identity guard is not aware of the global path and rejects the sentinel UUID set on `beads_global`.

## Repro

```bash
# From any beads-enabled workspace where the shared Dolt server is running
$ bd remember --global "test cross-project memory" --key test
Error: failed to open database: PROJECT IDENTITY MISMATCH — refusing to connect

  Local project ID (metadata.json):  <local-project-uuid>
  Database project ID:               00000000-0000-0000-0000-000000000000
```

`bd memories --global` fails the same way.

## Expected

`--global` should bypass project-identity verification (or accept the sentinel UUID), letting clients read/write the shared `beads_global` database from any project workspace.

## Root cause

In v1.0.2 (`internal/storage/dolt/store.go:1151` `verifyProjectIdentity`), the check skips only when `dbID` is empty:

```go
dbID, err := s.GetMetadata(ctx, "_project_id")
if err != nil || dbID == "" {
    return nil // old database without project_id — skip
}

if localID != dbID {
    return fmt.Errorf("PROJECT IDENTITY MISMATCH ...")
}
```

`beads_global._project_id` is set to `00000000-0000-0000-0000-000000000000` — non-empty, so the check fires and rejects on every `--global` invocation.

PR #3270 ("feat: add global beads_global database for shared-server mode", `fa87dd2b`) added the `--global` flag in `cmd/bd/main.go` and the database setup in `internal/doltserver/doltserver.go` + `cmd/bd/init.go`, but did not modify `verifyProjectIdentity` to handle the global-mode path. Two subsequent PRs touched `store.go` (`65dbdbf5`, `50f715b1`) for unrelated changes; neither addressed this.

## Suggested fix (any one)

1. **In `verifyProjectIdentity`**: also treat the null UUID `00000000-0000-0000-0000-000000000000` as "no ID, skip check" alongside the empty-string case. Defensive, no flag wiring needed.
2. **Plumb the global flag** into `DoltStore` config and skip `verifyProjectIdentity` when `cfg.Global == true`.
[...]

PR #4250, opened 2026-05-29
fix(cli): honor shared-server mode when no project config is found (#3817)

### LINKED ISSUE #3817: cfg == nil fallback in cmd/bd/main.go ignores BEADS_DOLT_SHARED_SERVER, silently creates phantom embedded DBs

maphew:
## What

Honors shared-server mode in the one resolution path that dropped it: when `configfile.Load(beadsDir)` returns `(nil, nil)` because the resolved beads directory has no `metadata.json`/`config.yaml`.

## Why

The shared-server override in `cmd/bd/main.go` lived only inside the `if cfg != nil` branch:

```go
if cfg != nil {
    doltCfg.ServerMode = cfg.IsDoltServerMode()
    if !doltCfg.ServerMode && !doltCfg.ProxiedServer && doltserver.IsSharedServerMode() {
        doltCfg.ServerMode = true
    }
    // ...
} else if cfgErr == nil {
    // cfg == nil: only the default database name was set — ServerMode left false
}
```

So when a user exported `BEADS_DOLT_SHARED_SERVER=1` but the resolved dir had no project config, `doltCfg.ServerMode` stayed `false`. `newDoltStore` keys the backend purely off `cfg.ServerMode` (`store_factory.go`), so bd fell through to `embeddeddolt.Open`. With a leftover embedded data directory present, bd silently read that phantom embedded DB and reported success — fragmenting data away from the shared server instead of connecting to it (originally reported by @mieubrisse).

This is the third concern that was bundled into the (closed) #3865. The identity-verification half was fixed by #3932 (#3818 / #3476); this mode-resolution half was not.

## The fix

Hoist the override so it also runs in the `cfg == nil` / parse-failure cases. The check is idempotent — when the in-branch override already ran, `ServerMode` is already `true`, so nothing changes for the `cfg != nil` paths.

It respects the existing guards (`!doltCfg.ProxiedServer`) and stays on the beads side of the storage boundary: it's a routing/resolution fix (which backend to open), not a new lock/retry or a weakened integrity check.

## Verification

Before/after, with `BEADS_DOLT_SHARED_SERVER=1`, no project config, no reachable server:

| | result |
|---|---|
| before | `No issues found.` (exit 0) — silently used embedded |
| after  | `Error: ... database "beads" not found on Dolt server` (exit 1) — honors shared-server |

Added `TestSharedServerCfgNilHonorsSharedServer` — a hermetic regression test (no container): config-less dir + `BEADS_DOLT_SHARED_SERVER=1` + unreachable server must attempt the server rather than silently fall back to embedded. Confirmed it fails on `main` and passes with this change.

Fixes #3817

### DIFF
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -1007,6 +1007,20 @@ var rootCmd = &cobra.Command{
 			fmt.Fprintf(os.Stderr, "warning: no beads configuration found in %s; using default database name %q\n", beadsDir, configfile.DefaultDoltDatabase)
 			doltCfg.Database = configfile.DefaultDoltDatabase
 		}
+		// Honor shared-server mode even when no project config was found
+		// (cfg == nil) or the parse failed. The override inside the
+		// cfg != nil branch above is skipped in those cases, so without this
+		// an exported BEADS_DOLT_SHARED_SERVER is silently ignored and bd
+		// falls through to embeddeddolt.Open, creating a phantom embedded DB
+		// that subsequent writes fragment into (GH#3817). This is idempotent:
+		// when the override above already ran, ServerMode is already true.
+		if !doltCfg.ServerMode && !doltCfg.ProxiedServer && doltserver.IsSharedServerMode() {
+			doltCfg.ServerMode = true
+			serverMode = doltCfg.ServerMode
+			if cmdCtx != nil {
+				cmdCtx.ServerMode = doltCfg.ServerMode
+			}
+		}
 		// If config parse failed (cfgErr != nil), still default the database
 		// name so the store-open error is about the real problem (the parse
 		// failure warning already printed) rather than a confusing "database
--- a/cmd/bd/shared_server_cfgnil_test.go
+++ b/cmd/bd/shared_server_cfgnil_test.go
@@ -1,0 +1,86 @@
+//go:build cgo
+
+package main
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+)
+
+// TestSharedServerCfgNilHonorsSharedServer is a regression test for GH#3817.
+//
+// When the resolved beadsDir has no metadata.json/config.yaml, configfile.Load
+// returns (nil, nil). The shared-server override in cmd/bd/main.go used to live
+// only inside the `cfg != nil` branch, so an exported BEADS_DOLT_SHARED_SERVER
+// was silently ignored: doltCfg.ServerMode stayed false and newDoltStore fell
+// through to embeddeddolt.Open. With a leftover embedded data directory present
+// (e.g. a clone or recovery dir that has no metadata.json yet) bd would read
+// that phantom embedded DB and silently report success, fragmenting data away
+// from the shared server instead of connecting to it.
+//
+// The fix honors shared-server mode regardless of whether project config was
+// found. This test reproduces the config-less case with no reachable server and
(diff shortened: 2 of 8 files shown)


maphew:
mybd-yc2l: this PR still mergeable, but parked as main.go open path being rewired by Steve and coffeegoddd;

maphew:
Pushed `7a742853e` to remove the stale generated docs refresh from this branch.

The shared-server cfg-nil fix does not change CLI help. The red doc freshness check was caused by the prior docs commit being generated from a CGO-enabled binary, while the current PR doc gate builds `bd` with `CGO_ENABLED=0` and expects the no-CGO federation surface.

Local validation after the fix:

- `CGO_ENABLED=0 go build -tags gms_pure_go -o bd ./cmd/bd/ && ./scripts/check-doc-flags.sh ./bd && ./scripts/check-doc-freshness.sh`
- `make ci-pr-lint`
- `go test -tags 'cgo,gms_pure_go' ./cmd/bd -run TestSharedServerCfgNilHonorsSharedServer -count=1`

GitHub checks are rerunning on `7a742853e`.

maphew:

PR #4266, opened 2026-05-30
fix(deps): deterministic dependency primary keys for cross-clone merge-safety (#4259)

### LINKED ISSUE #4259: bd dolt pull unmergeable after multi-machine upgrade — migration 0043 forks the dependencies primary key (conditional DDL + DEFAULT (UUID()))
### Environment
- `bd` v1.0.5 (Homebrew), two machines (both upgraded at ~the same time)
- Dolt 2.0.7, embedded mode
- Sync: shared beads DB over a `git+ssh` remote via `bd dolt push` / `bd dolt pull` (`refs/dolt/data`)
- macOS (Darwin 25.x)

### Summary
After both machines independently upgraded `bd` and applied schema migrations, `bd dolt pull` fails permanently with a Dolt schema-merge error. The `dependencies` table ended up with **different primary key definitions on the two clones**, which Dolt's 3-way merge cannot reconcile. There is no clean recovery path through `bd` — the only fix was to make one machine canonical and have the other discard and re-adopt its DB.

### Steps to reproduce
1. Two clones of a beads DB syncing via `bd dolt push`/`pull`, both on a pre-0043 schema, with some divergent local issue/dependency activity.
2. Upgrade `bd` on both; each applies migrations through `0043` independently against its own working copy.
3. Run `bd dolt pull` on either machine.

### Expected
Either the migration is deterministic so both clones reach an identical `dependencies` schema and the pull merges, **or** `bd` detects the schema incompatibility before merging and reports it with recovery guidance.

### Actual
```
Error: merge origin/main: Error 1105: error: cannot merge because table dependencies has different primary keys in its common ancestor
```
The error is unhandled/unclassified, so no guidance is given. Pull is blocked indefinitely.

### Root-cause analysis
`internal/storage/schema/migrations/0043_drop_dependencies_generated_column.up.sql` rebuilds the `dependencies` primary key and has two properties that make it unsafe across clones:

1. **Entirely conditional on live state.** The whole migration is gated on a runtime probe (`@needs_drop`, lines 3–9: "does column `depends_on_id` still exist?"). If two clones differ even slightly in pre-migration state (interaction with the equally-conditional `0041`/`0042`, or a partially-applied state), one runs the PK rebuild and the other no-ops (`SELECT 1`). Both then record the **same** migration version, but the table's PK **definition** differs — exactly the "different primary keys in its common ancestor" failure.
2. **Nondeterministic surrogate PK.** Line 59: `ALTER TABLE dependencies ADD COLUMN id CHAR(36) NOT NULL DEFAULT (UUID()) PRIMARY KEY FIRST`. A `UUID()`-defaulted surrogate PK generates different values on each clone, which is hazardous for a database designed to merge across clones (the CHANGELOG frames UUID PKs as enabling "federation-safe multi-clone operation", but a per-clone-random PK undermines row identity across clones).

**Why it's invisible:** the migration cursor records only `MAX(version)` (`internal/storage/schema/schema.go`, via `INSERT IGNORE` into `schema_migrations`) with no per-migration content/schema hash, so two clones at the same `MAX(version)` are assumed identical. The divergence stays silent until a merge fails.

**No safeguard catches it:** `checkSchemaSkew` (`schema.go`) only compares local-DB-vs-local-binary version; `bd doctor` checks are local-only; the pull path (`internal/storage/dolt/store.go` → `CALL DOLT_PULL`) merges blindly; and error classification (`cmd/bd/dolt.go`, `isDivergedHistoryErr`) matches only "no common ancestor", not "different primary keys". CI (`cross-version-smoke`) covers single-machine upgrades only — there is no two-clone post-migration merge test.

### Impact
Silent, unrecoverable-via-`bd` data-sync break for any multi-machine user who upgrades both clones — the core supported topology. Recovery requires manual canonical-adoption and discarding one clone's local history.

### Suggested resolutions
1. **Make migrations deterministic / merge-safe.** Avoid `DEFAULT (UUID())` for a PK that must merge across clones — backfill a deterministic key (e.g. derived from `issue_id` + target) so independent runs produce identical PK values. Prefer unconditional, idempotent DDL over state-probed `IF` branches that can diverge.
2. **Record a per-migration schema/content hash** in `schema_migrations` (not just the version number) and compare it pre-merge so divergence is detectable.
3. **Add a remote-vs-local schema compatibility gate** to `bd dolt push`/`pull`, and classify the "different primary keys in its common ancestor" error with actionable recovery guidance.
4. **Add a CI recipe** that migrates two clones independently and then merges, to catch this class of regression.

maphew:
## Why

Fixes #4259. `dependencies.id` (and `wisp_dependencies.id`) were filled by a DB-side `DEFAULT (UUID())` — a **per-clone-random** surrogate primary key. Because beads syncs by merging independent Dolt clones, two clones that create the *same logical edge*, or that apply migration `0043` independently, end up with the same row under two different primary keys. Dolt then either refuses the merge:

```
Error 1105: cannot merge because table dependencies has different primary keys in its common ancestor
```

or pulls in both rows and trips `uk_dep_issue_target`. Either way `bd dolt pull` breaks with no `bd`-level recovery — the core multi-machine topology. This was the v1.0.5 release-blocker.

The natural identity of a dependency edge is `(issue_id, target)` where `target` is the single non-null of `depends_on_issue_id` / `depends_on_wisp_id` / `depends_on_external` — exactly the columns the `uk_dep_*` unique keys enforce. `type` is deliberately not part of the identity. Deriving `id` from that identity makes the same edge map to the same primary key on every clone.

## What this PR does

- **`internal/storage/depid`** — single source of truth for the key: `depid.New(issueID, target)` returns an RFC-4122 v5 UUID over the natural edge key. Golden vectors pin the value.
- **Every insert site sets it explicitly:** the embedded/`issueops` paths (`AddDependencyInTx`, `replaceDependencyTargetInTx` which re-derives from the new target, the bulk-import path in `CreateIssueInTx`, and `AddWispDependency`) **and** the server-mode/use-case path (`domain/db` `dependencySQLRepositoryImpl.Insert`). Promotion (`promote.go`) and demotion (`ephemeral_routing.go`) carry `id` across their `INSERT ... SELECT` (both tables key `id` identically, so a promoted edge keeps a clone-stable key).
- **Migration `0050`** (idempotent, unconditional — the convergence point every clone reaches when upgrading): drops the random default on `dependencies.id` so a missed insert fails loudly instead of silently re-forking, and re-asserts the `uk_dep_*` unique keys with `CREATE UNIQUE INDEX IF NOT EXISTS` so a clone whose `0043` no-opped on a divergent pre-state converges.
- **`0043` hardened** to drop the default as well, so a not-yet-upgraded clone reaches the no-default schema directly. (Belt-and-suspenders; `0050` is the load-bearing convergence and works on its own.)
- **`MigrateUp` backfill (`rekeyDependencyIDs`)** rewrites existing rows' random ids to the deterministic value, so independently-migrated clones converge to byte-identical, merge-safe `dependencies`. Idempotent; runs only on a migration pass.

## What this fixes, and what is a follow-up

- **Migration-fork hazard (the reported incident) — fully fixed.** Two clones that independently migrate *previously-synced* data converge to byte-identical schema **and** rows, so the merge is clean.
- **Steady-state same-edge hazard ("Hazard B") — fully fixed.** With a deterministic id the same edge is one primary key, so the *unrecoverable* `uk_dep_*` constraint violation no longer occurs. `bd dolt pull` now **auto-resolves** a same-edge dependency conflict that differs only in audit columns (`created_at`/`created_by`/`metadata`/`thread_id`) — the same way it already does for the `metadata` table (`--theirs`, convergent across clones pulling from one remote). A conflict where the dependency *type* differs, or one side deleted the edge, is left for the operator. So two machines that each run `bd dep add X Y` between syncs merge cleanly. (The migrate-vs-adopt prevention gate and a per-migration content hash are separate follow-up PRs.)

## Migration safety

- `ALTER COLUMN id DROP DEFAULT` and `CREATE UNIQUE INDEX IF NOT EXISTS` were verified idempotent on Dolt 2.0.7 (no-op when already in the target state). Both `0050` and the hardened `0043` converge to the same final schema for never-migrated and already-migrated clones. Dropping the default needs no value for existing rows (they already have ids). `0050` is irreversible by design (down is a documented no-op), like `0043`.
- The backfill is committed via the normal migration-commit path for `dependencies`; `wisp_dependencies` is `dolt_ignore`'d so its re-key stays clone-local (it only escapes on promotion, which copies the id).

## Testing

- `depid` golden / determinism / collision unit tests.
- `rekeyDependencyIDs` unit tests: re-keys only divergent rows, skips a missing table, and is idempotent.
- `TestAddDependencyUsesDeterministicID` — the production `AddDependency` path writes the deterministic id (not a random UUID).
- `TestDependencyDeterministicIDMergesAcrossBranches` — two divergent branches creating the same edge: deterministic ids converge to one row on merge; a negative control shows random ids collide on `uk_dep_issue_target`.
- `TestTryAutoResolveMergeConflicts_DependencyAuditOnly` / `…_DependencyTypeConflictLeftAlone` — the pull auto-resolver resolves an audit-only same-edge conflict and leaves a type conflict for the operator; existing metadata auto-resolve tests still pass.
- Full build, `go vet`, and `gofmt` clean; the full Beads/Dolt suite is queued in CI.

## Notes for reviewers

- `wisp_dependencies` keeps its (now dormant) `DEFAULT (UUID())`: it is `dolt_ignore`'d and recreated from its `0021` definition by `EnsureIgnoredTables`, so a migration can't durably drop it — and all wisp insert sites set `id` explicitly, so it never fires in production.
- Existing test fixtures that seed `dependencies` directly now pass an explicit `id` (kept as `UUID()` — opaque setup ids, behaviorally identical to the old default). The canonical `depid.New` pattern is exercised by the production-path and merge tests.
- Both rename sides re-derive `id`: target rename is delete+reinsert, and source rename re-keys via `rekeyDependencySourceInTx` after the `fk_dep_issue` `ON UPDATE CASCADE` moves `issue_id` (the cascade does not recompute the surrogate `id`). This keeps `id == depid.New(issue_id, target)` invariant on every path.

## Update — review round 2 (7bb486e58)

Two blocking review findings, both fixed in this PR:

1. **git-protocol pulls skipped the resolver.** `pullFromRemote` returned out of `doltCLIPull` for git-protocol / credentialed / cloud-auth remotes before `tryAutoResolveMergeConflicts` ran, so the `git+https` topology #4259 targets never got auto-resolution. All three CLI-pull paths now route through `finishCLIPull` → `autoResolveConflictsAfterCLIPull`, which resolves on the store connection (`s.db`, on the pull's branch — a fresh connection defaults to the base branch and misses the conflicts).
2. **Source-rename left a stale `id`.** Fixed by `rekeyDependencySourceInTx` (above), plus `dependencyConflictsAreAuditOnly` now verifies `issue_id` + resolved target from `dolt_conflicts_dependencies` rather than trusting the primary key, so a stale id can never auto-collapse two different edges.

New tests: `TestUpdateIssueIDRekeysDependencySource`, `TestTryAutoResolveMergeConflicts_DependencyStaleIDLeftAlone`, `TestAutoResolveConflictsAfterCLIPull_DependencyAuditOnly`.

_claude-opus-4-8-xhigh on behalf of maphew_

_round 2: claude-opus-4-8-max on behalf of maphew_


maphew (commented):
I found two blocking issues that should be addressed before merge.

1. `bd dolt pull` still skips the new dependency conflict resolver for git-protocol remotes. `pullFromRemote` returns directly from `doltCLIPull` for `isGitProtocolRemote` at `internal/storage/dolt/store.go:2212-2216`, and `doltCLIPull` only shells out to `dolt pull` at `internal/storage/dolt/store.go:2051-2063`. That bypasses `pullWithAutoResolve` / `tryAutoResolveMergeConflicts`, so the `git+ssh` topology described in #4259 still does not get the same-edge dependency auto-resolution this PR claims. Deterministic IDs make the failure recoverable, but `bd dolt pull` still fails instead of resolving audit-only conflicts on the main supported git remote path. Please either route CLI pulls through the resolver after pull errors/conflicts or narrow the claim/tests and add git-protocol coverage.

2. Issue ID renames leave `dependencies.id` stale, which invalidates the resolver's same-PK assumption. `updateIssueIDInTx` updates `issues.id` first at `internal/storage/issueops/bulk_ops.go:227`, so the FK cascade can update `dependencies.issue_id` / `depends_on_issue_id` before `UpdateIssueIDInDependenciesInTx` runs. The later replacement query only looks for the old target (`internal/storage/issueops/dependencies.go:434-480`) and the source update at `internal/storage/issueops/dependencies.go:404-406` does not recompute `id`, so a row can become `(new_issue_id, target)` while keeping `depid.New(old_issue_id, target)`. After that, `dependencyConflictsAreAuditOnly` treats matching primary keys as proof of same edge and only compares `type` (`internal/storage/dolt/store.go:2427-2429`), which can auto-resolve a real semantic conflict if one side has a stale pre-rename id. Rename paths need to rederive dependency ids, or the conflict checker needs to verify `issue_id` and resolved target explicitly before using `--theirs`.

Local checks:

- `timeout 120s go test ./internal/storage/depid ./internal/storage/schema ./internal/storage/domain/db ./internal/storage/issueops`
- `timeout 180s go test ./internal/storage/dolt -run 'Test(AddDependencyUsesDeterministicID|DependencyDeterministicIDMergesAcrossBranches|TryAutoResolveMergeConflicts_DependencyAuditOnly|TryAutoResolveMergeConflicts_DependencyTypeConflictLeftAlone|UpdateIssueIDUpdatesPersistentDependencyTargets)$' -count=1`

maphew:
Both findings confirmed and fixed in 7bb486e58. Thanks — these were real and both are now closed in this PR (you picked "both in this PR").

## 1. git-protocol pulls now run the resolver

You were right: `pullFromRemote` returned straight out of `doltCLIPull` for `isGitProtocolRemote` (and the credentialed / cloud-auth CLI paths), so the `git+https` topology #4259 actually targets never reached `tryAutoResolveMergeConflicts`. All three CLI-pull paths now route through a new `finishCLIPull`, which reconciles the pull error with the resolver result (a pull that fails *because* of conflicts is recoverable once they resolve).

One subtlety worth flagging: the resolver runs on the **store connection (`s.db`)**, not a fresh one. A brand-new connection defaults to the base branch (`main`) and never sees the conflicts, which live on the branch the CLI pull merged into (`s.branch`). `s.db` is already on that branch, and the network transfer already happened in the subprocess, so no long-timeout connection is needed for the local resolve. (This also caught a latent gap: a fresh-connection resolver would have silently no-op'd on any non-`main` branch.)

New test: `TestAutoResolveConflictsAfterCLIPull_DependencyAuditOnly` persists an audit-only conflict into the working set (what a CLI pull leaves) and resolves it out-of-transaction, exactly as `finishCLIPull` does.

## 2. Source-rename no longer leaves a stale id

Confirmed via the FK constraints (`fk_dep_issue` / `fk_dep_issue_target` carry `ON UPDATE CASCADE` since 0042/0043): `updateIssueIDInTx` updates `issues.id` first, the cascade rewrites `dependencies.issue_id` before `UpdateIssueIDInDependenciesInTx` runs, and the old source-side `UPDATE dependencies SET issue_id = ? WHERE issue_id = old` was a no-op that never touched `id` — so the row kept `depid.New(oldID, target)`.

Two changes, belt and suspenders:

- **`rekeyDependencySourceInTx`** re-derives `id` from the post-rename `(newID, target)` for every source row (matching both the post-cascade `newID` and `oldID` defensively). Restores the `id == depid.New(issue_id, target)` invariant everywhere.
- **`dependencyConflictsAreAuditOnly`** no longer trusts the primary key as proof of a shared edge. It now reads `our_/their_issue_id` and the resolved `our_/their_` target from `dolt_conflicts_dependencies` and requires `issue_id` + target (and type) to match on both sides before `--theirs`. So even a stale id that slips through can't auto-collapse two different edges.

New tests: `TestUpdateIssueIDRekeysDependencySource` (rename re-derives the source id) and `TestTryAutoResolveMergeConflicts_DependencyStaleIDLeftAlone` (same-PK / different-`issue_id` is left for the operator — fails on the old PK+type-only check, passes now).

## Local checks

- `go test -tags gms_pure_go ./internal/storage/depid ./internal/storage/schema ./internal/storage/domain/db ./internal/storage/issueops` — pass
- `go test -tags gms_pure_go ./internal/storage/dolt -run 'Test(AddDependencyUsesDeterministicID|DependencyDeterministicIDMergesAcrossBranches|TryAutoResolveMergeConflicts_DependencyAuditOnly|TryAutoResolveMergeConflicts_DependencyTypeConflictLeftAlone|TryAutoResolveMergeConflicts_DependencyStaleIDLeftAlone|UpdateIssueIDUpdatesPersistentDependencyTargets|UpdateIssueIDRekeysDependencySource|AutoResolveConflictsAfterCLIPull_DependencyAuditOnly)$'` — pass
- broader `-run 'Pull|Conflict|Dependenc|Rename|Merge|IssueID|Promote'` — pass; `gofmt`/`go vet` clean

Full Beads/Dolt suite queued in local verification. The PR body's old "Notes for reviewers" line claiming the source rename intentionally leaves `id` stable is now obsolete and updated.

PR #4236, opened 2026-05-28
fix(remote): remove cli invocations of 'dolt remote' and exclusively use store method and sql

coffeegoddd:
On main, 'dolt remote' cli calls are made aggressively, even when using a dolt server. This pr removes these calls and replaces them with an equivalent server query, which reduces io load as the prior to dolt 2.0.8, mixed cli and server usage would always load the full database for the cli command.

Issue #2118, opened 2026-02-25
`bd dolt pull` seems to use the wrong dolt config


smurfix:
Tested on current main, 4e938f7:
```
$ bd dolt pull
Pulling from Dolt remote...
Error: failed to pull from origin/main: Error 1105 (HY000): fatal: remote 'origin' not found.
Please make sure the remote exists.
$ cd .beads/dolt/
.beads/dolt$ dolt pull origin
\ Pulling...
Everything up-to-date
$
```

maphew:
FYI — PR #2178 addresses the root cause here: \d dolt push\/\pull\ now fall back to CLI \dolt push\/\pull\ for SSH remotes, and \d dolt remote add\ writes to both the SQL server and CLI filesystem config so remotes are visible from both surfaces.

sksaranraj:
**bd dolt push**
Pushing to Dolt remote...
Error: failed to push to origin/main: Error 1105 (HY000): fatal: remote 'origin' not found.
Please make sure the remote exists.


**bd dolt pull**
Pulling from Dolt remote...
Error: failed to pull from origin/main: Error 1105 (HY000): fatal: remote 'origin' not found.
Please make sure the remote exists.

what will be the solution for this issue, 

As of now i have to add **dolt remote add origin** and then start exectuing the above commands.

It could be better if this happend as part of bd init

m-u-xyz:
@maphew what's with the mojibake?

@sksaranraj `git init` doesn't set up a remote. So why should `bd init`?

IMHO `bd push`/`pull` should notice that there's no default remote and tell the user to run `bd doctor --fix`; the latter should notice this situation and offer to set the dolt remote to whatever the git "origin" remote is.

sksaranraj:
> [@maphew](https://github.com/maphew) what's with the mojibake?
> 
> [@sksaranraj](https://github.com/sksaranraj) `git init` doesn't set up a remote. So why should `bd init`?
> 
> IMHO `bd push`/`pull` should notice that there's no default remote and tell the user to run `bd doctor --fix`; the latter should notice this situation and offer to set the dolt remote to whatever the git "origin" remote is.

My git operations were working fine, not sure what im missing but i face error with bd dolt push / pull

maphew:
@smurfix heh, thanks for the new term  (mojibake). It looks like something in my tool kit is incorrectly escaping backticks. (I use commandline `gh` a lot for managing github issues and PRs.) Thanks for the flag.

[...]

PR #4268, opened 2026-05-30
feat(schema): prevention gate for silent in-place migration of remote-backed databases (#4259)

maphew:
## Why

Follow-up to #4259 (deterministic dependency ids fix #4266). `bd` auto-runs pending schema migrations the first time a new binary opens an existing database. When two clones that sync through a shared remote each upgrade independently, **both migrate in place and the schema forks**, so `bd dolt pull` breaks with no `bd`-level recovery. The supported flow is *"only ONE machine migrates the database; every other client adopts the migrated database from the remote."* Today nothing enforces that — the fork is created silently by auto-migration on first launch.

This adds the prevention gate (layer 1 of the #4259 fix plan): turn the safe-upgrade headline from a doc people don't read into something `bd` enforces.

## What this PR does

`schema.CheckRemoteMigrateGate` (in the `schema` package, alongside `SchemaSkewError`): before applying pending migrations on a read/write store open, it refuses if the database **already has a recorded schema version, has pending migrations, AND has a remote configured** — and returns a `RemoteMigrateGateError` telling the operator to choose:

- **Migrate** as the single designated migrator: `BD_ALLOW_REMOTE_MIGRATE=1 bd <command>`, then `bd dolt push`.
- **Adopt** the already-migrated database from the remote: `bd bootstrap` (re-clone).

The gate is a **no-op** for: fresh databases (version 0 / missing `schema_migrations`), databases already at the binary's version (no pending migrations), databases with no remote, and read-only opens.

It is wired into **both** store modes:
- Server mode: `DoltStore.initSchema` (before `MigrateUpWithLock`).
- Embedded mode: `EmbeddedDoltStore.initSchema` (before `MigrateUp`).

Embedded mode matters specifically because the **original report was filed against embedded mode** — gating only server mode would leave the primary case exposed. (An earlier draft did exactly that; review caught it.)

The error surfaces with a dedicated `UserMessage` / JSON block, mirroring the existing schema-skew UX. The post-clone warmup in `bd bootstrap` treats the gate as non-fatal so it surfaces on the next command instead of an unrelated wisp-table warning. A missing `dolt_remotes` system table is treated as "no remote" so it can never wedge every store open.

## Testing

- `schema.CheckRemoteMigrateGate` unit tests (sqlmock): escape hatch, fresh DB, at-latest, pending-but-no-remote, **blocked** (pending + remote), and missing-`dolt_remotes` → allowed.
- Server-mode integration test: reopening a behind, remote-backed DB returns the gate error end-to-end.
- Embedded-mode integration test (opt-in `BEADS_TEST_EMBEDDED_DOLT=1`): same, through `embeddeddolt.Open`.
- JSON-handler shape test; existing `TestDoltNew_*` open tests still pass (the gate does not affect normal opens).

## Notes for reviewers

- **By design:** auxiliary, optional-store paths that already swallow store-open errors (e.g. `bd doctor` KV status, ADO config lookup) keep degrading gracefully rather than aborting — the gate surfaces on the operator's primary command via the central store-open handler. Making `bd doctor` hard-fail on a gated DB would be worse.
- **Escape-hatch asymmetry:** schema-skew has both `--ignore-schema-skew` and `BD_IGNORE_SCHEMA_SKEW`; this PR ships only the `BD_ALLOW_REMOTE_MIGRATE` env var (the minimal local form). A matching `--allow-remote-migrate` flag is an easy follow-up if desired.
- **Behind-remote bootstrap:** if you `bd bootstrap` a remote that is itself behind this binary, the gate fires (you'd be the first migrator) — this is intentional "make the choice" behavior, not a regression; the common recipe-correct case (remote already at latest) has no pending migrations and is unaffected.

_claude-opus-4-8-xhigh on behalf of maphew_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- **Remote-migrate prevention gate.** `bd` now refuses to silently auto-apply pending schema migrations to an existing database that has a remote configured (in both server and embedded mode), and tells the operator to choose: migrate (as the single designated migrator, then `bd dolt push`) or adopt the already-migrated database from the remote (`bd bootstrap`). Migrating each clone independently forks the schema and breaks `bd dolt pull` ([#4259](https://github.com/gastownhall/beads/issues/4259)). The gate is a no-op for fresh databases, databases already at the binary's version, databases with no remote, and read-only opens. The designated migrator proceeds with `BD_ALLOW_REMOTE_MIGRATE=1`.
+- **Remote-migrate prevention gate.** `bd` now refuses to silently auto-apply pending schema migrations to an existing database that has a remote configured (in both server and embedded mode), and tells the operator to choose: migrate (as the single designated migrator, then `bd dolt push`) or adopt the already-migrated database from the remote (`bd bootstrap`). Migrating each clone independently forks the schema and breaks `bd dolt pull` ([#4259](https://github.com/gastownhall/beads/issues/4259)). The gate is a no-op for fresh databases, databases already at the binary's version, databases with no remote, and read-only opens. The designated migrator proceeds with `BD_ALLOW_REMOTE_MIGRATE=1`. In server mode the gate also detects remotes persisted on disk in `.dolt/config`, so a freshly (auto-)started server — whose in-memory `dolt_remotes` table is not yet populated — cannot slip a remote-backed database past the gate.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Remote-migrate prevention gate.** `bd` now refuses to silently auto-apply pending schema migrations to an existing database that has a remote configured (in both server and embedded mode), and tells the operator to choose: migrate (as the single designated migrator, then `bd dolt push`) or adopt the already-migrated database from the remote (`bd bootstrap`). Migrating each clone independently forks the schema and breaks `bd dolt pull` ([#4259](https://github.com/gastownhall/beads/issues/4259)). The gate is a no-op for fresh databases, databases already at the binary's version, databases with no remote, and read-only opens. The designated migrator proceeds with `BD_ALLOW_REMOTE_MIGRATE=1`.
+
 ### Fixed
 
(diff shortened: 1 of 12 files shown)


maphew (commented):
Review finding:

- P1: Server-mode remote detection can miss configured remotes after a Dolt server restart. `internal/storage/dolt/store.go:1105` runs `store.initSchema()` before `internal/storage/dolt/store.go:1139` runs `store.syncCLIRemotesToSQL()`, but `syncCLIRemotesToSQL` documents that `dolt_remotes` is empty after restart while CLI remotes persist in `.dolt/config`. `internal/storage/schema/remote_migrate_gate.go:126` only checks `SELECT COUNT(*) FROM dolt_remotes`, so the first write open after upgrading an old remote-backed server-mode DB can see `hasRemote=false` and proceed to `initSchemaOnDBWithRetry`, silently applying the migrations this PR is meant to block. The new server-mode test keeps the SQL remote registered in the same server session, so it does not cover the restarted/auto-started path. Please either sync CLI remotes before the gate or have the gate detect persisted CLI remotes too, and add a regression test that clears SQL remotes while a CLI remote remains.

Validation:

- `go test ./internal/storage/schema -run TestCheckRemoteMigrateGate`
- `go test ./internal/storage/dolt -run TestDoltNew_RemoteMigrateGate_BlocksReopen`
- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run TestHandleRemoteMigrateGateJSON`
- Default `go test ./cmd/bd -run TestHandleRemoteMigrateGateJSON` fails locally because ICU headers are unavailable (`unicode/uregex.h`).

maphew:
Good catch — confirmed and fixed in c5e90b7.

**Root cause (exactly as described).** `DoltStore.initSchema` runs the gate (`store.go:1105` → `CheckRemoteMigrateGate`) before `syncCLIRemotesToSQL` (`store.go:1139`), and the gate only read `SELECT COUNT(*) FROM dolt_remotes`. As `federation.go` documents, a freshly (auto-)started server starts with an empty `dolt_remotes` table while CLI remotes persist in `.dolt/config`, so the first write open after upgrading an old, remote-backed server-mode DB saw `hasRemote=false` and migrated in place — the exact fork #4259 is meant to block. Embedded mode is unaffected: a fresh `Open` already reflects persisted remotes in `dolt_remotes` (its gate test registers via `CALL DOLT_REMOTE`, reopens fresh, and still trips).

**Fix.** New `schema.CheckRemoteMigrateGateWithRemoteCheck(ctx, db, extraHasRemote func() bool)`: when `dolt_remotes` reports no remote, it consults an on-disk fallback before allowing migration. Server mode passes `DoltStore.hasPersistedCLIRemote`, which lists CLI remotes in `CLIDir()` and the server root — the same locations `syncCLIRemotesToSQL` inspects (including the #2118 server-root case). The probe runs only when there is a pending migration **and** the SQL table shows no remote, so the `dolt remote -v` subprocess stays off the common open path. Embedded mode keeps the SQL-only `CheckRemoteMigrateGate`.

I chose "gate detects persisted CLI remotes" over "sync remotes before the gate" to avoid reordering the load-bearing open sequence — `syncCLIRemotesToSQL` also performs the #2118 server-root filesystem migration over the main pool.

**Tests.**

- `schema` (sqlmock, deterministic): fallback-true blocks; fallback-false allows; an SQL remote short-circuits the fallback (probe not called); the fallback is not consulted when there is no pending migration.
- `dolt`: `TestDoltStore_hasPersistedCLIRemote` covers the on-disk probe directly. The test you flagged — `TestDoltNew_RemoteMigrateGate_BlocksReopen` — now exercises the **disk-only path first** (CLI remote in `.dolt/config`, `dolt_remotes` asserted empty), then the SQL path, reusing its single `CREATE`/`DROP DATABASE`.

On the regression test specifically: I folded the disk-path assertion into the existing test rather than adding a second integration test, because two consecutive `CREATE`/`DROP DATABASE` cycles destabilize the test dolt-sql-server — the existing test fails on its own second run with `invalid connection` / `unexpected EOF`, independent of this change. One create/drop, both paths covered.

**Validation.** `go test ./internal/storage/schema -run TestCheckRemoteMigrateGate` and the two `dolt` gate tests pass locally (pure-Go, `CGO_ENABLED=0 -tags gms_pure_go`); full `make test` is queued. As you noted, the default `./cmd/bd` path can't run locally here (ICU `unicode/uregex.h` unavailable).

steveyegge:
Maintainer rebase before merge: resolved conflicts with #4266/#4326, and adapted to #4236 (which removed `syncCLIRemotesToSQL`/`migrateServerRootRemotes`) — kept only the new `hasPersistedCLIRemote` on-disk probe and updated comments referencing the deleted sync. Dropped the t.Setenv test-fix commit that already landed as #4269 (main's variant avoids the CREATE/DROP cycle). All gate unit + integration tests verified locally (server-mode testcontainer and embedded-mode opt-in).

PR #4269, opened 2026-05-30
test(dolt): fix t.Setenv-before-t.Parallel panic in TestCheckForwardDrift_EscapeHatch

maphew:
## Why

`TestCheckForwardDrift_EscapeHatch_ReturnsNil` (added in 7da4f7817) calls `t.Setenv("BD_IGNORE_SCHEMA_SKEW", "1")` and then `setupTestStore`, which marks the test parallel via `t.Parallel()`. Go forbids `t.Setenv` in a parallel test (or one with parallel ancestors), so the test panics the entire `dolt` test binary whenever the isolated test Dolt server is running. `skipIfNoDolt` only skips when the server is absent, so `go test ./internal/storage/dolt/` (or `make test`) on a Dolt-equipped machine reliably hits the panic:

```text
panic: testing: test using t.Setenv, t.Chdir, or cryptotest.SetGlobalRandom can not use t.Parallel
--- FAIL: TestCheckForwardDrift_EscapeHatch_ReturnsNil
```

## Fix

Give the test its own dedicated database so it is non-parallel and may call `t.Setenv`. The test closes the store without `DROP DATABASE`: rapid CREATE/DROP cycles can crash the Dolt testcontainer, and the random test database is discarded when the container terminates.

Reviewed follow-up: feba2481 avoids the mid-package `DROP DATABASE` cleanup that could make the next schema-skew test lose the Dolt server connection.

### DIFF
--- a/internal/storage/dolt/schema_skew_test.go
+++ b/internal/storage/dolt/schema_skew_test.go
@@ -64,10 +64,10 @@ func TestCheckForwardDrift_CurrentVersion_NoError(t *testing.T) {
 // TestCheckForwardDrift_EscapeHatch_ReturnsNil verifies that
 // BD_IGNORE_SCHEMA_SKEW=1 suppresses the forward-drift error.
 //
-// Uses a dedicated CREATE/DROP database rather than setupTestStore: this test
-// calls t.Setenv, which Go forbids in a parallel test, and setupTestStore marks
-// the test parallel. (setupTestStore + t.Setenv panics the dolt test binary
-// whenever the test Dolt server is running.)
+// Uses a dedicated database rather than setupTestStore: this test calls
+// t.Setenv, which Go forbids in a parallel test, and setupTestStore marks the
+// test parallel. (setupTestStore + t.Setenv panics the dolt test binary whenever
+// the test Dolt server is running.)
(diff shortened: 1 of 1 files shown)


maphew:
Reviewed current head `feba2481`.

I found one issue during review: the new dedicated database cleanup used `DROP DATABASE`, and the local two-test selector reproduced a Dolt testcontainer connection loss in the next schema-skew test:

```text
go test ./internal/storage/dolt -run 'TestCheckForwardDrift_EscapeHatch_ReturnsNil|TestDoltNew_ReadOnly_ForwardDrift_EscapeHatch_Succeeds' -count=1 -v
```

I pushed `feba2481` to keep the useful non-parallel `t.Setenv` fix while closing the store without a mid-package `DROP DATABASE`, matching the existing Dolt test harness guidance about rapid CREATE/DROP cycles.

Local validation now passes:

```text
go test ./internal/storage/dolt -run 'TestCheckForwardDrift_EscapeHatch_ReturnsNil|TestDoltNew_ReadOnly_ForwardDrift_EscapeHatch_Succeeds' -count=1 -v
go test ./internal/storage/dolt -run 'TestCheckForwardDrift|TestDoltNew_' -count=1 -v
```

No remaining code issues found. GitHub checks are still queued on the new head.

PR #4270, opened 2026-05-30
feat(schema): per-migration content hash + bd doctor skew detection (#4259 reporter fix No.2)

maphew:
## Why

Reporter fix No.2 for #4259. `schema_migrations` recorded only `(version, applied_at)`, so two clones at the same `MAX(version)` were assumed identical even if a conditional migration sent them to different schemas — the divergence stayed invisible until a merge failed. Record a per-migration content hash so that divergence is detectable.

## What this PR does

- `schema_migrations` gains a `content_hash CHAR(64)` column holding the **SHA-256 of each migration's file content**, written at apply time in the migrate loop.
- The column is added to **fresh** databases via `bootstrapSQL`, and **idempotently** to existing databases at migrate time (`ensureContentHashColumn`, an `INFORMATION_SCHEMA`-guarded `ALTER`). Already-applied rows keep a `NULL` hash — their content is not re-read.
- `AllMigrationsSQL` (the fast-bootstrap bundle) embeds the hashes too.
- **Detection (`bd doctor`):** a new `Migration Content Skew` check compares the local `schema_migrations` against the **cached remote-tracking ref** (`remotes/<remote>/<branch>`, *no network fetch*) and **warns** when this database and its remote applied different content for the same migration version — surfacing the silent #4259 fork as a clear advisory instead of a cryptic merge failure. The comparison is a pure, reusable primitive (`schema.ContentHashSkew`). It is read-only and **does not gate push/pull**, and skips cleanly when there's no remote, no cached ref, or no recorded hashes.

**Not a new fork hazard:** the column definition and the hashes are deterministic and the primary key is unchanged, so two clones independently upgrading both reach an identical `schema_migrations` schema and identical hashes — it still merges cleanly (this is the opposite of the `DEFAULT (UUID())` problem #4259 is about).

**No numbered migration** is used (the column is added via bootstrap + the idempotent ensure step), deliberately so this doesn't collide with the migration number used by the deterministic-id PR.

## Testing

- `ensureContentHashColumn` unit tests (sqlmock): adds the column when missing, no-op when present.
- `TestAllMigrationsSQLRecordsContentHashes` (dolt CLI integration): applies the full bundle and asserts every recorded migration's `content_hash` equals the SHA-256 of its file content.
- Existing `schema` package tests updated for the new INSERT/probe and pass.
- `schema.ContentHashSkew` unit tests (identical / one-diverges / multiple-sorted / one-sided / empty-hash / empty-map) and the `bd doctor` check (sqlmock): no-database, no-remote, no-cached-ref, matches, and diverges→warning.

## Scope

Recording **and** the `bd doctor` detection are included. The remaining follow-up is wiring the comparison into a **`bd dolt push`/`pull` gate** (block/guide before the merge) — that touches the storage/merge boundary and is intentionally left for confirmation with the Dolt storage maintainer before building.


maphew (commented):
Found one issue that should be fixed before merge:

- `schema_migrations.content_hash` is not added for databases already at the latest numbered migration. `MigrateUp` exits when `migrationWorkNeeded` returns false (`internal/storage/schema/schema.go:198-205`), and that helper only checks cursor versions/backfill (`internal/storage/schema/schema.go:286-290`). The new `ensureContentHashColumn` call is inside `migrationSource.migrate` (`internal/storage/schema/schema.go:762-768`), so an already-upgraded v1.0.5 database with no custom status/type backfill never runs the `ALTER`. In that common path, the new doctor check later queries `content_hash` (`cmd/bd/doctor/migration_skew.go:91-97`) and silently reports "schema_migrations unavailable" / no comparison, so the PR does not actually install the recording/detection surface for existing up-to-date databases. The migration work decision needs to account for a missing `content_hash` column, and the `ALTER` needs to be treated as schema work so it is staged/committed even when no numbered migrations or backfills ran.

Local validation:

- `go test ./cmd/bd/doctor -run TestCheckMigrationContentSkew -count=1`
- `go test ./internal/storage/dolt -run TestCheckForwardDrift_EscapeHatch_ReturnsNil -count=1`

Note: full `go test ./cmd/bd/doctor` failed locally in existing tests unrelated to this check (`TestCheckRepoFingerprint_UsesTargetRepoOutsideCWD`, `TestCheckTestPollution_NoTestIssues_EmptyDB`). Preflight is also blocked while the latest CI run is pending and because the PR has no closing issue reference, for example `Fixes #4259`.

maphew:
Thanks — confirmed and fixed in `0dfcc6d1b`. You were right: this was a silent no-op on exactly the databases the PR is meant to protect.

## Root cause (as you traced it)

`MigrateUp` returns early when `migrationWorkNeeded` is false (`schema.go:198-205`), and that helper only considered cursor versions + backfill (`schema.go:286-290`). `ensureContentHashColumn` lived inside `migrationSource.migrate`, which is reached only *after* that gate. So a database already at the latest numbered migration with no pending backfill — the common post-1.0.5 upgrade path — never ran the `ALTER`, the column was never added, and `readMigrationContentHashes` (`cmd/bd/doctor/migration_skew.go`) then errored on the `content_hash` SELECT, so the skew check reported `schema_migrations unavailable` with no comparison.

## Fix

1. `migrationWorkNeeded` now reports work-needed when the `content_hash` column is missing on an at-latest database. The `INFORMATION_SCHEMA` probe is extracted into `hasContentHashColumn` and short-circuits on the main source, so the common (already-converged) path costs one extra catalog query.
2. `ensureContentHashColumn` now returns whether it actually added the column; `migrate` threads that up and `MigrateUp` folds it into the "did we do schema work?" gate (`applied == 0 && !backfilled && appliedIgnored == 0 && !mainColumnAdded && !ignoredColumnAdded`). The `ALTER` is therefore staged + committed through the existing `stageSchemaTables` / `DOLT_COMMIT` path even when no numbered migration or backfill ran.

It stays idempotent: once the column exists, `migrationWorkNeeded` falls through to the backfill check and `MigrateUp` short-circuits as before — no re-`ALTER` on every open.

## Tests

- `TestMigrationWorkNeededAddsContentHashColumnOnUpToDateDB` — at-latest + missing column => work needed (the regression).
- `TestMigrationWorkNotNeededWhenContentHashColumnsPresent` — at-latest + columns present + no backfill => not needed (locks ordering / no ALTER loop).
- `ensureContentHashColumn` tests now assert the added-bool return.

Local validation, all green: `go test ./internal/storage/schema`, `go test ./cmd/bd/doctor -run TestCheckMigrationContentSkew -count=1`, `go vet ./internal/storage/schema`.

## On your secondary notes

- **Pre-existing test failures:** confirmed unrelated. I stashed this branch's changes and re-ran at the clean PR HEAD — `TestCheckTestPollution_NoTestIssues_EmptyDB` fails identically there (`Status "ok" want "warning"` / `"N/A (no database)" want "N/A (query failed)"`), so it is environment/ordering, not this PR. `TestCheckRepoFingerprint_UsesTargetRepoOutsideCWD` passed in my isolated run.
- **Closing-issue reference:** deliberately omitted. #4259 is split across four PRs (#4266 / #4268 / #4269 / #4270); the release-blocker #4266 is the natural closer, and this one is slated to land last. A `Fixes #4259` here would close the umbrella issue early. This PR references #4259 as reporter fix No.2 without a closing keyword on purpose.

## One landing note

The `content_hash` install is itself a tiny in-place schema change on an at-latest DB, so once #4268 (refuse silent in-place migrate of a remote-backed DB) lands, this ensure step must sit on the **allowed** side of that gate — otherwise remote-backed clones (the ones that sync and most need skew detection) would never get the column. Flagging for the rebase rather than solving it here.

steveyegge:
Maintainer rebase onto main (now including #4268): resolved conflicts in `lock_test.go` (main grew a generic `expectColumnExists` helper — `expectContentHashColumnExists` now delegates to it), `cmd/bd/doctor.go` (#4236 removed `CheckRemoteConsistency`, so only the new migration-content-skew check is added), and CHANGELOG. Dropped the t.Setenv test-fix commit already merged as #4269. Unit tests, sqlmock tests, and the dolt-CLI integration tests (`TestAllMigrationsSQLRecordsContentHashes`) verified locally.

PR #3740, opened 2026-05-06
fix(dolt): skip git hooks on internal push of refs/dolt/data (GH#3724)

pmgledhill102:
## Summary

Closes #3724.

`bd dolt push` fails with `fatal: this operation must be run in a work tree` when the user's `init.templateDir` has populated the embedded Dolt cache-mirror's `hooks/` dir with a pre-push hook (typically via the [pre-commit framework](https://pre-commit.com/)). Dolt's internal `git push --porcelain --force-with-lease=… refs/dolt/data` against the bare-style cache repo at `.beads/embeddeddolt/<db>/.dolt/git-remote-cache/<hash>/repo.git/` fires that hook, which then calls `git diff` and bombs out — the cache-mirror has no work tree.

This is the **push-side sibling** of the commit-side bugs already fixed by PR #3626 (GH#3340) and PR #3599 (GH#3598) — same root pattern, different commit site.

## Fix (Option A from the issue)

Set `GIT_CONFIG_PARAMETERS='core.hooksPath=/dev/null'` on the subprocess env in `doltCLIPush` (`internal/storage/dolt/store.go`). Every `git` invocation made by the `dolt push` subprocess inherits the override and skips client-side hooks.

The shape mirrors PR #3626 (`commitBeadsConfig` → `git commit --no-verify`), but at the push site we can't pass `--no-verify` directly because bd shells out to `dolt`, not `git` — Dolt invokes git internally. The env-var path is the supported way to override config for nested git invocations. I verified the quoting works against vanilla git (`GIT_CONFIG_PARAMETERS="'core.hooksPath=/dev/null'" git push` reliably suppresses the pre-push hook).

The same one-liner is also applied to `doltCLIPushRefToPeer` in `federation.go` — same file, same root cause: federation pushes against git-protocol peers go through the same git-remote-cache mirror and have identical exposure.

Option B (sentinel env var) and Option C (scrub cache-mirror hooks at init) from the issue are intentionally **not** included — single-concern PR. C in particular would be a reasonable stacked follow-up.

## What changed

- `internal/storage/dolt/credentials.go` — new `applyNoGitHooksToCmd(cmd)` helper next to `applyS3ChecksumEnvToCmd`, with a comment explaining the cache-mirror / `init.templateDir` interaction.
- `internal/storage/dolt/store.go` — call the helper in `doltCLIPush` after credentials and S3 env are applied.
- `internal/storage/dolt/federation.go` — call the helper in `doltCLIPushRefToPeer` for the federation peer-push path.
[...]

PR #4281, opened 2026-05-31
fix(dolt): disable git hooks on the internal Dolt push SQL path (GH#4272)

### LINKED ISSUE #4272: GH#3724 fix is incomplete: hook-skip only covers the CLI push fallback, not the primary CALL DOLT_PUSH (SQL) path — `bd dolt push` still fails with "must be run in a work tree"
## Summary

PR #3740 (fix for GH#3724) added `applyNoGitHooksToCmd` to disable client-side git hooks during bd's internal `refs/dolt/data` push, so templated cache-mirror hooks don't crash the push with `fatal: this operation must be run in a work tree`.

**But the fix only covers the CLI shell-out fallback path.** It is *not* applied to the primary `CALL DOLT_PUSH` SQL path, which is what bd actually uses for a `git+ssh` remote in embedded mode. As a result, on `bd v1.0.5` (latest) the original symptom is **still reproducible**: `bd dolt push` fails with `fatal: this operation must be run in a work tree` whenever `init.templateDir` has populated the cache-mirror `hooks/` dir with pre-commit-framework hooks.

This is a follow-up to GH#3724 (closed/completed by PR #3740). The deadlock siblings (GH#3340/#3598, PR #3599) are a *separate* and correctly-resolved family — this is purely the push-side hook crash.

## Environment

- `bd version`: **`1.0.5 (Homebrew)`**
- OS: macOS Darwin 25.5.0 (Apple Silicon)
- Mode: embedded (`dolt_mode=embedded`, `serverMode=false`)
- Remote: `git+ssh://git@github.com/<owner>/<repo>.git`
- `init.templateDir` = `~/.git-templates` (hooks dispatch into the [pre-commit framework](https://pre-commit.com/))

## Repro

1. With `init.templateDir` hooks present (pre-commit framework, or any `pre-push` that runs `git diff`/`git status`), in an embedded bd workspace with a `git+ssh` remote.
2. Ensure the cache-mirror hooks dir is populated (happens automatically after `bd init` / `bd bootstrap`; or plant manually to isolate the push path):

   ```sh
   CACHE=".beads/embeddeddolt/<db>/.dolt/git-remote-cache/<hash>/repo.git"
   mkdir -p "$CACHE/hooks"
   cat > "$CACHE/hooks/pre-push" <<'EOF'
   #!/bin/sh
   echo "HOSTILE HOOK FIRED: pre-push" >&2
   git diff --quiet   # fails with "must be run in a work tree" in a bare repo
   exit 1
   EOF
   chmod +x "$CACHE/hooks/pre-push"
   ```
3. Run `bd dolt push`.

### Actual

```text
$ bd dolt push
Pushing to Dolt remote...
Error: push to origin/main: Error 1105: unknown push error; addTableFiles, updateManifestAddFiles: HOSTILE HOOK FIRED: pre-push
fatal: this operation must be run in a work tree
error: failed to push some refs to 'ssh://github.com/<owner>/<repo>.git'
```

The hook **fired** (marker emitted), and the failure surfaces as Dolt **`Error 1105`** — i.e. it came through the in-process SQL engine, *not* the `doltCLIPush` shell path (which would format the error as `dolt push failed: …`).

### Expected

`bd dolt push` succeeds; the bd-internal push does not run user hooks regardless of which push path executes.

Removing the cache-mirror `hooks/` dir (`find .beads/embeddeddolt -type d -name hooks -exec rm -rf {} +`) makes the same push succeed — confirming the hooks are the cause and that the fix simply isn't reaching this path.

## Root cause

`applyNoGitHooksToCmd` (which sets `GIT_CONFIG_PARAMETERS='core.hooksPath=/dev/null'`) is only wired into the two **CLI shell-out** push helpers:

- `internal/storage/dolt/store.go:2040` — `doltCLIPush` (`exec … "dolt", "push", …`)
- `internal/storage/dolt/federation.go:477` — `doltCLIPushRefToPeer` (`exec … "dolt", "push", …`)

Both are documented as fallbacks: *"Used for git-protocol remotes where `CALL DOLT_PUSH` times out through the SQL connection."* They only run on timeout (or cloud-auth drift in server mode per `shouldUseCLIForCloudAuth`, `credentials.go:746`).

The **primary** path for a `git+ssh` remote in embedded mode is `CALL DOLT_PUSH` executed by the **in-process Dolt SQL engine**. That engine spawns `git` for the `refs/dolt/data` push itself, and nothing sets `core.hooksPath=/dev/null` on the bd process (or on the engine's git child) — so the templated `pre-push` hook in the bare cache-mirror fires and dies with the work-tree error. The CLI fallback that *does* disable hooks is never reached, because the SQL push fails fast (hook returns non-zero) rather than timing out.

So the hook-disable needs to apply to the `CALL DOLT_PUSH` path too — not just the CLI shell-out fallback.

## Suggested fix

Any of (roughly in order of smallest blast radius):

- **A.** Set `GIT_CONFIG_PARAMETERS='core.hooksPath=/dev/null'` in the bd process environment around the in-process `CALL DOLT_PUSH` (and `CALL DOLT_PULL`/clone) so the embedded engine's git children inherit it — the env-var analogue of the existing `applyNoGitHooksToCmd`, applied at the SQL call site rather than only the CLI subprocess. For external server mode, the server must be started with the same env (it inherits at startup — see `credentials.go:740-745`).
- **C (defence-in-depth, from the original #3724 discussion).** Scrub the cache-mirror `hooks/` dir during `bd init` / `bd bootstrap`, mirroring PR #3342's precedent — so templated hooks never land there in the first place. This closes the upstream cause regardless of push path.

A + C together would make this robust against both the SQL and CLI paths and against future internal git ops.

## References

- GH#3724 — original report (closed/completed)
- PR #3740 — `fix(dolt): skip git hooks on internal push of refs/dolt/data (GH#3724)` — the fix that's incomplete here (only CLI path)
- GH#3340 / #3598, PR #3599 — sibling *deadlock* family on the commit side (separate; correctly resolved via `--no-verify` + `BD_GIT_HOOK` export-skip)
- PR #3342 — `fix(init): skip hooks during from-jsonl bootstrap commit` — precedent for Option C

pmgledhill102:
## What

Disable client-side git hooks for bd's internal `refs/dolt/data` push on the **primary `CALL DOLT_PUSH` path** — both when the Dolt engine runs in-process (embedded mode) and when it runs as an auto-started `dolt sql-server` (server mode) — not just the CLI shell-out fallback.

Fixes #4272.

## Why

PR #3740 (GH#3724) added `applyNoGitHooksToCmd` — `GIT_CONFIG_PARAMETERS='core.hooksPath=/dev/null'` — but only wired it into the two CLI shell-out helpers (`doltCLIPush`, `doltCLIPushRefToPeer`). Those are documented fallbacks, used only when `CALL DOLT_PUSH` times out through the SQL connection.

The **primary** path runs `CALL DOLT_PUSH/PULL/FETCH` through the Dolt SQL engine, which shells out to git for the `refs/dolt/data` transfer against the bare cache-mirror (`.dolt/git-remote-cache/<hash>/repo.git/`). Nothing disabled hooks there, so a templated `pre-push` hook (from `init.templateDir` / the pre-commit framework) still fired and crashed with `fatal: this operation must be run in a work tree`, surfacing as Dolt `Error 1105`. The CLI fallback that *does* disable hooks is never reached, because the SQL push fails fast rather than timing out.

## Fix

The engine runs in one of two ways, so the override is applied in both:

1. **Embedded mode** runs the engine **in-process**, and the git child it spawns inherits the **bd process** environment. `withRemoteOperationEnv` — the single choke point that all SQL push/pull/fetch funnel through (directly in `store.go`, or via `withEnvCredentials` in `federation.go`) — now sets `GIT_CONFIG_PARAMETERS` for the duration of the op, **unconditionally**. (Previously the no-credentials git+ssh / `file://` case took the helper's lock-free fast path and set no env at all — exactly the case the bug bit.)
2. **Server mode** runs an auto-started `dolt sql-server` subprocess. `serverEnv()` (`internal/doltserver`) sets the override in the spawned server's environment at startup, so every git child it spawns skips hooks.

A pre-existing `GIT_CONFIG_PARAMETERS` is **preserved** in both paths: the hooks override is appended after it (git applies parameters left-to-right, so ours wins) rather than added as a second, ambiguous entry for the same key.

All git-hook knowledge stays in the storage/server layer; `applyNoGitHooksToCmd` (CLI fallback) is unchanged. The `internal/doltserver` copy of the constant can't import `internal/storage/dolt` (that package imports this one), so it's defined locally. Scope is the SQL push path only — Option C (scrub cache-mirror hooks at init) and the `internal/remotecache` clone path remain deferred follow-ups, matching PR #3740.

> **Server-mode upgrade note:** a `dolt sql-server` already running before the upgrade won't have the new env. Run `bd dolt stop` once so the next push starts a server that picks it up. (Embedded mode needs no such step.)

## What changed

- `internal/storage/dolt/credentials.go` — `withRemoteOperationEnv` now always disables hooks (new `setNoGitHooksEnv`, merge-preserving + restoring under `federationEnvMutex`); removed the no-creds/no-s3 early return that skipped env setup.
- `internal/doltserver/doltserver.go` — server launch env built via new `serverEnv()` / `withGitHooksDisabled()`.
- `internal/storage/dolt/credentials_test.go`, `internal/doltserver/doltserver_test.go` — unit tests for both paths (override set during op, restored after, preserves existing value, composes with credentials). Run on regular PR CI (no Docker).

## Verification

Reproduced and fixed end-to-end with two locally-built binaries (macOS, Apple Silicon, `dolt` on PATH, **embedded mode** — the default), using the issue's repro shape: a `file://` bare-git remote, a hostile `pre-push` hook planted in the materialised cache-mirror, then `bd dolt push`.

**Before (build from `main`):**

```
$ bd dolt push
Pushing to Dolt remote...
Error: push to origin/main: Error 1105: unknown push error; addTableFiles, updateManifestAddFiles: HOSTILE HOOK FIRED: pre-push
fatal: this operation must be run in a work tree
error: failed to push some refs to 'file:///.../remote.git'
```

**After (build with this fix)** — identical setup, same hostile hook still present:

```
$ bd dolt push
Pushing to Dolt remote...
✓ Push complete.
```

Other checks (all green):

```
gofmt -l <changed .go files>                                                # clean
go vet -tags gms_pure_go ./internal/storage/dolt/ ./internal/doltserver/    # clean
go test -tags gms_pure_go ./internal/storage/dolt/ ./internal/doltserver/   # ok
CGO_ENABLED=1 golangci-lint run --build-tags=gms_pure_go --timeout=5m ./... # 0 issues
go build -tags gms_pure_go ./...                                            # ok
```

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -684,6 +684,65 @@ const doltServerLogLevel = "warning"
 // Debug mode also raises --loglevel from the default warning to debug;
 // the connection-log spam concern that motivated the warning floor is
 // the price of opting into debug.
+// gitConfigParametersEnv is git's environment variable for injecting one-off
+// config values into every git invocation in the process subtree.
+const gitConfigParametersEnv = "GIT_CONFIG_PARAMETERS"
+
+// noGitHooksConfigParam is the GIT_CONFIG_PARAMETERS entry that disables
+// client-side git hooks (it points core.hooksPath at /dev/null). This mirrors
+// applyNoGitHooksToCmd in internal/storage/dolt, which can't be imported here
+// (that package imports this one).
+const noGitHooksConfigParam = "'core.hooksPath=/dev/null'"
+
+// serverEnv builds the environment for the spawned dolt sql-server. It starts
+// from the current process environment (so the server inherits PATH, HOME,
+// credentials, cloud auth, etc.) and disables client-side git hooks.
+//
+// bd runs every CALL DOLT_PUSH/PULL/FETCH inside this server process, which
+// shells out to git for the `refs/dolt/data` transfer against the embedded
+// cache-mirror (`.dolt/git-remote-cache/<hash>/repo.git/`). That mirror is a
+// bare-style repo with no work tree, but `git init` still copies the user's
+// `init.templateDir` hooks into its `hooks/` dir; a templated `pre-push` hook
+// that runs `git diff` / `git status` then crashes the push with "fatal: this
+// operation must be run in a work tree" (GH#4272).
+//
+// PR #3740 (GH#3724) only covered the CLI shell-out push fallback
+// (applyNoGitHooksToCmd). The primary push path runs CALL DOLT_PUSH inside this
+// server, so the override has to be in the server's environment at startup; the
+// server then inherits it once and every git child it spawns skips hooks. Same
+// intent as the commit-side `--no-verify` fix (GH#3340 / GH#3598).
+func serverEnv() []string {
+	return withGitHooksDisabled(os.Environ())
+}
+
+// withGitHooksDisabled returns env with GIT_CONFIG_PARAMETERS set so git skips
+// client-side hooks. Any pre-existing GIT_CONFIG_PARAMETERS value is preserved
+// and the hooks override appended after it (rather than added as a second,
+// ambiguous entry for the same key); git applies parameters left to right, so
+// the override still wins.
+func withGitHooksDisabled(env []string) []string {
+	const prefix = gitConfigParametersEnv + "="
+	out := make([]string, 0, len(env)+1)
+	merged := false
+	for _, e := range env {
+		if strings.HasPrefix(e, prefix) {
+			existing := strings.TrimPrefix(e, prefix)
+			if existing == "" {
+				out = append(out, prefix+noGitHooksConfigParam)
+			} else {
+				out = append(out, prefix+existing+" "+noGitHooksConfigParam)
+			}
+			merged = true
+			continue
+		}
+		out = append(out, e)
+	}
+	if !merged {
+		out = append(out, prefix+noGitHooksConfigParam)
+	}
+	return out
+}
+
 func buildDoltServerArgs(host string, port int, debug bool, profDir string) []string {
 	var args []string
 	if debug {
--- a/internal/doltserver/doltserver_test.go
+++ b/internal/doltserver/doltserver_test.go
@@ -1961,3 +1961,89 @@ func TestEnsureGlobalDatabase_ServerNotReachable(t *testing.T) {
 		t.Error("expected error when server is not reachable")
 	}
 }
+
+// envValue returns the value of the last entry for key in env, or "" if absent.
+// Mirrors the "last wins" semantics that exec applies for duplicate keys.
+func envValue(env []string, key string) (string, bool) {
+	prefix := key + "="
+	val, found := "", false
+	for _, e := range env {
+		if strings.HasPrefix(e, prefix) {
+			val, found = strings.TrimPrefix(e, prefix), true
+		}
+	}
+	return val, found
+}
+
+func countEnvKeys(env []string, key string) int {
+	prefix := key + "="
+	n := 0
+	for _, e := range env {
+		if strings.HasPrefix(e, prefix) {
+			n++
+		}
+	}
+	return n
+}
+
+// TestServerEnvDisablesGitHooks is the regression test for GH#4272: the
+// auto-started dolt sql-server must run with git client-side hooks disabled, so
+// CALL DOLT_PUSH's internal git push of refs/dolt/data against the cache-mirror
+// doesn't fire templated pre-push hooks.
+func TestServerEnvDisablesGitHooks(t *testing.T) {
+	t.Setenv(gitConfigParametersEnv, "")
+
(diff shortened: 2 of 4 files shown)


pmgledhill102:
Been too long. Might re-raise in the future, as I still think it's relevant, but been ready to merge and passing checks for a month without movement.

PR #4284, opened 2026-06-01
fix: bound embedded Dolt opens

krantiutils:
## Summary
- add a default 30s timeout around embedded Dolt opens so lock/connect waits cannot run forever without a parent deadline
- support `BEADS_EMBEDDED_OPEN_TIMEOUT`, with `BEADS_EMBEDDED_LOCK_TIMEOUT` as a compatibility fallback
- keep the no-CGO `sanitizeDBName` helper available after the storage refactor

## Verification
- `go test -tags gms_pure_go ./cmd/bd -run '^$'`
- `go test -tags gms_pure_go ./internal/storage/embeddeddolt`
- installed locally with `make install-force`
- live town-root smoke: bounded `bd list`, `gt hook --json`, and `gt mail inbox --identity mayor/ --json` returned without hanging

Note: a broader `go test -tags gms_pure_go ./cmd/bd ./internal/storage/embeddeddolt` run was stopped after `cmd/bd` sat in its test binary for over two minutes with negligible CPU; replaced with the build-only `cmd/bd` check above plus focused embedded-Dolt tests.

### DIFF
--- a/cmd/bd/store_factory_nocgo.go
+++ b/cmd/bd/store_factory_nocgo.go
@@ -5,6 +5,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/steveyegge/beads/internal/configfile"
 	"github.com/steveyegge/beads/internal/storage"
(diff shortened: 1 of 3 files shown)


maphew:

maphew (commented):
Reviewed exact head `f99c62ec4b81608a5742a4dd1327e4a1ccb8b73f`.

The embedded-open timeout remains useful and is not superseded on current main; later open PRs touching the no-CGO factory are unrelated. Preserve this first-filed contribution and maintainer-refresh it onto the current driver/v2 code.

There is one compile blocker: `cmd/bd/store_factory_nocgo.go` redeclares `sanitizeDBName`, which already exists in `cmd/bd/database_name.go`. A real `CGO_ENABLED=0` compile fails with `sanitizeDBName redeclared`. Delete the duplicate helper and its `strings` import.

The tests also stop at timeout parsing and deadline arithmetic. Add the missing behavioral repro: hold an embedded connector on a directory, attempt a second `OpenSQL` with a short timeout, and assert it returns within the bound. That validates the actual driver path and answers the earlier request for a reproducible lock-wait case.

Disposition: **maintainer-fix/rebase, retain the contributor attribution, and rerun current CI**.

bee-ghosttrack:
CI triage (bee, maintainer seat): the compile red is a collision with main, not the diff itself — main now has `sanitizeDBName` in `cmd/bd/database_name.go`, so the copy in `store_factory_nocgo.go:81` redeclares it. Rebase onto current main and drop the duplicate; the branch is CONFLICTING as well.

PR #4288, opened 2026-06-01
fix(storage): skip empty Dolt commit when working set has no staged changes

realies:
## Problem

The selective-staging Dolt commit helpers — `versioncontrolops.StageAndCommit`,
`dolt.doltAddAndCommit`, and `dolt.doltAddAndCommitInTx` — `DOLT_ADD` a fixed or
dirty-tracked set of tables and then `DOLT_COMMIT('-m', …)`. A write statement can
succeed and mark its table dirty **without changing any rows**:

- an idempotent `INSERT … ON DUPLICATE KEY UPDATE value = VALUES(value)` re-writing the
  same value (e.g. `SetMetadata`/`SetConfig` heartbeats),
- an `INSERT IGNORE` that hits a duplicate (re-adding an existing label/dependency),
- an `UPDATE … WHERE` that matches nothing (e.g. a same-actor re-claim).

In those cases staging produces an empty set, and `DOLT_COMMIT('-m', …)` is rejected
by Dolt with a server-side **"nothing to commit"** warning logged on *every* call. The
Go error is already swallowed via `isDoltNothingToCommit`, but that does not suppress
the server-side log line. At reconcile/heartbeat/agent-retry cadence this floods the
Dolt server log (observed thousands of warnings/hour on a busy coordination database).

A *global* "is anything pending?" check is **not sufficient** for selective staging:
when an unrelated table is concurrently dirty in the working set, the global check
reports pending, yet staging only the (clean) target tables stages nothing — so the
empty `-m` commit still fires.

## Fix

Add `issueops.HasStagedChanges` (`SELECT COUNT(*) FROM dolt_status WHERE staged = 1`)
and gate each helper on the **staged** set — exactly what `-m` will commit — after the
`DOLT_ADD` loop. `StageAndCommit` keeps a cheap global `HasPendingChanges` fast-path
first to avoid touching the staging machinery when the whole working set is clean.

Also consolidate the six hand-rolled `DOLT_ADD` + `DOLT_COMMIT('-m')` blocks in
`dolt/issues.go` (`UpdateIssue`, `ClaimIssue`, `ClaimReadyIssue`, `CloseIssue`,
`DeleteIssue`, `DeleteIssues`) onto the now-guarded `doltAddAndCommitInTx`, removing
the duplicated staging logic and giving those sites the guard for free.

### Why a staged-set check and not `RowsAffected`

`Update`/`Close` co-write an `events` row, so a zero-rows *main-table* write can still
be a real, committable change. `dolt_status` captures the true staged state where a
per-statement `RowsAffected` check would incorrectly skip those.

### Not over-suppressing

Real `Update` (always bumps `updated_at`), `Claim`, `Close` (rewrites
`closed_at`/`updated_at`), `Delete`, and dependency add/remove all stage > 0 rows, so
they still commit and HEAD advances. Only true empty no-ops are skipped.

## Tests

- **sqlmock unit tests** (`versioncontrolops`): assert `DOLT_COMMIT` is **not** issued
  on an empty staged set (including the unrelated-dirty-table case) and **is** issued
  on a real change. These are discriminating — they fail if the guard is reverted.
- **Integration tests** (`dolt`, real Dolt server): assert no regression /
  over-suppression for dependency add/remove and claim/close with an unrelated table
  (config) dirty in the working set, and that `HasStagedChanges` behaves correctly.

`go build`, `go vet`, `gofmt`, and the full `internal/storage/...` suites pass.

### DIFF
--- a/internal/storage/dolt/empty_commit_test.go
+++ b/internal/storage/dolt/empty_commit_test.go
@@ -1,0 +1,252 @@
+//go:build cgo
+
+package dolt
+
+import (
+	"context"
+	"regexp"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/steveyegge/beads/internal/storage/issueops"
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// TestDoltAddAndCommitSkipsEmptyCommitWithUnrelatedDirtyTable is a regression
+// test for the high-frequency "nothing to commit" Dolt warning spam.
+//
+// Root cause: doltAddAndCommit / doltAddAndCommitInTx stage only a FIXED table
+// list (e.g. ["dependencies"]) and then issue DOLT_COMMIT('-m', ...). An
+// idempotent no-op write (re-adding an existing dependency via INSERT IGNORE, or
+// removing a non-existent one) leaves that table clean, so the DOLT_ADD stages
+// nothing and the '-m' commit fails server-side with "nothing to commit" —
+// logged as a warning on every call. At reconcile cadence this floods dolt.log.
+//
+// Crucially, a *global* "any pending changes" check (issueops.HasPendingChanges,
+// as used by StageAndCommit which stages ALL dirty tables) is NOT sufficient
+// here: when an UNRELATED table (config, issues, …) is dirty in the working set,
+// the global check reports "pending" yet the selective DOLT_ADD still stages
+// nothing, so the empty '-m' commit fires anyway. The fix checks the STAGED set
+// (hasStagedChanges) after staging the target tables.
+//
+// This test reproduces that exact condition: it leaves `config` dirty, then
+// asserts no-op dependency writes neither error nor advance HEAD, while real
+// writes still commit (no over-suppression).
+func TestDoltAddAndCommitSkipsEmptyCommitWithUnrelatedDirtyTable(t *testing.T) {
+	store, cleanup := setupTestStore(t)
+	defer cleanup()
+
+	ctx, cancel := testContext(t)
+	defer cancel()
+
+	a := &types.Issue{ID: "pc-a", Title: "A", Status: types.StatusOpen, Priority: 2, IssueType: types.TypeTask}
+	b := &types.Issue{ID: "pc-b", Title: "B", Status: types.StatusOpen, Priority: 2, IssueType: types.TypeTask}
+	c := &types.Issue{ID: "pc-c", Title: "C", Status: types.StatusOpen, Priority: 2, IssueType: types.TypeTask}
+	for _, iss := range []*types.Issue{a, b, c} {
+		if err := store.CreateIssue(ctx, iss, "tester"); err != nil {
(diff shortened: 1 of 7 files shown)


maphew:

maphew:
The staged-set-vs-pending-set distinction here is the exact right call, and it took real tracing to find: you're correct that a global `HasPendingChanges` gate is defeated whenever an unrelated table is dirty in the working set, and that `RowsAffected` would wrongly skip the co-written `events` row. The sqlmock tests that assert `DOLT_COMMIT` is *not issued* (rather than merely not erroring) are exactly the discriminating coverage this needs. Thank you for the depth of work in this pair of PRs — and apologies both for the wait and for the June comment asking for a repro your PR body already contained; that was our mistake, not a gap in your report.

**Disposition: we want this.** Re-verified against current main (c989b6b87): the problem is fully intact — `StageAndCommit` (internal/storage/versioncontrolops/commit.go:41) still commits unconditionally, `HasStagedChanges` was never added, and `internal/storage/dolt/issues.go` now has *more* hand-rolled `DOLT_ADD`+`DOLT_COMMIT('-m')` blocks than when you wrote this (e.g. issues.go:216, 290, 343, 377, 494, 529, 558). Nothing here was absorbed by other commits.

The branch conflicts because `issues.go` was substantially refactored under you (withRetryTx wrappers, wisp routing, CAS claim paths), so the consolidation half needs redoing rather than mechanical rebasing. Two notes for the re-port:

1. The refactored issues.go sites intentionally ignore `DOLT_ADD` errors (`_, _ = tx.ExecContext(...)`, e.g. issues.go:343) because some tables may be dolt_ignore'd, while `doltAddAndCommitInTx` (internal/storage/dolt/ephemeral_routing.go:303) returns add errors — consolidating onto it changes that behavior, so the helper likely needs best-effort add semantics first.
2. The `HasStagedChanges` guard itself ports cleanly; `versioncontrolops/commit.go` has barely moved.

Per our current policy, rebases onto a moved main are maintainer work, not something we push back on contributors for. Would you like to redo the re-port yourself, or shall we rebase it on our side with your commits and attribution preserved? Either way, this and its companion (4348, which is complementary — it covers the working-set and `-Am` paths this one doesn't) are both headed for merge.

steveyegge:
Same disposition as #4348: picking up the re-port maintainer-side per the offer in the review above, with your commits and attribution preserved (`Co-authored-by` where the consolidation half has to be rewritten against the refactored `issues.go`). Tracked as bd-oibnz, sequenced after the #4348 re-port (bd-ga8p7). If you'd rather do it yourself, say so and I'll stand down.

steveyegge:
Closing the loop: this fix is now on main via #5403 (squash-merged as f8df36db5), a maintainer-side re-port of your work — the HasStagedChanges guard ports directly, and the DOLT_ADD/DOLT_COMMIT consolidation half was re-done against the refactored issues.go (the remaining unification is tracked separately). Thank you @realies — the guard and the diagnosis were yours; attribution is preserved in the landed commit.

PR #4290, opened 2026-06-02
fix(init): initialize planning Dolt schema in fork auto-config (be-wqt8)

quad341:
## What this changes

`bd init` on a fork now correctly initializes the planning Dolt schema before returning. Previously, `autoConfigureForkContributor` created `~/.beads-planning/.beads` but never ran schema initialization. Any command that opened the planning store immediately after `bd init` (e.g., `bd migrate-personal`) would hit "Dolt server unreachable" because the schema tables didn't exist yet.

The fix calls `newDoltStoreFromConfig` on the fresh planning directory right after it is created. In CGO/embedded builds this runs `initSchema` so the store is ready for first use. In no-CGO and server-mode builds the call is a no-op (schema initializes on first use once a server is running) — this path is explicitly non-fatal.

A second, independent fix extends the idempotency guard for `autoConfigureForkContributor` to cover YAML-configured contributor setups. Previously, only the DB config key was checked; contributors configured via `config.yaml` or `BD_ROUTING_CONTRIBUTOR` would trigger a spurious re-setup. The new check on `config.GetValueSource` covers both cases.

## Review notes

- Path for no-CGO/server builds: `newDoltStoreFromConfig` returns `nocgoEmbeddedErrMsg`; the error is intentionally discarded with `_ = planningStore.Close()` (non-fatal by design).
- The idempotency guard is intentionally broader than the description says: catches both `SourceConfigFile` and `SourceEnvVar`, which is correct.
- New test `fork_auto_contributor` is gated `//go:build cgo`; no-CGO builds compile clean.

## Test plan

- [x] `make test` — zero regressions vs main (exit 0, 37.3% total coverage)
- [x] Fork auto-config regression test (`fork_auto_contributor`) asserts `embeddeddolt` dir exists after `bd init`, proving `initSchema` ran
- [x] Release gate: [`release-gates/be-wqt8-planning-init-gate.md`](release-gates/be-wqt8-planning-init-gate.md)

🤖 Deployed by actual-factory

### DIFF
--- a/cmd/bd/init_contributor.go
+++ b/cmd/bd/init_contributor.go
@@ -282,7 +282,8 @@ func autoConfigureForkContributor(ctx context.Context, store storage.DoltStorage
 		return nil
 	}
 
-	// Already configured: idempotent re-init.
+	// Already configured: idempotent re-init. Check both DB config and YAML
+	// config — a YAML-configured contributor setup would be missed by DB-only check.
 	if existing, err := store.GetConfig(ctx, "routing.contributor"); err == nil && existing != "" {
 		if !quiet {
 			fmt.Printf("\n  %s Fork detected (upstream: %s)\n", ui.RenderWarn("⚠"), upstreamURL)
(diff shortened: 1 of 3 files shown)


maphew (approved):
Review outcome: approve.

I checked the fork auto-config path, the config-source guard, and the embedded init regression coverage. The change is narrowly scoped, preserves contributor routing idempotency, and the PR CI is green across the required gates, including embedded Dolt, no-CGO compile, lint, formatting, regression, and upgrade smoke checks.

Local verification:
- `CGO_ENABLED=0 go test ./cmd/bd -run '^$'` passed in a detached PR worktree.
- `go test -tags cgo ./cmd/bd -run 'TestEmbeddedInit/fork_auto_contributor' -count=1` could not run on this Windows host because the local CGO toolchain is missing ICU headers for `github.com/dolthub/go-icu-regex`; I treated that as an environment limitation, with upstream embedded CI covering the path.

No blocking findings.

_kilocode-gpt-5.5-unknown-reasoning on behalf of matt wilkie_

PR #4294, opened 2026-06-02
feat(gitlab): sync issue dependency links on push

maxinflection:
After issue content syncs, mirror local beads dependency edges as GitLab issue links: blocks -> GitLab blocks/is_blocked_by (direction-normalized), related/relates-to -> a single relates_to link per unordered pair (deduped). Children of synced epics also keep the epic's milestone during a dependency-only sync. Reports links_pushed in --json.

Additive only (never deletes remote links). License-aware: blocks/ is_blocked_by needs GitLab Premium/Ultimate; on instances without it the link is reported as a single curated, non-fatal message (links_license_skipped in JSON) while relates_to and milestones still apply. Genuine (non-license) link failures remain real warnings, kept distinct from the license case.

## What

`bd gitlab sync` now pushes beads **dependency relationships** to GitLab as issue
links — not just issue content. After content syncs, each edge becomes the matching
GitLab link:

| beads edge | GitLab link |
|---|---|
| `A blocks B` | `A` **blocks** `B` / `B` **is blocked by** `A` |
| `A related-to C` | `A` **relates to** `C` (single link, deduped) |

It also keeps issues under an epic pointed at that epic's milestone during a
dependency-only sync. `links_pushed` is reported in `--json`.

## Why

The current sync mirrors titles/descriptions/state but drops the structure *between* issues,
so GitLab boards and reviewers can't see what blocks what. This makes the beads
dependency graph first-class in GitLab.

## Scope & design

- **Additive** — only creates missing links, never deletes (stale-link pruning is a
  deliberate follow-up, so we don't clobber links added by hand in GitLab).
- **License-aware & transparent** — `blocks`/`is_blocked_by` requires GitLab
  Premium/Ultimate. Where it's unavailable, those links are reported once as a clear,
  non-fatal message (and `links_license_skipped` in `--json`) while `relates_to` and
  milestones still apply. Genuine (non-license) link failures remain real warnings.
- **Unsupported `blocks` is omitted, not downgraded to `relates_to`.** A fallback might
  seem better than nothing, but it would harm the bidirectional sync: the pull side
  imports GitLab links back into beads dependencies, and an issue link carries no
  metadata, so a fallback `relates_to` is indistinguishable from a genuine one on the
  next pull. It would be written back as a beads `related` edge — silently corrupting
  the source-of-truth dependency graph and dropping the `blocks` direction. It would
  also assert a misleading relationship to anyone reading GitLab, and (since the sync is
  additive) linger after a license upgrade alongside the real `blocks` link. Omitting
  keeps the beads graph clean and is fully reversible: enable the license, re-sync, and
  `blocks` links appear with nothing to undo.
- **Out of scope** — GitLab Epic/work-item hierarchy, cross-project links, stale-link
  removal.

## Verification

- `go test ./internal/gitlab/... ./cmd/bd/` — direction normalization, `relates_to`
  dedup, link idempotency, the epic-milestone pass, and the license paths (license-403
  → graceful skip with `relates_to` still applied; non-license error → surfaced as a
  genuine failure).
- Validated against a live GitLab.com (Ultimate) project and a self-hosted instance:
  correct link direction, `relates_to` deduped to one link, epic milestone applied,
  `--dry-run` writes nothing, re-sync creates nothing new; on the unlicensed instance
  the `blocks` links degrade to the curated non-fatal message while `relates_to` and
  milestones sync normally.

### DIFF
--- a/cmd/bd/gitlab.go
+++ b/cmd/bd/gitlab.go
@@ -574,7 +574,7 @@ func runGitLabSync(cmd *cobra.Command, args []string) error {
 	}
 
 	if jsonOutput {
-		outputJSON(gitlabSyncResult{
+		return outputJSON(gitlabSyncResult{
 			DryRun:              gitlabSyncDryRun,
 			Pulled:              result.Stats.Pulled,
 			Pushed:              result.Stats.Pushed,
@@ -588,7 +588,6 @@ func runGitLabSync(cmd *cobra.Command, args []string) error {
 			MilestonesUpdated:   milestonesUpdated,
 			Warnings:            append(result.Warnings, linkWarnings...),
 		})
-		return nil
 	}
 
 	// Output results
--- a/cmd/bd/gitlab.go
+++ b/cmd/bd/gitlab.go
@@ -15,6 +15,7 @@ import (
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/gitlab"
 	"github.com/steveyegge/beads/internal/metrics"
+	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/tracker"
 	"github.com/steveyegge/beads/internal/types"
 )
@@ -225,6 +226,22 @@ func getGitLabConfig() GitLabConfig {
 
 // getGitLabConfigValue reads a GitLab configuration value from store or environment.
 func getGitLabConfigValue(ctx context.Context, key string) string {
+	// Secret/yaml-only keys (e.g. gitlab.token) live in config.yaml, not the
+	// Dolt database, to avoid leaking secrets when the DB is pushed to remotes.
+	// Read them from config.yaml first, then env, and never touch the store.
+	// Mirrors internal/gitlab/tracker.go getConfig after upstream 99653e059.
+	if config.IsYamlOnlyKey(key) {
+		if val := config.GetString(key); val != "" {
+			return val
+		}
+		if envKey := gitlabConfigToEnvVar(key); envKey != "" {
+			if val := os.Getenv(envKey); val != "" {
+				return val
+			}
+		}
+		return ""
+	}
+
 	// Try to read from store (works in direct mode)
 	if store != nil {
 		value, _ := store.GetConfig(ctx, key)
@@ -422,6 +439,22 @@ func runGitLabProjects(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// gitlabSyncResult holds the JSON output for the gitlab sync command.
+type gitlabSyncResult struct {
+	DryRun              bool     `json:"dry_run"`
+	Pulled              int      `json:"pulled"`
+	Pushed              int      `json:"pushed"`
+	Created             int      `json:"created"`
+	Updated             int      `json:"updated"`
+	Skipped             int      `json:"skipped"`
+	Conflicts           int      `json:"conflicts"`
+	Errors              int      `json:"errors"`
+	LinksPushed         int      `json:"links_pushed"`
+	LinksLicenseSkipped int      `json:"links_license_skipped,omitempty"`
+	MilestonesUpdated   int      `json:"milestones_updated,omitempty"`
+	Warnings            []string `json:"warnings,omitempty"`
+}
+
 // runGitLabSync implements the gitlab sync command.
 // Uses the tracker.Engine for all sync operations.
 func runGitLabSync(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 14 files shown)


maxinflection:
Friendly bump on this one 🙂 and I've freshened it up: rebased on latest main (it was ~120 commits behind) with no conflicts, full test suite green locally.

Context for reviewers: this adds GitLab dependency-link sync on the push side — it's the symmetric other half of #3202, which already landed the pull direction for GitLab. The same links-as-dependencies pattern is already established in Linear (#3519) and Azure DevOps, so this just brings GitLab to parity.

I don't think CI ever triggered here (no checks on the PR) — happy to do whatever's needed to get workflows approved/running, and glad to split the +1.6k diff into smaller commits if that helps review.

Anything you need from me?

maxinflection:
Rebased again onto v1.1.0-rc.2 (this branch was ~240 commits behind). The only conflict was a one-line import-ordering collision in cmd/bd/gitlab.go from the RunE/usage-metrics refactor (#4419) — internal/gitlab/* had no changes across the RC range, so there's no functional overlap with this feature. Build clean, GitLab + link test suites green locally, and the PR is mergeable again.

I think this should be a clean addition to 1.1 whenever it clears RC status. Happy to re-target or re-verify against the final release tag if that's easier for the merge.

maphew (commented):
This lands the push half of GitLab dependency-link sync: after content syncs, beads `blocks` edges become GitLab blocks/is_blocked_by links (direction-normalized), `related`/`relates-to` collapse to one deduped `relates_to` per pair, and epic children keep their milestone on dependency-only passes. It brings GitLab to parity with the Linear and ADO integrations, and the design choices are the right ones: additive-only (consistent with the unmanaged-link preservation direction we took in #4523), and the decision to omit rather than downgrade unlicensed `blocks` links is well argued - a `relates_to` fallback would round-trip back into beads as a corrupted `related` edge.

I verified locally at head 9e71d8b: `internal/gitlab` and the GitLab-named `cmd/bd` tests pass, a test-merge with current main is clean and builds, and dependency-relationship sync is explicitly in-scope per the integration charter. CI has never run on this branch, so it needs a workflow approval before merge.

Findings, none blocking:

- [minor] Link push only fires from `bd gitlab sync`. `runGitLabPush` in `cmd/bd/sync_push_pull.go` gains the new `ContentEqual` hook but not the link pass, so `bd gitlab push <ids>` syncs content without links. Fine as a follow-up, but worth a note so users aren't surprised.
- [minor] `RelationMap` in `internal/gitlab/mapping.go` is now dead: the rewritten `issueLinksToDependencies` in `links.go` ignores `MappingConfig`, so the changed `"is_blocked_by": "blocks"` default is never read. Suggest deleting the GitLab `RelationMap` entirely rather than leaving config surface that implies configurability.
- [minor] The diff carries real fixes beyond the title: milestone identifiers now get a `milestone:` prefix in `ExtractIdentifier`/`FetchIssue`, `updateMilestone` resolves IID to API ID (main was passing a URL-extracted IID to an endpoint that wants the global milestone ID), milestone push gains content-equality to stop repeat PUTs, and `runGitLabSync` now sets `commandDidWrite`. All look correct and are tested (`TestUpdateMilestoneResolvesIIDToAPIID`, `TestFetchIssueMilestoneIdentifier`), but please make sure the squash message names them so the changelog doesn't lose them.

Recommendation: merge-fix. Next steps: we approve workflows and let CI run; if green, this merges as-is and the two minor cleanups (dead `RelationMap`, `bd gitlab push` link parity) land as small follow-ups - happy to absorb those ourselves, or take a quick commit from you if you'd rather.

Thanks for a genuinely well-built PR - the license-aware degradation, the direction-normalization tests, and keeping it rebased twice without being asked all made this an easy review.

maxinflection:
I'm knocking out the recommended fixes this morning. should have a fresh commit shortly and thanks again for looking at this!

maxinflection:
Pushed a follow-up commit on top of this PR (and rebased the branch onto current `main`): **carry closed state to GitLab on push**.

### Why
The dependency-link work here syncs issue *content* and *links*, but push never mapped a bead's **closed** state onto the GitLab issue/milestone. GitLab's `POST /issues` (and milestone/work-item creation) can't set state, so closed beads landed as `opened` and closed epics as `active` milestones — milestone progress bars read 0/X even when every child was done. Reproduced on a real project (~880 beads: 619 closed issues all `opened`, 27 closed epics still active).

### What the commit does
- **Close-after-create** on all three create paths (`CreateIssue`, `createMilestone`, `createTaskWorkItem`) via a best-effort follow-up `state_event=close`, since the create APIs can't set state.
- `BeadsIssueToGitLabFields` now emits `state_event` `close`/`reopen`, so the **update** path flips both directions (reopening a bead reopens the GitLab issue). `deferred` deliberately stays open — only "done" work counts toward milestone completion.
- On a failed best-effort close, the created issue and its `external_ref` are still returned so the next push doesn't create a duplicate; the failure is recorded on a new generic `TrackerIssue.Warnings` field, which the sync engine drains into the sync result. Previously a failed close was only a stderr print, so `--json` could report `errors: 0` while the issue was left open.

### Tests
`internal/gitlab/tracker_state_test.go` (new) covers: closed→closed / open+`deferred`→open, closed epic→closed milestone, update flips both directions, and a failed follow-up close returning the issue while surfacing a warning. Also updated `mapping_test.go`.

Rebased on current `main`; `golangci-lint --build-tags=gms_pure_go ./...`, the tagged build, and `go test -tags gms_pure_go` for the affected packages all pass locally.

maphew:
LGTM, thanks!

PR #4296, opened 2026-06-02
fix(dolt): distinguish a real dolt repo from the sql-server runtime dir

reggieperry:
## What

Add rootDirHasDoltRepo(), keyed on `.dolt/repo_state.json` (written by `dolt init`, absent for a bare sql-server runtime dir), and gate the remote migration on it.

## Why

migrateServerRootRemotes checked only for `.dolt/` existence to decide whether the server root holds a dolt repo. The dolt sql-server, when its data_dir points at the bead store server root, writes `.dolt/sql-server.info` for its own runtime state — so the `.dolt/` check matched a non-repo directory, fell through to a 5–7s `dolt remote -v` walk that errored "not a valid dolt repository", and wedged the gascity reconciler tick on every bd write from the city root.

## Verification

Test: TestRootDirHasDoltRepo covers no-.dolt, sql-server-runtime-only, and real-repo cases — green.

### DIFF
--- a/internal/storage/dolt/federation.go
+++ b/internal/storage/dolt/federation.go
@@ -180,6 +180,23 @@ func (s *DoltStore) syncCLIRemotesToSQL(ctx context.Context) {
 	}
 }
 
+// rootDirHasDoltRepo reports whether rootDir contains a dolt repository
+// initialized via `dolt init`. It is distinguished from a directory that
+// merely contains the sql-server's runtime files (e.g.
+// `.dolt/sql-server.info`).
+//
+// A real dolt repo always has `.dolt/repo_state.json` after `dolt init`.
+// The dolt sql-server, when its data_dir points at the server root, creates
+// only `.dolt/sql-server.info` for its own runtime state. The earlier check
+// `os.Stat(rootDir/.dolt)` could not distinguish the two and fell through
+// to a 5–7 s `dolt remote -v` walk that errored out with "not a valid
+// dolt repository" — wedging the gascity reconciler tick on every bd write
+// from the city root.
+func rootDirHasDoltRepo(rootDir string) bool {
+	_, err := os.Stat(filepath.Join(rootDir, ".dolt", "repo_state.json"))
+	return err == nil
+}
+
 // migrateServerRootRemotes checks the dolt server root directory (dbPath) for
 // remotes that should be propagated to the database CLI directory (CLIDir).
 // This handles GH#2118: users run `dolt remote add` in .beads/dolt/ (server root)
@@ -191,8 +208,7 @@ func (s *DoltStore) migrateServerRootRemotes(cliDir string) []storage.RemoteInfo
 	if rootDir == "" || rootDir == cliDir {
 		return nil
 	}
-	// Server root must have .dolt/ (from dolt init)
-	if _, err := os.Stat(filepath.Join(rootDir, ".dolt")); err != nil {
+	if !rootDirHasDoltRepo(rootDir) {
 		return nil
 	}
 	rootRemotes, err := doltutil.ListCLIRemotes(rootDir)
--- a/internal/storage/dolt/federation_repo_check_test.go
+++ b/internal/storage/dolt/federation_repo_check_test.go
@@ -1,0 +1,96 @@
+package dolt
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestRootDirHasDoltRepo verifies the discriminator that distinguishes a
+// real dolt repository (initialized via `dolt init`) from a directory that
+// merely contains the dolt sql-server's runtime files.
+//
+// Regression context: the dolt sql-server, when its `data_dir` points at
+// the bead store's server root, writes `.dolt/sql-server.info` for its
+// own runtime state. The previous check in migrateServerRootRemotes
+// stat'd only `.dolt/` and so treated this runtime directory as a real
+// repo, falling through to a 5–7 s `dolt remote -v` invocation on every
+// bd write — wedging the gascity reconciler tick.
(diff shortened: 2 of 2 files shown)


maphew:

reggieperry:
Repro details below, with timing measured on the host.

Environment:
- `bd` 1.0.5 (`d11b50f81`, `main@d11b50f818eb`)
- dolt 2.0.7
- Ubuntu 24.04.4 LTS, kernel 6.17.0-35
- Storage: Dolt server mode. One `dolt sql-server` whose `data_dir` is the shared bead-store server root that holds the per-city databases as subdirectories. The server root itself is not `dolt init`'d; the sql-server writes `<root>/.dolt/sql-server.info` there for its own runtime state.

What triggers it. `migrateServerRootRemotes` (`internal/storage/dolt/federation.go`) runs on a store open with `rootDir = s.dbPath`, the server root. Before the fix it gated only on `os.Stat(rootDir/.dolt)`, then called `ListCLIRemotes(rootDir)`, which shells `dolt remote -v` in `rootDir`. The sql-server's `.dolt/sql-server.info` makes `.dolt/` exist without the directory being a repo, so that `dolt remote -v` errors "not a valid dolt repository", and because the migration runs in the per-write path the cost lands on every bd write from the city root.

Measured on this host, with the server root confirmed as `.dolt/sql-server.info` present and `repo_state.json` absent:
- A single `dolt remote -v` in the server root, with the `dolt sql-server` running on that `data_dir`: ~1.8 s (5 samples, 1.73–1.87 s), exit 2, "not a valid dolt repository". The running-server interaction is the cost; the CLI handshakes to the server before erroring.
- The same command against an isolated directory with no running server: 0.06 s. So the latency is the live-server interaction, not the bare wrong-directory error.
- The 5–7 s in the original description is the cost as it manifested in the per-write hot path under the reconciler, not a single isolated call.

Minimal repro of the mechanism (the discriminator the fix keys on):
```bash
# a server root shaped like a live sql-server data_dir, but not dolt init'd
mkdir -p /tmp/sqlsrv-root/.dolt
: > /tmp/sqlsrv-root/.dolt/sql-server.info        # what the sql-server writes into its data_dir
ls /tmp/sqlsrv-root/.dolt                          # sql-server.info present, repo_state.json absent
( cd /tmp/sqlsrv-root && dolt remote -v )          # exit 2, "not a valid dolt repository ..."
```
With a `dolt sql-server` actually running on that `data_dir`, the same `dolt remote -v` rises to ~1.8 s per call, and the pre-fix `migrateServerRootRemotes` ran it on every write.

Expected vs. actual:
- Before the fix: every bd write from the city root incurs the server-root `dolt remote -v` (~1.8 s per call here, the 5–7 s aggregate under the reconciler), erroring "not a valid dolt repository".
- After the fix: `rootDirHasDoltRepo` (keyed on `.dolt/repo_state.json`, `federation.go:195`) returns false for the sql-server runtime directory, so `migrateServerRootRemotes` returns early and walks nothing. `TestRootDirHasDoltRepo` covers the no-`.dolt`, sql-server-only, and real-repo cases.

maphew:
Thank you for this — the diagnosis and the reproduction you posted are excellent, and the root cause is exactly right: gating only on `os.Stat(rootDir/.dolt)` matched the sql-server's runtime `.dolt/sql-server.info` and fell through to a live-server `dolt remote -v` on the per-write path.

I'm closing this as **superseded**, because the code path it patches no longer exists on `main`.

## What changed on main

`migrateServerRootRemotes` and its caller `syncCLIRemotesToSQL` were removed in d00b890a3 (#4236, *"remove cli invocations of 'dolt remote' and exclusively use store method and sql"*). The store-open remote-migrate gate now goes through `hasPersistedCLIRemote` → `HasPersistedRemote` → `doltutil.PersistedRemotes`, which reads `.dolt/repo_state.json` directly with `os.ReadFile` — **no `dolt remote -v` subprocess anywhere on the store-open / per-write path**. So the ~1.8 s live-server stall you measured can't occur on current `main`; the whole CLI-subprocess remote-discovery step that produced it is gone.

Because `migrateServerRootRemotes` is deleted, merging this would reintroduce a function nothing calls (it also conflicts against `main` for the same reason).

## Your insight was independently confirmed

The replacement keys the "is this a real repo?" decision on exactly the artifact you identified: `.dolt/repo_state.json`. `PersistedRemotes` treats a missing `.dolt/repo_state.json` as "not a dolt repository here," which is precisely the discriminator your `rootDirHasDoltRepo` implements — so `main` arrived at the same conclusion you did, via removing the subprocess rather than gating it.

## If it still reproduces

Your trace was on `bd` 1.0.5 (`d11b50f81`). If you build `bd` from current `main` and still see a `dolt remote -v` stall from the city root, please reopen with a fresh trace — the path is different now (the remaining `ListCLIRemotes` call sites are `bd doctor` and the CLI push/pull mirror helpers, not store-open), so a new repro would point at a genuinely separate issue worth its own fix.

Thanks again for the careful report and the timing measurements — they made verifying the supersession straightforward.

maphew:
@reggieperry you answered our repro request in 23 minutes, and then heard nothing for 22 days. Worse: unknown to us, the code path your patch gated had already been deleted from main eleven days before we asked — the request was moot the moment we made it, and checking your patch against current main would have shown that. That's on our process, and I'm sorry for the wait.

Your diagnosis outlived the PR: main's replacement keys "is this a real dolt repo" on exactly the .dolt/repo_state.json discriminator you identified. The offer from the close stands — if a dolt remote -v stall reproduces on a current build, that's a genuinely new issue and it gets fast-tracked.

Thanks for the careful work, and for the patience it shouldn't have taken.

This lapse was uncovered due to the circumstances covered in #5094.

reggieperry:
No worries. I was glad to contribute.

Issue #3212, opened 2026-04-12
bd list: default --limit 50 silently truncates output with no indicator


RaddSnake:
## Summary
`bd list` defaults to `--limit 50` with no visible indicator when results are truncated. This makes a partial view indistinguishable from a complete view, which can cause users (and AI assistants consuming bd output) to conclude a bead store is empty, corrupted, or missing data when it's actually just paginated.

## Reproduction
Bead store with 569 issues:

```
$ bd list --json | jq length
50           # silent truncation — no warning, no "N more hidden"

$ bd list --all --limit 0 --json | jq length
569          # actual count
```

`bd list --status=open` can also return fewer results than the store actually holds with no indicator that the filter-plus-limit combination produced a subset.

## Observed vs expected
Currently there's nothing in `bd list` output that signals "this is a subset". A user seeing 50 rows assumes that's all the data. When a filter returns 0 or a small number, it reads as "the filter matched nothing" rather than "the store is truncated at the limit".

## Suggested fix
Print a one-line hint at the bottom of `bd list` (and other paginated commands) when `len(results) == limit` and `limit != 0`:

```
... showing 50 of N (use --limit 0 for all, or --limit N to raise the cap)
```

Ideally also show a hint on near-empty results when a limit-plus-filter combination could be hiding matches.

## Version
bd 0.62.0

## Priority
Low — functional, not broken. But this is a UX footgun that repeatedly causes false \"my data is gone\" panics. The fix is a single conditional in the list output formatter.

maphew:
Thanks for the report! A fix should land on main shortly via #3243

hwaterer:
The fix from #3243 is gated to interactive terminals, so it doesn't reach the programmatic / `--json` consumers this issue was originally about.

On v1.1.0 (`8e4e59d3`):
- `bd list --limit 5 --json` → a bare 5-element array, no `truncated`/`total` field.
- `bd list --limit 5` → prints 5 rows + `Total: 5 issues`, with nothing indicating more matched; stderr empty.

Root cause (source at v1.1.0): `printTruncationHint` (`cmd/bd/list_output.go:16`) early-returns unless `ui.IsStderrTerminal()` — i.e. `term.IsTerminal(os.Stderr.Fd())` (`internal/ui/terminal.go:20`). So the hint only reaches a human at a TTY; any piped/scripted/CI/agent invocation never sees it, and the `--json` payload carries no truncation field, so a non-interactive `--json` consumer gets no signal at all.

[...]

PR #4300, opened 2026-06-03
wip(list): proxied server list command support

coffeegoddd:


maphew (changes requested):
Thanks for pushing this forward. The proxied list path is valuable, but I found several regressions that should be fixed before this lands.

- `cmd/bd/list_input.go`: direct/embedded mode now passes exactly `effectiveLimit` into the storage query, while the later truncation check still depends on receiving `limit+1` rows. For `bd list --limit N` with more than `N` matches, `len(issues) > effectiveLimit` can no longer become true, so the #3212 warning is never emitted. This matches the current `TestEmbeddedList/limit_truncation_hint` failure.

- `cmd/bd/list_input.go`: direct/embedded list queries still need to fetch all rows before any non-default user sort. `SearchIssuesInTx` ignores `filter.SortBy` and orders/limits by priority before returning rows, so `bd list --sort title --limit 10` now sorts only the first priority-ordered page instead of the full result set. The old behavior fetched all rows for requested sorts, not just for `--sort id`.

- `internal/storage/domain/db/issue_descendants.go`: descendant walking now enables wisp traversal whenever the wisp dependency table exists, even when the caller's filter has `SkipWisps` set. That makes proxied `bd list --parent X --pretty/tree` able to include ephemeral or wisp descendants that the equivalent flat/default list hides.

- `internal/storage/domain/dependency.go`: the output dependency loader always calls `ListByIssueIDs`, which reads only the permanent `dependencies` table. Proxied list rendering calls this for each returned item, so returned wisps lose outgoing edges stored in `wisp_dependencies` in pretty/tree/dot/template output.

- `internal/storage/domain/db/pagination.go`: `--limit 0` with a positive `--offset` returns no pagination clause, so proxied `bd list --offset N --limit 0` ignores the offset and returns the full result set. Unlimited pages still need to honor offset.

PR #4303, opened 2026-06-04
db-proxy connection pooling (opt-in) + proxied-server external deploy path

cstar:
## Problem

Multi-agent orchestrators (e.g. gascity/gc, #1978) run beads against a persistent `dolt sql-server` and fork `bd` for **every** operation. Each `bd` process opens a fresh `*sql.DB` → dials dolt (TCP + MySQL auth handshake + session setup) → exits, closing the pool. The in-process pool never survives the process, so there is zero cross-invocation reuse. Measured on a 1-rig city (7 agents): **71 new connections/sec** to dolt, sql-server at 85–170% CPU while the processlist is mostly `Sleep` — pure connection-setup cost. The existing db-proxy was a byte-transparent 1:1 forwarder (dial + `io.Copy` + close per client), so it didn't help.

Hard constraint: **keep Dolt** (bead versioning + cross-machine sync via dolt remotes).

## Approach

Make the db-proxy a **session-pooling, MySQL-protocol-aware** forwarder. It terminates the client handshake itself (skip-auth: clients are loopback-only and already trusted by the proxy's design), borrows an already-authenticated backend from a pool, forwards the command phase **byte-transparently** (the proven `io.Copy` path — bd's storage/uow/transaction code is unchanged), and on client disconnect sends `COM_RESET_CONNECTION` and returns the backend to the pool instead of closing it.

Pooling is **opt-in** via `BEADS_PROXY_POOL_SIZE` (unset/0 = today's transparent behavior). Embedded and direct ServerMode are unaffected.

### Correctness
- **Capability-flag parity**: byte-transparent forwarding requires client↔proxy and proxy↔backend to negotiate the same caps (result-set encoding depends on `DEPRECATE_EOF`, `PROTOCOL_41`, `QUERY_ATTRIBUTES`, `SESSION_TRACK`, …). The pool is keyed by `(capabilities, database)`; mismatched clients fall back to a fresh dial. All `bd` clients share one driver+DSN → one key → full reuse.
- **Session isolation** via `COM_RESET_CONNECTION` (dolt releases locks, recreates the session, restores DB). go-sql-driver's `ResetSession` does *not* send it, so the proxy does. Reset failure → connection destroyed, not reused.
- **Misalignment guard**: the reset reply must be sequence 1; an out-of-sequence packet (e.g. client killed mid-result) destroys the connection rather than lending a corrupted one.
- **COM_QUIT interception**: go-sql-driver sends COM_QUIT on Close; the client→backend direction is frame-aware enough to swallow it and reclaim the backend instead of forwarding it (which would kill the pooled connection).

A `bd daemon` alternative was rejected (routing the whole domain layer over IPC, or reinventing the wire protocol over a socket; transactions still force per-invocation session pinning → no gain).

## Deployable end-to-end
- Unfenced `bd init --proxied-server` for **external** mode (the full `runInitProxiedServer` already existed behind a blanket "not yet implemented" guard; managed-local stays rejected pending its `.dolt` lifecycle).
- Routed a server-mode store through the proxy in proxied mode (`newProxiedServerRoutedStore`) so the legacy store-based commands (list/ready/stats/update/close/…) work — previously only `bd create` (uow-based) did; the rest dereferenced a nil store.

## Validation
- **Unit** (no deps): wire round-trips; pool reuse/cap/key-isolation/drain/eviction/lifetime against a fake MySQL backend.
- **Integration** (cgo + dolt): real proxyServer query/tx/`DOLT_COMMIT`, rollback, 8 concurrent clients keep private sessions, session-state isolation; container-backed `bd init --proxied-server` external + create/list/ready/update/close.
- **Benchmark** `BenchmarkConnectionChurn`: dolt `Connections` delta over 50 sequential sessions — pooling **off = 50** (1:1 churn), **on = 0** (50 hits, 1 dial).
- **Real CLI** vs a quiet dolt: **7.02 → 1.02** new dolt connections per `bd` invocation (~7×); functional correctness against a live managed dolt; zero pool reset/misalignment events. (The CLI residual ~1/invocation is per-process uow schema-init, not a pool defect; pure-pool ceiling is 0.)

See `docs/CONNECTION_POOLING.md` (design) and `docs/CONNECTION_POOLING_DEPLOYMENT.md` (deploy + validation numbers + orchestrator integration).

Refs gascity#1978.

### DIFF
--- a/cmd/bd/brief_flag_test.go
+++ b/cmd/bd/brief_flag_test.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// TestReadyBriefFlag verifies the --brief seam on `bd ready` (be-yvci): the flag
+// exists with a safe default (off), and setting it maps onto
+// WorkFilter.BriefBodies — the field that selects the body-stripped work-probe
+// projection. This is the seam gascity's supervisor work_query (ga-arn) uses.
+func TestReadyBriefFlag(t *testing.T) {
+	t.Parallel()
+
+	briefFlag := readyCmd.Flags().Lookup("brief")
+	if briefFlag == nil {
+		t.Fatal("--brief flag should exist on readyCmd")
+	}
+	if briefFlag.DefValue != "false" {
+		t.Errorf("--brief default should be 'false', got %q", briefFlag.DefValue)
+	}
+
+	newCmd := func() *cobra.Command {
+		c := &cobra.Command{Use: "probe", Run: func(*cobra.Command, []string) {}}
+		c.Flags().Bool("brief", false, "")
+		return c
+	}
+
+	// --brief set ⇒ WorkFilter.BriefBodies == true
+	withBrief := newCmd()
+	if err := withBrief.Flags().Parse([]string{"--brief"}); err != nil {
+		t.Fatalf("parse --brief: %v", err)
+	}
+	var f types.WorkFilter
+	applyReadyBriefFlag(withBrief, &f)
+	if !f.BriefBodies {
+		t.Error("--brief should set WorkFilter.BriefBodies=true")
+	}
+
+	// default (no --brief) ⇒ BriefBodies stays false (no behavior change)
+	def := newCmd()
+	var f2 types.WorkFilter
+	applyReadyBriefFlag(def, &f2)
+	if f2.BriefBodies {
+		t.Error("absent --brief should leave WorkFilter.BriefBodies=false")
+	}
+}
--- a/cmd/bd/compact.go
+++ b/cmd/bd/compact.go
@@ -81,6 +81,10 @@ Examples:
 	Run: func(_ *cobra.Command, _ []string) {
 		// Block mutating operations in embedded mode; allow --stats, --analyze, --dry-run read-only paths.
 		if !compactStats && !compactAnalyze && !compactDryRun {
+			// S5: compaction rewrites Dolt history and needs exclusive direct
+			// repository access — impossible through the proxy. Read-only paths
+			// (--stats/--analyze/--dry-run) remain serviceable over the routed store.
+			guardUnsupportedInProxiedMode(CapabilityCompaction)
 			if err := requireServerMode("compact"); err != nil {
 				FatalError("%v", err)
 			}
--- a/cmd/bd/db_proxy_child.go
+++ b/cmd/bd/db_proxy_child.go
@@ -71,7 +71,7 @@ not intended to be invoked directly by users.`,
 		// non-empty password; the managed loopback server uses root with no
 		// password.
 		backendPassword := ""
-		if backend == proxy.BackendExternal {
+		if backend == proxy.BackendExternal || backend == proxy.BackendLocalSharedServer {
 			backendPassword = os.Getenv(configfile.ExternalDoltPasswordEnvVar)
 		}
 
@@ -98,10 +98,12 @@ func newDatabaseServer(backend proxy.Backend, rootDir, configPath, logPath, dolt
 	switch backend {
 	case proxy.BackendLocalServer:
 		return server.NewDoltServer(doltBin, rootDir, configPath, logPath, 0)
-	case proxy.BackendExternal:
+	case proxy.BackendExternal, proxy.BackendLocalSharedServer:
+		// The shared backend fronts the managed dolt through the same external
+		// server. The N+1 → 1 collapse comes from pointing every scope at one
+		// shared rootDir (so the parent's spawn-or-reuse yields a single child),
+		// not from a distinct server type.
 		return server.NewExternalDoltServer(external)
-	case proxy.BackendLocalSharedServer:
-		return nil, fmt.Errorf("backend %q: not yet implemented", backend)
 	}
 	return nil, fmt.Errorf("unknown backend %q", backend)
 }
--- a/cmd/bd/db_proxy_child.go
+++ b/cmd/bd/db_proxy_child.go
@@ -29,6 +29,8 @@ var (
 	dbProxyChildExternalTLSCertPath string
 	dbProxyChildExternalTLSKeyPath  string
 	dbProxyChildExternalKeepAlive   time.Duration
+	dbProxyChildPoolSize            int
+	dbProxyChildBackendUser         string
 )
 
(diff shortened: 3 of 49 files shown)


cstar:
### Note: multi-database topologies and the pool key

Flagging an invariant this PR already satisfies, for reviewers running a different topology than the one #1978 measured.

The original measurement was a **1-rig city = one database**. A common multi-agent topology instead runs **one `dolt sql-server` hosting one database per rig** (`hq`, `vw`, `va`, …), each scope's `metadata.json` pinning a distinct `dolt_database` that `bd` selects at connect time.

That makes borrow/reset dangerous if the pool were keyed only per-server: `COM_RESET_CONNECTION` restores the session to the backend's **handshake database**, so a backend opened against `vw` then lent to a client wanting `hq` would resolve unqualified table names against `vw` → silent cross-store reads.

This PR keys the pool by `(capabilities, database)` (`pooledconn.go` builds the key from the client's handshake response; `pool.go` authenticates each backend with `CONNECT_WITH_DB` on `key.db`), so a `vw`-authed backend is never lent to an `hq` client. It relies on `go-sql-driver/mysql` carrying the database in the MySQL handshake when `DBName` is set in the DSN (which `bd` does).

Proven against a real dolt (added in this PR):
- `TestSpike_MultiDatabaseIsolation` — 20 interleaved iterations across two databases through a size-1-per-key pool; each client reads only its own sentinel row, zero cross-read.
- `TestSpike_HandshakeCarriesDatabase` — `SELECT DATABASE()` returns the handshake db on both a fresh and a reused (reset) backend.
- plus unit-level `TestPool_KeyIsolation`.

Single-DB deployments don't hit this; multi-DB-per-server ones are covered by the database-keyed pool.

cstar:
## Follow-up commit: configurable proxy idle-timeout (`a9aefe27`)

Live deployment on a real gascity city (portharbour) surfaced a churn driver the per-invocation benchmark can't see: the **30s proxy idle-timeout**. An orchestrator that touches many scopes sparsely (controller patrol probing dozens of rigs once per interval) starves each per-scope proxy below 30s, so it spawns → serves one bd op → idle-dies → respawns on the next touch. Proxies respawned continuously (every PID different within 35s); ~21–29 new dolt conns/sec — pooling never reached the warm-pool steady state.

This commit adds `IdleTimeoutFromEnv` reading **`BEADS_PROXY_IDLE_TIMEOUT`** (Go duration; `0` disables), mirroring the existing `PoolSizeFromEnv`, wired into the external + local dolt-server UOW providers. Default unchanged (30s) when unset.

**Result** at `BEADS_PROXY_POOL_SIZE=2` + `BEADS_PROXY_IDLE_TIMEOUT=10m` on the same city: proxies stay warm (PIDs stable, no respawn), churn drops to the ~2–3 conns/sec baseline, `Threads_connected` bounded (≤22). Sweet spot is **small pool + long idle-timeout**, not pool-size tuning.

Unit test: `TestIdleTimeoutFromEnv` (unset/blank/unparseable→fallback, valid durations, `0`→disabled).

coffeegoddd:
proxied server mode is not implemented, and we arent taking contributions related to its code at the moment. if you are hitting a specific gascity or beads bug, please file an issue, but this and related proxied server prs are not mergable

cstar:
Thanks @coffeegoddd — I can see it was wired up and then deliberately disabled (`298563044 init.go: disable again`); my only change there was re-enabling the external path (`01c37ca0`).

What I actually built on top of it is a separate **connection-pooling** layer (`internal/storage/dbproxy/proxy`: a MySQL-aware proxy that keeps warm, authenticated backend connections and `COM_RESET_CONNECTION`s between borrowers). It just happens to need proxied-server as its host.

The real goal is killing per-`bd`-invocation connection churn — ~71 new Dolt connections/sec on a live gascity (gastownhall/gascity#1978), most of Dolt's CPU spent on connection setup. Benchmark: 7.02 → 1.02 connections/invocation.

So two genuine questions:
1. Is proxied-server disabled permanently, or being reworked? (i.e. is re-enabling it a dead end I should abandon?)
2. Is there a blessed direction for the churn itself — persistent `bd` session, a `bd` daemon, something else — that I should target instead?

Happy to file the churn as an issue with the numbers and drop the proxied-server re-enable entirely. The pooling proxy is self-contained if it's ever useful as reference.

maphew:
converted to draft because as noted in https://github.com/gastownhall/beads/pull/4303#issuecomment-4622545053 proxied server mode is on hold / not implemented (however @coffeegoddd please see @cstar's follow up questions https://github.com/gastownhall/beads/pull/4303#issuecomment-4623107685)

maphew:

rbriski:
We hit this same Dolt connection churn on a live Gas Town deployment and independently built a very similar session-pooling proxy before finding this PR — happy to consolidate here rather than run a parallel implementation. Offering what we have in case it's useful:

**Production validation on a different topology.** We soaked our pooler on a live multi-rig workspace — an `hq` database plus several product databases on one shared `dolt sql-server` (not the 1-rig city in the description) — deployed as a hybrid: one shared pooler service fronting the server, with existing `bd` clients routed through it by a thin socket wrapper that auto-falls-back to a direct connection if the pooler socket is absent. Results across multiple consecutive steady-state windows:

- **Connection churn: ~557 → ~13–22 new conns/min (~96% collapse)**, sustained.
- **`Threads_connected` pinned ~8**, rising only toward the pool cap under concurrent load — backends pinned, not proportional to client rate.
- Reads ~55–65 ms, writes ~220–260 ms through the pool; a real create+close write probe verified end-to-end each window.
- Pooler stable: single long-lived process, zero respawns; mail and all `bd` ops healthy throughout.

This corroborates the connection-setup-cost thesis on a shared multi-DB server, where the failure mode is one rig's storm wedging the listener for every other database.

**One edge case worth comparing.** Our version terminated the handshake but then ran the command phase through a vitess `Handler` rather than byte-transparent `io.Copy`, specifically because `bd`'s server-mode DSN does not set `interpolateParams` — so go-sql-driver issues server-side `COM_STMT_PREPARE`/`COM_STMT_EXECUTE`. We handled that by interpolating bound params back into SQL. Flagging it in case the byte-transparent path needs prepared-statement coverage (and the capability-flag parity around `QUERY_ATTRIBUTES`/`SESSION_TRACK` you already note). Your `io.Copy` command phase is cleaner if the prepared-statement case is covered or ruled out — happy to compare notes.

Glad to **run this branch on the live town** to validate under real multi-rig load whenever it's helpful. Whatever lands here, we'll consolidate onto it.

PR #4313, opened 2026-06-06
fix(dolt): proxied routed store must not clobber dolt-server.port

bourgois:
**Note**: This fix builds on the connection-pooling work in PR #4303
(\`cstar:feat/connection-pooling\`). The first 8 commits in this diff come from
that PR; the single fix commit \`48118a1e5\` is the change being added. If
you prefer, merge #4303 first and I will rebase this PR onto main to show
only the one-commit fix.

## Problem

\`newProxiedServerRoutedStore\` (introduced in PR #4303) opens a \`ServerMode\`
DoltStore pointed at the db-proxy's ephemeral listener so legacy store-based
commands (\`list\`, \`ready\`, \`update\`, \`--claim\`, \`close\`) work through the proxy.
\`newServerMode\` then persists that listener port into \`.beads/dolt-server.port\`
via \`EnsurePortFile\`, because the connection looks like a localhost server.

That file is the **canonical-server discovery hint**. Writing the proxy's
ephemeral port there means:
- It is clobbered on the next reconcile cycle.
- It goes stale the moment the proxy idle-dies and respawns on a new port.
- Any reader that resolves the endpoint from the file then targets a dead
  listener — manifesting downstream as silent connection failures and a
  fleet-wide claim stall.

Observed in production: port-file churn, proxy-pool leak, and 0 claims
transitioning to \`in_progress\` across all rigs until a supervisor restart.

## Fix

Add \`Config.RoutedThroughProxy\` (bool). Set it in \`newProxiedServerRoutedStore\`.
Gate the port-file persist on it via \`shouldPersistPortFileForConfig\` so only
the canonical local \`dolt sql-server\` records its port — never an ephemeral
proxy listener.

The guard layers on top of the existing \`shouldPersistResolvedPortFile\` check:
a proxied store never persists regardless of host/port values, and an env-override
still suppresses the persist for a canonical server.

## Tests

New pure-Go unit test in \`internal/storage/dolt/port_file_persist_test.go\`:
- \`canonical local server persists\` — happy path unchanged
- \`proxied routed store does not persist\` — the fix
- \`remote host never persists\` — unchanged
- \`nil config is safe\` — defensive nil guard
- \`EnvOverrideSuppresses\` — env-override still wins

All 6 subtests pass. \`go test ./...\` + \`go build ./...\` + \`go vet ./...\` clean.

### DIFF
--- a/cmd/bd/db_proxy_child.go
+++ b/cmd/bd/db_proxy_child.go
@@ -29,6 +29,8 @@ var (
 	dbProxyChildExternalTLSCertPath string
 	dbProxyChildExternalTLSKeyPath  string
 	dbProxyChildExternalKeepAlive   time.Duration
+	dbProxyChildPoolSize            int
+	dbProxyChildBackendUser         string
 )
 
 var dbProxyChildCmd = &cobra.Command{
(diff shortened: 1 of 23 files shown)


maphew:
I convert to draft because proxied server / connection pooling implementation is held back pending https://github.com/gastownhall/beads/pull/4303#issuecomment-4622545053 and  https://github.com/gastownhall/beads/pull/4303#issuecomment-4623107685.

maphew:

bourgois:
Closing this myself — it cannot land here, and the reason is worth recording rather than leaving it to sit in draft.

**The bug it fixes cannot occur upstream.** The guard is gated entirely on `cfg.RoutedThroughProxy`, and that field is set in exactly one place:

```
cmd/bd/uow_factory.go →  RoutedThroughProxy: true
```

inside `newProxiedServerRoutedStore`, which does not exist in this repository — it belongs to the proxied-server routed store carried downstream. So for any build of `main`, `RoutedThroughProxy` is always false and `shouldPersistPortFileForConfig` reduces to the pre-existing `isLocalHost(cfg.ServerHost) && shouldPersistResolvedPortFile()`. The change is a behavioural no-op here.

That also explains why the repro asked for twice never arrived, and I should have worked this out then rather than leaving the request unanswered: the failure needs a store that connects to an ephemeral db-proxy listener, so there is no command sequence that reproduces it against upstream `main`. Apologies for the silence.

@maphew — your call to hold this behind #4303 was right, and more right than it looked: this isn't merely sequenced after the pooling work, it is only meaningful inside it. With the db-proxy work staying downstream, the port-file guard goes with it and is now recorded as permanent fork carry on our side.

Nothing is lost by closing — the fix is already on our fork's `main`. Thanks for the reviews on this and the neighbouring PRs; the 1105-retry and phase-aware-retry re-lands in particular came back better than they went out.

Issue #4051, opened 2026-05-20
Feature Request: bd attachment


MovGP0:
- It should be possible to associate an attachement files to a bead.
- The files should be stored in the `.beads/attachements/<BEADID>/` folder
- when showing a bead, the attachements should be listed
- files might be identified by their hash (similar to git commit hashes) instead of their filename

**Example:**
```sh
# attach a file to a bead
bd attachment add <beadid> '/path/to/fiile'

# list all file attachements
# e.g. hash, filename, MIME type, size in KiB/MiB/GiB
bd attachment list <beadid>

# copy the file attachement to a file or folder
bd attachment copy <beadid> '<filename>' '/target/path/'
bd attachment copy <beadid> '<filename>' '/target/path/new_filename'

# remove/delete the file attachement
bd attachment remove <beadid> '<filename>'
bd attachment remove <beadid> <hash>
```

deanputney:
What would be really nice is if descriptions were able to accept markdown, including file attachments. Then you could write a short doc inside an item and fully explain what's needed.

MovGP0:
> What would be really nice is if descriptions were able to accept markdown

This is already supported, just broken on some terminals. Will be fixed with https://github.com/gastownhall/beads/pull/4406

PR #4316, opened 2026-06-07
Feature/4051

MovGP0:
## What

Adding minimum-viable support for handling file attachements for beads.

## Why

Users usually want to attach files to beads that are useful for reproducing bugs and errors or provide additional context for a featuture or task.

## Verification

Tested the functionality, but review & feedback would be nice.

## Not in scope

Support for file attachement sync for Jira/Linear is not implemented yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,11 @@ named `origin` automatically. Cross-machine sync uses `bd dolt push` and
 for viewers and interchange, not the source of truth or a full database
 backup.
 
+Attachment metadata syncs with the Dolt database, but attachment bytes are
+plain local files under `.beads/attachments`. See
+[docs/ATTACHMENTS.md](docs/ATTACHMENTS.md) before relying on sync or backup for
+attached files.
+
 ### Server Mode
 
 ```bash
@@ -171,6 +176,8 @@ migration instructions.
 for review, migration, and interoperability, but they do not capture Dolt
 branches, commit history, working-set state, or non-issue tables. Use
 `bd backup` or a manual Dolt backup when you need a restorable database backup.
+Attachment bytes under `.beads/attachments` are outside the Dolt database and
+need a separate filesystem backup.
 
 ## 🌐 Community Tools
 
--- a/cmd/bd/attachment.go
+++ b/cmd/bd/attachment.go
@@ -1,0 +1,546 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	attachfs "github.com/steveyegge/beads/internal/attachments"
+	"github.com/steveyegge/beads/internal/types"
+	"github.com/steveyegge/beads/internal/ui"
+)
+
+type attachmentListItem struct {
+	ID               string    `json:"id"`
+	IssueID          string    `json:"issue_id"`
+	HashAlgorithm    string    `json:"hash_algorithm"`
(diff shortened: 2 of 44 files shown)


maphew:

maphew:
First, an apology: the only reply you'd received on this PR was an automated bug-triage template ("couldn't find a concrete reproduction recipe") that misfired onto a feature PR — that was our tooling, not a judgment of your work, and the same misfire hit your other PRs. Sorry. Here is the real review; the length reflects that a new first-class data type in a tracker deserves this much scrutiny, not that the work is being audited.

**What's genuinely good here** — and it's a lot: the write path streams through SHA-256 into a same-directory temp with sync+rename so a partial write never gets a metadata row; `cleanRelPath`/`ensureWithin` path handling rejects traversal in both separator flavors; and most importantly, you did the honest thing about sync — `docs/ATTACHMENTS.md` states plainly that bytes are local-only and `list`/`show` surface absent files as `missing` instead of failing. The CLI shape from #4051 is good and deserves to survive whatever happens next.

**Why this can't merge in its current shape.** The project charter (`engdocs/PROJECT_CHARTER.md`) puts import/export/sync/backup *inside* core scope for beads data. This PR introduces a class of first-class beads data excluded from all four: on any second clone, every attachment row exists and every byte is absent. For a tool whose core promise is "a teammate or fresh agent pulls the tracker and sees the same state," documented-local-only is still the wrong default. Three concrete blockers underneath that:

1. **`bd attachment prune` can destroy the only copy of data.** Reachability is computed from the current Dolt branch's working set, but the bytes live outside Dolt: attach on branch `feat`, `bd checkout main`, `prune` — bytes gone, permanently, no confirmation, no quarantine. A GC that deletes unversioned data using a versioned index is unsound.
2. Migration `0050` now collides with main's `0050_dependencies_deterministic_id` (main is at 0059) — mechanical renumber.
3. ~660 lines target `website/` and `docs/` paths that no longer exist (the Docusaurus site was deleted for Mintlify; internal docs moved to `engdocs/`) — most of this branch's conflict state, and not reasonably your job to redo against a tree that moved underneath you.

**The decision that's actually pending** — and it's queued with the project owner now, not left hanging: *must an attachment's bytes survive `bd dolt push` + clone on another machine?*

- **If yes**: keep your `attachments` table and CLI essentially as designed, and put the bytes in Dolt (blob column or sibling table, size-capped). Dolt then provides sync/versioning/backup for free and `prune`/`fsck` — including blocker 1 — disappear entirely. This deletes more than it adds relative to the current branch, and keeps your schema, CLI, MIME and hashing work intact.
- **If no**: attachments aren't beads data, and the honest minimal version is metadata-only references (`bd attachment add <id> <path-or-url>`), with the Jira/Linear trackers (#4317/#4318) becoming the *primary* home for real bytes — squarely inside the integration boundary.

Either way, the offer: once the call is made, we'll carry the salvageable portion forward against current main **under your attribution** rather than asking you to rebuild 660 lines of docs against a reorganized tree. If you have a preference between the two routings, argue it here — the sync question is the whole ballgame and you've clearly thought about it (you wrote the local-only warnings yourself).

#4317 and #4318 are explicitly parked behind this decision — same review will apply once the base shape settles. (A fuller findings list — wisp routing, LIKE-metachar selector on the destructive remove path, the count-only `bd show` JSON contract, conformance coverage — is written up and will follow the design call; none of it changes the disposition above.)

---
Review provenance: author-clustered maintainer sweep (your four open PRs reviewed together today; #4325 was fix-merged and is heading in). Design review ran against the PR head + current main.

PR #4317, opened 2026-06-07
Feature/4051 Jira

MovGP0:
## What

Adds Jira support for attachements

## Why

Attachements need to be synchronized with 3rd party issue trackers

## Verification

Unit tests implemented against mock API

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,11 @@ named `origin` automatically. Cross-machine sync uses `bd dolt push` and
 for viewers and interchange, not the source of truth or a full database
 backup.
 
+Attachment metadata syncs with the Dolt database, but attachment bytes are
+plain local files under `.beads/attachments`. See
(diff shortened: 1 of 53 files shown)


MovGP0:
Depends on PR #4316

maphew:

maphew:
The attachment-sync work here is parked on a design decision now queued on the base PR: see the full review on #4316 (which also carries an apology owed to you — the earlier auto-comment on this thread was a misfired bug-triage template, not a judgment of the work). Short version: whether attachment *bytes* must ride `bd dolt push`, or live in external trackers with beads holding references, decides the base storage shape — and this PR's tracker-side sync becomes *more* central, not less, in the reference-only routing. Either way the plan preserves your work under your attribution rather than asking for blind rebases. One coherent answer will land on #4316 once the owner call is made.

PR #4323, opened 2026-06-07
test(dolt): add TestNoPushSkipsDoltPush for push guard coverage

quad341:
## What this changes

Adds a direct unit test for the no-push guard at `cmd/bd/dolt.go:247-249`. When `BD_NO_PUSH=true` (the `no-push: true` config key), `bd dolt push` now has explicit coverage asserting:

1. `Push()` is never called on the store — the guard is genuine, not just a print statement.
2. Output contains `"skipping push"` — the user gets a clear signal.

The test mirrors the existing `TestNoPushDoesNotSkipDoltPull` pattern (same `saveAndRestoreGlobals`/`t.Setenv`/`config.ResetForTesting` structure), confirming the no-push guard is push-only and pull is unaffected.

This branch also carries the underlying no-push feature (guard dolt push/pull and strip push from rendered templates) that is pending review in PR #4212. PR #4212's branch has stale merge history; this branch is a clean rebase onto current main with the test added.

## Review notes

- New test: `TestNoPushSkipsDoltPush` at `cmd/bd/dolt_test.go:1506` — the primary deliverable of this bead.
- `minimalPushStore` (a small test double in dolt_test.go) records whether `Push()` was called; no changes to production types.
- The no-push guard itself is at `dolt.go:247-249` (unchanged from PR #4212).
- No config, CLI surface, or behavior changes beyond test coverage.

## Test plan

- [x] `TestNoPushSkipsDoltPush`: PASS — Push() not called, "skipping push" in output
- [x] `TestNoPushDoesNotSkipDoltPull`: PASS — no regression on pull behavior
- [x] Release gate: [`release-gates/be-i9e2-gate.md`](release-gates/be-i9e2-gate.md)

🤖 Deployed by actual-factory

### DIFF
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -413,10 +413,6 @@ variables for authentication.
 Use --remote to pull from a specific named remote instead of the default.
 The remote must already exist (see 'bd dolt remote add').`,
 	Run: func(cmd *cobra.Command, args []string) {
-		if config.GetBool("no-push") {
-			fmt.Println("skipping pull: rig is local-only (no-push: true)")
-			return
-		}
 		ctx := context.Background()
 		st := getStore()
 		if st == nil {
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -336,6 +336,10 @@ uncommitted changes in its working set).
 Use --remote to push to a specific named remote instead of the default.
 The remote must already exist (see 'bd dolt remote add').`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if config.GetBool("no-push") {
+			fmt.Println("skipping push: rig is local-only (no-push: true)")
+			return
+		}
 		ctx := context.Background()
 		st := getStore()
 		if st == nil {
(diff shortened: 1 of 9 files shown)


maphew:
Picked this up in an upstream review sweep as the merge vehicle for the no-push guard (it carries the extra `TestNoPushSkipsDoltPush` coverage). The green checks here were stale, though — cached from before a `main` refactor — so I rebased onto current `main` and fixed the one defect that surfaced. Pushed as `2854c534`.

**The hidden defect:** `main` now routes `bd dolt pull` failure handling through `isConfirmedNoRemote`, which calls `st.ListRemotes` during error classification (`cmd/bd/dolt.go`). `minimalPullStore` embeds a nil `storage.DoltStorage` and overrode only `Pull`, so once merged into current `main`, `TestNoPushDoesNotSkipDoltPull` nil-panics on the new `ListRemotes` call. `mergeable=CLEAN` is textual-only and the checks were cached, so GitHub didn't show it. This is the same drift @quad341 already fixed on #4212 — I applied the equivalent here: `ListRemotes` reporting no remotes, which keeps the simulated failure on the benign no-remote path so the test still proves the guard doesn't short-circuit `bd dolt pull`.

Verified against current `main`: `TestNoPushSkipsDoltPush`, `TestNoPushDoesNotSkipDoltPull`, `TestIsConfirmedNoRemote`, `TestDoltPushPullCommitNeedStore` all pass; build, `gofmt`, and `generate-cli-docs.sh --check` clean.

Since this branch now has both the drift fix and the superset test coverage, it supersedes #4212 (same feature, fewer tests) — noting that there so the two don't both land.

Issue #4324, opened 2026-06-08
Build: Makefile should support MSVC


MovGP0:
In addtion to GCC support, on Windows the MSVC linker should also be supported.

PR #4325, opened 2026-06-08
fix(build) fixing build when using MSVC (#4324)

MovGP0:
## What

Fixes the makefile for Windows

## Why

The project should also build on Windows

## Verification

Works on my machine 😇


Zireael:
Beads gets daily commits merged, but no official build emitted for a month = Is the entire CI pipeline down?
I ended up cloning and making my own build workflow and invoke it from time to time to get the latest build.

maphew:

maphew:
This fix solves a real onboarding trap: Windows contributors with Visual Studio installed hit opaque CGO flag errors because `cl.exe` can't take GCC-style flags, and nothing said so. Your Makefile toolchain discovery (CC override → PATH gcc → windows-gnu clang → known MSYS2/MinGW locations, with an actionable error at the end) plus the honest docs note is exactly the right shape — thank you, and sorry it sat unreviewed this long.

I've just pushed a maintainer merge commit to your branch (`b71bdf162`) resolving the one conflict that had accumulated: `docs/INSTALLING.md` moved to `docs/getting-started/installation.md` on main, so your wording was carried into the new location unchanged. Nothing in your Makefile change needed touching — it merged cleanly. Linux `make build` verified unaffected.

Next step is ours, not yours: once CI comes back green this merges. If you'd rather adjust anything first, say so and we'll hold.

A note on your other three PRs (#4316/#4317/#4318, the attachments feature): a substantive design review is running now and you'll get a real answer on those shortly — they haven't been forgotten either.

PR #4326, opened 2026-06-08
/{.github,docs,go,internal,scripts}: bump driver/dolt version to 2.1.4

coffeegoddd:


julianknutsen:
@mayor, any other dolt-versions we need to update in homebrew, CI, etc to match the new minimum?

PR #4107, opened 2026-05-22
fix(query,ready,list): optimize reads used by gc/gt

coffeegoddd:
This pr optimizes `bd ready` `bd list` and `bd query` for the json read path used by gas city and gas town. Blocked issues are no longer calculated on read, and instead a denormalized `is_blocked` field is maintained on the write path so reads are fast.

A 500MB beads database was used for all benchmarks.

Simple benchmarks from main:

```bash
➜  beads_sample_beads git:(main) ✗ time bd ready --json --limit 0 > old_world_ready.json
bd ready --json --limit 0 > old_world_ready.json  0.17s user 0.14s system 36% cpu 0.854 total
```
```bash
➜  beads_sample_beads git:(main) ✗ time bd list --json --limit 0 > old_world_list.json
bd list --json --limit 0 > old_world_list.json  0.29s user 0.22s system 34% cpu 1.481 total
```
```bash
➜  beads_sample_beads git:(main) ✗ time bd query "ephemeral=true" --json --limit 0 > old_world_query.json
bd query "ephemeral=true" --json --limit 0 > old_world_query.json  0.18s user 0.17s system 34% cpu 1.038 total
```

throughput (completed commands in duration):

```bash
duration: 180s
  bd list:  58
  bd ready: 58
  bd query: 58
  total:    174
```

```bash
duration: 180s, workers: 4
  bd list:  212
  bd ready: 212
  bd query: 212
  total:    636
```

write amplification:

```bash
➜  beads_sample_beads git:(main) ✗ ./write_amp_bench.sh
bd       : bd version 1.0.4 (dev: main@09a38908771a)
config   : WIDTH=4 DEPTH=4 CROSS_BLOCKS=50 SEED=42
tag      : bench-335572-1779475986
log      : /tmp/bd-write-amp-h6ITf9.log

=== Phase A: build tree ===
root: test-o8ho9
layer 1: 4 children
layer 2: 16 children
layer 3: 64 children
layer 4: 256 children
total nodes: 341  (leaves: 256)
cross-tree blocks edges added: 50
Phase A wall: 156.060 s

=== Phase B: cascade updates ===
Phase B wall: 189.836 s

=== Phase C: cascade delete ===
Phase C wall: 0.625 s

=================================================
Phase A (create):   156.060 s
Phase B (updates):  189.836 s
Phase C (delete):     0.625 s
Total:              346.521 s
=================================================
```

Benchmarks on this PR branch:

```bash
[...]

PR #4327, opened 2026-06-08
fix(storage): set updated_at explicitly in is_blocked recompute (GH#4298)

kevglynn:
## Summary

Fixes GH#4298: `bd export` emits `updated_at` with local wall-clock numerals suffixed `Z`, off by the machine's UTC offset, for any issue/wisp touched by the blocked-state recompute. `created_at` on the same row stays correct because it is written explicitly.

Root cause: `RecomputeIsBlockedInTx` (`internal/storage/issueops/blocked_state.go`) was the **only** issueops UPDATE that omitted `updated_at` and relied on the schema's `updated_at DATETIME ... ON UPDATE CURRENT_TIMESTAMP`. Under the embedded Dolt path the session query time is set from local `time.Now()` (`dolthub/driver` `conn.go` → `SetQueryTime(time.Now())`, still present in v2.1.4), so the auto-filled `updated_at` carries local numerals tagged UTC. This breaks UTC-based downstream tooling (sync, "recently changed" queries, conflict resolution).

## Change

- Bind `updated_at = time.Now().UTC()` explicitly in the mark/unmark batched recompute UPDATEs (issues + wisps), matching the existing pattern in `claim.go`, `close.go`, `update.go`, `bulk_ops.go`, and `compaction.go`. This removes the recompute's reliance on `ON UPDATE CURRENT_TIMESTAMP` and the driver's local query time.
- Add `TestIsBlocked_RecomputeWritesUTCUpdatedAt`, which forces a non-UTC local zone and asserts (in SQL via `TIMESTAMPDIFF`) that `updated_at` stays within seconds of `created_at` after the recompute, so a timezone-mislabeled value is caught regardless of host/CI timezone.

## Test plan

```
CGO_ENABLED=1 go test -tags gms_pure_go -run '^TestIsBlocked_' ./internal/storage/dolt/
```

All `TestIsBlocked_*` pass. The change is behavior-preserving for the affected row set (the same rows the recompute already touched via `ON UPDATE`); it only makes the written timestamp an explicit UTC value instead of the driver's local query time.

### DIFF
--- a/internal/storage/dolt/is_blocked_test.go
+++ b/internal/storage/dolt/is_blocked_test.go
@@ -2,7 +2,9 @@ package dolt
 
 import (
 	"context"
+	"database/sql"
 	"testing"
+	"time"
 
 	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/types"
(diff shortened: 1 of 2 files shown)


maphew:

kevglynn:
**Reproduction:**

```bash
# Storage: embedded Dolt

# Setup: create two issues with a blocking dependency
bd init --prefix test
bd create "blocker" -p 1
bd create "blocked" -p 1
bd dep add test-2 test-1

# Trigger: close the blocker (triggers is_blocked recompute)
bd close test-1 --reason "done"

# Observe: bd export | jq '.updated_at' on test-2
# Shows local wall-clock time with Z suffix (e.g., "2026-06-16T17:00:00Z"
# when actual UTC would be "2026-06-16T21:00:00Z" — off by UTC offset)
```

**Expected**: `updated_at` in the recompute path uses explicit `time.Now().UTC()`
**Actual (before fix)**: Relies on Dolt's `ON UPDATE CURRENT_TIMESTAMP` which uses the driver's local `time.Now()` numerals tagged as UTC — off by the machine's UTC offset

Test: `TestIsBlocked_RecomputeWritesExplicitUTC` verifies UTC correctness.

kevglynn:
Closing in favor of a tracking issue. The is_blocked recomputation was rewritten in PR #4107 and the storage layer is undergoing further redesign by @coffeegoddd (Dolt Server v2 / Schema v2 on the 2.0 roadmap). Filed tracking issue to revisit the updated_at concern after the rewrite lands. Per dustin-dolt's guidance on Discord: tracking issues over PRs for areas being actively rewritten.

PR #4328, opened 2026-06-08
feat(claim): make claimable statuses configurable via claim.from-statuses (#4164)

### LINKED ISSUE #4164: claim() refuses any status except 'open' even when custom statuses are configured
## What

\`bd update <id> --claim\` (and \`bd update <id> --status in_progress --claim\` in one call) refuses to act on an issue whose current status is anything other than \`open\` — even when the project has configured custom statuses via \`bd config set status.custom\`. The hard-coded SQL \`WHERE status = 'open'\` in \`internal/storage/issueops/claim.go\` is the root cause.

## Why this matters

Downstream tools (e.g. [factoryskills](https://github.com/nicholasspencer/factoryskills)) configure custom statuses to encode an opinionated lifecycle:

\`\`\`
draft → in_spec → spec_review → ready → in_progress → code_review → closed
\`\`\`

In that lifecycle, \`ready\` is the canonical post-deliberation state where work should be claimable. The natural marshal-loop flow is \`fs forge <id>\` which validates \`status==ready\` then runs \`bd update <id> --status in_progress --claim\`. The combined call fails because \`--claim\` is evaluated against the pre-existing \`ready\` status rather than the about-to-be-set \`in_progress\`.

## Reproduce (bd built from main @ \`f8b9400\`)

\`\`\`bash
bd init --prefix sbox
bd config set status.custom "draft,in_spec,spec_review,ready,code_review,needs_work,recorded"
ID=\$(bd create --type bug --priority 1 --title "claim test" --description "x" | grep -o 'sbox-[a-z0-9]*' | head -1)
bd update \$ID --status ready          # ✓
bd update \$ID --claim                  # ✗ Error claiming \$ID: issue not claimable: status ready
bd update \$ID --status in_progress --claim  # ✗ same error
\`\`\`

Workaround currently in use:
\`\`\`bash
bd update \$ID --status open
bd update \$ID --status in_progress --claim
\`\`\`

## Root cause

\`internal/storage/issueops/claim.go:53,59\` hard-codes the claimable predecessor as \`open\`:

\`\`\`go
// Lines 50-54 (and 56-60 for the StartedAt-already-set branch):
result, err = tx.ExecContext(ctx, fmt.Sprintf(\`
    UPDATE %s
    SET assignee = ?, status = 'in_progress', updated_at = ?, started_at = ?
    WHERE id = ? AND status = 'open' AND (assignee = '' OR assignee IS NULL OR assignee = ?)
\`, issueTable), actor, now, now, id, actor)
\`\`\`

The conditional UPDATE only matches rows where \`status = 'open'\`, so claim is structurally one-status.

## Proposed fix (sketch — happy to defer to maintainers)

A few shapes that would resolve this without breaking existing behavior:

1. **Make the claimable predecessor configurable.** Add \`claim.from-statuses\` to bd config (defaulting to \`open\`), and substitute it into the WHERE clause.
2. **Accept any non-terminal status.** Expand the WHERE to \`status IN ('open', 'ready', 'in_progress', ...)\` — drops the gate to "not closed/blocked/deferred."
3. **Honor an explicit \`--from-status <s>\` flag.** Lets callers narrow or broaden per-call.
4. **Evaluate \`--status\` ordering before \`--claim\`.** If \`--status in_progress\` is also supplied in the same call, apply it first so claim sees \`in_progress\` and the idempotent branch at \`claim.go:87\` kicks in.

Option 1 or 4 is probably the least surprising. Happy to send a PR for whichever direction you prefer.

## Verification

The factoryskills sandbox test in \`/tmp/bd-race-test.sh\` step 7 reproduces this on every run.

kevglynn:
## Summary

`ClaimIssueInTx` hardcoded `WHERE status = 'open'`, so an issue parked in a project's custom claimable status (e.g. `ready`) could never be claimed — `bd update <id> --claim` (and `--status in_progress --claim`) failed for any custom-status lifecycle. Reported in #4164.

This makes the claimable-status set configurable while keeping the default behavior identical:

- New `claim.from-statuses` config key lists **additional** statuses an issue may be claimed from.
- `open` is **always** claimable; it's prepended and de-duplicated, so behavior is unchanged when the key is unset.
- The CAS `UPDATE` now uses `status IN (...)` built from that set (parameterized; no SQL injection surface).
- Registered the `claim.` config prefix so `bd config set claim.from-statuses ...` isn't flagged as an unrecognized key.

### Usage

```sh
# default: only "open" is claimable (unchanged)
bd config set status.custom "ready:active"
bd config set claim.from-statuses "ready"
bd update <id> --claim   # now works when the issue is in "ready"
```

## Test plan

Added `cmd/bd/claim_custom_status_embedded_test.go` (embedded-dolt, gated on `BEADS_TEST_EMBEDDED_DOLT=1`):

- [x] `default_open_only_claimable` — claiming a `ready` issue fails when `claim.from-statuses` is unset (pins backward-compatible default)
- [x] `custom_status_claimable_when_configured` — with `claim.from-statuses=ready`, claim succeeds and transitions to `in_progress`
- [x] `open_still_claimable_with_config` — `open` remains claimable even when the key lists only custom statuses

Existing claim coverage still passes:

```
go test -tags gms_pure_go -run 'TestEmbeddedUpdate/update_claim' ./cmd/bd/   # update_claim, update_claim_already_claimed PASS
go test -tags gms_pure_go -run '^TestEmbeddedClaimFromCustomStatus$' ./cmd/bd/  # PASS
```

Fixes #4164

### DIFF
--- a/cmd/bd/claim_custom_status_embedded_test.go
+++ b/cmd/bd/claim_custom_status_embedded_test.go
@@ -1,0 +1,78 @@
+//go:build cgo
+
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// TestEmbeddedClaimFromCustomStatus is the regression test for
+// gastownhall/beads#4164: claim hardcoded "status = 'open'", so issues parked
+// in a project's custom claimable status (e.g. "ready") could never be claimed.
+// claim.from-statuses makes the claimable set configurable while keeping "open"
+// always claimable (default behavior unchanged).
+func TestEmbeddedClaimFromCustomStatus(t *testing.T) {
+	if os.Getenv("BEADS_TEST_EMBEDDED_DOLT") != "1" {
+		t.Skip("set BEADS_TEST_EMBEDDED_DOLT=1 to run embedded dolt integration tests")
+	}
+	t.Parallel()
+
+	bd := buildEmbeddedBD(t)
+	dir, _, _ := bdInit(t, bd, "--prefix", "claim")
+
+	// Register a custom "ready" status so issues can be parked there.
+	if out, err := bdRunWithFlockRetry(t, bd, dir, "config", "set", "status.custom", "ready:active"); err != nil {
+		t.Fatalf("config set status.custom failed: %v\n%s", err, out)
+	}
+
+	t.Run("default_open_only_claimable", func(t *testing.T) {
+		issue := bdCreate(t, bd, dir, "Default claimable", "--type", "task")
+
+		// Park it in the custom "ready" status.
+		bdUpdate(t, bd, dir, issue.ID, "--status", "ready")
+		if got := bdShow(t, bd, dir, issue.ID); got.Status != types.Status("ready") {
+			t.Fatalf("expected status ready after update, got %s", got.Status)
+		}
+
+		// With claim.from-statuses unset, only "open" is claimable, so claiming
+		// a "ready" issue must fail — this pins the backward-compatible default.
+		out := bdUpdateFail(t, bd, dir, issue.ID, "--claim")
+		if !strings.Contains(out, "claimable") {
+			t.Errorf("expected not-claimable error, got: %s", out)
+		}
+	})
+
+	t.Run("custom_status_claimable_when_configured", func(t *testing.T) {
+		if out, err := bdRunWithFlockRetry(t, bd, dir, "config", "set", "claim.from-statuses", "ready"); err != nil {
+			t.Fatalf("config set claim.from-statuses failed: %v\n%s", err, out)
+		}
+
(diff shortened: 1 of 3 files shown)


kevglynn:
Closing in favor of #4334, which fixes the same issue (#4164) with a cleaner approach.

Both make `ClaimIssueInTx` claim from more than just `open`, but #4334 derives the claimable set from the existing status **category** — `open` plus any custom status whose category is `active` (the same set `bd ready` already treats as actionable) — instead of introducing a new `claim.from-statuses` config key as this PR does. That avoids making users re-declare statuses they already categorized via `status.custom`, keeps claim consistent with `bd ready`, and aligns with the charter's preference for using existing metadata over adding new config surface. #4334 also preserves the GH#3570 anti-steal protection (wip/done/frozen excluded) and adds end-to-end dolt-store tests.

Superseded by #4334.

PR #4329, opened 2026-06-08
fix(github): content-based dedup on push to stop re-PATCHing every issue (#4214)

### LINKED ISSUE #2: RFC: Optional Turso backend for branching, real-time sync, and vector search
**Proposal**  
Keep Beads exactly as-is for the default path (**JSONL in Git as source of truth + local SQLite cache**). Add an **optional** Turso/libSQL backend that unlocks:
- **Database Branching** for safe experiments and big refactors
- **Near-real-time sync** between agents and humans via embedded replicas
- **Native Vector Search** for semantic issue lookup
- **Browser-ready** persistence (WASM/OPFS) for future web UIs/agent workbenches
- **Client-side encryption at rest** (embedded replicas)
- **Concurrent writes** beyond SQLite’s single-writer limit

This preserves Beads’ zero-setup workflow and Git-first philosophy while offering a scalable, collaboration-friendly path when teams (or lots of agents) need it.

---

### Motivation

Beads’ current dual-storage design (JSONL in Git + SQLite cache) is elegant and low-friction. Where it may start to creak is **multi-agent, multi-machine** collaboration and **AI-assisted retrieval**:

- Multiple writers can churn Git and create merge friction.
- Some tasks need **shared state fast** (without a Git merge queue, we live in a fast world now).
- AI/agent workflows benefit from **vector similarity search** over issues/notes/context.
- A future Beads **browser UI** would be much simpler if the DB can run locally in the browser and sync when online.

Turso gives us an opt-in path to these capabilities while retaining SQLite compatibility and the local-first ethos.

---

### Proposal (opt-in; default remains unchanged)

1. **Introduce a storage driver interface**
   - Keep the existing `sqlite+git` path as the default driver.
   - Add a `turso` driver that can be enabled via config/env.

2. **Two connection modes**
   - **Local-first (embedded replica):** keep `.beads/local.db` for fast offline reads/writes; periodically **sync** with a remote Turso primary (read-your-writes semantics).
   - **Remote-only:** connect directly to the Turso database (useful for CI, headless agents, shared runners).

3. **Database Branching as a workflow primitive**
   - Map Beads branch commands to DB branches for safe, isolated change sets:
     - `bd branch create <name>` → create branch from `main`
     - `bd branch switch <name>` → point the driver at that branch
     - `bd branch merge <name>` → controlled merge back to `main`
   - Use branches for **big schema/issue reshapes**; use normal sync for routine edits.

4. **Native Vector Search for issues**
   - Add an `embeddings` table (or a vector column on issues) behind a feature flag.
   - Provide `bd search --similar "text"` when the `turso` driver is active.

5. **Browser-ready**
   - Allow a WASM/OPFS replica for a web UI / agent workbench that runs entirely in the browser and syncs when connectivity returns.

6. **GUI client**
   - Document a one-click connection via **Outerbase Studio** for folks who want a visual client right now.

---

### Migration (optional, one-time)

Projects that want Turso can do:

```bash
# 1) (Optional) Import existing SQLite data into a new Turso DB
turso db import .beads/beads.db

# 2) Configure credentials (local dev or CI)
export TURSO_DATABASE_URL="https://<db>.turso.io"
export TURSO_AUTH_TOKEN="<token>"

# 3) Switch Beads to the turso driver
bd config set storage.driver=turso
# or for new repos:
bd init --driver=turso

### LINKED ISSUE #4214: bd github sync --push-only: per-issue dirty state never clears, every invocation re-pushes the entire backlog (O(N) API calls)
## Symptom

On a beads repo with ~1,633 issues, every invocation of `bd github sync
--push-only` makes ~3,450 GitHub REST API calls and takes ~6 minutes —
**regardless of whether any issue has actually changed locally since the last
successful sync**.

Run on a schedule across multiple hosts sharing one GitHub App installation
token, this drains the installation's primary rate-limit pool (6,250/hr on
this user account) and pegs it indefinitely, causing 403s for all other
consumers of the same token (CI, other automation).

## What's happening

`bd github sync --push-only` (and `bd github push` without explicit IDs)
treats every local issue as needing reconciliation against GitHub. It does
this even when the previous sync completed successfully and nothing has been
written to dolt since.

You can verify the per-issue dirty signal directly without making any API
calls. Pick any synced issue ID:

```text
$ bd github push <id> --dry-run
[dry-run] Would update in GitHub: <title>
Dry run mode - no changes will be made
✓ Pushed 1 issues
```

Actually push it:

```text
$ bd github push <id>            # rc=0, succeeds; PATCHes the GitHub issue
```

Re-run the dry-run immediately:

```text
$ bd github push <id> --dry-run
[dry-run] Would update in GitHub: <title>   # STILL "would update"
```

The per-issue dirty state did not clear after a successful push. Sampled the
first 12 IDs in `.beads/issues.jsonl` (a mix of older and newer, including
freshly-synced ones); **all 12 report `Would update` in `--dry-run`, every
time, perpetually**.

So the unscoped `--push-only` sees N dirty issues and reconciles+pushes each,
producing O(N) API calls every run.

`--dry-run` itself makes 0 REST API calls (verified by accounting against
GitHub's `/rate_limit`, which is exempt from quota and so safe to sample
freely). bd is computing the dirty set from local state — the bug is in that
local state, not in a remote scan.

For comparison, `.beads/export-state.json` *does* track an incremental
cursor (`last_dolt_commit`) that advances correctly for the JSONL export
path, and the JSONL export is incremental as advertised in the 1.0.3
CHANGELOG. The GitHub-sync path appears not to have an equivalent cursor (or
its dirty marker is never updated on successful PATCH).

## Repro

1. A beads repo with at least a few hundred issues, all already synced to
   GitHub.
2. Run `bd github sync --push-only`. Observe O(issues) API calls and minutes
   of runtime, even with nothing changed locally.
3. Pick any synced issue ID. `bd github push <id>` → succeeds. Then
   `bd github push <id> --dry-run` → still reports "Would update".
4. Optional: confirm against `GET /rate_limit` (exempt from quota) that the
   calls are real and consume the primary REST pool.

Tested on bd `1.0.4 (dev)`.

## Distinction from #3623

[#3623](https://github.com/gastownhall/beads/issues/3623) addressed the
*secondary* rate-limit hit when bulk-**creating** new issues
(`POST /issues`), focused on retry/backoff and error reporting. That work is
independent of this one.

This issue is about the **read + PATCH cost on every invocation**, scaling
with backlog size, even when no issue has changed. It hits the **primary**
REST rate limit, not the secondary creation limit, and reproduces on a
steady-state repo with zero new local writes.

## Suggested fix shape

Two complementary improvements; either helps, both together make repeat
pushes essentially free:

1. **Clear per-issue dirty state after a successful push.** A successful
   `bd github push <id>` doesn't update whatever marker `--push-only`
   consults, so the same id remains "dirty" indefinitely. Recording the
   successful round-trip should mark that id clean until the next local
   write.

2. **Track a `last_github_sync` cursor** analogous to `last_dolt_commit` in
   `.beads/export-state.json`. `--push-only` could then derive its push set
   as "dolt changes since `last_github_sync`" rather than scanning every
   issue.

Either way, a no-op `bd github sync --push-only` (nothing has changed since
last sync) should early-exit cheaply, not do an O(N) reconcile.

## Impact

Running beads syncs from multiple machines against one shared GitHub
installation token is a common pattern. With this behavior the per-host cost
grows linearly with backlog size and saturates the installation's hourly
quota long before the backlog grows to a count that would feel problematic
by other measures.

kevglynn:
## Summary

`bd github sync --push-only` (and `bd github push`) re-PATCHed the **entire backlog on every run** — O(N) GitHub REST calls — even when nothing changed locally, draining the shared installation's primary rate-limit pool (reported in #4214).

Root cause: the GitHub sync path never set `engine.PushHooks`, so `doPush` fell back to timestamp-only dedup. Linking (writing `external_ref` after create) and any local edit bump a beads issue's `updated_at` past GitHub's second-granularity `updated_at`, so the timestamp check almost always concludes "local is newer" and PATCHes again. Linear and Jira already wire a `ContentEqual` push hook; GitHub was the gap.

## Changes

- **`internal/github/mapping.go`** — add `PushFieldsEqual(local, remote, config)`, comparing only the fields a push actually mutates: `title`, `body`, `state`, and the label set `BeadsIssueToGitHubFields` would send (order-insensitive, since GitHub doesn't preserve label order).
- **`internal/github/tracker.go`** — add `MappingConfig()` accessor (mirrors Linear's `MappingConfig()`).
- **`cmd/bd/github.go`** — add `buildGitHubPushHooks` and wire `engine.PushHooks` in `runGitHubSync`.
- **`cmd/bd/sync_push_pull.go`** — wire `engine.PushHooks` in `runGitHubPush`.

When an issue's pushable fields already match GitHub, the engine now skips it (`stats.Skipped++`) instead of issuing a redundant PATCH.

## Scope / follow-ups

- This eliminates the redundant **PATCH** storm. Dry-run behavior is unchanged and still makes **0 REST calls**.
- A push-side incremental cursor (analogous to `last_dolt_commit` in `export-state.json`) to also skip the per-issue `FetchIssue` for unchanged issues is a sensible follow-up (issue's suggested fix #2); it can build on this hook.

## Test plan

- [x] `internal/github` unit test `TestPushFieldsEqual` (new) — identical (incl. reordered labels) → equal; title/body/state/label/priority differences → not equal; in_progress status label, closed→state, and non-scoped labels handled.
- [x] `go test -tags gms_pure_go ./internal/github/... ./internal/tracker/...` pass (engine `TestEnginePushWithContentEqual` covers the skip mechanism this hook plugs into).

Fixes #4214

### DIFF
--- a/cmd/bd/github.go
+++ b/cmd/bd/github.go
@@ -419,8 +419,9 @@ func runGitHubSync(cmd *cobra.Command, args []string) error {
 	engine.OnMessage = func(msg string) { _, _ = fmt.Fprintln(out, "  "+msg) }
 	engine.OnWarning = func(msg string) { _, _ = fmt.Fprintf(os.Stderr, "Warning: %s\n", msg) }
 
-	// Set up GitHub-specific pull hooks
+	// Set up GitHub-specific pull and push hooks
 	engine.PullHooks = buildGitHubPullHooks(ctx)
+	engine.PushHooks = buildGitHubPushHooks(gt)
 
 	// Build sync options from CLI flags
 	pull := !githubSyncPushOnly
@@ -475,6 +476,29 @@ func runGitHubSync(cmd *cobra.Command, args []string) error {
 	}
 
 	return nil
+}
+
+// buildGitHubPushHooks creates PushHooks for GitHub-specific push behavior.
+// The ContentEqual hook lets the engine skip issues whose pushable fields
+// already match GitHub, so repeated `github sync --push-only` / `github push`
+// runs don't re-PATCH unchanged issues (gastownhall/beads#4214).
+func buildGitHubPushHooks(gt *github.Tracker) *tracker.PushHooks {
+	config := gt.MappingConfig()
+	if config == nil {
+		config = github.DefaultMappingConfig()
+	}
+	return &tracker.PushHooks{
+		ContentEqual: func(local *types.Issue, remote *tracker.TrackerIssue) bool {
+			if remote == nil {
+				return false
+			}
+			gh, ok := remote.Raw.(*github.Issue)
+			if !ok || gh == nil {
+				return false
+			}
+			return github.PushFieldsEqual(local, gh, config)
+		},
+	}
 }
 
 // buildGitHubPullHooks creates PullHooks for GitHub-specific pull behavior.
--- a/cmd/bd/github.go
+++ b/cmd/bd/github.go
@@ -498,6 +498,13 @@ func buildGitHubPushHooks(gt *github.Tracker) *tracker.PushHooks {
 			}
 			return github.PushFieldsEqual(local, gh, config)
 		},
+		// ContentHash lets the engine skip the per-issue GitHub fetch entirely
+		// when an issue is unchanged since its last push, so a no-op
+		// `github sync --push-only` makes ~zero REST calls instead of one GET
+		// per linked issue (gastownhall/beads#4214).
+		ContentHash: func(local *types.Issue) string {
+			return github.PushContentHash(local, config)
+		},
 	}
 }
 
--- a/cmd/bd/github.go
+++ b/cmd/bd/github.go
@@ -505,6 +505,10 @@ func buildGitHubPushHooks(gt *github.Tracker) *tracker.PushHooks {
 		ContentHash: func(local *types.Issue) string {
 			return github.PushContentHash(local, config)
 		},
+		// TargetScope supplies the host and repository omitted by shorthand refs
+		// such as github:42, so changing GitHub target configuration invalidates
+		// the local no-op cache.
+		TargetScope: gt.PushTargetScope,
 	}
 }
 
--- a/cmd/bd/sync_push_pull.go
+++ b/cmd/bd/sync_push_pull.go
@@ -605,6 +605,7 @@ func runGitHubPush(cmd *cobra.Command, args []string) error {
 	engine := tracker.NewEngine(gt, store, actor)
 	engine.OnMessage = func(msg string) { fmt.Println("  " + msg) }
 	engine.OnWarning = func(msg string) { fmt.Fprintf(os.Stderr, "Warning: %s\n", msg) }
+	engine.PushHooks = buildGitHubPushHooks(gt)
 
 	result, err := engine.Sync(ctx, tracker.SyncOptions{
 		Push:     true,
--- a/internal/github/mapping.go
+++ b/internal/github/mapping.go
@@ -194,6 +194,57 @@ func BeadsIssueToGitHubFields(issue *types.Issue, config *MappingConfig) map[str
 	return fields
 }
 
+// PushFieldsEqual reports whether a GitHub push would be a no-op by comparing
+// only the fields a push can actually mutate: title, body, state, and the
+// label set that BeadsIssueToGitHubFields would send. When these already match
+// the remote issue, the push is redundant and can be skipped — this is what
+// prevents `bd github sync --push-only` from re-PATCHing every issue on every
+// run (gastownhall/beads#4214). Mirrors linear.PushFieldsEqual.
+func PushFieldsEqual(local *types.Issue, remote *Issue, config *MappingConfig) bool {
+	if local == nil || remote == nil {
+		return false
+	}
+	if local.Title != remote.Title {
+		return false
+	}
+	if local.Description != remote.Body {
+		return false
+	}
+
+	desiredState := "open"
+	if local.Status == types.StatusClosed {
+		desiredState = "closed"
+	}
+	if !strings.EqualFold(desiredState, remote.State) {
+		return false
+	}
+
+	// Compare the label set we would push against the remote's current labels.
+	// Both include the scoped type::/priority::/status:: labels plus any
+	// non-scoped labels, so an unchanged issue produces an identical set.
+	desiredLabels, _ := BeadsIssueToGitHubFields(local, config)["labels"].([]string)
+	return labelSetsEqual(desiredLabels, remote.LabelNames())
+}
+
+// labelSetsEqual reports whether a and b contain the same labels, ignoring
+// order (GitHub does not preserve label order across a round-trip).
+func labelSetsEqual(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	counts := make(map[string]int, len(a))
+	for _, l := range a {
+		counts[l]++
+	}
+	for _, l := range b {
+		counts[l]--
+		if counts[l] < 0 {
+			return false
+		}
+	}
+	return true
+}
+
 // priorityToLabel converts beads priority (0-4) to GitHub priority label value.
 func priorityToLabel(priority int) string {
 	switch priority {
--- a/internal/github/mapping.go
+++ b/internal/github/mapping.go
@@ -2,7 +2,12 @@
 package github
 
 import (
+	"crypto/sha256"
+	"encoding/binary"
+	"encoding/hex"
 	"fmt"
+	"sort"
+	"strconv"
 	"strings"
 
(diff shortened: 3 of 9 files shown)


kevglynn:
Follow-up pushed (b89570a): the first commit added the `ContentEqual` hook, which stops the redundant PATCH but still issues one `FetchIssue` GET per linked issue — so a no-op `github sync --push-only` still drains the REST rate limit (the core complaint in #4214).

This commit closes that gap with a `ContentHash` push hook:

- The engine persists each issue's pushed content hash in the dolt-ignored `local_metadata` table (keyed `github.pushhash.<id>`) and consults it **before** `FetchIssue`. Issues unchanged since their last push skip the fetch *and* the update — a clean re-sync now makes ~zero REST calls instead of N GETs.
- The hash is also recorded when `ContentEqual` reports a match, so the first pass after a clone/restart (when `local_metadata` is empty) repopulates hashes via the existing fetch path, and subsequent passes are free.
- Dry-run consults the hash too, so an unchanged issue previews as a skip instead of always printing "Would update".

`local_metadata` is dolt-ignored, so these hashes are clone-local and reset on clone/branch-checkout/server-restart — worst case is one extra fetch+compare pass to repopulate, never a missed update. New coverage: `TestEnginePushWithContentHash` (engine short-circuit + dry-run) and `TestPushContentHash` (hash stability/sensitivity).

maphew:

kevglynn:
**Reproduction:**

```bash
# Storage: embedded Dolt
# Requires: GitHub integration configured (bd github sync)

# Trigger: run push-only sync twice in a row with no local changes
bd github sync --push-only
bd github sync --push-only

# Observe: both runs PATCH every issue to GitHub (O(N) API calls)
# even though nothing changed locally between runs.
# GitHub API rate limit drains rapidly on large backlogs.
```

**Expected**: Second push is a no-op (content hasn't changed)
**Actual (before fix)**: Timestamp-only dedup always concludes "local is newer" because linking writes `external_ref` which bumps `updated_at`. No `PushHooks.ContentEqual` was wired for GitHub (Linear and Jira already had this).

Fix adds `PushFieldsEqual` comparing title, body, state, and labels — matching the existing pattern from Linear/Jira push hooks.

kevglynn:
Rebased onto current `main`. One real conflict in `internal/tracker/engine.go`: `main` had added a warnings-surfacing loop (`for _, w := range created.Warnings { e.warn(...) }`) right after issue creation since this branch was cut, at the same insertion point where this PR's `e.recordPushHash(ctx, issue)` call landed. Both are additive and unrelated — kept both, warnings-surfacing first (matches the existing comment above it about degraded pushes), then the push-hash recording, mirroring the order already used on the update path a few lines below.

Fresh CI running on the rebased head now (`mergeable: true`). Flagging the non-mechanical resolution explicitly since this touches the sync path — happy to walk through it further if useful before merge.

_cursor-sonnet-5-unknown-reasoning on behalf of kevglynn_

maphew:
Cross-note from a review of #5065 (halaprix, "include bead metadata in synced issue bodies"), so whoever lands second isn't surprised.

#5065 renders a managed block into the GitHub issue body — bead ID marker plus dependency/dependent lines — and strips that block back out on pull, so local bead descriptions stay clean. That interacts directly with the dedup here:

- `PushFieldsEqual` compares `local.Description != remote.Body`. After #5065, the local description is the *stripped* text and the remote body carries the appended block, so that comparison is permanently unequal.
- `PushContentHash` hashes `local.Description`, so the stored hash never matches what was actually pushed.

Net effect if both land without reconciliation: the dedup this PR exists to provide is defeated and every sync re-PATCHes every issue — the exact behaviour you're fixing.

The fix is small and belongs to whichever PR lands second: hash and compare the *rendered* body rather than the raw description. There's already precedent for the shape in the Linear path, which compares against `BuildLinearDescription` output at `internal/linear/mapping.go:587`.

No action needed from you right now — #5065 is not merge-ready (it has two round-trip blockers I've asked about there), so this PR is likely to land first. I'm flagging it so that if it does, we know to update these two comparators when #5065 follows, rather than discovering it as a mystery re-PATCH loop later. Same note posted on #5065.

maphew (approved):
Reviewed exact head `04add8b70b4c05b5d3b916eb075b98b487a41628` against current upstream.

I found no correctness issues. The outbound cache is bound to content, external reference, and canonical target scope; focused GitHub/tracker tests and `git diff --check` pass, and the branch merges cleanly with current upstream.

One integration constraint: later-filed #5065 changes the same hook and mapping path to render managed metadata into GitHub bodies. #4329 should land first. #5065 should then derive equality and hashing from the exact rendered outbound body so metadata-only changes cannot be hidden by this cache.

Recommendation: **approve and absorb #4329**, preserving the contributor's attribution.

PR #4332, opened 2026-06-09
feat(init): add --init-if-missing for idempotent init (#3490)

### LINKED ISSUE #3490: bd init: add --init-if-missing flag for idempotent orchestration
## Summary

Add a flag to `bd init` that skips (exit 0) when the workspace is already initialized, instead of aborting with exit 1. Motivated by orchestrator tooling that wants to run `bd init` as part of idempotent project-scaffold workflows.

## Current behavior

First run succeeds; second run aborts with exit 1:

```
$ mkdir /tmp/demo && cd /tmp/demo && git init -q
$ bd init --prefix foo --non-interactive --skip-hooks --skip-agents
# ... initializes successfully
$ bd init --prefix foo --non-interactive --skip-hooks --skip-agents
Error:
⚠ Found existing Dolt database: /tmp/demo/.beads/embeddeddolt/foo
This workspace is already initialized.
  ...
Aborting.
$ echo $?
1
```

The error message is helpful; the exit code forces every caller to pre-check `.beads/embeddeddolt/<prefix>/` existence to distinguish "already initialized" (benign) from real errors (permission, missing git, bad prefix).

## Proposed behavior

**Option A (preferred): new `--init-if-missing` flag**

```
$ bd init --prefix foo --init-if-missing --non-interactive
Skipping init: workspace already initialized at .beads/embeddeddolt/foo
$ echo $?
0
```

Rationale: opt-in, non-breaking, self-documenting at the call site.

**Option B: distinct exit code for "already initialized"**

Keep the current abort behavior, but exit with a stable code (e.g. `2` or `64`) when the failure reason is "already initialized". Orchestrators can then treat that code as benign without parsing stderr.

Either option removes the pre-check dance; Option A is more ergonomic.

## Why not `--force`?

`--force` does the opposite — it re-initializes destructively. The request here is the benign complement: "leave it alone if it already exists."

## Use case

[Kyber](https://github.com/IanGordonOne/kyber) has a project-scaffold workflow that runs ~10 steps including `bd init`, each of which must be idempotent so partial failures are recoverable by re-running. The orchestrator currently checks `existsSync('.beads/embeddeddolt/<prefix>/')` before invoking `bd init`, which works but depends on bd's internal storage layout — a contract the orchestrator shouldn't need to know.

## Non-goals

- Not asking for behavior change on bare `bd init` (keep current protective error).
- Not asking for destructive force-like semantics.
- Not asking for automatic retry on transient errors.

Happy to send a PR if Option A (or B) sounds right to you.

kevglynn:
## Summary

Closes #3490.

`bd init` aborts with exit 1 when the workspace is already initialized. Scaffold/bootstrap scripts that just want "ensure a workspace exists" must therefore pre-`stat` `.beads/embeddeddolt/<prefix>` to distinguish the benign "already initialized" case from a real error — a footgun.

This adds an opt-in **`--init-if-missing`** flag for idempotent init:

- When the workspace is already initialized, init prints `Skipping init: workspace already initialized.` to stderr (suppressed under `--quiet`) and exits **0** instead of aborting.
- Default behavior (flag absent) is **unchanged**: exit 1 with the existing "already initialized … Aborting" message.

This is framed as CLI ergonomics (generic idempotency), not orchestration policy, so it stays within the project charter's core scope.

## Implementation

- The flag is intercepted **only** at the existing `checkExistingBeadsData(prefix)` guard call site in `cmd/bd/init.go`. The detection logic (`checkExistingBeadsDataAt`) is untouched.
- `--force` / `--reinit-local` paths are unaffected (they bypass the guard as before).

## Non-goals

- Not changing the default abort behavior.
- Not auto-repairing/migrating an existing workspace.
- Not touching reinit-local paths.

## Tests

- `TestInitIfMissing` in `cmd/bd/init_test.go` (mirrors `TestInitAlreadyInitialized`): inits a workspace, then re-runs with `--init-if-missing` and asserts `Execute()` returns nil (exit 0) and emits the benign skip message. The default abort path calls `os.Exit(1)` and cannot be exercised in-process (documented limitation already noted in this file).
- Verified passing against a real Dolt container: `--- PASS: TestInitIfMissing`.

## Docs

- `docs/CLI_REFERENCE.md`, website CLI reference pages, and `website/static/llms-full.txt` regenerated via `scripts/generate-cli-docs.sh` + `scripts/generate-llms-full.sh`; `./scripts/generate-cli-docs.sh --check` passes.

## Test plan

- [ ] `bd init --prefix foo` then `bd init --prefix foo --init-if-missing` exits 0 with the skip message
- [ ] `bd init --prefix foo` (no flag) on an initialized workspace still exits 1
- [ ] `--init-if-missing` appears in `bd init --help` and `docs/CLI_REFERENCE.md`

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -372,6 +372,17 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 				// --init-if-missing makes init idempotent: when the workspace
 				// is already initialized, skip and exit 0 instead of aborting.
 				if initIfMissing {
+					// Guard against masking a genuine mismatch: if the caller
+					// explicitly requested a --prefix that does not match the
+					// existing workspace, silently skipping would ignore the
+					// request. Abort as usual in that case. (Skipped when
+					// --database is set, since that overrides prefix-based
+					// naming and is the authoritative database selector.)
+					if cmd.Flags().Changed("prefix") && !cmd.Flags().Changed("database") {
+						if existing := existingWorkspaceDBName(); initIfMissingPrefixMismatch(existing, prefix) {
+							FatalError("workspace already initialized as database %q, but --prefix %q was requested.\nRemove --prefix (or pass a matching value) to reuse the existing workspace.", existing, prefix)
+						}
+					}
 					if !quiet {
 						fmt.Fprintln(os.Stderr, "Skipping init: workspace already initialized.")
 					}
@@ -460,19 +471,12 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 
 		// Normalize prefix before storing it in config or deriving the Dolt
 		// database name. Dots are not valid in issue prefixes and must match the
-		// underscore form used for DoltDatabase/metadata.json.
-		// Leading dots produce invalid names (e.g. ".claude" -> "claude"), and
-		// the trailing hyphen is added automatically during ID generation.
-		prefix = strings.TrimLeft(prefix, ".")
-		prefix = strings.TrimRight(prefix, "-")
-		prefix = strings.ReplaceAll(prefix, ".", "_")
-
-		// Sanitize prefix for use as a MySQL database name.
-		// Directory names like "001" (common in temp dirs) are invalid because
-		// MySQL identifiers must start with a letter or underscore.
-		if len(prefix) > 0 && !((prefix[0] >= 'a' && prefix[0] <= 'z') || (prefix[0] >= 'A' && prefix[0] <= 'Z') || prefix[0] == '_') {
-			prefix = "bd_" + prefix
-		}
+		// underscore form used for DoltDatabase/metadata.json. Leading dots
+		// produce invalid names (e.g. ".claude" -> "claude"), the trailing
+		// hyphen is added automatically during ID generation, and a leading
+		// non-letter is prefixed with "bd_" so the derived MySQL identifier is
+		// valid (directory names like "001" are common in temp dirs).
+		prefix = normalizeIssuePrefix(prefix)
 
 		// Cross-boundary safety (bd-q83 / ADR 0002): check remote state
 		// BEFORE any filesystem side-effects so a refusal exits cleanly.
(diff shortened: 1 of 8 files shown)


maphew:
Reviewed this as part of an upstream review sweep — nice, clean feature, thanks. I pushed one maintainer fix commit (`4828ea4c0`) to the branch for two idempotency-correctness edge cases; details below so you can see what changed.

1. **Operational errors were masked as a benign skip.** `checkExistingBeadsData` returned an error both for the benign "already initialized" case *and* for operational failures (e.g. an unreadable `.beads/embeddeddolt`). With `--init-if-missing`, any error was treated as a successful skip, so a real failure exited 0. The check now tags the benign case (`errWorkspaceAlreadyInitialized`, via a small error type that keeps your user-facing message intact); `--init-if-missing` skips only that case and still aborts on operational errors.

2. **Explicit `--database` mismatch wasn't guarded.** The `--prefix` mismatch guard was good, but `--database` — which is the authoritative database selector later in init — was silently skipped on mismatch. Added `initIfMissingDatabaseMismatch` and check `--database` first (then `--prefix`), so an explicit mismatch aborts rather than quietly reusing a different database.

Tests added alongside yours: a `initIfMissingDatabaseMismatch` table, and a `checkExistingBeadsDataAt` test asserting an existing DB matches the sentinel while an unreadable `embeddeddolt` does not (so it can't be masked).

Validated locally: build, `go vet`, the init tests (`//go:build cgo`, run with `CGO_ENABLED=1 -tags gms_pure_go`), `gofmt`, and `generate-cli-docs.sh --check` all pass. The branch merges cleanly onto `main`. Please sanity-check the commit; happy to adjust if you'd cut it differently.

kevglynn:
Thanks for the review and the fix commit, Matt. Good catches... agree with the sentinel error pattern. 

Reviewed `4828ea4c` locally: the `workspaceExistsError` / `errors.Is` approach cleanly separates benign from operational errors without leaking sentinel text into the user message, and the `--database` mismatch guard mirrors the `--prefix` one correctly. 

Tests look thorough, especially the operational-error-not-masked case with the file-instead-of-dir trick.

Issue #3570, opened 2026-04-28
bd ready --claim is not atomic — two workers can claim the same bead concurrently


amshancodes:
## Summary

In bd 1.0.3, `bd ready --claim` (and equivalently the pattern `bd ready --json --limit=1` followed by `bd update --assignee --status in_progress`) is not atomic across concurrent callers. Two workers polling within the same window can both observe the same ready bead, both claim it, and both begin work. Result: duplicated work, near-identical PRs, wasted compute.

## Repro evidence (production, 2026-04-28)

Gas Town overnight convoy at amshancodes/vixxa-ai produced two PRs from one bead:

- PR #114: `[WS2.01] GlucoseSource protocol + GlucoseManager P0 conformance (vx-6i8.1)` — branch `feature/ws2-glucose-source-protocol`, 5 files / ~355 insertions
- PR #115: `[WS2.01] GlucoseSource protocol + GlucoseManager P0 conformance (vx-6i8.1)` — branch `feature/ws2-sensor-ownership`, 5 files / ~355 insertions, near-identical content

Both PRs target bead `vx-6i8.1`. Two `vixxa/swift-implementer` polecat sessions claimed it within minutes of each other. There is no record of the second claim being rejected.

The convoy was running with 2 implementer pool slots active and standing-order beads polling on a 5-minute LOOP DIRECTIVE. The two implementers' polls hit `bd ready` before either had updated the bead status to `in_progress` — both saw it ready, both claimed.

## Expected

`bd ready --claim` should be atomic: exactly one caller wins the claim, all others see the bead as already-claimed (or get the next ready bead). At minimum, `bd update --status in_progress --assignee X` should fail (or no-op) if the bead is already `in_progress` with a different assignee.

## Suggested fix area

A `SELECT ... FOR UPDATE` (Dolt/MySQL) or atomic compare-and-swap on the status+assignee fields when transitioning `open → in_progress` would close the race. The `bd ready --claim` flag specifically should be a single transaction that filters + claims atomically, not a two-step query+update.

## Why this matters

For human-driven workflows the race is rare (one human, one terminal). For autonomous agent pools (the use-case Gas Town is built around), it manifests in production within hours of running multiple workers against the same standing-order bead. Without atomic claim, agent pools cannot be scaled past 1 worker per bead-pool without external coordination.

cc @yegge — flagging per standard practice for Gas Town pipeline / beads upstream issues.

---

[...]

PR #4334, opened 2026-06-09
fix(claim): honor custom active statuses in bd update --claim (#4164)

### LINKED ISSUE #4164: claim() refuses any status except 'open' even when custom statuses are configured

kevglynn:
## Summary

`bd update <id> --claim` could not claim an issue sitting in a **custom active-category status**, even though `bd ready` surfaces such issues as actionable work. The compare-and-swap UPDATE in `ClaimIssueInTx` hardcoded `status = 'open'`, so a `draft -> ready -> in_progress` lifecycle (where `ready` is a custom `active` status) was un-claimable.

This makes claim consistent with `bd ready`: an issue is claimable from `open` **plus** any configured custom status whose category is `active`.

## What changed

- New helper `ClaimableSourceStatusesInTx` resolves the claimable source set: built-in `open` plus custom statuses in the `active` category (read from the normalized `custom_statuses` table).
- `ClaimIssueInTx` now uses a `status IN (...)` predicate instead of `status = 'open'`.
- `wip` / `done` / `frozen` and unspecified-category custom statuses remain **excluded**, so the GH-3570 anti-steal protection is preserved (an `in_progress`/`blocked` issue — or a custom alias for one — is never silently re-claimable).
- Scope is intentionally limited to the direct `bd update <id> --claim` path (`ClaimIssueInTx`). `ClaimReadyIssueInTx` keeps its existing `StatusOpen` filter to respect `bd ready`'s current non-goals around custom-status selection.

## Tests

New end-to-end dolt-store tests in `internal/storage/dolt/claim_custom_status_test.go`:

- claim **succeeds** from a custom `active` status and transitions to `in_progress`
- claim is **rejected** (`ErrNotClaimable`) from a custom `wip` status, leaving the issue untouched
- built-in `open` stays claimable (regression guard)
- cross-actor steal protection still holds (`ErrAlreadyClaimed`), same-actor re-claim is idempotent

```
go test -tags gms_pure_go -run 'Claim' ./internal/storage/dolt/ ./internal/storage/issueops/ ./internal/storage/embeddeddolt/
ok  	github.com/steveyegge/beads/internal/storage/dolt
```

Fixes #4164


maphew:

kevglynn:
**Reproduction:**

```bash
# Storage: embedded Dolt

# Setup: define a custom active status (e.g., "ready")
# In .beads/config.toml:
#   [[custom_statuses]]
#   name = "ready"
#   category = "active"

bd create "test task" -p 1
bd update bd-1 --status ready   # Move to custom active status

# Trigger: try to claim it
bd update bd-1 --claim
# Error: issue not found or not in claimable state
```

**Expected**: `--claim` works from any status in the `active` category (matching what `bd ready` surfaces)
**Actual (before fix)**: `ClaimIssueInTx` hardcodes `status = 'open'`, rejecting custom active statuses like `ready`, `draft`, etc.

Fix uses `status IN (...)` predicate including `open` + custom statuses with `active` category.

kevglynn:
Bump for review — fixes claim to respect custom active statuses defined in metadata. Light issueops behavioral fix, no storage layer impact.

PR #4335, opened 2026-06-09
fix(prime): inject memories under custom PRIME.md; fix --memories-only (#3941)

kevglynn:
## Summary

`bd prime` mishandled persistent memories when a custom `.beads/PRIME.md` exists:

- With a custom `PRIME.md`, `bd prime` emitted **only** the custom content and silently dropped the persistent memories section that the default template appends.
- `bd prime --memories-only` returned the **full custom `PRIME.md`** instead of just the memories — which defeats memory injection in the PreCompact hook path, the primary consumer of `--memories-only`.

This makes memory injection robust to a custom template, matching the default-template behavior.

## What changed (`cmd/bd/prime.go`)

- `--memories-only` is now handled **first** and returns only the memories section, regardless of any custom `PRIME.md` or `--export`. A custom template can no longer suppress hook memory injection.
- When a custom `PRIME.md` is found, the memories section (`formatMemoriesForPrime(false)`) is **appended** when non-empty — matching what the default template does.
- Extracted `readCustomPrimeContent()` to dedupe the three `PRIME.md` read sites (local `.beads/`, redirected workspace, global `~/.config/beads/`).
- Updated the cobra `Long` help that claimed a custom `PRIME.md` overrides output "entirely".

## Tests (`cmd/bd/prime_gemini_hook_test.go`)

Binary-level tests against the package Dolt harness:

- `TestPrime_CustomPrimeMd_AppendsMemories` — with a stored memory and a custom `PRIME.md`, output contains **both** the custom content and the `Persistent Memories` section.
- `TestPrime_MemoriesOnly_WithCustomPrimeMd` — `--memories-only` returns the memory content and does **not** leak custom `PRIME.md` content.
- Existing verbatim-override tests (`Local`/`Global`/`Redirected`/`Default`) still pass: a fresh workspace has no memories, so nothing is appended.

```
go test -tags gms_pure_go -run 'Prime|Codex' ./cmd/bd/
ok  	github.com/steveyegge/beads/cmd/bd
```

## Non-goals

- Not adding `--no-memories` (that is #3961, sequenced after this).
- Not changing memory storage or adding `--memories-cap`.

Fixes #3941

Made with [Cursor](https://cursor.com)

Issue #3941, opened 2026-05-14
bd prime --memories-only outputs full custom PRIME.md instead of just memories


nerdroid23:
### Description

When a custom `.beads/PRIME.md` file exists, `bd prime` replaces the entire default output — including dropping the persistent memories section that is normally appended. The `--memories-only` flag, which should output *only* the memories, instead outputs the full custom `PRIME.md` content (also without memories).

### Steps to Reproduce

```bash
# 1. Add a memory and confirm it exists
$ bd remember "test memory for reproduction"
$ bd memories # Shows the stored memory ✅

# 2. Confirm default template includes memories
$ bd prime --export | grep "Persistent Memories"
## Persistent Memories    # ✅ present

# 3. Confirm memories appear without a custom PRIME.md
# (skip if you don't have one yet)

# 4. Create a custom PRIME.md
$ bd prime --export > .beads/PRIME.md
# (or use any custom content)

# 5. With custom PRIME.md: memories silently dropped
$ bd prime | grep "Persistent Memories" # ❌ nothing

# 6. --memories-only outputs full custom PRIME.md, not memories
$ bd prime --memories-only | grep "Persistent Memories" # ❌ no actual memories
$ bd prime --memories-only | head -5
# Shows the custom PRIME.md content, not memories
```

### Expected Behavior

- `bd prime` with custom `PRIME.md`: should output custom content plus the persistent memories section appended, same as the default template does.
- `bd prime --memories-only`: should output only the persistent memories section, regardless of whether a custom `PRIME.md` exists.

### Actual Behavior

- `bd prime` with custom `PRIME.md`: outputs custom content only. Memories are silently dropped — no warning, no error.
- `bd prime --memories-only`: completely ignores the flag. Outputs the full custom PRIME.md instead of the stored memories. Zero actual memory content is included.

### Impact

This breaks the primary use case for `--memories-only`: injecting memories in hook contexts (e.g., PreCompact). Users who customize `PRIME.md` lose all persistent memory injection, which defeats the purpose of bd remember.

There's no indication memories are being dropped, so users won't know their agent is operating without context they explicitly saved.

### Environment

- bd version 1.0.4 (Homebrew)
- macOS Tahoe 26.4.1

PR #3779, opened 2026-05-07
feat(cli): expose checkout, merge, and branch -d commands (GH#3680)

kevglynn:
## Summary
- Adds `bd checkout <branch>` to switch Dolt branches
- Adds `bd merge <branch>` to merge a branch into current, with conflict reporting and `--strategy` flag
- Adds `bd branch -d <branch>` to delete branches (with current-branch safety guard)
- All three operations wrap existing storage API methods that were already implemented but had no CLI surface
- Works with both server-mode and embedded-mode Dolt backends
- Full `--json` support for agent use

Closes #3680

## Test plan
- [x] `make build` succeeds
- [x] `go vet` passes
- [x] `bd branch -d` prevents deleting the current branch, calls DeleteBranch
- [x] `bd checkout` calls store.Checkout with JSON output of previous/current branch
- [x] `bd merge` calls store.Merge, reports conflicts with resolution hints
- [x] All commands support `--json` output
- [x] Merge `--strategy ours|theirs` auto-resolves conflicts via ResolveConflicts

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3779"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

PR #4336, opened 2026-06-09
feat(prime): add --no-memories to omit the persistent memories section (#3961)

### LINKED ISSUE #3961: bd prime: add --no-memories flag (or --memories-cap=N) to omit the memories section
## Problem

`bd prime` always emits the full memories section. For workspaces with many
memories this can be very large — measured >150 KB of memories in the prime
output of a real-world memory-heavy workspace. That payload lands in every
SessionStart, every PreCompact, and every subagent dispatch that inherits
the parent's prime output.

## Use cases that motivate this

1. **Subagent dispatches.** A worker spawned to run `bd close <id>` or
   `bd update <id> --claim` rarely needs every tribal memory of the
   workspace; it needs the workflow protocol and the bead state.
2. **Cold-start hot paths.** SessionStart fires `bd prime` and the result
   is injected into the model context. The memories section can dominate
   the payload for memory-heavy workspaces and competes with the agent's
   working context budget.
3. **Compaction.** PreCompact prime + the new context's prime add up.

## Workaround in use

A downstream wrapper hook (in loom — https://github.com/fkberthold/loom)
post-processes `bd prime` stdout and caps the memories section. Measured
~10× reduction in the worst case. The wrapper works via text-pattern
matching of section markers, which is brittle — any future change to the
`bd prime` output format silently breaks the cap.

## Proposed shape

Either of:

- `bd prime --no-memories` — omit the memories section entirely.
- `bd prime --memories-cap=N` — limit memories output to N most recent
  entries (or first N bytes; whichever is easier).

Downstream wrappers would migrate to shell out with the flag and drop
the text-pattern post-processor.

## Acceptance from a downstream POV

- Flag documented in `bd prime --help`.
- `bd prime --no-memories` exits 0 and emits all sections except memories.
- Existing `bd prime` behavior unchanged when the flag is absent.

---

Filed on behalf of downstream tooling that wraps `bd prime` output.
Cross-reference: loom-45i in https://github.com/fkberthold/loom.

kevglynn:
## Summary

Adds `bd prime --no-memories` to omit the persistent memories section. In memory-heavy workspaces that section can dominate SessionStart / PreCompact / subagent context budgets, and wrappers currently strip it with brittle text-pattern post-processors. This caps context cost at the source.

The flag suppresses memory injection at **every** site:
- the custom-`PRIME.md` append (added in #3941),
- the CLI path (`outputCLIContext`),
- the MCP path (`outputMCPContext`).

`--memories-only` takes precedence when both flags are set (it is handled first and returns early), so the dedicated memory-injection path is never silently emptied.

## ⚠️ Stacked on #3941

This branch is **built on top of #4335** (`fix(prime): inject memories under custom PRIME.md; fix --memories-only`), since both touch the same `primeCmd.Run` orchestration. The first commit here is that fix. **Merge #4335 first**; this PR will then show only the `--no-memories` commit. Reviewing just `2f90bfd0e` shows the net change for this PR.

## What changed (`cmd/bd/prime.go`)

- New `primeNoMemories` var + `--no-memories` flag registration.
- Memory injection guarded by `!primeNoMemories` in the custom-PRIME append, `outputCLIContext`, and `outputMCPContext`.
- `Long` help updated; precedence with `--memories-only` documented.

## Tests (`cmd/bd/prime_gemini_hook_test.go`)

Binary-level tests against the package Dolt harness:

- `TestPrime_NoMemories_DefaultPath` — omits memories on the default generated output; a **control run** without the flag proves the memory is otherwise present (signal).
- `TestPrime_NoMemories_CustomPrimeMd` — omits the appended memories under a custom `PRIME.md` while preserving the custom content.
- `TestPrime_NoMemories_MemoriesOnlyWins` — `--memories-only --no-memories` still returns memories.

```
go test -tags gms_pure_go -run 'Prime' ./cmd/bd/
ok  	github.com/steveyegge/beads/cmd/bd
```

## Docs

Regenerated `docs/CLI_REFERENCE.md`, `website/.../prime.md` (all versions), and `website/static/llms-full.txt`. Both `generate-cli-docs.sh --check` and `generate-llms-full.sh --check` pass.

## Non-goals

- Not implementing `--memories-cap=N` (separate).
- No change to default behavior when the flag is absent.

Fixes #3961


maphew (commented):
Thanks for this - `bd prime --no-memories` is a clean, well-motivated addition (fixes #3961) that lets memory-heavy workspaces cap SessionStart/PreCompact/subagent context cost at the source instead of relying on brittle downstream text-pattern stripping. The flag's precedence with `--memories-only` (memories-only wins, handled first, so a custom template can never suppress hook memory injection) is correct and the binary-level tests (`TestPrime_NoMemories_DefaultPath`, `_CustomPrimeMd`, `_MemoriesOnlyWins`) cover the right cases. Docs were regenerated for `bd prime` and pass `--check`.

However, there's a mechanical problem blocking merge:

- **[blocker]** The base commit this PR stacks on (`aed144e77`, nominally #4335) also modifies `cmd/bd/branch.go` to add a `bd branch -d/--delete` flag - 36 lines, identical to the diff already in the separate open PR #3779 ("expose checkout, merge, and branch -d commands"), which you also authored. It isn't mentioned in either PR description, and because only the `bd prime` docs were regenerated, `website/docs/cli-reference/branch.md` is now stale. That's the exact reason `Check doc flags freshness`, `Build Artifacts`, and the required `CI Gate` are red right now - the log shows the drift is entirely in `docs/CLI_REFERENCE.md` + `website/docs/cli-reference/branch.md` for the new `-d` flag. This looks like a rebase artifact: the prime branch appears to have been cut from a local branch that still carried the un-merged #3779 commit rather than from current `upstream/main`.
- **[major]** Please fix this at the source - rebase the #4335 branch onto current `upstream/main` and drop the `branch.go` hunk so it carries only the #3941 prime.go/tests/docs changes, then rebase this branch on the corrected #4335. That should leave #4336 as a single clean commit adding `--no-memories`, with CI green.
- **[nit]** No need to re-add `branch -d` here once it's dropped - #3779 already owns that feature and can land separately whenever it's ready.

Once the stack is rebased and CI is green, this is straightforward to merge - the `--no-memories` logic itself needs no changes from what I can see. Suggested order: rebase #4335 first, confirm CI, then rebase #4336 on top.

Recommendation: merge-fix. Next steps for the contributor: rebase #4335 onto `upstream/main` dropping the stray `branch.go` commit, rebase #4336 on top, confirm `Check doc flags freshness` / `CI Gate` go green, then it should be ready to merge (squash preferred).

maphew:
Correction to my review above: I re-verified against the actual branch heads, and the stray `branch.go` hunk is **not** in #4335 itself. The current #4335 head (`1e4a356d2` + `e18ebbb6c`, pushed 2026-06-29) carries only the prime.go/tests/docs changes and its CI is fully green, so no rebase or hunk-dropping is needed there. The 36-line `branch -d` change (identical to #3779) lives only in this PR's embedded copy of that commit (`aed144e77`), where it is folded directly into the prime-fix commit - likely a stale copy from before #4335 was cleaned up.

Corrected recommendation: leave #4335 alone and simply rebase this branch onto the current #4335 head (or restack it cleanly on `upstream/main` + the two #4335 commits). That drops the stray `branch.go` hunk and should turn `Check doc flags freshness` / `Build Artifacts` / `CI Gate` green. The blocker itself still stands - this PR as currently pushed does carry the #3779 `branch.go` change, and that is exactly what CI is failing on.

kevglynn:
Rebased onto the freshly-rebased #4335, and removed the stray `cmd/bd/branch.go` hunk (branch deletion support from #3779) that had leaked into this PR's history.

The PR now contains only the `--no-memories` feature for `bd prime` — no unrelated diffs. Build and prime tests pass.

PR #4338, opened 2026-06-09
fix(close): update .beads/last-touched on close per documented contract (#3965)

### LINKED ISSUE #3965: bd update no longer touches .beads/last-touched on writes (regression vs pre-1.0 contract)
## Summary

`bd update` (and likely other write subcommands) no longer touches `<workspace>/.beads/last-touched` on bd v1.0.2 writes. The file's content + mtime are now only updated by `bd show` reads, repurposing it as a read-marker.

Downstream consumers that relied on last-touched as a write-marker (file-watch fingerprints, cache validators) now show stale state for any external bd writes.

## bd version

```
bd version 1.0.2 (Homebrew)
```

## Reproducer (6 lines)

```bash
stat -f "%m %z" .beads/last-touched
bd update <issue-id> --add-label test-marker
stat -f "%m %z" .beads/last-touched
# mtime + size IDENTICAL despite the write
bd update <issue-id> --remove-label test-marker
stat -f "%m %z" .beads/last-touched
# still identical
```

For comparison, the same write DOES flip `<workspace>/.beads/dolt/<db>/.dolt/noms/journal.idx` (mtime + size both change), which is what we're switching our downstream consumer to use.

## Expected

`bd update` (and other writes) touch last-touched, OR docs are updated to make clear last-touched is now a read-marker only.

kevglynn:
## Summary

`bd close` documents that a close is one of the operations that updates the **last-touched** issue (see `bd close --help` and `CLI_REFERENCE.md`: *"closes the last touched issue ... from most recent create, update, show, or close operation"*). But `SetLastTouchedID` was only called on the `--claim-next` path — a plain `bd close <id>` left the marker pointing at a stale issue. Downstream write-marker consumers (file-watch fingerprints, cache validators) therefore missed closes.

## Note on the issue title

#3965 is titled around `bd update`, which **already** touches last-touched on `main` (`update.go`). This PR fixes the still-broken **`bd close`** case of the *same documented contract*, and hardens the marker write for all callers. Flagging that explicitly so the scope is clear.

## What changed

- **`cmd/bd/close.go`** — after a successful close, record the first closed ID as last-touched (mirrors `bd update`'s `firstUpdatedID` pattern). A later `--claim-next` still overwrites it with the claimed issue (the newer touch), so that path is unchanged.
- **`cmd/bd/last_touched.go`** — `SetLastTouchedID` now calls `os.Chtimes` after the write, so the marker's mtime **always advances** even when the same ID is rewritten. This directly addresses the reporter's "mtime IDENTICAL" complaint and hardens the write-marker for every caller.

## Tests (`cmd/bd`)

- `TestClose_UpdatesLastTouched` (binary-level): create an issue, seed `last-touched` with a sentinel, run `bd close <id>`, assert the marker now equals the closed ID. **Fails before the fix** — a plain close never rewrote the marker.
- `TestSetLastTouchedIDAdvancesMtime`: rewriting the same ID advances mtime.

```
go test -tags gms_pure_go -run 'Close|LastTouched' ./cmd/bd/
ok  	github.com/steveyegge/beads/cmd/bd
```

## Non-goals

- Not adding last-touched to every other write command (`defer`/`reopen`/`label`/etc.) — that would be a broader change.
- No central post-write hook.

Fixes #3965


maphew:

kevglynn:
**Reproduction:**

```bash
# Storage: embedded Dolt

# Setup:
bd init --prefix test
bd create "some issue" -p 1

# Trigger: close the issue, then check last-touched
bd close test-1 --reason "done"
bd show --last
# Before fix: shows a STALE issue (whatever was last created/updated/shown)
# After fix: shows test-1 (the issue we just closed)
```

**Expected**: `bd close` updates last-touched marker per the documented contract (see `bd close --help`: "closes the last touched issue ... from most recent create, update, show, or close operation")
**Actual (before fix)**: `SetLastTouchedID` only called on the `--claim-next` path — a plain `bd close <id>` leaves the marker stale. Downstream consumers (file-watch fingerprints, cache validators) miss closes.

kevglynn:
Bump for review — small fix ensuring .beads/last-touched gets updated on close. Behavioral correctness fix, no storage layer changes. Aligns with 2.0 roadmap code hygiene goal.

PR #4339, opened 2026-06-09
bd list: support configurable default --limit via list.limit config key

mccraigmccraig:
## Summary

Add a configuration key `list.limit` so users can set a default value for `--limit` via `~/.config/bd/config.yaml` (or `BD_LIST_LIMIT` env var), avoiding the need to type `--limit 0` on every invocation.

## What changed

The `--limit` flag on `bd list` had a hardcoded cobra default of 50 with no viper binding. This follows the same `Changed()` + `config.Get*()` pattern used by `--json`, `--readonly`, `--actor`, and `--dolt-auto-commit`.

### Code
- **`internal/config/config.go`**: added `v.SetDefault("list.limit", 50)` in `Initialize()` — matches the current flag default, so no behavior change for unconfigured setups
- **`cmd/bd/list_input.go`**: in `gatherListInput()`, fall back to `config.GetInt("list.limit")` when `--limit`/`-n` is not explicitly passed. Viper's `AutomaticEnv` (prefix `BD_`, `.`→ `_` replacer) automatically picks up `BD_LIST_LIMIT` from the environment

### Tests (`cmd/bd/list_embedded_test.go`)
- `TestEmbeddedList/list_limit_config_env` — sets `BD_LIST_LIMIT=1`, verifies at most 1 result; verifies `--limit 0` takes precedence
- `TestEmbeddedList/list_limit_config_file` — writes `list.limit: 1` to `~/.config/bd/config.yaml`, same assertions

### Docs
- `docs/CONFIG.md`: added `list.limit` to supported settings table
- `website/docs/reference/configuration.md`: added `list.limit` with flag, env var, and default

## Config usage after this change

```yaml
# ~/.config/bd/config.yaml
list:
  limit: 0
```

Or: `export BD_LIST_LIMIT=0`

The `--limit`/`-n` flag still takes precedence when explicitly passed.

### DIFF
--- a/cmd/bd/list_embedded_test.go
+++ b/cmd/bd/list_embedded_test.go
@@ -293,17 +293,21 @@ func TestEmbeddedList(t *testing.T) {
 		if len(allIssues) <= 1 {
 			t.Errorf("--limit 0 should override BD_LIST_LIMIT=1, got %d", len(allIssues))
 		}
+
+		// --all is also an explicit list request and should override the configured default.
+		allFlagIssues := bdListJSONWithEnv(t, bd, dir, []string{"BD_LIST_LIMIT=1"}, "--all")
+		if len(allFlagIssues) <= 1 {
+			t.Errorf("--all should override BD_LIST_LIMIT=1, got %d", len(allFlagIssues))
+		}
 	})
 
 	t.Run("list_limit_config_file", func(t *testing.T) {
-		// Write config.yaml with list.limit: 1 (HOME=dir so ~/.config/bd/config.yaml).
-		configDir := filepath.Join(dir, ".config", "bd")
-		if err := os.MkdirAll(configDir, 0o755); err != nil {
-			t.Fatalf("mkdir config dir: %v", err)
-		}
-		if err := os.WriteFile(filepath.Join(configDir, "config.yaml"), []byte("list:\n  limit: 1\n"), 0o644); err != nil {
+		// Write project config.yaml with list.limit: 1.
+		configPath := filepath.Join(dir, ".beads", "config.yaml")
+		if err := os.WriteFile(configPath, []byte("list:\n  limit: 1\n"), 0o644); err != nil {
 			t.Fatalf("write config.yaml: %v", err)
 		}
+		defer func() { _ = os.Remove(configPath) }()
 
 		// Config file should limit to at most 1 issue.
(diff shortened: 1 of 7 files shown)


maphew:

mccraigmccraig:
@maphew  there's no bug - this is a feature addition, which allows the default limit for a `bd list`  to be configured

maphew:
Ah, yes, the check for repo sweep set up was too simplistic. I'll redirect to a better queue @mccraigmccraig. Thanks for pushing back.

maphew (commented):
I found two small issues while testing this locally. I am going to fix these in-place on the PR branch since maintainer edits are enabled.

1. `list.limit` is currently ignored for `bd list --json` and other non-TTY output. The code reads `config.GetInt("list.limit")`, but leaves the limit source looking like the default flag, so the later non-TTY branch sets the effective limit back to unlimited. Repro:

   `CGO_ENABLED=1 BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go ./cmd/bd -run '^TestEmbeddedList$' -count=1 -v`

   The new `list_limit_config_env` and `list_limit_config_file` subtests both returned 12 issues instead of at most 1.

2. `list.limit` is documented as config.yaml/tool-level configuration, but the YAML-only routing code does not include the `list.` namespace. Without that, `bd config set list.limit ...` would route to the database path even though `bd list` reads only the Viper/config.yaml value during command setup.

The feature itself fits the existing `bd list` surface and is worth landing after these fixes.

maphew (approved):
Maintainer fix is pushed in place and the focused validation now passes.

Validated:

- `CGO_ENABLED=1 BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go ./cmd/bd -run '^TestEmbeddedList$' -count=1 -v`
- `CGO_ENABLED=1 go test -tags gms_pure_go ./internal/config -run '^TestIsYamlOnlyKey$' -count=1 -v`

The broader local `go test -tags gms_pure_go ./cmd/bd ./internal/config` did not complete within 5 minutes on this Windows worktree; the `internal/config` package also has unrelated local environment failures when run alone here. No GitHub checks are reported for this PR.

maphew:
Thanks @mccraigmccraig. This is merged now.

I preserved the feature and tests, then added a small maintainer fix for non-TTY/config precedence and config.yaml routing before squash-merging.

PR #4340, opened 2026-06-09
fix(context): tolerate non-git scopes in GetRepoContext

nicholasspencer:
## What

`GetRepoContext()` (and therefore `bd context` and every other caller) now falls back to the directory that contains `.beads` when there is no enclosing git repository, instead of failing with `cannot determine repository root`.

## Why

In dolt-server mode bd does not need a git root — it operates on the `.beads` store directly. But a `.beads` store can legitimately live in a directory that is **not itself version-controlled**, e.g. a Gas City / HQ scope that holds a store but is not a git repo. In that scope, `git.GetMainRepoRoot()` fails and `buildRepoContext()` returned a hard error, so `bd context --json` (and anything that resolves repo context) blew up there.

Concretely this blocks orchestrators that shell out to `bd context --json` per scope — the gascity supervisor's `bd_context_agreement` preflight fails at the non-git city scope, taking the native store offline. The redirect/external-`BEADS_DIR` paths are unchanged; only the "no git root at all" case now degrades gracefully to the `.beads` parent.

## Verification

- New test `TestGetRepoContext_NonGitScope` (exercises a `.beads` store in a plain non-git dir; asserts `RepoRoot` falls back to the `.beads` parent, no error, `IsRedirected == false`).
- `CGO_ENABLED=1 go test -tags gms_pure_go ./internal/beads/` — passes.
- `CGO_ENABLED=1 go build -tags gms_pure_go ./cmd/bd/` — builds.
- `gofmt -l` clean.

Verified live: with this fix, `bd context --json` at a non-git city scope returns `backend=dolt, database=hq` instead of erroring, and the gascity supervisor's per-scope `bd context` handshake passes.

### DIFF
--- a/internal/beads/context.go
+++ b/internal/beads/context.go
@@ -128,11 +128,16 @@ func buildRepoContext() (*RepoContext, error) {
 		// Beads dir is in a different repo - use that repo's root
 		repoRoot = repoRootForBeadsDir(beadsDir)
 	} else {
-		// Normal case - find repo root via git
+		// Normal case - find repo root via git.
 		var err error
 		repoRoot, err = git.GetMainRepoRoot()
 		if err != nil {
-			return nil, fmt.Errorf("cannot determine repository root: %w", err)
+			// No enclosing git repository (e.g. a non-git city/HQ scope that
+			// holds a .beads store but is not itself version-controlled). bd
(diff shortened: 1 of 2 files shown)


maphew:

PR #4344, opened 2026-06-09
Isolate prime output tests from live memories

quad341:
## What this changes

`TestOutputContextFunction` now suppresses live stored memories while it checks
the structural output of `bd prime`. That keeps the CLI and MCP stealth-mode
cases deterministic on machines where the local Beads store contains arbitrary
memory text, including command strings that stealth mode is supposed to hide.

This does not change production `bd prime` behavior. It only isolates the test
fixture in `cmd/bd/prime_test.go` so the suite validates the intended output
shape instead of the operator's local memory contents.

## Review notes

- Scope is test-only, plus the release gate checklist.
- `stubNoMemories()` uses the existing `ensureStoreActiveForPrime` test hook and
  returns `context.Canceled`, which follows the current non-timeout empty-memory
  path.
- No CLI flags, config, storage behavior, or memory rendering code changes.

## Test plan

- [x] `make build`
- [x] `go test ./cmd/bd/ -run '^TestOutputContextFunction$' -v`
- [x] `make test`
- [x] Release gate: [`release-gates/cli-stealth-test-isolation-gate.md`](release-gates/cli-stealth-test-isolation-gate.md)

### DIFF
--- a/cmd/bd/prime_test.go
+++ b/cmd/bd/prime_test.go
@@ -11,6 +11,11 @@ import (
 )
 
 func TestOutputContextFunction(t *testing.T) {
+	// Isolate from live store: structural content must not depend on what
+	// memories happen to be stored locally (they can contain any text,
+	// including the git-operation strings that stealth mode rejects).
+	defer stubNoMemories()()
+
 	tests := []struct {
 		name          string
 		mcpMode       bool
@@ -326,6 +331,26 @@ func stubPrimeNoPushConfigured(noPush bool) func() {
 	}
 	return func() {
 		primeNoPushConfigured = original
+	}
+}
+
+// stubNoMemories disables live store access so structural-content tests
+// remain deterministic regardless of local store state. Memories can contain
+// arbitrary text (including git-operation strings rejected by stealth mode),
+// so tests that check prime's structural output must suppress them.
+func stubNoMemories() func() {
+	oldStore := store
+	oldActive := storeActive
+	oldEnsure := ensureStoreActiveForPrime
+	store = nil
+	storeActive = false
+	ensureStoreActiveForPrime = func(_ context.Context) error {
+		return context.Canceled // non-timeout → formatMemoriesForPrime returns ""
+	}
+	return func() {
+		store = oldStore
+		storeActive = oldActive
+		ensureStoreActiveForPrime = oldEnsure
 	}
 }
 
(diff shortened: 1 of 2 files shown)


maphew:

quad341:
Thanks @maphew — and good news for the sweep: this is already fixed on `main`, by your own #4552.

## Repro recipe

- **`bd version`:** any build at this PR's base (`origin/main` @ `9a1c88b` and earlier). It's a **test-harness** leak, not shipped `bd prime` behavior.
- **Storage mode:** embedded Dolt, **`CGO_ENABLED=1`** — the embedded store must be openable, which is exactly why the pure-Go CI build never hits it.
- **OS:** any (observed on Linux).
- **Setup:** run the test from a checkout that is **nested inside a real beads workspace** — i.e. an ancestor directory has a `.beads` store — and that store holds at least one persistent memory whose body contains a stealth-rejected git phrase:
  ```
  bd remember "remember to run git pull before starting"
  ```
  (any memory text containing `git pull` / `git status` / `git commit` will do)
- **Command:**
  ```
  CGO_ENABLED=1 go test ./cmd/bd/ -run '^TestOutputContextFunction$' -v
  ```
- **Expected:** all subtests pass.
- **Actual (pre-fix):** the `CLI Stealth` (and, for a `git pull` memory, `CLI Local-only`) subtest fails with `Unexpected text found: git pull` — or whichever rejected git-op phrase the local memory happens to contain.

## Mechanism

`TestOutputContextFunction` calls `outputPrimeContext` without stubbing the package-global `store`, so `formatMemoriesForPrime` (`cmd/bd/prime.go:426`) sees `store == nil`, lazily opens whatever DB `FindBeadsDir()`'s ancestor-walk finds from the test CWD, and injects that workspace's live persistent-memory text **verbatim** into the prime output. Memory bodies aren't run through stealth's reject filter, so a memory mentioning a git-op phrase trips the `rejectText` assertions. CI checkouts have no ancestor DB, so `store` stays nil, `formatMemoriesForPrime` returns `""`, and the leak never surfaces there.

## Why this PR is now redundant

That is exactly the root cause #4552 documents and fixes: it adds `stubPrimeStoreUnavailable()` to `TestOutputContextFunction` **plus five sibling prime tests** — a strict superset of what this PR does (`stubNoMemories()` on the single test). #4552 landed on `main` on Jul 2, after this branch was cut (Jun 8), which is also why the PR now shows as conflicting on `TestOutputContextFunction`.

So #4344 is superseded by #4552. Please go ahead and close it as superseded — or let me know and I'll close it. Thanks for the repro sweep.

PR #4346, opened 2026-06-09
fix(storage): set updated_at to UTC in all blocked-state recompute paths

### LINKED ISSUE #4298: bd export: updated_at contains local time mislabeled as UTC for issues touched by blocked-state recompute
## Summary

`bd export` emits NDJSON where `updated_at` is rendered in the machine's local wall-clock time but incorrectly suffixed with `Z` (the UTC designator). The value is therefore wrong by the local UTC offset. `created_at` on the same record is correct (true UTC). This corrupts any downstream tool that performs UTC-based comparison of `updated_at` (e.g., sync, "recently changed" queries, conflict resolution).

## Environment

- `bd` version 1.0.5 (Homebrew)
- macOS, machine timezone PDT (UTC−7)
- Default embedded Dolt storage backend

## Steps to reproduce

```bash
bd init --quiet --prefix bd          # fresh temp directory
A=$(bd create "Task A" -p 1 -t task --silent)
B=$(bd create "Task B" -p 2 -t task --silent)
C=$(bd create "Task C" -p 3 -t task --silent)
bd dep add "$B" "$A"                 # B depends on A (A blocks B)
bd dep add "$C" "$A"                 # C depends on A
bd export -o export.jsonl
```

Inspect the exported JSONL. All three issues were created within ~2 real seconds of each other:

```jsonl
{"id":"A", "created_at":"2026-06-03T00:29:44Z", "updated_at":"2026-06-03T00:29:44Z", ...}
{"id":"B", "created_at":"2026-06-03T00:29:45Z", "updated_at":"2026-06-02T17:29:46Z", ...}
{"id":"C", "created_at":"2026-06-03T00:29:46Z", "updated_at":"2026-06-02T17:29:47Z", ...}
```

Issue A (never became blocked) is correct — both timestamps are true UTC. Issues B and C (became blocked via `bd dep add`) have `updated_at` values that are local PDT (17:29) mislabeled with `Z`.

## Expected behavior

`updated_at` is true UTC, consistent with `created_at` and with the actual wall-clock moment of the update.

## Actual behavior

`updated_at` carries local wall-clock numerals with a `Z` suffix, off by the machine's UTC offset, for any issue whose row was touched by the blocked-state recompute path. The corruption correlates exactly with `bd dep add` operations on the dependent issues.

## Root cause analysis

Three layers combine:

1. **Schema** — `internal/storage/schema/migrations/0001_create_issues.up.sql` defines:

   ```sql
   updated_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
   ```

   The `ON UPDATE CURRENT_TIMESTAMP` clause auto-fills `updated_at` on any row UPDATE whose SET clause omits it.

2. **Blocked-state recompute** — The most likely trigger is `RecomputeIsBlockedInTx` in `internal/storage/issueops/blocked_state.go`, which issues `UPDATE issues SET is_blocked = ... WHERE id IN (...)` without an explicit `updated_at` in the SET clause. Empirically, the corruption correlates exactly with issues that gained a dependency via `bd dep add` — issues never run through the recompute (e.g. the blocker) keep correct timestamps. When this UPDATE fires, the `ON UPDATE CURRENT_TIMESTAMP` auto-fill kicks in.

3. **Dolt driver query time** — In the embedded Dolt path, `github.com/dolthub/driver` (`conn.go`) sets the session query time via `queryCtx.SetQueryTime(time.Now())` — local time, not `time.Now().UTC()`. go-mysql-server then constructs a time value carrying local wall-clock numerals but a UTC location tag, which serializes as `<local-time>Z`.

**Why `created_at` is immune:** It is set only on INSERT from an explicit `time.Now().UTC()` in the create path and has no `ON UPDATE` clause.

**Why normal edits are immune:** The regular update path (`internal/storage/issueops/update.go`) always includes `updated_at = ?` bound to `time.Now().UTC()`, which suppresses the `ON UPDATE` auto-fill. Only UPDATEs that omit `updated_at` (like the blocked-state recompute) hit the bug.

## Suggested fix

Two options (not mutually exclusive):

1. **Narrow fix:** Include `updated_at = ?` explicitly (bound to `time.Now().UTC()`) in the SET clause of the blocked-state recompute UPDATE, and audit any other UPDATE statements on the `issues` table that omit it. This matches the pattern already used in `update.go`.

2. **Broader fix:** The `ON UPDATE CURRENT_TIMESTAMP` clause on `updated_at` appears vestigial — every application-level update path already sets `updated_at` explicitly. Removing the clause via a new migration would close the entire class of implicit-update timezone corruption.

The deepest root cause lives upstream in `dolthub/driver` (`SetQueryTime` using local time instead of UTC). You may wish to report it there as well, but beads can fully mitigate on its own side via either option above.

## Notes / Impact

Any tool that compares `updated_at` across issues (or against `created_at`) for ordering, conflict resolution, or change detection will produce incorrect results when any affected issue is in a non-UTC timezone. The magnitude of the error equals the machine's UTC offset. On a machine already in UTC the bug is silent (offset = 0), which likely explains why it escapes CI and most maintainer testing.

ksletmoe-aws:
## What

`bd export` emitted `updated_at` in **local wall-clock time mislabeled with a `Z` (UTC) suffix** for any issue or wisp touched by the blocked-state recompute path, so the value was wrong by the machine's UTC offset. `created_at` on the same record was always correct. This corrupts any downstream tool that does UTC-based comparison of `updated_at` (sync, "recently changed" queries, conflict resolution). Reported in #4298.

This PR sets `updated_at = time.Now().UTC()` **explicitly** in the five affected UPDATEs on both the `issues` and `wisps` tables, matching the pattern already used in `update.go`.

## Why

Three layers combined to produce the corruption:

1. **Schema** — `updated_at` is declared `DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`. Any UPDATE whose SET clause omits `updated_at` triggers the auto-fill.
2. **Blocked-state recompute** — the `is_blocked` toggle UPDATEs omitted `updated_at`:
   - `internal/storage/issueops/blocked_state.go` — mark/unmark issue templates
   - `internal/storage/issueops/blocked_state.go` — mark/unmark wisp templates
   - `internal/storage/issueops/dependencies.go` — direct-blocking source update (parameterized over source table)
3. **Dolt driver** — the embedded driver sets session query time via `SetQueryTime(time.Now())` (local, not `.UTC()`), so `ON UPDATE CURRENT_TIMESTAMP` resolves to local wall-clock numerals carrying a UTC location tag, which serializes as `<local-time>Z`.

`created_at` is immune (INSERT-only, set from an explicit `time.Now().UTC()`, no `ON UPDATE`). Normal edits are immune because `update.go` always binds `updated_at = time.Now().UTC()`. Only the UPDATEs that omitted `updated_at` hit the bug — which is exactly why the corruption correlates with `bd dep add`. On a machine already in UTC the offset is 0, so the bug is silent — which is why it escapes CI and most maintainer testing.

This follows up #1123 (the earlier UTC-normalization fix); the blocked-state recompute path was the remaining gap.

## Scope

- Covers ALL blocked-state recompute paths on both the `issues` and `wisps` tables — the latter matters because NoHistory beads (stored in `wisps` with `ephemeral=0`) are included in `bd export` output.
- Five affected UPDATEs fixed:
  - `blocked_state.go`: `markBlockedTemplateForIssues` (issues mark)
  - `blocked_state.go`: `unmarkBlockedTemplateForIssues` (issues unmark)
  - `blocked_state.go`: `markBlockedTemplateForWisps` (wisps mark)
  - `blocked_state.go`: `unmarkBlockedTemplateForWisps` (wisps unmark)
  - `dependencies.go`: `markDirectBlockingDependencySourceInTx` (direct-blocking, parameterized over source table)
- Every other update path (`update.go`, `claim.go`, `close.go`, `compaction.go`, `bulk_ops.go`) already sets `updated_at` explicitly and is untouched.
- Does **not** remove the `ON UPDATE CURRENT_TIMESTAMP` clause. That is a broader, separate hardening option (would close the whole class of implicit-update corruption) and is better handled on its own, since the blocked-state paths currently rely on that clause to bump the timestamp at all.

## Testing

Two complementary layers of regression coverage:

1. **Per-PR guard (runs on every PR)** — `internal/storage/issueops/blocked_state_updated_at_test.go`: pure `go-sqlmock` white-box tests (`TestRecomputeIsBlockedSetsUpdatedAtUTC`, `TestMarkIsBlockedSetsUpdatedAtUTC`, `TestMarkDirectBlockingDependencySetsUpdatedAtUTC`, `TestMarkDirectBlockingDependencyWispSetsUpdatedAtUTC`) that invoke the fixed functions and assert each UPDATE binds `updated_at = ?` to a `time.Time` whose location is UTC. A custom matcher checks `Location() == time.UTC`, so the test fails if `updated_at` is dropped from the SET clause **or** reverted to local `time.Now()` — deterministically, regardless of the CI machine's timezone. No CGO/Docker/embedded-Dolt; runs in the standard `go test -race -short ./...` of `pr-core`. Verified to fail when the fix is reverted.
2. **Behavioral reproduction (embedded-storage tier)** — `internal/storage/embeddeddolt/blocked_state_tz_test.go` `TestBlockedStateUpdatedAtIsUTC`: forces a non-UTC `time.Local`, triggers the recompute via a real dependency add on the in-process embedded driver, and asserts the stored `updated_at` is true UTC. **Fails before the fix (~7h drift), passes after.** This needs the embedded driver, so it runs in `main.yml` → `test-embedded-storage` (gated by `BEADS_TEST_EMBEDDED_DOLT=1`, merge queue / post-merge), not in the per-PR checks. Run locally with:
> ```bash
> BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go \
>   -run '^TestBlockedStateUpdatedAtIsUTC$' ./internal/storage/embeddeddolt/...
> ```

`golangci-lint run --build-tags=gms_pure_go` (v2.9.0) → 0 issues. `gofmt` / `go vet` clean.

## Upstream note

The deepest root cause is in `github.com/dolthub/driver` (`SetQueryTime(time.Now())` should use UTC). Worth reporting upstream, but beads fully mitigates on its own side here.

Fixes #4298


maphew (commented):
Thanks for the careful root-causing here - the three-layer analysis (schema ON UPDATE clause, the five omitted-updated_at UPDATE sites, and the embedded Dolt driver's local-time query clock) is exactly right, and the sqlmock + embedded-driver regression tests are a genuinely nice pair of coverage. I appreciate the effort.

However, on checking upstream/main I found this exact set of call sites was already fixed on 2026-06-11 by commit `03cdc6c86` ("fix(sync): stop the is_blocked recompute from stamping per-clone wall clock into issues.updated_at", bd-578h9.19) - and it took a different, more conservative approach than this PR.

- **[blocker]** All five templates this PR touches (`blocked_state.go` mark/unmark for issues and wisps, and `dependencies.go`'s direct-blocking fast path) are already modified on main, but instead of `updated_at = time.Now().UTC()` they self-assign `updated_at = updated_at`, which suppresses `ON UPDATE CURRENT_TIMESTAMP` without stamping any new value at all.
- **[major]** That distinction matters: main's commit explains that stamping *any* fresh `updated_at` on a derived-state (`is_blocked`) flip - even a correctly-UTC one - causes independent clones to diverge on `updated_at` for the same converged row, and `issues` isn't in the merge auto-resolve allowlist, so every sync surfaces an operator-level conflict. It also poisons the pull conflict guard, import stale guard, and incremental sync, which all read the bump as a real edit. Merging this PR as-is would reintroduce that regression while fixing the narrower UTC-mislabeling symptom.
- **[minor]** #4298 is still open; it's worth closing with a pointer to `03cdc6c86`, since the exact repro in that issue no longer reproduces on main.
- **[minor]** This PR's new tests (`blocked_state_updated_at_test.go`, `blocked_state_tz_test.go`) assert `updated_at` changes on these paths, while main's own regression tests from the same commit (`blocked_merge_test.go`, `cross_upgrade_merge_test.go`) assert it must *not* change, for the merge-safety reason above.

**Conflict:** This is an ordinary content conflict, not the pre-rewrite no-common-ancestor case - `git merge-base upstream/main <head>` resolves to `9a1c88b63a`, which is a real ancestor of both branches. The conflict is a direct collision: main independently rewrote the same five UPDATE templates for a related but different reason, so a plain rebase would need a design decision, not just a diff replay.

Recommendation: retire. The bug this PR targets is already fixed on main via a design that also closes a merge-safety hole this PR's approach would reopen. Suggested next step for the contributor: close this PR (with thanks and credit for the diagnosis, since the root-cause writeup mirrors what landed - and notably predates it by two days), and separately close or narrow #4298 pointing at `03cdc6c86`. If timezone-correct stamping of `updated_at` is still wanted for some other update path, that'd be a fresh, narrower PR against current main rather than a rebase of this one.

maphew:
Following up on the review from July 5 — apologies that this then sat another three weeks without a resolution; that's on us, not you.

To restate the situation plainly: your root-cause analysis here was exactly right, and it *predates* the fix that landed on main. Commit `03cdc6c86` (merged 2026-06-11) modified the same five UPDATE templates you identified — the mark/unmark templates in `internal/storage/issueops/blocked_state.go` (now lines 179, 223, 285, 329 on main) and the direct-blocking fast path in `internal/storage/issueops/dependencies.go` (lines 373, 388) — but took a different approach: `updated_at = updated_at` self-assignment, which suppresses `ON UPDATE CURRENT_TIMESTAMP` without stamping any new value. That choice was deliberate: stamping a fresh timestamp (even a correctly-UTC one) on a derived-state `is_blocked` flip makes independent clones diverge on `updated_at` for the same converged row, and `issues` isn't in the merge auto-resolve allowlist, so every sync surfaces an operator-level conflict. Main's regression tests (`blocked_merge_test.go`) now assert `updated_at` must *not* change on these paths — the inverse of what this PR's tests assert — so the two designs can't coexist and a rebase wouldn't be a mechanical fix.

Given that, we're closing this PR rather than asking you to rework it. The value you contributed is real and we want it recorded: (1) we'll close #4298 with a pointer to `03cdc6c86` and credit your diagnosis, which nailed the mechanism (embedded Dolt's `SetQueryTime(time.Now())` inheriting `time.Local`) two days before the fix landed; (2) the broader driver-clock concern — local time leaking into *other* persisted timestamps — is live in open PR gastownhall/beads#4730, which normalizes create-path audit timestamps in `dependencies.go`; if you'd like to weigh in there or cut a narrow follow-up against current main for any path still affected, we'd welcome it.

This will stay open for a few days in case you'd like to respond before the close. Thanks again for a genuinely well-diagnosed report and patch — the outcome here is a timing collision, not a quality problem.

maphew:
Superseded by 03cdc6c86 on main, which fixed the same five UPDATE templates with a deliberately divergence-safe design (self-assignment, not fresh stamps); merging this PR would reintroduce the merge-divergence regression its own tests would fight. Author's diagnosis credited in GH#4298 (closed). Thank you!

PR #4347, opened 2026-06-09
feat(list): add --external-ref exact-match filter

### LINKED ISSUE #4341: Feature: `bd list --external-ref` filter for direct lookup by external reference
## Problem

When using beads as cross-session context for AI coding agents (Claude Code in my case), the primary lookup pattern is: "find the bead associated with this GitHub PR branch or Jira ticket." The `external_ref` field is perfect for this — I store refs like `jira-PET-354` or `gh:live-infrastructure/pet-354-irsa-grant`.

However, there's no way to look up a bead by `external_ref` without listing all beads and filtering client-side:

```bash
# What I want:
bd list --external-ref "jira-PET-354"

# What I have to do today:
bd list --all --json --limit 200 | jq '.[] | select(.external_ref == "jira-PET-354")'
```

`bd search --external-contains` exists (great!), but it requires a full-text query term too — it's a filter on search results, not a standalone lookup. For programmatic hook usage, I need the ref lookup without knowing a title keyword.

## Use Case

I maintain a hook stack for Claude Code (`SessionStart`, `PostToolUse`) that surfaces relevant beads contextually:

- **On session start**: check if the current git branch has a matching bead (`external_ref=gh:<repo>/<branch>` or `jira-<TICKET>`)
- **On git push**: check if a bead already exists for this branch before suggesting `bd create`

Both hooks currently do `bd list --all --json --limit 200` and filter client-side by `external_ref`. This works but:
1. Fetches all beads when we need 0-1
2. Scales poorly as the DB grows
3. Can't use `bd search --external-contains` because we don't have a title keyword at that point

## Proposed Solution

Add `--external-ref <value>` as a filter flag on `bd list`:

```bash
bd list --external-ref "jira-PET-354" --json
# Returns matching beads (typically 0 or 1) without fetching the full list
```

And/or allow `bd search --external-contains` to work without a positional text query:

```bash
bd search --external-contains "jira-PET" --status open --json
# Currently requires a non-empty positional query arg even with the filter
```

Either approach solves the problem — the core need is: "give me beads matching this external_ref without requiring a title keyword."

## Context

This comes from running beads as a global shared-context tracker across ~100 parallel Claude Code sessions/worktrees. The hooks that work around this limitation spawn `bd list --all --limit 200` on every session start and every `git push`, then filter in Node.js. With a native flag, each lookup would be a single indexed query instead.

Happy to submit a PR if this aligns with your roadmap — the index likely already exists given `--external-contains` works on the search path.

## Related

- #1295 (closed — added `--external-contains` to search, the substring filter half)
- #2883 (open — full-text search expansion)

mlushpenko:
## What

Add `--external-ref` flag to `bd list` for exact-match lookup by external reference value. Also adds `--external-contains` to `bd list` for parity with `bd search`.

## Why

AI coding agent hooks that track work by external ref (e.g. `jira-PET-354`, `gh:repo/branch`) currently have to `bd list --all --json --limit 200` and filter client-side to find the bead associated with a specific PR or ticket. This is inefficient and scales poorly.

With this flag, the lookup becomes:
```bash
bd list --external-ref "jira-PET-354" --json
```

One indexed query instead of fetching the entire bead list.

## Use case

I run beads as cross-session context for ~100 parallel Claude Code sessions. Two hooks need this:
- **SessionStart**: find the bead for the current git branch (`external_ref=gh:<repo>/<branch>`)
- **PostToolUse (git push)**: check if a bead already exists before suggesting `bd create`

Both currently do the full list+filter workaround.

## Changes

- `cmd/bd/list_input.go`: add `externalRef` and `externalContains` fields + flag reading
- `cmd/bd/list_filter.go`: wire both into `buildListFilter`
- `cmd/bd/list.go`: register `--external-ref` and `--external-contains` flags
- `internal/types/types.go`: add `ExternalRef *string` to `IssueFilter`
- `internal/storage/issueops/filters.go`: add WHERE clause for exact match
- `internal/storage/dolt/transaction.go`: same for Dolt path
- `internal/storage/domain/db/issue_search.go`: add `eqStrPtr` call
- `internal/storage/dolt/queries_test.go`: add test cases for exact match (positive + negative)

## Testing

```
$ CGO_ENABLED=0 go test -tags "gms_pure_go" -short -run "TestSearchIssues_ByExternalRef" ./internal/storage/dolt/
ok  	github.com/steveyegge/beads/internal/storage/dolt	0.816s

$ CGO_ENABLED=0 go test -tags "gms_pure_go" -short ./internal/storage/issueops/
ok  	github.com/steveyegge/beads/internal/storage/issueops	0.534s

$ CGO_ENABLED=0 go test -tags "gms_pure_go" -short ./internal/storage/domain/db/
ok  	github.com/steveyegge/beads/internal/storage/domain/db	0.671s
```

All pass locally (Go 1.26.2, macOS arm64).

Closes #4341


maphew (commented):
Thanks for this - it's a clean, well-scoped addition. `bd list --external-ref <value>` for exact-match lookup (plus `--external-contains` for parity with `bd search`) directly closes #4341 and is exactly the kind of "one indexed query instead of full-list-and-filter" improvement that agent-hook use cases need. Nice test coverage on the Dolt path too (positive + negative exact-match cases).

**Findings**

- **[major] CONFLICTING because of a since-landed refactor, not competing intent.** main is 229 commits ahead of this branch's base. In that window, `internal/storage/domain/db/issue_search.go` and `internal/storage/issueops/filters.go` were consolidated - both now just delegate to a single shared builder, `internal/storage/sqlbuild.BuildIssueFilterClauses` (`internal/storage/sqlbuild/filter.go`). That's exactly where this PR's parallel edits to the two old builders now collide. Good news: I checked, and the exact-match `ExternalRef` filter has *not* landed independently on main - `sqlbuild/filter.go` still only has the `ExternalRefContains` substring clause, so there's no duplicate-effort concern, just a mechanical relocation.
- **[minor] Rebase target is now a single file.** After rebasing, add the one WHERE-clause line (`external_ref = ?` gated on `filter.ExternalRef != nil`) only in `internal/storage/sqlbuild/filter.go`, right next to the existing `ExternalRefContains` LIKE clause (~line 48). The changes to `cmd/bd/list.go`, `list_input.go`, `list_filter.go`, `internal/types/types.go`, `internal/storage/dolt/transaction.go`, and `internal/storage/dolt/queries_test.go` are all outside the refactor's footprint and should apply cleanly.
- **[nit]** New list flags get mirrored into the generated CLI reference (`docs/CLI_REFERENCE.md` via `scripts/generate-cli-docs.sh`), so a follow-up doc regen commit would round this out - not a blocker.

**Conflict and remedy:** this is an ordinary conflict, not a pre-rewrite orphan branch - `git merge-base upstream/main <head>` resolves cleanly to a real common ancestor (`9a1c88b`), so a plain rebase (no `--onto`) is all that's needed:

```bash
git fetch upstream main
git rebase upstream/main
# resolve issue_search.go / filters.go conflicts by moving the
# ExternalRef exact-match clause into internal/storage/sqlbuild/filter.go instead
```

Recommendation: fix-merge. Once rebased onto current main with the exact-match clause relocated into `sqlbuild/filter.go` (and the two now-conflicting per-stack copies dropped), this should be ready to merge - the core design and tests are already solid. Thanks again for the contribution and for filing #4341 with such a clear use case.

mlushpenko:
Rebased onto current main and force-pushed. Conflicts resolved as you described — `ExternalRef` exact-match clause now lives in `internal/storage/sqlbuild/filter.go` only (plus `dolt/transaction.go` which still has its own builder). The old copies in `issueops/filters.go` and `domain/db/issue_search.go` are gone since they delegate to sqlbuild now.

Build + tests pass locally (Go 1.26.2, `CGO_ENABLED=0 -tags gms_pure_go`). Ready for re-review.

mlushpenko:
Hey, just pinging in case my last message got buried — rebased and force-pushed as requested. Hoping not to do another rebase 😅 Let me know if anything else is needed!

PR #4348, opened 2026-06-09
fix(storage): stop guaranteed-empty DOLT_COMMITs flooding the Dolt log on busy coordination DBs

realies:
## Problem

On a busy multi-database coordination deployment (gascity-managed, 5 DBs, perpetual wisp churn), **~99% of all Dolt server log lines** were `error="nothing to commit"` warnings (~0.7–2/sec sustained, peaks 17/sec), with the Dolt sql-server burning 1–3 CPU cores evaluating empty commits. Every single `bd` **write** command emitted one guaranteed-empty `DOLT_COMMIT`.

## Root cause (traced via dolt debug query log on the live system)

Two independent unconditional-commit sites:

1. **`DoltStore.Commit`** (`internal/storage/dolt/store.go`) builds its staging list from `dolt_status` excluding only `config`. **`dolt_ignore`'d tables (`wisp_%`, `local_metadata`) sit permanently dirty in `dolt_status` under wisp churn but can never be staged** — `DOLT_ADD` on them fails and is skipped. So whenever only ignored tables are dirty (i.e. *constantly* on a coordination DB), the list is non-empty, the `len(tables)==0` early-return never fires, and the function issues an empty `DOLT_COMMIT('-m')` that the server rejects. The post-write auto-commit takes this path on **every write command**. The captured query trail of one `bd update`:
   ```
   UPDATE issues … ; INSERT INTO events …
   DOLT_ADD ×2 ; DOLT_COMMIT('-m', …)   -- real commit, succeeds
   COMMIT
   DOLT_ADD ×6 ; DOLT_COMMIT('-m', …)   -- 6 = the perpetually-dirty ignored tables; always empty → server error
   ```

2. **`doltServerTx.Commit`** (`internal/storage/uow/doltserver_tx.go`) issued `CALL DOLT_COMMIT('-Am', ?)` unconditionally with no working-set check and returned the raw error.

## Fix

- `DoltStore.Commit`: filter the `dolt_status` scan with the same `dolt_ignore` predicate as `issueops.HasPendingChanges`, so the existing `len(tables)==0` early-return fires as intended (its comment already said “or dolt_ignore'd” — the implementation just never excluded them).
- `doltServerTx.Commit`: gate on `issueops.HasPendingChanges` (correct for `-Am`, which sweeps the whole working set and respects `dolt_ignore`), and swallow a residual benign nothing-to-commit instead of returning it. + discriminating sqlmock regression test.

Uses only helpers already on `main` (`HasPendingChanges`, `IsNothingToCommitError`) — independent of, and complementary to, #4288 (which guards the selective-staging `-m` helpers).

## Live validation (production coordination DB)

| Metric | Before | After |
|---|---|---|
| Background `nothing to commit` per 15s | 11 | **0** |
| Burst of 20 idempotent no-op writes | 27 errors | **0** |
| Sustained 60s, normal orchestrator load | ~40–50 | **0** |

`go build ./...`, `go vet`, `gofmt` clean; `internal/storage/uow` package tests pass including the real-Dolt integration path.

### DIFF
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -1690,7 +1690,23 @@ func (s *DoltStore) Commit(ctx context.Context, message string) (retErr error) {
 	// GH#2455: Stage all dirty tables EXCEPT config. Query dolt_status for
 	// dirty tables and stage each one individually, skipping config to avoid
 	// sweeping up stale issue_prefix changes from concurrent operations.
-	rows, err := conn.QueryContext(ctx, "SELECT table_name FROM dolt_status")
+	//
+	// Exclude dolt_ignore'd tables in the scan itself (same predicate as
+	// issueops.HasPendingChanges). Ignored tables (wisp_%, local_metadata, …)
+	// can sit PERMANENTLY dirty in dolt_status under wisp churn but can never
+	// be staged — DOLT_ADD on them fails and is skipped below. Without this
+	// exclusion the table list is non-empty whenever only ignored tables are
+	// dirty, so every write command issues a guaranteed-empty DOLT_COMMIT that
+	// the server rejects with "nothing to commit" — flooding the Dolt log and
+	// burning server CPU at orchestrator reconcile cadence (observed ~99% of
+	// all Dolt log lines on a busy coordination DB).
+	rows, err := conn.QueryContext(ctx, `
+		SELECT s.table_name FROM dolt_status s
+		WHERE NOT EXISTS (
+			SELECT 1 FROM dolt_ignore di
+			WHERE di.ignored = 1
+			AND s.table_name LIKE di.pattern
+		)`)
 	if err != nil {
 		// If dolt_status fails, fall back to nothing (rare edge case).
 		return fmt.Errorf("failed to query dolt_status: %w", err)
--- a/internal/storage/uow/doltserver_tx.go
+++ b/internal/storage/uow/doltserver_tx.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 
 	"github.com/steveyegge/beads/internal/storage/domain/db"
+	"github.com/steveyegge/beads/internal/storage/issueops"
 )
 
 type doltServerTx struct {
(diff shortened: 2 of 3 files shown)


maphew:

maphew:
This is an exemplary bug report wrapped in a PR — the captured query trail showing the second `DOLT_ADD ×6; DOLT_COMMIT` after the real commit, the observation that the comment at the `len(tables)==0` early-return already *claimed* to exclude dolt_ignore'd tables while the query never did, and the before/after production numbers make the root cause undeniable. Thank you, and apologies both for the long wait and for the June comment asking for a repro recipe when your body already had a traced root cause and a live-validation table; that reply missed the mark.

**Disposition: we want both fixes; the second needs adapting to code that moved under you.** Verified against current main (c989b6b87):

1. **The store.go fix is still needed and re-targets cleanly.** `Commit` was refactored into `commitWorkingSet` with config-staging modes, but the dolt_status scan is still unfiltered (internal/storage/dolt/store.go:2385), so the ignored-table flood you traced is fully intact. Your exact predicate already exists on main in `CommitPending` (store.go:2582), which makes this an easy, precedented port. One interaction to preserve: with the filter, the empty-tables path fires more often, and in `configIncludeAll` mode that path deliberately routes to `concludeOpenMerge` (store.go:2423) — your change is compatible with that; just keep it intact.
2. **The uow fix needs rework — not because you were wrong, but because main moved.** `doltServerTx.Commit` was rewritten (commit 794ff0790) so a failed `DOLT_COMMIT` rolls back before releasing the pinned connection — returning a session with an open transaction to the pool was itself a bug. Your `!pending -> return nil` skip would release the connection with the `START TRANSACTION` still open, reintroducing that. The guard is still right; the skip path just needs to close the SQL transaction (plain `COMMIT` or rollback) before release. Separately, I'd drop the belt-and-suspenders swallow of nothing-to-commit inside `Commit`: `RunTx` already swallows it at the call site (internal/storage/uow/tx.go:63), and the uow layer deliberately lets that error surface as a lost-update signal (see internal/storage/uow/errors.go and lostupdate_dolt_test.go) — swallowing it one layer down could mask a silently lost write.

Rebasing onto moved main is maintainer work under our current policy, and the doltServerTx integration is the fiddly part. Want to take the re-port yourself with the notes above, or shall we do it on our side with your commits and attribution preserved? Together with 4288 (which covers the selective-staging `-m` helpers this PR doesn't), both are headed for merge.

steveyegge:
Picking this up maintainer-side per the offer above (no rush on your end — if you'd still like to do the re-port yourself, say so here and I'll stand down and hand you the branch). Plan: cherry-pick your commits where they apply cleanly, redo the moved parts per the review notes, attribution preserved throughout (`Co-authored-by` where a commit has to be rewritten). Tracked as bd-ga8p7; the companion #4288 re-port is bd-oibnz and will follow. The superseding PR will reference this one.

steveyegge:
Re-port update, with a happy surprise: **half of this landed independently yesterday.** #5341 (merged 2026-08-06) applied the exact dolt_ignore anti-join predicate to `commitWorkingSet` that your store.go commit pioneered — a cherry-pick of `2eb78b205` onto current main now produces a zero-length diff, and #5341 also added the discriminating test this path needed (`TestCommitToleratesDirtyIgnoredWisp`). Your diagnosis was correct twice over: two contributors converged on the identical fix.

The uow half is now up as **#5401**, re-cut against main's current pinned-connection invariants per the review notes above, with your co-authorship recorded and the original commit cited. #4348 will close automatically when it merges; your production trace and before/after numbers remain the canonical description of the bug.

PR #4141, opened 2026-05-23
fix(cli): skip redundant server-mode write auto-commits

julianknutsen:
## What

This updates CLI write paths to flush pending Dolt commits only when `bd` is running embedded storage. In server/proxied-server mode, claim, assign, update, and close no longer schedule redundant command-level auto-commits; no-history and ephemeral create post-writes also avoid extra server-mode Dolt commits.

## Why

Server-mode writes are already handled through the server storage path, so the old command-level write marker could trigger unnecessary follow-up commits after mutations. Embedded mode still needs explicit pending-write flushing, so this keeps that behavior through a focused helper while avoiding redundant work in server mode.

## Verification

- `./scripts/test.sh -run '^Test(CommitPendingIfEmbedded|ShouldCommitCreatePostWrites)' ./cmd/bd`

## PR Hygiene

- Focused on one behavior fix: redundant server-mode write auto-commits.
- No `.beads/` data or generated artifacts included.
- No documentation update needed; this does not change user-facing CLI flags or output.

Issue #3231, opened 2026-04-13
fix: auto-sanitize hyphenated dolt_database on embedded-mode upgrade (post-#2142 gap)


kevglynn:
## Summary

Projects initialized before #2142 (or via `--database` with hyphens) retain hyphenated `dolt_database` names in `metadata.json`. When upgrading to bd 1.0.0 — which defaults to embedded mode — embedded Dolt rejects these names with an unhelpful `invalid database name` error. All beads become inaccessible until the user manually edits `metadata.json`.

This is a remaining gap from the #2142 fix, which only addressed **new** `bd init` runs.

### How it happens
1. User has project `my-project` initialized before #2142, or used `--database my-project` (which passes `ValidateDatabaseName` since it allows hyphens)
2. `metadata.json` contains `"dolt_database": "my-project"`
3. User upgrades to bd 1.0.0 (embedded mode default)
4. Every `bd` command fails: `invalid database name: "my-project"`

### Gaps in current code

| Component | Issue |
|-----------|-------|
| `newDoltStoreFromConfig` | Passes hyphenated name directly to embedded engine — no sanitization |
| `embeddeddolt` error | Says `invalid database name` with no guidance on how to fix it |
| `bd doctor` | Hyphen check only runs in **server mode** (`RunServerHealthChecks`); embedded-default projects skip it entirely |
| `ValidateDatabaseName` | Allows hyphens (`[a-zA-Z0-9_\-]`), but embedded mode rejects them — inconsistent |
| `checkMetadataConfigValues` | Does not check `dolt_database` for embedded-incompatible characters |

### Impact

Any pre-#2142 project in a hyphenated directory (very common: `my-project`, `cool-app`, `jtbot-core`) becomes completely unusable after upgrading to embedded mode. The workaround — manually adding `"dolt_mode": "server"` or editing `dolt_database` — is undiscoverable.

### 1. Auto-sanitize on embedded store open (`store_factory.go`)
When `newDoltStoreFromConfig` is about to open an embedded store with a hyphenated database name:
- Sanitize hyphens → underscores
- Rename the existing database if it exists with the old name
- Persist the sanitized name back to `metadata.json`
- Log the migration for visibility

### 2. Better error in embedded mode (`embeddeddolt/open.go`, `store.go`)
If identifier validation fails and the name contains hyphens, add actionable guidance:
```
invalid database name "my-project": hyphens are not allowed in embedded mode.
[...]

PR #4350, opened 2026-06-10
fix(mol bond): route operands to the target's database

### LINKED ISSUE #4714: bd mol bond ignores prefix routing: routed-prefix issue ID not found (works in bd show)
### Summary

`bd mol bond <formula> <id>` fails to resolve an issue ID that lives in a **prefix-routed** database, erroring with

```
'<id>' not found (not an issue ID or formula name)
```

while `bd show <id>` resolves the exact same ID from the exact same directory. This is the same shape as #3608 (`bd close` failed while `show`/`update` worked) and consistent with #4647: prefix routing appears to have been added command-by-command, and `mol bond` was not included.

This makes Gas Town's default `gt sling <bead> <rig>` path fail whenever the bead lives in the rig's routed store, because `gt sling` auto-applies `mol-polecat-work` via `bd mol bond`. `gt sling --dry-run` does **not** catch it, because dry-run never executes the bond.

### Environment

- `bd version 1.0.3 (1b2dd2cb)`
- Dolt **server** mode (`dolt_mode: server`, shared server on one port), multiple databases
- `routing.mode: "explicit"`, `.beads/routes.jsonl`:
  ```
  {"prefix":"hq-","path":"."}
  {"prefix":"tef-","path":"tef/mayor/rig"}
  ```
  i.e. prefix `tef-` routes to a different database than the local (`hq-`) one.

### Reproduction

All commands run from the town root (the dir whose local DB is the `hq-` store):

```console
$ bd show tef-a87 --json
[ { "id": "tef-a87", "title": "…", … } ]          # resolves — routing honored

$ bd mol bond mol-polecat-work tef-a87 --json --ephemeral --var issue=tef-a87
{ "error": "'tef-a87' not found (not an issue ID or formula name)", "schema_version": 1 }

# Control: same bond against a LOCAL (hq-) bead gets past ID resolution to variable validation
$ bd mol bond mol-polecat-work hq-5ej --json --ephemeral --var issue=hq-5ej
{ "error": "bonding: missing required variables: base_branch, … (use --var)", "schema_version": 1 }
```

The contrast in the last two commands isolates it: with a local-prefix bead, resolution succeeds and bonding proceeds to variable validation; with a routed-prefix bead, resolution fails outright. So the formula loads fine — it's the **target issue** resolution that ignores `routes.jsonl`.

### Expected

`bd mol bond` should resolve the target issue ID through the same prefix-routing path as `bd show`/`bd update`, so a routed-prefix bead can be bonded to a formula from the town root.

### Notes

- There is no cwd from which the bond succeeds: from the town root the formula is found but the routed bead isn't; from inside the rig dir the bead is local but the formula (which lives in the town's `.beads/formulas`) isn't found.
- Downstream impact: `gt sling <bead> <rig>` (Gas Town) fails on its default path. Workaround is `gt sling … --hook-raw-bead` (skips the formula/bond), but that spawns a polecat whose `gt prime --hook` then reports no hooked work.
- Not fixed in v1.1.0 as far as the changelog shows (routing fixes there are enumerated per-command: `edit`/`reopen`/`delete`, `show`/`graph`/`blocked`, `dep add/rm/blocks/tree`; `mol bond` is not among them).

j-s-au:
## Why

`bd mol bond` resolved operands only in the active database. In a centralized-Dolt Gas Town workspace, `bd show <rig-bead>` followed `routes.jsonl`, but `bd mol bond mol-polecat-work <rig-bead>` failed with `not found`, causing `gt sling` to roll back the polecat.

This is a real, independently reproduced routing-parity bug. It was first diagnosed and validated end-to-end by Johan Snyman in this PR and gastownhall/gastown#4220, then independently reproduced and modernized by Rody Vilchez in #4714 and #4720.

Closes #4714. Addresses Bug 1 in gastownhall/gastown#4220.

## What changed

This branch consolidates both contributions:

- preserves Johan's original diagnosis, live Gas Town validation, one-work-store design, and original authored commit;
- preserves Rody's current-main write-intent routing, issue-first formula ambiguity handling, routed dry-run behavior, cross-store policy, changelog, tests, and original authored commit;
- rebases the combined work onto current `main` and adds maintainer fixes for store identity and safe store opening.

The final flow has two phases:

1. Discover both operands through read-only routing, record their logical database homes, reject contributor auto-routes and cross-store pairs, and use the same validation for `--dry-run`.
2. Close discovery handles, reopen only the single accepted home writable, verify it is still the same logical database, reload both issue operands there, then perform the bond.

This prevents unsupported cross-rig commands from triggering writable-open migrations in either foreign project. Matched-route configuration, migration-gate, permission, and lock failures are also propagated instead of being mislabeled as `not found`.

## Attribution

- Johan Snyman (`j-s-au`): original commit `86be4d7f6`, production diagnosis, gastownhall/gastown#4220 reproduction, live end-to-end verification, and work-store/test design.
- Rody Vilchez (`R0SEWT`): carried commit `2ef514122`, #4714 reproduction, current routing primitives, write-intent/dry-run split, issue-first behavior, cross-store policy, changelog, and regression coverage.

Both original authored commits remain in the branch history.

## Verification

Passed locally:

```text
go vet -tags gms_pure_go ./cmd/bd/...
go test -tags gms_pure_go ./cmd/bd -run 'TestMolBondPrefixRouting|TestResolveMolBondOperandRouting' -count=1
BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go ./cmd/bd -run TestEmbeddedMolBondRoutesFormulaMutationToTarget -count=1
```

Coverage includes:

- formula + explicit-prefix-routed issue mutating only the target database;
- same-rig issue pairs succeeding through one writable home;
- true cross-rig pairs rejected during read-only discovery;
- cross-rig dry-run rejection without advancing either database;
- contributor auto-route mutation refusal;
- issue IDs that resemble formula names retaining issue-first semantics;
- matched-route failures remaining actionable;
- writable-reopen identity drift aborting before mutation.

The full `make test` suite passed locally for exact head `bcafe5bdd889eaaf97d231275a3fed85f10f615c`. Two independent agent reviews report no remaining blockers. This nontrivial PR still requires substantive human review before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   every remote-ahead case, as before
   ([#4259](https://github.com/gastownhall/beads/issues/4259)).
 
+### Fixed
+
+- **`bd mol bond <A> <B>` ignored prefix routing** — operand resolution now uses
+  the same routing fallback as `bd show`/`bd update`/`bd close`, so a bead that
+  lives in a routed store (a prefix-routed rig via `routes.jsonl`, or the
+  contributor planning store) resolves instead of failing with `'<id>' not found
+  (not an issue ID or formula name)`. Previously `bd mol bond` resolved operands
+  against the local store only, so `gt sling <bead> <rig>` — which cooks the
+  default formula and bonds it to a rig-routed bead — died at `bd mol bond` even
+  though `bd show <bead>` resolved the same ID. Resolution is write-intent: a
+  prefix-routed target opens writable and the bond commits where the bead lives.
+  A bond whose two operands resolve to different stores is rejected rather than
+  written to the wrong database. Same routing-parity fix as
+  [#3608](https://github.com/gastownhall/beads/issues/3608) for `bd close`.
+  Fixes [#4714](https://github.com/gastownhall/beads/issues/4714).
+
 ## [1.1.0] - 2026-07-04
 
 First stable release of the 1.1.0 line. It consolidates everything from
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,7 +104,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   A bond whose two operands resolve to different stores is rejected rather than
   written to the wrong database. Same routing-parity fix as
   [#3608](https://github.com/gastownhall/beads/issues/3608) for `bd close`.
-  Fixes [#4714](https://github.com/gastownhall/beads/issues/4714).
+  Consolidates the complementary work in
+  [#4350](https://github.com/gastownhall/beads/pull/4350) and
+  [#4720](https://github.com/gastownhall/beads/pull/4720), following the live
+  Gas Town reproduction in
+  [gastown#4220](https://github.com/gastownhall/gastown/issues/4220). Fixes
+  [#4714](https://github.com/gastownhall/beads/issues/4714).
 
 ## [1.1.0] - 2026-07-04
 
--- a/cmd/bd/dep.go
+++ b/cmd/bd/dep.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -623,7 +622,7 @@ func validateBulkDepEdges(ctx context.Context, edges []bulkDepEdge) ([]bulkDepEd
 		current.Cleanups = append(current.Cleanups, fromCleanup)
 		current.IssueID = fromID
 		current.Store = fromStore
-		current.StoreKey = dependencyStoreKey(fromStore)
+		current.StoreKey = storeIdentityKey(fromStore)
 
 		if strings.HasPrefix(edge.DependsOnID, "external:") {
 			if err := validateExternalRef(edge.DependsOnID); err != nil {
@@ -672,18 +671,6 @@ func validateBulkDepEdges(ctx context.Context, edges []bulkDepEdge) ([]bulkDepEd
 
 func bulkDepValidationError(errs []string) error {
 	return fmt.Errorf("bulk dependency validation failed:\n  %s", strings.Join(errs, "\n  "))
-}
-
-func dependencyStoreKey(s storage.DoltStorage) string {
-	if locator, ok := storage.UnwrapStore(s).(storage.StoreLocator); ok {
-		if cliDir := strings.TrimSpace(locator.CLIDir()); cliDir != "" {
-			return "cli:" + filepath.Clean(cliDir)
-		}
-		if path := strings.TrimSpace(locator.Path()); path != "" {
-			return "path:" + filepath.Clean(path)
-		}
-	}
-	return fmt.Sprintf("instance:%p", s)
 }
 
 var depListCmd = &cobra.Command{
--- a/cmd/bd/mol_bond.go
+++ b/cmd/bd/mol_bond.go
@@ -216,9 +216,9 @@ func runMolBond(cmd *cobra.Command, args []string) error {
 		if routeErr != nil {
 			return HandleErrorRespectJSON("%v", routeErr)
 		}
-		if rr.MutationForbidden {
+		if routeErr := verifyMolBondWritableHome(rr, targetStoreKey); routeErr != nil {
 			rr.Close()
-			return HandleErrorRespectJSON("cannot bond issue %s: contributor auto-routing forbids mutation; run the bond from the project that owns the issue", targetID)
+			return HandleErrorRespectJSON("%v", routeErr)
 		}
 		activeStore = rr.Store
 		closeActiveStore = rr.Close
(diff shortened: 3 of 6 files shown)


maphew (commented):
Thanks for tracking down the `bd mol bond` cross-database routing gap - the diagnosis is solid and the value is clear: this is the fix that unblocks `gt sling` for rigs in a centralized-Dolt Gas Town install (gastownhall/gastown#4220), and the regression test structure (modeled on `TestResolveCloseTargets`) is the right shape for this kind of routing fix.

**Findings**

- **[blocker]** `resolveBondStore` (cmd/bd/mol_bond.go) resolves the target via `resolveAndGetIssueWithRouting`. For a prefix-routed target (the town-root -> rig scenario this PR is fixing), that function opens the routed store *read-only* (`resolveViaPrefixRoutingWithAccess(..., writable=false)` -> `newReadOnlyStoreFromConfig` -> `embeddeddolt.OpenReadOnly`). `internal/storage/embeddeddolt/store.go`'s `withConn` rejects any commit against such a store with `errReadOnly`. `runMolBond` then runs the actual bond write (`bondMolMol`/`bondProtoMol`/etc.) against that same `workStore` - so the operand now resolves, but the write that used to fail with "not found" will now fail with a read-only storage error instead. Every mutating command elsewhere in the tree (`assign.go`, `comment.go`, `human.go`, `delete.go`, `note.go`, `heartbeat.go`, `edit.go`, `update.go`, `priority.go`, `reopen.go`, `tag.go`) uses `resolveAndGetIssueForMutation` for exactly this reason, and `close.go`'s `resolveCloseTargets` - the function this PR's test says it's modeled on - explicitly opens prefix-routed targets writable (`resolveViaPrefixRoutingWithAccess(ctx, id, true)`, "a prefix-routed target opens writable so the close commits on the target head (#4141)"). `resolveBondStore` needs the same write-intent path. Worth also extending `TestResolveBondStore` to assert an actual write lands in the routed store, since the current cases only check which store reference comes back, which is why this didn't surface in testing.
- **[minor]** The reported CONFLICTING state is an ordinary, easily-resolved conflict, not the pre-rewrite orphan-branch case - `git merge-base upstream/main <head>` finds a real, if 219-commits-stale, ancestor. The only conflicting file is `cmd/bd/mol_bond.go`, where `d7d7a3460` (RunE + metrics refactor) touched the same lines this PR does. A normal `git rebase upstream/main` should resolve it cleanly; no `--onto` needed.

Once the operand resolution is switched to the write-intent routing path (mirroring `resolveCloseTargets`) and the test covers an actual write, and the branch is rebased onto current `main`, this should be good to land - the underlying diagnosis and approach are right, it just needs to route writes, not just reads, to the target database.

Recommendation: fix-merge. Next steps for @j-s-au: (1) swap `resolveAndGetIssueWithRouting` for `resolveAndGetIssueForMutation` (or replicate `resolveCloseTargets`'s explicit writable=true prefix-routing branch) in `resolveBondStore`; (2) add a test case that performs a real bond against a routed target and asserts the wisp lands in the target DB; (3) rebase onto current `upstream/main` to pick up the `d7d7a3460` RunE refactor. Happy to take another look once those are in.

maphew:
Correction to the blocker in my review above, after re-verifying against the actual PR head rather than `main`. The read-only chain I described (`resolveViaPrefixRoutingWithAccess(..., writable=false)` -> `newReadOnlyStoreFromConfig` -> `embeddeddolt.OpenReadOnly`) only exists on current `main`; on this branch's merge-base (`99c70904f`, before #3231/#4141 landed read-only routed opens), `newReadOnlyStoreFromConfig` opens embedded stores writable via `embeddeddolt.Open`, and the server-mode `ReadOnly` flag only skips schema init and pull auto-commit rather than rejecting writes. So the PR as submitted routes writes correctly, and @j-s-au's end-to-end verification is legitimate - my "which is why this didn't surface in testing" remark was wrong, and I apologize for implying the testing was inadequate.

The finding still matters, but as a rebase hazard, not a defect in the submitted code: the mandatory rebase onto current `main` (to resolve the `d7d7a3460` conflict) will pull in the read-only routed opens, at which point bond writes through `resolveAndGetIssueWithRouting` will fail. Recommendation unchanged in substance: after rebasing, switch `resolveBondStore` to the write-intent path (`resolveAndGetIssueForMutation`, which only exists post-rebase) and add the write-assertion test case.

maphew:
Maintainer consolidation update: I rebased this contributor branch onto current `main`, preserved Johan's original authored commit, and carried Rody's #4720 commit onto the same branch with its original authorship intact.

The combined implementation now:

- discovers and validates both operand homes read-only before any writable foreign-store open;
- rejects cross-rig and contributor-auto-routed mutations consistently in normal and dry-run modes;
- reopens only one accepted target writable and verifies its logical identity before mutation;
- preserves issue-first formula resolution and propagates matched-route failures;
- includes explicit-prefix server-mode tests and an embedded end-to-end formula-to-routed-issue mutation test.

Targeted tests and `go vet` pass. Full `make test` is queued for exact head `bcafe5bdd889eaaf97d231275a3fed85f10f615c`. Independent reviews found no remaining blockers, but this remains pending substantive human review and the full-suite result.

maphew:
Full local verification has now passed for the exact published head:

```text
head: bcafe5bdd889eaaf97d231275a3fed85f10f615c
command: make test
result: exit=0
```

The PR body has been updated from “queued” to “passed.” No agent merge will be attempted; substantive human review remains required for this nontrivial routing/storage-open change.

maphew:
An update that reverses what this thread was told on July 12, with the reasoning in the open: back then the plan was for this PR to be the surviving branch for #4714, with R0SEWT's work carried onto it. Since then the situation inverted — this branch is now ~425 commits behind main with real content conflicts in `mol_bond.go`/`routed.go` (its green CI predates all of that), while #4720 was kept current through two review rounds and now sits 19 commits behind, conflict-free, with the routing blockers verified fixed. Landing cost decides it: **#4720 becomes the surviving branch.**

What survives from here — because two parts of this PR's design are better than what #4720 has, and they're coming along: the read-only-discovery-then-writable-reopen phasing (so a bond that gets rejected never writable-opens a foreign rig during resolution — the exact hazard your design existed to prevent) and the matched-route failure propagation (a route that fails on config/lock reports the real error, not "not found"). Both get ported onto #4720 with your attribution preserved in the commit trailers and PR body, alongside the embedded end-to-end test if it transplants cleanly.

Thank you for the original diagnosis of #4714 and for a design that was, on the operand-safety axis, ahead of both later branches. This PR will close once the ports land on #4720 — if you'd rather do the porting yourself on that branch, say so there and it's yours.

maphew:
Superseded by #4720, the canonical fix for #4714; unique embedded e2e test coverage is being ported with attribution (see disposition comment 2026-07-26). Thanks for the two-phase open design — it survives in the merged line.

PR #4237, opened 2026-05-28
docs: align agent maintainer conventions

maphew:
## What

Align agent-facing maintainer guidance across the upstream beads docs:

- document `Agent-Signature:` and GitHub comment/review signing conventions
- add `scripts/gh-body-lint` for Markdown bodies posted through `gh --body-file`
- add final PR-state reread guidance to maintainer PR rules
- refresh the checked-in managed `AGENTS.md` marker to the current version/profile/hash format and keep the generated template aligned

## Why

Agent-maintained PR work now crosses Codex, Claude, and orchestration repos. The existing upstream instructions already cover project scope, storage boundaries, contributor protection, and generated bd integration blocks. This PR carries over the remaining portable conventions from the mybd coordination repo without moving mybd-specific coordination policy into beads core.

The goal is to make maintainer actions auditable and less drift-prone while keeping generated AGENTS.md content refresh-safe.

Concrete cases this is meant to prevent:

- A maintainer review comment says "looks good" but gives no clue which agent/runtime produced it. With the signing convention, the rendered comment ends with something like `_codex-gpt-5.5-medium on behalf of matt wilkie_`, while the matching commit can carry `Agent-Signature: codex-gpt-5.5-medium on behalf of matt wilkie`.
- An agent posts a PR body with escaped newline text, so GitHub renders the validation section as one ugly line instead of a real list. `scripts/gh-body-lint` catches that before posting and also nudges non-linking issue refs to `#123` or `owner/repo#123` so GitHub links them.
- A PR is reviewed, then another maintainer or contributor adds a late comment before the agent closes or merges it. The final reread rule makes that last pass explicit: re-check latest comments, review threads, and linked issues before taking the maintainer action.
- A checked-in generated `AGENTS.md` block still uses the legacy marker `<!-- BEGIN BEADS INTEGRATION -->`. The refreshed output is `<!-- BEGIN BEADS INTEGRATION v:1 profile:full hash:e07235d4 -->`, where `v:1` is the marker-format version and `hash:` tracks the rendered body for safe refresh detection.

## Validation

- `scripts/pr-preflight.sh --search "agent instructions signing body lint execution metadata AGENTS" --repo gastownhall/beads`
- `scripts/pr-preflight.sh --search "Agent-Signature gh-body-lint PR_MAINTAINER_GUIDELINES execution_agent_type" --repo gastownhall/beads`
- `go test ./internal/templates/agents ./cmd/bd/setup`
- `scripts/check-doc-freshness.sh`
[...]

Issue #3541, opened 2026-04-27
Add structured execution metadata keys for agent delegation hints


maphew:
## Summary

Add documented structured metadata keys for execution/delegation hints on bead items so agents and runners can make model, reasoning, and scheduling decisions without parsing prose from descriptions, design notes, or comments.

## Rationale

Agents currently need to infer delegation details from unstructured text when deciding which execution agent, model, reasoning effort, or parallelization group should handle a bead. That is brittle and difficult to standardize across CLI workflows, MCP clients, and external agent runners.

The existing `Issue.Metadata json.RawMessage` field is a good fit for this enhancement:

- `internal/types/types.go` documents `Issue.Metadata` as an extension point for tool annotations.
- Metadata participates in the content hash, so these annotations are included in bead identity/history.
- `bd update --set-metadata key=value` already supports incremental metadata edits.
- `bd list`, `bd ready`, and `bd search` already support top-level metadata filtering with `--metadata-field` and `--has-metadata-key`.

Recommendation: start with metadata keys rather than adding new top-level `Issue` fields. These values are currently agent hints, not core bead scheduling semantics.

## Proposed keys

Use flat top-level metadata keys:

- `execution_agent_type`
- `execution_suggested_model`
- `execution_reasoning_effort`
- `execution_mode`
- `execution_parallel_group`

Example usage:

```bash
bd update bd-123 \
  --set-metadata execution_agent_type=codex \
  --set-metadata execution_suggested_model=gpt-5 \
  --set-metadata execution_reasoning_effort=high \
  --set-metadata execution_mode=delegate \
  --set-metadata execution_parallel_group=docs-batch
```

## When to promote to first-class fields

Keep these as metadata until core workflows begin treating them as scheduling or execution semantics across tools.

Promote to top-level `Issue` fields only if workflows such as `bd ready`, `bd claim`, MCP clients, agent runners, or other core integrations need to interpret these values consistently as first-class execution behavior rather than optional annotations.

## Acceptance criteria

- Documentation describes the recommended execution/delegation metadata keys and their intended meanings.
- Documentation includes at least one `bd update --set-metadata` example showing incremental edits.
- Tests cover preserving these metadata keys through create/update/list or equivalent metadata round trips.
- Tests or documented examples show filtering beads with `--metadata-field` and/or `--has-metadata-key` for these execution keys.
[...]

PR #4358, opened 2026-06-10
docs(metadata): document portability of execution model/effort hints

maphew:
## What

Two small wording changes to `docs/METADATA.md`, in the agent execution metadata section:

- Document the portability rule for `execution_reasoning_effort` and `execution_suggested_model`: stored effort values are a canonical advisory scale that consumers map to their runtime's nearest native level, and suggested models are capability-tier hints that consumers on a different provider substitute with a same-tier model rather than drop.
- Update the gh-3541 reference from "open question" phrasing to its resolution: execution hints stay metadata-only, with the JSON/JQ snippet as the supported access path.

## Why

Agent runtimes disagree on effort vocabularies and model names. The doc currently exemplifies one runtime's effort scale (`low`/`medium`/`high`/`xhigh`) without saying what a consumer on a runtime with a different scale should do, and `execution_suggested_model` does not say whether a model value binds a consumer on another provider. In cross-runtime use (one agent writes the hint, a different runtime's agent consumes it), that ambiguity makes the hints either silently dropped or wrongly treated as hard pins.

The fix is deliberately a rule, not a table: no model names and no per-vendor scale mappings are added, because those age with every model release. The rule (canonical scale, consumer maps; tier suggestion, consumer substitutes) stays correct as runtimes evolve.

This follows the project charter's metadata-before-schema boundary - it only clarifies semantics of the existing advisory keys, no schema or CLI changes.

## Collision check

`scripts/pr-preflight.sh --search "execution metadata reasoning effort model"` found no open PRs on this topic. This PR shares no files with #4237 (agent maintainer conventions) and is independent of it. #3541 is the issue that originated these keys; it closed 2026-06-03 with the metadata-only resolution this PR now reflects.

### DIFF
--- a/docs/METADATA.md
+++ b/docs/METADATA.md
@@ -33,14 +33,22 @@ These keys are advisory metadata, not core issue fields. When a workflow uses
 them, they take precedence over free-form notes for execution routing. Notes
 remain useful for rationale, ownership, and exact prompts.
 
+Model and effort values are portable hints, not runtime bindings. The
+`execution_reasoning_effort` values above are a canonical advisory scale:
+writers should store canonical values rather than runtime-local ones, and a
+consumer whose runtime uses a different scale should map the stored value to
+its nearest native level instead of dropping the hint. Likewise,
+`execution_suggested_model` is a capability-tier suggestion: a consumer on a
+different provider should substitute a model of the same tier rather than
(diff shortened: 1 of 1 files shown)


maphew:
doc only change that didn't need formal review

PR #4312, opened 2026-06-06
fix(docgen): guard generate-cli-docs.sh against CGO-binary federation drift (be-3cse)

quad341:
## Problem

`generate-cli-docs.sh` selects its `bd` binary as: `$1`, else `$PROJECT_ROOT/bd`, else builds `CGO_ENABLED=0`. When a CGO-enabled `./bd` is left in `PROJECT_ROOT` (e.g. from `make build` during a builder session), a no-arg or implicit-PROJECT_ROOT invocation regenerates docs from it. A CGO `bd` exposes the `bd federation` subcommands, writing federation pages into `docs/CLI_REFERENCE.md`, `website/{docs,versioned_docs/version-*}/cli-reference/federation.md` and `website/static/llms-full.txt`. CI's `check-doc-flags` builds `CGO_ENABLED=0`, whose `bd` emits no federation subcommands → doc-flags fails and cascades to Build Artifacts / PR Policy / CI Gate. Hit PRs #3710, #4055, #4153 simultaneously on 2026-06-06.

## Fix

After resolving the `bd` binary (regardless of which path was taken), probe `"$BD" federation --help`. If it does **not** print the pure-go stub (`Federation commands require CGO`), the binary is CGO-enabled: rebuild a fresh `CGO_ENABLED=0 -tags gms_pure_go` binary internally and use that for doc generation. The explicit-arg fast path is preserved for callers that already pass a pure-go binary.

The oracle comment points to `scripts/ci/pr-policy.sh build_docs_binary` (`env CGO_ENABLED=0 go build`) so future maintainers know what CI does.

## Testing

- `LC_ALL=C ./scripts/generate-cli-docs.sh --check /tmp/bd-pure-test` → `PASS` (zero churn, no rebuild triggered)
- `LC_ALL=C ./scripts/generate-cli-docs.sh --check /tmp/fake-cgo-bd.sh` → guard fires, rebuilds pure-go, `PASS`
- `./scripts/check-doc-flags.sh /tmp/bd-pure-test && ./scripts/check-doc-freshness.sh` → both `PASS`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4243, opened 2026-05-29
fix(ci): build doc-check bd with CGO so federation docs stay fresh (#4203)

maphew:
## Why

Closes #4203. The `check-doc-flags` CI job built `bd` with `CGO_ENABLED=0`, which compiles the `bd federation` **stub** (`cmd/bd/federation_nocgo.go`, `//go:build !cgo`) rather than the full subcommand tree (`cmd/bd/federation.go`, `//go:build cgo`). It then regenerated docs from that stub binary and diffed them against the committed docs — which were generated with a CGO build. Any PR whose docs reflected a CGO build (the default a contributor gets from `go build`) failed the freshness check.

The drift became visible once #4181 / #4182 (merged 2026-05-26) widened the check to website per-command pages, versioned snapshots, and `llms-full`.

## What

- **`.github/workflows/ci.yml`** — drop `CGO_ENABLED=0` from the doc-check build (Option 1 in #4203). This aligns it with every other `go build` in the file and makes the doc-generating binary match the feature-complete artifact users actually run. Added a comment so it doesn't silently regress.
- **Regenerated docs** with a CGO `bd` (`generate-cli-docs.sh` + `generate-llms-full.sh`), restoring the full `bd federation` reference (`add-peer` / `list-peers` / `remove-peer` / `status` / `sync`) that the merged #4196 workaround had collapsed to the stub message.

## Verification

Built with the exact new CI command and ran both checks locally — green:

```
go build -tags gms_pure_go -o ./bd ./cmd/bd/
./scripts/check-doc-flags.sh ./bd        # PASS: generated CLI docs are fresh
./scripts/check-doc-freshness.sh         # PASS: freshness markers current
```

## Notes for reviewers

- **Federation is intrinsically CGO-only** (requires the Dolt backend), so the full reference is only actionable on a CGO build; no-CGO users still get "requires CGO" when they actually invoke it. Documenting the full tree loses nothing and gives federation users their reference back.
- **`.goreleaser.yml` ships both CGO and no-CGO artifacts**, so the rationale here is "match the feature-complete binary," not "match every shipped binary."
- **Versioned snapshots** (`website/versioned_docs/version-1.0.0`, `version-1.0.4`) are regenerated because `generate-cli-docs.sh` compares them; this reverses the stub-ification from #4196. If the project would rather freeze versioned snapshots to what each release shipped, that's a separate tooling change.
[...]

Issue #4203, opened 2026-05-27
Docs freshness CI compares no-CGO bd output to CGO-generated committed docs (federation tree drifts)


coffeegoddd:
## Problem

The `check-doc-flags` job in `.github/workflows/ci.yml` builds `bd` with:

```
CGO_ENABLED=0 go build -tags gms_pure_go -o bd ./cmd/bd/
```

then runs `scripts/check-doc-flags.sh` + `scripts/check-doc-freshness.sh`, which regenerate docs from that binary and diff against the committed docs.

The committed docs were generated at some prior point with a **CGO-enabled** `bd`, so they include the full `bd federation` subcommand tree (`add-peer`, `list-peers`, `remove-peer`, `status`, `sync`). A no-CGO `bd` replaces that tree with a single "requires CGO" stub.

Net result: any time the committed docs reflect a CGO build (the default a contributor gets locally with `go build`), the CI freshness check fails. Any time they reflect a no-CGO build, the website CLI reference loses the federation subcommand pages that federation users would look for.

## When this became visible

The drift has existed for weeks, but the older freshness check only diffed `docs/CLI_REFERENCE.md`, where federation's footprint is small. Two PRs that merged on 2026-05-26 widened the check:

- #4181 (commit `a90671ce0`) added `scripts/check-doc-freshness.sh` and called it from the job.
- #4182 (commit `14d9d2880`) widened `generate-cli-docs.sh` to compare website per-command pages and versioned snapshots, and added a `--check` mode for `generate-llms-full.sh`.

After those merged, the previously-invisible CGO/no-CGO inconsistency in federation docs started failing every PR build whose committed docs were generated with CGO.

## Reproduction

On `db/provider` (PR #4196) at HEAD `2bfb5ad15`:

```
# Matches what CI does — fails
CGO_ENABLED=0 go build -tags gms_pure_go -o ./bd ./cmd/bd/
./scripts/check-doc-flags.sh ./bd
# → FAIL: docs/CLI_REFERENCE.md out of sync, website/* out of sync

# Locally natural — passes
go build -tags gms_pure_go -o ./bd ./cmd/bd/
./scripts/check-doc-flags.sh ./bd
# → PASS (until something on the no-CGO path regenerates llms-full)
```

The diff is the federation subcommand tree (~107 lines per copy) being collapsed to/expanded from a "requires CGO" stub.

## Workaround currently in PR #4196

Regenerate everything with the exact CI build command and commit:

```
[...]

PR #4359, opened 2026-06-10
ci(docs): blame-scoped doc freshness gate + exact-fix patch artifact (#4301)

### LINKED ISSUE #4301: CI doc generation is a nightmare to deal with, just have CI push a commit to the PR with the generated docs
The CI doc check pipeline is still busted/ inconsistent and wastes a ton of time having me try and generate the correct docs that CI wants,
which seems to vary across environments.

See these commits: https://github.com/gastownhall/beads/pull/4300

If the generated docs are important, just push a commit to my branch in CI with them generated so I dont have to futz with it.

maphew:
## Why

Closes #4301. The doc freshness gate regenerates every generated doc artifact and fails the PR on any diff, even drift the contributor did not cause. Two failure modes have been burning contributor time:

- **Inherited drift**: when a staleness window opens on `main` (like the 2026-06-06 cascade that hit #3710, #4055 and #4153 simultaneously), every open PR goes red until someone regenerates docs, regardless of what the PR touches.
- **Environment mismatch**: the contributor's local regeneration (default `go build` is CGO-enabled, unlocking the `bd federation` tree; locale-dependent file ordering; stale `./bd` binaries) produces different bytes than CI's canonical pinned build, so "just run the script" turns into a guessing loop (#4300, #4203).

The principle this PR implements: **contributors only own docs for the code they actually touched**. And when they do owe a regeneration, CI should hand them the exact fix rather than making them reproduce CI's environment.

## What

New `scripts/check-cli-docs-drift.sh`, wired into `check-doc-flags.sh` check 4 (so both the `check-doc-flags` job and the `pr-policy` wrapper pick it up):

1. **Fast path unchanged**: the strict regenerate-and-diff probe (`generate-cli-docs.sh --check`) runs first. Fresh docs pass exactly as before, no extra cost.
2. **Attribution on failure (PRs only)**: regenerate the artifacts at HEAD and at the merge-base, each in a scratch git worktree with the canonical pinned build (`CGO_ENABLED=0 go build -tags gms_pure_go`). If the PR neither changes the regenerated CLI surface nor touches generated files, the drift is inherited from the base branch: emit a `::warning::` annotation and pass.
3. **The fix as an artifact**: when drift IS attributable (the PR changed the CLI surface, or hand-edited generated files), the check fails and writes the regenerated diff to a patch, uploaded as the `cli-docs-freshness-patch` artifact. Download, `git apply`, push. No local environment involved. This is the practical version of the issue's "just push a commit to my branch": CI produces the canonical commit content; it can't safely push to fork branches with the read-only `pull_request` token, so it hands over the patch instead.
4. **Strict mode preserved** wherever there is no PR base ref: pushes to `main`, releases, and plain local runs behave exactly as today. A binary-mismatch false alarm (e.g. a CGO `bd` passed to the check) is also self-correcting: if the committed docs are fresh under the canonical build, the check passes and says why.

Workflow changes in `.github/workflows/pr.yml`: `fetch-depth: 0` on the doc-check job (merge-base needs history), `DOC_DRIFT_PATCH_OUT` env, and an `if: failure()` artifact upload step.

## What this PR deliberately does not do

- It does not regenerate the committed docs or take a side on the CGO-vs-pure-Go canonical surface question (#4203 / #4243 / #4312). It is compatible with either outcome: whatever build CI pins, attribution uses the same pinned build at both ends.
- It does not touch `check-doc-freshness.sh` (the `Last reviewed` marker ages). Note that an expiring marker fails unrelated PRs the same way inherited drift did; that deserves the same treatment in a follow-up.

Complements #4312 (pinning the docgen binary in generate mode); the attribution path here independently never trusts a caller-supplied binary.

## Test plan

- [x] `bash -n` both scripts
- [x] Clean tree, fast path: `./scripts/check-doc-flags.sh ./bd` passes with no behavior change
- [x] Inherited drift: synthetic base commit with a hand-staled generated file + unrelated PR commit on top → strict probe fails, attribution regenerates at HEAD and merge-base, verdict "inherited", exit 0 with warning
- [x] Attributable drift (CLI surface change without regen) → fails with the surface-change reason, patch written; the patch also heals pre-existing inherited drift; `git apply` + commit turns the check green via the fast path
- [x] Hand-edited generated file (surface unchanged) → fails with the touched-files reason
- [x] Strict fallback with no base ref → fails as today

Validation details for each scenario are in the PR comments.

_claude-unknown-model-high on behalf of matt wilkie_

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -147,6 +147,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          # Full history so check-cli-docs-drift.sh can regenerate docs at the
+          # merge-base and only fail PRs for drift they actually introduced.
+          fetch-depth: 0
 
       - name: Set up Go
         uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
@@ -157,9 +161,22 @@ jobs:
         run: CGO_ENABLED=0 go build -tags gms_pure_go -o bd ./cmd/bd/
 
       - name: Validate docs against CLI
+        env:
+          DOC_DRIFT_PATCH_OUT: ${{ runner.temp }}/cli-docs-freshness.patch
         run: |
           ./scripts/check-doc-flags.sh ./bd
           ./scripts/check-doc-freshness.sh
+
+      # When the drift check fails, it writes the exact regenerated-docs fix
+      # (produced with CI's canonical build) so contributors can apply it with
+      # `git apply` instead of reproducing CI's environment locally.
+      - name: Upload docs fix patch
+        if: failure()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: cli-docs-freshness-patch
+          path: ${{ runner.temp }}/cli-docs-freshness.patch
+          if-no-files-found: ignore
 
   # Fast check to catch accidental .beads/issues.jsonl changes from contributors
   check-no-beads-changes:
--- a/scripts/check-cli-docs-drift.sh
+++ b/scripts/check-cli-docs-drift.sh
@@ -1,0 +1,231 @@
+#!/bin/bash
+# check-cli-docs-drift.sh — Blame-scoped freshness check for generated CLI docs.
+#
+# Strict probe first: regenerate the generated doc artifacts and diff against
+# the committed copies, exactly like generate-cli-docs.sh --check. When the
+# probe fails AND a diff base is known (a PR), the drift is attributed before
+# failing anyone:
+#
+#   * regenerate the same artifacts at HEAD and at the merge-base, each in a
+#     scratch worktree with the canonical pinned build
+#     (CGO_ENABLED=0 go build -tags gms_pure_go);
+#   * if the change neither alters the regenerated CLI surface nor touches the
+#     generated files, the drift was inherited from the base branch: warn and
+#     pass — contributors only own docs for the code they actually touched;
+#   * otherwise fail, and write the exact regeneration diff as a patch
+#     (--patch-out / $DOC_DRIFT_PATCH_OUT) so the fix never depends on the
+#     contributor's local build environment.
+#
+# Without a diff base (push to main, releases, plain local runs) the strict
+# probe result stands, preserving the historical behavior.
+#
+# Usage: check-cli-docs-drift.sh [--base <ref>] [--patch-out <file>] [bd-binary]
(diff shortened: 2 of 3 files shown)


maphew:
Validation transcript (local, scratch clone of this branch at `24c6d7ae5`; synthetic commits simulate the base branch and PR):

**A. Clean tree, fast path** — `./scripts/check-doc-flags.sh ./bd` on this branch: `PASSED: All documentation references are consistent with CLI`. No behavior change, no extra cost.

**B. Inherited drift passes with a warning** — base commit hand-stales `website/docs/cli-reference/index.md`; PR commit on top touches only `README.md`:

```
WARN: generated docs are stale, but the drift is inherited from the base branch:
this change neither alters the regenerated CLI surface nor touches generated files.
   website/docs/cli-reference/index.md | 2 +-
Not failing this PR. The base branch needs a docs regeneration:
  ./scripts/generate-cli-docs.sh && ./scripts/generate-llms-full.sh
PASS: no doc drift attributable to this change   (exit 0)
```

**C. Attributable surface change fails and hands over the fix** — PR commit edits the `bd version` `Short:` text without regenerating; base also carries the inherited staleness from B:

```
FAIL: this change affects the generated CLI docs and the committed copies are stale.
  - the regenerated CLI surface differs from the merge-base
   docs/CLI_REFERENCE.md                                         | 4 ++--
   website/docs/cli-reference/index.md                           | 2 +-
   website/docs/cli-reference/version.md                         | 2 +-
   website/static/llms-full.txt                                  | 2 +-
   website/versioned_docs/version-1.0.0/cli-reference/version.md | 2 +-
   website/versioned_docs/version-1.0.4/cli-reference/version.md | 2 +-
   website/versioned_docs/version-1.0.5/cli-reference/version.md | 2 +-
The exact fix (regenerated with CI's canonical build) was written to:
  /tmp/cli-docs-freshness.patch   (exit 1)
```

`git apply` of that patch + commit turns the check green via the fast path (and note the patch healed the inherited `index.md` staleness too).

**D. Hand-edited generated file fails** — PR commit edits `docs/CLI_REFERENCE.md` directly, CLI surface unchanged:

```
FAIL: this change affects the generated CLI docs and the committed copies are stale.
  - generated doc files were modified by this change:
      docs/CLI_REFERENCE.md   (exit 1)
```

**E. Strict fallback** — no `--base`, no `GITHUB_BASE_REF` (push-to-main / release / local shape): `FAIL: generated CLI docs are out of sync (strict result stands)` (exit 1), matching today's behavior.

Also ran `actionlint` on the modified `pr.yml` (clean) and `bash -n` on both scripts.

Cost note: the attribution path adds two scratch-worktree builds + regenerations, but only runs when the strict probe already failed; green PRs pay nothing.

_claude-unknown-model-high on behalf of matt wilkie_

Issue #2946, opened 2026-04-01
Beads is using embedded Dolt despite shared-server being set


mieubrisse:
I ran a `bd init --prefix pwriting --shared-server` yesterday. Finally got it working, I thought, but then saw weird stuff where beads I thought were added weren't added.

I think it's Beads using an embedded server despite shared-server being connected:

```
$  tail -10 config.yaml
# Integration settings (access with 'bd config get/set')
# These are stored in the database, not in this file:
# - jira.url
# - jira.project
# - linear.url
# - linear.api-key
# - github.org
# - github.repo

dolt.shared-server: true

$ bd dolt start
Error: 'bd dolt start' is not supported in embedded mode (no Dolt server)

$  bd version
bd version 0.63.3 (Homebrew)
```

mieubrisse:
Looks like it's completely ignoring `--shared-server`. I blew away `.beads`, redid `bd init --prefix pwriting --shared-server`, and it's still using embedded Dolt:

```
$  bd init --prefix pwriting --shared-server
Note: origin has an existing beads database (refs/dolt/data).
  Run 'bd bootstrap' instead to clone it.
  Continuing with fresh database initialization.

  ✓ Configured Dolt remote: origin → git+ssh://git@github.com/mieubrisse/personal-writing.git
  Repository ID: 72ffaecd
  Clone ID: da637d031ad31e02
  ✓ Shared server mode enabled
Contributing to someone else's repo? [y/N]: n
  Hooks installed to: .beads/hooks/
  AGENTS.md already has current agent instructions
  ✓ Committed beads files to git

✓ bd initialized successfully!

  Backend: dolt
  Mode: embedded
  Database: pwriting
  Issue prefix: pwriting
  Issues will be named: pwriting-<hash> (e.g., pwriting-a3f2dd)

Run bd quickstart to get started.
```

mieubrisse:
looks like switching the metadata.json to "server" mode per the PR still says that there's no database inside the shared Dolt server. so `bd init` is doing something that can't just be recreated by switching "embedded" -> "server" in metadata.json

PR #4366, opened 2026-06-10
fix(storage): fail loud when metadata.json exists but cannot be loaded, instead of silently opening the embedded store

Wldc4rd:
## Problem

On server-mode deployments, a transient failure to load `.beads/metadata.json` silently degrades store selection to the embedded default. The command then answers every query from an empty embedded relic with **exit 0** — a false-empty that orchestrators read as "no work".

Observed live (multi-agent Gas City deployment, dolt server under load): all three label-list queries (`approved` / `needs-review` / `needs-refinery`) returned exit 0 + empty **simultaneously**, self-cleared ~30s later, and nearly caused a missed-merge decision. The only trace was one unscanned stderr line (`warning: failed to load beads config from ...`).

## Root cause

`configfile.Load` already distinguishes *absent* (`nil, nil` — legitimate fresh-repo default) from *present-but-unloadable* (`nil, err` — torn read mid-rewrite, EINTR/EMFILE under load, corrupt file). Four call sites collapsed the two:

| site | pre-patch behavior on load error |
|---|---|
| `main.go` `loadServerModeFromBeadsDir` | early return; `serverMode` stays false → embedded |
| `main.go` store-init (~line 960) | stderr warning, then **continues with zero-value config** → embedded + default db name |
| `store_factory.go` `newDoltStoreFromConfig` | `err == nil &&` guard skips server branch → `embeddeddolt.Open` |
| `store_factory.go` `newReadOnlyStoreFromConfig` | same; surfaces a misleading downstream "database not found" |

Two non-atomic writers provide the torn-read trigger: `configfile.Save` and `doltserver.writePortFile` both used `os.WriteFile` (in-place truncate window).

## Fix

One contract, applied at all four sites: **metadata.json exists but cannot be loaded → hard error naming the file and cause; absent → unchanged fresh-repo default.** Plus write-temp-then-rename for `configfile.Save` and `writePortFile` so concurrent readers never observe a torn file. `store_factory_nocgo.go` gets the same treatment so the error names the real cause instead of "embedded requires CGO".

This is consistent with the existing protection philosophy: bd already refuses loudly on project-identity mismatch (GH#2372) and guards auto-import against clobbering (#3691); the corrupt-config path was the remaining silent hole in the same story.

## Repro / verification

Server-mode rig, `metadata.json` truncated to a mid-write shape (`{"dolt_mode":"serv`):

```
before: bd list --label approved --json  -> exit 0, []   (false-empty)
after:  bd list --label approved --json  -> exit 1,
        load .../metadata.json: parsing config: unexpected end of JSON input
        (refusing to guess the storage mode; the embedded fallback would mask server-backed data)
```

- New tests in `cmd/bd/store_selection_failloud_test.go` (red pre-patch: the cgo factory silently created and returned an embedded store; the mode loader swallowed the error). Absent-metadata case pinned green to protect the fresh-repo default.
- `go build` clean under `CGO_ENABLED=1` and `0`; `go vet` clean; `internal/configfile`, `internal/doltserver`, and the targeted `cmd/bd` selection/config/port suites pass.
- Dogfooded against a live production-scale store: patched binary returns identical results and exit codes to the running 1.0.5 on real label queries (server mode, correct metadata), and fails loud on the corrupt-metadata repro where 1.0.5 returns the silent empty.


maphew (commented):
Thanks for this - it is a well-diagnosed fix for a nasty failure mode. A present-but-unloadable `metadata.json` currently degrades silently to the embedded store, which on server-mode deployments answers every query from an empty relic with exit 0. We have been bitten by exactly this false-empty class locally, and I confirmed the bug is still live on current main (`cmd/bd/main.go` warning-and-continue at the store-init site, `err == nil &&` guards in both store factories). The four-site contract (absent = fresh-repo default, unloadable = hard error naming file and cause) is the right shape, and the red-first tests pin it.

Verified: the branch merges cleanly onto current main - the `warnSharedServerEmbeddedMismatch` rename and the #2946 shared-server override that landed since your base compose fine with the error-return change (checked via three-way merge-tree, zero conflicts).

Findings:

- **[major]** The `FatalError` in rootCmd's `skipsStoreInit` branch (the `loadServerModeFromConfig()` call in `cmd/bd/main.go`) fires for every command in `noDbCommands` - including `bd doctor`, `bd init`, `bd version`, and `bd help`. With a corrupt `metadata.json`, the user is locked out of the exact tools the error message tells them to use to fix it. Suggest keeping the hard error on the store-opening paths (`prepareSelectedCommandContext`, the store-init site, both store factories) and downgrading the skipsStoreInit path to a loud non-fatal stderr warning, or exempting the no-DB commands.
- **[minor]** `internal/atomicfile.WriteFile` already exists (predates this branch) and does temp+rename **with fsync**. The new `writeFileAtomic` in `internal/configfile/configfile.go` and the inline temp+rename in `writePortFile` skip the fsync, so a crash can still leave an empty renamed file on some filesystems - the corruption class this PR targets. Swapping both to `atomicfile.WriteFile` deletes code and strengthens the guarantee.
- **[nit]** `store_selection_failloud_test.go` is `//go:build cgo` only, so the `store_factory_nocgo.go` changes are untested; a mirrored test would pin them.
- **[nit]** The same PreRun now uses `return HandleError(...)` style on main; `FatalError` bypasses that flow. Worth aligning while you are in the hunk.

CI has not run on the branch yet (needs workflow approval); we will trigger it once the above lands.

Recommendation: **fix-merge**. Address the [major] doctor/init lockout and swap in `atomicfile.WriteFile`, push, and we will approve the workflow run and land it. The core contract change is exactly what we want - thank you.

Wldc4rd:
@maphew Sorry for letting this lie for so long. Unfortunately, we've switched things around and I'm not sure our current environment can repro this issue (plus it may have already been fixed upstream after so long) AND I'm low on tokens at this time so I'm punting this one to you guys. :)

maphew:
Thanks @Wldc4rd — no need to apologize, and this was worth picking up. I reproduced the bug from scratch against a binary built from today's `main` tip: with a truncated `metadata.json`, `bd list --json` answered `[]` with exit 0 while the store held data, and the fallback even minted a brand-new empty default-name database on disk next to the real one. The hole you patched is real and was still fully open upstream.

**Taking this over as maintainer.** Summary of what I did on your branch:

1. **The red CI was a mirage — your patch never broke anything.** Today's failing checks are attempt-2 re-runs of the June 10 CI, pinned to your original merge commit (`ddbd25e` into `de7d3aed5`):
   - The two `Embedded Dolt Cmd` shard failures (`TestEmbeddedUpdateRoutedStoreCommitsTargetHead`, `TestEmbeddedRoutedSiblingWritesCommitTargetHead`, "store is read-only") reproduce at the June 10 base **without** your patch — a since-fixed main bug of that era.
   - The three `Upgrade smoke` failures resolve the release matrix at runtime, so the re-run pitted late-June v1.1.0-family data against a candidate binary frozen at June 10 code — a downgrade in disguise, unrelated to this change.
2. **Merged current `main` into the branch** (clean merge; your commit is intact) and validated locally: your four unit tests, both routed integration suites, `internal/configfile`, `internal/doltserver`, `go vet`, and builds under `CGO_ENABLED=0` and `1` all pass.
3. **One real fix on top** (`64cdff3`), flagged independently by two reviewers: the fatal also fired in the pre-run contexts that serve no-DB commands, so `bd doctor`, `bd init`, `bd version`, `bootstrap`, and `context` all exited 1 on a corrupt `metadata.json` — bricking the exact repair path the error message points at ("fix or restore metadata.json and retry"). Those contexts now warn-and-continue; the hard error stays where a store is actually selected (store init + both store factories), which is where the false-empty risk lives. Your unit tests and the error contract are unchanged.
4. **Added regression coverage**: an end-to-end test (corrupt `metadata.json` → `version`/`doctor` still run, `list` fails loud naming the file, `bd init` repairs, `list` works again with data intact) and a nocgo variant asserting the factory error names the real cause instead of "embedded requires CGO".

The write-temp-then-rename hardening in `configfile.Save` and `writePortFile` reviewed clean (symlink/Windows-rename/crash-leftover angles checked). Plan: merge once fresh CI is green. Thanks for the report, the diagnosis, and the tests — the false-empty analysis in the PR description made this easy to verify.

PR #4367, opened 2026-06-10
fix(embeddeddolt): give routed writes a writable target store (be-cibr)

quad341:
## Problem

Since 3f2b4e08d (`fix(embeddeddolt): give read-only opens a real read-only path`, bd-6dnrw.32), main is red on two shards:

- **Test (Embedded Dolt Cmd 8/20):** `TestEmbeddedUpdateRoutedStoreCommitsTargetHead`
- **Test (Embedded Dolt Cmd 12/20):** `TestEmbeddedRoutedSiblingWritesCommitTargetHead` (comment/note/reopen subtests)

All fail with `embeddeddolt: store is read-only` when a routed write (`bd update`, `bd comment`, `bd note`, `bd reopen`, `bd close`) tries to commit the target head.

**Root cause:** `resolveViaPrefixRouting` opens the target rig's database via `newReadOnlyStoreFromConfig` → `embeddeddolt.OpenReadOnly`. Before bd-6dnrw.32, `OpenReadOnly` returned a writable store (the name was aspirational). After bd-6dnrw.32 it returns a genuinely read-only store — correct for cross-repo hydration, but it broke the routed-write path that shares the same factory.

## Fix

Three-file change:

1. **`internal/storage/embeddeddolt/store.go`** — add `OpenWritable`: identical to `OpenReadOnly` (skips `initSchema`, only checks forward drift) but does **not** set `readOnly: true`. Preserves the no-foreign-migration contract from GH#3231 while allowing writes.

2. **`cmd/bd/store_factory.go` + `store_factory_nocgo.go`** — add `newWritableRoutedStoreFromConfig` mirroring `newReadOnlyStoreFromConfig` but using `OpenWritable` / non-ReadOnly server config.

3. **`cmd/bd/routed.go`** — add `forWrite bool` to `resolveViaPrefixRouting`. Write callers (`resolveAndGetIssueWithRouting`, `resolveCloseTargets`) pass `true`; read caller (`getIssueWithRouting`) passes `false`.

The read-only path (`newReadOnlyStoreFromConfig` → `OpenReadOnly`) is unchanged for cross-repo hydration and `openRoutedReadStore`.

## Verification

Both previously-failing tests now pass locally:

```
--- PASS: TestEmbeddedUpdateRoutedStoreCommitsTargetHead (70.85s)
--- PASS: TestEmbeddedRoutedSiblingWritesCommitTargetHead (78.20s)
    --- PASS: .../comment
    --- PASS: .../note
    --- PASS: .../reopen
```

### DIFF
--- a/cmd/bd/close.go
+++ b/cmd/bd/close.go
@@ -613,10 +613,7 @@ func resolveCloseTargets(ctx context.Context, localStore storage.DoltStorage, id
 			cleanup()
 			return nil, func() {}, fmt.Errorf("resolving ID %s: %w", id, err)
 		}
-		// Write-intent: a prefix-routed target opens writable so the close
-		// commits on the target head (#4141). Contributor auto-routing below
-		// stays read-only: it hydrates foreign projects that must not be mutated.
-		if r, err := resolveViaPrefixRoutingMode(ctx, id, true); err == nil {
+		if r, err := resolveViaPrefixRouting(ctx, id, true); err == nil {
 			results = append(results, r)
 			continue
 		}
--- a/cmd/bd/close.go
+++ b/cmd/bd/close.go
@@ -594,7 +594,7 @@ func resolveCloseTargets(ctx context.Context, localStore storage.DoltStorage, id
 			return nil, fmt.Errorf("no auto-routed store available")
 		}
 		sharedRoutedTried = true
-		rs, routed, err := openRoutedReadStore(ctx, localStore)
+		rs, routed, err := openRoutedStore(ctx, localStore)
 		if err != nil {
 			return nil, err
 		}
@@ -613,7 +613,7 @@ func resolveCloseTargets(ctx context.Context, localStore storage.DoltStorage, id
 			cleanup()
 			return nil, func() {}, fmt.Errorf("resolving ID %s: %w", id, err)
 		}
-		if r, err := resolveViaPrefixRouting(ctx, id, true); err == nil {
+		if r, err := resolveViaPrefixRouting(ctx, id); err == nil {
 			results = append(results, r)
 			continue
 		}
--- a/cmd/bd/list.go
+++ b/cmd/bd/list.go
@@ -483,7 +483,7 @@ var listCmd = &cobra.Command{
 
 		activeStore := store
 		// Contributor auto-routing: read from the same target repo as bd create.
-		routedStore, routed, err := openRoutedReadStore(ctx, activeStore)
+		routedStore, routed, err := openRoutedStore(ctx, activeStore)
 		if err != nil {
 			FatalError("%v", err)
 		}
--- a/cmd/bd/ready.go
+++ b/cmd/bd/ready.go
@@ -185,7 +185,7 @@ This is useful for agents executing molecules to see which steps can run next.`,
 			CheckReadonly("ready --claim")
 		} else {
 			// Contributor auto-routing: read from the same target repo as bd create.
-			routedStore, routed, err := openRoutedReadStore(ctx, activeStore)
+			routedStore, routed, err := openRoutedStore(ctx, activeStore)
 			if err != nil {
 				FatalError("%v", err)
 			}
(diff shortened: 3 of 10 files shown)


quad341:
Pushed `9809349c` to clear the only red checks (Lint / Build Artifacts / PR Lint / CI Gate). They were all the same `unparam` finding:

```
cmd/bd/routed.go:72: resolveAndGetIssueWithRoutingMode - forWrite is unused (unparam)
```

When the routed prefix open was unified onto `newWritableNoMigrateStoreFromConfig` (be-jb55), the inner `resolveViaPrefixRoutingMode(ctx, id, forWrite)` call became `resolveViaPrefixRouting(ctx, id)`, which left `forWrite` on the outer `resolveAndGetIssueWithRoutingMode` read nowhere. The commit drops the dead parameter plus its two pass-through args, and refreshes the two wrapper doc-comments that still described the old read-only-vs-writable split (the writable-no-migrate open keeps the bd-6dnrw.32 / GH#3231 no-open-time-mutation invariant intact).

Behavior is unchanged — it's a dead-parameter removal. Verified locally: `golangci-lint --build-tags=gms_pure_go` reports 0 issues, `go build` + `go vet` pass; the Embedded Dolt test shards were already green at the prior head.

maphew:

quad341:
Thanks @maphew. The reason a repro is hard to find today is that **current `main` already fixes this independently** — so this PR is effectively superseded. Details:

### Original repro (buggy window)

- **Storage mode:** embedded Dolt, cross-rig **prefix routing** via `.beads/routes.jsonl` (a bead ID whose prefix routes to a *sibling* rig's database).
- **OS:** not OS-specific — it's about which store-open mode the routed path picks (reproduced on the Linux CI shards).
- **Buggy commit window:** starting at `3f2b4e08d` (`fix(embeddeddolt): give read-only opens a real read-only path`, bd-6dnrw.32), which made `OpenReadOnly` return a genuinely read-only store. The routed **write** path shared that same factory.

The smallest concrete repro is already committed as two end-to-end tests (each builds throwaway scratch repos):

```
BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 \
  go test -tags gms_pure_go ./cmd/bd -count=1 -v \
  -run 'TestEmbeddedUpdateRoutedStoreCommitsTargetHead|TestEmbeddedRoutedSiblingWritesCommitTargetHead'
```

- `TestEmbeddedUpdateRoutedStoreCommitsTargetHead` (`cmd/bd/update_embedded_test.go`)
- `TestEmbeddedRoutedSiblingWritesCommitTargetHead` (`cmd/bd/autocommit_embedded_test.go`) — `comment` / `note` / `reopen` subtests

What they automate, in prose: init a source rig (`--prefix src`), init a sibling `target-repo` rig (`--prefix tgt`), create an issue in the target, write `{"prefix":"tgt-","path":"target-repo"}` to the source rig's `.beads/routes.jsonl`, then run a routed **write** (`bd update <tgt-id> --title …`, or `comment`/`note`/`reopen`) **from the source rig**.

- **Expected:** the routed write commits to the target rig — `tgt` HEAD advances and the change persists.
- **Actual (buggy window):** `embeddeddolt: store is read-only` — the routed write resolved the target through the read-only factory (`OpenReadOnly`), and the commit guard refused it.

### Why it no longer reproduces on `main`

The routed read/write split now lives in `resolveViaPrefixRoutingWithAccess(ctx, id, writable bool)` (`cmd/bd/routed.go`): mutation commands go through `resolveAndGetIssueForMutation` / `close.go` with `writable=true` and open `newDoltStoreFromConfig` (writable), while reads keep `newReadOnlyStoreFromConfig`; auto-routed *contributor* stores stay read-only regardless (GH#2345). That's a different (and broader) mechanism than this PR's `OpenWritable`, and it rewrote the same `routed.go` / `close.go` code this branch touches — which is why the branch now shows conflicts.

I re-ran both tests against current `origin/main` (no changes from this PR) and both pass, so the regression is closed on `main`.

Given that, I think this PR is best **closed as superseded** rather than rebased. Happy to close it, or leave it to you — whichever you prefer. Thanks for the sweep!

PR #4372, opened 2026-06-11
metrics: instrument commands and migrate to GA4 via eventkit

coffeegoddd:
## Summary

- Refactor every `cmd/bd` command to use Cobra `RunE` and emit a `cli_command` event via eventkit, attributed with the command name.
- Promote the `config` package to a process-wide global so command code can read settings without threading a viper instance through every call site.
- Send metrics via the GA4 Measurement Protocol. No secrets ship in the binary or user config.
- `bd init` emits one of `init-embedded`, `init-server`, `init-shared-server`, `init-proxied-server` as the `command` value (no separate `dolt_mode` param) so GA4 dashboards can group by `command` alone.
- Ensure `~/.config/bd/config.yaml` exists with a complete `metrics:` block on every command (pre-run), not just `bd init`. Existing values are preserved; only missing leaves are filled. Failures (broken `~/.config`, malformed YAML) fail loudly.
- Fix `bd config set` for dotted keys against nested YAML (e.g. `bd config set metrics.disabled true` against a file with a `metrics:` block was creating a duplicate flat key instead of updating the nested leaf).

## Config surface

- Default endpoint: `https://gastownhall-eventsapi.com/mp/collect`.
- Env override: `BEADS_METRICS_ENDPOINT` (highest precedence).
- Config key: `metrics.endpoint`.
- Disable: `BD_DISABLE_METRICS=1` or `metrics.disabled: true`. Env beats config; resolution is fresh on every invocation.
- Bootstrap (any command, missing file):
  ```yaml
  metrics:
    disabled: false
    endpoint: https://gastownhall-eventsapi.com/mp/collect
  ```

## Test plan

- [x] `go build ./...` clean; `golangci-lint run ./internal/config/... ./internal/metrics/... ./cmd/bd/...` reports 0 issues
- [x] `go test ./cmd/bd/... ./internal/metrics/... ./internal/config/...` green
- [x] `bd send-metrics` consumes `.evtq` files successfully
- [x] `cli_command` events visible in GA4 Realtime with per-mode `init-*` command values
- [x] Pre-run bootstrap creates `~/.config/bd/config.yaml` on any command (verified with `bd version` from a fresh `HOME`); preserves user values when only some `metrics:` leaves are present
- [x] `BD_DISABLE_METRICS=1` set on a later command (after `bd init` ran with metrics enabled) suppresses new `.evtq` files
- [ ] Manual smoke on a fresh checkout (no pre-existing config) by a reviewer


maphew:

maphew:
Did a review pass on this as part of an upstream review-requested sweep. There's clearly useful work here, and since it's your own branch this is meant as "things I'd want resolved before it lands," not a request to hand it off.

The things that stood out:

1. **Telemetry default.** The instrumentation appears to default to sending to a hosted endpoint (GA4 via eventkit), while the existing docs state there's no outbound telemetry. That's the highest-risk item — I'd want the default to be explicitly opt-in (or an explicit, documented decision to the contrary) before any events leave the machine.
2. **Privacy/schema docs.** There's no event-schema doc/registry or a check to keep it honest. Worth landing the schema + privacy doc *before* telemetry is enabled, so what's collected is reviewable.
3. **Command-surface churn.** ~231 `NewCommandEvent` call sites across ~173 `cmd` files is a large blast radius to review and maintain. A centralized Cobra command-path instrumentation point would get most of the coverage with far less surface.
4. **Reviewability.** The PR is currently conflicting with `main` and large enough that Copilot declined (>20k lines). Splitting into reviewable pieces — (a) the dotted-YAML config fix, (b) `internal/metrics` + a hidden `send-metrics`, (c) consent/default redesign, (d) docs/privacy/schema, (e) the command instrumentation itself — would make each piece landable on its own merits.
5. **CI blockers** to clear regardless: a cgo-tagged helper used from a non-cgo path, Nix vendor-hash drift, and the embedded routed-update failure.

Happy to help with any of the split pieces if that's useful. The config fix and the metrics scaffolding look like the easy wins to land first.

maphew:
Closing as superseded, with thanks — this work shipped via #4419 (merge 8cdb83d7e, 2026-06-23), which cherry-picked this branch's commit lineage with your authorship preserved (d7d7a3460, author: coffeegoddd), plus maintainer follow-ups on top:

- first-run consent notice + `bd metrics` command (00d317dde)
- `DO_NOT_TRACK` honored as an opt-out (#4938)
- output-helper and config precedence/unset fixes addressing the review items raised here

The branch itself is 630 commits behind and fully conflicting with the absorbed variant on `main`, so a rebase would be a re-implementation of what already landed.

One residual from the review discussion that #4419 only partially addressed: a formal event-schema/privacy doc (what exactly is emitted, with what identifiers, and where that's documented). If you think that gap is real, it deserves a fresh small issue/PR against main rather than a revival of this branch — happy to help land it.

Thanks for the feature — it's shipped and in daily use.

PR #4373, opened 2026-06-12
Fix rig bead durability

silvanshade:
## Summary

- Remove `rig` from the default infra type set so newly-created `type=rig`
  beads are durable issues instead of ephemeral wisps.
- Keep `rig` excluded from ready work through explicit ready-work filtering.
- Fix the infra-to-wisps migration behavior so migration code does not classify
  rig identity beads as infra.
- Update CLI/help/reference text that described rigs as default infra types.

## Root Cause

`domain.DefaultInfraTypes()` included `rig`. Create paths use infra type
classification to mark beads ephemeral and route them to the wisp path. That
made rig identity beads transient even though rig identity is durable workspace
state.

Ready-work exclusion is a separate concern. `type=rig` should remain hidden from
`bd ready`, but it should not be ephemeral.

## Change Details

- Removes `rig` from default infra types.
- Keeps `rig` out of ready work via explicit ready exclusions.
- Updates the infra-to-wisps migration path so it does not treat `rig` as infra.
- Adds tests proving rig beads are persistent but not ready work.
- Updates docs and command help for the corrected infra type defaults.

## Validation

- PASS:
  `go test -tags gms_pure_go ./internal/storage/schema ./internal/storage/domain ./internal/storage/domain/db ./internal/storage/dolt -run 'TestMigration0035|TestDefaultInfraTypes|TestDomainDB/TestIssueGetReadyWork/ExcludesDefaultTypes|TestRigIssueIsPersistentButHiddenFromReady'`

- Earlier focused validation also passed:
  `go test ./internal/storage/domain ./internal/storage/domain/db ./internal/storage/dolt -run 'TestDefaultInfraTypes|TestDomainDB/TestIssueGetReadyWork/ExcludesDefaultTypes|TestRigIssueIsPersistentButHiddenFromReady'`

- Broader `make test` was attempted earlier and failed in unrelated/local
  environment-heavy tests recorded on `bds-vwz`.

- `golangci-lint run ./...` was not run because `golangci-lint` was not
  installed in the worker environment.

## Related

- Umbrella post-mortem issue: https://github.com/gastownhall/gastown/issues/4254
- Gastown companion PR: https://github.com/gastownhall/gastown/pull/4251

### DIFF
--- a/cmd/bd/count.go
+++ b/cmd/bd/count.go
@@ -283,7 +283,7 @@ Examples:
 //     --include-templates);
 //   - gate beads are excluded unless gates are explicitly requested via
 //     --type gate (list's default without --include-gates);
-//   - counting an infra type (agent/rig/role/message, or the store-configured
+//   - counting an infra type (agent/role/message, or the store-configured
 //     set) routes to the ephemeral wisps tier, like list's infra-type listing.
 //
 // The non-flag path never calls this function: bd count without
--- a/cmd/bd/export.go
+++ b/cmd/bd/export.go
@@ -30,7 +30,7 @@ For supported full backup/restore flows, use 'bd backup init', 'bd backup sync',
 and 'bd backup restore'.
 
 By default, exports only regular issues (excluding infrastructure beads
-like agents, rigs, roles, and messages). Use --all to include everything.
+like agents, roles, and messages). Use --all to include everything.
 
 Memories (from 'bd remember') are excluded by default because they may
 contain sensitive agent context. Use --include-memories or --all to
@@ -92,7 +92,7 @@ func runExport(cmd *cobra.Command, args []string) error {
 	// Build filter for issues table. Export all statuses by default.
 	filter := types.IssueFilter{Limit: 0}
 
-	// Exclude infra types by default (agents, rigs, roles, messages)
+	// Exclude infra types by default (agents, roles, messages).
 	if !exportAll && !exportIncludeInfra {
 		var infraTypes []string
 		if store != nil {
--- a/cmd/bd/export.go
+++ b/cmd/bd/export.go
@@ -58,7 +58,7 @@ var (
 func init() {
 	exportCmd.Flags().StringVarP(&exportOutput, "output", "o", "", "Output file path (default: stdout)")
 	exportCmd.Flags().BoolVar(&exportAll, "all", false, "Include all records (infra, templates, gates, memories)")
-	exportCmd.Flags().BoolVar(&exportIncludeInfra, "include-infra", false, "Include infrastructure beads (agents, rigs, roles, messages)")
+	exportCmd.Flags().BoolVar(&exportIncludeInfra, "include-infra", false, "Include infrastructure beads (agents, roles, messages)")
 	exportCmd.Flags().BoolVar(&exportScrub, "scrub", false, "Exclude test/pollution records")
 	exportCmd.Flags().BoolVar(&exportIncludeMemories, "include-memories", false, "Include persistent memories (from 'bd remember') in the export")
 	exportCmd.Flags().BoolVar(&exportNoMemories, "no-memories", false, "Exclude persistent memories (deprecated: now the default)")
--- a/cmd/bd/list.go
+++ b/cmd/bd/list.go
@@ -723,8 +723,8 @@ func init() {
 	// Gate filtering: exclude gate issues by default (bd-7zka.2)
 	listCmd.Flags().Bool("include-gates", false, "Include gate issues in output (normally hidden)")
 
-	// Infra type filtering: exclude agent/rig/role/message by default
-	listCmd.Flags().Bool("include-infra", false, "Include infrastructure beads (agent/rig/role/message) in output")
+	// Infra type filtering: exclude agent/role/message by default
+	listCmd.Flags().Bool("include-infra", false, "Include infrastructure beads (agent/role/message) in output")
 
 	// Explicit type exclusion
 	listCmd.Flags().StringSlice("exclude-type", nil, "Exclude issue types from results (comma-separated or repeatable, e.g., --exclude-type=convoy,epic)")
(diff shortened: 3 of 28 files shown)


maphew (commented):
Thanks. The product intent looks right to me: `type=rig` should be durable issue state, while still hidden from `bd ready`. I do not think this branch should merge as-is, but it is a good fix-merge candidate.

Findings:

1. This edits frozen migration `0035` instead of adding a new forward migration. The changed up migration excludes `rig`, but already-upgraded databases will not re-run `0035`; migration application stops once the recorded version is current. `scripts/check-migration-hygiene.sh` rejects this pattern for shipped migrations. Existing `type=rig` wisps need a new next-version migration that moves the rows and side tables back to durable `issues`.

2. Until that repair migration exists, legacy rig wisps are less protected. The PR removes `rig` from default infra types, while wisp cleanup protection still keys off configured infra types. That is correct after durable repair, but unsafe for existing `rig` rows left in `wisps`.

3. The branch is stale against current `upstream/main`. Merge simulation conflicts in the ready-work code. Current main moved ready-work predicate construction into `internal/storage/sqlbuild`, so the durable-but-hidden `rig` exclusion should be rebased into that shared builder rather than the older file.

Non-blocking: `cmd/bd/export.go` still has a comment saying default infra includes rigs, even though the code no longer does.

Recommended path: fix this on the contributor branch if practical. Restore frozen `0035`, add a new forward migration for existing rig wisps, and move the ready-work exclusion into the current `sqlbuild` path. The related `gastownhall/gastown#4254` context still supports this direction.

Checks run locally:

- `bd-main/scripts/pr-preflight.sh 4373 --repo gastownhall/beads`
- `scripts/check-migration-hygiene.sh` failed on the frozen `0035` edit
- `go test -tags gms_pure_go ./internal/storage/schema ./internal/storage/domain ./internal/storage/domain/db ./internal/storage/dolt -run 'TestMigration0035|TestDefaultInfraTypes|TestDomainDB/TestIssueGetReadyWork/ExcludesDefaultTypes|TestRigIssueIsPersistentButHiddenFromReady' -count=1`
- same package set without `-run`

No code was pushed from this review.

maphew:

silvanshade:
> Could you add the smallest command sequence or setup that triggers this, including the `bd version`, storage mode, OS, and the expected vs. actual output? A scratch-repo repro is ideal if one is possible.

Thanks for the responses. I don't have the capacity to investigate further and provide a full from-scratch minimal reproducible case now.

Here's what I can tell you though, I was using:

- beads `1.0.4` (downloaded from GitHub releases)
- gastown `1.2.1` (downloaded from GitHub releases)

Both installed through `mise` using the `github:` purl for the release artifacts. OS was macOS 26.5.1.

I basically just followed the instructions [here](https://github.com/gastownhall/gastown#setup-docker-compose-below) to set up a new gastown:

```sh
# Create workspace with git initialization
gt install ~/gt --git
cd ~/gt

# Add your first project
gt rig add myproject https://github.com/you/repo.git
```

This would always immediately complain about the `bd init` part of the `gt install` although I don't remember the details now.

Then `gt doctor` would always complain about a few things too, like stale Claude settings (which it had just created), maybe beads again, can't remember.

I'd be able to eventually resolve those problems and get a fully green `gt doctor`.

But then after actually performing `gt mayor attach` and asking it to sling some beads, eventually things would fall over.

The polecats wouldn't hook the beads properly, there'd be a mis-assigned deacon (missing the trailing `/`, another known issue). Eventually `gt doctor` would start complaining mysteriously about things like the missing rig identity beads. I don't recall when exactly that happened and if it required a `gt down` then `gt up` first or not.

Most of the time I tried using `gt doctor --fix` it would not fix the problems, nor would it report why (honestly I think the failures to fix should force a `--verbose` display or something).

I also noticed that when I tried to add a second rig, I encountered a different class of failure modes. The rig config was missing the schema types and I'd get other error messages.

Maybe I somehow was doing things incorrectly but I tried setting up 3 completely fresh towns, even with separate fresh repos as rigs, and kept encountering those problems.

Eventually I just decided to not use gastown because I couldn't get it to work reliably. The issue I linked:

https://github.com/gastownhall/gastown/issues/4254

with all the referenced PRs, appeared to potentially fix the main problems I kept encountering, but by then I'd burned a lot of usage capacity just trying to get the system to work well enough to pass some basic smoke tests, and determined it wasn't reliable enough to continue with.

Outside of gastown integration, beads seems to be working okay for me and I'm using it to manage several projects.

maphew:
Thank you for the as detailed-as-can-remember followup, it's appreciated.

Yeah, Gastown has many rough edges. I don't use it myself (yet) as I have my head and hands full just grappling with beads itself. Gas City is much easier to get up and running with, though it's been a month since I ran through that set up.

maphew:
Maintainer update: I took this over and pushed a rebased branch at `747c3a25`.

What changed:

- Rebased onto current `upstream/main` and resolved the ready-work conflict in the current `sqlbuild` path.
- Kept frozen migration `0035` at its upstream content, then added `0053_repair_rig_wisps` as the forward data repair for existing `type=rig` rows already routed into `wisps`.
- Kept `type=rig` durable while excluding it from ready work through the shared ready-work filter.
- Updated the remaining CLI/docs wording that still grouped rigs with transient infra beads.

Validation run locally:

- `scripts/check-migration-hygiene.sh`
- `go test -tags gms_pure_go ./internal/storage/schema -count=1`
- `go test -tags gms_pure_go ./internal/storage/domain ./internal/storage/domain/db ./internal/storage/sqlbuild -count=1`
- `go test -tags gms_pure_go ./cmd/bd -run 'TestApplyCountIncludeInfra|TestEmbeddedExport|TestExportToFile|TestExportToStdout|TestWispFilteringFromExport|TestEmbeddedList|TestListCommandSuite|TestWispCreate' -count=1`

I also invoked the Dolt-server-only rig tests, but this Windows session skipped them because the test Dolt server/Docker path is unavailable. I queued the long local verifier as `mybd-m3yi` for `make test` at head `747c3a25`.

This now matches the charter direction I was looking for: rig identity is durable issue data, while ready-list hiding stays workflow/presentation filtering rather than making rig an ephemeral orchestration primitive.
_codex-unknown-model-unknown-reasoning on behalf of matt wilkie_

Issue #593, opened 2025-12-16
bd prime opinionated git ops


mhyrr:
**Tl;dr**: Add some flexibility to how/whether bd executes git commands.

The close session protocol described in https://github.com/steveyegge/beads/blob/8c07a4df44da8aaaad2480060e81ab1044c4b9f0/cmd/bd/prime.go#L192 drives the agent to execute the described git ops, especially after compacting. Even if your CLAUDE.md file has something like: 

```
## Git Policy
  - NEVER commit changes unless I explicitly ask you to commit
  - Only stage and commit when I say "commit", "make a commit", or similar
```

The close session protocol seems to always override this. Would prefer to extract the close session protocol to define a user-driven set of steps.

steveyegge:
There's already a `--stealth` flag for this:

```bash
bd prime --stealth  # No git operations, flush only
```

If you set up Claude with stealth mode, the hooks use this:
```bash
bd setup claude --stealth
```

This makes `bd prime` output the "close session" section without git commands - it'll just say to run `bd sync --flush-only`.

**However**, I understand you might want more granular control - like keeping beads git-tracked but still controlling when commits happen. 

For now, the workaround is to use `--stealth` mode or edit your hooks to use `bd prime --stealth`. 

A config option like `prime.no-git-ops: true` could be added if there's demand. Adding the triaged label for now - let me know if `--stealth` works for your use case!

mhyrr:
Thanks for the reply and thanks for beads!

For what it's worth, I'm not interested in running in stealth mode, I simply don't want the close session protocol to run. So your `prime.no-git-ops: true` would be perfect.

The place I notice this the most: after compaction. If whatever work I'm doing kicks past the context window and compacting occurs, Claude Code tries to commit immediately after completing the next time around and I find myself writing "don't run the close session protocol!" during the compaction because I'm not ready for the commits/push to occur until work is verified, etc.

steveyegge:
Fixed in commit 2ff17686.

You can now disable git operations in the session close protocol via config:

```bash
bd config set no-git-ops true
```

After setting this, `bd prime` will output:
```
[ ] bd sync --flush-only    (export beads to JSONL only)
```

Instead of the full git add/commit/push workflow.

This lets you control when commits happen manually, preventing the auto-commit behavior after context compaction that you mentioned.

PR #4375, opened 2026-06-12
fix(stealth): keep --stealth out of the tracked .gitignore

HackAttack:
bd init --stealth correctly routed .beads/ and .claude/settings.local.json into .git/info/exclude, but the later GH#2034 project-root .gitignore step (doctor.EnsureProjectGitignore) ran unconditionally, creating or modifying a tracked .gitignore with the Dolt-file patterns. That change shows up in git status and is committable, defeating the whole point of stealth mode.

In stealth mode those patterns now go to .git/info/exclude instead, via a path-aware addProjectPatternsToGitExclude (resolved through --git-common-dir so worktrees hit the main repo's exclude file). The non-stealth path is unchanged.

bd doctor was leaking the same way: the "Project Gitignore" check and --fix called CheckProjectGitignore/FixProjectGitignore unconditionally, so doctor would re-create the .gitignore that stealth init deliberately avoided. Detection keys off the persisted no-git-ops flag (the same signal bd prime uses); stealth repos now check and fix .git/info/exclude instead. The doctor package stays config-agnostic, so the stealth branching lives in the cmd layer and reuses the "Project Gitignore" check name to keep --fix dispatch and ordering stable.

Tests: unit coverage for the exclude routing, stealth detection, and the stealth doctor check; end-to-end TestEmbeddedInit subtests assert that both bd init --stealth and bd doctor --fix leave git status --porcelain empty.

### DIFF
--- a/cmd/bd/doctor.go
+++ b/cmd/bd/doctor.go
@@ -681,9 +681,14 @@ func runDiagnostics(path string) doctorResult {
 	result.Checks = append(result.Checks, gitignoreCheck)
 	// Don't fail overall check for gitignore, just warn
 
-	// Check 14a: Project-root .gitignore has Dolt exclusion patterns (GH#2034)
-	projectGitignoreCheck := convertWithCategory(doctor.CheckProjectGitignore(path), doctor.CategoryGit)
-	result.Checks = append(result.Checks, projectGitignoreCheck)
+	// Check 14a: Project-root Dolt exclusion patterns (GH#2034). In stealth mode these live in
+	// .git/info/exclude, so check that location instead to avoid recreating .gitignore.
+	if isStealthRepo(path) {
+		result.Checks = append(result.Checks, convertWithCategory(checkProjectExcludeStealth(path), doctor.CategoryGit))
+	} else {
+		projectGitignoreCheck := convertWithCategory(doctor.CheckProjectGitignore(path), doctor.CategoryGit)
+		result.Checks = append(result.Checks, projectGitignoreCheck)
+	}
 	// Don't fail overall check for project gitignore, just warn
 
 	// Check 14b: redirect file tracking (worktree redirect files shouldn't be committed)
--- a/cmd/bd/doctor_fix.go
+++ b/cmd/bd/doctor_fix.go
@@ -255,10 +255,13 @@ func applyFixList(path string, fixes []doctorCheck) {
 		case "Gitignore":
 			err = doctor.FixGitignore(path)
 		case "Project Gitignore":
-			// Stealth repos must not get a tracked .gitignore; route the patterns into
-			// .git/info/exclude instead (matches bd init --stealth).
+			// Stealth / no-git-ops repos must not get a tracked .gitignore; route the patterns into
+			// .git/info/exclude instead (matches bd init --stealth) and strip any beads section a
+			// previous run leaked into the tracked .gitignore so stealth leaves no trace.
 			if isStealthRepo(path) {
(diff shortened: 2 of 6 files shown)


maphew:

HackAttack:
`bd init --stealth` modifies .gitignore, which is not stealthy because .gitignore is shared by all users of the repo. The appropriate place for user-specific ignores is .git/info/excludes.

HackAttack:
I expect `--stealth` to mean “no evidence of beads will be exposed to anyone else.”

maphew:
Thanks for the clarification. The goal this far hasn't been for perfect invisibility, rather just not being noisy about it. I'll take another look.

maphew:
## Review + maintainer follow-ups (pushed to this branch)

Thanks @HackAttack — you're right that "stealth" should mean *no trace for collaborators*, not just "less noisy". I reviewed the PR, agreed with that stronger definition, and pushed maintainer edits directly to this branch (your original commit is preserved; mine is on top with a `Co-authored-by` trailer).

### Review of the original change
- **Correct root-cause fix.** Both leak paths are closed: `bd init --stealth` no longer calls `EnsureProjectGitignore`, and `bd doctor --fix` no longer re-creates the tracked `.gitignore`.
- `resolveGitExcludePath` correctly handles worktrees via `--git-common-dir` and anchors git's relative output back to `repoPath` — a real subtlety, handled well.
- Strong regression tests (`git status --porcelain` empty + byte-for-byte `.gitignore` unchanged) for both init and doctor.

### What I changed on top
1. **Remediate already-leaked repos.** Prevention alone left a gap: a repo initialized as stealth by an older `bd` still has the Dolt patterns in a *tracked* `.gitignore`. The stealth doctor check now flags that leak, and both `bd doctor --fix` and a re-run of `bd init --stealth` strip the bd-managed section (header + the patterns beads added beneath it). Unrelated user patterns are preserved, and the file is removed entirely when beads was its only content — restoring true stealth.
2. **De-duplicated the exclude writers.** `setupGitExclude`, `setupForkExclude`, and `addProjectPatternsToGitExclude` now share one `addExcludePatterns` core built on `resolveGitExcludePath` (was three near-identical read/dedupe/append/write blocks).
3. **Clarified the stealth signal.** Detection keys off the persisted `no-git-ops` flag — the same signal `bd prime` uses (#593) — so an explicit `no-git-ops` user is covered too. Made the exclude section header neutral since `bd doctor --fix` writes it as well, not just `bd init --stealth`.
4. **Fixed CI.** Resolved the `gosec` G702 finding on the variadic `git` invocation, and rebased onto `main` to clear the unrelated `docs/CLI_REFERENCE.md` drift that was failing the policy/build/doc-flags checks.
5. **Tests** for remediation (preserve user content, delete-when-only-beads, no-op without a section) and the leaked-`.gitignore` doctor warning.

### Verification (local)
`gofmt` clean · `go vet` clean · `golangci-lint` 0 issues · de-dupe + remediation unit tests pass · `check-doc-flags` and full `ci-pr-policy` pass. Embedded-Dolt e2e tests are blocked locally by the schema-migration gate (#4259), so I'm relying on CI for those.

Will squash-merge once CI is green, crediting you as author.

maphew:
Merged via squash. Thanks again @HackAttack — your bug report and fix are the substance of this change; you're credited as author on the squashed commit. The maintainer additions (leak remediation, exclude-writer de-dupe, gosec fix, rebase for the unrelated doc drift) ride on top. Full CI is green, including the embedded-Dolt matrix. The one Lint failure along the way was a transient Go-proxy fetch error, cleared by a re-run.

PR #4376, opened 2026-06-12
fix(resolver): use last-segment prefix match to prevent wisp substring collision

sarendipitee:
## Problem

When two beads share overlapping hash fragments — e.g. a real bead `gcy-oqf` and an ephemeral wisp `gcy-wisp-goqfo` — the fuzzy resolver's `strings.Contains` check matched `"goqfo"` as containing the substring `"oqf"` and silently returned the **wrong** bead. No ambiguity error was raised because there was only one (wrong) match.

Repro (observed in production, gascity-source rig):
```
# Two beads exist: gcy-oqf (real bug, open) and gcy-wisp-goqfo (ephemeral, closed)
bd show gcy-oqf   # → silently showed gcy-wisp-goqfo
bd close gcy-oqf  # → re-closed the already-closed wisp; real bead stayed open
```

Root cause: the exact-ID fast path can fail on transient cache lag. The substring fallback then finds `gcy-wisp-goqfo` because `"wisp-goqfo"` contains `"oqf"`, and returns it silently as the single match.

## Fix

Change the fuzzy match from substring containment to **last-dash-segment prefix comparison**:

| Input | Issue hash | Old (Contains) | New (last-seg HasPrefix) |
|-------|-----------|----------------|--------------------------|
| `oqf` | `oqf` | ✓ exact (unchanged) | ✓ exact (unchanged) |
| `oqf` | `wisp-goqfo` | ✓ (BUG: contains "oqf") | ✗ (last seg "goqfo" not prefixed by "oqf") |
| `t3st` | `wisp-t3st` | ✓ | ✓ (last seg "t3st" = prefix "t3st") |
| `t3` | `wisp-t3st` | ✓ | ✓ (last seg "t3st" starts with "t3") |

This preserves the intended use case (resolving wisps by their trailing segment) while eliminating false substring hits.

## Also: BD_NO_FUZZY env var

Added `BD_NO_FUZZY=1` environment variable. When set, partial resolution is entirely skipped after the exact-ID fast path. This lets callers opt out of all fuzzy matching (e.g. gc's internal beads store calls that should never silently redirect).

## Tests

- `TestHashSegmentPrefixMatch` — unit tests for the new helper covering the repro case and valid cases
- `TestResolvePartialID_WispSubstringCollision` — integration test for the exact gcy-g4o repro
- `TestResolvePartialID_WispLastSegment` — confirms wisp-by-last-segment still works
- `TestResolvePartialID_BdNoFuzzy` — confirms BD_NO_FUZZY behavior

Fixes: gcy-g4o

## Prefix-vs-substring semantics (intentional)

`hashSegmentPrefixMatch` narrows plain (non-wisp, no-dash) partial-ID matching from substring-anywhere (old `strings.Contains`) to last-dash-segment prefix-only (`strings.HasPrefix`). This is deliberate: substring-anywhere is what allowed the wisp-collision footgun. Tail/middle fragments (e.g. `f8e9` of `a3f8e9`) no longer match; only genuine prefixes do. Pinned by new test `TestHashSegmentPrefixMatchPlainHashPrefixOnly`.

_Note: branch was rebased onto upstream/main; the #4150 loop-variable rename (`issue.ID`/`w.ID` → `id`/`wID`) conflict was resolved keeping `hashSegmentPrefixMatch`._


maphew (commented):
Thanks for tracking down this one - the repro is convincing and nasty: a fuzzy ID lookup silently returning the *wrong* bead (rather than erroring on ambiguity) is a correctness footgun, and the fix is small and well targeted. I confirmed on current upstream/main that internal/utils/id_parser.go still uses strings.Contains(issueHash, hashPart) in both the issue and wisp fallback loops, so the bug this PR fixes is still live - it hasn't been superseded by other work. I also checked #4472 (cross-table issue/wisp ID collision rejection): that PR prevents a *new* collision from being created at write time, while this PR hardens the *read-time* resolver against any pre-existing or promotion-path collision. The two are complementary, not competing - no overlap to reconcile.

**Findings**

- [major] The last-dash-segment prefix comparison (hashSegmentPrefixMatch) is applied to every hash, not just compound wisp hashes. For a plain hash with no dash, the trailing-segment is the whole hash, so this quietly changes ordinary partial-ID matching from "substring anywhere" (old Contains) to "prefix only" (HasPrefix). A lookup like resolving bd-a3f8e9 by typing the tail fragment f8e9 worked before and would stop matching after this lands. That may well be intentional - the function's own doc comment only illustrates prefix-style partial IDs - but it's a broader behavior change than the PR description states (which frames this purely as a wisp-collision fix), and there's no test pinning the new prefix-only behavior for a non-wisp hash. Suggest a sentence in the description plus one small test asserting this is deliberate.

**Conflict**

This is an ordinary rebase conflict, not the pre-rewrite orphan-branch situation - git merge-base upstream/main <head> finds a real, recent common ancestor (2026-06-10), well after the history rewrite. #4150 (merged 2026-07-04) refactored the same two fallback loops this PR touches, renaming the loop variables from full objects (issue.ID, w.ID) to ID-only strings (id, wID) for a performance win on large stores. This PR's hunks still reference the old names. Fix: git fetch upstream main && git rebase upstream/main, then in each conflict keep this PR's hashSegmentPrefixMatch(...) call but swap in the current loop variable (id/wID) in place of issue.ID/w.ID. It's a two-line mechanical resolution per hunk, no logic rework needed.

**Recommendation:** merge-fix. Please rebase onto upstream/main and resolve the two variable-name hunks as described above, and consider adding the one plain-hash test noted above (or a short description update if the broadened behavior is intentional). Once green, this looks ready - thank you for the clear repro and the regression tests, they made this fast to verify.

Next steps for the contributor:
1. git fetch upstream main && git rebase upstream/main
2. Resolve internal/utils/id_parser.go keeping hashSegmentPrefixMatch(...), using id/wID (not issue.ID/w.ID)
3. Re-run go test ./internal/utils/...
4. Optionally note the prefix-vs-substring semantics broadening in the PR description

sarendipitee:
@maphew Rebased onto upstream/main and resolved the two loop-variable hunks per your guidance (`id`/`wID`), keeping `hashSegmentPrefixMatch`. Added `TestHashSegmentPrefixMatchPlainHashPrefixOnly` pinning the intentional prefix-only (vs old substring-anywhere) semantics. `go test ./internal/utils/...` green.

maphew:
Your diagnosis here was exactly right, and it's worth restating precisely because of what happened next: `strings.Contains` in partial-ID resolution silently returned the wrong bead when a search term appeared *inside* another hash (your gcy-g4o case: "oqf" matching inside wisp hash "goqfo"), and prefix-anchored matching was the correct repair. Thanks also for the July 6 rebase and the pinned-semantics test after the earlier guidance — that follow-through was noticed.

The reason this PR is now due to retire rather than merge: main independently landed the same protection in `3f2d64c2e` ("reject substring matches in ID resolution", #4939, tracking issue 4234) — leading-prefix matching with the `wisp-` infix stripped, and regression tests covering exactly your bug shape (a wisp hash containing the search term as an interior substring now errors instead of resolving). Your branch and main converged on the same semantics by different mechanisms (your last-dash-segment helper vs. main's prefix-after-strip), and with main's version already shipped with tests, carrying a second mechanism would just be drift risk.

If you see a case your `hashSegmentPrefixMatch` handles that the landed fix doesn't — compound-hash searches like `wisp-goq` are the one marginal difference I could construct — say so here and we'll pin it with a test on main. Otherwise this closes in a few days with the diagnosis credited to you and gcy-g4o.

sarendipitee:
@maphew It's pretty frustrating that a PR 5 days old gets merged and supersedes this one (over 40 days old).

That means other people were having this same issue for months and lord knows how many more that didn't report or say anything until it took a second person to PR _the same thing_. 

It also causes me more burden to maintain PRs on my forks as they drift further and further from main. Every PR I make is rebased off of upstream/main at the time of open so every time I update my systems I have dozens of PR branches I have to carefully rebase indepedently against different lines of work. Fine, that's software development, I'm not going to be pushy about how fast things get reviewed in OSS. 

But why the hell wasn't this merged in the first place and why was that other PR merged _**first**_? Why are we not looking at PR candidates _**chronologically**_? Why are there no automatic checks for new PRs for already open PRs / issues before they are getting fully merged. That's what is so frustrating.

I have 50+ PRs across gastownhall repos and stuff like this makes me question whether i want to keep contributing.

maphew:
Thanks for taking the time to express your frustration @sarendipitee . Truly, I appreciate it. 

I fully admit the PR triage and processing is all over the place, haphazard. I've been watching the number tick steadily up with dread, especially since the mess I made at release v1.0.4 which put the brakes on most of it and we saw the total double in a few weeks. Issue processing is even worse, which is, well, worse! 

Anyway, I can't speak for the other maintainers but I personally been working across themes not PR age. Starting with the ones I understand the best and then working outward. Formula, molecules and wisps are in the outer reaches for me and consequently I haven't touched them much.

I do start sessions along the lines of _"bring beads repo up to date with origin and upstream, then process ready items. Check for freshness against gh repo before initiating work"_ and _"before selecting each task, reconcile ready and in-progress Beads with open GitHub issues, branches, and PRs. Identify duplicate, overlapping, conflicting, or already-implemented work"_. However I've seen agents simply stop at 50 or 100 comparisons though and I'm guessing that's what's happened here with yours.

After a concerted effort in the last week, several hours a day, I've finally gotten the total PR into the low hundreds and will turn the dial towards working more on the oldest regardless of theme.

maphew:

Resolution notes: main had since refactored the resolver signature (`PartialIDResolverStore` interface instead of concrete storage) and consolidated the prefix-match branches into an if/else-if chain; your `hashSegmentPrefixMatch` fix is applied inside that newer structure at both the plain-ID and wisp-fallback sites, and your `BD_NO_FUZZY` doc comment is kept. All resolver tests pass locally, including the wisp substring-collision cases this PR exists for.

maphew (commented):
Reviewed exact head `ae08eb538210d234b88947be30de20fe6e2b6443`.

Thanks for the follow-through. The open-PR sweep confirms that later #4415 duplicated these resolver hunks inside unrelated flat-file work. Under first-filed policy, #4376 should remain canonical and #4415's transformed successor should drop those hunks.

One test still needs repair before landing: `TestResolvePartialID_WispSubstringCollision` inserts `gcy-oqf` and resolves that exact full ID, so the exact-search fast path returns before the changed fuzzy fallback runs. The test therefore passes even with the old buggy matcher. Please simulate the reported cache-lag state—such as exposing only `gcy-wisp-goqfo` and asserting `gcy-oqf` is not resolved, or using a wrapper that hides the real bead—and verify that the pre-fix implementation fails.

Disposition: **maintainer-refresh onto current main, repair that regression test, preserve the contributor commits, and merge**. The later #4939 landing should not erase the original attribution.

PR #4377, opened 2026-06-12
fix(codex): correct hook path to resolve from plugin root

### LINKED ISSUE #4210: Codex-only hooks.json at package root causes non-blocking errors in Claude Code
## Summary

The beads plugin (v1.0.4) places `hooks/hooks.json` at the package root. This file calls `bd codex-hook <event>`, which is a subcommand that only exists on `main` (not in the v1.0.4 Homebrew binary). Claude Code's plugin loader discovers the root-level `hooks/hooks.json` and fires those hooks, producing a non-blocking error on every session start:

```
Failed with non-blocking status code: Error: unknown command "codex-hook" for "bd"
```

## Root Cause

Two issues combine:

1. **Version skew:** `codex_hook.go` was added to `main` after the v1.0.4 tag, so the Homebrew binary doesn't include the `codex-hook` subcommand.

2. **Hook file placement:** The Codex manifest (`.codex-plugin/plugin.json`) references `./hooks/hooks.json`, but the `hooks/` directory sits at the package root rather than inside `.codex-plugin/`. Claude Code's plugin loader discovers it there and registers the hooks alongside its own inline hooks from `.claude-plugin/plugin.json`.

## Current Behavior

- **Claude Code** fires both its inline `bd prime` hooks (from `.claude-plugin/plugin.json`) AND the Codex `bd codex-hook` hooks (from root `hooks/hooks.json`). The former succeed; the latter fail with "unknown command."
- **Codex** would also fail since the installed binary lacks `codex-hook`, but that resolves once v1.0.5 ships.

## Suggested Fix

Either:

- **Move `hooks/hooks.json` inside `.codex-plugin/hooks/hooks.json`** so Claude Code's loader doesn't discover it, OR
- **Gate the hooks on the binary version** — check that `bd codex-hook` exists before calling it, OR
- **Ship v1.0.5** with `codex-hook` compiled in, so both platforms work

## Environment

- `bd` v1.0.4 (Homebrew)
- beads plugin v1.0.4 (marketplace)
- Claude Code + Codex both installed
- macOS

christso:
## What

Fixes the Codex hooks path in `plugins/beads/.codex-plugin/plugin.json`. PR #4188 moved hook files from `plugins/beads/hooks/hooks.json` to `plugins/beads/.codex-plugin/hooks/hooks.json` to isolate Codex-only hooks from Claude. However, the manifest path was left as `"./hooks/hooks.json"`, which Codex resolves relative to the **plugin root** (`plugins/beads/`), not relative to `.codex-plugin/`.

This causes a runtime warning:

```
failed to read /home/<user>/.codex/plugins/cache/beads-marketplace/beads/1.0.5/hooks/hooks.json: No such file or directory
```

The actual file is at `.codex-plugin/hooks/hooks.json`. Correcting the path to `"./.codex-plugin/hooks/hooks.json"` lets Codex resolve it correctly, and preserves the harness-specific isolation introduced by #4188 (Claude's loader does not scan inside `.codex-plugin/`).

## Changes

- `plugins/beads/.codex-plugin/plugin.json`: `"hooks"` path `./hooks/hooks.json` -> `./.codex-plugin/hooks/hooks.json`
- `cmd/bd/setup/plugin_layout_test.go`: update `TestPluginLayoutUsesSharedBeadsRoot` assertion to match the corrected path

## Verification

- `jq empty plugins/beads/.codex-plugin/plugin.json plugins/beads/.codex-plugin/hooks/hooks.json` - both files parse cleanly
- `go test ./cmd/bd/setup -run TestPluginLayoutUsesSharedBeadsRoot -count=1` - **PASS**
- File still exists at `plugins/beads/.codex-plugin/hooks/hooks.json` (unchanged)
- No file at `plugins/beads/hooks/hooks.json` (unchanged, Claude protection intact)

Note: CI checks are pending maintainer approval to run (standard GitHub Actions behavior for first-time external contributors).

Closes #4210. Follow-up to #4188.

### DIFF
--- a/cmd/bd/setup/plugin_layout_test.go
+++ b/cmd/bd/setup/plugin_layout_test.go
@@ -47,8 +47,8 @@ func TestPluginLayoutUsesSharedBeadsRoot(t *testing.T) {
 	if codexManifest.Skills != "./skills/" {
 		t.Fatalf("Codex manifest skills path = %q, want ./skills/", codexManifest.Skills)
 	}
-	if codexManifest.Hooks != "./hooks/hooks.json" {
-		t.Fatalf("Codex manifest hooks path = %q, want ./hooks/hooks.json", codexManifest.Hooks)
+	if codexManifest.Hooks != "./.codex-plugin/hooks/hooks.json" {
+		t.Fatalf("Codex manifest hooks path = %q, want ./.codex-plugin/hooks/hooks.json", codexManifest.Hooks)
 	}
 
 	requireRepoFile(t, root, "plugins", "beads", "skills", "beads", "SKILL.md")
--- a/plugins/beads/.codex-plugin/plugin.json
+++ b/plugins/beads/.codex-plugin/plugin.json
@@ -16,7 +16,7 @@
     "agent-memory"
   ],
   "skills": "./skills/",
-  "hooks": "./hooks/hooks.json",
+  "hooks": "./.codex-plugin/hooks/hooks.json",
   "interface": {
     "displayName": "Beads",
     "shortDescription": "Project task tracking with bd",


kevglynn (commented):
Reviewed. The fix is straightforward and correct.

Verified locally:
- `plugins/beads/.codex-plugin/hooks/hooks.json` exists
- `plugins/beads/hooks/hooks.json` does NOT exist (confirmed the path was broken)
- The manifest's `hooks` field resolves relative to plugin root (`plugins/beads/`), so `./.codex-plugin/hooks/hooks.json` is the correct path

The PR description clearly explains the regression from #4188 (moved hooks for Claude isolation but didn't update the manifest path). Test assertion is updated to match.

Easy win — targeted bug fix, test updated, no side effects. LGTM.

PR #4382, opened 2026-06-12
fix(dolt): retire idle pool conns before server reaps them + retry reads

brandonmartin:
## Problem
After the dolt sql-server `wait_timeout` was lowered to 30s, long-lived daemons (e.g. an orchestrator's reconcile loop) intermittently fail reads with:

    begin read tx: invalid connection

The connection pool only configures `ConnMaxLifetime` (1h) and never sets `ConnMaxIdleTime`. So a connection that sits idle in the pool is reaped **server-side** after 30s, but the pool still believes it's good. The next query handed that dead connection fails with `invalid connection`, and because `withReadTx` calls `BeginTx` directly with no retry, the error propagates straight to the caller and can stall the daemon.

## Fix
1. **Retire idle connections before the server does.** Add `Config.ConnMaxIdleTime` and `defaultConnMaxIdleTime = 20 * time.Second`, applied via `db.SetConnMaxIdleTime` in `applyPoolLimits`, mirroring the existing `ConnMaxLifetime` default-fallback. The pool now closes an idle connection at 20s — before the 30s server reap — so a reaped connection is never handed out.
2. **Retry the transient case.** Wrap `withReadTx`'s `BeginTx`+`fn` in the existing `withRetry` helper. `isRetryableError` already matches `"invalid connection"`, so a residual race (or any other transient connection blip) is retried instead of surfaced. Safe because `fn` is read-only and the transaction is always rolled back, so re-running has no side effects. This matches how the write paths already use `withRetry`.

## Why 20s
It must be strictly below the server `wait_timeout` (30s) so the client retires the connection first, with enough margin for clock skew and the pool's sweep granularity. 20s gives a comfortable 10s buffer while keeping connections warm for the common case. Overridable via `Config.ConnMaxIdleTime`.

## Testing
- `go build ./...`, `go vet ./internal/storage/dolt/...` — pass
- `go test ./internal/storage/dolt/...` — pass (embedded Dolt, ~11s, no skips)
- Deployed in a downstream consumer (a long-running orchestrator) via a `replace`: the recurring `begin read tx: invalid connection` storm stopped immediately after rollout (0 new occurrences post-deploy vs ~1/min before).


maphew:

brandonmartin:
Sorry the description didn't carry a recipe — here's a scratch repro. The reason a normal sweep won't surface it: **it only reproduces against the Dolt _sql-server_ store, not embedded.** Embedded Dolt has no server-side `wait_timeout` connection reaping, so the unit tests (`go test ./internal/storage/dolt/...`, embedded) can't trigger it — you need a real sql-server whose `wait_timeout` is shorter than the pool's idle time.

Compress `wait_timeout` so it fires in seconds instead of 30s:

```bash
mkdir /tmp/bd-idle && cd /tmp/bd-idle && dolt init -q
dolt sql-server --host 127.0.0.1 --port 3307 &
sleep 2
# prod was 30s; 5s just shrinks the idle wait you have to sit through
mysql -h 127.0.0.1 -P 3307 -u root -e "SET GLOBAL wait_timeout=5; SET GLOBAL interactive_timeout=5;"

# point bd's native store at 127.0.0.1:3307 (sql-server mode), then warm the pool:
bd create "warm the pool" >/dev/null
bd list >/dev/null

sleep 8        # let the pooled read conn sit idle past wait_timeout -> server reaps it

bd list        # next read is handed the dead connection
```

- **Expected:** the final `bd list` returns normally.
- **Actual (pre-fix):** it fails with `begin read tx: invalid connection` (the pool still believes the reaped conn is good, and `withReadTx` calls `BeginTx` with no retry, so it propagates).

Env: `bd 1.0.5`, native Dolt **sql-server** store, Linux x86_64.

Prod was the same shape with the real numbers: server `wait_timeout=30s` vs an orchestrator reconcile loop holding read conns idle up to ~1h (`ConnMaxLifetime=1h`, no `ConnMaxIdleTime`) → ~1 failure/min. With this PR (`ConnMaxIdleTime=20s` < `wait_timeout`, plus the `withReadTx` retry) the final read stays green even at `wait_timeout=5s`, because the pool retires the idle conn before the server reaps it.

maphew:
Verified this locally — the fix is correct and low-risk. Thanks for the clear repro write-up; it pointed straight at the mechanism.

**Setup:** dolt 2.1.8 sql-server, go 1.26.4, `go-sql-driver/mysql v1.9.3` (the driver bd uses), reproducing at the `database/sql` pool layer where the fix lives.

**Proactive half (`SetConnMaxIdleTime`) — proven.** With `ConnMaxIdleTime=0` (pre-fix) the pool reuses the same server connection id after an idle period — exactly the connection the server reaps. With `ConnMaxIdleTime` set, the pool retires the idle connection and opens a fresh one before the server's `wait_timeout` can reap it. With prod's `20s < 30s` this is preventive and never reaches the error path.

**Retry half (`withReadTx` → `withRetry`) — correct.** `isRetryableError` already matches both `invalid connection` and `gone away` (MySQL 2006 idle timeout), so the wrap actually fires. The replay is safe here specifically because the read tx is read-only and always rolled back — no side effects to double-apply.

**One thing worth confirming:** `withReadTx` now flows through `withRetry`, which consults the circuit breaker. Reads previously didn't touch the breaker; they now contribute failures to it and can return `ErrCircuitOpen`. That looks intentional (parity with the write path), but flagging it as a behavior change for the read path in case it isn't.

**Caveat on my verification:** I couldn't reproduce the end-to-end *symptom* on dolt 2.1.8 — its server-side `wait_timeout` reaping didn't fire in my test, and a `KILL`-based simulation got absorbed by `database/sql`'s built-in bad-connection retry. So I verified the mechanism rather than the raw failure. CI on the branch is currently showing no reported checks — worth a green run before merge.

maphew:
Pushed a maintainer update to the contributor branch that merges current `upstream/main` into #4382 while preserving Brandon's original fix commit. The previous red checks were from the stale base branch (plus the doc freshness drift); after refreshing the branch, the failing embedded cmd shards now pass.

Local verification before push:

- `go test ./internal/storage/dolt/...`
- `BEADS_TEST_EMBEDDED_DOLT=1 go test -tags=gms_pure_go ./cmd/bd -run '^(TestEmbeddedDepBulkNoCycleCheckSkipsPerEdgeCycleValidation|TestEmbeddedUpdateRoutedStoreCommitsTargetHead|TestEmbeddedRoutedSiblingWritesCommitTargetHead)$' -count=1 -timeout=10m`

CI after push is green across PR, PR Risk, Cross-Version Smoke Tests, and Regression Tests.

_kilocode-unknown-model-unknown-reasoning on behalf of matt wilkie_

PR #3052, opened 2026-04-06
fix: auto-commit config table writes in remember, forget, and config commands

ryanRfox:
## Summary

Fixes #3028

`bd remember`, `bd forget`, `bd config set`, `bd config unset`, and `bd config set-many` write to the Dolt config table via `store.SetConfig()` / `store.DeleteConfig()`, but none of them call `store.CommitPending()` afterward. This means the config changes are staged in the Dolt working set but never committed to Dolt history — they silently vanish on server restart or get clobbered by the next unrelated commit.

## Changes

- **cmd/bd/memory.go**: Add `CommitPending()` after `SetConfig` in `remember` and after `DeleteConfig` in `forget`
- **cmd/bd/config.go**: Add `CommitPending()` after `SetConfig` in `config set`, after `DeleteConfig` in `config unset`, and after the batch loop in `config set-many`

All commit failures are non-fatal warnings since the primary operation (set/delete) already succeeded.

## Test plan

- [x] `make build` passes
- [x] Relevant tests pass (`TestRemember`, `TestForget`, `TestConfig`)
- [x] `go vet ./...` passes
- [x] Pre-existing test failures in `cmd/bd` (timeout), `cmd/bd/doctor`, and `internal/configfile` are unrelated to these changes (caused by live Dolt server interference)

PR #4383, opened 2026-06-12
fix: commit config-table writes in server mode + truthful explicit commits (GH#4078)

### LINKED ISSUE #4078: v1.0.4 regression: bd remember does not auto-commit on server mode (regression of #3028 from 7e0d09cbf)
## Summary

In **v1.0.4**, `bd remember` (and `bd forget` / `bd config set|unset`) silently fail to auto-commit when running against **server mode** (the `internal/storage/dolt` `DoltStore` backend used for shared / multi-tenant deployments — e.g. anything sharing a `dolt sql-server` across projects, including Gas City). Memory writes land in the working set but never get persisted as Dolt commits, so subsequent `bd dolt pull` rejects with `cannot merge with uncommitted changes` until the user manually commits via raw SQL.

`bd dolt commit` and `bd vc commit` also silently no-op on config-only working sets in server mode — they report success ("Committed." / "Created commit <hash>") but no commit is created and `dolt_status` stays dirty. The returned hash is the existing HEAD.

This is a **regression of #3028** (which was fixed by PR #3052 in v1.0.1) introduced by commit [`7e0d09cbf`](https://github.com/gastownhall/beads/commit/7e0d09cbf) ("fix CommitPending", May 1 2026) — six days before the v1.0.4 tag.

**Embedded-dolt mode is unaffected.** The asymmetry is in the storage-implementation `Commit()` method, not in the bd binary.

## Affected

- **Version:** v1.0.4 (`ce242a8...`). The two relevant commits — PR #3052's fix and `7e0d09cbf`'s refactor — both pre-date the v1.0.4 tag.
- **Storage mode:** **server mode** (`mode: server` per `docs/DOLT.md:346`) — `internal/storage/dolt.DoltStore`, used whenever a `dolt sql-server` is reached over a TCP socket. This includes both project-owned external servers and shared-server deployments at `~/.beads/shared-server/` (`BEADS_DOLT_SHARED_SERVER=1` / `dolt.shared-server: true`). Both are first-class supported configurations — not deprecated.
- **Embedded mode is NOT affected** (`internal/storage/embeddeddolt.EmbeddedDoltStore` — see "Root cause" for why).
- **Commands:** `bd remember`, `bd forget`, `bd config set`, `bd config unset` for the auto-commit path; `bd dolt commit` and `bd vc commit` for the explicit-commit path.

## Why this still matters even though server-mode auto-commit defaults to off

Server mode defaults `dolt.auto-commit` to off for a deliberate concurrency-safety reason (`cmd/bd/main.go:979-991`):

> Server mode: OFF — the server handles commits via its own transaction lifecycle; firing DOLT_COMMIT after every write under concurrent load causes 'database is read only' errors.

Upstream's intent is that server-mode users either (a) explicitly set `dolt.auto-commit: on` to opt in to the trade-off, or (b) leave it off / on `batch` and flush at logical boundaries via `bd dolt commit`. **Both intended escape hatches are broken for config-table writes:**

1. **`dolt.auto-commit: on` opt-in path** — `bd config get dolt.auto-commit` returns `on`, `--dolt-auto-commit=on` flag is read and validated, `maybeAutoCommit` proceeds past the `mode != doltAutoCommitOn` guard — but `st.Commit()` then filters `config` out at `internal/storage/dolt/store.go:1674` and returns nil early. No commit fires on `bd remember`.

2. **Default batch-commit path** — `bd dolt commit` / `bd vc commit` invoke the same `DoltStore.Commit()` and also silently no-op when only `config` is dirty. Both commands report success ("Committed." / "Created commit <hash>") but the returned hash is the *existing* HEAD and `dolt_status` stays dirty. This actively misleads users into thinking they've reconciled the working set, then their next `bd dolt pull` fails with `cannot merge with uncommitted changes`.

PR #3052's original fix used explicit `CommitPending` calls in the command handlers that worked correctly in *both* modes regardless of the auto-commit setting (CommitPending sees all dirty tables and stages them individually). The `7e0d09cbf` refactor lost that property by routing memory/config writes through the generic `Commit()` path.

## Reproducer

```bash
# Against a shared dolt sql-server (e.g. dolt.auto-start=false, server already running):
bd dolt show              # → "Mode: shared server"
bd config get dolt.auto-commit   # → "on"

# Write a memory:
bd remember --key test-2026-05-21 "Repro marker"
# → "Remembered [test-2026-05-21]: Repro marker"

# Status:
mysql -h 127.0.0.1 -P <port> -u root <db> -e "SELECT * FROM dolt_status;"
# → config | 0 | modified              ← uncommitted

mysql -h 127.0.0.1 -P <port> -u root <db> -e "SELECT message, date FROM dolt_log ORDER BY date DESC LIMIT 1;"
# → no commit corresponding to the remember above

# Workaround:
mysql ... -e "CALL DOLT_ADD('config'); CALL DOLT_COMMIT('-m', 'memory: ...');"
# → commit lands, dolt_status clean
```

`bd --dolt-auto-commit=on remember ...` also reproduces (flag is read but not honored on this path).

## Root cause

1. **PR #3052** (commit `a58149d36`, v1.0.1) fixed #3028 by adding explicit `store.CommitPending(ctx, getActor())` calls after `SetConfig`/`DeleteConfig` in `cmd/bd/memory.go` and `cmd/bd/config.go`. `CommitPending` sees all dirty tables and commits unconditionally.

2. **Commit `7e0d09cbf`** ("fix CommitPending", May 1) removed those explicit `CommitPending` calls — see the relevant hunks in `cmd/bd/memory.go` (lines around 91 and 231) and `cmd/bd/config.go`. They were replaced with `commandDidWrite.Store(true)`, deferring commit to the unified `PersistentPostRun` → `maybeAutoCommit` hook (`cmd/bd/dolt_autocommit.go`).

3. **`maybeAutoCommit`** (in `cmd/bd/dolt_autocommit.go:42-76`) calls `st.Commit(ctx, msg)` at line 69. It does *not* call `CommitWithConfig` or `CommitPending`.

4. **The two `Commit()` implementations diverge on whether they include the config table:**

   **Embedded** — `internal/storage/embeddeddolt/version_control.go:42-52`:
   ```go
   func (s *EmbeddedDoltStore) Commit(ctx context.Context, message string) error {
       return s.withConn(ctx, true, func(tx *sql.Tx) error {
           if _, err := tx.ExecContext(ctx, "CALL DOLT_ADD('-A')"); err != nil { ... }
           if _, err := tx.ExecContext(ctx, "CALL DOLT_COMMIT('-m', ?)", message); err != nil { ... }
           return nil
       })
   }
   ```
   `DOLT_ADD('-A')` stages everything including `config` → memory auto-commit works.

   **Server mode** — `internal/storage/dolt/store.go:1645-1705`:
   ```go
   // Lines 1659-1681:
   // GH#2455: Stage all dirty tables EXCEPT config. ...
   for rows.Next() {
       var table string
       if err := rows.Scan(&table); err != nil { ... }
       if table != "config" {                  // ← line 1674: filter
           tables = append(tables, table)
       }
   }
   ...
   if len(tables) == 0 {
       return nil // Nothing to commit (all changes were config-only or dolt_ignore'd)
   }
   ```
   When `config` is the only dirty table, `tables` is empty and `Commit()` returns `nil` early at line 1684. No commit is created, no error is surfaced.

The config-table exclusion in `DoltStore.Commit` is intentional per GH#2455 (to prevent sweeping up stale `issue_prefix` changes from concurrent ops). The bug is that `maybeAutoCommit` should use `CommitWithConfig` — or call `CommitPending` — when the write being committed *was* a config-table write.

## Why both bugs are connected

This is also what causes the **`bd vc commit` / `bd dolt commit` UX bug** mentioned in #3216 (residual issue, never separately tracked):

- `bd vc commit` and `bd dolt commit` both ultimately invoke `DoltStore.Commit()`.
- With only `config` dirty, both report success ("Committed." / "Created commit <hash>") but no new commit lands and `dolt_status` stays dirty. The "Created commit" hash is actually the existing HEAD.
- This actively misleads users into thinking they've reconciled the working set, then their next `bd dolt pull` fails.

Both batch-mode (`bd dolt commit`) and auto-commit-mode users on the shared-server backend hit this.

## Suggested fix

Three options, in order of risk:

- **(a) Cheap & low-risk — restore explicit calls.** Revert the `cmd/bd/memory.go` and `cmd/bd/config.go` hunks of `7e0d09cbf` (the `bd remember` / `bd forget` / `bd config set/unset` paths), going back to the v1.0.1 pattern of calling `store.CommitPending(ctx, getActor())` explicitly after `SetConfig`/`DeleteConfig`. This matches what shipped in v1.0.1 through v1.0.3 and was already validated to fix #3028.

- **(b) Cleaner — plumb intent through `maybeAutoCommit`.** Add a "this command touched config" flag (similar to `commandDidWrite` and `commandDidWriteTipMetadata`) that `maybeAutoCommit` checks to choose between `Commit()` and `CommitWithConfig()`. Keeps the unified PostRun hook architecture from `7e0d09cbf` intact.

- **(c) Most invasive — symmetric Commit.** Give `DoltStore.Commit` an opt-in "include config" parameter (or a new `CommitAll` method) and let callers express intent. Probably overkill given (b) achieves the same outcome.

Also: the `bd vc commit` and `bd dolt commit` user-facing paths should surface a clear message when the only dirty table is `config`, pointing the user at `bd dolt commit --include-config` or equivalent. Today's silent success is the part that bites.

## Related

- #3028 (CLOSED) — original "bd remember does not auto-commit" bug. **This issue is a regression.**
- PR #3052 (`a58149d36`) — original fix for #3028.
- #3216 (CLOSED) — `bd init` config-table commit fix (added `CommitWithConfig`); body explicitly flags the residual `bd vc commit` UX bug, but that piece was never tracked.
- GH#2455 — design rationale for excluding config from `DoltStore.Commit`.
- `7e0d09cbf` — refactor that introduced the regression.

## Workaround for users on v1.0.4 + server mode

Until fixed, after any session that writes memories or config:

```bash
mysql -h 127.0.0.1 -P <port> -u root <db> \
  -e "CALL DOLT_ADD('config'); CALL DOLT_COMMIT('-m', 'memory/config writes');"
```

Then `bd dolt pull` / `bd dolt push` will work normally.

aaronlippold:
> **Updated 2026-06-13:** rebased onto current `main` and narrowed. Two things this PR originally bundled are now resolved upstream by #4389, so they've been dropped and this PR is purely the #4078 config-commit fix:
> - the **routed-sibling read-only fix** — #4389's `routed.go` change fixes the same `store is read-only` bug;
> - the **CLI doc regeneration** — #4389 regenerated the docs, so the doc-flags jobs are green on `main` (companion PR #4385 closed as superseded).
>
> CI now runs against today's green `main`.

## What

Makes server-mode config-table writes actually commit, and makes the explicit
commit commands (`bd dolt commit` / `bd vc commit`) config-inclusive and
truthful.

## Why

Fixes #4078. In server mode, `bd remember` / `bd forget` /
`bd config set|unset|set-many` strand the working set dirty forever:
`maybeAutoCommit` skips SQL-server modes entirely, and generic `Commit()`
excludes `config` by design (#2455) — so the explicit flush commands silently
no-op on a config-only working set while still printing success. Every
subsequent `bd dolt pull` fails with `cannot merge with uncommitted changes`
until the operator runs raw `DOLT_ADD('config')`/`DOLT_COMMIT` SQL (we hit
this daily on a shared-server deployment).

Approach is per the claim comment on #4078 (option (b), modernized to current
`main`):

1. **`CommitConfigOnly`** (new `VersionControl` method): stages ONLY the
   config table via the existing `doltAddAndCommit` helper. Unlike
   `CommitWithConfig` (`-Am`), it cannot sweep a concurrent operation's dirty
   tables — the same guarantee that motivated #2455. Embedded aliases
   `Commit` for the same reason its `CommitWithConfig` does.
2. **Config writes commit immediately in server mode** (`commitConfigWrite`
   helper wired into the five write sites; `set-many` commits once per
   batch). Embedded mode is untouched — its `-Am` auto-commit and the
   unflagged-writes sweep already cover config. This restores the v1.0.1
   behavior from #3052 that the unified-PostRun refactor regressed.
3. **Truthful explicit commits**: `bd dolt commit` / `bd vc commit` share
   `explicitDoltCommit` — config-inclusive (`CommitPending` /
   `HasPendingChanges`+`CommitWithConfig`) and honest about
   `Nothing to commit` vs `Committed.` (previously `bd vc commit` printed
   `Created commit <hash>` where the hash was the unchanged HEAD).
   `--json` keeps its exact `{committed, hash, message}` shape.

`Commit()`'s #2455 config exclusion is intentionally unchanged — its ~60
callers keep their semantics.

## Verification

- New tests: 3 storage tests against a real `dolt-sql-server` container
  (scoped commit lands; concurrent dirty `issues` row survives — #2455
  pinned; generic `Commit()` exclusion pinned), 6 unit tests for
  `commitConfigWrite` (server/proxied/embedded modes, error surfacing,
  nothing-to-commit), 6 for `explicitDoltCommit` (truthfulness matrix).
- Full embedded group (memory, config, dep/link + direct-commit head
  invariants, VC, dolt) passes against current `main`.
- Live end-to-end on a disposable `dolthub/dolt-sql-server:2.1.0` container
  with a `file://` remote: `bd remember` → `dolt_status` clean → `bd dolt
  pull` → `Pull complete.` with zero manual SQL; truthful-output matrix for
  both explicit commit commands verified against `dolt_status`/`dolt_log`
  after each step.
- `go build` (default tags + `gms_pure_go` + `CGO_ENABLED=0`), `go vet`, and
  `./scripts/test.sh` clean.

Possible follow-ups (not in this PR, one issue per PR):

- `migrate.go` and `rename_prefix.go` have `SetConfig` sites with the same
  stranded-config shape.
- The v49→v51 `events` rekey migration panics with
  `invalid hash length: 19` on a database carrying 0.62-era event rows
  (clean databases migrate fine) — can file separately with the repro.
- macOS dev nit: `TestExpandPath/tilde_only` fails under `scripts/test.sh`
  because `test-env.sh` builds `HOME` from `$TMPDIR`, which carries a
  trailing slash on macOS. Passes bare; Linux CI unaffected.

---

## 📍 PR series map

Coordinated series from the same contributor fixing the config-commit bug
family and (next) proposing a centralized mode-policy architecture. Each item
is independently reviewable.

| # | PR / item | What | Status |
|---|-----------|------|--------|
| — | #4389 (merged) | `count --include-infra`; also regenerated the CLI docs and fixed routed-sibling writes + the bulk `--no-cycle-check` test | **merged** — resolved the doc-staleness and routed-sibling pieces this series originally carried |
| — | #4385 | standalone CLI doc regeneration | **closed** — superseded by #4389 |
| 1 | **this PR (#4383)** | GH#4078 config-commit core — server-mode config commits + truthful explicit commits | ready; CI green against current `main` |
| 2 | *(planned)* | RFC + Phase 1–5: centralize deployment-mode policy so the bug family becomes structurally impossible (each phase small, independently shippable, stacked on this PR) | after this PR, in phase order |

*"Stacked on" is a CI/merge-order relationship only — the diffs don't overlap.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Server-mode config writes commit again — `bd dolt pull` no longer needs a
+  manual flush.** `bd remember`, `bd forget`, and `bd config set|unset|set-many`
+  in server mode left the `config` table dirty forever: `maybeAutoCommit` skips
+  SQL-server modes entirely, and generic `Commit()` excludes config
+  ([#2455](https://github.com/gastownhall/beads/issues/2455)), so the next
+  `bd dolt pull` failed with `cannot merge with uncommitted changes` until the
+  operator ran raw `DOLT_ADD('config')`/`DOLT_COMMIT` SQL. Config writes now
+  commit immediately via the new scoped `CommitConfigOnly` (stages ONLY the
+  config table — concurrent operations' dirty tables are never swept),
+  restoring the v1.0.1 behavior from
+  [#3052](https://github.com/gastownhall/beads/pull/3052)
+  ([#4078](https://github.com/gastownhall/beads/issues/4078)).
+- **`bd dolt commit` and `bd vc commit` are truthful and include config.** On a
+  config-only working set both commands printed success (`Committed.` /
+  `Created commit <hash>` — the hash being the unchanged HEAD) while committing
+  nothing. Explicit commits are intentional flush points: they now include the
+  config table and honestly report `Nothing to commit` when the working set is
+  clean ([#4078](https://github.com/gastownhall/beads/issues/4078)).
 - **Cross-clone issue-delete vs child-row-insert merges now converge.** The
   synced child tables (`dependencies`, `labels`, `comments`, `events`,
   snapshots) carry `FOREIGN KEY ... ON DELETE CASCADE` to `issues` (migrations
--- a/cmd/bd/config.go
+++ b/cmd/bd/config.go
@@ -208,6 +208,10 @@ var configSetCmd = &cobra.Command{
 			os.Exit(1)
 		}
 		commandDidWrite.Store(true)
+		if err := commitConfigWrite(ctx, store, "config set"); err != nil {
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			os.Exit(1)
+		}
 
 		if jsonOutput {
 			outputJSON(map[string]string{
@@ -479,6 +483,10 @@ var configUnsetCmd = &cobra.Command{
 			os.Exit(1)
 		}
 		commandDidWrite.Store(true)
+		if err := commitConfigWrite(ctx, store, "config unset"); err != nil {
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			os.Exit(1)
+		}
 
 		if jsonOutput {
 			outputJSON(map[string]string{
@@ -743,6 +751,11 @@ Examples:
 				}
 			}
 			commandDidWrite.Store(true)
+			// One scoped commit for the whole batch, not one per key.
+			if err := commitConfigWrite(ctx, store, "config set-many"); err != nil {
+				fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+				os.Exit(1)
+			}
 		}
 
 		// Phase 7: Output results
--- a/cmd/bd/config_commit_write_test.go
+++ b/cmd/bd/config_commit_write_test.go
@@ -1,0 +1,114 @@
+//go:build cgo
+
+package main
+
+import (
+	"context"
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/storage"
+)
+
+type fakeConfigCommitStore struct {
+	storage.DoltStorage
+	configOnlyCalls int
+	messages        []string
+	err             error
+}
+
+func (f *fakeConfigCommitStore) CommitConfigOnly(_ context.Context, message string) error {
+	f.configOnlyCalls++
+	f.messages = append(f.messages, message)
+	return f.err
+}
+
+// GH#4078: server-mode config writes (bd remember, bd config set) must be
+// committed immediately and scoped — maybeAutoCommit skips server mode
+// entirely and generic Commit() excludes config, so nothing else ever
+// commits them.
+func TestCommitConfigWriteServerModeCommitsScoped(t *testing.T) {
+	saveStorageMode(t)
+	serverMode = true
+
+	fake := &fakeConfigCommitStore{}
+	if err := commitConfigWrite(context.Background(), fake, "remember"); err != nil {
+		t.Fatalf("commitConfigWrite: %v", err)
+	}
+	if fake.configOnlyCalls != 1 {
+		t.Fatalf("CommitConfigOnly calls = %d, want 1 in server mode", fake.configOnlyCalls)
+	}
+	if !strings.HasPrefix(fake.messages[0], "bd: remember (auto-commit) by ") {
+		t.Fatalf("commit message = %q, want prefix %q", fake.messages[0], "bd: remember (auto-commit) by ")
+	}
+}
+
+func TestCommitConfigWriteProxiedServerModeCommitsScoped(t *testing.T) {
+	saveStorageMode(t)
+	proxiedServerMode = true
+
+	fake := &fakeConfigCommitStore{}
+	if err := commitConfigWrite(context.Background(), fake, "config set"); err != nil {
+		t.Fatalf("commitConfigWrite: %v", err)
+	}
+	if fake.configOnlyCalls != 1 {
+		t.Fatalf("CommitConfigOnly calls = %d, want 1 in proxied server mode", fake.configOnlyCalls)
+	}
+}
+
+// Embedded mode already commits config via the '-Am' auto-commit and the
+// unflagged-writes sweep — the helper must not double-commit there.
+func TestCommitConfigWriteEmbeddedModeIsNoOp(t *testing.T) {
+	saveStorageMode(t)
+	serverMode = false
+	proxiedServerMode = false
+
+	fake := &fakeConfigCommitStore{}
+	if err := commitConfigWrite(context.Background(), fake, "remember"); err != nil {
+		t.Fatalf("commitConfigWrite: %v", err)
+	}
+	if fake.configOnlyCalls != 0 {
+		t.Fatalf("CommitConfigOnly calls = %d, want 0 in embedded mode", fake.configOnlyCalls)
+	}
+}
+
+// A real commit failure must surface — silently losing the commit is the
+// exact bug class this helper exists to fix (GH#4078's silent no-op half).
+func TestCommitConfigWriteSurfacesCommitError(t *testing.T) {
+	saveStorageMode(t)
+	serverMode = true
+
+	fake := &fakeConfigCommitStore{err: errors.New("connection refused")}
+	err := commitConfigWrite(context.Background(), fake, "remember")
+	if err == nil {
+		t.Fatal("commitConfigWrite returned nil, want the commit error surfaced")
+	}
+	if !strings.Contains(err.Error(), "connection refused") {
+		t.Fatalf("error = %q, want it to wrap %q", err.Error(), "connection refused")
+	}
+}
+
+// "Nothing to commit" is a benign race (e.g. the same key re-written to an
+// identical value) and must not fail the user's write.
+func TestCommitConfigWriteSwallowsNothingToCommit(t *testing.T) {
+	saveStorageMode(t)
+	serverMode = true
+
+	fake := &fakeConfigCommitStore{err: errors.New("nothing to commit")}
+	if err := commitConfigWrite(context.Background(), fake, "remember"); err != nil {
+		t.Fatalf("commitConfigWrite: %v, want nothing-to-commit swallowed", err)
+	}
+	if fake.configOnlyCalls != 1 {
+		t.Fatalf("CommitConfigOnly calls = %d, want 1", fake.configOnlyCalls)
+	}
+}
+
+func TestCommitConfigWriteNilStoreIsNoOp(t *testing.T) {
+	saveStorageMode(t)
+	serverMode = true
+
+	if err := commitConfigWrite(context.Background(), nil, "remember"); err != nil {
+		t.Fatalf("commitConfigWrite with nil store: %v, want no-op", err)
+	}
+}
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -473,16 +473,16 @@ For more options (--stdin, custom messages), see: bd vc commit`,
 			os.Exit(1)
 		}
 		msg, _ := cmd.Flags().GetString("message")
-		if msg == "" {
-			msg = fmt.Sprintf("bd: dolt commit (auto-commit) by %s", getActor())
-		}
-		if err := st.Commit(ctx, msg); err != nil {
-			if isDoltNothingToCommit(err) {
-				fmt.Println("Nothing to commit.")
-				return
-			}
+		// GH#4078: explicit commits include config and report truthfully.
+		// Without -m, CommitPending builds a descriptive summary message.
+		committed, err := explicitDoltCommit(ctx, st, msg)
+		if err != nil {
 			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 			os.Exit(1)
+		}
+		if !committed {
+			fmt.Println("Nothing to commit.")
+			return
 		}
 		commandDidExplicitDoltCommit = true
 		fmt.Println("Committed.")
--- a/cmd/bd/dolt_autocommit.go
+++ b/cmd/bd/dolt_autocommit.go
@@ -83,6 +83,31 @@ func commitPendingIfEmbedded(ctx context.Context, st storage.DoltStorage, actor 
 	return maybeAutoCommitStore(ctx, st, p)
 }
 
+// commitConfigWrite commits the config table immediately after an
+// intentional config write (bd remember/forget, bd config set/unset).
+// Server modes only: maybeAutoCommit skips SQL-server modes entirely, and
+// generic Commit() excludes config (GH#2455) — so without this the write
+// strands the working set dirty and the next 'bd dolt pull' fails until a
+// manual flush (GH#4078). The commit is SCOPED to the config table, so a
+// concurrent operation's dirty tables are never swept. Embedded mode is a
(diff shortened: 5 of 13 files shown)


aaronlippold:
CI note: the three failing jobs here (Build Artifacts, PR Policy, Check doc flags freshness) reproduce identically on `main` HEAD `1825cf357` (run 27426301286) — `docs/CLI_REFERENCE.md` wasn't regenerated when 49a803050/8cd650308 changed the `dep --no-cycle-check` / `import --allow-stale` help text. Not introduced by this PR (it touches no CLI surface). Fix: #4385.

maphew (commented):
Thanks for this one - it is a real, carefully diagnosed fix for #4078, and I confirmed the bug is still live on today's `main`: `maybeAutoCommit` skips server modes, `Commit()` still excludes config (`configExclude`), and `bd dolt commit` / `bd vc commit` still silently no-op on a config-only working set. Notably, the merged #4412 only fixed the pre-pull wedge for `kv.*` rows - its refusal message for dirty internal config keys tells operators to run `bd dolt commit`, which today cannot commit them. So this PR is the missing half of #4412, not superseded by it. The scoped `CommitConfigOnly` design (explicit `DOLT_ADD`, never `-Am`) and the test coverage against a real `dolt-sql-server` are exactly right.

Findings, all rebase-related:

- **[major]** `explicitDoltCommit`'s pending-changes check uses an optional interface assertion for `HasPendingChanges` - but #4418 (a59e75325) removed `DoltStore.HasPendingChanges` and the embedded one from `main`. After a mechanical rebase the check silently vanishes, and since server-mode `CommitWithConfig` swallows nothing-to-commit, `bd dolt commit -m x` on a clean set would report `Committed.` again - the exact bug you're fixing, and your mock-based tests won't catch it. Reinstate a concrete check (`issueops.HasPendingChanges` still exists) or make it a required method.
- **[major]** The custom-message path relies on `CommitWithConfig`'s `DOLT_COMMIT('-Am')`, which #4412 explicitly documents as unreliable for staging config under the server stored-procedure path (that's why `commitBeforePull` stages via `DOLT_ADD`). Post-rebase, route explicit commits through the new `commitWorkingSet` with a config-including mode instead. `CommitConfigOnly` itself is fine - `doltAddAndCommit` still exists on `main`.
- **[minor]** `CommitConfigOnly` stages the whole config table, so in server mode it can sweep a concurrent writer's half-applied internal config row (the row-level cousin of #2455). #4412's `kv.*` screen (`assertDirtyConfigUserKVOnly`) is a ready-made guard for the `remember`/`forget` sites.
- **[minor]** `vc.go` still sets `commandDidExplicitDoltCommit` before the outcome is known, while your `dolt.go` correctly sets it after - Copilot's earlier comment still applies.
- **[nit]** Soften the interface doc: embedded `CommitConfigOnly` commits the full working set, not "ONLY config".

On the conflict: this is an ordinary rebase, not the pre-rewrite situation - your branch shares merge-base e8ae7a2 (#4389) with `main`. `git rebase upstream/main` hits four hunks in `cmd/bd/vc.go` and the two `version_control.go` files, all from #4412/#4418 touching the same region.

**Recommendation: fix-merge.** Rebase onto current `main`, reconcile the two major items above with #4412's `commitWorkingSet`/`configCommitMode` structure, and re-run your server-container tests. Happy to review the rebased version promptly - this closes a gap `main`'s own error guidance currently points into.

maphew:
Coming back to this after too long a gap — apologies for the three weeks of silence since the July 5 review, and thanks for staying engaged back in June (including tracking the CI red on main down to #4385, which was above and beyond).

Re-verified today against `main` @ c989b6b87: **the bug you diagnosed is still live and this PR is still the fix of record for #4078.** `bd vc commit` (cmd/bd/vc.go:229) and `bd dolt commit` (cmd/bd/dolt.go:615) still route through `store.Commit` → `commitWorkingSet(..., configExclude)` (internal/storage/dolt/store.go:2317), so a config-only working set still reports "Nothing to commit". #4412 only covers `kv.*` rows on the pre-pull path (`configIncludeUserKVOnly`, store.go:2343) — and its refusal message for dirty internal config keys tells operators to run the very `bd dolt commit` that can't commit them. Nothing else open or merged touches this; your scoped `CommitConfigOnly` design remains exactly right.

Disposition, plainly: **we want this in, and the remaining work is ours if you'd like.** Since the July review, project policy changed — rebasing stale-but-wanted PRs is now maintainer work, not something we bounce back to contributors (#5094). The reconciliation needed is the same as before:

1. Rebase onto current `main` (conflicts are the four hunks in cmd/bd/vc.go and the two version_control.go files, from #4412/#4418).
2. Route the explicit-commit custom-message path through `commitWorkingSet` with a config-including mode instead of `CommitWithConfig`'s `DOLT_COMMIT('-Am')`, which #4412 documents as unreliable for staging config under the server stored-procedure path.
3. Replace the optional `HasPendingChanges` interface assertion with a concrete check — #4418 removed `DoltStore.HasPendingChanges`, so after a mechanical rebase the assertion silently never fires and `bd dolt commit` on a clean set would report "Committed." again; `issueops.HasPendingChanges` (internal/storage/issueops/commit_pending.go:18) is still there to build on.
4. Minor: screen the `remember`/`forget` `CommitConfigOnly` sites with #4412's `kv.*` guard, and fix the `commandDidExplicitDoltCommit` ordering in vc.go to match your dolt.go version.

**Want to take the rebase yourself, or shall I do it?** If I don't hear back within a week or so I'll do the rebase and reconciliation on a maintainer branch with your commits and authorship preserved, credit you in the merge, and link it here — this closes a gap that `main`'s own error guidance currently points users into, so it shouldn't wait on either of our calendars any longer than it already has.

maxinflection:
bumping this. would love to see it in the next release. Let me know if you want help rebasing.

maphew:
> Want to take the rebase yourself, or shall I do it? If I don't hear back within a week or so I'll do ...

I've changed roles from maintainer to normal contributor, handing off to @steveyegge, @julianknutsen, @csells to bring this home.

seanmartinsmith:
@aaronlippold - is this still on your plate? it's been sitting conflicting since the 08-24 handoff and nobody's picked it up since.

for context on why i'm asking: #4078 is still live on main @ `d530cddfa` on the sql-server route (`maybeAutoCommit`, `cmd/bd/dolt_autocommit.go:112-115`, returns nil whenever the store is not embedded, so `bd remember` writes never get a dolt commit), maphew re-verified this PR as the fix of record on 07-27, and hexmode added an independent confirmation on #4078 on 08-19. so it's not one deployment, and the fix is already written and reviewed.

if you don't have the cycles, i'm happy to rebase it onto current main as a fresh PR - your commits kept as-is with the conflict resolution on top, so authorship stays yours - and carry it through review. just say either way; i don't want to step on it if you're mid-rebase.

aaronlippold:
Hi Sean,

Thanks for reaching out. I'm happy to do whatever is best for the project.
If it's easier for you to rebase it and carry it forward, please feel free
to do so! Otherwise, if you'd prefer me to pick it back up, I'm happy to
finish it out. Let me know what is best for the group.

Best,
Aaron Lippold
***@***.***
260-255-4779

--------
Aaron Lippold

***@***.***

260-255-4779

twitter/aim/yahoo,etc.
'aaronlippold'

PR #4384, opened 2026-06-12
Preflight graph apply blocking cycles

julianknutsen:
## Summary
- preflight the planned graph for blocking dependency cycles before inserting edges
- skip per-edge SQL cycle checks for cycle-relevant graph apply dependencies after the graph-level preflight passes
- cover mixed local/external fanout edges with focused tests

## Verification
- go test ./cmd/bd -run 'TestExecuteGraphApplyUnit(SkipsSQLCycleChecksAfterGraphPreflight|RejectsMixedLocalExternalBlockingCycle|RejectsBlockingChildToParentDuplicate|RejectsReverseParentToChildBlockingEdge)' -count=1
- go test ./cmd/bd -run 'TestExecuteGraphApplyRejectsMixedLocalExternalBlockingCycle|TestExecuteGraphApplyRejectsStoredPrefixParentBlockingPath|TestExecuteGraphApplyAllowsExplicitParentChildDuplicate' -count=1

### DIFF
--- a/cmd/bd/assign.go
+++ b/cmd/bd/assign.go
@@ -27,9 +27,7 @@ Examples:
 
 		ctx := rootCtx
 
-		// Write-intent routing: a prefix-routed target must open writable so the
-		// assignee update commits on the target head (#4141).
-		result, err := resolveAndGetIssueWithRoutingForWrite(ctx, store, id)
+		result, err := resolveAndGetIssueForMutation(ctx, store, id)
 		if err != nil {
 			if result != nil {
 				result.Close()
--- a/cmd/bd/close.go
+++ b/cmd/bd/close.go
@@ -621,7 +621,7 @@ func resolveCloseTargets(ctx context.Context, localStore storage.DoltStorage, id
 		// Write-intent: a prefix-routed target opens writable so the close
 		// commits on the target head (#4141). Contributor auto-routing below
 		// stays read-only: it hydrates foreign projects that must not be mutated.
-		if r, err := resolveViaPrefixRoutingMode(ctx, id, true); err == nil {
+		if r, err := resolveViaPrefixRoutingWithAccess(ctx, id, true); err == nil {
 			results = append(results, r)
 			continue
 		}
--- a/cmd/bd/comment.go
+++ b/cmd/bd/comment.go
@@ -62,7 +62,7 @@ Examples:
 
 		ctx := rootCtx
 
-		result, err := resolveAndGetIssueWithRoutingForWrite(ctx, store, id)
+		result, err := resolveAndGetIssueForMutation(ctx, store, id)
 		if err != nil {
 			if result != nil {
 				result.Close()
--- a/cmd/bd/comments.go
+++ b/cmd/bd/comments.go
@@ -156,9 +156,7 @@ Examples:
 		}
 		ctx := rootCtx
 
-		// Write-intent: comments add writes through the routed target store,
-		// matching the `bd comment` shorthand (#4141).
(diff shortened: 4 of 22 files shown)


maphew:

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds a whole-graph preflight that rejects blocking dependency cycles before `graph apply` inserts any edges, and skips the now-redundant per-edge SQL cycle checks once that preflight passes — which closes a real data-integrity gap where a multi-edge `graph apply` could otherwise commit a blocking cycle that the per-edge checks missed.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The first review pass found that routed dependency/link target resolution had started opening prefix-routed foreign stores for writing when only the target ID was needed; the final patch restores read-only resolution for dependency/link targets, keeps write-intent resolution only for the mutated source issue/store, and adds routed `dep` coverage pinning that invariant.
- The embedded and domain (server-side) graph-apply paths enforced different blocking-cycle strategies; the final patch makes the domain use-case run its own whole-graph preflight, so both paths validate all planned blocking edges before any insert and the prior caller-assumption gap for fully-local server plans is closed.
- The domain cycle walk consulted only one dependency table; the final patch loads both the regular and wisp dependency tables so cross-table regular/wisp blocking cycles are detected, matching the SQL probe coverage it replaced.
- The bulk `--no-cycle-check` path had lost test coverage; the final patch restores tests for the positive path, cross-batch cycle-closing rejection, atomic rollback, and the "no edges added" contract.

Review findings addressed:
- Behavioral correctness (validated, all reviewers): the whole-graph blocking-cycle preflight is correct — embedded and domain graph-apply both validate planned blocking edges before inserts and only skip the per-edge SQL check afterward.
- Cross-table cycle detection (validated, high confidence): the domain graph walk now reads both dependency tables, so mixed regular/wisp blocking cycles are rejected before writes.
- Routed read/write boundary (validated, high confidence): source mutations use writable resolution while dependency/link targets stay on the read-only lookup path, reclosing the routed-write guardrail.

Remaining non-gating notes:
- Routing-helper access-mode docs were trimmed during the rename, and the local-cycle helper `graphApplyEdgeCanSkipSQLCycleCheck` now carries a slightly stale name; these are readability/guardrail-doc follow-ups, not correctness issues.
- The domain preflight performs two existing-dependency reads per uncached node — a latency-only consideration if graph-apply preflight cost ever becomes material.
- Parent-path validation now spans both dependency tables without a dedicated regression test; behavior is believed correct and is exercised indirectly by the cross-table cycle tests.
- No remaining gating issues were found.

### Maintainer Changes
`fix: maintainer review fixes for adopted graph-apply cycle PR` — 22 files changed, 537 insertions(+), 184 deletions(-). The maintainer fixes restore the read-only routed dependency/link resolution, add the domain whole-graph preflight with cross-table (regular + wisp) cycle detection, and add the domain and `dep` regression tests. The original contributor commit is preserved.

### Final Review Status
Ready for the merge queue: final head `18ea1abc1` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/27859777694 (all required checks green on `18ea1abc1`)

### Review Iterations
3 review passes performed (2 fix iterations): the first two blocked on the routed-resolution and cross-path/cross-table cycle-detection gaps above, and the third confirmed every required fix landed with no remaining blocker or major findings.

---

PR #4388, opened 2026-06-12
ci: run all tests, drop -short, add embedded job, pull Dolt image

coffeegoddd:
## Summary
- Drops `-short` and `-skip '^TestEmbedded'` from the required PR Go test gate (`scripts/ci/pr-core.sh`) and from the post-merge Linux/macOS test matrix (`.github/workflows/main.yml`)
- Pre-pulls `dolthub/dolt-sql-server:2.1.0` in the pr-core and main test jobs so the testcontainers-backed integration tests actually start instead of silently returning `doltNoImage`
- Opts pr-core's test env into the real Dolt path (`BEADS_TEST_ENV_RUN_DOLT=1`) so `scripts/ci/lib/test-env.sh` stops auto-adding `BEADS_TEST_SKIP=dolt`
- Drops `-short` from the `CGO_ENABLED=0` cmd/bd compile guard (no behavior change — none of the explicitly-listed tests use `testing.Short()`)
- Adds `-race`, log capture (`TEST_LOGFILE`), and `pipefail` to `scripts/test.sh` for parity with how this gets exercised locally

Coverage gap this closes: previously the dolt-server-backed integration suites silently self-skipped in CI (via `BEADS_TEST_SKIP=dolt` and the missing Docker image). Those suites will now run.

Embedded-Dolt coverage is unchanged — `pr-risk.yml` and `main.yml` already build the embedded test binaries and run a 20-way sharded `test-embedded-cmd` matrix plus `test-embedded-storage`, gated on `detect-ci-tier.outputs.full_embedded == 'true'`. A first revision of this PR added a redundant single-job `pr-embedded-wrapper`; that has been reverted.

## Test plan
- [ ] `pr-core-wrapper` runs full `./...` with race + 20m timeout against a live Dolt testcontainer
- [ ] Post-merge `test` matrix (Linux + macOS) runs the full suite without `-short`
- [ ] CGO_ENABLED=0 guard still passes the named pure-Go subset without `-short`
- [ ] Existing `test-embedded-storage` and 20-shard `test-embedded-cmd` in `pr-risk.yml` continue to run unchanged when `detect-ci-tier` selects the full embedded tier


maphew (changes requested):
The telemetry configuration can be overridden by project config despite being presented as a user-level opt-out, and the new user-config unset path both fails for nested metrics keys and loosens file permissions. These are functional/privacy regressions that should be fixed before merging.

Findings:

- [P1] Resolve metrics settings from user scope only - `cmd/bd/main.go:1452`  
  When a repository `.beads/config.yaml` contains `metrics.disabled` or `metrics.endpoint`, Viper's project-over-user precedence makes these lookups override the user's `~/.config/bd/config.yaml`; a user who ran `bd config set metrics.disabled true` can therefore be opted back into telemetry, or have events sent to a repo-specified endpoint, just by running `bd` in that repo. Metrics privacy settings should ignore project config or ensure user/env settings win.

- [P2] Unset nested metrics keys instead of flat names - `internal/config/yaml_config.go:263`  
  For configs bootstrapped or updated in nested form (`metrics: { disabled: ... }`), `commentOutYamlKey` only searches for a literal top-level `metrics.disabled:` line, so `bd config unset metrics.disabled` or `metrics.endpoint` reports success while leaving the active nested value in place. Use the same nested-YAML handling as `SetUserYamlConfig` when unsetting dotted user keys.

- [P2] Preserve private permissions when rewriting user config - `internal/config/yaml_config.go:265`  
  When unsetting a metrics key, this rewrites the entire user config with mode `0644`, unlike the `0600` mode used by bootstrap/set. If the same `~/.config/bd/config.yaml` contains secrets such as `github.token` or `linear.api_key`, unsetting a metrics key makes them readable by other local users; preserve the existing mode or keep `0600`.

maphew:
Closing as superseded, with thanks — the value in this branch is already on `main` with your authorship preserved.

What landed where:

- The usage-metrics + RunE refactor bundled in this branch was cherry-picked to `main` as d7d7a3460 (author: coffeegoddd, "cherry picked from 9596c723") via #4419, then hardened by follow-ups: first-run consent notice (00d317dde), `DO_NOT_TRACK` opt-out (#4938), and the config precedence/unset/permissions fixes that resolved all three findings from the CHANGES_REQUESTED review here (metrics opt-out can no longer be overridden by a repo config; nested `metrics:` keys unset correctly; user config keeps 0600). The stale review state on this PR is therefore resolved-on-main, not actionable here.
- The CI intent (run the full suite, drop `-short`) was superseded by main's current architecture: sharded CI gates plus the `Short()`-allowlist enforced by `check-testing-short.sh` — main adopted the opposite design deliberately, so the residual `-short` removal in this branch now contradicts upstream policy.
- The remaining diff is unlandable mechanically: ~701 commits behind with ~123 conflicted paths (the RunE refactor era plus six weeks of drift), no CI on the current head, and ~20 tests blanket-skipped in intermediate commits.

If you still want something from this branch that main doesn't have (e.g. an embedded-Dolt CI job variant), a fresh small PR against current main would be very welcome. Thanks again — the metrics work this carried is now a shipped feature.

PR #4393, opened 2026-06-13
fix: Dolt /tmp/mysql.sock down → TCP fallback (gt-28itz)

athosmartins:
## Problem (gt-28itz)
Dolt's `/tmp/mysql.sock` is intermittently down. When it is, `gt mq submit` and cross-rig `bd` reads fail with `connection refused` on the socket path, even though the Dolt server is reachable via TCP (port still works). This blocked merge flows + cross-rig bead reads in Gas City (2026-06-13).

## Fix
`fix/gt-28itz-dolt-socket-tcp-fallback` @8d55d7fc5 — when the Unix socket connection fails, fall back to TCP automatically. Built + tested + pushed.

### DIFF
--- a/internal/storage/dolt/socket_fallback_test.go
+++ b/internal/storage/dolt/socket_fallback_test.go
@@ -1,0 +1,78 @@
+package dolt
+
+import (
+	"errors"
+	"net"
+	"testing"
+	"time"
+)
+
+// resolveSocketTransport implements a socket-first / TCP-fallback policy for
+// Dolt server connections (gt-28itz). These tests stub dialProbe so they run
+// without a live Dolt server.
+
+func withStubbedDialProbe(t *testing.T, reachable map[string]bool) {
+	t.Helper()
+	orig := dialProbe
+	t.Cleanup(func() { dialProbe = orig })
+	dialProbe = func(network, addr string, _ time.Duration) error {
+		if reachable[network+"|"+addr] {
+			return nil
+		}
+		return errors.New("stub: unreachable " + network + " " + addr)
+	}
+}
+
+// No socket configured → pure TCP, no probing, returns "".
+func TestResolveSocketTransport_NoSocket(t *testing.T) {
+	withStubbedDialProbe(t, map[string]bool{})
+	got := resolveSocketTransport("", "127.0.0.1", 52756, 100*time.Millisecond)
+	if got != "" {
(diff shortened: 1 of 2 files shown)


maphew:

kevglynn:

Ran a complexity-only review on this diff. The socket-fallback fix itself (`resolveSocketTransport` + test, ~130 lines) is clean and minimal — no unnecessary abstractions, no speculative config.

The main observation is scope rather than complexity: this PR bundles three unrelated concerns behind a single fix title:

1. **Socket→TCP fallback** (the titled fix) — lean, well-tested, ships as-is.
2. **Auto-convoy feature** (`auto_convoy.go`, ~329 lines) — unrelated to socket transport, separate concern.
3. **Post-mortem document** (`scripts/post-mortem-beads-fork-sync.md`, 102 lines) — operational writeup unrelated to the socket fix.

Per the project's "one PR, one coherent concern" convention, splitting these would make each independently reviewable and mergeable. The socket fix is ready now; the other two don't need to wait on it or vice versa.

One minor code nit: `auto_convoy.go` uses `crypto/rand` + base32 + ToLower + truncate for a non-security random ID. `math/rand/v2` with a direct alphabet pick would be simpler.

**net: -102 lines possible** (post-mortem deletion), but the bigger win is the split.

maphew (approved):
Maintainer branch surgery, explained: this branch bundled three concerns, and review feedback (kevglynn's pass) plus the project charter pointed at shipping just one of them. I've force-pushed a re-cut:

- **Ships as-authored, your commit, your authorship**: the socket-first/TCP-fallback fix (`dialProbe`/`resolveSocketTransport`), cherry-picked cleanly onto current main with one keep-both conflict resolution against the recently-landed test-mode guard.
- **Maintainer additions**: a `debug.Logf` on the fallback branch so operators can see socket-down/TCP-used in the wild, and test hardening from a cross-vendor review gate — the suite now guards the `newServerMode` wiring itself (probe-recording stubs; the test fails if the wiring is removed) and asserts the no-socket path never probes.
- **Dropped from this PR**: the auto-convoy orchestration piece — gastown orchestration concepts belong outside beads core per the project charter. Worth re-proposing against gastown or as a beads plugin discussion if you still want it; happy to help route that.

Checks pending on the new head; merge tail goes to the pr-babysit patrol once green. Thanks for the fallback fix, athosmartins — it's a real operational papercut solved.

PR #4403, opened 2026-06-14
fix(query): lex digit-led identifier-shaped tokens as identifiers

Rome-1:
## Summary

The query lexer's \`readNumberOrDuration\` unconditionally claimed any digit-led token, then erred at the first non-digit / non-duration-suffix byte. Identifier-shaped values like \`1-alpha\` or \`42d-sla\` therefore couldn't appear unquoted on the value side of a comparison:

\`\`\`
$ bd query 'assignee=1-alpha'
Error: parsing query: expected digit at position 11

$ bd query 'label=42d-sla'
Error: parsing query: expected digit at position 12
\`\`\`

Every caller that wanted to filter on such a value had to quote the string or fall back to a client-side filter. The DSL already treats identifiers and values uniformly (the existing \"identifier with hyphen\" / \"unquoted identifier with colon\" cases under \`TestLexer\` exercise that for letter-led tokens) — digit-led was the only outlier.

## Changes

Two lookahead adjustments in \`readNumberOrDuration\`:

- **Duration suffix is terminal-only.** \`isDurationSuffix\` only fires when the suffix character isn't followed by another identifier-continuation character. \`updated>7d\` still tokenises as \`updated\` / \`>\` / \`7d\` (duration); \`label=7day-sla\` falls through to the identifier path.
- **Unsigned digit-led identifier fallback.** If, after reading the digit run, the next character is an identifier-continuation char, the lexer rewinds to the token start and re-reads via \`readIdent\`. Signed forms (\`-3-foo\`) still error so arithmetic-shaped input doesn't silently morph into an identifier — quote them if you need them.

## Testing

- \`go test ./internal/query/\` — full package passes (existing cases unchanged).
- Five new \`TestLexer\` cases:
  - \`digit-led identifier on the value side\` (\`assignee=1-alpha\`)
  - \`digit-led identifier in a compound expression\` (\`assignee=2-bravo AND status=open\`)
  - \`digit-led identifier with duration-suffix prefix\` (\`label=7day-window\`)
  - \`duration still wins when the suffix stands alone\` (\`updated>30d\`)
  - \`duration with trailing space stays a duration\` (\`updated>30d AND status=open\`)
- \`go test ./internal/storage/issueops/\` passes.
- End-to-end: built \`cmd/bd\` against this branch, ran \`bd query 'assignee=1-alpha'\` against a real store — returns cleanly instead of the parser error. \`bd query 'priority=2'\` and \`bd query 'updated>30d'\` continue to work as before.

## Notes for reviewers

- No external API or schema change; this is strictly a lexer-fallback widening.
- The pre-commit hook needed \`GOTOOLCHAIN=go1.26.2\` and a bumped golangci-lint timeout to run on this machine; lint result was clean (\`0 issues\`). Happy to file a separate PR bumping the pinned \`golangci-lint@v2.10.1\` to a release that supports the project's Go target if maintainers want one.

### DIFF
--- a/internal/query/lexer.go
+++ b/internal/query/lexer.go
@@ -229,12 +229,19 @@ func (l *Lexer) readString(quote rune, startPos int) (Token, error) {
 }
 
 // readNumberOrDuration reads a number or duration (e.g., 7d, 24h).
+//
+// Unsigned digit-led tokens that continue into identifier characters
+// (e.g. "1-alpha", "42day-sla", "9.3.1") are re-lexed as identifiers so
+// they can stand in unquoted on the value side of comparisons like
+// "label=1-alpha". Signed forms ("-3-foo") still error — the user has
+// to quote them.
 func (l *Lexer) readNumberOrDuration(startPos int) (Token, error) {
 	var sb strings.Builder
 
 	// Handle optional sign
 	r := l.next()
-	if r == '-' || r == '+' {
+	hadSign := r == '-' || r == '+'
+	if hadSign {
 		sb.WriteRune(r)
 		r = l.next()
 	}
@@ -256,13 +263,26 @@ func (l *Lexer) readNumberOrDuration(startPos int) (Token, error) {
 		sb.WriteRune(r)
 	}
 
-	// Check for duration suffix
-	if r != 0 && isDurationSuffix(r) {
+	// Check for duration suffix. Only commit to a duration when the suffix
+	// stands alone — if more identifier characters follow (e.g. "7day"),
+	// fall through to the identifier-fallback below.
+	if r != 0 && isDurationSuffix(r) && !isIdentChar(l.peek()) {
 		sb.WriteRune(r)
 		return Token{Type: TokenDuration, Value: sb.String(), Pos: startPos}, nil
 	}
 
-	// Not a duration suffix, back up
+	// Identifier-continuation fallback: an unsigned digit-led run that
+	// butts against more identifier characters is an identifier, not a
+	// number. Restart the lex at the original position and read it as an
+	// identifier so "1-alpha", "2bravo", "42d-sla" all tokenize as one
+	// TokenIdent.
+	if !hadSign && r != 0 && isIdentChar(r) {
+		l.pos = startPos
+		return l.readIdent(startPos)
+	}
+
+	// Not a duration suffix and not an identifier — back up the lookahead
+	// rune.
 	if r != 0 {
 		l.backup()
 	}
--- a/internal/query/query_test.go
+++ b/internal/query/query_test.go
@@ -115,6 +115,36 @@ func TestLexer(t *testing.T) {
 			expected: []TokenType{TokenIdent, TokenEquals, TokenIdent, TokenEOF},
 			values:   []string{"label", "=", "gt:merge-request", ""},
 		},
+		{
+			name:     "digit-led identifier on the value side",
+			input:    "assignee=1-alpha",
+			expected: []TokenType{TokenIdent, TokenEquals, TokenIdent, TokenEOF},
+			values:   []string{"assignee", "=", "1-alpha", ""},
+		},
+		{
+			name:     "digit-led identifier in a compound expression",
+			input:    "assignee=2-bravo AND status=open",
+			expected: []TokenType{TokenIdent, TokenEquals, TokenIdent, TokenAnd, TokenIdent, TokenEquals, TokenIdent, TokenEOF},
+			values:   []string{"assignee", "=", "2-bravo", "AND", "status", "=", "open", ""},
+		},
+		{
+			name:     "digit-led identifier with duration-suffix prefix",
+			input:    "label=7day-window",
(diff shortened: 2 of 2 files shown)


kevglynn (commented):
Reviewed and tested locally — all lexer tests pass (22 TestLexer cases, 2 TestLexerErrors cases).

The approach is sound:
- The lookahead adjustment (`isDurationSuffix(r) && !isIdentChar(l.peek())`) correctly preserves `30d` as a duration while letting `7day-window` fall through to the identifier path
- The rewind-and-relex via `l.pos = startPos; return l.readIdent(startPos)` cleanly reuses the existing identifier logic rather than duplicating it
- Signed forms (`-3-foo`) still error, which prevents arithmetic-shaped input from silently becoming identifiers — good design boundary

The five new test cases cover the important edge cases well: basic digit-led identifiers, compound expressions, duration-prefix collision (`7day-window`), and regression tests confirming durations still win when standalone.

One observation: the change is slightly more complex than a typical "easy win" since it touches parser internals with subtle state-machine semantics, but the scope is tight (one function, one file, clear test coverage). Fix-merge candidate at most. LGTM.

maphew:

Rome-1:
@maphew Thanks — here's a self-contained scratch-repo repro.

**Environment**
- `bd version`: `1.0.5 (dev)`
- Build points: base = `7f30ac3c1` (this PR's parent, pre-fix), fixed = `9b51223f6` (PR head)
- Storage mode: default — **dolt backend, embedded mode** (the bug is in the query lexer, so it's storage-independent; reproduces identically under sqlite/jsonl)
- OS: Linux 6.17.0-1017-aws (Ubuntu 24.04, x86_64)

**Setup (fresh scratch repo)**
```bash
git clone https://github.com/gastownhall/beads && cd beads
git checkout 7f30ac3c1 && go build -o /tmp/bd-base ./cmd/bd     # pre-fix
git checkout 9b51223f6 && go build -o /tmp/bd-fix  ./cmd/bd     # this PR

mkdir /tmp/scratch && cd /tmp/scratch && /tmp/bd-base init
```

**Reproduce the bug (pre-fix binary)** — any digit-led, identifier-shaped value on the RHS of a comparison fails to parse:
```
$ /tmp/bd-base query "assignee=1-alpha"
Error: parsing query: expected digit at position 11

$ /tmp/bd-base query "label=42d-sla"
Error: parsing query: expected digit at position 10

$ /tmp/bd-base query "label=7day-window"
Error: parsing query: unexpected token "ay-window" at position 8 (expected end of query)
```
- **Expected:** the value lexes as an identifier and the query runs (zero matches in an empty repo).
- **Actual:** parse error — the lexer commits to a number/duration on seeing the leading digit, then chokes on the trailing identifier characters.

**Verify the fix (PR binary)** — same commands, same repo:
```
$ /tmp/bd-fix query "assignee=1-alpha"
No issues found matching query: assignee=1-alpha
$ /tmp/bd-fix query "label=42d-sla"
No issues found matching query: label=42d-sla
$ /tmp/bd-fix query "label=7day-window"
No issues found matching query: label=7day-window
```

**End-to-end (proves it matches, not just parses):**
```
$ /tmp/bd-fix create "demo issue" -l 1-alpha          # -> bd-scratch-1k3
$ /tmp/bd-fix query "label=1-alpha"
Found 1 issues:
  ○ bd-scratch-1k3 [● P2] [task] [1-alpha] - demo issue

$ /tmp/bd-base query "label=1-alpha"                  # same query, pre-fix
Error: parsing query: expected digit at position 8
```

**Regression check** — bare durations still lex as durations (no behavior change):
```
$ /tmp/bd-fix query "updated>30d"
No issues found matching query: updated>30d
```

Signed digit-led forms (e.g. `-3-foo`) still error by design and must be quoted, as noted in the lexer comment.

maphew (commented):
Thanks for the fix and for the thorough repro follow-up, @Rome-1 - the base/fix binary comparison with a scratch repo was exactly what I asked for and made this easy to verify independently.

I applied the diff to current `main` in a scratch worktree and confirmed:
- `go test ./internal/query/...` passes, including your five new `TestLexer` cases.
- The lookahead logic holds up under a couple of adjacent cases I probed: `updated>30d` still lexes as a duration, `label=7day-window` and `assignee=1-alpha` lex as identifiers, and signed forms remain rejected as designed.
- Two incidental (non-regressing) side effects worth knowing about, not blockers: digit-led decimals like `priority=1.5` now lex as an identifier and fail later at `strconv.Atoi` with a clearer error instead of erroring in the lexer on the stray `.`; and unquoted date-shaped values like `created>2026-01-01` now lex as one identifier and actually parse via `ParseRelativeTime`, where before they split into three tokens and never worked unquoted at all. Both are improvements or neutral, not breaks.

`kevglynn`'s review already covers the design tradeoffs well (rewind-and-relex reusing `readIdent`, terminal-only duration suffix check, signed forms still erroring) and I agree with the LGTM.

Recommendation: merge as-is. No changes requested.

PR #4406, opened 2026-06-14
fix(cli): render markdown with terminal capability gates (#3881)

### LINKED ISSUE #3881: bd show: Markdown formatting is broken 🐞
- The bd CLI used to support markdown formatting in the console, but it seems tob be currently broken.

### Steps to reproduce

1. create a bd bead
2. create a markdown document for the bead body (e.g. `body.md`)
3. set the bead's body with the markdown file `bd update <beadid> --body-file 'body.md'`
4. execute `bd show <beadid>`

### Current behavior

The CLI output is in plaintext.
Headers like `DESCRIPTION` and `COMMENT` are formatted, but the bead content itself is not formatted.

### Expected behavior

- when the terminal supports ANSI codes, the output should be properly formatted as markdown
- headers should be formatted properly using ANSI colors
- inline code should be properly formatted using language-aware formatting
- markdown tables should be properly formatted
- italics and boldface should be properly formatted
- links should be clickable when the terminal supports it
- images should be properly drawn using [sixel](https://en.wikipedia.org/wiki/Sixel) format

MovGP0:
Fixes the existing Glamour formatting for ANSI-Capable terminals.

## What

- checks the terminal if ANSI / OSC8 is supported
- if supported, renders the output using the ANSI / OSC8 formatting using Glamour

## Why

Fixes #3881 

## Verification

- Unit tests where implemented

### DIFF
--- a/internal/ui/terminal.go
+++ b/internal/ui/terminal.go
@@ -3,6 +3,8 @@ package ui
 
 import (
 	"os"
+	"strconv"
+	"strings"
 
 	"golang.org/x/term"
 )
@@ -25,6 +27,7 @@ func IsStderrTerminal() bool {
 //   - NO_COLOR: https://no-color.org/ - disables color if set
 //   - CLICOLOR=0: disables color
 //   - CLICOLOR_FORCE: forces color even in non-TTY
+//   - TERM=dumb: disables color unless explicitly forced
 //   - Falls back to TTY detection
 func ShouldUseColor() bool {
 	// Git hook context - disable color to prevent termenv OSC 11 terminal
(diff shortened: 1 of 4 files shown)


maphew:

MovGP0:
> Could you add the smallest command sequence or setup that triggers this, including the `bd version`, storage mode, OS, and the expected vs. actual output? A scratch-repo repro is ideal if one is possible.

It's easy to reproduce on Windows:
```sh
bd show <beadid>
```
The problem is the broken detection for ANSI support detection on Windows.

MovGP0:
@maphew any other questions?

maphew:
how does the broken ansi support manifest? Is this with powershell (what version) or cmd, and what terminal are you using?

For me there aren't any issues with `bd show ...` display, but I've been using this win11 machine for serveral years and I have starship, nerdfonts, and some other things installed that might be hiding it. 

I'm using _bd version 1.0.5_ built from _(6a3f515ce: HEAD@6a3f515ced18)_, which is newer than the published release.  However there have been no changes in terminal handling that I remember.

maphew:
I've taken another look. "ANSI detection" framing is a misdiagnosis, but the PR does contain the real fix almost incidentally.

Every platform, every terminal, every build since 2026-03-10 has been getting the breakage, quietly. It's not Windows-specific, and nothing to do with ANSI/OSC8 capability detection. Anyone whose terminal shows colored bd section labels is also getting an uncolored body — they just may not realize the body was supposed to be styled. Fix incoming.

maphew:
Thanks @MovGP0 — merging this. Your report and fix are what surfaced the actual root cause, which turned out to be broader than terminal detection.

**Summary of what's going in**

- **Root cause:** the glamour v1->v2 dependency migration dropped the style option from the `NewTermRenderer` call. In glamour v2 that constructor does *not* auto-select a style, so bodies rendered with an empty `StyleConfig` (plaintext) on every platform since the bump — not a Windows ANSI-detection issue. Section labels (DESCRIPTION, COMMENT) stayed colored because they go through `ui.RenderBold` rather than glamour, which is exactly the "headers formatted, body not" symptom.
- **Fix (yours):** restores rendering via `WithEnvironmentConfig()` and adds terminal-capability gating — ANSI honoring `NO_COLOR` / `CLICOLOR` / `TERM=dumb`, OSC 8 hyperlinks gated to a known-capable terminal allowlist with a `FORCE_HYPERLINK` override, and a glamour notty fallback for no-color terminals.
- **Added on merge:** a dedicated `internal/uimd/markdown_test.go` regression guard (`TestRenderMarkdownStylesBodyContentRegression3881`) that pins the #3881 symptom directly — it asserts a heading/bold/inline-code body renders with ANSI styling and consumes the markdown markers. Verified it fails against the empty-`StyleConfig` bug and passes with the fix, so future refactors of the OSC 8 logic can't silently re-break body styling.

Squash-merging once checks are green.

PR #4411, opened 2026-06-15
test(deps): pin insert-vs-backfill id agreement for #4259

rjc123:
## What this pins

`AddDependency` (the production insert path) and `rekeyDependencyIDs` (the backfill in migration 0050) must compute the same dependency primary key from the same edge. The existing tests proved each path individually but not that they **agree with each other**:

- `TestAddDependencyUsesDeterministicID` — insert path calls `depid.New()` ✓
- `TestRekeyDependencyTableRewritesOnlyDivergentRows` — backfill calls `depid.New()` ✓
- **No test that the two calls are over the same input** ✗

A refactor that changes what `AddDependency` writes to `depends_on_issue_id` — or changes the `COALESCE` column order in `rekeyDependencyTable` — would silently diverge. Two independently-upgraded clones would then produce different ids for the same logical edge, re-introducing the #4259 fork with no warning until `bd dolt pull` fails in production.

## What the test does

`TestInsertBackfillIDPathsAgree` in `internal/storage/dolt/dependencies_merge_test.go`:

1. Creates a real edge via `AddDependency` (the production insert path)
2. Reads back the stored `id`
3. Reads the exact column values `rekeyDependencyTable` would `COALESCE` — `(issue_id, COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external))`
4. Recomputes `depid.New(issueID, coalesceTarget)` — what the backfill would derive
5. Asserts stored id == backfill-derived id

This is the test called out as missing in the #4259 Opus review: *"The code is consistent (verified above), but this is the exact invariant a future refactor could break silently, so pinning it is cheap insurance."*

## Test plan

- [x] Runs under existing `BEADS_TEST_EMBEDDED_DOLT=1` gate (uses `setupTestStore`)
- [x] Fails immediately if `AddDependency` and the backfill's COALESCE disagree
- [x] No new dependencies or build tags required

### DIFF
--- a/internal/storage/dolt/dependencies_merge_test.go
+++ b/internal/storage/dolt/dependencies_merge_test.go
@@ -100,6 +100,76 @@ func TestDependencyDeterministicIDMergesAcrossBranches(t *testing.T) {
 	})
 }
 
+// TestInsertBackfillIDPathsAgree pins the invariant that the production insert
+// path (AddDependency) and the backfill path (rekeyDependencyIDs / migration
+// 0050) compute the same dependency primary key from the same edge.
+//
+// Insert path:  depid.New(dep.IssueID, dep.DependsOnID); dep.DependsOnID is
+//
+//	stored in depends_on_issue_id.
+//
+// Backfill path: SELECT issue_id, COALESCE(depends_on_issue_id,
+//
+//	depends_on_wisp_id, depends_on_external) FROM dependencies;
+//	then depid.New(issue_id, coalesced_target).
+//
+// The two paths agree because AddDependency writes DependsOnID to
+// depends_on_issue_id and the backfill reads that column first via COALESCE.
+// If AddDependency ever begins encoding the target differently, or the
+// backfill's COALESCE column order changes, the stored id diverges from what
+// the backfill would derive — and two independently-upgraded clones would fork
+// again (#4259). This test creates a real edge via AddDependency, reads back
+// the exact column values the backfill would use, recomputes depid.New over
+// them, and asserts they match.
+func TestInsertBackfillIDPathsAgree(t *testing.T) {
+	store, cleanup := setupTestStore(t)
+	defer cleanup()
+
+	ctx, cancel := testContext(t)
+	defer cancel()
+
+	db := store.db
+	for _, id := range []string{"ibp-src", "ibp-dst"} {
+		if _, err := db.ExecContext(ctx,
+			"INSERT INTO issues (id, title, description, design, acceptance_criteria, notes, status, priority, issue_type) VALUES (?, ?, '', '', '', '', 'open', 2, 'task')",
+			id, id); err != nil {
+			t.Fatalf("seed issue %s: %v", id, err)
+		}
+	}
+	if err := store.AddDependency(ctx, &types.Dependency{
+		IssueID: "ibp-src", DependsOnID: "ibp-dst", Type: types.DepBlocks,
+	}, "test"); err != nil {
+		t.Fatalf("AddDependency: %v", err)
+	}
+
+	// What the insert path stored.
+	var insertID string
+	if err := db.QueryRowContext(ctx,
+		"SELECT id FROM dependencies WHERE issue_id = 'ibp-src' AND depends_on_issue_id = 'ibp-dst'",
+	).Scan(&insertID); err != nil {
+		t.Fatalf("read insert id: %v", err)
+	}
+
+	// What the backfill would compute from the stored column values:
+	// rekeyDependencyTable selects (issue_id, COALESCE(depends_on_issue_id,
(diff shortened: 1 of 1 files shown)


rjc123:
Hey @maphew — just a nudge that the version bump to cut a release is important here. The (from OPUS review) fix for #4259 is all merged to main but without a tagged release users (like me) are still running in a broken state. Would be great to get that out the door soon. 

Adding this test to fill some holes presented by my agent review of the build work done while I was away from the office.

kevglynn (commented):
Reviewed. This is a well-motivated invariant pin — exactly the kind of test that prevents silent regressions across independent upgrades.

The test logic is sound:
1. Creates a real edge via `AddDependency` (production insert path)
2. Reads back the stored `id`
3. Reads the same columns the backfill's `COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external)` would read
4. Recomputes `depid.New(issueID, coalesceTarget)` — what migration 0050 would derive
5. Asserts they match

This directly addresses the gap identified in the #4259 review — the two paths were individually tested but never tested for agreement. A future refactor that changes the COALESCE order or what `AddDependency` writes to `depends_on_issue_id` would trip this immediately.

Test-only PR, no production code changes, runs under existing `BEADS_TEST_EMBEDDED_DOLT=1` gate. Easy win. LGTM.

maphew:
**Automated agent review** (Kilo / kilo-auto/frontier — not a human maintainer)

Good invariant pin overall, and the test structure is sound. One concrete issue worth addressing before merge:

**Duplicated COALESCE expression — test does not track the real backfill**

`TestInsertBackfillIDPathsAgree` (line 154–160) hardcodes the COALESCE SQL string:

```go
"SELECT issue_id, COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external) "+
    "FROM dependencies WHERE issue_id = 'ibp-src'",
```

`internal/storage/schema/dep_id_backfill.go` uses the identical literal independently (line 58). If someone later changes the COALESCE order in the backfill and forgets to update the test, both sides still compile and the test passes — but the invariant it was meant to pin is silently broken, putting us back in the #4259 situation.

The fix is to replace the hardcoded string with the existing shared constant:

```go
// internal/storage/sqlbuild/sqlbuild.go:32
const DepTargetExpr = "COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external)"
```

```go
import "github.com/steveyegge/beads/internal/storage/sqlbuild"

if err := db.QueryRowContext(ctx,
    "SELECT issue_id, "+sqlbuild.DepTargetExpr+
        " FROM dependencies WHERE issue_id = 'ibp-src'",
).Scan(&issueID, &coalesceTarget); err != nil {
```

CI is all green and the test logic is otherwise correct. Recommend fixing the expression duplication before merge so the pin is actually load-bearing.

rjc123:
Thanks for the review.

**Pushing back on this one** — the duplication is real, but the divergence risk it implies isn't load-bearing here. `ck_dep_one_target` enforces that exactly one of the three target columns is non-null per dependency row. With that invariant in place, the `COALESCE` order is cosmetic: any permutation returns the same value for any well-formed row.

The "someone reorders the COALESCE and the test silently passes" scenario only materialises if `ck_dep_one_target` is also broken — at which point we have a much bigger problem this test was never the right place to catch.

**What the test does pin is intact:**

- `AddDependency` writes `DependsOnID` to `depends_on_issue_id`
- `depid.New(IssueID, DependsOnID)` agrees with `depid.New(issueID, COALESCE_target)` on the natural edge key
- The test fails immediately if either side regresses

Swapping in `sqlbuild.DepTargetExpr` would be a cosmetic DRY tidy, not a strengthening of the invariant. It would also leave the underlying duplication untouched — `schema/dep_id_backfill.go:58` also hardcodes the same string, as do several other call sites. A single source of truth for `DepTargetExpr` is a separate, broader refactor worth doing in its own right; I don't think it should ride on this test PR. Happy to file a follow-up issue if you want it picked up.

---

**Meta point on AI reviewers**

Kilo's suggestion follows the right shape (extract constant, dedupe SQL) but isn't load-bearing here. As human maintainers I think we have to be deliberate that automated reviewers' suggestions get accepted on technical merit rather than on autopilot — the natural bias of these agents is to add code, and tidiness that doesn't strengthen an invariant accumulates without improving the codebase.

For symmetry: this reply was drafted with Claude Opus 4.7 under my direction — same class of tool as Kilo, just on the contributor side. The accept/reject call sits with me as the human.

Will leave the PR as-is for now — let me know if I've missed something.

maphew:
Thanks for the contribution @rjc123 , and I appreciate the pushback. I do tend to let the clankers do what they want, especially when I'm on the edge of my understanding like here.

PR #4412, opened 2026-06-15
fix(dolt): pre-pull commit must include config (kv.memory.* pull-wedge)

steveyegge:
## Problem

`bd dolt pull` always fails with `Error 1105 (HY000): cannot merge with uncommitted changes`, and `bd doctor` reports `Dolt Locks: config: modified` right after every command — even immediately after `bd dolt commit` and with `--dolt-auto-commit on`. Push is unaffected; only the pull (cross-clone receive) path is wedged. (Live repro: Wyvern wy-71t, server mode, git-protocol remote.)

## Root cause — two correct changes collide over the `config` table

1. **Memories live in `config`.** `bd remember` stores each persistent memory as a `kv.memory.<slug>` row in the **synced** `config` table.
2. **`Commit` deliberately excludes `config`** (GH#2455): `DOLT_COMMIT('-Am', …)` once swept stale `issue_prefix` changes from concurrent writers into commits, so `Commit` now stages every dirty table *except* `config`.
3. **The pre-pull auto-commit uses `Commit`** (GH#2474) in both pull paths — `pullFromRemote` (default remote) and `PullFrom` (named-peer) — so it skips `config`.

Net effect: once any memory exists, the `kv.memory.*` rows sit permanently uncommitted (nothing ever commits `config`), the pre-pull "clean the working set" step leaves `config` dirty, and `DOLT_MERGE` refuses to start. GH#2474's intent is silently defeated by GH#2455 whenever `config` is the dirty table — which is always, after the first `bd remember`.

## Fix

1. **`store.go` / `federation.go`:** a new `commitBeforePull` stages `config` **explicitly** via `DOLT_ADD` (not `-Am`) before the pre-pull commit. `Commit` and `commitBeforePull` share `commitWorkingSet(includeConfig bool)`.

   > Why not the literal `Commit`→`CommitWithConfig` swap from the proposal: on the live server-mode DB, `CommitWithConfig`'s `DOLT_COMMIT('-Am', …)` did **not** reliably stage `config` under the SQL stored-procedure path, whereas explicit `DOLT_ADD('config')` did. A new embedded test proves `-Am` *does* stage config in the in-process engine, so the `-Am` flaw is **server-mode-specific** — explicit staging is correct for both.

2. **`mergesettle.go`:** `TryAutoResolveMergeConflicts` now auto-resolves `config` conflicts with `--theirs` **only when every conflicted row is a `kv.memory.*` key** (convergent, same policy already used for `metadata`). Any other config key (e.g. `issue_prefix`) is left for the operator. New helper `configConflictsAreMemoryConvergent`, const `memoryConfigKeyPrefix = "kv.memory."`.

3. **Embedded mode unchanged** in behavior — its pre-pull already staged config; it gains the resolver hardening and the proof test.

## Tests

`pull_config_wedge_test.go` added in both:
- `internal/storage/dolt` — server-mode, needs Docker → runs in CI.
- `internal/storage/embeddeddolt` — in-process, green locally.

`gofmt` + `go vet` clean; full embedded settle suite shows no regressions.

## Why this doesn't reintroduce GH#2455

GH#2455 guards against `-Am` sweeping a concurrent writer's half-applied `issue_prefix` change into an unrelated commit. The pre-pull commit is different in kind: an explicit user action (`bd dolt pull`) that *requires* a clean working set, committing *this clone's own* config as the merge basis — exactly what `bd dolt commit` already does via `CommitWithConfig`. Staging only happens on the pull path; the convergent resolver touches only `kv.memory.*`.

### DIFF
--- a/PROPOSAL-pull-config-wedge.md
+++ b/PROPOSAL-pull-config-wedge.md
@@ -1,0 +1,173 @@
+# Proposal: fix `bd dolt pull` "cannot merge with uncommitted changes" wedge
+
+**Status:** for review (diagnosis + patch). Authored 2026-06-14 from a live repro in
+the Wyvern beads DB (server mode, git-protocol remote on Bitbucket).
+
+## Symptom
+
+`bd dolt pull` always fails:
+
+```
+Error 1105 (HY000): cannot merge with uncommitted changes
+```
+
+`bd doctor` reports `Dolt Locks: config: modified` immediately after every command,
+even right after `bd dolt commit`, and even with `--dolt-auto-commit on`. Tracked as
+Wyvern wy-71t. Push is unaffected; only pull (the cross-clone *receive* path) is wedged.
+
+## Root cause — two correct PRs collide over the `config` table
+
+1. **Memories live in `config`.** `bd remember` stores each persistent memory as a
+   `kv.memory.<slug>` row in the **synced** `config` table. Confirmed via the working
+   set on the live DB — the only dirty rows were exactly the 5 memories:
+
+   ```
+   SELECT to_key, diff_type FROM dolt_diff('main','WORKING','config');
+   kv.memory.beads-jsonl-sync-procedure                       added
+   kv.memory.bitbucket-altssh-dead                            added
+   kv.memory.local-dev-ssl-posture                            added
+   kv.memory.react-client-e2e-against-the-local-game-server   added
+   kv.memory.react-e2e-same-character-parallel-logins         added
+   ```
+
+2. **`Commit` deliberately excludes `config`** (GH#2455, `store.go` ~1731):
+   `DOLT_COMMIT('-Am', …)` once swept stale `issue_prefix` changes from concurrent
+   operations into commits, so `Commit` now stages every dirty table *except* `config`.
+
+3. **The pre-pull auto-commit uses `Commit`** (GH#2474). Both pull paths try to clean
+   the working set before merging, but call `s.Commit`, which skips `config`:
+
+   - `internal/storage/dolt/store.go:2309` — `pullFromRemote` (default remote; what
+     `bd dolt pull` invokes via `st.Pull`)
+   - `internal/storage/dolt/federation.go:65` — `PullFrom` (named-peer pull)
+
+**Net effect:** once any memory exists, the `kv.memory.*` rows sit permanently
+uncommitted (nothing ever commits `config`), so the pre-pull "clean the working set"
+step leaves `config` dirty, and `DOLT_MERGE` refuses to start — the exact case
+`abortMerge`'s comment already calls out: *"a merge that REFUSED TO START on a dirty
+working set."* GH#2474's intent is silently defeated by GH#2455 whenever `config` is
+the dirty table, which is always after the first `bd remember`.
+
+This is independent of the v42→v51 schema-migration fork (#4259) seen in the same DB;
+that was a separate, newer problem. Migrating + force-publishing fixed the fork and
+restored `push`, but `pull` stays wedged because of this config issue.
+
+## Empirical validation (live DB)
+
+Confirmed the mechanism end-to-end without any source change:
+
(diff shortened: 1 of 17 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @steveyegge! This PR fixes a Dolt pull/sync **config wedge**: before it, the pre-pull auto-commit excluded the `config` table, so a dirty `kv.memory.*` row (for example after `bd remember`) could leave `bd pull` and `bd federation sync` permanently unable to merge. The fix pre-commits pending user config before merge entry and auto-resolves convergent `kv.memory.*` conflicts, which keeps Dolt-backed federation usable when local memory/config writes are present — a real operability win for anyone syncing a database that also holds agent memory.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- `bd federation sync` (server `pullFromRemote`, peer `PullFrom`, and federation `Sync`) still hit the original dirty-`config` merge wedge; the final patch pre-commits pending user `kv.*` config before merge entry on all three paths.
- The pre-pull auto-commit could have committed unsafe internal `config` rows such as `issue_prefix`; the final patch screens the pre-pull commit to user `kv.*` keys only and refuses dirty non-`kv.*` config, preserving the GH#2455 stale-internal-config guard.
- `bd vc merge --strategy` / `sync --strategy` could leave a resolved config-only conflict uncommitted in server mode; the final patch routes those resolutions through the config-inclusive `CommitMergeResolution` path.
- Generic `bd kv set memory.*` data was being treated as a convergent memory; auto-resolution is now limited to `kv.memory.*` (with an operator notice), and the `kv.memory.` prefix is centralized in `internal/storage/kvkeys` so it can no longer drift between packages.
- Embedded `bd federation sync` could still hit the dirty-working-set refusal; the final patch adds a pre-merge commit guard plus an embedded regression test.

Review findings addressed:
- Behavioral correctness (major, high, claude+codex): the dirty-`config` sync wedge is now closed across server, peer, federation, and embedded sync paths, backed by new `pull_config_wedge_test.go` coverage in both `internal/storage/dolt` and `internal/storage/embeddeddolt`.
- Pre-pull commit safety (major, high, codex): unsafe internal `config` rows are no longer auto-committed; only user `kv.*` keys pass the pre-pull screen.
- Strategy-resolution commit (major, high, claude+gemini): config-only conflict resolutions from `--strategy` are committed through the config-inclusive path instead of being silently dropped in server mode.
- Memory namespace boundary (major, high, claude+gemini): convergent auto-resolution is scoped strictly to `kv.memory.*` and the prefix is centralized to prevent drift.

Remaining non-gating notes:
- Legacy `bd kv set memory.*` rows remain readable but can no longer be removed via `bd kv clear` (minor) — that namespace now belongs to `bd remember`/`bd forget`.
- Embedded federation `Sync --strategy` is correct today but still relies on incidental `Commit` behavior rather than the explicit `CommitMergeResolution` contract used elsewhere (minor).
- `bd doctor` can briefly report `config: modified` after a memory write; that state now self-clears on the next pull (nit).

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch as a single commit on top of your contributor commits:
- `fix(dolt): adopt-pr maintainer fixes for pull/sync config wedge` — 17 files changed, +840/-62 (config-inclusive commit paths across server/peer/federation/embedded sync, `internal/storage/kvkeys` prefix centralization, `CommitMergeResolution` routing for `--strategy`, and server + embedded `pull_config_wedge` regression tests).

### Final Review Status
Ready for the merge queue: final head `e052f7d0` has passing GitHub checks (51 checks green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/27859809225

### Review Iterations
5 review passes — 4 fix iterations resolving the major findings above, then a clean approving pass with only minor/nit follow-ups remaining.

---

PR #4414, opened 2026-06-16
feat(storage): backend-agnostic conformance test suite

MarkAtwood:
## Summary

- Adds `internal/storage/conformance/` package with 40 behavioral contract tests for the `Storage`/`DoltStorage` interface
- Any backend implementation can verify behavioral parity by calling `conformance.RunAll(t, factory)`
- No changes to existing code — pure addition

## Motivation

Storage backends (dolt, embedded, and potentially future backends) each write their own tests with no shared contract verification. A conformance suite ensures behavioral parity across implementations and catches semantic drift.

## Test coverage

| Category | Tests |
|---|---|
| Issue CRUD | 12 (create, get, update, delete, close, reopen, external ref, batch) |
| Search/filter | 6 (text query, status, priority, limit, count, group-by) |
| Dependencies | 3 (add, remove, counts) |
| Ready/blocked | 5 (no deps, blocked by open, unblocked by close, blocked list, epic closure) |
| Labels | 3 (add/remove, idempotent, query by label) |
| Comments | 2 (add + chronological order, count) |
| Config | 2 (set/get/getall, local metadata) |
| Metadata slots | 1 (set, get, clear) |
| Statistics | 1 (aggregate counts) |
| Iterators | 2 (issues, comments) |
| Transaction | 1 (create + label in single tx) |

## Usage

```go
func TestConformance(t *testing.T) {
    conformance.RunAll(t, func(t *testing.T) storage.DoltStorage {
        return newTestStore(t) // backend-specific factory
    })
}
```

## Test plan

- [x] `go build ./internal/storage/conformance/` compiles
- [x] `go vet ./internal/storage/conformance/` clean
- [ ] Wire into existing Dolt backend tests (follow-up)


maphew:
Thanks for this @MarkAtwood - the conformance suite is exactly the executable contract the storage layer has been missing. I pushed one commit (`4e6aaad`) to wire it into the real Dolt backends and to make it run green, since the "wire into existing Dolt backend tests" box was still unchecked and the suite had never actually executed against a backend.

What the commit does:

**Wiring**
- `internal/storage/embeddeddolt/conformance_test.go` runs `conformance.RunAll` against embedded Dolt. It's gated by `BEADS_TEST_EMBEDDED_DOLT=1` like the rest of that package, so it runs in the existing `test-embedded-storage` CI job and self-skips in the fast job. **38/38 green** locally.
- `internal/storage/dolt/conformance_test.go` runs it against server-mode Dolt via `setupTestStore`. It self-skips with no live test server, so enforcing it on every PR needs a dedicated CI job with a `dolt-sql-server` container - I'll track that as a follow-up rather than bolt CI infra onto this PR.

**Fixes the first real run surfaced (suite side, your design preserved):**
- Factory contract documented + embedded factory now sets `issue_prefix`. The storage layer validates it and does not default it (that's the CLI layer's job).
- `withDefaults()` fills the `Status`/`IssueType` the storage layer requires, so the terse literals stay valid.
- Not-found checks use `errors.Is(err, storage.ErrNotFound)` rather than strict `==`, since backends wrap the sentinel with context.
- `CreateDuplicate` now asserts the data-integrity invariant (exactly one row for the ID after a duplicate create) instead of expecting an error: beads' storage layer intentionally upserts on duplicate ID (#3955 import idempotency). The test deliberately does not bless last-writer-wins overwrite.

**One backend parity fix the suite caught (its whole point):**
- `issueops/delete.go`: `DeleteIssue` now wraps `storage.ErrNotFound` to match `GetIssue`/`UpdateIssue`. `CloseIssue` has the same latent gap; follow-up.

Happy to split the `delete.go` change out if you'd rather keep this PR test-only - let me know. All edits preserve your suite structure and coverage.

maphew:
Merging this now with the conformance suite wired into CI against **both** Dolt backends. Summary of what landed on top of @MarkAtwood's suite:

**CI enforcement**
- `test-embedded-conformance` — runs `conformance.RunAll` (`TestConformance`) against embedded Dolt, split out of `test-embedded-storage` (which now `-test.skip`s it) so neither job blows its timeout. Kept main's bump of `test-embedded-storage` to `-test.timeout=20m`.
- `test-server-storage` — the server-mode twin: builds a dedicated `dolt-conformance-test` binary, pulls `dolthub/dolt-sql-server:2.1.0`, and runs the same suite against server Dolt. Both are wired into `ci-gate` and gated on the existing `full_embedded` tier.

**Bug the suite immediately caught (and this PR fixes)**
- The server-conformance job hung `TestConformance/IterIssues` to the timeout on its first run. Root cause: `dolt.IterIssues` streamed rows on a dedicated connection and hydrated each row's labels on a **second** pool connection. That works at `MaxOpenConns=2` (embedded) but server-mode pins to `MaxOpenConns=1` (branch isolation is session-level via `DOLT_CHECKOUT`), so the label query deadlocked waiting for the cursor's connection.
- Fix: `IterIssues` now buffers rows and batch-hydrates labels (`WHERE issue_id IN (...)`) inside a single read transaction on one connection — the same pattern `SearchIssues` already uses — then returns a `SliceIter`. Same query/filter/ordering; labels follow the existing nil-when-empty convention.

This is exactly the value of a backend-agnostic conformance contract: it turned an invisible single-connection deadlock into a failing test. Thanks @MarkAtwood for the suite.

All 53 checks green (server + embedded conformance included) prior to the final `main` merge; conflict resolution was limited to `pr-risk.yml`.

Issue #4331, opened 2026-06-09
Concurrent-mutate ephemeral-jsonl-import race clobbers persisted field edits (Dolt server mode)


daver787:
**Labels:** bug, data-loss, dolt-server-mode, concurrency
**Severity:** High — silently reverts committed writes; corrupts orchestration state under concurrent access
**Affected version:** `bd 1.0.4` (Homebrew), Dolt `sql-server` mode (DB on `:3307`)

## Summary

In Dolt server mode, every `bd` **mutate** rebuilds an ephemeral DB by importing `.beads/issues.jsonl` and then blind-upserts. When a second process mutates concurrently (or the on-disk jsonl is stale due to the export throttle), that import **reverts field edits another process already committed** — including edits that were written, read back, and verified as persisted. Additive rows (new issues, new dependency bonds) survive; **field updates on existing rows are silently lost.**

This is not theoretical: it has clobbered live orchestration state repeatedly (multi-agent dispatch, where many short-lived workers each run `bd` at startup).

## Impact

- **Silent data loss on field edits.** `status`, custom fields (e.g. an attached-work pointer), and even a previously-committed `closed` state get reverted by an unrelated concurrent mutate.
- **Read-after-write is not durable.** A write can be confirmed persisted on reread, then reverted seconds later by another process's startup import.
- **Asymmetric:** read paths do **not** import, so reads look consistent while writes quietly regress. Additive rows persist; field mutations on existing rows do not.
- Forces operators into fragile, purely-operational mitigations (serialize all writes, pre-warm the server, avoid concurrent startups) that don't scale.

## Environment

- `bd version 1.0.4 (Homebrew)`
- Dolt running as a server: `dolt sql-server --config .../.dolt-data/config.yaml`, port `:3307`
- `.beads/issues.jsonl` present as the passive export, with an export throttle (`export.interval`, default `60s`)
- Workload: many concurrent short-lived `bd` clients (each runs `bd` commands at process startup)

## Root cause (as diagnosed)

The **mutate path differs from the read path**: reads connect to the server, but mutates first **reconstruct an ephemeral DB from `issues.jsonl` and blind-upsert into it**. Two failure modes follow:

1. **Stale-jsonl revert (single-writer).** The on-disk jsonl lags real state by up to the export-throttle window (`60s` default). A mutate that imports that stale jsonl upserts old field values over newer committed ones — reverting them.
[...]

Issue #4135, opened 2026-05-23
bd 1.0.4 embedded mode: 6 observed failure modes (split-brain, auto-flush, identity, chain drops) — consolidated report


olgasafonova:
Consolidated report of six failure modes observed across `bd` 1.0.3 → 1.0.4 in a multi-repo workflow (Mar–May 2026). Each is reproducible in isolation. I'm filing them together because they share two root drivers (the Dolt/JSONL dual-source identity model + the auto-flush write semantics during reads) and the upstream fixes are easier to design as a set than one-by-one.

Documented in our internal rules at `rules/beads-tooling.md` over the period the failures emerged. Six concrete fix asks (a)-(f) listed at the end.

---

### Environment

- bd versions affected: 1.0.3 and 1.0.4 (Homebrew install)
- Mode: embedded Dolt (no `bd dolt start` running)
- OS: macOS 14.x, Darwin 24.5.0
- Use case: solo workflow across ~25 repos, occasional multi-agent parallel work via `git worktree` (with the `claude-code` CLI as the runner)
- All repos use `.beads/issues.jsonl` as the git-tracked source of truth

---

### Failure mode 1 — Daemon + worktree split-brain (observed 23-04-2026)

Worktrees plus the `bd` daemon's `.beads/` auto-discovery can latch onto the wrong location after a branch switch. Symptoms:

- `bd list` from the main checkout returns "No issues found" while the database still has issues
- `embeddeddolt/` ends up with two sub-directories (e.g. `beads/` and `<repo_name>/`), suggesting database identity got split
- Branch-switch sequences (main → audit branch → feature branch) silently change which `.beads/` the daemon serves
- A `bd status` warning about "permissions issue" on the worktree path appears while running from the main repo

**Reproduction:** create a worktree of a repo that has `.beads/`; switch branches in either checkout; run `bd list` and observe count mismatch with the JSONL.

---

### Failure mode 2 — Split-brain triggers without worktrees (observed 04-05-2026)

Same `embeddeddolt/` two-subdirectory pattern reproduces in a single checkout (no worktrees). When the database is empty or stale, four sharper failure modes surface:

**2a. Auto-flush wipes JSONL on every read against an empty DB.**

[...]

PR #4415, opened 2026-06-16
feat(storage): flat-file storage backend

MarkAtwood:
## Why

Beads is powerful but its Dolt dependency is a barrier to adoption. Small teams and solo developers who just want git-tracked issue management alongside their code have to run a database server, manage 50MB+ of embedded storage, and deal with a complex sync model.

The flat-file backend makes beads work like a `.editorconfig` or `.github/` — plain files in git, no infrastructure. `git push` syncs your issues. `git clone` gets them. `git log -p .beads/issues/PROJ-abc.json` shows the full history of one issue. No server, no ports, no migration system, no 65KB text limit. For teams already using git, this is zero-overhead issue tracking.

## What

Flat-file storage backend for beads. One JSON file per issue, git-tracked, no server process. Drop-in replacement that eliminates 8 known Dolt/JSONL bugs.

**Depends on:** #4414 (conformance test suite)

**New package** `internal/storage/flatfile/` (~5,700 lines):
- Full `DoltStorage` interface implementation (70+ methods, 14 sub-interfaces)
- Passes all 40 conformance tests from #4414
- 75+ additional unit/integration/concurrency tests
- Atomic write-tmp-rename on all persistent state
- Path traversal protection, git flag injection prevention
- Hash-based ID generation matching Dolt's algorithm

**CLI integration** (~125 lines of existing-file changes):
- `bd init --backend=flatfile` — new init path
- `bd migrate flatfile` — migration from Dolt with `--dry-run`
- `bd doctor` — 7 flatfile-specific health checks
- Store factory dispatch in `newDoltStore`, `newDoltStoreFromConfig`, `newReadOnlyStoreFromConfig`

**Zero impact on existing Dolt users** — flatfile only activates when `metadata.json` says `backend: "flatfile"`.

## How it works

```
.beads/
  metadata.json        # {"backend":"flatfile"}
  config_kv.json       # config key-value store
  issues/
    proj-abc.json      # one file per issue (git-tracked)
    proj-def.json
  comments/
    proj-abc/
      2026-06-15T10-30-00.json
  events/
    proj-abc.jsonl     # append-only audit log
```

Each issue is a standalone JSON file. Dependencies are stored as an array in the depending issue — reverse lookups computed at read time by scanning all files. At <10K issues this is instant (~2MB in memory). Comments and events are separate files to minimize churn on the issue file itself.

Concurrency: different issues = different files = git auto-merges with no conflict. Same issue edited by two people = standard git merge conflict (visible and explicit, unlike Dolt's silent last-writer-wins). This is strictly better than every Dolt concurrency bug filed to date.

## Bugs eliminated

| Bug | Root cause | Fixed? |
|---|---|---|
| #4128 write-path OOM + lock fights | Bypassing running Dolt server | Yes — no server |
| #3931 silent issue drops | JSONL overwrite from stale session | Yes — no JSONL |
| #4331 concurrent mutation race | Single-file JSONL clobber | Yes — per-issue files |
| #3406 watcher race on delete | JSONL file-watch race | Yes — no watcher |
| #3948 auto-import overwrite race | JSONL auto-import | Yes — no import step |
| #4135 embedded mode failures | 6 failure modes in embedded Dolt | Yes — no Dolt |
| #4259 migration PK fork | Dolt migration system | Yes — no migrations |
| #3925 TEXT column truncation | Dolt 65KB cap | Yes — no cap |

## Files changed (existing code)

| File | Lines | What |
|---|---|---|
| `store_factory.go` | +13 | Flatfile dispatch (cgo) |
| `store_factory_nocgo.go` | +11 | Flatfile dispatch (nocgo) |
| `init.go` | +70 | `runInitFlatfile` function |
| `doctor.go` | +25 | Flatfile branch + embedded-mode bypass |
| `configfile.go` | +21 | `BackendFlatfile` constant, `GetBackend()` reads field |
| `.golangci.yml` | +9 | gosec exclusions for flatfile paths |

Everything else is new files in `internal/storage/flatfile/` and `cmd/bd/doctor/flatfile.go`.

## Test plan

- [x] `go test ./internal/storage/flatfile/` — 75+ tests pass
- [x] `golangci-lint run` — 0 issues
- [x] Conformance suite — 40/40 pass
- [x] End-to-end CLI: init → create → list → dep add → ready → close → doctor
- [x] Three independent code review passes (49 findings, all resolved)
- [ ] Dog-food on real project


maphew:
Thanks for the PR. It's driven some interesting exploration and welcome thinking. 1st  pass comments:

_The flat-file model would buy real adoption (no CGO, files-in-git, no Dolt edges) and would cost beads its automatic multi-agent merge and its decoupled multi-repo story. Accepting it into core would contradict the storage-as-driver direction we've been holding everyone else to. So the value is worth absorbing, but the right vessel is a conformance-gated driver behind the boundary,_

The axis here isn't server vs. no-server; it's git-native versioning vs. database-native versioning. The flat-file backend isn't a novel idea competing with Dolt — it's a smarter redo of the model Dolt replaced: one JSON file per issue instead of one giant JSONL, which fixes the single-file-clobber class because different issues become different files that git auto-merges. So the real question isn't "Dolt vs. new thing," it's: now that we know how to do git-file storage without the JSONL clobber bug, does git-native versioning recover enough of what Dolt gave us to justify shedding Dolt's costs?

**Flat-file user gains:**

- Zero infrastructure / no CGO - pure Go. That means trivial cross-compilation, smaller static binaries
- Transparency / tool-ability. Issues are plain JSON you can cat, grep, jq, and review in a PR diff. git log -p .beads/issues/PROJ-abc.json is a more accessible history than dolt blame. For a lot of solo/OSS users this "issues are just files" model is the entire appeal (cf. git-bug, Fossil).
- Drops Dolt's hard edges (the bugs and issues mentioned)
- Conflicts become visible. Same-issue concurrent edits surface as a normal git merge conflict 

**Flat-file user loses:**

- Cell-level merge -- and this is the load-bearing loss. Beads markets "cell-level merge" and "Zero Conflict … multi-agent/multi-branch workflows" as core (README, ARCHITECTURE.md, DOLT.md). With Dolt, two agents updating different fields of the same issue merge cleanly. With flat-file, that's a whole-file conflict a human has to resolve. The beads+dolt flagship use case is many agents hammering issues in parallel, so here it's a regression, not an improvement
- SQL queries, indexes, and scale. Reverse-dependency lookups and search are "scan every file into memory." Fine at <10K issues (~2MB, instant — and most target users live well under that), but there's no index story, so it's O(n) where Dolt has things like the (status, updated_at) composite index. The scale ceiling is real but probably above where the intended audience sits.
- Dolt-native branching, time-travel (AS OF), and cross-entity ACID transactions.

The above is to reframe, and not to place value judgement. For someone choosing between the models, some things to consider.

**My concerns as a project maintainer:**

1. Charter tension: "Beads should not become a storage engine" and concurrency/locking/crash-safety are the driver's job, not beads.  A flat-file backend belongs behind the driver boundary as its own driver/plugin, not as a second engine welded into core. There's messiness here because Dolt is not cleanly behind a driver boundary, however this the direction we're headed with it and the intention behind much of the last month's activity.

2. Permanent 2× maintenance, and the PR doesn't even gate it. This is a 70-method, 14-sub-interface contract. Every future storage capability now needs two implementations forever, or the backends silently drift. And as submitted, the PR touches only .golangci.yml — no CI workflow change wires the conformance suite into the gate. So parity is asserted ("40/40 pass") but not enforced on future PRs.

3. Support-surface forking. The moment two backends ship, every bug report, doc page, and feature ("does branching work? does AS OF work? why is search slow?") gains a "which backend?" axis. Things built Dolt-first either won't exist or behave differently on flat-file. That cost is paid continuously, by maintainers in triage.

4. Migration fidelity. bd migrate flatfile needs to round-trip events, comments, metadata slots, and the full audit trail losslessly -- and ideally reverse (flatfile→Dolt) too, or users who try it and hit the scale/merge ceiling are stuck.

**Value to capture regardless:**

- Decouple and land #4414 (the conformance suite) on its own merits, now. It's a pure addition, it hardens the interface, and it's the prerequisite for any second backend being safe. Wire it into CI against the Dolt backend

- For the backend itself, the charter-consistent shape is a driver/plugin that sits behind the storage boundary, conformance-gated in CI, shipped experimental with docs that state the losses plainly (whole-file merge on same-issue edits, no multi-repo decoupling, O(n) scan ceiling). That keeps the storage-engine logic out of core, honors the Dolt storage ownership line, and still gives solo/OSS users the thing they actually want.

_analysis assisted by claude opus 4.8 xhigh_

---

_Mine is just one maintainer's viewpoint. I'm philosophically aligned with local-first and flat-files for simpler use cases, so I like the intention. I'm unsure about the implementation as it sits. Some of the burden is on our side, for not yet having a clear storage driverboundary and plugin system for people to worth from._

cc @steveyegge @coffeegoddd @julianknutsen

kevglynn:

Ran a complexity-only review on this diff (not correctness, security, or performance — just "what can get shorter"). The core storage implementation is well-structured and lean. The findings are all in the interface-compliance ceremony layer, and a maintainer could absorb most of these on-branch if they're welcome:

`internal/storage/flatfile/transaction.go`: **yagni:** `flatFileTx` is 182 lines of pure delegation — every method calls `tx.s.SameMethod()` with no transactional behavior. A `type flatFileTx struct{ *FlatFileStore }` with only the 2 methods that differ in signature would be ~5 lines.

`internal/storage/flatfile/*.go` (all exported methods): **shrink:** `if err := s.checkClosed(); err != nil { return ..., err }` appears 70+ times (~210 lines). A `closedGuard` wrapper or code-generated decorator at the dispatch level would collapse this.

`internal/storage/flatfile/extensions.go` (CompactionStore, FederationStore): **yagni:** Stub methods all returning nil/false/"not supported". Embedding a `NoopCompactionStore` / `NoopFederationStore` mixin would delete ~50 lines of boilerplate.

`internal/storage/flatfile/gitops.go` (RemoteStore): **native:** `AddRemote`/`RemoveRemote`/`HasRemote`/`ListRemotes` wrap `git remote` commands. If no CLI caller uses `store.AddRemote()`, these exist only to satisfy the interface — consider whether the interface should have a noop mixin too.

`internal/storage/conformance/conformance.go`: **delete:** `var _ = fmt.Sprintf` / `var _ = time.Now` — unused-import anchors are dead code.

**net: ~-650 lines possible** with zero behavioral change. The core issue/dep/search/ready-work logic is solid — this is all ceremony trimming.

Not a blocker — just flagging opportunities in case they're useful before or after merge.

maphew:
Thanks @kevglynn. The four flat-file findings here are for the backend authors, but your `internal/storage/conformance/conformance.go` one (the `var _ = fmt.Sprintf` / `var _ = time.Now` anchors) belongs to #4414, which owns that file - I absorbed it there in commit `9d4076` (dropped the anchors and the now-unused `fmt` import; `time` stays, it's used by `testUpdatePreservesCreatedAt`), credited to you. #4414 also got the conformance suite wired into the embedded + server Dolt backends and running green.

MarkAtwood:
# feat: flat-file storage backend (files-in-git, no CGO)

Supersedes the earlier revision of #4415. Rebased on current `main`, history curated to 6 logical commits.

## What this is

A pure-Go storage backend that keeps one JSON file per issue under `.beads/`, versioned by git itself — no CGO, no Dolt, `cat`/`grep`/`jq`-able, and diffable in a PR. It's the low-infrastructure end of the backend spectrum this repo is already building toward (sqlite/postgres/mysql), not a Dolt replacement: Dolt keeps its cell-level-merge / many-agent story, flatfile serves solo/OSS/files-in-git users, and the two are bridged by lossless bidirectional migration.

It enters through the **same conformance gate** as sqlite — `TestConformance` (160 subtests), interface-completeness, and unsupported-contract — all green, plus a `-race` pass on the concurrency and transaction tests.

## Commits

1. `fix(storage)` — shared-backend correctness fixes surfaced while building parity (see **Cross-backend changes** below)
2. `feat(storage)` — the flat-file backend (store, locking, journaling/rollback, full SQL parity)
3. `feat(bd)` — backend registration, init/doctor wiring
4. `feat(bd)` — bidirectional dolt↔flatfile migration
5. `ci` — conformance gate wiring
6. `docs`

## Process note (why the diff is trustworthy)

After the functional work, the backend was put through **two independent adversarial code-review passes** (13 modules each, cross-cutting-invariant lens included) and every confirmed finding fixed — **~99 findings across both rounds, all fixed except one open design question below**, with a handful refuted-with-rationale rather than force-fixed. The review found and fixed real defects the test suite couldn't see: a `SourceRepo` field that was never persisted (`json:"-"`), two comment data-loss bugs in migration, and a rollback path that couldn't restore files whose parent directory had been removed. Round 2 specifically hunted the seams *between* round-1's fixes. The history is curated so those fix commits are folded into the code they correct — none of those bugs ever existed in shipped code.

## One open question for maintainers

**Cross-prefix dependency classification.** In `computeBlockedSet`, flatfile counts a blocking/waits-for edge whose target has a different ID prefix but exists locally. This matches the Dolt **import** path — but Dolt's two write paths disagree with *each other*: `AddDependency` (command) classifies a cross-prefix target as external and never blocks, while the create/import path (`issueops/create.go`, `ClassifyDepTarget(..., false)`) stores the same edge as local and *does* block. So under the finding's own trigger (`bd import` of another rig's JSONL), flatfile and Dolt already agree.

We've left this as-is rather than guess. A faithful "always match Dolt" fix would require persisting per-edge classification at write time (a flatfile JSON schema addition) — worth doing only once upstream decides **which of Dolt's two write paths is canonical**. Which is correct: should a cross-prefix edge to a locally-present issue block, or not? flatfile will follow that.

## Design positions worth flagging

- **Backend switching is operator-only.** `bd init --server` (or `--database`) without `--backend=dolt` now **errors** instead of silently producing a flatfile workspace. We deliberately rejected auto-selecting Dolt from server flags: an agent must never be able to flip a workspace's backend when the operator asked for flatfile. Switching happens only via an explicit `--backend`, in either direction. (Errors are legible to agents and self-correct in one turn; silent backend drift compounds.)
- **Reverse migration regenerates audit events rather than transferring them** — no backend currently exposes an event-import API, and portable JSONL carries no events even Dolt→Dolt. Everything else round-trips losslessly. A portable-events capability would benefit all backends and is a natural follow-up.
- **`RunInTransaction` gained real rollback** via pre-image journaling; a documented residual gap is directory *moves* (UpdateIssueID's comment-dir/event-file renames) which sit outside the file-level journal.

## Cross-backend changes (commit 1, review separately)

Building parity surfaced fixes that touch the Dolt/SQLite paths, isolated into commit 1 so they can be reviewed on their own merits: exported `issueops.LeaseTTL` for reuse; **LIKE-escaping in `sqlbuild` filters** (a latent wildcard-injection in search affecting all SQL backends); id-parser hardening.

## Not addressed (pre-existing)

`TestInitNonInteractiveAutoExportDefaultOffAndOptIn` fails on `main` before any change in this branch (auto-export flush timing); left untouched.

## Test plan

- `go test ./internal/storage/flatfile/ -count=1` — 160-subtest conformance + unit + concurrency, green
- `... -race` on conformance — green
- `BEADS_TEST_EMBEDDED_DOLT=1 go test ./cmd/bd/ -run TestFlatfileMigrationRoundTrip` — dolt→flatfile→dolt, identical export dumps
- `scripts/conformance.sh` — flatfile gated alongside sqlite
- `golangci-lint` (repo config) — 0 issues; `staticcheck` clean on the backend

MarkAtwood:
Rebased onto current main (`3ecbf5bc9`); CI is green across the board — every code-bearing workflow including the required gate, the embedded-Dolt shards, package gates, corpus, and the real-DB conformance matrix. (The PR Risk bot is the one red mark: it gets no secrets on fork PRs, and fails on this PR's runs regardless of content.) The rebase absorbed the new `UnclaimIssue`/dependency-validation/`LabelsAny` contracts, and full CI shook out four real fixes worth noting for review: a git-identity fallback so flat-file auto-commits work on machines with no `git config user.email`, backend-aware agent templates (Dolt workspaces keep the Dolt guidance verbatim — the earlier unconditional rewrite is gone), cross-backend JSON shape parity for `bd blocked` / `bd dep tree`, and the e2e reference profile now pins `--backend=dolt` explicitly rather than inheriting the default.

The one open design question from the PR description still needs a maintainer call: which of Dolt's two cross-prefix write paths is canonical. Everything else is ready for review.

MarkAtwood:
Correction to my note above: the "PR Risk" red was not a fork-secrets limitation — that workflow risk-tiers CI and runs the full embedded-Dolt suite for large diffs, and it was failing on real harness gaps (embedded tests inheriting the new flat-file default instead of asking for Dolt). Those are fixed as of `8e3659257`, plus one real catch from the same tier: a pre-existing `config.yaml` with `dolt.mode: server` now refuses a defaulted init instead of being silently shadowed by a flat-file workspace. Two residuals are pre-existing on main, not introduced here: the `TestEmbeddedInitConcurrent` flake (Dolt engine close panic under concurrent embedded init — reproduces on main at a higher rate than on this branch) and the embedded-conformance job's occasional 30m timeout (full suite passes in ~7m locally). All six workflows — PR Risk included — are now green on the current head.

MarkAtwood:
Ready to adjust and rebase as soon as #4859 lands. The registry is exactly the landing pattern this backend wants: flatfile becomes one self-registering registrant file behind the blank import, and I'll wire a `Provision` hook so `bd init --backend=flatfile --backend-opt key=value` works through the new contract. Holding off on the conversion until #4859 merges so I'm rebasing onto the final shape rather than chasing it through review.

maphew:
Status signal so you don't spend effort against a moving target — two things have changed under this PR's holding pattern:

1. **Don't hold for #4859.** The registry PR you're waiting on got its review on July 23, and the dominant finding was that main had already landed its own answer while that branch aged: backend *selection* is now the fail-closed tombstone policy in `cmd/bd/backend_support.go` (`RemovedBackendError`/`UnknownBackendError`), not a name-keyed registry. Whether #4859 re-cuts against that is a direction call sitting with its author. Rebasing your backend onto #4859's shape today would mean chasing a branch that may never land in that form.

2. **The harder signal, said plainly: main's storage direction went the other way while this PR waited.** Since mid-July, main removed the SQLite backend and the direct Postgres/MySQL adapters ("Dolt-only" simplification) — the project deliberately contracted from N in-tree backends to one. A new in-tree flat-file backend swims directly against that current, and you should know that before investing in the registry conversion, because the review conversation would likely start there rather than at your code.

The path that *does* fit where the project is heading: duncan4123's external-backend plugin work (#4561 plugin architecture, #4736 configured-backend opener) is exactly the seam a flat-file backend belongs behind — out-of-tree, self-contained, no core storage surface. Those PRs are in active review (changes-requested July 24, author engaged). If flat-file ships as the first real plugin backend, it becomes the proof case for that API instead of a second exception to the simplification.

To be clear about what this comment is and isn't: your PR hasn't had the deep code review it deserves yet (that's on us — 40 days is too long), and nothing above is a verdict on your implementation, which by your own CI notes has real test-harness discipline behind it. It's a direction disclosure so your next move is informed. If you want to make the case for in-tree — the charter conversation is open to that — say so and we'll do the full design review against that question directly. If the plugin path appeals, we can loop you into the #4561 review thread so the API grows the hooks flat-file needs (your `Provision` hook idea is exactly the kind of requirement that should shape it).

MarkAtwood:
@maphew — thanks for the direction disclosure. That was more useful than silence, and I'd rather know now.

Before I spend another rebase, I need a yes or no from a human on one question: does a non-Dolt storage backend belong in this tree at all?

I'm asking because the answer changes everything and nothing else does. The PR is 45 days old with no formal review and no line comments. It was green on all six workflows on July 16 and sat for ten more days. Main has moved 377 commits since my last rebase, and that rebase now crosses the Postgres/MySQL/SQLite removals, so it's days of work touching sqlbuild, the conformance profiles, and init. I'll do it, and I'll do whatever restructure you want on top of it — driver, plugin, out-of-tree, in-tree behind a registrant. But I'd rather not pay it twice, and I'd rather not pay it at all if the charter answer is no.

@steveyegge — this is really a call for you. If it's no, say no and I'll maintain the fork and stop asking. Nobody loses anything.

One thing worth weighing that isn't about my code. #4881's own commit message says that dropping the SQLite profile leaves the CLI differential runner idling — "skip, not fail — until another backend profile is registered." That harness compares two backends against each other, and Dolt is now the only backend, so it has nothing to differentiate against. This PR is the only candidate second profile that exists: 160/160 conformance subtests green, CI gate already wired the way the sqlite profile was, zero conformance tests touched. Whatever you decide about flat-file as a shipped backend, you currently have a test harness with no second operand, and that's a real hole in your Dolt signal.

Also, #4847 kept SQLite explicitly as "a lightweight pure-Go option" two days before #4881 removed it. If that niche is now empty on purpose, fine — but I don't think it was decided, I think it was collateral.

Two smaller notes so they're on the record rather than as an argument:

#4859 merged today. If the answer is "in-tree behind the registrant seam," that's the landing pattern and flatfile fits it — one registrant file behind a blank import, `WorkspaceIsBeadsDir` is exactly our shape. The gap is that the seam covers open and discover but not create, so `bd init --backend=flatfile` is rejected by design and the workspace-creation path has nowhere to live. Same gap in #4561: `Provider` is Name/Capabilities/Open, and `InitParams` sits in the v1alpha1 protocol unused, with no `init` RPC among the 169 methods. For an out-of-tree plugin the current answer is `bd init` provisions Dolt and then `bd backend install` flips the name — which for flat-file means requiring Dolt to create the workspace whose point is not needing Dolt. Wiring `init` to the `InitParams` that's already there would close that for any backend, not just mine.

And the cross-prefix dependency question from my July 12 message is still open and still needs a maintainer call: `AddDependency` classifies a cross-prefix target as external while the create/import path stores it local, so Dolt's two write paths disagree. Flatfile matches the import path. I'm not guessing at which one is canonical.

maphew (commented):
Reviewed exact head `8e3659257ad495ba3a9160d5146bb419d22183ae` against current upstream.

Decisive answer: please do not spend time rebasing this in-tree implementation. Flat-file should not land as a second storage engine inside Beads core under the current Dolt-only storage charter.

The implementation also currently:

- makes flat-file the silent `bd init` default while help still promises Dolt;
- advertises multi-process safety despite relying only on in-process locks;
- can silently strand comments/events during issue-ID renames;
- hard-codes `.beads/` in several custom-directory commit paths; and
- fails its concurrent atomic-write test on Windows.

The contributor value should be preserved through transformation, not discarded. #4414 already retained the conformance work. The remaining implementation, migration requirements, and tests should become the first external backend/plugin after the registry from #4859 gains a narrow provisioning/init lifecycle. Later-filed #4561 points in that materially better direction but itself needs simplification onto the merged registry; #4736 is not a viable base.

Recommendation: replace the current PR with an external-driver implementation carrying explicit co-author attribution from #4415. The current branch conflicts with upstream in 17 paths, so asking for a rebase before that architectural move would create substantial throwaway work. I am not opening a replacement or closing this PR in this review sweep.

maphew:
@MarkAtwood - after off channel discussion on the ideas and code in this PR the decision is: we want it, but not in main tree. Our job is make the external path real and open the door for you.

 @steveyegge will followup with technical details on what that means.

steveyegge:
@MarkAtwood — following up on maphew's note with the mechanics. First, the direct answer you asked for on July 31: **no, a non-Dolt backend doesn't belong in this tree** — not because the work is lacking (it isn't; more below), but because a second in-tree backend puts the semantic-drift tax on core forever, and the charter says core doesn't carry that. What we *do* want is this backend alive, external, and conformance-gated — and the door-building is our job, not yours.

**The path, concretely.** #4859 merged July 31, so the registry seam is real in main: flatfile becomes its own Go module registering via the blank-import registrant pattern you described on July 17, living in your repo. You offered to wire a `Provision` hook so `bd init --backend=flatfile` works through the contract — yes please, PR it against the merged seam; we commit to reviewing it in days, not weeks. On our side, filed and owned by us:

1. The conformance suite currently sits under `internal/` where your module can't import it — we're exporting it (together with the backend interface surface it exercises) as a public package, so an external backend can implement the contract and CI against the same assertions core uses.
2. The differential harness gets optional second-profile support so flatfile can run harnessed.
3. Conformance CI gets a macOS leg — we reproduced a real APFS failure (below) that Linux-only CI structurally can't see.

**What the deep review found** (three parallel passes plus a local suite run — this is the detail you were owed on the record, not in a back channel). Ship-blockers before we'd link flatfile as a conformance-passing community backend:

1. **No cross-process locking.** Two concurrent `bd` processes can double-win `--claim`, double-acquire the merge slot, and silently lose a created issue (counter read-modify-write). Needs flocks.
2. **Journal is memory-only** — SIGKILL leaves torn state, contrary to a code comment claiming otherwise.
3. **Raw ID = filename**, so `test-pc`/`test-PC` collide on APFS — conformance fails on macOS today, exactly where your target users live. Needs case-encoded filenames.
4. **Five semantic parity gaps** that got past the gate: ready-sort tie-break inverted (different result set under `--limit`), custom active statuses unclaimable, stale `started_at` on unclaim, no-history wisps vanish, ListWisps sort inverted.
5. Every git op except Commit hardcodes `.beads/` (pathspec), and `bd graph` reparses O(V·n).

**Credit where due:** the differential run against your backend found two real bugs in the *Dolt* backend — the cross-prefix self-disagreement you reported, plus `bd ready --parent` leaking unrelated wisps. Both are filed on our side, and we're ruling on cross-prefix canonicalization at the interface level. A second backend is the best oracle this interface will ever have — that's the argument for keeping yours alive and harnessed, and it's why the door matters to us and not just to you.

On the strategic question underneath: the pure-Go niche didn't die on purpose with the SQLite removal — the external path is our answer to it.

#4415 itself is your call: leave it open as the anchor until the Provision hook and conformance export land, or close it now and we track the door work on our side. Either way you'll hear from us when each piece merges, starting with review on your Provision PR whenever you're ready.

— drafted and driven by claude-fable-5 (bee) on behalf of steveyegge

steveyegge:
First door piece is up: #5321 exports the storage contract (`backend/`) and the conformance suite (`backend/conformance/`) out of `internal/`, with an out-of-tree stub module as the acceptance test. Once it merges, your module can implement `backend.DoltStorage` and run `conformance.RunAll` directly — and the Provision hook PR you offered has a public surface to land against whenever you're ready.

steveyegge:
Following up on the provisioning seam ahead of the offered `bd init --backend=flatfile --backend-opt k=v` PR — here is the hook shape we'd expect, so the PR lands clean on the first pass. *(Posted by claude-fable-5 ("bee") on behalf of @steveyegge.)*

**Registry:** one optional field on `Backend` in `internal/storage/backends/backends.go`:

```go
// Provision creates a new workspace for this backend under beadsDir.
// Optional: nil means the backend opens existing workspaces only, and
// bd init keeps failing closed for it.
Provision func(ctx context.Context, beadsDir string, opts map[string]string) error
```

**Dispatch:** `cmd/bd/init.go` currently rejects every registered backend fail-closed (the check at the top of backend selection). With this hook, the rejection keys on `Provision == nil` instead — same error text for hookless backends, a minimal provisioning path otherwise.

**Division of labor (the important part):**
- `bd init` owns: flag validation, creating the empty `.beads` dir, invoking `Provision`, writing `metadata.json` with `backend: <name>` *only after* `Provision` returns nil, then opening the store via the backend's registered `Open` and writing `issue_prefix` through the normal storage-config surface — so every backend shares one prefix path.
- `Provision` owns: the backend's own substrate under `beadsDir` — files, remote refs, its own config — including persisting whatever `--backend-opt` values it needs. Opaque to core; unknown or invalid opts are the hook's job to reject.
- **Postcondition:** after `Provision` returns nil, `backend.Open(ctx, beadsDir)` succeeds and the resulting store passes the conformance surface (#5321's harness is the acceptance bar).
- **Fail-closed:** if `Provision` errors, init removes the `.beads` dir it created — `TestInitRejectsRegisteredBackend`'s invariant (no workspace written on failure) extends to provisioning failure; a sibling test should pin it.

**`--backend-opt k=v`:** repeatable string flag on `bd init`, parsed to `map[string]string`, passed verbatim to `Provision`, never persisted or validated by core.

**Scope fences for v1:** the provisioning path supports `--prefix`, `--quiet`, `--non-interactive`, `--backend-opt`; combining a registered backend with Dolt-only surfaces (`--server`, `--shared-server`, `--proxied-server-*`, `--database`, `--contributor`, `--team`) errors out fail-closed. `bd bootstrap` stays Dolt-only for now.

**One open choice** we'll happily take your view on in the PR: whether `issue_prefix` is written by init through `Open` (as above — keeps one shared prefix path) or by `Provision` itself. We lean init-owns-it.

The review commitment upthread stands: days, not weeks, once the PR is up.

steveyegge:
Update: #5321 is merged (`227d0f5f4`). The conformance harness and contract surface are now public on main — `github.com/steveyegge/beads/backend` (contract types + registry) and `backend/conformance` (the acceptance harness). The package docs carry the implementer recipe; the provision-hook sketch two comments up still stands as the expected shape for the `bd init --backend=flatfile` wiring. Looking forward to the PR. *(claude-fable-5 ("bee") on behalf of @steveyegge.)*

rassie:
@MarkAtwood is your work available anywhere as pluggable storage already?

PR #3669, opened 2026-05-03
feat(linear): ambient staleness signal — bd prime auto-pull + --pull-if-stale

kevglynn:
## Summary

- Add ambient staleness detection so agents and devs automatically get fresh Linear data without retry infrastructure
- `--pull-if-stale` flag on `bd linear sync` with configurable `--threshold` (default 20m) and 5-minute debounce
- `bd prime` auto-pulls from Linear when data is stale and `LINEAR_API_KEY` is set
- Per-session stderr warning on any `bd` command when Linear data is stale

## Motivation

In a local-first system, cron-based pull can silently go stale (laptop sleep, network issues). Rather than adding retry infrastructure, this makes staleness _ambient_ — the tool itself notices and acts. Agents get `--pull-if-stale` for zero-cost freshness checks; humans get a one-time-per-session warning.

## Design

Core staleness library (`internal/linear/staleness.go`):
- `WriteLastPullTimestamp` / `ReadLastPullTimestamp` — file-based timestamp in `.beads/last_pull`
- `IsPullStale(dir, threshold)` — checks if data exceeds configurable threshold
- `IsWithinDebounce(dir, window)` — prevents agent loops with 5-min debounce
- `GetStalenessInfo` / `FormatAge` — human-readable status

Integration points:
- `cmd/bd/linear.go`: `--pull-if-stale` + `--threshold` flags, debounce short-circuit, timestamp write after successful pull
- `cmd/bd/prime.go`: auto-pull during `bd prime` when stale
- `cmd/bd/main.go`: per-session stderr warning via `maybeWarnLinearStaleness`

## Scalability

Validated by adversarial architecture + performance review:
- Morning storm scenario (50 devs, all `bd prime` at 9am) = ~50 API calls = 1% of Linear rate limits
- Scales to 500 devs safely with current design
- Debounce prevents agent loop amplification

## Test plan

14 new tests covering all staleness functions:
- Round-trip write/read
- Missing/empty/malformed file handling
- Stale/fresh detection
- Debounce window behavior
- `FormatAge` formatting
- All 51 tests pass in `internal/linear/`

---
[...]

PR #4419, opened 2026-06-16
feat(metrics): anonymous CLI usage metrics + RunE refactor

julianknutsen:
## What

Adds anonymous CLI usage metrics and converts every command to cobra's `RunE`
(return errors instead of `os.Exit`). Cherry-picked from @coffeegoddd's PR #2
(`refactor to use RunE, emit usage metrics, skip metrics in CI`); his commit and
authorship are preserved. A second commit applies fixes from an exhaustive code
review.

- **`internal/metrics`** (new): emits a `cli_command` event per command via
  dolthub/eventkit — the command name only, with each batch carrying the bd
  version and OS platform, keyed by a machine-derived HMAC-protected distinct ID.
  Written under `~/.beads/eventsData`, POSTed to
  `https://gastownhall-eventsapi.com/mp/collect`.
- **RunE conversion**: commands return errors so the deferred metrics event is
  recorded on the way out; CI lanes set an env var to skip emission.
- **Consent UX** (added here): a friendly first-run notice + a `bd metrics`
  command (status / on / off / example) to see and control metrics with one
  command — no config editing, no env var, no restart.

## Why

Usage telemetry for `bd` (owner: @coffeegoddd). The feature was developed on a
separate branch (PR #2) off the same base as `main`; this lands it cleanly on
current `main` with merge resolutions and review fixes so it can ship as a
single, focused change.

## Provenance & merge resolutions

Cherry-picked the **content commit** `9596c7233` only — the PR-#2 merge wrapper
`762bdba66` is tree-identical and would duplicate. Nine `cmd/bd` files conflicted
because the feature predates recent `main` work; each was resolved to combine the
RunE refactor with the kept logic:

- `delete/dep/edit/human/link/reopen` — kept the `#4141` write-intent routing (`…ForWrite`) under the RunE wrapper
- `children` — kept the `--pretty` passthrough + `return listCmd.RunE(...)`
- `main` — kept the `#4259` remote-migrate-gate handling + `return HandleError(...)`
- `prime` — took only the `metrics` import (the kept `#3669` linear-staleness removal stands)
- 4 leftover `FatalError`/`FatalErrorRespectJSON` calls in newer kept code (`count/defer/vc`) → the refactor's `return Handle…` form

## Review findings (exhaustive review run; confirmed findings fixed)

**Fixed (commit 2):**
- `metrics.NewCommandEvent` panicked on an empty command name → falls back to `"unknown"` (a telemetry helper must not crash a command).
- `cmd/bd/init` `#4259` gate path exited via `os.Exit(1)`, skipping the deferred metrics event, post-run flush, and `initLock.Unlock` → `return &exitError{Code: 1}` like every sibling path.
- Telemetry docs referenced a nonexistent `--global` flag (would error for users opting out) → corrected; endpoint/payload/disable now documented.

**Consent UX added (3rd commit):**
- The telemetry consent posture (previously silent opt-out) is now transparent and friction-free:
  - A one-time, friendly **first-run notice** (stderr) tells users metrics are on and how to opt out; it records `metrics.notice_shown` so it never repeats, and is suppressed in JSON/hook/quiet and metrics-command contexts.
  - A new **`bd metrics`** command: `bd metrics` (status + what's collected + endpoint), `bd metrics on` / `bd metrics off` (writes user-global config directly — no manual editing, no env var, takes effect next command with no restart), and `bd metrics example` (shows a concrete `cli_command` payload matching the real wire shape and pretty-prints the **real queued events** buffered under `~/.beads/eventsData`).
  - `bd metrics` is store-independent, so users can always check/change it even if the DB is unavailable. `BD_DISABLE_METRICS` still works and is surfaced if it disagrees with the saved config.

**Deferred as follow-ups (kept the PR focused / preserved contributor design):**
- `CheckReadonly()` `os.Exit(1)` drops the usage event when a write command is rejected in read-only mode (~10 commands). A correct fix changes `CheckReadonly`'s contract across all callers — out of scope for this cherry-pick; worth a follow-up (and a CI grep gate barring `os.Exit` in RunE bodies).
- Minor consistency nits from the refactor (a few `SilenceUsage` omissions, a `Run`-vs-`RunE` holdout in the `human` parent command, JSON-stream choices on a few error paths). Logged, not blocking.

## Verification

- `make build`, `go build -tags gms_pure_go ./...`, `go vet ./...` — clean
- `internal/metrics` tests — pass
- Full `make test` — **60 packages ok**; the only failure is the pre-existing `cmd/bd` flake noted below (not a regression).

> Note: any `cmd/bd` flakiness under the parallel runner (`TestCommitBeadsConfigSkipsGitHooks`, etc.) is a pre-existing `BEADS_DIR`/cwd test-isolation issue that reproduces on untouched `main` and passes in isolation — not introduced here.

### DIFF
--- a/.github/workflows/ci-measurements.yml
+++ b/.github/workflows/ci-measurements.yml
@@ -46,6 +46,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.suite }}
   cancel-in-progress: false
 
+env:
+  BD_DISABLE_METRICS: "1"
+  BD_DISABLE_EVENT_FLUSH: "1"
+
 jobs:
   pr-linux:
     name: Measure PR Linux commands
--- a/.github/workflows/cross-version-smoke.yml
+++ b/.github/workflows/cross-version-smoke.yml
@@ -22,6 +22,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+env:
+  BD_DISABLE_METRICS: "1"
+  BD_DISABLE_EVENT_FLUSH: "1"
+
 jobs:
   smoke:
     name: Upgrade smoke (${{ matrix.prev_version }} → candidate)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+env:
+  BD_DISABLE_METRICS: "1"
+  BD_DISABLE_EVENT_FLUSH: "1"
+
 jobs:
   detect-ci-tier:
     name: Detect CI tier
--- a/.github/workflows/migration-test.yml
+++ b/.github/workflows/migration-test.yml
@@ -19,6 +19,10 @@ on:
           - direct-only
           - stepping-only
 
+env:
+  BD_DISABLE_METRICS: "1"
+  BD_DISABLE_EVENT_FLUSH: "1"
+
 jobs:
   migration-test:
     name: Migration fidelity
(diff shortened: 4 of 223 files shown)


coffeegoddd (approved):
LGTM!

PR #4424, opened 2026-06-16
fix(remember): reject subcommand-like bare word instead of storing it

### LINKED ISSUE #4401: beads CLI: remember treats a subcommand-like first argument as memory content
## Summary

`bd remember <word>` can interpret a subcommand-like first argument as memory **content** rather than as a subcommand, leading to a surprising accidental memory write when a user expects a subcommand to run.

## Environment

- Observed on `bd` 1.0.4 (and 1.0.3). Please confirm against latest `main`.

## Reproduction

Invoking `bd remember` with a bare word that a user might intend as a subcommand stores that word as the remembered content instead of dispatching a subcommand:

```
bd remember <word>
# stores "<word>" as memory content (no subcommand dispatch / no confirmation)
```

(The exact word that triggered this in our case was an accidental invocation and was not captured verbatim; the symptom is the verb/noun ambiguity of `remember`'s first positional argument.)

## Expected

Either `remember` reserves recognizable subcommand tokens, or it requires/echoes an explicit confirmation of what is being stored, so a mistaken invocation does not silently write memory.

## Actual

The first positional argument is treated as content; a mistaken invocation silently writes a memory entry.

## Workaround

No known workaround beyond careful invocation discipline (treat `bd remember`'s content argument explicitly, e.g. quote it and double-check before running).

## Severity / impact

**Low–Medium** — usability/safety. No data loss, but it produces unexpected memory writes that a user did not intend.

harry-miller-trimble:
## Summary

`bd remember` is a leaf command (`cobra.ExactArgs(1)`), so a mistaken `bd remember recall` — or any bare bd command name like `status`, `forget`, `memories` — silently stored that word as a memory instead of doing what the user intended. No subcommand dispatch, no confirmation.

This guards the content argument: if it is a **single bare word** that matches the name or alias of a top-level bd command, `remember` refuses with an actionable hint instead of writing a memory.

- Genuine insights are phrases (they contain whitespace), so they pass through untouched — zero impact on normal use.
- An explicit `--key` signals deliberate intent and bypasses the guard, so storing the literal word is still possible.

## Behavior

```
$ bd remember recall
Error: "recall" looks like a command, not something to remember
Hint: Did you mean 'bd recall'? To store "recall" as a memory anyway, give it an explicit key: bd remember "recall" --key <key>

$ bd remember "always run tests with the -race flag"
Remembered [always-run-tests-with-the-race-flag]: always run tests with the -race flag

$ bd remember recall --key my-recall-note
Remembered [my-recall-note]: recall
```

`--json` mode emits a structured `{"error":..., "hint":...}` object (via `FatalErrorWithHintRespectJSON`).

## Testing

- New pure-Go regression test `TestMatchesKnownCommand` (no cgo/Dolt dependency) covering sibling commands, case-insensitivity, multi-word insights, a leading-command-word phrase, non-command words, and empty/whitespace.
- `go build -tags gms_pure_go ./cmd/bd/` clean; `gofmt`/`go vet` clean.
- Verified end-to-end against an embedded workspace (refuse / store / escape-hatch / JSON cases above).

Fixes #4401

### DIFF
--- a/cmd/bd/memory.go
+++ b/cmd/bd/memory.go
@@ -44,6 +44,29 @@ func slugify(s string) string {
 	return slug
 }
 
+// matchesKnownCommand reports whether insight is a single bare word that
+// matches the name or an alias of a top-level bd command. It is used to catch
+// `bd remember <subcommand>` mistakes before they become accidental memories.
+// Multi-word insights (the normal case) always pass, since they contain
+// whitespace and so cannot be a single command token.
+func matchesKnownCommand(cmd *cobra.Command, insight string) (string, bool) {
+	word := strings.TrimSpace(insight)
+	if word == "" || strings.ContainsAny(word, " \t\r\n") {
+		return "", false
+	}
+	for _, c := range cmd.Root().Commands() {
+		if strings.EqualFold(c.Name(), word) {
+			return c.Name(), true
+		}
+		for _, alias := range c.Aliases {
+			if strings.EqualFold(alias, word) {
+				return c.Name(), true
+			}
+		}
+	}
+	return "", false
+}
+
 // rememberCmd stores a memory.
 var rememberCmd = &cobra.Command{
 	Use:   `remember "<insight>"`,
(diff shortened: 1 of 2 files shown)


maphew:

PR #4426, opened 2026-06-16
docs(skill): fix decision skill referencing nonexistent `bd decision record`

### LINKED ISSUE #4291: bd decision record - Does not exist but exist in SKILLs causing misfire.
Kindly check this file for the source of the issue.

plugins/beads/skills/beads/commands/decision.md

-----

Assessment 👍 
-------

Confirmed — line 12 is the only place bd decision appears in the entire file. Every other command block uses the correct bd create --type decision, bd 
  list --type decision, etc.

  Why that line specifically causes it

  When the user wants to record a decision (or you invoke `bd decision record`):

  The parenthetical reads as a second, equivalent trigger: "…or you invoke bd decision record." An agent skimming the skill sees a backtick-wrapped bd 
  decision record presented as an action it can take, and runs it literally as a shell command → unknown command "decision". The intended meaning was "or
  when the /beads:decision record slash-command path fires" — but the prose dropped the slash and wrote it like a CLI verb.

  It's purely a wording bug: the parenthetical names the slash-command action using CLI-command syntax, and those don't match because no bd decision
  subcommand exists.

  The fix

  Remove or reword just that parenthetical. Two clean options:

  # Option A — drop it entirely
  When the user wants to record a decision:

  # Option B — keep intent, fix the reference
  When the user wants to record a decision (i.e. the `record` action of this command):

harry-miller-trimble:
## Summary

The decision skill (`plugins/beads/skills/beads/commands/decision.md`) had its "Record a Decision" section read:

> When the user wants to record a decision (or you invoke `bd decision record`):

There is no `bd decision` subcommand. An agent skimming the skill sees a backtick-wrapped `bd decision record` presented as an action and runs it literally → `unknown command "decision"`. The intended meaning was the `record` action of this slash-command (frontmatter: `argument-hint: record|list|show|supersede`), not a CLI verb.

## Fix

Reword the parenthetical to name the slash-command action instead of a fake CLI command:

```diff
-When the user wants to record a decision (or you invoke `bd decision record`):
+When the user wants to record a decision (i.e. the `record` action of this command):
```

This is the only `bd decision` reference in the file (and the repo); every other command block already uses the correct `bd create --type decision` / `bd list --type decision` forms.

Fixes #4291

### DIFF
--- a/plugins/beads/skills/beads/commands/decision.md
+++ b/plugins/beads/skills/beads/commands/decision.md
@@ -9,7 +9,7 @@ Decisions use `--type decision`. The description field holds the structured deci
 
 ## Record a Decision
 
-When the user wants to record a decision (or you invoke `bd decision record`):
+When the user wants to record a decision (i.e. the `record` action of this command):
 
 1. Gather the following (ask if not provided):
    - **Title**: Short summary of what was decided (required)


maphew (commented):
Thanks for the careful fix, Harry. I've verified that `bd decision` is not a subcommand anywhere in beads — `decision` is only valid as a `--type` argument to `bd create` and `bd list`. Your reword from "or you invoke `bd decision record`" to "i.e. the `record` action of this command" correctly points agents to the skill's documented actions instead, preventing them from trying to run a nonexistent command. I've confirmed this is the only instance of this pattern in the repo; all other sections already use the correct forms. CI is green. This is ready to merge.

PR #4429, opened 2026-06-16
fix(migrate): compute repo fingerprint from -C target, not process cwd

### LINKED ISSUE #4361: bd -C <dir> migrate --update-repo-id computes the new fingerprint from the process cwd, not the -C target
## summary

`bd -C <repo> migrate --update-repo-id` derives the **new** repository fingerprint from the process working directory instead of the `-C` target. if you run it from a different repo's directory, it offers to stamp the target DB with the *wrong repo's* fingerprint - and since the metadata table syncs, the bad value would propagate to every machine on the next push.

`bd doctor` honors `-C` correctly for the same computation, which is how i caught it: dry-run and doctor disagreed.

## version

`bd 1.0.5 (c6802bcc2)` - tip of main as of 2026-06-10. shared-server mode, windows (also confirmed the doctor side on wsl + macos).

## repro

two repos, A and B, each with its own beads DB. from inside A:

```
bd -C /path/to/B migrate --update-repo-id --dry-run
```

observed: `New:` shows **A's** fingerprint. i had two DBs with stale fingerprints (remote URLs had changed) and ran the dry-run via `-C` from a third repo - both reported the *identical* `New:` value (the third repo's hash), while `bd -C <repo> doctor` correctly reported each repo's own expected current fingerprint:

```
# bd -C ~/tools/sym doctor        -> stored: 7e3cad59, current: 7d035d82
# bd -C ~/tools/tpane doctor      -> stored: cef2bed2, current: 5cb2b0cc
# from ~/.files:
bd -C ~/tools/sym  migrate --update-repo-id --dry-run   -> New: f898cd35   (= ~/.files' fingerprint)
bd -C ~/tools/tpane migrate --update-repo-id --dry-run  -> New: f898cd35   (same wrong value)
```

cd'ing into each repo and running the same command without `-C` produces the correct per-repo values (matching doctor), and the update then works as intended.

## impact

without the dry-run + doctor cross-check this writes a wrong-repo fingerprint into a synced table - every other machine inherits it on pull and their doctors go red with a misleading "database belongs to different repository". easy to misdiagnose as the wrong-database case and reach for `rm -rf .beads && bd init`.

## related

#4241 looks like the same underlying class - `-C` resolving repo-derived state from the caller's cwd instead of the target - but that one's a spurious warning on a read path, this one's a wrong value on a write path to a synced table. filing separately for severity; if you'd rather fold them into one -C audit, happy to consolidate there. (i've also seen one unconfirmed case of `bd -C <dir> info` reporting the cwd project - same smell.) worth a quick audit of wherever the fingerprint helper gets its path.

harry-miller-trimble:
## Problem

`bd -C <dir> migrate --update-repo-id` derives the **new** repository fingerprint from the process working directory, not the `-C` target. `-C` doesn't `chdir` (it resolves `BEADS_DIR`), but `beads.ComputeRepoID()` shells out to git in the caller's cwd. So:

- Run from a **different repo**, it stamps the target DB with the *caller's* fingerprint.
- Run from a **non-git cwd**, it fails outright with `not a git repository` despite a valid `-C` target.

Because `repo_id` lives in the **synced metadata table**, the bad value propagates to every clone on the next push and their `bd doctor` goes red with a misleading "database belongs to different repository". `bd doctor` already computes this correctly (`ComputeRepoIDForPath(beadsDir)`); migrate disagreed, which is how the reporter caught it.

Fixes #4361.

## Repro (before)

Two repos A and B, each with its own beads DB. From **inside A**:

```
$ bd -C /tmp/repoB migrate --update-repo-id --dry-run   # New: 1814c071  (A's!)
$ bd -C /tmp/repoA migrate --update-repo-id --dry-run   # New: 1814c071  (A's)
$ cd /tmp && bd -C /tmp/repoA migrate --update-repo-id --dry-run
Error: failed to compute repository ID: not a git repository
```

Both `-C` targets report the **same** fingerprint (the cwd's).

## Fix

Compute repo-derived metadata from the resolved `.beads` directory (which already honors `-C`) instead of the process cwd, mirroring `bd doctor`:

- `handleUpdateRepoID` → `ComputeRepoIDForPath(beadsDir)`
- `handleDoltMetadataUpdate` repo_id/clone_id backfill → `ComputeRepoIDForPath(beadsDir)` / `GetCloneIDForPath(beadsDir)` (same defect on the plain `bd migrate` path; threaded `beadsDir` in)

## After

```
# from inside A:
bd -C /tmp/repoB ... --dry-run   # New: 09b30495  (B's)
bd -C /tmp/repoA ... --dry-run   # New: 1814c071  (A's)
# from /tmp (no git repo): both targets report their own fingerprint, no error
```

The `-C` values now match the per-repo control values (`cd` in, no `-C`) exactly.

## Test

Adds `TestEmbeddedMigrate/migrate_update_repo_id_honors_C`: two repos, asserts `bd -C <B>` from cwd A reports B's fingerprint, not A's. Verified it **fails** on the pre-fix code (`reported A's fingerprint … computed from cwd, not -C target`) and passes with the fix. `gofmt`/`go vet` clean; pure-Go (`CGO_ENABLED=0`) compile gate passes.

### DIFF
--- a/cmd/bd/migrate.go
+++ b/cmd/bd/migrate.go
@@ -92,11 +92,15 @@ BD_ALLOW_REMOTE_MIGRATE=1 remains supported for scripted/CI use.
 			return HandleError("failed to load config: %v", err)
 		}
 
-		return handleDoltMetadataUpdate(cfg, dryRun)
+		return handleDoltMetadataUpdate(cfg, beadsDir, dryRun)
 	},
 }
 
-func handleDoltMetadataUpdate(cfg *configfile.Config, dryRun bool) error {
+// handleDoltMetadataUpdate handles version metadata updates for Dolt backends.
+// beadsDir is the resolved .beads directory (honoring -C); repo-derived metadata
+// is computed from it rather than the process cwd so `bd -C <dir> migrate`
+// fingerprints the target repo, not the caller's (GH#4361).
+func handleDoltMetadataUpdate(cfg *configfile.Config, beadsDir string, dryRun bool) error {
 	ctx := rootCtx
 	store := getStore()
 	if store == nil {
@@ -207,7 +211,7 @@ func handleDoltMetadataUpdate(cfg *configfile.Config, dryRun bool) error {
 
 	// Set repo_id if missing (non-fatal — may fail in non-git environments)
 	if needsRepoID {
-		computed, err := beads.ComputeRepoID()
+		computed, err := beads.ComputeRepoIDForPath(beadsDir)
 		if err != nil {
 			if !jsonOutput {
 				fmt.Fprintf(os.Stderr, "Warning: could not compute repo_id: %v\n", err)
@@ -228,7 +232,7 @@ func handleDoltMetadataUpdate(cfg *configfile.Config, dryRun bool) error {
 
 	// Set clone_id if missing (non-fatal — may fail in non-git environments)
 	if needsCloneID {
-		computed, err := beads.GetCloneID()
+		computed, err := beads.GetCloneIDForPath(beadsDir)
 		if err != nil {
 			if !jsonOutput {
 				fmt.Fprintf(os.Stderr, "Warning: could not compute clone_id: %v\n", err)
@@ -304,7 +308,11 @@ func handleUpdateRepoID(dryRun bool, autoYes bool) error {
 		return HandleErrorWithHint("no beads database found", diagHint())
 	}
 
-	newRepoID, err := beads.ComputeRepoID()
+	// Compute new repo ID from the resolved .beads directory (honoring -C),
+	// not the process cwd. Otherwise `bd -C <dir> migrate --update-repo-id`
+	// stamps the target DB with the caller repo's fingerprint and the bad
+	// value propagates to every clone on the next sync (GH#4361).
+	newRepoID, err := beads.ComputeRepoIDForPath(beadsDir)
 	if err != nil {
 		if jsonOutput {
 			if jerr := outputJSON(map[string]interface{}{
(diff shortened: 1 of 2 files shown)


maphew (commented):
Thanks for tracking this down, @harry-miller-trimble - nice catch that bd doctor and bd migrate --update-repo-id disagreed on the same computation, and the repro in the issue (#4361) makes the bug obvious.

The fix looks correct: routing the repo_id/clone_id computation through ComputeRepoIDForPath(beadsDir)/GetCloneIDForPath(beadsDir) instead of the cwd-based ComputeRepoID()/GetCloneID() mirrors what bd doctor already does, and it closes the exact gap described - both the bd -C <dir> migrate --update-repo-id path and the plain migrate metadata backfill (repo_id/clone_id). The new TestEmbeddedMigrate/migrate_update_repo_id_honors_C test is a solid regression check: it sets up two repos with distinct fingerprints and asserts a -C-targeted dry-run from a different cwd reports the target's fingerprint, not the caller's. All CI (including the embedded-Dolt suite and upgrade smoke tests) passed at the time of the run.

Since this PR was opened, main picked up an unrelated refactor of cmd/bd/migrate.go (RunE returning error), so GitHub now shows this branch as conflicting. From a look at current main, the conflict is purely mechanical - handleUpdateRepoID already resolves its own beadsDir locally, and handleDoltMetadataUpdate's caller already has beadsDir in scope - so it should rebase cleanly with a small edit. I'll pull this in as a fix-merge: reapply the fix on top of current main, keep your test as-is, and push so this can land with your authorship preserved. Will follow up with the rebased branch shortly.
</draft_comment>

PR #4430, opened 2026-06-16
fix(wisp): never age-GC active (blocked/in-progress) molecule steps

### LINKED ISSUE #4394: bd mol wisp gc --age deletes active (BLOCKED) molecule steps
### Observed Behavior

Running `bd mol wisp gc --age 1h --force` during patrol cycles deletes BLOCKED steps that were created minutes ago, not hours. This causes active molecules to self-destruct mid-execution.

**Evidence from reproduction:**
- mol-deacon-patrol (26 steps) → 1 step (25 steps deleted mid-execution)  
- mol-deacon-patrol (26 steps) → 2 steps (24 steps deleted)
- Same command run on other patrol formulas (witness, refinery) likely caused similar molecule kills

### Root Cause Analysis

The `--age` filter is defined as: 
> A wisp is considered abandoned if: It hasn't been updated in --age duration and is not closed

**Expected behavior:** BLOCKED steps created minutes ago should NOT match "hasn't been updated in --age duration" since they're recent.

**Suspected root causes:**
1. Timestamp tracking broken: When a step transitions to BLOCKED status, `wisp.updated_at` is not being updated to current time
2. Wrong field used: The comparison may be using `created_at` instead of `updated_at`  
3. BLOCKED status treated as terminal: BLOCKED steps incorrectly grouped with closed/abandoned wisps
4. Timezone/clock issue: Time comparison logic uses wrong timezone or system clock skew

### Impact

- Production patrol molecules (deacon, witness, refinery) were self-destructing every cycle
- Prevents any formula from safely running `bd mol wisp gc --age` while active molecules are running
- Requires workaround: use only `--closed` gc or exclude active molecules via `--exclude-type mol`

### Workaround Applied

Removed `bd mol wisp gc --age 1h --force` from all patrol formulas in gas-town. Now using only `--closed` gc which is safe.

### Steps to Reproduce

1. Start an active molecule with multiple steps (especially a patrol formula with 20+ steps)
2. Run `bd mol wisp gc --age 1h --dry-run` while molecule is executing
3. Verify BLOCKED steps <5min old are listed as candidates for deletion  
4. Run `--force` to confirm deletion of active steps

### Proposed Fixes

**Option 1 (preferred):** Fix timestamp tracking
- Ensure BLOCKED steps update `updated_at` to current time when status changes
- Review wisp lifecycle code to confirm all status transitions update the timestamp

**Option 2:** Add molecule protection
- Introduce `--exclude-type mol` default behavior or smart detection
- Protect all molecule-type wisps from gc --age deletion

**Option 3:** Add status awareness
- Make gc --age smart about BLOCKED status: don't consider BLOCKED as abandoned
- Require explicit `--include-blocked` flag to clean BLOCKED steps by age

**Option 4:** Default behavior change  
- Change --age to only clean truly abandoned wisps (last_activity > age AND not in active graph)

Recommend Option 1 (timestamp fix) as the root cause, but Option 3 (status awareness) as a safer band-aid while Option 1 is being fixed.

harry-miller-trimble:
## Problem

`bd mol wisp gc --age <dur>` deletes **active** molecule steps, self-destructing running molecules mid-execution. From the report: `mol-deacon-patrol (26 steps) → 1 step` — 25 active steps deleted in a single GC pass. The workaround was to drop `gc --age` from all patrol formulas entirely.

Fixes #4394.

## Root cause

GC judged a wisp's age purely by `updated_at`, and the only state it excluded was `closed`. But:

- A step waiting its turn is **blocked**, which in beads is the derived `is_blocked` column — and its recompute **deliberately preserves `updated_at`** (`SET is_blocked = 1, updated_at = updated_at`, see `RecomputeIsBlocked`) so blocking churn doesn't reset activity timestamps.
- So a step that has been blocked longer than `--age` looks "stale" by `updated_at` and was reclaimed — even though it is live work.
- The dependent-cascade (added to avoid orphaning step children) then swept the rest of the molecule along with the abandoned parent.

Reproduced: two ephemeral wisps where the child is blocked on the parent; `gc --age 1ms --dry-run` listed **both** the blocked child and an in-progress step as candidates.

## Fix

Treat a wisp as **active** — and exempt from age-based GC — when it is:
- blocked on an open dependency (`GetBlockedIssues` / `is_blocked`), or
- in an in-flight status: `in_progress`, `blocked`, `hooked`, `pinned`.

The guard is applied in both the direct age scan **and** the dependent-cascade, so active steps are never deleted nor cascaded into. Genuinely idle (`open` + stale) and `closed` wisps are still collected exactly as before. After the fix the same repro protects the blocked child and the in-progress step, leaving only the idle open wisp as a candidate.

This implements the issue's recommended Option 3 (status/activity awareness). Note: bumping `updated_at` on block (Option 1) was rejected — it's intentional that derived-state recompute doesn't touch activity timestamps, and it wouldn't protect a step legitimately blocked longer than `--age`.

## Test

`TestWispGCProtectsActiveWisps` (embedded): a blocked step and an in-progress step survive `gc --age`, while an idle open wisp remains a candidate. Verified it **fails** on pre-fix code (blocked + in-progress wrongly listed) and passes with the fix. Existing `TestWisp*` and storage-layer `TestWispGC`/`TestFindWispDependents`/`TestDeleteWispBatch` still pass. `gofmt`/`go vet` clean; pure-Go compile gate passes; CLI reference regenerated for the updated help text.


maphew (commented):
Thanks Harry - this is a solid fix for a real incident. The root-cause analysis is right: `is_blocked` recompute deliberately preserves `updated_at` so blocking churn doesn't reset activity timestamps, which meant a step blocked longer than `--age` looked stale and got reclaimed, taking the rest of its molecule with it via the dependent cascade. Guarding both the direct age scan and the cascade with `isActiveWisp` (blocked-on-open-dependency OR in-flight status) is the right shape for the fix, and the new embedded test (verified failing pre-fix, passing post-fix) is a good regression guard alongside the existing wisp GC tests.

One mechanical issue: this PR shows as conflicting with `main`. It looks like the branch predates main's RunE/error-return refactor of `wisp.go` (the `runWispGC` signature changed from void + `FatalError` to `error` + `HandleError`/`RunE`), so a straight rebase will need the same guard logic re-applied to the current function shape - the logic itself doesn't need to change.

Minor, non-blocking note: `GetBlockedIssues(ctx, types.WorkFilter{})` fetches blocked issues database-wide rather than scoping to wisps/ephemerals; fine for correctness, just worth a follow-up if it becomes a hot path on large DBs.

I'll pull this to a local worktree, rebase onto current main, and push the fix-up so this can land - will keep your commits and attribution intact and note here when it's ready to merge.

maphew:
Reviewed this with two independent models. Both agreed the fix is correct for the case it targets; they disagreed on disposition, and working through that disagreement is what produced the commit I've just pushed to this branch.

The short version: your diagnosis is right and the fix is a strict improvement, so I absorbed the follow-ups rather than sending you a review round trip for them. Your commit is untouched.

**What I changed, and why**

The guard hand-writes the live-status list (`in_progress`, `blocked`, `hooked`, `pinned`). That covers the two dominant cases from #4394, but it leaves the same class of hole in three other places:

- `deferred` is reachable for wisps (`bd defer` has no ephemeral guard and the update layer routes to the `wisps` table automatically), and reclaiming something a user explicitly deferred defeats the point of deferring it.
- Custom statuses (`status.custom`) in wip/frozen categories are real workflow states and were unprotected.
- The `Issue.Pinned` flag is separate from the `pinned` status, and the closed-purge branch of this same command already honors it — so the age branch was internally inconsistent with its own sibling.

Rather than extend the list, I switched the predicate to the status *category*: `CategoryWIP` is work in flight, `CategoryFrozen` is work on ice, so only plain `open` (active) and `closed` (done) are age-reclaimable. Custom statuses then participate for free by their configured category. This closes all three gaps and is shorter than enumerating them.

The other change is that the blocked-set lookup now fails closed. It previously warned and continued with an empty set, which silently re-exposes every live blocked step to deletion — and `status` is not a usable fallback, because `is_blocked` is derived state independent of it, so a blocked-but-`status=open` step loses all protection. `purge.go` already has a comment stating the equivalent lookup there is "required, not best-effort"; this now matches that.

Also renamed `isActiveWisp` to `isProtectedWisp`. The old name collided with `(*DoltStore).isActiveWisp`, which is in this same delete path but means only "a row exists in the `wisps` table".

**On your tests** — they were genuinely non-tautological, which made this easy to build on. I extended them with deferred, pinned-status and custom wip-category cases, plus a custom *active*-category case that asserts the guard does not over-protect. Reverting the production change fails the deferred and custom-wip assertions, so the new coverage is load-bearing too.

One thing worth knowing if you write more tests here: I had to reorder the setup so that every bead expected to be *reclaimed* is created first. The age predicate compares against an `updated_at` written by the database, so the most recently touched bead can briefly appear not-yet-stale against a `1ms` threshold if the DB clock runs slightly ahead of the test process. That produced a real intermittent failure before I reordered it. Protected beads are excluded regardless of age, so only the reclaimable assertions are sensitive to this.

**Deliberately not done here:** the scan-then-delete window. Candidate status is read outside the deletion transaction, and in server mode there is no process-wide lock, so a wisp that becomes blocked after the scan can still be deleted. That is pre-existing — the `--age` predicate itself was never revalidated at delete time either — and it does not exist in the default embedded backend, which holds an exclusive flock. It is not yours to fix in this PR; it wants its own change.

I have not approved this myself, since I have now committed to the branch. Handing the merge tail to our patrol so it lands once checks are green.

PR #4449, opened 2026-06-18
bootstrap: guard git code-repo sync.remote + transient-not-found retry

bourgois:
## Problem

Two failure modes when `sync.remote` is misconfigured or a managed Dolt server restarts:

1. **Wrong URL type**: If `sync.remote` is set to a git code-repository URL (e.g. `git+ssh://github.com/org/repo.git`), `bd bootstrap` fires `DOLT_CLONE` against it. Dolt spins to ~1000% CPU and requires manual SIGKILL.

2. **Transient restart**: When a managed Dolt server restarts it briefly reports "database not found" for up to ~30 s. Without retry logic, `bd bootstrap` immediately falls through to sync/init, potentially triggering an unnecessary `DOLT_CLONE`.

### Layer 1 — `isGitCodeRepoURL()` guard in `detectBootstrapAction()`

Added `isGitCodeRepoURL(rawURL string) bool` (priority order):
- Dolt-native schemes (`dolthub://`, `s3://`, `gs://`, `az://`, `file://`) → **false** (always safe)
- `.git` suffix → **true** (Dolt DBs never use a `.git` suffix)
- Known code forges (`github.com`, `gitlab.com`, `bitbucket.org`, `codeberg.org`) or their subdomains → **true**
- Everything else → **false** (self-hosted `git+ssh://` Dolt remotes are valid)

When `isGitCodeRepoURL` returns true, `detectBootstrapAction` logs a structured error and returns `action=none`.

### Layer 2 — transient-not-found retry

When the server is reachable but the DB appears absent, retry 3× with exponential back-off (10 s, 20 s, 40 s) before falling through to sync/init. Spans 70 s total, covering all observed managed-Dolt restart windows (~30 s).

The delay is injected via `bootstrapRetryDelay` (`var` holding `func(time.Duration)`) so tests run at zero-sleep speed.

## Tests

New file `cmd/bd/bootstrap_guard_test.go`:
- `TestIsGitCodeRepoURL` — 21 cases: `.git` suffix, forge hosts, subdomains, Dolt-native schemes, self-hosted remotes, SCP-style, empty string
- `TestDetectBootstrapAction_GitCodeRepoSyncRemoteBlocked` — `action=none` for 4 git forge URLs
- `TestDetectBootstrapAction_ValidDoltSyncRemoteUnchanged` — no regression for 3 valid Dolt remotes
- `TestDetectBootstrapAction_ServerTransientRestart_DBFoundAfterRetry` — `action=none` + `HasExisting=true` when DB appears on 3rd probe
- `TestDetectBootstrapAction_ServerGenuinelyAbsent_FallsThrough` — falls through to `init` when DB is absent after all retries

Two existing server-mode tests stub `bootstrapRetryDelay` to avoid 70 s of real sleep.

## Test results

863 tests pass. 2 pre-existing failures in `TestIsBackupAutoEnabled` (unrelated, also fails on `main`).

### DIFF
--- a/cmd/bd/bootstrap.go
+++ b/cmd/bd/bootstrap.go
@@ -46,6 +46,10 @@ type bootstrapServerDBCheck struct {
 	Err       error
 }
 
+// bootstrapRetryDelay is the sleep function used between server-DB-not-found
+// retries. Injectable for testing.
+var bootstrapRetryDelay = func(d time.Duration) { time.Sleep(d) }
+
 var checkBootstrapServerDB = func(probeCfg bootstrapServerProbeConfig) bootstrapServerDBCheck {
 	host := probeCfg.host
 	port := probeCfg.port
@@ -284,6 +288,15 @@ func detectBootstrapAction(beadsDir string, cfg *configfile.Config) BootstrapPla
 	// Check sync.remote (primary) or sync.git-remote (deprecated fallback)
 	syncRemote := resolveSyncRemote()
 	if syncRemote != "" {
+		if isGitCodeRepoURL(syncRemote) {
+			// Cloning from a git code-repo URL via DOLT_CLONE spins dolt to
+			// 1000% CPU and requires manual SIGKILL (vc-8djyca). Reject and
+			// surface the misconfiguration rather than attempting the clone.
+			fmt.Fprintf(os.Stderr, "error: sync.remote %q looks like a git code-repository URL, not a Dolt remote — skipping clone\n", syncRemote)
+			plan.Action = "none"
+			plan.Reason = fmt.Sprintf("sync.remote %q rejected: git code-repository URL (not a Dolt remote)", syncRemote)
+			return plan
+		}
 		// User-provided sync.remote — trust the URL format as-is.
 		// normalizeRemoteURL would convert http:// to git+http://,
 		// breaking Dolt remotesapi endpoints (GH#3339).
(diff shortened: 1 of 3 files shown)


bourgois:
## Platform-architect review — changes required before merge

**Implementation verdict: correct.** L1 (`isGitCodeRepoURL` guard) and L2 (transient-not-found retry) both match the plan exactly. URL classifier priority order is right (Dolt-native schemes → `.git` suffix → known forges → allow-all-else). Injectable `bootstrapRetryDelay` is a clean testing pattern. 21-case URL classifier tests, four detection-path integration scenarios — thorough. The 2 pre-existing `TestIsBackupAutoEnabled` failures are confirmed pre-existing on `main`.

**One blocking finding before merge: Voxist-internal bead IDs in code comments.**

Six occurrences of `vc-8djyca` (a Voxist-internal incident tracking identifier) are embedded in code comments and will be permanently visible in upstream git history. Upstream contributors should not see internal incident IDs.

| File | Line | Fragment to remove |
|---|---|---|
| `bootstrap.go` | 293 | `(vc-8djyca)` |
| `bootstrap.go` | 399 | `(vc-8djyca)` |
| `bootstrap.go` | 1012 | `(vc-8djyca)` |
| `bootstrap_guard_test.go` | 14 | `(vc-8djyca)` |
| `bootstrap_guard_test.go` | 95 | `(vc-8djyca Layer 1)` |
| `bootstrap_guard_test.go` | 177 | `(vc-8djyca Layer 2)` |

**Fix:** Strip only those parentheticals. The surrounding text is accurate and should stay. No other changes required — once these six fragments are removed, this is ready to merge.

maphew (commented):
Thanks @bourgois - this addresses a real and nasty failure mode (DOLT_CLONE spinning to 1000% CPU against a git code-repo URL) and the URL-classifier guard (Layer 1) is a clean, well-tested fix worth landing.

A few things need to happen before merge, though:

1. **Internal IDs still need scrubbing.** Your own review comment (from 2026-06-18) already flagged six `vc-8djyca` fragments in `bootstrap.go` and `bootstrap_guard_test.go` as a blocking issue and said removing them was the only remaining step - but no follow-up commit has landed since, and the diff still contains all six. Could you push that cleanup?

2. **Layer 2 (the retry-with-backoff around a server DB-not-found probe) is the wrong layer for this project.** Our storage-boundary policy (`PROJECT_CHARTER.md`) is explicit that storage-specific retry loops and crash-recovery workarounds for Dolt server behavior belong in the Dolt driver, not in beads' bootstrap code - beads should only put data in and pull data out through the storage interface. The managed-Dolt-restart issue you're working around here is a real problem, but the fix belongs in `dolthub/driver` (or the interface needs to widen to expose a driver-level retry/backoff option), not as a beads-side sleep loop. There's also a concrete cost: in shared-server mode the probe's data dir is machine-global, so a first-ever bootstrap of a new project on a machine whose shared server already hosts other projects' databases passes the entries check and eats the full 10+20+40s backoff before falling through to init - your own `TestDetectBootstrapAction_ServerGenuinelyAbsent_FallsThrough` exercises exactly that fall-through path (with sleeps stubbed; real runs wait the full 70s).

Our plan: split this into two pieces. Layer 1 (the `isGitCodeRepoURL` guard, once the internal IDs are scrubbed) can land as-is here with attribution to you. Layer 2 should come out of this PR - please file it as a driver issue instead (or we can file it on your behalf and link back here) so the restart-tolerance behavior gets fixed at the right layer.

No action needed on the Layer 1 tests/logic itself - they look solid (21 URL-classifier cases plus 4 integration scenarios, no regressions to the existing valid-remote paths). Once the IDs are scrubbed we're happy to merge just that part; let us know if you'd like us to push a Layer-2-removal commit to your branch to speed this along.

maphew:
Filed the driver-side follow-up for Layer 2 here: dolthub/driver#453.

That keeps the managed-Dolt restart tolerance tracked at the storage boundary while this PR can focus on landing the Layer 1 URL guard after the internal IDs are scrubbed.

steveyegge:
Landed via #5400 (`8501a27c3` on main), with your commit preserved as an ancestor and authorship intact — the same org-fork transport pattern as #5335. The only deltas from your branch were the six review-requested internal-ID scrubs and two test-harness adaptations to code that moved since June (`snapshotBootstrapEnv` idiom + cgo build-tag gating). Both guard layers are now live. Thanks for the thorough work, @bourgois — the 21-case URL classifier suite was a pleasure to land.

PR #4428, opened 2026-06-16
fix(config): derive tracker config namespaces from the registry

harry-miller-trimble:
## Summary

`bd config set ado.org <org>` emitted a spurious warning even though ADO is a fully registered tracker and the value writes/reads correctly:

```
$ bd config set ado.org MyOrg
Warning: "ado.org" is not a recognized config key. Did you mean "ai.org"?
Run 'bd config --help' for valid namespaces.
Set ado.org = MyOrg
```

## Root cause

`recognizedConfigPrefixes` in `cmd/bd/config.go` hardcoded `jira.`/`linear.`/`github.` but never gained `ado.` (or `gitlab.`/`notion.`). The allowlist had drifted from the tracker registry — a recurring class of bug that bites every new tracker.

## Fix

Make the tracker registry the single source of truth instead of just patching in `ado.`:

- Drop the hardcoded tracker prefixes from `recognizedConfigPrefixes`.
- Derive them at runtime from `tracker.List()` via a new `allRecognizedConfigPrefixes()` helper, used by both `isRecognizedConfigKey` and `suggestConfigKey`.

This eliminates the drift class — the next registered tracker is covered automatically. **Bonus:** a genuine near-miss like `add.org` now suggests `ado.org` instead of the unhelpful `ai.org`.

Also adds the missing tracker namespaces (`gitlab`, `ado`, `notion`) to the `Common namespaces` help text.

## Behavior

```
$ bd config set ado.org MyOrg
Set ado.org = MyOrg                                    # no warning

$ bd config set add.org Oops
Warning: "add.org" is not a recognized config key. Did you mean "ado.org"?   # still catches typos, better suggestion
Set add.org = Oops
```

## Testing

- `TestIsRecognizedConfigKey` extended with `ado.org`/`ado.project` plus `github`/`linear`/`gitlab`/`notion` keys (guards the prefixes removed from the static list).
- New `TestTrackerConfigPrefixesRecognized` asserts **every** registered tracker's namespace is recognized — directly guards the drift class.
- `TestSuggestConfigKey` gains the `add.org -> ado.org` case.
- Verified end-to-end against an embedded workspace; `go build`/`gofmt`/`go vet` clean; `go test -run Config ./cmd/bd/` green.

Fixes #4427

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4453, opened 2026-06-19
fix(config): recognize ado.* as valid config namespace

### LINKED ISSUE #4427: bd config set ado.* warns "not a recognized config key" despite ADO being a registered tracker
## Summary

`bd config set ado.org <org>` (and `ado.project`, etc.) emits a spurious warning even though ADO is a fully registered tracker and the value is written/read correctly. The config-key recognizer's allowlist was never updated when the ADO integration was added.

## Repro

```
$ bd config set ado.org MyOrg
Warning: "ado.org" is not a recognized config key. Did you mean "ai.org"?
Run 'bd config --help' for valid namespaces.
Set ado.org = MyOrg

$ bd ado status
Organization: MyOrg          # <- value DID persist and is read back
...
Status: ✓ Configured
```

So the write succeeds and `bd ado status` reads it — the warning is purely cosmetic, but it's misleading (it implies the key is a typo / no-op) and the levenshtein suggester unhelpfully proposes `ai.org` (edit distance 2 from `ado`).

Observed on installed `bd version 1.0.3`; the bug is still present on `main` (`v1.0.5-128-ga59e75325`).

## Root cause

`cmd/bd/config.go` — `recognizedConfigPrefixes` hardcodes the tracker namespaces and includes `jira.`, `linear.`, `github.` but **not `ado.`**:

```go
var recognizedConfigPrefixes = []string{
	"export.", "import.", "dolt.", "jira.", "linear.", "github.", "custom.",
	"status.", "doctor.suppress.", "routing.", "sync.", "git.",
	"directory.", "repos.", "external_projects.", "validation.",
	"hierarchy.", "ai.", "backup.", "federation.",
}
```

Meanwhile ADO is a properly registered tracker — `internal/ado/tracker.go` calls `tracker.Register("ado", ...)` and `ConfigPrefix()` returns `"ado"`. The allowlist simply drifted from the tracker registry. (`bd config --help`'s "Common namespaces" list omits ADO for the same reason.)

## Suggested fix

Minimal, matches the existing idiom — add `"ado."` to `recognizedConfigPrefixes` (and ideally to the `Common namespaces` help text in `configCmd.Long`):

```go
"export.", "import.", "dolt.", "ado.", "jira.", "linear.", "github.", "custom.",
```

`cmd/bd/config_validate_key_test.go`'s `TestIsRecognizedConfigKey` covers `jira.url` but no ADO key — add `ado.org` / `ado.project` to its recognized list to guard against regression.

### Optional, more robust

There is a `tracker.List()` registry (`internal/tracker/registry.go`). Deriving the tracker prefixes from registered trackers instead of hardcoding them would prevent the *next* tracker from hitting this same drift. Larger surface, but eliminates the class of bug.

## Verified

Applied the one-line fix + test case locally and `go test ./cmd/bd/` passes (build needs ICU CGO flags on macOS: `icu4c` via brew + `CGO_CXXFLAGS`/`CGO_LDFLAGS`).

jfjrh2014:
## Summary

`bd config set ado.org <value>` (and `ado.project`, `ado.projects`, etc.) emits a spurious warning:

```
$ bd config set ado.org MyOrg
Warning: "ado.org" is not a recognized config key. Did you mean "ai.org"?
```

The value is actually written and read correctly by the ADO integration (cmd/bd/ado.go), but the config-key allowlist in config.go was never updated when ADO support was added.

## Fix

Add `"ado."` to `recognizedConfigPrefixes` in cmd/bd/config.go. This is a one-line change that mirrors how `jira.`, `linear.`, and `github.` are already handled.

## Test

Added `ado.org`, `ado.project`, `ado.projects` to the recognized-key list in `TestIsRecognizedConfigKey`.

Fixes #4427

### DIFF
--- a/cmd/bd/config.go
+++ b/cmd/bd/config.go
@@ -812,7 +812,7 @@ var recognizedConfigPrefixes = []string{
 	"export.", "import.", "dolt.", "jira.", "linear.", "github.", "custom.",
 	"status.", "doctor.suppress.", "routing.", "sync.", "git.",
 	"directory.", "repos.", "external_projects.", "validation.",
-	"hierarchy.", "ai.", "backup.", "federation.",
+	"hierarchy.", "ai.", "backup.", "federation.", "ado.",
 }
 
 // recognizedConfigKeys lists valid non-namespaced config keys.
--- a/cmd/bd/config_validate_key_test.go
+++ b/cmd/bd/config_validate_key_test.go
@@ -10,7 +10,7 @@ func TestIsRecognizedConfigKey(t *testing.T) {
 		"export.auto", "dolt.auto-push", "jira.url", "custom.anything",
 		"doctor.suppress.git-hooks", "no-git-ops", "beads.role",
 		"status.custom", "ai.model", "backup.enabled", "import.path",
-		"dolt.local-only",
+		"dolt.local-only", "ado.org", "ado.project", "ado.projects",
 	}
 	for _, key := range recognized {
 		if !isRecognizedConfigKey(key) {


maphew (commented):
Thanks for tracking down this bug and for the clean, well-tested fix - the repro and root cause writeup made this easy to verify.

While reviewing, I found that #4428 (opened a few days before this PR, by a different contributor) fixes the identical root cause (#4427) but goes one step further: instead of hardcoding `ado.` into the recognized-prefix allowlist, it derives all tracker namespaces from the tracker registry at runtime. That removes the whole class of drift this bug came from - the same allowlist was also missing `gitlab.` and `notion.`, which #4428 catches too - and adds a dedicated regression test (`TestTrackerConfigPrefixesRecognized`) that asserts every registered tracker's namespace is recognized, plus doc regeneration for the new namespaces in the help text.

#4428 is currently mergeable; this branch has since drifted into a merge conflict with `main`.

Given the overlap, the recommendation is to retire this PR in favor of #4428, since it delivers the same fix for #4427 in a form that also prevents recurrence for future trackers. Thank you again for finding and diagnosing the bug clearly - that description is genuinely what made triage straightforward. If #4428 stalls or is withdrawn for any reason, I'll come back to this one and help get it rebased and merged.

maphew:
Closing as superseded: gastownhall/beads#4428 (merged 2026-07-08) resolved this class of problem structurally by deriving valid tracker config namespaces from the tracker registry itself, so `ado.*` — and any future tracker's namespace — is recognized without a hand-maintained allowlist. That landed after this PR was opened, and it covers the case this PR fixed.

Thank you for the report and the fix — the PR correctly identified the gap, and the registry-derived approach is where the fix ended up living. If you spot any `ado.*` key that is still rejected on current `main`, please open an issue and we'll treat it as a bug in the registry derivation.

PR #4011, opened 2026-05-17
test: prevent Cobra command flag cache races

quad341:
Cherry-picks d8a97c6bb (originally authored by Julian Knutsen) as a
standalone PR on main. The fix removes t.Parallel() from
TestStaleCommandInit and TestNotionCommandsRegistered, and extends
the cobra parallel-policy guard to flag .Find( and .InheritedFlags(.

Root cause: both methods lazily merge persistent parent flags into a
shared pflag cache, racing under -race when the two tests run in
parallel workers. Race trace example (PR #4009 macOS-latest job):
pflag.(*FlagSet).AddFlag write vs. updateParentsPflags read on the
same FlagSet.

Currently sitting on fix/ready-wisp-filter-parity and runtime/perf-*
branches without its own PR — landing this on main fixes the flake
for every open PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4461, opened 2026-06-20
Add append-only provenance_events log and bd provenance CLI

### LINKED ISSUE #4460: Feature: append-only provenance event log (provenance_events) for agent-driven work
## What

Add a small, generic primitive to core bd: an append-only `provenance_events` table that binds causal facts about how a bead was worked — what base commit it started from, which actor touched it when, what commit/PR landed it, what other work it drew on — to the `work_id`, plus a `bd provenance` command group to record and query them.

It's **inert by default**: nothing writes it unless a producer (an orchestrator, a git hook, a CI webhook) opts in, so it adds a table and a command and changes no existing behavior.

```
bd provenance record --issue <id> --kind <cut|claim|commit|land|used|…> --source <src> \
       [--actor <id>] [--ref <sha|pr|work_id>] [--ref-kind <k>] [--payload <json>] [--at <iso>]
bd provenance log <issue-id> [--kind <k>]
bd provenance by-ref <ref>
```

There is deliberately no `update`/`delete` verb — append-only is enforced by the absence of a mutating command. Corrections are new rows.

## Why

Tools that drive agents against bd (orchestrators, CI, retrieval/memory systems) all need to answer "how did this work actually happen?" Today that's reconstructed after the fact from side-channels — git history by date, commit-message trailers, transcript matching — and each reconstruction is lossy. Some facts are *destroyed* at work-time and unrecoverable later:

- the **exact base commit** a worktree was cut from (once the branch advances, it's gone — you can only approximate it by date),
- **which of several concurrent sessions** produced a given commit (no per-commit actor stamp → time-window guessing),
- the ordered **interrupt/resume/handoff** history (the assignee field only keeps the end state).

These are cheap to record at the one moment they're true and impossible to recover otherwise. A provenance log is the natural home, and it's generically useful to any agent-driving consumer, not specific to one.

## Design sketch (full rationale would go in a proposed ADR-0003)

- **Sibling table, not an extension of `events`.** The `0005` `events` table is a two-value field-mutation audit (`old_value`/`new_value`/`comment`); a provenance fact is a typed binding to a structured external artifact (a `land` is commit + PR + CI). This follows bd's own precedent of purpose-specific event tables (`wisp_events`, `0031`) rather than overloading the central audit log, and keeps the hot audit-write path untouched.
- **Opaque identifiers.** `actor`/`ref` are opaque strings bd doesn't interpret — no session-UUID or orchestrator-specific id semantics in core bd. The runtime defines identity; bd only binds it. This is what keeps the primitive usable without any particular orchestrator.
- **Append-only**, FK→`issues` CASCADE, indexed on `(issue_id, occurred_at)`, `ref`, and `kind`. One deliberate deviation from the `events` idiom: a separate `occurred_at` (event-time) alongside `created_at` (ingest-time), because hooks may record a fact after it happened.
- bd validates only **structure** (known `kind`, allowed `ref_kind`) — never what a SHA *means* (ZFC).

## Scope / questions for maintainers

1. Does a provenance primitive belong in core bd, or would you prefer it live as an extension/consumer against the dolt schema?
2. Naming: `bd provenance` (avoids colliding with the `events`/audit surface) vs. folding into an existing command — preference?
3. Migration `0040` + `bd provenance` CLI + store methods + `//go:build cgo` tests + ADR-0003 as one focused PR — does that match your one-issue-per-PR expectation, or would you split the schema and CLI?

Happy to bring a full PR; opening this first to confirm the primitive is wanted in core before writing it.

sjarmak:
## What

Implements the append-only provenance primitive proposed in #4460: a `provenance_events` table that binds causal facts about how a bead was worked — the base commit a worktree forked from, which actor touched it when, what landed it — to an issue, plus a `bd provenance record|log|by-ref` command group.

Closes #4460.

## Why

Tools that drive agents against beads need to answer "how did this work actually happen?" Today that's reconstructed after the fact (git history by date, commit trailers, transcript matching) and each reconstruction is lossy — some facts (the exact fork SHA, per-commit attribution) are *destroyed* at work-time and unrecoverable later. They're cheap to record at the moment they're true. This is the generic, inert-by-default home for them; nothing writes the table unless a producer (git hook, orchestrator, CI) opts in.

## Design

- **Sibling table, not an extension of `events`.** `events` is a two-value field-mutation audit; a provenance fact is a typed binding to a structured external artifact (a `land` is commit + PR + CI). Follows the `wisp_events` precedent for purpose-specific event tables; leaves the audit write-path untouched.
- **Append-only at the event level** — no UPDATE/DELETE operation on individual events. An event's lifecycle is bound to its issue (`ON DELETE CASCADE ON UPDATE CASCADE`), identical to the `events` audit table.
- **Opaque `actor`/`ref`** — bd never interprets them; only `kind` and `ref_kind` are structurally validated (ZFC), so the primitive is usable by any runtime without baking in orchestrator/session semantics.
- **`occurred_at` (event-time) separate from `created_at`** — a producer may record a fact after it happened.
- **Idempotent record** via a content-addressed deterministic id (`source:issue:kind:(ref or occurred_at)`, second-truncated to match the `DATETIME` column), so a producer firing twice is a no-op (`INSERT IGNORE`). An event with neither a ref nor an `occurred_at` is rejected at the store boundary (no stable id otherwise). A reserved `ingest-backfill` source is rejected so a producer can't forge a derived event.
- Migration `0052`; shared logic in `issueops`, both Dolt stores delegate (mirrors the `events` pattern). ADR-0003 records the rationale, including why not extend `events`, why opaque ids, and the distributed-merge note.

## Test plan

- [x] `make build`, `go vet ./...`, `gofmt`, `make check-docs` clean
- [x] CGO tests (`./cmd/bd/ -run Provenance`): record→log→by-ref round trip; idempotent re-record (incl. sub-second `occurred_at`); reject unknown kind / reserved source (case-insensitive) / non-40-hex git-sha / ref-kind-without-ref / ref-less-without-`--at`; opaque fields round-trip exactly
- [x] migration applies via the schema runner
- [ ] maintainer review

## Notes

- `ref` is `VARCHAR(255)` (directly indexable) — fits a SHA/PR-url; very long refs aren't indexed.
- Provenance events are not yet in the merge auto-resolve allowlist (a `bd pull` with concurrent inserts from two clones leaves the conflict for the operator); safe to auto-resolve later (append-only + deterministic id + `INSERT IGNORE` is commutative), deferred as a separate change.


maphew (commented):
Thanks @sjarmak - this is an unusually thorough contribution. The design does real work to stay a primitive (opaque `actor`/`ref`, closed structural sets, inert by default, content-addressed idempotent ids), the ADR honestly records alternatives including the distributed-merge deferral, and the CGO tests are meaningful: round trip, idempotency including sub-second `--at`, NULL round-trip, and every validation guard. The storage plumbing correctly mirrors the `events` pattern across both Dolt stores, telemetry, and the jira test stub.

All the CI red plus the merge conflict trace to three mechanical issues:

1. **`DEFAULT (UUID())` must go** (`0052_create_provenance_events.up.sql` line 21). This is what fails the migration-hygiene check, and it directly contradicts migration 0051 (`drop_aux_id_defaults`), which just removed exactly this default from every aux table because of the #4259 per-clone-divergence corruption class. Since `RecordProvenanceEventInTx` always mints the id app-side, the default is dead weight - and 0051's rationale is that a future path omitting the id should fail loudly with a NOT NULL violation rather than silently mint per-clone-random keys. One-line fix.

2. **Renumber the migration to 0054.** Main has since landed `0052_add_date_indexes` and `0053_repair_rig_wisps`, which is also why the PR shows conflicting. Please update the "migration 0052" references in `issueops/provenance.go` and ADR-0003 to match.

3. **Regenerate docs; revert the versioned_docs edits.** The doc-freshness jobs fail because the committed generated docs are stale. Note that `scripts/generate-cli-docs.sh` only writes `website/versioned_docs/` via `--versioned` at release-snapshot time - the hand-added `provenance.md` under `version-1.0.0`, `version-1.0.4`, and `version-1.0.5` documents the command in releases that don't have it, and `llms-full.txt` derives from the latest released snapshot. Revert those, rebase, and rerun the generator; the current-docs output (`docs/CLI_REFERENCE.md`, `website/docs/cli-reference/`) is all that should change.

Two smaller notes, neither blocking:

- `ProvenanceEventID` joins `source:issue:kind:disc` with `:` unescaped. `source` is arbitrary and refs are often URLs, so two distinct events can theoretically collide onto one id and the second gets silently `INSERT IGNORE`d. Length-prefixing or hashing each field separately would make the id basis unambiguous - fine as a follow-up.
- `provKind` is shared between `record --kind` and `log --kind`; separate vars would be more robust given the flag-state races #4011 dealt with.

On scope: this is the one genuine judgment call. The charter prefers metadata before schema for workflow-specific data, and the kind vocabulary here is workflow-flavored (the #4460 thread is already proposing more kinds). That said, the things this table provides - indexed `by-ref` lookup, an append-only invariant, deterministic idempotent ids, FK cascade - are exactly what issue metadata cannot, and `events`/`wisp_events` set the precedent for purpose-specific event tables. We'll confirm the schema-vs-metadata call on #4460, but the design has earned its case.

**Recommendation:** the three blockers above are mechanical, so we're happy to apply them on this branch if "Allow edits by maintainers" is enabled - otherwise a rebase plus those three fixes from your side should turn CI green. Either way we'd like to land this.

sjarmak:
Thanks for the careful review, @maphew. I've rebased onto current main and addressed all three blockers. Summary of what changed:

**1. Dropped `DEFAULT (UUID())` on `provenance_events.id`.** The column stays `CHAR(36) NOT NULL PRIMARY KEY`; only the default is gone. `RecordProvenanceEventInTx` already mints the id app-side from the content-addressed basis, so the default was dead weight, and as you noted it both fails the hygiene check (Check B) and reintroduces exactly the per-clone-divergence class that 0051 removed from the other aux tables (#4259). A future path that omits the id now fails loudly on the NOT NULL constraint instead of silently minting a per-clone-random key.

**2. Renumbered the migration.** One wrinkle: since your review, main has landed `0054_add_lease_columns` on top of `0052_add_date_indexes` and `0053_repair_rig_wisps`. Renumbering to 0054 would collide with the lease-columns migration and trip the duplicate-version guard (Check A), so I took the next free number and made it **0055** rather than 0054. Both files are renamed (`0055_create_provenance_events.{up,down}.sql`), and the "migration 0052" references in `internal/storage/issueops/provenance.go` and `docs/adr/0003-provenance-event-log.md` are updated to 0055. Grep confirms no other file references the old number for this migration (the remaining `0052` hit in CHANGELOG.md is the date-indexes migration, which is correct). Happy to shift the number again if something else lands before this merges.

**3. Regenerated docs; reverted the versioned snapshots.** Reverted the hand-added `provenance.md` and the `index.md` edits under `versioned_docs/version-1.0.0`, `1.0.4`, and `1.0.5`, plus the `llms-full.txt` changes, back to their pre-PR state, since those releases don't have the command. Then re-ran `scripts/generate-cli-docs.sh` (no `--versioned`) so only the current-docs output is regenerated. `git diff origin/main..HEAD` touches no `versioned_docs` or `llms-full.txt` files, and `scripts/check-doc-freshness.sh` reports all four doc gates green, including "llms-full.txt is fresh" and "generated CLI docs are fresh".

One more thing surfaced by the rebase: main's `5d87c3001` migrated CLI commands off `FatalError*` onto `HandleError*` with `RunE`, so the provenance command didn't compile as-is. I converted the three subcommands to `RunE`/`HandleErrorRespectJSON` to match the current convention; exit-1 semantics are preserved.

On your non-blocking notes: I took the `provKind` one, since it was a clean two-variable split. `provenance log --kind` now binds its own `provLogKind` var instead of sharing `provKind` with `record --kind`, so the two flags can't race a shared package var across runs, which is the failure mode #4011 dealt with. The ID `:`-escaping I've left as the tracked follow-up we agreed on.

Gates run locally, all green: `go build ./...`, `go vet` on the touched packages (CGO on and off), the migration-hygiene check (`BASE_SHA=origin/main`), the doc-flag and doc-freshness checks, and the CGO provenance test suite (11 tests, including the round-trip, idempotency, sub-second `--at` idempotency, opaque-field round-trip, and the validation guards) plus the `issueops` and `jira` package tests.

maphew (approved):
@sjarmak — you fixed all three blockers on 2026-07-08, said so precisely, and then waited 18 days while main moved 500+ commits under you. That's on us, so we did the drift-repair maintainer-side rather than sending you a fourth round of asks. Your two commits are intact with your authorship; one maintainer commit (304c2c8d6) on top contains only rebase-tail mechanics:

- **Migration renumbered 0055 → 0060** (upstream took 0055–0059 in the interim) with both prose refs updated; `check-migration-hygiene.sh` passes.
- **Generated doc artifacts dropped, not regenerated** — since #4876 the CLI docs are pinned to release v1.1.0 (`docs/cli-docs.pin`) and new-command docs land at the next pin bump (same as `bd sync` and `bd conflicts`, which merged doc-less). Your `bd provenance` docs will appear with the next release regen.
- **ADR moved to `engdocs/adr/`** (where ADRs live now) with a `DOC_INVENTORY.md` row, plus one added paragraph documenting that wisp demotion drops provenance via the same delete-cascade as issue deletion — an accepted consequence, no wisp counterpart table.
- **`provenance_events` added to the Dolt staging lists** (`issues.go` delete paths + `permanentIssueAuxTables`) so cascaded deletions don't sit unstaged in the working set in server mode.
- Rebase conflict in `embeddeddolt/store.go` resolved by keeping upstream's new `EventsSince` and your three methods side by side.

Verification at the new head: vet clean; all 11 provenance tests pass; issueops suite passes; binary builds. The scope question is also finally answered — maintainer position posted on #4460 (schema over metadata: accepted; minimal-and-opaque over rich schema: accepted for v1). The ADR reads `Proposed` in this branch and flips to Accepted when this merges.

Deferred to a follow-up (tracked on our side): union-merge auto-settle for `provenance_events`, `--kind` validation on the log path, and the ID colon-escaping we'd already agreed to punt.

Thanks for the patience and the discipline of the fix round — this is how we wish every PR response looked. Handing to our merge automation now; it merges when checks pass on the new head.

steveyegge:
Drift re-repair round 2, maintainer-side (the 7/27 approval stands; merge automation was disabled repo-wide on 7/31, which is why this sat — sorry again for the wait, @sjarmak). The branch is rebased onto current main with all three commits' authorship preserved verbatim; the only rewording is the fix-merge subject (0060 → 0063, matching the renumber).

What moved under it this time:
- **Migration renumbered 0060 → 0063** (main took 0060–0062 in the interim); both prose refs updated; `check-migration-hygiene.sh` green.
- **Two new staging surfaces on main since 7/27**, both now wired with `provenance_events`: `DeleteCascadeTables` in `issueops/wisp_routing.go` (main grew a ratchet test deriving CASCADE targets from shipped migrations — it catches the omission, and passes now) and `sweptTables` in `dolt/sweeper.go` (feeds both the sweeper and the new batched deleter).
- `deleteIssue`/`deleteIssues` staging loops were refactored on main into `doltAddAndCommitInTx` table lists; the table rides those lists now.

Verified at the new head: build + vet clean, migration hygiene green, all 11 provenance tests green, issueops suite green (including the cascade ratchet). The dolt-server integration suites need testcontainers and are CI's to arbitrate — runs approved on the new head.

**One design note for the record** (doesn't block this merge): since the approval, migration 0062 moved `events` to the dolt-ignored plane, so the two audit tables now sit on different versioning/replication planes, and `RecordProvenanceEvent`'s commit-per-event is the history-churn class 0062 eliminated. Acceptable at introduction tempo for an inert-by-default primitive; tracked as **bd-gw6m0** to decide (keep versioned vs. follow `events`) before provenance runs at fleet tempo.

Merging when checks pass on this head.

steveyegge:
Merged as c0f18bc03. Closing the loop on the long arc here: approved 7/27, then two rounds of maintainer-side drift repair to keep the branch current against main's churn (backend simplification, then the facade waves) — the final green run is on the exact head that merged (ad0867995).

Thanks @sjarmak for the provenance plane and for staying with it across the rebases. The versioning-plane follow-up discussed in review is tracked maintainer-side as bd-gw6m0 and will be decided before fleet-tempo adoption rather than blocking this landing.

_claude-fable-5 on behalf of bee (beads-lane steward)_

PR #4432, opened 2026-06-16
fix(import): honor imported dependency created_by

orlandobrent:
## What

JSONL import now honors `created_by` on imported dependency edges when the dependency object includes it.

## Why

JSONL import already parsed dependency `created_by`, but `PersistDependencies` stamped new dependency rows with the current actor. This caused imported dependency attribution to be lost. Dependencies without imported `created_by` still fall back to the current actor, preserving interactive creation behavior.

## Verification

- `go test ./internal/storage/issueops -run 'TestPersistDependencies(HonorsImportedCreatedBy|DefaultsCreatedByToActor)'`
- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestImportFromLocalJSONL/preserves_dependency_created_by_from_JSONL_import'`

PR #4472, opened 2026-06-22
fix(create): reject cross-table ID collisions between issues and wisps (#4455)

### LINKED ISSUE #4455: bd create allows cross-table ID collision (same --id as an issue and a --no-history wisp), bricking bd ready/search
## Summary

`bd create` will happily put the **same ID in both the `issues` and `wisps` tables**. There is no cross-table uniqueness check at create time, so a normal issue and a `--no-history`/`--ephemeral` wisp created with the same explicit `--id` both succeed. The resulting dual-presence row then makes `bd ready` and `bd search` hard-error for the *entire* store.

This is the same dual-presence failure mode tracked as `be-iabdi` and addressed for the **lookup** side by #4163. #4163 makes lookups tolerant of the dup (prefers the wisp) and adds a `bd doctor` cleanup — but it does **not** prevent the dup from being created. This issue tracks the remaining root-cause/prevention gap.

## Reproduction

```bash
git init repro && cd repro
bd init -p prev --non-interactive

# 1. A normal issue -> issues table
bd create "perm copy" --id prev-aaa --type task          # ✓ Created  (rc=0)

# 2. A no-history wisp with the SAME id -> wisps table
bd create "wisp copy" --id prev-aaa --no-history --type task   # ✓ Created  (rc=0)  <-- no guard

# 3. Lookups that merge issues+wisps now break for the whole store
bd ready          # Error: ready work id "prev-aaa" exists in both issues and wisps   (rc=1)
bd search prev    # Error: id "prev-aaa" exists in both issues and wisps              (rc=1)
```

(`--ephemeral` instead of `--no-history` reaches the wisps table the same way.)

## Expected

The second `bd create` should refuse to create a row whose ID already exists in the other table (or otherwise prevent the cross-table collision), so a single store can never reach the dual-presence state through normal CLI use.

## Actual

Both creates succeed silently. `bd ready` and `bd ready --json` and `bd search <term>` then fail with `id "<id>" exists in both issues and wisps`. (`bd list`/`bd blocked` happen to survive — they use a query path that doesn't trip the merge guard — which makes the breakage inconsistent and easy to misdiagnose.)

## Root cause

`InsertIssueIfNew` (`internal/storage/issueops/create.go`) checks for an existing ID **only in the target table**:

```go
SELECT COUNT(*) FROM <issueTable> WHERE id = ?   // issueTable is "issues" OR "wisps", never both
```

There is no check against the sibling table, and nothing else in the create path enforces cross-table ID uniqueness. A natural fix would be to reject (or dedup) a create whose ID already exists in the other table, inside `InsertIssueIfNew` / the batch-create path.

## Relationship to #4163

- #4163 = lookup robustness + `bd doctor` repair (treats the symptom; recommended to merge).
- This issue = prevent the dual-presence from being created in the first place (closes the hole).

## Environment

- Reproduced on `main` @ `1d4dbb5ff` (`bd 1.0.5` dev build), embedded (single-DB) store.
- No migration or external sync required — plain `bd create` reaches it.

---
*Filed via Claude Code while triaging open PRs (root-cause companion to #4163).*

harry-miller-trimble:
Fixes #4455.

## Problem

`bd create` would put the **same ID in both the `issues` and `wisps` tables**. A normal issue and a `--no-history`/`--ephemeral` wisp created with the same explicit `--id` both succeed silently:

```
$ bd create "perm copy" --id prev-aaa --type task                 # ✓ rc=0
$ bd create "wisp copy" --id prev-aaa --no-history --type task     # ✓ rc=0  <-- no guard
```

The resulting dual-presence row makes the merge-based lookups (`bd ready`, `bd search`) hard-error for the entire store. #4163 made those lookups *tolerant* of an already-corrupted store (and added a `bd doctor` repair) but did **not** prevent the dup from being created. This PR closes the prevention gap.

## Root cause

The existence check in `InsertIssueIfNew` only ever queries the **target** table (`issues` OR `wisps`, never both), and nothing else in the create path enforces cross-table ID uniqueness.

## Fix

Add `checkCrossTableIDCollision` to the create path (`CreateIssueInTxWithResult`): a create whose ID already lives in the sibling table is refused with a clear error.

```
$ bd create "wisp copy" --id prev-aaa --no-history --type task
Error: cannot create "prev-aaa": ID already exists in the issues table (issues and wisps share one ID space)
```

Two paths are deliberately carved out:

- **Promotion** (`PromoteFromEphemeralInTx`) inserts into `issues` while the wisp row still exists, then deletes the wisp. It calls `InsertIssueIfNew` directly and bypasses `CreateIssueInTxWithResult`, so its transient dual-presence window is unaffected.
- **Auto-import upgrade-recovery** (`ConflictSkip`, GH#3955) must never hard-fail on startup, so it *skips* the colliding row instead of erroring (lookups stay tolerant via #4163).

Guard runs in both collision directions and reuses the existing single-transaction create path, so it works in embedded mode and for sequential `bd create` invocations in server mode.

## Testing

Added `internal/storage/embeddeddolt/cross_table_id_collision_test.go` covering:
- issue-then-wisp and wisp-then-issue collisions are both rejected and write nothing;
- same-table re-create still upserts (no regression);
- promotion of an ephemeral wisp still succeeds with the guard in place.

Verified the test **fails without the fix** and passes with it. Full `internal/storage/issueops` and `internal/storage/embeddeddolt` suites pass (`BEADS_TEST_EMBEDDED_DOLT=1`, `-tags gms_pure_go`).

### DIFF
--- a/internal/storage/embeddeddolt/cross_table_id_collision_test.go
+++ b/internal/storage/embeddeddolt/cross_table_id_collision_test.go
@@ -1,0 +1,114 @@
+//go:build cgo
+
+package embeddeddolt_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// TestCreateRejectsCrossTableIDCollision is the regression test for GH#4455:
+// `bd create` must never put the same ID in both the issues and wisps tables.
+// Issues and wisps share one ID space but live in separate tables, so a dual
+// presence row makes merge-based lookups (bd ready/search) hard-error for the
+// whole store. The fix rejects a create whose ID already exists in the sibling
+// table; #4163 only made the lookups tolerant of an already-corrupted store.
+func TestCreateRejectsCrossTableIDCollision(t *testing.T) {
+	skipUnlessEmbeddedDolt(t)
+
+	ctx := t.Context()
+
+	// Direction 1: permanent issue first, then a wisp with the same ID.
+	t.Run("issue then wisp", func(t *testing.T) {
+		te := newTestEnv(t, "ct")
+		if err := te.store.CreateIssue(ctx, &types.Issue{
+			ID: "ct-aaa", Title: "perm", Status: types.StatusOpen, Priority: 2, IssueType: types.TypeTask,
+		}, "tester"); err != nil {
+			t.Fatalf("create permanent issue: %v", err)
+		}
+		err := te.store.CreateIssue(ctx, &types.Issue{
+			ID: "ct-aaa", Title: "wisp", Status: types.StatusOpen, Priority: 2, IssueType: types.TypeTask, Ephemeral: true,
+		}, "tester")
+		if err == nil {
+			t.Fatal("expected cross-table collision to be rejected, got nil error")
+		}
+		if !strings.Contains(err.Error(), "already exists in the issues table") {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		// The wisp must not have been written.
+		var wispCount int
+		te.queryScalar(t, ctx, "SELECT COUNT(*) FROM wisps WHERE id = ?", []any{"ct-aaa"}, &wispCount)
+		if wispCount != 0 {
+			t.Fatalf("colliding wisp was written: wisps rows = %d, want 0", wispCount)
+		}
+	})
+
+	// Direction 2: wisp first, then a permanent issue with the same ID.
+	t.Run("wisp then issue", func(t *testing.T) {
+		te := newTestEnv(t, "ct")
+		if err := te.store.CreateIssue(ctx, &types.Issue{
+			ID: "ct-bbb", Title: "wisp", Status: types.StatusOpen, Priority: 2, IssueType: types.TypeTask, Ephemeral: true,
+		}, "tester"); err != nil {
+			t.Fatalf("create wisp: %v", err)
+		}
+		err := te.store.CreateIssue(ctx, &types.Issue{
+			ID: "ct-bbb", Title: "perm", Status: types.StatusOpen, Priority: 2, IssueType: types.TypeTask,
+		}, "tester")
+		if err == nil {
+			t.Fatal("expected cross-table collision to be rejected, got nil error")
+		}
+		if !strings.Contains(err.Error(), "already exists in the wisps table") {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		var issueCount int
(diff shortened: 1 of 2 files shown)


maphew:
Thanks @harry-miller-trimble - this closes the prevention half of #4455 exactly where it needs closing, and the write-up is a model PR description: root cause, the two deliberate carve-outs (promotion's transient dual-presence window, ConflictSkip's never-hard-fail recovery path), and a regression test verified to fail without the fix.

I checked current main before reviewing: there is still no create-time cross-table guard there (#4163 only made lookups tolerant and added the `bd doctor` repair), #4455 remains open, and no other PR targets it - so this is live and wanted, not superseded. The base is ~139 commits behind but merges cleanly; the only change to `create.go` on main since your base is the unrelated `dependencyCreatedBy` fix (#4432) in a different function. CI is green across the board.

On the layering question we weighed: create-time rejection is the right layer. Dolt can't express a cross-table unique constraint, so the create transaction is the only enforcement point for the shared ID space, and it composes cleanly with the merged tolerance (#4163) and repair into prevention/tolerance/repair defense-in-depth. The resolver work in #4376 is orthogonal (resolution ambiguity, not dual-presence creation). I also confirmed `InsertIssueIfNew` has exactly two callers - this guarded path and the documented promotion carve-out - so nothing else slips past the guard.

Three small observations, none blocking:

1. **Regular imports now fail loudly on a cross-table dup** (`internal/storage/issueops/create.go`, the non-ConflictSkip error return). Plain `bd import` runs with `ConflictSkip=false` (UPSERT semantics), so restoring a pre-promotion JSONL snapshot onto a store where that wisp was since promoted now aborts the import instead of silently writing the corrupt row. Erroring beats corrupting, so I read this as an improvement - flagging it so the behavior change on the restore path is a conscious acceptance rather than a surprise.
2. **Auto-generated IDs that collide cross-table error rather than regenerate.** `GenerateIssueIDInTable` checks only the target table, and promoted wisps leave `<prefix>-wisp-*` IDs in `issues`, so a rare hash collision would surface your new error instead of trying another nonce. Vanishingly unlikely and strictly better than the status quo; a follow-up could fold a sibling check into the generator's collision loop.
3. **The acknowledged concurrent-create race** (read-then-write, no DB constraint) is a residual the PR body already scopes honestly; tolerance + `bd doctor` cover that leak.

Recommendation: **merge**. This is a small, well-tested fix for a release-blocking corruption path, with both collision directions, the same-table upsert non-regression, and the promotion carve-out locked in by tests.

PR #4473, opened 2026-06-22
feat(dolt): query-level connection pooler to collapse Dolt connection churn

rbriski:
Core is built, tested, and proven end-to-end on a live town. An extended soak is running on a live workspace; see the testing comment below for full results.

## Problem
`bd`/`gt` use a connection-per-invocation model: every short-lived `bd` process opens its own Dolt sql-server connection. Under multi-process patrol/agent load this storms the listener's accept loop until it wedges ("alive but unresponsive"). Measured churn on a live town: hundreds of new Dolt connections/min, while concurrent connections stay tiny (sub-second open/close).

## Fix
A query-level MySQL connection **pooler** that fronts Dolt: it holds K persistent backend connections and multiplexes the many short-lived `bd` clients onto them, so **Dolt sees ~K connections regardless of client rate.**

- **`internal/storage/dbproxy/pool/`** — the pooler core: vitess `go/mysql` `Listener` + `Handler`, a persistent backend pool with keepalive + reconnect-on-borrow (Dolt idle-reaps loopback conns ~40s), bounded borrow-wait, and `COM_RESET_CONNECTION` between clients (proven to fully reset Dolt session state — branch/vars/autocommit/DB — so backends are safely reused with zero bleed).
- **`proxy/pooled_server.go`** — reuses the existing `dbproxy` lifecycle (lock/pidfile/log/idle-shutdown, signal handling) for a new pooled serving mode; the byte-passthrough proxy is untouched.
- **`internal/storage/dolt/pool_activation.go`** — `maybeActivatePooler`, hooked at the top of `newServerMode`, spawns one **shared, long-lived pooler per Dolt endpoint** and routes the server-mode store connection through its socket. **OFF by default** (opt-in via `BEADS_DOLT_POOL`), strict no-op otherwise, falls back to a direct connection on spawn failure.

### Prepared statements
`bd`'s server-mode DSN does not set `interpolateParams`, so go-sql-driver sends server-side `COM_STMT_PREPARE/EXECUTE`. The pooler implements these by interpolating bound params back into SQL via vitess `GenerateQuery` (SQL-encoded/escaped) and running plain text — so prepared statements work through the pool without a server-side prepare.

## Enabling
`BEADS_DOLT_POOL=1` (`BEADS_DOLT_POOL_SIZE=16`, `BEADS_DOLT_POOL_SOCKET=...` optional). See `internal/storage/dbproxy/pool/doc.go`.

## Evidence (live town Dolt)
- **Churn A/B, normal `bd` invocations:** plain ×60 → ~120–183 new Dolt connections; pooled ×60 → **~1–5.** ~30–180× collapse.
- **Shared + long-lived:** `bd` from multiple rigs uses the same pooler PID; one child process; per-rig DB isolation intact (correct `ComInitDB` per session).
- **Isolation/commit:** K=1 forced-reuse shows no session/branch/var bleed; a `DOLT_COMMIT` written through the pool is visible to a separate pooled reader.
- Tests + `golangci-lint`: green (incl. `-race`). Live-Dolt integration tests gated on `BEADS_POOL_TEST_DOLT_ADDR` (skipped in CI).

## Status / soak
Running on a live town now; churn collapse + stability confirmed across multiple steady-state windows (full results in the testing comment). Extended soak ongoing.

### DIFF
--- a/cmd/bd/db_proxy_child.go
+++ b/cmd/bd/db_proxy_child.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
@@ -29,6 +30,8 @@ var (
 	dbProxyChildExternalTLSCertPath string
 	dbProxyChildExternalTLSKeyPath  string
 	dbProxyChildExternalKeepAlive   time.Duration
+	dbProxyChildPoolSocket          string
+	dbProxyChildPoolSize            int
 )
 
 var dbProxyChildCmd = &cobra.Command{
@@ -58,6 +61,13 @@ not intended to be invoked directly by users.`,
 			KeepAlivePeriod: dbProxyChildExternalKeepAlive,
 		}
 
+		// Pooled mode: serve the QUERY-LEVEL connection pooler on a unix socket
+		// instead of the TCP byte-passthrough proxy, reusing the same lifecycle
+		// scaffolding (lock, pidfile, log, signal handling, idle shutdown).
+		if backend == proxy.BackendExternalPooled {
+			return runPooledProxyChild(cmd.Context(), external)
+		}
+
 		srv, err := newDatabaseServer(backend, dbProxyChildRoot, dbProxyChildConfig, dbProxyChildLogPath, dbProxyChildDoltBin, external)
 		if err != nil {
 			return err
@@ -77,6 +87,35 @@ not intended to be invoked directly by users.`,
 		}
 		return nil
 	},
+}
+
+// runPooledProxyChild runs the pooled (query-level) serving mode. The upstream
+// Dolt address comes from the external config; clients are served on the
+// pool socket recorded in the pidfile.
+func runPooledProxyChild(ctx context.Context, external configfile.ExternalDoltConfig) error {
+	if dbProxyChildPoolSocket == "" {
+		return fmt.Errorf("backend %q requires --pool-socket", proxy.BackendExternalPooled)
+	}
+	doltAddr := ""
+	if external.Host != "" {
+		doltAddr = fmt.Sprintf("%s:%d", external.Host, external.Port)
+	}
+	p := proxy.NewPooledServer(proxy.PooledOpts{
+		RootDir:     dbProxyChildRoot,
+		Socket:      dbProxyChildPoolSocket,
+		DoltAddr:    doltAddr,
+		DoltSocket:  external.Socket,
+		DoltUser:    external.User,
+		PoolSize:    dbProxyChildPoolSize,
+		IdleTimeout: dbProxyChildIdleTimeout,
+	})
+	if err := p.ListenAndServe(ctx); err != nil {
+		if errors.Is(err, proxy.ErrLockHeld) {
+			os.Exit(proxy.LockHeldExitCode)
+		}
+		return err
+	}
+	return nil
 }
 
 func newDatabaseServer(backend proxy.Backend, rootDir, configPath, logPath, doltBin string, external configfile.ExternalDoltConfig) (server.DatabaseServer, error) {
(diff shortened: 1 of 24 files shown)


rbriski:
## Testing

Validated in three stages — protocol/correctness proof, an isolated A/B churn measurement, and a live multi-database town soak. Bottom line: **~96% reduction in Dolt connection churn with no correctness regression**.

### 1. Protocol & session-isolation correctness

The pooler terminates the client handshake with vitess `go/mysql` (`AuthServerNone`) and multiplexes short-lived clients onto K persistent Dolt backends, resetting each backend between checkouts with `COM_RESET_CONNECTION`. Verified against a live Dolt sql-server:

- **go-sql-driver / `bd` compatibility**: real `bd` clients connect, query, and write through the pooler unchanged (unix-socket DSN).
- **No session-state bleed**: `COM_RESET_CONNECTION` fully clears per-session state between borrowers — current database, session variables, autocommit, and the active Dolt branch all reset. A client that switches branch/vars does not leak that state to the next client.
- **Dolt write semantics preserved**: `DOLT_COMMIT()` and the working set round-trip correctly through the pool; writes land and are durable (validated by creating + closing real beads through the pool).
- **Prepared statements**: `bd`'s server-mode DSN does not set `interpolateParams`, so go-sql-driver issues server-side `COM_STMT_PREPARE`/`COM_STMT_EXECUTE`. The pooler interpolates bound parameters back into SQL (vitess `GenerateQuery`) and executes plain text on the backend, so prepared-statement clients work without leaving per-statement state on a shared backend.

### 2. Isolated A/B churn measurement

Pooler activated via `BEADS_DOLT_POOL=1`, with `Connections` sampled from `SHOW GLOBAL STATUS`:

- **Direct (no pooler)**: 60 sequential `bd` invocations → **183** new Dolt connections.
- **Pooled**: 60 sequential `bd` invocations → **1** new Dolt connection.

Backend connection count is decoupled from client invocation rate, as designed.

### 3. Live town soak (hybrid deployment)

Deployed across a live multi-rig workspace (an `hq` database plus several product databases on a shared Dolt server) as a hybrid: one shared pooler service fronting the server, with existing clients routed through it by a thin wrapper that **auto-falls-back to a direct connection if the pooler socket is absent** — so the pooler is never a hard dependency.

Baseline churn before the pooler: **~557 new Dolt connections/min** — the level that periodically wedged the listener (alive-but-unresponsive, connection-pool exhaustion, `i/o timeout` on client queries).

Measured across consecutive steady-state windows with the pooler live:

| Window | Churn (conns/min) | Threads_connected | bd read | bd write |
|--------|-------------------|-------------------|---------|----------|
| 1 | 21 | 8  | OK | OK |
| 2 | 22 | 8  | OK (57 ms)  | OK (226 ms) |
| 3 | 13 | 8  | OK (58 ms)  | OK (222 ms) |
| 4 | 20 | 15 | OK (64 ms)  | OK (263 ms) |

- **Churn: ~557 → ~13–22/min (~96 % collapse)**, sustained across every window.
- **`Threads_connected` stays bounded** around 8, rising only toward the pool cap (16) under concurrent load — backends are pinned, not proportional to client rate.
- **Reads ~55–65 ms, writes ~220–260 ms** through the pool — healthy, no latency regression.
- **Pooler stability**: a single long-lived process, zero respawns over the soak; mail and all `bd` operations healthy throughout; no new Dolt escalations.
- Each window verified a real read **and** a real write (create + close a bead) end-to-end through the pool.

The schema-migration guard is not a factor: the pooler never opens a beads store (it only relays MySQL packets), so it fronts databases at any schema version without triggering migration logic — which is what lets the soak run a current-`main` pooler in front of older-schema clients.

An extended soak on the live town is ongoing. The wrapper's direct-fallback plus a one-symlink revert keep the deployment fully reversible at any point.

### Reproducing the churn measurement

```sh
sample() { mysql -h127.0.0.1 -P3307 -uroot -N \
  -e "SHOW GLOBAL STATUS LIKE 'Connections';" | awk '{print $2}'; }

C0=$(sample); for i in $(seq 60); do bd list --limit 1 >/dev/null; done; echo $(( $(sample) - C0 ))
# repeat with BEADS_DOLT_POOL=1 to compare pooled vs direct
```

rbriski:
Converting this back to draft. After opening it I found #4303 (db-proxy connection pooling + proxied-server external deploy path), which solves the same Dolt connection churn with the same session-pooling-proxy approach, is further along (federation, external deploy path, docs), and is part of the broader proxied-server workstream.

Rather than maintain a parallel implementation, we're taking our production validation — a live multi-rig town soak showing ~96% connection-churn collapse — over to #4303 and consolidating there. Keeping this open as draft only as a reference for the prepared-statement (`COM_STMT_PREPARE`/`EXECUTE`) handling, which differs from #4303's byte-transparent command phase; will close in favor of #4303 once that lands.

PR #4474, opened 2026-06-22
fix(docgen): generate CLI docs in one process

maphew:
Add a bulk help docs writer that emits the root CLI reference and website command pages from a single Cobra walk. Update the shell generator to call that mode, preserve the pure-Go docs binary guard, and refresh historical versioned CLI docs only when release snapshotting requests a specific version.

_Agent-Signature: codex-gpt-5.5-medium on behalf of matt wilkie_

### DIFF
--- a/cmd/bd/help_all.go
+++ b/cmd/bd/help_all.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"unicode"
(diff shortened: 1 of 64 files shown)


maphew:
The refactor is behavior-preserving for live docs and removes ~150 lines of duplicated shell logic. Two design observations a maintainer may want to weigh — both appear intentional, neither is a defect:

* scripts/check-cli-docs-drift.sh / scripts/check-doc-flags.sh — versioned-docs CI coverage is dropped entirely. Both scripts had their website/versioned_docs/*/cli-reference scanning removed, and the generator no longer rewrites those dirs on normal runs. Net effect: after this PR, nothing in CI validates the historical snapshots — they're only ever written by an explicit --versioned release run. This is the stated intent ("historical snapshots are release artifacts"), and is arguably more correct than the old behavior of rewriting a v1.0.0 snapshot to match dev main. Flagging only so the coverage reduction is a conscious choice.

* Altitude: bd help --docs-root makes the help command mutate the filesystem (writes files and, via removeMarkdownFiles, deletes *.md). Putting destructive doc-tree regeneration behind a help flag is mildly surprising, but it's consistent with the existing --doc/--all flags already hosted there, so it fits the established pattern.

Overall this looks solid and well-tested — I'd approve. The two notes above are context, not blockers.

PR #4478, opened 2026-06-23
Retire legacy Homebrew tap publishing

maphew:
## Why

Beads currently has two Homebrew formulas that can drift: Homebrew core `beads`
and the legacy tap formula `bd`. The tap formula is currently stale, so users
can install the wrong binary depending on which path they follow.

## What

- Remove the release workflow job that publishes `Formula/bd.rb` to
  `gastownhall/homebrew-beads`
- Document Homebrew core `beads` as the only supported Homebrew package
- Add migration commands for old `bd` tap installs
- Correct an old newsletter upgrade snippet from `brew upgrade bd` to
  `brew upgrade beads`

## Validation

- `git diff --check`
- `npx.cmd --yes js-yaml .github/workflows/release.yml`
- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd/doctor -run TestUpgradeCommandForPath -count=1`

Note: full `go test -tags gms_pure_go ./cmd/bd/doctor` hit existing Windows
temp/Dolt cleanup failures unrelated to this workflow/docs change.


maphew:
## Review: Retire legacy Homebrew tap publishing

Reviewed together with its paired tap change gastownhall/homebrew-beads#9 — they should land as a unit (stop publishing the tap formula, then convert the orphaned formula into a deprecation shim).

### Verified
- The `update-homebrew-formula` job is fully removed on the branch — no leftover `needs:`, `HOMEBREW_TAP`, or `homebrew-beads` references remain in `release.yml`.
- `TestUpgradeCommandForPath` (cited in validation) exists at `cmd/bd/doctor/version_test.go`.
- `NEWSLETTER.md` fix (`brew upgrade bd` → `brew upgrade beads`) is correct.
- `RELEASING.md` "Verify Homebrew" section already uses `brew upgrade beads` / `bd version` — no stale `bd` formula reference, nothing to change there.
- Migration snippet correctly untaps both legacy taps (`gastownhall/beads`, `steveyegge/beads`).

### Minor
- The migration snippet is duplicated verbatim across `docs/INSTALLING.md`, `website/docs/getting-started/installation.md`, and `website/docs/getting-started/upgrading.md`. Not worth blocking, but it will drift. Filed as a follow-up bead to consolidate into a single shared source.

### Verdict
Looks ready. The only caveat is ordering: land this (or at least stop the publishing job) together with gastownhall/homebrew-beads#9 so the shim isn't overwritten by a subsequent release.

maphew:
Hat tap to mike on Discord for bringing the homebrew version mismatch to my attention.

PR #4479, opened 2026-06-23
perf(schema): replace per-connection content_hash IS.COLUMNS probe with SHOW COLUMNS

johnzook:
## Human Commentary

During heavy load in Gas City this showed up as a major consumer of CPU cycles. All indications are that it is safe and performant. We are requesting the necessary data instead of grabbing significantly more from Dolt/database and narrowing down the desired data. I have not been able to find a reason not to make the change. No signs that there are any configurations where the change would not still function.

## What

Replace the per-connection `content_hash` existence probe in schema migration: `migrationSource.hasContentHashColumn` now uses `SHOW COLUMNS FROM <cursorTable> LIKE 'content_hash'` instead of `SELECT COUNT(*) FROM INFORMATION_SCHEMA.COLUMNS`.

## Why

Dolt does not push the `INFORMATION_SCHEMA.COLUMNS` predicate down, so the old probe materialized **every column of every table across all databases** on each call. `migrationWorkNeeded` runs the probe **twice on every connection**, so on a multi-database deployment (5 DBs / ~1,843 columns) it cost ~19–60 ms/conn and was a measured ~34% of Dolt CPU — a recurring connection-storm driver. `SHOW COLUMNS` answers from a single table's schema without scanning `INFORMATION_SCHEMA`.

Correctness is preserved:

- The `Field` name is compared **exactly**, because `_` is a LIKE single-character wildcard.
- A missing cursor table makes `SHOW COLUMNS` error with 1146 where the old probe returned count 0; that error is swallowed (via `dberrors.IsTableNotExist`) so a not-yet-created table still reports *no content_hash column*.
- The separate id-column re-key probe in `dep_id_backfill.go` still uses `INFORMATION_SCHEMA.COLUMNS` and is intentionally unchanged (it runs only during a migration pass, not per-connection).

## Verification

```
go build ./...
go test ./internal/storage/schema/...
go test ./internal/storage/dolt/ -run='^$' -bench=BenchmarkContentHashColumnProbe -benchmem
```

- `BenchmarkContentHashColumnProbe` (before/after A/B): `INFORMATION_SCHEMA.COLUMNS` ~9.49 ms/op vs `SHOW COLUMNS … LIKE` ~0.22 ms/op (~43× faster). Reproduced locally on Dolt 2.1.8 at ~18 ms → ~0.43 ms.
- `TestHasContentHashColumnUsesShowColumns` (sqlmock): present, absent, missing-table-reports-false, LIKE-wildcard-rejected, error-propagation.
- `TestHasContentHashColumnMatchesInformationSchemaOnDolt` (real Dolt, gated by `RequireDoltBinary`): the new probe agrees with the retired `INFORMATION_SCHEMA` probe in both the present and absent states.

---
_Claude Code-claude-opus-4-8-high on behalf of John Zook

### DIFF
--- a/internal/storage/dolt/dolt_benchmark_test.go
+++ b/internal/storage/dolt/dolt_benchmark_test.go
@@ -1902,3 +1902,46 @@ func BenchmarkGetIssuesByIDs_SmallNLargeW_10_10K(b *testing.B) {
 func BenchmarkGetIssuesByIDs_SmallNLargeW_100_5K(b *testing.B) {
 	benchmarkGetIssuesByIDsSmallN(b, 5000, 100)
 }
+
+// =============================================================================
+// Schema Probe Benchmarks
+// =============================================================================
+
+// BenchmarkContentHashColumnProbe compares the retired INFORMATION_SCHEMA.COLUMNS
+// existence probe against the SHOW COLUMNS probe that replaced it.
+// schema.MigrateUp's migrationWorkNeeded runs this probe twice on every
+// connection. INFORMATION_SCHEMA.COLUMNS forces Dolt to materialize the full
+// column catalog because the predicate is not pushed down; SHOW COLUMNS reads a
+// single table's schema directly. The gap measured here understates production:
+// on a multi-database deployment the old probe was measured at 19-60 ms/conn
+// versus ~0 ms on the new one.
+func BenchmarkContentHashColumnProbe(b *testing.B) {
+	store, cleanup := setupBenchStore(b)
+	defer cleanup()
+	ctx := context.Background()
+
+	b.Run("InformationSchema", func(b *testing.B) {
+		b.ReportAllocs()
+		for i := 0; i < b.N; i++ {
+			var count int
+			if err := store.db.QueryRowContext(ctx,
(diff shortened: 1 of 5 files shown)


maphew:

This is the best-tested of the batch and the mechanism checks out: the bool semantics are provably equivalent to the old `INFORMATION_SCHEMA.COLUMNS` probe (the exact Field comparison neutralizes the `_` LIKE wildcard, and both the DATABASE()-scoped probe and unqualified `SHOW COLUMNS` resolve against the current database).

On the cross-platform/DB concern you flagged in the description: it is well-contained. The missing-table swallow rides `dberrors.IsTableNotExist`, which `currentVersion()` already depends on in the identical runtime paths, and embedded and sql-server modes share the same GMS engine, so this PR does not introduce a new trust surface. The sqlmock units plus the gated real-Dolt parity test and benchmark cover the rest.

Plan: we are at v1.1.0-rc.2 and holding main until stable ships. Since this touches the per-connection migration-work-needed path, it lands just after 1.1.0 stable once the newly approved CI runs are green; it is also the cleanest candidate to pull into an early 1.1.x patch if we want the per-connection CPU win sooner.

PR #4481, opened 2026-06-23
fix(dolt): close db pool on any failure path in newServerMode/openServerConnection

johnzook:
## Human Commentary

Original fix was adding the db.Close() that's missing in the following block, which leads to a connection leak (the underlying bug being fixed):
```
		if err := store.initSchema(ctx); err != nil {
			return nil, fmt.Errorf("failed to initialize schema: %w", err)
		}
```
but that just patches what is a structural problem that allows this class of errors to exist. So rewrote to protect all early returns in both functions. Claude's audit of other similar places in the code to apply this structural fix came up with nothing else.

## What

Installs a single deferred close guard at the point of pool acquisition in
both `newServerMode` and `openServerConnection` (`internal/storage/dolt/store.go`),
so the Dolt connection pool is closed on every early-return failure path and
released only once ownership transfers to the caller.

## Why

`newServerMode`'s `initSchema` branch returned without closing the pool, leaking
it. In a long-lived process these leaks accumulate toward `max_connections`
(`ConnMaxLifetime=1h` does not promptly reap idle leaked conns) — the suspected
driver of the multi-day connection creep. `openServerConnection` had the same
shape: six separate error returns each repeating `_ = db.Close()`, one bad merge
away from the same leak.

Rather than patch the one missing close, the guard makes the path forget-proof:

```go
ready := false
defer func() {
    if !ready {
        _ = db.Close()
    }
}()
```

`ready` is set true only at the success return, where ownership passes to the
caller (`newServerMode` -> `store.Close()`; `openServerConnection` -> the returned
`*sql.DB`). All the call-site `_ = db.Close()` closes are removed in favor of the
guard, so a future failure path cannot reintroduce the leak.

## Verification

- `go build ./internal/storage/dolt/`
- `go test ./internal/storage/dolt/`

_Claude Code-claude-opus-4-8-high on behalf of John Zook_

### DIFF
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -1125,9 +1125,16 @@ func newServerMode(ctx context.Context, cfg *Config) (*DoltStore, error) {
 		return nil, err
 	}
 
+	// Close the pool on any failure path below; cleared once ownership passes to the caller.
+	storeReady := false
+	defer func() {
+		if !storeReady {
+			_ = db.Close()
+		}
+	}()
+
 	// Test connection
 	if err := db.PingContext(ctx); err != nil {
-		_ = db.Close()
 		return nil, fmt.Errorf("failed to ping Dolt database: %w", err)
 	}
 
@@ -1161,7 +1168,6 @@ func newServerMode(ctx context.Context, cfg *Config) (*DoltStore, error) {
 	// would fail with cryptic unknown-column errors instead of a clear
 	// "upgrade bd" message (the stale-binary incident behind #4135/#4137).
 	if err := schema.CheckForwardDrift(ctx, db); err != nil {
-		_ = db.Close()
 		return nil, err
 	}
 	if !cfg.ReadOnly {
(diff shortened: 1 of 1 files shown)


maphew:
Triaged as part of a maintainer sweep of your five open PRs (full report: [johnzook-pr-triage-2026-07-03](https://github.com/maphew/mybd/blob/main/reports/johnzook-pr-triage-2026-07-03.md)). Thanks for these, and for your patience - a gap in our triage automation let them sit unseen; that is filed and being fixed.

Verified correct: the diff base is byte-identical to current main (zero drift), the sentinel-guarded defer is sound (no double-close, ownership transfer via the ready flag is right, `initDB` keeps its own independent close), and it fixes a genuine defect - the `initSchema` failure branch in `newServerMode` returned without closing the pool. The other rewritten branches were already closing correctly, so for them this is a forget-proofing refactor.

One framing note: the leak fires only on failure paths, not steady-state, so urgency is lower than a multi-day connection creep would imply. That affects timing, not correctness.

Plan: we are at v1.1.0-rc.2 and holding main until stable ships. This is the safest of your five and lands first in the post-stable batch, once the newly approved CI runs come back green. Optional but welcome: a regression test asserting the pool closes on an induced `initSchema` failure (for example via `sql.DB.Stats().OpenConnections`); if you do not get to it we can absorb that at merge time with credit.

PR #4484, opened 2026-06-23
fix: honor DOLT_REMOTE_USER during clone

iamthebot:
## What

This makes fresh clones honor `DOLT_REMOTE_USER`, matching the authentication behavior Beads already uses for pull, push, and fetch.

When the variable is set, the CLI clone paths now call `dolt clone --user ...`, and the embedded/external-server paths call `DOLT_CLONE('--user', ...)`. When it is unset, Beads uses the same unauthenticated clone calls it uses today.

## Why

This failure only appears on a clean machine or checkout, which makes it easy to miss.

A user can configure an authenticated Dolt remote, export both `DOLT_REMOTE_USER` and `DOLT_REMOTE_PASSWORD`, and use that setup successfully for normal pull and push operations. But if the local Beads database does not exist yet, `bd init` or `bd bootstrap` takes one of the clone paths instead. Those paths inherited the password environment variable but never passed Dolt the username.

Dolt only emits the remote's Basic authorization metadata when `--user` is present, so the server sees the clone as anonymous and rejects it with its `CLONE_ADMIN` authorization check. The confusing part is that the same credentials work once a database has already been cloned, because the later pull/push code already handles the username correctly.

This change closes that gap across all of the places Beads can initiate a clone:

- embedded `DOLT_CLONE`
- external-server `DOLT_CLONE`
- owned-server/CLI bootstrap
- multi-repository remote cache hydration

The password stays in `DOLT_REMOTE_PASSWORD`; it is not added to process arguments, logs, or configuration. The existing unauthenticated behavior remains the default and can still be selected simply by leaving `DOLT_REMOTE_USER` unset.

## Reproduction

Captured from the unpatched parent commit:

```bash
bd version
# bd version 1.1.0-rc.1 (26be6aa9: main@26be6aa94c89)

# Storage mode: embedded (the default; no --server flag)
# OS: macOS 26.4 (25E246), arm64
# Go: go1.26.2 darwin/arm64
```

The failure can be reproduced in a scratch Git repository against any Dolt
remotes API endpoint that enforces username/password authentication:

```bash
export REMOTE_URL=http://127.0.0.1:59962/test_database
export DOLT_REMOTE_USER=beads_user
export DOLT_REMOTE_PASSWORD=correct-password

tmp="$(mktemp -d)"
git init --quiet "$tmp/repo"
cd "$tmp/repo"

bd init --non-interactive --skip-agents --skip-hooks \
  --prefix repro \
  --remote "$REMOTE_URL"
```

### Actual behavior before this PR

The command exits `1`. The relevant stderr is:

```text
Error: failed to clone remote "http://127.0.0.1:59962/test_database":
clone from remote: dolt clone http://127.0.0.1:59962/test_database:
Error 1105: could not access dolt url
'http://127.0.0.1:59962/test_database':
rpc error: code = PermissionDenied desc = unauthenticated user does not
have privilege CLONE_ADMIN
```

The local HTTP/2 mock recorded the first Dolt RPC as:

```text
RPC: /dolt.services.remotesapi.v1alpha1.ChunkStoreService/GetRepoMetadata
Authorization: <absent>
```

`DOLT_REMOTE_PASSWORD` is present in the process environment, but without
`--user` Dolt does not generate Basic authorization metadata.

### Expected behavior

When both credential variables are set, the clone request should carry:

```text
Authorization: Basic <redacted>
```

With this PR, the same scratch-repo command reaches the mock with that header.
A real server can then validate the supplied username/password and continue the
clone. Leaving `DOLT_REMOTE_USER` unset preserves the existing unauthenticated
behavior.

The automated regression tests exercise the same before/after distinction with
mock SQL and remotes API endpoints, so reviewers do not need to provision an
authenticated server to run the test suite.

## Verification

- focused tests for authenticated and unauthenticated SQL clone calls
- focused tests for CLI and remote-cache clone arguments
- `make fmt-check`
- canonical `gms_pure_go` build
- focused package tests with `-tags gms_pure_go`
- end-to-end fresh `bd init --remote` test against a mock Dolt remotes API, confirming the expected Basic authorization metadata is sent

I also ran the repository's full `make test`. The packages changed here passed, while current `main` hit two unrelated failures: `cmd/bd` exceeded its three-minute timeout in `TestInitMetricsEnvConfigPrecedence`, and `cmd/bd/doctor` failed the existing `TestExpandPath/tilde_only` path assertion.

## AI disclosure

This contribution was substantially prepared with OpenAI Codex. AI assistance
was used to investigate the existing authentication paths and upstream history,
implement the change, write tests, run validation, and draft this pull request.
The commit also carries the repository's required `Agent-Signature` trailer so
that the assistance is visible in Git history as well as in this PR.

### DIFF
--- a/cmd/bd/bootstrap.go
+++ b/cmd/bd/bootstrap.go
@@ -846,7 +846,7 @@ func cloneViaEmbedded(ctx context.Context, beadsDir, remoteURL, dbName string) e
 	}
 	defer func() { _ = cleanup() }()
 
-	if err := versioncontrolops.DoltClone(ctx, db, remoteURL, dbName); err != nil {
+	if err := versioncontrolops.DoltClone(ctx, db, remoteURL, dbName, os.Getenv("DOLT_REMOTE_USER")); err != nil {
 		return fmt.Errorf("clone from remote: %w", err)
 	}
 	fmt.Fprintf(os.Stderr, "Synced database from %s\n", remoteURL)
@@ -883,7 +883,7 @@ func cloneViaServer(ctx context.Context, beadsDir, remoteURL, dbName string, cfg
 			cfg.GetDoltServerHost(), port, err)
 	}
 
-	if err := versioncontrolops.DoltClone(cloneCtx, db, remoteURL, dbName); err != nil {
+	if err := versioncontrolops.DoltClone(cloneCtx, db, remoteURL, dbName, os.Getenv("DOLT_REMOTE_USER")); err != nil {
 		return fmt.Errorf("clone from remote via server: %w", err)
 	}
 	fmt.Fprintf(os.Stderr, "Synced database from %s (via server at %s:%d)\n",
--- a/docs/DOLT.md
+++ b/docs/DOLT.md
@@ -447,8 +447,8 @@ dolt:
 | `BEADS_DOLT_SERVER_TLS` | Enable TLS (set to "1" or "true") |
 | `BEADS_DOLT_SERVER_USER` | MySQL connection user |
 | `BEADS_DOLT_SHARED_SERVER` | Enable shared server mode (set to "1" or "true") |
-| `DOLT_REMOTE_USER` | Push/pull auth user |
-| `DOLT_REMOTE_PASSWORD` | Push/pull auth password |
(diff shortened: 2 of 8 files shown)


maphew (commented):
Test

PR #4476, opened 2026-06-23
chore: prepare v1.1.0-rc.1 release

maphew:
## Why

v1.0.5 was pulled after #4259 because it could damage cross-clone sync. We need an RC from current main, but the current release workflow publishes Homebrew/npm/PyPI for every `v*` tag. That is not safe for `v1.1.0-rc.1`.

## What

- Bumps source/package/plugin versions to `1.1.0-rc.1`.
- Stamps the changelog and `bd info --whats-new` for the #4259 recovery candidate.
- Allows prerelease SemVer in `scripts/update-versions.sh`.
- Leaves stable docs on the latest stable snapshot for prereleases.
- Keeps Windows PE fixed resource and manifest versions numeric (`1.1.0`) while string metadata reports `1.1.0-rc.1`.
- Skips Homebrew, npm, and PyPI publishing jobs for prerelease tags while still allowing GitHub prerelease binaries.

## Validation

- `GITHUB_REF=refs/tags/v1.1.0-rc.1 GITHUB_REF_TYPE=tag ./scripts/check-versions.sh`
- `npm pack --dry-run` in `npm-package`
- `python -m build --wheel --sdist` in `integrations/beads-mcp`
- `go build -tags gms_pure_go -o build/bd-rc-smoke.exe ./cmd/bd`
- `./build/bd-rc-smoke.exe version`
- `git diff --check`
- `bash -n scripts/update-versions.sh scripts/gen-winres.sh scripts/check-docs-version.sh`

Local Windows caveat: `go test -tags gms_pure_go ./scripts` and a focused `cmd/bd` test run hit existing Windows path assumptions unrelated to this change.

_codex-gpt-5.5-xhigh on behalf of matt wilkie_

PR #4486, opened 2026-06-24
chore: re-land v1.1.0-rc.1 release prep

maphew:
## Why

#4476 auto-merged the v1.1.0-rc.1 release-prep change before maintainers
could review it. #4477 backed that out so the change could land through
normal review instead of by auto-merge. This PR re-lands the RC prep on top
of the post-revert `main` tip, so maintainers can review it here before the
candidate is tagged and published.

This is the review-and-publish gate: merging this PR puts the RC metadata on
`main`; cutting the `v1.1.0-rc.1` tag afterward is what actually publishes the
prerelease. v1.0.5 was pulled from the default release path after #4259; this
candidate carries the cross-clone data-safety fixes for validation before
stable promotion.

## Highlights since 1.0.5

- **Data-safety recovery for #4259.** Dependency and history-table row IDs now
  converge deterministically across clones (migration `0050` rekeys existing
  rows), `schema_migrations` records per-migration content hashes, and
  `bd doctor` surfaces migration-content skew instead of leaving operators
  with cryptic Dolt merge failures.
- **Safer upgrades for shared remotes.** Remote-backed databases no longer
  silently apply pending schema migrations in place unless a designated
  migrator opts in with `BD_ALLOW_REMOTE_MIGRATE=1`; recovery guidance now
  recognizes Dolt primary-key merge refusals and points to the canonical-clone
  bootstrap path (`docs/RECOVERY.md#pk-fork-refused`).
- **More reliable sync/merge behavior.** `bd dolt pull` repairs supported
  foreign-key cascade conflicts, handles mixed-version migration rows where
  safe, recomputes `is_blocked` after merges, and adds `bd recompute-blocked`
  plus doctor repair coverage for stale blocker state.
- **Broader server/proxied mode coverage.** More core commands gained
  server-mode implementations and parity coverage
  (list/update/show/delete/close/dep/reopen/config/ready), with connection
  lifecycle and retry fixes around embedded Dolt / server mode.
- **CLI and workflow polish.** `bd init --init-if-missing`,
  `bd count --include-infra`, `children --pretty`, `defer --reason`,
  non-truncated piped `bd list`, safer worktree/where/no-db routing, and Jira
  custom-field support.
- **Recovery, docs, and release safety.** Compaction archives content before
  destructive cleanup; docs/community tooling/recovery guidance refreshed;
  CI/release gates catch version/docs/migration drift; RC release automation
  builds GitHub prerelease artifacts without updating stable Homebrew / npm /
  PyPI channels.

See `CHANGELOG.md` and the `cmd/bd/info.go` entry restored in this PR for the
full per-change detail and upgrade notes.

## What

Re-applies the exact change reverted in #4477 (revert-of-the-revert on the
current tip), restoring:

- `v1.1.0-rc.1` version metadata across all package manifests (Go, npm,
  PyPI/uv, Nix, Windows PE resources, plugin manifests, marketplace).
- The prerelease-aware release workflow and version scripts
  (`release.yml`, `update-versions.sh`, `check-docs-version.sh`,
  `gen-winres.sh`): prerelease tags skip the stable docs-snapshot
  requirement and stay off the stable Homebrew/npm/PyPI channels.
- The `CHANGELOG.md` `1.1.0-rc.1` section and the `cmd/bd/info.go`
  agent-actionable version entry.

Also regenerates `integrations/beads-mcp/uv.lock` to `1.1.0rc1`. The original
#4476 / #4477 round left `uv.lock` out of sync with `pyproject.toml`, which
reddened the Package Gate (MCP) check; this keeps them consistent.

## Validation

- `uv lock --locked` passes (pyproject and lockfile agree on `1.1.0rc1`).
- `scripts/check-versions.sh`
- Diff vs `main` is exactly the 18 release-prep files, no unrelated changes.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
       "name": "beads",
       "source": "./plugins/beads",
       "description": "AI-supervised issue tracker for coding workflows",
-      "version": "1.0.5"
+      "version": "1.1.0-rc.1"
     }
   ]
 }
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,7 +46,14 @@ jobs:
           echo "OK: tag and version.go agree on $code_version"
 
       - name: Check version + docs consistency
-        run: BEADS_REQUIRE_RELEASE_DOCS=1 ./scripts/check-versions.sh
+        run: |
+          tag_version="${GITHUB_REF_NAME#v}"
+          if [[ "$tag_version" == *-* ]]; then
+            echo "Prerelease tag detected; stable docs snapshot may remain on latest stable release."
+            ./scripts/check-versions.sh
+          else
+            BEADS_REQUIRE_RELEASE_DOCS=1 ./scripts/check-versions.sh
+          fi
 
   release-package-mcp:
     name: Release Package Gate (MCP)
@@ -338,7 +345,7 @@ jobs:
   publish-pypi:
     runs-on: ubuntu-latest
     needs: [release-package-mcp, goreleaser]
-    if: ${{ github.repository == 'gastownhall/beads' && startsWith(github.ref, 'refs/tags/v') }}
+    if: ${{ github.repository == 'gastownhall/beads' && startsWith(github.ref, 'refs/tags/v') && !contains(github.ref_name, '-') }}
     steps:
       - name: Checkout
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v6
@@ -373,7 +380,7 @@ jobs:
   publish-npm:
     runs-on: ubuntu-latest
     needs: [release-package-npm, goreleaser, goreleaser-macos]
-    if: ${{ github.repository == 'gastownhall/beads' && startsWith(github.ref, 'refs/tags/v') }}
+    if: ${{ github.repository == 'gastownhall/beads' && startsWith(github.ref, 'refs/tags/v') && !contains(github.ref_name, '-') }}
     permissions:
       contents: read
       id-token: write  # Required for npm provenance/trusted publishing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.1.0-rc.1] - 2026-06-23
+
 ### Upgrade Notes
 
 - **Mixed bd versions sharing one Dolt remote.** Pre-1.0.6 binaries record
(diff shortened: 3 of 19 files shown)


maphew:
## Review

Traced the prerelease CI gating end-to-end. The core logic holds up:

- The `tag_version == *-*` dash check in `verify-version-consistency` is safe because that job is tag-only (`startsWith(github.ref, 'refs/tags/v')`), so `GITHUB_REF_NAME` is always `v…`.
- The `release.yml` prerelease branch correctly relies on the new `IS_PRERELEASE` guard in `check-docs-version.sh` to suppress the tag-triggered `REQUIRE_RELEASE_DOCS=1` auto-set; the two changes are coordinated.
- `publish-pypi` / `publish-npm` gates (`!contains(github.ref_name, '-')`) skip prereleases. Homebrew needs no new gate: goreleaser already has `brews.skip_upload: true` + `release.prerelease: auto`, so it never pushes the formula and marks the GitHub release as prerelease automatically.
- `check-versions.sh` compares every manifest to `CANONICAL` (`1.1.0-rc.1`) exactly, and the diff sets them all to that value, so this transition validates. `uv.lock` PEP 440 normalization (`1.1.0rc1`) is consistent with `pyproject.toml`.

### Finding (latent): `scripts/update-versions.sh` — winres/manifest numeric-version search uses `CURRENT_VERSION`, silently no-ops on cross-base prerelease bumps

The write side now strips the prerelease suffix:

```sh
update_file "cmd/bd/winres/winres.json" "\"file_version\": \"$CURRENT_VERSION\"" "\"file_version\": \"$BASE_VERSION\""
update_file "cmd/bd/winres/manifest.xml" "version=\"$CURRENT_VERSION.0\"" "version=\"$BASE_VERSION.0\""
```

but the **search** pattern still uses the full `CURRENT_VERSION` (read from `version.go`, which carries the prerelease suffix). After a prerelease bump, `winres.json` holds the base form (`"file_version": "1.1.0"`) while `version.go` holds `1.1.0-rc.1`. On the next bump across a *different* base (e.g. `1.1.0-rc.1` → `1.2.0-rc.1`), the pattern `"file_version": "1.1.0-rc.1"` will not match the file's actual `"file_version": "1.1.0"`. `update_file` is a plain `sed -i` with no match assertion, so it silently does nothing and the Windows PE numeric file/product version stays stuck at the old base (`1.1.0`) — with no error. The string `FileVersion` / `ProductVersion` fields (written with `NEW_VERSION`) update correctly, so the file ends up internally inconsistent, and `check-versions.sh` does not inspect the numeric fields to catch it.

This PR's `1.0.5 → 1.1.0-rc.1` transition is unaffected (base unchanged), which is why it passes here; the trigger is a future prerelease-to-different-base bump.

**Fix applied on this branch:** derive `CURRENT_BASE="${CURRENT_VERSION%%-*}"` and match the numeric winres/manifest fields on the base form rather than on `CURRENT_VERSION`.

Nothing else surfaced as maintainer-actionable; the remainder of the diff is mechanical version-string bumps that line up with the validation gates.

PR #4489, opened 2026-06-24
bd sql: proxied-server support

coffeegoddd:
## Summary

Adds proxied-server support for `bd sql`, routing raw SQL through a new domain use-case interface (`RawSQLUseCase`) like the other proxied commands. The non-proxied path is untouched; all proxied logic lives in a new `sql_proxied_server.go`, following the `update.go` / `update_proxied_server.go` split.

The local-vs-proxied difference is confined to the `uow` transaction layer (the mode-polymorphic `db.Runner`), so the new use case is the same code in both modes — consistent with how create/update/query/list already work.

## What's included

- **`internal/storage/domain/raw_sql.go`**: `RawSQLUseCase` (+ `RawSQLRepository`) with `Query`/`Exec` and a `RawSQLResult{Columns, Rows}`. `Query` materializes rows inside the open transaction (the `*sql.Conn` lifetime is bound to the UOW); `[][]any` aligned to `Columns` preserves order and tolerates duplicate column names.
- **`internal/storage/domain/db/raw_sql.go`**: repository over `db.Runner` — generic scan loop with `[]byte`→`string` normalization for `Query`, `RowsAffected` for `Exec`.
- **`internal/storage/uow/uow.go`**: `RawSQLUseCase()` added to the `UnitOfWork` interface and `baseUOW` (lazy accessor), mirroring the other six aggregates.
- **`cmd/bd/sql.go`**: early `usesProxiedServer()` dispatch before the existing `RawDBAccessor`/`store` path (which is `nil` in proxied mode).
- **`cmd/bd/sql_proxied_server.go`**: `runSQLProxiedServer` — read path opens a read-only UOW and calls `RawSQLUseCase().Query`; write path runs `CheckReadonly` then `uow.RunInTxMsg(...)` so the write commits via `CALL DOLT_COMMIT`. JSON/CSV/table rendering matches the embedded `bd sql` output.
- **`cmd/bd/sql_proxied_integration_test.go`**: parity with the embedded `TestSqlCommand` (select count, filtered select, empty result/table) plus proxied-specific coverage — CSV rendering, the exec/`rows_affected` write path, cross-connection persistence (verifies the `DOLT_COMMIT` is visible from an independent `*sql.DB`), and an invalid-SQL error case.

## Notes

- The `bd init --proxied-server` gate ("not yet implemented") is left **intact**; the sql proxied path stays dormant behind it, like the already-merged proxied commands. Lifting the gate is a separate, deliberate change.
- The new integration test is opt-in via `BEADS_TEST_PROXIED_SERVER=1` (requires a Dolt binary), same as the existing proxied command tests. It passes locally end-to-end.
- Behavioral note for review: a proxied raw write becomes an explicit Dolt commit (`bd sql: <query>`), unlike server/embedded auto-commit — consistent with every other proxied mutation, but worth a look since `bd sql` is the escape hatch.

### DIFF
--- a/cmd/bd/sql.go
+++ b/cmd/bd/sql.go
@@ -47,6 +47,10 @@ WARNING: Direct database access bypasses the storage layer. Use with caution.`,
 		query := args[0]
 		csvOutput, _ := cmd.Flags().GetBool("csv")
 
+		if usesProxiedServer() {
+			return runSQLProxiedServer(cmd, rootCtx, query, csvOutput)
+		}
+
 		if store == nil {
 			return HandleErrorRespectJSON("no database connection available (%s)", diagHint())
 		}
--- a/cmd/bd/sql.go
+++ b/cmd/bd/sql.go
@@ -48,7 +48,7 @@ WARNING: Direct database access bypasses the storage layer. Use with caution.`,
 		csvOutput, _ := cmd.Flags().GetBool("csv")
 
 		if usesProxiedServer() {
-			return runSQLProxiedServer(cmd, rootCtx, query, csvOutput)
+			return runSQLProxiedServer(rootCtx, query, csvOutput)
 		}
 
 		if store == nil {
--- a/cmd/bd/sql_proxied_integration_test.go
+++ b/cmd/bd/sql_proxied_integration_test.go
@@ -1,0 +1,180 @@
+//go:build cgo
+
+package main
+
+import (
+	"encoding/json"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+func bdProxiedSQL(t *testing.T, bd, dir string, args ...string) (string, string) {
+	t.Helper()
+	fullArgs := append([]string{"sql"}, args...)
+	stdout, stderr, err := bdProxiedRunBuffers(t, bd, dir, fullArgs...)
+	if err != nil {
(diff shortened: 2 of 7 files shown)


kevglynn:

Ran a complexity-only review on this diff. The PR follows project conventions faithfully — the bloat here is mostly inherited pattern duplication, not invention.

`internal/storage/domain/raw_sql.go`: **yagni:** `RawSQLUseCase` has identical signatures to `RawSQLRepository` and is pure delegation with zero business logic. Could wire `RawSQLRepository` through the UOW directly. (-17 lines)

`cmd/bd/sql_proxied_server.go` (renderRawSQLResult): **shrink:** The renderer is copy-pasted from `sql.go` — width calc, 60-char cap, separator, truncation all identical. A shared renderer over `RawSQLResult` would deduplicate both paths. (-78 lines, but ~80 of the duplicate is pre-existing in `sql.go`)

`cmd/bd/sql_proxied_server.go` (isRead): **shrink:** `isRead` prefix classification duplicates `sql.go` verbatim. A shared `isReadQuery(string) bool` helper would cover both. (-5 lines)

**net: -95 lines possible**, though most of the savings require touching the existing `sql.go` too — better as a follow-up than a prerequisite.

Not a blocker. The only novel over-engineering is the empty use-case layer; the rest is inherited.

PR #4491, opened 2026-06-25
fix(dolt): retry mutate+commit methods on serialization conflict (1213/1205)

steveyegge:
## Problem

`ClaimIssue`, `ClaimReadyIssue`, `UpdateIssue`, and `CloseIssue` each hand-roll their own transaction (`BeginTx` + `DOLT_ADD`/`DOLT_COMMIT` + `Commit`) **outside** `withRetryTx`. Under concurrent writers, a transaction that loses Dolt's optimistic commit-time working-set merge gets **MySQL 1213/1205** (`ErrLockDeadlock` wrapping `ErrRetryTransaction`) — a *guaranteed server-side rollback* — and these methods surface it as a hard failure instead of retrying.

These four are exactly the multi-agent **work-queue hot paths**: `bd ready --claim` (dequeue), claim, update, close. With N agents pulling from one shared Dolt server, claim/close contention is the common case, not the edge case.

## Why retry is the only option

Dolt has **no real row locking** — `FOR UPDATE` / `FOR UPDATE OF ... SKIP LOCKED` are *parse-only no-ops*. Straight from the engine:

> `buildForUpdateOf` ... `FOR UPDATE` in general is a no-op, so `FOR UPDATE OF` is also a no-op: https://www.dolthub.com/blog/2023-10-23-hold-my-beer/
> — `go-mysql-server` `sql/planbuilder/select.go`

So the classic `SELECT ... FOR UPDATE SKIP LOCKED` queue pattern can't help here; concurrency safety has to come from **retry on the serialization error**, which `withRetryTx` (`store.go`) already implements with backoff (it explicitly classifies 1213/1205 as retryable). The claim/update/close paths just weren't using it.

## Fix

Move each method body into a `withRetryTx` closure. `withRetryTx`→`withWriteTx` owns `BeginTx` and the final `Commit` and retries serialization failures; the closure keeps the existing **single-tx CAS/mutation + `DOLT_COMMIT`** atomicity exactly (no behavior change on the happy path). On retry, `ClaimReadyIssue` naturally re-scans the ready front from a fresh snapshot and claims the next available issue — the correct dequeue semantics.

No new helpers, no signature changes; `wrapTransactionError` stays in use by other call sites.

## Testing

- `go build` + `go vet ./internal/storage/dolt/` clean
- Concurrency tests pass: `TestEmbeddedReadyClaimConcurrent`, `TestEmbeddedReadyConcurrent`, `TestEmbeddedCloseConcurrent`, `TestEmbeddedUpdateConcurrent`
- Non-concurrent embedded paths pass: `TestEmbeddedUpdate`, `TestEmbeddedClose`, `TestEmbeddedReady` (+ routed/auto-commit variants)
- One unrelated pre-existing failure observed (`TestWatchIssueDetectsFieldUpdate`, a watch-snapshot timing test) — fails identically on the clean base, not touched here.

## Context

Found while building a multi-agent Java→Kotlin port that coordinates several agent clones through one shared Beads DB as a dependency-ordered work queue. `bd ready --claim` + atomic claim is already a great dequeue primitive; this closes the gap that made it fail under the very concurrency it's meant for.

### DIFF
--- a/internal/storage/dolt/concurrent_test.go
+++ b/internal/storage/dolt/concurrent_test.go
@@ -955,3 +955,127 @@ func TestSerializationConflictRetry(t *testing.T) {
 		t.Fatalf("expected %d labels, got %d: %v", numGoroutines, len(labels), labels)
 	}
 }
+
+// =============================================================================
+// Test: Concurrent Work-Queue Drain (the "Gas Station" scenario)
+//
+// N workers concurrently dequeue from ONE shared ready-front via
+// ClaimReadyIssue until it returns nil, exactly as N agent clones draining a
+// shared Beads work queue would. This is the core Gas Station claim-queue
+// invariant and the scenario the multi-agent port harness depends on:
+//
+//   - every issue is claimed by EXACTLY ONE worker (no double-claim / lost work)
+//   - every issue is claimed (no stranded ready work left behind)
+//   - the count of distinct claims equals the number of issues
+//
+// Dolt has no SKIP LOCKED, so the safety comes from the claim CAS (UPDATE ...
+// SET assignee WHERE assignee IS NULL) colliding on the same cell, surfacing as
+// a 1213/1205 serialization conflict, which ClaimReadyIssue's withRetryTx
+// re-scans and retries. This test is the regression guard for that guarantee.
+// =============================================================================
+
+func TestConcurrentWorkQueueDrain(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping work-queue drain test in short mode")
+	}
+
+	store, cleanup := setupConcurrentTestStore(t)
+	defer cleanup()
+
+	ctx, cancel := concurrentTestContext(t)
+	defer cancel()
+
+	// Seed a ready-front: all open, unassigned, no blockers => all ready.
+	const numIssues = 40
+	want := make(map[string]bool, numIssues)
+	for i := 0; i < numIssues; i++ {
+		id := fmt.Sprintf("queue-%03d", i)
+		issue := &types.Issue{
+			ID:          id,
+			Title:       fmt.Sprintf("Queue item %d", i),
+			Description: "ready work for the shared drain",
+			Status:      types.StatusOpen,
+			Priority:    (i % 4) + 1,
+			IssueType:   types.TypeTask,
+		}
+		if err := store.CreateIssue(ctx, issue, "seeder"); err != nil {
+			t.Fatalf("seed issue %s: %v", id, err)
+		}
+		want[id] = true
+	}
+
+	const numWorkers = 6
+
+	var mu sync.Mutex
+	claimedBy := make(map[string]string, numIssues) // issueID -> worker that claimed it
+	var doubleClaim atomic.Int32
+	var claimErrs atomic.Int32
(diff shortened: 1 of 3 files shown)


steveyegge:
For the Dolt folks (cc @timsehn) — this PR is the concrete artifact behind the leasing / multi-agent work-queue email thread.

The thing we confirmed by reading source (`go-mysql-server` `sql/planbuilder/select.go`): `FOR UPDATE` / `SKIP LOCKED` are parse-only no-ops ([the 2023 "hold my beer" post](https://www.dolthub.com/blog/2023-10-23-hold-my-beer/)), so concurrency safety for a claim queue can't come from pessimistic locking — it has to come from **retry on the 1213/1205 optimistic-merge conflict**. Beads already had that primitive (`withRetryTx`); the four queue hot paths (`claim`, `ready --claim`, `update`, `close`) just weren't using it. This wires them up.

No engine change needed for *this* fix. The open Dolt-side questions from the email still stand and aren't blocked by this PR:
- durability/crash semantics of `dolt_nonlocal_tables` `'immediate'` tables (where we'd park lease/heartbeat state, zero-commit)
- whether real row-level locking / `SKIP LOCKED` is on the roadmap (would let the claim queue avoid the retry-storm entirely)
- the FK-across-the-boundary nuance (engine tests show FK→nonlocal works; the `dolt_ignore` overlay is what forbids it)

Happy to feed back what we see running this under sustained multi-writer claim load once the shared-server work-queue is live.

timsehn:
@coffeegoddd is definitely the best person to review this. I'll see if I can grab him.

julianknutsen:
Thanks for this contribution, @steveyegge! This PR taught the Dolt `mutate + commit` paths to retry on serialization conflicts (the `1213` / `1205` errors) instead of surfacing them to callers, and it backed that up with the `TestConcurrentWorkQueueDrain` regression test plus the matching `testing.Short` allowlist entry. That's exactly the right hardening for the concurrent work-queue path.

While this was in the maintainer adoption queue, the same fix converged onto `main` through #4760 ("fix: add transaction retry with exponential backoff for serialization errors"), which merged on 2026-07-13. `main` now carries the equivalent `withRetryTx` handling in `internal/storage/dolt/issues.go` and a byte-identical `internal/storage/dolt/concurrent_test.go`. When we rebased this branch onto the current base, every commit dropped out as "patch contents already upstream" — the net diff against `main` is now empty.

### Original PR Review
The maintainer review did look at this change and found it sound: clean retry wrapping on the update / claim / ready-claim / close mutation paths, a solid concurrency regression test that checks for no surfaced claim errors, no double claims, and no stranded ready work, and the correct short-test allowlisting for that test. There were no gating issues — the change was accepted on its merits. The only reason there is nothing left to merge is that `main` already carries the same behavior.

### Disposition
I'm closing this as **superseded by #4760**, not because anything was wrong with it — your fix and #4760 landed on the same solution, and #4760 happened to merge first. Thank you again for chasing down the serialization-conflict handling; it's good to have two independent contributors agree on how this path should behave.

---
_Reviewed via the `/adopt-pr` workflow. Original contributor commits preserved on the branch; closing only to avoid a duplicate, conflicting merge surface for a change that is already on `main`._

PR #4493, opened 2026-06-25
feat(lease): claim-TTL + heartbeat + reclaim for dead-worker recovery

steveyegge:
## Summary

Gas Station v1.1 dead-worker recovery (wy-5r9j), stacked on #4491 (whose `withRetryTx` it depends on). A claim was previously **permanent** — a worker that died mid-task stranded its issue `in_progress` forever. This gives a claim a **lease** that can expire and be recovered.

## What's here

- **Migration 0054** — adds `lease_expires_at`, `heartbeat_at` (DATETIME) and `row_lock` (BIGINT) to `issues` (and `wisps`, for uniform `issueops` routing), plus `idx_issues_lease`. `assignee` doubles as the lease owner. CLI-bundle parity variant added.
- **Claim** stamps `lease_expires_at = now + TTL` (default 5m; `WithLeaseTTL` to override), `heartbeat_at = now`, fresh `row_lock`.
- **`bd heartbeat <id>`** — owner-only; pushes the lease forward, rewrites `row_lock`. Fails once the lease is gone so a preempted worker learns to stop.
- **`bd reclaim --older-than <dur>`** — reverts `in_progress` issues whose lease expired more than `<dur>` ago back to ready (clears assignee/started_at, records a `lease_reclaimed` event). Default grace `2×TTL`.
- Implemented in shared `issueops`, so both server-mode `DoltStore` and `EmbeddedDoltStore` get it. All lease paths are `withRetryTx`/`withConn` wrapped.

## The `row_lock` trick

Dolt has no row locking and merges concurrent commits **cell-by-cell**, so a heartbeat (writes `heartbeat_at`) racing a reclaim (writes `status`) would silently cell-merge into a *zombie* — an open/unassigned issue still carrying the worker's fresh heartbeat. Every mutating path (claim, close, update, heartbeat, reclaim) now rewrites the shared `row_lock` cell, forcing the loser to a 1213/1205 conflict that `withRetryTx` replays.

## Tests (real Dolt)

- `TestRowLockForcesConflictOnDisjointCellWrites` — proves the bug **without** `row_lock` (silent merge → zombie) and correctness **with** it (1213 conflict).
- `TestConcurrentHeartbeatReclaimClose` — integration race: live workers heartbeat+close vs a continuous reaper vs dead workers; asserts exactly-once, no lost close, no zombie.
- Claim/heartbeat/reclaim unit coverage on both backends; CLI smoke + docs regenerated.

Also de-couples `TestMigration0053PromotesRigWisps` from `LatestVersion()` (it assumed the rig-wisp repair was always the latest migration).

## Notes / follow-ups

- Lease columns are second-granular DATETIME (Dolt **rounds**), so sub-second TTLs are meaningless — production TTL is minutes.
- Heartbeat writes a Dolt commit; cadence must stay ≪ claim cadence to avoid history bloat (revisit `dolt_squash_history()` later if needed).
- Worker-loop wiring (heartbeat each iteration; supervisor cron `bd reclaim --older-than 2×TTL`) is the consumer-side follow-up (wy-s068), not in this PR.

### DIFF
--- a/cmd/bd/heartbeat.go
+++ b/cmd/bd/heartbeat.go
@@ -1,0 +1,93 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/steveyegge/beads/internal/metrics"
+	"github.com/steveyegge/beads/internal/ui"
+)
+
+var heartbeatCmd = &cobra.Command{
+	Use:     "heartbeat <id>",
+	Aliases: []string{"hb"},
+	GroupID: "issues",
+	Short:   "Refresh the lease on an issue you hold in_progress",
+	Long: `Refresh the lease on an issue you currently hold in_progress.
+
+A claim carries a lease that expires after a TTL. A worker keeps its claim alive
+by heartbeating faster than the TTL; once it stops (because it died), the lease
+goes stale and 'bd reclaim' reverts the issue to ready so another worker can pick
+it up. Heartbeat pushes lease_expires_at forward and stamps heartbeat_at = now.
+
+Only the current owner may heartbeat. If the lease has already been reclaimed or
+the issue closed, heartbeat fails so the worker learns to stop.
+
+Heartbeat writes a Dolt commit, so heartbeat well below the TTL but not so fast
+it bloats history — cadence should be a small fraction of the TTL, not per-op.
+
+Examples:
+  bd heartbeat bd-123
+  bd hb bd-123`,
+	Args:          cobra.ExactArgs(1),
+	SilenceUsage:  true,
+	SilenceErrors: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		CheckReadonly("heartbeat")
+
+		evt := metrics.NewCommandEvent("heartbeat")
+		defer func() {
+			if c := metrics.Global(); c != nil {
+				c.CloseEventAndAdd(evt)
+			}
+		}()
+
+		ctx := rootCtx
+		id := args[0]
+
+		result, err := resolveAndGetIssueForMutation(ctx, store, id)
+		if err != nil {
+			if result != nil {
+				result.Close()
+			}
+			return HandleErrorRespectJSON("resolving %s: %v", id, err)
+		}
+		if result == nil || result.Issue == nil {
+			if result != nil {
+				result.Close()
+			}
+			return HandleErrorRespectJSON("issue %s not found", id)
+		}
+		defer result.Close()
+
+		issueStore := result.Store
+		if err := issueStore.HeartbeatIssue(ctx, result.ResolvedID, actor); err != nil {
+			return HandleErrorRespectJSON("heartbeat %s: %v", result.ResolvedID, err)
+		}
+
+		if err := commitPendingIfEmbedded(ctx, issueStore, actor, doltAutoCommitParams{
+			Command:  "heartbeat",
+			IssueIDs: []string{result.ResolvedID},
+		}); err != nil {
+			return HandleErrorRespectJSON("failed to commit: %v", err)
+		}
+
+		SetLastTouchedID(result.ResolvedID)
+
+		if jsonOutput {
+			return outputJSON(map[string]string{
+				"id":     result.ResolvedID,
+				"status": "heartbeat",
+				"owner":  actor,
+			})
+		}
+		fmt.Printf("%s Heartbeat %s (lease refreshed)\n", ui.RenderPass("✓"), formatFeedbackID(result.ResolvedID, result.Issue.Title))
+		return nil
+	},
+}
+
+func init() {
+	heartbeatCmd.ValidArgsFunction = issueIDCompletion
+	rootCmd.AddCommand(heartbeatCmd)
+}
--- a/cmd/bd/reclaim.go
+++ b/cmd/bd/reclaim.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/steveyegge/beads/internal/metrics"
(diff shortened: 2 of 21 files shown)


maphew:
## Review

High-quality work overall — excellent docs/comments, a properly idempotent-guarded migration, an index that matches the reaper's scan predicate, and `TestRowLockForcesConflictOnDisjointCellWrites` is a genuinely good regression guard that proves both the bug *and* the fix against real Dolt. Notes below, with the architectural question first since it's the one I'd resolve before merge.

### Architectural fit — `row_lock` in shared `issueops`

The storage-layer roadmap is moving concurrency/locking/crash-safety behind the driver interface, and beads packages should avoid storage-engine-specific retry/crash-recovery. This PR puts a **Dolt-specific concurrency workaround (`row_lock`) directly into shared `issueops`** — `update.go`, `close.go`, `claim.go`, and the new `lease.go` all rewrite `row_lock` purely to defeat Dolt's cell-by-cell merge. That's beads core simulating a row lock the engine doesn't provide.

Worth separating the two halves:

- The **lease feature** (TTL / heartbeat / reclaim columns + semantics) is legitimately product-level and fits beads.
- The **`row_lock` merge-defeat mechanism** is storage-engine-specific correctness plumbing that may belong behind the driver interface, or at least warrants a sign-off from the storage-layer owner (@coffeegoddd) before it's baked into every mutating path in `issueops`.

### Correctness

1. **Backend divergence: heartbeating a wisp.** `DoltStore.HeartbeatIssue` guards with `isActiveWisp` and returns `ErrNotClaimable`, but `EmbeddedDoltStore.HeartbeatIssue` calls `issueops.HeartbeatIssueInTx` with **no wisp guard** — and that function routes to the wisp table via `WispTableRouting` and will update a wisp's lease columns. The same call errors on one backend and succeeds on the other. Suggest moving the guard into `HeartbeatIssueInTx` so both backends share it.

2. **`TestRowLockForcesConflictOnDisjointCellWrites` self-disables its assertion.** When the no-`row_lock` race doesn't silently merge, the test calls `t.Skipf`, which aborts the whole test — so the critical *with-`row_lock`* conflict assertion never runs. If Dolt's merge semantics ever shift, coverage of the actual fix vanishes exactly when you'd want it. Consider splitting into two tests, or logging the precondition miss and continuing to the with-lock assertion rather than skipping.

3. **Embedded path has no serialization retry.** The summary says all lease paths are `withRetryTx`/`withConn` wrapped, but `withConn` isn't a retry wrapper — only the server `DoltStore` paths get `withRetryTx`. For single-process embedded this is probably fine (no concurrent committers to merge), but then the `row_lock` writes on the embedded path are inert overhead. A one-line confirmation that embedded can't hit the cell-merge race would settle it; if it can, it needs retry too.

### Minor / non-blocking

- **Clock skew:** lease expiry compares `time.Now()` across whatever machines run claim/heartbeat vs. reclaim. The 2×TTL grace window mitigates and the docs acknowledge it, but multi-host correctness rests on loosely-synced clocks — worth being explicit this is single-host-ish today.
- **`reclaim` is unbounded:** it reverts all stale issues in one transaction/commit. Fine at expected scale; a large dead backlog becomes one big commit. A `--limit` could come later.
- **`row_lock` cost on the hot path:** every `UpdateIssue` now does a `crypto/rand` read plus an extra changed cell (more Dolt history churn) on all updates, not just leased ones. Cheap per-op, but it's now paid universally.
- **Stacked PR:** depends on #4491's `withRetryTx`; the conflict-replay correctness story relies on it landing first.

### Verdict

Solid, well-tested, well-documented feature. The only real cross-backend bug is the wisp-guard divergence (#1). The thing to resolve before landing is the architectural question — the `row_lock` storage workaround in shared `issueops` — which per the storage-driver roadmap should get an explicit nod from the storage-layer owner rather than merging as beads-core plumbing.

_claude-opus-4-8-unknown-reasoning on behalf of matt wilkie_

maphew (commented):
Reviewed at `ea2bd682`.

Thanks for the original lease primitive—its value has already landed. Merged #4537 explicitly carried this stack into `main`, retained Steve's authorship on the lease commit, and layered the maintainer fixes for backend parity, migration symmetry, JSON shape, update semantics, and test coverage. This PR now targets the obsolete base of closed #4491 and has no remaining integration value to merge.

Later CAS/ownership PRs (#4682, #4697, #4715, #5245) build on the shipped lease surface; they do not displace the first-filed contribution. Recommended disposition is **retire as carried by #4537**, with thanks and an explicit attribution note. I am not closing it in this review sweep.

maphew:
Thanks for filing this — the lease-primitive design here (claim-TTL, heartbeat, reclaim, schema v54) was solid, and its value has already landed on `main`.

#4537 merged your lease commit (`eb1a0cc7`, authorship retained: steveyegge + claude) on top of the retry-on-serialization work, and layered maintainer follow-up fixes over it: cross-backend parity for `HeartbeatIssue`'s wisp guard, migration/JSON-shape symmetry, and extra coverage for the fields this PR introduced.

This PR's base (`fix/claim-retry-on-serialization`, PR #4491) closed without merging, so #4493 is now stacked on an obsolete base with no remaining diff to contribute beyond what #4537 already carries forward.

I'll close this in a few days as retired / carried by #4537, with attribution preserved on the original commit. Later CAS/ownership work (#4682, #4697, #4715, #5245) builds on the lease surface this PR introduced — it extends rather than replaces this contribution.

Thanks again for the diagnosis and the test coverage (`TestRowLockForcesConflictOnDisjointCellWrites` proved real value here).

maphew:
Retired / carried by #4537 (merged 2026-07-04 with authorship preserved); base PR #4491 closed unmerged (disposition comment 2026-08-04)

PR #4500, opened 2026-06-27
fix(gate): scope `bd gate list <issue-id>` to that issue's gates

itsandyking:
## Problem

`bd gate list` takes no positional argument, so `bd gate list <issue-id>` **silently ignores** the argument and returns every gate in the database — identical to the no-arg `bd gate list`. The `<issue-id>` looks bead-scoped but isn't.

This is a footgun for anything that trusts it to be scoped. In our case a janitor that listed "the gates on my lane" via `bd gate list <conductor-bead>` got the whole database back (~50 ambient gates from unrelated work) and a "purge stray gates" loop nearly mass-resolved other work's gates (caught before any damage).

### Reproduce (before)
```
bd gate create --blocks A --type gh:pr --await-id 111
bd gate create --blocks B --type gh:pr --await-id 222
bd gate list A      # returns BOTH 111 and 222 — the "A" arg is ignored
```

## Fix

Give `bd gate list` an optional `[issue-id]` argument. When provided, it lists **only the gates that block that issue** — the gate-type issues in its own dependency set (`store.GetDependencies`) — honoring `--all` and `--limit` exactly like the DB-wide path. No argument preserves the existing DB-wide behavior. An unknown id returns a clean `issue not found` error instead of a DB-wide dump.

The gate→issue relationship is the same edge `bd gate create` writes (`{IssueID: blocked, DependsOnID: gate, Type: blocks}`), so `GetDependencies(blocked)` is exactly the set of that issue's gates.

### After
```
bd gate list A      # ⏳ Open Gates (1):  ○ ... gh:pr 111
bd gate list B      # ⏳ Open Gates (1):  ○ ... gh:pr 222
bd gate list        # ⏳ Open Gates (2):  111 + 222   (unchanged)
bd gate list bad-id # Error: issue not found: bad-id
```

## Tests

- **`TestFilterIssueGates`** (unit) — the scoping/filter helper: only gate-type deps returned, `--all` controls closed visibility, `--limit` honored, nil-safe.
- **`TestEmbeddedGateListScoped`** (integration) — two gated issues; the scoped list returns only the named issue's gate and never the other's; DB-wide still lists both; unknown id errors.

Full gate suite green locally (`go test -tags gms_pure_go -run Gate ./cmd/bd` with `BEADS_TEST_EMBEDDED_DOLT=1`), `go vet` clean, `gofmt` applied.

Backward compatible: no behavior change for `bd gate list` / `--all` / `--limit` with no positional argument.


maphew (commented):
Thanks for tracking down and fixing this — the repro in the description (janitor purge loop nearly acting on unrelated gates) is a great illustration of why the silent argument-ignore was dangerous.

Reviewed the diff against `cmd/bd/gate.go` on main:

- The scoped path correctly reuses `store.GetDependencies(target.ID)`, which is exactly the edge set `bd gate create` writes (`{IssueID: blocked, DependsOnID: gate, Type: blocks}`), so scoping by dependency + gate-type filter is sound rather than guessed.
- `--all` and `--limit` are honored identically to the DB-wide path.
- The "issue not found" error on an unknown id matches the existing convention already used elsewhere in this file (e.g. the `blocksID` not-found case), so no new inconsistency introduced.
- Tests are solid: `TestFilterIssueGates` unit-tests the extracted helper (non-gate deps excluded, nil-safety, --all/--limit), and `TestEmbeddedGateListScoped` is a real integration regression for the reported leak, asserting the scoped list never contains the other issue's gate.

One minor, non-blocking observation: `filterIssueGates` truncates in the order `GetDependencies` returns rows, which has no `ORDER BY` in the underlying query — so `--limit` on the scoped path could return a different subset than the DB-wide path's ordering in edge cases with many gates on one issue. Not worth blocking on; happy to see this land as-is or with a trivial follow-up if you want deterministic ordering.

Recommendation: merge as-is. Nice, well-tested, backward-compatible fix for a real safety hole.

maphew:
Maintainer push to unblock this: merged current `main` into the branch and regenerated the CLI docs (`docs/CLI_REFERENCE.md`, `website/docs/cli-reference/gate.md`) so the doc-freshness gate can pass — the help-text change here requires committed regenerated docs. No changes to your code; your commit is intact.

The scoping fix itself reviewed well (real safety issue — the unscoped list nearly caused a mass gate-resolve; the unit + embedded-dolt regression coverage is appreciated). Two nits noted for a possible follow-up, neither blocking: `filterIssueGates` truncates in `GetDependencies` row order so `--limit` ordering differs from the DB-wide path, and an empty scoped result JSON-encodes as `null` where the DB-wide path emits `[]`.

Once checks go green on the refreshed head this can merge. Thanks itsandyking!

maphew (approved):
Approving: the scoping fix reviewed clean (real safety issue, good regression coverage — see earlier comment for the two optional nits), the doc-freshness blocker is resolved by the maintainer docs-regen commit, and checks are green on a branch refreshed onto current main (including the #4744 per-step-commit fix).

PR #4504, opened 2026-06-28
fix(schema): make CALL-bearing migrations idempotent and drain all proc result sets

Rome-1:
## Summary

A fresh `bd init` against a shared Dolt `sql-server` under concurrent load could permanently brick the database at migration `0040`. Two compounding causes:

1. **Undrained procedure result set.** A `CALL DOLT_*` result set left undrained on the pinned migration connection (`busy buffer` → `driver: bad connection`) then failed the *separate* version-record `INSERT` (and the lock `RELEASE`) that follow it, so `schema_migrations` never advanced past the partially-applied migration.
2. **Non-idempotent migration.** `0040` used a bare `INSERT INTO dolt_nonlocal_tables`, so when the init retry loop re-ran it, the replay died on `duplicate primary key given: [wisps]` (or `nothing to commit`). Every subsequent open re-ran `0040` and failed the same way.

Draining migration bodies alone was necessary but not sufficient: a load-induced transient (or a no-op commit) could still re-enter a non-idempotent migration.

## Changes

- **`drainCall()`** — run every `CALL DOLT_*` site (migration bodies, apply-migrations `DOLT_COMMIT`, `DOLT_RESET`, `DOLT_ADD`) via `QueryContext` and consume all result sets, leaving the pinned connection clean for the next statement. `GET_LOCK`/`RELEASE_LOCK` stay as single-result `SELECT`s.
- **`0040`** — `INSERT IGNORE` + `DOLT_COMMIT('--skip-empty')` so a replay is a clean no-op.
- **`0041`** — `DOLT_COMMIT('--skip-empty')` on the leading commit (its `DELETE` is already idempotent; the remainder is `@needs_migrate`-guarded).

Migration SQL edits preserve `content_hash` semantics.

## Testing

- `migration_body_drain_test.go` — guards that every `dolt_nonlocal_tables` migration uses `INSERT IGNORE` and `--skip-empty`.
- `initschema_idempotent_test.go` (integration) — forges the partial-`0040` brick state on a real `sql-server` and asserts re-init recovers to the latest schema version.
- Rebased on current `main`; `go build`, `go vet`, and the full `internal/storage/schema` suite pass (existing `preMigrationRepair` coverage runs alongside the new drain/idempotency tests).

### DIFF
--- a/internal/storage/dolt/initschema_idempotent_test.go
+++ b/internal/storage/dolt/initschema_idempotent_test.go
@@ -1,0 +1,183 @@
+//go:build integration && !windows
+
+package dolt
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+	"time"
+
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/steveyegge/beads/internal/storage/doltutil"
+	"github.com/steveyegge/beads/internal/storage/schema"
+)
+
+// TestSchemaInitRecoversFromPartialNonlocalMigration reproduces, against a real
+// shared dolt sql-server, the brick that bricked freshly-init'd rigs: migration
+// 0040 partially applied (its CALL DOLT_COMMIT committed a dolt_nonlocal_tables
+// row) but a transient ("busy buffer" -> "driver: bad connection") prevented its
+// schema_migrations version row from recording, so the init retry loop re-ran
+// 0040 from the top forever.
+//
+// Before the fix the replay died on "duplicate primary key given: [wisps]"
+// (0040's bare INSERT) — or "nothing to commit" (its DOLT_COMMIT after a no-op
+// INSERT). The fix leaves 0040/0041's shipped, content-hashed bytes untouched
+// (editing them would fork already-upgraded clones via schema_migrations
+// .content_hash — see scripts/check-migration-hygiene.sh check C). Instead it
+// (a) drains every CALL DOLT_* result set so the transient no longer arises
+// (drainCall in internal/storage/schema/schema.go), and (b) registers
+// pre-migration repairs keyed to versions 40 and 41 that normalise
+// dolt_nonlocal_tables before each frozen body replays
+// (internal/storage/schema/migration_repairs.go). Re-init is therefore
+// idempotent and recovers to the latest schema version.
+//
+// Would have caught the bd-1.0.x/1.1.0-rc.1 shared-server init brick
+// (ahdr-ovz / hq-oi0db).
+func TestSchemaInitRecoversFromPartialNonlocalMigration(t *testing.T) {
+	skipIfNoDolt(t)
+	acquireTestSlot()
+	t.Cleanup(releaseTestSlot)
+
+	if testServerPort == 0 {
+		t.Skip("no Dolt test server available")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	rootDSN := doltutil.ServerDSN{Host: "127.0.0.1", Port: testServerPort, User: "root"}.String()
+	rootDB, err := sql.Open("mysql", rootDSN)
+	if err != nil {
+		t.Fatalf("open root connection: %v", err)
+	}
+	defer rootDB.Close()
+
+	dbName := uniqueTestDBName(t)
+	if _, err := rootDB.ExecContext(ctx, "CREATE DATABASE IF NOT EXISTS `"+dbName+"`"); err != nil {
+		t.Fatalf("create database: %v", err)
+	}
+	t.Cleanup(func() {
+		// Skip DROP — rapid drop cycles can crash the Dolt container.
+	})
+
+	dbDSN := doltutil.ServerDSN{Host: "127.0.0.1", Port: testServerPort, User: "root", Database: dbName}.String()
+	db, err := sql.Open("mysql", dbDSN)
+	if err != nil {
+		t.Fatalf("open db connection: %v", err)
+	}
+	defer db.Close()
+
+	// 1. Fully initialize to the latest schema.
+	if _, err := initSchemaOnDB(ctx, db); err != nil {
+		t.Fatalf("initial initSchemaOnDB: %v", err)
+	}
+	latest := schema.LatestVersion()
+	assertSchemaVersionAtLeast(ctx, t, db, latest)
+
+	// 2. Forge the brick: rewind the cursor below 0040 and leave a *committed*
+	//    dolt_nonlocal_tables row that 0040 inserts. This is exactly the
+	//    partially-applied state a transient leaves behind — 0040's CALL
+	//    DOLT_COMMIT committed the side effect, but the schema_migrations version
+	//    row never recorded. The row must be committed (not just staged): a real
+	//    partial 0040 committed it, and an uncommitted row would instead trip the
+	//    separate "pending migration alters a dirty table" guard
+	//    (gastownhall/beads#4566) before the replay is ever reached.
+	if _, err := db.ExecContext(ctx, "DELETE FROM schema_migrations WHERE version >= 40"); err != nil {
+		t.Fatalf("rewind schema_migrations: %v", err)
+	}
+	if _, err := db.ExecContext(ctx,
+		"INSERT IGNORE INTO dolt_nonlocal_tables (table_name, target_ref, options) VALUES ('wisps', 'main', 'immediate')"); err != nil {
+		t.Fatalf("re-add nonlocal row: %v", err)
+	}
+	if _, err := db.ExecContext(ctx,
+		"CALL DOLT_COMMIT('-Am', 'forge partial 0040: committed nonlocal row, version row absent')"); err != nil {
+		t.Fatalf("commit forged partial state: %v", err)
+	}
+
+	// 3. Re-init. Pre-fix this died replaying 0040's non-idempotent INSERT
+	//    ("duplicate primary key given: [wisps]"); post-fix the version-40
+	//    pre-migration repair clears the committed stray row so the frozen 0040
+	//    replays cleanly and the chain reaches the latest version.
+	if _, err := initSchemaOnDB(ctx, db); err != nil {
+		t.Fatalf("re-init after forged partial migration (this is the brick): %v", err)
+	}
+	assertSchemaVersionAtLeast(ctx, t, db, latest)
+}
+
+// TestSchemaInitRecoversFromPartial0041NonlocalDelete covers the second frozen
+// non-idempotent migration. 0041 opens by clearing dolt_nonlocal_tables and
+// committing; if a transient interrupts it after that commit but before its
+// version row records, the retry re-runs 0041 against an already-empty table and
+// the shipped DELETE+COMMIT (no --skip-empty) dies with "nothing to commit". The
+// version-41 pre-migration repair restores the four rows 0040 leaves so the
+// frozen DELETE has a real diff again, and re-init recovers.
+func TestSchemaInitRecoversFromPartial0041NonlocalDelete(t *testing.T) {
+	skipIfNoDolt(t)
+	acquireTestSlot()
+	t.Cleanup(releaseTestSlot)
+
+	if testServerPort == 0 {
+		t.Skip("no Dolt test server available")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	rootDSN := doltutil.ServerDSN{Host: "127.0.0.1", Port: testServerPort, User: "root"}.String()
+	rootDB, err := sql.Open("mysql", rootDSN)
+	if err != nil {
+		t.Fatalf("open root connection: %v", err)
+	}
+	defer rootDB.Close()
+
+	dbName := uniqueTestDBName(t)
+	if _, err := rootDB.ExecContext(ctx, "CREATE DATABASE IF NOT EXISTS `"+dbName+"`"); err != nil {
+		t.Fatalf("create database: %v", err)
+	}
+
+	dbDSN := doltutil.ServerDSN{Host: "127.0.0.1", Port: testServerPort, User: "root", Database: dbName}.String()
+	db, err := sql.Open("mysql", dbDSN)
+	if err != nil {
+		t.Fatalf("open db connection: %v", err)
+	}
+	defer db.Close()
+
+	// 1. Fully initialize. After 0041 runs, dolt_nonlocal_tables is empty
+	//    (0040 added its four rows; 0041 deleted them all) — the exact table
+	//    state a partial 0041 leaves behind.
+	if _, err := initSchemaOnDB(ctx, db); err != nil {
+		t.Fatalf("initial initSchemaOnDB: %v", err)
+	}
+	latest := schema.LatestVersion()
+	assertSchemaVersionAtLeast(ctx, t, db, latest)
+
+	// 2. Forge the partial-0041 brick: rewind the cursor to 40 (0040 recorded,
+	//    0041 not) with dolt_nonlocal_tables already empty and committed.
+	if _, err := db.ExecContext(ctx, "DELETE FROM schema_migrations WHERE version >= 41"); err != nil {
+		t.Fatalf("rewind schema_migrations to 40: %v", err)
+	}
+	if _, err := db.ExecContext(ctx,
+		"CALL DOLT_COMMIT('-Am', 'forge partial 0041: cursor at 40, nonlocal table empty', '--skip-empty')"); err != nil {
+		t.Fatalf("commit forged partial state: %v", err)
+	}
+
+	// 3. Re-init. Pre-fix this died replaying 0041's DELETE over an empty table
+	//    ("nothing to commit"); post-fix the version-41 repair restores the rows
+	//    so the frozen DELETE+COMMIT has a diff and the chain reaches latest.
+	if _, err := initSchemaOnDB(ctx, db); err != nil {
+		t.Fatalf("re-init after forged partial 0041 (this is the brick): %v", err)
+	}
+	assertSchemaVersionAtLeast(ctx, t, db, latest)
+}
+
+func assertSchemaVersionAtLeast(ctx context.Context, t *testing.T, db *sql.DB, want int) {
+	t.Helper()
+	var got int
+	if err := db.QueryRowContext(ctx, "SELECT COALESCE(MAX(version), 0) FROM schema_migrations").Scan(&got); err != nil {
+		t.Fatalf("read schema version: %v", err)
+	}
+	if got < want {
+		t.Fatalf("schema version = %d, want >= %d", got, want)
+	}
+}
--- a/internal/storage/schema/lock.go
+++ b/internal/storage/schema/lock.go
@@ -101,7 +101,10 @@ func MigrateUpWithLock(ctx context.Context, conn *sql.Conn, databaseName string,
 		// retry can never converge (gastownhall/beads#5012).
 		fmt.Fprintf(stderr, "Discarding interrupted-bootstrap working set (%s) and re-running migrations…\n",
 			strings.Join(dirtyErr.Tables, ", "))
-		if _, resetErr := conn.ExecContext(ctx, "CALL DOLT_RESET('--hard')"); resetErr != nil {
+		// Drained, not Exec'd: the very next thing this path does is re-run the
+		// whole MigrateUp pass on this same pinned connection, so an
+		// undrained proc result set here would poison every statement of it.
+		if resetErr := drainCall(ctx, conn, "CALL DOLT_RESET('--hard')"); resetErr != nil {
 			return applied, errors.Join(err, fmt.Errorf("schema: fresh-bootstrap reset: %w", resetErr))
 		}
 		applied, err = MigrateUp(ctx, conn)
--- a/internal/storage/schema/lock_test.go
+++ b/internal/storage/schema/lock_test.go
@@ -120,10 +120,10 @@ func TestMigrateUpSeedsIgnorePatternsWhenNoWorkNeeded(t *testing.T) {
 	expectScalar(mock, "SELECT COUNT(*) FROM custom_statuses", "count", 1)
 	// The seed inserted rows and no migration pass follows to commit them, so
 	// MigrateUp must commit the heal itself, scoped and labeled.
-	mock.ExpectExec(regexp.QuoteMeta("CALL DOLT_ADD('dolt_ignore')")).
-		WillReturnResult(sqlmock.NewResult(0, 0))
-	mock.ExpectExec(regexp.QuoteMeta("CALL DOLT_COMMIT('-m', 'schema: seed dolt_ignore patterns')")).
-		WillReturnResult(sqlmock.NewResult(0, 0))
+	mock.ExpectQuery(regexp.QuoteMeta("CALL DOLT_ADD('dolt_ignore')")).
+		WillReturnRows(sqlmock.NewRows([]string{"status"}))
+	mock.ExpectQuery(regexp.QuoteMeta("CALL DOLT_COMMIT('-m', 'schema: seed dolt_ignore patterns')")).
+		WillReturnRows(sqlmock.NewRows([]string{"hash"}))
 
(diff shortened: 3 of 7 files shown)


kevglynn:

Ran a complexity-only review on this diff. The core fix — `execMigrationBody`, `procedureCallRe`, and the two test functions — is ~110 lines, lean, and precisely targeted.

`scripts/repro-dolt-migration-commit/` (INCIDENT-REPORT.md + main.go): **yagni:** 347 lines of investigation artifact. The incident report restates the PR description and the code comment on `execMigrationBody`; the standalone repro harness requires a `dolt` binary, won't run in CI, and duplicates the classification logic already pinned by `migration_body_drain_test.go`. The `isTransient()` function in the repro also duplicates the retry-classification logic from `initSchemaOnDBWithRetry` in the store layer.

**net: -347 lines possible** by deleting the `scripts/repro-dolt-migration-commit/` directory. The PR body is the incident report; the test file is the repro.

Not a blocker — the fix itself is clean. Just suggesting the lab notebook doesn't need to ship.

Rome-1:
Thanks for the over-engineering pass — agree the fix itself should stay lean.

Keeping `scripts/repro-dolt-migration-commit/` deliberately, mainly for consistency with the harnesses already in `scripts/` (`repro-dolt-hang`, `repro-dolt-prod-timeouts`): same shape (an INCIDENT-REPORT plus a `go run` harness) and the same "needs a live `dolt`, so it doesn't run in CI" tradeoff. If those belong in the tree, this one does too.

On the overlap with `migration_body_drain_test.go`: that unit test only pins the *classification* — which migration bodies take the draining path — and guards it so a future `CALL`-bearing migration is covered automatically. It deliberately doesn't (can't) reproduce the failure, which is a load/timing race on a shared sql-server. The harness runs the real `schema.MigrateUpWithLock` path concurrently across N fresh DBs and exits non-zero if any commits a migration without recording its version row. The test and the harness cover different layers — classification vs. end-to-end reproduction.

On the INCIDENT-REPORT overlapping the PR body: that's intentional — the PR description is ephemeral, the in-tree report is the durable record for the next person who hits an intermittent migration brick. And `isTransient` mirrors `initSchemaOnDBWithRetry` on purpose, so the harness reaches the `CALL` migrations under contention instead of bailing on a transient startup race; it's a standalone `main` package, so it can't import the unexported store helper.

Rome-1:
Update — revised this PR after finding the original fix was **necessary but not sufficient**, which also changes my earlier answer on the repro.

**On correctness:** draining the migration *body* alone did not actually stop the brick on a real shared sql-server under load. The deeper cause is that migration 0040 is non-idempotent (`INSERT INTO dolt_nonlocal_tables ...`), and the init retry loop re-runs the whole migration after *any* transient — so a load-induced "busy buffer"/"bad connection" (or a no-op "nothing to commit") still re-entered 0040 and died on `duplicate primary key given: [wisps]`. Body-draining couldn't prevent that.

The fix now does both halves:
- `drainCall` consumes the result set of **every** `CALL DOLT_*` site (migration bodies, the apply-migrations `DOLT_COMMIT`, `DOLT_RESET`, `DOLT_ADD`) — not just bodies — so the pinned connection is never left busy for the version-record `INSERT` or the lock release.
- 0040/0041 are made **idempotent**: `INSERT IGNORE` + `DOLT_COMMIT('--skip-empty')`, so a retry replay is a clean no-op instead of a dup-key/"nothing to commit" failure.

**On the repro:** you were right — I've **deleted** `scripts/repro-dolt-migration-commit/`. It was worse than redundant: it passed while the real `bd init` path failed (it didn't exercise the retry loop against a loaded shared server), so it was a false green. Replaced with CI-runnable tests:
- `migration_body_drain_test.go` now also guards that every `dolt_nonlocal_tables` migration uses `INSERT IGNORE` + `--skip-empty` (tracks the embedded migrations, not a hard-coded list).
- `initschema_idempotent_test.go` (integration) forges the partial-0040 brick state on a real sql-server and asserts re-init recovers to the latest version.

Validated end-to-end by taking a database that had actually hit this brick in production (stuck below 0040 with the nonlocal row already committed) from bricked → latest schema version with the patched binary, on a live shared server.

maphew (commented):
Thanks for chasing this down to the real root cause, and for pulling the repro harness once it turned out to be a false green - that's exactly the kind of course-correction that makes a fix trustworthy. The core change is lean and well targeted:

- `drainCall()` consistently consuming every `CALL DOLT_*` result set on the pinned migration connection is the right fix for the "busy buffer" -> "bad connection" symptom.
- Making 0040/0041 idempotent (`INSERT IGNORE` + `DOLT_COMMIT('--skip-empty')`) closes the second half - a load-induced transient can no longer re-enter a non-idempotent migration and dup-key.
- `migration_body_drain_test.go` pinning the classification regex against boundary cases, and guarding that every `dolt_nonlocal_tables` migration is idempotent going forward, is good defensive test design - it tracks the embedded migrations rather than a hard-coded list.
- `initschema_idempotent_test.go` forging the exact partial-0040 brick state on a real sql-server is a strong integration test for a hard-to-reproduce load/timing race.

One thing worth resolving before merge: rewriting the SQL text of migrations 0040/0041 changes their recorded `content_hash` for every future application (`schema.go` hashes the file bytes and stores it per-version at apply time). Those two migrations already shipped in v1.0.5 and v1.1.0, so plenty of production databases already have the *old* hash recorded for versions 40/41. The `content_hash` mechanism exists specifically so `ContentHashSkew` (used by the #4516 smart remote-migrate gate) can detect "same version, different content" as a genuine schema fork and block with a human data-loss decision (`smartForkSkew`). `ReadMigrationContentHashes` selects the full history from `schema_migrations`, not just pending versions, so this comparison isn't limited to fresh inits.

To be fair about reachability: hashes sync with the Dolt history, so this only fires when one clone of the same history applies the new text while another applied the old - and the smart gate's convergence floor (`LastNonDeterministicMigration = 43`) means 40/41 are never auto-applied on remote-backed databases. But the divergence paths are real (`BD_ALLOW_REMOTE_MIGRATE=1`, `BD_SMART_GATE=0`, a clone migrated before it starts comparing against the remote), and when they happen the mismatch surfaces in three places as a genuine fork: `smartForkSkew` blocks, `bd doctor` reports Migration Content Skew, and the merge-settle path refuses to auto-resolve two different non-empty hashes on the same `schema_migrations` row - all for a textual change that produces identical table state.

Could you confirm whether this interaction was considered? A couple of ways to sidestep it if not:
- Ship the idempotency fix as a *new* migration (e.g. 0044) that corrects the earlier state, rather than mutating 0040/0041's already-released bytes in place, or
- Special-case 0040/0041 in the skew comparison (a small "known-benign hash pairs" allowlist, similar in spirit to `nondeterminism-allowlist.txt`).

Also noticed CI hasn't run at all on the current head commit (no check-runs, no combined status) - not sure if that's a repo-side gating issue for external contributors or something else, but worth checking before merge since `initschema_idempotent_test.go` is integration-tagged and needs a live sql-server to actually exercise.

None of this is a knock on the core fix, which should land - just want the content-hash/fork-skew interaction resolved or explicitly accepted first, since it's the same class of "silent schema fork" bug this PR is otherwise so careful about.

Rome-1:
Thanks for catching this — you're right that it's the same silent-schema-fork class this PR is otherwise careful about, and it deserved a real resolution rather than an accept. It was considered; the interaction is now avoided entirely by **not changing the shipped bytes of 0040/0041 at all**, so there is no content_hash change and nothing for `ContentHashSkew` / `smartForkSkew` / `bd doctor` / merge-settle to (falsely) flag.

**Why not a new migration (option 1).** A database bricked at 40 never reaches a later version, so a forward migration can't heal it — the failing file aborts the pass before anything after it runs. That's exactly the doctrine already written down in `internal/storage/schema/migration_repairs.go`:

> a bug that makes a migration FAIL on drifted databases cannot be fixed forward with a new migration either — the failing file aborts the pass before any later version runs. Repairing the drift in code, keyed to the pending version, is the only path that heals affected databases without touching shipped SQL.

(0044 is also already taken on current `main` — migrations are at 0054 now.)

**Why not the skew allowlist (option 2).** It works, but the byte change to 0040/0041 itself trips `scripts/check-migration-hygiene.sh` check C (frozen migrations), and a "known-benign hash pairs" exception would have to live in *two* independent places — `schema.ContentHashSkew` (smart gate + `bd doctor`) and the separate SQL comparison in `versioncontrolops/mergesettle.go` — permanently widening the fork detector's benign-exception surface for a change we don't actually need to make.

**What landed instead.** The drain half is unchanged (`drainCall()` consuming every `CALL DOLT_*` result set — that's pure Go, no migration-content impact). The idempotency half is delivered as **pre-migration repairs keyed to versions 40 and 41** (the established `migration_repairs.go` pattern, alongside the existing v53 repairs) that normalise `dolt_nonlocal_tables` right before each frozen body replays:

- before 0040: clear any of its four rows a partial apply committed, so the bare `INSERT` no longer duplicate-keys;
- before 0041: restore those four rows if a partial apply already deleted them, so the frozen `DELETE`+`DOLT_COMMIT` has a real diff instead of "nothing to commit".

Both are strict no-ops on the common path. 0040/0041's released bytes — and their recorded content hashes — are untouched, so the divergence paths you listed (`BD_ALLOW_REMOTE_MIGRATE=1`, `BD_SMART_GATE=0`, a clone that migrated before comparing) simply can't produce a v40/v41 hash mismatch anymore.

**On CI / the integration test.** Good call — running it actually surfaced a bug in the test itself: it forged the stray `dolt_nonlocal_tables` row *uncommitted*, which trips the separate "pending migration alters a dirty table" guard (#4566) rather than the real dup-key brick (a genuine partial 0040 has already *committed* its row via `CALL DOLT_COMMIT`). Fixed to forge the committed state, and added a second case for the partial-0041 "nothing to commit" path. Both now pass against a real `sql-server` (`go test -tags integration ./internal/storage/dolt/...`), along with the full `internal/storage/schema` suite.

kevglynn:
Verified the revised approach against the codebase — this resolves the content_hash/fork-skew concern:

- 0040/0041's shipped bytes are untouched in this diff (only `schema.go`, `migration_repairs.go`, and tests changed) — no `content_hash` change, so `smartForkSkew`, `bd doctor`'s Migration Content Skew check, and merge-settle's hash comparison have nothing new to flag.
- The keyed `preMigrationRepairs` registry follows the exact precedent already in `migration_repairs.go` (the v53 repair), including its own doctrine comment about why forward-fix migrations can't heal a bricked database — this isn't a new pattern invented for this PR.
- `LastNonDeterministicMigration = 43` checks out in `smart_remote_migrate_gate.go`, confirming the convergence-floor claim.
- `TestFrozenNonIdempotentMigrationsHaveRepairs` and `TestMigrationsWithProcedureCallsTakeDrainPath` scan the real embedded migrations rather than a hardcoded list, so this stays correct as new migrations are added.
- CI is fully green on the current head (`d285748`).

This touches the schema/migration path, so per the merge-discipline rules a fresh human approval on the current commit is needed before merge — @maphew's review predates this revision (posted against `f4df12e`, before the pivot to pre-migration repairs). Looks ready for a re-look.

_cursor-sonnet-5-unknown-reasoning on behalf of kevglynn_

maphew:
Closing the loop on the July 5 concern — your revised approach resolves it, and resolves it the right way: by not changing the shipped bytes of 0040/0041 at all, there is no `content_hash` delta for `smartForkSkew`, doctor's content-skew check, or merge-settle to false-flag, and the keyed `preMigrationRepairs` registry follows the existing v53 precedent instead of inventing a parallel mechanism. kevglynn's July 11 independent verification checked exactly the load-bearing claims and they hold; with that plus the diff scope (schema.go, migration_repairs.go, tests only), this is approve-in-principle from our side. Sorry it sat two weeks after his verification — that wait was ours, not yours.

One mechanical gate and one sequencing note:

1. The branch now conflicts with main in the schema package (three files — main's migration machinery moved a lot this month: numbered migrations through 0059, ignored through 0016, and the #5012 bootstrap-heal changes). A rebase is needed before CI can speak. If you'd rather not chase it, say so and we'll do the rebase maintainer-side on your branch with your commits intact.
2. Sequencing: #4804 (migration-lock fast path) is also approve-pending-rebase and touches `migrateUp`/`schema.go` too. Whichever lands second rebases over the first; no design interaction expected (your repair runs inside the locked pass, which is inside #4804's sentinel-revoked interval — exactly where it should be), but flagging it so a second conflict doesn't surprise you.

maphew (commented):
Thanks for this — you independently diagnosed the busy-buffer poisoning chain that we later hit again in #5012, and #5042 (retry convergence) explicitly builds on your source-side diagnosis. We did a deep review at head d285748 against current main. Verdict: we want the drain pattern, and the diff needs a rebase plus some rework because main has moved substantially (19 commits in `internal/storage/schema/` since your base). Details below; **offer at the end.**

## The diff splits cleanly in two

**Part A — drain pattern** (`drainCall`, `execMigrationBody`, the 3 call-site conversions, classifier + tests): mechanically self-contained, auto-merges clean, tests pass at your head (`go test ./internal/storage/schema/...` green). We want this.

**Part B — 0040/0041 idempotency repairs** (`repairKey` registry, the two nonlocal repairs, integration tests): independently defensible but needs separate consideration (below). The `repairKey` map refactor carries nearly all the merge conflicts — main rewrote the same dispatch switch twice (added `case 47` and extended `case 53`).

## Part A findings

1. **Mechanism doc-comment needs correcting.** go-sql-driver's `mysqlConn.exec` does run `discardResults()` on the happy path — the real asymmetry is the **error path**: `exec` returns early without discarding when a mid-stream statement errors (e.g. `nothing to commit` on statement 4 of 0040's 8, with more results queued), while `rows.Close()` always drains. That's a solid argument for `execMigrationBody` on multi-statement CALL-bearing bodies; it's not an argument for the three single-statement conversions. Suggest rewriting the comment to the error-path argument.
2. **The conversion is now incomplete — and the gaps are the hot sites.** Since your base, main added more `CALL DOLT_*` Exec sites on the same pinned migration connection: `commitMigrationStep` (per-step `DOLT_ADD`/`DOLT_COMMIT` — N× more CALL traffic than everything the PR converts), `commitSeededDoltIgnore`, and #5042's `DOLT_RESET('--hard')` in `lock.go` (which is immediately followed by a full `MigrateUp` on that same conn — the strongest candidate in the codebase on your own theory). A rebased part A should extend to those.
3. **Rebase trap (would-be red merge):** main's `TestRunMigrationsLargeRigNoticeOnlyOnMainSource` calls `runMigrations` with bound 46 and sits *outside* every conflict region, so git merges it silently — post-merge it reaches 40/41, routes through `drainCall`/`anyNonlocalFrozenRowPresent`, and panics on the `mockDB` stubs. Both bounds need lowering to 39, exactly as you did for the two tests you knew about.

## Part B findings

- Re #4800 (dhruv-pratap reproduced `nothing to commit` on dolt 2.1.10 server mode with a build of this PR): the repairs are a strict **no-op** there — `anyNonlocalFrozenRowPresent` short-circuits on a pristine store — so B is *insufficient* for that failure, but not *wrong*. And B covers a real hole nothing on main covers: committed 0040 rows with no cursor row (death in the self-commit window) replay into `duplicate primary key`, which is not a `*DirtyTablesError`, so #5042's fresh-bootstrap heal never fires — and couldn't help anyway since the rows are committed, not working-set.
- One scoping fix needed: the repairs commit with `DOLT_COMMIT('-Am', ...)`, which on the bounded-migrate path (`upTo != 0`, where per-step commit is off) would sweep migrations 1–39's uncommitted working set into a repair-labelled commit. Scope the stage explicitly or gate the repair on `commitEachStep`.
- `TestFrozenNonIdempotentMigrationsHaveRepairs` is the best thing in the diff — deriving the needs-repair set from the migration bytes so future non-idempotent migrations fail the build. We'd keep that regardless.

## Offer

Want to take these (one PR, rebased, part A extended to the new call sites, both mock bounds at 39)? Or shall we land it maintainer-side in two commits with `Co-authored-by` attribution — part A+ first, part B after settling the scoping question? Happy either way; you've been responsive and this diagnosis was ahead of us by a week.

Rome-1:
Taking these — rebased onto current main and reworked in one commit (`b2918fa`), rather than splitting maintainer-side. Point by point:

## Part A

**1. Mechanism doc-comment — corrected, and you're right.** Confirmed against the pinned driver (`go-sql-driver/mysql v1.10.0`): `mysqlConn.exec` ends in `return handleOk.discardResults()`, so the happy path was never the gap and the old comment was wrong to imply it was. The asymmetry is the error path — `exec` returns early *without* discarding when `readResultSetHeaderPacket` / `skipColumns` / `skipRows` fails mid-stream (`connection.go:466-483`), while `mysqlRows.Close()` skips unread rows and discards remaining result sets unconditionally (`rows.go:100-120`). The comment now argues that instead.

Worth adding to your framing: the asymmetry is reachable here *specifically* because these sites tolerate an error and keep using the same pinned connection — `MigrateUp` and `commitMigrationStep` both swallow `nothing to commit` and carry on. That's what turns "early return without discard" into a poisoned connection rather than a fatal error, and it's why the single-statement `CALL` sites are worth converting too, not just the multi-statement bodies.

**2. Extended to the new call sites.** `commitMigrationStep` (per-step `DOLT_ADD`/`DOLT_COMMIT`), `commitSeededDoltIgnore`, and #5042's `DOLT_RESET('--hard')` in `lock.go` all drain now. Agreed the `lock.go` one is the strongest candidate in the tree — it's immediately followed by a full `MigrateUp` on that same conn, so an undrained result set there poisons the entire retry pass.

**3. Rebase trap — caught, and it was real.** `TestRunMigrationsLargeRigNoticeOnlyOnMainSource` merged silently exactly as you predicted. Both bounds lowered to 39, matching the two tests already lowered.

## Part B

**Scoping fix applied.** The repairs now stage `dolt_nonlocal_tables` by name instead of `DOLT_COMMIT('-Am', ...)`, so the bounded-migrate path can't sweep migrations 1..upTo's uncommitted working set into a repair-labelled commit. I verified the behavior against a real dolt server rather than assuming it:

- `CALL DOLT_ADD('dolt_nonlocal_tables')` stages only that table — a separate dirty table stayed at `staged=0`.
- The scoped commit committed only the staged table; the other table remained uncommitted afterward.
- `--skip-empty` with nothing staged is a clean no-op, not an error.

I went with explicit scoping over gating on `commitEachStep` because it needs no new plumbing through `preMigrationRepair`'s signature and is correct on both paths rather than just avoiding the bad one.

Also dropped the stale justification in `anyNonlocalFrozenRowPresent`'s comment that leaned on the old `-A` behavior — the guard is still there, but its reason is now just "no-op on the common path," not commit-sweep avoidance.

## Conflict resolution

Main's repairs at versions 47 and 53 are preserved in the keyed registry, including #4690's `ensureDependenciesIDColumn` in the v53 chain (main had extended that `case 53` since my base). Also corrected a stale comment claiming `ignoredSource` has 11 migrations — it has 16.

## Verification

`go build`, `go vet`, `scripts/check-migration-hygiene.sh` (frozen-migration check C passes — no shipped bytes changed), and the full `internal/storage/schema` suite are green. Both `initschema_idempotent_test.go` integration tests now actually run and pass against a real dolt sql-server (`-tags integration`, `dolthub/dolt-sql-server:2.2.0`) — the partial-0040 dup-key state and the partial-0041 "nothing to commit" state both recover.

Two notes on the local test environment, neither caused by this diff — both reproduce identically on a clean `origin/main` worktree:
- Six integration tests fail locally (`TestMultiProcessSchemaInit*`, four `TestDoltNew_*RemoteMigrateGate*`), same set on main.
- Three `schema_skew` unit tests fail if `BD_IGNORE_SCHEMA_SKEW=1` is set in the ambient environment; they pass with it unset. Arguably those tests should clear the var themselves rather than inherit it, but that's unrelated to this PR and I left it alone.

**On sequencing (#4804):** happy to rebase over it whichever way it lands. Agreed there's no design interaction — the repair runs inside the locked pass, within #4804's sentinel-revoked interval.

Not merging this myself, per your merge-discipline note — it's yours to land, and it'll want fresh approval on `b2918fa` since this is a substantive rework rather than a pure rebase.

Rome-1:
Amended to `bb34153` — the only CI failure on `b2918fa` was a `misspell` hit on a comment I wrote in the scoping change (`labelled` → `labeled`); all four red checks were that same one line. Re-ran `golangci-lint v2.10.1` locally with CI's exact invocation (`--config=.golangci.yml --build-tags=gms_pure_go`): 0 issues. No functional change from `b2918fa`.

maphew (approved):
This round closes it out — approving. You addressed all four items from the last review in one clean rebase, corrected the mechanism argument instead of hand-waving it, and pulled the repro harness when it turned out to be a false green. That last one is the kind of intellectual honesty that makes reviews easy, and it is appreciated.

We re-verified independently before approving: the patch applies clean to current main (all 8 `CALL DOLT_*` sites in `internal/storage/schema/` unmoved, and the diff converts all 8 — `unstageIgnoredTables` delegates to a converted site, so there is no ninth); the driver claim holds in the pinned source (`mysqlConn.exec` has three early error returns before its terminal `discardResults`, while `mysqlRows.Close` drains unconditionally — Exec-on-error dirties the wire, Query+Close does not); the drain loop plus deferred Close covers multi-result-set procedures on both the clean and mid-stream-error paths; and `drainCall` with args stays on the server-side prepared path exactly as `ExecContext` did, so no protocol change. On part B, the `repairKey` registry is a behavior-preserving refactor of main's switch (no competing repair has landed since), `--skip-empty` and the `dolt_nonlocal_tables` column list check out against the pinned Dolt, and the by-name staging correctly keeps the bounded-migrate working set out of repair-labeled commits.

Three follow-ups, none blocking — say the word if you would rather we take any of them maintainer-side after the merge:

1. `migration_body_drain_test.go` — the guard test's ground truth is `strings.Contains(upper, "CALL DOLT_COMMIT")` while `procedureCallRe` matches any `CALL`. A future migration using only `CALL DOLT_ADD`/`DOLT_RESET` trips the `detected && !hasCall` branch and fails CI for doing the correct thing. One-line fix: ground truth `"CALL DOLT_"`.
2. `repairPartial0041NonlocalDelete` cannot distinguish "0041 died right after its leading DELETE+COMMIT" from "0041 died after the `depends_on_wisp_id` ALTER" — in the second case the replay's `@needs_migrate` guard then skips the remaining body and records v41 over a `dependencies` table with no rebuilt PK/FK/indexes, where pre-patch that state hard-failed visibly. The window is narrow (per-step commits leave those ALTERs as uncommitted changes the dirty guard catches first), so either a comment saying exactly that, or an additional gate on `depends_on_wisp_id` being absent, would do.
3. `commitNonlocalRepair`'s `CALL DOLT_ADD('dolt_nonlocal_tables')` is not tolerant of Dolt's "table not in working set" error, so on an engine where nonlocal mutations never enter the working root it converts a `nothing to commit` failure into a differently-worded staging failure. One line: tolerate the no-op the way `commitMigrationStep` tolerates `nothing to commit`.

Two nits at your discretion: `drainCall` drops the deferred `rows.Close()` error (a drain failure inside Close returns nil while the conn is still dirty — the exact class the function exists to prevent; `defer func() { err = errors.Join(err, rows.Close()) }()` closes it), and the comment-only `t.Cleanup` body in `initschema_idempotent_test.go` can go.

One small ask before merge: the PR description is now stale — it still describes editing 0040 to `INSERT IGNORE` and adding `--skip-empty` to the shipped migration bytes, which the final diff deliberately does not do (frozen bytes stay frozen; repairs live in the registry). Since the squash commit message inherits from the description, a quick refresh keeps the history honest.

For the record on scope: this does not close gastownhall/beads#4800 — on a pristine store both repairs correctly short-circuit before touching anything, and #4800's frozen-`DOLT_COMMIT`-finds-no-diff failure is a different mechanism. We will note that there so the correlation doesn't auto-retire it.

We also spotted that the same undrained-CALL shape exists at much higher call volume on transaction-pinned connections in `internal/storage/dolt/` (`issues.go`, `slots.go`, `store.go` — `_, _ = tx.ExecContext("CALL DOLT_ADD...")` followed by `CALL DOLT_COMMIT` on the same tx). That is out of scope here and tracked on our side; mentioning it in case you want first crack at it, since it is your pattern.

Merge mechanics: our patrol lands this once the base branch is green again (a flake-rerun is settling on main right now). Thanks for the persistence across three rounds — this is a genuinely valuable fix.

PR #4514, opened 2026-06-29
docs(upgrade): add validated recipe for remote-backed / multi-clone upgrades

maphew:
## Why

Upgrading 1.0.x → 1.1 across a remote-backed database is not "just replace the
binary." 1.1 adds the remote-migrate gate (#4259): `bd` refuses to silently
apply pending schema migrations to a database that has a Dolt remote, and the
`0050` dependencies primary-key reshape can leave independently-migrated clones
permanently un-mergeable. Today that procedure lives only as prose scattered
across the CHANGELOG upgrade note, the gate's own error text, and
`docs/RECOVERY.md` — there is no single ordered recipe in the user-facing
upgrade guide. A user who follows `docs/INSTALLING.md` literally hits the gate
with no forewarning.

This matters more at GA than for the RC: the RC is a GitHub prerelease few
people install, but 1.1.0 will land on stable Homebrew/npm/PyPI, so far more
users will swap the binary effortlessly and walk straight into the gate.

## What

- **`website/docs/getting-started/upgrading.md`** — new "Remote-backed databases
  and multiple clones" section with a single-clone recipe and a multi-clone
  recipe. Written keyed on the *behavior* (remote + pending migration), not on a
  specific version, so it stays correct when the RC is promoted to 1.1.0 and for
  every future migration.
- **`CHANGELOG.md`** — the 1.1.0-rc.1 multi-clone upgrade note rewritten as a
  numbered recipe that links to the guide.
- **`docs/INSTALLING.md`** — "After Upgrading" now points at the recipe and notes
  that prereleases ship only as GitHub prereleases (not on stable channels).

## Validated, not just written

I ran the recipe end-to-end on a live embedded database (v49 → v53): the gated
`BD_ALLOW_REMOTE_MIGRATE=1 bd migrate` + `bd dolt push` succeeded and all 464
issues survived. The dry run corrected two things the existing prose got wrong:

- The gate blocks `bd dolt push` / `bd dolt pull` too, not only `bd migrate`. So
  all pre-upgrade syncing must be done with the **old** binary, and other clones
  adopt the migrated database with **`bd bootstrap`**, not `bd dolt pull`.
- The migration-content-skew check ships inside `bd doctor`, which is **server
  mode only** ("not yet supported in embedded mode"). Embedded users verify with
  `bd version` / `bd ready` instead. (Filed separately as a product gap.)

## Notes

- Docs-only; no code or schema changes.
- Aware of #4286 (split `dependencies` into six target-typed tables). That is a
  separate, unmerged code direction; this PR documents the behavior as shipped in
  `v1.1.0-rc.1` (migration `0050`). If #4286 lands and changes the migration
  story, the recipe should be updated to follow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,17 +23,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   upgrade all clones of a shared remote together, letting one designated
   machine migrate and `bd dolt push` first.
 
-- **Upgrading from 1.0.4/1.0.5 with multiple clones: sync before AND
-  immediately after.** This release reshapes the `dependencies` primary key
-  (migration `0050`). If two clones cross that boundary with un-synced
-  dependency edits on both sides, their histories become permanently
-  un-mergeable — Dolt refuses the merge outright (`cannot merge because table
-  dependencies has different primary keys in its common ancestor`) before the
-  pull auto-resolver can run. To stay safe: `bd dolt push` + `bd dolt pull`
-  on every clone *before* upgrading, let one designated machine upgrade,
-  migrate, and `bd dolt push`, then on each remaining clone upgrade and
-  `bd dolt pull` *before* doing tracked work. If clones have already forked,
-  see the recovery playbook:
+- **Upgrading a remote-backed database: one machine migrates, the rest adopt.**
+  This release reshapes the `dependencies` primary key (migration `0050`), and
+  `bd` now refuses to silently migrate a database that has a Dolt remote
+  configured. If two clones cross the `0050` boundary with un-synced dependency
+  edits on both sides, their histories become permanently un-mergeable — Dolt
+  refuses the merge outright (`cannot merge because table dependencies has
+  different primary keys in its common ancestor`) before the pull auto-resolver
+  can run. Replacing the binary alone is not enough; follow the ordered recipe:
+
+  1. With your **current** binary, on **every** clone: `bd dolt push` +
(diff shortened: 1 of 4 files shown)


maphew (commented):
## Review finding

[P2] Update the recovery playbook prevention flow to match the new gate behavior.

`docs/RECOVERY.md` still tells non-migrator clones to upgrade the binary and then run `bd dolt pull` in the prevention section. That now contradicts this PR's validated recipe: once the new binary sees pending migrations on a remote-backed clone, `bd dolt pull` is refused too, and the clone must adopt the migrated database with `bd bootstrap` instead.

Because the updated changelog links readers into this recovery playbook, the stale prevention bullets can send users back into the unsafe path this PR is trying to prevent. Please update that section to match the "one machine migrates, every other clone bootstraps" flow.

No other issues found. I checked the PR diff, #4259 / #4286 drift, and `git diff --check`; CI is green.

maphew:
Fixed in 888e928cc. The `pk-fork-refused` Prevention section now matches the validated recipe:

- "Sync before upgrading" notes that once the new binary is installed, `bd dolt push`/`bd dolt pull` are gated too, so syncing must happen on the old binary first.
- The third bullet changed from "upgrade, then `bd dolt pull`" to "every other clone adopts with `bd bootstrap`" — `bd dolt pull` is refused while the clone still has pending migrations, and the sync-before step is what preserves work since bootstrap replaces the local DB.

So the changelog → playbook link no longer routes users back into the gated path.

PR #4515, opened 2026-06-29
fix(gate): make the remote-migrate gate agent-safe, not a footgun

maphew:
## Why

Agents are the highest-volume bd users, and the #4259 remote-migrate gate hands
them a loaded gun. When a write is refused on a remote-backed database, the JSON
error's top-level `hint` is the literal runnable command
`BD_ALLOW_REMOTE_MIGRATE=1 bd migrate`. An agent optimizing to unblock the
command will just run it:

- single clone → harmless (it *is* the designated migrator), but
- shared remote / multiple clones → a second clone migrating independently
  forks the schema unrecoverably — the exact hazard the gate exists to prevent.

The gate stops naive humans safely (the terminal message frames the choice), but
the machine-facing payload pre-decides the dangerous action for agents.

## What

De-fangs the agent-facing surfaces. **The gate's decision is unchanged** — this
is messaging only.

- **JSON error (`handleRemoteMigrateGateJSON`)** — `hint` is now a non-runnable
  directive (`AgentDirective()`): "Coordination decision required … do NOT
  auto-run a migration — surface remote_migrate_gate.options to the operator."
  The runnable migrate command moves into
  `remote_migrate_gate.options[migrate]`, gated on its *single designated
  migrator* precondition and annotated with its `risk`, beside the `adopt`
  option. Adds `severity` / `human_decision_required` / `observed` / `expected`
  / `docs`, matching the existing `AgentDiagnostic` contract ("Go observes and
  reports, the agent decides and acts"). Backward-compatible: `error` and
  `current_version`/`latest_version`/`pending` are preserved.
- **Embedded lenient read-only warning** — this fires during `bd prime` and
  every read, i.e. *before* an agent attempts a write. Reframed from
  "To resolve … runs: `<command>`" to a conditional coordination decision
  listing both options, so the agent is steered correctly at session start.

## Tests

- `cmd/bd`: `hint` must equal the directive and must **not** equal the escape
  command; `options` must carry both paths; the migrate command must appear
  **only** under the conditional `migrate` option.
- `internal/storage/schema`: a source-level test locks the same safety contract
  (`AgentDirective` is not runnable; escape command lives only under `migrate`).
- Note: four `dolt`-CLI migration tests in `internal/storage/schema` fail in my
  WSL environment; they fail **identically on pristine `main`** with zero
  changes, so they are pre-existing/environmental and unrelated to this change.

## Scope and follow-up

This is the safe, contained half of a two-part plan to make the agent path
**footgun-free now** and **smooth soon**:

- This PR: stop handing agents an auto-runnable migrate. (footgun-free)
- Follow-up (separate, reviewed): a **smart gate** that fetches the remote
  schema head and resolves the safe cases automatically — adopt when the remote
  is ahead and content-matches, auto-migrate when the remote equals local and
  all pending migrations are deterministic (the hygiene-guarded majority post
  #4259), and stop only on a genuine `content_hash` fork. The determinism work
  already shipped (`rowid`/`depid`, `nondeterminism-allowlist.txt`) makes this
  feasible. Happy to open a design issue if the direction is welcome.

Docs-side companion (recipe users follow once gated): gastownhall/beads#4514.

### DIFF
--- a/cmd/bd/remote_migrate_gate.go
+++ b/cmd/bd/remote_migrate_gate.go
@@ -2,20 +2,44 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 
 	"github.com/steveyegge/beads/internal/storage/schema"
 )
 
 // handleRemoteMigrateGateJSON renders the #4259 remote-migrate gate error as a
-// structured JSON error block, mirroring handleSchemaSkewJSON.
+// structured JSON error block for agent consumption.
+//
+// The top-level "hint" is deliberately a non-runnable directive, NOT the
+// `BD_ALLOW_REMOTE_MIGRATE=1 bd migrate` escape command: handing an agent a
+// ready-to-run migrate as "the fix" is the footgun that forks shared remotes on
+// multi-clone setups. The migrate command lives only inside
+// remote_migrate_gate.options[migrate], gated on its "single designated
+// migrator" precondition and annotated with its risk, so the agent surfaces a
+// human decision instead of auto-running it.
 func handleRemoteMigrateGateJSON(e *schema.RemoteMigrateGateError) {
-	outer := buildJSONError(e.Error(), e.EscapeHint())
+	outer := buildJSONError(e.Error(), e.AgentDirective())
 	if m, ok := outer.(map[string]interface{}); ok {
+		opts := make([]map[string]interface{}, 0, len(e.Options()))
+		for _, o := range e.Options() {
+			opts = append(opts, map[string]interface{}{
+				"id":       o.ID,
+				"when":     o.When,
+				"commands": o.Commands,
+				"risk":     o.Risk,
+			})
+		}
 		m["remote_migrate_gate"] = map[string]interface{}{
-			"current_version": e.CurrentVersion,
-			"latest_version":  e.LatestVersion,
-			"pending":         e.Pending,
+			"current_version":         e.CurrentVersion,
+			"latest_version":          e.LatestVersion,
+			"pending":                 e.Pending,
+			"severity":                "blocking",
+			"human_decision_required": true,
+			"observed":                fmt.Sprintf("%d pending schema migration(s) and a configured remote", e.Pending),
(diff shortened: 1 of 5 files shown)


maphew:
Opened the design issue for the "smooth" follow-up referenced above: #4516 — a state-aware gate that auto-adopts/auto-migrates the safe cases (using the already-shipped determinism + `schema.ContentHashSkew` machinery) and stops only on a genuine fork, reusing the non-runnable `AgentDiagnostic` options this PR introduces.

This PR remains the contained, messaging-only half (footgun-free now); #4516 is the just-works half (smooth, behind review).

maphew (commented):
Thanks, this looks directionally right to me. I found one small merge-order issue to fix before landing:

- `cmd/bd/remote_migrate_gate.go` emits a `docs` URL to `website/docs/getting-started/upgrading.md#remote-backed-databases-and-multiple-clones`, but that anchor is not present on `main` in this PR. It is added by companion PR #4514, which is still open. If this PR merges first, the agent-facing JSON points blocked users at a missing section. Please either merge #4514 first, include the doc change here, or point at an already-existing recovery doc.

Local verification:

- `git diff --check`
- `go test ./internal/storage/schema -run 'TestRemoteMigrateGate|TestCheckRemoteMigrateGate'`

I could not run the touched `cmd/bd` or embedded-Dolt package tests locally on this Windows machine because the local build environment is missing ICU headers (`unicode/uregex.h`). GitHub CI is still running several embedded-Dolt/regression jobs.

maphew:
Good catch — and it's the cross-PR coupling the split introduced. The `docs` anchor (`#remote-backed-databases-and-multiple-clones`) lands with companion PR #4514, whose CI is green and which is mergeable/clean.

Recommend simply **merging #4514 before #4515** (your option 1) — no code churn here, and the agent-facing JSON points at a live section the moment this lands. If you'd rather decouple so order can't matter, I can repoint the `docs` URL at `docs/RECOVERY.md#pk-fork-refused` (already on `main`), at the cost of a less-ideal target for a not-yet-forked agent. Your call.

PR #4518, opened 2026-06-30
fix(schema): seed canonical dolt_ignore patterns idempotently at MigrateUp (GH#4378)

### LINKED ISSUE #4378: dolt_ignore patterns are one-shot migration side effects: out-of-band-copied DBs stay under-seeded forever (wisp/local churn feeds the dirty-table migration gates)
## Summary

4 of the 5 canonical `dolt_ignore` patterns (`wisps`, `wisp_%`, `local_metadata`, `repo_mtimes`) are seeded **only** as one-shot side effects of migrations `0019_wisps_dolt_ignore.up.sql` and `0028_local_state_dolt_ignore.up.sql`. They execute exactly once, when those versions cross. The only pattern re-asserted idempotently is `ignored_schema_migrations` (`internal/storage/schema/schema.go`, the `REPLACE INTO dolt_ignore` inside the migration pass).

Any database whose `schema_migrations` cursor arrives at-latest **without actually executing** those migrations never gets the other 4 patterns — permanently.

## How a database gets into that state

Out-of-band materialization. We hit it with a table-by-table copy used to rename a database on a shared `dolt sql-server` (Dolt has no `RENAME DATABASE`): `CREATE DATABASE new` + copy all base tables from the old DB. The copy carries `schema_migrations`/`ignored_schema_migrations` (cursor at latest) but misses the `dolt_ignore` system table (not a base table) and the dolt commit history. Same end state for dump-restores or any external provisioning that stamps the cursor.

On the next `bd` open, `migrationWorkNeeded` correctly reports nothing to do, so 0019/0028 never run → 4 patterns missing forever.

## Consequences

- `wisp_%`, `local_metadata`, `repo_mtimes` churn shows up in `dolt_status` as permanent dirty/new tables.
- That feeds the dirty-table migration gates: `pendingMigrationDirtyTables` blocks future migration passes ("pending schema migrations alter pre-existing dirty tables"), and `stageSchemaTables` (which skips pre-dirty tables) can end up committing nothing — in our case a freshly-materialized DB stayed entirely working-set-only for a week and blocked a fleet-wide migration drain.
- If the churn ever gets committed, clone-local state lands in history.

Forensic tell that a DB was copied rather than created (useful for triage): `MIN(applied_at)` in `ignored_schema_migrations` **predates** the `Initialize data repository` root commit in `dolt_log`.

## Fix we ship in our fork (offer to PR)

[`marcodelpin/beads@082796b711`](https://github.com/marcodelpin/beads/commit/082796b7112b63a7efe3577efbfb40886adb21c9):

1. Package-level canonical list `doltIgnorePatterns` (all 5 patterns).
2. `seedDoltIgnorePatterns()` — `INSERT IGNORE` per pattern — called at the **top of `MigrateUp`, before the `migrationWorkNeeded` short-circuit**, so an under-seeded database heals at its first write-mode open. `INSERT IGNORE` (not `REPLACE`) keeps it a no-op on healthy DBs and respects an explicit operator override (`ignored=false` row).
3. `dolt_ignore` is exempted from the pre-existing-dirty guards (`delete(dirtyBefore, "dolt_ignore")`, same treatment as the aux-rekey tables) — required because `pendingMigrationDirtyTables` text-scans pending up.sql, and a pre-dirty `dolt_ignore` plus any future migration touching `dolt_ignore` would otherwise hard-block the gate. `stageSchemaTables` then commits the seeded rows with the pass.
4. Regression test (`TestMigrateUpSeedsIgnorePatternsWhenNoWorkNeeded`: the seed must fire before the no-work short-circuit) + the dirty-guard canary test moved from `dolt_ignore` to a real user table.

Running in production on our fleet (~140 DBs, single shared sql-server) since 2026-06-12; verified end-to-end (delete a pattern → one write-mode open restores it).

Happy to open a PR if you want this upstream — or adapt the approach if you'd prefer the seed somewhere else (e.g. store init).

marcodelpin:
Fixes #4378.

## Problem

`dolt_ignore` patterns are seeded as one-shot side effects of migrations `0019_wisps_dolt_ignore` / `0028_local_state_dolt_ignore`. A database materialized **out-of-band** (table-by-table copy, dump restore, rename -- Dolt has no `RENAME DATABASE`) arrives with its `schema_migrations` cursor already at-latest, so `migrationWorkNeeded` short-circuits and those migrations never run.

The DB then stays under-seeded **forever**: `wisp_%` / `local_metadata` / `repo_mtimes` churn as permanent dirty/new tables in `dolt_status`, which feeds the dirty-table migration gates (`pendingMigrationDirtyTables`) and can block future migration passes.

Full reproduction recipe is in #4378 (out-of-band `CREATE TABLE ... LIKE` + `INSERT ... SELECT` per base table -> `ws_copy.dolt_ignore` ends up empty, 0 of 5 patterns, where a freshly-`bd init`'d DB has all 5).

## Fix

`MigrateUp` now re-asserts the full canonical pattern set via `INSERT IGNORE INTO dolt_ignore` **before** the `migrationWorkNeeded` short-circuit:

- `INSERT IGNORE` is a no-op on healthy DBs (existing rows untouched) and respects an explicit operator override (a pattern present with `ignored=false`).
- `dolt_ignore` is exempted from the pre-existing-dirty guards as pass-owned state (the same treatment the aux-rekey tables already get), so on an under-seeded DB the newly-inserted rows are committed by the pass.
- This replaces the narrower `REPLACE INTO dolt_ignore VALUES ('ignored_schema_migrations', true)` that previously ran only **inside** the (skipped) migration pass.

## Tests

- `TestMigrateUpSeedsIgnorePatternsWhenNoWorkNeeded` -- new regression guard: the no-work-needed path still seeds the full pattern set before short-circuiting.
- `TestSchemaMigrationRejectsChangedPreExistingDirtyTable` updated to use a real user table (`custom_statuses`) as its dirty canary; it previously dirtied `dolt_ignore`, which this change intentionally exempts from the guard.
- `go build`, `go vet`, and the schema-package tests pass locally (`-tags gms_pure_go`).

Branch built on top of current `gastownhall/main`.


maphew (commented):
Thanks for tracking this down and for the clean regression test - the core diagnosis in #4378 (out-of-band-materialized DBs never re-running the one-shot 0019/0028 seeding side effects) and the fix direction (idempotent INSERT IGNORE re-assert ahead of the migrationWorkNeeded short-circuit) are both right, and generalizing from the single ignored_schema_migrations pattern to the full canonical set is a good simplification.

One small observation, take it or leave it: on the migrationWorkNeeded == false short-circuit path (the exact scenario the fix targets - an out-of-band DB with its cursor already at latest), seedDoltIgnorePatterns runs and then MigrateUp returns with no DOLT_COMMIT, and no caller of schema.MigrateUp (embedded or sql-server via MigrateUpWithLock) commits that path either. So when the seed actually inserts rows, dolt_ignore sits as an uncommitted working-set diff until something else commits. In practice this is transient rather than harmful: the patterns take effect immediately from the working set, embedded mode's per-command auto-commit and the pre-pull CommitPending both use DOLT_COMMIT('-Am'), and server-mode commitWorkingSet stages every dirty table from dolt_status - so the next write or pull sweeps it up, and the mergesettle preMergeClean gate is unaffected because every pull path commits pending changes before DOLT_MERGE. The residual cost is tidiness: the seed rides along inside an unrelated "bd: commit pending changes" or issue-write commit instead of a labeled schema commit, and a repo running only read-only commands (or dolt.autocommit batch mode) carries the diff until then.

If you want the heal to converge in one pass, a scoped commit on the !needed short-circuit when seedDoltIgnorePatterns actually changed something (track RowsAffected) - DOLT_ADD('dolt_ignore') followed by DOLT_COMMIT('-m', 'schema: seed dolt_ignore patterns') - would do it, with a test asserting the commit happens only when rows changed. Entirely optional; fine as-is or as a follow-up.

Everything else here looks solid - the dolt_ignore exemption from the pre-existing-dirty guards is consistent with the existing aux-rekey-table treatment, the updated TestSchemaMigrationRejectsChangedPreExistingDirtyTable canary swap to custom_statuses is a reasonable adaptation given the exemption, and CI is fully green including the migration-hygiene and differential-regression gates.

Happy to help land this - since maintainer edits are enabled on the branch, we can push a small follow-up commit for the optional scoped-commit tidy-up if you'd like, or merge as-is.

marcodelpin:
Thanks for the thorough review — and for spelling out the working-set residue on the short-circuit path; agreed it's worth closing in one pass rather than leaving the seed to ride along in an unrelated commit.

Implemented the scoped-commit tidy-up as d384660e6:

- `seedDoltIgnorePatterns` now tracks `RowsAffected` and reports whether it changed anything;
- on the `migrationWorkNeeded == false` short-circuit, when the seed actually inserted rows, `MigrateUp` runs `CALL DOLT_ADD('dolt_ignore')` + `CALL DOLT_COMMIT('-m', 'schema: seed dolt_ignore patterns')` so the heal converges immediately with a labeled commit;
- the existing no-work regression test now asserts the scoped commit fires, and a new negative test (`TestMigrateUpSkipsSeedCommitWhenNothingChanged`) asserts a healthy database (all `INSERT IGNORE` no-ops) stages/commits nothing.

One small judgment call: a `RowsAffected` error degrades to "unchanged" (skip the commit), which is exactly the pre-tidy-up behavior — the seed stays a working-set diff swept up by the next commit — rather than failing the open or risking an empty `DOLT_COMMIT`.

Verified locally on go1.26.4: `gofmt`/`go vet` clean, `go build ./...`, and the `internal/storage/schema` + `internal/storage/embeddeddolt` suites green. Ready from my side once CI agrees.

marcodelpin:
CI note: the first run after the tidy-up went red in `TestMigrateUpReturnsDirtyTablesErrorForPreExistingDirtyTable` — that's the #4567 dirty-guard test (527ea0c23), which landed on `main` after this branch was cut and mocks `MigrateUp` without the GH#4378 seed, so it mismatched in the PR merge context (not a regression from the tidy-up itself).

Fixed by merging current `main` into the branch (f79c0c5ac, clean) and adding `expectIgnorePatternSeed` to that test (f2a6b2f70) — seed rows there are owned by the migration pass, so no scoped-commit expectations apply. `internal/storage/schema` + `internal/storage/embeddeddolt` suites verified green locally on the merged tree.

maphew (approved):

Verified the mechanism end to end: the canonical five dolt_ignore patterns are re-asserted idempotently at the top of MigrateUp on both commit paths, including the no-work short-circuit, with a scoped labeled DOLT_COMMIT only when rows actually changed. Repeated opens are safe, operator `ignored=false` overrides are respected, the ExecContext-on-CALL idiom matches the rest of the file, and read-only opens and non-Dolt backends are untouched. The earlier test collision with the schema-lock work is reconciled, and the positive+negative tests cover the heal path. CI is fully green at this head and base main is green today.

Non-blocking observations, fine as follow-ups or to leave as-is:
- The no-work heal path commits with a generic label; a more specific label would make dolt history easier to read.
- `gofmt` nits none; behavior notes recorded in the tracker.

Thanks marcodelpin — this closes a real self-heal gap for databases materialized out-of-band (fixes gastownhall/beads issue 4378).

maphew:
Post-merge note: this broke main at 615d68a50 — `TestEmbeddedMigrateConvergesUnderConcurrentInterruptedRetries_4566` fails deterministically because the seed's dirty `dolt_ignore` rows ride the migration pass, and a pass killed between per-step commits now leaves a dirty working set (the #4566 self-heal contract).

Not something your CI could have caught: the convergence test landed on main (#4611) after this PR's checks ran — a stale-green-checks gap on the maintainer side, and the local pre-merge validation ran the schema suite but not the embedded-dolt convergence suite. Fix-forward is gastownhall/beads#4744, which commits the seed in its own scoped, labeled commit before the pass on the needed path too (same helper as your no-work-path commit). The seed mechanism itself is unchanged and sound.

PR #4520, opened 2026-06-30
feat(spool): offline write-spool for transient server-unreachable writes (GH#4379)

### LINKED ISSUE #4379: Server mode: writes are lost when the Dolt server is unreachable — offering an offline write-spool (shipped on our fork)
## Problem

In server mode, every write command (`create`, `update`, `note`, `close`, …) fails hard if the Dolt sql-server is briefly unreachable — server restart, network blip, `i/o timeout`. For interactive use that's an error message and a retry; for **agent/automation use the write is simply lost** — the agent has already moved on and nobody replays it.

With the current investment in remote/proxied server topologies (the bd-6dnrw series), more deployments will have bd talking to a server that can be momentarily away. There's currently no offline-write fallback.

## What we ship on our fork (offer to PR)

We've been running an offline write-spool in production since 2026-05-15 on a single shared sql-server hosting ~140 databases:

- **Transient vs permanent classification** — network timeouts, 5xx, Dolt `i/o timeout` get spooled; SQL/constraint errors still surface immediately (no masking of real failures).
- **Durable queue** — BLAKE3-keyed JSONL entries in `.beads/spool/`, atomic temp-rename writes, 100MB cap.
- **Replay** — opportunistic non-blocking `MaybeDrain` in `rootCmd` `PersistentPreRun` (any later bd invocation drains the queue once the server is back) + manual `bd spool status|drain|clear --confirm`.
- **Safety** — cross-platform locking (flock / LockFileEx), crash-safe drain with seen-set dedup, cursor advances only past processed entries on permanent error.
- Integration tests + operator doc (`docs/spool.md`).

Key commits on the fork:
- [`25dfec080e`](https://github.com/marcodelpin/beads/commit/25dfec080e181b56aa610e3c47dfbd24b6c35bfa) — `internal/spool/` package (entry/append/canonical + tests)
- [`e14fe3043c`](https://github.com/marcodelpin/beads/commit/e14fe3043cda469d7d51f3462f80cfb04e9e17af) — wiring: 4 write callsites + admin subcommand + `MaybeDrain` hook

## Question

Would you take a PR for this? Happy to adapt naming/UX to your preferences (e.g. gate it behind a config key, default-off, different admin-command shape). If you'd rather solve the gap differently, treat this as the problem report.

marcodelpin:
Fixes #4379.

## Problem

In server mode, a write that hits a transient error (server restart, i/o timeout, connection reset mid-write) is lost. `bd create` / `update` / `note` / `close` surface the store error and exit non-zero -- there is no record of the attempted write and no replay once the server is reachable again. On a shared sql-server under load (a restart blip across many DBs) this silently drops user writes.

Two deterministic reproductions are in #4379:

- cold-start (server already down at invocation -> store-open fails), and
- mid-write (a tiny TCP fault-proxy RSTs the connection on the write `INSERT` after store-open succeeds) -- the variant this spool targets.

## Fix

An offline write-spool:

- `internal/spool`: a crash-safe, file-locked, disk-capped append/replay queue with canonical (sorted-key, whitespace-free) JSON + blake3 dedup. Transient failures are durably queued; permanent failures are dead-lettered.
- `cmd/bd/write_with_spool.go`: `writeWithSpool(ctx, op, payload, directWrite)` runs the write; on a transient error (`IsTransientErr`) it spools the op and returns success, on a permanent error it returns the error as-is.
- The four write commands (create / update / note / close) route their store write through `writeWithSpool`.
- `bd spool status | drain | clear` admin subcommand.
- A non-blocking opportunistic `MaybeDrain` in the root PreRun replays pending entries before each command runs (TryLock: skipped if another process is draining; errors logged, never fatal).

## Tests

- `internal/spool` ships unit + crash + disk-cap + parity + replay tests; all pass (`go test -tags gms_pure_go ./internal/spool/`).
- `go build` + `go vet` clean against current `gastownhall/main` (verified on a Linux clean-room build, go1.26.4).

## Dependency

Adds `github.com/zeebo/blake3 v0.2.3` (already present transitively in `go.sum`) as a direct dependency for the canonical-hash dedup.

Branch built on top of current `gastownhall/main`.

### DIFF
--- a/cmd/bd/close.go
+++ b/cmd/bd/close.go
@@ -149,7 +149,17 @@ the flags appear in the command line.`,
 				}
 			}
 
-			if err := activeStore.CloseIssue(ctx, id, reason, actor, session); err != nil {
+			if err := writeWithSpool(ctx, "close",
+				spoolPayload(map[string]interface{}{
+					"id":      id,
+					"reason":  reason,
+					"actor":   actor,
+					"session": session,
+				}),
+				func() error {
+					return activeStore.CloseIssue(ctx, id, reason, actor, session)
+				},
+			); err != nil {
 				fmt.Fprintf(os.Stderr, "Error closing %s: %v\n", id, err)
 				continue
 			}
--- a/cmd/bd/create.go
+++ b/cmd/bd/create.go
@@ -545,7 +545,15 @@ var createCmd = &cobra.Command{
 			// If error getting parent or parent has no source_repo, continue with default
 		}
 
-		if err := store.CreateIssue(ctx, issue, actor); err != nil {
+		if err := writeWithSpool(ctx, "create",
+			spoolPayload(map[string]interface{}{
+				"issue": issue,
+				"actor": actor,
+			}),
+			func() error {
+				return store.CreateIssue(ctx, issue, actor)
+			},
+		); err != nil {
 			return HandleError("%v", err)
 		}
 
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/steveyegge/beads/internal/hooks"
 	"github.com/steveyegge/beads/internal/metrics"
 	"github.com/steveyegge/beads/internal/molecules"
+	"github.com/steveyegge/beads/internal/spool"
 	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/storage/dolt"
 	"github.com/steveyegge/beads/internal/storage/schema"
@@ -1206,6 +1207,20 @@ var rootCmd = &cobra.Command{
 
 		// Sync all state to CommandContext for unified access.
 		syncCommandContext()
+
+		// Opportunistic spool replay: if there are pending writes from a
+		// previous offline session, drain them now before the command runs.
+		// Non-blocking (TryLock): if another process is draining, skip.
+		// Errors are logged but never fatal -- the command proceeds normally.
+		if store != nil && !useReadOnly && beadsDir != "" {
+			go func() {
+				spoolDir := filepath.Join(beadsDir, "spool")
+				sp := spoolSingleton(spoolDir)
+				if err := spool.MaybeDrain(rootCtx, sp, spoolDispatch(rootCtx)); err != nil {
+					debug.Logf("spool MaybeDrain: %v", err)
+				}
+			}()
+		}
 
 		// Tips (including sync conflict proactive checks) are shown via maybeShowTip()
 		// after successful command execution, not in PreRun
--- a/cmd/bd/note.go
+++ b/cmd/bd/note.go
@@ -100,7 +100,16 @@ Examples:
 		updates := map[string]interface{}{
 			"notes": combined,
 		}
-		if err := issueStore.UpdateIssue(ctx, result.ResolvedID, updates, actor); err != nil {
+		if err := writeWithSpool(ctx, "note",
+			spoolPayload(map[string]interface{}{
+				"id":      result.ResolvedID,
+				"updates": updates,
+				"actor":   actor,
+			}),
+			func() error {
+				return issueStore.UpdateIssue(ctx, result.ResolvedID, updates, actor)
+			},
+		); err != nil {
 			return HandleErrorRespectJSON("updating %s: %v", id, err)
 		}
 		if err := commitPendingIfEmbedded(ctx, issueStore, actor, doltAutoCommitParams{
--- a/cmd/bd/spool_cmd.go
+++ b/cmd/bd/spool_cmd.go
@@ -1,0 +1,180 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	"github.com/steveyegge/beads"
+	"github.com/steveyegge/beads/internal/spool"
+)
+
+// spoolCmd is the root of the "bd spool" subcommand tree.
+var spoolCmd = &cobra.Command{
+	Use:     "spool",
+	GroupID: "maint",
+	Short:   "Manage the offline write spool",
+	Long: `Manage the offline write-spool (.beads/spool/).
+
+The spool buffers bd write commands (create/update/note/close) when Dolt is
+temporarily unreachable. Entries are replayed automatically at the start of
+the next bd command (MaybeDrain), or you can trigger a drain manually.
+
+Subcommands:
+  status   Show queue depth, oldest entry, and disk usage
+  drain    Force-drain the spool now (replay all pending entries)
+  clear    Wipe the spool (requires --confirm)`,
+}
+
+// spoolStatusCmd prints spool diagnostics without modifying state.
+var spoolStatusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Show spool queue depth, oldest entry timestamp, and disk usage",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		sp, err := openSpool()
+		if err != nil {
+			return err
+		}
+
+		count, err := sp.QueueLineCount()
+		if err != nil {
+			return fmt.Errorf("count queue: %w", err)
+		}
+		bytes, err := sp.QueueDiskBytes()
+		if err != nil {
+			return fmt.Errorf("stat queue: %w", err)
+		}
+		deadCount, err := sp.DeadLetterCount()
+		if err != nil {
+			return fmt.Errorf("count dead-letter: %w", err)
+		}
+		inflightAge, err := sp.InflightOldestAge()
+		if err != nil {
+			return fmt.Errorf("inflight age: %w", err)
+		}
+
+		cursor, err := sp.LoadCursor()
+		if err != nil {
+			return fmt.Errorf("load cursor: %w", err)
+		}
+
+		if jsonOutput {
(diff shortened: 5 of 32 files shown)


marcodelpin:
Heads-up on the overlap with #4462 (phase-aware write-tx retries): this PR is built on top of it (fa4dce454 base), and `go build` + the `internal/spool` tests pass with both in place. I want to surface the relationship up front rather than have it come up in review.

**How I see the layering**

- #4462 handles transient errors *within* a write transaction, retried in-process pre-commit. That covers the common server blip on an already-open connection.
- This spool is the *cross-invocation* durability layer for the cases #4462 cannot reach: the server unreachable at store-open (cold-start, `connection refused` before any tx begins), or an outage longer than the in-process retry budget. It durably queues the op so a later `bd` invocation replays it. Different layer, not a replacement for the retry path.

**The design point I want your call on (since #4462 deliberately stopped replaying ambiguous commit-result losses)**

- The spool's `IsTransientErr` classifies `connection reset` / `EOF` / `i/o timeout` as transient and queues for replay. For the cold-start path (`connection refused` at open) the write provably never reached the server, so replay is safe.
- For a mid-write `connection reset`, the error string alone cannot distinguish "reset before the server applied the write" (safe to replay) from "reset after a successful commit" (ambiguous -- replay would duplicate, e.g. a second `bd create`).
- The canonical-hash SeenSet dedups the spool against replaying the *same queued entry* twice, but it does not by itself prevent a duplicate when the original ambiguous write actually committed server-side (the spool never saw the ack) -- which is exactly the case #4462 is careful about.

**Proposal**

Scope the spool to the unambiguous pre-send failures (cold-start / `connection refused` at open) and defer mid-write ambiguous resets to #4462's retry-or-surface semantics. That keeps the two mechanisms cleanly separated and avoids the duplicate-on-ambiguous-commit case you designed #4462 to prevent. I'm happy to narrow `IsTransientErr` accordingly, and/or reuse the reconciled uow/dolt transient classifier so there's a single source of truth for "transient" rather than a second one in the spool. Let me know which boundary you prefer and I'll adjust.

maphew (commented):
Thanks @marcodelpin - this is an unusually well-built PR: deterministic repros in #4379, a thorough test suite (crash, parity, disk-cap, cross-platform locks), and an honest self-review of the #4462 overlap. You asked for a design call on the layering, so here it is, along with what a review of the current branch found.

## The design call

We're going to pass on the always-on CLI write-spool. Two reasons:

1. **It changes bd's write contract.** Today exit 0 means the write is durably in the database. With the spool, `bd create`/`update`/`note`/`close` can return success for a write that hasn't happened yet and might dead-letter. That's a product-surface change we don't want as default behavior, and it bites hardest exactly in the agent-automation case #4379 is about (see the create-ID issue below).
2. **Charter storage boundary.** `docs/PROJECT_CHARTER.md` says to avoid beads-side flocks, storage-specific retry loops, and crash-recovery workarounds - reconnect/retry-until-reachable for transient connection failures is the Dolt driver's job. #4462 (which you built on, now merged) plus the follow-up read-retry work is the in-tree shape of that; the remaining cold-start gap should be routed toward the driver rather than a parallel retry/WAL layer in beads. The spool also introduces a second transient-error classifier alongside `internal/storage/uow/errors.go` - you flagged this yourself, and a single source of truth is the right instinct.

On your specific question: yes, scoping replay to unambiguous pre-send failures is the only safe boundary - a mid-write `connection reset` can follow a server-side commit, and the SeenSet can't dedup against a write the spool never saw acknowledged. But note that once narrowed that far, most store-open failures abort before the four `writeWithSpool` callsites even run, so there isn't much spool left - which is part of why we'd rather solve this in the driver/retry layer.

## Concrete issues found in the branch (for completeness)

- **Spooled `bd create` succeeds with an empty ID.** Issue IDs are generated inside the store's create transaction, so when the write is spooled, `issue.ID` is empty: the command prints an empty ID, runs dependency adds against an empty ID, and exits 0. Replay later assigns an ID nobody ever sees. (`cmd/bd/write_with_spool.go`, `cmd/bd/create.go`)
- **`Drain` can loop forever.** `replayEntries` returns a nil error on a transient dispatch failure, the cursor doesn't advance, and the outer loop only breaks when a batch comes back short - so with >= 10 queued entries and the server still down, `bd spool drain` spins indefinitely re-pulling the same batch. (`internal/spool/replay.go`)
- **The background `MaybeDrain` goroutine** in `PersistentPreRun` runs concurrently with the foreground command against the same global store (bd/dolt writes need to stay serial, and this fires in embedded mode too), and it's never waited on - a fast command exiting after dispatch but before `seen.set`/inflight persistence makes duplicate replay a normal-exit event, not just a crash case. (`cmd/bd/main.go`)
- **`context.Canceled` is classified transient**, so a Ctrl-C'd write gets queued and silently executed later. (`internal/spool/append.go`)
- **`queue.jsonl` is never compacted**, so the 100MB cap measures lifetime volume and eventually refuses all appends until `bd spool clear` - which itself doesn't take the drain lock and discards pending entries wholesale. (`internal/spool/spool.go`)
- Replay bypasses command-level post-write logic (embedded-mode commit, hooks, flush), and post-create dependency/label writes aren't spooled, so a spooled create still loses data.
- The branch is currently conflicting with main after the #4462 squash and needs a rebase regardless.

## What we'd like to keep

The problem report and repros in #4379 are genuinely valuable and already sharpened the retry work. The remaining unhandled case - server unreachable at store-open - is real; the direction we want is driver-level reconnect/retry (dolthub/driver) plus the existing in-process retry budget, not a durable queue with success-on-spool semantics. If you'd like to pursue that, a driver-side issue referencing your cold-start repro would be very welcome, and we're happy to link it from #4379.

Closing this one rather than asking for rework, since the changes needed amount to a different design - but with real appreciation for the engineering quality and the transparency about the #4462 overlap. Thank you.

marcodelpin:
Thanks for the design call — the write-contract reasoning is convincing (the empty create-ID alone is disqualifying for the always-on shape), and the branch findings are all fair. We'll keep the spool fork-side and triage those.

Following up on the driver-side routing before filing anywhere, because the pointer needs one correction:

**bd's server mode doesn't go through `dolthub/driver`** — it speaks MySQL wire via `go-sql-driver/mysql` (`sql.Open("mysql", …)` throughout `internal/storage/dolt/store.go`); `dolthub/driver` is the embedded engine, which has no server to be unreachable. Retry-at-open has historically been treated as application policy at `go-sql-driver/mysql`, so I'd expect a decline there as well.

Here's what current `main` (1914af585) actually does in the cold-start case, verified on a scratch repo:

- **localhost server config**: `doltserver.EnsureRunning` auto-starts the managed server (GH#2049/#2050) — cold start heals itself, no gap.
- **external server config** (the #4379 topology — shared/proxied server): the fail-fast TCP pre-dial probe (3c167b82e) returns immediately:

  ```
  $ bd create "cold start lost write"    # .beads/metadata.json -> external server, briefly down
  Error: failed to open database: Dolt server unreachable at dolt-server.internal:13309:
  dial tcp 192.0.2.5:13309: connect: connection refused
  # ~0.04s, no retry engages — the write is gone, the agent has moved on
  ```

- every **post-open** layer already retries `connection refused` with backoff: `initSchemaOnDBWithRetryAndGate` (30s), `withRetry` (30s + circuit breaker), `withRetryTx` (15s, #4462), the post-CREATE catalog ping (10s).

So the one surface where `connection refused` is *not* retried is the pre-dial probe at open — exactly where a brief server restart loses an agent's write. Since that fail-fast is deliberate, the smallest in-tree shape looks like a bounded, config-gated retry budget around the probe for external-server mode only (reuse `newServerRetryBackoff`/`isRetryableError`, e.g. `dolt.open-retry-budget: 15s`, default off so today's fail-fast UX is untouched). That stays inside the charter's "in-process retry budget" and touches no write contract.

Happy to PR that, or to file it as a fresh issue for triage first — whichever you prefer. (If you'd still rather see something filed against a driver repo, the realistic target would be `go-sql-driver/mysql`, but I'd expect them to call retry-at-open application policy.)

maphew:
Your driver-routing correction is accepted — you're right that server-mode bd speaks MySQL wire via `go-sql-driver/mysql` and `dolthub/driver` is only the embedded engine, so the earlier "file it against the driver" pointer was wrong, and filing retry-at-open against `go-sql-driver/mysql` would indeed be declined as application policy. Thanks for verifying the cold-start behavior on a scratch repo instead of arguing from memory — the ~0.04s fail-fast trace showing the pre-dial probe as the one unretried surface is exactly the evidence that makes this easy to decide.

**Yes to the narrow shape — please PR it directly** (no need to file an issue first): a bounded, config-gated retry budget around the pre-dial probe, external-server mode only, reusing `newServerRetryBackoff`/`isRetryableError`, default **off** so today's fail-fast UX is untouched. That stays inside the charter's in-process retry budget and touches no write contract, precisely as you framed it. Two asks for the PR: keep the config key self-describing (`dolt.open-retry-budget` works), and include a test that the budget does NOT engage in embedded/localhost-managed modes (where `EnsureRunning` already heals).

With that path agreed, I'm closing this PR now — the design call from July 5 stands (always-on spool changes the write contract), you've kept the spool fork-side as discussed, and the surviving in-tree value has a better home in the new small PR. The deterministic repro suite you built here (#4379) remains the reference for validating the probe-retry work. Thanks for taking a "no" this constructively and coming back with a sharper shape — that's the collaboration this project runs on.

PR #4523, opened 2026-06-30
fix(ado): preserve unmanaged links on push instead of clobbering them

### LINKED ISSUE #4522: bd ado push deletes Related/Predecessor-Successor links (full relations replace)
## Summary

`bd ado sync --push-only` (and therefore `bd ado push`) **deletes existing
Related and Predecessor/Successor links** on the work items it pushes. The push
appears to send a full work-item update that reconciles the `relations` array
down to only the link types bd models (the parent/child hierarchy), discarding
any link types bd does not manage. Because a scoped push also writes the parent
link, the **parent work item's** relations get clobbered too — not just the
pushed item's.

This is silent data loss in Azure DevOps: links a human created in ADO disappear
the first time bd pushes a status change.

## Environment

- `bd version 1.0.3 (7fb454cd7: main@7fb454cd7f75)`
- Tracker: Azure DevOps (`bd ado`)

## Repro

1. In ADO, give a work item (and/or its children) some **Related** and
   **Predecessor/Successor** links in addition to the normal parent/child
   hierarchy.
2. Pull it into beads (`bd ado pull <id>`), change a status locally.
3. Push the status back, scoped to the item:
   ```
   bd ado sync --push-only --parent <id> --prefer-newer
   ```
4. Re-open the work item (and its parent) in ADO.

**Expected:** the status change is pushed; all existing links are preserved.
**Actual:** the Related and Predecessor/Successor links are gone from the pushed
item *and* from its parent. Only the parent/child hierarchy links survive.

## Likely root cause

The push builds the outgoing `relations` array from the set of links bd tracks
and sends it as the work item's complete relations, so ADO removes every relation
not present in that array. bd doesn't model Related / Predecessor-Successor /
hyperlink / attachment links, so they are dropped. The `--parent` scoping also
re-asserts the parent link on the parent work item, which is why the parent's
unrelated links are wiped as well.

## Suggested fix

On push, **read-merge-write** the relations array rather than replacing it: fetch
the work item's current relations, modify only the link types bd owns
(parent/child), and leave all other relation types untouched. Equivalently, emit
JSON-Patch `add`/`remove` ops only for the links bd manages instead of a wholesale
`replace` of `relations`.

## Impact / workaround

Any team using `bd ado` on work items that carry Related or dependency links in
ADO loses those links on the first status push. Our downstream workaround was to
gate ADO status push-back off by default so bd's push is never invoked, but that
also disables status mirroring — the real fix needs to be here in bd's push.

harry-miller-trimble:
## Problem

Fixes #4522. `bd ado push` (and `bd ado sync --push-only`) silently **deletes Related and Predecessor/Successor links** on the work items it pushes — and, because a scoped push re-asserts the parent link, on the **parent** work item too. Links a human created in ADO disappear the first time bd pushes a status change.

## Root cause

`LinkResolver.PushLinks` built the *desired* relation set only from the forward-direction link types beads emits — `Dependency-Forward`, `Hierarchy-Forward`, `Related` — targeting beads-tracked items, then removed **any** `System.LinkTypes.*` relation not in that set. That swept away two classes of links beads does not manage:

- **Reverse-direction links**: `Dependency-Reverse` (predecessor/successor) and `Hierarchy-Reverse` (parent-on-child). beads represents these relationships via the *forward* link on the **other** work item, so they never appear in the desired set — and were unconditionally deleted on every push.
- **Forward links to untracked items**: e.g. a `Related` link a human added in ADO to a work item outside the beads-synced set. The push call site filters these out of the desired set, so they were deleted too.

## Fix

Make push **read-merge-write** (as suggested in the issue): only remove a current relation when beads actually owns it —

1. its relation type is one beads emits (`removableRelTypes`: `Dependency-Forward`, `Hierarchy-Forward`, `Related`), **and**
2. its target is a beads-tracked work item (`managedTargets`, threaded in from `pushADOLinks`).

Every other relation — reverse-direction links, links to untracked items, hyperlinks, attachments, custom link types — is left untouched. Genuine deletion-sync of beads-owned links between tracked items still works.

## Tests

- New `TestPushLinks_PreservesUnmanagedLinks` asserts a push removes **zero** links when the work item carries only reverse-direction links (to tracked items) and a `Related` link to an untracked item. Verified it fails before the fix (3 spurious `remove` ops — exactly the reported data loss) and passes after.
- Existing `PushLinks` tests updated for the new `managedTargets` parameter; the remove-stale tests now pass the tracked-target set so deletion-sync of owned links is still covered.

### DIFF
--- a/cmd/bd/ado.go
+++ b/cmd/bd/ado.go
@@ -767,6 +767,24 @@ func pushADOLinks(ctx context.Context, resolver *ado.LinkResolver, at *ado.Track
 	var warnings []string
 	linkCount := 0
 
+	// Build the set of ADO work item IDs beads tracks. PushLinks only removes a
+	// current relation when its target is in this set, so links to items beads
+	// does not track (e.g. human-created Related / Predecessor-Successor links)
+	// are preserved rather than clobbered. See GH#4522.
+	managedTargets := make(map[int]bool)
+	for _, issue := range allIssues {
+		if issue.ExternalRef == nil {
+			continue
+		}
+		ref := *issue.ExternalRef
+		if !at.IsExternalRef(ref) {
+			continue
+		}
+		if id, cerr := strconv.Atoi(at.ExtractIdentifier(ref)); cerr == nil {
+			managedTargets[id] = true
+		}
+	}
+
 	for _, issue := range allIssues {
 		if issue.ExternalRef == nil {
 			continue
@@ -822,7 +840,7 @@ func pushADOLinks(ctx context.Context, resolver *ado.LinkResolver, at *ado.Track
 			continue
 		}
 
-		errs := resolver.PushLinks(ctx, workItemID, items[0].Relations, desired)
+		errs := resolver.PushLinks(ctx, workItemID, items[0].Relations, desired, managedTargets)
 		for _, e := range errs {
 			msg := fmt.Sprintf("Link sync ADO #%d: %v", workItemID, e)
 			warnings = append(warnings, msg)
--- a/internal/ado/links.go
+++ b/internal/ado/links.go
@@ -184,12 +184,34 @@ type adoLinkKey struct {
 	Commented bool // true if the link has a discovered-from comment
 }
 
+// removableRelTypes are the forward-direction ADO relation types beads emits
+// and therefore owns. Only links of these types may be removed during push.
+//
+// Reverse-direction link types — Hierarchy-Reverse (parent) and
+// Dependency-Reverse (predecessor/successor) — are deliberately excluded:
+// beads represents those relationships via the forward link on the *other*
+// work item, so they never appear in the desired set and must not be deleted.
+// Likewise, any other relation type (hyperlinks, attachments, custom link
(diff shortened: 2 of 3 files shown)


maphew (commented):
Thanks for tracking down this data-loss bug and fixing it - I confirmed the root cause against `PushLinks`/`isLinkRelation` on `main`: `isLinkRelation` matches any `System.LinkTypes.*` relation, and the old removal loop deleted anything not in the forward-only `desired` set, which is exactly the reverse-direction and human-created-link clobbering described in #4522.

The fix (removableRelTypes + managedTargets, read-merge-write) is minimal and correctly scoped: `managedTargets` is built from the full local issue store in `pushADOLinks`, not just the current push batch, so links to beads-tracked items outside this call's scope are still correctly preserved as "managed". `TestPushLinks_PreservesUnmanagedLinks` is a solid regression test, and the existing remove-stale tests were updated (not weakened) to thread through the new parameter. All CI checks are green.

Recommendation: merge as-is. Nice catch and clean fix.

Issue #4516, opened 2026-06-29
Smart remote-migrate gate: auto-resolve safe cases so the agent upgrade path just works (#4259 follow-up)


maphew:
## Summary

Make the remote-migrate gate ([#4259](https://github.com/gastownhall/beads/issues/4259)) **state-aware** so the common agent upgrade path just works, instead of blocking every write on a remote-backed database with pending migrations. This is the "smooth" follow-up to the messaging-only fix in #4515 (which made the gate *footgun-free* but still blocking).

## Background

Agents are the highest-volume bd users. After a binary upgrade, a database that has a remote and pending migrations refuses **every** write until a human intervenes. The current gate decides on three local facts only:

```
current > 0  &&  pending > 0  &&  hasRemote  →  refuse
```

It never looks at the **remote's** schema state, so it cannot tell apart:

1. **Safe first-mover** — remote is at the same version as me; no one has migrated yet.
2. **Already migrated** — remote is ahead; I should *adopt*, not migrate.
3. **Genuine fork** — remote applied *different content* for the same version (the real [#4259](https://github.com/gastownhall/beads/issues/4259) hazard).

All three get the same blanket refusal, and (before #4515) the same auto-runnable escape command. #4515 fixed the messaging; this issue proposes fixing the **decision**.

## Why this is feasible now

The data-safety work that shipped for 1.1 already provides everything needed to distinguish the three cases:

- **Deterministic row IDs** (`internal/storage/rowid`, `internal/storage/depid`) + migration `0050` make independently-migrated clones converge to byte-identical synced tables. The catastrophic fork was specific to the *legacy* non-deterministic migrations; new migrations are kept deterministic by the CI hygiene guard (`internal/storage/schema/migrations/nondeterminism-allowlist.txt`).
- **Per-migration content hashes** + `schema.ContentHashSkew(local, remote)` already detect case 3 precisely.
- `cmd/bd/doctor/migration_skew.go` already reads remote migration hashes from the **cached remote-tracking ref with no network fetch** (`readMigrationContentHashes(ctx, db, ref)`), then compares via `ContentHashSkew`. The smart gate can reuse this directly.

## Proposal

When the gate would currently fire, consult remote state and route:

| Remote vs local | Pending migrations | Action |
|---|---|---|
| remote **ahead**, content matches | (n/a — adopt) | **Adopt automatically** (fast-forward / bootstrap-equivalent). Agent unblocked, no migration. |
[...]

PR #4524, opened 2026-06-30
feat(gate): state-aware smart remote-migrate gate, opt-in (#4516)

maphew:
## Why

Agents are the highest-volume bd users, and after a binary upgrade the #4259 remote-migrate gate refuses **every** write on a remote-backed database with pending migrations until a human intervenes. #4515 made that refusal agent-*safe* (no auto-runnable footgun), but it is still a hard block on the common upgrade path. This makes the gate **state-aware** so the cases that are provably safe resolve automatically, and only the genuinely ambiguous case stops — implementing the design in #4516.

This is opt-in (`BD_SMART_GATE`); default behavior is the unchanged #4515 block.

## The key realization

The original #4516 proposal routed on "are the pending migrations deterministic?", which has no clean runtime signal. Tracing what migration `0050` actually guarantees collapsed that: the migrator always runs the **entire** pending batch, and any binary new enough to carry this gate also carries `0050` — *"the convergence point every clone reaches when upgrading to the fixed release"*. So if a legacy non-deterministic migration (`0037`/`0043`, random `UUID()` PKs) is pending, its convergence step is pending right behind it; two clones migrating independently diverge then **re-converge to byte-identical tables**.

That makes the safety input a single **version floor**, not a determinism manifest: at or above the highest genuinely-non-deterministic migration (`LastNonDeterministicMigration = 43`, cross-checked against `nondeterminism-allowlist.txt` by a test), every pending migration is deterministic by CI-hygiene construction, so a first-mover migrate is convergent. The proposal's "non-deterministic pending → stop a human" case is therefore **dropped as redundant** — a human there can only "run the migration," which is exactly what is already safe. Full reasoning: #4516 (design comment).

## What it does

When the blunt gate would fire and the designated-migrator escape hatch is not set, the smart gate consults the remote's **cached** `schema_migrations` ref (no network fetch — the same source the doctor skew check uses) and routes:

| Remote vs local (cached ref) | Action |
|---|---|
| `ContentHashSkew` non-empty | **stop** — genuine fork; human data-loss decision (which clone is canonical) |
| remote **ahead**, no skew | **stop with an adopt directive** — adoption is a destructive re-clone, never performed silently |
| remote **== local**, `current >= floor` | **auto-migrate** — safe first-mover; the full batch converges |
| below floor / uncached ref | **fall back to the blunt #4515 block** |

The stop cases reuse #4515's `AgentDiagnostic` shape (non-runnable directive + conditional `options`), tailored per decision so an agent surfaces the right human choice and never auto-runs a destructive action.

## Notable changes

- `internal/storage/schema/smart_remote_migrate_gate.go` — the router, the convergence-floor constant, and the opt-in flag.
- `RemoteMigrateGateError` gains a `Decision` (+ `SkewVersions`); `Error`/`UserMessage`/`AgentDirective`/`Options` and the JSON handler branch on it. **The default (smart-off) output is byte-identical to #4515.**
- Lifts the cached-remote-ref hash reader into the `schema` package (`ReadMigrationContentHashes` + ref-error classifiers); the doctor skew check now reuses it instead of a private copy.

## Testing

- Unit tests (sqlmock) for every routing branch, escape-hatch precedence, and that smart-off issues no extra reads and preserves #4515 exactly.
- `TestConvergenceFloorMatchesAllowlist` cross-checks the floor against `nondeterminism-allowlist.txt` so it cannot silently drift.
- `TestDoltNew_SmartRemoteMigrateGate_RealDolt` — a real-Dolt cross-clone test walking auto-migrate / adopt / fork-skew against a genuine pushed `remotes/origin/main` ref (guarded by the Docker test server). Its SQL (`DOLT_PUSH`, the `AS OF 'remotes/origin/main'` read, working-set divergence) was validated against Dolt 2.1.10.

## Open questions for review

- **Auto-migrate staging.** Default is opt-in via `BD_SMART_GATE`. Is a per-decision opt-in (e.g. auto-adopt-only vs auto-migrate) wanted before any default-on?
- **Non-destructive adopt.** This stops-with-directive on remote-ahead. A future fast-forward adopt via `bd dolt pull` (valid only when local never migrated and has no unpushed work) is a reasonable follow-up but wants its own cross-clone test.
- **Server/embedded parity** for the cached remote-state read.

Builds on #4515; same root issue #4259.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -856,7 +856,7 @@ jobs:
         env:
           BEADS_TEST_EMBEDDED_DOLT: "1"
           BEADS_TEST_EMBEDDED_TEST_BINARY: /tmp/embeddeddolt-test
-        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=10m
+        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=20m
 
   test-embedded-cmd:
     name: Test (Embedded Dolt Cmd ${{ matrix.shard }}/${{ strategy.job-total }})
--- a/.github/workflows/pr-risk.yml
+++ b/.github/workflows/pr-risk.yml
@@ -151,7 +151,7 @@ jobs:
         env:
           BEADS_TEST_EMBEDDED_DOLT: "1"
           BEADS_TEST_EMBEDDED_TEST_BINARY: /tmp/embeddeddolt-test
-        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=10m
+        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=20m
 
   test-embedded-cmd:
     name: Test (Embedded Dolt Cmd ${{ matrix.shard }}/${{ strategy.job-total }})
--- a/cmd/bd/doctor/migration_skew.go
+++ b/cmd/bd/doctor/migration_skew.go
@@ -11,7 +11,6 @@ import (
 	"github.com/steveyegge/beads/internal/configfile"
 	"github.com/steveyegge/beads/internal/storage/dberrors"
 	"github.com/steveyegge/beads/internal/storage/embeddeddolt"
-	"github.com/steveyegge/beads/internal/storage/issueops"
 	"github.com/steveyegge/beads/internal/storage/schema"
 )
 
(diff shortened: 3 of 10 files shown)


maphew (commented):
Reviewed this as maintainer and fixed the two smart-gate routing issues in `941eef1b8` on this PR branch.

Findings addressed:

- The smart gate treated `remote < local` as the safe first-mover case because it only checked whether the remote was ahead. It now requires the cached remote's max schema version to exactly match the local current version before auto-migrating; remote-behind falls back to the human coordination block.
- Server mode always compared `remotes/origin/<branch>` even when the store had a different configured default remote. The server-mode gate now passes `s.remote` through to the smart router, while existing default wrappers still use `origin`.

Validation run locally:

- `go test -tags gms_pure_go ./internal/storage/schema`
- `go test -tags gms_pure_go ./cmd/bd/doctor -run MigrationContentSkew`
- `go test -tags gms_pure_go ./cmd/bd -run TestHandleRemoteMigrateGateJSON`
- `go test -tags gms_pure_go ./internal/storage/dolt -run TestDoltNew_SmartRemoteMigrateGate_RealDolt`

Full `go test -tags gms_pure_go ./cmd/bd/doctor` was not used as the gate here because this Windows checkout still has unrelated pre-existing failures in `TestResolveBeadsDirForRepo_CachesFallbackResult` and `TestCheckGitUpstream`.

PR #4528, opened 2026-06-30
feat(prime): remind when AGENTS.md and CLAUDE.md diverge

maphew:
## Why

When a workspace keeps both `AGENTS.md` and `CLAUDE.md` as separate real files (rather than one symlinked to the other), and both carry the bd integration block, it is easy to edit one and forget the other. The two agent-instruction surfaces then drift apart silently. `bd prime` is already the place agents read for workflow context, so it is the natural spot to surface this.

## What

`bd prime` now appends a single one-line reminder to its generated output, but only when ALL of these hold:

- both `AGENTS.md` and `CLAUDE.md` exist in the workspace,
- both are regular files (symlinks are rejected via `os.Lstat`),
- they are not hardlinked to the same inode (`os.SameFile`),
- each contains the `BEGIN BEADS INTEGRATION` marker.

In every other case (the common case) it emits nothing and performs only a couple of cheap `Lstat` calls, so workspaces that do not meet all conditions pay no added output and negligible cost. The override path (a custom `PRIME.md`) is untouched.

The check lives in a small, side-effect-free helper `primeDivergenceReminder(workspaceDir)` in `cmd/bd/prime_divergence.go`, wired into `primeCmd` after the context is generated.

## Testing

```
CGO_ENABLED=0 go build -tags gms_pure_go ./...
CGO_ENABLED=0 go test -tags gms_pure_go -run Prime ./cmd/bd/
gofmt -l (changed files)   # clean
go vet -tags gms_pure_go ./cmd/bd/   # clean
git diff --check   # clean
```

New tests in `cmd/bd/prime_divergence_test.go` cover: both present + independent + marker => reminder; one file missing => nothing; marker missing in one => nothing; symlink case => nothing; hardlink (shared inode) => nothing; neither present => nothing; empty-arg/cwd path. Symlink and hardlink tests skip gracefully where the OS cannot create them.

### DIFF
--- a/cmd/bd/prime.go
+++ b/cmd/bd/prime.go
@@ -168,6 +168,10 @@ Config options:
 			}
 			return nil
 		}
+		// Append the AGENTS.md/CLAUDE.md divergence reminder only when both
+		// files are independent regulars carrying the bd marker; otherwise this
+		// adds nothing (zero output, negligible cost).
+		buf.WriteString(primeDivergenceReminder(""))
 		emit(buf.String())
 		return nil
 	},
--- a/cmd/bd/prime_divergence.go
+++ b/cmd/bd/prime_divergence.go
@@ -12,7 +12,7 @@ const beadsIntegrationMarker = "BEGIN BEADS INTEGRATION"
 
 // primeDivergenceReminder returns a one-line reminder when both AGENTS.md and
 // CLAUDE.md in workspaceDir exist as independent regular files (not symlinks,
-// not hardlinked to each other) that each contain the bd integration marker.
+// not sharing an inode) that each contain the bd integration marker.
 //
 // In every other case it returns the empty string and performs no work beyond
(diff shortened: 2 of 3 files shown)


maphew (commented):
Two small blockers:

- `cmd/bd/prime_divergence_test.go:13`: `writeFile` collides with the CGO-only package helper in `explicit_db_nodb_test.go`, so the embedded build fails. Rename this helper or use the existing one with `[]byte(...)`.
- `cmd/bd/prime_divergence.go:15`: `misspell` rejects `hardlinked` at lines 15, 48, and 59, breaking lint/artifact jobs. Reword to avoid that token, for example "share an inode".

Verified: no-CGO Prime tests pass; `CGO_ENABLED=1 go test -tags gms_pure_go -race -c ./cmd/bd` reproduces the compile failure.

Issue #4137, opened 2026-05-23
`bd 1.0.4`: schema migration 28 (`0028_local_state_dolt_ignore`) fails fatally with "nothing to commit" on every fresh Dolt-server-mode database


blairsilverberg:
# Summary

`bd 1.0.4` cannot complete schema migration 28 against a Dolt server when autocommit is on. The migration's `INSERT INTO dolt_ignore` statements run successfully and are auto-committed by Dolt, leaving the working set clean. bd then calls `CALL DOLT_COMMIT(...)` to seal the migration, Dolt returns `Error 1105 (HY000): nothing to commit`, and bd treats this as a fatal migration failure — it never inserts the `schema_migrations.version = 28` row. Every subsequent `bd` open retries migration 28, hits the same error, and fails open.

The database is effectively bricked for writes from the moment the first migration 28 attempt happens. We hit this 100% reproducibly on a fresh database created via `gt dolt init-rig <name>` (which is the centralized-Dolt setup path).

# Affected versions

- `bd` version: **1.0.4** (commit `ce242a879`)
- `dolt` version: 2.0.5
- `gt` version (creating the database): 0.12.0

# Reproduction

1. Create a fresh database on a centralized Dolt server. The exact path we used was `gt dolt init-rig openclaw`, but anything that runs `bd`'s migrations against a Dolt SQL server with autocommit on should reproduce it.
2. The new database now has:
   - `schema_migrations` populated with versions 1-27, each with an `applied_at` timestamp.
   - `dolt_ignore` populated with the four entries migration 28 is supposed to add (`local_metadata`, `repo_mtimes`, `wisp_%`, `wisps`).
   - **No** `schema_migrations.version = 28` row.
3. Run any bd write — e.g. `bd config set issue_prefix op`, `bd create ...`, `bd config set status.custom "..."` — from a `cwd` whose `.beads/metadata.json` points at the server database.

# Observed behavior

```
Error: failed to open database: failed to initialize schema:
schema migration: migration 0028_local_state_dolt_ignore.up.sql:
statement failed: Error 1105 (HY000): nothing to commit
```

The same error on every subsequent invocation, indefinitely. Read-only commands like `bd config show` and `bd list` succeed.

# Expected behavior

bd should complete migration 28 successfully on a fresh database and record the `schema_migrations.version = 28` row. Subsequent opens should observe migration 28 as applied and skip it.

# Root cause hypothesis

[...]

PR #4531, opened 2026-06-30
fix(storage): fail fast on writable opens of a schema-newer DB

maphew:
## Why

A stale `bd` that opens a database whose schema is **newer** than the binary
should fail fast with a clear "upgrade bd" message, not a cryptic SQL error.
This is the stale-binary incident behind #4135/#4137: an old `bd` opened a
forward-drifted DB and died with `column "depends_on_id" could not be found`
(a column dropped by a later migration) instead of telling the user the real
problem.

#4152 added the `SchemaSkewError` forward-drift guard and the
`BD_IGNORE_SCHEMA_SKEW` / `--ignore-schema-skew` escape hatch, but it wired
`CheckForwardDrift` only into the **read-only** store paths. The **writable**
open paths were left uncovered:

- `MigrateUp` no-ops when the DB is already past the binary's latest migration
  (`atLatest` uses `>=`), so a writable open of a forward-drifted DB succeeds
  and the failure only surfaces later, as a cryptic unknown-column error.
- The reported incident was in embedded mode, where writes (`bd create`,
  `bd update`, `bd close`) are the common case, so the uncovered writable path
  is the one users actually hit when their binary is behind.

## What

Extends the existing #4152 guard to both writable open paths; no new error
type, message, or flag (the `SchemaSkewError` UX and escape hatch are reused):

- **Server mode** (`internal/storage/dolt`): run `schema.CheckForwardDrift` on
  read-only **and** writable opens, before `initSchema`. Read-only behavior is
  unchanged; the writable branch now fails fast on forward drift.
- **Embedded mode** (`internal/storage/embeddeddolt`): run
  `schema.CheckForwardDrift` in `initSchema`, after the `USE <db>` switch and
  before the remote-migrate gate / `MigrateUp`. The read-only embedded open
  (`OpenReadOnly`) already guarded this; the writable open did not.
- `checkSchemaSkew` now reads the version via `CurrentVersion`, which treats a
  missing `schema_migrations` table as version 0. This makes the guard safe to
  run before `initSchema` creates the table on a fresh database (a fresh DB is
  version 0, i.e. no forward drift).
- `CheckForwardDrift` now accepts a `DBConn` (a pooled `*sql.DB` or a pinned
  `*sql.Conn`) so the embedded path can pass its pinned connection. Existing
  callers pass `*sql.DB`, which satisfies `DBConn`, so they are unaffected.

On the version-hint question (mapping a schema version to a required `bd`
release): this PR keeps #4152's deliberate choice to **not** hard-code a
version table. The message points users to rebuild from `main` or install the
latest release. A hand-maintained schema-version-to-release table would be
brittle and is better introduced separately if maintainers want it.

## Testing

```
CGO_ENABLED=0 go build -tags gms_pure_go ./...
CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/schema/...   # ok
CGO_ENABLED=0 go vet  -tags gms_pure_go ./internal/storage/schema/... \
  ./internal/storage/dolt/... ./internal/storage/embeddeddolt/...        # clean
gofmt -l <changed files>   # clean
git diff --check           # clean
```

Added two focused unit tests in
`internal/storage/schema/schema_skew_test.go`:

- `TestCheckSchemaSkew_MissingTable_NoError`: a table-not-exist error is
  treated as version 0 (no drift), covering the new safe-before-`initSchema`
  behavior on the writable path.
- `TestCheckForwardDrift_Conn_Ahead`: the exported guard accepts a `DBConn`
  and still reports forward drift.

The heavier dolt/embeddeddolt integration tests are server-gated; the touched
test binaries compile under `-tags gms_pure_go`.

Related: #4152 (forward-drift guard this builds on), #4135, #4137, #4259.

### DIFF
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -1154,12 +1154,17 @@ func newServerMode(ctx context.Context, cfg *Config) (*DoltStore, error) {
 		autoStartedServerDir: autoStartedDir,
 	}
 
-	if cfg.ReadOnly {
-		if err := schema.CheckForwardDrift(ctx, db); err != nil {
-			_ = db.Close()
-			return nil, err
-		}
-	} else {
+	// Forward-drift guard runs on read-only AND writable opens. A binary older
+	// than the database's schema cannot migrate it forward: MigrateUp no-ops
+	// when the DB is already past the binary's latest migration (atLatest uses
+	// >=), so without this guard a writable open would proceed and later queries
+	// would fail with cryptic unknown-column errors instead of a clear
+	// "upgrade bd" message (the stale-binary incident behind #4135/#4137).
+	if err := schema.CheckForwardDrift(ctx, db); err != nil {
+		_ = db.Close()
+		return nil, err
+	}
+	if !cfg.ReadOnly {
 		if err := store.initSchema(ctx); err != nil {
 			return nil, fmt.Errorf("failed to initialize schema: %w", err)
 		}
--- a/internal/storage/embeddeddolt/store.go
+++ b/internal/storage/embeddeddolt/store.go
@@ -273,6 +273,17 @@ func (s *EmbeddedDoltStore) initSchema(ctx context.Context) error {
 		}
 	}
 
+	// Forward-drift guard: if this database's schema is AHEAD of the binary,
+	// fail fast with a clear "upgrade bd" message before MigrateUp no-ops and a
+	// later query dies on a dropped/renamed column. Embedded mode is the mode
+	// the stale-binary incident (#4135/#4137) was observed in. The read-only
+	// embedded open (OpenReadOnly) already guards this; the writable open did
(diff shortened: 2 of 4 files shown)


maphew:
Reviewed the storage/schema changes and pushed one maintainer nit commit (58d76114b) so the DBConn test actually exercises a pinned *sql.Conn.

Local validation:
- go test ./internal/storage/schema
- go test ./internal/storage/dolt
- CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/embeddeddolt

I also ran CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/...; unrelated Windows/environment failures occurred in internal/storage/dbproxy/server and internal/storage/uow (POSIX absolute-path test assumptions, a locked temp server.log, and a proxy readiness timeout). I did not find blockers in the PR diff. Remaining GitHub build/regression/smoke jobs are still pending.

PR #4532, opened 2026-06-30
fix(scripts): document CGO override + stderr-warn in CLI docgen guard

maphew:
## Why

Regenerating the CLI doc artifacts (`scripts/generate-cli-docs.sh`, which also
drives `scripts/generate-llms-full.sh`) on a developer machine that has a C
compiler links CGO by default. A CGO-enabled `bd` exposes the full
`bd federation` subcommand tree, whereas CI builds the docs binary with
`CGO_ENABLED=0` (see `scripts/ci/pr-policy.sh`), which stubs federation
("built without CGO support"). The committed docs therefore come from the
stubbed build. A naive local regen produces ~hundreds of lines of spurious
federation churn (`federation.md`, `CLI_REFERENCE.md`, `llms-full.txt`) that
has nothing to do with the change being made — an easy trap to commit by
accident.

`generate-cli-docs.sh` already guards this by force-rebuilding a pure-go binary
when the resolved `bd` lacks the federation stub marker (added in #4474). This
PR hardens that guard's ergonomics: it offered no escape hatch for an
intentional full-federation regen, and announced the binary swap on stdout.

## What

- Route the CGO-mismatch notice to **stderr** so `--check`/CI stdout stays clean
  and the signal reads as a proper warning.
- Add a documented **`BD_DOCS_ALLOW_CGO=1`** override hook: set it to trust the
  supplied binary as-is (e.g. to deliberately regenerate the full federation
  tree). Default behavior is unchanged and safe — rebuild a pinned
  `CGO_ENABLED=0 -tags gms_pure_go` binary that matches CI.
- Make the warning name the pure-go recipe and reference the stub marker
  `Federation commands require CGO` (`cmd/bd/federation_nocgo.go`).
- Document the override in the script's `--help` text.

`generate-llms-full.sh` builds no `bd` binary (it only concatenates markdown),
so it needs no change here.

## Testing

- `bash -n` on both scripts: clean.
- `shellcheck scripts/generate-cli-docs.sh`: no findings. (Pre-existing
  informational/style findings in `generate-llms-full.sh` are untouched.)
- Built a pure-go binary (`CGO_ENABLED=0 go build -tags gms_pure_go`) and
  confirmed its `federation --help` contains the stub marker
  `Federation commands require CGO`; confirmed the CGO `federation.go` Long text
  does not, so the marker reliably distinguishes the two builds.
- Ran `./scripts/generate-cli-docs.sh --check <pure-go-bd>`: guard accepts the
  binary without rebuilding and reports docs fresh (`PASS`).
- `git diff --check`: clean.

(A CGO build could not be produced in this environment — missing ICU headers,
the same reason CI uses pure-go — so the negative-path rebuild was validated by
source inspection of the marker rather than execution.)

### DIFF
--- a/scripts/generate-cli-docs.sh
+++ b/scripts/generate-cli-docs.sh
@@ -31,6 +31,11 @@ Usage: scripts/generate-cli-docs.sh [--check] [--versioned VERSION] [path-to-bd]
 Generate live CLI docs from one bd process. Historical Docusaurus snapshots are
 left untouched unless --versioned VERSION is supplied by the release snapshot
 workflow.
+
+If the resolved bd binary is CGO-enabled it emits the full `bd federation` help
+tree that CI (CGO_ENABLED=0) stubs out; the script rebuilds a pinned pure-go
+binary to keep committed docs in sync. Set BD_DOCS_ALLOW_CGO=1 to bypass that
+rebuild and trust the supplied binary as-is.
 EOF
             exit 0
             ;;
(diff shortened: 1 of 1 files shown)


maphew (commented):
Review complete. I reviewed the one-file `scripts/generate-cli-docs.sh` change and did not find any blocking issues.

The `BD_DOCS_ALLOW_CGO=1` escape hatch is scoped to the intentional full-federation-docs case, and the default path still rebuilds the same `CGO_ENABLED=0 -tags gms_pure_go` binary used for CI-consistent docs. The stderr warning placement also behaves correctly for `--check`.

Local verification:

- `bash -n scripts/generate-cli-docs.sh`
- `./scripts/generate-cli-docs.sh --check`
- simulated a CGO-looking `bd` without the pure-Go federation stub marker; the script rebuilt pure-Go, exited 0, kept stdout to the PASS lines, and sent the warning to stderr
- confirmed the marker text exists in `cmd/bd/federation_nocgo.go`

I could not run local `shellcheck` because it is not installed in this environment, but the GitHub lint/check suite is green.

I cannot submit this as a formal approval from this account because GitHub rejects self-approval on the PR author account, so posting the review result as a comment.

PR #4536, opened 2026-07-01
feat(create): add status flag for initial issue status

thewoolleyman:
## What

Adds `bd create --status/-s` so a new issue can be created directly in a built-in or configured custom status.

The default remains unchanged: without `--status`, new issues start as `open`, except future `--defer` still starts them as `deferred`. When both `--status` and `--defer` are present, the explicit status wins and `defer_until` is still preserved.

## Why

This completes the existing custom-status workflow rather than adding a new status concept: `bd update --status` already validates built-in and configured custom statuses, and `bd create` should be able to start an issue in the same lifecycle state in one step. This avoids create-then-update transitions for agents or workflows that model their lifecycle in status.

## Notes

- Reuses `types.Status(...).IsValidWithCustom(...)` with configured custom statuses for CLI validation.
- Threads the initial status through the shared create builder and proxied-server input path.
- Normalizes typed `status.custom` entries during create insertion, so config like `review:wip` validates new issues with status `review`.
- PR preflight found #4023, but it is about prune/export/migrate-personal and does not overlap this create/status change.

## Validation

- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestBuildCreateIssueFromInput'`
- `CGO_ENABLED=1 go test -tags gms_pure_go ./internal/storage/issueops -run 'TestGetCustomStatusesTxParsesTypedConfigNames'`
- `CGO_ENABLED=1 go build -tags gms_pure_go ./...`
- Temp-repo smoke: `bd create --status blocked`, configured `status.custom review:wip` then `bd create --status review`, and invalid `--status nope`
- `./scripts/check-doc-flags.sh ./bd`
- `./scripts/check-doc-freshness.sh`

Known local validation issues unrelated to this diff:

- `make test` timed out in existing doctor tests: `TestDoctorWithBeadsDir` and `TestCheckDocumentationBdPrimeReference`.
- `golangci-lint run ./...` could not start because installed `golangci-lint` was built with Go 1.24 while this module targets Go 1.26.2.

### DIFF
--- a/cmd/bd/create.go
+++ b/cmd/bd/create.go
@@ -162,6 +162,16 @@ var createCmd = &cobra.Command{
 
 		issueType, _ := cmd.Flags().GetString("type")
 		assignee, _ := cmd.Flags().GetString("assignee")
+		statusFlag, _ := cmd.Flags().GetString("status")
+		if statusFlag != "" {
+			var customStatuses []string
+			if cs, err := store.GetCustomStatuses(rootCtx); err == nil {
+				customStatuses = cs
+			}
+			if !types.Status(statusFlag).IsValidWithCustom(customStatuses) {
+				return HandleErrorRespectJSON("invalid status %q (built-in: open, in_progress, blocked, deferred, closed, pinned, hooked; or configure custom statuses via 'bd config set status.custom')", statusFlag)
+			}
+		}
 
 		labels, _ := cmd.Flags().GetStringSlice("labels")
 		labelAlias, _ := cmd.Flags().GetStringSlice("label")
@@ -344,6 +354,7 @@ var createCmd = &cobra.Command{
 				Labels:             labels,
 				MolType:            molType,
 				WispType:           wispType,
+				InitialStatus:      statusFlag,
 				DueAt:              dueAt,
 				DeferUntil:         deferUntil,
 				Metadata:           metadata,
@@ -503,6 +514,7 @@ var createCmd = &cobra.Command{
 			Actor:              eventActor,
 			Target:             eventTarget,
 			Payload:            eventPayload,
+			InitialStatus:      statusFlag,
 			DueAt:              dueAt,
 			DeferUntil:         deferUntil,
 			Metadata:           metadata,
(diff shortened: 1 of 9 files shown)


maphew (commented):
Thanks Chad, this is a nice, tightly-scoped completion of the custom-status workflow.

What I checked:
- `--status/-s` on `bd create` reuses the exact same `types.Status(...).IsValidWithCustom(...)` validation shape already used for `--type`, in both the direct (`cmd/bd/create.go`) and proxied-server (`cmd/bd/create_proxied_server.go`) paths - consistent with existing conventions, not a new pattern.
- Explicit `--status` correctly wins over `--defer`-derived status while `defer_until` is still preserved (`buildCreateIssue`/`buildCreateIssueFromInput`), and there's a dedicated test for that interaction (`initial_status_wins_over_defer`, `TestBuildCreateIssueFromInput_ExplicitStatusWinsOverDefer`).
- The `GetCustomStatusesTx` refactor is a real bug fix, not just plumbing: it now normalizes typed `status.custom` entries (e.g. `review:wip`) down to the bare name (`review`) via `ResolveCustomStatusesDetailedInTx` + `CustomStatusNames`, which is what makes `bd create --status review` validate correctly against that config. Covered by `TestGetCustomStatusesTxParsesTypedConfigNames`. Confirmed this function has only one other caller (`issueops/create.go`), so the change is well-contained.
- Docs (`docs/CLI_REFERENCE.md`, `website/docs/cli-reference/create.md`) were regenerated for the new flag.
- Product-surface fit: issue lifecycle/status is explicitly listed as core surface area in `PROJECT_CHARTER.md`, and nothing here touches storage-engine internals (no flocks/retry/crash-recovery), so this stays on the right side of the driver/core boundary.

Two very minor notes, neither blocking:
- In `create_proxied_server.go`, `configUW.Close(ctx)` moved later in the function so the new status-validation block can call `configUW.ConfigUseCase().GetCustomStatuses(ctx)` before closing. That's fine, just flagging that a future cleanup could add `CustomStatuses` to `domain.CreateContext` (mirroring the existing `CustomTypes` field) so this path doesn't need to hold the UOW open for an extra query.
- Worth a follow-up doc/help-text mention of which statuses are "built-in" for discoverability, though the error message already lists them.

Recommendation: merge as-is. Nice work.

PR #4543, opened 2026-07-02
fix(notion): drop in_trash from data-source query body (Notion 400)

### LINKED ISSUE #4542: bd notion sync fails with Notion 400: body.in_trash should be not present
### Summary

Every `bd notion sync` invocation (`--pull`, `--push`, `--prefer-local`, …) fails with:

```
Error: fetching issues: Notion API error validation_error (400):
body failed validation: body.in_trash should be not present, instead was `false`.
```

### Cause

`Client.QueryDataSource` (`internal/notion/client.go`) includes `"in_trash": false` in the
`POST /v1/data_sources/{id}/query` request body:

```go
request := map[string]interface{}{
    "page_size":   maxPageSize,
    "result_type": "page",
    "in_trash":    false,
}
```

Under the API version the client pins (`DefaultNotionVersion = "2026-03-11"`), the query
endpoint rejects `in_trash` as a body property — it is valid on page **PATCH** bodies
(as used correctly in `ArchivePage`), but not on data-source **query** bodies. Notion
appears to have tightened server-side validation recently (observed failing from
2026-06-30; the same binary synced fine before that), so this now breaks every
Notion-sync user regardless of bd version — the field is still present at HEAD.

Because `QueryDataSource` is the enumeration step for both directions, **pull and push
both fail**; no Notion sync path works.

### Fix

Drop the key. Notion's query endpoint already excludes trashed pages by default, so
behavior is unchanged:

```diff
 		request := map[string]interface{}{
 			"page_size":   maxPageSize,
 			"result_type": "page",
-			"in_trash":    false,
 		}
```

### Repro

1. Configure Notion sync against any workspace.
2. `bd notion sync --pull --dry-run` → 400 as above (exit 1).

### Environment

- Observed on v1.0.3 and v1.0.4 (Homebrew); field present through current `main`
- macOS arm64; Notion API `2026-03-11`

### Verification of the fix

Rebuilt v1.0.3/v1.0.4 with only this line removed → `bd notion sync --pull --dry-run`
succeeds against a live workspace (3,397 pages queried); trashed pages remain excluded
from results.

uschtwill:
## What

Removes the `"in_trash": false` key from the `QueryDataSource` request body in
`internal/notion/client.go`. One-line deletion, no behavior change: Notion's
data-source query endpoint excludes trashed pages by default.

## Why

Fixes #4542.

Notion's query endpoint (under the pinned `Notion-Version: 2026-03-11`) rejects
`in_trash` as a body property with:

```
validation_error (400): body failed validation: body.in_trash should be not present
```

`in_trash` is valid on page PATCH bodies (kept as-is in `ArchivePage`) but not on
query bodies. Since `QueryDataSource` is the enumeration step for both sync
directions, every `bd notion sync` invocation currently fails — pull and push.

## Verification

- Rebuilt v1.0.3 and v1.0.4 with only this change; `bd notion sync --pull --dry-run`
  against a live Notion workspace succeeds (previously: 400, exit 1). 3,397 pages
  queried; a subsequent real pull + push round-trip completed cleanly.
- Trashed pages remain excluded from query results (endpoint default).
- `go build -tags gms_pure_go ./cmd/bd` clean.

### DIFF
--- a/internal/notion/client.go
+++ b/internal/notion/client.go
@@ -124,7 +124,6 @@ func (c *Client) QueryDataSource(ctx context.Context, dataSourceID string) ([]Pa
 		request := map[string]interface{}{
 			"page_size":   maxPageSize,
 			"result_type": "page",
-			"in_trash":    false,
 		}
 		if cursor != "" {
 			request["start_cursor"] = cursor


maphew (commented):
Thanks for the fast turnaround on this, @uschtwill - and for the thorough verification writeup (live workspace pull/push round-trip against 3,397 pages, plus confirming trashed-page exclusion is unaffected).

Reviewed against `internal/notion/client.go`: the removed `"in_trash": false` key is indeed query-body-only cruft - `ArchivePage`'s PATCH body correctly keeps `in_trash` since that's the field Notion expects there. The fix is behavior-preserving and matches the linked issue's repro and root-cause analysis exactly.

CI is green across the board (build, lint, full test matrix, upgrade smoke tests). Existing tests in `client_test.go` don't assert on the removed field, so nothing needed updating there - though a small follow-up asserting the query body does *not* contain `in_trash` would guard against this regressing again; not a blocker.

Recommendation: merge as-is.

PR #4551, opened 2026-07-02
feat(gate): turn the smart remote-migrate gate on by default (#4516)

### LINKED ISSUE #4516: Smart remote-migrate gate: auto-resolve safe cases so the agent upgrade path just works (#4259 follow-up)

maphew:
## Why

#4524 shipped the state-aware remote-migrate gate opt-in behind `BD_SMART_GATE`, deliberately leaving default-on as an open question. Two days of real-world use answered it: nothing in the blunt block's message, `--json` output, or docs mentions that the variable exists, so no one discovers it. In practice every clone still hits the always-block wall - today a second clone of a personal database showed the full v49 -> v53 blunt refusal and the operator had to research designated-migrator semantics by hand, even though the smart gate sitting dormant in the same binary would have routed the case correctly.

A safety feature that is provably at-least-as-safe as the behavior it replaces, but that nobody can find, protects no one. This flips the default.

## What

- `SmartGateEnabled()` now returns true when `BD_SMART_GATE` is unset. `BD_SMART_GATE=0` (any parseable boolean false) opts out and restores the unconditional #4515 block. An unparseable value keeps the default rather than silently disabling the gate - a typo must not turn off a safety feature.
- No routing changes. The verdict table from #4524 is untouched, and every fallback path (unreadable cached remote state, below the convergence floor, opted out) still degrades to the blunt block, whose output remains byte-identical to #4515.
- Blunt-gate tests (`schema`, `dolt`, `embeddeddolt`) now pin `BD_SMART_GATE=0` explicitly, since "unset" no longer means "blunt". New tests pin the default-on contract: a `SmartGateEnabled` truth table (unset/1/true/0/false/FALSE/garbage) and a routing test proving an unset env resolves the safe first-mover.
- Docs: the upgrade guide's remote-backed section now describes the default routing and the opt-out; CHANGELOG entry under Unreleased. The existing recipes are kept as the explicit path - they work identically in either mode.

## Safety argument (unchanged from #4524)

The only case that auto-resolves is remote-at-same-version with no content skew at/above the convergence floor: concurrent first-movers re-converge to byte-identical tables via migration 0050, so auto-migrating cannot fork. Remote-ahead still stops with an adopt directive (adoption is destructive, never silent), content skew still stops for a human, and anything unreadable falls back to the blunt block.

## Testing

- `CGO_ENABLED=0 go build -tags gms_pure_go ./...`
- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/schema/` - all gate tests pass, including the new default-on routing and truth-table tests
- `go test ./internal/storage/schema/` (default CGO test build) - pass
- `go vet` on the three touched storage packages - clean
- Server-mode and embedded real-Dolt gate tests need the CI harness (test Dolt server / cgo); the changes there are test-env pinning only. Full `make test` queued locally.

## Follow-up candidates (out of scope here)

- Surface smart-gate state in the blunt block when it fires as a fallback ("smart gate could not read the remote's cached schema state"), so the remaining blunt cases are self-explaining.
- The non-destructive fast-forward adopt for the remote-ahead case, as noted in #4524's open questions.

Closes #4516.

https://claude.ai/code/session_0167xsjJDukVKE7BE1i4W2wE

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **The state-aware remote-migrate gate is now on by default.** rc.2 shipped
+  the smart gate behind an opt-in env var, but nothing in the gate's block
+  message or docs surfaced that the var existed, so in practice everyone kept
+  hitting the blunt always-block behavior. The provably-safe first-mover case
+  (remote at the same schema version as this clone) now auto-migrates without
+  `BD_ALLOW_REMOTE_MIGRATE`; remote-ahead still stops with an adopt directive
+  and content skew still stops for a human. Set `BD_SMART_GATE=0` to opt out
+  and restore the unconditional block
+  ([#4516](https://github.com/gastownhall/beads/issues/4516)).
+
 ## [1.1.0-rc.2] - 2026-07-02
 
 Second release candidate for 1.1.0. Fixes the two upgrade-breaking migration
--- a/cmd/bd/init_embedded_test.go
+++ b/cmd/bd/init_embedded_test.go
@@ -98,6 +98,18 @@ func bdEnv(dir string) []string {
 	)
 }
 
+// envWithout returns env minus any entries for the named variable.
+func envWithout(env []string, name string) []string {
+	out := make([]string, 0, len(env))
+	for _, e := range env {
+		if strings.HasPrefix(e, name+"=") {
+			continue
+		}
+		out = append(out, e)
+	}
+	return out
+}
+
 func isEmbeddedLockOutput(out string) bool {
 	out = strings.ToLower(out)
 	return strings.Contains(out, "one writer at a time") ||
@@ -671,11 +683,16 @@ func TestEmbeddedInit(t *testing.T) {
 		}
 	})
 
-	t.Run("remote_behind_schema_gates_with_guidance", func(t *testing.T) {
-		// bd-4mpy7: bootstrapping from a remote whose database is behind this
-		// binary's schema must fail with designated-migrator guidance and
-		// leave a finalized workspace where the guidance commands can run —
-		// not a half-initialized directory with a raw gate error.
+	t.Run("remote_behind_schema_gate", func(t *testing.T) {
+		// bd-4mpy7 / #4516: bootstrapping from a remote whose database is
+		// behind this binary's schema. Shared fixture: a published remote
+		// regressed one migration below LatestVersion. Two paths against it:
+		// the default smart gate auto-migrates the clone as a safe
+		// first-mover (remote at the same version — no one has migrated),
+		// while the BD_SMART_GATE=0 opt-out must fail with
+		// designated-migrator guidance and leave a finalized workspace where
+		// the guidance commands can run — not a half-initialized directory
+		// with a raw gate error.
 		remoteDir := filepath.Join(t.TempDir(), "behind-remote")
 		remoteURL := "file://" + remoteDir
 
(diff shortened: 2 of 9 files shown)


maphew:
I found one merge blocker in the test coverage:

- `cmd/bd/init_embedded_test.go` still assumes the old blunt gate in `TestEmbeddedInit/remote_behind_schema_gates_with_guidance`. The test sets `BD_ALLOW_REMOTE_MIGRATE=0` but leaves `BD_SMART_GATE` unset, so on this branch it exercises the new default-on smart path and succeeds as a safe first-mover. CI shows the failure in `Test (Embedded Dolt Cmd 15/20)`:

  ```text
  Smart gate (BD_SMART_GATE): auto-applying 1 pending deterministic schema migration to a remote-backed database (v52, remote at same version -- safe first-mover, concurrent migrators converge; #4516). Run `bd dolt push` after.
  init_embedded_test.go:715: bd init --remote against a behind-schema remote should fail
  ```

  If this test is meant to preserve blunt-gate guidance coverage, it should also set `BD_SMART_GATE=0` beside `BD_ALLOW_REMOTE_MIGRATE=0`. If it is meant to cover the default behavior, it needs to assert the new auto-migrate result and leave separate opt-out coverage for the blunt message.

Everything else I checked lines up with the PR's stated behavior. `go test ./internal/storage/schema/` passes locally, and the schema-level default-on/opt-out cases look covered.

maphew:
Good catch - fixed in 3729f5567, taking the "both" option: the test now splits over the shared behind-remote fixture into

- `default_smart_gate_auto_migrates_first_mover` - asserts the new default end-to-end: `bd init --remote` against a same-version-behind remote succeeds as the safe first-mover, prints the smart-gate notice directing a follow-up `bd dolt push`, and the clone is immediately usable (`bd list` sees the source issue with no unlock). It strips any ambient `BD_SMART_GATE` from the env so it genuinely exercises the unset default.
- `opt_out_gates_with_guidance` - the original blunt-gate coverage unchanged in substance, now under `BD_SMART_GATE=0`: init fails with designated-migrator guidance, leaves a finalized workspace, and the guided `BD_ALLOW_REMOTE_MIGRATE=1 bd migrate` recovery works.

Both subtests pass locally under `CGO_ENABLED=1 BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go -run "TestEmbeddedInit/remote_behind_schema_gate" ./cmd/bd`. The branch is also rebased onto current main (it now sits atop the merged #4552).

As a side note, this subtest turned out to be the first real-database end-to-end check of the default flip - the fresh-clone-adopts-behind-remote flow is exactly the scenario that motivated default-on.

maphew:
Follow-up candidate No.1 from this PR's description has landed: #4573 (merged as 94dd4dcf4) makes the blunt fallback block self-explaining - one appended human-message line plus a `fallback_reason` JSON field naming which case degraded to the blunt stop (unreadable-remote-state, below-convergence-floor, opted-out, or unparseable-env). Verdict table and the #4515 block body are unchanged.

Follow-up candidate No.2 (non-destructive fast-forward adopt for remote-ahead) now has a full design written down in the maintainer tracker (mybd-ae1i): new `smartAdoptFastForward` verdict as a strict refinement of the adopt stop, driver-side ancestry primitives, and a three-piece landable split. Implementation not started yet.

PR #4556, opened 2026-07-02
fix(codex): stop noisy prompt refresh hook

christso:
## Withdrawn

This PR is withdrawn.

Follow-up testing showed the installed Codex marketplace plugin was still on the original hook manifest. Normal prompts did not inject Beads context on every turn.

`UserPromptSubmit` still runs every prompt, but `bd codex-hook UserPromptSubmit` only emits Beads context when a post-compaction refresh marker exists. The remaining behavior is a transient status label, not repeated context insertion.

The branch has been restored to the original hook behavior.


christso:
Withdrawing this.

The installed Codex marketplace plugin was still using the original hook manifest, and normal prompts did not inject Beads context every turn. `UserPromptSubmit` does run on each prompt, but `bd codex-hook UserPromptSubmit` only emits Beads context when a post-compaction refresh marker exists.

The branch has been restored to the original hook behavior. This is not the context-waste defect I originally reported.

christso:
Closing as withdrawn. The branch has been restored to original hook behavior.

PR #4558, opened 2026-07-03
fix(schema): repair v53 wisp dependency drift

### LINKED ISSUE #4555: migration 0053_repair_rig_wisps fails: wd.depends_on_issue_id missing (rc.2)
## Summary

Migration `0053_repair_rig_wisps` fails on a remote-backed embedded Dolt database because it references a column (`depends_on_issue_id`) that does not exist on the `dependencies` table (`wd`). The failure leaves four tables in a dirty half-migrated state that no `bd` command can recover from, because every command triggers auto-migrate-on-open and dies on the dirty check.

## Environment

- `bd version`: `1.1.0-rc.2 (bc9f10a5b: HEAD@bc9f10a5bc16)`
- OS: Windows 11 (MINGW64_NT-10.0-26200, MSys x86_64)
- Backend: embedded Dolt, sync.remote = `git+https://github.com/maphew/mdo.git`
- DB was created on this machine; this is the only clone (designated migrator)

## Reproduction

```sh
BD_ALLOW_REMOTE_MIGRATE=1 bd migrate
```

Starting schema version: **v49** (from a DB last touched by an older bd).

## What happened

1. `BD_ALLOW_REMOTE_MIGRATE=1 bd migrate` ran migrations v49 → v52 successfully.
2. Migration **v53** (`0053_repair_rig_wisps.up.sql`) failed:

   ```
   Error: failed to open database: embeddeddolt: init schema: embeddeddolt: migrate:
     migration 0053_repair_rig_wisps.up.sql: Error 1105: table "wd" does not have column "depends_on_issue_id"
   ```

3. The failed run left tables `comments`, `dependencies`, `events`, `issues` in a **dirty** state.
4. Every subsequent `bd` command now fails on open with:

   ```
   Error: failed to open database: embeddeddolt: init schema: embeddeddolt: migrate:
     pending schema migrations alter pre-existing dirty tables: comments, dependencies, events, issues
   ```

## Why this is a hard trap

No `bd` command can repair the database, because each one runs auto-migrate-on-open and hits the dirty-table check before any repair logic runs. Attempted recovery paths, all blocked by the same open-time failure:

- `bd migrate` → fails as above
- `bd migrate schema` (documented idempotent/recovery path) → fails as above
- `bd --ignore-schema-skew ...` → auto-migrate runs before the flag can take effect
- `BD_ALLOW_REMOTE_MIGRATE=1 bd migrate --inspect` → fails (can't open)
- `bd doctor` → "not yet supported in embedded mode"
- `bd sql "SHOW COLUMNS FROM dependencies"` → can't open DB

## Hypothesis

Migration 0053 expects the `dependencies` table (aliased `wd`?) to already have a `depends_on_issue_id` column, but no prior migration (v49 → v52) added it for this DB. This suggests either:

- the migration assumes base-schema state that an older `bd` version did not create, or
- a prior migration that should add the column is missing from the v49→v52 set, or
- the column was expected to be added by 0053 itself but a preceding `ALTER` in the same file didn't land before the statement that uses it.

## Workaround / data safety

I exported all 40 issues + 1 memory via `bd export --all -o backup.jsonl` **before** the dirty state set in, so data is safe. Planned recovery once a fix is available: `bd admin reset` + `bd bootstrap` / `bd import` from JSONL.

## Ask

- Guidance on repairing the dirty state in place without a full nuke-and-reimport (ideally an env var or flag to **skip auto-migrate-on-open**, so `bd migrate schema` or manual SQL can run).
- Confirmation of whether 0053 should be re-ordered or whether a prior migration needs backfilling for DBs coming from v49.

Happy to provide the JSONL export or more diagnostics if helpful.

maphew:
## Why

Fixes #4555.

Migration `0053_repair_rig_wisps` can run on mixed-vintage embedded Dolt databases where the clone-local `wisp_dependencies` table exists and has the newer `id` column, but still lacks one or more split target columns. The migration selected the `id` branch and then read `depends_on_issue_id`, `depends_on_wisp_id`, and `depends_on_external`, which caused the migration to fail before later code could repair the database.

Already-failed databases also get stuck behind the dirty-table guard because opening the database retries migrations before any manual repair command can run.

## What Changed

- Extends the v53 pre-migration repair hook to add missing `wisp_dependencies` split target columns and backfill them from legacy `depends_on_id` when available.
- Adds a narrow failed-v53 recovery gate that ignores dirty tables only when the main schema is at v52, every dirty committable table is one that v53 can dirty, and `wisp_dependencies` still needs the split-target repair.
- Adds sqlmock regression coverage for the new repair and recovery checks.

## Validation

- `go test ./internal/storage/schema -count=1`
- `git diff --check`

I also ran `go test ./internal/storage/...`; it failed outside this change on Windows/environment prerequisites: Unix absolute-path expectations in `internal/storage/dbproxy/server`, a temp `server.log` cleanup lock, and missing ICU headers for embedded Dolt CGO packages.

### DIFF
--- a/internal/storage/schema/lock_test.go
+++ b/internal/storage/schema/lock_test.go
@@ -121,7 +121,7 @@ func expectOnePendingMigration(t *testing.T, mock sqlmock.Sqlmock) {
 		// issue columns and no local wisp_dependencies table, so no ALTERs follow.
 		for _, col := range []string{"hook_bead", "role_bead", "agent_state", "last_activity", "role_type", "rig"} {
 			mock.ExpectQuery(`SELECT COUNT\(\*\) FROM INFORMATION_SCHEMA\.COLUMNS`).
-				WithArgs(col).
+				WithArgs("issues", col).
 				WillReturnRows(sqlmock.NewRows([]string{"count"}).AddRow(1))
 		}
 		mock.ExpectQuery(`SELECT COUNT\(\*\) FROM INFORMATION_SCHEMA\.TABLES`).
--- a/internal/storage/schema/lock_test.go
+++ b/internal/storage/schema/lock_test.go
@@ -116,13 +116,17 @@ func expectOnePendingMigration(t *testing.T, mock sqlmock.Sqlmock) {
 	expectContentHashColumnExists(mock)
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM schema_migrations", "version", latest-1)
 	if latest == 53 {
-		// The v53 pre-repair (#4502) probes the six rig/agent columns on
-		// issues; this mocked world has them all, so no ALTERs follow.
+		// The v53 pre-repair probes the six rig/agent columns on issues and
+		// then the local wisp_dependencies table; this mocked world has all
+		// issue columns and no local wisp_dependencies table, so no ALTERs follow.
 		for _, col := range []string{"hook_bead", "role_bead", "agent_state", "last_activity", "role_type", "rig"} {
 			mock.ExpectQuery(`SELECT COUNT\(\*\) FROM INFORMATION_SCHEMA\.COLUMNS`).
 				WithArgs(col).
 				WillReturnRows(sqlmock.NewRows([]string{"count"}).AddRow(1))
 		}
+		mock.ExpectQuery(`SELECT COUNT\(\*\) FROM INFORMATION_SCHEMA\.TABLES`).
+			WithArgs("wisp_dependencies").
+			WillReturnRows(sqlmock.NewRows([]string{"count"}).AddRow(0))
 	}
 	mock.ExpectExec("(?s).*").
 		WillReturnResult(sqlmock.NewResult(0, 0))
--- a/internal/storage/schema/migration_repairs.go
+++ b/internal/storage/schema/migration_repairs.go
@@ -18,7 +18,10 @@ import (
 // preMigrationRepair dispatches any repair registered for (source, version).
 func (m migrationSource) preMigrationRepair(ctx context.Context, db DBConn, version int) error {
 	if m.cursorTable == "schema_migrations" && version == 53 {
-		return ensureIssuesRigColumns(ctx, db)
+		if err := ensureIssuesRigColumns(ctx, db); err != nil {
+			return err
+		}
+		return ensureWispDependenciesSplitTargets(ctx, db)
 	}
 	return nil
 }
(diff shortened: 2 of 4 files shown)


maphew:
This cleanly unblocks the #4555 wedge — nice work. The backfill classification faithfully reproduces the shipped ignored `0003` / main `0047` UPDATE sequence (external-prefix → wisp-join → issue-join → catch-all), so it inherits the same ordering semantics rather than introducing a new classification rule. The extra `IS NULL` guards are a correct improvement for re-runnability over the single-shot canonical block.

The dirty-table recovery gate is the heart of this and it's done right: `failed0053DirtyTablesAreRecoverable` runs after `committableDirtyTables` strips the ignored `wisp_*`/`wisps` tables and before `pendingMigrationDirtyTables` would reject them, so it's positioned early enough to actually let the repair fire on an already-failed DB. The three-way gate (version == 52 AND dirty tables ⊆ {child_counters, comments, dependencies, events, issues, labels} AND wisp_dependencies still needs repair) is tight — I confirmed the allowlist is exactly the non-ignored table set that 0053 writes, and the `wispDependenciesNeed0053Repair` conjunct is what stops a genuinely-corrupt DB from being silently recovered past the dirty guard. Ordering against `ensureIssuesRigColumns` is safe since the issue-join only reads `issues.id`.

Approving. A few optional hardening suggestions, none blocking:

- The sqlmock tests assert the four UPDATEs fire in order but can't prove the *classification* resolves id collisions deterministically (wisp-before-issue). The invariant is currently carried by convention-matching with 0003/0047. A real Dolt-backed test (alongside `TestMigration0053RepairsRigWispsThroughDoltCLI`) seeded with a `depends_on_id` that matches both a wisp and an issue would lock that in end-to-end.
- The recovery gate's safety claim — "reject when wisp_dependencies does NOT need repair, even at v52 with allowlisted dirty tables" — isn't tested directly. A subtest where all three split columns are already present and the gate still returns false would be valuable, since that's the property keeping unrelated v52 dirty-table conflicts from being masked.
- A wrong-version rejection subtest (`current != 52`) for the gate would round out coverage of the three conjuncts in isolation.

Minor: `ensureIssuesRigColumns` could now reuse the new `schemaColumnExists` helper instead of its inline probe, but that's pure cleanup for a follow-up.

Note on local validation: the `make test` run on the verify queue failed only on environmental packages (release-script tests need a prebuilt `bd`/`go.mod` on PATH in the clean worktree; `repro-dolt-prod-timeouts` needs a built `bd` binary). The `internal/storage/schema` package itself passed, as did the broader storage/conformance/cmd suites on CI.

_zcode-unknown-model-unknown-reasoning on behalf of matt wilkie_

Issue #4547, opened 2026-07-02
Proposal: pluggable storage backends — one seam, capability-gated commands, Postgres first


julianknutsen:
## Goal

Let users choose their storage backend, while `bd` keeps ONE core command set that behaves
identically everywhere:

- `bd init --backend=dolt|postgres|sqlite` — engine chosen at init, recorded in the workspace
  locator; opening reads the locator, never the environment.
- **Core commands** (create/list/show/ready/update/close/dep/query/…) work the same on every
  backend — proven by a differential conformance harness, not code review.
- **Storage-specific commands are capability-gated addons**: `bd history`, `bd dolt push/pull`,
  `bd vc`, `bd branch`, `bd diff`, `bd federation`, `bd backup`, `bd sql` exist when the opened
  store advertises the capability, and are deterministic stubs otherwise. Pick Dolt → you get
  history and remotes. Pick Postgres → those commands stub out (v1).
- **First proof:** beads running on a Postgres backend, with the CLI simply not having the
  history functionality.

Dolt stays the default and richest backend. Existing workspaces keep opening exactly as today.

## Where we are: three plumbings, one engine

Today there is only ONE engine (Dolt) reached three different ways, through TWO disjoint
code paths:

```
                     cmd/bd (~271 commands)
                            │
        ┌───────────────────┴─────────────────────────┐
        │ every command                  13 commands ALSO have a second body,
        ▼                                picked by usesProxiedServer():
  global store (storage.DoltStorage,                  ▼
  the 144-method composed interface)          global uowProvider
        │                                             │
  ┌─────┴──────┐                              uow.UnitOfWork (one tx per command)
  ▼            ▼                                      │
EmbeddedDolt  DoltStore                       domain use-cases (IssueUseCase…)
(in-process,  (MySQL wire to                          │
 CGO, flock=  dolt sql-server)                domain/db SQL repositories
 1 writer)         │                                  │
  └─────┬──────────┘                          db.Runner → dbproxy → dolt sql-server
        ▼                                             │
  issueops/versioncontrolops                          │
  (shared MySQL/Dolt SQL, ~12.5k LOC)                 │
        └──────────────────┬──────────────────────────┘
                           ▼
                  the same Dolt database
```

1. **Embedded** — in-process engine, single-writer flock. 2. **Server** — `DoltStore` speaks
MySQL wire to a `dolt sql-server`. 3. **Proxied-server** — the newer unit-of-work stack
[...]

PR #4560, opened 2026-07-03
store: unify command paths behind Store interface, add OCC, remove 13 duals

### LINKED ISSUE #4517: Write commands should verify durable row state before reporting success
## Summary

`bd` write commands such as close/update/note should not print success until the durable canonical row (or an equivalent commit/transaction invariant) has been verified.

In a private multi-agent Gas City workspace, we observed success output for close-style writes where the requested state was not the durable final row state. The dangerous behavior is the success surface: agents and orchestration trust `rc=0` / `✓ Closed`, but the canonical issue row can later show a different status or a different close reason.

## Why this matters

This is a data-integrity issue for AI-agent workflows. Beads is treated as the durable work store; if a write command can return success without proving the requested mutation persisted, agents make decisions from false state.

## Observed classes

- A close command printed success, but later readback showed the issue still open/blocked.
- Concurrent close-style writes can both return success, while the final durable row can only contain one close reason. The losing caller still saw success.
- Events/audit may record attempted writes even when the durable issue row does not match the requested mutation.
- The workspace is a managed multi-writer Dolt server setup; endpoint topology was separately verified healthy/canonical, so this should not be solved by starting ad hoc Dolt servers or changing endpoint ownership.

## Requested behavior

For close/update/note-like writes:

1. Success means the durable row now matches the requested status/reason/field change, or a documented equivalent invariant has been verified.
2. Serialization losers, stale endpoint, Dolt unreachable, branch/transaction races, and post-write mismatches must not be hidden behind success output.
3. If retry is safe and canonical for the command, retry. Otherwise exit nonzero and report what persisted instead.
4. Programmatic callers need a clear machine-detectable failure when requested state != durable state.

## Environment where observed

- `bd version 1.0.5 (Homebrew)`
- `dolt version 2.1.4`
- Gas City managed Dolt server topology, multiple concurrent agents/writers

I can provide private workspace evidence to maintainers if needed, but avoided publishing customer/workspace details here.

### LINKED ISSUE #4541: bug(update): --notes silently replaces the field; agent fleets lose audit history
## Summary

`bd update <id> --notes "..."` **replaces** the `notes` field with no warning, no confirmation, no backup. For agent fleets that treat `notes` as an append-only audit surface (chain decisions, escalations, handoffs), every call wipes the prior thread. The data loss is **silent** — no stderr, no exit-code change, no permission prompt.

## Repro

```bash
$ bd create "demo" -t task -p 2 --json | jq -r '.[0].id'
bd-abc

$ bd update bd-abc --append-notes "agent A: Tier-1 passed"
$ bd update bd-abc --append-notes "agent B: Tier-2 scope check OK"

$ bd show bd-abc --json | jq '.[0].notes'
"agent A: Tier-1 passed\nagent B: Tier-2 scope check OK"

# So far so good. Now the footgun:

$ bd update bd-abc --notes "agent C: re-running Tier-2"
$ bd show bd-abc --json | jq '.[0].notes'
"agent C: re-running Tier-2"

# agent A and agent B's contributions are gone.
```

## Impact

- **Silent.** No permission prompt (unlike the heredoc case in #3102), no shell error, no stderr. The next `bd show` returns the new content as if the prior thread never existed.
- **Scales to fleets.** Any agent that conflates "update" with "append" hits it. We hit it in our own coding-orchestrator: it was calling `bd update <id> --notes` for chain decisions and silently destroying the prior thread on every bead close. The same footgun will hit every chain driver, dispatch loop, and CI automation that treats `notes` as an append surface.
- **Dolt history hides it.** The prior notes content is in Dolt commit history (`bd history <id>` / `bd dolt log`), so it is technically recoverable — but no agent reads that on resume. From the agent's perspective the data is lost.

## Why `--append-notes` is not enough

The `--append-notes` flag exists and is the documented workaround. But agents reading `bd update --help` see `--notes` first, infer "this is the field I want to update," and call it. The shorter flag name wins by default in agent prompts; the safer flag requires explicit awareness. We now mandate `bd note <id> "..."` (the `--append-notes` shorthand) and `bd comment <id> "..."` (the attributed-thread primitive) in our agent policy, but that is local hardening — it does not prevent the next fleet from hitting the same trap.

## Suggested fix

Pick one (in order of preference):

1. **Default `--notes` to append for agents.** Add `--replace-notes` (or `--notes-replace`) as the explicit destructive flag. Matches the principle of least surprise for a field most users treat as append-only.
2. **Add a non-destructive alias** (`--notes-append` / `--add-notes`) that is shorter / more discoverable than `--append-notes`.
3. **Add a strict-mode lint warning** when `--notes` is used on a bead whose prior `notes` field is non-empty — print to stderr but proceed. Agents running `--strict` or with `BD_SAFE=1` would refuse.

Any of these is a few lines of code. #3102 (open feature request) adds file-flag variants but preserves the destructive default — the same fleets that misuse `--notes` will misuse `--notes-file`.

## Cross-references

- **#3102** — open feature request for `--notes-file` / `--append-notes-file` (related; not a replacement for this bug)
- **CLI_REFERENCE.md** — currently documents `--notes` as the primary way to set notes, with `--append-notes` as an alternative

## Local mitigation we adopted

In case it helps other fleets (from our `AGENTS.md`):

```text
# Append-only shorthand — survives across calls
bd note <id> "..."              # alias for --append-notes

# Attributed, timestamped audit log — for handoffs / gate outcomes / escalations
bd comment <id> "..." --author <agent>

# Never: bd update <id> --notes "..."  (replaces the field)
```

We pinned this in agent policy and now mandate `bd comment` for any cross-agent audit trail. Happy to send a PR for option (1) or (3) above if helpful.

Amilliox:
The three storage backends (embedded, server, proxied) each had their own code path in cmd/bd, forcing 13 commands into dual files leaving the rest tied to the 144-method DoltStorage interface.\n\nThis adds a thin Store seam (RunRead/RunWrite) wrapping the existing UoW. All three backends now live behind it.\n\nConcurrent writes: update checks updated_at before writing (closes #4517), notes append instead of replacing (closes #4541).\n\nOracle A harness compares old vs new binary across 10 scenarios for refactoring safety.\n\nSee related proposal: #4547

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ SHELL := $(subst cmd,bin,$(subst git.exe,bash.exe,$(GIT_BASH)))
 endif
 endif
 
-.PHONY: all build test test-icu-path test-full-cgo test-regression test-upgrade test-cross-version test-migration bench bench-quick clean clean-test-tmp install install-force help check-up-to-date fmt fmt-check check-testing-short
+.PHONY: all build test test-icu-path test-full-cgo test-regression oracle-a test-upgrade test-cross-version test-migration bench bench-quick clean clean-test-tmp install install-force help check-up-to-date fmt fmt-check check-testing-short
 .PHONY: ci-pr-core ci-pr-policy ci-pr-lint ci-package-mcp ci-package-npm ci-website
 
 # Default target
@@ -104,6 +104,12 @@ ci-website:
 test-regression:
 	@echo "Running regression tests (baseline vs candidate)..."
 	go test -tags=regression,$(BUILD_TAGS) -timeout=$(REGRESSION_TIMEOUT) -v ./tests/regression/...
+
+# Run Oracle A conformance harness (10 basic scenarios, exit+stdout comparison).
+# Builds both binaries, compares output with normalization.
+oracle-a: build
+	@echo "Running Oracle A conformance harness..."
+	@go test -tags=regression,$(BUILD_TAGS) -timeout=10m -run TestOracleA -count=1 -v ./tests/regression/...
 
 # Run upgrade smoke tests (release stability gate).
 # Tests that upgrading from previous release preserves data, role, and mode.
--- a/cmd/bd/close_proxied_server.go
+++ b/cmd/bd/close_proxied_server.go
@@ -36,42 +36,70 @@ type closeProxiedOutcome struct {
 }
 
 func runCloseProxiedServer(cmd *cobra.Command, ctx context.Context, args []string) {
+	in, reasons, err := parseCloseProxiedArgs(cmd, args)
+	if err != nil {
+		FatalErrorRespectJSON("%v", err)
+	}
+
+	uw, err := openCloseProxiedUOW(ctx)
+	if err != nil {
+		FatalErrorRespectJSON("%v", err)
+	}
+	defer uw.Close(ctx)
+
+	outcomes, closedIssues := processCloseProxiedIssues(ctx, uw, args, reasons, in)
+	unblocked, continueResult, claimedNextIssue := processCloseProxiedPostActions(ctx, uw, args, outcomes, in)
+
+	commitCloseProxiedWithHooks(ctx, uw, outcomes, claimedNextIssue, continueResult)
+	renderCloseProxiedOutput(outcomes, closedIssues, unblocked, continueResult, claimedNextIssue, in)
+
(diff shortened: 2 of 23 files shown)


julianknutsen:
Thank you for your submission, but we are going to land the full branch already in progress for this.

PR #4561, opened 2026-07-03
Add configured backend API and external backend plugins

duncan4123:
## Summary

This PR gives Beads one reusable, backend-neutral opening path for built-in and external storage backends while keeping Dolt as the default.

- Adds public `beads.OpenConfigured(...)`, returning the opened `Storage` and a backend descriptor.
- Adds backend capability modeling for transactions, leases, raw SQL, maintenance, versioning, remotes, and concurrent writers.
- Registers the built-in Dolt, PostgreSQL, MySQL, and SQLite providers behind the same internal factory.
- Adds a versioned external backend protocol and process-backed `storage.DoltStorage` adapter.
- Makes `OpenFromConfig`, `OpenBestAvailable`, and CLI configured-store construction delegate to the shared factory.
- Supports configured backends in both CGO and non-CGO builds; only embedded Dolt requires CGO.
- Preserves existing Dolt command-specific startup, gateway, proxy, and migration behavior.

## Backend selection and trust

The backend name remains workspace metadata, but executable plugin commands are resolved only from local/user trust configuration or environment. A committed `metadata.json` cannot authorize code execution.

For a non-Dolt backend, Beads first checks for a locally trusted external provider, allowing a plugin to implement or override that backend. Otherwise it uses a registered built-in provider. Unknown backends fail closed when no trusted provider exists.

The legacy committed plugin command fields are retained only for configuration compatibility and are never executed.

## Public API

```go
store, info, err := beads.OpenConfigured(ctx, beadsDir, beads.OpenConfiguredOptions{
    ReadOnly: true,
})
```

`info` identifies the selected backend, whether it is external, and its advertised capabilities. Callers such as Gas City can use this information to select supported behavior without importing Beads' internal database implementations.

`OpenFromConfig` remains source-compatible and now honors all configured backends by delegating to `OpenConfigured`.

## External plugin boundary

Beads owns backend selection, local trust resolution, process lifecycle, protocol negotiation, and the storage adapter. Backend-specific implementation remains outside the repository.

A DoltLite proof implementation is available at https://github.com/duncan4123/beads-backend-doltlite.

## Verification

Passed focused CGO coverage for the public API, configured factory, plugin trust/read-only behavior, and configuration normalization.

Passed focused non-CGO coverage for configured backend registration and embedded-Dolt failure guidance.

The initialization metrics regression passed independently across embedded, server, shared-server, and proxied-server modes. The complete `cmd/bd` package run exceeded its cumulative 10-minute timeout under parallel load; the test active at timeout passed in isolation in 166 seconds.


maphew (commented):
Thanks @duncan4123 - this is a serious, well-shaped piece of work: the provider registry is exactly the seam we want for backend construction, defaults are preserved when no plugin is configured, the docs sketch is honest about scope, and the out-of-tree DoltLite proof (duncan4123/beads-backend-doltlite) demonstrating the boundary end-to-end is exactly the right way to propose a plugin architecture. We also had in-progress work on this same registry + `GetBackend` un-stub seam, and your PR takes priority; we'll build on this branch rather than compete with it.

There are two blocker-level issues plus a few smaller items we need to resolve before merge. We expect to handle most of these as maintainer commits on your branch rather than asking for rounds of changes, but flagging them so the reasoning is visible:

**1. Plugin command trust (blocker).** `pluginprocess.Start` execs `backend_plugin_command` from `.beads/metadata.json` (`internal/backend/pluginprocess/client.go`, the `#nosec G204 - trusted project metadata` line). But metadata.json is committed to git in supported workflows (our inherited-artifacts flows commit it; the canonical `.beads/.gitignore` excludes only the db and runtime files, not metadata.json), and `bd` is frequently auto-run by agent hooks on session start. `bd prime` still opens the store lazily via `formatMemoriesForPrime` -> `ensureStoreActive` -> `newDoltStoreFromConfig`, so even the hook path execs the plugin. As written, cloning a hostile repo and running bd (including hook-fired `bd prime`) executes an attacker-controlled command. The plugin command needs to come from a non-committed trust source - local-only config, user-global config.yaml, an env var, or an explicit one-time trust confirmation per repo (git's `core.hooksPath`-style posture). Happy to discuss which shape you prefer.

**2. Read-only sessions (blocker).** `OpenParams` has no read-only concept, so `newReadOnlyStoreFromConfig`'s plugin branch (`cmd/bd/store_factory.go`) opens a fully writable plugin session, and the plugin branch in `cmd/bd/main.go` ignores `doltCfg.ReadOnly`. That silently drops the guarantee from #3231 that a read-only open of a foreign project must not mutate it. Adding `read_only` to the plugin `OpenParams`/`openParams` and plumbing it through both call sites should be enough for v1alpha1.

Smaller items:

- **Nil embedded interface** (`internal/backend/pluginprocess/store.go`): `Store` embeds a nil `storage.DoltStorage`, and exactly one method - `Sync` - is unimplemented, so `bd federation sync` against a plugin backend would nil-panic; future interface additions would panic too instead of failing to compile. You're at 143/144 methods, so dropping the embed and implementing `Sync` (even as a clean "unsupported" error) gets compiler enforcement almost for free.
- **`*WithCounts` methods return wrong data**: `SearchIssuesWithCounts` and `GetReadyWorkWithCounts` in `store.go` call the non-counting methods and wrap `IssueWithCounts{Issue: issue}`, silently dropping the dependency/comment/label counts that list/ready output relies on. Either make these first-class protocol methods or stop advertising them.
- **Error mapping**: plugin errors come back as flat strings, so `errors.Is`-based caller logic (`ErrNotFound`, `ErrAlreadyClaimed`, `ErrNotClaimable`, and friends) degrades across the process boundary and would break restore/rename/promote/claim control flow. A code-to-sentinel table is worth a v1alpha1 roadmap note at minimum.
- **Call lifecycle / hangs**: `client.request` checks `ctx` only before writing, then holds `c.mu` across a blocking `dec.Decode`; a plugin that never sends hello or wedges mid-request hangs the whole `bd` process. There's no startup/request timeout or health check, and `Close`'s `cmd.Wait()` can block if the child ignores stdin close or leaves a process tree (notably on Windows). Bounded graceful-shutdown-then-kill plus a request deadline would harden this.
- **`GetBackend` compat**: unknown stale values like `postgres` used to coerce to dolt (the deleted test asserted this deliberately) and now hard-fail at provider lookup. A warn-and-fallback, or a `bd doctor` fix, would protect existing repos with junk in that field.
- **Test depth**: the helper-process test is a nice pattern; crash-mid-request, protocol-version mismatch, and store_factory routing cases would round it out.

One owner-level design point to settle rather than a defect: `backend/plugin` aliases ~50 `internal/types`/`internal/storage` types as the public wire surface, and `pluginprocess.Store` claims the full `storage.DoltStorage` (versioning, federation, remotes, merge slots, advanced queries) - which effectively freezes those internal Dolt-shaped shapes as the plugin protocol. The `Capabilities` structs exist but don't yet gate command paths, so unsupported operations surface only after dispatch. That may be the pragmatic alpha choice, but we want to make that commitment deliberately - the fallback is to land the registry + `GetBackend` + store_factory routing now (which we want regardless) and iterate with you on a smaller, capability-gated, versioned backend contract before the protocol package goes public.

**Recommendation:** fix-and-merge. We'll aim to push the mechanical fixes (2, the `Sync`/`*WithCounts` items) to this branch if maintainer edits are enabled, and coordinate with you on the trust model for (1) since it shapes the config surface. CI hasn't run on this branch yet, so we'll kick that off as part of the same pass.

maphew:
Flagging @harry-miller-trimble as he was/is doing work on what form(s) our plugin architecture should take. We haven't chatted about it for some time and I'm unsure how his investigations have/are turning out.

_edit:_ my review above is about the shape of the current PR, it does not look into architectural considerations.

maphew:
Also see #4601

maphew:
Coordination note, as promised when we said this PR takes priority on the configured-backend seam: #4859 (the narrow name-keyed registry salvaged from the #4601 rollback) is moving toward merge. It deliberately claims only the in-process registration seam — no exec path, no capability model, no public `OpenConfigured` — so it does not compete with the plugin-process design here; rather, this PR would rebase onto it and register its provider through that seam instead of carrying its own registry. The exec-trust question that holds this PR is untouched by #4859. Flagging it now so the overlap is discovered here rather than via rebase conflict.

maphew (commented):
Reviewed at `d8de9872`.

Thanks @duncan4123—the provider/process proof remains valuable, and we intend to preserve it rather than compete with it. Rechecking the current head confirms that the trust, read-only, protocol-correctness, and lifecycle risks from the July 5 review remain unresolved; this comment does not duplicate those detailed findings.

The branch now also overlaps the narrower configured-backend registry in #4859. The safest consolidation is to transplant this PR's configured-backend opener, process-protocol proof, tests, and out-of-tree validation onto that registry, with explicit local trust and a smaller capability-gated interface. Preserve your commits/design and attribution, but do not merge this old branch unchanged.

bee-ghosttrack:
CI triage (bee, maintainer seat): Lint fails on two findings inside this diff — `internal/storage/issueops/close.go:37` (unparam: the `sqlite` parameter of closeIssueInTx is unused) and `cmd/bd/backend.go:150` (gosec G304, file inclusion via variable). The branch is also CONFLICTING with main now, so a rebase is needed before CI can give a real signal.

PR #4567, opened 2026-07-03
fix(embedded): let working-set reconcile commands open past dirty-table migration guard (#4566)

### LINKED ISSUE #4566: Embedded mode: dirty working set deadlocks schema migration — bd dolt commit can't clear it because it also triggers init schema
## Summary

In **embedded** mode, a dirty working set on the content tables (`comments`, `dependencies`, `events`, `issues`) deadlocks the v52→v53 schema migration *and* every recovery path. The dirty-table guard (`pendingMigrationDirtyTables`) correctly refuses to migrate, but `init schema` runs on **every** store open — including `bd dolt commit` / `bd vc commit` — so the user cannot commit the working set to clear the dirty state. There is no in-tool escape hatch (`--ignore-schema-skew`, `bd doctor --fix` unsupported in embedded mode, `bd admin` all do not bypass it).

This is a distinct trigger of the same gate discussed in #4378, but in **embedded** mode and from a **failed-migration recovery** rather than an out-of-band DB copy. The novel, actionable observation is the **deadlock**: the guard blocks the only command that could satisfy it.

Cross-ref: the local store reached this state during the #4555 window (rc.2 migration 0053 was failing); a prior recovery attempt left the four content tables with uncommitted working-set edits and no clean Dolt commit. The #4555 fix (now in main, bf4c7568c) lands correctly — with `BD_ALLOW_REMOTE_MIGRATE=1` the remote-backed guard is passed and the migration is genuinely attempted — only to hit this dirty-table gate underneath.

## Environment

- `bd version 1.1.0-rc.2 (bf4c7568c: main@bf4c7568ca5e)` (built from current beads `main` after #4555 merge)
- storage mode: **embedded** (`dolt_mode: embedded`, `.beads/embeddeddolt/`)
- OS: Windows 10 26200 x64
- git remote: `github.com/maphew/mdo`, `refs/dolt/data` present (remote-backed)

## Smallest command sequence

The store must already have (a) pending schema migrations and (b) a dirty working set on a table the migration alters. In this case v49→v53 was pending and `comments/dependencies/events/issues` were dirty in the working set (a leftover from a failed recovery during the #4555 window).

```bash
# any store-open triggers init schema, which attempts the migration:
BD_ALLOW_REMOTE_MIGRATE=1 bd migrate
# Error: failed to open database: embeddeddolt: init schema: embeddeddolt:
#        migrate: pending schema migrations alter pre-existing dirty tables:
#        comments, dependencies, events, issues

# the documented recovery — commit the working set first — also triggers init schema:
BD_ALLOW_REMOTE_MIGRATE=1 bd dolt commit -m "checkpoint before v53 migration"
# same error → deadlock: cannot commit because opening the store runs the migration
```

`bd dolt commit`, `bd vc commit`, `bd migrate schema`, and plain `bd list` all fail or warn-and-degrade through the same `init schema` path.

**Expected**: a path to reconcile the dirty working set short of wiping the store — e.g. `bd dolt commit` succeeds (committing the dirty tables at the *old* schema), after which `bd migrate` can run on a clean working set; or an explicit `--discard-working-set` / repair flag on `bd migrate`.

**Actual**: no in-tool recovery. Every store-open runs `init schema`, which refuses to migrate dirty tables, and the commit that would clean them also runs `init schema`.

## Workaround I used

Since the data was safe (JSONL export + Dolt `.darc` backups intact) and the remote was the canonical source:

1. `bd export --all -o backup.jsonl`  (pre-rebuild safety copy)
2. `mv .beads/embeddeddolt .beads/embeddeddolt.dirty-pre-v53`
3. `bd bootstrap --yes`   (re-clone from origin → clean working set, schema at remote's v49)
4. `BD_ALLOW_REMOTE_MIGRATE=1 bd migrate`   (v49→v53, applies cleanly)
5. `bd dolt push`   (publish v53 schema)

This worked and lost no data, but it requires the operator to recognize that the only fix is destroying and re-cloning the local store — and to trust that nothing local-only is unsynced. In embedded mode with a single machine, that's a real footgun.

## Suggested fix (for discussion)

One or both of:

1. **Don't run the dirty-table guard during commit opens.** `bd dolt commit` / `bd vc commit` opening the store should not require migration to complete — its whole purpose is to reconcile pending changes. Letting it commit at the current (old) schema would unblock the operator.
2. **An explicit `bd migrate --discard-dirty` / `--force`** that drops or checkpoints the dirty working set before applying the migration, with a clear warning. Useful when (as here) the dirty state is itself the debris of a prior failed migration rather than user data.

Option 1 seems preferable because it makes the existing documented recovery (`bd dolt commit`) actually function, without adding a footgun flag.

## Artifacts

- Preserved dirty store available for inspection: `.beads/embeddeddolt.dirty-pre-v53/`
- Pre-rebuild export (40 issues, stale relative to remote) and post-bootstrap export (69 issues) captured; remote was *ahead* of the local dirty store, so nothing was lost.

maphew:
## Why

In embedded mode, every store open runs `init schema` -> `schema.MigrateUp`, and MigrateUp correctly refuses to run when pending migrations would alter tables that already have uncommitted working-set changes. But the documented recovery for that refusal - commit the working set - is itself a `bd` command (`bd dolt commit` / `bd vc commit`) that opens the store and hits the same refusal. The guard blocks the only command that can satisfy it: a deadlock with no in-tool recovery, reported in #4566 (hit in the wild during the #4555 rc.2 window; the only escapes were raw `dolt add . && dolt commit` inside `.beads/embeddeddolt/<db>`, or destroying and re-cloning the store).

This implements option 1 from the issue discussion: make the working-set reconcile commands able to open the store without completing the migration, so the existing documented recovery actually functions. No new footgun flag.

## What

- **`schema.DirtyTablesError`** (new typed error): the pre-flight dirty-table guard in `MigrateUp` now returns this instead of a bare `fmt.Errorf`, and its message tells the operator the recovery path (`bd dolt commit`, then re-run the migration). The guard still fires before any migration statement runs, so a caller that skips on this error skips a fully un-started pass.
- **Open intents in embeddeddolt**: the previous `lenientGate bool` becomes a small open-intent enum (`openStrict` / `openReadOnlyCommand` / `openWorkingSetReconcile`). Read-only commands behave as before for the #4259 remote-migrate gate, and now also survive a dirty-table refusal (a read listing issues should not be bricked by a dirty working set either). The new `OpenForWorkingSetReconcile` intent gets warning text that matches what is actually happening ("committing the working set at the current schema") instead of the read-only wording, which would have been wrong for a commit.
- **`bd dolt commit` and `bd vc commit` open leniently**: a new `dolt.Config.LenientOpen` flag, set only for those two commands (leaf `commit` under `dolt`/`vc`), routes the embedded open through `OpenForWorkingSetReconcile`. On a dirty-table (or remote-gate) refusal the open warns, skips the migration, and lets the commit proceed at the current schema. After the commit the working set is clean and a normal `bd migrate` applies.

Deliberately unchanged:

- The **ignored-source** dirty guard stays a fatal plain error: it fires mid-pass after main-source migrations have applied, so a lenient skip there could checkpoint a half-applied pass. Dirty ignored tables are bd-internal state, not user data (commented in code).
- **Server mode** is untouched; `LenientOpen` is embedded-only and documented as such. `bd migrate` and `bd bootstrap` still hit the guard strictly - the typed error just makes the failure actionable.

## Recovery flow after this change

```bash
bd migrate
# Error: ... pending schema migrations alter pre-existing dirty tables: comments, dependencies, events, issues;
#        run 'bd dolt commit' to commit the working set at the current schema, then re-run the migration (#4566)

bd dolt commit -m "checkpoint before migration"   # now succeeds (was the deadlock)
bd migrate                                        # clean working set -> migration applies
```

## Testing

- `schema`: `TestMigrateUpReturnsDirtyTablesErrorForPreExistingDirtyTable` - full MigrateUp harness asserting the guard now yields `*DirtyTablesError` with the offending tables populated.
- `embeddeddolt`: `TestEmbeddedDirtyTablesGate_BlocksReopenThenWorkingSetReconcileRecovers` - end-to-end reproduction against real embedded Dolt (gated behind `BEADS_TEST_EMBEDDED_DOLT=1`, matching the sibling remote-migrate gate tests): store pinned at schema v51 with a dirty `issues` table -> plain `Open` fails with `DirtyTablesError` -> `OpenForWorkingSetReconcile` succeeds without migrating -> `Commit` clears the working set -> plain reopen migrates to latest. (Pinned to v51, not `LatestVersion()-2`, so future migrations cannot shift the start onto v52 where the failed-0053 recovery path would swallow the guard.)
- Existing remote-migrate gate tests pass against the reworked open path; no code string-matched the old guard message.
- `gofmt`, `go vet` (cgo and non-cgo stub), `go build ./cmd/bd`, full `./internal/storage/schema` package tests: clean.

Fixes #4566.

https://claude.ai/code/session_01FWVnjn3Bq4PSG1akpFpKnY

### DIFF
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -143,6 +143,25 @@ func isReadOnlyCommand(cmdName string) bool {
 	return readOnlyCommands[cmdName]
 }
 
+// isWorkingSetReconcileCommand reports whether cmd's whole purpose is to
+// reconcile the Dolt working set: "bd dolt commit" or "bd vc commit". These
+// commands are the documented recovery from a pending-migration dirty-table
+// refusal, but they also open the store, and an open runs the migration -
+// hitting that same refusal before the commit that would clear the dirty
+// state ever runs. Opening leniently (embeddeddolt.OpenForWorkingSetReconcile)
+// breaks that deadlock by skipping the migration instead of failing the open
+// (gastownhall/beads#4566).
+func isWorkingSetReconcileCommand(cmd *cobra.Command) bool {
+	if cmd.Name() != "commit" {
+		return false
+	}
+	parent := cmd.Parent()
+	if parent == nil {
+		return false
+	}
+	return parent.Name() == "dolt" || parent.Name() == "vc"
+}
+
 // loadBeadsEnvFile loads .beads/.env into process environment for per-project
 // Dolt credentials (GH#2520). Uses gotenv.Load which is non-overriding —
 // existing shell env vars always take precedence.
@@ -985,8 +1004,9 @@ var rootCmd = &cobra.Command{
 		// on a different filesystem (e.g., ext4 for performance on WSL).
 		doltPath := doltserver.ResolveDoltDir(beadsDir)
 		doltCfg := &dolt.Config{
-			ReadOnly: useReadOnly,
-			BeadsDir: beadsDir,
+			ReadOnly:    useReadOnly,
+			BeadsDir:    beadsDir,
+			LenientOpen: isWorkingSetReconcileCommand(cmd),
 		}
 
 		// Load config to get database name and server connection settings
--- a/cmd/bd/store_factory.go
+++ b/cmd/bd/store_factory.go
@@ -60,6 +60,13 @@ func newDoltStore(ctx context.Context, cfg *dolt.Config) (storage.DoltStorage, e
 		// already skip migration entirely.
 		return embeddeddolt.OpenForReadOnlyCommand(ctx, cfg.BeadsDir, cfg.Database, "main")
 	}
+	if cfg.LenientOpen {
+		// Working-set-reconcile commands (bd dolt commit, bd vc commit) must
+		// not be bricked by a pending-migration dirty-table refusal: that
+		// refusal's documented recovery is exactly the commit these commands
+		// run, so failing the open here would deadlock (#4566).
+		return embeddeddolt.OpenForWorkingSetReconcile(ctx, cfg.BeadsDir, cfg.Database, "main")
+	}
 	return embeddeddolt.Open(ctx, cfg.BeadsDir, cfg.Database, "main")
 }
 
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -207,6 +207,14 @@ type Config struct {
 	Database       string // Database name within Dolt (default: "beads")
 	ReadOnly       bool   // Open in read-only mode (skip schema init)
 
+	// LenientOpen opens the store leniently: embedded mode only. A migration
+	// gate refusal (#4259) or a dirty-working-set refusal (#4566) skips the
+	// migration instead of failing the open. Set for working-set-reconcile
+	// commands (bd dolt commit, bd vc commit; #4566), whose entire purpose is
+	// to clear the working set that the migration would otherwise refuse to
+	// touch. Ignored in server mode.
+	LenientOpen bool
+
 	// Server connection options
 	ServerSocket   string // Unix domain socket path (overrides Host/Port when set)
 	ServerHost     string // Server host (default: 127.0.0.1)
--- a/internal/storage/embeddeddolt/cache.go
+++ b/internal/storage/embeddeddolt/cache.go
@@ -29,7 +29,7 @@ type cacheEntry struct {
 // This prevents redundant engine initializations when multiple code paths open
 // connectors against the same data directory in the same process.
 func Open(ctx context.Context, beadsDir, database, branch string) (*EmbeddedDoltStore, error) {
-	return openCached(ctx, beadsDir, database, branch, false)
+	return openCached(ctx, beadsDir, database, branch, openStrict)
 }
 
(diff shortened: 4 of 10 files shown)


maphew (commented):
No code-review findings.

I reviewed the patch against #4566 and the storage/schema recovery path. The approach looks sound: `bd dolt commit` / `bd vc commit` get a lenient embedded open, skip only the safe pre-migration dirty-table refusal, then commit the working set at the current schema. The typed `schema.DirtyTablesError` is limited to that pre-flight guard, while the mid-pass ignored-table guard remains fatal, which avoids checkpointing a half-applied migration pass.

Validation I ran:

- `bd-main/scripts/pr-preflight.sh 4567 --repo gastownhall/beads`
- `CGO_ENABLED=0 go test ./cmd/bd -run TestIsDoltNothingToCommit`

I could not run the CGO embedded package locally because this machine is missing ICU headers for `github.com/dolthub/go-icu-regex` (`unicode/uregex.h`). The embedded end-to-end test behind `BEADS_TEST_EMBEDDED_DOLT=1` should still run in an ICU-capable CGO environment, but I do not see a blocker in the code.

Issue #3687, opened 2026-05-03
bd dolt stop falsely reports 'server not running' in shared-server mode


mieubrisse:
When run from a repo configured for shared-server mode (`dolt.shared-server: true` in `.beads/config.yaml`), `bd dolt stop` fails with `Error: dolt server is not running` — even when the shared server is actively running.

## Reproduction

1. From a repo with shared-server mode enabled, confirm the shared server is up:
   ```
   $ lsof -i :3308
   COMMAND   PID    USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
   dolt    14126 odyssey   37u  IPv4 0xbf7d48906bc89982      0t0  TCP localhost:3308 (LISTEN)
   ```
2. Run:
   ```
   $ bd dolt stop
   Error: dolt server is not running
   ```
3. Other `bd` subcommands in the same directory work against the shared server fine (`bd list`, `bd dolt show`, etc.)

## Expected

`bd dolt stop` should follow the same shared-server detection path the rest of `bd` uses, find the running shared dolt process (default port 3308, PID file under `~/.beads/shared-server/`), and stop it.

## Likely cause

The `stop` subcommand appears to look for a *per-repo* dolt process (probably a local PID file under `.beads/`) and exit early when it doesn't find one — without falling through to the shared-server lookup path.

## Impact

Blocks the standard rename-DB workflow (stop server → `mv` data dir → restart) when run from inside a project repo. Workarounds: stop via direct `kill` on the shared-server PID, or run from a non-bd working directory.

## Environment

- bd version: 0.61.0
- macOS (Darwin 25.2.0)
- Shared server at `~/.beads/shared-server/` listening on port 3308
- Repo's `.beads/config.yaml` has:
  ```yaml
  dolt:
    shared-server: true
  ```

PR #3550, opened 2026-04-27
fix(dolt): bd dolt status probes SQL when local server is externally-managed (be-0eyj)

quad341:
## Summary

`bd dolt status` previously printed **\"Dolt server: not running\"** when the rig pointed at a **local** Dolt sql-server whose lifecycle was owned by something other than bd (an orchestrator like gc, a systemd unit, etc.). Every other CRUD command (`bd list`, `bd ready`, …) connected to the same server fine — only `dolt status` lied. This caused agents in the field to halt thinking Dolt was down.

This PR widens the routing in `doltStatusCmd` so the SQL-probe path covers both flavors of externally-managed server:

- non-local host (already covered, GH#3362)
- **local host with bd auto-start disabled** — new (be-0eyj)

## Root cause

The current condition routes to `runExternalDoltStatus` only when the host is non-local:

\`\`\`go
if cfg.IsDoltServerMode() && !isLocalHost(cfg.GetDoltServerHost()) {
    runExternalDoltStatus(beadsDir, cfg)
    return
}
// ↓ falls through to PID-file path even when nothing wrote a PID file
serverDir := doltserver.ResolveServerDir(beadsDir)
state, err := doltserver.IsRunning(serverDir)
\`\`\`

A rig configured as \`dolt_mode: server\` + \`dolt.auto-start: false\` (with the orchestrator running the dolt sql-server itself) lands on \`localhost\`. \`isLocalHost\` returns true, so the dispatcher falls through to \`IsRunning\`, which reads a PID file that no one ever wrote and returns \`Running: false\`.

## Fix

Extract the routing into a small pure predicate:

\`\`\`go
func shouldUseExternalDoltStatus(cfg *configfile.Config, autoStartDisabled bool) bool {
    if cfg == nil || !cfg.IsDoltServerMode() {
        return false
    }
    if !isLocalHost(cfg.GetDoltServerHost()) {
        return true
    }
    return autoStartDisabled
}
\`\`\`

When auto-start is disabled, bd is by definition not the lifecycle owner — there is no PID file to consult, so SQL probe is the right answer regardless of host. The PID-file path is preserved unchanged for genuinely bd-managed servers (auto-start enabled, local host) so PID/log/data details continue to appear there.

## Verification

Against a live rig that hit the bug (gc-managed shared dolt on \`127.0.0.1:28231\`, \`BEADS_DOLT_AUTO_START=0\`):

**Before:**
\`\`\`
$ bd dolt status
Dolt server: not running
  Expected port: 28231
[...]

PR #4568, opened 2026-07-04
fix(dolt): probe SQL for bd dolt status in shared-server mode (GH#3218)

### LINKED ISSUE #3218: External Dolt server not recognized by `bd dolt status` / `bd dolt show` in server mode
## Problem

When `dolt_mode: server` is configured in `.beads/metadata.json` and the Dolt SQL server is managed externally (e.g. by a macOS launchctl agent), the following commands produce incorrect output:

| Command | Observed | Expected |
|---------|----------|----------|
| `bd dolt status` | `running: false, pid: 0, port: 0` | `running: true, managed: false, host: 127.0.0.1, port: 3307` |
| `bd dolt status --json` | `{"running":false,"pid":0,"port":0}` | JSON reflecting actual reachability |
| `bd dolt show` | `Mode: per-project` | `Mode: server` |

Meanwhile, `bd dolt test` correctly connects and reports the server as reachable.

## Root cause

`bd dolt status` checks for a Beads-managed Dolt process (PID file / process table). It does not probe the configured `dolt_server_host:dolt_server_port` for liveness when no managed process is found.

`bd dolt show` appears to derive mode from the presence of a local `.beads/dolt/<db>` directory rather than reading `dolt_mode` from metadata.

## Expected behavior

1. If metadata says `dolt_mode: server`, attempt a TCP connect or minimal SQL ping to `dolt_server_host:dolt_server_port`.
2. If reachable, report `running: true` with `managed: false` to distinguish external servers from Beads-managed ones.
3. `bd dolt show` should read `dolt_mode` from metadata as the authoritative mode value.

### Target output shape

```json
{
  "running": true,
  "managed": false,
  "host": "127.0.0.1",
  "port": 3307,
  "data_dir": "/path/to/shared/beads-dolt",
  "mode": "server"
}
```

## Environment

- beads v1.0.0 (Homebrew)
- macOS, Dolt SQL server managed by launchctl
- `.beads/metadata.json` with `dolt_mode: server`, `dolt_server_host: 127.0.0.1`, `dolt_server_port: 3307`

## Workaround

Use `bd dolt test` as the authoritative connectivity check and ignore `bd dolt status` output for external servers.

maxinflection:
## What

Route `bd dolt status` to the external SQL probe whenever shared-server mode is active, instead of consulting the local PID file. Fixes the case where a shared Dolt server on localhost is falsely reported as "not running."

## Why

Fixes #3218. With a shared Dolt server on localhost and bd auto-start **enabled**, `bd dolt status` printed `Dolt server: not running` even though `bd dolt test`, `bd dolt show`, and normal CRUD all connected fine (last reproduced 2026-06-14 on v1.0.5).

`shouldUseExternalDoltStatus` (`cmd/bd/dolt.go`) only probed SQL for non-local hosts or auto-start-disabled. A shared server on localhost with auto-start enabled matched neither, so status fell back to the PID-file lookup — but bd has no PID file for a server owned by Homebrew/systemd/launchd or a sibling clone. (PR #3550 fixed the non-local / auto-start-disabled variants; this closes the remaining shared-server-on-localhost case.)

There is also a subtler slice: `Config.IsDoltServerMode()` does not consult `dolt.shared-server` from `config.yaml` (only `dolt.mode` and the `BEADS_DOLT_SHARED_SERVER` env var). So `dolt.shared-server: true` in config.yaml + a stale `metadata.json` pinning `dolt_mode="embedded"` reports `IsSharedServerMode() == true` but `IsDoltServerMode() == false`, and still fell through.

The fix places the `sharedServerMode` branch **before** the `IsDoltServerMode` guard, so shared-server mode wins over stale embedded metadata — mirroring the existing `loadServerMode` override in `main.go` (#2946: `if !sm && !psm && IsSharedServerMode() { sm = true }`). Proxied-server mode is excluded (`!cfg.IsDoltProxiedServerMode()`) to match that override's `!psm` guard. The flag is passed in (like `autoStartDisabled`) to keep the predicate pure/testable; the caller supplies `doltserver.IsSharedServerMode()`. The doc comment and `status` help text are updated accordingly.

`bd dolt stop` has the same shared-server gap (#3687) but lives in `doltserver.StopWithForce` — left for a separate change.

## Verification

Extends `TestShouldUseExternalDoltStatus` with:
- shared-server + localhost + auto-start **enabled** → external (primary #3218 regression)
- shared-server + empty host → external
- shared-server flag + stale `embedded` metadata → external (the config.yaml slice; previously fell through)
- proxied-server + shared-server flag → non-external (precedence, matches main.go `!psm`)

```
go test -tags gms_pure_go ./cmd/bd/ -run 'ShouldUseExternalDoltStatus|RenderLocalDoltStatus|RunExternalDoltStatus' -v   # all pass
go test -tags gms_pure_go ./cmd/bd/ -run Dolt                                                                            # full Dolt suite green
go vet -tags gms_pure_go ./cmd/bd/ && gofmt -l cmd/bd/dolt.go cmd/bd/dolt_test.go                                        # clean
```

### DIFF
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -573,10 +573,10 @@ var doltStatusCmd = &cobra.Command{
 In embedded mode, reports that the Dolt engine runs in-process and shows
 the on-disk data directory. For beads-managed (local) servers, displays
 PID, port, and data directory from the local PID file. For externally-
-managed servers — either a remote dolt_server_host or a local server
-managed outside bd (dolt.auto-start: false, e.g. an orchestrator-shared
-sql-server) — pings the configured endpoint via SQL and reports
-reachability, server version, and database.`,
+managed servers — a shared server (dolt.shared-server: true), a remote
+dolt_server_host, or a local server managed outside bd (dolt.auto-start:
+false, e.g. an orchestrator-shared sql-server) — pings the configured
+endpoint via SQL and reports reachability, server version, and database.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		beadsDir := selectedDoltBeadsDir()
 		if beadsDir == "" {
@@ -608,7 +608,7 @@ reachability, server version, and database.`,
 			// — which is the exact failure mode this PR addresses.
 			fmt.Fprintf(os.Stderr, "Warning: cannot load .beads config (%v); falling back to PID-file status path\n", cfgErr)
 		}
-		if cfg != nil && shouldUseExternalDoltStatus(cfg, doltserver.IsAutoStartDisabled()) {
+		if cfg != nil && shouldUseExternalDoltStatus(cfg, doltserver.IsAutoStartDisabled(), doltserver.IsSharedServerMode()) {
 			runExternalDoltStatus(beadsDir, cfg)
 			return
 		}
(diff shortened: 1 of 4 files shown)


maphew (approved):
Reviewed the status-routing change and pushed a maintainer docs fix directly to this branch.

What I checked:

- The shared-server status predicate now matches the existing `loadServerModeFromBeadsDir` shared-server override and keeps proxied-server excluded.
- The stale `metadata.json` / `dolt.shared-server: true` path can reach the new predicate because no-store `bd dolt status` setup binds the selected config before the command runs.
- The failing doc freshness check was caused by the generated `bd dolt status` help text drift, so I regenerated `docs/CLI_REFERENCE.md` and `website/docs/cli-reference/dolt.md` in commit `742a7d29c`.

Local verification:

- `go test -tags gms_pure_go ./cmd/bd/ -run 'ShouldUseExternalDoltStatus|RenderLocalDoltStatus|RunExternalDoltStatus' -v`
- `gofmt -l cmd/bd/dolt.go cmd/bd/dolt_test.go`
- `go vet -tags gms_pure_go ./cmd/bd/`
- `CGO_ENABLED=0 go build -tags gms_pure_go -o ./bd-docs ./cmd/bd/ && ./scripts/check-doc-flags.sh ./bd-docs`

No remaining code-review findings from me. Approving; final merge should still wait for the refreshed CI run to complete.

PR #4569, opened 2026-07-04
feat(prime): cap injected memories with count/char budgets and a visible elision banner

Wldc4rd:
## Problem

`bd prime` injects every persistent memory, whole, in alphabetical key order, with no cap. SessionStart hook hosts cap what they ingest (Claude Code persists a tool result around ~25K tokens and previews far less), so a large memory set silently loses everything past the cut — and because the order is alphabetical, whether a memory survives is an accident of its key's first letter, unrelated to importance or freshness.

Real numbers from the workspace that hit this: 102 memories, ~160KB / ~40K tokens of `--memories-only` output. Roughly 80% of memories were never ingested on any session start. It surfaced as agents "forgetting" in-flight handoffs after every compaction — the fresh session declared all-quiet while two active threads sat stalled, their context sitting past the truncation line. The existing truncation directive ("read the full persisted hook output") mitigates but asks a fresh session to read 40K tokens to find the one thing blocked on it.

## Change

Two opt-in caps, resolved flag-over-config so a repo can adopt them without editing every hook line:

| Flag | Config key | Meaning |
|---|---|---|
| `--max-memories N` | `prime.max-memories` | at most N memories injected |
| `--max-memory-chars N` | `prime.max-memory-chars` | total injected entry characters ≤ N |

- Caps apply at **whole-memory boundaries**; at least one memory is always emitted (a single oversized memory exceeds the char budget rather than vanishing).
- When anything is elided, a **banner ahead of the entries** — where host truncation cannot hide it — reports how many were cut, which cap did it, and points at `bd memories <keyword>` / `bd remember <key>` for the rest. A capped prime never silently drops context.
- An explicit flag value of `0` forces unlimited over a configured cap.
- Defaults (no flag, no config) leave output **byte-identical** to today, verified against a 110-memory store in both `--memories-only` and full prime modes.
- The config keys are registered in `YamlOnlyKeys` (prime can run before the database is reachable, so they must be readable from `config.yaml`) and in the recognized-keys list, so `bd config set prime.max-memories 20` routes to yaml and reads back without warnings.
- Both caps apply in compact (MCP) mode too, with a matching one-line elision note.

`formatMemoriesForPrime` is split so the selection/formatting is a pure function (`renderPrimeMemories`) with direct unit tests; the store-fetch half is unchanged.

## Testing

- 7 new tests (`cmd/bd/prime_memory_caps_test.go`): uncapped legacy shape, count cap + banner-precedes-entries, char cap at whole-memory boundary, oversized-first-memory floor, compact capped/uncapped shapes, and the full flag/config/explicit-zero/negative-clamp resolution matrix.
- All `Prime*` tests pass; `internal/config` suite passes.
- Full `cmd/bd` suite: one failure, `TestIsBackupAutoEnabled`, which fails identically on a pristine `origin/main` checkout in the same environment (env-sensitive, unrelated).
- Behavioral verification on a seeded 110-memory (~158KB) store: byte-identical uncapped output vs a main-built binary; `--max-memories 15` → 22.2KB; `--max-memory-chars 25000` → 25.1KB; config-key path, flag-beats-config, and `--hook-json` envelope validity (banner arrives inside `additionalContext`).
- A `.beads/PRIME.md` override continues to win over everything, so repos curating their own resume-head are unaffected.

## Non-goals / follow-up

Recency- or pinned-first selection would be strictly better than alphabetical-capped, but the memory store keeps no per-key timestamps (memories are rows in the `config` table), so it needs sidecar metadata — left as follow-up so this stays small and backward-compatible.

While testing I also hit a pre-existing, unrelated bug: `bd config unset` of yaml-routed dotted keys (e.g. `backup.git-push`) reports success but leaves the nested yaml entry live — reproducible on main with keys that predate this PR. Filing separately.

### DIFF
--- a/cmd/bd/config.go
+++ b/cmd/bd/config.go
@@ -878,6 +878,7 @@ var recognizedConfigKeys = map[string]bool{
 	"create.require-description": true, "beads.role": true,
 	"auto_compact_enabled": true, "schema_version": true,
 	"output.title-length": true,
+	"prime.max-memories": true, "prime.max-memory-chars": true,
 }
 
 func isRecognizedConfigKey(key string) bool {
--- a/cmd/bd/config.go
+++ b/cmd/bd/config.go
@@ -878,7 +878,7 @@ var recognizedConfigKeys = map[string]bool{
 	"create.require-description": true, "beads.role": true,
 	"auto_compact_enabled": true, "schema_version": true,
 	"output.title-length": true,
-	"prime.max-memories": true, "prime.max-memory-chars": true,
+	"prime.max-memories":  true, "prime.max-memory-chars": true,
 }
 
 func isRecognizedConfigKey(key string) bool {
--- a/cmd/bd/prime.go
+++ b/cmd/bd/prime.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"sort"
 	"strings"
 	"time"
 
@@ -111,7 +110,9 @@ Memory injection caps:
 	prime.max-memory-chars config keys; an explicit flag wins, and an explicit
 	0 forces unlimited). Caps apply at whole-memory boundaries, at least one
 	memory is always emitted, and a banner ahead of the entries reports how
-	many were elided and how to browse the rest with bd memories.`,
+	many were elided and how to browse the rest with bd memories.
+	--max-memory-chars caps the total bytes of the injected memory entries;
+	the section header and elision banner are excluded from the budget.`,
 	SilenceUsage:  true,
 	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
@@ -202,7 +203,7 @@ func init() {
 	primeCmd.Flags().BoolVar(&primeMemoriesOnly, "memories-only", false, "Output only persistent memories for compact hook contexts")
 	primeCmd.Flags().BoolVar(&primeHookJSONMode, "hook-json", false, "Wrap output in the SessionStart hook JSON envelope (Claude Code, Gemini CLI, Codex)")
 	primeCmd.Flags().IntVar(&primeMaxMemories, "max-memories", 0, "Cap injected persistent memories to N entries (0 = unlimited; falls back to the prime.max-memories config key)")
-	primeCmd.Flags().IntVar(&primeMaxMemoryChars, "max-memory-chars", 0, "Cap the total characters of injected persistent memories, at whole-memory boundaries (0 = unlimited; falls back to the prime.max-memory-chars config key)")
+	primeCmd.Flags().IntVar(&primeMaxMemoryChars, "max-memory-chars", 0, "Cap the total bytes of injected memory entries, at whole-memory boundaries; section header and banner are not counted (0 = unlimited; falls back to the prime.max-memory-chars config key)")
 	rootCmd.AddCommand(primeCmd)
 }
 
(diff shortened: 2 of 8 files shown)


maphew:
You're pointing out the dumb method we're using for memory management is significant - thanks! 

What work do you have pending on the PR? (What is keeping it in draft)

Wldc4rd:
> You're pointing out the dumb method we're using for memory management is significant - thanks!
> 
> What work do you have pending on the PR? (What is keeping it in draft)

I think the only "wierdness here" is that I'm using doltlite and that may be a factor (I'm not sure but Dolt might have timestamping) but if this same behavior also occurs in Dolt then it's a pretty big deal. Beads memories with no timestamps are dangerous. The fix is probably good to do anyway though so I'll kick it out of draft. Thanks!

maphew (commented):
Thanks for this — the problem statement is exactly right (silent host truncation of alphabetical-ordered memories is a real failure mode we've hit here too), the whole-memory-boundary capping with a banner *ahead* of the entries is the right shape, and splitting `renderPrimeMemories` out as a pure function with direct tests made this easy to review. Uncapped byte-identical output was verified on our side as well.

A multi-agent review (8 finder angles, adversarially verified) surfaced the items below. Since maintainer edits are enabled on this branch, **I'll push fix commits for these directly to this PR** rather than asking for another round — you'll see them land shortly; flag anything you disagree with.

## Blocking (CI is currently red on these)

1. **`docs/CLI_REFERENCE.md` not regenerated.** The PR adds two flags and edits `primeCmd`'s Long help, so the `check-doc-flags` job fails on docs drift. Will regenerate with the pinned pure-Go build.
2. **`gofmt` failure in `cmd/bd/config.go`** (the added `recognizedConfigKeys` line breaks alignment). `make fmt` fixes it.

## Correctness

3. **No value validation for the new keys** (`internal/config/yaml_config.go` — `validateYamlConfigValue` has cases for peers like `hierarchy.max-depth` but none for `prime.*`). `bd config set prime.max-memory-chars 25k` reports success, viper's `GetInt` coerces `"25k"` to 0, and 0 means uncapped — so a typo silently disables the cap the user believes is active, which is the exact failure this feature exists to prevent. Fix: non-negative-integer validation cases for both keys.
4. **The elision banner names every configured cap, not the one that fired.** With `max-memories=50`, `max-memory-chars=500`, and 10 stored memories, the char cap binds but the banner reads `showing 3 of 10 ... capped by max-memories=50, max-memory-chars=500` — self-contradictory, and raising `max-memories` visibly does nothing. The fill loop already knows which `break` it took; the banner should report only the binding cap.

## Documentation

5. **`docs/CONFIG.md`** — the hand-maintained key table documents direct peers (`create.require-description`, `output.title-length`) but not the two new keys; nothing auto-generates that file, so they're invisible to anyone auditing config. Will add entries including the yaml-only rationale and 0-means-unlimited semantics.
6. **`--max-memory-chars` counts bytes, not characters** (`len(entry)` on a Go string), and the budget covers entry strings only — the section header, the `Stored via ...` line, and the banner itself (~380 chars in full mode) are uncounted. Both are defensible (error direction is conservative; your PR description already scopes the cap to entry characters), but the flag help reads like an exact total-character cap. Will clarify the help text rather than change behavior.

## Non-blocking notes

- The flag-explicitness state (`primeMaxMemoriesSet` / `primeMaxMemoryCharsSet`) lives in package globals written only inside the cobra `RunE`. Every current caller routes through `RunE` so this is correct today, but a future in-process caller (e.g. MCP serving prime context directly) would silently lose the "explicit 0 = force unlimited" semantics. Not fixing in this pass; noting it for a follow-up.
- `cmd/bd` already has `sortedKeys[V any]` (`delete_proxied_server.go`) and `truncate` (`worktree_cmd.go`) that the new/moved code re-implements inline — cosmetic, may fold into the fix commits if they stay small.
- On your comment about missing timestamps: agreed that's the deeper issue (alphabetical-capped means late-alphabet keys are elided forever), and your call to keep recency selection out of this PR is right. Worth its own issue so it doesn't get lost.

Findings were adversarially verified against the diff and a main checkout before posting; a suspected test-ordering flake was checked and refuted, so no concern there.

maphew:
Maintainer fix commits pushed to this branch, per the review above — your original commit is untouched, these sit on top:

- `85cc4f0` — the correctness/cleanup items: `validateYamlConfigValue` cases for both keys (a typo'd value like `25k` now fails at `bd config set` time instead of silently meaning unlimited), the elision banner now names only the cap that actually fired (new test covers the both-configured/one-binding case), help text and `docs/CONFIG.md` clarified to "bytes of memory entries, header/banner excluded", plus the gofmt fix and two small helper reuses (`sortedKeys`, `truncate`) that don't change behavior.
- `5db07ed` — regenerated `docs/CLI_REFERENCE.md` and the website prime page for the two new flags, which is what `check-doc-flags` was failing on.

Verified locally: `gofmt -l` clean, `go test ./cmd/bd -run 'Prime|RenderPrimeMemories'` and the full `internal/config` suite pass (pure-Go build), and `scripts/check-cli-docs-drift.sh` passes. All six red checks traced back to the docs drift and the gofmt slip, so CI should go green on this push.

Nothing about your cap semantics changed: whole-memory boundaries, the at-least-one-memory floor, explicit-0-forces-unlimited, and byte-identical uncapped output are all as you wrote them. If any of the wording tweaks feel off, push over them — happy to rebase around you.

The no-timestamps-on-memories issue you raised in the thread deserves its own tracking issue rather than living only here; we'll file it upstream so the recency-selection follow-up doesn't get lost.

PR #4575, opened 2026-07-05
fix(resolve): map embeddeddolt data dir back to its beads dir (GH#4574)

### LINKED ISSUE #4574: BEADS_DIR embedded store not named .beads silently resolves to empty ~/.beads (returns 0 issues) in 1.1.0
**Summary:** In 1.1.0, an embedded Dolt store pointed to via `BEADS_DIR` whose directory leaf is not literally `.beads` is silently ignored — data commands resolve to the auto-created empty `~/.beads` and return 0 issues. It looks like total data loss; the data is actually intact.

## Repro (bd 1.1.0, Homebrew)

```sh
# 1) seed a store OUTSIDE $HOME
BEADS_DIR=/tmp/bd-seed bd init --prefix seed
BEADS_DIR=/tmp/bd-seed bd create "hello world" -t task
BEADS_DIR=/tmp/bd-seed bd count      # => 1  (correct)

# 2) same store copied UNDER $HOME with a non-".beads" leaf
cp -R /tmp/bd-seed ~/bd-under-home
BEADS_DIR=~/bd-under-home bd count    # => 0  (BUG)
# also prints: warning: no beads configuration found in ~/.beads; using default database name "beads"

# 3) same store, leaf renamed to ".beads"
BEADS_DIR=~/somedir/.beads bd count   # => 1  (works)
```

Identical data; the only variables are the directory *name* and whether an ancestor `.beads` (the auto-created `~/.beads`) exists.

## Root cause (main)

- `Config.DatabasePath()` always returns `<beadsDir>/dolt`, but the 1.0+ embedded engine stores data at `<beadsDir>/embeddeddolt/` (`internal/storage/embeddeddolt/store.go`; `dolt/` is the pre-0.56 layout).
- In `resolveBeadsDirForDBPath` (`cmd/bd/main.go`), the candidate check `PathsEqual(cfg.DatabasePath(beadsDir), dbPath)` never matches (`…/dolt` != `…/embeddeddolt`), so it returns "".
- `resolveCommandBeadsDir` then falls back to a parent-directory walk that grabs the first ancestor `.beads` — the auto-created empty `~/.beads`.
- A store whose dir *is* named `.beads` works only because of the basename early-return in `resolveBeadsDirForDBPath`, which bypasses the broken `DatabasePath` compare.

## Impact

Any custom `BEADS_DIR` embedded store not named `.beads` (common when a tool embeds bd and keeps its store under e.g. `~/.local/state/<app>/beads`) silently returns zero results after upgrading to 1.1.0. It's worse under `$HOME`, where `~/.beads` is always an ancestor and gets auto-created.

## Suggested fix

`DatabasePath()` should return the `embeddeddolt/` path for the embedded backend (or the resolver should compare against the actual embedded data dir), so `resolveBeadsDirForDBPath` matches regardless of the store's directory name.

## Environment

- `bd version 1.1.0 (Homebrew)`, macOS (arm64)

enieuwy:
Fixes #4574.

## Problem

An embedded Dolt store pointed to via `BEADS_DIR` whose directory leaf is not literally `.beads` is silently ignored in 1.1.0 — data commands resolve to the auto-created empty `~/.beads` and return 0 issues (looks like data loss; data is intact).

Root cause: in `resolveBeadsDirForDBPath` (`cmd/bd/store_reopen.go`) a candidate beads dir was accepted only when `cfg.DatabasePath(beadsDir) == dbPath`. `Config.DatabasePath` returns `<beadsDir>/dolt`, but the 1.0+ embedded engine stores data at `<beadsDir>/embeddeddolt/`, so the compare never matched. `resolveCommandBeadsDir` then fell back to its ancestor-walk and returned the first `.beads` it found — the auto-created empty `~/.beads`. A store whose dir *is* named `.beads` only works because of the basename early-return that bypasses the broken compare.

## Fix

Match on the parent relationship: `dbPath` is always the data directory living directly inside `beadsDir` (`<beadsDir>/embeddeddolt` embedded, `<beadsDir>/dolt` server), so accept a candidate when `filepath.Dir(dbPath) == beadsDir`. This is guarded by `configfile.Load(beadsDir)` succeeding, so it can't false-match an unrelated parent (the empty `~/.beads` has no `metadata.json` and is skipped). `Config.DatabasePath`'s broad contract is left untouched.

## Test

`TestResolveBeadsDirForDBPath_EmbeddedNonDotBeads` builds an embedded store whose beads dir is *not* named `.beads` and asserts `resolveBeadsDirForDBPath(<dir>/embeddeddolt)` returns `<dir>`. It fails on `main` (returns `""`) and passes with this change.

## Repro (before fix)

```sh
BEADS_DIR=/tmp/bd-seed bd init --prefix seed
BEADS_DIR=/tmp/bd-seed bd create "hello world" -t task
BEADS_DIR=/tmp/bd-seed bd count      # => 1
cp -R /tmp/bd-seed ~/bd-under-home
BEADS_DIR=~/bd-under-home bd count    # => 0  (BUG); fixed => 1
```

### DIFF
--- a/cmd/bd/store_reopen.go
+++ b/cmd/bd/store_reopen.go
@@ -96,6 +96,14 @@ func resolveBeadsDirForDBPath(dbPath string) string {
 		if utils.PathsEqual(beadsDir, dbPath) || utils.PathsEqual(beadsDir, actualDBPath) {
 			return beadsDir
 		}
+		// dbPath is the data directory living directly inside beadsDir
+		// (<beadsDir>/embeddeddolt for embedded, <beadsDir>/dolt for server).
+		// Match on that parent relationship so resolution works for embedded
+		// stores regardless of the beads dir name; Config.DatabasePath alone
+		// returns the "dolt" name and misses embeddeddolt/ (GH#4574).
+		if utils.PathsEqual(filepath.Dir(dbPath), beadsDir) || utils.PathsEqual(filepath.Dir(actualDBPath), beadsDir) {
+			return beadsDir
+		}
 		if utils.PathsEqual(cfg.DatabasePath(beadsDir), dbPath) || utils.PathsEqual(cfg.DatabasePath(beadsDir), actualDBPath) {
 			return beadsDir
 		}
--- a/cmd/bd/store_reopen_embedded_resolution_test.go
+++ b/cmd/bd/store_reopen_embedded_resolution_test.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/configfile"
+	"github.com/steveyegge/beads/internal/utils"
+)
+
+// TestResolveBeadsDirForDBPath_EmbeddedNonDotBeads guards GH#4574: an embedded
+// store stores its data under <beadsDir>/embeddeddolt/, but Config.DatabasePath
+// returns <beadsDir>/dolt. When the beads dir is not literally named ".beads"
+// the early-return in resolveBeadsDirForDBPath does not fire, so resolution
+// must still map <beadsDir>/embeddeddolt back to <beadsDir> via the parent
+// relationship — otherwise commands silently fall back to an ancestor ~/.beads
+// and report zero issues.
+func TestResolveBeadsDirForDBPath_EmbeddedNonDotBeads(t *testing.T) {
+	root := t.TempDir()
+	// leaf is intentionally NOT ".beads" (mirrors an embedding tool that keeps
+	// its store at e.g. ~/.local/state/<app>/beads).
+	beadsDir := filepath.Join(root, "beads")
+	dbPath := filepath.Join(beadsDir, "embeddeddolt")
+	if err := os.MkdirAll(dbPath, 0o755); err != nil {
+		t.Fatalf("mkdir embeddeddolt dir: %v", err)
+	}
+
+	cfg := &configfile.Config{
+		Database:     "dolt",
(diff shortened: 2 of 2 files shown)


maphew (commented):
Thanks for the clear repro and root-cause writeup, @enieuwy - this is exactly the kind of "looks like data loss but isn't" bug that's worth fixing fast.

I reproduced the issue independently:
- Checked out this PR's commit, reverted just `store_reopen.go` back to `main` while keeping your new test: `TestResolveBeadsDirForDBPath_EmbeddedNonDotBeads` fails on `main` (`resolveBeadsDirForDBPath` returns `""`), and passes with your fix applied.
- Ran the full `cmd/bd` test suite plus `doctor`/`doctor/fix` packages after the fix: all green, no regressions.
- Confirmed the new parent-match check (`filepath.Dir(dbPath) == beadsDir`) can only match the true owning directory of `dbPath`, since it's gated behind `configfile.Load(beadsDir)` succeeding and directory dedup in the candidate list - it can't accidentally grab an unrelated ancestor `.beads` (like the auto-created empty `~/.beads` from the bug report).
- `Config.DatabasePath`'s existing contract (`<beadsDir>/dolt`, or the custom `dolt_data_dir` override) is untouched, so server-mode and custom-data-dir resolution paths are unaffected.

This is a self-contained fix to `bd`'s own path-resolution logic - no storage-engine internals touched - and stays well within scope for beads core.

Only note: CI hasn't run yet on the fork branch (`gh pr checks` reports nothing), but local build and full test run confirm the fix cleanly with no side effects.

Recommend merging as-is once CI comes back green.

PR #4576, opened 2026-07-05
fix(doltremote): fold host case, strip creds, recognize user-less SCP in CanonicalForComparison

maphew:
## Why

Follow-up to #4153 (git-origin collision guard in `bd dolt remote add` + `bd doctor`). A re-review of that PR found the shared URL comparator `doltremote.CanonicalForComparison` misses three equivalences, all **false negatives** that silently weaken the collision guard and the `bd doctor` remotes check (flagged in https://github.com/gastownhall/beads/pull/4153#issuecomment-4883818549):

1. **Host case not folded** - `https://GitHub.com/org/repo` did not match `https://github.com/org/repo`, despite DNS being case-insensitive.
2. **Embedded credentials not stripped** - a PAT-in-URL git origin (`https://user:pass@github.com/org/repo`) never matched a clean Dolt remote for the same repo.
3. **User-less SCP form not recognized** - `github.com:org/repo.git` was not treated as SCP-style, so it never canonicalized to the same form as `git@github.com:org/repo.git`.

The direction is fail-safe (the guard fails to fire rather than wrongly blocking), so this was non-blocking for #4153 landing - but each gap lets a genuine same-repo collision slip past the guard.

## What

In `internal/doltremote/remote.go`:

- `CanonicalForComparison` now, after the existing trailing-slash / `.git` trimming, strips any embedded `user[:pass]@` credentials from the authority and lowercases **only** the host (via new helper `stripCredentialsAndFoldHostCase`). Scheme and path are left untouched.
- `isSCPStyleGitURL` now recognizes the user-less SCP form `host:path` when the pre-colon token looks like a hostname (contains a `.`), in addition to the existing `user@host:path` form. Dotless tokens (e.g. a Windows drive letter `C:foo`) are still rejected, and `isWindowsDrivePath` continues to run first in `Normalize`.

Deliberately preserved:
- **`http` vs `https` stay distinct** - the scheme is never folded (different endpoints).
- The org/repo **path case is never folded** (only the host).

## Tests

Extends the existing `TestCanonicalForComparison_*` unit tests in `cmd/bd/dolt_remote_origin_guard_test.go` with the three equivalence classes (host case, credentials incl. token-only form, user-less SCP three-way equivalence) plus a negative guard asserting `http://` != `https://`.

```
CGO_ENABLED=0 go test -tags gms_pure_go -run 'TestCanonicalForComparison|TestDoltRemoteMatchesGitOrigin' ./cmd/bd/   # PASS
CGO_ENABLED=0 go build -tags gms_pure_go ./...                                                                        # clean
gofmt -l ...                                                                                                          # clean
```

### DIFF
--- a/cmd/bd/dolt_remote_origin_guard_test.go
+++ b/cmd/bd/dolt_remote_origin_guard_test.go
@@ -134,6 +134,36 @@ func TestCanonicalForComparison_UserlessSCP(t *testing.T) {
 	}
 }
 
+func TestCanonicalForComparison_NonDefaultSSHUserPreserved(t *testing.T) {
+	// Non-default SSH users select the remote account/home directory and
+	// must NOT be stripped like HTTP(S) transport credentials - otherwise
+	// distinct remotes for different accounts on the same host would
+	// falsely canonicalize to the same string.
+	alice := doltremote.CanonicalForComparison("alice@example.com:repo.git")
+	bob := doltremote.CanonicalForComparison("bob@example.com:repo.git")
+	if alice == bob {
+		t.Errorf("distinct SSH users must not canonicalize equal, both got %q", alice)
+	}
+
+	aliceGitSSH := doltremote.CanonicalForComparison("git+ssh://alice@example.com/repo.git")
+	if alice != aliceGitSSH {
+		t.Errorf("SCP and git+ssh forms of the same non-default user must match: %q vs %q", alice, aliceGitSSH)
+	}
+	if bob == aliceGitSSH {
+		t.Errorf("git+ssh://alice@... must not match a bob@... origin: %q vs %q", aliceGitSSH, bob)
+	}
+}
+
+func TestCanonicalForComparison_DefaultGitSSHUserStillFolded(t *testing.T) {
+	// The conventional "git" SSH user remains the documented default-equivalent
+	// case: bare host:path and explicit git@host:path must still match.
+	userless := doltremote.CanonicalForComparison("github.com:org/repo.git")
+	withGitUser := doltremote.CanonicalForComparison("git@github.com:org/repo.git")
+	if userless != withGitUser {
+		t.Errorf("default git@ user must still fold with userless form: %q vs %q", withGitUser, userless)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


maphew:


**[P2] Preserve non-default SSH users in comparisons — internal/doltremote/remote.go:129-130** — _confirmed_

When both remotes are SSH/SCP and the username selects the account or home directory, for example git origin `alice@example.com:repo.git` and Dolt remote `bob@example.com:repo.git`, this strips `alice@`/`bob@` from the authority and makes `CanonicalForComparison` return the same string. That causes `doltRemoteMatchesGitOrigin` and the doctor check to reject or warn on distinct remotes; limit credential stripping to HTTP(S) userinfo or preserve SSH users except explicitly default-equivalent cases.

**[P2] Normalize dotless SCP host aliases — internal/doltremote/remote.go:78-78** — _softened_

When a git origin uses a userless SSH config alias such as `github:org/repo.git` or `localhost:repo.git`, the new userless-SCP path still returns false because the host token lacks a dot. The origin remains raw while equivalent Dolt forms like `git+ssh://github/org/repo.git` canonicalize to `git+ssh://...`, so the collision guard and doctor check miss the match; Git's `[user@]host:path` syntax allows dotless host aliases, so only Windows drive-like cases should be excluded.

Verification note: Known, documented limitation rather than an unnoticed bug: the PR's userless-SCP recognition deliberately requires a dotted host, so dotless ssh-config aliases (github:org/repo.git, localhost:repo.git) are still not canonicalized and a cross-form collision with an equivalent git+ssh:// or user@alias Dolt remote is silently missed (identical-string pairs still match). This weakens the fail-open guard for a real usage pattern; a follow-up could adopt git's own rule - treat any host:path as SCP-style except a single-letter pre-colon token (Windows drive), since isWindowsDrivePath alone does not cover slashless forms like C:foo. Suggest P3/follow-up rather than P2.

maphew:
Pushed a hardened canonical comparison fix in commit 5b1327e4a, addressing the 2026-07-07 cross-vendor review finding on this PR. The change tightens equality handling in the canonical comparison path so edge cases flagged during review no longer produce inconsistent results. Local build and targeted tests pass against the updated branch.

PR #4581, opened 2026-07-05
fix(dolt): borrow ignored-tx from pool + per-dial credential connector (hosted-gateway churn)

### LINKED ISSUE #1: `bd dep tree` mostly broken
A bit difficult to pinpoint, but I've had the following symptoms:

1. An epic contains item-1 and item-2, both have dependant issues, but only item-2 is expanded, item-1 does not have any subitems
2. Subitems tend to repeat multiple times in the same list level
3. Subitems with more than one blocker appear multiple times on multiple levels (might be correct, because it's a graph)

In general, a tree graph might not be a perfect idea if the database is a graph, but I think you know that way better than me.

### LINKED ISSUE #2: RFC: Optional Turso backend for branching, real-time sync, and vector search

julianknutsen:
Kills the ~10–20s/op connection churn against the hosted beads-gateway.

**Measured: ~37 ms/write saved (2.0×) at an 18 ms RTT** (latency-injected local bench; a lower bound — prod TLS makes each fresh dial costlier).

- **FIX #1** — `beginIgnoredTxOnBranch` borrows a warm connection from the main pool instead of a fresh single-conn `sql.DB` per write (MaxOpenConns==1 deadlock guard, InUse exhaustion pre-check, 250 ms bounded borrow; falls back to today's fresh-dial path when unsafe).
- **FIX #2** — `openSQLDB` builds a go-sql-driver `BeforeConnect` connector that re-resolves the cached credential token per new physical dial (hosted credential-command path only). Pooled connections survive EIA rotation. Local/static path byte-identical.
- **Follow-ups** — `resolveCredentialToken` honors the dial context; a MySQL 1045 invalidates the cached token so the retry re-mints.

Unit tests mutation-verified (deleting the BeforeConnect attachment / dropping the fallback credCmd fails the teeth tests); `go vet` clean; dolt-package tests pass in isolation (monolithic-run failures are docker-contention schema-init timeouts).

### DIFF
--- a/internal/configfile/central_config_test.go
+++ b/internal/configfile/central_config_test.go
@@ -209,6 +209,57 @@ func TestApplyCentralDefaults_NilCentralIsNoOp(t *testing.T) {
 	}
 }
 
+func TestTrustedDoltCredentialCommand(t *testing.T) {
+	const helper = "gasworks getToken beads --org o_1"
+
+	t.Run("env var is the highest-priority trusted source", func(t *testing.T) {
+		t.Setenv("BEADS_DOLT_CREDENTIAL_COMMAND", helper)
+		// A central config naming a different helper must still lose to the env var.
+		central := filepath.Join(t.TempDir(), "server.json")
+		writeCentralConfig(t, central, &Config{DoltCredentialCommand: "central-helper"})
+		t.Setenv("BEADS_CENTRAL_CONFIG", central)
+
+		if got := TrustedDoltCredentialCommand(); got != helper {
+			t.Fatalf("TrustedDoltCredentialCommand() = %q, want env value %q", got, helper)
+		}
+	})
+
+	t.Run("central server config is honored when no env var", func(t *testing.T) {
+		t.Setenv("BEADS_DOLT_CREDENTIAL_COMMAND", "")
+		central := filepath.Join(t.TempDir(), "server.json")
+		writeCentralConfig(t, central, &Config{DoltCredentialCommand: helper})
+		t.Setenv("BEADS_CENTRAL_CONFIG", central)
+
+		if got := TrustedDoltCredentialCommand(); got != helper {
+			t.Fatalf("TrustedDoltCredentialCommand() = %q, want central value %q", got, helper)
+		}
+	})
+
+	t.Run("no trusted source yields empty; project metadata is never a source", func(t *testing.T) {
+		t.Setenv("BEADS_DOLT_CREDENTIAL_COMMAND", "")
+		// Absent central config, and there is deliberately no parameter through which a
+		// project's tracked metadata.json could ever supply the helper.
+		t.Setenv("BEADS_CENTRAL_CONFIG", filepath.Join(t.TempDir(), "absent.json"))
+
+		if got := TrustedDoltCredentialCommand(); got != "" {
+			t.Fatalf("TrustedDoltCredentialCommand() = %q, want empty (no trusted helper)", got)
+		}
+	})
+
+	t.Run("broken central config resolves to no helper", func(t *testing.T) {
+		t.Setenv("BEADS_DOLT_CREDENTIAL_COMMAND", "")
+		central := filepath.Join(t.TempDir(), "server.json")
+		if err := os.WriteFile(central, []byte("{bad json"), 0o600); err != nil {
+			t.Fatal(err)
+		}
+		t.Setenv("BEADS_CENTRAL_CONFIG", central)
+
+		if got := TrustedDoltCredentialCommand(); got != "" {
+			t.Fatalf("TrustedDoltCredentialCommand() = %q, want empty (broken central is opt-out)", got)
+		}
+	})
+}
+
 func TestDefaultCentralConfigPath(t *testing.T) {
 	path := DefaultCentralConfigPath()
 	if path == "" {
--- a/internal/configfile/configfile.go
+++ b/internal/configfile/configfile.go
@@ -344,14 +344,40 @@ func (c *Config) GetDoltServerUser() string {
 	return DefaultDoltServerUser
 }
 
-// GetDoltCredentialCommand returns the credential-helper command bd runs to obtain a
-// short-lived server-mode credential. Checks BEADS_DOLT_CREDENTIAL_COMMAND env first, then
-// config. Empty means no helper — the static BEADS_DOLT_SERVER_USER / credentials-file path applies.
-func (c *Config) GetDoltCredentialCommand() string {
+// TrustedDoltCredentialCommand returns the server-mode credential-helper command bd runs to
+// obtain a short-lived credential (presented as the MySQL username), resolved from a TRUSTED,
+// user-local source only, in priority order:
+//
+//  1. the BEADS_DOLT_CREDENTIAL_COMMAND environment variable, then
+//  2. the central per-user server config (BEADS_CENTRAL_CONFIG or ~/.config/beads/server.json).
+//
+// SECURITY: it deliberately IGNORES the dolt_credential_command field of a project's tracked
+// .beads/metadata.json. That file is committed and travels with a git clone, so honoring an
+// executable helper named there would let any checked-out repository run arbitrary shell code
+// through the credential runner (sh -c) during ordinary store open — a reachable CWE-78 OS
+// command injection. Project metadata may point bd at connection endpoints, never at an
+// executable to run; a helper must be established by the machine's operator out-of-band.
+//
+// Empty means no trusted helper is configured — the static BEADS_DOLT_SERVER_USER /
+// credentials-file path applies.
+func TrustedDoltCredentialCommand() string {
 	if v := os.Getenv("BEADS_DOLT_CREDENTIAL_COMMAND"); v != "" {
 		return v
 	}
-	return c.DoltCredentialCommand
+	centralPath := os.Getenv("BEADS_CENTRAL_CONFIG")
+	if centralPath == "" {
+		centralPath = DefaultCentralConfigPath()
+	}
+	if centralPath != "" {
+		// A helper is strictly opt-in, so a missing or broken central config resolves to
+		// "no trusted helper" rather than surfacing an error here.
+		if central, err := LoadCentralConfig(centralPath); err == nil && central != nil {
+			if central.DoltCredentialCommand != "" {
+				return central.DoltCredentialCommand
+			}
+		}
+	}
+	return ""
 }
 
 // GetDoltDatabase returns the Dolt SQL database name.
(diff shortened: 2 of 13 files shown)


maphew:
Thanks Julian - this is a sharp fix for a real cost, and the workmanship shows. The value here is concrete: the ignored-tx path was paying a full fresh dial (TLS + MySQL auth + Dolt session setup) on every mixed write, and borrowing from the warm pool removes that entirely in the common case, with the old fresh-dial path preserved as a fallback so nothing regresses when borrowing is unsafe. The BeforeConnect connector is the right tool for per-dial credential refresh - it is the driver's official extension point, keeps the token out of the DSN's hot path, and the fail-closed behavior on mint errors is exactly right. The test suite is genuinely good: the mock-driver teeth tests document what mutation each one catches, and the error paths assert the borrowed conn goes back to the pool (InUse==0).

Findings, most important first:

1. **Borrowed conn returns to the pool with its session branch switched** (`transaction.go`, borrow fast path + cleanup closure). `DOLT_CHECKOUT` is session-level, and the cleanup returns the borrowed conn without restoring its prior branch - every other production checkout site (federation.go staging, compact, flatten) restores before release, and before this PR no ordinary write flow mutated a pooled session's branch. I'll be straight that this is *latent* today rather than reachable: `DoltStore.Checkout` has no non-test caller on main, so pool sessions can't actually diverge in any shipped flow - the exposure is multi-conn-pool tests and future callers of `Checkout`. So: normal severity, not a blocker, but worth the cheap hardening as defense in depth - only borrow when `currentBranch == s.branch` and fall back otherwise (one field compare, no extra round trip), or restore the branch in cleanup. A comment acknowledging the invariant would also help the interface-split work later.

2. **Stacking/sequencing**: the base `feat/dolt-credential-command` has no PR of its own and isn't on main, so no CI ran here and this can't land as filed. Suggest opening/landing the base PR first, or folding the stack into one PR retargeted at main. Good news on that path: I checked a 3-way merge of the PR-head `store.go` onto current main and it auto-merges cleanly - both #4481's `connReady` cleanup and your `openSQLDB` changes survive - so the retarget is mechanical, no real conflict with #4481.

3. **`isAuthError` string fallback** (`store.go`): the `"access denied"` substring match can misclassify a non-auth error that happens to contain the phrase, turning a fail-fast permanent error into invalidate+retry on the credential path. Bounded by backoff, but tightening to the typed 1045 check where possible would be safer.

4. Nit: the side pools (`execWithLongTimeout*`, `openMigrationDB`, `openLongTimeoutConn`) now resolve the cached token via BeforeConnect but have no 1045 invalidation, so a revoked-mid-life token fails there without re-mint until a `withRetry` op heals the cache. A shared comment or helper at those dial sites would close the loop.

Also worth cross-referencing #4102 in the description - this addresses the per-write churn half of that report (not the per-process cold start), and the linkage will help whoever picks up the rest.

Recommended outcome: **fix-merge** - address (1), resolve the stacking so CI runs, then this is a clear win. Fit-wise this is squarely adapter-internal connection lifecycle, the right side of the storage boundary, and it composes cleanly with the interface-split direction.

maphew:
Update: finding 1 from the review is now closed on this branch - I pushed 9c769896b directly (the branch lives in this repo).

Rather than restoring the branch in cleanup or gating on a store-field compare, the borrow fast path now verifies the borrowed session is already on the requested branch (`SELECT active_branch()` - same round-trip count as the `DOLT_CHECKOUT` it replaces, so the fast path stays free) and refuses otherwise, falling back to the existing fresh dial. A borrowed pool session's branch is never mutated, which keeps the same invariant every other production checkout site preserves by restore-before-release. New teeth test `TestIgnoredTxBorrowRefusesForeignBranch` proves the refusal, the untouched pool return, and that the write still completes via the fallback; the reuse test now asserts zero `DOLT_CHECKOUT` on the pool. Full pure-Go dolt package suite passes locally; build/vet/gofmt clean.

Findings 3 (typed-1045-only auth classification) and 4 (side-pool token invalidation) remain open as normal-severity follow-ups - happy to fold them here if you want them in this PR, otherwise they can trail.

What still blocks the merge is only the stacking, and that call is yours, @julianknutsen:

1. **Land the base first**: open a PR for `feat/dolt-credential-command` -> `main`, land it, and this PR then re-bases to a clean main-targeted diff with CI.
2. **Fold the stack**: retarget this PR to `main` so it carries both commits as one reviewed unit. I re-confirmed the review's finding that the head 3-way-merges onto current main cleanly (no #4481 conflict).

Say the word (or just do either - you have the keys) and I'll take it the rest of the way, same day. Everything else from the review is resolved or tracked.

maphew:
Following through on the July 7 offer to take this the rest of the way: I attempted the fold-the-stack path (option 2) today, and it surfaced something you should know about before either of us goes further — so this is a status-and-decision comment rather than a merge notification.

**The fold itself went fine.** Retargeting onto current `main` produced conflicts in four files (`configfile.go`, `open.go`, `open_test.go`, `store.go`) from the backend-simplification refactor; all resolved, build and vet clean, and FIX 1 (the ignored-tx pool borrow) came through untouched — nothing main landed overlaps it.

**What the tests then exposed is a genuine feature collision, not a merge artifact.** While this PR sat, main independently grew a second credential-command subsystem: `internal/creds` + `ApplyGatewayCredential` (landed with the backend work), which mints the token once at open and presents it as the connection username. Your FIX 2 is the per-dial `BeforeConnect` re-mint. The two compile side-by-side but interact badly, and one failing test in each direction proves it:

1. Main's gateway path runs first and never sets `cfg.CredentialCommand` — so your per-dial connector would silently never activate in exactly the hosted-gateway scenario this PR targets. Your own `TestApplyResolvedConfig_SetsCredentialCommand` catches this.
2. Your `resolveServerCredential` predates main's now-real embedded/server mode distinction and runs the credential command unconditionally — main's regression test (`embedded mode: command is not run even when set`) rightly fails.

Worth saying plainly: **the problem FIX 2 solves is still real on main.** Mint-once-at-open means pooled reconnects after token rotation dial with a stale username; your per-dial re-mint is the correct shape for that. The work isn't wasted — it needs to be unified with `internal/creds` (single resolver/runner, gateway path populating `CredentialCommand`, mode-gating) rather than merged alongside it. There's also a small trust-boundary divergence to settle: your `TrustedDoltCredentialCommand` honors central config as a command source; main's is env-only.

Two ways forward, your pick:

1. **Split**: land FIX 1 (pool borrow + the credential-token invalidation follow-ups) now — it's independent, reviewed, and the measured 2× win stands — and fold FIX 2 into a credential-unification change as a second step.
2. **Unify in one go**: rework FIX 2 against `internal/creds` first, land everything together.

I'd take path 1 and am happy to do the extraction with your commits preserved. If I don't hear from you in about a week I'll proceed that way rather than let this sit another cycle. The full merge attempt (with both failing tests reproducible) is preserved locally if you want the details.

---
Provenance: author-clustered maintainer sweep; the fold attempt and test evidence are from a maintainer working session today, not CI.

steveyegge:
Executing path 1 from the 7/26 comment (window lapsed at 12 days): FIX 1 is extracted as #5402 — the ignored-tx pool borrow, the branch-refusal hardening, the fresh-pool metric, and the mock-driver teeth tests, with your commits credited via Co-authored-by (@julianknutsen, @maphew).

One scope refinement vs the 7/26 phrasing, worth stating plainly: the credential-token invalidation follow-ups (5c2b9bad8 / 99bb19bb8 / d13434282) turned out NOT to be extractable standalone — they invalidate the per-dial credential cache, which doesn't exist on main (main's internal/creds mints once at open, so there is no per-dial cache to invalidate and the code would be semantically dead). They now travel with the FIX-2 unification, tracked maintainer-side as bd-5osca along with the seam test and the unification requirements from the fold-attempt analysis.

This PR stays open until the FIX-2 decision — nothing in it is being dropped.

_claude-fable-5 on behalf of bee (beads-lane steward)_

steveyegge:
FIX 1 is now on main: #5402 merged at 6985b3862. Remaining scope here is exactly the FIX-2 unification (bd-5osca). _— bee_

PR #4586, opened 2026-07-05
feat(gate): non-destructive fast-forward adopt detection (adopt-ff, #4551 follow-up 2)

maphew:
## Why

When the smart remote-migrate gate (#4551) finds the remote schema ahead with zero content skew, today's best verdict is `adopt` - a destructive recommendation that discards local state even when there is provably nothing to discard. The common fleet case is exactly that: a fresh or idle clone whose Dolt history is a strict ancestor of the remote head with a clean working set. Telling agents "adopt (destructive)" there causes needless hesitation and human escalations for an operation that is loss-free by construction.

This PR teaches the gate to recognize that case and say so: a new `adopt-ff` decision, a strict refinement of `adopt`, emitted only when local is a strict ancestor of the remote head AND the working set is clean. Any weaker evidence (callbacks unavailable, ancestry inconclusive, dirty working set, errors) falls back to exactly today's `adopt` behavior - the refinement can never produce a worse outcome than the status quo.

Follows the #4524/#4551 follow-up list (this is follow-up 2) and the storage-boundary charter: ancestry primitives live driver-side in `versioncontrolops`, injected into the schema gate via callbacks, since `schema` cannot import the version-control surface.

## What

**Piece 1 - ancestry primitives (`internal/storage/versioncontrolops`):**
- `LocalIsStrictAncestorOf`: full-history commit-hash set-difference; true only when `ahead == 0 && behind >= 1` (equal heads and divergence are both false).
- `WorkingSetClean` and `FastForwardAdopt` helpers (the latter is wired but never called yet - see scope note).

**Piece 2 - detection + directive wiring (`internal/storage/schema`):**
- New verdict `smartAdoptFastForward` (`Decision: "adopt-ff"`) in the smart gate router, with tailored `Error`/`userBody`/`AgentDirective`/`Options` cases; the `adopt-fast-forward` option carries `Risk: none`, justified by the guards.
- `FastForwardAdopter` callback struct; new exported entry points `CheckRemoteMigrateGateWithAdopt` / `CheckRemoteMigrateGateForRemoteWithRemoteCheckAndAdopt`. All pre-existing entry points are unchanged and route with `adopt=nil`, i.e. byte-identical behavior for existing callers.
- Injection at both store construction sites (server `dolt` and `embeddeddolt`).
- 7 routing subtests covering the full fallback matrix (happy path, dirty working set, non-ancestor, ancestry error, working-set error, nil callbacks, fork-skew short-circuit) plus a messaging/options contract test.

## Scope note

Piece 2 only detects and reports. The AUTO fast-forward execution (piece 3) is deliberately not included: `FastForward` is injected but the router never calls it, so `adopt-ff` still stops and surfaces options. Piece 3 (auto-resolve + real embedded-Dolt integration tests) follows once this lands.

## Verification

`CGO_ENABLED=0 go build/vet/test -tags gms_pure_go` clean on `./internal/storage/{schema,versioncontrolops,dolt,embeddeddolt}`; golangci-lint 0 issues; reviewed pre-push by an independent opus agent (verdict: all safety invariants hold).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Remote-ahead adopts fast-forward automatically when it is provably
+  loss-free AND lands exactly at this binary's latest migration.** When the
+  smart remote-migrate gate
+  ([#4516](https://github.com/gastownhall/beads/issues/4516)) finds the
+  remote ahead with no content skew, this clone's local Dolt history a
+  strict ancestor of the remote's with a clean working set, AND the
+  fast-forward would land exactly at the binary's own latest migration
+  (nothing pending after it, nothing beyond what this binary supports), `bd`
+  now fast-forwards to the remote's migrated schema automatically instead of
+  stopping with an adopt directive — nothing local is discarded. An unpushed
+  local commit, a dirty working set, or a remote that is not exactly at this
+  binary's latest migration all disqualify the automatic fast-forward and
+  fall back to the existing manual `bd bootstrap` adopt directive (with the
+  sharper loss-free guidance whenever ancestor+clean still hold), never a
+  forced write. Set `BD_SMART_GATE=0` to opt out and keep the manual wall for
+  every remote-ahead case, as before
+  ([#4259](https://github.com/gastownhall/beads/issues/4259)).
+
 ## [1.1.0] - 2026-07-04
 
 First stable release of the 1.1.0 line. It consolidates everything from
--- a/internal/storage/dolt/smart_remote_migrate_gate_test.go
+++ b/internal/storage/dolt/smart_remote_migrate_gate_test.go
@@ -2,12 +2,15 @@ package dolt
 
 import (
 	"context"
+	"database/sql"
 	"errors"
 	"fmt"
 	"path/filepath"
 	"testing"
 
+	"github.com/steveyegge/beads/internal/storage/doltutil"
 	"github.com/steveyegge/beads/internal/storage/schema"
+	"github.com/steveyegge/beads/internal/storage/versioncontrolops"
 )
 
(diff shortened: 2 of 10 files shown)


maphew:
Piece 3 pushed (`97eaab4ed`): the gate now executes the auto fast-forward instead of stopping with a directive, completing the design in the PR description.

What changed:

- On the `smartAdoptFastForward` verdict, the gate calls the injected `FastForward` callback (`DOLT_MERGE('--ff-only', ref)`), prints a one-line stderr notice, and returns nil - `bd` proceeds with HEAD already at the remote's migrated state, no migrations run locally.
- Auto-execution requires `remoteMax == latest` in addition to strict-ancestor + clean working set. This was the key hardening out of review: without it, a partial fast-forward (`current < remoteMax < latest`) would let `MigrateUp` run the remainder un-gated across the nondeterminism floor (the #4259 fork class), and `remoteMax > latest` would silently advance HEAD past the binary's supported schema (the #4135 class). Both windows are now closed; the TOCTOU re-check re-reads `remoteMax` fresh in the same session immediately before the merge.
- Disqualified-but-loss-free cases (read-only open, unwired callback, `remoteMax != latest`) return the `adopt-ff` directive - the operator learns a loss-free adopt is possible - while genuine merge failures degrade to the plain destructive `adopt` directive, never forcing.
- Unit tests pin the safety matrix (FF called exactly once only in the ancestor+clean+atLatest case; never on read-only/unwired/partial; degrade on merge failure). New real-Dolt cross-clone tests cover auto-FF success (HEAD == remote, ahead=0/behind=0), unpushed-commit degrade, and the below-latest partial case asserting no data motion.

Review provenance: implementation and blocker fixes by claude-sonnet-5 builder agents; adversarial review by a claude-opus reviewer and an independent `codex` (gpt-5.5) reviewer on the same diff - both independently flagged the missing `remoteMax == latest` guard, which is what piece 3's hardening addresses. Full `make test` is queued in the local verification pipeline; the real-Dolt tests need a test server and will run there.

PR #4587, opened 2026-07-05
build: auto gms_pure_go tag + make doctor-build preflight

maphew:
## Why

Omitting `-tags=gms_pure_go` is the single most common per-session build trap. A bare `CGO_ENABLED=1 go build ./cmd/bd` fails with a cryptic C-linker error (`unicode/uregex.h: No such file or directory` from go-icu-regex) because go-mysql-server links ICU by default under cgo; the `gms_pure_go` tag swaps in a pure-Go regex and avoids it. `CGO_ENABLED=0` sidesteps ICU but can't open embedded Dolt at runtime. The correct local build is `CGO_ENABLED=1 -tags gms_pure_go` — but nothing makes that automatic for a bare `go` command, and there's no diagnostic when someone gets it wrong. (This one bit me this session: a plain CGO build hit the exact ICU error.)

## What

1. **Persist the tag for bare `go build`/`go test`.** `.devcontainer/setup.sh` now runs `go env -w GOFLAGS=...-tags=gms_pure_go` so a later shell that never sourced `.buildflags` still picks up the tag. It's idempotent (skips if already present) and merges with any existing `GOFLAGS`. Placed **before** `source .buildflags`, because that source exports `GOFLAGS` into the script's shell and would otherwise shadow the on-disk `go env` value the check reads. The existing `.buildflags` sourcing is unchanged.

2. **`make doctor-build` preflight.** New target reporting `GOFLAGS` / `CGO_ENABLED` / `CC` (and whether CC is on PATH), with a PASS/WARN verdict. On WARN it names the exact `uregex.h` failure it prevents and prints the remedy (`go env -w GOFLAGS=-tags=gms_pure_go`, or the one-shot `CGO_ENABLED=1 go build -tags gms_pure_go ./cmd/bd`). It's a diagnostic (exit 0 on both paths), not a CI gate. Added to `.PHONY` and `make help`.

## Testing

```
GOFLAGS= make doctor-build                        # WARN path: names uregex.h + remedy
GOFLAGS=-tags=gms_pure_go make doctor-build       # PASS path
bash -n .devcontainer/setup.sh                    # clean
./scripts/check-build-tags.sh                     # repo's own ICU-tag guard: all clear
```
The `setup.sh` persist logic was unit-tested in isolation with a sandboxed `GOENV` (idempotent on rerun); the full `setup.sh` (which does `sudo mv`, `bd init`, etc.) was not run.

### DIFF
--- a/.devcontainer/setup.sh
+++ b/.devcontainer/setup.sh
@@ -10,7 +10,15 @@ set -e
 echo "⚙️  Persisting -tags=gms_pure_go in go env..."
 PERSISTED_GOFLAGS="$(go env GOFLAGS)"
 if [[ "$PERSISTED_GOFLAGS" != *gms_pure_go* ]]; then
-  go env -w GOFLAGS="${PERSISTED_GOFLAGS:+$PERSISTED_GOFLAGS }-tags=gms_pure_go"
+  if [[ "$PERSISTED_GOFLAGS" == *-tags=* ]]; then
+    # Merge into the existing -tags value instead of appending a second
+    # -tags flag: Go does not merge repeated -tags flags, so a second one
+    # would silently replace the first and disable the existing tags.
+    PERSISTED_GOFLAGS="$(printf '%s' "$PERSISTED_GOFLAGS" | sed -E 's/-tags=([^[:space:]]*)/-tags=\1,gms_pure_go/')"
+  else
+    PERSISTED_GOFLAGS="${PERSISTED_GOFLAGS:+$PERSISTED_GOFLAGS }-tags=gms_pure_go"
+  fi
+  go env -w GOFLAGS="$PERSISTED_GOFLAGS"
 fi
 
 # Canonical build flags (GOFLAGS=-tags=gms_pure_go, CGO_ENABLED=1).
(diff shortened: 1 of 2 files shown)


maphew:


**[P2] Merge existing -tags instead of appending another one — .devcontainer/setup.sh:13-13** — _confirmed_

When the devcontainer already has persisted GOFLAGS such as `-tags=integration`, this writes `-tags=integration -tags=gms_pure_go`. The Go command's `-tags` flag is not additive in that form: the later `-tags` replaces the earlier tag list, so existing tags are silently disabled for future bare `go` commands. Please merge `gms_pure_go` into any existing `-tags` value instead of appending a second `-tags` flag.

**[P3] Avoid passing doctor-build on overridden tag lists — Makefile:85-87** — _confirmed_

This substring check can report PASS for unsafe GOFLAGS, for example `GOFLAGS='-tags=gms_pure_go -tags=integration' make doctor-build`. Because Go applies the later `-tags` value, `gms_pure_go` is not actually active and a bare CGO build can still take the ICU path, but the diagnostic says it is safe. The check should parse the effective final tag list rather than just looking for the string anywhere.

maphew:
Pushed a fix for the build-doctor preflight check in commit 9bcc3d7b0. This addresses the finding surfaced by the 2026-07-07 cross-vendor review of this PR. The branch `feat/build-doctor-preflight` has been updated with the change; please re-review when convenient.

PR #4591, opened 2026-07-05
fix(doltserver): stop test sql-server children on parent death

maphew:
## Why

Interrupted Go test runs can leave managed `dolt sql-server` children alive after the test binary is gone. Those servers keep running against deleted tempdirs and accumulate until they are manually reaped.

Production server survival is intentional, so the fix must not make normal managed servers die with a single `bd` invocation.

## What

- Preserve the existing detached Unix process group behavior.
- On Linux only, when `BEADS_TEST_MODE=1`, add `Pdeathsig=SIGTERM` to the managed `dolt sql-server` child.
- Keep non-Linux Unix and Windows behavior unchanged.
- Cover the Linux `SysProcAttr` contract with focused tests.

Coordination: `mybd-q6cz`.

## Verification

- `go test ./internal/doltserver`
- `go test ./internal/storage/dolt -run 'TestShouldStopAutoStartedServerOnClose|TestApplyConfigDefaults'`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c ./internal/doltserver`

### DIFF
--- a/internal/doltserver/doltserver_linux.go
+++ b/internal/doltserver/doltserver_linux.go
@@ -1,0 +1,14 @@
+//go:build linux
+
+package doltserver
+
+import (
+	"os"
+	"syscall"
+)
+
+func applyTestModeParentDeathSignal(attr *syscall.SysProcAttr) {
+	if os.Getenv("BEADS_TEST_MODE") == "1" {
+		attr.Pdeathsig = syscall.SIGTERM
+	}
+}
--- a/internal/doltserver/doltserver_linux_test.go
+++ b/internal/doltserver/doltserver_linux_test.go
@@ -1,0 +1,32 @@
+//go:build linux
+
+package doltserver
+
+import (
+	"syscall"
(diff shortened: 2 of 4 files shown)


maphew:
Consolidating: closing in favor of #4592, which contains this PR's exact fix (Linux-only, `BEADS_TEST_MODE=1`-gated `Pdeathsig: SIGTERM`, detached process-group behavior preserved, production spawn unchanged) plus the suite-exit sweep this PR's description anticipated as follow-up - `SweepOrphanedTestServers` after `m.Run()` in the server-spawning TestMains, with reviewed cross-suite safety (no global temp root, deleted-cwd signature or caller-owned roots only).

Both PRs came out of coordination bead `mybd-q6cz`; two agent sessions raced on it in parallel and this one landed first - the parent-death design here matches what #4592 shipped, so nothing from this branch is lost. One consolidated PR keeps upstream review load down.

PR #4592, opened 2026-07-05
fix(doltserver): reap leaked test dolt sql-servers (test-gated Pdeathsig + suite-exit sweep)

maphew:
## Why

Interrupted test runs leak detached embedded `dolt sql-server` processes forever. Evidence (2026-07-05, Linux dev machine): 13 orphaned `dolt sql-server` processes, ages 30 minutes to 3+ days, every one reparented to init and serving a now-deleted `/tmp` test tempdir (traceable to specific Go tests: `TestFixMissingMetadata_*`, `TestCheckDatabaseVersion*`, and the `cmd/bd` guard suite's `beads-bd-tests-*` dirs).

Root cause: the server is deliberately started detached (`Setpgid: true`, no `Pdeathsig`) because the production shared server must outlive the `bd` process that starts it. When a test binary is SIGKILLed (timeout, resource limit), `t.Cleanup`/`defer` teardown never runs, `KillStaleServers` only fires at the next server *startup*, and nothing ever collects the orphans.

## What (defense in depth, production behavior unchanged)

1. **Test-gated parent-death signal (Linux):** when `BEADS_TEST_MODE=1`, `procAttrDetached` now also sets `Pdeathsig: SIGTERM`, so a test-spawned server dies with the test process even on SIGKILL. `doltserver_unix.go` (`//go:build !windows`) is split so this only compiles on Linux (`procattr_linux.go`); other unixes keep the exact current attr (`procattr_other_unix.go`); Windows path untouched. With `BEADS_TEST_MODE` unset the attr is byte-identical to before (`Setpgid: true` only) - the shared-server-outlives-bd production model is preserved. The Go-runtime thread-retirement caveat is documented in a comment; this is best-effort test hygiene, not a correctness guarantee.
2. **Suite-exit sweep:** `SweepOrphanedTestServers` runs after `m.Run()` in the `TestMain`s of the server-spawning suites (`cmd/bd` guard suite, `cmd/bd/doctor` e2e, `cmd/bd/doctor/fix`, `internal/beads`, `internal/storage/dolt`, plus a new `TestMain` for `internal/doltserver`'s `integration`-tagged tests, which previously had none). Selection logic is a pure, unit-tested function; the `/proc`-based gatherer is Linux-only with a no-op stub elsewhere.

## Safety design (the part to review hardest)

A machine-wide process sweep must never kill a real project's live shared server, and must not kill sibling test suites' live servers under parallel `go test -p N`. The sweep therefore only signals processes that:

- match the `dolt ... sql-server` cmdline shape, and
- have a cwd (== data dir) that is **deleted** (`/proc/<pid>/cwd` ends in ` (deleted)` - the observed leak signature), **or** lies under one of the *caller-suite-owned* temp roots passed in explicitly.

There is deliberately **no global `os.TempDir()` root**: an earlier draft had one, and independent review showed it would let a finishing suite reap sibling suites' still-live servers under the repo's own `scripts/test.sh -p 4`. Production servers (cwd like `.../.beads/embeddeddolt/<db>`, dir exists) match neither criterion. Signals are SIGTERM then bounded SIGKILL escalation on positive PIDs only (never a process group), with a short bounded wait, so `TestMain` cannot hang.

## Verification

- `go build -tags gms_pure_go ./...`; `go vet -tags "gms_pure_go integration" ./...`; cross-compile checks `GOOS=darwin` and `GOOS=windows` - all clean.
- New unit tests: `Pdeathsig` set iff `BEADS_TEST_MODE=1` (Linux); sweep selection pins the safety-critical negatives (live production dir, live dir outside caller's roots, non-dolt cmdline) and positives (deleted cwd; live dir under the caller's own root).
- Fast `internal/doltserver` package suite passes; the slow embedded-Dolt/integration matrices are left to CI.

Implementation by a delegated builder agent (claude-sonnet-5); independently reviewed by a delegated reviewer agent (claude-opus-4-8) - the review's blocker (global temp-dir root) and test-contract findings are incorporated.

### DIFF
--- a/cmd/bd/doctor/dolt_e2e_test.go
+++ b/cmd/bd/doctor/dolt_e2e_test.go
@@ -14,6 +14,7 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/steveyegge/beads/internal/doltserver"
 	"github.com/steveyegge/beads/internal/storage/dolt"
 	"github.com/steveyegge/beads/internal/testutil"
 )
@@ -80,6 +81,11 @@ func testMainInner(m *testing.M) int {
 	}
 
 	code := m.Run()
+
+	// Best-effort reap of any dolt sql-server left running under a temp dir
+	// this suite created (e.g. a SIGKILLed run) — see
+	// gastownhall/beads mybd-q6cz.
+	doltserver.SweepOrphanedTestServers(testBDDir)
 
 	os.Unsetenv("BEADS_DOLT_PORT")
 	os.Unsetenv("BEADS_TEST_MODE")
--- a/cmd/bd/doctor/fix/testmain_cgo_test.go
+++ b/cmd/bd/doctor/fix/testmain_cgo_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"testing"
 
+	"github.com/steveyegge/beads/internal/doltserver"
 	"github.com/steveyegge/beads/internal/testutil"
 )
 
@@ -20,8 +21,24 @@ func TestMain(m *testing.M) {
 		defer testutil.TerminateDoltContainer()
 	}
 
+	// Suite-owned root for the orphan-server sweep below. Must never be a
+	// shared/global temp dir (see SweepOrphanedTestServers) — this one is
+	// unique to this test run and removed when it exits.
+	suiteTempRoot, tempRootErr := os.MkdirTemp("", "beads-fix-tests-*")
+	if tempRootErr != nil {
+		fmt.Fprintf(os.Stderr, "WARN: failed to create suite temp root: %v\n", tempRootErr)
+	}
+
 	code := m.Run()
 
+	// Best-effort reap of any dolt sql-server left running under this
+	// suite's own temp root (e.g. a SIGKILLed run) — see
+	// gastownhall/beads mybd-q6cz.
+	doltserver.SweepOrphanedTestServers(suiteTempRoot)
+
+	if suiteTempRoot != "" {
+		os.RemoveAll(suiteTempRoot)
+	}
 	os.Unsetenv("BEADS_DOLT_PORT")
 	os.Unsetenv("BEADS_TEST_MODE")
 	os.Exit(code)
--- a/cmd/bd/test_repo_beads_guard_test.go
+++ b/cmd/bd/test_repo_beads_guard_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/steveyegge/beads/internal/config"
+	"github.com/steveyegge/beads/internal/doltserver"
 )
 
 // beforeTestsHook is set by CGO-tagged test files to perform setup before tests run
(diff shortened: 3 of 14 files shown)


maphew:
Supersedes #4591 (closed, consolidated): that PR carried the identical test-gated `Pdeathsig` piece, authored in a parallel agent session from the same coordination bead (`mybd-q6cz`). This PR adds the suite-exit sweep on top.

maphew:


**[P2] Avoid tying test CLI servers to the bd subprocess — internal/doltserver/procattr_linux.go:38-40** — _confirmed_

When `BEADS_TEST_MODE=1` is inherited by exec-based CLI tests, this runs inside the short-lived `bd` subprocess that starts the server, not the long-running test binary. `Pdeathsig` then SIGTERMs the `dolt sql-server` as soon as that command exits, so `bd dolt start` and auto-started repo-local servers no longer persist for the next `bd` invocation in test-mode scenarios that exercise cross-command lifecycle behavior. Use a more specific parent-death guard for direct test binaries or strip this flag from child CLI environments.

Verification reproduced the break concretely: `TestDoctorCheckHealthReportsVersionMismatchOnRepoLocalPort` (cmd/bd/doctor_health_integration_test.go) appends `BEADS_TEST_MODE=1` to an exec'd bd subprocess, runs `bd dolt start`, then `bd sql` / `bd doctor --check-health` against that server - with this PR the server's death-signal parent is the short-lived `bd dolt start` process, and integration-tagged tests do run on Linux CI. Telling detail: `TestE2E_AutoStartedRepoLocalServerPersistsAcrossCommands` survives only because it explicitly clears `BEADS_TEST_MODE=` in its subprocess env. One scope correction to the Codex wording: *auto-started* repo-local servers are NOT affected (resolveAutoStart forces auto-start off under BEADS_TEST_MODE=1, and test-mode auto-started servers are stopped on close). The affected surface is the explicit lifecycle paths run in a BEADS_TEST_MODE=1 subprocess: `bd dolt start` (dolt.go:541), `bd init --server` (init.go:977), and the config-apply restart (config_apply.go:408).

**[P2] Revalidate the PID before force-killing after SIGTERM — internal/doltserver/sweep_linux.go:70-73** — _confirmed_

After the 300ms grace period this sends `SIGKILL` to every PID that accepted `SIGTERM`, even if the original server exited during the grace period. In a PID-reuse window, the sweep can kill an unrelated process despite the surrounding logic being careful to identify only orphaned test servers. Recheck that the PID still refers to the same `dolt sql-server` candidate before escalating.

Verification note: Severity caveat: the window requires the kernel to cycle the entire PID space within ~300ms at test-suite exit, so real-world likelihood is very low; this is a hardening gap in a safety-critical path rather than a probable production kill. Note the same no-revalidation pattern also exists between candidate gathering and the initial SIGTERM (sweep_linux.go:48-57), which any fix should cover too.

maphew:
This branch has been updated to fix the flaky test-server reap behavior, pushed at commit 23c6205f3. The change addresses the finding surfaced in the 2026-07-07 cross-vendor review of unreviewed upstream PRs, where the test-server teardown was identified as a source of intermittent failures. The fix is verified clean and ready for another look.

PR #4595, opened 2026-07-06
fix(import): respect import.auto=false in the empty-DB import-on-write path

### LINKED ISSUE #4304: bd update auto-imports JSONL into empty DB despite import.auto=false
From Gas Town recovery gt-dolt-hook-mail-persistence-recovery with bd 1.0.4 (ce242a879):

Repro observed in /home/coder/gt/gastown/polecats/chrome/gastown:

1. Run: BD_DOLT_AUTO_COMMIT=on bd update gt-dolt-hook-mail-persistence-recovery --notes ...
2. bd prints: auto-importing /home/coder/gt/.beads/issues.jsonl into empty database; auto-imported N issues.
3. bd reports success.
4. The authoritative gastown issue does not show the note.
5. bd -C /home/coder/gt show gt-dolt-hook-mail-persistence-recovery shows the note on an hq-imported duplicate gt-* issue.

Also reproduced with BEADS_NO_AUTO_IMPORT=1 and with BD_IMPORT_AUTO=false; auto-import still occurred. bd config show recognizes BD_IMPORT_AUTO=false, but the write path ignores it.

Expected: import.auto=false / no-auto-import must prevent auto-import, and routed writes should target the authoritative prefix database or fail loudly rather than importing stale JSONL into an empty/default database and reporting success.

maphew:
Fixes #4304.

## Why

With `import.auto=false` (env `BD_IMPORT_AUTO=false` or `config.yaml`), any write command (`bd update`, `bd close`, ...) against an empty database with a `.beads/issues.jsonl` present still silently auto-imported the JSONL - the exact data-safety hazard #4304 reports. Reproduced on current main (c4e51986b): `bd config show` reports `import.auto = false (env_var)`, yet `bd update` prints `auto-imported 2 issues from .../issues.jsonl`.

## Root cause

`shouldRunAutoImportJSONL` (cmd/bd/main.go), the gate in front of the empty-DB upgrade-recovery import that every write command runs through, checked read-only mode, `--global`, server mode, and `cmd.Name() != "import"` - but never consulted `import.auto`. Only the separate git-hook sync path (`importJSONLForSync` in cmd/bd/hooks.go) honored the key, so disabling auto-import silenced the hook importer while the import-on-write path kept firing.

## What

- Add the same `config.GetBool("import.auto")` guard to `shouldRunAutoImportJSONL`, making the two auto-import paths consistent. The registered default is `true` (internal/config/config.go), so default behavior - empty-DB recovery import on write - is unchanged; only an explicit opt-out is now respected everywhere.
- Regression test `TestShouldRunAutoImportJSONL_RespectsImportAutoFalse`, plus config initialization in the pre-existing table test so its default-on cases assert the registered default rather than ambient state.
- Update the `import.auto` entry in docs/CONFIG.md: the key is now the master auto-import switch, gating both the post-merge/checkout hook fallback and the empty-database import-on-write path.

## Scope note for reviewers

This deliberately broadens `import.auto` from "legacy hook fallback switch" to "all auto-import" - a user who sets `import.auto=false` expects no silent imports anywhere, per #4304. Explicit `bd import` is unaffected (it short-circuits on command name independent of the flag), as are server mode and read-only commands.

Follow-up commit 24dc3af also folds in the adjacent mybd-gai7 bug: `import.auto` is now marked YAML-only like `import.path`, so `bd config set import.auto false` writes to `.beads/config.yaml` where the startup auto-import gates read it.

Verified: pure-Go build, targeted `go test` on the auto-import gate/import tests, gofmt clean; end-to-end repro confirmed suppressed with the fixed binary (env var and config.yaml variants), and default-on behavior confirmed intact. Follow-up local checks: `go test ./internal/config -run TestIsYamlOnlyKey` and `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run TestConfigSetMany`.

Tracked in mybd as beads mybd-3fvn and mybd-gai7.

_claude-fable-5-high on behalf of matt wilkie_ (implementation: claude-sonnet-5 builder agent; review: claude-opus reviewer agent)

_codex-gpt-5.5-high on behalf of matt wilkie_ (follow-up mybd-gai7 fix)

### DIFF
--- a/cmd/bd/auto_import_upgrade_unit_test.go
+++ b/cmd/bd/auto_import_upgrade_unit_test.go
@@ -224,6 +224,49 @@ func TestShouldRunAutoImportJSONL_RespectsImportAutoFalse(t *testing.T) {
 	}
 }
 
+// TestIsDisablingImportAutoViaConfigCommand is the regression test for the
+// P2 finding on gastownhall/beads#4595 (Codex cross-vendor review,
+// 2026-07-07): shouldRunAutoImportJSONL runs in PersistentPreRun before
+// "bd config set import.auto false" writes the new value, so the command
+// meant to disable auto-import triggered it on its own invocation whenever a
+// stale .beads/issues.jsonl sat next to an empty database (GH#4304). This
+// exemption must fire for "config set import.auto false"/"0" and for
+// "config set-many ... import.auto=false", but not for unrelated config
+// keys, truthy values, or commands outside the config subtree.
+func TestIsDisablingImportAutoViaConfigCommand(t *testing.T) {
+	configCmd := &cobra.Command{Use: "config"}
+	setCmd := &cobra.Command{Use: "set"}
+	setManyCmd := &cobra.Command{Use: "set-many"}
+	configCmd.AddCommand(setCmd, setManyCmd)
+	writeCmd := &cobra.Command{Use: "update"}
+
+	tests := []struct {
+		name string
+		cmd  *cobra.Command
+		args []string
+		want bool
+	}{
+		{name: "config set import.auto false", cmd: setCmd, args: []string{"import.auto", "false"}, want: true},
+		{name: "config set import.auto 0", cmd: setCmd, args: []string{"import.auto", "0"}, want: true},
+		{name: "config set import.auto true", cmd: setCmd, args: []string{"import.auto", "true"}, want: false},
+		{name: "config set other key", cmd: setCmd, args: []string{"export.auto", "false"}, want: false},
+		{name: "config set-many import.auto=false", cmd: setManyCmd, args: []string{"jira.url=x", "import.auto=false"}, want: true},
+		{name: "config set-many import.auto=true", cmd: setManyCmd, args: []string{"import.auto=true"}, want: false},
+		{name: "unrelated command", cmd: writeCmd, args: []string{"import.auto", "false"}, want: false},
+		{name: "set without config parent", cmd: &cobra.Command{Use: "set"}, args: []string{"import.auto", "false"}, want: false},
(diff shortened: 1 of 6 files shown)


maphew:


**[P2] Skip auto-import before setting import.auto — cmd/bd/main.go:1350-1352** — _confirmed_

When an empty database still has a non-empty `.beads/issues.jsonl`, running `bd config set import.auto false` does not actually prevent the recovery import on that invocation: `shouldRunAutoImportJSONL` runs in `PersistentPreRun` before the config command writes the new YAML value, so it sees the old default `true` and imports first. This makes the documented master switch impossible to enable safely from the CLI in the exact stale-JSONL scenario it is meant to guard; the config command should bypass auto-import when it is setting `import.auto` (or otherwise apply this value before the pre-run import check).

Verification note: One nuance: 'impossible to enable safely from the CLI' is marginally overstated - viper's AutomaticEnv binding (SetEnvPrefix("BD") + '.'->'_' replacer, internal/config/config.go:165-169) means `BD_IMPORT_AUTO=false bd config set import.auto false` is honored on that same invocation, and editing .beads/config.yaml directly is also safe. The plain `bd config set import.auto false` path is broken exactly as described, so the P2 stands; the suggested fix (exempt config-set-of-import.auto from the pre-run import, or apply the pending value before the check) is apt.

maphew:
Pushed a follow-up fix in commit bda985024 that addresses the auto-import edge case flagged in this PR. The change has been verified clean in the working worktree and force-updates the `fix/3fvn-4304-auto-import` branch.

This also resolves the finding raised in the 2026-07-07 cross-vendor review of this PR.

PR #4598, opened 2026-07-06
fix(bootstrap): retry failed clone cleanup

### LINKED ISSUE #4310: [Windows] bd bootstrap fails to clean up failed dolt clone due to LOCK file being in use
## Bug Report

On Windows, when `bd bootstrap` encounters a `dolt clone` failure, the subsequent cleanup step fails with a file locking error, leaving the workspace in a broken state.

### Error Message
```text
Error: Bootstrap failed: clone from remote: dolt clone git+https://github.com/maphew/mybd.git: Error 1105: clone failed; manifest referenced table file for which there is no chunkSource.: unable to clean up failed clone in directory 'beads': unlinkat A:\dev\mybd\.beads\embeddeddolt\beads\.dolt\noms\LOCK: The process cannot access the file because it is being used by another process.
```

### Root Cause
When `dolt clone` fails, it attempts to clean up the partially created `.dolt` directory. However, Windows prevents the deletion of the `noms/LOCK` file because a process (often the crashed `dolt` process itself, or an overzealous antivirus scanner) still holds an active file handle to it.

### Expected Behavior
`beads` should handle Windows file locking gracefully during cleanup. Potential improvements include:
1. Implementing a retry mechanism with a short delay before attempting cleanup, giving the OS time to release the file handle.
2. Detecting and terminating orphaned `dolt` processes before attempting to clean up a failed clone.
3. Providing a clearer, actionable error message (e.g., "Cleanup failed due to a Windows file lock. A previous dolt process may still be running. Try closing it or manually deleting the `.beads` directory and retrying.")

### Workaround
Manually kill any running `dolt` or `bd` processes, delete the `.beads` directory, and retry `bd bootstrap`.

### Environment
- OS: Windows
- `bd` version: latest

maphew:
Fixes #4310.

## Why

On Windows, a failed `dolt clone` during `bd bootstrap` can leave a partial clone target behind when Dolt's own failure cleanup cannot delete `.dolt/noms/LOCK` because another process or scanner still has the file open. The user then sees the clone failure plus a nested cleanup failure and may be left with a broken partial target.

## What

- After `dolt clone` fails in the CLI/owned-server bootstrap path, retry cleanup of the clone target briefly.
- Only remove the target when it contains a real `.dolt` directory, so a pre-existing non-Dolt directory with the configured database name is preserved.
- If cleanup still fails, add actionable Windows guidance about stuck `dolt`/`bd` processes, antivirus file handles, manual deletion, and retrying `bd bootstrap`.
- Preserve the original clone error and output.

## Verification

- `go test ./internal/storage/dolt -run 'Test(BootstrapFromRemote|DoltCloneArgs|RemoveFailedCloneTarget|FormatFailedCloneTarget)'`
- `git diff --check`

Tracked in mybd as bead mybd-kj2v.

### DIFF
--- a/internal/storage/dolt/bootstrap.go
+++ b/internal/storage/dolt/bootstrap.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/steveyegge/beads/internal/configfile"
@@ -54,8 +53,8 @@ func BootstrapFromRemoteWithDB(ctx context.Context, doltDir, remoteURL, database
 		return false, fmt.Errorf("invalid remote URL: %w", err)
 	}
 
-	if strings.TrimSpace(database) == "" {
-		return false, fmt.Errorf("database name must not be empty; use cfg.GetDoltDatabase() to resolve the configured name")
+	if err := ValidateDatabaseName(database); err != nil {
+		return false, fmt.Errorf("invalid database name %q (use cfg.GetDoltDatabase() to resolve the configured name): %w", database, err)
 	}
 
 	// Verify dolt CLI is available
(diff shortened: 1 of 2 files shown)


maphew:


**[P1] Avoid deleting pre-existing clone targets — internal/storage/dolt/bootstrap.go:103-104** — _confirmed_

This cleanup runs after any `dolt clone` failure and recursively removes `path` whenever `path/.dolt` exists, even if that target was not created by this clone attempt. Because `database` is accepted from config/env without being constrained to a single path element, values like `foo/bar` or `../other-db` can make `doltExists(doltDir)` miss an existing Dolt repo; a normal "target already exists" clone failure would then delete that repo. Please validate the database name as a basename and/or only remove targets known not to have existed before cloning.

Verification traced the full chain and reproduced it: the pre-clone guard `doltExists(doltDir)` reads only immediate children (`doltDir/<entry>/.dolt`), so with `dolt_database="foo/bar"` it checks `doltDir/foo/.dolt` (absent) while the clone target is `doltDir/foo/bar`; with `"../other-db"` the cleaned path is a sibling `ReadDir` never sees. The value reaches the clone unvalidated - `GetDoltDatabase` returns `BEADS_DOLT_SERVER_DATABASE` or metadata.json `dolt_database` verbatim, and `ValidateDatabaseName` (which would reject these) is applied only on `bd init` flags and the `CREATE DATABASE` path, never before the CLI clone. End-to-end: first bootstrap clones into `doltDir/foo/bar`; a later bootstrap misses it, re-clones, `dolt clone` fails "target already exists", and the new cleanup `RemoveAll`s the pre-existing repo including local commits. Pre-PR the same failure left the repo intact, so this is a data-loss regression introduced by the patch. Scope caveat: reaching it requires a hand-edited metadata.json or hand-set env var (default and `bd init`-created setups are name-validated), but hand-edited metadata is exactly the recovery scenario bootstrap serves. Cheapest fix: run `ValidateDatabaseName` on the value at the top of `BootstrapFromRemoteWithDB`, and/or stat the target before cloning and skip cleanup when it pre-existed.

maphew:
Pushed the fix for this issue as commit 106658185, which cleans up the bootstrap handling identified in the report. This addresses the finding from the 2026-07-07 cross-vendor review of own upstream PRs. The branch `fix/kj2v-bootstrap-cleanup` is ready for review.

PR #3812, opened 2026-05-07
test(cmd/bd): tolerate timeout-killed processes in TestEmbeddedInitConcurrent

seanmartinsmith:
Fixes #3811

## What

Makes `TestEmbeddedInitConcurrent` tolerant of timeout-killed processes under CI runner pressure. Two changes, both in `cmd/bd/init_embedded_test.go`:

- **Invariant fix**: a third outcome category (`timeoutKills`) is introduced. Timeout-killed processes are now logged rather than immediately failing the test. The final assertion becomes `successes + lock_errors + timeoutKills == N`. A cap (`timeoutKills > 2`, i.e. > N/5 for N=10) still flags systemic runner problems. A new "at least 1 lock error" guard is added alongside the existing "at least 1 success" guard so the test cannot silently pass if every process is killed.
- **Timeout bump**: per-process timeout raised from 45 s to 90 s. Defense-in-depth; does not replace the invariant fix.

## Why

PR #3555 (`Test (Embedded Dolt Cmd 16/20)`, 2026-04-27) failed with:

```
init_embedded_test.go:832: process 0 timed out after 45s running concurrent bd init: signal: killed
init_embedded_test.go:850: expected successes (2) + lock errors (7) = 10, got 9
```

Run: https://github.com/gastownhall/beads/actions/runs/25016789724/job/73267367100

Process 0 was killed by the test's own 45 s timeout before it could resolve. The old invariant (`successes + lock_errors == N`) has no bucket for that outcome, so the test counted 9/10 and failed. PR #3555's actual diff was 4 lines in an unrelated package — it cannot causally affect concurrency in `cmd/bd/`. The failure is infrastructure noise surfacing a brittle test invariant.

## Verification

```bash
# build
make build

# test (requires embedded dolt runner with BEADS_TEST_EMBEDDED_DOLT=1)
go test -tags embedded_dolt -run TestEmbeddedInitConcurrent -v ./cmd/bd/...

# without embedded dolt: test skips cleanly
go test -tags gms_pure_go -run TestEmbeddedInitConcurrent -v ./cmd/bd/...
```

One file changed, no production code touched.

---
[...]

PR #3806, opened 2026-05-07
test(hooks): skip 5 POSIX-shebang tests on Windows (GH#3800)

seanmartinsmith:
Fixes #3805

## What

Skip 5 POSIX-shebang hook tests on Windows with explanatory comments and GH#3800 references. Five tests, five skip guards, no other changes. Companion to PR #3802 (`TestUpdateCloseHookFiring`), same root cause, same fix pattern.

Tests skipped:
- `TestRunSync_Success`
- `TestRunSync_HookFailure`
- `TestRun_Async`
- `TestRunSync_ReceivesJSON`
- `TestRunSync_Timeout`

## Why

Each test writes an extensionless `#!/bin/sh` script as a hook file, then asserts execution side effects (file created, stdin captured, non-zero exit code, timeout error). On Windows the hook runner (`internal/hooks/hooks_windows.go`) delegates to `exec.Command` → `CreateProcess`, which has no shebang dispatch. The hook silently no-ops: no file is written, stdin is never read, the expected error from `exit 1` is never returned, and `RunSync` returns immediately rather than hitting the timeout path.

Root cause in one line: tests assume POSIX shell execution; the Windows runner provides none.

The right long-term fix is PATHEXT-aware extension lookup and interpreter dispatch in the Windows runner. That is tracked in GH#3800 and is deliberately out of scope here, same call as PR #3802.

## Verification

Before (Windows, HEAD `8b7cbfc3b`):

```
=== RUN   TestRunSync_Success
    hooks_test.go:171: Failed to read output file: open ...\output.txt: The system cannot find the file specified.
--- FAIL: TestRunSync_Success (0.00s)
=== RUN   TestRunSync_HookFailure
    hooks_test.go:329: RunSync should have returned error for failed hook
--- FAIL: TestRunSync_HookFailure (0.00s)
=== RUN   TestRun_Async
    hooks_test.go:365: Failed to read output file after retries: open ...\async_output.txt: The system cannot find the file specified.
--- FAIL: TestRun_Async (0.00s)
=== RUN   TestRunSync_ReceivesJSON
    hooks_test.go:210: Hook did not receive JSON input
--- FAIL: TestRunSync_ReceivesJSON (0.00s)
=== RUN   TestRunSync_Timeout
    hooks_test.go:267: RunSync should have returned error for timeout
--- FAIL: TestRunSync_Timeout (0.00s)
```

After:

```
=== RUN   TestRunSync_Success
[...]

PR #3802, opened 2026-05-07
test: skip TestUpdateCloseHookFiring on Windows (GH#3800)

seanmartinsmith:
Fixes #3800

## What

Skip `TestUpdateCloseHookFiring` on Windows with an explanatory comment and issue reference. One test, one skip guard, no other changes.

## Why

The test writes a POSIX shell script (`#!/bin/sh`) as the hook file, then calls `runner.RunSync(...)` and asserts the hook created a marker file. On Windows the hook runner (`internal/hooks/hooks_windows.go`) executes hooks via `exec.Command(ctx, hookPath, ...)`, which delegates to `CreateProcess`. Windows has no shebang dispatch: an extensionless file starting with `#!/bin/sh` is not runnable as a shell script, the exec silently no-ops (no error, no output), and the marker file is never created.

Root cause in one line: test setup assumes POSIX shell execution; the Windows runner provides none.

The right long-term fix is to add PATHEXT-aware extension lookup and interpreter dispatch to the Windows runner (look for `on_close.ps1`, `on_close.cmd`, etc. and invoke the appropriate interpreter). That is tracked in GH#3800 and is a non-trivial runner change, deliberately out of scope here.

## Verification

Before (Windows, HEAD `8b7cbfc3b`):

```
=== RUN   TestUpdateCloseHookFiring/on_close_fires_for_status_transition_to_closed
    update_close_hook_test.go:49: on_close hook did not create marker file: ...The system cannot find the file specified.
--- FAIL: TestUpdateCloseHookFiring (0.00s)
```

After:

```
=== RUN   TestUpdateCloseHookFiring
    update_close_hook_test.go:33: hook script execution not supported on Windows - see GH#3800
--- SKIP: TestUpdateCloseHookFiring (0.00s)
PASS
ok  	github.com/steveyegge/beads/cmd/bd	0.218s
```

`TestUpdateCloseHookCondition` (pure logic, no exec) continues to pass on Windows unchanged. `make build` clean.

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/beads/pr/3802"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
[...]

PR #4600, opened 2026-07-06
test: port Windows hardening fixes

maphew:
## Why

Port the live Windows test-hardening fixes from contributor PRs #3797, #3801, #3802, #3806, and #3812 onto current `main`.

This keeps the contributor value while avoiding stale-branch drift in the orphan-band queue. #3801 now conflicts because the Dolt server test moved to `internal/storage/dbproxy/server/doltserver_test.go`, so that fix is re-applied at the current path. The release-script test also needs Windows-hosted Bash path handling so the fake `bd` helper is selected reliably during tests.

## What Changed

- Let release-script tests use Bash-native paths for the fake `bd` helper.
- Stop Windows hook-script tests that rely on POSIX shebang execution until the runner supports it.
- Ensure Dolt server cleanup closes the log handle before Windows temp-dir removal.
- Tolerate bounded timeout kills in `TestEmbeddedInitConcurrent`.

## Validation

- `CGO_ENABLED=0 go test -tags gms_pure_go ./scripts`
- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^TestUpdateCloseHookFiring$'`
- `go test ./internal/hooks -run '^(TestRunSync_Success|TestRunSync_ReceivesJSON|TestRunSync_Timeout|TestRunSync_HookFailure|TestRun_Async)$'`
- `go test ./internal/storage/dbproxy/server -run '^(TestDoltServer_|TestNewDoltServer_)'`
- `BEADS_TEST_EMBEDDED_DOLT=1 go test -tags 'cgo gms_pure_go' ./cmd/bd -run '^TestEmbeddedInitConcurrent$' -count=1`

Co-authored-by: sms <114885497+seanmartinsmith@users.noreply.github.com>

### DIFF
--- a/cmd/bd/init_embedded_test.go
+++ b/cmd/bd/init_embedded_test.go
@@ -1877,7 +1877,7 @@ func TestEmbeddedInitConcurrent(t *testing.T) {
 	for i := 0; i < N; i++ {
 		go func(idx int) {
 			defer wg.Done()
-			ctx, cancel := context.WithTimeout(context.Background(), 45*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
 			defer cancel()
 
 			cmd := exec.CommandContext(ctx, bd, "init", "--prefix", "conc", "--force", "--quiet", "--skip-agents")
@@ -1889,10 +1889,11 @@ func TestEmbeddedInitConcurrent(t *testing.T) {
 	}
 	wg.Wait()
 
-	successes, lockErrors := 0, 0
+	successes, lockErrors, timeoutKills := 0, 0, 0
 	for _, r := range results {
 		if r.timedOut {
-			t.Errorf("process %d timed out after 45s running concurrent bd init: %v\n%s", r.idx, r.err, r.out)
+			t.Logf("process %d timed out after 90s running concurrent bd init: %v\n%s", r.idx, r.err, r.out)
+			timeoutKills++
 			continue
 		}
 		if strings.Contains(r.out, "panic") {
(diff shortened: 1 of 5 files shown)


maphew:


**[P2] Keep release fixtures out of the source tree — scripts/release_script_test.go:200-204** — _confirmed_

When the tests run from a read-only checkout or packaging build directory, this now fails before the release-script tests execute because it tries to create `.tmp-release-tests` under `sourceRepoRoot`. These fixtures were previously under `t.TempDir()`, so the tests did not require write access to the repository; keep the temp fixture outside the source tree and convert paths for Bash separately.

**[P2] Preserve the caller PATH when adding fake bd — scripts/release_script_test.go:229-229** — _confirmed_

On systems where `bash` or the core utilities used by `release.sh` are not in `/usr/bin` or `/bin` (for example Nix/Guix-style environments), the inner `bash scripts/release.sh` command now fails because this replaces PATH instead of prepending the fake `bd` directory to the existing environment. The previous test preserved `os.Getenv("PATH")`; append the converted existing PATH here rather than hard-coding only `/usr/bin:/bin`.

maphew:
Pushed 95d6159c0 addressing both P2 findings from the 2026-07-07 cross-vendor review: test fixtures now live in `t.TempDir()` instead of the source tree, and the test PATH prepends the fake bin to the caller's converted PATH rather than replacing it. A follow-up review pass caught that the reconstruction initially dropped the `/usr/bin:/bin` baseline git-bash coreutils need on Windows; the pushed commit keeps that baseline appended.

PR #4601, opened 2026-07-06
feat: choose your own storage backend (Postgres / MySQL / SQLite behind the Dolt-parity seam)

julianknutsen:
## What

This branch makes the bd storage backend **user-selectable**. Alongside the existing Dolt backend (which stays the default, byte-identical on the default path), `bd init` now supports:

```bash
bd init --backend=postgres --pg-url="postgres://bd:pw@db.example.com:5432/beads?sslmode=require" --pg-schema=myrepo
bd init --backend=mysql    --mysql-url="bd:pw@tcp(db.example.com:3306)/" --mysql-database=myrepo_beads
bd init --backend=sqlite   [--sqlite-path=beads.db]
bd init --backend=dolt     # default; unchanged
```

(Passwords in the init URLs are used for provisioning only and are redacted — via the driver's own parser, verified by re-parse — before the DSN is persisted to `.beads/metadata.json`.)

All three SQL backends implement the **full semantic core** of bd — issue CRUD, search/counts, dependencies and cycle detection, ready/blocked work, claim/lease, labels, comments, events, wisps, config, slots, statistics, streaming iterators, transactions — with behavior conformance-tested to be **byte-for-byte identical to the Dolt reference**. What they deliberately do *not* implement is version control: history, branches, remotes, sync, federation, and compaction are Dolt-only capabilities, and on a SQL backend each returns a clean typed error instead of a panic or a silent no-op:

```
$ bd history demo-9q9
Error: failed to get history: operation "History" not supported by the sqlite backend
```

`bd init` sets the expectation up front on every non-Dolt backend: *"History: not tracked (SQLite backend has no version control; use Dolt for history)."*

## Why

- **Deployment fit.** Teams that already run Postgres or MySQL can point bd at infrastructure they operate, back up, and secure today. SQLite gives a zero-server, zero-credential, single-file workspace (pure-Go driver).
- **Dolt remains the reference.** Dolt is the default, the only backend with history, and the correctness oracle every other backend is tested against. This feature does not fork bd's semantics — it proves the other backends match them.
- **Scope discipline.** Per `docs/PROJECT_CHARTER.md`, beads "should not become a storage engine" — and this PR doesn't add storage-engine surface area. It swaps the substrate *behind the existing storage seam*: same interface, same commands, same semantics, minus the commit graph.

## How this was built: oracle-driven development

The methodology is the headline, because it's what makes a 64-commit backend swap reviewable. The loop, repeated per backend:

1. **Build a conformance oracle** — a corpus that runs against the embedded-Dolt reference and captures its *exact* behavior (output bytes, error strings, event contracts, sort orders, NULL placement).
2. **Introduce a backend** — a schema port plus a thin SQL dialect.
3. **Run the oracle differentially** against the new backend.
4. **Fix divergences until byte-for-byte compliant.** Then start the loop again for the next backend.

The commit history reads in exactly this order: oracle rig first, then Postgres, then MySQL, then SQLite, each followed by fix-until-compliant commits. The audit-corpus commit states the result: running it "flushed out ~11 real divergences, all fixed in the preceding commits… The corpus is green on all four backends."

Concrete divergences the oracle caught — none of which a unit test was ever going to find:

- **SQLite `LOCATE` vs `INSTR`.** The recursive parent-descendant CTE uses MySQL `LOCATE(needle, haystack)`; SQLite only has `INSTR` with *reversed* arguments. Effect: `bd ready --parent <epic>` silently returned an empty set on SQLite. The audit case `ReadyParentTransitiveDescendants` caught it; the fix is an arg-swapping rewrite in the SQLite dialect.
- **Postgres strictness and NULL ordering.** `COALESCE(priority, '')` mixes integer and text — MySQL shrugs, Postgres errors; and Postgres sorts NULLs opposite to Dolt on `DESC`, so `closed_at`/assignee sorts diverged. Fixed with a portable `CAST(... AS CHAR)` (translated to `::text`) and an explicit `(col IS NULL)` leading sort key.
- **Reopen semantics, not SQL at all.** `sqlkit.ReopenIssue` used `WHERE status='closed'`, making reopen-of-an-open-issue a silent no-op — but the Dolt reference always mints a status event, bumps `updated_at`, and wraps `storage.ErrNotFound`. The oracle flagged the missing event; the fix mirrors the reference's composition exactly.
- **Determinism hardening that improved Dolt itself.** The ambiguous-partial-ID error listed matches in `created_at DESC` order — same set, different display order between Postgres's whole-second timestamps and Dolt's sub-second ones. Only the deep 523-scenario differential caught it; the fix (`sort.Strings`) made the output deterministic on *every* backend, including Dolt.

Every backend profile's expected-failure list is empty: **no known divergences remain.**

## Testing (regression prevention)

Two tiers, one entry point, **local == CI** (`.github/workflows/conformance.yml` runs the same `scripts/conformance.sh` a contributor runs):

**Tier 1 — in-process conformance.** `conformance.RunAll` runs the same ~40 behavior subtests against embedded-Dolt, Postgres, MySQL, and SQLite: CRUD, search, deps, ready/blocked, claim/lease, labels, comments, config, slots, stats, iterators, transactions, plus:
- the **audit corpus** (~94 cases derived from a systematic audit of the Dolt reference — close/reopen idempotency, delete cascade/force/dry-run, cycle/self/external dependency routing, ready-work exclusions, NULL placement, custom-status normalization, molecule rollups, …). Its doc comment states the contract: "a failure here on a SQL backend is a genuine divergence from the reference, not a bad test";
- the **claim/lease oracle** (9 cases, including dead-worker lease reclaim, made deterministic via an injectable lease TTL);
- **`TestInterfaceCompleteness`** — parses the generated unsupported shell and asserts it equals the audited allowlist *in both directions* (nothing silently unsupported, no stale entries);
- **`TestUnsupportedContract`** — reflectively calls all 44 allowlisted methods and asserts each returns `*storage.ErrUnsupported` (not a panic, not nil, not some other error), DB-free;
- dialect gates: a corpus test that `PREPARE`s every statement family bd emits against live Postgres, and password-redaction round-trip tests.

**Tier 2 — end-to-end CLI differential** (build tags `gms_pure_go e2e`). Builds a real `bd` binary, runs `bd init` plus a ~30-scenario CLI corpus in an isolated workspace per backend, and asserts byte-equal normalized output against the `dolt-embedded` reference profile. Only timestamps/paths/schema handles are normalized (IDs are pinned with `--id`). Expected-failure entries are visible-never-masked, and an XFAIL that *starts passing* fails the run — the list can only shrink. A companion round-trip test pins export → import → re-export JSONL as a lossless multiset per backend.

**Adding a backend = one profile entry.** `test/conformance/profiles.go` is a declarative registry — `Name`, `Available` (env-gated), `InitArgs` (the exact `bd init` flags), `Env`, `NewHandle`/`Teardown`, `XFail` — and both Tier-2 suites iterate it automatically. A future backend needs one registry entry plus a store-factory arm, and the whole harness picks it up.

**Real databases in CI.** The conformance workflow runs `postgres:16` and `mysql:8` service containers on every PR (`BEADS_PG_TEST_URL`, `BEADS_MYSQL_TEST_URL`); Dolt-embedded and SQLite run in-process with no env at all. There is also an optional deep gate (`CONFORMANCE_DEEP=1`, ~50 min, 523 scenarios) outside the per-PR loop; it caught the two subtlest divergences above.

## Override layers & dialects

**One shared implementation, thin dialects.** All three SQL backends share a single `sqlkit.Store` (~104 methods, ~2,070 LOC) parameterized by a strategy bundle `{Dialect, Readiness, Claim}`, which itself delegates to the backend-neutral `issueops` layer (~11,400 LOC) where query/graph/ordering/ID-mint semantics live *once*. A concrete backend package is deliberately tiny — embed `*sqlkit.Store`, add a ~20-line dialect, DSN/credential handling, `Provision`, and the generated unsupported shell. Non-test hand-written glue is on the order of 200–400 lines per backend.

Dialects translate bd's canonical MySQL-dialect SQL at `PREPARE` time: Postgres rewrites `CONCAT`/`LOCATE`/`DATE_FORMAT`/`INSERT IGNORE`/`ON DUPLICATE KEY`/placeholders (`?`→`$n`); SQLite handles `INSERT OR IGNORE`, the `INSTR` arg reversal, and JSON aggregate renames; MySQL is nearly the identity dialect (one rewrite to dodge error 1093, plus UTC session pinning). Crucially, the translators are **fail-loud**: they handle only the finite set of constructs bd actually emits, and anything unrecognized dies as a syntax error at `PREPARE` — a divergence is a build failure, not a subtly wrong row set. This is what makes "dialects save enormous work" safe rather than scary.

**Credential ladder (per backend, fail-closed).** Secrets never reach disk: `bd init` redacts the password from `--pg-url`/`--mysql-url` *before* persisting the DSN to `.beads/metadata.json`, using the driver's own parser and re-verifying nothing survived (refusing init rather than persisting a secret in an unanticipated DSN shape). Every subsequent command resolves the password at open time, highest configured rung wins, and a configured-but-failing rung **aborts the open** — it never silently falls through:

1. Password embedded in an explicit DSN (`BEADS_POSTGRES_URL` / `BEADS_MYSQL_URL` / init flags) — wins outright
2. `BEADS_PG_PASSWORD_COMMAND` / `BEADS_MYSQL_PASSWORD_COMMAND` — the standard credential-process idiom (kubectl `ExecCredential` / AWS `credential_process` / git credential-helper): bd runs the command, accepts a bare token or a JSON envelope with an expiry, and caches until just before expiry. Works with any secret manager or short-lived-token issuer (e.g. Vault, RDS IAM auth tokens)
3. `BEADS_PG_PASSWORD` / `BEADS_MYSQL_PASSWORD` env vars
4. Credentials file (`$BEADS_CREDENTIALS_FILE`, default `~/.config/beads/credentials`), INI sections keyed by `[host:port]`, with an ssh-style permissions warning
5. Driver-native fallbacks — Postgres honors `PGPASSWORD`/`~/.pgpass` via pgx; MySQL deliberately has no ambient fallback and connects with an empty password only when nothing is configured

The Dolt *server* backend additionally supports an **identity** credential command (`BEADS_DOLT_CREDENTIAL_COMMAND`): a short-lived token presented as the connection username to an authenticating gateway server, which verifies it and routes to the database — bd stays a passive client. Credentials are typed at the config slot (secret vs identity), so an identity token can never land in a password field or vice versa; the direct Postgres/MySQL backends reject `BEADS_PG_CREDENTIAL_COMMAND` / `BEADS_MYSQL_CREDENTIAL_COMMAND` outright. All `*_COMMAND` slots are read from the environment only, never from workspace metadata — a persisted command would be arbitrary code executed on open, and that waits on a workspace-trust gate. SQLite needs no credentials at all.

## Core vs capabilities

```mermaid
flowchart TD
    CLI["bd CLI"] --> SEAM["Storage interface<br/>one seam — every backend satisfies all of it"]

    SEAM --> CORE["CORE — issueops<br/>CRUD · search · deps · ready/claim · ordering · id-mint<br/>written once, shared by EVERY backend"]
    SEAM --> CAPS["CAPABILITIES — version control<br/>history · branch · remotes · sync"]

    CORE --> DS["Dolt store<br/>runs the core natively"]
    CORE --> KIT["sqlkit.Store<br/>SQL-family adapter · Dialect / Readiness / Claim"]

    DS --> DOLT[("Dolt — embedded / server<br/>the conformance ORACLE")]
    KIT --> PGD["pgdialect"] --> PG[("Postgres")]
    KIT --> MYD["mysqldialect"] --> MY[("MySQL")]
    KIT --> LTD["sqlitedialect"] --> LT[("SQLite")]

    CAPS -->|"implemented in the Dolt store"| DS
    CAPS -->|"generated typed-unsupported shell"| SHELL["Postgres / MySQL / SQLite<br/>operation X not supported"]

    DOLT -. "conformance-tested: SQL family matches Dolt byte-for-byte" .-> KIT
```

<details>
<summary>Plain-text version of the diagram</summary>

```
                              bd CLI
                                |
                       Storage interface
                                |
      +-------------------------+--------------------------+
      |                                                    |
  CORE  (every backend)                        CAPABILITIES  (Dolt-only)
  issueops -- CRUD, search, deps,              version control:
  ready/claim, ordering, id-mint,              history, branch, remotes, sync
  cycle checks: written ONCE                     |
      |                                          +-> Dolt store: implemented
      +--------------------+                     +-> Postgres/MySQL/SQLite:
      |                    |                          generated typed-"unsupported"
  Dolt store         sqlkit.Store                     (operation "X" not supported)
  runs the core      SQL-family adapter
  natively           (Dialect/Readiness/Claim)
      |                    |
 [Dolt engine]      +------+------+---------+
 embedded/server    pgdialect mysqldialect sqlitedialect
 = conformance          |         |            |
    ORACLE          Postgres    MySQL       SQLite

  issueops is shared by EVERY backend (Dolt included); each backend is a store
  adapter over it. Dolt runs the core's SQL natively and adds version control;
  the SQL family runs it through one thin dialect per engine, and is
  conformance-tested to match the Dolt oracle byte-for-byte.
```

</details>

The backend contract is `storage.DoltStorage` = the core `Storage` interface plus capability sub-interfaces (`VersionControl`, `HistoryViewer`, `RemoteStore`, `SyncStore`, `FederationStore`, `CompactionStore`, …). The SQL backends implement the entire core through the shared `sqlkit.Store`; the 44 Dolt-only capability methods come from a **generated** typed-unsupported shell (`unsupported_gen.go`, `go generate`, never hand-stubbed), each returning `*storage.ErrUnsupported{Op, Backend}`. Three interlocking guards keep the boundary honest: the generator's strict skip-list validation, the two-directional completeness test, and the reflective behavioral contract.

Two deliberate nuances: the `Commit*` family is a **no-op, not unsupported**, on SQL backends — every write already committed its own SQL transaction, so opportunistic `store.Commit()` calls must not error. And the runtime seam mirrors the test seam: non-Dolt backends implement the `storage.NonCommitGraphBackend` marker, which `cmd/bd`'s post-run hook checks to skip the entire Dolt-only maintenance tail (auto-commit/export/backup/push). A store *without* the marker is assumed to be Dolt, so the default path is provably untouched.

## Not in scope / follow-ups

- **Stale help text.** The user-facing help is refreshed in this PR (`bd init` long help + the `--backend` flag now list postgres/mysql/sqlite; the sqlite "deprecated" wording is gone). A couple of internal comments and `docs/CLI_REFERENCE.md` still mention the old single-backend world — a minor sweep before undraft.
- **Backend docs.** `docs/` needs a POSTGRES/MYSQL/SQLITE page (in `docs/DOLT.md`'s voice) covering setup and the credential ladder — currently the ladder is documented only in code comments. The storage-backends guide in this PR is the start.
- **Per-connection token refresh for long-lived daemons.** The credential-command cache re-runs the helper shortly before expiry on *open*; a daemon holding a pooled connection past token expiry may need per-connect refresh (pgx `BeforeConnect`-style).
- **Credentials-file IPv6 keying.** `HostPort` returns IPv6 hosts unbracketed, so an IPv6 endpoint never matches a file section. Known sharp edge, noted in code.
- **History emulation on SQL backends** — explicitly not a goal. "History absent" is the design, not a gap; Dolt remains the backend for version control.
- **Workspace-trust gate** that would allow persisting credential commands in metadata.

## Verification

```bash
# Build + full test suite (applies the gms_pure_go tag automatically; Go per go.mod)
make build && make test

# Both conformance tiers. Dolt-embedded + SQLite need no env;
# Postgres/MySQL profiles self-skip unless the test URLs are set.
docker run -d -p 5432:5432 -e POSTGRES_USER=bts -e POSTGRES_PASSWORD=bts -e POSTGRES_DB=bdwedge postgres:16
docker run -d -p 3306:3306 -e MYSQL_ROOT_PASSWORD=bts mysql:8   # then create user 'bts'@'%' as in the workflow
BEADS_PG_TEST_URL="postgres://bts:bts@127.0.0.1:5432/bdwedge" \
BEADS_MYSQL_TEST_URL="bts:bts@tcp(127.0.0.1:3306)/" \
./scripts/conformance.sh

# Individual gates
go test -tags gms_pure_go ./internal/storage/postgres/ ./internal/storage/mysql/ ./internal/storage/sqlite/ \
  -run 'TestInterfaceCompleteness|TestUnsupportedContract'
CGO_ENABLED=1 go test -tags "gms_pure_go e2e" ./test/conformance/    # Tier 2 needs BOTH tags

# Hands-on smoke (run outside any existing beads workspace)
go build -tags gms_pure_go -o bd ./cmd/bd
./bd init --backend=sqlite --prefix demo
./bd create "first issue" -p 1 && ./bd ready
./bd history <id>     # -> typed: operation "History" not supported by the sqlite backend

# Optional deep gate (~50 min, excluded from per-PR CI)
CONFORMANCE_DEEP=1 BTS_RS_DIR=/path/to/oracle ./scripts/conformance.sh
```

CI runs the same script with real `postgres:16` / `mysql:8` service containers on every PR (`.github/workflows/conformance.yml`).

---

**Draft status:** marked DO NOT MERGE pending the stale-help-text sweep, the backend docs pages, and maintainer review of the capability-shell allowlist.


maphew:
@julianknutsen also see proposal at #4561 from @duncan4123

duncan4123:
> @julianknutsen also see proposal at #4561 from @duncan4123

I built this branch on top of my plugin idea and ran the tests/oracle tests (which are great) for the plugins. I also added doltlite to the oracle tests, founds some bugs in the doltlite and postgres plugins I made.
I havent tried running postgres or sqlite from this branch yet. Will do so today.

duncan4123:
I think this PR gives us the right foundation for DoltLite too, but probably not by making DoltLite another built-in backend in the main `bd` release.

The key distinction is build/linking:

- Postgres and MySQL use normal Go drivers and connect to an external server.
- The SQLite backend in this PR uses the pure-Go SQLite path.
- DoltLite currently needs the process that opens the database to link/load `libdoltlite`.

Because of that, I do not think the default `bd` binary should be released linked against DoltLite. That would make the main Beads release CGO/native-library/platform-sensitive for one optional backend. The plugin boundary is a better fit:

```text
bd
  normal portable Beads binary
  speaks backend plugin protocol

bd-backend-doltlite
  separate plugin binary
  links/loads libdoltlite
  opens the DoltLite store
```

That said, the DoltLite plugin should not have to carry a copied/forked Beads storage implementation forever. It should be able to conform to the SQL backend shape introduced here.

The shape I would like to get to is:

```text
doltlite plugin process
  backend protocol server
  doltliteDialect
    Open() -> *sql.DB using sqlite3_doltlite / libdoltlite
    SQLite-compatible SQL translation
    DoltLite-specific busy timeout, connection, retry, and lock policy
  DoltLiteStore
    embeds/reuses sqlkit.Store for portable Beads behavior
    overrides/adds versioned operations:
      Commit / CommitWithConfig
      branches
      remotes
      history/diff/as-of
      sync/fetch/push
      flatten/gc/maintenance
```

So DoltLite would reuse the same core semantics as the new SQL backends for CRUD, deps, ready work, claims, labels, comments, events, config, slots, etc., while keeping DoltLite-specific version-control behavior in the plugin.

The main thing this PR could do to enable that cleanly is separate the backend implementor kit from the built-in backends. Today the useful pieces appear to live under `internal/storage/...`, which means an external plugin cannot import them directly without being built inside the Beads module or vendoring/copying code. It would be useful to expose supported packages along these lines:

```text
backend/plugin        wire protocol types
backend/sqlkit        reusable SQL-family storage adapter
backend/sqlitekit     SQLite translation/dialect helpers
backend/conformance   reusable backend/plugin conformance harness
```

Then a DoltLite plugin can stay out-of-process, keep the native DoltLite dependency isolated, and still run the exact same SQL/storage behavior and conformance/oracle expectations as SQLite/Postgres.

One capability detail is worth making explicit: DoltLite's concurrency model is not the same as Postgres. It is embedded/file-backed, so the plugin will likely serialize writes, use a busy timeout, retry known lock/conflict errors, and possibly hold a file lock around write transactions. That is still a valid backend, but it should not advertise the same capability as a pooled server backend. A capability like this would make the distinction clearer:

```text
writer_model = "server_concurrent" | "embedded_serialized" | "single_writer"
```

or similar.

So the proposal is:

1. Keep DoltLite out of the main `bd` binary by default.
2. Keep DoltLite as a backend plugin binary that owns `libdoltlite`.
3. Expose the SQL/sqlite backend kit from this PR so plugins can reuse it instead of copying storage logic.
4. Add/keep conformance profiles for plugin-backed DoltLite so it is tested against the same Dolt reference/oracle behavior for portable Beads operations.
5. Model concurrency and versioning as explicit backend capabilities rather than hardcoding those assumptions into `bd`.

That would let DoltLite fit into this backend architecture without forcing every Beads user onto a native DoltLite build.

PR #4602, opened 2026-07-06
gate: add multi-machine ordering to remote-migrate gate messages (#4259)

rjc123:
Follow-up to #4259 (I'm the reporter).

## Motivation

Upgrading beads across two machines, the point where I had to stop and work out what to do was the remote-migrate gate message itself. It tells you how to unblock the machine in front of you — migrate as the designated migrator, or `bd bootstrap` — but not the safe **end-to-end multi-machine ordering**, so it isn't obvious what the overall sequence should be.

## Changes

1. **Short multi-machine ordering added to the human-readable gate messages** (`internal/storage/schema/remote_migrate_gate.go` and `internal/storage/embeddeddolt/store.go`): only one clone migrates; the others sync on their current binary and pull the migrated schema after the migrator pushes, before upgrading. No links — just the instruction. (The JSON/agent error already links the full doc; these plain-text messages didn't carry the ordering at all.)

2. **Docs: a non-destructive adoption path for in-sync clones** (`website/docs/getting-started/upgrading.md`): the shipped recipe has every non-migrator clone `bd bootstrap` (a full re-clone). A clone that is fully in sync can instead fast-forward onto the migrated schema by pulling on its current binary before upgrading — no re-clone, local Dolt history preserved.

## ⚠️ Please validate change 2 before merging

Change 2 is **not verified against the v1.1.0 gate**, and it differs from the current "every other clone adopts, does not pull" guidance. It's based on a **pre-v1.1.0** two-clone/one-remote observation where a clean trailing clone adopted the migrated schema via a plain `bd dolt pull` on the old binary (after the migrator pushed), and the subsequent upgrade found nothing to migrate.

Treat it as a question, not a claim:

- Does an old binary cleanly **fast-forward** a PK-reshaping migration when the clone has no un-synced local commits?
- Does the subsequent upgrade then correctly see **zero pending migrations** (gate stays silent)?

If yes, it saves a re-clone. If there's a reason it isn't safe (e.g. an old binary must not pull a schema it doesn't understand), I'll drop change 2 and keep only change 1 (the message ordering). Happy to add a two-binary sandbox repro if that would help confirm it.

### DIFF
--- a/internal/storage/embeddeddolt/store.go
+++ b/internal/storage/embeddeddolt/store.go
@@ -323,7 +323,9 @@ func (s *EmbeddedDoltStore) initSchema(ctx context.Context) error {
 				"  you are the single designated migrator (only ONE clone may migrate a\n" +
 				"  shared remote, else the schema forks; #4259):\n" +
 				"    • designated migrator (only ONE machine): %[3]s=1 bd migrate && bd dolt push\n" +
-				"    • every other clone (another already migrated): bd bootstrap\n"
+				"    • every other clone (another already migrated): bd bootstrap\n" +
+				"    • several machines: only ONE migrates; sync each other clone and run\n" +
+				"      bd dolt pull after the migrator pushes, before upgrading it\n"
 			switch s.intent {
 			case openWorkingSetReconcile:
 				fmt.Fprintf(os.Stderr,
--- a/internal/storage/schema/remote_migrate_gate.go
+++ b/internal/storage/schema/remote_migrate_gate.go
@@ -192,7 +192,12 @@ func (e *RemoteMigrateGateError) userBody() string {
 			"        bd bootstrap\n" +
 			"      Re-cloning replaces your local database: any local issues you have not\n" +
 			"      pushed are LOST. Push first (`bd dolt push`) or save a copy\n" +
-			"      (`bd export --all -o backup.jsonl`) before re-cloning.\n"
+			"      (`bd export --all -o backup.jsonl`) before re-cloning.\n" +
+			"\n" +
+			"  Several machines on one remote? Only ONE migrates. Get every other clone\n" +
+			"  fully in sync on its current binary, then after the migrator pushes run\n" +
+			"  `bd dolt pull` on each before upgrading it — its upgrade then has nothing\n" +
+			"  to migrate and needs no re-clone.\n"
 	}
 }
 


maphew:
Thanks for this - the multi-machine ordering gap in the plain-text gate messages is a real, well-scoped fix, and it's good to see it come from the person who filed #4259 in the first place.

**Change 1 (gate message ordering hints, `store.go` / `remote_migrate_gate.go`)**: this is clean and low-risk. I checked both insertion points against current `main` (post-#4515/#4524/#4551): the surrounding text in each case is the decision-agnostic default/shared guidance block, not the `adopt`- or `fork-skew`-specific branches, so the new bullet doesn't contradict or duplicate anything the smart gate already says for those decisions - it just fills in ordering context that, as you note, only the JSON/agent path's linked doc previously carried. No objection to this half.

**Change 2 (the "Adopting without a re-clone" docs section)**: thank you for flagging this as unverified rather than asserting it - that's the right instinct. Two things worth tightening, neither of them a safety hazard (I checked: the recipe runs the upgrade in a fresh process, so the remote-migrate gate re-evaluates on the next open either way - no un-gated migration path is opened):

- The closing promise ("there are no pending migrations and the remote-migrate gate stays silent") is only guaranteed in the single-hop, whole-fleet-on-one-new-binary case this section sits under. If the migrator's push landed the remote *below* the new binary's own latest supported version (several migration releases queued, or an adopter installing a newer binary than the migrator used), the adopter's first open still has pending migrations, and the gate re-engages: either the #4516 smart gate's sanctioned auto-migrate path (with an stderr notice) or a block with migrate-or-adopt guidance. Suggested fix: scope the prose to the single-hop recipe explicitly, or add a one-line caveat so readers aren't surprised when the gate speaks up.
- There's an open PR (#4586) automating exactly this idea (detect a fully-in-sync clone, fast-forward it instead of re-cloning), with auto-execution already pushed on that branch. Once it lands, the gate itself does this dance with a stricter TOCTOU-checked guard, and this manual recipe likely needs a rewrite or trimming shortly after.

Recommendation: land change 1 now - it's ready. For change 2, either add the single-hop caveat and land it as interim guidance, or hold it until #4586 settles and fold the automated behavior into the docs instead. I'd lean toward the first if you want this available now, the second if you'd rather not document a workflow about to be automated. Thanks again for closing the loop on your own original report.

rjgeng:
Change 1 is clearly the right fix. The gate message gap is real from an operator perspective — when you're mid-upgrade across two machines the message tells you what to do on *this* clone but not the safe ordering across clones, so you have to go find the docs under pressure. Embedding the one-line sequence removes that lookup.

On Change 2: the reasoning is sound for the stated precondition. A trailing clone with no local diverging commits has nothing to 3-way merge — it's a pure fast-forward onto the migrated schema, same as any Dolt `dolt pull` where the local branch hasn't diverged. The old binary doesn't need to *understand* the new schema to move the ref pointer forward; it just receives the commit and advances HEAD. The question of whether the subsequent `bd upgrade` then correctly detects zero pending migrations is the gating uncertainty, and I haven't tested that specific interaction against the v1.1.0 gate.

If the gate logic reads the current schema version from the Dolt commit rather than re-deriving it by diffing table definitions, it should see the migrated state and stay silent. If it re-derives, it could misfire. Worth a targeted test with `bd upgrade --dry-run` (if that flag exists) before merging change 2.

maphew:
@rjgeng - good question, and it has a clean answer in the source: the gate does not re-derive schema state from table definitions. Pending detection reads the version cursor directly - `SELECT COALESCE(MAX(version), 0) FROM schema_migrations` (`internal/storage/schema/schema.go`), and `CheckRemoteMigrateGate` builds on the same `CurrentVersion` helper. Since the fast-forward pull delivers the migrated `schema_migrations` rows along with everything else, the next open sees max(version) == the binary's latest and the gate stays genuinely silent in the single-hop case. No misfire path from re-derivation.

On the targeted test: there's no `bd upgrade` command (upgrades are just installing a new binary; migration runs at store open), but `bd migrate --dry-run` exists and shows what would be applied without touching anything - that plus any read command after the pull is the check you described.

Status update that changes my earlier recommendation: **#4586 merged this morning** (ce3c040), so the gate itself now detects a fully-in-sync clone and fast-forwards it automatically, with a TOCTOU-checked guard. That makes the manual recipe in change 2 documentation of a workaround the tool no longer needs. Suggested path for this PR:

- Keep change 1 exactly as-is (the message ordering hints) - still correct post-#4586, still wanted.
- Replace the change 2 section with a short paragraph describing the new automatic behavior (trailing in-sync clones are fast-forwarded by the gate on next open; only diverged clones need the adopt/re-clone decision), rather than the manual `dolt pull` recipe.

@rjc123 if you'd rather land change 1 now and leave the docs paragraph to a follow-up, I'm happy to take it either way - just say which you prefer and I'll get it moving.

rjc123:
Thanks @maphew — appreciate the careful review, and the source-level confirmation on the version-cursor detection (that also settles @rjgeng's question cleanly). Agreed with the plan: I've amended this down to change 1 (the gate message ordering) only. Since #4586 now fast-forwards in-sync clones automatically, change 2's manual recipe documents a workaround the tool no longer needs, so I've dropped it rather than reworking it — happy to leave a docs paragraph describing the new automatic behaviour to a follow-up. Thanks again for all the work on beads; it's become a core part of my workflow.

PR #4603, opened 2026-07-06
feat(beads-mcp): add comment, comments, and note MCP tools

banozz0:
## What

Adds three MCP tools to `beads-mcp`, each wrapping an existing `bd` CLI command:

| Tool | Wraps | Purpose |
|------|-------|---------|
| `comment` | `bd comment <id> <text>` | Add a durable, timestamped comment |
| `comments` | `bd comments <id> --json` | Read an issue's comments back |
| `note` | `bd note <id> <text>` | Append to the issue's `notes` field |

## Why

The MCP server currently exposes no way to add or read comments. MCP-only agents can therefore only overwrite the single `notes` field via `update`, and `show()` returns `comment_count` but not the comment bodies.

In multi-agent setups this means agents that drive the board purely through the MCP leave no readable per-turn trail of what they did — a human (or another agent) has to read the raw transcript or dig through files to find out. The `bd` CLI already supports `comment` / `comments` / `note`; this just surfaces them through the MCP so agents that don't shell out to the CLI reach parity.

## How

- `comment` / `note` reuse a small `_run_text_command` helper, since those `bd` subcommands print a confirmation line rather than JSON (mirrors the existing `add_dependency` approach).
- `comments` parses `bd comments <id> --json` into a new `Comment` model (`id`, `issue_id`, `author`, `text`, `created_at`).
- New param models: `AddCommentParams`, `ListCommentsParams`, `AddNoteParams`.
- Registered in `server.py` with the same `@with_workspace` / `@require_context` decorators as the other write tools, and added to `_TOOL_CATALOG` / `get_tool_info`.

Purely additive — no existing behavior changes.

## Testing

- `ruff check` clean.
- Added unit tests for `add_comment`, `add_note`, and `list_comments` (success, failure, not-found, invalid-response) mirroring the existing `add_dependency` tests. `pytest tests/test_bd_client.py tests/test_tools.py tests/test_config.py` → 87 passed.
- Manual end-to-end against a real `bd` database: `comment` / `note` / `comments` round-trip correctly, and the `Comment` model parses real `bd comments --json` output.

### DIFF
--- a/integrations/beads-mcp/CHANGELOG.md
+++ b/integrations/beads-mcp/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- New `comment`, `comments`, and `note` MCP tools, wrapping the existing
+  `bd comment` / `bd comments` / `bd note` CLI commands. Previously the MCP
+  server exposed no way to add or read comments, so MCP-only agents could only
+  overwrite the single `notes` field — `show()` reports `comment_count` but not
+  the bodies. `comment` leaves a durable, timestamped work trail, `comments`
+  reads them back, and `note` appends to the notes field.
+
 ### Fixed
 - `mcp__beads__show()` no longer crashes with `pydantic literal_error` when an
   issue has dependency types beyond the original four (`blocks`, `related`,
--- a/integrations/beads-mcp/README.md
+++ b/integrations/beads-mcp/README.md
@@ -218,6 +218,9 @@ await beads_ready_work(workspace_root="/Users/you/project-a")
 - `update` - Update issue (status, priority, design, notes, etc). Note: `status="closed"` or `status="open"` automatically route to `close` or `reopen` tools to respect approval workflows
 - `close` - Close completed issue
 - `dep` - Add dependency (blocks, related, parent-child, discovered-from)
+- `comment` - Add a durable, timestamped comment to an issue (a record of work/decisions)
+- `comments` - List all comments on an issue (`show` reports comment_count but not the bodies)
+- `note` - Append a note to an issue's notes field
 - `blocked` - Get blocked issues
 - `stats` - Get project statistics
 - `reopen` - Reopen a closed issue with optional reason
--- a/integrations/beads-mcp/src/beads_mcp/bd_client.py
+++ b/integrations/beads-mcp/src/beads_mcp/bd_client.py
@@ -10,14 +10,18 @@ from typing import Any
 
 from .config import load_config
 from .models import (
+    AddCommentParams,
     AddDependencyParams,
+    AddNoteParams,
     BlockedIssue,
     BlockedParams,
     ClaimIssueParams,
     CloseIssueParams,
+    Comment,
     CreateIssueParams,
     InitParams,
     Issue,
+    ListCommentsParams,
     ListIssuesParams,
     ReadyWorkParams,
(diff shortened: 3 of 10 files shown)


maphew:
Thanks for this — real value here. MCP-only agents currently have no way to leave a readable trail (only the single `notes` field, and `show()` reports `comment_count` without the bodies), and this closes that gap cleanly by wrapping the existing `bd comment` / `bd comments --json` / `bd note` commands rather than inventing anything new. I checked the argument construction, the `_run_command` auto-`--json` behavior, and the `Comment` model fields against the actual Go CLI source (`cmd/bd/comment.go`, `comments.go`, `note.go`) and they line up correctly. The read-vs-write `@with_workspace`/`@require_context` decorator split on `comments` vs `comment`/`note` also matches the existing pattern exactly.

One gap worth closing before merge: test coverage stops one layer short of the tool it says it mirrors. `tests/test_bd_client.py` gets good unit coverage for the three new methods (six tests, including error and invalid-response paths), but:

- `tests/test_tools.py` has no tests for `beads_add_comment`, `beads_list_comments`, or `beads_add_note`, while the analogous `beads_add_dependency` has a wrapper-level success/error pair there (lines ~252/267). To be fair, a handful of other existing tools also lack `test_tools.py` coverage, so this isn't breaking a hard convention — but `add_dependency`'s wrapper also has an error-handling branch these new passthrough wrappers lack, which is exactly what a wrapper-level test would surface.
- Neither `tests/test_bd_client_integration.py` nor `tests/test_mcp_server_integration.py` gained a real-`bd`/real-MCP-client test for the new tools, while `add_dependency` has one in each.

Given the description says testing "mirrors the existing add_dependency approach," it'd be great to actually mirror it at all three tiers, not just the mocked-subprocess one. Also worth noting: there's no PR-triggered CI for `integrations/beads-mcp` (only a manual test-pypi workflow), so the `ruff check` / `pytest ... → 87 passed` claims in the description haven't been independently verified — please re-run locally before merge.

Small non-blocking nit: the new `_run_text_command` helper generalizes exactly the subprocess boilerplate that `add_dependency` still has inlined separately (`bd_client.py` ~line 633) — could fold that into a call to the new helper as a follow-up cleanup, no need to hold this PR for it.

Recommend: add the missing `test_tools.py` (and ideally integration) coverage for the three new tools, confirm tests/ruff pass locally, then merge.

maphew:
Following up on the review above to get this landed today: I pushed the missing test coverage directly to this branch (378f6d46c, maintainer edits are enabled), completing the three-tier mirror of `add_dependency` that the description cites:

- `tests/test_tools.py`: success + error tests for `beads_add_comment`, `beads_list_comments`, and `beads_add_note`. The error tests document that `BdError` propagates out of these passthrough wrappers (they intentionally have no catch-and-stringify branch like `add_dependency`'s - propagation matches the majority of existing write tools).
- `tests/test_bd_client_integration.py`: real-bd round-trip - two comments added and read back in chronological order, note appended and verified via `show`.
- `tests/test_mcp_server_integration.py`: end-to-end MCP-client exercise of the `comment`, `comments`, and `note` tools.

Local verification per the review's ask: `pytest tests/test_bd_client.py tests/test_tools.py tests/test_config.py` -> 93 passed (87 before), `ruff check` and `ruff format` clean, and both integration modules collect cleanly. The integration tiers need a `bd` binary and run under the manual workflow as before; they mirror the adjacent `add_dependency` tests structurally.

Merging now. Thanks @banozz0 - this closes a real gap for MCP-only agents, and the wrapping is faithful to the CLI.

PR #4599, opened 2026-07-06
fix(prune): speed up reference scan

maphew:
## Why

Upstream `main` is failing the `Main` workflow and nightly full tests after #4023. Both failures share the same root cause:

- `TestPruneLargeFixture`
- `buildReferencedSet` takes about 15-16s
- the test requires the 10K-bead fixture to complete in under 5s

The reference-aware prune path compiled all candidate IDs into one large regexp alternation and applied it to every open bead body. On the large fixture that made the protection scan too slow for CI.

## What Changed

- Replaced the regexp alternation with a linear literal matcher grouped by first byte.
- Kept the existing word-boundary behavior for candidate ID references.
- Added a matcher boundary test for punctuation, child IDs, and embedded-word false positives.

## Verification

```text
CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestCandidateIDMatcherWordBoundaries|TestPruneLargeFixture' -count=1 -v
CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'Test.*Prune|TestCandidateIDMatcher' -count=1 -v
CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^$'
```

`TestPruneLargeFixture` now completes locally in about 0.46s.

PR #4610, opened 2026-07-06
feat(prune): prune and purge proxied-server support

coffeegoddd:
## What

Adds proxied-server support for `bd prune` (and, since they share the code path, `bd purge`) by routing the delete-closed-beads flow through a `UnitOfWork` and the domain use cases instead of the embedded store.

## How

- **Domain (`DeleteIssuesParams` gains `Cascade`)**: `deleteMany` previously always expanded to dependents via `FindAllDependents` (cascade-always). It now only expands when `params.Cascade` is set; otherwise it operates on the exact `IDs`. The `DeleteIssue`/`DeleteWisp` convenience wrappers and the existing `bd delete` proxied callers set `Cascade: true` to preserve their current always-cascade contract. Prune sets `Cascade: false` to delete exactly the matched closed beads — matching the embedded store's `cascade=false, force=true` semantics.
- **Dispatch**: `runPurgeOrPrune` forks to a new `runPurgeOrPruneProxied` when `usesProxiedServer()`.
- **New handler (`purge_proxied_server.go`)**: mirrors the embedded flow over the UOW — `IssueUseCase().SearchIssues` for candidates, `ConfigUseCase().GetCustomStatuses` + batched `CommentUseCase().GetCommentsForIssues` for the reference-aware skip scan, and `IssueUseCase().DeleteIssues(Cascade:false)` for dry-run counts and the real delete, followed by `uw.Commit`. Reuses the existing pure helpers (`parseHumanDuration`, `filterClosedDeletionCandidates`, safety-skip warnings, output formatting).

No new domain use-case methods were needed — `SearchIssues`, `GetCustomStatuses`, and `GetCommentsForIssue(s)` already existed; only `DeleteIssuesParams` was extended.

## Testing

`prune_proxied_integration_test.go` (runs with `BEADS_TEST_PROXIED_SERVER=1`, gated behind the existing `bd init --proxied-server` guard like the other proxied integration tests): requires-filter, dry-run no-op, force-deletes-closed, skips-open, reference-aware-skip, `--ignore-references`, JSON output, and dolt-commit — all green. Domain delete unit tests and embedded `prune`/`purge`/`delete` tests pass.

### DIFF
--- a/cmd/bd/delete_proxied_server.go
+++ b/cmd/bd/delete_proxied_server.go
@@ -78,6 +78,7 @@ func runDeleteProxiedServer(cmd *cobra.Command, ctx context.Context, args []stri
 
 	res, err := issueUC.DeleteIssues(ctx, domain.DeleteIssuesParams{
 		IDs:                  in.ids,
+		Cascade:              true,
 		UpdateTextReferences: true,
 	}, actor)
 	if err != nil {
@@ -105,8 +106,9 @@ func runDeleteProxiedPreview(ctx context.Context, issueUC domain.IssueUseCase, i
 	}
 
 	res, err := issueUC.DeleteIssues(ctx, domain.DeleteIssuesParams{
-		IDs:    in.ids,
-		DryRun: true,
+		IDs:     in.ids,
+		Cascade: true,
+		DryRun:  true,
 	}, actor)
 	if err != nil {
 		FatalErrorRespectJSON("preview counts: %v", err)
--- a/cmd/bd/prune_proxied_integration_test.go
+++ b/cmd/bd/prune_proxied_integration_test.go
@@ -1,0 +1,165 @@
+//go:build cgo
+
+package main
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+func issueRowExists(t *testing.T, db *sql.DB, id string) bool {
+	t.Helper()
+	var n int
+	if err := db.QueryRowContext(context.Background(),
+		"SELECT COUNT(*) FROM issues WHERE id = ?", id).Scan(&n); err != nil {
+		t.Fatalf("count issue %s: %v", id, err)
+	}
+	return n > 0
+}
+
+func bdProxiedPrune(t *testing.T, bd, dir string, args ...string) string {
+	t.Helper()
+	fullArgs := append([]string{"prune"}, args...)
+	stdout, stderr, err := bdProxiedRunBuffers(t, bd, dir, fullArgs...)
+	if err != nil {
+		t.Fatalf("bd prune %s failed: %v\nstdout:\n%s\nstderr:\n%s",
+			strings.Join(args, " "), err, stdout, stderr)
+	}
+	return stdout
+}
+
+func bdProxiedPruneFail(t *testing.T, bd, dir string, args ...string) string {
+	t.Helper()
+	fullArgs := append([]string{"prune"}, args...)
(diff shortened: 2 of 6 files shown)


maphew:
Thanks Dustin - this closes a real gap (prune/purge simply didn't work against a proxied server) and it does so the right way: through the UOW and existing domain use cases, with only a one-field widening of `DeleteIssuesParams` rather than any new storage-side surface. The `Cascade` refactor is careful - I checked every non-test `DeleteIssuesParams` constructor on main and all of them are updated to preserve the always-cascade contract, and `Cascade: false` for prune matches the embedded `cascade=false, force=true` orphan semantics. The integration test coverage (dry-run, force, reference protection, `--ignore-references`, JSON, dolt commit) is exactly what this path needed. CI is green across the board.

Two things before/around merge:

1. **Coordinate with #4599 on the reference scan.** `buildReferencedSetProxied` (cmd/bd/purge_proxied_server.go:160) copies the embedded regexp-alternation scan that #4599 replaces with a linear matcher - that alternation is the ~15s-on-10K-beads scan behind the `TestPruneLargeFixture` CI failures. As written, proxied mode reintroduces that profile, and `TestPruneLargeFixture` only guards the embedded helper. Whichever PR lands second, the proxied path should reuse the shared matcher - happy to have that be a fast follow-up if you'd rather land this first.

2. **JSON parity: `referenced_ids_sample` (Copilot caught this too).** `emitProxiedPruneResult` (cmd/bd/purge_proxied_server.go:318) drops `referencedSample`, so a proxied `bd prune --json --force` omits `referenced_ids_sample` where the embedded path emits it (purge.go final-result block). Your empty and dry-run emitters already include it, so it looks like an oversight rather than a choice - passing the sample through is a ~5-line fix.

One observation, no action needed unless you disagree: the proxied dolt commit reads `bd: prune N bead(s)` while embedded prune's commit (via `DoltStore.DeleteIssues`) reads `bd: delete N issue(s)`. Yours is the more honest message; just flagging that history now differs by mode.

Recommendation: fix the `referenced_ids_sample` parity here (small), land, and follow up on the shared matcher once #4599 merges. Nice work - this is the kind of change that moves the storage-seam migration forward.

maphew:
Following up on the review above: I pushed the small JSON-parity fix directly to this branch (maintainer edits are enabled) so this can land today - commit 3d0910be7.

What it does:

- `emitProxiedPruneResult` now takes and emits `referencedSample`, so proxied `bd prune --json --force` includes `referenced_ids_sample` exactly like the embedded path and your own empty/dry-run emitters.
- Extended the `json_output` integration subtest to cover a referenced skip (`referenced_skipped` and `referenced_ids_sample` assertions), so the parity has a test guarding it.

Local verification: with the `bd init --proxied-server` "not yet implemented" guard temporarily lifted (uncommitted, same dormant-suite situation as the rest of the proxied integration tests), `TestProxiedServerPrune` passes across all 8 subtests under `BEADS_TEST_PROXIED_SERVER=1` with a real dolt binary. Build, `go vet`, and `gofmt` clean with `CGO_ENABLED=0 -tags gms_pure_go`.

The #4599 matcher coordination will be handled on the #4599 side after this merges: its linear matcher will be extended to `buildReferencedSetProxied` so the regexp-alternation profile doesn't come back through the proxied path.

Merging once CI is green on the new head. Thanks again Dustin - this closes a real gap the right way.

maphew:
Follow-up delivered: #4599 is merged (428a9db) and includes the coordination commit switching `buildReferencedSetProxied` to the shared linear `candidateIDMatcher`, so the proxied path no longer carries the regexp-alternation scan profile. Both halves of the coordination are now on main.

PR #4611, opened 2026-07-07
finish #4566: self-heal the migrate/bootstrap path, commit each migration atomically

csauer02-personal-user:
## Summary

`#4567` made the **hand-commit recovery path** (`OpenForWorkingSetReconcile` + `bd dolt commit`/`bd vc commit`) able to open a store whose pending migration touches a dirty table. It did **not** fix the **automatic** path — and its own scope note says so ("`bd migrate`/`bd bootstrap` still hit the guard strictly; server mode untouched").

`MigrateUp` applies every pending `ALTER` + its `schema_migrations` row to the Dolt working set, then runs a **long fallible tail** (backfills, dependency/aux rekeys, `dolt_ignore`) before a **single terminal `DOLT_COMMIT`**. A crash / kill / timeout anywhere in that tail leaves the applied migrations **uncommitted**; Dolt persists the working set across restarts. On retry, `pendingMigrationDirtyTables` + `DirtyTablesError` refuse — the guard cannot distinguish the migration's **own** half-finished debris from user working-set changes. A managed supervisor re-invokes the migrate path every attempt, re-dirties, and **never converges** without a human stopping it to hand-commit.

## Fix — commit each migration atomically

Commit each numbered migration **as it applies** (`commitMigrationStep`): snapshot working-set dirtiness before the migration's `ExecContext`, then after apply + cursor-`INSERT`, force-stage only the tables this migration newly dirtied (plus the cursor table) and `DOLT_COMMIT`. A crash/kill after any step then leaves a **clean** working set at the last committed version, so a plain retry resumes at the next version on a clean set and the dirty-table guard can never trip on the migration's own debris.

This is data-safe because `MigrateUp`'s pre-flight guard already proved no migration-touched table holds uncommitted **user** rows before `migrate()` runs; selective staging never sweeps an untouched user-dirty table into the commit.

Scoping: gated to the **production main-source path** (`upTo==0`, main cursor). `MigrateUpTo` (test-support) and the dolt-ignored source keep the single terminal commit, preserving their semantics and the "ignored tables never enter shared history" invariant.

## Consolidation (one band-aid deleted, one honestly retained)

- **Deleted** `failed0053DirtyTablesAreRecoverable` + its caller + use-site + its 5 unit tests. It is genuinely dead post-fix: the multi-migration `v49→v53` batch it recovered can no longer strand `0050/0051/0052` uncommitted (each commits per-step).
- **Retained** `auxRekeyResumePending` — verified **not** dead. It guards a *different* crash site: the aux-rekey runs in the tail under the terminal commit, which per-step commit doesn't touch. On a legacy DB whose aux-rekey crashed mid-pass, deleting it would silently regress convergence (the resumed rekey's `events/comments` writes would trip `changedDirtyTableSignatures` + `stageSchemaTables`). Left intact.

## Tests / proof (`BEADS_TEST_EMBEDDED_DOLT=1`)

New `internal/storage/embeddeddolt/migrate_selfheal_test.go`, using a nil-in-prod step-boundary fault hook (a clean between-operations kill — the faithful shape of SIGKILL+restart):

- **Repro A** — seed real schema below a multi-step jump, kill right after a migration applies, plain retry → converges to latest + clean working set, no hand-commit.
- **Repro B** (the race #4567 never tested) — 4 concurrent supervisor-retry loops with a shared fault hook killing ~55% of steps → every worker converges to latest+clean, and **no killed attempt ever leaves a dirty working set** (the per-step-commit invariant).
- **Unit** — pre-existing user dirt on an untouched table stays dirty after per-step commits (selective-staging safety).

**Pre-fix (per-step commit toggled off) the repros fail** exactly as #4566 describes:
```
retry MigrateUp returned *DirtyTablesError [comments issues] — the migrate path did NOT self-heal
worker 0: killed pass left a DIRTY working set [comments issues schema_migrations] at version 49
```
**Post-fix all pass:**
```
--- PASS: TestEmbeddedMigrateSelfHealsAfterMidPassFault_4566 (1.59s)
--- PASS: TestEmbeddedMigrateConvergesUnderConcurrentInterruptedRetries_4566 (4.04s)
--- PASS: TestEmbeddedMigrateStepCommitLeavesUntouchedUserDirt_4566 (1.66s)
--- PASS: TestEmbeddedDirtyTablesGate_BlocksReopenThenWorkingSetReconcileRecovers  [existing #4566/#4567 test, unregressed]
```
No regression: `go build ./...` + `go vet` clean; `go test ./internal/storage/schema/...` ok (sqlmock tests updated for the per-step `DOLT_ADD`/`COMMIT` sequence); full embeddeddolt store suite ok. (Docker-gated dolt-server tests skip locally but exercise the same `MigrateUp`.)

Refs #4566, #4567.

### DIFF
--- a/internal/storage/embeddeddolt/migrate_selfheal_test.go
+++ b/internal/storage/embeddeddolt/migrate_selfheal_test.go
@@ -31,6 +31,15 @@ import (
 // schema.SetMigrateStepFaultHookForTest, which models a process killed between
 // operations (the working set is whatever the last completed SQL statement
 // left) — the faithful shape of a SIGKILL/timeout that a supervisor restarts.
+//
+// Scope caveat: these repros call schema.MigrateUp directly, below the
+// remote-migrate gate. On a remote-backed store the retry's open path re-runs
+// the #4516 smart gate first, and a killed pass leaves the working-set cursor
+// ahead of the remote, which routes to the undetermined verdict and blocks the
+// unattended retry at RemoteMigrateGateError (a pre-existing gate limitation,
+// not introduced here). "Plain retry converges" therefore holds for local
+// stores and for gate-sanctioned paths, not for an unattended supervisor loop
+// on a remote-backed store.
 
 const selfHealSeedVersion = 48 // migrations 49..LatestVersion() form the killed jump
 
--- a/internal/storage/embeddeddolt/migrate_selfheal_test.go
+++ b/internal/storage/embeddeddolt/migrate_selfheal_test.go
@@ -1,0 +1,375 @@
+//go:build cgo
+
+package embeddeddolt_test
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"math/rand"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/storage/embeddeddolt"
+	"github.com/steveyegge/beads/internal/storage/schema"
+)
+
+// These tests cover the migrate/bootstrap self-heal that finishes
+// gastownhall/beads#4566. #4567 made the hand-commit recovery path
+// (OpenForWorkingSetReconcile + bd dolt commit) able to open a store whose
+// pending migration touches a dirty table. This suite proves the automatic
+// path: when a migration pass is killed mid-flight, a plain retry converges to
+// the latest schema with a clean working set and NO manual commit, because each
+// numbered migration is now committed atomically as it applies.
+//
+// The fault is injected at a migration-step boundary via
+// schema.SetMigrateStepFaultHookForTest, which models a process killed between
+// operations (the working set is whatever the last completed SQL statement
+// left) — the faithful shape of a SIGKILL/timeout that a supervisor restarts.
+
+const selfHealSeedVersion = 48 // migrations 49..LatestVersion() form the killed jump
+
+func requireEmbedded(t *testing.T) {
+	t.Helper()
+	if os.Getenv("BEADS_TEST_EMBEDDED_DOLT") != "1" {
+		t.Skip("set BEADS_TEST_EMBEDDED_DOLT=1 to run embedded dolt tests")
+	}
+}
+
+// TestEmbeddedMigrateSelfHealsAfterMidPassFault_4566 is Repro A: pin a store
+// below a multi-step jump, kill the pass right after a numbered migration
+// applies (before the terminal commit), then prove a plain retry converges with
+// a clean working set and no hand-commit.
+//
+// PRE-fix (per-step commit removed) this same test fails at the retry: MigrateUp
+// returns *schema.DirtyTablesError because the killed pass left migration 49's
+// ALTER (touching `issues`, which pending migrations 53/54 also touch)
+// uncommitted, and the guard refuses to entangle it.
+func TestEmbeddedMigrateSelfHealsAfterMidPassFault_4566(t *testing.T) {
+	requireEmbedded(t)
+	ctx := t.Context()
+	dataDir := seedMainSchemaAt(t, ctx, selfHealSeedVersion)
+
+	// Attempt 1: kill the pass immediately after migration 49 applies.
+	const killAfter = 49
+	restore := schema.SetMigrateStepFaultHookForTest(func(_ context.Context, _ schema.DBConn, version int) error {
+		if version == killAfter {
+			return fmt.Errorf("injected fault: process killed after migration %d", killAfter)
+		}
+		return nil
+	})
+	conn, closeConn := openPinnedConn(t, ctx, dataDir)
+	_, err := schema.MigrateUp(ctx, conn)
+	closeConn()
+	restore()
+	if err == nil || !strings.Contains(err.Error(), "injected fault") {
+		t.Fatalf("attempt 1 MigrateUp err = %v, want the injected fault", err)
+	}
+
+	// Attempt 2 (the supervisor's plain retry): no fault, no hand-commit.
+	conn2, closeConn2 := openPinnedConn(t, ctx, dataDir)
+	defer closeConn2()
+	applied, err := schema.MigrateUp(ctx, conn2)
+	if err != nil {
+		// This is the PRE-fix failure: *schema.DirtyTablesError.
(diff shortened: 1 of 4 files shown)


maphew:
Thanks for this - it is a genuinely strong finish to #4566. #4567 deliberately fixed only the hand-commit recovery path, and this closes the automatic half with the right core idea: commit each numbered migration atomically as it applies, so the fallible backfill/rekey tail can no longer strand a whole pass's migrations for the retry to trip over. Things I verified and want to call out as done well:

- **The #4516 smart-gate is respected structurally.** The gate runs at store open before `MigrateUp` is reachable (embeddeddolt `store.go`, dolt `store.go`), and this PR changes only `MigrateUp`'s interior, so it cannot auto-apply migrations the gate would block.
- **schema_migrations semantics are preserved**: each step commits its DDL together with its `(version, content_hash)` cursor row, which is exactly what the gate's skew comparison reads; the ignored source and `MigrateUpTo` keep the terminal-commit semantics, so dolt-ignored tables still never enter shared history.
- **Convergence is not harmed**: above the floor, pending migrations are deterministic, so per-step table content is byte-identical across concurrent first-movers, and Dolt merges by content (commit graphs already diverged across clones before this change). Retaining `auxRekeyResumePending` was the right call and the reasoning in the description is correct.
- The concurrent interrupted-retry repro (Repro B) and the selective-staging safety test are exactly the tests this change needed.

Four findings before merge, roughly in order of importance:

1. **The `failed0053DirtyTablesAreRecoverable` deletion regresses stores already stranded in the wild.** "Genuinely dead post-fix" holds only for strandings created by binaries containing this fix. A store sitting at v52 with dirty `comments/dependencies/events/issues` from an *old* binary's failed v49->v53 batch (the #4555 incident class) is auto-recovered on current main by this exemption; post-PR it gets `DirtyTablesError` and needs the manual `bd dolt commit` path from #4567. That may be an acceptable trade given the short rc.2 exposure window, but it should be a conscious maintainer decision rather than a deadness claim - either restore the exemption for one more release or note the downgrade explicitly in the PR body.

2. **The "plain supervisor retry converges" claim is oversold for remote-backed stores - the exact #4566 environment.** If the process dies between steps, the retry's open path hits the smart gate with `remoteMax < current`, which routes to `smartUndetermined` and the blunt block, so the retry stalls at `RemoteMigrateGateError` unless `BD_ALLOW_REMOTE_MIGRATE=1` is set. To be fair, this parking predates the PR (the gate reads the working-set `schema_migrations`, which a pre-PR killed pass also advanced uncommitted), so it is a pre-existing gate limitation this PR neither causes nor fixes - and still a clear improvement over the old `DirtyTablesError` deadlock. But both repros call `schema.MigrateUp` directly and bypass the gate, so this path is untested, and the description's supervisor-convergence narrative doesn't hold on remote-backed stores. Please temper the claim; a follow-up issue to teach the smart gate a "resume own partial migration" verdict (no skew, remote strictly behind, at/above floor) would be a good companion.

3. **A residual window remains within a single step**: a SIGKILL between a migration's SQL and its `commitMigrationStep` still leaves debris the guard trips on (the fault hook fires only after the commit, so tests model only the between-steps kill). The window shrinks from the whole tail to one step, which is the real win - just soften "the dirty-table guard can never trip on the migration's own debris" in the doc comment and description.

4. **Fresh init cost**: with `commitEachStep` gated only on `upTo==0` + main source, a brand-new database now per-step commits the whole 1..latest run - roughly 54 root commits and two `dolt_status` scans per step on every `bd init`. Self-heal has value on fresh init too, so maybe this is fine, but consider gating to `current > 0` or note the trade deliberately.

Nit: give the per-step commit message the migration name (`schema: apply migration 0050_...`) - forensically useful in `dolt_log` for exactly the crash scenarios this PR targets.

Housekeeping: CI has not run yet (first-contribution approval); the base is ~28 commits behind main but none of the touched files drifted, so it applies clean. Once CI is approved and items 1-3 are addressed (1 may just be a decision + description edit), I would be glad to land this.

Recommended outcome: **fix-merge**.

maphew:
Following up on the review above to get this landed today. I pushed a maintainer commit to this branch (18be953f8, maintainer edits are enabled) covering the mechanical items, and here are the explicit decisions on the judgment ones:

**Pushed in 18be953f8** (items 3, the nit, and the test-scope half of 2):

- Per-step commit messages now name the migration (`schema: apply migration 0050_...`) instead of the generic `apply migration step` - `dolt_log` forensics for exactly the crash class this PR targets. `lock_test.go` sqlmock expectation updated for the parameterized commit.
- `runMigrations` doc comment softened: the strandable window shrinks from the whole pass tail to a single step (a kill between a migration's SQL and its `commitMigrationStep` still leaves that step's debris); it does not close entirely.
- The self-heal suite's header now carries a scope caveat: the repros call `schema.MigrateUp` below the remote-migrate gate, and on a remote-backed store an unattended retry parks at the #4516 smart gate's undetermined verdict - a pre-existing gate limitation this PR neither causes nor fixes.

**Item 1 - `failed0053DirtyTablesAreRecoverable` deletion: accepted, consciously.** The exemption's remaining value is auto-recovering stores *already* stranded by an old binary's failed v49-v53 batch (the #4555 incident class). Given the short rc.2 exposure window, the documented manual recovery path from #4567 (`OpenForWorkingSetReconcile` + `bd dolt commit`) remaining available, and the exemption being a targeted band-aid we do not want to carry indefinitely, I am taking the downgrade-to-manual trade rather than restoring it for one more release. This decision will be recorded in the merge commit body, not buried.

**Item 2 follow-up filed:** #4628 proposes the "resume own partial migration" gate verdict (no skew, remote strictly behind, at/above floor) so killed remote-backed migrations converge unattended too.

**Item 4 - fresh-init per-step commits: accepted as-is.** Self-heal has value on `bd init` too (a killed first init is exactly the confusing case), and ~54 root commits in a brand-new database's history is a cosmetic cost. If it bothers anyone later, gating on `current > 0` is a one-line follow-up.

Local verification of the maintainer commit: schema package tests pass; embedded repros A/B, the selective-staging unit test, and the #4566/#4567 dirty-gate test pass under `BEADS_TEST_EMBEDDED_DOLT=1`; build/vet/gofmt clean with `-tags gms_pure_go`.

Merging once CI is green on the new head. Thanks @csauer02-personal-user - this is a strong, carefully-proven finish to #4566.

PR #4617, opened 2026-07-07
feat(dolt): interpolate params in the server DSN

julianknutsen:
## What

Set `InterpolateParams=true` on the Dolt server DSN (`ServerDSN.String()`), so parameterized statements are rendered client-side into the SQL and sent as a single round-trip instead of a server-side PREPARE + EXECUTE pair.

## Why

Over a high-latency connection (for example a remote, TLS-fronted Dolt server), a write that issues many parameterized statements inside one transaction — creating an issue and its labels, events, and dependencies — otherwise pays a full network round-trip per statement just to prepare it. Interpolating parameters roughly halves the statement count on the wire for these write batches.

The driver falls back to a server-side prepare (`driver.ErrSkip`) whenever it cannot safely interpolate an argument, so query results never change; this only removes the extra round-trip when interpolation is safe. It is independent of `MultiStatements` and is rejected by the driver only with custom unsafe collations, which this DSN never sets.

## Verification

```
go test -tags gms_pure_go ./internal/storage/doltutil/ -run TestServerDSN
```

Added `TestServerDSN_InterpolatesParams`, which asserts `interpolateParams=true` is present in the formatted DSN. Existing `TestServerDSN_*` cases continue to pass.

### DIFF
--- a/internal/storage/doltutil/dsn.go
+++ b/internal/storage/doltutil/dsn.go
@@ -36,13 +36,26 @@ func (d ServerDSN) String() string {
 	}
 
 	cfg := mysql.Config{
-		User:                 d.User,
-		Passwd:               d.Password,
-		Net:                  net,
-		Addr:                 addr,
-		DBName:               d.Database,
-		ParseTime:            true,
-		MultiStatements:      true,
+		User:            d.User,
+		Passwd:          d.Password,
+		Net:             net,
+		Addr:            addr,
+		DBName:          d.Database,
+		ParseTime:       true,
+		MultiStatements: true,
+		// InterpolateParams renders bound parameters into the SQL client-side, so
+		// a parameterized query is a single round-trip instead of a server-side
+		// PREPARE + EXECUTE pair. Over a high-latency connection (e.g. a remote
+		// TLS-fronted Dolt server), a write that issues many parameterized
+		// statements — such as creating an issue and its labels, events, and
+		// dependencies inside one transaction — otherwise pays a full round-trip
+		// per statement for the prepare alone. The driver falls back to a
+		// server-side prepare (driver.ErrSkip) whenever it cannot safely
+		// interpolate an argument, so results never change; this only removes the
+		// extra round-trip when interpolation is safe. Independent of
+		// MultiStatements. The driver rejects it only with custom unsafe
+		// collations, which this DSN never sets.
+		InterpolateParams:    true,
 		Timeout:              timeout,
 		AllowNativePasswords: true,
 	}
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR sets `InterpolateParams=true` on the Dolt server DSN (`ServerDSN.String()`), so eligible parameterized statements are rendered client-side and sent as a single round-trip instead of a server-side PREPARE + EXECUTE pair. For multi-statement write batches over a high-latency, TLS-fronted Dolt server — creating an issue together with its labels, events, and dependencies in one transaction — that roughly halves the statement count on the wire, which is a real latency win for remote server-mode deployments and changes no query results.

### Original PR Review
Decision: approve.

What the review validated (no blocking gaps required fixing):
- Server-mode DSNs now request client-side interpolation: `internal/storage/doltutil/dsn.go` sets `InterpolateParams: true`, and `internal/storage/doltutil/dsn_test.go` asserts the formatted DSN contains `interpolateParams=true`.
- Prior behavior is preserved for unsafe cases: the go-sql-driver/mysql driver returns `driver.ErrSkip` whenever it cannot safely interpolate an argument, so it falls back to the normal server-side prepare path and results are unchanged.
- The "unsafe collation" caveat does not apply here: the driver rejects interpolation only when a custom unsafe collation is configured, and `ServerDSN.String()` never sets `Collation`.

Review findings addressed:
- Change impact / blast radius (minor, high confidence — Codex + Claude): the parallel dbproxy DSN helper (`internal/storage/dbproxy/util/dsn.go`) still omits `InterpolateParams` and would miss this optimization. This is non-blocking because proxied-server/dbproxy mode is currently rejected before normal use (`cmd/bd/init.go`, `cmd/bd/store_factory.go`), so no pre-merge change was required.

Remaining non-gating notes:
- If proxied-server/dbproxy mode is ever re-enabled, align `internal/storage/dbproxy/util.DoltServerDSN.String()` with `doltutil.ServerDSN` — either set `InterpolateParams: true` there with matching DSN coverage, or route dbproxy through the shared `doltutil.ServerDSN` helper once the TLS fields can be reconciled.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `09d134fa1` has passing required GitHub checks — build, lint, the full Embedded and Server Dolt test matrix, the contract corpus, differential regression, and the upgrade smokes are all green — and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/28851834527

### Review Iterations
One review pass. The multi-model review (Claude + Codex) approved the change on the first pass with no blocker or major findings, so no fix iterations were needed.

---

PR #4630, opened 2026-07-07
feat(preflight): report base-branch CI health before merge decisions

maphew:
## Why

Between 2026-07-05 and 2026-07-07 main was red for about 33 hours while a high-throughput salvage train merged dozens of stale-branch PRs. Once main went red, every open PR inherited red checks, "the failures are pre-existing" became the default reasoning, and three additional independent breakages (errcheck, testing.Short policy, unclaim test regressions, plus a -race prune flake) landed under that cover before #4623 and #4624 batch-fixed main. Nothing in the existing preflight looks at the base branch, which is exactly the condition under which its per-PR check verdicts become misleading.

## What

`scripts/pr-preflight.sh` now reports base-branch CI health alongside the PR's own checks:

- For each workflow on the base branch, judge by its newest completed run that finished success or failure. Cancelled runs are superseded pushes and carry no signal, and a green unrelated workflow that happened to run last (e.g. a docs deploy) cannot mask a red test workflow.
- Red base: `[warn]` by default, spelling out the pre-existing-failure trap and the merge-only-the-fix rule. `PR_PREFLIGHT_BLOCK_RED_BASE=1` escalates to a blocking `[block]` finding for stricter autonomous-agent policies.
- Green base: `[pass]` with the workflow count. Indeterminate (no Actions, no decisive runs, or `gh run list` failure): `[warn]` to check manually, never a hard error.

## Verification

- `bash -n` and `shellcheck` clean.
- Live green path against this repo (3 workflows, all green as of this run).
- Red path simulated with a fixture where a docs workflow succeeded after Main failed: Main's failure is surfaced, the later green docs run does not mask it, cancelled runs are ignored.
- Empty/no-Actions path degrades to the manual-check warning.

### DIFF
--- a/scripts/pr-preflight.sh
+++ b/scripts/pr-preflight.sh
@@ -21,6 +21,8 @@ Before implementing a feature/fix or opening a replacement PR:
 What this checks:
   - whether an existing PR is external/cross-repository contributor work
   - draft, review, mergeability, and check status
+  - CI health of the base branch (red base makes "failures are pre-existing"
+    reasoning unsafe; set PR_PREFLIGHT_BLOCK_RED_BASE=1 to block instead of warn)
   - risky diff signals: .beads data, missing tests for code changes, large diffs
   - contributor-protection next steps and attribution reminders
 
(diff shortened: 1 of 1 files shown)


maphew:
Found one blocker while reviewing this:

`scripts/pr-preflight.sh` currently treats only `success` and `failure` base workflow conclusions as decisive. If the latest base run for a workflow ends `timed_out` or `action_required`, the script skips it and can fall back to an older `success` for the same workflow, reporting the base as green even though the latest terminal result is red. That should match the existing PR check handling: terminal red conclusions count as red, while superseded `cancelled` runs stay ignored.

I am fixing this in place on this branch before merge.

PR #4420, opened 2026-06-16
feat(setup): add bd setup cursor with MCP integration

kevglynn:
## Summary

- Implements `bd setup cursor` generating two files: `.cursor/rules/beads.mdc` (prohibition-first agent safety rails) and `.cursor/mcp.json` (registers `bd mcp` as a native tool server)
- MCP merge uses `json.RawMessage` to preserve all unknown fields (`env`, `cwd`, `url`, `type`) from existing server entries during round-trip — no data loss
- Smart detection skips MCP injection when a beads-capable server already exists under any key (supports Jawnt and custom setups)
- Removal only deletes the `beads` MCP entry if it was installed by `bd setup cursor` (won't touch third-party registrations)

## What Gets Installed

**`.cursor/rules/beads.mdc`** — always-applied rules covering:
- Agent safety (commands that will hang)
- Prohibited patterns (fake commands, wrong flag syntax)
- Workflow (ready → claim → work → close)
- Session close checklist

**`.cursor/mcp.json`** — registers `bd mcp` stdio server, merged non-destructively into existing config.

## Test plan

- [x] `TestCursorRulesTemplate` — validates required content in generated rules
- [x] `TestCursorHasBeadsMCP` — detection logic for existing beads servers
- [x] `TestInstallCursor_FreshWorkspace` — clean install creates both files
- [x] `TestMergeCursorMCP_PreservesExistingServers` — doesn't clobber other entries
- [x] `TestMergeCursorMCP_PreservesUnknownFields` — round-trips `env`, `cwd`, etc.
- [x] `TestMergeCursorMCP_SkipsWhenBeadsKeyExists` — defers to existing `beads` key
- [x] `TestMergeCursorMCP_SkipsWhenBDMCPUnderCustomKey` — detects `bd mcp` under any name
- [x] `TestMergeCursorMCP_InvalidJSON` — returns actionable error, leaves file unchanged
- [x] `TestRemoveCursor_SkipsThirdPartyBeadsKey` — won't delete Jawnt's registration
- [x] `TestRemoveCursor_PreservesOtherServers` — only removes beads entry
- [x] `TestRemoveCursor_DeletesMCPFileWhenOnlyBeads` — cleans up empty config
- [x] `TestInstallCursorIdempotent` — second run is a no-op

Made with [Cursor](https://cursor.com)

PR #4649, opened 2026-07-07
feat(cursor): bring Cursor integration to parity with Claude/Codex

### LINKED ISSUE #4417: feat: add `bd setup cursor` for Cursor IDE integration
Beads has `bd setup aider`, `bd setup claude`, and `bd setup opencode`, but no Cursor equivalent. This came up in [Discussion #206](https://github.com/gastownhall/beads/discussions/206) (unanswered since Nov 2025).

**Proposed scope:**

- Generate `.cursor/rules/beads-workflow.mdc` with always-applied rules (use `--json`, no `bd edit`, no markdown TODOs, session completion = push)
- Optionally generate a starter MCP server config pointing to bd commands
- Mirror the `bd setup aider` pattern: single command, documented output, non-destructive

Happy to implement this.

HackAttack:
Cursor support is anemic compared to Claude Code and Codex. It ships only a static rules file: no agent lifecycle hooks, so `bd prime` context is never injected at session start and is silently lost after context compaction, and it has no `bd init` auto-install, no `--global` scope, no shared agent skill, and no `bd doctor` coverage.

This PR brings Cursor up to the same footing as the tier-1 integrations.

- `bd setup cursor` now installs agent lifecycle hooks (`.cursor/hooks.json`) alongside the rules file, wiring sessionStart/preCompact/postToolUse to a hidden `bd cursor-hook` command so Beads context survives compaction:
  * sessionStart injects full `bd prime` context into new sessions
  * preCompact arms a one-shot refresh marker
  * postToolUse re-injects `bd prime` once after a compaction, then no-ops
- `bd init` auto-installs Cursor (rules + skill + hooks) and stages `.cursor/`
- `bd setup cursor --global` writes `~/.cursor/hooks.json` + `~/.agents/skills/beads`
- Canonical rules: `.cursor/rules/beads.mdc` wraps the shared `recipes.Template` instead of a hand-maintained copy
- Shared agent skill (`.agents/skills/beads`), removed only once neither Cursor nor Codex still relies on it
- `bd doctor` reports Cursor Integration, Settings Health, and Hook Completeness checks with agent-mode enrichment

Shared hook mechanics (prime runner + one-shot refresh marker) are factored into `cmd/bd/agent_hook.go` and reused by codex-hook.

Resolves #4417


maphew (commented):
Thanks for this — it's an unusually complete parity PR. The hook design is well reasoned (working within Cursor's constraint that only `sessionStart`/`postToolUse` can inject `additional_context`, using `preCompact` to arm a one-shot marker), and the surrounding engineering is careful:

- Merge/preserve semantics for user hooks in `.cursor/hooks.json` are correct and idempotent, with tests covering upsert, double-install, preserve-on-remove, and remove-when-only-ours.
- The shared `.agents/skills/beads` refcounting is handled **in both directions** (`RemoveCursor` keeps it while Codex is installed, `removeCodex` keeps it while Cursor is installed), with tests for each.
- Factoring the marker mechanics into `cmd/bd/agent_hook.go` also fixes a latent bug in the old `codexHookMarkNeedsRefresh` (it swallowed the `MkdirAll` error and returned nil without ever writing the marker).
- Doctor checks delegate bd-hook detection to the setup package (`CursorManagedHookEvents`), so installer and diagnostics can't drift — nice.
- The docs update (PATH resolution pitfalls, `--continue` behavior, the bare-prompt limitation) is genuinely useful.

Two things to fix before merge, both small:

1. **Windows home-dir isolation in the new tests.** `TestInstallCursorGlobalHooksOnly` / `TestRemoveCursorGlobal` (`cmd/bd/setup/cursor_test.go`) and the new `cmd/bd/doctor/cursor_test.go` tests set only `t.Setenv("HOME", ...)`. On Windows, `os.UserHomeDir()` reads `USERPROFILE`, so these tests would install/remove bd hooks in the developer's *real* `~/.cursor/hooks.json` and `~/.agents/skills/beads` — and `TestRemoveCursorGlobal` passes silently while doing it. CI doesn't catch this because the Windows job is smoke-only. Please add `t.Setenv("USERPROFILE", home)` beside each `HOME` override, matching the existing pattern in `cmd/bd/doctor/claude_test.go`.
2. **Rebase.** The branch is conflicting with `main`, but I checked with `git merge-tree` — the only conflict is `CHANGELOG.md`; everything else auto-merges.

Two questions, no code change required:

- The hook event names / payload / response schema (`cmd/bd/cursor_hook.go`) can't be verified from code alone. Can you confirm the exact Cursor IDE and `cursor-agent` builds you verified against (the CHANGELOG says the 2026.06 line)? Also, does Cursor tolerate comments (JSONC) in `hooks.json`? `parseHooksJSON` is strict JSON, so a commented user file would hard-fail install and trip the new doctor error.
- Coordination with #4420: that PR adds Cursor **MCP registration** and its own rules content, touching the same `cmd/bd/setup/cursor.go` / `internal/recipes/recipes.go` / `website/docs/integrations/cursor.md`. The two are complementary (this PR doesn't touch MCP; that one doesn't touch hooks/doctor/init). My inclination is to land this one first — wrapping the shared `recipes.Template` is the right call for the rules file — and have #4420 rebase its MCP work on top. Flagging so the landing order is deliberate rather than accidental.

One nit you can take or leave: `RemoveCursor` prints "✓ Removed Cursor integration" even when the shared skill was kept for Codex (the "Keeping Beads agent skill" line mitigates it), and `codexIntegrationInstalled` only looks at the AGENTS.md managed section — checking `.codex/hooks.json` too would make it symmetric with `cursorIntegrationInstalledAt`.

For traceability: this closes the gap tracked in #4417 — worth linking it in the PR description so the issue closes on merge.

With the `USERPROFILE` test fix and the rebase, I'm happy to merge this. If it's easier, I can push those two fixes to your branch directly (maintainer edits are enabled) — say the word.

HackAttack:
Thanks for pointing out #4417. I linked it and added `USERPROFILE` everywhere I was setting `HOME`. Also fixed a [tiny docs gap](https://github.com/gastownhall/beads/compare/318c99761a85254207918edbd14c022d179ea19a..806ba7d9085819c71fcf99d1a2ec78c219aefbea).

The Cursor documentation does not state that comments are allowed in hooks.json so I think strict parsing is ok. I don’t have an exact version of when I was doing this work but my current version (Cursor CLI) is `2026.07.08-0c04a8a`.

maphew (approved):
The requested Windows home-directory isolation is now present beside every new `HOME` override, the branch is rebased and clean against `main`, and the contributor answered the Cursor-version/strict-JSON questions. The full CI matrix is green at `806ba7d`.

Approved for the integration value described in #4417: lifecycle hooks, shared-skill coordination, init/setup support, and doctor coverage remain within the project charter’s integration boundary.

PR #4653, opened 2026-07-07
ci(docs): auto-apply CLI docs regeneration to PR branches

maphew:
## Why

Generated CLI docs (`docs/CLI_REFERENCE.md`, website CLI pages, `llms-full.txt`) are a recurring friction point: every change to CLI help text requires a manual regeneration step, and when it's forgotten the `check-doc-flags` gate goes red — most recently on 2026-07-07, when the `agent.profile` drift briefly turned main red after a merge. The gate is well designed (it attributes drift to the PR and even uploads the exact fix as the `cli-docs-freshness-patch` artifact), but a human still has to notice the artifact, download it, and apply it by hand.

Moving the check to nightly was considered and rejected: main-green is load-bearing for merge decisions (#4630 preflight base-CI health), and the check is cheap and deterministic. The right fix is to automate the regen, not defer the signal.

## What

A `workflow_run` job (`docs-autofix.yml`) that closes the loop when a PR run fails and left the patch artifact:

- **Same-repo PR branches**: applies the patch and pushes a `docs: auto-regenerate CLI reference` commit, then leaves/updates one marker comment.
- **Fork PRs**: leaves an idempotent comment with the one-liner apply recipe (`gh run download … && git apply --index … `) — no token we hold can push to a fork.
- Skips cleanly when the artifact is absent, the PR is closed, or the head moved since the failing run.
- Circuit breaker: if the failing head is already an autofix commit, it refuses to stack another and falls back to the recipe comment.

The apply logic lives in `scripts/docs-autofix-push.sh` so it is shellcheckable and locally testable.

## Security model

The privileged side never checks out or executes PR code — it checks out base `main` for scripts and treats the artifact strictly as data:

- anchored path allowlist: only the generated doc files, single path segment where a wildcard appears, conservative filename charset, `..` and symlink file modes refused before `git apply`;
- `git apply --index` supplies the underlying escape guards (rejects traversal/absolute paths and writes through in-patch symlinks);
- PR-controlled values (`head_branch`, `head_repository`, …) reach the shell only via `env:`, never inline `${{ }}` interpolation;
- the push token is passed per git command via `http.extraheader`, never written to on-disk config.

Worst case for a hostile patch: rewriting generated doc files on its own PR branch.

## Token note

Pushes with the default `GITHUB_TOKEN` do not retrigger PR checks — the fix lands but checks need a manual re-run (the bot comment says so). Configuring a `DOCS_AUTOFIX_TOKEN` repo secret (fine-grained PAT or App token, `contents:write` on this repo) makes same-repo PRs fully hands-off. Worth doing if this lands.

## Verification

- `shellcheck -S warning scripts/docs-autofix-push.sh` clean; workflow YAML parses.
- Functional tests with a stubbed `gh` against the script: workflow-file patch, `cli-reference/../../../` traversal, nested-subdir `.md`, and symlink-mode patches all REFUSED; allowed-paths patch proceeds; closed-PR / moved-head / deleted-fork inputs no-op with exit 0; circuit-breaker path posts the fallback comment.
- Cross-vendor review before opening: Codex (`codex review`, gpt-5.5, high) found two P2s (fallback recipe losing newly-added files with `commit -am`; fork+PAT falling through to a doomed push) — both fixed. An independent Claude reviewer confirmed the privilege model and found the allowlist glob laxity (`*` spanning `/`), the `--paginate | head -1` SIGPIPE-under-pipefail hazard, and the symlink-mode gap — all fixed as above; its P3 suggestions (token-in-URL, URL-encoding, deleted-fork guard, circuit breaker) are incorporated.

`workflow_run` behavior can only be exercised end-to-end after this merges (it runs the workflow file from the default branch); a follow-up drift PR can serve as the live test.

### DIFF
--- a/.github/workflows/docs-autofix.yml
+++ b/.github/workflows/docs-autofix.yml
@@ -1,0 +1,83 @@
+# Auto-fix stale generated CLI docs on PRs.
+#
+# The PR workflow's check-doc-flags job already computes the exact
+# regeneration patch with CI's canonical pinned build and uploads it as the
+# cli-docs-freshness-patch artifact (scripts/check-cli-docs-drift.sh). This
+# workflow closes the loop: when a PR run fails and left that artifact, it
+# pushes the regen commit to the PR branch (same-repo PRs) or leaves an
+# idempotent comment with the one-liner apply recipe (fork PRs).
+#
+# SECURITY MODEL: this runs with write permissions via workflow_run, so it
+# never checks out or executes PR code. It checks out the base default branch
+# for scripts, and treats the artifact strictly as data: every path in the
+# patch must match an anchored generated-docs allowlist (single path segment,
+# no traversal, no symlink modes) and `git apply --index` supplies the
+# underlying escape guards (scripts/docs-autofix-push.sh), so a hostile patch
+# can at most rewrite generated doc files on its own PR branch.
+#
+# TOKEN NOTE: pushes made with the default GITHUB_TOKEN do not retrigger PR
+# checks. Configure a DOCS_AUTOFIX_TOKEN repo secret (fine-grained PAT or
+# GitHub App token with contents:write on this repo) for fully hands-off
+# operation on same-repo PRs; without it the pushed fix still lands but
+# checks need a manual re-run. Fork PRs always get the apply-recipe comment -
+# no token we hold can push to a fork.
+
+name: Docs Autofix
+
+on:
+  workflow_run:
+    workflows: ["PR"]
+    types: [completed]
+
+permissions:
+  contents: write
+  pull-requests: write
+  actions: read
+
+concurrency:
+  group: docs-autofix-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+jobs:
+  autofix:
+    name: Apply CLI docs regeneration to PR
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'failure'
+    runs-on: ubuntu-latest
+    steps:
+      # Trusted side only: base repo default branch, never the PR head.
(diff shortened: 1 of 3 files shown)


maphew (commented):
I found one issue to fix before relying on the optional PAT path:

- In `.github/workflows/docs-autofix.yml`, the script receives `GH_TOKEN: ${{ secrets.DOCS_AUTOFIX_TOKEN || github.token }}`, but the PR text tells maintainers that `DOCS_AUTOFIX_TOKEN` only needs `contents:write`. Once that secret is configured, all of `scripts/docs-autofix-push.sh`'s `gh api` calls use the PAT, including the marker-comment list/create/update paths. GitHub's issue-comment endpoints require Issues or Pull requests read/write permissions, so a contents-only token will push the fix and then fail while posting/updating the status comment; fork fallback comments would fail too. Either split the API/comment token from the push token (keep `github.token` for `gh api`, use the PAT only in the git extraheader) or document/request the additional PR/comment permissions for `DOCS_AUTOFIX_TOKEN`.

Verification: read the PR diff and maintainer guidelines; checked current CI status; ran `shellcheck -S warning` on `scripts/docs-autofix-push.sh` clean.

maphew:
Fixed in 081d12071 with the first option you suggested: the tokens are now split. All `gh api` calls (PR lookup, marker-comment list/create/update) always use the workflow's `github.token`, which carries this job's `pull-requests: write`; `DOCS_AUTOFIX_TOKEN` flows only into the git `http.extraheader` for the push itself, via a new `PUSH_TOKEN` env var that defaults to the workflow token when the secret is absent. So `contents: write` remains the complete, accurate requirement documented for the PAT, and fork-fallback comments work regardless of PAT configuration.

Shellcheck clean; the stubbed-`gh` functional tests (refusal paths, no-op paths, circuit breaker) still pass with the default `PUSH_TOKEN` fallback.

PR #4654, opened 2026-07-07
feat(migrate): add --force flag as CLI twin of BD_ALLOW_REMOTE_MIGRATE

maphew:
# Why

The remote-migrate gate (#4259 / #4515 / #4516) has one unlock for the designated migrator: the `BD_ALLOW_REMOTE_MIGRATE=1` env var. An env var is the natural shape for scripted/CI use, but as the *primary* interactive escape hatch it is un-idiomatic: every comparable safety bypass in bd is a flag (`bd close --force`, `bd dolt push --force`, `bd init --force`, `bd reset --force`, ~20 more), and operators reaching this gate are typing the command by hand at a decision point the gate itself constructed for them.

This adds `bd migrate --force` (and `bd migrate schema --force`) as the CLI twin of the env var. `--force` rather than `--yes` because in bd's own conventions `--yes` means "skip a confirmation prompt" (migrate already has one for `--update-repo-id`), while `--force` consistently means "bypass a safety check" — and the gate is a refusal, not a prompt.

# What

- `--force` on `bd migrate` and `bd migrate schema` (no `-f` shorthand — a fork-risk bypass should be typed deliberately).
- The gate fires at store open in the root `PersistentPreRunE` (and in `autoMigrateOnVersionBump`), long before `migrateCmd.RunE`, so the flag is consulted there via a new `isForcedMigrate()` helper (same pattern as `isWorkingSetReconcileCommand`, #4566).
- It sets a process-local override in `internal/storage/schema` (`SetForceAllowRemoteMigrate`) rather than `os.Setenv`: an env mutation would leak the bypass into child processes (git hooks, dolt subprocesses).
- The override is consulted exactly where the env var is — only once the gate would otherwise fire — and prints the same one-migrator warning, attributed to the flag.
- User-facing guidance (gate text, JSON `remote_migrate_gate.options`, `EscapeHint`, bootstrap/init hints, the embeddeddolt lenient-gate warning, upgrading.md, RECOVERY.md) now shows `bd migrate --force` as the primary form. **The env var remains fully supported** and is documented as the scripted/CI equivalent; nothing is deprecated.
- Historical records (`CHANGELOG` history, `info.go` version notes, `website/versioned_docs`) are deliberately untouched.

- `--force` is rejected when combined with the preview flags (`--dry-run`, `--inspect`), and the rejection happens *before* the store opens: with the gate overridden, the open itself applies pending migrations, so honoring the preview flag afterward would be a lie.

# Agent-safety posture unchanged

The #4259 design point that agents must not receive an unconditionally runnable fix is preserved: the migrate command (now spelled `bd migrate --force`) still appears only inside the conditional `options[migrate]` entry with its `when` precondition ("single designated migrator, confirmed with the operator") and `risk` annotation, never as a top-level hint, and `AgentDirective()` is unchanged.

# Testing

- New schema-level unit test: the programmatic override bypasses the gate where it would otherwise fire (sibling of the env-var test).
- Updated string assertions in bootstrap/init embedded tests.
- New unit test for the prerun helpers (`isForcedMigrate` scoping, preview-flag conflict detection).
- The agent-safety tests now also assert `AgentDirective()` never embeds the runnable `bd migrate --force` form, mirroring the existing env-form assertion.
- `CGO_ENABLED=0 go build -tags gms_pure_go ./...`, `go vet`, and targeted gate/migrate/bootstrap tests pass; CLI reference docs regenerated with the pinned CGO-free flow.
- Reviewed pre-submission by two independent model families; the preview-flag conflict above was a cross-vendor review catch.

Not done here: `cmd/bd/info.go`'s `versionChanges` has no entry for the new flag — those entries look release-curated, so I left that to the release process (the CHANGELOG Unreleased entry is in).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `bd migrate --force` (and `bd migrate schema --force`): CLI flag twin of
+  `BD_ALLOW_REMOTE_MIGRATE=1` for bypassing the remote-migrate gate (#4259)
+  as the single designated migrator; process-local so it cannot leak into
+  child processes (git hooks, dolt subprocesses).
+
 ### Changed
 
 - **Remote-ahead adopts fast-forward automatically when it is provably
--- a/cmd/bd/bootstrap.go
+++ b/cmd/bd/bootstrap.go
@@ -711,7 +711,7 @@ func printBootstrapRemoteBehindGuidance(w io.Writer, e *schema.RemoteMigrateGate
 			"  Re-running `"+rerunCmd+"` will NOT fix this — the remote itself is behind.\n"+
 			"  Choose one:\n"+
 			"    • This machine is the designated migrator (exactly ONE machine should be):\n"+
-			"        "+schema.AllowRemoteMigrateEnv+"=1 bd migrate\n"+
+			"        bd migrate --force\n"+
 			"        bd dolt push\n"+
 			"      then other machines re-run `bd bootstrap` to adopt the migrated database.\n"+
 			"    • Another machine is the designated migrator: wait for it to push, then\n"+
--- a/cmd/bd/bootstrap_noninteractive_test.go
+++ b/cmd/bd/bootstrap_noninteractive_test.go
@@ -131,7 +131,7 @@ func TestPrintBootstrapRemoteBehindGuidance(t *testing.T) {
 	for _, want := range []string{
 		"needs 1 schema migration (v49 -> v50)",
 		"Re-running `bd bootstrap` will NOT fix this",
-		schema.AllowRemoteMigrateEnv + "=1",
+		"bd migrate --force",
 		"bd dolt push",
 	} {
 		if !strings.Contains(out, want) {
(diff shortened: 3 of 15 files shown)


maphew:
Reviewed PR #4654.

No actionable code-review findings.

What I checked:

- The new `bd migrate --force` / `bd migrate schema --force` path is scoped to the migrate commands and sets a process-local override before store open, avoiding env leakage to child processes.
- The `--force` conflict with preview modes (`--dry-run`, `--inspect`) is rejected before store open, which preserves the safety meaning of those preview flags.
- The remote-migrate gate still keeps the runnable escape hatch inside conditional operator guidance, while `AgentDirective()` remains non-runnable.
- User-facing guidance and CLI docs consistently prefer `bd migrate --force` while preserving `BD_ALLOW_REMOTE_MIGRATE=1` for scripted/CI use.
- The change fits beads core scope as CLI/storage recovery surface.

Verification notes:

- Inspected the PR patch and relevant source context locally in `A:\dev\mybd\.worktrees\beads\pr-4654-review`.
- Plain local `go test` hit the expected missing CGO ICU header on this machine; a CGO-free targeted test retry was started but did not complete within my review wait window.
- CI snapshot showed the required gate passing; I would merge once the remaining matrix jobs are settled.

PR #4656, opened 2026-07-07
feat(security): report-only malware-attachment detector

julianknutsen:
## Report-only malware-attachment detector

Adds a GitHub Action that scores new issue/PR/discussion comments and bodies for the active spam campaign (burner accounts posting an LLM-written "here's a fix" comment with a `*_fix.zip` attached via GitHub's `user-attachments` uploader) and **logs the moderation tier it would assign**.

### It takes no action
`permissions: contents: read` only — no minimize, delete, label, lock, or block anywhere. The goal is to bake against real traffic and measure the false-positive rate **before** any enforcement is wired up. Graduating to minimize-only, then human-gated delete/block, is documented in `.github/scripts/README-comment-moderation.md`.

### Coverage
Triggers on issue comments, issue/PR bodies (via `pull_request_target`, default-branch checkout only, never runs PR code), PR reviews + review comments, discussions, discussion comments, and commit comments — created **and edited** (payloads are often edited in). Bot/app authors are skipped. Comment bodies reach the detector only through the step `env:` block (never interpolated into a shell command).

### Detection
Tiered scoring gated on author association + account age. Two corrections from adversarial review are baked in:
1. `github.com` is not blanket-trusted — `/raw/`, `/blob/*?raw=true`, gist `/raw/`, foreign `releases/download`, and `objects.githubusercontent.com` archives are extension-gated, so hosting the zip on a throwaway repo does not evade it.
2. Timing / "third-party" never alone justify the top tier — a payload-specific signal is required, so an eager helper's `repro.zip` cannot be auto-actioned.

### Verification
`python3 .github/scripts/moderate_comment.py --selftest` passes: both real malware samples → tier A, three evasion variants → A, and the legitimate patterns (contributor `.tar.gz`, screenshots, logs, release links, self-issue `repro.zip`, bots) stay out of the destructive tier.

> Note: the new triggers only take effect once this is on the default branch, so this PR does not run the detector. Please review and merge when ready; it is safe (report-only).


maphew:
## Review — report-only malware-attachment detector

Careful design overall (env-only body passing, `pull_request_target` with default-branch checkout, adversarial-review corrections baked in). Because it's report-only nothing here is exploitable today, but these matter for the PR's stated goal: the bake period is meant to measure accuracy, and these are true-negative misses and false positives that would corrupt that data (and let payloads through once enforcement graduates). The dominant theme is that `classify()`'s host/path/extension matching is a brittle exact-match ladder with many trivial bypasses.

### Findings (most severe first)

1. **`comment-moderation.yml` — `BODY`/author `||` chains resolve to the PR, not the review, for `pull_request_review` events.** That payload always contains a `pull_request` object, and the chains put `pull_request.body`/`pull_request.user.*` ahead of `review.body`/`review.user.*`. So a malicious review body is never scored (the clean PR description is scored instead), and author/association/account-age are taken from the PR author, not the reviewer. A fresh burner reviewing a maintainer's PR is scored as `MEMBER` → exempt; conversely an empty-body approval on a PR whose description legitimately links a `.zip` scores that text under the reviewer's fresh-account identity. An entire trigger surface is mis-scored in both directions.

2. **`moderate_comment.py` (`classify`, github.com `/blob/` branch) — blob links are unconditionally `ALLOW`.** Only `/blob/*?raw=true` is extension-gated; a plain blob URL to an archive returns tier C with zero findings, while the identical `/raw/` URL is tier A. GitHub's blob page has a one-click "Download raw," so this is a trivial hosting variant of the exact campaign targeted.

3. **`moderate_comment.py` — `/user-attachments/assets/` is blanket-`ALLOW` with no extension gate.** Any file served under an `assets/` URL is exempt regardless of extension or account freshness, contradicting stated correction #1. A payload reachable via an `assets/` path drops to tier C — fully invisible.

4. **`moderate_comment.py` — `objects.githubusercontent.com` archives score only `SUSP` (2), never `CORE`.** That host is exactly where `github.com/user-attachments/files/...` downloads redirect to. Posting the resolved signed object-store URL serves the same zip but yields `hard_core=False` → tier B, escaping the only actionable tier.

5. **`moderate_comment.py` (`ext_of` / `classify` / basename) — extension and bait matching are defeated by URL decoration.** `basename`/`ext` are taken from the raw last path segment, so `bd_fix_v1.zip?x=1` (query), `.../bd_fix_v1.zip/` (trailing slash), `//github.com/...` (protocol-relative → empty netloc), and `github.com.` (trailing-dot FQDN) each cause the `$`-anchored `BAIT_NAME`/`MASQUERADE` regexes to miss and/or the github.com files branch to be skipped — every one demotes the real sample from A to B. One appended character defeats the hard tier.

6. **`moderate_comment.py` — `MD_LINK` regex exhibits quadratic backtracking on unclosed `[text](` fragments.** ~20s at 16k chars; GitHub bodies reach 65,536, extrapolating past the job's `timeout-minutes: 5`. A body of repeated `[`/`(` sequences (adversarial, or just pasted code) times out the run, burning a full runner allocation and scoring nothing — repeatable on every edit.

7. **`moderate_comment.py` — the `NAV_PREFIXES` allow-check is dead code and mis-fires.** `path.startswith("/" + p.split("/")[1])` tests for `/tree`, `/issues`, `/releases` at the path *root*, which never matches real `/owner/repo/tree/...` navigation, so legitimate source-tree links (e.g. `github.com/owner/repo/tree/main/dist/tool.zip`) fall through to `SUSP` and, with a young third-party account, cross the tier-B line — a false positive on ordinary helpful comments. Meanwhile it `ALLOW`s top-level paths like `/pulls` it was never meant for.

8. **`moderate_comment.py` — `host.endswith("dropbox.com")` has no dot boundary.** `notdropbox.com/scl/x/file.zip` returns `("CORE", 5, flag_host=True)`, making `corroborated`/`hard_core` true even in code context. An attacker-registered `*dropbox.com`-suffixed lookalike reaches auto-hard tier A on a host unrelated to Dropbox. Should be `== "dropbox.com" or endswith(".dropbox.com")`.

### Also worth a note (not ranked)

- **Bake-data integrity:** `commit_comment` events can structurally never reach tier A (no thread object → `third_party=False`, which the A gate requires), and `EVENT_CREATED_AT` uses `comment.created_at` for *edited* events, so the timing bonus describes original post time, not when the payload was edited in. Both skew the false-positive/negative measurements the report-only period is meant to produce. The `concurrency` group can also drop a still-pending intermediate edit's scoring despite `cancel-in-progress: false`.
- **`normalize_url` over-decodes** up to 3× before parsing, so `%23` → `#` reclassifies a real `.zip` filename as a fragment, dropping the extension gate (A→B).
- **Self-test caveat:** the SAMPLES pass, but several bypasses above (assets path, blob, object-store, query/slash decorations) are not represented in the self-test — the README calls the self-test the pre-merge check, so consider adding these as regression cases before merge.
- **ZFC (debatable):** `CONTRIBUTING.md` says "Do not add heuristics, keyword matching, ranking logic … in application code," but scopes it to "code that involves AI decision-making or orchestration." This detector is pure heuristics with no AI, so whether the rule binds is a maintainer judgment call, not a clear violation.
- **Cleanup:** the github.com `/raw/` branch and gist `/raw/` guard are redundant with their fallthroughs; `BODY_B64` is dead (no caller sets it); the author-`||`-chain is duplicated across both workflow steps; `OWN_ORG` is hardcoded rather than derived from `github.repository_owner` (breaks on forks).

PR #4658, opened 2026-07-07
perf(ready): make `bd ready --json` O(page), not O(N²) (~25× at 10k)

julianknutsen:
## What

`bd ready --json` was **O(N²)** at scale — ~**4.2 s** at 10k issues on the SQLite backend — and agents call it on **every work cycle**. This makes it **O(page)**: **~4.2 s → 0.16 s (~25×)** at 10k on SQLite, with **byte-identical output**. The bug is in backend-agnostic code, so every backend benefits (SQLite most; Postgres/MySQL/Dolt were already faster because they compile the same SQL to O(N) hash joins).

## Root cause

`bd ready --json` builds a six-way LEFT-JOIN "counts" mega-query (`sqlbuild.SearchCountsSQL`). The reverse-blocker `rc` subquery keys on a computed `COALESCE(depends_on_issue_id, …) AS dep_id`; SQLite can't auto-index that materialized expression, so it degrades to an `O(candidates × distinct_blockers)` nested loop (`EXPLAIN QUERY PLAN` → `SCAN rc LEFT-JOIN`). At 10k that's ~11 M inner visits. Separately, `cmd/bd/ready.go` re-ran the **entire mega-query unbounded** just to size the "Showing X of N" total.

## Fix

- **Page pushdown** (`sqlbuild.SearchCountsByIDsSQL`, `issueops`): resolve the ≤N ready IDs first via the cheap indexed ready query, then constrain the six count subqueries to those IDs. Per-issue counts are a function of the target issue's subgraph, so restricting the subquery inputs to the page cannot change any surviving row's count — output is byte-identical. Falls back to the unbounded query for pages > 1000 (placeholder safety).
- **Kill the double-count** (`cmd/bd/ready.go`): replace the unbounded re-count with a single indexed `COUNT(*)` over the ready predicate (`CountReadyWorkInTx`), exposed via `CountReadyWork` on the sqlkit / dolt / embedded-dolt stores (unwrapped past `HookFiringStore`).

## Proof

| | old | new |
|---|---|---|
| `ready -n 1 --json` @10k SQLite | 4.15 / 4.24 / 4.10 s | **0.18 / 0.16 / 0.16 s** |
| output (`-n 1`, `-n 20`, `--limit 0`) | — | **byte-identical** (643 B / 12,512 B / 6 MB) |

**Conformance:** embedded-Dolt (the oracle) **fully green (290 s)**; SQLite green; Postgres + MySQL ready/counts suites **`ok`**. The one Postgres failure in the full run — `Audit/…/LabelCollationOrder` — is **pre-existing** (it fails identically on the clean base: `GetLabels` binary-vs-case-insensitive collation divergence, unrelated to this change). Build / gofmt / golangci-lint clean on the changed files.

## Notes

- Base is `feat/choose-your-own-backend` because `CountReadyWork` is exposed on `sqlkit.Store` (this branch). The **page-pushdown** and **double-count** parts are backend-agnostic and backportable to `main`.
- Discovered while front-door-benchmarking the storage backends; this was the standout hot-path regression at scale.

### DIFF
--- a/cmd/bd/ready.go
+++ b/cmd/bd/ready.go
@@ -16,6 +16,14 @@ import (
 	"github.com/steveyegge/beads/internal/utils"
 )
 
+// readyWorkCounter is the optional store capability that sizes the total ready
+// count cheaply (SELECT COUNT(*) over the ready predicate) instead of re-running
+// the counts mega-query unbounded. Local backends implement it; callers fall
+// back to GetReadyWorkWithCounts when a store does not.
+type readyWorkCounter interface {
+	CountReadyWork(ctx context.Context, filter types.WorkFilter) (int, error)
+}
+
 var readyCmd = &cobra.Command{
 	Use:   "ready",
 	Short: "Show ready work (open, no active blockers)",
@@ -237,12 +245,26 @@ This is useful for agents executing molecules to see which steps can run next.`,
 			totalReady := len(results)
 			truncated := false
 			if filter.Limit > 0 && len(results) == filter.Limit {
-				countFilter := filter
-				countFilter.Limit = 0
-				all, countErr := activeStore.GetReadyWorkWithCounts(ctx, countFilter)
-				if countErr == nil && len(all) > len(results) {
-					totalReady = len(all)
-					truncated = true
+				// The page is full, so there may be more ready work. Size the true
+				// total N with a cheap COUNT(*) over the same ready predicate rather
+				// than re-running the whole counts mega-query unbounded. The count is
+				// byte-identical to len(GetReadyWorkWithCounts(Limit=0)); fall back to
+				// that method only if the store predates the ReadyCounter capability.
+				// Unwrap past the HookFiringStore decorator so the assertion reaches
+				// the concrete store that carries the optional ReadyCounter method.
+				if counter, ok := storage.UnwrapStore(activeStore).(readyWorkCounter); ok {
+					if n, countErr := counter.CountReadyWork(ctx, filter); countErr == nil && n > len(results) {
+						totalReady = n
+						truncated = true
+					}
+				} else {
+					countFilter := filter
+					countFilter.Limit = 0
+					all, countErr := activeStore.GetReadyWorkWithCounts(ctx, countFilter)
+					if countErr == nil && len(all) > len(results) {
+						totalReady = len(all)
+						truncated = true
+					}
 				}
 			}
 			if results == nil {
--- a/internal/storage/issueops/ready_work.go
+++ b/internal/storage/issueops/ready_work.go
@@ -14,7 +14,11 @@ import (
 )
 
 type readyWorkPredicates struct {
-	whereSQL         string
+	whereSQL string
+	// whereArgs binds only the WHERE placeholders (no ORDER BY params). It is
+	// what a bare COUNT(*) over the ready predicate needs; args carries the
+	// WHERE params followed by the ORDER BY params for the full page query.
+	whereArgs        []interface{}
 	orderBySQL       string
 	limitSQL         string
 	args             []interface{}
@@ -60,12 +64,14 @@ func buildReadyWorkPredicates(ctx context.Context, tx DBTX, filter types.WorkFil
 		inputs.ParentDescendantIDs = descendantIDs
 	}
 
-	whereSQL, args, err := sqlbuild.BuildReadyWorkWhere(filter, tables, inputs)
+	whereSQL, whereArgs, err := sqlbuild.BuildReadyWorkWhere(filter, tables, inputs)
 	if err != nil {
 		return nil, err
 	}
 
 	orderBy := buildReadyWorkOrder(filter.SortPolicy)
+	args := make([]interface{}, 0, len(whereArgs)+len(orderBy.Args))
+	args = append(args, whereArgs...)
 	args = append(args, orderBy.Args...)
 
 	var limitSQL string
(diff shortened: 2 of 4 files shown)


maphew (commented):
Ran a multi-angle review of this PR (8 finder angles → 36 candidates → adversarial verification; 14 findings survived, consolidated to the 8 actionable ones below).

**Overall:** the page-pushdown half is a real, well-diagnosed win and worth landing — the WHERE predicates are alias-safe for the bare ID page query, the args/placeholder ordering in `SearchCountsByIDsSQL` checks out, and the wisp-merge path is handled consistently. The `CountReadyWork` capability half, however, is not wired up at this head: despite the PR body's claim that it is "exposed via `CountReadyWork` on the sqlkit / dolt / embedded-dolt stores", no store type implements the method anywhere in the tree at this revision.

## Findings (most severe first)

1. **`internal/storage/issueops/ready_work_counts.go:191` — the "kill the double-count" half is dead code (CONFIRMED).** No type implements `CountReadyWork`, so the `UnwrapStore(...).(readyWorkCounter)` assertion at `cmd/bd/ready.go:255` never succeeds and every truncated `bd ready` still takes the unbounded-fallback branch — the exact re-count the PR removes. `CountReadyWorkInTx` has zero callers. Relatedly, the private interface lives in `cmd/bd` instead of `internal/storage` beside the other optional capabilities (`GarbageCollector`, `Flattener`, `BackupStore`, ...), so backends have no compile-time contract to implement against. Either wire the method onto the stores or drop this half.

2. **`cmd/bd/ready.go:256` — caller passes the filter with `Limit > 0` into `CountReadyWork`; the zero-the-limit contract exists only in comments (CONFIRMED).** Both fallback paths explicitly zero `Limit`, but the capability call does not. A future implementer that honors `filter.Limit` returns exactly `n == len(results)`, the `n > len(results)` guard never fires, and the "Showing X of N" warning silently disappears on that backend with no test or compile failure. Zeroing `countFilter.Limit` at the call site makes conformance structural.

3. **`internal/storage/issueops/ready_work_counts.go:199` — one wisp row (even a closed, never-ready one) evaporates the count fix (CONFIRMED).** `CountReadyWorkInTx` gates its fast `COUNT(*)` only on the wisps table being empty/missing; otherwise it falls back to `GetReadyWorkWithCountsInTx(Limit=0)` — the original unbounded mega-query materializing all counts/labels/deps_json just to call `len()`. Any workspace that has ever created a wisp silently regains the pre-PR latency. Two predicate `COUNT(*)`s plus overlap dedup would size N without the counts projection.

4. **`internal/storage/issueops/ready_work_counts.go:117` — `--limit` above 1000 is strictly slower than pre-PR (CONFIRMED).** `runReadyCountsInTx` executes the full ID page query, discards the result at the `len(pageIDs) > readyCountsPushdownMaxPage` check, then runs the unchanged mega-query anyway. Since `len(pageIDs) <= filter.Limit` by construction, checking `limit > readyCountsPushdownMaxPage` before issuing the ID query avoids the wasted pass — twice per call (issues + wisps families).

5. **`internal/storage/sqlbuild/counts.go:132` — `SearchCountsByIDsSQL` duplicates the ~100-line mega-query instead of generalizing `SearchCountsSQL` with an optional ID constraint, and the copies are already drifting (CONFIRMED).** The new form uses the `DepTargetExpr` constant while `SearchCountsSQL` still hardcodes the `COALESCE(depends_on_issue_id, ...)` literal three times. Because `runReadyCountsInTx` picks between them on limit/page size, a single-sided future edit breaks the "byte-identical" invariant only for `limit<=0` or pages >1000 — a mode split no single test run covers. One builder taking an optional ids param keeps the projection order in one place.

6. **`internal/storage/sqlbuild/counts.go:132` — no tests accompany code whose central claim is an equivalence (CONFIRMED).** Neither "by-IDs output is byte-identical to the unbounded query" nor "`CountReadyWorkInTx` == `len(unbounded result)`" is exercised by any test; the proof is a manual benchmark. An equivalence test across both paths would also catch findings 3–5 structurally.

7. **`internal/storage/issueops/ready_work_counts.go:148` — `readyCountsPushdownMaxPage=1000` hardcodes a cross-backend placeholder-limit claim in a layer that cannot know backend limits (CONFIRMED).** The repo already chunks large IN-lists via `sqlbuild.QueryBatchSize=200` precisely because parameter limits are backend-specific; this constant bakes in "8×1000 placeholders is safe everywhere" (false for e.g. an engine built with `SQLITE_MAX_VARIABLE_NUMBER=999`), and exceeding the cap silently falls back to the quadratic query instead of chunking.

8. **`internal/storage/issueops/ready_work_counts.go:154` — `scanReadyCountsRows` is a verbatim copy of the scan/dedupe loop inside `runSearchQueryInTx` (`search_counts.go:117-146`), differing only in the error prefix (CONFIRMED).** A future scan/dedupe fix applied to one copy makes bounded pages hydrate differently from unbounded/large ones. Extract a shared scan helper.

Minor, below the cut: `whereArgs` in `readyWorkPredicates` is derivable state whose only consumer is the currently-dead `CountReadyWorkInTx` (`ready_work.go:21`); the reorder-to-page-order block duplicates the same pattern in `getReadyIssuesInTx` (`ready_work_counts.go:130`); and the doc claim that deps_json element order is "preserved" overstates what `JSON_ARRAYAGG` without an ORDER BY guarantees (`counts.go:194`).

**Bottom line:** the pushdown is worth landing. Findings 1–3 mean the "Showing X of N" improvement never executes as shipped and would degrade or vanish under easy-to-hit conditions once wired — store wiring plus an equivalence test would resolve the top items.

julianknutsen:
Superseded by the reworked PR targeting `main` — that one implements CountReadyWork on the stores (this PR's Fix A was dead code, as @maphew caught), incorporates all 8 review findings, adds equivalence tests, and retargets main (the bug is on main, not just the feature branch).

Issue #4657, opened 2026-07-07
bd update --claim is not a hard CAS under true sub-second concurrency (load-tested)


Caiuse:
## Finding

`bd update <id> --claim` is documented as "Atomically claim the issue ...
fails if already claimed" and behaves correctly for the realistic sequential
case: claiming an already-`in_progress` bead a moment after the first claim
committed reliably errors ("issue already claimed by <assignee>", exit 1).

However, load-testing against a scratch bead by launching N `bd update <id>
--claim` calls as close to simultaneously as possible (shell background jobs,
same process launch instant) showed the claim is NOT a hard single-writer CAS
under true sub-second concurrency:

- 2-way race, 5 iterations: both callers reported success every time (5/5).
- 8-way race, 3 iterations: 2/8, 3/8, and 7/8 callers reported success
  (non-deterministic — not "exactly one wins").

This indicates the underlying implementation is likely a check-then-act
(SELECT status, then UPDATE) rather than a single `UPDATE ... WHERE status =
'open'` compare-and-swap with rows-affected verification, leaving a genuine
TOCTOU window when multiple callers hit it inside that window.

bd version tested: 0.58.0 (ae14933d)

## Context

Found while load-testing an orchestration harness (racework-v2) that
dispatches many concurrent developer-agent processes, each of which calls
`bd update --claim` on a `bd ready` task shortly after spin-up. In practice
the two racing callers are always full agent dispatches (subagent spin-up,
`bd show`, model round-trips) before either reaches its `bd update --claim`
line — this naturally imposes hundreds-of-ms-to-seconds of separation
between the two real claim calls, far wider than the sub-millisecond gap
between two `&`-backgrounded shell invocations used in this test. So this
did not cause an observed double-claim in that harness's realistic dispatch
cadence — but it would be expected to under denser/faster concurrent
dispatch.

## Suggested fix direction

If bd's claim is implemented as read-then-write, harden it server-side to a
single conditional `UPDATE ... WHERE status = 'open'` (or equivalent) with a
rows-affected check (or a DB-level transaction/row lock spanning the
check+write), so that under N simultaneous callers exactly one succeeds and
N-1 receive the "already claimed" error deterministically.

## Repro sketch

```bash
bd create "scratch claim-race bead"   # note the id, e.g. BD-xxxx
for i in $(seq 1 8); do
  bd update BD-xxxx --claim & 
done
wait
# Expect: exactly 1 success, 7 "already claimed" errors.
[...]

PR #4675, opened 2026-07-08
fix(claim): enforce unclaim ownership, stamp lease on proxied claim, non-zero exit on lost batch claim

sjarmak:
## Why

The claim CAS in `ClaimIssueInTx` is a correct single-writer compare-and-set, but three write surfaces around it are not safe for concurrent multi-agent use, and a load test on the claim path (context below) confirmed the gaps are reachable:

- `bd unclaim` accepts an `actor` but never compares it to the current assignee, so any process can release any worker's live claim. It also writes `status`/`assignee` without rewriting `row_lock` or clearing the lease columns, violating the invariant documented in `lease.go` (any status/assignee/lease mutation must rewrite `row_lock` so a racing heartbeat conflicts instead of silently cell-merging into a zombie).
- The proxied-server (uow) claim path runs the CAS but omits the lease and `row_lock` columns, so claims made through it are invisible to `bd reclaim` and reintroduce the cell-merge class.
- A batch `bd update --claim` where one id loses the claim but another wins exits 0, hiding the lost claim from exit-code-driven automation.

## What changed

- **`bd unclaim` ownership + invariant** (`internal/storage/issueops/unclaim.go`, `storage.go`, `dolt/issues.go`, `embeddeddolt/issues.go`): the UPDATE re-checks ownership at write time (`AND assignee = ?`), clears the lease columns, and rewrites `row_lock`; a non-owner is rejected with a typed `ErrNotOwner` instead of a silent no-op. A new `--force` flag bypasses the check for admin/reaper use. `UnclaimIssue` is now wrapped in `withRetryTx` like the other claim-family writes.
- **Proxied-server claim** (`internal/storage/domain/db/issue.go`, `issueops/lease.go`): the uow claim reuses the shared `LeaseSetClause`/`LeaseTTL` so it stamps the same lease and `row_lock` as `ClaimIssueInTx`; a proxied claim is now visible to `bd reclaim`. The fresh-vs-idempotent `ClaimRowResult` distinction is preserved.
- **Batch claim exit code** (`cmd/bd/update.go`, `cmd/bd/update_proxied_server.go`): a lost `--claim` in a batch now forces a non-zero exit in both the direct and proxied paths, while winning claims still commit.
- **Tests**: CLI-level coverage for non-owner rejection and `--force`; storage-level coverage that unclaim clears the lease columns and rewrites `row_lock` and that a post-release heartbeat fails; a proxied-claim test asserting the lease is stamped and the bead is reclaimable.

## Test plan

- `go build ./...`, `go vet ./cmd/bd/ ./internal/storage/...`, `gofmt -l`: clean
- `go test ./internal/storage/issueops/...`: pass
- Against a real dolt sql-server (testcontainers): `TestUnclaimOwnershipAndLease`, `TestIssueSQLRepository/Claim/StampsLeaseAndIsReclaimable`, and the claim/unclaim/lease family: pass
- Embedded (`BEADS_TEST_EMBEDDED_DOLT=1`): `TestEmbeddedUnclaim` (incl. the two new ownership cases), `update_claim_batch_partial_loss_exits_nonzero`: pass

## Context

Found during a concurrency audit of a multi-agent orchestrator that dispatches many concurrent `bd` claimers. Independent of #4461. The related claim-identity question (same-actor idempotency reported in #4657) is a separate concern; this PR hardens the release/transfer and proxied-claim surfaces, which are unsafe regardless of claimer identity.


maphew:
Maintainer push: merged current `main` into the branch to clear the conflict with the lease-semantics changes that landed in the meantime (gastownhall/beads#4718). Resolution details:

- `internal/storage/dolt/lease_test.go`: kept **both** sides — your `TestUnclaimOwnershipAndLease` plus main's new `TestBareUpdateClaimDoesNotArmLease` / `TestUpdatePreservesLeaseOnSameClaim` guards.
- `internal/storage/sqlkit/capabilities.go` (backend added on main after you branched): adapted to this PR's widened `UnclaimIssue(..., force bool)` signature.

No semantic conflict: lease-stamping on explicit/proxied claims coexists with main's bare-update rule — the full claim/lease/unclaim suite passes locally on the merged tree (`internal/storage/dolt`, `issueops`, `cmd/bd`). Your commits are intact. CLI docs verified fresh after the merge.

Thanks sjarmak — once checks rerun green this is in good shape for final judgment.

maphew (approved):
Approving after a full correctness pass on the merged tree. All three behavior changes verified: ownership enforcement is defense-in-depth (Go pre-check + write-time `AND assignee = ?` predicate + re-read disambiguation, with the rejected-unclaim-preserves-row_lock invariant asserted); lease stamping is reached only via the explicit claim verb (`spec.Claim` → `ClaimIssue`), so main's bare-update-claims-must-not-arm rule (gastownhall/beads#4718) is preserved — a bare status+assignee update routes through `ManageLeaseOnUpdate` (clear-only) and never touches this path; batch-claim exit-code change keeps same-actor idempotent re-claims at exit 0, so script-compat risk is low. Signature threading covers every backend including the new sqlkit; docs regen covers the only help-text change.

One should-fix, post-merge: no `[Unreleased]` CHANGELOG entry for the new `--force` flag and the batch exit-code change — we'll fold that into a pending changelog follow-up. Two nits (direct/proxied exit asymmetry on non-conflict errors; force-hint wording when the resolved actor is empty) recorded in the tracker, neither blocking.

Thanks sjarmak — solid, well-tested work across a tricky surface.

PR #4679, opened 2026-07-08
perf(ready): fix O(N²) `bd ready --json` counts (page-push + COUNT(*) sizing)

julianknutsen:
## What

`bd ready --json` runs an O(N²) counts query on `main` today (independent of the multi-backend work). Two wastes in backend-agnostic code:

- The six-way counts mega-query (`sqlbuild.SearchCountsSQL`) joins **all** ready candidates against a reverse-blocker `rc` subquery keyed on `COALESCE(depends_on_issue_id, …)` — un-indexable, so it degrades to an `O(candidates × blockers)` nested loop.
- `cmd/bd/ready.go` re-ran the **entire mega-query unbounded** just to size the "Showing X of N" total.

This fixes both: **page-push** the counts (resolve the ≤N ready IDs, then constrain every count subquery to that page, chunked by `QueryBatchSize`), and **size the total with cheap indexed `COUNT(*)`s** over the ready predicate (issues + wisps families, minus overlap) instead of the mega-query. Output is **byte-identical** — the per-issue counts are a function of the target's subgraph, so constraining subquery *inputs* to the page can't change any surviving row's count.

## Why `main` (not the feature branch)

The bug is on `main`: `counts.go:33` has the `rc` COALESCE nested loop and `ready.go` has the unbounded re-count — and `counts.go` is byte-identical between `main` and the multi-backend branch. `main`'s backends (`dolt`, `embeddeddolt`) both get the fix here. The one feature-only piece — a ~6-line `sqlkit.Store.CountReadyWork` so sqlite/postgres/mysql inherit it — is a **follow-up on the backend-refactor branch** (sqlkit doesn't exist on `main`).

## Impact

- **Dolt (embedded), 10k issues:** `ready -n 1 --json` ~2.3 s → ~1.65 s. Modest in absolute terms because embedded-Dolt's per-invocation cost is dominated by ~1.6 s of engine startup — but the structural change is **O(N²) → O(page)**, so it removes a cliff that grows with repo size rather than a fixed constant.
- **SQLite backend** (feature branch, once the sqlkit follow-up lands): the same fix is ~**25×** (SQLite has no startup floor and no hash join, so the O(N²) counts was the entire cost).
- Byte-identical results/counts across the board.

## Correctness

- New conformance tests (`internal/storage/conformance/conformance.go`): `ReadyCountsPageEquivalence` (page bytes == unbounded), `ReadyCountsWithWisps`, `ReadyCountsPageChunking` (page > `QueryBatchSize`) — each asserts by-IDs byte-identity **and** `CountReadyWork == len(unbounded)`. Run under every conformance backend.
- **embedded-Dolt conformance green**; `go build`/`go vet` clean.

## Review

This supersedes #4658 and incorporates its full review (thanks @maphew) — all 8 findings addressed: real `CountReadyWork` on the stores with the capability interface moved into `internal/storage`; predicate-`COUNT(*)` sizing so a single wisp no longer disables the fast path; call-site limit zeroing; cap-before-query; one generalized counts builder; IN-list chunking instead of a quadratic fallback; shared scan/dedupe helper; and the previously-missing equivalence tests.

### DIFF
--- a/cmd/bd/ready.go
+++ b/cmd/bd/ready.go
@@ -236,12 +236,26 @@ This is useful for agents executing molecules to see which steps can run next.`,
 			totalReady := len(results)
 			truncated := false
 			if filter.Limit > 0 && len(results) == filter.Limit {
+				// The page is full, so there may be more ready work. Size the true
+				// total N over the same ready predicate, zeroing the limit so the
+				// count is the full ready set (byte-identical to
+				// len(GetReadyWorkWithCounts(Limit=0))). Prefer the cheap COUNT(*)
+				// capability; unwrap past decorators (e.g. HookFiringStore) so the
+				// assertion reaches the concrete store. Fall back to the unbounded
+				// mega-query only when a store predates ReadyWorkCounter.
 				countFilter := filter
 				countFilter.Limit = 0
-				all, countErr := activeStore.GetReadyWorkWithCounts(ctx, countFilter)
-				if countErr == nil && len(all) > len(results) {
-					totalReady = len(all)
-					truncated = true
+				if counter, ok := storage.UnwrapStore(activeStore).(storage.ReadyWorkCounter); ok {
+					if n, countErr := counter.CountReadyWork(ctx, countFilter); countErr == nil && n > len(results) {
+						totalReady = n
+						truncated = true
+					}
+				} else {
+					all, countErr := activeStore.GetReadyWorkWithCounts(ctx, countFilter)
+					if countErr == nil && len(all) > len(results) {
+						totalReady = len(all)
+						truncated = true
+					}
 				}
 			}
 			if results == nil {
--- a/internal/storage/conformance/conformance.go
+++ b/internal/storage/conformance/conformance.go
@@ -11,7 +11,9 @@ package conformance
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"sort"
 	"testing"
 	"time"
@@ -66,6 +68,9 @@ func RunAll(t *testing.T, factory Factory) {
 	t.Run("ReadyUnblockedByClose", func(t *testing.T) { testReadyUnblockedByClose(t, factory) })
 	t.Run("BlockedIssues", func(t *testing.T) { testBlockedIssues(t, factory) })
 	t.Run("EpicsEligibleForClosure", func(t *testing.T) { testEpicsEligible(t, factory) })
+	t.Run("ReadyCountsPageEquivalence", func(t *testing.T) { testReadyCountsPageEquivalence(t, factory) })
+	t.Run("ReadyCountsWithWisps", func(t *testing.T) { testReadyCountsWithWisps(t, factory) })
+	t.Run("ReadyCountsPageChunking", func(t *testing.T) { testReadyCountsPageChunking(t, factory) })
 
 	// Labels
 	t.Run("Labels", func(t *testing.T) { testLabels(t, factory) })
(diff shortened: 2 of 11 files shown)


maphew (approved):
No findings.

I reviewed the ready-count page-push path, the new `ReadyWorkCounter` capability wiring, the shared counts SQL builder, and the conformance coverage. The bounded by-ID hydration preserves the old per-family top-N behavior before the existing issues/wisps merge, and the `COUNT(*)` total uses the same ready predicates with ORDER BY args correctly excluded from the count query.

Local validation:

- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/sqlbuild ./internal/storage/issueops`
- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/conformance ./internal/storage/dolt ./internal/storage/embeddeddolt`
- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run 'Test.*Ready|TestJSONContract_Ready|TestProtocol_Ready'`
- `bd-main/scripts/pr-preflight.sh 4679 --repo gastownhall/beads`

Preflight only warned that the PR has no closing issue reference; I do not see that as a code blocker for this perf fix.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes an O(N²) counts path in `bd ready --json` — it page-pushes the six-way counts (resolve the ≤N ready IDs first, then constrain every count subquery to that page, chunked by `QueryBatchSize`) and sizes the "Showing X of N" total with cheap indexed `COUNT(*)`s over the ready predicate instead of re-running the whole counts mega-query. That matters because it turns a query that degraded to an `O(candidates × blockers)` nested loop (the reverse-blocker `rc` COALESCE subquery) plus an unbounded re-count into `O(page)` work, removing a cliff that grew with repo size on one of the hottest CLI paths — while keeping output byte-identical.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Reverse-blocker counts hotspot: review validated the page-bounded counts query as result-preserving — the ID page, by-ID hydration, IN-list chunking, and page-order restoration paths all produce the same per-issue counts, because constraining the subquery *inputs* to the page can't change any surviving row's count.
- JSON ready-total sizing hotspot: review validated the inclusion-exclusion `COUNT(*)` total against the ready/wisp merge semantics, and confirmed both Dolt stores (embedded and server) expose the new `ReadyWorkCounter` capability that `bd ready --json` now uses instead of the unbounded fetch.

Review findings addressed:
- Correctness / behavioral (blocker & major): none introduced — review found no blocking or major issues; the two performance rewrites were confirmed byte-identical to the prior results and counts.
- Test evidence (minor, high confidence): the PR adds `ReadyCountsPageEquivalence`, `ReadyCountsWithWisps`, and `ReadyCountsPageChunking` conformance tests that assert by-IDs byte-identity **and** `CountReadyWork == len(unbounded)` across every conformance backend, closing the equivalence-coverage gap that #4658's review had flagged.

Remaining non-gating notes (follow-ups, not merge blockers):
- Ready issue/wisp overlap arithmetic is currently exercised only with disjoint IDs; a fixture with one ID present in both `issues` and `wisps` would harden the overlap-subtraction term.
- The new equivalence tests use the oldest-sort policy; a Hybrid-policy case straddling the 48-hour cutoff would cover the user-facing default ordering.
- The `deps_json` intra-array order comment is slightly stronger than SQL guarantees; either add multi-dependency coverage or soften/sort on scan if order is meant to be contractual.
- Non-JSON `bd ready` still materializes all ready work to size its total; routing it through `ReadyWorkCounter` is an optional consistency follow-up.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `ab2dbcd` has passing required GitHub checks (Build, Lint, the full Embedded Dolt test + conformance + storage matrix, Regression, and upgrade smokes all green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/28979318553

### Review Iterations
1 review pass performed; approved with no fix iterations required. This PR supersedes #4658 and carries forward its full review — all eight findings from that thread (real `CountReadyWork` on the stores behind a capability interface in `internal/storage`, predicate-`COUNT(*)` sizing, call-site limit zeroing, cap-before-query, one generalized counts builder, IN-list chunking, a shared scan/dedupe helper, and the previously-missing equivalence tests) are already incorporated here.

---

PR #4682, opened 2026-07-09
feat(cas): optimistic concurrency for beads — metadata + whole-row compare-and-swap

julianknutsen:
## What

Adds a lock-free **compare-and-swap** capability to beads, on two surfaces:

- **Single-key metadata CAS** (`CompareAndSetMetadataKey` / `CompareAndClearMetadataKey`, exposed as `bd cas set/unset`) — guarded `JSON_SET`/`JSON_REMOVE` for claim-once fences, epoch counters, and guarded release. No migration.
- **Whole-row CAS** — migration `0055` adds an opaque `revision` nonce to `issues`+`wisps`; every content write stamps a fresh one. `UpdateIssueIfMatch` / `DeleteIssueIfMatch` / `CloseIssueIfMatch` apply a mutation iff the current revision matches, and `--if-revision N` surfaces it on `bd update`/`close`/`assign`/`delete`.

Supporting machinery: a build-time AST guard proving **every** issues/wisps content write across both stacks (issueops + proxied `domain/db`) stamps a revision (lease keepalive and the `is_blocked` recompute are the documented exemptions); a federation degrade gate that fail-closes whole-row CAS when a single timeline can't be guaranteed; and a Dolt merge auto-resolver for the revision-only conflicts the nonce now produces on disjoint cross-clone edits.

## Why

Beads had no CAS primitive: any read-modify-write (dispatch epoch fences, drain reservations, the API's `If-Match`/`resourceVersion`) could only read-then-write and silently clobber a concurrent peer. A guarded write now fails a lost race with a typed precondition error instead of overwriting.

The merge resolver is required because the nonce is load-bearing: once every write stamps a fresh revision, disjoint-column edits that used to auto-merge cleanly across clones always conflict on `revision`. The resolver rebuilds such a row **cell-wise** — it enumerates the live schema dynamically (so a newly-added column can never be silently dropped), overlays the peer's disjoint columns, mints a fresh revision, and escalates any genuine non-revision conflict (including `updated_at` time-gaps) or add/delete to the operator.

## Verification

All verified against **real Dolt 2.1.8** (server + embedded) and end-to-end via the `bd` binary:

- Metadata + whole-row CAS one-winner concurrency races (exactly one winner, N−1 typed `PreconditionFailedError`).
- The merge resolver: both disjoint edits survive + a fresh third revision (single- and multi-column peer edits); the `updated_at` time-gap wedge, a real content conflict, and modify/delete all escalate; a schema-parity guard pins that `dolt_conflicts_issues` exposes every physical column.
- The AST completeness guard (`TestAllIssueRowWritesBumpRevision`) and the domain/db scan parity.
- CLI exit codes: matched writes succeed; a stale `--if-revision` exits **9** with `code:"precondition_failed"` + `expected/current_revision`; a gate refusal exits the distinct **13**; the single-issue validations (multi-ID, `--claim`, label/parent, `--cascade`, delete without `--force`) exit 1.
- `bd show --long` renders `Revision`; `bd show --json` already carries it. `JSONSchemaVersion` unchanged (additive).

```
make test  # + targeted real-Dolt suites:
go test ./internal/storage/{issueops,dolt,embeddeddolt,domain/db,versioncontrolops}/...
go test ./cmd/bd/ -run 'IfRevision|Cas|Conditional'
```


julianknutsen:
**Review pipeline: parked pending author rebase.** Five automated review attempts failed in preflight (`rebase_and_merge_conflicts_against_main` — this branch, like its CAS-stack siblings #4715/#4724, cannot be cleanly rebased/merged onto current main). To resume: rebase onto main, push, and re-add `status/needs-review-auto`.

maphew (commented):
First: sixteen days of silence on a PR that clearly took real effort is on us, not you — and worse, main moved under this branch four separate times while it waited, including your own #4911. Apologies. Deep review done today (head 745de00); here's the honest picture and a concrete path.

**What's genuinely excellent here, by name:** the one-winner real-Dolt concurrency tests asserting exactly-1 win / N−1 typed errors / zero raw errors (that last assertion is what makes the serialization claim testable instead of asserted); the AST completeness guard *with its own teeth-test* (`TestRevisionGuardHasTeeth`); and the `is_blocked`-recompute / lease-heartbeat exemption analysis on the stamping policy — correct, non-obvious, and worth preserving as documentation regardless of anything else.

**The collision, specifically.** Since this branch's base, main landed: #4911 (yours, 07-22: `RowVersion` on `row_lock` + `MergeMetadata`), #5008 (07-24: `--if-assignee`/`--if-status` with exit 13 = guard mismatch), field-level three-way automerge (34695bd, 07-24), and #5006 (verified claim writes). Net effect: `revision` now duplicates `row_lock`/`RowVersion` (same type, same opacity, same equality-only contract, same two tables — the only semantic delta is stamping breadth, which is a policy change to `freshRowLock()` call sites, not a second column); `CompareAndSetMetadataKeyInTx` duplicates `MergeMetadataInTx` (same transaction shape, same retry) while bypassing metadata-schema validation and full event recording; and the mergesettle work is a special case of the shipped automerge.

**Two blockers independent of the duplication** (they'd block even a conflict-free version):
1. Serializing `revision` breaks the interchange protocol — CI is red on §J6.5/§J6.6 round-trip equivalence, and `TestRowVersionNeverSerialized` locks in the deliberate opposite decision for the identical token. Fix is `json:"-"`.
2. Exit 13 collision: #5008 shipped 13 as "guard mismatch — retryable"; this PR uses 13 as "gate refusal — do NOT retry" and puts mismatch on 9. Semantically inverted on the same number.

Also worth knowing: import currently rewrites `revision` on every row (`issueUpsertColumns`), so re-importing an identical snapshot dirties every row and manufactures cross-clone conflicts — and the guarded write paths skip the metadata-schema validation and wisp demotion the unguarded paths perform.

**What we'd like to absorb NOW, with `Co-authored-by: Julian Knutsen`, whether or not you take the re-scope:**
- `freshRevisionDistinctFrom` applied to `row_lock` inside automerge — this fixes a live hazard on main *today* (an LWW-settled merge can keep one side's old token, letting a stale `ExpectedVersion` win a CAS it should lose). Highest-value idea in the diff.
- The AST completeness guard, retargeted at `row_lock` (with the `strings.Contains(body, "revision")` false-negative tightened).
- The exemption analysis as doc on `freshRowLock`.

**Suggested re-scope — two small PRs on current main instead of rebasing 3700 lines:**
- **PR A — metadata CAS:** add `expected *string` / `ifAbsent` preconditions to the existing `MergeMetadataInTx`; put `bd cas set/unset` on top. No interface change, no migration, keeps validation + events. ~200 lines with tests.
- **PR B — `--if-revision` on the existing `RowVersion`:** widen `row_lock` stamping to all content writes (safe once the AST guard lands), expose `ExpectedVersion` as `--if-revision`, reuse exit 13. No new column, no resolver, no gate.

**One design question only you can answer** (you have real-Dolt data none of us do): the federation gate says whole-row CAS is unsafe with a remote configured "until the revision-cell merge resolver lands" — and the resolver is in this same diff. Which did you conclude is true? That answer decides whether `--if-revision` is usable for anyone with a remote (i.e., nearly everyone), and it matters for PR B too.

**Coordination:** you have two open concurrency PRs (this + #4697) that both mint `PreconditionFailedError` and both claimed migration 0055. Rather than reconciling them independently and colliding a third time, we're proposing #4697 as the single reconciliation thread — review landing there too today. Happy to pair on the re-scope in either thread.

_claude-fable-5-high on behalf of maphew_ (deep-review agent findings; probes and CI analysis verified in a local worktree)

quad341:
@julianknutsen — before we scope Phase 3 of the versioned-beads work (#6136: versioned reads + `expectedRevision` CAS), I want to ask how you'd like this PR to relate to it, because Phase 3 walks straight through ground you already covered here.

Concretely: Phase 1 (#6304, up today) adds `issues.current_revision` — functionally the whole-row revision your migration 0055 introduced, now a `BIGINT` sequence tied to a per-issue version log rather than an opaque nonce. Phase 3 then has to guarantee that *every* content write across both stacks stamps it, and has to survive Dolt merges that conflict on that cell. Those are exactly the two hardest pieces in this PR: the build-time AST completeness guard, and the merge auto-resolver for revision-conflict cells. Rewriting either would be redoing your work five weeks later, which is neither good engineering nor good manners.

Three shapes that would work from our side — your call which, or something else:

1. You rebase this onto post-0067 `main` and we scope #6136 around it; your commits land as the core of Phase 3.
2. We carry the auto-resolver and AST-guard commits into the Phase 3 PR with your authorship intact, and you review the result.
3. You'd rather Phase 3 start clean, and review the #6136 design before we build.

Either way the metadata half (`bd cas set/unset`, no migration) is orthogonal to History and could split out and land on its own if it's still wanted.

No rush this week — Phase 2 (#6135) is ahead of this regardless. We'll hold #6136 scoping until we hear from you.

PR #4686, opened 2026-07-09
fix(list): nest bd list --tree by parent-child edges only, not any dep on an epic

rjc123:
## Problem

`bd list --tree` (the default list view) rendered **any** dependency whose
target is an epic as a parent-child relationship. So a task that merely
`blocks` — or `waits-for`, `discovered-from`, etc. — an epic was drawn as
that epic's *child*. A genuinely 2-layer parent tree could therefore display
as a 6+ level tangle, repeatedly causing false "the hierarchy is broken"
conclusions when the parent tree was actually clean.

## Impact on agents consuming the output

beads is heavily driven by coding agents, and `bd list --tree` is a primary
way they read the shape of an epic. The false nesting was especially costly
here: an agent reading the tangled tree would conclude the hierarchy was
broken and start "fixing" a structure that was already correct — re-parenting
tasks, opening grooming spikes, second-guessing a clean plan. In practice
this drove real churn (wasted grooming passes and reorg work) that traced
back entirely to the rendering, not the data. Making the view match the
underlying parent-child structure removes that whole class of false signal.

## Fix

`buildIssueTreeWithDeps` now nests strictly on the `parent-child` edge type
(plus the existing dotted-ID fallback). This makes the tree view consistent
with the rest of beads:

- The storage layer already scopes an epic's children to `parent-child`
  edges only (`epic_closure.go` — `WHERE type = 'parent-child'`).
- Genuine parenthood is always written as a `parent-child` edge
  (`create.go`, `update.go`), so no real children are dropped.
- Non-hierarchical edges stay off the tree but remain fully visible via
  `bd show`, `bd dep tree`, and the compact list view (`blocked by:` /
  `blocks:`).
- The prior `relates-to` special-case (#3936) is naturally subsumed by the
  single parent-child guard, and its regression test still passes.

## Tests

Adds `TestListBuildIssueTree_NonParentChildDepOnEpicDoesNotNest`, a
table-driven regression over `blocks`, `waits-for`, `conditional-blocks`,
`discovered-from`, and `related` edges on an epic — each must leave the
source as a root rather than nesting it. The test fails on the previous
behaviour and passes with this change. All existing tree/list/epic tests
pass.

### DIFF
--- a/cmd/bd/list_helpers_test.go
+++ b/cmd/bd/list_helpers_test.go
@@ -156,6 +156,44 @@ func TestListBuildIssueTree_NoDuplicateChildrenFromMultipleDeps(t *testing.T) {
 	}
 }
 
+// A non-parent-child dependency on an epic (blocks / waits-for / discovered-from)
+// is a workflow edge, not membership. It must not nest the source under the epic
+// in `bd list --tree`. The storage layer already scopes an epic's children to
+// parent-child edges only (epic_closure.go), so the tree view must match: nesting
+// a mere blocker as a child made 2-layer parent trees render as 6+ level tangles
+// and triggered false "the hierarchy is broken" conclusions during grooming.
+func TestListBuildIssueTree_NonParentChildDepOnEpicDoesNotNest(t *testing.T) {
+	epic := &types.Issue{ID: "bd-epic", Title: "Epic", Status: types.StatusOpen, Priority: 2, IssueType: types.TypeEpic}
+
+	// Each non-parent-child edge type that AffectsReadyWork or is otherwise a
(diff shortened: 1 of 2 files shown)


maphew (approved):
Reviewed the tree-construction change and its regression coverage. Restricting nesting to `parent-child` edges correctly keeps workflow dependencies out of the structural tree, while preserving the existing dotted-ID fallback. The table-driven cases cover the affected non-hierarchical edge types, and the full PR check suite is green.

Approving the change. Upstream `main` is currently red, so this should be refreshed and rechecked after the base-branch fix lands before merge.

PR #4694, opened 2026-07-10
fix(is_blocked): de-alias UPDATE targets so the recompute runs on SQLite/DoltLite

Wldc4rd:
## Summary

The `is_blocked` mark/unmark recompute in `internal/storage/issueops/blocked_state.go` emits MySQL-style **target-table-aliased** `UPDATE`s:

```sql
UPDATE issues i SET i.is_blocked = ... WHERE i.id IN (...) ...
UPDATE wisps  w SET w.is_blocked = ... WHERE w.id IN (...) ...
```

SQLite's grammar does not permit aliasing the `UPDATE` **target** table, so on the SQLite-backed DoltLite backend every recompute fails to parse:

```
recompute is_blocked (mark): near "i": syntax error
```

Because it is a *parse* error, it fires on **every** execution regardless of data — on any active⇄closed transition that recomputes `is_blocked` (`bd update --status closed`, `bd reopen`, `bd remove-dependency`, ...). `bd close` takes a different recompute path and succeeds, which makes the failure look intermittent when it is actually deterministic per command path.

The `DBTX` abstraction deliberately shares these templates between the embedded (`*sql.Tx`, SQLite/DoltLite) path and the server/proxied (MySQL / go-mysql-server) path, so the fix must remain valid on **both** dialects.

## Fix

De-alias all **four** templates (issue + wisp, mark + unmark):

- SQLite forbids a table-qualified `SET` LHS, so the `SET` columns are left **bare** (`SET is_blocked = 1, updated_at = updated_at`). The `updated_at = updated_at` self-assignment that suppresses `ON UPDATE CURRENT_TIMESTAMP` is preserved.
- Outer references in the `WHERE` clause and the correlated `EXISTS` subqueries are qualified **by table name** (`issues.id` / `wisps.id`) instead of by alias.
- Every subquery `FROM` clause is unchanged, so the MySQL **ERROR 1093** posture is identical to before (the templates already select from the target table inside `EXISTS`; this change does not add any new such reference).
- The shared `waitsForGateBlockedSQL` fragment only references the subquery alias `d.` and is untouched.

## Testing

The existing `issueops` suite uses **go-sqlmock**, which never parses real SQL — which is why this defect shipped. This PR adds the first real-engine coverage of the recompute:

- **`internal/storage/issueops/blocked_state_sqlite_test.go`** drives `RecomputeIsBlockedInTx` against a real SQLite engine (the grammar DoltLite uses). It **fails before this change** with `recompute is_blocked (mark): near "i": syntax error`, and passes after, asserting correct `is_blocked` flags for both the mark path and the unmark path. (`JSON_UNQUOTE`/`JSON_EXTRACT`, which DoltLite provides on top of the SQLite grammar, are registered as trivial prepare-time stand-ins; the grammar bug under test is independent of them.)
- **`internal/storage/dolt/blocked_state_dolt_test.go`** is the MySQL/Dolt (go-mysql-server) companion, exercising the de-aliased mark and unmark templates on the real engine to guard against a regression on the shared server/proxied path (runs under the Dolt test server in CI).

Additionally verified end-to-end on a real DoltLite store: the aliased `UPDATE` reproduces `near "i": syntax error`; the de-aliased `UPDATE` succeeds and sets `is_blocked` correctly.

`gofmt` and `go vet` are clean; the full `issueops` package suite is green.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/dolthub/driver/v2 v2.1.10
 	github.com/dolthub/eventkit v0.0.0-20260611184414-99f5693e696a
 	github.com/go-sql-driver/mysql v1.10.0
+	github.com/mattn/go-sqlite3 v1.14.8
 	github.com/olebedev/when v1.1.0
 	github.com/spf13/cobra v1.10.2
 	github.com/spf13/viper v1.21.0
--- a/internal/storage/dolt/blocked_state_dolt_test.go
+++ b/internal/storage/dolt/blocked_state_dolt_test.go
@@ -1,0 +1,53 @@
+package dolt
+
+import (
+	"context"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/storage/issueops"
+)
+
+// TestRecomputeIsBlockedInTx_DoltDialect is the go-mysql-server / Dolt
+// (MySQL-dialect) companion to TestRecomputeIsBlockedInTx_SQLiteGrammar in
+// internal/storage/issueops. The is_blocked recompute templates are shared by
+// both the embedded (SQLite/DoltLite) and server/proxied (MySQL/gms) backends
+// (see the DBTX doc comment in blocked_state.go), so de-aliasing the UPDATE
+// targets to fix the SQLite grammar must not regress the MySQL path — in
+// particular it must not re-trigger MySQL ERROR 1093 on the same-table EXISTS
+// subqueries. This drives the de-aliased mark and unmark templates against the
+// real Dolt engine.
+func TestRecomputeIsBlockedInTx_DoltDialect(t *testing.T) {
+	store, cleanup := setupTestStore(t)
+	defer cleanup()
+	ctx := context.Background()
+
+	// mark: bm-w blocked on open bm-x, established through the normal write
+	// path (AddDependency), which runs the shared mark template on Dolt.
+	seedBlockedPair(ctx, t, store, true)
+	if !isBlocked(ctx, t, store.db, "bm-w") {
+		t.Fatalf("bm-w should be blocked by open bm-x (mark template on Dolt)")
+	}
+	if isBlocked(ctx, t, store.db, "bm-x") {
+		t.Fatalf("bm-x blocks others but is not itself blocked")
+	}
+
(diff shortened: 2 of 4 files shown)


maphew (commented):
The SQLite regression coverage and the underlying de-aliasing goal are valuable, but there are two blockers to resolve before this can merge.

1. The alias-free recompute statements no longer match the MySQL error-1093 rewrite, whose pattern expects `UPDATE <table> <alias> ... WHERE <alias>.id IN (...)`. As a result the driver returns the raw alias-free statement and reintroduces the self-reference failure the rewrite protects against. Please either extend the rewrite and its test to support the new alias-free form, or retain an alias where the SQLite translation requires it.
2. Adding `go-sqlite3` changes the module closure but leaves `default.nix`'s vendor hash stale. The failing Nix job supplies the replacement hash: `sha256-+wLmfWZAsYi8sZqIm4380rgpaaBDcgPOs9EcY0ylQtY=`.

The branch is also currently conflicted, and upstream `main` is red. Once the compatibility fix, Nix hash, and rebase are in place, please rerun the checks; the existing SQLite tests should be retained.

maphew:
Thanks @Wldc4rd for digging into the aliased-UPDATE recompute problem — the analysis of where the `is_blocked` recompute goes wrong on SQLite was useful and correct.

Retiring this PR as superseded rather than merging: the backends feature that landed on main on 2026-07-10 (1fc38ba77, c79bb32bc — hours after this PR opened) added `internal/storage/sqlitedialect/translate.go`, whose `addUpdateAlias` / `rewriteUpdateSetTargets` now translate the aliased UPDATE recompute for the SQLite backend at the dialect layer. That covers the real SQLite case this PR targeted, and it does so on the storage-driver side of the boundary, which is where the roadmap wants dialect differences handled. The embedded backend is Dolt/go-mysql-server (MySQL dialect), which accepts aliased UPDATE targets, so no change is needed there — and as noted in the earlier review, the de-aliased form would break the mysqldialect error-1093 rewrite pattern.

No action needed on your side. If you hit a remaining repro of this on current main, please open an issue with the recipe and we'll treat it as a live bug in the sqlitedialect translator.

Issue #4690, opened 2026-07-09
bd 1.1 migration 0053 fails after dependencies.id removal


Bella-Giraffety:
During Gas Town upgrade to @beads/bd 1.1.0, an HQ Dolt database at schema 52 had dependencies already in split-target/content-derived-key shape with no dependencies.id column. Migration 0053_repair_rig_wisps.up.sql still emits INSERT/REPLACE into dependencies(id, ...) and SELECT d.id. Even though the database had zero issue_type=rig wisps and zero rig-wisp dependencies, bd migrate/doctor/write paths failed with Unknown column 'id' in 'dependencies' and left schema migration lock/bad connection symptoms.\n\nObserved checks before workaround:\n- DESCRIBE dependencies: no id column; columns issue_id, type, created_at, created_by, metadata, thread_id, depends_on_issue_id, depends_on_wisp_id, depends_on_external.\n- DESCRIBE wisp_dependencies: id column still present.\n- SELECT COUNT(*) FROM wisps WHERE issue_type='rig': 0.\n- rig wisp deps and deps-to-rig-wisps counts: 0 across hq/gastown/dotfiles/beads/gt.\n\nWorkaround used only after confirming 0053 was a no-op for this data: inserted schema_migrations version 53 with content hash f13909f0db89f48db694a782d1805eee7ee1b2b2e4dcb3d4b12352333cbcb330 for HQ, then bd migrate completed metadata update to 1.1.0.\n\nExpected fix: make 0053 schema-aware for the dependencies table shape too, and no-op cleanly when no rig wisps exist. Cleanup-first: fix the migration SQL itself rather than adding wrappers or runtime bandaids.

maphew:
Merged #4878 (squash): hardens the schema migration chain — the dropped dependencies.id repair restores the surrogate key with COALESCE precedence matching rekeyDependencyTable, so the post-migration rekey is a genuine no-op. Reviewed clean (opus reviewer subagent); two test-coverage nits noted on the PR for follow-up.

_claude-fable-5-high on behalf of matt wilkie_

PR #4697, opened 2026-07-10
feat(ownership): claim_fence + guarded verbs + tier-complete requested leases

julianknutsen:
## What

Work-ownership integrity primitives, re-authored from current main as three
self-contained slices (one per commit) after the 2026-07-27 review:

1. **`claim_fence`** — a monotonic ownership fence on `issues` and `wisps`
   (synced migration `0062_add_claim_fence` for issues; wisps ride
   `ignored/0019_add_wisp_claim_fence`, the `row_lock`/ignored-0013 precedent,
   so fresh clones materialize it too). Bumped on ownership transitions —
   claim, unclaim/release, lease reclaim, assignee change (verb or
   import-upsert), reopen via the verb path — and paired with a `row_lock`
   rewrite in the same SQL statement so Dolt's cell merge can never silently
   converge two divergent bumps (`TestFenceBumpAlwaysPairsRowLock` enforces
   this at the source level, including the fragment-composed proxied SQL).
   The claim CAS bumps **unconditionally** — including a same-actor claim of an
   open row already assigned to them, so one session of a user can fence out
   another — while the idempotent same-actor re-claim of an `in_progress` row
   misses the CAS and does not bump. Both pinned by tests. `claim_fence`
   serializes as `claim_fence,omitempty` (readable via `bd show --json`), rides
   `IssueSelectColumnsLite`, and survives import without ever moving backward
   (fresh rows adopt the imported fence; existing rows bump only on
   assignee change).

2. **`--if-fence` guards, folded into the shipped surface** — `ExpectedFence
   *int64` on `UpdateIssueOptions` and `CloseIssueOptions` beside
   `ExpectedVersion`/`ExpectedAssignee`/`ExpectedStatus` (#4911/#5008 pattern),
   `ErrFenceMismatch` beside `ErrVersionMismatch`, re-exported at the root.
   `bd update --if-fence` and `bd close --if-fence` follow #5008's exit
   taxonomy (13 iff every failure was a stale guard, nothing written; 1
   otherwise); `bd update --if-fence` also works in proxied-server mode via
   `domain.UpdateSpec`, exactly like the shipped guards. `bd unclaim
   --if-fence` is an additional conjunct on both release paths and keeps
   unclaim's shipped guard contract (mismatch = exit 1; `--if-assignee`
   semantics byte-unchanged). Security rule per the review: **a fence guard
   establishes freshness, not authority** — `--if-fence` alone never lifts the
   unclaim owner check, and `--force` never skips a supplied guard.

3. **`lease.auto` + `bd lease disarm`** — a DB-stored config key (default
   **on**, preserving shipped behavior byte-for-byte) that gates automatic
   lease stamping on claim, and a one-shot `bd lease disarm` that flips the
   key and sweeps armed lease rows in one transaction (bounded re-sweeps for
   claims in flight). Heartbeat on a disarmed store returns a typed
   `storage.ErrUnleased` instead of silently re-arming. Disarm is a sweep, not
   a standing rejection: an import can restore live lease rows and `bd
   reclaim` handles them normally.

## Why

A claim in beads is atomic, but every other ownership transition (release,
reclaim, reassignment) is unguarded, and an assignee string cannot distinguish
two processes with the same name. A stale release can stomp a fresh re-claim,
and a reclaimed-but-still-live worker can close a bead it no longer owns.
`claim_fence` gives every ownership transition a monotonic marker so a caller
that read `(assignee, fence)` can release or complete against exactly the
state it observed — and a holder fenced out by an intervening transition gets
a typed mismatch instead of silently winning the write. Unlike
`ExpectedVersion` (which rides `row_lock` and moves on every content write), a
fence guard survives unrelated edits and fails only when ownership actually
moved.

## Changes from the previous revision (responding to the 07-27 review)

- Rebased/re-authored onto current main; the #4675 commits are gone (merged).
- Migration renumbered to the next free number on each independent sequence:
  synced **0062** and **ignored/0019** (main owns synced through 0061 —
  `0060_add_storage_class`, `0061_content_derived_aux_ids` — and ignored
  through 0018). Both number lines stay gapless, so the CLI-bundle test keeps
  main's strict `COUNT(*) == MAX(version) == LatestVersion()` assertion
  unchanged. `check-migration-hygiene.sh` gains a monotonic-landing check
  (Check D): any added migration whose version is not strictly above the base
  branch's per-track max now fails, naming the failure mode — because the
  cursor is `MAX`-based, a below-max migration silently never applies on
  already-migrated stores while fresh clones still get it. That check caught
  this branch's own collision when main landed 0060/0061 mid-review.
- The import-upsert `row_lock` regression is undone (fresh INSERTs keep
  `freshRowLock()`).
- Guard slice re-built on `UpdateIssueOptions`/`CloseIssueOptions` — the
  context transport, `PreconditionFailedError`, exit 9, and typed
  `{code:...}` stderr bodies are gone; no `--if-assignee` re-registration.
- Cross-actor release now requires `--if-assignee` or `--force`; the fence is
  never a credential.
- Slice 4 (wisp leasing, `--lease-ttl`, batch renew APIs) is **out**, parked
  pending the maintainer ruling; the migration-repair helpers are **dropped**
  (evidence in the review thread — 0054 is self-guarded, 0055 drops those
  columns, and main's #4878/#4690 already cover the v53 drift and the commit
  boundary).
- All doc regeneration dropped (CLI docs are release-pinned); anti-steamroller
  copy and shipped guard semantics untouched by construction.

## Verification

Built and tested at the tip of main (`423afdcb2`), all via real
infrastructure — no mocked Dolt:

```
go build ./... && go vet ./...                        # clean
go test ./internal/storage/{issueops,sqlbuild,schema}/ ./internal/types/
go test ./internal/storage/dolt/ -run 'Fence|Guard|Checked|Disarm|Lease|Reclaim|Conformance'
BEADS_TEST_EMBEDDED_DOLT=1 go test ./internal/storage/embeddeddolt/   # incl. 0062+ignored/0019 convergence/selfheal
go test ./cmd/bd/ [-run ...] (+ BEADS_TEST_PROXIED_SERVER=1)          # incl. proxied fence-bump + guard pins
go test ./cmd/bd/protocol/                            # corpus green; only golden churn = claim_fence:1 on the reopened bead
golangci-lint run ./...                               # only the 2 pre-existing cmd/bd/main.go G304s
BASE_SHA=origin/main scripts/check-migration-hygiene.sh && make check-docs
```

Known pre-existing flake, measured on pristine main (fails ~2/5 there):
`TestReclaimRevertsExpiredOnly` — unrelated to this change.

The branch was additionally taken through an adversarial multi-lens review
(fence/SQL correctness, guard-surface conformance, security, shipped-contract
regressions, migration hygiene, review fidelity, test adequacy, code quality)
with independent refutation of every finding; all 11 confirmed findings and 11
selected nits are fixed in these commits (details in the review-response
comment).

Compatibility notes (also in CHANGELOG): `Storage.UnclaimIssue` /
`UnclaimIssueIfAssignee` gained an `expectedFence *int64` parameter (breaking
for direct interface consumers); out-of-tree `Storage` implementations ignore
`ExpectedFence` fail-open, same caveat #5008 documented for the sibling
guards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,115 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`claim_fence`: a monotonic claim fence on every claimable row**
+  ([#4697](https://github.com/gastownhall/beads/pull/4697)). A `BIGINT NOT NULL
+  DEFAULT 0` column on `issues` (migration 0062) and on `wisps` (ignored
+  migration 0019 — `wisps` is `dolt_ignored`, so fresh clones never execute the
+  synced track and the column has to ship on the ignored one, following the
+  `0013_add_wisp_row_lock` precedent). It advances only on ownership
+  transitions: claim; unclaim/release (including `--force` and the
+  `--if-assignee` CAS form); lease reclaim; an assignee change, through the
+  generic update path *or* an import/upsert that changes the stored owner; and
+  reopen (closed→open) through the reopen verb or the generic update path.
+  Content mutations never move it, and neither does close. One exception on the
+  import side is deliberate: the upsert bump keys on an assignee change alone,
+  so an import that flips a stored closed row back to open with the *same*
+  assignee does not bump, where the reopen verb does — import is convergence
+  toward a peer's snapshot, not an ownership verb performed here, and a routine
+  sync must not retire the `--if-fence` guards live workers hold on this
+  replica. Unlike the `RowVersion` token — random per write and deliberately
+  unreadable — the fence is deterministic and monotonic, so it rides
+  `bd show --json` and the JSONL interchange as `claim_fence` (with
+  `omitempty`, so a never-claimed issue exports byte-identically to before).
+  That readability is the whole point: a caller can snapshot the ownership
+  state it decided on and later prove that state is still current.
+
+  Two decisions here are not the obvious ones and are pinned by tests. The
+  claim compare-and-swap bumps **unconditionally** — a successful claim of an
+  open row *already assigned to the claiming actor* still advances the fence,
+  because two sessions of one user are two ownership holders and the arriving
+  session must be able to fence out whatever the earlier one snapshotted;
+  keying the bump on "did the assignee string change" would leave exactly that
+  case silently shared. (The idempotent same-actor re-claim of an already
+  `in_progress` row is a different path: `in_progress` is not a claimable
+  source status, so it never matches the CAS, writes nothing, and does not
+  bump.) And every statement that bumps the fence rewrites `row_lock` in the
+  same statement. A bare monotonic counter is precisely the write pattern Dolt
+  cell-merges silently — two concurrent N→N+1 bumps produce identical cell
+  values and therefore no conflict — so the paired random `row_lock` rewrite is
+  what forces racing ownership transitions to serialize into the 1213/1205
+  failure the retry loop replays. A source-scanning test
+  (`TestFenceBumpAlwaysPairsRowLock`) stops a later edit from dropping it.
+
+  `bd reclaim --json` reports the post-bump fence for each recovered issue
+  (`claim_fence` on every element of `reclaimed`), so a reaper's log names the
+  generation that fenced the dead worker out.
+
+- **`--if-fence`: the claim fence as a guard on `bd update`, `bd close` and
+  `bd unclaim`** ([#4697](https://github.com/gastownhall/beads/pull/4697)). The
+  fence becomes a guard by joining the surface bd already ships (`--if-assignee`
+  / `--if-status`, and the engine's `ExpectedVersion` row CAS) rather than
+  building a second one beside it: `ExpectedFence *int64` is a new field on
+  `UpdateIssueOptions` and `CloseIssueOptions`, `nil` meaning "no check" and a
+  pointer to `0` being the real "expected never claimed" assertion. There is no
+  parallel guard interface and no new exit code — a stale fence is
+  `storage.ErrFenceMismatch`, a sentinel sitting beside `ErrVersionMismatch` and
+  `ErrStatusMismatch`. Guards stay conjunctive: assignee, status, version and
+  fence are ANDed, every read shares the writing transaction, and a refusal
+  rolls that transaction back so nothing — not even the event row — is written.
+  What the fence adds over `--if-assignee` is that it survives the holder's own
+  content writes and still catches a release-and-re-claim by the *same*
+  assignee, which an assignee guard cannot see.
+
+  Exit codes follow the existing taxonomy rather than the fence's own
+  preferences: `bd update` and `bd close` exit **13** when every failed ID lost
+  on a stale guard and nothing was written, and 1 for anything mixed in (close
+  keeps its partial-success and already-closed exit-0 contracts intact).
+  `bd unclaim` deliberately does not adopt 13 — its shipped `--if-assignee`
+  contract is exit 1 for any failed release, and a fence miss is just another
+  failed release. A guard establishes **freshness, not authority**: on unclaim
+  the fence is an additional conjunct on both release forms and never a
+  substitute for either, so a stranger quoting the correct live fence still gets
+  `ErrNotOwner`, and `--force` — which waives the ownership check, not the
+  guard — never skips a supplied fence. In proxied-server mode `bd update`
+  threads the guard through to the re-read inside the unit of work's
+  transaction; `bd close` and `bd unclaim` have no compare-and-set on their
+  proxied paths at all, so `--if-fence` there is refused loudly rather than
+  silently dropped.
+
+  Library consumers get the guard through the existing `UpdateIssueChecked` /
+  `CloseIssueChecked` calls; out-of-tree `Storage` implementations that ignore
+  the new option field simply do not enforce it and should add support before
+  advertising fence semantics — the same caveat the `--if-assignee` guards
+  recorded. `UnclaimIssue` / `UnclaimIssueIfAssignee` did change signature; see
+  Changed.
+
+- **Opt out of automatic claim leases: `lease.auto` and `bd lease disarm`**
+  ([#4697](https://github.com/gastownhall/beads/pull/4697)). By default every
+  claim stamps a lease and a supervisor's `bd reclaim` recovers the work of
+  dead workers — the right default for a fleet with no other recovery
+  authority, and unchanged here. Deployments where an orchestrator already
+  holds its own liveness evidence can now turn stamping off, so a fleet that
+  does not heartbeat is never one stray `bd reclaim` away from mass-reverting
+  live work. `bd lease disarm` is the safe form of that flip: it sets the new
+  `lease.auto` config key to off and clears the lease rows of the claims
+  already holding one in a single transaction, then re-sweeps a bounded number
+  of times to catch claims whose transactions read `lease.auto` before the flip
+  committed (they touch disjoint rows, so nothing forces them to conflict).
+  Nothing is released — status, assignee and the claim fence (`claim_fence`)
+  are all untouched, because disarming is lease bookkeeping, not an ownership
+  transition. `bd config set lease.auto off` does the flip alone, without the
+  sweep.
+
+  With leases off, claims carry no lease row and are invisible to the reaper,
+  and `bd heartbeat` on such a claim returns the new `storage.ErrUnleased`
+  instead of arming one: heartbeat is strictly a renewal there, since arming a
+  lease as a side effect would silently re-create the very reclaim exposure the
+  disarm removes. A claim that still holds a lease row keeps renewing normally,
+  so disarming is a one-shot sweep rather than a standing rejection. Only an
+  explicitly falsy value (`off`, or anything `strconv.ParseBool` reads as false)
+  disarms; unset, `on`, and a typo all leave the shipped default armed.
+
 - **Replica-aware leases: `bd reclaim` no longer reverts a lease another
   replica granted** (wy-jpd3.7). A lease is only meaningful on the replica that
   granted it — the other machine's liveness view is stale by up to one sync
@@ -85,6 +194,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   one.
 
 ### Changed
+
+- **`beads.Storage.UnclaimIssue` / `UnclaimIssueIfAssignee` gained an
+  `expectedFence *int64` parameter, and `beads.BulkIssueStore` gained
+  `DisarmAutoLeases`** ([#4697](https://github.com/gastownhall/beads/pull/4697)).
+  Callers that release without a fence guard pass `nil` (unchanged behavior);
+  any external type that *implements* either interface must update the two
+  unclaim signatures and add `DisarmAutoLeases(ctx) (int64, error)` to compile.
+  Decorators that embed the interface for passthrough are unaffected.
 
 - **`beads.BulkIssueStore.ReclaimExpiredLeases` gained a `types.ReclaimFilter`
   parameter** (wy-jpd3.3), threaded through the domain use-case/repository
--- a/beads.go
+++ b/beads.go
@@ -378,6 +378,7 @@ var (
 	ErrNotClaimable    = storage.ErrNotClaimable
 	ErrCloseBlocked    = storage.ErrCloseBlocked
 	ErrVersionMismatch = storage.ErrVersionMismatch
+	ErrFenceMismatch   = storage.ErrFenceMismatch
 	ErrSelfDependency  = domain.ErrSelfDependency
 	ErrDependencyCycle = domain.ErrDependencyCycle
 	ErrFieldTooLong    = types.ErrFieldTooLong
--- a/cmd/bd/claim_fence_guard_embedded_test.go
+++ b/cmd/bd/claim_fence_guard_embedded_test.go
@@ -1,0 +1,424 @@
+//go:build cgo
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// fenceOf reads an issue's claim fence back through the same surface a caller
+// would use to construct --if-fence: bd show --json.
+func fenceOf(t *testing.T, bd, dir, id string) int64 {
+	t.Helper()
+	return bdShow(t, bd, dir, id).ClaimFence
+}
+
+// TestUpdateIfFenceCLI drives `bd update --if-fence` end-to-end against the
+// embedded backend. The fence guard joins the bd-wsqvw guard family, so a stale
+// fence writes nothing and exits ExitGuardMismatch; what it adds over
+// --if-assignee is that it survives the holder's own content writes and still
+// catches a release-and-re-claim by the SAME assignee.
+func TestUpdateIfFenceCLI(t *testing.T) {
+	if os.Getenv("BEADS_TEST_EMBEDDED_DOLT") != "1" {
+		t.Skip("set BEADS_TEST_EMBEDDED_DOLT=1 to run embedded dolt integration tests")
+	}
+	t.Parallel()
+
+	bd := buildEmbeddedBD(t)
+	dir, _, _ := bdInit(t, bd, "--prefix", "uif")
+
+	t.Run("live_fence_applies_stale_fence_exits_13", func(t *testing.T) {
+		issue := bdCreate(t, bd, dir, "Fence guard", "--type", "task")
+		bdUpdate(t, bd, dir, issue.ID, "--assignee", "alice", "--status", "in_progress")
+		fence := fenceOf(t, bd, dir, issue.ID)
+		if fence == 0 {
+			t.Fatalf("claim_fence = 0 after an assignee change; --if-fence has nothing to guard on")
+		}
+
+		// The live fence: the update applies.
+		bdUpdate(t, bd, dir, issue.ID, "--if-fence", fmt.Sprint(fence), "--notes", "still mine")
+		if got := bdShow(t, bd, dir, issue.ID).Notes; got != "still mine" {
+			t.Errorf("notes = %q after a matching --if-fence, want %q", got, "still mine")
+		}
+
+		// A content write does not move the fence, so the same guard still holds.
+		if got := fenceOf(t, bd, dir, issue.ID); got != fence {
+			t.Errorf("claim_fence = %d after a content write, want unchanged %d", got, fence)
+		}
+
+		// Ownership moves and comes back to the SAME assignee: --if-assignee
+		// would still pass here, --if-fence must not.
+		bdUnclaim(t, bd, dir, issue.ID, "--actor", "alice")
+		bdUpdate(t, bd, dir, issue.ID, "--assignee", "alice", "--status", "in_progress")
+
+		out, code := bdUpdateFailCode(t, bd, dir, issue.ID,
+			"--if-fence", fmt.Sprint(fence), "--notes", "should-not-apply")
+		if code != ExitGuardMismatch {
+			t.Errorf("stale fence exit code = %d, want %d\n%s", code, ExitGuardMismatch, out)
+		}
+		if !strings.Contains(out, "fence") {
+			t.Errorf("mismatch error should name the fence, got:\n%s", out)
+		}
+		if got := bdShow(t, bd, dir, issue.ID).Notes; got != "still mine" {
+			t.Errorf("stale --if-fence wrote through: notes = %q, want unchanged", got)
+		}
+	})
+
+	t.Run("fence_zero_is_a_real_assertion", func(t *testing.T) {
+		issue := bdCreate(t, bd, dir, "Never claimed", "--type", "task")
+		bdUpdate(t, bd, dir, issue.ID, "--if-fence", "0", "--notes", "pristine")
+		if got := bdShow(t, bd, dir, issue.ID).Notes; got != "pristine" {
+			t.Errorf("notes = %q after --if-fence 0 on a never-claimed bead, want %q", got, "pristine")
+		}
+
+		bdUpdate(t, bd, dir, issue.ID, "--assignee", "alice", "--status", "in_progress")
+		out, code := bdUpdateFailCode(t, bd, dir, issue.ID, "--if-fence", "0", "--notes", "should-not-apply")
+		if code != ExitGuardMismatch {
+			t.Errorf("--if-fence 0 on a claimed bead: exit code = %d, want %d\n%s", code, ExitGuardMismatch, out)
+		}
+	})
+
+	t.Run("mixed_batch_failure_exits_1_not_13", func(t *testing.T) {
+		issue := bdCreate(t, bd, dir, "Mixed fence batch", "--type", "task")
+		bdUpdate(t, bd, dir, issue.ID, "--assignee", "alice", "--status", "in_progress")
+
+		out, code := bdUpdateFailCode(t, bd, dir, issue.ID, "uif-nope99",
+			"--if-fence", "99", "--priority", "1")
+		if code != 1 {
+			t.Errorf("mixed fence+lookup failure exit code = %d, want 1\n%s", code, out)
+		}
+	})
+}
+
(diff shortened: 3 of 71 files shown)


julianknutsen:
**Review pipeline: parked pending author decision.** Six review runs ended with `lease_storage_model_conflict_needs_human_decision` — this branch's lease/storage model conflicts with current main (same as stack-siblings #4682/#4715/#4724). To resume: rebase the CAS stack onto main, push, re-add `status/needs-review-auto`.

maphew (commented):
Fifteen days of silence on 4265 lines of unusually careful work — that's on us, and worse, upstream shipped a competing guarded-write design (#5008) *while this sat unanswered*. You did not create this collision; we did. Deep review done today at head dfa9bbb; here's what's landing, what needs reconciling, and what needs a maintainer ruling. **This thread is also the proposed single reconciliation point for this PR + your #4682** — they currently collide with each other (both mint `storage.PreconditionFailedError`, both claimed migration 0055) as well as with main, and resolving them independently would just collide a third time.

**Credit first, specifically:** the Dolt cell-merge analysis (a bare monotonic counter cell-merges silently; the paired `row_lock` rewrite is what forces serialization) with an AST-level test enforcing the invariant; the honest "scope and honest limits" comments; the refusal to map read errors to `not_found` in `RenewLeasesInTx`. These are load-bearing insights.

**Fastest win — split out the migration repairs now.** `ensureDependenciesSplitTargets` + `ensureLeaseColumns` have nothing to do with fencing and are the most immediately landable code in the diff — their own PR could merge this week. One fix needed: the v54 pre-repair currently issues ALTER/CREATE INDEX outside the per-step commit boundary, which is exactly what's failing `TestEmbeddedMigrateConvergesUnderConcurrentInterruptedRetries_4566` in CI (the red is real, not flake). Bring the DDL inside the step commit or drop `ensureLeaseColumns` from the split. (Also: the doc comment on `ensureDependenciesSplitTargets` names the wrong function and table — copy-paste artifact.)

**Slice 1 (`claim_fence`) is wanted.** It fills a gap nothing upstream covers: #4911's `ExpectedVersion` rides `row_lock`, which every content write rewrites, so a guard on it fails on unrelated edits — a token that moves *only* on ownership transitions is a genuinely different and useful semantic. Reconciliation needed to land it:
- Renumber 0056 → 0060 (main now owns 0056–0059), moving the `cli_migrations.go` case with it; stale "migration 0055" references in types.go/scan.go/the SQL header follow.
- One real regression to undo: the import upsert drops `row_lock` from `issueUpsertColumns` — fresh INSERTs land with `row_lock = 0` (identical values on two nodes = the exact silent cell-merge the invariant exists to prevent) and status-only imports no longer rewrite it. Keep `row_lock` + `freshRowLock()` and add `claim_fence` as an additional column.
- Classify `claim_fence` into `IssueSelectColumnsLite` (main's new lite-shape parity test will be red otherwise) and move it out of `LeaseSelectColumns` — it's an issues/wisps column and belongs on the qualifiable side; note the scan order now includes `leases.granted_node`.
- Decide the bump policy explicitly: `fence.go` documents "ownership transitions only," but `ClaimIssueInTx` bumps unconditionally — including same-actor re-claims. We'd actually argue FOR unconditional bump (it's what lets one session of the same user fence out another, which matters a lot in multi-session agent setups where same-user claims are idempotent) — but then say so and test it; right now two paths implement opposite policies under one doc.

**Slice 2 needs folding into the shipped surface, not a parallel one.** #5008 (merged 07-24) deliberately put guards on `UpdateIssueOptions` "rather than minting a parallel interface method," with exit 13 = mismatch. This PR's context-value transport can't cross the proxy boundary — your own `validateGuardInvocation` has to refuse proxied mode with exit 13 (which simultaneously means "mismatch" upstream: unshippable contract, and the duplicate `--if-assignee` registration would panic the CLI at init after a naive merge). Ask: fold `--if-fence` in as `ExpectedFence *int64` beside `ExpectedVersion`/`ExpectedAssignee`, `ErrFenceMismatch` beside `ErrVersionMismatch`, exit 13 for mismatch. One more security-shaped fix while you're there: `--if-fence` alone currently lifts the owner check on unclaim — but the fence is a printable small integer, not a credential; a guard establishes freshness, not authority. Require the assignee guard (or actor == assignee) for cross-actor release.

**Slices 3 (lease.auto/disarm) — independently valuable, low controversy,** could be its own small PR after slice 1.

**Slice 4 (wisp leasing) needs a maintainer ruling before you touch the 247 conflict lines in lease.go:** shipped migration 0055 says in prose that wisps are never leased and dropped their lease columns as "never-reclaimed, entirely unread"; this PR reverses that. That's a product decision — we're routing it to the owner now and will come back with a yes/no rather than let you rebase against an unsettled question. Related: the justification currently lives in a gascity design doc beads maintainers can't read — please inline the core argument in the PR body.

Small stuff that deletes a third of your conflict volume for free: restore the deliberate anti-steamroller unclaim message (it encodes incident wy-yuclk), and revert the doc-comment rewrites on untouched functions (one now names the wrong method).

Want to run with the split (migration repairs first, then slice 1 reconciled, then slice 2 folded into #5008's surface)? Or we can absorb the migration-repairs split maintainer-side with `Co-authored-by` if you'd rather focus on the fencing core. Either way nothing here gets closed-and-forgotten — this thread stays the reconciliation point until both PRs have a landing path.

_claude-fable-5-high on behalf of maphew_ (deep-review agent findings; CI failure and merge-tree analysis verified in a local worktree)

julianknutsen:
Thanks for the depth of this review — taking the split exactly as proposed, with
two corrections discovered while executing it, one new hazard found in the
process, and one contract choice to confirm. The branch is re-authored from
current main as three commit-sized slices. No textual rebase survived contact:
every one of the 19 conflicted files was re-authored from main's side, so the
#4675 duplication, the `--if-assignee` rebind, the unclaim-copy regression, and
the doc-comment rewrites are gone by construction rather than restored by patch.

New head: `48f3a4909` (fence) → `35e2fadb9` (guard) → `7fc3d0df9` (lease.auto),
on `423afdcb2`.

**1. Migration numbers: synced 0062 and ignored/0019 — and the numbering
question earned a permanent tripwire.** Main owns the synced track through
**0061** (`0060_add_storage_class`, `0061_content_derived_aux_ids`) and the
ignored track through **0018** (`ignored/0018_aux_row_id_rekey2_marker`), so
this branch takes the next free number on each independent sequence: synced
`0062_add_claim_fence` and `ignored/0019_add_wisp_claim_fence`. Wisps get the
column on the ignored track rather than via a wisps ALTER inside the synced
migration: wisps is `dolt_ignored`, fresh clones never execute synced
migrations, and `ignored/0013_add_wisp_row_lock` is the exact precedent (the old
branch's synced wisps ALTER left fresh clones without the column).

The numbering churn on the way here turned up something worth keeping. **The
cursor is MAX-based** (`SELECT COALESCE(MAX(version), 0)` with an
`if current >= target` short-circuit), so a migration numbered *below* an
already-applied version silently never runs on that database, while fresh
clones still get it — a divergence no existing gate catches. So
`scripts/check-migration-hygiene.sh` gains a **Check D**: any added migration
whose version is not strictly above the base branch's per-track max fails,
naming that failure mode. It promptly earned its keep by catching this
branch's own collision when main landed 0060/0061 mid-review — hence 0062.
Self-tested on both tracks. (Methodology note for anyone re-running it: it
diffs against the base, so an unstaged file is invisible to it — stage the
file or the self-test proves nothing. CI is unaffected.)

No test expectation was loosened for any of this: an earlier revision of this
work had relaxed the CLI-bundle count assertion to tolerate a reserved-number
hole, and that relaxation is reverted — the suite keeps main's strict
`COUNT(*) == MAX(version) == LatestVersion()`, and both number lines are
gapless.

Related, for #4461: it currently carries `0060_create_provenance_events`,
which main's own 0060 now occupies, so it needs its own renumber independent
of this PR.

**2. The migration-repairs split dissolves under verification — proposing to
drop both helpers rather than split them out.** Digging into your "bring the DDL
inside the step commit or drop it" fork:
- `ensureLeaseColumns` (v54 pre-repair): 0054's frozen SQL is already fully
  self-guarded (INFORMATION_SCHEMA + PREPARE/EXECUTE per column and index, both
  tables) — there is no partial-apply brick for it to fix — and 0055 drops those
  columns one step later, so on every ≤53→latest upgrade it is pure
  add-then-drop churn. It is also the direct cause of the
  `TestEmbeddedMigrateConvergesUnderConcurrentInterruptedRetries_4566` red you
  verified: on the old branch the repair ran before the `dirtyBeforeStep`
  snapshot, stranding its ALTERs outside the per-step commit. Main already fixed
  that ordering in #4878 (snapshot-before-repair, with the #4690 comment), so
  even the "fix the boundary" half of the fork is already upstream.
- `ensureDependenciesSplitTargets` (v53): a different table and aspect than 0058
  (synced `dependencies` columns + backfill vs `wisp_dependencies`
  constraints), but the drift it guards is covered upstream — guarded 0041
  creates the split columns before 0053 in any pass, and the observed real drift
  (#4690, missing `dependencies.id`) is handled by `ensureDependenciesIDColumn`
  inside `repairV53RigAndSplitTargets`, which fails loudly if the split columns
  were ever truly absent. (Its doc header was, as you spotted, a copy of the
  wisp helper's — reconstructed mechanically during the July-19 rebase attempt,
  not designed against a field drift report.)

If you know of a live database exhibiting the v53 split-column drift, the
correct port is into `repairV53RigAndSplitTargets` before
`ensureDependenciesIDColumn`, re-entrant (no `len(missing)==0` early return),
registered via the `preMigrationRepairs` map — happy to do that as a follow-up.
Absent a field report it's dead weight, so this PR ships no repair changes:
`migration_repairs.go` and `schema.go` are untouched at main's version.

**3. Slice 1 (claim_fence) — landed per your reconciliation list.**
- Import upsert: `insertIssueIntoTable` keeps `row_lock` + `freshRowLock()`
  exactly as on main and adds `claim_fence`; fresh INSERTs never land with
  `row_lock = 0`. One deliberate refinement to "add `claim_fence` as an
  additional column": it is *not* in the ON-DUPLICATE assignment list —
  `claim_fence = VALUES(claim_fence)` would let an older snapshot import move a
  fence backward and re-validate retired guards. Fresh rows adopt the imported
  fence; existing rows move only via the assignee-change bump, which pairs with
  main's existing `row_lock = VALUES(row_lock)` rewrite.
- Classification: `claim_fence` rides `IssueBaseColumns` +
  `IssueSelectColumnsLite` (small routing/claim column, the `row_lock`
  precedent), not `LeaseSelectColumns`; both scanners, lite-merge widths, and
  the parity tests updated, with the scan-order note about `leases.granted_node`
  honored.
- Bump policy: ruled explicitly your way — the claim CAS bumps
  unconditionally, including a same-actor claim of an open row already assigned
  to them, so one session of a user can fence out another. Documented in
  `fence.go`, the 0062 header, and `types.Issue.ClaimFence` (all three now
  enumerate one canonical list), and pinned by tests on both sides: the open
  same-assignee claim bumps, the idempotent `in_progress` re-claim misses the
  CAS and does not. One documented exception the review surfaced: an
  import-upsert that flips a stored closed row back to open with the same
  assignee does *not* bump — import is convergence, not an ownership verb.
- `TestFenceBumpAlwaysPairsRowLock` ports, and got strengthened twice: it now
  requires a real `row_lock =` assignment in the SET clause (a WHERE-only
  mention used to satisfy it), and `TestFenceTransitionCoverage` now requires
  `FenceBumpExpr` in `domain/db/issue.go` — the proxied claim SQL composes the
  bump as a fragment, so the literal scan's `domain/db` leg was silently
  vacuous. The proxied bump also gained a behavioral pin (claim through the
  proxied server, assert the fence advanced).
- `claim_fence` serializes as `claim_fence,omitempty` — readable via
  `bd show --json` so callers can construct `--if-fence`, and zero-fence issues
  export byte-identically, so the entire golden churn is one line on the
  reopened corpus bead.

**4. Slice 2 — folded, not parallel.** `ExpectedFence *int64` on
`UpdateIssueOptions` and `CloseIssueOptions` beside `ExpectedVersion` /
`ExpectedAssignee` / `ExpectedStatus`; `ErrFenceMismatch` beside
`ErrVersionMismatch` (re-exported at the root); one shared primitive extended
rather than a new one minted. `--if-fence` on `bd update` (exit 13 on pure
stale-guard, threaded through the proxied path via `domain.UpdateSpec` exactly
like #5008's guards), on `bd close` (same taxonomy; proxied close refuses
loudly, the shipped unclaim `--if-assignee` precedent), and on `bd unclaim` as
an additional conjunct on both release paths. The context transport,
`PreconditionFailedError`, exit 9, and the typed `{code:...}` stderr bodies are
all gone, and there is no duplicate flag registration.

Security rule adopted verbatim: a fence guard establishes freshness, not
authority. `--if-fence` alone never lifts the unclaim owner check — cross-actor
release still requires `--if-assignee` or `--force` — and `--force` never skips
a supplied guard, on any verb. Negative `--if-fence` values are rejected as
usage errors (exit 1) rather than reported as mismatches, so a typo can't
masquerade as the racer-won verdict scripts branch on.

Two contract choices to confirm:
- **Unclaim's fence mismatch exits 1**, matching unclaim's shipped
  `--if-assignee` contract, while update/close use 13. Unifying unclaim on 13
  would change a shipped contract, so I left it and documented it.
- **`--if-fence` does not join #5066's `AssigneeNotStolen` skip set** — only
  `--if-assignee` and `--claim` sanction overwriting a live foreign claim. A
  matching fence is arguably a stronger observation than a named holder, so
  this may be worth extending, but it's your policy call; the help text now
  states the shipped behavior plainly instead of implying otherwise.

**5. Slice 3 ships here as its own commit** (`lease.auto` + `bd lease disarm`),
re-expressed against the leases-table model: disarm is a config flip plus a
lease-row sweep in one transaction, with bounded re-sweeps for claims in
flight; heartbeat on a disarmed store returns a typed `ErrUnleased` instead of
silently re-arming. The wisp upgrade interlock
(`TestDisarmClearsLegacyWispLeases`) is dropped as moot — main's 0055 already
removed wisp lease columns, so there are no legacy wisp leases to clear. Also
honest in the help now: disarm is a one-shot sweep, not a standing rejection —
an import can restore live lease rows and `bd reclaim` handles them normally.

**6. Slice 4 stays out** pending the wisp-leasing ruling (none published as of
today). The renew/count APIs and `--lease-ttl` went with it. On inlining the
design justification: that belongs in the slice-4 PR, since this branch no
longer touches wisp leasing at all — happy to write it there once the ruling
lands either way.

**7. For #4682's reconciliation** (this thread as the single point): this PR no
longer mints `PreconditionFailedError`, claims no 005x migration number, and
uses no exit 9 — so #4682's re-scope onto `RowVersion` / exit 13 can proceed
without colliding with anything here.

**Note on #4715:** it stacks on this branch's old head, which this force-push
replaces (old head `dfa9bbb93`, still reachable via this PR's previous-head
SHA). It needs re-stacking regardless of its own review outcome, and its
`0055_add_claim_fence` / `0056_add_holder_token` numbering now has Check D to
answer to.

**Verification.** Rebuilt and gated at the tip of main (`423afdcb2`):
build/vet/gofmt clean; `golangci-lint` shows only the two pre-existing
`cmd/bd/main.go` G304s; the unit suites (issueops, sqlbuild, types, schema)
green; the dolt-server behavioral suite green across fence/guard/disarm/
conformance over several full runs; the embedded cgo suite green including 0062
+ ignored/0019 through the migration convergence and self-heal harnesses;
`cmd/bd` embedded and proxied suites green including the proxied fence-bump and
proxied guard exit-13 pins; protocol corpus green; `make check-docs` and
migration hygiene green. Known pre-existing flake, measured failing ~2/5 on
pristine main and therefore not from this branch:
`TestReclaimRevertsExpiredOnly`.

The branch also went through an adversarial multi-lens review (fence/SQL
correctness, guard-surface conformance, security, shipped-contract regressions,
migration hygiene, review fidelity, test adequacy, code quality), with every
finding independently refutation-checked before acceptance. All 11 confirmed
findings and 11 selected nits are folded into the three commits — the two most
load-bearing being the Check D tripwire above and the vacuous-AST-scan hole on the
proxied claim path.

_claude-fable-5 on behalf of @julianknutsen (re-authoring, verification, and
adversarial review run in a local worktree; every gate above executed against
real Dolt, not mocks)_

julianknutsen:
CI is green at `9123e0612` (87 passed, 2 skipped, 0 failed; mergeable/clean) —
one commit past the head named above, for a fix worth calling out because it
touches a shared harness rather than this feature's own code.

**Differential Regression (v0.49.6 baseline)** failed on four scenarios
(`TestReopenRoundTrip`, `TestReopenPreservesDeps`, `TestStatusTransitions`,
`TestUpdateDoesNotClobberRelationalData`) with the same root cause: each
transitions ownership, so the candidate's JSONL carries `"claim_fence": 1`
while the pinned baseline binary has no such column, and the differential
export diff reported the feature as a regression.

Fixed by adding `claim_fence` to `versionSpecificFields` in
`tests/regression/regression_test.go` — the harness's existing mechanism for
fields that exist in one version and not the other, stripped from *both* sides
before the diff, alongside `_type`, `deleted_at`, the epic rollup counters, and
friends. Explicitly *not* fixed by skipping the scenarios or by dropping
`claim_fence` from the interchange: the field needs to round-trip (it's how a
caller learns the generation to pass to `--if-fence`, and import-time fence
adoption is what keeps a restored row's guards honest), and the four scenarios
are load-bearing for exactly the reopen/update paths this branch touches — they
now compare everything except the one column the baseline cannot know about.

Worth noting for future additive columns: any new exported field trips this
suite the same way, and `versionSpecificFields` is the intended landing spot.

maphew:
Re-reviewed at `9123e061`. Every ask from the 07-27 pass was executed — migration renumbered, the `row_lock` import-upsert regression undone, `claim_fence` classified into `IssueSelectColumnsLite` and out of `LeaseSelectColumns`, the parallel guard interface replaced by #5008's shipped surface, the context-value transport and exit-9 taxonomy deleted, the `--if-assignee` re-registration removed, cross-actor release re-secured. Thank you for taking the re-author rather than patching over it.

**The important structural change, which I verified directly:** this branch is no longer blocked on anything on our side.

Slice 4 is gone. The diff touches only `0062_add_claim_fence` and `ignored/0019_add_wisp_claim_fence`, both additive — `0055_move_leases_to_table` is untouched, so **nothing here reverses a migration**. That was the one thing holding this PR behind an unanswered maintainer question about whether wisps should be leasable work. `ignored/0019` adds a `claim_fence` column to `wisps`, which is not a lease column; wisps have been *claimable* on main since before this PR, and the file's own header preserves "wisps are never leased". So fencing wisp claims is consistent either way that question eventually lands, and it no longer needs answering first.

One housekeeping ask: **the PR title is now stale.** "tier-complete requested leases" advertises slice 4, which is out. Worth retitling before merge so the commit log doesn't imply a decision that was never made.

## What I checked rather than assumed

The fence is a real CAS where it matters: `unclaim.go:104-112` and `:220-232` put it in the WHERE clause as a conjunct with a re-read disambiguation on zero rows, and `claim.go:100-116` rides the bump inside the existing claim CAS paired with `RowLockClause()`, so two concurrent claimers take a Dolt cell conflict and the loser's CAS matches nothing. Both cannot win. The `update`/`close --if-fence` paths use precheck-in-same-tx plus the row_lock replay — the same structure the already-merged `ExpectedVersion`/`ExpectedAssignee` guards use, so no new TOCTOU window.

I went looking specifically for a bypass and didn't find one: every non-test SQL site that writes `assignee` bumps, both import/upsert sites prepend `UpsertFenceAssignments` first in the `ON DUPLICATE KEY UPDATE` list (which is load-bearing, since assignments evaluate in order), and where the proxied transport genuinely can't carry the guard it refuses loudly instead of silently dropping it. That last point was the failure mode flagged last time and it's handled.

Migrations are `INFORMATION_SCHEMA`-guarded and additive, numbering is gapless above the base max on both tracks, and CI is green (87 success, 0 failures).

## Findings — all improvements, none invalidating the design

1. **`fence_invariant_test.go:143` — the coverage test has a whitelist.** `TestFenceTransitionCoverage` claims to enforce "every ownership-transition entry point carries a fence bump", but `mustBump` is a hardcoded `{claim.go, unclaim.go, lease.go, reopen.go}`. A new file in `issueops` that writes `assignee` passes silently — precisely the drift the test exists to catch. Walking all non-test files in the package and requiring a bump wherever the SQL assigns `assignee` would close it. (`TestFenceBumpAlwaysPairsRowLock` does scan everything, but enforces the reverse implication.)
2. **`UnclaimIssue` gained a positional guard param**, giving `UnclaimIssue(..., force bool, expectedFence *int64)` alongside `UnclaimIssueIfAssignee(..., expectedAssignee string, expectedFence *int64)`. This is the shape #5008 explicitly rejected. An `UnclaimIssueOptions{Force, ExpectedAssignee, ExpectedFence}` would collapse both methods into one, delete the caller-side branch in `cmd/bd/unclaim.go:112-117`, and break the interface once rather than once per future guard — it deletes more than it adds. Given you've already done one full re-author, I'm happy to take this maintainer-side rather than send the branch back again; say the word.
3. **The concurrency claims are argued, not demonstrated.** The cell-merge reasoning is correct and `TestFenceBumpAlwaysPairsRowLock` is a good structural guard, but it's a source-level scan — no test spawns two racing claimers and asserts exactly one wins with fence `N+1` rather than both winning or the fence landing at `N+2`. Since "many parallel agents against one tracker" is the target operating condition, one behavioral race test would materially strengthen this. Mitigating: the serialization mechanism is `row_lock`, which is concurrency-tested on main — so under-evidenced rather than unevidenced.
4. **`bd lease disarm` can't distinguish convergence from budget exhaustion.** `DisarmAutoLeasesWith` breaks on `n == 0`, but exhausting `maxDisarmResweeps` while still clearing rows prints the same `cleared N lease(s)`. On a command whose purpose is removing reclaim exposure, the operator should get a "re-run `bd lease disarm`" signal — returning a `converged bool` would do it.
5. Smaller: `--force` + `--if-fence` is mutually exclusive in `update.go:874` but deliberately allowed in `unclaim.go:171` / `close.go:435`, with two contradictory rationales stated for the same flag pair — pick one and explain the odd verb. Batch `--if-fence` applies one fence to every ID, which unlike `--if-assignee` is nearly always a user error given a fence is a strictly per-row generation. And there's no fresh-clone assertion that `wisps.claim_fence` exists — the 0013 failure class `ignored/0019` is there to prevent — since the schema-test fixture runs `AllMigrationsSQL()`, which by design omits the ignored track.

Worth calling out separately: the migration-hygiene **Check D** addition catches a below-`MAX(version)` silent-skip class the repo had no gate for at all, and it caught its own collision on the way in. That's a contribution beyond this feature.

**Approve with comments.** The correctness core is sound. None of the above blocks merge; (1) and (3) are the two I'd most like to see, and (2) I'll absorb myself if you'd rather not touch the branch again.

sjarmak:
Checking in on this from the Gas City side, with one piece of information that has not been on this thread: what it is holding.

Gas City registered `beads.guarded_release` as an inert gate (gastownhall/gascity#4274) deliberately waiting on a bd pin that carries the guarded verbs, with `BD_GUARDED_RELEASE_MIN_VERSION` left out of `deps.env` on purpose until one exists. `claim_fence` is not in any shipped bd release (v1.2.1 is current), so that gate and every consumer behind it stay inert. This PR is the pin blocker for the whole ownership-fencing lane downstream, not just for beads.

State as I read it today: approved with comments on 2026-07-30 at `9123e0612`, CI green at 87 passed and 0 failed, and explicitly not blocked on the reviewer's side. Nothing since, and the branch has gone back to `CONFLICTING` against main, so the three-slice re-author from `423afdcb2` has drifted out of date again. Worth saying plainly that this is a stalled PR rather than a contested one.

Two things that would otherwise ship wrong, both already noted in the 07-30 review:

- The title still advertises "tier-complete requested leases", which is slice 4 and was cut. Merged as-is, the commit log implies a decision nobody made.
- maphew offered to absorb the `UnclaimIssueOptions` collapse maintainer-side rather than send the branch back a third time. That offer is two weeks old and may still stand.

No pressure on the two open improvements from that review, the coverage-test whitelist and the racing-claimers test. Both are real and neither was called blocking.

The concrete ask: do you want to refresh this against main yourself, or would you rather hand off the conflict resolution? I have no merge authority here and I am not proposing to take over your branch. If a second pair of hands on the rebase is useful, say so and I will ask on our side; if you would rather it wait, that is a fine answer and I will stop pinging it, but we would then plan around the pin staying where it is.

PR #4699, opened 2026-07-10
feat: auto-resolve issues-table modify/modify conflicts by LWW (updated_at)

### LINKED ISSUE #4698: Auto-resolve modify/modify issues-table conflicts by last-write-wins (updated_at) — biggest remaining manual wall in multi-clone pull
## Summary

`bd dolt pull` bails to manual conflict resolution for the `issues` table even when every conflict is a provably-safe modify/modify where one side is strictly newer by `updated_at`. The existing auto-resolver already handles `metadata`, `dependencies`, `schema_migrations`, and `config` with table-specific safety checks, but `issues` hits the default case (`return false, nil`) and dumps every conflict to the operator. This is the biggest remaining manual wall in the core multi-clone sync workflow.

## Background

`TryAutoResolveMergeConflicts` in `internal/storage/versioncontrolops/mergesettle.go` is the dispatcher that inspects `dolt_conflicts` after a CLI pull and auto-resolves safe conflicts. It has a allowlist of four tables, each with its own safety predicate:

| Table | Safety check | Resolution |
|-------|-------------|------------|
| `metadata` | always convergent | `--theirs` |
| `dependencies` | same logical edge, audit-only diff (`dependencyConflictsAreAuditOnly`) | `--theirs` |
| `schema_migrations` | vintage-only (`schemaMigrationsConflictsAreVintageOnly`) | row-wise keep hash |
| `config` | memory-convergent keys (`configConflictsAreMemoryConvergent`) | `--theirs` |

Any other table — including `issues` — hits the `default` case and the entire pull bails: `return false, nil` ("not resolvable, leave for operator").

## The scenario

Two clones around a central remote. Clone A migrates v49→v54 and pushes. Clone B (lagging since June 21) upgrades, migrates, and pulls. The merge produces 36 conflicts in the `issues` table — every one a modify/modify where the remote side has a strictly newer `updated_at` (clone A had been actively working; clone B hadn't touched these issues since the last sync). All 36 required manual `dolt conflicts resolve --theirs`.

This is not a schema fork (#4259 is fixed). The deterministic migrations converge cleanly. It is a **data conflict** on the primary issue table — the one every clone writes to — and the auto-resolver has no case for it.

## Suggested fix

Add an `issues` case to `TryAutoResolveMergeConflicts`, following the existing per-table pattern:

1. **Safety predicate** (mirrors `dependencyConflictsAreAuditOnly`): only handle **modify/modify** conflicts (both sides have the row — `our_id` and `their_id` both valid). Reject add/add and delete/modify → leave for the operator.
2. **Resolution**: row-wise last-write-wins by `updated_at` — take whichever side has the strictly newer timestamp. If timestamps are equal, leave for the operator (ambiguous).
3. **Staging**: `DOLT_ADD('issues')` after resolution, same as the other tables.

This is **never less safe** than today: today every `issues` conflict bails to the operator regardless. The change only auto-resolves the provably-safe subset (one side strictly newer, both sides have the row).

The `issues` table is core to bd (bd owns the schema and maintains `updated_at` as a reliable timestamp it writes on every mutation), so this lives within the existing storage-boundary pattern — it's the same shape as `dependencies`' natural-identity check and `config`'s memory-key prefix check.

## Risk note: last-write-wins loses field-level changes from the older side

LWW on the **whole row** loses field-level changes from the losing side. Example: local updates `notes` at T1, remote updates `status` at T2 (T2 > T1) → the remote's entire row wins, and the local `notes` edit is silently lost.

This is the same convergent trade-off that `metadata` and `config` already make, and bd is primarily single-user-multi-machine (so concurrent edits to the same issue across machines are rare). But it should be stated explicitly for `issues` because issue field-level merge is more visible than metadata.

A possible mitigation for a follow-up: only auto-resolve when the losing side's `updated_at` predates the merge-base commit (i.e., the losing side hasn't touched the issue since the last successful sync). That narrows the loss window to "both sides edited since last sync" cases, which are rarer. This adds complexity and could be a separate enhancement on top of the basic LWW fix.

## Related

- #4259 — the original cross-clone merge fork (fixed by deterministic ids + migration 0050)
- #4516 — the smart remote-migrate gate (auto-resolves the *schema* side; this issue is the *data* side)
- #4551 — smart gate follow-up (non-destructive fast-forward adopt handles the non-divergent remote-ahead case; this issue handles the divergent case where conflicts actually exist)

maphew:
## Why

`bd dolt pull` bails to manual conflict resolution for the issues table even when every conflict is a provably-safe modify/modify where one side is strictly newer by `updated_at`. The existing auto-resolver (`TryAutoResolveMergeConflicts`) handles metadata, dependencies, schema_migrations, and config with table-specific safety checks, but issues hits the default case and dumps every conflict to the operator.

Observed 2026-07-10: after migrating v49→v54 and pulling, 36 issues-table conflicts required manual `dolt conflicts resolve --theirs`. Every one was modify/modify with the remote strictly newer by `updated_at`. This is the biggest remaining manual wall in the core multi-clone sync workflow.

Closes gastownhall/beads#4698.

## What

1. **Safety predicate** (`issuesConflictsAreLWWSafe`): only handle modify/modify (base row exists, both sides have the row). Reject add/add (no base row), delete/modify (one side deleted), and equal `updated_at` (ambiguous).
2. **Resolution** (`resolveIssuesLWWConflicts`): row-wise last-write-wins by `updated_at`. For rows where ours wins, delete the conflict row from `dolt_conflicts_issues` (the working set already holds our values — Dolt's manual-resolution path). For remaining rows (theirs wins), resolve with a single `DOLT_CONFLICTS_RESOLVE('--theirs', 'issues')`.
3. **Stage** with `DOLT_ADD('issues')` after resolution.

Never less safe than today: today every issues conflict bails regardless. The change only auto-resolves the provably-safe subset.

## Risk

LWW on the whole row loses field-level changes from the losing side (local updates notes at T1, remote updates status at T2 → remote row wins, local notes lost). **Mitigation implemented:** the safety predicate now also checks that the losing side's `updated_at` predates the merge base (equals the base row's `updated_at`). If the losing side also moved its `updated_at` past the base value, both sides made real edits since the last sync and the conflict is left for the operator. This preserves the common case (one side's conflict is a migration artifact — the row was rewritten by a schema change but `updated_at` did not move — and the other side made the only real edit) while declining the rarer concurrent-edit case where row-level LWW would silently drop field-level changes. See [follow-up comment](#issuecomment-4939374832).

## Tests

- `TestTryAutoResolveMergeConflicts_IssuesLWWTheirsWins` — theirs has newer `updated_at`, verify title matches theirs
- `TestTryAutoResolveMergeConflicts_IssuesLWWOursWins` — ours has newer `updated_at`, verify title matches ours
- `TestTryAutoResolveMergeConflicts_IssuesAmbiguousLeftAlone` — equal timestamps, verify unresolved
- `TestTryAutoResolveMergeConflicts_IssuesAddAddLeftAlone` — add/add, verify unresolved
- Updated `TestPullFromReportsOperatorConflicts` to pin `updated_at` to equal values (ambiguous → operator path still exercised)

### DIFF
--- a/internal/storage/dolt/federation_pull_settle_test.go
+++ b/internal/storage/dolt/federation_pull_settle_test.go
@@ -121,7 +121,9 @@ func TestPullFromSettlesFKCascadeViolations(t *testing.T) {
 // peerPullOutcome's post-hoc GetConflicts can never see it — the conflicts
 // must arrive captured pre-rollback inside MergeConflictsError and surface
 // through PullFrom's conflict-reporting contract. Peer setup mirrors
-// TestPullFromSettlesFKCascadeViolations.
+// TestPullFromSettlesFKCascadeViolations. Both sides' updated_at is pinned to
+// the same value so the LWW auto-resolver (GH#4698) declines the conflict as
+// ambiguous, leaving it for the operator.
 func TestPullFromReportsOperatorConflicts(t *testing.T) {
 	store, cleanup := setupConcurrentTestStore(t)
 	defer cleanup()
@@ -159,7 +161,7 @@ func TestPullFromReportsOperatorConflicts(t *testing.T) {
 	defer peerConn.Close()
 	peerConn.SetMaxOpenConns(1)
 	for _, q := range []string{
-		"UPDATE issues SET title = 'theirs' WHERE id = 'opcpeer-x'",
+		"UPDATE issues SET title = 'theirs', updated_at = '2026-01-01 00:00:00' WHERE id = 'opcpeer-x'",
 		"CALL DOLT_COMMIT('-Am', 'peer retitles')",
 		"CALL DOLT_PUSH('origin', 'main')",
 	} {
@@ -168,7 +170,11 @@ func TestPullFromReportsOperatorConflicts(t *testing.T) {
 		}
 	}
 
-	if _, err := db.ExecContext(ctx, "UPDATE issues SET title = 'ours' WHERE id = 'opcpeer-x'"); err != nil {
+	// Pin updated_at to the same value as the peer's update so the LWW
+	// safety check (issuesConflictsAreLWWSafe) rejects it as ambiguous —
+	// this test exercises the operator-conflict reporting path, which
+	// requires a conflict the auto-resolver declines (GH#4698).
+	if _, err := db.ExecContext(ctx, "UPDATE issues SET title = 'ours', updated_at = '2026-01-01 00:00:00' WHERE id = 'opcpeer-x'"); err != nil {
 		t.Fatalf("retitle locally: %v", err)
 	}
 	if _, err := db.ExecContext(ctx, "CALL DOLT_COMMIT('-Am', 'local retitles')"); err != nil {
--- a/internal/storage/dolt/issues_conflict_test.go
+++ b/internal/storage/dolt/issues_conflict_test.go
@@ -2,7 +2,7 @@ package dolt
 
 import "testing"
 
-func setupIssueMergeConflict(t *testing.T, issueID, ourTitle, ourUpdatedAt, theirTitle, theirUpdatedAt string, seed bool) (*DoltStore, string) {
+func setupIssueMergeConflict(t *testing.T, issueID, baseTitle, baseUpdatedAt, ourTitle, ourUpdatedAt, theirTitle, theirUpdatedAt string, seed bool) (*DoltStore, string) {
 	t.Helper()
 	store, cleanup := setupTestStore(t)
 	t.Cleanup(cleanup)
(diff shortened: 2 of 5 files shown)


maphew:
## Follow-up: merge-base-predates mitigation (pushed)

The Risk section flagged a possible follow-up: LWW on the whole row loses field-level changes from the losing side when **both** sides edited the same issue since the last sync. I've implemented that mitigation in `579fd7ef9`.

### What changed

`issuesConflictsAreLWWSafe` now also reads `base_updated_at` from `dolt_conflicts_issues` and, after the existing modify/modify + unequal-timestamp checks, verifies the **losing side's** (older) `updated_at` does not exceed the base value:

- If the losing side's `updated_at` == base `updated_at` → the losing side hasn't touched the issue since the last sync (its "modification" is a migration artifact, e.g. a schema change rewrote the row without moving `updated_at`) → **auto-resolve by LWW** (the one real edit wins).
- If the losing side's `updated_at` > base `updated_at` → both sides made real edits since the merge base → **left for the operator** (row-level LWW would silently drop the losing side's field-level changes).

`resolveIssuesLWWConflicts` is unchanged — it already does row-wise LWW; only the safety gate is stricter.

### Why this preserves the original fix's value

The observed real-world case (36 conflicts after v49→v54 migration + pull) is exactly the migration-artifact case: clone B's rows were rewritten by schema migrations but `updated_at` didn't move, while clone A's active edits did. Those still auto-resolve. The mitigation only declines the rarer concurrent-edit case (both clones actively edited the same issue since the last sync), which is the case where silent field-level loss is a real risk.

### Tests

- `setupIssueMergeConflict` now seeds a controlled base `updated_at` (explicit instead of `CURRENT_TIMESTAMP`, which was timing-dependent).
- Existing theirs-wins / ours-wins tests set the losing side's `updated_at` equal to the base → still auto-resolve.
- New `TestTryAutoResolveMergeConflicts_IssuesConcurrentEditLeftAlone`: both sides' `updated_at` > base → left for the operator.

_amp-gpt-5.4-2026-03-05-high on behalf of maphew_

PR #4712, opened 2026-07-10
fix(security,ci): redact all DSN password forms in telemetry; require pg/mysql conformance (post-merge #4601)

julianknutsen:
## Post-merge remediation for #4601

Follow-up PR for the post-merge review of #4601 ("choose your own storage
backend"). Reviewed landed range `812ad6d..3f0474a`. This PR fixes the two
gating findings the post-merge review surfaced; it does not re-land #4601.

### 1. Security — telemetry leaks Postgres/MySQL passwords (major)

`scrubArgsForTelemetry` redacted only `user:PASS@host` URL userinfo, so
`bd init --backend=postgres --pg-url ...` leaked the password into the
`bd.args` telemetry span for two accepted DSN shapes:

- URL query params — `?password=` / `?sslpassword=`
- libpq keyword/value — `password=` / `sslpassword=`

Fix: route `--pg-url` / `--mysql-url` values (both `--flag=value` and
`--flag value`) through the parser-backed pgdialect scrubber, exposed as a new
`pgdialect.ScrubDSNString` (the primitive `ScrubDSNError` already used). Every
other arg keeps the narrow userinfo scrub, so ordinary text is never
over-redacted.

### 2. CI — pg/mysql conformance can silently self-skip to green (major)

The embedded-Dolt reference step hard-fails if its suite skips, but the
Postgres/MySQL Tier-1 steps in `scripts/conformance.sh` self-skipped when
`BEADS_*_TEST_URL` was unset and CI never set `BEADS_CONFORMANCE_REQUIRE`, so
the storage-parity gate could report green with reduced backend coverage.

Fix: add `require_backend` + `run_sql_conformance` that mirror the embedded-Dolt
top-level `TestConformance` skip/PASS detector for pg/mysql, and set
`BEADS_CONFORMANCE_REQUIRE=postgres,mysql` in `conformance.yml` (which also arms
the pre-existing Tier-2 e2e hard-fail).

### Tests / verification

- `pgdialect.TestScrubDSNString` — every password form is removed; passwordless
  and empty DSNs pass through.
- `cmd/bd.TestScrubArgsForTelemetry` — all DSN password forms across both flag
  spellings; `TestScrubArgsForTelemetryLeavesOrdinaryArgs` proves no
  over-redaction plus userinfo defense-in-depth.
- `scripts/conformance.sh` helpers unit-tested, and `run_sql_conformance`
  integration-tested against real `go test`: not-required self-skips (exit 0),
  required-but-absent hard-fails on the detected top-level `--- SKIP` (exit 1).
- `gofmt`, `go vet`, `golangci-lint` (0 issues), and full `go build ./...` clean.

Non-gating review follow-ups (Postgres collation enforcement, bounded pg/mysql
pool limits, SQLite one-target dependency `CHECK`, extra conformance coverage)
are intentionally out of scope for this remediation.


maphew (commented):
The required-backend conformance gate and the broad DSN redaction coverage are valuable. There is one security blocker, though: percent-encoded password query-key spellings bypass the new scrubber.

`pgx` accepts query keys after URL decoding, so `postgres://u@h/db?pass%77ord=TOPSECRET` (and `sslpass%77ord`) configures a password. `dsnPasswordValues` currently compares the raw key only, so it returns no value for those valid aliases; the raw `--pg-url` is then emitted in `bd.args` telemetry, and userinfo scrubbing does not apply.

Please decode or otherwise canonicalize query keys before the password-key comparison, and add regressions through both `ScrubDSNString` and `scrubArgsForTelemetry` for encoded query keys. PR CI is green, but upstream `main` is red, so refresh/recheck after the base fix before merging.

maphew:
Maintainer push (4b56677e8): closes the encoded-query-key blocker from the 2026-07-10 review, plus four more bypasses a paired cross-vendor review (Claude + Codex) surfaced. All five were reproduced test-first against `pgx.ParseConfig` before fixing:

1. **Percent-encoded query keys** — `pass%77ord=`/`%70assword=` are live password params after `url.Query()` decoding; `dsnPasswordValues` now compares the unescaped key as well as the raw one.
2. **pflag shorthand clusters** — `-qpSECRET` parses as `-q -p SECRET` but the raw token reached telemetry; the shorthand scan now walks the leading letter run (over-redacts, never under-redacts).
3. **Escaped whitespace in unquoted libpq values** — `password=SUPER\ SECRET` leaked ` SECRET`; the value pattern now consumes backslash-escaped sequences.
4. **Vertical tab** — pgx treats `\v` as DSN whitespace, Go regexp `\s` does not; the whitespace class is now spelled out explicitly.
5. **Overlapping secrets** — a short secret that prefixes a longer one mangled the longer before its replacement pass, leaking the tail; `ScrubDSNString` now replaces longest-first.

Regression cases added for every form in `redact_error_test.go` and `telemetry_scrub_test.go`; `gofmt`/`go vet` clean, pgdialect + cmd/bd scrub suites green locally. Your commits are intact — this only extends your redaction machinery in the direction the PR already established. Nice remediation work on the conformance gate, julianknutsen.

maphew (approved):
Approving. Review history: independent cross-vendor review (Claude + Codex) confirmed the PR's core redaction machinery and conformance gating are sound, surfaced five bypasses, and all five were fixed test-first in 4b56677e8 (details in the earlier comment). Checks are green on the branch refreshed onto current main. The remaining known limitation (over-redaction when a password equals a structural substring) is cosmetic-only and documented in review.

PR #4713, opened 2026-07-10
fix(label): support comma-separated multi-label add/remove; fail hard on bad issue IDs

steveyegge:
## Problem (bd-vu5kv)

`bd label add <id> a b c` silently applies only `c`. The grammar is `add [issue-id...] [label]` — the last arg is the only label, and everything before it is treated as an issue ID. `a` and `b` fail to resolve, print a stderr warning, get **skipped**, and the command exits 0. A Fable agent hit this in the wild and only caught it because `bd ready --label program:castellan` came back empty.

The warn-and-continue also meant a plain typo'd issue ID in a batch call could exit 0 having labeled nothing.

## Fix

1. **Comma-separated multi-label**: the final positional arg now splits on commas, so `bd label add bd-123 a,b,c` adds all three labels in one transaction. Matches the existing `bd create --labels a,b,c` convention. Same for `bd label remove`.
2. **Fail hard on unresolvable issue IDs**: resolution failure is now an error *before any mutation* (resolution happens ahead of the transaction, so nothing is partially applied). When multiple ID-position args fail — the space-separated-labels mistake — the error hints at the comma form:
   ```
   Error: resolving issue ID "a": no issue found matching "a" (to add multiple
   labels, pass one comma-separated argument: bd label add <issue-id> label1,label2)
   ```

Space-separated positional labels can't be made to Just Work: ID resolution is fuzzy (partial hashes, prefix-less forms), so guessing the ID/label boundary risks applying labels to the wrong issue. Loud error + hint is the safe behavior.

The reserved `provides:` check now runs against every label in the list.

## Behavior changes

- Batch calls with an unresolvable ID now exit non-zero and apply nothing (previously: warn, skip, exit 0).
- A label containing a literal comma can no longer be added/removed positionally — already impossible via `bd create --labels`, which splits on commas.

## Testing

- New parser unit tests (comma splitting, whitespace, empties).
- 4 new embedded end-to-end tests, including a regression test for the exact reported invocation (`add <id> space-a space-b space-c` → hard error, comma hint, zero labels applied). Full `TestEmbeddedLabel` suite: 22/22 pass.
- Full `go test ./cmd/bd/` suite green; `golangci-lint` clean.
- CLI reference regenerated with `scripts/generate-cli-docs.sh` (pinned pure-go build) for the doc-freshness probe; docs/LABELS.md updated.

Closes bd-vu5kv

https://claude.ai/code/session_012VBUR4qkHbssPomBDuJgkQ

### DIFF
--- a/cmd/bd/label.go
+++ b/cmd/bd/label.go
@@ -4,7 +4,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"os"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


maphew (commented):
The implementation and regression coverage address the silent partial-success path cleanly: IDs now resolve before the transaction, and multi-label parsing follows the existing comma-separated convention.

One merge blocker remains in the required checks. `Check doc flags freshness` reports that the `docs/LABELS.md` additions have not been mirrored to `website/docs/core-concepts/labels.md` and `website/versioned_docs/version-1.1.0/core-concepts/labels.md`. Running `./scripts/sync-website-docs.sh` and committing those two mirrors should clear the failure. Upstream `main` is also red, so refresh/recheck after the base fix lands.

maphew:
Pushed a maintainer fix for the CI-only documentation drift on this PR: the current and v1.1.0 label-page mirrors plus the generated `website/static/llms-full.txt` are now synchronized in `aadc1b635`.

Local `scripts/check-website-docs-drift.sh` passes, and the refreshed `Check doc flags freshness` job is green. The full CI matrix is still running on that exact head; merge should wait for its required gate.

PR #4715, opened 2026-07-10
feat(ownership): holder_token + advisory enforcement mode

julianknutsen:
## What

Binds a claim to an incarnation-unique **holder token** so a stale process cannot pass as the current owner of a claim whose assignee name was re-used, and adds an **advisory** enforcement mode that classifies non-holder mutations for observability.

- Migration v56: `holder_token VARCHAR(64)` on `issues` and `wisps`, recorded at claim from the caller's ambient `BEADS_HOLDER_TOKEN` (via a context transport, like `WithLeaseTTL`/`WithGuard`).
- The token is **enforcement-only state**: kept out of the issue read surface (`IssueSelectColumns`, JSON) **and** out of event `old_value` — a worker that could read the current token via `bd show`/`bd history` could just present it. Guard tests pin its absence from both surfaces.
- Cleared on unclaim, reclaim, ownership-transition update, and import assignee-change (issueops + proxied `domain/db`); an idempotent same-actor re-claim refreshes it so a fresh incarnation owns its own token.
- `claims.enforcement` config (`off` default | `advisory`): under `advisory`, an in-place mutation of a claimed row by other than the clean holder emits an `ownership_advisory` event classed `cross_actor_infra` / `empty_token_legacy` / `tokenless_caller` / `token_mismatch` (the genuine two-incarnation zombie). Advisory never blocks.

## Why

A claim's assignee is a reusable display name; two processes with the same name are indistinguishable to beads. The token distinguishes runtime incarnations, so a stale process cannot pass as the current owner. `token_mismatch` is reserved strictly for two token-bearing incarnations disagreeing, so a later slice's require-mode flip can gate on a clean zombie count. This slice ships the column, recording, and advisory telemetry only — require-mode enforcement and the fenced transfer verb are deliberately deferred to gate on the advisory soak this produces.

Design context (in the consuming fork): `engdocs/plans/ownership-fencing/DESIGN.md` §2.1/§2.3.

## Stacking and scope (please read)

- **Stacked on the tier-complete lease work (#4697), which is stacked on #4675.** This branch is based on `feat/claim-fence`; since those are fork/feature branches that cannot be set as a GitHub base here, this PR targets `main` and its diff currently includes the commits from #4675 and #4697. **Review only the single `feat(ownership): holder_token…` commit at the tip**; the diff cleans up once the parents merge and this rebases.
- **Advisory only.** Require-mode enforcement (blocking) and the `bd claim transfer` verb are out of scope (a follow-up gated on advisory-soak data); the reserved sentinel `!` is defined but not yet wired.
- **Migration numbering:** takes `0056_add_holder_token` on top of `0055_add_claim_fence` (from #4697). When all three land in order the sequence is contiguous.

## Verification

Beads-only (real Dolt + conformance), no orchestrator:

```
# holder-token recording/clearing + never-surfaced (issue JSON and events) + advisory classes
go test ./internal/storage/dolt/ -run 'Holder|Advisory' -timeout 20m

# hidden-surface guard (holder_token stays out of the read columns)
go test ./internal/storage/issueops/ -run TestHolderTokenNotInSelectColumns

# regression: fence bumps, guarded verbs, lease/reclaim unaffected
go test ./internal/storage/dolt/ -run 'Fence|Guard|Reclaim|Unclaim|Lease' -timeout 25m

go test ./internal/storage/issueops/ ./internal/storage/domain/db/ ./internal/storage/schema/
golangci-lint run ./...
make check-docs
```

The never-surfaced invariant is enforced by tests against **both** the issue read surface and the events surface (a pre-commit review caught and this fixes an `old_value` token leak; the class in `new_value` carries the full telemetry).


julianknutsen:
**Review pipeline: parked pending author rebase.** Automated review attempts (rebase_gate_conflict (lease-model architecture conflict vs main)) fail in preflight because this branch cannot be cleanly rebased/merged onto current main. Parking to stop restart churn. To resume: rebase onto main, push, and re-add the `status/needs-review-auto` label.

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace through the pinned Dolt transaction implementation; Claude adjudication re-verified the migration collision at the PR head). Verdict: **needs changes** — ownership fencing is a real need for fleets, but the design has been overtaken by main's lease restructuring and has two synchronization holes.

**Blocker 1 — migrations collide AND target a schema main removed.** Verified: this PR's 0055/0056 collide with main's (main owns 0055-0058, and the loader panics on duplicates at store open). Renumbering isn't enough: main's 0055 deliberately moved leases out of `issues`/`wisps` into the clone-local `leases` table (to stop version-history churn), while this PR's lease SQL still writes the removed columns — including leasing wisps, which main explicitly excludes. The fence/holder-token design needs rebuilding against the `leases` table.

**Blocker 2 — the "hidden" holder credential is readable via main's raw-SQL surface.** Since the branch point, main added `bd sql` (local and proxied). `SELECT holder_token FROM issues` recovers another incarnation's token and defeats the token-mismatch fence. Either tokens aren't credentials, or they need an access boundary raw-SQL can't reach.

**Blocker 3 — advisory telemetry can't support the safety conclusion.** At the head, `WithHolderToken` exists only on claim paths; ordinary update/close use unenriched contexts and the proxied update path records nothing — so token-bearing workers classify as `tokenless_caller`, stale tokens never classify as `token_mismatch`, and `--claim`-then-update generates false telemetry immediately. The soak data the PR proposes to collect would be systematically wrong.

**Blocker 4 — `lease disarm` is timing-based, not synchronized.** A claim transaction that snapshots `lease.auto=on` before the flip can commit an armed lease after disarm's bounded sweeps return success — the pinned Dolt three-way-merges disjoint config/issue writes without conflict. Needs an epoch/fence or claim-side post-flip validation.

**Should-fixes:** proxied claim paths accept and silently discard `--lease-ttl` (and drop the token context — same proxied-parity gap flagged across this sweep); and the `unclaim --if-assignee` guard conflicts with the contract main already landed (registered flag, empty-value rejection, `--force` exclusivity, exit 1) — reconcile rather than replace, and beware duplicate Cobra registration on rebase.

PR #4718, opened 2026-07-11
fix(lease): stop arming worker leases on bare status/assignee updates

### LINKED ISSUE #4716: ManageLeaseOnUpdate arms a worker lease on plain 'bd update -s in_progress -a X', making interactive claims reclaim-bait
## Summary

`issueops.ManageLeaseOnUpdate` arms a fresh TTL lease (`lease_expires_at = now + TTL`) whenever a generic `bd update` leaves an issue `in_progress` with a non-empty assignee. For worker/automation flows that heartbeat, that's coherent — but an **interactive** claim (`bd update <id> -a <me> -s in_progress`, the documented alternative to `--claim` for humans/resident agents) never heartbeats, so the lease silently expires and `bd reclaim` reverts the human's live work to open+unassigned.

## Observed impact (production multi-agent constellation, bd 1.1.0 @ 86ccd7ce7)

- Every interactively-claimed issue cycled **stale-lease → reclaim revert → supervisor restore → fresh lease (from the restore update itself) → revert** every ~15 min (5m TTL + 10m reclaim grace). Six crew-held issues churned this way for a full day (36+ restore events), each restore appending an audit comment (comment spam) and each cycle opening a seconds-wide open+unassigned window in which a polling worker's `bd ready --claim` legitimately grabbed a human's in-flight issue — a fleet worker duplicated ~2h of a resident agent's work this way.
- There is **no way** to hold an issue `in_progress`+assigned WITHOUT a live lease: any update that would 'fix' it re-arms one. Downstream we now proxy-heartbeat interactive claims from our supervisor (`bd --actor <assignee> heartbeat <id>` — works because heartbeat's owner check is `assignee == actor`), but that's a workaround.

## Suggested direction (any of)

1. Only claim/heartbeat paths arm leases; a manual `bd update` that sets in_progress+assignee should NOT (or should clear an existing lease — manual state supersedes lease machinery).
2. An explicit opt-in, e.g. `bd update --lease`, for automations that want the lease armed via update.
3. `bd reclaim` skips issues whose assignee never heartbeated on the current lease (lease armed by update, heartbeat_at == lease-arm time).

## Related

- `ReclaimedLease.PreviousOwner` is `COALESCE(assignee,'')` — the assignee at revert time, not the lease holder. Fine for our restore heuristics (that's what saved assignee-carrying claims), but the field name is misleading when the lease was armed by an update actor different from the assignee.

steveyegge:
Fixes #4716 (local bead: bd-9hpgf, P1).

## Problem

`ManageLeaseOnUpdate` armed a 5-minute worker lease whenever a generic update left an issue `in_progress` + assigned — it gated on the **resulting state**, not the **verb**. So a plain interactive claim (`bd update -s in_progress -a <who>`) silently entered the heartbeat/reclaim protocol that nobody was running for it:

1. Interactive session claims a bead by hand — lease armed, no one heartbeats.
2. Lease lapses; `bd reclaim` reverts the bead to open while the owner keeps working.
3. Supervisor restores the assignment — which re-arms the lease, making the churn self-sustaining (~15 min revert/restore cycles across every crew-held bead).
4. Eventually a headless fleet worker legitimately claims a reverted bead whose owner is mid-implementation → duplicate work, force-close + collision cleanup.

Observed live on the wyvern shared-dolt constellation (their wy-nx8pt); every multi-agent deployment mixing interactive and fleet claims would hit it.

## Fix

Leases now arm **only** via the lease-aware verbs — `ClaimIssueInTx` (`bd update --claim`, `bd ready --claim`) and heartbeat. `ManageLeaseOnUpdate` is clear-only:

- **Leaving the claimed state** (not in_progress, or unassigned): clears the lease, as before.
- **Changing who holds the claim** (assignee transfer, or a fresh transition into in_progress): clears the previous owner's lease so it can't count down against the new holder. The new holder gets a lease only by claiming/heartbeating through the lease-aware verbs.
- **Same-claim edit** (already in_progress, same assignee): lease columns untouched — a worker's live lease survives unrelated edits to its issue.

This also defuses the supervisor-restore churn amplifier: a restore via `bd update` no longer re-arms anything. Real workers restored that way self-heal — their next heartbeat re-arms the lease (heartbeat only requires `in_progress` + owner match, not an existing lease).

`updateIssueInTx` still rewrites `row_lock` on every mutation, so the reclaim/heartbeat conflict invariant is unchanged. The domain/db backend never managed leases on update (pre-existing parity gap tracked in bd-6dnrw.44), so no changes there.

## Tests

- `TestBareUpdateClaimDoesNotArmLease` (new, real Dolt): the bd-9hpgf repro — hand-dole claim carries no lease and survives a reclaim sweep whose cutoff would reap any leased issue.
- `TestUpdatePreservesLeaseOnSameClaim` (new, real Dolt): a same-claim edit doesn't disarm a worker's live lease.
- `TestUpdateIssueMaintainsLeaseOwnership` (flipped): previously pinned the buggy transfer-stamps-fresh-lease behavior; now asserts transfer clears the lease, old owner can't heartbeat, new owner re-arms via heartbeat.
- `TestManageLeaseOnUpdate` (new, pure unit table-test): pins the clear-only contract — a generic update never emits lease-arming SQL.

Verified end-to-end with the built binary: bare `bd update -s in_progress -a crow` leaves `lease_expires_at`/`heartbeat_at` NULL and `bd reclaim` leaves it alone; `bd update --claim` still stamps the 5-minute lease.

Full `internal/storage` (non-dolt) suite passes; dolt-package and `cmd/bd` suites show only failures identical to `origin/main` baselines run on the same machine (known container-contention flakes + pre-existing failures, lists diffed by name).

https://claude.ai/code/session_01U5mtSXJuNGs4tmeKDbwNif

### DIFF
--- a/internal/storage/dolt/lease_test.go
+++ b/internal/storage/dolt/lease_test.go
@@ -137,6 +137,9 @@ func TestUpdateIssueMaintainsLeaseOwnership(t *testing.T) {
 	seedClaimedIssue(t, ctx, store, "lease-update", "alice", time.Hour)
 	before := readLeaseState(t, ctx, store, "lease-update")
 
+	// An assignee transfer through generic update clears the old owner's lease
+	// rather than stamping a fresh one: the new holder did not claim through the
+	// lease-aware verb, so their hold is durable until they opt in (bd-9hpgf).
 	if err := store.UpdateIssue(ctx, "lease-update", map[string]interface{}{"assignee": "bob"}, "dispatcher"); err != nil {
 		t.Fatalf("transfer assignee: %v", err)
 	}
@@ -144,17 +147,22 @@ func TestUpdateIssueMaintainsLeaseOwnership(t *testing.T) {
 	if transferred.assignee.String != "bob" {
 		t.Fatalf("assignee after transfer = %q, want bob", transferred.assignee.String)
 	}
-	if !transferred.leaseExpires.Valid || !transferred.heartbeatAt.Valid {
-		t.Fatalf("transfer did not stamp a fresh lease: %+v", transferred)
+	if transferred.leaseExpires.Valid || transferred.heartbeatAt.Valid {
+		t.Fatalf("transfer should clear the old owner's lease, got %+v", transferred)
 	}
-	if !transferred.heartbeatAt.Time.After(before.heartbeatAt.Time) && transferred.rowLock == before.rowLock {
-		t.Fatalf("transfer did not refresh heartbeat or row_lock: before=%+v after=%+v", before, transferred)
+	if transferred.rowLock == before.rowLock {
+		t.Fatalf("transfer did not rewrite row_lock: before=%+v after=%+v", before, transferred)
 	}
 	if err := store.HeartbeatIssue(ctx, "lease-update", "alice"); !errors.Is(err, storage.ErrAlreadyClaimed) {
 		t.Fatalf("old owner heartbeat err = %v, want ErrAlreadyClaimed", err)
 	}
+	// The new owner can opt back into lease semantics by heartbeating.
 	if err := store.HeartbeatIssue(ctx, "lease-update", "bob"); err != nil {
 		t.Fatalf("new owner heartbeat: %v", err)
+	}
+	rearmed := readLeaseState(t, ctx, store, "lease-update")
+	if !rearmed.leaseExpires.Valid || !rearmed.heartbeatAt.Valid {
+		t.Fatalf("heartbeat by new owner should re-arm the lease, got %+v", rearmed)
 	}
 
 	if err := store.UpdateIssue(ctx, "lease-update", map[string]interface{}{"status": string(types.StatusOpen)}, "dispatcher"); err != nil {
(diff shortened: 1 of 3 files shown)


maphew (approved):
The clear-only lease update semantics are correct and well covered. Generic updates now clear stale ownership leases without arming new ones, while same-owner/same-status edits preserve a live worker lease; claim and heartbeat remain the only arming paths. The unconditional `row_lock` rewrite still preserves heartbeat/reclaim conflict behavior.

Freshness review found no exact duplicate. This should land before the broader ownership work in #4697; that branch must rebase and retain this clear-only behavior. #4675 should also refresh after this merge, and #4493 is stale as an as-is merge candidate.

Verification on head `39bed11aa5624`:

```text
go test ./internal/storage/issueops/...
go test -tags gms_pure_go -run 'Test(BareUpdateClaimDoesNotArmLease|UpdatePreservesLeaseOnSameClaim|UpdateIssueMaintainsLeaseOwnership)$' ./internal/storage/dolt
```

Both passed locally. All GitHub checks and the blocking base-health preflight are green.

PR #4720, opened 2026-07-11
fix(mol): route bond operand resolution like show/update/close (closes #4714)

### LINKED ISSUE #4714: bd mol bond ignores prefix routing: routed-prefix issue ID not found (works in bd show)

R0SEWT:
## What

`bd mol bond <A> <B>` now resolves operands through the same local → prefix route → auto-route fallback used by other issue commands. Bonds run in the validated owning store, while unsupported cross-store and contributor-planning-store mutations fail before any writable foreign-store open.

## Why

Fixes #4714. `mol bond` previously resolved only against the local store, so `bd show <id>` could find a routed issue that `bd mol bond <formula> <id>` rejected. This breaks the default `gt sling` path.

The final design also incorporates the two applicable safety ideas from the older #4350 implementation: discover routed homes read-only before reopening one writable, and preserve actionable failures after a prefix route matches. Johan Snyman is credited as co-author on the maintainer integration commit.

## How

- Discover both operands read-only, with unconditional formula fallback so valid plain formulas and formulas colliding with ambiguous issue prefixes still work.
- Apply the same discovery and store-policy validation to `--dry-run` and execution.
- Compare logical stores using the existing `dependencyStoreKey`, which unwraps hook/telemetry decorators before consulting `StoreLocator`.
- Reject genuinely cross-store bonds and contributor auto-routed mutations before opening a foreign store writable.
- After validation, close read handles, reopen the accepted prefix/maintainer/default route writable, and verify its logical identity has not changed.
- Run routed mutations through the normal storage decorator chain so hooks and telemetry remain active.
- Distinguish “no matching route” from a matched route whose metadata or store open failed; the latter error is returned instead of being collapsed into “not found.”

The proxied-server `mol bond` path remains local-only and is intentionally outside this PR’s direct-store routing change.

## Verification

Focused cgo routing tests pass:

- `TestDiscoverMolBondOperandRouting`
- `TestDiscoverMolBondOperandFormulaWinsAmbiguousIssuePrefix`
- `TestMolBondPrefixRoutedWriteThrough`
- `TestSameBondStoreSeparateHandles`
- `TestMolBondAutoRoutedReadOnlyFailsFast`

Existing routing compatibility tests also pass:

- `TestResolveCloseTargets`
- `TestDepRoutedTargetOpensReadOnly`
- `TestDepListCrossRigRouting`

The affected package compiles under the pure-Go `gms_pure_go` configuration. A fresh cross-vendor review found no remaining code issues; the full local suite is queued through the repository verifier.

---

Original implementation and tests by Rody Vilchez; maintainer integration with Johan Snyman attribution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -242,7 +242,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   message before any mutation is attempted. Two operands that live in the same
   routed rig share one store handle and bond normally; only a bond whose
   operands resolve to genuinely different databases is rejected rather than
-  written to the wrong one. Same routing-parity fix as
+  written to the wrong one. As part of this change, maintainer/default
+  auto-routed stores now open writable for any validated write-intent resolver
+  (`bd update`/`bd close` target resolution included, not just `mol bond`);
+  contributor auto-routes remain read-only and mutation-forbidden. Same
+  routing-parity fix as
   [#3608](https://github.com/gastownhall/beads/issues/3608) for `bd close`.
   Fixes [#4714](https://github.com/gastownhall/beads/issues/4714).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -236,9 +236,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   against the local store only, so `gt sling <bead> <rig>` — which cooks the
   default formula and bonds it to a rig-routed bead — died at `bd mol bond` even
   though `bd show <bead>` resolved the same ID. Resolution is write-intent: a
-  prefix-routed target opens writable and the bond commits where the bead lives.
-  A bond whose two operands resolve to different stores is rejected rather than
-  written to the wrong database. Same routing-parity fix as
+  prefix-routed target opens writable and the bond commits where the bead lives,
+  while a bond pinned to the contributor auto-routed store — which always opens
+  read-only because it hydrates a foreign project — fails fast with a clear
+  message before any mutation is attempted. Two operands that live in the same
+  routed rig share one store handle and bond normally; only a bond whose
+  operands resolve to genuinely different databases is rejected rather than
+  written to the wrong one. Same routing-parity fix as
   [#3608](https://github.com/gastownhall/beads/issues/3608) for `bd close`.
   Fixes [#4714](https://github.com/gastownhall/beads/issues/4714).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -228,6 +228,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   emphasis (`**bold**`, `*italic*`), `SELECT * FROM t`, code spans and fenced
   blocks all render exactly as before.
 
+- **`bd mol bond <A> <B>` ignored prefix routing** — operand resolution now uses
+  the same routing fallback as `bd show`/`bd update`/`bd close`, so a bead that
+  lives in a routed store (a prefix-routed rig via `routes.jsonl`, or the
+  contributor planning store) resolves instead of failing with `'<id>' not found
+  as issue or formula`. Previously `bd mol bond` resolved operands
+  against the local store only, so `gt sling <bead> <rig>` — which cooks the
+  default formula and bonds it to a rig-routed bead — died at `bd mol bond` even
+  though `bd show <bead>` resolved the same ID. Resolution is write-intent: a
+  prefix-routed target opens writable and the bond commits where the bead lives.
+  A bond whose two operands resolve to different stores is rejected rather than
+  written to the wrong database. Same routing-parity fix as
+  [#3608](https://github.com/gastownhall/beads/issues/3608) for `bd close`.
+  Fixes [#4714](https://github.com/gastownhall/beads/issues/4714).
+
 ### Documentation
 
 - **The heartbeat/re-home invariant and the two states it can strand are now
--- a/cmd/bd/mol_bond.go
+++ b/cmd/bd/mol_bond.go
@@ -12,7 +12,6 @@ import (
 	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/types"
 	"github.com/steveyegge/beads/internal/ui"
-	"github.com/steveyegge/beads/internal/utils"
 )
 
 var molBondCmd = &cobra.Command{
@@ -110,11 +109,11 @@ func runMolBond(cmd *cobra.Command, args []string) error {
 	}
 
 	if in.dryRun {
-		issueA, formulaA, err := resolveOrDescribe(ctx, store, in.argA, in.vars)
+		issueA, formulaA, err := resolveOrDescribeWithRouting(ctx, store, in.argA, in.vars)
 		if err != nil {
 			return HandleErrorRespectJSON("%v", err)
 		}
-		issueB, formulaB, err := resolveOrDescribe(ctx, store, in.argB, in.vars)
+		issueB, formulaB, err := resolveOrDescribeWithRouting(ctx, store, in.argB, in.vars)
 		if err != nil {
 			return HandleErrorRespectJSON("%v", err)
 		}
@@ -122,38 +121,56 @@ func runMolBond(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	subgraphA, cookedA, err := resolveOrCookToSubgraph(ctx, store, in.argA, in.vars)
+	opA, err := resolveMolBondOperand(ctx, store, in.argA, in.vars)
 	if err != nil {
 		return HandleErrorRespectJSON("%v", err)
 	}
-	subgraphB, cookedB, err := resolveOrCookToSubgraph(ctx, store, in.argB, in.vars)
+	defer opA.Close()
+	opB, err := resolveMolBondOperand(ctx, store, in.argB, in.vars)
+	if err != nil {
+		return HandleErrorRespectJSON("%v", err)
+	}
+	defer opB.Close()
+
+	// The bond mutation runs against the store that owns the operands. When an
+	// operand was resolved via routing (e.g. a prefix-routed rig bead), that is
+	// the routed store, not the local one — otherwise the bond would write to
+	// the wrong database or fail on a cross-store reference (mirrors #3609).
+	activeStore, err := pickBondStore(store, opA, opB)
 	if err != nil {
 		return HandleErrorRespectJSON("%v", err)
 	}
 
+	// No cleanup needed - in-memory subgraphs don't pollute the DB
+	subgraphA, cookedA := opA.subgraph, opA.cooked
+	subgraphB, cookedB := opB.subgraph, opB.cooked
 	issueA := subgraphA.Root
 	issueB := subgraphB.Root
 	aIsProto := issueA.IsTemplate || cookedA
 	bIsProto := issueB.IsTemplate || cookedB
 
+	// Dispatch based on operand types
+	// All operations use activeStore; phase flags determine ephemeral vs persistent.
 	var result *BondResult
 	switch {
 	case aIsProto && bIsProto:
-		result, err = bondProtoProto(ctx, store, issueA, issueB, in.bondType, in.customTitle, actor)
+		// Compound protos are templates - always persistent
+		// Note: Proto+proto bonding from formulas is a DB operation, not in-memory
+		result, err = bondProtoProto(ctx, activeStore, issueA, issueB, in.bondType, in.customTitle, actor)
 	case aIsProto && !bIsProto:
 		if cookedA {
-			result, err = bondProtoMolWithSubgraph(ctx, store, subgraphA, issueA, issueB, in.bondType, in.vars, in.childRef, actor, in.ephemeral, in.pour)
+			result, err = bondProtoMolWithSubgraph(ctx, activeStore, subgraphA, issueA, issueB, in.bondType, in.vars, in.childRef, actor, in.ephemeral, in.pour)
 		} else {
-			result, err = bondProtoMol(ctx, store, issueA, issueB, in.bondType, in.vars, in.childRef, actor, in.ephemeral, in.pour)
+			result, err = bondProtoMol(ctx, activeStore, issueA, issueB, in.bondType, in.vars, in.childRef, actor, in.ephemeral, in.pour)
 		}
 	case !aIsProto && bIsProto:
 		if cookedB {
-			result, err = bondProtoMolWithSubgraph(ctx, store, subgraphB, issueB, issueA, in.bondType, in.vars, in.childRef, actor, in.ephemeral, in.pour)
+			result, err = bondProtoMolWithSubgraph(ctx, activeStore, subgraphB, issueB, issueA, in.bondType, in.vars, in.childRef, actor, in.ephemeral, in.pour)
 		} else {
-			result, err = bondMolProto(ctx, store, issueA, issueB, in.bondType, in.vars, in.childRef, actor, in.ephemeral, in.pour)
+			result, err = bondMolProto(ctx, activeStore, issueA, issueB, in.bondType, in.vars, in.childRef, actor, in.ephemeral, in.pour)
 		}
 	default:
-		result, err = bondMolMol(ctx, store, issueA, issueB, in.bondType, actor)
+		result, err = bondMolMol(ctx, activeStore, issueA, issueB, in.bondType, actor)
 	}
 	if err != nil {
 		return HandleErrorRespectJSON("bonding: %v", err)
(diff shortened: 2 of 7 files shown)


maphew:
Thank you, Rody. This contribution identified an independent reproduction and supplied the modern current-main routing design that the earlier #4350 branch needed.

Following contributor-first consolidation, I carried your original commit onto #4350 after Johan's earlier authored commit, then added maintainer fixes for same-store identity, read-only discovery before writable reopen, dry-run policy parity, actionable route errors, and reopen identity verification. Your commit, changelog work, #4714 reproduction, issue-first behavior, writable prefix-routing design, and tests remain attributed in the combined history and PR body.

#4350 is now the consolidation branch at `bcafe5bdd889eaaf97d231275a3fed85f10f615c`. I am leaving this PR open until the combined branch passes its remaining gates and lands; afterward this PR can be retired with a concrete pointer rather than closed prematurely.

maphew (dismissed):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the direction change in repo history). Verdict: **needs changes** — the routing goal (bond operands resolve like show/update/close) is right; two structural collisions need resolving first.

**Blocker 1 — the resolver conflicts with the landed #3874 direction.** Verified in history: upstream deliberately *removed* the `looksLikeFormulaName` gate from mol bond's resolution ("resolve plain formula names like every other subcommand", with regression coverage) — while this PR's rewritten resolver reintroduces gating that rejects valid plain formula names. merge-tree conflicts in both resolver paths; the rebase must preserve main's unconditional parser fallback rather than reapplying the gate.

**Blocker 2 — routed operands resolve but can't be bonded.** Contributor auto-routing deliberately opens read-only stores (embedded writes are refused at the store layer), yet `runMolBond` selects that store and dispatches mutations. The new test verifies only *resolution*, so it passes while the actual bond fails at write time. Either open the routed store writable for bond (if that's intended policy) or fail fast with a clear message before mutating.

**Should-fix — same routed database compares as two different stores.** `pickBondStore` compares Go handle identity, but each route-open constructs a fresh store in server mode, so two operands in the same routed rig are wrongly rejected as cross-store. The locator-based identity pattern already exists (`dependencyStoreKey` in `dep.go`) — reuse it.

Plus: end-to-end coverage for formula+routed / same-routed / different-store bond cases, and hook-isolate the test's `git init` per the repo convention.

R0SEWT:
Thanks for the thorough cross-vendor review — all three points are addressed. Rebased onto current main (`222a5df` rebase + `57a9f2a` review response; PR description updated to match).

**Blocker 1 — resolver conflicted with the landed #3874 direction.** The rebase preserves main's unconditional parser fallback: `looksLikeFormulaName` is gone from the branch entirely (zero occurrences). Both the mutation path (`resolveMolBondOperands`) and the dry-run path (`resolveOrDescribe`) fall through to the formula registry unconditionally and let `parser.LoadByName` decide. `TestBdMolBond_ResolvesPlainFormulaName` (#3874's regression test) passes against the new resolver.

**Blocker 2 — routed operands resolved but couldn't be bonded.** Resolution now mirrors `resolveCloseTargets` explicitly (local → prefix-routed **write-intent** → shared auto-routed **read-only** handle) and records writability per operand. `pickBondStore` rejects a bond pinned to a read-only store with a clear message *before* any mutation is attempted — since upstream policy is that auto-routed stores hydrate foreign projects and must never be mutated (GH#3231/#4141), fail-fast is the chosen behavior rather than opening them writable. On the test gap you called out: `TestMolBondPrefixRoutedWriteThrough` now exercises the actual dependency write and verifies it committed in the routed rig database via a fresh routed read; `TestMolBondAutoRoutedReadOnlyFailsFast` covers the fail-fast path.

**Should-fix — same routed database compared as two different stores.** Two changes: both operands of a bond now share one routed handle (an already-opened routed store is tried before opening another), and `sameBondStore` falls back to `StoreLocator` identity (`CLIDir`, which pins the concrete database) for separately-opened handles. `TestSameBondStoreSeparateHandles` covers the exact scenario — two distinct handles from consecutive route-opens in server mode compare as the same store, and handles onto different databases do not.

maphew:
Both structural blockers from the July 24 review are verified delivered — no third round on those. The gating is genuinely gone (the single remaining `looksLikeFormulaName` occurrence is a comment in the #3874 regression test, which your branch correctly leaves untouched), both paths fall through unconditionally to `parser.LoadByName`, the routing order mirrors `resolveCloseTargets` step for step, and `TestMolBondPrefixRoutedWriteThrough` closes the exact gap the review named — a real bond against the routed store re-verified through a fresh routed read, not resolution-only. Two rounds of precise, complete responses; thank you.

**And an apology plus a decision you're owed**: on July 12 this PR was told it would be retired in favor of #4350, and then the July 24 review engaged your branch on its merits anyway — two maintainer signals pointing in opposite directions while you kept working. Resolving that now, in your favor: **#4720 is the surviving branch.** The mergeability picture inverted since July 12 — #4350 is now 425 commits behind main with real content conflicts, while your branch is 19 behind and conflict-free. We'll port the two genuinely better pieces of #4350's design onto this branch (with Johan's attribution preserved): the read-only-discovery-then-writable-reopen phasing, so a cross-store bond that gets rejected never writable-opens a foreign rig first, and matched-route error propagation, so a route that fails on config/lock doesn't collapse into "not found".

What's actually left, in order:

1. **Rebase onto current main** — one-character fix: `openRoutedReadStore` now returns 4 values (#4866), so `resolveMolBondOperands`'s call site needs `rs, routed, _, err`. Your branch doesn't compile against main until then; this also explains the silence from CI (which I've now approved to run — your head had zero executed checks, so nothing about it was validated server-side yet).
2. Should-fix: dry-run parity — `resolveOrDescribe` applies routing but skips the `pickBondStore` writability/cross-store validation, so `--dry-run` can print "Result: compound molecule" for a bond the real path refuses. Given #4714's framing was exactly "dry-run doesn't catch it", the dry-run should surface the same rejection.
3. Should-fix: `sameBondStore` should unwrap decorators (`storage.UnwrapStore(...)` + `filepath.Clean`, matching `dependencyStoreKey` in `cmd/bd/dep.go`) — the global store is decorator-wrapped, so the direct type assertion silently mis-answers when a prefix route resolves back onto the local database.
4. The two #4350 ports above — we're happy to do those maintainer-side on your branch unless you want them.

Say the word on who takes 2–4; either way this lands under your and Johan's names together.

maphew:
Rebased your branch onto current main locally and ran the four new tests against `c989b6b87` — all pass, including `TestMolBondPrefixRoutedWriteThrough`, which is the one that actually proves the bond commits in the routed rig rather than the local store. That test is the difference between this PR and every earlier attempt at #4714, and it holds up under a real rebase. Thank you — two rounds, both complete, and the second one landed the hard part.

Since you haven't picked up items 2–4 from the last comment, we're taking them maintainer-side so this stops waiting on you. Attribution stays yours (and j-s-au's for the #4350 ports); the branch keeps your authorship. Here's what's being applied:

1. **Rebase.** `openRoutedReadStore` grew a fourth return value in #4866, so `resolveMolBondOperands` needs `rs, routed, _, err := ...`. Without it the branch doesn't compile against main — which is also why the green CI on your head isn't the whole story: it ran against your base (`ad7a5bc8e`), before that change. The only textual conflict is one CHANGELOG hunk.
2. **Dry-run parity.** `resolveOrDescribe` routes now, but the dry-run block never calls `pickBondStore`, so `--dry-run` can print `Result: compound molecule` for a bond the real path refuses (cross-store, or pinned to the read-only auto-routed store). Given #4714's framing was literally "dry-run doesn't catch it", dry-run should surface the same rejection.
3. **`sameBondStore` decorator unwrap.** The global store is wrapped in `HookFiringStore` (`cmd/bd/storage_chain.go:28`), which doesn't implement `StoreLocator` — so the direct type assertion silently answers "different stores" when a prefix route resolves back onto the local database. Switching to the existing `dependencyStoreKey` helper (`cmd/bd/dep.go:703`, `storage.UnwrapStore(...)` + `filepath.Clean`) deletes the bespoke comparison rather than adding to it. Your `TestSameBondStoreSeparateHandles` passes today only because it compares raw handles; a wrapped-store case is being added.
4. **The two #4350 ports**: read-only discovery before writable reopen (so a bond that gets rejected as cross-store never writable-opens a foreign rig first — the open-time-migration hazard from #4141/#3231), and matched-route error propagation (a route that matched but failed on config or lock currently collapses into "not found as issue or formula").

Two smaller notes, neither blocking: the handle-reuse step is tried before an operand's own prefix route, which is only reachable with partial IDs but binds to the wrong rig when it is; and `pickBondStore` checks writability of the first pinned operand only — safe as written, worth a comment saying why.

One thing to expect downstream: #5027 (proxied-server support for `bd mol`) touches the same functions and keeps `resolveOrCookToSubgraph`, which this PR removes. This lands first since it's the smaller, green, bug-fixing change; #5027 will rebase onto it, and its new proxied bond path will need this routing too or #4714 reproduces under proxied-server mode. Tracking that separately, not your problem.

Also: the multi-hyphen prefix limitation in #5048 (`extractBeadPrefix` cuts at the first `-`) applies to bond as well — but only because you correctly reused the shared resolver instead of writing a private one, so bond gets that fix for free when #5048 lands. That's the right architecture, and it's why this approach beat the alternatives.

Next: the four items go on your branch as maintainer commits (or a superseding branch with your commits preserved, if maintainer-push to your fork isn't permitted), CI re-runs, and the merge patrol takes it from green. If you'd rather push any of them yourself, say so in the next day or so and we'll hold.

R0SEWT:
Sounds good — please go ahead with the remaining items maintainer-side. The proposed changes and the ports from #4350 with Johan's attribution make sense to me.

Thank you for the consolidation decision, the thorough reviews, and for validating the branch against current main.

maphew:
Thanks, Rody — I’ve pushed the maintainer fix-merge to your branch at `729046786`.

What changed:

- rebased onto current `main`, preserving your two commits;
- made dry-run and execution share read-only routed discovery and store-policy validation;
- reused the unwrapping `dependencyStoreKey` identity check for separately opened/decorated handles;
- ported the applicable #4350 safety ideas: read-only discovery before writable reopen and propagation of matched-route failures;
- reopened only the validated prefix/maintainer/default route writable, then verified its identity before mutation;
- preserved contributor auto-routes as mutation-forbidden;
- restored the normal hook/telemetry decorator chain on routed writes;
- credited Johan Snyman on the maintainer integration commit.

The focused five-test cgo routing suite and three existing routing compatibility tests pass, the affected package compiles in the pure-Go configuration, and a fresh cross-vendor review found no remaining code issues. The full `make test` suite is queued against the exact pushed head.

I also updated the PR body to match the final design and called out that proxied-server route parity remains a separate follow-up. The stale automated changes-requested review has been dismissed because its findings are addressed. I have not handed this to merge automation: repository policy still requires a substantive human review for this code change.

maphew (commented):
Coming back to this as a requested reviewer, with two disclosures up front: I stepped down as maintainer on 2026-08-10, so I can no longer merge this or push to the branch; and the current head commit (729046786) is one I pushed myself in July, so a formal approve would be self-review. Taking the comment register instead.

The design still holds up on re-read: read-only discovery of both operands with the unconditional formula fallback preserved (#3874), same-store validation via the decorator-unwrapping `dependencyStoreKey`, fail-fast on mutation-forbidden contributor auto-routes, writable reopen with the TOCTOU identity re-check, and the hook/telemetry decorators rewired onto the routed store. The test file covers the right things, including a real write-through bond in a routed rig.

What has changed is `main`. The branch is now **CONFLICTING** (`cmd/bd/mol_bond.go`, `CHANGELOG.md` vs current main), and the rebase is **not purely mechanical**: #5253 (2af843a60) landed after the branch base, and a textual conflict resolution would compile and pass this PR's own tests while silently regressing it. Two specific landmines:

1. **Dry-run var validation.** Main's `resolveOrDescribe` now calls `formula.ValidateProvidedVars` so `--dry-run` fails the same way execution would. This PR's direct-store dry-run path (`discoverMolBondOperand` -> `renderMolBondDryRun`) never validates `--var` values; its formula branch only calls `parser.LoadByName`. Post-rebase, `bd mol bond --dry-run <formula> <x> --var bad=...` would preview success while the real bond fails - exactly the false preview #5253 fixed. The discovery/dry-run path needs the same `ValidateProvidedVars` call.
2. **`ErrVarValidation` double-wrap.** Main's `resolveOrCookToSubgraph` now passes `formula.ErrVarValidation` through unwrapped, but this PR's `materializeMolBondOperand` wraps every `resolveAndCookFormulaWithVars` error as `'%s' not found as issue or formula`. Post-rebase, a `--var` enum/pattern violation on a valid formula would again be reported as formula-not-found. It needs the same `errors.Is(err, formula.ErrVarValidation)` passthrough.

Two smaller notes for the same pass:

- `openAutoRoutedStore` now opens maintainer/default auto-routed stores writable for *any* write-intent caller of `resolveAndGetIssueWithRoutingAccess` (update/close as well, not just mol bond). That matches the accepted design, but no test exercises update/close through the newly writable auto-route and the CHANGELOG entry only mentions mol bond - worth one test or one sentence.
- `routed.go`: `os.IsNotExist(err)` works only while `loadPrefixRoutes` returns the raw `os.Open` error; `errors.Is(err, fs.ErrNotExist)` is the robust form.

No CI ran on the current head (statusCheckRollup is empty for 729046786), so the green claims in the PR body are local runs against an older main; CI needs the rebase anyway.

@R0SEWT the rebase plus the two #5253 reconciliations fall to you or a current maintainer now - sorry the maintainer-side help I offered earlier is no longer something I can deliver, and thanks for the precise turnarounds across the earlier rounds. @csells @julianknutsen: once rebased with the two items above addressed, this is close to landable from my side.

R0SEWT:
Rebased onto current main (`d871b5c54`) and both #5253 reconciliations are in, plus the two smaller notes. The branch keeps its three commits (my two + the maintainer integration commit with Johan's attribution) and adds two on top: `fix(mol): reconcile routed bond discovery with #5253 var validation` and `fix(mol): review polish`.

**Landmine 1 — dry-run `--var` validation.** Resolved one level lower than main's `resolveOrDescribe`: `discoverMolBondOperand` now takes the provided vars and runs `formula.ValidateProvidedVars` on its formula fallback, so discovery itself rejects an enum/pattern/provided-empty violation. Since dry-run and execution both go through discovery, they fail identically — and before any writable reopen. `resolveOrDescribe` keeps main's 4-arg signature and `ValidateProvidedVars` body verbatim (the proxied-server dry-run path calls it with vars, so resolving that conflict any other way doesn't compile). New coverage: `TestMolBondDryRunRejectsEnumViolatingVar` and `TestMolBondExecutionRejectsEnumViolatingVar` assert the violation surfaces and is *not* masked as `not found as issue or formula` — #5253 left no CLI-level bond dry-run test, so these guard the new path directly.

**Landmine 2 — `ErrVarValidation` double-wrap.** `materializeMolBondOperand` now passes `formula.ErrVarValidation` through unwrapped before the not-found wrap, matching `resolveOrCookToSubgraph` and pour/wisp. `TestMaterializeMolBondOperandPassesVarValidationThrough` pins it with `errors.Is`.

**Smaller note 1 — writable maintainer/default auto-routes beyond mol bond.** Took both halves of "one test or one sentence": the CHANGELOG entry now names the widening (maintainer/default auto-routes open writable for validated write-intent resolvers; contributor auto-routes stay read-only and mutation-forbidden), and `TestMaintainerAutoRouteMutationWritesThrough` drives `resolveAndGetIssueForMutation` — the `bd update` resolver — through a maintainer auto-route, commits a real update there, and re-verifies it via a fresh routed read.

**Smaller note 2 — `routed.go`:** `errors.Is(err, fs.ErrNotExist)` replaces `os.IsNotExist`.

Also picked up along the way: the routing test repos now pin `core.hooksPath` to `.git/hooks` after `git init` (the `initGitRepoAt` convention — the July review's nit), and the CHANGELOG entry was re-slotted into `## [Unreleased] → ### Fixed`; resolving the conflict in place would have filed it inside the released 1.2.1 section.

**Verification:** the full `make test` suite passed locally with zero failures against `f6a191fd6`, alongside the focused battery (the five routing tests, the four new var-validation/auto-route tests, `TestResolveCloseTargets`, `TestDepRoutedTargetOpensReadOnly`, `TestDepListCrossRigRouting`, `TestBdMolBond_ResolvesPlainFormulaName`, and #5253's pour/wisp suite). Main then moved again this morning, so the branch was re-rebased onto `d871b5c54` — only a CHANGELOG append conflicted; every code and test file is byte-identical to the fully-verified state — and `cmd/bd` was re-built in both the cgo and `CGO_ENABLED=0` pure-Go `gms_pure_go` configurations with the mol-bond battery re-run green.

@csells @julianknutsen — this needs CI approval on the new head (no checks have ever run server-side on this PR) and the substantive human review repository policy requires. Per matt's Aug 16 comment it's close to landable once the two #5253 items were addressed; they're in.

_claude-fable-5 on behalf of R0SEWT (AI-assisted; all changes verified by local test runs)_

PR #4721, opened 2026-07-11
test(embeddeddolt): stabilize operator conflict fixture

maphew:
## Why

The embedded-Dolt operator-conflict fixture relied on automatic timestamps. After the issues-table LWW resolver in gastownhall/beads#4699, differing timestamps can make the conflict safely auto-resolve, leaving the test intermittently expecting an error from a successful settle.

## What

Pin both divergent retitles to the same `updated_at` value. This keeps the conflict ambiguous by construction, so it deterministically exercises the operator-conflict reporting path.

## Validation

- `go test -tags gms_pure_go ./internal/storage/embeddeddolt -run '^TestEmbeddedMergeAndSettleReportsOperatorConflicts$' -count=10 -v`


maphew:
the fixture now only exercises the tie→ambiguous path. Worth confirming #4699 has a companion test asserting that differing timestamps do auto-settle — that path just lost its accidental coverage here. If not, I'll add one in a follow-up.
Also ran -count=10 locally; a -race pass would add confidence since the flake was timestamp-granularity-dependent.

Fable-5-thinking

PR #4723, opened 2026-07-11
fix(postgres): serialize concurrent schema initialization

duncan4123:
## Summary

- serialize PostgreSQL schema bootstrap with a database-and-schema-scoped advisory lock
- prevent concurrent `bd` processes from racing PostgreSQL catalog DDL
- add a regression test using eight independent PostgreSQL sessions

## Root cause

Every native PostgreSQL backend open runs `InitSchema`. Gas City can start several short-lived `bd` commands concurrently, and PostgreSQL `IF NOT EXISTS` / `CREATE OR REPLACE` DDL is not concurrency-safe across sessions. This produced catalog errors such as `tuple concurrently updated` and duplicate `pg_namespace` keys.

## Validation

- `go test ./internal/storage/postgres -count=1`
- live PostgreSQL `TestInitSchemaConcurrent` and `TestInitSchema`
- 12 concurrent `bd list --json` calls against a Gas City native PostgreSQL workspace
- `gc session list` no longer reports schema initialization errors

### DIFF
--- a/internal/storage/postgres/schema.go
+++ b/internal/storage/postgres/schema.go
@@ -476,6 +476,21 @@ func InitSchema(ctx context.Context, db *sql.DB, schema string) error {
 	}
 	defer conn.Close()
 
+	// InitSchema runs on every backend open. Separate bd processes can open the
+	// same workspace concurrently, and PostgreSQL's IF NOT EXISTS / CREATE OR
+	// REPLACE forms are not safe against concurrent catalog updates (for
+	// example pg_namespace uniqueness and function replacement can race). Hold
+	// a session advisory lock for this database+schema while applying bootstrap
+	// DDL. The lock is automatically released when conn closes even if the
+	// explicit unlock cannot run because ctx was canceled.
+	const lockKey = `hashtextextended(current_database() || ':' || $1, 0)`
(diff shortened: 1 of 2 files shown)


harry-miller-trimble:
Thanks for identifying and fixing this — the concurrent `InitSchema` race (`tuple concurrently updated`, XX000) is real, and your advisory-lock approach is the right shape.

We're going to consolidate on #4725, which fixes the same bug with a couple of additional robustness properties, so I'm closing this as superseded. To be transparent about *why* #4725 was chosen (not a knock on this PR — the diagnosis here is sound):

- **Lock key derivation.** #4725 computes the advisory-lock key in Go (FNV-64a of `bd.initschema:<schema>`) and pins it with a golden-value test, which makes the key a stable cross-binary/rolling-upgrade contract. This PR derives the key in-DB via `hashtextextended`, whose internal-function stability is a bit less guaranteed across versions, and there's no key-contract test.
- **Test coverage.** #4725 adds both the key-contract test and a heavier concurrent-open storm test.

Both PRs independently arrived at the same core fix (session-scoped `pg_advisory_lock` + cancel-safe unlock), so credit to you for the catch. If #4725 doesn't land for some reason, please reopen this — happy to revisit.

_This is an automated agent review, not a maintainer approval._

PR #4725, opened 2026-07-11
fix(postgres): serialize InitSchema DDL with a per-schema advisory lock

julianknutsen:
## What

Serializes `InitSchema` DDL application with a per-schema Postgres session advisory lock (`pg_advisory_lock`), keyed by an FNV-64a hash of the workspace schema name.

- `internal/storage/postgres/schema.go`: acquire the lock on the pinned connection before applying DDL; release via `pg_advisory_unlock` under `context.WithoutCancel` so a canceled ctx can never return a lock-holding connection to the pool and wedge every later open of the schema. Distinct schemas hash to distinct keys, so per-workspace inits serialize against themselves but never against each other.
- `internal/storage/postgres/schema_test.go`: `TestSchemaAdvisoryLockKey` pins the key derivation with golden values (the key is a cross-process contract — changing the hash silently breaks lock interop with already-deployed binaries), and `TestInitSchemaConcurrentOpens` is the regression test: 8 separate `*sql.DB` pools (mimicking separate bd processes) hammering `InitSchema` on one fresh schema, gated on `BEADS_PG_TEST_URL` like the existing Postgres tests.

## Why (live incident: concurrent bd opens race pg catalog DDL -> tuple concurrently updated -> orphan-drained sessions)

`InitSchema` runs on EVERY open and re-applies idempotent DDL (`CREATE OR REPLACE FUNCTION`, `CREATE INDEX IF NOT EXISTS`, ...). Idempotent is not concurrency-safe: concurrent opens rewriting the same pg catalog tuples collide with `ERROR: tuple concurrently updated (SQLSTATE XX000)`.

Diagnosed live on a running gc fleet: the dispatcher, session reconciler, and agent subprocesses all open bd concurrently, so the race is routine — and it surfaced as the reconciler's assigned-work probe failing on exactly this error and orphan-draining healthy worker sessions mid-workflow. The failure looks like a bd/store outage, but it is a DDL race on open.

## Verification

- `go test -tags gms_pure_go ./internal/storage/postgres/` — pass (`TestSchemaAdvisoryLockKey` runs everywhere; the concurrency test additionally ran against live Postgres via `BEADS_PG_TEST_URL`).
- `go vet` and `golangci-lint run` clean on the package.
- Red-team storm (`bd list` first-op opens, 8 concurrent processes per round, same workspace schema, real Postgres):
  - **Unpatched bd: 19/80 opens failed** with `tuple concurrently updated (SQLSTATE XX000)`.
  - **Patched bd: 0/120 failures** on the pre-existing schema and **0/40** on a fresh schema.
  - Mixed fleet (patched + unpatched binaries concurrently): 4/40 failures — expected, unpatched binaries bypass the lock, so full protection lands only once all binaries in a deployment are upgraded (rolling-upgrade caveat, no worse than today).

### DIFF
--- a/internal/storage/postgres/schema.go
+++ b/internal/storage/postgres/schema.go
@@ -5,9 +5,19 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"regexp"
 	"strings"
 )
+
+// schemaAdvisoryLockKey derives a stable Postgres advisory-lock key from a
+// workspace schema name. Distinct schemas map to distinct keys so per-workspace
+// InitSchema calls serialize against themselves but never against each other.
+func schemaAdvisoryLockKey(schema string) int64 {
+	h := fnv.New64a()
+	_, _ = h.Write([]byte("bd.initschema:" + schema))
+	return int64(h.Sum64())
+}
 
 // ddl is the Postgres backend schema, embedded verbatim from
 // internal/storage/postgres/schema.sql. It is applied statement-by-statement by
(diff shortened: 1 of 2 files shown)


harry-miller-trimble:
## Automated agent review — recommend MERGE (and close #4723 as superseded)

> This is an automated agent review, not a maintainer approval, and does **not** satisfy the human-review requirement in PR_MAINTAINER_GUIDELINES §77. It is posted to inform a maintainer's review. This touches storage init — a sensitive path — so it warrants a real human review.

### The race (real, present on main)
`InitSchema` (`internal/storage/postgres/schema.go`) runs on every backend open and re-applies idempotent-but-not-concurrency-safe DDL (`CREATE SCHEMA IF NOT EXISTS`, plus `CREATE OR REPLACE FUNCTION` / `CREATE INDEX IF NOT EXISTS`) with no cross-session serialization. Concurrent `bd` processes collide on the catalog with `tuple concurrently updated (XX000)`.

### The advisory lock is correct
- **Right type:** `pg_advisory_lock` (session-scoped), not `pg_advisory_xact_lock` — each DDL is autocommit (no enclosing txn), so an xact lock would release after the first statement. Session lock spans the whole DDL sequence.
- **Key:** FNV-64a of `"bd.initschema:"+schema`. Postgres advisory locks are namespaced per-database internally, so keying on schema name alone is sufficient; distinct schemas don't block each other.
- **Acquired before `CREATE SCHEMA`** so it covers `pg_namespace` creation.
- **Released on all paths:** `defer pg_advisory_unlock` registered immediately after acquisition; defer LIFO ordering runs unlock before `conn.Close()` returns the connection to the pool — no leak on DDL/stamp errors.
- **Canceled-ctx handling:** unlock uses `context.WithoutCancel(ctx)` — without it, `database/sql` would short-circuit on an already-done ctx and return a healthy connection to the pool still holding the lock, wedging future opens. Sound.
- No deadlock risk (single lock, never nested, dead sessions auto-release).

### Tests — good
`TestSchemaAdvisoryLockKey` pins the key with golden values (a real cross-binary/rolling-upgrade contract; runs without PG). `TestInitSchemaConcurrentOpens` (gated on `BEADS_PG_TEST_URL`) storms 8 pools × 3 opens on one fresh schema. Consistent with the existing Dolt precedent (per-database advisory lock around schema init).

### Relationship to #4723
#4723 (duncan4123) fixes the same bug but is weaker: key computed in-DB via `hashtextextended` (less robust as a cross-binary contract than the Go-side FNV + golden pin test), and no key-contract test. Recommend merging this and closing #4723 as superseded, with credit.

PR #4727, opened 2026-07-11
feat(unclaim): add --if-assignee conditional release (CAS inverse of claim)

julianknutsen:
## What

Adds `bd unclaim --if-assignee <who>`: an atomic compare-and-swap release, the inverse of the claim-side CAS. The issue is released only while it is still assigned to the expected assignee; otherwise bd exits nonzero with a typed `storage.ErrAssigneeMismatch` naming the current holder and leaves the claim untouched.

## Why

The concurrency audit flagged the conditional-release gap: bd claims atomically (`ClaimIssueInTx`: `UPDATE ... WHERE assignee = ''` + RowsAffected) but can only release unconditionally. A supervisor returning a dead worker's issue (e.g. Gas City's orphan release) is forced into read-then-unclaim — a TOCTOU window in which the claim can be reclaimed and re-taken by another worker, whose live claim the stale releaser then silently clobbers. The only atomic workaround was raw `bd sql`, which is not portable across backends and not part of the CLI contract. Severity: a correctness hole in multi-agent claim ownership — silent lost work under contention, no error on either side.

The new verb reuses the proven claim-side idiom (`UPDATE ... WHERE id = ? AND assignee = ?` in the same mutation tx, RowsAffected as the verdict), wired through issueops/sqlkit (sqlite, dolt server) and embedded dolt, and declared unsupported on the direct mysql/postgres stores like the other portable claim ops.

## Verification

TDD tests (written first, failed before the fix):

```
go test -tags gms_pure_go ./internal/storage/sqlite/ -run TestUnclaimIfAssigneeTwoHandles -count=1   # PASS
CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd/ -run TestUnclaimIfAssigneeCLI -count=1              # PASS (end-to-end CLI)
BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go ./internal/storage/embeddeddolt/ -run 'TestConformance/UnclaimIfAssignee' -count=1  # PASS (both new conformance subtests)
go test -tags gms_pure_go ./internal/storage/issueops/ ./internal/storage/mysql/ ./internal/storage/postgres/ ./internal/storage/ ./internal/telemetry/ ./internal/jira/ -count=1  # PASS
```

Red-team empirical re-run of the audit race (rebuilt patched bd; per issue, claim as workerA then race 4 releasers `--if-assignee workerB` (wrong) against 4 `--if-assignee workerA` (right)):

- 40 issues raced, 8 racers per issue, 320 total release attempts
- right-assignee wins: 40 (exactly 1 per issue) — wrong-assignee wins: 0
- loud mismatch losers: 280/280 (nonzero exit + "assignee mismatch ... held by" on stderr); quiet/mislabeled failures: 0
- exactly one `unclaimed` event per issue; `PRAGMA integrity_check` ok; 0 foreign_key_check rows

### DIFF
--- a/.github/workflows/pr-risk.yml
+++ b/.github/workflows/pr-risk.yml
@@ -203,9 +203,9 @@ jobs:
         # Conformance suite (conformance.RunAll via TestConformance) against the
         # embedded Dolt backend. Split from test-embedded-storage so each job
         # stays well under its timeout; the server twin runs in mybd-665b's job.
-        # 30m matches the repo's integration timeout: the audit suite is large and
-        # embedded Dolt under -race is slow, so 15m no longer left headroom.
-        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=30m -test.run '^TestConformance$'
+        # 45m gives the audit-heavy embedded run enough headroom without making
+        # the workflow open-ended.
+        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=45m -test.run '^TestConformance$'
 
   test-server-storage:
     name: Test (Server Dolt Conformance)
--- a/cmd/bd/unclaim.go
+++ b/cmd/bd/unclaim.go
@@ -31,6 +31,8 @@ assignee. If the holder differs — e.g. the claim was already reclaimed and
 re-taken by another worker — nothing is changed and bd exits nonzero with an
 error naming the current holder. Use this from supervisors that must return a
 specific worker's issue without ever clobbering someone else's live claim.
+--if-assignee requires a non-empty assignee and cannot be combined with --force
+(they encode contradictory intent).
 
 Exit status: 0 when every issue was released; 1 when any release failed
 (including an --if-assignee mismatch).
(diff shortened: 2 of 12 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds `bd unclaim --if-assignee <name>`, a compare-and-swap conditional release that only unclaims an issue when its current assignee still matches the expected value — the inverse of claim's CAS. That gives supervisors and automation a race-free way to release a claim they believe they hold without clobbering a claim that has since moved to a different holder.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- An empty `--if-assignee ""` previously slipped past the conditional path and fell back to unconditional unclaim semantics. The CLI now selects conditional mode by flag *presence* and rejects an empty expected assignee before resolving or mutating any issue, and the storage layer rejects an empty expected assignee as well.
- `--force` and `--if-assignee` could be combined ambiguously. They are now mutually exclusive and rejected before `RunE`, with regression tests asserting the existing claim is left intact when the conflicting flags are supplied.

Review findings addressed:
- correctness (major, high confidence — codex + claude): `bd unclaim --if-assignee ""` could bypass conditional release and fall back to unconditional unclaim. The final patch dispatches conditional mode by flag presence, rejects an empty expected assignee up front, makes `--force`/`--if-assignee` mutually exclusive, and adds CLI tests covering the empty-assignee and combined-flag cases.

Remaining non-gating notes:
- No remaining gating issues were found. The supervisor-safe CAS path was validated end to end — it prechecks the expected holder, performs a transactional `UPDATE ... assignee = ?`, rewrites `row_lock`, and returns `ErrAssigneeMismatch` with no release side effects on a stale-holder mismatch — and both the conditional and unconditional paths share `finishUnclaimInTx`, preserving lease deletion and `unclaimed` event recording.

### Maintainer Changes
The change set was reconstructed onto current `main`: the original PR branch had merged a stale `main` that predated the storage refactor removing the sqlkit/SQLite/MySQL/Postgres backends, so the feature was re-expressed against the Dolt-only `Storage` interface — adding `storage.ErrAssigneeMismatch`, `issueops.UnclaimIssueIfAssigneeInTx` (a CAS mirror of the existing unclaim-in-tx path), and the dolt / embedded-dolt / telemetry wiring. Maintainer fixups then addressed the review finding (`fix(unclaim): reject empty --if-assignee and conflicting --force`) and extended the embedded-conformance test timeout (`fix: extend embedded conformance timeout`). Final patch: 12 files changed, 386 insertions(+), 14 deletions(-).

### Final Review Status
Ready for the merge queue: final head `7b9721961` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/29639370848

### Review Iterations
2 review passes performed (1 fix iteration): request_changes → approve.

---

PR #4728, opened 2026-07-11
fix(sqlkit): run withReadTx read-only to keep WAL reader concurrency

julianknutsen:
## What

`withReadTx` now begins its transaction with `sql.TxOptions{ReadOnly: true}` instead of `nil` options, so reads no longer take SQLite's single write lock; a regression test pins reader/writer independence across two independent store handles on one database file.

## Why

SQLite concurrency audit caveat #5 (MED severity): the DSN sets `_txlock=immediate`, so a default `BeginTx` issues `BEGIN IMMEDIATE` and acquires the write lock — even in `withReadTx`, which only reads and always rolls back. Every read therefore queued behind any in-flight writer and surfaced `SQLITE_BUSY` after `busy_timeout`, forfeiting WAL's reader/writer independence. This was the amplifier that turned any long write into a full-store outage: in the audit hammer, a 26,238-issue `bd import` held the write lock for 118s, during which **100% of concurrent bd operations failed, reads included**. `TxOptions{ReadOnly: true}` suppresses the immediate begin mode (plain deferred `BEGIN` on SQLite; `START TRANSACTION READ ONLY` / `BEGIN READ ONLY` on the mysql/pgx backends — safe, since every `withReadTx` callback is a pure issueops read). Immediate mode bought nothing in `withReadTx` anyway: it always rolls back.

## Verification

TDD regression test (fails on the old code with `SQLITE_BUSY`, passes with the fix):

```
go test -tags gms_pure_go -count=1 -run TestReadConcurrentWithHeldWriteLock ./internal/storage/sqlite/   # PASS (0.07s)
go test -tags gms_pure_go -count=1 ./internal/storage/sqlite/                                            # ok (17.3s)
```

Red-team empirical re-run (real bd binaries, real OS processes, 6k-issue `bd import` holding the write lock ~28s while a second process loops `list`/`show`/`ready`):

- **Baseline binary (cc5d0563a):** reads during hold: 1 ok / 1 fail (each read stalled on the write lock, then errored)
- **Patched binary:** reads during hold: **190 ok / 0 fail** over the same ~28s window
- Write-path semantics preserved: 8-way cross-process claim race on one issue → exactly 1 winner (rc=0), 7 loud losers (rc=1, "issue already claimed")
- Post-run: `PRAGMA integrity_check` = ok, 0 foreign-key violations, row counts exact (6,005/6,005 imported issues present) in both runs

### DIFF
--- a/internal/storage/sqlite/locking_test.go
+++ b/internal/storage/sqlite/locking_test.go
@@ -61,6 +61,56 @@ func TestConcurrentAccessNoLockError(t *testing.T) {
 	}
 }
 
+// TestReadConcurrentWithHeldWriteLock guards WAL reader/writer independence
+// across separate store handles — the cross-process analog: two bd processes on
+// one database file. Handle A holds SQLite's single write lock in an open
+// transaction (the DSN's `_txlock=immediate` takes it at BEGIN); a read through
+// handle B must still succeed, because under WAL readers never wait on the
+// writer. Before withReadTx passed sql.TxOptions{ReadOnly: true}, B's read also
+// issued BEGIN IMMEDIATE, queued on the write lock, and surfaced SQLITE_BUSY
+// after busy_timeout — so during any long write (e.g. a large import) every
+// concurrent read failed, forfeiting WAL's reader concurrency entirely.
+func TestReadConcurrentWithHeldWriteLock(t *testing.T) {
+	ctx := context.Background()
+	path := filepath.Join(t.TempDir(), "wal-read.db")
+	stA, err := Provision(ctx, path)
+	if err != nil {
+		t.Fatalf("Provision: %v", err)
+	}
+	t.Cleanup(func() { _ = stA.Close() })
+	if err := stA.SetConfig(ctx, "issue_prefix", "test"); err != nil {
+		t.Fatalf("SetConfig(issue_prefix): %v", err)
+	}
+
+	// Second, independent handle (own pool/connection) on the same file.
+	stB, err := New(ctx, Config{DSN: dsn(path)})
+	if err != nil {
+		t.Fatalf("New (second handle): %v", err)
+	}
+	t.Cleanup(func() { _ = stB.Close() })
+
+	// Hold the write lock on A: with _txlock=immediate, BeginTx issues BEGIN
+	// IMMEDIATE. The uncommitted INSERT makes it a real in-flight write.
+	writeTx, err := stA.(*Store).DB().BeginTx(ctx, nil)
+	if err != nil {
+		t.Fatalf("begin write tx on handle A: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes SQLite read
transactions actually read-only so that concurrent readers keep working while a
writer holds the lock, which restores WAL's reader/writer independence for the
`sqlkit`-backed stores and removes a real source of spurious `SQLITE_BUSY`
stalls under load.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Read transactions were inheriting the connection's `_txlock=immediate`
  setting, so every `withReadTx` began with `BEGIN IMMEDIATE` and grabbed
  SQLite's single write lock. That made pure reads queue behind an in-flight
  writer and time out with `SQLITE_BUSY`, defeating WAL reader/writer
  concurrency. The final patch begins read work with
  `sql.TxOptions{ReadOnly: true}`, which suppresses the immediate-lock mode on
  SQLite (plain deferred `BEGIN`) and maps to `START TRANSACTION READ ONLY` /
  `BEGIN READ ONLY` on the MySQL and Postgres backends. Reviewers confirmed
  every current `withReadTx` caller is a pure read, so the read-only mode is
  safe across all `sqlkit` consumers.
- Test evidence: the new regression test `TestReadConcurrentWithHeldWriteLock`
  holds a write lock open on one handle and proves a second handle can still
  read committed config concurrently, locking in the behavior the fix restores.

Review findings addressed:
- correctness (approve, Claude + Codex): both reviewers independently validated
  the premise and the fix — default transactions under `_txlock=immediate`
  contend for the write lock, and `ReadOnly: true` is the correct, minimal way
  to keep reads concurrent. No blocker or major findings were raised.

Remaining non-gating notes:
- Forward-looking only: a future `sqlkit` read-labeled method that accidentally
  issues DML through `withReadTx` would not be caught as strictly on SQLite's
  deferred read transaction as on the enforcing server backends. Adding
  conformance coverage that read-labeled methods stay pure reads is a good
  follow-up; it is not a gate for this change because all current callers were
  verified to be read-only.

### Maintainer Changes
No maintainer changes -- PR accepted as-is. The contributor's single commit was
rebased onto the latest `main` before merge-ready (no content change; patch id
preserved) and re-validated by CI.

### Final Review Status
Ready for the merge queue: final head `4516f453b` has passing required GitHub
checks, and the merge-ready metadata is in place. I am marking this PR
`status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/29208690146 (final head `4516f453b`; all required checks green)

### Review Iterations
1 review pass performed with 0 fix iterations — the change was approved on the
first pass and accepted without maintainer code changes.

---

PR #4729, opened 2026-07-11
fix(import): write large imports in bounded transactions

julianknutsen:
## What

`bd import` now writes large imports (>250 issues) in bounded transactions with a short pause between commits, ordering rows so readiness-affecting dependency edges commit in the same transaction as their row. Small imports keep the exact pre-existing single-transaction path.

## Why

The audit found that `bd import` wrote the entire batch in one transaction, holding the store's write lock for the whole import — a full-store outage for every concurrent `bd` command. Severity: high (fleet-visible outage on any large import; a 26k-issue import measured ~2 minutes of total write-lock hold on SQLite, far past the 5s `busy_timeout`, so concurrent operations fail rather than wait).

Empirical numbers from the red-team harness (10k-issue import with four concurrent `bd` probes — two readers, a `show` sampler, a writer):

- **Before:** 226 concurrent ops, 10 failed (4.4%), longest availability gap **46.8s** (the entire import), max successful-op latency 21.9s.
- **After:** 234 concurrent ops, **0 failed**, longest gap **1.66s**, max latency 6.1s. Import wall time 47.2s → 50.8s (~8% overhead).

Two measured pitfalls shaped the fix: chunking alone is not enough (SQLite busy-polling has no fairness queue; without the 150ms inter-chunk pause the import loop re-took the lock before any waiter polled — unchanged 100% failure, zero acquisitions across 20 chunk commits), and chunk size must keep per-transaction lock holds under `busy_timeout` (250 issues ≈ 1.4s hold vs ~6s at 1000). Correctness across chunk boundaries: rows land in dependency order (Kahn over readiness edges), so `bd ready` mid-import can never offer blocked work; cross-chunk non-readiness edges are wired by a `ConflictSkip` dependency pass that never rewrites rows, so a racing update can neither be clobbered nor stale-drop the edges; a mid-import failure leaves a durable committed prefix and re-running the same import converges.

## Verification

TDD tests (12 new tests in `cmd/bd/import_chunking_test.go`, covering chunk boundaries, same-transaction blocking edges, mid-failure durable prefix, inter-chunk pause, real-store resume convergence, no-ready-window at every freeze point, concurrent-update races, stale-rejection edge policy, and concurrent availability):

```
go test -tags gms_pure_go -run 'TestImportChunked|TestImportIssuesCore' ./cmd/bd/
ok  github.com/steveyegge/beads/cmd/bd  16.369s   (14 tests, all PASS)
```

Red-team empirical re-run (harness: 10k-issue JSONL import against a live SQLite store with concurrent reader/writer/sampler probes, per-op exit codes analyzed for failure rate and availability gaps):

```
BASE:  total ops=226 ok=216 fail=10 fail_rate=4.4%   longest gap=46.79s  max latency=21.92s
FIX:   total ops=234 ok=234 fail=0  fail_rate=0.0%   longest gap=1.66s   max latency=6.08s
```

Kill-and-resume re-run: import killed mid-flight, re-run completed `10000/10000 issues committed` + `dependencies wired for 186/186 issues` and converged on the full set.


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR rewrites large `bd import` writes so that rows and dependencies are committed in bounded, paced transactions instead of one giant all-or-nothing transaction — which matters because a big import previously held a full-store write lock for the entire operation, stalling every other writer until it finished.

### Original PR Review
Decision: approve.

What the review validated:
- **Full-store write-lock outage on large imports** — confirmed fixed: large imports now write in bounded row/dependency transactions with inter-chunk pacing, while small imports keep the existing single-transaction fast path, so the lock is no longer held for the whole import.
- **Readiness-safe chunking for normal imports** — confirmed correct: readiness-affecting dependencies are ordered so non-cycle rows carry their blockers inline before each chunk's `is_blocked` recompute, so chunking doesn't briefly mark blocked work as ready.
- **Deferred dependency pass and retry convergence** — confirmed correct: `ConflictSkip` preserves row writes while wiring deferred dependencies, and a mid-import failure can be resumed idempotently without corrupting state.

Review findings addressed:
- Readiness edge case (minor, medium confidence — claude): a duplicate-ID import can transiently expose a row as ready before a later duplicate row wires its blocker. The review confirmed the final state still converges correctly, so this is a brief transient window rather than a lasting inconsistency.

Remaining non-gating notes:
- The duplicate-ID transient-readiness window above is the only residual item and is non-gating; no blocking gaps were found. Worth a follow-up if strict readiness during import of duplicate IDs ever becomes load-bearing.

### Maintainer Changes
No maintainer code changes -- PR accepted as-is. The branch was rebased onto the latest `main` to keep it current and re-run CI against the current base; the change set is byte-for-byte identical (patch-id preserved).

### Final Review Status
Ready for the merge queue: final head `0077eeafe` was rebased onto the latest `main` and has passing required GitHub checks (Conformance, Cross-Version Smoke Tests, PR, PR Risk, and Regression Tests all green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/29684543675

### Review Iterations
1 review pass performed; the PR was approved with no fix iterations required.

---

PR #4730, opened 2026-07-11
fix(storage): normalize persisted audit timestamps to UTC

jakelindsay87:
## What

Normalize imported issue, comment, and dependency timestamps to UTC before persistence. Fresh dependency inserts now bind the same explicit UTC `created_at` value instead of relying on database-side `NOW()`.

## Why

Serialized timestamps can carry a non-UTC offset even when they represent the correct instant. Persisting those values unchanged makes audit fields backend/session dependent and can break byte-stable export and cross-clone comparisons.

This keeps imported dependency attribution from #4432 intact: a serialized `created_by` is preserved, while dependencies without one still fall back to the current actor. It also keeps the deterministic dependency identity introduced in #4266 unchanged.

This is separate from #4298 / #4346, which concern implicit `updated_at` changes during blocked-state recomputation.

## Verification

- `go test ./internal/storage/issueops` passes.
- Direct `AddDependencyInTx` sqlmock coverage asserts that an offset timestamp is bound as the exact equivalent UTC instant in the fifth INSERT argument, and that a zero timestamp falls back to a current `time.Time` with `Location() == time.UTC`.
- `go test ./internal/storage/...` passed the changed package plus `internal/storage/embeddeddolt`, SQLite, schema, and most storage packages. The aggregate Windows run remains red on existing platform/external-tool assumptions: Unix absolute-path fixtures, `printf`, missing external `dolt`, and generated-file newline drift.
- `git diff --check` passes.

The required PR wrappers were also attempted on Windows with the repository-pinned Go 1.26.2 toolchain:

- `make ci-pr-core`, `make ci-pr-policy`, and `make ci-pr-lint` cannot start directly because GNU Make is not installed.
- Direct `./scripts/ci/pr-core.sh` reaches `go test` but cannot build `runtime/cgo` because `gcc` is unavailable.
- Direct `./scripts/ci/pr-policy.sh` passes build-tag policy and Go-install guidance, then cannot read JSON version files because `jq` is unavailable.
- Direct `./scripts/ci/pr-lint.sh` stops when its formatting step invokes missing GNU Make; `golangci-lint` is also unavailable. A direct Windows `gofmt -l .` reports the repository-wide CRLF checkout, while the changed files were formatted with `gofmt` and the focused package test passes.

Linux CI should establish the complete required wrapper results.

### DIFF
--- a/internal/storage/issueops/create.go
+++ b/internal/storage/issueops/create.go
@@ -425,6 +425,13 @@ func PrepareIssueForInsert(issue *types.Issue, customStatuses, customTypes []str
 	} else {
 		issue.UpdatedAt = issue.UpdatedAt.UTC()
 	}
+	normalizeTimePtrToUTC(issue.StartedAt)
+	normalizeTimePtrToUTC(issue.ClosedAt)
+	normalizeTimePtrToUTC(issue.LeaseExpiresAt)
+	normalizeTimePtrToUTC(issue.HeartbeatAt)
+	normalizeTimePtrToUTC(issue.DueAt)
+	normalizeTimePtrToUTC(issue.DeferUntil)
+	normalizeTimePtrToUTC(issue.CompactedAt)
 
 	// Ensure closed issues have a closed_at timestamp.
 	if issue.Status == types.StatusClosed && issue.ClosedAt == nil {
@@ -443,6 +450,12 @@ func PrepareIssueForInsert(issue *types.Issue, customStatuses, customTypes []str
 		issue.ContentHash = issue.ComputeContentHash()
 	}
 	return nil
+}
+
+func normalizeTimePtrToUTC(value *time.Time) {
+	if value != nil && !value.IsZero() {
+		*value = value.UTC()
+	}
 }
 
 // ValidateIssueIDPrefix validates that the issue ID matches the configured prefix
@@ -665,6 +678,8 @@ func PersistComments(ctx context.Context, tx *sql.Tx, issue *types.Issue) (Creat
 		createdAt := comment.CreatedAt
 		if createdAt.IsZero() {
 			createdAt = time.Now().UTC()
+		} else {
+			createdAt = createdAt.UTC()
 		}
 		// Check for existing identical comment to prevent duplicates on re-import.
 		// The UUID PK means ON DUPLICATE KEY UPDATE would never fire,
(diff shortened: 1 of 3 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the signature drift at the PR head and current main). Verdict: **needs changes** — the UTC-normalization fix is correct and the dependency trace proves it necessary (the pinned driver formats `time.Time` without an offset); it needs a semantic rebase and one attribution fix.

**Blocker — the target functions were rewritten under this PR.** Verified: current main changed `AddDependencyInTx` to `(bool, error)` with an `AddDependencyOpts` parameter and conditional `dependency_added` event staging, and rewrote batch persistence into hierarchy-first phases. merge-tree conflicts in `dependencies.go` and `create_test.go`. The timestamp normalization must be ported into the new implementation without losing the event staging or the phased validation — not resolved mechanically.

**Should-fix — personal migration still rewrites the historical creator.** The PR preserves `dep.CreatedAt` but the insert continues writing `actor` as `created_by`, while the batch-import path correctly uses the `dependencyCreatedBy(dep, actor)` fallback. So `bd migrate personal` keeps the historical timestamp but stamps today's operator as creator. Use the same fallback in `AddDependencyInTx`, and add a migration round-trip test — the current sqlmock test omits `CreatedBy` and literally expects `"current.user"`, which is how this slipped through.

Nit: the timestamp cases want the table-driven shape.

With the rebase and the creator fallback, this closes a real imported-data-fidelity hole.

maphew:
Coordination note before anything else lands in this area — your PR is the older art here and we want the precedence handled in the open, not discovered after the fact.

Newly opened #5297 (cosentinode) changes the same dependency-INSERT line this PR changes, swapping `NOW()` for `UTC_TIMESTAMP()`. On that one statement your bound `dependencyCreatedAt(dep)` is the strictly better behavior — it fixes the same timezone defect *and* preserves an imported `dep.CreatedAt`, which `UTC_TIMESTAMP()` discards. Nobody is silently superseding that: the import-fidelity half remains yours.

The plan we're proposing, given this branch is currently in conflict with main (the target functions were rewritten into `addDependencyInTx` with `AddDependencyOpts` since you branched):

1. #5297's minimal `UTC_TIMESTAMP()` one-liner lands first — it's small, applies cleanly, and stops the user-visible bleeding from #5233.
2. This PR rebases on top and keeps its genuinely additive scope, which no other open PR covers: `normalizeTimePtrToUTC` over StartedAt/ClosedAt/LeaseExpiresAt/HeartbeatAt/DueAt/DeferUntil/CompactedAt in `PrepareIssueForInsert`, the comment-timestamp normalization, and `dep.CreatedAt` preservation on the dependency insert (which would then supersede the `UTC_TIMESTAMP()` interim on that line).

Per project policy, the rebase across the `addDependencyInTx` rewrite is maintainer work if you'd rather not chase it — the earlier review's semantic-rebase ask plus the `created_by` attribution question are the two open items either way. If we don't hear back we'll carry the rebase ourselves with your commits preserved; nothing here gets closed out from under you.

maphew:
#5635 now carries this PR's two commits rebased onto current main (authorship preserved), per the 2026-08-03 coordination note - main's addDependencyInTx rewrite and #5297 had left this unmergeable. @jakelindsay87 if you would rather pick it back up yourself, say the word and #5635 closes in favor of your branch.

PR #4726, opened 2026-07-11
fix(update): exit nonzero when a multi-ID update partially fails

julianknutsen:
## What

Makes `bd update <id...>` exit nonzero whenever any ID in the batch fails, with a per-ID failure report on stderr (a machine-parseable JSON line in `--json` mode). Successful IDs stay applied — the command remains per-ID, not atomic — and this contract is now documented in the command help.

## Why

Audit defect: multi-ID `bd update` only signaled failure when *every* ID failed (`len(args) > 0 && firstUpdatedID == ""`). One successful ID reset the whole batch to exit 0, so a bogus or unresolvable ID mid-batch was silently swallowed. Per-ID errors were printed inline on stderr, but the process reported success — scripts and orchestrators could not detect a partial batch. Severity: high for automation callers (silent partial writes with a success exit code).

Empirical baseline evidence: `bd update good1 bogus good2 --priority 1` exited rc=0 on baseline (both in text and `--json` mode, where stderr had no parseable failure report at all). Only the all-fail case exited nonzero. Patched: the same command exits rc=1, reports `1 of 3 issues failed to update` with the failed ID and reason, and in `--json` mode emits a final stderr line `{"error":..., "failed":[{"id":...,"error":...}], "schema_version":1}` (envelope-aware) while stdout keeps the array-of-updated-issues success shape.

## Verification

TDD tests (new, `cmd/bd/update_multi_id_exit_test.go`):

```
go test -tags gms_pure_go ./cmd/bd/ -run 'MultiIDUpdate' -v
--- PASS: TestMultiIDUpdatePartialFailureExitsNonzero (19.55s)
--- PASS: TestMultiIDUpdatePartialFailureJSONReportsFailedIDs (4.69s)
--- PASS: TestMultiIDUpdateAllFailStillExitsNonzero (3.24s)
--- PASS: TestMultiIDUpdateAllGoodPathUnchanged (4.25s)
```

Updated testscript (`cmd/bd/testdata/update_errors.txt` now expects nonzero exit):

```
go test -tags 'gms_pure_go scripttests' ./cmd/bd/ -run 'TestScripts/update_errors' -v
--- PASS: TestScripts (14.92s)  (subtest TestScripts/update_errors)
```

Red-team empirical re-run (baseline vs patched, scenarios S1-S7):

- S1 text partial: baseline rc=0 → patched rc=1, both good IDs still applied (prio=1)
[...]

PR #4732, opened 2026-07-11
fix(update): resolve metadata/note merges inside the mutation transaction (concurrent lost-update)

julianknutsen:
## What

Fixes the concurrent metadata/notes lost-update defect by moving all merge-shaped edits (`bd update --metadata/--set-metadata/--unset-metadata/--append-notes`, `bd note`) from client-side read-merge-write into the mutation transaction, on **both** the embedded path and the proxied-server path.

- **Operation keys instead of pre-merged values.** The CLI now passes the operation (`issueops.OpMergeMetadata`, `OpSetMetadata`, `OpUnsetMetadata`, `OpAppendNotes`); `issueops.ResolveMergeOps` resolves it against the row re-read inside the mutation transaction — in `updateIssueInTx` (embedded) and in the `domain/db` repository `Update` (proxied-server mode). The shared merge helpers (`MergeMetadataJSON`, `ApplyMetadataEdits`, `MetadataEditValue`) move to `internal/storage`; the old `cmd/bd` copies are thin aliases.
- **Locking read.** Merge-op updates read the row via new `GetIssueForUpdateInTx` (`SELECT … FOR UPDATE`): Postgres (READ COMMITTED) and MySQL (REPEATABLE READ) serialize the read itself; `sqlitedialect` strips the clause (safe — `_txlock=immediate` already takes the DB write lock at BEGIN); Dolt parses it as a no-op and relies on commit-time conflict detection plus the retry below. `pgdialect` corpus pins that `FOR UPDATE` survives translation.
- **Proxied server: whole-attempt retry, no silent swallow.** On a Dolt serialization failure the server redoes the WHOLE read-merge-write in a fresh unit of work (new exported `uow.IsSerializationError`; exponential backoff, 15s budget), commits each attempt exactly once, and when retries exhaust it fails **loudly** to stderr instead of exiting 0.

## Why

**The audit defect (HIGH).** The multiprocess-safety audit found that metadata edits and note appends merged against a snapshot read in an *earlier* transaction: two concurrent writers touching **different keys** of the **same issue** both exit 0, but the later commit erases the earlier writer's committed keys. Proven empirically on the stock binary: **7 of 200 exit-0 `--set-metadata` writes silently vanished** in an 8-writer same-issue hammer.

**The red-team block that forced this revision.** The first fix moved the merge in-transaction on the direct path only. Red-team review blocked it with two findings:

1. **Proxied-server mode still snapshot-merged** in `buildUpdateSpecForIssue` (against the unit-of-work's MVCC snapshot), so the daemon/Dolt sql-server route kept the lost update.
2. **An even nastier silent swallow:** after Dolt rolled a conflicted transaction back server-side, `uow.CommitWithRetries` retried only the COMMIT — re-committing the now-empty session. Dolt answered "nothing to commit", `isDoltNothingToCommit` swallowed it, and bd printed "✓ Updated" and exited 0 for a write that never landed.

A follow-up red-team round also blocked on Postgres/MySQL merge-op serialization (a plain in-tx read does not block on a concurrent writer's row lock, so the merge input was a stale MVCC snapshot — hence the `FOR UPDATE` read) and on notes (`bd note` / `--append-notes` had the same clobber — hence the append operation). This revision addresses all blocks.

## Verification

**Tests (all new, red on the unpatched paths):**
- `internal/storage/conformance/lostupdate.go` — cross-backend regression body: two independent store handles (separate pools, modeling two `bd` processes) plus a third lock-holding pool that keeps both writers in flight before either commits; wired into sqlite (`update_mergeops_test.go`), embedded dolt, Dolt sql-server (`uow/lostupdate_dolt_test.go`), mysql, and postgres suites.
- `internal/storage/domain/db/issue_mergeops_test.go` — repository-level merge-op resolution.
- `cmd/bd/update_metadata_race_test.go`, `cmd/bd/note_race_test.go` — real multi-process races through the CLI: concurrent `--set-metadata` writers and concurrent `bd note` appenders, no lost keys/notes.
- `cmd/bd/update_proxied_server_test.go` — whole-attempt retry on conflict, loud failure on exhaustion, and the legitimate "nothing to commit" (wisp-only update) still succeeding.
- `cmd/bd/update_proxied_integration_test.go` — proxied-server concurrent metadata + append-notes coverage.
- `internal/storage/sqlitedialect/translate_test.go` + `pgdialect/corpus_test.go` — `FOR UPDATE` stripped on SQLite, preserved with intact placeholder numbering on Postgres.

**Red-team hammer (8 writers × 25 iterations, same issue, one `bd` process per write):**

| Scenario | Stock binary | This fix |
|---|---|---|
| counter keys, widened lock window | **LOST-UPDATES** (k5, k6 reverted 25→23, all writes exit 0) | CLEAN (0/200 lost) |
| unique key per write, natural contention | **LOST-UPDATES** (3 exit-0 writes erased) | CLEAN (0/200 lost) |
| unique key per write, widened lock window | — | CLEAN (0 lost; the single contention exhaustion surfaced as a **nonzero exit**, not a silent loss) |

`go vet` clean, `gofmt` clean, golangci-lint 0 issues; full `./internal/storage/...` sweep green apart from known environment-bound `storage/dolt` sql-server slowness (those tests pass in isolation, verified at base and branch).


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a real concurrent
lost-update hole in `bd update`/`bd note` by resolving metadata and note merges
*inside* the mutation transaction (against the transaction-local row) instead of
against a stale pre-read, and it makes the proxied path retry the whole
read-merge-write unit of work on serialization conflicts. That matters because
two agents updating the same issue concurrently could previously clobber each
other's metadata keys or note appends and the CLI would still exit 0 — exactly
the silent data loss this change eliminates.

This PR was reviewed and adopted with maintainer fixes pushed directly to the PR
branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Rebasing the branch onto current `main` (after the SQLite/Postgres/MySQL
  backend removal) had surfaced a few regressions; the maintainer fixup restores
  the always-string `--set-metadata` contract (GH#4146) so `--set-metadata k=v`
  stays a JSON string rather than inferring numbers/booleans, and restores the
  proxied multi-ID partial-failure behavior so a batch with any failed ID exits
  non-zero (previously regressed #4726).
- The dead `internal/storage/sqlitedialect` package and the `FOR UPDATE` /
  row-locking scaffolding (a parse-only no-op on Dolt) were removed; correctness
  now rests on commit-time conflict detection plus the whole-attempt retry, which
  the tests exercise directly.

Review findings addressed:
- readability/complexity (major): the newly introduced `ResolveMergeOps` sat in
  the complexity major band, so it was split into `resolveMetadataMergeOps` and
  `resolveNotesAppendOp` (with an `isMergeOpKey` helper). Pure structural change —
  the merge-op key set, resolution order, direct-replacement conflict handling,
  schema validation, and copy-not-mutate semantics are all preserved.
- test hygiene (minor): a test-only retry-budget global was noted as potentially
  race-prone if the mutating test were ever made parallel — non-gating, left as a
  follow-up note.

Remaining non-gating notes:
- No remaining gating issues were found. The concurrent metadata/note lost-update
  and proxied whole-UOW retry behaviors were each re-validated against the final
  head.

### Maintainer Changes
Maintainer fixes were squashed into a single commit on top of the contributor
work — `fix(update): string --set-metadata, proxied partial-failure exit, drop
dead locking scaffolding` (14 files, +189/−1312, dominated by removing the dead
sqlitedialect package). It also aligns the test/policy layer with the new merge
semantics: allowlists the `TestUOW_ConcurrentMergeOps_NoLostUpdate` Dolt-container
stress test in `check-testing-short`, and removes two now-unreachable multi-ID
tests that seeded non-object metadata via `--metadata "[]"` (the new
in-transaction merge correctly rejects non-object metadata, so that scenario can
no longer occur). The four per-ID exit-code tests from #4726 are retained.

### Final Review Status
Ready for the merge queue: final head `2186227d7` has passing required GitHub
checks, and the merge-ready metadata is in place. I am marking this PR
`status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/29686127343

### Review Iterations
Four review passes were performed; the loop converged to `approve` after the
complexity split, and the final head passes all required CI (one proxied-Dolt
integration shard was re-run after an infrastructure timeout and passed cleanly).

---

Issue #4549, opened 2026-07-02
tracker: dolt_history fast path gated on DB() presence, not a history capability (breaks non-Dolt SQL backends; already mis-fires on embedded)


julianknutsen:
### Summary

`tracker.Engine.externalRefChangedAfter` gates a Dolt-specific fast path — a query against the `dolt_history_issues` system table — on the presence of a raw `*sql.DB` handle, rather than on whether the backend actually provides history / Dolt system tables. This conflates "backend speaks SQL" with "backend has Dolt's `dolt_history_*` system tables."

### Where

`internal/tracker/engine.go`:

```go
// engine.go:696
type dbProvider interface {
	DB() *sql.DB
}

// engine.go:700
func (e *Engine) externalRefChangedAfter(ctx context.Context, local *types.Issue, currentRef string, lastSync time.Time) (bool, error) {
	if local == nil {
		return false, nil
	}
	provider, ok := e.Store.(dbProvider)          // engine.go:704 — gate keyed on DB() presence
	if !ok || provider.DB() == nil {
		return local.CreatedAt.After(lastSync) || local.UpdatedAt.After(lastSync), nil  // graceful fallback
	}

	var previousRef sql.NullString
	err := provider.DB().QueryRowContext(ctx, `
		SELECT external_ref
		FROM (
			SELECT id, external_ref, commit_date FROM dolt_history_issues  // engine.go:713 — Dolt system table
		) h
		WHERE h.id = ? AND h.commit_date <= ?
		ORDER BY h.commit_date DESC
		LIMIT 1
	`, local.ID, lastSync.UTC()).Scan(&previousRef)
	...
}
```

The gate is `e.Store.(dbProvider)` + `provider.DB() != nil`. If it passes, the code queries `dolt_history_issues`. If it fails, it falls back to a coarser `CreatedAt`/`UpdatedAt` timestamp comparison.

### Why this is a robustness problem

The correct predicate for taking this path is **"the backend has queryable Dolt history"**, but the code uses **"the backend exposes a `*sql.DB`"**. Those are not the same property, and they are already decoupled in the codebase today:

[...]

PR #4733, opened 2026-07-11
feat(storage): add ExternalRefHistoryQuerier capability for external-ref history (#4549)

vishnujayvel:
## What

Adds `ExternalRefHistoryQuerier` — a storage capability interface for backends
that can resolve an issue's `external_ref` as of a point in time via Dolt's
`dolt_history_issues` system table. Implemented on
`DoltStore` and `EmbeddedDoltStore`; the consumer PR (#4734) discovers the
capability by unwrapping decorator-wrapped stores with the existing
(non-generic) `storage.UnwrapStore` helper.

This is the **primitive half** of the #4549 fix. The tracker wiring that consumes
it is in a stacked follow-up PR, per `CONTRIBUTING_PR_GUIDELINES.md`'s
primitive-before-wiring rule.

## Why

The sync fast-path needs to ask "did this issue's `external_ref` change since we
last synced?" — a question only history-capable Dolt backends can answer.
Exposing that as a *named capability*, rather than having callers type-assert a
raw `*sql.DB`, ties the fast-path to the behaviour it actually depends on instead
of to an implementation detail of how a store is built. The bug this enables
fixing is described in the follow-up PR.

## What changed

- `internal/storage/history_viewer.go` — `ExternalRefHistoryQuerier` (embeds
  `HistoryViewer`, adds `PreviousExternalRef`).
- `internal/storage/issueops/history.go` — `PreviousExternalRefInTx`, the
  `dolt_history_issues` query used by both stores' `PreviousExternalRef`
  implementations.
- `internal/storage/dolt/versioned.go`, `internal/storage/embeddeddolt/store.go`
  — implement `PreviousExternalRef` against `dolt_history_issues`, plus
  compile-time `var _ storage.ExternalRefHistoryQuerier` asserts alongside the
  existing interface checks.

## Verification

- `go build ./...` clean (cgo, `gms_pure_go` tag).
- Exercised and tested end-to-end in the stacked follow-up PR.

Root-cause analysis credited to the #4549 reporter.

---
_Consumer / wiring PR: **#4734** (stacked on this one)._

### DIFF
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -153,6 +153,7 @@ var _ storage.GarbageCollector = (*DoltStore)(nil)
 var _ storage.Flattener = (*DoltStore)(nil)
 var _ storage.Compactor = (*DoltStore)(nil)
 var _ storage.SchemaMigrator = (*DoltStore)(nil)
+var _ storage.ExternalRefHistoryQuerier = (*DoltStore)(nil)
 
 // DoltStore implements the Storage interface using Dolt
 type DoltStore struct {
--- a/internal/storage/dolt/versioned.go
+++ b/internal/storage/dolt/versioned.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"time"
 
 	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/storage/issueops"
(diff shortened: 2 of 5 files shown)


maphew (approved):

This is exactly the sanctioned direction for the storage boundary: a capability interface (`ExternalRefHistoryQuerier` embedding `HistoryViewer`) instead of type-asserting a raw `*sql.DB` across the boundary. Verified locally on today's main tip (672d94208): builds standalone, `go vet` clean, implemented only by `DoltStore` and `EmbeddedDoltStore` with compile-time asserts, and sqlite/mysql/sqlkit cleanly lack `PreviousExternalRef` so they fall through to the heuristic. No import cycles.

One doc nit, no code change needed: the PR description says this adds a generic `storage.UnwrapStore[T]` helper — the code actually (and preferably) reuses the existing non-generic `storage.UnwrapStore`. Feel free to update the body.

CI runs were pending workflow approval; I've approved them. Merge order: this one first, then gastownhall/beads PR 4734. Thanks vishnujayvel!

vishnujayvel:
Thanks for the review, maphew!

You're right about the `storage.UnwrapStore[T]` line — the PR body was
inaccurate. This PR doesn't add a new helper at all; #4734 (the consumer)
discovers the `ExternalRefHistoryQuerier` capability through the existing,
non-generic `storage.UnwrapStore` (`internal/storage/hook_decorator.go`).
I've corrected the description to say that.

PR #4734, opened 2026-07-11
fix(tracker): gate externalRefChangedAfter on capability, not DB() presence (#4549)

vishnujayvel:
> **Stacked on #4733** (the `ExternalRefHistoryQuerier` capability PR) — please
> review/merge that base first. Because #4733 isn't in `main` yet, this PR's diff
> currently also shows its storage commit; the **net-new change here** is
> `internal/tracker/engine.go` plus the two tracker test files. Once #4733 lands,
> this diff reduces to just the tracker change.

## What

Fixes the `externalRefChangedAfter` sync fast-path to gate on the
`ExternalRefHistoryQuerier` capability (added in the base PR) instead of
type-asserting a raw `*sql.DB` via `DB()`.

## Why — the bug (#4549)

The fast-path decided whether it could use Dolt history by checking for a raw
`*sql.DB`. That is the wrong signal, and it is wrong in two opposite directions:

- **`EmbeddedDoltStore`** (the default local backend) has the history data but
  exposes no raw `*sql.DB` — it runs SQL through its own transaction-scoped
  connection helper. So the check failed and the fast-path silently fell back to
  a coarse timestamp heuristic, producing wrong "changed / not-changed" verdicts.
- A backend that *does* expose `DB()` but is not Dolt would pass the old check
  and then query `dolt_history_issues`, a Dolt-only system table — an error, not
  a graceful fallback.

Gating on the capability fixes both: history-capable backends take the fast path;
everything else takes the intended fallback path.

## What changed

- `internal/tracker/engine.go` — `externalRefChangedAfter` resolves the store via
  `storage.UnwrapStore[ExternalRefHistoryQuerier]` and falls back cleanly when the
  capability is absent.
- `internal/tracker/engine_external_ref_history_test.go` — 11 pure-Go tests
  (fakes) covering capability present/absent, decorator unwrap, fast-path
  same/different-ref, not-found, error propagation, embedded-Dolt fast path, and
  the no-capability fallback.
- `internal/tracker/engine_history_gate_cgo_test.go` — 2 cgo tests against the
  real embedded-Dolt and Dolt-server stores.

## Verification

- `go build ./...` clean (cgo, `gms_pure_go` tag).
- **11 pure-Go gate tests: PASS.**
- cgo tests pass locally against embedded-Dolt; the Dolt-**server** path is
  exercised by CI (Docker was unavailable in my local environment — stated
  honestly rather than claimed).
- **revert-confirms-fail:** forcing the gate to `false` makes 7 gate tests fail;
  restoring the fix returns them to green — the tests genuinely catch the bug.

Root-cause analysis credited to the #4549 reporter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,18 @@ remote-migrate gate from a blunt block into a state-aware one.
 
 ### Fixed
 
+- **`--label-any` is no longer silently dropped by `bd ready` and
+  `bd ready --claim`.** The ready-work WHERE builder emitted clauses for
+  `--label` and `--exclude-label` but none for `--label-any`, so the OR-set
+  filter was ignored on the ready/claim path (with or without `--parent`) on
+  every backend — while `bd list`/`bd search` honored it. On an *atomic claim*
+  this was dangerous rather than merely wrong: a worker fencing itself to its
+  own lane (`bd ready --claim --label-any lane-a --parent epic-1`) would
+  happily claim another lane's issue and believe it was fenced. `--label-any`
+  now emits an OR-set membership clause that AND-combines with `--label`,
+  `--exclude-label`, and `--parent`, exactly as the flag help promises; an
+  exhausted lane now claims nothing instead of falling back to unfenced work.
+
 - **A failed v53 migration no longer traps the database, and the v53 repair
   now covers `wisp_dependencies` split-column drift.** rc.2 repaired the
   `issues` rig columns, but a database coming from schema v49 could still fail
--- a/cmd/bd/label.go
+++ b/cmd/bd/label.go
@@ -4,7 +4,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"os"
 	"sort"
 	"strings"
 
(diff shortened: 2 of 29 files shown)


maphew (approved):

Correct fix for gastownhall/beads issue 4549, and validated more deeply than CI will: locally on today's main tip I ran the pure-Go tracker gate tests, the embedded-Dolt fast-path test, **and** the Docker-backed Dolt-server test (`TestDoltStoreExternalRefChangedAfterUsesHistoryFastPath`, 2.47s) — all green. The capability-absent fallback preserves the exact prior heuristic (`CreatedAt/UpdatedAt` vs `lastSync`), and `found=false → changed=true` preserves the old `sql.ErrNoRows` behavior. Decorator unwrap via the existing `storage.UnwrapStore` is covered by tests.

One non-blocking nit: the extracted path maps a NULL historical `external_ref` to `("", found=true)` where the old inline code said "changed" unconditionally — divergent only when `currentRef` is empty, which the sole caller already guards against. A one-line test locking the NULL-historical contract would be nice, not required.

CI was pending workflow approval; approved. Will merge after gastownhall/beads PR 4733. Thanks vishnujayvel!

vishnujayvel:
Thanks for the review, maphew!

Added the test locking the NULL-historical-ref contract:
`TestEngineExternalRefChangedAfter_FastPathNullHistoricalRef` in
`internal/tracker/engine_external_ref_history_test.go`. It confirms a NULL
historical `external_ref` surfaces as `("", found=true)` and gets compared
against `currentRef` like any other value, rather than being treated as an
unconditional "changed" the way the old inline `!previousRef.Valid` check
did. Pure-Go suite passes (`CGO_ENABLED=0 go test -tags gms_pure_go
./internal/tracker/...`), `go vet` clean.

Heads-up: this pushes a new commit, so CI will need your workflow approval
once more — sorry for the extra click. Happy to drop the commit instead if
you'd rather merge as-is.

PR #4736, opened 2026-07-12
Expose configured backend opener

duncan4123:
## Summary

Expose a public `beads.OpenConfigured` API that opens the built-in backend selected by `.beads/metadata.json` and returns backend capabilities.

This supports the upstream-native Dolt, PostgreSQL, MySQL, and SQLite backends. Unknown backend names fail closed. It intentionally does **not** introduce external backend-plugin discovery or executable trust/configuration.

## Consumer

Gas City uses this API to open its native store against the configured Beads backend; see gastownhall/gascity#4154.

## Validation

- `go test . ./internal/backend`
- `CGO_ENABLED=0 go test . ./internal/backend`
- `go build ./cmd/bd`

### DIFF
--- a/beads.go
+++ b/beads.go
@@ -10,6 +10,7 @@ package beads
 import (
 	"context"
 
+	"github.com/steveyegge/beads/internal/backend"
 	"github.com/steveyegge/beads/internal/beads"
 	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/storage/dolt"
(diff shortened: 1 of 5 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the drift at the PR head). Verdict: **needs changes — really a redesign call**, same situation as #4859: main has moved to Dolt-only with fail-closed tombstones.

**Blocker 1 — the backend set this PR exposes no longer exists.** Verified: `internal/backend/configured.go` imports `internal/storage/{mysql,postgres,sqlite}` — all deleted on main (#4857, #4881), with the names retained only as tombstoned errors. merge-tree conflicts in `beads.go`, and even a resolved merge leaves uncompilable imports. Main also already exposes its own configured-open seam (`OpenBestAvailable`), so the surviving idea here — a public capability-descriptor API — would need rebuilding on that, Dolt-only.

**Blocker 2 — proxied-server mode selects the wrong store.** The CGO opener checks only `IsDoltServerMode()` and falls through to `embeddeddolt.Open` when proxied mode is active (proxied is deliberately rejected by the real factory because it requires the UOW provider) — while the returned descriptor claims `Embedded=false, ConcurrentWriters=true`. That's a wrong-database/false-empty risk wearing proxy capabilities.

**Blocker 3 — `ReadOnly` is three different promises.** Embedded Dolt gets strict read-only; server Dolt gets schema-only (a `CreateIssue` still opens a writable `BeginTx` — default transaction characteristics are ReadWrite in the pinned driver/engine); the SQL-family constructors discarded the flag entirely and provision schema at open. A public API can't ship an option whose meaning varies silently by topology.

**Should-fix:** `RawSQL=true` for every Dolt topology contradicts the proposal's own topology-aware capability requirement (embedded exposes no raw DB surface; the SQL stores that did got `false`). Plus: coverage tests only SQLite selection + unknown-name; no public docs; the nine capability booleans exist in triplicate.

Suggested disposition: close or redraft against current main's Dolt-only seam — the capability-descriptor idea is worth keeping; everything else has been overtaken.

PR #4738, opened 2026-07-12
feat(create): add status.default config for default initial status

thewoolleyman:
## Summary

Adds a `status.default` config key that sets the initial status of **single-issue `bd create`** when no explicit `--status` flag is given, instead of the built-in `open`. Complements the per-call `--status` flag (#4536) with an opt-out tenant default.

Precedence (highest → lowest): explicit `--status` flag → `--defer`-derived `deferred` → `status.default` → built-in `open`.

The configured value must be a built-in status or a configured custom status; it is validated at create time (matching the `--status` flag's `IsValidWithCustom` check), not at `bd config set` time. An empty/unset value falls through to `open` unchanged. Both create paths — embedded/direct and proxied/sql-server — read and validate it identically, **including the `--dry-run` preview**.

## Scope (intentional — narrow by design)

This deliberately covers **single-issue `bd create` only**, which is the path the requesting consumer (livespec's ledger-status-conformance track) uses. The following intentionally continue to start issues in `open` and do **not** consult `status.default`:

- **Batch create:** `bd create --file` (markdown) and `bd create --graph` (JSON)
- **Cross-repo routing:** `--repo` and contributor/maintainer auto-routing (the default is resolved against the local store)
- **Sibling commands:** `bd q`, `bd create-form`, `bd batch create`, `bd todo add`

Extending to those paths is straightforward follow-up if the project wants it, but it isn't required for the use case this unblocks, and keeping the change minimal contains the blast radius. Happy to file a tracking issue for broader coverage.

## Notable behavior fix

The proxied create path now resolves and validates `status.default` **before** the `--dry-run` preview is rendered. Previously the proxied dry-run always previewed `open` and skipped validation, diverging from the real create and from the embedded path (which was already correct). Now the proxied dry-run reflects the status the real create would apply and rejects an invalid default.

## Tests

- **Builder precedence unit tests** for both builders (`buildCreateIssue`, `buildCreateIssueFromInput`): flag > defer > default > open.
- **Embedded end-to-end** (`TestEmbeddedCreate/default_status_*`): default applies, flag wins, unset → open, invalid default rejected.
- **Proxied end-to-end** (`TestProxiedServerCreate_DefaultStatus`, new): config applies, flag wins, invalid default rejected, and `--dry-run` preview reflects the default — covering the proxied config-read/validate wiring the unit tests don't reach.

CLI reference docs regenerated from `bd help`.

### DIFF
--- a/cmd/bd/config.go
+++ b/cmd/bd/config.go
@@ -72,6 +72,23 @@ Custom Status States:
   This enables issues to use statuses like 'awaiting_review' in addition to
   the built-in statuses (open, in_progress, blocked, deferred, closed).
 
+  Default Initial Status:
+    Set status.default to change the status new issues receive from a
+    single-issue 'bd create' when no explicit --status flag is given.
+    Precedence: an explicit --status flag wins, then a --defer-derived
+    'deferred', then status.default, then the built-in 'open'. The configured
+    value must be a built-in status or a configured custom status; it is
+    validated when 'bd create' runs.
+
+    Scope: this applies only to single-issue 'bd create'. Batch create
+    (--file / --graph), cross-repo routing (--repo and contributor/maintainer
+    auto-routing), and the sibling commands 'bd q', 'bd create-form',
+    'bd batch create', and 'bd todo add' intentionally still start issues in
+    'open' and do not consult status.default.
+
+    Example:
+      bd config set status.default backlog
+
 Suppressing Doctor Warnings:
   Suppress specific bd doctor warnings by check name slug:
     bd config set doctor.suppress.pending-migrations true
--- a/cmd/bd/create.go
+++ b/cmd/bd/create.go
@@ -165,17 +165,27 @@ var createCmd = &cobra.Command{
 		issueType, _ := cmd.Flags().GetString("type")
 		assignee, _ := cmd.Flags().GetString("assignee")
 		statusFlag, _ := cmd.Flags().GetString("status")
-		if statusFlag != "" {
+		// store is nil for `create --repo=<remote URL>` with no local
+		// .beads/; fall back to built-in statuses only.
+		var defaultStatus string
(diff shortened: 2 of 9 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the retry-architecture drift at the PR head and current main). Verdict: **needs changes** — the feature is reasonable; the implementation predates two structural changes it must adopt.

**Blocker 1 — stale proxied transaction architecture.** Verified: the PR's proxied create still opens one UOW and calls `uow.CommitWithRetries` (two call sites), while current main rewrote the path to `uow.RunTxResult`, replaying the whole read/write closure on serialization failure — main's own `uow/errors.go` documents why commit-only retry can silently lose writes. merge-tree conflicts here; a mechanical resolution either doesn't compile or reintroduces the unsafe retry. The default lookup, validation, construction, and create all belong inside the retry closure.

**Blocker 2 — cross-repo creates consume the *source* tenant's default.** `status.default` is read before routing resolves, and the captured value survives the store swap — so `--repo <other>` and auto-routed creates apply (or reject on) the source repository's setting, despite the PR explicitly excluding cross-repo from the feature. Resolve routing first; skip the default whenever the target isn't `.`.

**Should-fixes:** an invalid `status.default` is rejected before `--defer` gets to supply its documented higher-precedence `deferred` status; and embedded `GetConfig` errors are discarded as "unset" (issue silently created as `open`) while the proxied path correctly propagates them — only missing/empty should fall back.

Nit: the duplicated three-case tests for both builders want one table with two adapters.

The batch/sibling-command exclusions otherwise hold, and store-open behavior is unaffected. Rebase onto the RunTxResult shape and this is a nice quality-of-life feature.

PR #4740, opened 2026-07-12
fix(config): stale legacy backend:"sqlite" metadata must not shadow a live Dolt workspace (bd-oyvc2.7)

steveyegge:
## What

Restores the pre-#4601 guard for legacy workspaces whose `.beads/metadata.json` still says `backend:"sqlite"` from the pre-#3151 SQLite era:

- `GetBackend()` honors `sqlite` only together with `sqlite_path` — the positive marker `bd init --backend=sqlite` always writes (pinned by new init tests), so every new-era SQLite workspace keeps working while bare legacy `backend:"sqlite"` resolves to Dolt again.
- Belt-and-suspenders: `sqlite.NewFromConfig` now refuses (with an actionable error naming both remedies) to provision a fresh SQLite database when the file is missing but a Dolt database already exists on disk — covering embedded, server-mode, and custom `dolt_data_dir` layouts.
- A one-line stderr notice (existing `warnSharedServerEmbeddedMismatch` idiom) nudges legacy users to remove the stale field.
- The `GetBackend` doc comment now matches the code (it previously claimed postgres/mysql-only while the switch honored sqlite).

## Why

Regression from #4601: it rewrote the `TestGetBackend` **"stale sqlite value"** pin — which existed precisely to protect these workspaces — into an assertion that sqlite is honored. On upgrade, a legacy Dolt workspace routed to the SQLite factory, silently created an empty `.beads/beads.db`, false-emptied every issue, and divorced new writes from the real store. Found in the bd-oyvc2 post-merge review of #4601; fixes **bd-oyvc2.7** (P2).

## How verified

- Unit tests: restored stale-sqlite pin + honored-with-path cases in `configfile_test.go`; new `fromconfig_test.go` guard matrix (refusal in both Dolt layouts, no file left behind, existing-file opens, fresh-clone provisions, custom `dolt_data_dir`); `store_factory_test.go` regression test proving legacy metadata opens embedded Dolt with no `beads.db` created; init tests pinning that `sqlite_path` is always persisted.
- `go build -tags gms_pure_go ./...` (plus CGO_ENABLED=0 build), `go vet`, `go test ./internal/configfile/ ./internal/storage/sqlite/` and the cmd/bd backend-selection tests under `BEADS_TEST_EMBEDDED_DOLT=1` — all green.
- Hands-on smoke: new-era `bd init --backend=sqlite` + create + list works; a Dolt workspace hand-edited to `backend:"sqlite"` lists its Dolt issues (notice printed, no beads.db created); the ambiguous marker-without-file case exits 1 with both remedies; the suggested `bd init --backend=sqlite --reinit-local` remedy works.

### DIFF
--- a/internal/configfile/configfile.go
+++ b/internal/configfile/configfile.go
@@ -17,7 +17,7 @@ const ConfigFileName = "metadata.json"
 
 type Config struct {
 	Database string `json:"database"`
-	Backend  string `json:"backend,omitempty"` // Storage backend: "dolt" (default), "postgres", "mysql", or "sqlite". Read via GetBackend().
+	Backend  string `json:"backend,omitempty"` // Storage backend: "dolt" (default), "postgres", "mysql", or "sqlite" (honored only with sqlite_path set). Read via GetBackend().
 
 	// Deletions configuration
 	DeletionsRetentionDays int `json:"deletions_retention_days,omitempty"` // 0 means use default (3 days)
@@ -250,9 +250,17 @@ func (c *Config) GetCapabilities() BackendCapabilities {
 }
 
 // GetBackend returns the configured storage backend. Only the explicitly-allowlisted
-// non-default backends ("postgres", "mysql") are honored; "", "dolt", and any legacy
-// or unknown value resolve to dolt so the default path stays byte-identical and a
-// typo fails safe to Dolt.
+// non-default backends are honored: "postgres", "mysql", and "sqlite" — the latter
+// only when sqlite_path is also set, the positive marker that `bd init
+// --backend=sqlite` always writes. Everything else — "", "dolt", unknown values, and
+// a bare backend:"sqlite" without sqlite_path — resolves to dolt so the default path
+// stays byte-identical and a typo fails safe to Dolt.
+//
+// The bare-sqlite rule protects pre-#3151 workspaces whose metadata.json still says
+// backend:"sqlite" but which have operated as Dolt ever since SQLite was removed:
+// honoring the stale field would provision a fresh empty SQLite database next to the
+// live Dolt data, making every issue vanish (false-empty) and divorcing new writes
+// (bd-oyvc2.7, regression from #4601). See HasStaleSQLiteBackend.
 func (c *Config) GetBackend() string {
 	if c != nil {
 		switch c.Backend {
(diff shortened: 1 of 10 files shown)


maphew (commented):
I reproduced two regressions despite the targeted tests passing. Both need correction before merge: legacy SQLite-only upgrades can lose access to their existing issues, and fresh SQLite provisioning can be blocked by unrelated Dolt environment state.

internal/configfile/configfile.go
```diff
@@ -261,12 +269,25 @@ func (c *Config) GetBackend() string {
 		case BackendMySQL:
 			return BackendMySQL
 		case BackendSQLite:
-			return BackendSQLite
+			if c.SQLitePath != "" {
```
maphew, line 272: **[P1] Preserve legacy SQLite-only workspaces**

This makes a pre-removal SQLite workspace (`backend:"sqlite"`, an existing `.beads/beads.db`, no Dolt data, and no `sqlite_path`) resolve to Dolt. `bd list` then reports no database, making the existing issues inaccessible. Backend selection needs to distinguish a live legacy SQLite database from stale SQLite metadata next to live Dolt data.

internal/storage/sqlite/fromconfig.go
```diff
@@ -25,11 +33,63 @@ func NewFromConfig(ctx context.Context, beadsDir string) (storage.DoltStorage, e
...
 
+// existingDoltDataDir returns the path of a Dolt database directory already present
+// in the workspace, or "" when none exists. It checks the embedded layout
+// (.beads/embeddeddolt/<db>/.dolt) and the server-mode data dir (default .beads/dolt,
+// or the configured dolt_data_dir), where databases live either directly
+// (<dir>/.dolt) or one level down (<dir>/<db>/.dolt).
+func existingDoltDataDir(beadsDir string, cfg *configfile.Config) string {
+	roots := []string{filepath.Join(beadsDir, "embeddeddolt")}
+	doltDir := filepath.Join(beadsDir, "dolt")
+	if custom := cfg.GetDoltDataDir(); custom != "" {
```
maphew, line 60: **[P2] Ignore Dolt environment when opening SQLite**

`GetDoltDataDir()` consumes `BEADS_DOLT_DATA_DIR`. In a fresh SQLite clone without its local `beads.db`, an ambient variable pointing to any populated Dolt root makes this guard scan unrelated databases and refuse provisioning. Dolt environment overrides must be inert for an explicitly selected SQLite backend; restrict this detection to workspace-local or config-owned paths.

PR #4741, opened 2026-07-12
fix(sqlite): make LIKE case-sensitive to match the other backends (bd-oyvc2.10)

steveyegge:
## What

Adds `_pragma=case_sensitive_like(1)` to the SQLite DSN so LIKE on the sqlite backend matches Dolt/MySQL (`utf8mb4_0900_bin`) and Postgres (collation "C"). Adds three Dolt-reference conformance audit cases plus a DSN unit test.

## Why

The new SQLite backend (#4601) left LIKE at SQLite's default ASCII-case-insensitive behavior (**bd-oyvc2.10**, from the post-merge review). Raw-cased emitters silently diverged: `bd list --id-prefix` with mixed case returned extra rows, parent-descendant listing (`id LIKE parent || '.%'`) leaked a same-lettered-different-case sibling parent's children, and `GetNextChildID` could count another parent's children into the counter. Intentionally case-insensitive searches wrap both sides in `LOWER()` and are unaffected — a full classification of every LIKE emitter is in the review notes; none relied on SQLite's case-insensitivity. No GLOB usage exists in the codebase.

## How verified

- modernc.org/sqlite applies `_pragma` DSN params on every new connection; the new `TestDSNCaseSensitiveLike` asserts `'a' LIKE 'A'` = 0 across a 4-connection pool.
- New conformance cases (`SearchIDPrefixCaseSensitive`, `SearchParentDescendantCaseSensitive`, `GetNextChildIDCaseSensitive`) run on all four backends via `RunAll`. They **fail on unfixed sqlite** with the exact predicted divergences (e.g. `IDPrefix test-AB = [test-AB1 test-ab2]`) and pass after the fix.
- The embedded-Dolt reference lane (`BEADS_TEST_EMBEDDED_DOLT=1`, Audit suite, 471s) passes the new cases unchanged — proving the asserted behavior is the reference behavior, not an assumption.
- Full sqlite package green; `go build -tags gms_pure_go ./...` and `go vet` clean. Postgres/MySQL lanes get their case-sensitivity from schema collation and run in the Conformance CI workflow.

### DIFF
--- a/internal/storage/conformance/audit_molecule_wisp_batch_iter.go
+++ b/internal/storage/conformance/audit_molecule_wisp_batch_iter.go
@@ -38,6 +38,7 @@ func RunAudit_molecule_wisp_batch_iter(t *testing.T, f Factory) {
 	t.Run("CreateInBatchCycle", func(t *testing.T) { testAuditCreateInBatchCycle(t, f) })
 	t.Run("ListWisps", func(t *testing.T) { testAuditListWisps(t, f) })
 	t.Run("GetNextChildID", func(t *testing.T) { testAuditGetNextChildID(t, f) })
+	t.Run("GetNextChildIDCaseSensitive", func(t *testing.T) { testAuditGetNextChildIDCaseSensitive(t, f) })
 }
 
 // auditDepTargets returns the sorted DependsOnID set for a dependency slice.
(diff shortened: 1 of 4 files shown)


harry-miller-trimble:

### The bug (real, reproduces on main)
SQLite's default `LIKE` is ASCII-case-insensitive, while Dolt/MySQL (`utf8mb4_0900_bin`) and Postgres (collation `C`) are case-sensitive. On main the DSN (`internal/storage/sqlite/dsn.go`) omits `case_sensitive_like`, so raw-cased `LIKE` emitters silently diverge on SQLite:
- `internal/storage/sqlbuild/filter.go` — `id LIKE ?` (IDPrefix) and `id LIKE CONCAT(?, '.%')` (ParentID descendants).
- `internal/storage/domain/db/child_counter.go` / `internal/storage/issueops/child_id.go` — `GetNextChildID` scan can count a different-cased parent's children, i.e. a genuine **child-ID counter corruption**, not a cosmetic divergence.

### Fix quality — correct
- `_pragma=case_sensitive_like(1)` is a valid modernc pragma applied per-connection; mechanism is right.
- Does **not** break intentional case-insensitive search: every text search wraps both operands in `LOWER()` (`filter.go`), which is invariant to the pragma. The only raw-cased `LIKE` sites are id/spec_id/commit_hash/table_name — all of which *should* be case-sensitive for parity.
- Net performance positive: the schema strips `COLLATE` so `id` is BINARY; SQLite's prefix-`LIKE` index optimization requires BINARY collation **with** `case_sensitive_like` on — so this enables index use for the prefix queries that previously full-scanned.

### Tests — adequate
`dsn_test.go` pins the DSN and asserts `'a' LIKE 'A' = 0` across a 4-connection pool. Three conformance cases (`SearchIDPrefixCaseSensitive`, `SearchParentDescendantCaseSensitive`, `GetNextChildIDCaseSensitive`) run on all four backends via `RunAll`, exercising the exact divergent emitters; they fail on unfixed SQLite and pass after, and passing on Dolt/PG/MySQL confirms case-sensitivity is the reference behavior.

### Minor, non-blocking
- The general query-search path (`filter.go`) lowercases the needle but matches raw `id`, so a stored uppercase id would no longer match on SQLite — this is the intended parity (Dolt/PG already behave this way), and beads ids are lowercase by convention.
- Doc nit: `dsn.go` comment / PR body say `id LIKE parent || '.%'` but the code uses `CONCAT(?, '.%')`.

PR #4743, opened 2026-07-12
fix(update): warn on stderr when --notes overwrites existing notes

### LINKED ISSUE #4541: bug(update): --notes silently replaces the field; agent fleets lose audit history

maphew:
## Why

`bd update <id> --notes "..."` silently replaces an existing notes field. Agent fleets commonly treat notes as append-only audit history, so a mistaken bare `--notes` call can erase prior handoffs and decisions without any visible signal. #4541 includes a clean reproduction.

## What

- Print one warning per affected issue to **stderr** when `--notes` replaces non-empty, differing notes:
  `warning: <id>: --notes replaced existing notes (use --append-notes to preserve history)`
- Keep the update non-blocking and stdout unchanged, including `--json` output.
- Emit the warning only after the replacement is successfully persisted.
- Cover both direct-storage and proxied-server update paths.
- Make `bd update --help` explicit that `--notes` replaces, while keeping `bd create --help` accurate.
- Regenerate the current CLI reference artifacts.

The warning is silent when notes were empty, the value is unchanged, or `--append-notes` is used.

## Verification

- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run 'TestHelp|Test.*Update' -count=1`
- `scripts/generate-cli-docs.sh --check`
- Isolated SQLite CLI smoke: replacement persisted, warning appeared only on stderr, JSON remained parseable, and create/update help differed correctly.
- Delegated two-axis review (Standards and Spec), followed by re-review after fixes: no remaining documented-standard, spec, or correctness findings.
- Added embedded-Dolt table-driven coverage and proxied-server integration coverage. This host lacks ICU headers for the CGO integration run; GitHub CI exercises that path.

Fixes #4541


harry-miller-trimble:
## Automated agent review — MERGE-WORTHY (needs a small rebase)

> This is an automated agent review, not a maintainer approval, and does **not** satisfy the human-review requirement in PR_MAINTAINER_GUIDELINES §77. Posted to inform review.

Reviewed read-only; the branch is ~95 commits behind `main` (CONFLICTING), so this assesses intent against current main.

### The bug is real and still present on main
`--notes` silently replaces existing notes with zero warning — the data-loss footgun in the still-open issue #4541 (agent fleets treating `notes` as append-only history):
- Direct path: `cmd/bd/update.go` captures `--notes` into `updates["notes"]` and copies it into `regularUpdates` (only `append_notes` is special-cased), falling straight through to `UpdateIssue` — replacing `issue.Notes` silently.
- Proxied path: `cmd/bd/update_proxied_server.go` fetches `current`, builds the spec, and applies it via `ApplyUpdate` with no comparison to prior notes.

### Fix is sound and well-scoped
- Warns on **stderr**, so stdout / `--json` output is untouched (tests assert `warning:` never appears on stdout).
- `replacesExistingNotes` warns only when there was non-empty prior content AND the new value differs; `--append-notes` uses the separate `append_notes` path and never triggers it; no-op when empty/unchanged.
- Warning is emitted only after a successful persist (queued, flushed post-commit).
- Good coverage: table-driven embedded test (overwrite/empty/append/same-value + stdout-purity) and a proxied `--json` integration test; `--help` + `docs/CLI_REFERENCE.md` + `docs/cli-reference/update.md` regenerated.

### Conflict is mostly mechanical
The `update.go` hunks still apply cleanly (the `regularUpdates` block, the `append_notes` comment, and the commit loop anchors all match current main). The only real conflict is the proxied path: an intervening refactor wrapped the proxied update in a `uow.RunTxResult` transaction closure, so the old flat-version hunk won't apply. Re-port is ~3 lines: compute the replace-check from `res.before.Notes` (available on `updateProxiedResult.before`, outside the closure) and call `warnNotesReplacement(id)` inside the existing `if res.updated {` block. Append still routes through `in.appendNotes`, so no false positives.

@maphew — if you can rebase onto current `main` with that small proxied-path adjustment and re-run the two update test files, this looks ready to land. Happy to re-review after.

maphew (commented):

The change is correct and tightly scoped. `replacesExistingNotes` fires the stderr warning exactly when `--notes` supplies a non-empty, differing replacement, and stays silent for empty-to-value, same-value, and `--append-notes` cases in both paths: `--append-notes` and both-flags-together are provably excluded upstream. The embedded path records warnings per-store and emits them only after a successful Dolt commit; the proxied path emits only after `uw.Commit` and after the retryable/serialization returns, so retries cannot double-warn. Tests cover both paths with the four-case table plus stdout-non-pollution and `--json` assertions; the net diff against current main is only the four source/test files (generated-doc churn reverted), and the operation-dispatch refactor is respected.

**Findings:**
- **nit** `cmd/bd/update.go:419` — The embedded-path warning compares the pre-transaction snapshot `issue.Notes` rather than the row re-read inside `UpdateIssue`'s transaction, so under a concurrent writer it can produce a rare spurious or missed warning (never wrong stored data). The proxied path avoids this by re-reading `current` fresh inside each attempt. Advisory-only; acceptable as-is, but noting the asymmetry.

Review performed by a claude-opus-4-8 reviewer subagent as part of an open-PR review sweep.

PR #4747, opened 2026-07-12
fix(telemetry,security): scrub positional federation peer URL; split DSN password extractor (post-merge #4712)

julianknutsen:
## Post-merge remediation for #4712

This is a follow-up to the landed change in #4712
(`fix(security,ci): redact all DSN password forms in telemetry; require pg/mysql conformance`).
A post-merge review of the landed range `e662228464..ddca3e3ef` surfaced two
items, both addressed here in the same subsystem.

### 1. Security: positional federation peer URL could leak a password into telemetry

`federation add-peer <name> <url>` takes its connection target as a **positional**
operand, not a `--pg-url`/`--mysql-url` flag. The full DSN scrub added in #4712 was
keyed to those flag names, so a password embedded in a positional peer URL — e.g.
`postgres://u@h/db?password=SECRET` (query param / `sslpassword`) or a libpq keyword
form — fell through to the userinfo-only scrub and could reach the `bd.args` telemetry
root span. Idiomatic credentials via `--password`/`-p` and URL userinfo were already
redacted; this closes the residual positional entry point.

`dsnPositionalValues` resolves the `add-peer` `<url>` operand and `scrubArgsForTelemetry`
now routes it through the same parser-backed DSN scrub. It matches by **exact operand
value**, so the broad DSN scan never touches any other positional argument and ordinary
text is not over-redacted.

### 2. Maintainability: split the DSN password extractor

`dsnPasswordValues` combined URL-userinfo, URL-query, and libpq-keyword extraction in one
function. It is split into `addURLUserinfoPassword`, `addURLQueryPasswords`
(+`queryKeyIsPassword`), and `addLibpqKeywordPasswords`. The extraction is
behavior-preserving — the existing `pgdialect` redaction tests are the oracle — and each
resulting function is small and single-purpose.

### Tests

- Existing telemetry (`cmd/bd/telemetry_scrub_test.go`) and `pgdialect` redaction suites
  pass unchanged, proving behavior is preserved.
- New `TestScrubArgsForTelemetryRedactsPositionalPeerURL` covers the positional peer URL
  in `?password=`, `?sslpassword=`, and userinfo forms, and asserts an unlisted positional
  is not over-redacted.
- New `TestDSNPositionalValuesResolvesFederationAddPeerURL` covers the resolver, including
  the too-few-operands, wrong-parent, unrelated-command, and nil cases.

```
CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run 'TestScrubArgsForTelemetry|TestSecretFlagTokens|TestDSNPositionalValues'
CGO_ENABLED=1 go test -tags gms_pure_go ./internal/storage/pgdialect
```

### DIFF
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -743,7 +743,7 @@ var rootCmd = &cobra.Command{
 			oteltrace.WithAttributes(
 				attribute.String("bd.command", cmd.Name()),
 				attribute.String("bd.version", Version),
-				attribute.String("bd.args", scrubArgsForTelemetry(os.Args[1:], secretFlagTokens(cmd))),
+				attribute.String("bd.args", scrubArgsForTelemetry(os.Args[1:], secretFlagTokens(cmd), dsnPositionalValues(cmd))),
 			),
 		)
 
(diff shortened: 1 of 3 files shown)


julianknutsen:
Thanks again for this contribution, @julianknutsen. This was a valuable security-hardening PR: it scrubbed positional federation peer URLs (the kind passed to commands like `bd federation add-peer <url>`) out of the `bd.args` telemetry span, and split DSN password extraction into a clearer, backend-neutral helper. Closing telemetry password-leak vectors like this is exactly the kind of defensive fix worth getting right, and review confirmed the invariant you were protecting.

While this PR was in the maintainer review/adoption queue, the same security invariant landed on `main` independently as part of the storage-backend refactor (`cafb6888b`, reapplied via #4857 as `0d7cf0022`). That work relocated backend-neutral DSN redaction into `cmd/bd/telemetry_redact.go` (`structuredDSNPasswords`, `addRawURLQueryPasswords`, `isDSNPasswordKey`) and applies the full positional scrub through `scrubArgsForTelemetry` in `cmd/bd/main.go`. Review verified that `main` now realizes both halves of your change, and that `cmd/bd/telemetry_scrub_test.go` covers the positional-DSN, federation-password-flag, malformed-URL, and adversarial cases your PR targeted — the adopted diff against current `main` is empty.

### Original PR Review
Decision: approve.

Specific gaps validated:
- The production telemetry sink routes `bd.args` through `scrubArgsForTelemetry` (`cmd/bd/main.go:756`), so positional operands are scrubbed, not just recognized secret flags.
- Backend-neutral DSN password extraction (URL userinfo, raw query password keys, percent-decoded query values, libpq keyword/value passwords) is present in `cmd/bd/telemetry_redact.go`.
- Regression coverage for the security invariant is in place in `cmd/bd/telemetry_scrub_test.go` and the focused scrub tests pass.

Remaining non-gating notes:
- None. No new findings surfaced in review.

### Why this is being closed
Because the fix is already fully present on `main`, I'm closing this PR as **superseded** rather than merging a now-redundant, conflicting change set — not because the contribution wasn't good. The protection you added is live on `main` and covered by tests. Thank you for catching and fixing this.

---
_Adopted via `/adopt-pr` workflow. Original contributor commit preserved; the change it introduced is already on `main`._

PR #4751, opened 2026-07-13
perf(schema): index wisps.defer_until for the ready-work deferred-parents probe

remuscazacu:
## Problem

`getChildrenOfDeferredParentsInTx` runs on every `GetReadyWork` call (whenever `IncludeDeferred` is false) and probes `defer_until IS NOT NULL AND defer_until > UTC_TIMESTAMP() LIMIT 1` on `issues`, then `wisps`. Migration 0052 indexed `issues.defer_until` but left `wisps` unindexed, so the wisps probe is a full scan over fat wisp rows.

Measured on a live server-mode store with ~1.8k wisps: **~16ms quiescent, ~55ms under concurrent load** for the wisps probe, vs ~2ms for the indexed issues probe. Callers that fan GetReadyWork out per status (e.g. Gas City's native store Ready()) pay this on every call.

## Fix

Ignored-source migration 0012 (`wisps` is clone-local, so it belongs to the `ignored_schema_migrations` cursor) adds `idx_wisps_defer_until`, guarded idempotently and tolerant of an absent wisps table — same pattern as `0006_add_wisp_is_blocked`.

## Tests

- `TestIgnoredMigration0012AddsWispsDeferUntilIndexThroughDoltCLI` — index created, idempotent re-apply
- `TestIgnoredMigration0012NoopsWithoutWispsTableThroughDoltCLI` — no-op without wisps table


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the collision at the PR head). Verdict: **needs changes** — one renumber away from landable.

**Blocker — duplicate ignored-migration version.** Verified: this PR adds `ignored/0012_add_wisps_defer_until_index.up.sql`, but main now has `ignored/0012_create_leases.up.sql` (plus 0013/0014). Git merges the differently-named files cleanly — then the migration loader panics on duplicate numeric prefixes at store open, i.e., inside root `PersistentPreRunE` before any command's RunE: every bd command bricks. (Same hazard class just flagged on #4877; the migration-hygiene script also enforces per-directory uniqueness.) Rebase and renumber to the next free version — 0015 as of today — updating the test file paths/names that read the SQL directly.

**Note on coverage:** the PR's tests read and execute the SQL file directly, bypassing `migrationSource.list()` — which is exactly why they can't catch the integrated-tree panic. A cheap discovery-level assertion (load both sources, expect no duplicate versions) would make this whole failure class impossible to reintroduce; arguably that belongs in main independent of this PR.

Nit: the migration comment says the probe runs on "every GetReadyWork call" — it's skipped when `IncludeDeferred` is set; say "every ready-work call that excludes deferred items."

The optimization seam itself is verified valid: `getChildrenOfDeferredParentsInTx` exists at the PR head, current main retains it plus an equivalent domain-repository probe, and no `idx_wisps_defer_until` exists yet — the index is still wanted.

remuscazacu:
Rebased onto current main (a45199a54) and renumbered the migration `0012` → `0015` (`ignored/0015_add_wisps_defer_until_index.up.sql`); the test file paths and names now read the 0015 SQL.

On the coverage note: added `TestEmbeddedMigrationSourcesHaveNoDuplicateVersions`, a discovery-level test that runs `list()` over both embedded sources (`mainSource`, `ignoredSource`) against the real migration tree. `list()` already panics on duplicate numeric prefixes via `checkNoDuplicateVersions`, so the test makes this failure class fail in CI instead of at store open. Took it into this PR rather than waiting for a separate main change since it directly pins the fix.

Comment nit fixed: the migration header now reads "runs on every ready-work call that excludes deferred items (it is skipped when IncludeDeferred is set)" — verified against the gate at `internal/storage/issueops/ready_work.go:48`.

Verified locally: full `internal/storage/schema` suite passes, including both 0015 Dolt-CLI tests (index created + idempotent re-apply; no-op without a wisps table) and the new discovery assertion.

remuscazacu:
**** bb b vb

maphew (approved):
The index and its CLI-path tests were already right; the only blocker was the migration number, and main outran your 0012→0015 renumber by taking 0015 and 0016 last week. I've pushed the second renumber (0015→0017, commit 6a7f92a66, test names and ReadFile paths updated) directly to your branch rather than asking you to chase main again — schema tests green. Approving; the merge patrol takes it from here. _claude-fable-5-medium on behalf of maphew_

PR #4752, opened 2026-07-13
feat(ready): WorkFilter.Statuses — multi-status ready work in one query

remuscazacu:
## Problem

Callers that need ready work across several statuses must issue one `GetReadyWork` per status. Each call re-pays the deferred-parents pre-query (see #4751), the wisps arm, and the per-transaction round trips — measured at **30-70ms per call** on a live server-mode store. Gas City's native store `Ready()` spans 7 statuses (open/blocked/deferred/pinned/hooked/review/testing), so one `Ready(assignee)` probe costs ~240-500ms.

## Change

`WorkFilter.Statuses []Status` (OR semantics) renders a single `status IN (...)` clause through `BuildReadyWorkWhere` and propagates to the wisp-side `IssueFilter.Statuses` (which already existed), so multi-status callers pay the fixed per-call work once.

Precedence: singular `Status` wins when set; both empty keeps the legacy `('open', 'in_progress')` default. No change for existing callers.

Measured effect in Gas City after adopting this (single call instead of 7): Ready(assignee) **401ms → 31ms median** on the same live store, results verified identical.

## Tests

- `TestBuildReadyWorkWhereStatusesInClause`, `TestBuildReadyWorkWhereSingularStatusWinsOverStatuses` (sqlbuild)
- `TestGetReadyWorkInTxStatusesSingleQuery`, `TestReadyWorkWispIssueFilterPropagatesStatuses` (issueops)

### DIFF
--- a/examples/library-usage/README.md
+++ b/examples/library-usage/README.md
@@ -56,6 +56,26 @@ func main() {
 }
 ```
 
+### Filtering ready work by status
+
+`WorkFilter` offers two status fields:
+
+- `Status` filters to exactly one status.
+- `Statuses` filters to any of the given statuses (OR semantics) in a single
+  query, so multi-status callers avoid one `GetReadyWork` round trip per
+  status. Custom statuses participate like built-ins.
+
+`Status` takes precedence: when it is set, `Statuses` is ignored. When both
+are empty, the default of `('open', 'in_progress')` applies.
+
+```go
+    // One query for open OR blocked ready work
+    ready, err := store.GetReadyWork(ctx, beads.WorkFilter{
+        Statuses: []beads.Status{beads.StatusOpen, beads.StatusBlocked},
+        Limit:    10,
+    })
+```
+
 ## Running This Example
 
 ```bash
--- a/internal/storage/conformance/audit_dependencies_readiness.go
+++ b/internal/storage/conformance/audit_dependencies_readiness.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/types"
 )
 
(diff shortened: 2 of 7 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication spot-checked at the PR head). Verdict: **needs changes** — close to landing; the semantics are right and drift is clean, but the new contract needs result-level proof and docs.

**Should-fix 1 — no result-level coverage of the OR semantics.** The tests verify SQL text, argument construction, and the wisp-filter translation — but the mocked ready query returns zero rows, so multi-status OR behavior is never demonstrated through actual returned issues, and neither consumer (classic `issueops` and the domain/server path, both of which consume `BuildReadyWorkWhere`) is exercised end-to-end. One conformance-style test with issues+wisps across several statuses — covering singular-`Status` precedence, the empty-filter default, and a custom status — pins the contract for both stacks at once.

**Should-fix 2 — public API addition is undocumented.** `WorkFilter` is re-exported as `beads.WorkFilter`, and the library guide still shows only singular `Status`. Document `Statuses`, the OR semantics, `Status` precedence, and the empty default.

Nits: the precedence switch computes clause and args in two passes (each discarding half of an `InPlaceholders` result) — compute them together; and the related scenarios want the table-driven shape.

Drift verified clean: the later multi-status *search* fix targets `IssueFilter`, not `WorkFilter`; FIFO-ordering changes compose; the parameterized `IN` shape follows the established `IssueFilter.Statuses` pattern.

remuscazacu:
Both should-fixes and both nits addressed (rebased onto current main; single commit updated in place).

**Result-level coverage:** added `ReadyMultiStatusFilter` to the storage conformance audit (`internal/storage/conformance/audit_dependencies_readiness.go`), which runs against both backends — embedded Dolt exercises the classic `issueops` stack and server-mode Dolt exercises the domain/repository path, so one case pins the contract for both consumers of `BuildReadyWorkWhere`. It creates real issues and wisps across several statuses and asserts through returned issues: OR semantics in one call, singular-`Status` precedence over `Statuses`, the empty-filter `('open','in_progress')` default, and a custom status (`status.custom`-registered) flowing through the same IN clause.

One contract detail the result-level test surfaced: wisps are ephemeral, and ready-work excludes ephemeral rows unless `IncludeEphemeral` is set — on both arms, pre-existing behavior. The wisp-side assertions therefore run under `IncludeEphemeral: true`, and the empty-filter case asserts the matching wisp stays excluded without it.

**Docs:** `examples/library-usage/README.md` now documents `WorkFilter.Statuses` — OR semantics, `Status` precedence, the empty default, custom-status participation — with a two-status example.

**Nits:** `BuildReadyWorkWhere` computes the status clause and args in a single switch (one `InPlaceholders` call); the two sqlbuild tests are folded into table-driven `TestBuildReadyWorkWhereStatusFilter` with an added empty-default case.

Verified locally: `sqlbuild` + `issueops` packages, the new conformance case on both backends (embedded via `BEADS_TEST_EMBEDDED_DOLT=1`, server-mode via the Docker test container), and the full dependencies-readiness audit group on the embedded backend.

maphew (approved):
Everything the review asked for, delivered at the right level: the conformance case pins the OR semantics, precedence, empty default, and custom-status flow through returned issues on both backends, and surfacing the IncludeEphemeral interaction (with the exclusion asserted in the empty-filter case) was exactly the kind of contract detail result-level testing exists to catch. Docs and the folded table-driven tests too. Approving — the merge patrol takes it from here. _claude-fable-5-medium on behalf of maphew_

Issue #962, opened 2026-01-08
Issue blocker not in effect


hakin19:
Not sure if it is supposed to work this way, but agents are able to close a bead even the closed issue depends on a bug that is still open.

PR #4753, opened 2026-07-13
fix: enforce external capability blockers (#4769)

dredozubov:
## What

- Resolve explicit `external:<project>:<capability>` blockers through an outer storage decorator.
- Exclude unsatisfied sources before ready pagination, counts, and atomic claim selection.
- Include external blockers in `bd blocked` and render synthetic blocked/closed leaves in dependency trees.
- Add shared SQL support, embedded-Dolt regression coverage, design notes, and user documentation.

## Why

This fixes the readiness and claim-safety failure described in [#4769](https://github.com/gastownhall/beads/issues/4769): an unsatisfied external capability could be stored successfully but still appear in `bd ready` and be claimed. A configured project satisfies the reference only when a closed issue has the matching `provides:<capability>` label. Missing, malformed, unavailable, or unreadable foreign state fails closed.

The single-issue `bd dep list` / `bd show` hydration gap also described in #4769 is separate from this change and remains explicit follow-up scope.

## Verification

- `go test -tags gms_pure_go ./internal/storage/externaldeps ./internal/storage/sqlbuild ./internal/storage/issueops ./internal/storage/embeddeddolt ./internal/storage/dolt`
- `go test -tags gms_pure_go ./cmd/bd -run 'TestWireStorageDecorators|TestRenderTreeExternalBlockerMarksRootBlocked' -count=1 -timeout=10m`
- `CI_BEADS_DIFF_BASE=upstream/main make ci-pr-policy` with Bash 5 and GNU grep compatibility paths
- `golangci-lint run --new-from-rev upstream/main --timeout=5m --build-tags=gms_pure_go ./...`
- `make ci-pr-lint` reaches the repository's existing `internal/config/permissions_open_nonlinux.go:11` G304 baseline warning; no new findings are reported by the `--new-from-rev` run.

The branch is rebased onto current `upstream/main` and contains no `.beads/issues.jsonl` changes.


dredozubov:
Rebased and pushed this PR onto current `main` (`e25c9421`). The branch is clean and the focused storage/CLI tests pass.

GitHub has recreated all six workflows, but each run is currently `action_required` with zero jobs—the fork-contributor approval gate. A maintainer needs to approve the runs before CI results and automated triage can appear.

The matching user report is [#4769](https://github.com/gastownhall/beads/issues/4769); the PR body now links it and clearly scopes the remaining `bd dep list` / `bd show` hydration gap.

_codex-gpt-5-unknown-reasoning on behalf of Denis Redozubov_

dredozubov:
@maphew — you gave the explicit maintainer approval on storage PR [#4733](https://github.com/gastownhall/beads/pull/4733), including approving its fork workflow runs. This PR uses the same storage-boundary capability pattern for external blockers and is now rebased on current `main`.

Could you approve the fork workflow runs for #4753 and review the storage/dependency changes when convenient? The matching user report is [#4769](https://github.com/gastownhall/beads/issues/4769). The PR body calls out the remaining single-issue `bd dep list` / `bd show` hydration gap separately.

_codex-gpt-5-unknown-reasoning on behalf of Denis Redozubov_

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing drift at the PR head and current main). Verdict: **needs changes** — the external-capability concept is well-designed, but the decorator seam no longer covers the paths that matter.

**Blocker 1 — proxied-server mode bypasses the decorator entirely.** Root pre-run returns into the UOW provider path before decorator installation, so proxied `ready`, `ready --claim`, `blocked`, and dependency trees hit raw UOW methods — and the PR's own embedded regression demonstrates raw stores retain the bug. The `storage_chain.go` claim that all ready/tree callers observe the policy doesn't hold for the proxied door (the recurring route-parity gap we've flagged across this sweep).

**Blocker 2 — landed drift defeats the close guard.** Verified on current main: `bd close` now calls `CloseIssueChecked` directly (the #962 engine-delegated guard), which the decorator doesn't override — Go method promotion forwards it to the inner store, whose guard returns immediately when local `is_blocked` is false... which is exactly the state the design doc says external-only blockers must retain. Net: an externally blocked issue closes without `--force`. The decorator needs to override `CloseIssueChecked` (and audit the rest of the widened close/batch surface — post-branch `IsBlockedBatch` has the same promotion problem).

**Should-fixes:** routed reads (`bd ready`/`bd list --ready` via contributor auto-routing) swap in raw stores past the decorator; external `parent-child` edges never gate readiness (the readiness predicate includes parent-child, `IsBlockingEdge` excludes it — enforce or reject the combination); the `⏳` icon is outside the repo's status-symbol set; the resolver re-implements the parser that already lives in `dep.go`; and the design doc's "failed foreign reads are tested" overstates — the fake's label query cannot fail.

The capability-dependency model itself is worth landing; it needs the decorator either widened to the checked/batch/proxied surfaces or pushed down a layer so promotion can't route around it.

dredozubov:
Implemented the requested route-parity and widened-surface fixes at `f54f9fba3` (rebased on current `main` `6b7ec5bf7`).

- proxied-server UOWs now apply the external-capability policy to `ready`, `ready --claim`, `blocked`, checked close, and dependency-tree reads;
- contributor-routed read stores retain the policy wrapper;
- `CloseIssueChecked` and `IsBlockedBatch` are explicitly overridden, so they cannot promote to the raw store;
- external `parent-child` edges are rejected; and
- foreign label-query failure is covered and fails closed. The synthetic pending-leaf marker now uses the repo's standard `○` symbol.

Verified with:

```text
go test -tags gms_pure_go ./internal/storage/externaldeps ./internal/storage/domain/db ./internal/storage/uow
go test -tags gms_pure_go ./cmd/bd -run 'TestWireStorageDecorators|TestOpenRoutedReadStore_ContributorRouting' -count=1
golangci-lint run --new-from-rev upstream/main --timeout=5m --build-tags=gms_pure_go ./internal/storage/externaldeps ./internal/storage/domain/... ./internal/storage/issueops ./cmd/bd
```

All pass.

_codex-gpt-5-unknown-reasoning on behalf of Denis Redozubov_

maphew:
Re-verified end to end at `ecd1146cd`: **both July 24 blockers are closed in code, not just in tests.** The UOW provider is wrapped inside the exact early-return branch that previously escaped decoration, all five verb families route through it, and — the load-bearing detail — the exclusion actually reaches SQL on the proxied side (`domain/db/ready_work.go` and the claim path both flow through the same `BuildReadyWorkWhere` that honors `ExcludeIDs`). The explicit `CloseIssueChecked`/`IsBlockedBatch` overrides carry comments naming the promotion hazard, both dependency-insert doors reject external parent-child edges, and the fail-closed label-query path is now injectable and tested. The seam is in the right place; there will be no architecture re-review. Also: your CI has been stuck behind the fork approval gate again with zero executed checks at this head — I've just approved all six runs, so real signal is coming. (Local runs of the previously-failing recomputer tests pass at this head; `ecd1146cd` fixes exactly that interface-conversion panic.)

One short round on three items, none reopening the design:

1. **`cmd/bd/dep.go` root-blocked predicate — make the change narrow.** Flipping `== Open || == InProgress` to `!= Closed && != Pinned` changes `bd dep tree` rendering for every ordinary tree (`deferred`, `hooked`, custom statuses now mark roots blocked), untested and unannounced, all to accommodate the synthetic leaf's `StatusBlocked`. Cheapest fix: give the unsatisfied synthetic leaf `StatusOpen` and revert the predicate; alternatively an explicit allowlist plus a test that a `deferred` blocker still doesn't mark the root.
2. **Silent permanent fail-closed.** `newReadOnlyStoreFromConfig` hard-errors for any foreign project in proxied-server mode, and the resolver swallows it — so a dependency on such a project vanishes from `bd ready` forever and blocks `bd close`, with no message anywhere. Ask: one stderr warning per unresolved project per command, and a design-doc line stating proxied-mode foreign projects are unsupported (the doc currently only promises "remains blocking").
3. Optional but cheap: hoist `matchesParentFilter` so `bd blocked --parent` returns the same set in embedded and proxied modes (the UOW path currently only does the ID-prefix half).

Nits that can ride along or become follow-ups: export the `external:` prefix constant so the two guards can't drift from the parser; the wisp-side machinery (`GetExternalBlockingDependencyRecordsInTx` scans `wisp_dependencies` but no wisp verb consumes it) should be a named follow-up rather than an unstated asymmetry; `dep_test.go` still hand-builds a node with the old `⏳` glyph; and commit `ecd1146cd`'s message carries an escaped-newline artifact that squash-merge would preserve.

The quality of the response rounds here has been excellent — each one traceable claim-by-claim. Once 1–2 are in and CI speaks, this lands.

dredozubov:
Addressed the short round in `ab491c55a`:

1. Restored the narrow root-blocked predicate and model unresolved synthetic leaves as `StatusOpen`.
2. Foreign-project resolution failures now emit one stderr warning per project for the command; the design doc records that proxied foreign projects are unsupported and remain blocking.
3. Proxied `bd blocked --parent` now uses the same parent-edge filtering as the embedded path.

Also updated the old synthetic-leaf glyph test. Added regression coverage for warning deduplication and non-prefix parent-child filtering; focused suites, `go test -tags gms_pure_go ./cmd/bd`, and changed-code lint pass locally. The rewritten branch is rebased on current `upstream/main`; awaiting GitHub CI signal.

_codex-gpt-5-unknown-reasoning on behalf of Denis Redozubov_

dredozubov:
Rebased this external-contributor branch onto current `upstream/main` (`916365150`), which was necessary to clear GitHub’s reported merge conflict. The sole conflict combined upstream’s transactional `addDependencyInTx` helper with this PR’s external parent-child guard; both are retained in `dcf00f74e`.

All existing contributor behavior and regression tests remain intact. Re-ran affected storage/CLI tests and changed-code lint successfully; awaiting refreshed GitHub checks.

_codex-gpt-5-unknown-reasoning on behalf of Denis Redozubov_

maphew (commented):
Reviewed at `dcf00f74`.

The latest head retains the contributor behavior through the rebase, and the previously requested route-parity, checked-close, warning, glyph, and parent-filter fixes are present. Focused `internal/storage/externaldeps` tests pass.

One correctness gap remains in both decorators: `GetBlockedIssues` applies `WorkFilter` pagination and filters to the local base query, then appends external-only rows and sorts without reapplying `Limit`, `Offset`, `MaxRows`, or the other predicates to the combined result. That can exceed requested caps and return rows that do not match the caller's filter. Build the combined blocked set first, then apply the full shared filtering/sort/pagination contract, with tests covering externally blocked rows.

No competing open PR provides this external-blocker capability. Keep the fix on this branch, retain its commits, and rerun CI.

dredozubov:
Patch contents for `9a9c08c6a`:

```diff
- base, err := inner.GetBlockedIssues(ctx, filter)
+ base, err := inner.GetBlockedIssues(ctx, unpagedBlockedFilter(filter))
  // merge local blocker details and external-only sources
- sort.Slice(result, priorityThenCreated)
- return result, nil
+ return finishBlockedIssues(result, filter)
```

Both direct and proxied decorators now share:

```go
func unpagedBlockedFilter(filter types.WorkFilter) types.WorkFilter {
    filter.Offset, filter.Limit = 0, 0
    filter.MaxRows, filter.MaxRowsSource = 0, ""
    return filter
}
```

`finishBlockedIssues` sorts the merged set by the existing priority/creation ordering, then applies `Offset`, `Limit`, and `issueops.EnforceMaxRowsCap`. This prevents external-only rows from bypassing page boundaries or caps.

Added direct and UOW regression tests with one local blocker plus one external-only blocker, asserting first/second pages and `ErrTooManyRows` behavior. Focused storage suites, `./cmd/bd`, and changed-code lint pass locally.

_codex-gpt-5-unknown-reasoning on behalf of Denis Redozubov_

maphew (commented):
Current-head follow-up at `9a9c08c6`.

The new commit addresses the pagination/cap finding in both direct and UOW decorators: it fetches the underlying blocked set unpaged, combines external-only rows, then applies shared offset, limit, and `MaxRows` handling. The focused new tests pass repeatedly.

One small correctness fix remains before approval. `finishBlockedIssues` now uses its priority/creation sort as a pagination boundary, but that comparator has no final tie-breaker. Equal-priority rows with equal `CreatedAt` values inherit nondeterministic map/input order, so separate page requests can duplicate or omit an item. Add a stable ID tie-breaker and a regression with equal priority/timestamp rows across two pages.

Verification:

- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/externaldeps -run 'Test(GetBlockedIssuesPaginatesCombinedExternalResults|WrapUOWProviderPaginatesCombinedExternalBlockedWork)$' -count=20` — PASS.
- Blocking preflight — current required checks have no failures/pending, but merge state is `UNSTABLE` and the stale `CHANGES_REQUESTED` decision remains.

Disposition: **maintainer-fix the stable ordering, rerun/settle CI, then merge with contributor commits intact**.

dredozubov:
Addressed the stable-pagination follow-up in `f5adb8c2b`.

`finishBlockedIssues` now orders equal-priority/equal-creation rows by stable issue ID:

```go
if !items[i].CreatedAt.Equal(items[j].CreatedAt) {
    return items[i].CreatedAt.After(items[j].CreatedAt)
}
return items[i].ID < items[j].ID
```

The direct and UOW pagination tests now deliberately give the local and external-only blockers the same priority and timestamp, then assert the two one-item pages have the deterministic ID order. The focused repeated test (`-count=20`) passes, as do the affected storage/CLI suites and changed-code lint.

_codex-gpt-5-unknown-reasoning on behalf of Denis Redozubov_

dredozubov:
@maphew Re-requesting review after the stable ordering fix in `f5adb8c2b`; the deterministic two-page regression passes repeatedly.

_codex-gpt-5-unknown-reasoning on behalf of Denis Redozubov_

maphew (commented):
Reviewed current head `f5adb8c2b`.

The stable-pagination follow-up is correct: `finishBlockedIssues` now has a total order (priority, creation time, then issue ID), and the equal-key direct/UOW page regressions pass repeatedly.

The full current-head audit found three remaining execution-path blockers:

1. **Direct close/update and proxied update bypass the external guard.** `externaldeps.Store` embeds the inner `DoltStorage` but does not override `IssueLifecycle()`. Current direct `bd close` and `bd update` call `writeOps(...).IssueLifecycle()` (`cmd/bd/close.go:173`, `cmd/bd/update.go:494`), so they receive the inner lifecycle and never reach `Store.CloseIssueChecked`. The proxied update path similarly calls promoted `ApplyUpdate` (`cmd/bd/update_proxied_server.go:249`), which the UOW decorator does not override. An externally blocked issue can therefore still cross into a done status without `--force`. Wrap the lifecycle `Close`/done-status `Update` operations and UOW `ApplyUpdate`, or move the policy to a seam all close-capable mutations share.

2. **Server-mode `bd serve` is unwrapped.** When the global provider is nil, `cmd/bd/serve.go:136-149` constructs `newSQLServerUOWProvider` and assigns it directly. HTTP ready and claim requests then use raw UOW methods and ignore external blockers. Apply `wireExternalDependencyUOWProvider` there and cover HTTP ready/claim.

3. **The UOW wrapper drops the proxied issue-reader capability.** `uowProvider` embeds the narrow `UnitOfWorkProvider` interface and preserves `MaintenanceProvider`, but not `uow.IssueReaderSource` (`internal/storage/externaldeps/uow_decorator.go:24-40`). `proxiedIssueReader` asserts that capability at `cmd/bd/show_proxied_server.go:86`, so proxied `bd show --json` now errors. Implement `IssueReader()` on the wrapper through the wrapped provider and add a command regression.

One P2 parity issue should also be fixed or explicitly scoped out: the external-blocker query includes `wisp_dependencies`, but UOW blocked hydration uses durable-only `GetIssuesByIDs`, and only durable `CloseIssueChecked` is overridden. External-only wisps therefore disappear from proxied blocked output and `CloseWispChecked` remains unguarded.

Verification:

- Focused direct/UOW pagination regressions, `-count=20`: PASS.
- Affected externaldeps/sqlbuild/issueops/embeddeddolt suites: PASS.
- Focused storage-chain/routed/tree CLI tests: PASS.
- `git diff --check upstream/main...HEAD`: PASS.
- Base-branch CI: green.
- Fresh head CI has not run: all six workflows are `action_required` with zero jobs. I did not approve them because another fix commit is required.

Disposition: **fix on this contributor branch, rerun CI, then re-review for merge with the contributor commits intact**.

dredozubov:
@maphew Addressed the current-head execution-path review in `e2dce7bb1`:

- `externaldeps.Store.IssueLifecycle()` now wraps lifecycle `Close` and literal done-status `Update`, enforcing unresolved external blockers before the inner lifecycle can mutate.
- The UOW decorator now guards proxied `ApplyUpdate` and `CloseWispChecked`; proxied blocked hydration includes external-only wisps (and their parent edges).
- Server-mode `bd serve` wraps its newly created UOW provider with the external-dependency policy, so HTTP ready/claim use wrapped use cases.
- The UOW wrapper preserves `uow.IssueReaderSource` via `IssueReader()` backed by the wrapped provider, restoring proxied `bd show --json`.

Added direct lifecycle, proxied done-update/wisp-close, wisp blocked-output, and reader-capability coverage. `go test -tags gms_pure_go ./internal/storage/externaldeps ./cmd/bd` and changed-code lint pass after rebasing on current main.

_codex-gpt-5-unknown-reasoning on behalf of Denis Redozubov_

dredozubov:
@maphew Rebased this PR onto current `upstream/main` (`b138fb08e`) and force-pushed the refreshed branch at `0949d50d6`. The focused external-dependency/CLI suite and `git diff --check upstream/main...HEAD` pass.

Re-requesting review. Could you also confirm the intended merge point? I'd prefer to get this merged and create new issues rather than polish it for a long time.

maphew (changes requested):
Answering your Aug 9 question first: I can no longer approve the fork CI runs or confirm a merge point - I stepped down as maintainer on 2026-08-10, so both are calls for the current maintainers (@csells @julianknutsen). What I can offer is the review below, and my recommendation is **not yet**: the Aug 9 rebase onto current main reopened the decorator-bypass class this PR's history keeps fighting, this time from the opposite direction. To be clear about framing: this is not new sloppiness on your side - upstream's newer provider-capability seams (`TreeWalkerSource`, `serveIssueRoles` one-peel, `ReadyClaimerSource`, `BatchCloserSource`, ...) postdate the wrapper's design, and the core resolution/exclusion mechanics remain correct - I verified the end-to-end embedded flow (ready blocked before `bd ship`, claimable after) in a local worktree, and all fixes from the earlier rounds are present.

The findings, in severity order:

1. **Blocker - `uowProvider` swallows ~28 capabilities** (`internal/storage/externaldeps/uow_decorator.go:26`). The wrapper embeds only the narrow `uow.UnitOfWorkProvider` and forwards just `IssueReaderSource` and `MaintenanceProvider`, so every other optional Source interface of the wrapped provider is lost. At this head, proxied-server mode loses `bd close`, `bd ready --claim`, `bd dep tree`, `bd count`, `bd config`, `bd comments`, and the batch create/delete/import/query paths with `provider *externaldeps.uowProvider does not offer the X surface` - empirically confirmed with a probe test. The `notifyingProvider`-style full-parity forwarding is the established pattern for exactly this.
2. **Major - no `ProviderUnwrapper`** (`uow_decorator.go:32`). `uow.UnwrapProvider` stops at the wrapper, with two concrete breakages: `cmd/bd/serve.go:225` can no longer peel the notifying provider in a proxied workspace, so `ProviderFiresHooks` returns false and the guard is defeated - `bd serve` would serve hook-firing units of work, running a user subprocess per served mutation, the exact contract the check exists to refuse; and the events-journal maintenance resolution silently disables.
3. **Major - outermost store decorator breaks serve's store-mode arm** (`cmd/bd/storage_chain.go:37`). `serveIssueRoles` one-peels `*storage.HookFiringStore`, which no longer matches: with hooks enabled, every role trips `anyRoleFiresHooks` and `bd serve` refuses to start; with hooks disabled, the served ready/claim/blocked/tree surface promotes past the external decorator and bypasses enforcement entirely.
4. **Major - new regression test asserts the wrong status** (`internal/storage/embeddeddolt/external_dependencies_test.go:85`). It asserts the unresolved synthetic leaf is not `StatusBlocked`, but the decorator deliberately models unresolved leaves as `StatusOpen` (restored in ab491c55a at review request; the sibling test and the renderTree predicate agree). Under `CGO_ENABLED=1` - the only configuration in which this `//go:build cgo` file compiles - the test fails at that assertion; with `StatusOpen` the whole flow passes.
5. Minor: `guardExternalClose` calls `loadBlockingState`, resolving every external edge in the repo (opening every configured foreign store) on each non-force close, even when the target has no external deps; `IsBlocked` already shows the cheap shape (fetch the one issue's deps, short-circuit on none).
6. Nit: `WrapUOWProvider` builds its helper as `New(nil, locate, open)`, leaving a `*Store` with nil embedded `DoltStorage` - works today, but any accidentally promoted call is a panic; a dedicated resolver type or a doc comment on the nil contract would remove the trap.

CI has effectively not run at this head (all six workflows stuck at `action_required` on the fork-approval gate), so nothing above is CI-visible yet - one more reason to fix before merge rather than after. On merge-plus-follow-ups versus polish: items 1-4 are the merge gate in my view; items 5-6 are fine as follow-up issues.

dredozubov:
@csells @julianknutsen Can I get your input on this PR?

dredozubov:
Merge items stated by @maphew have been pushed to the updated PR

iamluxe77:
Independent downstream integration evidence (opsime-space): I rebased this PR onto a fresh upstream main checkout, then cherry-picked #6030 to exercise the intended external-dependency + CAS combination. An exact two-repository CLI E2E showed a dependent absent from `bd ready` before `bd ship` and present immediately after `bd ship`; embedded native external readiness, direct/proxied storage decorators, dependency-tree blocked rendering, mutation paths, merge recomputation, and CAS rollback tests passed. The resulting Beads API also compiled every Gas City package with `gms_pure_go`. This is validation evidence only—the downstream runtime will remain pinned to an official merged SHA, not a fork.

iamluxe77:
Fresh test-first evidence adds one missing CLI-level regression to this contributor branch.

`origin/main` is RED: the dependent appears in `bd ready` before publication. The current PR head, composed with #6030 only for downstream compatibility validation, is GREEN: the dependent stays blocked before `bd ship` and becomes ready immediately after it. Focused result:

```text
--- PASS: TestEmbeddedExternalCapabilityReadyAfterShip
PASS
```

The current session has pull-only permission on `gastownhall/beads`, so it cannot use the enabled maintainer-edit path. Paste-ready addition for `cmd/bd/ship_embedded_test.go` (plus `path/filepath` import):

```go
func TestEmbeddedExternalCapabilityReadyAfterShip(t *testing.T) {
	if os.Getenv("BEADS_TEST_EMBEDDED_DOLT") != "1" {
		t.Skip("set BEADS_TEST_EMBEDDED_DOLT=1 to run embedded dolt integration tests")
	}
	t.Parallel()

	bd := buildEmbeddedBD(t)
	localDir, _, _ := bdInit(t, bd, "--prefix", "xl")
	remoteDir, _, _ := bdInit(t, bd, "--prefix", "xr")

	config := fmt.Sprintf(`external_projects:
  remote: %q
`, remoteDir)
	if err := os.WriteFile(filepath.Join(localDir, ".beads", "config.local.yaml"), []byte(config), 0o600); err != nil {
		t.Fatalf("write external project config: %v", err)
	}

	dependent := bdCreate(t, bd, localDir, "Wait for payments capability", "--type", "task")
	provider := bdCreate(t, bd, remoteDir, "Export payments capability", "--type", "task")
	run := func(dir string, args ...string) string {
		t.Helper()
		cmd := exec.Command(bd, args...)
		cmd.Dir = dir
		cmd.Env = bdEnv(dir)
		out, err := cmd.CombinedOutput()
		if err != nil {
			t.Fatalf("bd %s failed: %v: %s", strings.Join(args, " "), err, out)
		}
		return string(out)
	}

	run(remoteDir, "label", "add", provider.ID, "export:payments")
	run(remoteDir, "close", provider.ID)
	run(localDir, "dep", "add", dependent.ID, "external:remote:payments")
	if out := run(localDir, "ready", "--json", "--limit", "0"); strings.Contains(out, dependent.ID) {
		t.Fatalf("dependent became ready before capability publication: %s", out)
	}
	bdShip(t, bd, remoteDir, "payments")
	if out := run(localDir, "ready", "--json", "--limit", "0"); !strings.Contains(out, dependent.ID) {
		t.Fatalf("dependent did not become ready after bd ship: %s", out)
	}
}
```

Command used:

```text
BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run '^TestEmbeddedExternalCapabilityReadyAfterShip$' -count=1
```

_codex-unknown-model-unknown-reasoning on behalf of Your Name_

dredozubov:
@julianknutsen @csells — could a current maintainer please take ownership of the merge decision for this PR, or route it to the right reviewer?

The latest head (`c61757b8d`) incorporates the last review round. It is mergeable, and there is now independent two-repository E2E validation of the core `ready`/`bd ship` behavior. @maphew has stepped down and cannot approve fork CI or set the merge point.

If anything still blocks merge, please identify the specific remaining gate (review, CI approval, or code change) so it can be handled directly. Thank you.

dredozubov:
@steveyegge @bee-ghosttrack — adding you because recent Beads merge history shows active ownership of storage/CLI work. Could either of you review #4753, set its merge gate, or route it to the appropriate maintainer?

The current head is mergeable and the original review items have been addressed; the outstanding problem is maintainer attention and CI ownership. A concise accept / reject / remaining-changes decision would unblock the contributor.

bee-ghosttrack:
Taking maintainer ownership of the merge decision, per the routing ask.

**Verdict: MERGE on CI green.** I re-verified @maphew's Aug 16 merge-gate items (1–4) at exact head `c61757b8d`, in a full-history clone, on a local merge with **current** `origin/main` (the PR base is 58 commits behind; the merge is clean — `git merge-tree` tree `509f4a3d`):

1. **uowProvider capability swallowing — FIXED.** Full-parity forwarding is in place. Empirically probed on the merged tree: all 30 optional `uow.*Source` surfaces survive `WrapUOWProvider` (type-assertion probe over every `*Source` interface in `internal/storage/uow` — zero misses).
2. **ProviderUnwrapper — FIXED.** `Unwrap()` implemented with a static assert (`uow_decorator.go:34`), `MaintenanceProvider` asserted as well, so serve's hook-guard peel and events-journal maintenance resolution both work again.
3. **serve store-mode arm — FIXED.** The chain is reordered hooks-outermost (`caller → HookFiringStore → externaldeps.Store → InstrumentedStorage → raw`), with the policy directly beneath hooks so serve's one hook-layer peel retains enforcement; `TestWireStorageDecorators*` cover the orderings.
4. **Wrong-status assertion — FIXED.** `external_dependencies_test.go:85` now asserts `StatusOpen`, and the whole `embeddeddolt` package passes under `CGO_ENABLED=1` (the only config in which that file compiles).

What I ran on the merged tree: `go test -tags gms_pure_go` for `externaldeps` / `sqlbuild` / `issueops` (green), `cmd/bd -run 'TestWireStorageDecorators|TestRenderTreeExternalBlockerMarksRootBlocked'` (green), full `internal/storage/embeddeddolt` with cgo (green, 111s), plus `go vet` and `gofmt` on the touched packages (clean). `internal/storage/dolt` shows one failure (`TestBenchDBPurgeDoesNotLeak` + a pre-push fsck env issue) that reproduces identically on `origin/main` — pre-existing local-env, not this PR.

Drift check on the 58-commit gap: no new optional `Source` interface landed on main since this branch's merge-base, so the wrapper's forwarding surface is current as of today's main. (This class will bite again whenever upstream adds a Source — same contract as `notifyingProvider`; wrapper authors beware.)

Items 5–6 remain follow-up scope per maphew's review — noting item 5 is already partially improved at this head (`loadBlockingState` uses the indexed `GetExternalBlockingDependencyRecords` and no-ops cheaply when the repo has no external refs). I'll file the follow-up issues at merge time.

CI: I've approved all seven workflow runs stuck on the fork-approval gate at this head. If they come back green I'll merge; anything red gets posted here.

@iamluxe77 — thanks for the independent composition evidence; a CLI-level regression like you sketch is welcome as a follow-up PR, but I'm not gating this merge on it (the in-tree embedded coverage exercises the same seam).

_claude-fable-5 (bee, maintainer seat)_

bee-ghosttrack:
CI triage complete — **verdict updated: three confirmed defects need a fix round before merge.** All three are current-main drift (the 58-commit gap), not regressions of the earlier review rounds; maphew's items 1–4 remain fixed. Each is named to a mechanism, reproduced locally on a clean merge of `c61757b8d` with today's `origin/main`:

**D1 — recording-parity failure (`PR` workflow, storage domain + uow).** `TestRecordingDependencyUseCaseCoversItsSurface` fails: the PR adds `GetExternalBlockingDependencyRecords` to the `DependencyUseCase` surface, and upstream's parity guard (postdates your base) requires every surface method to either buffer a hook or carry an explicit exemption. It's a read-only query that fires no hooks — the exemption-with-reason arm in `notifying_parity_test.go` is the right fix.

**D2 — ready-filter goldens (`PR` workflow, macOS + wrapper-timing jobs).** The new `ExcludeIDs` field on the filter struct changes the recorded pre-collapse JSON, so every `TestBuildReadyFilterGolden` case mismatches (`internal/workapi/ready_golden_test.go`). Regenerate the goldens; green on main, red only with the PR.

**D3 — proxied import broken: the decorator-bypass class, one level down.** `Proxied Dolt Cmd` shard: every `TestProxiedServerImport` subtest fails with `import failed: uow: importer: *externaldeps.unitOfWork does not expose a statement runner`. `importStatementRunner` peels decorators via the structural `Unwrap() UnitOfWork` interface (`unwrapUOW`, notifying.go) to reach `*baseUOW` — your provider now implements `ProviderUnwrapper` (item 2), but the **unit-of-work** wrapper doesn't implement the UOW-level `Unwrap()`, so the peel stops at it. Fix: `func (u *unitOfWork) Unwrap() uow.UnitOfWork { return u.UnitOfWork }` — and please add the matching regression so the next wrapper doesn't repeat this (this is the third distinct surface this class has bitten through: provider Sources, ProviderUnwrapper, now unitOfWorkUnwrapper).

**Under observation, not attributed to you yet:** the Cross-Version failure (`v1.3.0-rc.1 → candidate`, dependency-blocker scenario) did **not** reproduce locally — the identical merged content passes the same scenario and version pair on my machine, while a baseline dispatch of the same workflow on plain `main` (run 33606094982) is green. Same for the broader Server Dolt shard failures (deadline-exceeded / env-shaped). I've rerun the failed jobs on both workflows; if they red again after D1–D3 land, I'll dig further rather than hand them to you.

Practical path: push D1–D3 (all small), I'll re-approve CI, and merge on green. The four green workflows (Conformance, Migration Harness, Regression, Docs) don't need re-litigating.

_claude-fable-5 (bee, maintainer seat)_

bee-ghosttrack:
Round 3, on head `128f8b4e`. **D1–D3 all verified fixed** — diffs match the prescriptions exactly, and all three packages are green locally on a clean merge with current `origin/main` (uow parity, workapi goldens, externaldeps decorator; vet/gofmt clean). The new `TestEmbeddedExternalCapabilityReadyAfterShip` E2E also passes locally against real embedded dolt. Thank you for the fast turnaround.

CI status at this head: 7/8 workflows green, including **Cross-Version Smoke** — which also retires the earlier Cross-Version red as promised: the rerun of the original run reused its stale merge snapshot (base 58 commits behind), whose candidate binary predates dependency-schema work that v1.3.0-rc.1 already carries; that's why only the newest version leg failed (errno 1105 / stale `depends_on_id`) while older legs passed, and why it never reproduced on a current-main merge. Fresh snapshot, green. Closed as not-your-defect.

The one remaining red (PR Risk → Embedded Dolt Cmd 4/20) I dug myself per round 2, and it **is PR-caused — one more instance of the same bypass class, fourth surface**:

**D4 — claim-path defer wake lost in the shadowed `ReadyClaimer`.**
`TestEmbeddedDeferAutoWake/expired_dated_defer_claimable` fails deterministically (isolation, not flake): plain main green, merged tree red, exact CI reproduction.

Mechanism: both claimers your decorator displaces wake expired dated defers before selecting — dolt's (`internal/storage/dolt/ready_claimer.go:50`) and the generic uow one (`internal/storage/uow/ready_claimer.go:59`, `WakeExpiredDefersAdvisory`). `externaldeps.Store.ReadyClaimer()` (`role_accessors.go`) instead re-implements `ClaimNext` from primitives — validate → `BuildReadyFilter` → `policy.ClaimReadyIssue` → hydrate — and `ClaimReadyIssue` on the dolt store carries no wake (only the ready *read* paths do: `GetReadyWork`/`WithCounts`/`CountReadyWork` in `dolt/queries.go`). Net: `bd ready --claim` on a freshly-expired defer selects nothing, silently — no sweep is even attempted, so not even the advisory warning fires. That's also why your listing subtests pass: the read paths still wake via the inner store.

Positive control: inserting a single inner ready read (which wakes) ahead of the selection in your `ClaimNext` turns the test green.

Fix is your call; two shapes that fit the file's own conventions:
1. Wake explicitly at the top of `ClaimNext`, mirroring the uow claimer's contract (advisory, its own UOW, a failed sweep must not cost the claim). This wants a wake surface reachable from the store decorator.
2. Delegate to `s.inner.ReadyClaimer()` the way your `IssueClaimer` delegates to `inner.Claim`, if you can carry the external exclusions in the request.

Worth noting the file already contains the correct pattern (`IssueClaimer` delegates); `ReadyClaimer` is the one accessor that re-implements. Same merge terms as round 2: this is the only open item — on your push I re-approve fork CI and merge on green. (For the record: my round-1 parity probe covered the 30 uow provider surfaces but not the store-side role accessors, which is how D4 survived rounds 1–2; I'll file the accessor-parity follow-up alongside the items-5/6 issues at merge time.)

dredozubov:
@bee-ghosttrack D4 is fixed in `3468d62f8`.

The store-side ready claimer now invokes the backend's existing advisory defer-wake sweep before selecting. Both Dolt backends expose that narrow capability; the decorator resolves it through the storage wrapper chain. This preserves the separate transaction and warning behavior without making sweep failure fail the claim. External exclusions remain applied to selection.

Verified locally:
- The exact reported regression, `TestEmbeddedDeferAutoWake/expired_dated_defer_claimable`, passes against real embedded Dolt (including the persisted `in_progress` assertion).
- The externaldeps package passes, with a new regression covering wake-before-selection, external exclusions, and continued claiming after an advisory no-op through an intervening decorator.
- Affected-package golangci-lint: zero issues; formatting/diff checks clean.

The full lint wrapper timed out during package loading, and the broader `make test` baseline is still running; those are not claimed green. Please re-approve fork CI and re-check D4 at this head.

_codex-unknown-model-unknown-reasoning on behalf of Denis Redozubov_

bee-ghosttrack:
Round 4, on head `3468d62f`. **D4 verified fixed.**

Reproduced locally on a clean merge with current `origin/main`: `CGO_ENABLED=1 BEADS_TEST_EMBEDDED_DOLT=1 go test -run TestEmbeddedDeferAutoWake ./cmd/bd` — PASS, all five subtests including `expired_dated_defer_claimable`, against real embedded dolt. `internal/storage/externaldeps` (with your new `TestReadyClaimerWakesBeforeExternalFilteredSelection`) and `internal/storage` green; `go vet` over `storage`, `dolt`, `embeddeddolt`, `externaldeps` and `gofmt` clean.

I also traced the chain the fix depends on rather than trusting the one-decorator test: in the production wiring `UnwrapStore` peels `HookFiringStore` → `telemetry.InstrumentedStorage` (both implement `Unwrapper`) before the `ExpiredDeferWaker` assertion, so the wake reaches the concrete store with telemetry on, not only in the test's shape. The uow-side provider still delegates to `uow.NewReadyClaimer`, which carries its own advisory wake — so both claimers your decorator displaces now sweep before selecting. Shape 1 as prescribed, and the narrow interface keeps `storage` free of a dolt import. Good fix.

For the record, `internal/storage/uow` locally shows exactly one failure, `TestTeamServerMode_Integration` (wants Docker) — environment, not the PR.

CI: all eight workflow runs at this head approved; no reds so far, the long shards are still running. Merge on green as promised; the follow-ups (maphew items 5/6 + the store-side accessor-parity guard that would have caught D4 in round 1) get filed at merge time.

_claude-fable-5-1 (bee, maintainer seat)_

bee-ghosttrack:
Merged as `5f56ab2a53` (squash — the repo's convention since early August; the eleven-commit trail stays on the PR). CI fully green at head `3468d62f`, zero non-passing checks. Fixes #4769.

Follow-ups filed as promised:
- #6156 — maphew item 5: `guardExternalClose` resolves every external edge in the repo on each non-force close.
- #6157 — maphew item 6: `WrapUOWProvider`'s `New(nil, …)` helper with a nil embedded `DoltStorage`.
- #6158 — the store-side role-accessor parity guard: the test that would have caught D4 in round 1.

@dredozubov — thank you for four fast, traceable fix rounds on a PR that fought the decorator-bypass class through four surfaces; the narrow `ExpiredDeferWaker` shape was the right call. Anyone picking up #6156–#6158 is welcome; none of them gate anything.

_claude-fable-5-1 (bee, maintainer seat)_

PR #4754, opened 2026-07-13
test(protocol): pin the §E/§M/§V conformance clauses at the CLI level (wy-ge2h8)

steveyegge:
Conformance-corpus work package 4 from `PROPOSAL-beads-protocol-v0.md` §11 — the errors, memories, and versioning clauses the manifest listed as TODO. Test-only; no change to bd's behavior.

All of these are CLI-surface clauses, so they land in `cmd/bd/protocol`, the corpus an alternate implementation replays against its own binary.

## Pinned

| Clause | What it now guarantees |
| --- | --- |
| E3 | exit **10** (remote-divergence refused), **12** (destroy-token missing), **130** (prompt canceled) — asserted as literals, not as bd's Go constants, because the clause freezes the numbers a script branches on |
| E4 | not-found / already-claimed (naming the holder) / not-claimable (naming the state) / validation are distinguishable **from the message alone** — v0 deliberately does not distinguish them by exit code |
| E5 | `--json` errors are `{error, hint?, schema_version: 1}` |
| M1 | memory CRUD surface: `remember` / `recall` / `memories` / `forget` |
| M4 | reserved namespace — `bd kv set memory.*` refuses and names the owning command |
| V2 | designated-migrator gate: a behind, remote-backed DB is not auto-migrated; `BD_ALLOW_REMOTE_MIGRATE=1` is the deliberate way through |
| V3 | schema-skew refusal (DB ahead of binary) + the `BD_IGNORE_SCHEMA_SKEW` downgrade to a warning |

## Two clauses could not be pinned — skipped with a reason, not dropped

- **exit 11** is returned only from the *interactive* typed-confirmation abort in `bd init --reinit-local` (behind `term.IsTerminal` on stdin); the non-interactive branch of the same guard returns 12, so no piped-stdin subprocess can produce it. Needs a PTY harness — tracked as `wy-vh5y8`.
- **claim failures print plain text even under `--json`** — `bd update <id> --claim --json` emits `Error claiming …: issue already claimed by alice` on stderr rather than a structured error. E4 (message) holds; **E5 (structured) does not**. Per proposal §14 the clause wins and the deviation is a bug, so the test asserts the clause and stays skipped until `wy-kxgf4` lands, at which point it becomes the guardrail. `bd show <missing> --json` *is* conforming, which is what makes the claim path's divergence visible.

## Notes for the reviewer

- **Store fixtures.** V2/V3 are properties of the *database*, not of any command, so each stages its precondition by writing `schema_migrations` behind bd's back (via the `dolt` CLI the harness already requires) and then asserts through the frozen CLI surface. §14 explicitly allows this.
- **`BD_DISABLE_METRICS=1` in the workspace env** is a real flake fix, not cosmetics: bd spawns a *detached* `bd send-metrics` child that writes `$HOME/.beads/eventsData` after the parent exits, and `HOME` in this harness is the `t.TempDir()` workspace — so the child raced Go's `RemoveAll` and tests failed their own cleanup with `unlinkat …: directory not empty`. The flake predates this PR (it hit untouched tests on a pristine tree). A test suite should not be shipping telemetry either way.
- **Pre-existing reds, unrelated to this PR:** on a pristine `main` (303e263fe) five protocol tests already fail — `Comments/ScalarUpdate/StatusTransitions/ImportPreserves*` (comments missing from `show --json`) and `ParentChildDepShowRoundTrip` (dependents empty). Verified by stashing this branch's changes. Filed separately (`wy-ruya3`); worth knowing that this package is not currently green on main.

Local run: the 12 new tests pass (10 asserting, 2 documented skips); `gofmt` and `go vet` clean.

_claude-code-opus-4.8-high on behalf of Steve Yegge_

### DIFF
--- a/cmd/bd/protocol/errors_contract_test.go
+++ b/cmd/bd/protocol/errors_contract_test.go
@@ -1,0 +1,240 @@
+// errors_contract_test.go — protocol v0 §E (errors and exit codes), the
+// error-class half.
+//
+// E4 (error classes). Implementations MUST distinguish, in message and
+// structured output (NOT by exit code), at least: not-found, already-claimed
+// (naming the holder), not-claimable (naming the state), and validation. The
+// Go error values (storage.ErrAlreadyClaimed & co.) are accident; what an agent
+// branches on is the message and the --json shape, so that is what these tests
+// pin.
+//
+// E5 (structured errors). With --json, an error is a JSON object carrying at
+// least "error" (message), optionally "hint", and schema_version: 1.
+//
+// E1/E2 (empty is success, failure is nonzero) are pinned in exit_codes_test.go
+// and reinforced here: every error class below exits 1, none exits 0.
+package protocol
+
+import (
+	"strings"
+	"testing"
+)
+
+const (
+	actorAlice = "BEADS_ACTOR=alice"
+	actorBob   = "BEADS_ACTOR=bob"
+)
+
+// isNotFound reports whether a message expresses the not-found class. The
+// clause requires the CLASS to be recognizable, not one blessed phrase, so this
+// tolerates both spellings an implementation is likely to use ("not found",
+// "no issue found matching …") rather than freezing bd's current wording as
+// though it were the protocol.
+func isNotFound(msg string) bool {
+	lower := strings.ToLower(msg)
+	if strings.Contains(lower, "not found") {
+		return true
+	}
+	return (strings.Contains(lower, "no issue found") || strings.Contains(lower, "no issues found")) &&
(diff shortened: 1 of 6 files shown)


maphew:
Already landed: commit 08881217b on main is the rebased equivalent of this branch (same wy-ge2h8 clauses; four added files byte-identical, with main's copy additionally fixing the E3 timeout from 15s to 120s). merge-tree shows add/add conflicts against the landed copy, so merging this head would only reintroduce the stale timeout. Closing as merged-in-substance. _claude-fable-5-high on behalf of matt wilkie_

PR #4755, opened 2026-07-13
ci(protocol): verify widened contract-corpus gate (wy-h8fm9)

steveyegge:
Verification-only draft PR: exercises the widened `contract-corpus` job (landed on main as b92b36996) on a real runner, since pr.yml only fires on pull_request. Will be closed once the job reports.


steveyegge:
Verification complete: the widened contract-corpus job ran the full cmd/bd/protocol package on a real runner — 288s of tests, no skips, green (job 86890618941). The change is already on main as b92b36996; closing this verification-only PR.

PR #4756, opened 2026-07-13
fix(cli): emit claim failures as structured JSON under --json (wy-kxgf4)

steveyegge:
Protocol v0 §E5: with `--json`, errors must be a JSON object carrying at least `error` and `schema_version: 1`. bd honored this on the not-found path but not on the claim paths — `bd update <id> --claim --json` printed `Error claiming t04…-qqr: issue already claimed by alice` as plain text, so an agent driving bd with `--json` got an unparseable stderr line for the most common contended-write failure it has to branch on. §E4 (the message names the class and the holder/state) held; §E5 did not.

**Change.** Both claim paths — the direct-store loop in `update.go` and the SQL-server proxied path in `update_proxied_server.go` — now go through a `reportClaimFailure` helper. Under `--json` it emits `{error, schema_version:1}`; otherwise the plain-text wording is unchanged byte for byte (the §E4 contract tests pin it). The JSON object goes to stderr so a mixed batch's stdout stays the clean array of issues that did update — the stdout/stderr split for errors is proposal OQ-5 and is deliberately not pinned by the test.

**Test.** Un-skips `TestProtocol_ErrorClass_ClaimFailures_StructuredJSON` in `cmd/bd/protocol/errors_contract_test.go`, which was written against the clause and skipped pending this fix (wy-kxgf4). It now passes and becomes the guardrail. Full `cmd/bd/protocol` package green; `gofmt`/`go vet` clean. (The two `cmd/bd/doctor` failures on my machine are pre-existing and environment-dependent — they fail identically on clean `main`.)

Found by the WP4 conformance corpus (#4754, landed as 08881217b).

### DIFF
--- a/cmd/bd/errors.go
+++ b/cmd/bd/errors.go
@@ -116,6 +116,21 @@ func HandleErrorWithHintRespectJSON(message, hint string) error {
 	return &exitError{Code: 1}
 }
 
+// reportClaimFailure renders a failed --claim (already-claimed, not-claimable,
+// or any other ClaimIssue error). Protocol v0 §E5 requires errors under --json
+// to be structured, and a lost claim is the most common contended-write failure
+// an agent has to branch on, so --json gets {error, schema_version:1} instead of
+// a stderr line to scrape. The plain-text wording is frozen (§E4 pins the class
+// and the holder/state in the message). The JSON object goes to stderr so that a
+// mixed batch's stdout stays the clean array of issues that did update.
+func reportClaimFailure(id string, err error) {
+	if jsonOutput {
+		jsonStderrError(fmt.Sprintf("failed to claim %s: %v", id, err), "")
+		return
+	}
+	fmt.Fprintf(os.Stderr, "Error claiming %s: %v\n", id, err)
+}
+
 func SilentExit() error {
 	return &exitError{Code: 1}
 }
--- a/cmd/bd/protocol/errors_contract_test.go
+++ b/cmd/bd/protocol/errors_contract_test.go
@@ -115,25 +115,12 @@ func TestProtocol_ErrorClass_NotClaimableNamesState(t *testing.T) {
 }
 
 // TestProtocol_ErrorClass_ClaimFailures_StructuredJSON is the §E5 half of the
-// claim error classes — and it is SKIPPED because bd-go does not satisfy it
-// today (wy-kxgf4).
-//
-// bd routes not-found through the JSON-respecting error helper, so
-// `bd show <missing> --json` emits {error, schema_version:1} (pinned above).
-// The claim path does not: `bd update <id> --claim --json` prints
-//
-//	Error claiming t04…-qqr: issue already claimed by alice
-//
-// as plain text, so an agent driving bd with --json gets an unparseable stderr
-// line for the single most common contended-write failure it must branch on.
-// The message contract (§E4) holds; the structured contract (§E5) does not.
-//
-// Per proposal §14, where bd's behavior deviates from a clause the clause wins
-// and the deviation is a bug — so this test asserts the clause and stays
-// skipped until wy-kxgf4 lands, at which point it becomes the guardrail.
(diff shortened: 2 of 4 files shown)


harry-miller-trimble:
## Automated agent review — MERGE-WORTHY (needs a trivial rebase)

> This is an automated agent review, not a maintainer approval, and does **not** satisfy the human-review requirement in PR_MAINTAINER_GUIDELINES §77. Posted to inform review.

Reviewed read-only; the branch is ~96 commits behind `main` (CONFLICTING), so this assesses intent against current main.

### The bug still exists on current main
Both claim-failure paths print plain text to stderr with no `jsonOutput` guard, so `bd update <id> --claim --json` emits an unparseable line instead of a JSON object (violates Protocol v0 §E5):
- `cmd/bd/update.go:380` — `fmt.Fprintf(os.Stderr, "Error claiming %s: %v\n", id, err)` in the direct-store `ClaimIssue` loop (matches the PR's target context).
- `cmd/bd/update_proxied_server.go:124` — same `Fprintf`, now guarded by `if claimLost` after the UOW transaction returns.

Contrast the not-found path (`show.go` → `HandleErrorRespectJSON` → `jsonStdoutError`) which already emits `{error, schema_version:1}` — the inconsistency this PR closes.

### Fix is well-formed
`reportClaimFailure` routes `--json` through `jsonStderrError`/`buildJSONError` (same helper as the not-found path), emitting `{error, schema_version:1}`; plain-text wording unchanged so the §E4 message-class tests still hold. The stdout→stderr choice (vs the not-found path) is deliberate to keep a mixed batch's stdout a clean issue array (OQ-5).

### Test is meaningful
Un-skips `TestProtocol_ErrorClass_ClaimFailures_StructuredJSON` (`cmd/bd/protocol/errors_contract_test.go`), asserting a parseable JSON error object with non-empty `error` + `schema_version==1` for both already-claimed and not-claimable cases; the harness handles stderr placement and both envelope forms.

### Conflict is mechanical (one line)
The `update_proxied_server.go` conflict is from the `uow.RunTxResult` refactor that relocated the claim-failure `Fprintf` to line 124 (guarded by `if claimLost`), semantics unchanged. Resolution: replace that `fmt.Fprintf(os.Stderr, "Error claiming %s: %v\n", id, err)` with `reportClaimFailure(id, err)`. The `update.go`, `errors.go`, and test hunks apply as-is.

Since this is your own PR, a quick rebase + the one-line swap should make it green.

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication verified the landed reporter). Verdict: **needs changes** — the claim-failure detail (holder / state) is worth keeping, but the delivery mechanism has been superseded.

**Blocker — main centralized failure reporting after this branched.** Verified: `reportUpdateFailures` now aggregates all per-ID failures into one final envelope-aware JSON line on stderr (with a regression pinning that single-document shape), while this PR emits an eager JSON object per claim failure at both call sites — merge-tree conflicts at both. Keeping the eager helper alongside main's reporter emits multiple JSON documents per command; dropping it fails this PR's own E4/E5 assertions (which expect holder/state in the top-level `error`, where main now puts only "N of M failed" with details under `failed[].error`). The right integration: carry claim class/holder/state into `reportUpdateFailures`' `failed[]` entries and re-pin E4/E5 against that shape; the proxied claim branch has also moved into the retry-aware `applyUpdateProxiedAttempt`.

**Should-fix — JSON-mode detection is inconsistent on the proxied path:** the helper honors the global `jsonOutput` (set by `--format json` and configured `json: true`) while the proxied success path checks only the literal `--json` flag — a mixed batch can emit JSON failures with human successes. Main's `jsonOut := jsonOutput` seam is the compatible fix; the new contract test covers only literal `--json` on a direct workspace.

The storage-side plumbing (holder/state preserved through `ErrAlreadyClaimed`/`ErrNotClaimable`) is real and worth landing inside the new shape.

PR #4757, opened 2026-07-13
feat(labels): exclusive label namespaces, flag-gated (bd-7u5ki)

steveyegge:
## Summary

Implements **bd-7u5ki**: exclusive label namespaces, flag-gated.

Labels are free-form strings, so a bead can carry both `tier:fable` and `tier:opus` — and the two downstream consumers disagree about which one counts: the ready-front listing matches the label you query for (bead looks routable) while the claim filter excludes on the other one (bead is unclaimable). A second label in a routing namespace silently narrows eligibility to zero. Motivating incident: two dep-chained wyvern constellations stalled invisibly on 2026-07-13.

A workspace can now opt in:

```bash
bd config set labels.exclusive-prefixes "tier:,review:"
```

after which an issue may carry **at most one label per configured prefix**. Off by default — an empty/unset key is byte-for-byte today's behavior.

## Design decisions (per the spec's open questions)

- **Reject vs replace:** reject by default, with `bd label add --replace` to swap in one step (the spec's suggested middle ground). The error names the existing label.
- **Enforcement point:** the two storage choke points, so every mutation path is covered without per-command duplication:
  - `issueops.AddLabelInTx` — `bd label add`, `bd update --add-label`, `bd label propagate`
  - `issueops.PersistLabels` (via `BatchContext`) — `bd create -l`, `create --file/--graph`, import, repo-sync, doctor-fix migrate
- **Import replays history:** import and other replay paths set `ExclusiveLabelConflictWarn` — violations are reported per conflict on stderr and the labels are **kept** (no silent data loss). `bd doctor` is the cleanup tool.
- **Adoption lint:** new `bd doctor` check *Exclusive Label Namespaces* (warn-only, Data category) lists non-closed violators so adopters can clean up before flipping enforcement on.

## Ergonomics

- `bd update` now processes `--remove-label` before `--add-label`, so a one-command swap (`--remove-label tier:x --add-label tier:y`) works under enforcement.
- `bd create --parent`: an explicit label evicts the inherited parent label in the same exclusive namespace instead of failing the create.
- `bd config set` validates the key; `provides:` is rejected (reserved and inherently multi-valued). `tier` and `tier:` normalize to the same prefix.

## Testing

- Pure unit tests: `internal/labelns` (parse/validate/match/conflicts).
- Integration tests (Dolt test server): rejection on add, idempotent re-add, unconfigured workspaces untouched, create-path rejection, one-command update swap, `--replace` swap, import warn-and-keep with callback, doctor violation scan (closed issues skipped).
- Verified end-to-end against a real workspace: config validation, create/label-add rejection, `--replace`, update swap, inheritance override, import warning, and the doctor check (85-check run, reports OK when unconfigured).
- `golangci-lint run` clean on touched packages; docs gates pass (`test/docsync`, CLI-docs drift, freshness).
- Pre-existing failures `TestImportFromLocalJSONL_LegacyFormats` (cmd/bd) and `TestValidation*` (cmd/bd/doctor) reproduce on clean `main` and are unrelated.

## Docs

`docs/reference/configuration.md` gains an *Exclusive Label Namespaces* section; CLI reference regenerated from the updated `config`/`label` help text.

### DIFF
--- a/cmd/bd/config.go
+++ b/cmd/bd/config.go
@@ -14,6 +14,7 @@ import (
 	"github.com/steveyegge/beads/internal/beads"
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/git"
+	"github.com/steveyegge/beads/internal/labelns"
 	"github.com/steveyegge/beads/internal/metrics"
 	"github.com/steveyegge/beads/internal/remotecache"
 	"github.com/steveyegge/beads/internal/tracker"
@@ -72,6 +73,18 @@ Custom Status States:
   This enables issues to use statuses like 'awaiting_review' in addition to
   the built-in statuses (open, in_progress, blocked, deferred, closed).
 
+Exclusive Label Namespaces:
+  Labels are free-form, but routing conventions like tier:<model> assume one
+  label per prefix; a second one silently narrows fleet eligibility to zero.
+  Declare prefixes exclusive (at most one label per issue) with:
+
+    bd config set labels.exclusive-prefixes "tier:,review:"
+
+  Adding a second label in an exclusive namespace is then rejected on every
+  write path ('bd label add --replace' swaps instead). Import warns and keeps
+  violating labels; 'bd doctor' reports existing violations. Unset the key to
+  restore fully free-form labels (the default).
+
 Suppressing Doctor Warnings:
   Suppress specific bd doctor warnings by check name slug:
     bd config set doctor.suppress.pending-migrations true
@@ -88,6 +101,7 @@ Examples:
   bd config set jira.url "https://company.atlassian.net"
   bd config set jira.project "PROJ"
   bd config set status.custom "awaiting_review,awaiting_testing"
+  bd config set labels.exclusive-prefixes "tier:,review:"
   bd config set doctor.suppress.pending-migrations true
   bd config set dolt.debug true                        # Enable Dolt sql-server debug mode (loglevel=debug, --prof cpu)
   bd config set dolt.local-only true                   # Skip wiring a Dolt sync remote during bd init
@@ -207,6 +221,12 @@ var configSetCmd = &cobra.Command{
 		if key == "status.custom" && value != "" {
 			if _, err := types.ParseCustomStatusConfig(value); err != nil {
 				return HandleError("invalid status.custom value: %v", err)
+			}
+		}
+
+		if key == labelns.ConfigKey && value != "" {
+			if _, err := labelns.ValidatePrefixes(value); err != nil {
+				return HandleError("invalid %s value: %v", labelns.ConfigKey, err)
 			}
 		}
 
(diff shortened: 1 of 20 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the exclusive-namespace feature is valuable for fleets, and precisely because of that, its invariant needs to hold under the two conditions fleets actually create: concurrency and the proxied path.

**Blocker 1 — the exclusivity check is racy under concurrent writers.** The check-then-insert in `issueops` has no structural backing: the labels schema is just `PRIMARY KEY (issue_id, label)`, so `tier:a` and `tier:b` are distinct rows, and the pinned Dolt engine snapshots at tx start and three-way-merges at commit — two writers each see no conflict and both merge cleanly, recreating exactly the invalid state the flag promises to prevent. Needs a shared CAS/lock target (e.g. a per-(issue, namespace) row the way claims now use `row_lock`) or structural uniqueness, plus a two-writer regression.

**Blocker 2 — proxied/UOW writes bypass the invariant entirely.** Verified: the domain label repository does a direct `INSERT IGNORE` and never routes through `AddLabelInTx`/`labelns`; proxied update adds before removing with no exclusivity check; and post-branch main now dispatches `bd label` mutations to a proxied handler that neither enforces the invariant nor implements `--replace`. (Same route-parity gap we've flagged across this sweep — this PR is the strongest case for pushing the invariant down into the shared storage layer rather than the CLI.)

**Should-fixes:** proxied `config set` and `config set-many` skip `ValidatePrefixes` (malformed values like `:` or reserved `provides:` get stored, then silently dropped by the parser — quietly disabling enforcement); repo-sync and doctor-migrate enable conflict-warn mode without wiring the callback, so violations are preserved silently; `doctor` passes `UnderlyingDB()` straight into issueops (storage-boundary rule); and the 895-line span violates the repo's own stacking rule (storage/issueops vs cmd/bd as separate PRs) — splitting would also make the rebase tractable, since merge-tree already reports five conflict hunks including a semantic one against #4729's import chunking.

marcodelpin:
We've been running this branch and carried it forward on our fork: merged current main, addressed review blocker 1 (per-(issue,namespace) lock row serializing exclusive-namespace writers, on both routes, with a two-writer regression test), and added hardening on top (lock-row cascade on delete, bounded retention, a `bd doctor` check for pre-cascade orphan lock rows). Branch: `marcodelpin/beads:pr/4757-blockers-fix`. Happy to open a PR against your branch, or split out whatever subset is useful — whichever shape works best for you.

One design point we'd value your read on: policy ENABLE has a disclosed race window (a writer whose snapshot predates the enable can commit a violation; the doctor check detects it after the fact) — we chose disclosure + doctor over a global rendezvous to keep the per-namespace lock granularity.

bee-ghosttrack:
CI triage (bee, maintainer seat): the pure-Go compile check fails inside the PR's new test — `cmd/bd/label_exclusive_test.go:20` and `:71` reference `newTestStoreWithPrefix`, which does not exist in the tree. Branch is CONFLICTING with main as well.

PR #4722, opened 2026-07-11
docs: Mintlify overhaul — docs/ becomes the doc site, all-in

csells:
# docs: Mintlify overhaul — docs/ becomes the doc site

Replaces the Docusaurus/GitHub Pages site with a Mintlify site rooted at
`docs/`. Decisions behind the design are recorded in
[`engdocs/decisions/2026-07-10-mintlify-docs-overhaul.md`](https://github.com/gastownhall/beads/blob/docs/mintlify-overhaul/engdocs/decisions/2026-07-10-mintlify-docs-overhaul.md).

## What's in this PR

- **`website/` is deleted** (Docusaurus app, its 4 version snapshots,
  `deploy-docs.yml`, the ci-website gate, release-time snapshotting). `docs/`
  itself is now the site: `docs/docs.json` defines nav and 46 redirects (42 page renames + 4 wildcards for the old /next/ and version-prefixed route spaces)
  covering the old top-level page names. Versioning is current-only.
- **CLI reference is generated, vendor-neutrally.** `bd help --docs-root`
  emits generic Markdown (committed `docs/CLI_REFERENCE.md` + a staging tree);
  the new `tools/docsmint` (Go, unit-tested) turns staging into the 113
  committed `docs/cli-reference/` pages and splices the docs.json nav. No
  site-generator specifics live in the bd binary. The existing drift model
  (`generate-cli-docs.sh --check`, `check-cli-docs-drift.sh`, autofix bot) is
  preserved over the new artifacts.
- **`docs/` ⇄ `engdocs/` split**: docs/ is user-facing only; ~35
  contributor/maintainer files (adr/, design/, staged-for-removal/ — nothing
  deleted) moved to engdocs/ with inbound links fixed.
- **Content overhaul**: 48 authored pages ported to Mintlify components; the
  two divergent FAQs merged into one; new canonical "How Beads Works" concept
  page; section overviews; README trimmed 254 → 188 lines; 9 mermaid diagrams;
  Excalidraw pipeline ported.
- **Fact corrections throughout** — the port surfaced heavy drift: phantom
  commands (`bd pour`, `bd pin`, `bd hook`, `bd stats`, …), phantom config
  keys, wrong data paths, a routing page describing a system that doesn't
  exist. Every factual claim in the new corpus was verified against the binary
  or source (620 claims checked, 45 corrected).
- **bd string fixes**: printed doc paths/URLs (prime help, init-safety,
  recovery URL, doctor hints, recipe/agent templates) now point at the new
[...]

PR #4759, opened 2026-07-13
docs: redirect legacy GitHub Pages site to Mintlify

maphew:
## What changed

- replace the frozen Docusaurus GitHub Pages deployment with a two-file redirect artifact (`index.html` and `404.html`) built from one committed source
- preserve bookmarked deep-link paths, query strings, and fragments while moving from `gastownhall.github.io/beads` to `beads.gascity.com`
- cover the retired `1.1.0` version route and repoint remaining live repository links to the Mintlify site

## Why

The Mintlify cutover in #4722 intentionally stopped publishing Docusaurus, but the last Pages artifact is still live and silently serves stale documentation. Disabling Pages would break saved links. A redirect-only artifact keeps those bookmarks useful and sends readers to the current docs.

The deployment remains separate from Mintlify: after this PR merges, the `main` push publishes only the redirect stubs to GitHub Pages. The new documentation continues to be hosted at `https://beads.gascity.com/`.

## Validation

- `go test ./test/docsync`
- `go test -tags=gms_pure_go ./test/docsync`
- dependency-free Node VM checks for `/beads`, `/beads/`, deep paths, query strings, fragments, and exact-prefix handling
- generated artifact check confirms identical `index.html` and `404.html`
- all replacement target URLs return HTTP 200
- independent standards and spec reviews: no findings
- `git diff --check`

`actionlint` was not installed locally; the workflow reuses the exact immutable Pages action pins and deployment pattern from the retired workflow.

### DIFF
--- a/.github/pages-redirect/index.html
+++ b/.github/pages-redirect/index.html
@@ -1,0 +1,42 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="robots" content="noindex" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Beads documentation has moved</title>
+  </head>
+  <body>
+    <p>
+      The Beads documentation has moved to
+      <a id="destination" href="https://beads.gascity.com/">beads.gascity.com</a>.
+    </p>
+    <noscript>
+      <p>
+        JavaScript is required to preserve this page's path. Continue to the
+        <a href="https://beads.gascity.com/">current Beads documentation</a>.
+      </p>
+    </noscript>
+    <script>
+      (function () {
+        var path = window.location.pathname;
+        var destinationPath = path;
+
+        if (path === "/beads" || path === "/beads/") {
+          destinationPath = "/";
+        } else if (path.indexOf("/beads/") === 0) {
+          destinationPath = path.slice("/beads".length);
+        }
+
+        var destination =
+          "https://beads.gascity.com" +
+          destinationPath +
+          window.location.search +
+          window.location.hash;
+
+        document.getElementById("destination").href = destination;
+        window.location.replace(destination);
+      })();
+    </script>
+  </body>
+</html>
--- a/.github/pages-redirect/index.html
+++ b/.github/pages-redirect/index.html
@@ -4,6 +4,8 @@
     <meta charset="utf-8" />
     <meta name="robots" content="noindex" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="redirect-destination" content="https://beads.gascity.com" />
+    <meta http-equiv="refresh" content="1; url=https://beads.gascity.com/" />
     <title>Beads documentation has moved</title>
   </head>
   <body>
@@ -29,7 +31,7 @@
         }
 
         var destination =
-          "https://beads.gascity.com" +
+          document.querySelector('meta[name="redirect-destination"]').content +
           destinationPath +
           window.location.search +
           window.location.hash;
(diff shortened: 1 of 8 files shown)


maphew:
## Review: 8-angle finder sweep + verification

Ran an 8-angle multi-agent review (line-by-line, removed-behavior, cross-file trace, reuse, simplification, efficiency, altitude, conventions), then verified surviving candidates with greps against `main` and live HTTP probes of both the old and new docs sites. The design is sound overall — the `/1.1.0/:slug*` addition completes the frozen site's version set (`/next`, `/1.0.0`, `/1.0.4`, `/1.0.5` are already on `main`). Six findings survived, ordered by severity.

### Act-before-merge

**1. `.github/pages-redirect/index.html:26` — redirect is JavaScript-only; deep legacy links will be served with HTTP status 404 and no fallback.**
Every deep path goes through the copied `404.html`, which GitHub Pages serves with status 404 — no `Location` header, no `<meta http-equiv="refresh">`. Verified live: `https://gastownhall.github.io/beads/llms-full.txt` and deep doc pages return 200 today; after this deploys, curl, link checkers, crawlers, and AI agents fetching them get a 404 HTML page and treat the docs as dead, even though `https://beads.gascity.com/llms-full.txt` serves 200. The decision record (`engdocs/decisions/2026-07-10-mintlify-docs-overhaul.md:127`) specified "meta-refresh redirect stubs." The 404 status itself is a Pages limitation, but adding a meta-refresh (even root-only) alongside the JS path-preserver covers no-JS clients and most crawlers. Relatedly, the `<noscript>` link drops the path entirely (points at the root) while the page text claims JavaScript is required to preserve it.

**2. `npm-package/bin/README.md:13` — missed README copy ships stale links in the npm tarball.**
Lines 13 and 182 still link `https://gastownhall.github.io/beads/`, including the now-false "(versioned)" claim this PR removed from the top-level README. `npm-package/package.json` `files` includes `bin/`, so this copy ships inside published `@beads/bd` platform packages, and its links will only work via the JS redirect this same PR deploys.

### Optional polish

**3. `.github/workflows/deploy-pages-redirect.yml:40`** — the `github.repository` guard sits only on `deploy`, so every fork that syncs `main` runs the full `build` job (checkout + copy + artifact upload) whose output is guaranteed to be discarded. Moving the guard to `build` lets `deploy` skip via `needs`.

**4. `.github/workflows/deploy-pages-redirect.yml:19`** — `cancel-in-progress: true` on the `pages` group can abort a deployment mid-publish; two quick pushes and a transient failure on the second leave the site stale until a manual re-run. GitHub's Pages starter workflow uses `cancel-in-progress: false` for this reason.

**5. `.github/workflows/deploy-pages-redirect.yml:22`** — a two-job split with a full repo checkout is heavy for copying one 42-line committed file: two runner boots plus an artifact handoff per deploy. A single job with `sparse-checkout: .github/pages-redirect` then upload + deploy is equivalent and half the YAML.

**6. `.github/pages-redirect/index.html:18`** — `https://beads.gascity.com` is spelled out four times in one file (visible anchor, noscript anchor, prose, JS string). The next domain move needs four coordinated edits; missing the noscript one leaves no-JS users on a stale domain — invisible to any tester with JS enabled. Hoist into one constant and derive or co-locate the rest.

### Refuted along the way

- Missing `/1.0.4`, `/1.0.0`, `/next` Mintlify redirects — all already present on `main`.
- `/beads/search` breaking — already a 404 on the frozen site today.
- Capital-case `/Beads/` paths regressing — already 404 today.
- CLAUDE.md/AGENTS.md convention violations — none; docsync tests pass on these edits.

PR #4761, opened 2026-07-13
ci: shard embedded Dolt conformance gate

maphew:
## What

- Split embedded-Dolt conformance into `core` and `audit` matrix partitions.
- Keep the existing 30-minute Go timeout per partition.
- Add a 35-minute Actions timeout so a real Go timeout can still emit diagnostics before the runner stops.
- Apply the same partitioning to `main.yml` and `pr-risk.yml` so PR coverage matches main.

## Why

[Main run 29278903172](https://github.com/gastownhall/beads/actions/runs/29278903172) timed out `TestConformance` at exactly 30 minutes. The suite was not hung: it was still entering new subtests, with only the tail of `Audit` plus three short top-level tests remaining.

Recent green runs took 21-25 minutes, while the failed runner was uniformly about 1.4x slower. Raising the timeout again would repeat the recent 15-to-30-minute ratchet. `RunAll` already has a natural `Audit` boundary, and splitting there gives each partition about 2x headroom while roughly halving wall-clock time.

The earlier red lint wrapper was a transient Go proxy HTTP/2 download error and passed on the replacement run. The conformance timeout is the remaining reproducible main failure.

## Verification

- `actionlint` v1.7.12 passes both changed workflows.
- A synthetic nested Go test confirms the `-test.run`/`-test.skip` filters assign all non-Audit tests to `core` and the full Audit subtree to `audit`.
- `scripts/check-build-tags.sh` passes.
- `git diff --check` passes.
- Independent review confirmed the PR-risk aggregate gate consumes the matrix job's combined result.

Coordination: `mybd-ie8n`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -844,10 +844,18 @@ jobs:
         run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=30m -test.skip '^TestConformance$'
 
   test-embedded-conformance:
-    name: Test (Embedded Dolt Conformance)
+    name: Test (Embedded Dolt Conformance - ${{ matrix.partition }})
     needs: [detect-ci-tier, build-embedded]
     if: needs.detect-ci-tier.outputs.full_embedded == 'true'
     runs-on: ubuntu-latest
+    timeout-minutes: 35
+    strategy:
+      fail-fast: false
+      matrix:
+        partition: [core, audit]
+    env:
+      BEADS_TEST_EMBEDDED_DOLT: "1"
+      BEADS_TEST_EMBEDDED_TEST_BINARY: /tmp/embeddeddolt-test
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v6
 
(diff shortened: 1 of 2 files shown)


maphew (commented):

The PR shards the embedded-Dolt conformance gate into core/audit matrix partitions in both main.yml and pr-risk.yml. I verified against source that RunAll has exactly one top-level Audit seam and that the -test.run/-test.skip filters produce complete, non-overlapping coverage with no stranded subtests. The pr-risk ci-gate correctly aggregates the matrix job's combined result and its required check name is unchanged, so PR merges are unaffected. The change is correct, idiomatic (mirrors the existing test-embedded-cmd matrix), and I found no blocker or major issues.

**Findings:**
- **minor** `.github/workflows/pr-risk.yml:218` — pr-risk's per-shard Go timeout drops from 45m (single combined run) to 30m per shard, with a 35m Actions job timeout. Correctness of the split is fine, but the ~15m-per-half estimate in the comment is unverified; if the core shard (all non-Audit subtests) runs long under runner variance it can hit 30m and fail. Given the 5m job-timeout buffer this is thin if the estimate is off. Worth monitoring the first few runs; not a defect in the diff itself.
- **nit** `.github/workflows/main.yml:847` — main.yml renames the status check from 'Test (Embedded Dolt Conformance)' to '... - core'/'... - audit' and, unlike pr-risk, has no aggregating gate job. pr-risk is safe (CI Gate / Required is unchanged and aggregates via needs). Confirm no branch-protection rule requires the old exact job name directly, otherwise that required check would sit pending forever. Low risk since main.yml is the push workflow, not the PR gate.

PR #4762, opened 2026-07-13
test(protocol): pin §E3 exit 11 with a PTY harness (wy-vh5y8)

steveyegge:
## What

Un-skips `TestProtocol_ExitCode11_LocalExistsRefused` in `cmd/bd/protocol/exit_codes_init_test.go`, the last unpinned §E3 exit code from the WP4 conformance corpus (wy-ge2h8).

## Why it was skipped

Exit 11 (`ExitLocalExistsRefused`) is returned **only** from the interactive typed-confirmation abort in `bd init --reinit-local` (`cmd/bd/init.go`, behind `term.IsTerminal(os.Stdin.Fd())`). The non-interactive branch of the same guard returns 12, so no piped-stdin subprocess could ever produce 11.

## How

Allocate a PTY (`creack/pty` — a **test-only** dependency, already in `go.sum`), drive `bd init --reinit-local` against a workspace seeded with one issue, answer the `Type 'destroy N issues' to confirm:` prompt with the **wrong** text, and assert exit 11 — distinct from the 12 a bad `--destroy-token` yields. Also asserts the abort is a no-op (the seeded issue survives). Windows skips (no unix PTY).

## Acceptance

- ✅ The formerly-skipped test runs and passes (`go test -run TestProtocol_ExitCode11_LocalExistsRefused`, ~40s incl. Dolt container).
- ✅ No new **non-test** dependency in `go.mod`: `creack/pty` is imported only from `_test.go`.

### DIFF
--- a/cmd/bd/protocol/exit_codes_init_test.go
+++ b/cmd/bd/protocol/exit_codes_init_test.go
@@ -31,6 +31,8 @@ import (
 	"sync"
 	"testing"
 	"time"
+
+	"github.com/creack/pty"
 )
 
 // Literal exit codes frozen by §E3. Deliberately NOT bd's Go constants: the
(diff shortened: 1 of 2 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication verified the one blocker). Verdict: **needs changes — one-hunk rebase.** The PTY harness itself checks out end-to-end (init skips root store init; `countExistingIssues` is the store-opening step; `pty.Start` provides the controlling TTY; wrong confirmation maps to literal exit 11; seeded-issue survival asserted), and current main retains the same exit-11 behavior, so no semantic conflict.

**Blocker — `go.mod` conflict:** the branch adds `creack/pty v1.1.24` beside the stale `dolthub/driver/v2 v2.1.10`; main is now on driver v2.2.0 (verified) and Go 1.26.5. Rebase keeping pty and main's driver/Go lines.

Nit: the prompt-driving mutex/reader/deadline machinery is duplicated between the two tests — a small helper would halve it.

bee-ghosttrack:
CI triage (bee, maintainer seat): the red is `nix build .#default` / Test Nix Flake; the run's logs have expired, but the diff adds a go.mod dependency (creack/pty) and the `Update default.nix vendorHash` job was skipped, so the flake's pinned vendorHash no longer matches. Regenerate it alongside the go.mod change. Branch is CONFLICTING with main as well.

PR #4764, opened 2026-07-13
perf(schema): skip migration lock on already-migrated opens

outdoorsea:
## Summary

`schema.MigrateUpWithLock` acquires the per-database schema-migration advisory lock (`SELECT GET_LOCK`) on **every** server-mode database open (via `dolt.initSchemaOnDB`). Because `bd` opens a fresh connection per operation and the lock is a single named lock per database, **every concurrent open on a shared Dolt sql-server serializes through that one lock** — even when the schema is already at the latest version and no migration will run.

Under load on a busy shared database the lock queue deepens until acquisition exceeds the 5s `GET_LOCK` timeout and opens fail with:

```
schema: acquire migration lock: schema migration lock unavailable: timeout
```

A hot-but-healthy server degrades into one that drops writes and thrashes on retries.

## Where

- `internal/storage/schema/lock.go` — `MigrateUpWithLock` unconditionally `AcquireMigrationLock` → `MigrateUp` → `ReleaseMigrationLock`.
- Caller: `internal/storage/dolt/store.go` `initSchemaOnDB`, run on every server-mode open.

The lock is only actually needed to serialize a migration pass (or the `dolt_ignore` self-heal commit). An open on an already-migrated, already-seeded database runs `MigrateUp` as a pure no-op, so it never needed the lock.

## Fix

Read-only double-checked locking. `MigrateUpWithLock` now probes `migrationWorkNeeded` and a new `doltIgnorePatternsSeeded` — **both plain reads** — and returns immediately without touching `GET_LOCK` when the open is provably a no-op. Any uncertainty (a pending migration, or an under-seeded out-of-band copy that still needs the `dolt_ignore` heal) falls through to the unchanged locked `MigrateUp`, so the #4566 self-heal contract and concurrent-migration serialization are preserved exactly.

Single layer (`internal/storage/schema`), +158 lines.

## Proof (beads-only, no orchestrator)

- `TestMigrateUpWithLockSkipsLockWhenAtLatestAndSeeded` — an at-latest, seeded open issues **no `GET_LOCK`** (sqlmock fails on any unexpected query, so an errant lock acquisition fails the test).
- `TestMigrateUpWithLockTakesLockWhenUnderSeeded` — an at-latest but under-seeded copy **still takes the lock** and commits the heal under it (the zero-regression / #4566 guard).
- `TestMigrateUpWithLockUsesDatabaseScopedLockOnly` — the unchanged pending-migration path still locks.

`go test ./internal/storage/schema/`, `go vet ./internal/storage/schema/`, and `gofmt -l` are all clean.

## Real-world corroboration

Surfaced in a high-concurrency deployment (~two dozen processes plus tight-cooldown health probes all opening one shared coordination database): `SHOW FULL PROCESSLIST` showed multiple connections parked in `SELECT GET_LOCK(...)` at once; the server log flooded with `Error reading packet from client ... i/o timeout`; trivial queries oscillated between ~3s and >40s (timeout). Restarting the server rebuilt the same contention within minutes — because the lock is re-acquired on every reconnect, not a wedged-process issue. The mechanism and the fix are entirely within beads; the tests above need no orchestrator.

## Out of scope

The same deployment also shows high connection **churn** (a fresh connect/close per `bd` invocation, no pooling). That is process/orchestrator-level, not a `schema`-layer concern, so it is intentionally excluded here (one layer per PR). This change targets the serialization that made opens *fail*; churn is a separate follow-up.

I couldn't add a wall-clock benchmark that is meaningful under `sqlmock` (no lock latency to measure); happy to add a real-server concurrent-open benchmark if you'd like one for the record.

### DIFF
--- a/internal/storage/schema/lock.go
+++ b/internal/storage/schema/lock.go
@@ -53,7 +53,17 @@ func MigrationLockName(databaseName string) string {
 // server session. Embedded Dolt intentionally uses bare MigrateUp because it
 // relies on the embedded driver's file/concurrency controls instead of
 // sql-server session locks.
+//
+// A read-only fast path skips the lock entirely when the database is provably
+// at-latest and seeded, so the common open no longer serializes every concurrent
+// connection on the per-database GET_LOCK.
 func MigrateUpWithLock(ctx context.Context, conn *sql.Conn, databaseName string) (applied int, err error) {
+	if noop, err := migrateUpWouldNoop(ctx, conn); err != nil {
+		return 0, err
+	} else if noop {
+		return 0, nil
+	}
+
 	lockName := MigrationLockName(databaseName)
 	if err := AcquireMigrationLock(ctx, conn, lockName); err != nil {
 		return 0, err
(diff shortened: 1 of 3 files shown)


maphew:
Thanks for identifying this hot path — skipping the migration lock when the schema is already current is a real win, and your draft got the shape right (unlocked probe of migration + seed state, locked path otherwise).

The same optimization has since landed in review-ready form in #4804, which covers both of your probe states and adds the piece review surfaced as necessary: a pass-completion sentinel (deleted before the first migration mutation, stamped only after the full migrate/rekey/commit tail), so the fast path can't engage in the window after migration cursors are current but before the dependency rekey finishes. One deliberate difference: where this draft fails the open on a probe error, #4804 falls through to the authoritative locked path instead, which is the safer degradation.

So rather than ask you to rebase the draft, I'm proposing to retire it in favor of #4804 and will close it in a few days if there's no objection. If you see a case your version handles that #4804 does not, say the word and we'll reopen the comparison.

maphew:
Retiring in favor of #4804, which covers both unlocked-probe states and adds the pass-completion sentinel; see disposition comment 2026-07-28. Thanks for mapping the fast path first.

PR #4773, opened 2026-07-14
docs: Fix flowchart references in README.md

chr1sto14:
## What

Fix the "Unable to render rich display" error in the mermaid rendering because `graph` is a reserved word.

## Why

image

## Verification

Rendered the README.md confirming the fix.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,12 +16,12 @@ Beads provides a persistent, structured memory for coding agents. It replaces me
 
 ```mermaid
 flowchart LR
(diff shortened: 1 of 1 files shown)


maphew:
Overtaken: the README flowchart node-id fix landed via #4994 (renamed to depgraph), and this PR touches only README.md. The remaining broken diagram in docs/core-concepts/index.md is being fixed by #4958. Thanks for spotting and fixing this — you had it right before the fix that landed. _claude-fable-5-high on behalf of matt wilkie_

PR #4374, opened 2026-06-12
fix: consolidate 8 critical-bug-sweep branches (5 fixes, 4 duplicate)

kevglynn:
## Summary

The daily "critical bug sweep" automation repeatedly discovered and fixed the same bugs across 8 separate `cursor/critical-bug-investigation-*` branches without landing any of them, spawning duplicates each run.

This PR consolidates all 8 branches:

### Duplicate fixes (4 branches → 1 cherry-pick)
- **Linear OAuth authHeader()** — `createIssueSingleAttempt` used `c.APIKey` directly instead of `c.authHeader()`, so OAuth client-credentials setups sent an empty Authorization header. Picked branch `672f` (most comprehensive regression test using `NewOAuthClient` with a real token server). Deleted branches `30a4`, `92ee`, `df0e`.

### Distinct fixes (4 branches → 4 cherry-picks)
- **Rate-limit circuit-breaker false-positive** (`5e8b`) — The breaker tripped on the current request that revealed low remaining quota, discarding a successful response. Now records state and trips on the *next* call instead. Adds persistent `rateLimitState` struct with mutex.
- **BatchPush double-format skip mismatch** (`7e25`) — Pre-formatted descriptions from `FormatDescription` hook caused `PushFieldsEqual` to see a mismatch (structured fields still populated on local issue). Clears `AcceptanceCriteria`/`Design`/`Notes` before comparison.
- **bd doctor conflicting project_id repoint** (`1a3d`) — When the configured database has its own `project_id` that disagrees with `metadata.json`, and a different database matches the local `project_id`, doctor now errors instead of silently switching databases.
- **Shared-server sole-candidate auto-adopt** (`cef0`) — Gates sole-candidate database adoption on `!doltserver.IsSharedServerMode()` to prevent cross-project data attachment when no `project_id` anchor exists.

All 8 branches have been deleted from the fork remote.

## Test plan

- [x] `go test ./internal/linear/... -count=1` — all pass (OAuth, rate-limit breaker, BatchPush)
- [x] `go test ./cmd/bd/doctor/fix/... -count=1 -run TestReconcile` — all 8 reconciliation tests pass
- [x] Each fix includes its own regression test
- [x] No conflicts between cherry-picks

Made with [Cursor](https://cursor.com)

PR #4778, opened 2026-07-14
fix(linear): arm rate-limit breaker after successful calls

kevglynn:
## Summary

The rate-limit circuit-breaker tripped on the *current* request that revealed low remaining quota, discarding a successful response. Now records state and trips on the *next* call instead, via a persistent `rateLimitState` struct with mutex.

## Split context

Split out of #4374, which consolidated 8 duplicate `cursor/critical-bug-investigation-*` branches into one PR. Splitting into independent, single-concern PRs per review feedback — see [PR_MAINTAINER_GUIDELINES.md](https://github.com/gastownhall/beads/blob/main/PR_MAINTAINER_GUIDELINES.md).

## Test plan

- `go test ./internal/linear/... -run TestRateLimit -count=1`

### DIFF
--- a/internal/linear/client.go
+++ b/internal/linear/client.go
@@ -119,6 +119,11 @@ type rateLimitState struct {
 	resetsAt  time.Time
 }
 
+// maxRateLimitResetHorizon accommodates Linear's short quota windows and
+// clock skew without allowing a nonsensical timestamp to block requests
+// effectively forever.
+const maxRateLimitResetHorizon = 24 * time.Hour
+
 func newRateLimitState() *rateLimitState {
 	return &rateLimitState{remaining: -1}
 }
@@ -257,7 +262,8 @@ func (c *Client) recordRateLimitHeaders(info RateLimitInfo) {
 	}
 	c.rateLimitState.mu.Lock()
 	defer c.rateLimitState.mu.Unlock()
-	if info.RequestsReset.IsZero() || !time.Now().Before(info.RequestsReset) {
+	now := time.Now()
+	if info.RequestsReset.IsZero() || !now.Before(info.RequestsReset) || info.RequestsReset.After(now.Add(maxRateLimitResetHorizon)) {
 		c.rateLimitState.remaining = -1
 		c.rateLimitState.resetsAt = time.Time{}
 		return
--- a/internal/linear/client.go
+++ b/internal/linear/client.go
@@ -234,9 +234,14 @@ func (c *Client) circuitBreakerError() *ErrRateLimitExhausted {
 	if c.rateLimitState == nil {
 		return nil
 	}
-	c.rateLimitState.mu.RLock()
-	defer c.rateLimitState.mu.RUnlock()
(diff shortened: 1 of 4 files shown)


harry-miller-trimble:

On main the rate-limit circuit breaker runs *after* a successful HTTP response is fully read, inside the retry loop (`internal/linear/client.go`): it parses `X-RateLimit-Requests-Remaining` from the just-succeeded response and, if below the floor, returns `ErrRateLimitExhausted` — **discarding the good payload** the request already paid for.

### The fix
Converts the stateless per-call breaker into a persistent one (`rateLimitState` struct, mutex-guarded, shared by pointer across `With*` clones):
- `recordRateLimitHeaders` stores remaining/resetsAt after each response; the current successful response is now returned normally.
- `circuitBreakerError` is checked at the *top* of the execute loop, so the *next* call trips before hitting the network (matches the PR title).
- Auto-disarms once `resetsAt` passes, then re-arms on the next low reading.

Correct and idiomatic: nil-guarded (`rateLimitState == nil` → breaker disabled, no panic), RWMutex with write-lock on expiry, per-clone floor honored while sharing the state pointer. All six Client constructors (`NewClient`, `NewOAuthClient`, `WithEndpoint`, `WithHTTPClient`, `WithProjectID`, `WithRateLimitFloor`) propagate the shared state; none missed.

### Coupled header-parse change is necessary, not churn
It switches `X-RateLimit-Requests-Reset` parsing from `time.Parse(RFC3339, …)` to `time.UnixMilli(…)`. Per Linear's docs this header is epoch **milliseconds**, so the RFC3339 parse *always failed* on main (left `RequestsReset` zero — cosmetic before, but the new stateful breaker depends on `resetsAt` to disarm). The overflow/negative/beyond-horizon guards are sound.

### Tests — adequate
`CircuitBreakerTripsOnSubsequentRequest` (first call returns data, second trips locally with one request made), `CircuitBreakerAllowsRequestAfterResetThenRearms`, and a `CircuitBreakerRequiresValidFutureReset` table (missing/malformed/overflow/negative/max-int64/beyond-horizon/valid). `TestParseRateLimitHeaders` updated to millis; `parent_reconcile_test.go` updated to send the now-required reset header.

### Minor, non-blocking
The breaker no longer arms when the reset header is absent/unparseable (main armed on remaining alone). Deliberate tradeoff — without a reset time it can't know when to disarm, so it degrades to normal 429/backoff rather than blocking forever. Safe since Linear always sends the header.

kevglynn:
Pushed `b86c1090` fixing the `TestReconcileParents_AbortsOnRateLimit` failure from the last run: your `48b0c672` (require valid reset to arm breaker) changed the arming contract, but the parent-reconcile mock only sent `X-RateLimit-Requests-Remaining` with no reset header — so the breaker never armed and the abort never happened. The mock now sends a valid future `X-RateLimit-Requests-Reset` alongside the low remaining count, matching real Linear responses. Test-fixture-only change; no production code touched.

PR #4779, opened 2026-07-14
fix(linear): avoid double-format skip mismatch in batch push

kevglynn:
## Summary

Pre-formatted descriptions from the `FormatDescription` hook caused `PushFieldsEqual` to see a mismatch, since structured fields (`AcceptanceCriteria`/`Design`/`Notes`) were still populated on the local issue. Clears those fields before comparison so `BatchPush` doesn't skip issues it shouldn't.

Includes a follow-up commit adding a `TeamLabels` mock handler to the regression test, needed because the label-push feature already on `main` builds a per-team label cache during `BatchPush`.

## Split context

Split out of #4374, which consolidated 8 duplicate `cursor/critical-bug-investigation-*` branches into one PR. Splitting into independent, single-concern PRs per review feedback — see [PR_MAINTAINER_GUIDELINES.md](https://github.com/gastownhall/beads/blob/main/PR_MAINTAINER_GUIDELINES.md).

## Test plan

- `go test ./internal/linear/... -run TestBatchPush -count=1`

### DIFF
--- a/internal/linear/tracker.go
+++ b/internal/linear/tracker.go
@@ -503,7 +503,14 @@ func (t *Tracker) BatchPush(ctx context.Context, issues []*types.Issue, forceIDs
 			fetched, lookupErr := routeClient.FetchIssueByIdentifier(ctx, externalID)
 			if lookupErr == nil && fetched != nil {
 				remoteIssue = fetched
-				if PushFieldsEqual(issue, remoteIssue, t.config, teamLabelCache) {
+				// BatchPush receives pre-formatted descriptions from the sync
+				// engine (FormatDescription hook). Clear structured fields before
+				// comparison so PushFieldsEqual does not re-append them.
+				comparableIssue := *issue
+				comparableIssue.AcceptanceCriteria = ""
+				comparableIssue.Design = ""
+				comparableIssue.Notes = ""
+				if PushFieldsEqual(&comparableIssue, remoteIssue, t.config, teamLabelCache) {
 					result.Skipped = append(result.Skipped, issue.ID)
 					continue
 				}
--- a/internal/linear/tracker_test.go
+++ b/internal/linear/tracker_test.go
@@ -157,6 +157,79 @@ func TestBatchPush_SkipsUnchangedIssue(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


harry-miller-trimble:

On the Linear **batch push** path the engine pre-formats each issue's description but leaves the structured fields populated; the skip-comparison then re-appends them — a double-format that makes unchanged issues appear different and get needlessly re-pushed on every sync.

Chain on main:
1. Linear implements `BatchPushTracker`, so the engine routes through the batch path (`internal/tracker/engine.go`), calling `collectBatchPushIssues` → `formatPushIssue`, which sets `copy.Description = FormatDescription(issue)` (= `BuildLinearDescription`, merging AcceptanceCriteria/Design/Notes) **but does not clear the structured fields**.
2. Inside `BatchPush`, the skip check calls `PushFieldsEqual(...)` (`internal/linear/tracker.go`) which recomputes `BuildLinearDescription(local)` (`internal/linear/mapping.go`). Since `local.Description` is already merged AND the structured fields are still set, the criteria sections are appended a **second** time → mismatch vs remote → not skipped → a needless `UpdateIssue` API call every sync (write amplification).

This matches the pattern already handled correctly on the CreateIssue path and the update mapper; only the skip comparator re-formatted.

### Fix quality — correct, minimal, idiomatic
Shallow-copies the issue and clears the three structured fields before comparison, so `BuildLinearDescription` returns just the already-formatted Description. Original untouched; issues without structured fields are unaffected (clearing empty strings is a no-op); legitimate content changes still trigger updates. The existing `NormalizeIssueForLinearHash` helper can't be reused (it re-runs `BuildLinearDescription`, reintroducing the double-format), so the inline approach is right. The other `PushFieldsEqual` caller (`ContentEqual` hook, non-batch path) receives the un-formatted issue and is unaffected.

### Tests — adequate
`TestBatchPush_SkipsUnchangedIssueWithPreformattedDescription` reproduces the exact state `formatPushIssue` produces and asserts the issue is skipped and `issueUpdate` is never called. Without the fix, the double-append makes it mismatch → update fires → test fails. Genuinely gates the bug.

### Minor, non-blocking
The test simulates `formatPushIssue` output by hand rather than driving the engine end-to-end; reasonable at the `BatchPush` boundary, though an engine-level test would be marginally stronger.

PR #4780, opened 2026-07-14
fix(doctor): harden project_id/metadata reconciliation against unsafe auto-switch

kevglynn:
## Summary

Two related safety guards on `reconcileAuthoritativeServerMetadata`'s auto-switch logic, kept together as one PR since the second commit's diff is directly adjacent to the first's in the same function:

1. **Block unsafe cross-project auto-switch.** When the configured database has its own `project_id` that disagrees with `metadata.json`, and a different database matches the local `project_id`, doctor now errors instead of silently switching databases — the two signals conflict and auto-switching could silently attach the workspace to a different project.
2. **Avoid shared-server sole-candidate adoption.** Gates sole-candidate database adoption on `!doltserver.IsSharedServerMode()`, since a sole schema candidate on a shared server can belong to a different workspace, and there's no local `project_id` anchor to safely resolve the ambiguity.

## Split context

Split out of #4374, which consolidated 8 duplicate `cursor/critical-bug-investigation-*` branches into one PR. Splitting into independent, single-concern PRs per review feedback — see [PR_MAINTAINER_GUIDELINES.md](https://github.com/gastownhall/beads/blob/main/PR_MAINTAINER_GUIDELINES.md). These two commits are kept together (rather than split further) because they touch adjacent hunks of the same function in `cmd/bd/doctor/fix/metadata.go`.

## Test plan

- `go test ./cmd/bd/doctor/fix/... -run TestReconcile -count=1`

### DIFF
--- a/cmd/bd/doctor/fix/metadata.go
+++ b/cmd/bd/doctor/fix/metadata.go
@@ -344,6 +344,7 @@ func reconcileAuthoritativeServerMetadata(cfg *configfile.Config, databases []se
 			schemaCandidates = append(schemaCandidates, db)
 		}
 	}
+	current, hasCurrent := byName[cfg.GetDoltDatabase()]
 
 	if cfg.ProjectID != "" {
 		var matches []serverDatabaseMetadata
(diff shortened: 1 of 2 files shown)


harry-miller-trimble:

`reconcileAuthoritativeServerMetadata` (`cmd/bd/doctor/fix/metadata.go`) reconciles `metadata.json` (`cfg.ProjectID` + `cfg.DoltDatabase`) against the authoritative server DB inventory. Two silent-misattachment paths exist on main:

1. **Cross-project auto-switch.** When `cfg.ProjectID` matches exactly one *other* database, doctor blindly repoints `cfg.DoltDatabase` to it — even when the *currently configured* DB carries a coherent, different `project_id`. That is a genuine conflict of two authoritative signals, and main silently switches the workspace to the other project's database.
2. **Sole-candidate adoption in shared-server mode.** When `cfg.ProjectID == ""` and exactly one schema DB exists, doctor adopts it. In shared-server mode all projects share one server, so a lone foreign schema DB can belong to another workspace — with no local anchor, adoption redirects this workspace to another project's data.

### Fix quality — good, correctly scoped
- Guard 1 fires only in the truly-ambiguous case (`hasCurrent && current.HasSchema && current.ProjectID != "" && current.ProjectID != cfg.ProjectID`) and returns a clear actionable error without mutating `cfg`. Erroring for a human to resolve is the right conservative choice for two conflicting authoritative signals.
- Guard 2 (`&& !doltserver.IsSharedServerMode()`) is well-targeted: legit shared-server bootstrap where `dolt_database` is already correct is handled by the *earlier* backfill block (reached first); the sole-candidate fallback only fires when the DB name is *also* wrong — exactly the misattachment case worth blocking. Legitimate bootstrap/backfill preserved.
- The `current, ok` → `current, hasCurrent` hoist is a clean behavior-preserving refactor so both blocks share one lookup; `doltserver` already imported.

### Tests — good
- `ErrorsOnConflictingIdentitySignals` constructs the guard-1 scenario, asserts the error, `changed=false`, and no mutation of `cfg` (fails on main).
- `DoesNotAdoptSoleCandidateInSharedServerMode` sets shared-server via `t.Setenv`, single foreign schema DB, asserts no change (fails on main).
- Existing `UsesProjectIDToRepairDatabaseName` updated honestly (its old scenario is now correctly a conflict error, so the test keeps the still-valid clean-repair path).

### Minor, non-blocking
A user who hand-edits `metadata.json`'s `project_id` to "switch projects" while leaving `dolt_database` on the old coherent DB now gets an error instead of an auto-switch. That is not a documented switch flow, and a clear error beats a silent redirect. Scope is limited to `doctor --fix`; no runtime-path impact.

PR #4787, opened 2026-07-14
fix(comments): reject swapped-order `comments <id> add` instead of silent no-op (#4642)

### LINKED ISSUE #4642: bd comments <id> add <text> silently no-ops with exit 0 (swapped subcommand order)
## problem
`bd comments <id> add <text>` (issue-id before the `add` subcommand) silently no-ops with exit 0. It doesn't add the comment — it falls through to the parent `comments` command's own list behavior, printing existing comments and ignoring the trailing `add <text>` args entirely. Correct form is `bd comments add <id> <text>`.

## versions

## location
`cmd/bd/comments.go:14-100` — the `commentsCmd` parent command's `RunE`. It's registered with `Args: cobra.MinimumNArgs(1)` and only ever reads `args[0]` as the issue ID (line 44); everything after that is silently ignored. Cobra only dispatches to the `add` subcommand (`commentsAddCmd`, registered at line 120) when `add` is the *first* positional token. If the issue ID comes first, cobra has no matching subcommand name to route on, so it falls through to the parent's own list-and-print `RunE` — which happily accepts (and drops) any extra args.

## repro
```
$ bd create --silent "test issue for comments repro"
zz-4dq

$ bd comments add zz-4dq "hello from correct form"
Comment added to zz-4dq

$ bd comments zz-4dq add "hello from swapped form - should NOT appear"
Comments on zz-4dq:

[actor] at 2026-07-07 18:42
  hello from correct form
$ echo $?
0
```
The swapped invocation exits 0, prints only the pre-existing comment, and never writes "hello from swapped form" anywhere. In a scripted/agent pipeline (output often piped through `tail`/`head` or discarded on success), this reads as a successful comment-add and is only caught by re-listing comments afterward.

## suggested fix
Make the parent `comments` command reject unexpected trailing positional args instead of silently accepting them, e.g. in `cmd/bd/comments.go`'s `commentsCmd.RunE`: if `len(args) > 1`, error out with something like `unexpected extra arguments after issue-id %q — did you mean 'bd comments add %s %s'?`. This turns the silent no-op into an immediate, actionable error and costs nothing for the legitimate single-arg list case.

harry-miller-trimble:
Closes #4642

## Problem

`bd comments <issue-id> add <text>` — issue id *before* the `add` subcommand — silently no-ops with exit 0 (GH#4642). Cobra only routes to the `add` subcommand when `add` is the first positional token; with the id first it falls through to the parent `comments` command, whose `RunE` reads only `args[0]` (the issue id) and ignores everything after it. So it lists existing comments and exits 0 without ever writing the intended comment.

In scripted/agent pipelines (output often piped through `tail`/`head` or discarded on success) this reads as a successful add and is only caught by re-listing comments afterward.

## Repro (before)

```
$ bd comments zz-4dq add "hello from swapped form"
Comments on zz-4dq:
  ... (only pre-existing comments) ...
$ echo $?
0            # comment never stored
```

## Fix

The parent `comments` command only lists comments for a single issue, so any trailing positional arg is a mistake. Reject it with a clear error instead of dropping it. The common swapped-`add` case gets a targeted hint:

```
$ bd comments zz-4dq add "hello"
Error: "bd comments <issue-id> add ..." is not valid — the subcommand must come first.

To add a comment, run:
  bd comments add <issue-id> <text>

Example:
  bd comments add zz-4dq "your comment"

See: bd comments --help
$ echo $?
1
```

## Tests

Adds embedded regression tests (`comments_swapped_add_rejected`, `comments_extra_arg_rejected`) verifying the swapped form exits non-zero, surfaces the hint, and stores nothing.

### DIFF
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -1149,6 +1149,62 @@ func TestCLI_CommentsListMisplacedSyntax(t *testing.T) {
 	}
 }
 
+// TestCLI_CommentsSwappedAddRejectedBeforeStoreOpen is a regression test for
+// GH#4642's should-fix: the swapped-order rejection (`bd comments <id> add
+// <text>`) must fire from commentsCmd's Args validator, before
+// PersistentPreRunE ever opens a store or runs a migration, and before
+// RunE's usesProxiedServer() dispatch — not from a check inside RunE that
+// only the local-store branch reaches. Using a directory with no .beads/ at
+// all proves the store was never touched: the old RunE-only check needed an
+// already-initialized store (PersistentPreRunE would have failed first with
+// "no beads database found"), so this would fail with the wrong error before
+// the fix.
+func TestCLI_CommentsSwappedAddRejectedBeforeStoreOpen(t *testing.T) {
+	noBeadsCwd := t.TempDir()
+	if _, err := os.Stat(filepath.Join(noBeadsCwd, ".beads")); err == nil {
+		t.Fatalf("test setup: %s unexpectedly has a .beads dir", noBeadsCwd)
+	}
+
+	stdout, stderr, err := runBDInProcessAllowError(t, noBeadsCwd, "comments", "bd-123", "add", "should not be stored")
+	if err == nil {
+		t.Fatalf("expected non-zero exit for swapped-order add, got success:\nstdout: %s", stdout)
+	}
+	combined := stdout + stderr
+	if strings.Contains(combined, "no beads database found") {
+		t.Fatalf("Args validation did not run before store open: got the pre-fix error.\nOutput:\n%s", combined)
+	}
+	if !strings.Contains(combined, "bd comments add") {
+		t.Errorf("expected hint pointing to `bd comments add`, got:\n%s", combined)
+	}
+}
+
+// TestCLI_CommentsSwappedAddRejectedInProxiedServerMode is the proxied-server
+// counterpart: forcing proxiedServerMode simulates the dispatch that GH#4642's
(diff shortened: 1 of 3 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — right fix, wrong layer; one shared change fixes both findings.

**Blocker 1 — the rejection runs after store-open side effects.** Verified: the command keeps `Args: cobra.MinimumNArgs(1)` and rejects the swapped-order syntax only inside RunE. Cobra validates `Args` *before* `PersistentPreRunE`, but RunE runs after — and root pre-run performs version migration and opens the store. So the malformed invocation this PR exists to catch can still trigger migration work, hit the remote-migrate gate, or die with "no database found" before the helpful syntax hint ever prints. The embedded tests use an initialized store, so they can't see this.

**Blocker 2 — merge-base drift routes proxied mode around the fix.** Since this branched, main added `runCommentsProxiedServer`, dispatched before the auto-merge location of this PR's validation — and the proxied handler consumes only `args[0]`, ignoring trailing `add <text>`. The three-way merge is textually clean, so nothing will flag it: proxied `bd comments <id> add <text>` still silently lists comments. (This is the recurring direct-vs-proxied parity gap we're seeing across many PRs.)

**One fix for both:** move the tailored rejection into a custom `Args` validator on `commentsCmd` — it then fires before store open *and* before the proxied dispatch, covering every mode. Add one proxied regression test alongside.

maphew:

maphew (approved):

PR #4788, opened 2026-07-14
fix(create): reject whitespace-only titles (#4771)

### LINKED ISSUE #4771: bd create silently accepts an empty/missing title, defaulting it to the bare type string
Discovered 2026-07-14 in a downstream consumer repo (mlcc-tcad) during a PM tracker-hygiene pass.

**Bug:** `bd create --type task ...` (and presumably other issue types) silently ACCEPTS a missing/empty `--title` and defaults the stored title to the bare type string itself (e.g. a `task` created with no title ends up with `title: "task"`, a `bug` ends up with `title: "bug"`). No warning, no error, no non-zero exit.

**Why this matters:** a bead minted this way is functionally invisible in any title-based view — board scans, `bd search` (title/description only), title-sorted listings all show a meaningless `task`/`bug`/`epic` row indistinguishable from every other title-less mint. In our repo tonight, one agent repeated the same `bd create` invocation (omitting `--title`) four times before self-correcting, producing 6 title-less beads under one epic. Two of them were nearly closed as apparent noise/duplicates before a reviewing agent's own context caught that they carried real (but only comment-recorded) scope.

**Requested fix:** `bd create` should REFUSE (non-zero exit, clear error message) when title is empty/whitespace-only, rather than silently defaulting to the type string. If a default-title convenience is intentionally desired for some workflow, it should at minimum not collide with the bare type name (which reads as "no title" to any downstream consumer) — but our preference is a hard refusal, matching how most trackers treat a missing required field.

**Workaround in place downstream:** we're building a repo-local pre-mint guard (mirroring an existing duplicate-title guard) to catch this before a title-less bead ever lands, but the actual fix belongs here.

harry-miller-trimble:
Closes #4771

## Problem

`bd create` accepts a whitespace-only title, minting a bead with a blank title (GH#4771). Such a bead is functionally invisible in every title-based view — board scans, `bd search` (title/description only), title-sorted listings — and reads as "no title" to any downstream consumer. In the reporter's repo this produced multiple indistinguishable title-less beads under one epic, two of which were nearly closed as noise.

The existing guards reject a truly empty/absent title, but a title made of spaces or tabs slips past both the positional and `--title` paths:

```
$ bd create --title "   " --type bug
✓ Created issue: tt-vd1 —
$ bd create "  "
✓ Created issue: tt-36b —
```

## Fix

Add a single `strings.TrimSpace(title) == ""` check right after title resolution, so both the positional and `--title` paths refuse a blank title with a clear, non-zero-exit error:

```
$ bd create --title "   " --type bug
Error: title cannot be empty or whitespace-only
$ echo $?
1
```

Valid titles (including intentionally dash-prefixed `--title` values) are unaffected.

## Tests

`TestCLI_CreateRejectsEmptyTitle` covers whitespace-only titles via `--title` (spaces, tab) and via a positional arg.

## Note

The issue also reports the title defaulting to the bare *type* string (e.g. `title: "task"`). That exact symptom no longer reproduces on `main` — a missing/empty title is already rejected — but the whitespace-only gap remained, which is the same class of "title-less bead" the issue asks to prevent.

### DIFF
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -1350,6 +1350,49 @@ func TestCLI_CreateRejectsEmptyTitle(t *testing.T) {
 	}
 }
 
+// TestCLI_CreateRejectsEmptyTitle_ProxiedServerMode is a proxied-server
+// regression for GH#4771. Before this fix, the whitespace-only guard lived
+// only in create's local-store RunE branch: gatherCreateInput/
+// runCreateProxiedServer (the path taken when usesProxiedServer() is true)
+// never re-checked, so proxied-mode create still minted a blank-titled bead.
+// The fix moved the check into resolveTitle, invoked from createCmd's Args
+// validator — which runs for every invocation regardless of backend, before
+// RunE ever branches on usesProxiedServer(). Setting proxiedServerMode here
+// exercises that dispatch without needing a real proxied server: Args
+// validation must reject before gatherCreateInput/runCreateProxiedServer (or
+// any store open) ever runs.
+func TestCLI_CreateRejectsEmptyTitle_ProxiedServerMode(t *testing.T) {
+	origProxied := proxiedServerMode
+	t.Cleanup(func() { proxiedServerMode = origProxied })
+	proxiedServerMode = true
+
+	// createCmd is a shared package-level *cobra.Command, so a --title value
+	// set by an earlier in-process test invocation (e.g. TestCLI_CreateRejectsEmptyTitle's
+	// own FlagTab case) survives on the FlagSet across rootCmd.Execute() calls.
+	// Reset it explicitly so this test's outcome doesn't depend on suite
+	// ordering — a pre-existing gap, not something this test should also fall
+	// victim to.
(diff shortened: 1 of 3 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified at the PR head). Verdict: **needs changes** — one should-fix; the guard itself is right.

**Should-fix — proxied-server creation bypasses the new validation.** Verified: RunE returns through `gatherCreateInput`/`runCreateProxiedServer` at the top, before the whitespace-title guard; the input gatherer returns whitespace unchanged and the type-level check rejects only zero-length. So both advertised invocations still create blank-looking issues in proxied mode, and the new test covers only the ordinary backend. Since a title never configures store opening, the cleanest home is a shared validator both branches call — or, better, a custom `Args`/pre-open check, which also stops migrations running for an invocation that's guaranteed to fail.

Nit: the three-line inline rationale restates the commit message; the repo prefers inline comments for non-obvious invariants only.

Merge-base drift is clean and the finding survives on current main. Small fix, then this lands.

maphew:

maphew (approved):

PR #4790, opened 2026-07-14
fix(actor): let BEADS_ACTOR outrank deprecated BD_ACTOR (#4645)

### LINKED ISSUE #4645: BD_ACTOR (deprecated) unintentionally outranks BEADS_ACTOR due to viper env-prefix auto-binding
## problem
`BD_ACTOR` (documented as the deprecated alias) actually outranks `BEADS_ACTOR` (documented as the primary override) when both are set, because viper's automatic env-prefix binding picks up `BD_ACTOR` and pre-populates the `actor` package var before `getActorWithGit()` ever runs its own explicit priority check.

## versions

## location
- `cmd/bd/main.go:460-478` — `getActorWithGit()`'s doc comment and body state the intended priority: `--actor flag > BEADS_ACTOR env > BD_ACTOR env (deprecated) > git config user.name > $USER > "unknown"`. But it only reaches the `BEADS_ACTOR`/`BD_ACTOR` checks (lines 471-478) when the package-level `actor` var is still empty (line 466: `if actor != "" { return actor }`).
- `cmd/bd/main.go:736-738` (also `423-424` in `refreshBoundCommandConfig`) — before any command's `RunE` calls `getActorWithGit()`, this pre-run hook does `actor = config.GetString("actor")` whenever the `--actor` flag wasn't explicitly passed. This treats the viper-resolved value as equivalent to an explicit flag.
- `internal/config/config.go:165,169` — `v.SetEnvPrefix("BD")` + `v.AutomaticEnv()` makes viper auto-bind env var `BD_ACTOR` to the `actor` config key (prefix `BD` + key `ACTOR`). So `config.GetString("actor")` silently returns `BD_ACTOR`'s value whenever it's set, with no awareness of `BEADS_ACTOR` at all.

Net effect: if `BD_ACTOR` is set, `actor` gets populated from it during pre-run *before* `getActorWithGit()` runs. Then `getActorWithGit()`'s `if actor != ""` short-circuits and returns that value immediately, skipping the `BEADS_ACTOR` check entirely — even though `BEADS_ACTOR` is set and is supposed to win.

## repro
Reproduced against the raw `bd` binary (bypassing shell wrappers that might set their own env) in a scratch `.beads` repo:

```
$ env -u BD_ACTOR -u BEADS_ACTOR BD_ACTOR=bdactor BEADS_ACTOR=beadsactor bd create "repro both" --json | grep created_by
  "created_by": "bdactor"       # WRONG: BEADS_ACTOR should win per documented priority

$ env -u BD_ACTOR -u BEADS_ACTOR BEADS_ACTOR=beadsactor bd create "repro beads only" --json | grep created_by
  "created_by": "beadsactor"    # correct in isolation

$ env -u BD_ACTOR -u BEADS_ACTOR BD_ACTOR=bdactor bd create "repro bd only" --json | grep created_by
  "created_by": "bdactor"       # correct in isolation (deprecated alias still works)
```

Only when both are set does the documented precedence invert.

## suggested fix
Keep `BD_ACTOR` working as a deprecated alias (it's a deliberate channel — don't remove it). Fix the precedence instead:
- Stop letting viper's `AutomaticEnv` auto-bind `BD_ACTOR` onto the `actor` config key. The `actor` key's only legitimate sources should be the `--actor` flag and `config.yaml`'s `actor:` value — `getActorWithGit()` already owns the `BEADS_ACTOR` > `BD_ACTOR` > git > `$USER` fallback chain explicitly and correctly.
- Concretely: either unbind `actor` from automatic env (so `config.GetString("actor")` no longer surfaces `BD_ACTOR`), or have `refreshBoundCommandConfig`/the pre-run hook only assign `actor = config.GetString(...)` from the YAML-file-backed value, not viper's auto-env-resolved value, leaving env-var precedence entirely to `getActorWithGit()`.

harry-miller-trimble:
Closes #4645

## Problem

The documented actor priority (in `getActorWithGit`'s doc comment and `TestGetActorWithGit`) is:

```
--actor flag > BEADS_ACTOR env > BD_ACTOR env (deprecated) > git user.name > $USER > "unknown"
```

But when both `BEADS_ACTOR` and `BD_ACTOR` are set, the **deprecated** `BD_ACTOR` won.

## Root cause

`internal/config` calls `v.SetEnvPrefix("BD")` + `v.AutomaticEnv()`, which binds `BD_ACTOR` → the `actor` config key. viper consults the prefixed auto-binding *ahead of* any explicit `BindEnv`, so `config.GetString("actor")` returns `BD_ACTOR`'s value even when `BEADS_ACTOR` is also set (`BEADS_ACTOR` isn't under the `BD` prefix, so viper never maps it to `actor`).

The pre-run hooks (`refreshBoundCommandConfig` and the persistent pre-run) pre-populate the global `actor` from `config.GetString("actor")` when `--actor` wasn't passed. `getActorWithGit()` then hits its `if actor != "" { return actor }` early return and hands back the `BD_ACTOR` value, never reaching its own `BEADS_ACTOR` check.

## Fix

Add `resolveConfiguredActor()`, which checks `BEADS_ACTOR` explicitly before falling back to `config.GetString("actor")`, and use it in both pre-run hooks. This restores the documented precedence without touching viper's global env wiring (which other keys depend on). `BD_ACTOR` and a `config.yaml` `actor:` value still apply when `BEADS_ACTOR` is unset.

## Verification

End-to-end with an embedded workspace:

```
# before (upstream/main)
$ BEADS_ACTOR=beads-wins BD_ACTOR=bd-loses bd create "x" --json | jq .created_by
"bd-loses"      # bug

# after
"beads-wins"    # BEADS_ACTOR wins
$ BD_ACTOR=bd-only bd create "y" --json | jq .created_by
"bd-only"       # deprecated alias still works as fallback
```

Adds `TestResolveConfiguredActor` covering both cases.

### DIFF
--- a/cmd/bd/actor_test.go
+++ b/cmd/bd/actor_test.go
@@ -5,6 +5,8 @@ import (
 	"os/exec"
 	"strings"
 	"testing"
+
+	"github.com/steveyegge/beads/internal/config"
 )
 
 // TestGetActorWithGit tests the actor resolution fallback chain.
@@ -153,6 +155,51 @@ func TestGetActorWithGit(t *testing.T) {
 			}
 		})
 	}
+}
+
+// TestResolveConfiguredActor verifies that BEADS_ACTOR (primary override)
+// outranks the deprecated BD_ACTOR alias when the --actor flag is not set,
+// even though viper's AutomaticEnv binds BD_ACTOR to the "actor" config key
+// (GH#4645). This is the resolution the pre-run hooks use to pre-populate the
+// global actor before getActorWithGit runs.
+func TestResolveConfiguredActor(t *testing.T) {
+	origBd, bdSet := os.LookupEnv("BD_ACTOR")
+	origBeads, beadsSet := os.LookupEnv("BEADS_ACTOR")
+	defer func() {
+		if bdSet {
+			os.Setenv("BD_ACTOR", origBd)
+		} else {
+			os.Unsetenv("BD_ACTOR")
+		}
+		if beadsSet {
+			os.Setenv("BEADS_ACTOR", origBeads)
+		} else {
+			os.Unsetenv("BEADS_ACTOR")
+		}
+		_ = config.Initialize()
+	}()
+
+	t.Run("BEADS_ACTOR outranks BD_ACTOR", func(t *testing.T) {
+		os.Setenv("BD_ACTOR", "from-bd-actor")
+		os.Setenv("BEADS_ACTOR", "from-beads-actor")
+		if err := config.Initialize(); err != nil {
+			t.Fatalf("config.Initialize(): %v", err)
+		}
+		if got := resolveConfiguredActor(); got != "from-beads-actor" {
+			t.Errorf("resolveConfiguredActor() = %q, want %q (BEADS_ACTOR must win)", got, "from-beads-actor")
+		}
+	})
+
+	t.Run("BD_ACTOR used when BEADS_ACTOR unset", func(t *testing.T) {
+		os.Unsetenv("BEADS_ACTOR")
+		os.Setenv("BD_ACTOR", "from-bd-actor")
+		if err := config.Initialize(); err != nil {
+			t.Fatalf("config.Initialize(): %v", err)
+		}
+		if got := resolveConfiguredActor(); got != "from-bd-actor" {
+			t.Errorf("resolveConfiguredActor() = %q, want %q (BD_ACTOR fallback)", got, "from-bd-actor")
+		}
+	})
 }
 
 // TestGetActorWithGit_PriorityOrder tests that the priority order is respected
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -248,6 +248,37 @@ func TestCLI_Create(t *testing.T) {
 	}
 }
 
+// TestCLI_CreateActorPrecedence is a CLI-level regression test for GH#4645:
+// with both BEADS_ACTOR and the deprecated BD_ACTOR set, `bd create` must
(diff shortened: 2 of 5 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace through pinned Viper; Claude adjudication spot-checked at the PR head). Verdict: **needs changes** — the precedence fix is right and correctly placed (the helper runs before migration and store open); two should-fixes before landing.

**Should-fix 1 — `bd config show` now contradicts the actor actually used.** The runtime path correctly prefers `BEADS_ACTOR` over deprecated `BD_ACTOR` and feeds it through to `created_by`. But `config show` still resolves the actor through the Viper-backed config layer, whose automatic `BD_`-prefix environment binding surfaces `BD_ACTOR` first (traced through pinned Viper v1.21.0's env-binding order; the config layer itself only mentions the deprecation in a comment). With both variables set, mutations use the `BEADS_ACTOR` value while `config show` reports the `BD_ACTOR` value with `env: BD_ACTOR` provenance — breaking show's "all effective configuration" contract. The show path needs the same precedence helper.

**Should-fix 2 — the regression test doesn't capture the wiring.** It calls `resolveConfiguredActor()` directly, so reverting the root pre-run assignment or the target-context rebinding leaves it green. Add a CLI-level create with both env vars set asserting `created_by`, and a `config show` assertion for value + provenance.

Nits: the old `getActorWithGit` doc comment is now attached to the new helper — move it back; and the two repeated test cases want a table.

Merge-base drift is clean; main's intervening changes don't touch actor precedence.

maphew:

maphew (approved):

PR #4791, opened 2026-07-14
fix(bootstrap): commit .beads workspace files so adopt-from-remote leaves a clean tree (#4644)

### LINKED ISSUE #4644: bd bootstrap leaves .beads/.gitignore dirty after appending missing required patterns (no commit)
## problem
When `bd bootstrap` (sync-clone / adopt-from-remote flow) finds an existing `.beads/.gitignore` missing newer required patterns, it appends them via a bare `os.WriteFile` and leaves the file — and the working tree — dirty, with no commit and no notice to the user. This differs from plain `bd init` on a fresh repo, which explicitly stages and commits `.beads/` (including a freshly-templated `.gitignore`) as part of its own flow. Repos that were bootstrapped/adopted rather than freshly `bd init`'d accumulate this dirty hunk every time a bd upgrade adds a new required pattern.

## versions

## location
`cmd/bd/doctor/gitignore.go:172-212` — `EnsureGitignoreForBeadsDir`. When the file exists but is missing required patterns, it appends them and calls `os.WriteFile(gitignorePath, []byte(newContent), 0600)` at line 207 with no accompanying git operation. `EnsureGitignoreForBeadsDir` has exactly one caller in the whole codebase: `cmd/bd/bootstrap.go:754`, inside `finalizeSyncedBootstrap` (the function that "writes metadata.json and config.yaml after a successful sync clone, matching the on-disk layout that `bd init` produces" — per its own doc comment at line 721-725). Unlike `bd init`'s normal flow, which auto-stages and commits with message `"bd init: initialize beads issue tracking"` (`cmd/bd/init.go:1583-1629`, guarded on `!stealth && isGitRepo() && useLocalBeads`), `finalizeSyncedBootstrap` has no equivalent `git add`/`git commit` step after calling `EnsureGitignoreForBeadsDir`. The existing unit test (`cmd/bd/doctor/gitignore_test.go:360-394`, `TestEnsureGitignoreForBeadsDir_AppendsMissingRuntimePatterns`) only asserts the file content is correct — it doesn't touch git state, so this gap has no test coverage either.

## repro
Reproduces on any repo that went through `bd bootstrap` (or an equivalent sync-clone adopt) against a `.beads/.gitignore` predating a newer required pattern (e.g. `.local_version` or `backup/`, both added at some point after initial rollout — see the test's own `stalePatterns` exclusion list at `gitignore_test.go:369`). Observed in practice across multiple real repos: after a bd version bump that added new required patterns, the next bootstrap/sync-adopt appended a `# Added by bd (missing required patterns)` hunk to `.beads/.gitignore` with no commit, leaving `git status` dirty until a human or agent manually ran `git commit --only .beads/.gitignore`. This recurred independently per-repo and per-machine (4 of 7 beads repos dirty with the identical hunk on one machine alone), because each affected clone hits the append path once and is never auto-committed.

## suggested fix
Either:
1. Generate the complete, current required pattern set in `.beads/.gitignore` at the point a repo is first set up (both `bd init`'s template write and `finalizeSyncedBootstrap`'s path), so there's nothing left to append later on the common path.
2. For the migration case (older `.beads/.gitignore` missing newer patterns): make `finalizeSyncedBootstrap` mirror `bd init`'s existing auto-stage-and-commit block (`cmd/bd/init.go:1583-1629`) after calling `EnsureGitignoreForBeadsDir`, or at minimum print a one-line notice ("appended N required ignore patterns to .beads/.gitignore — commit it") instead of silently dirtying the tree.

harry-miller-trimble:
Closes #4644

## Problem

The standalone `bd bootstrap` sync path (`executeSyncAction` → `finalizeSyncedBootstrap`) writes `.beads/config.yaml` and `metadata.json`, and appends any newer required patterns to an existing `.beads/.gitignore` via `EnsureGitignoreForBeadsDir` — but never commits them. The working tree (and a *tracked* `.beads/.gitignore`) is left dirty with no notice.

`bd init` on a fresh repo, by contrast, explicitly stages and commits `.beads/`. So repos that were **bootstrapped/adopted** rather than freshly `bd init`'d accumulate a dirty `.gitignore` hunk every time a bd upgrade adds a new required pattern.

## Fix

Add `commitBeadsWorkspaceFiles`, called right after `finalizeSyncedBootstrap` in the sync path. It brings the bootstrap flow to parity with init, but is deliberately conservative:

- **no-ops** outside a (non-bare) git repo;
- only acts when `git status --porcelain -- .beads/` shows actual changes;
- commits with an **explicit `.beads/` pathspec**, so any unrelated staged changes the user had in flight are left untouched (this is stricter than init's bare `git commit`);
- uses `--no-verify` (mirroring bd init) so bootstrap-installed hooks can't recurse into bd while the embedded Dolt lock may still be held.

The `bd init` sync path (which calls `finalizeSyncedBootstrap` and then does its own `git add .beads/` + commit) is unaffected — by the time its commit runs there's nothing new under `.beads/`, so it's a clean no-op.

## Tests

- `TestCommitBeadsWorkspaceFiles`: after a stale committed `.gitignore` is appended-to and fresh workspace files are written, the tree is clean afterward, an unrelated staged file remains staged, and the commit is labeled `bd bootstrap: …`.
- `TestCommitBeadsWorkspaceFiles_NoGitRepo`: safe no-op outside a git repo.

### DIFF
--- a/cmd/bd/bootstrap.go
+++ b/cmd/bd/bootstrap.go
@@ -740,6 +740,12 @@ func executeSyncAction(ctx context.Context, plan BootstrapPlan, cfg *configfile.
 		return err
 	}
 
+	// Commit the workspace files bootstrap just wrote/updated (config.yaml,
+	// metadata.json, and any appended .beads/.gitignore patterns) so an
+	// adopt-from-remote flow leaves a clean working tree, matching bd init
+	// (GH#4644).
+	commitBeadsWorkspaceFiles(plan.BeadsDir)
+
 	// Open and close the store to ensure dolt_ignore'd wisp tables are
 	// created in the working set. Clone does not include these tables
 	// (they are never committed), so they must be recreated after clone.
--- a/cmd/bd/bootstrap.go
+++ b/cmd/bd/bootstrap.go
@@ -740,12 +740,6 @@ func executeSyncAction(ctx context.Context, plan BootstrapPlan, cfg *configfile.
 		return err
 	}
 
-	// Commit the workspace files bootstrap just wrote/updated (config.yaml,
-	// metadata.json, and any appended .beads/.gitignore patterns) so an
-	// adopt-from-remote flow leaves a clean working tree, matching bd init
-	// (GH#4644).
-	commitBeadsWorkspaceFiles(plan.BeadsDir)
-
 	// Open and close the store to ensure dolt_ignore'd wisp tables are
 	// created in the working set. Clone does not include these tables
 	// (they are never committed), so they must be recreated after clone.
@@ -779,6 +773,17 @@ func executeSyncAction(ctx context.Context, plan BootstrapPlan, cfg *configfile.
 	}
 	configureInitDoltRemote(ctx, warmupStore, plan.SyncRemote, false)
 	_ = warmupStore.Close()
+
+	// Commit the workspace files bootstrap just wrote/updated (config.yaml,
+	// metadata.json, and any appended .beads/.gitignore patterns) so an
+	// adopt-from-remote flow leaves a clean working tree, matching bd init
+	// (GH#4644).
+	//
+	// Deliberately last: every way this bootstrap can still fail (most
+	// notably the remote-migrate gate above, which returns a non-zero exit)
+	// happens before it, so a failed bootstrap never leaves a commit behind
+	// advertising a workspace it did not finish setting up.
+	commitBeadsWorkspaceFiles(plan.BeadsDir)
 
 	return nil
 }
--- a/cmd/bd/commit_beads_workspace_test.go
+++ b/cmd/bd/commit_beads_workspace_test.go
@@ -1,0 +1,117 @@
+package main
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// gitRun runs a git command in dir and fails the test on error.
+func gitRun(t *testing.T, dir string, args ...string) {
+	t.Helper()
+	cmd := exec.Command("git", args...)
+	cmd.Dir = dir
+	if out, err := cmd.CombinedOutput(); err != nil {
+		t.Fatalf("git %s failed: %v\n%s", strings.Join(args, " "), err, out)
+	}
+}
+
+// gitPorcelain returns `git status --porcelain` output for dir.
+func gitPorcelain(t *testing.T, dir string) string {
+	t.Helper()
+	cmd := exec.Command("git", "status", "--porcelain")
+	cmd.Dir = dir
+	out, err := cmd.Output()
+	if err != nil {
+		t.Fatalf("git status failed: %v", err)
+	}
+	return strings.TrimSpace(string(out))
+}
+
+// TestCommitBeadsWorkspaceFiles verifies that the bootstrap workspace-commit
+// helper leaves a clean working tree after .beads/ files are written/updated,
+// and that it does not sweep up unrelated staged changes (GH#4644).
+func TestCommitBeadsWorkspaceFiles(t *testing.T) {
+	repo := t.TempDir()
+	gitRun(t, repo, "init")
+	gitRun(t, repo, "config", "user.email", "test@example.com")
+	gitRun(t, repo, "config", "user.name", "Test")
+
+	beadsDir := filepath.Join(repo, ".beads")
+	if err := os.MkdirAll(beadsDir, 0o750); err != nil {
+		t.Fatal(err)
+	}
+	gitignore := filepath.Join(beadsDir, ".gitignore")
+
+	// Simulate an already-adopted repo: a committed but stale .gitignore.
+	if err := os.WriteFile(gitignore, []byte("dolt/\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	gitRun(t, repo, "add", ".beads/.gitignore")
+	gitRun(t, repo, "commit", "-m", "initial beads gitignore")
+	if s := gitPorcelain(t, repo); s != "" {
+		t.Fatalf("expected clean tree after initial commit, got:\n%s", s)
+	}
+
+	// A concurrent, unrelated staged change the user is preparing.
+	unrelated := filepath.Join(repo, "unrelated.txt")
+	if err := os.WriteFile(unrelated, []byte("wip\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	gitRun(t, repo, "add", "unrelated.txt")
+
+	// Bootstrap appends a missing pattern (as EnsureGitignoreForBeadsDir does)
+	// and writes fresh workspace files, leaving them dirty/untracked.
+	if err := os.WriteFile(gitignore, []byte("dolt/\nembeddeddolt/\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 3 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the goal (clean tree after adopt-from-remote) is right; the commit helper's blast radius is too wide.

**Blocker 1 — the helper commits unrelated `.beads` content.** Verified: status/add/commit all target the whole `beadsDir` pathspec, while the helper's stated ownership is `metadata.json`, `config.yaml`, `.gitignore`. Legitimate tracked content lives under `.beads/` (formulas, git-tracked `issues.jsonl`), so a user's pending `.beads` edit is silently swept into the bootstrap commit. The test protects only a file *outside* `.beads`. Use exact file pathspecs and add a collateral-file-inside-`.beads` test.

**Blocker 2 — bootstrap from a linked worktree can commit on the main checkout's branch.** Fresh-clone bootstrap can resolve `plan.BeadsDir` to the main checkout's shared `.beads`; the helper derives its git dir from that path and commits there — advancing the *other* checkout's HEAD. The sibling code in the same file already binds git operations via `beads.GetRepoContext()`; this helper should too, with a linked-worktree regression asserting the other checkout's HEAD never moves.

**Should-fix — semantic drift on hook skipping:** since this branched, main fixed exactly this misconception (`git -c core.hooksPath= commit --no-verify` — `--no-verify` alone still runs `prepare-commit-msg`), with a regression test. Adopt the same invocation.

**Should-fix — the commit lands before the store-open migration gate:** git history is written, then the warmup open can fail on `RemoteMigrateGateError`, leaving a failed bootstrap that nonetheless created a commit. Commit after the store opens cleanly, or document+test the partial-failure shape.

harry-miller-trimble:
Thanks — all four points addressed, and the branch is rebased onto current `main`. One of them I implemented differently than suggested; reasoning below.

**Blocker 1 — commits unrelated `.beads` content: fixed.** The helper now builds exact file pathspecs for the three files bootstrap owns (`metadata.json`, `config.yaml`, `.gitignore`) instead of the `.beads/` directory. New test `TestCommitBeadsWorkspaceFiles_LeavesCollateralBeadsContentAlone` commits a tracked `.beads/issues.jsonl` and `.beads/formulas/mol-x.formula.toml`, dirties both, and asserts they survive uncommitted. Confirmed it fails against the old directory pathspec:

```
--- FAIL: TestCommitBeadsWorkspaceFiles_LeavesCollateralBeadsContentAlone
    expected issues.jsonl to stay uncommitted (bootstrap does not own it), status:
    expected formulas/mol-x.formula.toml to stay uncommitted (bootstrap does not own it), status:
```

**Blocker 2 — linked worktree: fixed, but not via `GetRepoContext()`.** The helper now resolves the checkout that physically contains `beadsDir` with `git -C <beadsDir> rev-parse --show-toplevel` and refuses to act if `beadsDir` resolves outside that root (both sides through `EvalSymlinks`, since git reports `/private/var/...` on macOS while the caller holds `/var/...`).

I did not route it through `beads.GetRepoContext()`, because for a linked worktree that is the mechanism that *causes* the hazard rather than preventing it. In `buildRepoContext`:

- `isExternalBeadsDir` compares **git common dirs**, which are identical for a worktree and its main checkout — so a worktree's own `.beads` is not "external" and `isExternal` stays false;
- the non-external branch sets `repoRoot, err = git.GetMainRepoRoot()`, which for a worktree returns `filepath.Dir(ctx.commonDir)` — **the main checkout**;
- `RepoContext.GitCmd` then pins `cmd.Dir`, `GIT_DIR`, and `GIT_WORK_TREE` to that `RepoRoot`.

So `rc.GitCmd(...)` on a worktree's `.beads` paths runs git against the main checkout's work tree — which is exactly the "advances the other checkout's HEAD" failure the blocker asks to prevent. `TestCommitBeadsWorkspaceFiles_LinkedWorktreeDoesNotTouchMainCheckout` pins the intended behavior: it creates a linked worktree with its own `.beads`, runs the helper, and asserts the main checkout's HEAD **and** branch are unchanged while the worktree's HEAD advances and its `.beads` files come out clean. That test fails under a `GetRepoContext()`-bound implementation, which is the point of keeping it. Happy to revisit if I've misread the intended `RepoRoot` semantics for worktrees.

**Should-fix — hook skipping: adopted.** Now `-c core.hooksPath= commit --no-verify`, matching `init.go:2167` and its comment that `--no-verify` alone does not skip `prepare-commit-msg`. `TestCommitBeadsWorkspaceFiles_BypassesHooks` installs a `prepare-commit-msg` that appends `.hook-ran` and exits 1, then asserts the marker is absent and the commit still lands — same shape as init's `auto_commit_bypasses_hooks`.

**Should-fix — commit before the migration gate: fixed.** The call moved from before the warmup store open to after `warmupStore.Close()`, so the `RemoteMigrateGateError` path (and the non-fatal store-init warning path) both return before any commit is made. A bootstrap that exits non-zero no longer leaves a commit behind advertising a workspace it did not finish setting up.

Also worth noting: because the status probe now runs against the owned files only, a repo that gitignores `.beads/` entirely lands in the existing "nothing changed" early return rather than erroring on `git add` of an ignored path.

Tests: `TestCommitBeadsWorkspaceFiles*` and the bootstrap suite pass locally (`go test -tags gms_pure_go ./cmd/bd/ -run 'CommitBeadsWorkspaceFiles|Bootstrap'`), `gofmt` and `go vet` clean.

maphew (approved):
Re-review of the rebased two-commit branch: all four points from my 2026-07-24 review are resolved. Approving.

- **Blocker 1** (whole-directory pathspec sweeping unrelated `.beads` content): fixed as asked - `beadsWorkspaceFiles` is now the exact trio (`metadata.json`, `config.yaml`, `.gitignore`) with per-file pathspecs after `--`, plus the collateral-content regression test. I empirically confirmed unrelated staged changes survive and the pathspec commit works on an unborn HEAD.
- **Blocker 2** (linked-worktree bootstrap committing on the main checkout): done differently, and your way is better than what I suggested. `beads.GetRepoContext()` resolves a linked worktree to the MAIN checkout root (the common-dir comparison makes the worktree's `.beads` non-external), so my suggested mechanism would itself have caused the cross-checkout commit the blocker forbids. Your `gitWorktreeRootContaining()` via `git -C <beadsDir> rev-parse --show-toplevel` binds to the checkout that physically contains `.beads`, is symlink-safe (EvalSymlinks-normalized containment), no-ops in bare repos, and the regression test pins that the main checkout's HEAD/branch never move. Verified against `internal/beads/context.go` and `internal/git/gitdir.go`.
- **Hook bypass**: now byte-for-byte the same `-c core.hooksPath= commit --no-verify` invocation as `init.go`, with its own regression test.
- **Commit ordering**: `commitBeadsWorkspaceFiles` moved after the store opens cleanly (after `configureInitDoltRemote` and the warmup close), so a `RemoteMigrateGateError` bootstrap no longer leaves a commit behind.

New-commit scan found nothing gating. Two residual edges, noted only for the record since both preserve pre-PR behavior: the non-fatal store-open warning path returns before the commit call (consistent with "commit only after a clean open"), and a pathspec-limited commit during an in-progress merge degrades to the stderr warning and a dirty tree. CI is fully green and the branch is mergeable.

One process note: I stepped down as maintainer on 2026-08-10 and can no longer land this. @csells @julianknutsen - ready for merge from my side.

PR #4792, opened 2026-07-14
fix(context): resolve from .beads when outside a git repository (#4772)

### LINKED ISSUE #4772: bd context fails outside a git repository, contradicting its own docs
## Problem

`bd context --json` errors out when the working directory is not inside a git repository, even though the command's own help says it "reads directly from config files and does not require the database to be open, making it useful for diagnostics in degraded states." A `.beads/` directory with valid `metadata.json` and config is present and sufficient to answer the question, but context resolution hard-requires a git repo root first.

## Impact

Any consumer that uses `bd context` as a health/identity probe gets a false "unreachable" in perfectly healthy non-git scopes. Concretely: Gas City's native-store preflight cross-verifies backends via `bd context --json`; in a city directory without `.git`, the probe fails, the preflight degrades, and the whole deployment silently falls back to a fork-per-operation store — costing a sustained CPU core on the Dolt server (downstream: gastownhall/gascity#4245, gastownhall/gascity#4246). The fix on my machine was literally `git init` in the scope directory.

## Expected vs actual

- **Expected:** with a discoverable `.beads/` directory, `bd context` resolves the scope from it (or from `--db`/env) and reports backend identity; git is a fallback for finding the repo root, not a hard requirement.
- **Actual:**
  ```
  $ bd context --json      # in a dir with .beads/ but no .git
  {
    "error": "cannot resolve repo context: cannot determine repository root: not a git repository: exit status 128",
    "schema_version": 1
  }
  ```

## Evidence

Same directory after `git init` (no commits, nothing else changed):

```
$ git init && bd context --json
{
  "backend": "dolt",
  "bd_version": "1.1.0",
  "beads_dir": "/home/.../gascity-hq/.beads",
  "dolt_mode": "server",
  ...
}
```

---

<details>
<summary>Machine detail</summary>

**Repro**
1. `mkdir scope && cd scope` — do NOT `git init`.
2. Create a working `.beads/` (dolt server backend in my case; likely reproduces with any backend).
3. `bd context --json` → `cannot resolve repo context: cannot determine repository root: not a git repository: exit status 128`.
4. `git init` → command succeeds with full context output.

**Notes**
- bd 1.1.0, Linux.
- Other bd commands in the same non-git scope work fine when pointed at the store (e.g. `bd list --json` against the dolt server), so the git requirement appears specific to repo-context resolution rather than store access.
- Suggested resolution order: explicit flag/env → walk up for `.beads/` → git repo root as last resort. Alternatively, degrade gracefully: report the fields that don't need a repo root (backend, dolt_mode, bd_version, schema_version) and null the repo-root-derived ones.

**Post-merge verification**
- The repro above returns backend identity (exit 0) in a `.beads/`-bearing directory with no `.git`.

</details>

harry-miller-trimble:
Closes #4772

## Problem

`bd context`'s own help says it "reads directly from config files and does not require the database to be open, making it useful for diagnostics in degraded states." But it hard-failed when the working directory wasn't inside a git repository, even with a valid `.beads/` present:

```
$ bd context --json      # dir with .beads/ but no .git
{
  "error": "cannot resolve repo context: cannot determine repository root: not a git repository: exit status 128",
  "schema_version": 1
}
```

Consumers using `bd context --json` as a health/identity probe get a false "unreachable" in perfectly healthy non-git scopes. (Reporter: Gas City's native-store preflight degraded to a fork-per-operation store — a sustained CPU core — because the probe failed in a city directory without `.git`.)

## Fix

`beads.GetRepoContext()` finds and boundary-validates `.beads/` *before* it resolves the git repo root, so a `cannot determine repository root` failure means the only missing piece is git. That one failure now returns a typed `*beads.NoRepoRootError` carrying the already-validated `.beads` directory, and `beads.GetRepoContextAllowingNoGit()` selects it with `errors.As` and synthesizes the context with the `.beads` parent as the repo root. Backend identity then resolves from config, as documented.

Scoped deliberately:
- Only the typed `*NoRepoRootError` triggers the fallback, and it is emitted at the single `GetRepoMainRoot` failure site — after `.beads` discovery and after the SEC-003 boundary check. `no .beads directory found` and unsafe-location errors still surface unchanged. `TestNoRepoRootError_DoesNotMatchUnsafeLocation` pins that a hostile path spelling the error phrase no longer selects the fallback.
- Only the config-reading callers opt in. `bd context` calls `GetRepoContextAllowingNoGit` on the direct route, and the `domain/fs` context provider (`contextRepositoryImpl.RepoContext`/`.Role`) does so for the proxied route. Everything that actually runs git commands still calls `GetRepoContext`, so a missing repository stays an error where it matters.
- One knock-on worth naming: `bd serve` resolves its workspace through that same `domain/fs` context provider (`cmd/bd/serve.go:266`), so `bd serve` in a non-git directory with a valid `.beads/` now proceeds instead of failing at context resolution. That is the same config-only identity resolution this PR is about, and consistent with it — but it is a behavior change beyond `bd context`, so it should not go unmentioned. (Thanks @maphew for catching that the earlier "other commands' behavior is untouched" claim had gone stale.)
- And the shape of that knock-on: the fallback inherits `FindBeadsDir`'s ancestor walk, which outside a git repo has no `walkBoundary` and so climbs until it finds a `.beads/` with project files or reaches `/`. That is the same answer you already get from a subdirectory *inside* a git repo, so it is the consistent one rather than a new behavior — but it means `bd serve` in a scratch directory nested under a project now binds that project's workspace instead of failing. It stays bounded in practice: `hasBeadsProjectFiles` rejects a registry-only `~/.beads`, and the SEC-003 boundary rejects system paths. (Thanks @bee-ghosttrack for chasing this one down and confirming it is not a defect.)

**What the synthesized context can and cannot know.** `RepoRoot` is the `.beads` parent, `CWDRepoRoot` is empty, `IsWorktree` is false (nothing to be a worktree of), and `IsRedirected` is true exactly when `BEADS_DIR` names the resolved directory. That last one is not a guess: `isExternalBeadsDir` compares git *common dirs*, and the CWD side of that comparison needs a repository — which is precisely the state this branch serves — so an explicit `BEADS_DIR` is the only evidence left, and it is the same signal `FindBeadsDir`'s step 1 used to choose the directory in the first place. `engdocs/REPO_CONTEXT.md` now carries this as a table.

## Verification

Reproduced the exact error on `main` in a non-git dir with `.beads/`; with the fix `bd context --json` reports `backend: dolt`, a populated `beads_dir`, and `repo_root` = the `.beads` parent.

`*NoRepoRootError` means "git could not tell us a root", which is broader than "there is no repo" — a dangling gitfile, a dubious-ownership refusal, or git missing from `PATH` all arrive the same way. Recovery is equally safe in all of them (the workspace is located and boundary-checked either way), but the command whose pitch is diagnostics in degraded states shouldn't swallow the fact, so `bd context`'s text output prints `cwd repo: git: unavailable` rather than nothing — the counterpart of the omitted `cwd_repo_root` JSON field.

Regression coverage decodes `ContextInfo` and asserts exact values through a shared `assertNonGitContext` helper (`beads_dir`, `repo_root`, backend, non-empty `bd_version`, `is_worktree=false`) on both the embedded route (`context_json_without_git_repo`) and the proxied route (`TestProxiedServerContext`).

### DIFF
--- a/cmd/bd/context_cmd.go
+++ b/cmd/bd/context_cmd.go
@@ -2,8 +2,6 @@ package main
 
 import (
 	"fmt"
-	"path/filepath"
-	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/beads/internal/beads"
@@ -77,23 +75,7 @@ Examples:
 			prepareSelectedNoDBContext(selected)
 		}
 
-		rc, err := beads.GetRepoContext()
-		if err != nil && strings.Contains(err.Error(), "cannot determine repository root") {
-			// bd context is documented to work in degraded states and reads
-			// only from config files. Reaching this error means a valid,
-			// boundary-checked .beads directory was already found and the sole
-			// failure was the absence of a git repository root. Git is only a
-			// means of locating the repo root, not a hard requirement here, so
-			// fall back to the .beads parent as the repo root and continue —
-			// backend identity still resolves from config (GH#4772).
-			if beadsDir := beads.FindBeadsDir(); beadsDir != "" {
-				rc = &beads.RepoContext{
-					BeadsDir: beadsDir,
-					RepoRoot: filepath.Dir(beadsDir),
-				}
-				err = nil
-			}
-		}
+		rc, err := beads.GetRepoContextAllowingNoGit()
 		if err != nil {
 			if jsonOutput {
 				if jerr := outputJSON(map[string]string{"error": fmt.Sprintf("cannot resolve repo context: %v", err)}); jerr != nil {
--- a/cmd/bd/context_cmd.go
+++ b/cmd/bd/context_cmd.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"fmt"
+	"path/filepath"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/beads/internal/beads"
@@ -76,6 +78,22 @@ Examples:
 		}
 
 		rc, err := beads.GetRepoContext()
+		if err != nil && strings.Contains(err.Error(), "cannot determine repository root") {
+			// bd context is documented to work in degraded states and reads
+			// only from config files. Reaching this error means a valid,
+			// boundary-checked .beads directory was already found and the sole
+			// failure was the absence of a git repository root. Git is only a
+			// means of locating the repo root, not a hard requirement here, so
+			// fall back to the .beads parent as the repo root and continue —
+			// backend identity still resolves from config (GH#4772).
+			if beadsDir := beads.FindBeadsDir(); beadsDir != "" {
+				rc = &beads.RepoContext{
+					BeadsDir: beadsDir,
+					RepoRoot: filepath.Dir(beadsDir),
+				}
+				err = nil
+			}
+		}
 		if err != nil {
 			if jsonOutput {
 				if jerr := outputJSON(map[string]string{"error": fmt.Sprintf("cannot resolve repo context: %v", err)}); jerr != nil {
--- a/cmd/bd/context_embedded_test.go
+++ b/cmd/bd/context_embedded_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"sync"
 	"testing"
@@ -43,6 +44,31 @@ func TestEmbeddedContext(t *testing.T) {
 		}
 		if len(strings.TrimSpace(stdout.String())) == 0 {
 			t.Error("expected non-empty context --json output")
+		}
+	})
+
+	// bd context must resolve from a discoverable .beads directory even when
+	// the scope is not inside a git repository — git is a means of finding the
+	// repo root, not a hard requirement for this diagnostic (GH#4772).
+	t.Run("context_json_without_git_repo", func(t *testing.T) {
+		ngDir, _, _ := bdInit(t, bd, "--prefix", "ng")
+		if err := os.RemoveAll(filepath.Join(ngDir, ".git")); err != nil {
+			t.Fatalf("remove .git: %v", err)
+		}
+
+		cmd := exec.Command(bd, "context", "--json")
+		cmd.Dir = ngDir
+		cmd.Env = bdEnv(ngDir)
+		stdout, stderr, err := runCommandBuffers(t, cmd)
+		if err != nil {
+			t.Fatalf("bd context --json failed outside git repo: %v\nstdout:\n%s\nstderr:\n%s", err, stdout.String(), stderr.String())
+		}
+		out := stdout.String()
+		if strings.Contains(out, "cannot determine repository root") || strings.Contains(out, `"error"`) {
+			t.Errorf("expected context to resolve without a git repo, got error output:\n%s", out)
+		}
+		if !strings.Contains(out, `"beads_dir"`) {
+			t.Errorf("expected beads_dir in context output:\n%s", out)
 		}
 	})
 }
--- a/cmd/bd/context_embedded_test.go
+++ b/cmd/bd/context_embedded_test.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
@@ -10,6 +11,8 @@ import (
 	"strings"
 	"sync"
 	"testing"
+
+	"github.com/steveyegge/beads/internal/configfile"
 )
 
 func TestEmbeddedContext(t *testing.T) {
@@ -51,26 +54,64 @@ func TestEmbeddedContext(t *testing.T) {
 	// the scope is not inside a git repository — git is a means of finding the
 	// repo root, not a hard requirement for this diagnostic (GH#4772).
 	t.Run("context_json_without_git_repo", func(t *testing.T) {
-		ngDir, _, _ := bdInit(t, bd, "--prefix", "ng")
-		if err := os.RemoveAll(filepath.Join(ngDir, ".git")); err != nil {
+		nonGitDir, _, _ := bdInit(t, bd, "--prefix", "ng")
+		if err := os.RemoveAll(filepath.Join(nonGitDir, ".git")); err != nil {
 			t.Fatalf("remove .git: %v", err)
 		}
 
 		cmd := exec.Command(bd, "context", "--json")
-		cmd.Dir = ngDir
-		cmd.Env = bdEnv(ngDir)
+		cmd.Dir = nonGitDir
+		cmd.Env = bdEnv(nonGitDir)
 		stdout, stderr, err := runCommandBuffers(t, cmd)
 		if err != nil {
 			t.Fatalf("bd context --json failed outside git repo: %v\nstdout:\n%s\nstderr:\n%s", err, stdout.String(), stderr.String())
 		}
 		out := stdout.String()
-		if strings.Contains(out, "cannot determine repository root") || strings.Contains(out, `"error"`) {
-			t.Errorf("expected context to resolve without a git repo, got error output:\n%s", out)
+
+		// Decode rather than probing for key names: ContextInfo's fields are
+		// not omitempty, so a wrong or empty beads_dir still emits the key.
+		var info ContextInfo
+		if jerr := json.Unmarshal([]byte(out), &info); jerr != nil {
+			t.Fatalf("decoding context JSON: %v\noutput:\n%s", jerr, out)
 		}
-		if !strings.Contains(out, `"beads_dir"`) {
-			t.Errorf("expected beads_dir in context output:\n%s", out)
-		}
+		assertNonGitContext(t, info, nonGitDir)
 	})
+}
+
+// assertNonGitContext holds `bd context` to the exact answer it must give for
+// a workspace at dir that is not inside a git repository: the workspace is
+// identified from .beads, the repo root falls back to the .beads parent, and
+// the backend still resolves from config.
+//
+// Paths are compared through EvalSymlinks because the test's temp dir is
+// reported canonically (/private/var/... on macOS) by the binary under test.
+func assertNonGitContext(t *testing.T, info ContextInfo, dir string) {
+	t.Helper()
+
+	resolve := func(p string) string {
+		if r, err := filepath.EvalSymlinks(p); err == nil {
+			return r
+		}
+		return p
+	}
+
+	wantBeadsDir := resolve(filepath.Join(dir, ".beads"))
+	if got := resolve(info.BeadsDir); got != wantBeadsDir {
+		t.Errorf("beads_dir = %q, want %q", info.BeadsDir, wantBeadsDir)
+	}
+	wantRepoRoot := resolve(dir)
+	if got := resolve(info.RepoRoot); got != wantRepoRoot {
+		t.Errorf("repo_root = %q, want %q (the .beads parent, since there is no git root)", info.RepoRoot, wantRepoRoot)
+	}
+	if info.Backend != configfile.BackendDolt {
+		t.Errorf("backend = %q, want %q — backend identity must still resolve from config", info.Backend, configfile.BackendDolt)
+	}
+	if info.BdVersion == "" {
+		t.Error("bd_version should be populated")
+	}
+	if info.IsWorktree {
+		t.Error("is_worktree should be false outside a git repository")
+	}
 }
 
 func TestEmbeddedContextConcurrent(t *testing.T) {
--- a/cmd/bd/context_proxied_integration_test.go
+++ b/cmd/bd/context_proxied_integration_test.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"encoding/json"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -56,6 +57,31 @@ func TestProxiedServerContext(t *testing.T) {
 		}
 		if info.ServerHost != "" || info.ServerPort != 0 {
 			t.Errorf("server host/port must be empty in proxied-server mode: host=%q port=%d", info.ServerHost, info.ServerPort)
+		}
+	})
+
+	// The proxied route must survive a workspace outside git for the same
+	// reason the direct route does — its context provider reads config files
+	// only (GH#4772). This gets its own project because it removes .git, which
+	// the shared one's parallel subtests still need.
+	t.Run("json_without_git_repo", func(t *testing.T) {
+		t.Parallel()
+		np := newSharedProxiedProject(t, bd, "ctxng")
+		if err := os.RemoveAll(filepath.Join(np.dir, ".git")); err != nil {
+			t.Fatalf("remove .git: %v", err)
+		}
+
+		stdout, stderr, err := bdProxiedRunBuffers(t, bd, np.dir, "context", "--json")
+		if err != nil {
+			t.Fatalf("bd context --json failed outside git repo: %v\nstdout:\n%s\nstderr:\n%s", err, stdout, stderr)
+		}
+		var info ContextInfo
+		if jerr := json.Unmarshal([]byte(stdout), &info); jerr != nil {
+			t.Fatalf("decoding context JSON: %v\noutput:\n%s", jerr, stdout)
+		}
+		assertNonGitContext(t, info, np.dir)
+		if info.DoltMode != configfile.DoltModeProxiedServer {
+			t.Errorf("dolt_mode = %q, want %q", info.DoltMode, configfile.DoltModeProxiedServer)
 		}
 	})
 
--- a/internal/beads/context.go
+++ b/internal/beads/context.go
@@ -99,6 +99,53 @@ func GetRepoContext() (*RepoContext, error) {
 	return repoCtx, repoCtxErr
 }
 
+// NoRepoRootError reports that a .beads directory was found and cleared the
+// SEC-003 safe-boundary check, but no git repository root could be determined
+// for it — i.e. the workspace simply is not inside a git repo.
+//
+// It is deliberately distinct from the other two failures buildRepoContext can
+// return ("no .beads directory found" and the unsafe-location rejection):
+// those mean bd does not know which workspace it is looking at, while this one
+// means it does. Only this failure is safe for a git-independent caller to
+// recover from, and only a typed error can say so — the unsafe-location
+// message embeds the offending path verbatim, so any substring test over the
+// message text is a path-controlled discriminator.
+type NoRepoRootError struct {
+	// BeadsDir is the resolved, boundary-checked .beads directory.
+	BeadsDir string
+	// Err is the underlying git failure.
+	Err error
+}
+
+func (e *NoRepoRootError) Error() string {
+	return fmt.Sprintf("cannot determine repository root: %v", e.Err)
+}
+
+func (e *NoRepoRootError) Unwrap() error { return e.Err }
+
+// GetRepoContextAllowingNoGit returns the repository context for callers that
+// only need to locate the workspace, not operate on git — `bd context` and the
+// context provider behind the proxied route, both of which read config files
+// and are documented to answer in degraded states.
+//
+// It behaves exactly like GetRepoContext except for NoRepoRootError, where it
+// synthesizes a context rooted at the .beads parent. Every other failure,
+// including the unsafe-location rejection, propagates untouched. Callers that
+// actually run git commands must keep using GetRepoContext, so that a missing
+// repository stays an error at the point where it matters (GH#4772).
+func GetRepoContextAllowingNoGit() (*RepoContext, error) {
+	rc, err := GetRepoContext()
+	var noRoot *NoRepoRootError
+	if !errors.As(err, &noRoot) {
+		return rc, err
+	}
+	return &RepoContext{
+		BeadsDir:    noRoot.BeadsDir,
+		RepoRoot:    filepath.Dir(noRoot.BeadsDir),
+		CWDRepoRoot: git.GetRepoRoot(), // "" outside a git repo
+	}, nil
+}
+
 // buildRepoContext constructs the RepoContext by resolving all paths.
 // This is called once per process via sync.Once.
 func buildRepoContext() (*RepoContext, error) {
@@ -133,7 +180,12 @@ func buildRepoContext() (*RepoContext, error) {
 		var err error
 		repoRoot, err = git.GetMainRepoRoot()
 		if err != nil {
-			return nil, fmt.Errorf("cannot determine repository root: %w", err)
+			// Typed so callers that can work without git select this exact
+			// failure with errors.As instead of matching the message text.
+			// beadsDir is carried along because it has already cleared steps
+			// 1 and 2 above (found, and inside the safe boundary), which is
+			// precisely what makes the no-git fallback safe to take.
+			return nil, &NoRepoRootError{BeadsDir: beadsDir, Err: err}
 		}
 	}
 
--- a/internal/beads/no_repo_root_test.go
+++ b/internal/beads/no_repo_root_test.go
@@ -1,0 +1,113 @@
+package beads
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/git"
+)
+
+// TestGetRepoContextAllowingNoGit_RecoversOutsideGitRepo verifies the
+// git-independent entry point resolves a workspace that is not inside a git
+// repository, rooting the context at the .beads parent (GH#4772).
+func TestGetRepoContextAllowingNoGit_RecoversOutsideGitRepo(t *testing.T) {
+	tmpDir := t.TempDir()
+	beadsDir := filepath.Join(tmpDir, ".beads")
+	if err := os.MkdirAll(beadsDir, 0o750); err != nil {
+		t.Fatalf("failed to create .beads dir: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(beadsDir, "beads.db"), []byte{}, 0o600); err != nil {
+		t.Fatalf("failed to create beads.db: %v", err)
+	}
+
+	origWD, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() {
+		_ = os.Chdir(origWD)
+		ResetCaches()
+		git.ResetCaches()
+	})
+	ResetCaches()
+	git.ResetCaches()
+	if err := os.Chdir(tmpDir); err != nil {
+		t.Fatal(err)
+	}
+
+	// Baseline: the git-requiring entry point still refuses, and does so with
+	// the typed error rather than a bare fmt.Errorf.
+	if _, err := GetRepoContext(); err == nil {
+		t.Fatal("GetRepoContext should still fail outside a git repository")
+	} else {
+		var noRoot *NoRepoRootError
+		if !errors.As(err, &noRoot) {
+			t.Fatalf("GetRepoContext error = %v, want a *NoRepoRootError", err)
+		}
+	}
+
+	rc, err := GetRepoContextAllowingNoGit()
+	if err != nil {
+		t.Fatalf("GetRepoContextAllowingNoGit failed outside a git repository: %v", err)
+	}
+	wantBeadsDir := resolveSymlinks(beadsDir)
+	if rc.BeadsDir != wantBeadsDir {
+		t.Errorf("BeadsDir = %q, want %q", rc.BeadsDir, wantBeadsDir)
+	}
+	if want := filepath.Dir(wantBeadsDir); rc.RepoRoot != want {
+		t.Errorf("RepoRoot = %q, want %q (the .beads parent)", rc.RepoRoot, want)
+	}
+	if rc.CWDRepoRoot != "" {
+		t.Errorf("CWDRepoRoot = %q, want empty outside a git repository", rc.CWDRepoRoot)
+	}
+}
+
+// TestNoRepoRootError_DoesNotMatchUnsafeLocation is the regression for the
+// discriminator itself.
+//
+// The no-git fallback must fire on exactly one failure: "a valid,
+// boundary-checked .beads was found, but there is no git root". The other
+// failure buildRepoContext can return — the SEC-003 unsafe-location rejection
+// — embeds the offending path verbatim in its message, so a substring test
+// over the message text is controlled by the path being rejected: a workspace
+// whose path contains the probe phrase would take the fallback and have its
+// unsafe-location error silently cleared.
+//
+// errors.As over a typed error cannot be spoofed that way, which is why the
+// selection is typed.
+func TestNoRepoRootError_DoesNotMatchUnsafeLocation(t *testing.T) {
+	const probe = "cannot determine repository root"
+
+	hostilePath := filepath.Join("/etc", probe, ".beads")
+	unsafeErr := fmt.Errorf("BEADS_DIR points to unsafe location: %s", hostilePath)
+
+	// The old discriminator: a path-controlled false positive.
+	if !strings.Contains(unsafeErr.Error(), probe) {
+		t.Fatalf("test setup no longer reproduces the substring collision: %v", unsafeErr)
+	}
+
+	// The current one: type-based, so the unsafe-location error is never
+	// mistaken for a recoverable missing-git-root.
+	var noRoot *NoRepoRootError
+	if errors.As(unsafeErr, &noRoot) {
+		t.Error("unsafe-location error must never select the no-git fallback")
+	}
+
+	realErr := &NoRepoRootError{BeadsDir: "/tmp/ws/.beads", Err: errors.New("not a git repository")}
+	if !errors.As(realErr, &noRoot) {
+		t.Error("NoRepoRootError must select the no-git fallback")
+	}
+	if noRoot.BeadsDir != "/tmp/ws/.beads" {
+		t.Errorf("BeadsDir = %q, want the boundary-checked dir carried by the error", noRoot.BeadsDir)
+	}
+	if !strings.Contains(realErr.Error(), probe) {
+		t.Errorf("error message changed: %q — the wording is user-facing", realErr.Error())
+	}
+	if !errors.Is(realErr, realErr.Err) {
+		t.Error("NoRepoRootError must unwrap to the underlying git failure")
+	}
+}
--- a/internal/storage/domain/fs/context.go
+++ b/internal/storage/domain/fs/context.go
@@ -26,8 +26,12 @@ func (r *contextRepositoryImpl) beadsDir(ctx context.Context) string {
 	return r.fsRepo.ResolveBeadsDirPath(ctx).BeadsDir
 }
 
+// RepoContext resolves the workspace paths behind `bd context`'s proxied
+// route. It tolerates the absence of a git repository for the same reason the
+// direct route does: this provider only reads config files, and a workspace
+// outside git is a supported degraded state, not a failure (GH#4772).
 func (r *contextRepositoryImpl) RepoContext(ctx context.Context) (domain.RepoPaths, error) {
-	rc, err := beads.GetRepoContext()
+	rc, err := beads.GetRepoContextAllowingNoGit()
 	if err != nil {
 		return domain.RepoPaths{}, err
 	}
@@ -41,10 +45,12 @@ func (r *contextRepositoryImpl) RepoContext(ctx context.Context) (domain.RepoPat
 }
 
 func (r *contextRepositoryImpl) Role(ctx context.Context) (string, bool, error) {
-	rc, err := beads.GetRepoContext()
+	rc, err := beads.GetRepoContextAllowingNoGit()
 	if err != nil {
 		return "", false, err
 	}
+	// Outside a git repo there is no `git config beads.role` to read, so
+	// rc.Role() reports "not configured" rather than failing the command.
 	role, ok := rc.Role()
 	return string(role), ok, nil
 }


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claim at the PR head). Verdict: **needs changes** — the no-Git fallback is wanted, but the trigger is unsafe and the proxied route misses it.

**Blocker — substring error-matching can suppress SEC-003 unsafe-location errors.** Verified at the PR head: the fallback triggers on `strings.Contains(err.Error(), "cannot determine repository root")`. The unsafe-location rejection embeds the offending *path verbatim* in its error text — so a project under a path containing that phrase matches the substring, takes the fallback, and `FindBeadsDir` (which checks project-file presence but not the safe-boundary rule) clears the error. Free-form error text as a domain discriminator is the root problem: introduce a typed/sentinel error for "no repository root" and select the fallback with `errors.Is`, or perform the no-Git fallback inside `buildRepoContext` after its boundary check.

**Should-fix — proxied mode still fails outside Git.** The proxied dispatch runs before the new fallback, and its context provider propagates `GetRepoContext()`'s error unchanged (root pre-run isn't the obstacle — `context` is classified no-store). Centralizing the typed fallback in repository-context resolution covers both modes at once.

**Should-fix — the regression only checks a JSON key exists.** Fields aren't `omitempty`, so an empty or wrong `beads_dir` passes. Decode `ContextInfo` and assert exact `beads_dir`, `repo_root`, and backend.

Nits: inline rationale duplicates the commit message; `ngDir` → `nonGitDir`.

harry-miller-trimble:
Thanks — all three addressed plus the nits, and the branch is rebased onto current `main`.

**Blocker — substring error-matching: fixed with a typed error.** `buildRepoContext` now returns `*beads.NoRepoRootError` for the missing-git-root case, and selection is `errors.As`. The error carries the `.beads` directory that has already cleared steps 1 and 2 (found, and inside the SEC-003 boundary) — which is both what makes recovering from *this* failure safe, and what removes the second `FindBeadsDir()` call the old fallback needed. Message text is unchanged.

`TestNoRepoRootError_DoesNotMatchUnsafeLocation` pins the discriminator directly: it builds the unsafe-location error with a hostile path (`/etc/cannot determine repository root/.beads`), asserts the old substring test still matches it — so the collision is documented, not just assumed — and asserts `errors.As` does not.

**Should-fix — proxied mode: fixed by centralizing.** The recovery lives in `beads.GetRepoContextAllowingNoGit`, and both routes now go through it: `context_cmd.go` directly, and the proxied provider via `domain/fs` `contextRepositoryImpl.RepoContext` / `.Role`, which previously propagated `GetRepoContext()`'s error unchanged. Everything that actually runs git commands still calls `GetRepoContext`, so a missing repository stays an error at the point where it matters rather than becoming a silently synthesized context process-wide.

**Should-fix — the regression only checked a key exists: fixed.** Both the embedded and the new proxied subtest decode `ContextInfo` and assert exact values via a shared `assertNonGitContext` helper: `beads_dir` is the workspace's `.beads`, `repo_root` is its parent, `backend` is still `dolt` (from config), `bd_version` is populated, `is_worktree` is false. Paths go through `EvalSymlinks` on both sides since the binary reports the canonical `/private/var/...` form on macOS. The proxied case builds its own project rather than reusing the shared one, because it removes `.git` and the shared project's other subtests run in parallel.

**Nits:** `ngDir` → `nonGitDir`; the inline rationale that duplicated the commit message is gone — what remains is the doc comment on `NoRepoRootError`/`GetRepoContextAllowingNoGit` explaining which of the three `buildRepoContext` failures is recoverable and why.

Tests run locally: `./internal/beads/`, `./internal/storage/domain/...`, and `TestEmbeddedContext` with `BEADS_TEST_EMBEDDED_DOLT=1` all pass. `TestProxiedServerContext` skips here (no shared proxied server available locally) but compiles; CI covers it.

maphew (approved):
Re-review: all three points plus both nits are genuinely addressed, and the blocker fix took the stronger of the two shapes I sketched. Approving - this supersedes my earlier CHANGES_REQUESTED.

- **Blocker** (substring-matched fallback spoofable by a path containing the error phrase): now a typed `*NoRepoRootError`, emitted at the exact `GetRepoMainRoot` failure site and only after `.beads` discovery and the SEC-003 boundary check, selected via `errors.As`, with a regression test pinning the original substring collision. Verified in source that the error site sits after boundary validation, and that the synthesized context's zero-valued `IsRedirected`/`IsWorktree` are correct in the only reachable scenario.
- **Proxied route**: both entry points moved onto `GetRepoContextAllowingNoGit` via the domain/fs provider (`contextRepositoryImpl.RepoContext`/`.Role`), which is the provider behind the proxied path - so the fix is centralized rather than duplicated, as asked.
- **Test rigor**: `assertNonGitContext` now decodes `ContextInfo` and asserts exact `beads_dir`, `repo_root`, backend, non-empty `bd_version`, and `is_worktree=false`. The new `TestProxiedServerContext` subtest runs in CI's proxied shards, which are green.
- Both nits done (duplicated inline rationale replaced by doc comments that actually explain which of the three failures is recoverable; `ngDir` renamed).

One non-gating observation worth a one-line PR-body edit: the domain/fs provider also serves `bd serve` (`cmd/bd/serve.go:266`), so serve in a non-git directory with `.beads` now proceeds instead of failing. That is consistent with the config-only identity-resolution intent and probably an improvement, but the body's claim that other commands' behavior is untouched is now stale.

CI fully green at head, mergeable. Process note: I stepped down as maintainer on 2026-08-10 and can no longer land this. @csells @julianknutsen - ready for merge from my side.

harry-miller-trimble:
Thanks for the re-review — and for the `bd serve` catch. Confirmed: `cmd/bd/serve.go:266` resolves its workspace through `contextinfo.NewContextProvider(...).ContextUseCase()`, which wires `domainfs.NewContextRepository(...)` (`internal/storage/contextinfo/provider.go:24-31`) — the exact provider this PR moved onto `GetRepoContextAllowingNoGit`. So `bd serve` in a non-git directory with a valid `.beads/` now proceeds instead of failing at context resolution.

I've replaced the stale "other commands' behavior is untouched" line in the PR body with a scoped bullet that names `bd serve` explicitly and says why the change is consistent with the config-only identity resolution intent, rather than silently in scope. No code change — the behavior is the one you described and I agree it's the improvement, it just needed to be stated.

_claude-code-opus-5-unknown-reasoning on behalf of Harry Miller_

bee-ghosttrack (changes requested):
Round 1. I reviewed this in a throwaway worktree at `ad50827f16`, built it, ran the tests it touches, and mutation-probed every one of them. The core fix is right and the scoping argument holds up under attack — the typed error really does keep the SEC-003 rejection out of the fallback, and behavior inside a git repo is byte-identical. Two things block: a wrong answer on the `BEADS_DIR`-outside-git path, and one new test that does not guard the thing its own doc comment says it guards.

## Merge cleanliness / staleness

The PR is from 2026-07-14 and sits 119 commits behind `main`, so I checked the ground under it rather than trusting `MERGEABLE`.

```
git merge-tree --write-tree origin/main ad50827f1   → rc=0 (tree 1808a21e3)
git diff --check origin/main..ad50827f1             → rc=0
git rev-list --count ad50827f1..origin/main         → 119
```

- The resolution path it patches still exists at `main` head, unmoved: `internal/beads/context.go` `buildRepoContext()` still does find-`.beads` → SEC-003 boundary → redirect/external → `git.GetMainRepoRoot()`, and the single failure site the PR types is still the one at step 4. `git diff origin/main..head` over the three non-test files shows only the PR's own hunks — no drift underneath it.
- The PR body names two consumers of the `domain/fs` context provider (`cmd/bd/context_proxied_server.go`, and the `bd serve` knock-on at `cmd/bd/serve.go:266`). I re-enumerated against **current main**, not the PR's base:

```
grep -rn "NewContextProvider" --include=*.go (origin/main) →
  cmd/bd/serve.go:266
  cmd/bd/context_proxied_server.go:24
```

Still exactly those two. The disclosure has not gone stale in 119 commits. Good.

- `go build ./...` rc=0 · `go vet ./internal/beads/... ./internal/storage/domain/fs/... ./cmd/bd/...` rc=0 · `gofmt -l` on all six files: clean.

## What I could not break

Worth recording, because these were my three main attack lines and they all failed:

**1. Callers inside a git repo.** I built `bd` from `origin/main` and from the head and diffed `bd context --json` in three shapes: a plain git repo with `.beads` at the root; a `git worktree` whose main repo holds the `.beads`; and a nested git repo whose `.beads` lives **above** the git root (`$T/proj/.beads` with the repo at `$T/proj/inner`). All three: identical output except `bd_version` (main is 1.2.2, the PR head predates the bump). `GetMainRepoRoot()` succeeds in all of them, so `errors.As` never matches and the old path runs untouched. The worktree case still reports `repo_root` = main repo, `is_worktree: true`.

**2. The SEC-003 rejection leaking into the fallback.** Not theoretical — I hit it by accident. With `.beads` under `/private/tmp/...`, head still refuses:

```
{"error": "cannot resolve repo context: BEADS_DIR points to unsafe location: /private/tmp/.../proj/.beads"}
```

Identical on `main` and on head. The unsafe-location error propagates through `GetRepoContextAllowingNoGit` unchanged, exactly as `internal/beads/context.go:130-146` claims.

**3. Unbounded walk-up.** The fallback inherits `FindBeadsDir`'s ancestor walk, and outside a git repo `walkBoundary` is `""` (`internal/beads/beads.go:798-808`), so the walk does climb to `/`. But it is bounded in practice: `hasBeadsProjectFiles` rejects a registry-only `~/.beads`, the SEC-003 boundary rejects system paths, and the loop stops at `/`. Verified: from `$T/proj/sub/deep/er` (no git anywhere) head resolves `repo_root=$T/proj` from the ancestor `.beads` — which is the same answer you'd get from a subdirectory of a git repo, so it is the consistent one. Not a defect. It is worth one line in the PR body, though, since `bd serve` now inherits it: `bd serve` in a scratch directory nested under a project will now bind that project's workspace instead of failing.

**4. `-C`.** There is no global `-C`/`--directory` flag on `bd`; scope comes from CWD and `BEADS_DIR`. Nothing to interact with.

## Mutation probes

Every test the PR adds or changes, with the fix it guards reverted:

| Probe | Mutation | Test | Result |
|---|---|---|---|
| A | `if true \|\| !errors.As(...)` — fallback disabled | `TestGetRepoContextAllowingNoGit_RecoversOutsideGitRepo` | **FAIL** (`cannot determine repository root: not a git repository: exit status 128`) ✅ |
| C | `buildRepoContext` back to `fmt.Errorf` (untyped) | same test's baseline assertion | **FAIL** (`want a *NoRepoRootError`) ✅ |
| D | `context_cmd.go:78` back to `GetRepoContext()` | `TestEmbeddedContext/context_json_without_git_repo` | **FAIL**, and it reproduces #4772's exact error text ✅ |
| E | `fs/context.go:34` back to `GetRepoContext()` | `TestProxiedServerContext/json_without_git_repo` | **FAIL** (`cannot resolve context: cannot determine repository root`) ✅ |
| **B** | typed selection → `strings.Contains(err.Error(), "cannot determine repository root")` | `TestNoRepoRootError_DoesNotMatchUnsafeLocation` | **PASS — vacuous** ❌ |

Green runs on the unmutated head:

```
go test ./internal/beads -run 'TestGetRepoContextAllowingNoGit_RecoversOutsideGitRepo|TestNoRepoRootError_DoesNotMatchUnsafeLocation' -count=1   rc=0  (2 PASS)
CGO_ENABLED=1 BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go ./cmd/bd -run 'TestEmbeddedContext$' -count=1   rc=0  (3 PASS, 227s)
CGO_ENABLED=1 BEADS_TEST_EMBEDDED_DOLT=1 BEADS_TEST_PROXIED_SERVER=1 go test -tags gms_pure_go ./cmd/bd -run 'TestProxiedServerContext$' -count=1   rc=0  (4 PASS, 111s)
```

Working tree `git status --porcelain` clean after every probe.

## BLOCKING

### B1 — `is_redirected` and `role` come out wrong on the `BEADS_DIR`-outside-git path

`internal/beads/context.go:139-145` synthesizes the context with `IsRedirected` left at its zero value. But `BEADS_DIR` pointing at another repo's `.beads` from a non-git CWD reaches this branch: `isExternalBeadsDir` calls `git.GetGitCommonDir()` on the **CWD**, which fails outside git, so `isExternal` stays false and we fall through to the `GetMainRepoRoot` error site. The result is a `bd context` that denies it is redirected.

Same `BEADS_DIR`, two CWDs, `bd` built from each side:

```
# CWD = a different git repo — main (baseline, unchanged by this PR)
is_redirected: true
repo_root:     $T/gitrepo
role:          contributor

# CWD = a non-git dir — main
{"error": "cannot resolve repo context: cannot determine repository root: ..."}

# CWD = a non-git dir — this PR
is_redirected: false
repo_root:     $T/gitrepo          ← right
role:          (absent)            ← wrong
```

That contradicts the field's own contract two screens up in the file it edits — `internal/beads/context.go:57-59`, "IsRedirected is true if BeadsDir resolves to a different repository than CWD. This covers explicit BEADS_DIR usage and redirect files" — and `RepoContext.Role()` at `internal/beads/context.go:623-626`, "BEADS_DIR implies contributor (external repo mode)". `bd context --json` is being sold in this very PR as a health/identity probe for consumers; shipping `is_redirected: false` for a workspace that *is* a `BEADS_DIR` redirect makes the probe lie about identity, which is a worse failure mode than the error it replaces. It rides the proxied route too, via `fs/context.go` `RepoContext`/`Role`.

I checked the blast radius before calling this blocking: `IsContributor()`/`IsMaintainer()` have no non-test callers on main, so nothing gates on it today — this is a wrong *answer*, not a wrong *action*. It is still the answer the command exists to give.

Fix is small: in the fallback, set `IsRedirected` when `BEADS_DIR` names the resolved dir (the same condition `FindBeadsDir` step 1 took), and assert it in `TestGetRepoContextAllowingNoGit_RecoversOutsideGitRepo` — plus the negative case, `is_redirected: false` when the `.beads` was found by the CWD walk. Right now no test in the PR asserts `IsRedirected` at all on this path.

### B2 — `TestNoRepoRootError_DoesNotMatchUnsafeLocation` does not test the discriminator

`internal/beads/no_repo_root_test.go:64-113`. Its doc comment says it "is the regression for the discriminator itself" and that "errors.As over a typed error cannot be spoofed that way, which is why the selection is typed". It never calls the selection. Every assertion is over two hand-built error values and the stdlib's `errors.As` — `errors.As(fmt.Errorf(...), **NoRepoRootError)` is false as a property of Go, no matter what `GetRepoContextAllowingNoGit` does.

Probe B: I replaced the typed selection in `GetRepoContextAllowingNoGit` with the exact substring test the comment warns against —

```go
if err == nil || !strings.Contains(err.Error(), "cannot determine repository root") {
    return rc, err
}
```

— i.e. reintroduced the path-controlled false positive the test is named after. Both new `internal/beads` tests stayed green (rc=0). The regression this test claims to hold is not held.

It is not worthless — it does pin `Unwrap`, the user-facing message wording, and that `BeadsDir` is carried — but the headline assertion is missing. Cheapest honest fix: pull the two-line selection out of `GetRepoContextAllowingNoGit` into a helper (`func recoverNoGit(rc *RepoContext, err error) (*RepoContext, error)`), have the exported function call it, and drive the helper directly with `fmt.Errorf("BEADS_DIR points to unsafe location: /etc/cannot determine repository root/.beads")`, asserting the error comes back out. That version fails under probe B.

## NITS (non-blocking)

**N1 — `NoRepoRootError` fires on *any* `git rev-parse` failure, not just "not a repo".** `git.GetMainRepoRoot()` bottoms out in one `git rev-parse --git-dir --git-common-dir --show-toplevel` (`internal/git/gitdir.go:32-36`); every failure of that command becomes `gitCtx.err`, and every one of those now takes the fallback. Not hypothetical — a directory with a dangling gitfile (`.git` containing `gitdir: /nonexistent/nope`) plus a valid `.beads`:

```
main:  {"error": "cannot resolve repo context: cannot determine repository root: not a git repository: exit status 128"}
head:  repo_root = $T/broken, is_worktree=false, backend=dolt   (no mention of git at all)
```

Same for a broken `safe.directory`/dubious-ownership repo, or `git` missing from `PATH`. The type's doc (`context.go:99-108`) says it means "the workspace simply is not inside a git repo"; it actually means "git could not tell us". For `bd context` — the command whose whole pitch is diagnostics in degraded states — that is the one fact worth surfacing and it is now silently swallowed. JSON consumers can still infer it (`cwd_repo_root` is `omitempty` and comes back absent), but `printContextText` prints nothing: `cwd repo` only prints when non-empty *and* different. Suggest either reword the type doc, or have `printContextText` say `git: unavailable` when `CWDRepoRoot == ""`.

**N2 — the new comment at `internal/storage/domain/fs/context.go:51-52` is factually wrong.** It says "Outside a git repo there is no `git config beads.role` to read, so rc.Role() reports 'not configured'". `git config --get` reads global and system config outside a repo. With `beads.role = maintainer` in `~/.gitconfig`, head in a non-git dir reports `"role": "maintainer"`. The behavior is fine; the comment explains it wrongly, and a wrong comment on a security-adjacent identity path is worse than none.

**N3 — no CHANGELOG entry, no `engdocs/REPO_CONTEXT.md` update.** `## [Unreleased]` on main carries per-change entries with issue links; this is a user-visible behavior change plus a `bd serve` knock-on and gets none — an `### Fixed` entry citing #4772 is the house style. And `engdocs/REPO_CONTEXT.md` is the reference for exactly this module: it documents `GetRepoContext()`'s boundary validation and the `IsRedirected` field semantics that B1 bends, and gains neither the new entry point nor the no-git case.

**N4 — trivial.** `GetRepoContextAllowingNoGit` builds a fresh `RepoContext` on every call where `GetRepoContext` returns the process-cached pointer; `fs/context.go` calls it twice per command (`RepoContext` + `Role`). Two allocations and two `git.GetRepoRoot()` calls (itself cached). Mention only.

## Summary

The mechanism is right and the fences around it hold — typed selection, unsafe-location still refused, no change inside a git repo, both route tests non-vacuous. Fix B1 (`IsRedirected` on the `BEADS_DIR` path, with assertions) and B2 (make the discriminator test drive the discriminator) and I'll re-review; N1–N4 are yours to take or leave.

harry-miller-trimble:
Thanks — this is the review I'd want on a security-adjacent identity path, and both blockers were right. Fixed, all four nits taken, rebased onto current `main`. Pushed as `8421b8d7a`.

## B1 — `is_redirected` and `role` on the `BEADS_DIR`-outside-git path

Confirmed. Your reasoning about *why* it happens is the load-bearing part and I'd missed it: `isExternalBeadsDir` compares git **common dirs**, and the CWD side of that comparison cannot be computed without a repository — which is precisely the state this branch exists to serve. So the normal path's answer isn't merely unset here, it's unavailable, and zero-valuing the field quietly asserts the opposite of what the environment plainly says.

An explicit `BEADS_DIR` is the only evidence left, and it's the same signal `FindBeadsDir`'s step 1 used to choose the directory, so the fallback mirrors that condition:

```go
IsRedirected: beadsDirNamedByEnv(noRoot.BeadsDir),
```

`beadsDirNamedByEnv` resolves `BEADS_DIR` through the same `canonicalizeBeadsDirPath` + `FollowRedirect` the finder used, so it says "this directory was named" rather than "some `BEADS_DIR` exists". A `.beads` found by the working-directory walk still reports false. `IsWorktree` stays false with a comment saying why, rather than by omission.

`TestRecoverNoGit_RedirectProvenance` covers all three cases you asked for and one more — named by `BEADS_DIR` (asserting `Role()` returns `Contributor`, not just the flag), found by the CWD walk, and a `BEADS_DIR` that names a *different* directory than the one resolved. `TestGetRepoContextAllowingNoGit_RecoversOutsideGitRepo` now pins the negative on the real path too. Reverting the one line:

```
--- FAIL: TestRecoverNoGit_RedirectProvenance/named_by_BEADS_DIR_is_redirected
    IsRedirected = false for a workspace named by BEADS_DIR
    Role() = ("", false), want ("contributor", true) — BEADS_DIR implies contributor
```

Your blast-radius check matches mine: `IsContributor()`/`IsMaintainer()` have no non-test callers, so this was a wrong answer rather than a wrong action — but as you say, it's the answer the command exists to give.

## B2 — the discriminator test didn't drive the discriminator

You're right, and the probe is the proof: asking `errors.As` about two hand-built values tests the standard library, and passes whatever this package actually uses. I took your suggested fix exactly — `recoverNoGit(rc, err)` split out of `GetRepoContextAllowingNoGit`, with the test driving it directly. It now asserts that all three non-recoverable outcomes come back **unchanged** (unsafe-location, no-`.beads`, and a successful context), not just that `errors.As` says no.

Under your probe B it fails, and usefully — it names the context the substring version wrongly recovered:

```
--- FAIL: TestNoRepoRootError_DoesNotMatchUnsafeLocation
    unsafe-location error was recovered into a context (&{BeadsDir:/Users/…/beads/.beads
    RepoRoot:/Users/…/beads CWDRepoRoot:/Users/…/pr-4792-context-nogit IsRedirected:false
    IsWorktree:false}); it must propagate
```

## N1 — `NoRepoRootError` fires on any `git rev-parse` failure

Taken, both halves. The type doc no longer says "the workspace simply is not inside a git repo"; it says git could not tell us a root, names the cases you listed (dangling gitfile, dubious-ownership, no git on `PATH`), and says why recovering is equally safe in all of them while a caller reporting on the workspace shouldn't claim to know which it was.

And `printContextText` now says so. You spotted that `cwd repo` printed only when non-empty *and* different, so the degraded state vanished from text output entirely while `--json` consumers could still infer it. It now prints `cwd repo: git: unavailable` when `CWDRepoRoot` is empty — the text counterpart of the omitted field.

## N2 — the wrong comment on `fs/context.go`

Fixed, and I confirmed your claim by accident while running probe B: `Role()` came back `"maintainer"` from my own `~/.gitconfig` with no repository anywhere. `git config --get` reads global and system config, exactly as you said. The comment now says `Role()` still answers outside a repo — `BEADS_DIR` resolves to `Contributor` directly, otherwise the global/system files are read — and reports "not configured" only when none of them carry a role. Behavior unchanged; the comment was the bug.

## N3 — CHANGELOG and `engdocs/REPO_CONTEXT.md`

Both added. The CHANGELOG entry goes under `### Fixed` citing #4772. `REPO_CONTEXT.md` gains a "Resolving Without Git" section: why the no-git failure is the one that's safe to recover from (the workspace is *known*, not unknown), why the selection is typed rather than by message text, and a table of what the synthesized context can and cannot know — which is where the `IsRedirected` semantics B1 bends now live in prose.

## N4 — the allocation

Noted and left. Two allocations and two cached `git.GetRepoRoot()` calls on a config-reading path isn't worth a cache whose invalidation would be a new thing to get wrong, and `GetRepoContext()`'s own `sync.Once` already absorbs the expensive half. Happy to revisit if it ever shows up in a profile.

## Also from your review

The ancestor-walk observation (N3 under "what I could not break") is now a line in the PR body, as you suggested — including that it's the same answer a subdirectory inside a git repo already gets, so it's the consistent behavior rather than a new one, and that `bd serve` inherits it.

## Local gates on the rebased head

`gofmt -l`, `go build ./...`, `go vet ./internal/beads/... ./internal/storage/domain/fs/... ./cmd/bd/...` clean; `go test ./internal/beads/ ./internal/storage/domain/...` green; `BEADS_TEST_EMBEDDED_DOLT=1 go test ./cmd/bd/ -run TestEmbeddedContext` green; `generate-cli-docs.sh --check` PASS; `go test ./test/docsync/` green.

One thing I did **not** fix because it isn't mine: `TestPrepareSelectedCommandContext_RebindsTargetConfig` and `..._DoesNotMergeCallerConfigForUnsetKeys` fail on my machine with `actor = "harry", want "target-actor"`. I verified they fail identically with my changes stashed, so they predate this branch — my own `~/.beads/config.yaml` leaks into them. Amusingly that's the same legacy-config shape behind #4795's bypass. Worth its own issue if it isn't already known; say the word and I'll file it.

_claude-code-opus-5-unknown-reasoning on behalf of Harry Miller_

bee-ghosttrack (changes requested):
Round 2 from me (bee). Re-reviewed `8421b8d7aa` in a throwaway worktree off the PR head, rebased onto current `main` (0 behind, 3 ahead), merge-tree clean, `go build ./...` / `go vet` / `gofmt -l` clean. Both Round-1 blockers are fixed **at the unit level and the tests now hold under mutation** — but the B1 fix comes out wrong in the built binary, for a reason this same author just fixed in #4795. One blocker, then approve.

## What I ran

| Command | Result |
|---|---|
| `go test ./internal/beads/ -count=1` (whole package) | rc=0, 8.0s |
| `go test ./internal/storage/domain/fs/ -count=1` | rc=0 |
| `CGO_ENABLED=1 BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go ./cmd/bd -run 'TestEmbeddedContext$' -count=1 -v` | rc=0 — 3/3 incl. `context_json_without_git_repo`, 32s |
| same + `BEADS_TEST_PROXIED_SERVER=1`, `-run 'TestProxiedServerContext$'` | rc=0 — 4/4 incl. `json_without_git_repo`, 27s |

## Mutation probes (tree restored after each)

| Probe | Result |
|---|---|
| **P1** — `IsRedirected: false` (B1 reverted) | rc=1 — `TestRecoverNoGit_RedirectProvenance/named_by_BEADS_DIR_is_redirected` FAILS ✅ |
| **P2** — `recoverNoGit` back to `strings.Contains(err.Error(), "cannot determine repository root")` (Round-1 probe B) | rc=1 — `TestNoRepoRootError_DoesNotMatchUnsafeLocation` FAILS with *"unsafe-location error was recovered into a context … it must propagate"*, and `TestRecoverNoGit_RedirectProvenance` + `TestGetRepoContextAllowingNoGit_RecoversOutsideGitRepo` fail alongside ✅ |
| **P3** — delete the new `cwd repo: git: unavailable` branch in `printContextText` | rc=0 over every `-run Context` test — the N1 text line is **unpinned** (nit, below) |

So the B2 test now drives the discriminator — good, exactly the shape I asked for.

## BLOCKING — B1 is wrong in the other direction once the real command runs

I built `bd` from the head and ran it from a SEC-003-safe location under `$HOME` (anything under `/private/tmp` is still refused, as it should be):

```
# BEADS_DIR=$T/gitrepo/.beads, CWD = $T/nogit (non-git)     — the Round-1 case
  "is_redirected": true,   "role": "contributor"            ← right

# NO BEADS_DIR in the environment, CWD = $T/proj/sub, .beads found by the walk at $T/proj
  "is_redirected": true,   "role": "contributor"            ← WRONG
  text:  cwd repo: git: unavailable / redirected: yes
```

The second run had `BEADS_DIR` unset (`env -u BEADS_DIR`). `beadsDirNamedByEnv` reads `os.Getenv("BEADS_DIR")` **live** — and by the time `recoverNoGit` runs, `bd` has exported one for itself: `context_cmd.go:75` calls `prepareSelectedNoDBContext(selected)` immediately before `GetRepoContextAllowingNoGit()`, and that does `os.Setenv("BEADS_DIR", …)` with the directory discovery just found. So on the no-git path every walk-found workspace now reports itself as a redirect and claims `Contributor`. The proxied route gets it the same way through the root pre-run.

This is the identical provenance bug you fixed an hour ago in #4795 (`beadsDirFromCaller`): a `BEADS_DIR` in the process env is not evidence the caller named one. The unit test passes because `TestRecoverNoGit_RedirectProvenance` drives `recoverNoGit` directly with the env you set, never through the pre-run — and neither route test asserts `is_redirected` at all (`assertNonGitContext` checks `beads_dir`, `repo_root`, backend, version, `is_worktree`). Same gap as Round 1's B1, one layer up.

Fix, in preference order:

1. **Have the finder record provenance.** `FindBeadsDir` already knows whether step 1 chose the directory; carry that bit (or the step-1 result) so `recoverNoGit` asks "was this dir named by the caller" rather than "does the env have a value". Or, cheaper and symmetrical with #4795: an `internal/beads` package-init snapshot of the inherited `BEADS_DIR`, and compare against *that*. Either way the live env is the wrong source.
2. **Pin it end to end:** `assertNonGitContext` asserts `is_redirected == false` for the walk-found case (both routes already have one), plus one route subtest that sets `BEADS_DIR` in the subprocess env and asserts `true` + `"role": "contributor"`. Under the current head the first assertion fails, which is the regression this needs.

I checked blast radius the same way as Round 1: `IsContributor()`/`IsMaintainer()` still have no non-test callers, so this is a wrong answer, not a wrong action — but `bd context --json` is being shipped here precisely as an identity probe, and it now says "redirected, contributor" for every non-git workspace it finds by walking.

## Nits (non-blocking)

- **N1 — `git: unavailable` is unpinned** (probe P3). One assertion on the text route in `TestEmbeddedContext/context_json_without_git_repo`'s sibling, or a `printContextText` unit test with `CWDRepoRoot: ""`, would hold it.
- **N2 — `TestPrepareSelectedCommandContext_RebindsTargetConfig` / `_DoesNotMergeCallerConfigForUnsetKeys`:** they fail on my box too with the same `actor` leak, with your changes and without. Yes, please file it — pre-existing, and the legacy `~/.beads/config.yaml` shape behind #4795 is exactly what leaks into them.

Everything else from Round 1 is taken as asked — the `REPO_CONTEXT.md` section and the `NoRepoRootError` doc are both accurate now, and the `fs/context.go` comment is right. Fix the provenance and add the end-to-end `is_redirected` assertions and this is an approve from me.

harry-miller-trimble:
CI note: the one failure, `Test (storage domain + uow)`, is the test Dolt server dying mid-job, not this branch. A maintainer rerun should clear it — I have no rerun lever on this repo.

The server dropped at 14:39:47 and everything after it failed to connect:

```
[mysql] 2026/09/11 14:39:47 packets.go:58 unexpected EOF
--- PASS: TestEnginePushWithShouldPush (11.91s)
--- FAIL: TestEnginePushWithShouldPushCanInspectLabels (0.00s)
    engine_test.go:1309: Failed to create dolt store: Dolt server unreachable at
    127.0.0.1:32772: dial tcp 127.0.0.1:32772: connect: connection refused
```

**37** tests fail with that same `connection refused`, every one in `0.00s` — a cascade from one dead container, not 37 independent breaks.

Why it isn't attributable here:

- The failures are all in `github.com/steveyegge/beads/internal/tracker`. This branch touches eight files and none of them is in that package (`internal/beads`, `internal/storage/domain/fs`, `cmd/bd/context_cmd.go`, two test files, CHANGELOG, engdocs). `internal/tracker` does not import `internal/beads` at all, so there is no path from the changed code to it.
- My packages in that same job passed *before* the server died: `internal/storage/domain/fs` in 1.370s, `internal/storage/domain` 1.013s, `internal/storage/domain/db` 51.890s, `internal/storage/uow` 332.968s.
- `internal/tracker` itself was healthy until the EOF — `TestEnginePushWithShouldPush` passed in 11.91s immediately before the first failure.
- The only change this round in `internal/storage/domain/fs` is the N2 comment correction; no executable change.

Everything else on this head is green, and #4795 — pushed minutes earlier off the same base — is fully green including this job.

_claude-code-opus-5-unknown-reasoning on behalf of Harry Miller_

harry-miller-trimble:
You're right, and the sting is that this is the same bug I fixed in #4795 an hour before writing it. Fixed, both nits taken, N2 filed. Pushed as `ef69b571d`, rebased onto current `main`.

## BLOCKING — provenance, one layer up

Reproduced against a built binary before touching anything, from a SEC-003-safe location under `$HOME`:

```
$ cd $T/proj/sub && env -u BEADS_DIR bd context --json
  "is_redirected": true,
  "role": "contributor"
```

No `BEADS_DIR` anywhere in the caller's environment. `context_cmd.go:75` calls `prepareSelectedNoDBContext(selected)` immediately before `GetRepoContextAllowingNoGit()`, and that ends in `os.Setenv("BEADS_DIR", …)` with whatever discovery just found — so `beadsDirNamedByEnv`'s live read saw bd's own answer. Worse than the Round-1 bug, as you say: it's the common case rather than the unusual one.

`internal/beads` now takes the same package-init snapshot `package main` does, and `beadsDirNamedByCaller` compares against that. I went with your cheaper option deliberately: having `FindBeadsDir` record provenance doesn't actually fix it, because `FindBeadsDir` reads the live env too — by the time it runs, step 1 *would* have chosen bd's exported value, so the finder would report "named by the caller" just as wrongly. The snapshot is the only place the distinction still exists.

One thing that fell out while checking the edges, now a comment rather than an assumption: a redirect **file** cannot be missed by this branch. `buildRepoContext` checks `GetRedirectInfo()` at step 3, and a redirect sends it down the `isExternal` branch that never asks git for a root — so by construction the only redirect `recoverNoGit` can be looking at is a caller-named `BEADS_DIR`. That's what makes the single signal sufficient rather than merely convenient.

**Coverage, both halves, both routes.** Your diagnosis of why the unit tests passed was exact — they drive `recoverNoGit` with an env they set themselves, which is precisely the distinction that cannot be tested in process. So:

- `assertNonGitContext`, shared by the embedded and proxied route tests, now asserts `is_redirected == false` for the walk-found case. Reverting the fix: `is_redirected should be false for a .beads found by the working-directory walk`.
- a new embedded subtest runs the binary from an unrelated directory with `BEADS_DIR` supplied in the **subprocess** environment — the only place a caller-supplied value is distinguishable from a bd-exported one — and asserts `is_redirected == true` and `role == "contributor"`.

## N1 — `git: unavailable` unpinned

Your probe P3 was right. `TestPrintContextText_*` now pins all three branches: an empty CWD repo root reports git unavailable, a matching one prints nothing, and a differing one prints the path and must *not* say unavailable. Deleting the branch now fails with the output inline. Kept as an in-process stdout capture rather than another subprocess, given `cmd/bd` has no runtime budget to spare — #6253.

## N2 — filed as #6560, and I owe you a correction

I told you this was a legacy `~/.beads/config.yaml` leaking in, you reasonably repeated it, and it's wrong — I have no such file. It's `BEADS_ACTOR` in the environment. `resolveConfiguredActor` (`main.go:688`) checks the env *ahead* of `config.GetString("actor")` deliberately, as the fix for #4645, so with `BEADS_ACTOR` set no `config.yaml` actor value can ever win and the assertion is unsatisfiable:

```
$ go test ./cmd/bd/ -run TestPrepareSelectedCommandContext -count=1
FAIL   (actor = "harry")
$ env -u BEADS_ACTOR go test ./cmd/bd/ -run TestPrepareSelectedCommandContext -count=1
ok     1.620s
```

The issue suggests clearing both `BEADS_ACTOR` and `BD_ACTOR` in the two tests' setup — the deprecated alias reaches the same key through `AutomaticEnv` and would take over once the primary is empty. Sorry for sending you down the wrong path on that one; the symptom looked like config precedence and I didn't check my own environment before naming a cause.

## Local gates on the rebased head

`gofmt -l`, `go build ./...`, `go vet ./internal/beads/... ./cmd/bd/...` clean; `go test ./internal/beads/ ./internal/storage/domain/fs/` green; `BEADS_TEST_EMBEDDED_DOLT=1 go test ./cmd/bd/ -run 'TestEmbeddedContext$|TestPrintContextText'` green, 4/4 embedded subtests including the new one.

_claude-code-opus-5-unknown-reasoning on behalf of Harry Miller_

bee-ghosttrack (approved):
**Verdict: APPROVE — round 3** (bee; head `ef69b571d`, reviewed in a throwaway worktree with a binary built from it).

The round-2 blocker is fixed, and it is fixed in the built binary, which is where it was broken. From a SEC-003-safe directory under `$HOME`, workspace made by `bd init` with `.git` removed:

```
# env -u BEADS_DIR, CWD = $T/proj/sub, .beads found by the walk at $T/proj
  "is_redirected": false            (no "role" key at all)      text: cwd repo: git: unavailable

# BEADS_DIR=$T/proj/.beads, CWD = $T/elsewhere (non-git)
  "is_redirected": true, "role": "contributor"                  text: redirected: yes / role: contributor
```

## Probes (tree restored after each)

| Probe | Result |
|---|---|
| **P1** — `beadsDirNamedByCaller` back to a live `os.Getenv` | `TestEmbeddedContext/context_json_without_git_repo` FAILS at `context_embedded_test.go:77` ("is_redirected should be false for a .beads found by the working-directory walk"). The assertion lives in the shared `assertNonGitContext`, so the proxied route holds it too ✅ |
| **P2** — delete the `git: unavailable` branch in `printContextText` | `TestPrintContextText_SaysWhenGitCouldNotAnswer` FAILS ✅ — N1 is pinned |

Gates on this head: `go build -tags gms_pure_go ./...`, `go vet ./internal/beads/... ./cmd/bd/...`, `gofmt -l` on the four files, `go test -count=1 ./internal/beads/` — all rc=0. `BEADS_TEST_EMBEDDED_DOLT=1 … -run 'TestEmbeddedContext$|TestPrintContextText'` rc=0 (4 subtests + 3 text cases); `BEADS_TEST_PROXIED_SERVER=1 … -run 'TestProxiedServerContext$'` rc=0.

Things I tried to break and could not: no `os.Setenv` runs inside any `init()` or package-level initializer in the repo, so the snapshot really does predate bd's own export; both `bd context` routes compute in the client process, so a long-lived server's environment never stands in for the caller's; and the comparison mirrors the finder's canonicalization — trailing slash, relative `../.beads`, and a symlinked path all report `true`, and `BEADS_DIR=` (empty) reports `false` like unset. Your argument for the snapshot over finder-recorded provenance is right as far as `BEADS_DIR` goes: the finder reads the live env too.

## One follow-up — not blocking, and partly my doing

I offered the snapshot as the cheaper option, and it is narrower than I said. `BEADS_DIR` is not the only way a caller names a workspace: `--db` / `BEADS_DB` / `BD_DB` and `-C` reach the same `os.Setenv("BEADS_DIR", …)` (`cmd/bd/main.go` `prepareSelectedCommandContext`, and the `-C` handling), and the snapshot cannot see those. So:

```
# CWD = gitA (a git repo):      bd --db $T/gitB/.beads context --json
  "is_redirected": true,  "role": "contributor"
# same command after rm -rf gitA/.git:
  "is_redirected": false, "role": "maintainer"
```

Same explicit selection, opposite identity. It is a wrong answer, not a wrong action — `IsContributor()` / `IsMaintainer()` still have no non-test callers, `is_redirected` from this path feeds only the two display sites, and `GET /v0/beads/context` does not serve it — and before this PR that second command had no answer at all, so I am not holding the fix for it. But the identity probe should not say "maintainer" of a workspace the caller named from outside. The durable shape is for the selection site to record *that the caller chose* (one bit set where `--db` / `-C` / inherited `BEADS_DIR` are resolved) rather than inferring it from the env afterwards. Fold it in if you are pushing anyway, or file it and I'll review it there; either is fine by me. Related: the comment at `internal/beads/context.go:178-180` ("mirrors FindBeadsDir's step 1, the same condition that chose the directory") is no longer quite true for exactly these cases, since step 1 reads the live env.

## Nits

- `context.go:203-204` points at "`beadsDirFromCaller` in package main (GH#4635)". That symbol is not on `main` — #4795 is still open — so today the reference dangles. When both land, derive one from the other rather than keeping two snapshots.
- `stubCallerBeadsDir` (`no_repo_root_test.go`) mutates a package var; restored with `t.Cleanup` and nothing on that path is parallel today. A one-line "not safe under `t.Parallel()`" comment would keep it that way.

## Merge state

`git merge-tree --write-tree origin/main HEAD` → rc=1, one conflict, `CHANGELOG.md` only; no code conflicts. A rebase will move the head past this approval — I'll re-verify the rebased head when it lands, and if it is the CHANGELOG resolution alone that is a range-diff read, not another round. Merge is @julianknutsen's call.

Thanks for the #6560 correction, and for checking your own environment before I repeated the wrong cause a second time.

— bee

harry-miller-trimble:
Merged current main at `961aa6239d`. The only conflict was CHANGELOG.md. The resolution takes main's file and places this PR's entry under `## [Unreleased]` / `### Fixed`. No code changed. Build, vet, `internal/beads` and the `bd context` tests pass on the merged tree.

_claude-code-opus-5-5-unknown-reasoning on behalf of Harry Miller_

bee-ghosttrack (approved):
Re-read at `961aa6239d` against my approval at `ef69b571d0`: outside CHANGELOG.md the PR's own diff against main is byte-identical, and the entry now sits under `## [Unreleased]` (#6643). Approval stands. CI is still pending on this head; it's for the maintainer to confirm green before merging.

PR #4789, opened 2026-07-14
docs(create): document that --deps 'blocks:id' reverses edge direction (#4693)

harry-miller-trimble:
Closes #4693

## Problem

`bd create --deps` wires opposite edge directions depending on how the dep is spelled, and the flag help documented none of it (GH#4693):

- `--deps X` (bare id) → new issue **depends on** X
- `--deps depends-on:X` / `blocked-by:X` → new issue **depends on** X
- `--deps blocks:X` → **X depends on the new issue** (direction reversed)

The swap is intentional (verb reading: "new issue blocks X"), but the old help string —

```
Dependencies in format 'type:id' or 'id' (e.g., 'discovered-from:bd-20,blocks:bd-15' or 'bd-20')
```

— put `blocks:bd-15` right next to bare-id usage, reading as "the same thing with an explicit type". Nothing hinted that `blocks:` inverts.

Field report from the issue: an agent ran `--deps blocks:<gate>` intending "new task is blocked by the gate"; instead the gate silently ended up depending on the new task (gate went BLOCKED). The follow-up `bd dep add` in the intended direction then failed with "would create a cycle" — two steps removed from the cause. This is a nasty shape for agent workflows: silent at create time, corrupts the ready-set, surfaces later as a confusing cycle error.

## Fix

Documentation only — no behavior change. Rewrite the `--deps` flag help to spell out both directions explicitly:

```
Dependencies as 'type:id' or bare 'id'. Bare 'id', 'depends-on:id', and
'blocked-by:id' all make THIS issue depend on id; 'blocks:id' reverses
direction (id depends on this issue). E.g. 'blocked-by:bd-20,discovered-from:bd-15'
```

Regenerated CLI docs (`docs/CLI_REFERENCE.md`, `docs/cli-reference/create.md`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4793, opened 2026-07-14
fix(dolt): derive circuit-breaker temp dir from os.TempDir(), not /tmp (#4636)

### LINKED ISSUE #4636: Windows: dolt circuit-breaker hardcodes /tmp, writes to C:\tmp instead of os.TempDir()
## problem
the dolt circuit-breaker's state-file directory is a hardcoded unix literal (`"/tmp/beads-circuit"`), not derived from `os.TempDir()`. on windows, `filepath.Join("/tmp/beads-circuit", ...)` resolves the leading `/tmp` against the current drive root, so every bd invocation against a dolt server writes circuit-breaker json to `C:\tmp\beads-circuit\...` instead of `%TEMP%`. it's silent (writes succeed, no error) and fires on every bd command that touches a dolt server, so `C:\tmp` accumulates stale `beads-dolt-circuit-*.json` files indefinitely (1,011 legacy flat files + a live subdir observed on one machine).

## versions
bd v1.1.0 (commit 8e4e59d39); confirmed present at main @187ba85f3. Windows

## location
- `internal/storage/dolt/circuit.go:88` — `const circuitBreakerDir = "/tmp/beads-circuit"` (the writer, used by `newCircuitBreaker`)
- `internal/storage/dolt/circuit.go:329` — `_ = os.Remove("/tmp/beads-dolt-circuit-0.json")` (legacy port-0 cleanup)
- `cmd/bd/doctor/circuit.go:14` — `dir := "/tmp/beads-circuit"` (doctor check, independent hardcoded copy)

all three are compile-time literals — no env var (`%TEMP%`, `$TMPDIR`) can redirect them. verified present unchanged at upstream/main @187ba85f3 (identical to fleet HEAD 8c958d225).

## repro
on windows:
```
bd dolt start   # or any bd command against a running dolt server
dir C:\tmp\beads-circuit
```
`beads-dolt-circuit-*.json` files appear under `C:\tmp` (drive root), not `%TEMP%\beads-circuit`.

## suggested fix
derive the directory from `os.TempDir()` at all three sites:
```go
var circuitBreakerDir = filepath.Join(os.TempDir(), "beads-circuit") // const -> var, runtime call
```
apply the same `filepath.Join(os.TempDir(), ...)` to the doctor check (`cmd/bd/doctor/circuit.go:14`) and the legacy port-0 cleanup literal (`circuit.go:329`). `os.TempDir()` returns `%TEMP%` on windows and honors `$TMPDIR` on unix (default `/tmp`) — correct cross-platform, no other behavior change.

caveat: `CleanStaleCircuitBreakerFiles`/`cleanStaleCircuitBreakerFilesIn` only sweep the dedicated subdirectory plus the single legacy port-0 file — they won't retroactively clean the flat `beads-dolt-circuit-*.json` files already accumulated at `C:\tmp` on affected machines. worth widening the legacy-cleanup sweep to glob `os.TempDir()`'s root for `beads-dolt-circuit-*.json` too as a one-time migration, but that's separate from the core fix.

harry-miller-trimble:
Closes #4636

## Problem

The Dolt circuit-breaker's state directory was a hardcoded unix literal, `"/tmp/beads-circuit"` — as were the doctor check's independent copy and the legacy port-0 cleanup path. On Windows, `filepath.Join("/tmp/beads-circuit", …)` resolves the leading `/tmp` against the current drive root, so **every** bd command that touches a Dolt server wrote circuit-breaker JSON to `C:\tmp\beads-circuit` instead of `%TEMP%`. It's silent (writes succeed) and fires on every invocation, so `C:\tmp` accumulates stale `beads-dolt-circuit-*.json` files indefinitely (the report observed 1,011 legacy flat files + a live subdir on one machine).

Three hardcoded sites:
- `internal/storage/dolt/circuit.go` — the writer (`circuitBreakerDir`) and the legacy port-0 removal
- `cmd/bd/doctor/circuit.go` — the doctor check's copy

## Fix

- Replace the `circuitBreakerDir` const with a `circuitBreakerDir()` function returning `filepath.Join(os.TempDir(), "beads-circuit")`, and route the writer + startup cleanup through it.
- `CleanStaleCircuitBreakerFiles` now sweeps **both** the correct temp root and the legacy hardcoded `/tmp` locations (i.e. `C:\tmp\beads-circuit` on Windows), so the accumulated droppings get cleaned up on upgrade, and removes the legacy port-0 file from both roots.
- The doctor check derives the same `os.TempDir()`-based path, so it inspects where files are actually written.

No behavior change on Linux (where `os.TempDir()` is typically `/tmp`); on macOS files move to `$TMPDIR` (`/var/folders/…`) and on Windows to `%TEMP%`.

## Tests

- `TestCircuitBreakerDir_DerivedFromTempDir`: with `TMPDIR`/`TMP`/`TEMP` pointed at a temp dir, the breaker directory follows `os.TempDir()` and is no longer the `/tmp` literal.
- Updated `TestCircuitBreakerDir_UsesSubdirectory` for the new function form.


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claim at the PR head). Verdict: **needs changes** — one should-fix; the temp-dir derivation itself is right.

**Should-fix — the legacy `/tmp` cleanup never removes the files users actually accumulate.** Verified: the legacy directory is swept with `cleanStaleCircuitBreakerFilesIn`, which deliberately skips every state that isn't open/half-open — but the common steady-state artifact is a *closed* JSON (every successful open calls `RecordSuccess`, and ephemeral ports mint distinct filenames, so closed sidecars accumulate). An existing test even pins that closed files survive the cleaner. Meanwhile doctor scans only the new `os.TempDir()` location, so legacy survivors become permanently invisible orphans — contradicting the commit's cleanup claim. The legacy sweep (unlike the live-dir sweep) should remove closed files too, with a test covering a legacy closed sidecar.

Nits: the comment at the top of `circuit.go` still says state lives "in /tmp"; and the doctor + circuit packages now independently encode the same directory contract — the seam that drifted originally.

Merge-base drift: main is 143 commits ahead but none touch these files; the finding stands on current main.

maphew:
Consolidated update on your reviewed batch — all five PRs from the 07-24 reviews now carry the requested fixes as maintainer commits pushed to your branches (your commits untouched underneath), and each is approved and queued for the merge patrol:

- **#4793** (this PR): merged current main (the circuit-breaker test-override plumbing landed there in parallel; both wired together), and the legacy `/tmp/beads-circuit` sweep now removes closed sidecars too (`removeClosed` param) since nothing live ever reuses them there — plus the stale doc-comment nit. Commit `47a947d5d`.
- **#4790**: `bd config show` now reuses `resolveConfiguredActor()` so its `actor` entry reports `env: BEADS_ACTOR` provenance instead of silently showing the deprecated variable's view; wiring covered end-to-end through `rootCmd.Execute()`, not just the helper. Commit `03cdecbfe`.
- **#4789**: kept your help-text fix, dropped the two generated-doc edits per the release-pin policy. Commit `b245c5372`.
- **#4788**: whitespace-title rejection moved into shared `resolveTitle()` + a `cobra.Args` validator, so it fires before store-open/migration and covers the proxied-server path; proxied regression test added. Commit `1957fd6de`.
- **#4787**: same pattern — rejection moved into `commentsCmd.Args`, firing before store-open and proxied dispatch, with tests proving the store never opens on the swapped form. Commit `d1f5af812`.

The recurring theme in your batch (validation running after expensive store-open, missing the proxied path) was worth fixing uniformly — thanks for the clean, single-purpose PRs that made that pattern visible. Your remaining four open PRs (#4798, #4795, #4792, #4791) are author-owned per the earlier reviews and stay in your court.

One heads-up from working in your branches: `TestCLI_CreateRejectsEmptyTitle`'s `FlagTab` subtest leaves the shared `--title` cobra flag dirty for later tests in full-suite runs — pre-existing on your branch and on main, not caused by your change; we've noted it maintainer-side.

maphew (approved):

Issue #4769, opened 2026-07-14
external dependencies are stored but ignored by ready, single dep list, and show


DevMasterDru:
## Bug description

In `bd 1.1.0`, adding an `external:<project>:<capability>` blocking dependency reports success and stores the edge in `dependencies.depends_on_external`, but an unshipped capability does not block the consumer from `bd ready`. A single-issue `bd dep list` and `bd show` also omit the stored external edge.

This contradicts the current `bd dep add --help` contract:

> External references are stored as-is and resolved at query time using the external_projects config. They block the issue until the capability is shipped in the target project.

## Reproduction

Using two isolated embedded-Dolt repositories:

```sh
root=$(mktemp -d)
producer="$root/producer"
consumer="$root/consumer"
mkdir -p "$producer" "$consumer"
git -C "$producer" init -q
git -C "$consumer" init -q
(cd "$producer" && bd init --non-interactive --skip-hooks --skip-agents --prefix producer)
(cd "$consumer" && bd init --non-interactive --skip-hooks --skip-agents --prefix consumer)

producer_issue=$(bd -C "$producer" create "Produce test capability" --labels export:test-capability --silent)
consumer_issue=$(bd -C "$consumer" create "Consume test capability" --silent)
bd -C "$consumer" config set external_projects.producer "$producer"
bd -C "$consumer" dep add "$consumer_issue" external:producer:test-capability

bd -C "$consumer" dep list "$consumer_issue" --json
bd -C "$consumer" show "$consumer_issue" --json
bd -C "$consumer" ready --json

bd -C "$producer" close "$producer_issue"
bd -C "$producer" ship test-capability
bd -C "$consumer" ready --json
```

Observed output, abbreviated:

```text
Added dependency: consumer-13o depends on external:producer:test-capability (blocks)
dep list before shipment: []
show dependencies before shipment: []
ready before shipment: [consumer-13o]
ready after shipment:  [consumer-13o]
```

Direct Dolt inspection confirms the edge exists:

```json
{"rows":[{"depends_on_external":"external:producer:test-capability","issue_id":"consumer-13o","type":"blocks"}]}
```

## Expected behavior

- Before shipment, the external edge is visible in dependency output and excludes the consumer from `bd ready` and ready-based claim workflows.
- After the producer capability is shipped, the consumer becomes ready.

## Actual behavior

- The edge is stored but omitted by single-issue dependency display.
- The consumer is ready both before and after shipment.
- The configured external project is never consulted on the ready path.

[...]

PR #4795, opened 2026-07-14
fix(init): refuse to initialize directly in the home directory (#4635)

### LINKED ISSUE #4635: bd init creates a git repo + agent scaffolding in an unintended dir (e.g. $HOME) with no confirmation
## summary

`bd init` run in a directory that is not a git repo (and has no git ancestor) unconditionally `git init`s the directory, writes agent scaffolding (CLAUDE.md, AGENTS.md, .gitignore, .agents/, .codex/, .claude/settings.json), and commits — with no confirmation. run in `$HOME` this silently turns your home directory into a git repo and can clobber an existing `~/CLAUDE.md`. it's worse under `--non-interactive` (auto-engaged whenever stdin isn't a TTY, i.e. whenever an *agent* runs `bd init`): all of it happens with zero prompt.

## versions
- bd **v1.1.0** (commit `8e4e59d39`); present unchanged at **main @ `187ba85f3`**
- linux + macos + windows (path-agnostic)

## root cause

`cmd/bd/init.go:727-728`:
```go
// Ensure git is initialized — bd requires git for role config, sync branches, hooks, ...
if !isGitRepo() && !hasExplicitBeadsDir {
    gitInitCmd := exec.Command("git", "init")
```
the only guards are "already a git repo" and "explicit BEADS_DIR set". there is **no guard on where this runs** — no check that the cwd is a place the user meant to initialize (e.g. refuse/confirm when cwd is `$HOME`, a filesystem root, or a non-empty dir that isn't already a repo). the scaffolding + commit at ~L1587+ then follow unconditionally, and under `--non-interactive` there is no prompt at all.

## what it is NOT

not a bulk `git add -A`. i verified the commit is already **scoped** to bd-generated files — decoy/secret/nested files seeded in the dir beforehand are left untracked. so the fix is **not** commit-scoping; the defect is *creating a git repo + beads project in an unintended dir without consent*.

## repro
```sh
cd "$HOME"        # or any non-repo dir you didn't mean to initialize
bd init --non-interactive
# -> $HOME is now a git repo, scaffolded with agent files, .beads committed — no prompt
```

## impact
- a new user (or an agent) can turn `$HOME`, or any wrong directory, into a git repo + beads project silently
- can clobber an existing `~/CLAUDE.md`
- if `~/.beads` already exists (any shared-server box), the scoped commit still sweeps a large pre-existing tree into the new repo

## suggested fix

gate repo-creation on consent before `git init`: refuse — or require `--yes` / an interactive confirm — when the target is `$HOME`, a filesystem root, or a non-empty directory that isn't already a git repo. under `--non-interactive`, refuse rather than proceed silently. `BEADS_DIR`-set already opts out, which is a good precedent.

harry-miller-trimble:
Closes #4635

## Problem

`bd init` in a directory that isn't a git repo unconditionally `git init`s it, writes agent scaffolding (`CLAUDE.md`, `AGENTS.md`, `.gitignore`, `.agents/`, `.codex/`, `.claude/settings.json`), and commits — with no confirmation. Run in `$HOME` this silently turns the home directory into a git repository and clobbers whatever agent files are already there.

It's worse under `--non-interactive`, which auto-engages whenever stdin isn't a TTY — i.e. whenever an *agent* runs `bd init`. All of it happens with zero prompt.

## Fix

Refuse, when the working directory **is** the user's home directory, it isn't already a git repo, and the caller didn't name a `BEADS_DIR`:

```
$ cd ~ && bd init
bd init refuses: /Users/me is your home directory, and it is not a git repository.

  Why: init would 'git init' your home directory and scaffold agent files
       into it (CLAUDE.md, AGENTS.md, .gitignore, .claude/, .codex/,
       .agents/), overwriting any you already keep there. A project is
       virtually never the home directory itself.

  Next:
    Initialize a project instead (recommended):
      cd ~/some-project && bd init

    Keep the workspace elsewhere, with no scaffolding in your home:
      BEADS_DIR=<path>/.beads bd init
      (this is a DIFFERENT init: an explicit BEADS_DIR skips the git init
       and the working-directory-local .beads, so nothing is written here)

    Recovery playbook:
      docs/recovery/init-safety.md#init-home-refused
$ echo $?
13
```

**Placement.** The guard is the first statement in init's `RunE`, ahead of every path that has an effect: the proxied dispatch (`runInitProxiedServer` → `EnsureGitRepo` → `git init`), the `--reinit-local`/`--force` pre-checks (`countExistingIssues` opens a real store, creating the data directory and able to run migrations), and the mode globals `RunE` sets on itself.

**Opt-in is caller intent, not the process environment.** The guard reads `beadsDirFromCaller` — `BEADS_DIR` as the process *inherited* it, captured at package initialization — not `os.Getenv("BEADS_DIR")`. By the time `RunE` runs, bd has usually exported one itself: the root `PersistentPreRunE` resolves a beads dir for every no-DB command and exports it, and that resolution accepts any ancestor `.beads/` holding so much as a `config.yaml` — including this project's own legacy user-level `~/.beads/config.yaml`. `bd -C <dir>` exports one too, and that one doesn't even move the working directory the scaffolding lands in.

**Refusal conventions.** Exit code `13` (`ExitHomeDirRefused`), alongside init's existing 10/11/12, so a script can branch on the number instead of grepping stderr. `docs/recovery/init-safety.md#init-home-refused` is the playbook the message points at, and `bd help init-safety` lists the code.

## Scope

Deliberately narrow. It fires only when **all** of these hold:

- the working directory is *exactly* your home directory — `mkdir proj && cd proj && bd init` is unaffected, and so is any other subdirectory;
- your home is not already a git repository — a dotfiles repo is unaffected;
- the caller did not name a `BEADS_DIR`.

`--force` and `--reinit-local` do **not** override it: they bypass the local data-safety guard, which is a different question from "is this the right directory". `/`, `/Users`, `/home` and the like stay unguarded — this is about the one directory people actually land in by accident.

## Tests

- `TestIsUserHomeDir`, `TestIsUserHomeDir_UnresolvableHome` — the decision, with the home resolver stubbed so no test touches the real home directory.
- `TestGuardInitInHomeDir` — the fail-open cases (caller-supplied `BEADS_DIR`, an already-tracked home, an ordinary project dir), plus the case that keeps the provenance fix honest: a `BEADS_DIR` that *bd* exported is not caller opt-in.
- `TestInitCommand_RefusesInHomeDirBeforeAnySideEffect` — runs init's `RunE` across plain / `--reinit-local` / `--force` / `--proxied-server` and asserts not just the refusal but that **nothing** exists afterwards: `.git`, `.beads`, `CLAUDE.md`, `AGENTS.md`, `.gitignore`, `.agents`, `.codex`, `.claude`. This is what pins the guard's *position*; moving it back down the file fails all four.
- `TestInitThroughRootCmd_RefusesInHomeDirWithDiscoveredBeadsDir` — through `rootCmd.Execute`, in a home carrying only `.beads/config.yaml`, because the bypass lives in a pre-run hook that `RunE`-only tests skip. It asserts the precondition (pre-run really does discover that directory) before asserting the refusal, and names a backend that cannot exist as a circuit breaker, so a future fail-open stops at backend selection rather than building a store in someone's home directory.

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"runtime"
 	"strconv"
@@ -321,6 +322,69 @@ func seedInitWorkspaceIdentity(
 	return nil
 }
 
+// resolveUserHomeDir returns the user's home directory.
+//
+// It prefers the OS account database over $HOME so the guard keys off the real
+// account home even when $HOME has been repointed — which happens constantly in
+// this repo's own tests, and in any sandbox that fakes a home. Both sources are
+// best-effort: in a CGO-free build user.Current() falls back to os.UserHomeDir()
+// (i.e. $HOME on Unix) itself, so this is a preference, not a guarantee of
+// $HOME-independence. Mirrors the home resolution in internal/beads'
+// isPathInSafeBoundary.
+//
+// A package var so tests can stub it.
+var resolveUserHomeDir = func() string {
+	if u, err := user.Current(); err == nil && u.HomeDir != "" {
+		return u.HomeDir
+	}
+	home, _ := os.UserHomeDir()
+	return home
+}
+
+// isUserHomeDir reports whether dir is the user's home directory (path-boundary
+// comparison, symlink/case tolerant). Used to guard `bd init` against turning
+// $HOME into a git repo (GH#4635).
+func isUserHomeDir(dir string) bool {
+	home := resolveUserHomeDir()
+	return home != "" && utils.PathsEqual(dir, home)
+}
+
+// guardInitInHomeDir refuses `bd init` directly in the user's home directory
+// when it is not already a git repo and no explicit BEADS_DIR was given.
+//
+// `bd init` would otherwise `git init` the home directory and write agent
+// scaffolding (CLAUDE.md, AGENTS.md, .gitignore, .agents/, .codex/,
+// .claude/settings.json) there, potentially clobbering existing files — a
+// silent footgun, especially under --non-interactive, which auto-engages
+// whenever stdin is not a TTY, i.e. whenever an agent runs init (GH#4635). A
+// real project is virtually never the home directory itself.
+//
+// This runs as the first thing in init's RunE, ahead of every path that has an
+// effect: the proxied dispatch (runInitProxiedServer -> EnsureGitRepo -> git
+// init), the --reinit-local/--force pre-checks (countExistingIssues opens a
+// real store, which creates the data directory and can run migrations), and
+// the mode globals RunE sets on itself. A guard placed after any of those
+// refuses only after the damage it exists to prevent.
+//
+// Fail-open by construction: an unresolvable home, an explicit BEADS_DIR, or an
+// already-tracked home all mean "not the case this guards", and init proceeds.
+func guardInitInHomeDir() error {
+	if os.Getenv("BEADS_DIR") != "" {
+		return nil
+	}
+	if isGitRepo() {
+		return nil
+	}
+	cwd, err := os.Getwd()
+	if err != nil || !isUserHomeDir(cwd) {
+		return nil
+	}
+	return fmt.Errorf("refusing to 'bd init' directly in your home directory (%s):\n"+
+		"  this would turn it into a git repository and scaffold agent files here,\n"+
+		"  overwriting any existing ones. cd into a project directory and re-run, or set\n"+
+		"  BEADS_DIR to an explicit location if you really mean to initialize here", cwd)
+}
+
 var initCmd = &cobra.Command{
 	Use:           "init",
 	GroupID:       "setup",
@@ -361,6 +425,12 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
   • --contributor and --team flags are rejected (wizards require interaction)
   Also auto-detected when stdin is not a terminal or CI=true is set.`,
 	RunE: func(cmd *cobra.Command, _ []string) (retErr error) {
+		// First, before anything with an effect: never turn $HOME into a git
+		// repo (GH#4635). See guardInitInHomeDir for why the position matters.
+		if err := guardInitInHomeDir(); err != nil {
+			return err
+		}
+
 		prefix, _ := cmd.Flags().GetString("prefix")
 		quiet, _ := cmd.Flags().GetBool("quiet")
 		contributor, _ := cmd.Flags().GetBool("contributor")
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -366,10 +366,26 @@ func isUserHomeDir(dir string) bool {
 // the mode globals RunE sets on itself. A guard placed after any of those
 // refuses only after the damage it exists to prevent.
 //
-// Fail-open by construction: an unresolvable home, an explicit BEADS_DIR, or an
-// already-tracked home all mean "not the case this guards", and init proceeds.
+// Fail-open by construction: an unresolvable home, a caller-supplied BEADS_DIR,
+// or an already-tracked home all mean "not the case this guards", and init
+// proceeds.
+//
+// The opt-in check reads beadsDirFromCaller, NOT os.Getenv("BEADS_DIR"). By the
+// time this runs the root PersistentPreRunE has already exported a beads dir it
+// discovered for itself, and beads.FindBeadsDir() accepts any ancestor .beads/
+// holding a config.yaml — including this project's own legacy user-level
+// ~/.beads/config.yaml. Keying off the live environment therefore made the
+// guard fail open for exactly the long-standing users most likely to have
+// scaffolding in $HOME worth not clobbering. `bd -C <dir> init` exported one
+// too, and that one does not even move the working directory the scaffolding
+// lands in.
+//
+// Refusals carry ExitHomeDirRefused so a script can branch on the number, like
+// init's three other refusal classes. The message is printed here rather than
+// returned so the exit code is the error's whole payload; docs/recovery/
+// init-safety.md#init-home-refused is the recovery playbook it points at.
 func guardInitInHomeDir() error {
-	if os.Getenv("BEADS_DIR") != "" {
+	if beadsDirFromCaller != "" {
 		return nil
 	}
 	if isGitRepo() {
@@ -379,10 +395,37 @@ func guardInitInHomeDir() error {
 	if err != nil || !isUserHomeDir(cwd) {
 		return nil
 	}
-	return fmt.Errorf("refusing to 'bd init' directly in your home directory (%s):\n"+
-		"  this would turn it into a git repository and scaffold agent files here,\n"+
-		"  overwriting any existing ones. cd into a project directory and re-run, or set\n"+
-		"  BEADS_DIR to an explicit location if you really mean to initialize here", cwd)
+	fmt.Fprintf(os.Stderr, "%s\n", homeDirRefusalMessage(cwd))
+	return &exitError{Code: ExitHomeDirRefused}
+}
+
+// homeDirRefusalMessage is the What/Why/Next refusal text for a `bd init` run
+// directly in an untracked home directory.
+//
+// The BEADS_DIR escape hatch is described honestly: setting it does not run the
+// init the caller asked for in $HOME, it runs a different one — an explicit
+// BEADS_DIR makes hasExplicitBeadsDir true, which skips the `git init` and the
+// CWD-local .beads semantics. Saying "set BEADS_DIR if you really mean it"
+// without that caveat sends the user somewhere they did not intend to go.
+func homeDirRefusalMessage(cwd string) string {
+	return fmt.Sprintf(`bd init refuses: %s is your home directory, and it is not a git repository.
+
+  Why: init would 'git init' your home directory and scaffold agent files
+       into it (CLAUDE.md, AGENTS.md, .gitignore, .claude/, .codex/,
+       .agents/), overwriting any you already keep there. A project is
+       virtually never the home directory itself.
+
+  Next:
+    Initialize a project instead (recommended):
+      cd ~/some-project && bd init
+
+    Keep the workspace elsewhere, with no scaffolding in your home:
+      BEADS_DIR=<path>/.beads bd init
+      (this is a DIFFERENT init: an explicit BEADS_DIR skips the git init
+       and the working-directory-local .beads, so nothing is written here)
+
+    Recovery playbook:
+      docs/recovery/init-safety.md#init-home-refused`, cwd)
 }
 
 var initCmd = &cobra.Command{
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -153,6 +153,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   configured. Auto-start warns and retargets, as it did before. Shared-server
   mode still fails closed there — a repo-local auto-start is a different
   database, not a port refresh — but now says so in its own words.
+- **`bd init` refuses to initialize directly in your home directory**
+  ([#4635](https://github.com/gastownhall/beads/issues/4635)). Run there, `bd
+  init` would `git init` your home directory and scaffold agent files into it —
+  `CLAUDE.md`, `AGENTS.md`, `.gitignore`, `.claude/`, `.codex/`, `.agents/` —
+  overwriting any you already keep there, and it did so silently under
+  `--non-interactive`, which engages automatically whenever stdin is not a TTY
+  (so: whenever an agent runs it). A project is virtually never the home
+  directory itself, so this now refuses with exit code `13`
+  (`ExitHomeDirRefused`), documented at
+  `docs/recovery/init-safety.md#init-home-refused`.
+
+  Deliberately narrow: it fires only when the working directory is *exactly*
+  your home directory, your home is not already a git repository (tracked
+  dotfiles are unaffected), and you did not name a `BEADS_DIR` yourself. A
+  subdirectory of home is an ordinary project and is never refused. `--force`
+  and `--reinit-local` do not override it — they bypass the local data-safety
+  guard, which is a different question from "is this the right directory". The
+  guard runs first in init's `RunE`, ahead of every path with an effect,
+  including the proxied dispatch and the store-opening `--reinit-local`
+  pre-checks.
 
 - **`bd reclaim` summarizes the leases its replica guard declined instead of
   naming every one, every run** (wy-sp2l4). A lease granted by another replica
--- a/cmd/bd/init_home_guard_test.go
+++ b/cmd/bd/init_home_guard_test.go
@@ -1,0 +1,200 @@
+package main
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// TestIsUserHomeDir covers the guard decision used to refuse `bd init` directly
+// in the user's home directory (GH#4635). The resolver is stubbed so the test
+// does not depend on (or touch) the real home directory.
+func TestIsUserHomeDir(t *testing.T) {
+	home := t.TempDir()
+	stubHomeDir(t, home)
+
+	if !isUserHomeDir(home) {
+		t.Errorf("isUserHomeDir(home) = false, want true for %q", home)
+	}
+	if sub := filepath.Join(home, "project"); isUserHomeDir(sub) {
+		t.Errorf("isUserHomeDir(%q) = true, want false (subdir of home)", sub)
+	}
+	if parent := filepath.Dir(home); isUserHomeDir(parent) {
+		t.Errorf("isUserHomeDir(%q) = true, want false (parent of home)", parent)
+	}
+}
+
+// TestIsUserHomeDir_UnresolvableHome verifies the guard never fires when the
+// home directory cannot be resolved (fail-open: better to allow init than to
+// wrongly refuse every directory).
+func TestIsUserHomeDir_UnresolvableHome(t *testing.T) {
+	stubHomeDir(t, "")
+
+	if isUserHomeDir("/some/dir") {
+		t.Error("isUserHomeDir should be false when home is unresolvable")
+	}
+}
+
+// TestGuardInitInHomeDir covers the conditions under which the guard declines
+// to fire: an explicit BEADS_DIR, an already-tracked home, and any directory
+// that simply is not home.
+func TestGuardInitInHomeDir(t *testing.T) {
+	t.Run("refuses_in_untracked_home", func(t *testing.T) {
+		home := t.TempDir()
+		stubHomeDir(t, home)
+		chdir(t, home)
+		t.Setenv("BEADS_DIR", "")
+
+		err := guardInitInHomeDir()
+		if err == nil {
+			t.Fatal("guardInitInHomeDir() = nil, want a refusal in an untracked home")
+		}
+		if !strings.Contains(err.Error(), "refusing to 'bd init' directly in your home directory") {
+			t.Errorf("unexpected error text: %v", err)
+		}
+	})
+
+	t.Run("allows_with_explicit_beads_dir", func(t *testing.T) {
+		home := t.TempDir()
+		stubHomeDir(t, home)
+		chdir(t, home)
+		t.Setenv("BEADS_DIR", filepath.Join(home, "elsewhere", ".beads"))
+
+		if err := guardInitInHomeDir(); err != nil {
+			t.Errorf("guardInitInHomeDir() = %v, want nil when BEADS_DIR is explicit", err)
+		}
+	})
+
+	t.Run("allows_when_home_is_already_a_git_repo", func(t *testing.T) {
+		home := t.TempDir()
+		stubHomeDir(t, home)
+		chdir(t, home)
+		t.Setenv("BEADS_DIR", "")
+		gitInitAt(t, home)
+
+		if err := guardInitInHomeDir(); err != nil {
+			t.Errorf("guardInitInHomeDir() = %v, want nil when home is already tracked", err)
+		}
+	})
+
+	t.Run("allows_in_an_ordinary_project_dir", func(t *testing.T) {
+		home := t.TempDir()
+		stubHomeDir(t, home)
+		project := filepath.Join(home, "project")
+		if err := os.MkdirAll(project, 0o750); err != nil {
+			t.Fatal(err)
+		}
+		chdir(t, project)
+		t.Setenv("BEADS_DIR", "")
+
+		if err := guardInitInHomeDir(); err != nil {
+			t.Errorf("guardInitInHomeDir() = %v, want nil for a project dir under home", err)
+		}
+	})
+}
+
+// TestInitCommand_RefusesInHomeDirBeforeAnySideEffect runs the actual init
+// command, which is where the guard's *position* is under test rather than its
+// decision. Both flag shapes below reach a side effect before the guard's
+// original location did:
+//
+//   - --reinit-local / --force run countExistingIssues, which opens a real
+//     store (creating the data directory, and able to run migrations);
+//   - --proxied-server dispatches to runInitProxiedServer, which calls
+//     EnsureGitRepo — a `git init` in the current directory.
+//
+// So the assertion is not merely "it errors": nothing may exist afterwards.
+func TestInitCommand_RefusesInHomeDirBeforeAnySideEffect(t *testing.T) {
+	cases := []struct {
+		name  string
+		flags map[string]string
+	}{
+		{name: "plain", flags: nil},
+		{name: "reinit_local", flags: map[string]string{"reinit-local": "true"}},
+		{name: "force", flags: map[string]string{"force": "true"}},
+		{name: "proxied_server", flags: map[string]string{"proxied-server": "true"}},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			home := t.TempDir()
+			stubHomeDir(t, home)
+			chdir(t, home)
+			t.Setenv("BEADS_DIR", "")
+			t.Setenv("BD_NON_INTERACTIVE", "1")
+			for name, value := range tc.flags {
+				setInitFlag(t, name, value)
+			}
+
+			err := initCmd.RunE(initCmd, nil)
+			if err == nil {
+				t.Fatal("bd init in an untracked home returned nil, want a refusal")
+			}
+			if !strings.Contains(err.Error(), "refusing to 'bd init' directly in your home directory") {
+				t.Fatalf("unexpected error: %v", err)
+			}
+
+			// A refusal that still left something behind is not a refusal.
+			for _, name := range []string{".git", ".beads", "CLAUDE.md", "AGENTS.md", ".gitignore", ".agents", ".codex", ".claude"} {
+				if _, statErr := os.Stat(filepath.Join(home, name)); statErr == nil {
+					t.Errorf("init created %s in the home directory despite refusing", name)
+				}
+			}
+		})
+	}
+}
+
+// stubHomeDir points the guard's home resolver at dir for the duration of the
+// test. Stubbing rather than setting $HOME is deliberate: the resolver prefers
+// the OS account database, so $HOME alone would not move it.
+func stubHomeDir(t *testing.T, dir string) {
+	t.Helper()
+	orig := resolveUserHomeDir
+	resolveUserHomeDir = func() string { return dir }
+	t.Cleanup(func() { resolveUserHomeDir = orig })
+}
+
+// gitInitAt makes dir a git repository.
+func gitInitAt(t *testing.T, dir string) {
+	t.Helper()
+	cmd := exec.Command("git", "init")
+	cmd.Dir = dir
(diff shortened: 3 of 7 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the ordering at the PR head). Verdict: **needs changes** — right guard, placed after the hazards it guards.

**Blocker 1 — a store can open and migrate before the refusal.** Verified: `--reinit-local`/`--force` call `countExistingIssues` (init.go:513) well before the home guard (~:699), and that helper opens a real store — data-directory creation, schema init, remote-migrate gate, `MigrateUp`. So `bd init` in an untracked home with existing beads data can mutate it before refusing. The guard must run ahead of every store-opening path — first thing in RunE.

**Blocker 2 — merge-base drift leaves proxied init unguarded.** At the PR head proxied init was disabled; upstream has since enabled it, and current main returns through `runInitProxiedServer` (which calls `EnsureGitRepo` → `git init`) *before* the guard's merge location — plus there's a textual conflict in init.go anyway. The rebase needs one shared guard before the proxied dispatch, not a mechanical conflict resolution.

**Should-fix — the CGO-free rationale doesn't hold.** The comment claims `user.Current` is `HOME`-independent, but Go's pure-Go fallback is `os.UserHomeDir` → `$HOME` on Unix, same as the explicit fallback — and CGO-free binaries are a supported configuration. Don't over-promise; treat the env-derived home as best-effort in both builds.

**Should-fix — tests cover only the helper.** No test runs the command to verify refusal ordering or absence of `.git`/`.beads` side effects — which is exactly where both blockers hide. Add command-level coverage for ordinary and proxied init.

harry-miller-trimble:
Thanks — rebased onto current `main` (the `init.go` conflict was an add/add in the helper block, resolved by keeping both) and all four points addressed.

**Blocker 1 — a store can open and migrate before the refusal: fixed.** The guard is now the first statement in init's `RunE`, ahead of `countExistingIssues`, ahead of the mode globals `RunE` sets on itself, ahead of everything. It's extracted into `guardInitInHomeDir()` so the placement is a one-liner at the top rather than a block wedged into the middle of the path validation.

**Blocker 2 — proxied init unguarded: fixed by the same move.** Being first in `RunE` puts it ahead of the `--proxied-server` dispatch (`runInitProxiedServer` → `EnsureGitRepo` → `git init`), so one guard covers both routes rather than two copies drifting apart.

**Should-fix — tests cover only the helper: fixed, and this is what proves the ordering.** `TestInitCommand_RefusesInHomeDirBeforeAnySideEffect` runs the actual command (in-process `initCmd.RunE`, the pattern already used by `metrics_test.go`, `gate_test.go`, `list_ready_brief_test.go`) across plain / `--reinit-local` / `--force` / `--proxied-server`, and asserts not just the refusal but that **nothing** exists afterwards: `.git`, `.beads`, `CLAUDE.md`, `AGENTS.md`, `.gitignore`, `.agents`, `.codex`, `.claude`.

I confirmed it catches the old placement. Moving the guard back to where it was and re-running:

```
--- FAIL: TestInitCommand_RefusesInHomeDirBeforeAnySideEffect/proxied_server
panic: runtime error: invalid memory address or nil pointer dereference
  ...doltversion.Probe
  ...cmd/bd.resolveAndProbeDolt (uow_factory.go:85)
  ...cmd/bd.runInitProxiedServer (init_proxied_server.go:80)
  ...cmd/bd.init.func205 (init.go:679)
```

— i.e. the old position let proxied init get all the way into `runInitProxiedServer` before the guard was ever consulted. The helper-level decision tests are still there (`TestIsUserHomeDir`, plus a new `TestGuardInitInHomeDir` covering the three fail-open cases: explicit `BEADS_DIR`, an already-tracked home, and an ordinary project dir under home). The resolver is stubbed throughout, so no test touches the real home directory.

**Should-fix — the CGO-free rationale: corrected.** You're right that the pure-Go `user.Current()` falls back to `os.UserHomeDir()` → `$HOME` when the account is absent from the passwd database. The comment no longer claims `$HOME`-independence; it now says the account database is *preferred* over `$HOME` — which is still worth doing, since this repo's own test harness (`bdEnv`) sets `HOME=<workspace>` and inits there, and reading `$HOME` first would make the guard fire on every one of those — while stating plainly that both sources are best-effort.

Tests: the new tests plus the existing `TestInit*` suite pass locally (`go test -tags gms_pure_go ./cmd/bd/ -run 'TestIsUserHomeDir|TestGuardInitInHomeDir|TestInitCommand_RefusesInHomeDir|TestInit'`), `gofmt` and `go vet` clean.

harry-miller-trimble:
Status check on this one: all four points from the 2026-07-24 review were addressed in the rebase pushed on 2026-08-14 (details in the comment above), and the branch is still `MERGEABLE`/`CLEAN` against current `main` with CI green.

@maphew — you re-reviewed and approved my other three from that batch (#4791, #4792, #4798) on 2026-08-16 but this one wasn't in the sweep, so I suspect it was just missed rather than held back. If you'd rather not pick it up now that you've stepped down, no problem.

@csells @julianknutsen — this is the sibling of those three and is in the same state; it needs a fresh reviewer since the standing `CHANGES_REQUESTED` predates the fixes.

_claude-code-opus-5-unknown-reasoning on behalf of Harry Miller_

cmd/bd/init.go
```diff
@@ -321,6 +322,69 @@ func seedInitWorkspaceIdentity(
...
+// effect: the proxied dispatch (runInitProxiedServer -> EnsureGitRepo -> git
+// init), the --reinit-local/--force pre-checks (countExistingIssues opens a
+// real store, which creates the data directory and can run migrations), and
+// the mode globals RunE sets on itself. A guard placed after any of those
+// refuses only after the damage it exists to prevent.
+//
+// Fail-open by construction: an unresolvable home, an explicit BEADS_DIR, or an
+// already-tracked home all mean "not the case this guards", and init proceeds.
+func guardInitInHomeDir() error {
+	if os.Getenv("BEADS_DIR") != "" {
```
maphew: [P1] Distinguish caller-supplied BEADS_DIR from pre-run discovery

By this point, a nonempty BEADS_DIR is not proof that the user supplied it. Root PersistentPreRunE calls selectedNoDBBeadsDir -> prepareSelectedNoDBContext -> prepareSelectedCommandContext, which exports the discovered path with os.Setenv. In an untracked home containing existing .beads project files, this branch therefore returns nil even when the caller supplied no BEADS_DIR. Reinit can still reach countExistingIssues and open the existing store, the data-mutation case raised in the July review. Preserve the original opt-in state or check before that export, after directory selection. The RunE-only tests omit the hook that causes this.

maphew (commented):
Thanks for the follow-up, Harry. Re-reviewed `300d39ab91e3f74898d57a96585e82e67da32ace` and its merge with current main, `2f5795ca265cf6caca7864352436e6359d8b0e60`.

The first-in-`RunE` placement and CGO wording address those parts of the earlier review. All the added tests pass locally, and the seven reported CI workflows are green. However, one command-lifecycle bypass remains, detailed inline, so I cannot clear the earlier review yet.

With an existing `~/.beads/config.yaml`, root pre-run discovery sets `BEADS_DIR` itself. The new guard mistakes this for explicit user opt-in. The new tests invoke `RunE` directly and therefore miss that setup.

I reproduced the bypass using a temporary, stubbed home, no git repo, and an initially empty `BEADS_DIR`. Through `rootCmd.ExecuteContext`, `init --backend=review4795-safety-sentinel` returns the unknown-backend error instead of the expected home refusal. That deliberately invalid backend stops the experiment before storage initialization. The same bypass occurs at the PR head and the current merge.

Please preserve whether `BEADS_DIR` was caller-supplied, or run the guard after `-C` selection but before automatic discovery exports it. Add coverage through Cobra execution with pre-existing beads data, including reinit, rather than only an empty home and direct `RunE`.

Validation: Windows, Go 1.27.1, `CGO_ENABLED=0`, `-tags gms_pure_go`; focused home-guard tests pass, added lifecycle regression fails as described. Full suite not run.

bee-ghosttrack (changes requested):
Round 1 from me (bee). Adversarial pass on `300d39ab91e3f74898d57a96585e82e67da32ace`, in a throwaway worktree off the PR head. I like the shape of this — the guard is in the right place and its tests are not decorative — but it has a hole that makes it inert for a realistic class of users, and it skips three `bd init` refusal conventions this repo enforces on every other refusal.

## What I ran

| Command | Result |
|---|---|
| `git merge-tree --write-tree origin/main 300d39ab9` | rc=0 — clean merge (main is 119 commits ahead; PR is 1 ahead) |
| `git diff --check origin/main..300d39ab9` | rc=0 |
| `go build ./...` | rc=0 |
| `go vet ./cmd/bd/` | rc=0 |
| `go test ./cmd/bd -run 'TestIsUserHomeDir\|TestGuardInitInHomeDir\|TestInitCommand_RefusesInHomeDirBeforeAnySideEffect' -count=1 -v` (cgo) | rc=0 — PASS, 10/10 incl. subtests, 150s |
| same, `CGO_ENABLED=0 -tags gms_pure_go` | rc=0 — PASS, 13.9s |

Diff is the two files the PR claims (`cmd/bd/init.go` +70, `cmd/bd/init_home_guard_test.go` +200). No undisclosed files.

## Mutation probes — the tests are NOT vacuous

Four probes, each restored to a clean `git status --porcelain` afterwards:

| Probe | Result |
|---|---|
| A: delete the `guardInitInHomeDir()` call from `RunE` (init.go:428) | rc=1 — `TestInitCommand_RefusesInHomeDirBeforeAnySideEffect` FAILS on all four subtests |
| B: `guardInitInHomeDir` returns nil unconditionally | rc=1 — `TestGuardInitInHomeDir/refuses_in_untracked_home` FAILS |
| C: `isUserHomeDir` made ancestor-tolerant (`\|\| strings.HasPrefix(dir, home)`) | rc=1 — `TestIsUserHomeDir` FAILS: `isUserHomeDir(".../001/project") = true, want false (subdir of home)` |
| D: **position probe** — move the guard call from first-in-`RunE` down to just above the `if !isGitRepo() && !hasExplicitBeadsDir` block (init.go:1207) | rc=1 — all four subtests FAIL with `init created .beads in the home directory despite refusing` / `init created .gitignore …` |

Probe D is the one I care about: the test file's central claim is that the guard's *position* is under test, and it is. Moving the guard ~780 lines later leaves `.beads/` and `.gitignore` behind in the home dir and the test catches it. That is a well-built test, and it answers maphew's earlier "tests cover only the helper" point properly.

I also measured the resolver rather than trusting the comment. On darwin, `user.Current().HomeDir` returns the real account home under **both** `CGO_ENABLED=1` and `CGO_ENABLED=0` with `HOME=/tmp/fakehome`, while `os.UserHomeDir()` returns `/tmp/fakehome` — so the guard does not key off `$HOME` here and cannot trip this repo's own `HOME=t.TempDir()` harnesses on macOS. The rewritten comment at init.go:324-332 is now honest about this being a preference, not a guarantee. And `resolveUserHomeDir` genuinely mirrors the existing precedent at `internal/beads/context.go:373-379`, so it is not a novel policy. Good.

---

## BLOCKING 1 — the guard is inert whenever root pre-run exports `BEADS_DIR` for you

`guardInitInHomeDir` treats `os.Getenv("BEADS_DIR") != ""` (init.go:372) as "the user explicitly opted in". It is not. Cobra's root `PersistentPreRunE` runs **before** `RunE`, and for a no-DB command like `init` it does this (main.go:1156-1157):

```go
beadsDir := selectedNoDBBeadsDir(cmd)
prepareSelectedNoDBContext(beadsDir)   // → os.Setenv("BEADS_DIR", beadsDir), main.go:599
```

`selectedNoDBBeadsDir` ends in a bare `return beads.FindBeadsDir()` (main.go:1148), which walks up from CWD and accepts any `.beads/` for which `hasBeadsProjectFiles` is true — and **`config.yaml` alone is enough** (`internal/beads/…`, `hasBeadsProjectFiles`). `~/.beads/config.yaml` is not hypothetical: it is this project's own **legacy user-level config path**, still read today — `internal/config/user_config_path.go:36`, `candidates.legacy = <home>/.beads/config.yaml`. So a user carrying the legacy global config gets `BEADS_DIR` set for them, and the guard fails open.

I reproduced it in-process rather than taking it on faith (throwaway probe test, since deleted; `resolveUserHomeDir` stubbed to a temp home, CWD = that home, no git repo, `~/.beads/config.yaml` present):

```
baseline refusal OK: refusing to 'bd init' directly in your home directory (/private/var/folders/…/001): …
selectedNoDBBeadsDir(initCmd) = "/private/var/folders/…/001/.beads"
BYPASS CONFIRMED: guard returned nil after PreRun exported BEADS_DIR=".../001/.beads"
```

**And the damage is not fully self-cancelling.** Yes, `hasExplicitBeadsDir` (init.go:1049) reads the same env, so the `git init` at init.go:1207 also skips — the "$HOME becomes a git repo" half is accidentally covered. But the *other* half the PR body promises to prevent — "can clobber an existing `~/CLAUDE.md`" — is **not** gated on `hasExplicitBeadsDir` at all:

- init.go:2140 — `if !stealth && !skipAgents { … addAgentsInstructions(resolvedAgentsFile, …) }` → writes `AGENTS.md` relative to CWD
- init.go:2175 — `if !stealth && !skipAgents && !isBareGitRepo() { setup.InstallClaudeProject(…); InstallCodexProject(); InstallCursorProject() }` → `CLAUDE.md`, `.claude/settings.json`, `.codex/`, `.cursor/`, all relative to CWD
- init.go:1170 — project-root `.gitignore` update, gated on `useLocalBeads`, which is true when the beads dir is CWD-local

So in the bypass state `bd init` in an untracked home still scaffolds over `~/CLAUDE.md`, `~/AGENTS.md`, `~/.claude/settings.json` — with no refusal. The same hole opens through `bd -C <dir> init` run from the home directory (`applyChangeDirSelection`, main.go:828, sets `BEADS_DIR` too): different beads dir, same CWD, same clobber.

This is the unresolved point in maphew's 2026-09-09 comment, and I reached it independently. Fix, in preference order:

1. Snapshot whether `BEADS_DIR` was **caller-supplied** — read it in `main()` before any `os.Setenv` and have the guard consult that snapshot — or
2. run the guard after `-C` selection but before discovery exports `BEADS_DIR`.

Either way, the guard should key off "the caller named a beads dir", not "the process env happens to have one".

Coverage this needs: a test through `rootCmd` execution (not a direct `initCmd.RunE(...)` call), in a home that already has `.beads/config.yaml`, asserting the refusal and asserting `AGENTS.md`/`CLAUDE.md`/`.claude` are not created. The four existing command-level subtests all call `RunE` directly, which is exactly why they miss this.

## BLOCKING 2 — a new `bd init` refusal that skips all three refusal conventions

`cmd/bd/init_safety.go:21-38` is explicit: *"Exit codes for init-safety refusals. Stable values so CI scripts can branch on them without grep'ing stderr"* — 10 `ExitRemoteDivergenceRefused`, 11 `ExitLocalExistsRefused`, 12 `ExitDestroyTokenMissing`. This PR adds a fourth refusal class to `bd init` and returns a bare `fmt.Errorf` → generic exit 1, so the only way to detect it is to grep stderr. That is precisely what the existing contract exists to avoid, and `cmd/bd/protocol/exit_codes_init_test.go` pins the others.

`docs/recovery/init-safety.md` states its own contract in its header: *"each named refusal in `bd init` and `bd dolt push`/`pull` points here to a labeled anchor with step-by-step recovery instructions"* — with a TOC of `init-force-refused` / `init-token-missing` / `init-local-exists`, each carrying its exit code. The new refusal gets no anchor, no `bd help init-safety` entry, and no note in `engdocs/adr/0002-init-safety-invariants.md`.

Please add a named code (`ExitHomeDirRefused` or similar) + a `docs/recovery/init-safety.md` anchor the message points at, and a protocol test alongside the other three.

## BLOCKING 3 — no CHANGELOG entry

`CHANGELOG.md` has a live `## [Unreleased]` section and every comparable merged change touches it (#6024, #4769, #5859, #5666, #6022, #6011, #5826, #5854, #5908, #5806 — all ten most recent `CHANGELOG.md` commits on main are feature/fix PRs). This is a user-visible behavior change: a command that used to succeed now refuses. It needs a `### Changed` entry under `## [Unreleased]`.

---

## NITs (not blocking)

- **NIT — the PR body is stale against its own head.** It says the tests are `TestInitRefusesHomeDirectory` and `TestInitAllowsHomeSubdirectory`; neither name exists anywhere in the tree (`grep -rn` over the head: no hits). The actual tests are `TestIsUserHomeDir`, `TestIsUserHomeDir_UnresolvableHome`, `TestGuardInitInHomeDir`, `TestInitCommand_RefusesInHomeDirBeforeAnySideEffect`. The body also quotes an error message (`this would turn $HOME into a git repository…`) that differs from the shipped one (`this would turn it into a git repository…`). Please refresh the description — the current tests are *better* than the ones advertised, and the body undersells them.
- **NIT — `--force` does not override, and nothing does.** `TestInitCommand_RefusesInHomeDirBeforeAnySideEffect/force` pins that, which I read as deliberate and I am fine with it. But the message's escape hatch ("set `BEADS_DIR` … if you really mean to initialize here") does not do what the user meant: with `BEADS_DIR` set, `hasExplicitBeadsDir` is true, so `git init` and the local-`.beads` semantics are skipped — it is a different init, not the one they asked for. Either say so in the message, or give the deliberate case a real flag.
- **NIT — exact-`$HOME` only.** `/`, `/Users`, `/home` and the like stay unguarded. Deliberately narrow per the PR body, and I agree with the scoping; noting it so the next reader does not assume broader coverage.
- **NIT — CI is stale.** The green matrix on this head is from 2026-08-14; main has moved 119 commits since. `mergeStateStatus` is CLEAN and it builds and vets clean here, but a re-run is warranted before merge regardless — and will be needed anyway once the blockers are addressed.

## What I did not run

I did not run the existing `cmd/bd` init suites for regressions. Rationale: I measured that `resolveUserHomeDir` is `$HOME`-independent on darwin under both CGO settings, so the repo's `HOME=t.TempDir()` harnesses (`init_embedded_test.go:92`, `proxied_integration_helpers_test.go:46`, `tests/regression/regression_test.go:329`, …) cannot trip the guard there, and the full CI matrix at this head was green. If the fix for BLOCKING 1 changes how the home is resolved, that reasoning has to be redone — flagging it now so it is not forgotten.

Happy to re-review promptly once the `BEADS_DIR` provenance fix lands. The core of this is good work.

harry-miller-trimble:
Thanks both — @maphew and @bee-ghosttrack reached the same bypass independently, from opposite directions, and you were both right. All three blockers addressed plus the nits, rebased onto current `main`. Pushed as `ea66cb26d`.

## BLOCKING 1 — `BEADS_DIR` provenance

Confirmed and fixed. The guard asked "does the process env have a `BEADS_DIR`", which by the time `RunE` runs is a question bd has usually answered for itself: root `PersistentPreRunE` → `selectedNoDBBeadsDir` → `prepareSelectedNoDBContext` → `prepareSelectedCommandContext`, which does `os.Setenv("BEADS_DIR", beadsDir)` at `main.go:599`. And `selectedNoDBBeadsDir` bottoms out in `beads.FindBeadsDir()`, which accepts any ancestor `.beads/` for which `hasBeadsProjectFiles` is true — and `config.yaml` alone is enough (`internal/beads/beads.go:688`). So the legacy user-level `~/.beads/config.yaml` failed the guard open for exactly the long-standing users most likely to have `~/CLAUDE.md` worth not clobbering.

@bee-ghosttrack's point about the damage not being self-cancelling is the one that made this worth blocking on, and I checked it rather than taking it on faith: `hasExplicitBeadsDir` gates the `git init`, but `addAgentsInstructions` (`init.go:2140`) and `InstallClaudeProject`/`InstallCodexProject`/`InstallCursorProject` (`init.go:2175`) are gated only on `!stealth && !skipAgents` and write relative to CWD. So the bypass still scaffolded over `~/CLAUDE.md`, `~/AGENTS.md` and `~/.claude/settings.json`.

Fixed with option 1 from both of your lists — snapshot the caller-supplied value:

```go
// beadsDirFromCaller is BEADS_DIR as the process INHERITED it, captured at
// package initialization — which the language guarantees runs before main, and
// therefore before any bd code exports a BEADS_DIR of its own.
var beadsDirFromCaller = os.Getenv("BEADS_DIR")
```

I preferred it over "run the guard before discovery exports it" because it keys off what the *caller* said rather than on the pre-run's ordering staying put — and because it also covers `bd -C <dir> init`, which @bee-ghosttrack flagged and which the ordering fix alone would not have, since `applyChangeDirSelection` runs earlier still and names a beads dir without moving the working directory the scaffolding lands in.

**Coverage, exactly as asked.** `TestInitThroughRootCmd_RefusesInHomeDirWithDiscoveredBeadsDir` goes through `rootCmd.Execute` rather than `initCmd.RunE`, in a home carrying only `.beads/config.yaml`, and asserts `AGENTS.md`/`CLAUDE.md`/`.claude` are not created. Two details worth naming:

- it asserts the **precondition** first — that pre-run really does still discover a `config.yaml`-only `.beads` — so if that ever changes, the test says it has stopped reproducing the bypass instead of passing for the wrong reason;
- it borrows @maphew's `--backend` sentinel as a circuit breaker, so a future fail-open stops at backend selection rather than building a real store in someone's home directory.

Against the old code it reproduces your observation verbatim:

```
init_home_guard_test.go:314: init ran past the home guard and got as far as
backend selection: unknown backend "review4795-home-guard-sentinel"
```

Plus a unit case, `refuses_when_bd_exported_beads_dir_for_itself`, pinning the distinction directly.

## BLOCKING 2 — refusal conventions

`ExitHomeDirRefused = 13` joins 10/11/12, the message prints and the error carries only the code (the `ExitLocalExistsRefused` pattern at `init.go:885`), `docs/recovery/init-safety.md` gains an `init-home-refused` anchor — TOC entry, symptom, why, three recovery paths, and a script-branching snippet — which the message points at, and `bd help init-safety` lists the code.

**One deliberate omission, flagging it rather than quietly skipping it.** I did not add a `cmd/bd/protocol` case. That file opens by saying §E3 freezes 10/11/12/130 as scriptable API and that *"New stable codes require a spec revision"*, so a test asserting 13 there would be claiming wire status the spec hasn't granted, in a corpus downstream consumers vendor. The constant plus the docs anchor give the refusal the same in-repo scriptability without that claim — and the exit code is asserted at the `cmd/bd` level instead. If you'd rather 13 become part of v0, say so and I'll add the §E3 case in the same shape as the other three.

## BLOCKING 3 — CHANGELOG

Added under `## [Unreleased]` → `### Changed`, citing #4635, with the scope limits stated.

## NITs

- **Stale PR body** — rewritten. You're right that it advertised tests that don't exist (`TestInitRefusesHomeDirectory`, `TestInitAllowsHomeSubdirectory`) and quoted a message that isn't the shipped one. The new body describes the real four, the placement argument, the provenance rule, and the exit code.
- **The escape hatch didn't do what the user meant** — good catch, and I took "say so in the message" rather than adding a flag. It now reads: *"this is a DIFFERENT init: an explicit BEADS_DIR skips the git init and the working-directory-local .beads, so nothing is written here"*. The recovery doc spells it out at more length, including the case where you actually do want your home tracked (`git init` it yourself first, deliberately, and the guard stands aside).
- **Exact-`$HOME` only** — deliberate, and now said in the PR body and the recovery doc so the next reader doesn't assume broader coverage.
- **Stale CI** — the rebase onto current `main` takes care of it.

## Local gates

`gofmt -l`, `go build ./...`, `go vet ./cmd/bd/...` clean; `scripts/generate-cli-docs.sh --check` → `PASS`; `go test ./test/docsync/` green; `go test ./cmd/bd/ -run 'Init|Home|Guard|Safety'` green (175s). The rebase took conflicts in `CHANGELOG.md` and in `docs/recovery/init-safety.md`'s TOC — main reworded two entries there — both resolved by keeping main's wording and adding the new entry alongside.

@bee-ghosttrack — on your "what I did not run": the fix does not change how the home is resolved (`resolveUserHomeDir` is untouched), only what counts as caller opt-in, so your reasoning about the `HOME=t.TempDir()` harnesses still holds. The full `cmd/bd` init suites pass here regardless.

_claude-code-opus-5-unknown-reasoning on behalf of Harry Miller_

bee-ghosttrack (approved):
Round 2 from me (bee). Re-reviewed `ea66cb26dc` in a throwaway worktree off the PR head. It is rebased onto current `main` (0 behind, 2 ahead), merges clean, and the three blockers are cleared with tests that hold under mutation. **Approving.** Two non-blocking nits at the end, one of which is worth a look before merge because it is cheap now and not later.

## What I ran

| Command | Result |
|---|---|
| `git merge-tree --write-tree origin/main ea66cb26dc` · `git diff --check` · `go build ./...` · `go vet ./cmd/bd/` | all rc=0 |
| `go test ./cmd/bd -run 'TestIsUserHomeDir\|TestGuardInitInHomeDir\|TestInitCommand_RefusesInHomeDirBeforeAnySideEffect\|TestInitThroughRootCmd' -count=1 -v` (cgo) | rc=0 — PASS 5/5, 5.6s |
| same, `CGO_ENABLED=0 -tags gms_pure_go` | rc=0 — PASS, 1.9s |
| `go test ./cmd/bd/protocol/ -count=1` | rc=0 — PASS, 389s. The §E3 corpus is untouched by 13 and still green |

Diff is the seven files the PR now claims. Working tree clean after every probe.

## Mutation probes

| Probe | Result |
|---|---|
| **E** — guard reads `os.Getenv("BEADS_DIR")` again, i.e. the Round-1 bypass reintroduced | rc=1 — `TestGuardInitInHomeDir/allows_when_the_caller_supplied_beads_dir`, `/refuses_when_bd_exported_beads_dir_for_itself`, **and** `TestInitThroughRootCmd_RefusesInHomeDirWithDiscoveredBeadsDir` FAIL ✅ |
| **A** — delete the `guardInitInHomeDir()` call from `RunE` | rc=1 — `TestInitCommand_RefusesInHomeDirBeforeAnySideEffect/plain` FAILS ✅ |
| **G** — `return fmt.Errorf("refused")` instead of the coded `exitError` | rc=1 — all three command-level tests FAIL on the exit code ✅ |

Probe E is the one this round is about, and the through-`rootCmd` test fails under it, so the regression it is named after is actually held. The precondition assertion (pre-run really does still discover a `config.yaml`-only `.beads`) means it also cannot go green for the wrong reason if discovery ever changes. Good test.

## Blockers

1. **`BEADS_DIR` provenance — cleared.** The package-init snapshot is the right lever: it keys off what the caller said rather than on the pre-run's ordering staying put, and it covers `bd -C <dir> init` for free, which the ordering fix would not have. I also checked how the coded error surfaces: `main.go:2233` sends an `exitError` straight to `os.Exit(code)` and prints nothing else, so the message the guard writes is the whole of stderr — the same shape as `ExitLocalExistsRefused`. No stray `Error: exit code 13` line.
2. **Refusal conventions — cleared.** Named constant, `bd help init-safety` entry, `docs/recovery/init-safety.md#init-home-refused` with the script-branching snippet. The protocol-corpus omission is the right call by my reading too: `exit_codes_init_test.go` says new stable codes need a spec revision, so a §E3 case for 13 would be the PR granting itself wire status. In-repo scriptability via the constant and the docs anchor is what I asked for. If maintainers want 13 in v0, that is a spec change first.
3. **CHANGELOG — cleared.** Under `## [Unreleased]` → `### Changed`, scope limits stated.

The body is refreshed and names the real four tests. The escape-hatch message now says what `BEADS_DIR` actually does. Both nits from Round 1 taken as I would have.

## Nits (non-blocking)

- **N1 — exit 13 already has a twin in this binary.** `cmd/bd/update.go:823` has `const ExitGuardMismatch = 13` — `bd update`'s "guard mismatch, racer won" code. Different command, so no script that branches per command can confuse them, and nothing in the protocol catalog claims the numbers are global. But "bd exited 13" now means two things in one binary's documentation, and a future generic exit-code table has a collision baked in. 14 is unused (so are 15 and 16). Renumbering is a four-file constant swap today and a breaking change the day after it ships. Your call and maphew's — I am not holding the PR on it, and if you take 14 it does not need another round from me.
- **N2 — the ADR table.** `engdocs/adr/0002-init-safety-invariants.md` § "Exit codes (stable, grep-safe)" lists 10/11/12 only. One row for the new code keeps the ADR from going stale on the day this lands.

On your "what I did not run" reply: agreed, `resolveUserHomeDir` is untouched so the `HOME=t.TempDir()` harness reasoning from Round 1 stands.

harry-miller-trimble:
Merged current main at `6ea8da6246`. There were two changes:

- **`cmd/bd/init_safety_help.go`**: the one conflict. Main extended the exit-12 description ("also returned when the interactive --discard-remote typed-token confirmation is declined"), and this PR adds exit `13` directly below it. The resolution keeps both. Exit code `13` is still unused on main.
- **CHANGELOG**: the entry moved under `## [Unreleased]` / `### Changed`. Without the move it would merge cleanly into the released `[1.3.0]` section (#6643).

No other code changed. Build, vet and the home-guard tests (`TestGuardInitInHomeDir`, `TestInitCommand_RefusesInHomeDirBeforeAnySideEffect`, `TestInitThroughRootCmd_RefusesInHomeDirWithDiscoveredBeadsDir`, `TestIsUserHomeDir*`) pass on the merged tree.

_claude-code-opus-5-5-unknown-reasoning on behalf of Harry Miller_

bee-ghosttrack (approved):
Re-read at `6ea8da6246` against my approval at `ea66cb26dc`: outside CHANGELOG.md the PR's own diff against main is byte-identical, and the entry now sits under `## [Unreleased]` (#6643). Approval stands. CI is still pending on this head; it's for the maintainer to confirm green before merging.

PR #4798, opened 2026-07-14
fix(storage): carry closed_by_session through the domain/UOW writer (#4662)

### LINKED ISSUE #4662: PromoteFromEphemeral drops closed_by_session when promoting closed wisps
## Summary

Gastown PR gastownhall/gastown#4414 is blocked because compact wisp promotion delegates to the Beads SDK `PromoteFromEphemeral` path. PR Sheriff evidence from the Gastown blocker `gt-cbs1` shows the SDK promotion path can drop `closed_by_session` for closed wisps.

## Evidence

Checked Beads versions:

- v1.0.5
- v1.1.0
- main pseudo `v1.1.1-0.20260708062845-1c1f2501efbb`

The SDK defines `ClosedBySession` and close/update paths write `closed_by_session`, but promotion omits the column from the select/scan and insert/upsert path:

- `IssueSelectColumns` / `ScanIssueFrom`
- `InsertIssueIntoTable` / `issueUpsertColumns`

As a result, `PromoteFromEphemeral` hydrates a zero `ClosedBySession` and inserts/promotes without preserving the original closure session metadata.

## Expected

`PromoteFromEphemeral` should preserve `closed_by_session` when promoting closed wisps.

## Acceptance Criteria

- Promotion preserves `closed_by_session` for closed wisps.
- Regression test covers closed wisp promotion preserving `closed_by_session`.
- Gastown can consume the fix via normal release/pseudo dependency update.
- No Gastown raw SQL, vendoring, or `go.mod replace` workaround is needed.

harry-miller-trimble:
Closes #4662

> **Scope changed after rebase (2026-08-14).** The original diff added `closed_by_session` to the shared select/scan/insert column sets. All of that landed upstream independently in #5191, so those hunks are dropped. What remains — and what is still broken on current `main` — is the second writer.

## Problem

`db.insertIssueRow` (`internal/storage/domain/db/issue.go`), the statement behind proxied create, batch create and import, has no `closed_by_session` in its INSERT column list, its VALUES, or its `ON DUPLICATE KEY UPDATE`.

The classic writer (`issueops.insertIssueIntoTable`) does carry it, and the shared scanner (`issueops.ScanIssueFrom`) reads it back. So the two writers disagree, and the disagreement is silent: the schema defaults the column to `''`, the insert succeeds, the read succeeds, and only the closing-session provenance is gone.

That is the GH#4662 data loss — `PromoteFromEphemeral` reads a wisp and re-inserts it, dropping the original closing session — still live, just relocated to the writer the proxied stack uses.

## Fix

Add `closed_by_session` to `insertIssueRow`'s insert columns, values, and upsert assignments, with a note at the chokepoint on why the two writers' column lists have to stay in step.

## Tests

Three cases in the domain/db suite (real Dolt via the test container):

- `Insert/PreservesClosedBySession` — issues table;
- `Insert/PreservesClosedBySessionOnWisp` — wisps table, since promotion reads the wisp side first;
- `Insert/UpsertRewritesClosedBySession` — the `ON DUPLICATE KEY UPDATE` half, which import re-enters; a column in the INSERT but absent from the upsert would carry the value on first write and keep a stale one forever after.

All three fail against `main` and pass with the fix:

```
--- FAIL: TestDomainDB/TestIssueSQLRepository/Insert/PreservesClosedBySession
        closed_by_session must survive a domain/UOW insert
--- FAIL: TestDomainDB/TestIssueSQLRepository/Insert/PreservesClosedBySessionOnWisp
        closed_by_session must survive a domain/UOW wisp insert
--- FAIL: TestDomainDB/TestIssueSQLRepository/Insert/UpsertRewritesClosedBySession
        the upsert must rewrite closed_by_session, not keep the stale value
```

Plus the conformance case `PromoteClosedWispPreservesSession`, kept from the original commit: it runs on every backend and is the end-to-end regression for the reported bug (close a wisp with a session, promote, assert the session survives). Confirmed passing under embedded Dolt.

### DIFF
--- a/backend/conformance/portable.go
+++ b/backend/conformance/portable.go
@@ -42,6 +42,7 @@ func RunPortableMethods(t *testing.T, factory Factory) {
 	t.Run("ImportIssueComment", func(t *testing.T) { testImportIssueComment(t, factory) })
 	t.Run("AddIssueCommentBurstOrder", func(t *testing.T) { testAddIssueCommentBurstOrder(t, factory) })
 	t.Run("PromoteFromEphemeral", func(t *testing.T) { testPromoteFromEphemeral(t, factory) })
+	t.Run("PromoteClosedWispPreservesSession", func(t *testing.T) { testPromoteClosedWispPreservesSession(t, factory) })
 	t.Run("UpdateIssueID", func(t *testing.T) { testUpdateIssueID(t, factory) })
 	t.Run("DeleteIssuesBySourceRepo", func(t *testing.T) { testDeleteIssuesBySourceRepo(t, factory) })
 	t.Run("CreateIssuesWithFullOptions", func(t *testing.T) { testCreateIssuesWithFullOptions(t, factory) })
@@ -574,6 +575,33 @@ func testPromoteFromEphemeral(t *testing.T, f Factory) {
 	// Second promote fails (no longer a wisp).
 	if err := s.PromoteFromEphemeral(c, "test-w", "a"); err == nil {
 		t.Error("double promote: want error, got nil")
+	}
+}
+
+// testPromoteClosedWispPreservesSession verifies that promoting a *closed*
+// ephemeral wisp preserves its closed_by_session provenance. Before the fix
+// (GH#4662), closed_by_session was absent from the shared issue select/scan
+// and insert column set, so promotion (read wisp -> insert issue) hydrated a
+// zero value and dropped the original closing session.
+func testPromoteClosedWispPreservesSession(t *testing.T, f Factory) {
+	s := f(t)
+	c := ctx()
+
+	must(t, s.CreateIssue(c, withDefaults(&types.Issue{ID: "test-cw", Title: "Closed wisp", Ephemeral: true}), "a"))
+	must(t, s.CloseIssue(c, "test-cw", "done", "a", "sess-xyz"))
+
+	must(t, s.PromoteFromEphemeral(c, "test-cw", "a"))
+
+	got, err := s.GetIssue(c, "test-cw")
+	must(t, err)
+	if got.Ephemeral {
+		t.Error("promoted issue still Ephemeral")
+	}
+	if got.Status != types.StatusClosed {
+		t.Errorf("promoted issue status = %q, want %q", got.Status, types.StatusClosed)
+	}
+	if got.ClosedBySession != "sess-xyz" {
+		t.Errorf("closed_by_session after promote = %q, want %q (GH#4662)", got.ClosedBySession, "sess-xyz")
 	}
 }
 
--- a/internal/storage/domain/db/issue.go
+++ b/internal/storage/domain/db/issue.go
@@ -754,6 +754,14 @@ func insertIssueRow(ctx context.Context, runner Runner, table string, issue *typ
 	if err := types.CheckFieldLen("owner", issue.Owner); err != nil {
 		return err
 	}
+	// The column list must stay in step with the classic writer
+	// (issueops.insertIssueIntoTable) — the shared scanner hydrates whatever
+	// either writer stored, so a column present there and missing here is not
+	// a harmless omission but a silent, writer-dependent data loss. That is
+	// what closed_by_session was: the schema defaults it to '', so a proxied
+	// create/import read back an empty closing session while the same issue
+	// created through the classic path kept it (GH#4662).
+	//
 	// Stamp a fresh non-zero row_lock at create, exactly like the classic
 	// insertIssueIntoTable (issueops/helpers.go). Without it a proxied-server
 	// (uow) create leaves row_lock at the schema DEFAULT 0, so the row's
@@ -767,7 +775,7 @@ func insertIssueRow(ctx context.Context, runner Runner, table string, issue *typ
 			created_at, created_by, owner, updated_at, started_at, closed_at, external_ref, spec_id,
 			compaction_level, compacted_at, compacted_at_commit, original_size,
 			sender, ephemeral, no_history, wisp_type, pinned, is_template,
-			mol_type, work_type, source_system, source_repo, close_reason,
+			mol_type, work_type, source_system, source_repo, close_reason, closed_by_session,
 			event_kind, actor, target, payload,
 			await_type, await_id, timeout_ns, waiters,
 			due_at, defer_until, metadata,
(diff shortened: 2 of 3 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the shared-seam fix is the right shape; one more writer needs it, and the rebase is semantically loaded.

**Blocker 1 — the proxied/domain writer still drops `closed_by_session`.** Verified at the PR head: `IssueSQLRepository.Insert` uses its own `insertIssueRow`, whose column list contains zero occurrences of `closed_by_session` — not in INSERT, not in the upsert. The schema defaults it to `''`, and the newly shared scanner now faithfully reads back the empty value this writer discarded. This is live wiring (UOW installs this repository; proxied create invokes it). Add the column to `insertIssueRow`'s insert/values/upsert with an insert→get regression through the domain/UOW stack.

**Blocker 2 — the rebase onto current main is semantic, not textual.** Since the branch point, main moved leases out of issue/wisp rows into a joined overlay (migration 0055 removed columns the PR projection still selects) and added `row_lock`/`RowVersion` CAS hydration. merge-tree conflicts in all four core files; taking either side wholesale loses something (the PR's projection queries removed columns; the PR's scanner side loses `row_lock`). Reconciliation = `closed_by_session` added to current main's `IssueBaseColumns` + scan destination alongside `row_lock`/lease overlay, and the parity fixture becomes 50 columns.

Nit: two test comments asserting `closed_by_session` is absent from `IssueSelectColumns` are made false by this PR — update them.

The PR-head internals are otherwise consistent (49/49 select/scan, 47/47 insert/args, `updated_at` last in the stale-guarded upsert) and the promotion regression is welcome — the wisp-promotion data-loss bug is real and worth landing once the two writers agree.

harry-miller-trimble:
Thanks — both blockers addressed. The rebase changed what this PR should be, so the diff is now much smaller and points at the part that is still broken.

**Blocker 2 — the rebase is semantic, not textual: resolved by dropping, not merging.** You were right that taking either side wholesale loses something. It turns out there's a third option: the entire select/scan/insert half of this PR **already landed upstream** in #5191 ("one guarded issue-lifecycle surface"). On current `main`:

- `sqlbuild.IssueBaseColumns` (and `IssueBaseColumnsLite`) already list `closed_by_session`, alongside `row_lock`/`storage_class` and with leases moved to the `LeaseSelectColumns` overlay;
- `issueops.ScanIssueFrom` already scans it into `issue.ClosedBySession`;
- `issueops.insertIssueIntoTable` and `issueUpsertColumns` already carry it;
- the scan-parity fixture is already 52 columns with `closed_by_session` in it.

So rather than reconciling those four files against the lease overlay, I dropped those hunks. Nothing is lost and there is no merge to get subtly wrong.

**Blocker 1 — the proxied/domain writer still drops it: confirmed and fixed, and it's the whole PR now.** `db.insertIssueRow` (`internal/storage/domain/db/issue.go`) — the statement behind proxied create, batch create and import — has no `closed_by_session` in its INSERT column list, its VALUES, or its `ON DUPLICATE KEY UPDATE`. Combined with the now-shared scanner, that is exactly the failure you described: the schema defaults the column to `''`, the insert succeeds, the read succeeds, and only the provenance is gone. The GH#4662 data loss is still live on main — it just moved writers.

Fixed in all three places, with a note at the chokepoint on why the two writers' column lists have to stay in step.

**Regression coverage, as requested — insert → get through the domain/UOW stack.** Three cases in the domain/db suite (real Dolt via the test container):

- `Insert/PreservesClosedBySession` — issues table;
- `Insert/PreservesClosedBySessionOnWisp` — wisps table, since promotion reads the wisp side first;
- `Insert/UpsertRewritesClosedBySession` — the `ON DUPLICATE KEY UPDATE` half, which import re-enters; a column in the INSERT but absent from the upsert would carry the value on first write and keep a stale one forever after.

Verified against main (fix reverted, tests kept):

```
--- FAIL: TestDomainDB/TestIssueSQLRepository/Insert/PreservesClosedBySession
        closed_by_session must survive a domain/UOW insert
--- FAIL: TestDomainDB/TestIssueSQLRepository/Insert/PreservesClosedBySessionOnWisp
        closed_by_session must survive a domain/UOW wisp insert
--- FAIL: TestDomainDB/TestIssueSQLRepository/Insert/UpsertRewritesClosedBySession
        the upsert must rewrite closed_by_session, not keep the stale value
```

All three pass with the fix. The conformance case `PromoteClosedWispPreservesSession` from the original commit is kept — it runs on every backend and is the end-to-end GH#4662 regression; confirmed passing under embedded Dolt (`TestConformance/Portable/PromoteClosedWispPreservesSession`).

**Nit — test comments asserting `closed_by_session` is absent from `IssueSelectColumns`:** no longer applicable, those comments are gone from main along with the rest of #5191.

Local runs: `./internal/storage/domain/db/` (Dolt container), `./internal/storage/issueops/`, and the embedded-Dolt conformance suite all pass; `gofmt` and `go vet` clean.

maphew (approved):
Re-review of the reshaped PR: both blockers from my 2026-07-24 review are resolved, and the reshape is a genuine improvement rather than an evasion. Approving.

- **Blocker 1** (domain/UOW writer drops `closed_by_session`): fixed exactly as asked in `internal/storage/domain/db/issue.go` - column in the INSERT list, placeholder row, upsert assignment `closed_by_session = VALUES(closed_by_session)`, and the arg in the right slot (I checked the 49/49/49 column/placeholder/arg alignment). The insert-then-get regressions are present as three domain-suite tests shown failing on main, and `UpsertRewritesClosedBySession` covering the import re-entry path is a catch a naive fix would have missed.
- **Blocker 2** (semantic rebase onto the moved column sets): resolved by a third option my review didn't anticipate, and the right one - the conflicting select/scan/insert hunks already landed on main via #5191, so you dropped them instead of merging. Verified against current main that `sqlbuild`, `ScanIssueFrom`, and the classic writer/upsert all carry the column now. The #4662 data loss survived #5191 precisely by relocating into the domain/UOW writer, which this PR now fixes surgically - hence the much smaller diff, which is the correct size for what remains.
- The nit (test comments asserting the column's absence) is moot: #5191 removed those comments along with the old column-set state.

The kept conformance case (`PromoteClosedWispPreservesSession`) gives the end-to-end #4662 regression on every backend, which is the durable protection here. CI is fully green across the matrix and the PR is mergeable.

Process note: I stepped down as maintainer on 2026-08-10 and can no longer land this. @csells @julianknutsen - ready for merge from my side.

PR #4804, opened 2026-07-15
perf(schema): skip the migration lock and pass when the database is current

atbrace:
## Problem

Every store open in server mode runs `MigrateUpWithLock`, which unconditionally takes the per-database `GET_LOCK` before discovering — in steady state, always — that there is no migration work. A fleet of concurrent bd processes therefore serializes **every command's open** on one advisory lock. Observed live on a shared Dolt sql-server under multi-agent load (load avg 64 on a 6-core host, Dolt pegged ~77% CPU): `SHOW FULL PROCESSLIST` showed connections parked on `SELECT GET_LOCK(?, ?)` plus the per-open `INSERT IGNORE` dolt_ignore seed writes, all to conclude no-op.

## Fix

Add a strictly read-only currency probe before the lock: the canonical dolt_ignore seed patterns are all present (`doltIgnoreSeedCurrent`, the read-only counterpart of `seedDoltIgnorePatterns`) **and** `migrationWorkNeeded` is false. When current, `MigrateUpWithLock` returns without touching `GET_LOCK`, without the 5 seed `INSERT IGNORE`s, and without the pass.

Anything less than fully-current — including a probe error — falls through to the locked pass unchanged (double-checked locking): it re-checks under the lock and surfaces genuine errors itself. Concurrent processes taking the read-only fast path together need no coordination, and a process racing a real migration sees not-current and queues on the lock exactly as before. Both call sites (`dolt.initSchemaOnDB`, `uow.dolt_sql_provider`) get the fast path.

## Benchmarks

New `BenchmarkMigrateUpWithLockCurrent[Parallel]` (real throwaway `dolt sql-server`, database migrated to current — the per-command open cost). i5-8500B, 8 threads, dolt 2.1.10:

| bench | main | this PR |
|---|---|---|
| sequential | 15.80 ms/op | 5.44 ms/op (2.9×) |
| 8-way parallel | 16.81 ms/op | 0.93 ms/op (18×) |

The parallel row is the finding: on main, 8-way parallel is **no faster than sequential** — fully serialized on the advisory lock. With the fast path it scales with parallelism, which is exactly the fleet shape (N agents × several bd commands/sec).

## Tests

- `TestMigrateUpWithLockSkipsLockWhenCurrent` — sqlmock ordered expectations prove no GET_LOCK / seed / commit is issued when current
- `TestMigrateUpWithLockReseedsUnderLockWhenSeedMissing` — under-seeded DB skips the fast path and heals under the lock as before
- `TestMigrateUpWithLockFallsThroughOnProbeError` — probe failure falls through to the locked pass instead of failing the open
- Existing `TestMigrateUpWithLockUsesDatabaseScopedLockOnly` updated for the pre-lock probe; full schema package green (141 tests)

Note: `internal/storage/uow` `TestNewDoltServerUOWProvider_ConcurrentInstantiation` fails identically on unmodified main in my environment (fresh-DB concurrent migration racing the 5s lock timeout on a loaded box) — pre-existing, not introduced here; anecdotally this PR shortens it (254s → 19s) since only the first winner holds the lock for the full pass.

## Update: pass-completion sentinel (review response, 40c7773)

The paired review found the probe's inputs are all satisfied **mid-pass** by a running migration's per-step commits — before `rekeyDependencyIDs` and the final staging commit have run — so a concurrent opener could fast-path into another process's half-finished pass. (`rekeyAuxRowIDs` was never exposed: it only runs while ignored migration 9 is pending, which keeps `migrationWorkNeeded` true and the probe blocked.)

Fixed with a **pass-completion sentinel**: a `migration_pass_complete` row in `local_metadata` (dolt-ignored — database-local on a shared server, the probe's exact visibility scope). Deleted before the pass's first mutation, restamped with the binary's `main/ignored` latest versions (monotonic compare) only after a fully successful `MigrateUp`. The probe requires it first, keeping the cursor/seed checks for out-of-band regressions. A crashed pass leaves the sentinel absent — every open takes the locked pass until one completes, the safe direction. Pre-sentinel databases pay one locked no-op pass to earn the stamp.

New tests: `TestMigrationProbeRefusesFastPathMidPassOnRealServer` (real-server upgrade-shaped pass paused between the last cursor commit and the rekey tail; proves the window is real, asserts the probe refuses — fails pre-fix), `TestMigrateUpWithLockSentinelGatesFastPathOnRealServer` (sentinel-absent state queues on `GET_LOCK`, restamps under it), plus sqlmock sentinel-missing/stale and clear-before-pass/stamp-after-final-commit ordering. Benchmarks unchanged within noise (sentinel = one extra single-row PK read).


maphew (changes requested):
Cross-vendor paired agent review (Claude Opus + Codex gpt-5.6-sol, independent traces; the vendors disagreed and the contradiction was adjudicated against the code before posting). Verdict: **needs changes** — one real interleaving hazard; everything else about the PR is strong.

**Blocker (adjudicated and confirmed at the PR head): the fast path can pass while another process's migration pass is still running its tail.** The unlocked probe `migrationStateCurrent` = seeds present + `!migrationWorkNeeded`, and `migrationWorkNeeded` checks exactly: both cursor tables at-latest, content-hash columns present, and the custom-statuses/types backfill. But `MigrateUp`'s pass does more than that, in this order: `mainSource.migrate` (which inserts the cursor row and runs `commitMigrationStep` **per step** — plain autocommitted statements, immediately visible to other sessions on a shared server), then the custom-tables backfill, then `rekeyDependencyIDs` (#4259 deterministic dep-key convergence), then `rekeyAuxRowIDs`, then the ignored source and the final staging commit. Take an upgrade whose ignored source is already at latest: the moment the last numbered migration's per-step commit lands and the backfill finishes, every probe check passes — while the dep/aux rekey rewrites are still in flight. Pre-PR, a second opener waited on `GET_LOCK` for the whole pass; post-PR it proceeds lock-free mid-rekey, free to read half-rekeyed dependencies or write rows the rekey is concurrently rewriting — the merge-divergence class the rekey exists to eliminate. The three new tests (steady-current, missing-seed, probe-error) don't cover this interleaving. Fix shape: the probe must not be satisfiable until the *complete* pass has finished — e.g. a pass-completion sentinel committed at the end of `MigrateUp` that `migrationStateCurrent` requires, or fold the rekey-pending checks into `migrationWorkNeeded`. A concurrency regression test (fault hook pausing between cursor commit and rekey while a second opener probes) would pin it.

**What holds up (verified):** the fast path skips nothing on the gate side — `CheckRemoteMigrateGate` runs before `MigrateUpWithLock` in server mode and returns nil precisely for the no-pending case, so the version floor can't be bypassed; probe errors correctly fall through to the locked pass (`probeErr == nil && current` guard, with a test); missing-table maps to fall-through; embedded mode is untouched; and the real dolt-sql-server benchmarks capture the fleet-parallel motivation.

**Duplicate:** draft #4764 implements the same idea and edits the same lines — they cannot both land. This PR is the stronger of the pair (its probe-error fall-through vs #4764's fail-the-open, missing-table resilience, and real-server tests), so once the blocker above is fixed we'd recommend landing this one and closing #4764 with credit. Note #4764 has the same tail-interleaving hazard.

Nits: the three fast-path tests could be table-driven; `doltIgnoreSeedCurrent` scans all of `dolt_ignore` rather than a bounded `WHERE pattern IN (...)` (immaterial — tiny table); `startBenchDoltServer`'s initial `sql.DB` lacks cleanup on setup failure before the first defer.

atbrace:
Thanks for the adjudicated review — the blocker is real, and it's now fixed at the head (40c7773).

**Confirming the hazard, with one correction to its scope.** Verified against source and reproduced live: with an upgrade-shaped database (main cursor one behind, ignored source already at latest — the common release shape), a second session probing at the moment the last numbered migration's per-step commit lands sees every probe input satisfied while the pass tail has not run. One detail differs from the review: `rekeyAuxRowIDs` was never exposed — it only runs while ignored migration 9 (its clone-local marker) is *pending*, and any pending ignored version makes `migrationWorkNeeded` true, which blocks the probe for the whole aux-rekey window. The genuinely exposed tail is `rekeyDependencyIDs` (gated on nothing but "work was needed") plus the staging/final-commit tail — where a lock-free opener could also trip the changed-signature guard or have its writes swept into the pass's force-staged commit. Same blocker, slightly narrower window; the fix covers the whole pass either way.

**Fix shape: the pass-completion sentinel** (your first suggested shape). A `migration_pass_complete` row in `local_metadata` — dolt-ignored, so database-local on a shared server, which is exactly the probe's visibility scope; precedent is the aux-rekey crash sentinel in the same table. `migrateUp` deletes it before the pass's first mutation; `MigrateUp` restamps it only after a fully successful return, recording the binary's `main/ignored` latest versions (compared monotonically, so a newer binary's completed stamp is never downgraded by an older binary and mixed-version fleets don't ping-pong). The probe requires the sentinel *first* and keeps the cursor/seed checks ANDed in, since the sentinel can't see out-of-band regressions (a table copy that rewinds cursors leaves `local_metadata` untouched). "Current" is now unsatisfiable while any pass is in flight — or died mid-flight, which degrades to every open taking the locked pass until one completes and restamps: the safe direction. Pre-sentinel databases pay one locked no-op pass to earn the stamp, then fast-path again. I didn't take the fold-into-`migrationWorkNeeded` shape because detecting pending dependency rekeys read-only means scanning `dependencies` and computing `depid` per row on every open — it would give back much of what the fast path buys.

The pass-start clear deliberately does **not** create `local_metadata` (bare `DELETE`, errno-1146-tolerant): the dirty-table guards grep pending migration SQL — main `0029`/`0030` and ignored `0001` all reference the table — against the pre-pass dirty set, and materializing a perpetually-untracked ignored table before those guards run would surface it to them on databases that never had it.

**The requested concurrency regression test** is `TestMigrationProbeRefusesFastPathMidPassOnRealServer`: a real dolt sql-server, a real upgrade-shaped pass, paused by the existing per-step fault hook exactly between the last cursor commit and the rekey tail, while a second session (a) proves the legacy inputs are all current — the window is genuinely reproduced, not assumed — and (b) asserts `migrationStateCurrent` refuses. On the pre-fix probe it fails with `migrationStateCurrent() = true mid-pass`. `TestMigrateUpWithLockSentinelGatesFastPathOnRealServer` covers the state side: sentinel absent (what a prober sees mid-pass) queues on `GET_LOCK` — proven by timing out against a held lock — then restamps under the lock and fast-paths again. sqlmock ordering tests pin sentinel-missing/stale → locked pass, and that a work-needed pass issues the `DELETE` before its first migration statement and the `REPLACE` only after the final `DOLT_COMMIT`.

**Benchmarks** are unchanged within noise (i5-8500B, dolt 2.1.10, 300x, 4 interleaved runs): parallel 1.6–3.9 ms/op with the sentinel vs 2.0–3.9 ms/op at the previous head — the sentinel adds one single-row PK read to the probe's 7 queries; both remain far below main's fully-serialized ~16.8 ms/op.

**Nits:** `startBenchDoltServer` is generalized to `testing.TB` for the new tests, and its setup-phase `*sql.DB` now registers cleanup before anything can fail. Left the three fast-path tests non-table-driven and the `dolt_ignore` full scan as-is per the immaterial call, but happy to change either. Agreed on the #4764 duplicate resolution — same hazard there, so this supersedes it once merged.

maphew:
Re-verified at `40c7773`, adversarially: the sentinel design closes the window. The load-bearing ordering — DELETE the sentinel before the pass's first mutation, stamp only after the final `DOLT_COMMIT` — held up against every mid-pass probe timeline we could construct; a prober at the old exploit point (last numbered migration's per-step commit landed, tail not run) now sees the sentinel absent and falls to the lock. Crash mid-pass fails safe (sentinel stays absent, everyone locks until a pass completes), the ANDed cursor/seed checks make the stamp necessary-but-not-sufficient, `>=` keeps mixed-binary fleets monotonic, and the nonlocal-table visibility reasoning is right. This is the correct shape for a fast path: cheap, layered, and never trusted alone.

**Your scope correction is accepted, with the review's error acknowledged**: `rekeyAuxRowIDs` early-returns unless ignored 9 is pending, and while it's pending `migrationWorkNeeded` is true anyway (ignored cursor ≤ 8 < 16), so it was never the exposed tail — that was `rekeyDependencyIDs` plus staging/final-commit, exactly as you said. Thanks for correcting the record with a reproduction instead of letting it slide.

Three things left, only the first two gating:

1. **Rebase** — the branch conflicts with main in exactly one hunk (`lock.go`: main's #5012 added `MigrateLockOption`/fresh-bootstrap heal to `MigrateUpWithLock`). Resolution order: probe first, then the option loop, then lock, then main's heal body — no semantic interaction, since a freshly-bootstrapped DB has no sentinel. Everything else auto-merges; the week's migration-number churn is absorbed by your `"%d/%d"` format.
2. **Make the sentinel write non-fatal.** `MigrateUp` currently returns `ensureMigrationPassComplete`'s error, so a stamp failure (read-only working set, disk) fails a migration that fully succeeded and committed — a new failure mode purchased by an optimization. The sentinel is a hint: log and continue, never propagate.
3. Confirm-only: `TestMigrationProbeRefusesFastPathMidPassOnRealServer` skips when `dolt` isn't on PATH — worth one look at the covering CI job's log to confirm it executes rather than skips, since it's the whole regression guard.

Two non-gating nits for whenever: the `lock.go` doc comment should say "no pass was in flight *at probe time*" (the interval phrasing overstates), and the `!needed && seedChanged` branch commits `dolt_ignore` without revoking the sentinel — harmless, but either wrap it or document the exemption.

Your call on who does 1–2: your turnaround has been same-day, so it's yours if you want it; otherwise we'll fold both into a maintainer push on this branch mid-week and approve. Either way this is landing.

maphew:

Resolution notes: main gained a variadic `MigrateLockOption` parameter (`WithFreshBootstrapHeal`) since you branched; your sentinel-gated fast path now runs as the first check in `MigrateUpWithLock`, before the option processing (the fast path needs none of it), and probe errors still fall through to the locked path rather than failing the open. Tests adapted to main's newer `drainCall`-style stored-proc mocking, and the dirty-guard mocks now account for the sentinel-revocation `DELETE` your design issues before the first mutation. Full schema package passes locally. Please double-check the resolution preserves your invariants.

maphew (commented):
Reviewed at `3a017a89`.

Thanks @atbrace—the sentinel closes the mid-pass race, and the maintainer refresh preserved that invariant. One gating item remains: `MigrateUp` still returns the sentinel-stamp error after the actual migration has completed. Because the row only enables an optimization, a failed stamp must warn and fall back to future locked opens, not fail an already-successful migration.

After that fix, resolve current drift and obtain fresh CI. This is the stronger, earlier implementation than duplicate #4764; preserve this contributor's commits and credit #4764 when consolidating.

jacobhausler:
Cherry-picked these two commits onto current main and verified the pass-completion sentinel against a live `dolt sql-server` rather than sqlmock: `TestMigrationProbeRefusesFastPathMidPassOnRealServer` and `TestMigrateUpWithLockSentinelGatesFastPathOnRealServer` both pass, and the degradation path holds — a pass that dies mid-flight leaves the sentinel absent, so later openers take the lock instead of fast-pathing into a half-finished pass. The race raised in review looks closed.

One item from that review round appears unaddressed on the current head (`3a017a89b`): `MigrateUp` still returns `applied, ensureMigrationPassComplete(ctx, db)` (`internal/storage/schema/schema.go:473`), so a sentinel-stamp failure after an otherwise successful migration still reaches the caller as a migration error.

Conflicts against current main are mechanical, confined to `lock.go`/`lock_test.go` from the `MigrateLockOption` and `ignoredSource` probe work that landed since this branched.

atbrace:
@maphew Both remaining items are addressed at `f947c3d5c`.

**1. The gating item — the stamp error no longer fails a successful migration** (`1e5fe6c1f`). `MigrateUp` now logs the failure, naming the consequence, and returns success:

```
schema migration completed, but recording the pass-completion sentinel failed: <err>; later opens will take the migration lock instead of the lock-free fast path
```

Your reasoning is the comment now sitting above it: the row only enables an optimization, so its absence degrades to locked opens — the pre-sentinel behavior, always correct. Returning the error reported failure to a caller whose database was in fact fully current, and invited a retry of a pass with nothing left to do. Both `MigrateUpWithLock` paths route through `MigrateUp`, so one change covers them.

`TestMigrateUpSucceedsWhenPassSentinelStampFails` pins it: a no-work pass whose `REPLACE INTO local_metadata` errors must still return nil.

**2. Drift resolved** (`f947c3d5c`) — merged current `main` (`7505e173f`), the same shape as your earlier maintainer refresh, so this contributor's commits stay intact. All conflicts were in `lock_test.go`; resolution notes are in the merge commit message. Two worth calling out because they were behavioral, not textual:

- `expectDirtyGuardRefusal` was restructured on main (`cursor`/`eventsFlipVersion`, dirty `events` rather than `issues`) — took main's version and kept this branch's sentinel-revoke expectation inside it.
- `TestMigrateUpWithLockFreshBootstrapHealCapabilityIsOneShot` needed a new expectation: the post-reset **rerun** is a second `migrateUp` pass, so it revokes the sentinel too. Without it sqlmock desynced and the test failed on a closed connection. Confirmed this test passes on unmodified `main`, so it was this branch's interaction, not pre-existing.

**Verification** (linux/amd64, go 1.26): `go build ./...` clean, `go vet ./internal/storage/schema/` clean, and the full `./internal/storage/...` tree green — 26 packages, including `schema`, `dolt`, `embeddeddolt`, `uow`, and `issueops`.

Ready for fresh CI.

maphew:
Handing the baton to @csells,  @julianknutsen on this one. I've stepped down as maintainer and resumed role of normal contributor and can't trigger CI etc.

atbrace:
Closing: superseded by #6022 (merged 2026-08-27), which answers the steady-state "nothing to migrate" question without GET_LOCK via alreadyConverged and keeps the locked path for anything that can actually migrate. Same goal, already on main. Thanks maphew, jacobhausler, and codex for the reviews that got the pass-completion race nailed down here.

PR #4808, opened 2026-07-15
fix(windows): report Dolt run-state correctly in bd config drift

marcodelpin:
## What

Fix `bd config drift` misreporting the shared Dolt server's run state on Windows.

**Scope** (narrowed per review): `pidAlive` has exactly one production caller — `bd config drift`. `bd doctor`'s server run-state goes through `doltserver.IsRunning`, whose Windows helper was separately fixed on main by 594c464cc (crediting this PR); that path is NOT claimed here.

## Why

`isServerProbablyRunning()` (`cmd/bd/config_drift.go`) checks the server PID with `proc.Signal(syscall.Signal(0))` — the Unix null-signal existence probe. Windows has no signals: `os.Process.Signal` rejects everything but `Kill`, so the probe always errors and **every live PID is treated as dead**. On Windows, drift detection therefore always emits

```
dolt.shared-server is enabled but no server appears to be running
```

even when the server is up.

An existence-only replacement (trusting `os.FindProcess` / `OpenProcess` success) is **not** correct either: on Windows a terminated process's kernel object stays openable for as long as any handle to it lingers anywhere on the system (an un-`Wait()`ed parent, an AV/EDR scanner, a debugger, Task Manager), so it reports a **dead** server as alive for a window after it exits.

## Fix

Split the liveness probe into a build-tagged `pidAlive(pid)` helper, following the pattern already used in `internal/doltserver` and `internal/linear`:

- **Unix** keeps `Signal(0)` (`os.FindProcess` always succeeds there, so the signal is the real check).
- **Windows** uses `OpenProcess(SYNCHRONIZE | PROCESS_QUERY_LIMITED_INFORMATION)` then `WaitForSingleObject(h, 0)` and `CloseHandle`. A terminated process's handle is **signaled** (`WAIT_OBJECT_0`); a running one **times out** (`WAIT_TIMEOUT`). This distinguishes live from dead even when `OpenProcess` still succeeds on a lingering handle.

Unix behavior is unchanged.

## Tests

Adds `TestIsServerProbablyRunningReportsDeadPIDWithLingeringHandle` (Windows): it starts a child that exits immediately, deliberately never `Wait()`s it so `os/exec` keeps a lingering handle open, confirms the child has terminated, then asserts the drift probe reports it dead. The test **fails against the naive `OpenProcess`-success probe and passes with this fix** — it is a genuine regression guard, not a tautology.

The pre-existing `TestIsServerProbablyRunningOwnPID` (live PID) continues to pass on both platforms.

## Verification

```
go test -tags gms_pure_go -run 'TestIsServerProbablyRunning' ./cmd/bd/   # Windows: both pass
go vet -tags gms_pure_go ./cmd/bd/                                        # clean; each build-tagged file compiles on its platform
GOOS=linux go vet -tags gms_pure_go ./cmd/bd/                             # unix variant compiles, byte-identical probe
```

## Note

`internal/doltserver/doltserver_windows.go`'s `isProcessAlive` uses the same bare-`OpenProcess`-success heuristic and has the same latent lingering-handle issue; happy to follow up on that separately to keep this PR to one change.

---

_Prepared with AI assistance (Claude Code); reviewed and submitted by the author. Disclosed via the `Agent-Signature:` trailer per `engdocs/AGENT_SIGNING.md`._


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified at the PR head). Verdict: **needs changes** — the wait-based liveness algorithm is sound against the pinned `x/sys/windows`; three accuracy/coverage items.

**Should-fix 1 — the fix's claimed scope overstates.** Verified: `processAlive` has exactly one production caller — `bd config drift`. `bd doctor`'s server run-state still goes through `doltserver.IsRunning` → the existence-only `OpenProcess` helper on Windows (that helper was separately fixed by 594c464cc, which credits this PR). Narrow the commit/PR claim to `bd config drift`, or note the sibling landing.

**Should-fix 2 — the Windows regression test never executes in CI.** The PR workflow selects the test by name on Ubuntu (where the `//go:build windows` file is excluded), and the Windows job runs only on pushes to main and does build/smoke, not `go test`. A targeted Windows PR-time test run (just this test) closes the gap cheaply.

**Should-fix 3 — the test's precondition isn't established deterministically.** It discards `WaitForSingleObject`'s status and checks only `err` — but the pinned x/sys returns nil error for `WAIT_TIMEOUT`; only `WAIT_FAILED` errors. Require `WAIT_OBJECT_0` (as main's newer Windows tests do), else a slow child exit turns into a spurious failure.

Nit: `h`/`s` → `processHandle`/`waitResult`.

Drift is clean: the `IsSharedServerMode`/`SharedServerPath` move composes, and 594c464cc doesn't supersede the `cmd/bd` fix here. Good, needed fix — tighten the claims and the test and it lands.

marcodelpin:
Addressed the review in a53db4370; PR title/body narrowed accordingly.

1. **Scope narrowed (should-fix 1)** — title and body now claim `bd config drift` only (`processAlive`'s single production caller), with an explicit note that `bd doctor`'s run-state goes through `doltserver.IsRunning`, separately fixed on main by 594c464cc.
2. **Windows PR-time test lane (should-fix 2)** — new `test-windows-liveness` job in `pr.yml` (windows-latest, runs `-run '^TestIsServerProbablyRunning'` — 5 tests including the lingering-handle regression guard), wired into the CI gate (`needs` + `CI_GATE_REQUIRED` + env mapping) so it actually gates rather than running advisory-only. Rationale in the job comment: the `//go:build windows` file is compiled out on the Ubuntu name-selected run, and `main.yml`'s Windows job is push-to-main build/smoke only.
3. **Deterministic precondition (should-fix 3)** — the test now requires `WAIT_OBJECT_0` from `WaitForSingleObject` (the pinned x/sys errors only on `WAIT_FAILED`, so a `WAIT_TIMEOUT` slipped through the err-only check), matching the pattern of the newer Windows tests in `internal/doltserver` and `internal/linear`, same failure message format.

nit: `h`/`s` renamed to `processHandle`/`waitResult`.

Verified on Windows (go1.26.5, gms_pure_go): `go test -count=1 -run '^TestIsServerProbablyRunning' ./cmd/bd` — 5/5 pass (NoPIDFile, BadPID, DeadPID, OwnPID, ReportsDeadPIDWithLingeringHandle); `gofmt -l` and `go vet` clean.

marcodelpin:
CI red root-caused and fixed in e9668b8e1: upstream #5004 (bb9bb7487, merged this morning) added a test-only `processAlive` helper in `cmd/bd/proxied_local_helpers_test.go` (`//go:build cgo && unix`); on the PR merge ref both declarations coexisted in package main, so the cgo unix lanes failed to compile (`processAlive redeclared` — Build (Embedded Dolt) + PR Core). Renamed this PR's production helper to `pidAlive`; the #5004 test helper is untouched (minimal diff, 3 files, 5 lines).

Verified before pushing: merged tree (PR head + current main) compiles clean on Linux with cgo (`go vet ./cmd/bd` + `go test -c ./cmd/bd`), and the Windows drift tests are still 5/5. PR body updated to the new name.

marcodelpin:
PR Risk shard note: `Test (Proxied Dolt Cmd 1/15)` failed at **setup** — `bd init --proxied-server` → `pre-repair for migration 0053_repair_rig_wisps.up.sql: checking wisp_dependencies.depends_on_issue_id: Error 1105 (HY000): no root value found in session` (run 30082257990). Unrelated to this diff (a windows-tagged test + a unix-side symbol rename, compile-verified on the merge ref): the error fires in init/migrate before any changed path executes, and 20/21 sibling subtests in the same family passed, including the heavier cascade/delete paths. The signature has no hits in either repo's issue search — flagging here in case it recurs elsewhere.

maphew:
On your shard note: that `Error 1105 … no root value found in session` at proxied init/migrate is a known flake family in the proxied fresh-init path — the same signature class was fixed on main via #5042 (merged 07-25), so a rerun against an updated branch should clear it. Good instinct flagging it with the run link rather than retrying silently.

Status otherwise: the three items from the 07-24 review still stand (they're accuracy/coverage, not blockers to your approach), and the branch has since picked up a conflict with main. If you'd like, we can handle the rebase maintainer-side once you've pushed the review items — or push everything together, your call.

maphew:

The only conflict was `.github/workflows/pr.yml`: your `test-windows-liveness` job and main's newer `worktree-remove-windows` job landed in the same location — kept both, and wired both into the `ci-gate` needs/required lists. Full build passes and the `pidAlive` tests are green locally.

marcodelpin:
All three should-fix items are addressed at the current head (`262aa0c2a`), and the review gate is
the only thing still red — `CHANGES_REQUESTED` predates the work that answered it. Here is each item
with what it can be checked against, so this is verifiable rather than asserted.

**SF1 — claimed scope.** The PR/commit title is now narrowed to the single caller:
`fix(windows): report Dolt run-state correctly in bd config drift`. In `cmd/bd`, `pidAlive` has
exactly one production caller, `cmd/bd/config_drift.go:331`. (`internal/workspacegate/gate.go:300`
declares a same-named but unrelated package-scoped function — different package, not a second caller
of this one.) `bd doctor` still goes through `doltserver.IsRunning`, unchanged by this PR, as you noted.

**SF2 — the Windows regression test never executes in CI.** There is now a PR-time Windows lane,
`Test (Windows) cmd/bd liveness regression`, and on this head it **ran and passed** — an executed
lane rather than a claim about one. Thanks for keeping it alongside `worktree-remove-windows` and
wiring both into `ci-gate` when you resolved the `pr.yml` conflict.

**SF3 — non-deterministic precondition.** `cmd/bd/config_drift_windows_test.go:48` now requires
`WAIT_OBJECT_0` explicitly and fails with the observed status otherwise, instead of checking only
`err` — so a slow child exit no longer reads as a pass.

**Nit** — `h`/`s` are now `processHandle`/`waitResult`.

On the red you saw last: the failing jobs were `errcheck` on
`cmd/bd/config_drift_windows.go:26` (the unchecked `windows.CloseHandle`), not the flake. It is fixed
in `262aa0c2a` with the same `defer func() { _ = ... }()` form the five sibling Windows files already
use. The three bare calls in the `_test.go` are deliberately untouched, since `.golangci.yml` sets
`tests: false`, which is why the linter named only line 26. CI is now 85 success / 1 skipped /
**0 failures**, and the PR is `MERGEABLE`.

Happy to squash or rebase if you would prefer a different shape before it lands.

maphew (approved):
Reviewed at `262aa0c2ae6` using the correlated bead's staged `gpt-5.6-terra` / medium routing, then adjudicated here.

No correctness findings. The Windows helper correctly distinguishes a live process (`WAIT_TIMEOUT`) from a terminated process whose handle remains open (`WAIT_OBJECT_0`), preserves Unix signal-0 behavior, and is limited to `bd config drift`. The implementation matches the established liveness handling in `internal/doltserver` and `internal/linear`; the required Windows CI lane covers the lingering-handle regression.

Verification:

- `go test -tags gms_pure_go -count=1 -run '^TestIsServerProbablyRunning' ./cmd/bd` — PASS (`ok`, 1.657s).
- `go vet -tags gms_pure_go ./cmd/bd` — PASS.
- GitHub — `CLEAN`, `MERGEABLE`, with all reported checks successful; the aggregate changes-requested state is from an older head.
- Blocking upstream preflight — blocked only because the stale changes-requested decision remains; no code/check failure was reported.

Disposition: **approve**. Refresh against current main and let CI revalidate before merge because this head's merge base is substantially behind.

steveyegge:
Post-merge note: the `test-windows-liveness` job this PR added predated the Go-cache-ownership invariant that landed on main while this PR awaited merge (`TestGoCacheOwnershipTopology`, #5278), so its `setup-go` pin tripped that check once merged — fixed forward in #5404 (041f1d64d): released v7.0.0 SHA + `cache: false`, job behavior otherwise unchanged. Nothing needed from you — the liveness regression guard is live on every PR as intended. The merge-timing miss was mine, not this PR's.

Issue #770, opened 2025-12-28
rename-prefix: 8-char limit not enforced elsewhere


jdillon:
## Summary

`bd rename-prefix` enforces an 8-character prefix limit that isn't enforced by `bd init` or `bd create`. This creates a situation where you can create issues with longer prefixes but can't rename to consolidate them.

## Reproduction

```bash
# This works fine:
bd config set issue_prefix claudecode
bd create "Test issue"  # Creates claudecode-xxx

# But this fails:
bd rename-prefix claudecode-
# Error: prefix too long (max 8 characters): claudecode
```

## Location

The limit is in `cmd/bd/rename_prefix.go:191`:
```go
if len(prefix) > 8 {
    return fmt.Errorf("prefix too long (max 8 characters): %s", prefix)
}
```

## Suggestion

Either:
1. Remove the limit from `rename-prefix` (since the DB handles longer prefixes fine)
2. Or add consistent validation to `init`/`create` if there's a technical reason for the limit

Not urgent - just a consistency nit. Happy to PR if you have a preference on which approach.

steveyegge:
Fixed in ab3feda0 - removed the arbitrary 8-char limit. Thanks for the report!

PR #4813, opened 2026-07-15
docs: remove stale 8-char prefix limit claim from rename-prefix help (GH#770)

kevglynn:
## Summary

- Commit ab3feda00 removed the `len(prefix) > 8` check from `validatePrefix` (closing #770), but the `Max length: 8 characters` line survived in the command's `Long:` help text, so `bd rename-prefix --help` and every generated doc still document a limit that no longer exists. Observed in the wild: the dry run happily accepts 9- and 16-char prefixes while the help says the max is 8.
- Removes the stale line from the Go source (`cmd/bd/rename_prefix.go`) and the hand-written plugin skill page, then regenerates the CLI docs with `./scripts/generate-cli-docs.sh`.
- The regen also picks up pre-existing drift in the `bd backup` pages (the "Auto-backup default" help section had not been regenerated on main), which keeps `generate-cli-docs.sh --check` green.

## Test plan

- [x] `./scripts/generate-cli-docs.sh --check` passes (generated docs fresh)
- [x] `CGO_ENABLED=1 go test -tags gms_pure_go -run '^TestValidatePrefix$' ./cmd/bd/` passes (existing test already asserts `verylongprefix-` is valid, "No length limit (GH#770)")
- [x] `scripts/pr-preflight.sh --search` found no open PRs or issues covering this

### DIFF
--- a/cmd/bd/rename_prefix.go
+++ b/cmd/bd/rename_prefix.go
@@ -34,7 +34,6 @@ USE CASES:
 - Migrating to team naming standards
 
 Prefix validation rules:
-- Max length: 8 characters
 - Allowed characters: lowercase letters, numbers, hyphens
 - Must start with a letter
 - Must end with a hyphen (e.g., 'kw-', 'work-')
--- a/docs/reference/advanced.md
+++ b/docs/reference/advanced.md
@@ -30,7 +30,7 @@ bd rename-prefix kw-            # Every knowledge-work-* ID becomes kw-*
 ```
 
 The rename updates all issue IDs and all text references across all fields.
-Prefixes are at most 8 characters of lowercase letters, numbers, and hyphens,
+Prefixes are lowercase letters, numbers, and hyphens,
 must start with a letter, and must end with a hyphen. If a corrupted database
 contains issues with multiple prefixes, `bd rename-prefix <prefix> --repair`
 consolidates them. See [bd rename-prefix](/cli-reference/rename-prefix).
--- a/plugins/beads/skills/beads/commands/rename-prefix.md
+++ b/plugins/beads/skills/beads/commands/rename-prefix.md
@@ -9,7 +9,6 @@ Updates all issue IDs and all text references across all fields.
 
 ## Prefix Rules
 
-- Max length: 8 characters
 - Allowed: lowercase letters, numbers, hyphens
 - Must start with a letter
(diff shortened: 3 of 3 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the pin policy). Verdict: **needs changes** — the source fix is right; the generated files now collide with the release-pin policy that landed after you branched.

**Blocker (drift, not your doing) — generated docs violate the release pin.** Main now pins generated CLI docs to `v1.1.0` (`docs/cli-docs.pin` + the 2026-07-17 docs-release-pin decision); the generator produces from the pinned release and the freshness gate rejects drift. v1.1.0's binary still emits the eight-character sentence, so this PR's edits to `docs/CLI_REFERENCE.md` / `docs/cli-reference/rename-prefix.md` / `backup.md` are guaranteed check failures (the backup hunks are also unrelated pre-existing drift swept up by the old full-tree regen). Keep the `cmd/bd/rename_prefix.go` help fix and the plugin correction; drop the three generated files — they regenerate when the pin advances.

**Should-fix — one more user-facing copy of the stale limit survives:** `docs/reference/advanced.md` (hand-written, not generated) still says prefixes are limited to eight characters, while the validator has no length check and the tests accept `verylongprefix-`. Fix that page in this PR since it's the actual documentation debt #770 complained about.

Side observation (pre-existing, not verdict-driving): the help text says prefixes must end with a hyphen while the tests accept `kw` — worth reconciling someday.

kevglynn:
Addressed review feedback:

- **Dropped all 3 generated doc files** — they're release-pinned to v1.1.0 and will regen on next release advancement. Only source changes remain.
- **Fixed `docs/reference/advanced.md`** — removed the "at most 8 characters" claim (the hand-written page the review flagged as the actual documentation debt from #770).

Branch rebased onto current main. Three files changed: `cmd/bd/rename_prefix.go`, the plugin skill page, and `docs/reference/advanced.md`.

maphew (approved):
Clean response — generated docs dropped per the release pin, and fixing docs/reference/advanced.md got the actual documentation debt from #770, which was always the real payload here. Approving; merge patrol takes it from here. _claude-fable-5-medium on behalf of maphew_

Issue #4816, opened 2026-07-15
bd close falsely reports success (✓ Closed, exit 0) on already-closed issues; new --reason silently dropped


kevglynn:
## Summary

`bd close <id> --reason "..."` on an **already-closed** issue prints `✓ Closed <id>: <reason>` and exits 0, but the new reason is never persisted (first close wins, by design since #4025). The operator's evidence-mapped close reason is silently dropped while the CLI reports success. The CLI also does all of its "a close happened" bookkeeping for the no-op: it increments `closedCount`, updates last-touched, triggers `--suggest-next` / `--claim-next` / molecule auto-close, attempts a Dolt commit, and calls the audit sidecar hook as if a state change occurred.

This is the same class as #4517 (write commands must verify durable row state before reporting success) — the durable-state signal exists at the storage layer but is dropped on the way to the CLI.

## Repro (stock beads, no orchestrator; HEAD 5f685959c)

```bash
mkdir /tmp/bd-repro && cd /tmp/bd-repro && git init -q && bd init
bd create "t" -t task -p 2 --description d        # → repro-vcl
bd close repro-vcl --reason "FIRST"               # ✓ Closed repro-vcl — t: FIRST
bd close repro-vcl --reason "SECOND"              # ✓ Closed repro-vcl — t: SECOND, exit 0
bd show repro-vcl --json                          # close_reason is still "FIRST"
```

Observed on a fresh build of `main` (5f685959c): the second close prints `✓ Closed repro-vcl — t: SECOND` and exits 0; `close_reason` remains `FIRST`.

A common agent flow that hits this: `bd epic close-eligible` closes an epic with its hardcoded reason, then an agent runs `bd close <epic> --reason "<detailed evidence>"` — false success, reason lost (the missing `--reason` on close-eligible is filed separately as #4817).

## Root cause (verified at HEAD)

- `internal/storage/issueops/close.go` (`closeIssueInTx`): the UPDATE has `WHERE id = ? AND status != 'closed'`. On an already-closed row it correctly returns `CloseResult{AlreadyClosed: true}` with nil error. **This idempotency is deliberate** — it fixed #4025 (re-close used to surface a bogus "issue not found"), and `closeIdempotent` in `internal/storage/domain/db/close_test.go` pins first-reason-wins with no error. The storage contract should not change.
[...]

PR #4818, opened 2026-07-15
feat(storage): add CloseIssueWithResult to expose the already-closed no-op (GH#4816)

kevglynn:
## Summary

Storage-layer half of #4816 (PR 1 of a 2-PR stack; the `cmd/bd` wiring follows in a separate PR).

`CloseIssue` is deliberately idempotent per #4025: re-closing a closed issue returns nil, keeps the first `close_reason`, and mints no second `closed` event. `issueops.CloseIssueInTx` already reports this via `CloseResult.AlreadyClosed`, and the newer domain layer propagates it (`CloseIssueResult.Closed`), but the `storage.DoltStorage` boundary the CLI uses returns only `error` — so `cmd/bd` cannot distinguish a real close from a no-op and falsely reports success (#4816; principle from #4517: write commands must verify durable state before reporting success).

This PR adds to `storage.Storage`:

```go
CloseIssueWithResult(ctx, id, reason, actor, session) (*CloseResult, error)
```

with `CloseResult{AlreadyClosed bool}`, implemented by threading the existing `issueops` result through every backend:

- `dolt` (including the wisp path), `embeddeddolt`, `sqlkit` (shared by postgres/mysql/sqlite)
- `HookFiringStore`: fires `on_close` only for a real close; the already-closed no-op fires no hook. `CloseIssue` keeps its old unconditional firing so existing callers see zero behavior change.
- `telemetry.InstrumentedStorage`: straight passthrough with a span
- postgres/mysql/sqlite `unsupported_gen.go` regenerated (skip-list addition only)

`CloseIssue` now delegates to the new method and is otherwise unchanged — the #4025 contract (nil error, first reason wins, no duplicate `closed` event) is preserved byte-for-byte, and the conformance suite asserts it stays that way.

## Why a new method instead of changing CloseIssue's signature

Changing `CloseIssue` to return `(*CloseResult, error)` touches every call site, test fake, and SDK consumer in one diff. The additive method follows the existing `SearchIssues`/`SearchIssuesWithCounts` precedent and lets the CLI adopt the signal incrementally.

## Tests

- New conformance subtest `Audit/issue-lifecycle/CloseWithResultReportsAlreadyClosed`: first close reports `AlreadyClosed=false`; re-close reports `true` with nil error, first reason intact, exactly one `closed` event; missing ID is still `storage.ErrNotFound`. Runs against every backend the conformance gate covers.
- Hook-decorator unit tests: `on_close` fires on a real close, is skipped on the already-closed no-op.
- Fail-on-HEAD note: these tests exercise a method that does not exist on HEAD, so they fail to compile there rather than fail at runtime. The CLI-visible regression tests (which do fail on HEAD) land with the stacked `cmd/bd` PR.

## Validation

- `go vet ./...`, `golangci-lint run` on touched packages: clean
- `go test ./internal/storage/... ./internal/telemetry/... ./internal/jira/...`: pass, except the `embeddeddolt` package which hits the 10m go-test timeout on my machine **with or without this change** (verified against a pristine `origin/main` worktree — the full `BEADS_TEST_EMBEDDED_DOLT=1` package run is simply slower than 10m here). The new conformance subtest passes in isolation in ~5s.

Refs #4816. Contract preserved from #4025; principle from #4517.


kevglynn:
**Superseded by upstream work — proposing to close.**

Since this PR opened, `main` landed [`CloseIssueChecked`](https://github.com/gastownhall/beads/pull/4893) (`CloseIssueResult.Unchanged` for the already-closed no-op). That is the storage-layer signal this PR was adding as `CloseIssueWithResult`.

A further CLI delegation that consumes `Unchanged` (and skips audit / closedCount / suggest-claim side effects) is in open [#4911](https://github.com/gastownhall/beads/pull/4911) (consolidates closed #4904). Note: #4911 deliberately keeps printing `✓ Closed` for output parity; the remaining delta vs [#4816](https://github.com/gastownhall/beads/issues/4816) is truthful messaging / an `already_closed` JSON indicator — that would be a thin follow-up on top of #4911, not this storage method.

I am not rebasing this branch onto current `main` (sqlkit/postgres stubs were removed; the additive method is obsolete). Happy to close this PR unless a maintainer prefers otherwise.

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the supersession on current main). Verdict: **needs changes — suggested disposition: retire as superseded.**

The implementation matched its spec at its base (the storage-wide `CloseIssueWithResult` wiring and the conformance coverage both check out). But since the branch point, main landed the same capability as `CloseIssueChecked` returning `CloseIssueResult.Unchanged` (built on the same `issueops.CloseResult.AlreadyClosed`), wired into `bd close` with idempotent-re-close side-effect suppression (verified at `cmd/bd/close.go:148+`). Rebasing this PR would introduce a second result-bearing close API for the same fact.

Seven touched files also no longer exist on main (sqlkit and the MySQL/PostgreSQL/SQLite `unsupported` stubs were deleted with the backend removal) — modify/delete conflicts that must resolve as deletions.

**One piece worth carrying forward:** this PR's hook decorator deliberately skips `on_close` for a no-op re-close, while main currently fires the checked-close hook unconditionally. If that conditional semantics is desired (it likely is — sibling #4819 makes the same point), it should be ported to `CloseIssueChecked` as a small focused PR rather than via this branch.

kevglynn:
Closing as superseded — `CloseIssueChecked` on main delivers this capability. The one piece worth preserving (hook decorator conditional skip of `on_close` for no-op re-closes) will be a small focused follow-up PR against the current `CloseIssueChecked` API.

Issue #4052, opened 2026-05-20
Shared-server mode: write commands report success when Dolt is unreachable; status change never lands


SeraphimKa:
## TL;DR

When the Dolt SQL server isn't reachable in shared-server mode, `bd close` (and other write commands) print `✓ Closed`, exit 0, and append a "status closed" event to `.beads/interactions.jsonl`, but the actual `issues.status` UPDATE doesn't reach Dolt. When the server comes back online, `bd ready` / `bd list` shows the pre-close state and the user thinks the issue zombied back.

Same shape as #3360, #3383, #3905. Posting a clean repro.

## Repro

```sh
systemctl --user start beads-shared    # or whatever brings your dolt sql-server up
bd doctor | grep Connected
#   ✓ Dolt Connection Connected successfully

systemctl --user stop beads-shared     # take the backend down

bd close <some-issue-id> --reason "test"
# prints close_reason, exits 0
# .beads/interactions.jsonl: gets a status-closed audit row
# .beads/issues.jsonl: status is NOT updated

systemctl --user start beads-shared
bd show <same-issue-id>
# status: IN_PROGRESS (not CLOSED, despite the success message earlier)
```

## How I noticed

WSL2 rebooted at 10:07. The `beads-shared` systemd user unit started at 10:08:08. I'd opened a terminal in that window and run two `bd close` commands at 08:24 and 08:29, before the unit was up. Both printed `✓ Closed` and exited 0. A few hours later when I checked the queue, both were still `in_progress`. The audit log had both close events with `int-XXXX` ids, but `issues.status` in Dolt was untouched.

The boot race itself is probably WSL-specific (systemd user services lag a bit after WSL2 brings the session up). The silent drop reproduces on any platform via the steps above.

## Env

- bd 1.0.4 (ce242a879)
- Arch Linux on WSL2
- `.beads/metadata.json`:
  ```json
  {
    "backend": "dolt",
    "dolt_mode": "server",
    "dolt_server_host": "127.0.0.1",
    "dolt_server_port": 3308
  }
  ```

## What would help

Non-zero exit with a clear "backend unreachable" message in the write path would have saved me a few hours of suspecting my git hooks. Bonus: gate the interactions.jsonl append on a successful Dolt commit (covers #3360 too).

## Related

#3360, #3383, #3905, #3760, #3545

## Workaround I'm using

[...]

PR #4819, opened 2026-07-15
fix(close): stop reporting success when the issue was already closed (GH#4816)

### LINKED ISSUE #4816: bd close falsely reports success (✓ Closed, exit 0) on already-closed issues; new --reason silently dropped

kevglynn:
## Summary

CLI half of #4816. **Stacked on #4818** — this branch contains that PR's storage commit plus one commit of `cmd/bd` wiring; please review/land #4818 first (after it merges this PR rebases to a single commit).

Before (repro from #4816, stock beads):

```
$ bd close bd-1 --reason "FIRST"
✓ Closed bd-1: FIRST
$ bd close bd-1 --reason "SECOND"
✓ Closed bd-1: SECOND        # false — nothing changed, exit 0
$ bd show bd-1 --json | jq .close_reason
"FIRST"
```

The storage layer's idempotent close (#4025) is a no-op here, but `bd close` reported success, silently dropped the new `--reason`, wrote the audit status-change entry, and ran every "we closed it" side effect: `closedCount` (which drives last-touched, `--suggest-next`, `--claim-next`, `--continue`, molecule auto-close) and the embedded auto-commit. Principle from #4517: a write command must verify durable state before reporting success.

After — `cmd/bd/close.go` uses `CloseIssueWithResult` (from #4818) and on `AlreadyClosed`:

- stderr: `<id> already closed — close reason unchanged; use `bd reopen <id> && bd close <id> --reason ...` to rewrite it`; no `✓ Closed` line
- skips the audit entry, `closedCount` increment, last-touched update, molecule auto-close, suggest/claim-next, and the embedded commit for that ID
- JSON: `{"closed": [...], "already_closed": [...]}` when any target was already closed; the existing plain-array shape is unchanged when none were
- exit code stays 0 when every attempted target was an already-closed no-op (idempotent-close scripts keep working); real failures still exit non-zero, and mixed batches report each ID truthfully

## Open questions for maintainers

1. **Exit code with an explicit `--reason` that could not be applied.** This PR keeps exit 0 for all already-closed no-ops to preserve idempotency for scripts/agents that re-close defensively. An argument exists for non-zero when the user explicitly passed `--reason` (their input was dropped). Happy to change or gate behind config — flagging rather than deciding unilaterally since it is a behavior change.
2. **Proxied-server path.** `runCloseProxiedServer` is untouched; the server RPC path has the same false-success family tracked in #4052 and would need the signal threaded through the protocol. Out of scope here.

## Tests (`cmd/bd/close_embedded_test.go`)

All verified to fail on the pre-fix code (ran with the `close.go` change stashed) except the marked guard:

- `reclose_reports_already_closed_and_preserves_reason` — truthful stderr, no success line, exit 0, first reason wins
- `reclose_json_reports_already_closed` — `already_closed` JSON shape
- `reclose_skips_claim_next` — no-op close must not claim the next ready issue
- `close_mixed_batch_reports_each_id_truthfully` — one open + one already-closed target
- `reclose_mints_no_second_closed_event` — green-by-design guard that the #4025 contract (no duplicate event, no error) survives this change

Full `TestEmbeddedClose` suite passes with the fix (`BEADS_TEST_EMBEDDED_DOLT=1`). `golangci-lint run ./cmd/bd/`: clean.

Fixes #4816. Depends on #4818. Related: #4025 (contract preserved), #4517 (principle), #4052 (proxied-path sibling, not addressed here).


kevglynn:
**Largely superseded — proposing to close (or retarget as a thin follow-up).**

`CloseIssueChecked` + `Unchanged` landed in [#4893](https://github.com/gastownhall/beads/pull/4893). Open [#4911](https://github.com/gastownhall/beads/pull/4911) (includes the former #4904) already:
- delegates `bd close` to `CloseIssueChecked`
- skips audit / closedCount / molecule auto-close / suggest-claim on `Unchanged`
- keeps exit 0

What #4911 deliberately does **not** do (output parity with the old false-success path):
- still prints `✓ Closed <id>: <reason>` on a no-op re-close
- still emits the issue in the plain JSON array (no `already_closed` key)

So the remaining piece of [#4816](https://github.com/gastownhall/beads/issues/4816) after #4911 lands is truthful messaging only. Options:
1. Close this PR as superseded by #4911; open a small follow-up for the message/JSON shape once #4911 merges.
2. Rewrite this PR against `main`+`CloseIssueChecked` as that thin follow-up now (conflicts with #4911's intentional parity choice — coordinate with that author).

I am holding the rebase until a maintainer/author preference is clear; the storage half (#4818) should close either way.

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the supersession on current main). Verdict: **needs changes — retire or substantially redesign atop `CloseIssueChecked`.**

**The core has been superseded:** since this branched, main landed transactional `CloseIssueChecked` (f9b20203e + 607df586d), which closes the check-then-close race this PR still has (its `IsBlocked` and `CloseIssueWithResult` run as separate transactions — the pinned driver isolates within, not across, that gap) and already suppresses real-close-only side effects on idempotent re-close. Ten conflict hunks plus deleted-backend files make a mechanical rebase a non-starter.

**Also, at the PR head:** proxied mode bypasses the fix entirely — the proxied close path still runs post-close actions and prints `✓ Closed` regardless (its regression test discards stdout/stderr/error); the early-continue suppresses `--continue` / `--claim-next` / molecule auto-close on re-close, which strands crash-retry workflows (current main deliberately implements — and tests — the opposite contract, including stranded-molecule healing); a reload error can make a confirmed already-closed target exit 0 with *no* JSON at all; and the new `already_closed` JSON object conflicts with main's preserved historical array shape.

**Worth preserving:** the hook decorator's conditional skip of `on_close` for no-op re-closes — main currently fires the checked-close hook unconditionally, and porting that one behavior to `CloseIssueChecked` as a small focused PR (with a proxied-parity check) would capture this PR's real value. Same recommendation as sibling #4818.

kevglynn:
Closing as superseded — `CloseIssueChecked` + `Unchanged` on main covers the storage signal, and #4911 handles the CLI delegation. The hook decorator's conditional `on_close` skip for no-op re-closes (the one piece main still lacks) will be a small focused follow-up PR.

Issue #3816, opened 2026-05-08
Feature: Pin/Protect epics from bd epic close-eligible


Wermeling:
## Problem

`bd epic close-eligible` auto-closes any epic where all children are `closed`. This is problematic for **permanent/dauer/epic-level epics** that should remain open permanently as organisational containers, for example:

- `Epic: Bug Fixes` — umbrella for all bug-fix beads
- `Epic: Documentation Fixes` — umbrella for all docs beads
- `Epic: Tests` — umbrella for all test-coverage beads
- `Epic: Flipper Capability Program` — ongoing program epic

These are never "done" in a traditional sense — they are **living buckets** that continuously receive new child beads. Once their current children close, the epic itself gets auto-closed, even though more work will arrive under it.

## Proposed Solution

Introduce a **pinned / protected** state for epics that exempts them from `bd epic close-eligible`.

### Option A: `--pinned` flag on `bd update <epic-id> --pinned`
```bash
bd update <epic-id> --pinned
bd update <epic-id> --unpinned   # remove protection
```

### Option B: `--protect` flag on `bd epic close-eligible`
```bash
bd epic close-eligible --exclude-pinned   # default behaviour, skip pinned epics
bd epic close-eligible                    # current: closes all eligible epics
```

### Option C: Dedicated `type=dauer-epic` or `lifespan=permanent`
Distinguish between **delivery epics** (which are done when children close) and **organisational epics** (which live forever).

## Expected Behaviour

- Pinned/organisational epics are **never auto-closed** by `bd epic close-eligible`
- Pinned epics **still appear in `bd epic status`** and normal list commands
- `bd epic close-eligible` shows a count that **excludes pinned epics** (when filtering)
- The pin/protect state is **visible in `bd list`** and `bd epic status`

## Alternatives Considered

1. **Keep a permanent dummy child open** — works but is a hack that pollutes the tracker
2. **Move completed epics to a "Archive" parent epic** — loses the organizational grouping
3. **Don't use `bd epic close-eligible`** — manual discipline doesn't scale

## See Also

- PR #3276: "Keep regular epics open after the last child closes" — partial improvement, but regular epics are still closed when all children complete
- `bd epic close-eligible --dry-run` preview before running

---
Would be useful for: teams using bd as a long-running organisational layer (not just one-shot project tracking).

harry-miller-trimble:
[...]

PR #4820, opened 2026-07-15
feat(epic): add --reason to bd epic close-eligible (GH#4817)

### LINKED ISSUE #4817: bd epic close-eligible hardcodes close reason; add --reason flag
## Summary

`bd epic close-eligible` hardcodes the close reason `"All children completed"` for every epic it closes (`cmd/bd/epic.go`, `closeEligibleEpicsCmd` calls `store.CloseIssue(ctx, epicID, "All children completed", "system", "")`). There is no `--reason` flag, so an operator or agent cannot attach an evidence-mapped reason at close time.

Because close reasons are first-write-wins at the storage layer (#4025, by design), the hardcoded reason is permanent: a follow-up `bd close <epic> --reason "<detailed reason>"` is a silent no-op that currently even reports success (filed separately as #4816). The only recovery is `bd reopen` + re-close.

## Repro (stock beads, no orchestrator; HEAD 5f685959c)

```bash
mkdir /tmp/bd-repro && cd /tmp/bd-repro && git init -q && bd init
bd create "e" -t epic -p 2 --description d          # → epic E
bd create "c" -t task -p 2 --parent=E --description d
bd close <child> --reason "child done"
bd epic close-eligible                              # closes E, reason "All children completed"
bd close E --reason "DETAILED"                      # false success; reason unchanged
bd show E --json                                    # close_reason: "All children completed"
```

## Proposal

Add `--reason` to `bd epic close-eligible`: one shared reason applied to every epic it closes in that run. Default stays `"All children completed"` so existing behavior is unchanged. `--dry-run` output and the `--json` result should surface the reason that will be / was used.

Not in scope: pinning/protecting epics from close-eligible — that is #3816 and stays separate.

kevglynn:
## Summary

Fixes #4817. Independent of the #4816 stack (#4818/#4819); based directly on `main`.

`bd epic close-eligible` hardcoded the close reason `"All children completed"`, so an agent finishing an epic's children could not attach an evidence-mapped reason to the epic close. Combined with #4816, a follow-up `bd close <epic> --reason ...` silently lost that reason too (first close wins at the storage layer), making close-eligible effectively reason-lossy.

Changes:

- New `-r/--reason` flag: one shared reason applied to every epic the command closes. Default stays `"All children completed"`, so existing behavior is unchanged; an explicitly empty reason falls back to the default.
- Human dry-run preview and the final summary line now show the reason.
- Non-dry-run JSON gains a `"reason"` key: `{"closed": [...], "count": N, "reason": "..."}` (additive).
- Dry-run `--json` shape (array of epic statuses) is **deliberately unchanged**: the reason there is an input the consumer already knows, and reshaping the array into an object would break existing consumers. Flagging in case maintainers prefer the reshape.
- CLI reference regenerated via `scripts/generate-cli-docs.sh` (epic pages only; unrelated pre-existing backup.md drift on main was left out of this diff).

## Tests (`cmd/bd/epic_embedded_test.go`)

Verified to fail on current `main` (unknown `--reason` flag) except the marked guard:

- `close_eligible_reason_flag_persists` — `--reason` value lands in `close_reason`
- `close_eligible_dry_run_shows_reason` — preview shows the reason, epic stays open
- `close_eligible_json_includes_reason` — JSON `reason` key + closed IDs
- `close_eligible_default_reason_unchanged` — green-by-design guard that the default is untouched

Full `TestEmbeddedEpic` suite passes (`BEADS_TEST_EMBEDDED_DOLT=1`). `golangci-lint run ./cmd/bd/`: clean.

Related: #3816 (close-eligible pinning) is adjacent but distinct and not touched here.

### DIFF
--- a/cmd/bd/epic.go
+++ b/cmd/bd/epic.go
@@ -2,12 +2,18 @@ package main
 
 import (
 	"fmt"
+	"os"
+	"strings"
+
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/beads/internal/metrics"
 	"github.com/steveyegge/beads/internal/types"
 	"github.com/steveyegge/beads/internal/ui"
-	"os"
 )
+
+// defaultCloseEligibleReason is the close reason `bd epic close-eligible` has
+// always used; --reason overrides it (GH#4817).
+const defaultCloseEligibleReason = "All children completed"
 
 var epicCmd = &cobra.Command{
 	Use:     "epic",
@@ -90,6 +96,14 @@ func filterEligibleEpics(epics []*types.EpicStatus) []*types.EpicStatus {
 	return filtered
 }
 
+func resolveCloseEligibleReason(cmd *cobra.Command) string {
+	reason, _ := cmd.Flags().GetString("reason")
+	if strings.TrimSpace(reason) == "" {
+		return defaultCloseEligibleReason
+	}
+	return reason
+}
+
 var closeEligibleEpicsCmd = &cobra.Command{
 	Use:           "close-eligible",
 	Short:         "Close epics where all children are complete",
@@ -104,9 +118,10 @@ var closeEligibleEpicsCmd = &cobra.Command{
 		}()
 
 		dryRun, _ := cmd.Flags().GetBool("dry-run")
+		reason := resolveCloseEligibleReason(cmd)
 
 		if usesProxiedServer() {
-			return runCloseEligibleEpicsProxiedServer(rootCtx, dryRun)
+			return runCloseEligibleEpicsProxiedServer(rootCtx, dryRun, reason)
 		}
 
 		if !dryRun {
@@ -121,11 +136,11 @@ var closeEligibleEpicsCmd = &cobra.Command{
 			return outputNoEligibleEpics()
 		}
 		if dryRun {
-			return outputCloseEligibleDryRun(eligibleEpics)
+			return outputCloseEligibleDryRun(eligibleEpics, reason)
 		}
 		closedIDs := []string{}
 		for _, epicStatus := range eligibleEpics {
-			err := store.CloseIssue(rootCtx, epicStatus.Epic.ID, "All children completed", "system", "")
+			err := store.CloseIssue(rootCtx, epicStatus.Epic.ID, reason, "system", "")
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Error closing %s: %v\n", epicStatus.Epic.ID, err)
 				continue
@@ -135,7 +150,7 @@ var closeEligibleEpicsCmd = &cobra.Command{
 		if len(closedIDs) > 0 {
 			commandDidWrite.Store(true)
 		}
-		return outputCloseEligibleResult(closedIDs)
+		return outputCloseEligibleResult(closedIDs, reason)
 	},
 }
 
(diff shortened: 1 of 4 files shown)


kevglynn:
Rebased onto current `main`.

Notes from the rebase:
- Main had refactored `cmd/bd/epic.go` (helpers + proxied-server path). `--reason` is now threaded through both the embedded and proxied `close-eligible` paths.
- Committed CLI docs are release-pinned via `docs/cli-docs.pin` (`v1.1.0`), so this PR no longer touches `docs/CLI_REFERENCE.md` / `docs/cli-reference/epic.md` — the flag will show up in help/docs when the next release regenerates them.

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified at the PR head). Verdict: **needs changes** — the flag is wanted and both mode wirings exist; four consistency gaps before landing.

**Should-fix 1 — `validation.on-close` is bypassed.** Verified: `epic.go` resolves the reason and sends it straight to storage with no validation reference; ordinary `bd close` validates the reason in both direct and proxied modes. With `validation.on-close=error`, `bd close --reason x` fails while `bd epic close-eligible --reason x` persists `x`. Run the same validator (the default `"All children completed"` already passes).

**Should-fix 2 — the promised non-dry JSON shape isn't total.** The commit promises `{"closed","count","reason"}`, but the nothing-eligible path returns `[]` via `outputNoEligibleEpics` on both routes. Return the documented object (with `count: 0`) or document the alternate shape.

**Should-fix 3 — the reason isn't on the final output line** (epic IDs print after the reason-bearing summary), and there's no non-dry human-output assertion to pin whichever contract you intend.

**Should-fix 4 — proxied behavior has zero reason coverage.** The proxied persistence trace exists (UOW → domain issue → db write) but every new reason test is embedded-only, and the existing proxied test omits `reason` entirely. One proxied assertion on persisted `CloseReason` closes it.

Nit: the four repeated epic/child/close fixtures want a helper or table. Drift is clean (the guarded-close work changed ordinary close, not epic closure); generated docs correctly untouched under the release pin.

kevglynn:
Addressed all 4 review items. Rebased onto current main.

**1. validation.on-close is no longer bypassed** — `validateCloseReasons([]string{reason})` is now called before either the embedded or proxied close path runs, matching `bd close` behavior.

**2. JSON shape is now total** — the nothing-eligible path returns `{"closed":[],"count":0,"reason":"..."}` instead of `[]`, so the JSON contract is consistent regardless of whether epics were found.

**3. Reason is on the final output line** — human output now lists epic IDs first, then prints the `✓ Closed N epic(s) with reason "..."` summary on the last line. Added `close_eligible_human_output_ends_with_reason` test to pin this contract.

**4. Proxied-mode reason coverage added** — `write_closes_with_reason` in `epic_proxied_integration_test.go` verifies `--reason` persists to `CloseReason` through the full UOW transaction path.

Build compiles clean (`CGO_ENABLED=0 go build -tags gms_pure_go ./cmd/bd/...`).

maphew (approved):
All four items landed as described: validateCloseReasons ahead of both close paths, the total JSON shape on the nothing-eligible branch, the summary-line ordering, and the proxied parity with its integration test. Approving; merge patrol takes it from here. _claude-fable-5-medium on behalf of maphew_

PR #1228, opened 2026-01-21
fix(list): stabilize tree ordering for consistent --tree output

mrsimpson:
## Motivation

The `bd list --tree` command produced non-deterministic ordering between consecutive executions, making `watch bd list --tree` unusable due to constantly changing output. Root issues and children with identical priorities appeared in different orders across runs, breaking the user experience for monitoring workflows.

## Key Changes

- Add `compareIssuesByPriority()` function with primary sort by priority and secondary sort by ID for deterministic behavior
- Apply stable sorting to root issues in `buildIssueTreeWithDeps()` after line 248
- Apply stable sorting to children in `childrenMap` during tree building for complete consistency
- Update `printPrettyTree()` to use same comparison function for uniformity
- Added comprehensive `TestStableTreeOrdering` test with 5-run stability verification covering both root and children ordering

## Side-Effects

- Tree output now consistently orders by priority (P0→P1→P2→P3→P4) with alphabetical ID fallback
- Items with identical priority are sorted deterministically by ID
- `watch bd list --tree` now provides stable, usable output
- Minor performance overhead from sorting (negligible for typical issue counts)
- Enhanced test coverage prevents future regressions
- Fixes pre-existing indentation inconsistencies in test code via gofmt

## Contribution Checklist

- ✅ **Project database excluded**: No `.beads/issues.jsonl` changes included
- ✅ **Both modes supported**: Fix works in both direct and daemon modes  
- ✅ **New tests**: Added `TestStableTreeOrdering` with comprehensive coverage
- ✅ **No regression**: All existing tests pass, verified with `go test ./cmd/bd`
- ✅ **Linting**: Code follows Go standards, formatted with `gofmt` and `goimports`

**Before:**
```
# Multiple runs produce different orders
Run 1: bd-e3r.5, bd-e3r.4, bd-e3r.1, bd-e3r.2, bd-e3r.3
Run 2: bd-e3r.3, bd-e3r.1, bd-e3r.4, bd-e3r.2, bd-e3r.5
```

**After:**
```
# Identical output across all runs
○ bd-e3r ● P2 Responsible-Vibe Development: beads
├── ○ bd-e3r.6 ● P1 Test feature A
├── ○ bd-e3r.7 ● P1 Test feature B  
├── ○ bd-e3r.8 ● P2 Test identical priority A
├── ○ bd-e3r.9 ● P2 Test identical priority B
└── ○ bd-e3r.1 ● P3 Reproduce
```

PR #4821, opened 2026-07-15
feat(list): add --deps to annotate and order bd list --tree by dependencies

rjc123:
## Problem

`bd list --tree` nests parent-child hierarchy only (since #4686). That correctly stopped dependency edges being mis-drawn as hierarchy — but it left them **invisible** in the tree. When an issue depends on, blocks, or waits for another, nothing in `--tree` shows it, and siblings sort by priority/ID even when a real dependency dictates order.

## What this adds

An opt-in `--deps` flag, **off by default** (the default tree and its deliberate priority sort from #1228 are untouched):

- **Annotations** — each node's non-parent-child edges render as distinct rows with a `╌╌▷` glyph that can't be confused with the `├──/└──` connectors. In-view targets get a full row; out-of-view targets collapse to one `↗ N outside this view: …` summary so filtered views (`--parent`/`--type`) stay legible.
- **Ordering** — siblings are topologically sorted by scheduling edges (`depends-on`, `conditional-blocks`, `waits-for`) so a dependency prints above its dependent. Priority→ID tiebreak; cycle-safe (falls back, never drops or hangs).
- **Modes** — bare `--deps` = scheduling family; `--deps=all` also shows knowledge-graph edges (`related`, `discovered-from`, `duplicates`, `supersedes`, `replies-to`). Ordering is always driven by scheduling edges only.

It's the natural complement to #4686: that PR stopped conflating dependencies with hierarchy; this gives them an honest, ordered presentation.

## Testing

Unit tests for the ordering (topological, dependency-overrides-priority, cycle no-drop, out-of-group), edge classification, and rendering (in-view rows, out-of-view collapse, mode filtering). Verified manually on a real 3,000-issue tree at epic and leaf altitude.

_Note: the `docs/cli-reference/backup.md` + `CLI_REFERENCE.md` backup hunks are pre-existing doc drift the mandatory CLI-docs regen picked up, not part of this change._


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified representative claims at the PR head). Verdict: **needs changes** — the tree-dependency rendering design is sound (sibling-topological ordering, cycle fallback, edge classification all check out); wiring breadth and the docs pin need work.

**Blocker (drift) — generated docs violate main's release pin.** Same class as several sibling PRs: `docs/cli-docs.pin` now pins generated CLI docs to `v1.1.0`; the required freshness gate regenerates from the pin and will reject the unreleased `--deps` flag docs (plus the unrelated backup hunks the old full-tree regen swept in). Drop the generated-doc changes; keep the source.

**Should-fix — `--deps` is silently ignored on most paths.** The flag is accepted everywhere but threaded only through the two direct, non-watch render calls: direct `--watch` returns earlier into the mode-less renderer; the proxied parent/watch/refresh/pretty paths all use the mode-less wrapper; and `--deps` with `--flat`/`--json`/`--format` is accepted and does nothing. Thread `depsMode` through every tree renderer or reject unsupported combinations — and add command-level wiring tests (this is the same accept-and-ignore class we've flagged on several other PRs today).

**Should-fix — dependency-read errors render as dependency-free trees.** Both `GetAllDependencyRecords` errors are discarded; a nil map skips ordering and annotations but still prints the enabled legend — the user sees a confident, wrong tree. When `--deps` was requested, return the error.

Standards: reuse the existing `captureStdout` helper (the new local copy drops the stdio mutex and concurrent pipe drain); name the eleven table cases with `t.Run`.

rjc123:
Addressed the review. Dropped the generated CLI-doc commit and rebased onto current `main` — PR is now source-only.

- **Accept-and-ignore:** `--deps` now implies the tree view for a bare `--deps` and is rejected with a clear error alongside `--json`/`--format`/`--flat` (approach: reject the non-tree combinations rather than thread `depsMode` through every renderer — so `--deps --watch` errors too).
- **Swallowed dep-read errors:** now returned when `--deps` was requested, instead of rendering a dependency-free tree under the legend.
- **Tests:** switched to the shared `captureStdout` helper, named the `depEdgeDisplay` cases with `t.Run`, and added command-level wiring tests for the rejected combinations + implies-tree.

Thanks for the thorough review.

rjc123:
The `PR Core (wrapper timing)` red here looks like a flaky test teardown, unrelated to this change:

```
--- FAIL: TestSharedServerCfgNilHonorsSharedServer (0.10s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat .../002: directory not empty
```

The test body itself passes (0.10s, no assertion failure) — the failure is in Go's automatic `t.TempDir()` cleanup: `RemoveAll` hits "directory not empty" because the shared Dolt sql-server the test starts hasn't released its files before the temp dir is torn down. A teardown race, and this test (shared-server config) is unrelated to the `--deps` tree annotation in this PR.

Could you re-run the failed job when you get a chance? I'll also open a separate issue + PR for the underlying teardown race so it stops biting unrelated PRs.

maphew (approved):
Re-review after the 2026-07-24 rebase. Verdict: **approve**.

All four points from the previous review are addressed and verified at the new head:

- **Accept-and-ignore**: `--deps` with `--json`/`--format`/`--flat`/`--watch` is now rejected in `gatherListInput` with clear errors, and a bare `--deps` implies the tree view — with command-level wiring tests for each combination. Reject-not-thread is a reasonable call; it keeps the renderer surface small.
- **Swallowed dep-read errors**: both `GetAllDependencyRecords` errors now propagate when `--deps` was requested.
- **Standards**: shared `captureStdout` helper reused; the `depEdgeDisplay` table cases named with `t.Run`.

One residual gap from the earlier review remained: the two **proxied-server tree paths** (hierarchical `--parent` and the pretty list in `renderProxiedListText`) still called the mode-less wrapper, so `--deps` was accepted and silently ignored under a proxied deployment. Rather than send this back for another round, I pushed a two-line maintainer commit to your branch threading `in.depsMode` through both call sites (9e1ca5fc3) — the deps map those paths already load is keyed identically and their load errors already propagate. Build + the deps/wiring tests pass locally. Please shout if you'd rather it were done differently.

On the red `PR Core (wrapper timing)` check: verified from the job log it is exactly the `TestSharedServerCfgNilHonorsSharedServer` TempDir teardown race you root-caused — which your own #5032 fixes. Nice cross-PR closure. Merge tail: this PR merges once the fresh CI run on the new head is green (the teardown flake may still bite until #5032 lands; a rerun clears it).

Why approve: the feature is the honest complement to #4686 (dependencies stopped masquerading as hierarchy there; here they get a legible, ordered, opt-in presentation), the ordering is cycle-safe and default-preserving, and the review loop converged.

PR #4823, opened 2026-07-15
Adopt server identity in bd init against an authenticating gateway

julianknutsen:
## What

Makes `bd init --server` work correctly against an authenticating SQL gateway: it routes through the gateway credential machinery, adopts the server's project identity instead of generating one, and stops writing client-local state into the shared hosted database. Plain (embedded / non-gateway) `bd init` is byte-identical to before.

## Why

Today `bd init --server` against a gateway connects as `root` and takes the create-database path, and — worse — the gateway credential command ran *unconditionally*, so an ambient `BEADS_DOLT_CREDENTIAL_COMMAND` in the environment broke plain embedded `bd init` (it would fail with a spurious "hosted database" error and leave a half-initialized `.beads/`). This change gates the credential command on server mode (mirroring the canonical open path in `internal/storage/dolt/open.go`) and has gateway init adopt the server-authoritative `_project_id` / `issue_prefix` rather than minting a divergent identity. It also skips the per-client metadata writes (`repo_id`, `clone_id`, `bd_version`, `last_import_time`) and the initial-state commit in gateway mode, so multiple clients don't stomp a shared hosted database.

## Verification

```
make build                                                    # PASS
go vet -tags gms_pure_go ./cmd/bd/                            # clean
go test -tags gms_pure_go ./cmd/bd/ -run \
  'TestApplyInitGatewayCredential|TestResolveInitIssuePrefix|TestResolveInitProjectID|TestShouldWriteProjectIDLocally|TestShouldWriteInitStateToDB' -v   # 17/17 PASS
```

Includes `TestApplyInitGatewayCredentialSkipsEmbeddedMode` (the server-mode gate — a credential command set with `ServerMode=false` is not run and the config is untouched) and a non-gateway-identical guard. Reproduced the fix by hand: `BEADS_DOLT_CREDENTIAL_COMMAND=false bd init -p t --stealth --quiet` in a fresh dir now exits 0 (command not executed) instead of erroring.

Every existing `TestInit*` failure is a pre-existing shared-test-dolt-server contention issue ("database not found on Dolt server"), reproduced identically on `origin/main` with this change reverted; the changed code is covered by the scoped tests above.

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -32,6 +32,133 @@ import (
 	"golang.org/x/term"
 )
 
+// applyInitGatewayCredential routes the hand-built init Dolt config through the
+// gateway credential machinery. When BEADS_DOLT_CREDENTIAL_COMMAND is configured,
+// its short-lived token becomes the connection username, the config is marked as
+// targeting an authenticating gateway server (so the store skips schema init and
+// the SHOW/CREATE DATABASE probe), and local auto-start is disabled — a gateway
+// server is externally managed, and spawning a local dolt server would shadow it.
+//
+// It runs only when the config targets a server (ServerMode). An embedded init
+// never presents a username, so an ambient credential command must not run — or
+// fail — an embedded open. This mirrors the canonical open path
+// (internal/storage/dolt/open.go), which gates the command on
+// IsDoltServerMode()||IsSharedServerMode(). Without this gate, a plain embedded
+// `bd init` on a host that exports BEADS_DOLT_CREDENTIAL_COMMAND would drag the
+// gateway machinery into an ordinary local init and abort it with a misleading
+// provisioning-contract error, leaving a half-initialized .beads/. Gateway init
+// always reaches here with ServerMode set: init forces server mode for a shared
+// server and hard-fails a remote dolt host that lacks server mode.
+//
+// The main command path applies this via applyResolvedConfig; init builds its own
+// config and must mirror the behavior. Within server mode it is still a no-op when
+// no command is configured or when a --server-user flag already preset the
+// username. It fails closed: a configured-but-failing command aborts init.
+func applyInitGatewayCredential(ctx context.Context, beadsDir string, doltCfg *dolt.Config) error {
+	if !doltCfg.ServerMode {
+		return nil
+	}
+	fileCfg, _ := configfile.Load(beadsDir)
+	if fileCfg == nil {
+		fileCfg = configfile.DefaultConfig()
+	}
+	applied, err := dolt.ApplyGatewayCredential(ctx, fileCfg, doltCfg)
+	if err != nil {
+		return err
+	}
+	if applied {
+		// ApplyGatewayCredential sets DisableAutoStart, but the hand-built init
+		// config path never runs ApplyCLIAutoStart to translate that into the
+		// AutoStart field the store's auto-start block actually consults.
+		doltCfg.AutoStart = false
+	}
+	return nil
+}
+
+// resolveInitIssuePrefix decides how init sets the issue_prefix. readErr is the
+// error (if any) from reading the current issue_prefix out of the database.
+//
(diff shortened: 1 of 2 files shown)


julianknutsen:
Field report from running this branch end-to-end against a Dolt server behind an authenticating TLS-required gateway: the core flow works — `bd init --server` adopts the server-side project identity, and list/create work over TLS with the credential command, no CA bundle needed.

One non-blocking follow-up surfaced: the `cfg.Gateway` short-circuit in `internal/storage/dolt/store.go` covers the main store path, but `bd doctor`'s checks (Sync Staleness / Federation Conflicts / Dolt Mode) and the init post-check rebuild their store config without the gateway/TLS settings, so they fall into the legacy `databaseExistsOnServer` probe on a plain-TCP DSN and report `Error 1045 (28000): TLS required` — i.e. a fresh, perfectly working init prints "Setup incomplete: Failed to connect". Cosmetic (the workspace works), but it lands on exactly the first-run path this PR enables. Happy to send a follow-up PR threading the resolved server config into those checks rather than widening this one.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes `bd init --server` work correctly against an authenticating SQL gateway — it routes through the gateway credential machinery, adopts the server's project identity instead of minting a divergent one, and stops writing client-local state into the shared hosted database. Just as valuable, it fixes a real embedded-mode regression along the way: an ambient `BEADS_DOLT_CREDENTIAL_COMMAND` no longer breaks plain `bd init`. The non-gateway path stays byte-identical, so this hardens hosted-gateway init and multi-client safety without touching the embedded workflow.

### Original PR Review
Decision: approve.

Specific gaps validated:
- **Stale local project identity on gateway re-init** — `bd init` now always consults `_project_id` in gateway mode and reconciles the server-authoritative value over stale local metadata, so a re-init can't diverge from the server's identity.
- **Ambient credential command on embedded init** — `applyInitGatewayCredential` exits before running the command when `ServerMode` is false, so an ambient `BEADS_DOLT_CREDENTIAL_COMMAND` no longer breaks plain embedded `bd init` or leaves a half-initialized `.beads/`.
- **Gateway mode writing server-owned state** — schema init, database creation/probing, remote writes, clone-local metadata, `_project_id` write-back, shared-global provisioning, and the initial Dolt commit are all gated off under gateway mode, so multiple clients no longer stomp a shared hosted database.
- **Transient reads misreported as provisioning violations** — `resolveInitIssuePrefix` and `resolveInitProjectID` now distinguish missing keys from read errors and surface read failures as transient errors rather than false provisioning failures.
- **Gateway auto-start suppression** — the hand-built init config translates `DisableAutoStart` into `AutoStart=false` after credential adoption.

Review findings addressed:
- Test Evidence Quality (minor, high confidence): the new tests cover the pure helper behavior well; a storage-backed regression for the absent-key (`"", nil`) contract that gateway resolution relies on is recorded as optional follow-up coverage.
- Change Impact / Blast Radius (minor, low confidence): in gateway + shared-server mode, init still advertises `beads_global` metadata it deliberately does not provision; recorded as an optional follow-up to either gate or document that write.

Remaining non-gating notes:
- Add a storage-backed regression test proving absent `issue_prefix` / `_project_id` reads return empty values without errors.
- Confirm/document that gateway shared-server deployments never consult the locally advertised `beads_global`, or gate that metadata write under the same gateway-aware condition used by `shouldInitSharedGlobalDB`.
- No blocking or major issues were found; both notes above are optional and do not gate the merge.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `6ce51f64` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/29677229809

### Review Iterations
1 review pass performed; approved with no fix iterations required.

---

PR #4825, opened 2026-07-15
docs: regenerate backup help

maphew:
## Why

`main` is red at [Actions run 29434366508](https://github.com/gastownhall/beads/actions/runs/29434366508): the strict CLI-doc freshness check finds the recently added `bd backup` auto-backup help in the live CLI but not in the generated reference. This blocks every non-main-fix PR.

## Change

Regenerate the two canonical CLI reference artifacts. The update records the auto-backup default: enabled for embedded mode with a Git remote, disabled for shared server mode unless explicitly enabled.

## Verification

- `./scripts/generate-cli-docs.sh --check <CGO_ENABLED=0 -tags gms_pure_go binary>`
- `./scripts/check-doc-flags.sh <same binary>`
- `git diff --check`

The doc-flags check retains its existing legacy-SQLite prose warnings but exits successfully.

### DIFF
--- a/docs/CLI_REFERENCE.md
+++ b/docs/CLI_REFERENCE.md
@@ -2358,6 +2358,16 @@ DoltHub is recommended for cloud backup:
   bd backup init https://doltremoteapi.dolthub.com/&lt;user&gt;/&lt;repo&gt;
   Set DOLT_REMOTE_USER and DOLT_REMOTE_PASSWORD for authentication.
 
+Auto-backup default:
+  When backup.enabled is unset, auto-backup turns ON in embedded mode if a
+  git remote exists, and stays OFF in sql-server / shared-server mode. In
+  server mode many bd clients share one Dolt server, and each would register
+  a server-side backup remote under the same name pointing at its own local
+  dir and full-sync the whole database — a self-amplifying storm. To back up
+  a shared server, run 'bd backup' explicitly (or set backup.enabled=true and
+  coordinate destinations). 'bd config get backup.enabled' shows the effective
+  value and its source.
+
 ```
 bd backup [command]
 ```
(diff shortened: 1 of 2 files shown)


maphew:
## Review

### Overview
Docs-only stop-the-line fix. Main went red because the strict CLI-doc freshness check ([Actions run 29434366508](https://github.com/gastownhall/beads/actions/runs/29434366508)) found the recently added `bd backup` auto-backup help in the live binary but not in the generated reference. This PR regenerates the two canonical artifacts (`docs/CLI_REFERENCE.md` and `docs/cli-reference/backup.md`), adding the identical 10-line "Auto-backup default" block to each. No hand-written prose, no source changes.

### Correctness
- ✅ **Verbatim match with source of truth.** The new block is character-identical to the `Long` string in `cmd/bd/backup.go` (embedded-mode ON with git remote, server-mode OFF, the self-amplifying-storm rationale, the `bd config get backup.enabled` pointer). This is a faithful regeneration, not a manual edit.
- ✅ **Both mirrors updated consistently.** The monolithic reference and the per-command page carry the same block — no drift between the two generated artifacts.
- ✅ **CI confirms the fix works**: "Check doc flags freshness" passes on the PR, along with all other completed checks at review time; remaining heavy checks (Build Artifacts, Contract corpus, Storage backend conformance) were still pending, none failed.

### Conventions
- ✅ **Pure-Go generation was used correctly.** The diff contains zero federation churn — the tell-tale of a CGO build — consistent with `CGO_ENABLED=0 -tags gms_pure_go` generation. The PR body confirms `generate-cli-docs.sh --check` and `check-doc-flags.sh` were run against the right binary.
- ✅ **PR body answers "why"** and documents verification.
- ✅ **Correct stop-the-line posture**: while main is red, the only mergeable PR is the main fix — this is that fix, appropriately minimal and scoped.

### Minor observations (no action needed)
- The `&lt;user&gt;`/`&lt;repo&gt;` HTML entities visible in context lines are pre-existing generator escaping, consistent across both files — not introduced here.
- The em-dash in "self-amplifying storm" comes from the Go source string; keeping it verbatim is correct for a generated artifact (the freshness check would flag any normalization).

### Verdict
**Approve.** This is exactly what a doc-freshness main fix should look like: mechanically generated, verified against the binary, matching source verbatim, and green on the check it exists to fix. The only gate before merge is letting the pending heavy checks finish.

Issue #4638, opened 2026-07-07
Windows test-suite portability: NTFS exec bit, Signal(0), USERPROFILE vs HOME, worktree walk-up boundary


seanmartinsmith:
## problem
five independent Windows test-portability failures on upstream/main, all pre-existing (not introduced by any specific feature branch), all narrow/self-contained. bundleable into one PR — each item below is either a small windows-aware branch in production code or a test-harness-only fix; none require touching unrelated code paths.

## versions
bd v1.1.0 (commit 8e4e59d39); confirmed present at main @187ba85f3. Windows

### 1. NTFS has no exec bit — `HookExists`/`Run`/`RunSync` check `Mode()&0111`, always false on Windows
- location: `internal/hooks/hooks.go:110` (`HookExists`, the check `return info.Mode()&0111 != 0`), also `:64` (`Run`) and `:90` (`RunSync`) apply the same check.
- test: `TestHookExists_Executable`, `internal/hooks/hooks_test.go:86-99` — creates a hook file via `os.WriteFile(path, ..., 0755)`; NTFS doesn't honor unix perm-bit semantics, so the requested 0755 never produces an exec bit Go can observe, and `Mode()&0111` is always 0 on windows regardless of the file's actual usability as a hook.
- fix: skip the POSIX-mode exec check on windows (`runtime.GOOS == "windows"` branch, or a `_windows.go`/`_unix.go` split like `doltserver` already uses) and treat "file exists, not a directory" as sufficient — or align with whatever interpreter/shebang dispatch logic the hook runner (`runHook`) actually uses on windows, so `HookExists` and the dispatcher agree on what counts as runnable.

### 2. `isServerProbablyRunning` uses `Signal(0)`, unsupported on Windows
- location: `cmd/bd/config_drift.go:324` — `err = proc.Signal(syscall.Signal(0))`. Go's windows `os.Process.Signal` only supports `os.Kill`; any other signal (including `Signal(0)`, the standard unix liveness probe) returns an "not supported by windows" error, so `err == nil` is never true there.
- test: `TestIsServerProbablyRunningOwnPID`, `cmd/bd/config_drift_test.go:95-108` — expects `true` for the test's own live PID; fails on windows because the `Signal(0)` probe always errors.
[...]

PR #4828, opened 2026-07-15
fix(metrics): move telemetry queue out of workspace state (#4807)

### LINKED ISSUE #4807: bug(metrics): DataDir() hardcodes ~/.beads, ignores BEADS_DIR — creates phantom directory that confuses discovery
I could not determine if this was by oversight or by design.

I'm busy automating the `bd` setup for our org and ran into this issue, as I would like a clean home directory and for tools to use XDG paths, where possible. A little stubborn of me. Bumped my head against this dual directory issue and got the digital twin to deep dive it.

## Summary

`internal/metrics/metrics.go` hardcodes the telemetry data directory to `$HOME/.beads/eventsData`, ignoring the `BEADS_DIR` environment variable. When a user sets `BEADS_DIR` to a non-default location (e.g. `~/.local/share/beads`), every `bd` invocation creates `~/.beads/eventsData/` as a side effect. This phantom `~/.beads/` directory then interferes with workspace discovery in subsequent commands.

## Versions

- bd **v1.1.0** (commit `8e4e59d39`)
- Still present on **main** (`internal/metrics/metrics.go` unchanged)
- linux + macOS (path-agnostic)

## Root cause

`internal/metrics/metrics.go:15,37-41`:

```go
const dataDirName = ".beads"

func DataDir() (string, error) {
    home, err := os.UserHomeDir()
    if err != nil { return "", err }
    return filepath.Join(home, dataDirName, "eventsData"), nil
}
```

`DataDir()` always returns `$HOME/.beads/eventsData` regardless of `BEADS_DIR`. Every command that initializes telemetry creates this directory.

## Impact

When `BEADS_DIR` points elsewhere (e.g. `~/.local/share/beads`), a phantom `~/.beads/` is created containing only `eventsData/`. This directory has no `config.yaml`, `metadata.json`, or database — but its existence is enough to confuse discovery:

- `bd config set` (and potentially other commands that resolve workspace via CWD walk) can find `~/.beads`, see no `config.yaml`, and fail with: `Error: setting config: no config.yaml found in BEADS_DIR (/root/.beads)`
- The error message itself is misleading — it reports `BEADS_DIR` as `~/.beads` even though the actual env var points elsewhere, suggesting that the phantom directory is being picked up and overriding the env var somewhere in the discovery chain

Note: PR #3104 (`fix(config): respect BEADS_DIR for YAML-only config resolution`) fixed `findProjectConfigYaml` to check `$BEADS_DIR` first. But the phantom `~/.beads/` created by telemetry appears to interfere with earlier discovery stages (e.g. `loadEnvironment` → `FindBeadsDir` CWD walk) that run before `findProjectConfigYaml` is called.

## Repro

```sh
# Clean slate
rm -rf ~/.beads ~/.local/share/beads

export BEADS_DIR=~/.local/share/beads
cd $HOME

# Init works — data goes to BEADS_DIR
bd init --prefix home --stealth --non-interactive --skip-agents --skip-hooks

# Verify: data is at BEADS_DIR, but ~/.beads/eventsData also exists
ls ~/.local/share/beads/config.yaml   # exists (correct)
ls ~/.beads/eventsData/               # exists (side effect)
ls ~/.beads/config.yaml               # does not exist

# This fails:
bd config set export.auto false
# Error: setting config: no config.yaml found in BEADS_DIR (/root/.beads) (run 'bd init' first)
```

## Suggested fix

`DataDir()` should respect `BEADS_DIR` when set:

```go
func DataDir() (string, error) {
    if beadsDir := os.Getenv("BEADS_DIR"); beadsDir != "" {
        return filepath.Join(beadsDir, "eventsData"), nil
    }
    home, err := os.UserHomeDir()
    if err != nil { return "", err }
    return filepath.Join(home, dataDirName, "eventsData"), nil
}
```

Alternatively, telemetry could use a fully separate path (e.g. `$XDG_STATE_HOME/beads/eventsData`) that never overlaps with the `.beads` project directory — but the simpler fix is to follow `BEADS_DIR`.

## Related

- PR #3104 — `fix(config): respect BEADS_DIR for YAML-only config resolution` (merged, in v1.0.2+) — fixed one layer but this telemetry side-effect creates the precondition that defeats it
- Issue #4635 — `bd init creates a git repo + agent scaffolding in an unintended dir` (open) — different root cause but same theme of `$HOME` being treated as a safe default target

### LINKED ISSUE #4937: Windows: metrics tests can modify the real user profile
## Summary

Several `internal/metrics` tests try to isolate profile writes by setting only `HOME`. On Windows, Go's `os.UserHomeDir()` uses `USERPROFILE`, while `os.UserConfigDir()` uses `APPDATA`. The tests therefore read and write the developer's real profile instead of their `t.TempDir()` fixtures.

This is both a portability failure and a test-safety defect: a focused run on current `upstream/main` modified the existing `%USERPROFILE%\.config\bd\config.yaml` and created two `.evtq` files under `%USERPROFILE%\.beads\eventsData`.

## Environment and affected tests

Observed on Windows amd64 with Go 1.26.5 at `upstream/main` commit `2a63eb3182f7f878893da36c99dd3b5d48f60206`.

Affected tests include:

- `TestInitEnabledFlipsEnabledTrue`
- `TestCloseAndFlushPersistsQueuedEvents`
- the `TestEnsureUserConfigDefaults_*` suite

The same HOME-only setup appears in disabled/flusher tests and should be fixed consistently so future behavior changes cannot leak writes.

Do not reproduce this on a Windows workstation without first using a disposable user profile or explicitly redirecting `USERPROFILE` and `APPDATA`; the defect is that the tests themselves do not provide that isolation.

## Root cause

`metrics.DataDir()` calls `os.UserHomeDir()`, and `config.UserConfigYamlPath()` combines `os.UserHomeDir()` with `os.UserConfigDir()`. Setting `HOME` alone redirects those APIs on Unix but not Windows.

## Expected behavior

The package test process must never touch a developer's real profile. A package-level `TestMain` should redirect `HOME`, `USERPROFILE`, `HOMEDRIVE`/`HOMEPATH`, `XDG_CONFIG_HOME`, `APPDATA`, `LOCALAPPDATA`, and inherited `BEADS_DIR` before any test runs. Tests that assert filesystem output should additionally use a per-test helper with a unique profile root.

## Verification

With that isolation in place, the complete package passes with `-race` on Windows. When run under an external profile sentinel, its marker remains unchanged, no `.config`, `.beads`, `bd` app-data, or inherited `BEADS_DIR` path is created, and the package-level disposable profile is removed after the run.

## Related

#4638 describes the general `HOME` versus `USERPROFILE` test-harness problem for other packages. PR #4828 adds metrics `DataDir` coverage and still contains a HOME-only default-path test; it should use the same helper or equivalent isolation.

vishnujayvel:
## Summary

`internal/metrics.DataDir()` always wrote telemetry under `$HOME/.beads/eventsData`. Every
metrics-enabled `bd` run could create a phantom `~/.beads/` that interfered with workspace
discovery when the real database lived elsewhere.

This change moves the telemetry queue to `~/.config/bd/eventsData` — a pure per-user path that
never depends on, and never creates, a workspace directory. `BEADS_DIR` is no longer consulted for
the queue at all: an earlier revision of this PR made the queue follow `BEADS_DIR`, but review
established that tying telemetry to workspace selection was the underlying bug rather than the
fix, so the queue is now resolved independently of the workspace entirely.

Because the queue path is now fixed and machine-scoped rather than workspace-derived, `Init` can
construct the file emitter unconditionally and immediately — there is no ordering dependency on
workspace selection left to defer around. `TestInitEnabledCreatesDataDirImmediately` pins that
behavior.

Fixes #4807.
Fixes #4937.

Thanks to the issue reporters for the clear repros and suggested shape, and to the reviewers whose
design ruling reshaped this from a `BEADS_DIR` passthrough into the `~/.config/bd` move.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./internal/metrics/... ./cmd/bd` — clean
- [x] `go test ./internal/metrics/...` — pass
- [x] `go test -run 'Metric|Telemetry' ./cmd/bd/` — pass
- [x] Profile isolation (#4937): `TestMain` + `isolateUserProfile(t)` in
      `internal/metrics/testenv_test.go` redirect every platform home/config lookup
      (`HOME`, `USERPROFILE`, `XDG_CONFIG_HOME`, `APPDATA`, `LOCALAPPDATA`, `HOMEDRIVE`,
      `BEADS_DIR`) into a test-owned temp dir, so the package can no longer read or write the
      developer's real profile.
- [x] `cmd/bd` per-test profile normalization (@ecuthiell's `facd1890f`, absorbed): `metricsTestEnv`
      now overrides the full profile environment rather than `HOME` alone, so CGO-enabled Windows
      runs no longer have subprocesses queue events under the suite-wide profile root while
      assertions inspect the per-test one.
- [x] Rebased onto `main` @ `634cbbc4b`; conflicts resolved keep-both, no commits dropped.

### DIFF
--- a/AGENT_INSTRUCTIONS.md
+++ b/AGENT_INSTRUCTIONS.md
@@ -503,9 +503,11 @@ gh issue view 201
 record carrying only the command name; each batch also carries the bd version
 and OS platform, keyed by a machine-derived, HMAC-protected distinct ID. No
 email, repo path, remote URL, issue content, or user-supplied strings are
-collected. Events are written under `~/.beads/eventsData` by default, or under
-`$BEADS_DIR/eventsData` once the workspace is resolved (via `-C`, `.beads/.env`,
-or `BEADS_DIR`), and POSTed to `https://gastownhall-eventsapi.com/mp/collect`.
+collected. Events are written under `$BEADS_DIR/eventsData` once the workspace is
+resolved (via `-C`, `.beads/.env`, or `BEADS_DIR`) and that directory exists,
+and under `~/.beads/eventsData` otherwise — telemetry never creates a workspace
+that a command did not. They are POSTed to
+`https://gastownhall-eventsapi.com/mp/collect`.
 
 Metrics are enabled by default (opt-out). The friendliest way to see or change
 them is `bd metrics` (`bd metrics on` / `bd metrics off` / `bd metrics example`),
--- a/AGENT_INSTRUCTIONS.md
+++ b/AGENT_INSTRUCTIONS.md
@@ -503,9 +503,9 @@ gh issue view 201
 record carrying only the command name; each batch also carries the bd version
 and OS platform, keyed by a machine-derived, HMAC-protected distinct ID. No
 email, repo path, remote URL, issue content, or user-supplied strings are
-collected. Events are written under `~/.beads/eventsData` by default, or
-under `$BEADS_DIR/eventsData` when `BEADS_DIR` is set, and POSTed to
-`https://gastownhall-eventsapi.com/mp/collect`.
+collected. Events are written under `~/.beads/eventsData` by default, or under
+`$BEADS_DIR/eventsData` once the workspace is resolved (via `-C`, `.beads/.env`,
+or `BEADS_DIR`), and POSTed to `https://gastownhall-eventsapi.com/mp/collect`.
 
 Metrics are enabled by default (opt-out). The friendliest way to see or change
 them is `bd metrics` (`bd metrics on` / `bd metrics off` / `bd metrics example`),
--- a/AGENT_INSTRUCTIONS.md
+++ b/AGENT_INSTRUCTIONS.md
@@ -503,10 +503,10 @@ gh issue view 201
 record carrying only the command name; each batch also carries the bd version
 and OS platform, keyed by a machine-derived, HMAC-protected distinct ID. No
 email, repo path, remote URL, issue content, or user-supplied strings are
-collected. Events are written under `$BEADS_DIR/eventsData` once the workspace is
-resolved (via `-C`, `.beads/.env`, or `BEADS_DIR`) and that directory exists,
-and under `~/.beads/eventsData` otherwise — telemetry never creates a workspace
-that a command did not. They are POSTed to
+collected. Events are written to a single machine-scoped queue beside the
+user-global config, `~/.config/bd/eventsData` — not under any workspace's
+`.beads/` directory, and independent of `BEADS_DIR` — so telemetry never
+creates or depends on a workspace. They are POSTed to
 `https://gastownhall-eventsapi.com/mp/collect`.
 
 Metrics are enabled by default (opt-out). The friendliest way to see or change
--- a/AGENT_INSTRUCTIONS.md
+++ b/AGENT_INSTRUCTIONS.md
@@ -503,7 +503,8 @@ gh issue view 201
 record carrying only the command name; each batch also carries the bd version
 and OS platform, keyed by a machine-derived, HMAC-protected distinct ID. No
 email, repo path, remote URL, issue content, or user-supplied strings are
-collected. Events are written under `~/.beads/eventsData` and POSTed to
+collected. Events are written under `~/.beads/eventsData` by default, or
+under `$BEADS_DIR/eventsData` when `BEADS_DIR` is set, and POSTed to
 `https://gastownhall-eventsapi.com/mp/collect`.
 
 Metrics are enabled by default (opt-out). The friendliest way to see or change
--- a/cmd/bd/init_metrics_test.go
+++ b/cmd/bd/init_metrics_test.go
@@ -27,7 +27,7 @@ type metricsEvent struct {
 
 func readInitEvent(t *testing.T, home, expectedCommand string) metricsEvent {
 	t.Helper()
-	dir := filepath.Join(home, ".beads", "eventsData")
+	dir := filepath.Join(home, ".config", "bd", "eventsData")
 	entries, err := os.ReadDir(dir)
 	if err != nil {
 		t.Fatalf("read eventsData dir: %v", err)
@@ -188,7 +188,7 @@ func writeUserMetricsDisabled(t *testing.T, home string, disabled bool) {
 
 func evtqFilesIn(t *testing.T, home string) []string {
 	t.Helper()
-	dir := filepath.Join(home, ".beads", "eventsData")
+	dir := filepath.Join(home, ".config", "bd", "eventsData")
 	entries, err := os.ReadDir(dir)
 	if err != nil {
 		return nil
@@ -355,7 +355,7 @@ func TestInitMetricsDisabledSuppresses(t *testing.T) {
 		t.Fatalf("bd init failed: %v\n%s\n%s", err, stdout.String(), stderr.String())
 	}
 
-	dir := filepath.Join(home, ".beads", "eventsData")
+	dir := filepath.Join(home, ".config", "bd", "eventsData")
 	if _, err := os.Stat(dir); err == nil {
 		entries, _ := os.ReadDir(dir)
 		var evtqs []string
@@ -376,7 +376,7 @@ func TestInitMetricsDisabledSuppresses(t *testing.T) {
 // one cli_command event).
 func allCommandEvents(t *testing.T, home string) []string {
 	t.Helper()
-	dir := filepath.Join(home, ".beads", "eventsData")
+	dir := filepath.Join(home, ".config", "bd", "eventsData")
 	entries, err := os.ReadDir(dir)
 	if err != nil {
 		return nil
@@ -442,7 +442,7 @@ func TestMetricsTodoAliasEmitsSingleEvent(t *testing.T) {
 	}
 
 	// Isolate the next invocation by dropping init's queued event.
-	if err := os.RemoveAll(filepath.Join(home, ".beads", "eventsData")); err != nil {
+	if err := os.RemoveAll(filepath.Join(home, ".config", "bd", "eventsData")); err != nil {
 		t.Fatalf("clear eventsData: %v", err)
 	}
 
@@ -476,7 +476,7 @@ func TestMetricsReadyGatedAliasEmitsSingleEvent(t *testing.T) {
 	}
 
 	// Isolate the next invocation by dropping init's queued event.
-	if err := os.RemoveAll(filepath.Join(home, ".beads", "eventsData")); err != nil {
+	if err := os.RemoveAll(filepath.Join(home, ".config", "bd", "eventsData")); err != nil {
 		t.Fatalf("clear eventsData: %v", err)
 	}
 
@@ -511,7 +511,7 @@ func TestMetricsWispAliasEmitsSingleEvent(t *testing.T) {
 	}
 
 	// Isolate the next invocation by dropping init's queued event.
-	if err := os.RemoveAll(filepath.Join(home, ".beads", "eventsData")); err != nil {
+	if err := os.RemoveAll(filepath.Join(home, ".config", "bd", "eventsData")); err != nil {
 		t.Fatalf("clear eventsData: %v", err)
 	}
 
@@ -603,7 +603,7 @@ func TestMetricsOffEmitsNoEvent(t *testing.T) {
 	if got := allCommandEvents(t, home); len(got) == 0 {
 		t.Fatalf("precondition: metrics-enabled `bd init` produced no event; env may be misconfigured")
 	}
-	if err := os.RemoveAll(filepath.Join(home, ".beads", "eventsData")); err != nil {
+	if err := os.RemoveAll(filepath.Join(home, ".config", "bd", "eventsData")); err != nil {
 		t.Fatalf("clear eventsData: %v", err)
 	}
 
--- a/cmd/bd/init_metrics_test.go
+++ b/cmd/bd/init_metrics_test.go
@@ -84,14 +84,34 @@ func (e metricsEvent) attr(key string) (string, bool) {
 
 func metricsTestEnv(home string, extra ...string) []string {
 	base := bdEnv(home)
-	out := make([]string, 0, len(base)+len(extra)+1)
+	// bdEnv overrides HOME only. Keep the platform-specific profile variables
+	// aligned too, or os.UserHomeDir still resolves USERPROFILE on Windows and
+	// writes the queue outside this test's fixture.
+	out := make([]string, 0, len(base)+len(extra)+8)
 	for _, e := range base {
-		if strings.HasPrefix(e, "BD_DISABLE_METRICS=") || strings.HasPrefix(e, "DO_NOT_TRACK=") {
+		key, _, _ := strings.Cut(e, "=")
+		switch strings.ToUpper(key) {
+		case "HOME", "USERPROFILE", "HOMEDRIVE", "HOMEPATH",
+			"XDG_CONFIG_HOME", "APPDATA", "LOCALAPPDATA",
+			"BD_DISABLE_METRICS", "DO_NOT_TRACK":
 			continue
 		}
 		out = append(out, e)
 	}
-	out = append(out, "BD_DISABLE_EVENT_FLUSH=1")
+	out = append(out,
+		"HOME="+home,
+		"USERPROFILE="+home,
+		"XDG_CONFIG_HOME="+filepath.Join(home, ".config"),
+		"APPDATA="+filepath.Join(home, "AppData", "Roaming"),
+		"LOCALAPPDATA="+filepath.Join(home, "AppData", "Local"),
+		"BD_DISABLE_EVENT_FLUSH=1",
+	)
+	if volume := filepath.VolumeName(home); volume != "" {
+		out = append(out,
+			"HOMEDRIVE="+volume,
+			"HOMEPATH="+strings.TrimPrefix(home, volume),
+		)
+	}
 	return append(out, extra...)
 }
 
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -946,15 +946,6 @@ var rootCmd = &cobra.Command{
 			return err
 		}
 
-		// Attach the metrics emitter now that the workspace is resolved, so
-		// events land under the workspace bd selected rather than whatever
-		// BEADS_DIR held at process start.
-		if dir, err := metrics.DataDir(); err != nil {
-			debug.Logf("warning: resolve metrics data dir failed: %v", err)
-		} else if err := metrics.AttachFileEmitter(dir); err != nil {
-			debug.Logf("warning: attach metrics file emitter failed: %v", err)
-		}
-
 		// Block dangerous env var overrides that could cause data fragmentation (bd-hevyw).
 		if err := checkBlockedEnvVars(); err != nil {
 			return HandleError("%v", err)
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -946,6 +946,15 @@ var rootCmd = &cobra.Command{
 			return err
 		}
 
+		// Attach the metrics emitter now that the workspace is resolved, so
+		// events land under the workspace bd selected rather than whatever
+		// BEADS_DIR held at process start.
+		if dir, err := metrics.DataDir(); err != nil {
+			debug.Logf("warning: resolve metrics data dir failed: %v", err)
+		} else if err := metrics.AttachFileEmitter(dir); err != nil {
+			debug.Logf("warning: attach metrics file emitter failed: %v", err)
+		}
+
 		// Block dangerous env var overrides that could cause data fragmentation (bd-hevyw).
 		if err := checkBlockedEnvVars(); err != nil {
 			return HandleError("%v", err)
--- a/cmd/bd/metrics.go
+++ b/cmd/bd/metrics.go
@@ -248,8 +248,9 @@ func marshalIndentNoEscape(v any) ([]byte, error) {
 }
 
 // showQueuedEvents pretty-prints the real, locally-buffered event payloads from
-// ~/.beads/eventsData (the same files the flusher reads before sending), so the
-// user sees exactly what would leave their machine. Returns how many event
+// metrics.DataDir() (~/.beads/eventsData by default, or $BEADS_DIR/eventsData
+// when BEADS_DIR is set; the same files the flusher reads before sending), so
+// the user sees exactly what would leave their machine. Returns how many event
 // files were shown.
 func showQueuedEvents(out interface{ Write([]byte) (int, error) }) int {
 	dir, err := metrics.DataDir()
--- a/cmd/bd/metrics.go
+++ b/cmd/bd/metrics.go
@@ -249,9 +249,10 @@ func marshalIndentNoEscape(v any) ([]byte, error) {
 
 // showQueuedEvents pretty-prints the real, locally-buffered event payloads from
 // metrics.DataDir() (~/.beads/eventsData by default, or $BEADS_DIR/eventsData
-// when BEADS_DIR is set; the same files the flusher reads before sending), so
-// the user sees exactly what would leave their machine. Returns how many event
-// files were shown.
+// once the workspace is resolved; the same files the flusher reads before
+// sending), so the user sees exactly what would leave their machine. Safe to
+// call here because it runs from a command's RunE, after PersistentPreRunE has
+// already resolved the workspace. Returns how many event files were shown.
 func showQueuedEvents(out interface{ Write([]byte) (int, error) }) int {
 	dir, err := metrics.DataDir()
 	if err != nil {
--- a/cmd/bd/metrics.go
+++ b/cmd/bd/metrics.go
@@ -248,11 +248,9 @@ func marshalIndentNoEscape(v any) ([]byte, error) {
 }
 
 // showQueuedEvents pretty-prints the real, locally-buffered event payloads from
-// metrics.DataDir() (~/.beads/eventsData by default, or $BEADS_DIR/eventsData
-// once the workspace is resolved; the same files the flusher reads before
-// sending), so the user sees exactly what would leave their machine. Safe to
-// call here because it runs from a command's RunE, after PersistentPreRunE has
-// already resolved the workspace. Returns how many event files were shown.
+// metrics.DataDir() (~/.config/bd/eventsData, the same machine-scoped queue the
+// flusher reads before sending), so the user sees exactly what would leave
+// their machine. Returns how many event files were shown.
 func showQueuedEvents(out interface{ Write([]byte) (int, error) }) int {
 	dir, err := metrics.DataDir()
 	if err != nil {
--- a/internal/metrics/flusher.go
+++ b/internal/metrics/flusher.go
@@ -13,6 +13,11 @@ import (
 const (
 	EnvEndpoint = "BEADS_METRICS_ENDPOINT"
 
+	// EnvDataDir carries the parent's already-resolved event queue directory to
+	// the detached flusher, so the child never re-derives it from an inherited
+	// BEADS_DIR that may not match the workspace the parent selected.
+	EnvDataDir = "BD_EVENTS_DIR"
+
 	flushTimeout = 30 * time.Second
 )
 
@@ -21,10 +26,13 @@ func RunSendMetrics() int {
 		return 0
 	}
 
-	dir, err := DataDir()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "send-metrics: %v\n", err)
-		return 1
+	dir := os.Getenv(EnvDataDir)
+	if dir == "" {
+		var err error
+		if dir, err = DataDir(); err != nil {
+			fmt.Fprintf(os.Stderr, "send-metrics: %v\n", err)
+			return 1
+		}
 	}
 
 	// Bound the queue before flushing: TTL out stale batches and orphaned
--- a/internal/metrics/flusher.go
+++ b/internal/metrics/flusher.go
@@ -13,11 +13,6 @@ import (
 const (
 	EnvEndpoint = "BEADS_METRICS_ENDPOINT"
 
-	// EnvDataDir carries the parent's already-resolved event queue directory to
-	// the detached flusher, so the child never re-derives it from an inherited
-	// BEADS_DIR that may not match the workspace the parent selected.
-	EnvDataDir = "BD_EVENTS_DIR"
-
 	flushTimeout = 30 * time.Second
 )
 
@@ -26,13 +21,10 @@ func RunSendMetrics() int {
 		return 0
 	}
 
-	dir := os.Getenv(EnvDataDir)
-	if dir == "" {
-		var err error
-		if dir, err = DataDir(); err != nil {
-			fmt.Fprintf(os.Stderr, "send-metrics: %v\n", err)
-			return 1
-		}
+	dir, err := DataDir()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "send-metrics: %v\n", err)
+		return 1
 	}
 
 	// Bound the queue before flushing: TTL out stale batches and orphaned
--- a/internal/metrics/machineid.go
+++ b/internal/metrics/machineid.go
@@ -17,6 +17,13 @@ import (
 // it is still written 0600 like the rest of our per-user state.
 const machineIDCacheName = "machine-id"
 
+// machineIDDataDirName is the ~/.beads directory the machine-id cache lives
+// under. metrics.go's DataDir moved the telemetry queue to ~/.config/bd and
+// dropped its own dataDirName constant (GH#4807), but the machine-id cache
+// stays put in ~/.beads: relocating it would orphan every existing install's
+// cached distinct ID, which this fix does not need to do.
+const machineIDDataDirName = ".beads"
+
 // maxMachineIDLen bounds what the cache read will accept. ProtectedID today is
 // a 64-char hex HMAC; the bound is loose so an upstream format change does not
 // silently invalidate every cache, while still refusing to feed a corrupt or
@@ -28,7 +35,7 @@ func machineIDCachePath() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return filepath.Join(home, dataDirName, machineIDCacheName), nil
+	return filepath.Join(home, machineIDDataDirName, machineIDCacheName), nil
 }
 
 // validMachineID accepts a cached or freshly computed ID for (re)use: one
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -46,10 +46,11 @@ func Endpoint() string {
 // so this deliberately ignores BEADS_DIR and the resolved workspace, which is
 // what fixes GH#4807 on a fresh install where no workspace exists yet.
 func DataDir() (string, error) {
-	// UserConfigYamlPath returns a literal "~/..." display string when the home
-	// dir is unresolvable; joining that would make a directory named "~".
-	if _, err := os.UserHomeDir(); err != nil {
+	// os.UserHomeDir can return a non-absolute value (e.g. HOME=~) with no error, creating a literal "~" dir under cwd.
+	if home, err := os.UserHomeDir(); err != nil {
 		return "", fmt.Errorf("metrics: resolve home dir: %w", err)
+	} else if !filepath.IsAbs(home) {
+		return "", fmt.Errorf("metrics: resolve home dir: not absolute: %q", home)
 	}
 	return filepath.Join(filepath.Dir(config.UserConfigYamlPath()), "eventsData"), nil
 }
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -3,17 +3,16 @@ package metrics
 import (
 	"context"
 	"fmt"
-	"os"
 	"path/filepath"
-	"sync"
 	"time"
 
 	"github.com/dolthub/eventkit"
+
+	"github.com/steveyegge/beads/internal/config"
 )
 
 const (
-	AppName     = "beads"
-	dataDirName = ".beads"
+	AppName = "beads"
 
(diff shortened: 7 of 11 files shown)


ecuthiell:
While validating this PR on Windows, I found a separate test-safety defect: setting only `HOME` does not redirect Go's `os.UserHomeDir()` or `os.UserConfigDir()` there. The current metrics tests can therefore write telemetry queues and user config into the real Windows profile. I filed #4937 for the exact problem.

I prepared commit [`290eee721`](https://github.com/ecuthiell/beads/commit/290eee72194dbff4e23c6e2e4ebd88da8d7c4698) directly on this PR's current head, `ab80cffe4`. It preserves both contributed `DataDir` tests, adapts them to a shared cross-platform profile helper, and adds a package-level `TestMain` safety net so future tests cannot reach the developer's profile if they omit the helper.

Validation on Windows amd64 / Go 1.26.5:

```text
go test -count=1 -race -short ./internal/metrics  PASS
go vet ./internal/metrics                         PASS
```

An external profile sentinel and the real-profile hash/listing both remained unchanged, and the package-owned temporary profile was removed.

To preserve contributor priority, I have not opened a competing PR. The patch commit can be cherry-picked onto this branch; alternatively, if #4828 lands first, I will rebase that single commit onto `main` and open the one-issue #4937 follow-up.

vishnujayvel:
Thanks for validating this on real Windows hardware, @ecuthiell — that's exactly the kind of gap that's invisible on a Unix dev box. Appreciate you filing #4937 with the repro details instead of just noting it in passing.

I looked through `290eee721`. It's a correct and complete fix for what #4937 describes: it redirects `HOME`, `USERPROFILE`, `HOMEDRIVE`/`HOMEPATH`, `XDG_CONFIG_HOME`, `APPDATA`, and `LOCALAPPDATA`, which covers both `os.UserHomeDir()` (which falls back to `USERPROFILE`, then `HOMEDRIVE`+`HOMEPATH`, on Windows) and `os.UserConfigDir()` (which reads `%AppData%`). The added `TestMain` gives the whole `internal/metrics` package a safety net even for tests that don't call the helper directly, and `TestPackageUserProfileIsIsolated` self-checks that isolation actually took effect — good belt-and-suspenders design. It also keeps both of the `DataDir` tests from this PR intact, just routed through the shared helper.

To be clear on scope, since your issue calls this out itself: this fixes the isolation gap for `internal/metrics` specifically. The same `t.Setenv("HOME", ...)`-only pattern shows up in a couple dozen other test files across the tree (`cmd/bd/`, `internal/config`, the tracker packages, etc.), which lines up with what you flagged in #4937 as the broader, separate problem (#4638). So #4937 as filed is resolved by this commit; the wider sweep is correctly out of scope here.

One logistics note: `290eee721` is on your fork (`ecuthiell/beads`), not pushed onto this PR's branch — its parent is `ab80cffe4`, which is exactly this branch's current head, so it applies as a clean cherry-pick / fast-forward. Since you deliberately held off opening a competing PR, I've cherry-picked it onto `fix/metrics-beads-dir-4807` and pushed, with you preserved as the commit author. I haven't run the Windows suite myself, so I'm taking your validation output (`go test -race -short ./internal/metrics`, `go vet`) at face value rather than re-asserting anything I haven't checked — flag if you'd rather I wait for CI to confirm before you sign off.

Once that's pushed, this should be ready for another look.

ecuthiell:
@vishnujayvel Thanks! I'm happy with this - don't delay on my account!

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the ordering at the PR head). Verdict: **needs changes** — the intent (honor BEADS_DIR for telemetry) is right; startup ordering defeats it.

**Blocker — telemetry binds its queue before workspace selection.** Verified: `metrics.Init` runs (~main.go:746) before `applyChangeDirSelection` (~:768), before `.beads/.env` loading, and long before final context binding — and the pinned eventkit emitter calls `os.MkdirAll` immediately. So `bd -C <workspace>` still creates and writes `$HOME/.beads/eventsData`. Worse, when `.beads/.env` or explicit-target flows later change `BEADS_DIR`, the spawned flusher inherits the *changed* environment and recomputes `DataDir()` — parent writes the home queue, child flushes the workspace queue, events stranded. The new test calls `DataDir()` directly with `BEADS_DIR` pre-set, so it can't see the lifecycle. Fix shape: defer emitter construction until after workspace selection (or re-init on selection), and pass the resolved dir to the flusher explicitly rather than via inherited env.

**Should-fix — stale docs:** `AGENT_INSTRUCTIONS.md` (and the hint in `cmd/bd/metrics.go`) still say events always live under `~/.beads/eventsData`.

Drift is clean (`DO_NOT_TRACK` support is preserved by the merge), but current main has the same init ordering — a rebase alone won't fix the blocker.

vishnujayvel:
Thanks for the review — the ordering issue is real and worth getting right before this lands.

Confirmed: `metrics.Init` runs in `PersistentPreRunE` (cmd/bd/main.go, around line 745) before `applyChangeDirSelection()` (around line 768), so `DataDir()` is evaluated against whatever `BEADS_DIR` happens to be in the ambient environment at process start, not the workspace `bd` actually resolves via `-C`/`.beads/.env`/explicit-target flows. That confirms the blocker as described, including the flusher-inherits-changed-env split-write scenario.

I don't want to guess at the right shape for this under review pressure — deferring emitter construction vs. re-initializing on selection vs. passing the resolved dir into the flusher explicitly are meaningfully different designs with different blast radius on the startup path (telemetry.Init, span setup, and the `BD_DISABLE_METRICS`/user-config bootstrap all sit in the same `PersistentPreRunE` block). I'd rather a maintainer weigh in on which shape fits the intended lifecycle before I touch that ordering.

What I did do in this pass, since it was unambiguous:
- Updated the stale `~/.beads/eventsData`-only docs in `AGENT_INSTRUCTIONS.md` and the `showQueuedEvents` comment in `cmd/bd/metrics.go` to reflect the `BEADS_DIR` override, so at least the documentation matches what `DataDir()` actually does today (independent of the ordering fix).

I also checked the two failing required checks:
- **PR Core (wrapper timing)**: the only test failure is `TestInitBackendFlag/sqlite_is_no_longer_supported` in `cmd/bd/init_test.go`, asserting `bd init --backend sqlite` writes no workspace state. That subtest doesn't exist on this branch's own tree — it only exists on current `main` post `6fd9dbda7` ("remove the SQLite backend and dialect"), and CI runs against the PR-into-main merge commit. This PR's diff doesn't touch backend selection or `cmd/bd/init.go` at all, so this looks like a pre-existing regression on `main` surfaced by the merge, unrelated to this change — worth its own issue rather than blocking here.
- **CI Gate / Required**: fails solely because it aggregates `PR_CORE_WRAPPER`, so it's downstream of the same failure above.

Given the ordering fix is a real design decision, I'll leave that for maintainer input rather than push a guess through review. Happy to implement whichever shape you'd prefer once there's agreement on it.

maphew (commented):
Two things, and the first one matters for the red check.

**The failing `PR Core` check is caused by this PR, not pre-existing on main.** We re-ran the failing subtest on clean main: it passes. The mechanism: the init test env sets `BEADS_DIR` and strips `BD_*` vars (so metrics are enabled), then asserts the workspace dir does not exist after a rejected init — but with this PR, `DataDir()` resolves to `$BEADS_DIR/eventsData`, and `metrics.Init` runs before `applyChangeDirSelection`, where the file emitter eagerly `MkdirAll`s. The test is not stale; it is a canary for exactly the harm the earlier review described — telemetry materializing directories at a path resolved before the workspace decision is made. A useful habit that would have caught this: before declaring a red check pre-existing, reproduce it on the merge base — if it passes there, it's ours.

**On the design ruling you asked for** — that's a fair ask, and here is our recommendation: split `metrics.Init` so the collector registers early with a null emitter, and attach the file emitter only after `applyChangeDirSelection`, passing the resolved data dir explicitly rather than via inherited env. That one move fixes the eager mkdir, the failing check, and the parent/child split-write problem together. If you'd like to take that shape, we're glad to review it; if you'd rather we absorb it maintainer-side with your attribution preserved (the ecuthiell cherry-pick is already handled nicely, and the metrics test-env isolation work here is independently valuable), say the word.

One small note: the doc additions describe the `BEADS_DIR` behavior as settled, but while the ordering bug stands it's only intermittently true — worth holding those lines until the lifecycle fix lands.

vishnujayvel:
Thanks — both points landed, and the baseline habit especially. You were right that the red check was ours, and I'd been about to call it pre-existing. Reproducing on the merge base before making that claim is now the rule I work by.

I took the shape you recommended. `Init` registers the collector against a switchable emitter wired to `NullEmitter{}`, and `AttachFileEmitter(dataDir)` is called from `PersistentPreRunE` after `applyChangeDirSelection`, with the resolved dir passed in rather than read from the environment.

**That shape alone did not fix the check, though**, and the reason seems worth writing down. I built it, merged current `main`, and ran `TestInitBackendFlag/sqlite_is_no_longer_supported` — still red. Deferring the attach moves the `MkdirAll` later, but a rejected `bd init --backend sqlite` still reaches `PersistentPreRunE`, still emits its command event, and `eventkit.NewFileEmitter` still creates the queue under the `BEADS_DIR` the rejected command was aiming at. So `.beads` existed and the assertion fired. On `main` the check passes only because `DataDir()` there ignores `BEADS_DIR` entirely — the queue lands in `$HOME`, safely outside the directory under assertion. In other words the canary isn't really guarding the mkdir timing; it's guarding a rule this PR's premise collides with:

**telemetry may write into a workspace, but must never be the thing that brings one into being.**

Two changes on top of your shape make that true:

- The file emitter is now built lazily, on the first event that actually needs writing. `NewFileEmitter` calls `MkdirAll`, so constructing it eagerly materialized a queue directory even for commands that emit nothing.
- A `BEADS_DIR` that does not exist yet is ignored in favor of the home queue. A command that hasn't created its workspace doesn't get one created on its behalf by the metrics path.

That last one is the load-bearing bit, and it's a deliberate narrowing of what this PR originally claimed: `BEADS_DIR` is honored for telemetry *once the workspace exists*, not unconditionally. If you'd rather the queue always stay in `$HOME` and drop the workspace-local behavior altogether, that's a smaller and even safer change — say so and I'll cut it back.

I also took the flusher half of your first review. `send-metrics` was re-deriving `DataDir()` from its inherited environment; it now receives the parent's already-resolved dir as `BD_EVENTS_DIR`, mirroring how `flusherChildEnv` already pins the resolved endpoint instead of trusting an inherited one. Parent and child can no longer disagree about which queue they're using.

Docs: both stale spots are updated, and they now state the workspace-exists condition rather than describing `BEADS_DIR` as unconditional — your note about not over-claiming while the lifecycle was still broken was fair.

Verification, run against this branch merged with current `main` rather than the branch alone, since CI tests the merge commit and the canary subtest doesn't exist on our merge base:

- `TestInitBackendFlag/sqlite_is_no_longer_supported` — passes (fails without the two changes above)
- full `cmd/bd` package — passes
- `internal/metrics` — passes, including under `-race` for the new switchable emitter
- new regressions: `Init` alone creates no queue dir; a non-existent `BEADS_DIR` falls back to home and is not created; the flusher child env pins the resolved dir

Local runs skip the Dolt-server-gated subtests (no Docker here), so CI is still the real check on those.

On your offer to absorb this maintainer-side: still happy either way. The shape is small now and I've done the work, so merging it here is easy — but if you'd rather carry it yourself, take it, no attribution needed on my end.

maphew (commented):
The shape you landed is the right one, and the write-up that came with it is the most useful thing in this thread: "telemetry may write into a workspace, but must never be the thing that brings one into being" is a rule worth keeping in the codebase regardless of where this PR ends up. Reproducing the canary on the merge base before calling it pre-existing, and then reporting that your first fix *didn't* make it pass, is exactly the debugging discipline that makes a review cheap. Thank you — and thanks for carrying @ecuthiell's isolation commit with authorship intact rather than re-typing it.

Re-verified at head `a662a83` merged with current `main`: `internal/metrics` passes under `-race`, `TestInitBackendFlag/sqlite_is_no_longer_supported` passes, all required checks are green.

Two things left, both about scope rather than breakage.

**1. The fresh-setup case in #4807 is still not fixed.** The existence gate means the first command in a not-yet-created `BEADS_DIR` falls back to `$HOME/.beads` — which is the phantom directory the issue is about. Reproduced against your build in an isolated HOME:

```
$ BEADS_DIR=$H/.local/share/beads bd version
$ find $H/.beads
$H/.beads/eventsData
$H/.beads/eventsData/qVwnVh6Pg0ciwbuYkl6y0g.evtq
```

That is step one of the issue's own repro (`rm -rf ~/.beads ~/.local/share/beads; export BEADS_DIR=...; bd init ...`). Post-init runs do go to the right place — I confirmed `$BEADS_DIR/eventsData` gets the queue once the directory exists — so this is a real improvement, just not a complete one. Either narrow the body to `Refs #4807` and say the fresh case is out of scope, or take the alternative below.

**2. The `.beads/.env` route is documented but not honored.** `AttachFileEmitter` is called at `cmd/bd/main.go:775`, and `loadSelectionEnvironment()` — which is what reads `BEADS_DIR` out of `.beads/.env` — runs a few lines below it. Reproduced: with `proj/.beads/.env` naming a workspace and no ambient `BEADS_DIR`, the event landed in `$HOME/.beads/eventsData`, not the workspace. But `AGENT_INSTRUCTIONS.md:500` and the `showQueuedEvents` comment both list `.beads/.env` as a supported source. Smallest fix: move the attach block to just after `loadSelectionEnvironment()`; otherwise drop the `.beads/.env` clause from both doc spots.

Smaller items, take or leave:

- `internal/metrics/metrics.go:100` — `os.Stat` accepts a regular file, so a `BEADS_DIR` naming a file yields `<file>/eventsData` and the `MkdirAll` failure surfaces only inside `Send`, dropping the event. `info.IsDir()` closes it.
- Same function — the raw `os.Getenv("BEADS_DIR")` diverges from the canonical `beads.FindBeadsDir()` (`internal/beads/beads.go:689`), which canonicalizes the path, follows `.beads/redirect`, and validates project files. With a redirect in play telemetry writes to the pre-redirect directory. There's no import cycle in that direction if you'd rather resolve in `cmd/bd` and pass the result in.
- `cmd/bd/metrics.go:241` — `showQueuedEvents` re-derives `DataDir()` at RunE time, which can differ from the dir the emitter actually armed (fallback case, or `BEADS_DIR` created in the meantime by `bd init`). You already have `switchableEmitter.attachedDir()`; exporting it and preferring it here makes the display always match the queue in use.
- Nothing pins the *positive* ordering — that `bd -C <ws>` queues under `<ws>/eventsData`. The canary only catches the negative, so a future reorder of `PersistentPreRunE` would regress this quietly.

**The simpler road, if you want it.** You floated it yourself, and the issue author floated it first: put the queue somewhere that is not a `.beads` directory at all — e.g. beside `config.UserConfigYamlPath()` under `~/.config/bd/`, which bd already creates in `EnsureUserConfigDefaults`. Telemetry is machine-scoped (`eventkit.MachineID`), so a per-workspace queue buys nothing that a single global one doesn't, and that version fixes #4807 in every case including the fresh one while deleting the switchable emitter, the `BD_EVENTS_DIR` pinning, the lazy mkdir and the Stat gate. It's roughly 90 fewer lines than what's here, and the ordering question disappears entirely. The only cost is a one-time decision about existing queued files under `~/.beads/eventsData`.

That is a design call, not a demand — your current shape is defensible and green, and if you'd rather land it as-is with the doc claim corrected and the scope narrowed to `Refs #4807`, say so and we'll take that. Either way, tell us whether you want to make these edits or would rather we apply them maintainer-side with your attribution preserved; the ecuthiell absorption you handled shows you'd rather drive, so we won't push commits onto your branch uninvited.

One bookkeeping note: this branch also resolves #4937 via `3991752b5`, but the body only says `Fixes #4807` — adding `Fixes #4937` will close that issue on merge.


vishnujayvel:
@maphew Taking the simpler road — you're right that it's better, and the fresh-install gap you found is the decisive reason: no amount of ordering surgery on the workspace-aware version fixes a case where there is no workspace yet.

Landed at `bb2798307`, plus two follow-ups (`1fc0a455f`, `81ca32517`) — both described below. Net `-96` across 8 files.

- `DataDir()` is now a single machine-scoped queue beside the user-global config: `~/.config/bd/eventsData`, resolved via `config.UserConfigYamlPath()` so it stays in lockstep with wherever that already lives on a given platform. It no longer reads `BEADS_DIR` at all.
- `Init` goes back to building the file emitter eagerly and directly — no more switchable wrapper. That's safe now for the reason you gave: this directory is never a workspace, so there's nothing for a rejected command to falsely "create." Startup ordering relative to `applyChangeDirSelection` stops mattering entirely, which is what let me delete the switchable emitter, `AttachFileEmitter` and its call site, the `BD_EVENTS_DIR` flusher-child pinning, the lazy-mkdir Send-time construction, and the `BEADS_DIR`-existence Stat gate. Net diff is -115 lines, in the neighborhood of what you estimated.
- `.beads/.env`'s `BEADS_DIR` clause is gone from both doc spots (AGENT_INSTRUCTIONS.md and the `showQueuedEvents` comment) rather than fixed, since it's no longer relevant to telemetry — the smaller-diff option you offered.
- Your smaller items (`os.Stat` vs `info.IsDir()`, the `FindBeadsDir()` canonicalization gap, `showQueuedEvents` re-deriving instead of reading `attachedDir()`) are all moot under this design — there's no `BEADS_DIR` read left in the metrics path for any of them to apply to.

One thing the move surfaced that I want to flag explicitly, since it's a regression I introduced and then caught rather than something you asked for. `config.UserConfigYamlPath()` returns the literal display string `"~/.config/bd/config.yaml"` — tilde and all — when `os.UserHomeDir()` fails (`internal/config/yaml_config.go:541`, comment says "fallback display string"). Joining that unchecked means `DataDir()` hands back the relative path `~/.config/bd/eventsData`, and `NewFileEmitter`'s `MkdirAll` then creates a directory *literally named* `~` under the process's cwd. `os.UserHomeDir()` fails whenever `$HOME` is unset — containers, systemd units, cron — which is exactly where `bd` runs unattended. The code this PR replaced returned an explicit error on that path; `81ca32517` restores it, with a test that fails with `DataDir() = "~/.config/bd/eventsData"` if the guard is removed (I checked that by removing it, not by assuming).

On your one open question — existing queued files under `~/.beads/eventsData` from the old location — I did not write a migration. Anything already queued there is orphaned rather than moved forward; a future run's flusher only reads the new location. That felt like the right call for keeping the diff minimal and matching the "simpler road" spirit, but it's a real decision, not something I want to have made unilaterally — happy to add a one-time migration (or just leave it, the volume is presumably tiny) if you'd rather.

Test coverage: `internal/metrics` rewritten so `DataDir()`'s test asserts `BEADS_DIR` — unset, pointing at an existing dir, and pointing at a not-yet-created dir — never changes the result, which is the direct regression test for the fresh-install case you reproduced. `TestInitBackendFlag/sqlite_is_no_longer_supported` passes when I merge this branch with current `main` (verified locally, not just asserted) — the canary that motivated the whole redesign is now permanently un-triggerable rather than newly satisfied, since the metrics path never touches `.beads` at all anymore.

What I verified locally, merging this branch with current `origin/main` (no Docker here, so Dolt-server-backed subtests skip both directions):
- `go vet ./internal/metrics/... ./cmd/bd/...` — clean
- `go test -race ./internal/metrics/...` — full package passes
- `go test -run TestInitBackendFlag ./cmd/bd/` — passes, including the canary subtest
- `go test -run 'Metrics|ShowQueuedEvents' ./cmd/bd/` — every metrics-touching test in `cmd/bd` passes, including the double-emit regressions and the env/config precedence matrix

That last run caught something the internal/metrics-only pass missed: `cmd/bd/init_metrics_test.go` had its own hardcoded `$HOME/.beads/eventsData` assumptions (separate from anything in `internal/metrics`), which failed loudly once I actually ran the package instead of stopping at `go vet`. Fixed in the follow-up commit `1fc0a455f`. Flagging this myself since I'd rather you hear it from me than find it — I did initially report only the `internal/metrics` and canary results and had to come back and add this.

Also exported the two-commit patch series and confirmed it applies cleanly to the PR's own merge-base.

Still outstanding: your and CI's own re-verification, since I don't have Windows or a Dolt server to check the two things covered by @ecuthiell's isolation commit and the `dolt_backend_succeeds` subtest myself. I did not attempt the full, unscoped `cmd/bd` package run (it needs ~9 minutes and I only have a bounded window here) — the scoped runs above cover everything this change touches.

Thanks again for pushing on this past "defensible and green" — the fresh-install repro you ran made it obvious the workspace-aware version was solving an ordering problem instead of the actual bug.

vishnujayvel:
Two things from a follow-up pass on this branch. The first is a correction to something I told
you in this thread yesterday.

**The home-dir guard I added in `81ca32517` was incomplete.** I described it as restoring the
explicit error on an unresolvable home, and it does — but it only checks `os.UserHomeDir()`'s
*error*, not its *value*. `os.UserHomeDir()` on Unix just reads `$HOME`, so with `HOME=~` it
returns `("~", nil)`: no error, guard passes, and `DataDir()` hands back the relative path
`~/.config/bd/eventsData` again. `MkdirAll` then creates a directory literally named `~` under
the cwd — the exact failure I'd claimed the guard prevented. I confirmed it with a throwaway
probe rather than reasoning about it: `t.Setenv("HOME", "~")` gives
`DataDir() = "~/.config/bd/eventsData"` and `filepath.IsAbs(...) = false`.

`f0315ad0f` requires the resolved home to be absolute, with a test covering `HOME=~`. I checked the
test is load-bearing by neutralising the guard and confirming it fails with the bad path, not
just that it passes with the fix. To be clear about how reachable this is: no shell or OS sets
`HOME=~` — a login shell always resolves it to an absolute path. It takes a non-shell
env-loading context that doesn't tilde-expand: a Docker `ENV HOME=~`, an `EnvironmentFile` line,
a mis-set CI variable. Narrow, but it's the same class of unattended environment the original
GH#4807 report came from, and the guard was already claiming to cover it.

**Second, a design question I don't want to answer unilaterally, since it's about the shape you
asked for.** Tying the queue to `config.UserConfigYamlPath()` also inherits that function's
existence checks. It isn't a pure function of the environment — `internal/config/yaml_config.go:526-541`
stats `~/.config/bd/config.yaml` first, then `os.UserConfigDir()/bd/config.yaml`, and returns
whichever exists:

```go
if _, err := os.Stat(xdgPath); err == nil {   // :528
    return xdgPath
}
if configDir, err := os.UserConfigDir(); err == nil {
    osPath := filepath.Join(configDir, "bd", "config.yaml")
    if _, err := os.Stat(osPath); err == nil { // :535
        return osPath
    }
}
return xdgPath
```

So on macOS the answer can change between two runs with identical env. Concretely: a user with
no config file anywhere queues events to `~/.config/bd/eventsData`; a
`~/Library/Application Support/bd/config.yaml` later appears — a leftover from a build that only
wrote the `os.UserConfigDir()` location, or a dotfiles restore — and the next run resolves to the
Library path instead. Events already queued at the first location are then orphaned, since the
flusher only reads wherever `DataDir()` currently points.

I have deliberately **not** changed this. The fix would be to pin the queue to one stable
machine-scoped directory independent of where the config file happens to live, or to have the
flusher sweep both candidates — and either one moves the design away from "beside the
user-global config", which is the property you asked for. It's also not a regression this PR
introduces; it's a consequence of the location itself, and it would apply equally to the
`BEADS_DIR`-based version. Your call whether it's worth a follow-up issue or is simply
acceptable — the volume of orphaned events would be tiny, and it only bites macOS users whose
config location moves.

maphew (commented):
Reviewed at `f0315ad0`.

The simplified machine-scoped design fixes the telemetry-queue problem. One small path-stability issue remains: `DataDir` follows `UserConfigYamlPath`, whose result depends on which config file currently exists, so macOS can silently switch queues and strand pending files. Pin telemetry to one pure, stable machine path instead.

This supersedes the more complex alternatives discussed around #4807 rather than overlapping another open implementation PR. After the path fix, rerun the unrelated failing proxied shard and clear the stale changes-requested state.

vishnujayvel:
Pinned `DataDir()` to a pure `~/.config/bd/eventsData` path, no existence probing.

Confirmed the issue: `DataDir()` called `filepath.Dir(config.UserConfigYamlPath())`, and `UserConfigYamlPath()` checks `~/.config/bd/config.yaml` first, then falls back to checking `os.UserConfigDir()/bd/config.yaml` (`~/Library/Application Support/bd/config.yaml` on macOS) via `os.Stat` on each call — so the resolved directory could flip between runs as either file appeared or disappeared, exactly as described.

Fix pins `DataDir()` to `~/.config/bd/eventsData`, computed straight from `os.UserHomeDir()` with no probing — same shape as `configfile.DefaultCredentialsPath()`, the existing convention for other machine-scoped state in this repo. `UserConfigYamlPath()` itself is untouched; it's still used for the actual config.yaml reads/writes where checking both candidates is intentional.

Added `TestDataDirStableRegardlessOfConfigFileExistence`: resolves `DataDir()`, creates a `config.yaml` at the OS-native candidate, resolves again, removes it, creates one at the `~/.config/bd` candidate instead, resolves again — asserts the result never moves. Ran on macOS, where the two candidates are genuinely distinct paths, so the test exercises the real divergence rather than coinciding trivially.

No migration needed: the `UserConfigYamlPath()`-linked `DataDir()` only ever existed on this open PR branch (introduced in `bb2798307`) — it never merged to `main`, so no released build ever queued telemetry there. Nothing is stranded for real users.

This also closes out the config-path instability we'd flagged as an open design question on 2026-07-30.

The proxied-shard rerun and clearing the stale changes-requested state both need maintainer access, so I've left those for you — let me know if anything else on the branch needs changing first.

ecuthiell:
Current follow-up status from the #5602 review:

- #4828 still carries the `internal/metrics` profile-isolation fix for #4937 as commit `3991752b54`; I will not open a competing implementation while that work remains incorporated here.
- The PR is currently reported as conflicting with `main`, so its final implementation and checks need a current-base rebase/re-review before it can land.
- The body still closes only #4807. Please add `Fixes #4937` (or equivalent close bookkeeping) when rebasing so the incorporated isolation issue does not remain open after merge.

No new design request is hidden in this note; it is ownership and landing bookkeeping for the already-reviewed work.

vishnujayvel:
Thanks for the status note, and for holding off on a competing implementation — appreciated.

Rebased onto `main` @ `634cbbc4b`. All 14 commits replayed, none dropped. Five conflicts, all in
`internal/metrics/metrics.go` except the first, resolved keep-both or by re-applying our change on
top of the new upstream structure:

- `userconfig_test.go` — upstream added `USERPROFILE`/`APPDATA` to the inline test setup; our
  `isolateUserProfile(t)` helper from the #4937 isolation work is a superset of it, so the helper
  now covers that case too.
- `Init()` ×4 — upstream made `distinctID`/`cachedMachineID` lazy and renamed the emitter API twice
  (`set(e)` → `set(e, dir)` → `useNull()`/`useDir()`). Our deferred-emitter change re-applied on
  each new signature; the lazy distinct-ID block kept intact.

I've also added `Fixes #4937` to the body as you asked. While in there I corrected the Summary —
it still described the original `$BEADS_DIR/eventsData` passthrough, not the `~/.config/bd` move
this branch actually settled on after your design ruling. Happy to revert that part to a pure
one-line bookkeeping edit if you'd rather keep the diff to exactly what you asked for.

## One thing worth flagging — a silent cross-file break

This is not a conflict git could show you, so I want it on the record rather than buried in the
diff. `internal/metrics/machineid.go` (added upstream since this branch forked) resolves its cache
path with:

```go
filepath.Join(home, dataDirName, machineIDCacheName)
```

`dataDirName` is the `".beads"` constant in `metrics.go` that this PR's commit
"move telemetry queue to `~/.config/bd`, drop `BEADS_DIR` entirely" deletes. After the rebase the
tree therefore did not compile — no conflict marker anywhere, just an `undefined: dataDirName` at
build time.

There are two honest resolutions and I deliberately took the conservative one:

- **What I did:** kept the machine-id cache exactly where you put it, `~/.beads/machine-id`, by
  declaring the directory constant local to `machineid.go`. Upstream's `machineid_test.go` passes
  completely unmodified. No behavior change.
- **What I did not do:** co-locate the cache at `~/.config/bd/machine-id` alongside the relocated
  queue. That is arguably the tidier end state — it is the same category of per-user state, and the
  XDG argument that moved the queue applies to it equally — but it would orphan every existing
  install's cached distinct ID, so each existing machine would recompute a fresh one and read as
  brand-new in telemetry. That is a product decision, not a rebase decision, and you said plainly
  there was no design request in your note.

If you'd like the co-location, say so and I'll do it as a separate commit here or a follow-up PR —
your call which.

## Test plan

Rebased onto `origin/main` @ `634cbbc4b`. ICU flags per the repo's CGO requirements.

- [x] `go build ./...` — clean
- [x] `go vet ./internal/metrics/... ./cmd/bd` — clean
- [x] `go test ./internal/metrics/...` — pass
- [x] `go test -run 'Metric|Telemetry' ./cmd/bd/` — pass
- [x] `git diff origin/main -- internal/metrics/machineid_test.go` — empty, i.e. the upstream test
      is untouched, which is the check that the resolution above really is behavior-preserving

**Limitation stated plainly:** Docker isn't available on this machine, so the Dolt-backed suites
report `WARN: Docker not available, skipping Dolt tests`. Nothing in this branch touches that path,
but the rebase is verified against the Go suites above rather than a full CI run.

ecuthiell:
Exact-head Windows follow-up at `d689d4c2d`:

The replayed #4937 package-isolation commit (`a902fdf68`) is intact and correct. An uncached native-Windows `go test -count=1 -tags=gms_pure_go -race ./internal/metrics/...` passes under an external profile sentinel; its marker stayed byte-identical and no `.config`, `.beads`, app-data `bd`, or inherited `BEADS_DIR` path was created.

I did find one adjacent `cmd/bd` harness gap while reproducing this PR body's focused test. With CGO enabled on Windows, `go test -count=1 -tags=gms_pure_go -run 'Metric|Telemetry' ./cmd/bd/` fails at this head because `metricsTestEnv(home)` changes only `HOME` through `bdEnv`, while retaining the package-level `USERPROFILE`. `os.UserHomeDir()` follows `USERPROFILE` on Windows, so subprocesses queue events under the package profile while the assertions inspect the per-test profile. The required Windows lanes do not exercise this exact CGO-enabled slice.

A test-only fix is available as [`facd1890f`](https://github.com/ecuthiell/beads/commit/facd1890fc33dbd2ea120c5b921d6dca21490036), directly atop the current PR head. It normalizes the complete per-test profile environment without changing production behavior. After that commit, these pass on native Windows:

- `go test -count=1 -tags=gms_pure_go -race ./internal/metrics/...`
- `go test -count=1 -tags=gms_pure_go -run 'Metric|Telemetry' ./cmd/bd/`
- `go test -count=1 -tags=gms_pure_go -run '^TestInitBootstrapsUserConfigWhenMissing$' ./cmd/bd/`

Please absorb that commit before landing.

Separately, the final implementation no longer matches two pieces of PR metadata: the title says it honors `BEADS_DIR`, although the queue now deliberately ignores it, and the body says the emitter is deferred, although `Init()` now constructs it eagerly and `TestInitEnabledCreatesDataDirImmediately` enforces that behavior. A current title such as `fix(metrics): move telemetry queue out of workspace state (#4807)` plus removal of the deferred-emitter sentence would make the metadata match the final diff. `Fixes #4937` is correct and should remain.

_Codex-unknown-model-unknown-reasoning on behalf of Ewen Cuthiell_

vishnujayvel:
Absorbed — thanks, and thanks for going back to real Windows hardware on the exact head rather
than the branch generally. Cherry-picked `facd1890f` with `-x` so the provenance and your
authorship stay on the commit.

The mechanism reproduces on reading. `TestMain` points `HOME`/`USERPROFILE`/`APPDATA`/
`XDG_CONFIG_HOME` at one suite-wide temp root, and `metricsTestEnv` then overrides only `HOME`
for the subprocess — so `USERPROFILE` still names the suite root, not the per-test fixture. On
Windows `os.UserHomeDir()` follows `USERPROFILE`, so the subprocess queues events under the suite
root while the assertions read the per-test `home`. My local runs cover the Unix suites only — the
CGO-enabled `cmd/bd` slice is untested on my side, so that part rests on your diff.

## On the metadata — both corrections land

Retitled to your wording: `fix(metrics): move telemetry queue out of workspace state (#4807)`. The
old title described the abandoned first revision.

The deferred-emitter sentence was wrong, not just stale — it contradicts
`TestInitEnabledCreatesDataDirImmediately` (`internal/metrics/metrics_test.go:246`), which documents
the opposite: the queue path is "a fixed, machine-scoped location … so it is safe for Init to
construct the file emitter … unconditionally and immediately." Replaced with a sentence that keeps
the point — telemetry never creates a workspace directory — via the mechanism that actually ships.
`Fixes #4937` stays, as you said.

vishnujayvel:
Rebased onto `upstream/main` @ `71c4cd08b`. New head: `054125efd`.

- Kept ecuthiell's absorbed Windows validation commit (`054125efd`, author Ewen Cuthiell). Harness gap fix is present after rebase: `metricsTestEnv` still overrides `HOME`/`USERPROFILE`/`HOMEDRIVE`/`HOMEPATH`/`XDG_CONFIG_HOME`/`APPDATA`/`LOCALAPPDATA`.
- `git range-diff` vs the pre-rebase series is not all `=`: 8/16 commits identical (including the Windows isolation commit), 8 replayed with keep-both against main's GH#5712/GH#5871/GH#5900 spawn/prune work (main's tests and `startDetached` retained; DataDir still `~/.config/bd/eventsData`).
- Local gates: `go test -tags gms_pure_go ./internal/metrics/...` 67 passed; `go vet -tags gms_pure_go ./internal/metrics/... ./cmd/bd` clean; `scripts/ci/pr-lint.sh` with `BD_LINT_NEW_FROM_MERGE_BASE=upstream/main` native golangci-lint 0 issues (bare tree lint is red on files this PR does not touch, also on main). `cmd/bd` subprocess tests timed out here on embedded-dolt init (7–8 min per case) — environmental, not an assertion failure.

Push: `--force-with-lease` of `054125efd` replacing `c70d55d17`.

PR #4829, opened 2026-07-15
fix(storage): total ORDER BY on dependency bulk-loaders (#4749)

### LINKED ISSUE #4749: Dependency bulk-load has no ORDER BY tiebreaker — `dependencies` arrays come back in arbitrary, plan-dependent order (unstable `bd ready --json`, phantom export churn)
**Version:** bd 1.1.0 (dev) · Dolt 2.1.9, server mode · Darwin arm64
**Affected at current tip (`303e263`):** `internal/storage/issueops/dependency_queries.go:34` and `:113`

## Summary

Both dependency bulk-loaders order by `issue_id` alone:

```go
// dependency_queries.go:113 — getDependencyRecordsIntoFromTable (the export / ready path)
SELECT issue_id, %s AS depends_on_id, type, created_at, created_by, metadata, thread_id
FROM %s WHERE issue_id IN (%s) ORDER BY issue_id

// dependency_queries.go:34 — getAllDependencyRecordsIntoFromTable
SELECT issue_id, %s AS depends_on_id, ... FROM %s ORDER BY issue_id
```

`issue_id` is the *grouping* key, not a total order. Rows are then appended per issue into `result[dep.IssueID]`, so the order of the dependencies **within** a single issue is whatever the engine happens to return. It is not random per run — so a repeat-export test passes — but it is **arbitrary and plan-dependent**, and it shifts when the plan or the physical row order shifts.

The sibling bulk-loaders in the same package both get this right, which is what makes it look like an oversight rather than a decision:

| loader | `ORDER BY` | total? |
|---|---|---|
| `labels.go:110` | `issue_id, label` | ✅ |
| `bulk_ops.go:147` (comments) | `issue_id, created_at ASC, id ASC` | ✅ |
| `dependency_queries.go:34`, `:113` | `issue_id` | ❌ |

## The two shipped queries disagree with each other

This part needs no history and no speculation about storage internals. The two queries above select the same rows for a given issue, both say `ORDER BY issue_id` — and they return them in **different orders**. Measured against an 839-issue DB (201 issues carry ≥2 dependencies):

- the two queries disagree on **115 of the 201** issues
- e.g. `bd-a1qs.3.10` — full-scan returns `[bd-a1qs.3.9, bd-a1qs.3]`, IN-list returns `[bd-a1qs.3, bd-a1qs.3.9]`; set-equal, order-different

Within the *single* IN-list query the engine also mixes access paths: 86 of the 201 issues come back in primary-key order (the PK is a content-derived UUIDv5, so effectively hash order) and 115 in secondary-index order. That accounts exactly for the observed distribution — of the 86 PK-order issues, 68 happen to coincide with `depends_on_id` order and 18 do not, so precisely 18 arrays are non-canonical. Predicted set == observed set.

## Consequences

**1. `bd export` churn.** The export serializes this order verbatim (`cmd/bd/export.go:186`); I confirmed the exported arrays reproduce the IN-list query's order for 201/201 issues. When the order shifts, git sees a full-line rewrite for every affected issue with byte-identical *content*. One commit in our repo rewrote **18 of 201** dependency arrays — all set-equal, all noise — alongside just 6 real changes.

**2. It is not confined to export.** The defect is in the storage layer, so the same unpinned order reaches `bd ready` (`ready.go:475` assigns the slice straight to `issue.Dependencies`, which `--json` then emits), `bd list`, `bd dep`, and `bd migrate`. Worth separating from #4231 / #4436, which were closed and deferred on the grounds that JSONL export is being deprecated as source of truth in 2.0 — **this query outlives that deprecation.**

**3. (Latent) `bd ready`'s `childToParent` is last-write-wins.** `ready.go:412` does `childToParent[issueID] = dep.DependsOnID` inside the loop, so an issue with more than one `parent-child` edge would get an arbitrary parent, picked by this same unpinned order. The unique keys are `(issue_id, target)`, so the schema does not forbid multi-parent. No such rows exist in my database, so I have **not** observed this — flagging it as latent rather than claiming it.

## Why the existing test doesn't catch it

`TestExportByteStabilityAllRecordTypes` (`cmd/bd/export_test.go:645`, GH#4086) is the guard here, but it cannot see this bug, for two independent reasons:

1. **It fans the dependencies the wrong way.** `for _, issue := range issues[1:] { AddDependency(issue → issues[0]) }` gives every issue exactly **one** outgoing dependency. `GetDependencyRecordsForIssues` groups by the dependency's *source*, so every array the test builds has length 1 — the intra-issue tie order, the only thing `ORDER BY issue_id` fails to determine, is never exercised.
2. **Repeat-export identity is the wrong invariant.** The order is stable for a fixed storage state, so N consecutive exports over unchanged data are byte-identical *even with the bug present*. The property that actually matters for a VCS-tracked artifact is that the serialization is a pure function of the logical content — canonical — not merely repeatable.

## Suggested fix

Give both queries a tiebreaker, mirroring the labels loader:

```sql
ORDER BY issue_id, depends_on_id, type
```

`(issue_id, target)` is unique (`uk_dep_issue_target`), so this is a total order. Verified against a live Dolt 2.1.9 server: ordering by the `depends_on_id` alias is accepted and returns canonical order.

A test that would actually catch it: fan the dependencies **out** (give one issue N deps) and assert the resulting array is sorted — rather than asserting only that two exports agree with each other.

## Related

- **#4127** — export ordering, but at the **line** level (issue vs memory lines). A different bug; sorting lines by `id` does not touch intra-array order.
- **#4231** (closed) / **#4436** (tracking) — deferred because JSONL is being deprecated as source of truth. As noted above, this defect is in the storage layer and is not confined to JSONL.

---

<sub>🤖 Investigated and written by a **[Claude Code](https://claude.com/claude-code) session** (Opus 4.8), filed under @ek's account at their direction. Flagging that plainly so you can weigh it accordingly — and because the useful counterpart to the disclosure is that every claim here is a *measurement*, not an inference, and all of them are reproducible: the counts come from a live `bd 1.1.0 (dev)` / Dolt 2.1.9 server-mode database (839 issues, 201 with ≥2 dependencies) and can be re-derived with `bd sql`; the source references are against tip `303e263`. Where I could not verify something I have said so explicitly (the `childToParent` item is marked latent because no multi-parent rows exist in my database). Happy to supply the exact queries, or to open a PR with the tiebreaker plus a fan-out test, if that would help.</sub>

vishnujayvel:
## Summary

Dependency bulk-loaders ordered only by `issue_id`, which groups rows but does not totally order dependencies *within* an issue. The engine could return different intra-issue orders depending on plan/access path, which showed up as pure reorder noise in `bd export` and unstable dependency arrays in ready/list JSON.

Both loaders now use:

```sql
ORDER BY issue_id, depends_on_id, type
```

matching the labels bulk-loader style (`issue_id, label`). `(issue_id, target)` is unique, so this is a total order.

Fixes #4749.

Thanks for the detailed measurement and diagnosis in the issue.

## Test plan

- [x] `go test ./internal/storage/issueops/ -count=1 -run 'DependencyRecords|GetAllDependency'` — pass
- [x] Updated sqlmock regex for full-scan `ORDER BY`
- [x] New `TestGetDependencyRecordsForIssuesOrdersByDependsOnID` — asserts IN-list query includes the tiebreaker and preserves row order

### DIFF
--- a/internal/storage/issueops/dependency_queries.go
+++ b/internal/storage/issueops/dependency_queries.go
@@ -30,11 +30,12 @@ func GetAllDependencyRecordsInTx(ctx context.Context, tx DBTX) (map[string][]*ty
 func getAllDependencyRecordsIntoFromTable(ctx context.Context, tx DBTX, depTable string, result map[string][]*types.Dependency) error {
 	// Total order: issue_id alone is only a grouping key; without a tiebreaker the
 	// intra-issue dependency slice is plan-dependent (export churn, unstable --json).
-	// Mirrors labels bulk-load (issue_id, label). Unique key is (issue_id, target).
+	// Mirrors labels bulk-load (issue_id, label). The separate typed-target unique
+	// keys don't make (issue_id, depends_on_id, type) total, so `id` closes it.
 	rows, err := tx.QueryContext(ctx, fmt.Sprintf(`
 			SELECT issue_id, %s AS depends_on_id, type, created_at, created_by, metadata, thread_id
 			FROM %s
-			ORDER BY issue_id, depends_on_id, type
+			ORDER BY issue_id, depends_on_id, type, id
 		`, DepTargetExpr, depTable))
 	if err != nil {
 		return fmt.Errorf("get all dependency records from %s: %w", depTable, err)
@@ -113,7 +114,7 @@ func getDependencyRecordsIntoFromTable(ctx context.Context, tx DBTX, depTable st
 		}
 		rows, err := tx.QueryContext(ctx, fmt.Sprintf(
 			`SELECT issue_id, %s AS depends_on_id, type, created_at, created_by, metadata, thread_id
-			 FROM %s WHERE issue_id IN (%s) ORDER BY issue_id, depends_on_id, type`,
+			 FROM %s WHERE issue_id IN (%s) ORDER BY issue_id, depends_on_id, type, id`,
 			DepTargetExpr, depTable, strings.Join(placeholders, ",")), args...)
 		if err != nil {
 			return fmt.Errorf("get dependency records from %s: %w", depTable, err)
--- a/internal/storage/issueops/dependency_queries.go
+++ b/internal/storage/issueops/dependency_queries.go
@@ -28,10 +28,13 @@ func GetAllDependencyRecordsInTx(ctx context.Context, tx DBTX) (map[string][]*ty
 
 //nolint:gosec // G201: depTable is "dependencies" or "wisp_dependencies" (hardcoded by caller).
 func getAllDependencyRecordsIntoFromTable(ctx context.Context, tx DBTX, depTable string, result map[string][]*types.Dependency) error {
+	// Total order: issue_id alone is only a grouping key; without a tiebreaker the
+	// intra-issue dependency slice is plan-dependent (export churn, unstable --json).
+	// Mirrors labels bulk-load (issue_id, label). Unique key is (issue_id, target).
 	rows, err := tx.QueryContext(ctx, fmt.Sprintf(`
 			SELECT issue_id, %s AS depends_on_id, type, created_at, created_by, metadata, thread_id
 			FROM %s
-			ORDER BY issue_id
+			ORDER BY issue_id, depends_on_id, type
 		`, DepTargetExpr, depTable))
 	if err != nil {
 		return fmt.Errorf("get all dependency records from %s: %w", depTable, err)
@@ -110,7 +113,7 @@ func getDependencyRecordsIntoFromTable(ctx context.Context, tx DBTX, depTable st
 		}
 		rows, err := tx.QueryContext(ctx, fmt.Sprintf(
 			`SELECT issue_id, %s AS depends_on_id, type, created_at, created_by, metadata, thread_id
-			 FROM %s WHERE issue_id IN (%s) ORDER BY issue_id`,
+			 FROM %s WHERE issue_id IN (%s) ORDER BY issue_id, depends_on_id, type`,
 			DepTargetExpr, depTable, strings.Join(placeholders, ",")), args...)
 		if err != nil {
 			return fmt.Errorf("get dependency records from %s: %w", depTable, err)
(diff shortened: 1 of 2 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace through the pinned engine; Claude adjudication re-verified at the PR head). Verdict: **needs changes** — one small should-fix and this lands.

**Should-fix — the order isn't schema-total; add the primary key as final tie-breaker.** Verified: both loaders order by `issue_id, depends_on_id, type` only. The schema keeps three *separate* typed-target uniqueness constraints (not one over the coalesced target), so two distinct rows can legally carry the same `(issue_id, COALESCE(target), type)` via different typed columns — main's own migration 0058 handles exactly the legacy case of an identifier resolving as both issue and wisp. The pinned go-mysql-server applies `sort.Stable` over only the supplied fields — no hidden PK tie-break — so the advertised total order can still tie. `ORDER BY issue_id, depends_on_id, type, id` closes it for both queries.

Nits (tests): `flex` → a descriptive regex name; the long projection/order regex is duplicated in two places (shared builder prevents drift); and the partitioning commentary describes an abandoned approach while the test deliberately calls the FromTable fast path.

Drift: 128 commits behind but textually clean; main's newer target-keyed readers sit beside (not instead of) these export/ready loaders, and the constraint-restore commit retains the separate typed unique keys — the finding stands on current main. Good, wanted fix: deterministic export/ready output is exactly what the merge-safety work needs.

vishnujayvel:
Thanks for the thorough review — the cross-vendor trace made the should-fix easy to verify.

Addressed in 92d1351: added `id` as the final tie-breaker on both loaders — `ORDER BY issue_id, depends_on_id, type, id`. Your reading matches what I found: because the schema keeps three separate typed-target uniqueness constraints rather than one over the coalesced target, two distinct rows can legally share `(issue_id, COALESCE(target), type)`, so the advertised order could still tie.

On the nits: renamed `flex` → `depsForIssuesRegex`, pulled the duplicated SELECT projection into a single shared const so the two query shapes can't drift, and replaced the stale partitioning comment (the test calls the `FromTable` fast path, so there's no partition query to expect).

`go test ./internal/storage/issueops/...` and `gofmt` are clean. One focused commit on top of the existing history — no rebase needed.

maphew (approved):
Both ORDER BY clauses now carry the `id` tie-breaker and the dead regex helper cleanup addressed the nit too — exactly the requested shape, thanks for the quick fix. Checks are green; approving, and the merge patrol takes it from here.

PR #4830, opened 2026-07-15
fix(dolt): show remotes from persisted repo_state when store is nil (#4619)

### LINKED ISSUE #4619: embedded mode: bd dolt show reports "Remotes: (none)" while bd dolt remote list shows origin
## Summary

In embedded mode, `bd dolt show` always prints `Remotes: (none)` even when an `origin` remote is configured and fully working (`bd dolt remote list` shows it; push/pull succeed).

## Versions

- bd 1.1.0 (linux/amd64, WSL2)
- embedded Dolt backend (default), database cloned from a DoltHub remote

## Repro

```bash
bd init --remote https://doltremoteapi.dolthub.com/<org>/<db> --non-interactive
bd dolt remote list
# origin               https://doltremoteapi.dolthub.com/<org>/<db>
bd dolt show
# ...
# Remotes:
#   (none)
```

## Root cause (from reading main)

`bd dolt show` is registered as a no-DB command (`"dolt"` is in `noDbCommands` in `cmd/bd/main.go`, and `show` is not in `needsStoreDoltSubcommands`), so store initialization is skipped and `getStore()` returns nil inside `showDoltConfig` (`cmd/bd/dolt.go`). The remotes section is:

```go
st := getStore()
var remotes []storage.RemoteInfo
if st != nil {
    remotes, _ = st.ListRemotes(ctx)
}
if len(remotes) == 0 {
    fmt.Println("  (none)")
}
```

With `st == nil` this unconditionally prints `(none)`. `bd dolt remote list` works because `remote` is in `needsStoreDoltGrandchildren`, so it gets a real store.

Related but distinct from #3689 (server-mode SQL/CLI remote divergence): this one is purely the embedded/no-store display path.

## Expected

`bd dolt show` should report the same remotes that `bd dolt remote list` reports. The codebase already has an engine-free source of truth for exactly this cold-start situation: `doltutil.PersistedRemotes` reads `<db>/.dolt/repo_state.json` (used by the remote-migrate gate and the push/pull no-remote skip, GH#2118 / bd-578h9.10). `showDoltConfig` can fall back to it when the store is nil or answers empty.

Happy to send a PR with that fix plus a regression test.

vishnujayvel:
## Summary

In embedded mode, `bd dolt show` always printed `Remotes: (none)` even when `bd dolt remote list` showed a working `origin`.

`dolt show` is a no-store diagnostic (`getStore()` is nil), so the remotes block never called `ListRemotes`. This change resolves remotes for show via:

1. `store.ListRemotes` when a store happens to be present, else  
2. `doltutil.PersistedRemotes` under `.beads/embeddeddolt[/database]` (and server-mode `dolt[/database]` paths) — the same on-disk `repo_state.json` source used by the remote-migrate gate.

Fixes #4619.

Thanks for the clear root-cause analysis and suggested approach.

## Test plan

- [x] `go test ./cmd/bd/ -count=1 -run 'ResolveDoltShowRemotes'` with icu4c CGO flags — pass  
- [x] `TestResolveDoltShowRemotesFromPersistedState` — fixture `repo_state.json` with origin  
- [x] `TestResolveDoltShowRemotesNoneWhenNoState` — empty when no `.dolt` state

### DIFF
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -1441,27 +1441,36 @@ func selectedDoltBeadsDir() string {
 // `show` is a no-store diagnostic command, so getStore() is usually nil and
 // ListRemotes is unavailable. Fall back to on-disk repo_state.json (same
 // source as the remote-migrate gate) so remotes match `bd dolt remote list`
-// in embedded mode (GH#4619).
-func resolveDoltShowRemotes(beadsDir string, cfg *configfile.Config, embeddedDataDir string) []storage.RemoteInfo {
+// (GH#4619).
+//
+// Only the candidate path(s) for the active mode (embedded vs. server) are
+// probed; a repo in one mode must not surface stale remotes persisted under
+// the other mode's data directory. Within the mode-appropriate candidates,
+// the first repo_state.json found on disk is authoritative: an empty
+// remotes list there means "no remotes", not "keep looking" — this stops
+// an authoritative-but-empty active database from falling through to a
+// stale candidate. A corrupt or unreadable repo_state.json is surfaced as a
+// warning rather than silently rendered as "(none)".
+func resolveDoltShowRemotes(beadsDir string, cfg *configfile.Config, embeddedDataDir string, embedded bool) []storage.RemoteInfo {
 	ctx := context.Background()
 	if st := getStore(); st != nil {
 		if remotes, err := st.ListRemotes(ctx); err == nil && len(remotes) > 0 {
 			return remotes
 		}
 	}
-	// Probe persisted remotes under common Dolt data roots without opening a store.
 	dbName := ""
 	if cfg != nil {
 		dbName = cfg.GetDoltDatabase()
 	}
 	var candidates []string
-	if embeddedDataDir != "" {
-		candidates = append(candidates, embeddedDataDir)
-		if dbName != "" {
-			candidates = append(candidates, filepath.Join(embeddedDataDir, dbName))
+	if embedded {
+		if embeddedDataDir != "" {
+			candidates = append(candidates, embeddedDataDir)
+			if dbName != "" {
+				candidates = append(candidates, filepath.Join(embeddedDataDir, dbName))
+			}
 		}
-	}
-	if beadsDir != "" {
+	} else if beadsDir != "" {
 		candidates = append(candidates, filepath.Join(beadsDir, "dolt"))
 		if dbName != "" {
 			candidates = append(candidates, filepath.Join(beadsDir, "dolt", dbName))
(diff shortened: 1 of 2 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified at the PR head). Verdict: **needs changes** — the repo_state.json decoding approach is structurally sound (pinned Dolt's dolt_remotes reads the same state); the resolver's selection logic needs tightening.

**Should-fix 1 — remotes can come from an inactive database.** Verified: the resolver builds a candidate list that always tries embedded paths before legacy/server paths and returns the *first non-empty* result, ignoring the active mode the same function computes later. A repo switched to server mode can display stale embedded remotes; an authoritative-but-empty active database falls through to stale `.beads/dolt` state. Pick the one mode-appropriate path and honor an authoritative empty result.

**Should-fix 2 — corrupt state renders as "(none)".** Verified: every `PersistedRemotes` error hits `continue`. That helper deliberately distinguishes missing from unreadable/malformed — its own regression test says corruption must "never silently none" — and the existing remote-migrate caller warns instead of swallowing. Surface a warning line rather than "(none)".

**Should-fix 3 — the new tests mutate the package-global `store` without restoring it** and don't control `cmdCtx` (which `getStore()` prefers when non-nil), making them order-dependent. The save/restore pattern already exists a few hundred lines up in the same file.

Drift is clean; `dolt show` correctly skips store init so no open-time concerns. With mode-aware selection and error surfacing, this is a genuinely useful visibility fix.

maphew:
Following the 07-24 review, I've pushed the specified fix directly to this branch as maintainer commit `8701e7514` (your commits untouched underneath): `resolveDoltShowRemotes` now probes only the active mode's candidate paths, a directory with an on-disk `repo_state.json` is authoritative even when its remotes map is empty (no more falling through to a stale candidate), and a corrupt `repo_state.json` warns instead of silently rendering `(none)`. Three new tests cover mode-appropriateness, the authoritative-empty case, and the corrupt-state warning. Once CI is green this merges — with your two sibling PRs (#4832, #4829) approved on your own fixes today, this clears your whole reviewed batch.

maphew (approved):
Approving with maintainer commit 8701e7514 applied per the 07-24 review (see comment). _claude-fable-5-medium on behalf of maphew_

vishnujayvel:
Thanks for pushing `8701e7514` rather than bouncing it back — the mode-appropriate probe plus treating an on-disk `repo_state.json` as authoritative even when its remotes map is empty is cleaner than what I had been putting together, and the corrupt-state warning is the right call given `PersistedRemotes` already distinguishes missing from unreadable.

I had a fix for the same three findings in progress and discarded it in favour of yours to avoid a duplicate commit on an approved branch.

PR #4831, opened 2026-07-15
fix(dolt): align show config-source priority with DefaultConfig (#4511)

### LINKED ISSUE #4511: bd dolt show advertises a config-source priority order that contradicts DefaultConfig port resolution
**Problem.** `bd dolt show` ends with a "Config sources (priority order)" block that misdescribes how the port it just printed was actually resolved. An operator debugging a wrong-port / cross-project-leak situation who trusts this block will look in the wrong place and reach a wrong conclusion.

**Root cause (origin/main, `fa4dce454`).** The advisory text is hardcoded in `showDoltConfig` (`cmd/bd/dolt.go:1372-1376`):

```
Config sources (priority order):
  1. Environment variables (BEADS_DOLT_*)
  2. metadata.json (local, gitignored)
  3. config.yaml (team defaults)
```

but the resolution `show` actually consumes one screen earlier (`dsCfg := doltserver.DefaultConfig(beadsDir)`, `dolt.go:1293-1295`) lives in `DefaultConfig` (`internal/doltserver/doltserver.go:459-525`):

1. `BEADS_DOLT_SERVER_PORT` env (`:472`)
2. **port file `.beads/dolt-server.port`** — comment "primary persistent source… Elevated to top priority (after env var)" (`:481-487`)
3. `config.yaml` / global `dolt.port` (`:491-498`)
4. `metadata.json DoltServerPort` — comment "Deprecated: metadata.json DoltServerPort is **git-tracked** and propagates to all contributors… Emit a one-time warning but still use the value as a fallback" (`:503-516`)

Four concrete contradictions: (a) the port file — the documented *primary* persistent source — is **absent** from the `show` list; (b) `metadata.json` is ranked #2 in `show` but is the **lowest** fallback in code; (c) `show` labels `metadata.json` "(local, gitignored)" while the code comment says it is git-tracked and was deprecated *precisely because* it is not local/gitignored (GH#2372) — the label is factually inverted; (d) `config.yaml` is ranked below `metadata.json` in `show` but above it in code.

**Why it matters to a downstream consumer.** Gas City runs a shared, gc-managed dolt sql-server and triages port/endpoint drift by reading the port file and the resolved config. An operator running `bd dolt show` during a drift incident is pointed at `metadata.json` as a high-priority "local" source and the port file is hidden — directly counter to the GH#2372 hardening this workspace depends on.

**Dedup.** Searched `dolt show config sources priority`, `port resolution order`, `metadata.json gitignored port`: closest are #2073, #2372, #2555 (all closed, about resolution *logic*, not the `show` output text). No issue/PR about the `show` advisory text. `git log -15` on `cmd/bd/dolt.go` and `doltserver.go` shows no in-flight change to either the print block or `DefaultConfig` ordering.

**Fix candidates.**
1. Replace the static block with text generated from the same precedence `DefaultConfig` walks (including the port file and the deprecated/git-tracked status of `metadata.json`) — single source of truth.
2. If a generated block is too heavy, correct the static text: insert `.beads/dolt-server.port (gitignored, primary)` after the env var, move `config.yaml` above `metadata.json`, relabel `metadata.json` `(deprecated, git-tracked)`.
3. Add a test asserting the `show` ordering string matches the order enumerated by `DefaultConfig`, so the two cannot drift again.

---
_Filed by the Gas City maintainer fleet from a downstream interdependency audit; verified against `gastownhall/beads` origin/main @ `fa4dce454`._

vishnujayvel:
## Summary

`bd dolt show` printed a "Config sources (priority order)" block that did not match how `doltserver.DefaultConfig` actually resolves the server port:

- omitted `.beads/dolt-server.port` (primary gitignored source after env)
- ranked `metadata.json` above `config.yaml` and labeled it "local, gitignored" (it is git-tracked / deprecated for ports)

The banner now lists: env → port file → config.yaml/global → metadata.json (deprecated), with a small contract test so the strings cannot drift silently.

Fixes #4511.

## Test plan

- [x] `go test ./cmd/bd/ -count=1 -run TestDoltShowConfigSourceLinesMatchDefaultConfigOrder` (icu4c CGO) — pass


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified at the PR head). Verdict: **needs changes** — the banner is closer than before but still doesn't match `DefaultConfig`, and the guard-test doesn't guard.

**Should-fix 1 — the displayed chain is still inaccurate on three points.** Verified against the hard-coded slice: (a) it advertises `BEADS_DOLT_*` broadly, but `DefaultConfig` checks exactly `BEADS_DOLT_SERVER_PORT`; (b) `.beads/dolt-server.port` is wrong for shared mode, where the port file lives under the resolved `~/.beads/shared-server` directory; (c) two distinct YAML sources with distinct precedence (Dolt data-dir `config.yaml` `listener.port`, then Beads project/global `dolt.port`) are collapsed into one entry labeled `dolt.port`. The real chain: exact env override → resolved port file → Dolt server YAML → Beads project/global YAML → metadata fallback.

**Should-fix 2 — the "contract test" checks token membership plus one ordering (port file before metadata) on the copied display slice.** It never asserts env-before-portfile or portfile-before-YAML, and it tests the copy, not the resolver — several wrong permutations pass, and future `DefaultConfig` changes drift silently. Either derive the display lines from a shared representation of the resolution sources, or write a behavioral test against `DefaultConfig` itself.

That drift seam is the real fix here: this is the second manually-maintained copy of the resolution order (the #4511 complaint was exactly this class of divergence). Kill the copy and the bug class dies with it.

vishnujayvel:
Thanks for the careful re-read, @maphew — both should-fixes check out, and you're right about the drift seam being the real bug.

**Should-fix 1 (accuracy).** Verified all three points against `internal/doltserver.DefaultConfig`:

- (a) The env check is `os.Getenv("BEADS_DOLT_SERVER_PORT")` specifically — the "`BEADS_DOLT_*` broadly" wording was wrong.
- (b) `DefaultConfig` reassigns `beadsDir` to `SharedServerDir()` (`~/.beads/shared-server`) *before* calling `readPortFile` when shared mode is on, so the port file is not always `.beads/dolt-server.port` — confirmed.
- (c) There are indeed two distinct YAML sources with distinct precedence: `configYamlPort(beadsDir)` reads `listener.port` from the Dolt server's own `config.yaml` in the resolved dolt data directory (via `servercfg.YamlConfigFromFile`), checked *before* `config.GetYamlConfig("dolt.port")`, which reads Beads' own project/global `config.yaml`. The old banner collapsed these into one "config.yaml" line.

The true chain, read directly out of `DefaultConfig`, matches what you described: exact env override (`BEADS_DOLT_SERVER_PORT`) → resolved port file (shared-server dir in shared mode) → Dolt server config.yaml (`listener.port`) → Beads project/global config.yaml (`dolt.port`) → metadata.json `dolt_server_port` (deprecated fallback, with a one-time stderr warning).

**Should-fix 2 / the drift seam.** Agreed that hand-correcting the copy would just leave a corrected copy that can drift again. Instead of a second manually-maintained slice, `DefaultConfig` now walks an ordered slice of `(label, resolve)` pairs (`internal/doltserver.portSources`); `PortSourceLabels()` exposes the same slice's labels, and `bd dolt show-config`'s `printDoltShowConfigSources` renders those labels directly. The CLI text is generated from the same data the resolver walks, so it can't diverge from actual behavior again.

Added `TestDefaultConfigPrecedenceChain` in `internal/doltserver`, a behavioral test against `DefaultConfig` itself (not the display copy): it layers each source on top of the previous ones in a temp dir and asserts the higher-priority source wins at every step — including the two previously-untested transitions (env-before-port-file, port-file-before-either-YAML-source) and the split between the two YAML sources that the old contract test collapsed. I verified this test is load-bearing by temporarily swapping the two YAML source entries in `portSources`: the test failed with a clear precedence mismatch (`port = 5002, want 5003`), then I reverted the swap.

Commands run, from a clean worktree at the PR's current head:
- `gofmt -l` on all touched files — clean, no output.
- `go build ./...` — clean (only pre-existing "ignoring duplicate libraries" linker warnings, unrelated to this change).
- `go test ./internal/doltserver/...` — all pass, including the new `TestDefaultConfigPrecedenceChain` and existing `TestDefaultConfig*`/`TestDefaultConfigReadsConfigYamlPort` cases.
- `go test ./cmd/bd/...` — all pass, including the rewritten `TestDoltShowConfigSourcesRendersPortSourceLabels`, which checks the CLI output against `doltserver.PortSourceLabels()` directly.
- `go test ./cmd/bd/doctor/...` — `TestCheckBeadsRole_NotConfigured` and `TestCheckBeadsRole_NotGitRepo` fail on this machine; this is pre-existing and environmental (a global `git config beads.role=maintainer` leaks in), reproduces identically on an unrelated branch, and is unrelated to this change.

Files touched: `internal/doltserver/doltserver.go` (portSources chain + PortSourceLabels + DefaultConfig refactor), `internal/doltserver/portsources_test.go` (new behavioral precedence test), `cmd/bd/dolt.go` (printDoltShowConfigSources now renders PortSourceLabels), `cmd/bd/dolt_test.go` (contract test rewritten to check the rendering, not a hand-copied slice).

One housekeeping note: GitHub now reports this branch as conflicting with `main`. I'll rebase and push that as its own step rather than mixing an unrelated merge into this review round — flagging so the red state isn't mistaken for a problem with the change above.

maphew (approved):
Approving — this round did the right thing in the strongest way: instead of hand-correcting the duplicated priority list, you deleted the duplication and derived the banner from the resolver itself. We verified `portSources` matches `DefaultConfig`'s actual resolution order exactly, `PortSourceLabels()` is pure (no deprecation warning fires from rendering), and the two previously-untested transitions are now genuinely covered by `TestDefaultConfigPrecedenceChain`.

Also: no rebase needed — the PR reports clean/mergeable as-is, so please don't push a merge commit; we'll land it from here.

Two nits, neither blocking:

1. Behavior change hiding in the refactor: `BEADS_DOLT_SERVER_PORT=0` used to return immediately after a successful `Atoi`, skipping the shared-mode fallback; the new loop breaks and falls through, so under shared mode the port becomes 3308 instead of 0. Nothing in-tree exercises it, but `return port, err == nil && port > 0` in the env resolver would restore the old semantics and make env consistent with every other source's `>0` guard.
2. The banner is now titled "Config sources for server port", but `bd dolt show-config` also prints host/user/database/mode with no documented sources. A one-line footer (or explicitly scoping the banner) would keep it honest.

Happy to take either as a follow-up ourselves.

PR #4832, opened 2026-07-15
fix(swarm): exclude closed issues from ready-front waves (#4564)

### LINKED ISSUE #4564: bd swarm validate lists closed beads in ready-fronts / wave breakdown (status=closed not filtered)
## Summary

`bd swarm validate <epic>` lists **closed** beads in the "Ready Fronts (waves of parallel work)" breakdown. The wave/ready-front computation does not filter `status=closed`, even though the epic header's closed-count is computed correctly.

## Environment

- `bd version 1.0.5 (6a3f515ce: HEAD@6a3f515ced18)` (build ahead of the v1.0.4 stable release)
- Backend: Dolt

## Reproduction

Running `bd swarm validate <epic>` on an epic that has closed leaf children (closed, with no open blockers) produces a "Ready Fronts" breakdown whose Wave 1 includes those closed children.

Observed concretely: an epic with two children closed as resolved still had both listed in Wave 1 of the ready-fronts breakdown. Both were confirmed `CLOSED` via `bd show`. The epic header itself correctly reported the closed count (e.g. `Total issues: 40 (18 closed)`) — so the underlying status is right; only the ready-fronts/wave computation fails to exclude `status=closed`.

## Expected

Closed issues should be excluded from wave membership and from the max-parallelism count in `bd swarm validate`, consistent with the (correct) epic-header closed count.

## Suggested test spec

- **Unit:** the swarm ready-front computation should exclude `status=closed` issues from wave membership — construct an epic fixture with a closed child that has no open blockers, assert it is absent from all waves.
- **Integration:** run `bd swarm validate <epic>` against a real Dolt DB containing a closed leaf child; assert the closed ID appears in neither the wave listings nor the max-parallelism count.

vishnujayvel:
## Summary

`bd swarm validate` listed **closed** children in "Ready Fronts" waves. The wave algorithm treated every child as work to schedule and counted closed nodes as unresolved blockers.

This change:

- Excludes `status=closed` issues from ready-front membership
- Treats closed dependencies as already satisfied for open dependents
- Sets estimated sessions from open issues only

Fixes #4564.

## Test plan

- [x] `go test ./cmd/bd/ -count=1 -run TestComputeReadyFronts` (icu4c CGO) — pass
  - closed leaf not in wave 0
  - closed blocker does not hold open dependent forever; open-only path advances

### DIFF
--- a/cmd/bd/swarm.go
+++ b/cmd/bd/swarm.go
@@ -314,6 +314,13 @@ func analyzeEpicForSwarm(ctx context.Context, s SwarmStorage, epic *types.Issue)
 	return analysis, nil
 }
 
+// issueIsClosed reports whether id is closed. Closed issues count as satisfied
+// and are excluded from cycle detection and ready-front scheduling (GH#4564).
+func issueIsClosed(analysis *SwarmAnalysis, id string) bool {
+	n, ok := analysis.Issues[id]
+	return ok && n.Status == string(types.StatusClosed)
+}
+
 // detectStructuralIssues looks for common problems in the dependency graph.
 //
 //nolint:unparam // issues reserved for future use
@@ -402,6 +409,7 @@ func detectStructuralIssues(analysis *SwarmAnalysis, _ []*types.Issue) {
 
 	// 5. Detect cycles using simple DFS
 	// (The main DetectCycles in storage is more sophisticated, but we do a simple check here)
+	// Closed issues are excluded: a closed cycle must not block open children (GH#4564).
 	inProgress := make(map[string]bool)
 	completed := make(map[string]bool)
 	var cyclePath []string
@@ -409,6 +417,9 @@ func detectStructuralIssues(analysis *SwarmAnalysis, _ []*types.Issue) {
 
 	var detectCycle func(id string) bool
 	detectCycle = func(id string) bool {
+		if issueIsClosed(analysis, id) {
+			return false
+		}
 		if completed[id] {
 			return false
 		}
@@ -434,6 +445,9 @@ func detectStructuralIssues(analysis *SwarmAnalysis, _ []*types.Issue) {
 	}
 
 	for id := range analysis.Issues {
+		if issueIsClosed(analysis, id) {
+			continue
+		}
 		if !completed[id] {
 			if detectCycle(id) {
 				break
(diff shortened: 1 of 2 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the call ordering at the PR head). Verdict: **needs changes** — one should-fix; the wave-filtering itself is sound.

**Should-fix — closed-node cycles still suppress all ready fronts.** Verified: `detectStructuralIssues` runs before `computeReadyFronts`, its cycle detection traverses closed nodes, and `computeReadyFronts` returns early on any recorded error — before the new closed-node filtering gets a chance. Concrete counterexample: closed A ⇄ open B. With A satisfied and excluded (this PR's own semantics), B should be wave 0; instead validation reports a cycle and no fronts are produced. An all-closed cycle likewise suppresses unrelated open children. The tests call `computeReadyFronts` directly, bypassing `detectStructuralIssues`, so they can't catch it. Exclude closed nodes from the active cycle graph (the same closure predicate the fronts now use), and add a full-`analyzeEpicForSwarm` test for the closed-cycle case.

No standards findings; wiring verified; zero merge-base drift (main's swarm.go blob is identical to the merge base). With the cycle graph aligned to the same closed-exclusion rule, this lands.

vishnujayvel:
Thanks — you're right that the wave-filtering alone wasn't enough, and your counterexample is exactly the hole.

Addressed in e12a56f: `detectStructuralIssues`'s cycle DFS traversed every child regardless of status and runs before `computeReadyFronts`, which bails out on any recorded error — so a closed-node cycle (or an all-closed cycle) suppressed ready fronts even for unrelated open children. Pulled the closed-status predicate `computeReadyFronts` already used into a shared `issueIsClosed` helper and reused it in the cycle DFS, skipping closed issues both as DFS roots and as traversal targets, so a closed node can no longer participate in a reported cycle.

Added `TestAnalyzeEpicForSwarmClosedCycleDoesNotSuppressOpenFronts`, a full-path `analyzeEpicForSwarm` test over exactly your case (closed A ⇄ open B plus an unrelated open C). Confirmed it fails on the pre-fix code (cycle error reported, zero ready fronts) and passes after the fix (open-b and open-c both land in wave 0, closed-a excluded).

`go build ./...`, the swarm tests in `cmd/bd`, and `gofmt` are clean.

maphew (approved):
Your follow-up commit does exactly what the review asked — closed nodes excluded from the cycle-detection graph itself, so the predicate and the wave-builder now agree — and the fast turnaround is appreciated. Checks are green; approving, and the merge patrol takes it from here.

PR #4833, opened 2026-07-15
fix(create): hard-fail multi-type same-target --deps (#4626)

### LINKED ISSUE #4626: bd create --deps with a comma-separated multi-type value (e.g. discovered-from:X,blocked-by:X) silently registers only one dep type
## Observed (mlcc-tcad, 2026-07-07, bead mlcc-zrn6x)

`bd create --deps 'discovered-from:X,blocked-by:X'` (a single `--deps` flag value containing two comma-separated `type:id` pairs targeting the same base issue) silently registers only ONE of the two dependency types — the other is dropped with no warning, error, or partial-application notice.

This produced two mint-parked TRACKER-ONLY follow-up issues (in our repo: `mlcc-zrn6x`, `mlcc-su32p`) that both the issue text and the minting agent's intent stated should be `blocked-by` their base issue, but neither actually carries that gate — `bd ready`/dependency resolution would have let them be dispatched against the pre-merge trunk, colliding with the base once it landed. It was caught by a human/agent noticing the missing dependency before dispatch, not by any tooling.

## Why this matters

`--deps` accepting a comma-separated list is presumably intended to support exactly this multi-type-against-one-target case (per the CLI help text: `'type:id' or 'id' (e.g., 'discovered-from:bd-20,blocked-by:bd-15' ...)`) — but that documented example is two DIFFERENT ids; it's not clear from the docs whether two different TYPES against the SAME id is meant to be supported, and empirically only one wins with no signal either way.

## Suggested fix

Either:
1. Make `--deps` genuinely support multiple dependency types against the same target id in one comma-separated value (matching what a reasonable reading of the help text implies), or
2. If that's not intended, have `bd create`/`bd update --deps` detect a dropped/collapsed entry (e.g., by comparing the number of parsed `type:id` pairs against what was actually persisted) and emit a warning or hard error rather than silently accepting a partial application.

Either fix would prevent a silently-ungated issue from being dispatched as if it had a real block on its base.

vishnujayvel:
## Summary

`bd create --deps 'discovered-from:X,blocked-by:X'` could silently keep only one edge: dependency uniqueness is per `(issue_id, target)` (`uk_dep_issue_target`), not per type, and `AddDependency` failures were only warned.

This change:

1. Expands comma-separated `--deps` tokens and parses via `parseDepSpecs`
2. Rejects multi-type / duplicate edges that would collide on the same target **before** create
3. Hard-fails create if `AddDependency` errors (no silent partial apply)

Schema still does not allow two types to the same target; the CLI now fails loudly instead of minting an un-gated issue.

Fixes #4626.

## Test plan

- [x] `go test ./cmd/bd/ -count=1 -run 'TestParseDepSpecs|TestDiscoveredFromParent'` — pass
  - multi-type same target → error
  - comma-separated different targets still parse

## Follow-up hardening
- Validate `--deps` (including multi-type same-target uniqueness) **before** `CreateIssue` so a rejection cannot leave an orphan issue.


vishnujayvel:
Merged latest `main` into this branch to clear the conflict (merge commit, history preserved).

The only conflict was in `cmd/bd/create.go`: `main` refactored issue creation into the atomic `createIssueWithDeps` path, which overlapped this PR's change to parse `--deps` before creating the issue. Resolved by keeping `main`'s atomic path while preserving this PR's fail-closed behavior — `--deps` is still fully parsed and validated up front (the multi-type same-target hard-fail from #4626 lives in `parseDepSpecs`/`create_deps.go`, which the atomic path calls), so no orphan issue can be created on a bad `--deps` spec. No other files conflicted.

Focused tests on the affected logic (`TestCreate*Dep*`/`TestParseDepSpecs`) pass locally with `BEADS_TEST_SKIP=dolt` (no Docker daemon on this machine for the dolt integration harness); CI should confirm the full suite.

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the multi-type same-target hazard is worth fixing, but the current guard is both too strict and porous, and the orphan rationale has drifted.

**Should-fix 1 — duplicate same-type edges become fatal.** The validator rejects every repeated target/direction, including aliases that normalize identically — but both storage paths deliberately treat same-pair/same-type additions as idempotent and reject only *differing* types. `--deps blocked-by:X --deps depends-on:X`, previously accepted, now fails. Deduplicate identical normalized specs; keep the hard failure for genuinely conflicting types.

**Should-fix 2 — comma values are parsed twice.** `--deps` is a pflag `StringSlice`, which already CSV-decodes; the parser then splits on commas again, so a CSV-quoted target like `related:external:foo,bar` becomes two specs. The unit test bypasses Cobra and masks the production representation.

**Should-fix 3 — parsed specs aren't reused for `source_repo` inheritance.** Verified: `discoveredFromParent(deps)` receives the *raw* flag values on both the direct and proxied paths, so for the newly supported combined form (`discovered-from:bd-20,blocks:bd-15`) the parent lookup silently misses and `source_repo` isn't inherited — an earlier commit claimed parsed-spec reuse but the merge dropped that wiring.

**Should-fix 4 — layering:** the CLI error hard-codes the `uk_dep_issue_target` schema index; keep the message domain-oriented (the storage layer already owns same-pair conflict behavior). And add a stock-CLI regression through the real-binary harness — the direct `parseDepSpecs` test proves neither exit code nor absence of an orphan.

**Drift note:** the orphan-prevention rationale is obsolete — atomic create (issue + edges in one transaction with rollback) is an ancestor of this PR's own base, and current main has since landed uniform typed dependency conflicts while preserving same-type idempotency. The PR needs semantic adaptation to that, not just a rebase.

vishnujayvel:
Thanks for the thorough re-review — all four should-fix items are addressed:

1. **Duplicate same-type edges** — `dedupeDepSpecs` now collapses a repeated *identical* normalized edge (e.g. `blocked-by:X` + `depends-on:X`, both aliasing to the same type/target) instead of failing. Only a genuine type conflict on the same target (e.g. `discovered-from:X` + `blocked-by:X`) is still rejected.
2. **Double comma-split** — removed the manual re-split; `--deps` is a cobra `StringSlice` flag that already CSV-decodes, so re-splitting was double-decoding and would mangle a CSV-quoted target containing a literal comma. `parseDepSpecs` now only trims/skips-empty on the already-split elements.
3. **Parsed specs not reused for `source_repo` inheritance** — confirmed: an earlier commit on this branch had that wiring and the merge with `main` silently dropped it. Restored via a new `discoveredFromParentSpec(specs)` that consumes the already-parsed specs, wired into both the direct (`cmd/bd/create.go`) and proxied (`cmd/bd/create_proxied_server.go`) paths you named.
4. **Error message / regression coverage** — reworded the same-target-conflict error to drop the internal `uk_dep_issue_target` index name (domain-oriented wording only), and added real-`bd`-binary regression tests in `create_deps_atomic_test.go` covering both the reject-with-no-orphan case and the dedup-not-reject case, since a direct `parseDepSpecs` unit test can't prove exit code or orphan absence.

Also updated the `TestParseDepSpecs` comma-handling cases to reflect the actual production representation (cobra hands the parser an already-split slice, not one comma-joined string), per your note that the previous test bypassed cobra.

Not touched: `cmd/bd/create_form.go`'s `discoveredFromParent(fv.Dependencies)` — that's a separate, pre-existing lenient raw-string parsing path for form input that never goes through `parseDepSpecs`, so it's out of scope here. The "drift note" on atomic-create ancestry is noted but is context rather than an actionable request.

Tests run locally, all passing:

- `go test ./cmd/bd/ -run 'TestParseDepSpecs|TestDiscoveredFromParent|TestBuildWaitsFor'` — 25/25 subtests.
- `go test ./cmd/bd/ -run TestCreateDepsAtomicity -count=1` — 8/8 subtests (53s), including the two new ones: `multi_type_same_target_rejected_before_create_no_orphan` and `duplicate_identical_dep_is_deduped_not_rejected`.

One honest gap: the Dolt-backed subtests skip on this machine (`WARN: Docker not available`), so the atomic-harness coverage above is the SQLite/in-process path only.

This branch also currently conflicts with `main` — I'll rebase and push that separately rather than fold an unrelated merge into this review round.

maphew (approved):
Approving — all four asks from the last round are genuinely addressed, and we verified rather than assumed: the `SwapDirection|TargetID` dedupe key exactly models the `uk_dep_issue_target` uniqueness for a new issue (so `blocked-by:X`+`depends-on:X` dedupe, `discovered-from:X`+`blocked-by:X` conflict, and `blocks:X`+`blocked-by:X` both survive — matching storage semantics); the CSV handling nets out to no change either way; `source_repo` wiring is restored on both `parseDepSpecs` callers, which are the only two; and the real-binary regressions are in.

No rebase needed — the PR is clean/mergeable as-is; please don't push a merge commit, we'll land it from here. Heads-up that our just-opened #5116 (dep add `--type` alias normalization) touches the same file for a different bug: we ran a cross-merge check and the hunks compose cleanly in either landing order, so neither of us needs to wait on the other.

Two follow-ups we're filing on our side, no action needed from you:

1. The conflict error reports normalized types — a user who typed `blocked-by:X` sees the error name `blocks`. Carrying the raw spelling through to the message would remove one head-scratch.
2. The interactive form path (`create_form.go`) bypasses `parseDepSpecs` entirely, so form-created issues keep the old silent multi-type drop. You called it out of scope, which is reasonable — we're tracking it as its own issue.

Thanks for the disciplined iteration on this one.

PR #4838, opened 2026-07-16
feat(setup): redirect beads block to AGENTS.md when CLAUDE.md is a stub

maphew:
## What

When `CLAUDE.md` is a thin stub that imports `AGENTS.md` via the `@-include` convention (e.g. `@AGENTS.md` on its own line), `bd setup claude` now writes the managed `BEADS INTEGRATION` block to `AGENTS.md` instead of duplicating it in the stub.

## Why

Some repos keep `AGENTS.md` as the single authoritative agent-instruction file and reduce `CLAUDE.md` to a short stub that imports it (`@AGENTS.md`). Claude Code expands `@`-imports, so the `AGENTS.md` content is already in context. Previously, `bd setup claude` wrote its managed block into `CLAUDE.md`, which fought this pattern: re-running setup re-inserted the block into the stub, duplicating content already present in `AGENTS.md` via `bd setup codex` or other recipes.

## How

`claudeAgentsEnv()` now checks whether `CLAUDE.md` contains a standalone `@AGENTS.md` directive (the configured agents file name) and whether that agents file exists. If both are true, it redirects the beads section target to the agents file. The redirect applies to install, check, and remove operations. If the agents file does not exist, the fallback behavior (writing to `CLAUDE.md`) is preserved.

The detection is conservative: only a standalone `@AGENTS.md` line (after trimming whitespace) triggers the redirect. Inline references in prose (e.g. "See @AGENTS.md for details") do not.

Closes beads issue mybd-xq68.

### DIFF
--- a/cmd/bd/setup/claude.go
+++ b/cmd/bd/setup/claude.go
@@ -138,10 +138,13 @@ func stripStaleClaudeBlock(env claudeEnv) error {
 // the file is a thin stub that imports shared agent instructions from the
 // agents file rather than carrying its own content.
 func isAgentsImportStub(content, agentsFile string) bool {
-	directive := "@" + agentsFile
+	directives := []string{"@" + agentsFile, "@./" + agentsFile}
 	for _, line := range strings.Split(content, "\n") {
-		if strings.TrimSpace(line) == directive {
-			return true
+		trimmed := strings.TrimSpace(line)
+		for _, directive := range directives {
+			if trimmed == directive {
+				return true
+			}
 		}
 	}
 	return false
@@ -277,7 +280,11 @@ func installClaude(env claudeEnv, global bool, stealth bool) error {
 		_, _ = fmt.Fprintf(env.stderr, "Warning: failed to update %s: %v\n", claudeInstructionsFile, err)
 	}
 
-	if redirected {
+	// Only strip the stale CLAUDE.md block once the redirect has actually
+	// written the replacement to AGENTS.md. If installAgents skipped injection
+	// (e.g. AGENTS.md is a symlink), stripping here would delete-before-write
+	// and leave the project with no beads section anywhere.
+	if redirected && !agentsSkipped {
 		if err := stripStaleClaudeBlock(env); err != nil {
 			// Non-fatal: the redirect itself already succeeded above
 			_, _ = fmt.Fprintf(env.stderr, "Warning: failed to clean stale beads block from %s: %v\n", claudeInstructionsFile, err)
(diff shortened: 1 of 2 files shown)


maphew (commented):
**Agent review — verdict: concerns**

The core redirect logic is sound, well-structured, and reuses existing agents.go primitives cleanly; detection is appropriately conservative and test coverage for the happy paths is good. Two correctness gaps in the install/remove interaction warrant attention before merge: the stale-block cleanup can delete the CLAUDE.md block without a replacement being written when AGENTS.md is a symlink, and `bd remove claude` now deletes the project-authoritative shared beads block from AGENTS.md. Both are edge/teardown scenarios but represent content loss that the previous behavior did not cause.

**Findings:**
- **major** `cmd/bd/setup/claude.go:280` — In installClaude, stripStaleClaudeBlock runs whenever `redirected` is true, ignoring `agentsSkipped`. installAgents skips injection and calls markSkipped() when the target (AGENTS.md) is a symlink (agents.go:105-113), returning nil. In that case the redirect never actually wrote the beads block to AGENTS.md, yet the code still strips the block from CLAUDE.md - leaving the project with no beads section anywhere (delete-before-write). Gate the strip on `redirected && !agentsSkipped` so the CLAUDE.md block is only removed once the replacement is confirmed written.
- **major** `cmd/bd/setup/claude.go:460` — removeClaude with the redirect active calls removeAgents against AGENTS.md, which deletes the entire `<!-- BEGIN BEADS INTEGRATION -->` section from AGENTS.md (TestRemoveClaudeRedirectsToAgentsMD asserts this). That section is the project-authoritative block created by `bd init`/createNewAgentsFile (agents.go:349-356), not a Claude-specific pointer. Before this PR, `bd remove claude` only touched CLAUDE.md; now uninstalling Claude integration silently removes the shared beads block that the AGENTS.md-authoritative pattern exists to centralize. There is no ownership guard and no test for removing when AGENTS.md carries a full-profile/shared block. Consider limiting the redirected remove to the stale CLAUDE.md cleanup and leaving the shared AGENTS.md section intact, or at minimum documenting/gating this.
- **minor** `cmd/bd/setup/claude.go:145` — isAgentsImportStub only matches a line that trims to exactly `@AGENTS.md`. Claude Code also expands relative import forms such as `@./AGENTS.md`, which will not trigger the redirect and will fall back to writing the block into the stub. This is called out as intentionally conservative in the PR body, but `@./AGENTS.md` is a common, legitimate stub form worth accepting.
- **minor** `cmd/bd/setup/claude.go:90` — When redirecting to an existing AGENTS.md that has no beads block yet, installAgents writes claudeAgentsIntegration's ProfileMinimal. A repo that treats AGENTS.md as the authoritative agent-instruction file generally wants the full profile there (as `bd init` and `bd setup codex` produce). Only pre-existing full blocks are preserved (agents.go:134-140); a blank AGENTS.md gets the minimal profile, which may under-populate the authoritative file.

maphew:
Thanks for the review — both major findings are addressed in 413b8c57b.

**Major 1 (delete-before-write on symlinked AGENTS.md):** `installClaude` now
only strips the stale CLAUDE.md block when `redirected && !agentsSkipped`, so
a symlinked AGENTS.md (which skips injection) no longer leaves the project
with no beads section anywhere. Added
`TestInstallClaudeRedirectSkipsStripWhenAgentsSymlink` as a regression test.

**Major 2 (removeClaude deleting the shared AGENTS.md block):** `removeClaude`
no longer calls `removeAgents` against AGENTS.md when the redirect is active.
It now only cleans up a stale block left directly in CLAUDE.md and leaves the
project-authoritative AGENTS.md section untouched. Updated
`TestRemoveClaudeRedirectsToAgentsMD` and
`TestRemoveClaudeRedirectCleansStaleClaudeBlock` to assert preservation
instead of deletion, and added
`TestRemoveClaudeRedirectPreservesFullProfileSharedBlock` for the full-profile
shared-block case.

**Minor 1 (`@./AGENTS.md`):** Fixed — `isAgentsImportStub` now accepts both
`@AGENTS.md` and `@./AGENTS.md`.

**Minor 2 (blank AGENTS.md profile on redirect):** Deferred. Whether a blank
AGENTS.md should get the full profile instead of minimal on redirect is a
product-intent question (what belongs in an authoritative shared file vs. a
Claude-specific pointer), not a straightforward bug fix — leaving it for a
maintainer call rather than guessing.

Verified: `go build`, `gofmt -l`, `go vet`, and `go test ./cmd/bd/setup/...`
all clean (CGO_ENABLED=0, `-tags gms_pure_go`). I also confirmed the new/
updated tests fail without the fix (reverted the gating locally, re-ran, saw
the expected failures, restored) to make sure they're real regression
coverage and not just green by construction.

PR #4839, opened 2026-07-16
fix(update,close): gate the no-ID last-touched fallback to interactive sessions (bd-m00pb)

steveyegge:
## Summary

`bd update` / `bd close` with no issue ID fall back to the "last touched" issue. Convenient at an interactive prompt — dangerous in scripts, where an empty shell variable silently mutates whatever issue was touched last. In a real multi-agent session (wyvern `wy-c9uoz`, forensics `wy-yuclk`), a failed `bd create --silent` printed nothing and the follow-on `bd update $ID ...` rewrote an unrelated **closed** bead's assignee/status/close_reason, requiring hand restoration.

This PR gates the fallback to interactive sessions:

- Fallback applies only when **stdin is a terminal**.
- Explicit overrides, mirroring the `isNonInteractiveBootstrap` precedence: `BD_LAST_TOUCHED_FALLBACK=1/true` always allows (script opt-in), `=0/false` always denies, then `BD_NON_INTERACTIVE` / `CI` deny, then TTY detection.
- The non-interactive refusal names the opt-in env var, so any scripted caller that relied on the fallback gets a one-line migration path.
- Read-only consumers (`bd show --current`) are unchanged; the proxied-server paths already hard-errored on a missing ID.

## Testing

- New unit tests cover the env-precedence table and the default non-TTY deny.
- New end-to-end test (`TestNoIDFallbackGuard`) drives the real binary: no-ID `update`/`close` refused without a TTY, refused under `BD_NON_INTERACTIVE=1`, honored with `BD_LAST_TOUCHED_FALLBACK=1` (mutating exactly the last-touched issue), and explicit IDs unaffected.
- Full `TestClose*` / `TestUpdate*` suites pass; `gofmt` and `go vet` clean.

## Notes

- CLI reference pages regenerated via `scripts/generate-cli-docs.sh`; this also picks up one pre-existing generated-drift hunk from the auto-backup help text landed in 5f685959c.
- Tracked as bd-m00pb (P1); requested by the wyvern rig (wy-c9uoz).
- Sibling safety ask bd-at6rc (unclaim `--force` guard) will follow as its own PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   note: in server mode that was behaving like `on`; it now genuinely stops
   minting version commits until an explicit `bd dolt commit`.
 
+- **The no-ID "last touched issue" fallback on `bd update` / `bd close` is
+  now interactive-only** (bd-m00pb,
+  [#4839](https://github.com/gastownhall/beads/pull/4839)).
+  Previously a scripted `bd update $ID ...` with an accidentally empty `$ID`
+  silently mutated whatever issue was touched last — a real agent session
+  corrupted an unrelated closed bead this way. The missing-ID refusal happens
+  in argument validation, before any store open, migration, or auto-import
+  side effect. The fallback now requires stdin to be a terminal;
+  `BD_NON_INTERACTIVE=1` and `CI=1/true` also disable it. Scripts that
+  intentionally relied on it can set `BD_LAST_TOUCHED_FALLBACK=1` (or `=0` to
+  disable the fallback everywhere). Explicit IDs and `bd show --current` are
+  unaffected.
+
 
 - **`bd delete --cascade` no longer marks LIVE issues as deleted in other
   issues' text** (bd-x82so). When the deletion named a wisp, the set used to
--- a/cmd/bd/close.go
+++ b/cmd/bd/close.go
@@ -27,7 +27,11 @@ var closeCmd = &cobra.Command{
 	Long: `Close one or more issues.
 
 If no issue ID is provided, closes the last touched issue (from most recent
-create, update, show, or close operation).
+create, update, show, or close operation). This fallback only applies in
+interactive sessions (stdin is a terminal); in scripts and agent sessions a
+missing ID is an error, so a command built from an empty variable cannot
+silently close an unrelated issue. Set BD_LAST_TOUCHED_FALLBACK=1 to allow
+the fallback anywhere, or =0 to disable it entirely.
 
 When closing multiple issues, provide one --reason for all IDs or repeat
 --reason once per ID. Reasons map positionally: the first --reason applies
@@ -50,8 +54,11 @@ the flags appear in the command line.`,
 			return runCloseProxiedServer(cmd, rootCtx, args)
 		}
 
-		// If no IDs provided, use last touched issue
+		// If no IDs provided, use last touched issue (interactive only)
 		if len(args) == 0 {
+			if !AllowLastTouchedFallback() {
+				return HandleErrorRespectJSON("no issue ID provided (the last-touched fallback only applies in interactive sessions; pass an explicit issue ID or set BD_LAST_TOUCHED_FALLBACK=1)")
+			}
 			lastTouched := GetLastTouchedID()
 			if lastTouched == "" {
 				return HandleErrorRespectJSON("no issue ID provided and no last touched issue")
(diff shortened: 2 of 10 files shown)


jacobhausler:
Second independent occurrence of this failure class, filed before we found this PR:
argless `bd close` in our session (city "randland") closed a live P0 (ra-vdisw) that
was last-touched only incidentally — the command was never typed directly, it fired
via backtick command-substitution inside a double-quoted shell string in a bead note.
Reopened ~2 min later, fully restored, no collateral dispatch. Full incident: our
internal bead ra-jrw6a / upstream-candidate writeup ra-ql6hv. The TTY-gated fix here
(explicit BD_LAST_TOUCHED_FALLBACK opt-in) would have prevented it — our failure was
in a non-interactive backtick-substitution context, not an interactive shell, so the
default-deny direction matches. Any chance this can get a review pass? CI is green but
it's been open a week with a merge conflict now blocking it.

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication spot-verified at the PR head). Verdict: **needs changes** — the interactive-gating idea is right; three integration items.

**Blocker — textual conflicts with landed work.** The branch is ~150 commits behind; merge-tree conflicts in `cmd/bd/update.go` and `CHANGELOG.md`. The update.go resolution must preserve the multi-ID partial-failure contract paragraph that landed via b2b153b7b alongside the new fallback documentation.

**Should-fix — the no-ID refusal runs after store-open side effects.** Both commands keep `MinimumNArgs(0)` and refuse only in RunE — but root pre-run has already run `autoMigrateOnVersionBump`, opened the store (remote-migrate gate), and possibly auto-imported JSONL by then. A non-interactive no-ID invocation can therefore migrate/import (or die with a store error) instead of failing fast with the advertised message. Cobra validates `Args` before `PersistentPreRunE`: putting the conditional no-ID denial in argument validation (with the interactive fallback still resolved in RunE) gives the fail-fast behavior the PR describes. The guard test asserts exit/message but not that the store stayed unopened.

**Should-fix — generated docs hit the release-pin policy** landed after this branched (`docs/cli-docs.pin` → v1.1.0): drop the regenerated `update.md`/`close.md` hunks (and the unrelated backup hunk, which landed independently) and keep source help + changelog.

The proxied handlers already reject missing IDs — that parity is in place.

steveyegge:
Review response pushed (rebased onto main @ 92ad0295f, then e5a5ccb66):

- **Rebase blocker**: resolved; `update.go` long-help keeps the multi-ID partial-failure contract paragraph from b2b153b7b alongside the fallback-gating doc. Changelog entry re-homed under the current Unreleased/Fixed.
- **Fail-fast should-fix**: the non-interactive no-ID refusal now lives in cobra `Args` validation (which runs before `PersistentPreRunE`, cobra v1.10.2 `command.go:968` vs `:985`), so it fires before `autoMigrateOnVersionBump`, store open, or JSONL auto-import. The interactive fallback is still resolved in RunE. New guard subtest `refusal fires before store open` runs the no-ID command in a bare non-workspace dir and asserts the advertised message appears AND the directory stays empty — under the old ordering that case died on a store error.
- **Docs pin should-fix**: dropped the regenerated `docs/cli-reference/{update,close}.md` + `CLI_REFERENCE.md` hunks and the unrelated backup hunk; branch now carries source help + changelog only, per `docs/cli-docs.pin`.

`TestNoIDFallbackGuard` green locally (all six subtests, verified RUN lines, no skips).

steveyegge:
**Fable adversarial review verdict: APPROVE** (fresh-context reviewer, charged to refute; gate per Steve — maphew's re-review not required to land).

maphew-findings ledger at e5a5ccb660: all three 7/24 items genuinely addressed — (1) clean rebase, merge-base = main tip, the b2b153b7b multi-ID partial-failure paragraph preserved in update.go's help, CHANGELOG under current Unreleased; (2) refusal moved into cobra Args validation — verified against cobra v1.10.2 source that ValidateArgs precedes PersistentPreRunE with no OnInitialize hooks registered to sneak side effects earlier, and the new bare-dir subtest proves the fail-fast leaves the store unopened; (3) all four release-pinned docs hunks dropped, interdiff confirms nothing else lost or duplicated across the 647-commit rebase (last_touched.go byte-identical to the reviewed version). Proxied handlers' hard-error parity holds; no in-tree script relied on the argless fallback; the new tests genuinely run in the required PR gate (-race keeps the cgo e2e compiled in; the only skip is TTY-conditional).

**Merge is mechanically blocked by an environmental CI red this PR did not cause:** the 90-day doc-staleness tripwire fires on azure-devops.md / json-schema.md / otel-data-model.md — files this PR never touches; any branch fails it today. Fix belongs on main; this PR re-runs green after. Follow-ups worth filing, not re-review: Args-phase refusal honors --json but not the --format json alias (normalized in pre-run, which now correctly runs later); the BD_LAST_TOUCHED_FALLBACK=1 advice is a dead end in proxied mode; oracle scenario close_no_id_no_lasttouched's pins prose is stale.

PR #4844, opened 2026-07-16
feat(dolt): add `bd dolt rebase` to reconcile colliding child IDs (#4796)

### LINKED ISSUE #4796: `bd dolt pull` unmergeable when two machines create children under the same parent in parallel — colliding child ids + `child_counters` conflict, no `bd dolt rebase` path
### Environment
- `bd` v1.1.0 (Homebrew), embedded Dolt mode
- Two machines syncing a shared beads DB over a `git+ssh` remote via `bd dolt push` / `bd dolt pull` (`refs/dolt/data`)
- macOS (Darwin 25.x)

### Summary
When two machines each run `bd create --parent <epic>` under the **same** parent before syncing, both independently allocate the **same** child id (e.g. `epic.71`) for **different** work. On the next `bd dolt pull` the 3-way merge hits an add/add primary-key collision on `issues` **and** a both-changed conflict on the `child_counters` row for that parent. Neither is auto-resolvable, so the pull aborts and sync is blocked indefinitely. The only recovery today is a heavy manual runbook (SQL renumber + conflict resolution, or make one machine canonical and re-adopt).

This is the **data-level** sibling of the now-closed schema-level fork in #4259. It is distinct from #4750 (archived-suffix recycling, single machine) and #728 (counter-not-backfilled, single machine): here both sides are fully healthy and correctly numbered *locally* — the collision is purely a product of concurrent offline creation under a shared parent.

### Steps to reproduce
1. Two clones A and B of a beads DB syncing via `bd dolt push`/`pull`, both current, sharing a parent epic `pa-vubk` whose `child_counters.last_child = 70`.
2. Without syncing between them:
   - On A: `bd create --parent pa-vubk ...` twice → `pa-vubk.71`, `pa-vubk.72` (A's work); `last_child` → 72.
   - On B: `bd create --parent pa-vubk ...` three times → `pa-vubk.71`, `pa-vubk.72`, `pa-vubk.73` (B's *different* work); `last_child` → 73.
3. A pushes first (becomes canonical). B runs `bd dolt pull`.

### Expected
Either sequential child numbering never blocks reconciliation, or `bd` offers a first-class command to resolve it (detect the colliding children, renumber the non-dominant side, fix references, complete the merge).

### Actual
The pull aborts. Two separate unresolvable conflicts are involved:

1. **`issues` PK add/add collision.** `pa-vubk.71` and `.72` exist on both sides with different content. Dolt records these in `dolt_conflicts_issues`; beads' auto-resolver (`internal/storage/versioncontrolops/mergesettle.go`) only auto-resolves `metadata`, audit-only `dependencies`, and `schema_migrations` — never `issues` — so the merge aborts.
2. **`child_counters` both-changed conflict.** After the colliding issue rows are renumbered away, the pull still aborts on the counter row:
   ```
   Error: merge origin/main: merge conflicts in child_counters require operator resolution; merge aborted and working set restored
   ```
   (base `last_child=70`, ours `=73`, theirs `=72` — a genuine both-changed cell that beads does not auto-resolve to the max.)

### Root cause
`internal/storage/issueops/child_id.go` (`GetNextChildIDTx`) derives the next child number from `child_counters.last_child` plus a scan of existing children on the **local** clone only. There is no cross-machine coordination and no merge-time reconciliation, so two offline clones deterministically mint the same ids. The child id *is* the parent/child relationship (there is no `parent_id` column), so a collision is a hard PK conflict rather than a soft one.

### Proposed solution (I've been asked to build/propose a fix and am happy to contribute a PR)
A `bd dolt rebase` command (naming it after `git rebase`, since it replays the losing side's ids onto the dominant history) with two mutually exclusive flags:

- `--local-dominates` — on collision, local ids win; the conflicting **remote** children are renumbered.
- `--remote-dominates` — the reverse; the conflicting **local** children are renumbered.

Behaviour:
1. Detect child-id PK collisions between the two heads and the both-changed `child_counters` rows.
2. Renumber the non-dominant side's colliding children to the next free numbers **above the max of both counters** (so a local-only non-colliding child like `.73` is never clobbered).
3. Rewrite references. Notably, the FK constraints already carry `ON UPDATE CASCADE` (`dependencies.issue_id`/`depends_on_issue_id`, `labels`, `comments`, `events`, `issue_snapshots`), so renaming `issues.id` propagates automatically; freetext id mentions in `description`/`notes` are reported for manual review rather than blindly rewritten.
4. Bump `child_counters` to the true high-water mark and complete the merge (resolving the counter conflict to the max).
5. Backup-first, idempotent, and prints exactly what was renumbered (old id → new id).

I verified this algorithm by hand against a live occurrence today (two machines both created `pa-vubk.71`/`.72`; renumbering the local pair to `.74`/`.75`, keeping the non-colliding `.73`, then resolving the counter to `75`, produced a clean merge). Happy to turn that into `bd dolt rebase` if the maintainers are open to it.

### Related
- #4259 (schema-level PK fork — this is the data-level counterpart)
- #4750 (child_counters not a true high-water mark)
- #728 (colliding child ids when counter not backfilled)
- #4640 (orphan child-counter on import)

rjc123:
Closes #4796.

## Summary

Adds `bd dolt rebase [--remote-dominates | --local-dominates]`, which reconciles cross-clone **hierarchical child-ID collisions** that a plain `bd dolt pull` cannot auto-merge.

When two clones each add a child under the same parent while offline, both mint the same `parent.N` id (the child counter is purely local). On sync this is an **add/add PK collision on `issues`** *and* a **both-changed `child_counters` row** — neither of which the pull settle auto-resolves, so the pull aborts and the operator is left with a manual re-ID runbook. This is the data-level sibling of the #4259 schema-level PK fork.

## Behaviour

`bd dolt rebase` fetches the remote, renumbers the losing side's colliding children (and their subtrees) to free ids, resolves the `child_counters` row to the true high-water mark, and completes the merge. A backup tag is written before any mutation; the push is left to the user.

- `--remote-dominates` (default): the remote keeps its ids (it's the shared copy other clones have already pulled), your local colliding children are renumbered.
- `--local-dominates`: your local ids are kept, the remote's colliding children are renumbered.

## Design decisions worth a reviewer's eye

- **Dependency rekey.** The renumber routes through the canonical rename primitive (`issueops.UpdateIssueIDInTx`) rather than a bare `UPDATE issues`, so the derived-PK dependency edges (`depid.New(issue_id, target)`, #4259) are rekeyed. An FK cascade alone updates the edge's columns but leaves the surrogate key stale, which re-forks it on the next merge.
- **Local-branch only.** Both directions run entirely on the local working branch. A checked-out remote ref cannot be used, because the `wisp_*` tables are session-local working-set tables that are never committed to history — resetting the working set to a committed ref loses them, and the rename primitive then fails. `--local-dominates` therefore renumbers locally (the remote-dominates result) and then **swaps** each contested pair so the local row reclaims the id.
- **child_counters.** The both-changed counter conflict is resolved to the true high-water mark — the half the standard settle deliberately refuses.

## Testing

- Real embedded-Dolt integration tests reproducing a two-clone `.71/.72/.73` collision in **both** directions, asserting: winning-side ids kept, losing-side renumbered, FK-cascaded dependents and rekeyed dependency edges follow the renumber, `child_counters` at the true max, no residual conflicts, clean working set.
- Unit tests for collision-root detection and ref validation.
- Manually exercised end-to-end with the built binary against a two-clone `file://` collision in both directions, each confirming a clean `pull` **and** `push` round-trip afterwards.

## Note on the root cause (deliberately *not* addressed here)

This is a pragmatic unblock, not the root-cause fix. The deeper issue is that the visible `parent.N` numbering is being used as the unique key. The durable identifier should be **stable and unique**, while the visible hierarchical number is **derived dynamically from parentage**. That is a larger schema change worth tracking separately; this command unblocks sync today without it.

Happy to adjust naming, flag semantics, or split the change if that suits the project's direction better.


rjc123:
A small DX suggestion on the `Check doc flags freshness` gate, prompted by this PR.

The check already regenerates the CLI reference at both HEAD and the merge-base to detect drift, and it writes the exact corrective diff to `cli-docs-freshness.patch` (uploaded as an artifact). In other words, the job **builds the correct docs in order to prove the committed ones are stale** — and the generated reference is fully deterministic from the command tree. Given that, it could just as easily *produce* the docs (push the regenerated files to the PR branch, or have a bot apply the patch) instead of red-failing and asking the contributor to regenerate locally.

Concretely: this PR tripped the gate not because of its own change, but because the branch was one commit behind `main`. The merge-base regeneration picked up a small pre-existing `bd gate` help drift that has nothing to do with `bd dolt rebase`. A rebase + regenerate fixed it, but the failure is confusing when the drift isn't yours — and it's odd to be asked to hand-run a step the CI just ran itself.

Not blocking, and I've folded the regeneration into this PR either way — just flagging it as a possible workflow improvement (auto-commit the regenerated docs, or at least surface the freshness artifact as a one-click suggested commit).

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace through the pinned Dolt engine; Claude adjudication spot-verified at the PR head). Verdict: **needs changes** — `bd dolt rebase` addresses a real pain (#4796 child-ID collisions), but the current machinery can lose or corrupt data in the failure paths.

**Blocker 1 — mixed wisp subtrees aren't renumbered.** Descendant enumeration reads only `issues`, so hierarchical wisp children under a renumbered durable root keep their `oldRoot.*` IDs — identity and hierarchy diverge. (Wisp children under durable parents are an explicitly supported shape.)

**Blocker 2 — unresolved constraint violations can be committed.** The rebase enables `dolt_force_transaction_commit`, then discards `TryRepairFKCascadeViolations`' repaired-result before committing (verified: the first return value is dropped). The sibling merge-settle path does this right — it rejects `hadViol && !repairedViol` — and the pinned Dolt confirms forced commits accept outstanding violations. Same guard needed here.

**Blocker 3 — a failed rebase can leave committed renumbering behind, with recovery guidance that doesn't exist.** Renumbering commits *before* the merge attempt; `abortMerge` restores the pre-merge state, which already contains that commit. On error the CLI discards the backup report, prints the recovery tag only on success, and recommends `bd dolt reset --hard` — a command that doesn't exist (the only reset is the unrelated destructive `bd admin reset`). Either stage renumbering and merge in one commit boundary, or make the error path print the backup tag and a real recovery sequence.

**Should-fixes:** the `parent.1000000+i` temporary IDs aren't checked for existence (a legal occupant makes the swap fail mid-flight, triggering blocker 3); the "two clones + pull/push convergence" verification claim isn't in the tests (both call `RebaseChildCollisions` directly on a peer branch — no `RebaseRemote`, transport, or convergence coverage); the command ignores `--json` and has no CLI tests; recovery-command policy routes fixes through `bd doctor --fix` — integrate or document the exception. Plus a rebase over the backend-removal drift (modify/delete conflicts on the deleted SQL-backend stubs) and dropping the unrelated gate-doc hunk.

rjc123:
Thanks @maphew — all three data-integrity blockers are addressed, and the branch is rebased onto current `main` (it was well behind, which also resolves the SQL-backend-removal drift and the docs-pin). Squashed to one clean commit.

1. **Wisp subtrees** — `subtreeIDs` now enumerates both `issues` and the session-local `wisps` table, and the rename loads each row from whichever table owns it. Added an embedded integration test for the wisp-under-durable shape (it orphans the wisp child without the fix).
2. **FK violations** — we now capture the repaired result and guard `hadViol && !repairedViol` with an abort, matching the pull-settle path; you're right that `dolt_force_transaction_commit=1` defeats the commit-gate we were relying on.
3. **Atomicity** — the mechanic is now atomic: any failure hard-resets to the pre-rebase backup tag, so a failed merge can't strand the renumber commit. Dropped the incorrect `bd dolt reset --hard` guidance (that command doesn't exist) in favour of reporting the automatic restore.

Also dropped the unrelated gate-doc hunk and the CLI-reference additions (they regenerate at the next release under the docs pin). CI docs-freshness passes locally.

rjc123:
The three red checks look like infra flakes rather than anything in this change — all three are in code paths this PR doesn't touch:

- **Storage backend conformance (embedded Dolt oracle)** — a timeout, not an assertion: `panic: test timed out after 10m0s` with `TestConformance` alone having run the full `10m0s` (so nothing new in this PR even executed before it fired). The run came in at `600.082s` — 0.08s over the cap. I ran that same suite locally on this branch's HEAD and it passes in **529s (8m49s)**, and the conformance suite is backend-agnostic — this change doesn't alter the CRUD paths it exercises. Looks like the job is simply sitting right on the 10-minute ceiling ("each test builds an isolated Dolt-backed workspace") and tipped over on a slow runner.
- **Test (Proxied Dolt Cmd 1/15)** — `TestProxiedServerFindDuplicates` fails at `bd init --proxied-server` with `uow: migrate: rekey dependency ids: wisp_dependencies: Error 1105: no root value found in session`. That migration is in `internal/storage/schema/schema.go` (untouched here, landed in #4259), and "no root value found in session" reads as a Dolt proxied-server session flake.
- **CI Gate / Required** — the aggregator, red only because of the two above.

Could you re-run the failed jobs when you get a chance? I expect them to go green.

Separately: that conformance job running ~9–10 min against a hard 10-minute timeout is a latent flake for any contributor, not just this PR. Happy to open a separate issue for it (shard the job by a `go test -list` partition, or bump the timeout) if that'd be useful — didn't want to fold an unrelated CI change into this PR.

maphew:
Interim status so this doesn't look stalled — thanks for the substantial rework.

Your CI triage checks out where I could verify it: the `Storage backend conformance` failure is the suite hitting the job's 10-minute ceiling (`panic: test timed out after 10m0s`, `600.082s` — 0.08s over), not an assertion, and the `Proxied Dolt Cmd 1/15` error you quoted (`no root value found in session` during a #4259 migration this PR doesn't touch) reads as a known-shape proxied-session flake. One caveat worth keeping in mind: this PR adds a 362-line embedded-Dolt integration test to the same package the conformance job runs, so if that job executes the whole package rather than `-run TestConformance` alone, the branch does nudge an already-at-the-ceiling job closer to it — the shard-or-bump issue you offered to file is worth filing either way.

On the review itself: your response addresses all three data-integrity blockers head-on (wisp-aware subtree enumeration with a new orphan-reproducing test, the `hadViol && !repairedViol` abort matching the pull-settle path, and atomic hard-reset-to-backup on any failure — plus dropping the nonexistent `bd dolt reset --hard` guidance). Because those three sit squarely on storage correctness and Dolt failure-path semantics, the re-verification is getting a dedicated deep review pass rather than a drive-by in this sweep — same rigor as the first round, on current main. That is queued now; expected within a day or two, and the failed jobs will get a rerun alongside it.

Not asking anything further of you in the meantime.

maphew (commented):
Deep re-review at head d484d2e (two independent review passes, cross-vendor, with empirical probes against real embedded Dolt run in a scratch copy). The design is sound and the three fixes from the previous round are real work — wisp-aware enumeration, abort parity, and the backup-tag defer are all genuinely there. But two of the fixes are incomplete in ways we confirmed by running the code, not just reading it, and both are in this PR's own named risk class: silent data corruption on a sync failure path. Details below.

## Blocker 1 — free-ID selection is wisp-blind: rebase mints an ID a session-local wisp already owns

`maxChildNumber` (rebase.go:422-442) scans `issues` only; `renumberCollisions` seeds the free-number high-water solely from that scan and never consults `wisps` or `child_counters`. A session-local wisp legitimately occupies `parent.N` without appearing in `issues` (`GetNextChildIDTx` numbers wisp children of durable parents off the shared space). Probe (durable parent, wisp child at `.74`, remote-dominates rebase):

```
DUAL PRESENCE: xt-p.74 exists in BOTH issues ("Local71") and wisps ("WispSibling")
```

...and the rebase returned nil — "Rebase complete" over a store that now hard-errors on `bd ready`/`bd search` (the #4455 dual-presence class this command exists to clean up). Same blind spot at rebase.go:562, where `resolveChildCountersToHighWater` also uses a bare `SET last_child = ?` (the sibling path at :294-297 correctly uses `GREATEST`), so the counter can be *lowered* below an existing wisp child, re-arming duplicate minting.

Fix shape: `maxChildNumber` takes max over `issues` AND `wisps` (working set), seed from `GREATEST(scan, child_counters.last_child)` on all three refs, `GREATEST` in the :566 update.

## Blocker 2 — the atomic restore strands renamed wisps; CLI then reports a clean rollback

`DOLT_RESET('--hard', <tag>)` in the defer (rebase.go:121-129) only reverts *tracked* tables; `wisps`/`wisp_*` are `dolt_ignore`'d. End-to-end probe driving the real failure path (unresolvable `config` conflict):

```
AFTER FAILED REBASE: issues ap-x.71=1 ap-x.74=0 | wisps ap-x.71.1=0 ap-x.74.1=1
```

The durable child is back at `.71`; its wisp child is stranded at `.74.1` — orphaned under an ID that no longer exists, which is the exact hazard the wisp-enumeration fix was added to prevent, reintroduced on the failure path *by* that fix. Worse: `updateWispIDInTx` rewrites both `wisp_dependencies` (untracked, survives) and `dependencies` (tracked, reverted), so the two dependency tables end mutually inconsistent. And cmd/bd/dolt.go:635-641 prints "No changes were kept" over it.

Fix shape (preference order): (1) snapshot affected wisp rows pre-tx and restore them in the defer; (2) defer the wisp renames until after merge+swap succeed, so the failure path has no wisp mutation to undo (`renameSubtreeInTx` already separates the tables — a reordering, not a rewrite); (3) minimum: refuse to start with wisps inside a colliding subtree and fix the :114-120 contract comment + CLI message.

## Blocker 3 — LIKE wildcards in subtree root IDs are unescaped

rebase.go:333-334 binds the root into a `LIKE CONCAT(?, '.%')` pattern. Explicitly-assigned IDs may contain `_`/`%` (current ID validation permits them), so a collision under `bd-a_b.71` can sweep and rename descendants of `bd-axb.71` — silent relocation of unrelated data. Escape wildcards with an `ESCAPE` clause or use a prefix-safe comparison.

## Should-fixes

- **Restore message wrong in both directions:** rebase.go:141,:145 return `nil, err` after `BackupTag` is set, so on renumber failure the operator is never told the tag exists; conversely store.go:3160/version_control.go:465 return `report, err` when only `recomputeBlockedAfterPull` failed — rebase committed, CLI still prints "restored to pre-rebase state". Add `Restored bool` to `RebaseReport`, set only by the defer; gate the message on it; return `report, err` at :141/:145.
- **Deferred reset runs on the caller's context** (rebase.go:125): a canceled API context means the rollback itself can't execute. Use a bounded detached context for cleanup.
- **Merge-error handling diverges from pull-settle** (rebase.go:461-468): bails immediately where `SettleMerge` deliberately carries `mergeErr` through the settle pass because some Dolt versions error on conflicts. On such builds `bd dolt rebase` could never complete (the `child_counters` conflict is always present). Route through the same shape.
- **Temp ID `parent.1000000+i` has no occupancy check** (rebase.go:401-413): a bystander at `parent.1000000.5` gets silently swept into the swapped subtree (probe-confirmed). Assert `temp`/`temp.%` free in issues+wisps, or allocate from the same free-number pool.
- **Counter delete/modify conflict no-ops silently** (rebase.go:531-578): our-side deletion → `UPDATE` affects 0 rows, `--ours` keeps the deletion, counter gone. Upsert like :294-297 + check `RowsAffected`. Also :537's `strings.Contains(err.Error(), ...)` swallows real errors — use `dberrors.IsTableNotExist`.
- **Header comment claims `content_hash` is nulled on renumber** (rebase.go:47-51) — nothing does that (and it's unnecessary: `ComputeContentHash` excludes ID). On a corruption-adjacent path, an asserted-but-absent safety measure is worse than none. Delete the paragraph.
- **Coverage gaps from the previous round remain:** no test exercises `RebaseRemote` on either store; no two-clone convergence test; `--json` prints prose (report has json tags, CLI never checks `jsonOutput`); registration tests not extended; and nothing routes an operator here — the pull-failure path never mentions `bd dolt rebase`. A one-line hint there is the minimum for discoverability.
- Nits: `CountersSet` map iteration makes report output non-deterministic (sort keys); `--dry-run` would be cheap and worth it for an operator recovery command; backup tags accumulate with no documented cleanup.

CI-cost note: the new tests land in `test-embedded-storage` (~4s wall, negligible), not the at-ceiling conformance shards — no concern there.

## Verdict

Not mergeable at this head — blockers 1 and 2 are confirmed corruption on the command's own recovery path. But the deltas are small relative to the value and the invariants that matter (derived-PK dep rekey via `issueops.UpdateIssueIDInTx`, sorted collision processing, backup-tag mechanics) are genuinely satisfied. Want to take these? If you'd rather we absorb them, we're happy to land maintainer-side fixes on top of your branch with your commits intact — say the word.

_claude-fable-5-high on behalf of maphew_ (findings merged from two independent review agents, claude-opus + gpt-5.6-sol, probes verified on embedded Dolt)

maphew (changes requested):
Third-round deep review at head d484d2e7b, and first the good news: **all three blockers from the last round are genuinely fixed**, and we verified each mechanically rather than taking the diff's word for it.

- *Wisp-aware enumeration*: `subtreeIDs`/`subtreeIDsFromTable` cover both `issues` and `wisps` with clean table-not-exist degradation, `renameSubtreeInTx` picks the source table per id via `IsActiveWispInTx`, and the new test's wisp-child assertions would fail without the fix. Real coverage, not decorative.
- *Abort parity*: the `hadViol && !repairedViol` path now mirrors `mergesettle.go` faithfully, and your added `GetConflicts` check is actually stricter than `SettleMerge`. Session-flag hygiene checks out in both callers (fresh conn per call, nothing leaks to the pool).
- *Atomicity*: we traced every post-tag return; the named-`retErr` defer catches all of them. One correction to your own comment though — `DOLT_RESET('--hard', tag)` does *not* destroy session wisps: Dolt's `resetHardTables` carries working-but-not-staged tables onto the target root, and the wisp tables are `dolt_ignore`d and never staged. The stay-on-local-branch design is still right; the stated justification just isn't, and it's worth rewording before someone "fixes" it the wrong way.

Also confirmed the #4259 invariant holds: every rename routes through `issueops.UpdateIssueIDInTx`, both directions of the derived dependency PK are rekeyed, and `assertDepRekeyed` pins it on both sides.

What blocks this round is narrow — two functions, a handful of lines, but the failure class is the one this command exists to repair:

1. **`maxChildNumber` seeds from an issues-only scan** (its comment says it mirrors `GetNextChildIDTx`, but that primitive seeds from `child_counters.last_child` first, then maxes with the scan). Two concrete failures: a deleted child (`bd delete` never lowers the counter) makes the renumberer re-mint an id a peer may still hold — precisely the #4796 hazard; and a session wisp under a durable parent consumes a `P.N` slot without appearing in `issues`, so the renumberer can mint `issues.P.N` colliding with `wisps.P.N` — no PK conflict fires (different tables), and from then on `IsActiveWispInTx` routes every durable operation on that id to the wisps table. Fix: seed per parent from `GREATEST(max(issues), max(wisps), child_counters.last_child)`, and read `child_counters AS OF ref` for the remote refs too.

2. **`resolveChildCountersToHighWater` writes a live-max, not a high-water mark**: the plain `UPDATE ... SET last_child = ?` can move the counter *down* for any parent whose highest children were deleted or are wisps, after which the next `bd create --parent` re-mints an id a peer already holds. Cheap fix, fully local: you already query `dolt_conflicts_child_counters` for these parents — select `our/their/base_last_child` in that same query and write `GREATEST(trueMax, our, their, base)`. That also closes the ordering issue where the function runs before conflict auto-resolve and so can't see a theirs-only row.

An extended fixture with a session-wisp sibling plus a deleted-then-renumbered child would pin both. If you would rather stop here after three rounds — completely understandable — say the word and we will absorb these two fixes maintainer-side with your authorship preserved; the PR is otherwise ready and we want it.

Worth knowing, not blocking (follow-up material, we are filing beads for these on our side):

- Renaming a wisp orphans its `wisp_labels`/`wisp_comments`/`wisp_events` rows — but that is a pre-existing gap in `updateWispIDInTx` (only `wisp_dependencies` has an FK), not your doing; your PR is just the first bulk caller. We are splitting it into its own issue.
- `detectCollisions` treats identical add/add rows (same id, same content via out-of-band import) as collisions and forks one issue into two — a content-equality guard would skip those.
- The temp-id `parent.1000000+i` occupancy probe from last round is still open, and each swap now writes `renamed` audit events to and from the fake temp id.
- `RebaseRemote` on the raw `*sql.DB` relies on `SetMaxOpenConns(1)` for session pinning; `db.Conn(ctx)` would make it a guarantee, matching `dolt/versioned.go`.
- `doltRebaseCmd` never consults `jsonOutput` even though `RebaseReport` carries full json tags; the two early-failure paths return a nil report so the "restored to backup tag" message is suppressed exactly when it matters; and the `CountersSet` map iteration order is nondeterministic.
- Adding `RebaseRemote` to `storage.RemoteStore` widens the exported `beads.RemoteStore` alias — a break for out-of-tree implementers; a narrow capability interface (the codebase's own documented pattern) is cheap now and expensive later.
- Small doc nits: the `content_hash`-set-NULL paragraph describes something the code doesn't do (and renames don't invalidate the hash anyway — it excludes the id); the test-file header points at `internal/storage/dolt` but the tests live in `embeddeddolt`; the wisp "convergence" second pass is vacuous (after the first merge, `DOLT_MERGE_BASE('HEAD', peer)` *is* peer's head, so the collision set is empty by construction).

Three rounds in, this is high-quality, responsive work on a genuinely hard primitive — thank you for staying with it. Two small functions and it lands.

rjc123:
Thanks for the thorough round-3 pass — both blockers are fixed in b0958bdf6.

- **`maxChildNumber` is now a true high-water**: `GREATEST(max(issues), max(wisps), child_counters.last_child)`, reading the counter and — on the working set only — the session-local `wisps`/`wisp_child_counters`, with `AS OF` applied only to the durable side (wisps are never at a committed ref). So the renumberer no longer re-mints a deleted-child slot the counter still reserves, nor mints `issues.P.N` over an existing `wisps.P.N`.
- **`resolveChildCountersToHighWater`** now selects `our/their/base_last_child` in the same conflict query and writes `GREATEST(trueMax, our, their, base)` via a never-lower `INSERT … ON DUPLICATE KEY UPDATE last_child = GREATEST(last_child, VALUES(last_child))` upsert (mirroring `renumberCollisions`), which also lands the theirs-only row you flagged.
- **New embedded-Dolt regression** `TestEmbeddedRebaseRemoteDominatesHighWaterSlots` — a fixture with a deleted highest child *and* a session wisp sibling. Verified it bites: without the fix it re-mints `.73` and collides on `.74`; with it, the locals renumber clear to `.75/.76` and the counter holds at the high-water.
- **Comment reworded**: you're right that `DOLT_RESET '--hard'` carries the dolt-ignored wisps onto the target root — the restore defer relies on exactly that. The stay-on-local-branch rationale now says what it actually is (keep the rename on the live wisp-bearing working set), not the incorrect "a reset would drop them".

One heads-up on mechanics: the branch has drifted behind `main` since your 25 Jul merge (unrelated churn in `cmd/bd/dolt.go` now conflicts with our `bd dolt rebase` wiring). Happy to merge `main` in and resolve it, or leave it to you as before — whichever you prefer.

maphew:
Round-3 verification: both fixes check out, and I traced each claim rather than taking the summary.

**`maxChildNumber` is a true high-water.** `GREATEST(max direct child in issues, child_counters.last_child)` is the right seed, and it does mirror `issueops.GetNextChildIDTx` — that function reads the counter first and only then raises it from live rows (`internal/storage/issueops/child_id.go:17-48`), so a renumberer that read only live rows was strictly weaker than the minter it has to stay clear of. The deleted-child hazard is real and now covered.

**`resolveChildCountersToHighWater`.** Folding `our/their/base_last_child` into the mark is correct and the `GREATEST(...)` upsert makes it monotone. The theirs-only case you flagged does need the `INSERT`: a plain `UPDATE` has no working-set row to match.

**The new regression test bites.** I confirmed the direction it fails in without the fix (`.73/.74` instead of `.75/.76`), and `assertRebaseSettled` plus the dep-rekey assertion still cover the pre-existing invariants.

## One thing the theirs-only `INSERT` introduced — fixed and pushed

`child_counters.parent_id` carries `CONSTRAINT fk_counter_parent FOREIGN KEY (parent_id) REFERENCES issues(id) ON DELETE CASCADE` (`internal/storage/schema/migrations/0008_create_child_counters.up.sql:4`). So a counter row cannot exist without its parent, and this shape reaches the new `INSERT`:

- our side deletes parent `P` — the cascade removes our `child_counters` row;
- their side bumps `P`'s counter;
- the merge leaves `P` deleted, `dolt_conflicts_child_counters` has a theirs-only row, `COALESCE(our_parent_id, their_parent_id, base_parent_id)` yields `P`;
- the `INSERT` tries to create `child_counters(P, hw)` with no `issues` row to reference.

I didn't want to assert this from reading alone, since `@@dolt_force_transaction_commit = 1` is set just above and might have suppressed the check — so I probed the real embedded engine with both session flags the rebase path sets:

```
UPDATE form: no error, RowsAffected=0
INSERT form: Error 1452: cannot add or update a child row -
             Foreign key violation on fk: `fk_counter_parent`,
             table: `child_counters`, referenced table: `issues`,
             key: `[no-such-parent]`
```

So `force_transaction_commit` only stops `DOLT_COMMIT` refusing an already-violated working set; it does not suppress statement-level FK enforcement. The old `UPDATE` was accidentally correct here — zero rows matched, and `DOLT_CONFLICTS_RESOLVE('--ours')` then kept our deletion. The new form makes it an error that fails the whole rebase and hard-resets to the backup tag. Safe, but a working reconcile becomes an aborted one.

Worth noting the loop walks **every** row in `dolt_conflicts_child_counters`, not just the parents in the collision that prompted the rebase — so an unrelated deleted-parent-vs-bumped-counter elsewhere in the same merge is enough to trip it.

The canonical bulk path already guards exactly this, and says why:

> Orphaned hierarchical IDs are valid import input when the parent was deleted before export. Their auxiliary counter has no owner and must not be inserted: both counter tables enforce a parent foreign key.
> — `internal/storage/issueops/create.go:931-933`

Pushed as `bd50374bf` rather than sending you a round 4 for seventeen lines: skip a conflict row whose parent is absent from `issues`, and let `--ours` keep the deletion. Revert or force-push over it if you'd rather shape it differently. One piece I deliberately left with you: I added no test for it, and you have the fixture harness in hand — a theirs-only counter conflict over a locally-deleted parent is the case to pin.

The commit before it, `da22574b4`, merges current `main` and resolves the `cmd/bd/dolt.go` conflict you flagged — both sides had only added a flag registration in the same `init()` block, so it keeps both.

Local validation on the pushed head: `go build`, `go vet`, `golangci-lint run` (0 issues), `go test ./internal/storage/versioncontrolops`, and all four `TestEmbeddedRebase*` integration tests pass.

## Smaller notes, none blocking

- **`wisp_child_counters` in `maxChildNumber` looks unreachable.** Both callers pass a durable parent: `detectCollisions` queries `issues`, and the conflict loop's parents come from `child_counters`, whose FK targets `issues`. Since `GetNextChildIDTx` and the bulk path in `issueops/create.go:920-928` both route on whether the *parent* is a wisp, a durable parent's counter always lives in `child_counters` — so that lookup should always return `sql.ErrNoRows`. If it's deliberate cover for a promoted wisp whose `wisp_child_counters` row outlived the promotion (`internal/storage/schema/cli_migrations.go:243-250` migrates and `:293` deletes those rows, so a survivor is conceivable), the comment is the place to say so. The `wisps` *table* scan beside it is a different matter and I do think it earns its place.
- **The new fixture's state isn't one the create paths can produce.** `insertRebaseWisp` does a bare `INSERT INTO wisps` with no counter bump, so the test sits at counter 73 with a wisp at `.74`. In production that wisp is minted through `GetNextChildIDTx` with a *durable* parent, which routes to `child_counters` and leaves the counter at 74 — where the counter read alone catches it. That makes the `wisps` scan defense-in-depth rather than the thing this fixture proves, which is fine, but the comment currently reads as a reproduced production hazard. Either bump the counter in the fixture to model reality (and keep a separate synthetic case for the cross-table guard), or say "synthetic: counter deliberately left behind the wisp".
- **`err == sql.ErrNoRows`** in `lastChildCounter` matches the surrounding code (`child_id.go:20` does the same), so I've left it alone — flagging only because `errors.Is` is the form that survives a driver that wraps.
- **Heads-up on a collision coming the other way:** #5146 introduces a `DrainCall` helper for tx-pinned `CALL DOLT_*` sites. `rebase.go` adds several (`DOLT_COMMIT`, `DOLT_RESET`, `DOLT_MERGE`, `DOLT_CONFLICTS_RESOLVE`), so whichever lands second gets a small mechanical follow-up. Nothing to do here.
- **CI.** Base `main` is red right now for unrelated reasons, so this PR's checks aren't a clean signal until that clears; I'll rerun once it does rather than reading through the noise.

On the conformance-timeout flake you offered to file separately — please do, and thanks for keeping it out of this PR. Your read on it stands up: that job sits at ~9–10 min against a hard 10-minute cap, and this branch adds a 362-line embedded-Dolt test to the same package, which nudges it closer.

rjc123:
Pushed `134565c33`. This addresses the test you left us and the two comment corrections, and — rather than another reactive round — we ran an independent review of `rebase.go` and **probed each candidate finding on the real embedded engine before acting**. Two of them turned out to be non-issues, and it seemed more useful to show the probes than to "fix" them:

**The orphaned-counter test you left us — added, and it bites.**
`TestEmbeddedRebaseRemoteDominatesOrphanedCounterConflict`: the P.71/.72 collision that triggers the rebase, plus an unrelated parent Q that our side deletes and the peer bumps the counter for — so the settle walks a theirs-only `dolt_conflicts_child_counters(Q)` whose parent this merge leaves deleted. With your guard removed it reproduces the exact `Error 1452 fk_counter_parent` abort you measured; with it, the rebase completes and `--ours` keeps our deletion.

**Confirmed the FK is actually present — your guard's rationale holds.**
Worth flagging since it cuts against a reading of the migrations: `0039_drop_child_counters_fk` exists, but on the migrated embedded schema the constraint is still there (`INFORMATION_SCHEMA` shows `fk_counter_parent = 1`; DELETE-parent cascades the counter to 0; the orphan INSERT errors). So the guard is load-bearing, not belt-and-braces, and your `bd50374bf` reasoning is correct as written.

**Durable subtree-parent counter — checked, and it's fine (not the gap it looked like).**
A colliding child that is itself a parent with its own `child_counters` row: I expected the rename to strand that counter, but on the real engine renaming `.71 → .74` carries both the durable grandchild `.71.1 → .74.1` **and** its counter (the FK cascades the id change; no orphan at `.71`, a live counter at `.74`). Added `TestEmbeddedRebaseRemoteDominatesDurableSubtreeCounter` to pin that, since the existing wisp-subtree test only moved a wisp grandchild and never asserted a counter.

**Fix — nil report on the restore path.**
`RebaseChildCollisions` returned `nil, err` on the two error paths *after* the backup tag is created. The defer had already hard-reset to the tag, but the CLI only prints "restored to backup" when the report is non-nil — so the reassurance was suppressed exactly when a real restore happened (e.g. a commit-renumber failure after the renumber tx committed rows). Now returns the report.

**Comment corrections you flagged.**
`content_hash`-set-NULL header removed (the rename primitive never touches it, and `ComputeContentHash` excludes the id); `maxChildNumber`'s wisp reads now carry their real defense-in-depth rationale; `errors.Is` at the two `ErrNoRows` sites; high-water fixture labelled synthetic; the "convergence" re-run reworded to the weak idempotency invariant it is; test header pointed at `embeddeddolt`.

Left as follow-ups on your side as agreed: json output, the narrow capability interface vs widening `RemoteStore`, the content-equality guard in `detectCollisions`, and the temp-id occupancy/audit-noise in the swap.

Local gate on the pushed head: `go vet`, all six `TestEmbeddedRebase*`, the `versioncontrolops` unit tests, and `make ci-pr-lint` (rebase files clean) pass.

maphew (commented):
Reviewed at `134565c3`.

The data-integrity fixes through this head hold and CI is green. Three maintainer-side fixes remain before merge:

- Allocate and vacancy-check the swap temporary prefix; `parent.1000000+i` can capture an unrelated existing subtree.
- Distinguish a restored failure from a successful rebase whose post-merge `is_blocked` recompute failed, since the latter currently prints a false rollback message.
- Treat identical add/add content as already reconciled instead of duplicating it.

No open PR implements the same end-to-end `bd dolt rebase` operation; adjacent Dolt migration and pull work should remain separate. We should absorb these fixes on this contributor branch with its commits intact.

rjc123:
Pushed `c0e6224dc`. You offered to absorb the remaining three maintainer-side; we've done them on the branch instead, plus the capability interface. Take or leave any of it.

**Branch is level with `main` again.** Merged `upstream/main` in (`29419af8f`) rather than rebasing, since you own commits on this head — 0 behind, no conflicts.

**1. Swap scratch id is allocated and vacancy-checked.** `parent.1000000+i` was only ever improbable, never verified free. Since `renameSubtreeInTx` moves a root *and* everything under `root.`, the case that bites isn't an occupied scratch id — it's an occupied scratch *subtree*: `parent.1000000.1` with no `parent.1000000` row of its own passes a bare row-existence probe, then gets dragged to the remote's new id by the final rename. `allocateSwapTemp` probes upward and takes the first candidate where `subtreeIDs` finds nothing in `issues` **or** `wisps`, marking each taken so pairs can't share. Exhausting the range is a hard error, never a fall-back onto an occupied id.

**2. Identical add/adds are left to the merge.** Same id with the same `content_hash` on both sides is one issue that reached two clones out of band, not two contesting a slot — and renumbering forks it irreversibly, where nothing downstream can tell the halves were ever one row. The test is deliberately one-sided: a candidate is only skipped when *both* sides carry a non-empty hash and they agree, so an absent hash falls through to the existing behaviour. Renumbering a genuinely-identical pair costs a spurious fork; not renumbering a genuinely-different pair costs an issue. Only the first is recoverable by hand.

One subtlety worth calling out, because it's the part that would silently rot: the guard runs **before** the collision-root filter, not after. That filter suppresses a colliding id whose ancestor also collides, on the grounds the ancestor's subtree rewrite will carry it — but if the ancestor has just been excluded as identical, no rewrite is coming, and the descendant is dropped on the floor. There's a test for exactly that shape; with the guard moved after the filter it reports `report.Renumbered has 0 entries, want 1`.

**3. `RebaseReport.Restored` separates a rollback from a kept rebase.** The reconcile is atomic and hard-resets to the backup tag on its own failure, but a failure *after* it returns — the store wrapper's `is_blocked` recompute — leaves the merge committed and standing. The CLI printed "the database was restored… no changes were kept" for both, sending the operator hunting for changes that were live. The flag is set by the restore defer itself; the CLI prints the rollback message only for a real rollback, and the other path says the rebase was kept, lists the renumbers, and points at `bd doctor --fix` for the stale blocked state.

**Capability interface, per your round-3 note.** `RebaseRemote` is off `RemoteStore` and onto `ChildCollisionRebaser`, a single-method optional interface alongside `Flattener`/`GarbageCollector`/`Compactor`, reached from `cmd/bd` through `storage.UnwrapStore` like the other cmd/bd optional interfaces — the category `beads.go` already documents. Net effect, and the easiest thing to check:

```
git diff main...HEAD -- internal/storage/remote.go beads.go backend/backend.go
→ (empty)
```

This PR no longer touches the public surface at all.

That also resolved a merge-surfaced failure: syncing to `main` lit up `TestPublicSurfaceComplete`, which landed since our last sync — `RebaseReport`/`RebaseRenumber` had become reachable from the public backend contract *precisely because* `RebaseRemote` sat on `RemoteStore`. Off the engine interface they're internal again and no aliases are needed.

The cost of moving off `RemoteStore` is that the compiler no longer forces either store to keep the method, and that failure would be silent — `bd dolt rebase` reporting an incapable backend on a capable one. Guarded with `var _ storage.ChildCollisionRebaser` on both stores, plus tests pinning that the capability is reachable through a `HookFiringStore` decorator via `UnwrapStore` but **not** by direct assertion. If that direct assertion ever starts passing, the method has leaked back onto `DoltStorage` and the public surface has widened again.

**Tests.** Four new cases in `embeddeddolt`, against the real engine. Each was verified to fail with its fix reverted — identical add/add renumbered anyway; the colliding child under an identical parent dropped entirely; the bystander subtree dragged off its id (`read title of rbp-swap.1000000.1: sql: no rows in result set`); `Restored` false after a genuine rolled-back failure. Plus the two decorator/capability tests in `internal/storage`.

**Local gate on the pushed head.** `make test` (full, `TEST_TIMEOUT=15m`) and `make ci-pr-core` both green; `make api-check`, doc-flags, doc-freshness, version and `testing.Short` checks pass; all ten `TestEmbeddedRebase*` pass. Two caveats, both environmental rather than clean results: `make ci-pr-lint` reports 5 gosec hits, all of which reproduce identically on a clean `main` checkout in a scratch worktree and none of which are in files this branch touches; and `ci-pr-policy`'s first step dies on macOS's bash 3.2 (`mapfile: command not found`), so the remaining steps were run individually and pass — that script only scans `*.sh`, workflows, hooks and the `Makefile`, none of which this branch changes.

**Still open, not taken.** `jsonOutput` on `doltRebaseCmd` and the `CountersSet` iteration order; and the swap's renames still write `renamed` audit events to and from the scratch id — the ids are now vacancy-checked but the audit noise is unchanged. Happy to take either, or leave them to you.

maphew (approved):
Re-review of c40c1d075 and c0e6224dc: all three items I had offered to absorb maintainer-side are done on the branch - better than a carry, since each comes with a regression test I verified bites with the fix reverted. Approving.

- **Swap-temp vacancy**: `allocateSwapTemp` probes each candidate as a *subtree* prefix over both issues and wisps, takes the first with zero occupants, marks it taken per pair, and hard-errors rather than guessing - which closes the capture-an-unrelated-subtree hazard of the fixed `parent.1000000+i` scheme.
- **`Restored` flag**: set only by the restore defer after a successful `DOLT_RESET`, so a successful rebase whose post-merge `is_blocked` recompute failed now gets its own distinct message instead of the false "database was restored / no changes were kept" rollback text.
- **Identical add/add guard**: `dropIdenticalAddAdds` skips a candidate only when both sides carry a non-empty `content_hash` and they agree (absent hash falls through to renumber, erring toward the recoverable side), and the ordering subtlety - identical filter before the collision-root filter - is handled and pinned.
- **Round-3 capability interface** (c0e6224dc): `ChildCollisionRebaser` as a narrow optional interface reached via `storage.UnwrapStore`, same pattern as `Flattener`/`Compactor`; `remote.go`/`beads.go`/`backend.go` untouched, so the exported surface stays clean for out-of-tree implementers.

Two residual non-gating items for either a tiny follow-up push or a maintainer touch-up:

1. **Double-failure messaging** (`cmd/bd/dolt.go` ~770): when the reconcile fails *and* the automatic `DOLT_RESET` restore also fails, `Restored` stays false with `Renumbered` non-empty, so the CLI prints "The rebase itself completed and was committed; nothing was rolled back... run bd doctor --fix" directly under an error saying the restore to the backup tag failed. That contradictory state (committed renumber, aborted merge, failed restore) deserves its own message pointing at the backup tag.
2. Nit: the identical guard covers collision candidates, not identical descendants inside a genuinely colliding parent's subtree - `renameSubtreeInTx` renames indiscriminately, so such a child still forks. Contrived shape, and everything forked before the guard existed, so noting it is enough.

The mutually agreed follow-ups (`--json` on `doltRebaseCmd`, deterministic `CountersSet` ordering, scratch-id audit noise) remain open by explicit agreement. Branch is level with main, CI fully green including both required gates, mergeable.

Process note: I stepped down as maintainer on 2026-08-10 and can no longer land this. @csells @julianknutsen - ready for merge from my side.

rjc123:
Still `MERGEABLE` / `CLEAN` at `c0e6224dc` with CI green — nothing needed from me, but say the word if you'd like `main` merged in before you get to it and I'll refresh the branch.

rjc123:
@pcresswell's [report on #4796](https://github.com/gastownhall/beads/issues/4796#issuecomment-5542789240)
looks right to me — same collision on current stable (`bd`
v1.2.2), and the top-level ID case they hit suggests the gap is offline ID allocation
generally rather than `child_counters` specifically.

No chase on this PR. v1.3.0 looks like it is fairly taking the attention: the last stable sits
around 1,350 commits behind `main`, the oldest of them from mid-May, so it is a big release
with a lot of old in it.

So I'm making no changes to this branch — including the wider scope @pcresswell's report
points at — until v1.3.0 has landed and the newer PRs have had a look. Happy to merge `main`
in whenever it's useful.

PR #4851, opened 2026-07-16
refactor(storage): name-keyed backend registry for storage backend selection

julianknutsen:
Follow-up to #4847: convert backend selection from hardcoded dispatch to a name-keyed registry.

Today adding (or removing) a backend means touching every dispatch site: the `store_factory*.go` if-chains, the primary open ladder in `main.go`, workspace discovery, and the `configfile` backend enum. This PR replaces all of that with a small registry (`internal/storage/backends`): a backend is one registrant file that calls `backends.Register` from `init()`, plus a blank import in `cmd/bd/backends_builtin.go`. No shared dispatch code changes.

**What's in the registry contract:**
- `Open` / `OpenReadOnly` — open a workspace from `.beads` config
- `Provision` — `bd init` provisioning, taking a backend-agnostic `ProvisionRequest{BeadsDir, Options}` and returning config entries to persist in `metadata.json` (`backend_config` map)
- `WorkspaceIsBeadsDir` — workspace-discovery predicate

**CLI surface:**
- New generic `bd init --backend-opt key=value` (repeatable) passes provisioning options to the selected backend; values may contain `=` (DSNs etc.). Option keys are owned by the backend.
- `--sqlite-path` stays as sugar for the sqlite backend's `path` option (conflicting values in both spellings are rejected explicitly).
- Existing workspaces are untouched: sqlite keeps reading/writing the legacy `sqlite_path` field so older `bd` versions and current workspaces interoperate.

**Behavior guarantees, tested:**
- SQLite converts to the registry with byte-identical behavior; the dispatch conversion is covered by contract tests in `cmd/bd` plus registry unit tests.
- Precedence is pinned: a *registered* backend name is classified as supported ahead of the removed-backend tombstones from #4847, so a backend can be reintroduced additively by registration alone — while unregistered removed names keep failing closed with the migration guidance.
- Dolt paths are unchanged (defense-in-depth guard test included).

This keeps the core simple while making backend membership a build-time composition decision — local-first backends (e.g. the flat-file backend proposed in #4415) can plug into the storage seam and the conformance oracle (`internal/storage/conformance`) without touching shared dispatch code.

### DIFF
--- a/cmd/bd/backend_registry_contract_test.go
+++ b/cmd/bd/backend_registry_contract_test.go
@@ -1,0 +1,106 @@
+package main
+
+// Extension-contract tests for the backend registry: registering a backend
+// by name (one additive registrant file in a real extension) is enough for
+// the metadata-driven store factories to dispatch to it, wins over the
+// removed-backend tombstones, and leaves unregistered names failing closed
+// exactly as before.
+
+import (
+	"context"
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/configfile"
+	"github.com/steveyegge/beads/internal/storage"
+	"github.com/steveyegge/beads/internal/storage/backends"
+)
+
+var (
+	errContractOpen         = errors.New("contract backend: read-write open")
+	errContractOpenReadOnly = errors.New("contract backend: read-only open")
+)
+
+// registerContractBackend registers a fixture backend whose open functions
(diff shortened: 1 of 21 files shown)


julianknutsen:
Superseded: #4857 restored the PostgreSQL/MySQL adapters, so this branch's assumptions (removed-backend tombstones, sqlite-only registration) no longer match main. The registry refactor is being reworked on top of current main with all four backends registering through the same seam; a fresh PR will follow.

Issue #4534, opened 2026-07-01
bd create fails with fk_counter_parent FK violation on all inserts (even fresh top-level issues) — regression after schema migration


eitsupi:
## Summary

`bd create` fails on every invocation — even creating a brand-new top-level issue with no `--parent` — with a foreign key violation on `child_counters`.
Reads (`bd show`, `bd list`, `bd stats`) are unaffected; only inserts are broken.

## Environment

- bd version: `1.1.0-rc.1 (fa4dce454)`
- OS: macOS (Darwin 25.5.0, arm64)
- Storage backend: Dolt (embedded)
- `bd doctor`: 0 errors, all checks pass (including "Pending Migrations: None required", "Dolt Schema: All required tables present")

## Steps to reproduce

```bash
bd create --title="test" --type=task --priority=4
```

## Expected

A new issue is created successfully.

## Actual

```
Error: failed to insert issue proj_name-x46: insert issue into issues: Error 1105 (HY000): cannot add or update a child row: a foreign key constraint fails (`proj_name`.`child_counters`, CONSTRAINT `fk_counter_parent` FOREIGN KEY (`parent_id`) REFERENCES `issues` (`id`))
```

Same result with `--parent=<existing-valid-parent-id>` — the parent issue itself exists and is readable via `bd show <parent-id>`.

## Investigation notes

- `bd doctor`, `bd migrate --inspect`, `bd migrate schema` all report clean (schema already at v53, zero pending/registered migrations).
- `ignored_schema_migrations` has exactly 10 rows, all timestamped the same moment a `dolt log` commit "schema: apply migrations" landed — i.e. the DB went through a migration recently, and `bd create` broke starting right after that commit.
- Since the FK violation reproduces for a *brand-new* top-level issue (no `--parent`, no pre-existing relationship to any parent row), the violation can't be about a bad/missing parent value — it looks like an insert-ordering regression where the `child_counters` row is written (or the FK check evaluated) before the new issue's own row in `issues` is visible to the same transaction/connection.
- Direct `SELECT` against `child_counters` and `issues` shows a structurally intact schema (normal FK definition, no orphaned rows).

## Impact

This blocks all issue creation on an otherwise healthy, actively-used Dolt database (259 existing issues, all reads fine). No data corruption observed so far, but the DB is currently unusable for new work.

## What we have NOT done

[...]

Issue #1393, opened 2026-01-30
CreateIssueImport does not update child_counters table


wanthmen:
## Summary

When importing issues with hierarchical IDs (e.g., `bd-t8a.1`) via the transaction path, the `child_counters` table is not updated. This causes UNIQUE constraint failures when later creating new child issues with `--parent`.

## Steps to Reproduce

1. Export issues to JSONL (includes hierarchical IDs like `bd-parent.1`, `bd-parent.2`)
2. Delete the database
3. Import from JSONL: `bd import -i .beads/issues.jsonl`
4. Try to create a new child: `bd create "New task" --parent bd-parent`

## Expected Behavior

New child should be created with ID `bd-parent.3` (next available number).

## Actual Behavior

```
Error: operation failed: failed to create issue: insert issue: failed to insert issue: 
sqlite3: constraint failed: UNIQUE constraint failed: issues.id
```

The system tries to create `bd-parent.1` because `child_counters` table has no entry for `bd-parent`.

## Root Cause

`CreateIssueImport` in `internal/storage/sqlite/import_tx.go` does not call `ensureChildCounterUpdatedWithConn` after inserting hierarchical issues.

Compare the two code paths:

### ✅ `CreateIssuesWithFullOptions` (batch_ops.go) - CORRECT

```go
// Phase 4: Bulk insert issues
if err := bulkInsertIssues(ctx, conn, issues); err != nil {
    return wrapDBError("bulk insert issues", err)
}

// Phase 4.5: Update child counters for hierarchical IDs (GH#728 fix)
if err := updateChildCountersForHierarchicalIDs(ctx, conn, issues); err != nil {
    return wrapDBError("update child counters", err)
}
```

### ❌ `CreateIssueImport` (import_tx.go) - MISSING

```go
// Insert issue (strict)
if err := insertIssueStrict(ctx, t.conn, issue); err != nil {
    return fmt.Errorf("failed to insert issue: %w", err)
}
// Record event
if err := recordCreatedEvent(ctx, t.conn, issue, actor); err != nil {
    return fmt.Errorf("failed to record creation event: %w", err)
}
// Mark dirty
if err := markDirty(ctx, t.conn, issue.ID); err != nil {
    return fmt.Errorf("failed to mark issue dirty: %w", err)
}
// ❌ NO child counter update!
return nil
```

## Proposed Fix

Add child counter update in `CreateIssueImport` after `insertIssueStrict`:

```go
// Insert issue (strict)
[...]

Issue #728, opened 2025-12-24
bd create --parent allocates colliding child IDs when child_counters not backfilled


garitar:
### Bug

When a repo already contains hierarchical issue IDs like `<parent>.<n>` created via explicit `--id` (or imported from JSONL), `bd create --parent <parent>` can allocate a colliding child ID starting at `.1`. In my case the command exited 0, returned the existing child ID, and emitted a UNIQUE constraint warning on dependencies.

### Repro (v0.35.0, direct mode)

```bash
mkdir repro && cd repro
git init -q
bd --no-daemon init --skip-hooks --skip-merge-driver --prefix bd

bd --no-daemon create --id bd-test "Parent" -t epic -p 2
bd --no-daemon create --id bd-test.1 "Existing child 1" -t task -p 2 --deps parent-child:bd-test
bd --no-daemon create --id bd-test.2 "Existing child 2" -t task -p 2 --deps parent-child:bd-test

bd --no-daemon create "New child via --parent" -t task -p 2 --parent bd-test --json
```

Observed:
- Output `id` is `bd-test.1` (collision) instead of expected `bd-test.3`.
- Warning:
  - `sqlite3: constraint failed: UNIQUE constraint failed: dependencies.issue_id, dependencies.depends_on_id, dependencies.type`

### Root cause (code pointers)

In direct mode, `bd create --parent` generates the child ID by calling `store.GetNextChildID(ctx, parentID)` in `cmd/bd/create.go`.

SQLite `GetNextChildID` uses the `child_counters` table (see `internal/storage/sqlite/hash_ids.go`):

```sql
INSERT INTO child_counters (parent_id, last_child)
VALUES (?, 1)
ON CONFLICT(parent_id) DO UPDATE SET
  last_child = last_child + 1
RETURNING last_child
```

The `child_counters` table is created by migration `internal/storage/sqlite/migrations/014_child_counters_table.go`, but it is **not backfilled** from existing `issues.id` values. So repos that already contain children created via explicit IDs (or JSONL import) can have `child_counters` missing/low; the first allocation then starts at 1 and collides.

(As a reference point: the in-memory backend has a test showing `LoadFromIssues` initializes child counters based on existing IDs: `internal/storage/memory/ready_blocked_nodb_test.go`.)

### Suggested fix options

Any one of these would address the collision:

[...]

PR #4858, opened 2026-07-17
fix(doctor): detect orphaned child_counters rows (#4539)

### LINKED ISSUE #4539: bd doctor: detect orphaned child_counters rows (invisible create-bricking drift, #4534 follow-up)
## Summary

Follow-up from #4534: a single orphaned `child_counters` row (a `parent_id` with no matching `issues` row) can fail Dolt constraint validation on subsequent writes, bricking every `bd create` — while `bd doctor` reports 0 errors, "Pending Migrations: None required", and "Dolt Schema: All required tables present". The operator has no tool-side signal that the database is in this state; the reporter had to find it with a manual `LEFT JOIN`.

#4538 heals the known legacy window (rows orphaned while `fk_counter_parent` was dropped between migration 0039 and clone-local 0002) with a one-time cleanup migration. But the drift class that creates such rows (#728, #1393 — `child_counters` not kept in sync with `issues` on all code paths) predates that window and may recur, and any future recurrence would again be invisible until the next insert fails.

## Proposal

Add a `bd doctor` check:

```sql
SELECT cc.parent_id, cc.last_child
FROM child_counters cc
LEFT JOIN issues i ON cc.parent_id = i.id
WHERE i.id IS NULL;
```

- Non-empty result → actionable error naming the dangling `parent_id`s, with the repair (`DELETE` of those rows, matching the FK's cascade semantics) either suggested or offered behind `bd doctor --fix`.
- Same check for `wisp_child_counters` vs `wisps` would close the twin gap.

Credit: suggested by @eitsupi in #4534.

vishnujayvel:
## What

Adds a `bd doctor --check=validate` check ("Orphaned Child Counters") that
detects `child_counters` rows whose `parent_id` has no matching `issues`
row, and the twin case: `wisp_child_counters` rows whose `parent_id` has no
matching `wisps` row. Both are fixable via `bd doctor --fix`, which deletes
the dangling rows.

## Why

Follow-up from #4534. A single orphaned `child_counters` row (a `parent_id`
with no matching `issues` row) can fail Dolt constraint validation on a
later write, silently bricking every `bd create` — while `bd doctor`
reported 0 errors, since no check looked at this table. #4538 (merged)
healed the known legacy window with a one-time cleanup migration
(`migrations/ignored/0011_cleanup_orphaned_child_counters.up.sql`), but the
drift class that creates such rows (#728, #1393 — `child_counters` not kept
in sync with `issues` on all code paths) predates that window and can
recur, with no ongoing guard until now.

Fixes #4539.

Credit: reported by @maphew; the doctor-check approach was suggested by
@eitsupi in #4534.

## Implementation

Follows the existing "Orphaned Dependencies" check pattern exactly:

- `cmd/bd/doctor/validation.go`: `CheckOrphanedChildCounters` /
  `checkOrphanedChildCountersDB` — same `openStoreDB` + core-logic-on-`*sql.DB`
  split as `CheckOrphanedDependencies` / `checkOrphanedDependenciesDB`, so the
  core logic is unit-testable against a live Dolt store without going
  through `ResolveBeadsDirForRepo`.
- `cmd/bd/doctor/checks_nocgo.go`: non-CGO stub (`"Skipped: requires CGO"`),
  matching every other Dolt-backed check.
- `cmd/bd/doctor/fix/validation.go`: `OrphanedChildCounters` — same
  transaction + `DOLT_COMMIT` shape as `OrphanedDependencies`, deletes by
  `parent_id` from whichever table the orphan came from.
- `cmd/bd/doctor_validate.go`: registered in `collectValidateChecks` as
  `fixable: true`, next to `CheckOrphanedDependencies`.
- `cmd/bd/doctor_fix.go`: wired into `applyFixList`'s dispatch switch under
  case `"Orphaned Child Counters"`.

Both orphan classes (child_counters↔issues, wisp_child_counters↔wisps) are
reported as one check via a `UNION ALL`, matching how "Orphaned
Dependencies" already unions `dependencies` and `wisp_dependencies` into a
single check rather than two.

## Verification

- New unit tests against a live Dolt store, mirroring the existing
  `checkOrphanedDependenciesDB` tests exactly:
  - `cmd/bd/doctor/validation_test.go`:
    `TestCheckOrphanedChildCountersDB_NoOrphans`,
    `TestCheckOrphanedChildCountersDB_ChildCountersOrphanDetected`,
    `TestCheckOrphanedChildCountersDB_WispChildCountersOrphanDetected`
    (the FK-checks-disabled insert trick these use is the same one
    #4538's migration and the existing orphan-dependency tests use).
  - `cmd/bd/doctor_validate_test.go`:
    `TestValidateCheck_DetectsOrphanedChildCounters` exercises the full
    `collectValidateChecks` wiring (not just the core-logic function).
  - `cmd/bd/doctor/fix/validation_test.go`:
    `TestFixFunctions_RequireBeadsDir` extended with `OrphanedChildCounters`,
    plus a new `TestOrphanedChildCounters_FixDeletesOnlyOrphans` covering the
    fix's DELETE itself (not just the missing-`.beads`-dir error path) — seeds
    a live store with both valid and orphaned `child_counters`/
    `wisp_child_counters` rows and asserts the fix removes exactly the
    orphans, leaves the valid rows alone, the check reports clean afterward,
    and a second run is a safe no-op. Unlike its siblings in this file, it
    does **not** depend on Docker: it uses `dolt.New(..., AutoStart: true)`
    to auto-start a real local `dolt sql-server` via the `dolt` binary on
    PATH — the same mechanism `bd init` uses in production and the one CI's
    `RequireDoltBinary` check already requires — so it actually executes
    (not skips) in this sandboxed dev environment, and ran green as its own
    first-class test rather than a throwaway substitute. Revert-proofed
    directly against the shipped code: temporarily replaced both `DELETE`
    statements in `OrphanedChildCounters` with a no-op query, reran — the
    test failed at every orphan-removal assertion as expected — then
    restored the original code (`git diff` clean) and reran — passed again.
  These all use the package's standard `newTestDoltStore`/
  `setupValidateTestDB` helpers, which need a local
  `dolthub/dolt-sql-server:2.2.0` Docker test container — same as every
  sibling orphan-check test in this codebase. Sandboxed dev environment
  here could not pull that image (Docker Hub pulls hung entirely, even for
  `hello-world`), so these skip locally exactly like every other
  Docker-backed test in the suite.
  - **Docker-free companions (shipped, not throwaway):** each of the four
    tests above now has a permanent `_LocalDolt` companion —
    `TestCheckOrphanedChildCountersDB_NoOrphans_LocalDolt`,
    `_ChildCountersOrphanDetected_LocalDolt`,
    `_WispChildCountersOrphanDetected_LocalDolt` (`cmd/bd/doctor/validation_test.go`),
    and `TestValidateCheck_DetectsOrphanedChildCounters_LocalDolt`
    (`cmd/bd/doctor_validate_test.go`) — dual-harnessed alongside the
    Docker-backed originals (which are untouched), using the same
    `dolt.New(..., AutoStart: true)` local-server pattern as
    `TestOrphanedChildCounters_FixDeletesOnlyOrphans`. All four ran green
    for real in this sandbox. Revert-proofed directly: temporarily replaced
    `checkOrphanedChildCountersDB`'s two `WHERE ... IS NULL` predicates with
    `WHERE 1 = 0` (detection disabled) — the two orphan-detected `_LocalDolt`
    tests and the validate-check `_LocalDolt` test failed as expected, the
    no-orphans test still passed; restored the original code (`git diff`
    clean) and reran — all four passed again.
  - **Important caveat, disclosed rather than hidden:** none of the four
    original tests, and — despite being Docker-free — none of the four new
    `_LocalDolt` companions either, currently execute in any
    `pull_request`-triggered beads CI job. This isn't a Docker-availability
    problem, it's two independent, structural CI gaps we found while
    verifying this: (1) `scripts/ci/lib/test-env.sh`'s `beads_test_env_enter`
    unconditionally sets `BEADS_TEST_SKIP=dolt` for the one PR job whose
    package scope even includes `cmd/bd/doctor(/fix)` (`pr.yml`'s "PR Core"),
    since nothing in the repo ever sets the `BEADS_TEST_ENV_RUN_DOLT=1`
    escape hatch; (2) `pr-risk.yml`'s "Build (Embedded Dolt)" job only
    compiles `./cmd/bd/` (not `/...`) into its test binary, so
    `cmd/bd/doctor` and `cmd/bd/doctor/fix` tests aren't just skipped there,
    they're never even built. We've filed this as a separate upstream issue
    (draft: `beads-fix-application-tests-issue.md`) with a candidate CI-YAML
    fix (draft: `beads-ci-dolt-tests.patch`) rather than bundling a
    CI-workflow change into this test PR.
- `go build ./cmd/bd/...`, `go vet ./cmd/bd/...`, `gofmt -l` (clean),
  `golangci-lint run ./cmd/bd/...` (0 issues).
- `go test ./cmd/bd/...` — full suite passes except two **pre-existing,
  unrelated** failures in `cmd/bd/doctor` (`TestCheckBeadsRole_NotConfigured`,
  `TestCheckBeadsRole_NotGitRepo`), confirmed present on unmodified
  `origin/main` in this same environment (git-config-dependent, unrelated to
  `child_counters`/doctor-validate). Not touched by this PR.

## Project-identity guard on the destructive fix

`OrphanedChildCounters` DELETEs rows, so it is a destructive `bd doctor --fix`
entry point and now goes through `guardFixTarget` like its siblings. Without
it, the fresh connection `openDoltDB` dials could land on a different
project's database (the stale `.beads/dolt-server.port` scenario) and delete
that project's counter rows. The guard aborts with `PROJECT IDENTITY MISMATCH`
before any DELETE runs.

Test coverage for the guard:

- `OrphanedChildCounters` added to
  `TestDestructiveFix_AbortsOnProjectIdentityMismatch`'s entry-point table
  (`cmd/bd/doctor/fix/remotes_test.go`), which enumerates every destructive
  fix that shares `openDoltDB`/`openFixDB`.
- `TestOrphanedChildCounters_AbortsOnProjectIdentityMismatch`
  (`cmd/bd/doctor/fix/validation_test.go`) — a Docker-free companion to that
  table entry. The table's store needs the Dolt test container, so it skips
  wherever Docker is absent; this one uses the same local auto-start pattern
  as `TestOrphanedChildCounters_FixDeletesOnlyOrphans`, plants a real orphan
  row, and asserts both that the fix aborts and that the orphan survives.
  Mutation-checked: with the `guardFixTarget` call removed it fails at the
  `got nil error` check — which fires before the orphan-survival assertion, so
  the deletion itself is visible in the fix's stdout rather than in a second
  failing assertion; restored, it passes.

`TestOrphanedChildCounters_FixDeletesOnlyOrphans` seeds a *matching*
`project_id` so the guard admits it — which is also why it alone could not
pin the guard down, hence the dedicated mismatch test above.

## Test-environment isolation

The orphan-counters tests construct their own local Dolt server, so they must
not be redirected to an ambient one. Clearing the legacy `BEADS_DOLT_PORT`
alone was not enough: `BEADS_DOLT_SERVER_PORT` outranks it in
`applyConfigDefaults`, and `BEADS_DOLT_SERVER_HOST`, `_SERVER_SOCKET`,
`_SERVER_MODE`, `_SHARED_SERVER` and `_AUTO_START` can redirect `dolt.New`
just as effectively. Because this package's `TestMain` sets
`BEADS_TEST_SERVER=1`, a reachable ambient endpoint could have received the
test's `CreateIfMissing` plus its schema and data writes — i.e. running the
suite could write to a developer's real Dolt server.

`BEADS_DOLT_DATA_DIR` is cleared for the same reason by a different route: the
auto-start path resolves its data directory through `ResolveDoltDir` →
`projectDoltDirPath` (`internal/doltserver/physical_root.go:62`), which honours
that variable first. With it set, the test passed *silently* while laying
`.dolt/`, `.doltcfg/`, `.bd-dolt-ok` and its own database directory down in
whatever directory the variable named.

No shared helper existed for this; every isolated-Dolt call site either uses
the Docker container helpers or hand-rolls a partial scrub (the list in
`internal/storage/dolt/active_database_size_test.go` misses
`BEADS_DOLT_AUTO_START`). Added `testutil.ClearAmbientDoltEnv`
(`internal/testutil/testdoltcommon.go`) as the single place that list lives,
and used it. Existing partial scrubs are left alone to keep this PR scoped.

Verified with the hostile-env repro
`BEADS_DOLT_SERVER_HOST=192.0.2.1 BEADS_DOLT_SERVER_PORT=59999`: previously
the test failed at `dolt.New` with `Dolt server unreachable 192.0.2.1:59999`;
it now passes.

No unrelated files touched.

**Pre-existing note (not touched by this PR):** `TestValidateCheck_AllClean`
in `cmd/bd/doctor_validate_test.go` asserts `collectValidateChecks` returns
exactly 4 entries; it already returns 5 on `origin/main` (stale since
`CheckCrossTableDuplicates` was added), and now 6 with this PR's addition.
The assertion appears to have gone stale independently of this change —
flagging for visibility, left alone per "don't fix unrelated things."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2091,6 +2091,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `Storage` implementations that ignore the new fields simply do not enforce
   them and should add support before advertising guard semantics.
 
+- **`bd doctor --check=validate` now detects orphaned `child_counters` /
+  `wisp_child_counters` rows** ([#4539](https://github.com/gastownhall/beads/issues/4539),
+  follow-up to [#4534](https://github.com/gastownhall/beads/issues/4534)).
+  A `child_counters` row whose `parent_id` has no matching `issues` row (or a
+  `wisp_child_counters` row with no matching `wisps` row) can fail Dolt
+  constraint validation on a later write, silently bricking every
+  `bd create` — while `bd doctor` previously reported 0 errors. #4538 healed
+  the known legacy window with a one-time cleanup migration, but the drift
+  class that creates such rows (#728, #1393) predates that window and can
+  recur with no ongoing guard. The new "Orphaned Child Counters" check
+  flags dangling `parent_id`s and is fixable via `bd doctor --fix`.
+  Credit: reported by @maphew, check suggested by @eitsupi in #4534.
+
 - **Pool-aware claiming via the `claim.pools` config key** (bd-bguz6).
   Dispatcher fleets pre-assign issues to a pool pseudo-assignee (e.g.
   `fable-crew`); `--claim` previously refused those ("already assigned"),
--- a/cmd/bd/doctor.go
+++ b/cmd/bd/doctor.go
@@ -890,6 +890,11 @@ func runDiagnostics(path string) doctorResult {
 	result.Checks = append(result.Checks, cloneLocalFKCheck)
 	// Don't fail overall check for severed clone-local FKs, just warn
 
+	// Check 21a: Orphaned child counters (twin gap to orphaned deps, #4539)
+	orphanedChildCountersCheck := convertDoctorCheck(doctor.CheckOrphanedChildCounters(path))
+	result.Checks = append(result.Checks, orphanedChildCountersCheck)
+	// Don't fail overall check for orphaned child counters, just warn
+
 	// Check 22a: Child→parent dependencies (anti-pattern)
 	childParentDepsCheck := convertDoctorCheck(doctor.CheckChildParentDependencies(path))
 	result.Checks = append(result.Checks, childParentDepsCheck)
--- a/cmd/bd/doctor/agent.go
+++ b/cmd/bd/doctor/agent.go
@@ -118,6 +118,7 @@ var agentEnrichers = map[string]enricher{
 	"Test Pollution":               enrichTestPollution,
 	"Orphaned Dependencies":        enrichOrphanedDeps,
 	"Clone-Local FKs":              enrichCloneLocalFKs,
+	"Orphaned Child Counters":      enrichOrphanedChildCounters,
 	"Child-Parent Dependencies":    enrichChildParentDeps,
 	"Classic Artifacts":            enrichClassicArtifacts,
 	"Pending Migrations":           enrichPendingMigrations,
@@ -468,6 +469,17 @@ func enrichCloneLocalFKs(dc DoctorCheck) agentEnrichment {
 		expected:    "Every FK on clone-local tables (events, wisp_dependencies, wisp_labels, wisp_comments, wisp_events, wisp_child_counters) present and enforcing",
 		commands:    []string{"bd doctor --fix"},
 		sourceFiles: []string{"cmd/bd/doctor/clone_local_fks.go:CheckCloneLocalFKs"},
+	}
+}
+
+func enrichOrphanedChildCounters(dc DoctorCheck) agentEnrichment {
+	return agentEnrichment{
+		severity:    "blocking",
+		explanation: fmt.Sprintf("Orphaned child counter rows found: %s. A child_counters/wisp_child_counters row with no matching issues/wisps parent can fail Dolt constraint validation on a later write, silently bricking every 'bd create' (#4539).", dc.Message),
+		observed:    dc.Message + "\n" + dc.Detail,
+		expected:    "Every child_counters/wisp_child_counters row has a live parent in issues/wisps",
+		commands:    []string{"bd doctor --check=validate --fix"},
+		sourceFiles: []string{"cmd/bd/doctor/validation.go:CheckOrphanedChildCounters"},
 	}
 }
 
--- a/cmd/bd/doctor/fix/remotes_test.go
+++ b/cmd/bd/doctor/fix/remotes_test.go
@@ -158,6 +158,7 @@ func TestDestructiveFix_AbortsOnProjectIdentityMismatch(t *testing.T) {
 		"CrossTableDuplicates":    func(p string) error { return CrossTableDuplicates(p, false) },
 		"OrphanedDependencies":    func(p string) error { return OrphanedDependencies(p, false) },
 		"ChildParentDependencies": func(p string) error { return ChildParentDependencies(p, false) },
+		"OrphanedChildCounters":   func(p string) error { return OrphanedChildCounters(p, false) },
 		"RecomputeBlocked":        RecomputeBlocked,
 	} {
 		t.Run(name, func(t *testing.T) {
--- a/cmd/bd/doctor/fix/validation.go
+++ b/cmd/bd/doctor/fix/validation.go
@@ -116,6 +116,105 @@ func OrphanedDependencies(path string, verbose bool) error {
 	return nil
 }
 
+// OrphanedChildCounters removes child_counters/wisp_child_counters rows whose
+// parent_id has no matching issues/wisps row. A single such row can fail Dolt
+// constraint validation on subsequent writes, silently bricking every
+// `bd create` (#4539, follow-up to #4534). Credit: check suggested by
+// @eitsupi in #4534.
+// If verbose is true, prints each removed row; otherwise shows only summary.
+func OrphanedChildCounters(path string, verbose bool) error {
+	beadsDir, err := resolvedWorkspaceBeadsDir(path)
+	if err != nil {
+		return err
+	}
+
+	db, err := openDoltDB(beadsDir)
+	if err != nil {
+		fmt.Printf("  Orphaned child counters fix skipped (%v)\n", err)
+		return nil
+	}
+	defer db.Close()
+
+	// Find child_counters rows dangling from issues and wisp_child_counters
+	// rows dangling from wisps.
+	query := `
+		SELECT 'child_counters' AS counter_table, cc.parent_id
+		FROM child_counters cc
+		LEFT JOIN issues i ON cc.parent_id = i.id
+		WHERE i.id IS NULL
+		UNION ALL
+		SELECT 'wisp_child_counters' AS counter_table, wcc.parent_id
+		FROM wisp_child_counters wcc
+		LEFT JOIN wisps w ON wcc.parent_id = w.id
+		WHERE w.id IS NULL
+	`
+	rows, err := db.Query(query)
+	if err != nil {
+		return fmt.Errorf("failed to query orphaned child counters: %w", err)
+	}
+	defer rows.Close()
+
+	type orphan struct {
+		counterTable string
+		parentID     string
+	}
+	var orphans []orphan
+
+	for rows.Next() {
+		var o orphan
+		if err := rows.Scan(&o.counterTable, &o.parentID); err == nil {
+			orphans = append(orphans, o)
+		}
+	}
+	if err := rows.Err(); err != nil {
+		return fmt.Errorf("row iteration error: %w", err)
+	}
+
+	if len(orphans) == 0 {
+		fmt.Println("  No orphaned child counters to fix")
+		return nil
+	}
+
+	// Delete orphaned child counter rows.
+	// Uses explicit transaction so writes persist when @@autocommit is OFF
+	// (e.g. Dolt server started with --no-auto-commit).
+	showIndividual := verbose || len(orphans) < 20
+	tx, err := db.Begin()
+	if err != nil {
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	var removed int
+	for _, o := range orphans {
+		var err error
+		switch o.counterTable {
+		case "child_counters":
+			_, err = tx.Exec("DELETE FROM child_counters WHERE parent_id = ?", o.parentID)
+		case "wisp_child_counters":
+			_, err = tx.Exec("DELETE FROM wisp_child_counters WHERE parent_id = ?", o.parentID)
+		default:
+			fmt.Printf("  Warning: skipped orphaned child counter from unexpected table %s\n", o.counterTable)
+			continue
+		}
+		if err != nil {
+			fmt.Printf("  Warning: failed to remove %s row for %s: %v\n", o.counterTable, o.parentID, err)
+		} else {
+			removed++
+			if showIndividual {
+				fmt.Printf("  Removed orphaned %s row: parent_id=%s\n", o.counterTable, o.parentID)
+			}
+		}
+	}
+	if err := tx.Commit(); err != nil {
+		return fmt.Errorf("failed to commit orphaned child counter removals: %w", err)
+	}
+
+	// Commit changes in Dolt
+	_, _ = db.Exec("CALL DOLT_COMMIT('-Am', 'doctor: remove orphaned child counter rows')") // Best effort: commit advisory; schema fix already applied in-memory
+
+	fmt.Printf("  Fixed %d orphaned child counter row(s)\n", removed)
+	return nil
+}
+
 // ChildParentDependencies removes child→parent blocking dependencies.
 // These often indicate a modeling mistake (deadlock: child waits for parent, parent waits for children).
 // Requires explicit opt-in via --fix-child-parent flag since some workflows may use these intentionally.
--- a/cmd/bd/doctor/fix/validation.go
+++ b/cmd/bd/doctor/fix/validation.go
@@ -1,12 +1,14 @@
 package fix
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"path/filepath"
 
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/steveyegge/beads/internal/configfile"
+	"github.com/steveyegge/beads/internal/storage/issueops"
 )
 
 // getDatabasePath returns the actual database directory path, respecting dolt_data_dir.
@@ -121,6 +123,19 @@ func OrphanedDependencies(path string, verbose bool) error {
 // constraint validation on subsequent writes, silently bricking every
 // `bd create` (#4539, follow-up to #4534). Credit: check suggested by
 // @eitsupi in #4534.
+//
+// The pinned engine defaults to repeatable-read with `FOR UPDATE` as a no-op,
+// so the delete below re-checks parent absence at delete time via a NOT
+// EXISTS predicate rather than keying off a pre-scanned list — a parent
+// created between the scan and the delete keeps its (now valid) counter.
+//
+// Mirrors DoltStore.runDoltTransaction (internal/storage/dolt/transaction.go,
+// GH#2455): the DML transaction, staging, and DOLT_COMMIT all run on one
+// pinned *sql.Conn, since each pool connection has an independent working set
+// in Dolt SQL server mode and mixing connections would let DOLT_COMMIT see
+// stale or unrelated state. Only the tables actually touched are staged, so
+// an unrelated dirty working set is not swept under this commit.
+//
 // If verbose is true, prints each removed row; otherwise shows only summary.
 func OrphanedChildCounters(path string, verbose bool) error {
 	beadsDir, err := resolvedWorkspaceBeadsDir(path)
@@ -135,8 +150,11 @@ func OrphanedChildCounters(path string, verbose bool) error {
 	}
 	defer db.Close()
 
+	ctx := context.Background()
+
 	// Find child_counters rows dangling from issues and wisp_child_counters
-	// rows dangling from wisps.
+	// rows dangling from wisps, for reporting only — the deletes below
+	// re-check parent absence themselves rather than keying off this list.
 	query := `
 		SELECT 'child_counters' AS counter_table, cc.parent_id
 		FROM child_counters cc
@@ -148,11 +166,10 @@ func OrphanedChildCounters(path string, verbose bool) error {
 		LEFT JOIN wisps w ON wcc.parent_id = w.id
 		WHERE w.id IS NULL
 	`
-	rows, err := db.Query(query)
+	rows, err := db.QueryContext(ctx, query)
 	if err != nil {
 		return fmt.Errorf("failed to query orphaned child counters: %w", err)
 	}
-	defer rows.Close()
 
 	type orphan struct {
 		counterTable string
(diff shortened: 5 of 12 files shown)


maphew:
Whips. The status change to Ready for Review was a fat finger slip. Switch back of not ready

vishnujayvel:
No worries — switched it back to draft. (It's queued behind a couple of our other open PRs anyway; will mark it ready once those clear.)

vishnujayvel:
Marking this ready after all — the test suite firmed up since my last comment (fix-application test + Docker-free companion tests for all checks, all executing locally against a real `dolt sql-server`). Review whenever convenient, no rush.

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the orphan-counter check is a genuinely good doctor addition; the destructive repair path needs hardening before it ships.

**Blocker 1 — scan-then-delete can remove a counter that became valid concurrently.** Orphans are identified before the transaction begins, and the delete keys only on `parent_id` without re-checking parent absence. The pinned engine defaults to repeatable-read and implements `FOR UPDATE` as a no-op, so a parent created between scan and delete loses its (now valid) counter. Use an atomic conditional `DELETE ... WHERE NOT EXISTS (SELECT 1 FROM issues ...) AND NOT EXISTS (... wisps ...)` instead of the pre-scan list.

**Blocker 2 — the Dolt commit is unsafe and its failure is invisible.** Verified: `_, _ = db.Exec("CALL DOLT_COMMIT('-Am', ...)")` — through the *pool*, error discarded. Beads' own transaction layer documents that DML and `DOLT_COMMIT` must share one pinned connection (working sets are session-local), and `-Am` stages *every* table, so unrelated dirty state rides along in the repair commit — or the commit silently fails while doctor still reports "Fixed." Pin one connection, stage only `child_counters`, check the result, and test both Dolt history and unrelated-dirty-table preservation.

**Should-fix — the advertised remediation doesn't discover the check.** The user-facing message and CHANGELOG say `bd doctor --fix`, but bare doctor never calls `CheckOrphanedChildCounters` — only `--check=validate` collects it. Wire it into bare doctor or fix every remediation string.

**Should-fix — none of the new regression tests execute in PR CI** (`RequireDoltBinary` skips under `BEADS_TEST_SKIP=dolt`, and the risk shards run only `TestEmbedded*`), leaving the destructive path without executed coverage. Also: doctor's command help and generated docs still omit the new check.

Wiring for detection/dispatch/enrichment all verified present; drift clean. Fix the two transaction-safety points and this is a solid #4539 close.

vishnujayvel:
Thanks for the cross-vendor review — all four items are addressed, and the two blockers were both real.

**Blocker 1 — scan-then-delete race.** The deletes keyed off the stale pre-scan `parent_id` list with no re-check at delete time. Both deletes in `OrphanedChildCounters` are now atomic `DELETE ... WHERE NOT EXISTS (SELECT 1 FROM issues/wisps ...)`, matching the shape you suggested; the pre-scan is now reporting-only.

**Blocker 2 — `DOLT_COMMIT` via the pool with `-Am`, error discarded.** Now pins a single `*sql.Conn` (`db.Conn(ctx)`) for the whole DML + stage + commit sequence, stages only `child_counters` and `wisp_child_counters` via `CALL DOLT_ADD` rather than a blanket `-Am`, and checks the commit error via `issueops.IsNothingToCommitError` — the same helper `cmd/bd/doctor/fix/blocked.go` already uses. This mirrors the existing idiom in `internal/storage/dolt/transaction.go:62-71` (`runDoltTransaction`, GH#2455) rather than inventing a new pattern.

**Should-fix — bare `bd doctor --fix` didn't discover the check.** Confirmed: the sibling `CheckOrphanedDependencies` was wired in at `cmd/bd/doctor.go:837` but `CheckOrphanedChildCounters` was not, so it only ran under `--check=validate`. Wired in one block below it, same style.

**Should-fix — new tests don't execute in PR CI.** No change here, deliberately: the PR description already discloses this and files it as separate work with a draft CI patch, and bundling a workflow change into a test PR is the thing that disclosure was avoiding.

Two things worth flagging honestly rather than leaving you to find them:

1. **The first version of Blocker 1's fix was wrong, and the tests caught it.** I'd written `DELETE FROM child_counters cc WHERE NOT EXISTS (...)` — Dolt rejects a table alias in a single-table `DELETE` (`Error 1105 (HY000): syntax error at position 30 near 'cc'`), so `TestOrphanedChildCounters_FixDeletesOnlyOrphans` failed outright. Fixed in `f768fdedd` by qualifying the correlated subquery with the table name instead of an alias. Both commits are on the branch; squash if you'd rather not carry the intermediate step.
2. **This branch was rebased** onto `origin/main` (`4c63e1510`) to clear a conflict, so the history changed under you. The only conflict was in `CHANGELOG.md` and was purely additive — both entries kept, upstream's first.

Test results, run locally:

- `go test ./cmd/bd/doctor/fix/...` — **ok** (16.7s), including `TestOrphanedChildCounters_FixDeletesOnlyOrphans`.
- `go test ./cmd/bd/doctor/...` — two failures, both **pre-existing and environmental**, not from this change: `TestCheckBeadsRole_NotConfigured` and `TestCheckBeadsRole_NotGitRepo`. They reproduce identically on an unmodified base in this checkout (a global `git config beads.role` leaks into the test environment).
- `gofmt -l` on all touched files — clean.

One limitation: Docker isn't available on this machine, so the Dolt-backed subtests skip. The coverage above is the in-process path only.

maphew (commented):
Status update after a full re-verification: the code is right, and what remains is a maintainer scoping call — not more work for you right now.

What we confirmed: both blockers from the last round are genuinely fixed. The deletes are now atomic `NOT EXISTS` predicates with the pre-scan demoted to reporting; the commit sequence pins a single connection, stages only the two touched tables, and tolerates no-op commits through the shared error check — mirroring the existing transaction discipline rather than inventing a new pattern. On policy: the detection half is one read-only query, wired behind the embedded-mode gate, so it cannot touch embedded storage; the `--fix` path dials the server over TCP only. Structurally it's a clone of the shipped orphaned-dependencies check — exactly the "one check at a time" shape the doctor policy asks for.

The open question is ours to decide: the destructive `--fix` half ships with zero executed CI coverage (the Dolt-backed subtests skip everywhere in CI today — a gap you disclosed honestly, and which is tracked on our side as pre-existing debt). The choice is between landing the whole PR accepting that disclosed gap, or splitting — landing the read-only detection now and holding the fix until the CI coverage exists. "Safe checks first" literally reads as the second option. We're putting that to the repo owner and will follow up here shortly; no rebase or changes needed from you in the meantime.

Two small things you could optionally fold in whenever you next touch it: the pre-scan prints "Removing orphaned row..." for candidates the atomic predicate may then spare — "Candidate:" wording would keep the output honest; and the enrichment text still advertises `--check=validate --fix` while the check now also runs in bare doctor.

maphew:
@vishnujayvel this went CONFLICTING against main a while back (last update 2026-07-28) and #4539 is otherwise unclaimed - do you plan to rebase? If you have moved on, happy to carry the rebase with your commits and credit intact; just say the word.

vishnujayvel:
Thanks for the nudge, and for offering to carry it — appreciated, but no need; I've rebased it.

The conflict was self-inflicted on my side: #5132 dropped `//go:build cgo` from the doctor
validation files and deleted the `checks_nocgo.go` stubs, and this branch predated that. The
rebase resolves it in the direction #5132 pointed:

- **Dropped** the `CheckOrphanedChildCounters` "Skipped: requires CGO" stub along with the rest
  of `checks_nocgo.go`. It was obsolete after #5132, and arguably wrong — the check only uses the
  server-mode Dolt client, so like its siblings it should run in pure-Go builds rather than
  reporting a cosmetic skip.
- Re-applied the check registration on top of the current wiring in `doctor.go`,
  `doctor_fix.go`, and `doctor/agent.go`.

The rebase also surfaced something worth calling out: #5142 made `openDoltDB` return the
config and established `guardFixTarget` as the convention *every destructive fix entry point
must follow*. `OrphanedChildCounters` is one, and predated it — so the fix path was missing the
project-identity check. Adopting the two-value call alone would have compiled cleanly while
silently skipping that guard, so I've wired it in to match the sibling fixes:

```go
if skip, err := guardFixTarget("Orphaned child counters fix", db, beadsDir, cfg); skip {
    return err
}
```

Net effect: same detection, now live for `CGO_ENABLED=0` operators, and the destructive half
now honours the identity guard the rest of `doctor --fix` does.

## On scope — happy to split this if you'd prefer

The PR currently carries both halves: the read-only detection and the destructive `--fix` that
deletes orphaned rows. If you'd rather land them separately, say so and I'll split it — detection
only now, `--fix` held back until there's Dolt-backed CI coverage exercising the delete path. I
don't have a way to run that coverage locally, and shipping an unexercised destructive path
against a live database is the kind of thing I'd rather you got to opt into than inherit. That
also matches the safe-checks-first posture the doctor surface already takes.

Either way is fine by me — whole thing or detection-first, your call.

## Test plan

Rebased onto `origin/main` @ `18c24a3ac`. ICU flags per the repo's CGO requirements.

- [x] `go build ./cmd/bd` — clean
- [x] `go vet ./cmd/bd/doctor/...` — clean
- [x] `go test ./cmd/bd/doctor/fix/` — `ok 14.097s` (whole package, incl. the sibling guarded fixes)
- [x] `go test ./cmd/bd/` — `ok 550.186s`
- [x] `go test ./cmd/bd/doctor/` — fails **only** on the two pre-existing tests below

**Two pre-existing failures, reproduced on unmodified `origin/main` today.** On a clean
`origin/main` @ `18c24a3ac` worktree, `TestCheckBeadsRole_NotConfigured` (`role_test.go:18`,
`:24`) and `TestCheckBeadsRole_NotGitRepo` (`role_test.go:104`) fail with identical assertions;
the other four `TestCheckBeadsRole_*` pass. A global `git config beads.role=maintainer` on this
machine leaks into them. Not introduced by this branch, and this branch touches nothing in that
code path.

**Limitation worth stating plainly:** Docker isn't available here, so the Dolt-backed tests
report `WARN: Docker not available, skipping Dolt tests`. The `--fix` delete path is therefore
covered by the local Dolt-server tests above but *not* by the Docker-gated suite — which is the
concrete reason the split offer above is on the table rather than rhetorical.

maphew:
Rebased and self-diagnosed — even better than a carry. Thanks for the quick turnaround; I'll leave #4539 to this PR then. Withdrawn on my side, and good luck with the checks.

vishnujayvel:
Checks are green now, so: all clear on that front.

Full CI passed on `692947be7` — including the `Test (Embedded Dolt Cmd */20)` shards where
`TestOrphanedChildCounters_FixDeletesOnlyOrphans` had been failing, plus `CI Gate / Required`.
The failure was environmental rather than a defect in the check: the test asked `dolt.New` for an
auto-started instance and swallowed the start error, so it ran on port 0 and the assertion tripped
downstream of the real cause. It now fails hard on that error instead of limping on.

Thanks again for the offer to carry the rebase — and for closing the loop rather than leaving it
ambiguous. Ready for review whenever you get to it.

maphew (commented):
Latest-delta review (`4a6d839a` through `692947be7`): the product-side identity guard is correct, but the tests leave two safety regressions unguarded.

**P1 - scrub the full Dolt endpoint environment before claiming local auto-start** (`cmd/bd/doctor/fix/validation_test.go:180`)

The test clears only legacy `BEADS_DOLT_PORT`, while `dolt.New` gives `BEADS_DOLT_SERVER_PORT` higher precedence and also honors ambient host, socket, server-mode, shared-server, and auto-start settings. That means this test can be redirected away from the per-test local server described in its comment. An unreachable ambient endpoint makes the test fail before exercising the fix; a reachable one can receive `CreateIfMissing` and schema/data writes because this package's `TestMain` sets `BEADS_TEST_SERVER=1`.

Please either use the standard isolated local-Dolt test helper or clear the complete endpoint/lifecycle override set before constructing the store, including both port spellings and the host/socket/server-mode/shared-server/auto-start overrides. This will make the test prove the path it says it proves and avoid touching an ambient Dolt endpoint. Reproduction with `BEADS_DOLT_SERVER_HOST=192.0.2.1 BEADS_DOLT_SERVER_PORT=59999 go test ./cmd/bd/doctor/fix -run '^TestOrphanedChildCounters_FixDeletesOnlyOrphans$' -count=1 -v` fails at `validation_test.go:216`; the same target passes with endpoint overrides unset.

**P2 - exercise the new guard against a mismatched project identity** (`cmd/bd/doctor/fix/remotes_test.go:157`)

The updated happy-path test supplies matching project IDs, but it still passes if the newly added `guardFixTarget` call is removed. `TestDestructiveFix_AbortsOnProjectIdentityMismatch` already enumerates destructive fix entry points and omits this exact one. Add `OrphanedChildCounters` to that table, ideally with an orphan row and an assertion that the row remains untouched, so a future removal of this safety guard fails a regression test.

Everything else in the latest delta checks out:

- `guardFixTarget` follows the sibling destructive-fix convention and runs before the orphan query or either delete.
- Changing the auto-start error from `Skipf` to `Fatalf` closes the prior silent-green coverage hole.
- The target test passes locally in a clean endpoint environment, and the PR head is green across the embedded Dolt shards and both required CI gates.

I am no longer a Beads maintainer, so this is a contributor review, not a maintainer approval or merge decision. @gastownhall/beads-maintainers, please take ownership of the finding's disposition and the PR's merge decision from here.

vishnujayvel:
Thanks @maphew — both findings reproduced before I touched anything, and both were real.

**P1.** Your repro fails exactly as you described:

```
validation_test.go:216: dolt.New with local auto-start: Dolt server unreachable 192.0.2.1:59999
--- FAIL: TestOrphanedChildCounters_FixDeletesOnlyOrphans (1.51s)
```

There is no standard isolated local-Dolt helper to reach for — I looked before hand-rolling. Every isolated-Dolt call site either uses the Docker container helpers (which themselves *set* `BEADS_DOLT_PORT`) or hand-rolls a partial scrub; the most complete existing list, `internal/storage/dolt/active_database_size_test.go:284`, has 7 vars and misses `BEADS_DOLT_AUTO_START`. So I added the helper your review assumed existed: `testutil.ClearAmbientDoltEnv(t)` in `internal/testutil/testdoltcommon.go`. One place for the list to grow, which is your fragility objection. I did not refactor the existing partial scrubs — that felt like a separate change.

Reviewing that helper turned up a **second route to the same hazard that neither of us had named**: `BEADS_DOLT_DATA_DIR`. The auto-start path is `EnsureRunningDetailed` → `Start` with `DataDir: ResolveDoltDir(beadsDir)` (`doltserver.go:910`), and `ResolveDoltDir` → `projectDoltDirPath` honours that variable first (`internal/doltserver/physical_root.go:62`). With it set to an empty scratch directory, the test **passed** while writing `.dolt/`, `.doltcfg/`, `.bd-dolt-ok` and its own database directory into that directory. Silent, unlike the endpoint case, which fails loudly. It is now the 9th entry, and the probe leaves the directory empty.

**P2.** Confirmed — with the `guardFixTarget` call deleted, the happy-path test stays green.

I added `OrphanedChildCounters` to the table as you asked, but I want to be straight that **I could not run it**: that suite is Docker-gated and skips here, and `fixTestServerPort()` reads a package-private var set only by `startDoltContainer()`, so a local `dolt sql-server` cannot be substituted. That entry is verified by construction and compilation only.

Given `bd-nxt5e` — every DB-backed test in this package skipping for months with nothing noticing — a table row nobody can run seemed like thin protection for a safety guard. So I also added `TestOrphanedChildCounters_AbortsOnProjectIdentityMismatch` on the same Docker-free auto-start path, with the orphan row present and the untouched assertion you asked for. That one is mutation-checked:

```
--- PASS: TestOrphanedChildCounters_AbortsOnProjectIdentityMismatch (9.60s)     [guard present]

  Removing orphaned child_counters row: parent_id=tst-missing-parent
  Fixed 1 orphaned child counter row(s)
    validation_test.go:419: OrphanedChildCounters should have aborted on project identity mismatch, got nil error
--- FAIL: TestOrphanedChildCounters_AbortsOnProjectIdentityMismatch (9.64s)     [guard deleted]
```

Happy to drop the extra test if you would rather carry only the table entry.

Two caveats worth stating rather than burying:

- The helper clears variables with `t.Setenv(name, "")`, which sets empty rather than unsetting. Every production reader of these nine is `os.Getenv` with a comparison where empty is equivalent to unset (`isFalsyBool("")` is `false`, so an emptied `BEADS_DOLT_AUTO_START` does not disable auto-start), so this is safe today — but it is a property of the readers, not of the technique.
- `golangci-lint` here is 2.12.2 against CI's pinned v2.10.1. Clean, but not a version-identical run.

The PR body has been updated — it previously never mentioned the project-identity guard at all, even though `4a6d839a` added it.

PR #4859, opened 2026-07-17
feat(storage): add extension-safe backend registry seam

julianknutsen:
## What

Adds a process-local, name-keyed registry for storage backends behind the
existing `storage.DoltStorage` seam.

- A downstream build can register read-write and read-only open functions plus
  its workspace-discovery behavior.
- `configfile` recognizes registered names through a standard-library-only leaf
  package, avoiding an import cycle into the typed registry.
- The root command path and both CGO and non-CGO factories dispatch registered
  names through the same contract.

The stock Beads build registers no alternate backend. Its existing Dolt
embedded, server, and proxied-server paths remain hand-wired and unchanged.

## Why

This preserves the useful registry and registration seam from the original
contribution while rebasing it onto the current Dolt-only `main`. Downstream
distributions can add a backend without extending shared backend-name
if-chains, while the OSS binary keeps its current storage policy.

## Deliberately not included

- No SQLite, PostgreSQL, or MySQL backend implementation or registrant.
- No generic provisioning hook, `--backend-opt`, or `backend_config` metadata.
- No schema, migration, dependency, CI-workflow, or generated-doc changes.
- Removed SQLite/PostgreSQL/MySQL names remain fail-closed tombstones in stock
  Beads; unknown names also continue to fail before a store opens.

## Verification

- `CGO_ENABLED=1 go test -tags gms_pure_go ./internal/storage/backends ./internal/configfile`
- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/backends ./internal/configfile`
- Focused registry and fail-closed `cmd/bd` tests in both CGO and non-CGO builds
- `go test -race` for the registry and config packages
- `make ci-pr-policy`
- `make ci-pr-lint` (native and Windows/non-CGO cross-lint)

The shared local host's full `make test` and `make ci-pr-core` runs also exposed
two pre-existing test-harness problems: order-dependent backup-config state and
unbounded host-process waits. Every implicated test passes in isolation; those
follow-ups are tracked as `bd-yrn` and `bd-9y7`. GitHub CI will provide the
clean-run result for this rebased head.

### DIFF
--- a/beads_cgo.go
+++ b/beads_cgo.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/steveyegge/beads/internal/configfile"
+	"github.com/steveyegge/beads/internal/storage/backends"
 	"github.com/steveyegge/beads/internal/storage/dolt"
 	"github.com/steveyegge/beads/internal/storage/embeddeddolt"
 )
@@ -31,6 +32,13 @@ func OpenBestAvailable(ctx context.Context, beadsDir string) (Storage, error) {
 	}
 	if !configfile.IsSupportedBackend(cfg.Backend) {
 		return nil, configuredBackendUnavailable(cfg.Backend)
+	}
+
+	// Dispatch to a registered extension backend before any Dolt path, mirroring
+	// the CLI store factories so SDK callers get the backend they registered
+	// instead of a silently-opened embedded Dolt store.
+	if backend, ok := backends.Lookup(cfg.GetBackend()); ok {
+		return backend.Open(ctx, beadsDir)
 	}
 
 	if cfg.IsDoltServerMode() {
--- a/beads_nocgo.go
+++ b/beads_nocgo.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/steveyegge/beads/internal/configfile"
+	"github.com/steveyegge/beads/internal/storage/backends"
 	"github.com/steveyegge/beads/internal/storage/dolt"
 )
 
@@ -25,6 +26,13 @@ func OpenBestAvailable(ctx context.Context, beadsDir string) (Storage, error) {
 	}
 	if !configfile.IsSupportedBackend(cfg.Backend) {
 		return nil, configuredBackendUnavailable(cfg.Backend)
+	}
+
+	// Dispatch to a registered extension backend before any Dolt path, mirroring
+	// the CLI store factories so SDK callers get the backend they registered
+	// instead of the embedded-Dolt-requires-CGO error.
+	if backend, ok := backends.Lookup(cfg.GetBackend()); ok {
+		return backend.Open(ctx, beadsDir)
 	}
 
 	if cfg.IsDoltServerMode() {
--- a/cmd/bd/backend_registry_contract_test.go
+++ b/cmd/bd/backend_registry_contract_test.go
@@ -1,0 +1,85 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/configfile"
+	"github.com/steveyegge/beads/internal/storage"
+	"github.com/steveyegge/beads/internal/storage/backends"
+)
+
+var (
+	errRegistryReadWrite = errors.New("registry read-write open")
+	errRegistryReadOnly  = errors.New("registry read-only open")
+)
+
+func registerContractBackend(t *testing.T, name string) {
+	t.Helper()
+	backends.Register(name, backends.Backend{
+		Open: func(context.Context, string) (storage.DoltStorage, error) {
+			return nil, errRegistryReadWrite
+		},
+		OpenReadOnly: func(context.Context, string) (storage.DoltStorage, error) {
+			return nil, errRegistryReadOnly
+		},
+		WorkspaceIsBeadsDir: true,
+	})
+	t.Cleanup(func() { backends.Deregister(name) })
+}
+
+func writeContractBackendConfig(t *testing.T, backend string) string {
+	t.Helper()
+	beadsDir := t.TempDir()
+	if err := (&configfile.Config{Backend: backend}).Save(beadsDir); err != nil {
+		t.Fatalf("save metadata.json: %v", err)
+	}
+	return beadsDir
+}
+
+func TestRegisteredBackendDispatchesReadWriteAndReadOnly(t *testing.T) {
+	const name = "contract"
+	registerContractBackend(t, name)
+	beadsDir := writeContractBackendConfig(t, name)
+
+	if err := validateConfiguredBackend(&configfile.Config{Backend: name}); err != nil {
+		t.Fatalf("validateConfiguredBackend() rejected registered backend: %v", err)
+	}
+	if _, err := newDoltStoreFromConfig(t.Context(), beadsDir); !errors.Is(err, errRegistryReadWrite) {
+		t.Fatalf("read-write factory error = %v, want %v", err, errRegistryReadWrite)
+	}
+	if _, err := newReadOnlyStoreFromConfig(t.Context(), beadsDir); !errors.Is(err, errRegistryReadOnly) {
+		t.Fatalf("read-only factory error = %v, want %v", err, errRegistryReadOnly)
+	}
+}
+
+func TestRegisteredBackendDrivesWorkspaceDiscovery(t *testing.T) {
+	const name = "contract-discovery"
+	registerContractBackend(t, name)
+
+	if !registeredBackendWorkspaceIsBeadsDir(&configfile.Config{Backend: name}) {
+		t.Fatal("registered backend did not expose its .beads workspace")
+	}
+	if registeredBackendWorkspaceIsBeadsDir(&configfile.Config{Backend: "unregistered"}) {
+		t.Fatal("unregistered backend exposed a .beads workspace")
+	}
+	if registeredBackendWorkspaceIsBeadsDir(&configfile.Config{Backend: configfile.BackendDolt}) {
+		t.Fatal("Dolt must retain its existing database discovery path")
+	}
+}
+
+func TestOSSRegistersNoRemovedBackends(t *testing.T) {
+	for _, name := range []string{
+		configfile.BackendPostgres,
+		configfile.BackendMySQL,
+		configfile.BackendSQLite,
+	} {
+		if backends.Registered(name) {
+			t.Errorf("OSS unexpectedly registered removed backend %q", name)
+		}
+		if err := validateConfiguredBackend(&configfile.Config{Backend: name}); err == nil {
+			t.Errorf("OSS unexpectedly accepted removed backend %q", name)
+		}
+	}
+}
--- a/cmd/bd/backend_support.go
+++ b/cmd/bd/backend_support.go
@@ -30,8 +30,7 @@ func registeredBackendWorkspaceIsBeadsDir(cfg *configfile.Config) bool {
 	if cfg == nil {
 		return false
 	}
-	backend, ok := backends.Lookup(cfg.GetBackend())
-	return ok && backend.WorkspaceIsBeadsDir
+	return backends.WorkspaceIsBeadsDir(cfg.GetBackend())
 }
 
 func requireDoltBackend(cfg *configfile.Config) error {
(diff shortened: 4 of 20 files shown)


maphew (commented):
Cross-vendor paired agent review (Claude Opus + Codex gpt-5.6-sol, independent traces, adjudicated before posting). Verdict: **needs changes — but really a direction call for you** rather than a fix list, since main has moved since this branched.

**The dominant finding (both reviewers converged): the PR is superseded by main's own landed direction.** Merge-base is from 2026-07-16; one day later main landed 0d7cf0022 (roll back the direct PostgreSQL/MySQL adapters) and 6fd9dbda7 via #4881 (remove the SQLite backend and dialect), then replaced backend *selection* with the fail-closed tombstone policy in `cmd/bd/backend_support.go` (`RemovedBackendError` / `UnknownBackendError`), wired through the factory, init, doctor, bootstrap, and dolt paths. This PR registers exactly the three deleted backends (`cmd/bd/backends_builtin.go` blank-imports `internal/storage/{sqlite,postgres,mysql}` — packages that no longer exist), and its unknown-name → silent Dolt fallback is the opposite of main's deliberate hard-error. The GitHub CONFLICTING flag understates it: `backend_support.go` is an add/add with different purposes, `sqlite/fromconfig.go` is modify/delete, and even a textually resolved merge cannot compile. So the question isn't "rebase?" — it's whether the registry seam should land at all in a Dolt-only, fail-closed world, and that's yours to answer as the owner of the storage-boundary direction (#4547 Route A).

**What's worth salvaging regardless (both reviewers agree the engineering is good):** the `internal/storage/backendnames` leaf package cleanly breaks the `configfile → backends` import cycle; `Register` panics on empty/reserved/duplicate names; dispatch moved to build-agnostic `store_factory_config.go` preserving the CGO/nocgo split (nocgo build verified green locally). If Route A eventually wants a registry, this scaffolding — with zero registrants and reconciled against the tombstone gate instead of the silent-fallback — is the piece to carry forward.

**Base-level defects found while tracing (relevant only if any of this is revived):** plain `bd init` can orphan a registered *custom* backend — `checkExistingSQLBackend` guards only the three literal names, so a registry-recognized custom backend falls through and init overwrites `cfg.Backend` without `--reinit-local`, contradicting the fail-closed invariant documented above the guard; the registrants map `OpenReadOnly` to `NewFromConfig`, which provisions/`InitSchema`s — so read-classified commands would run DDL at open; the claimed no-CGO end-to-end test never executes in CI (workflow regexes select `RegisteredBackend.*` but the test is named `TestRegistryBackendEndToEndSQLite`); and the SQLite init rollback restores `metadata.json` but leaves the created database behind, so the retry the comment promises is blocked by the existing-file guard.

Suggested disposition: close or convert to draft pending the #4547 direction decision, extracting `backendnames` + the registry skeleton into a fresh branch if/when a registry is wanted on top of the tombstone policy.

julianknutsen:
Looking at this now

julianknutsen:
Thanks, @julianknutsen — the registry's leaf name-classification package and
build-agnostic registration contract are the valuable core of this
contribution, and those are now preserved on current `main`.

I rebased and force-pushed the PR branch as commit `be3a85240`. The rewritten
diff keeps registry validation, read-write/read-only dispatch, workspace
discovery, CGO/non-CGO coverage, and contributor attribution. It removes the
SQLite/PostgreSQL/MySQL implementations and registrants, generic provisioning
and init flags, backend-specific metadata, and their CI/docs changes. Stock
Beads therefore remains Dolt-only and fail-closed for removed or unknown
backend names.

Local focused tests, race tests, PR policy, formatting, native lint, and the
Windows/non-CGO cross-lint pass. The full shared-host suites encountered two
pre-existing harness issues; the affected tests pass in isolation and are
tracked separately as `bd-yrn` and `bd-9y7`. GitHub CI is now running against
the clean rebased head.

_codex-gpt-5-unknown-reasoning on behalf of CI Bot_

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98637017

maphew (commented):
The fail-closed discipline in this rewrite is the part I want to call out first: `requireDoltBackend` in `internal/storage/dolt/open.go` still re-checks `GetBackend() != BackendDolt` *after* `IsSupportedBackend`, so a registered name can never be reinterpreted as a Dolt workspace, and `autoMigrateOnVersionBump` already skips non-Dolt backends. Those are exactly the two places where a naive registry seam would have silently opened an empty Dolt database, and both are closed. The `Register` contract (panic on empty/reserved/incomplete/duplicate) is the right shape for process-start wiring, and `Deregister` being scoped to tests keeps it from becoming a runtime knob.

Thanks for taking the direction call and re-cutting this to a narrow, Dolt-only-preserving seam — that was the harder path and it is the right one.

I verified locally rather than reading the diff alone: `CGO_ENABLED=1` and `CGO_ENABLED=0` builds of `./...` are clean, `./internal/storage/backends/... ./internal/configfile/...` and the focused `cmd/bd` contract tests pass, and the contract test is load-bearing — removing the `backends.Lookup` dispatch from `newDoltStoreFromConfig` makes `TestRegisteredBackendDispatchesReadWriteAndReadOnly` fail, and it fails by *succeeding* to open a Dolt store, which is precisely the false-empty case this whole design is built to prevent. Nothing here leaks orchestration policy into core, adds no `dolthub/` imports outside storage, and touches no schema. Design fit is good.

Three findings. None of them are reachable in the stock binary (nothing registers), so none block main's health — but each one is a trap laid for the exact downstream this seam serves, and all three are cheap.

**1. `bd init --backend=<registered>` provisions a Dolt workspace anyway — `cmd/bd/init.go:419`** (should-fix)

The guard is `if !configfile.IsSupportedBackend(backendFlag)`, which now accepts registered names. Init then falls through to the Dolt provisioning path and writes `"backend": "<name>"` into `metadata.json` over a freshly created embedded Dolt database. Since this PR deliberately ships no provisioning hook, the only correct outcome is a hard failure. Suggested fix — gate init on the literal Dolt set rather than the widened predicate:

```go
if backendFlag != "" && backendFlag != configfile.BackendDolt {
    if backends.Registered(backendFlag) {
        return fmt.Errorf("backend %q is registered for opening existing workspaces but has no provisioning path; create the workspace with its own tooling", backendFlag)
    }
    // existing tombstone / unknown switch
}
```

**2. Public `OpenBestAvailable` widens without dispatching — `beads_cgo.go:32`, `beads_nocgo.go:26`** (should-fix)

Both call `configfile.IsSupportedBackend(cfg.Backend)` (now true for a registered name) and then unconditionally open embedded Dolt (CGO) or return "embedded Dolt requires CGO; use server mode" (non-CGO). This is the public API extenders call, so a registered backend gets a wrong-store open or a misleading error. Either mirror the factory dispatch:

```go
if backend, ok := backends.Lookup(cfg.GetBackend()); ok {
    return backend.Open(ctx, beadsDir)
}
```

or, if you'd rather keep the public API Dolt-only for now, fail closed with an explicit "registered backend %q is not reachable through OpenBestAvailable" — but don't let it reach `embeddeddolt.Open`.

**3. `internal/storage/backendnames` is not needed for the reason it states** (should-fix, and it deletes code)

The package doc says it exists so `configfile` can classify names "without importing the typed backend registry", avoiding an import cycle. There is no cycle to avoid: `internal/storage` does not depend on `internal/configfile` (`go list -deps ./internal/storage` yields only `types`, `hooks`, `storage`). I changed `configfile` to import `internal/storage/backends` directly and call `backends.Registered` — `go build ./...` is clean under both `CGO_ENABLED=1` and `CGO_ENABLED=0 -tags gms_pure_go`, and the `backends` and `configfile` test packages both still pass (the external `backends_test` package importing `configfile` is fine).

Dropping it removes 36 lines, a second mutex, and the `Register`/`Deregister` window where the two maps disagree. If the actual motive is dependency weight — keeping `configfile`, a widely imported leaf, from pulling in `storage`+`hooks`+`types` — that's a legitimate reason and I'd keep the package; just fix the doc comment, because "import cycle" sends the next reader looking for something that isn't there.

Smaller items, take or leave:

- `cmd/bd/bootstrap.go:282` — `requireBootstrapDoltBackend` is named for Dolt but only calls `validateConfiguredBackend`, which now returns nil for a registered backend; bootstrap then runs Dolt database detection. Either rename it or point it at `requireDoltBackend`.
- `internal/storage/backends/backends.go` — the reserved name is the literal `"dolt"`; since there's no cycle (see #3 above), `configfile.BackendDolt` works.
- No listing accessor, so messages like init's `the supported backend is "dolt" (default)` and `bd doctor` output become untrue in a downstream build. A `backends.Names()` would let them tell the truth.
- `TestOSSRegistersNoRemovedBackends` checks only the three tombstones; asserting the registry is empty outright is a stronger guard.
- The head commit message says "Salvage the name-keyed extension contract from PR #4859" while being the head of #4859 — name the pre-rebase commit instead. Its `Agent-Signature` trailer also carries `unknown-reasoning` and `on behalf of CI Bot`.
- Branch is 38 commits behind main. Checks are all green and the merge state is clean, but please `gh pr update-branch` and let CI re-run before this goes to the merge queue.

One coordination note, not a code finding: #4561 proposes an overlapping provider registry, and it was told it had precedence on this seam. That PR is much broader (external plugin process, capability modeling, public `OpenConfigured`) and is held on a separate exec-trust blocker, so I don't read this as competing — but I'll leave a note there explaining how the narrow in-process seam relates before this merges, so nobody discovers it by rebase conflict.

Happy to push fixes for (1) and (2) as maintainer commits on this branch if you'd prefer — or they're yours if you'd rather take them. Just say which.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98637619

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds an extension-safe backend registry seam that lets downstream builds register additional storage backends by name, which gives the project a clean, additive extension point for storage selection without forking the store-open path — and it stays a no-op for the shipped OSS binary, where Dolt remains the only provisioned backend.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- `bd bootstrap` now fails closed on registered (non-Dolt) backends: it rejects them before metadata repair, action detection, or plan execution, and re-checks the guard inside `executeBootstrapPlan`, so a registered backend can never be provisioned through bootstrap.
- `bd init` accepts registered names for classification but rejects *creation*, returning open-existing-workspace-only guidance, so registered backends stay open/discover-only rather than being newly created.
- `OpenBestAvailable` dispatches registered backends ahead of the Dolt/server paths in both the CGO and non-CGO builds, so the public open path routes to the right backend.
- Workspace discovery returns the `.beads` directory for registered backends whose workspace *is* the `.beads` directory, so discovery no longer mis-derives their location.
- Hook/molecule path resolution reuses the resolved `.beads` directory instead of deriving it from `filepath.Dir(dbPath)`, so registered-workspace layouts resolve correctly.

Review findings addressed:
- Architectural Consistency (minor, high confidence): the review noted that `requireDoltBackend` and `requireBootstrapDoltBackend` encode the same fail-closed invariant in two places and could drift over time. This is non-gating — the final patch keeps both guards correct and consistent today; consolidating or cross-referencing them is an optional post-merge cleanup.

Remaining non-gating notes:
- Only the maintainability note above remains, and it is non-gating. No blocking or major issues were found: the seam is additive, gated behind registered-backend predicates, and revertible.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch: `fix(storage): dispatch and fail-closed for registered backends` (12 files changed, +473/-7). This commit adds the registered-backend dispatch in `OpenBestAvailable`, the fail-closed guards in `bd bootstrap` and `bd init`, workspace discovery and hook/molecule path resolution for registered workspaces, plus dedicated tests. Your original registry-seam commit is preserved.

### Final Review Status
Ready for the merge queue: final head `c38dbc1d1` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/30620139623

### Review Iterations
4 review passes performed; the change reached an approve decision after maintainer fixes closed the bootstrap, init, dispatch, discovery, and path-resolution gaps above.

---

PR #4862, opened 2026-07-17
fix(claim): serialize concurrent claims with a per-database advisory lock

itsandyking:
## The bug
`ClaimIssue` / `ClaimReadyIssue` are a correct compare-and-swap — `UPDATE issues SET status='in_progress', assignee=? WHERE id=? AND status='open' AND assignee IN ('',NULL,actor)`, then a `RowsAffected` check. But under Dolt's snapshot isolation, each concurrent claim transaction evaluates that `WHERE` against its **own start snapshot**, so N racers all still see `status='open'`, all get `RowsAffected==1`, and all commit. The CAS is silently defeated and **one issue is handed to multiple callers**. Any setup running N processes against a shared Dolt server (e.g. a fan-out of agents doing `bd ready --claim`) double-claims work.

## Measured
Against a Dolt 2.1.10 `sql-server`, 6 concurrent claims of one open issue:

| approach | winners (correct = 1) |
|---|---|
| baseline (today) | **6 / 6** |
| `SELECT … FOR UPDATE` | **6 / 6** (Dolt does not serialize on it) |
| `GET_LOCK()` | **1 / 6** ✅ |

So the usual `FOR UPDATE` fix does **not** work under Dolt — a named advisory lock does.

## The fix
Hold a MySQL named advisory lock (`GET_LOCK`) for the duration of each claim:
- The lock **and** the claim transaction run on **one pinned `*sql.Conn`** (`conn.BeginTx`), so it works under `SetMaxOpenConns(1)`.
- The lock name is **per-database** (`bd_claim_<db>`) so claims in different repos on a shared server never block each other.
- Release confirms `RELEASE_LOCK==1` then returns the connection to the pool; on any release failure it physically drops the driver session (`conn.Raw` + `driver.ErrBadConn`, then `Close`) so a failed release can never leave the lock held and deadlock later claims.

## Alternatives considered
- **`SELECT … FOR UPDATE`** — the standard fix, and the first thing most reviewers will reach for. Dolt evaluates the predicate against each transaction's start snapshot rather than taking a serializing row lock, so it does not serialize concurrent claimants (measured 6 / 6 above). Kept in the table precisely to pre-empt that suggestion.
- **Dolt commit-time conflict detection (optimistic retry)** — the most Dolt-idiomatic option: let both CAS updates run and rely on the second `COMMIT` failing on a same-cell write-write conflict, then retry. Not dependable here — whether a same-cell conflict errors vs. auto-merges depends on transaction/conflict configuration, and in the baseline the concurrent claims committed silently (6 / 6). Making it reliable would need explicit multi-statement transactions with conflict-on-commit enabled plus a bounded retry loop: more surface area and more Dolt-version coupling than a single advisory lock.
- **`GET_LOCK` is not a MySQL-only bolt-on** — it is implemented natively by go-mysql-server (the engine under Dolt) as a session-scoped named lock held in the server, independent of the versioned storage and snapshot isolation. That orthogonality is exactly why it serializes where row locks do not, and it behaves identically whether `bd` is pointed at Dolt or upstream MySQL — so the fix is portable across both backends rather than Dolt-specific.

## Verification
- Existing `Ready|Claim|Proxied|Update` CLI suite: **pass** (embedded-Dolt, `gms_pure_go`).
- **End-to-end A/B on a real Dolt server**, 12 concurrent claimers over 3 beads × 4 rounds: current `bd` double-claims every round (48 "successes" for 12 real slots); the patched `bd` claims each bead **exactly once** (12 successes total, **0 double-claims**), no deadlock.
- Reviewed adversarially by two independent models, which drove fixes for an unbounded release context, a pooled-session lock leak, a `MaxOpenConns=1` self-deadlock, and an under-specified release error path before it passed.

### DIFF
--- a/internal/storage/dolt/issues.go
+++ b/internal/storage/dolt/issues.go
@@ -3,8 +3,10 @@ package dolt
 import (
 	"context"
 	"database/sql"
+	"database/sql/driver"
 	"encoding/json"
 	"fmt"
+	"io"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


maphew (commented):
Cross-vendor paired agent review (Claude Opus + Codex gpt-5.6-sol, independent traces, adjudicated against current main before posting). Verdict: **needs changes — and probably retire in favor of what main has since shipped.** The diagnosis in this PR was valuable; the mechanism has been overtaken.

**The decisive fact (verified on current main):** since this PR branched, main rewrote both claim paths to wrap in `withRetryTx` **and** to rewrite a shared `row_lock` fence inside every claim (`internal/storage/issueops/claim.go` — "a concurrent reclaim or close on the same row is forced to conflict rather than silently cell-merge"). That fence is a direct answer to this PR's core concern ("whether a same-cell conflict errors vs auto-merges depends on transaction/conflict configuration"): main no longer relies on incidental cell overlap — it manufactures a guaranteed same-cell write so the loser gets 1213/1205 and retries against a fresh snapshot. `TestConcurrentWorkQueueDrain` pins exactly-once claims. The GitHub CONFLICTING flag reflects this: both claim-method bodies conflict semantically (lock-based vs fence+retry), not mechanically. Your measured 6-on-1 double-claim repro predates the fence; if you can re-run that harness against current main and it still double-claims, that would be a headline finding worth its own issue — but the burden has shifted.

**Even on its own terms, the paired traces found gaps worth recording:** (1) wisp claims bypass the lock — `ClaimIssue` returns through `claimWisp` before `acquireClaimLock`, and `ClaimReadyIssue` can select wisps, so a locked ready-claim races an unlocked direct wisp claim; (2) the per-database lock name has no 64-byte guard — the in-tree `internal/storage/schema/lock.go` helper (`MigrationLockName` fnv-hashes long names, plus `discardConn`) already solves this and should have been reused rather than re-implemented (~50 duplicated lines); (3) the coarse per-database lock held across `DOLT_COMMIT` serializes *all* claims — a real throughput regression vs optimistic retry under fleet fan-out; (4) `GET_LOCK` timeout surfaces as a hard, non-retryable error to the caller; (5) beads-side `GET_LOCK`/`Conn.Raw` session surgery in the store sits on the wrong side of the storage-boundary charter — if an advisory lock were ever needed, it belongs behind the driver interface. Credit where due: the lock lifecycle itself is careful — deferred release with a fresh bounded context, `driver.ErrBadConn` teardown on failed release so a pooled session can't retain the lock.

**Suggested disposition:** close with thanks (or repurpose as the re-measurement issue against the fence), preserving the 6-on-1 harness and the GET_LOCK-vs-FOR-UPDATE measurements — that diagnostic work is the durable value here, and it's exactly what should gate any future revisit.

maphew:
Closing this one with real thanks — you isolated a genuine defect class (double-claims under Dolt snapshot isolation) and brought quantitative evidence with the 6-concurrent-claims harness (baseline 6/6 double-claim; `SELECT ... FOR UPDATE` also 6/6; `GET_LOCK` down to 1/6). That's exactly the kind of report that moves things.

The reason for closure rather than merge: upstream main has since landed a different fix for the same root cause. `--claim` now goes through a conditional-UPDATE CAS whose `row_lock` cell is rewritten with every claim, forcing concurrent claim-family writes on the same row to collide at Dolt commit time instead of silently cell-merging; the loser retries and its CAS predicate then fails deterministically (`internal/storage/issueops/claim.go`, regression-tested for exactly-one-winner in `internal/storage/dolt/concurrent_test.go` — `TestConcurrentWorkQueueDrain`). The recently merged #5006 (claim verify-by-re-read under degraded servers) and #5008 (`--if-assignee`/`--if-status` guards) harden the same path from other angles. An advisory lock on top would serialize claims that the commit-conflict mechanism already resolves, and this was confirmed by an independent cross-vendor review before closing.

One ask, because your harness is valuable: if you re-run the 6-concurrent-claims test against current main and still see more than one winner with **distinct** actors (same-actor re-claim is documented idempotent success, so it will always report "multiple successes" by design), please open an issue with the numbers — that would be a real regression and we'd want it loudly.

PR #4863, opened 2026-07-17
feat(leases): ephemeral dolt_ignored leases table — heartbeats mint zero commits (bd-lrgn1)

steveyegge:
## Summary

Moves claim leases (`holder`, `granted_at`, `lease_expires_at`, `heartbeat_at`) off the versioned `issues` table into a new `leases` table registered in `dolt_ignore` (precedent: `wisps`). Claims and heartbeats no longer mint Dolt commits or history: a claim commits exactly once (the history-worthy status/assignee transition) and heartbeats commit nothing. At fleet scale this removes ~13k coordination commits/day of pure chatter — the dominant source of unbounded reachable history and of the write pressure that starves large catch-up merges (bd-lrgn1; evidence from the wyvern deployment).

Leases are deliberately node-local (dolt_ignored tables don't replicate) — matching what leases already were in reality: only enforceable on the replica that granted them. Cross-machine claim visibility still rides `status`/`assignee` on `issues`. Bonus: heartbeats stop stamping `issues.updated_at`, so `updated_at` regains its merge/LWW meaning.

## Design

- `leases` table on all five backends: dolt via migration 0055 + ignored/0012 (fresh clones), sqlite/mysql `schema.sql`, postgres `schema.go`. Always-on, not a config mode (no dual query paths).
- Invariant: a lease row exists iff its issue is a live claim on this node. Every path ending or transferring a claim (close, unclaim, reclaim, delete, status/assignee update, import accepting a newer non-claimed snapshot) deletes the row.
- All full-issue query sites hydrate via `sqlbuild.LeaseJoin` (LEFT JOIN); a missed site fails loudly at PREPARE.
- `row_lock` STAYS on issues — it is the general write-serialization cell, not lease state. Wisps are never leased.
- Protocol L1.2 preserved: lease fields still round-trip JSONL via export-time JOIN; import restores into `leases` but never clobbers a live (unexpired) local lease (`RestoreLeaseOnImportInTx` + new test).
- Migration 0055 moves live lease values from the old columns, then drops them (single combined ALTER per table); guarded for re-runs and partial states. Old binaries against a migrated DB fail loudly (columns gone), never corrupt.

## Fixes surfaced while validating

- Both `pgdialect` and `sqlitedialect` needed the `leases → issue_id` ON CONFLICT arbiter entry — without it every claim upsert on sqlite/postgres failed at PREPARE (translator tests added for both upsert forms).
- sqlite/mysql schema loaders split DDL on `;` before stripping comments — the leases DDL comment contained one and broke every subsequent statement.
- uow `initSchema` waiter budget 15s → 60s: it must outwait a peer's full cold-start migration pass (grows with every migration), not a transient blip. Same reasoning for the dolt package's `testTimeout` 30s → 45s.

## Acceptance (from bd-lrgn1)

1. ✅ Claim + N heartbeats → zero new dolt commits (`TestHeartbeatMintsNoDoltCommits`, asserts via dolt_log count)
2. ✅ status/assignee transitions still commit
3. ✅ Migration test from the column schema with live leases preserved (`TestMigration0055MovesLiveLeases`)
4. ✅ Lease protocol tests green in updated form (`cmd/bd/protocol`, incl. new never-clobber-live-lease test)
5. ✅ Reclaim of an expired lease across a server restart (`TestReclaimExpiredLeaseSurvivesRestartEmbedded` — real embedded-engine close/reopen; lease row durability asserted explicitly)
6. ✅ `./scripts/conformance.sh` fully green (both tiers); `make test` green except two failures reproduced identically on main (internal/audit hang and cmd/bd 3m package timeout — environmental, see below)

## Testing notes

Full local `internal/storage/dolt` package runs crawl on the dev machine for **both this branch and origin/main** with the identical failure signature (store init deadline; long-lived shared test container degradation, consistent with the dolt GC connection-slot leak dolthub/dolt#11196) — all lease/claim/migration tests pass in isolated and targeted runs; CI is the arbiter for the full package. Rebased onto current main (post backend-revert + TLS proxy changes); build/vet/lint clean (one pre-existing gosec finding untouched by this branch).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -838,9 +838,10 @@ jobs:
           BEADS_TEST_EMBEDDED_DOLT: "1"
           BEADS_TEST_EMBEDDED_TEST_BINARY: /tmp/embeddeddolt-test
         # Conformance runs in the dedicated test-embedded-conformance job
-        # below (mirrors pr-risk.yml); skip it here so this job stays inside
-        # its 20m envelope as the base embedded suite grows.
-        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=20m -test.skip '^TestConformance$'
+        # below (mirrors pr-risk.yml); skip it here. The budget must cover
+        # every store-open re-running the full migration chain, which grows
+        # with each migration (bd-lrgn1 pushed a slow runner past 20m).
+        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=30m -test.skip '^TestConformance$'
 
   test-embedded-conformance:
     name: Test (Embedded Dolt Conformance)
--- a/.github/workflows/pr-risk.yml
+++ b/.github/workflows/pr-risk.yml
@@ -169,9 +169,10 @@ jobs:
           BEADS_TEST_EMBEDDED_DOLT: "1"
           BEADS_TEST_EMBEDDED_TEST_BINARY: /tmp/embeddeddolt-test
         # Conformance runs in the dedicated test-embedded-conformance job; skip
-        # it here. Keeps upstream's 20m bump (base embedded tests grew) while
-        # this job stays inside its envelope without the conformance suite.
-        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=20m -test.skip '^TestConformance$'
+        # it here. The budget must cover every store-open re-running the full
+        # migration chain, which grows with each migration (bd-lrgn1 pushed a
+        # slow runner past the old 20m).
+        run: /tmp/embeddeddolt-test -test.v -test.count=1 -test.timeout=30m -test.skip '^TestConformance$'
 
   test-embedded-conformance:
     name: Test (Embedded Dolt Conformance)
--- a/cmd/bd/doctor/dolt.go
+++ b/cmd/bd/doctor/dolt.go
@@ -288,7 +288,7 @@ func checkSchemaWithDB(conn *doltConn) DoctorCheck {
 	// Check dolt_ignore'd tables — these only exist in the working set and
 	// must be recreated each server session. (GH#2271)
 	ignoredTables := []string{
-		"local_metadata", "repo_mtimes",
+		"leases", "local_metadata", "repo_mtimes",
 		"wisps", "wisp_labels", "wisp_dependencies", "wisp_events", "wisp_comments",
 	}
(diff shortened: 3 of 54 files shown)


steveyegge (commented):
Fable review (wyvern constellation, requested via mail m-1784278297-938a; independent of the authoring session — formal GitHub approve impossible, single shared account). Line-reviewed the full diff with the wyvern shared-server deployment as the reference case. VERDICT: APPROVE, merging.

Verified in the diff:
- Migration 0055: idempotent on re-run/partial state (temp+rename dance, per-clause guards), live-lease carryover scoped exactly right (in_progress + non-empty assignee + non-null expiry), INSERT IGNORE so earlier partial runs win, single combined ALTER, down migration present. Old binaries fail loudly post-drop — correct fail-closed posture for a mixed-binary fleet, and the three upgrade-smoke CI jobs validate it.
- Claim atomicity: lease upsert rides the claim tx; no claimed-without-lease window. Wisps excluded.
- Heartbeat: leases-table-only write (zero-commit property pinned by TestHeartbeatMintsNoDoltCommits via dolt_log count, not just asserted); the no-lease-row disambiguation that re-arms lease semantics for a hand-doled claim (bd-9hpgf) quietly fixes a previously-unreclaimable class. Nice.
- Reclaim: expiry re-check inside the lease DELETE honors a racing rescue; issue revert re-checks in_progress; row_lock retained where it still matters (status/assignee races); heartbeat-vs-reclaim now contend on the same lease row so the old cell-merge hazard is structurally gone. Grace = 2xTTL documented.
- Invariant coverage: DeleteLeaseInTx on every claim-ending path I could enumerate (close, unclaim, delete, generic status/assignee update, bulk ops, import reconcile).
- Import/export: L1.2 round-trip via export-time JOIN; RestoreLeaseOnImportInTx never clobbers an unexpired local lease (the ON DUPLICATE KEY assignment-ordering trick is correct MySQL semantics — that ordering comment is load-bearing, keep it); orphaned lease rows reconciled against the stored row, not the snapshot.
- stale.go: heartbeat-aware staleness with correct placeholder ordering.

Deployment notes for the cutover (not blockers): (1) resume order is binaries-first — any pre-0055 binary errors hard against the migrated DB by design; (2) updated_at regains LWW meaning — anything that treated updated_at as worker liveness must switch to lease/heartbeat signals (bd stale already has).

PR #4864, opened 2026-07-17
docs(recovery): history-squash runbook (bd-okqpv / wy-7lmdp)

steveyegge:
New recovery runbook: shedding reachable Dolt history that `dolt gc` cannot reclaim, via a fenced-window squash (backup → soft-reset squash to a baseline commit → drop stale refs → `gc --full` → force re-publish → re-clone everywhere else). Follows the house runbook format (Symptoms / Diagnosis / Solution ≤5 steps / Prevention); nav + recovery index updated.

Motivated by the wyvern deployment's 3.6G of reachable lease-churn history (wy-7lmdp); generic to any workspace with high-frequency-write history bloat. Companion to PR #4863, which removes the churn source going forward.

Docs-only. Gates: `go test ./test/docsync` and `check-doc-freshness.sh` green.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -271,6 +271,7 @@
           "recovery/merge-conflicts",
           "recovery/circular-dependencies",
           "recovery/sync-failures",
+          "recovery/history-squash",
           "recovery/uninstalling"
         ]
       },
--- a/docs/recovery/history-squash.md
+++ b/docs/recovery/history-squash.md
@@ -95,6 +95,16 @@ bd backup remove && bd backup init <path>   # fresh destination, then:
 bd backup sync
 ```
 
+<Warning>
+A Dolt remote accumulates chunks monotonically: the force-push re-points the
+remote's refs at the squashed chain but deletes nothing, so the *remote's*
+storage does not shrink. To reclaim the published side too, replace the
+remote — clear its storage (or pick a fresh path/prefix) before the push:
+`bd dolt remote remove <name>`, `bd dolt remote add <name> <fresh-url>`,
+then `bd dolt push --force`. Every other clone must re-clone after a squash
+regardless, so replacing the remote costs nothing extra.
+</Warning>
+
 **Step 5:** Verify, then re-clone everywhere else. On this machine:
 
 ```bash
--- a/docs/recovery/history-squash.md
+++ b/docs/recovery/history-squash.md
@@ -1,0 +1,103 @@
+---
+title: History Bloat
+description: Shed reachable Dolt history that dolt gc cannot reclaim
+---
+
+Every bead write mints a Dolt commit, and Dolt keeps every byte a reachable
+commit references — `dolt gc` only reclaims what nothing points to. A
+workspace that has accumulated months of high-frequency writes can grow far
+past its live data (gigabytes of storage for a few thousand beads) while
+`dolt gc` reclaims nothing, because the entire chain is still reachable from
+your branch. This runbook squashes that chain to a single baseline commit so
+the old history becomes collectable, without touching your live data.
+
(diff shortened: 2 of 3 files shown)


steveyegge (commented):
Fable review (companion to my #4863 review; requested via m-1784302036-7ba8). Runbook is well-shaped — correct squash recipe (reset --soft to root + add -A + commit), server-stopped fencing, dolt-native backup as rollback, honest re-clone warning, and the Prevention section is accurate post-#4863. ONE BLOCKING GAP + two nits before merge:

BLOCKING — remote-tracking refs defeat the gc as written. Step 3 deletes stale local branches and tags, then runs `dolt gc --full`; Step 4 force-pushes AFTER. But any workspace that has ever pushed/fetched carries remote-tracking refs (`dolt branch -r`) still pointing at the pre-squash chain, and gc treats those as reachable — so on exactly the workspaces this runbook targets (big, long-lived, remoted), Step 3's gc reclaims little or nothing and the operator has no cue that it silently failed. Fix: in Step 3, also delete or re-point remote-tracking refs (`dolt branch -rd <remote>/<branch>`), or reorder so `dolt gc --full` runs AFTER Step 4's force-push has moved the remote refs — and either way add a verification line (`du -sh .` again) so the operator sees the space actually come back. If dolt's gc provably ignores remote refs, say so explicitly in the doc instead — the reader needs one sentence either way.

Nit 1: Step 2's `dolt log --oneline | tail -1` assumes the root commit is the desired baseline parent; a one-line note on WHY root (keeps a valid chain with a single ancestor) would stop operators from "simplifying" to an orphan branch.
Nit 2: the index-table symptom line is good; consider cross-linking the Prevention paragraph to the leases/wisps docs from #4863 so the two land as one story.

Happy to re-review the turnaround immediately — the wyvern squash window is live today and will hit the remote-refs case for real.

steveyegge:
Review addressed in 079a6bf49: Step 3 now deletes remote-tracking refs (`dolt branch -rd`) before `dolt gc --full` — Step 4's force-push recreates them on the new chain — and verifies reclamation with `du -sh .` plus a survivors-recheck note. One deliberate deviation from the review text: a failed gc would NOT cause Step 4 to republish the bloat (push only sends chunks reachable from the new baseline), so the doc says the machine merely keeps the bloat locally until gc succeeds. Both nits taken: root-commit rationale added to Step 2; Prevention now links the heartbeat (leases) and wisps pages. docsync + freshness gates green.

steveyegge (commented):
Re-review of 079a6bf49 + cbe8eface (verdict: APPROVE — recorded as a comment because the shared account authored the PR): the Step 3 blocking gap is closed (remote-tracking refs deleted before gc --full, du -sh verification, survivors recheck), Step 2 carries the root-commit rationale, and the new Step 4 warning correctly states the monotonic-remote property with a workable replace-the-remote recipe. Both earlier nits taken. Your correction to my framing is right — a failed gc cannot make Step 4 republish the bloat, since push sends only chunks reachable from the new baseline; the doc's phrasing now matches that.

One non-blocking wording nit for a follow-up: Step 5 says old clones 'can no longer pull — the histories no longer share a root.' They DO share exactly the root (Step 2 keeps it as the sole ancestor — that's the point of not orphaning). The real hazard is worse than an error: a pull from an old clone can 'succeed' as a cross-merge whose merge commit re-anchors the entire old chain, and a subsequent push resurrects the bloat on the squashed remote. Suggest: 'old clones must not pull — a pull cross-merges the divergent chains and re-anchors the old history; re-create them from the squashed remote.' Fine to land tonight as-is; the operative instruction (re-clone, unfence only after) is correct either way.

PR #4865, opened 2026-07-17
fix(show): flag comment_count as omitted when comments are not streamed

athosmartins:
## What

`bd show --json`'s count-only default returns `comment_count` but no `comments` key unless `--include-comments` is passed. Adds a `comments_omitted` (bool, `omitempty`) field to the payload, set `true` only when `comment_count > 0` and the array was left out.

## Why

A caller that checks only `.comments` sees it absent and concludes "no comments" — wrong whenever `comment_count > 0`. A caller that reads `comment_count` alone has no signal the array was omitted on purpose rather than dropped by a client bug. Measured cost: 3 agents (in a downstream deployment) independently misread this exact payload shape in one session — one twice, including missing a comment where a colleague had already diagnosed the same issue being investigated.

This does **not** revert the count-only default or its perf rationale (`cfcc95799`, pinned by `TestJSONContract_ShowOmitsOptInPayloadsByDefault`) — the array is still opt-in via `--include-comments`. It only makes the partial representation self-announcing instead of silently indistinguishable from "no comments" or "you got everything."

The identical gap exists for `dependent_count`/`dependents` — left out of this PR as a separate, symmetric follow-up rather than scope-creeping this fix.

**Deliberately not done:** did not bump `JSONSchemaVersion` or run `make corpus-regen`. This is an additive `omitempty` field (no existing consumer's parse breaks; no existing corpus fixture has `comment_count > 0` for `show`, so the golden files are untouched), and regenerating the corpus needs a live Dolt+Docker environment I didn't have available to verify the diff myself. Deferring that call to whoever can run it and review the output.

## Verification

- `go build -tags gms_pure_go ./cmd/bd/...` — clean
- `go vet -tags gms_pure_go ./cmd/bd/... ./internal/types/...` — clean
- `golangci-lint run --build-tags gms_pure_go ./cmd/bd/... ./internal/types/...` — 0 issues
- `go test -tags gms_pure_go ./internal/types/...` — all pass (no Dolt/Docker dependency)
- Added `TestJSONContract_ShowCommentsOmittedFlag` to `cmd/bd/protocol/json_contract_test.go`, following the existing `TestJSONContract_ShowOmitsOptInPayloadsByDefault` pattern — could not run it in this sandbox (no Docker for the testcontainers-backed Dolt test server) but it will run under normal CI
- Manually verified end-to-end against a locally-built binary in an isolated scratch workspace (not the system `bd`):
  - 3-comment issue, default `show --json`: `comment_count: 3, comments: null, comments_omitted: true`
  - same issue, `--include-comments`: `comment_count: 3, comments_omitted: null, comments: [...3 real texts...]`
  - 0-comment issue, default `show --json`: `comment_count: 0, comments: null, comments_omitted: null` (true-empty stays distinguishable from the omitted case, no regression)
  - non-JSON `bd show` text output unaffected (was never in scope of the bug)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claim at the PR head). Verdict: **needs changes** — right contract, missing on the proxied route.

**Blocker — proxied `bd show` never sets the new flag.** Verified at the PR head: `CommentsOmitted` appears zero times in `show_proxied_server.go`; the proxied path computes `comment_count` and attaches comments only under `--include-comments`, so the default proxied response still returns `comment_count > 0`, no `comments`, and no `comments_omitted` — exactly the ambiguity this PR fixes on the direct path. The proxied dual needs the same one-line assignment, plus a proxied test for a positive count without `--include-comments`.

**Should-fix — the predicate inherits the known wisp-count asymmetry.** The flag keys off `cmtCount > 0`, but both count implementations query only the permanent `comments` table, and the conformance suite explicitly pins that `CountIssueComments` returns zero for wisps. A commented wisp therefore reports `comment_count: 0` with no omission flag. Pre-existing, but the new contract makes it visible — worth a note or a follow-up issue rather than silence.

**Should-fix — wire-contract artifacts not updated.** `docs/reference/json-schema.md` doesn't mention count-only behavior or `comments_omitted`, and the protocol corpus's `show` capture has no comments so it can't exercise the field. There's also a genuine policy tension to resolve explicitly: `protocol/CATALOG.md` says every deliberate wire change bumps the version, while the schema doc says additive optional fields don't — pick one in the PR body and add a positive-comment corpus capture.

Nit: the new contract-test helper repeats a shape already partly centralized in `helpers_test.go`.

maphew (approved):
Approving — this supersedes my 2026-07-23 changes-requested review. The blocker (proxied `bd show` never set `comments_omitted`) is fixed as a maintainer commit on this branch (`407a5cd81`), mirroring the direct path's ga-clgh else-branch in `proxiedBuildDetails`, with three proxied-path subtests (flag set when count>0 and comments not included; absent when included; absent at zero) and a `comments_omitted` entry in docs/reference/json-schema.md. No schema version bump: additive optional field the direct path already emits.

The wisp comment-count asymmetry from my review is real but pre-existing and orthogonal — filed separately rather than held against this PR.

Validation: build/vet/gofmt clean, new subtests plus the full proxied show suite green, independent review + cross-vendor Codex gate signed off. Merge tail goes to the pr-babysit patrol. Thanks athosmartins.

PR #2385, opened 2026-03-05
fix: add contributor auto-routing fallback to show/update/close (GH#2345)

steveyegge:
## Summary
- Add contributor auto-routing fallback to bd show, bd update, and bd close
- When an issue is not found via prefix-based routing, these commands now fall back to openRoutedReadStore/determineAutoRoutedRepoPath
- Fixes GH#2345: bd show/update/close fail to find issues created via contributor auto-routing

## Test plan
- bd create + bd show works with routing.mode = auto
- bd update works with auto-routed issues
- bd close works with auto-routed issues
- Non-auto-routed workflows unchanged

Issue #2345, opened 2026-03-04
bd show/update/close fail to find issues created via contributor auto-routing


eitsupi:
## Summary

When contributor auto-routing is enabled (`routing.mode = auto`, `routing.contributor = <path>`), `bd create` and `bd list` correctly route to the planning repo's database, but `bd show`, `bd update`, and `bd close` do **not** implement contributor auto-routing. They query only the local project's database, which is empty, resulting in "no issue found matching" errors.

This is a regression/gap related to #942 (same symptom, different root cause) and #1165 (routing confusion).

## Environment

- bd version: 0.58.0 (ae14933d)
- OS: Linux (WSL2)
- Dolt mode: server
- Project prefix: multi-hyphen (e.g. `my-cool-project`)

## Reproduction

```bash
# Config has routing enabled:
# routing.mode = auto
# routing.contributor = <planning-repo-path>

bd create --title="Test issue" --type=bug --priority=2
# ✓ Created issue: my-cool-project-avo

bd list
# ○ my-cool-project-avo [● P1] [bug] - ...   (WORKS)

bd show my-cool-project-avo
# Error fetching my-cool-project-avo: no issue found matching   (FAILS)

bd update my-cool-project-avo --status=in_progress
# Error resolving my-cool-project-avo: no issue found matching  (FAILS)
```

## Root Cause Analysis

Verified via direct SQL queries:

```bash
bd sql "SELECT id FROM beads.issues"
# my-cool-project-avo   ← data is HERE (planning repo's default DB)

bd sql "SELECT id FROM my_cool_project.issues"
# (0 rows)              ← local project DB is EMPTY

bd sql "SELECT DATABASE()"
# my_cool_project        ← default connection uses local DB
```

### Why this happens

1. **`bd create`** (create.go:383-411): Implements contributor auto-routing via `routing.DetectUserRole()` + `routing.DetermineTargetRepo()`. When the user is a contributor, it opens a new store at the planning repo's `.beads/`, which has no `metadata.json` and defaults to the `"beads"` database name. Issue is created in `beads.issues`.

2. **`bd list`** (list.go:706-716): Implements contributor auto-routing via `openRoutedReadStore()` → `determineAutoRoutedRepoPath()`. Same routing logic, same `"beads"` database. Successfully finds the issue.

[...]

PR #4866, opened 2026-07-17
fix(routing): warn when contributor auto-routing swaps bd list/ready to the planning store

RaviTharuma:
## Problem

When `beads.role` resolves to `contributor` and auto-routing is active (`routing.mode=auto` + `routing.contributor`, as set up by contributor-style init — this also happens implicitly when initializing in fork checkouts), `bd list` and `bd ready` silently read from the routed contributor planning store instead of the local project database.

`bd stats` and `bd show <id>` do **not** go through this routing path, so they keep reporting the local project's data. The result looks exactly like data loss:

```
$ bd stats        # Total Issues: 38
$ bd show x-19h   # works, shows the issue
$ bd list --all   # "No issues found."
$ bd ready        # "No open issues"
```

There is no indication anywhere that a different store was consulted. Diagnosing this from the CLI alone took considerable effort — the fix turned out to be `git config beads.role maintainer`, but nothing points there.

## Change

Print a stderr note at the exact point where `openRoutedReadStore` swaps the active store for these two commands, including the local issue count that is being hidden and the fix command:

```
note: beads.role=contributor routes bd list/ready to the contributor planning store, not this project (38 issue(s) in this project are hidden as a result).
  Fix: git config beads.role maintainer
```

- stderr only; stdout/JSON output unchanged
- only fires when the store is actually swapped (`routed == true`)
- not gated on `--quiet`, matching the existing unconditional stderr truncation notice in `bd ready --json`, since this communicates a change of data source

## Repro

1. Project with issues, `git config beads.role contributor`, auto-routing active (e.g. initialized in a fork checkout)
2. `bd list` → empty, no explanation (before) / empty + stderr note (after)

Observed in the wild on two real projects (both initialized inside `~/Code/forks/*` checkouts) on bd v1.1.0 (Homebrew, macOS arm64).

## Precise trigger conditions

`beads.role=contributor` alone does **not** trigger this — verified on a fresh init. The swap requires auto-routing to be active as well: `routing.mode=auto` + `routing.contributor=<path>` (as set up by `bd init --contributor`; see engdocs/CONTRIBUTOR_NAMESPACE_ISOLATION.md "Auto-Routing Based on User Role"). `openRoutedReadStore()` then swaps the active store unconditionally for `bd list` (list.go) and `bd ready` (ready.go) before any query runs. `bd stats` never uses this path, and `bd show <id>` only consults routing as a not-found fallback — which is exactly why the symptoms look so contradictory.

## Prior art

#2345 / PR #2385 fixed the same class of silent-routing invisibility for the `show`/`update`/`close` not-found fallback path (March 2026). This PR closes the analogous gap for the unconditional `list`/`ready` path, which that fix did not cover. The notice style follows the existing `warning: beads.role not configured (GH#2950)` message.

Tests: `go test ./internal/beads/... ./internal/routing/...` and `go test -run 'Rout|List|Ready|Contributor' ./cmd/bd/` pass with the change; `go vet ./cmd/bd/` is clean.

## Why this hits fork checkouts without anyone opting in

On this machine, three projects initialized inside fork checkouts all ended up with `routing.contributor` pointing at the same auto-provisioned planning store, without it ever being set by hand:

```
$ bd config get routing.contributor     # in each of 3 fork-based projects
~/.beads-planning

$ bd -C ~/.beads-planning list --all --json | jq length
0                                        # the planning store is empty
```

Combined with `beads.role` resolving to `contributor` in those same checkouts, `bd list`/`bd ready` route to an empty store and report nothing — while `bd stats` reports the real local count (38 in one case). Nothing in the output connects the two. That is the practical impact: the default fork setup can silently look like total data loss.


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the warning is genuinely wanted (silent store-swaps have burned people), but as written it can misdiagnose and miscount.

**Blocker — every routed read is labeled contributor routing.** Verified at the PR head: `printContributorRoutingNotice` unconditionally prints "beads.role=contributor routes bd list/ready…" and recommends `git config beads.role maintainer` — but the route opener consults `routing.default` and `routing.maintainer` too, and the callers in `list.go`/`ready.go` fire the notice for *any* non-`.` target. For maintainer- or default-routed setups the diagnosis is false and the recommended fix is a no-op (or already true). The opener needs to surface *which* rule routed, and the notice should say that.

**Should-fix — violates the documented `--quiet` contract.** Quiet mode is defined as "errors only," and this non-error notice writes to stderr with direct `Fprintf`, ignoring `quietFlag`.

**Should-fix — the hidden-count is wrong and silently optional.** The notice calls all of `stats.TotalIssues` "hidden as a result" — that's an unfiltered `COUNT(*)` including closed/deferred/blocked issues that `bd ready`'s predicates would never show. Say "this project contains N total issues" (or compute the real delta), and decide what happens when `GetStatistics` errors instead of silently dropping the count.

Also: no tests changed for a user-visible behavior change (contributor/maintainer/default routes × both commands × `--quiet` × JSON isolation is the matrix), and the comment claiming `show` doesn't route is contradicted by `show.go`'s routed resolution path.

Worth one more round — a correctly-attributed, quiet-respecting version of this warning is a real usability win.

maphew (approved):
Approving — supersedes my 2026-07-23 changes-requested review; fixes applied as maintainer commits. The notice now attributes the swap to the actual matched rule (a `RoutingRule` enum threaded out of `internal/routing`), with per-rule remediation commands; the contributor wording covers both explicit `beads.role=contributor` and origin-URL inference (the gate caught that the old text could claim a config key that wasn't set); and the notice honors `--quiet`. Wiring tests cover contributor, maintainer, and default rules end-to-end. Your underlying idea — never silently swap read stores — ships intact and is a genuine usability win.

Validation: build/vet/gofmt clean, routing/notice tests green by name; cross-vendor Codex gate signed off after the wording fix. Merge tail to the pr-babysit patrol. Thanks RaviTharuma.

PR #4870, opened 2026-07-17
feat: export DependencyAddOptions from the root package

julianknutsen:
## Summary
`Transaction.AddDependencyWithOptions` and `Transaction.CycleThroughEdges` are both on the public `Transaction` interface, but `DependencyAddOptions` lived only in `internal/storage` — so the designed bulk-write path (`SkipCycleCheck` per edge + one whole-graph `CycleThroughEdges` pass before commit, bd-6dnrw.8) was **uncallable by embedders**.

## Why now
Gas City's native graph-apply pours molecule-sized batches (67 nodes / ~100 blocking edges). Each edge pays the recursive cycle-reachability query, which cannot finish inside a 120s per-command budget: reproduced 2026-07-17, the batch died at the deadline mid-edges twice (`failed to check for dependency cycle: context deadline exceeded`), then fell back to per-bead creates — a 9-minute pour. Gas City has an interim deadline-scaling fix (gastownhall/gascity#4391); this export is the prerequisite for the real fix (skip per-edge, one batch check).

Gas City will pick this up at the next version pin bump (go.mod + deps.env lockstep).

## Change
One root alias in `beads.go` + a compile-time proof that the alias satisfies the `Transaction` method signature.

### DIFF
--- a/beads.go
+++ b/beads.go
@@ -23,6 +23,15 @@ type Storage = beads.Storage
 // Use Storage.RunInTransaction() to obtain a Transaction instance.
 type Transaction = beads.Transaction
 
+// DependencyAddOptions controls transaction-scoped dependency insertion for
+// Transaction.AddDependencyWithOptions. Exported so embedders' bulk graph
+// writers can set SkipCycleCheck per edge and run one whole-graph
+// Transaction.CycleThroughEdges pass before commit (bd-6dnrw.8) instead of
+// paying the recursive per-edge cycle query — which cannot finish inside a
+// per-command budget on molecule-sized graphs (observed: a 67-node/100-edge
+// batch blowing a 120s deadline mid-transaction, gascity 2026-07-17).
+type DependencyAddOptions = storage.DependencyAddOptions
+
 // RemoteStore provides dolt remote management and replication operations.
 // Use type assertion on a Storage value to access these methods:
 //
--- a/beads.go
+++ b/beads.go
@@ -30,6 +30,11 @@ type Transaction = beads.Transaction
 // paying the recursive per-edge cycle query — which cannot finish inside a
 // per-command budget on molecule-sized graphs (observed: a 67-node/100-edge
 // batch blowing a 120s deadline mid-transaction, gascity 2026-07-17).
+//
+// Callers that set SkipCycleCheck MUST run Transaction.CycleThroughEdges before
+// commit and fail on new blocks/conditional-blocks/parent-child cycles
+// (waits-for is excluded); skipping the per-edge check trades per-edge cost for
+// one whole-graph check, never graph integrity.
 type DependencyAddOptions = storage.DependencyAddOptions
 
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR exports `DependencyAddOptions` from the root `beads` package as a type alias for the internal `storage.DependencyAddOptions`, which lets embedders' bulk graph writers name that type to set `SkipCycleCheck` per edge and run a single whole-graph `Transaction.CycleThroughEdges` pass before commit. That batch dependency-add path was previously uncallable from outside the module, so molecule-sized batches were forced through the recursive per-edge cycle query — the exact path that blew a 120s transaction deadline on a 67-node/100-edge batch. Exporting the alias makes the designed batch path reachable for external embedders.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The exported alias's doc comment described only the happy path and dropped the mandatory whole-graph cycle-check guardrail that the internal `storage.DependencyAddOptions` carries. The final patch mirrors that contract onto the public surface, so an embedder reading the root-package doc is told that setting `SkipCycleCheck` obligates them to run `Transaction.CycleThroughEdges` before commit and fail on new block / conditional-block / parent-child cycles — rather than dropping the per-edge check with nothing replacing it and silently committing a cycle.

Review findings addressed:
- Contract & interface fidelity (minor, high confidence): the public alias docs understated the mandatory whole-graph cycle-check guardrail. The final patch now documents the `SkipCycleCheck` → `CycleThroughEdges` obligation directly on the exported alias, matching the internal type's contract.
- Test clarity (nit, high confidence): the runtime field round-trip in the added test is tautological, and the compile-time assertion is the real contract check. This is a non-gating note — the compile-time assertion that the alias is accepted by the exported `Transaction.AddDependencyWithOptions` is the meaningful proof of the export surface.

Remaining non-gating notes:
- The runtime `TestDependencyAddOptionsIsExported` field check is redundant with the compile-time assertion; it is harmless and left in place. No remaining gating issues were found.

### Maintainer Changes
`docs(beads): mirror SkipCycleCheck whole-graph guardrail on public alias` — beads.go, 1 file changed, 5 insertions(+). Documentation-only change to the exported alias's doc comment; no behavior change.

### Final Review Status
Ready for the merge queue: final head `e728ce65` has passing required GitHub checks (Conformance, Cross-Version Smoke Tests, PR, PR Risk, and Regression Tests all green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/29627865311

### Review Iterations
2 review passes performed. The first surfaced the public-doc guardrail gap (fixed by the maintainer doc commit) and the non-gating test-clarity note; the second re-reviewed the fixed branch and validated the export surface, batch-path reachability, cycle-safety behavior, and test evidence, reaching approval.

---

PR #3051, opened 2026-04-06
refactor: remove redirect file creation from bd worktree create

osamu2001:
## Summary

- Remove `.beads/` directory and redirect file creation from `bd worktree create`
- `findDatabaseInTree` already discovers the main repo's `.beads` via `GetWorktreeFallbackBeadsDir` (git common dir), making the redirect redundant
- Update help text and doc strings to reflect automatic discovery
- Remove now-unnecessary redirect computation tests

## Test plan

- [x] `CGO_ENABLED=0 go build ./cmd/bd/` passes
- [x] `CGO_ENABLED=0 go vet ./cmd/bd/` passes
- [x] Existing `TestGetRedirectTarget` and `TestAddToGitignore` pass

PR #4875, opened 2026-07-17
fix(beads): don't follow .beads redirect to an invalid target (#4692)

### LINKED ISSUE #4692: bd 1.1: graceful server→embedded fallback writes a worktree-depth redirect into a rig-root .beads, silently hiding all issues
## Summary

On a shared Dolt `sql-server` setup, `bd` can silently write a stray `.beads/redirect` file into a **rig-root** `.beads/` directory whose `metadata.json` validly points at a reachable server database. The redirect is computed with **worktree-relative depth** (e.g. `../../.beads`, or `../../../../.beads`), so it climbs *past* the rig's own store and lands on an unrelated/nonexistent location. Every subsequent `bd list` / `bd show` then reports **"No issues found"** even though all data is intact on the Dolt server. It has recurred multiple times on the same repo (the artifacts regenerate after a manual fix).

## Environment

- **bd v1.1.0-rc.1** (`fa4dce4`). The introducing code path ("graceful server-to-embedded fallback", the worktree redirect writer) is still present on `main`.
- **Backend:** a single shared Dolt `sql-server` (port 3307) serving many per-rig databases. Each rig's `.beads/metadata.json` has `dolt_mode: server`, `dolt_database: <rig>`, `dolt_server_port: 3307`, and the server is up/reachable.
- **Layout:** a monorepo ("town") with a **rig-root** `.beads/` (the canonical store, which must have **no** redirect) plus sub-workspaces (crew / refinery / polecat git **worktrees**), each of which legitimately carries a `.beads/redirect` pointing *up* to the rig root (depth = nesting level).

## Symptom

- `bd list` / `bd show <id>` → **"No issues found"** for a rig whose beads were live moments earlier.
- A stray file appears at `<rig>/.beads/redirect` containing a worktree-depth relative path. Observed both `../../.beads` and `../../../../.beads` on the same repo on different days.
- **Data is safe:** `SELECT COUNT(*) FROM <rig>.issues` on the server returns the full, intact count — only bd's path resolution is diverted.

## Mechanism (traced)

1. **Precursor:** a **0-byte `<rig>/.beads/beads.db`** (an uninitialized embedded SQLite file) sits at the rig root. Only rigs with this 0-byte file are affected; rigs with a *populated* or *absent* `beads.db` are fine.
2. **Trigger:** a **real crew git checkout/merge** (PR-merge cycle) runs the worktree's `post-checkout` / `post-merge` hooks, which invoke `bd hooks run …`. bd resolves through the worktree redirect to the rig-root store, hits the 0-byte db, and the **graceful server→embedded fallback** writes a `redirect` into the **rig-root** `.beads/` — but with the *worktree's* relative depth, so it points outside the rig.
3. **Result:** bd now follows that rig-root redirect on every call → wrong/empty location → silent "No issues found".

Additional observations:
- **It regenerates.** Removing the redirect + the 0-byte `beads.db` fixes it, but the 0-byte `beads.db` reappeared within ~4 days (next merge cycle) and the redirect returned.
- **Correlates with merge churn.** The repos that break run many active worktrees; low-churn repos with the same 0-byte precursor but unwired crew hooks never trigger it.
- Plain reads/writes and a *sterile* `bd hooks run post-checkout` (old == new HEAD) do **not** trigger it — it needs a real HEAD change.

## Why the recent related fix doesn't cover this

`ddbd25e26` — *fix(storage): fail loud when metadata.json exists but cannot be loaded, instead of silently opening the embedded store* — fires when `metadata.json` **cannot be loaded**. Here `metadata.json` is valid and the server is reachable; a **redirect file takes precedence** over the valid metadata, so that guard is never reached.

## Suggested fixes (any one would break the cycle)

1. The graceful server→embedded fallback should **never write a worktree-depth redirect into a resolved rig-root** `.beads/` (a directory that has a valid server-mode `metadata.json`). Writing a redirect there is always wrong.
2. When a rig-root `.beads/` has a valid `metadata.json` (`dolt_mode: server`) pointing at a **reachable** server DB, a co-located `redirect` file should be ignored (or flagged), not silently followed.
3. Following a redirect whose target does not resolve to a valid store should **fail loud**, not silently fall back to an empty embedded store.
4. Avoid leaving/creating a 0-byte `beads.db` at a server-mode rig root (the precursor).

## Workaround

`mv <rig>/.beads/redirect` and `mv <rig>/.beads/beads.db` (the 0-byte one) aside; bd reads from the server again. Reversible, no data loss.

maphew:
## Why

Fixes #4692. In bd 1.1, when a shared-server connection failed, a graceful fallback could leave a `.beads/redirect` file pointing at a directory with no database. bd followed it unconditionally and silently landed on an empty embedded store, so `bd list`/`bd show` reported **zero issues** with no error - every tracked issue appeared gone.

The redirect *write*-site that produced this was already removed upstream in #3051 (redirect creation deleted from `bd worktree create`), but the *follow* side still honors any redirect it finds, so a stale redirect file left behind by an older binary still triggers the silent-empty-store failure.

## What

Guard the follow side in `FollowRedirect` (`internal/beads/beads.go`), the single choke point for every redirect consumer (`ResolveBeadsDirForRepo`, `FindDatabasePath`, worktree fallbacks):

- After resolving the redirect target, validate it with the existing `hasBeadsProjectFiles` helper (checks for metadata.json / config.yaml / dolt / embeddeddolt / `*.db`). If the target has **none** of those, the redirect is ignored: the source dir is returned unchanged and a one-line warning naming both paths is emitted (deduped once per source dir per process via a package-level `sync.Map`, since `FollowRedirect` is called many times per invocation).
- A present-but-corrupt target `metadata.json` still counts as "has metadata" and the redirect is followed, so the existing hard-error in `store_factory.go` surfaces the corruption loudly rather than this guard swallowing it.

This validates the **target**, not the source's mode. An earlier draft keyed the guard on "source is server-mode", but review (cross-vendor: Claude opus + Codex gpt-5.6) found that breaks the documented multi-clone redirect topology preserved by #4592 (`fb51196f7`, "preserve source rig's dolt_database across .beads/redirect") - a server-mode rig legitimately redirects to a shared Gas Town `.beads` while keeping its own `dolt_database`. Target-validation matches the documented contract ("The target directory must exist and contain a valid database", docs/reference/advanced.md) and leaves that topology untouched; `TestResolveRedirect_PreservesSourceDatabase` passes unmodified.

`bd doctor` is updated to match: a redirect whose target has no database is now an actionable finding (not "safe to delete" cruft), and `doctor --clean` will no longer delete `metadata.json` from a `.beads` dir that also holds a redirect file.

## Tests

New in `internal/beads` (all pin `BEADS_DOLT_SERVER_MODE`/`BEADS_DOLT_SHARED_SERVER` empty for determinism): redirect to a nonexistent target is ignored with warning; redirect to a dir with no database is ignored; redirect to a valid target is followed; redirect to a valid target from a **server-mode** source is still followed (the #4592-topology regression test). Plus doctor findings for invalid-target redirects and a `doctor --clean` test asserting metadata.json is preserved alongside a redirect. `go build`, full `internal/beads` and `cmd/bd/doctor{,/fix}` test packages, and `gofmt` all clean.

### DIFF
--- a/cmd/bd/doctor/artifacts.go
+++ b/cmd/bd/doctor/artifacts.go
@@ -396,5 +396,48 @@ func validateRedirect(beadsDir string, report *ArtifactReport) {
 			Description: fmt.Sprintf("redirect target is not a directory: %s", target),
 			SafeDelete:  false,
 		})
+		return
 	}
+
+	// gastownhall/beads#4692: FollowRedirect ignores a redirect whose target
+	// has no metadata.json and no recognizable database (a stray
+	// worktree-depth redirect, e.g. a relic of the "bd worktree create"
+	// write-site removed in #3051, can point past the real .beads dir into
+	// an empty location). Flag that here too as an actionable warning, not
+	// SafeDelete cruft -- the fix is to correct or remove the redirect file,
+	// not to delete anything automatically.
+	if !hasDatabaseOrMetadata(resolvedTarget) {
+		report.RedirectIssues = append(report.RedirectIssues, ArtifactFinding{
+			Path:        redirectPath,
+			Type:        "redirect",
+			Description: fmt.Sprintf("redirect target has no database or metadata.json (ignored by bd): %s", target),
+			SafeDelete:  false,
+		})
+	}
+}
+
+// hasDatabaseOrMetadata reports whether dir contains a metadata.json or a
+// recognizable Dolt/SQLite database. Mirrors the target-validity check
+// FollowRedirect (internal/beads) applies before honoring a redirect;
+// duplicated locally (rather than importing internal/beads) to keep this
+// package's existing pattern of small, self-contained filesystem checks
+// (see isRedirectExpectedDir/hasRedirectFile above).
+func hasDatabaseOrMetadata(dir string) bool {
+	if _, err := os.Stat(filepath.Join(dir, "metadata.json")); err == nil {
+		return true
+	}
+	if info, err := os.Stat(filepath.Join(dir, "dolt")); err == nil && info.IsDir() {
+		return true
+	}
+	if info, err := os.Stat(filepath.Join(dir, "embeddeddolt")); err == nil && info.IsDir() {
+		return true
+	}
+	dbMatches, _ := filepath.Glob(filepath.Join(dir, "*.db"))
+	for _, match := range dbMatches {
+		baseName := filepath.Base(match)
+		if !strings.Contains(baseName, ".backup") && baseName != "vc.db" {
+			return true
+		}
+	}
+	return false
 }
--- a/cmd/bd/doctor/artifacts_test.go
+++ b/cmd/bd/doctor/artifacts_test.go
@@ -227,6 +227,11 @@ func TestScanForArtifacts_ValidRedirect(t *testing.T) {
 	if err := os.MkdirAll(targetDir, 0755); err != nil {
 		t.Fatal(err)
 	}
+	// A valid redirect target must have a metadata.json or database
+	// (gastownhall/beads#4692 guard).
+	if err := os.WriteFile(filepath.Join(targetDir, "metadata.json"), []byte(`{"database":"beads.db"}`), 0644); err != nil {
+		t.Fatal(err)
+	}
 
 	// Create redirect pointing to valid target
 	if err := os.WriteFile(filepath.Join(beadsDir, "redirect"), []byte(targetDir), 0644); err != nil {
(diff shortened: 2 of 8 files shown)


maphew (commented):

The fix guards FollowRedirect with the existing hasBeadsProjectFiles target-validity check, correctly placed after the exists/is-dir check and before the chain check, with once-per-source stderr deduping. It validates the target (not source mode), preserving the #4592 multi-clone redirect topology, and the regression tests (server-mode source still follows a valid target; invalid target ignored with warning; valid target followed) are meaningful. Doctor and doctor --clean companion changes are consistent with the stated intent, and CI is fully green with a clean mergeable state. One minor divergence: the doctor-side hasDatabaseOrMetadata omits the config.yaml branch that hasBeadsProjectFiles (which it claims to mirror) includes.

**Findings:**
- **minor** `cmd/bd/doctor/artifacts.go:425` — hasDatabaseOrMetadata's doc comment says it "Mirrors the target-validity check FollowRedirect (internal/beads) applies," but it omits the config.yaml branch that hasBeadsProjectFiles actually checks (beads.go:620). config.yaml is a live, supported config file (jira secrets, custom types, issue-prefix), so a redirect target holding only config.yaml (no metadata.json / db) would be FOLLOWED by bd yet flagged by doctor with the message "redirect target has no database or metadata.json (ignored by bd)" -- a false positive whose text is also factually wrong for that case. Fix: add a `config.yaml` os.Stat branch to hasDatabaseOrMetadata (and consider naming config.yaml in the finding message) so it truly mirrors FollowRedirect.
- **nit** `cmd/bd/doctor/artifacts.go:380` — validateRedirect checks the raw resolvedTarget while FollowRedirect canonicalizes via canonicalizeBeadsDirPath/preferStableBranchWorktreeBeadsDir before validating. For a detached-snapshot-worktree redirect, FollowRedirect may retarget to a stable branch worktree that has a db while doctor inspects the detached path that may not, yielding an inconsistent finding. Pre-existing divergence in doctor (not introduced here) and a niche case, but worth noting since this PR newly makes target-content a doctor finding.

PR #4877, opened 2026-07-17
feat(create): make every native issue field addressable from --graph JSON plans

### LINKED ISSUE #4362: bd create --graph silently drops unrecognized node keys (parent/deps hierarchy lost)
## Version

bd version 0.63.3 (884cc117: HEAD@884cc1171e3e)

## Summary

`bd create --graph FILE` silently ignores unrecognized node-level keys (`parent`, `deps`) and creates all issues as flat, disconnected nodes. No warning or error is emitted and the command reports success. The parent-child hierarchy and blocking edges from the input file are lost.

## Minimal Repro

**Input file (`graph.json`):**
```json
{
  "nodes": [
    {
      "key": "A",
      "title": "Parent Epic",
      "type": "epic",
      "description": "This is the parent epic"
    },
    {
      "key": "B",
      "title": "Child Task One",
      "type": "task",
      "description": "First child task",
      "parent": "A"
    },
    {
      "key": "C",
      "title": "Child Task Two",
      "type": "task",
      "description": "Second child task, depends on B",
      "parent": "A",
      "deps": ["B"]
    }
  ]
}
```

**Steps:**
```bash
mkdir /tmp/repro && cd /tmp/repro
git init -q
bd init -p repro
bd create --graph graph.json
bd children repro-<epic-id>    # shows 0 children
bd dep tree repro-<epic-id>    # shows no dependencies
```

## Actual Behavior

```
Created 3 issues
  A -> repro-xxx
  B -> repro-yyy
  C -> repro-zzz
```

`bd children` returns only the epic itself (0 children). `bd dep tree` shows no edges. `bd show` on the child issues shows no parent or blocking relationship. No warning about unrecognized keys.

## Expected Behavior

One of:
1. **Reject:** Fail with a clear error: `unrecognized node key "parent" — did you mean "parent_key"? See: bd create --graph --help`
2. **Warn:** Emit a warning per unrecognized key: `warning: node B: unrecognized key "parent" (ignored)`

The current behavior is a silent data loss: the user believes hierarchy was created, but it was not.

## Context / Correct Schema

The correct schema for parent-child hierarchy uses `parent_key` at the node level (not `parent`). For blocking dependencies, use a top-level `edges` array with `from_key`, `to_key`, and `type: "blocks"`. The commonly intuited `parent` and `deps` node-level keys are accepted without error but have no effect.

A working equivalent of the above:
```json
{
  "nodes": [
    { "key": "A", "title": "Parent Epic", "type": "epic" },
    { "key": "B", "title": "Child Task One", "type": "task", "parent_key": "A" },
    { "key": "C", "title": "Child Task Two", "type": "task", "parent_key": "A" }
  ],
  "edges": [
    { "from_key": "B", "to_key": "C", "type": "blocks" }
  ]
}
```

Unknown keys at the node level should be rejected or warned about rather than silently dropped.

### LINKED ISSUE #4890: bd create --graph: plan nodes can only set a fraction of native issue fields, forcing create-then-patch
## Problem

`bd create --graph <plan.json>` is the only way to create an issue graph atomically, but plan nodes expose just a handful of fields: title, type, description, assignee, priority, labels, string-valued metadata, and parent. Everything else a single `bd create` can set is unreachable from a plan:

- **status** — graphs are born `open`; seeding an in-progress/closed graph means a follow-up `bd update` per node
- **planning/content fields** — `design`, `acceptance_criteria`, `notes`, `spec_id`, `owner`, `estimated_minutes`, `due_at`, `defer_until`, `external_ref`
- **explicit `id`** — plans can't mint predetermined IDs
- **storage class** — no per-node `ephemeral` / `no_history`, `wisp_type`, `mol_type`
- **`pinned`**, **event fields** (`event_kind`, `actor`, `target`, `payload`)
- **metadata** is limited to string values; numbers/objects are rejected
- **edges** can't carry waits-for fanout-gate metadata (`gate`, `spawner`) or `thread_id`

So any orchestrator or importer building a realistic graph has to create-then-patch: N extra update calls, and the graph is observable in a half-initialized state in between — exactly what the atomic `--graph` call exists to avoid.

## Proposal

Bring `--graph` plan nodes to full parity with single-issue `bd create` by routing node materialization through the same `buildCreateIssue` path, and keep the parity from regressing with a reflection guardrail test that fails whenever `createIssueParams` grows a field that isn't addressable from a plan node (or explicitly excluded with a stated reason). Add the edge-side fields (`gate`, `spawner_key`/`spawner_id`, `thread_id`) with the same guardrail treatment.

## Related

While auditing the waits-for gate paths for this, a latent bug surfaced: the blocked-state SQL compares `JSON_EXTRACT(metadata,'$.gate')` directly, so a waits-for dependency stored without gate metadata (e.g. via `bd dep add --type waits-for`) evaluates NULL, poisoning the enclosing `NOT(... AND ...)` and unblocking the parent when *any* child closes instead of all. Fixing the gate handling (COALESCE default + always-write gate metadata + repair migration) belongs with this change since the new edge fields formalize gate metadata.

anisoptera:
Closes #4890. Closes #4362.

## What

Brings `bd create --graph <plan.json>` to full parity with single-issue `bd create`: every native issue field the CLI can set is now addressable from a plan node, plus two graph-only conveniences (`status` and `pinned`) that `bd create` has no flags for. Graph node creation is routed through the same `buildCreateIssue` path single-create uses, the duplicated gate/metadata logic is hoisted into shared `internal/types` helpers, and a NULL-safe waits-for fanout-gate predicate is fixed (with repair migrations).

Newly addressable **node** fields: `id` (explicit ID), `status`, `design`, `acceptance_criteria`, `notes`, `spec_id`, `external_ref`, `owner`, `estimated_minutes`, `due_at`, `defer_until`, per-node `ephemeral`/`no_history`, `wisp_type`, `mol_type`, `pinned`, and the event fields (`event_kind`, `actor`, `target`, `payload`). `metadata` now accepts arbitrary JSON values, not just strings. **Edges** gain waits-for fanout-gate metadata (`gate`, `spawner_key`/`spawner_id`) and `thread_id`.

## Why

`--graph` previously exposed only a handful of fields (title, type, description, assignee, priority, labels, string metadata, parent), so building a real issue graph in one atomic call couldn't set status, planning fields, or event payloads — you had to create and then patch. This closes that gap and keeps it closed: a reflection guardrail test (`TestGraphApplyNodeCoversCreateIssueParams`) fails if `createIssueParams` grows a field that isn't addressable from a plan (or explicitly excluded with a reason).

It also fixes a latent correctness bug in waits-for gate evaluation. The blocked-state SQL compared `JSON_EXTRACT(metadata,'$.gate') = 'any-children'` directly, so a waits-for dependency created without gate metadata (e.g. `bd dep add --type waits-for`, which stores none) yielded NULL and poisoned the enclosing `NOT(... AND ...)` — unblocking the gate as soon as *any* child closed instead of waiting for all. The fix:

- Runtime SQL (`issueops/blocked_state.go`) now `COALESCE`s a missing gate to the `all-children` default.
- The write paths always store gate metadata for waits-for deps, so new rows can't hit this.
- Two migrations (`0056` main + `ignored/0013` wisp twin) recompute `is_blocked` to repair rows mis-set by the old predicate.

### Notable internals

- New shared helpers in `internal/types`: `NewGraphEdgeDependency` (which owns waits-for gate-metadata serialization), `NewWaitsForDependency`, `NewGraphNodeDependency`, `MergeMetadataRefs`, `IsValidWaitsForGate`, and `ValidMolTypeNames`/`ValidWispTypeNames` — used by both the embedded and proxied/domain apply paths so the two can't drift.
- Unknown-field warning sets are now derived from struct JSON tags via reflection (`jsonTagSet`), replacing three hand-maintained maps and their guardrail test; unknown-key matching is case-insensitive (encoding/json binds case-variant keys anyway).
- Plan validation runs against the *same* materialized issue the apply path stores (with the real plan-wide options), so plan-time and insert-time validation can't diverge. Proxied-server mode enforces a uniform storage class across the plan, and `domain.applyGraph` guards its own single-table routing against mismatched node storage classes.

## Verification

Built and tested with the project's `gms_pure_go` path:

- `go build -tags gms_pure_go ./cmd/bd/... ./internal/storage/... ./internal/types/...` — clean.
- `go test -tags gms_pure_go ./internal/types/... ./internal/storage/domain/... ./internal/storage/issueops/...` — pass.
- `go test -tags gms_pure_go -run 'GraphApply|ValidateGraph|EmbeddedCreate|...' ./cmd/bd/...` — pass.

New tests cover: full-field round-trip (`graph_full_fields`), per-node storage-class routing to the wisps table, waits-for gate metadata, explicit-ID prefix checks, node-field and edge-gate validation rules, the NULL-safe ungated-waits-for-still-gets-metadata invariant, case-insensitive unknown-field detection, and two parity guardrails: `TestGraphApplyNodeCoversCreateIssueParams` (every `createIssueParams` field addressable from a plan node) and `TestBuildDomainGraphPlanCoversEdgeFields` (every edge/node-dep field survives the proxied projection). The `--graph` plan schema is documented in an embedded help supplement shown by `bd create --help` (the committed CLI reference pages are pinned to the released bd by main's docs pipeline, so the section appears there once this ships in a release), and `TestDocsGraphPlanExampleValidates` pins the supplement's example to the validator.

### Manual verification

Beyond the automated suite, the feature was driven end-to-end against a live embedded-Dolt database (`bd` built via `make build`):

- **Full-field round-trip** — one plan setting every new node field, confirmed via `bd show --json`: explicit `id`, `status`, all content/planning fields, `mol_type`, `pinned`, and event fields all round-tripped; non-string `metadata` values survived (a number stayed a number, a nested object stayed an object); `started_at`/`closed_at` were auto-stamped for `in_progress`/`closed` nodes; the `feat` alias normalized to `feature`; `metadata_refs` resolved to the minted IDs; the parent-child dep was created.
- **Fanout gates** (observed via `bd blocked` state transitions): an `any-children` gate unblocked after the *first* child closed; an `all-children` gate stayed blocked until *all* children closed.
- **NULL-gate regression fix** — a `waits-for` dep created via `bd dep add` (which stores no gate metadata) stayed blocked after the first child closed and cleared only once all closed, i.e. it now defaults to `all-children` instead of wrongly unblocking on the first close.
- **Guardrails** — invalid status, event-fields-on-a-non-event node, foreign-prefix explicit ID (and its `--force` override), duplicate explicit ID, and gate-on-a-non-`waits-for` edge are all rejected with actionable messages; unknown fields warn with corrective hints; per-node `ephemeral` routed to the wisps table.

Note: `--dry-run` validates graph *structure* only — it does not resolve stored-dependency blocking cycles, so a plan that makes a parent wait on its own child passes dry-run but is (correctly) rejected at apply time. This limitation is called out in the dry-run output note.


maphew (dismissed):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the migration collision at the PR head). Verdict: **needs changes** — a lot of genuinely useful capability here (every native field addressable from `--graph` plans), gated behind one hard landing blocker and a set of drift adaptations.

**Blocker — duplicate migration versions.** Verified: main already ships migration 0056 (and ignored 0013/0014); this PR adds its own differently-named 0056 + ignored 0013. Both files survive a textual merge, and the migration loader panics on duplicate version discovery *before any SQL runs* — bricking every command. Rebase onto current main and take the next free version numbers (this is also a protected migration path per the maintainer guidelines).

**Drift should-fixes:** the new prefix-selection helper always prefers project YAML, but current main deliberately uses the opened *global* store's prefix in `--global` mode (pinned by an existing test) — preserve that for both single and graph creation, else valid global IDs get rejected when project YAML exists.

**Correctness should-fixes:** (1) inline `deps` bypass graph validation — the local-cycle adjacency includes parents and top-level edges but not `node.Deps`, and inline deps skip the parent-conflict check, so dry-run accepts plans that live apply then rejects and rolls back (an existing accepted fixture demonstrates the parent-shadow case, contradicting the schema doc); (2) the claim that the schema stays available via `bd create --help` is false — the custom help path never consults `commandDocSupplements` and `createCmd` has no `Long`; (3) dry-run's "performs no writes" comment is too broad — root pre-run still auto-migrates and opens the store before RunE sees `--dry-run` (say "no graph-plan writes").

**Structural:** the five-pass graph orchestration now exists in duplicate (`graph_apply.go` and the domain layer, including duplicated pair helpers) — worth converging on one implementation during the rebase.

maphew (commented):
Thank you @anisoptera — this is the kind of contribution we hope for: a real gap closed end-to-end with guardrails so it stays closed, a latent correctness bug found and fixed along the way, and a branch history showing disciplined self-review. We want this in.

This is a paired cross-vendor maintainer review (a Claude reviewer and a Codex reviewer run independently on the same diff at merge-base 38031c5, every finding re-verified against the branch head before posting). Since maintainer edits are enabled on the branch, we will push fix commits here for the items below rather than round-tripping a request-changes — your commits and attribution stay intact.

## What we verified holds up

- **The NULL-gate fix is complete.** The runtime `$.gate` SQL reader exists in exactly one place (`internal/storage/issueops/blocked_state.go`) and the COALESCE default lands there; the `bd doctor` consistency checker (`blocked_consistency.go`) interpolates the same `waitsForGateBlockedSQL` constant, so it inherits the fix, and the Go-side reader (`types.ParseWaitsForGateMetadata`) already defaults to all-children. No unfixed reader — including on current main, where the storage work merged since your branch point added no new gate readers.
- **Three-route parity is real.** The embedded, proxied, and domain apply paths all materialize nodes through the shared `buildCreateIssue` helpers; the old proxied projection that dropped fields is gone; `resolveEdgeRef` prefers `id` over `key` consistently; the reflection guardrails pass.
- **The repair migration recomputes `is_blocked` from scratch** — repairing both wrongly-unblocked and wrongly-blocked rows, computed from status-based reachability, order-insensitive under renumbering.
- CI is fully green on the head.

## Fixes we will push to this branch

1. **Migration 0056 can wedge legacy-shaped clones at startup** *(found by the Codex reviewer)*. The `@has_wisps` guard checks table existence only, then the full CTE references post-split `wisp_dependencies` columns (`depends_on_issue_id` / `depends_on_wisp_id`). A clone whose clone-local wisp tables predate `ignored/0003` (the split migration) while its synced main cursor is current fails the main migration pass on an unknown column — and since the main pass runs before the ignored pass, `ignored/0003` never gets the chance to split the table. Migration 0047 guards this exact hazard with an `INFORMATION_SCHEMA.COLUMNS` shape check; this migration needs the same.
2. **An explicit plan `id` that collides with an existing issue silently overwrites it** *(found by the Codex reviewer; adjudicated as inherited)*. Plan validation catches intra-plan duplicates only; the insert path then runs an unconditional `INSERT ... ON DUPLICATE KEY UPDATE`. We traced this as inherited behavior — `bd create --id` with an existing ID upserts the same way on main today — but plans mint IDs in bulk, so this PR amplifies the exposure. We will add a store-existence preflight to plan validation on both transports.
3. **`ignored/0013` omits the `updated_at = updated_at` self-assignment** on both of its `is_blocked` writes *(flagged independently by both reviewers)*. `wisps.updated_at` carries `ON UPDATE CURRENT_TIMESTAMP`; the `0056` twin and the runtime recompute both deliberately self-assign so a derived-state repair does not masquerade as a user edit.
4. **Rebase housekeeping.** The branch now conflicts with main in `cmd/bd/create.go` (#4957 landed global-ID prefix validation there), and both migration numbers have since been taken on main (`0056_add_comments_keyset_index`, `ignored/0013_add_wisp_row_lock`), so the two new migrations get renumbered to the next free slots.

Two minors also picked up in the fix commits: the `--mol-type` rejection message says graph plans do not carry molecule semantics, but the per-node `mol_type` field is honored — reworded to point at the node field; and a comment on `TestGraphApplyNodeCoversCreateIssueParams` noting it asserts field presence, not wiring into the `createIssueParams` literal.

One more connection: this PR also resolves the complaint in #4362 (unknown plan keys silently dropped, parent/deps hierarchy lost) via the unknown-field warnings and full parent/deps support, so we are adding `Closes #4362` to the description for it to close alongside #4890.

_Reviewed by claude-opus (reviewer agent) and codex-gpt-5.6-sol-xhigh; verified and adjudicated by claude-fable-5-xhigh on behalf of maphew_

maphew:
Maintainer fix commits pushed, as promised in the review above. Your commits are untouched; five commits were added on top:

1. **Merge of current main** — resolves the `cmd/bd/create.go` conflict with #4957 by making `loadEmbeddedIDPrefixes` delegate to main's tested `selectCreateIDPrefix`, so the `--global` prefix rule now applies to both the single-create explicit-ID path and `--graph` plan validation.
2. **Migration renumber + review fixes** — `0056` → `0059` and `ignored/0013` → `ignored/0015` (both numbers were taken on main while the PR was in flight); the legacy `wisp_dependencies` shape guard; and the `updated_at = updated_at` self-assignment in the wisp repair.
3. **`0059` restructured to direct SQL** — while landing the review fixes, the renumber unblocked main's full-chain migration acceptance test (`TestFullChainFromPreWispsAndMissingDependenciesIDConvergesThroughDoltCLI`), and it caught a latent bug neither reviewer had found: on the Dolt CLI batch path (`dolt sql -q/-f`, the fresh-bundle route) **prepared DML silently fails to apply** — a `PREPARE`'d `UPDATE` reports success and changes nothing (observed on dolt 2.2.0; same limitation class `cli_migrations.go` documents for prepared `ALTER TABLE`). The migration's direct zeroing followed by a dynamic recompute meant every issue's `is_blocked` was zeroed and never restored on that path. The recompute now runs as one direct unconditional statement reading fixed-shape stand-in copies of the wisp tables; the only dynamic statements are best-effort guarded `INSERT..SELECT` copies into the stand-ins, which degrade to issues-only semantics (your fresh-clone compromise) instead of corrupting. The commit message on `28325c654` has the full analysis, including why pre-adding split columns was rejected (it would flip `ignored/0003`'s `@needs_migrate` guard and skip its data backfill).
4. **Explicit-ID collision preflight** — plan validation now rejects an explicit `id` that already exists in the store (the insert path upserts, so a collision silently rewrote the existing issue while reporting creation; inherited from `bd create --id`, but plans mint IDs in bulk). Embedded probes the store; proxied probes through the unit of work inside the apply transaction, so the live path cannot race a concurrent create. Deliberately not `--force`-overridable.
5. **Small polish** — `--mol-type` rejection message now points at the per-node `mol_type` field, the plan supplement documents the no-overwrite rule, and the parity guardrail's doc comment notes it checks field names, not wiring.

Verification: `go build -tags gms_pure_go ./...` clean; targeted `cmd/bd` graph/create/proxied suites, `BEADS_TEST_EMBEDDED_DOLT=1 TestEmbeddedCreate` (full embedded round-trip, 89s), `internal/storage/schema` (including the previously failing full-chain test), `internal/types`, `internal/storage/domain`, and `internal/storage/issueops` all pass; `scripts/check-migration-hygiene.sh` OK. A handful of `internal/storage/dolt` PATH-dolt tests fail identically on current main in this environment (dolt 2.2.0 local quirks) and are left to CI's pinned toolchain.

Thanks again — the core design needed no changes at all; everything above is staleness repair and hardening around the edges of a solid contribution.

Issue #4353, opened 2026-06-10
embedded dolt lib pin predates dolthub/dolt#11126 - bd migrate schema can still flip a TEXT column's storage encoding without rewriting rows (invalid hash length: 19 on read / 1 on insert+merge)


seanmartinsmith:
## tl;dr

dolt changed the default on-disk format for TEXT columns in 2.0.7, and until
dolthub/dolt#11126 (merged 5/30) an `ALTER TABLE ... MODIFY` would silently re-stamp a
column's format tag to match the running binary — without converting the rows. my events
tables got hit twice: once in the 2.0.7/2.0.8 server era (tag flipped to adaptive over
legacy rows → `invalid hash length: 19` on any read that dereferences them — sat silent for
weeks because nothing scans event old/new values), and again last week when a dep-id repair
re-ran migrations through **bd's embedded engine**, whose dolt lib is pinned to May-7 —
**pre-fix** — flipping the tag back the other way (`invalid hash length: 1` on insert/merge;
killed `bd create` and cross-machine merges fleet-wide). no data was ever damaged — every
value is intact under the wrong tag, and i reproduced both flip directions with a probe
built from bd's own go.mod. the asks, in one line: cut a release carrying #4326's pin bump
(main fixed the engine half on 6/8, but every *released* build still ships the pre-fix pin),
guard migration 0048 (still two bare ALTERs on HEAD as of today; 0049 is guarded), and
consider teaching `bd doctor` to spot tag/row drift — the pin bump prevents new flips but
can't heal a DB that already drifted.

## upstream state (dup check, so you don't have to)

- **dolthub/dolt#11131** (closed 5/31) — same `: 19` signature, same mechanism (community
  repro: metadata-only encoding flip on `ALTER ... MODIFY ... LONGTEXT` over a legacy
  `StringAddrEnc` column). **#11017** made adaptive the default in 2.0.7 (the exposure era);
  **#11126** (merged 5/30) preserves legacy encoding on ALTER and adds a
  `dolt admin schema-encoding-drift` heal toolkit (`check`/`repair`/`recover-rows`/
  `migrate-adaptive`). my standalone 2.1.x servers carry the fix (probed: their ALTERs
  preserve); the bd embedded pin doesn't.
- **#4326** (merged to main 6/8): bumps driver→v2 2.1.4 and the embedded dolt lib to a June-5
  snapshot — **post-#11126**, so tip-of-main bd already closes the flip going forward. but
[...]

Issue #4176, opened 2026-05-26
Fresh Dolt-server clone crashes in migration `0047` (`table not found: wisps`) when remote schema cursor is below the binary's latest


seanmartinsmith:
## Summary

A fresh `bd` clone of a Dolt-server-mode project crashes on open with `Error 1146 (HY000): table not found: wisps` during schema migration `0047_recompute_mixed_is_blocked`, whenever the cloned `schema_migrations` cursor arrives **below** the cloning binary's embedded latest. The `wisps` table is `dolt_ignore`'d, so it never syncs to a clone, and the migration runner executes the main sequence (which reaches `0047`, referencing `wisps` unconditionally) *before* the ignored/local sequence that recreates the wisp tables. The database fails to open for every command that touches it — including `bd dolt pull`, `bd show`, `bd list`.

A clone that arrives at the **same** cursor as the binary's latest does **not** hit this (no main migration runs), which is why it only bites during the post-upgrade skew window.

## Affected versions

- **bd:** 1.0.4 (`d7469d5ab`) — also observed on `8ae4c3c67`
- **dolt:** reproduced on 2.0.6; Dolt-version-independent (the failure is an absent table, not a Dolt defect — 2.0.7's go-mysql-server leak fix does not affect this path)
- **Backend:** Dolt sql-server (shared/centralized server mode)
- **Trigger window:** any clone whose remote `schema_migrations` MAX is in roughly `[20, latest-1]`

## Reproduction (minimal)

1. **Producer** has a project whose Dolt remote (`refs/dolt/data`) was last pushed by a `bd` at schema cursor *N* (e.g. 46) — i.e. before the current binary's latest (47). This is the normal state right after a schema bump, before every machine has re-pushed.
2. **Consumer** clones/pulls that remote with a newer `bd` whose embedded latest is 47: `bd bootstrap` (or `bd dolt pull origin` against an existing clone).
3. Consumer's `bd` opens the DB → crashes.

(Equivalently: producer inits on an old schema `< 20`, migrates to HEAD locally — which creates `wisps` *locally*, advances `schema_migrations` to 46+ — configures a git remote and `bd dolt push`. Any second machine that clones then fails. A single-machine `bd init` never hits this, because `wisps` is created locally during init before it is `dolt_ignore`'d.)

## Observed behavior

```
Error: failed to open database: failed to initialize schema:
schema migration: migration 0047_recompute_mixed_is_blocked.up.sql:
[...]

PR #4878, opened 2026-07-17
fix(schema): harden migration chain against missing wisp tables, dropped dependencies.id, and 0048 re-application (#4695, #4690, #4353)

### LINKED ISSUE #4695: Migration 0047 fails on databases without wisps table (pre-1.0 databases upgrading to v53)
## Environment
- bd version: 1.1.0 (Homebrew, stable)
- Database schema: v23 (created pre-1.0, partially migrates to v46)
- Backend: embedded Dolt
- Remote: yes (git+ssh to GitLab)
- OS: macOS ARM64

## Problem

Running `BD_ALLOW_REMOTE_MIGRATE=1 bd migrate` on a database originally created in the pre-1.0 era fails at migration 47:

```
Error: failed to open database: embeddeddolt: init schema: embeddeddolt: migrate:
migration 0047_recompute_mixed_is_blocked.up.sql: Error 1146: table not found: wisps
```

The database never had `wisps` or `wisp_dependencies` tables — they were introduced in a later schema version. The migration script assumes they already exist and tries to query/update them unconditionally.

## Steps to reproduce

1. Have a database at schema v23 (created with bd ~v0.6x era, synced to a remote)
2. Install bd v1.1.0 (stable, via Homebrew)
3. Run `BD_ALLOW_REMOTE_MIGRATE=1 bd migrate`

Migrations 24–46 apply successfully (after a `bd dolt commit` to clear the dirty working set). Migration 47 then fails with table-not-found.

## Attempted workarounds

Manually creating the `wisps` table via `dolt sql` (bypassing bd) gets past the initial error, but then:
- Migration 47 fails on missing `status` column in wisps
- After adding that: fails on missing `wisp_dependencies` table
- After creating that: fails on missing `issue_id` column in wisp_dependencies
- After fixing that: migration 53 (`0053_repair_rig_wisps.up.sql`) fails on missing `issue_type` column

It is a cascade of missing schema that was never present in the database history. The migration chain assumes an intermediate state that this database never passed through.

## Expected behavior

Migration 47+ should either:
- Create the missing tables as part of the migration chain before referencing them (e.g., `CREATE TABLE IF NOT EXISTS wisps ...` prior to the UPDATE/SELECT)
- Skip wisps-related operations when the tables do not exist
- Or the migration runner should detect the gap and route through an alternate migration path for pre-wisps databases

## Current workaround

Using the bd v1.0.3 binary works with schema v23 databases without issue. We downloaded the v1.0.3 release binary and use it for all operations on this database.

## Related issues

- v1.1.0 release notes mention `fix(schema): repair migration-state drift that breaks v53 upgrades` (#4502, #4534) — does not cover the missing-tables case
- v1.1.0 final: `fix(schema): repair v53 wisp dependency drift` (#4558) — possibly related but did not fix this scenario
- v1.1.0-rc.1 warning about #4502: "v53 migration fails on databases whose issues table predates the rig/agent columns" — similar class of problem but different tables

maphew:
## Why

Three separate reports of the schema-migration chain hard-failing or corrupting data on upgrade, all sharing the same root shape (a clone whose schema cursor lands mid-chain, or a table drifted from canonical). Each blocks or damages an upgrade with a raw SQL error:

- **#4695 / #4176** - migration 0047 dies with `table not found: wisps` on databases where the `wisps`/`wisp_dependencies` tables don't exist. Those tables are `dolt_ignore`'d and never sync to a clone; `MigrateUp` runs the main sequence (reaching 0047) before the ignored sequence that would recreate them, so any clone whose cursor arrives below the binary's latest hits 0047's unconditional `WITH RECURSIVE` JOIN on missing tables.
- **#4690** - migration 0053 fails on databases where `dependencies.id` was already dropped; 0053's `REPLACE INTO dependencies (id, ...)` references a column that no longer exists.
- **#4353 (residual)** - migration 0048 is two bare `ALTER TABLE events MODIFY ... LONGTEXT` statements with no idempotency guard. 0049 guards its equivalent MODIFYs; 0048 does not, so a re-applied 0048 through a drifted engine can flip a column's storage-encoding tag without rewriting rows (`invalid hash length: 19` on read).

## What

All fixes follow the established pattern (merged #4538/#4558): shipped migrations are frozen (`check-migration-hygiene.sh` blocks editing them), so repairs go in the pre-migration repair hook (`internal/storage/schema/migration_repairs.go`), and new guarantees go in a new fix-forward migration.

- **#4695/#4176** - v47 repair hook creates missing `wisps`/`wisp_dependencies` at their **forward-canonical shape** (0020/0021 base plus every column/index added by migrations through 0052 that 0047-0053 read: `no_history` from 0023, `started_at` from 0027, `idx_wisp_dep_type` from 0022), or delegates to the existing split-target repair when the tables exist but are pre-split. 0047's split ALTERs are gated and no-op against the repaired shape.
- **#4690** - v53 repair hook restores `dependencies.id` to 0043's canonical end state (`PRIMARY KEY`), backfilling with `depid.New(issue_id, target)` using the same target precedence (`issue → wisp → external`) as `rekeyDependencyIDs`, so the post-migration rekey pass is a true no-op. Restoring the id as a real key closes a `REPLACE INTO` duplicate-edge hole that a plain non-key column would leave open.
- **#4353 residual** - new migration `0056_events_value_columns_idempotent_longtext` mirrors 0049's per-column `INFORMATION_SCHEMA` guard for `events.old_value`/`new_value` (two independent guards + MODIFYs, so mixed TEXT/LONGTEXT states are handled). Note: 0056 closes the gap within bd's cursor-gated migrate path; it cannot protect a raw re-execution of frozen 0048 outside that path, which would need a code-side v48 execution guard (out of scope).
- **Atomicity** - the pre-migration repair mutates synced tables (`dependencies`), but `runMigrations` took its dirty-table snapshot *after* calling the repair, so a repair mutation was misclassified as pre-existing dirt and excluded from the step's atomic commit - a crash could leave committed history claiming v53 while `dependencies.id` was still absent, with the repair never re-running. Fixed by snapshotting before the repair. Verified behavior-neutral for the ~54 migrations with no repair hook (the repair is a genuine no-op there).

## Tests

- `TestFullChainFromPreWispsAndMissingDependenciesIDConvergesThroughDoltCLI` (real dolt CLI) - seeds a DB simultaneously missing the wisp tables **and** `dependencies.id` at cursor 46, drives the full main + ignored sequence to latest, asserts canonical shape / PK / backfill / correct `is_blocked` recompute, and re-run idempotency. This exercises the 0047→0053→0056→ignored path the original per-migration tests skipped.
- `TestEmbeddedMigrateRepairedDependenciesIDColumnCommitsAtomicallyWithVersion53_4690` (real embedded Dolt) - fault-injects a kill right after v53's step-commit and reads `dependencies AS OF HEAD` to prove the repaired column is durably committed with the cursor; fails under the pre-fix ordering.
- Plus targeted repair-hook and 0056 guard tests. `go build`, full `internal/storage/schema` package, and `check-migration-hygiene.sh` all clean.

Reviewed cross-vendor (Claude opus + Codex gpt-5.6); an earlier revision's incomplete wisps shape and missing-PK repair were caught and corrected before this PR. Known nit (not reachable in these bugs' shapes): if a drifted `dependencies` had `id` inside a *composite* PK, the repair leaves the composite PK rather than reasserting a sole-`id` PK.

Closes #4695, #4176, #4690. Addresses the migration-guard residual of #4353.

### DIFF
--- a/internal/storage/embeddeddolt/migrate_repair_atomicity_test.go
+++ b/internal/storage/embeddeddolt/migrate_repair_atomicity_test.go
@@ -1,0 +1,152 @@
+//go:build cgo
+
+package embeddeddolt_test
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/storage/schema"
+)
+
+// TestEmbeddedMigrateRepairedDependenciesIDColumnCommitsAtomicallyWithVersion53_4690
+// is the atomicity regression for the migration-chain-hardening branch's
+// finding #6: preMigrationRepair (registered for main-source version 53,
+// see internal/storage/schema/migration_repairs.go's ensureDependenciesIDColumn,
+// #4690) can itself mutate a synced table -- ALTER TABLE dependencies ADD
+// COLUMN id / MODIFY ... NOT NULL / ADD PRIMARY KEY. Before the fix,
+// runMigrations snapshotted the "before" dirty-table state AFTER the repair
+// ran, so the repair's own mutation to `dependencies` was misclassified as
+// pre-existing dirt to exclude from version 53's per-step commit: the cursor
+// row for v53 would land in one Dolt commit while the repaired
+// `dependencies` table sat uncommitted in the working set, picked up (if at
+// all) only by MigrateUp's much-later end-of-pass catch-all commit. A
+// process killed anywhere between the v53 step-commit and that later commit
+// would leave committed history claiming v53 applied while
+// dependencies.id was never durably recorded -- and the repair cannot
+// re-run to fix it, because its version is no longer pending once the
+// cursor row is committed.
+//
+// The fix (schema.go's runMigrations) snapshots dirty tables BEFORE calling
+// preMigrationRepair, not after, so the repair's mutations are attributed to
+// version 53's own step and land in the SAME atomic commit as its cursor
+// row. This test proves that directly: fault-inject a "kill" immediately
+// after version 53's step-commit (the same seam #4566's self-heal tests use)
+// and read dependencies back AS OF HEAD -- the commit the fault fires right
+// after -- to prove the repaired column is already durably there, not
+// stranded in a working set the injected kill would otherwise lose.
+func TestEmbeddedMigrateRepairedDependenciesIDColumnCommitsAtomicallyWithVersion53_4690(t *testing.T) {
+	requireEmbedded(t)
+	ctx := t.Context()
+	dataDir := seedMainSchemaAt(t, ctx, 52)
+
+	conn, closeConn := openPinnedConn(t, ctx, dataDir)
+	// Reproduce #4690: dependencies already in split-target/content-derived-
+	// key shape (0043 has run, so the uk_dep_* unique keys are present) but
+	// its surrogate id column and primary key are gone -- a different
+	// historical migration path than this repo's 0043, which adds them.
+	if _, err := conn.ExecContext(ctx, "ALTER TABLE dependencies DROP PRIMARY KEY"); err != nil {
+		closeConn()
+		t.Fatalf("drop dependencies primary key: %v", err)
+	}
+	if _, err := conn.ExecContext(ctx, "ALTER TABLE dependencies DROP COLUMN id"); err != nil {
+		closeConn()
+		t.Fatalf("drop dependencies.id: %v", err)
+	}
+	if _, err := conn.ExecContext(ctx, "CALL DOLT_ADD('-A')"); err != nil {
+		closeConn()
+		t.Fatalf("stage dependencies drop: %v", err)
+	}
+	if _, err := conn.ExecContext(ctx, "CALL DOLT_COMMIT('-m', 'test: reproduce #4690 missing dependencies.id')"); err != nil {
+		closeConn()
+		t.Fatalf("commit dependencies drop: %v", err)
+	}
+	closeConn()
+
+	const killAfter = 53
+	restore := schema.SetMigrateStepFaultHookForTest(func(_ context.Context, _ schema.DBConn, version int) error {
+		if version == killAfter {
+			return fmt.Errorf("injected fault: process killed after migration %d", killAfter)
+		}
+		return nil
+	})
+	defer restore()
+
+	conn2, closeConn2 := openPinnedConn(t, ctx, dataDir)
+	_, err := schema.MigrateUp(ctx, conn2)
+	if err == nil || !strings.Contains(err.Error(), "injected fault") {
+		closeConn2()
+		t.Fatalf("MigrateUp err = %v, want the injected fault after migration %d", err, killAfter)
+	}
+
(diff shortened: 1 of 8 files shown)


maphew (commented):

The migration-chain hardening is correct and well-guarded. All three repairs (missing wisp tables, dropped dependencies.id, 0048 re-application) detect their broken states and no-op on healthy databases; each repair is re-entrant/idempotent (WHERE id IS NULL scoping, CREATE TABLE IF NOT EXISTS, per-column INFORMATION_SCHEMA guards); numbering is contiguous after the 0056 to 0057 rebase and auto-registered via the embed glob; the repair fires only for the pending version actually being applied, so it cannot touch a converged database; and the dirty-table snapshot reordering is sound. Verified in particular: the forward-canonical wisps DDL exactly matches 0020 base plus 0023/0027 additions minus 0038 drops; the depid precedence matches rekeyDependencyTable's COALESCE order; and the v47 delegate hybrid path still converges structurally through ignored 0005. Issues #4695, #4690, #4353 all match their fixes. No blocker or major issues.

**Findings:**
- **nit** `internal/storage/schema/schema_test.go:1328` — Only 0057's no-op branch is exercised: on any fresh chain the events columns are already LONGTEXT from 0048, so the MODIFY conversion branch is never run by any test. Suggest seeding a DB with old_value/new_value forced back to TEXT and asserting 0057 converts each independently, including the mixed case the header comment motivates.
- **nit** `internal/storage/schema/schema_test.go:933` — The 0047 delegate path is only covered by a mock with split columns already present; the real legacy interaction (repair adds split columns, 0047 defers the structural split to ignored 0005) is untested end-to-end. Confirmed by inspection to converge, but a CLI test seeding the legacy pre-split shape and asserting canonical constraint convergence would lock it in.

Review performed by a claude-opus-4-8 reviewer subagent as part of an open-PR review sweep.

maphew:
Follow-up: the review's two test-coverage nits were pursued in #4987, and closing them surfaced a real gap in this PR's delegate path. Engine-based tests (which execute the actual frozen migration SQL - the dolt CLI harness turns out to silently no-op PREPARE/EXECUTE-driven DDL, so it could only ever exercise no-op branches) showed that a database taking the `ensureWispDependenciesSplitTargets` delegate path converges structurally but permanently lacks `fk_wisp_dep_wisp_target`, `fk_wisp_dep_issue_target`, and `ck_wisp_dep_one_target`: the ignored-chain restore guards can never fire for genuinely legacy sources. The review's "confirmed by inspection to converge" was wrong on exactly this point. #4987 adds an idempotent healing migration (0058) that also repairs already-affected databases, plus the engine-based coverage.

maphew:
Update: #4987 is now merged (squash). It ships the 0058 healing migration (adds the three missing wisp_dependencies constraints on delegate-path databases, pre-cleaning FK orphans and normalizing zero/multi-target rows per the backfill precedence first, so constraint adds cannot brick store-open) plus engine-based frozen-guard tests that execute the embedded migration SQL production runs. This closes out both review nits from this PR and the constraint gap they uncovered.

PR #4879, opened 2026-07-17
feat(proxied-server): support 18 more commands (13 read + 5 mutation) with embedded-parity tests

coffeegoddd:
## Summary

Adds proxied-server-mode support to 18 more `bd` commands, plus full embedded-parity integration tests. Each command gains a `run<Cmd>ProxiedServer` dual dispatched via `usesProxiedServer()`, backed by the `uow.UnitOfWork` use-case layer instead of the direct `storage.Store`.

### Commands now supported in proxied-server mode

**Tier 1 — read/query (13):** `count`, `history`, `status`, `statuses`, `types`, `lint`, `info`, `stale`, `orphans`, `find-duplicates`, `duplicates`, `epic`, `graph`

**Tier 2 — mutations (5):** `set-state`, `defer`, `undefer`, `unclaim`, `reclaim`

### New use-case / repository methods

Added to `domain.IssueUseCase`/`IssueSQLRepository` (and `ConfigUseCase`), each backed by an `issueops` helper — several widened from `*sql.Tx` to `issueops.DBTX` (the established pattern, mechanical and body-unchanged):

- `CountIssues`, `CountIssuesByGroup` (count)
- `History`, `IterEvents` (history)
- `GetStaleIssues` (stale)
- `GetEpicsEligibleForClosure` (epic)
- `GetLocalMetadata` on `ConfigUseCase` (info `--schema`; reads `local_metadata`, distinct from `metadata`)
- `Unclaim` (unclaim), `ReclaimExpiredLeases` (reclaim)

`epic close-eligible` and `set-state`/`defer`/`undefer`/`unclaim`/`reclaim` writes run inside `uow.RunTx`/`RunTxResult`.

### Bugs surfaced by the parity tests and fixed

1. **`create --no-history` routed to the wrong table.** `create_proxied_server.go` chose `CreateWisp` only on `Ephemeral`, but the canonical rule (`issueops/create.go`) is `Ephemeral || NoHistory`. no_history beads (GC-eligible wisps) were landing in the durable `issues` table — a data-correctness bug in the already-shipped `create` dual, caught by the `count --include-infra` parity test.
2. **No-op status updates failed in proxied mode.** The uow MySQL connection used *changed-rows* semantics, but the domain repo's `Update` uses `RowsAffected()==0 → not found` (which assumes *matched-rows*, like the embedded engine). Re-deferring an already-deferred issue (and any `update --status X` when already `X`) failed with `sql: no rows in result set`. Fixed by setting `ClientFoundRows=true` on the uow DSN — also fixes the already-shipped proxied `update` dual.
3. **`info --json` printed human output.** A redundant local `--json` flag on `info` shadowed the persistent one and got clobbered by the `PersistentPreRunE` reset (in both direct and proxied modes). Removed it.
4. **Custom statuses dropped in proxied mode.** The domain `GetCustomStatuses` read only the `custom_statuses` table, while proxied `config set status.custom` writes just the config row. Now resolves via `issueops.ResolveCustomStatusesDetailedInTx` (table → config → YAML), matching the store and mirroring `GetCustomTypes` — also fixes the proxied `create`/`list`/`purge` paths.

### Tests

- Every command has a `*_proxied_integration_test.go` suite ported to match its embedded counterpart's scenarios (run against a real Dolt sql-server via the shared testcontainers harness).
- Deliberate non-ports: concurrency-stress tests (storage-layer, not command behavior); cross-actor `--force`/`--assigned` cases (the proxied harness strips `BEADS_*`, so a second actor identity can't be set); `duplicates --auto-merge` write path (rejected in proxied mode); `graph check` on a cyclic graph (`dep add` refuses to create a cycle; `DetectCycles` is covered by the dep/link/ready duals).
- `.github/scripts/proxied-cmd-test-shards.txt` updated with committed, subtest-weight-balanced shard assignments for all new `TestProxiedServer*` functions (validated: 63/63 tests assigned, `fallback: 0`, no duplicates).

### Known proxied-mode limitations (consistent with existing duals)

Full issue IDs only (no partial-prefix resolution); cross-repo external-dep routing dropped in `graph` (single-DB); `duplicates --auto-merge` non-dry-run rejected.

## Test plan

- `go build ./...` clean; `cmd/bd` test package compiles.
- Full proxied integration suite green against a real Dolt container (one shared server, ~150 subtests, 0 failures).
- Direct-path (`*_embedded_test.go`) and touched unit suites (`issueops`, `domain`, `domain/db`, `dbproxy/util`) pass.

### DIFF
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -2,84 +2,96 @@
 #
 # Format: <total_shards> <shard_number> <top_level_test_name>
 #
-# 15-shard split, greedily bin-packed by estimated cost
-# (per-subtest project inits dominate; shared-project read subtests are cheap).
-# proxied-test-shard.sh hash-distributes any newly-added test not listed here.
+# 15-shard split, bin-packed longest-processing-time-first by
+# estimated cost (bd-init count; per-init migration chains dominate).
+# Regenerate with scripts/ci/gen_proxied_shard_manifest.py after adding,
+# splitting, or reweighting TestProxiedServer* functions. Newly-added
+# tests not listed here hash-distribute via proxied-test-shard.sh.
 
-15 1 TestProxiedServerUpdate
+15 1 TestProxiedServerChildren
+15 1 TestProxiedServerConfigSetMany
+15 1 TestProxiedServerDeferUndefer
+15 1 TestProxiedServerDelete
+15 1 TestProxiedServerFindDuplicates
+15 1 TestProxiedServerPriority
+15 1 TestProxiedServerUpdateConcurrentClaim
 
-15 2 TestProxiedServerShow
+15 2 TestProxiedServerCloseConcurrent
+15 2 TestProxiedServerDeleteConcurrent
+15 2 TestProxiedServerEpic
+15 2 TestProxiedServerLink
+15 2 TestProxiedServerQuery
+15 2 TestProxiedServerUpdate
+15 2 TestProxiedServerUpdateConcurrentMetadata
 
-15 3 TestProxiedServerCreate
+15 3 TestProxiedServerConfigConcurrent
+15 3 TestProxiedServerDepConcurrent
+15 3 TestProxiedServerGateList
+15 3 TestProxiedServerList
+15 3 TestProxiedServerReadyClaimConcurrent
+15 3 TestProxiedServerUpdate2
 
-15 4 TestProxiedServerClose
+15 4 TestProxiedServerClose2
+15 4 TestProxiedServerContext
+15 4 TestProxiedServerDoltRemoteRemove
+15 4 TestProxiedServerEdit
+15 4 TestProxiedServerReadyConcurrent
+15 4 TestProxiedServerTodo
 
-15 5 TestProxiedServerDep
+15 5 TestProxiedServerAssign
+15 5 TestProxiedServerCreate2
+15 5 TestProxiedServerDuplicates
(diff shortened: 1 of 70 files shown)


idirectships:
I found a narrow count-contract bug while validating the new `info` paths: a zero-value `IssueFilter` includes ephemeral wisps, so `bd info` can report a larger total than `bd status`, JSONL/export, and the durable `issues` table.

I prepared a scoped commit on top of this PR's exact head `10cca499ee5baa47fdff0e96615021ce97be5b5e`:

- commit: https://github.com/idirectships/beads/commit/2a3f3e6a92adf22e773a3a1c74b61530e4dddd5c
- direct and proxied `info` now use `CountIssues(..., IssueFilter{SkipWisps: true})`
- both paths gain a durable-issue-plus-wisp regression
- contributor commits and existing tests are unchanged

Focused validation passed:

```text
BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go -run '^TestEmbeddedInfo$' ./cmd/bd
BEADS_TEST_PROXIED_SERVER=1 CGO_ENABLED=1 go test -tags gms_pure_go -run '^TestProxiedServerInfo$' ./cmd/bd
gofmt
git diff --check
```

GitHub reported `maintainerCanModify=true`, but a direct non-force push to the contributor fork was denied with HTTP 403 for the `idirectships` credential. I therefore did not open a competing PR or alter this branch. If the contract looks right, the commit can be cherry-picked onto `db/more-commands`.

_codex-unknown-model-unknown-reasoning on behalf of Drew Garman_

PR #3971, opened 2026-05-16
fix(filters): Dolt label-filter regression fix — LabelPattern + LabelRegex + JSON truncation hint suppression

quad341:
## What this changes

Fixes two regressions in `bd list` on the Dolt backend.

**Label filter routing** (`--label-pattern` / `--label-regex`): `bd list --label-pattern 'bug*'` and `bd list --label-regex 'P[01]'` were silently ignored — the filter arguments were parsed correctly at the CLI but never wired into the SQL `WHERE` clause in `internal/storage/issueops/filters.go`. After this fix, `BuildIssueFilterClauses` emits the appropriate `LIKE ESCAPE '|'` (for glob patterns) or `REGEXP` (for regex) clause. Four new unit tests lock in the behavior.

**JSON truncation hint** (`--json` mode): When result sets were truncated, `bd list --json` was emitting a human-readable "more results matched" hint to stderr even in machine-readable output mode. Operators piping output to `jq` would see unexpected stderr chatter. The fix suppresses the hint when `--json` is active.

## Why one branch

Both fixes were authored together as part of the same investigation (be-ucslk4 + be-acnquj) and cherry-pick cleanly onto a fresh branch off main. Shipping together keeps the label-filter fix atomic.

## Review notes

- The PG-side label filter wiring (`internal/storage/postgres/issues.go`) is **intentionally excluded** from this PR. That change requires the full PG backend rollup which is gated separately.
- A pre-existing test build conflict in `cmd/bd` under `gms_pure_go,dolt_only,cgo=1` (duplicate `test_helpers` symbols) is pre-existing and not introduced here. The affected tests are in a different package from the label-filter changes.
- Gate checklist: [`release-gates/be-ucslk4-dolt-ship-gate.md`](release-gates/be-ucslk4-dolt-ship-gate.md)

## Test plan

- [x] `go test ./internal/storage/issueops/` — PASS (covers `TestBuildIssueFilterClauses_LabelPattern`, `_LabelPatternWispsTable`, `_LabelRegex`, `TestGlobToLikePattern`)
- [x] `make build` clean
- [x] `golangci-lint run ./internal/storage/issueops/ ./cmd/bd/` — 0 issues
- [ ] Reviewer: run `bd list --label-pattern 'bug*'` and confirm non-empty results on a Dolt-backend city

PR #4882, opened 2026-07-18
fix(storage): wire --label-pattern into the WHERE builder

athosmartins:
## What

`bd list --label-pattern` (and `bd list --ready --label-pattern`, which
shares the same filter) now actually filters by the glob pattern. Before
this change, a non-matching pattern silently returned every issue instead
of an empty/filtered set.

## Why

`IssueFilter.LabelPattern` / `WorkFilter.LabelPattern` were parsed from the
`--label-pattern` flag and stored on the filter struct, but neither
`BuildIssueFilterClauses` (`internal/storage/sqlbuild/filter.go`) nor
`BuildReadyWorkWhere` (`internal/storage/sqlbuild/ready.go`) ever read the
field — it was dead code. Concretely:

```
$ bd list --label-pattern 'warrant*' --json
```

returned 5 arbitrary issues sharing no substring with `warrant*`, instead of
`[]`. Root-caused by grepping for every reference to `LabelPattern` in
`internal/storage` — there was exactly one write site (the CLI flag parse)
and zero read sites.

Fix: wire `LabelPattern` into a label-subquery clause in both builders,
mirroring the existing `Labels`/`LabelsAny`/`ExcludeLabels` idiom. The glob
(`*`, `?`) is translated to a SQL `LIKE` pattern, escaping any literal `%`,
`_`, or `\` in the input first so they can't act as unintended wildcards
(new `globToSQLLike` helper, with its own table-driven test).

### Scope note: `--label-regex` is NOT touched here

`IssueFilter`/`WorkFilter` also carry a `LabelRegex` field with the
identical dead-code bug. An earlier version of this branch fixed it too,
via SQL `REGEXP` — but `engdocs/ICU-POLICY.md` states plainly that "bd does
not use SQL REGEXP functions," which is the explicit justification for
shipping without libicu at all. Introducing real `REGEXP` usage undercuts
that invariant, even though it happens to still work under `gms_pure_go`.
Fixing `--label-regex` properly wants a Go-side `regexp.MatchString`
evaluation instead of a SQL push-down — a different enough implementation,
and outside what ga-hqchm reported, that per `CONTRIBUTING.md`'s "one issue
per PR" rule it belongs in its own PR rather than piggybacking here. Left
`LabelRegex` exactly as dead-code-but-harmless as it was before this PR.

### Also noticed, not fixed here

The standalone `bd ready` subcommand doesn't expose `--label-pattern`/
`--label-regex` as flags at all (confirmed via `bd ready --help`) — separate
from this bug. `bd list --ready`, which shares `WorkFilter`/
`BuildReadyWorkWhere`, works correctly with this fix. Flagging in case a
maintainer wants a tracking issue; happy to file one if useful.

## Verification

```
go test ./internal/storage/sqlbuild/...          # new + existing tests, all pass
go build -tags gms_pure_go ./...                  # canonical build, clean
go vet -tags gms_pure_go ./...                     # clean
otool -L bd | grep -i icu                          # empty — binary stays ICU-free
```

Also verified end-to-end against a real built binary (not just unit tests),
against a scratch workspace:

```
$ bd list --label-pattern 'tech-*' --json   # issue labeled tech-debt
[{"id":"e2e-bkh", ...}]
$ bd list --label-pattern 'warrant*' --json # no issue has a matching label
[]
$ bd list --ready --label-pattern 'tech-*' --json
[{"id":"e2e-bkh", ...}]
$ bd list --ready --label-pattern 'warrant*' --json
[]
```

Strengthened the existing `label_pattern` case in
`cmd/bd/list_embedded_test.go`, which only asserted the matching issue was
present and never that a non-matching issue was absent — which is why it
kept passing despite the filter doing nothing.

(Note: `tests/regression/discovery_test.go` already had
`TestDiscovery_LabelPatternFilterDeadCode` tracking this exact bug, but
that file is explicitly excluded from CI via its `regression && discovery`
build tag, per its own header — "intentionally fail while known bugs are
still open... not in CI gates." Left it as-is since graduating/pruning
discovery-tracked bugs looked like a maintainer convention call, not mine
to make; the new coverage in `sqlbuild_test.go` and `list_embedded_test.go`
is what actually gates CI now.)


athosmartins:
Rebased onto `main` (merge, not force-push) to pick up 369751643 ("docs(recovery): drop the dead bd-heartbeat link from the squash runbook"), which fixes the sole CI failure on this PR (`TestDocsSiteLinks` — broken `/cli-reference/heartbeat` link). That failure was pre-existing on `main` at the time this branch forked and unrelated to this PR's diff (only touches `internal/storage/sqlbuild/*` and `cmd/bd/list_embedded_test.go`). Verified locally post-merge: `TestDocsSiteLinks` now passes, and the sqlbuild label-pattern tests (`TestBuildIssueFilterClausesLabelPattern`, `TestBuildReadyWorkWhereLabelPattern`) still pass. No changes to the actual fix.

maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the `--label-pattern` wiring is real progress on BUG-28, but the PR leaves its sibling flag as an advertised no-op.

**Blocker — `--label-regex` is still a silent no-op.** Verified at the PR head: `bd list` advertises regex filtering and `list_filter.go` stores the value, but grep finds zero consumers of `LabelRegex` anywhere in `internal/storage/` — only `LabelPattern` reaches the WHERE builder and `ready.go`. The checked-in BUG-28 description (`tests/regression/DISCOVERY.md`) covers *both* flags. A user passing `--label-regex` gets unfiltered results with no warning — the exact failure class this PR exists to fix, one flag over. Either implement it (Go-side regex per the ICU policy, since the pinned go-mysql-server has no regex-safe LIKE) or fail closed / remove the flag until it exists.

**Should-fix — the glob→LIKE conversion drops bracket expressions.** The converter translates only `*` and `?`; `[dl]` stays literal. The original SQLite implementation used `GLOB`, under which `tech-[dl]*` matched `tech-debt` and `tech-legacy` — the pinned go-mysql-server LIKE recognizes only `%`, `_`, and escapes, so bracket classes can't survive this translation. Worth either emulating brackets in the conversion, documenting the narrowed semantics, or routing bracket-bearing patterns to the same Go-side matcher as regex. Current tests cover only `back*`-style patterns, so the regression is invisible.

Nits: the duplicated predicate construction in `filter.go`/`ready.go` could share a helper; one commit message omits the issue-suffix convention.

Merge-base drift is clean (pagination/blocked-filter work doesn't collide). Land the pattern wiring by all means — just don't leave `--label-regex` advertising something the storage layer never sees.

maphew (approved):
Approving — this supersedes my 2026-07-23 changes-requested review. Since upstream #3971 landed the list-path implementation of `--label-pattern`/`--label-regex` while this PR was open, I've reshaped the branch with maintainer commits rather than asking for another round:

- Merged main; the `filter.go`/`list_embedded_test.go` hunks resolve to main's #3971 implementation.
- Your surviving unique value — the ready path — is upgraded to main's idiom: `BuildReadyWorkWhere` gains LabelPattern via `globToLikePattern` + `LIKE ? ESCAPE '|'` and a LabelRegex `REGEXP ?` clause, with clause-level tests.
- Both flags are registered on the standalone `bd ready` subcommand (absorbing the surviving scope of your #4884, credited in the commit).
- A cross-vendor review gate caught that the flags were dead under `--proxied-server` (reads after the early return, `gatherReadyInput` not copying them) — fixed and pinned with a `gatherReadyInput` unit test plus a proxied integration subtest covering both filters.

Validation: build/vet/gofmt clean, sqlbuild suite, ready input test, and the proxied subtest all green locally. Merge tail goes to the pr-babysit patrol. Thanks athosmartins.

PR #4884, opened 2026-07-18
fix(storage): wire --label-regex into the query pipeline; register on bd ready

athosmartins:
## What

- Fixes `--label-regex` being silently ignored (parsed but never applied) across every `bd list`/`bd ready` row-fetch path — the `LabelRegex` analog of ga-hqchm/#4882's `LabelPattern` fix, implemented as a Go-side post-fetch filter since SQL `REGEXP` isn't available here (see Design below).
- Adds `--label-regex` flag registration to the standalone `bd ready` subcommand, which didn't expose it (or `--label-pattern`) at all.

## Why

Follow-up from ga-hqchm/#4882, split out per `CONTRIBUTING.md`'s one-issue-per-PR rule — both #4882's own scope note and this bug's description call this out explicitly. Two adjacent gaps:

1. `IssueFilter.LabelRegex`/`WorkFilter.LabelRegex` were parsed from `--label-regex` and stored on the filter structs, but no query builder ever read them — a non-matching pattern silently returned every issue, the exact shape #4882 fixed for `LabelPattern`.
2. The standalone `bd ready` subcommand doesn't register `--label-pattern`/`--label-regex` as flags at all (confirmed via `bd ready --help`), separate from the dead-filter bug above.

## Design: Go-side filter, not SQL

Unlike `LabelPattern` (which #4882 pushes down as a SQL `LIKE` clause), `LabelRegex` can't be pushed into the WHERE clause: `bd` does not use SQL `REGEXP` functions (`engdocs/ICU-POLICY.md` — the explicit justification for shipping without libicu). So this fix applies `regexp.MatchString` as a post-fetch, in-Go filter over already-hydrated issue labels.

That has one real architectural consequence, handled explicitly: filtering *after* the SQL layer means the SQL-level `LIMIT` can't run before the regex filter without risking dropping real matches that happen to sort after the first `--limit` *unfiltered* rows — silently wrong, no error, the exact bug class this whole chain exists to eliminate. So wherever `LabelRegex` is set, this PR suppresses the SQL `LIMIT`, fetches every row matching every *other* (SQL-pushed) filter, regex-filters in Go, then applies `--limit` itself. Verified end-to-end below that `--label-regex X --limit 1` returns a real match, not an empty/wrong result.

## Scope: which paths are fixed

Fixed, both the plain and `--json`/counts variants — `bd list --json` and `bd ready --json` turned out to route through entirely separate row-fetch/hydrate/sort/truncate implementations from the non-json path, which the first pass of this PR missed until manual end-to-end testing caught it returning unfiltered results under `--json`:

- `bd list --label-regex` (`SearchIssuesInTx` / `searchTableInTxT`)
- `bd list --label-regex --json` (`SearchIssuesWithCountsInTx`)
- `bd ready --label-regex` / `bd list --ready --label-regex` (`GetReadyWorkInTx`)
- `bd ready --label-regex --json` / `bd list --ready --label-regex --json` (`GetReadyWorkWithCountsInTx`)

Explicitly **not** fixed, documented inline at each site:

- `CountIssuesInTx` / `CountReadyWorkInTx` — `SELECT COUNT(*)`-only paths that never fetch or hydrate rows. Applying `LabelRegex` here would mean fetching every candidate just to discard most of it, defeating the point of a cheap count. A "Showing X of N" hint combined with `--label-regex` will overcount N (it counts pre-regex matches). This does **not** affect the correctness of the issues actually returned — only that informational total.
- `IterIssues`/`IterWisps` (`internal/storage/dolt`) — Go library API surface (`storage.Iter[T]`), not reachable from any `bd` CLI command.
- `--label-pattern` (glob) on `bd ready` — deliberately **not** registered as a flag yet. Unlike `--label-regex`, `WorkFilter.LabelPattern` still isn't read by `BuildReadyWorkWhere` until #4882 lands; registering it now would silently no-op — reproducing the exact dead-flag bug this PR exists to eliminate, just on a third callsite. Trivial follow-up once #4882 merges (mirrors this PR's `--label-regex` registration in shape).

## Testing

```
go test ./internal/storage/sqlbuild/... ./internal/storage/issueops/...   # all pass
go build -tags gms_pure_go ./...                                          # clean
go vet -tags gms_pure_go ./...                                            # clean
otool -L bd | grep -i icu                                                 # empty — binary stays ICU-free
```

New coverage: table-driven unit tests for the regex-matching primitives (`label_regex_test.go`, mirrors `TestGlobToSQLLike`'s style), a focused test proving `buildReadyWorkPredicates` suppresses the SQL `LIMIT` exactly when `LabelRegex` is set (`TestBuildReadyWorkPredicates_SuppressesLimitSQLWhenLabelRegexSet`), and a `finishSearchIssuesWithCounts` test covering the `--json` path directly (`TestFinishSearchIssuesWithCountsAppliesLabelRegex`).

Also verified end-to-end against a real built binary, against a scratch workspace (issues labeled `tech-debt`, `tech-legacy`, `frontend`, `biotech-debt`):

```
$ bd list --label-regex '^tech-' --json
{"id":"e2e-djm","labels":["tech-debt"]}
{"id":"e2e-g39","labels":["tech-legacy"]}
$ bd list --label-regex 'nonexistent-.*' --json
[]
$ bd list --label-regex '[invalid' --json
Error: invalid --label-regex pattern "[invalid": error parsing regexp: missing closing ]: `[invalid`

$ bd ready --label-regex '^tech-' --plain
📋 Ready work (2 issues with no active blockers):
1. [● P2] [task] e2e-djm: Tech debt issue
2. [● P2] [task] e2e-g39: Legacy tech issue

$ bd ready --label-pattern 'tech-*' --plain
Error: unknown flag: --label-pattern

$ bd ready --label-regex '^tech-' --json --limit 1
{"id":"e2e-djm","labels":["tech-debt"]}   # a real match — proves limit-then-filter isn't silently dropping matches
```

### DIFF
--- a/cmd/bd/ready.go
+++ b/cmd/bd/ready.go
@@ -89,6 +89,7 @@ This is useful for agents executing molecules to see which steps can run next.`,
 		labels, _ := cmd.Flags().GetStringSlice("label")
 		labelsAny, _ := cmd.Flags().GetStringSlice("label-any")
 		excludeLabels, _ := cmd.Flags().GetStringSlice("exclude-label")
+		labelRegex, _ := cmd.Flags().GetString("label-regex")
 		issueType, _ := cmd.Flags().GetString("type")
 		issueType = utils.NormalizeIssueType(issueType) // Expand aliases (mr→merge-request, etc.)
 		parentID, _ := cmd.Flags().GetString("parent")
@@ -141,6 +142,7 @@ This is useful for agents executing molecules to see which steps can run next.`,
 			Labels:           labels,
 			LabelsAny:        labelsAny,
 			ExcludeLabels:    excludeLabels,
+			LabelRegex:       labelRegex,
 			IncludeDeferred:  includeDeferred,  // GH#820: respect --include-deferred flag
 			IncludeEphemeral: includeEphemeral, // bd-i5k5x: allow ephemeral issues (e.g., merge-requests)
 			ExcludeTypes:     excludeTypes,
@@ -754,6 +756,12 @@ func init() {
 	readyCmd.Flags().StringSliceP("label", "l", []string{}, "Filter by labels (AND: must have ALL). Can combine with --label-any")
 	readyCmd.Flags().StringSlice("label-any", []string{}, "Filter by labels (OR: must have AT LEAST ONE). Can combine with --label")
 	readyCmd.Flags().StringSlice("exclude-label", []string{}, "Exclude issues that have ANY of these labels")
+	readyCmd.Flags().String("label-regex", "", "Filter by label regex pattern (e.g., 'tech-(debt|legacy)')")
+	// --label-pattern (glob) is intentionally NOT registered here yet: unlike
+	// --label-regex, BuildReadyWorkWhere doesn't read WorkFilter.LabelPattern
+	// until #4882 lands (ga-hqchm's SQL LIKE wiring), so exposing it on this
+	// command now would silently no-op — the exact dead-flag bug ga-7r884
+	// exists to eliminate, just on a third callsite. Add it once #4882 merges.
 	readyCmd.Flags().StringP("type", "t", "", "Filter by issue type (task, bug, feature, epic, decision, merge-request). Aliases: mr→merge-request, feat→feature, mol→molecule, dec/adr→decision")
 	readyCmd.Flags().String("mol", "", "Filter to steps within a specific molecule")
 	readyCmd.Flags().String("parent", "", "Filter to descendants of this bead/epic")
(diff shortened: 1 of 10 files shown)


maphew:
Closing this one with thanks — its problem is now solved upstream, in two parts, and your ready-path work is credited in the landing:

- **The list-path half** (wiring `--label-regex` into the query pipeline) landed via #3971, which implements label regex as a SQL `REGEXP` subquery in `BuildIssueFilterClauses`. That merge also settles this PR's core design question: the Go-side post-fetch filter, LIMIT-suppression, and count-gap machinery here were built on the premise that SQL `REGEXP` wasn't available on the embedded backend — green CI on #3971 shows it is, so the simpler SQL approach won.
- **The ready-path half** (your unique surviving scope: registering `--label-regex`/`--label-pattern` on `bd ready` and honoring them in ready-work queries) is landing through your own #4882, where I've added maintainer commits that wire both filters into `BuildReadyWorkWhere` and register both flags on the standalone `bd ready` subcommand — with credit to this PR in the commit message.

So nothing of value here is lost: half was independently implemented, and the other half ships under your name via #4882. Sorry these two sat unreviewed long enough for the queue to catch up with them — that's on us, not you.

PR #4885, opened 2026-07-18
fix(storage): route bd history through a long-timeout connection (ga-ahnxx)

athosmartins:
## What

`bd history <id>` routes its `dolt_history_issues` scan through a dedicated one-shot connection with a 5-minute read timeout, instead of the shared connection pool's hardcoded 10s `ReadTimeout`.

## Why

`DoltStore.History()` ran its query through the shared pool (`buildServerDSN`'s 10s `ReadTimeout`). Issues with many revisions (thousands, in the case that surfaced this) can take 5-13s+ to scan, so the client-side 10s deadline fires intermittently — surfacing as a confusing "MySQL i/o timeout / invalid connection" error on `bd history` while `bd show` on the *same* id succeeds instantly, because `bd show` is a fast PK point-lookup on the live `issues` table, not a scan of the history table.

The fix mirrors the existing `execWithLongTimeout`/`openLongTimeoutConn` pattern already used for `DOLT_PUSH`/`PULL`/`FETCH` — those operations legitimately run long too, and already get a dedicated connection rather than fighting the shared pool's timeout.

This only removes the *client-side* ceiling. The Dolt server's own `read_timeout_millis` (30s, deliberately short to bound orphaned-connection pileup) still applies server-side and is unaffected — a `dolt_history_issues` scan whose per-row production stalls past that window can still fail via that separate, server-side mechanism. That's a distinct, harder problem (likely needs a query-side fix to avoid `ORDER BY`-driven full-buffer-before-first-row behavior, or a deliberate server-config tradeoff) and is intentionally out of scope here.

Also confirmed while investigating: the `max1Row`-workaround subquery wrapper added in a prior commit is *not* a contributing cause — direct-vs-wrapped timing was statistically indistinguishable (both 5-13s+ against a live db with many-revision issues).

## Verification

- `go build ./internal/storage/dolt/...` and `go vet ./internal/storage/dolt/...` — clean.
- Added `TestHistory_UsesDedicatedLongTimeoutConnection`: creates an issue, confirms `History` succeeds, then corrupts `store.connStr` to an address that fails DNS fast (RFC 2606 `.invalid`) and confirms `History` now fails — proving it re-resolves `connStr` per call via `openLongTimeoutConn` rather than silently reading through the still-healthy shared pool (which would mask a regression back to the 10s ceiling this PR removes).
- Verified against a live Dolt db with an issue carrying ~9000 revisions via a throwaway standalone program: reproduced the original 10s timeout pre-fix, confirmed it's gone post-fix. That same run did trip the independent 30s *server-side* `read_timeout_millis` once (errno 1105) — expected, and explicitly called out above as unaffected/out of scope for this PR.
- This sandbox has no Docker, so the package's own container-gated suite can only skip here (not run) — CI should run it.


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — right diagnosis (ga-ahnxx timeouts are real), one branch-correctness gap in the dedicated connection.

**Blocker — the long-timeout connection loses the store's checked-out branch.** Verified at the PR head: `openLongTimeoutConn` only re-parses the DSN and raises `ReadTimeout`; `withReadTxLongTimeout` then runs `HistoryInTx` on that brand-new session. Dolt resolves an unqualified database on a fresh session to the *default* branch until that session runs `dolt_checkout` — but `DoltStore.Checkout` is session-scoped (`DOLT_CHECKOUT` + `s.branch`). So after an in-process checkout to a feature branch, `bd history` on this path silently reads main's history — missing or stale entries, no error. The fix is small: after opening, select the store's active branch on the same session (`CALL DOLT_CHECKOUT(s.branch)` or a branch-qualified database name) before the transaction begins.

**Should-fix — the regression test can pass while returning nothing.** The test asserts only that `History` returns no error; since the test harness checks out a unique session-local branch while the new connection defaults to main, an *empty* result passes. Assert the created issue's content is present in the returned history — that assertion would have caught the blocker above.

Wiring verified: `cmd/bd/history.go` and `restore.go` do route through `DoltStore.History`; merge-base drift to current main is clean (the proxy-history generalization doesn't collide).

With the branch selection added and the test asserting content, this is a solid fix for a real intermittent failure.

maphew (approved):
Approving — this supersedes my 2026-07-23 changes-requested review. Since the author hasn't been back, I've applied the fixes as maintainer commits on this branch (fix-merge per our maintainer policy, attribution preserved):

- `a4f5b28bb`: the blocker — the dedicated long-timeout connection now selects the pooled session's *actual* branch (`SELECT active_branch()` on the pooled session, then `DOLT_CHECKOUT` on the fresh connection before the read tx; falls back to the store's tracked branch if the query fails). Reading the live session branch rather than the store field also covers checkouts made behind the store's back.
- The regression test now asserts the created issue's content is returned; verified it fails with the checkout call removed and passes with it — it genuinely guards the bug.
- The deployment-specific comment ("this city's dolt-config.yaml") is reworded neutrally.

Validation: build, vet, gofmt, and the history test set green under embedded Dolt; independent review + a cross-vendor Codex gate both signed off on the maintainer commits. Merge tail goes to the pr-babysit patrol once checks are green on the new head. Thanks for isolating the timeout path cleanly, athosmartins — the dedicated-connection design is the right shape.

PR #4886, opened 2026-07-18
fix(dolt): stop misreporting an interrupted fsck as dangling-chunk corruption (bd-f2b15)

steveyegge:
## Summary

`classifyFSCKFailure` (internal/storage/dolt/store.go) routed **any** non-empty fsck output that wasn't a could-not-open phrasing to `ErrDanglingReference`. When the bd process group is killed mid-fsck (observed: a background `bd dolt push` stopped during the bd-lrgn1 cutover), `dolt fsck` prints the literal text `context canceled` to its combined output before dying — so the user saw:

    Error: dangling chunk reference: aborting push to prevent propagating corrupt chunks: context canceled

a false corruption report for a plain interrupt. Same genus as the dolthub/dolt#10915 could-not-open misclassification that branch (a) already guards against. The store was fine (the remote pushed clean minutes later).

## Fix

- New `fsckOutputInterrupted` helper recognizes cancellation phrasings in fsck output: `context canceled`, `context deadline exceeded`, `signal: killed`, `signal: terminated`.
- Branch (a) now falls through to the context-state branches for such output instead of declaring corruption — so parent-cancel still maps to plain cancellation (b), caller deadline to caller-timeout guidance (c), fsck's own timeout to `BEADS_FSCK_TIMEOUT` guidance (d).
- New branch (e): cancellation phrasing with **no** live context error (process group killed out from under fsck — both contexts look healthy from the classifier's view) returns a plain cancellation error wrapping `context.Canceled`, not `ErrDanglingReference`.
- Real corruption output (no cancellation phrasing) still aborts the push exactly as before.

## Tests

- bd-f2b15's required pin: `output='context canceled'` + `parentErr=nil` must NOT map to `ErrDanglingReference` — plus `signal: killed`, fall-through-to-(b), fall-through-to-(d), and a negative sentinel that real dangling-reference output still aborts.
- `TestFsckOutputInterrupted` table covering all four phrasings and non-interrupt output.
- Full `./internal/storage/dolt/` package green locally; `go vet` + `go build ./...` clean.

Closes bd-f2b15.

### DIFF
--- a/internal/storage/dolt/fsck_test.go
+++ b/internal/storage/dolt/fsck_test.go
@@ -200,7 +200,59 @@ func TestClassifyFSCKFailure(t *testing.T) {
 			wantIsNot:    []error{ErrDanglingReference},
 			wantContains: "BEADS_FSCK_TIMEOUT",
 		},
-		// (e) Generic non-zero exit, empty output, no context error → ErrDanglingReference.
+		// (e) Cancellation phrasing in output with no context error (bd-f2b15
+		//     pin): a killed process group leaves "context canceled" in the
+		//     output while both contexts look healthy. Must NOT be reported
+		//     as corruption.
+		{
+			name:         "output 'context canceled' + nil ctx errors → interrupted, not ErrDanglingReference",
+			parentErr:    nil,
+			fsckErr:      nil,
+			output:       "context canceled",
+			wantIs:       context.Canceled,
+			wantIsNot:    []error{ErrDanglingReference, ErrFSCKTimeout},
+			wantContains: "interrupted",
+		},
+		{
+			name:         "output 'signal: killed' + nil ctx errors → interrupted, not ErrDanglingReference",
+			parentErr:    nil,
+			fsckErr:      nil,
+			output:       "signal: killed",
+			wantIs:       context.Canceled,
+			wantIsNot:    []error{ErrDanglingReference, ErrFSCKTimeout},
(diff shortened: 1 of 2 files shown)


steveyegge:
**Review verdict (lion): APPROVE — no changes requested.** (Posting as a comment; GitHub refuses a formal approval since the PR author is the same account as my gh auth.)

Reviewed the diff and independently verified in a fresh checkout of the PR head: full `./internal/storage/dolt/` package green (`-count=1`), `go vet` clean, and all six new classifier cases plus the `TestFsckOutputInterrupted` table pass.

The logic is right, and importantly it stays fail-closed: every branch except could-not-open still aborts the push, so the worst case for a mixed output (real corruption finding AND cancellation noise in the same combined output, e.g. fsck killed mid-report) is a mislabeled abort — the retry re-runs fsck and reports the corruption honestly. No propagation window is opened.

Points I checked specifically:
- The (a) content-wins race comment still holds for genuine findings: cancellation phrasing is the only thing that falls through, and it is by definition not an integrity finding.
- The new (e) branch correctly covers fsckErr=context.Canceled with a nil parent (killed process group) — previously that mapped to a false ErrDanglingReference.
- Substring phrasings match Go's literal error strings ("context canceled", not the double-l spelling), and a hex chunk hash can't collide with them.
- The negative sentinel pins that dangling-reference output without cancellation noise still aborts as corruption.

PR #4888, opened 2026-07-18
docs(recovery): squash fence must cover every syncing machine's scheduled jobs (bd-g3di3)

steveyegge:
## Summary

Amendment to the history-squash runbook from its first production run (the wy-7lmdp window, 2026-07-18). The window was executed with every **local** writer fenced — agents, launchd pushers, the server itself — but a **peer machine's** scheduled bridge-sync job was never stopped. One tick after the clean force-push landed, that job pulled the new baseline, cross-merged it into its old chain, and pushed: the freshly replaced remote went from 65MB back to 16.4GB within a minute. (Recovered without data loss: re-fence, capture the delta, re-wipe, re-push, re-clone.)

The runbook said "stop every writer" but nothing told the operator that writers include scheduled sync jobs on other machines, or that a single live peer undoes the whole procedure.

## Changes

- **Warning**: fenced window means writers stopped *on every machine that syncs this database*.
- **Step 1**: enumerate the machines and their sync units by name and verify each is stopped; explains the one-live-peer failure mode (pull baseline → cross-merge → push old history back).
- **Step 5**: re-enable each machine's sync jobs only after that machine has re-cloned; unfence writers last.

Credit: hole found and recovered by crow (wyvern rig); amendment is his ask.

## Gates

- `go test ./test/docsync` green; `check-doc-freshness.sh` green.

Closes bd-g3di3.

### DIFF
--- a/docs/recovery/history-squash.md
+++ b/docs/recovery/history-squash.md
@@ -14,7 +14,8 @@ the old history becomes collectable, without touching your live data.
 <Warning>
 This procedure rewrites history. Every other clone of the database becomes
 unmergeable and must re-clone, and remotes and backups must be re-pointed.
-Run it in a fenced window: all writers stopped, backup verified first.
+Run it in a fenced window: all writers stopped **on every machine that syncs
+this database**, backup verified first.
 </Warning>
 
 ## Symptoms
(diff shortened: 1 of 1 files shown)


steveyegge:
**Review verdict (lion): APPROVE — no changes requested.**

The amendment is the field incident's lesson correctly generalized, not just patched:

- The Warning now scopes the fence to *every machine that syncs the database*, which is the actual invariant — the original wording let a reader believe fencing the squash machine sufficed.
- Step 1 names the writer class that actually bit (scheduled sync units — cron/launchd/systemd), requires enumerating machines and units by name with per-unit verification, and the spelled-out failure mode is mechanically accurate: the peer's pull cross-merges old chain + new baseline, and its push then lands as a fast-forward of the fresh remote (the merge commit has the new baseline as ancestor), resurrecting the entire old history one tick later.
- Step 5's re-enable ordering (each machine's sync jobs only after that machine re-clones; writers last) closes the same hole on the exit path, and the language stays consistent with the doc's existing cross-merge explanation.

Docs-only, no code surface. Merge when ready.

PR #4889, opened 2026-07-18
fix(gc): prune remote-tracking refs before post-squash GC so flatten/compact actually reclaim (bd-agctw)

steveyegge:
## Problem (bd-agctw, P1)

`bd flatten` and `bd compact` squash history and then run `DOLT_GC` — but neither deletes remote-tracking refs first, and any workspace that has ever pushed or fetched has them. GC treats those refs as live roots, so on exactly the workspaces these commands target (big, long-lived, synced) the GC reclaims ~nothing and the operator gets no cue. Observed in prod: the wyvern store stayed ~10G through a compression attempt (2026-07-17); lion independently diagnosed the same hole for the manual runbook path in #4864.

## Fix

- **versioncontrolops**: new `ListRemoteRefs` / `PruneRemoteRefs` / `ListTags` (`dolt_remote_branches` + `CALL DOLT_BRANCH('-D','-r', ...)`; tags list-only). `Flatten`/`Compact` doc comments now name the required prune so future callers don't reintroduce the bug.
- **storage**: optional `RemoteRefPruner` interface, implemented on `DoltStore` and `EmbeddedDoltStore` (same conn-pinning discipline as `DoltGC`).
- **`bd flatten` / `bd compact`**: prune between the squash and the GC; report store size before → after (freed bytes) so a no-op reclaim is visible; warn about tags (they anchor history the same way but are user-created and never deleted); surface refs/tags in `--dry-run` and JSON (`remote_refs_pruned`, `tags_anchoring`, `size_*_bytes`).
- **`bd gc` (audited, deliberately not pruning)**: `bd gc` never squashes — its Compact phase is informational only — and its remote-tracking refs double as the smart migrate gate's cached remote tip (`AS OF 'remotes/origin/main'`, #4516). Pruning there would weaken the gate for zero reclaim. It now reports size before → after and notes anchoring refs/tags with a pointer to flatten/compact.

Pruning touches only the local ref cache: nothing is deleted on the remote, and the next push/fetch re-creates the refs at the new tip.

## Testing

- New real-Dolt regression test `TestDoltNew_PruneRemoteRefs_RealDolt`: proves the pre-flatten chain stays reachable through `remotes/origin/main` after `Flatten` (the bug precondition), that prune removes the anchor (listing empty, `AS OF` read fails), tags survive, GC runs clean, and prune is idempotent.
- E2E on a scratch workspace: `bd flatten --force` printed `Pruned 1 remote-tracking ref(s): remotes/origin/main` and `Store: 1.7 MB → 42.3 KB (freed 1.6 MB)` — space previously retained silently. `bd gc` now shows `freed 0 B` when there is nothing to reclaim.
- `versioncontrolops`, `embeddeddolt` suites green locally; `golangci-lint` 0 issues. Full `internal/storage/dolt` and `cmd/bd` suites exceed the local 10m `go test` default (hence the sharded CI lanes); 30m solo runs in progress locally, CI is authoritative.

### DIFF
--- a/cmd/bd/compact_dolt.go
+++ b/cmd/bd/compact_dolt.go
@@ -33,7 +33,9 @@ How it works:
   2. Creates a squashed base commit from all old history
   3. Cherry-picks recent commits on top
   4. Swaps main branch to the compacted version
-  5. Runs Dolt GC to reclaim space
+  5. Prunes remote-tracking refs (they would keep the old history alive;
+     the next push or fetch re-creates them at the new tip)
+  6. Runs Dolt GC to reclaim space
 
 Examples:
   bd compact --dry-run               # Preview: show commit breakdown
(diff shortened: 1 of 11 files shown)


steveyegge:
**Review verdict (lion): APPROVE — no changes requested.**

Independently verified in a fresh checkout of the PR head: `versioncontrolops` and `embeddeddolt` suites green, `TestDoltNew_PruneRemoteRefs_RealDolt` passes against real Dolt (`-count=1 -v`), `go vet` clean over `cmd/bd` and `internal/storage/...`. The regression test is the right shape — it proves the bug precondition (old chain reachable through `remotes/origin/main` after Flatten), not just the fix, and pins the anchor as actually gone (`dolt_log` errors) rather than merely unlisted, with tags surviving and prune idempotent.

**On the flagged judgment call — `bd gc` not pruning: I agree.** Without a squash the remote-tracking refs are ancestors of (or at) local main, so they anchor nothing GC could otherwise reclaim — pruning there would weaken the #4516 migrate gate's cached remote tip for zero benefit. And on flatten/compact the cached tip is semantically dead anyway (history was just rewritten; every peer must re-clone), so pruning is unambiguously correct there. The informational note in `bd gc` output routing operators to flatten/compact is the right compromise.

Other points checked:
- Prune sits between squash and GC in both flatten and compact paths; failure is warn-and-continue (GC still runs, just reclaims less) — the right severity for a local cache cleanup.
- Tags are list-and-warn only, never deleted, enforced by the test.
- Conn discipline is consistent: pinned conn for the mutating prune on both stores (`withMutatingDBConn` on embedded), pool reads for listings.
- Size reporting degrades gracefully (-1 → fields omitted / line suppressed), and measuring after the squash correctly isolates the GC-attributable reclaim.

One cosmetic non-blocker: `remote_refs_pruned`/`tags_anchoring` marshal as JSON `null` when empty (nil slices) rather than `[]` — worth normalizing someday if any consumer scripts against it, not worth holding the PR.

PR #4891, opened 2026-07-18
fix(schema): carry wisps.row_lock on the ignored migration track — fresh bootstraps since 0054 can't insert wisps (bd-rotq2)

steveyegge:
## Problem (bd-rotq2, P1 — live in prod)

`wisps` is dolt-ignored (migration 0019), so its schema is clone-local, materialized by `ignored/0001`. But `row_lock` was added only by **synced** migration 0054. A workspace that bootstraps or re-clones from a remote whose `schema_migrations` cursor is already ≥ 0054 adopts the cursor without ever executing 0054 — leaving a wisps table the post-0054 binary cannot insert into (`Error 1054: Unknown column 'row_lock' in 'wisps'`).

Live occurrence: claude-code-vm's Gargoyle lane soft-failed 54 wisp inserts after its post-squash re-clone (wy-pt82l, reported by lion), silently dropping its wisp trail while the lane stayed green. **Every fresh synced bootstrap since the 0054/0055 cutover is affected**, including new community clones; in-place upgrades are not (0054 executed locally there), which is why it went unnoticed.

## Fix

New guarded `ignored/0013_add_wisp_row_lock` (same INFORMATION_SCHEMA + PREPARE pattern as `ignored/0006`): adds the column iff wisps exists and lacks it. Because the ignored track runs at store open, already-broken workspaces **self-heal on their next connect**; upgraded workspaces where 0054 already added the column no-op. Nothing edits shipped migrations (content hashes untouched).

## Testing

- New real-Dolt regression test `TestDoltNew_WispRowLockSelfHeal_RealDolt`: fresh workspace has the column; then reproduces the bootstrap-broken state exactly (drop column, rewind ignored cursor, synced cursor stays at latest) and proves a reopen restores it, an insert naming `row_lock` succeeds, and further opens are no-ops. Includes a probe subtlety worth knowing: ignored-table schema lives in the working set, so a pooled connection can hold a pre-ALTER read view — probes force a fresh view first.
- `internal/storage/schema` suite, `embeddeddolt` suite, `test/docsync`, `check-migration-hygiene.sh`, and lint all green locally.

## Process note

Future synced migrations must not alter dolt-ignored tables without a paired ignored/ migration; a doctor check comparing local-table shape to binary expectation is a candidate follow-up (tracked on the bead).

### DIFF
--- a/internal/storage/dolt/wisp_row_lock_heal_test.go
+++ b/internal/storage/dolt/wisp_row_lock_heal_test.go
@@ -1,0 +1,113 @@
+package dolt
+
+import (
+	"context"
+	"fmt"
+	"testing"
+)
+
+// TestDoltNew_WispRowLockSelfHeal_RealDolt is the regression test for
+// bd-rotq2: wisps is dolt-ignored, so its schema is clone-local, and a
+// workspace that bootstraps from a remote whose synced cursor is already
+// >= 0054 never executes the synced migration that added wisps.row_lock —
+// leaving a wisps table the post-0054 binary cannot insert into (Error 1054,
+// observed in prod on claude-code-vm). The fix carries the column on the
+// ignored track (ignored/0013), which runs at store open, so an affected
+// workspace self-heals on its next connect. The test reproduces the broken
+// state directly (drop the column, rewind the ignored cursor past 0013 —
+// the synced cursor stays at latest, exactly like a bootstrap) and asserts
+// a reopen restores the column.
+func TestDoltNew_WispRowLockSelfHeal_RealDolt(t *testing.T) {
+	skipIfNoDolt(t)
+
+	ctx, cancel := testContext(t)
+	defer cancel()
+
+	tmpDir := t.TempDir()
+	dbName := uniqueTestDBName(t)
+
+	cfg := &Config{
+		Path:            tmpDir,
+		CommitterName:   "test",
+		CommitterEmail:  "test@example.com",
+		Database:        dbName,
+		CreateIfMissing: true,
+		MaxOpenConns:    1,
+	}
+
+	store, err := New(ctx, cfg)
+	if err != nil {
+		t.Fatalf("New (create): %v", err)
+	}
+	defer func() {
+		dropCtx, dropCancel := context.WithTimeout(context.Background(), testTimeout)
+		defer dropCancel()
+		_, _ = store.db.ExecContext(dropCtx, fmt.Sprintf("DROP DATABASE IF EXISTS `%s`", dbName))
+		store.Close()
+	}()
+
+	hasRowLock := func(stage string) bool {
+		t.Helper()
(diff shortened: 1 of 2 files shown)


steveyegge:
**Review verdict: PASS** (lion — reviewer on the originating prod incident, wy-pt82l/wy-4jjz.6; posting as comment, GH blocks same-identity approve).

Verified, not trusted:

- **Root-cause match**: the migration's mechanism statement matches what I observed on claude-code-vm — post-0054 binary (1.1.0 dev aceab79806bd), pre-0054 clone-local wisps table, 54× Error 1054 on wisp INSERT after the read path recovered. Carrying the column on the ignored track is the right fix surface: it heals exactly the workspaces synced 0054 can never reach.
- **Definition parity**: 0013's `BIGINT NOT NULL DEFAULT 0` is byte-identical to synced 0054's wisps clause — no schema fork between healed and in-place-upgraded workspaces on this column.
- **Fresh-init double-ALTER**: safe — 0054's wisps clauses are guarded on column absence, so ignored-0013-then-synced-0054 ordering on a brand-new workspace no-ops cleanly; the test's fresh-workspace phase exercises this for real.
- **Idempotence vs the wy-pt82l interim ALTER**: the INFORMATION_SCHEMA guard makes 0013 a no-op on crow's hand-fixed VM, either ordering.
- **Test quality**: reproducing the bootstrap state directly (drop column + rewind ignored cursor, synced cursor untouched) is exactly the broken invariant, and asserting a *usable* insert naming row_lock — the shape the prod binary emits — is the assertion that matters. Idempotent third open covered.

**One non-blocking observation**: synced 0054 also added `lease_expires_at`/`heartbeat_at` to wisps, which bootstrap-broken clones will now permanently lack. I traced all live query shapes: post-cutover code never reads or writes those columns on wisps (lease state lives in the ignored/0012 `leases` table; wisp IDs never get lease rows; only row_lock is written on wisp mutating paths), so the divergence is dead-weight-only today. Worth a one-line comment in 0013 saying they're deliberately not healed — otherwise the next person auditing 0054-vs-clone schema diffs will think the heal is incomplete. If wisps ever grow a reader of those columns, that's a new 001x.

CI is green. Merge on green per protocol — and once a binary with this ships to claude-code-vm, the Gargoyle's wisp trail (dark since 07-17 14:30Z, live-flipped lane) lights back up; I'll verify from the wyvern side.

steveyegge:
**Re-verdict: PASS** (lion). Verified the delta (9e3cb619e3) is comment-only — sole file is ignored/0013, guard SQL unchanged — and verified the new claim against 0055 itself: guarded `DROP COLUMN lease_expires_at/heartbeat_at` blocks confirm upgraded workspaces shed those columns too, so bootstrapped clones aren't merely harmlessly divergent, they're already in the correct final shape and the heal is *complete*. Better than my original note assumed. Merge on green.

PR #4892, opened 2026-07-18
feat(errors): typed dependency error taxonomy on the public beads package

julianknutsen:
## What

Adds a typed error taxonomy for dependency-write failures on the public `beads` package. Every production site that rejects a **self-dependency** or a **dependency cycle** now returns a shared sentinel (`ErrSelfDependency` / `ErrDependencyCycle`), and the root package re-exports the dependency + not-found + claim sentinels (`beads.ErrNotFound`, `ErrAlreadyClaimed`, `ErrNotClaimable`, `ErrSelfDependency`, `ErrDependencyCycle`) as aliases. Callers can switch from string-matching error messages to `errors.Is` — and every message stays **byte-identical**, so existing string matchers keep working during the migration.

## Why

The `bd` CLI classifies these dependency-write failures with `strings.Contains`, which is brittle and drifts across the three plumbings that can produce them (`issueops`, the Dolt cross-tier check, and the `domain`/`db` use-case path). Typing the errors at the source lets the CLI — and the library's own call sites — classify by identity (`errors.Is`) instead of message text. This is the first slice of refactoring the `bd` CLI into a thin adapter over the library, with the library owning error semantics; the message text is preserved so nothing that still string-matches breaks mid-migration.

A pre-merge review caught that converting only the same-tier `issueops` cycle check was **not enough**: `doltTransaction.checkCrossTierSchedulingCycle` runs a *separate* cycle check on cross-tier/wisp edges (the `bd create --deps blocks:<wisp>,depends-on:<regular>` shape) and sets `SkipCycleCheck`, bypassing the `issueops` path — so `errors.Is(err, beads.ErrDependencyCycle)` was true for same-tier cycles but false for cross-tier ones on the *same* `DoltStore.AddDependency`. All plumbings are now routed through the sentinels so the taxonomy is uniform.

Scope is deliberately narrow (one concern): only the dependency self-dep/cycle taxonomy plus the re-export. Sentinels for verbs that have no return site yet (e.g. close-blocked) are intentionally excluded to avoid dead placeholder values.

## Verification

- `gofmt -l` clean on all touched files; `go build ./...` and `go vet` clean.
- Wrap-preserving contract tests assert **exact** message byte-identity **and** `errors.Is` on real production returns:
  - `internal/storage/issueops` — `CheckDependencyCycleInTx` self-dep + cycle.
  - root `beads` package — sentinel identity + a real-production-return test through the `domain` use case.
  - `internal/storage/dolt` — `errors.Is(domain.ErrDependencyCycle)` regression assertions on the cross-tier and mixed-tier cycle paths (`TestRunInTransactionCrossTierCycleRejected`, `TestRunInTransactionMixedTierCycleSameTierClosingEdgeRejected`).
  - `internal/storage/domain/db` — `errors.Is(domain.ErrSelfDependency)` on the SQL repository self-dep path.
- The pre-existing self-dependency test passes **unchanged**, which is the proof the canonical messages stayed byte-identical.
- `golangci-lint run` (pre-commit) reports **0 new issues**.
- CI (`make ci-pr-core` / `ci-pr-policy` / `ci-pr-lint`) will run the full suite; the Dolt-backed integration tests above were run locally against a live Dolt container and pass.

Reviewer sanity check:
```bash
go test ./internal/storage/issueops/ . ./internal/storage/domain/ \
  -run 'TestCheckDependencyCycleInTx|TestReExportedSentinel|TestPersistDependenciesSkipsValidationErrorsWhenConfigured' -count=1
go test ./internal/storage/dolt/ \
  -run 'TestRunInTransactionCrossTierCycleRejected|TestRunInTransactionMixedTierCycleSameTierClosingEdgeRejected' -count=1
```

### DIFF
--- a/beads.go
+++ b/beads.go
@@ -14,6 +14,7 @@ import (
 	"github.com/steveyegge/beads/internal/configfile"
 	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/storage/dolt"
+	"github.com/steveyegge/beads/internal/storage/domain"
 	"github.com/steveyegge/beads/internal/types"
 )
 
@@ -195,4 +196,15 @@ const (
 	EventLabelAdded        = types.EventLabelAdded
 	EventLabelRemoved      = types.EventLabelRemoved
 	EventCompacted         = types.EventCompacted
+)
+
+// Re-exported error sentinels so consumers match on errors.Is rather than
+// on message text. Each is an ALIAS of the internal sentinel, so the identity
+// is preserved across the package boundary.
+var (
+	ErrNotFound        = storage.ErrNotFound
+	ErrAlreadyClaimed  = storage.ErrAlreadyClaimed
+	ErrNotClaimable    = storage.ErrNotClaimable
+	ErrSelfDependency  = domain.ErrSelfDependency
+	ErrDependencyCycle = domain.ErrDependencyCycle
 )
--- a/cmd/bd/dep.go
+++ b/cmd/bd/dep.go
@@ -14,6 +14,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/beads/internal/metrics"
 	"github.com/steveyegge/beads/internal/storage"
+	"github.com/steveyegge/beads/internal/storage/domain"
 	"github.com/steveyegge/beads/internal/types"
 	"github.com/steveyegge/beads/internal/ui"
 )
@@ -527,7 +528,11 @@ func addBulkDependenciesInTx(ctx context.Context, tx storage.Transaction, edges 
 		return fmt.Errorf("final cycle check failed (no edges added): %w", cycleErr)
 	}
 	if cyclePath != "" {
-		return fmt.Errorf("dependency cycle would be created: %s (no edges added; run 'bd dep cycles' for analysis)", cyclePath)
+		// Type this rejection so callers can errors.Is(err, domain.ErrDependencyCycle),
+		// matching the proxied/domain bulk final gate. NewCycleError renders the
+		// message verbatim (no sentinel text appended), so the user-facing string
+		// is unchanged.
+		return domain.NewCycleError("dependency cycle would be created: %s (no edges added; run 'bd dep cycles' for analysis)", cyclePath)
 	}
 	return nil
 }
--- a/cmd/bd/dep_test.go
+++ b/cmd/bd/dep_test.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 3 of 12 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds a typed error taxonomy for dependency-write failures on the public `beads` package: self-dependency and dependency-cycle rejections now return shared sentinels (`ErrSelfDependency` / `ErrDependencyCycle`), re-exported from the root package as aliases, so callers can classify with `errors.Is` instead of brittle `strings.Contains` — while every error message stays byte-identical. That replaces fragile string-matching in the `bd` CLI with identity-based classification across all three plumbings that can produce these failures (`issueops`, the Dolt cross-tier check, and the `domain`/`db` use-case path), and lands the first slice of turning `bd` into a thin adapter over the library with the library owning error semantics.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Bulk scheduling self-edges were misreported as cycles — bulk validation now rejects `IssueID == DependsOnID` before the hierarchy, per-edge, and final cycle checks, returning a self-dependency sentinel (`internal/storage/domain/dependency.go`).
- Cross-tier scheduling cycles returned an untyped string — the cross-tier cycle path now returns `domain.ErrDependencyCycle` (`internal/storage/dolt/transaction.go`), with cross-tier regression coverage.
- The public sentinels did not compose across the internal→public boundary — `beads.ErrSelfDependency` and `beads.ErrDependencyCycle` are now identity-preserving aliases of the internal sentinels (`beads.go`).
- Embedded bulk final-gate cycle rejections were untyped while the proxied/domain paths were typed — the domain bulk final gate and the embedded CLI path now share one cycle wrapper that keeps the rendered message (`internal/storage/domain/dependency.go`, `cmd/bd/dep.go`).

Review findings addressed:
- Behavioral correctness (major, high, Claude + Codex): the bulk dependency final cycle guard now satisfies `errors.Is(err, ErrDependencyCycle)` while preserving its message text.
- Contract & interface fidelity (major, high, Codex + Claude): proxied CLI error text stays byte-identical via a string-preserving wrapper whose `Unwrap()` returns the sentinel.
- Behavioral correctness (major, high, Codex): bulk self-edges are now reported as self-dependencies rather than as cycles.

Remaining non-gating notes:
- One minor, non-gating maintainability note: the defensive SQL-repository self-edge message at `internal/storage/domain/db/dependency.go` now carries the sentinel-prefixed wording for direct repository callers. Normal use-case and CLI entry points still reject self-edges earlier with the preserved user-facing text, so this only matters to an out-of-repo caller that exact-matches that internal defensive string.

### Maintainer Changes
`fix(errors): type bulk dependency self-dep and cycle rejections without changing their text` — 5 files changed, +124/-6, pushed directly to `feat/guarded-ops-s1-error-taxonomy`. It applies the self-dependency guard, routes the embedded bulk final cycle gate through the shared typed wrapper, and adds the exact-message + `errors.Is` assertions the review asked for.

### Final Review Status
Ready for the merge queue: final head `8725e99d` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/29667437190

### Review Iterations
4 review passes performed: 3 fix iterations that resolved the bulk self-dependency/cycle typing and byte-identical message-preservation findings, then a final pass with no remaining blocker or major findings.

---

PR #4893, opened 2026-07-18
feat(close): CloseIssueChecked — atomic guarded close in the engine

julianknutsen:
> Stacked on #4892 (S1 typed error taxonomy). Base is `feat/guarded-ops-s1-error-taxonomy`; retarget to `main` once #4892 merges. The diff below is S2 only.

## What

Adds `CloseIssueChecked` — a close that runs the `is_blocked` guard **and** the close inside **one transaction**, so the guard is atomic. A `Force` option bypasses it (mirrors `bd close --force`). New `storage.ErrCloseBlocked` sentinel (re-exported as `beads.ErrCloseBlocked`), plus `CloseIssueOptions{Reason,Session,Force}` and `CloseIssueResult{Unchanged}`.

## Why

`bd close` today reads blockers (`store.IsBlocked`) and then calls `store.CloseIssue` in a **separate** transaction — a TOCTOU: a blocker can clear, or the bead can be closed by someone else, between the check and the close. Moving the guard into a single engine transaction removes that window and lets the `bd` CLI delegate its close guard to the engine instead of hand-rolling the check-then-close dance. This is a step in making the library the single enforcement layer with the CLI as a thin adapter.

Scope is one concern. Category-aware reopen (reopening custom done-category statuses) and folding the open-children close guard (GH#3681) into the same op are deliberately split into follow-up PRs.

## Verification

- Guard semantics: `IsBlockedInTx` reads the denormalized transitive `is_blocked` column — an open blocking dependency or an open blocking gate. On a blocked refusal, `CloseIssueChecked` returns `ErrCloseBlocked` and the transaction **rolls back**: no close, no `closed` event.
- Implemented on the `Storage` interface across `DoltStore` (perm + wisp paths, mirroring `CloseIssue`'s DOLT_COMMIT / wisp handling), `EmbeddedDoltStore`, `HookFiringStore` (fires `on_close` on success, like `CloseIssue`), and `InstrumentedStorage`; both cgo and `CGO_ENABLED=0` builds compile.
- Integration tests assert the **atomic-refuse** property (blocked → refused, still open, **zero** closed events recorded) on the perm, wisp, and embeddeddolt paths, plus: Force closes a blocked bead; a non-blocking dep type does not trip the guard; already-closed → `Unchanged=true`; missing id → `ErrNotFound`.
- `gofmt`/`go vet` clean; `golangci-lint` (pre-commit) 0 issues.

```bash
go test ./internal/storage/dolt/ -run TestCloseIssueChecked -count=1 -v
BEADS_TEST_EMBEDDED_DOLT=1 go test ./internal/storage/embeddeddolt/ -run TestEmbeddedCloseIssueChecked -v
```

### DIFF
--- a/beads.go
+++ b/beads.go
@@ -48,6 +48,16 @@ type Transaction = beads.Transaction
 // one whole-graph check, never graph integrity.
 type DependencyAddOptions = storage.DependencyAddOptions
 
+// CloseIssueOptions carries the optional inputs to Storage.CloseIssueChecked —
+// an atomic, guarded close that refuses a still-blocked issue with
+// ErrCloseBlocked unless Force is set. Exported so consumers can name it without
+// importing internal/storage.
+type CloseIssueOptions = storage.CloseIssueOptions
+
+// CloseIssueResult reports the outcome of Storage.CloseIssueChecked. Unchanged
+// is true when the issue was already closed (idempotent no-op).
+type CloseIssueResult = storage.CloseIssueResult
+
 // RemoteStore provides dolt remote management and replication operations.
 // Use type assertion on a Storage value to access these methods:
 //
@@ -205,6 +215,7 @@ var (
 	ErrNotFound        = storage.ErrNotFound
 	ErrAlreadyClaimed  = storage.ErrAlreadyClaimed
 	ErrNotClaimable    = storage.ErrNotClaimable
+	ErrCloseBlocked    = storage.ErrCloseBlocked
 	ErrSelfDependency  = domain.ErrSelfDependency
 	ErrDependencyCycle = domain.ErrDependencyCycle
 )
--- a/errors_test.go
+++ b/errors_test.go
@@ -3,6 +3,7 @@ package beads_test
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/steveyegge/beads"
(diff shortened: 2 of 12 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds `CloseIssueChecked` — an atomic guarded-close primitive in the storage engine. The blocked-guard and the close now run inside a single transaction, and a blocked refusal returns *before* any close or versioning side effects, so a "don't close while blocked" rule can be enforced atomically at the engine layer instead of racing a check-then-close in application code. The Dolt and embedded-Dolt wrappers surface the idempotent no-op cleanly by mapping `issueops.CloseResult.AlreadyClosed` onto the public `CloseIssueResult.Unchanged`. This is a solid, well-tested foundation for guarded closes.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The guarded close evaluated the denormalized `is_blocked` flag before checking whether the row was already closed. An already-closed issue carrying a stale `is_blocked=1` value would return `ErrCloseBlocked` instead of the idempotent `Unchanged=true` no-op that the close contract promises. The final patch short-circuits the already-closed row first and delegates to `CloseIssueInTx`, so re-closing an already-closed issue — even one with a stale blocked flag — is once again an idempotent success, and a Dolt regression fixture pins the behavior.

Review findings addressed:
- Contract correctness (major, reviewer): a non-force close of an already-closed row with a stale `is_blocked` flag returned a blocked error rather than an idempotent no-op; the final patch detects the closed row before applying the blocked guard, restoring the documented idempotency.
- Test evidence quality (minor, high confidence): the permanent-issue stale-`is_blocked` regression is now pinned; wisp and embedded-wrapper coverage still leans on the shared `issueops` path — recorded as a non-gating follow-up rather than a merge blocker.

Remaining non-gating notes:
- A wisp-specific stale-`is_blocked=1` already-closed regression test would extend coverage beyond the shared `issueops.CloseIssueCheckedInTx` case (minor).
- The transitive / no-direct-blockers `ErrCloseBlocked` message branch has no direct assertion yet (nit).

None of these block merge.

### Maintainer Changes
`fix(close): short-circuit already-closed before the is_blocked guard` — 2 files changed, 110 insertions(+), 9 deletions(-) (`internal/storage/issueops/close.go` plus a Dolt regression test). The original contributor commit is preserved; this is a single maintainer fixup on top.

### Final Review Status
Ready for the merge queue: final head `3806417c` has passing required GitHub checks (PR, PR Risk, Regression Tests, Conformance, and Cross-Version Smoke Tests all green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/29701651853

### Review Iterations
2 review passes performed: attempt 1 requested the idempotent-re-close fix, and attempt 2 approved after the fix landed and was validated.

---

PR #4894, opened 2026-07-18
feat(validate): typed length bounds for assignee, owner, and label

julianknutsen:
> Stacked on #4893 (S2) → #4892 (S1). Base is `feat/guarded-ops-s2-checked-close`; retarget down the stack as each lands. The diff below is S3 only.

## What

Adds typed length validation for the `assignee`, `owner`, and `label` fields (all `VARCHAR(255)` columns). Over-length values now return a typed `beads.ErrFieldTooLong` (`errors.Is`-able) instead of a raw backend "data too long" error (assignee/owner) or a silent `INSERT IGNORE` truncation (labels). Length is counted in **runes** to match the column's character semantics.

## Why

`bd create -a <long>`, `bd assign`, `bd label add <long>`, and imports could feed the engine an over-255 field. Today `bd` surfaces either a raw SQL error the CLI can't classify, or — for labels — silently stores a truncated value the user never typed (a full-array label PATCH then never converges). Validating up front with a typed error lets `bd` (and the library's own call sites) reject cleanly and consistently. Part of making the library the single enforcement layer with the CLI as a thin adapter.

A pre-merge review caught that the engine has **two** write stacks — the embedded `issueops`/`DoltStore` path and the proxied-server `domain/db` (uow) path — and a first pass only guarded the first, so the silent-truncation bug persisted in proxied-server mode. All raw assignee/owner/label write sites on **both** stacks (create, update, label insert, and the claim CAS that writes `assignee = actor`) are now guarded through the shared `types.CheckFieldLen`.

Scope is one concern (field bounds). The closed-boundary-as-category status guard is split into a follow-up.

## Verification

- `types.CheckFieldLen` uses `utf8.RuneCountInString` vs `MaxFieldLen = 255`; a 255-rune multibyte value (`é`×255, ~510 bytes) round-trips **unchanged** through the real store, a 256-rune one is rejected — proving rune-count, not byte-count, end to end.
- Guarded on both stacks; tests reject over-length create/update/label/claim with `ErrFieldTooLong` and assert **nothing is persisted** (single and bulk create), on the `dolt` and the `domain/db` (proxied) harnesses.
- `gofmt`/`go vet` clean; `go build ./...` under cgo and `CGO_ENABLED=0`; `golangci-lint` (pre-commit) 0 issues.

```bash
go test ./internal/storage/dolt/ -run TestFieldLengthValidation -v
go test ./internal/storage/domain/db/ -run TestDomainDB/TestFieldLengthGuards -v
go test ./internal/types/ . -count=1
```

### DIFF
--- a/beads.go
+++ b/beads.go
@@ -208,6 +208,10 @@ const (
 	EventCompacted         = types.EventCompacted
 )
 
+// MaxFieldLen re-exports the maximum length (in characters) of the assignee,
+// owner, and label fields, paired with the ErrFieldTooLong sentinel below.
+const MaxFieldLen = types.MaxFieldLen
+
 // Re-exported error sentinels so consumers match on errors.Is rather than
 // on message text. Each is an ALIAS of the internal sentinel, so the identity
 // is preserved across the package boundary.
@@ -218,4 +222,5 @@ var (
 	ErrCloseBlocked    = storage.ErrCloseBlocked
 	ErrSelfDependency  = domain.ErrSelfDependency
 	ErrDependencyCycle = domain.ErrDependencyCycle
+	ErrFieldTooLong    = types.ErrFieldTooLong
 )
--- a/errors_test.go
+++ b/errors_test.go
@@ -44,6 +44,7 @@ func TestReExportedSentinelIdentity(t *testing.T) {
 		{"ErrNotClaimable", beads.ErrNotClaimable, storage.ErrNotClaimable},
 		{"ErrSelfDependency", beads.ErrSelfDependency, domain.ErrSelfDependency},
 		{"ErrDependencyCycle", beads.ErrDependencyCycle, domain.ErrDependencyCycle},
+		{"ErrFieldTooLong", beads.ErrFieldTooLong, types.ErrFieldTooLong},
 	}
 	for _, tc := range cases {
 		if tc.exported != tc.internal {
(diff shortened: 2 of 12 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds typed 255-character length bounds for the `assignee`, `owner`, and `label` fields — a rune-counted typed error contract (`ErrFieldTooLong` via `types.CheckFieldLen`) wired through the embedded and proxied create, update, claim, and label write paths — which replaces silent label truncation and raw backend `VARCHAR(255)` errors with one clear, consistent validation error.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Over-length labels could be silently truncated by `INSERT IGNORE` writes while over-length `assignee`/`owner` values surfaced raw backend errors. The typed 255-char guards now reject these consistently at every write chokepoint, and the review independently validated the premise against the `INSERT IGNORE` label writes and the `VARCHAR(255)` assignee/owner columns.
- The guards were traced through both storage stacks — embedded and proxied create, update, claim, and label writes — and the rune-counted `types.CheckFieldLen` / `ErrFieldTooLong` contract preserves `errors.Is` behavior through the public aliases.

Review findings addressed:
- Architectural consistency (minor, high confidence / Claude): the new `ValidateForImport` comment and test can imply live import-path coverage even though the production import path does not call that function. This is a clarity note, not a behavior gap — the enforced create/update/claim/label write paths are unaffected.

Remaining non-gating notes:
- Consider clarifying that `ValidateForImport` field-length parity is intentional but currently unused in production, or dropping the guard/test from that unused path (`internal/types/types.go`, `internal/types/types_test.go`). No gating issues were found.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `b0ec80dce` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/29703967095

### Review Iterations
1 review pass performed — approved on the first pass with no fix iterations required.

---

PR #4896, opened 2026-07-18
fix(query): make slash and mixed-case metadata keys queryable

### LINKED ISSUE #4895: Metadata keys with '/' or mixed case are accepted on write but silently unqueryable
## Summary

Two related gaps let `bd` accept a metadata key at write time that can never be matched by a `metadata.<key>=<value>` query — the query returns zero rows with no error.

### 1. Slash keys rejected at query time

`storage.ValidateMetadataKey` did not allow `/`, and the query lexer did not treat `/` as an identifier character, so path-style keys like `jira/sprint` could not be stored or queried even though they are natural for namespaced tooling.

### 2. Mixed-case keys silently lowercased

Keys like `Sprint` or `reviewedBy` pass validation and are stored case-sensitively in JSON metadata, but the query parser lowercased the entire field token. `metadata.Sprint=Q1` was rewritten to `MetadataFields["sprint"]`; both `JSON_EXTRACT(metadata, '$."sprint"')` and the in-memory predicate are case-sensitive, so the query silently matched nothing. Inconsistent with `has_metadata_key=Sprint` and the `--metadata` flag, which both preserve case.

## Repro

```
bd update bd-1 --set-metadata Sprint=Q1
bd list --query 'metadata.Sprint=Q1'       # zero rows, no error
bd list --query 'has_metadata_key=Sprint'  # matches — inconsistent
```

## Root causes

- The lexer's identifier alphabet (`isIdentChar`) and the storage key regex (`validMetadataKeyRe`) are maintained independently, so a key can be valid to store but unlexable in a query.
- `parser.go` lowercased the whole field token including the case-sensitive metadata key suffix.
- `JSONMetadataPath` only quoted dotted keys and spliced keys unescaped, relying on callers validating first.

anisoptera:
Closes #4895

## Problem

Metadata keys could be accepted at write time yet be silently unqueryable via `metadata.<key>=<value>` (zero rows, no error):

- Path-style keys (`jira/sprint`) were rejected by both the key validator and the query lexer.
- Mixed-case keys (`Sprint`) were stored case-sensitively but the parser lowercased the whole field token, so the query targeted the wrong JSON key — inconsistent with `has_metadata_key` and `--metadata`, which preserve case.

## Changes

- **Allow `/` in metadata keys**: added to `validMetadataKeyRe` and to the lexer's identifier alphabet (same precedent as `:` for namespaced labels).
- **Preserve metadata key case**: field names remain case-insensitive, but the key suffix of a `metadata.<key>` field keeps its original case.
- **Always quote + escape JSON paths**: `JSONMetadataPath` now always emits `$."<key>"` (no charset list to keep in sync) and escapes `\` and `"` locally, so it is correct for any input rather than correct-only-if-pre-validated.
- **Invariant test**: `TestMetadataKeysAreQueryable` sweeps every ASCII char and enforces that any key accepted by `ValidateMetadataKey` parses into the filter under its original case — guarding the independently-maintained lexer/validator charsets against future drift.

## Testing

`go test -tags gms_pure_go` passes for `internal/query`, `internal/storage`, `internal/storage/sqlbuild`, and `cmd/bd` metadata tests.

End-to-end, against a real `bd` binary (built with `gms_pure_go`) and a fresh Dolt database:

```
bd update test-5qz --set-metadata Sprint=Q1
bd update test-5qz --set-metadata jira/sprint=42

bd query 'metadata.Sprint=Q1'                 # ✓ matches (returned 0 rows before this fix)
bd query 'metadata.jira/sprint=42'            # ✓ matches (slash key, SQL JSON_EXTRACT path)
bd query 'metadata.Sprint=Q2'                 # ✓ no match (values still filter)
bd query 'metadata.sprint=Q1'                 # ✓ no match (JSON keys stay case-sensitive)
bd query 'has_metadata_key=Sprint'            # ✓ matches — now consistent with metadata.Sprint
bd query 'metadata.Sprint=Q1 OR status=closed'  # ✓ matches (in-memory predicate path)
bd list --metadata-field jira/sprint=42       # ✓ matches (flag path)
```


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the drift at the PR head). Verdict: **needs changes** — the slash/mixed-case queryability itself is sound and wanted; landing needs a careful rebase plus two accuracy fixes.

**Landing blocker — the metadata-helper conflict must keep both sides.** Verified: current main added `MergeMetadataJSON` / `ApplyMetadataEdits` / `MetadataEditValue` (required by the landed transaction-safe metadata update path) exactly where this PR declares `jsonPathEscaper`; merge-tree conflicts there. Either side taken wholesale breaks compilation or regresses the concurrent-update fix — the resolution keeps the escaper *and* all three upstream helpers.

**Should-fix — `JSONMetadataPath` overclaims "correct for any input."** Every production SQL caller first runs `ValidateMetadataKey`, which rejects quotes and backslashes — so the escaping cases are unreachable through real callers. And through the pinned stack they don't actually work end-to-end: go-mysql-server's fallback delegates to the pinned jsonpath tokenizer, which doesn't recognize escaped quotes or doubled backslashes (its own tests leave those disabled), and Dolt's optimized parser unescapes only `\"`. Either narrow the helper's contract to validated keys (and drop the unreachable cases), or add extraction-level tests proving the escaping through the real engines.

**Should-fix — the newly supported query surface is undocumented.** `bd query --help` mentions neither `metadata.<key>` nor `has_metadata_key`; add the syntax + a slash-key example and regenerate.

Verified clean: lexer/parser/evaluator/SQL-builder wiring all present; valid slash and mixed-case keys flow correctly to bound `JSON_EXTRACT`; no operator ambiguity from `/`.

maphew (approved):
Approving — supersedes my 2026-07-24 changes-requested review. Rather than another round, I merged main into the branch as a maintainer commit: the conflict with #4732 resolves by keeping both your always-quote+escape `JSONMetadataPath`/`jsonPathEscaper` and main's metadata-edit functions. The review's proof demand is met with two new query-path subtests showing slash and mixed-case keys round-trip through the real Dolt/go-mysql-server JSON path parser; quote/backslash escaping is documented as defense-in-depth (all production callers gate through `ValidateMetadataKey`, which rejects both — verified, including the `doltTransaction.SearchIssues` call site).

Validation: build/vet clean, metadata suites green; cross-vendor Codex gate signed off after a doc-accuracy fix. Merge tail to the pr-babysit patrol. Thanks anisoptera — two solid landings this week.

PR #4897, opened 2026-07-18
feat(ready): add pagination metadata to BD_JSON_ENVELOPE=1 output when truncated

jacobhausler:
## Problem

`bd ready --json` returns a bare JSON array capped at 100 items with no indication that the result is truncated. The text output path already emits:

```
Showing 100 of 149 ready issues. Use --limit 0 for all, or --limit N to raise the cap.
```

But the JSON output path — which is what agents, `gc`, and automated scripts actually consume — has no equivalent signal. A programmatic consumer reading stdout cannot distinguish \"149 ready, here are 100\" from \"exactly 100 ready.\"

## Solution

When `BD_JSON_ENVELOPE=1` is set and the result set is capped, include a `\"pagination\"` key in the envelope:

```json
{
  \"schema_version\": 1,
  \"data\": [...],
  \"pagination\": {
    \"returned\": 100,
    \"total\": 149,
    \"truncated\": true
  }
}
```

When results are not truncated, the `pagination` key is absent entirely (no noise for the common case). The existing stderr text hint is unchanged. The bare-array path (default, without `BD_JSON_ENVELOPE`) is also unchanged.

## Compatibility

- **Non-breaking**: no change to the default bare-array output.
- Consumers that want machine-readable truncation info opt in via `BD_JSON_ENVELOPE=1`, which is already the documented forward-compatible JSON format.
- The `pagination` key follows the same envelope pattern as `schema_version` and `data`.

## Verification

Built from source on a throwaway 105-issue store with `--limit` at default (100):

**Before** (`BD_JSON_ENVELOPE=1`):
```json
{\"schema_version\": 1, \"data\": [...100 items...]}
# no pagination key — no way to detect truncation
```

**After** (`BD_JSON_ENVELOPE=1`):
```json
{
  \"schema_version\": 1,
  \"data\": [...100 items...],
  \"pagination\": {\"returned\": 100, \"total\": 105, \"truncated\": true}
}
```

`--limit 0` returns all 105 items with `pagination: null` (no truncation).

Text banner unchanged: `Showing 100 of 105 ready issues. Use -n to show more.`

Test suite passes (pre-existing failures in `TestIsBackupAutoEnabled` / `TestConfigGetBackupEnabled_EffectiveValue` — git-remote detection — are unrelated to this change and fail on unmodified HEAD too).

## Scope

Signal only. Default limit value and sort policy are out of scope per the upstream issue discussion.

## Alternatives considered

- **Breaking envelope** (`--json` always returns `{\"items\":[...],\"total\":N}`): cleaner but breaks every existing caller. Left for a major-version bump.
- **Stderr JSON object**: agent consumers typically don't capture stderr; prose-on-stderr already exists and is not sufficient for automated use.
- **`--limit 0` default for `--json`**: changes semantics and performance profile, separate discussion.


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claim at the PR head). Verdict: **needs changes** — good contract addition, but the mode that most needs it is skipped.

**Blocker — proxied-server JSON still truncates silently.** Verified at the PR head: `cmd/bd/ready.go` routes proxied invocations away from the modified path, and `ready_proxied_server.go` calls `outputJSON(results)` with no pagination metadata even though it holds definitive truncation evidence in `page.HasMore` (used only afterward). Non-CGO builds are steered to proxied mode, so this isn't an obscure backend — it's the exact silent-truncation ambiguity the PR exists to remove, on one of the two front doors. (`PaginationMeta.Total` may need an API extension there since the domain page carries no total; `returned` + `has_more` alone would already be honest.)

**Should-fix — the command-level wiring is untested.** The only new test verifies serialization of caller-supplied metadata; nothing runs `ready` against a genuinely truncated result — which is precisely how the proxied gap slipped through.

**Should-fix — the public JSON contract doc is now stale.** `docs/reference/json-schema.md` documents the envelope as `schema_version` + untouched `data` and says `ready` matches `list`; the new top-level `pagination` key needs documenting (additive-optional, so no version bump per the schema doc's own rule — but it must be written down). Also the commit's `Closes #4892` reference is stale — main already merged an unrelated #4892.

With the proxied dual emitting the same metadata and a truncation test on both routes, this is a clean, agent-friendly improvement.

maphew (approved):
Approving — supersedes my 2026-07-23 changes-requested review. The blocker is fixed as a maintainer commit: the proxied ready route now emits the same pagination metadata as the direct route under `BD_JSON_ENVELOPE=1`, with exact parity (key present only when truncated; `total` omitted via `omitempty` since the proxied backend has no cheap count — the Codex gate caught that a bare int would emit a false `"total": 0`). Proxied subtests cover truncated (meta asserted, no total key) and non-truncated (key absent). Your envelope/output.go design ships as-authored — nice clean seam.

Validation: build/vet/gofmt clean, envelope unit tests + proxied subtests green by name; cross-vendor gate signed off. Merge tail to the pr-babysit patrol. Thanks jacobhausler.

PR #4898, opened 2026-07-19
feat(metadata): MergeMetadata — atomic single-key metadata merge

julianknutsen:
> Stacked on #4894 (S3) → #4893 (S2) → #4892 (S1). Base is `feat/guarded-ops-s3-field-validation`; retarget down the stack as each lands. The diff below is S4 only.

## What

Adds `MergeMetadata` — an **atomic** single-key merge into an issue's metadata JSON — and reimplements `SlotSet`/`SlotClear` on top of it. The value is stored as raw JSON, so nested objects/arrays are preserved (not stringified).

## Why

`SlotSet` (and `SlotClear`) did a read-modify-write across **two** transactions: `GetIssue`, then `UpdateIssue`. Two concurrent calls — even on **different** keys — each read the same base metadata and wrote back the whole blob, so the second silently clobbered the first's key. Doing the read and the write in **one** transaction closes that window: on a Dolt optimistic-commit conflict the write is retried and re-reads the latest metadata, so a concurrent merge of a different key survives. This is the engine primitive the `bd` CLI (and the library's own metadata call sites) should use instead of hand-rolling a fragile read-then-write.

A pre-merge review caught that reimplementing `SlotSet` on a **bare** metadata `UPDATE` would silently drop what the old `UpdateIssue` path did — the `EventUpdated` history event (with actor attribution) and the configured metadata-schema validation — and would leave `SlotClear` on the old clobber. The fix routes the merge/delete **through `UpdateIssueInTx`** in the one transaction, so every side effect is preserved (behavior-identical to the old `SlotSet`, now clobber-free), and `SlotClear` is rebuilt on the same atomic path.

## Verification

- **B1 no-clobber**: an 8-goroutine test merges distinct keys concurrently on one issue and asserts every key (plus a pre-seeded one) survives — the property the old cross-tx `SlotSet` violated; a matching test interleaves concurrent `SlotSet`/`SlotClear`.
- **Behavior preserved**: tests assert the `EventUpdated` history event (actor-attributed) is recorded and the metadata-schema validation still rejects a violating value; `SlotSet` stays byte-compatible; nested JSON round-trips as an object.
- Cross-backend **conformance** covers `MergeMetadata` + atomic `SlotClear` on both server and embedded Dolt.
- `gofmt`/`go vet` clean; `go build ./...` under cgo and `CGO_ENABLED=0`; `golangci-lint` (pre-commit) 0 issues.

```bash
go test ./internal/storage/dolt/ -run 'TestMergeMetadata|TestSlotSet|TestSlotClear' -v
go test ./internal/storage/conformance/ -run 'TestConformance/.*/Slots' -v
```

### DIFF
--- a/internal/jira/tracker_test.go
+++ b/internal/jira/tracker_test.go
@@ -735,6 +735,9 @@ func (s *configStore) MergeSlotAcquire(_ context.Context, _, _ string, _ bool) (
 }
 func (s *configStore) MergeSlotRelease(_ context.Context, _, _ string) error { return nil }
 func (s *configStore) SlotSet(_ context.Context, _, _, _, _ string) error    { return nil }
+func (s *configStore) MergeMetadata(_ context.Context, _, _ string, _ json.RawMessage, _ string) error {
+	return nil
+}
 func (s *configStore) SlotGet(_ context.Context, _, _ string) (string, error) {
 	return "", nil
 }
--- a/internal/storage/conformance/portable.go
+++ b/internal/storage/conformance/portable.go
@@ -1,6 +1,7 @@
 package conformance
 
 import (
+	"encoding/json"
 	"errors"
 	"slices"
 	"sort"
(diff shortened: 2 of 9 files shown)


julianknutsen:

PR #4899, opened 2026-07-19
feat(concurrency): expose a per-row version on Issue reads (RowVersion)

julianknutsen:
> Stacked on #4898 (S4) → #4894 (S3) → #4893 (S2) → #4892 (S1). Base is `feat/guarded-ops-s4-merge-metadata`; retarget down the stack as each lands. The diff below is S5 only.

## What

Surfaces the engine's internal `row_lock` column read-only as `types.Issue.RowVersion` — an opaque, equality-only optimistic-concurrency token. It is a **Go-only** field (`json:"-"`); it never appears on any `bd --json` or `bd export` surface.

## Why

`row_lock` is a random non-zero int64 the engine rewrites on every status/ownership-mutating write (claim, close, unclaim) and on the generic update path, but it was not surfaced on `types.Issue`. A caller that wants a fine-grained change token had only `updated_at`, which is stored at **second granularity** — so two writes in the same second are indistinguishable. Exposing `RowVersion` lets a caller tell same-second writes apart (the concrete gap: a change token built from `updated_at` + status + labels can't detect a second same-second edit to the same field; `RowVersion` can). This is the read-side foundation the `bd` CLI and the library's own call sites need; an optional in-transaction CAS on the guarded ops is a follow-up.

A pre-merge review caught that serializing the token onto the JSON wire (`json:"row_version,omitempty"`) broke the pinned `TestCorpusGolden` and the export round-trip test — and couldn't be regenerated stably because the value is random per write. The fix makes it Go-only (`json:"-"`): callers read the field directly, and no golden or export output changes.

## Verification

- Single-point change: `row_lock` added to the canonical `sqlbuild.IssueBaseColumns` and a matching dest in `issueops.ScanIssueFrom` (verified 49-column/49-dest positional alignment). Both write stacks, wisps, and the counts/dependents parallel scans derive from those, so hydration propagates once; broad issue-read tests (GetIssue/Search/Create/counts/dependents) confirm no read is corrupted.
- `RowVersion` hydrates on read and the list path, changes on a mutating write, and distinguishes two same-second writes (identical `updated_at`, distinct `RowVersion`) — the B4 property.
- `RowVersion` never appears on any JSON surface (`TestRowVersionNeverSerialized`); `TestCorpusGolden` and the export round-trip are green with **no** golden/testdata changes.
- Content hash is unaffected (`RowVersion` is not folded into `ComputeContentHash`). `gofmt`/`go vet` clean; `go build ./...` under cgo and `CGO_ENABLED=0`; `golangci-lint` 0 issues.

```bash
go test ./internal/storage/dolt/ -run TestRowVersion -v
go test ./internal/types/ -run RowVersion
go test ./cmd/bd/protocol/ -run TestCorpusGolden
```

### DIFF
--- a/internal/storage/dolt/row_version_test.go
+++ b/internal/storage/dolt/row_version_test.go
@@ -1,0 +1,135 @@
+package dolt
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/steveyegge/beads/internal/storage"
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// TestRowVersion exercises the read-only RowVersion token — the issues/wisps
+// row_lock cell surfaced on types.Issue. RowVersion is stable across reads with
+// no intervening write, changes on every status/ownership-mutating write, is
+// hydrated by the list path as well as the point read, and — the reason the
+// token exists — distinguishes two writes that land in the same wall-clock
+// second, where updated_at (DATETIME, second granularity) is identical.
+func TestRowVersion(t *testing.T) {
+	store, cleanup := setupTestStore(t)
+	defer cleanup()
+	ctx, cancel := testContext(t)
+	defer cancel()
+
+	get := func(t *testing.T, id string) *types.Issue {
+		t.Helper()
+		iss, err := store.GetIssue(ctx, id)
+		if err != nil {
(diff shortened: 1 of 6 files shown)


julianknutsen:

PR #4900, opened 2026-07-19
feat(deps): return a typed error on dependency type-conflict

julianknutsen:
> Stacked on #4899 (S5) → #4898 (S4) → #4894 (S3) → #4893 (S2) → #4892 (S1). Base is `feat/guarded-ops-s5-row-version`; retarget down the stack as each lands. The diff below is S6 only.

## What

Makes the dependency type-conflict error typed on **both** engine write paths and re-exports the two dependency-conflict types on the public `beads` package, so a caller classifies an `AddDependency` rejection with `errors.As` instead of matching the message text.

## Why

Adding a dependency between a pair that already has an edge of a *different* type is a deterministic rejection. One write path returned a typed `*DependencyTypeConflictError`; the other returned a plain `fmt.Errorf` string — so `errors.As` worked on one path and not the other, and `bd` had to fall back to string-matching (or treat the deterministic rejection as a retryable fault). Typing it consistently lets the `bd` CLI (and the library's own call sites) map it to a clean, deterministic outcome. Completes the dependency-write error taxonomy: self-dependency and cycle are already typed sentinels, and the hierarchy/cross-type error was already typed on both paths — type-conflict was the last formatted-string gap.

The message is **byte-identical** (`DependencyTypeConflictError.Error()` is the exact former string), so no string-matcher changes.

## Verification

- `errors.As(err, &beads.DependencyTypeConflictError{})` matches at the public `AddDependency` boundary on both write paths (verified against a real Dolt server and the alternate stack), with the four fields readable; the re-exports are `=` type aliases so the public type is the engine type.
- Covered on the permanent `issues→issues` route **and** the wisp-source `BeginTx`/`Commit` seam; `err.Error()` asserted byte-identical to the historical string; the deterministic conflict is returned immediately (not retried as a transient serialization error).
- Existing dependency string-match tests still pass (message unchanged). `gofmt`/`go vet` clean; `go build ./...` under cgo and `CGO_ENABLED=0`; `golangci-lint` 0 issues.

```bash
go test ./internal/storage/dolt/ -run TestAddDependencyTypeConflict -v
go test ./internal/storage/domain/db/ -run TestDependencySQLRepository -count=1
go test . -run TestReExportedDependencyConflictTypes
```

### DIFF
--- a/beads.go
+++ b/beads.go
@@ -224,3 +224,13 @@ var (
 	ErrDependencyCycle = domain.ErrDependencyCycle
 	ErrFieldTooLong    = types.ErrFieldTooLong
 )
+
+// DependencyTypeConflictError is returned by AddDependency when an edge of a
+// different type already exists between the pair; callers errors.As it to read
+// the existing/requested types instead of parsing the message.
+type DependencyTypeConflictError = domain.DependencyTypeConflictError
+
+// DependencyHierarchyConflictError is returned by AddDependency when a blocking
+// edge would gate an issue on its own ancestor/descendant (a gate that can
+// never clear).
+type DependencyHierarchyConflictError = domain.DependencyHierarchyConflictError
--- a/errors_test.go
+++ b/errors_test.go
@@ -94,6 +94,70 @@ func TestReExportedSentinelCatchesRealProductionError(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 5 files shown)


julianknutsen:

PR #4902, opened 2026-07-19
feat(deps): emit dependency_added / dependency_removed events

julianknutsen:
> Stacked on #4900 (S6) → #4899 (S5) → … → #4892 (S1). Base is `feat/guarded-ops-s6-dep-errors`; retarget down the stack as each lands. The diff below is S6b only.

## What

Restores `dependency_added` / `dependency_removed` event emission on `bd dep add` / `bd dep remove`, so the events feed surfaces dependency changes. Emitted on both Dolt write paths, committed atomically with the edge, gated to the explicit dep verb.

## Why

The event types have existed since the original SQLite backend, which emitted them on every add/remove — but the emission was dropped when that backend was removed for the Dolt-only migration and never ported. The constants (and their re-exports) survived with no writer, so `bd dep add`/`remove` recorded nothing and the history/activity view silently omitted dependency changes. This restores the original behavior on the current `bd` write paths (an events row on the source issue with the original descriptive text).

Two correctness details the restore has to get right (both caught pre-merge):
- **Committed with the edge.** The store-level add/remove and the transaction-path (`RunInTransaction` / batch / graph-apply) methods now stage the source's event table (`events` or `wisp_events`) before commit — otherwise the event row dangles in the working set and is lost on the next reset/reconcile.
- **Explicit verb only, consistently.** Only the explicit `bd dep add`/`bd dep remove` verb emits. Create-with-deps (implicit parent-child, `--deps`, waits-for) creates the edge without an event on both backends, so `bd create --parent X` yields identical history regardless of storage backend.

## Verification

- Add/remove emit exactly one event and **commit** it (asserted via `events AS OF 'HEAD'` + clean `dolt_status`); idempotent re-add and no-op remove of a missing edge emit nothing; wisp sources route to `wisp_events`, never `events`.
- Consistency proof: explicit `dep add`/`remove` emit; `bd create --parent`/`--deps` create the edge but emit no `dependency_added` — on both write plumbings.

```bash
go test ./internal/storage/dolt/ -run 'TestDependencyEvent|TestRunInTransaction.*Dependency' -v
go test ./internal/storage/domain/db/ -run 'TestDependencyEventConsistency|TestDependencySQLRepository' -count=1
```

### DIFF
--- a/internal/storage/dolt/dependencies.go
+++ b/internal/storage/dolt/dependencies.go
@@ -52,7 +52,10 @@ func (s *DoltStore) AddDependency(ctx context.Context, dep *types.Dependency, ac
 		return err
 	}
 	// GH#2455: Use explicit DOLT_ADD to avoid sweeping up stale config changes.
-	return s.doltAddAndCommit(ctx, []string{"dependencies"}, "dependency: add "+string(dep.Type)+" "+dep.IssueID+" -> "+dep.DependsOnID)
+	// Stage events alongside dependencies: AddDependencyInTx now records a
+	// dependency_added event, and DOLT_COMMIT only commits explicitly staged
+	// tables — omitting events would leave the event row in the working set.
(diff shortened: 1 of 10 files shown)


julianknutsen:

PR #4903, opened 2026-07-19
feat(close): optional ExpectedVersion CAS on CloseIssueChecked

julianknutsen:
> Stacked on #4902 (S6b) → #4900 (S6) → … → #4892 (S1). Base is `feat/guarded-ops-s6b-dep-events`; retarget down the stack as each lands. The diff below is S5b only.

## What

Adds an optional compare-and-swap to `CloseIssueChecked`: the caller may pass the row's expected version (the `RowVersion` / `row_lock` token), and the close proceeds only if the row still has that version — otherwise it refuses with a typed `beads.ErrVersionMismatch`, atomically. When no version is supplied, behavior is unchanged.

## Why

`CloseIssueChecked` already guards a close on `is_blocked` in one transaction. But a caller that reads an issue and then closes it can't express "only if it hasn't changed since I read it" — a lost-update window. The optional `ExpectedVersion` closes that: the version read and the close share one transaction, so it's a true compare-and-swap with no read-then-write gap. This is the engine primitive the `bd` CLI (and the library's own call sites) need for optimistic concurrency on close.

`Force` bypasses only the `is_blocked` guard, never the version check — the CAS is an orthogonal precondition. The check has two limbs: the read-side comparison catches a writer that committed before the close began, and on the retry-wrapped path a commit-time `row_lock` conflict is replayed so the re-read refuses.

## Verification

- Match closes; stale version refuses atomically — `errors.Is(err, ErrVersionMismatch)`, issue still open, zero `closed` events (the tx rolled back); a committed concurrent write invalidates a captured version; nil `ExpectedVersion` is unchanged behavior; `Force: true` with a stale version still refuses; missing id → `ErrNotFound`.
- Wisp sources route to the `wisps` table; an already-closed re-close with the post-close version stays idempotent (`Unchanged`), not a spurious mismatch. Covered on both DoltStore and EmbeddedDoltStore.

```bash
go test ./internal/storage/dolt/ -run TestCloseIssueCheckedVersionCAS -v
go test . -run TestReExportVersionMismatch
```

### DIFF
--- a/beads.go
+++ b/beads.go
@@ -220,6 +220,7 @@ var (
 	ErrAlreadyClaimed  = storage.ErrAlreadyClaimed
 	ErrNotClaimable    = storage.ErrNotClaimable
 	ErrCloseBlocked    = storage.ErrCloseBlocked
+	ErrVersionMismatch = storage.ErrVersionMismatch
 	ErrSelfDependency  = domain.ErrSelfDependency
 	ErrDependencyCycle = domain.ErrDependencyCycle
 	ErrFieldTooLong    = types.ErrFieldTooLong
--- a/errors_test.go
+++ b/errors_test.go
@@ -28,6 +28,22 @@ func TestReExportCloseBlocked(t *testing.T) {
 	}
 }
 
+// TestReExportVersionMismatch proves the public beads.ErrVersionMismatch alias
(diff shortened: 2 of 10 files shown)


julianknutsen:

PR #4904, opened 2026-07-19
refactor(cli): bd close delegates to the engine's guarded close

julianknutsen:
> Stacked on #4903 (S5b) → #4902 (S6b) → … → #4892 (S1). Base is `feat/guarded-ops-s5b-version-cas`; retarget down the stack as each lands. The diff below is S7 only.

## What

`bd close` (embedded path) now delegates to the engine's `CloseIssueChecked` instead of running its own is_blocked pre-check and closing in a separate call. The verb builds `CloseIssueOptions{Reason, Session, Force}` and calls the op; the duplicated CLI guard is deleted.

## Why

The CLI's check-then-close was a read-then-write TOCTOU: an issue could become blocked between the `IsBlocked` read and the `CloseIssue` write. `CloseIssueChecked` runs the guard and the close in one transaction, so the refusal is atomic (still open, no `closed` event). This is the delegation pattern for the `bd` verbs: argument parsing in the CLI, enforcement in the library.

**Behavior-preserving, deliberately.** A pre-merge review compared old-vs-new binaries and caught two deltas, both fixed before merge:
- The engine guard originally refused on the bare denormalized `is_blocked` column, which would have refused transitively-blocked children (blocked parent, zero direct blockers) and stale `is_blocked` rows that the old CLI closed. The guard now uses the exact historical predicate — refuse only on a **live direct blocker** — so the delegation changes no close semantics. (Whether transitive blocks *should* refuse a close is a real question, but it belongs to a deliberate semantics change, not a refactor.)
- An already-closed issue still appears in the `--json` array and text report (idempotent success, exit 0) — but no longer emits a spurious `closed→closed` audit entry or the real-close side effects.

## Verification

- Direct blocker: refuses atomically (issue stays open, blockers named, `--force` hint), closes with `--force`. Transitively-blocked and stale-`is_blocked` beads close without `--force` (old-binary parity). Already-closed: in `--json` output alone and in mixed batches, exit 0, no spurious audit.
- Covered at the engine level (both stores, incl. new transitive/stale cases) and through the built `bd` binary.

```bash
go test ./internal/storage/dolt/ -run TestCloseIssueChecked -v
BEADS_TEST_EMBEDDED_DOLT=1 go test ./cmd/bd/ -run Close -count=1
```

### DIFF
--- a/cmd/bd/close.go
+++ b/cmd/bd/close.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -106,6 +107,7 @@ the flags appear in the command line.`,
 		// Direct mode
 		closedIssues := []*types.Issue{}
 		closedCount := 0
+		alreadyClosed := 0
 		firstClosedID := ""
 
 		for i, id := range resolvedIDs {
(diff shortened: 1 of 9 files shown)


julianknutsen:

PR #4845, opened 2026-07-16
Enable verified public v0.62 migration apply

julianknutsen:
## What

Enable the public v0.62 migration bridge to perform an explicit, plan-bound server-to-embedded upgrade. `--inspect` remains the default and returns an immutable plan digest; effectful work requires `--apply --yes --expect-plan <digest>` plus explicit checksum-pinned v0.62.0 `bd` and Dolt 1.84.0 executables.

The bridge copies and requalifies the historical source, exports and enriches the exact supported core corpus, initializes a side-by-side embedded-Dolt target, imports and independently reopens it, compares canonical issue and relationship data, retains the original `.beads` tree as `.beads-v0.62.0-rollback`, and atomically publishes the verified target. Hidden init-only plumbing preserves the authenticated project, repository, clone, database, and issue-prefix identities without exposing a new user-facing `bd` command.

This remains opt-in and manual. Ordinary current commands continue to refuse the historical workspace; there is no automatic upgrade.

Parent: #4840

## Why

The official v0.62 lane previously proved only an internal test recipe, while the public bridge could inspect but truthfully returned `apply_not_available`. Users on v0.62 need a deterministic path that either completes with verified fidelity and a retained rollback or fails closed before publishing a target.

The authentic lane also caught a fake/real discrepancy during this slice: current `bd export` represents parenthood through the `parent-child` dependency and omits the redundant legacy top-level `parent` projection. The focused fake now reproduces that shape, and the bridge derives the projection only when exactly one matching dependency exists. Missing, conflicting, malformed, or multiple parent edges remain rejected.

## Qualification boundary

The first public apply path is deliberately bounded to the checksum-pinned official v0.62.0 source and its qualified five-issue/eight-feature corpus:

- epic, child task, blocking bug, standalone issue, and closed issue
- exact `parent-child` and `blocks` edges, including supported dependency audit fields
- rich description, notes, design, acceptance criteria, external reference, label, and comment fields
- authenticated issue prefix, Dolt database, and project identity
- correct `blocked`, `ready`, `close`, reopen, and post-migration create behavior

Sources outside that proven semantic envelope are refused rather than migrated speculatively.

## Safety, rollback, and provider boundary

- Binds inspect and apply to source-tree/content hashes, source/target executable hashes, source identity, semantic baseline, and the exact plan digest.
- Reinspects under an authenticated workspace operation fence before cutover.
[...]

PR #4810, opened 2026-07-15
Refuse legacy Dolt workspaces before migration mutation

julianknutsen:
## What

Refuse legacy v0.50-v0.58 Dolt-server workspaces before version tracking, compatibility-marker writes, server startup, or store open. The error asks the user to preserve a byte-for-byte `.beads` backup and use a qualified, version-specific migration bridge.

This is a focused prerequisite for the checksum-pinned v0.57 historical-upgrade E2E lane. It does not migrate data, change schema, or alter current embedded/proxied/Postgres/MySQL/SQLite provider behavior.

## Why

The current candidate's ordinary `bd list` against an official v0.57.0 workspace timed out after rewriting `.local_version`, creating `.bd-dolt-ok`, starting/touching server state, and changing legacy Noms manifests. A migration bridge cannot be considered safe if an ordinary current command mutates the source before refusing it.

Tracked by `bd-ldt0f.3.26`.

Stacked on PR #4805, which adds the preceding v0.55.4 historical lane.

## Safety and rollback

- Classification uses the durable server-era storage shape and missing post-v0.58 `project_id`; mutable version/compatibility markers cannot bypass it.
- The version witness is bounded, no-follow, and cosmetic only.
- Ambiguous/symlinked/special Dolt roots fail closed.
- Current server workspaces with project identity and all non-server providers bypass the guard.
- The change performs no migration. Reverting this commit restores the prior behavior; no data rollback is required.
- This migration-adjacent change requires substantive human review and must not be self-merged.

## Verification

- `make build`
- Shipped-config focused CGO suite covering v0.55.4/v0.57 refusal, current provider shapes, mutable marker bypasses, ambiguous roots, store-free commands, and canonical SQLite metadata: PASS
- Real v0.57 subprocess refusal: about 0.14 seconds, source tree byte-identical, no `.bd-dolt-ok`
- `git diff --check`
- Exact-diff council at SHA-256 `86187f3d6bf4aaf815652ca910f1f9ee5d0fb722756047e778ebd370019c1d47`: 0 Critical, 0 Important

_codex-gpt-5-unknown-reasoning on behalf of Test User_

PR #4801, opened 2026-07-15
test(migration): gate real v0.49.6 SQLite upgrade and rollback safety

julianknutsen:
## What

Adds a required black-box PR gate that creates a real workspace with the checksum-pinned official v0.49.6 Linux amd64 release, probes it with the current multi-provider candidate, and runs the explicit `sqlite_to_current` bridge.

The strict lane fails on an unavailable or tampered release, skip, outcome/recipe drift, incomplete fixtures or snapshots, source mutation, unsafe rollback preservation, data-fidelity loss, or broken blocked/ready/close behavior.

## Why

Historical upgrades have failed for users without a deterministic release gate. This is the first thin end-to-end spine: a real released SQLite-era binary, a real legacy database, the current candidate, an explicit bridge, and fail-closed assertions in normal PR CI.

The expected result remains `MANUAL(sqlite_to_current)`; this PR does not make upgrades automatic.

Bead: `bd-ldt0f.3.22`

## Migration safety

- The direct candidate probe must leave the complete classic `.beads` tree unchanged.
- The bridge preserves byte-identical rollback copies of the SQLite DB, WAL/SHM, metadata, `config.json`, and `config.yaml` before invoking historical commands.
- Existing stale or conflicting rollback destinations fail before any copy or source removal.
- Rollback artifacts are verified both immediately after the bridge and after post-upgrade mutation checks.
- Exact IDs, rich fields, closed state, dependency, label, comment author/text, item cardinality, and empty collections are compared.
- The strict PR job runs for PRs into any target branch, including the multi-provider integration branch; only documentation/website-only changes are ignored.

## Verification

- `make build`
- `./scripts/migration-test/strict-mode-test.sh`
- `CANDIDATE_BIN="$PWD/bd" GIT_CONFIG_NOSYSTEM=1 ./scripts/migration-test/run.sh --strict --expect MANUAL v0.49.6`
  - 5/5 issues preserved
  - 8 required source features present
  - 0 fidelity/blocker violations
  - result: `MANUAL(sqlite_to_current)`
- `CANDIDATE_BIN="$PWD/bd" GIT_CONFIG_NOSYSTEM=1 ./scripts/migration-test/run.sh --self-test`
- `actionlint .github/workflows/migration-test.yml`
- `shellcheck -x -S error scripts/migration-test/run.sh scripts/migration-test/lib/*.sh scripts/migration-test/recipes/sqlite_to_current.sh scripts/migration-test/strict-mode-test.sh`
- `make ci-pr-policy`
[...]

PR #4905, opened 2026-07-19
Plan deterministic universal upgrades for v1.2

julianknutsen:
## What

Adds the final requirements, architecture, executable Gas City bead DAG, and qualification contract for deterministic upgrades from every prior public Beads release revision to the exact packaged v1.2/HEAD candidate set.

This is the plan-only PR for epic `bd-ldt0f`; it changes no production behavior.

## Why

Users on old releases have encountered nondeterministic or failed upgrades. The plan defines a zero-skip historical E2E denominator, authentic old-binary workspace creation, one storage-neutral migration planner, driver-owned safety and recovery boundaries, separate smoke and focused result namespaces, and an immutable one-build qualification epoch on top of the multi-provider branch.

## Final scope

- Four plan artifacts, `+4659/-0`, against `feature/backend-provider-change-20260713` at `f7d0c26ec8c1e7b6b075cc49b07cb2f0f41c3a47`.
- 173 historical tag names and 179 frozen revision identities: 178 currently visible cross-remote identities plus the superseded first-public `v1.1.0` revision.
- One owned root convoy, 11 executable children, and 22 static blocking edges.
- Initial ready set exactly `{u0-integrate-history, u1-lock-denominators}`; U10 is an explicitly read-only final eligibility barrier.
- Provider identity separates platform and engine runtime, and historical smoke identities remain runtime-keyed.
- Migration inspection is atomic for every open; durable fenced/CAS recovery stays behind the driver/adapter boundary.
- Release eligibility, bundle sealing, claim attestation, exact-tag dispatch, publication, and postpublication verification are separate fail-closed stages.

## Contributor intake

- Preserves accepted history, tests, authorship, and attribution from #4801, #4810, and #4845 through U0's explicit ancestry merges on corrected PR #4907.
- Treats external contributor PR #4878 contributor-first before any overlapping U4 schema work.
- Does not close, replace, or supersede contributor work.

## Verification

- Unique frontmatter and fenced-YAML parsing passes.
- The active Gas City materializer dry-run accepts one convoy with 11 mapped children.
- The dependency graph is acyclic; all 22 static edges are present; the ready set is exactly U0/U1.
- Exact mappings are `.14`, `.15`, and `.17`-`.25`; `.16` is explicitly excluded and never inferred.
- The root carries `owned`; U10 has `files: []`.
- Historical inventory and materializer behavior were independently re-audited against live refs, GitHub evidence, and the active parser.
- Final hardening staged diff SHA-256: `6cccf057469ff4b41cb702191e8f255626cca18e6ef1246facd2c5dfa26f3ab0`.
- Approved index tree and committed tree: `5f3e1e00f49c6df593e93ea6c17ebfcb5220d58c`.
- Three fresh independent Sol/Ultra reviewers approved those identical staged bytes with zero Critical or Important findings.
- Post-commit first-parent full-index diff digest and commit tree exactly match the approved digest/tree.
- `git diff --cached --check` passed before commit; no source tests were required because this PR changes plan documents only.

## Review policy

Each implementation slice remains subject to TDD and three independent Sol/Ultra approvals of identical staged bytes before commit. Each new PR receives `status/needs-review-auto` exactly once and verifies its historical label event. Nontrivial work requires substantive accountable-human review of current HEAD; migration, schema, and sync paths never merge with bot-only approval.

### DIFF
--- a/.gc/plans/v12-universal-upgrades/.council/ultra-architecture.md
+++ b/.gc/plans/v12-universal-upgrades/.council/ultra-architecture.md
@@ -1,0 +1,212 @@
+## Architecture verdict
+
+Build one storage-neutral upgrade planner/executor, backed by driver-owned migration operations, plus a black-box historical-release harness. Do not create 173 production code paths: production selects routes from probed storage generation, provider, topology, and schema version—not from the Beads release tag.
+
+### 1. Architecture and module boundaries
+
+Use four boundaries:
+
+1. **`internal/storage` and `dolthub/driver`**
+   - Own read-only probing, engine capabilities, quiescence, snapshots, transactional behavior, cutover, rollback, and verification.
+   - Expose only storage-neutral state and capability results. Do not leak Dolt/PostgreSQL/MySQL internals into CLI or public SDK types.
+   - Missing capabilities return a route-specific unsupported result.
+
+2. **Internal migration core**
+   - Maintains the small route graph and deterministic ordered plan.
+   - Classifies each plan as `NoOp`, `StartupSafe`, `ManualRequired`, or `Unsupported`.
+   - Coordinates driver operations but contains no engine inspection, flock implementation, generic retry loop, or crash-recovery framework.
+   - Uses the same planner for startup and `bd migrate`; there must not be two migration implementations.
+
+3. **Startup and `bd migrate` adapters**
+   - Startup selects only `NoOp` or `StartupSafe`.
+   - Existing `--inspect`, `--dry-run`, and `--yes` are sufficient; add no persisted plan format or approval database.
+
+4. **Test-only release catalog and harness**
+   - Catalogs tags/releases/assets and acquires historical binaries.
+   - Runs old and candidate binaries strictly as black boxes.
+   - Never becomes production code or a release/review controller.
+
+The production route graph should normalize historical workspaces into format families such as legacy SQLite, legacy Dolt, server Dolt, embedded Dolt, and current provider-local formats. Every catalog row must map to a probed route; an unrecognized generation is a hard failure.
+
+Provider-local schema upgrades and provider-changing migrations are separate concerns. The only admitted provider-changing route remains embedded Dolt → PostgreSQL. SQLite → PostgreSQL, MySQL conversions, PostgreSQL → Dolt, and arbitrary provider pairs remain unsupported.
+
+### 2. Startup and manual-migration behavior
+
+Every normal workspace open follows these invariants:
+
+1. Probe is read-only and must not initialize storage, launch an engine, rewrite config, or partially open a legacy database.
+2. Construct and validate the complete plan before the first mutation. Every
+   mutating path, including `StartupSafe`, then acquires driver-owned
+   quiescence, re-probes, recomputes the complete plan, and byte-matches source,
+   capabilities, configuration, topology, and plan before its first write.
+3. Startup may auto-apply only when every step:
+   - stays on the same local provider and topology;
+   - has fixed work independent of workspace cardinality;
+   - performs no bulk scan, copy, backfill, remote coordination, or destructive rewrite;
+   - is declared atomic or fully rollback-safe by the driver;
+   - requires no snapshot/cutover protocol.
+4. Index builds, data-dependent transformations, historical bridges, server/embedded transitions, provider changes, remote schema operations, and destructive steps are always manual.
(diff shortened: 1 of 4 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication spot-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the deterministic-upgrade planning framework is substantial work, but it's built against a storage world main has since rejected, plus four independent correctness blockers.

**Blocker 1 — restores removed backends.** Init provisioning, runtime store-open branches, and the migration selector all speak PostgreSQL/MySQL/SQLite; main deleted those adapters (#4857, #4881) and rejects the names fail-closed. This is a product-direction conflict, not a rebase chore — the plan needs re-scoping around Dolt-only storage or an explicit new decision.

**Blocker 2 — fresh `bd init` breaks on FreeBSD before RunE.** Verified: the init backend-preflight runs unconditionally in root pre-run and reaches `safefile.ObserveMetadataNoFollow`, whose non-linux/darwin/windows stub returns `ErrUnsupported` — and goreleaser still ships `bd-freebsd-amd64`.

**Blocker 3 — the redirect rewrite drops #4875's invalid-target guard.** The strict path verifies directory identity but not `hasBeadsProjectFiles(target)`, which main added after the branch point (with a regression test) precisely to stop false-empty `bd list` on redirects to empty directories. Both sides changed `beads.go`; the guard must be manually preserved.

**Blocker 4 — the topology witness can misclassify shared-server as embedded.** The planner classifies from metadata + filesystem artifacts but ignores `dolt.shared-server` config/env, which the real store-open treats as authoritative over stale `dolt_mode=embedded` metadata — the planner can admit an embedded migration the runtime would never perform.

**Blocker 5 — the new lifecycle has no production caller.** `InspectSourceShape`/`BindProviderConfiguration`/`WithBoundProviderConfiguration` are used only inside the package; startup and `bd migrate` still run the legacy path, contradicting the checked-in one-planner-before-telemetry/version/store-open requirement.

Also: the linux workspace-identity probe opens `embeddeddolt` directly and embeds ext4/xfs policy — engine layout belongs behind the storage driver (the PR's own boundary rule); and the source-identity binding hard-codes `branch: "main"` while bd can operate on other branches.

maphew:
## Chain-level disposition (this PR + #4907)

Following up the two changes-requested reviews posted earlier today on this PR and #4907, here is the maintainer view of the chain as a whole, so the per-PR blockers don't read as a rejection of the program.

**The program is wanted.** Deterministic, qualified upgrades from every old public release is exactly the pain our issue queue keeps surfacing (failed/nondeterministic upgrades, schema-incompatible installs). The authentic-release qualification harness, the fail-closed pre-store guards, and the public v0.62 bridge from #4801/#4810/#4845 are the right shape, and the CI lanes on #4907 are already green against their base.

**The base is the problem, and it's shared by the whole chain.** Everything here — including the merged sub-PRs — sits on `feature/backend-provider-change-20260713`, which is 29 ahead / 308 behind `main` and still carries the PostgreSQL/MySQL/SQLite storage world that `main` removed and now rejects fail-closed (#4857, #4881). Nothing on that base can reach `main` without re-scoping around Dolt-only storage; rebasing alone won't do it, since the plan artifacts, the migration selector, and the topology witness all encode the multi-backend world.

**Two paths, either with same-week maintainer support:**

1. **You re-cut (preferred).** Extract the qualification harness + v0.62 bridge onto Dolt-only `main`: drop the PostgreSQL/MySQL/SQLite provisioning and selection surfaces, keep the runtime-keyed historical smoke identities, requalify the strict lanes against main's current driver/Dolt pins, and re-scope the plan docs in this PR accordingly. Your Gas City program structure and authorship stay fully intact; we commit to prompt review of the re-cut.
2. **Maintainer-side extraction.** If you'd rather focus elsewhere, we extract the bridge + harness onto `main` with your commits preserved as literal ancestry (or Co-authored-by where a re-cut is unavoidable), and the per-PR blockers fixed in the process. Nothing gets closed or superseded without your work credited.

Could you say which path you prefer by **2026-07-31**? If we don't hear back we'll ping once, then take the question to the repo owner rather than act unilaterally.

The per-PR blockers from today's reviews stand independently of the path choice (here: FreeBSD init preflight, the dropped #4875 redirect guard, topology-witness vs `dolt.shared-server` config, no production caller for the new lifecycle; on #4907: the direct engine introspection that the storage charter reserves to the driver).

PR #4906, opened 2026-07-19
feat(update): UpdateIssueChecked — optional ExpectedVersion CAS on the update

julianknutsen:
> Stacked on #4904 (S7) → #4903 (S5b) → … → #4892 (S1). Base is `feat/guarded-ops-s7-cli-delegation`; retarget down the stack as each lands. The diff below is S5c only.

## What

Adds `UpdateIssueChecked` — the generic field-map update with an optional compare-and-swap on the row's version (`RowVersion` / `row_lock`), mirroring the `CloseIssueChecked` CAS. When `ExpectedVersion` is supplied, the version read and the update share one transaction and a moved row refuses with the typed `beads.ErrVersionMismatch`. Nil disables the check; `UpdateIssue` itself is untouched.

## Why

A caller that reads an issue and then updates it couldn't express "only if it hasn't changed since I read it" — the read and the write were separate calls, a lost-update window. This closes it at the engine, the same way the checked close did, so `bd` (and the library's own call sites) get optimistic concurrency on the update verb without hand-rolling a read-then-write.

One route needed real care: an update carrying `no_history`/`wisp` routes through the demote-to-wisp flow (a multi-table row move). A pure extraction (`demoteToWispInTx`, verified byte-identical to the old body under `git diff -w`) lets the version check compose atomically with the demote in the same transaction — and tests now lock that seam.

## Verification

- CAS match updates and bumps `RowVersion`; stale refuses atomically (`errors.Is(ErrVersionMismatch)`, field unchanged, zero `updated` events); a committed concurrent write invalidates a captured version; nil is byte-identical to `UpdateIssue` (line-by-line parity: same validation, routing, commit); missing id → `ErrNotFound`.
- All routes covered on both stores: permanent, wisp (match + stale), and the demote route (match migrates + applies; stale leaves the row in `issues`, unchanged). The hook decorator fires `on_update` exactly once on success and never on a refused update.
- `gofmt`/`go vet` clean; `go build ./...` under cgo and `CGO_ENABLED=0`; `golangci-lint` 0 issues; `UpdateIssue`/`DemoteToWisp` regression suites green.

```bash
go test ./internal/storage/dolt/ -run TestUpdateIssueCheckedVersionCAS -v
go test ./internal/storage/ -run TestHookFiringStoreUpdateIssueChecked -v
```

### DIFF
--- a/beads.go
+++ b/beads.go
@@ -58,6 +58,12 @@ type CloseIssueOptions = storage.CloseIssueOptions
 // is true when the issue was already closed (idempotent no-op).
 type CloseIssueResult = storage.CloseIssueResult
 
+// UpdateIssueOptions carries the optional inputs to Storage.UpdateIssueChecked —
+// an update with an optional ExpectedVersion compare-and-swap that refuses a
+// concurrently-modified issue with ErrVersionMismatch. Exported so consumers can
+// name it without importing internal/storage.
+type UpdateIssueOptions = storage.UpdateIssueOptions
+
 // RemoteStore provides dolt remote management and replication operations.
 // Use type assertion on a Storage value to access these methods:
 //
--- a/errors_test.go
+++ b/errors_test.go
@@ -44,6 +44,24 @@ func TestReExportVersionMismatch(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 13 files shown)


julianknutsen:

PR #4909, opened 2026-07-19
fix(import): report accurate create/update/skip counts for --dry-run

### LINKED ISSUE #4901: bd import --dry-run reports every row as "created" — existence check skipped, output misleads in both directions
`bd import --dry-run` is the tool an operator reaches for to answer "what would this import actually do" — before running an upsert with side effects, or while diagnosing suspected divergence between a JSONL snapshot and the database. As shipped, its output answers neither: every input row is reported as `created`, whether or not the id already exists.

Minimal repro (the file is the database's own export, moments old):

```
$ bd export -o issues.jsonl        # 561 issues
$ bd import --dry-run --json -i issues.jsonl
{"created": 561, "dry_run": true, "skipped": 0, ...}
```

A byte-identical snapshot of current state should dry-run as ~561 updates/no-ops and 0 creates. Reporting 561 creates means the dry-run path returns before the id-existence/upsert check runs.

How we hit it: a data-loss audit ("does this old backup JSONL contain issues the server lost?"). Dry-run said a 507-row backup would create 507 issues — a catastrophic-loss signal. An independent id-set diff showed all 507 ids present in the database; the true answer was 0. The output misleads in both directions: it can fabricate a loss scare, or mask that a real import would touch hundreds of existing rows.

Expected: dry-run performs the same id lookup the real import performs and partitions rows into would-create / would-update / would-skip(stale) — the real import evidently has this logic, since its documented semantics are upsert with stale-row skipping.

Observed on bd via Homebrew (recent 1.x, macOS, Dolt server mode; happy to add exact version output if useful). Additive fix only — no existing behavior needs to change except the dry-run report.

mohamedramadan14:
## What

Fixes `bd import --dry-run` so it reports accurate created / updated / unchanged / stale-skipped counts, instead of reporting every parsed row as a create.

## Why

Fixes #4901. The dry-run branch in `runImportFromReader` returned before ever checking whether an incoming row's id already existed locally, so it always reported `len(issues)` as "created" — a byte-identical re-import of the current database showed hundreds of creates instead of zero. This makes dry-run misleading in both directions: it can fabricate a data-loss scare, or hide that a real import would touch hundreds of existing rows.

The fix reuses the same id-lookup the real (non-dry-run) import path already performs (`filterStaleImportIssues`), so dry-run now classifies rows the same way a real import would.

## Verification

- Added unit tests in `cmd/bd/import_shared_unit_test.go` covering: the exact repro from the issue (byte-identical snapshot → 0 creates), a mix of new/updated/stale/unchanged rows, `--allow-stale`, and the "first import into an empty db" edge case.
- `go build -tags gms_pure_go ./cmd/bd/...` and `go test -tags gms_pure_go ./cmd/bd/...` pass locally.


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the dry-run counting is wrong for two supported input classes and diverges from the execute path's JSON contract.

**Blocker 1 — title-only imports dry-run as "unchanged" but execute as created.** The classifier excludes rows with an empty ID or zero `updated_at` from `NewIDs` and then counts them as `Unchanged` — but `bd import` documents title-only rows as valid (the importer generates IDs and defaults timestamps via `PrepareIssueForInsert`). So `{"title":"x"}` reports `0 new, 0 updated, 1 unchanged` in dry-run and creates an issue on execute. The new tests only use fully-populated rows.

**Blocker 2 — `--allow-stale` reports every row as "new."** Verified at the PR head: the `allowStale` branch returns `Created: len(issues)` with no existence lookup, and the added unit test codifies `Created=1` for an ID that already exists. `--allow-stale` disables the stale guard; it doesn't turn UPSERTs into INSERTs — the real path returns `isNew=false` for existing IDs.

**Should-fix — dry-run and execute JSON disagree on field meaning.** Dry-run's `created`/`ids` carry only truly-new rows; execute's copy them from `ImportResult`, which the core fills with every accepted row including updates. Same fields, different semantics depending on `dry_run` — consumers can't rely on either. Also: equal-timestamp ties are added to `updated` in dry-run even though the store keeps local columns (they already have their own `tie_kept_local_ids`), and the textual "Would import N issues" uses the pre-stale-filter count so the categories don't sum.

Drift note: merges cleanly over the landed chunked-write change (dry-run returns before that path); a rebase is still advisable since `import_shared.go` moved substantially.

The direction — accurate dry-run counts — is exactly right; it just needs the classifier to mirror the real path's `isNew` semantics (including generated IDs and the allow-stale branch) rather than approximating them.

maphew (approved):
Approving — this supersedes my 2026-07-23 changes-requested review. Both blockers and the cleanups are applied as maintainer commits on this branch:

- Title-only / zero-`UpdatedAt` rows now classify through an existence lookup (Created vs Updated), never Unchanged — in the main loop, and (caught by a cross-vendor review gate) also in the all-title-only short-circuit branch that returned before any classification.
- `--allow-stale` classifies existing IDs as Updated via a pure existence classifier instead of blanket Created.
- Cleanups: `TieKeptLocal` out of Updated (reported separately, counted Unchanged), the human-text total now equals `Created+Updated+Unchanged`, `NewIDs` deduped with a separate raw `NewCount` keeping sums exact.
- Tests: untimestamped-row classification (both paths), the rewritten allow-stale test, title-only, tie-kept, dedupe, and an all-title-only-batch regression.

Validation: build/vet/gofmt clean, full Import test set green. Merge tail goes to the pr-babysit patrol. Thanks mohamedramadan14 — the dry-run/execution mismatch you flagged was real and broader than it first looked.

Issue #3502, opened 2026-04-26
bd history <id> --json returns prose 'No history found' instead of valid JSON when bead has no history


seanmartinsmith:
## bug

`bd history <id> --json` returns prose, not JSON, when the bead has no history. breaks any consumer that pipes the output to `jq` or otherwise expects structured output.

## repro (verified on origin/main @ cc31c6046)

```
bd history <id-with-no-history> --json
# stdout: "No history found for issue <id>"
# exit: 0
```

expected: `[]` on stdout, exit 0
actual: prose string, exit 0 — `jq` and friends fail to parse

## root cause

`cmd/bd/history.go` lines 35-48: the empty-history early-return fires *before* the `if jsonOutput` branch. so when `--json` is set AND history is empty, the consumer gets prose. every other input shape (non-empty history) honors the flag correctly — only the empty case slips through.

```go
if len(history) == 0 {
    fmt.Printf("No history found for issue %s\n", issueID)  // L36 — fires first
    return
}

if historyLimit > 0 && historyLimit < len(history) {
    history = history[:historyLimit]
}

if jsonOutput {                                              // L45 — never reached when empty
    outputJSON(history)
    return
}
```

## proposed fix

reorder so the json branch runs above the empty-check. small move, no behavior change for non-empty history, no behavior change for the human-readable path. result: `bd history <id> --json` returns `[]` for empty, full array otherwise — consistent with `bd list --json` / `bd show --json` behavior.

happy to send a PR if this is the shape you want, or if you'd prefer a structured envelope (`{"id":"<id>","history":[]}`) lmk and i'll match that instead.

## why bother

came up while planning a cross-project tool that wants to compose `bd history --json` for a bead-event timeline. without this, every consumer has to defensively detect-and-discard the prose case, which is ugly and error-prone for what should be a clean json contract.

seanmartinsmith:
PR up: #3503

PR #4910, opened 2026-07-19
fix(history): resolve partial/short issue IDs in bd history

### LINKED ISSUE #4868: `bd history <id>` does not resolve partial/short IDs — silently reports "No history found"
## Summary

`bd history <short-id>` prints **"No history found"** (exit 0) for *every* issue, even one with a
long history, because `bd history` never resolves a partial ID to a full one. Every sibling id-command
(`show`, `update`, `close`, `reopen`, `state`, `ready`, …) resolves; `history` is the outlier.

```
$ bd history jt0j                      # short id
No history found for issue jt0j
$ bd history myproj-jt0j               # same issue, full id
📜 History for myproj-jt0j (59 entries)
```

## This is a false negative, not empty history

`jt0j` above is a **null-free** issue whose full id renders 59 entries — so this is not the NULL-scan
crash (that's a separate bug). The short id simply matches nothing and is reported as absence.

## Root cause (released code, v1.1.0 @ `8e4e59d`)

`cmd/bd/history.go:37-39` passes the raw arg straight to the store:

```go
issueID := args[0]
history, err := store.History(ctx, issueID)   // no utils.ResolvePartialID, no routing
```

`store.History` runs an exact match — `HistoryInTx` (`internal/storage/issueops/history.go:30`) does
`WHERE h.id = ?`. A short id matches zero rows, `len(history)==0`, and the command renders
"No history found" (`cmd/bd/history.go:48`). Confirmed there is no resolution anywhere else on the
path: no `PersistentPreRunE` rewrite of args, and the `HookFiringStore` wrapper embeds the store and
passes `History` through unchanged (`internal/storage/hook_decorator.go:27`).

Contrast the ~20 commands that call `utils.ResolvePartialID` (e.g. `cmd/bd/state.go:52`,
`cmd/bd/ready.go:624`). `history` is simply missing the call.

## Why this is a known-shaped gap

The maintainers have already noted (#4714, #4647) that prefix routing / ID resolution "was added
command-by-command" and specific commands were skipped — #3608 fixed `bd close`, #4720 is fixing
`bd mol bond`, etc. This is the same family: `bd history` was left one level lower still — it lacks
even *local* `ResolvePartialID`, not just cross-store routing. #3502 (closed) already touched this exact
"No history found" line to make it emit `[]` under `--json`; the underlying reason a valid id reaches
that branch — no partial-id resolution — was never addressed.

## Fix

Resolve before querying, mirroring the sibling commands:

```go
issueID := args[0]
fullID, err := utils.ResolvePartialID(ctx, store, issueID)
if err != nil {
    return HandleErrorRespectJSON("failed to resolve issue: %v", err)
}
history, err := store.History(ctx, fullID)
```

(If parity with routed/contributor stores is also wanted, use the `cmd/bd/routed.go` primitives the
same way #4720 does for `mol bond` — but the minimum fix is the local `ResolvePartialID` above.)

## Repro

Any repo, any issue: `bd history <short-id>` → "No history found"; `bd history <full-id>` → works.
No special state required.

mohamedramadan14:
## What

`bd history <id>` now resolves partial/short issue IDs before querying, the same way every sibling id-command (`show`, `update`, `close`, `state`, `ready`, ...) already does.

## Why

Fixes #4868.

`bd history` was the one id-command that skipped `utils.ResolvePartialID`. It passed the raw arg straight to `store.History`, which does an exact match on the full ID. A short id like `jt0j` matched zero rows and printed "No history found" (exit 0) even for an issue with dozens of history entries — a false negative, not an empty result.

The fix resolves the id once up front and uses the resolved id for both the default history path and the `--events` path. If resolution fails (the id truly doesn't exist, or is ambiguous), the raw input is left as-is and falls through to the existing "No history found" / empty-JSON-array output instead of erroring. That preserves the exit-0/empty-result contract from #3502, which existing tests depend on for genuinely unknown ids — this change is only about fixing false negatives for ids that *do* resolve.

## Verification

- `go build -tags gms_pure_go ./...`
- `go vet -tags gms_pure_go ./cmd/bd/...`
- `BEADS_TEST_EMBEDDED_DOLT=1 go test -tags cgo,gms_pure_go -run TestEmbeddedHistory ./cmd/bd/` — added two subtests reproducing the issue (`short_id_resolves_like_full_id`, `partial_id_resolves_via_events_flag_too`); all 15 subtests in `TestEmbeddedHistory` pass, including the pre-existing nonexistent-id cases.
- `go test -tags cgo,gms_pure_go,integration -run TestCLI_ ./cmd/bd/...` — full fast CLI suite passes, no regressions.


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head and current main). Verdict: **needs changes** — the feature idea is good, but main has moved under it.

**Blocker — merge-base drift makes the auto-merged result uncompilable and bypasses the new backend seam.** Current main's `cmd/bd/history.go` no longer has the code shape this PR patches: it dispatches through `runHistory(rootCtx, store, ...)` with a proxied dual (`runHistoryProxiedServer`), and the local `ctx` this PR's inserted `ResolvePartialID(ctx, store, issueID)` relies on is gone — verified against main. A mechanical `ctx → rootCtx` fix is not enough: in proxied mode pre-run creates only the `uowProvider` and the global `store` stays nil, which `ResolvePartialID` rejects — and the PR discards that error. The resolution needs to move inside `runHistory`/the UOW-capable seam so embedded and proxied both get it.

**Should-fix — ambiguity and storage errors are silently eaten.** `ResolvePartialID` deliberately returns an ambiguity error for multi-match prefixes, but the PR discards every error and falls through to exact-match queries with the unresolved text — so an ambiguous short ID renders "No history" / empty JSON instead of asking the user to disambiguate. The test covers only a single unique ID.

**Should-fix — short IDs still can't reach deleted issues' history.** Resolution searches live issue tables; deletion removes the row, while `dolt_history_issues` still serves the full ID. So `bd history <full-id>` works on a deleted issue but `bd history <short-id>` resolves nothing and queries the wrong ID. Worth either documenting or resolving against history when the live lookup misses.

Rebase onto `runHistory`, propagate resolution errors, and add embedded+proxied coverage — then this is a nice usability win.

maphew (approved):
Approving — this supersedes my 2026-07-23 changes-requested review. The rebase-and-finish is applied as maintainer commits on this branch, preserving your resolution work:

- Rebuilt against main's current `runHistory` seam (the merged tree previously didn't compile — undefined `ctx`).
- Ambiguous prefixes now surface the candidate list through the standard error path (`utils.ErrAmbiguousID`, wrapped with `%w`), in both plain and `--json` output, instead of a silent "No history found"; not-found keeps the exit-0 contract (#3502).
- Proxied mode (no local store) skips resolution explicitly with the raw-ID passthrough documented.
- Your two embedded subtests are kept; an `ambiguous_partial_id_errors` subtest asserts both candidate IDs appear in plain stderr and the JSON error body (tightened after a cross-vendor review gate flagged the looser first version).

Validation: build/vet/gofmt clean; embedded history suite including the new subtest green; `TestResolvePartialID` suite unaffected. Merge tail goes to the pr-babysit patrol. Thanks mohamedramadan14.

Issue #94, opened 2025-10-21
"Error getting epic status ... converting NULL to string is unsupported"


neongreen:
After upgrading from beads 0.9.10 (or 0.9.11?) to 0.10.1:

```
❯ bd epic status --no-daemon
Error getting epic status: sql: Scan error on column index 9, name "assignee": converting NULL to string is unsupported
```

PR #4912, opened 2026-07-19
fix(history): coalesce NULL text columns in dolt_history_issues scan

### LINKED ISSUE #4867: `bd history <id>` crashes on any historical row with a NULL text column
## Summary

`bd history <full-id>` aborts with a scan error whenever a row in `dolt_history_issues` has NULL in
`description`, `design`, `acceptance_criteria`, or `notes`:

```
Error: failed to get history: get issue history: query error: failed to scan history:
sql: Scan error on column index 2, name "description": converting NULL to string is unsupported
```

This is the same class as #94 (which fixed it for `assignee`) — four more columns in the *history*
scan were missed.

## Root cause (released code, v1.1.0 @ `8e4e59d`)

`internal/storage/issueops/history.go` — and its duplicate `internal/storage/dolt/history.go` —
select those four columns (`history.go:22` / `:71`) and scan them straight into the plain `string`
fields of `types.Issue` (`history.go:50` / `:98`):

```go
&issue.ID, &issue.Title, &issue.Description, &issue.Design, &issue.AcceptanceCriteria, &issue.Notes, ...
```

`types.Issue.Description/Design/AcceptanceCriteria/Notes` are all plain `string`
(`internal/types/types.go`). Every *other* nullable text column in the very same scan
(`assignee`, `owner`, `created_by`, `close_reason`, `mol_type`) already uses `sql.NullString` — that
was the #94 fix. These four were never converted.

## When the NULL arises (not hypothetical)

Normal usage does not trigger it — bd stores `''`, not NULL, for an empty description (verified: live
`issues` table had 0 NULL / 50 empty-string descriptions). NULL appears in `dolt_history_issues` after
a **schema migration that re-creates a column**: Dolt's `dolt_history_*` tables project historical
rows against the *current* schema, so a column that gained new identity in the migration reads back
NULL at every pre-migration commit. This is exactly what #94 hit on the 0.9→0.10 upgrade.

Measured on one migrated DB (schema v32→v53): pre-migration commits had 188,261 rows with NULL
`description`; post-migration commits had 0. Every issue that existed before the migration is now
unreadable via `bd history`; new issues are fine.

## Impact

The data is intact and SQL-reachable — this is a read-path crash, not corruption. But `bd history`
becomes unusable for any issue older than the migration, with no workaround short of raw SQL.

## Fix

Single chokepoint — both stores route through `HistoryInTx`. COALESCE the four columns in its SELECT:

```sql
SELECT
  id, title,
  COALESCE(description,'')          AS description,
  COALESCE(design,'')               AS design,
  COALESCE(acceptance_criteria,'')  AS acceptance_criteria,
  COALESCE(notes,'')                AS notes,
  status, priority, ...
```

(or switch those four scan targets to `sql.NullString` to match their neighbours — same effect). The
duplicated SELECT/scan in `internal/storage/dolt/history.go` needs the same change.

## Repro

Any Dolt-backed repo whose schema migration re-creates one of these columns, then `bd history <id>`
on a pre-migration issue. Minimal SQL-level repro: any `dolt_history_issues` row with
`description IS NULL` makes `bd history <that id>` crash.

mohamedramadan14:
## What

`bd history <id>` no longer crashes when a historical row has NULL in `description`, `design`, `acceptance_criteria`, or `notes`.

## Why

Fixes #4867.

`dolt_history_issues` projects old commits against the *current* schema, so a column that gets recreated by a migration reads back NULL at every pre-migration commit, even though the live `issues` table enforces NOT NULL on it. The history scan (`internal/storage/issueops/history.go`, and its duplicate in `internal/storage/dolt/history.go`) scanned those four columns straight into plain Go `string` fields with no NULL handling, so any such row crashed with `sql: Scan error ... converting NULL to string is unsupported`. Every other nullable column in the same query (`assignee`, `owner`, `created_by`, ...) already handles this correctly via `sql.NullString` — these four were simply missed, the same class of bug as #94.

This is a read-path crash, not data loss — the data is intact and reachable via raw SQL, but `bd history` becomes permanently unusable for any issue that predates a migration touching one of these columns.

Fix: `COALESCE(col, '') AS col` on all four columns in both scan sites, so NULL reads back as the same empty string bd already uses for "no description."

## Verification

- `go build -tags gms_pure_go ./...`, `go vet -tags gms_pure_go ./internal/storage/...`
- Added regression tests across all three affected code paths (`DoltStore.History`, the dead-code duplicate `DoltStore.getIssueHistory`, and `EmbeddedDoltStore.History`), each reproducing a genuine NULL row in `dolt_history_issues` by relaxing the column's NOT NULL constraint and writing NULL directly, then committing — not just an empty string, which was never broken.
- `BEADS_TEST_EMBEDDED_DOLT=1 go test -tags cgo,gms_pure_go -run TestHistory_NullTextColumns ./internal/storage/embeddeddolt/...` passes.
- Confirmed the test actually catches the bug: temporarily reverted the COALESCE fix, reran, and got the exact error string from this issue (`sql: Scan error on column index 2, name "description": converting NULL to string is unsupported`). Restored the fix and confirmed it passes again.
- The `internal/storage/dolt` package tests need the external `dolt` CLI binary, which isn't available in my sandbox, so I couldn't execute those locally — they follow the same pattern as the embeddeddolt test that I did verify, and exercise the identical SQL fix.


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head before posting). Verdict: **needs changes** — the SQL repair is correct; the regression coverage doesn't prove the bug it fixes, and one new test covers dead code.

**The fix itself checks out.** Migration 0049 widened the four content columns `TEXT → LONGTEXT`; pinned Dolt's `dolt_history` projection (`history_table.go`) copies a historical value only when old and current column types are *exactly* equal, so pre-migration rows project as `nil`, and `database/sql` refuses to scan `NULL` into `string`. `COALESCE` in the scan query is the right repair.

**Should-fix 1 — the tests don't reproduce the defect.** Both new tests (`internal/storage/dolt/versioned_test.go` and the embedded twin) call `nullOutTextColumns` to write literal NULLs into the *current* row, then assert on `history[0]` — verified at the PR head. That exercises "current row happens to be NULL," which COALESCE also fixes, but never the reported failure mode: a *pre-migration historical* row projected to NULL by the type mismatch. The faithful shape: create + commit the issue, apply the `TEXT → LONGTEXT` ALTER, leave current values non-NULL, then assert the four fields on the *older* commit's history entry. As written, the tests would keep passing even if the history projection regressed differently.

**Should-fix 2 — 56-line regression test for a dead method.** `getIssueHistory` (`internal/storage/dolt/history.go:64`) has no production callers at the PR head — verified by grep; the shipped history paths go through `issueops.HistoryInTx` (both dolt and embeddeddolt stores). The new `history_test.go` coverage duplicates the versioned/embedded tests against a method nothing calls. Drop that test; retiring the unused method itself would make a good separate cleanup.

Drift note: the PR is ~42 commits behind main; the only relevant landed change (broadening `HistoryInTx` to `DBTX`) merges cleanly with the COALESCE patch — no semantic conflict.

With the test reshaped to the historical-row scenario and the dead-path test dropped, this is a clean land.

maphew (approved):
Approving — this supersedes my 2026-07-23 changes-requested review. Your COALESCE fix is correct and ships unchanged; the review's concern was that the tests wrote literal NULLs the schema never permits. I've reshaped them as a maintainer commit on this branch (`33238cc39`) to reproduce the *real* scenario: rows committed under the pre-0049 TEXT schema project as NULL through `dolt_history_issues` after the LONGTEXT migration —

- ALTER the four text columns down to TEXT + commit, create an issue with real values + commit, replay migration 0049's exact DDL + commit; assert the pre-migration entry's fields coalesce to `""` while the newest keeps its values.
- Applied in both the dolt and embeddeddolt suites; the literal-NULL helper and the duplicate scan-path test are removed.
- Guard proof: reverting your COALESCE makes the reshaped tests fail with the exact `converting NULL to string` scan error; restoring it passes.

Validation: build/gofmt clean, dolt + embedded history suites green, independent review + cross-vendor Codex gate signed off. Merge tail goes to the pr-babysit patrol. Thanks mohamedramadan14 — good catch on the projection behavior, and the fix itself needed no changes.

PR #4913, opened 2026-07-19
docs(async-gates): fix nonexistent gate subcommands in skill resource

vishnujayvel:
## What

`plugins/beads/skills/beads/resources/ASYNC_GATES.md` documents `bd gate eval`, `bd gate approve`, `bd gate close`, and a combined `--await <type>:<id>` flag. None of these exist on `main` — cobra falls through to parent help for all of them. This PR corrects every command in the doc to match the real subcommand tree (`create`, `list`, `show`, `check`, `resolve`, `add-waiter`, `discover`) and its actual flags.

## Why

Discovered this while building a community skill against `bd` 1.0.4 (`bd gate --help` there lists `add-waiter/check/create/discover/list/resolve/show`, no `eval`/`approve`/`close`) and confirmed the same surface is current on `main`. An agent following this doc verbatim gets a cobra usage error on the very first `bd gate create --await ...` example.

## How each replacement was verified

Ground truth is `cmd/bd/gate.go` on `main` (cross-checked against the auto-generated `docs/cli-reference/gate.md`):

- `bd gate eval` → `bd gate check` — `gateCheckCmd` (`cmd/bd/gate.go:523`, `Use: "check"`, `Short: "Evaluate gates and close resolved ones"`); `--dry-run` flag registered at `cmd/bd/gate.go:1015`.
- `bd gate approve <id>` → `bd gate resolve <id>` — `gateResolveCmd` (`cmd/bd/gate.go:469`, `Short: "Manually resolve (close) a gate"`); doc-comment explicitly says human gates require `bd gate resolve` (`cmd/bd/gate.go:290-296`). No `approve` command or `--comment` flag exists; `resolve` takes `--reason` (`cmd/bd/gate.go:1011`).
- `bd gate close <id>` → `bd gate resolve <id>` — there is no standalone `close` subcommand; `resolve` is documented as "equivalent to `bd close <gate-id>` but with a more explicit name" (`cmd/bd/gate.go:471-473`), so it also covers the doc's "manual close (any gate)" and "force close if stuck" cases.
- `--await <type>:<id>` → `--type <type> --await-id <id>` — `gateCreateCmd` has no `--await` flag; it registers `--type` (default `human`; allowed values `human`, `timer`, `gh:run`, `gh:pr`) and `--await-id` separately (`cmd/bd/gate.go:1024-1029`).
- `--title` / `--notify` → dropped, replaced with `--reason` where a description was implied — neither flag exists on `gate create`; the closest real equivalent is `--reason string` (`cmd/bd/gate.go:1027`).
- "Mail" gate type row → removed — grepped `cmd/bd/gate.go` and `internal/` for any `mail:` await-type handling; none exists. The only `mail` command in the codebase (`cmd/bd/mail.go`) is unrelated to gates.

## Scope

Deliberately minimal — no restructuring, no new sections, no touching the sibling resources. Every code block that referenced a nonexistent command/flag was corrected in place; everything else (tables like "Gates vs Issues", the troubleshooting flow, prose) is untouched.

Note for maintainers: `plugins/beads/skills/beads/CLAUDE.md` flags CLI-syntax duplication in resources as a staleness risk ("CLI command syntax ❌ No — Use `bd <cmd> --help`") — which is exactly what happened here. Out of scope for this PR, but might be worth a follow-up to trim the hardcoded flag lists in `ASYNC_GATES.md` down to conceptual guidance + a pointer to `bd gate --help`, the way `SKILL.md` already does for the resource index entry.

## Testing

Docs-only change to a plugin resource (not part of the Mintlify `docs/` tree, so `docs-render-check` doesn't cover it). Verified by reading `cmd/bd/gate.go` directly rather than running the CLI, since every replacement command's existence and flags are visible in the source.

### DIFF
--- a/plugins/beads/skills/beads/resources/ASYNC_GATES.md
+++ b/plugins/beads/skills/beads/resources/ASYNC_GATES.md
@@ -14,7 +14,7 @@
 | CI | `gh:run` | GitHub Actions run ID | Wait for GitHub Actions completion |
 | PR | `gh:pr` | PR number | Wait for PR merge |
 | Timer | `timer` | — | Deployment propagation delay |
-| Bead | `bead` | `<rig>:<bead-id>` | Wait for a cross-rig bead to close |
+| Bead | `bead` | `<bead-id>` (local) | Wait for another local bead to close |
 
 ---
 
@@ -37,10 +37,14 @@ bd gate create --type gh:pr --blocks bd-abc \
 bd gate create --type timer --blocks bd-abc \
   --timeout 15m
 
-# Cross-rig bead gate
+# Bead gate (waits for a local bead to close)
 bd gate create --type bead --blocks bd-abc \
-  --await-id other-project:op-abc123
+  --await-id bd-xyz
 ```
+
+> Note: cross-rig await IDs (`<rig>:<bead-id>`) are accepted at create time
+> but are not currently evaluable — the gate stays pending forever. Use a
+> local bead ID.
 
 **Required options**:
 - `--blocks <issue-id>` — Issue that stays blocked until the gate resolves
--- a/plugins/beads/skills/beads/resources/ASYNC_GATES.md
+++ b/plugins/beads/skills/beads/resources/ASYNC_GATES.md
@@ -12,42 +12,44 @@
 |------|-----------------|--------------|----------|
 | Human | `human` (default) | — | Cross-session human approval |
 | CI | `gh:run` | GitHub Actions run ID | Wait for GitHub Actions completion |
-| PR | `gh:pr` | PR number | Wait for PR merge/close |
+| PR | `gh:pr` | PR number | Wait for PR merge |
 | Timer | `timer` | — | Deployment propagation delay |
+| Bead | `bead` | `<rig>:<bead-id>` | Wait for a cross-rig bead to close |
 
 ---
 
 ## Creating Gates
 
 ```bash
-# Human approval gate
+# Human approval gate (timeout is stored but not enforced by check)
 bd gate create --type human --blocks bd-abc \
-  --reason "Approve production deploy" \
-  --timeout 4h
+  --reason "Approve production deploy"
 
 # CI gate (GitHub Actions)
 bd gate create --type gh:run --blocks bd-abc \
-  --await-id 123456789 \
-  --timeout 30m
+  --await-id 123456789
 
 # PR merge gate
 bd gate create --type gh:pr --blocks bd-abc \
-  --await-id 42 \
-  --timeout 24h
+  --await-id 42
 
-# Timer gate (deployment propagation)
+# Timer gate (deployment propagation) — --timeout is enforced by check
 bd gate create --type timer --blocks bd-abc \
   --timeout 15m
+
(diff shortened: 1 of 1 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication spot-verified against the PR head). Verdict: **needs changes** — the fix direction is right (the old text described nonexistent subcommands), but three behavioral claims in the corrected guide still don't match the code.

**Should-fix 1 — failure/closure semantics.** The guide says failed workflow runs and closed PRs *resolve* their gates. The code resolves success/skipped runs and merged PRs, and **escalates** failures, cancellations, and closed-unmerged PRs (escalated gates stay open; only `resolved` gates close) — there's even an explicit test pinning that a closed-unmerged PR does not resolve. The guide should say "resolve on success/merge; escalate on failure/closed-unmerged."

**Should-fix 2 — timeout guidance promises enforcement that doesn't exist.** The examples put timeouts on human and GitHub gates as forever-open protection, but timeout evaluation dispatches only for `timer` gates; human gates are skipped and the GitHub checkers never read `Timeout`. Either drop timeouts from those examples or state plainly that only timer gates enforce them.

**Should-fix 3 — the suggested lookup recipe can't find what it tells you to write.** The guide recommends descriptive reasons, then suggests `bd gate list | grep` to find them — but the reason is stored in `Description`, and list output prints neither title nor description. Point at JSON output with a filter instead.

Nit: the corrected Gate Types table still omits the implemented `bead` gate type.

For a guide whose purpose is to fix inaccurate gate instructions, these are worth one more pass — the structural corrections (real subcommand names, discover registration) all check out.

vishnujayvel:
Thanks @maphew — addressed the three should-fixes (and the bead-type nit) in the latest commit:

1. **Failure/closure semantics** — guide now says resolve on success/merge/timer; escalate (gate stays open) on failed runs / closed-unmerged PRs. Table in the Monitoring section.
2. **Timeouts** — only `timer` gates enforce `--timeout` via `bd gate check`; examples no longer imply human/GitHub forever-open protection from timeout alone.
3. **Lookup recipe** — plain `bd gate list` only prints id + await type/id; find-by-reason uses `bd gate list --json | jq …` against `description`/`title`.
4. **Nit** — added the implemented `bead` gate type row.

Please take another look when you have a moment.

maphew (approved):
Approving — this supersedes my 2026-07-23 changes-requested review. Your follow-up addressed all three should-fixes and the nit, verified against `cmd/bd/gate.go` on current main: resolve-vs-escalate semantics, timer-only `--timeout` enforcement, and the JSON-based find-gate-by-reason recipe are all now accurate.

One maintainer commit added on top (`308867f0f`): a cross-vendor review pass caught that the bead-gate row and example documented cross-rig await IDs (`other-project:op-abc123`), which `bd gate` accepts at create time but treats as permanently unevaluable — the gate would pend forever. The doc now points at local bead IDs with an explicit warning about the colon-qualified form.

Merge tail goes to the pr-babysit patrol. Thanks vishnujayvel — docs that match the binary are worth this much scrutiny.

PR #4911, opened 2026-07-19
feat(engine): guarded write ops — atomic metadata merge, row-version CAS, typed dep errors + events, bd close delegation

julianknutsen:
> **Consolidated PR** — replaces the fine-grained stack #4898, #4899, #4900, #4902, #4903, #4904, #4906 (closed, unmerged; their commits are all in this branch). Base is `main`; the diff below is the complete guarded-writes group. Follow-up groups: reads/keysets (#4922) and events/dependents/claim (#4926) stack on this.

## What

The complete guarded write surface for the engine, as one reviewable unit (each slice was developed and verified independently; commit history preserves the slices):

- **Atomic metadata merge** (`MergeMetadata`) — replaces whole-blob read-modify-write; `SlotSet`/`SlotClear` rebuilt on it; events staged correctly (was #4898)
- **Row version** — `row_lock` exposed as the Go-only `RowVersion` concurrency token; create stamps it, lifecycle writes rewrite it (was #4899)
- **Typed dependency errors** — cycle/self/type-conflict as sentinels + structured types on the root package; wrap-preserving, byte-identical messages (was #4900)
- **Dependency events restored** — `dependency_added`/`dependency_removed` emission (a regression from the storage migration), on explicit dep verbs only, wisp-aware event-table staging, both write plumbings (was #4902)
- **Close CAS** (`CloseIssueOptions.ExpectedVersion`) — optimistic-concurrency precondition inside the close transaction; version check runs before the guard and before Force (was #4903)
- **`bd close` delegation** — the CLI verb delegates to the engine's guarded close; duplicated guard logic deleted; historical refuse predicate and already-closed output preserved exactly (was #4904)
- **Update CAS** (`UpdateIssueChecked`) — same precondition for updates, composing atomically with ephemeral demotion (was #4906)
- **Proxied-mode close delegation** — the proxied-server `bd close` path delegates to the domain checked close; same sentinel, byte-identical messages (this branch's tip slice)

## Why

`bd` verbs and library callers each re-implemented guard/merge/version logic with read-then-write races. This moves every guard into the engine as single-transaction checked operations with typed errors (`errors.Is`/`errors.As` instead of string-matching), keeping user-facing behavior byte-identical.

## Verification

[...]

PR #4915, opened 2026-07-19
feat(comments): keyset-paged comments read (GetIssueCommentsPage)

julianknutsen:
> Stacked on #4911 (S7b) → #4906 (S5c) → … → #4892 (S1). Base is `feat/guarded-ops-s7b-proxied-close`; retarget down the stack as each lands. The diff below is S10 only.

## What

Adds `GetIssueCommentsPage` — a keyset-paged comments read (`(created_at, id)` ascending, index-backed resume predicate, limit default 100 / cap 500) — plus the composite index migration it needs, on both stores. The whole-thread `GetIssueComments` is untouched.

## Why

The comments surface was whole-thread only: any consumer wanting a page of a long thread had to materialize the entire thread and page in memory — with a cap, long threads get served incomplete. A keyset page read makes paging a bounded engine call: stable order, exact resume, `O(page)` per request. The pattern (sargable `created_at >= ? AND ((created_at > ?) OR (id > ?))`, same-second `id` tie-break, page-walk-equals-full-read) mirrors the engine's existing keyset reads.

Three non-obvious correctness points, all caught by pre-merge review and covered by tests:
- **Clone-safe migration.** Wisp tables are clone-local, and main migrations run before the clone-local chain materializes them — so the wisp half of the index migration is table-guarded (the 0035/0037/0053 precedent, with a matching no-op pin test), and fresh clones get the wisp index from the canonical clone-local schema instead. A bare `CREATE INDEX` would have broken every fresh clone's first writable open.
- **Cursor round-trip.** `AddIssueCommentInTx` returned an un-truncated Go time while the column stores `DATETIME(0)`; using that as a cursor could skip same-second rows. The returned `CreatedAt` is now truncated to match storage (stored value unchanged).
- **Honest plan guard.** The EXPLAIN test pins the composite index by its column-list signature (single-sourced from the exported production query) — it fails if the index is dropped, not just if the query errors.

Note on the migration number: 0056 is the only contiguous slot on this base (a gap-free version invariant is test-enforced). Other in-flight branches also carry an 0056; the loader panics loudly on duplicates, so whichever lands second renumbers on rebase — documented in the migration header.

## Verification

- Page-walk equals the full read exactly (5 same-second groups of 5, page size 4 — no dupes, no gaps); same-second tie-break asserted; first/past-end/empty/default-limit; wisp comments route to `wisp_comments`; migration no-ops cleanly on a DB without wisp tables (pin test through the dolt CLI); EXPLAIN shows an index seek on `(issue_id, created_at, id)`.
- Both stores covered; existing comments tests and the full schema suite green. `gofmt`/`go vet` clean; `go build ./...` under cgo and `CGO_ENABLED=0`; `golangci-lint` 0 issues.

```bash
go test ./internal/storage/dolt/ -run 'CommentsPage' -v
go test ./internal/storage/schema/ -run 'TestMigration0056|TestAllMigrationsSQL' -v
```

### DIFF
--- a/beads.go
+++ b/beads.go
@@ -64,6 +64,12 @@ type CloseIssueResult = storage.CloseIssueResult
 // name it without importing internal/storage.
 type UpdateIssueOptions = storage.UpdateIssueOptions
 
+// CommentPageCursor is the resume position for Storage.GetIssueCommentsPage — the
+// (created_at, id) of the last comment already returned, with the zero value
+// starting a walk from the beginning of the thread. Exported so consumers can
+// name it without importing internal/storage.
+type CommentPageCursor = storage.CommentPageCursor
+
 // RemoteStore provides dolt remote management and replication operations.
 // Use type assertion on a Storage value to access these methods:
 //
--- a/internal/jira/tracker_test.go
+++ b/internal/jira/tracker_test.go
@@ -714,6 +714,10 @@ func (s *configStore) AddIssueComment(_ context.Context, _, _, _ string) (*types
 func (s *configStore) GetIssueComments(_ context.Context, _ string) ([]*types.Comment, error) {
 	return nil, nil
 }
+
+func (s *configStore) GetIssueCommentsPage(_ context.Context, _ string, _ storage.CommentPageCursor, _ int) ([]*types.Comment, error) {
+	return nil, nil
+}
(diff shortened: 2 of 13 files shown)


julianknutsen:
Consolidated into #4922 for review throughput — same branch lineage, same commits, unchanged content; only the PR boundary moved. Closing unmerged.

PR #4916, opened 2026-07-19
bd: transactional events journal + hooks on both write plumbings

julianknutsen:
> **Re-port note.** This branch was rebuilt from scratch on current `main` (the previous head `3666e5026` is preserved in the PR history). The 2026-07-24 review found six blockers on a branch that had meanwhile fallen 600+ commits behind; rather than rebase through the guarded-ops/lifecycle re-architecture, the feature was re-ported final-shape-first, folding in the hardening and API-3.1 work staged on #5182 (attribution preserved via `Co-authored-by`). All six blockers are addressed — point-by-point reply on the review thread. The hooks rider is **split out to #5437** per the review.

## What this is

A durable, replayable **local journal of committed issue mutations** — a binlog/WAL-style feed for external tooling. **Off by default; zero overhead when off** (one atomic load per mutation). Enable per workspace with `bd config set events-journal true` or `BD_EVENTS_JOURNAL=1`.

Every durable bead mutation, on **both** write plumbings, inserts one row into `bd_events_journal` **inside the mutation's own transaction**: `{seq, ts, op, issue_id, issue_json, dep_json, comment_json}`. `issue_json` is the full post-mutation snapshot (issue + labels + `is_blocked`), `NULL` on delete. Ops: `create|update|close|delete|dep_add|dep_remove` on the wire, plus `comment` records carrying replayable comment payloads (`source: structured|audit` — a closed set exported from `issueops`). `seq` is **gapless and commit-ordered**, drawn from a single-row counter (`bd_events_seq`) bumped in-transaction; the counter self-heals from both a missing and a stale row. Journal failure fails the mutation — the journal can never lag, skip, or invent.

Consumers: `bd events tail --since <checkpoint>` (JSONL), `bd events export`, `bd events prune --before N`. A checkpoint below the retained window — or any interior sequence hole — returns a **typed truncation error** (`{"code":"events_journal_truncated","since":…,"floor":…,"head":…}`, exit 1) instead of silently skipping. Retention is **bounded by default**: a writer-pays, throttled auto-prune (modeled on SQLite's WAL auto-checkpoint / Dolt's auto-GC — no daemon required) trims the journal to the retention floors (`events-journal-retain-days`=7, `events-journal-retain-rows`=100k) in capped prefix batches after mutating commands, plus a ticker in `bd serve`. Floors are keep-more constraints; disabling both floors makes auto-prune a no-op (an explicitly unbounded ledger). `bd events prune` remains for earlier, on-demand cuts below the floors; `events-journal-auto-prune=false` disables the automation. **Prune — manual or automatic — only ever deletes a prefix of the sequence.**

## The riskiest things in this diff, named plainly

- **Two new migrations** — `0064_create_events_journal` + ignored `0022` (renumbered from the reviewed branch's colliding 0056/0014). Both tables are registered in `dolt_ignore` before the migration runs: no Dolt history growth, no sync/federation traffic, rows survive `dolt reset --hard`, clones start empty with `seq` at 1. Payload columns are `LONGTEXT` (a >64KB comment must not fail the user's write).
- **An emit call in every bead mutator.** Coverage is enforced structurally, not by convention: AST guards fail the build if any current or future bead-table mutator (either plumbing) stops journaling without a reasoned exemption, if a raw dolt-package transaction reaches a mutator without journal scoping, or if a store/provider construction site skips activation. Activation lives in the store/provider **factories** (the target workspace's own config governs), not at call sites.
- **`is_blocked` is now a documented optional field on the `/v0` OpenAPI Issue schemas.** The snapshot needs the field on `types.Issue`; with `x-go-type`-pinned schemas that puts it on the wire, so it is documented rather than left as an undocumented leak (`TestWireTagBijection` pins it).

## Deliberate contract boundaries (documented in help + docs page)

Per-branch working-set state; **per-replica** (seq spaces differ across replicas; checkpoints are not portable; `bd dolt pull`/merge-settle changes are not journaled — mirrors re-baseline after sync); raw `bd sql` DML, store-open-time migrations, and compaction/restore rewrites are not journaled (the compaction gap is inherited from the reference design and flagged as a follow-up).

## Fixed en route (found by adversarial review of this port)

- **A latent upstream defect in the API-3.1 reference**: the retain-days floor was a per-row `ts <` predicate over client-stamped clocks, so a prune under writer clock skew could carve a **mid-journal hole** the left-edge truncation check cannot see — silent record loss. Here, retain-days resolves to a seq bound (prune stays a pure prefix delete) **and** the truncation check sweeps batches for interior gaps. The enterprise port carries the latent form; flagged for the next sync.
- Wisp slot/metadata writes and `doltTransaction` comment writes that would have bypassed the journal; creation-time (imported) comments now journal so imports are replayable.

## Verification

`go build`/`vet` clean; migration hygiene clean; the full storage suites (issueops, domain/db, dolt targeted batches, embeddeddolt, schema, uow, config, httpapi) green; `-race` on the seq/concurrency/accessor paths green; `scripts/test.sh ./cmd/bd/` green; consumer golden rendered twice byte-identical; `make api-check` no drift; every commit passed the pre-commit lint hook. The AC03 end-to-end guard (restart-resume + pruned-past-must-be-told) is ported and green. Concurrency proofs: server-mode commit-order conflict/retry and embedded serialized writers, both `-race`-clean. **CI note:** the full `internal/storage/dolt` shard exceeds local timeouts under its 2-slot semaphore and must be exercised by CI on this PR.

## Provenance

Design and review by Claude Fable sessions, implementation by Claude Opus agents, four adversarially-verified multi-agent review councils across the slices (every blocker/major finding verified by two independent skeptics before action; ~60 reviewer/verifier agents total). Tracking: bd-opisf. Prior lineage: #4916 original + the #5182 residue (`3bfed2431`, `0a3fe0f9b`), attribution preserved.


maphew (dismissed):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the migration collision and representative claims at the PR head). Verdict: **needs changes** — the transactional events journal is a significant capability and the counter-concurrency reasoning checks out against the pinned engine, but six blockers stand between here and landable.

**Hard landing blockers:** (1) migration numbers collide — this PR's 0056/ignored-0014 are now owned by main's comments-index migrations (verified; main is through 0058), and the loader panics on duplicates at store open with *no* textual conflict to warn you; (2) the dependency-write sites conflict with main's newly landed `EmitEvent`-gated audit writes — resolution must keep both the conditional audit records and the unconditional journal records.

**Completeness blockers:** (3) comment mutations are unreplayable — both write seams record a generic update whose snapshot hydrates only the issue row + labels, so author/text/ID/time of comments can't be reconstructed, contradicting the "every mutation, exact history" contract; (4) active-wisp deletes (single and batch) route through `deleteWisp` paths that never call `RecordDeleteInTx` — and the structural guards inspect only `issueops`/`domain/db`, so they can't see it; (5) the UOW hook-parity rider overrides only selected methods — unoverridden wisp/claim-ready/bulk-dependency/wisp-label/wisp-comment methods dispatch straight to the embedded implementation on live proxied paths; (6) documented config doesn't work — the three new keys aren't in the YAML-only startup set, so `bd config set events-journal true` stores to the DB while enablement/retention read only Viper: journaling stays off and retention floors resolve to zero (unintended pruning).

**Standards:** journal SELECT/DELETE SQL duplicated in the CLI instead of extending the accessor; a new top-level command + wire contract with zero Markdown; the hook-parity rider is self-described as separate (split it); implementation-coupled coverage guards; missing Agent-Signature trailers on the four commits.

The core design — journal rows in the same transaction as the mutation — is right, and worth this rework.

julianknutsen:
@maphew — thank you for the 2026-07-24 review; every blocker is addressed. The branch was **re-ported onto current main** rather than rebased (it had fallen 600+ commits behind, across the guarded-ops/lifecycle re-architecture), folding in the #5182 residue with attribution. Point-by-point:

1. **Migration collision** → renumbered to `0064` / ignored `0022` (re-verified free against `main` at push time). Both tables `dolt_ignore`d before the migration runs, following the 0062/ignored-0019 audit-events precedent. Payload columns widened to `LONGTEXT` while the table has never shipped (a >64KB comment must not roll back the user's write — the reviewed branch and the enterprise port both carry that latent `TEXT` trap).

2. **Coexistence with the audit `events` writes** → journal emission is unconditional at every bead-table write leaf and never gated by `EmitEvent`; the audit writes are untouched. Both rows land in the same mutation transaction. The docs page adds an orientation section distinguishing hooks / audit history / journal.

3. **Comment mutations unreplayable** → there is now an engine-side `comment` op carrying the replayable payload (`comment_json`, `source: structured|audit` as a closed exported set). Creation-time (imported) comments journal too — the golden pins the shapes.

4. **Active-wisp deletes never journaled** → wisp create/update/close/delete/promote and wisp slot/metadata writes all journal; the structural guards now scan both plumbings plus raw dolt-package transactions, so a wisp path cannot silently regress again.

5. **UOW hook-parity rider** → **split out to #5437** as requested — full-surface NotifyingProvider with a reflection partition guard, already green (100/100 checks) and awaiting review. This PR contains no hooks changes.

6. **`bd config set events-journal true` did not work** → fixed via the YAML startup-set (`YamlOnlyKeys`) and pinned by an end-to-end test that enables the journal purely through `bd config set` and observes rows. Env (`BD_EVENTS_JOURNAL*`) works alongside; retention floors now default 7d/100k instead of 0.

Standards items: query bodies live once at the issueops seam (both stores' accessors and the proxied path share them); a docs page ships in this PR; all commits carry `Agent-Signature` trailers.

Also fixed en route, found by adversarial review of the port: a latent defect in the API-3.1 retain-days design (per-row client-clock predicate → possible mid-journal hole invisible to the left-edge truncation check). Prune now resolves to a pure seq-prefix delete and the truncation check sweeps for interior gaps. The enterprise lineage carries the latent form; flagged for the next sync.

One policy call surfaced for you rather than decided silently: the docs release-pin decision (2026-07-17) says hand-written pages document the pinned release (v1.1.0), and the new `docs/reference/events-journal.md` page describes a main-only feature. It ships in this PR because the review asked for docs; if you'd rather hold the page until the pin bumps, it's an isolated `docs(events)` commit that can be dropped cleanly (the CHANGELOG entry sits under Unreleased either way). `docs/reference/configuration.md` deliberately does not list the three new keys yet — its freshness-marker contract makes that a full re-review, filed as follow-up for pin-bump time.

Re-requesting review. The full `internal/storage/dolt` shard needs CI (locally it exceeds the 2-slot semaphore timeout); everything else listed in the PR body ran green locally.

julianknutsen:
## Review record (multi-agent, adversarially verified)

Per the project owner's direction, this comment records the review this branch received before merge. All reviews were performed by Claude agent councils — each finding independently attacked by two skeptic agents before being accepted — and every confirmed finding was fixed and mutation-verified (revert the fix → the guarding test fails → restore → green). This record is agent work and does not claim to be a human review.

**Coverage, by slice:**

- **Engine (S1)**: 3-lens council (transactional/seq correctness, coverage falsification both directions, schema/dolt_ignore/snapshot fidelity), 17 agents. Confirmed and fixed: two unscoped wisp-slot transactions that silently skipped journaling (plus a structural scoping guard so the class cannot recur), and `comment_json`/`dep_json` widened to LONGTEXT (a >64KB comment must not roll back the user's write).
- **Surface (S2)**: 3-lens council, 11 agents, 4 confirmed / 0 refuted. Headline: a **latent defect inherited from the reference design** — the retain-days floor was a per-row client-clock predicate, so a prune under writer clock skew could carve a mid-journal hole invisible to the left-edge truncation check (silent record loss). Fixed by resolving retain-days to a seq bound (prune is provably prefix-only) plus an interior-gap sweep in the truncation check. Also fixed: `bd serve` and routed/cross-workspace opens ran with the journal silently off — activation moved into the store/provider factories with a construction-site guard.
- **Fix-wave closure**: independent closure round — Y1 prune fixes verified on all counts (bound math, boundary cases, engine-level skew-pair and interior-hole tests confirmed to fail on the reference form); the factory-activation audit then caught a false guard exemption (three `bd doctor --fix` handlers do mutate beads — they now journal through an activating factory).
- **Auto-prune (S5, the final commit)**: 3-lens council, 5 agents. Confirmed and fixed: `bd events tail` passed the trigger gate (reader-pays violation; `export` escaped only via a leaf-name collision — both now excluded by independent defences, each mutation-verified). Hardened from the minors: watermark stamped in its own transaction (failing bound reads degrade to no-maintenance instead of per-command retries), no-retry 30s-bounded maintenance transactions, forward-clock-skew watermark guard, YAML `no/off` spellings honored for the opt-out (a hand-written `no` previously parsed as "keep deleting"), `ts` index added while the table is unshipped, config keys registered with `bd config set`, and docs/help wording corrected ("earlier cut below the floors", not "tighter than the floors").

**Verification state at head `d05d94626`:** full storage suites, `-race` on all concurrency paths, consumer golden rendered twice byte-identical, AC03 end-to-end guard, e2e coverage for embedded/proxied/serve/routed topologies, `make api-check` clean, docs render checks clean, every commit through the pre-commit lint hook. One pre-existing failure in `internal/storage/dolt` (`TestSchemaMigrationRejectsChangedPreExistingDirtyTable`, the `custom_statuses` backfill guard) reproduces at the branch base with the journal's `doltIgnorePatterns` entries removed — unrelated to this PR; the full dolt shard runs in CI.

PR #4918, opened 2026-07-19
fix(create): reject --waits-for-gate when --waits-for is absent; add regression tests

### LINKED ISSUE #4917: bd create --waits-for-gate: silent no-op without --waits-for; half-write on invalid value
## Summary

Two bugs in `bd create`'s `--waits-for-gate` flag, both of the silent-drop family — the instrument and the operator's belief diverge with nothing surfacing the divergence.

### Defect A — silent no-op without `--waits-for`

```bash
mkdir /tmp/wfg && cd /tmp/wfg && bd init --prefix pt
SPAWN=$(bd create "spawner" -t task --json | jq -r .id)
bd create "child" --parent $SPAWN -t task

# Both exit 0 and create a normal bead with no gate, no dependency, no warning:
bd create "probe" -t task --waits-for-gate all-children    ; echo $?   # -> 0
bd create "probe" -t task --waits-for-gate TOTALLY-BOGUS   ; echo $?   # -> 0
```

A value that is rejected outright when `--waits-for` is present is accepted silently when it is absent. An operator who writes `--waits-for-gate` believing they armed a hold gets a success message and a fully-dispatchable bead instead.

**Root cause:** `buildWaitsFor` returns `nil, nil` when `spawnerID` is empty, regardless of whether `--waits-for-gate` was explicitly supplied by the caller.

### Defect B — half-write when both flags are present with an invalid gate value

> **Note:** This defect is already fixed structurally by the `create_atomic.go` refactor that moved validation before `createIssueWithDeps`. The issue and PR both document it because (a) it was present in v1.1.0, (b) the fix has no regression test, and (c) the two defects share the same root area.

```bash
before=$(bd list --limit 0 --json | jq length)
bd create "jsonpath probe" -t task --waits-for $SPAWN --waits-for-gate TOTALLY-BOGUS
# -> exit 0, issue CREATED with NO dependency — appears in bd ready
after=$(bd list --limit 0 --json | jq length)
# before < after despite exit 0 with no dep added
```

Any caller treating a nonzero exit as "nothing was created" is left with a dep-less, permanently-ready orphan bead every time — the exact shape that re-offers work nobody thinks is dispatchable.

## Why this matters

Both defects manufacture dispatchable beads that a caller believes do not exist (A) or believes are properly held (B). In an orchestrated multi-agent system, this translates directly to premature dispatch.

## Binary

```
go version -m $(which bd)
  path  github.com/steveyegge/beads/cmd/bd
  mod   github.com/steveyegge/beads  (devel)  vcs.modified=true
```

## Fix

See the linked PR.

jacobhausler:
Fixes #4917.

## What changed

Three files changed in production code, two in tests.

### `cmd/bd/create_deps.go` — `buildWaitsFor` signature change

Added a `gateExplicit bool` parameter. When `spawnerID` is empty AND the flag was
explicitly supplied (`gateExplicit=true`), return an error instead of silently
returning `nil`. This fixes **Defect A**.

```go
// before
func buildWaitsFor(spawnerID, gate string) (*domain.WaitsForSpec, error) {
    if spawnerID == "" {
        return nil, nil  // silently drops the explicit flag
    }
    ...
}

// after
func buildWaitsFor(spawnerID, gate string, gateExplicit bool) (*domain.WaitsForSpec, error) {
    if spawnerID == "" {
        if gateExplicit {
            return nil, fmt.Errorf("--waits-for-gate requires --waits-for (no spawner ID specified)")
        }
        return nil, nil
    }
    ...
}
```

### `cmd/bd/create_input.go` — `waitsForGateSet` field

Added `waitsForGateSet bool` to `createInput` (mirroring the existing `repoOverrideSet`
pattern), populated via `cmd.Flags().Changed("waits-for-gate")` in `gatherCreateInput`.

### `cmd/bd/create.go` and `cmd/bd/create_proxied_server.go` — call site updates

Both call sites of `buildWaitsFor` now pass the explicit-flag boolean:
- Direct (non-proxied) path: `cmd.Flags().Changed("waits-for-gate")`
- Proxied-server path: `in.waitsForGateSet`

### `cmd/bd/create_deps_test.go` — unit tests for `buildWaitsFor`

Updated existing call sites to pass the new parameter. Added a new sub-test
`explicit_gate_without_spawner_returns_error` that confirms both `"all-children"`
(valid value, wrong context) and `"TOTALLY-BOGUS"` (invalid value) are rejected
when `gateExplicit=true` and `spawnerID=""`.

### `cmd/bd/create_deps_atomic_test.go` — integration tests (real bd binary + embedded Dolt)

Two new sub-tests added to `TestCreateDepsAtomicity`:

**`waits_for_gate_without_waits_for_is_rejected`** — Defect A regression test.
Runs `bd create --waits-for-gate <gate>` (no `--waits-for`) for both a valid and
invalid gate value. Asserts exit ≠ 0 AND that no issue persists in `bd list`.
This test **fails on the unfixed code** (both commands exit 0 today).

**`invalid_waits_for_gate_value_is_rejected_before_write`** — Defect B regression test.
Runs `bd create --waits-for <id> --waits-for-gate TOTALLY-BOGUS`. Asserts exit ≠ 0
AND that the issue count does not change. Guards the existing `create_atomic.go`
fix against regressions.

Both sub-tests are also added to the `ready_never_offers_a_failed_create` guard.

## Demonstrated failing cases (FALSIFIABLE-CHECK FLOOR)

Against the current unpatched binary:

```
# Defect A — exits 0, creates bead with no gate or dependency
$ bd create "probe" -t task --waits-for-gate all-children   ; echo $?
✓ Created issue: pt-v1p — probe  exit=0
$ bd create "probe" -t task --waits-for-gate TOTALLY-BOGUS  ; echo $?
✓ Created issue: pt-8wt — probe  exit=0

# Defect B — exits 0, creates bead without the waits-for dependency
$ bd create "probe" -t task --waits-for $SPAWN --waits-for-gate TOTALLY-BOGUS ; echo $?
✓ Created issue: pt-e5u — probe  exit=0
$ bd dep list pt-e5u --json
[]    # no dependencies
$ bd ready --json | jq '.[].title'
"probe"   # dispatchable
```

Against the patched binary:

```
$ bd create "probe" -t task --waits-for-gate all-children
Error: --waits-for-gate requires --waits-for (no spawner ID specified)  exit=1

$ bd create "probe" -t task --waits-for $SPAWN --waits-for-gate TOTALLY-BOGUS
Error: invalid --waits-for-gate value "TOTALLY-BOGUS" (valid: all-children, any-children)  exit=1
# bead count unchanged

$ bd create "probe" -t task --waits-for $SPAWN --waits-for-gate all-children
✓ Created issue: fx-77d — probe   # works as expected
$ bd show fx-77d | grep -A2 DEPENDS
DEPENDS ON
  → ○ fx-riw: spawner  # dependency wired correctly
```

## Binary used for verification

```
go version -m $(which bd)
  path  github.com/steveyegge/beads/cmd/bd
  mod   github.com/steveyegge/beads  (devel)  vcs.modified=true
```


maphew (changes requested):
Cross-vendor paired agent review (Claude Opus + Codex gpt-5.6-sol, independent traces, adjudicated against the PR head before posting). Verdict: **needs changes** — the single-create fix is correct and well-tested, but the same defect survives on two other routes.

**The fix itself is right (verified):** threading `cmd.Flags().Changed("waits-for-gate")` into `buildWaitsFor` is the only reliable way to distinguish explicit use from cobra's `all-children` default; the rejection semantics match the flag's meaning (a gate type is meaningless without a `--waits-for` spawner — `domain.WaitsForSpec` always pairs them, and pre-fix the gate was silently dropped, i.e. Defect A); all three callers are updated including the proxied path via `createInput.waitsForGateSet`; the regression matrix covers rejected and still-valid combos plus the no-half-write contract; the error message names both flags.

**Should-fix (adjudicated and confirmed): the batch routes still silently drop the flag.** `--file` and `--graph` dispatch and return at the top of RunE (`cmd/bd/create.go:57-85`), before the waits-for flags are ever read — so `bd create --file plan.md --waits-for-gate TOTALLY-BOGUS` succeeds and creates ungated issues, which is exactly Defect A on another door. The proxied route rejects these combinations via `singleIssueOnlyFlags` (`create_input.go:270-298`), so the local and proxied routes now disagree. The cheap, consistent fix is to hoist a `singleIssueOnlyFlags`-style rejection above the `--file`/`--graph` dispatch (which would close the same hole for any other single-issue-only flag, not just this one).

**Should-fix: with `--parent`, an invalid combo still advances the child counter before rejection.** `GetNextChildID` runs at `create.go:470`, before the validation at `:537`, and its upsert commits — so a rejected create burns a child ID. Pre-existing ordering, but since this PR is about rejecting the combo cleanly, moving the flag validation above the parent/child-ID work is in scope and free.

**Nit:** the help text and generated `docs/cli-reference/create.md` don't mention that `--waits-for-gate` requires `--waits-for`; one line in the flag help (then the docs regen) closes it. Also `--dry-run` on the single path returns before this validation, so a dry-run won't warn about the bad combo — worth a look while in here.

With the batch-route rejection hoisted and the help text updated, this is a clean land.

maphew (approved):
Approving — supersedes my 2026-07-23 changes-requested review; both should-fixes are applied as maintainer commits. Batch routes (`--file`/`--graph`) now reject `--waits-for-gate` via the existing `rejectSingleIssueFlags*` helpers (mirroring the proxied path), and child-ID reservation moved after waits-for/dependency validation so a failed create no longer burns an ID. Three regression subtests pin all of it. Your core `buildWaitsFor` fix ships unchanged.

Validation: build/vet/gofmt clean, Create/WaitsFor set green; independent review + cross-vendor Codex gate signed off. Merge tail to the pr-babysit patrol. Thanks jacobhausler.

PR #4921, opened 2026-07-20
feat(storage): (created_at, id) keyset issue-list predicate on IssueFilter

julianknutsen:
> Stacked on #4915 (S10) → … → #4892 (S1). Base is `feat/guarded-ops-s10-comments-keyset`; retarget down the stack as each lands. The diff below is S11 only. Per the chain's review model, the label is added only when this PR becomes head-of-chain.

## What

Ports the proven `(created_at, id)` keyset issue-list predicate onto the stack: `IssueFilter` gains `AfterCreatedAt *time.Time` + `AfterID string`, emitted through the shared `sqlbuild.BuildIssueFilterClauses` path as the single-sourced sargable `KeysetCreatedAtIDPredicate`, honored by `SearchIssues`/`SearchIssuesWithCounts` on both backends. Authorship preserved (cherry-pick).

## Why

`bd list` and the library's own callers had only `CreatedBefore + Limit` — with second-granularity timestamps, a same-second group larger than one page cannot page without dropping or duplicating its overflow. The `(created_at DESC, id ASC)` tuple cursor (id = primary key, total tie-break) closes that, matching the keyset pattern the engine already uses for the comments page read.

## Verification

- Same-second-overflow paging regressions through the public `SearchIssues` on **both** server and embedded Dolt; a dolt EXPLAIN guard pins the plan against the exact production predicate constant; a sqlbuild unit test pins clause/arg emission.
- Composes with every existing filter; no interaction with the stack's column-list changes (WHERE-only). `gofmt`/`go vet`/builds clean; `golangci-lint` 0 issues.

```bash
go test ./internal/storage/dolt/ -run 'TestSearchIssuesKeyset' -v
BEADS_TEST_EMBEDDED_DOLT=1 go test ./internal/storage/embeddeddolt/ -run TestSearchIssuesKeysetEmbedded -v
```

### DIFF
--- a/internal/storage/dolt/keyset_search_test.go
+++ b/internal/storage/dolt/keyset_search_test.go
@@ -1,0 +1,163 @@
+package dolt
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/steveyegge/beads/internal/storage/sqlbuild"
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// ksIDs projects issue IDs in result order.
+func ksIDs(issues []*types.Issue) []string {
+	out := make([]string, len(issues))
+	for i, iss := range issues {
+		out[i] = iss.ID
+	}
+	return out
+}
+
+func ksEqual(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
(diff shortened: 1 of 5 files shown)


julianknutsen:
Consolidated into #4922 for review throughput — same branch lineage, same commits, unchanged content; only the PR boundary moved. Closing unmerged.

PR #4922, opened 2026-07-20
feat(storage): keyset reads — comments page, issue-list cursor, IsBlocked filter

julianknutsen:
> **Consolidated PR** — replaces #4915, #4921 (closed, unmerged; their commits are in this branch). Stacks on the guarded-writes group #4911; retarget to `main` when it lands. The diff vs its base is the complete keyset-reads group.

## What

The bounded read surface, as one unit (commit history preserves the slices):

- **Keyset comments read** (`GetIssueCommentsPage`, was #4915) — `(created_at, id)` page read with a sargable resume predicate, composite index migration 0056 (clone-safe: wisp half table-guarded per the 0035/0037/0053 precedent + no-op pin test; fresh clones get the wisp index from the canonical clone-local schema), EXPLAIN plan guard, cursor-safety fix (`AddIssueComment` returns second-truncated `CreatedAt`)
- **Keyset issue-list predicate** (`IssueFilter.AfterCreatedAt`/`AfterID`, was #4921) — exact resume for same-second groups through the shared filter-clause builder, both backends
- **`IsBlocked` filter** (this branch's tip slice) — the denormalized `is_blocked` column becomes a real server-side filter for search and grouped counts

## Why

Whole-thread/whole-list reads force callers to materialize and page in memory (with caps that silently truncate long threads). Keyset reads make paging a bounded engine call with exact resume.

## Verification

Page-walk == full-read under same-second collisions on both backends; migration no-op pin test through the dolt CLI; EXPLAIN guards pin index seeks against the exact production predicates; conformance coverage for the grouped-count filter. Builds/vet/lint clean.

PR #4924, opened 2026-07-20
feat(storage): durable events reads — EventsSince keyset feed with per-bead scope

julianknutsen:
> Stacked on #4922 (S12) → … → #4892 (S1); base `feat/guarded-ops-s12-isblocked-filter`. Diff below is this slice only. Per the chain's review model, the label arrives when this PR becomes head-of-chain.

## What
Ports the durable events read surface: `EventsSince` — a keyset `(created_at, id)` feed over the events table with an optional per-bead scope — plus `EventCursor`, the `EventClaimed` constant, and `GetDependentRecords`. Sargable predicate, single-sourced into the EXPLAIN guard. Authorship preserved (cherry-picks).

## Why
`bd` verbs that render event history (`bd show`'s per-bead history, activity views) and library callers had only limit-bound newest-first reads or a timestamp watermark — no stable resumable cursor. The keyset feed gives exact resume with same-second correctness, matching the engine's established keyset pattern.

## Verification
Events keyset tests + EXPLAIN plan guard (single-sourced from production SQL) on real Dolt; embedded feed test; `Event` regression sweep; builds cgo+nocgo, vet, lint 0 issues.
```bash
go test ./internal/storage/dolt/ -run 'TestEventsSince' -v
```

### DIFF
--- a/internal/storage/dependency_queries.go
+++ b/internal/storage/dependency_queries.go
@@ -10,6 +10,14 @@ import (
 type DependencyQueryStore interface {
 	GetDependencyRecords(ctx context.Context, issueID string) ([]*types.Dependency, error)
 	GetDependencyRecordsForIssues(ctx context.Context, issueIDs []string) (map[string][]*types.Dependency, error)
+	// GetDependentRecords returns raw dependency rows whose target is targetID
+	// (the inbound edges), without hydrating the source issues. depType filters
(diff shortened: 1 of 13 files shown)


julianknutsen:
Consolidated into #4926 for review throughput — same branch lineage, same commits, unchanged content; only the PR boundary moved. Closing unmerged.

PR #4925, opened 2026-07-20
feat(storage): bounded dependents reads + batched transitive-blocked read

julianknutsen:
> Stacked on the S13 events PR → … → #4892. Diff below is this slice only. Per the chain's review model, the label arrives when this PR becomes head-of-chain.

## What
Ports the bounded dependents read family — `GetDependentRecords` id-cursor paging + `CountDependentRecords` + batched `GetDependentRecordsForIssues` (with the cross-table depid dedup fix) — and `IsBlockedBatch` (one batched read of the transitive is_blocked column, issues-wins on id collision, matching single-row `IsBlocked`). Storage layer only; the root re-exports land in the next slice with the interfaces they extend. Authorship preserved.

## Why
`bd dep list`/dependents views and library callers materialized unbounded dependent sets and did N per-row `IsBlocked` calls for a page of beads. Bounded, batched, index-backed reads close both.

## Verification
Dependents cursor/count/batch + collision-dedup + IsBlockedBatch parity tests on real Dolt and embedded; regression sweep; builds/vet/lint clean.
```bash
go test ./internal/storage/dolt/ -run 'TestGetDependentRecords|TestIsBlockedBatch|TestCountDependentRecords' -v
```

### DIFF
--- a/internal/storage/dependency_queries.go
+++ b/internal/storage/dependency_queries.go
@@ -20,6 +20,14 @@ type DependencyQueryStore interface {
 	// group membership reads that must see dangling children. RAW: applies no
 	// visibility policy — the caller filters at hydration.
 	GetDependentRecords(ctx context.Context, targetID string, depType string, limit int, afterID string) ([]*types.Dependency, error)
+	// GetDependentRecordsForIssues returns raw dependency rows keyed by TARGET id:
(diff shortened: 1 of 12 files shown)


julianknutsen:
Consolidated into #4926 for review throughput — same branch lineage, same commits, unchanged content; only the PR boundary moved. Closing unmerged.

PR #3632, opened 2026-05-01
feat(storage): refuse test DBs on production dolt servers (AD-01)

quad341:
## What this changes

Adds two layered guards that prevent test data from polluting a production Dolt server, and generalizes the existing production-port check so it works on deployments that don't use the legacy default port.

**1. Production-port detection beyond `3307`.** `isProductionPort` previously only flagged `cfg.ServerPort == DefaultSQLPort`. It now also matches a port set via the new `BEADS_PRODUCTION_PORT` env var, or a `dolt-server.port` file in an operator-set `BeadsDir`. The existing `BEADS_TEST_MODE=1` panic and the lower-layer firewall in `openServerConnection` both delegate to this helper, so installs on non-3307 ports gain the same protection without callsite changes.

**2. Database-name firewall in `New()`.** Refuses to open any database whose name matches `testDatabasePrefixes` (now extended with `benchdb_`) on any server, unless the operator has explicitly opted into a dedicated test-server lane via `BEADS_TEST_SERVER=1`. Returns an `error` (not a panic) so tests can assert on it. The error names the DB and the address for forensics.

Both defenses share the single `BEADS_TEST_SERVER=1` opt-in — a test harness sets one env var and is consistently allowed through every layer.

## Review notes

- **New env vars**:
  - `BEADS_TEST_SERVER` — strict `=='1'` opt-in for test-server connections. Not "truthy" — stray values do not open the firewall.
  - `BEADS_PRODUCTION_PORT` — extra production-port indicator, parsed as int, matched against `cfg.ServerPort`.
- **Defaults are safe.** The firewall is closed unless explicitly opted in. There is no implicit fallback.
- **Why error rather than panic in `New()`.** The DB-name firewall returns `error` so callers (including tests) can assert. The production-port hard guard still panics — it is defense-in-depth that is only reachable if `applyConfigDefaults`'s sentinel-port rewrite is bypassed.
- **Why Rule 3 doesn't read `filepath.Dir(cfg.Path)`.** When `BeadsDir` is empty, `productionPortReasons` deliberately does NOT fall back to `filepath.Dir(cfg.Path)`. Test fixtures commonly set `cfg.Path` under `/tmp` without a real `BeadsDir`; falling back there would treat any leaked `/tmp/dolt-server.port` from a stray dev server as production. See the doc-comment at `internal/storage/dolt/store.go:107-114`.
[...]

PR #4928, opened 2026-07-20
fix(backup): report embedded database size

### LINKED ISSUE #4920: backup status reports 0 B for non-empty embedded Dolt databases
## Summary

In the default embedded Dolt mode, `bd backup status` reports:

```text
Database size: 0 B
```

and `bd backup status --json` reports:

```json
"database_size": {
  "bytes": 0,
  "human": "0 B"
}
```

The active database and filesystem backup are both non-empty. `bd backup sync` completes successfully and writes non-empty `.darc` artifacts, so this appears to be a status-reporting defect rather than a sync failure.

## Environment

- beads: `1.1.0` (`8e4e59d39f34`)
- Storage: default embedded Dolt
- Backup: local `file://` destination
- OS: Windows 11 Pro amd64, build 26200
- Go: not installed; reproduced with the official prebuilt Windows binary
- No external orchestrator involved

The observed active embedded database contained approximately 32.4 MB of files. Four synchronized `.darc` files totalled 6,968,193 bytes, while status reported zero bytes.

## Minimal reproduction

Using stock Beads without an orchestrator:

```sh
git init beads-backup-size-repro
cd beads-backup-size-repro

bd init --non-interactive --skip-agents --skip-hooks
bd create "size probe" --type task --priority 2

bd backup init ../beads-backup-size-destination
bd backup sync

bd backup status
bd backup status --json
```

Inspecting the destination confirms that the sync wrote non-empty `.darc` files, while both status forms report a zero-byte database.

## Expected behavior

`bd backup status` should report a non-zero approximate size for a non-empty embedded database.

If the active database directory cannot be resolved or read, status should surface that condition rather than returning a successful zero-byte measurement indistinguishable from an empty database.

The reported source-database size does not need to equal the compressed backup-artifact size.

## Suspected cause

[`doltBackupSize`](https://github.com/gastownhall/beads/blob/v1.1.0/cmd/bd/backup_dolt.go#L351-L373) resolves the data directory through `doltserver.ResolveDoltDir`, while the default embedded backend stores its data under `.beads/embeddeddolt`.

The resulting server-oriented path does not exist in this embedded workspace. `dirSize` then suppresses filesystem walk errors, including the missing-root error, causing the absent directory to become `(0, nil)` and appear as a valid measurement.

Current [`main`](https://github.com/gastownhall/beads/blob/main/cmd/bd/backup_dolt.go#L358-L379) appears to retain the same resolution and error-suppression path.

## Suggested regression coverage

- Initialize a default embedded database and create an issue.
- Assert that `bd backup status --json` reports `database_size.bytes > 0`.
- Assert that human-readable status does not report `Database size: 0 B`.
- Verify that a missing or unreadable resolved root does not silently become a successful zero-byte result.
- Preserve coverage for server mode and custom Dolt data-directory configuration.

If this diagnosis and expected behavior match the intended contract, I would be happy to prepare a focused, regression-tested fix.

Would you prefer the size to come from the active storage backend, or is mode-aware filesystem resolution the intended design?

---

ecuthiell:
## What

Report `backup status` database size through a narrow, optional, per-instance `storage.ActiveDatabaseSizer` capability.

- embedded and owned/shared local Dolt stores measure only their active database directory, excluding sibling databases;
- external, proxied, gateway, TLS, socket, remote-host, and otherwise ambiguous instances report the capability as unsupported;
- unsupported size is omitted from both human and JSON status while the rest of the status still renders;
- a store that declared authoritative local ownership but cannot measure its directory fails visibly instead of returning a false zero; and
- decorated stores are unwrapped before capability selection.

The filesystem walk is context-aware, overflow-checked, tolerant of live file deletion, and does not follow reparse-point/symlink directories.

## Why

Fixes #4920.

Embedded Dolt stores their active database below `.beads/embeddeddolt`. The old server-oriented resolver selected the wrong root and suppressed filesystem errors into a successful zero-byte result. Conversely, treating every server-mode client path as local would measure unrelated or nonexistent data for external servers. The per-instance capability makes local authority explicit and keeps unsupported storage optional.

## Verification

At exact head `26dbf3e0b0f8f32cfb69bc2a3fc94c28998eaaac`:

- focused command/process tests cover human and JSON streams, omitted unsupported values, declared-local failures, and a real embedded backup;
- storage tests cover embedded siblings, owned/shared local instances, external/stale-local paths, gateway/proxied/socket/TLS/remote cases, decorator unwrapping, live-walk churn, cancellation, and overflow;
- a mixed-instance regression retains owned instance A and external instance B across legacy-endpoint, shared-mode, and shared-plus-explicit-endpoint flips, proving each instance keeps its captured authority;
- focused normal and race suites, `go vet`, `git diff --check`, and pinned `golangci-lint` with `--new-from-rev=upstream/main` pass with zero issues.

The fresh exact-head required remote checks remain stop conditions before merge.

---


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the embedded fix is right; server modes regress.

**Blocker — external server mode now fails `bd backup status` outright.** Verified at the PR head: `doltBackupSizeFor` walks `locator.Path()` and returns an error on any failure, which the status path treats as fatal. But in server mode `Path()` is a client-local `ResolveDoltDir` result, and the repo itself documents that no local database directory is expected there (`doctor/fresh_clone_server.go`). Previously the sizing error was suppressed and the rest of status rendered; now an external server with no local `.beads/dolt` exits 1 — and a vestigial local directory silently reports an unrelated size. Storage sizing needs to distinguish locally-measurable stores from external ones (a capability or a "size unavailable" render), not fail the whole command.

**Should-fix — `Path()` measures the storage root, not the active database.** Embedded `Path()` is `.beads/embeddeddolt` (all databases) while `CLIDir()` is the active `<root>/<database>`; walking `Path()` includes sibling databases under a "Database size" label. The test fake returns the same value for both, and the integration test asserts only `> 0`, so overcounting is invisible. Size the active database dir, and add a sibling-database test case.

Nit: the test fake embeds all of `storage.DoltStorage` to fake one locator method — injecting the narrow sizing dependency would keep nil promoted methods out of reach.

The narrow embedded-path goal (resolve via the active locator, propagate walk errors) is correctly implemented and the claimed wiring all exists — it's the cross-mode behavior that needs the capability split.

ecuthiell:
Addressed the external-server and active-database findings in `26dbf3e0b0f8f32cfb69bc2a3fc94c28998eaaac`.

Sizing is now a narrow optional `storage.ActiveDatabaseSizer` capability captured per store instance. Embedded and provably owned/shared local stores measure only the active database directory; external, proxied, gateway, socket, TLS, remote-host, explicit-endpoint, and ambiguous instances return typed unsupported. `backup status` omits unsupported size in both human and JSON output while preserving the rest of the status, but a declared-local measurement failure remains fatal.

The permanent tests include larger sibling databases, stale local paths for external stores, decorator unwrapping, live-walk churn, and mixed-instance authority churn. The latter keeps local A and external B alive while legacy endpoint/shared/shared-plus-explicit global state changes, then rechecks both instances after each flip; this guards against reintroducing call-time global inference.

The exact head passed focused normal/process/race suites, vet, diff hygiene, lint with zero issues, and two independent contract reviews. Fresh required remote checks remain stop conditions before merge.

maphew (approved):
Approving — this supersedes my earlier changes-requested review (which predates your force-push). All three items are verified addressed at 26dbf3e0:

1. **External-server blocker**: the optional `storage.ActiveDatabaseSizer` capability with construction-time authority resolution is the right shape — external/proxied/gateway/socket/TLS/remote/env-endpoint instances return a typed `ErrUnsupported`, `bd backup status` omits size but still renders, and a declared-local measurement failure stays visibly fatal (no false zeros).
2. **Path() over-measurement**: embedded now sizes `CLIDir()`, server sizes root/database, with sibling-database scoping tests in both packages.
3. **Test fake**: `stubChainStore` implements `ActiveDatabaseSize` explicitly.

Maintainer assist: I pushed merge commit 47831a41c to this branch resolving the one conflict with main — an additive-additive collision in `internal/storage/dolt/store.go` between AD-01's `buildTestModeProductionPortPanic` (#3632) and your `resolveLocalActiveDatabaseDir`; both functions kept verbatim. Build, vet, and the focused ActiveDatabaseSize/BackupStatus/MeasureDirectorySize/sibling-scoping tests pass locally on the merged tree.

Merge tail goes to the pr-babysit patrol once the fresh checks on the merged head are green. Thanks for the careful capability design, ecuthiell — this whole batch of five was a pleasure to land.

PR #4929, opened 2026-07-20
fix(worktree): add explicit merge containment

### LINKED ISSUE #4919: bd worktree remove misclassifies a missing upstream as unpushed commits
## Summary

`bd worktree remove` refuses to remove a clean worktree when its branch has no configured upstream, reporting that the worktree has unpushed commits.

This affects short-lived local worker branches whose commits have already been integrated into a local branch but which were never intended to track or push to a remote branch.

## Environment

- beads: `1.1.0` (`8e4e59d39f34`)
- Git: `2.54.0.windows.1`
- OS: Windows 11 Pro, build 26200
- Go: not installed; reproduced with the official prebuilt Windows binary

The same implementation remains present on `main`.

## Minimal reproduction

Using stock Beads without an orchestrator:

```sh
tmp="$(mktemp -d)"

git -C "$tmp" init
git -C "$tmp" config user.name "Repro"
git -C "$tmp" config user.email "repro@example.invalid"
git -C "$tmp" commit --allow-empty -m base

git -C "$tmp" worktree add "$tmp-lane" -b repro-lane

# Empty output: the worktree is clean.
git -C "$tmp-lane" status --porcelain

# Empty output: the branch has no upstream.
git -C "$tmp-lane" for-each-ref \
  --format='%(upstream)' \
  refs/heads/repro-lane

# Exit zero: the lane HEAD is already contained in the local integration ref.
git -C "$tmp" merge-base --is-ancestor repro-lane HEAD
echo $?

cd "$tmp"
bd worktree remove "$tmp-lane"
```

## Actual behavior

```text
safety check failed: worktree has unpushed commits
Use --force to skip safety checks
```

## Expected behavior

A missing upstream should not be reported as evidence of unpushed commits.

The command should fail closed with an accurate, actionable diagnostic unless the caller supplies another explicit ref against which containment can be verified.

## Root cause

[`checkWorktreeSafety`](https://github.com/gastownhall/beads/blob/main/cmd/bd/worktree_cmd.go#L661-L683) runs:

```go
git log @{upstream}.. --oneline
```

using `CombinedOutput`, discards the exit status, and treats any resulting text as commit output:

```go
output, _ = gitCmd.CombinedOutput() // Ignore error (no upstream is ok)
if len(strings.TrimSpace(string(output))) > 0 {
    return fmt.Errorf("worktree has unpushed commits")
}
```

When no upstream is configured, Git exits nonzero and writes a diagnostic such as `fatal: no upstream configured for branch ...` to stderr. Because stdout and stderr are combined, that diagnostic is interpreted as an unpushed commit.

The same implementation is present in [v1.1.0](https://github.com/gastownhall/beads/blob/v1.1.0/cmd/bd/worktree_cmd.go#L643-L665).

## Proposed contract

Would maintainers be open to an explicit comparison option such as:

```sh
bd worktree remove <name> --merged-into <ref>
```

Suggested semantics:

- The worktree must still be clean.
- With `--merged-into <ref>`, removal proceeds only if the worktree's `HEAD` is an ancestor of `<ref>`.
- An invalid ref, ancestry-check error, or non-contained `HEAD` fails closed.
- Without the option, a configured upstream retains the existing ahead-commit check.
- Without either a configured upstream or an explicit comparison ref, the command reports that it cannot verify containment instead of reporting unpushed commits.
- `--force` retains its existing behavior.

An explicit comparison ref avoids temporarily assigning a local integration branch as the branch's upstream merely to satisfy the removal check.

If this interface—or a maintainer-preferred equivalent—fits the intended command contract, I would be happy to prepare a focused PR.

## Suggested regression coverage

1. Clean branch with a configured upstream and no ahead commits succeeds.
2. Clean branch ahead of its configured upstream is refused.
3. Clean branch with no upstream receives an accurate fail-closed diagnostic.
4. No-upstream branch contained in `--merged-into <ref>` succeeds.
5. No-upstream branch not contained in that ref is refused.
6. Invalid comparison ref is refused.
7. Dirty worktree is refused regardless of the comparison ref.

---

ecuthiell:
## What

Add `bd worktree remove --merged-into <ref>` as an explicit positive-containment guard, and harden the destructive removal transaction around it.

Containment and grammar:

- omission uses the target branch's configured upstream; no resolvable upstream fails closed with guidance to configure one or use `--merged-into`;
- an explicitly empty `--merged-into=` is rejected rather than falling back;
- explicit refs are resolved without Git DWIM ambiguity, including short-name collisions, root refs, pseudorefs, and symbolic-ref chains, then pinned to one commit OID;
- target `HEAD` must be an ancestor of that pinned comparator immediately before mutation; and
- `--force` preserves its existing bypass of cleanliness/containment while still using the identity transaction below.

Removal safety:

- resolve the exact Git worktree registration, physical directory, `.git` marker, administrative gitdir, common dir, HEAD/branch, status bytes, and lock/prune presence plus exact reasons;
- scrub hostile Git routing/config environment and revalidate the complete pinned snapshot immediately before invoking Git;
- on Windows, use `os.SameFile` when both paths exist; refuse one-existing/one-missing identity; and, when both leaves are missing, require identical ordinal unresolved suffixes anchored at stat-proven `os.SameFile` ancestors;
- invoke the final destructive Git selector with the exact registry spelling and command-scoped `core.ignorecase=false`, so a last-boundary disappearance/config flip cannot redirect it to a case-variant sibling;
- distinguish unchanged Git failure from partial/indeterminate removal state;
- remove only an exact `# bd worktree` / path pair from a stable regular `.gitignore`, reporting post-removal cleanup failure as partial success; and
- require a fail-closed native Windows process suite in `CI Gate / Required`.

## Why

Fixes #4919.

A short-lived local worktree branch may already be integrated into another local branch without ever having a remote upstream. The old check interpreted Git's missing-upstream error text as commit output and reported false "unpushed commits." `--force` is too broad when the caller can provide positive containment evidence.

Because removal is destructive, the comparator alone is not enough: concurrent target/ref/config changes, case-sensitive NTFS identities, prunable registrations, symlink/reparse replacement, force-mode byte changes, and `.gitignore` cleanup need explicit fail-closed semantics.

The previous hosted Windows run also exposed an 8.3-path boundary: the runner created fixtures under `C:\Users\RUNNER~1\...`, while Git recorded the physically identical long ancestor spelling. Four failures were raw-string oracle errors; the two-prunable case found a real production gap when both leaves were absent. The revised physical-ancestor rule and independent oracle cover both.

This remains a userspace transaction, not an atomic exclusion boundary against arbitrary concurrent writers. The final selector hardening closes the demonstrated wrong-registration redirection; other state changes in the unavoidable revalidation-to-mutation gap are handled fail-closed where Git exposes them, but cannot be ruled out absolutely.

## Verification

Rebased onto upstream `b7b6075917a40d9d52a3742bbea6de3cfdd0a319`; exact head: `fd335d52f8487ba1f81e3c0a50c9916a97b0cdca`.

Permanent real-process coverage includes:

- omitted/empty/invalid/ambiguous refs, branch/tag and pseudoref collisions, symbolic refs, comparator races, and non-containment;
- dirty, untracked, ignored, and force-mode changed-byte cases;
- target/admin/`.git` identity and symlink replacement, partial Git removal, unchanged failure, and exact `.gitignore` ownership including leading-space names;
- hostile Git environment; and
- native NTFS case-sensitive both-existing, existing/missing, one/two-prunable, 8.3/long-ancestor aliases, restoration-hook, final-selector race, and bare lock-transition cases with registry/branch/filesystem/`.gitignore` zero-mutation assertions.

Exact-head local checks passed:

```text
Git 2.55.0.windows.2, actual 8.3 TEMP ancestor:
  CGO_ENABLED=0 go test -tags gms_pure_go -count=1 -run '^TestWorktreeRemove' ./cmd/bd
  ok github.com/steveyegge/beads/cmd/bd 159.129s

Git 2.55.0.windows.2, actual 8.3 TEMP ancestor:
  go test -tags gms_pure_go -count=1 -run '^TestWorktreeRemoveWindows' ./cmd/bd
  ok github.com/steveyegge/beads/cmd/bd 22.519s

go vet -tags gms_pure_go ./cmd/bd
golangci-lint run --build-tags gms_pure_go --new-from-rev upstream/main ./cmd/bd/...
actionlint .github/workflows/pr.yml
git diff --check upstream/main...HEAD
```

The exact hosted `Worktree remove boundary (Windows)` job, `CI Gate / Required`, and maintainer re-review remain stop conditions before merge.

---


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace with an empirical git probe; Claude adjudication re-verified at the PR head). Verdict: **needs changes** — two safety edges on a safety feature.

**Should-fix 1 — pseudoref collisions validate containment against the wrong commit.** The resolver's comment claims `show-ref --verify` recognizes pseudorefs like `ORIG_HEAD`, but an empirical probe on git 2.43 shows `show-ref --verify --quiet -- ORIG_HEAD` exits 1 — while `rev-parse ORIG_HEAD^{commit}` succeeds, and `refs/ORIG_HEAD` is a legal competing ref. Git's revision rules give `$GIT_DIR/ORIG_HEAD` precedence, so when both exist the resolver sees only `refs/ORIG_HEAD`, returns it, and the containment check runs against the wrong commit — removal can pass where the caller's actual `ORIG_HEAD` would fail. The ambiguity tests cover branch/tag but not this.

**Should-fix 2 — explicitly-empty `--merged-into=` silently falls back to upstream mode.** Verified: the code selects explicit-containment only on a nonempty string and never consults `Changed("merged-into")` (zero occurrences at the PR head). So `--merged-into="$TARGET"` with an unset shell variable checks the configured upstream instead of erroring — precisely the accidental-invocation shape this flag exists to guard. The pinned pflag records explicitly-supplied empty values as `Changed()`; erroring on changed-but-empty closes it. The existing test covers omission, not `--merged-into=`.

Generated CLI docs are correctly untouched (pinned to v1.1.0). Drift is clean; the flag needs no pre-store-open handling. With those two edges closed this is a good containment guard.

ecuthiell:
Addressed both requested changes—and expanded the permanent boundary around this destructive command—in exact head `f81e728b36264c800f4c4026a38b332ef9c7230b`.

The explicit flag now rejects `--merged-into=`; ref resolution handles root refs/pseudorefs and rejects ambiguous short names before pinning a commit OID, so `ORIG_HEAD` cannot silently resolve to a competing `refs/ORIG_HEAD`.

The repeated contract review also found several target-identity cases worth fixing before asking you to re-review. Removal now pins and revalidates the Git registry plus physical target/admin/`.git` identities, status bytes, comparator, and lock/prune presence/reasons; unproven Windows paths never case-fold. Required native NTFS tests cover `Lane`/`lane` as both-existing, existing/missing, one/two-prunable, and a deterministic restoration race, plus bare lock transitions. `.gitignore` cleanup is exact-pair, stable-file, and partial-outcome aware.

The exact focused suite passed independently in 86–92 seconds, the adversarial set passed three consecutive runs, and vet/lint/actionlint/diff hygiene are green. A required hosted Windows boundary job now feeds `CI Gate / Required`; those fresh exact-head results remain mandatory before merge.

ecuthiell:
I investigated the failed hosted Windows boundary run and pushed a correction at `fd335d52f8487ba1f81e3c0a50c9916a97b0cdca`.

The original diagnosis needed one correction: Git did **not** remove the wrong `Lane`/`lane` registration in four of the five failing tests. The runner fixture lived under the short `C:\Users\RUNNER~1\...` spelling while Git recorded the long ancestor spelling, so the tests' raw full-path comparisons falsely reported preserved registrations as absent. The two-prunable case did expose a real production gap: with both leaf paths missing, raw absolute equality could not recognize identical unresolved suffixes beneath physically identical short/long ancestors.

The revision now:

- proves both-missing identity only by peeling identical ordinal suffix components to existing ancestors that `os.SameFile` proves identical;
- continues to refuse one-existing/one-missing and non-not-found stat errors;
- uses the exact registered spelling plus command-scoped `core.ignorecase=false` for the final destructive Git selector, preventing a post-revalidation disappearance/config flip from selecting a case sibling;
- replaces the raw-string test oracle with exact leaf plus physical-parent identity; and
- makes the restoration hook observable with a sentinel, while starting from both registered variants.

Independent exact-head validation passed the full `^TestWorktreeRemove` process suite under Git for Windows `2.55.0.windows.2` with an actual 8.3 TEMP ancestor (`159.129s`), plus the Windows boundary subset (`22.519s`), vet, lint, actionlint, formatting, and diff hygiene. An independent contract review also passed the named selector-race, two-prunable, restoration-hook, and ordinary `core.ignorecase=true` falsifiers.

I have updated the PR description with the precise identity rule and the remaining userspace race boundary. The new hosted Windows job, required aggregate, and maintainer re-review remain the stop conditions.

ecuthiell:
The patch-specific hosted gates are green on `fd335d52f8487ba1f81e3c0a50c9916a97b0cdca`, including [Worktree remove boundary (Windows)](https://github.com/gastownhall/beads/actions/runs/30089571435/job/89469545101) and the primary [CI Gate / Required](https://github.com/gastownhall/beads/actions/runs/30089571435/job/89472285273).

The separate PR Risk aggregate failed only because [Proxied Dolt Cmd 1/15](https://github.com/gastownhall/beads/actions/runs/30089571462/job/89471001099) hit a transient MySQL `busy buffer` while creating the fixture for `TestProxiedServerUpdateConcurrentClaim`; the partial init then surfaced the dirty-`issues` migration guard. That path is unrelated to this PR's four changed files. I attempted a failed-job-only rerun, but GitHub reports that my fork contributor account lacks repository admin permission to rerun it.

Could a maintainer please rerun the failed jobs in [PR Risk run 30089571462](https://github.com/gastownhall/beads/actions/runs/30089571462)? I have recorded the distinct proxied-fixture failure separately so it does not get lost or conflated with the closed #4566 recovery issue.

maphew (approved):
Approving — this supersedes my earlier changes-requested review (2026-07-23); both items are verified resolved at head fd335d52f8:

1. **Explicitly-empty `--merged-into=`** now errors instead of falling back to upstream: `singleWorktreeStringFlag.Set` rejects empty values and duplicate flags, covered by `TestWorktreeRemoveCobraGrammar`.
2. **ORIG_HEAD pseudoref collision** is handled fail-closed — `isWorktreeLocalPseudoref` rejects HEAD and all-caps `*_HEAD` selectors outright with guidance to use a full ref or commit OID, and short-name ambiguity across ref namespaces errors explicitly. Stricter than I asked for, which is the right call for a destructive command.

Verified by independent review + adversarial re-check (including a dedicated Git 2.43 ORIG_HEAD regression test on the branch). The one red check was a proxied-Dolt fixture flake (`TestProxiedServerUpdateConcurrentClaim`, mysql "busy buffer" — unrelated to this PR's four files); I've rerun the failed job as requested.

One real but non-blocking find to note: `addToGitignore` splits on LF without stripping a trailing CR, so a CRLF .gitignore can defeat the duplicate-entry check (usually masked by the `isIgnoredByGit` early return). I'll file it as a follow-up rather than hold this PR.

Merge tail goes to the pr-babysit patrol once checks are green (squash, attribution preserved).

PR #3710, opened 2026-05-04
perf(dolt): eliminate 12s slow path on bd commands when .local_version is stale

quad341:
## What this changes

A user running multiple `bd` binaries from different paths (e.g. a stale `~/go/bin/bd` and a current `~/.local/bin/bd`) sees `.local_version` flap between versions on every command. That triggers `autoMigrateOnVersionBump`, which opens the store **writeable**, which calls `syncCLIRemotesToSQL → migrateServerRootRemotes`, which runs `dolt remote -v` against the multi-database server root (`.beads/dolt/`). That subprocess takes **~12 seconds** to fail with "current directory is not a valid dolt repository" — because the server root has `.dolt/sql-server.info` but no `.dolt/repo_state.json`.

End-user effects this fixes:

- `gc mail inbox` going from 60–80 s back down to single-digit seconds.
- `gc status` no longer hangs past its 30 s timeout.
- `gc mail send` no longer hits the 60 s subprocess timeout repeatedly.
- `bd list` / `bd ready` / `bd show` already used a read-only store and didn't hit the slow path; this PR keeps that property and adds the same defense on the writeable hot path.

The bug was introduced by [the well-intentioned `migrateServerRootRemotes` migration](https://github.com/gastownhall/beads/issues/2118), which assumed `rootDir/.dolt` implied a real Dolt repo. On a multi-DB server install it doesn't.

## What ships in this PR

This PR ships two of the three defenses originally scoped (see maintainer-edit note below for why):

- **Layer 2 — defensive**: `ListCLIRemotes` wraps the `dolt` subprocess in a 2 s `context.WithTimeout`. Backstops against the broken-parent-dir case (and any future hung-dolt scenario) wherever `ListCLIRemotes` is still called today — `cmd/bd/doctor/federation.go`'s federation check and the CLI push/pull/fetch remote-routing helpers in `internal/storage/doltutil/remotes.go`.
- **Layer 3 — read path**: `autoMigrateOnVersionBump` does a read-only probe of `bd_version` before opening the store writeable. Read-only stores skip the writeable migration path entirely, so a stale `.local_version` no longer drags users through an unnecessary writeable open on every command.

Layer 2 and Layer 3 are independent, low-risk changes that each stand on their own merits. No new config knobs, no new endpoints, no schema changes, no breaking changes.

## Review notes

[...]

PR #4930, opened 2026-07-20
fix(readonly): prevent validation side effects

idirectships:
## What

Make strict `--readonly` validation observationally pure across root initialization, validator-owned stores, Dolt server lifecycle state, circuit-breaker files, port-file repair, and post-run maintenance.

Strict readonly now fails closed for proxied Dolt, whose current provider bypasses the direct store's readonly config. Current upstream already rejects removed PostgreSQL/MySQL/SQLite backend tombstones before maintenance or store startup; this patch preserves that architecture. Ordinary command-classified reads retain their existing maintenance and auto-start compatibility.

The patch also isolates circuit state for package tests and child `bd` processes under suite-owned temporary roots while preserving the exact production defaults.

## Why

`bd config validate --readonly` could still open the main store writable, track CLI version changes, auto-migrate metadata, start a server, clean or record circuit-breaker state, repair the resolved port file, and run post-command backup/export/push work. A read-only audit observed stale circuit state being deleted.

This is distinct from #4367, which addresses routed writable-store behavior, and it does not modify `version_tracking.go` from #3710.

## Verification

- Fresh focused strict-policy and post-run tests: pass.
- Fresh validator-owned store-options test: pass.
- Fresh circuit isolation, readonly server-open, port-file, auto-start, and writable-companion tests: pass.
- `git diff --cached --check`: pass.
- `make ci-pr-core`: changed storage package passed; two `cmd/bd` failures reproduce unchanged on the exact base.
- `make test`: changed storage package passed; the same exact-base failures and a contention timeout remain named local blockers.
- `gofmt`: pass.
- Hermetic child-process canary: compiles but SKIPs locally because Docker-backed Dolt test infrastructure is unavailable; CI must produce an explicit PASS.
- Local policy/lint scripts are blocked by missing Bash `mapfile` and missing `golangci-lint`; no local tool workaround was installed.

Separate-source exact-candidate review receipt and full local proof are recorded in the handoff packet. Producer will not self-merge.

### DIFF
--- a/cmd/bd/doctor/config_values.go
+++ b/cmd/bd/doctor/config_values.go
@@ -406,7 +406,7 @@ func checkDatabaseConfigValues(repoPath string) []string {
 
 	// Open Dolt store in read-only mode
 	ctx := context.Background()
-	store, err := dolt.NewFromConfigWithCLIOptions(ctx, beadsDir, &dolt.Config{ReadOnly: true})
+	store, err := dolt.NewFromConfigWithCLIOptions(ctx, beadsDir, configValidationStoreOptions())
 	if err != nil {
 		issues = append(issues, fmt.Sprintf("database: failed to open Dolt store: %v", err))
 		return issues
@@ -414,6 +414,10 @@ func checkDatabaseConfigValues(repoPath string) []string {
 	defer func() { _ = store.Close() }()
 
 	return checkDatabaseConfigValuesWithStore(store)
+}
+
+func configValidationStoreOptions() *dolt.Config {
+	return &dolt.Config{ReadOnly: true, DisableAutoStart: true}
 }
 
 func checkDatabaseConfigValuesWithStore(store *dolt.DoltStore) []string {
--- a/cmd/bd/doctor/config_values_test.go
+++ b/cmd/bd/doctor/config_values_test.go
@@ -7,6 +7,16 @@ import (
 	"testing"
 )
 
+func TestConfigValidationStoreOptionsAreStrictReadOnly(t *testing.T) {
+	cfg := configValidationStoreOptions()
+	if !cfg.ReadOnly {
+		t.Fatal("config validation store must be read-only")
+	}
+	if !cfg.DisableAutoStart {
+		t.Fatal("config validation store must disable auto-start")
+	}
+}
+
 func setupDoctorSharedWorktree(t *testing.T) (mainRepoDir, worktreeDir string) {
 	t.Helper()
 
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -151,6 +151,37 @@ func isReadOnlyCommand(cmdName string) bool {
 	return readOnlyCommands[cmdName]
 }
 
+type rootStorePolicy struct {
+	readOnly         bool
+	disableAutoStart bool
+	runMaintenance   bool
+}
+
+// effectiveRootStorePolicy separates strict --readonly/config policy from
+// command classification. Classified reads retain their compatibility
+// maintenance and auto-start behavior; strict readonly is mutation-free.
+func effectiveRootStorePolicy(cmdName string, strictReadonly bool) rootStorePolicy {
+	return rootStorePolicy{
+		readOnly:         strictReadonly || isReadOnlyCommand(cmdName),
+		disableAutoStart: strictReadonly,
+		runMaintenance:   !strictReadonly,
+	}
+}
+
+// backendSupportsStrictReadonly reports whether the live backend path can open
+// without provisioning or lifecycle changes. Unsupported SQL backends are
+// rejected earlier by validateConfiguredBackend; proxied Dolt remains writable-only.
+func backendSupportsStrictReadonly(cfg *configfile.Config) bool {
+	return cfg == nil || !cfg.IsDoltProxiedServerMode()
+}
+
+var (
+	runPostRunAutoCommit = maybeAutoCommit
+	runPostRunAutoBackup = maybeAutoBackup
+	runPostRunAutoExport = maybeAutoExport
+	runPostRunAutoPush   = maybeAutoPush
+)
+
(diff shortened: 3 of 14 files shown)


idirectships:
Maintainer action requested for exact head `d1c2f807356209a7550501a38bb5a0290974cce0`:

The pull request check rollup shows no check runs for this public-fork pull request. The upstream `main` workflows include unfiltered `pull_request` triggers for this Go change, and the workflow blobs are identical at the base and head. If the fork-workflow banner is present, please select **Approve and run workflows**.

The exact-head Actions query now shows five withheld runs, each completed with `conclusion=action_required`:

- [Regression Tests](https://github.com/gastownhall/beads/actions/runs/29729306207)
- [PR](https://github.com/gastownhall/beads/actions/runs/29729305372)
- [Conformance](https://github.com/gastownhall/beads/actions/runs/29729305394)
- [PR Risk](https://github.com/gastownhall/beads/actions/runs/29729305389)
- [Cross-Version Smoke Tests](https://github.com/gastownhall/beads/actions/runs/29729305449)

Please do not request or create another push solely to trigger CI; that would change the exact head already covered by the independent review receipt. After checks run, a separate maintainer code-review approval is still required. No merge is requested by this comment.

## Dispatch Receipt

- Posted and exact-run evidence appended: `2026-07-20 MST`
- Comment: https://github.com/gastownhall/beads/pull/4930#issuecomment-5020492188
- Branch/head changed: no
- Merge requested or performed: no

maphew (changes requested):
Cross-vendor paired agent review (Claude Opus + Codex gpt-5.6-sol, independent traces, adjudicated and re-verified against the PR head before posting). Verdict: **needs changes** — the intent is right and much of the wiring is clean (post-run maintenance gating, the doctor's own `{ReadOnly, DisableAutoStart}` store for `config validate --readonly`, proxied fail-closed), but there is one regression and one large coverage gap, both stemming from the same root: the PR keys several decisions off the store's `ReadOnly` flag, which means "classified read command" (`bd list/show/ready/...`), when the intent is the strict `--readonly` signal.

**Blocker 1 (both reviewers converged independently): classified read commands lose server auto-start.** `serverOpenCanAutoStart` (`internal/storage/dolt/store.go:1304`) rejects every `cfg.ReadOnly` store — but `effectiveRootStorePolicy` sets `readOnly` for all classified reads, not just strict `--readonly`. So `bd show` against a stopped managed server now errors ("start manually") instead of auto-starting. This contradicts the PR body's own compatibility claim and is pinned by an existing test the diff doesn't touch: `cmd/bd/dolt_autostart_lifecycle_integration_test.go:88-99` (stops the server, runs `bd show`, asserts the auto-started server survives). It's `cgo && integration`-gated, which is why local runs didn't surface it. The new `TestServerOpenCanAutoStartHonorsReadOnly` encodes the wrong invariant — `{ReadOnly:true, AutoStart:true}` is also an ordinary `bd list`. The machinery for the correct distinction already exists and is strict-only: `policy.disableAutoStart` → `ApplyCLIAutoStart` clears `cfg.AutoStart`. Fix: drop the `!cfg.ReadOnly` clause (or gate on `!cfg.DisableAutoStart`), and re-point the unit test at `DisableAutoStart`. The same conflation, lower blast radius, applies to `persistResolvedPortFile` and `initializeServerCircuitBreaker` (classified reads lose port-file repair and fast-fail breaker protection; the `BEADS_TEST_MODE` short-circuit also disables the breaker more broadly than the state-isolation goal needs).

**Blocker 2 (maintainer-verified at the PR head): embedded strict `--readonly` still opens a writable store and runs migrations.** `newDoltStore` (`cmd/bd/store_factory.go:57-61`) routes `cfg.ReadOnly` in embedded mode to `embeddeddolt.OpenForReadOnlyCommand`, whose own doc comment says the returned store "is otherwise a normal writable store" — its open path creates the data dir, initializes schema, and calls `MigrateUp`. The strict write-refusing behavior this PR adds exists only on the server-mode path, and the regression canary only exercises server mode — so `bd config validate --readonly` on the default embedded backend still performs the class of side effects the PR exists to prevent.

**Should-fix — pre-policy writers (single-reviewer trace, spot-checked plausible; please confirm):** several writes run before the readonly policy is even constructed in root pre-run: `configfile.Load`'s legacy `.beads/config.json` → `metadata.json` migration (writes new file, deletes old), telemetry/metrics file initialization under HOME, and the molecule template auto-import (`Loader.LoadAll` inserts missing templates, not gated on strict policy). The test canary seeds modern metadata and disables metrics, which masks these. Worth either gating them on the strict signal or explicitly documenting them as out of scope for `--readonly`.

**Scope clarification worth stating in the PR body:** command bodies are not centrally write-protected (`bd config set`-style YAML/git writes, audit sidecar appends), so `--readonly` is a store-open policy, not a general write barrier. Fine if intentional — say so, since the flag name promises more.

The direction is good and the fix path is small: key every new decision off the strict signal (`DisableAutoStart` / a threaded strict flag), give embedded mode a genuinely write-refusing strict open, and extend the canary to embedded mode. With those, this lands.

maphew (approved):
Approving — supersedes my 2026-07-23 changes-requested review; both blockers are fixed as maintainer commits on your (rebased) branch. Auto-start/circuit-breaker/port-file gating is re-keyed from `cfg.ReadOnly` to `cfg.DisableAutoStart` (strict-only signal), and strict `--readonly` now routes to the genuinely write-refusing `embeddeddolt.OpenReadOnly` — which already existed — while ordinary classified reads keep their writable-open semantics. The Codex gate additionally caught that `Close()` could still delete leaked `tmp_pack_*` files under strict readonly (now early-returns) and that the hermeticity canary's own fixture was missing the now-required `federation.remote` (fixed; the canary passes). Your observational-purity framing was the right call throughout.

Validation: policy/factory/embedded tests green by name, revert-sensitivity confirmed; cross-vendor gate signed off. Merge tail to the pr-babysit patrol. Thanks idirectships.

PR #4933, opened 2026-07-20
fix(doctor): bound dependency-cycle check queries, match bd dep cycles semantics

swedeinasia-flow:
## What

Rewrites the `bd doctor` "Dependency Cycles" check to load the dependency graph with small keyset-paginated queries and detect cycles client-side (iterative Tarjan SCC), instead of one unbounded `WITH RECURSIVE` path-enumeration query. It also restricts traversal to blocking edge types (`blocks`, `conditional-blocks`), matching `bd dep cycles` / `issueops.DetectCyclesInTx` / the add-time cycle guard.

## Why

Two defects in the current check:

1. **It can never finish on real-world stores served by a shared dolt sql-server.** The recursive CTE enumerates every *simple path* in the graph (the `path NOT LIKE …` dedup makes it exponential on dense graphs). Dolt server configs commonly set `read_timeout_millis` to reap slow per-call connections; any single query over that limit is killed mid-stream and the check permanently reports `Unable to check for cycles: invalid connection`. Observed on a store with ~2.4k issues / 3.5k dependency rows behind a server with `read_timeout_millis: 15000` — raising the server timeout is not an option when it's deliberately low to keep dead connections from piling up. Paged reads (1000 rows/query, keyed on the deterministic `dependencies.id` PK from migration 0050) keep every query bounded; the whole check now completes in a few hundred milliseconds on that same store.

2. **Its semantics disagree with the rest of bd.** The recursive SQL traverses *every* dependency type, while `bd dep cycles`, `DetectCyclesInTx`, and cycle prevention on `bd dep add` all restrict to blocking types because non-blocking links (`tracks`, `related`, …) legitimately form loops ("Non-blocking edge types cannot form blocking cycles"). On the store above, the moment the query could complete it reported **1753 "cycles"** — all through `tracks` edges — while its own fix hint ("Run 'bd dep cycles'…") then reports none. The check now filters to blocking types, client-side, exactly like `AppendBlockingGraphInTx`.

User-visible messages, statuses, and the flagged node set (nodes participating in ≥1 blocking cycle) are unchanged; the node set is now sorted, so `First cycle involves:` is deterministic. A `dependencyCycleMaxEdges` guard (1M edges) degrades to a warning instead of unbounded memory growth.

## Verification

```bash
go test -tags gms_pure_go -run 'TestDependencyCycleNodes|TestCheckDependencyCycles' ./cmd/bd/doctor/
golangci-lint run --build-tags gms_pure_go ./cmd/bd/doctor/   # 0 issues
```

New table-driven tests cover DAGs, self-edges, disjoint cycles, cycles behind DAG prefixes, and a 200k-node cycle that guards the iterative traversal against stack-depth limits a recursive DFS would hit.

Also verified live against the affected store: `bd doctor --readonly` previously failed with `invalid connection` on every run; with this change it reports `No circular dependencies detected`, agreeing with `bd dep cycles` on the same store, and passes on healthy stores unchanged.


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the bounded-pagination idea is right; four correctness gaps in the new loader.

**Should-fix 1 — wisp cycles are invisible.** Verified at the PR head: `loadDependencyEdgePage` selects `FROM dependencies` only. The reference implementation (`issueops.DetectCyclesInTx`) and the store contract cover `dependencies` UNION `wisp_dependencies` — and the same package already has a union helper (`doctor/dependency_sql.go`). Doctor can report "No circular dependencies detected" for wisp-only or cross-table cycles that `bd dep cycles` would find, contradicting the PR's match-semantics goal.

**Should-fix 2 — pages don't share a snapshot.** Each page runs on the raw `*sql.DB` (implicit per-statement transactions in the pinned engine), so concurrent writes on a shared server can skip inserted edges or retain deleted ones across pages — false negatives or false cycles. The repo's established multi-query read pattern wraps one `BeginTx` around the whole read.

**Should-fix 3 — `dependencies.id` isn't guaranteed on doctor's path.** Post-branch drift: #4878 added `ensureDependenciesIDColumn` precisely because supported databases exist where the column never materialized, and doctor skips normal root store init, auto-migrates only on version bump, and opens read-only (schema init skipped). On that shape, the new `WHERE id > ?` pagination degrades the check to "Unable to check for cycles" where main's repair path would have it work.

**Should-fix 4 — a scan error silently truncates the graph.** A failed `rows.Scan` is ignored; since termination is `rowsRead < pageSize`, one malformed row ends pagination early and silently drops every later edge. The reference loader returns scan errors.

Also: the new million-edge cap introduces a warning status the commit message says doesn't exist — fine safeguard, but state and test it; and the tests exercise only the in-memory SCC helper, never multi-page loading, either table, or the error paths — which is how all of the above got through.

swedeinasia-flow:
Thanks for the thorough cross-vendor review — all four should-fixes were real. Addressed in 135428cb (amended in place since the original commit message itself needed correcting; branch also rebased onto current main so the change is reviewed against the post-#4878 reality).

**1. Wisp cycles** — the loader now traverses both `dependencies` and `wisp_dependencies` (`dependencyCycleTables`), matching `issueops.DetectCyclesInTx`. New tests cover a wisp-only cycle and a cross-table cycle (half in each table), both of which the previous head missed.

**2. Shared snapshot** — all pages of both tables now run inside one `BeginTx`, per the repo's established multi-query read pattern. The sqlmock tests use ordered expectations, so any query issued outside that transaction fails them.

**3. `dependencies.id` on doctor's path** — the loader probes each table (information_schema column check + a NULL-id probe) and falls back to the reference loader's single streaming scan when id keyset pagination would be unsound; the exponential recursive query was the problem, not the plain edge scan `bd dep cycles` itself performs. The probe also covers a wrinkle one step past the missing-column shape: `ensureDependenciesIDColumn`'s backfill means rows can be NULL mid-backfill with the column present, and NULL never satisfies `id > ?` — that shape would not even degrade to "Unable to check", it would silently drop edges. Both shapes are tested end-to-end against the package's Dolt harness (`DROP PRIMARY KEY` + `MODIFY ... NULL`, and `DROP COLUMN id`), asserting the cycle is still reported.

**4. Scan errors** — a failed `rows.Scan` now fails the check with the wrapped error (`Unable to check for cycles` / `scan <table>: ...`), pinned by a sqlmock test that injects a bad row mid-page.

On the extras:
- The million-edge cap is now stated in the commit message as a behavior change, enforced inside the row loop (so the unpaginated fallback scan can't exceed memory before a post-hoc check), and tested (`more than N edges`).
- New coverage beyond the SCC helper: multi-page loading at page sizes 1/2/1000 asserting the exact adjacency map across both tables, non-blocking loops ignored in both tables, and query-error propagation — e.g. a missing `wisp_dependencies` table on an unrepaired clone surfaces as the warning, the same way `bd dep cycles` errors on that shape, never as a silent "no cycles".

Local verification: no Docker on this machine, so the committed container-harness tests skip locally — I ran the identical scenarios against a manually started `dolt sql-server` and all pass, including the degraded-schema ALTERs. The sqlmock and pure-Go SCC tests pass everywhere. CI should exercise the container path.

maphew (approved):
Approving — this supersedes my 2026-07-23 changes-requested review. All four should-fixes are verified present and correct at the current head (independent review plus a cross-vendor verification pass): wisp_dependencies edges included in the cycle scan, pagination wrapped in a single transaction/snapshot, no reliance on `dependencies.id` existing on the pre-migration read-only doctor path, and Scan failures surfaced instead of swallowed.

Apologies that CI never ran until now — first-contribution workflow runs need maintainer approval and that slipped past us; I've approved them, so checks are running against your head. Merge tail goes to the pr-babysit patrol once green.

Thanks swedeinasia-flow — responsive, careful turnaround on all four items.

PR #4939, opened 2026-07-21
fix: reject substring matches in ID resolution, require exact hash match only

joshuaguyervs:
## Summary
- \`bd close\`/\`bd show\` (and the shared ID-resolution helper they call) silently substring-matched a nonexistent ID to an unrelated existing bead instead of erroring "not found" — could silently close/mutate the wrong issue.
- Removed substring matching from \`ResolvePartialID()\` in \`internal/utils/id_parser.go\`. Now requires exact hash match only.

## Changes
- Removed substring match collection (\`strings.Contains(issueHash, hashPart)\`)
- Removed substring match in wisp fallback search
- Added test case verifying substring matches are rejected

## Test plan
- [x] Full test suite passing, incl. new case


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — the interior-substring hazard is real, but the cure removes a documented feature.

**Blocker — leading partial-hash resolution is removed wholesale, not just interior substrings.** Verified at the PR head: only the `issueHash == hashPart` exact comparison remains and the `matches` slice is never populated, so any non-exact input falls through to "no issue found." But leading-prefix abbreviation is documented, promised in code comments, and pinned by tests: `docs/core-concepts/hash-ids.md` ("`bd show a1b2`" resolves the full hash), the `id_parser.go` header contract (`a3f8 → bd-a3f8e9`), `cmd/bd/cli_fast_test.go` (mutates through a 4-char truncated hash), and the oracle scenario corpus (`zzqxk → sh-zzqxk1`). This reaches `show`, `close`, and every mutation command routed through `ResolvePartialID`. The right shape is to keep **prefix** matches (`strings.HasPrefix(issueHash, hashPart)`) with the existing ambiguity error, and reject only non-prefix interior substrings — that closes the reported hazard without breaking the documented UX.

**Should-fix — an existing wisp test fails under the new logic.** For input `t3st`, `bd-wisp-t3st` reduces to `wisp-t3st`, the exact comparison fails, and the existing `id_parser_test.go` wisp-resolution expectation breaks (that suite provisions a real server — it's exercised, not dead).

**Should-fix — dead machinery left behind.** With `matches` permanently empty, the sorting/ambiguity/single-match paths are unreachable, the comments still describe fallback behavior that no longer exists, and the always-true `len(matches)==0` triggers a redundant wisp search. If exact-only were the intent, all of that should go; with prefix-matching restored, most of it comes back to life instead.

Nit: the new regression fixture only covers a durable issue; add the ephemeral/wisp negative case alongside it.

Merge-base drift is clean. The bug (interior substring matching surprising users) deserves fixing — restore prefix semantics and this lands.

maphew (approved):
Approving — supersedes my 2026-07-23 changes-requested review; the fixes are applied as maintainer commits. Interior-substring matches are rejected (your #4234 fix, kept) but leading-prefix resolution is restored via `strings.HasPrefix` — the outright deletion had also killed the documented abbreviation UX and left the ambiguity machinery dead. The wisp fallback strips the `wisp-` infix before comparing rather than a mechanical swap. Your exact-match tests ship unchanged, joined by prefix-resolution and wisp-prefix tests.

Validation: build/vet/gofmt clean, full ResolvePartialID suite green; independent review + cross-vendor Codex gate signed off. Merge tail to the pr-babysit patrol. Thanks joshuaguyervs.

PR #4941, opened 2026-07-21
feat: add public schema package re-exporting the migration API

eric-richardson1:
Adds a public schema package                                                                                                                                                                 
                                                                                                                                                                                              
Exposes bd's schema migration API to external Go modules. The migration engine lives under internal/storage/schema, so a separate module can't import it. This adds a thin wrapper at github.com/steveyegge/beads/schema re-exporting the minimal surface needed to create or upgrade a beads database over a standard database/sql connection


coffeegoddd (approved):
LGTM

PR #4949, opened 2026-07-22
fix(make): bootstrap Git Bash on Windows

### LINKED ISSUE #4948: Native GNU Make cannot reliably bootstrap Git Bash on Windows
## Summary

The repository Makefile cannot reliably bootstrap its POSIX shell when it is run by native GNU Make on Windows with a standard Git for Windows installation under `C:\Program Files`.

The current discovery command runs before `SHELL` has been overridden:

```make
GIT_BASH := $(shell where git 2>/dev/null)
```

Native Windows Make initially evaluates that command with Windows shell semantics. `cmd.exe` treats `/dev/null` as a path rather than a null device, discovery returns no usable Bash path, and Make falls back to Windows command execution for POSIX recipes. GNU Make can also launch simple commands and shebang interpreters directly, so selecting Bash alone is insufficient when Git's `usr/bin` is not already on the workstation `PATH`.

## Environment

- Beads: `main` at `607df586d198400e2a65681b824b7d6e21a14d97`
- Windows 11 Pro, build 26200
- Git for Windows 2.54.0, installed at `C:\Program Files\Git`
- GNU Make 4.4.1 for Windows from the EzWinPorts distribution
- The Make archive SHA-256 was `fb66a02b530f7466f6222ce53c0b602c5288e601547a034e4156a512dd895ee7`
- Git's `cmd` directory was on `PATH`; its `bin` and `usr/bin` directories were not

## Reproduction

From a clean checkout, run native Windows Make against a target which proves Bash and Git's POSIX tools are active:

```powershell
make.exe --no-print-directory --eval 'windows-shell-smoke:; @test -n "$$BASH_VERSION" && test "$$(command -v sed)" = /usr/bin/sed && test "$$(command -v env)" = /usr/bin/env && echo "Windows Make shell smoke passed"' windows-shell-smoke
```

Current `main` exits 2 with:

```text
The system cannot find the path specified.
process_begin: CreateProcess(NULL, sed -n "s/^go //p" go.mod, ...) failed.
Makefile:39: pipe: Bad file descriptor
'test' is not recognized as an internal or external command,
operable program or batch file.
make: *** [<builtin>: windows-shell-smoke] Error 1
```

`make help` likewise prints its recipe through `cmd.exe` and emits the parse-time `sed` failure.

## Expected result

Native GNU Make should run repository contribution targets through Git for Windows Bash without requiring contributors to add Git's POSIX directories to their global `PATH`. Missing Git or Bash should fail during Makefile parsing with a specific diagnostic.

## Proposed fix

Within the existing `Windows_NT` branch:

1. Bootstrap explicitly with `cmd.exe`.
2. Ask `git.exe --exec-path` for one installation path, ignoring any ambient `GIT_EXEC_PATH` override.
3. Canonicalize the Git installation root with `cd /d`, verify `bin/bash.exe`, and assign that raw canonical path to `SHELL`.
4. Prepend Git's `usr/bin` to the Make environment's `PATH`, because GNU Make may directly launch commands such as `sed` or the `env` interpreter from a script shebang.
5. Fail immediately with clear errors when Git for Windows or Bash cannot be resolved.

The change is contained entirely inside the Windows conditional; non-Windows Make behavior is unchanged.

With the proposed patch, both `make help` and the reproduction above exit 0, and the smoke prints:

```text
Windows Make shell smoke passed
```

---

ecuthiell:
## What

Make the Windows GNU Make bootstrap host-aware and deterministic:

- enter the `cmd.exe` → Git-for-Windows bootstrap only for native GNU Make `MAKE_HOST` identities;
- clear hostile Bash startup options and `GIT_EXEC_PATH` before resolving one Git-for-Windows installation;
- validate Bash, `sed`, and `env`, then expose Git's POSIX tools for direct process lookup;
- leave MSYS2 and Cygwin Make on their existing POSIX shell and PATH semantics; and
- require identity-checked native, MSYS2, and genuine Cygwin smoke lanes in the PR aggregate.

Non-Windows Make behavior is unchanged.

## Why

Fixes #4948.

The original native-Windows bootstrap evaluated POSIX redirection before Bash was active and could not reliably handle a normal Git installation under `C:\Program Files`. The first patch fixed that path but applied the native `cmd.exe` transition to every `OS=Windows_NT` Make host, which broke MSYS2/Cygwin during parsing. The revised patch uses GNU Make's actual host identity instead of the outer OS value as a proxy.

## Verification

Rebased onto upstream `b9b437894272f6fb1c0a0d0422962b0a6050b6d2`; revised head: `d45a7abb7730f624bf89b36ba3acb075e473cce5`.

Local checks passed:

- native GNU Make 4.4.1 with Git under `C:\Program Files`, restricted PATH, spaced profile, and poisoned `BASH_ENV`, `BASHOPTS`, `SHELLOPTS`, and `GIT_EXEC_PATH`;
- genuine MSYS2-hosted GNU Make with its POSIX PATH and startup/routing variables preserved;
- `make help`, direct Bash/`sed`/`env` lookup, a real recipe, `actionlint` v1.7.12, and `git diff --check`.

The first hosted run passed native and MSYS2 and exposed a Cygwin-only harness transport defect before Make ran: PowerShell stdin delivered CRLF, so Cygwin parsed `pipefail\r`. The exact head avoids stdin translation by LF-normalizing the script and passing it as one `bash -c` argument; the full multiline/quoted/heredoc smoke was rechecked through PowerShell into MSYS2's Cygwin-derived Bash.

The required `Windows Make shell` matrix supplies the permanent hosted native, MSYS2, and genuine Cygwin evidence. A fresh exact-head matrix and aggregate remain required before merge.

---

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -534,6 +534,188 @@ jobs:
           version: latest
           args: --timeout=5m --build-tags=gms_pure_go
 
+  windows-make-shell:
+    name: Windows Make shell (${{ matrix.host }})
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        host:
+          - native
+          - msys2
+          - cygwin
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v6
+
+      - name: Install native GNU Make
+        if: matrix.host == 'native'
+        shell: pwsh
+        run: choco install make --version=4.4.1 --yes --no-progress
+
+      - name: Set up MSYS2 GNU Make
+        if: matrix.host == 'msys2'
+        uses: msys2/setup-msys2@66cd2cce69caa17b53920067426061ca1de3a884 # v2.32.0
+        with:
+          msystem: MSYS
+          install: git make
+          path-type: minimal
+
+      - name: Set up Cygwin GNU Make
+        if: matrix.host == 'cygwin'
+        id: cygwin
+        uses: cygwin/cygwin-install-action@3f0a3f9f988f7e96b8c18098ae05eaec175f5b52 # v6
+        with:
+          packages: git make
+          add-to-path: 'false'
+
+      - name: Exercise native Windows Make
+        if: matrix.host == 'native'
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = 'Stop'
+          $make = (Get-Command make.exe -ErrorAction Stop).Source
+          $hostOutput = & $make --no-print-directory -f NUL --eval '$(info MAKE_HOST=$(MAKE_HOST))' --eval 'noop:;' noop
+          $makeHost = (($hostOutput | Where-Object { $_ -like 'MAKE_HOST=*' }) -replace '^MAKE_HOST=', '').Trim()
+          if ($LASTEXITCODE -ne 0 -or ($makeHost -ne 'Windows32' -and $makeHost -notmatch '-mingw32$')) {
+            throw "Expected native Windows GNU Make, found '$makeHost' at '$make'"
+          }
+
+          $git = (Get-Command git.exe -ErrorAction Stop).Source
+          if ($git -notmatch ' ') {
+            throw "Expected the hosted Git for Windows path to contain a space, found '$git'"
+          }
+
+          $profile = Join-Path $env:RUNNER_TEMP 'profile with spaces'
+          New-Item -ItemType Directory -Force -Path $profile | Out-Null
+          $poison = Join-Path $profile 'poison bash env.sh'
+          Set-Content -LiteralPath $poison -Encoding utf8 -Value 'exit 97'
+
+          $env:USERPROFILE = $profile
+          $env:BASH_ENV = $poison
+          $env:BASHOPTS = 'failglob'
+          $env:SHELLOPTS = 'nounset'
+          $env:GIT_EXEC_PATH = Join-Path $profile 'missing git exec path'
+          $env:PATH = "$(Split-Path $git);$env:SystemRoot\System32"
+
+          & $make --no-print-directory help | Out-Null
+          if ($LASTEXITCODE -ne 0) {
+            throw 'Native Make failed while parsing or running help'
+          }
+
+          @'
+          include Makefile
+          .RECIPEPREFIX := >
+          .PHONY: windows-make-shell-smoke
+          windows-make-shell-smoke:
+          >@case "$(SHELL)" in */bin/bash.exe) ;; *) exit 1;; esac
+          >@test -n "$$BASH_VERSION"
+          >@test "$$(command -v sed)" = /usr/bin/sed
+          >@test "$$(command -v env)" = /usr/bin/env
+          >@test -z "$${BASH_ENV+x}"
+          >@case "$$-" in *u*) exit 1;; esac
+          >@shopt -q failglob && exit 1 || :
+          >@test -z "$${GIT_EXEC_PATH+x}"
+          >@env printf '%s\n' 'native Windows Make shell smoke passed'
+          '@ | & $make --no-print-directory -f - windows-make-shell-smoke
+          if ($LASTEXITCODE -ne 0) {
+            throw 'Native Make shell smoke failed'
+          }
+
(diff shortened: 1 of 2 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claim at the PR head). Verdict: **needs changes** — the native-Windows bootstrap looks right, but it breaks the other Windows Make population.

**Should-fix — POSIX-hosted GNU Make on Windows (MSYS2/Cygwin) breaks at parse time.** The `ifeq ($(OS),Windows_NT)` branch unconditionally sets `SHELL := cmd.exe` and immediately evaluates `$(shell set GIT_EXEC_PATH=&& git.exe --exec-path)` (verified at the PR head). MSYS2/Cygwin Make passes commands with `-c`, which `cmd.exe` doesn't accept (`/c`), so the `$(shell ...)` fails while the Makefile is still being parsed — before any target runs. `OS=Windows_NT` is set in both environments, so the condition can't distinguish native-MinGW Make from POSIX-hosted Make; the repo's own `install.sh` explicitly recognizes MINGW/MSYS/CYGWIN. Gate the `cmd.exe` bootstrap (and the semicolon-delimited `PATH`) on native Windows Make — e.g. also test `$(SHELL)`/`MSYSTEM` — and keep the working POSIX path otherwise.

**Should-fix — no PR-time coverage.** The existing Windows job runs only on pushes to main and calls `go build` directly, so PR CI never exercises `make`, the `cmd.exe → Git Bash` transition, space-bearing paths, or the exported PATH. A minimal PR-time Windows job invoking one Make target (ideally under both native-MinGW and MSYS2 Make) would pin this.

The intended native path itself checks out (Git-for-Windows layout with `mingw64/libexec/git-core`, `bin/bash.exe`, `usr/bin` tools). Merge-base drift is clean — the Makefile hasn't changed since the merge base.

ecuthiell:
Addressed the change request in `62a19f1cec7bfac28576b5c1de6cbe5855186e7e`.

The native bootstrap is now gated by `MAKE_HOST`, so MSYS2 and Cygwin retain their POSIX shell/PATH behavior. The revision also neutralizes inherited Bash startup options, and a required three-host Windows matrix exercises native GNU Make, MSYS2 Make, and a genuine Cygwin Make runtime. Local native and genuine MSYS2 smokes, `actionlint`, and diff checks pass; the new hosted matrix is the remaining pre-merge proof.

ecuthiell:
The first required hosted matrix did its job: native and MSYS2 passed, while Cygwin failed before Make because the PowerShell here-string reached Cygwin Bash over stdin with CRLF (`set: pipefail\r: invalid option name`).

That contribution-owned harness defect is fixed in `d45a7abb7730f624bf89b36ba3acb075e473cce5`. The Cygwin step now LF-normalizes the complete script and passes it as one `bash -c` argument, avoiding stdin newline conversion while preserving nonzero-exit detection. I re-ran `actionlint`, diff hygiene, and the full multiline/quoted/heredoc transport through PowerShell into MSYS2's Cygwin-derived Bash; an independent exact-head review also passed.

A fresh native/MSYS2/Cygwin matrix and `CI Gate / Required` remain stop conditions before merge.

maphew (approved):
Approving — this supersedes my earlier changes-requested review (at superseded head 7a2d0d4e9). Both demands are verified resolved at head d45a7abb7:

1. **MSYS2/Cygwin parse-time breakage**: the native cmd.exe bootstrap is now gated on `$(filter Windows32 mingw32 %-mingw32,$(MAKE_HOST))`, so msys/cygwin Make hosts keep their POSIX shell and PATH untouched — and the new CI lanes assert exactly that preserved behavior.
2. **PR-time Windows Make coverage**: the required 3-lane `windows-make-shell` matrix (native/msys2/cygwin) runs host-identity checks under poisoned BASH_ENV/BASHOPTS/SHELLOPTS/GIT_EXEC_PATH and spaced paths, wired into the CI gate.

Verified by independent review + adversarial re-check: all 73 checks pass at head including all three Windows lanes; zero commits on main touch the Makefile or pr.yml since the merge-base, so no conflict or drift risk. Non-blocking note for later: the matrix runs on every PR — a paths filter could trim runner cost if it ever matters.

Merge tail goes to the pr-babysit patrol (squash, attribution preserved).

PR #4950, opened 2026-07-22
test(windows): isolate multi-ID update subprocess home

maphew:
## What

- Pin both `HOME` and `USERPROFILE` for multi-ID update subprocess tests.
- Initialize each test directory as a Git repository with repo-local hooks before running `bd init`.

## Why

The multi-ID update contract tests added in #4726 were not hermetic on Windows. `t.TempDir()` lives under the real user profile, and without a temporary Git boundary Beads repository discovery could walk upward and find the user's global `.beads` directory. The test then aborted after reporting the existing database instead of exercising the intended isolated workspace.

This keeps the tests inside their temporary directory on every platform and prevents them from consulting real user state.

## Validation

- `go test -tags gms_pure_go ./cmd/bd -run '^TestMultiIDUpdateAllGoodPathUnchanged$' -count=1`
- `go test -tags gms_pure_go ./cmd/bd -run '^TestMultiIDUpdate' -count=1`
- `git diff --check`

### DIFF
--- a/cmd/bd/update_multi_id_exit_test.go
+++ b/cmd/bd/update_multi_id_exit_test.go
@@ -28,8 +28,8 @@ import (
 )
 
 // multiIDUpdateEnv returns a hermetic environment for bd subprocess runs:
-// no inherited BEADS_* variables, HOME pinned to the test dir, metrics and
-// daemons disabled.
+// no inherited BEADS_* variables, HOME and USERPROFILE pinned to the test dir,
+// metrics and daemons disabled.
 func multiIDUpdateEnv(dir string) []string {
 	var env []string
 	for _, e := range os.Environ() {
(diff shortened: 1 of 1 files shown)


maphew (commented):

Test-only hardening that pins USERPROFILE (not just HOME) for the multi-ID update subprocess env and initializes each test dir as a git repo with repo-local hooks before bd init. Both changes are correct and directly address the stated non-hermeticity on Windows: FindBeadsDir's non-worktree walk-up boundary is the git root, so git init stops discovery from reaching the user's global ~/.beads, and bd resolves the user home via os.UserHomeDir() (USERPROFILE on Windows), which the original HOME-only pin missed. The git init + core.hooksPath idiom matches existing init_safety_test.go usage, and scope is confined to the two hunks. No blocker/major/minor findings.

PR #4951, opened 2026-07-22
fix(search): validate and combine status filters

### LINKED ISSUE #4834: `bd search --status` silently ignores comma-separated values (and invalid statuses) that `bd list --status` accepts
## Summary

`bd list --status` accepts a comma-separated list of statuses (`--status open,in_progress`) and this is documented in its help text. `bd search --status` does **not** — it silently returns zero results for the comma-separated form (and for any unrecognized status value), with exit code 0 and no error. This is a surprising inconsistency between two sibling filter commands, and the silent-empty behavior makes it easy to misdiagnose as "no matching issues."

## Reproduction

Given a store with issues titled/containing the word `prod` across multiple statuses:

```
$ bd search "prod" --status open --json | jq length
3
$ bd search "prod" --status in_progress --json | jq length
2
$ bd search "prod" --status open,in_progress --json | jq length
0        # ← expected 5; silently returns empty
$ echo $?
0        # ← no error signalled
```

For comparison, `bd list` handles the same input correctly:

```
$ bd list --status open,in_progress --json | jq length   # works as expected
```

Any unrecognized status value shows the same silent-empty behavior:

```
$ bd search "prod" --status bogusvalue --json
[]       # exit 0, no validation error
```

## Expected

One of:

1. **Preferred:** `bd search --status` accepts comma-separated statuses identically to `bd list --status` (consistent filter semantics across commands).
2. **At minimum:** an unparseable/unrecognized `--status` value produces a validation error (non-zero exit + message) rather than silently returning `[]`.

## Actual

- `--status open,in_progress` on `search` returns `[]`
- Any invalid status returns `[]`
- Exit code is 0; nothing is logged
- The comma-separated form is documented and works for `bd list`, so the divergence is unexpected

## Why this matters

Tooling that filters search results by status has to work around the gap by fetching a broad result set and filtering client-side, which defeats the point of a server-side filter. The silent-empty behavior (rather than an error) also masks the problem — a typo or a `list`-style comma argument looks like "no results found."

## Environment

- **bd version:** 1.1.0
- **Backend:** embedded Dolt

Happy to send a PR if the maintainers agree on the preferred direction (comma-separated support vs. validation error).

maphew:
## Why

`bd search --status` treated a comma-separated value as one unknown status and silently returned no matches. It also accepted arbitrary invalid values with exit code 0, unlike `bd list --status`. That inconsistency makes typos look like legitimate empty searches and forces callers to filter results themselves.

## What changed

- share `bd list`'s validated status parser with direct and proxied search
- support comma-separated OR filters, including configured custom statuses
- return a useful error for invalid status values
- document comma-separated search statuses
- cover parser semantics and embedded CLI behavior

## Validation

- `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^(TestEmbeddedSearch|TestApplyStatusFilter|TestListCommandSuite|TestListQueryCapabilitiesSuite)$' -count=1`
- `git diff --check`

Fixes #4834

### DIFF
--- a/cmd/bd/list_filter.go
+++ b/cmd/bd/list_filter.go
@@ -128,22 +128,8 @@ func buildListFilter(in listInput, cfg listFilterConfig) (types.IssueFilter, err
 		s := types.StatusOpen
 		filter.Status = &s
 	} else if in.status != "" && in.status != "all" {
-		names := cfg.customStatusNames()
-		statusParts := strings.Split(in.status, ",")
-		if len(statusParts) == 1 {
-			s := types.Status(strings.TrimSpace(statusParts[0]))
-			if !s.IsValidWithCustom(names) {
-				return filter, fmt.Errorf("invalid status %q (valid: %s)", in.status, validStatusList(names))
-			}
-			filter.Status = &s
-		} else {
-			for _, part := range statusParts {
-				s := types.Status(strings.TrimSpace(part))
-				if !s.IsValidWithCustom(names) {
-					return filter, fmt.Errorf("invalid status %q in multi-status filter (valid: %s)", strings.TrimSpace(part), validStatusList(names))
-				}
-				filter.Statuses = append(filter.Statuses, s)
-			}
+		if err := applyStatusFilter(&filter, in.status, cfg.customStatusNames()); err != nil {
+			return filter, err
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


maphew (commented):

The change correctly fixes `bd search --status` by sharing `bd list`'s validated, comma-aware status parser (`applyStatusFilter`) across the direct and proxied search paths. Logic is sound: the proxied early-branch simplification to `if status == ""` is exactly equivalent to the old two-branch form, the OR filter genuinely reaches the SQL layer (`sqlbuild/filter.go` emits `status IN (...)`), and invalid statuses now error instead of returning silent empty results. Tests (parser unit test plus embedded CLI subtests) are present and meaningful. No blocker or major issues; scope is tight with no smuggled changes.

**Findings:**
- **minor** `cmd/bd/search_embedded_test.go:29` — The PR body's stated validation command runs with CGO_ENABLED=0, but this file (like all *_embedded_test.go) is gated `//go:build cgo`. Under CGO_ENABLED=0 the entire embedded suite is excluded from compilation, so `-run '^(TestEmbeddedSearch|...)$'` matches nothing for the embedded name and exits 0 without running the two new subtests (search_status_comma_separated, search_status_invalid). The tests are correctly written and will run under a cgo build (CI), but the quoted local validation did not actually exercise them. Re-run with CGO_ENABLED=1 to confirm the embedded coverage, or correct the command in the PR body.
- **nit** `cmd/bd/search_proxied_server.go:191` — In the proxied path the status is validated only after `openProxiedListUOW` succeeds, so an invalid `--status` value pays the cost of opening the unit of work before erroring. This is unavoidable given config must be loaded from the uow, and matches the list command's pattern, so it is acceptable; noting only for completeness.

PR #4952, opened 2026-07-22
fix(config): read shared server PID for drift

### LINKED ISSUE #4836: config drift reads the project PID instead of the shared-server PID
## Description

`bd config drift` reports that no server is running when `dolt.shared-server=true`, even while the managed shared server is healthy and `bd dolt status` reports it as running.

## Environment

- `bd` 1.1.0
- `dolt` 2.1.10
- macOS arm64

## Reproduction

1. Configure a project with `dolt.shared-server: true` and `dolt_mode: server`.
2. Run `bd dolt start`.
3. Confirm `bd dolt status --json` reports `running: true`, port 3308, and data under `~/.beads/shared-server/dolt`.
4. Confirm the Dolt process is listening on `127.0.0.1:3308`.
5. Run `bd config drift`.

Observed:

```text
✗ server: dolt.shared-server is enabled but no server appears to be running
    expected: running
    actual:   not running
```

Expected: the server drift check resolves the shared-server PID location and reports the running shared server.

## Root cause

At current `main`, `checkServerDrift()` calls `isServerProbablyRunning(beadsDir)`. That helper always reads:

```go
filepath.Join(beadsDir, doltserver.PIDFileName)
```

The supported shared server writes its PID beneath `~/.beads/shared-server/`, not the project `.beads/` directory. A missing project PID therefore creates a false negative; an unrelated stale project PID can create a false positive.

## Suggested fix

When `dolt.shared-server=true`, resolve the managed shared-server directory through the same `doltserver` path helper used by `bd dolt status/start`. Retain the current project-local PID check for standalone server mode. Do not copy or mirror PID files between directories.

Add tests for:

- healthy shared server with no project-local PID;
- stopped shared server with a stale project-local PID;
- unchanged standalone-server behavior;
- non-mutating drift checks in all cases.

maphew:
## Why

`bd config drift` always read the project-local `dolt-server.pid`, even when the effective configuration selected the managed shared server under `~/.beads/shared-server`. That produced false negatives when the shared server was healthy and could produce false positives from an unrelated project PID.

## What changed

- add a canonical, non-creating shared-server path resolver
- retain the existing create-on-use lifecycle helper behavior
- select the shared PID directory from the effective shared-mode predicate, including `BEADS_DOLT_SHARED_SERVER`
- keep standalone server checks project-local
- cover healthy shared state, ignored project PIDs, env-only shared mode, standalone behavior, and the read-only guarantee

## Validation

- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/doltserver -run '^TestSharedServer(Path|Dir)' -count=1`
- Linux/WSL: `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^TestCheckServerDrift(SharedServerHealthyWithoutProjectPID|SharedServerEnvOverride|SharedServerMissingIgnoresProjectPID|StandaloneRunningPreservesInfo)$' -count=1`
- `git diff --check`

Windows PID liveness remains tracked separately by #4808; this PR deliberately does not overlap its platform helper changes.

Fixes #4836

### DIFF
--- a/cmd/bd/config_drift.go
+++ b/cmd/bd/config_drift.go
@@ -262,10 +262,24 @@ func checkServerDrift() []DriftItem {
 		}}
 	}
 
-	sharedServerEnabled := config.GetString("dolt.shared-server")
-	wantServer := strings.EqualFold(sharedServerEnabled, "true")
+	wantServer := doltserver.IsSharedServerMode()
 
-	serverRunning := isServerProbablyRunning(beadsDir)
+	serverDir := beadsDir
+	if wantServer {
+		var err error
+		serverDir, err = doltserver.SharedServerPath()
+		if err != nil {
+			return []DriftItem{{
+				Check:    "server",
+				Status:   driftStatusDrift,
+				Message:  fmt.Sprintf("dolt.shared-server is enabled but its state directory cannot be resolved: %v", err),
+				Expected: "resolvable shared server state directory",
+				Actual:   "unavailable",
+			}}
+		}
+	}
+
+	serverRunning := isServerProbablyRunning(serverDir)
 
 	if wantServer && !serverRunning {
 		return []DriftItem{{
--- a/cmd/bd/config_drift_test.go
+++ b/cmd/bd/config_drift_test.go
@@ -4,8 +4,10 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 
+	"github.com/steveyegge/beads/internal/doltserver"
 	"github.com/steveyegge/beads/internal/git"
 )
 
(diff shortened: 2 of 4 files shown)


maphew (commented):

The fix correctly aligns `bd config drift`'s server-PID lookup with where the shared server actually writes its PID: it selects the shared-server directory via `IsSharedServerMode()` (now also honoring `BEADS_DOLT_SHARED_SERVER`) and reads the PID from the non-creating `SharedServerPath()`, matching the production `resolveServerDir`/`pidPath` location. The `SharedServerDir` refactor preserves create-on-use behavior, and the added tests meaningfully cover healthy shared state, env-only shared mode, ignored project PIDs, standalone behavior, and the read-only guarantee. No correctness or scope problems found; the only note is a rare error-path behavioral divergence.

**Findings:**
- **nit** `cmd/bd/config_drift.go:272` — When SharedServerPath() errors (only when os.UserHomeDir fails), the drift check reports drift ("state directory cannot be resolved"), whereas the production path resolveServerDir() warns and falls back to the project .beads dir, meaning the running server's PID would actually live in the project dir. This is a benign divergence for an extremely rare failure mode, but it means drift could report a false positive in the one case where production silently degraded to per-project mode. Consider mirroring the fallback-to-beadsDir behavior for consistency, or leave as-is since surfacing the unresolved-home condition as drift is defensible.

PR #4953, opened 2026-07-22
fix(setup): migrate legacy Codex session hook

### LINKED ISSUE #4608: `bd setup codex` silently duplicates hook execution when a pre-existing, non-`bd codex-hook`-prefixed entry occupies the same event — including the shape `docs/CODEX_INTEGRATION.md` itself documents as the "Manual Fallback"
## Summary

`bd setup codex` upserts its native hooks (`SessionStart`, `PreCompact`, `PostCompact`, `UserPromptSubmit`) into `.codex/hooks.json` by removing only the entries it recognizes as self-managed, then appending its own. "Recognized as self-managed" means "a command string starting with `bd codex-hook `" (`codexHookEntryManaged`, `cmd/bd/setup/codex.go`). Any existing entry under the same event whose command does **not** match that prefix is left untouched, and the new managed entry is appended alongside it — so the event ends up running twice per session, silently, with no warning at install time and no detection by `bd setup codex --check` afterward.

This isn't a hypothetical pre-`bd setup codex` artifact: the shape that collides is exactly what `docs/CODEX_INTEGRATION.md`'s own "Manual Fallback" section used to recommend for projects wiring up the hook by hand before `bd codex-hook` existed, e.g.:

```json
{
  "hooks": {
    "SessionStart": [
      {
        "matcher": "startup|resume",
        "hooks": [
          { "type": "command", "command": "bd prime --stealth | jq -Rs '{hookSpecificOutput:{hookEventName:\"SessionStart\",additionalContext:.}}'" }
        ]
      }
    ]
  }
}
```

Running `bd setup codex` against a project with this hooks.json does not replace it — it appends a second `SessionStart` entry, so both the old jq-based command and the new `bd codex-hook SessionStart` run on every session start, each independently invoking `bd prime` and injecting Beads context. We reproduced this identically across three separate projects that had set up the hook this way before `bd codex-hook` existed.

## Where this lives in the code (`origin/main` @ `8e4e59d39`, same in v1.1.0)

- `codexHookEntryManaged` (`cmd/bd/setup/codex.go`) treats an entry as "ours" only if some command in it has the prefix `"bd codex-hook "`. Anything else — including the project's own prior `bd prime`-based hook — is invisible to this check.
- `removeCodexManagedHookEvent` filters an event's entry list down to non-managed entries only (i.e., it deliberately preserves anything not matching the check above).
- `upsertCodexManagedHooks` calls `removeCodexManagedHookEvent` then appends the new managed entries — so a pre-existing non-managed entry for the same event survives, and the new entry is added next to it rather than replacing it.
- `codexManagedHooksCurrent`, used by `bd setup codex --check`, only verifies that the wanted managed entries are *present* — it never checks for the absence of extra/duplicate entries for the same event, so the check reports a clean `✓ Codex native hooks installed` even with the duplicate sitting right next to it.

## Reproduction

```bash
mkdir -p demo/.codex && cd demo
cat > .codex/config.toml <<'EOF'
[features]
hooks = true
EOF
cat > .codex/hooks.json <<'EOF'
{
  "hooks": {
    "SessionStart": [
      {
        "matcher": "startup|resume",
        "hooks": [
          { "type": "command", "command": "bd prime --stealth | jq -Rs '{hookSpecificOutput:{hookEventName:\"SessionStart\",additionalContext:.}}'" }
        ]
      }
    ]
  }
}
EOF
bd setup codex
cat .codex/hooks.json
```

Expected: `SessionStart` ends up with a single, canonical entry.
Actual: `SessionStart` has two entries — the original `jq`-based one, untouched, plus a new one running `bd codex-hook SessionStart`. Both fire on every session start.

## Suggested fixes (any of)

1. Broaden `codexHookEntryManaged` (or add a sibling check) to also recognize commands that pipe through `bd prime` directly — the shape the project's own docs told users to hand-write — so upgrading from the documented manual fallback to native hooks is a clean replacement, not an addition.
2. At minimum, have `bd setup codex` detect *any* other entry present for an event it's about to write and print a warning (e.g., "SessionStart already has an unmanaged hook; you may be double-invoking bd prime — remove `<command>` manually if so") instead of silently appending.
3. Have `bd setup codex --check` flag events that contain more than the expected number of entries, so post-install drift like this doesn't require reading `hooks.json` by hand to notice.

## Impact

Not destructive, but it means every session start (and, since the pattern generalizes, potentially other events too) runs `bd prime` twice and injects the resulting Beads context twice into the model's input — silent token waste that compounds per session, with no signal from either the install command or `--check` that anything is off.

## Environment

- `bd` v1.1.0 (also present on `origin/main` @ `8e4e59d39`)
- Reproduced identically across three separate local projects, each of which had wired up the officially-documented manual `SessionStart` hook before `bd codex-hook` existed, then later ran `bd setup codex` to pick up the new `PreCompact`/`PostCompact`/`UserPromptSubmit` hooks.

maphew:
## Why

Projects that followed the former documented `bd prime --stealth | jq` Codex `SessionStart` fallback receive a second canonical hook when they later run `bd setup codex`. Both commands then run on every session start, duplicating Beads context and token use while `--check` still reports the installation as current.

## What changed

- recognize only the exact historical SessionStart command, not arbitrary `bd prime` hooks
- migrate/remove that nested command while preserving unrelated commands and entries
- make legacy-plus-canonical configurations report stale
- recognize legacy-only hooks as an existing Beads integration
- cover migration, idempotency, mixed entries, unrelated near-matches, check behavior, and removal

## Validation

- `go test ./cmd/bd/setup -run '^(TestCodexHooks|TestCodexManagedHooksCurrent|TestRemoveCodexHooks)' -count=1`
- Linux/WSL: `go test ./cmd/bd/setup -count=1`
- `git diff --check`

Fixes #4608


maphew (commented):

The PR does what its title/body claims: it migrates the exact historical Codex SessionStart jq-pipeline hook to the canonical `bd codex-hook SessionStart`, makes legacy+canonical configs report stale via `--check`, recognizes legacy-only hooks as an existing integration, and removes only the exact legacy command on uninstall while preserving unrelated hooks. The legacy constant matches the command documented in issue #4608 exactly; detection is command-only and correctly matcher-agnostic. Filter-in-place logic handles all entry shapes (managed-only, legacy-only, mixed, non-map, near-miss) without aliasing bugs, and idempotency holds. Tests are meaningful and cover the stated cases; scope is limited to codex.go/codex_test.go with no smuggled changes. No blocker or major issues found.

**Findings:**
- **nit** `cmd/bd/setup/codex.go:805` — Migration is intentionally exact-match-only (documented in the PR body), so it implements only fix option #1 from issue #4608, not #2/#3 (warn/flag on any unmanaged extra entry). A project whose hand-written SessionStart hook differs from the exact historical string (e.g. added a flag or altered internal spacing) will still silently double-invoke `bd prime` with no warning from either `bd setup codex` or `--check`. This is a reasonable conservative tradeoff, but the residual double-run case for varied hand-written hooks remains uncovered; a follow-up warning on any non-managed SessionStart entry would fully close the issue. Not blocking.

Issue #2636, opened 2026-03-16
v0.61.0: bd doctor infinite restart loop with zombie dolt processes


graffhyrum:
## Summary

`bd doctor` in v0.61.0 enters an infinite Dolt server restart loop, accumulating zombie processes and stale lock files. The server lifecycle changed to start-use-stop per command, but `bd doctor` retries indefinitely when the server shuts down between checks.

## Repro

1. Install bd v0.61.0
2. Have a `.beads/` project with Dolt backend
3. Run `bd doctor` (or trigger it via a session startup hook)
4. Observe: process never exits, zombie `dolt` processes accumulate

## Observed Behavior

Each cycle in the log:

```
Starting server with Config HP="127.0.0.1:<random-port>"|T="28800000"|R="false"|L="info"
Server ready. Accepting connections.
NewConnection ... connectionID=1
ConnectionClosed connectionID=1
... (schema init, auto-commit on all databases)
error running query ... error="nothing to commit"
Server closing listener. No longer accepting connections.
stats stopped: context canceled
Starting server with Config HP="127.0.0.1:<different-random-port>"|T="...
```

This repeats indefinitely. After ~2 minutes: 60+ zombie `[dolt] <defunct>` processes, stale `noms/LOCK` files in every database directory, 85K+ lines in `dolt-server.log`.

### Evidence

- **668 server restarts** logged in a single session
- All zombie processes share `bd doctor` as parent (never reaped)
- Port file and PID file become stale within seconds (server dies, new one starts on different port)
- `bd doctor` ignores `timeout` signal (SIGTERM) — required `kill -9`
- `bd dolt killall` reports "No orphan dolt servers found" despite dozens of zombies

### Collateral damage

- Stale `LOCK` files block subsequent dolt startups
- Circuit breaker triggers, blocking even manually-started servers
- Other `bd` commands fail with port mismatch (`bd dolt start` reports port X, `bd status` tries port Y)

## Expected Behavior

`bd doctor` should complete within seconds (as it does in v0.60.0: 67 checks passed in ~3s).

## Workaround

Downgrade to v0.60.0:

```bash
# Kill stuck doctor
pkill -9 -f "bd doctor"

# Clean up zombie state
find .beads/dolt -name "LOCK" -delete
rm -f .beads/dolt-server.{pid,port}

# Install v0.60
gh release download v0.60.0 --repo steveyegge/beads --pattern "beads_0.60.0_linux_amd64.tar.gz" -D /tmp
[...]

PR #4954, opened 2026-07-22
fix(init): use server-aware diagnostics

### LINKED ISSUE #4553: bd init --server: false "Setup incomplete: No dolt database found" after successful init — init diagnostics stat the local dolt dir in server mode
## Summary

A fully successful `bd init --server` ends with a false failure banner:

```
✓ bd initialized successfully!
  Backend: dolt
  Mode: server
  Server: root@127.0.0.1:3308
  Database: ts_inspec_profile_parser
  ...

⚠ Setup incomplete. Some issues were detected:
  • Database: No dolt database found

Run bd doctor --fix to see details and fix these issues.
```

The database was in fact created correctly on the server (full ~28-table schema), `bd status` / `bd list` work immediately, and a full `bd doctor` run seconds later passes every database check. The init-time diagnostic is wrong, and it points users at `bd doctor --fix` (and the check's own Fix text points at `bd bootstrap`) on a healthy brand-new board.

## Environment

- bd 1.0.5 (Homebrew), macOS (darwin/arm64)
- dolt sql-server (Homebrew dolt) on 127.0.0.1:3308
- Bug still present at HEAD `7d8063d` (2026-07-02) per code inspection below

## Reproduction

```bash
mkdir repro && cd repro && git init
bd init --server --server-host 127.0.0.1 --server-port 3308 --server-user root \
  --prefix repro --role maintainer --non-interactive
```

Init succeeds, then prints the "Setup incomplete … No dolt database found" banner. Verified immediately afterward:

- `SHOW TABLES` on the server database: full schema present (~28 tables)
- `bd status`, `bd list`: work fine
- `bd doctor` (read-only): database checks pass, 0 errors

Adding `--external` makes no difference — the identical false warning fires (tested both ways).

## Root cause

The init-time diagnostics use the standalone check variants, which require a **local** dolt directory that server-mode repos intentionally don't have:

1. `cmd/bd/init.go:1706` — server-mode init calls `runInitDiagnostics(cwd)`
2. `cmd/bd/doctor.go:899` — `runInitDiagnostics` calls the standalone `doctor.CheckDatabaseVersion(path, Version)`
3. `cmd/bd/doctor/database.go:65-74` — `CheckDatabaseVersion` short-circuits on `os.Stat(getDatabasePath(beadsDir))` (i.e. `.beads/dolt`) **before attempting any server connection**. In server mode that directory doesn't exist by design, so it returns `StatusError: "No dolt database found"` even though the configured server database is healthy.

The full `bd doctor` path doesn't have this bug because GH#2636 introduced `CheckDatabaseVersionWithStore` (`cmd/bd/doctor/database.go:93-116`), which guards the local-dir stat with `sharedStoreNeedsLocalDoltDir()` (`cmd/bd/doctor/shared_store.go:43`) — i.e. it only requires a local dolt dir when the repo is *not* in server mode. That's exactly why `bd init` warns and `bd doctor` passes on the same board seconds apart.

## Secondary effect: init diagnostics skip real validation in server mode

The other standalone checks used by `runInitDiagnostics` have the inverse problem — they treat "no local dolt dir" as "no database → N/A (OK)" and return early:

- `CheckSchemaCompatibility` (`database.go:174`)
- (and the same pattern in `CheckDatabaseIntegrity:236`, `CheckDatabaseSize:515`, `CheckProjectIdentity:345` for the full doctor's standalone paths)

So in server mode, init diagnostics simultaneously **false-error** on the version check and **silently skip** schema validation of the database that was actually created.

## Suggested fix

Apply the same server-mode guard the `*WithStore` variants already use: in the standalone checks, only treat a missing local dolt dir as "no database" when `sharedStoreNeedsLocalDoltDir(beadsDir)` is true; in server mode, proceed to `dolt.NewFromConfigWithCLIOptions(...)`, which connects to the configured server and validates the real database.

Alternatively, have `runInitDiagnostics` open a `NewSharedStore` and call the `*WithStore` variants, matching what full `bd doctor` does since GH#2636.

## Impact

This is a first-contact failure: the very first command a new user runs on a shared-server team setup ends by telling them setup failed and to run `bd doctor --fix` / `bd bootstrap`. Users (and AI coding agents) follow that advice and start "repairing" a healthy board — re-initing, adding flags, tearing down databases. On our team this class of false alarm has directly caused developers to refuse to adopt beads.

maphew:
## Why

`bd init --server` successfully creates and connects to a server database, but its final diagnostics still used path-only checks that require a local `.beads/dolt` directory. Healthy server-mode setups therefore ended with a false "No dolt database found" failure and advice to repair an already-working board.

Fixes #4553.

## What changed

- Open one shared, server-aware doctor store for init-time database checks.
- Route version, schema compatibility, and permissions checks through their existing `WithStore` variants.
- Add a server-backed regression proving init diagnostics succeed when `.beads/dolt` is intentionally absent.

The installation, Dolt-format, connection, and schema-health checks retain their existing behavior.

## Validation

- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^$' -count=1`
- `go test -tags gms_pure_go ./cmd/bd -run '^TestRunInitDiagnosticsServerModeWithoutLocalDoltDir$' -count=1` (compiles locally; assertions require the Docker-backed CI Dolt server and skip when unavailable)

### DIFF
--- a/cmd/bd/doctor.go
+++ b/cmd/bd/doctor.go
@@ -996,22 +996,28 @@ func runInitDiagnostics(path string) doctorResult {
 		result.OverallOK = false
 	}
 
+	// Open one shared store for the database-backed init checks. Server-mode
+	// workspaces may not have a local .beads/dolt directory, so the path-only
+	// checks would otherwise report a false missing database.
+	sharedStore := doctor.NewSharedStore(path)
+	defer sharedStore.Close()
+
 	// Check 2: Database version
-	dbCheck := convertWithCategory(doctor.CheckDatabaseVersion(path, Version), doctor.CategoryCore)
+	dbCheck := convertWithCategory(doctor.CheckDatabaseVersionWithStore(sharedStore, Version), doctor.CategoryCore)
 	result.Checks = append(result.Checks, dbCheck)
 	if dbCheck.Status == statusError {
 		result.OverallOK = false
 	}
 
 	// Check 3: Schema compatibility
-	schemaCheck := convertWithCategory(doctor.CheckSchemaCompatibility(path), doctor.CategoryCore)
+	schemaCheck := convertWithCategory(doctor.CheckSchemaCompatibilityWithStore(sharedStore), doctor.CategoryCore)
 	result.Checks = append(result.Checks, schemaCheck)
 	if schemaCheck.Status == statusError {
 		result.OverallOK = false
 	}
 
 	// Check 4: Permissions
-	permCheck := convertWithCategory(doctor.CheckPermissions(path), doctor.CategoryCore)
+	permCheck := convertWithCategory(doctor.CheckPermissionsWithStore(path, sharedStore), doctor.CategoryCore)
 	result.Checks = append(result.Checks, permCheck)
 	if permCheck.Status == statusError {
 		result.OverallOK = false
--- a/cmd/bd/doctor_init_diagnostics_cgo_test.go
+++ b/cmd/bd/doctor_init_diagnostics_cgo_test.go
@@ -1,0 +1,32 @@
+//go:build cgo
+
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestRunInitDiagnosticsServerModeWithoutLocalDoltDir(t *testing.T) {
+	projectDir := t.TempDir()
+	doltPath := filepath.Join(projectDir, ".beads", "dolt")
+	store := newTestStoreIsolatedDB(t, doltPath, "initdiag")
+	if err := store.Close(); err != nil {
+		t.Fatalf("close setup store: %v", err)
+	}
+
+	if info, err := os.Stat(doltPath); err == nil || !os.IsNotExist(err) {
+		t.Fatalf("expected no local .beads/dolt directory, got info=%v err=%v", info, err)
+	}
+
+	result := runInitDiagnostics(projectDir)
+	for _, check := range result.Checks {
+		if check.Message == "No dolt database found" {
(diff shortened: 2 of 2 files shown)


maphew (commented):

The PR faithfully reuses the established #2636 shared-store pattern (already present in the main runDiagnostics at doctor.go:509+) to fix the server-mode false "No dolt database found" in init diagnostics. I traced all six init checks: the three rerouted checks now correctly succeed against a server-backed store while preserving the local-mode missing-database behavior, and the four unchanged checks are genuinely unaffected in server mode. The regression test is meaningful and follows the repo's existing //go:build cgo Dolt-server test convention. No blocker or major findings.

**Findings:**
- **nit** `cmd/bd/doctor_init_diagnostics_cgo_test.go:24` — The test guards against the specific message string "No dolt database found". This couples the regression guard to that exact wording; if the message is later reworded, this loop silently stops guarding the original bug while still passing. The OverallOK assertion below is the real backstop, so this is belt-and-suspenders and acceptable, but consider asserting the Database check reaches StatusOK/StatusWarning rather than matching a literal error string.
- **nit** `cmd/bd/doctor_init_diagnostics_cgo_test.go:1` — The test only compiles under a cgo build and skips without the Docker-backed Dolt server, so the default pure-go CI lane (CGO_ENABLED=0 -tags gms_pure_go) does not exercise it. This matches the existing convention for Dolt-server tests in cmd/bd and is guarded by the cgo CI lane, so it is not a defect - just noting the regression is not covered in the pure-go lane.

PR #4955, opened 2026-07-22
fix(config): honor disabled Dolt auto-start

### LINKED ISSUE #4509: Centralize and test the BEADS_DOLT_AUTO_START=0 externally-managed-server contract — bd config apply bypasses the fence
**Problem.** `BEADS_DOLT_AUTO_START=0` (and `dolt.auto-start: false`) is the documented signal that an external process owns the Dolt sql-server and `bd` must not start or kill it. `IsAutoStartDisabled()` (`internal/doltserver/doltserver.go:165`) is checked in the two main lifecycle paths — `EnsureRunningDetailed` (`doltserver.go:640`) and `KillStaleServers`/`killStaleServersForDir` (`doltserver.go:1133`, `:1140`) — but the gate is enforced per-call-site rather than centrally, and at least one server-start path skips it. There is also no single regression test asserting "no doltserver code path starts or kills a server when the flag is set," so a new start trigger can silently regress the contract.

**Root cause (origin/main, `fa4dce454`).** `cmd/bd/config_apply.go` `applyServer()` decides to start a server purely from `wantServer := config.GetString("dolt.shared-server") == "true"` (`config_apply.go:386`) plus drift (`config_apply.go:99`), then calls `doltserver.Start(serverDir)` at `config_apply.go:408` with **no** `IsAutoStartDisabled()` / `ResolveServerMode == ServerModeExternal` check. A consumer with `dolt.shared-server: true` and `BEADS_DOLT_AUTO_START=0` (external manager owns even the shared server) will have a duplicate server spawned by `bd config apply`. The existing gate is correct where present but is copy-pasted into individual call sites; nothing prevents the next `Start()` caller from forgetting it. (The other two direct `Start()` callers — `cmd/bd/dolt.go:517` `bd dolt start` and `cmd/bd/init.go:976` shared-init — are explicit user commands and intentionally ungated; only the `config apply` reconciliation path is a silent bypass.)

**Why it matters to a downstream consumer.** Gas City runs an external, supervisor-managed Dolt sql-server and sets `BEADS_DOLT_AUTO_START=0` as its primary fence against a duplicate lifecycle start (a second server on a different ephemeral port silently splits the bead store / corrupts the LOCK). A bypassing start path means a routine `bd config apply` can breach that fence without any error or log — precisely the failure class `BEADS_DOLT_AUTO_START=0` is supposed to make impossible.

**Existing coverage.** `doltserver_test.go:827` `TestIsAutoStartDisabled`, `:876` `TestIsAutoStartDisabled_Sources`, `:796` `TestKillStaleServersSkipsAutoStartDisabled` cover the gate piecemeal; none assert "no path starts a server when the flag is set," and the `config apply` path is untested for this. Dedup: #3057 / PR #3058 (merged) fixed `resolveAutoStart` priority in `internal/storage/dolt/open.go` (different path); PR #2676 (GH#2641) fixed `KillStaleServers`; PR #3473 fixed `ApplyCLIAutoStart`. None centralize the contract or test the `config apply` start path; no open/closed issue requests the contract doc + comprehensive test.

**Fix candidates (at least two):**
1. **Centralize the fence.** Make `Start()` itself return a typed `ErrAutoStartDisabled` when the flag is set, so every caller (including `config apply`) inherits the guard instead of re-implementing it. Explicit-override commands (`bd dolt start`, shared-init) opt in via an internal `startForce`/`AllowExternalStart` parameter.
2. **Gate the specific call site** (minimal fix): in `applyServer()` (`config_apply.go`), short-circuit to an `applyStatusSkipped` result when `IsAutoStartDisabled()` is true, mirroring `EnsureRunningDetailed`, and document that `config apply` never starts an externally-managed server.
3. **Add a contract test** (ships with the fix): a table-driven test that, with `BEADS_DOLT_AUTO_START=0`, asserts every public start/kill entry point (`EnsureRunning`/`EnsureRunningDetailed`, `KillStaleServers`, the `config apply` server path) performs no spawn/kill — using injected spawn/kill hooks so it needs no real `dolt` binary, extending `TestKillStaleServersSkipsAutoStartDisabled` (`doltserver_test.go:796`).
4. **Document the contract** in a doc comment / `docs/`: `BEADS_DOLT_AUTO_START=0` is a stable public contract — bd will neither start nor kill a server regardless of socket/PID-file state — listing the env var + `dolt.auto-start` config as the two equivalent sources.

---
_Filed by the Gas City maintainer fleet from a downstream interdependency audit; verified against `gastownhall/beads` origin/main @ `fa4dce454`._

maphew:
## Why

`BEADS_DOLT_AUTO_START=0` and `dolt.auto-start: false` mark a Dolt server as externally managed. Most implicit lifecycle paths honor that fence, but `bd config apply` could silently bypass it and start a duplicate shared server during drift reconciliation.

Fixes #4509.

## What changed

- Skip shared-server reconciliation before dry-run or startup when either auto-start policy source disables management.
- Keep explicit `bd dolt start` and server initialization behavior unchanged.
- Cover environment- and workspace-config disable sources independently in both apply and dry-run modes.
- Assert the skipped path creates no shared-server state directory.

## Validation

- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^TestApplyServer' -count=1 -v`
- Independent correctness review: no blocking findings; repeated focused runs passed

### DIFF
--- a/cmd/bd/config_apply.go
+++ b/cmd/bd/config_apply.go
@@ -394,6 +394,18 @@ func applyServer(drifted bool, dryRun bool) ApplyResult {
 		}
 	}
 
+	// Reconciliation must honor the same auto-start policy as implicit storage
+	// opens. A disabled auto-start means the server is externally managed; keep
+	// Action="start" to describe the skipped reconciliation action consistently.
+	if doltserver.IsAutoStartDisabled() {
+		return ApplyResult{
+			Check:   "server",
+			Action:  "start",
+			Status:  applyStatusSkipped,
+			Message: "Dolt shared server not started because auto-start is disabled; the server is externally managed",
+		}
+	}
+
 	serverDir := doltserver.ResolveServerDir(beadsDir)
 
 	if dryRun {
--- a/cmd/bd/config_apply_test.go
+++ b/cmd/bd/config_apply_test.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


maphew (commented):

The fix adds an IsAutoStartDisabled() guard in applyServer so that bd config apply's drift reconciliation no longer starts a shared Dolt server when auto-start is disabled via env (BEADS_DOLT_AUTO_START) or workspace config (dolt.auto-start). I confirmed the bug is real: doltserver.Start does not check IsAutoStartDisabled at entry, so the pre-fix path would genuinely launch a server past the externally-managed fence. The guard is placed correctly (after the wantServer gate) and follows the existing skipped-return idiom. The new table test exercises both disable sources independently, and its assertions (status skipped, and no shared-server dir created even in dry-run) are meaningful because config.yaml makes FindBeadsDir resolve the temp workspace and BEADS_DIR/config.yaml is the top-priority config source. No blocker/major findings; change is well-scoped with no smuggled-in edits.

PR #4956, opened 2026-07-22
fix(dep): allow matching explicit parent edges

### LINKED ISSUE #4398: beads CLI: dep add / link refuses an explicit parent-child edge for an implicit child
## Summary

`bd dep add` / `bd link` refuse to create an explicit `parent-child` edge for an issue that is already an implicit (dot-numbered) child of the target parent, erroring with a "would create a deadlock" message. This makes it impossible to materialize an explicit parent edge for an implicit child through the normal dependency-add path.

## Environment

- Observed on `bd` 1.0.4 (and 1.0.3). Please confirm against latest `main`.

## Reproduction

```
bd link P.1 P --type parent-child
# → Error: cannot add dependency: P.1 is already a child of P.
#   Children inherit dependency on parent completion via hierarchy.
#   Adding an explicit dependency would create a deadlock

bd dep add P.1 P --type parent-child --no-cycle-check
# → same refusal
```

## Expected

Either the explicit `parent-child` edge is added idempotently (it agrees with the existing implicit hierarchy), or the refusal message points at the supported way to materialize the edge.

## Actual

Both `bd link` and `bd dep add` refuse; `--no-cycle-check` does not help.

## Workaround

`bd update <id> --parent <parent>` succeeds where `bd dep add` / `bd link` refuse — it sets the explicit parent via the reparent channel rather than the dependency-add guard. For hierarchy queries, rely on `bd children`.

## Severity / impact

**Medium** — correctness/usability. The reparent workaround exists, but the `dep add`/`link` refusal is surprising and blocks the obvious path to an explicit parent edge.

maphew:
## Why

The CLI rejected every dependency from a dotted child to its parent as a deadlock, even when the requested edge was the matching structural `parent-child` relationship. Storage already supports materializing that edge idempotently, so `bd dep add` and `bd link` contradicted the underlying model and the `bd update --parent` path.

Fixes #4398.

## What changed

- Centralize hierarchy conflict detection for dependency commands.
- Allow only `parent-child` to the dotted ID's immediate parent.
- Continue rejecting blocking/other edge types to any parent or ancestor.
- Continue rejecting `parent-child` links to a grandparent or higher ancestor.
- Apply the shared rule to direct, bulk, link, and proxied-server paths.

## Validation

- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^TestIsDisallowedHierarchicalDependency$' -count=1 -v`
- `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run '^TestEmbeddedExplicitHierarchicalParentEdges$' -count=1 -v`
  - six subtests passed: direct add, link, bulk, blocking guards, and grandparent rejection

### DIFF
--- a/cmd/bd/dep.go
+++ b/cmd/bd/dep.go
@@ -65,18 +65,36 @@ func resolveIDForMutation(ctx context.Context, localStore storage.DoltStorage, i
 // isChildOf returns true if childID is a hierarchical child of parentID.
 // For example, "bd-abc.1" is a child of "bd-abc", and "bd-abc.1.2" is a child of "bd-abc.1".
 func isChildOf(childID, parentID string) bool {
+	_, isAncestor := hierarchicalParentRelation(childID, parentID)
+	return isAncestor
+}
+
+func hierarchicalParentRelation(childID, targetID string) (immediateParent string, isAncestor bool) {
 	// A child ID has the format "parentID.N" or "parentID.N.M" etc.
 	// Use ParseHierarchicalID to get the actual parent
 	_, actualParentID, depth := types.ParseHierarchicalID(childID)
 	if depth == 0 {
-		return false // Not a hierarchical ID
+		return "", false // Not a hierarchical ID
 	}
 	// Check if the immediate parent matches
-	if actualParentID == parentID {
-		return true
+	if actualParentID == targetID {
+		return actualParentID, true
 	}
-	// Also check if parentID is an ancestor (e.g., "bd-abc" is parent of "bd-abc.1.2")
-	return strings.HasPrefix(childID, parentID+".")
+	// Also check if targetID is an ancestor (e.g., "bd-abc" is an ancestor of "bd-abc.1.2")
+	return actualParentID, strings.HasPrefix(childID, targetID+".")
+}
+
+// isDisallowedHierarchicalDependency reports whether an explicit dependency
+// conflicts with hierarchy encoded in a dotted issue ID. The one allowed match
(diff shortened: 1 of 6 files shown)


maphew (commented):

The change does exactly what the title/body claims: it narrows the hierarchy guard so a parent-child edge to a dotted ID's immediate parent is allowed while blocking/other edge types and parent-child edges to higher ancestors stay rejected. The refactor is correct, all eight former isChildOf call sites are converted with the right dependency type, and the new logic matches existing implicit-hierarchy semantics (parent-child is a blocking type, so no new deadlock is introduced). Unit and cgo/embedded tests are present and meaningful, using pre-existing helpers with correct build tags. No blocker or major issues found; only minor test-coverage and message-clarity nits.

**Findings:**
- **nit** `cmd/bd/dependency_hierarchy_test.go:10` — The pure-predicate unit test omits two branches worth locking in: a depth-0 (non-hierarchical fromID) case that should return false, and a non-blocking custom/well-known type (e.g. related or discovered-from) to the immediate parent that should return true. These distinguish the new narrowing from the old blanket rejection; adding them would guard against future regressions cheaply.
- **nit** `cmd/bd/dep.go:178` — The reused error text ("...would create a deadlock") is now imprecise for the narrowed rejections: e.g. `bd link child parent --type related` still fails with a deadlock message even though a related edge is non-blocking. This is pre-existing behavior (isChildOf rejected all types), not introduced here and out of scope for #4398, but the message could be clarified in a follow-up since parent-child is now the one accepted type.

Issue #4646, opened 2026-07-07
bd --global correctness gaps: create --id prefix validation reads CWD config instead of resolved store, and where ignores --global entirely


seanmartinsmith:
## problem
Two separate `--global` correctness gaps: `bd create --global --id/--parent` validates the explicit ID's prefix against the wrong config source, and `bd where --global` ignores the flag entirely and reports the CWD database instead of the global one.

## versions
bd v1.1.0 (commit 8e4e59d39); confirmed present at main @187ba85f3.

### 1. `bd create --global --id=...` validates against the CWD prefix, not the resolved store's prefix
- **location**: `cmd/bd/create.go:469-474`
  ```go
  var dbPrefix, allowedPrefixes string
  if yamlPrefix := config.GetString("issue-prefix"); yamlPrefix != "" {
      dbPrefix = yamlPrefix
  } else {
      dbPrefix, _ = store.GetConfig(ctx, "issue_prefix")
  }
  ```
  `config.GetString("issue-prefix")` reads the process-global viper config, which is populated from whatever `.beads/config.yaml` was found walking up from the CWD at startup — independent of `--global` or any other store-redirect. It's checked *first*, so it wins over `store.GetConfig(ctx, "issue_prefix")` (the actual resolved target store, e.g. `beads_global` when `--global` is set) whenever the CWD happens to have its own `issue-prefix` configured, which is the common case for any already-initialized repo.
- **repro**: in a repo whose `.beads/config.yaml` has `issue-prefix: myrepo`, running `bd create --global --id=global-123 --title=x` validates `global-123`'s prefix (`global`) against `myrepo` (the CWD's prefix) instead of against `beads_global`'s actual prefix (`global`) — the wrong config source is consulted even though the store being written to is correct.
- **fix**: check the resolved store's `issue_prefix` first (swap the branch order), falling back to `config.GetString("issue-prefix")` only when the store has none set. Or short-circuit the `yamlPrefix` branch whenever `globalFlag` (or any explicit store redirect) is active.
- Not a dup of #3897 — that's a different path (`ReadConfigPrefix` in shared-server, `internal/storage/issueops/helpers.go:537`), unrelated to this CLI-side branch ordering in `create.go`.

### 2. `bd where --global` never reads the global flag — reports the CWD database
[...]

Issue #2469, opened 2026-03-09
issue-prefix from config.yaml ignored when projects share a Dolt server


marcodelpin:
## Summary

When multiple beads projects share a single Dolt server (via common data directory or port discovery), the `issue-prefix` value from each project's `config.yaml` is ignored. All new issues get the prefix from the **database's** `config` table — which belongs to whichever project first initialized the shared server.

This means projects cannot have distinct issue prefixes (`tm-`, `orin-`, `sys-`, etc.) even though `config.yaml` explicitly sets different values.

## Reproduction

1. Project A: `.beads/config.yaml` has `issue-prefix: "sys"`, runs Dolt server on port 13342
2. Project B: `.beads/config.yaml` has `issue-prefix: "tm"`
3. `bd` in Project B discovers the running server on 13342 (from Project A)
4. `bd create --title "test"` in Project B → creates issue with `sys-xxxx` prefix, not `tm-xxxx`

## Root cause

In `cmd/bd/create.go` lines ~489-494:
```go
dbPrefix, _ = store.GetConfig(ctx, "issue_prefix")
if dbPrefix == "" {
    dbPrefix = config.GetString("issue-prefix")
}
```

The DB config value always wins. Since the DB belongs to Project A's server, it returns `sys`. The YAML config.yaml `issue-prefix` is only used as a fallback when the DB has no value — which never happens because `bd init` always writes it.

## Expected behavior

When a project's `config.yaml` specifies `issue-prefix`, that value should be used for issue creation in that project, regardless of what the shared DB's `config` table says.

## Proposed fix

The YAML `config.yaml` `issue-prefix` should take precedence over the DB `config` table when both are set. The config.yaml is the per-project declaration of intent; the DB config table is an implementation detail that can be stale when shared.

Alternatively, this is naturally resolved by PR #2416 (shared server mode) which creates separate databases per project within a single server, so each DB has its own correct `issue_prefix`.

## Related

- #2372 — Dolt server port cross-project leakage (same root cause family)
- #2377 — Single Dolt server feature request
- PR #2416 — Shared server mode (would fix this architecturally)

## Environment

- bd v0.59.0 (dev: main@4dd397ac)
- Multiple projects under a common directory tree sharing Dolt server
- Each project has its own `.beads/config.yaml` with distinct `issue-prefix`

steveyegge:
[...]

PR #4957, opened 2026-07-22
fix(create): validate global IDs against global prefix

maphew:
## Why

`--global` switches the opened store to `beads_global`, but explicit-ID validation still preferred the current project's YAML prefix. From a project using `proj0`, `bd create --global --id global-...` was therefore rejected before it could write to the correctly selected global database.

This addresses the create-prefix portion of #4646. The separate `where --global` output contract remains intentionally unresolved.

## What changed

- In global mode, validate explicit IDs against the selected store's `issue_prefix`.
- In ordinary project mode, preserve YAML-first then store-fallback precedence from #2469.
- Add pure precedence tests and a shared-server integration regression with differing project/global prefixes.

## Validation

- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^TestSelectCreateIDPrefix$' -count=1 -v`
- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run '^TestGlobalDBIdentityCheck$' -count=1 -v` (compiles locally; existing Windows guard skips the Linux/container assertion)

### DIFF
--- a/cmd/bd/create.go
+++ b/cmd/bd/create.go
@@ -483,12 +483,9 @@ var createCmd = &cobra.Command{
 
 			ctx := createCtx
 
-			var dbPrefix, allowedPrefixes string
-			if yamlPrefix := config.GetString("issue-prefix"); yamlPrefix != "" {
-				dbPrefix = yamlPrefix
-			} else {
-				dbPrefix, _ = store.GetConfig(ctx, "issue_prefix")
-			}
+			storePrefix, _ := store.GetConfig(ctx, "issue_prefix")
+			dbPrefix := selectCreateIDPrefix(globalFlag, config.GetString("issue-prefix"), storePrefix)
+			var allowedPrefixes string
 			allowedPrefixes, _ = store.GetConfig(ctx, "allowed_prefixes")
 
 			if err := validation.ValidateIDPrefixAllowed(explicitID, dbPrefix, allowedPrefixes, forceCreate); err != nil {
@@ -698,6 +695,16 @@ func mergeCreateLabels(labels, inheritedLabels []string) []string {
 		return nil
 	}
 	return merged
+}
+
+func selectCreateIDPrefix(global bool, yamlPrefix, storePrefix string) string {
+	if global {
+		return storePrefix
+	}
+	if yamlPrefix != "" {
+		return yamlPrefix
+	}
+	return storePrefix
 }
 
 func renderCreateDryRunPreview(issue *types.Issue, labels, deps []string) {
--- a/cmd/bd/create_prefix_test.go
+++ b/cmd/bd/create_prefix_test.go
@@ -1,0 +1,41 @@
+package main
+
+import "testing"
+
+func TestSelectCreateIDPrefix(t *testing.T) {
+	tests := []struct {
+		name        string
+		global      bool
+		yamlPrefix  string
+		storePrefix string
+		want        string
+	}{
+		{
+			name:        "global ignores YAML",
+			global:      true,
+			yamlPrefix:  "proj0",
+			storePrefix: "global",
+			want:        "global",
+		},
+		{
+			name:        "ordinary prefers YAML",
+			yamlPrefix:  "proj0",
+			storePrefix: "store",
+			want:        "proj0",
+		},
+		{
+			name:        "ordinary falls back to store",
+			storePrefix: "store",
+			want:        "store",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := selectCreateIDPrefix(tt.global, tt.yamlPrefix, tt.storePrefix); got != tt.want {
+				t.Fatalf("selectCreateIDPrefix(%t, %q, %q) = %q, want %q",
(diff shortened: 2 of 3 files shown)


maphew (commented):
**Agent review — verdict: concerns**

The production fix in create.go is correct and idiomatic: selectCreateIDPrefix returns the selected store's prefix in --global mode and preserves the YAML-first/store-fallback precedence otherwise, and the pure unit test covers all three branches well. However, the integration regression test is built on a false assumption about where the project prefix is stored. bd init (non-no-db) writes the prefix to the DB, not to config.yaml (the file gets a commented `# issue-prefix: ""`), so the new assertion will fail when a container is available, and the create/show steps pass on both patched and unpatched code, meaning the test does not actually guard #4646.

**Findings:**
- **blocker** `cmd/bd/global_identity_integration_test.go:107` — The added assertion `if yamlPrefix := config.GetStringFromDir(beadsDir, "issue-prefix"); yamlPrefix != "proj0"` will t.Fatalf whenever the Dolt container is available. `bd init` (non-no-db) does not write the prefix into config.yaml: createConfigYaml is called as createConfigYaml(beadsDir, false, "") (cmd/bd/init.go:1321) and renderInitConfigYAML only emits an uncommented `issue-prefix:` line when `noDbMode && prefix != ""` (cmd/bd/init_templates.go:33-35). The prefix lives in the DB; config.yaml holds the commented `# issue-prefix: ""`, so config.GetStringFromDir returns "", not "proj0". The author's validation note ("compiles locally") suggests the container path never actually ran. Fix: after init, write the prefix into config.yaml explicitly, e.g. os.WriteFile(filepath.Join(beadsDir, "config.yaml"), []byte("issue-prefix: proj0" plus a trailing newline), 0o644) (the pattern used at cmd/bd/init_embedded_test.go:1636), then the assertion is valid.
- **major** `cmd/bd/global_identity_integration_test.go:110` — The create/show regression assertions do not actually exercise the bug. Because config.yaml carries no issue-prefix in this setup, config.GetString("issue-prefix") is "" in the create --global path even on unpatched code, so the old logic already falls back to store.GetConfig against the global store (prefix "global") and validation passes. Thus `bd create --global --id global-explicit` succeeds on both patched and unpatched binaries and the test cannot detect a regression of #4646. To make it a true guard, set issue-prefix=proj0 in the project config.yaml (see the blocker fix) so unpatched code would incorrectly prefer "proj0" and reject the global ID.
- **nit** `cmd/bd/global_identity_integration_test.go:112` — The test hardcodes the literal prefix in explicitGlobalID := "global-explicit" while using doltserver.GlobalDatabaseName in adjacent error messages. Prefer doltserver.GlobalIssuePrefix + "-explicit" so the test stays coupled to the constant it depends on (internal/doltserver/doltserver.go:98).
- **nit** `cmd/bd/create.go:486` — The refactor now always calls store.GetConfig(ctx, "issue_prefix") even in ordinary project mode when a YAML prefix is present, whereas the previous code skipped that DB read on the YAML-first path. selectCreateIDPrefix discards storePrefix in that case, so the result is unchanged; this is only a harmless extra config query, not a correctness issue.

maphew:
Addressed the review findings on `cmd/bd/global_identity_integration_test.go` in ded6f6235:

- **blocker**: after `bd init`, the test now writes `issue-prefix: proj0` into `.beads/config.yaml` explicitly (matching the pattern in `cmd/bd/init_embedded_test.go`), since non-no-db `bd init` only stores the prefix in the DB and leaves config.yaml with a commented `# issue-prefix: ""`. The `config.GetStringFromDir` assertion now checks a value that's actually present.
- **major**: verified the reasoning against `selectCreateIDPrefix` in `cmd/bd/create.go` and `ValidateIDPrefixAllowed` in `internal/validation/bead.go`. With `config.yaml` now carrying `issue-prefix: proj0`, unpatched code (which used the YAML prefix unconditionally, ignoring `--global`) would select `proj0` as `dbPrefix` and reject `global-explicit` since it doesn't match `proj0-`. So the create/show steps now exercise the exact regression path from #4646.
- **nit**: `explicitGlobalID` now derives from `doltserver.GlobalIssuePrefix + "-explicit"` instead of a hardcoded literal.

Validation:
- `gofmt -l cmd/bd/` - clean, no output.
- `CGO_ENABLED=1 go vet -tags gms_pure_go ./cmd/bd/...` - clean, no output.
- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd/ -run TestSelectCreateIDPrefix -count=1 -v` - PASS (all 3 subtests).
- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd/ -run TestGlobalDBIdentityCheck -count=1 -v` - SKIP, verbatim:
  ```
  global_identity_integration_test.go:30: skipping: Dolt container not available: cannot create container provider: Docker not available
  --- SKIP: TestGlobalDBIdentityCheck (7.37s)
  ```
  Docker isn't available in this sandbox, so the integration test only compiles and reaches the container-provider check here; it did not actually run the create/show/list assertions locally. It needs a Docker-backed Dolt container to run for real.

Pushed to `fix/bwpm-global-create-prefix`.

Issue #4640, opened 2026-07-07
bd import/export data-fidelity gaps: bare cross-project deps dropped on reimport, closed_by_session omitted, orphan child-counter aborts import


seanmartinsmith:
## problem
Three independent `bd export` / `bd import` round-trip data-fidelity gaps, surfaced during a fleet-wide canonical-DB rebuild that hit a broader class of bd storage/migration bugs. A fourth item from the same sweep — migration `0053_repair_rig_wisps` guarding on TABLE existence instead of COLUMN existence — was already filed as #4502 and fixed by #4538 (merged 2026-07-02, `ensureIssuesRigColumns`); that same PR also fixed the related #4534 orphan-child-counter brick. These three are still open, unfiled, and independently fileable — grouped here since they're all export/import fidelity gaps found in the same audit pass.

## versions
bd v1.1.0 (commit 8e4e59d39); confirmed present at main @187ba85f3.

## 1. bare cross-project dependency targets get reclassified as local on import and dropped

**location:** `internal/storage/issueops/create.go:683` — `PersistDependenciesWithOptionsResult` calls `kind := ClassifyDepTarget(ctx, tx, dep, false)`, hardcoding `isCrossPrefix=false`. Contrast the interactive single-add path, `internal/storage/dolt/dependencies.go:23,33` (`DoltStore.AddDependency`), which computes `isCrossPrefix := isCrossPrefixDep(dep.IssueID, dep.DependsOnID)` before classifying.

`ClassifyDepTarget` (`internal/storage/issueops/dependencies.go:55-63`) only recognizes a target as external if `isCrossPrefix` is true OR the string literally starts with `"external:"`. A bare cross-project dependency (e.g. `sym-3su` depends on `mkt-456`, added via plain `bd dep add`, no `external:` literal) is classified correctly at *creation* time — the interactive path computes `isCrossPrefixDep` and routes it to `depends_on_external` — but the column stores the *raw* id (`mkt-456`), with no `external:` marker, since the column itself is the type marker.

[...]

PR #4963, opened 2026-07-23
fix(import): preserve bare cross-prefix dependencies

maphew:
## Why

JSONL export preserves cross-project dependency IDs such as `mkt-456` without adding an `external:` marker. The two-phase bulk import path nevertheless classified every bare ID as local, so a valid exported edge like `sym-3su -> mkt-456` was treated as a missing local target and silently skipped.

This is a data-fidelity bug: the issue survives a round trip while part of its dependency graph disappears.

## What changed

- classify bulk-persisted dependency targets as cross-prefix before looking for a local issue or wisp
- store those targets in `depends_on_external`, matching the interactive dependency path
- cover the storage helper, direct JSONL import, and a literal export→fresh-import round trip

Same-prefix missing targets and explicit `external:` targets retain their existing behavior.

## Validation

- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/issueops`
- `CGO_ENABLED=1 BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go ./cmd/bd -run 'TestImportFromLocalJSONL/preserves_bare_cross-prefix_dependency|TestExportImportRoundTrip' -count=1 -timeout=8m`
- `git diff --check`
- repository commit hook: `0 issues`
- independent read-only review: no blockers

## Scope and overlap

This is the first independent acceptance-criterion leaf of #4640. It does not supersede #4286: that larger, currently conflicting dependency-table rewrite still leaves this bulk classifier call unchanged. Keeping this fix small and current-schema-compatible makes the round-trip correction reviewable and portable if that broader work later lands.

### DIFF
--- a/cmd/bd/export_test.go
+++ b/cmd/bd/export_test.go
@@ -344,6 +344,13 @@ func TestExportImportRoundTrip(t *testing.T) {
 		"exp-6", "Round Trip", "round trip test", "", "", "", "open", 1, "feature"); err != nil {
 		t.Fatalf("insert issue: %v", err)
 	}
+	if err := s.AddDependency(ctx, &types.Dependency{
+		IssueID:     "exp-6",
+		DependsOnID: "mkt-456",
+		Type:        types.DepRelated,
+	}, "test"); err != nil {
+		t.Fatalf("add cross-prefix dependency: %v", err)
+	}
 
 	// Export
 	exportFile := filepath.Join(tmpDir, "roundtrip.jsonl")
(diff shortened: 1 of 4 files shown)


maphew (commented):

The fix correctly makes the two-phase bulk import path classify bare cross-prefix dependency targets as external (storing them in depends_on_external and skipping the local existence probe), matching the interactive AddDependency path exactly. The change is minimal and idiomatic: isCrossPrefix is computed the same way as embeddeddolt/dolt/sqlkit callers, same-prefix and explicit external: behavior is preserved, and dep.IssueID is guaranteed set beforehand. Tests are present and meaningful (mock-level classification, direct JSONL import, and a real export→fresh-import round trip), and the mock-expectation updates in create_test.go accurately reflect the new external classification. No blocker or major issues found.

**Findings:**
- **minor** `internal/storage/issueops/create.go:774` — isCrossPrefix is derived purely from prefix inequality, so a genuinely-local issue stored under a different prefix in the same DB (e.g. a multi-prefix/federated database) is now written to depends_on_external instead of being linked as a local issue edge, meaning graph/blocked traversal won't follow into that target's own dependencies. This is not a regression - it mirrors the existing interactive AddDependency path (embeddeddolt/dependencies.go:16, dolt/sqlkit transaction.go) that this PR intentionally aligns with, and the export re-emits the bare id either way. Worth the author confirming that single-DB cross-prefix targets are always intended to be treated as external, since the fix leans on that assumption. No change required for the stated round-trip goal.

PR #4964, opened 2026-07-23
fix(import): allow children of deleted parents

maphew:
## Why

A valid export can contain a closed hierarchical child after its parent was deleted. Bulk import persists that child, then unconditionally tries to create an auxiliary `child_counters` row whose foreign key points at the absent parent. That counter write rolls back the entire transaction, including unrelated issues in the same import batch.

The child is durable history; the ownerless counter is disposable derived state. Import should preserve the former without inventing the latter.

## What changed

- route unknown hierarchy parents through the existing error-aware batched wisp lookup
- verify the routed parent row exists before reading or upserting its counter
- skip counter reconciliation only for a genuinely absent owner
- propagate wisp-routing and owner-lookup errors instead of confusing them with absence
- update the storage conformance contract for Strict and Skip orphan handling

Foreign-key checks remain enabled. Existing issue and wisp parent paths are unchanged.

## Validation

- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/issueops -count=1`
- `CGO_ENABLED=1 BEADS_TEST_EMBEDDED_DOLT=1 go test -tags gms_pure_go ./cmd/bd -run '^TestEmbeddedImportClosedChildWhoseParentWasDeleted$' -count=1`
- existing durable and mixed durable/wisp child-counter tests
- exact embedded-Dolt `CreateOrphanHandling` conformance case
- `git diff --check`
- repository commit hook: `0 issues`
- independent review found and verified the fix for a swallowed wisp-lookup error; final re-review found no blockers

The embedded regression imports an ordinary row and a closed orphan `.7`, proves the batch commits with no ownerless counter, recreates the parent, and verifies the next child allocation is `.8`.

## Scope and overlap

This is the third independent acceptance-criterion leaf of #4640. It is distinct from #4538: that merged migration repairs counter rows already present in a database, while this patch prevents a fresh import from attempting an invalid counter insert. Exact-topic preflight and branch searches found no active implementation.

### DIFF
--- a/cmd/bd/import_orphan_embedded_test.go
+++ b/cmd/bd/import_orphan_embedded_test.go
@@ -1,0 +1,102 @@
+//go:build cgo
+
+package main
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/steveyegge/beads/internal/storage/embeddeddolt"
+	"github.com/steveyegge/beads/internal/types"
+)
+
+func TestEmbeddedImportClosedChildWhoseParentWasDeleted(t *testing.T) {
+	if os.Getenv("BEADS_TEST_EMBEDDED_DOLT") != "1" {
+		t.Skip("set BEADS_TEST_EMBEDDED_DOLT=1 to run embedded dolt import tests")
+	}
+
+	ctx := context.Background()
+	beadsDir := t.TempDir()
+	store, err := embeddeddolt.Open(ctx, beadsDir, "beads", "main")
+	if err != nil {
+		t.Fatalf("open embedded Dolt store: %v", err)
+	}
+	t.Cleanup(func() { _ = store.Close() })
+	if err := store.SetConfig(ctx, "issue_prefix", "test"); err != nil {
+		t.Fatalf("set issue prefix: %v", err)
+	}
+
+	// A valid export can retain a closed child after its parent was deleted.
+	// The unrelated row proves a counter reconciliation failure does not roll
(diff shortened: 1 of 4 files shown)


maphew (commented):

The fix correctly resolves the stated bug: an orphaned hierarchical child (parent deleted before export) no longer forces an FK-violating child_counters insert that rolls back the whole import batch. The parent-existence gate skips only the genuinely ownerless counter while preserving behavior for present issue and wisp parents, and it propagates wisp-routing/owner-lookup errors instead of masking them as absence. Tests are thorough and meaningful (three sqlmock unit tests for the skip and both error paths, an embedded-Dolt integration regression, and an updated conformance contract), and the docs/comment changes match actual current behavior. Only a minor efficiency observation, no correctness or scope concerns.

**Findings:**
- **minor** `internal/storage/issueops/create.go:918` — The per-parent existence check `SELECT 1 FROM %s WHERE id=?` is redundant for wisp parents: any parent flagged `isWisp` was just proven to exist in the `wisps` table by `WispIDSetInTx` within the same transaction, so that branch of the existence query can never return ErrNoRows and adds a guaranteed round-trip. Only the `issues`-table check for non-wisp parents carries new information. Additionally that non-wisp check is issued one-per-parent rather than batched like the adjacent `WispIDSetInTx` call, so a bulk import spanning many parents pays N extra sequential round-trips during reconciliation. Correctness is unaffected; consider skipping the existence query when membership is already known and/or batching the `issues` existence lookup for large imports.

PR #4968, opened 2026-07-23
test(preflight): inject fake gh in Bash

### LINKED ISSUE #4967: PR preflight tests can invoke the real GitHub CLI on Windows
## Summary

`TestPRPreflightSupportsGitHubCLIWithoutClosingIssuesReferences` intermittently invokes the workstation's real `gh.exe` on Windows instead of its disposable fake `gh` script.

The fixture puts a newly created, extensionless POSIX script named `gh` at the front of a Git Bash `PATH`. That is not a deterministic shadow for a native `gh.exe` under Windows/MSYS executable lookup. When the fake is missed, the supposedly hermetic test contacts the real GitHub CLI, depends on authentication state, takes minutes, and can write CLI state into the checkout.

## Environment

- Beads: `main` at `02e38d0bdbd311d2113abbdb9b816a9308d79d93`
- Windows 11 Pro, build 26200
- Go 1.26.5 `windows/amd64`
- Git for Windows 2.54.0
- GitHub CLI 2.76.2 installed as `gh.exe`

## Reproduction

With Git Bash, `jq`, and a native `gh.exe` on the host PATH:

```powershell
go test ./scripts -run '^TestPRPreflightSupportsGitHubCLIWithoutClosingIssuesReferences$' -count=3 -v
```

One observed repeated run passed once and then failed twice after resolving the real CLI. Another single run took 113 seconds and failed across the fixture table with:

```text
To get started with GitHub CLI, please run: gh auth login
exit status 4
```

Escaped invocations also created `.local/state/gh/device-id` under the worktree because the fixture's translated temporary `HOME` was not stable in the stripped child environment.

## Expected result

The test must always call its fake implementation, perform no GitHub network/authentication work, and keep every generated file under a disposable test directory.

## Proposed fix

Run the production preflight script by sourcing it in a Bash process that already defines `gh` as a shell function. `command -v gh` and every subsequent call then resolve in the same shell without relying on filesystem executability, PATHEXT, or MSYS lookup timing.

Use a disposable working directory, keep the call log relative to it, and set the child `HOME` there. Inherit the host environment for the real `bash`, `git`, and `jq` dependencies while the in-shell fake completely intercepts `gh`.

This is test-fixture isolation only; `scripts/pr-preflight.sh` does not need a production change.

---

ecuthiell:
## What

Make the PR-preflight fixture deterministic at the real Bash process boundary:

- execute `scripts/pr-preflight.sh` as a script instead of sourcing it into the test wrapper;
- provide the fake `gh` function through one controlled `BASH_ENV` file;
- construct a curated child environment containing only the resolved Bash/`git`/`jq` paths and isolated HOME, XDG, Git, and GitHub state;
- poison and prove exclusion of inherited startup hooks, shell options, and host configuration; and
- require the integration-tagged process suite on Linux, macOS, and Windows through `CI Gate / Required`.

This remains test-only; production preflight behavior is unchanged.

## Why

Fixes #4967.

An extensionless fake executable is not a deterministic shadow for native `gh.exe` under Windows/MSYS lookup. The first revision avoided that lookup by defining a Bash function, but sourced the production script and inherited the host environment. In particular, inherited `BASH_ENV` executes before the fake is installed and can invoke the real CLI or alter shell behavior. The revised fixture controls that startup boundary and runs the production script normally.

## Verification

Rebased onto upstream `b9b437894272f6fb1c0a0d0422962b0a6050b6d2`; revised head: `cade6876f91b8a92f92e0812fe99c9542cc18a56`.

Local checks passed:

- focused real-process suite: 23.386s after the final base refresh;
- focused race suite: 30.625s;
- missing-tool and hostile-host-state falsifiers fail closed rather than skip or escape to the real CLI;
- `go vet`, `golangci-lint` 2.10.1 (0 issues), `actionlint` v1.7.12, and `git diff --check`.

The required platform matrix now supplies permanent Linux, macOS, and Git-Bash-on-Windows evidence. Those remote results remain required before merge.

---

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -184,6 +184,26 @@ jobs:
           path: ${{ runner.temp }}/cli-docs-freshness.patch
           if-no-files-found: ignore
 
+  pr-preflight-platforms:
+    name: PR preflight process (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v6
+
+      - name: Set up Go
+        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Exercise the real Bash process boundary
+        shell: bash
+        run: go test '-tags=integration,gms_pure_go' -count=1 -run '^TestPRPreflight' ./scripts
+
   # Fast check to catch accidental .beads/issues.jsonl changes from contributors
   check-no-beads-changes:
     name: Check for .beads changes
@@ -544,6 +564,7 @@ jobs:
       - check-version-consistency
       - check-migration-hygiene
       - check-doc-flags
+      - pr-preflight-platforms
       - check-no-beads-changes
       - detect-package-gates
       - package-mcp
@@ -569,6 +590,7 @@ jobs:
             CHECK_VERSION_CONSISTENCY
             CHECK_MIGRATION_HYGIENE
             CHECK_DOC_FLAGS
+            PR_PREFLIGHT_PLATFORMS
             CHECK_NO_BEADS_CHANGES
             DETECT_PACKAGE_GATES
             PACKAGE_MCP
@@ -586,6 +608,7 @@ jobs:
           CHECK_VERSION_CONSISTENCY: ${{ needs.check-version-consistency.result }}
           CHECK_MIGRATION_HYGIENE: ${{ needs.check-migration-hygiene.result }}
           CHECK_DOC_FLAGS: ${{ needs.check-doc-flags.result }}
+          PR_PREFLIGHT_PLATFORMS: ${{ needs.pr-preflight-platforms.result }}
           CHECK_NO_BEADS_CHANGES: ${{ needs.check-no-beads-changes.result }}
           DETECT_PACKAGE_GATES: ${{ needs.detect-package-gates.result }}
           PACKAGE_MCP: ${{ needs.package-mcp.result }}
--- a/scripts/pr_preflight_test.go
+++ b/scripts/pr_preflight_test.go
@@ -1,3 +1,5 @@
+//go:build integration
+
 package scripts_test
 
 import (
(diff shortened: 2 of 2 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claim at the PR head). Verdict: **needs changes** — a single small should-fix; the approach is right.

**Should-fix — inherited `BASH_ENV` defeats the hermeticity the test exists to provide.** The test builds its subprocess env with `append(os.Environ(), ...)` (verified at the PR head) and never clears `BASH_ENV`. Non-interactive Bash sources `$BASH_ENV` *before* executing the `-c` body — i.e., before the fake `gh` function is defined — so host startup code can invoke the real `gh`, touch auth state, or change shell options ahead of interception. One line fixes it: append `BASH_ENV=` (and arguably `ENV=` for defensive portability) to `cmd.Env` alongside the Windows-required variables.

Everything else verified: every production call in `pr-preflight.sh` is an unqualified `gh`, so the function interception covers them; positional-arg sourcing is correct; merge-base drift is clean (main's copy of the test equals the merge-base blob).

ecuthiell:
Addressed the change request in `cade6876f91b8a92f92e0812fe99c9542cc18a56`.

The fixture no longer inherits host `BASH_ENV` or sources the production script into its wrapper. It launches the exact script as a real process, supplies the fake `gh` through a controlled `BASH_ENV`, curates shell/Git/GitHub/path state, fails rather than skips when required tools are absent, and wires real Linux/macOS/Windows process runs into the required CI aggregate. Focused normal/race tests, vet, lint, actionlint, and the hostile-environment falsifier pass locally; the new hosted matrix remains the pre-merge proof.

maphew (approved):
Approving — this supersedes my earlier changes-requested review (at old head 6048177). The BASH_ENV hermeticity hole is fully closed at head cade6876:

- Child env is built from scratch (no `os.Environ` inheritance; Windows-only whitelist of SYSTEMROOT/WINDIR/COMSPEC).
- `BASH_ENV` now points at the controlled fixture that defines the fake `gh`; `ENV=` is cleared.
- The new falsifier `TestPRPreflightFixtureIgnoresHostShellAndGitHubState` poisons BASH_ENV/ENV/BASHOPTS/SHELLOPTS/GH_CONFIG_DIR/GIT_CONFIG_GLOBAL and proves the hostile startup file never executes.

Verified by independent review + adversarial re-check: all 5 `gh` call sites in `scripts/pr-preflight.sh` remain same-shell calls covered by function interception; the shared helpers exist on main in the same package (no compile risk); checks green at head, MERGEABLE at 42 behind with no overlapping main changes. The move of these tests behind `//go:build integration` is a net coverage upgrade — the required 3-OS `pr-preflight-platforms` job is stronger than the previous skippable unit path — and the fail-closed `t.Fatalf` on missing tools is deliberate per gastownhall/beads#4967.

Merge tail goes to the pr-babysit patrol (squash, attribution preserved).

PR #4969, opened 2026-07-23
fix(plugin): make command argument hints strings

maphew:
## Why

Copilot loads the shared plugin command frontmatter as YAML and requires `argument-hint` to be a string. Seven commands currently decode a bracketed value as a sequence, so those skills fail to load. Several other multi-bracket hints are not valid yaml.v3 scalars at all, leaving the same schema vulnerable to future consumers and refreshes.

## What changed

- normalize all 23 shared command `argument-hint` fields to explicit YAML strings
- preserve each displayed hint, except the two no-argument commands whose reported `[]` values correctly become empty strings
- scan every shared command Markdown file in a regression test and assert all 23 hints decode as strings
- normalize CRLF before parsing so the test cannot silently skip a normal Windows checkout

The cleanup expands mechanically beyond the seven reported files because a truthful scan-all test exposed invalid legacy multi-bracket YAML in the remaining command set.

## Validation

- `go test ./cmd/bd/setup -run 'TestPlugin(CommandArgumentHintsAreStrings|LayoutUsesSharedBeadsRoot)' -count=1`
- `go test ./cmd/bd/setup -skip '^TestInstallAgentsInspectError$' -count=1`
- `git diff --check`
- repository commit hook: `0 issues`
- independent read-only review: no remaining blockers

The complete setup package has one pre-existing Windows-only failure in `TestInstallAgentsInspectError`: its expected inspect-error wording is replaced by a temp-file path error. That exact failure reproduces on a separate upstream-main-derived worktree and is unrelated to these frontmatter changes.

Closes #4494.

_codex-gpt-5.6-sol-high on behalf of matt wilkie_

PR #4975, opened 2026-07-23
docs(plugin): replace removed sync command guidance

### LINKED ISSUE #4856: plugin docs: sync.md calls `bd sync` a no-op — it is actually removed (unknown command)
## Affected file
`plugins/beads/skills/beads/commands/sync.md` (main)

## Problem
Minor wording bug: the doc says

> `bd sync` is **deprecated** and is now a no-op.

But the command is fully removed, not a no-op:

```
$ bd sync
Error: unknown command "sync" for "bd"
```

An agent told "it's a no-op" may still emit `bd sync` calls (harmless if truly no-op) and then hit hard errors instead.

## Suggested fix
Change wording to "has been removed" and keep the pointer to `bd dolt push` / `bd dolt pull`.

Environment: bd 1.1.0 (Homebrew), macOS arm64.

maphew:
## Why

The bundled plugin described top-level `bd sync` as a harmless no-op. The command has been removed, so agents following that guidance fail with an unknown-command error. The old replacement advice also did not guarantee that pending off/batch/server working-set changes reached the remote.

## What changed

- state that top-level `bd sync` is removed while unrelated nested sync commands remain
- document pull-before-work and commit→pull→push before handoff
- explain safe automatic Git-origin adoption when no Dolt remote exists
- reserve manual remote setup for a distinct, absent custom remote
- clarify that auto-commit is local and opt-in auto-push publishes committed HEAD only

## Validation

- checked root and nested command registration against current Cobra definitions
- checked commit/pull/push behavior in embedded and server modes
- checked remote adoption, collision guard, and auto-push behavior against source/tests
- Markdown fence and stale-command scans; `git diff --check`
- independent final review: no blockers

Related active PR #4969 touches only this file's frontmatter quoting; its hunk is separate.

Closes #4856.

### DIFF
--- a/plugins/beads/skills/beads/commands/sync.md
+++ b/plugins/beads/skills/beads/commands/sync.md
@@ -1,17 +1,50 @@
 ---
-description: Synchronize issues (deprecated — use bd dolt push/pull)
-argument-hint: (deprecated)
+description: Synchronize issues with a configured Dolt remote
+argument-hint: ""
 ---
 
-`bd sync` is **deprecated** and is now a no-op.
+The top-level `bd sync` command has been removed. Running it returns an
+unknown-command error; it is not a compatibility no-op. Nested commands such
+as `bd backup sync` and tracker-specific sync commands are separate commands
+and remain supported.
 
-## Use Dolt commands instead
+## Supported workflow
 
-- **Push to remote**: `bd dolt push`
-- **Pull from remote**: `bd dolt pull`
-- **Commit pending changes**: `bd dolt commit`
-- **Check connection**: `bd dolt show`
+```bash
+# Before starting work or consuming changes from another clone
+bd dolt pull
+
+# After local issue updates and before handoff: commit, integrate, then publish
+bd dolt commit
+bd dolt pull
+bd dolt push
+```
+
+`bd dolt commit` creates an explicit commit boundary when auto-commit is off or
+in batch mode; it is a safe no-op when there is nothing pending. The following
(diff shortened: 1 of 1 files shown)


maphew (commented):

This is a docs-only change to plugins/beads/skills/beads/commands/sync.md replacing outdated "bd sync is a no-op" guidance with accurate replacement guidance. I verified every factual claim against cmd/bd/dolt.go, cmd/bd/dolt_autocommit.go, and cmd/bd/dolt_autopush.go: top-level `bd sync` is indeed gone (no such Use string registered on rootCmd), nested commands like `bd backup sync` remain, `bd dolt commit` prints "Nothing to commit." when there's nothing pending, auto-commit is local-only, auto-push is opt-in via dolt.auto-push (default false) and only pushes already-committed HEAD (via GetCurrentCommit, no commit call), and adoptGitOriginRemoteForPush only fires when zero Dolt remotes exist and a git origin is present. No scope creep (single file, matches PR's file list) and no tests needed for a docs change.

**Findings:**
- **nit** `plugins/beads/skills/beads/commands/sync.md:2` — argument-hint is set to an empty string ("") for a no-argument command. Other no-arg commands in this directory use "[]" (e.g. quickstart.md, blocked.md) rather than ""; for consistency with the local convention this should probably be `argument-hint: []`.

Review performed by a claude-sonnet-5 reviewer subagent as part of an open-PR review sweep; findings will be triaged by the maintainer before any merge.

PR #4979, opened 2026-07-23
fix(scripts): remove Python from doc freshness checks

### LINKED ISSUE #4978: Doc freshness check misreports a broken Python alias as an invalid review date
## Summary

`scripts/check-doc-freshness.sh` reports valid `Last reviewed` markers as invalid dates when `python3` exists on `PATH` but is not a working interpreter. A default Windows installation can have exactly that state: the Microsoft Store app-execution alias resolves as `python3.exe` but exits with an installation prompt.

Python is not documented as a prerequisite for the freshness check, and the script already depends on the platform `date` utility to determine today's date.

## Environment

- Beads: `main` at `02e38d0bdbd311d2113abbdb9b816a9308d79d93`
- Windows 11 Pro, build 26200
- Git 2.54.0.windows.1 with Git Bash
- `python3.exe`: Microsoft Store alias under `%LOCALAPPDATA%\Microsoft\WindowsApps`, with no Python runtime installed

## Reproduction

On that host:

```text
> python3 --version
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Apps > Advanced app settings > App execution aliases.
```

Run the freshness check with a deterministic current date:

```bash
DOC_FRESHNESS_TODAY=2026-07-22 ./scripts/check-doc-freshness.sh
```

The checked-in marker `Last reviewed: 2026-07-10` is valid and current, but the script reports:

```text
FAIL: invalid Last reviewed date: 2026-07-10
```

The same false diagnostic is emitted for every otherwise valid reviewed date. A poisoned `python3` shell function that exits nonzero reproduces the behavior deterministically, without relying on the Windows alias.

## Root cause

`date_age_days` invokes `python3` directly. Its caller maps every nonzero result from that function to `FAIL: invalid Last reviewed date`, so interpreter discovery and startup failures are indistinguishable from malformed document data.

## Expected result

A missing or nonfunctional optional interpreter must not be reported as invalid documentation data. Preferably the check should not require Python at all when its existing `date` dependency can perform the calculation portably.

## Proposed fix

Replace the inline Python calculation with a small, strict GNU/BSD `date` adapter:

- Git Bash and Linux use `date -d`;
- macOS/BSD use `date -j -f`;
- both parse dates at UTC midnight and return the epoch plus canonical `YYYY-MM-DD`;
- the canonical value must exactly round-trip to the input, so a parser cannot silently normalize an impossible date such as `2026-02-30`; and
- the existing invalid, future, and stale diagnostics remain unchanged.

Add process-level regression tests that poison `python3`, verify a valid 16-day age calculation succeeds without invoking it, preserve the malformed, future, and stale date diagnostics, force the BSD fallback, and prove that successful-but-normalized results from both GNU and BSD forms are rejected. On Windows, derive Bash from the selected Git for Windows installation and probe its MSYS environment instead of relying on the unrelated `bash.exe` app alias on `PATH`.

---

ecuthiell:
## What

Remove the undocumented Python dependency from `scripts/check-doc-freshness.sh` while preserving Python's calendar contract:

- validate strict `YYYY-MM-DD` values over the proleptic Gregorian `0001-01-01` through `9999-12-31` domain in Bash;
- compute monotonic day numbers with integer arithmetic, including Gregorian century/leap rules;
- use native `date` only to supply today's value when `DOC_FRESHNESS_TODAY` is absent;
- distinguish a failing or malformed native provider from an invalid explicit override; and
- keep malformed, future, and stale document diagnostics unchanged.

The real-process tests carry the `integration` build tag and run under an exact `/usr/bin:/bin` child PATH with Python poisoned. A required Linux/macOS/Windows matrix exercises calendar boundaries, native default-TODAY success, provider failure, provider success with malformed output, invalid overrides, and Git-Bash discovery; missing prerequisites fail instead of skipping.

## Why

Fixes #4978.

Windows can expose a Microsoft Store `python3.exe` app-execution alias even when no Python runtime is installed. The old script mapped any interpreter failure to an invalid `Last reviewed` value, turning a tooling failure into a misleading documentation-quality error.

An intermediate GNU/BSD `date` parser passed Windows but the required macOS lane proved that native BSD `date` does not cover the full year 1–9999 domain. The final design keeps platform date parsing out of validation entirely and needs no new runtime dependency.

## Verification

Rebased onto upstream `1bbca0c1fa545b990a03e2270b75158facef7a50`; exact head: `9c2b48767e7a9ec3f72018c8d6b0aa7c214429b4`.

Local exact-head checks passed:

```text
go test '-tags=integration,gms_pure_go' -count=1 -run '^TestDocFreshness' ./scripts
go vet '-tags=integration,gms_pure_go' ./scripts
shellcheck scripts/check-doc-freshness.sh
actionlint .github/workflows/pr.yml
git diff --check upstream/main...HEAD
```

Two independent adversarial reviews verified the calendar function against .NET's independent proleptic-Gregorian implementation through era, century, leap, and domain boundaries; reviewed patch identity remained byte-for-byte stable across the final unrelated upstream rebase.

The fresh exact-head hosted Linux/macOS/Windows matrix and `CI Gate / Required` remain stop conditions before merge.

---


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — one should-fix; the shell rewrite itself is sound and this is close to landing.

**Should-fix — the new test is an integration test running in the core suite.** `scripts/check_doc_freshness_test.go` launches real Bash subprocesses against temp-file fixtures, but carries no `//go:build integration` tag (verified at the PR head), so it runs under `ci-pr-core`'s `go test -short ./...`. The repo's testing philosophy classifies real filesystem/process tests as integration-gated. Tag it (and wire it into an appropriate gate) or provide lighter unit coverage for the parsing helpers.

Nits: the repository-root/MSYS-path helpers duplicate ones already in the same `scripts_test` package (`release_script_test.go`) — reuse avoids divergent Windows path behavior; and GNU date accepts year `0000` where the old Python `datetime.date` rejected it — explicitly reject `0000-*` if the prior diagnostic is worth preserving (confirmed on GNU date only).

Otherwise verified: GNU/BSD parsing + canonical round-trip validation, Python-free age calculation, both-backend test coverage, and Git Bash discovery all check out; merge-base drift is clean; `bash -n`, ShellCheck, and `gofmt` pass.

ecuthiell:
Addressed the requested integration boundary and the two review nits in `1f07f82db4b2a342ecf1cf1ac3d42601aec7c3d3`:

- the real-process tests now carry `//go:build integration` and run in a required Ubuntu/macOS/Windows matrix wired into `CI Gate / Required`;
- the fixture reuses the package's existing repository-root and MSYS-path helpers;
- year `0000` is rejected, while leap-day and the supported `0001-01-01` / `9999-12-31` boundaries are locked down;
- the script preflights GNU/BSD date capabilities and reports an unavailable backend as a tooling error rather than invalid document data; and
- every Bash/Git-Bash/date prerequisite in the required process suite now fails closed. A curated Windows PATH with no `git.exe` exits nonzero instead of producing an all-skipped green run.

The exact head passed the focused integration suite, `actionlint`, diff hygiene, and an independent adversarial re-review. The hosted three-platform matrix is running and remains required before merge.

ecuthiell:
The first required platform run exposed a real portability flaw in the intermediate revision: Windows passed, but macOS's native BSD `date` could not satisfy the preserved year 1–9999 domain. I replaced platform epoch parsing rather than weakening that contract.

Exact head `9c2b48767e7a9ec3f72018c8d6b0aa7c214429b4` now validates and computes proleptic-Gregorian day numbers in pure Bash. Native `date` owns only default-TODAY acquisition, with separate diagnostics for provider failure, provider success with malformed output, and an invalid explicit override. The required process suite now exercises a successful real native provider on every host, fails closed on missing prerequisites, and runs the child under exact `PATH=/usr/bin:/bin`.

The exact head passed the focused process suite, vet, ShellCheck, actionlint, diff hygiene, and two independent adversarial reviews, including comparison against .NET's calendar oracle. The final rebase changed only an unrelated upstream storage test and preserved the reviewed patch byte-for-byte.

A fresh Linux/macOS/Windows matrix and `CI Gate / Required` remain mandatory before merge.

maphew (approved):
Approving — this supersedes my earlier changes-requested review (pinned to old commit 7752eb7), whose three items are all verified resolved at head 9c2b487:

1. **Integration gating**: `scripts/check_doc_freshness_test.go` now carries `//go:build integration`, and the new required `check-doc-freshness-platforms` ubuntu/macos/windows matrix is wired into the CI gate — all three lanes green.
2. **Helper duplication**: the test reuses `sourceRepoRoot`/`msysPath` from `release_script_test.go` instead of redefining them.
3. **Year-0000**: platform date parsing is gone entirely; pure-Bash proleptic-Gregorian validation enforces year 1–9999, with an explicit year-zero FAIL test.

Verified by independent review + adversarial re-check (branch content, CI state, and main-drift: zero overlapping commits on main since the base). Two non-blocking notes for later, not this PR: the `freshnessDocuments` fixture duplicates the script's DOCS array (a drift fails loudly, so acceptable), and the 3-OS matrix could gain a paths filter if runner cost ever matters.

Merge tail goes to the pr-babysit patrol (squash, attribution preserved). Thanks for the thorough turnaround, ecuthiell.

maphew:
Maintainer update: merged current `main` into this branch to resolve the `.github/workflows/pr.yml` conflict.

The conflict was two independent job additions colliding textually: this PR's new `check-doc-freshness-platforms` matrix job, and `pr-preflight-platforms` which landed on `main` after the branch point. Both jobs now coexist, each wired into the PR baseline gate (`needs`, `CI_GATE_REQUIRED`, env mapping). While in there I bumped this PR's job to the same `actions/checkout` / `actions/setup-go` pins `main` now uses, so the pins stay consistent post actions-group bump.

`TestDocFreshness` and `TestPRPreflight` integration suites both pass locally on Linux after the merge. No changes to the script or test content of this PR.

PR #4984, opened 2026-07-23
fix(create): reject caller input that would be discarded

### LINKED ISSUE #4983: bd create silently discards extra positional args and explicit empty external input
## Summary

bd create accepts input that it silently discards in two closely related fail-open paths:

1. A second positional argument is ignored, so bd create bug REAL_TITLE creates or previews an issue titled bug and drops REAL_TITLE.
2. An explicitly supplied empty --body-file or stdin exits successfully and creates or previews an issue with no description, indistinguishable from intentionally omitting the description input.

Both are dangerous in automation because the command reports success after losing caller-supplied data.

## Reproduction

Tested with bd 1.1.0 on Windows PowerShell and confirmed in current main source.

```powershell
bd create bug SECOND_TITLE -d x --dry-run --json
# exit 0; JSON title is bug; SECOND_TITLE is absent

New-Item empty.md -ItemType File
bd create EMPTY_BODY --body-file empty.md --dry-run --json
# exit 0; description is absent
```

## Expected

- create accepts at most one positional title and exits nonzero before any write when more are supplied.
- Explicit empty stdin or file input exits nonzero by default with actionable guidance.
- Intentionally omitting a description remains valid when create.require-description is disabled.
- An explicit opt-in may preserve the rare intentional empty-file case.

## Root cause

Current cmd/bd/create.go declares cobra.MinimumNArgs(0), while both the normal and proxied input paths consume only args[0]. The create paths also call getDescriptionFlag without the existing validateDescriptionUpdate guard used by update.

## Proposed fix

- Use cobra.MaximumNArgs(1).
- Apply the existing external-description guard in both normal and proxied create paths.
- Register --allow-empty-description for intentional empty external input.
- Add regression tests and preserve zero/one positional title compatibility.

davevan2:
## What

Make create fail closed when caller-supplied input would otherwise be silently discarded:

- reject more than one positional title before RunE
- reject explicit empty stdin or file descriptions by default in both normal and proxied create paths
- preserve intentional empty external input behind --allow-empty-description
- preserve zero-title flag usage, one positional title, and intentionally omitted descriptions

## Why

Today bd create bug REAL_TITLE silently keeps bug as the title and discards REAL_TITLE. Explicit empty --body-file input similarly succeeds as though no description input was supplied. Both produce success-looking automation after data loss.

Fixes #4983.

## Verification

```text
go test -tags gms_pure_go ./cmd/bd -run Test(CreateCommand|GatherCreateInputRejectsEmptyBodyFile|ValidateDescriptionUpdate) -count=1
ok github.com/steveyegge/beads/cmd/bd
```

Isolated proxied-store CLI proof:

- extra positional: exit 1, accepts at most 1 arg
- empty body file: exit 1 with --allow-empty-description guidance
- empty body with opt-in: exit 0
- omitted description: exit 0
- persisted issue count after all dry runs: 0

### DIFF
--- a/cmd/bd/create.go
+++ b/cmd/bd/create.go
@@ -46,7 +46,7 @@ var createCmd = &cobra.Command{
 	GroupID:       "issues",
 	Aliases:       []string{"new"},
 	Short:         "Create a new issue (or batch from markdown/graph JSON)",
-	Args:          validateCreateArgs,
+	Args:          cobra.MatchAll(cobra.MaximumNArgs(1), validateCreateArgs),
 	SilenceUsage:  true,
 	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
@@ -124,9 +124,12 @@ var createCmd = &cobra.Command{
 			}
 		}
 
-		description, _, err := getDescriptionFlag(cmd)
+		description, descriptionChanged, err := getDescriptionFlag(cmd)
 		if err != nil {
 			return err
+		}
+		if err := validateDescriptionUpdate(cmd, description, descriptionChanged); err != nil {
+			return HandleError("%v", err)
 		}
 
 		skills, _ := cmd.Flags().GetString("skills")
@@ -782,6 +785,7 @@ func init() {
 	createCmd.Flags().String("mol-type", "", "Molecule type: swarm (multi-agent), patrol (recurring ops), work (default)")
 	createCmd.Flags().String("wisp-type", "", "Wisp type for TTL-based compaction: heartbeat, ping, patrol, gc_report, recovery, error, escalation")
 	createCmd.Flags().Bool("validate", false, "Validate description contains required sections for issue type")
+	createCmd.Flags().Bool("allow-empty-description", false, "Allow empty description input from stdin or file")
 	// Event-specific flags (only valid when --type=event)
 	createCmd.Flags().String("event-category", "", "Event category (e.g., patrol.muted, agent.started) (requires --type=event)")
 	createCmd.Flags().String("event-actor", "", "Entity URI who caused this event (requires --type=event)")
--- a/cmd/bd/create_dispatch_test.go
+++ b/cmd/bd/create_dispatch_test.go
@@ -1,0 +1,173 @@
+//go:build cgo
+
+package main
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// TestCreateDirectModeRejectsAllowEmptyDescriptionWithFile is a regression
+// test for the 2026-07-23 CHANGES_REQUESTED review on gastownhall/beads#4984:
+// direct-mode (non-proxied) --file dispatch returned from RunE before the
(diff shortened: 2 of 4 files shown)


maphew (changes requested):
Cross-vendor agent review (Codex gpt-5.6-sol primary trace; Claude adjudication re-verified the load-bearing claims at the PR head). Verdict: **needs changes** — right goal, but one door still open and thin coverage.

**Should-fix — direct-mode batch paths silently ignore the new flag.** Verified at the PR head: `gatherCreateInput` (where `--allow-empty-description` is declared single-issue-only and rejected in batch) runs only on the proxied route; direct mode's `--file`/`--graph` dispatch returns from RunE before the local validation at `create.go:124-130`. So `bd create --file plan.md --allow-empty-description` accepts-and-ignores in direct mode while proxied rejects it — the same silent-discard class this PR exists to eliminate, and the same route split just flagged on #4918. Hoisting a shared single-issue-only flag rejection above the mode/batch dispatch would fix both PRs' gap in one place.

**Should-fix — regression coverage stops at flag registration + proxied rejection.** No test covers direct execution, a successful `--allow-empty-description` create, the error text, or batch parity — which is exactly where the gap above hides.

Nit: `validateDescriptionUpdate` is now called from create; a create-neutral name (`validateDescriptionInput`) would stop the next reader from assuming update-only semantics.

Wiring otherwise checks out (`getDescriptionFlag`, the guard helper, cobra arg validation before store open); merge-base drift is clean.

maphew (approved):
Approving — supersedes my 2026-07-23 changes-requested review; the fixes are applied as maintainer commits on this branch. Direct-mode `--file`/`--graph` now reject `--allow-empty-description` (the silent-discard gap the review flagged — the proxied path already rejected it), with rejection tests for both routes plus a success round-trip for the opt-in, verified running against the containerized test server. Your original single-issue validation ships unchanged; thanks davevan2 — the input-discard class this PR targets is exactly right, and the broader flag-hoist for batch routes landed separately via #4918's fix-merge.

Validation: build/vet/gofmt clean, Create test set green; independent review + cross-vendor Codex gate (two rounds) signed off. Merge tail goes to the pr-babysit patrol.

maphew:
Maintainer rebase onto current main (was conflicting). Two adaptations were needed because main evolved past this branch:

- `Args` is now `cobra.MatchAll(cobra.MaximumNArgs(1), validateCreateArgs)` — main moved title validation into the `Args` phase (validateCreateArgs), and this PR's arg-count cap is composed with it rather than replacing it.
- The narrow `--allow-empty-description is not valid with --file/--graph` checks from the second commit are now covered by main's generalized `singleIssueOnlyFlags` rejection (this PR's flag is registered in that list), so the inline checks resolved away.
- `TestCreateCommandAllowsZeroOrOnePositionalArgument` was adapted: zero args now fails title resolution at `Args` time by design on main, so the test asserts that error instead of acceptance.

What survives is the substance: the `--allow-empty-description` opt-in flag on `create`, `validateDescriptionUpdate` wiring on both the direct and proxied-server paths, and the dispatch/input test coverage. Authorship preserved via rebase. Targeted tests green locally (`go test -run 'TestCreateCommand|TestGatherCreateInput|TestDispatch' ./cmd/bd/`).

PR #4985, opened 2026-07-23
fix: keep preview commands from migrating stores

davevan2:
## What

Treat command-local `--dry-run` and `--inspect` modes as true non-mutating previews. Embedded preview opens now skip schema initialization, refuse writes, and tolerate a behind-but-query-compatible schema cursor so validation can run without repairing the database.

Add regressions proving `create --dry-run` leaves the schema version, Dolt HEAD, and issue count unchanged, and that explicit `-C` targeting wins over an inherited `BEADS_DIR`.

## Why

Preview commands were identified as mutating commands until their `RunE` handlers executed. Before that point, version-bump auto-migration and embedded store initialization could migrate and commit the selected database—even when the operation was only a dry run.

## Verification

- Focused command and `-C` unit tests pass with `gms_pure_go`.
- Embedded dry-run immutability and `-C` target regressions pass with CGO.
- Existing create/migrate dry-run and inspect suites pass.
- Embedded read-only store tests pass.
- Explicit `migrate schema` regression passes.
- Patch-scoped `golangci-lint` reports 0 issues.
- Full `make test` was attempted; the Windows run remains red on pre-existing platform failures. A representative embedded hook failure reproduces unchanged at upstream `fcdcae1ce`.

_codex-unknown-model-unknown-reasoning on behalf of davevan2_


maphew (changes requested):
Cross-vendor paired agent review (Claude Opus + Codex gpt-5.6-sol, independently traced, findings adjudicated and re-verified against the PR head before posting). Verdict: **needs changes** — the core mechanism is right, but two paths dodge it.

The central guard is correctly placed and we verified it: `previewMode` is computed in root `PersistentPreRunE` before any store opens, `--force`+preview is rejected pre-open, `autoMigrateOnVersionBump` is skipped for previews, and the main store opens via the read-only preview path — exactly the placement the #4654 lesson requires for anything that must influence store-open behavior. The embedded and ordinary server paths are solid.

**Blocker 1 — proxied-server previews still migrate.** `cmd/bd/main.go` (proxied branch, ~line 1269) short-circuits to `newProxiedServerUOWProvider` + `reconcileVersionProxiedServer` with no `previewMode` gating. The provider chain (`uow_factory.go` → `doltserver_provider.go`/`external_doltserver_provider.go` → `dolt_sql_provider.go`) runs `openAndInitSchema` → `CREATE DATABASE` + `schema.MigrateUpWithLock`. So `create/delete --dry-run` in proxied mode still migrates during root pre-run — the exact side effect this PR exists to prevent, on the mode where it's least visible.

**Blocker 2 — cross-repo dry-run parent lookup opens the target writable.** `openDryRunTargetStore` (`cmd/bd/create.go` ~815–845) documents read-only intent ("Do not add writes here") but passes the writable `newDoltStoreFromConfig` factory on both the cached-remote and local-target paths, which runs schema init on the *target* repo (and the factory path can rename a legacy database / rewrite metadata). `create --dry-run --parent X --repo <other>` therefore mutates the other repo's store. The new regression test uses neither `--parent` nor `--repo`, so it can't catch this.

**Should-fix A — a preview consumes the one-shot upgrade signal.** `trackBdVersion()` runs before the preview check and rewrites `.beads/.local_version` immediately on version change. A preview command run first after an upgrade skips `autoMigrateOnVersionBump` (correct) but has already burned the version-change marker, so the next ordinary command sees a matching `.local_version` and never runs the version-bump reconciliation (including the pre-0.56 recovery path). Defer the `.local_version` write when `previewMode`, or persist a pending-upgrade marker. A two-invocation regression test (preview first, ordinary second) would pin it.

**Should-fix B — PostRun tip auto-commit errors on the preview store.** The tip-metadata auto-commit in PersistentPostRun (`main.go` ~1418–1442) is deliberately not gated by read-only classification, and read-only commands tolerate it because `OpenForReadOnlyCommand` is intentionally writable for incidental writes (`embeddeddolt/cache.go` comment). The strict write-refusing preview store breaks that: with `dolt.auto-commit=on` and an eligible tip (human, non-`--json` usage), `SetLocalMetadata` returns `errReadOnly` and the command exits non-zero *after* a successful dry-run. Smallest fix: treat `errReadOnly` as non-fatal there, consistent with `recordTipShown`'s own "ok to fail silently" contract.

**Nit:** `TestEmbeddedChangeDirOverridesInheritedBeadsDir` tests `-C` vs inherited `BEADS_DIR` with no corresponding production change in this PR — worth splitting out.

Relaxing `CheckBehindDrift` for previews (raw SQL error instead of the friendly behind-drift message in the post-upgrade window) reads as a considered, acceptable tradeoff and is documented in the code.

Thanks for attacking this — the migrate-at-open trap is real (it burned a prior `--force`/`--dry-run` combination in #4654), the embedded/server fix is the hard part and it's done right here. With the proxied branch and the dry-run target factory routed through the same preview policy, this lands.

maphew (approved):
Maintainer fix-merge, per the review timeline (changes requested 2026-07-23; no activity since, head unmoved). Your commit is untouched at the base of the branch — everything below is additive on top, so feel free to amend or push back on any of it.

What landed:

- **Merge of current `main`** (`34450e851`): `main` gained the `rootStorePolicy` split since this PR was written, reorganising exactly the lines the PR touched. Resolved by composition — `useReadOnly := policy.readOnly || previewMode`, and version-bump maintenance runs only when `policy.runMaintenance && !previewMode` — so strict `--readonly` and preview each keep their own effect.
- **Blocker 1** (`ced481612`): proxied-server previews now route through the preview policy via a variadic `uow.ProviderOption` seam (`uow.WithPreview()`); `initSchema` is USE-only under preview (no `CREATE DATABASE`, no `MigrateUpWithLock`), and `reconcileVersionProxiedServer` is skipped. Hermetic sqlmock tests cover the provider; `previewProviderOptions` (`338c5ec55`) pins the CLI wiring.
- **Blocker 2** (`61f54f205`): `openDryRunTargetStore` uses a new non-mutating `newPreviewStoreFromConfig` on both the cached-remote and local-target paths. `TestEmbeddedCreateDryRunCrossRepoDoesNotMigrateTarget` fails before the fix (target schema version and Dolt HEAD both moved) and passes after.
- **Should-fix A** (`79cd7b8c6`): previews no longer consume the one-shot `.local_version` marker; two-invocation regression test included, verified failing pre-fix.
- **Should-fix B** (`8c8bf3de6`): the PostRun tip auto-commit treats a read-only refusal as non-fatal, consistent with `recordTipShown`'s fail-silent contract.

Two notes for the record:

- The proxied preview provider suppresses **open-time** mutations (the bug here); the returned provider is otherwise ordinary. The embedded preview store refuses writes for its whole lifetime — the asymmetry is deliberate and scoped to what this PR is about.
- `TestConfigValidateReadOnlyIsHermetic` fails on the merge base itself (reproduced on clean `upstream/main` — a `dolt.gate.lock` file left by strict-readonly `bd config validate`, workspacegate #5046 territory). If that job is red here, it is pre-existing, not this branch.

Validation: cgo and nocgo builds clean, `go vet` clean, targeted `DryRun|Preview`, uow, embeddeddolt and version/factory suites green; full `make test` queued in our local verification lane. Approving — the migrate-at-open trap this PR closes is real, and with the proxied branch and dry-run target factory now under the same policy it is closed everywhere.

Issue #4249, opened 2026-05-29
Proposal: a build-tagged zstd seam so embedded Dolt can build without cgo (pure-Go bd is proven)


maphew:
## Summary

_Right now, building bd with embedded Dolt requires CGO - a C compiler toolchain - and that single fact causes the recurring weekly pain: wrong build incantations, doc drift, "requires CGO" runtime errors, a complicated release matrix. Work in PR #4408 proved the whole problem traces to one C library (a zstd compression dependency deep inside Dolt) and that a pure-Go replacement works._

Beads' entire cgo maintenance burden traces to a **single** transitive
dependency: `github.com/dolthub/gozstd` (a cgo wrapper over libzstd), imported by
exactly one package, `github.com/dolthub/dolt/go/store/nbs`. Once
`-tags=gms_pure_go` removes the other C dependency (`go-icu-regex`), `gozstd` is
the **only** cgo dependency left in a full `bd` binary.

That one dep forces the `//go:build cgo` / `!cgo` split that ripples through
beads — embedded-storage stubs, the gated `federation` command, ~100 cgo-only
test files, nondeterministic doc generation (#4203), the pre-commit typecheck
trap (#3402), and a split release matrix where android/Termux ships without
embedded mode (#3538).

**I built a pure-Go `bd` and it works.** Replacing `gozstd` with a ~120-line shim
over `github.com/klauspost/compress/zstd` (already an indirect dep) and
decoupling beads' code from the `cgo` build tag yields a statically-linked,
cgo-free `bd` that builds and **runs embedded Dolt** end to end.

This issue asks whether Dolt would accept a **build-tagged seam** in `store/nbs`
so this can be done cleanly upstream — defaulting to cgo/gozstd so prebuilt
binaries and big-data Dolt users are completely unaffected.

## Evidence (reproducible)

In a worktree, `replace github.com/dolthub/gozstd => ./gozstd-shim` (klauspost
backend) plus flipping beads' `//go:build cgo/!cgo` tags to a CGO-independent
`nocgo` tag (reproducing today's CGO=1 file set), then at
`CGO_ENABLED=0 -tags gms_pure_go`:

- `go build ./cmd/bd` → exit 0; `file bd` → **statically linked**, `ldd` → *not a
  dynamic executable* (no libzstd / ICU / libc).
- `bd init` → `Backend: dolt, Mode: embedded`; `bd create` / `bd list`
  round-trip issues. **Embedded Dolt runs pure-Go.**
[...]

PR #4986, opened 2026-07-23
Use Snappy output for Dolt garbage collection

maphew:
Pass archive level 0 through every Beads-owned GC path so new table files use Snappy instead of zstd archives. Existing zstd archive decoding remains unchanged.

Refs #4249.

## Why

Beads does not benefit enough from zstd archive compression to justify its costs. Zstd is the last dependency forcing embedded Dolt builds through CGO, complicating local builds, releases, cross-compilation, and platform support.

Testing also showed that replacing the zstd encoder with klauspost is unsafe: its trained dictionaries provide little benefit over Snappy and can produce archives that existing CGO/libzstd builds reject with a dictionary mismatch.

Dolt already provides a compatible alternative. Archive level `0` writes classic Snappy-compressed table files. Snappy is pure Go, interoperable, and sufficient for typical Beads database sizes. Existing zstd archives remain readable; this change only stops Beads-triggered garbage collection from creating new ones.

This is the safe first step toward pure-Go embedded builds: remove zstd from the write path now, then retain decode-only support for existing zstd archives during the remaining Dolt-side work.

Agent-Signature: amp-unknown-model-unknown-reasoning on behalf of matt wilkie
Amp-Thread-ID: https://ampcode.com/threads/T-019f8d27-e480-774e-bd5a-7051e8e6e378


maphew (commented):
**Paired agent review — verdict: concerns** (claude-opus-4-8 and codex gpt-5.6-sol reviewed independently; findings merged, with the contradiction between them resolved by direct verification against the pinned Dolt source)

The core mechanism is correct: `CALL DOLT_GC('--archive-level', '0')` parses via `CreateGCArgParser` in the pinned Dolt (`v0.40.5-0.20260715172757`), and level 0 maps to `chunks.NoArchive`, i.e. classic Snappy table files, exactly as the PR claims. All in-process explicit GC paths now funnel through the single `versioncontrolops.DoltGC` helper, the decode/zstd read path is untouched, and no unrelated changes were found. Two majors below need addressing before this fully delivers "no new zstd archives from Beads-owned GC".

**Findings:**
- **major** `internal/storage/dbproxy/server/doltserver.go:236` (also `internal/doltserver/doltserver.go` launcher) — Beads-managed `dolt sql-server` processes inherit Dolt's auto-GC defaults: `DefaultAutoGCBehaviorEnable = true` with `DefaultCompressionLevel = 1` (servercfg/serverconfig.go:64,81 in the pinned module), and neither Beads launcher sets `auto_gc_behavior.archive_level: 0` in the generated YAML config. Background auto-GC on managed servers therefore continues to create zstd archives, contradicting the "every Beads-owned GC path" goal. Configure both server-launch paths (found by codex; independently confirmed against the pinned Dolt source - the other reviewer's claim that no auto-GC path exists was incorrect).
- **major** `cmd/bd/compact.go:812` — the external-subprocess path runs `exec.Command("dolt", "gc", "--archive-level", "0")` against whatever `dolt` binary is on PATH, with no version guarantee (unlike the in-process paths pinned by go.mod). On a host whose external dolt predates the flag, compact now aborts with an unknown-flag error where it previously worked. Suggest detecting the unknown-flag error and retrying with plain `dolt gc`, or probing `dolt version` before appending the flag; at minimum document the minimum external Dolt version (found by opus).
- **minor** `internal/storage/versioncontrolops/gc_test.go:18` — the test asserts only the exact SQL string via sqlmock (parameter passing), not that GC output actually avoids archive files; there is also no coverage for the compact.go subprocess flag or the doltserver shutdown path now routing through the helper. Both reviewers flagged this independently. An embedded-store integration assertion that GC produces no archive table files would carry the behavioral claim.
- **nit** `internal/storage/dbproxy/server/doltserver.go:377` — error wrapping text changed from `dolt_gc: %w` to the helper's `dolt gc: %w`; harmless, noted so it is not mistaken for lost context.

maphew:
Addressed both majors on top of the existing commit (4fafc043d):

**Major 1 (managed servers auto-GC into zstd)** — 775d207b3

- `internal/storage/dbproxy/server` already generates a `config.yaml` (via `cmd/bd/proxied_server.go`'s `renderProxiedServerConfig`); it now sets `auto_gc_behavior.archive_level: 0` there, auto-GC left enabled.
- `internal/doltserver`'s legacy launcher turned out to have no YAML config path at all — it invokes `dolt sql-server -H -P --loglevel` purely via CLI flags, and Dolt's `commandLineServerConfig.AutoGCBehavior()` is a stub that hardcodes `archive_level` to the zstd default regardless of any flag, on any Dolt version. There is no CLI-flag way to express this. It now optionally switches to `--config <generated yaml>` (Dolt's own docs: `--config` makes it ignore all other `sql-server` flags, so host/port/log-level move into the YAML alongside `archive_level`).
- Both paths gate on a new `internal/doltserver.SupportsArchiveLevelConfig(doltBin)`, which runs `dolt version` and fails closed (omits the key, unchanged prior behavior) on any parse/exec error. This is not cosmetic: Dolt's YAML loader uses `yaml.UnmarshalStrict` (`servercfg/yaml_config.go`), so an older external dolt binary whose own compiled-in struct predates `auto_gc_behavior.archive_level` (`minver:"1.52.1"`) would hit a hard parse error and refuse to start, not silently ignore the key — confirmed by reading the pinned module source rather than assuming. Existing `config.yaml` files are left untouched; only fresh bootstraps set the key.

**Major 2 (external `dolt gc --archive-level 0` regression)** — 3e121f834

`runCompactDolt` now retries with plain `dolt gc` when the first attempt's output matches a new `isUnknownArchiveLevelFlagError` classifier, printing a one-line notice that the older dolt may still produce zstd archives. The classifier requires both an "unknown flag"-style phrase and the flag name in the output, so a genuine GC failure is never misclassified as a missing flag and swallowed. Verified the exact Dolt argparser message (`error: unknown option \`archive-level'`) by running a local dolt build against the flag.

**Minor (gc_test.go only asserts the SQL string)** — not changed. `internal/storage/versioncontrolops` has an established, documented precedent (see the comment atop `fastforward_test.go`, from a prior PR) that this package intentionally uses sqlmock only and defers a live embedded-Dolt harness to a separate "piece-2 integration tests" effort, since none exists here and building one would invent new test architecture out of scope for this fix. Kept consistent with that precedent rather than introducing a one-off harness.

**Nit (error-wrap text change)** — left as noted (harmless).

Validation: `CGO_ENABLED=1 go build -tags gms_pure_go ./...`, `gofmt -l` clean on touched dirs, `go vet -tags gms_pure_go ./cmd/bd/... ./internal/storage/... ./internal/doltserver/...` clean, and targeted `go test` passes for `internal/doltserver`, `internal/storage/versioncontrolops`, `internal/storage/dbproxy/server`, and the new/touched `cmd/bd` tests (compact, proxied_server). `TestCompactSuite` skips locally (no Docker-backed Dolt server available in this sandbox), unrelated to this change.

maphew (commented):
**Paired agent delta review of the two fix commits — verdict: concerns** (claude-opus-4-8 and codex gpt-5.6-sol independently; contradiction adjudicated against the pinned Dolt source)

Much of the delta verifies clean on both sides: the YAML shape is correct with auto-GC left enabled, the 1.52.1 version gate matches the `minver` on `archive_level` exactly and fails closed with robust parsing, the unknown-option classifier matches Dolt's real argparser message and cannot loop, real GC failures still propagate, the tests round-trip through Dolt's own config loader, and the Amp base commit is untouched with clean attribution. Two majors remain, both found by codex; the second was verified directly against the pinned source after the opus reviewer assessed the modes as equivalent.

**Findings:**
- **major** `internal/doltserver/doltserver.go:793` — Dolt's `setupDoltConfig` (cmd/dolt/commands/sqlserver/sqlserver.go:532-590 in the pinned module) returns immediately when `--config` is passed; the parent-directory `../.doltcfg` discovery (privileges.db users, branch_control.db) runs only in flag mode. A deployment whose legacy server previously discovered a parent `.doltcfg` would, on a new-enough dolt, silently start with a fresh `$data_dir/.doltcfg` - existing users and branch controls abandoned, new passwordless root. The generated YAML must replicate the flag-mode discovery: set `cfg_dir` explicitly (parent `../.doltcfg` when it exists relative to the data dir, else `$data_dir/.doltcfg`, and surface the both-exist error case), and add a test covering the parent-`.doltcfg` layout.
- **major** `cmd/bd/proxied_server.go:117` — pre-existing configs bypass the new plumbing in both directions: an upgraded Beads-managed config lacking `archive_level` silently keeps zstd auto-GC (undermining the PR's stated goal on every already-bootstrapped install until manual regeneration), and a managed config written with the key can hard-fail a later strict-loading dolt downgrade (`UnmarshalStrict`). For Beads-managed configs, reconcile at load: add `archive_level: 0` when the binary supports it and the key is absent; strip it when the binary does not; preserve everything else including the port. For user-custom configs, warn rather than rewrite.
- **nit** `internal/doltserver/doltserver.go:496` — pass the `--config` path through `filepath.Abs` (the child runs with its cwd set to the data dir; the sibling dbproxy path already abs's defensively).
- **nit** `internal/doltserver/gc_config.go` — `SupportsArchiveLevelConfig` forks `dolt version` uncached on every start/UOW-provider creation; a per-binary memoization would remove the repeated probe if the path gets hot.

maphew:
Addressed both round-2 majors and both nits on the same branch, on top of the round-1 commits.

**Major 1 (--config mode skips .doltcfg discovery)** — d04a39356

Confirmed against the pinned module: `setupDoltConfig` (`cmd/dolt/commands/sqlserver/sqlserver.go:534-536`) returns immediately when `--config` is present, so the parent `../.doltcfg` lookup only runs in flag mode. `internal/doltserver.resolveCfgDir` now replicates that discovery exactly for our generated YAML: parent `../.doltcfg` (relative to the data dir) if it exists and is a directory, else `$data_dir/.doltcfg`, and `ErrMultipleDoltCfgDirs` (mirroring Dolt's own ambiguous case) if both exist rather than guessing. `Start()` resolves it once per launch, before the port-retry loop, and writes it into the YAML's `cfg_dir`.

`cfg_dir`'s minver: `servercfg.YAMLConfig.CfgDirStr` (`yaml_config.go:162`) carries **no `minver` tag at all** — it predates the tagging scheme, so it's supported on every version we could plausibly encounter. The existing 1.52.1 gate (from `archive_level`) stays as the single version gate for `--config` mode; no change needed there.

Tests: `TestResolveCfgDir_ParentExists` (the actual bug — asserts resolution finds and prefers a seeded parent `.doltcfg`), `TestResolveCfgDir_CurrentExists`, `TestResolveCfgDir_NeitherExists` (default), `TestResolveCfgDir_BothExistIsAmbiguous` (asserts `errors.Is(_, ErrMultipleDoltCfgDirs)`), and `TestResolveCfgDir_FileNotDirIgnored`.

**Major 2 (pre-existing managed configs bypass everything)** — e610f6f95

Distinguishing Beads-managed from user-custom: fresh renders (`renderProxiedServerConfig`) now carry a one-line YAML comment marker (`# Managed by Beads - safe to delete...`), invisible to Dolt's own parser (YAML comment). `ensureProxiedServerConfig` checks the marker on every existing default-path file:

- **marker present** (Beads-managed): `reconcileManagedProxiedServerConfig` parses the file into the `servercfg` YAML structs, adds `auto_gc_behavior.archive_level: 0` if supported-and-absent, strips `auto_gc_behavior` entirely if unsupported-and-present, and re-renders — a genuine parse-modify-rewrite, so port and every other field survive untouched. No-op (skips the write) when already aligned, which is the steady-state case on every ordinary command.
- **marker absent** (a pre-marker file from before this fix, or a genuinely custom path via env var/sidecar): never rewritten, one-line stderr warning instead, covering both directions (zstd may still be active + how to opt in; or the file may cause a refuse-to-start on this dolt).

Tests: `TestEnsureProxiedServerConfig_ManagedWithoutKeyGainsIt`, `TestEnsureProxiedServerConfig_ManagedWithKeyOnOldBinaryLosesIt`, `TestEnsureProxiedServerConfig_CustomConfigUntouchedWithWarning`, `TestEnsureProxiedServerConfig_PreMarkerDefaultConfigUntouchedWithWarning` (the marker-less-default-path case), plus lower-level `TestReconcileManagedProxiedServerConfig_{AddsKey,StripsKey,NoOpWhenAlreadyAligned}` and `TestIsManagedProxiedServerConfig`.

**Nits** — both in d04a39356

- `filepath.Abs` the `--config` path before exec in `internal/doltserver` (the child's cwd is the data dir; matches the sibling dbproxy/server path's existing defensive `filepath.Abs`).
- `SupportsArchiveLevelConfig` now memoizes per `doltBin` via `sync.Map` (`TestSupportsArchiveLevelConfig_Memoizes` proves it behaviorally: rewriting the stub binary at the same path after the first call must not flip the cached result, and only flips after an explicit cache reset).

Note on scope: `internal/storage/dbproxy/server`'s `--config` launch predates this whole PR (it always used `--config`, unconditionally, before any of these commits), so it doesn't have the same regression shape as the legacy launcher's newly-added `--config` branch — I left it out of the `.doltcfg` fix as a pre-existing, separate concern rather than scope creep on this PR.

Validation: `CGO_ENABLED=1 go build -tags gms_pure_go ./...` clean, `gofmt -l` clean on touched dirs, `go vet -tags gms_pure_go ./...` clean, and targeted `go test` green for `internal/doltserver`, `internal/storage/dbproxy/server`, `internal/storage/versioncontrolops`, and the touched `cmd/bd` tests (proxied_server, compact). `TestCompactSuite` still skips locally (no Docker-backed Dolt server in this sandbox), unrelated to this round.

maphew (commented):
**Agent verification review of the round-2 commits — verdict: concerns** (codex gpt-5.6-sol; the block-detection and truncating-write findings were independently confirmed by the orchestrator against the branch source)

The cfg_dir key/versioning and abs-path handling verify correct, the range contains only the expected commits and files, and the ordinary parent/current/both-exist discovery cases are covered. Four majors remain in the new code:

**Findings:**
- **major** `internal/doltserver/doltserver.go:812` — `filepath.Join(doltDir, "..", ...)` lexically cleans the `..` before symlink resolution, unlike Dolt which resolves `../.doltcfg` from the child process's actual cwd. With a symlinked data directory the physical parent `.doltcfg` can be missed, re-abandoning existing privileges/branch control - the exact failure round 2 set out to prevent. Resolve from the evaluated data-directory path and add a symlink-layout test.
- **major** `cmd/bd/proxied_server.go:223` — validating custom configs via `NewYamlConfig` instead of `YamlConfigFromFile` drops Dolt's environment-variable interpolation; a valid custom config using `${PORT}`-style values is now rejected before startup. Use interpolation-aware validation while keeping custom files untouched.
- **major** `cmd/bd/proxied_server.go:68` — `hasKey` tests `AutoGCBehavior != nil`, i.e. the whole block, not the `archive_level` field. A managed config whose block carries only `enable` neither gains `archive_level` when supported nor keeps its `enable` setting when unsupported (the strip removes the entire block). Inspect and modify only `ArchiveLevel_`, preserving sibling fields, with tests in both directions.
- **major** `cmd/bd/proxied_server.go:252` — reconciliation rewrites with truncating `os.WriteFile`; a crash or concurrent reader mid-write leaves a partial config that blocks every future startup. Write a same-directory temp file, sync, then atomically rename.
- **minor** `internal/doltserver/gc_config.go:40` — the capability cache is keyed by binary path only, so an in-place dolt upgrade/downgrade keeps the stale verdict for the process lifetime (and a test codifies this). Key by path+mtime/size or invalidate per launch.

maphew:
Addressed all four majors and the minor from the verification review, on the same branch.

**1. MAJOR (confirmed) — hasKey tested the whole block, not archive_level** — 20780b0b5

Fixed. `reconcileManagedProxiedServerConfig` now inspects and mutates only `ArchiveLevel_`: creates the `auto_gc_behavior` block if needed to add `archive_level: 0` (preserving any existing sibling fields on an already-present block), and on strip clears only `ArchiveLevel_`, removing the whole block only if that leaves it completely empty. Tests: `TestReconcileManagedProxiedServerConfig_PreservesSiblingFields` covers all three directions requested — block-with-only-`enable` gains `archive_level` (enable preserved); same block on an unsupported binary is untouched; block-with-both on an unsupported binary keeps `enable`, loses `archive_level`.

**2. MAJOR (confirmed) — truncating write** — 20780b0b5

Fixed. Both the reconcile-rewrite path and the fresh-render write path now go through `atomicWriteFile` — turns out this helper already existed in `cmd/bd/backup_export.go` (temp file in the same dir, `Sync`, `Close`, `Rename`), so I reused it instead of adding a duplicate.

**3. MAJOR (verified TRUE) — env-var interpolation** — 20780b0b5

Confirmed against the pinned module: `servercfg.YamlConfigFromFile` calls the unexported `interpolateEnv(data)` (`servercfg/env_interpolate.go`) *before* `NewYamlConfig`, expanding `${VAR}` placeholders. My round-2 switch to raw-bytes + `NewYamlConfig` for custom/marker-less configs skipped that entirely. Fixed by restructuring: the marker check (a literal comment-line prefix Beads itself writes and never interpolates) runs on raw bytes first to decide the branch, then a Beads-managed file (never contains interpolation syntax) is parsed via `NewYamlConfig` inside `reconcileManagedProxiedServerConfig`, while any unmanaged file (custom path, or a pre-marker default-path file) is now parsed via `YamlConfigFromFile` for both the required validation and the warn-check — matching what `dolt sql-server` itself does at actual startup. Test: `TestEnsureProxiedServerConfig_CustomConfigEnvVarInterpolation` uses a real `${VAR}` placeholder in `listener.port` and asserts it's accepted and the file is left byte-for-byte untouched.

**4. MAJOR (verified TRUE) — lexical vs physical `..` under a symlink** — 998996418

Confirmed mechanically with a standalone repro (not just reading source): `filepath.Join(doltDir, "..")` on a symlinked `doltDir` returns the symlink's *lexical* parent, while a process chdir'd into that symlink has its cwd tracked *physically* by the Linux kernel — verified both `pwd -P` and Python's `os.getcwd()` after `chdir` into a symlink resolve to the physical target, diverging from `filepath.Join`'s output. Since Dolt's own `../.doltcfg` lookup is a bare relative path resolved against the child process's real (physical) cwd, a symlinked data directory could miss the actual parent `.doltcfg`. Fixed: `resolveCfgDir` now resolves `doltDir` via `filepath.EvalSymlinks` (falling back to the lexical path on error) before computing the parent. `TestResolveCfgDir_SymlinkedDataDirFindsPhysicalParent` reproduces the exact symlinked layout; I confirmed it fails without the fix (reverted temporarily, watched it fail, restored) before finalizing.

**5. MINOR — memo key staleness** — 998996418

Cache key now includes `size`+`mtime` (`archiveLevelCacheKey`), falling back to an uncached probe when the binary can't be stat'd. The test that previously codified the staleness (asserted a rewritten binary's stale result "sticks") is replaced with two tests: `TestSupportsArchiveLevelConfig_MemoizesUnchangedBinary` (same file identity → still cached — proven via an embedded call counter in the stub, not just the returned bool) and `TestSupportsArchiveLevelConfig_InvalidatesOnFileIdentityChange` (changed mtime, same-length content → re-probed; `os.Chtimes` forces a deterministic mtime delta rather than relying on real sleep/filesystem timestamp resolution).

Validation: `CGO_ENABLED=1 go build -tags gms_pure_go ./...` clean, `gofmt -l` clean, `go vet -tags gms_pure_go ./...` clean, targeted `go test` green for `internal/doltserver`, `internal/storage/dbproxy/server`, `internal/storage/versioncontrolops`, and touched `cmd/bd` tests (proxied_server, compact, plus the pre-existing `atomicWriteFile`/backup tests since that helper is now shared). `TestCompactSuite` still skips locally (no Docker Dolt server in this sandbox), unrelated.

maphew (commented):
**Agent verification review of the round-3 commits — verdict: concerns** (codex gpt-5.6-sol)

Round 3's core fixes verify sound: field-level reconciliation, physical cfg-dir resolution, memo invalidation, and clean two-commit ancestry. Remaining findings are narrower edge behavior in config handling:

**Findings:**
- **major** `cmd/bd/proxied_server.go:289` (also ~321) — `atomicWriteFile` renames over the final path, so a symlinked managed config gets its symlink replaced by a regular file; the previous write-through semantics are lost. Resolve the target with `filepath.EvalSymlinks` before the rename, or explicitly refuse to rewrite symlinked configs with a warning.
- **major** `cmd/bd/proxied_server.go:135` — the unmanaged-config warning tests block presence, not `ArchiveLevel_`: a sibling-only (`enable`-only) custom config misses the zstd-risk warning on capable dolt versions and gets a false incompatibility warning on older ones. Same field-vs-block fix as the reconciler received.
- **minor** `cmd/bd/proxied_server.go:280` — a marker does not prove the file lacks `${VAR}` placeholders (operators can hand-edit a managed file); the raw-bytes parse then errors. Ensure a managed file that fails the raw parse degrades to warn-without-rewrite rather than failing server startup, or reconcile without expanding/persisting placeholders.
- **nit** `cmd/bd/backup_export.go:82` — the helper's "crash-safe" doc overstates: it fsyncs the file but not the parent directory, and rename atomicity is not guaranteed off-Unix. Narrow the claim or strengthen the helper.

maphew:
Addressed all four round-4 findings, on the same branch.

**1. MAJOR — atomic rename replaces a symlinked config with a regular file** — 6dcfdd5b2

Confirmed: `os.Rename`'s destination does not follow a symlink — it replaces whatever sits directly at that path. Both `atomicWriteFile` call sites in `ensureProxiedServerConfig` (reconcile-rewrite and fresh-render) now resolve the target through `filepath.EvalSymlinks` first (`resolveConfigWriteTarget`), falling back to the given path when it doesn't exist yet (also covers a dangling symlink — self-healed by writing at its own location). Test: `TestEnsureProxiedServerConfig_ManagedConfigBehindSymlinkPreservesSymlink` reconciles a `config.yaml` that's a symlink and asserts the symlink itself survives (still pointing at the same physical file) while the physical target gets the reconciled content. Confirmed it fails without the fix — reverted the resolution call, watched the test fail on the "must still be a symlink" assertion, restored.

**2. MAJOR — warn function tests block presence, not `ArchiveLevel_`** — 6dcfdd5b2

Fixed with the same field-vs-block distinction the reconciler got in round 3. Since `warnUnmanagedProxiedServerConfig` only has the `servercfg.ServerConfig` interface (not the concrete struct) coming out of `YamlConfigFromFile`, added `autoGCArchiveLevelIsSet` — type-asserts to `*servercfg.YAMLConfig` (which both of the pinned module's loaders always return under that interface) for the field-level check, falling back to the coarser block check only if that assertion were ever to fail. Test: `TestWarnUnmanagedProxiedServerConfig_EnableOnlyFieldLevel` — an enable-only custom config gets the zstd-risk warning on a capable dolt, and gets *no* warning at all on an incapable one (the old block-level check would have produced a false incompatibility warning there).

**3. MINOR — managed-file parse failure should degrade, not fail startup** — 6dcfdd5b2

Verified the prior behavior first, as asked: a marker-bearing file with a `${VAR}` placeholder fails `reconcileManagedProxiedServerConfig`'s raw-bytes `NewYamlConfig` parse (deliberately non-interpolating), and that error propagated straight out of `ensureProxiedServerConfig`. Traced the caller: `newManagedProxiedServerUOWProvider` in `cmd/bd/uow_factory.go` returns it directly (`if err != nil { return nil, err }`), so this aborted the *entire* `bd` command before ever attempting to start the managed server — even though a real `dolt sql-server --config` process would happily interpolate and start against the same file. Fixed: on a managed-file parse failure, log a one-line warning and fall through to the same warn-only/never-rewrite treatment as an unmanaged file, returning success. Test: `TestEnsureProxiedServerConfig_ManagedFileWithPlaceholderDegradesGracefully` — marker-bearing file with a `${PORT}`-style placeholder, `ensureProxiedServerConfig` returns no error, emits the warning, and leaves the file byte-for-byte untouched.

**4. NIT — narrow `atomicWriteFile`'s doc comment** — 25047d7ae

Narrowed rather than strengthened, per the instruction not to silently change the existing `saveBackupState` caller's expectations: the comment now states plainly that only the temp file's contents are fsynced (not the parent directory entry), that rename atomicity is a POSIX/Unix property not guaranteed on Windows, and that it replaces whatever is at `path` — symlink or not — pointing at `proxied_server.go`'s `resolveConfigWriteTarget` for a case that needs the physical-target resolution. No behavior change; separate commit from the others.

Validation: `CGO_ENABLED=1 go build -tags gms_pure_go ./...` clean, `gofmt -l` clean, `go vet -tags gms_pure_go ./...` clean, targeted `go test` green for `internal/doltserver`, `internal/storage/dbproxy/server`, `internal/storage/versioncontrolops`, and the touched `cmd/bd` tests (proxied_server, compact, backup/`atomicWriteFile`).

PR #4987, opened 2026-07-23
fix(schema): restore missing wisp_dependencies constraints after 0047 delegate repair

maphew:
## Summary

Closes the two test-coverage gaps flagged in #4878's review, and fixes a real
defect one of them surfaced along the way.

- **Review nit 1**: migration 0057's guarded `MODIFY COLUMN ... LONGTEXT`
  conversion branch was untested (every existing test hits the no-op branch,
  since a fresh chain already has these columns at LONGTEXT via 0048).
- **Review nit 2**: the 0047 delegate repair path (`ensureWispDependenciesSplitTargets`
  adding split columns to a legacy pre-split `wisp_dependencies` table) was
  only covered by a sqlmock unit test with the split columns already present;
  the real end-to-end legacy interaction was untested.

## The dolt CLI discovery

Both gaps were first attempted as `dolt sql -f` tests, mirroring the existing
`ThroughDoltCLI` pattern in `schema_test.go`. That approach cannot work:
**dolt CLI v2.2.2 (the current "latest" release, matching what CI installs)
silently no-ops any `PREPARE`/`EXECUTE`-driven `ALTER TABLE` run through
`dolt sql -f`/`-q`/piped stdin, and raises no error either.** Confirmed by
direct repro outside the test suite (plain `ALTER TABLE ... MODIFY` always
applies; the identical statement wrapped in `SET @sql = ...; PREPARE stmt
FROM @sql; EXECUTE stmt;` never does, in any invocation mode).

Every guarded DDL branch these two gaps needed to exercise -- 0057's MODIFY,
0047's own inline `wisp_dependencies` split, `ignored/0003`, `ignored/0005`
-- uses exactly that shape. This is why **no existing `ThroughDoltCLI` test
has ever actually run one of these guards' "fires and converts" branch**:
each either substitutes hand-written direct DDL (`cli_migrations.go`) or only
proves the no-op branch. A substitute can't catch a bug in the real guarded
SQL either, which is precisely what the review asked this coverage to catch.

**This is not a production defect** -- `MigrateUp` applies migrations through
`database/sql`/go-mysql-server's real engine, never through the dolt CLI's
batch-file executor, which nothing in production uses.

## The fix: engine-based tests, which then found a real defect

New tests in `internal/storage/embeddeddolt` (not `schema_test.go`) run the
exact frozen migration file content through the same path production uses --
a single `db.ExecContext(ctx, string(data))` against a real embedded (cgo)
go-mysql-server/Dolt engine, reusing the package's existing
`seedMainSchemaAt`/`openPinnedConn`/`schema.MigrateUp` harness.
`PREPARE`/`EXECUTE` behaves normally there.

- `TestEmbeddedMigration0057ConvertsTextEventsColumnsToLongtext` -- passes.
  Confirms the CLI diagnosis: identical SQL that silently no-op'd via the CLI
  converts correctly through the real engine, for both the "both columns
  drifted" and the header comment's "mixed" case, with idempotency proven by
  a second pass.
- `TestEmbeddedMigration0047DelegatesToSplitTargetRepairAndConvergesThroughIgnoredChain`
  -- seeds a genuinely legacy pre-split `wisp_dependencies` table (with a
  realistic `fk_wisp_dep_issue` FK, so the FK preserve-and-restore path is
  actually exercised, not trivially satisfied by a FK-less seed) and drives
  the real `schema.MigrateUp` end to end (main sequence including the
  Go-level delegate repair at v47, then the ignored sequence's real
  `ignored/0003`/`ignored/0005` SQL). **This is what surfaced the defect**:
  `fk_wisp_dep_wisp_target`, `fk_wisp_dep_issue_target`, and
  `ck_wisp_dep_one_target` never converged on this path.

### The defect (mybd-mq9b)

`ensureWispDependenciesSplitTargets` (the Go-level delegate repair) adds the
three split-target columns and backfills them from the legacy
`depends_on_id` column, but never adds the two split-target foreign keys or
the one-target CHECK constraint. Neither of the two migrations that would
normally add them ever fires for this path:

- `ignored/0003`'s own inline split is gated on `depends_on_wisp_id` being
  *absent* -- already false, since the delegate added it first.
- `ignored/0005` only *restores* a foreign key it detects present *before
  its own drop* of `depends_on_id` -- a genuinely legacy table never had
  `fk_wisp_dep_wisp_target`/`fk_wisp_dep_issue_target` to begin with, since
  their target columns didn't exist pre-split.

Net effect: any real database that took the delegate path (wisp tables
present, `wisp_dependencies` genuinely legacy) converged structurally
(columns, surrogate id, primary key, unique keys) but silently ended up
**permanently missing referential integrity** on two FK columns and the
one-of-three-targets invariant -- contradicting the #4878 review's "confirmed
by inspection to converge" for this path.

### The fix: new migration 0058, not an edit to frozen files

`0047`, `ignored/0003`, and `ignored/0005` are shipped and frozen
(`scripts/check-migration-hygiene.sh` Check C forbids editing a migration
file that already exists on `main`). More importantly, **a database already
affected by this gap has its `schema_migrations` cursor sitting at latest**,
so none of those files would ever become pending on it again even if editing
them were allowed -- only a new migration can reach it.

`migrations/0058_heal_wisp_dependencies_split_constraints.up.sql` adds the
three missing constraints, each independently guarded on its own
`INFORMATION_SCHEMA` checks (constraint absence, and its required
table/columns present first, not assumed), following the 0053/0057
repair-migration precedent from #4878 itself. Guards are safe on every
population:
- fresh-recreate-path databases (constraints already present) -- no-op.
- delegate-path databases (constraints missing) -- adds them.
- a database where `wisp_dependencies` or a required column is for any
  other reason absent by the time 0058 runs -- left untouched.

`TestEmbeddedMigration0058HealsAlreadyAffectedDatabase` covers the scenario
that matters most: a database that already took the delegate path *before*
0058 existed (built through the real chain, then the three constraints
dropped to reproduce that end state), proving 0058 heals it directly, and
that a second pass is an idempotent no-op.

With 0058 in place, `TestEmbeddedMigration0047Delegates...`'s full
assertions (including the fk_/ck_ constraints) now pass end to end.

## Falsification checks

- 0057: corrupted a copy of the MODIFY target (`LONGTEXT` -> `MEDIUMTEXT`) --
  both new subtests failed as expected (`COLUMN_TYPE = mediumtext, want
  longtext`); reverted before committing.
- 0058: corrupted a copy of the `ck_wisp_dep_one_target` guard (inverted the
  absence check) -- the new heal test failed as expected
  (`ck_wisp_dep_one_target count = 0, want 1`); reverted before committing.

## Validation

```
./scripts/check-migration-hygiene.sh                                  # Migration hygiene OK.
gofmt -l internal/storage/schema/ internal/storage/embeddeddolt/      # clean
CGO_ENABLED=1 go vet -tags gms_pure_go ./internal/storage/schema/... ./internal/storage/embeddeddolt/...   # clean
CGO_ENABLED=1 go test -tags gms_pure_go ./internal/storage/schema/... -count=1                             # ok, 10.5s
BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go ./internal/storage/embeddeddolt/... \
  -run 'TestEmbeddedMigration0057ConvertsTextEventsColumnsToLongtext|TestEmbeddedMigration0047DelegatesToSplitTargetRepairAndConvergesThroughIgnoredChain|TestEmbeddedMigration0058HealsAlreadyAffectedDatabase' -v
--- PASS: TestEmbeddedMigration0057ConvertsTextEventsColumnsToLongtext (both subtests)
--- PASS: TestEmbeddedMigration0047DelegatesToSplitTargetRepairAndConvergesThroughIgnoredChain
--- PASS: TestEmbeddedMigration0058HealsAlreadyAffectedDatabase
```

## What's not in this PR

The dolt-CLI draft tests from the first attempt (against `schema_test.go`'s
`ThroughDoltCLI` pattern) are not included -- they cannot pass against dolt
CLI v2.2.2's `PREPARE`/`EXECUTE` limitation, and the engine-based tests above
replace them.

Refs: #4878, mybd-mq9b, mybd-i9un

### DIFF
--- a/internal/storage/embeddeddolt/migrate_frozen_guard_convergence_test.go
+++ b/internal/storage/embeddeddolt/migrate_frozen_guard_convergence_test.go
@@ -1,0 +1,405 @@
+//go:build cgo
+
+package embeddeddolt_test
+
+import (
+	"context"
+	"database/sql"
+	"os"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/storage/schema"
+)
+
+// This file covers two coverage gaps flagged in gastownhall/beads#4878's
+// review (mybd-i9un): migration 0057's guarded TEXT->LONGTEXT conversion
+// branch, and 0047's Go-level delegate repair converging through the ignored
+// sequence's own structural split (#4555).
+//
+// Both were first attempted as dolt-CLI (`dolt sql -f`) tests, matching the
+// existing ThroughDoltCLI pattern in internal/storage/schema/schema_test.go.
+// That approach does not work for either gap: dolt CLI v2.2.2 (the current
+// "latest" release, matching what CI installs) silently no-ops any
+// PREPARE/EXECUTE-driven ALTER TABLE run through `dolt sql -f`/`-q`/piped
+// stdin -- confirmed by direct repro, not just by these two tests failing.
+// A plain (non-prepared) ALTER TABLE always applies; a PREPARE'd one, guarded
+// or not, never does, and raises no error either. Every one of 0057, 0047's
+// own inline wisp_dependencies split, and ignored/0003 and ignored/0005 use
+// exactly this `SET @sql = IF(...); PREPARE stmt FROM @sql; EXECUTE stmt;`
+// shape, so no existing ThroughDoltCLI test has ever actually exercised one
+// of these guards' "fires and converts" branch: each either substitutes
+// hand-written direct DDL (cli_migrations.go's cliCompatibleMigrationSQL) or
+// only proves the no-op branch. A substitute would not catch a bug in the
+// real guarded SQL either, which is precisely what the #4878 review nits
+// asked this coverage to catch.
+//
+// These tests instead run the exact frozen migration file content through
+// the same path production uses: a single db.ExecContext(ctx, string(data))
+// against a real embedded (cgo) go-mysql-server/Dolt engine -- see
+// schema.go's runMigrations, and schema.MigrateUp for the full orchestration
+// including preMigrationRepair. PREPARE/EXECUTE works normally through that
+// real SQL engine; the limitation above is specific to the dolt CLI's
+// `sql -f`/`-q` batch executor, which nothing in production uses.
+//
+// Both tests require BEADS_TEST_EMBEDDED_DOLT=1 (requireEmbedded, defined in
+// migrate_selfheal_test.go), the same gate every other cgo embedded-dolt test
+// in this package uses.
+
+// TestEmbeddedMigration0057ConvertsTextEventsColumnsToLongtext exercises
+// 0057's actual MODIFY branch: on any fresh chain 0048 has already widened
+// both columns to LONGTEXT, so @old_value_needs_fix/@new_value_needs_fix
+// read 0 and the MODIFY statements never fire under any pre-existing test.
+// Force the columns back to TEXT after a full migrate, then run 0057's real
+// frozen SQL (read from migrations/0057_events_value_columns_idempotent_longtext.up.sql)
+// and assert each column converts independently.
+func TestEmbeddedMigration0057ConvertsTextEventsColumnsToLongtext(t *testing.T) {
+	requireEmbedded(t)
+	ctx := t.Context()
+
+	migrationSQL, err := os.ReadFile("../schema/migrations/0057_events_value_columns_idempotent_longtext.up.sql")
+	if err != nil {
+		t.Fatalf("read 0057 migration: %v", err)
+	}
+
+	t.Run("both columns drifted to TEXT", func(t *testing.T) {
+		dataDir := seedMainSchemaAt(t, ctx, schema.LatestVersion())
+		conn, closeConn := openPinnedConn(t, ctx, dataDir)
+		defer closeConn()
+
+		execFrozenGuard(t, ctx, conn, `
+ALTER TABLE events MODIFY COLUMN old_value TEXT;
+ALTER TABLE events MODIFY COLUMN new_value TEXT;
+`)
+		requireEventsColumnType(t, ctx, conn, "old_value", "text")
+		requireEventsColumnType(t, ctx, conn, "new_value", "text")
+
+		execFrozenGuard(t, ctx, conn, string(migrationSQL))
+		requireEventsColumnType(t, ctx, conn, "old_value", "longtext")
+		requireEventsColumnType(t, ctx, conn, "new_value", "longtext")
+
+		// Idempotent: a second pass, with both guards now reading 0, must
+		// take the no-op branch for both columns rather than re-issue MODIFY
+		// (the #4353 encoding-flip risk 0057 exists to guard against).
+		execFrozenGuard(t, ctx, conn, string(migrationSQL))
+		requireEventsColumnType(t, ctx, conn, "old_value", "longtext")
+		requireEventsColumnType(t, ctx, conn, "new_value", "longtext")
+	})
+
+	t.Run("mixed: only old_value drifted to TEXT", func(t *testing.T) {
+		dataDir := seedMainSchemaAt(t, ctx, schema.LatestVersion())
+		conn, closeConn := openPinnedConn(t, ctx, dataDir)
+		defer closeConn()
+
+		// The header comment's motivating case: a database that picked up
+		// 0048 unevenly, so one column already reads LONGTEXT while the
+		// other is still TEXT. The two guards must act independently -- a
+		// single shared check would either skip converting old_value here or
+		// re-issue a MODIFY on new_value despite it already being LONGTEXT.
+		execFrozenGuard(t, ctx, conn, "ALTER TABLE events MODIFY COLUMN old_value TEXT;")
+		requireEventsColumnType(t, ctx, conn, "old_value", "text")
+		requireEventsColumnType(t, ctx, conn, "new_value", "longtext")
+
+		execFrozenGuard(t, ctx, conn, string(migrationSQL))
+		requireEventsColumnType(t, ctx, conn, "old_value", "longtext")
+		requireEventsColumnType(t, ctx, conn, "new_value", "longtext")
+
+		execFrozenGuard(t, ctx, conn, string(migrationSQL))
+		requireEventsColumnType(t, ctx, conn, "old_value", "longtext")
+		requireEventsColumnType(t, ctx, conn, "new_value", "longtext")
+	})
+}
+
+// TestEmbeddedMigration0047DelegatesToSplitTargetRepairAndConvergesThroughIgnoredChain
+// covers #4555's delegate path end-to-end: wisp_dependencies present but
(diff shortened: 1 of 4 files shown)


maphew (commented):
**Paired agent review — verdict: concerns** (claude-opus-4-8 and codex gpt-5.6-sol reviewed independently; contradiction adjudicated against repo history)

The core of the fix is right: the three constraint definitions are byte-identical to the canonical `wispDependenciesTableDDLForMigration0047`/`ignored/0003` forms, guards are correctly `DATABASE()`-scoped, 0058 is contiguous with no upstream collision, and the subtle single-pass ordering (0058 runs before `ignored/0005`, which then re-drops and re-restores the FKs it now observes) converges to canonical - all traced by the opus reviewer. But the codex reviewer caught what the opus reviewer rated safe: adding constraints over data that accumulated during the unconstrained window. The repo has been bitten by exactly this before - `ignored/0011`'s header documents FKs re-added under `FOREIGN_KEY_CHECKS=0` accepting legacy orphans, after which "Dolt then fails constraint validation on subsequent writes, so one legacy orphan bricks every bd create on an otherwise healthy database" - which settles the severity dispute in codex's favor.

**Findings:**
- **major** `internal/storage/schema/migrations/0058_heal_wisp_dependencies_split_constraints.up.sql:115` — with `FOREIGN_KEY_CHECKS=0`, both FK adds accept existing orphaned rows (targets deleted while no CASCADE was in force). Per the `ignored/0011` precedent, such declared-but-violated FKs later fail Dolt constraint validation on writes. Pre-clean orphans before adding the FKs - delete rows whose issue/wisp target no longer resolves, i.e. exactly what `ON DELETE CASCADE` would have done - mirroring 0011's idiom, and add an engine test seeding an orphan (found by codex; severity confirmed against `ignored/0011`'s documented failure class).
- **major** `...0058...up.sql:134` — `ADD CONSTRAINT ck_wisp_dep_one_target` validates existing rows regardless of `FOREIGN_KEY_CHECKS=0`; any zero-target or multi-target row accumulated during the unconstrained window makes MigrateUp fail before recording 0058, repeatedly aborting store-open with an opaque engine error. Normalize deterministically first (delete zero-target rows; for multi-target rows apply the delegate backfill's own precedence), and add an engine test seeding invalid rows (flagged by both reviewers; opus rated it minor, codex major - the store-open-bricking consequence makes it major).
- **minor** `internal/storage/embeddeddolt/migrate_frozen_guard_convergence_test.go:332` — assertions count constraint names/types only; wrong referenced tables/columns, delete/update actions, or CHECK expressions would still pass. Assert via `KEY_COLUMN_USAGE`/`REFERENTIAL_CONSTRAINTS`/`CHECK_CONSTRAINTS` or exercise the constraints behaviorally, including a mixed population with only one constraint missing (codex).
- **nit** `...migrate_frozen_guard_convergence_test.go:1` — tests skip without `BEADS_TEST_EMBEDDED_DOLT=1`; CI sets it, but a bare local `go test ./...` silently runs none of this (opus).

maphew:
Addressed the paired review's findings on the same branch (2 new commits, no force-push):

**Major 1 (FK orphans)**: 0058 now deletes the whole `wisp_dependencies` row when its `depends_on_wisp_id`/`depends_on_issue_id` target no longer resolves -- exactly what `ON DELETE CASCADE` would have done -- immediately before each `ADD CONSTRAINT ... FOREIGN KEY`. Mirrors `ignored/0011`'s guarded `LEFT JOIN ... WHERE <target>.id IS NULL` idiom, gated on the same `@needs_fk_*` flags the constraint adds use.

**Major 2 (CHECK over invalid rows)**: before `ADD CONSTRAINT ck_wisp_dep_one_target`, zero-target rows are deleted (semantically meaningless) and multi-target rows are normalized to exactly one target. Precedence is **external > wisp > issue**, verified against `wispDependenciesSplitTargetBackfillSQL`'s own statement order (external-prefix match first, then wisp, then issue, each guarded on the earlier ones not already matching) and its sibling test `TestWispDependenciesSplitTargetBackfillPrefersWispOverIssueThroughDoltCLI` (an id resolving as both a wisp and an issue backfills to the wisp reading). Documented in the migration header.

**Minor (assertion strength)**: added `requireForeignKeyReferences` (checks `KEY_COLUMN_USAGE` referenced table/column + `REFERENTIAL_CONSTRAINTS` `UPDATE_RULE`/`DELETE_RULE = CASCADE`) and `requireCheckRejectsInvalidTargetCounts` (behavioral -- insert zero-target/two-target rows and expect rejection, one-target and expect acceptance; `CHECK_CONSTRAINTS.CHECK_CLAUSE` isn't string-matched since the engine reformats it on storage). Applied to both existing engine tests. Added `TestEmbeddedMigration0058AddsOnlyMissingConstraintInMixedPopulation` for the mixed-population case (only `ck_wisp_dep_one_target` missing, both FKs already present).

**New majors coverage**: `TestEmbeddedMigration0058CleansOrphanedAndInvalidWispDependenciesRowsBeforeAddingConstraints` seeds a delegate-path DB at cursor-latest with an FK orphan, a zero-target row, and a multi-target row; runs 0058; asserts the orphan and zero-target rows are gone, the multi-target row is normalized to its wisp-precedence target, all three constraints converge, and a second pass is an idempotent no-op.

**Falsification**: corrupted a copy of the new wisp-orphan cleanup guard (its `DELETE` swapped for a no-op) -- the new test failed as expected (`orphan row count = 1, want 0`); reverted before committing.

**Validation**: `scripts/check-migration-hygiene.sh` OK, `gofmt`/`go vet` clean, `internal/storage/schema` package tests pass, all targeted `BEADS_TEST_EMBEDDED_DOLT=1` engine tests pass (0057, 0047 delegate, 0058 heal, 0058 mixed-population, 0058 orphan/invalid-row cleanup).

Commits: `ea18ab6ab` (fix), `4ac4ee3a6` (tests).

maphew:
CI failure fixed (run 29985853386, "Test (Embedded Dolt Storage)"): the four new engine tests read frozen migration SQL via `os.ReadFile("../schema/migrations/....up.sql")`, a package-relative path that resolved locally but not in the CI job's execution context (`no such file or directory`, 0.00s).

Fix: added `schema.MigrationSQL(name string) (string, error)`, a minimal read-only accessor over the package's existing `//go:embed migrations/*.up.sql` FS (`mainSource.files` -- the same embedded content `runMigrations` applies in production). Switched all four call sites to it and dropped the now-unused `os` import. This is strictly higher fidelity than the disk read it replaces: it's the literal embedded bytes the binary ships with, works identically regardless of the test binary's CWD, and can't drift from the build.

Validation: `gofmt`, `go vet`, `scripts/check-migration-hygiene.sh` (accessor lives in the schema package), `internal/storage/schema` package tests, and all five targeted `BEADS_TEST_EMBEDDED_DOLT=1` engine tests -- all pass locally.

Commit: `cecf10649`.

PR #4997, opened 2026-07-23
feat(gate): support bd gate check in proxied-server mode

coffeegoddd:
## Summary

Adds proxied-server support for `bd gate check`, which previously errored out with `gate check is not supported in proxied-server mode`. It now has behavioral and integration-test parity with embedded mode.

No new domain/use-case methods were needed — the flow maps onto the existing `IssueUseCase` surface (`SearchIssues`, `GetIssue`, `CloseIssue`, `UpdateIssue`).

## Changes

**`cmd/bd/gate.go` — shared, storage-agnostic plumbing**
- `checkGHRun` now takes a `persistAwaitID func(gateID, runID string) error` callback instead of a `bool` (nil = dry-run / no persistence), so the embedded path can write through `store` and the proxied path can defer the write into a transaction.
- Extracted helpers reused by both modes: `filterCheckableGates`, `printNoOpenGates`, `evaluateGates` (the gh/timer/bead switch, keyed off the `issueGetter` interface both `store` and `uw.IssueUseCase()` satisfy), `applyGateCheckResults` (render + close via a callback + escalate), and `printGateCheckSummary`. The embedded `RunE` collapsed to call these.

**`cmd/bd/gate_proxied_server.go` — the proxied dual**
- `runGateCheckProxiedServer`: evaluates gates under a read UOW *outside* any transaction (gate evaluation shells out to `gh` and must not run inside a retryable `RunTx`), then applies await-id updates and gate closes inside a single `uow.RunTxResult` returning `bd: gate check` as the commit message. Hooks fire (`fireProxiedCloseHooks`) and output renders after the commit, mirroring `close_proxied_server.go`; sets `commandDidWrite`.

**Dispatch + callers**
- `gate check` dispatches to the dual instead of erroring; updated the four `checkGHRun` call sites (`close.go` + three in `gate_test.go`) for the new signature, preserving the `updateGateAwaitIDFunc` test seam.

## Tests

New `cmd/bd/gate_check_proxied_integration_test.go` — `TestProxiedServerGateCheck`, 8 parallel subtests with parity to the embedded `gate_check_*` scenarios, asserting real resolution behavior (not just that the command runs):

- `no_gates`, `timer_expired_resolves`, `timer_pending_stays_open`, `dry_run_does_not_close`, `bead_gate_resolves_when_target_closes`, `type_filter_scopes_evaluation`, `limit_bounds_gates_checked`, `timer_escalate_is_noop_and_resolves`.

Uses the shared proxied harness (`requireSharedProxiedServer` / `newSharedProxiedProject` / `bdProxiedRunBuffers` / `openProxiedDB`). Timer/bead/gh gates are seeded by writing the `await_*` columns directly through `openProxiedDB` (the `lint_stale_proxied_integration_test.go` pattern), since `bd gate create` is not available in proxied mode. `//go:build cgo` and the `TestProxiedServer` name prefix mean the shard script discovers it and it hash-distributes into one CI shard automatically.

## Verification

- `go build`, `go vet`, `golangci-lint run ./cmd/bd/` — clean.
- `BEADS_TEST_PROXIED_SERVER=1 go test -run '^TestProxiedServerGateCheck$'` — all 8 subtests pass end-to-end against a real proxied Dolt server, in parallel.
- Existing gate unit tests (`GateCheck|CheckGHRun|CheckBeadGate|Gate`) pass.

### DIFF
--- a/cmd/bd/close.go
+++ b/cmd/bd/close.go
@@ -548,7 +548,7 @@ func checkGateSatisfaction(issue *types.Issue) error {
 
 	switch {
 	case strings.HasPrefix(issue.AwaitType, "gh:run"):
-		resolved, escalated, reason, err = checkGHRun(issue, true)
+		resolved, escalated, reason, err = checkGHRun(issue, func(gateID, runID string) error { return updateGateAwaitIDFunc(nil, gateID, runID) })
 	case strings.HasPrefix(issue.AwaitType, "gh:pr"):
 		resolved, escalated, reason, err = checkGHPR(issue)
 	case issue.AwaitType == "timer":
--- a/cmd/bd/gate.go
+++ b/cmd/bd/gate.go
@@ -561,7 +561,7 @@ Examples:
 	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if usesProxiedServer() {
-			return HandleErrorRespectJSON("gate check is not supported in proxied-server mode")
+			return runGateCheckProxiedServer(cmd, rootCtx)
 		}
 		CheckReadonly("gate check")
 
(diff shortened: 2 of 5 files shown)


maphew (changes requested):
Cross-vendor paired agent review (Claude Opus + Codex gpt-5.6-sol, independent traces, findings adjudicated and re-verified against the PR head before posting). Verdict: **needs changes** — structure is right, three concrete fixes wanted before merge.

What checks out (verified, not assumed): every PR-body wiring claim is real (`proxiedOpenReadUOW`, `RunTxResult`, `fireProxiedCloseHooks`, the `issueGetter` seam — all present and satisfied by both `store` and `uw.IssueUseCase()`); the `gate.go` extraction is behavior-preserving with the `checkGHRun` seam threaded through all call sites; dispatch is a plain `usesProxiedServer()` branch with no store-open ordering hazard; dry-run takes no transaction; and `TestProxiedServerGateCheck` is **active in CI** — both reviewers independently confirmed it hash-distributes into the proxied shard lane (shard 14/15) via `.github/scripts/proxied-test-shard.sh`, so the proxied path is genuinely exercised.

**1. Should-fix (both vendors converged; maintainer-verified against `uow/tx.go`): audit write inside the retryable transaction closure.** `audit.LogFieldChange` at `gate_proxied_server.go:126` runs inside the `RunTxResult` work function. `RunTxResult` re-invokes the whole closure when `uw.Commit` hits a serialization conflict, and the audit sidecar append is not transactional — so a commit-time retry duplicates the `status→closed` audit entry, and a permanently failed commit leaves a false one for a close that never landed. The `gateCheckApply` doc comment guards the `closed` slice against exactly this, but the audit call escapes the struct. The sibling `close_proxied_server.go:135-140` already does it right: stash old-status/reason in the outcome and emit audit post-commit, next to `fireProxiedCloseHooks`. Same pattern here — `proxiedGateClose` already carries `before`/`after`.

**2. Should-fix: the read UOW (and its DB connection/transaction) is held open across external `gh` calls.** `evaluateGates` runs before `readUW.Close(ctx)` (`gate_proxied_server.go:77-93`) and reaches `gh run view` for `gh:run` gates — an unbounded external process while a server transaction sits open. The PR body's "external evaluation occurs outside any transaction" holds for the *write* tx but not the read one. Cheap fix: snapshot the gates (`SearchIssues` + `filterCheckableGates`), close the read UOW, then evaluate — `evaluateGates` only needs the getter for parent lookups, which could ride a fresh short-lived UOW or the snapshot.

**3. Should-fix: persisted `await_id` discoveries skip update hooks.** The closure commits `await_id` updates for discovered run IDs (`:107-110`), but post-commit dispatch iterates only `applied.closed` — no `on_update` fires for a pending gate that gained an `await_id`. The embedded path emits `EventUpdate` via the hook decorator, and proxied update commands dispatch update hooks explicitly. The integration suite can't catch it (the gh gate is deliberately excluded from the harness).

**Design note, not a blocker (adjudicated down after checking embedded parity):** gates are closed from the pre-transaction evaluation snapshot, and a serialization retry replays the captured `results`; the in-tx `GetIssue` fetches `before` but never re-validates that the gate is still open with the same await config. The embedded flow has the same evaluate-then-close semantics, so this is inherited design, not a proxied regression — but the proxied split widens the window (external `gh` latency between read and write tx). Since you're already refetching `before` in-tx, a one-line status/await-config check before `CloseIssue` would make the proxied path strictly safer than embedded rather than slightly wider-windowed.

Nits: `before, _ :=` swallows the GetIssue error (matches sibling tolerance, effectively unreachable — fine); no user-facing doc mentions that `bd gate check` now works in proxied-server mode.

coffeegoddd:
Thanks for the thorough paired review — all three should-fix items plus the design-note safety check are addressed in 0c2628913.

**1. Audit inside the retryable closure** — moved `audit.LogFieldChange` out of the `RunTxResult` work function. `proxiedGateClose` now stashes `oldStatus`/`reason`, and audit is emitted post-commit next to `fireProxiedCloseHooks`, mirroring `close_proxied_server.go`. No more duplicate/false audit entries on retry or failed commit.

**2. Read UOW held across `gh` calls** — the read UOW is now closed right after `SearchIssues` + `filterCheckableGates`, before evaluation. Bead-gate lookups ride a fresh short-lived UOW via a new `proxiedFreshReadGetter` (opens/closes per `GetIssue`), so no server transaction is open across the external `gh` subprocess.

**3. `await_id` discoveries skip update hooks** — the tx now captures the post-update state of each persisted `await_id`, and post-commit fires `on_update` for them (before the close hooks, preserving embedded ordering). `commandDidWrite` keys off closed+updated.

**Design note** — added the one-line guard: `before` is refetched in-tx and a gate already closed in committed state is skipped rather than re-closed, so the proxied path is now strictly safer than embedded across the wider read→write window.

Nit (doc): there's no existing doc that claims gate check is unsupported in proxied mode (the `--proxied-server*` refs are just flag docs), so nothing stale to correct — happy to add an explicit mention if you'd like one.

`golangci-lint` clean; `TestProxiedServerGateCheck` still passes end-to-end against a live proxied Dolt server.

coffeegoddd:
Addressed both Copilot comments in 2b43fcb40:

- **`gate_proxied_server.go` (audit old-status)** — the close audit's old value now derives from the in-transaction `before.Status` (defaulting to `open` when empty) rather than the pre-tx evaluation snapshot (`r.gate.Status`), so a status change between evaluation and the close transaction can't record a stale old value. Matches `close_proxied_server.go`.
- **`gate.go` (stale `checkGHRun` doc)** — removed the doc comment that still referenced the removed `persistDiscoveredRunID` boolean; behavior is now controlled by whether the `persistAwaitID` callback is nil.

Lint clean; gate unit + `TestProxiedServerGateCheck` integration tests pass.

PR #5000, opened 2026-07-23
feat: support compact/gc/ping/dolt clean-databases in proxied-server mode

coffeegoddd:
## Summary

Adds proxied-server-mode support for four maintenance commands that previously hard-rejected (or silently mishandled) proxied mode: **`bd ping`**, **`bd gc`**, **`bd compact`** (both the top-level history-squash and `bd admin compact --dolt`), and **`bd dolt clean-databases`**.

In proxied mode the CLI talks to a per-workspace TCP proxy fronting a Dolt SQL server and runs everything through the `uowProvider` rather than the legacy `storage.Store`. Each command now follows the established `if usesProxiedServer() { return runXProxiedServer(...) }` dual pattern.

## Core infrastructure

Several of these operations (`CALL DOLT_GC`, `CALL DOLT_STATS_GC`, `DROP DATABASE`, and the `DOLT_BRANCH`/`DOLT_RESET`/`DOLT_CHERRY_PICK` squash sequence) **cannot run inside an explicit transaction**, but the proxied `uowProvider` opens `START TRANSACTION` eagerly. So we add a non-transactional path:

- `internal/storage/uow/maintenance.go` — `MaintenanceProvider` interface + `doltSQLProvider.RunNonTx`, which pins a raw `*sql.Conn` (no `START TRANSACTION`).
- `cmd/bd/proxied_maintenance.go` — `runProxiedNonTx` command-layer helper.

A `*sql.Conn` satisfies `versioncontrolops.DBConn`, so the existing SQL-based version-control ops (`Compact`, `Log`, `DoltGC`, `PruneRemoteRefs`, `ListTags`) are reused directly — no duplicated logic.

## Per-command

| Command | Behavior in proxied mode |
|---|---|
| `bd ping` | Trivial `SELECT 1` round-trip via `RunTxRead`; human + JSON output |
| `bd gc` | All 3 phases: decay (`IssueUseCase` search+delete), compact-info (`SELECT COUNT(*) FROM dolt_log`), Dolt GC (`DOLT_GC` + `DOLT_STATS_GC` on a non-tx conn) |
| `bd compact` (top-level) | Full commit-history squash via `versioncontrolops.Compact` on one pinned conn, then prune remote refs + Dolt GC |
| `bd admin compact --dolt` | Dolt GC via SQL; other modes return a clear scoped rejection |
| `bd dolt clean-databases` | `SHOW DATABASES` / `DROP DATABASE` on a non-tx conn (builds its own provider, since `bd dolt` subcommands skip provider init) |

## Testing

New gated integration suite `maintenance_commands_proxied_integration_test.go` (runs with `BEADS_TEST_PROXIED_SERVER=1`) covering, per command: dry-run, real mutation with data-preservation/idempotence checks, refuse-without-force, skip flags, JSON output, and the clean-databases non-stale-DB safety guarantee. This brings proxied coverage to meet-or-exceed the (weaker/absent) embedded tests for these commands.

Verified end-to-end on a real managed proxied workspace. `go build`, `go vet`, `gofmt`, and `golangci-lint` (0 issues) all clean.

### DIFF
--- a/cmd/bd/compact.go
+++ b/cmd/bd/compact.go
@@ -83,7 +83,10 @@ Examples:
 `,
 	RunE: func(_ *cobra.Command, _ []string) error {
 		if usesProxiedServer() {
-			return HandleErrorRespectJSON("admin compact is not supported in proxied-server mode")
+			if compactDolt {
+				return runCompactDoltProxiedServer(rootCtx)
+			}
+			return HandleErrorRespectJSON("only 'compact --dolt' is supported in proxied-server mode")
 		}
 		// Block mutating operations in embedded mode; allow --stats, --analyze, --dry-run read-only paths.
 		if !compactStats && !compactAnalyze && !compactDryRun {
--- a/cmd/bd/compact_dolt.go
+++ b/cmd/bd/compact_dolt.go
@@ -46,7 +46,7 @@ Examples:
 	SilenceErrors: true,
 	RunE: func(_ *cobra.Command, _ []string) error {
 		if usesProxiedServer() {
-			return HandleErrorRespectJSON("compact is not supported in proxied-server mode")
+			return runCompactProxiedServer(rootCtx)
 		}
 		evt := metrics.NewCommandEvent("compact")
 		defer func() {
--- a/cmd/bd/compact_dolt_proxied_server.go
+++ b/cmd/bd/compact_dolt_proxied_server.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/steveyegge/beads/internal/storage/versioncontrolops"
+)
+
+func runCompactDoltProxiedServer(ctx context.Context) error {
+	start := time.Now()
+
+	if compactDryRun {
+		if jsonOutput {
+			return outputJSON(map[string]interface{}{
+				"dry_run": true,
+			})
+		}
+		fmt.Printf("DRY RUN - Dolt garbage collection\n\n")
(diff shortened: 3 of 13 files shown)


coffeegoddd:
Addressed all six Copilot comments in 725abc06e:

1. **`scalarCount` types** (gc_proxied_server.go) — now handles `[]byte`, `uint64`, and `float64` in addition to `string`/`int64`. (Note: `RawSQLRepository.Query` already converts `[]byte`→`string`, so the compact-info count was correct on the current path, but the extra cases make `scalarCount` robust to driver/representation changes.)
2. **`runProxiedNonTx` JSON output** (proxied_maintenance.go) — both the nil-provider and unsupported-provider branches now use `HandleErrorRespectJSON` so `--json` callers get structured errors.
3. **Swallowed `rows.Scan` error** (clean-databases) — now returns `HandleError("scanning database name: ...")` instead of `continue`, so decode failures fail loudly rather than silently skipping databases.
4. **Ignored `fmt.Sscanf`** (test helper) — `proxiedCommitCount` now checks the scan error and `t.Fatalf`s on failure.
5 & 6. **'✓ GC complete' on Dolt GC failure** (gc) — added an overall `gcFailed` flag: the summary now prints `⚠ GC completed with errors`, the JSON summary sets `success: false`, and the command exits non-zero when the Dolt GC phase fails (non-dry-run).

`go build`, `go vet`, `gofmt`, `golangci-lint` (0 issues), and the full proxied integration suite all pass.

coffeegoddd:
Superseded by #5003, which is opened from the fork branch coffeegoddd:db/more-commands (same commits).

PR #5001, opened 2026-07-23
bd sql (proxied-server): multi-statement batches and --database flag

coffeegoddd:
## Summary

Two related improvements to `bd sql` in proxied-server mode, both landing entirely on the client side (the transparent proxy and Dolt server already support the wire behavior):

### 1. Multi-statement SQL

`bd sql` now accepts a `;`-separated batch of statements. Any multi-statement input is treated as a write: it executes atomically in a single unit-of-work transaction, commits via `DOLT_COMMIT`, and reports `OK` (JSON `{"status":"ok"}`). Single statements are unchanged — reads render a table, single writes still report `OK, N rows affected`.

Multi-statement detection is a lightweight top-level `;` scan that ignores semicolons inside `'…'`, `"…"`, `` `…` ``, `-- `, `#`, and `/* … */`. A single multi-line `SELECT` stays a single-statement read.

This routing was chosen after verifying `go-sql-driver/mysql` behavior against a real proxied Dolt server: `ExecContext` reports only the *last* statement's `RowsAffected` (not a sum), and `QueryContext` never surfaces write counts — so an honest per-batch row count isn't available, hence the plain `OK`.

### 2. `--database` flag

`bd sql --database <name> '<query>'` runs the query against a different database on the same proxied server — the equivalent of a session `USE <name>` — without changing the project's configured database. Implemented by switching the database on the pinned UOW connection before executing (verified that `USE` inside `START TRANSACTION` works and that a subsequent `DOLT_COMMIT` targets the switched database, leaving the default database untouched).

- Works for single reads, single writes, and multi-statement batches.
- The database name is validated (`^[a-zA-Z_][a-zA-Z0-9_]*$`) before interpolation — no injection via the flag.
- Only supported in proxied-server mode; direct-server mode returns a clear error rather than silently ignoring it.

## Implementation notes

- `internal/storage/uow/uow.go`: added `SwitchDatabase(ctx, database)` to the `UnitOfWork` interface / `baseUOW`, delegating to the existing identifier-validated `DDLSQLRepository.UseDatabase`.
- `cmd/bd/sql_proxied_server.go`: multi-statement detection + routing; both read and write paths switch database on the same pinned session.
- `cmd/bd/sql.go`: registered `--database`, help text, non-proxied guard.

## Testing

- `TestTopLevelStatementCount` / `TestIsMultiStatementSQL` — statement-counter unit tests (quotes, comments, escapes, multi-line).
- `TestProxiedServerMultiStatementSQL` — end-to-end: multi-statement batch commits atomically and prints `OK`; single write keeps its count; single read still renders; `--json` multi-statement omits `rows_affected`.
- `TestProxiedServerSQLDatabaseFlag` — end-to-end: `--database` write commits to the other database, unqualified read resolves there, invalid identifier rejected.

All pass against a real proxied Dolt server (`BEADS_TEST_PROXIED_SERVER=1`); `go vet`, `gofmt`, and full build clean.

### DIFF
--- a/cmd/bd/sql.go
+++ b/cmd/bd/sql.go
@@ -29,6 +29,11 @@ The query is passed directly to the database. SELECT queries return results as a
 table (or JSON/CSV with --json/--csv). Non-SELECT queries (INSERT, UPDATE, DELETE)
 report the number of rows affected.
 
+Multiple statements separated by ';' run as a single committed batch and report
+"OK". In proxied-server mode, --database runs the query against a different
+server database (equivalent to a session USE) without changing the project's
+configured database.
+
 WARNING: Direct database access bypasses the storage layer. Use with caution.`,
 	Args:          cobra.ExactArgs(1),
 	SilenceUsage:  true,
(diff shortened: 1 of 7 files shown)


coffeegoddd:
Superseded by #5002, which is opened from the coffeegoddd:db/sql fork branch.

PR #5003, opened 2026-07-23
feat: support compact/gc/ping/dolt clean-databases in proxied-server mode

coffeegoddd:
## Summary

Adds proxied-server-mode support for four maintenance commands that previously hard-rejected (or silently mishandled) proxied mode: **`bd ping`**, **`bd gc`**, **`bd compact`** (both the top-level history-squash and `bd admin compact --dolt`), and **`bd dolt clean-databases`**.

In proxied mode the CLI talks to a per-workspace TCP proxy fronting a Dolt SQL server and runs everything through the `uowProvider` rather than the legacy `storage.Store`. Each command now follows the established `if usesProxiedServer() { return runXProxiedServer(...) }` dual pattern.

## Core infrastructure

Several of these operations (`CALL DOLT_GC`, `CALL DOLT_STATS_GC`, `DROP DATABASE`, and the `DOLT_BRANCH`/`DOLT_RESET`/`DOLT_CHERRY_PICK` squash sequence) **cannot run inside an explicit transaction**, but the proxied `uowProvider` opens `START TRANSACTION` eagerly. So we add a non-transactional path:

- `internal/storage/uow/maintenance.go` — `MaintenanceProvider` interface + `doltSQLProvider.RunNonTx`, which pins a raw `*sql.Conn` (no `START TRANSACTION`).
- `cmd/bd/proxied_maintenance.go` — `runProxiedNonTx` command-layer helper.

A `*sql.Conn` satisfies `versioncontrolops.DBConn`, so the existing SQL-based version-control ops (`Compact`, `Log`, `DoltGC`, `PruneRemoteRefs`, `ListTags`) are reused directly — no duplicated logic.

## Per-command

| Command | Behavior in proxied mode |
|---|---|
| `bd ping` | Trivial `SELECT 1` round-trip via `RunTxRead`; human + JSON output |
| `bd gc` | All 3 phases: decay (`IssueUseCase` search+delete), compact-info (`SELECT COUNT(*) FROM dolt_log`), Dolt GC (`DOLT_GC` + `DOLT_STATS_GC` on a non-tx conn) |
| `bd compact` (top-level) | Full commit-history squash via `versioncontrolops.Compact` on one pinned conn, then prune remote refs + Dolt GC |
| `bd admin compact --dolt` | Dolt GC via SQL; other modes return a clear scoped rejection |
| `bd dolt clean-databases` | `SHOW DATABASES` / `DROP DATABASE` on a non-tx conn (builds its own provider, since `bd dolt` subcommands skip provider init) |

## Testing

New gated integration suite `maintenance_commands_proxied_integration_test.go` (runs with `BEADS_TEST_PROXIED_SERVER=1`) covering, per command: dry-run, real mutation with data-preservation/idempotence checks, refuse-without-force, skip flags, JSON output, and the clean-databases non-stale-DB safety guarantee. This brings proxied coverage to meet-or-exceed the (weaker/absent) embedded tests for these commands.

Verified end-to-end on a real managed proxied workspace. `go build`, `go vet`, `gofmt`, and `golangci-lint` (0 issues) all clean.


maphew:
Maintainer review sweep: this is solid work — it un-bricks four maintenance commands in proxied mode via a minimal non-tx connection primitive and reuses the `versioncontrolops` helpers instead of duplicating the squash recipe, with real integration coverage. To keep it moving I pushed one small maintainer commit (21eacaef0) with three mechanical alignments:

- `runCompactDoltProxiedServer` now calls `CheckReadonly("compact")` on the non-dry-run path (the embedded dispatch in `compact.go` and your other proxied handlers already do this; this one path could run Dolt GC in readonly mode).
- Its failure path now goes through `HandleErrorRespectJSON` so `--json` callers get a structured error, matching the convention you applied elsewhere in this PR.
- `runCompactProxiedServer` `--json` now emits `remote_refs_pruned` / `tags_anchoring` as arrays, matching the embedded `compact_dolt.go` schema (it was emitting counts under the same keys).

Things that genuinely need your judgment (none blocking merge, follow-ups welcome):

1. **Flake pressure on the shared proxied server**: the new parallel subtests run `CALL DOLT_GC` and `CREATE/DROP DATABASE` against the shared test server, and the first CI round saw sibling tests die in `bd init --proxied-server` with the known transient "no root value found in session" race. The legacy dolt store retries that error but the uow provider's init/migrate path apparently does not — worth either adding the transient-retry there or softening/serializing the GC-heavy subtests.
2. **Cross-mode exit-code divergence**: proxied `bd gc` exits non-zero when the Dolt GC phase fails; embedded `bd gc` warns and exits 0. Your stricter behavior is arguably better — decide whether to backport it to embedded or match embedded, but the modes should agree.
3. **`clean-databases` circuit-breaker**: the proxied path intentionally omits the embedded batch-pause/backoff/abort machinery. Fine for per-workspace proxies; please confirm shared multi-tenant servers can't accumulate enough stale DBs to need it.
4. **Parity ledger**: proxied gc/compact JSON lacks the embedded store-size before/after fields, and proxied gc omits the remote-refs/tags anchoring note. A line in the PR body or a follow-up issue would keep the parity story honest.

Plan is to merge once CI comes back green on the new head.

PR #4742, opened 2026-07-12
fix(proxied): fresh-UOW retry for Dolt serialization conflicts

medhatgalal:
## Summary

**Correctness fix for proxied-server (Dolt SQL server) mode:** when a Dolt serialization conflict (MySQL 1213/1205) occurs, replay the **entire high-level operation** on a **fresh unit of work / snapshot**, instead of retrying only `DOLT_COMMIT` on a transaction that has already lost.

This is a storage/CLI correctness slice. It is **not** the WAN performance fix. Lab verdict remains `GATEWAY_ARCHITECTURE_REQUIRED` until an integrated near-data cell meets promotion gates on current source.

## What was fixed

| Problem | Fix |
| --- | --- |
| `CommitWithRetries` only re-ran `DOLT_COMMIT` after 1213 | Added `uow.RunWithFreshUOWRetries` / `RunWithFreshUOWRetriesDynamicMessage` |
| Proxied update/claim/close/create/dep/reopen/delete could commit partial/stale state | Each path now reopens UOW, re-reads, re-applies, then commits |
| Failed `START TRANSACTION` could leak pinned connections | `discardSQLConn` on BeginTx failure |
| Canceled command context could skip ROLLBACK | UOW `Close` uses bounded `context.WithoutCancel` cleanup |
| Non-serialization commit failure could pool dirty sessions | ROLLBACK (or poison) before release — preserves existing tests |

## How it was fixed

1. **Kernel:** `internal/storage/uow/retry.go` — full-operation retry with backoff on serialization errors only; connection loss on commit is **not** blindly retried (ambiguous commit).
2. **CLI surface:** proxied handlers under `cmd/bd/*_proxied_server.go` call the kernel instead of `CommitWithRetries` for the listed ops.
3. **Tests:** `cmd/bd/fresh_uow_retry_test.go` proves 1213 forces a second `NewUOW` + re-apply; `internal/storage/uow` keeps pool-safety tests green.

## What this PR replaces / relates to

| PR | Relationship |
| --- | --- |
| Operator Draft **#1** (UOW on 4675 base) | **Superseded as the CI vehicle.** Same idea, wrong base for fork workflows (`pull_request` only targets `main`). Keep #1 as archaeology of the 4675 stack. |
| Upstream **#4675** (claim/lease) | **Not replaced.** Coordinate; this PR does **not** re-arm leases on bare status updates (compatible with main lease fixes). |
| Operator Draft **#2** (gateway lab) | **Not replaced.** Research/evidence only. |
[...]

Issue #4637, opened 2026-07-07
shared-server connect path performs no server-identity check; can silently connect to a foreign dolt server (WSL2/Windows loopback in the wild)


seanmartinsmith:
## problem
bd's server-mode connect path (`newServerMode`) treats "TCP dial to the configured host:port succeeded" as "this is our server" — it never verifies which data directory the answering process actually serves. the one check that does verify identity (`isProcessInDir`, comparing a candidate PID's CWD against the expected dolt data dir) is wired into exactly one place, `reclaimPort`, which only runs when `Start()` is asked to launch a *new* server and finds the port already occupied. when a server is already listening at the moment bd dials — the normal case for a shared/discovered server — the dial just succeeds and bd connects straight through; `reclaimPort`/`isProcessInDir` never run.

on windows this is structurally worse: `isProcessInDir` is hardcoded to return `false`, because CWD isn't exposed through the standard win32 process APIs. so even in the one path that does attempt verification, windows can never adopt its own server there (false negative) — and on the direct-connect path (the actual bug here), windows is exactly as blind as unix, since no identity check happens on any platform once the dial succeeds.

concretely: on a machine running both Windows and WSL2, WSL2's default networking mirrors a WSL-bound `127.0.0.1:<port>` onto the Windows loopback. if the WSL side's bd starts a shared dolt sql-server on the shared/expected port, Windows bd's dial to that same port succeeds against the *WSL* server, and bd proceeds to query/write against it with no warning — silently reading and writing the wrong project's database. confirmed in the field: Windows-side `bt` showed 4 projects (WSL's repo set) instead of the 18 in the Windows-side `~/.beads/shared-server/dolt`.

## versions
bd v1.1.0 (commit 8e4e59d39); confirmed present at main @187ba85f3. Windows + WSL2 on the same host

## location
- `internal/storage/dolt/store.go:1036-1047` (upstream/main @187ba85f3) — `newServerMode`'s fail-fast dial (`net.DialTimeout(...)`); when `dialErr == nil` (dial succeeds), execution falls straight through to opening the connection with zero identity check. auto-start (the only caller of `doltserver.EnsureRunningDetailed` → `reclaimPort`) only triggers when `dialErr != nil`.
- `internal/doltserver/doltserver.go:360` (`func reclaimPort`) / `:387` (`if isProcessInDir(pid, doltDir)`) — the only call site of the identity check anywhere in the codebase; unreachable from the connect path above.
[...]

Issue #4513, opened 2026-06-29
Expose a stable doltserver interface for external lifecycle managers to verify pid-birth coherence


sjarmak:
**Problem.** External supervisors that manage a beads dolt sql-server (Gas City) must answer "is the PID I recorded still the *same* process I started, or a reused PID?" Beads' own liveness check can't answer this, and exposes no API that can, so downstream re-implements process-birth-time validation independently and per-platform.

**Root cause (origin/main, `fa4dce454`).** Beads validates a managed server with `isProcessAlive` (a `kill -0` existence check, `internal/doltserver/doltserver_unix.go:120-127`) plus `isDoltProcess` (PID-in-dolt-name-list, `doltserver.go:398-406`). Neither captures nor compares a process start time, so a PID reused by an unrelated — or even a *different* — dolt sql-server passes both. The exported `State` (`doltserver.go:305-310`) carries `Running/PID/Port/DataDir` and **no** birth field, and there is no exported helper taking `(pid, birthTimestamp) -> bool`.

**Why it matters to a downstream consumer.** Gas City re-implements exactly this in `cmd/gc/dolt_start_managed.go` — `StartTimeTicks` read from `/proc/<pid>/stat` field 22 at registration and compared on every health/reap pass (`:38-44`, `:692-741`) — and repeats the pattern in `internal/workspacesvc/orphan_reap.go` and `cmd/gc/cmd_start_drift.go`. This duplicate is Linux-`/proc`-only, drifts from beads' internal lifecycle assumptions, and must be re-validated against every beads upgrade. A first-class beads API would let downstream delete it.

**Dedup.** `pid birth coherence interface`, `process start time lifecycle external` → no matching issue (hits #2050, #2685, #3142 are closed and unrelated). No in-flight work on a birth-coherence API in the doltserver history.

**Fix candidates.**
1. Add a `BirthToken` (opaque, platform-resolved start-time/ctime) to `State`, populated by `Start`/`IsRunning`, plus an exported `VerifyProcessIdentity(pid int, token BirthToken) (bool, error)` that confirms the live process matches — beads already has the platform-split files (`doltserver_unix.go`/`_windows.go`) to host the resolver.
2. Lighter: export the existing internal coherence check as `ProcessMatches(pid int) bool`, strengthened to include start-time, documented as the supported reattachment-validation entry point.
[...]

PR #5004, opened 2026-07-24
test: managed-local proxied lifecycle smoke lane (Linux, offline)

maphew:
## Why

Every proxied-server CI shard added by #4765 initializes test projects with `--proxied-server-external-host` against a shared Dolt testcontainer. That tier is valuable, but it means a green proxied run has never exercised the production **managed-local** path: `bd` launching its own loopback proxy, that proxy launching a locally installed `dolt sql-server`, and the topology surviving idle shutdown and transparent restart. This PR adds the first end-to-end lane for exactly that path. It is test-only — no product behavior changes.

This is the narrow managed-local slice of the broader cross-mode coverage gap tracked in #2764.

## What

- **`cmd/bd/proxied_local_helpers_test.go`** — a separate gate for this lane (`BEADS_TEST_PROXIED_LOCAL=1`). When the lane is requested, a missing `dolt` binary **fails** the test rather than skipping, so a broken prerequisite cannot produce a silently green run. Includes managed-local init (asserts no `External` block), pidfile/process/artifact inspection primitives, and a held-connection helper that keeps the proxy's idle watcher disarmed while the process tree is inspected. These helpers are deliberately reusable for future stale-artifact and process-identity lifecycle tests.
- **`cmd/bd/proxied_local_lifecycle_linux_test.go`** — `TestManagedLocalProxiedLifecycleSmoke`: init/create/read against the genuinely launched process tree (`bd db-proxy-child` → `dolt sql-server`); loopback-only proof both static (generated `config.yaml` marker, host, and port) and dynamic (enumerating the live processes' listening TCP sockets via `/proc`); process-identity sanity via `/proc/<pid>/cmdline`; observed idle shutdown (short `--proxied-server-idle-timeout` window); transparent restart with persisted data and a fresh proxy pid.
- **`.github/workflows/proxied-local-smoke.yml`** — pins the Dolt CLI (2.2.2), builds `bd` and the test binary online, then runs the test inside `unshare --net` with only loopback up — after probing that outbound really is unreachable — proving the whole lifecycle needs no network once binaries are installed.

## Evidence

Verified on Linux (dolt 2.2.0 locally):

- plain run passes in ~19–21s; the identical test passes inside a loopback-only network namespace;
- negative check: with `dolt` removed from `PATH` and the lane env set, the test fails with a clear message instead of skipping;
- full `make test` passes at this head.

One subtlety worth reviewer attention: releasing a held `sql.Conn` back to its pool keeps the TCP session to the proxy open, which still counts as an active connection and blocks idle shutdown indefinitely — the helper therefore closes the whole `sql.DB` and documents why (`heldProxiedConn.Release`).

## Non-overlap

- #2764 — broad cross-mode/CI coverage; this is the managed-local lane it calls for, not a competing scope.
- #4547 — capability seam for proxied command parity; untouched.
- #4513 / #4637 — PID/birth coherence and server identity design; this lane only adds *test-side* identity sanity checks and leaves the hardening design to those issues. The custom-config listener-policy question (validation currently parses but does not enforce the listener host of user-supplied Dolt YAML) is likewise left to follow-up hardening; this lane proves the default generated topology.
- #4742 — serialization retry draft; no interaction (this lane is single-writer by design; concurrent-writer coverage is a follow-up).

### DIFF
--- a/.github/workflows/proxied-local-smoke.yml
+++ b/.github/workflows/proxied-local-smoke.yml
@@ -14,6 +14,7 @@ name: Proxied Local Smoke
 
 on:
   workflow_dispatch:
+  merge_group:
   push:
     branches: [main]
     paths:
@@ -21,9 +22,17 @@ on:
       - 'cmd/bd/proxied_server*.go'
       - 'cmd/bd/*_proxied_server.go'
       - 'cmd/bd/db_proxy_child.go'
+      - 'cmd/bd/init*.go'
+      - 'cmd/bd/proxied_integration_helpers_test.go'
+      - 'cmd/bd/store_factory*.go'
+      - 'cmd/bd/test_dolt_server_cgo_test.go'
       - 'cmd/bd/uow_factory.go'
+      - 'internal/configfile/**'
+      - 'internal/doltserver/**'
       - 'internal/storage/dbproxy/**'
       - 'internal/storage/uow/**'
+      - 'go.mod'
+      - 'go.sum'
       - '.github/workflows/proxied-local-smoke.yml'
   pull_request:
     paths:
@@ -31,9 +40,17 @@ on:
       - 'cmd/bd/proxied_server*.go'
       - 'cmd/bd/*_proxied_server.go'
       - 'cmd/bd/db_proxy_child.go'
+      - 'cmd/bd/init*.go'
+      - 'cmd/bd/proxied_integration_helpers_test.go'
+      - 'cmd/bd/store_factory*.go'
+      - 'cmd/bd/test_dolt_server_cgo_test.go'
       - 'cmd/bd/uow_factory.go'
+      - 'internal/configfile/**'
+      - 'internal/doltserver/**'
       - 'internal/storage/dbproxy/**'
       - 'internal/storage/uow/**'
+      - 'go.mod'
+      - 'go.sum'
       - '.github/workflows/proxied-local-smoke.yml'
 
 permissions:
@@ -45,6 +62,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 25
     env:
+      CGO_ENABLED: "1"
       # Pinned so the lane tests a known dolt CLI, not whatever the installer
       # resolves on a given day. Bump deliberately.
       DOLT_VERSION: 2.2.2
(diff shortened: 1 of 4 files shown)


maphew (commented):
Automated adversarial review at `9bcfcbff` found a valuable lane with three CI-integrity blockers. The managed-local lifecycle itself is real and passed three consecutive local runs, but the gate can currently lose its signal:

1. **CGO-off is a silent green no-op.** Both new test files require `cgo`, while the workflow neither pins `CGO_ENABLED=1` nor asserts that the compiled binary contains `TestManagedLocalProxiedLifecycleSmoke`. Reproduced:

   ```text
   CGO_ENABLED=0 go test -tags gms_pure_go \
     -run '^TestManagedLocalProxiedLifecycleSmoke$' -v ./cmd/bd
   testing: warning: no tests to run
   PASS
   ```

2. **The path filters omit direct dependencies.** In particular, changes to `cmd/bd/init.go`, `cmd/bd/init_proxied_server.go`, `cmd/bd/proxied_integration_helpers_test.go`, `cmd/bd/test_dolt_server_cgo_test.go`, `cmd/bd/store_factory*.go`, `internal/configfile/**`, `internal/doltserver/**`, and `go.mod`/`go.sum` can break this topology without running its only lane. The workflow also lacks `merge_group`.

3. **The “no testcontainer” lane still enters the package-wide shared-Dolt setup.** `startTestDoltServer` exempts embedded and external-proxied tests, but not `BEADS_TEST_PROXIED_LOCAL`. My exact local smoke started the cached `dolthub/dolt-sql-server:2.2.0` container first; hosted CI only avoided that because the image was absent and logged a skip warning.

Two assertions should be tightened in the same pass:

- require each pidfile process to own its advertised loopback port, not merely some loopback listener;
- assert the requested five-second idle timeout was persisted, so falling back to the 30-second default cannot pass inside the 90-second shutdown allowance.

The core design held up: managed-local selection is genuine, the held pool is released correctly, cleanup is registered early, and config/socket/CRUD/SQL/shutdown/restart checks provide complementary signal.

Maintainer outcome: **fix-merge**. I’m applying these localized fixes directly to this writable branch now; no replacement PR is warranted.

maphew:
Fixes from the adversarial review are now pushed in `33ed64122`:

- makes the lane explicitly CGO-enabled and fails closed unless the exact lifecycle test is present in the compiled binary;
- adds `merge_group` coverage and expands the workflow's direct dependency paths;
- bypasses the package-wide Docker fixture only for the explicit managed-local smoke mode;
- proves the persisted idle timeout is exactly `5s`; and
- proves each pidfile process owns its advertised `127.0.0.1:port`, while still rejecting non-loopback listeners.

Local verification passed:

- focused listener-validation test;
- lifecycle smoke three consecutive times;
- CGO-on binary lists the target test;
- CGO-off binary omits it, which the workflow guard rejects; and
- `git diff --check`.

An independent adversarial re-review found no blocking findings. The refreshed hosted `Managed-local proxied lifecycle (Linux, offline)` lane is now queued.

maphew:
Status note: the sole failing check, Test Nix Flake, is pre-existing on `main` (its latest run fails the same job) and is unrelated to this test-only diff. The fix is #5007 (vendorHash for the Go 1.26.5 module set). Everything else is green at `33ed64122`, including the refreshed hosted Managed-local proxied lifecycle lane, and the local full `make test` verification queue is re-running against this head.

PR #5006, opened 2026-07-24
fix(claim): verify claim-family writes by re-read under a degraded server (bd-zccb9)

steveyegge:
## Summary

Fixes bd-zccb9 (upstream ask from wyvern incident wy-ejph3): under a degraded dolt sql-server, the exit status of a claim-family write is not truth in either direction —

- `bd update --claim` reported **success**, but the client hit an i/o timeout mid-transaction under server overload, the server-side txn rolled back, and the claim was gone ~30 minutes later. Cost: a duplicate full implementation of a bead (wy-ejph3).
- The inverse is also on record: `dolt commit: invalid connection` printed, but the write **actually applied** (wy-x543k).

Wyvern's fleet mandates a verify-by-re-read wrapper (`wh-take.sh`) as a workaround. This PR folds that protocol into bd natively so bare claims are safe.

## Design

New `internal/storage/dolt/claim_verify.go` — `verifiedClaimWrite` resolves claim-family outcomes against the database instead of trusting exit status:

1. **Reported success** → verify by re-read on a fresh connection; a mismatch fails LOUDLY ("treat the claim as NOT applied") and bumps `bd.claim_verify_lost_total`.
2. **Indeterminate commit-phase loss** (`errCommitPhase`, the #4462 phase-aware tag) → the re-read settles it: verified applied → accurate success (converts the wy-x543k false-failure); verified rolled back → replay once (safe — nothing landed; converts the wy-ejph3 storm case). Both conversions count in `bd.claim_verify_recovered_total` (`op`, `outcome=applied|replayed`).
3. **Any other error** → honest failure, exactly as before.

Wired into `ClaimIssue`, `UnclaimIssue`, `UnclaimIssueIfAssignee` (via wrapper) and `ClaimReadyIssue` (bespoke: the claimed ID is only known after the body runs, and a replay after a verified rollback legitimately may claim a *different* ready issue). Wisps and non-server (embedded) mode are exempt and unchanged. Ordinary field updates are out of scope by design — a lost label edit is an annoyance; a phantom claim is a duplicated implementation.

Also includes `PROPOSAL-cas-conditional-update.md` (separate commit) — fly's design doc for the downstream general CAS bead bd-wsqvw, committed here so the untracked doc survives workspace cleanup; bd-wsqvw depends on this PR's `verifiedClaimWrite`.

## Testing

- `claim_verify_test.go`, 5 tests against a real Dolt container: applied-indeterminate converts to success, verified-rollback replays, lying-success fails loudly, replay budget exhausts (indeterminate twice), e2e claim/unclaim + idempotent re-claim. All pass.
- Full `./internal/storage/dolt/` package suite, run locally (macOS/OrbStack laptop): 6 tests failed under full-parallel load (`TestSerializationConflictRetry`, `TestConcurrentWorkQueueDrain`, 4× `TestCrossProject_*`), every one at the same 45s `failed to initialize schema: context deadline exceeded` during embedded store *creation* — before any claim path runs, and in the mode `verifiedClaimWrite` explicitly exempts. All 6 pass in isolation on this branch (4–18s each). Attributed to local resource starvation, same class as the known load-sensitivity of this suite; CI's macOS full run is the authoritative check.
- `cmd/bd` Claim/Unclaim/Update suites (`claim_pools_embedded`, `unclaim*`, `update_embedded`, `update_proxied`): pass.
- `go build ./...`, `go vet`, `gofmt`, golangci-lint (pre-commit): clean.

Known pre-existing flake noted during development: `TestConcurrentHeartbeatReclaimClose` is load-sensitive under full parallel claim-suite load (real-time 500ms-heartbeat/2s-TTL deadlines vs shared-server tail latency); its race phase never touches the wrapped paths (heartbeat/reclaim/close are unwrapped). Investigated and attributed on bd-zccb9 — happy to sequentialize or loosen its TTL in this PR if preferred.

Closes bd-zccb9.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   every remote-ahead case, as before
   ([#4259](https://github.com/gastownhall/beads/issues/4259)).
 
+### Fixed
+
+- **Claim-family writes are now verified by re-read in Dolt server mode**
+  (bd-zccb9, from wyvern incident wy-ejph3). Under a degraded sql-server the
+  exit status of `bd update --claim` / `bd claim` / unclaim was not truth in
+  either direction: a claim could report success while the server-side
+  transaction died with the abandoned connection and rolled back (a phantom
+  claim that later cost a duplicate implementation), and conversely a
+  connection error could print with the write actually applied. After the
+  claim transaction, bd now re-reads the issue's assignee and status on a
+  fresh connection and resolves the outcome against the database: a reported
+  success that did not land fails loudly ("treat the claim as NOT applied");
+  an ambiguous commit-phase loss is settled by the re-read — verified applied
+  becomes an accurate success, verified rolled back is replayed once (safe:
+  nothing landed). Applies to claim, ready-claim, and unclaim in server mode;
+  wisps and embedded mode are unchanged. New metrics
+  `bd.claim_verify_lost_total` and `bd.claim_verify_recovered_total` count
+  loud failures and converted outcomes.
+
 ## [1.1.0] - 2026-07-04
 
 First stable release of the 1.1.0 line. It consolidates everything from
--- a/PROPOSAL-cas-conditional-update.md
+++ b/PROPOSAL-cas-conditional-update.md
@@ -1,0 +1,167 @@
+# PROPOSAL: general conditional (compare-and-set) update for `bd`
+
+**Status:** design, ready for implementation as a beads PR
+**Author:** design handoff (wyvern hostile-review epic wy-mdi5h)
+**Repo:** `github.com/steveyegge/beads` (`~/src/beads`, branch `main` @ `3a6dcff96`)
+**Motivating incidents:** wyvern wheelhouse review findings wy-mdi5h.3, .6, .7, .4 (TOCTOU park/restore/take/bulldog); upstream lineage bd-zccb9, #4727, wy-ejph3, wy-x543k.
+
+---
+
+## 1. Problem
+
+A hostile review of wyvern's wheelhouse fleet scripts found that its single most
+common bug class is **check-then-act on `bd` assignee/status**: park, interactive
+restore, queue-take, and bulldog-claim each read a bead's state, then issue a
+*blind* `bd update -a X` / `-s Y`, and lose to a racing writer in the gap. The
+scripts paper over it with re-reads (`wh-take.sh`), but re-reads only narrow the
+window; they cannot close it. The root cause is not a wheelhouse bug — it is that
+**`bd` exposes no general compare-and-set for the coordination fields.**
+
+`bd` already recognizes this need and has been closing it verb by verb:
+
+| primitive | CAS semantics | added |
+|---|---|---|
+| `bd update --claim` | `assignee ∈ {'', pool-alias} → me`, `status → in_progress` | (existing) |
+| `bd unclaim --if-assignee X` | `assignee == X → ''`, `status → open` | #4727 |
+| `verifiedClaimWrite` (WIP) | verify-after-write for the above under a degraded server | bd-zccb9 (uncommitted in tree) |
+
+The pattern is a `UPDATE … WHERE id=? AND status IN(…) AND assignee=?` with
+`RowsAffected` as the verdict and a `row_lock` rewrite so a racing reclaim/close
+conflicts instead of silently merging (`issueops.UnclaimIssueIfAssigneeInTx`).
+This proposal **generalizes that idiom once** instead of minting a third and
+fourth special-case verb, and covers the two transitions no existing verb can
+express.
+
+## 2. The gap
+
+Two wheelhouse transitions have no primitive:
+
+- **Reassign `X → Y`** (park): "set assignee to `mayor` only if it is still the
+  worker I just reclaimed." `--claim` only does `'' → me`; `unclaim --if-assignee`
+  only does `X → ''`. Nothing does `X → Y`.
+- **Claim-on-behalf with a status guard** (restore): "set assignee to the crew
+  owner *and* status to `in_progress`, only if the bead is still `open`/unassigned."
+  `--claim` always claims for `$BEADS_ACTOR`, never for a third party, and takes
+  no status precondition.
+
+Queue-take (`wh-take.sh`) and bulldog-claim are **not** new-primitive gaps — they
+are adoption gaps: `--claim` already handles pool aliases (`claim.pools`, e.g.
+`fable-crew`) atomically, so both should simply call `--claim` and honor its
+non-zero exit instead of a blind `-a` + swallowed refusal. This proposal fixes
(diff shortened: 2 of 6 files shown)


steveyegge:
@coffeegoddd — requesting a Dolt-semantics review on this one (and the stacked #5008). The design rests on four assumptions about dolt sql-server behavior that came out of incident forensics rather than documented guarantees, and you're far better placed to confirm or correct them:

1. **Commit-phase atomicity under connection loss.** When a client connection is abandoned mid-`DOLT_COMMIT` (i/o timeout, server overload), we assume the server resolves the transaction atomically — full rollback or full apply, never a partial state. The verify-by-re-read protocol depends on the re-read observing one of exactly those two outcomes.

2. **Cross-connection read-your-settled-writes.** After the write connection dies, we re-read on a *fresh* connection (`readClaimState`) and treat what it sees as the settled truth. Is there any window where a fresh connection can observe pre-commit state for a transaction the server has actually applied (or vice versa)?

3. **row_lock cell-rewrite conflict guarantee.** The whole CAS scheme (here and in the #5008 field guards, same invariant as #4911's version CAS) assumes that two transactions rewriting the same `row_lock` cell reliably produce a commit-time conflict (1213/1205) under Dolt's optimistic cell-level merge — never a silent cell merge. We believe this from dolthub's "hold my beer" post + observed behavior; a confirmation that it's a stable contract would be valuable.

4. **Serialization failure ⇒ server-side rollback.** `withRetryTx` replays on 1213/1205 assuming the server has already rolled the losing transaction back completely, so the replay starts from clean state.

If any of these are weaker than assumed, the failure modes are exactly the phantom-claim / false-failure incidents this PR exists to close (wy-ejph3 / wy-x543k in the PR body), so a "yes, guaranteed" or "no, here's the edge" on each would be the most useful possible review.

steveyegge:
## Review verdict (lion): APPROVE with one should-fix on the ready-claim replay leg

Full diff read; CI checked (all 68 substantive checks green, including the Embedded Dolt Storage/Cmd, Proxied Cmd, and Server Conformance lanes — see note on the Nix job below). The design is right: postcondition-by-re-read is exactly the wh-take.sh protocol, the three-way outcome resolution (loud lost-write / verified-applied success / verified-rollback single replay) matches both incident directions, and the exemptions (wisps, embedded mode, ordinary field updates) are the correct scope line. The five injection tests hit each protocol arm, and the concurrent-mutation-reads-as-lost-write trade-off is documented honestly in the `verifiedClaimWrite` doc comment. Your local 6-failure attribution (embedded store creation timeout, pre-claim-path, isolated-green) is consistent with CI's full green run.

### Finding (should-fix before merge): `ClaimReadyIssue`'s replay leg skips verification

In `verifiedClaimWrite`, a replay that itself dies at commit phase re-enters the loop: the second `errCommitPhase` gets its own re-read, so an applied-but-error-printed replay converts to an accurate success (`TestVerifiedClaimWriteConvertsAppliedIndeterminate` semantics), and a genuine second rollback exhausts the budget with the honest "did not land" error.

The bespoke `ClaimReadyIssue` path diverges on exactly that leg (issues.go, in the replay branch):

```go
if err = write(); err != nil {
    return nil, err
}
```

A replayed write that hits `errCommitPhase` returns the raw indeterminate error with **no verify pass** — reintroducing the wy-x543k false-failure for the replay attempt. Worse than cosmetic here: the replay may have claimed a *different* ready issue, and if it actually applied, that issue is now claimed by an actor who was told the claim failed — an orphaned claim that sits until lease reclaim. Two consecutive commit-phase losses is precisely the degraded-server regime this PR targets, so the bespoke path should keep the wrapper's semantics:

```go
err = write()
if err != nil && !(errors.Is(err, errCommitPhase) && claimed != nil) {
    return nil, err
}
if claimed == nil {
    if err != nil { return nil, err } // commit-phase loss with nothing selected: honest error
    return nil, nil
}
continue // re-enter the loop: attempt=1 verifies the (possibly new) claimed.ID; budget already spent
```

`continue` with `attempt` already incremented past the budget gives the same terminal behavior as the wrapper (applied → recovered success; rolled back → "did not land"). A sixth injection test mirroring `ConvertsAppliedIndeterminate` but on the replay leg would pin it.

### Non-blocking notes

- **Nix flake failure is not this PR.** The `Test Nix Flake` job fails on a go-modules `vendorHash` mismatch; this PR touches no `go.mod`/`go.sum`, so the pinned hash was invalidated by something already on main since the last green run of that workflow (2026-07-17). Wants its own one-line hash bump (`sha256-g3VILiRj/vRA04ahiPDEkUYZmjYp7X393tYEaM0TPbs=` per the job log) — file it separately, don't fold it in here.
- `TestConcurrentHeartbeatReclaimClose` load-sensitivity: agreed it's out of scope; leave it un-touched in this PR.
- The PROPOSAL doc riding along as a second commit is fine by me given #5008 consumes it and retargets on merge.

Fix the replay leg (+ test), and this is a clean merge on my verdict — no need for a full re-review round, just point me at the delta commit.

steveyegge:
Delta commit for the should-fix: a03b33c9b — the ready-claim replay leg now keeps the wrapper's verify semantics.

Shape is as suggested: the replay's `errCommitPhase`-with-selection falls through to the verify pass (attempt is past the replay budget, so a second verified rollback still exhausts with the honest "did not land" error), raw errors still return for anything else, and a commit-phase loss with nothing selected returns the honest indeterminate error. To make the sixth injection test possible I extracted the resolution loop as `verifiedReadyClaim(ctx, actor, write)` — `ClaimReadyIssue` is now just the real write closure plus that call, so the test drives the seam directly like the five wrapper tests. New test `TestVerifiedReadyClaimReplayConvertsAppliedIndeterminate` mirrors `ConvertsAppliedIndeterminate` on the replay leg: verified rollback → replay lands but reports commit-phase loss → recovered success returning the claimed issue, with the claim verified in the DB.

The Nix flake failure is confirmed not-ours and already resolved upstream: the stale vendorHash (bd-tub5q, my #5009) was superseded by #4701 regenerating default.nix on main — a rebase/rerun should come back green.

FYI #5008 was rebased onto this delta (its exit-code addition is commented there).

steveyegge:
**Re-verdict on delta a03b33c9b (lion): PASS — clear to merge.**

The replay leg now keeps the wrapper's semantics exactly as requested: `errCommitPhase`-with-selection falls through to the verify pass with the budget already spent (second verified rollback exhausts honestly; applied converts to accurate success), commit-phase-with-nothing-selected returns the honest indeterminate, and everything else stays raw. The `verifiedReadyClaim` extraction is the right testability move — the sixth injection test drives the write seam through both commit-phase losses and pins the DB state afterward, which is precisely the pin I asked for. One observed non-issue, no action needed: a replay-then-applied run counts `recovered{outcome=replayed}` + `recovered{outcome=applied}` (two conversion events for one logical claim) — fine for a conversion counter, just worth knowing when reading dashboards.

Verdict stands: **APPROVE**. Merge when ready; #5008 retarget noted, I'll review it against the rebased base.

Issue #2764, opened 2026-03-22
test: embedded mode, cross-mode, and CI infrastructure gaps


rsnodgrass:
## Summary

Embedded Dolt mode has critical concurrency bugs (GH#2571 — nil-pointer panics on parallel reads) but its test suite is gated behind build tags and env vars that may not run reliably. Additionally, no tests verify behavior when switching between embedded and server modes, and the test infrastructure itself has leaked resources (GH#2542 zombie servers from test processes).

This issue tracks embedded mode test hardening, cross-mode coverage, and CI infrastructure improvements.

## Context: Why These Gaps Matter

| Issue | What Happened | Why Tests Missed It |
|-------|--------------|---------------------|
| GH#2571 | Parallel read-only `bd show` commands panic with nil-pointer in embedded Dolt engine | Embedded concurrency tests require `embeddeddolt` build tag + `BEADS_TEST_EMBEDDED_DOLT=1`; CI does run them but close-after-use path is untested |
| GH#2542 | Test infrastructure leaked dozens of `dolt sql-server` processes (~600MB each) | No `TestMain` cleanup guardian; no post-suite process audit |
| GH#2489 | Switching to Dolt broke atomic git+issues model | No test validates data portability between storage backends |
| GH#2656 | Fresh `bd init` succeeds but Dolt is unreachable | No end-to-end init→use test for either mode |

### T1: Embedded Mode Close-Then-Use Safety Test
- **File:** `internal/storage/embeddeddolt/close_safety_test.go`
- **Build tag:** `//go:build embeddeddolt`
- **Priority:** P0

GH#2571 shows nil-pointer panics when the embedded engine is accessed concurrently. A related failure mode is accessing the store after `Close()`.

**What to test:**
1. Open `EmbeddedDoltStore`, call `Close()`, then call every public method (`CreateIssue`, `GetIssue`, `ListIssues`, `UpdateIssue`, `CloseIssue`, `GetDependencies`, etc.) — verify each returns `errClosed`, not panic
2. Open store, start a long read query in a goroutine, call `Close()` from main goroutine — verify the read either completes or returns a clean error
3. Open store, start a write in a goroutine, call `Close()` from main goroutine — verify no corruption
4. Call `Close()` twice — verify idempotent (no panic or double-free)

**Key files:**
- `internal/storage/embeddeddolt/store.go` — `closed` atomic bool, all public methods
- The existing `concurrency_test.go` tests multi-process CRUD but not close-during-use

[...]

PR #5013, opened 2026-07-24
fix(dbproxy): verified identity for the managed proxied-server lifecycle (adoption, cleanup, kills)

maphew:
## Why

Managed proxied-server mode currently trusts anything that answers a TCP dial: endpoint
discovery adopts any listener on the pidfile's loopback port with no identity check, stale
cleanup and `bd dolt stop` kill whatever process currently holds a recorded PID, the
pick-then-launch port allocation has a bind race, and a SIGKILLed proxy supervisor leaves an
orphan `dolt sql-server` that cleanup never inspects (its lock was held by the supervisor,
not the child). On a workstation these are wrong-database and kill-the-wrong-process
hazards; they block treating managed-local proxied mode as production-ready
(local-first readiness campaign; related upstream context: #2764).

## What

Five commits, staged to review in order:

1. **`internal/procid`** — versioned process-birth tokens (Linux `boot_id`+starttime,
   Windows creation FILETIME, Darwin `p_starttime`) with TOCTOU-safe signaling handles
   (pidfd on Linux, re-verified process handle on Windows). The `Token`/`Verify` shape
   deliberately matches what #4513 asks `internal/doltserver` to expose, so it can be
   adopted there later; this PR intentionally does not touch doltserver.
2. **Identity publication** — pidfile schema v2 (additive: `schema`, `kind`, `birth`,
   `root_id`, `control_port`), a per-start rotated 0600 `proxy.secret`, and a loopback
   IDENT control listener; a proxy/backend that cannot prove its own identity fails
   startup instead of publishing.
3. **Verified adoption and fail-closed cleanup** — typed adoption statuses (adoption
   requires v2 record + live birth match + authenticated handshake + root/pid/birth/port
   cross-checks), quarantine-by-rename instead of deletion, handle-only kills, orphan
   backend reaping regardless of child-lock state, and a spawn-marker/stop-epoch fix so
   `bd dolt stop` cannot report success while a mid-flight start publishes a listener.
4. **Cross-review hardening** — fixes from two independent adversarial reviews (details
   below), including a fail-open (a `boot_id` read failure classified live processes as
   dead) and restoring self-healing for provably-dead records so fail-closed does not
   become fail-wedged.
5. **`bd dolt stop --force` + end-to-end lane** — a narrowly-scoped operator escape that
   verifies executable identity before signaling and quarantines after, plus six new
   managed-local lifecycle tests (`BEADS_TEST_PROXIED_LOCAL=1`) against a real
   `dolt sql-server`.

## Behavior changes to note

- Records written by older versions (schema v1) are never dial-adopted. Dead-pid legacy
  records self-heal via quarantine; a *live* pre-upgrade proxy is refused with an
  actionable error, and `bd dolt stop --force` is the supported way to retire it.
- `IsRunning`-style discovery now reports false for a running-but-unverifiable proxy;
  the lock probe still backs up migration checks.
- Proxied roots gain two files: `proxy.secret` (rotated each start, purged by
  mode-migration cleanup) and, transiently, `*.stale-*` quarantine records (30-day sweep).

## Testing

- Full `internal/storage/dbproxy/...` + `internal/procid/...` suites, with `-race`, green.
- Managed-local lane against real dolt (all green, Linux): foreign-listener rejection,
  reused-PID never killed, legacy fail-closed (free lock / held-lock refusal of a foreign
  executable / held-lock `--force` of a genuine bd), orphan-backend reaping after
  supervisor SIGKILL, 5-process cold-start convergence, stop/start interleave.
- Darwin/Windows: cross-compile + compile-only test verification; lane equivalents are
  defined in the lane header and staged for follow-up. A macOS test run before merge would
  be welcome — the darwin procid path had two review-caught bugs, now fixed and unit-tested,
  but no macOS CI exists on PRs.

## Review provenance and follow-ups

The branch was adversarially reviewed pre-submission by two independent model reviewers
(structured findings, all fixed in commit 4; write-ups available on request). Known
follow-ups deliberately out of scope: proxy data-listener port-0 allocation and
comprehensive managed listener-host policy (next PR), and dolt-backend port-collision
recovery, which needs a config-ownership contract first (will arrive as an RFC issue).
Relationship to open issues: #4513 (procid is the requested primitive, adoptable by
doltserver later), #4637 (same disease class on the shared-server connect path — not
addressed here; the dbproxy-side handshake pattern may transfer).

https://claude.ai/code/session_01WZHiYc3wjcdy1wrUP2JiQi

### DIFF
--- a/.github/workflows/proxied-local-smoke.yml
+++ b/.github/workflows/proxied-local-smoke.yml
@@ -1,16 +1,17 @@
-# Managed-local proxied smoke lane.
+# Managed-local proxied lifecycle lane.
 #
 # The proxied shards in main.yml exercise command behavior against an
 # EXTERNALLY managed Dolt testcontainer. This lane covers the production
 # local-first path those shards cannot: bd launches its own loopback proxy,
 # the proxy launches a locally installed `dolt sql-server`, and the whole
 # lifecycle (create/read, loopback-only listeners, idle shutdown,
-# transparent restart with persistence) runs INSIDE a network namespace
-# with only loopback up — proving no outbound network is needed once the
-# binaries are installed. Linux first; Windows/macOS parity is tracked
-# separately.
+# transparent restart with persistence), identity failures, orphan recovery,
+# and start/stop convergence run INSIDE a network namespace with only
+# loopback up — proving no outbound network is needed once the binaries are
+# installed. Linux first; the test lane defines Windows/macOS parity but does
+# not implement those OS-specific observation helpers yet.
 
-name: Proxied Local Smoke
+name: Proxied Local Lifecycle
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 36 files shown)


maphew:
Pushed a9cfebc25, closing the findings from a two-vendor adversarial review (Claude Opus + GPT-5.6 review agents) of head d53a1ae19. The reviews found disjoint defect sets — the pairing continues to earn its keep on lifecycle code.

**Windows correctness** — `cleanupOrphanBackend` held the process handle across `waitForRecordedProcessExit`; an open handle keeps the PID allocated on Windows, so crashed-proxy recovery timed out unconditionally. Now kill → close → wait, mirroring shutdown.go. `procid.Verify` also checks `GetExitCodeProcess` so a terminated process with a third-party handle held (Task Manager, AV) no longer reads as alive — "Verify true implies running" now holds on all platforms.

**PID-reuse races** — Linux `Open` now opens the pidfd *before* token verification (the pidfd pins the PID, making the check race-free; ENOSYS keeps the old path with a documented residual). `--force` now uses one stable handle across inspection, workspace-scope check, kill, and exit confirmation instead of re-looking-up the PID, and refuses to signal any process whose command line cannot be tied to this workspace root (Windows can't read a foreign cmdline, so it refuses with the manual remedy named).

**Recovery gaps** — a pre-v2 upgrade topology has *paired* legacy records, which the previous `CanForceStopUnverified` gate rejected outright; forced recovery now covers proxy-only, backend-only, and paired legacy records, with the backend action reported in JSON and human output. `PidfdSendSignal` ESRCH on a fatal signal is success (target already gone), so stop no longer strands stale records.

**Resilience** — transient control-listener accept errors (e.g. EMFILE) no longer tear down a healthy data-serving proxy; the accept loop retries with capped backoff. Handshake schema check relaxed to v2-or-newer so a future v3 proxy isn't misreported as an identity mismatch.

Tests added for each: paired-legacy force-stop end-to-end, foreign-workspace refusal, ESRCH-after-exit, schema-v3 adoption, and a Windows Verify-after-terminate unit test. The `BEADS_TEST_PROXIED_LOCAL=1` gated lane passes all 7. Deferred with tracking: HMAC challenge-response for the control handshake, control-listener local-DoS surface, and `os.SameFile`-based root-identity relaxation.

Issue #5012, opened 2026-07-24
flaky CI: proxied shard 1/15 — fresh-database init fails with 'pending schema migrations alter pre-existing dirty tables: issues' (+ mysql busy buffer)


maphew:
## Summary

`Test (proxied-server cmd 1/15)` is failing intermittently — 4 hits on 2026-07-24 across unrelated changes (a dependency bump, a lint-only change, a Makefile-only change ×2, including two consecutive failures on the same head), making it the dominant source of red-main noise today.

Failing signature, always the same:

```
proxied_integration_helpers_test.go: bd init --proxied-server failed: exit status 1
stderr:
[mysql] connection.go:215 busy buffer
Error: failed to open uow provider: uow: init schema: uow: migrate: pending schema migrations alter pre-existing dirty tables: issues; run 'bd dolt commit' ... (#4566)
```

Observed in `config_proxied_integration_test.go` (`TestProxiedServerConfigSetMany` subtests) and `duplicates_proxied_integration_test.go`, i.e. whatever lands on shard 1's test mix.

Runs: 30075668194 (passed on rerun), 30095342676 (passed on rerun), 30114730858 (failed on rerun too), plus one earlier same-day occurrence.

## Why this looks like a harness/pool race, not a test bug

- Each shared-harness project gets a **unique database** (`uniqueProxiedDatabase()` in `proxied_shared_harness_test.go`), so a *fresh* database reporting a pre-existing dirty `issues` working set should be impossible — the `dirtyTables` probe (`internal/storage/schema/schema.go`, `SELECT ... FROM dolt_status`) is reading working-set state that cannot belong to the database just created.
- The `[mysql] busy buffer` line from go-sql-driver immediately before the error indicates a pooled connection was reused with unread result data (classic after a context cancellation mid-scan). A poisoned/mis-scoped session on the shared server would explain `dolt_status` answering for the wrong database or wrong session state.
- The shard runs many `TestProxiedServer*` tests in parallel against one testcontainers server, which is exactly the load shape that would surface a session-reuse race in the uow provider's `openAndInitSchema` → migrate path.

## Suggested directions

1. In the uow provider's init/migrate path, pin the dirty-table probe and migration to one connection with an explicit database selection (or verify `DATABASE()` matches before trusting `dolt_status`).
2. Investigate the `busy buffer` source — if a timeout/cancel fires during init on the shared server, ensure the driver connection is discarded rather than reused.
[...]

PR #5014, opened 2026-07-24
fix(dolt): support --purge-dropped in proxied-server mode (p1-9lf)

coffeegoddd:
## Problem

`bd dolt clean-databases` read `--purge-dropped` and then dropped it on the floor before dispatching to the proxied-server implementation:

```go
purgeDropped, _ := cmd.Flags().GetBool("purge-dropped")

if usesProxiedServer() {
    return runDoltCleanDatabasesProxied(rootCtx, beadsDir, dryRun)  // purgeDropped never passed
}
```

So under `--proxied-server`, stale databases were dropped but never purged: their directories stayed under `.dolt_dropped_databases/`, disk was never reclaimed, and nothing reported that the flag had done nothing.

The proxied dual had drifted further than that one flag. It also lacked the batching/circuit-breaker drop loop, the dry-run "purge ignored" notice, and the `DOLT_UNDROP` recoverability trailer. Its early return on "nothing stale found" was structurally incompatible with purge semantics: `shouldPurgeDroppedDatabases` deliberately ignores the dropped count, because a prior run's dropped-but-unpurged residue is already invisible to `SHOW DATABASES`, so `--purge-dropped` must still fire when this run drops nothing.

## Change

Both topologies now share one implementation keyed on the existing `versioncontrolops.DBConn`, which `*sql.DB` (direct server, via `openDoltServerConnection`) and the pinned non-transactional `*sql.Conn` (proxied server, via `uow.MaintenanceProvider.RunNonTx`) both satisfy. Neither `DROP DATABASE` nor `DOLT_PURGE_DROPPED_DATABASES()` may run inside an explicit transaction, so `RunNonTx` remains the right seam on the proxied side.

- **`cmd/bd/dolt_clean_databases.go`** (new) — `cleanDatabases(ctx, conn, opts)` carries the non-proxied flow verbatim as the canonical behavior. `listStaleDatabases` drains and explicitly closes the result set before any drop; on the proxied path the connection is a single pinned `*sql.Conn`, where the previous code only avoided sharing it with an open result set by accident (`rows.Next()` exhausting auto-closes).
- **`cmd/bd/dolt.go`** — `RunE` reduces to preconditions, options, dispatch. `purgeDroppedDatabases` widens from `*sql.DB` to `DBConn`; body unchanged, existing callers and tests compile untouched. `shouldPurgeDroppedDatabases` is untouched, and both paths now route through that already-unit-tested gate.
- **`cmd/bd/dolt_clean_databases_proxied_server.go`** — 85 lines to a `RunNonTx` delegation. It still builds its own provider, since the global `uowProvider` is nil for `bd dolt` subcommands.

One intentional behavior change on both paths: per-operation timeouts derive from the command's context instead of `context.Background()`, so Ctrl-C interrupts a long clean.

No flag, help text, or output-format changes, so no docs regeneration.

## Tests

New `TestProxiedServerCleanDatabasesPurgeDropped` runs against a dedicated proxied project (`bdProxiedInit`, which spawns its own proxy and child `dolt sql-server`) rather than the shared external harness — the purge is server-global and would destroy sibling tests' undrop state there. For the same reason its subtests are ordered and not parallelized against each other. It uses `DOLT_UNDROP` availability as the observable, mirroring `TestPurgeDroppedDatabasesRemovesUndropRecovery` on the direct-server side:

- `dry_run_ignores_purge` — reports the flag ignored, purges nothing, drops nothing.
- `without_flag_keeps_undrop_recovery` — prints the recoverability trailer, `DOLT_UNDROP` still succeeds.
- `purge_removes_undrop_recovery` — `DOLT_UNDROP` now fails, including for residue from the previous subtest, proving the purge is server-global.
- `zero_stale_still_purges_residue` — the regression this PR exists for: `--purge-dropped` fires on a run that finds nothing stale.

## Verification

```
go build ./... ; go vet ./cmd/bd/ ; golangci-lint run ./cmd/bd/...          → 0 issues
BEADS_TEST_PROXIED_SERVER=1 go test -run TestProxiedServerCleanDatabasesPurgeDropped  → PASS (4/4)
BEADS_TEST_PROXIED_SERVER=1 go test -run 'TestProxiedServerCleanDatabases$'           → PASS (3/3, unchanged)
go test -run 'TestShouldPurgeDroppedDatabases|TestPurgeDroppedDatabases|TestCleanDatabasesPurgeDroppedReclaimsResidue' → PASS
go test -run 'TestEmbeddedDolt$|TestEmbeddedAdminBlocked|TestDoltAdministrativeCommandsRejectRemovedBackends'          → PASS
```

The new test is discovered automatically by `.github/scripts/proxied-test-shard.sh` (it matches `^TestProxiedServer` and hash-falls-back to a shard, since it is not in the committed manifest).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Proxied-server CI shard 1 flake: `TestProxiedServerCleanDatabases` ran a
+  server-global destructive command against the shared test container**
+  (p1-9lf, hazard tracked in p1-8dz). The test used the shared external Dolt
+  container under `t.Parallel()`, and `bd dolt clean-databases` drops every
+  database matching `staleDatabasePrefixes` on the server it is pointed at.
+  The container is bootstrapped with a database named `beads_test`
+  (`internal/testutil/container_provider.go`), which matches one of those
+  prefixes — so every run of this test dropped the shared container's own
+  bootstrap database out from under whatever sibling tests were mid-flight,
+  which is timing-correlated with the `busy buffer` / "pending schema
+  migrations alter pre-existing dirty tables" failures seen in
+  `TestProxiedServerDelete` and `TestProxiedServerFindDuplicates`. The test now
+  runs against a dedicated proxied server (`bdProxiedInit`), like the new purge
+  test, so its blast radius is its own server.
+
 - **`bd dolt clean-databases --purge-dropped` now works in proxied-server
   mode** (p1-9lf). The command read the flag and then dropped it on the floor
   before dispatching to the proxied-server implementation, so under
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`bd dolt clean-databases --purge-dropped` now works in proxied-server
+  mode** (p1-9lf). The command read the flag and then dropped it on the floor
+  before dispatching to the proxied-server implementation, so under
+  `--proxied-server` stale databases were dropped but never purged: their
+  directories stayed under `.dolt_dropped_databases/`, disk was never
+  reclaimed, and nothing reported that the flag had done nothing. The proxied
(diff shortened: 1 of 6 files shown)


coffeegoddd:
aims to fix https://github.com/gastownhall/beads/issues/5012

PR #5024, opened 2026-07-24
fix(dbproxy): race-free proxy port allocation and managed loopback listener policy

maphew:
Stacks on #5013 (the verified-identity PR); only the final commit is new here. Review after
that one lands — happy to rebase.

## Why

Two remaining managed proxied-server safety gaps after #5013: the proxy data port is
allocated with a bind-close-launch race (pick a free port, release it, fork, child binds it
later — another process can take it in between), and a caller-supplied Dolt YAML can point
the managed backend at a non-loopback listener even though managed mode hardcodes a
root/empty-password account on the assumption of loopback-only exposure.

## What

- **Port 0 end to end**: when no explicit port is configured, the child binds
  `127.0.0.1:0` itself and publishes the OS-assigned port through its verified pidfile
  (the identity handshake from #5013 makes that publication trustworthy). The racy
  pre-pick is gone; explicitly configured ports keep exact fixed-port failure semantics.
  `PickFreePort` survives only for the dolt-backend config path, annotated for a follow-up
  RFC (backend port-collision recovery needs a config-ownership contract first).
- **Loopback listener policy**: one shared validator enforces numeric loopback literals for
  the effective listener host of every beads-managed dolt config — generated,
  `BEADS_PROXIED_SERVER_CONFIG`, sidecar, pre-marker default-path, hand-edited marker, and
  interpolation-fallback — at both `bd init` and runtime config resolution. Hostnames
  (including `localhost`) are rejected: resolver output is not loopback proof. An omitted
  host is accepted because dolt `servercfg` defaults it (`DefaultHost = "localhost"` inside
  dolt itself) — flagging that asymmetry openly; tightening it would break minimal custom
  configs that never set a host. There is deliberately no bypass env. External proxied
  topology (`--proxied-server-external-host/port/socket`) is exempt by construction and is
  the supported way to run a non-local backend.

## Testing

Unit: validator table (IPv4/IPv6 loopback pass; wildcard, public, hostname, `localhost`,
empty-string fail; omitted passes), port-0 publishes a nonzero dialable port with a real
proxied SQL round-trip, explicit-port-in-use fails without silent re-pick. Config-path
matrix covered at init and runtime. Managed-local lane (`BEADS_TEST_PROXIED_LOCAL=1`, real
dolt, Linux) green end to end including the new listener-policy lifecycle test and all
#5013 lane tests.

https://claude.ai/code/session_01WZHiYc3wjcdy1wrUP2JiQi


maphew:
Pushed d83434010 (rebased onto the updated #5013 head a9cfebc25 — this PR is stacked on #5013 and should merge after it), closing the findings from a two-vendor adversarial review.

The review confirmed the core claims of the original head: the port-0 bind-and-hold handoff genuinely closes the allocation race (the listener is held from bind through pidfile publish, no close-and-rebind window), and #4313's `dolt-server.port` clobbering concern is not reintroduced. The fixes address what surrounded that core:

**Whole-config listener policy** — the managed-mode check validated only `listener.host`, but dolt also honors `remotesapi.port` and `cluster.remotesapi.port` (both bind all interfaces) and `listener.socket` — so a config with a loopback host plus `remotesapi: {port: 8080}` passed policy while exposing the root/no-password DB to the network. The renamed `validateManagedServerConfigPolicy` now rejects all of these.

**Host policy made coherent** — the old code accepted an omitted host but rejected explicit `localhost`, and the accepted form was strictly worse: dolt's `CheckForUnixSocket` opens `/tmp/mysql.sock` whenever the effective host is `localhost` (the omitted-host default). Managed mode now requires an explicit numeric loopback IP (`127.0.0.0/8`, `::1`, `::ffff:127.0.0.1`); omitted, `localhost`, hostnames, and wildcards are all rejected with remediation text.

**One interpolation-aware validation** — five scattered validator calls collapsed into a single check on the resolved config path, which also fixes the contract violation where a marker config with `host: "${BEADS_HOST}"` hard-failed even when the variable expands to 127.0.0.1.

**Startup fenced by the stop epoch** — a `bd dolt stop` racing a slow managed startup could previously see the stop return, then have the child publish a running proxy afterward. The child now re-checks the stop epoch immediately before publishing its pidfile and aborts (stopping the backend) if it advanced; covered by a deterministic test.

Plus diagnosis fixes: lock-race child exits are no longer misreported as port-bind failures (and real failures name the log path), timeout messages no longer print "port 0", and the init-time config gate now sees `--proxied-server-root-path`. Policy table test covers 17 accept/reject cases; the `BEADS_TEST_PROXIED_LOCAL=1` gated lane passes end to end. Known limitation retained deliberately: a persisted ephemeral-range port that collides after reboot has no auto-repick — that allocator/ownership contract is tracked separately (PR-C/RFC).

Issue #2476, opened 2026-03-09
Reparenting a task does not update source epic's auto-close molecule calculation


MelsovCOZY:
## Description

When a task is reparented from Epic A to Epic B using `bd update <task-id> --parent=<epic-B-id>`, Epic A's internal auto-close calculation still includes the reparented task. Even after all remaining children of Epic A are closed, auto-close never fires.

## Root Cause

`loadDescendants()` in `cmd/bd/template.go` uses two strategies to find children:
1. **Strategy 1:** Check dependency records for parent-child relationships (correctly excludes reparented tasks)
2. **Strategy 2:** Find hierarchical children by ID pattern (`parentID.N`) — this still matches reparented tasks because the ID hasn't changed

When `getMoleculeProgress()` calculates completion via `loadTemplateSubgraph()`, Strategy 2 re-includes the reparented task, making the total count higher than the number of actual children. Auto-close never fires because `progress.Completed < progress.Total`.

## Steps to Reproduce

```bash
bd init
bd create --title="Epic A" --type=epic --priority=1
bd create --title="Task 1" --type=task --parent=bd-1
bd create --title="Task 2" --type=task --parent=bd-1
bd create --title="Task 3" --type=task --parent=bd-1
bd create --title="Epic B" --type=epic --priority=1

bd update bd-1.3 --parent=bd-5

bd close bd-1.1
bd close bd-1.2
# Expected: Epic A auto-closes
# Actual: Epic A remains open
```

## Expected Behavior

After reparenting Task 3 away from Epic A, closing Tasks 1 and 2 (the only remaining children) should trigger auto-close.

## Proposed Fix

In `findHierarchicalChildren()` (or Strategy 2's loop in `loadDescendants()`), verify that each hierarchical child doesn't have an explicit parent-child dependency pointing to a different parent. If it does, skip it — it was reparented away.

## Version

bd 0.58.0

PR #5027, opened 2026-07-24
feat(mol): support bd mol commands in proxied-server mode

coffeegoddd:
## Summary
- Routes 15 DB-touching `bd mol` subcommands (show, current, progress, last-activity, stale, ready --gated, distill, pour, wisp create/list/gc, bond, squash, burn) through proxied-server mode via a new `molReader`/`molWriter` port (`cmd/bd/mol_port.go`) shared with the embedded/server backend, instead of duplicating molecule logic. (`mol seed` is storage-free and needed no proxied path.)
- Dedupes drifted proxied-only reimplementations in `close_proxied_server.go` and `ready_proxied_server.go` that had fallen out of sync with the embedded logic (missing hierarchical-child discovery, reparent checks).
- Wires `FindWispDependentsRecursive` into `domain.IssueUseCase`/`IssueSQLRepository` so `bd mol wisp gc`'s cascade works under proxied mode.
- Adds 11 new proxied-server integration test files (47 subtests) covering every mol/wisp/pour subcommand, following the existing `t.Parallel()` + `newSharedProxiedProject` convention.
- Fixes two real bugs in the new mol port surfaced by running those tests against a live dolt sql-server container: missing label hydration on `GetIssue`/`GetIssuesByIDs`, and dependency-table mis-routing that caused an FK violation when linking a persistent digest to an ephemeral wisp root during squash. Also fixes a nil-pointer panic in `wisp gc --closed`.
- Regenerates the committed proxied-server CI shard manifest so the 14 new `TestProxiedServer*` entry points get pinned, cost-balanced shard assignments.
- Along the way, confirmed (via direct embedded-CLI testing) two pre-existing bugs shared identically by both backends — `bd mol stale --all` can never surface childless epics, and `bd mol bond` can never produce a `compound_proto` result via any real CLI invocation — and added tests that document these as parity guards rather than silently skipping them.

**Behavior change:** `bd ready --gated` (proxied mode) now shares `findGateReadyMolecules` with `bd mol ready --gated` instead of an inline reimplementation, so it now excludes hooked molecules the same way the embedded backend already does. Existing proxied-mode users may see fewer molecules in `bd ready --gated` output than before — this is a correction toward parity, not a regression.

## Test plan
- [x] `go build ./...`
- [x] `golangci-lint run ./...` — 0 issues
- [x] Embedded regression: `go test ./cmd/bd/... -run 'Mol|Wisp|Template|Pour|Bond|Squash|Burn|Distill|Ready' -count=1` — passing
- [x] `internal/storage/...` full suite — passing
- [x] `BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd/ -run 'TestProxiedServerMol|TestProxiedServerWisp|TestProxiedServerPour'` against a real dolt sql-server container — 47/47 subtests passing
- [x] Regenerated shard manifest validated via `proxied-test-shard.sh <N> 15` with `BEADS_TEST_SHARD_LIST_ONLY=1` across all 15 shards — zero fallback, no errors


maphew (commented):
Ran a cross-vendor review pair on this diff (Codex gpt-5.6-sol high + Claude opus, findings independently re-verified against the head `1a275d3` before posting). Overall: the port architecture is right — method names matching `storage.DoltStorage` so embedded satisfies `molReader`/`molWriter` structurally is elegant, the dedup direction is correct everywhere we traced it (proxied gains embedded behavior: hierarchical-child discovery, #2476 reparent check, wisp claims), the storage boundary is respected, and these are the best proxied integration tests in the repo. The items below are ordered by how much they should gate merge.

## CI red: root-caused, fix verified locally

Shard 11 fails deterministically, not flake. `TestProxiedServerReady/gated_empty_case` (`cmd/bd/ready_proxied_integration_test.go:438`) asserts `"No closed gates found"`, which this PR deletes along with `emitGatedEmpty()`; the shared empty-case wording is now `"No molecules ready for gate-resume dispatch"`. Updating the assertion to the new wording is the right fix (embedded wording is the parity target) — we applied exactly that one-line change locally and the subtest passes against a real dolt sql-server container (2.7s). We also swept the whole tree for other assertions on deleted proxied-only strings: this is the only one. (Cause of the slip: the test plan ran proxied `Mol|Wisp|Pour` but not proxied `Ready`.)

## Blockers (silent data/config loss on the new proxied paths)

1. **`uowMolWriter.CreateIssue` drops labels.** `cmd/bd/mol_port.go:314` builds `domain.CreateIssueParams{Issue: issue, ExplicitID: issue.ID}` without `Labels`; the domain create persists only `params.Labels` (`internal/storage/domain/issue.go:759`), never `issue.Labels`. Embedded persists `issue.Labels` via `issueops` `PersistLabels`. Since `cloneSubgraphInto` sets `Labels: oldIssue.Labels` on every clone and cook synthesizes `gate:<x>` labels, proxied `pour`/`wisp create`/`bond` produce molecules whose steps have no labels at all. Fix is one field (`Labels: issue.Labels`) plus a test asserting a cooked formula step's labels survive a proxied pour.

2. **Custom types never registered on proxied clones.** `ensureSubgraphCustomTypes` stayed in the embedded `cloneSubgraph` wrapper (`cmd/bd/template.go:562`); the proxied callers (`mol_proxied_server.go:111`, `wisp_proxied_server.go:75`) invoke `cloneSubgraphInto` directly, so a formula using e.g. type `gate` instantiates without updating `types.custom`. The uow create path doesn't validate type on insert, so it fails silently — until a later `bd update --type gate` is rejected. Moving the call inside `cloneSubgraphInto` compiles as-is (`molWriter` is a superset of `molConfigWriter`). Note `bondProtoMolAttachInto` lacks registration in both modes today, so that leg is a pre-existing shared gap rather than a regression.

3. **Cascade semantics diverge through the port on delete.** `uowMolWriter.DeleteIssue` (`mol_port.go:359`) routes to domain `DeleteIssue`/`DeleteWisp`, which hard-code `Cascade: true` (`internal/storage/domain/issue_delete.go:18,29`). The embedded `tx.DeleteIssue` it abstracts is strictly single-row (`internal/storage/issueops/delete.go:21`). So proxied `mol squash` deleting an ephemeral child can cascade into dependents outside the molecule that embedded would leave untouched. The port method should be non-cascading, with burn/GC opting into cascade explicitly.

4. **Proxied `mol burn` commits partial deletions and double-counts on retry.** `mol_proxied_server.go:600-609` logs a `burnWispsInto` error to stderr and lets `uow.RunTx` commit what already succeeded — contradicting the documented all-or-nothing contract at `mol_burn.go:375` that embedded has a test for (`critical_paths_test.go:110`). Also `batchResult` is declared outside the closure while `RunTx` wraps it in `backoff.Retry`, so a serialization retry appends duplicate results and double-counts `TotalDeleted`. Related: for a wisp root with children, cascade deletes several rows but the result reports only the operands, so `total_deleted` underreports even on success. The persistent-molecule loop just below returns the error correctly — same treatment here.

## Parity divergences worth a decision (the PR's own thesis is parity)

5. **`resolveMolID` (`mol_port.go:382`) reimplements `utils.ResolvePartialID` minus several behaviors** — known-prefix handling for multi-hyphen prefixes, the partial-ID-shape guard before fuzzy search, sorted/deterministic ambiguity output, and crucially the narrow `SearchIssueIDs` projection whose absence `id_parser.go:113` documents as a 60s+ pathology on a 23k-issue db. And it's now used on *embedded* paths too (`resolveOrDescribe`/`resolveOrCookToSubgraph` from `runMolBond`, `resolveProtoIDOrTitle`). Suggestion that deletes code instead of adding: add `SearchIssueIDs` to `molReader`, narrow `ResolvePartialID`'s parameter to that interface, drop `resolveMolID` entirely.

6. **`GetMoleculeProgress` port diverges from embedded SQL** (`mol_port.go:240-256` vs `issueops/molecule.go:94-109`): `InProgress` hard-coded to 1 vs a real count; subgraph-recursive totals vs direct-children totals (different `percent` for nested molecules); first vs last in-progress step as `CurrentStepID`. It also defeats the `LargeMoleculeThreshold` fast path by hydrating the whole subgraph over the wire per node. The proxied progress tests only cover flat single-in-progress molecules, so none of this trips.

7. **`GetMoleculeLastActivity`** (`mol_port.go:258`) seeds from root `UpdatedAt`/`molecule_updated`; embedded (`issueops/compaction.go:281`) ignores the root when children exist. A molecule whose epic was touched last answers differently per backend. The new `child_update_wins_over_root` test backdates the root so it can't catch this — the inverse case would.

8. **`bd ready --gated` vs `bd mol ready --gated` now render differently in proxied mode**: `runReadyProxiedGated` (`ready_proxied_server.go:421`) keeps an inline renderer (and still swallows the `outputJSON` error that 1a275d3 stopped swallowing elsewhere) instead of calling the extracted `renderGatedReadyMolecules` that `runMolReadyGatedProxiedServer` uses. One-line fix.

9. **`bd mol wisp gc --json` ships two shapes**: embedded emits the `deleteBatch` shape, proxied emits `{"cleaned_ids","cleaned_count"}` (`wisp_proxied_server.go:178`), and the new test pins the divergent one.

## Smaller notes

- `isWisp` (`mol_port.go:303`) never caches negatives (extra `GetWisp` round-trip per write on persistent issues) and treats any error as "not a wisp", so a transient backend error silently mis-routes a write to the persistent tables. Worth failing closed on error.
- `AdvanceToNextStep`'s `s == nil` guard is dead now that `close.go:282` passes a struct-boxed `newStandaloneStoreMolWriter`; a nil store would panic past the guard. Also worth a doc comment on `storeMolWriter` that reads bypass `tx` (no read-your-own-writes) and `ClaimStepIfOpen` opens its own transaction — fine for current callers, a trap for the next one.
- The two parity-guard tests (childless-epic stale, bond `compound_proto`) fail loudly on fix as intended — but their names read as spec. A tracking issue reference in each comment would keep a future reader from calcifying them.
- Behavior change worth a release-note line: old proxied `ready --gated` counted hooked molecules as dispatchable; shared `findGateReadyMolecules` excludes them. Correct direction, pinned by `hooked_molecule_excluded`, but output shrinks for existing proxied users.
- `mol seed` never touched the DB (filesystem-only), so "16 subcommands" slightly overcounts.
- Shard manifest reshuffle is verified correct (96 test functions = 96 entries, no dupes) but will conflict with any concurrent PR touching it; needs regeneration after rebase.

Happy to push the shard-11 test fix (and any of 1–4) as a follow-up commit if useful — no file overlap with #5013/#5024, so merge order is free.

PR #5028, opened 2026-07-25
fix(beads): allow /var/tmp in repo-context safety boundary

quad341:
## Problem

`isPathInSafeBoundary()` rejected any beads directory located under `/var/tmp`. On systems where Go's build tooling sets `GOTMPDIR=/var/tmp/...` for build-like subcommands, `t.TempDir()` roots test temp dirs under `GOTMPDIR` even though `os.TempDir()` itself still reports `/tmp`. That left the existing `os.TempDir()`-based carve-out unable to match, so any repo context resolved under a fresh `t.TempDir()` on such a system tripped the "unsafe location" check.

## Fix

Add a `/var/tmp` carve-out using the same symlink-safe resolution already used for the existing `/Users/Shared` carve-out, since `/var/tmp` is equally world-writable (`drwxrwxrwt`). This is a general FHS-standard fix (`/var/tmp` is the standard secondary temp directory that persists across reboots), not specific to any one environment.

## Testing

- Two table-test cases cover the `/var/tmp` root and a `GOTMPDIR`-style subdirectory.
- `TestIsPathInSafeBoundary_VarTmpSymlinkEscape` proves the carve-out rejects a symlink planted under `/var/tmp` that escapes to a disallowed directory. It skips gracefully when `/var/tmp` is absent or unwritable.
- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/beads -count=1`

## Current branch state

#5047 landed the separate macOS physical-temp-root fix on `main`. Merge commit `98051992f` takes that production form and removes the duplicate maintainer helper implementation from this branch. The effective diff against `main` is therefore back to this PR's original approved `/var/tmp` scope, with the contributor commit and tests intact.

### DIFF
--- a/internal/beads/context.go
+++ b/internal/beads/context.go
@@ -315,8 +315,9 @@ func isPathInSafeBoundary(path string) bool {
 	// escapes the boundary must be rejected, not followed into a system
 	// directory — same treatment as the /Users/Shared carve-out below
 	// (be-kghzr SEC-003 hardening).
-	tempDir := strings.TrimSuffix(os.TempDir(), "/")
-	if absPath == tempDir || strings.HasPrefix(absPath, tempDir+"/") {
+	tempDir := filepath.Clean(os.TempDir())
+	physicalTempDir := resolveLongestExistingAncestor(tempDir)
+	if pathWithinTempRoots(absPath, tempDir, physicalTempDir) {
 		return resolvedPathWithinRoot(absPath, tempDir)
 	}
 
@@ -405,6 +406,24 @@ func resolveLongestExistingAncestor(path string) string {
 		remainder = filepath.Join(filepath.Base(cur), remainder)
 		cur = parent
 	}
+}
+
+// pathWithinRoot performs a syntactic, path-boundary-aware containment check.
+// Both paths are cleaned first so sibling prefixes and ".." components cannot
+// make an unrelated path appear to be within root.
+func pathWithinRoot(path, root string) bool {
+	path = filepath.Clean(path)
+	root = filepath.Clean(root)
+	return path == root || strings.HasPrefix(path, root+string(filepath.Separator))
+}
+
+// pathWithinTempRoots gates the OS temp-directory carve-out for both forms seen
+// on macOS: the lexical /var/folders form returned by os.TempDir and the
+// canonical /private/var/folders form produced by symlink resolution. This is
+// only the syntactic gate; resolvedPathWithinRoot makes the final, symlink-safe
+// containment decision.
+func pathWithinTempRoots(path, lexicalRoot, physicalRoot string) bool {
+	return pathWithinRoot(path, lexicalRoot) || pathWithinRoot(path, physicalRoot)
 }
 
 // resolvedPathWithinRoot reports whether absPath, after symlink resolution, still
--- a/internal/beads/context.go
+++ b/internal/beads/context.go
@@ -327,6 +327,18 @@ func isPathInSafeBoundary(path string) bool {
 		return true
 	}
 
+	// Allow /var/tmp as the FHS-standard secondary temp directory (persists across
+	// reboots, unlike /tmp). This is distinct from the os.TempDir() carve-out
+	// above: a machine's build tooling can set GOTMPDIR to redirect Go's own
+	// test/compile temp dirs under /var/tmp even while os.TempDir() itself still
+	// reports /tmp, so t.TempDir() in a test binary can land here without the
+	// os.TempDir() check ever seeing it (be-odye4). Like /Users/Shared, /var/tmp is
+	// world-writable (drwxrwxrwt), so resolve symlinks before admitting (SEC-003):
+	// a symlink planted under it must not be followed into a rejected directory.
+	if absPath == "/var/tmp" || strings.HasPrefix(absPath, "/var/tmp/") {
+		return resolvedPathWithinRoot(absPath, "/var/tmp")
+	}
+
 	for _, prefix := range unsafePrefixes {
 		if strings.HasPrefix(absPath, prefix+"/") || absPath == prefix {
 			return false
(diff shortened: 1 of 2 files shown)


maphew (approved):
Approving. Reviewed the diff against `isPathInSafeBoundary` on current `main`, plus the `resolvedPathWithinRoot` / `resolveLongestExistingAncestor` helpers it reuses.

**Why this is correct:**

- Placement is right: the `/var/tmp` carve-out sits after the `os.TempDir()` and `/var/home/` allowances and *before* the `unsafePrefixes` loop, which is what lets it escape the blanket `/var` rejection — exactly the same structural position the problem requires.
- It reuses `resolvedPathWithinRoot` rather than a raw prefix check, so a symlink planted under world-writable `/var/tmp` that escapes to a rejected directory resolves outside the root and is refused. Since the carve-out returns the helper's result unconditionally, a failed resolution rejects rather than falling through — conservative in the right direction for a SEC-003 check.
- The boundary comparison (`== "/var/tmp" || HasPrefix(absPath, "/var/tmp/")`) avoids the sibling-prefix trap (`/var/tmpfoo` does not match).
- Tests mirror the established `/Users/Shared` symlink-escape pattern, and the escape test skips gracefully when `/var/tmp` is absent or unwritable, so it cannot flake on constrained runners.

CI is fully green, merge state CLEAN, and `pr-preflight` (blocking base-health mode) passes: base `main` green across all 7 workflows.

**One coordination note for whoever merges:** #5030 touches the same function. I ran a local `git merge-tree` simulation — `internal/beads/context.go` merges cleanly in either order, but `internal/beads/context_test.go` conflicts (both PRs append test cases at the same two insertion points). The resolution is trivial (keep both added blocks); whichever PR lands second will need that one-line-order conflict resolved, either by branch update or a maintainer push (maintainer edits are enabled on both). No semantic interaction between the two carve-outs.

maphew:
Jim, thank you for the `/var/tmp` fix and its symlink-escape coverage. I preserved your original commit and updated this branch in place rather than opening a competing PR.

#5030 landed first and then exposed the physical `/private/var/folders/...` form of macOS's temp root. #5044 intended to repair that regression, but its production hunk was lost during merge and `main` remained red. I merged current `main` into this branch, kept all three regression blocks, and added the missing production gate as a separate maintainer commit.

The combined change still makes `resolvedPathWithinRoot` the final containment decision, so the new physical spelling does not weaken your or #5030's symlink-escape protection. The full `internal/beads` package and targeted boundary/context tests pass locally. I also updated the PR title and body so the expanded stop-the-line scope is explicit.

_codex-gpt-5.6-sol-xhigh on behalf of maphew_

quad341:
Thanks @maphew — merged current `main` in and resolved the conflict. The branch is mergeable again.

**What conflicted.** #5047 landed the physical temp-root carve-out directly on `main` (inline `physTempDir` form) about half an hour after this branch got the same fix in a different shape (`pathWithinRoot` / `pathWithinTempRoots` helpers). Same behaviour, two spellings of the same gate — that's the whole conflict.

**How I resolved it** (98051992f, a plain merge commit — nothing force-pushed, your two commits are intact in history):

- Took `main`'s landed `physTempDir` form. Re-landing a refactor of already-merged code inside a `/var/tmp` PR is scope creep, and `main` is the one that has to stay green.
- Dropped the now-unreferenced `pathWithinRoot` / `pathWithinTempRoots` helpers and `TestPathWithinTempRoots_PhysicalForm` along with them, rather than leave production helpers reachable only from a test.
- `resolvedPathWithinRoot` is still the final containment decision, so #5030's symlink-escape rejection is untouched.

That leaves this PR at exactly its original approved scope — the diff against `main` is now only the `/var/tmp` carve-out (be-odye4) plus its boundary cases and `TestIsPathInSafeBoundary_VarTmpSymlinkEscape`. That part is still not on `main` by any other route.

If you'd rather keep the boundary-aware helper shape, I'd suggest it as its own follow-up against `main` — say the word and I'll restore it here instead.

**On the red checks** — none of the three were caused by this change:

| Check | Cause |
|---|---|
| `Test (Proxied Dolt Cmd 13/15)` | `docker pull dolthub/dolt-sql-server:2.2.0` → `registry-1.docker.io` client timeout. Infra flake. |
| `Storage backend conformance (embedded Dolt oracle)` | `panic: test timed out after 10m0s`. At the alarm the suite was mid-`dependencies-readiness/CrossTypeEpicTaskBlocking (1s)` — progressing, not wedged. No `internal/beads` frames in the dump; the suite just doesn't fit the 10m package timeout. |
| `CI Gate / Required` | Aggregate — failed only on `TEST_PROXIED_CMD=failure`, i.e. the row above. |

**Verification:** `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/beads -count=1 -run 'TestIsPathInSafeBoundary|TestGetRepoContextForWorkspace|TestRepoContext_Validate'` all green, including `TestIsPathInSafeBoundary_TempDirPhysicalForm` and both symlink-escape tests. `gofmt` clean. Two unrelated failures in the full package locally (`TestFindAllDatabases_Unit`, `TestGitCmd_WorktreeContext`) reproduce identically on unmodified `af03c6ee6`, so they're my sandbox, not this branch. macOS CI remains the platform confirmation.

maphew:
Thanks, Jim. I checked `98051992f` against current `main`, and this is the right resolution.

- Taking #5047's already-landed `physTempDir` form avoids reintroducing an equivalent refactor in the `/var/tmp` PR.
- The effective diff is back to exactly the original 52-line `/var/tmp` change and its symlink-escape coverage.
- Your original commit and the intervening maintainer commits remain intact in history.
- The full `internal/beads` package passes locally at the new head.

I restored the PR title/body to that effective scope and am refreshing the exact-head full-suite verification. We will treat the current CI failures according to the evidence you recorded rather than changing this patch for unrelated infrastructure/time-budget failures.

_codex-gpt-5.6-sol-xhigh on behalf of maphew_

PR #5029, opened 2026-07-25
fix(bench): stop dolt benchmarks from leaking test databases onto shared servers

quad341:
## Summary

The Dolt storage benchmarks could resolve their server connection from whatever Dolt server port was already configured in the environment, rather than always using an isolated throwaway instance. In shared or long-running environments where a real Dolt server is configured via environment variables, running the benchmarks could create throwaway test databases directly on that shared server instead of a disposable one — and never cleaned them up afterward.

## Changes

- Benchmarks now require an explicit opt-in (`BEADS_BENCH_DOLT_PORT`) pointing at a dedicated throwaway Dolt server before they will run against one at all. Without it, they skip cleanly with no network connection attempted.
- Ambient server-port configuration is no longer inherited by the benchmark setup path, closing the gap that allowed a shared server to be targeted unintentionally.
- Benchmark cleanup now explicitly drops the database it created, so a benchmark run (with the opt-in) no longer leaves anything behind.
- Added regression tests covering the opt-in decision logic (missing, valid, invalid, and zero/negative port values) and the default skip behavior.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] Full `internal/storage/dolt` package suite green (173/173)
- [x] New regression tests verified against a live environment with a shared server port configured — benchmarks skip cleanly, zero network I/O
- [x] Opt-in path verified end-to-end against a real disposable Dolt server

### DIFF
--- a/internal/storage/dolt/dolt_benchmark_safety_test.go
+++ b/internal/storage/dolt/dolt_benchmark_safety_test.go
@@ -1,0 +1,61 @@
+// Regression tests for be-cfm3z: the Dolt benchmark suite used to resolve
+// its server port from ambient BEADS_DOLT_SERVER_PORT/BEADS_DOLT_PORT, which
+// gc agent shells set to point at a shared production Dolt server. These
+// tests exercise the pure decision helper (benchDoltServerPort) directly —
+// no server, no network — so they can never themselves reach a real Dolt
+// server, production or otherwise, regardless of ambient env state.
+package dolt
+
+import "testing"
+
+func TestBenchDoltServerPort_SkipsWithoutOptIn(t *testing.T) {
+	t.Setenv("BEADS_BENCH_DOLT_PORT", "")
+	// Simulate the stock gc agent shell, which exports this pointing at the
+	// shared production server.
+	t.Setenv("BEADS_DOLT_SERVER_PORT", "28231")
+	t.Setenv("BEADS_DOLT_PORT", "")
+
+	port, skip, reason := benchDoltServerPort()
+	if !skip {
+		t.Fatalf("expected skip=true when BEADS_BENCH_DOLT_PORT is unset, got skip=false (port=%d)", port)
+	}
+	if reason == "" {
+		t.Fatal("expected a non-empty skip reason")
+	}
+}
+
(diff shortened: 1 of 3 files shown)


maphew (approved):
Approving. Reviewed the diff against current `main`, including how `applyConfigDefaults`/`New` resolve the server port.

**Why this is correct:**

- The leak vector is real: `applyConfigDefaults` reads `BEADS_DOLT_SERVER_PORT` (and legacy `BEADS_DOLT_PORT`) and overrides the benchmark's config, so any environment with those set — e.g. an agent shell pointed at a shared server — silently routed `beads_test_bench_*` databases onto that server, and the old cleanup never dropped them.
- I checked the "does this regress a working default?" question, and it doesn't: current `New()` always goes through `newServerMode` (there is no embedded open path today), so under a clean env the old benchmarks resolved port 0, failed the dial, and died with `b.Fatalf`. Skip-with-instructions under the default env is strictly better than fatal, and the explicit `BEADS_BENCH_DOLT_PORT` opt-in is the only way these benchmarks can run at all under the current architecture.
- Defense stacking is sensible: env scrub via `b.Setenv` (empty string behaves as unset in the resolution chain), `BEADS_TEST_MODE=1` keeps the production-port guard armed even for a mistyped opt-in port, and `BEADS_TEST_SERVER=1` is only reached after the operator explicitly designated a throwaway server.
- `dropBenchDatabase` opening its own short-lived admin connection (instead of the store's pool) means cleanup works regardless of store close state, and ignoring errors is right for best-effort cleanup.
- The four pure-logic regression tests are network-free by construction, so they can never themselves touch a real server. Good pattern.
- `release-gates/` gate record follows the established convention in that directory (cf. the AD-01 and be-x42v gates), and this one usefully documents the `scripts/test.sh` hermetic-env methodology point.

CI fully green, merge state CLEAN, `pr-preflight` (blocking base-health mode) passes: base `main` green across all 7 workflows. No file overlap with your #5028/#5030, so this one can land independently in any order.

One non-blocking observation, no action needed: `BenchmarkBootstrapEmbedded` now measures server-mode bootstrap (as does everything through `New()`), so the name is historical twice over. Worth a rename whenever someone next touches that file — not worth churn here.

PR #5030, opened 2026-07-25
fix(beads): make the temp-directory safety-boundary check symlink-safe

quad341:
## What changed

Hardens the OS temp-directory allowance in the repo-context safety-boundary check (`isPathInSafeBoundary`) so it can no longer be fooled by a symlink.

## Why

The temp directory (`/tmp` and equivalents) is world-writable, so any co-located local user can plant a symlink there. The old check tried to resolve symlinks first, but `filepath.EvalSymlinks` fails outright whenever the last path component doesn't exist yet — which is the normal case for a directory that's about to be created. When resolution failed, the code fell back to a raw prefix check on the *unresolved* path, with no symlink protection at all. A symlink planted under the temp dir pointing somewhere sensitive would sail through that fallback.

This is the same class of gap already closed for the `/Users/Shared` and `/var/tmp` carve-outs elsewhere in this function — the temp-dir check was the one remaining spot still using the weaker pattern.

## What to look at

The two-branch check collapses into a single call to the existing `resolvedPathWithinRoot` helper, which walks up to the longest existing ancestor, resolves symlinks there, and rejoins the non-existent remainder. That's safe for both symlink escapes and the common case of a not-yet-created path.

One thing worth double-checking in review: this returns the helper's result unconditionally once a path syntactically looks like it's under the temp dir, rather than falling through to later checks on a mismatch. That's deliberate — unlike `/Users/Shared` and `/var/tmp`, nothing later in the function covers the temp dir, so falling through on a failed check would silently readmit the escape via the function's fail-open default.

## How we know it works

Added a test that plants a real symlink under the temp dir pointing at `/etc` and confirms both the symlink itself and a not-yet-existent path routed through it are rejected. Also added a table case confirming a legitimate not-yet-created subdirectory under the temp dir still validates correctly. Full existing suite still passes; the only failures present are a pre-existing, unrelated sandbox/environment condition (confirmed identical on `main` before this change).

## Heads up for reviewers

PR #5028 touches this same function (adjacent, non-overlapping carve-out blocks) — no semantic conflict between the two, but landing order may be worth coordinating to avoid a needless textual merge conflict.

### DIFF
--- a/internal/beads/context.go
+++ b/internal/beads/context.go
@@ -309,17 +309,15 @@ func isPathInSafeBoundary(path string) bool {
 		return false
 	}
 
-	// Allow OS-designated temp directories (e.g., /var/folders on macOS)
-	// On macOS, TempDir() returns paths under /var/folders which symlinks to /private/var/folders
-	tempDir := os.TempDir()
-	resolvedTemp, _ := filepath.EvalSymlinks(tempDir)
-	resolvedPath, _ := filepath.EvalSymlinks(absPath)
-	if resolvedTemp != "" && strings.HasPrefix(resolvedPath, resolvedTemp) {
-		return true
-	}
-	// Also check unresolved paths (in case symlink resolution fails)
-	if strings.HasPrefix(absPath, tempDir) {
-		return true
+	// Allow OS-designated temp directories (e.g., /var/folders on macOS, which
+	// symlinks to /private/var/folders). World-writable, so resolve symlinks
+	// before admitting: a symlink planted under the temp dir whose target
+	// escapes the boundary must be rejected, not followed into a system
+	// directory — same treatment as the /Users/Shared carve-out below
+	// (be-kghzr SEC-003 hardening).
+	tempDir := strings.TrimSuffix(os.TempDir(), "/")
+	if absPath == tempDir || strings.HasPrefix(absPath, tempDir+"/") {
+		return resolvedPathWithinRoot(absPath, tempDir)
 	}
 
 	// Allow /var/home as a valid user home directory (Fedora Silverblue, Bluefin, etc.)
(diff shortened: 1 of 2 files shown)


maphew (approved):
Approving. Reviewed the diff against current `main` and traced the helper chain (`resolvedPathWithinRoot` -> `resolveLongestExistingAncestor`).

**Why this is correct:**

- The bug is real: `filepath.EvalSymlinks` fails on a not-yet-existent trailing component (the normal case for a BEADS_DIR about to be created), and the old fallback then did a raw `strings.HasPrefix` on the *unresolved* path — no symlink protection at all. Your red test captures exactly that shape (`<symlink-to-/etc>/.beads`).
- Collapsing both branches into `resolvedPathWithinRoot(absPath, tempDir)` closes the gap with the same mechanism already vetted for `/Users/Shared`, and it works on macOS: the helper resolves *both* sides through `resolveLongestExistingAncestor`, so `os.TempDir()` living under the symlinked `/var` -> `/private/var` still compares equal. The `TrimSuffix(os.TempDir(), "/")` also quietly fixes two latent defects in the old check — macOS's trailing-slash `TMPDIR` breaking the exact-root match, and the unanchored prefix admitting siblings like `/tmpfoo`.
- Returning the helper's result unconditionally (rather than falling through on mismatch) is the right call, for the reason you documented: nothing later in the function covers the temp dir, so fall-through would readmit the escape via the fail-open default. The cost is that a symlink under `/tmp` pointing at an otherwise-safe location is now rejected — acceptable conservatism for a security boundary; callers can pass the canonical path.
- The not-yet-created-subpath table case is the important regression guard here, and it is present.

CI fully green, merge state CLEAN, `pr-preflight` (blocking base-health mode) passes: base `main` green across all 7 workflows.

**Merge-order note** (same as posted on #5028): local `git merge-tree` simulation shows `context.go` merges cleanly with #5028 in either order, but `context_test.go` has a trivial both-appended-at-the-same-spot conflict. Whichever lands second needs a keep-both resolution; maintainer edits are enabled on both branches.

PR #5032, opened 2026-07-25
test(shared-server): disable metrics flusher to stop t.TempDir cleanup race

### LINKED ISSUE #5031: Flaky test: TestSharedServerCfgNilHonorsSharedServer fails its own cleanup (detached metrics flusher races t.TempDir)
### Symptom

`TestSharedServerCfgNilHonorsSharedServer` (`cmd/bd/shared_server_cfgnil_test.go`) intermittently fails in its **own cleanup**, not on an assertion — seen on the `PR Core (wrapper timing)` job:

```
--- FAIL: TestSharedServerCfgNilHonorsSharedServer (0.10s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat .../002: directory not empty
FAIL  github.com/steveyegge/beads/cmd/bd  244.590s
```

The test body passes (0.10s); the failure is Go's automatic `t.TempDir()` teardown hitting `ENOTEMPTY`.

### Root cause

`bd` spawns a **detached `bd send-metrics` child** (`internal/metrics/spawn.go` `MaybeSpawnFlusher`) that keeps writing `$HOME/.beads/eventsData` **after the parent process exits**. In this test `HOME` is a `t.TempDir()`, so the detached child races Go's `RemoveAll` and the temp dir is non-empty when `unlinkat` runs on it.

Confirmed by process inspection: after `bd list` returns, a `bd send-metrics` child is still alive; with `BD_DISABLE_EVENT_FLUSH=1` set, no child spawns.

This is already a **known hazard** — it's documented verbatim in `cmd/bd/protocol/helpers_test.go` (added in 08881217b, 2026-07-13), which calls it "a test-stability bug" and disables metrics for exactly this reason. The sibling `shared_server_integration_test.go` and `proxied_integration_helpers_test.go` also set the disable flag. `shared_server_cfgnil_test.go` simply omitted it.

### Fix

Add `BD_DISABLE_EVENT_FLUSH=1` to the test's env, matching the sibling shared-server/protocol tests. PR to follow.

### Broader note (optional follow-up)

Relying on every hermetic test to remember `BD_DISABLE_EVENT_FLUSH` / `BD_DISABLE_METRICS` is a footgun — this is at least the second test to hit it. A source-level guard would remove the whole class: e.g. have `MaybeSpawnFlusher` also return early when `BEADS_TEST_MODE=1` (already the canonical "tests own the lifecycle" signal, inherited across exec). No production impact, since `BEADS_TEST_MODE` is never set in real use. Happy to do that as a separate PR if you'd prefer it over per-test flags.

rjc123:
## Problem

`TestSharedServerCfgNilHonorsSharedServer` (`cmd/bd/shared_server_cfgnil_test.go`) intermittently fails in its **own cleanup**, not on an assertion (seen on the `PR Core (wrapper timing)` job):

```
--- FAIL: TestSharedServerCfgNilHonorsSharedServer (0.10s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat .../002: directory not empty
```

## Root cause

`bd` spawns a **detached `bd send-metrics` child** (`internal/metrics/spawn.go` `MaybeSpawnFlusher`) that keeps writing `$HOME/.beads/eventsData` **after the parent exits**. `HOME` in this test is a `t.TempDir()`, so the detached child races Go's automatic `RemoveAll` and the dir is non-empty when `unlinkat` runs.

Confirmed by process inspection: after `bd list` returns, a `bd send-metrics` child is still alive; with `BD_DISABLE_EVENT_FLUSH=1` set, none spawns.

This is already documented as "a test-stability bug" in `cmd/bd/protocol/helpers_test.go` (08881217b), and the sibling `shared_server_integration_test.go` / `proxied_integration_helpers_test.go` set the disable flag for the same reason. This test just omitted it.

## Fix

Add `BD_DISABLE_EVENT_FLUSH=1` to the test env, matching the sibling tests. Verified: the previously-flaky test passes repeatedly (`-count=5`) and no `send-metrics` child spawns.

Fixes #5031

## Note

Several hermetic tests hand-roll this flag to dodge the same race; a source-level guard (e.g. suppress the flusher under the existing `BEADS_TEST_MODE=1` signal) would remove the whole class. Kept out of this PR to stay minimal — happy to do it separately if preferred (noted in #5031).


maphew (approved):
Verdict: **approve**.

Verified the whole chain: the `PR Core` failure mode on #4821 matches your diagnosis byte-for-byte (`TempDir RemoveAll cleanup: unlinkat .../002: directory not empty`); the detached `bd send-metrics` flusher is real (`internal/metrics/spawn.go`); and the sibling tests (`shared_server_integration_test.go`, the protocol helpers per 08881217b) already set `BD_DISABLE_EVENT_FLUSH=1` for exactly this documented reason — this test was simply the one that omitted it. The seven-line env addition with the explanatory comment is the right minimal shape.

Why approve: this kills a teardown race that has been biting *unrelated* PRs (it red-flagged your own #4821), and it does so by matching an established in-tree convention rather than inventing a new mechanism.

On this PR's own red check: the only failure is `Windows Make shell (cygwin)` — verified from the log it is the Cygwin installer download failing all 5 retry attempts, i.e. precisely the infra flake your paired #5035 downgrades to a warning. Neither failure has anything to do with this diff. Merge tail: a rerun (or #5035 landing first) clears it.

Your note about a source-level guard (suppress the flusher under `BEADS_TEST_MODE=1`) is the right follow-up — nine test files now hand-roll this flag; tracked on our side per #5031.

PR #5035, opened 2026-07-25
ci(pr): soft-fail the Cygwin Make-shell setup on installer-download flakes

### LINKED ISSUE #5034: CI: windows-make-shell (cygwin) fails the required check on a Cygwin-installer download flake — should warn, not block
### Symptom

The `windows-make-shell` job's cygwin leg (`.github/workflows/pr.yml`, `name: Windows Make shell (${{ matrix.host }})`) intermittently fails the whole required check during **environment setup**, before any build or test runs. Example (PR #5032, run [30152168275](https://github.com/gastownhall/beads/actions/runs/30152168275/job/89664284348)):

```
Set up Cygwin GNU Make
  Attempt 1 failed. Retrying...
  ...
  Attempt 5 failed. Retrying...
  Failed to download setup-x86_64.exe after 5 attempts.
  ##[error]Process completed with exit code 1.
```

### Why this is wrong

The failing step is `Set up Cygwin GNU Make` (`.github/workflows/pr.yml:563`), which uses `cygwin/cygwin-install-action@…v6` to download `setup-x86_64.exe` from cygwin.com. When that download times out, the job exits 1 and the **required** check goes red — blocking PRs for a reason that has nothing to do with the change under test (here, a one-line test-env edit that doesn't even touch Windows). An external-dependency/network failure in toolchain setup should surface as a **warning**, not a hard red gate.

### Suggested fixes (any one, or combination)

1. **`continue-on-error: true`** on the Cygwin setup step (or the `host == 'cygwin'` matrix leg) so an installer-download failure renders as a neutral/yellow warning instead of failing the required check. The actual Make-shell smoke assertions still fail loudly if they run and break.
2. **Reduce the flakiness at source** — `cygwin/cygwin-install-action` supports a mirror `site:` and package caching; pinning a reliable mirror + caching the installer would cut the download failures.
3. **De-required** the cygwin leg (keep it informational) if a network-dependent third-party installer can't be made reliable enough for a blocking gate.

The intent: a code change should never be blocked by a transient failure to download the Cygwin installer. Happy to open a PR for option 1 (the smallest change) if that's the preferred direction.

rjc123:
Closes #5034.

## Problem

The `windows-make-shell` cygwin leg installs its toolchain with `cygwin/cygwin-install-action`, which downloads `setup-x86_64.exe` from cygwin.com. When that download times out, the `Set up Cygwin GNU Make` step exits 1, the job goes red, and — because `windows-make-shell` is in the `ci-gate` required set — it **blocks the PR** for a purely infrastructural reason, unrelated to the change under test. (Observed on #5032, whose only change was a one-line test-env edit that doesn't touch Windows.)

## Fix

Downgrade an installer-download failure to a non-blocking warning, without losing signal on real regressions:

- **`continue-on-error: true`** on the Cygwin setup step, so a failed install no longer fails the job.
- Gate the Cygwin smoke step on **`steps.cygwin.outcome == 'success'`** — `outcome`, not `conclusion`, because `continue-on-error` rewrites `conclusion` to `success` while `outcome` preserves the real result. So the smoke step skips cleanly on a failed install instead of erroring on a missing `bash`.
- Emit a **`::warning::`** annotation when the leg is skipped, so the flake stays visible rather than silently green.

A genuine Cygwin Make regression (install succeeds, smoke fails) still fails the job exactly as before — only the flaky third-party installer download is downgraded to a warning.

## Verification

YAML validated locally (parses cleanly); the behaviour relies on standard Actions `outcome`/`continue-on-error` semantics. I can't exercise the Windows/Cygwin runner locally, so the real proof is this PR's own CI — the cygwin leg should now go green-with-warning if the installer download flakes, and still run the smoke test normally when it doesn't.


rjc123:
Heads-up on the red checks — they aren't from this change.

This PR is workflow-only (`.github/workflows/pr.yml`, +18/-1); it touches no Go. The failing `Contract corpus` job is `TestProtocol_GrantingReplicaRoundTripsJSONL`, a **new** test added by the recent replica-guard work (`c00e4d6a9`, wy-jpd3.7) that sits at the tip of `main` this branch is cut from. The failure is that tightened guard skipping a cross-replica reclaim (`lease granted by replica "mini", not this node "laptop"`) — i.e. it fails inside the code that commit changed, not anything here. It surfaces on this PR only because the branch is based on the freshest `main` (needed to include the `windows-make-shell` job this change edits).

Flagging so it isn't attributed to the workflow change — and you may want to check whether that new guard test is flaky or genuinely red on `main`. (`CI Gate / Required` is just the aggregator going red off the back of it.)

maphew (approved):
Verdict: **approve**, and your heads-up comment about the red `Contract corpus` check is confirmed and root-caused below — it is not this PR, and it is currently redding *every* fresh PR.

## The workflow change

The `outcome` vs `conclusion` distinction is used correctly (`continue-on-error` rewrites `conclusion` to success; `outcome` keeps the real result), the smoke step's gate means a genuine Cygwin Make regression still fails the job, and the `::warning::` annotation keeps the flake visible instead of silently green. Verified the flake is real and current: on #5032 the installer download failed all 5 retries and took down the required gate for a one-line test-env change. Same pinned action SHA, +18/−1, workflow-only.

The honest tradeoff — a persistent (non-flake) breakage of the installer action itself would now warn instead of block — is worth it: that failure mode is third-party infrastructure either way, and the annotation plus a periodically-green-with-warning pattern is auditable.

## The Contract corpus red (not this PR)

Root-caused: `TestProtocol_GrantingReplicaRoundTripsJSONL` (`cmd/bd/protocol/lease_claim_test.go`) can currently never pass. The replica guard *works* — the reclaim correctly skips and reports `"count": 0` — but the guard's stderr audit line (`reclaim: skipping <id> ... lease granted by replica "mini"`) names the issue ID, and the test's helper `tryRunEnv` captures `CombinedOutput()`, so the assertion `strings.Contains(reclaimOut, leased)` at line 392 trips on the audit courtesy line rather than on any actual reclaim. The JSON payload on stdout proves nothing was reclaimed.

It reds every fresh PR but not `main` itself because the `Contract corpus` job exists only in `pr.yml` — the test landed on main (c00e4d6a9, wy-jpd3.7) without ever running in that job. Verified the identical failure on three other unrelated PR branches this afternoon. Fix belongs in the test: assert on the parsed JSON `reclaimed` list (stdout) instead of substring-matching combined output. Being routed as a stop-the-line fix on our side.

Merge tail for this PR: once that test fix lands on main, an update-branch/rerun here should go fully green.

PR #5042, opened 2026-07-25
fix(uow): self-heal fresh-bootstrap debris instead of failing proxied init (#5012)

maphew:
# fix(uow): self-heal fresh-bootstrap debris instead of failing proxied init

Fixes the dominant red-main flake tracked in #5012 (`Test (proxied-server cmd 1/15)`).

## Why

A fresh `bd init --proxied-server` can fail permanently with

```
[mysql] connection.go:215 busy buffer
Error: failed to open uow provider: uow: init schema: uow: migrate:
pending schema migrations alter pre-existing dirty tables: issues; run 'bd dolt commit' ... (#4566)
```

on a database that did not exist seconds earlier. The failure needs two moves, and current code supplies both:

1. **A mid-pass transient on the shared server.** Fresh bootstrap runs every migration (59 today) with a Dolt commit per step on one pinned session. Under CI's parallel-test load a step can die between its SQL and its per-step commit: a genuine 1213/1205 rollback, or a poisoned session whose next prepared statement hits the driver's `busy buffer` (that breadcrumb is logged at go-sql-driver `connection.go:215` when `ReleaseMigrationLock`'s `SELECT RELEASE_LOCK(?)` prepare finds unread result data on the connection — its `ErrMigrationLockRelease` is then `errors.Join`ed into the failure and classified retryable). Either way the step's DDL/DML is already applied to the working set but never Dolt-committed.
2. **The #4566 guard reads the debris on retry.** `initSchema`'s backoff re-runs `MigrateUpWithLock`; the pre-flight dirty-table guard sees `issues` dirty with pending migrations touching it, classifies the pass's *own half-applied step* as "pre-existing dirty user data", and returns `DirtyTablesError` — which the retry loop rightly treats as permanent. A retryable transient has been converted into a hard init failure, and no number of retries can converge.

The per-step commit contract (#4566) deliberately leaves this one-step window open, because in general working-set dirt may be user data worth protecting. But on a **fresh bootstrap** — a database created by this very init call — the guard's protection target cannot exist: `bd` hands the database to no writer until init returns. Refusing there is strictly wrong.

This is the same failure chain the external contributor's #4504 diagnosed months ago from the 0040-era symptom (undrained `CALL DOLT_*` result sets → busy buffer → failed version INSERT / lock release). #4504 removes a *poisoning source* (drain proc result sets, make 0040/0041 idempotent); this PR makes the *retry converge* whatever the mid-pass failure mode was (deadlock, kill, network, poisoned buffer). They are complementary, and #4504 remains worth landing on its own merits.

## What

- `internal/storage/uow/dolt_sql_provider.go` — `initSchema` proves database ownership atomically: the first attempt issues a bare `CREATE DATABASE` (no `IF NOT EXISTS`), so the server arbitrates the race — success proves this init created the database (heal eligible, sticky across retries), an already-exists refusal (1007) proves it did not (heal off, guard refusal unchanged). See the review-hardening comment below for why this replaced the earlier existence probe.
- `internal/storage/schema/lock.go` — `MigrateUpWithLock` gains variadic `MigrateLockOption`s (existing callers unchanged). Under `WithFreshBootstrapHeal`, a `*DirtyTablesError` refusal is healed **on the same locked session**: `CALL DOLT_RESET('--hard')` discards the interrupted bootstrap's working set (per-step commits already in history survive), then the pass re-runs once. Holding the migration lock through reset + re-run means no concurrent migrator can be mid-pass when the reset fires.
- Pre-existing databases keep the guard's refusal exactly as before; embedded Dolt (`MigrateUp` without lock) is untouched.

## Reproduction / verification

`TestNewExternalDoltServerUOWProvider_FreshInitSelfHealsAfterMidPassFailure` reproduces the CI interleaving deterministically against a real Dolt container: the per-step fault hook fires once after migration 0001 commits, dirties `issues` in the working set the way a mid-step death does (SQL applied, no Dolt commit), and fails the pass with a retryable 1213.

- On main (4fd05a574) the test fails with the exact shard-1 error string: `uow: init schema: uow: migrate: pending schema migrations alter pre-existing dirty tables: issues; ... (#4566)`.
- With this PR it converges: debris discarded, cursor at `LatestVersion()`, clean `dolt_status`.

sqlmock tests pin both directions: `TestMigrateUpWithLockFreshBootstrapHealResetsAndRetries` (reset + re-run under one lock session) and `TestMigrateUpWithLockDirtyGuardStaysFatalWithoutHeal` (no option → unchanged `*DirtyTablesError`, no reset). `TestNewExternalDoltServerUOWProvider_PreexistingDirtyDatabaseIsNotHealed` (container) pins the ownership-negative path: a non-creator init over a dirty database keeps the #4566 refusal and leaves the working-set dirt untouched.

Preflight run locally: `go build`/`go vet`/`golangci-lint` clean on the touched packages; full `internal/storage/schema` and `internal/storage/uow` suites pass (including the existing 10-way concurrent-instantiation test); 3 serial iterations of the shard-1 test set (`TestProxiedServerConfig*`, `TestProxiedServerDuplicates`, `TestProxiedServerFindDuplicates`) with `BEADS_TEST_PROXIED_SERVER=1` all green.

## Notes for review

- The heal is scoped to the main-source pre-flight guard (`*DirtyTablesError`). The mid-pass ignored-source guard stays fatal; a retry after that failure lands in the healed pre-flight guard anyway.
- `internal/storage/dolt`'s server-mode `initSchemaOnDBWithRetry` has the same theoretical window but its caller doesn't know whether it created the database; left for a follow-up rather than widening this PR.
- Why frequency rose recently (#5012's near-deterministic run): fresh-init cost grows with every added migration (a Dolt commit per step), so the mid-pass window widens monotonically; #4877's 0059 also added a heavy `UPDATE issues` recompute step. #5014's proxied-purge isolation reduced shared-server interference (Main is green since it merged), but the one-transient-from-permanent-failure structure remained.

### DIFF
--- a/internal/storage/domain/db/ddl.go
+++ b/internal/storage/domain/db/ddl.go
@@ -10,6 +10,12 @@ var validIdentifier = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*$`)
 
 type DDLSQLRepository interface {
 	CreateDatabaseIfNotExists(ctx context.Context, database string) error
+	// CreateDatabase issues a bare CREATE DATABASE (no IF NOT EXISTS) so the
+	// server arbitrates creation atomically: success proves this call created
+	// the database; an already-exists error (MySQL 1007) proves it did not.
+	// The error is returned unmapped (wrapped with %w) so callers can
+	// classify it against their driver.
+	CreateDatabase(ctx context.Context, database string) error
 	UseDatabase(ctx context.Context, database string) error
 }
 
@@ -30,6 +36,17 @@ func (r *ddlSQLRepository) CreateDatabaseIfNotExists(ctx context.Context, databa
 	}
 	if _, err := r.runner.ExecContext(ctx, "CREATE DATABASE IF NOT EXISTS "+ident); err != nil {
 		return fmt.Errorf("db: CreateDatabaseIfNotExists: %w", err)
+	}
+	return nil
+}
+
+func (r *ddlSQLRepository) CreateDatabase(ctx context.Context, database string) error {
+	ident, err := quoteIdentifier(database)
+	if err != nil {
+		return fmt.Errorf("db: CreateDatabase: %w", err)
+	}
+	if _, err := r.runner.ExecContext(ctx, "CREATE DATABASE "+ident); err != nil {
+		return fmt.Errorf("db: CreateDatabase: %w", err)
 	}
 	return nil
 }
--- a/internal/storage/schema/lock.go
+++ b/internal/storage/schema/lock.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"hash/fnv"
+	"strings"
 	"time"
 )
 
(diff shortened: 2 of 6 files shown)


maphew:
Cross-vendor review (Codex, structured diff review) raised a valid P1 on the first revision: the `SHOW DATABASES` existence probe was not atomic with `CREATE DATABASE IF NOT EXISTS`. Two concurrent `initSchema` calls racing on the same absent database could both pass the probe before either created it, so both recorded `createdFresh=true` while only one created it — and the losing caller could then enable the heal. The exploitable interleaving is narrow (it additionally needs the loser to later hold the migration lock, a `DirtyTablesError`, and dirt that isn't bootstrap debris — e.g. binary-version skew plus another actor's uncommitted writes), but the reviewer's core point stands: ownership should be **proven**, not inferred from a pre-lock probe.

Addressed in 9555e8281, with a stronger mechanism than lock-scoping the probe (which would still be TOCTOU against creators that don't take the migration lock, like server-mode stores): let the server arbitrate. The first attempt issues a **bare `CREATE DATABASE`** (no `IF NOT EXISTS`):

- success is the atomic, server-serialized proof that this init created the database → heal eligible;
- an already-exists refusal (MySQL 1007, with the message fallback this codebase already uses for Dolt's variant in `internal/doltserver` and `internal/storage/dolt`) proves it did not → heal stays off, guard refusal unchanged.

The `created` flag is sticky across retry attempts because it records the one fact no later probe could re-learn (attempt 2 always finds the database existing — that was the original bug's shape). Retries after a recorded create re-assert with `CREATE DATABASE IF NOT EXISTS` so a database dropped between attempts is recreated rather than failing the `USE`.

New regression: `TestNewExternalDoltServerUOWProvider_PreexistingDirtyDatabaseIsNotHealed` (real Dolt container) — init over a pre-existing database with dirty `issues` must fail with the #4566 refusal verbatim and leave the working-set dirt untouched (asserted via `dolt_status` after the failed init). The self-heal repro and the existing 10-way concurrent-instantiation test still pass — under the new scheme that concurrent test now exercises exactly one bare-create winner and nine 1007 losers, all converging.

Preflight on 9555e8281: gofmt/vet/golangci-lint clean; full `internal/storage/schema`, `internal/storage/domain/db`, and `internal/storage/uow` suites green (`CGO_ENABLED=1 -tags gms_pure_go`).

One behavior note for the record: two truly simultaneous `bd init`s of the same brand-new database are no longer both guaranteed to converge if the winner dies mid-pass at exactly the wrong moment — the loser keeps the guard's refusal (correctness over availability; only the creator may discard a working set). The winner's own retry still heals and converges, and the CI harness always uses per-test unique databases, so the #5012 flake fix is unaffected.

PR #5044, opened 2026-07-25
fix(beads): admit the physical temp-root form in the temp-dir carve-out

maphew:
## Why

Main is red at 5a88c2b48: `Test (macos-latest)` fails six `TestContext*` tests with `BEADS_DIR points to unsafe location: /private/var/folders/...`. #5030's symlink-safe hardening is correct in `resolvedPathWithinRoot` (both sides resolved), but the temp-dir carve-out **gate** still matches only the unresolved `os.TempDir()` prefix — on macOS that's the `/var/folders/...` symlink form. A caller-supplied path that has already been symlink-resolved arrives in the physical `/private/var/...` form, skips the carve-out, and falls through to the `/private` deny prefix.

## What

The gate now also matches the physical resolution of the temp root (`resolveLongestExistingAncestor(tempDir)`); containment is still decided by `resolvedPathWithinRoot`, so #5030's symlink-escape rejection is fully preserved — this only widens which spellings *enter* the carve-out, not what it admits.

Regression test added. It distinguishes on macOS, where the physical temp root sits under the denied `/private` prefix — i.e. exactly the macos-latest lane that caught the original failure; on other platforms it is satisfied either way and documents the contract.

## Verification

`CGO_ENABLED=0 go build -tags gms_pure_go`, `gofmt`, and the full boundary+context test selection green locally.

Stop-the-line per Base-Branch Health: while main is red this fix is the only mergeable PR.

### DIFF
--- a/internal/beads/context_test.go
+++ b/internal/beads/context_test.go
@@ -516,6 +516,41 @@ func TestIsPathInSafeBoundary_TempDirSymlinkEscape(t *testing.T) {
 	}
 }
 
+// TestIsPathInSafeBoundary_TempDirPhysicalForm covers the macOS shape where
+// $TMPDIR is itself a symlink (/var/folders/... -> /private/var/...): a
+// caller-supplied path that has already been symlink-resolved arrives in the
+// PHYSICAL form, does not share the unresolved os.TempDir() prefix, and must
+// still be admitted by the temp-dir carve-out. On macOS the physical form
+// falls under the denied /private prefix, so without the carve-out this test
+// fails (the exact TestContext* regression from the be-kghzr hardening); on
+// platforms whose physical temp root is not under a denied prefix the
+// assertion is satisfied either way — the macos-latest CI lane is the
+// distinguishing runner.
+func TestIsPathInSafeBoundary_TempDirPhysicalForm(t *testing.T) {
+	base := t.TempDir()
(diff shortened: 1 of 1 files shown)


maphew:
Signal from the merge-queue side: main's `Main` workflow is still red after this landed — [run 30169298748](https://github.com/gastownhall/beads/actions/runs/30169298748) on d91a57fad fails in two distinct ways:

1. **macOS (`Test (macos-latest)`)**: the same `TestContext*` / role-test family as before this fix. The failing output shows bd operating under `/var/folders/8j/.../T/...` — macOS's real per-user temp root, which resolves physically to `/private/var/folders/...`. If the carve-out admits only the `/tmp` → `/private/tmp` pair, macOS `os.TempDir()` never matches either form.

2. **Linux (`Test (proxied-server cmd 5/15, 13/15)`)**: `TestIsPathInSafeBoundary_TempDirPhysicalForm` — the test added by this PR — fails there too, alongside `TestIsContributor`/`TestRole_*` collateral in the same package, so the boundary check appears to misbehave in the proxied-server test env as well, not just on macOS symlink resolution.

Mentioning because the merge patrol holds every green PR (currently four dependabot-queue PRs) while main's latest `Main` run is red, so this is the head of the queue. Related in-flight work: #5046.

PR #5045, opened 2026-07-25
fix(cook): validate runtime formula variables

### LINKED ISSUE #5040: bd cook ignores declared formula variable constraints
## What

`bd cook` accepts runtime `--var` values that violate the formula's declared `enum` or `pattern` constraints.

## Reproduction

Current upstream `main`: `13e61051df853b9cf24c35b0dd28cb9a53cb0553`

Given:

```toml
formula = "enum-validation"
version = 1
type = "workflow"

[vars.policy]
required = true
enum = ["merge-completes", "tracking-only"]

[[steps]]
id = "publish"
title = "Publish with {{policy}}"
```

Both forms exit zero:

```text
bd cook ./enum-validation.formula.toml --var policy=merge-comtes --dry-run
bd cook enum-validation --search-path . --var policy=merge-comtes --dry-run
```

I reproduced the zero exit through both the exact-path and registry-name loaders on Windows 11 with the current upstream build.

## Expected

Runtime cooking should fail before dry-run output, ephemeral output, or persistence when a supplied value violates a declared formula constraint. A declared enum value should continue to succeed.

## Cause and proposed scope

`internal/formula.ValidateVars` already implements required, enum, and pattern validation, with unit coverage. The production `runCook` path resolves the formula but never calls it.

The focused fix is to call the existing validator once after resolution whenever cooking is in runtime mode, and add command-level regression coverage for exact-path and registry-name loading. This does not add a new validation model or change compile-time placeholder cooking.

_codex-gpt-5.6-sol-ultra on behalf of Ewen Cuthiell_

ecuthiell:
## What

Validate resolved runtime formula variables during `bd cook` by calling the
existing `formula.ValidateVars` helper before dry-run output, ephemeral output,
or persistence.

The command-level regression covers both exact-path and registry-name loading.
Invalid enum values fail without writing preview output; valid declared values
continue to preview successfully.

## Why

`bd cook` currently accepts supplied values that violate declared `required`,
`enum`, or `pattern` constraints because the runtime command path never invokes
the validator that already implements those rules.

Fixes #5040.

## Verification

- `go test -tags gms_pure_go ./internal/formula ./cmd/bd -run '^(TestRunCookRejectsInvalidEnumVariable|TestValidateVars)$' -count=1`
- `go test -race -tags gms_pure_go ./cmd/bd -run '^TestRunCookRejectsInvalidEnumVariable$' -count=1`
- `make ci-pr-policy`
- `make ci-pr-lint`

I also ran `make ci-pr-core` on native Windows. It reached the full Go suite,
where pre-existing Windows path assumptions and an unrelated graph-render test
timeout prevented the wrapper from passing; the focused packages above passed
and no failure implicated the changed formula-cooking paths.

_codex-gpt-5.6-sol-ultra on behalf of Ewen Cuthiell_

### DIFF
--- a/cmd/bd/cook.go
+++ b/cmd/bd/cook.go
@@ -381,6 +381,11 @@ func runCook(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return HandleError("%v", err)
 	}
+	if flags.runtimeMode {
+		if err := formula.ValidateVars(resolved, flags.inputVars); err != nil {
+			return HandleError("%v", err)
+		}
+	}
 
 	protoID := resolved.Formula
 	if flags.prefix != "" {
--- a/cmd/bd/cook_test.go
+++ b/cmd/bd/cook_test.go
@@ -2,7 +2,13 @@ package main
 
 import (
 	"encoding/json"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
+
+	"github.com/spf13/cobra"
 
 	"github.com/steveyegge/beads/internal/formula"
 	"github.com/steveyegge/beads/internal/types"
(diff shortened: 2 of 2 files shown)


maphew (approved):
Maintainer review — verified locally, not just read: the new `TestRunCookRejectsInvalidEnumVariable` passes under `-race` with both build-tag configurations, the touched files have no drift against current `main`, and validation is correctly placed after `loadAndResolveFormula` (so `extends`-inherited vars are covered) and before all three output branches. The second commit's stdio-mutex fix for the parallel-test race was the right call. Merging via the patrol once green.

Two adjacent gaps we noticed are pre-existing (present on `main` too), so they don't gate this PR; we're tracking them as follow-ups on our side and can file upstream issues if useful:

1. An explicitly empty value (`--var policy=`, e.g. from an unset shell var) still bypasses enum/pattern validation — `ValidateVars` skips validation when `val == ""` regardless of whether the var was provided.
2. `bd mol pour` / `bd mol wisp` validate only variable *presence*, never enum/pattern, even though cook's help steers users to those commands.

Minor polish, take or leave: the fake cook command in the test defaults `--dry-run` to true, so only the dry-run branch is exercised; and the verbatim full-message assertions will be brittle against future wording changes. Neither blocks.

Thanks @ecuthiell — tidy fix.

Issue #5043, opened 2026-07-25
proxied-server open bypasses the remote-migrate gate and silently creates missing databases


maphew:
## Summary

The proxied-server open path auto-creates a missing database and applies pending schema migrations **without consulting the remote-migrate gate**, unlike both other storage paths. Verified at `5a88c2b48` (current main).

`internal/storage/uow/dolt_sql_provider.go` — `initSchema` (called from `openAndInitSchema` on every proxied open):

```go
if err := ddl.CreateDatabaseIfNotExists(ctx, database); err != nil { ... }
if err := ddl.UseDatabase(ctx, database); err != nil { ... }
if _, err := schema.MigrateUpWithLock(ctx, conn, database); err != nil { ... }
```

There is no `CheckRemoteMigrateGate*` call anywhere in `internal/storage/uow`.

Contrast with the guarded paths, which both evaluate the gate (with the fast-forward adopter) before migrating:

- server store: `internal/storage/dolt/store.go:1920` — `schema.CheckRemoteMigrateGateForRemoteWithRemoteCheckAndAdopt(...)` wired into `initSchemaOnDBWithRetryAndGate`
- embedded store: `internal/storage/embeddeddolt/store.go:344` — `schema.CheckRemoteMigrateGateWithAdopt(...)`

The gate's own contract says it is mode-independent (`internal/storage/schema/remote_migrate_gate.go:28-48`: "It applies to both server mode and embedded mode.") — proxied-server mode is the odd one out.

## Consequences

1. **Designated-migrator policy bypass.** On a remote-backed database that is behind the binary's schema, embedded/server opens refuse with the gate error and direct the user to the designated-migrator workflow. The same database opened in proxied-server mode silently migrates on first open — exactly what the gate exists to prevent for coordinated multi-machine databases.
2. **Silent empty-database creation.** `CreateDatabaseIfNotExists` means any proxied open of a workspace whose named database is absent quietly creates an empty one and migrates it to latest, instead of failing loudly. Combined with the fact that `dolt_mode` in `.beads/metadata.json` is committed and propagates to clones (the `bd migrate from-server-to-proxied-server` help text documents this), a collaborator pulling a proxied-mode workspace without having bootstrapped data gets a fresh empty database that *looks* healthy. This bit us while designing an embedded→proxied migration path and is the sharpest edge of the two.

## Suggested direction

[...]

PR #5046, opened 2026-07-25
feat(workspacegate): two-level cooperative workspace/physical-root gate (part 1: primitive)

maphew:
## Why

Embedded mode is deliberately multi-process (`internal/storage/embeddeddolt/store.go:34-41`, `concurrency_test.go`) and holds no store-level lock, and every existing lock in the tree is special-purpose (init-only, migration-mutual-exclusion, daemon lifetime, start serialization). Nothing lets a maintenance operation say "no other bd process may touch this workspace's data until I finish."

That gap is the first fatal finding from the adversarial design review of the embedded ↔ proxied-server copy-migration design (the missing piece after #4691, which is metadata-flip-only): a copy migration without a quiescence fence can silently lose a concurrent writer's commits. It also blocks any future safe restore / destructive-repair verbs. Related context: #5043 (why blind auto-create makes this worse).

## What

New leaf package `internal/workspacegate` (stdlib + `internal/lockfile` only) plus `beads.OpenGated`:

- **Shared/exclusive advisory gate**: normal commands hold SHARED for their store lifetime; maintenance ops hold EXCLUSIVE. No SH→EX upgrade (not portable, can self-deadlock) — callers classify and acquire their final mode once.
- **Gate files live OUTSIDE every root a gated operation may replace** (`.beads.gate.lock` beside `.beads`, `<root>.gate.lock` beside a server root), because flock follows inodes and a lock inside a replaced directory splits identity. Gate files are never deleted; a symlinked guarded dir is refused rather than gated ambiguously.
- **Two levels**: workspace gate + physical-root gate — distinct workspaces legitimately share one dolt server root, and a workspace-only gate can't stop workspace B restarting the server workspace A is draining. `AcquireAll` acquires in sorted canonical-path order (one total order → no deadlocks between overlapping sets) under a single total wait budget, and ranks before every existing beads lock.
- **Honest diagnostics**: advisory holder-info sidecar written atomically (O_EXCL temp + rename; never follows a pre-planted symlink), removed before unlock so a successor's record is never clobbered; busy errors qualify the sidecar against blocker mode and process liveness instead of quoting a dead PID as fact. `ExclusiveHolder` probes with a shared lock only and creates nothing.
- **`beads.OpenGated`**: opt-in participation for library embedders. Per the decorator contract on `AsIssueClaimer`, the wrapper embeds `storage.DoltStorage` (capability assertions keep working) and exposes `Unwrap`; gates release on clean `Close` only. `ErrGateBusy` exported for `errors.Is`.
- Gitignore coverage wired, not just documented: `ProjectGitignorePatterns` (project root) + `GitignoreTemplate`/`requiredPatterns` (`.beads` internal).

Deliberately cooperative: processes predating the gate and ungated `beads.Open` consumers acquire nothing. Operations needing a hard guarantee must refuse when they cannot assume sole cooperative ownership — the gate makes cooperating activity visible and excludes new work; it does not pretend to fence strangers. (That refusal posture, plus CLI acquisition wiring at the command chokepoints, is part 2.)

## Test plan

- [x] `go test -race -count=2 ./internal/workspacegate/` — 20 tests: same-process + cross-process exclusion (re-exec helper binary), crash release, AcquireAll ordering/dedupe/partial-failure/total-budget, symlink canonicalization + refusal, sidecar symlink hardening + atomicity, probe non-interference with fail-fast acquirers, wait-budget capping, child-process handle non-inheritance
- [x] `go test -race -run TestOpenGated .` — gate respected/released around the storage lifetime; capability-preservation regression (IssueClaimer/EventQuerier/BlockedQuerier/RemoteStore/SyncStore/VersionControlReader survive the wrapper)
- [x] `go test ./cmd/bd/doctor/ -run Gitignore` after template/pattern additions
- [x] `golangci-lint run` — 0 issues on changed packages
- [x] Full `make test` running in local verification queue (result recorded on the coordination-repo bead)

## Provenance

Design and diff each went through cross-vendor adversarial review (Codex gpt-5.6-sol high/xhigh + Claude opus): the design review killed a v1 with five fatal flaws (in-root lock placement, epoch-as-fence, library-bypass hole, SH→EX upgrade, missing physical-root level); the diff review caught the narrow-interface decorator regression, probe-induced spurious failures (~5% measured), stale-PID diagnostics, torn sidecar reads, and per-gate wait inflation — all fixed here with regression tests.

Part 1 of 2: this lands the primitive with no behavior change to any existing command. Part 2 wires classification-driven acquisition into the CLI chokepoints (and is where migration's refuse-by-default posture lives).


maphew:
Pushed 4ddaf4e28, responding to a two-vendor adversarial review (Claude Opus + GPT-5.6 review agents) of the original head.

**Scope change: this PR now lands the `internal/workspacegate` primitive only.** The review found the primitive itself sound — lock ordering, crash recovery (kernel-released flock, advisory-only sidecar), symlink refusal, and release-before-unlock all held up under targeted attack — but found two blockers in the `OpenGated` public API that was riding along:

1. Its physical-root derivation skipped the store's real mode resolution (central config, env vars, config.yaml, shared-server root), so under shared-server mode it gated a directory the store never opens.
2. With no `metadata.json` but an existing database directory (a layout the open path explicitly supports), it silently skipped the physical gate entirely.

Both are fixed properly by a single shared physical-root resolver, which part 2's CLI wiring needs anyway — so `OpenGated` moves to part 2 rather than shipping here with a private, subtly-wrong resolver.

Also in this push, hardening the primitive per review: case-folded comparison keys for `AcquireAll` dedupe/ordering on Windows/macOS; `busyDetail` now stale-qualifies the recorded holder PID in Shared mode too; `ExclusiveHolder` probes read-only; a Windows process-handle leak in `pidAlive` fixed; honest starvation documentation on `Exclusive`/`Options.Wait`; and explicit docs for the cross-user 0600 gate-file limitation and the in-`.beads` gate-placement invariant. New tests cover the stale-holder reporting in both modes, the comparison key, and `ExclusiveHolder`'s error contract.

PR #5052, opened 2026-07-25
fix(backup): run backup sync/restore on an undeadlined one-shot connection (bd-2o8nv)

steveyegge:
## Problem

`bd backup sync` (and restore) execute `CALL DOLT_BACKUP(...)` on the default server connection pool, whose DSN pins `ReadTimeout=10s` (`buildServerDSN`). Any sync slower than 10s dies with a `packets.go` i/o timeout / `invalid connection`. Every first sync to a remote destination (e.g. `gs://`) is slower than that — and because the dolt server aborts the transfer when the client connection drops, **retrying never converges**: observed 247KB of a 46M store uploaded after three attempts against `gs://ghosttrack-wyvern/backups/wyvern`. `file://` destinations complete in <2s, which is why this never surfaced before backups moved off local disk.

## Fix

Follow the existing `DOLT_PUSH`/`DOLT_FETCH` pattern (one-shot connection, long deadline), generalized into `oneShotConn(readTimeout)`:

- `execWithLongTimeoutNoTx` keeps its 5m deadline (unchanged behavior, now shares the helper)
- `BackupSync`, `BackupDatabase`'s syncs, and `RestoreDatabase` use **no read deadline** (`readTimeout=0`)

Why unbounded rather than a bigger constant: a 5m deadline was tested against a real `gs://` destination and still timed out at exactly 300s, and any fixed budget fails the same way for a slow-enough destination while making retries useless (server-side abort on client drop). Backup ops are only reachable from explicit CLI commands — no daemon path — so the command context governs cancellation. `BackupAdd`/`BackupRemove` stay on the pool (fast metadata ops).

## Testing

- `go build ./...`, `go vet`, `go test ./internal/storage/dolt/ -run 'Backup|Conn'` green
- Real-world: with the 10s pool conn the wyvern admin sync to GCS died at 10s; at 5m, died at exactly 300s; with this fix the full first sync to `gs://` completes (46M store)

Bead: bd-2o8nv

### DIFF
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -934,8 +934,15 @@ func (s *DoltStore) BackupAdd(ctx context.Context, name, url string) error {
 }
 
 // BackupSync pushes the database to the named backup destination.
+// Runs on a long-timeout connection: a sync to a remote destination
+// streams the database and outlives the pool's 10s ReadTimeout.
 func (s *DoltStore) BackupSync(ctx context.Context, name string) error {
-	return versioncontrolops.BackupSync(ctx, s.db, name)
+	db, err := s.oneShotConn(0)
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+	return versioncontrolops.BackupSync(ctx, db, name)
 }
 
 // BackupRemove removes a configured Dolt backup destination.
@@ -960,6 +967,12 @@ func (s *DoltStore) BackupDatabase(ctx context.Context, dir string) error {
 	}
 	backupName := "backup_export"
 
+	syncDB, err := s.oneShotConn(0)
+	if err != nil {
+		return err
+	}
+	defer syncDB.Close()
+
 	// Register as a backup remote (idempotent — remove first if exists).
 	_ = versioncontrolops.BackupRemove(ctx, s.db, backupName)
 	if err := versioncontrolops.BackupAdd(ctx, s.db, backupName, backupURL); err != nil {
@@ -967,14 +980,14 @@ func (s *DoltStore) BackupDatabase(ctx context.Context, dir string) error {
 		// already point to this URL. In that case, sync using the existing
 		// remote name rather than failing.
 		if conflict := versioncontrolops.ExtractAddressConflictName(err); conflict != "" {
-			if syncErr := versioncontrolops.BackupSync(ctx, s.db, conflict); syncErr != nil {
+			if syncErr := versioncontrolops.BackupSync(ctx, syncDB, conflict); syncErr != nil {
 				return fmt.Errorf("sync to backup: %w", syncErr)
 			}
 			return nil
 		}
 		return fmt.Errorf("register backup remote: %w", err)
 	}
-	if err := versioncontrolops.BackupSync(ctx, s.db, backupName); err != nil {
+	if err := versioncontrolops.BackupSync(ctx, syncDB, backupName); err != nil {
 		return fmt.Errorf("sync to backup: %w", err)
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


steveyegge:
Real-world verification complete: the full first sync of the wyvern admin store to `gs://ghosttrack-wyvern/backups/wyvern` succeeded with this fix — **14m17s**, 101MB landed, `bd backup status` reports the sync clean. That duration fails at the pool's 10s timeout and also at a 5m deadline (tested, died at exactly 300s), so the unbounded read deadline is load-bearing, not caution.

maphew:
Reviewed 2026-07-26 (head ab1f97f03). **Approve-with-nits** — the diagnosis is right (a single long-blocking `CALL DOLT_BACKUP('sync')` under a per-I/O read deadline is exactly the wrong control), the real-world 14m17s/101MB validation is convincing, and CI is green. Mergeable as-is if you prefer; two items are worth folding in first:

1. **Helper duplication** — `oneShotConn` is byte-for-byte the same DSN-reparse logic as the pre-existing `openLongTimeoutConn` (store.go:3268), and open PR 5090 adds two more callers of the latter. Suggest deleting `oneShotConn` and parameterizing `openLongTimeoutConn(readTimeout time.Duration)`: pass `5*time.Minute` at the three existing sites, `0` at the backup sites. Net code deletion, and whoever lands second between this and 5090 avoids two near-identical helpers.
2. **Cross-session add-then-sync** — in `BackupDatabase`, `BackupRemove`/`BackupAdd` still run on the pooled `s.db` while the following `BackupSync` runs on the new `syncDB` session, relying on `dolt_backup('add')` being immediately visible across sessions. Passing `syncDB` to all three is strictly simpler and removes the question (the standalone store methods can stay pooled).

Noting for a follow-up bead rather than blocking: the PR body's "only reachable from explicit CLI commands" isn't quite true — `maybeAutoBackup` (cmd/bd/backup_auto.go:78) runs from `PersistentPostRun` and reaches `BackupDatabase` when `backup.enabled=true` on a shared-filesystem server, so with `readTimeout=0` a wedged server turns auto-backup into an indefinite hang of any ordinary `bd` command. A bounded context in `maybeAutoBackup` keeps the explicit sync/restore paths unbounded as designed. Also nit-level: a `oneShotDSN(readTimeout)` split would make the load-bearing zero unit-testable, and zeroing `WriteTimeout` alongside (as `openMigrationDB` does) or a one-line comment on why write stays bounded.

steveyegge:

Clean generalization of the proven DOLT_PUSH one-shot pattern. Points that carried the verdict:

- `ReadTimeout = 0` (no deadline) for backup sync/restore is the right call, and it's argued from evidence, not caution: the 5m variant died at exactly 300s against a real gs:// destination, and the verified real sync took 14m17s. A fixed budget on a monotonically-growing store is a time bomb; the server aborting the transfer on client drop means retries can never converge — so 'bigger timeout' was never a fix.
- Cancellation still exists: ExecContext carries the command context, and backup ops are only reachable from explicit CLI invocations, so an operator's interrupt is the governing bound. That's the same contract as running dolt's own CLI.
- Scope discipline is right: BackupAdd/Remove stay on the pool (quick metadata), only the streaming ops move; push/pull keep their tested 5m via the same extracted helper, so behavior there is bit-identical.
- The BackupDatabase address-conflict fallback path also got the one-shot conn — the easy path to miss; good that it's covered.

One cosmetic nit, fine to leave: `oneShotConn`'s error strings still say 'long-timeout connection', which reads odd for the no-deadline case — harmless, tidy up whenever that file is next touched.

CI 73/73 green + real-world receipt on the PR. Ship it — and glad this unblocks the backups-to-GCS ruling.

PR #5060, opened 2026-07-26
feat(cli): opt out of store init via bd:skip_store annotation

julianknutsen:
## What

Adds a cobra-annotation seam so a command can opt out of database/store
initialization in `PersistentPreRunE` by declaring
`Annotations{"bd:skip_store": "1"}` on itself (or any ancestor), instead of
being added to the central `noDbCommands []string` in `cmd/bd/main.go`.

Mechanism: a small `commandOptsOutOfStore(cmd)` helper (walks the command's
ancestor chain) OR-d into the existing store-skip gate. One new const, one
helper, one gate line.

## Why

`noDbCommands` is a hardcoded slice in one file. A command defined in another
file — or a build-tagged variant, or a downstream/plugin command — cannot add
itself to the exemption without editing that shared list, which is a recurring
merge-conflict point. An annotation lets a command declare "I don't need the
store" locally, where it's defined. This generalizes the existing behavior;
the name list still works and is untouched in spirit.

## Is this a no-op today?

For upstream, effectively yes — no command sets the annotation *by default*.
To avoid adding a seam with zero consumers, this **dogfoods** it on `doctor`:
`doctor` now opts out via the annotation on its `Command` literal instead of a
`noDbCommands` entry. Behavior is identical (the annotation walk yields the
same result for `doctor` and its subcommands as the old name/parent match, and
is actually more precise — it's pinned to the specific command literal). The
existing `doctor` `PersistentPreRunE` tests now exercise the annotation
end-to-end.

## Tests

- `commandOptsOutOfStore` unit coverage: self / parent / grandparent /
  non-"1" value / nil-map / nil-command.
- `TestPersistentPreRunHonorsSkipStoreAnnotation`: an end-to-end, self-validating
  gate test. It drives `PersistentPreRunE` with an annotated command (must skip
  store init and not open the store) **and** a non-annotated control command in
  the same environment (must reach store init and fail to open the absent DB) —
  so the annotated command's clean return is a meaningful discriminator, and a
  dropped gate fails loudly rather than silently.

`go vet` clean, `gofmt` clean, `go.mod` unchanged.

### DIFF
--- a/cmd/bd/doctor.go
+++ b/cmd/bd/doctor.go
@@ -68,7 +68,11 @@ const ConfigKeyHintsDoctor = "hints.doctor"
 var doctorCmd = &cobra.Command{
 	Use:     "doctor [path]",
 	GroupID: "maint",
-	Short:   "Check and fix beads installation health (start here)",
+	// doctor diagnoses installation health and must run without opening the
+	// store. It opts out of store init via the annotation seam rather than the
+	// noDbCommands list (see commandOptsOutOfStore in main.go).
+	Annotations: map[string]string{skipStoreAnnotation: "1"},
+	Short:       "Check and fix beads installation health (start here)",
 	Long: `Sanity check the beads installation for the current directory or specified path.
 
 This command checks:
@@ -495,7 +499,7 @@ func runDiagnostics(path string) doctorResult {
 	}
 
 	// bd-jgxi: Auto-migrate database version before checking it.
-	// Since doctor skips PersistentPreRun DB init (it's in noDbCommands),
+	// Since doctor skips PersistentPreRun DB init (via skipStoreAnnotation),
 	// trackBdVersion() and autoMigrateOnVersionBump() haven't run yet.
 	//
 	// Scope version tracking to the doctor target. Without this, `bd doctor <path>`
--- a/cmd/bd/doctor_conventions.go
+++ b/cmd/bd/doctor_conventions.go
@@ -11,7 +11,8 @@ import (
 // runConventionsCheck runs a composite conventions check: lint, stale, and orphans.
 // All findings are advisory (warning, never error) - conventions are a choice.
 func runConventionsCheck(path string) error {
-	// doctor is in noDbCommands so PersistentPreRun doesn't open the store.
+	// doctor opts out of PersistentPreRun DB init via skipStoreAnnotation, so
+	// PersistentPreRun doesn't open the store.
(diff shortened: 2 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98636277

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98636647

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds a `bd:skip_store` command annotation that lets an individual `bd` command opt out of store initialization in `PersistentPreRunE`, instead of every no-store command having to be enumerated centrally in `noDbCommands`. That matters because it gives commands defined outside `cmd/bd/main.go` a local, self-describing way to skip store init, and it converts `doctor` to that mechanism as the first consumer.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- No maintainer fixes were required. Review verified the opt-out is narrowly keyed to `bd:skip_store == "1"`, is nil-safe while walking the Cobra ancestor chain, and leaves the existing name-based `noDbCommands` skip logic and the protected Dolt-subcommand exceptions intact.
- Confirmed `doctor` now opts out via its command annotation while doctor subpaths that genuinely need storage (for example the conventions check) still initialize it lazily, so no doctor path loses its database access.

Review findings addressed:
- Behavioral correctness / architectural consistency (no finding, claude+codex): the annotation override only augments the final store-init skip decision; existing classification and Dolt-subcommand exceptions are preserved.
- Test evidence quality (no finding, claude+codex): the new `skip_store_annotation_test.go` covers the helper truth table, nil safety, and end-to-end `PersistentPreRunE` behavior, and the focused tests pass.

Remaining non-gating notes:
- One documented, non-gating footgun: because the helper walks the whole command ancestor chain, applying `bd:skip_store` to a parent command that has store-needing children would skip init for those children too. It is not a concern for the current `doctor` usage.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `5615f87` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/30180827709

### Review Iterations
1 review pass performed (multi-model Claude + Codex review with synthesis and a quality scorecard); no further iterations were required.

---

PR #5064, opened 2026-07-26
fix(schema): survive dolt#11131 encoding drift in the aux row re-key (#4380)

### LINKED ISSUE #4380: Aux row re-key crashes the migration pass on dolt#11131-class drifted storage (server panic 'invalid hash length: 19') — fix: skip-and-warn per table
## Problem

On databases carrying the dolthub/dolt#11131 class of storage drift (TEXT/LONGTEXT columns ALTERed under a pre-2.1 Dolt: row bytes left un-rewritten), **the aux row re-key crashes the whole migration pass mid-way**.

The re-key (bd-6dnrw.2 lineage) scans every column of `events`/`comments`/`issue_snapshots`/`compaction_snapshots`. Decoding one of the affected cells makes the Dolt **server** panic and drop the connection:

```
Error 1105 (HY000): panic recovered: invalid hash length: 19
  github.com/dolthub/dolt/go/store/val.AdaptiveValue.convertToTextStorage
  github.com/dolthub/dolt/go/store/val.(*TupleDesc).GetStringAdaptiveValue
```

The migration pass dies between cursor updates, leaving a dirty working set; subsequent passes hit the dirty-table gates. On our single shared sql-server (~140 databases, 28 with drifted tables) this blocked a fleet-wide migration drain until patched.

The underlying storage defect is dolt-side — we reported our variant as dolthub/dolt#11205 (follow-up to #11131, filed at @timsehn's request); affected cells are not even repairable from SQL (`UPDATE`/`DELETE` of the bad rows panic the same way). So bd cannot assume the data can be fixed first: **the migration path itself has to survive these tables.**

## Fix we ship on our fork (offer to PR)

[`767d5dcdef`](https://github.com/marcodelpin/beads/commit/767d5dcdefbaa497da30b8da1e37f44c4228e198): the re-key detects the panic signature per table, logs a loud warning, **skips that table and continues the pass**. Migrations complete; the skipped table keeps its old per-clone row ids.

Rationale for skip-over-fail: on a shared single sql-server the cross-clone PK convergence the re-key provides is moot for that table anyway, while a crashed migration pass takes the whole database out of upgradability. The warning keeps the skip visible to operators.

Given this week's bd-578h9 hardening series touches the same pass, happy to send the PR rebased on current main — or adapt the detection (it keys on the `invalid hash length` panic substring) to whatever shape you prefer.

maphew:
## Why

Fixes #4380.

A database whose `events` / `comments` storage carries dolthub/dolt#11131 encoding drift cannot be opened by bd 1.1.0 at all. The aux row-id re-key scans every column of the four aux tables; on a drifted table that scan panics inside Dolt (`invalid hash length: 19`), which aborted the whole `MigrateUp` pass — and since the pass is re-attempted on every open, the panic recurred on every start. The only build that could read the data was the pre-re-key one, so affected users had **no upgrade path onto 1.1.0**, forward or back.

Five independent environments confirmed it on #4380, four of them on released Homebrew `bd 1.1.0` + `dolt 2.1.10`, including single-user remote-backed databases — this is not a fleet-only or shared-server-only failure. @aaronlippold summarised it as "a breaking change for upgrade from 1.0.5 to 1.1.0".

The drift itself is not bd's to repair: the rows are physically undecodable, no SQL read or write can touch them, and healing them needs Dolt's `schema-encoding-drift recover-rows` (dolthub/dolt#11133), which @marcodelpin has since verified end-to-end on a live database. Dolt 2.1.10 stops *new* drift, and bd's own migration 0057 already guards the 0048 re-application that produced it locally. What was still missing is that **the migration path has to survive tables it cannot read**. That is all this PR does.

## Credit

The fix is @marcodelpin's, from [`767d5dcd`](https://github.com/marcodelpin/beads/commit/767d5dcdefbaa497da30b8da1e37f44c4228e198) on their fork, absorbed here with the semantic rework requested in [my review on #4380](https://github.com/gastownhall/beads/issues/4380#issuecomment-4897269733) rather than leaving them to do another round trip. The diagnosis behind it — the chunk-level characterisation, the negative result on the minimised fixture, the `dolt 2.1.10` re-test, the `recover-rows` verification — is theirs too, on the issue. `isSchemaEncodingDriftErr` and its table-driven test come from that commit.

## What changed vs. the fork commit

The fork's skip cleared the in-progress sentinel and returned nil, so `ignoredSource.migrate` recorded marker version 9 later in the same pass and `!markerPending` then barred re-entry forever. The warning promised "re-run the re-key after the storage drift is repaired", but no path existed to do so. On a single shared sql-server that loss is moot (one copy of every row, clones never merge), as the commit message says; upstream serves merge topologies, where the skipped table would keep its per-clone-random 0037 ids permanently.

So the skip is now **recorded and resumable**:

- **New clone-local record** `aux_row_rekey_drifted` in `local_metadata` (dolt-ignored, like the sentinel), naming the skipped tables. It re-admits the re-key after the marker is recorded, and a **post-marker pass re-keys only the recorded tables**. The others converged in the completed pass, and rows minted since carry app-minted ids already consistent across clones — re-keying those on one clone alone would manufacture the divergence the re-key exists to remove. Scoping keys on the marker, not on the sentinel, so a retry that itself dies on some unrelated error stays scoped rather than widening back to all four tables on the next pass.
- **The in-progress sentinel keeps its exact meaning** — "a rewrite is in flight" — and is cleared on a skip like any completed pass. It is deliberately *not* reused as the durable record: `MigrateUp` reads it to exempt the aux tables from the pre-existing-dirty and changed-signature guards, which is right for a crashed pass's partial UPDATEs sitting in the working set, but leaving it set indefinitely would relax those guards for the life of the clone.
- **`MigrateUp` extends that exemption to the recorded drifted tables**, and it matters more there than for the crash case: the changed-signature guard reads dirty tables through `dolt_diff`, which decodes exactly the cells that panic, so a drifted table left in `dirtyBefore` would fail the pass on the *read* and re-create the unopenable database this is meant to end.
- Both records are read in one round trip (`readAuxRekeyState`), so `MigrateUp` and the re-key each cost one existence probe instead of two apiece.

`TestRekeyAuxRowIDsKeepsSentinelOnFailure`'s invariant is untouched: a non-drift failure still aborts and still keeps the sentinel.

## Things worth knowing before you merge

- **A resumed pass re-keys the whole drifted table**, including rows minted between the skip and the repair, exactly as the existing crash-resume path does. That is why the retry rides every subsequent migration pass instead of waiting to be asked — the sooner it lands, the smaller that set.
- **The retry needs a pass to run.** `MigrateUp` short-circuits on `migrationWorkNeeded`, so an affected clone in steady state retries when the next migration ships, not on every open. Deliberate: the alternative is re-scanning a large `events` table on every command for a repair that may be months away. An explicit trigger (`bd doctor` surface, or a flag on `bd migrate`) is the obvious follow-up and I'm happy to take direction on it.
- **`local_metadata` is ephemeral** — migration 0030 says so explicitly: recreated empty on a working-set reset. A reset therefore loses the drift record and makes the skip permanent and silent. The in-flight sentinel has carried that same fragility since bd-578h9.16; flagging it rather than pretending otherwise.
- **Drift also produces readable garbage** (~10% of affected cells in @marcodelpin's characterisation return a zero payload). Those scan without error, so their rows are re-keyed from corrupt content and converge to an id no healthy clone derives. Nothing here can detect that; it is one more reason the storage repair is still required. The code comment says so.
- Warnings go to the standard logger, not the package's `stderr` progress writer, which is TTY-gated to `io.Discard`. That writer is right for progress chatter; a silent skip for piped/CI/agent callers — the ones least likely to notice divergent ids later — is not. This reverses the mechanical suggestion in my own review, for that reason.

## On the classifier

Kept as the bare `invalid hash length` substring rather than tightened to `panic recovered: invalid hash length` or to `: 19`, also reversing an optional suggestion in my review, because both narrower forms have a worse failure mode here — a miss falls through to the abort that leaves the database unopenable:

- The width varies with the direction of the tag flip. Migration 0057's own comment records both `invalid hash length: 19` on read and `invalid hash length: 1` on insert/merge for this same drift.
- The `panic recovered:` wrapper is added by the SQL engine; a path surfacing the panic unwrapped would not match.

The cost is bounded: dolt raises the phrase from `hash.New` generally, but reaching this classifier at all means the re-key could not read the table, so skipping and recording it is the same correct answer. Everything without the phrase still aborts the pass.

## On the storage boundary

Classifying a Dolt panic message inside `internal/storage/schema` brushes against the driver direction in `docs/PROJECT_CHARTER.md`. This package already hosts Dolt-message classifiers in the same style (`RemoteRefUnavailableErr`, `MissingMigrationObjectErr`), and this is migration survival for rows that are unreadable by design, so it seems acceptable here — flagging it so the storage-layer owners can say whether drift detection should eventually move behind the driver interface. @coffeegoddd.

## Testing

- `go test ./internal/storage/schema/` and `./internal/storage/...` green; `go vet` and `gofmt` clean.
- New in `aux_row_id_drift_test.go`: the classifier table (from the fork, plus the `: 1` and unwrapped-panic cases); a drifted table skipped while the pass completes and the record is written; **drift surfacing mid-rewrite** rather than on the scan, asserting the half-re-keyed table is recorded and committed; a non-drift error still aborting; a post-marker pass touching **only** the recorded table; the same when the sentinel is also set from a crashed retry; and the record surviving a still-unrepaired retry. sqlmock matches in order, so "exactly one table probe" is a real assertion that the other three are untouched.
- Two existing tests updated for the changed probe shape, not for changed behaviour.

Not exercised end-to-end against real drifted storage — the panic cannot be produced by any plain Dolt operation (see dolthub/dolt#11131), and @marcodelpin found their minimised fixture still carried recoverable fragments of production content, so it could not be shared. Their standing offer to run diagnostics by proxy against the drifted database is the way to close that gap; I'd like to take them up on it for this branch before it lands.

### DIFF
--- a/internal/storage/schema/aux_row_id_backfill.go
+++ b/internal/storage/schema/aux_row_id_backfill.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"log"
+	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


maphew:
Status note: this fix shipped today as hotfix release [v1.1.2](https://github.com/gastownhall/beads/releases/tag/v1.1.2), cherry-picked onto a v1.1.0-based branch (`hotfix/v1.1.1`, commit 564049bb8) so bricked 1.1.0 users regain an upgrade path — see the announcement on #4380. (The v1.1.1 tag exists but was never published; its release run failed on a stale MCP lockfile and the number was burned rather than rewriting a pushed release tag.)

This PR remains the main-line landing and its gate is unchanged: it still wants storage-owner input on the classifier-in-schema-package question, and reporter validation against real drifted storage — which the shipped release should now make possible.

maphew:
Reviewed at `2fd873e4`.

The classifier and skip/resume focused tests pass, but the resumed path has one dirty-working-set gap. `schema.go` removes every recorded drift table from `dirtyBefore` before knowing whether repair has made the table readable. Once readable, unrelated uncommitted edits can therefore be swept into the migration commit with the rekey.

Please retain the exemption only while `dolt_diff` still fails with the drift signature; after repair, signature-check and preserve pre-existing dirties normally. The branch also needs rebasing, storage-owner input, and the promised real drifted-store validation. This overlaps #4804's migration-open fast path but solves a distinct auxiliary-row corruption case; keep it separate and avoid broadening either patch across the storage boundary.

maphew:
Rebased onto current `main` (`1b67d0b1f`) across the multi-pass re-key restructure (`423afdcb2`), which is where the conflict lived: main split the re-key into gated passes (`auxRekeyPasses`) while this PR extends the sentinel with a drift record.

How the reconciliation resolves that: the carry keeps **one shared drift record across both passes**. Both passes run the identical `rekeyAuxRowTable` rewrite over the identical `auxRekeyTables` set - only their gates differ - so whichever pass retries first does the work, and clearing the record correctly makes the remaining pass a no-op. That invariant is now documented on `auxRowRekeyDriftedKey` and pinned by a new test at the `rekeyAuxRowIDsAllPasses` level.

Two hardening changes landed alongside, out of a second review pass on the reconciliation:

- `readAuxRekeyState` filters recorded names against `auxRekeyTables`, so a stray or hand-edited drift-record value can never exempt an arbitrary table from `MigrateUp`'s dirty guards.
- The state read moved ahead of the marker short-circuit (drift re-admission requires it). Cost: migrating opens pay one extra `INFORMATION_SCHEMA` probe per pass; steady-state opens pay nothing (`migrationWorkNeeded` short-circuits earlier).

`internal/storage/schema` suite green locally on the pushed head (`CGO_ENABLED=0`, `gms_pure_go`).

maphew:
Rebased onto current main to clear the merge conflict.

The conflict was in `lock_test.go`: main's cursor-existence probe (from #5847, be-bv7x) and this branch's multi-pass comment rewrite touched the same mock-expectation block. Resolution keeps both - the probe expectation plus the per-pass drift-state reads. The same probe expectation was also added to the drift/backfill tests this branch introduces (7 call sites), since `pendingVersions` now probes `ignored_schema_migrations` existence before reading the cursor.

`go test -tags gms_pure_go ./internal/storage/schema/` is green locally; no production-code changes, test-only reconciliation.

julianknutsen:
Maintainer rebase: retargeted onto `release/1.3.0` for the 1.3.0 release.

This is a release-critical safety fix — a database carrying dolthub/dolt#11131 encoding drift cannot be opened at all until the aux row re-key survives the tables it cannot read — so it should ship with 1.3.0 rather than wait on the main-line queue. Per the maintainer guidelines, rebases are our work, not yours; nothing is needed from you here.

## What moved

- Old base: `b7175be22` (ancestor of `main`)
- New base: `71377f276` = `origin/release/1.3.0`
- Old head: `cfca21760` → new head: `893f4f895`
- PR base changed `main` → `release/1.3.0`

## Conflicts

None. `release/1.3.0` was cut at the same commit as `main` (`71377f276`), and the branch was already `MERGEABLE` against `main`, so both commits replayed cleanly with no resolution required. The resulting diff is byte-identical to the pre-rebase one: 6 files, +754/-70, matching the counts GitHub showed before the retarget. Your earlier reconciliations — the multi-pass `auxRekeyPasses` carry with one shared drift record, and the `lock_test.go` cursor-existence-probe merge from #5847 — are all intact and untouched.

## Your commits are preserved

Both commits were rebased, not squashed, and keep their original authorship:

- `2d323db35` — `fix(schema): survive dolt#11131 encoding drift in the aux row re-key (#4380)`, authored by @maphew with `Co-authored-by: Marco Del Pin <marco.delpin@gmail.com>` intact
- `893f4f895` — `ci: retrigger - lint failed on a transient golangci schema fetch timeout`

No maintainer commit was added; the force-with-lease push carries only the rebase.

## Verification on the rebased head

Run with the repo's canonical flags (`CGO_ENABLED=1`, `-tags=gms_pure_go`):

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l internal/storage/schema/` — clean
- `./scripts/test.sh ./internal/storage/schema/...` — `ok` (the only package the diff touches)
- `./scripts/test.sh ./internal/storage/...` — all packages `ok`, including `dolt`, `embeddeddolt`, and `uow`, since `MigrateUp`'s dirty-guard change is consumed there

Every test this PR adds runs and passes on the new base, including `TestIsSchemaEncodingDriftErr`, `TestRekeyAuxRowIDsSkipsDriftedTable`, `TestRekeyAuxRowIDsSkipsDriftDuringRewrite`, `TestRekeyAuxRowIDsAbortsOnNonDriftError`, `TestRekeyAuxRowIDsResumesDriftedTableOnly`, `TestRekeyAuxRowIDsResumeStaysScopedAfterCrash`, `TestRekeyAuxRowIDsKeepsDriftRecordWhileUnrepaired`, and `TestRekeyAuxRowIDsAllPassesSharesDriftRecordAcrossPasses`.

## Still open from the thread (for the reviewer, not for you)

Carrying these forward so the retarget does not bury them:

- The 2026-08-02 review asked that the `dirtyBefore` exemption be retained **only** while `dolt_diff` still fails with the drift signature, so that after repair a readable table's pre-existing dirty edits are signature-checked normally instead of swept into the migration commit. On the current head `MigrateUp` still drops every recorded drifted table from `dirtyBefore` unconditionally. Reviewer's call whether that is a merge blocker or a follow-up bead.
- Storage-owner input on hosting the dolt-message classifier inside `internal/storage/schema` (@coffeegoddd), raised in the PR body.
- Validation against real drifted storage, which @marcodelpin offered to run by proxy.
- The explicit retry trigger (a `bd doctor` surface or a flag on `bd migrate`), noted in the PR body as an obvious follow-up.

## Next

Labeled `status/needs-review-auto`; the auto-review pipeline takes it from here. Not merged and not approved by this action. If `release/1.3.0` does not reintegrate into `main`, this will need a forward-port — flagging it so the main-line landing this PR was originally opened for is not lost.

Thanks for carrying @marcodelpin's fix through the semantic rework and three rebases rather than sending it back around.

_claude-opus-5-unknown-reasoning on behalf of Julian Knutsen_

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277258442

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @maphew! This PR makes `bd` survive the dolt#11131 encoding-drift condition in the aux row-id re-key migration (#4380). Before this change, a database that had accumulated that drift would panic in `rekeyAuxRowTable` and abort `MigrateUp` on every open, leaving the database effectively unopenable. Your fix detects the encoding-drift error, skips and records the affected aux table so the migration completes and the database opens, and re-keys the drifted tables on a later pass — which is exactly the kind of graceful-degradation-plus-recovery behavior that keeps existing users' databases from getting stuck.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The first version of the fix exempted **all** aux tables from the `dirtyBefore` edit-protection on a post-marker resume pass. On a resume where only some tables had drifted, that could drop a non-drifted table's pre-existing user edits from protection. The final patch scopes the exemption to exactly the drifted/rewritten tables by deriving it from the same `tablesToRekey` pass selector that the rewrite uses (`schema.go`), so the exempt set is a subset of the rewritten set by construction and a non-drifted table stays protected.
- The retry-state branching had grown into a complexity hotspot in `rekeyAuxRowIDsPending` / `MigrateUp`. The final patch lifts the gate and table-selection logic into a focused, pure `tablesToRekey` helper (`aux_row_id_backfill.go`), which also removes the risk of the exemption path and the rewrite path drifting apart.
- A regression test, `TestAuxRekeyExemptTablesScopedToRewrittenTables`, now pins the scoped exemption: exempt equals the drifted subset on a post-marker crashed-sentinel resume, all four tables on a marker-pending pass, and empty on an already-converged database.

Review findings addressed (fixed in commit `5d4023f3`):
- **Behavioral correctness (major):** `dirtyBefore` over-exemption on post-marker resume — exemption is now derived from the shared `tablesToRekey` selector, so it can no longer drop a non-drifted table's edits.
- **Maintainability / readability:** retry-state branching complexity hotspot — extracted into the pure `tablesToRekey` predicate.
- **Test evidence:** missing regression test for the scoped exemption — added (`TestAuxRekeyExemptTablesScopedToRewrittenTables`).
- **Error handling (minor):** `readAuxRekeyState`'s `rows.Scan` error path did not degrade on an encoding-drift error like its `QueryContext` / `rows.Err()` siblings — made symmetric.
- **Documentation (non-gating):** clarified in-code, on `migrationWorkNeeded`, that convergence after a repaired drift is deliberately opportunistic so steady-state opens stay cheap.

Non-blocking follow-ups (shared for visibility — NOT completed in this PR):
- No non-blocking follow-ups: every finding raised in review was addressed in commit `5d4023f3`. For awareness, the review noted three inherent limitations of the underlying dolt#11131 drift that are already acknowledged in code and are not action items for this PR: (1) a broad `invalid hash length` match can classify a non-#11131 error as drift, which still yields an openable database; (2) "silent-garbage" drift (cells that decode to a zero payload without panicking) re-keys corrupt content, which is why storage repair is still required; and (3) whole-table retry re-keys rows minted since the skip, matching the existing crash-resume path. None are introduced by this PR.

### Maintainer Changes
Maintainer fix pushed directly to the PR branch, on top of your commits:
- `5d4023f3` fix(schema): scope aux-rekey dirtyBefore exemption to the rewritten tables — 5 files changed, +273 / -85.

Your original commits are preserved unchanged.

### Final Review Status
Ready for the merge queue: final head `5d4023f3` has passing CI (Migration Test Harness is green; the `main`-targeted suites re-run automatically under the merge queue's `merge_group` gate at merge time), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/33228941991

### Review Iterations
2 review passes performed. Iteration 1 requested changes — scope the resume exemption to the drifted tables only, flatten the retry-state complexity, and add a regression test. Iteration 2 verified the fix commit and approved with no new findings.

---

Issue #30, opened 2025-10-14
Windows Defender blocks bd for ransomware behaviour


maphew:
Zealous corporate group policy setting for Windows Defender's anti-ransomware can interpret bd's rapid creation of files within user profile as behaviour to block.

```pwsh
❯ bd
Program 'bd.exe' failed to run: Access is deniedAt line:1 char:1
+ bd
+ ~~.
At line:1 char:1
+ bd
+ ~~
    + CategoryInfo          : ResourceUnavailable: (:) [], ApplicationFailedException
    + FullyQualifiedErrorId : NativeCommandFailed
```

<img width="640" height="369" alt="Image" src="https://github.com/user-attachments/assets/540f6074-aac5-43dd-b3a6-78f4e5a13bda" />

modifying `export.go` to create a temporary file first and then rename seems to be enough to quiet Defender down and allow Beads to run.

_Diagnosis and remedy supplied by Grok Code Fast 1. A PR to follow_ 

---

The file creation pattern in export.go centers on line 64 (f, err := os.Create(output)), which uses Go's standard os.Create function to create or truncate a file at the user-specified output path. This occurs only when the --output flag is provided (non-empty string), as shown in the conditional block starting at line 61.

Key Aspects of the Implementation:
File Handling Logic:

- If no output file is specified, output defaults to os.Stdout (line 62).
- When an output file is specified, os.Create opens the file for writing, creating it if it doesn't exist or truncating it if it does (line 64).
- A deferred close is applied to ensure the file is properly closed after the function exits, even on errors (lines 69-73).
- The file handle is assigned to out (line 74), which is then used for JSON encoding.

Writing Process:

- A json.NewEncoder is created using the out writer (line 78).
- Issues are encoded one by one in a loop (lines 79-84), writing JSON Lines format to either stdout or the file.

**Potential Windows Defender Triggers:**

Windows Defender may flag this pattern due to:

- File Creation in User-Specified Paths: os.Create directly creates files in potentially sensitive locations (e.g., system directories or downloads folder), which antivirus software often monitors for unauthorized modifications.
 
- Immediate Writing After Creation: The file is created and immediately populated with data, mimicking patterns seen in malware that drops payloads.
 
- JSON Content: While JSONL is benign text data, Defender might heuristically scan for suspicious structures or large data writes, especially if the output path or content resembles known threats.
 
[...]

PR #12, opened 2025-10-14
Fix quickstart EXTENDING.md reference to use full URL

elimelec:
The quickstart command referenced EXTENDING.md as if it were a local file, which was confusing for users running bd in their own projects. Now it clearly points to the upstream documentation with a full GitHub URL.

- Changed from "See EXTENDING.md" to "See database extension docs"
- Added full URL to the original repo's EXTENDING.md
- Preserved the original "integration patterns" context

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5065, opened 2026-07-26
fix(github): include bead metadata in synced issue bodies

### LINKED ISSUE #4307: GitHub sync: embed bd ID in issue body + sync dependencies as tasklists
## Problem

`bd github sync` pushes title, body, status, and labels to GitHub Issues, but three critical pieces of structural data are lost:

1. **bd ID not embedded** — A GitHub issue has no reference back to its `bd_xxx-yyy` ID. Impossible to trace from GitHub back to the canonical bd ticket.
2. **Dependencies not synced** — Issues with dependency edges (blocks/blocked-by) in bd have none on GitHub. The dependency graph is completely flattened.
3. **Comments not synced** — bd comments do not appear on GitHub and vice versa.

### 1. Embed bd ID in GitHub issue body
Append a metadata footer to the GitHub issue body:
```markdown
---
bd_id: bd_0x2c-tkg7
```
Or as a simple last line: `<!-- bd: bd_0x2c-tkg7 -->`

This enables bidirectional tracing and lets tooling resolve a GH issue back to its bd source.

### 2. Sync dependencies as GitHub tasklists
Translate bd dependency edges into GitHub-native tasklist syntax in the issue body:
```markdown
## Dependencies
- [ ] #123 (blocks)
- [ ] #456 (blocked by)
```

This makes the dependency graph visible and navigable on GitHub without any external tooling.

### 3. Bidirectional comment sync (optional)
- Push bd comments to GH issue comments
- Pull GH issue comments back to bd

## Context

Testing with 690 issues on a private repo. 38 issues have dependency edges. Current sync is a flat mirror — useful for visibility but loses the structural richness that makes bd valuable.

Running `bd version 1.0.0 (dev: feat/agentic-action-research@d8026878692d)`.

halaprix:
## Summary
- render a managed GitHub issue-body block with the source bead ID marker
- surface dependency and dependent edges as GitHub tasklists
- strip the generated block on pull so local bead descriptions stay clean

## Scope
- Closes #4307
- Comment/thread mirroring is intentionally out of scope per the integration charter.
- Adjacent PR #4329 is about push dedup; this PR changes the rendered body and preserves pre-rendered bodies so dependency tasklists are not dropped by create/update mapping.

## Test Plan
- `go test -tags gms_pure_go ./internal/github -count=1`
- `go test -tags gms_pure_go ./cmd/bd -run 'Test.*GitHub|TestGetGitHub|TestParseGitHub|TestGitHubIssue' -count=1`
- `make test`
- `make build`

## Notes
- `scripts/pr-preflight.sh --search "github sync bd id dependency tasklist 4307" --repo gastownhall/beads` could not complete in this environment because `gh` resolves to a non-GitHub wrapper. I used GitHub REST search instead and found no open duplicate PR for #4307 or the marker/tasklist scope.

### DIFF
--- a/cmd/bd/github.go
+++ b/cmd/bd/github.go
@@ -419,8 +419,9 @@ func runGitHubSync(cmd *cobra.Command, args []string) error {
 	engine.OnMessage = func(msg string) { _, _ = fmt.Fprintln(out, "  "+msg) }
 	engine.OnWarning = func(msg string) { _, _ = fmt.Fprintf(os.Stderr, "Warning: %s\n", msg) }
 
-	// Set up GitHub-specific pull hooks
+	// Set up GitHub-specific pull and push hooks
 	engine.PullHooks = buildGitHubPullHooks(ctx)
+	engine.PushHooks = buildGitHubPushHooks(ctx)
 
 	// Build sync options from CLI flags
 	pull := !githubSyncPushOnly
@@ -496,6 +497,22 @@ func buildGitHubPullHooks(ctx context.Context) *tracker.PullHooks {
 				issue.ID = generateIssueID(prefix)
 			}
 			return nil
+		},
+	}
+}
+
+func buildGitHubPushHooks(ctx context.Context) *tracker.PushHooks {
+	return &tracker.PushHooks{
+		FormatDescription: func(issue *types.Issue) string {
+			dependencies, depErr := store.GetDependenciesWithMetadata(ctx, issue.ID)
+			dependents, dependentErr := store.GetDependentsWithMetadata(ctx, issue.ID)
+			if depErr != nil {
+				dependencies = nil
+			}
+			if dependentErr != nil {
+				dependents = nil
+			}
+			return github.RenderGitHubIssueBody(issue, dependencies, dependents)
 		},
 	}
 }
--- a/internal/github/mapping.go
+++ b/internal/github/mapping.go
@@ -120,8 +120,9 @@ func GitHubIssueToBeads(gh *Issue, config *MappingConfig) *IssueConversion {
 	labelNames := gh.LabelNames()
 
 	issue := &types.Issue{
+		ID:           ExtractBDIDFromGitHubSyncBlock(gh.Body),
 		Title:        gh.Title,
-		Description:  gh.Body,
+		Description:  StripGitHubSyncBlock(gh.Body),
 		ExternalRef:  &htmlURL,
 		SourceSystem: sourceSystem,
 		IssueType:    types.IssueType(typeFromLabels(labelNames, config)),
@@ -151,9 +152,13 @@ func GitHubIssueToBeads(gh *Issue, config *MappingConfig) *IssueConversion {
 
 // BeadsIssueToGitHubFields converts a beads Issue to GitHub API update fields.
 func BeadsIssueToGitHubFields(issue *types.Issue, config *MappingConfig) map[string]interface{} {
+	body := issue.Description
+	if !strings.Contains(body, "<!-- bd-github-sync: start -->") {
+		body = RenderGitHubIssueBody(issue, nil, nil)
+	}
 	fields := map[string]interface{}{
 		"title": issue.Title,
-		"body":  issue.Description,
+		"body":  body,
 	}
 
 	// Build labels from type, priority, and status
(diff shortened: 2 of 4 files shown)


maphew (commented):
Thanks for this, and apologies for the silence — this sat unanswered on our side, which is a queue-rhythm failure, not a comment on the work. The feature is one we want: closing #4307 by making a synced GitHub issue carry its bead identity and edges is the right shape, and the plumbing fits existing patterns (the push hook, the create-path double-render guard, `tracker.go` sending `fields["body"]` so the hook's output isn't re-rendered).

I ran two independent reviews on different model families before writing this, because the change turns on a round-trip invariant and those are easy to get subtly wrong. Both reviews independently landed on the same two blockers, so I'm confident they're real rather than reviewer noise. One of them fetched `refs/pull/5065/head` and ran an actual probe; the transcripts below are measured, not hypothesized.

## Two blockers

**1. Strip is not the exact inverse of render, so an unpaired start marker deletes user text.**

`StripGitHubSyncBlock` (`internal/github/mapping.go:333`) removes from the *first* start marker to the first end marker, while `RenderGitHubIssueBody` (`:202`) always appends a fresh block at the *end*. If a body contains a start marker with no matching end, the next cycle spans that stray marker across everything between it and the generated block's end marker, and deletes it. Measured against your branch:

```
pull1 local:
    base

    <!-- ...start -->
    ## Beads
    Source: `bd-1`

    MY IMPORTANT NOTE

push  body:  ... two start markers, one end ...

pull2 local:
    base
          <- MY IMPORTANT NOTE permanently gone
```

This is reachable without anything adversarial: HTML comments are invisible in rendered Markdown, so someone tidying the visible `## Beads` section in the web editor can delete the closing comment without seeing it. The same mechanism fires on a bead whose description *legitimately* contains a well-formed block — a bead documenting this feature, or a support ticket quoting a synced body. `RenderGitHubIssueBody:206` strips it from the base before appending, the remote loses it, and the next pull writes the stripped version back over the local description.

A related case: dependency titles are interpolated verbatim (`:240-258`), so a dependency titled `x <!-- bd-github-sync: end --> y` terminates the strip early and leaks the remainder into the local description, growing the body on every round trip.

The fix that closes all three at once: anchor the strip to a single **well-formed, trailing** block with line-anchored markers (render only ever emits one, at the end), have `RenderGitHubIssueBody` refuse to append over an unpaired start, and escape or reject markers in interpolated titles.

**2. `ExtractBDIDFromGitHubSyncBlock` reads outside the block, so a GitHub issue can bind to the wrong bead.**

Despite the name, `mapping.go:351` searches the whole body for the first `<!-- bd:`, with no block requirement and no `sync=github-v1` requirement. Probe result on a body where user text mentions a marker before the real one:

```
extracted = "bd-victim"   (want "bd-real")
```

That value flows through `mapping.go:123` to `conv.Issue.ID`, and `internal/tracker/engine.go:434-437` then resolves `existing = localByID[localID]`. The unrelated local bead is updated with this issue's title, description, status and labels, **and has its `external_ref` repointed** to the imposter issue. The realistic trigger is mundane: a contributor copy-pastes a synced issue body into a new issue. On the create path, `GenerateID` only fills an empty ID (`cmd/bd/github.go:493`), so an arbitrary marker string can become a new bead's primary key with a foreign prefix and no validation.

Worth calling out that this makes pull change *issue identity matching*, which is the most consequential behaviour change in the PR and isn't mentioned in the description or in a code comment.

Fix: extract only from inside a paired block, require the `sync=github-v1` token, validate the ID against the local prefix, and refuse to adopt a local bead by ID when it already carries a different `external_ref`.

## Should-fix, not blocking

- **Dependency-query errors are swallowed** (`cmd/bd/github.go`, `if depErr != nil { dependencies = nil }`). A transient storage error pushes a block with the dependency sections silently *deleted*. The GitLab path warns on stderr (`cmd/bd/gitlab.go:733-736`); at minimum match that, better to skip the re-render and leave the previous body intact.
- **Push-dirtiness doesn't consider the rendered body.** The hook supplies `FormatDescription` but no `ContentEqual`, and the engine skips updates on timestamps. So adding or removing a dependency doesn't advance the issue timestamp and the changed tasklist may never reach GitHub; and if dependency B gets its external ref after A was rendered, A can stay permanently linked as `` `bd-B` `` rather than `#N`.
- **`strings.TrimSpace` on both sides** means `Strip(Render(D)) == TrimSpace(D)`, not `D`. Pull normalizes marker-free bodies byte-for-byte: a body of four leading spaces plus `code` plus a trailing CRLF comes back as bare `code`, which changes Markdown semantics for indented code.
- **The rendered lines are probably not GitHub tasklists.** `- [ ] #12 — blocking issue (`bd-blocker`, blocked by)` — GitHub only creates a tracked relationship when the item is a bare issue reference, so the trailing text makes these plain checkboxes. Closed dependencies also render `- [ ]`, so every synced issue gains a permanently-incomplete "0 of N tasks" badge, and a user who ticks a box has it reverted on the next push with no "generated, do not edit" notice.
- **`TestExtractBDIDFromGitHubSyncBlockSupportsLegacyMarker`** asserts that a bare `<!-- bd: ... -->` outside any block must parse. I grepped for `<!-- bd:` and `bd-github-sync` across the tree and found no prior convention — there is no legacy format to be compatible with, and this test is what pins the over-lenient parser behind blocker 2. It should go unless you can point at a real one.
- **Tests don't cover round-trip.** They assert that render output *contains* expected strings and that one canonical block strips cleanly; there's no `render → strip → compare` assertion and none of the adversarial cases. Double-push idempotency and interior CRLF do hold in practice (verified), but nothing pins them. The invariants worth asserting are `Strip(Render(D)) == D` and `Render(Strip(Render(D))) == Render(D)`, across unpaired/nested/truncated markers, markers in user text, hostile dependency titles, CRLF, leading whitespace, and empty descriptions.

## Coordination note

This collides with #4329 (push dedup), which you mention but the two aren't reconciled. That PR's `PushFieldsEqual` compares `local.Description != remote.Body` and `PushContentHash` hashes `local.Description`. After this PR, local descriptions are stripped and remote bodies carry the block, so the comparison is permanently false and the stored hash never matches what was pushed — the dedup is defeated and every sync re-PATCHes every issue. Whichever lands second needs to compare `RenderGitHubIssueBody(...)` output, the way Linear does at `internal/linear/mapping.go:587` with `BuildLinearDescription`. I'll cross-note this on #4329.

Also a release note we should write rather than something for you to fix: the first sync after this lands rewrites every previously-synced issue body, which is one-time notification churn for existing users. Pull on pre-block issues is safe (no marker, empty ID, normal path).

## Where I'd like to take this

The two blockers are surgical — anchored trailing-block framing plus a scoped extractor — and I'd rather not spend a review round trip on them. **My offer: I'll implement blockers 1 and 2 plus the round-trip test table maintainer-side, on top of your commits with your authorship preserved**, and open the result crediting you as the author of the feature. The tasklist-syntax question (item 7) becomes a follow-up issue rather than a gate.

If you'd rather do it yourself, that's genuinely welcome too — say so and I'll leave it with you and review promptly this time. Either way the design here is sound and the feature is going in; only tell me which path you prefer.

jms830:
Maintainer-side evidence on review item 7 (the lenient-marker compatibility test), since it's the one open question here that outside data can settle.

You wrote that you grepped for `<!-- bd:` and `bd-github-sync` across the tree, found no prior convention, and that the bare-marker test should go **unless someone can point at a real one**. There is a real one, and it isn't in the tree — it's in private repos, which is why the grep came back empty.

## Measured deployment

On one private repo (`bd` 1.1.2, GitHub integration in pointer-only mode — `external_ref` set, `bd github sync` deliberately never enabled):

- **21 open issues, 22 markers**, every one hand-placed as a bare trailing line.
- Form is byte-identical to this PR's grep target: `<!-- bd: br-0do -->`, `<!-- bd: br-x2a.1 -->`, and so on.
- No wrapping block, no `sync=github-v1` token — a bare HTML comment as the last line of the body.

The form isn't invented. It's the one proposed in #4307 itself:

> Or as a simple last line: `<!-- bd: bd_0x2c-tkg7 -->`

People read that issue and did it by hand while waiting for the feature. So the "legacy format" is a published suggestion that got adopted, not an accident. Expect other pointer-only installs to look the same.

## Recommendation: ship your strict fix anyway

Being one of the affected installs, our preference is still **strict** — paired block plus token, bare markers ignored.

Under a first-match extractor our bare footers become authoritative input to a parser that was never told about them. Under strict extraction they're inert prose, the `external_ref` path keeps working exactly as it does today, and the failure mode is "no ID found" rather than "wrong ID found". We'd rather have the former.

The one thing worth adding is a release note along the lines of *hand-placed `<!-- bd: … -->` footers are treated as ordinary body text, not as sync metadata* — because the format was published in #4307, and installs that copied it will otherwise assume it's understood.

Worth flagging separately: the first sync after this lands rewrites every previously-synced issue body, as you noted. For a pointer-only install whose markers live only on the GitHub side and never in the bead description, `strip(description) + block` drops them. That's survivable once it's expected, but it is silent, and it's another reason the release note matters.

## A fixture you can't synthesise

For the round-trip test table: one of our issues (`#30`) legitimately carries **two** bare markers, in this order:

```
<!-- bd: br-zbb -->
<!-- bd: br-zbb.2 -->
```

Both beads set `external_ref` to that same issue — parent track and child track, deliberately. It's the natural shape when an issue's scope spans a rollup and one of its children.

This breaks any "at most one marker" assumption in a way a synthetic fixture wouldn't:

- **First-match extraction** returns `br-zbb`. Correct here — but only because the parent happens to be written first. Reverse the two lines and the same code binds the child.
- **Strict extraction** returns empty, both fall through to `external_ref`, and both stay correct regardless of order.

Suggested cases, all cheap:

1. Bare marker present → not parsed as an ID, and preserved verbatim through render → pull → render.
2. **Two** bare markers → no ID extracted; specifically not "the first one".
3. Marker-shaped text inside ordinary prose or a fenced code block → untouched.
4. Paired block with token → extracted, and round-trips byte-stable.
5. Bare marker *plus* a valid paired block → the block wins irrespective of document order.

Case 2 is the one worth having; it's the case where a lenient parser is confidently wrong rather than merely unhelpful.

Happy to re-run any of this against the real install if a specific measurement would help. No opinion on the tasklist-syntax item.

maphew:
#5642 now carries this PR's commit (rebased across #4329/#4214, authorship intact) plus the two review blockers from the 2026-07-28 review, implemented as offered after the 2026-08-04 reply-by date passed. @jms830 your bare-marker evidence settled the strict-extraction question - those markers stay inert. @halaprix if you'd rather pick this back up yourself, say the word and #5642 closes in favor of your branch.

PR #5066, opened 2026-07-26
fix(update): refuse overwriting a live foreign claim without --force (bd-98s5c)

steveyegge:
## Summary

Plain `bd update -a <new>` and its shorthand `bd assign` silently overwrote another actor's live `in_progress` claim — rc=0, no warning, no audit trail. It was the **last unfenced cross-actor takeover path** in the CLI: `--claim`, `unclaim`, and `close` all already refuse without `--force` (bd-at6rc, be-035), so an agent that had correctly learned not to steamroll still had one verb that would do it without complaint. Filed by ganon from wyvern wy-8uaqw (two real Sol cross-actor steamrolls on 2026-07-16).

## Design

New `validation.AssigneeNotStolen` validator, wired at the cmd layer as the exact sibling of the be-035 close fence, covering all four surfaces: direct + proxied-server `update -a`, direct + proxied `assign`. Every predicate clause is load-bearing:

- `assignee != ""` — unassigned beads stay freely assignable
- `assignee != actor` — self-edits untouched
- `assignee != newAssignee` — idempotent re-assert stays a success (retry/replay safety)
- `status == in_progress` — **open-bead dispatch stays frictionless** (dispatchers hand-dole open beads constantly; this is why the unclaim/close ownership rule can't be reused verbatim)
- holder not a `claim.pools` alias — the same carve-out `--claim` has, via a new shared `issueops.ParseClaimPools` used by both claim duals and the fence, so `--claim` and `-a` can never give opposite answers on the same bead

Additional decisions (both recorded on the bead):
- **`-a ""` (unassign) of a live foreign claim is fenced** — it's unclaim in disguise.
- **The fence skips under an explicit `--if-assignee` guard.** `--force` and `--if-assignee` are mutually exclusive (mirroring unclaim), so without the skip the sanctioned bd-wsqvw park transition (`--if-assignee holder -a new`) would become impossible. Nothing silent about a CAS that names the holder; staleness still exits 13.
- Refusal is a **policy verdict: exit 1, never 13** — coordinating is the fix, not retrying. Copy names the holder, points at `bd mail`, names `--force` explicitly (deterministic for script authors), and never teaches the unclaim steamroller (wy-zs5s2 item 2).
- `bd batch` remains a raw-transaction surface that bypasses cmd-layer fences — pre-existing (it bypasses the close fence too), scoped out.

CLI reference docs are pinned to v1.1.0 (`docs/cli-docs.pin`), so the new flag help flows into docs at the next release pin bump — no docs change in this PR by design.

## Tests

- `TestAssigneeNotStolen` unit table: all clauses, pool carve-out, lazy pool-thunk contract, refusal-copy fragments.
- `TestReassignFenceCLI` (embedded): refuse/--force on update and assign, unassign fence, and the negative regression fences — open-bead reassign, self-edit, idempotent re-assert, guarded park, flag mutual exclusion, pool-alias take. **Passed** (273s).
- `TestProxiedServerReassignFence`: refusal exits 1 (not 13) on the shared-server topology where collisions actually happen, --force override, replay re-assert, open dispatch. **Passed**.
- Interacting suites re-run green: `TestUpdateIfGuardsCLI`, `TestEmbeddedClaimPools`, `TestUnclaimConditional`.
- Full default cmd/bd sweep: 769 pass; the one failure (`TestCreateDepsAtomicity`, create --waits-for-gate territory) is untouched by this diff and being verified pre-existing on main.

## Downstream

wyvern wy-er4ch: `wh-take.sh`'s --steal arm adopts `--force`; its queue arm can register `fable-crew` in `claim.pools` (no flags needed) or use `--if-assignee fable-crew`.

### DIFF
--- a/cmd/bd/assign.go
+++ b/cmd/bd/assign.go
@@ -17,6 +17,11 @@ var assignCmd = &cobra.Command{
 
 Shorthand for 'bd update <id> --assignee <name>'.
 
+Refuses to overwrite another actor's live in_progress claim without --force
+(bd-98s5c); issues assigned to a claim.pools alias are exempt, matching
+--claim. For a holder-aware transfer prefer
+'bd update <id> --if-assignee <holder> -a <new>'.
+
 Examples:
   bd assign bd-123 alice
   bd assign bd-123 ""      # unassign`,
@@ -33,8 +38,10 @@ Examples:
 			}
 		}()
 
+		force, _ := cmd.Flags().GetBool("force")
+
 		if usesProxiedServer() {
-			return runAssignProxiedServer(rootCtx, args)
+			return runAssignProxiedServer(rootCtx, args, force)
 		}
 
 		id := args[0]
@@ -60,6 +67,13 @@ Examples:
 		issueStore := result.Store
 
 		if err := validateIssueUpdatable(id, result.Issue); err != nil {
+			return HandleErrorRespectJSON("%s", err)
+		}
+
+		// bd-98s5c: bd assign is shorthand for an unguarded assignee update —
+		// same live-claim fence as bd update -a.
+		if err := validateIssueReassignable(id, result.Issue, actor, assignee,
+			storeClaimPoolAliases(ctx, issueStore), force); err != nil {
 			return HandleErrorRespectJSON("%s", err)
 		}
 
@@ -102,6 +116,7 @@ Examples:
 }
 
 func init() {
+	assignCmd.Flags().Bool("force", false, "Allow overwriting another actor's live in_progress claim (use only for abandoned claims — crashed agent, expired lease; prefer bd reclaim)")
 	assignCmd.ValidArgsFunction = issueIDCompletion
 	rootCmd.AddCommand(assignCmd)
 }
--- a/cmd/bd/assign_fence_embedded_test.go
+++ b/cmd/bd/assign_fence_embedded_test.go
@@ -180,6 +180,21 @@ func TestReassignFenceCLI(t *testing.T) {
 		}
 	})
 
+	t.Run("claim_conflict_copy_preserved_under_claim_flag", func(t *testing.T) {
+		// --claim -a X on a foreign live claim must keep failing through the
+		// claim CAS with the canonical "already claimed" copy — the fence
+		// defers to it (the CAS is itself the anti-steal gate, and existing
+		// automation keys on that message).
+		issue := claimAs(t, "holder")
+		out, _ := bdUpdateFailCode(t, bd, dir, issue.ID, "--actor", "thief", "--claim", "--assignee", "thief")
+		if !strings.Contains(out, "already claimed") && !strings.Contains(out, "already assigned") {
+			t.Errorf("expected the claim CAS conflict copy, got:\n%s", out)
+		}
+		if got := bdShow(t, bd, dir, issue.ID); got.Assignee != "holder" {
+			t.Errorf("claim+assignee combo clobbered the row: assignee=%q, want holder", got.Assignee)
+		}
+	})
+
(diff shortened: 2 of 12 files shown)


maphew:
Reviewed 2026-07-26 (head 50aa98027). **Approve-with-nits** — the predicate is well-chosen, each exemption is justified (I checked the `--claim` skip against `ClaimIssueInTx`'s CAS: a foreign live claim can't win it, so no hole), coverage is genuinely good across all four surfaces, CI green. Two small fixes before merge:

1. **`--force` is silently inert with `--claim`** (cmd/bd/update.go:811 + :414). The refusal copy tells users to "pass --force if the claim is abandoned", so the natural next command is `--claim -a me --force` — which skips the fence under claimFlag and fails with the unrelated "already claimed" copy. Either `MarkFlagsMutuallyExclusive("force", "claim")` or amend the `--force` help to say it doesn't apply to `--claim`.
2. **Refusal copy suggests an invalid command** — `bd mail <holder>` isn't a valid invocation (`bd mail` is a pass-through delegate with forms like `bd mail send <name>/ -s ...`, and errors when no delegate is configured). Drop the parenthetical or write `bd mail send <holder>/`.

Follow-up-bead material, not blockers: (a) the two-step bypass — `--status open` (unfenced) then `-a thief` — still silently succeeds, so the body's "last unfenced cross-actor takeover path" claim should either scope that out explicitly or fence status-away-from-in_progress on a live foreign claim; (b) `bd graph apply` (graph_apply.go:1051) writes assignee through a bulk map unfenced and is unmentioned alongside `bd batch`; (c) the `pool_alias_holder_is_takeable` subtest sets `claim.pools` on the shared dir without `t.Cleanup`; (d) no CHANGELOG entry for two core verbs newly exiting 1. Sequencing note: this should land before #4697/#4715 — smaller and self-contained, and #4697's `--if-fence` will need the same skip treatment as `--if-assignee`.

steveyegge:

Reviewed the full diff. The fence is correctly scoped and every predicate clause earns its place:

- The five-clause refusal condition is right, and the ordering matters: the lazy pool thunk is only consulted after the cheap clauses pass, and the laziness is contract-tested (`poolsCalled` assertion), so routine dispatch never pays the config read.
- All four surfaces are covered, and the proxied-server fence runs **inside** the UoW mutate closure — the direct path's check-then-write follows the existing close-fence pattern (same TOCTOU window as be-035; pre-existing architecture, not this PR's problem — the proxied path, where real collisions happen, is transactional).
- The two skips are the hard design calls and both are correct: `--if-assignee` names the holder (nothing silent; keeps the bd-wsqvw park expressible given the flag exclusion), and `--claim` defers to the claim CAS's canonical copy — with the CI-caught interaction now regression-fenced (`claim_conflict_copy_preserved_under_claim_flag`).
- `ParseClaimPools` shared parser kills the --claim/-a drift class structurally rather than by convention. Fail-closed on config read error is the right default for a fence.
- Fencing `-a ""` as unclaim-in-disguise closes the obvious workaround. Exit 1 (policy, never 13) is right — a steamroll refusal is not a retryable guard mismatch.
- Refusal copy: names holder, points at coordination, names `--force` deterministically, and the no-eviction-recipe rule is itself asserted in tests (wy-zs5s2 item 2 honored).

Notes, non-blocking:
1. `bd batch` bypassing cmd-layer fences is acknowledged pre-existing scope-out — fine here, but the fence family (close + reassign) is now big enough that a batch-surface parity bead is worth having on the books if one doesn't exist.
2. TestCreateDepsAtomicity flake claimed pre-existing on main — please confirm that verification landed before or alongside merge so the sweep stays a clean signal.
3. Downstream wy-er4ch (wh-take.sh --steal arm adopts --force) — good catch; wyvern side should land close behind the release that carries this.

Ship it.

PR #5067, opened 2026-07-26
fix(test): prebuild bd for subprocess tests; fit cmd/bd inside the package deadline (wy-4mtr0)

steveyegge:
## Problem (wyvern wy-4mtr0, filed by navi)

`./scripts/test.sh` with no flags could never produce a green full-suite receipt on a busy darwin fleet machine: cmd/bd tripped the 3m per-package go test deadline — `panic: test timed out after 3m0s` with one test running at 0s, i.e. the package alarm, not a hang. Acceptance lines were getting hand-waved because no green receipt was possible.

## Measurements (2026-07-26, the reporting machine class)

- cmd/bd's default suite: **~1090s of test time across ~1486 tests** — mostly serial, since the heavy prime/metrics/init families are `t.Setenv`-bound and cannot `t.Parallel`.
- Wall clock was dominated by **twelve separate in-test `go build` sites**, each linking the full cgo bd binary (top offender `TestClose_UpdatesLastTouched` = 142s of which nearly all was the once-build; `TestCreateDepsAtomicity` dropped 126s → 27s once prebuilt).
- CI never sees this: main.yml prebuilds bd and exports `BEADS_TEST_BD_BINARY` — but only 5 of the 12 builder sites honored it, and the local runner never got the prebuild at all.

## Fix — no naked timeout bump

1. **scripts/test.sh prebuilds bd once** and exports `BEADS_TEST_BD_BINARY` (same fast path as CI). Binary lands in the test-env root so cleanup is inherited; a caller-supplied value always wins; skipped when the requested packages can't include cmd/bd.
2. **All twelve in-test builder sites now honor `BEADS_TEST_BD_BINARY`** (six built unconditionally before). The **repo-root `./bd` reuse is removed** from `buildBDForInitTests` and `cli_fast_test`: it silently ran subprocess tests against whatever stale binary sat in the checkout root — a two-day-old one produced phantom `TestCreateDepsAtomicity` failures during this investigation (`--waits-for-gate` silently ignored by a binary predating the feature). Tests must exercise the checkout's source or an explicitly supplied binary, never an incidental artifact.
3. **Default TIMEOUT 3m → 25m, justified in the comment**: the deadline is a hang backstop, not a performance budget — it costs nothing while packages pass. cmd/bd measured 870s under `-p 4` in the passing acceptance run; 25m holds that plus fleet-load headroom.

## Acceptance (the bead's criterion, verified)

`./scripts/test.sh` with NO extra flags on the busy darwin machine class: **exit 0, 71 packages ok, zero failures** (cmd/bd 870s). One contention flake was chased down and cleared: `TestPrime_MemoriesOnly_WithCustomPrimeMd` has a 30s per-subprocess deadline that fired under deliberate parallel load; it passes 3/3 isolated.

## Follow-up (out of scope, noted on the bead)

The deep runtime cut — sharing the per-test embedded-Dolt bring-up and unwinding the `t.Setenv` serialization of the subprocess families — interacts with wy-s4zdj (leaked test dolt servers on darwin) and remains open.

### DIFF
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -705,19 +705,20 @@ func init() {
 	if os.Getenv(worktreeRemoveHelperEnv) != "" {
 		return
 	}
-	// Use existing bd binary from repo root if available, otherwise build once
+	// Prebuilt fast path (scripts/test.sh and CI export this), else build
+	// once. No repo-root ./bd reuse: a stale checkout-root binary silently
+	// substitutes itself for the source under test (wy-4mtr0).
+	if prebuilt := os.Getenv("BEADS_TEST_BD_BINARY"); prebuilt != "" {
+		if abs, err := filepath.Abs(prebuilt); err == nil {
+			if _, statErr := os.Stat(abs); statErr == nil {
+				testBD = abs
+				return
+			}
+		}
+	}
 	bdBinary := "bd"
 	if runtime.GOOS == "windows" {
 		bdBinary = "bd.exe"
-	}
-
-	// Check if bd binary exists in repo root (../../bd from cmd/bd/)
-	repoRoot := filepath.Join("..", "..")
-	existingBD := filepath.Join(repoRoot, bdBinary)
-	if _, err := os.Stat(existingBD); err == nil {
-		// Use existing binary
-		testBD, _ = filepath.Abs(existingBD)
-		return
 	}
 
 	// Fall back to building once (for CI or fresh checkouts)
--- a/cmd/bd/context_binding_integration_test.go
+++ b/cmd/bd/context_binding_integration_test.go
@@ -103,17 +103,10 @@ func TestListExplicitDBPathRebindsTargetContext(t *testing.T) {
 		t.Fatalf("create issue: %v", err)
 	}
 
-	binPath := filepath.Join(t.TempDir(), "bd-under-test")
-	packageDir, err := os.Getwd()
-	if err != nil {
-		t.Fatalf("getwd: %v", err)
-	}
-	buildCmd := exec.Command("go", "build", "-o", binPath, ".")
-	buildCmd.Dir = packageDir
-	buildOut, err := buildCmd.CombinedOutput()
-	if err != nil {
-		t.Fatalf("go build failed: %v\n%s", err, buildOut)
-	}
+	// Shared once-per-process binary (honors BEADS_TEST_BD_BINARY) instead of
+	// a per-test go build — the in-test link steps dominated cmd/bd's wall
+	// clock (wy-4mtr0).
+	binPath := buildBDForInitTests(t)
 
 	listCmd := exec.Command(binPath, "list", "--db", filepath.Join(targetBeadsDir, "dolt"), "--json")
 	listCmd.Dir = callerRepo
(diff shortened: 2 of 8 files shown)


maphew:
Reviewed 2026-07-26 (head 4f9e1405a). **Approve-with-nits** — right shape, net-deleting, and I verified the flag alignment (test.sh sources .buildflags so the prebuild gets `-tags=gms_pure_go` + CGO), the EXIT-trap cleanup, the coherence of removing `isCGOBuild`, and that all 12 builder sites honor the env var. One fix needed before merge:

1. **Windows breakage** (scripts/test.sh, prebuild block ~line 105): the output is hardcoded `"$PREBUILT_BD_DIR/bd"` with no `.exe`. Under Git Bash, `exec.LookPath` on Windows resolves through PATHEXT and won't find an extension-less `bd`, and since `findPrebuiltBDBinary`'s `os.Stat` *succeeds*, the fallback never triggers — every subprocess test fails outright. Fix: `PREBUILT_BD_BIN="$PREBUILT_BD_DIR/bd$(go env GOEXE)"` for both `-o` and the export.

Also worth doing in the same pass: engdocs/TESTING.md:44,269 and engdocs/CI_TEST_SURFACE_AUDIT.md:78 still document the 3m default this PR raises to 25m (doc-freshness CI doesn't cover these). Smaller polish, your call: the `mktemp -d` fallback when BEADS_TEST_ENV_ROOT is unset leaks a ~100MB dir per run (note the later shared-server block clobbers a naive EXIT trap); cli_fast_test.go's prebuilt path silently falls through on an unstat-able BEADS_TEST_BD_BINARY where the other eleven helpers hard-error; and the init() header comment still references the repo-root probe this PR deletes.

steveyegge:

The valuable part of this PR isn't the speedup, it's the correctness fix hiding inside it: the repo-root `./bd` reuse meant subprocess tests silently exercised whatever stale artifact sat in the checkout — 'tests must exercise the checkout's source or an explicitly supplied binary, never an incidental artifact' is exactly the right rule, and the new comment on `buildBDForInitTests` preserves the why. Bonus: this also closes the loop on #5066's sweep — the TestCreateDepsAtomicity 'pre-existing failure' there is now explained as this exact stale-binary phantom, which upgrades that sweep from 'one unexplained red' to clean.

Also right:
- `BEADS_TEST_BD_BINARY` as the single fast path, caller-supplied always winning, matching what CI already does — one mechanism, two consumers.
- 25m as a documented hang-backstop (with the 1090s/870s measurements inline) rather than a magic number; per-package deadlines cost nothing while green.
- scripttest only adopts the prebuilt when the basename is exactly `bd` — correct, since the scripts invoke plain `bd` through `sh -c`.
- The skip-guard so non-cmd/bd package selections don't pay the prebuild.

Two non-blocking notes:
1. test.sh's prebuild is a plain `go build` — no `-tags gms_pure_go` and no CGO check (the old `isCGOBuild` guard died with the repo-root path). On a normal dev box CGO is on and the binary has embedded Dolt, but an explicit `CGO_ENABLED=0 ./scripts/test.sh` would now hand init-style tests a binary with no embedded Dolt and no pure-go fallback tag. Cheap insurance: build the prebuilt with `-tags gms_pure_go`, matching what the in-test fallback builds use.
2. The mktemp'd prebuilt dir is never cleaned when BEADS_TEST_ENV_ROOT is unset — tmp litter per run, harmless but worth a trap-rm whenever test.sh is next touched.

Acceptance verified as claimed (no-flags run green, 71 packages, cmd/bd 870s). Ship it.

PR #5074, opened 2026-07-26
test: use native repro timeout fixtures

### LINKED ISSUE #5071: Native-Windows tests in repro-dolt-prod-timeouts use POSIX-only bd fixtures
# Native-Windows tests in repro-dolt-prod-timeouts use POSIX-only bd fixtures

## What happens

Two tests in `scripts/repro-dolt-prod-timeouts` create an extensionless file
named `bd` whose contents begin with a POSIX `#!/bin/sh` shebang:

- `TestResolveExecutablePathReturnsAbsolutePath`
- `TestBenchmarkSubprocessesStripDoltEnvOverrides`

That is executable on POSIX hosts, but native Windows neither resolves the
extensionless fixture through `exec.LookPath("./bd")` nor dispatches its
shebang through `CreateProcess`.

On current `main` (`d01d62eb21407fca3a16d95549fa3312280e7925`):

```text
=== RUN   TestResolveExecutablePathReturnsAbsolutePath
    main_test.go:75: find bd binary "./bd": exec: "./bd": executable file not found in %PATH%
--- FAIL: TestResolveExecutablePathReturnsAbsolutePath
=== RUN   TestBenchmarkSubprocessesStripDoltEnvOverrides
    main_test.go:126: runBD inherited denied env: err="exec: \"...\\bd\": executable file not found in %PATH%" stderr=""
--- FAIL: TestBenchmarkSubprocessesStripDoltEnvOverrides
```

The adjacent `TestControlQueryScriptPreservesReadyProbeFailure` uses the
intentional shell path and passes under Git Bash; this report is limited to the
two fixtures that production code executes directly.

## Expected behavior

The tests should use a platform-native executable fixture while preserving the
production `resolveExecutablePath` and `runBD` paths unchanged. A copied Go test
executable in narrowly bound self-reexec mode can act as `bd`/`bd.exe` on every
host and can retain the denied-environment assertion without adding a platform
branch to production code.

The helper mode should be bound so an ambient environment value cannot make the
ordinary test runner exit successfully without running its tests.

ecuthiell:
# test: use native repro timeout fixtures

Fixes #5071

## What

- Replace the two extensionless shebang-based fake `bd` files with a copied
  native Go test executable named `bd` or `bd.exe`.
- Give that executable a narrowly scoped self-reexec mode for the benchmark
  environment assertion.
- Bind helper activation to the expected executable name and live parent PID,
  and inject its control variables only into the intended child.
- Leave production executable resolution and benchmark behavior unchanged.

## Why

Native Windows cannot execute an extensionless POSIX shebang fixture through
`exec.LookPath`/`CreateProcess`, so two otherwise portable tests fail before
reaching their assertions.

## Validation

- Focused current-main failure reproduced on Windows.
- Full `scripts/repro-dolt-prod-timeouts` package passed on Windows (19 passed,
  one unrelated Dolt smoke skipped because Dolt is not installed).
- Race detector passed.
- `go vet ./scripts/repro-dolt-prod-timeouts`
- `golangci-lint run ./scripts/repro-dolt-prod-timeouts` (zero issues)
- Linux/amd64 CGO-disabled test binary compiled successfully.
- The unchanged exact head ran under WSL2's Linux kernel from native ext4 with
  the SHA-256-pinned official Go 1.26.5 linux/amd64 toolchain: exactly the
  three changed fixture tests passed, and the full package passed 19 tests
  with the one known Dolt-not-installed skip.
- `gofmt` and `git diff --check`

The exact-head `PR Core (wrapper timing)` and `CI Gate / Required` checks
remain mandatory ordinary regression and aggregate gates.

### DIFF
--- a/scripts/repro-dolt-prod-timeouts/main_test.go
+++ b/scripts/repro-dolt-prod-timeouts/main_test.go
@@ -4,12 +4,15 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"runtime"
 	"slices"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


maphew (approved):
Replacing the POSIX-shebang fake `bd` with a copied native test binary is the right portability fix, and the helper-activation guard is properly paranoid: mode env var AND executable basename AND live parent-PID binding, so an ambient env value can't turn an ordinary test run into a silent helper. The wrong-parent regression test closes the loop. Approving; the patrol will merge on green.

PR #5075, opened 2026-07-26
ci: lint Windows-tagged Go code

### LINKED ISSUE #4991: PR lint skips Windows-tagged Go files
## Summary

The required PR lint wrapper runs `golangci-lint` only for the runner's host Go target. The required lint jobs run on Ubuntu, so files selected only for Windows are never loaded and Windows-only lint regressions can merge unnoticed.

## Evidence

Environment:

- `upstream/main` at `2f9367d6a76e8bab2bf056e0a1c545014f5fe18f`
- golangci-lint 2.10.1, matching the version installed by the required wrapper job
- Go 1.26.5
- `gms_pure_go` build tag

The same whole-repository lint command is clean for the Linux target:

```bash
env GOOS=linux GOARCH=amd64 CGO_ENABLED=0 \
  golangci-lint run --timeout=5m --build-tags=gms_pure_go ./...
```

```text
0 issues.
```

Changing only the selected Go target exposes two findings:

```bash
env GOOS=windows GOARCH=amd64 CGO_ENABLED=0 \
  golangci-lint run --timeout=5m --build-tags=gms_pure_go ./...
```

```text
internal/doltserver/doltserver_windows.go:109:21: Error return value of `windows.CloseHandle` is not checked (errcheck)
internal/linear/synclock_windows.go:14:21: Error return value of `windows.CloseHandle` is not checked (errcheck)
2 issues
```

The wrapper currently invokes only:

```text
golangci-lint run --timeout=5m --build-tags=gms_pure_go ./...
```

On its Ubuntu runner, Go package selection excludes both files before the linter can inspect them.

## Expected behavior

The required lint contract should cover Windows-tagged Go code as well as the native CI target.

## Proposed fix

After the normal host-target lint, run the same pinned linter with `GOOS=windows CGO_ENABLED=0`. Skip the second invocation when the caller is already targeting Windows.

This is a cross-target analysis pass. It requires neither a Windows runner nor execution of Windows binaries, and it adds no runtime dependency. The CI cost is explicit: non-Windows callers perform one additional whole-repository lint pass.

## Dependency and preview patch

PR #4981 fixes the two current findings and supplies the clean baseline this gate needs.

A focused one-file implementation is available as a [preview commit](https://github.com/ecuthiell/beads/commit/d53bcd094043cf99c80ee867b55e440a2f2e2007) stacked directly on the exact #4981 head. It is intentionally not an upstream PR yet: targeting `main` now would either duplicate #4981's functional changes or make the new required gate fail on the known baseline findings.

After #4981 merges, the preview commit can be rebased onto current `main`, revalidated, and submitted as a clean one-issue/one-commit PR containing only the lint-gate change.

---

ecuthiell:
## Summary

Make the public PR lint route select Windows-tagged non-CGO Go files on every host, and make a green required job prove which target, Make boundary, shell/path dialect, and tool identities actually ran.

Closes #4991.

PR #4981 cleared the two pre-existing Windows-only findings and supplied the clean baseline for this gate.

## What changes

- Keep one native-target lint pass, then add a mandatory `GOOS=windows CGO_ENABLED=0` whole-repository pass unless that is already the selected target.
- Route maintained lint callers through one public `make -f Makefile ci-pr-lint` contract.
- Bind Bash, Git, Go/gofmt, GNU Make, compiler, sed, and private `golangci-lint` identities before the lint child starts, and curate inherited Make, Bash, Git, and Go controls.
- Add required capability lanes for Linux, macOS arm64, native Windows, MSYS2-on-Windows, and Cygwin-on-Windows, with mandatory selection markers and CI Gate propagation.
- Install `golangci-lint` 2.10.1 into one private runner-temp directory and clean only its proved direct file/directory shape.
- Propagate a failing `gofmt` producer instead of accepting captured output as success.
- Keep contributor, agent, lint, CI-topology, cleanup, audit, formula, and freshness guidance synchronized with the maintained route.

This changes development and CI validation only; it adds no application runtime dependency.

## Hosted-run repairs

The first published generation failed closed and exposed two defects in the new hosted authority:

- Stock macOS `/usr/bin/make` resolves to an Xcode target. The repaired contract binds the original `/usr/bin/make` invocation leaf, proves the canonical target is executable, requires exact first-line `GNU Make 3.81`, and deliberately does not parse the mutable free-form build banner. A fifteenth refusal case rejects an alternate invocation leaf.
- Native Windows no longer relies on the hosted image's stale MinGW `PATH`. The install step binds the exact Chocolatey 16.1.0 `gcc.exe` and `g++.exe` leaves, checks version and target machine, and publishes their paths as one explicitly typed UTF-8-no-BOM payload. This replaces the failing `AppendAllLines(object[])` overload path.

The repair also gives Cygwin Git a private, process-scoped `safe.directory` configuration for the canonical workspace spelling instead of mutating runner-global state. Permanent Go structural tests reject the obsolete macOS banner authority, fourteen-case refusal count, untyped workflow-output path, and missing protected dependency.

## Why

The existing Ubuntu lint pass can be green without loading `//go:build windows` files. Before #4981, the equivalent Windows-target invocation exposed two unchecked `CloseHandle` findings that the required PR gate did not see.

The routing fix crosses Make, Bash, Go, Git, path-translation, compiler, and hosted-runner boundaries. The protected assertions prevent a green result produced by a different target, startup hook, Makefile, executable, compiler, Git ownership decision, or shell dialect than the workflow declares.

## Exact-head local validation

Candidate `add7951651469412987b9598deb48f59f1e95d53` passed:

- the complete `scripts` Go package in 30.498 seconds;
- the full integration-tagged `scripts` suite in 96.279 seconds;
- the native Windows public `ci-pr-lint` route with Go 1.26.5, GNU Make 4.4.1, MinGW 16.1.0, eleven mandatory routing cases, four authority refusals, and the nonzero-`gofmt` falsifier;
- native Windows/amd64/CGO and Windows/non-CGO golangci-lint passes with zero issues;
- actionlint, ShellCheck with sourced-file analysis, Bash syntax, and `go vet` for the affected scripts package; and
- canonical diff and whitespace hygiene.

An independent exact-head review found no remaining issue and confirmed that the rebuild retains the newer upstream Windows compiler-selection logic without conflicting with the lint authority.

These local results prove the source, native public route, parser, routing, and falsifier contracts available on this workstation. They do not substitute for the macOS, MSYS2, or Cygwin hosted boundaries below.

## Required post-publication proof

Every prior hosted generation belongs to a superseded head and is not reusable. A wholly new exact-`add7951651469412987b9598deb48f59f1e95d53` `pull_request` generation must complete all of these jobs successfully:

- `Build Artifacts` for the bound Linux public lint route;
- `PR Lint (wrapper timing)` for real macOS arm64 and stock GNU Make 3.81 semantics;
- `Windows Make shell (native)` for the exact Chocolatey compiler leaves, typed workflow-output handoff, Git-for-Windows Bash, and full native lint route;
- `Windows Make shell (msys2)` for real MSYS2 Make/shell/path semantics; and
- `Windows Make shell (cygwin)` for real Cygwin Make/shell/path and process-scoped Git ownership semantics.

Missing, skipped, cancelled, stale, wrong-head, failed, or ambiguously associated evidence blocks verification.

---

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -13,7 +13,7 @@
 
 ## Tech Stack
 
-- **Language**: Go 1.21+
+- **Language**: exact version from `go.mod` (currently Go 1.26.5)
 - **Storage**: Dolt (version-controlled SQL database)
 - **CLI Framework**: Cobra
 - **Testing**: Go standard testing + table-driven tests
@@ -29,7 +29,8 @@
 - Never create test issues in production DB (use temporary DB)
 
 ### Code Style
-- Run `golangci-lint run ./...` before committing
+- Run `make ci-pr-lint`; direct lint and pre-commit hooks are not equivalent to
+  the repository-owned v2 gate
 - Follow existing patterns in `cmd/bd/` for new commands
 - Add `--json` flag to all commands for programmatic use
 - Update docs when changing behavior
--- a/.github/workflows/ci-measurements.yml
+++ b/.github/workflows/ci-measurements.yml
@@ -76,6 +76,15 @@ jobs:
           dolt config --global --add user.email "ci@beads.test"
 
       - name: Measure commands
+        env:
+          BASH_ENV: ""
+          ENV: ""
+          BEADS_CI_EXPECTED_OUTER_OS: linux
+          BEADS_CI_EXPECTED_TARGET_GOOS: linux
+          BEADS_CI_EXPECTED_TARGET_GOARCH: amd64
+          BEADS_CI_EXPECTED_TARGET_CGO_ENABLED: "1"
+          BEADS_CI_EXPECTED_GOLANGCI_LINT_VERSION: "2.10.1"
+          BEADS_CI_INSTALL_GOLANGCI_LINT: "1"
         run: |
           set -euo pipefail
           mkdir -p artifacts
@@ -83,11 +92,13 @@ jobs:
           source ./.buildflags
 
           ci_time "install gotestsum" -- go install gotest.tools/gotestsum@v1.13.0
-          ci_time "install golangci-lint" -- go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1
           status=0
           ci_time "pr-policy wrapper" -- make ci-pr-policy || status=$?
           ci_time "pr-core gotestsum" -- gotestsum --junitfile artifacts/pr-core-junit.xml --format testdox -- -race -short -skip '^TestEmbedded' ./... || status=$?
-          ci_time "pr-lint wrapper" -- make ci-pr-lint || status=$?
+          ci_time "pr-lint wrapper" -- \
+            env -u MAKE -u MAKEFLAGS -u MFLAGS -u GNUMAKEFLAGS -u MAKEFILES \
+              /usr/bin/make -f Makefile CI_BASH=/usr/bin/bash ci-pr-lint ||
+            status=$?
           exit "$status"
 
       - name: Upload measurement artifacts
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,14 +62,21 @@ jobs:
         with:
           go-version-file: 'go.mod'
 
-      - name: Install golangci-lint
-        run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1
-
       - name: Run policy checks
         run: make ci-pr-policy
 
-      - name: Run lint checks
-        run: make ci-pr-lint
+      - name: Run bound lint checks
+        shell: /usr/bin/bash --noprofile --norc -euo pipefail {0}
+        env:
+          BASH_ENV: ""
+          ENV: ""
+          BEADS_CI_EXPECTED_OUTER_OS: linux
+          BEADS_CI_EXPECTED_TARGET_GOOS: linux
+          BEADS_CI_EXPECTED_TARGET_GOARCH: amd64
+          BEADS_CI_EXPECTED_TARGET_CGO_ENABLED: "1"
+          BEADS_CI_EXPECTED_GOLANGCI_LINT_VERSION: "2.10.1"
+          BEADS_CI_INSTALL_GOLANGCI_LINT: "1"
+        run: env -u MAKE -u MAKEFLAGS -u MFLAGS -u GNUMAKEFLAGS -u MAKEFILES /usr/bin/make -f Makefile CI_BASH=/usr/bin/bash ci-pr-lint
 
       - name: Build reusable Linux artifacts
         run: |
@@ -380,7 +387,8 @@ jobs:
 
   pr-lint-wrapper:
     name: PR Lint (wrapper timing)
-    runs-on: ubuntu-latest
+    runs-on: macos-15
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v6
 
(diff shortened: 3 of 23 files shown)


maphew (commented):
The gap you found is real, your fix for it is right, and we want to land that part immediately: there are 15 `go:build windows` files no lint pass ever saw, and your `GOOS=windows CGO_ENABLED=0` cross-lint is the correct, cheap closure of #4991. Same for the `fmt-check` exit-status propagation (a genuine pre-existing bug) and the linter version pin. A maintainer PR extracting exactly those pieces is going up today with `Co-authored-by` credit to you, so the fix stops waiting on the rest of this discussion.

The review below is long because required-check changes deserve it — it is scrutiny of the packaging, not doubt about the finding.

**Two concrete bugs currently holding your own lanes red** (both deterministic, neither flake):

1. `Windows Make shell (native)` — `.github/workflows/pr.yml` (~line 1039): PowerShell can't bind `@(...)` (`object[]`) to the `IEnumerable<string>` overload of `[IO.File]::AppendAllLines`, so the step dies writing its own output ("Cannot find an overload… argument count: 3"). One-line fix: cast `[string[]]@(...)`, or use `Add-Content -LiteralPath $env:GITHUB_OUTPUT -Encoding utf8`. As written the native lane can never go green.
2. `PR Lint (wrapper timing)` — `scripts/ci/pr-lint-host.sh` (~462): on hosted macOS, `/usr/bin/make` is the xcode-select shim and `$(MAKE)` resolves to the Xcode Developer-dir path, so the `path == /usr/bin/make` equality can only pass locally. Your own engdocs describe the right check ("exact 3.81 **with an apple-darwin build**") — the code just doesn't implement it yet; check the banner instead of the path. **Important**: the same code path runs in `main.yml`'s `pr-lint-wrapper` job, which only triggers on push-to-main — this PR's CI cannot exercise it, and it would go red on `main` the moment this merges. Please prove that lane on a branch run before this goes in.

**The scope question** — genuinely a question, not a verdict: the host/tool identity-binding machinery (~1,000 lines in `pr-lint-host.sh`, ~730 in the routing test, ~370 of falsifier assertions) defends against a lint pass silently running the wrong target/toolchain. But the cheap measures already in this PR — pinned linter version, pinned `--config`, `GOWORK=off`, `-mod=readonly`, `env -u MAKE -u MAKEFLAGS … -f Makefile` — cover most of that threat, and the only two failures the machinery has produced so far are its own false-reds on required checks. One finding that argues for trimming regardless of where you land on that: **there are no `windows && cgo` files in the tree**, so the CGO=1 native-Windows lint pass — the heaviest new cost (Go setup + 102 MB pinned MinGW from Chocolatey + two extra whole-repo lint passes on the scarcest runner pool) — adds zero coverage over your own `GOOS=windows CGO_ENABLED=0` pass.

Smaller should-fixes if the fuller shape proceeds: `go-version-file: go.mod` instead of the `1.26.5` literal (next Go bump reds the lane otherwise); keep `GIT_BUILD :=` immediate (the `=` change re-forks `git rev-parse` per reference); trim the ~30-path `LINTING.md` freshness source list (any rename among those files reds "Check doc freshness" for unrelated PRs); note in the measurement doc that the wrapper timing series now includes a from-source linter build.

Suggested path: the core lands today (credit preserved); this PR then slims to the host-binding/falsifier work as a follow-up, gated on the two bug fixes above and a branch-run proof of the `main.yml` lane. If you disagree with the trim — especially on the identity-binding value — argue it here; the falsifier discipline is unusual and we'd rather debate it than dismiss it.

---
Review provenance: maintainer sweep continuation; deep review ran against the PR head, the failing job logs (run 30208202951), and current main. Your other three PRs today (#5073/#5074/#5076) are approved and with the merge patrol.

maphew:
Core extraction is up as #5083 (windows cross-lint, fmt-check exit fix, linter pin — `Co-authored-by` you). Merge patrol takes it from here; this PR now carries only the host-binding/falsifier work per the review above.

ecuthiell:
Thanks for extracting and landing the core in #5083 with attribution. I agree that is the right scope. The new hosted generation for this remainder confirmed two more deterministic false-reds: stock macOS Make normalizes its automatic MAKE value away from the original /usr/bin/make spelling, and the native Windows lane exposes the repository's incomplete EOL policy by materializing ordinary Go files as CRLF. Those results reinforce the maintenance-cost concern rather than justify carrying the approximately 2,700-line host-identity layer. The substantive lint gap, fmt exit propagation, and linter pin are now merged, and the native CGO lane adds no tagged-file coverage. I am therefore withdrawing the remaining host-binding/falsifier machinery instead of churning required checks further. I will keep the repository-wide EOL defect as a separate focused proposal. Thanks for the careful review and the split-merge. 

_codex-gpt-5.6-sol ultra on behalf of Ewen Cuthiell_

PR #5077, opened 2026-07-26
fix(dolt): recover persisted remotes in the GH#2118 cold-start window (wy-6k7f7)

steveyegge:
## Problem (wyvern wy-6k7f7, filed by austen from the wy-82hc5 adversarial pass)

A freshly (auto-)started sql-server can report an **empty `dolt_remotes`** while the remote is persisted on disk in `.dolt/repo_state.json` (GH#2118). wy-xtv17/wy-82hc5 taught the push/sync no-remote gate (`hasConfiguredRemote`) to consult the on-disk probe, but three **mutating** siblings still decided "no remote" from `len(ListRemotes)` alone:

1. `bd config apply` blind-added the federation remote over the invisible one.
2. `bd dolt remote add` (`ensureDoltRemote`): an empty listing made `existingURL == ""`, so the URL-disagreement confirmation was skipped and the add silently clobbered the persisted remote.
3. `prepareCLIRouteForGitProtocol` fell back to SQL `DOLT_PUSH`, which the cold server refuses with a bare `remote origin not found`.

## Fix: recovery, not just detection

The on-disk evidence carries **URLs**, not just existence — so the window is recoverable:

- New `DoltStore.PersistedRemoteInfos()` exposes the persisted name+URL enumeration with `HasPersistedRemote`'s exact directory-search semantics (`HasPersistedRemote` is now derived from it). Embedded rigs return nil — GH#2118 is a sql-server phenomenon.
- New `persistedRemoteInfosFor` peels storage decorators exactly like `persistedRemoteProberFor` (direct implementers honored first, so test doubles work).
- **config apply**: `currentOriginRemoteURL` resolves origin from evidence — SQL listing first, persisted enumeration as fallback, a failed listing stays an error — routing apply into the same consistent/update legs it takes after the window.
- **remote add**: the idempotent re-add of the persisted URL is a no-op; a real disagreement gets the same confirm flow as a visible remote; a confirmed overwrite tolerates the cold server refusing to remove a remote it can't see yet (the add establishes the new URL).
- **push route**: a persisted git-protocol remote recovers the CLI route outright — the CLI transport never needed the SQL listing, so **the push proceeds**; a persisted non-git remote fails with the cold-start explanation instead of the bare `remote not found` (the loud-beats-silent trade from wy-xtv17 kept, with better words — the bead's explicit ask).

Read-only doctor/drift diagnostics still judge from the listing alone (they misreport, they don't mutate) — noted in `hasConfiguredRemote`'s doc comment and tracked as the bead's lower-priority tail.

## Tests

- `TestEnsureDoltRemoteColdStartSameURLIsNoop` / `...DifferentURLPrompts` (declined writes nothing; confirmed tolerates the unremovable invisible remote) — fake store implementing the new lister directly.
- `TestCurrentOriginRemoteURL` — listing wins when populated; cold-start recovers persisted origin; no evidence → no origin; **list failure is an error, never evidence**.
- `TestPrepareCLIRouteForGitProtocolColdStartWindow` — sqlmock empty listing + real CLI repo: git-protocol recovers the route; non-git fails naming GH#2118; nothing persisted keeps the SQL fallback.

## Acceptance (as filed on the bead, verified)

`go build ./...` + `go vet ./...` rc=0; `go test ./internal/storage/dolt/ -count=1` ok; `go test ./cmd/bd/ -count=1 -timeout 40m` **ok in 435s**.

### DIFF
--- a/cmd/bd/config_apply.go
+++ b/cmd/bd/config_apply.go
@@ -12,6 +12,7 @@ import (
 	"github.com/steveyegge/beads/internal/doltserver"
 	"github.com/steveyegge/beads/internal/git"
 	"github.com/steveyegge/beads/internal/metrics"
+	"github.com/steveyegge/beads/internal/storage"
 	"github.com/steveyegge/beads/internal/storage/dolt"
 	"github.com/steveyegge/beads/internal/storage/doltutil"
 )
@@ -217,6 +218,40 @@ func applyHooks(drifted bool, dryRun bool) ApplyResult {
 }
 
 // applyRemote ensures the Dolt origin remote matches federation.remote config.
+// originRemoteEvidence is the store surface currentOriginRemoteURL needs:
+// the SQL-visible remotes plus the on-disk persisted enumeration. The
+// concrete *dolt.DoltStore that applyRemote opens satisfies both directly —
+// no decorator peel needed here (unlike cmd paths holding the wired store).
+type originRemoteEvidence interface {
+	ListRemotes(ctx context.Context) ([]storage.RemoteInfo, error)
+	PersistedRemoteInfos() []storage.RemoteInfo
+}
+
+// currentOriginRemoteURL resolves the origin remote's URL from evidence, not
+// from the SQL listing alone: a freshly (auto-)started sql-server can report
+// an empty dolt_remotes while the remote is persisted on disk in
+// .dolt/repo_state.json (GH#2118, wy-6k7f7). Recovering the persisted URL
+// routes `bd config apply` into the same consistent/update legs it would
+// take after the window, instead of blind-adding over an invisible remote.
+// A failed listing returns the error — it is never evidence of "no remote".
+func currentOriginRemoteURL(ctx context.Context, st originRemoteEvidence) (string, error) {
+	remotes, err := st.ListRemotes(ctx)
+	if err != nil {
+		return "", err
+	}
+	for _, r := range remotes {
+		if r.Name == "origin" {
+			return r.URL, nil
+		}
+	}
+	for _, r := range st.PersistedRemoteInfos() {
+		if r.Name == "origin" {
+			return r.URL, nil
+		}
+	}
+	return "", nil
+}
+
 func applyRemote(drifted bool, dryRun bool) ApplyResult {
 	if !drifted {
 		return ApplyResult{
@@ -260,7 +295,7 @@ func applyRemote(drifted bool, dryRun bool) ApplyResult {
 	}
 	defer func() { _ = st.Close() }()
 
-	remotes, err := st.ListRemotes(ctx)
+	currentURL, err := currentOriginRemoteURL(ctx, st)
 	if err != nil {
 		return ApplyResult{
 			Check:   "remote",
@@ -268,13 +303,6 @@ func applyRemote(drifted bool, dryRun bool) ApplyResult {
 			Status:  applyStatusError,
 			Message: "Failed to list Dolt remotes",
 			Error:   err.Error(),
-		}
-	}
-	var currentURL string
-	for _, r := range remotes {
-		if r.Name == "origin" {
-			currentURL = r.URL
-			break
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


maphew:
Reviewed 2026-07-26 (head 1c0ec642e). **Approve-with-nits** — the approach is right, extends the wy-xtv17/wy-82hc5 direction properly, and it rebases trivially (zero drift in the four touched non-test files). Two things before merge:

1. **`config apply`'s update leg missed the tolerance** (cmd/bd/config_apply.go:333-342, unchanged by the PR): `currentOriginRemoteURL` can now return a disk-only URL; when it disagrees with `federation.remote`, apply calls `st.RemoveRemote` — which a cold-started server rejects — so apply hard-errors on a case that previously (wrongly, but successfully) blind-added. `ensureDoltRemote` got exactly this tolerance (`existingFromDiskOnly` → non-fatal remove). Suggest `currentOriginRemoteURL` returning `(url, fromDiskOnly, err)` and treating the RemoveRemote failure as non-fatal when disk-only, mirroring dolt.go — or, if erroring is deliberate, issue-2118 cold-start wording on the error instead of bare Dolt output.
2. **Per-name lookup stops at the first directory** (new `PersistedRemoteInfos`): it keeps `HasPersistedRemote`'s "first directory yielding ANY remotes wins" semantics, so if CLIDir holds only `peer-mini`, an `origin` persisted under the store's dbPath is never seen and callers silently fall back to pre-fix behavior. Accumulate across both directories, first-directory-wins per name.

Follow-up-bead material: the retained warning text still says "remote-migrate gate could not inspect…" on paths now reached from `bd config apply`/`remote add` (misleading there); the confirmed-overwrite tolerance is only exercised against a fake store, so what a real cold server does on `DOLT_REMOTE('add')` with an invisible same-name entry is unverified; SQL `RemoveRemote` never clears the CLI mirror, so on external-server rigs a legitimately removed remote persists as a URL these paths will now act on; and `persistedRemoteInfosFor` could collapse with `persistedRemoteProberFor` (~20 lines deleted, two test doubles to update).

steveyegge:

This is the right shape for the GH#2118 class: the wy-xtv17/wy-82hc5 work taught one decider to *detect* the window; this converges the three mutating siblings and upgrades detection to *recovery*, exploiting the fact the on-disk evidence carries URLs. Specifics that carried it:

- `PersistedRemoteInfos` derives `HasPersistedRemote` rather than duplicating the directory walk — one first-hit semantics, structurally shared. The doc comment's note that read failures never masquerade as evidence (callers consult only after a *successful* empty listing) is the key invariant, and `list_failure_is_an_error_not_evidence` pins it for config apply.
- `ensureDoltRemote`: the `existingFromDiskOnly` tolerance on RemoveRemote is correctly narrow — a cold server can't remove what it can't see, the confirmed add establishes the new URL, and a remove failure on a *visible* remote still errors loudly. The declined-overwrite-writes-nothing assertion is exactly the audit that matters.
- Push route: recognizing that the CLI transport never needed the SQL listing makes the git-protocol case a full recovery (push proceeds), and the non-git case now fails with an actionable explanation instead of DOLT_PUSH's bare 'remote not found' — the loud-beats-silent trade kept, with the bead's asked-for better words.
- `persistedRemoteInfosFor` peels decorators identically to the existing prober (direct implementer first, so fakes work) — consistent with the established pattern rather than a second convention.
- Honest scope: read-only doctor/drift diagnostics still judge from the listing, and the updated `hasConfiguredRemote` comment says so explicitly rather than claiming the hole closed. Good that it stays tracked as the bead's tail.

One observation, non-blocking: `currentOriginRemoteURL` falls back to disk whenever the listing lacks an origin — including a *populated* listing that shows only peer remotes. On a warm server disk and SQL converge so this is unreachable in practice, and the conservative direction (treat persisted origin as real) is the safe side of the bet. Fine as is.

Tests hit all three legs with the right instruments (fakes for the evidence rule, sqlmock + real CLI repo for the route decider). Acceptance receipts check out. Ship it.

PR #5085, opened 2026-07-26
fix(embeddeddolt): use stored peer credentials for federation sync

### LINKED ISSUE #5080: Federation sync ignores credentials stored by add-peer
Federation against any authenticated Dolt server fails on first sync: the
credentials add-peer stores are never presented. The same gap was closed
for default remotes in #2543; this is the peer-path remainder, and it
looks similarly small.

**Expected** (`bd federation add-peer --help`): "Credentials are encrypted
and stored locally. They are used automatically when syncing with the peer."

**Actual:** in these trials, sync presented `root` despite the credentials
stored by add-peer. The credentials are saved but never read back: a
write-only store.

This resembles the default-remote credential issue fixed by #2543 (from
#2515). The embedded federation path authenticates through
`remoteAuthUser()`, which reads `$DOLT_REMOTE_USER` but does not consult
the peer store. The server-mode federation path already routes stored peer
credentials.

Relevant paths: `internal/storage/embeddeddolt/version_control.go:361-363`,
`internal/storage/embeddeddolt/federation.go:216,271`, and
`internal/storage/dolt/federation.go:154-176` (all at 0c5fa422d).

Tested against a TLS Dolt server whose remotesapi rejects anonymous
access; a plain `dolt clone --user` with the same credentials succeeds
(bd 1.1.0, dolt 2.2.2):

| Trial | Client config | Result |
|---|---|---|
| T1 | stored credentials (`add-peer --user/--password`) | fetch fails: `Access denied for user 'root'` |
| T2 | stored + `DOLT_REMOTE_PASSWORD` | same; still presents `root` |
| T3 | stored + `DOLT_REMOTE_USER` + `DOLT_REMOTE_PASSWORD` | full sync succeeds (push then denied by the server's SuperUser write policy, as expected) |

T3 confirms that sync succeeds when credentials reach the transport;
only the peer-store lookup is missing.

Could embedded federation sync use the credentials stored by add-peer,
routed the same way #2543 routes default-remote credentials? Peers without
stored credentials should retain their current behavior.

Related: #4837 covers the inverse case, where federation status rejects
credential-free remotes.

<details><summary>Self-contained repro: script + one fresh top-to-bottom run in a clean Ubuntu container (bd 1.1.0, dolt 2.2.2)</summary>

```bash
#!/usr/bin/env bash
# Repro: bd federation sync ignores stored peer credentials.
#
# `bd federation add-peer --help` promises: "Credentials are encrypted and
# stored locally. They are used automatically when syncing with the peer."
# This script shows they are not. Against a credential-gated dolt remotesapi
# endpoint:
#   T1  sync with stored credentials only            -> fetch fails
#   T2  sync + DOLT_REMOTE_PASSWORD                  -> fetch fails (no user)
#   T3  sync + DOLT_REMOTE_USER + DOLT_REMOTE_PASSWORD -> fetch authenticates
# T3 proves the transport handles credentials fine when they arrive via the
# (federation-undocumented) env pair: only the stored-credential plumbing is
# missing. A plain `dolt clone --user` control with identical credentials
# succeeds throughout.
#
# Requires: bd, dolt, git, openssl. Linux (relies on SSL_CERT_FILE to trust
# the self-signed CA; macOS Go binaries use the system trust store instead).
# Self-contained: temp dir, random high ports, server killed on exit.
set -uo pipefail

LAB="$(mktemp -d)"
SQL_PORT=$((20000 + RANDOM % 10000))
API_PORT=$((SQL_PORT + 1))
step()  { printf '\n\033[1m== %s ==\033[0m\n' "$1"; }
result(){ printf '\033[36m-> %s\033[0m\n' "$1"; }
sql()   { dolt --host 127.0.0.1 --port "$SQL_PORT" --no-tls -u root -p "$1" sql -q "$2"; }

cleanup() { [ -f "$LAB/server.pid" ] && kill "$(cat "$LAB/server.pid")" 2>/dev/null; }
trap cleanup EXIT

step "1. self-signed cert + dolt sql-server with TLS + remotesapi (ports $SQL_PORT/$API_PORT)"
openssl req -x509 -newkey rsa:2048 -keyout "$LAB/tls.key" -out "$LAB/tls.crt" \
    -days 2 -nodes -subj "/CN=localhost" \
    -addext "subjectAltName=DNS:localhost,IP:127.0.0.1" 2>/dev/null
export SSL_CERT_FILE="$LAB/tls.crt"
mkdir -p "$LAB/server"
cat > "$LAB/server/config.yaml" <<EOF
log_level: debug
behavior: {read_only: false, autocommit: true}
listener:
  host: 127.0.0.1
  port: $SQL_PORT
  tls_key: $LAB/tls.key
  tls_cert: $LAB/tls.crt
remotesapi: {port: $API_PORT}
data_dir: $LAB/server
cfg_dir: $LAB/server/.doltcfg
EOF
dolt sql-server --config "$LAB/server/config.yaml" > "$LAB/server.log" 2>&1 &
echo $! > "$LAB/server.pid"
for _ in $(seq 20); do (exec 3<>"/dev/tcp/127.0.0.1/$SQL_PORT") 2>/dev/null && break; sleep 1; done

step "2. bd workspace (embedded dolt)"
mkdir -p "$LAB/ws" && cd "$LAB/ws" && git init -q .
bd init -q --prefix demo >/dev/null 2>&1
bd create "issue born on the local workspace" >/dev/null 2>&1
grep -o '"dolt_mode": *"[^"]*"' .beads/metadata.json

step "3. seed the server's peerdb FROM the workspace database (shared history)"
# The embedded workspace is a real dolt repo on disk. Push it to a file
# remote (remote storage format), then clone that server-side, so a
# correctly-authenticated sync would actually have a common ancestor.
WS_DOLT="$(dirname "$(find "$LAB/ws/.beads" -maxdepth 4 -type d -name .dolt | head -1)")"
mkdir -p "$LAB/stage"
( cd "$WS_DOLT" && dolt remote add stage "file://$LAB/stage" \
    && dolt push stage main ) >/dev/null 2>&1
sql '' "CALL DOLT_CLONE('file://$LAB/stage', 'peerdb');" >/dev/null
# Drop the seeding remote: bd federation sync iterates every dolt remote,
# and a leftover 'stage' would pollute the trial output.
( cd "$WS_DOLT" && dolt remote remove stage ) >/dev/null 2>&1
sql '' "USE peerdb; SELECT COUNT(*) AS peer_issue_rows FROM issues;" | tail -3

step "4. peeruser (SELECT + CLONE_ADMIN) + root lockdown (password on root)"
sql '' "CREATE USER 'peeruser'@'%' IDENTIFIED BY 'peerpass';
GRANT SELECT ON peerdb.* TO 'peeruser'@'%';
GRANT CLONE_ADMIN ON *.* TO 'peeruser'@'%';
ALTER USER 'root'@'localhost' IDENTIFIED BY 'locked-root-pw';" >/dev/null
result "root now requires a password; peeruser is the only sanctioned remote identity"

step "5. CONTROL: anonymous clone refused, authenticated clone succeeds"
if dolt clone "https://localhost:$API_PORT/peerdb" "$LAB/clone-anon" >/dev/null 2>&1; then
    result "UNEXPECTED: anonymous clone succeeded"
else
    result "anonymous: refused (endpoint is credential-gated)"
fi
if DOLT_REMOTE_PASSWORD=peerpass dolt clone --user peeruser \
       "https://localhost:$API_PORT/peerdb" "$LAB/clone-auth" >/dev/null 2>&1; then
    result "dolt clone --user peeruser: SUCCEEDS (server + credentials are fine)"
else
    result "UNEXPECTED: authenticated control clone failed"
fi

step "6. add-peer with stored credentials (the documented contract)"
bd federation add-peer team "https://localhost:$API_PORT/peerdb" \
    --user peeruser --password peerpass 2>&1 | tail -2

trial() {  # trial <label> <extra env...>
    local label="$1"; shift
    : > "$LAB/server.log"
    local out
    out="$(env "$@" bd federation sync team 2>&1 | grep -v '^$' | tail -2)"
    local auth
    auth="$(grep -a -i -o "access denied for user [^\"]*\|authenticat[^\"]*user[^\"]*" "$LAB/server.log" | sort -u | head -2)"
    printf '%s\n  bd:     %s\n  server: %s\n' "$label" "$out" "${auth:-(no auth log line)}"
}

step "7. T1: stored credentials only (documented: 'used automatically')"
trial "T1" __NOOP=1

step "8. T2: plus DOLT_REMOTE_PASSWORD"
trial "T2" DOLT_REMOTE_PASSWORD=peerpass

step "9. T3: plus DOLT_REMOTE_USER + DOLT_REMOTE_PASSWORD (undocumented for federation)"
trial "T3" DOLT_REMOTE_USER=peeruser DOLT_REMOTE_PASSWORD=peerpass

step "10. versions"
bd --version; dolt version | head -1
echo; echo "Lab dir kept for inspection: $LAB"
trap - EXIT
```

Transcript of one fresh run:

```text
== 1. self-signed cert + dolt sql-server with TLS + remotesapi (ports 26566/26567) ==

== 2. bd workspace (embedded dolt) ==
"dolt_mode": "embedded"

== 3. seed the server's peerdb FROM the workspace database (shared history) ==
| 1               |
+-----------------+


== 4. peeruser (SELECT + CLONE_ADMIN) + root lockdown (password on root) ==
-> root now requires a password; peeruser is the only sanctioned remote identity

== 5. CONTROL: anonymous clone refused, authenticated clone succeeds ==
-> anonymous: refused (endpoint is credential-gated)
-> dolt clone --user peeruser: SUCCEEDS (server + credentials are fine)

== 6. add-peer with stored credentials (the documented contract) ==
Added peer team: https://localhost:26567/peerdb
  User: peeruser (credentials stored)

== 7. T1: stored credentials only (documented: 'used automatically') ==
T1
  bd:     🔄 Syncing with team...
  ✗ fetch failed: fetch from team: Error 1105: could not access dolt url 'https://localhost:26567/peerdb': rpc error: code = Unauthenticated desc = API Authentication Failure: Access denied for user 'root' (errno 1045) (sqlstate 28000)
  server: authentication failed: API Authentication Failure: Access denied for user 'root' (errno 1045) (sqlstate 28000)

== 8. T2: plus DOLT_REMOTE_PASSWORD ==
T2
  bd:     🔄 Syncing with team...
  ✗ fetch failed: fetch from team: Error 1105: could not access dolt url 'https://localhost:26567/peerdb': rpc error: code = Unauthenticated desc = API Authentication Failure: Access denied for user 'root' (errno 1045) (sqlstate 28000)
  server: authentication failed: API Authentication Failure: Access denied for user 'root' (errno 1045) (sqlstate 28000)

== 9. T3: plus DOLT_REMOTE_USER + DOLT_REMOTE_PASSWORD (undocumented for federation) ==
T3
  bd:       ✓ Merged
  ○ Push skipped: push to team/main: Error 1105: unknown push error; rpc error: code = PermissionDenied desc = API Authorization Failure: peeruser has not been granted SuperUser access
  server: (no auth log line)

== 10. versions ==
bd version 1.1.0 (8e4e59d39)
dolt version 2.2.2

Lab dir kept for inspection: /tmp/tmp.fhSmxoUAee
```

</details>

I found this while operating credential-gated Dolt servers
(ArcavenAE/callbook); happy to follow up with a PR extending #2543's
route to the peer path.

arcaven:
## What

Embedded federation sync now uses the credentials stored by
`bd federation add-peer`, falling back to the
`DOLT_REMOTE_USER`/`DOLT_REMOTE_PASSWORD` environment pair when a remote
has no stored credentials.

## Why

add-peer promises "credentials are encrypted and stored locally. They
are used automatically when syncing with the peer", but embedded sync
read only the environment, so first sync against any authenticated Dolt
server presented the wrong identity (#5080). This is the peer-path
remainder of the class fixed for default remotes in #2543.

Resolution order is peer store first, environment second, and unknown
remotes are untouched: federation sync iterates every dolt remote, so
plain remotes and credential-free peers keep their current behavior.

On concurrency: the in-process Dolt engine reads these variables from
the process environment, so a stored pair is exported only inside a
`federationEnvMutex` critical section and both variables are restored
afterward, including when the operation fails; a stored empty field
unsets its variable rather than inheriting an ambient value. Stored-peer
operations are therefore serialized against each other. Operations with
no stored credentials read the environment without the lock. This
matches the lock discipline of the existing server-mode path
(`withRemoteOperationEnv` also locks only when stored credentials are
present), not its semantics: server mode's lock-free branch talks to an
external server that never sees this process's environment, while the
embedded engine here genuinely reads the environment.

Fixes #5080.

## Verification

- `make test`, and targeted:
  `go test -tags gms_pure_go ./internal/storage/embeddeddolt/ -run TestWithPeerAuth -count=1`
  (this package is CGO-gated; run with CGO enabled, the default on a
  native toolchain. ICU headers are not required with the repo's build
  tags, per engdocs/ICU-POLICY.md.)
- New tests: table-driven credential-resolution cases (stored pair wins;
  stored empty password/username do not inherit ambient values; unknown
  remote and credential-free peer fall back to env), env restoration on
  callback failure, restoration when no prior env existed, and a
  concurrent two-peer test asserting stored-peer operations never mix
  credentials.
- `make ci-pr-lint`, `make ci-pr-policy`, `make ci-pr-core` pass locally.
- End-to-end: the repro script in #5080, unchanged, against a build of
  this branch: T1 (stored credentials only) now fetches and merges as
  the stored user; before this change it failed with
  `Access denied for user 'root'`. Push remains denied by the server's
  SuperUser write policy, as designed. Transcript below.

Note: open PR #4303 does not touch these files.

<details><summary>Fresh top-to-bottom run of the #5080 repro script against this branch (clean Debian container, cgo build, dolt 2.2.2)</summary>

```text

^[[1m== 1. self-signed cert + dolt sql-server with TLS + remotesapi (ports 24404/24405) ==^[[0m

^[[1m== 2. bd workspace (embedded dolt) ==^[[0m
"dolt_mode": "embedded"

^[[1m== 3. seed the server's peerdb FROM the workspace database (shared history) ==^[[0m
| 1               |
+-----------------+


^[[1m== 4. peeruser (SELECT + CLONE_ADMIN) + root lockdown (password on root) ==^[[0m
^[[36m-> root now requires a password; peeruser is the only sanctioned remote identity^[[0m

^[[1m== 5. CONTROL: anonymous clone refused, authenticated clone succeeds ==^[[0m
^[[36m-> anonymous: refused (endpoint is credential-gated)^[[0m
^[[36m-> dolt clone --user peeruser: SUCCEEDS (server + credentials are fine)^[[0m

^[[1m== 6. add-peer with stored credentials (the documented contract) ==^[[0m
Added peer team: https://localhost:24405/peerdb
  User: peeruser (credentials stored)

^[[1m== 7. T1: stored credentials only (documented: 'used automatically') ==^[[0m
T1
  bd:       ✓ Merged
  ○ Push skipped: push to team/main: Error 1105: unknown push error; rpc error: code = PermissionDenied desc = API Authorization Failure: peeruser has not been granted SuperUser access
  server: (no auth log line)

^[[1m== 8. T2: plus DOLT_REMOTE_PASSWORD ==^[[0m
T2
  bd:       ✓ Merged
  ○ Push skipped: push to team/main: Error 1105: unknown push error; rpc error: code = PermissionDenied desc = API Authorization Failure: peeruser has not been granted SuperUser access
  server: (no auth log line)

^[[1m== 9. T3: plus DOLT_REMOTE_USER + DOLT_REMOTE_PASSWORD (undocumented for federation) ==^[[0m
T3
  bd:       ✓ Merged
  ○ Push skipped: push to team/main: Error 1105: unknown push error; rpc error: code = PermissionDenied desc = API Authorization Failure: peeruser has not been granted SuperUser access
  server: (no auth log line)

^[[1m== 10. versions ==^[[0m
bd version 1.1.0 (dev)
dolt version 2.2.2

Lab dir kept for inspection: /tmp/tmp.Xyz675T7ts
```

</details>


maphew (approved):
Thank you for closing the add-peer promise gap — this is careful concurrency work and it shows. We verified the parts that worried us and they hold: the defer ordering in `withPeerAuth` registers `Unlock` before the restore closure, so LIFO restores the environment *before* releasing the mutex (and on panic); `overrideEnv`'s `os.LookupEnv` restores "was unset" as unset rather than `""`; and the empty-field-unsets rule is actually stricter than the cleanup in package `dolt`'s own `setFederationCredentials`, which unconditionally `Unsetenv`s and clobbers pre-existing ambient values — your `overrideEnv` shape is the fix that package wants too, if you ever feel like a follow-up there. The `errors.Is(..., ErrNotFound)` chain survives the `errors.Join` wrapping, and the unknown-remote fallback test pins it. Approving as a strict improvement over env-only.

Four follow-ups we would love to see — the first two matter, and we are happy to take any of them maintainer-side if you would rather stop here:

1. **The same bug survives in `bd sync --remote <peer>` and `bd dolt push|pull --remote <peer>`.** A federation peer is registered as an ordinary Dolt remote under the same name (`AddRemoteIfNotExists(peer.Name, ...)`), and `PushRemote`/`PullRemote` in `internal/storage/embeddeddolt/version_control.go` still read only `remoteAuthUser()`. So after this PR, `bd federation sync --peer team` presents stored credentials while `bd sync --remote team` still presents the env identity — the #5080 failure via a different verb. Routing those two through `withPeerAuth` closes it, and the `ErrNotFound` fallback makes it a no-op for plain remotes. It would also close the residual unsynchronized-env-reader window (`remoteAuthUser` call sites read env without the lock while the in-process engine does the same), which today is only reachable by embedders and parallel tests, but is free to fix with the same edit.

2. **Undecryptable stored credentials now hard-fail with a raw cipher error.** `federation_peers` rows replicate with the database, but `.beads-credential-key` is machine-local and gitignored — so machine B pulls the DB, silently generates its own key, and every peer sync dies with `resolve peer credentials: decrypt password: cipher: message authentication failed` (and `bd federation status` renders it as `Unreachable:`, pointing at the network instead of the local key). Hard-fail is a defensible choice — server mode already behaves this way, and silently falling back could present the wrong identity — but the error should say what actually happened and what to do: that the credential key is machine-local, and to re-run `bd federation add-peer <name> <url> --user ...`. Whatever semantics you settle on, a test pinning the decrypt-failure branch would keep it from being changed by accident later.

3. Two comment nits: `federationEnvMutex` here is a *separate* lock from package `dolt`'s mutex of the same name — the "mirroring package dolt's federationEnvMutex" comment reads like it is the same lock, and one clause ("a separate lock; the two stores never run remote ops concurrently") would prevent the misread. Similarly, the PR-body claim that lock-free env reads "match the existing server-mode semantics" is not quite equivalent: server mode's lock-free branch talks to an external server that cannot see this process's env, while embedded mode's genuinely reads env for the in-process engine.

4. Diagnostic nit: a peer saved with a username but an empty password (Enter at the interactive prompt) now deliberately suppresses an ambient `DOLT_REMOTE_PASSWORD` — correct as a security default, but a user whose pre-PR setup worked off the env gets an auth failure with no hint. A one-line stderr/debug warning when a peer has a username, no stored password, and `DOLT_REMOTE_PASSWORD` set would save them the puzzle.

One mechanical thing before merge: the three CI failures on this run (`errcheck` on `windows.CloseHandle` in `internal/procid` and `internal/storage/dbproxy`) are inherited from the branch point — current main already fixes them, and your branch is ~50 commits behind. A rebase onto main should take CI green without touching your change.

arcaven:
Branch updated (no history rewrite: merge of main @ bfdc54b06 plus one commit).

- The merge was clean and brings in #5088's errcheck fixes, so the three red lint checks from the stale merge snapshot should go green on this run (`git diff` against main for both windows files is empty).
- The only non-merge change is the comment clarification from your point 3: `federationEnvMutex` now states it is a separate lock from package dolt's same-named mutex, with the never-concurrent rationale. I also corrected the PR-body sentence you flagged: it now claims matching lock discipline, not matching semantics.
- After the merge and edit: `TestWithPeerAuth` (5 subtests) plus the restoration and concurrency tests pass, and the full `-tags gms_pure_go ./internal/storage/embeddeddolt/...` suite is green; gofmt and vet clean.

Your follow-ups 1 (`bd sync --remote` / `bd dolt push|pull` through `withPeerAuth`) and 2 (actionable decrypt-failure error + pinned test) are real; I reproduced both. I plan to carry them as separate PRs after this lands, along with the point-4 warning, so this one stays reviewable as-is. Happy for you to take any of them maintainer-side if that is faster.

arcaven:
Follow-up 1 is implemented as #5207, stacked on this branch (its head plus one commit). All seven non-federation verbs now route through `withPeerAuth`; plain remotes keep the env fallback via `ErrNotFound`. If you would rather fold it into this PR before merging, I will push that same commit here instead. #5207's body carries the behavior notes worth a look before either path, in particular the fail-closed case on a machine holding replicated peer rows without the local credential key.

arcaven:
All four follow-ups from the approval review now have artifacts, plus the aside. Thanks again for the specificity of that review; every item mapped cleanly to a change.

1. Non-federation verbs through `withPeerAuth`: #5207 (stacked here, fold-in offer stands).
2. Actionable decrypt-failure error plus pinning tests, both backends and the status line: #5214 (stacked here).
3. The two comment nits: already landed on this branch (the `federationEnvMutex` separate-lock clause is this branch's head commit, and the body's server-mode-semantics sentence was reworked the same day).
4. Warning when a stored username-with-empty-password suppresses an ambient `DOLT_REMOTE_PASSWORD`: #5215 (stacked here).

The aside about package dolt's `setFederationCredentials` cleanup clobbering ambient values: #5216, from main, since it is independent of this branch.

One adjacent find while testing the remediation path for item 2: a bare re-run of `add-peer` clears a stored sovereignty tier (upsert with no `Flags().Changed` check). Filed as #5213 with a repro; the new PRs' hint texts are worded to avoid steering anyone into it.

#5214 and #5215 both add imports to `federation_credentials_test.go`, so whichever lands second gets a trivial refresh from us. Happy to fold any of the stacked ones into this PR instead if that is easier on your end.

maphew (commented):
Current-head consolidation review at `e87e5cbe`.

The contributor's stored-credential implementation remains valuable, and the focused diff/test inspection confirms the original behavior. The sibling-PR audit, however, shows that two fixes belong in this shared `withPeerAuth` seam before the stack lands:

- Bind a stored peer credential to the actual canonical remote URL, not only to a matching remote name. Otherwise a same-named remote redirected to another URL can receive the peer's secret.
- Take `federationEnvMutex` for fallback/plain-remote callbacks too. Today those paths call `remoteAuthUser()` without the lock and can observe another peer operation's temporarily installed environment pair.

Fix these once on first-filed #5085. Then rebase #5207 and #5214 onto it; #5215 remains an independent warning increment, and #5216 is independent. This supersedes the operational readiness implied by my earlier approval without discarding it or duplicating its detailed review.

Verification at this head:

- Full three-file diff and surrounding remote verbs inspected.
- Pure-Go focused embeddeddolt package build — PASS (`[no tests to run]` for the CGO-only cases).
- Blocking preflight — blocked on GitHub merge state `UNKNOWN`; no branch action taken.

Disposition: **maintainer-fix on #5085, then merge/rebase the sibling stack**.

steveyegge:
Delta review of the post-approval commits before merging on maphew's standing approval (at `8744d81db`): the substantive one is `fdfa7c2d0` — verified sound. The URL-binding check fails closed correctly (`remote_url` is NOT NULL in 0015 and always written by add-peer, so no legacy peer row can get bricked), and moving the plain-remote/empty-credential fallbacks under `federationEnvMutex` closes the env-observation race the commit describes. The three regressions cover diverged-URL, missing-remote, and fallback-isolation. Stacked follow-ups #5207/#5214/#5215 noted — they rebase on this merge. CI fully green at `64f9428f8` (99 pass). Merging.

Issue #5084, opened 2026-07-26
Opening a fully-seeded database still writes: dolt_ignore seeding denies read-only accounts


arcaven:
Strictly read-only accounts cannot run `bd list` even once the open path
stops issuing CREATE DATABASE (#5079): migrate re-asserts the canonical
dolt_ignore patterns with `INSERT IGNORE` on every open, and the
privilege check fires even when every pattern is already present. A
read-before-write check would avoid this INSERT when every canonical
pattern already exists.

**Expected:** opening a fully-migrated, fully-seeded database should not
require INSERT privilege merely to reassert rows already present; per
the #2189 principle, open paths do not mutate unless initializing.

**Actual:** `MigrateUp` asserts six dolt_ignore patterns with
`INSERT IGNORE INTO dolt_ignore VALUES (?, true)` at the top of every
open, present or not. The seed's comment scopes this to write-mode
opens, but the proxied open path runs migrate for every command, reads
included. Dolt checks INSERT privilege before the IGNORE no-op (the
same shape noted in #5079 for `CREATE DATABASE IF NOT EXISTS`), so a
SELECT-only account fails with:

`uow: init schema: uow: migrate: seeding dolt_ignore pattern
"ignored_schema_migrations": Error 1105 (HY000): command denied to user
'reader'@'%'`

measured on a database whose six patterns were already present and
committed (a build carrying the probe-first change proposed for #5079,
now #5086; dolt 2.2.2).

Relevant paths: `seedDoltIgnorePatterns` and `doltIgnorePatterns` in
`internal/storage/schema/schema.go`, called from `MigrateUp`.

The seed exists for a good reason: it heals databases materialized
out-of-band, where the migration cursor is at-latest but the one-shot
0019/0028 side effects never ran (the #4566 self-heal contract). A probe
keeps that: read dolt_ignore, insert only the missing patterns. A healthy
database then sees zero writes on open; an under-seeded one heals exactly
as today when credentials allow, and when they do not, the error can say
the database is under-seeded and needs a writable maintenance run.

Could `seedDoltIgnorePatterns` check for missing patterns before
inserting? A row with ignored=false remains an explicit operator
override and should not be changed.

<details><summary>Self-contained repro: script + one fresh top-to-bottom run in a clean Debian container (bd built from the probe-first branch, dolt 2.2.2)</summary>

```bash
#!/usr/bin/env bash
# Repro: opening a fully-seeded proxied-server database still writes.
[...]

Issue #5079, opened 2026-07-26
Proxied-server mode issues CREATE DATABASE during read commands


arcaven:
Least-privilege accounts cannot use proxied-server mode at all; even
`bd list` fails, because every open starts with a bare `CREATE DATABASE`.
The direct stack already has the guard that fixes this, and adopting it
may also clear #5037, #5012, and #4861.

**Expected:** an account with read access to an existing database can run
read commands without requiring global CREATE. The direct-server stack
provides this through its CreateIfMissing guard, following the principle
established in #2189: do not create a database unless explicitly
initializing one.

Relevant paths: `databaseExistsOnServer` and the CreateIfMissing guard in
`internal/storage/dolt/store.go`; the flag is set only by
`cmd/bd/init.go`.

**Actual:** opening the proxied provider for a command first issues a bare
`CREATE DATABASE`. Accounts without global CREATE therefore fail before
the requested read begins.

Relevant paths: `initSchema` in
`internal/storage/uow/dolt_sql_provider.go`, reached via the open path in
`internal/storage/uow/external_doltserver_provider.go`.

Measured (script below; bd built from main at 0c5fa422d, dolt 2.2.2):

- **D1:** `bd list` is a query that ships DDL: 10 runs produced **20** bare
  `` CREATE DATABASE `demo` `` statements. The implementation confirms at
  least one attempt during each open.
- **D2:** SELECT-only account, existing database: `bd list` returns
  `uow: creating database: db: CreateDatabase: Error 1105 (HY000): command
  denied to user 'reader'@'%'`.
- **D3:** with otherwise identical grants, adding global CREATE allowed the
  command to succeed. Database-scoped grants did not substitute for it in
  this repro.

At current main the first open attempt still issues a bare `CREATE
DATABASE`, and a privilege denial (Error 1105) is still classified as
permanent; the recent retry-loop hardening (#5042, #5014) changed the
retry semantics but kept both behaviors. Only "already exists" (1007) is
handled. `CREATE DATABASE IF NOT EXISTS` would not help: the permission
check comes first, and Dolt still rejects it for a restricted account
(verified on dolt 2.2.2). The direct stack's probe-first approach, looking
before it leaps, avoids issuing DDL during read operations: it opens with
a SHOW DATABASES probe and only creates under an explicit CreateIfMissing
flag that bd init alone sets.

This may explain or contribute to #5037 (Error 1007 on an existing
database), #5012 (proxied init races), and #4861 (no sanctioned path to
[...]

PR #5086, opened 2026-07-26
fix(uow): probe for existing database before CREATE on proxied-server open

arcaven:
## What

Proxied-server opens probe for the database before any DDL: an existing
database is opened with USE plus migrate only. A
`WithCreateIfMissing(bool)` provider option gates creation, and it now
defaults to **false**, so callers opt in explicitly. `bd init` opts in
through the stacked cmd/bd change in #5087.

## Why

Every proxied open issued a bare `CREATE DATABASE` before the requested
command, so an account without the server-wide CREATE privilege failed
even `bd list` with Error 1105 on a database that already exists
(#5079). `CREATE DATABASE IF NOT EXISTS` would not help: the privilege
check precedes the existence check (verified on dolt 2.2.2).

The probe (`DatabaseExists`) iterates SHOW DATABASES rather than using
LIKE, since Dolt treats `_` as a wildcard. When creation happens, the
#5012/#5014 semantics are unchanged: the bare CREATE lets the server
arbitrate the race atomically, and only the proven creator passes
WithFreshBootstrapHeal.

Addresses the CREATE-on-open defect in #5079. Strictly SELECT-only
accounts remain blocked by migrate's dolt_ignore seed write, tracked in
#5084.

## Current head

`103d0948e`, a merge of main @ `bfdc54b06ab7` plus review commits, with
no history rewrite. Two commits on top of main:

```
$ git log --oneline --no-merges bfdc54b06..103d0948e
103d0948e fix(uow): createIfMissing defaults to false; not-found error points at bd bootstrap
8a210de87 fix(uow): probe before creating the proxied-server database

$ git diff --stat bfdc54b06..103d0948e | tail -1
 9 files changed, 254 insertions(+), 19 deletions(-)
```

The default flip lands in this layer because the mechanism does. One
consequence stated plainly: this PR is no longer standalone-safe, since
a fresh `bd init --proxied-server` now needs the cmd/bd wiring in #5087.
The standalone argument in my 2026-07-28 comment no longer applies, and
the stack should land as the two-commit pair you proposed.

## Verification

- Guard test: `WithCreateIfMissing(false)` against a missing database
  creates nothing and returns `uow: database "<name>" not found on Dolt
  server; run 'bd init' to create a new database, or 'bd bootstrap' to
  restore an existing project`. The test asserts both halves of that
  message.
- Regression test: an existing database opens under an account holding
  DML on its own database and no server CREATE privilege, which failed
  with Error 1105 before this change.
- Whole-repo `-tags gms_pure_go` build green, full uow suite green under
  Docker, and the #5079 least-privilege regression passes. Details are
  in the #5087 rundown rather than restated here:
  https://github.com/gastownhall/beads/pull/5087#issuecomment-5147345726
- `make ci-pr-*` was **not** run locally on this head. The default-cgo
  build hits a pre-existing vendored ICU header issue on this machine,
  unrelated to the diff, so upstream CI is the authority. This
  supersedes the "pass locally" line this body carried for the earlier
  head.
- On the lint failures visible earlier: the errcheck hits came from
  windows files added to main by #5013 after the branch point and fixed
  on main by #5088, so they are not attributable to this diff and clear
  on the merge of main. Recorded at
  https://github.com/gastownhall/beads/pull/5086#issuecomment-5146806208

Note: open PR #4303 touches the same provider files (connection
pooling); the changes are functionally independent, and I am happy to
rebase whichever lands second.

<details><summary>Fresh top-to-bottom run of the #5079 repro script at 8a210de87 (clean Debian container, dolt 2.2.2)</summary>

```text

^[[1m== 1. dolt sql-server (debug logging, port 24164) ==^[[0m

^[[1m== 2. workspace bound to the server (proxied mode, as an admin-tier account) + first bd command ==^[[0m
"dolt_mode": "proxied-server"

^[[1m== 3. D1: 10 read-only commands, counting CREATE DATABASE attempts ==^[[0m
^[[36m-> 10x 'bd list' -> 0 CREATE DATABASE statement(s) reached the server^[[0m

^[[1m== 4. restricted users: reader (SELECT only) and creator (ALL on demo + global CREATE) ==^[[0m

^[[1m== 5. D2: same workspace, SELECT-only user: even 'bd list' fails on CREATE ==^[[0m
Error: failed to open uow provider: uow: init schema: uow: migrate: seeding dolt_ignore pattern "ignored_schema_migrations": Error 1105 (HY000): command denied to user 'reader'@'%'

^[[1m== 6. D3: identical grants + global CREATE: everything works ==^[[0m
^[[36m-> bd list SUCCEEDS: the only delta from D2 is the global CREATE privilege^[[0m

^[[1m== 7. what the server refused (D2's denial, from the server log) ==^[[0m
command denied to user 'reader'@'%'
command denied to user 'reader'@'%' (errno 1105) (sqlstate HY000)

^[[1m== 8. versions ==^[[0m
bd version 1.1.0 (dev)
dolt version 2.2.2

Lab dir kept for inspection: /tmp/tmp.0KrBN5Zzub
```

</details>

Body rewritten 2026-07-31 to match head `103d0948e`. The previous
version described the pre-review shape, with `createIfMissing`
defaulting to true and an init-only not-found message. Earlier comments
are left as posted, corrected above where they went stale.

Part of #5079.

---


maphew:
Comparing this with #5087: the diff from `internal/storage/domain/db/ddl.go` onward is line-for-line identical there, and #5087 adds the CLI policy wiring on top (`bd init --proxied-server` keeps create-if-missing, ordinary commands and `dolt clean-databases` refuse and point at `bd init`). One semantic difference worth confirming: under this PR alone, ordinary commands retain the default `createIfMissing=true`; under #5087 they intentionally lose implicit creation.

Question before we route review effort: is this PR meant as a standalone narrower option (probe-before-CREATE but keep implicit creation), or as the extracted base of a #5087 stack? If it's the latter we'll review #5087 as the canonical change and retire this one once it lands.

arcaven:
> Question before we route review effort: is this PR meant as a standalone narrower option (probe-before-CREATE but keep implicit creation), or as the extracted base of a #5087 stack? If it's the latter we'll review #5087 as the canonical change and retire this one once it lands.

Option C: it's base of stack but designed to be standaline-safe.
Per  CONTRIBUTING_PR_GUIDELINES.md:19 it's one layer per PR and primitive unblocking application PR

From my internal notes (in beads) 
> Our design record names them as a stack. beads-pr-drafts.md labels them "PR 2a = #5086" and "PR 2b = #5087 (stacked on #5086)". ... So merging 2a alone was designed to be a no-behavior-change outcome, on purpose, as insurance.
 
I would recommend considering adopting rather than deferring to #5087 for these reasons:

> ... offered to "review #5087 as the canonical change and retire this one once it lands." I'd decline, for four reasons, and the mechanical one is decisive.
>
> - #5087 is a superset. Its commit list is 8a210de87 (the entirety of #5086) plus d2b24c168. Both PRs target main. So if #5087 merges first, #5086 becomes empty and gets closed, not merged, and the order has to be stated explicitly or someone will merge the wrong one first.
> - The storage guard test lives in #5086. That is why #5087's Verification section points at it rather than restating it. Reviewing #5087 alone means reviewing that test as part of a two-layer diff, which is the coupling their rule exists to prevent.
> - The risk profiles differ. #5086 fixes a hard failure (existing database plus an account without server-wide CREATE means even bd list dies with Error 1105) while preserving every default. #5087 removes implicit creation, which can surprise anyone relying on auto-create. If the policy half needs discussion, the defect half should not wait behind it.
> - CI is not a differentiator. Neither PR has run checks (zero statuses, both pending, likely awaiting fork-workflow approval), so unstable on both is just absent required checks. Our evidence remains the local make ci-pr-* runs already reported in the bodies.
> 
> The ask should be "merge #5086, then #5087," which is what current guidelines prescribe.
>
> The infra context changes the emphasis

This is the part worth attention, and it argues for both PRs (two-step) more strongly than the PR bodies do. Our own Dolt services have an ADR that reads:

>> var.databases is inventory + drop-protection only; grants stay instance-wide so bd's unconditional CREATE DATABASE IF NOT EXISTS keeps working. Scoping is a separate opt-in that is incompatible with beads.
>

I hit this in production design and paid for it by giving up per-database scoping. scoped_grants_enabled exists and is documented as incompatible with beads. The tiers are rw (the app-owner set migrations need) and ro (SELECT, EXECUTE, CLONE_ADMIN), on a shared multi-project instance.

Two consequences:
 
> - #5086 is the unblocker for a ** least-privilege goal. ** Once opens probe instead of issuing CREATE, a per-database grant works for the normal case, and scoped_grants_enabled stops being a dead option.

Granted, no benefit until release, but decoupling, and if this slips into the next release, there are very real velocity wins for us (and for others?)

> - #5087 is also real hardening for us, not just tidiness. On a shared instance with several projects, a typo'd workspace name silently creating a database is a live failure mode. But 5086 advances/unblocks real designs, on it's own, now, which I've actually reached for.
> - #5084 still blocks the ro tier (migrate's dolt_ignore seed write), so a SELECT-only account remains blocked either way. Worth keeping straight so I don't oversell what this stack delivers.

That is genuine independent motivation, and I think it's worth two-stepping this. 

At the same time, I wouldn't hesitate to close both (to prevent a half-fix) and re-bundle as a single PR, suggest it and it will be done. Thank you, grateful for beads.

arcaven:
Status update: my comment above has gone stale in two places.

1. "Neither PR has run checks" was true at posting (the first check-run landed a few hours later) but is now misleading: CI has run and both PRs show failing lint. The errcheck hits are in windows files added to main by #5013 after the branch point and fixed on main by #5088, so they are not attributable to either diff; a rebase clears them. Both PRs also now conflict with main after #5183 touched the same storage files.

2. "becomes empty and gets closed, not merged" overstated GitHub mechanics; the outcome depends on the merge method used. The durable point is only that the landing order must be explicit.

I will follow up with the rebase and the four review asks on #5087.

arcaven:
Branch updated alongside #5087 (merge of main @ bfdc54b06 resolving #5183's threading, plus the ask-1/ask-2 commit). Details and the four-ask rundown are in the #5087 comment; landing order stays 5086 then 5087, or land #5087 as the two-commit stack per your plan.

maphew (commented):
Reviewed at `103d0948`.

The probe-first mechanism and focused storage tests are sound, but this head is intentionally no longer standalone-safe: database creation defaults off while this PR's init caller does not opt in, so a fresh proxied init would fail.

#5087 contains these exact contributor commits plus the required CLI wiring and tests. The safest consolidation is to land #5087 as the complete stack and then retire this PR as carried forward, preserving its commits and attribution. Do not merge #5086 independently.

arcaven:
consolidated into #5087, thank you!

Issue #2189, opened 2026-02-28
CREATE DATABASE IF NOT EXISTS creates shadow database on wrong server


aaronlippold:
### Bug Description

When bd connects to a Dolt server that doesn't have the configured database, `openServerConnection()` unconditionally runs `CREATE DATABASE IF NOT EXISTS`, creating a fresh empty database. This shadows the real database which exists on a different server/data_dir.

### Reproduction Steps

1. Configure `metadata.json` with `dolt_server_port: 3307` pointing to an external Dolt server with data
2. Stop the server on 3307
3. Run any `bd` command
4. Auto-start launches a new server from `.beads/dolt/` on a derived port
5. `CREATE DATABASE IF NOT EXISTS` creates empty database on the new server
6. `bd list` shows 0 issues (data exists on disk but server serves empty shadow)

### Expected Behavior

- When configured server is unreachable and explicit port is set: error out, don't auto-start
- When connecting to any server: don't `CREATE DATABASE` unless explicitly initializing (`bd init`)

### Actual Behavior

- Auto-start silently launches different server
- `CREATE DATABASE IF NOT EXISTS` creates empty shadow database
- User sees 0 issues, thinks data is lost
- `bd doctor` does not detect this scenario

### Root Cause

- `store.go:752`: unconditional `CREATE DATABASE IF NOT EXISTS`
- `store.go:558-574`: auto-start doesn't respect explicit server configuration
- No verification that connected server is the expected one

### Additional Bugs Found and Fixed in PR #2188

While investigating and testing the shadow DB fix, two additional bugs were discovered:

**1. Backup git push fails across `.beads/redirect`** (`cmd/bd/backup_git.go`)
- When `.beads/redirect` points to a different project's `.beads/`, the backup directory resolves to that other project's repo
- `gitBackup()` ran `git add` in the CWD's repo, but the backup dir was outside it
- Error: `fatal: '/other-project/.beads/backup' is outside repository`
- Fix: `findGitRoot()` resolves the correct repo from the backup directory location

**2. macOS `CanonicalizePath` doesn't resolve case** (`internal/utils/path.go`)
- `resolveCanonicalCase()` used `realpath(1)` which only resolves symlinks, not filesystem case
- On case-insensitive HFS+/APFS, passing `/path/testcase` when the real name is `TestCase` returned the wrong case
- This breaks git operations that string-compare paths (GH#880)
- Fix: Walk directory entries with `os.ReadDir` and `strings.EqualFold` matching

[...]

PR #5087, opened 2026-07-26
fix(cmd/bd): create the proxied-server database only on bd init

### LINKED ISSUE #5079: Proxied-server mode issues CREATE DATABASE during read commands

arcaven:
Stacked on #5086 (probe-first open). Review #5086 first.

## What

`bd init --proxied-server` is the only command whose open may create the
database. Ordinary command opens of a missing database fail with `uow:
database "<name>" not found on Dolt server; run 'bd init' to create a new
database, or 'bd bootstrap' to restore an existing project` instead of
silently creating one. `dolt clean-databases` opens server-wide without
binding a database, so cleanup works in the broken state it exists for.

## Why

Follows #2189's principle: do not create a database unless explicitly
initializing one. #5086 supplies the mechanism (`WithCreateIfMissing`,
now defaulting to false) and its storage-layer tests; this PR is the
cmd/bd policy wiring, passing the option explicitly at every proxied
provider call site. Together they close the create half of #5079.

On a shared instance holding several projects, a mistyped workspace name
silently creating a database is a live failure mode, so this half is
hardening rather than tidiness.

## Current head

`2b1a929ee`, a merge of main @ `bfdc54b06ab7` plus review commits, with
no history rewrite. Five commits on top of main. `8a210de87` and
`103d0948e` are #5086; `d2b24c168`, `7fb680f2d`, and `2b1a929ee` are this
PR:

```
$ git log --oneline --no-merges bfdc54b06..2b1a929ee
2b1a929ee fix(cmd/bd): open clean-databases without binding a database
7fb680f2d test(cmd/bd): non-Docker guard for the proxied create policy
103d0948e fix(uow): createIfMissing defaults to false; not-found error points at bd bootstrap
d2b24c168 fix(cmd/bd): create the proxied-server database only on bd init
8a210de87 fix(uow): probe before creating the proxied-server database

$ git diff --stat bfdc54b06..2b1a929ee | tail -1
 15 files changed, 576 insertions(+), 29 deletions(-)
```

The "only the top commit is new here" line this body carried was true of
head `d2b24c168`. Three commits have landed since.

## Review status

Your four asks are answered in
https://github.com/gastownhall/beads/pull/5087#issuecomment-5147345726
covering the error text, the default flip, the non-Docker policy test,
and the clean-databases posture. I have not restated that rundown here.
One question in it is still open: whether `dolt clean-databases` should
keep the bind-and-check shape rather than the no-bind open. Say the word
and I will flip it back.

The CHANGES_REQUESTED review predates this head. It was submitted
against `d2b24c168`, before `7fb680f2d` and `2b1a929ee` landed, and
GitHub does not clear a review when the head moves.

## Verification

- Storage-layer behavior (guard error, nothing created) is tested in
  #5086.
- `cmd/bd/proxied_create_policy_test.go` drives the three call sites
  through a seam on the proxied open, without Docker: init resolves
  create-enabled, while the root dispatch and `dolt clean-databases`
  resolve create-disabled.
- Whole-repo `-tags gms_pure_go` build green, full uow suite green under
  Docker, and the new policy test passes without Docker.
- `make ci-pr-*` was **not** run locally on this head. The default-cgo
  build hits a pre-existing vendored ICU header issue on this machine,
  unrelated to the diff, so upstream CI is the authority. This
  supersedes the "pass locally" line this body carried for the earlier
  head. The lint failures visible earlier trace to windows files added
  to main by #5013 after the branch point and fixed on main by #5088.

Note: open PR #4303 touches cmd/bd/uow_factory.go (the constructor
signatures most likely to conflict); functionally independent, and I am
happy to rebase whichever lands second.

Body rewritten 2026-07-31 to match head `2b1a929ee`. The previous
version described head `d2b24c168`, with `WithCreateIfMissing` defaulting
to true and an init-only not-found message.

Fixes #5079.

---


maphew (changes requested):
Thanks for this pair — the stack split is exactly right (mechanism in #5086, policy here), the probe-first design with the bare-CREATE arbitration preserved is careful work, and the wildcard-safe `DatabaseExists` iteration with the comment explaining *why* `SHOW DATABASES LIKE` is unsafe is the kind of detail we love to see. The least-privilege regression test is a faithful repro of #5079. We want to land this stack, and ahead of #4303 to keep your rebase burden low.

Four asks before we do — three small, one design question:

1. **Point existing projects at `bd bootstrap`, not `bd init`.** The not-found error (`internal/storage/uow/dolt_sql_provider.go`, the `!createIfMissing` branch) says `run 'bd init' to create it`. The common way to hit this path is a machine that just cloned an existing project (or whose configured DB was dropped server-side) — for those, `bd init` is the wrong tool and can steer people into re-initializing a workspace that should be bootstrapped from the remote. Suggest wording like: `run 'bd init' to create a new database, or 'bd bootstrap' to restore an existing project`.

2. **`createIfMissing` defaulting to `true` re-opens the hole for future callers.** Every current call site passes the option explicitly, which is great — but the default in `newUOWProviderOptions` means the next caller of `NewExternalDoltServerUOWProvider`/`NewDoltServerUOWProvider` who forgets the option silently gets the implicit-create behavior this PR exists to remove. Since you already touched every call site, flipping the default to `false` (init passes `WithCreateIfMissing(true)` explicitly) makes the safe behavior the one you get by accident. If there is an external/embedder compat reason to keep `true`, a comment at the default saying so would do.

3. **The cmd/bd policy wiring has no test.** Both new tests exercise the storage layer via Docker (`StartIsolatedDoltContainer`), but nothing asserts the actual policy this PR adds: `bd init` opens with create enabled and every other command path (root command, `dolt clean-databases`) opens with create disabled. A lightweight non-Docker test — even just asserting which option `newProxiedServerUOWProvider` receives from each call site, or a fake-provider seam — would keep the wiring from regressing when #4303 reshapes `uow_factory.go` constructor signatures. That collision is likely, so a guard here is cheap insurance.

4. **Design question: should `bd dolt clean-databases` require the configured DB to exist?** `runDoltCleanDatabasesProxied` now passes `WithCreateIfMissing(false)`, so if the configured database was dropped server-side, the cleanup command fails at open with not-found. But cleaning up stale databases is precisely what you might reach for in that broken state. Should it open without binding to the configured database (probe-only / no `USE`), or is failing fast the intended behavior here? Either answer is fine — we would just like it to be deliberate, and a line in the PR body or a comment covering it.

Happy to help with any of these if you would rather we push follow-ups — say the word. Once these are settled we plan to land #5087 as the two-commit stack and close #5086 as carried-by-stack with attribution.

arcaven:
Both branches updated (merges of main @ bfdc54b06 plus review commits; no history rewrite). The conflicts were #5183's teamServer/expectedProjectID threading in the three uow providers and the factory reshape in cmd/bd; the resolution carries both threads, with the variadic options parameter kept last. All four asks are in:

1. **Error text**: the `!createIfMissing` not-found error now reads `run 'bd init' to create a new database, or 'bd bootstrap' to restore an existing project`, and the guard test asserts both halves.
2. **Default flipped**: `newUOWProviderOptions` defaults `createIfMissing` to false; `bd init` opts in explicitly on the cmd/bd layer, and every test call site that provisions a fresh database passes `WithCreateIfMissing(true)` explicitly. One consequence stated plainly: #5086 alone is no longer standalone-safe (a fresh `bd init --proxied-server` needs the cmd/bd wiring in this PR), so the earlier standalone argument no longer applies and the stack should land as the two-commit pair you proposed.
3. **Policy wiring test**: `cmd/bd/proxied_create_policy_test.go` is a non-Docker test using a seam on the proxied open; it drives the three call sites genuinely and asserts init resolves create-enabled while the root dispatch and `dolt clean-databases` resolve create-disabled.
4. **clean-databases decided as probe-only**: `WithNoDatabaseBind()` opens a server-wide connection with no probe, no create, no USE, no migrate, so cleanup works precisely in the broken state it exists for (configured database dropped server-side). One posture note so this is chosen, not slipped: on main the Adopting open ran the team-server schema/identity checks; the no-bind open runs neither, which we think is right for a command that only issues SHOW DATABASES / DROP / PURGE. If you'd rather keep the bind-and-check shape there, say so and I'll flip it back.

Verification: whole-repo `-tags gms_pure_go` build green on both branches; full uow suite (Docker) green on both; the #5079 least-privilege regression passes; the new policy test passes without Docker. `make ci-pr-*` was not run locally (default-cgo build hits a pre-existing vendored ICU header issue on this machine, unrelated); CI here is the authority.

_claude-code-claude-opus-5-high on behalf of arcaven_

maphew (approved):
Reviewed at `33c05bbc`.

The updated stack addresses all four prior review points, and the focused creation-policy and no-bind tests pass. Code review is approved.

GitHub currently reports a base conflict and this refreshed head has no required CI results yet, so refresh it onto current main and rerun checks before landing. #5087 contains the exact #5086 commits plus the required CLI policy and tests; the attribution-preserving consolidation is to land this complete stack and then retire #5086 as carried by it.

steveyegge:
Found why PR Core + macOS keep failing here while the diff itself is sound: `TestProxiedCreatePolicy_OrdinaryCommandOpensWithCreateDisabled` drives real root dispatch in a proxied workspace, which sets the `proxiedServerMode` package global — and neither `saveAndRestoreGlobals` nor `ensureCleanGlobalState` covers it. The leaked `true` diverts every later test's `PersistentPostRunE` into the proxied branch, so `TestPersistentPostRunE_ClearsOnErrorPath` fails deterministically in full-suite ordering:

```
go test ./cmd/bd -run 'TestProxiedCreatePolicy|TestPersistentPostRunE_ClearsOnErrorPath'
```

One-line-ish fix, verified green locally against your current head — in `interceptProxiedOpens` (covers all three tests):

```go
	t.Cleanup(func() { openProxiedServerUOWProviderFn = orig })
	// Root dispatch in a proxied workspace sets the proxiedServerMode global
	// (and saveAndRestoreGlobals does not cover it); a leaked true diverts
	// every later test's PersistentPostRunE away from the maintenance block.
	origProxied := proxiedServerMode
	t.Cleanup(func() { proxiedServerMode = origProxied })
```

I'd have pushed it directly (as with the branch refresh), but your fork is org-owned so maintainer-edit pushes 403. Push that and this merges on the standing approval — everything else is green at `9df81efcd`.

arcaven:
Thank you for running that down. That was not a diagnosis I would have reached from the failure text: the failing test and the cause were in different files, and the symptom only appears in full-suite ordering.

Pushed your fix verbatim at `1acdce6e6`, in the shared `interceptProxiedOpens` helper so it covers all three proxied create-policy tests.

Verified rather than assumed, at the pre-fix head first:

- `go test ./cmd/bd -run 'TestProxiedCreatePolicy|TestPersistentPostRunE_ClearsOnErrorPath'` reproduces `--- FAIL: TestPersistentPostRunE_ClearsOnErrorPath` exactly as you described, and passes with the fix.
- Full `cmd/bd` package run at both heads, compared by test name: 26 failures before, 25 after, zero new, and the one resolved is `TestPersistentPostRunE_ClearsOnErrorPath`. The rest are this machine's Docker and pure-Go-tag environment and fail identically at both heads.

Worth noting for whoever hits the next one: `saveAndRestoreGlobals` not covering `proxiedServerMode` is a general trap, not specific to these tests. Any future test that drives root dispatch in a proxied workspace inherits the same leak. Happy to file that as a separate issue proposing the global move into `saveAndRestoreGlobals` itself, if you would rather have it fixed at the source than at each call site. Not doing it in this PR, since that is a wider blast radius than a merge-blocker fix should carry.

arcaven:
Follow-up: main had moved since your comment, so the branch was `CONFLICTING` independently of the global-leak fix. Merged main at `c31b6bb4d` and the PR is `MERGEABLE` again.

One content conflict, in `internal/storage/uow/dolt_sql_provider.go`: both sides add a distinct field to `doltSQLProvider`, this branch's `createIfMissing` and main's `eventsJournalEnabled`. Kept both, with your `SetEventsJournalEnabled` unchanged. Nothing else conflicted.

Verified after the merge: `go build ./...` and `go vet` clean, the proxied create-policy tests plus `TestPersistentPostRunE_ClearsOnErrorPath` pass, and the full `internal/storage/uow` package passes against Docker, which exercises the merged struct rather than just compiling it.

arcaven:
Caught up to main and no longer conflicting. Main refactored this path in the interval, so it is a port rather than a textual merge, and one of your tests changed shape as a result. Flagging that explicitly rather than burying it in the diff.

Main moved the locked preparation out of the inline closure in `initSchema` into `bootstrapPreparer.prepare`, carrying the sticky `created`/`heal` state on the type. The conflict was that whole block and the resolution takes main's structure. This branch's change is re-applied inside `prepare`: probe with `DatabaseExists` before any DDL, open a database that is already there without attempting a CREATE, and refuse with the `bd init` / `bd bootstrap` message when it is missing and `createIfMissing` is off. The bare CREATE still arbitrates creation of a missing database, so the ownership proof and the heal capability are unchanged.

The test to look at is `TestInitSchemaConvergenceProbeRunsWithNoSessionDatabase`, which is new from main. It expected the locked preparation to issue a CREATE that loses with 1007. With the probe in front, an existing database is opened with no CREATE attempt at all, so that expectation is now a `SHOW DATABASES` returning the name. Everything the test asserts about the outcome is unchanged: this init did not create the database, captures no fresh-bootstrap heal authority, and still fails on the first-migration sentinel. If you would rather the probe not short-circuit the CREATE in that case, say so and I will invert it, since that is a behavior choice rather than a merge artifact.

gofmt and go vet clean; `go test ./internal/storage/uow/` passes in full.

bee-ghosttrack (changes requested):
First maintainer review, at head `11d105bc8`. The direction is right and the cmd/bd wiring is clean; two things in the storage half need a change before this lands, and the rest is non-blocking.

### Blocking

**1. `DatabaseExists` contradicts the server and the open's own existence proof.**
`ddl.go` walks `SHOW DATABASES` and compares `name == database` exactly. Dolt resolves database names case-insensitively: on a live Dolt server here, `SELECT COUNT(*) FROM information_schema.schemata WHERE schema_name = 'WYVERN'` returns 1 and `SELECT ... FROM WYVERN.issues` reads, while `SHOW DATABASES` prints `wyvern`. So a config that spells the name in a different case than the server stores it now gets the permanent `not found; run 'bd init'` refusal from a database that `USE` opens. The old `CREATE → 1007 → USE` path handled that case; this is the first probe in the tree whose false negative is a hard open failure. The repo already treats schema names as case-insensitive (`assertSessionDatabase` uses `strings.EqualFold`, with `session_database_test.go` pinning "a server that normalizes the name is not a mismatch").

Worse, `openAndInitSchema`'s fast path connects **straight to the target database** and treats the successful handshake as the existence proof; `prepare` then re-probes that already-bound session with a weaker mechanism (`SHOW DATABASES` is grant-filtered and case-exact) and, on disagreement, refuses.

Ask: reuse the probe main already has, `SELECT COUNT(*) FROM information_schema.schemata WHERE schema_name = ?` (`internal/storage/schema/converged.go`, the pre-lock convergence probe). It is parameterized, so the `_`/`%` wildcard concern the `ddl.go` comment raises does not arise, and it matches the server's own case rule. Alternatively `EqualFold` plus a short-circuit when `SELECT DATABASE()` already equals the target. Either way, add a unit test for `DatabaseExists` covering the case-variant name; today `internal/storage/domain/db` has no test for it, and a mutant flipping `==` to `EqualFold` is caught by nothing.

**2. The not-found message recommends `bd bootstrap`, which is not proxied-aware.**
`resolveRemoteCloneMode` (`cmd/bd/bootstrap.go`) picks external-server only via `IsDoltServerMode()` / shared-server env / host inference; `dolt_mode: proxied-server` is not one of those, and `executeInitAction`/`executeRestoreAction` build a `dolt.Config` with neither `ProxiedServer` nor `ServerMode` set, landing on the embedded open. A user following the message on a localhost proxied server populates `.beads/embeddeddolt` and re-runs into the same error. Either make bootstrap restore into the proxied server's database (and pin it), or drop the bootstrap clause from the message. (`bootstrap` is in `noDbCommands`, so it does not itself trip the create=false open; the problem is only that the advice is a dead end.)

### Non-blocking

3. **Test pins.** I ran mutants on the head:
   - `case !b.provider.createIfMissing:` → `case false:` is caught by exactly one test, `TestNewExternalDoltServerUOWProvider_MissingDatabaseWithCreateDisabled`, which is Docker-gated and skips silently without it. A sqlmock walk of the refusal (probe returns no row, `createIfMissing` unset, assert the message, no CREATE expected) would pin it everywhere CI runs.
   - Dropping `WithCreateIfMissing(true)` in `init_proxied_server.go` is caught (`TestProxiedCreatePolicy_InitOpensWithCreateEnabled`). Good.
   - The lock-test swap in `c61fd10b0` keeps the asserted outcome, as the commit message says, but it was the only walk of the `isDatabaseExistsError` (1007 create-race) branch; that branch and `isAccessDeniedError` now have no coverage on this path. A sqlmock case with `createIfMissing: true`, empty probe, `CREATE → 1007` restores it.
   - `cmd/bd/prime_proxied_server.go:34` opens with no options; it is the fourth proxied call site and the policy test does not cover it. (`serve.go:377` also reaches `newSQLServerUOWProvider` with the default; not a regression, since that path's DoltStore open already refuses a missing database, but worth a line in the policy test.)
4. The fast-path comment in `openAndInitSchema` ("the bare CREATE DATABASE inside the locked preparation is refused with 1007 exactly as it is on a no-database connection") now describes the old flow.
5. Hardening: `WithNoDatabaseBind` documents "per-database UnitOfWork/Tx use requires a bound database" but nothing enforces it; carrying `noDatabaseBind` on the struct and refusing in `BeginTx` would fail at the call rather than later with Error 1046.
6. On the open question in the body: the no-bind open for `clean-databases` is the right shape. I read `RunNonTx`, `Close` and the deferred events-journal activation; none touches the zeroed fields, and the non-proxied arm was never bound either. Keep it.

### Verification
- `go build ./...`, `go vet` on cmd/bd, internal/storage/uow, internal/storage/domain/db: clean.
- `go test ./internal/storage/uow/` at `2b71b9e55`: red on the four `TestInitSchemaSharedStoreGate` subtests (main's sqlmock expects the CREATE before the probe); I patched the expectations locally and got green, and your `11d105bc8` is the same fix. The full `./internal/storage/uow/` package at `11d105bc8` is green here (688 s, local dolt).
- `go test ./cmd/bd -run 'ProxiedCreatePolicy|Proxied'` and `./internal/storage/domain/db`: green.
- Live-Dolt case-sensitivity check as described in (1).
- Two adversarial subagent passes (semantics/callers, tests/mutants), findings re-verified by hand where cited.

_Claude Fable 5.1 (desk) with two Claude Opus 5 adversarial readers, on behalf of bee; cited lines and the live-Dolt check re-done by the desk before posting._

PR #5089, opened 2026-07-26
feat(dolt): make the shared pool's per-I/O read/write deadlines configurable (bd-vz0y9)

steveyegge:
## Summary

`buildServerDSN` pins the shared pool's `ReadTimeout`/`WriteTimeout` to a hardcoded 10s with no override. On a healthy local server that fast-fail is right; on an overloaded shared sql-server box the ceiling kills **ordinary** queries — link (wyvern fleet) measures ~300 `client connection went away while a query was executing` events/hour in dolt-server.log under load, including plain `bd list` and `bd mail send` — the latter meaning failure alerts for *other* bugs never got out. Today's confirmation: during a host load-avg-81 window, bridge-sync's `bd recompute-blocked` failed 24 consecutive retries on this ceiling (bd-bn8jo).

The known-long procedures are each hand-routed off the pool (`execWithLongTimeout`\* 5m, `openLongTimeoutConn` 5m, `openMigrationDB` uncapped, backup via bd-2o8nv/#5052), but the pool's own ceiling for everything else was invisible and untunable.

## Changes

- `Config.PoolReadTimeout` / `Config.PoolWriteTimeout` (0 = default 10s; **defaults deliberately unchanged** — the bug is that overloaded deployments have no knob, not the default)
- Resolution mirrors the sibling `dolt.max-conns` knob: caller override > `BEADS_DOLT_POOL_READ_TIMEOUT` / `BEADS_DOLT_POOL_WRITE_TIMEOUT` env > `dolt.pool-read-timeout` / `dolt.pool-write-timeout` config.yaml keys > 10s default
- Values accept `time.ParseDuration` strings (`90s`, `2m`) or bare seconds (`90`) via `parseTimeout`, extracted from the existing `timeoutFromEnv`
- Explicit yaml-registry entries for the two keys (the `dolt.` prefix already routed them to config.yaml)

No docs edit: docs describe the v1.1.0 pin; the knob reaches the site at the next release regeneration (same call as #5052).

## Testing

- `TestBuildServerDSN_PoolTimeouts`: defaults when unset; read-only, write-only, and both overrides carried into the DSN
- `TestParseTimeout`: duration strings, bare seconds, fallback on empty/invalid/non-positive
- `TestApplyResolvedConfig` subtests: env vars populate the fields (incl. bare-seconds form); caller-set values win over env
- Full `internal/storage/dolt`, `internal/config`, `internal/configfile` suites green; gofmt/vet/golangci-lint clean on touched packages

Validation offer: link reproduces the query-kill events on demand and offered to test patched builds (bd-vz0y9 / bd-na129).

Closes bd-vz0y9.

https://claude.ai/code/session_01KspC6Y46KKaU9ZNKupXLwk

### DIFF
--- a/internal/config/yaml_config.go
+++ b/internal/config/yaml_config.go
@@ -80,9 +80,11 @@ var YamlOnlyKeys = map[string]bool{
 	"import.path": true,
 
 	// Dolt server settings
-	"dolt.shared-server": true, // Shared Dolt server at ~/.beads/shared-server/ (GH#2377)
-	"dolt.max-conns":     true, // Connection pool size override (default 10, GH#3140)
-	"dolt.debug":         true, // Debug-mode dolt sql-server: --loglevel=debug + --prof cpu
+	"dolt.shared-server":      true, // Shared Dolt server at ~/.beads/shared-server/ (GH#2377)
+	"dolt.max-conns":          true, // Connection pool size override (default 10, GH#3140)
+	"dolt.pool-read-timeout":  true, // Pool per-I/O read deadline override (default 10s, bd-vz0y9)
+	"dolt.pool-write-timeout": true, // Pool per-I/O write deadline override (default 10s, bd-vz0y9)
+	"dolt.debug":              true, // Debug-mode dolt sql-server: --loglevel=debug + --prof cpu
 
 	// Secrets: tokens and API keys must NOT be stored in the Dolt database
 	// because that data is pushed to remotes, triggering secret-scanning
--- a/internal/storage/dolt/open.go
+++ b/internal/storage/dolt/open.go
@@ -301,6 +301,23 @@ func applyResolvedConfig(ctx context.Context, beadsDir string, fileCfg *configfi
 		}
 	}
 
+	// Pool per-I/O deadlines: env var > config.yaml > caller override > default
+	// (10s, see buildServerDSN). The default fast-fail is right for healthy
+	// local servers; overloaded shared-server deployments raise it so ordinary
+	// queries stop dying with "i/o timeout" under load (bd-vz0y9).
+	if cfg.PoolReadTimeout == 0 {
+		cfg.PoolReadTimeout = timeoutFromEnv("BEADS_DOLT_POOL_READ_TIMEOUT", 0)
+	}
+	if cfg.PoolReadTimeout == 0 {
+		cfg.PoolReadTimeout = parseTimeout(config.GetString("dolt.pool-read-timeout"), 0)
+	}
+	if cfg.PoolWriteTimeout == 0 {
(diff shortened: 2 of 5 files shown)


maphew:
Reviewed 2026-07-26 (head 2f414aba3). **Approve-with-nits** — implementation mirrors the dolt.max-conns resolution block exactly, both live open paths run `applyResolvedConfig` so the knob reaches every server-backed store, defaults unchanged, CI green. Two small fixes before merge:

1. **Precedence comment is inverted** (internal/storage/dolt/open.go, new block ~line 304): it says "env var > config.yaml > caller override > default" but the code checks `cfg.PoolReadTimeout == 0` first, so the real order is caller > env > config.yaml > default (which is what the PR body correctly claims). The wrong text is copied from the pre-existing stale comment on the max-conns block (open.go:287) — worth fixing both, since these comments are the knob's only precedence documentation.
2. **docs/reference/configuration.md needs two rows** — line 112 already documents `dolt.max-conns` in a hand-maintained table that no generator touches (check-doc-freshness.sh pins that file to other sources), so "the knob reaches the site at the next release regeneration" doesn't hold for this table; without rows the keys are undiscoverable indefinitely.

Follow-up bead rather than blocker: the long-timeout paths (`execWithLongTimeout`, `execWithLongTimeoutNoTx`, `openLongTimeoutConn`) set `ReadTimeout = 5m` unconditionally, so an operator raising PoolReadTimeout above 5m silently gets those paths *lowered* — clamp upward instead (`if cfg.ReadTimeout < longConnReadTimeout {...}`), coordinating with #5052's helper consolidation. Nit-level: the viper BD_* env aliasing gives the knob a second spelling at a different precedence tier (pre-existing, worth one docs sentence), and a `GetStringFromDir` fallback would serve library consumers that never call config.Initialize (the auto-start block a few lines down does exactly this).

steveyegge:
**Fable verdict (lion): APPROVE — merge, with one comment fix requested at merge time.**

Right-sized knob: defaults deliberately unchanged, DSN wiring mirrors dolt.max-conns, keys correctly land in YamlOnlyKeys (a timeout that only matters per-deployment has no business in the replicated DB), and the Config doc comment explicitly steers known-long operations away from leaning on this ('route through execWithLongTimeout' — exactly the right guidance to keep the relief valve from becoming a crutch). The parseTimeout extraction with the bare-number-is-seconds contract and negative/invalid-input fallback is well tested.

**The one thing to fix before/at merge** (comment only, no code change): the precedence comment in applyResolvedConfig reads 'env var > config.yaml > caller override > default', but the code — and your own 'caller-set pool timeouts win over env vars' test — implement **caller override > env var > config.yaml > default** (a non-zero cfg value short-circuits both lookups). Whoever next touches deployment config will believe the comment; make it match the test.

Non-blocking: the write-timeout knob ships untested against a real overloaded window (the bug receipts are all read-side). Fine — it's symmetric and cheap — just noting the asymmetry in evidence.

Ship it with the comment corrected.

PR #5090, opened 2026-07-26
fix(dolt): run is_blocked recomputes on the long-timeout connection (bd-bn8jo)

steveyegge:
## Summary

Both the full `is_blocked` recompute (`bd recompute-blocked` / `bd doctor --fix`) and the scoped post-merge recompute run their batched mark/unmark UPDATEs over the shared pool, whose per-I/O deadline (default 10s, `buildServerDSN`) kills any batch that outlives it. The UPDATEs carry five correlated EXISTS subqueries each, so on a loaded shared server the repair dies with `i/o timeout` / `invalid connection`.

Observed live today: bridge-sync's owed recompute against the wyvern DB failed **24 consecutive retries** during a host load-avg-81 window (wy-b72dj signature), reproduced manually twice. The two paths also compound: a pool-deadline kill of the post-merge recompute is exactly what creates the owed full recompute — which then dies the same way, so the repair never lands.

## Changes

Route both recompute transactions through `openLongTimeoutConn` (5m) — the path every other known-long maintenance op already uses (pull/merge conflict resolution, backup via bd-2o8nv/#5052, migrations):

- `RecomputeAllBlocked` (store.go): dedicated long-timeout connection for the guard + full mark/unmark pass
- `recomputeBlockedTx` (post-merge scoped recompute): same treatment

The trailing `DOLT_ADD`/`DOLT_COMMIT` stays on the pool: it stages a small computed diff, and the pool's ceiling is independently tunable via the bd-vz0y9 knob (#5089 — sibling PR, no code overlap conflict; both touch `store.go` in disjoint regions).

## Risk notes

- No session-state coupling: both recomputes already ran in their own transactions after the outer operation committed; the working set is server-global, and the established pattern (pull on long conn → commit on pool) is unchanged.
- Embedded mode unaffected — `EmbeddedDoltStore` has its own `RecomputeAllBlocked` with no network deadline.

## Testing

- Full `internal/storage/dolt` suite green (includes blocked_consistency, blocked_merge, blocked_recompute_guard tests exercising both changed paths)
- gofmt/build clean
- Field validation before the fix: direct no-timeout SQL probes of the exact mark/unmark predicates confirmed the failing owed recompute was a no-op *that run* (0 stale rows) — but only because local writes had kept flags consistent; a conflicted-pull window would have left `is_blocked` stale with no working repair path.

Closes bd-bn8jo.

https://claude.ai/code/session_01KspC6Y46KKaU9ZNKupXLwk

### DIFF
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -3438,7 +3438,18 @@ func (s *DoltStore) recomputeBlockedAfterPull(ctx context.Context, fromCommit st
 // operator resolved by hand — leaves is_blocked stale until this full pass runs.
 // Idempotent: a consistent database corrects nothing.
 func (s *DoltStore) RecomputeAllBlocked(ctx context.Context) (int, error) {
-	tx, err := s.db.BeginTx(ctx, nil)
+	// The full pass's batched UPDATEs carry five correlated EXISTS subqueries
+	// each; on a loaded shared server a single batch can outlive the pool's
+	// per-I/O deadline (default 10s, see buildServerDSN), killing the repair
+	// with "i/o timeout" — and the retry dies the same way, so the owed
+	// recompute never lands (bd-bn8jo). Run it on a dedicated long-timeout
+	// connection like the other known-long maintenance ops.
+	db, err := s.openLongTimeoutConn()
+	if err != nil {
+		return 0, err
+	}
+	defer db.Close()
+	tx, err := db.BeginTx(ctx, nil)
 	if err != nil {
 		return 0, fmt.Errorf("begin is_blocked recompute: %w", err)
 	}
@@ -3470,9 +3481,17 @@ func (s *DoltStore) RecomputeAllBlocked(ctx context.Context) (int, error) {
 }
 
 // recomputeBlockedTx runs the post-merge is_blocked recompute in its own
-// transaction.
+// transaction. Like RecomputeAllBlocked it runs on a long-timeout connection:
+// a heavy merge scopes the recompute over a large diff, and a pool-deadline
+// kill here is what turns into the owed full recompute in the first place
+// (bd-bn8jo).
 func (s *DoltStore) recomputeBlockedTx(ctx context.Context, fromCommit string) error {
-	tx, err := s.db.BeginTx(ctx, nil)
+	db, err := s.openLongTimeoutConn()
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+	tx, err := db.BeginTx(ctx, nil)
 	if err != nil {
 		return fmt.Errorf("begin is_blocked recompute: %w", err)
 	}


maphew:
Reviewed 2026-07-26; re-verified the central finding today at head 054c161d2. **Needs-changes** — the problem is real and the long-timeout routing matches precedent in this file, but there's one blocker:

**Blocker — the fresh sessions are never checked out to the store's active branch.** Both patched paths do `db, err := s.openLongTimeoutConn()` and use it directly. `withReadTxLongTimeout` (store.go:865-887) documents exactly this hazard for the read path ("the fresh one-shot connection defaults to the default branch…"), and `beginIgnoredTxOnBranch` (transaction.go:152-158) repeats it for writes. When the pool session is on a non-default branch (Store.Checkout, the federation staging-branch window at federation.go:432-477, branch-scoped transactions): (a) `GuardBlockedRecomputeWorkingSet` queries session-scoped `dolt_status` against the wrong branch's clean working set, silently passing the dirty-graph release-safety guard; (b) the recomputes then read and UPDATE the *default* branch's issues — a silent cross-branch write; (c) the trailing `doltAddAndCommit` runs on the pool (correct branch), stages nothing, and swallows it via `isDoltNothingToCommit`, so the operator sees a nonzero "corrected" count while their branch is unrepaired. Fix: extract the branch-reproduction block from `withReadTxLongTimeout` (store.go:872-887) into e.g. `openLongTimeoutConnOnActiveBranch(ctx)` (SELECT active_branch() on s.db → CALL DOLT_CHECKOUT on the new conn, s.branch fallback) and use it in both recompute paths.

**Major — the stated test evidence can't be right as described.** The suite's tests run on per-test isolated branches (dolt_test.go:132 → testutil.StartTestBranch, raw DOLT_CHECKOUT on store.db without updating s.branch), so with this patch TestRecomputeAllBlocked_RefusesDirty* should stop seeing a dirty working set and _RepairsAndCommitsWhenClean should see changed==0 — the suite should have *failed*. Likely explanation: TestMain skipped the package (no Dolt container). CI green isn't evidence either — pr-risk.yml runs embedded + proxied-cmd shards, not the server-backed internal/storage/dolt package. After the branch fix, please re-run `go test ./internal/storage/dolt -run 'Blocked'` against a live Dolt server. A regression test in the shape of history_test.go:30-96 (corrupt connStr to prove routing; sanity phase asserts the fresh conn reads the checked-out branch) would pin both properties.

Minor/nit: the guard-then-stage pair now spans two sessions where MaxOpenConns(1) previously made it tight — worth a sentence in the comment; and the bare `return 0, err` on open failure could name the operation (`fmt.Errorf("open long-timeout conn for is_blocked recompute: %w", err)`) like the sibling sites. Coordinate the helper consolidation with #5052, whichever lands second.

steveyegge:

Correct diagnosis and the established cure: the is_blocked recompute batches (five correlated EXISTS subqueries each) are exactly the 'known-long operation' class the pool's 10s per-I/O deadline exists to protect *against*, not to kill — and the failure compounding (the deadline kill is what creates the owed full recompute, which then dies the same way) makes this a convergence bug, not a performance nit. Moving both surfaces — the full pass AND recomputeBlockedTx, whose post-merge kill is the upstream source of the debt — is the complete fix; patching only RecomputeAllBlocked would have left the debt factory running.

One consolidation note, non-blocking: once #5052 merges, store.go will carry both its `oneShotConn(readTimeout)` and the `openLongTimeoutConn` this PR leans on — two spellings of the same mechanism. Worth a small follow-up to make openLongTimeoutConn delegate to oneShotConn(5m) (or rename), so the next long-op author has one obvious door. Pairs with the #5089 guidance comment that already points known-long ops here.

Ship it.

Issue #4512, opened 2026-06-29
Publish a stability/compat policy for the dolthub/dolt/go pseudo-version pin in go.mod


sjarmak:
**Problem.** `go.mod` (origin/main:135) pins `github.com/dolthub/dolt/go` at the pseudo-version `v0.40.5-0.20260605230755-1bf533220ab0` — an untagged commit (`1bf533220ab0`, 2026-06-05, after tag `v0.40.5`) of the upstream Dolt module, carried as a **direct** dependency. Downstream consumers that vendor beads as a library (Gas City) inherit this exact Dolt commit transitively, but have no documented contract for when it moves or how to reason about compatibility. A routine beads minor/patch bump can silently advance the embedded Dolt engine across a non-trivial commit range.

_(Framing note: `dolthub/dolt/go` is the upstream Dolt module itself, not a beads-owned fork. The fork-ownership concern that does exist is for `go-mysql-server`/ICU, tracked separately in #3303 — a different module and problem. This issue is specifically about the version *policy* of the upstream Dolt pin.)_

**Root cause.** The pin is a pseudo-version (not a release tag) and there is no policy statement in `CHANGELOG.md`, `CONTRIBUTING.md`, or `docs/` (`git grep dolt/go|pseudo-version|pin` returns nothing relevant). `.github/dependabot.yml` enables `gomod` updates with no ignore/group rule for `dolt/go`, so the pin can move via an automated bump with no changelog signal to consumers.

**Why it matters to a downstream consumer.** Gas City pins beads and runs a long-lived shared dolt sql-server; an unannounced jump in the embedded Dolt engine (storage-format, merge-semantics, or sql-server-flag changes) can surface as runtime drift in the gc-managed server with no upstream changelog line to trace it to. Today the only way to detect a Dolt engine move is to diff `go.mod` between beads tags.

**Dedup.** `dolt/go pin stability policy` → 0 results; `go.mod dolt fork pseudo-version` → only #3303 (closed, go-mysql-server build-tag/fork policy — related domain, not this module's version policy). No in-flight policy doc.

[...]

PR #4907, opened 2026-07-19
Guard and qualify historical upgrades to v1.2

julianknutsen:
## What

This PR establishes an explicit, fail-closed boundary for historical Beads
workspaces and proves the first direct historical-schema translation in the
current binary.

### Upgrade contract

- Recognized historical SQLite and legacy server-Dolt workspaces are refused
  before config migration, version tracking, init, bootstrap, doctor, or store
  construction can mutate them.
- The sealed-copy bridge and checksum-pinned migration harness qualify a fixed
  corpus of 13 releases: SQLite v0.9.1, v0.17.0, v0.49.6, and v0.50.3; legacy
  server Dolt v0.55.4, v0.56.1, v0.57.0, and v0.62.0; and embedded Dolt
  v0.63.3, v1.0.0, v1.0.1, v1.1.0, and v1.1.2. The authentic v0.55.4
  blank/default-mode layout is covered separately.
- Historical upgrades remain explicit. This PR does not make first command use
  mutate an old workspace automatically.

### First direct-reader slice

The current binary now exposes:

```text
bd migrate legacy-sqlite --source-db PATH --output PATH|-
```

For the authenticated v0.49.6/v0.50.3 SQLite schema family, this command:

- verifies the exact historical release marker, table definitions, and foreign
  keys before reading records;
- reads only a private sealed database/WAL copy and verifies that the source
  fingerprint did not change while sealing;
- translates issues, labels, dependencies, and comments to current canonical
  JSONL without executing a historical export binary;
- publishes output only after the complete conversion succeeds; and
- rejects schema drift or values the current importer would default,
  truncate, deduplicate, or otherwise lose.

Legacy timestamps are canonicalized to UTC using nearest-second rounding,
matching the current `DATETIME(0)` storage contract. Values that would round
outside the current range or trigger importer defaults are rejected. The
authenticated upgrade lanes verify timestamp equality after import into a
fresh current database and re-export.

### Final maintainer repairs

- The public direct bridge invokes the current authenticated reader against the
  sealed database; its regression fake exits if the historical generic
  exporter is called.
- No-store commands default to historical-workspace admission before their
  command body runs. Only static/global commands, specialized repair commands,
  and exact `migrate legacy-sqlite` are exempt.
- Diagnostic config reads remain nonmutating, and embedded-current precedence
  uses only a nonempty coarse `.dolt` marker behind the embedded storage
  adapter boundary.
- The legacy reader projection is aligned with the current 77-destination scan
[...]

PR #5092, opened 2026-07-26
feat(doltversion): dolt CLI discovery, hardened version probe, and warn-only version policy (contract part 1)

maphew:
## Why

Managed proxied-server mode spawns an external `dolt` binary that today is discovered by a bare `exec.LookPath("dolt")` with no override anywhere, no version awareness beyond the advisory archive-level probe, and poor failure behavior: a missing dolt errors *after* `bd init` has already created `.beads/`, and an incompatible one is spawned blind and dies as "exited before listener became ready" with the cause buried in server.log.

This is part 1 of a supported install/version contract for that binary (companion to the workspace-gate track, and distinct from the Go-module pin question in #4512 — this is about the CLI executable). The design went through an adversarial review pass whose two biggest reversals are baked in here:

- **No hard version floor.** A 2.2.0 floor was policy-without-evidence; the empirically established boundary (cross-version matrix, 2026-07-25: `dolt` < 2.0.0 cannot read module-written storage; 1.52.1 fails serve and read) becomes a **warn-only** `RecommendedMin = 2.0.0`. The only hard failures are a missing, non-executable, or probe-failing binary; even an unparseable dev-build version string is a warning, not a refusal.
- **Committed repo data never selects an executable.** Resolution order is `BEADS_DOLT_BIN` env → clone-local gitignored sidecar (hook point only in this PR; the sidecar setting itself lands with identity threading in part 2) → PATH. An explicitly-named broken binary is an error, never a silent fall-through to PATH, and explicit values are absolutized before validation (a bare relative name would otherwise validate the cwd file but exec a PATH lookup).

## What

New stdlib-only leaf package `internal/doltversion`:

- **Hardened probe** of `<binary> version`: timeout + process kill (`WaitDelay`), 64KiB output caps per stream, regular-file/exec-bit validation, unconditional symlink resolution (given and real path both recorded), errno-based exec-format detection that names a likely architecture mismatch, and file identity (size/mtime) captured for later change detection.
- **Version parsing** generalized from the archive-level probe's parser, tolerating prerelease/dev suffixes; canonical `errors.Is`-able error taxonomy (`ErrNotFound` / `ErrNotExecutable` / `ErrProbeFailed` / `ErrUnparseableVersion`).
- **Policy**: `ProbeWithPolicy` composes probe + recommended-version check into one call returning `(Identity, *Warning, error)`.

Wired at two entry points:

- `cmd/bd/uow_factory.go`: the managed proxied provider replaces its bare LookPath with the hardened probe. **Correction (2026-08-11, from review):** the original claim here — "no longer forks `dolt version` twice" — inverted in the re-merge: main's 68afa3d51 removed the archive-level probe consumer, so relative to current main the eager probe *added* one fork per command. The review-fix push resolves this properly with a cross-process probe cache keyed by (path, size, mtime): the steady state is now zero forks (one stat), re-probing only when the binary on disk changes. The recommendation warning prints at most once per day (warn stamp in the same cache) and respects `--quiet`/`--json`.
- `cmd/bd/init_proxied_server.go`: proxied init preflights the binary **before any `.beads/` write**, so a missing/broken dolt fails cleanly instead of after workspace creation. External proxied config is exempt (bd never spawns that server).

Docs: `BEADS_DOLT_BIN` added to the configuration reference; troubleshooting entry for the version recommendation.

## Deliberately out of scope (part 2)

- Threading the resolved identity through every spawn site — `doltserver.Start` still does its own `exec.LookPath("dolt")`, so `BEADS_DOLT_BIN` is honored by proxied mode but not yet by `bd dolt server start`. Part 2 threads one resolved identity everywhere, adds the clone-local sidecar setting, extends the pidfile with launched-binary identity, and consolidates `SupportsArchiveLevelConfig` onto this package.
- Any hard floor enforcement (needs the full CI-grade compatibility matrix first).

## Testing

24 package tests (`-race`), including: resolution precedence and the bare-name absolutization regression; probe stubs for hung (kill proven by wall-clock), garbage-output, over-cap-output, non-executable, symlinked, and missing binaries; warning demotion for unparseable versions vs hard errors for broken binaries; direct output-cap truncation test; parser tables incl. multi-line output and prereleases. `go build`/`go vet` clean; full `make test` queued in our local verification pipeline. Cross-vendor pre-review (Claude + GPT-5.6 review agents) applied before opening: absolutization, warning gating, duplicate-fork elimination, errno-based exec-format detection, and the unparseable-version demotion all came from that pass.

_claude-fable-5-medium on behalf of matt wilkie_


maphew:
Reviewed at `a84b9b5e`.

The leaf package and archive-level fail-closed logic look solid, and focused tests pass. Three fixes remain before merge:

- `init_proxied_server.go` must load persisted external proxied configuration before deciding that a local Dolt binary is required.
- Reuse the successful preflight identity/path during provider construction. A second `dolt version` probe can currently fail after `.beads` is written, defeating the pre-write guarantee.
- Remove the troubleshooting claim about clone-local sidecar resolution until that deferred wiring exists.

This PR is the right place for the binary preflight work; it overlaps #4907/#5100 only in historical-upgrade qualification, so keep the concerns separated and rebase after that stack settles. Then rerun CI.

maphew:
Rebased-by-merge onto current `main` — this PR was 255 commits behind and `CONFLICTING`, so `gh pr update-branch` could not resolve it. It is `MERGEABLE` again as of 3f0c5d72b.

One file conflicted, `cmd/bd/uow_factory.go`, because this PR and an upstream refactor changed the same function. Both sides are kept:

- **Ours:** `doltVersionWarnOnce`, `resolveAndProbeDolt`, `doltArchiveLevelSupported`.
- **Upstream:** `sqlServerUOWTopology`, `previewProviderOptions`, the `identityPosture` assert/adopt split, `newProxiedServerUOWProvider` / `...Adopting` / `openProxiedServerUOWProvider`, and `newManagedProxiedServerUOWProvider`'s new topology-struct signature.

**The one line worth checking on review** is in `newManagedProxiedServerUOWProvider`: it takes upstream's new signature and body, but the dolt lookup is ours — `resolveAndProbeDolt`, not `exec.LookPath`. A merge that let `LookPath` come back would have compiled and passed tests while quietly defeating the point of this PR.

**One behavioural change reviewers should know about, introduced by the merge rather than by the original PR:** upstream's server-mode path (`bd serve` → `newSQLServerUOWProvider`) did not exist when this PR was written, and it now also routes through `newManagedProxiedServerUOWProvider`. So the hardened probe's blast radius has widened — `bd serve` now forks a 10s-bounded `dolt version` probe where upstream only did `exec.LookPath`, and a broken/unresolvable dolt will fail `bd serve` rather than surfacing later. That is consistent with this PR's intent (fail loudly on an explicitly-named broken override), but it is new surface area and worth an explicit ack before merge. The error label was hardcoded to `newProxiedServerUOWProvider`, which would have reported a `bd serve` failure as a proxied-server one; it now names the function itself.

Verification: `gofmt` clean; `go build -tags gms_pure_go ./...` and `go vet` clean; the `cmd/bd` targeted suites (`TestUOW|TestProxied|Topology|Preview|ArchiveLevel`) and `./internal/doltversion/...` pass. Independent review confirmed the merged tree is exactly `upstream/main` plus this PR's patch: the diff against `upstream/main` is the same 14 files with the same per-file line counts as the diff against the merge base, +6 lines for the error-label comment above.

Not queued for auto-merge; the widened `bd serve` blast radius above deserves a maintainer ack first.

maphew:
Updated: merged current `main` (7e825441b) and resolved the drift. Three notes for reviewers:

1. **Archive-level adapter removed.** Main's 68afa3d51 deliberately deleted the `archive_level` gating/reconcile machinery from proxied-server mode, which was this PR's only consumer of the probed identity. The merge drops the `doltArchiveLevelSupported` adapter and its two tests rather than resurrecting ~200 deleted lines, and narrows `resolveAndProbeDolt` to return only the resolved binary path — part 2's revalidation consumer re-widens it when it lands. Net effect of this PR is now: discovery (env > sidecar hook > PATH), the hardened probe, and the warn-only version advisory.
2. **macOS CI failure fixed.** The one red check on the previous head was this PR's own probe tests: macOS's `/var/folders` temp root is a symlink and the tests compared `RealPath` against the raw `t.TempDir()` path. Fixed by resolving symlinks in the expectations (ef1898094); reproduced and verified against a symlinked `TMPDIR` on Linux.
3. **Worth knowing:** with the archive-level gate gone from main, `newManagedProxiedServerUOWProvider` forking `dolt version` again is exactly the path 68afa3d51 stripped it from. Here its only effect is the once-per-process warn-only advisory (plus failing hard on a non-executable/unprobeable binary, 10s timeout). If that posture is unwanted in proxied mode, the probe call is trivially droppable from that one call site without touching the rest of the contract.

Build, vet, `unparam`/golangci (CI's exact new-from-merge-base invocation), `internal/doltversion`, and the targeted `cmd/bd` proxied-server tests are all green locally with `-tags gms_pure_go`. Review appreciated.

steveyegge (changes requested):

Fable adversarial review (bee, beads-lane steward). This is a well-built, unusually well-reasoned PR — the probe package is genuinely hardened (timeout + WaitDelay kill proven by wall clock, per-stream 64KiB caps that never short-write the child, errno-based exec-format detection, symlink resolution, the bare-name absolutization fix with a real decoy regression test), the error taxonomy is clean and `errors.Is`-able, and warn-only correctly writes stderr gated on `--quiet`/`--json` so scripting stays clean. Verified on main (4e7cc88) in an isolated worktree: `go build ./...`, `go vet`, windows/amd64 cross-compile, and all 24 package tests green under `-race` (hermetic shell stubs, no real dolt needed). Prior art checked: the only existing probe is `gc_config.go`'s memoized `doltVersionAtLeast`, which this generalizes and explicitly defers consolidating — no unstated duplication. Three fixes before merge:

**1. should-fix — a per-command `dolt version` fork lands on the hot path, uncached, despite the type being designed for caching.** `newManagedProxiedServerUOWProvider` is reached from the store-open path — essentially every `bd` command in managed proxied mode — and `resolveAndProbeDolt` → `Probe` forks `<dolt> version` every time (measured 40–80ms with real dolt 2.2.3). Main today does only `exec.LookPath` (~µs) here, so relative to main this PR *adds* one subprocess per command; `bd init --proxied-server` forks twice by its own comment. The irony: `Identity`'s doc comment says the size/mtime fields exist so callers can cache probe results — but no caller caches anything; only the *warning* is deduped, not the fork. Minimal fix: probe only when the server actually needs spawning (an already-running server proves the binary spawnable), or memoize by (path, size, mtime) exactly like `archiveLevelSupportCache` already does.

**2. should-fix — Windows: explicit `BEADS_DOLT_BIN` without `.exe` fails ErrNotFound, and the platform branches ship untested.** `validateExplicitPath` uses raw `os.Lstat` — no `.exe`/PATHEXT completion, unlike `exec.LookPath` — so `BEADS_DOLT_BIN=C:\tools\dolt` reports "not found" while `dolt.exe` sits right there; neither code nor `docs/reference/configuration.md` says the extension must be spelled. Meanwhile every behavioral test calls `skipOnWindows`, so `windowsBadExeFormat = 193` and the exec-bit-skip branch have zero runtime coverage on the platform they target. `.cmd`/`.bat` stubs would make most of these tests runnable there. Pointed, given you're the project's Windows contributor — this is your home turf.

**3. should-fix — the warning fires on every command forever with no acknowledgment path.** Once-per-process is once-per-*command* for a CLI: an operator on old dolt sees the identical advisory on every `bd` invocation, trainable-to-ignore. A once-per-day stamp file or an ack knob fixes it; at minimum the troubleshooting entry should say it recurs. Folds naturally into fix 1.

**Nits:** the install hint wraps the deliberately-distinguished cancellation error — Ctrl-C during startup prints exactly the misleading "reinstall dolt" hint the package's own comment warns against (branch on `errors.Is(err, context.Canceled)` at the call site); `ParseVersion` accepts any first line ending in an integer ("dolt build 2023" → version "2023", silently *passing* the 2.0.0 floor — fine for warn-only, but this becomes the enforcement parser in part 2); the body's claim that the provider "no longer forks `dolt version` twice" inverted in the re-merge — the archive-level consumer was removed upstream, so net effect vs main is **+1** fork, not −2; correct the body so the merge record doesn't claim a perf win that inverted; minor taxonomy drift between Probe's stat branch (all→ErrNotFound) and validateExplicitPath's (non-ENOENT→ErrNotExecutable), and Probe's 10s timeout only applies when the outer ctx has no deadline.

**Verified clean for the record:** explicit-broken-path never falls through to PATH; typed-nil Warning trap avoided; init preflight correctly scoped; resolved doltBin genuinely reaches the spawn; dead-code risk low — wired at two real entry points, part-2 deferrals explicitly declared.

With 1–3 in, this merges on green — ping me on the re-push.

maphew:
All three fixes in, at `6c3d60d55` + `280908b5f` — pinging as asked. Package tests green under `-race`, `golangci-lint` clean, and the new Windows CI job in this push executes the platform branches on the PR run itself.

**1 — the hot-path fork is gone in the steady state, via the memoization option.** `ProbeWithPolicyCached` fronts the fork with a small JSON cache in the user cache dir, keyed by exactly the `(RealPath, FileSize, FileModTime)` fingerprint `Identity` documents for this purpose. A fingerprint hit costs one stat; any change to the binary on disk misses and re-probes; hard probe errors are never cached (a broken binary re-fails loudly on every command); every cache failure direction — no cache dir, corrupt file, failed write — fails open to the uncached behavior. I took memoization rather than probe-at-spawn deliberately: the spawn decision lives inside `proxy.GetCreateDatabaseProxyServerEndpoint`, and threading the probe down that layer crosses the boundary this PR intentionally leaves to part 2's launch-time revalidation — while the cache already achieves the same steady-state cost (zero forks). One honest consequence to name: a binary that breaks on disk while a detached server is still healthy fails commands at preflight rather than riding the live endpoint — that is this PR's existing fail-loudly posture, unchanged, and part 2's revalidation is the right place to relocate it if you want that changed.

**2 — Windows is now home turf, not a skip.** `validateExplicitPath` paths get PATHEXT completion via `exec.LookPath` when the exactly-spelled file doesn't exist (`BEADS_DOLT_BIN=C:\tools\dolt` finds `dolt.exe`; the exact spelled path wins when both exist; a decoy regression test pins that), and `docs/reference/configuration.md` documents it. The behavioral tests now run on Windows through `.cmd` stubs — `skipOnWindows` survives only where the subject genuinely doesn't exist there (the exec-bit check, symlink creation) — and a focused `windows-latest` job (same pattern as `test-windows-liveness`) runs the package at PR time, so `windowsBadExeFormat` and friends stop shipping untested. On the "can `CreateProcess` even run a `.cmd` directly" question my second-vendor reviewer raised: yes — implicitly spawning `cmd.exe` for batch files is long-standing `CreateProcess` behavior (it is the entire mechanism behind the 2024 BatBadBut advisory class), the stub arguments here are static, and this PR's own Windows job is the live proof either way.

**3 — the advisory now repeats at most once per day,** via a warn stamp in the same cache file, instead of once per process (= once per command for a CLI). In-process `sync.Once` dedup remains for multi-provider commands; the troubleshooting entry states the cadence and that upgrading the binary re-probes immediately.

**Nits:** Ctrl-C during the probe no longer prints the misleading install hint (the call site branches on `context.Canceled`, as the package's own comment warned); Probe's stat branch adopts `validateExplicitPath`'s not-found/not-executable taxonomy split; `ParseVersion` now documents the trailing-integer laxity ("dolt build 2023" → "2023") with the tightening required before it becomes part 2's enforcement parser; and the PR body's fork-count claim is corrected in place — you were right that it inverted in the re-merge, and the body now records +1-then-fixed rather than the phantom −2.

Second-vendor round on the fix itself also surfaced two things now fixed: `bd init`'s **local** `--quiet` shadows the persistent flag, so the advisory ignored `bd init --proxied-server -q` — `resolveAndProbeDolt` takes quiet as a parameter now; and the troubleshooting entry advertised the clone-local sidecar resolution source that this PR only stubs (`SidecarValue` is always empty), so that claim is deferred to part 2 where the reader lands.

maphew:
Follow-up: one more commit, `6ba5c0ea7`. The new Windows CI job caught the counting stub double-counting forks — cmd.exe includes the space before a mid-line `>>` in the written text, so `set /p =x >> file` appended two bytes per invocation and the byte-counting `forkCount` read every fork as two. All four Windows failures were exactly doubled expectations; the cache itself hit and skipped forks correctly. The commit switches the stub to a redirection-first batch line; production code untouched. All 104 checks are green on this head.

steveyegge:
## Review verdict: MERGE

Fable adversarial re-review (bee, beads-lane steward), following up my 2026-08-10 MERGE-AFTER-FIXES. All three should-fixes are verifiably in, each with tests that would catch a regression, and all 104 checks are green on this head:

1. **Hot-path fork — fixed.** `ProbeWithPolicyCached` keys on exactly the `(RealPath, FileSize, FileModTime)` fingerprint `Identity` documents for caching; steady state is one stat, zero forks. The fork-counting stub tests are the right proof shape — they'd catch a silent cache bypass, not just wrong output. Hard-errors-never-cached is the correct fail-loud choice, and recomputing `CheckRecommended` on cache hits means a future `RecommendedMin` bump needs no cache invalidation. Every failure direction fails open to uncached behavior, verified against the test set.
2. **Windows — fixed, and the new CI job already earned its keep** by catching the counting-stub double-count on the PR run itself (6ba5c0ea7). PATHEXT completion with exact-spelled-path-wins is pinned by a decoy test, `.cmd` stubs run the behavioral suite where it targets, and `skipOnWindows` survives only where the subject genuinely doesn't exist there.
3. **Warning cadence — fixed.** Once per day via the cache's warn stamp, fail-open toward showing it, docs state the cadence.

I also re-verified the codex round-1 items (external-config exemption matches main's own `externalConfig == nil` convention at init, and runtime external workspaces route around the probe entirely), that `doltBin` genuinely flows into `uow.NewDoltServerUOWProvider` so the `BEADS_DOLT_BIN` doc claim holds, and that the #5703 boundary is clean — the only part-2 artifact in this diff (`SidecarValue`) is inert and labeled.

Non-blocking findings, suggest folding into #5703:

- **Bare separator-free `BEADS_DOLT_BIN` resolves to $CWD** (`Resolve` + `filepath.Abs`, pinned by `TestResolveAbsolutizesBareEnvValue`). The absolutization correctly fixed the approve-one/exec-another split, but it re-enables the cwd-execution semantics Go's own `exec.ErrDot` exists to prevent: `BEADS_DOLT_BIN=dolt` plus a hostile repo containing an executable `dolt` runs repo content. Cleanest fix in part 2: refuse separator-free values with a clear error.
- **Single-entry machine-wide cache**: two workspaces alternating different dolt binaries regress to one fork per command, and each overwrite zeroes the warn stamp — reintroducing the per-command warning in exactly that configuration. Keying the cache filename by a hash of the resolved path fixes both. Accepted-by-design today per the code's one-binary-per-machine note.
- **Probe inherits bd's cwd** — `cmd.Dir = os.TempDir()` would make `dolt version` behavior environment-invariant for free (today's layout can't realistically bite, since bd's cwd isn't itself a dolt repo).
- **Windows corner**: a *directory* at the exact `BEADS_DOLT_BIN` path shadows a completable sibling `.exe` (Lstat short-circuit in `completeExecutableExt`); the resulting "is a directory" error is loud, so just a sharp corner of the correct exact-file-wins rule.
- Doc nit: `validateExplicitPath`'s "never from PATH" comment is outdated — `Probe` and `fingerprintMatches` apply it to PATH-resolved binaries too.

One behavior change to record for the log, disclosed in the thread and deliberate: a dolt that spawns fine but whose `version` subcommand fails or exceeds 10s now hard-fails managed-proxied commands and `bd serve` at preflight (main required only LookPath), and a binary that breaks on disk under a healthy detached server fails at preflight rather than riding the live endpoint. Both are the PR's fail-loudly posture, never cached, with part 2's launch-time revalidation as the named relocation point.

Merging on this verdict.

PR #5099, opened 2026-07-27
feat(gate): add --title flag to bd gate create

jacobhausler:
## Summary
- `bd gate create` always names an ad-hoc gate `Gate: <type>` (or `Gate: <type> <await-id>`), which is unfindable in a queue listing of many gates from the title alone.
- Adds an optional `--title` flag to override the generated title. When omitted, behavior is unchanged.

## Test plan
- [x] `go build ./cmd/bd` succeeds
- [x] `go test ./cmd/bd/...` passes (Dolt-server-dependent tests skip, as they do without this change)
- [x] Manual: `bd gate create --blocks <id> --title "Gate: custom"` creates a gate with the custom title (verified via `bd show`)
- [x] Manual: `bd gate create --blocks <id>` (no `--title`) still produces the default `Gate: <type>` title

### DIFF
--- a/cmd/bd/gate.go
+++ b/cmd/bd/gate.go
@@ -302,7 +302,8 @@ Examples:
   bd gate create --blocks bd-abc
   bd gate create --type=human --blocks bd-abc --reason="Need design review"
   bd gate create --type=timer --blocks bd-abc --timeout=2h
-  bd gate create --type=gh:pr --blocks bd-abc --await-id=42`,
+  bd gate create --type=gh:pr --blocks bd-abc --await-id=42
+  bd gate create --blocks bd-abc --title="Gate: awaiting owner sign-off"`,
 	SilenceUsage:  true,
 	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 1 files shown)


maphew (approved):
Approving. Sorry this sat unanswered — that is a queue-rhythm problem on our side, not anything about the change.

The change is right and minimal. Precedence is the part that matters and you got it in the correct order: the derived `Gate: <type>` and `Gate: <type> <await-id>` titles are computed first and `--title` overrides both, so an explicit title always wins regardless of whether `--await-id` is set. The help-text default you documented (`Gate: <type>`) matches what the code actually produces, the new example line is consistent with the others, and `--title` does not collide with any existing flag on this command.

One non-blocking gap: there is no test. `cmd/bd/gate_embedded_test.go:466` already asserts the derived title (`"Gate: gh:pr 42"`), so a `--title` case sits naturally right beside it — the useful assertion is that `--title` wins even when `--await-id` is supplied, since that is the interaction a future refactor could silently break. Happy to add it maintainer-side after this merges rather than sending you another round trip; say the word if you would rather add it here.

PR #5101, opened 2026-07-28
fix: preserve forced delete dry-runs and payload-blind previews

idirectships:
## Summary

- Propagate `--force` to single and batch delete dry-run previews without mutating records or dependencies.
- Keep quiet output empty and JSON previews structural-only across embedded and proxied-server output boundaries.
- Add hermetic regressions for quiet/JSON payload blindness and forced dry-run nonmutation.

## Verification

- `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run '^TestEmbeddedDelete$' -count=1 -v`\n- `CGO_ENABLED=1 go test -tags 'gms_pure_go integration' ./cmd/bd -run '^(TestDeletePreviewJSONIsPayloadBlind|TestDeleteBatchDryRunHonorsForce|TestOutputDeleteProxiedPreviewIsPayloadBlind|TestEmbeddedDelete)$' -count=1 -v` (payload-output cases passed; storage-backed cases skip when the local Dolt/Docker fixture is unavailable)\n- `CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run '^$' -count=1` and `CGO_ENABLED=1 go build -tags gms_pure_go ./cmd/bd`\n\n## Release request\n\nPlease include this contribution in the next stable canonical release. Before publishing, verify the release tag commit contains the eventual canonical merge commit for this PR (and, during review, this branch head `4fcfd55f4da416c13f13bbc1723c896bcaafedf4`) with `git merge-base --is-ancestor`. The release should use the existing canonical native `darwin/arm64` path and publish `beads_<version>_darwin_arm64.tar.gz` with its SHA-256 row in `checksums.txt`; install only through the checksum-verifying installer. Preserve the currently installed CLI as a recoverable regular file and record its version/embedded commit/SHA-256 before replacement; roll back by restoring that preserved file and re-verifying it.\n\n## Scope and negative proof\n\nThis PR carries only the five approved delete-path repair files, from a fresh canonical `main` base `dbbf3a9618aacaab20427f7bc1f7b35b6edab3ba`; it does not modify release workflows, version/changelog, credentials, permissions, signing, publishing, packages, or runtime data. It does not create a tag or release.

### DIFF
--- a/cmd/bd/delete.go
+++ b/cmd/bd/delete.go
@@ -133,7 +133,20 @@ Force: Delete and orphan dependents
 		idPattern := `(^|[^A-Za-z0-9_-])(` + regexp.QuoteMeta(issueID) + `)($|[^A-Za-z0-9_-])`
 		re := regexp.MustCompile(idPattern)
 		replacementText := `$1[deleted:` + issueID + `]$3`
-		if !force {
+		if dryRun || !force {
+			var previewResult *types.DeleteIssuesResult
+			if dryRun {
+				previewResult, err = activeStore.DeleteIssues(ctx, []string{issueID}, false, force, true)
+				if err != nil {
+					if previewErr := outputDeletionPreview([]string{issueID}, map[string]*types.Issue{issueID: issue}, false, true, previewResult, err, jsonOutput); previewErr != nil {
+						return previewErr
+					}
+					return HandleError("previewing deletion: %v", err)
+				}
+			}
+			if jsonOutput || isQuiet() {
+				return outputDeletionPreview([]string{issueID}, map[string]*types.Issue{issueID: issue}, false, dryRun, previewResult, nil, jsonOutput)
+			}
 			fmt.Printf("\n%s\n", ui.RenderFail("⚠️  DELETE PREVIEW"))
 			fmt.Printf("\nIssue to delete:\n")
 			fmt.Printf("  %s: %s\n", issueID, issue.Title)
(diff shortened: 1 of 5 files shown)


maphew (approved):
Approving. I verified the claims against the code rather than the description, with an independent cross-vendor review as a second pass, and both hold up. One note on framing first, because I think this PR undersells itself.

## This is a data-loss fix, not just an output fix

On current `main`, the single-issue delete path gates the preview on `if !force` (`cmd/bd/delete.go:136`) and the destructive branch below it never consults `dryRun`. So `bd delete <id> --force --dry-run` **actually deletes the issue** — `--dry-run` is silently ignored whenever `--force` is present. That is exactly the shape a user reaches for when they want to check a forced delete before committing to it. The PR's `if dryRun || !force` gate closes it. Worth saying out loud in the title/description, since "preserve forced delete dry-runs" reads as cosmetic.

## What I verified

- **`--force` reaches single and batch dry-run previews without mutating.** Single forwards `force` with `dryRun=true`; batch now passes `force` through to `DeleteIssues` instead of the hardcoded `false`. In `issueops/delete.go`, `force` only widens dependent discovery, and the `dryRun` return happens before any delete, dependency removal, reference rewrite, or blocked-state recomputation. Embedded Dolt opens the dry-run transaction non-committing; server Dolt suppresses wisp deletion and returns before staging/commit; the proxied domain path returns before all writes and closes through rollback.
- **Quiet previews are empty, JSON previews are structural-only.** The JSON preview carries IDs, mode flags, counts and orphan count — no titles or descriptions — and JSON correctly takes precedence over quiet. I checked the error path specifically: `outputDeletionPreview` writes the preview to stdout and `outputJSONError` writes to stderr, so the two calls in sequence do not produce two JSON documents on one stream.
- **No change to the destructive path.** The only incidental behavioural change is that proxied JSON encoding failures are now returned rather than discarded, which is an improvement.

## Two low-severity notes (non-blocking)

1. `TestOutputDeleteProxiedPreviewIsPayloadBlind` constructs a `deleteInput` and a synthetic result and calls only the output helper, so it bypasses flag gathering, `runDeleteProxiedPreviewTx`, and UOW rollback. Its name also falls outside the `TestProxiedServer*` selector in `.github/scripts/proxied-test-shard.sh`, so it will not run in the PR-risk proxied lane. The proxied quiet/privacy boundary is correct by inspection but is not regression-gated on PRs. Renaming it into the `TestProxiedServer*` namespace would be a cheap improvement — happy to do that maintainer-side rather than asking you for another round trip.
2. The proxied-server preview passes `Cascade:true, DryRun:true` but no `Force`. That is consistent with that backend's always-cascade policy, where `--force` is a confirmation gate rather than orphaning semantics — but it does mean there is no force-parity across backends, so the description shouldn't be read as claiming one.

Also noted (pre-existing, not yours to fix here): the embedded and proxied JSON preview schemas diverge — `issue_ids`/`would_remove_*`/`would_orphan` versus `ids`/`*_removed`/`connected`.

On the release request in the description: release cutting is a separate maintainer decision tracked on its own issues, so it will not be handled through this PR — but the fix itself is exactly the kind of thing that motivates a cut. Thanks for a tightly scoped patch with real regression coverage.

PR #5073, opened 2026-07-26
ci: give conformance explicit timeout budgets

ecuthiell:
# ci: give conformance explicit timeout budgets

Fixes #5070

## What

- Give the embedded-Dolt reference oracle an explicit 30-minute Go test
  timeout.
- Preserve the end-to-end CLI tier's existing ten-minute behavior explicitly.
- Add a 45-minute outer timeout to the standalone Conformance job.
- Add black-box tests for both exact Go argument vectors and for nonzero exit
  propagation from either tier.

## Why

The standalone entrypoint inherited Go's ten-minute default even though the
embedded oracle regularly needs longer and the dedicated conformance jobs
already budget 30 minutes. PR #4979 run `29973997194` timed out at exactly ten
minutes while the same merge commit passed in the explicitly budgeted
conformance lane.

The three deadlines are intentionally nested:

- Tier 1: 30 minutes for the long-running embedded oracle.
- Tier 2: 10 minutes, preserving its prior implicit behavior.
- Workflow: 45 minutes for both tiers plus setup and teardown.

The workflow placement is not a platform claim: these relative timeout and
failure-propagation semantics are host-independent. The exact-head standalone
Conformance result remains the required publication-time execution proof.

## Validation

Exact candidate `131c17fdfcc5f814b198292421a36917e2e811c9` passed:

- black-box, NUL-delimited argv assertions for the exact 30m and 10m Go calls;
- failure-propagation falsifiers for Tier 1 and Tier 2;
- focused, default, and integration Go tests;
- real Go command grammar checks;
- Bash syntax, ShellCheck, actionlint, golangci-lint, `gofmt`, and
  `git diff --check`;
- independent contract-centred review of the exact candidate.

The fresh exact-head `Conformance / Storage backend conformance (embedded Dolt
oracle)` pull-request job remains mandatory after publication.

---

_codex-tui-gpt-5.6-sol-ultra on behalf of Ewen Cuthiell_

PR #5115, opened 2026-07-28
fix(preflight): classify status checks latest-per-name, not raw rollup entries

maphew:
## Why

The pr-babysit merge patrol was permanently unable to merge three green, approved PRs (#5073/#5074/#5076): preflight reported "21 status check(s) failed or require action" while GitHub showed `mergeStateStatus: CLEAN` and `gh pr checks` showed 80/80 pass. Each needed a manual in-session merge.

Root cause: `statusCheckRollup` keeps every check run on the head SHA, including runs from cancelled or superseded check suites. Those heads carried 20 stale CANCELLED proxied-test runs plus one stale `CI Gate / Required` FAILURE from an older suite, next to the fresh green re-runs of the same checks. Preflight counted raw rollup entries, so the stale ones blocked forever.

## What

Collapse the rollup to the most recent entry per check name/context (`completedAt`, falling back to `startedAt`/`createdAt`) before classifying failed/pending, mirroring GitHub's own latest-per-name mergeability semantics.

- A genuinely failing or pending *latest* run still blocks — regression-tested both directions.
- The fake-gh test harness now accepts an injectable rollup via `PR_ROLLUP` (defaults to `[]`, existing fixtures unchanged).

## Validation

- `go test -tags integration -run TestPRPreflight ./scripts/` — all pass, including the two new subtests.
- Live run against an open PR with re-run suites reports checks correctly.

### DIFF
--- a/scripts/pr-preflight.sh
+++ b/scripts/pr-preflight.sh
@@ -300,9 +300,19 @@ fi
 # GitHub reports CLEAN (observed on gastownhall/beads#5073–5076: 20 stale
 # CANCELLED runs + 1 stale gate FAILURE alongside 80 green checks). Collapse
 # to the most recent entry per check name/context before classifying.
-rollup_latest=$(jq '[.statusCheckRollup[]?]
-  | group_by(.name // .context // "")
-  | map(max_by(.completedAt // .startedAt // .createdAt // ""))' <<<"$json")
+# Group key includes __typename and workflowName: a commit status and a
+# check run sharing a name are independent gates (GitHub requires both),
+# and identically named jobs in different workflows must not collapse into
+# one entry where a green run could mask the other workflow's red one.
+# gh exports absent CheckRun times as the zero time (0001-01-01...), so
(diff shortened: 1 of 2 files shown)


maphew:
Pushed a second commit addressing cross-vendor (Codex gpt-5.6-sol) review findings on the dedupe:

1. **Independent checks no longer merge.** The group key is now `(__typename, workflowName, name/context)` — a commit status and check run sharing a name are independent required gates, and identically named jobs in different workflows can no longer collapse into one entry where a green run masks the other workflow's red.
2. **gh zero-time exports handled.** `gh` exports absent CheckRun times as `0001-01-01T00:00:00Z` (non-pointer time.Time), which defeated the `//` fallback. Times are normalized and "latest" is the max of completed/started/created, so a fresh in-progress rerun supersedes an old completed failure and classifies as pending rather than stale-failed.

Both cases plus the status-vs-check-run case are regression-tested (five subtests total now).

PR #5120, opened 2026-07-28
fix(bootstrap): prefer existing DB over re-clone when sync.remote is set (GH#5037)

### LINKED ISSUE #5037: bd bootstrap with sync.remote set fails on an existing database (Error 1007) instead of validating or replacing it
## Summary

With `sync.remote` configured, `bd bootstrap` on a workspace that already has a local database always plans "clone from remote" and then fails when the clone hits the existing database:

```
Bootstrap plan: clone from remote
  Remote: http://<host>:7007/<db>
  Database: <db>
Error: Bootstrap failed: clone from remote: dolt clone http://<host>:7007/<db>: Error 1007: can't create database <db>; database exists
```

This contradicts both documented behaviors:

- `bd bootstrap --help`: "If database already exists: validates and reports status"
- Upgrade guide (`website/docs/getting-started/upgrading.md`, "Multiple clones sharing one remote", step 3): every *other* clone should run `bd bootstrap` to adopt the migrated database, and "`bd bootstrap` replaces the local database"

Practical impact: the documented v1.0 → v1.1 upgrade path for non-migrator clones fails on exactly the clones it is written for — ones that already have the database. That's how I hit it: the designated clone migrated and pushed, a second machine upgraded bd to 1.1.0 and ran `bd bootstrap` per the guide, and got Error 1007.

## Repro

bd 1.1.0 (built from the `v1.1.0` tag), embedded mode, Linux x86_64. Workspace git repo has `.beads/config.yaml` with `sync.remote: "http://<host>:7007/<db>"` (Dolt remotesapi).

```
git clone <workspace-repo> ws && cd ws
bd bootstrap --yes   # fresh clone from remote — works
bd bootstrap --yes   # Error 1007: can't create database <db>; database exists
```

## Cause

In `cmd/bd/bootstrap.go` at `v1.1.0`:

- `detectBootstrapAction` returns the `sync` plan unconditionally whenever `sync.remote` is set (~line 298), so the existing-database check (`existingBootstrapDBPlan`, line 365) is never reached and the help text's "validates and reports status" path cannot trigger.
- `executeSyncAction` (line 630) calls `cloneFromRemote` → `cloneViaEmbedded` (line 838), which runs `DOLT_CLONE` into the existing data dir without moving the current database aside — nothing implements the "replaces the local database" behavior the upgrade guide promises.

## Workaround

```
mv .beads/embeddeddolt .beads/embeddeddolt.bak
bd bootstrap
```

## Expected

Either behavior would work, as long as code and docs agree:

- existing database + `sync.remote` → validate and report status (help-text behavior), with an explicit flag to force a re-clone for the adopt-migrated-remote upgrade case; or
- the sync action moves the existing database aside (or drops it only after a successful clone) so `bd bootstrap` actually replaces it as the upgrade guide states.

vishnujayvel:
## Summary

Fixes `bd bootstrap` with `sync.remote` set on a workspace that **already has** a local database.

Previously `detectBootstrapAction` returned `Action=sync` whenever `sync.remote` was configured, skipping the existing-DB check. A second bootstrap then ran `DOLT_CLONE` into the existing dir and failed with `Error 1007: can't create database …; database exists` — contradicting help text ("If database already exists: validates and reports status") and the multi-clone upgrade guide.

Existing local DB detection now runs **before** the `sync.remote` early-return. Fresh clones (no local DB) still plan `sync` as before.

Closes #5037

## Test plan

- [x] `go test ./cmd/bd/ -run 'TestDetectBootstrapAction_ExistingEmbeddedDBWithSyncRemoteIsNoOp|TestDetectBootstrapAction_ExplicitSyncRemote' -count=1` (ICU cgo) — pass
- [x] Explicit sync.remote without local DB still plans `sync`

### DIFF
--- a/cmd/bd/bootstrap.go
+++ b/cmd/bd/bootstrap.go
@@ -309,6 +309,21 @@ func detectBootstrapAction(beadsDir string, cfg *configfile.Config) BootstrapPla
 	isSharedServer := bootstrapSharedServerMode(beadsDir)
 	isServer := cfg.IsDoltServerMode() || isSharedServer
 
+	// Prefer an existing local database over re-clone (GH#5037). Previously
+	// sync.remote returned Action=sync unconditionally, so a second
+	// `bd bootstrap` on an already-cloned workspace ran DOLT_CLONE into the
+	// existing dir and failed with Error 1007 (database exists) — contradicting
+	// help text ("If database already exists: validates and reports status")
+	// and the multi-clone upgrade guide. If the local beadsDir does not exist
+	// yet, still prefer sync recovery first for Action=="none" so a default
+	// shared-server "beads" DB from another project cannot mask a real clone.
+	if dbAction, ok := existingBootstrapDBPlan(beadsDir, cfg, isServer, isSharedServer); ok {
+		if beadsDirExists || dbAction.Action != "none" {
+			return dbAction
+		}
+		plan = dbAction
+	}
+
 	// Check sync.remote (primary) or sync.git-remote (deprecated fallback)
 	syncRemote := resolveSyncRemote()
 	if syncRemote != "" {
@@ -333,20 +348,6 @@ func detectBootstrapAction(beadsDir string, cfg *configfile.Config) BootstrapPla
 				return plan
 			}
 		}
-	}
-
-	if dbAction, ok := existingBootstrapDBPlan(beadsDir, cfg, isServer, isSharedServer); ok {
-		// If the local beadsDir does not exist yet, prefer recovering via sync
-		// first. This avoids false "nothing to do" results when the default
-		// shared-server database name happens to exist for another project.
-		if beadsDirExists || dbAction.Action != "none" {
-			return dbAction
-		}
-		// For synthesized paths with no local workspace directory yet, defer the
-		// existing-db no-op until we've ruled out all other recovery paths.
-		// This preserves the sync-precedence fix without downgrading the
-		// legitimate "database already exists" case into a fresh init.
-		plan = dbAction
 	}
 
 	// Check for backup JSONL files (must be non-empty to be useful)
--- a/cmd/bd/bootstrap_test.go
+++ b/cmd/bd/bootstrap_test.go
@@ -367,6 +367,62 @@ func TestDetectBootstrapAction_ExplicitSyncRemotePreservesRemotesAPIURL(t *testi
 	}
 }
 
+// TestDetectBootstrapAction_ExistingEmbeddedDBWithSyncRemoteIsNoOp is a
+// regression for GH#5037: when sync.remote is set AND embeddeddolt already
+// exists, plan must be Action=none (validate/report), not Action=sync which
+// would DOLT_CLONE into the existing database and fail with Error 1007.
+func TestDetectBootstrapAction_ExistingEmbeddedDBWithSyncRemoteIsNoOp(t *testing.T) {
+	restore := snapshotBootstrapEnv(t)
+	defer restore()
+	config.ResetForTesting()
+	defer config.ResetForTesting()
+
+	tmpDir := t.TempDir()
+	beadsDir := filepath.Join(tmpDir, ".beads")
+	if err := os.MkdirAll(beadsDir, 0o750); err != nil {
+		t.Fatal(err)
+	}
+	// Non-empty embedded database footprint (what a prior successful bootstrap left).
+	embedded := filepath.Join(beadsDir, "embeddeddolt")
+	if err := os.MkdirAll(filepath.Join(embedded, "mydb"), 0o750); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(embedded, "mydb", ".keep"), []byte("x"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	const syncRemote = "http://myserver:7007/mydb"
+	if err := os.WriteFile(filepath.Join(beadsDir, "config.yaml"), []byte("sync.remote: "+syncRemote+"\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	t.Setenv("BEADS_DIR", beadsDir)
+	t.Setenv("BEADS_TEST_IGNORE_REPO_CONFIG", "1")
+	if err := config.Initialize(); err != nil {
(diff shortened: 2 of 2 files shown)


maphew (commented):
The diagnosis in this PR and in #5037 is exactly right, and I verified each step of it against `upstream/main`: `detectBootstrapAction` does return `Action=sync` unconditionally once `sync.remote` is set, so `existingBootstrapDBPlan` is unreachable on that path, and `cloneViaEmbedded` (cmd/bd/bootstrap.go:857) really does run `DOLT_CLONE` into the existing data dir with nothing moving the current database aside. The regression test is also the real thing rather than decoration — I reverted `cmd/bd/bootstrap.go` to its pre-PR state and `TestDetectBootstrapAction_ExistingEmbeddedDBWithSyncRemoteIsNoOp` fails with `action = "sync"`, exactly as advertised. The branch merges cleanly with current main and `go test ./cmd/bd -run 'Bootstrap|bootstrap'` is green after the merge. Thanks for filing this with a reproduction and then following it with a patch; that is the useful shape of a bug report.

Two things need to change before this can land, both about the *breadth* of the fix rather than its direction.

**1. The hoist also short-circuits the server-probe-failure path.**

`cmd/bd/bootstrap.go:320-325` returns on any `ok == true` from `existingBootstrapDBPlan`, but that includes the failure branch at `cmd/bd/bootstrap.go:414-417`, which returns `Action="none"`, `HasExisting=false`, `Reason="Could not verify existing server database ..."` when the probe errors. I reproduced the consequence with an injected `checkBootstrapServerDB` returning `{Reachable:false, Err:"connection refused"}`, server mode, `.beads/` present, a non-empty data dir, and `sync.remote` configured:

- before this PR: `action="sync" reason="sync.remote configured — will clone from http://myserver:7007/mydb"`
- after this PR: `action="none" reason="Could not verify existing server database project_missing: connection refused"`

Because `bootstrapCmd` returns nil when `plan.Action == "none"`, `bd bootstrap` now exits 0 having done nothing when the Dolt server is unreachable — precisely the recovery scenario bootstrap exists for. Gating the early return on a database we positively found fixes it without giving up the fix:

```go
if dbAction, ok := existingBootstrapDBPlan(beadsDir, cfg, isServer, isSharedServer); ok && dbAction.HasExisting {
    if beadsDirExists {
        return dbAction
    }
}
```

and leaving the original block where it was, so the probe-error and synthesized-worktree cases keep their existing ordering (`e585a6187`, the shared-server-worktree fix, is what that later block is protecting).

**2. The documented upgrade path becomes a silent no-op.**

`docs/getting-started/upgrading.md:227-233` tells every non-migrator clone to run `bd bootstrap` to adopt the migrated remote, and states that "`bd bootstrap` replaces the local database". Those clones all have an existing local DB, so with this change that step prints "Database already exists at ..." and exits 0 while leaving the un-migrated database in place. Error 1007 was the wrong outcome, but it was at least a stop signal; a false success on a schema-migration path is worse. #5037 anticipated this — it asked for "validate and report status ... **with an explicit flag to force a re-clone** for the adopt-migrated-remote upgrade case". The PR ships the first half without the second, and without touching the doc.

Either half of that closes the gap: implement the force/adopt path (move `embeddeddolt` aside, clone, drop the backup only on success), or keep this PR's scope and update `docs/getting-started/upgrading.md:227-233` to the manual `mv .beads/embeddeddolt .beads/embeddeddolt.bak && bd bootstrap` procedure plus an actionable hint printed on the no-op.

Smaller points, none blocking on their own:

- `cmd/bd/bootstrap.go:105-110` — the `--help` bullet list is precedence-ordered, and this PR inverts the precedence for both the `sync.remote` and the `refs/dolt/data` bullets without updating it. The moved comment at `cmd/bd/bootstrap.go:316` cites that same help text as justification, so it should not be left describing the old order.
- `cmd/bd/bootstrap.go:321` — `dbAction.Action != "none"` is dead code: every `ok == true` return from `existingBootstrapDBPlan` sets `Action = "none"` (lines 415, 420, 428). Pre-existing, but the `HasExisting` gate above would replace it cleanly.
- `cmd/bd/bootstrap.go:324` — `plan = dbAction` now runs before the sync block, so `bd bootstrap --json` can emit `action:"sync"` together with `has_existing:true` for a shared-server synthesized dir. That combination was unreachable before. Nothing in-tree reads `HasExisting`, so it is cosmetic, but it is an unintended change to the JSON output.
- `cmd/bd/bootstrap.go:392-401` — the embedded existence test is just "`.beads/embeddeddolt` exists and is non-empty"; it never checks the configured database directory. As a last-resort fallback that was fine; as the gate in front of sync recovery it means a partial `embeddeddolt` left by an interrupted clone blocks the retry that would repair it. Narrowing to `filepath.Join(dbPath, cfg.GetDoltDatabase())` would make the new precedence considerably safer.
- Test coverage follows the same shape: the one new case is the embedded happy path. The two newly-preempted paths (server probe error + `sync.remote`; existing local DB + git origin carrying `refs/dolt/data`) are the ones that would catch a future re-break.

One coordination note: #4449 edits the same `sync.remote` block and the same server branch of `existingBootstrapDBPlan`. It predates this PR, so it has precedence on any textual conflict, but the two are conceptually independent and both can land — whichever lands second just needs a rebase.

Happy to push the `HasExisting` gate and the help-text ordering as commits on this branch if you would rather not respin it, or leave them to you — your call, just say which.

vishnujayvel:
Thanks for verifying the regression by reverting the file and for catching the probe-error short-circuit — that's the one that matters: our patch would have silently no-op'd `bd bootstrap` on an unreachable server, exactly the recovery case bootstrap exists for. No false claims in our body or test plan to correct.

We agree with both required changes. For the upgrade-path gap: rather than build the force/adopt-remote flag #5037 asked for, we'd rather keep this PR's scope narrow and update `docs/getting-started/upgrading.md:227-233` with the manual `mv .beads/embeddeddolt .beads/embeddeddolt.bak && bd bootstrap` procedure plus a hint on the no-op — the flag is a bigger feature that deserves its own PR/issue.

Please take the `HasExisting` gate (item 1), the docs update (item 2), and the smaller cleanup — help-text ordering, the dead `Action != "none"` check, and narrowing the embedded-existence test to the configured database dir — as maintainer commits. We'll leave the #4449 rebase to whoever lands second, as you suggested.

Appreciate the thoroughness here.

PR #5121, opened 2026-07-29
fix(export): stage basename under cmd.Dir to avoid double-root git add (GH#4351)

### LINKED ISSUE #4351: Auto-export 'git add' always fails: cmd.Dir + path are doubly-rooted
# Auto-export `git add` always fails with "exit status 128" because `cmd.Dir` + path are doubly-rooted

## Symptom

Every `bd` operation that triggers auto-export emits:

```
Warning: auto-export: git add failed: exit status 128
```

After the operation, `.beads/issues.jsonl` is left in a dirty git state (`M` in `git status`), polluting the working tree until manually committed or restored.

Reproduced on **bd 1.0.3 (Homebrew, macOS)** in any git repo with default config (`export.path = "issues.jsonl"`, `export.auto = true`, `export.git-add = true`). Also present in `main` as of bd v1.0.4.

## Root cause

In `cmd/bd/export_auto.go`, the `gitAddFile` function receives the full relative path from the repo root (e.g., `.beads/issues.jsonl`) but then sets `cmd.Dir` to the _directory_ portion of that path (`.beads/`), while passing the full path as the `git add` argument:

```go
// In maybeAutoExport (lines ~95–145):
exportPath := config.GetString("export.path")        // "issues.jsonl"
fullPath := filepath.Join(beadsDir, exportPath)       // ".beads/issues.jsonl"
// ...
gitAddFile(fullPath)                                  // called with ".beads/issues.jsonl"

// In gitAddFile (lines ~628–655):
cmd := exec.CommandContext(ctx, "git", "add", path)   // path = ".beads/issues.jsonl"
cmd.Dir = filepath.Dir(path)                          // ".beads"
```

The effect is: `cd .beads && git add .beads/issues.jsonl`. From inside `.beads`, the argument resolves to `.beads/.beads/issues.jsonl`, which doesn't exist. Hence `fatal: pathspec '.beads/issues.jsonl' did not match any files` → exit 128.

Verified manually outside `bd`:

```
$ cd /path/to/repo
$ (cd .beads && git add .beads/issues.jsonl)
fatal: pathspec '.beads/issues.jsonl' did not match any files
$ echo $?
128

$ git add .beads/issues.jsonl
$ echo $?
0
```

## Proposed fix

The simplest change is one line in `gitAddFile`:

```diff
-cmd := exec.CommandContext(ctx, "git", "add", path)
-cmd.Dir = filepath.Dir(path)
+cmd := exec.CommandContext(ctx, "git", "add", filepath.Base(path))
+cmd.Dir = filepath.Dir(path)
```

Pass `filepath.Base(path)` instead of `path` when `Dir` is set — `git add issues.jsonl` from inside `.beads` resolves correctly.

Alternatively, leave the argument as `path` but **don't set `cmd.Dir`** at all (let it inherit the caller's CWD, which is the repo root):

```diff
 cmd := exec.CommandContext(ctx, "git", "add", path)
-cmd.Dir = filepath.Dir(path)
```

Either fix should be a single-line PR. The first is more surgical and preserves the existing `cmd.Dir` for any locking/env behavior the caller depends on.

## Why it matters

The warning isn't fatal (the in-memory bd state is correct), but the resulting dirty `.beads/issues.jsonl` creates real friction:

- Repos using bd alongside git workflows get a permanently-dirty file that has to be repeatedly `git checkout`-ed or committed.
- Pre-commit hooks that scan `git status` see a phantom modification on every run.
- Downstream tooling that depends on a clean worktree (e.g. CI gates, release scripts) fails until the file is manually reset.

For multi-agent / pre-commit-gated workflows, every `bd update`, `bd create`, and `bd close` produces this dirt — it accumulates fast.

## Suggested regression test

```go
// Test that auto-export's git add resolves correctly when run from a directory
// other than the .beads parent.
func TestGitAddFileWithSubdirDir(t *testing.T) {
  // Setup: repo with .beads/issues.jsonl
  repo := setupTempRepo(t)
  exportPath := filepath.Join(repo, ".beads", "issues.jsonl")
  if err := os.WriteFile(exportPath, []byte("{}\n"), 0644); err != nil {
    t.Fatal(err)
  }

  // Act
  if err := gitAddFile(exportPath); err != nil {
    t.Fatalf("gitAddFile returned error: %v", err)
  }

  // Assert: file is now staged
  staged := mustRun(t, repo, "git", "diff", "--cached", "--name-only")
  if !strings.Contains(staged, ".beads/issues.jsonl") {
    t.Errorf("expected .beads/issues.jsonl staged, got: %q", staged)
  }
}
```

## Filing

Resolved upstream repo: `gastownhall/beads`

Exact ready-to-run filing command (requires `gh` auth to `gastownhall/beads`):

```sh
gh issue create --repo gastownhall/beads \
  --title "Auto-export 'git add' always fails: cmd.Dir + path are doubly-rooted" \
  --body-file docs/plans/lvl-upstream-issue.md
```

Alternatively, file manually at https://github.com/gastownhall/beads/issues/new using this document as the issue body.

vishnujayvel:
## Summary

Fixes an auto-export `git add` double-rooting path that would fail with exit 128 when the export path was relative (e.g. `.beads/issues.jsonl`).

`gitAddFile` set `cmd.Dir` to the parent of the path but passed the full path as the `git add` argument, so git effectively ran:

```
cd .beads && git add .beads/issues.jsonl
```

which looks for a non-existent nested path. Pass `filepath.Base(path)` (with `--`) so the pathspec is relative to `cmd.Dir`. Preserves `cmd.Dir = parent` so the GH#3311 hook worktree staging path still lands under `.beads/`, not bare `issues.jsonl` at repo root.

Refs #4351

(Corrected 2026-07-29: this was originally `Closes #4351`. It should not close the issue — as @maphew established in review, the regression fixed here is real but is not the reporter's exit-128 path, which looks more likely to be an `isGitRepo()` cwd-vs-`cmd.Dir` mismatch. The original claim was ours and it was wrong.)

## Test plan

- [x] `go test ./cmd/bd/ -run 'TestGitAddFile_' -count=1` (with ICU cgo flags) — all gitAddFile tests pass, including new `TestGitAddFile_RelativePathDoesNotDoubleRoot`
- [x] Existing GH#3311 worktree-hook regression still green

### DIFF
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -600,9 +600,12 @@ func gitAddFile(path string) error {
 
 	ctx, cancel := context.WithTimeout(context.Background(), gitAddTimeout)
 	defer cancel()
-	// Pass the basename only: cmd.Dir is the parent of path, so a full path
-	// argument would double-root (cd .beads && git add .beads/issues.jsonl →
-	// pathspec looks under .beads/.beads/). See GH#4351.
+	// Pass the basename only, defensively: cmd.Dir is the parent of path, so
+	// a full path argument would double-root (cd .beads && git add
+	// .beads/issues.jsonl → pathspec looks under .beads/.beads/) if a caller
+	// ever passed a relative path here. Both current callers pass absolute
+	// paths, so this guards against a regression rather than fixing a live
+	// failure. See GH#4351.
 	// Keep cmd.Dir = parent so GH#3311 hook worktree staging still resolves
 	// the index path under the repo root (not bare "issues.jsonl" at root).
 	cmd := exec.CommandContext(ctx, "git", "add", "--", filepath.Base(path))
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -600,7 +600,12 @@ func gitAddFile(path string) error {
 
 	ctx, cancel := context.WithTimeout(context.Background(), gitAddTimeout)
 	defer cancel()
-	cmd := exec.CommandContext(ctx, "git", "add", path)
+	// Pass the basename only: cmd.Dir is the parent of path, so a full path
+	// argument would double-root (cd .beads && git add .beads/issues.jsonl →
+	// pathspec looks under .beads/.beads/). See GH#4351.
+	// Keep cmd.Dir = parent so GH#3311 hook worktree staging still resolves
+	// the index path under the repo root (not bare "issues.jsonl" at root).
+	cmd := exec.CommandContext(ctx, "git", "add", "--", filepath.Base(path))
 	cmd.Dir = filepath.Dir(path)
 	cmd.Env = env
 	// Capture combined output so the caller's warning surfaces git's stderr
--- a/cmd/bd/export_auto_test.go
+++ b/cmd/bd/export_auto_test.go
@@ -584,6 +584,77 @@ func TestGitAddFile_NonHookContext_GuardDoesNotFire(t *testing.T) {
 	}
 }
 
+// TestGitAddFile_RelativePathDoesNotDoubleRoot is a regression test for
+// GH#4351: gitAddFile sets cmd.Dir to filepath.Dir(path). When path is
+// relative (e.g. ".beads/issues.jsonl"), passing that full path as the
+// git-add argument becomes `cd .beads && git add .beads/issues.jsonl`,
+// which looks for a non-existent nested path and exits 128. The fix is
+// to pass filepath.Base(path) so the pathspec is relative to cmd.Dir.
+func TestGitAddFile_RelativePathDoesNotDoubleRoot(t *testing.T) {
+	if _, err := exec.LookPath("git"); err != nil {
+		t.Skip("git not available")
+	}
+
+	tmpDir, err := os.MkdirTemp("", "bd-gh4351-*")
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { _ = os.RemoveAll(tmpDir) })
+	tmpDir, err = filepath.EvalSymlinks(tmpDir)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	repo := filepath.Join(tmpDir, "repo")
+	if err := os.MkdirAll(repo, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	runGit := func(args ...string) {
+		t.Helper()
+		c := exec.Command("git", args...)
+		c.Dir = repo
+		if out, err := c.CombinedOutput(); err != nil {
+			t.Fatalf("git %v: %v\n%s", args, err, out)
+		}
+	}
+	runGit("init", "-q")
+	runGit("config", "user.email", "t@t")
+	runGit("config", "user.name", "t")
+	if err := os.WriteFile(filepath.Join(repo, "README.md"), []byte("x\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	runGit("add", "README.md")
+	runGit("commit", "-qm", "init")
+
+	if err := os.MkdirAll(filepath.Join(repo, ".beads"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(repo, ".beads", "issues.jsonl"), []byte(`{"id":"x"}`+"\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := os.Unsetenv("GIT_DIR"); err != nil {
+		t.Fatal(err)
+	}
+	// Repo-relative path — the double-root trigger. Must stage successfully.
+	t.Chdir(repo)
+	relPath := filepath.Join(".beads", "issues.jsonl")
+	if err := gitAddFile(relPath); err != nil {
+		t.Fatalf("gitAddFile(%q): %v", relPath, err)
+	}
+
+	c := exec.Command("git", "diff", "--cached", "--name-only")
+	c.Dir = repo
+	data, err := c.CombinedOutput()
+	if err != nil {
+		t.Fatalf("git diff --cached: %v\n%s", err, data)
+	}
+	staged := strings.TrimSpace(string(data))
+	if !strings.Contains(staged, ".beads/issues.jsonl") && !strings.Contains(staged, filepath.ToSlash(relPath)) {
+		t.Errorf("expected .beads/issues.jsonl staged, got: %q", staged)
+	}
+}
+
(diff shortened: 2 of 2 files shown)


maphew (commented):
The regression test here is the good part: `TestGitAddFile_RelativePathDoesNotDoubleRoot` in `cmd/bd/export_auto_test.go` is a real test, not a decorative one. I checked out the branch, reverted only `cmd/bd/export_auto.go` to the parent commit, and it fails exactly as you describe:

```
--- FAIL: TestGitAddFile_RelativePathDoesNotDoubleRoot
    gitAddFile(".beads/issues.jsonl"): exit status 128: warning: could not open
    directory '.beads/.beads/': No such file or directory
    fatal: pathspec '.beads/issues.jsonl' did not match any files
```

It also follows the house conventions already in that file (`t.Chdir`, the `os.Unsetenv("GIT_DIR")` pattern from lines 568 and 700, `EvalSymlinks` on the temp dir). The `--` separator you added to the `git add` argv is a genuine bonus hardening against an export path beginning with `-`.

The code change itself is safe. For every path the callers actually produce it is behavior-identical, and for a relative path it is strictly better. I have no correctness objection to merging it.

**But it does not fix #4351, and the `Closes #4351` line needs to go.**

`gitAddFile` has exactly two production callers:

- `cmd/bd/export_auto.go:151` — `fullPath` built at `cmd/bd/export_auto.go:88` as `filepath.Join(beadsDir, exportPath)`
- `cmd/bd/hooks.go:1437` — `fullPath` built the same way at `cmd/bd/hooks.go:1388`

In both, `beadsDir` comes from `beads.FindBeadsDir()` (`internal/beads/beads.go:748`). Every return path of that function goes through `utils.CanonicalizePath`, which is `filepath.Abs` followed by `filepath.EvalSymlinks`. The result is always absolute, so `fullPath` is always absolute and `cmd.Dir = filepath.Dir(path)` is always an absolute directory.

Absolute pathspecs work fine from a subdirectory:

```
$ (cd repo/.beads && git add /abs/repo/.beads/issues.jsonl); echo $?
0
$ git -C repo diff --cached --name-only
.beads/issues.jsonl
```

Only the relative form reproduces the exit 128 — and no caller produces it. So the double-rooting is a latent bug in `gitAddFile`'s contract, worth fixing, but it is not the mechanism behind the exit 128 the reporter saw. Merging with `Closes #4351` would auto-close a live bug while the reporter's symptom persists. Please change it to `Refs #4351`.

**A more likely cause for #4351, if you want to chase it:** `isGitRepo()` (`cmd/bd/sync_git.go:19`) runs `git rev-parse --git-dir` in the **process cwd**, but `gitAddFile` runs `git add` with `cmd.Dir` set to the **beads directory**. When the beads dir lives outside the cwd's repository — `BEADS_DIR` pointing elsewhere, global mode, or a `.beads/redirect` — the gate at `cmd/bd/export_auto.go:150` passes and the `git add` then fails with `fatal: not a git repository`, which is exit 128. That would match the report exactly. The reporter was on 1.0.3, which predates #3505's stderr capture, so they would have seen only the bare `exit status 128` with no git message. Asking them to re-run on 1.1.x and paste the now-visible git stderr would settle which of these it is in one round trip.

Two small things on the patch itself:

- `cmd/bd/export_auto.go:602-605` — the comment states the double-root as live behavior. Given it is unreachable from the current callers, please phrase it defensively instead, e.g. "cmd.Dir is the parent of path, so the pathspec must be relative to it; a full path argument would double-root if path were ever relative." As written it will convince the next reader that the bug was live.
- The `--` separator is a behavior change independent of the basename fix and deserves a line in the summary.

The branch is 116 commits behind main but still merges clean with all checks green; the rebase is maintainer work, not something to ask you for.

Net: this is a real (if latent) hardening with a real test, and I would like to take it. Update the issue reference from `Closes` to `Refs` and adjust that comment, and it can go in — or say the word and I will make both edits on merge. Either way, thank you for the careful test; the diff was easy to verify precisely because you wrote one that actually fails without the fix.

vishnujayvel:
Thanks — the trace showing `gitAddFile`'s two callers always produce absolute paths, so the double-rooting can't actually fire today, is the important finding. Good catch, and better than our own read of #4351.

Correction: our body says `Closes #4351`; it should be `Refs #4351`. The regression fixed here is real and worth keeping, but it isn't the reporter's exit-128 path — agreed.

We'll take your suggested next step rather than guess further: ask the #4351 reporter to re-run on 1.1.x and paste the now-visible git stderr, which will confirm or rule out the `isGitRepo()` cwd-vs-`cmd.Dir` mismatch you flagged as the likelier cause.

Please push both small edits — the `Closes`→`Refs` change and rewording the `export_auto.go:602-605` comment to state the double-root defensively rather than as live behavior — as maintainer commits on the branch. Nothing here needs to stay with us; the test is the part we're glad to have gotten right, and you verified it properly.

maphew:
Pushed the comment reword as `e8f3cbf78`. It now says the basename-only call guards against a caller ever passing a relative path, notes that both current callers pass absolute paths, and states plainly that this is a regression guard rather than a fix for a live failure.

The `Closes` change turned out to need nothing from me — you had already corrected the description to `Refs #4351`, and the commit message never carried a `Closes` trailer. So that one is done.

Agreed on the next step for #4351 itself: the `isGitRepo()` cwd-versus-`cmd.Dir` mismatch is the likelier cause, and the reporter re-running on 1.1.x with the now-visible git stderr is what will settle it. This PR stands on its own as the guard plus its test either way.

vishnujayvel:
Thanks — `e8f3cbf78` reads exactly right. Framing it as a regression guard rather than a fix for a live failure is the honest version, and it's the framing I should have used in the first place: I asserted the double-rooting was the mechanism behind the reported exit 128 without tracing the callers, and your trace through `FindBeadsDir` → `CanonicalizePath` → `filepath.Abs`/`EvalSymlinks` shows every production call site passes an absolute path. That was my error, not a difference of judgement.

Agreed on the `isGitRepo()` cwd-versus-`cmd.Dir` mismatch as the likelier cause of #4351 — the gate at `cmd/bd/export_auto.go:150` checking the process cwd while `git add` runs with `cmd.Dir` set to the beads dir is a real gap whenever `BEADS_DIR`, global mode, or a `.beads/redirect` puts the beads dir outside the cwd's repo, and `fatal: not a git repository` is exit 128. The 1.0.3 detail fits too: pre-#3505 there was no stderr capture, so the reporter would have seen the bare status and nothing else.

Asking them to re-run on 1.1.x and paste the git stderr is the right next round trip. Happy to open that as a separate issue-thread comment or a follow-up PR once we know which it is — say which you'd prefer and I'll take it.

Nothing outstanding from me on this one.

Issue #4750, opened 2026-07-13
bd create --parent recycles child suffixes of ARCHIVED children: child_counters is not a true high-water mark


steveyegge:
## Summary

`bd create --parent <P>` can mint a child ID that **collides with an already-existing (archived) child** of the same parent. The `child_counters` high-water mark is supposed to prevent this, but it is not actually a high-water mark: it is only advanced by two of the several code paths that materialize a child ID, it is not exported to JSONL, and it is not consulted against issues that compaction/archive has swept out of the `issues` table.

Result: after a repo has (a) swept closed children out of `issues` into an archive JSONL, and/or (b) rebuilt its DB from `issues.jsonl`, `bd create --parent` restarts the suffix sequence at `.1` and silently re-issues IDs that already refer to different, historical issues. Grep of the archive, of commit messages, or of code comments citing the old IDs then resolves to two different issues.

## Observed

In a production repo (prefix `wy`, Dolt server backend, bd 1.1.0 / `86ccd7ce7`):

- Epic `wy-y5uv` had children `.1`–`.7`, all closed and swept to `.beads/archive/2026-07-02-closed-sweep.jsonl` (so no longer rows in `issues`).
- `bd create --parent wy-y5uv` then minted `wy-y5uv.1` and `wy-y5uv.2` — colliding with the archived `.1`/`.2`. A code comment in that repo (`rpc-constants.ts`) cites the *old* `wy-y5uv.6`, so the ID namespace is now ambiguous.
- Workaround used: `bd delete` the recycled IDs, recreate with `bd create --id wy-y5uv.8` (note `--id` and `--parent` are mutually exclusive), then reparent with `bd update --parent`.

The same DB **today** shows the counter is not a high-water mark:

```
child_counters: wy-y5uv -> last_child = 2
issues:         wy-y5uv.8, .9, .10, .12 all live
```

i.e. children created via explicit `--id` never raised the counter.

Auditing that DB against its archive JSONL: **36 of the 56 parents that have archived children are currently primed to recycle an archived suffix** on their next `bd create --parent` — including the repo's largest epic, whose next mint would be `.1` against an archive holding `.1`–`.45`.

## Root cause (paths that fail to maintain the counter)

[...]

PR #5122, opened 2026-07-29
fix(storage): advance child_counters on singular hierarchical create (#4750)

vishnujayvel:
# fix(storage): advance child_counters on singular hierarchical create

**Partial fix for #4750**

## Summary

`child_counters.last_child` is supposed to be a high-water mark for
`bd create --parent`, but only the **batch** create path called
`ReconcileChildCounters`. A singular `bd create --id P.8` left the counter
behind, so a later `--parent P` could mint a lower suffix after explicit IDs
or imports.

## Change

After a successful singular `CreateIssueInTx` of a new hierarchical ID, call
`ReconcileChildCounters` so `last_child` catches up.

## Out of scope (still open on #4750)

- IDs that exist only in archive JSONL (not in `issues`) are still invisible
  to the allocator — needs a separate high-water source from archives.

## Test plan

- [x] `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test ./internal/storage/embeddeddolt/ -run 'TestGetNextChildID/explicit_id_create'`
  — counter = 8 after `CreateIssue` with ID `P.8`; next mint is `P.9`

## Notes

Scope: this closes root cause 1 of #4750 only — the stale `child_counters` row on the single-issue explicit-ID create path. The `isNew` guard means a singular *upsert* of an existing dotted ID still does not raise the counter, and compaction-before-delete, JSONL export/import durability, and the `ON DELETE CASCADE` wipe all remain. #4750 stays open.

*(Corrected 2026-07-29 — the test command above originally read `-tags cgo`. As @maphew pointed out, `child_id_test.go` carries `//go:build cgo`, which is satisfied by `CGO_ENABLED=1` rather than a `-tags` value, and the subtests are additionally gated on `BEADS_TEST_EMBEDDED_DOLT=1`. The command as originally written would not have run the test. CI sets both, so the coverage itself is genuine.)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -998,6 +998,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `bd.claim_verify_lost_total` and `bd.claim_verify_recovered_total` count
   loud failures and converted outcomes.
 
+- **Single-issue `bd create --id P.N` no longer leaves `child_counters` stale**
+  (one root cause of [#4750](https://github.com/gastownhall/beads/issues/4750);
+  the issue stays open — three other drift paths remain). `CreateIssueInTxWithResult`
+  only reconciled `child_counters` when the batch path called
+  `ReconcileChildCounters`; an explicit hierarchical `--id` never advanced the
+  parent's `last_child` high-water mark. Once such children were later
+  archived out of `issues`, `GetNextChildIDTx` no longer saw them and could
+  re-mint an already-used suffix — an ID collision with an archived issue.
+  The single-issue path now calls `ReconcileChildCounters` too, and its
+  changed tables are merged into the result so the reconcile actually gets
+  staged for commit.
+
 ## [1.1.2] - 2026-07-26
 
 Hotfix release cut from v1.1.0. (There is no 1.1.1 release: its tag was cut
--- a/backend/conformance/portable.go
+++ b/backend/conformance/portable.go
@@ -677,9 +677,9 @@ func testCreateIssuesWithFullOptions(t *testing.T, f Factory) {
 
 // Batch-creating dotted hierarchical child IDs runs ReconcileChildCounters, which
 // emits `... ON DUPLICATE KEY UPDATE last_child = GREATEST(last_child, ?)` in the
-// child-counter upsert (issueops/create.go). That GREATEST path is reachable only
-// from batch create — the single-issue path never reconciles counters — so it was
-// dead in the rest of the suite, which is how a missing SQLite GREATEST translation
+// child-counter upsert (issueops/create.go). Both the batch and single-issue create
+// paths reach that GREATEST upsert (GH#4750), but this suite historically only
+// exercised it via batch create, which is how a missing SQLite GREATEST translation
 // (child creation aborted with "no such function: GREATEST") shipped green. This
 // scenario mints x-1, x-1.1, x-1.2 through the batch path and asserts every backend
 // both creates the children and reconciles the counter to the max direct child.
--- a/internal/storage/embeddeddolt/child_id_test.go
+++ b/internal/storage/embeddeddolt/child_id_test.go
@@ -108,6 +108,52 @@ func TestGetNextChildID(t *testing.T) {
 		}
 	})
 
+	// GH#4750: singular create with an explicit hierarchical ID must raise
+	// child_counters so subsequent --parent mints do not restart below it.
+	t.Run("explicit_id_create_advances_counter", func(t *testing.T) {
+		te := newTestEnv(t, "ex")
+		ctx := t.Context()
+
+		parent := &types.Issue{
+			ID:        "ex-parent",
+			Title:     "Parent",
+			Status:    types.StatusOpen,
+			Priority:  2,
+			IssueType: types.TypeEpic,
+		}
+		if err := te.store.CreateIssue(ctx, parent, "tester"); err != nil {
+			t.Fatalf("CreateIssue parent: %v", err)
+		}
+
+		// Simulate `bd create --id ex-parent.8` (no prior counter activity).
+		child := &types.Issue{
+			ID:        "ex-parent.8",
+			Title:     "Explicit child 8",
+			Status:    types.StatusOpen,
+			Priority:  2,
+			IssueType: types.TypeTask,
+		}
+		if err := te.store.CreateIssue(ctx, child, "tester"); err != nil {
+			t.Fatalf("CreateIssue explicit child: %v", err)
+		}
+
+		var lastChild int
+		te.queryScalar(t, ctx,
+			"SELECT last_child FROM child_counters WHERE parent_id = ?",
+			[]any{"ex-parent"}, &lastChild)
+		if lastChild != 8 {
+			t.Fatalf("child_counters last_child = %d, want 8 after explicit --id create", lastChild)
+		}
+
+		next, err := te.store.GetNextChildID(ctx, "ex-parent")
+		if err != nil {
+			t.Fatalf("GetNextChildID: %v", err)
+		}
+		if next != "ex-parent.9" {
+			t.Fatalf("GetNextChildID = %q, want ex-parent.9", next)
+		}
+	})
+
 	t.Run("wisp_parent_uses_wisp_counter", func(t *testing.T) {
 		te := newTestEnv(t, "wp")
 		ctx := t.Context()
--- a/internal/storage/issueops/create.go
+++ b/internal/storage/issueops/create.go
@@ -164,6 +164,21 @@ func CreateIssueInTxWithResult(ctx context.Context, tx DBTX, bc *BatchContext, i
 		return result, err
 	}
 	result.ChangedTables = mergeChangedTables(result.ChangedTables, commentResult.ChangedTables)
+
+	// Advance child_counters when a singular create materializes a hierarchical
+	// ID (e.g. bd create --id P.8). The batch path already calls
+	// ReconcileChildCounters after CreateIssuesInTx; without this, explicit --id
+	// creates leave last_child behind the live suffix high-water mark and the
+	// next bd create --parent can recycle lower suffixes (GH#4750).
+	if isNew {
+		if _, childNum, ok := ParseHierarchicalID(issue.ID); ok && childNum > 0 {
+			changedCounters, err := ReconcileChildCounters(ctx, tx, []*types.Issue{issue})
+			if err != nil {
+				return result, err
+			}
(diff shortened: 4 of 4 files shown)


maphew (commented):
Nice catch on the actual reachability here: this is the one path in `#4750` that is both real today and cheap to close, and the patch lands it in the right place. `CreateIssueInTxWithResult` really is what `bd create --id P.8` executes — `createIssueWithDeps` (cmd/bd/create_atomic.go:36-57) routes both the bare and the with-deps form into `CreateIssue`, which reaches this function through every backend — so the fix covers the CLI surface the issue reports, not just a library corner.

I verified the claims rather than taking the description on faith:

- The bug exists on `upstream/main`: `CreateIssueInTxWithResult` (internal/storage/issueops/create.go:85-167) writes issues, events, labels and comments and never touches `child_counters`; the only reconcile is the batch one at create.go:232. `GetNextChildIDTx` (internal/storage/issueops/child_id.go:15-51) takes `max(last_child, max suffix among live rows)`, so once the children are archived out of `issues` the stale counter is the only guard — matching the `wy-y5uv` drift in the issue.
- The test is load-bearing, not decorative. I reverted only the `create.go` hunk on the PR head and re-ran the new subtest: it fails with `queryScalar "SELECT last_child FROM child_counters WHERE parent_id = ?": sql: no rows in result set`. With the hunk it passes.
- No regression locally: `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go ./internal/storage/issueops/... ./internal/storage/embeddeddolt/...` is green (414s), and all 81 CI checks pass.
- Staging is handled correctly, which is easy to get wrong here: `CreateIssueDirtyTables` (create.go:255-266) previously learned about `child_counters` only via `storage.HasReservedChildCounter`, which is false for an explicit `--id`. Merging the reconcile's changed tables into `result.ChangedTables` is what makes the new write get staged, and `stageableChangedTables` still filters `wisp_child_counters` for a wisp parent. Orphan and same-batch-parent cases are safe too: `ReconcileChildCounters` skips on `sql.ErrNoRows` for the parent row, so no FK violation.

Two things I'd like to see before this merges, neither of which is a correctness problem:

1. **Skip the reconcile when called from the batch loop.** internal/storage/issueops/create.go:174-182 sits inside the function that `CreateIssuesInTxWithResult` calls per issue (create.go:207-214) before reconciling the entire slice at create.go:232. So a batch import now pays, per hierarchical issue, a `wispsTableEmptyOrMissingInTx` probe plus `SELECT id FROM wisps WHERE id IN (...)` (internal/storage/issueops/wisp_routing.go:72-90), a `SELECT 1 FROM issues WHERE id = ?`, and a `SELECT last_child ...` — 3-4 round trips whose result the trailing batch call recomputes anyway. Concrete fix: add `SkipChildCounterReconcile bool` to `BatchContext`, set it in `CreateIssuesInTxWithResult` right after `NewBatchContext`, and gate the new block on it.

2. **Update the now-false comment at internal/storage/conformance/portable.go:627-628**, which states "That GREATEST path is reachable only from batch create — the single-issue path never reconciles counters". After this change `bd create --id P.8` emits the GREATEST upsert as well. That comment is the recorded explanation for how a missing SQLite `GREATEST` translation shipped green, so leaving it stale costs a future reader real time.

Smaller notes:

- The PR body ends with "Stage-only until Vishnu says go." — that reads as a hold signal to a reviewer; worth removing.
- The stated test command won't run the test as written: `child_id_test.go` is `//go:build cgo` (satisfied by `CGO_ENABLED=1`, not `-tags cgo`) and the subtests are gated on `BEADS_TEST_EMBEDDED_DOLT=1` (internal/storage/embeddeddolt/create_issue_test.go:172-177). CI does set that env var, so coverage is genuine.
- Scope is honestly described. Worth stating explicitly in #4750 that this closes root cause 1 only: the `isNew` guard means a singular *upsert* of an existing dotted ID still doesn't raise the counter, and compaction-before-delete, JSONL export/import durability, and the `ON DELETE CASCADE` wipe all remain. #4750 should stay open.
- No CHANGELOG entry, unlike your #4858. CI doesn't enforce it, so take it or leave it.

Happy to push (1) and (2) as maintainer commits on your branch if you'd rather not round-trip — or they're yours if you want them. Either way this is good, targeted work on a bug that was genuinely hard to see.

vishnujayvel:
Thanks — the batch-loop double-reconcile catch is the useful one: `CreateIssuesInTxWithResult` calls the per-issue path before its own slice-wide reconcile, so a batch import was paying 3-4 extra round trips per hierarchical issue for a result it discards. Good find.

Correction: the stated test command doesn't run the test as written — `child_id_test.go` needs `CGO_ENABLED=1`, not `-tags cgo`, plus `BEADS_TEST_EMBEDDED_DOLT=1`. CI does set that env var so coverage is real, but the command in our test plan was wrong. Also, "Stage-only until Vishnu says go." in the body was a leftover internal note — will strip it.

Please push the `SkipChildCounterReconcile` flag on `BatchContext` (item 1) and the stale comment fix at `portable.go:627-628` (item 2) as maintainer commits. We'll add the CHANGELOG entry ourselves and state in #4750 that this closes root cause 1 only, per your scope note — #4750 stays open.

maphew:
Pushed both as `31388add6`.

`SkipChildCounterReconcile` is on `BatchContext` in `internal/storage/issueops/create.go`, set by `CreateIssuesInTxWithResult` right after it builds the context, and honoured by the per-issue path's `if isNew && !bc.SkipChildCounterReconcile` guard. Every other caller builds a fresh context through `NewBatchContext` and never touches the field, so the single-issue path is unchanged.

I checked the coverage claim rather than assuming it, since a wrong answer there would be a correctness regression rather than a perf win. `CreateIssuesInTxWithResult` calls `ReconcileChildCounters(ctx, tx, issues)` with the full `accepted` slice, and that function computes the max child suffix per parent across the whole set before upserting each parent once. The only issues missing from `accepted` are the stale-rejected ones, which the per-issue reconcile never reached either — `staleRejected` returns before the `isNew` branch. So the batch call strictly subsumes N single-issue calls: same end state, fewer round trips.

The `portable.go` comment now says both create paths reach the `GREATEST` upsert and that the suite historically only exercised it via batch create, rather than claiming the single-issue path never reconciles.

One thing to be aware of, not caused by either of us: `TestIgnoredMigration0011CleansOrphanedChildCounters` fails in `internal/storage/dolt` (`orphaned counter rows = 1, want 0`). I re-ran it with my diff stashed against your unmodified head and it fails identically, so it is pre-existing.

CHANGELOG and the #4750 scope note are yours as agreed.

vishnujayvel:
Thanks for `31388add6`, and for checking the subsumption claim rather than assuming it — that was the part that could have turned a perf win into a correctness regression, and your reasoning about `staleRejected` returning before the `isNew` branch is the piece I would have missed.

Taking the two items you left me:

**CHANGELOG** — added in `ec0771483`, in the newest `### Fixed` block under `[Unreleased]`. It states plainly that this closes one root cause and that #4750 stays open, following the "reporter fix No.2 for #4259" framing already in the file. No credit line, since the merged CHANGELOG has no `reported by` precedent — say the word if you'd rather it credited @steveyegge as the reporter and I'll add it.

**Body corrections, since the squash message is what survives** — I removed a leftover internal note ("Stage-only until Vishnu says go"), which should never have been in a PR body; thank you for flagging it. I also corrected the stated test command: it read `-tags cgo`, which as you point out does not satisfy `//go:build cgo` — that needs `CGO_ENABLED=1` — and it omitted `BEADS_TEST_EMBEDDED_DOLT=1`, so the command as written would not have run the test at all. It now reads `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test ...`. The coverage was genuine because CI sets both, but the command I published was wrong and would have wasted the time of anyone reproducing it. And I added the scope paragraph to the body naming the four paths that remain.

**#4750 scope note** — drafted, not yet posted; I'll get it up shortly.

Noting your `TestIgnoredMigration0011CleansOrphanedChildCounters` observation. I have a live interest in that class of claim: a "pre-existing on main" call of mine last week turned out to be my own machine's global `git config beads.role=maintainer` leaking into tests, so I now run with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`. Your stash-and-rerun against my unmodified head is the right control and I'm not disputing it — just flagging the failure mode in case it saves you a round somewhere else.

maphew (approved):
Reviewed at `ec077148360`.

No correctness findings. Singular hierarchical creates reconcile the child counter only when a row was newly materialized; batch creates suppress redundant per-item reconciliation and retain their existing slice-wide pass. Changed counter tables flow into dirty-table staging with correct wisp routing, and the changelog scope is accurate.

Verification:

- `go test -tags gms_pure_go ./internal/storage/embeddeddolt -run '^TestGetNextChildID$/explicit_id_create_advances_counter$' -count=1` — PASS (`ok`, 6.8s).
- Blocking upstream preflight — branch conflicts with current main but reported no failed or pending checks; no merge action was taken.

Disposition: **approve the change; merge after conflict resolution and a clean preflight**.

vishnujayvel:
Rebased onto current main (`4ebc989`). The CHANGELOG conflict resolved as pure concatenation — our #4750 bullet stays in [Unreleased], the 1.1.2 hotfix section follows unchanged.

One disclosed change beyond the reviewed diff: main's new conformance comment in `backend/conformance/portable.go` (the GREATEST-upsert scenario note) described the pre-#4750 behavior — "the single-issue path never reconciles counters" — which this PR changes. Left as-is it would have been false after merge, so I updated those three lines alongside the rebase. Code files are otherwise byte-identical to what you reviewed at `ec07714`; `TestGetNextChildID` (all six subtests) passes on the rebased head.

steveyegge:
Delta review of the post-approval commits (maphew's approval at `ec0771483` predates the rebase + perf commit, so re-verifying the delta before merging on it):

- `4ea7a4cba` (SkipChildCounterReconcile): verified the flag mutation is safe — `CreateIssuesInTxWithResult` builds its own function-local `BatchContext` (create.go:231), and the two `CreateIssueInTxWithResult` callers in embeddeddolt construct fresh contexts per call, so setting `bc.SkipChildCounterReconcile = true` inside the batch path can never leak into a later singular create.
- The disclosed `portable.go` comment fix is accurate post-#4750 (both paths now reach the GREATEST upsert).
- Rebase is clean concatenation in CHANGELOG as described.

CI fully green at `a12cf9978` (85 pass). Merging on the standing approval + this delta verdict.

Issue #2100, opened 2026-02-25
Inherit parent labels when creating child issues


m-gris:
First off — the shared-database-across-worktrees design is great. Combined with `branch:` labels for scoping, it gives us cross-branch visibility without the merge headaches of per-branch databases. We have been using this pattern heavily in a multi-worktree setup.

One friction point we have noticed: when creating child issues under an epic via `--parent`, labels do not carry over — so after breaking down an epic into, say, 8 tasks, we need a separate `bd label add .1 .2 .3 ... "branch:feature-x"` step to make the children visible to branch-scoped `bd ready --label 'branch:marc'`.

Since beads already propagates blocked/deferred status through the parent-child tree, it seemed like labels could follow a similar path — at least at creation time.

### Our thinking on the semantics

We would distinguish between **inheriting** and **propagating**:

- **Inherit** (at creation): when a child is created with `--parent`, it automatically gets the parent's current labels. One-time snapshot, no ongoing coupling. This feels like it could be the default — a child naturally belongs to the same scope as its parent.
- **Propagate** (after the fact): an explicit command like `bd label propagate <epic-id>` to push a label down to existing children. This is a deliberate act on things that already exist, so it makes sense as opt-in.

A `--no-inherit-labels` flag on `bd create` could serve as the escape hatch for cases where you specifically want a child with different labels.

### Use case for context

We use parallel git worktrees and `branch:` labels to scope beads work per-worktree. When breaking down an epic into child tasks, the branch label needs to be on every child for `bd ready` to surface them in the right worktree. Inheritance at creation would make this seamless.

PR #5123, opened 2026-07-29
docs: caveat --parent label inheritance with size/effort labels (#4562)

### LINKED ISSUE #4562: skill/docs: `--parent` examples silently double-label children with the epic's size label — suggest `--no-inherit-labels` in examples (interaction of #2100 default with LABELS.md size labels)
## Summary

The distributed beads skill (`beads-task-management` SKILL.md, v0.34.0) and docs show child-card creation with `--parent` in every example without mentioning that **children copy the parent's labels by default** (the intentional behavior from #2100). When teams follow `docs/LABELS.md`'s own recommendation to use size/effort labels, that combination silently double-labels every child with the epic's size marker. We suggest updating the skill/doc examples to include `--no-inherit-labels` (or a one-line caveat about the default) wherever `--parent` appears.

## What happens in practice

`docs/LABELS.md` recommends effort labels as a core label category ("Size/Effort Estimates: small / medium / large"). We follow that convention with finer granularity: epics carry an epic-scale marker (`sp:13`) and each child carries its own points label (`sp:2`, `sp:3`, `sp:5`, ...).

Because `bd create --parent <epic>` inherits the parent's labels by default, every child was also stamped with the epic's `sp:13`. Measured on two freshly created boards (bd 1.0.5, 2026-07-03): **22 of 22 parented task cards carried two size labels** — the epic's marker plus their own:

```
| issue_id                | issue_type | sp_labels  |
| claude-statusline-…​.1  | task       | sp:13,sp:2 |
| claude-statusline-…​.3  | task       | sp:13,sp:3 |
| …​ (22 rows total across 2 boards)                |
```

Consequences: a card's size becomes ambiguous at a glance, and any label-based query or report over effort labels double-counts (`bd list -l sp:13` returns every child of every epic, not the epic-scale items).

The skill's own "Epic with Children" examples produce exactly this state:

```bash
# from SKILL.md — every child inherits the epic's labels here
bd create "Research OAuth providers" -p 1 --parent myproject-abc
bd create "Implement auth endpoints" -p 1 --parent myproject-abc
```

## Suggested change

Not a behavior change — #2100's default is fine for team/domain labels, where inheritance is exactly what you want. This is a docs/skill-content request:

1. In the skill's `--parent` examples (SKILL.md "Epic with Children" and the extended OAuth walkthrough), add `--no-inherit-labels` with a one-line comment, e.g.:

   ```bash
   # --no-inherit-labels: keep epic-level labels (e.g. size markers) off the children
   bd create "Research OAuth providers" -p 1 --parent myproject-abc --no-inherit-labels
   ```

2. In `docs/LABELS.md`'s size/effort section, add a caveat: size labels + label inheritance interact — parented children inherit the epic's size label unless `--no-inherit-labels` is passed.

## Why it's worth doing

The two features are individually good and documented, but their interaction is a trap that only shows up later as dirty data: users discover it after boards are populated, then need a cleanup pass (ours: a `GROUP BY issue_id HAVING COUNT(*)>1` sweep over the labels table plus 22 `bd label remove` calls). One flag in the examples — or one caveat sentence — prevents the whole class.

Environment: bd 1.0.5 (Homebrew, macOS), server mode against a shared dolt sql-server; skill v0.34.0.

vishnujayvel:
# docs: caveat --parent label inheritance with size/effort labels

**Closes #4562**

## Summary

`bd create --parent` inherits parent labels by default (#2100). That is right for
team/domain labels, but combined with size/effort labels on epics it silently
double-labels every child and breaks effort queries (`bd list -l large` returns
the whole tree).

Docs/skill examples used bare `--parent` with no mention of this interaction.

## Change

- `docs/core-concepts/labels.md` — Size/Effort section caveat + example with
  `--no-inherit-labels`
- Parent-create examples in quickstart, molecules, multi-agent coordination,
  skill `epic` command, and integration patterns

No behavior change — documentation only.

## Test plan

- [x] Docs read-through for consistency
- [ ] Optional: follow quickstart epic/child steps and confirm children lack epic size labels when flag is used

### DIFF
--- a/docs/core-concepts/labels.md
+++ b/docs/core-concepts/labels.md
@@ -109,7 +109,7 @@ large     # > 3 days
 ```
 
 **Interaction with label inheritance:** `bd create --parent` copies the parent's
-labels onto the child by default (see #2100). If the epic carries a size marker
+labels onto the child by default (see GH#2100). If the epic carries a size marker
 (e.g. `large` or `sp:13`) and children have their own estimates, every child
 will also inherit the epic's size label — so `bd list -l large` returns the
 whole tree, not just epic-scale work. For per-child size labels, pass
@@ -122,8 +122,7 @@ shared can still be added explicitly).
 bd ready --json | jq '.[] | select(.labels[] == "small")'
 
 # Child keeps its own size label without inheriting the epic's
-bd create "Implement auth endpoints" -p 1 --parent bd-a3f8e9 --no-inherit-labels
-bd label add <child-id> small
+bd create "Implement auth endpoints" -p 1 --parent bd-a3f8e9 --no-inherit-labels -l small
 ```
 
 ### 4. Quality Gates
--- a/docs/core-concepts/labels.md
+++ b/docs/core-concepts/labels.md
@@ -108,10 +108,22 @@ medium    # 1-3 days
 large     # > 3 days
 ```
 
+**Interaction with label inheritance:** `bd create --parent` copies the parent's
+labels onto the child by default (see #2100). If the epic carries a size marker
+(e.g. `large` or `sp:13`) and children have their own estimates, every child
+will also inherit the epic's size label — so `bd list -l large` returns the
+whole tree, not just epic-scale work. For per-child size labels, pass
+`--no-inherit-labels` when creating children (team/domain labels you *do* want
+shared can still be added explicitly).
+
 **Example:**
 ```bash
 # Find small quick wins
 bd ready --json | jq '.[] | select(.labels[] == "small")'
+
+# Child keeps its own size label without inheriting the epic's
+bd create "Implement auth endpoints" -p 1 --parent bd-a3f8e9 --no-inherit-labels
+bd label add <child-id> small
 ```
 
 ### 4. Quality Gates
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -128,11 +128,10 @@ For large features, use hierarchical IDs to organize work:
 bd create "Auth System" -t epic -p 1
 # Returns: bd-a3f8e9
 
-# Create child tasks (use --parent to attach to the epic).
-# --no-inherit-labels keeps epic-level size/effort labels off children (GH#4562).
-bd create "Design login UI" -p 1 --parent bd-a3f8e9 --no-inherit-labels       # bd-a3f8e9.1
-bd create "Backend validation" -p 1 --parent bd-a3f8e9 --no-inherit-labels    # bd-a3f8e9.2
-bd create "Integration tests" -p 1 --parent bd-a3f8e9 --no-inherit-labels     # bd-a3f8e9.3
+# Create child tasks (use --parent to attach to the epic)
+bd create "Design login UI" -p 1 --parent bd-a3f8e9       # bd-a3f8e9.1
+bd create "Backend validation" -p 1 --parent bd-a3f8e9    # bd-a3f8e9.2
+bd create "Integration tests" -p 1 --parent bd-a3f8e9     # bd-a3f8e9.3
 
 # View hierarchy
 bd dep tree bd-a3f8e9
@@ -147,6 +146,8 @@ Dependency tree for bd-a3f8e9:
   > bd-a3f8e9.2: Backend validation [P1] (open)
   > bd-a3f8e9.3: Integration tests [P1] (open)
 ```
+
+Children inherit the epic's labels by default — if the epic carries a size/effort label, see [Labels](/core-concepts/labels) for how to keep it off the children.
 
 ## Add dependencies
 
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -128,10 +128,11 @@ For large features, use hierarchical IDs to organize work:
 bd create "Auth System" -t epic -p 1
 # Returns: bd-a3f8e9
 
-# Create child tasks (use --parent to attach to the epic)
-bd create "Design login UI" -p 1 --parent bd-a3f8e9       # bd-a3f8e9.1
-bd create "Backend validation" -p 1 --parent bd-a3f8e9    # bd-a3f8e9.2
-bd create "Integration tests" -p 1 --parent bd-a3f8e9     # bd-a3f8e9.3
+# Create child tasks (use --parent to attach to the epic).
+# --no-inherit-labels keeps epic-level size/effort labels off children (GH#4562).
+bd create "Design login UI" -p 1 --parent bd-a3f8e9 --no-inherit-labels       # bd-a3f8e9.1
+bd create "Backend validation" -p 1 --parent bd-a3f8e9 --no-inherit-labels    # bd-a3f8e9.2
+bd create "Integration tests" -p 1 --parent bd-a3f8e9 --no-inherit-labels     # bd-a3f8e9.3
 
 # View hierarchy
 bd dep tree bd-a3f8e9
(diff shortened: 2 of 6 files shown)


maphew (commented):
The caveat paragraph in `docs/core-concepts/labels.md:111-117` is the right catch and it holds up against the code — I checked rather than assumed:

- `cmd/bd/create.go:451-455` — with `--parent` set and `no-inherit-labels` false, `GetLabels(ctx, parentID)` copies **all** parent labels onto the child.
- `cmd/bd/create.go:773` — the `--no-inherit-labels` flag exists and does what the docs now say.
- `cmd/bd/list.go:735` — `-l/--label` is an AND filter, so the claim that `bd list -l large` returns the whole tree is accurate.

So #4562's report is real, the interaction was genuinely undocumented, and this paragraph is a good permanent home for it. That part I'd take as-is.

My hesitation is with the other five files, where the change is applying the flag to examples rather than documenting the interaction.

**1. The flag is a no-op in the examples it was added to.**

- `docs/getting-started/quickstart.md:128-134` — the epic is `bd create "Auth System" -t epic -p 1`, no labels.
- `docs/workflows/molecules.md:48-52` — `bd create "Feature X" -t epic`, no labels.
- `docs/multi-agent/coordination.md:80-83` — `bd-epic`, no labels.

In all three, nothing is inherited with or without the flag. The effect on a reader is to make `--no-inherit-labels` look like the standard way to create a child, which inverts the deliberate default from #2100 — and #4562 itself says that default is correct for team/domain labels. A quickstart is the worst place to teach an opt-out for a problem the reader does not have yet.

Suggested shape: keep the labels.md caveat and its example, revert quickstart/molecules/coordination, and if the hierarchy sections should mention it at all, add a one-line cross-link ("epics with size labels: see Labels → Size/Effort") rather than editing the commands.

**2. The sweep is incomplete either way.** `docs/core-concepts/issues.md:95` and `:123`, and `docs/reference/faq.md:85-86`, still show bare `--parent`. If the strategy is "flag in every `--parent` example", those are missing; if it's "one caveat plus one worked example", the five edits aren't needed. Right now it's half of each, which will read as inconsistent to the next person editing these files.

**3. Smaller things**

- `docs/core-concepts/labels.md:125-126` — two commands where one does the job: `bd create "Implement auth endpoints" -p 1 --parent bd-a3f8e9 --no-inherit-labels -l small`. With inheritance off, explicit `-l` labels pass straight through `mergeCreateLabels` (`cmd/bd/create.go:457`), and it avoids the `<child-id>` placeholder the reader can't fill in until after the first command runs.
- Issue-reference style is mixed within the PR: bare `#2100` at `labels.md:112` vs a `GH`-prefixed #4562 at `quickstart.md:132` and `plugins/beads/skills/beads/commands/epic.md:26`. Neither links in the rendered docs. Existing docs precedent is the `GH`-prefixed form (e.g. `docs/architecture/dolt.md:434`) — or just drop the reference from quickstart, where an internal issue number is noise for a new user.
- The PR body ends with "Stage-only until Vishnu says go." — looks like a leftover internal note.
- #4562 asked for the skill's SKILL.md "Epic with Children" examples; SKILL.md no longer carries `--parent` examples on main, so retargeting to `commands/epic.md` was the correct call. Worth saying so in the body, otherwise it reads as a missed file.

**Related work worth knowing about:** #4757 (open) adds flag-gated exclusive label namespaces (`labels.exclusive-prefixes`). That would let a workspace make `sp:`/size labels mutually exclusive and fix this class mechanically. It doesn't block this PR — docs should describe today's behavior — but if #4757 lands, the labels.md caveat should point at it.

Checks are green and the branch merges clean. Happy to trim this to the labels.md caveat plus cross-links and land it myself if you'd rather not do another round — say the word and it's yours either way. Documentation of a trap that only surfaces after a board is already dirty is worth this much attention; thanks for chasing the report all the way to a patch.

vishnujayvel:
Thanks — the point that `--no-inherit-labels` is a no-op in the three examples we added it to (no labels present, so nothing to inherit either way) is the real problem: it teaches the flag as the default way to create a child, which inverts the #2100 default we agree is correct. Good catch.

No false claims to correct here — the labels.md caveat itself checks out against the code as you verified.

We'll take your suggested shape: keep the `docs/core-concepts/labels.md` caveat and its worked example, revert the quickstart/molecules/coordination edits, and add a one-line cross-link instead of touching commands. We won't extend the sweep to `issues.md`/`faq.md`.

Please push, as maintainer commits: the two-command consolidation at `labels.md:125-126`, and fixing the issue-reference style to the `GH`-prefixed form (or dropping it from quickstart entirely). We'll add a line to the PR body noting the SKILL.md retarget to `commands/epic.md` was deliberate, not a miss.

maphew:
Pushed both as `bdf815444`, touching only `docs/core-concepts/labels.md` so it does not collide with the revert and cross-link work you are keeping.

The consolidation folds the follow-up `bd label add <child-id> small` into the create call as `-l small`, which also gets rid of the `<child-id>` placeholder the reader had to mentally substitute.

On the reference style, I went with the `GH#` form rather than dropping it. I checked what the docs tree already does outside this PR and the plain-text `(GH#NNNN)` pattern is consistent across five files that predate it — `docs/cli-reference/config.md`, `docs/CLI_REFERENCE.md`, `docs/reference/configuration.md`, `docs/architecture/dolt.md`, and a linked variant in `docs/reference/troubleshooting.md`. So the only non-conforming reference was the bare `#2100` in the caveat text, now `GH#2100`. Your quickstart line was already in the right form, so I left quickstart untouched.

Over to you for the reverts and the cross-link.

vishnujayvel:
Your point 1 is the one I should have caught myself: in all three of those examples the epic carries no labels, so the flag is inert and the only thing it does is teach a new reader that opting out is the standard way to create a child — inverting the #2100 default that #4562 itself says is correct. That's a docs change that makes the docs worse, and it took a reviewer to see it.

Pushed in `527a036770`, taking your "one caveat plus one worked example" shape:

- `quickstart.md`, `molecules.md`, `coordination.md` — command examples reverted to `main` byte-for-byte, each replaced with a one-line cross-link to the labels caveat in your suggested phrasing.
- `plugins/beads/skills/beads/resources/INTEGRATION_PATTERNS.md` — same no-op problem, which you didn't name but which fell out of the sweep. Reverted, no cross-link; a buried integration checklist isn't where a reader meets this trap.
- `labels.md` left alone — it's yours now. Noting only that `bdf815444` doesn't carry the #4757 pointer you mentioned; I didn't add it.

`commands/epic.md` I kept. Its edit is a prose caveat below the example rather than a change to the command, so it doesn't have the optics problem, and it's where SKILL.md's old "Epic Workflow" section — the exact thing #4562 reported against — now lives. It already used `GH#4562`.

Gates: `go test ./test/docsync`, `check-doc-freshness.sh`, `generate-cli-docs.sh --check`, and `check-doc-flags.sh` all pass. One honest gap: `./mint.sh broken-links` won't run here — this machine is on Node 25 and Mintlify wants 22 LTS. I confirmed `/core-concepts/labels` resolves in `docs.json` by hand, but I haven't seen the real link check green and would rather you knew that than assumed it.

**On `Closes #4562` after the reverts** — I think it still holds, but I'd rather you check my reasoning than take it. The issue asked for two things: the skill's `--parent` examples get the flag "or a one-line caveat about the default", and `LABELS.md`'s size/effort section gets a caveat. The second is satisfied by your `bdf815444` on top of `a9ec92621`; the first by the `epic.md` caveat, which is the alternative form the issue explicitly sanctioned. Your objection was to over-application, not under-application. If you read it as thinner than that, say so and I'll change it to `Refs` — I'd rather leave a live issue open than auto-close one on a partial.

PR #5126, opened 2026-07-29
refactor(cli): rename CPU profiling flag

julianknutsen:
## What

Rename the global CPU/trace diagnostic switch from `--profile` to
`--cpu-profile` and update its focused regression coverage and in-tree comment.

## Why

`--profile NAME` is the conventional namespace for selecting a named CLI
configuration. The existing boolean diagnostic flag prevents that interface
from being added and is less precise than `--cpu-profile` about what it does.

The profiling implementation and generated filenames are unchanged. This PR
intentionally does not edit the release-pinned generated CLI reference; it will
pick up the new spelling when the documentation pin advances.

## Validation

```text
GOCACHE=/tmp/bd-cpu-profile-gocache CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run '^TestCPUProfileFlagRegistered$' -count=1
git diff --check
pre-commit golangci-lint: 0 issues
```

_codex-gpt-5.6-terra-high on behalf of CI Bot_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,7 +101,7 @@ __pycache__/
 .Python
 .envrc
 
-# Performance profiling files (benchmarks, bd doctor --perf, and bd --profile)
+# Performance profiling files (benchmarks, bd doctor --perf, and bd --cpu-profile)
 *.prof
 *.out
 beads-perf-*.prof
--- a/cmd/bd/cpu_profile_test.go
+++ b/cmd/bd/cpu_profile_test.go
@@ -1,0 +1,19 @@
+package main
+
+import "testing"
+
+func TestCPUProfileFlagRegistered(t *testing.T) {
+	flag := rootCmd.PersistentFlags().Lookup("cpu-profile")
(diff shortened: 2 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98636947

maphew (approved):
Verified the rename is complete and self-consistent: `profileEnabled` had exactly the three uses this PR touches (cmd/bd/main.go:90, :686, :1063), and the only remaining `--profile` reference on main is the release-pinned `docs/CLI_REFERENCE.md:306`, which the PR correctly leaves for the next doc-pin advance. The new regression test pins registration, type, usage string, and absence of the old spelling.

Non-blocking thought: since `--profile` was user-visible, a hidden deprecated bool alias (`Flags().MarkHidden`) for one release would soften the break for anyone scripting it. Given it's a diagnostic flag, I don't consider that required.

PR #5128, opened 2026-07-29
fix(federation): treat missing peer row as credential-free remote

### LINKED ISSUE #4837: federation status rejects credential-free Dolt remotes in server mode
## Description

In Dolt server mode, `bd federation status` cannot fetch an ordinary credential-free remote even though `bd dolt pull` and `bd dolt push` succeed against the same remote.

## Environment

- `bd` 1.1.0
- `dolt` 2.1.10
- shared-server mode on macOS arm64

## Reproduction

1. Bootstrap a database from a Dolt remote using `sync.remote` without SQL username/password fields.
2. Confirm `bd federation list-peers --json` lists `origin` from Dolt remotes.
3. Confirm `bd dolt pull` and `bd dolt push` succeed using normal Dolt credential discovery.
4. Run `bd federation status --json`.

Observed:

```json
{
  "Reachable": false,
  "ReachError": "failed to get peer credentials: not found: federation peer origin"
}
```

Expected: absence of a `federation_peers` credential row means the remote has no peer-specific SQL credentials; fetch should proceed with empty credentials and normal Dolt credential discovery.

## Root cause

`runFederationStatus()` enumerates all Dolt remotes and calls `Fetch()`. `Fetch()` calls `withPeerCredentials()`, which treats `GetFederationPeer(...)=ErrNotFound` as fatal. However, `runFederationAddPeer()` calls only `AddRemote()` when `--user` is omitted, so this valid CLI path creates no `federation_peers` row.

## Suggested fix

Treat a missing credential row as empty credentials when the named Dolt remote exists. Keep malformed/decryption/database errors fatal. Alternatively, make credential-free `add-peer` persist a peer row, but existing bootstrapped remotes must continue to work without migration.

Add server- and embedded-mode tests covering:

- remote with no credential row;
- peer with valid stored SQL credentials;
- malformed/encrypted credential failures remaining explicit;
- status/fetch parity with `bd dolt pull` for a local file remote.

vishnujayvel:
## Summary

Fixes #4837.

In server mode, `bd federation status` reported:

```text
ReachError: failed to get peer credentials: not found: federation peer origin
```

for ordinary credential-free Dolt remotes, while `bd dolt pull`/`push` succeeded against the same remote. `withPeerCredentials` treated `GetFederationPeer` returning `ErrNotFound` as fatal, but `runFederationAddPeer` only calls `AddRemote()` when `--user` is omitted — so credential-free bootstrap (`add-peer` without `--user`, or `sync.remote` bootstrap) never creates a `federation_peers` row in the first place.

A missing peer row now maps to empty credentials (`fn(nil)`); decryption and database errors stay fatal.

Thanks to RaychevAngel for the clear repro and root-cause trace in the issue.

## Test plan

- [x] `go test ./internal/storage/dolt/ -count=1 -timeout 25m` (icu4c CGO flags) — pass, verified on a probe branch merged with current main (`3830f0a34`)


maphew (commented):
The root-cause trace here is correct, and I verified each step against the tree at `3830f0a34`:

- `runFederationAddPeer` (cmd/bd/federation.go:412-422) writes a `federation_peers` row only when `--user` is supplied; the `else` branch calls `store.AddRemote` alone. So a credential-free `add-peer`, and `sync.remote` bootstrap, genuinely leave no peer row.
- `GetFederationPeer` (internal/storage/dolt/credentials.go:328) turns `sql.ErrNoRows` into a wrapped `storage.ErrNotFound`, and `withPeerCredentials` (credentials.go:572-575 pre-patch) escalated that to a hard failure, which is exactly the `ReachError: failed to get peer credentials: not found: federation peer origin` in the issue. `runFederationStatus` enumerates peers from `ListRemotes` (cmd/bd/federation.go:279-281) and calls `Fetch` on each, so the failure is reached by every credential-free remote.

The fix itself checks out:

- `fn(nil)` is safe. `remoteCredentials.empty()` is explicitly nil-receiver-tolerant (credentials.go:436-438), and `withPeerCredentials` already passed `nil` for a peer row with blank username/password, so the nil path was pre-existing and exercised — this change just adds one more way to reach it.
- The best-effort `updatePeerLastSync` is already gated on `peer != nil`, so the missing-row path correctly skips it.
- The narrowing is genuinely narrow: `storage.ErrNotFound` is `errors.New("not found")` (internal/storage/storage.go:52) and is produced in exactly one branch of `GetFederationPeer`. Decryption failures (`failed to decrypt password`), credential-key init failures, and a missing/unmigrated `federation_peers` table all stay fatal, as claimed.
- I applied the diff on top of `3830f0a34` and `go test -tags gms_pure_go -run TestIsMissingFederationPeer ./internal/storage/dolt/` passes.

Two things before this lands.

**1. The test does not test the fix.** `internal/storage/dolt/credentials_peer_test.go` only exercises `isMissingFederationPeer`, which is a one-line wrapper around `errors.Is` — the table is effectively asserting that `errors.Is` unwraps `%w`. The behavior that changed is `withPeerCredentials` returning `fn(nil)` instead of an error, and nothing covers it: the test would still pass if the new branch in `withPeerCredentials` were deleted. Issue #4837 asked for coverage of the missing-row case, the stored-credential case, and malformed/decryption errors staying fatal; none of the three is covered.

A behavioral test is cheap here — `credentials_test.go:569` (`TestFederationPeerCredentialLifecycleLazyKeyInit`) already has the pattern: `skipIfNoServer(t)`, build a store with `New(...)`, then call `store.withPeerCredentials(ctx, "no-such-peer", fn)` and assert `fn` was invoked once with `nil` creds and no error returned. A second case adding a peer with credentials and asserting `fn` receives them keeps the non-regression side honest. With that in place I'd inline the `errors.Is` at the call site and drop `isMissingFederationPeer` entirely — it is indirection that exists only to be testable.

**2. Title.** `# fix(federation): treat missing peer row as credential-free remote` has a stray leading `# `, which would end up in the squash-merge subject. Please drop it.

One observation, not a blocker: `withPeerCredentials` now no longer validates that the named peer exists at all. `bd federation status` and `bd federation sync` without `--peer` enumerate from `ListRemotes`, so they are unaffected, but `--peer typo` previously failed with `not found: federation peer typo` and will now fall through to `CALL DOLT_FETCH` and surface Dolt's unknown-remote error instead. That is a small message regression, and #4837's own suggested fix scoped the leniency to "when the named Dolt remote exists". A `ListRemotes` membership check would restore the clear message, at the cost of one extra query per call; I don't think it blocks merging, but a sentence in the doc comment noting that peer-name validation now belongs to the caller would be worth having.

Once the behavioral test is in, this is good to merge. If you'd rather not build out the server-mode test, say so and I can push it onto your branch — your call.

vishnujayvel:
Thanks — the finding that the test doesn't test the fix is the one that matters: `credentials_peer_test.go` only exercises the `errors.Is` wrapper, and would still pass if the new `withPeerCredentials` branch were deleted. That's the gap #4837 actually asked to be covered.

No false claims to correct in the body itself.

On the title: our publishing script derived the PR title with `sed 's/^#\+ *//'`, and BSD sed doesn't support `\+`, so the leading `# ` was never stripped here (and on eight other PRs in this batch). Thanks for hand-editing it. The script now asserts the stripped title has no leading `#` before opening the PR, so this can't recur.

Please push the behavioral test — `skipIfNoServer`, a store built with `New(...)`, asserting `withPeerCredentials(ctx, "no-such-peer", fn)` invokes `fn` with nil creds and no error, plus the credentialed-peer case — as a maintainer commit, and inline `errors.Is` at the call site once it's in, dropping `isMissingFederationPeer`. We'll leave the `--peer typo` message regression as a documented tradeoff rather than add the extra query.

bee-ghosttrack:
The red **Server Dolt Full Suite 16/16** on `08e43561` is caused by this diff, not a flake — naming it so the next push can fix it in one go.

`TestFilteredPushUsesDedicatedLongTimeoutConnection` (`internal/storage/dolt/federation_test.go:643`, landed on main in #5342 on Aug 6, after this PR was opened) pushes to `"nonexistent-peer"` and asserts the error contains `failed to get peer credentials`. With `withPeerCredentials` now treating the missing `federation_peers` row as credential-free and calling `fn(nil)`, the push proceeds to Dolt and fails there instead:

```
federation_test.go:643: filtered push failed before peer lookup: failed to push to peer nonexistent-peer: Error 1105 (HY000): fatal: remote 'nonexistent-peer' not found.
```

That is the intended new contract (a name with no peer row is a plain Dolt remote, and Dolt says it does not exist), so the fix is on the test side: assert the Dolt `remote 'nonexistent-peer' not found` error rather than the credential-lookup one. The pool-wait observation the test actually exists for is unaffected — `GetFederationPeer` still runs on the shared pool after staging, so `waitForPoolWait` still fires — and `assertFederationStagingBranchAbsent` should still hold on the error path; please keep both assertions.

The rest of the earlier review thread (the unit test only covering the `errors.Is` predicate, not the fetch/pull path) still stands as the substantive ask.

vishnujayvel:
Thanks for pinning the red suite to `TestFilteredPushUsesDedicatedLongTimeoutConnection`. Agreed that it is this PR's contract: a name with no `federation_peers` row is a plain Dolt remote, so Dolt is the one that reports it missing.

Pushed 1b1b28048 with two changes:

1. `internal/storage/dolt/federation_test.go:640` — the assertion now expects the Dolt `remote 'nonexistent-peer' not found` error instead of `failed to get peer credentials`. `waitForPoolWait` and `assertFederationStagingBranchAbsent` are unchanged.
2. `internal/storage/dolt/credentials_peer_test.go` — added `TestWithPeerCredentialsMissingPeerRow` (server mode, `skipIfNoServer`, store built with `New(...)`), following `TestFederationPeerCredentialLifecycleLazyKeyInit`. Subtests: a name with no row invokes `fn` exactly once with `nil` creds and returns nil; an `fn` error on that path is propagated; a peer stored with username/password delivers those creds to `fn`. Deleting the new branch in `withPeerCredentials` fails the first subtest.

Verification on my side, honestly stated: this machine has no reachable Docker daemon, so the server-mode tests skip locally and I could not execute the two changed tests here. What I ran:

```
gofmt -l internal/storage/dolt/   -> (empty)
CGO_ENABLED=0 go vet -tags gms_pure_go ./internal/storage/dolt/   -> ok
BEADS_TEST_SKIP=dolt CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/dolt/ -run 'Federation|PeerCredentials' -count=1 -v
--- SKIP: TestWithPeerCredentialsMissingPeerRow (0.00s)
--- PASS: TestIsMissingFederationPeer (0.00s)
ok  	github.com/steveyegge/beads/internal/storage/dolt	0.522s
BEADS_TEST_SKIP=dolt CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/dolt/ -count=1
ok  	github.com/steveyegge/beads/internal/storage/dolt	9.072s
```

So the Server Dolt suite on this push is the real check for both tests; I will watch it.

I left `isMissingFederationPeer` and its predicate test in place for now rather than inlining `errors.Is` at the call site in the same push, since that also means removing `TestIsMissingFederationPeer`. Happy to do that in a follow-up commit if you still prefer it inlined, or you are welcome to drop it when merging.

bee-ghosttrack:
**Verdict: APPROVE** (bee, maintainer seat, re-review of `1b1b28048`).

Re-verified on this machine against a real Dolt container (`dolthub/dolt-sql-server`, the suite's own `TestMain`), since your run could only skip:

```
go test ./internal/storage/dolt/ -run 'TestWithPeerCredentialsMissingPeerRow|TestIsMissingFederationPeer' -count=1
--- PASS: TestWithPeerCredentialsMissingPeerRow (57.29s)   # all three subtests
--- PASS: TestIsMissingFederationPeer (0.00s)
go test -tags integration ./internal/storage/dolt/ -run TestFilteredPushUsesDedicatedLongTimeoutConnection -count=1
--- PASS: TestFilteredPushUsesDedicatedLongTimeoutConnection (31.80s)
```

Note the federation test is behind `//go:build integration`; without the tag it silently doesn't compile in, so a plain `go test` run would not have exercised your assertion change. It does with the tag, and the new Dolt `remote 'nonexistent-peer' not found` assertion holds on the error path with `waitForPoolWait` and `assertFederationStagingBranchAbsent` intact.

Mutation check, so the new test is proven to test the fix: with the `isMissingFederationPeer(err)` branch in `withPeerCredentials` (`credentials.go:623`) disabled, the first two subtests fail as they should (`want nil` / `want remote rejected`, both getting `failed to get peer credentials: not found: federation peer no-such-peer`); the stored-credentials subtest still passes. That closes the gap #4837 asked to be covered.

`go vet ./internal/storage/dolt/` clean. Branch is one commit behind main and mergeable.

One leftover from your own July note: `isMissingFederationPeer` is still a named helper with its own table test rather than an inlined `errors.Is` at the call site. Fine either way for me; not a blocker.

Merge is the maintainer gate's call.

PR #5129, opened 2026-07-29
fix(ado): map parent-child deps to hierarchy links (not Related)

### LINKED ISSUE #4961: ADO: parent-child dependencies pushed as Related links, System.Parent never set
## Summary

The ADO adapter maps beads dependency types using the string `"parent"`, but beads stores the hierarchy type as `"parent-child"` (`types.DepParentChild`). As a result every parent-child dependency is pushed to Azure DevOps as a **Related** link instead of a hierarchy link, and `System.Parent` is never set. The pull direction is broken symmetrically.

Confirmed against v1.1.0 and current `main` (identical), with a live round-trip against a real ADO project (Scrum process).

## Reproduce

```bash
bd create "parent epic" -t epic          # -> bd-aaa
bd create "child task"                   # -> bd-bbb
bd link bd-bbb bd-aaa --type parent-child
bd ado sync --push-only --parent bd-aaa
```

Reports `✓ Synced 1 dependency links`. In ADO the child has:

```
relations: [("System.LinkTypes.Related", <parent id>)]
System.Parent: null
```

Expected `System.LinkTypes.Hierarchy-Reverse` / `System.Parent` set.

## Root cause

Push path passes the storage vocabulary through unchanged — `cmd/bd/ado.go:826`:

```go
desired = append(desired, tracker.DependencyInfo{
    FromExternalID: extIDStr,
    ToExternalID:   targetExtID,
    Type:           string(dep.DependencyType),   // "parent-child"
})
```

`internal/types/types.go:857`: `DepParentChild DependencyType = "parent-child"`. There is no `"parent"` constant in `internal/types`.

That value reaches `internal/ado/links.go:91`:

```go
func beadsDepToADORel(depType string) string {
	switch depType {
	case "blocks":          return RelDependsOn
	case "parent":          return RelChild        // unreachable
	case "related":         return RelRelated
	case "discovered-from": return RelRelated
	default:                return RelRelated      // "parent-child" lands here
	}
}
```

`case "parent"` cannot match any value beads produces, so `RelChild` is never emitted.

Pull is broken the same way: `adoRelToBeadsDep` (links.go:60) returns `"parent"` for `RelChild`/`RelParent`, and `internal/tracker/engine.go:1218` casts it unchecked — `Type: types.DependencyType(dep.Type)` — writing a dep type no other beads code recognises.

The two combine into a warning on every subsequent sync:

```
Failed to create dependency X -> Y: dependency X -> Y already exists
with type "parent-child" (requested "related")
```

ADO appears to be the only adapter with this issue — Linear (`internal/linear/mapping.go:765`), GitHub (`internal/github/types.go:197`) and GitLab (`internal/gitlab/types.go:230`) all use `"parent-child"`.

Tests don't catch it because they assert the adapter against itself with literals production never produces — `internal/ado/links_test.go:253` `{"parent", RelChild}`, and line 388 `Type: "parent"`.

## Suggested fix

In `internal/ado/links.go`, use `"parent-child"` in both directions:

```go
// beadsDepToADORel
case "parent-child": return RelChild

// adoRelToBeadsDep
case RelChild:  return "parent-child", false
case RelParent: return "parent-child", true
```

Note the direction: beads stores the dep child -> parent, while `RelChild` (`Hierarchy-Forward`) means "this item is the parent of the target", so the push side needs the same swap `adoRelToBeadsDep` already performs on pull — otherwise the hierarchy is created inverted. The two test literals need updating alongside.

`tracks` also falls through to `default` and becomes a Related link, which may or may not be intended.

## Environment

- bd 1.1.0 (Homebrew), macOS
- Azure DevOps Services, Scrum process template, Epic -> Feature -> Product Backlog Item -> Task

vishnujayvel:
## Summary

Fixes #4961.

The ADO adapter mapped beads dependency types using the string `"parent"`, but beads stores the hierarchy type as `"parent-child"` (`types.DepParentChild`) — `"parent"` is not a constant that exists anywhere in `internal/types`. As a result, every parent-child dependency was pushed to Azure DevOps as a **Related** link instead of a hierarchy link, and `System.Parent` was never set. The pull direction was broken the same way, writing a `"parent"` dependency type back into beads that no other code recognizes.

`beadsDepToADORel` and `adoRelToBeadsDep` (`internal/ado/links.go`) now match on `"parent-child"` in both directions **and emit the correct hierarchy direction**.

*(Corrected 2026-07-29 — an earlier version of this paragraph said the change "preserves the direction swap the pull side already did." That was wrong on both halves, as @maphew established in review: the push path applies no swap at all (it builds the desired set from storage and simply used the wrong relation constant), and the pull side's swap flags were themselves inverted. Both are fixed in `e00fcfea3`; the original claim was ours and it was incorrect.)*

Because the desired set is built from the child's own outgoing row (`IssueID`=child, `DependsOnID`=parent), `PushLinks` always runs with `workItemID` as the child and the target as the parent, so the push must emit `RelParent` (Hierarchy-Reverse). On pull, the swap flags are flipped so both branches yield beads' child → parent storage direction.

Thanks to Ruben Fonseca for the detailed root-cause trace and live ADO round-trip repro in the issue.

## Test plan

- [x] `go test ./internal/ado/ -count=1 -timeout 25m` (icu4c CGO flags) — pass, verified on a probe branch merged with current main (`3830f0a34`)

### DIFF
--- a/internal/ado/links.go
+++ b/internal/ado/links.go
@@ -51,6 +51,8 @@ const discoveredFromComment = "beads:discovered-from"
 // adoRelToBeadsDep maps an ADO relation type to a beads dependency type.
 // Returns the dep type and whether the from/to should be swapped
 // (true for reverse link types that need direction normalization).
+// Hierarchy links use beads storage vocabulary "parent-child" (types.DepParentChild),
+// not the obsolete "parent" alias (GH#4961).
 func adoRelToBeadsDep(rel string, attributes map[string]interface{}) (depType string, swap bool) {
 	switch rel {
 	case RelDependsOn: // Dependency-Forward: this item is predecessor (blocks target)
@@ -58,9 +60,9 @@ func adoRelToBeadsDep(rel string, attributes map[string]interface{}) (depType st
 	case RelDependencyOf: // Dependency-Reverse: target blocks this item → swap
 		return "blocks", true
 	case RelChild: // Hierarchy-Forward: this item is parent of target
-		return "parent", false
+		return "parent-child", false
 	case RelParent: // Hierarchy-Reverse: target is parent of this item → swap
-		return "parent", true
+		return "parent-child", true
 	case RelRelated:
 		if hasDiscoveredFromAttribute(attributes) {
 			return "discovered-from", false
@@ -89,11 +91,13 @@ func hasDiscoveredFromAttribute(attributes map[string]interface{}) bool {
 }
 
 // beadsDepToADORel maps a beads dependency type to an ADO relation type.
+// Accepts both "parent-child" (canonical storage type) and the obsolete
+// "parent" alias so older in-memory values still map to hierarchy (GH#4961).
 func beadsDepToADORel(depType string) string {
 	switch depType {
 	case "blocks":
 		return RelDependsOn
-	case "parent":
+	case "parent-child", "parent": // "parent" kept as alias only
 		return RelChild
 	case "related":
 		return RelRelated
--- a/internal/ado/links.go
+++ b/internal/ado/links.go
@@ -52,7 +52,8 @@ const discoveredFromComment = "beads:discovered-from"
 // Returns the dep type and whether the from/to should be swapped
 // (true for reverse link types that need direction normalization).
 // Hierarchy links use beads storage vocabulary "parent-child" (types.DepParentChild),
-// not the obsolete "parent" alias (GH#4961).
+// not the legacy "parent" type, which was only ever produced by the pre-fix
+// pull path and is stored backwards (GH#4961, PR #5129).
 //
 // beads stores parent-child deps child → parent: IssueID is the child,
 // DependsOnID is the parent (see types.go ready-item parent computation).
@@ -99,7 +100,7 @@ func hasDiscoveredFromAttribute(attributes map[string]interface{}) bool {
 
 // beadsDepToADORel maps a beads dependency type to an ADO relation type.
 // Accepts both "parent-child" (canonical storage type) and the obsolete
-// "parent" alias so older in-memory values still map to hierarchy (GH#4961).
+// The legacy "parent" type is intentionally not mapped here (GH#4961, PR #5129).
 //
 // A parent-child dep's desired set is built from the child's own outgoing
 // dependency row (IssueID=child, DependsOnID=parent — see
@@ -111,7 +112,14 @@ func beadsDepToADORel(depType string) string {
 	switch depType {
 	case "blocks":
 		return RelDependsOn
-	case "parent-child", "parent": // "parent" kept as alias only
+	// The "parent" alias is deliberately NOT accepted here. Rows of that type were
+	// only ever written by the pre-fix pull path, and both old branches stored them
+	// parent -> child — backwards relative to the convention this file now uses. Mapping
+	// them to RelParent would assert the child is the parent's parent and reparent real
+	// hierarchies (maphew, PR #5129). They fall through to RelRelated, i.e. inert, which
+	// is where they have effectively been all along. Retyping and reversing those legacy
+	// rows needs a migration with its own tests, tracked separately.
+	case "parent-child":
 		return RelParent
 	case "related":
 		return RelRelated
--- a/internal/ado/links.go
+++ b/internal/ado/links.go
@@ -53,16 +53,23 @@ const discoveredFromComment = "beads:discovered-from"
 // (true for reverse link types that need direction normalization).
 // Hierarchy links use beads storage vocabulary "parent-child" (types.DepParentChild),
 // not the obsolete "parent" alias (GH#4961).
+//
+// beads stores parent-child deps child → parent: IssueID is the child,
+// DependsOnID is the parent (see types.go ready-item parent computation).
+// RelChild (Hierarchy-Forward) means "this item is the parent of target," so
+// the resulting dep must point from target (child) to this item (parent) —
+// swap. RelParent (Hierarchy-Reverse) means "target is the parent of this
+// item," so this item is already the child — no swap.
 func adoRelToBeadsDep(rel string, attributes map[string]interface{}) (depType string, swap bool) {
 	switch rel {
 	case RelDependsOn: // Dependency-Forward: this item is predecessor (blocks target)
 		return "blocks", false
 	case RelDependencyOf: // Dependency-Reverse: target blocks this item → swap
 		return "blocks", true
-	case RelChild: // Hierarchy-Forward: this item is parent of target
-		return "parent-child", false
-	case RelParent: // Hierarchy-Reverse: target is parent of this item → swap
+	case RelChild: // Hierarchy-Forward: this item is parent of target → swap so dep is child(target) → parent(this)
 		return "parent-child", true
+	case RelParent: // Hierarchy-Reverse: target is parent of this item; this item is already the child
+		return "parent-child", false
 	case RelRelated:
 		if hasDiscoveredFromAttribute(attributes) {
 			return "discovered-from", false
@@ -93,12 +100,19 @@ func hasDiscoveredFromAttribute(attributes map[string]interface{}) bool {
 // beadsDepToADORel maps a beads dependency type to an ADO relation type.
 // Accepts both "parent-child" (canonical storage type) and the obsolete
 // "parent" alias so older in-memory values still map to hierarchy (GH#4961).
+//
+// A parent-child dep's desired set is built from the child's own outgoing
+// dependency row (IssueID=child, DependsOnID=parent — see
+// GetDependenciesWithMetadata), so PushLinks always runs with workItemID as
+// the child and the target as the parent. RelParent (Hierarchy-Reverse) means
+// "target is parent of this item," which is exactly that relationship;
+// RelChild would instead assert the child is the parent's parent.
 func beadsDepToADORel(depType string) string {
 	switch depType {
 	case "blocks":
 		return RelDependsOn
 	case "parent-child", "parent": // "parent" kept as alias only
-		return RelChild
+		return RelParent
 	case "related":
 		return RelRelated
 	case "discovered-from":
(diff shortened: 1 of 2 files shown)


maphew (commented):
You found the real defect and located it precisely: `beadsDepToADORel`'s `case "parent"` is dead code, because beads only ever produces `types.DepParentChild = "parent-child"` (`internal/types/types.go:920`), so every hierarchy dep falls through to `default` and ships as a Related link. That diagnosis holds up against the code, and the ADO adapter is genuinely the odd one out — Linear, GitHub and GitLab all use `"parent-child"`.

The reason I can't take this as-is is the second half of #4961, the direction. Renaming the string without flipping the direction changes the failure mode from "inert" to "actively wrong", in both directions.

**Push (`internal/ado/links.go:99`)**

`cmd/bd/ado.go:826` builds the desired set from `GetDependenciesWithMetadata`, which returns only outgoing rows (`internal/storage/issueops/dependencies.go:1067`, `WHERE issue_id = ?`). For parent-child beads stores IssueID=child, DependsOnID=parent — see `internal/types/types.go:1379`, which computes `Parent = dep.DependsOnID`, and `internal/linear/mapping.go:761-766`, which emits From=child/To=parent for the same relationship. So `PushLinks` runs with `workItemID` = the child and `ToExternalID` = the parent, and at links.go:287-297 calls:

```go
r.Client.AddWorkItemLink(ctx, workItemID /* child */, targetURL /* parent */, RelChild, comment)
```

`RelChild = System.LinkTypes.Hierarchy-Forward` (`internal/ado/types.go:89`), which means *the target is the child*. The link that gets created says the Epic is a child of the Task — `System.Parent` ends up set on the wrong work item, and on a real backlog that reparents items. The issue's Expected result is `Hierarchy-Reverse` / `System.Parent` on the child, which is `RelParent`, not `RelChild`.

Suggested: `case "parent-child": return RelParent`. That change also needs `removableRelTypes` (links.go:190-194) revisited — it currently marks `RelChild` removable and deliberately excludes `RelParent` on the theory that beads never emits reverse links. Once beads owns the Hierarchy-Reverse link on the child, that ownership set has to flip for the hierarchy pair or stale hierarchy links will never be cleaned up.

**Pull (`internal/ado/links.go:63-65`)**

The swap flags are inverted relative to beads storage, which is why renaming the type here is not safe either. `ExtractLinkDeps` (links.go:140-149) sets From=source/To=target when `swap == false`, and the engine writes `IssueID=From, DependsOnID=To` (`internal/tracker/engine.go:1215-1219`). So work item A with `Hierarchy-Forward → B` (A is B's parent) currently produces IssueID=A, DependsOnID=B — beads records the *child* as A's parent. Today that row is typed `"parent"`, which no beads query recognizes, so it is garbage but harmless. After this patch it is a valid `parent-child` edge pointing backwards, and it feeds parent computation, blocked-state recompute (`internal/storage/issueops/dependencies.go:310-318`) and cycle detection (`internal/storage/issueops/cycles.go:111`).

Suggested: swap both flags — `RelChild → ("parent-child", true)`, `RelParent → ("parent-child", false)`.

**Tests**

`go test ./internal/ado/` passes on the branch (I ran it: `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/ado/ -count=1`, 4.3s), but it passes for the same reason the original bug survived. `links_test.go:388` renames the literal while keeping `From: "100", To: "300"` — parent→child, the direction production never generates — and the assertions at links_test.go:402-414 only check `op == "add"` and `path == "/relations/-"`; the emitted `rel` is never asserted. Two tests would pin this down and would fail on the current branch:

1. A push test whose `desired` uses the storage direction (From = the work item being pushed = the child, To = the parent), asserting the PATCH body carries `System.LinkTypes.Hierarchy-Reverse`.
2. A pull test on a work item carrying `RelChild`, asserting the resulting dep is From = the child, To = the source item.

**Smaller things**

- The PR title starts with a literal `# `, which would land in the squash commit subject.
- links.go:94-95 calls `"parent"` an "obsolete alias". It was never a valid beads type — the only source of such rows is the pre-fix pull path. Keeping the alias is fine, but the comment should say that, and existing `"parent"` rows in users' DBs stay invisible to every parent-aware query unless something normalizes them; worth a follow-up either way.
- The description says the change preserves "the direction swap the pull side already did" — the push side applies no swap, and the pull side's swap is the part that is wrong. Worth correcting so a later reader does not take it at face value.
- Branch base is `2e20a183f`, behind main (`3830f0a34`); still mergeable, just needs an update-branch before landing.
- Leaving `tracks` out of scope is the right call.

Thanks for picking this up, and for crediting the root-cause trace in the issue — that part of the report is exactly right, and the string mismatch half of the fix is correct. This is worth landing once the direction is handled. Want to take the two changes above plus the two tests, or would you rather I push them onto the branch?

vishnujayvel:
Thanks — B1 is a genuine bug and I verified the whole chain rather than taking it on faith. `cmd/bd/ado.go:812` calls `GetDependenciesWithMetadata(issue.ID)`, which returns outgoing rows, and `types.go:1298-1299` shows parent-child stored as `IssueID=child` / `DependsOnID=parent`. So `PushLinks` always runs with `workItemID` as the child and the target as the parent, and emitting `RelChild` (Hierarchy-Forward) from the child asserted the child is the parent's parent. Reparenting backwards, exactly as you described.

Correction to our description: it claims the fix preserves pull's existing swap. That's wrong on both halves — push applies no swap at all (it built the desired set from storage and used the wrong relation constant), and pull's swap flags were the part that was inverted.

Pushed in `e00fcfea3`:

- `beadsDepToADORel` now emits `RelParent` (Hierarchy-Reverse) for `parent-child`/`parent`.
- `adoRelToBeadsDep`: swap flags flipped for both `RelChild` and `RelParent`, so the resulting dep always points child → parent.
- `removableRelTypes` tracks `RelParent`, since beads now owns Hierarchy-Reverse on the child — the ownership flip you flagged as a consequence of B1, handled here rather than deferred.
- B3: added `TestPushLinks_HierarchyDirection` and `TestPullLinks_HierarchyChildDirection` to pin the emitted and parsed relation type in each direction. Five existing tests had the old backwards direction baked into their fixtures and assertions; those are corrected too, which is a fair sign the tests were confirming the bug rather than catching it.

`go test ./internal/ado/... -count=1` → `ok github.com/steveyegge/beads/internal/ado 4.459s`. `go vet` clean.

Took this one ourselves rather than handing it over — it was our bug.

maphew (commented):
Thanks for taking this one yourself, and for verifying the chain rather than accepting my trace on faith — that's the right instinct, and your walk through `cmd/bd/ado.go` → `GetDependenciesWithMetadata` → `types.go` reaches the same place I did. `e00fcfea3` fixes the direction problem properly.

**Resolved**

- **Push direction.** `beadsDepToADORel` now returns `RelParent` for `parent-child`. Since the desired set is built from the child's own outgoing row, `PushLinks` runs with `workItemID` = child and target = parent, and Hierarchy-Reverse on the child is exactly "the target is my parent". `System.Parent` now lands on the right work item.
- **Pull direction.** The swap flags are flipped the right way: `RelChild` → swap (the target is the child, so the dep points target → this item), `RelParent` → no swap. Both branches now produce beads' child → parent storage direction, which is what `types.go:1379` reads for parent computation and what `internal/linear/mapping.go:760-766` emits for the same relationship.
- **Ownership set.** Flipping `removableRelTypes` from `RelChild` to `RelParent` is correct and does more than I asked for. Under the old set, `RelChild` was removable, so pushing a *parent* work item would see the auto-mirrored Hierarchy-Forward link to its child, find it absent from the desired set (the parent has no outgoing hierarchy row), and delete it — quietly destroying a real hierarchy. Your revised `TestPushLinks_PreservesUnmanagedLinks` fixture pins that. Good catch that wasn't in my list.
- **Tests.** These are real direction tests, not assertion tweaks. I checked out `e00fcfea3`, reverted `internal/ado/links.go` to `822b249d2`, and re-ran: `TestPushLinks_HierarchyDirection`, `TestPullLinks_HierarchyChildDirection`, `TestPullLinks_DirectRelations`, `TestPullLinks_ReverseDirectionNormalization`, `TestAdoRelToBeadsDep`, `TestBeadsDepToADORel`, `TestPushLinks_RemoveStale`, `TestPushLinks_PreservesUnmanagedLinks` and `TestPushLinks_RemoveFailure` all fail. Restored, `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/ado/... -count=1` is green in 4.0s and `go vet` is clean. Correcting the five fixtures that had the old direction baked in was the right call rather than leaving them.

Placement is right too — hierarchy mapping is confined to `links.go`, and there's no second site setting `System.Parent` that could drift out of sync with it.

**One thing left, and it's my earlier advice that's now wrong**

In my first pass I said keeping the `"parent"` alias was fine. That was true when the canonical mapping was `RelChild`; it stopped being true the moment the direction flipped, and I should have flagged the consequence.

`internal/ado/links.go:114-115`:

```go
case "parent-child", "parent": // "parent" kept as alias only
	return RelParent
```

The only thing that ever wrote a `"parent"`-typed row is the pre-fix pull path, and `DependencyType.IsValid()` (`internal/types/types.go:969`) accepts any non-empty string under 50 chars, so those rows really are sitting in the DB of anyone who ran an ADO pull. Both old pull branches produced the same shape: `RelChild` gave From=source(parent)/To=target(child), and `RelParent` with the old `swap: true` gave From=target(parent)/To=source(child). So every legacy `"parent"` row is stored **parent → child** — backwards relative to the convention this PR just standardised on.

Concrete failure: a user synced an Epic 300 and Task 100 before this fix, and has a stale row `IssueID=300 (epic), DependsOnID=100 (task), type="parent"`. On the next `bd ado push`, `PushLinks` runs with `workItemID = 300` and a desired entry pointing at 100, the alias maps it to `RelParent`, and the PATCH asserts *the Task is the parent of the Epic*. That reparents the epic under the task — the exact corruption this PR is fixing, arriving through the legacy path. Before this PR the alias mapped to `RelChild`, which happened to be right for those backwards rows; after it, the alias is the one place that's newly wrong.

Simplest fix is to delete the alias:

```go
case "parent-child":
	return RelParent
```

Legacy rows then fall through to `default` → `RelRelated`, i.e. back to inert, which is where they've been all along and is strictly better than an active mis-parent. It also deletes code rather than adding it. A separate follow-up issue can handle the real cleanup, since a migration has to both retype `"parent"` → `"parent-child"` *and* reverse `IssueID`/`DependsOnID` — worth its own change with its own tests, definitely not this PR's job. Happy to open that issue, or push the one-line change onto your branch if you'd rather not spin another round.

**Two small things**

- The PR body still carries the original claim that the change preserves "the direction swap the pull side already did". Your comment corrects it, but the body is what a squash merge and a future `git log` reader see. Worth editing in place.
- `links.go:55` and `:101` still describe `"parent"` as an "obsolete alias". It was never a valid beads type — it only ever existed as output of the bug. If the alias survives, the comment should say that; if it goes, both comments can just drop the clause.

Base is 135 commits behind main, but nothing has touched `internal/ado` or `cmd/bd/ado.go` in that range, so an update-branch before landing is mechanical.

**Disposition: approve with one should-fix.** The core of this is right and the tests hold it in place. Resolve the `"parent"` alias — either drop it or, if you'd rather keep it, map it to `RelChild` with a comment explaining that legacy rows are stored backwards — and I'll land it. Everything else on my earlier list is either done or a follow-up.

vishnujayvel:
Thanks — took the should-fix, and you were right to flag your own earlier advice as now-wrong.

Pushed in `5a04ac174`: the `"parent"` alias is gone from `beadsDepToADORel`, which now matches only `"parent-child"`.

The reason it mattered rather than being cosmetic: `DependencyType.IsValid()` is `len(d) > 0 && len(d) <= 50`, so arbitrary strings validate and legacy `"parent"` rows really do persist in existing databases — this wasn't a dead branch. Those rows were only ever written by the pre-fix pull path, and both old branches stored them parent → child, i.e. backwards relative to the convention `links.go` now uses. Mapping them to `RelParent` would have asserted the child is the parent's parent and reparented real hierarchies — the exact bug B1 was about, reintroduced through the compatibility door. They now fall through to `RelRelated`, which is inert, and which is effectively where they have been all along. Retyping and reversing them is a migration with its own tests; not this PR.

Test updated to pin the behaviour rather than leave it implicit: `{"parent", RelRelated}`, with a comment explaining why. `go test ./internal/ado/... -count=1` → `ok github.com/steveyegge/beads/internal/ado 4.408s`, gofmt clean.

On `removableRelTypes` doing more than you asked — fair. It's in scope in the sense that leaving it tracking `RelChild` would have left ownership pointing at the relation we no longer emit, but I should have called it out as an extra rather than folding it in silently. Happy to split it out if you'd rather review it separately.

Also edited the PR body in place — you're right that the squash message is what survives. The paragraph claiming the fix "preserves the direction swap the pull side already did" is replaced with the correct description plus a dated note saying the original claim was ours and was wrong on both halves.

PR #5131, opened 2026-07-29
fix(doctor): detect parent→child hierarchy-blocking deps

### LINKED ISSUE #4814: checkChildParentDependenciesDB / --fix-child-parent only detects one direction of the parent-child+blocks deadlock
## Summary

`bd doctor`'s `checkChildParentDependenciesDB` (`cmd/bd/doctor/validation.go`) and its
`--fix-child-parent` remediation (`cmd/bd/doctor/fix/validation.go`,
`ChildParentDependencies`) detect and fix this SQL shape:

```
d.issue_id LIKE CONCAT(d.depends_on_id, '.%') AND d.type IN ('blocks', 'conditional-blocks', 'waits-for')
```

i.e. **child depends on parent** (`fromID=child, toID=parent`). This is the same direction
`bd dep add`'s own write-time guard already rejects (the "already a child of ... would
create a deadlock" error).

The **mirror-image shape -- parent depends on child** (`fromID=parent, toID=child`) -- is
not detected by either `bd doctor` or `bd dep add`'s pre-check, and not detected by `bd dep
cycles` either (confirmed: `DetectCycles`'s `cycleRelevantDepType` deliberately excludes
`DepParentChild`, so a `blocks`-based cycle routed through a `parent-child` edge is invisible
to it -- we read this as an intentional design tradeoff, not an oversight, since rejecting
every `parent-child`-adjacent `blocks` edge would over-reject a lot of legitimate graphs).
But `GetReadyWork` *does* treat `parent-child` as readiness-affecting (via
`DependencyType.AffectsReadyWork()`), so this shape fully, silently deadlocks both beads at
the ready-work layer while remaining completely invisible to every detection/prevention
mechanism bd currently ships.

## Isolated repro (3 commands, no proprietary data)

```
bd create fake-task-B --type=task
bd create fake-task-B1 --parent fake-task-B
bd dep add fake-task-B fake-task-B1
```

The third command **succeeds** with no rejection. After it:

```
bd list --parent fake-task-B --ready   # -> [] (both beads deadlocked)
bd blocked                              # -> correctly shows both mutually blocked
bd dep cycles                           # -> "No dependency cycles detected"
```

So the deadlock is real and immediately in effect, but the only tool that reports it
correctly (`bd blocked`) doesn't explain *why*, and the two tools that could plausibly
diagnose or prevent it (`bd dep cycles`, `bd dep add`'s pre-check) both say everything is
fine.

## Primary ask

Extend `checkChildParentDependenciesDB` / `ChildParentDependencies` to also detect the
mirror-image predicate:

```
d.depends_on_id LIKE CONCAT(d.issue_id, '.%') AND d.type IN ('blocks', 'conditional-blocks', 'waits-for')
```

You already ship the fix for one direction of this anti-pattern; the other direction is
just as real and currently invisible to both `bd dep add`'s pre-check and `bd dep cycles`.

## Secondary ask (framed as a question, not a bug report)

Should `bd dep add`'s own pre-check (`isChildOf` in `cmd/bd/dep.go`) also cover the
parent-depends-on-child direction at write time, the same way it already blocks
child-depends-on-parent? That would stop the bad edge from ever being written rather than
requiring a separate `bd doctor` pass to find it after the fact -- strictly stronger than
the primary ask, but a bigger behavior change (`AddDependencyInTx` currently only checks
cross-type compatibility plus the one `isChildOf` direction), so it's reasonable to prefer
the doctor-based approach as the safer opt-in default if a write-time guard feels too
strict.

## Two smaller, separately-worth-reporting issues found during the same investigation

1. **`bd list --parent a,b` (comma-separated multi-id) silently returns `[]`** instead of
   erroring or unioning. Confirmed isolated: two disposable single-bead roots each return 1
   row via `--parent <id>` alone, 0 rows via `--parent <id1>,<id2>`. Since `--parent` is
   documented/typed as a single string, this suggests comma-splitting was never
   implemented, not deliberately rejected -- worth either rejecting the malformed input
   loudly or implementing union semantics.
2. **`bd epic status <non-epic-id>` silently ignores its argument** and lists unrelated
   epics instead of erroring when given an id whose `issue_type` isn't `epic`. Confirmed
   against a disposable non-epic bead mixed in with real epics -- not an artifact of any
   specific dataset.

## Context note (not a fix request)

`bd doctor` itself is unavailable in embedded-Dolt mode ("not yet supported in embedded
mode," confirmed live) -- so even once the primary ask above lands, projects running
embedded mode have no `bd doctor` safety net at all today and need their own
application-level guard regardless of this issue's resolution. Flagging so the scope of
"projects currently unprotected by this check" is clear.

vishnujayvel:
## Summary

Fixes #4814.

`bd doctor`'s Child-Parent Dependencies check and `--fix-child-parent` only matched **child→parent** blocking edges. The mirror pattern, **parent→child** (`bd dep add parent child --type waits-for`), deadlocks ready-work through the parent-child hierarchy exactly the same way, but was invisible to both `bd doctor` and `bd dep cycles` (which deliberately excludes `parent-child` edges from cycle detection by design).

**Correction (2026-09-18, per bee-ghosttrack review):** an earlier version of this description claimed the write-time guard (`CheckBlockingHierarchyInTx`) only rejects the child→parent direction. That's wrong — it has rejected `blocks`/`conditional-blocks` symmetrically in **both** directions since 91336a3c8, whenever a real `parent-child` edge exists. What it can't catch is `waits-for` (a fanout-gate type deliberately excluded from that guard and from cycle detection, and reachable from the CLI today), plus rows that predate the guard or were written via import/direct SQL. This PR's parent→child arm targets that gap, not the already-guarded blocks/conditional-blocks path.

The SQL predicate (and the fix) now covers both directions. The parent→child arm additionally requires a **real** `parent-child` edge between the pair, not just a dotted-ID coincidence — IDs and edges diverge by design after a reparent or import, so a dotted-only pair with no structural link (e.g. an epic blocking on an orphaned-dotted subtask) is an ordinary "epic waits for its subtask" shape, not a deadlock, and must not be flagged or deleted:

```sql
d.type IN ('blocks', 'conditional-blocks', 'waits-for')
AND (
  d.issue_id LIKE CONCAT(d.depends_on_id, '.%')       -- child → parent (prefix-based; legacy/import rows only,
                                                       -- since the CLI already rejects new dotted child→ancestor edges)
  OR (
    d.depends_on_id LIKE CONCAT(d.issue_id, '.%')     -- parent → child
    AND EXISTS (                                      -- requires a REAL parent-child edge, not just an ID coincidence
      SELECT 1 FROM (...) pc
      WHERE pc.type = 'parent-child'
        AND pc.issue_id = d.depends_on_id
        AND pc.depends_on_id = d.issue_id
    )
  )
)
```

Messages reworded to "hierarchy-blocking anti-patterns" to cover both, and detail lines now tag which direction each flagged row is.

The issue's secondary ask — closing the `waits-for` gap at write time too, rather than catching it after the fact via doctor — would be a bigger behavioral change than this doctor-only detection/cleanup fix, and the issue itself frames it as a question rather than a bug. Left out of scope here.

Thanks to Akhil Kumar for the isolated repro and thorough analysis of why `bd dep cycles` and `bd dep add`'s pre-check both miss this direction, and to bee-ghosttrack for catching the write-time-guard mischaracterization and the missing real-edge requirement above.

## Test plan

- [x] `go test ./cmd/bd/doctor/ ./cmd/bd/doctor/fix/ -count=1 -timeout 25m` (icu4c CGO flags) — pass. (Correction, 2026-07-29: an earlier version of this test plan reported `TestCheckBeadsRole_NotConfigured` and `TestCheckBeadsRole_NotGitRepo` as failing on unmodified `main`. That was wrong. They are environment-local: our dev machine has a global `git config beads.role=maintainer`, and `cmd/bd/doctor/role_test.go:10-26` asserts a fresh repo has none. Both pass with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`. Upstream `main` is not red. Thanks to @maphew for catching it.)
- [x] `go test -tags gms_pure_go ./cmd/bd/doctor/... -run ChildParent -v` — all 11 cases pass, including two new ones added for this review round: a reversed-direction (parent→child) destructive case with a real parent-child edge, and a survivor case (dotted ID, no real parent-child edge) that must not be flagged/deleted. Both new tests were verified to fail against the pre-fix query and pass against the fix.

### DIFF
--- a/cmd/bd/doctor/fix/validation.go
+++ b/cmd/bd/doctor/fix/validation.go
@@ -112,9 +112,9 @@ func OrphanedDependencies(path string, verbose bool) error {
 	return nil
 }
 
-// ChildParentDependencies removes child→parent blocking dependencies.
-// These often indicate a modeling mistake (deadlock: child waits for parent, parent waits for children).
-// Requires explicit opt-in via --fix-child-parent flag since some workflows may use these intentionally.
+// ChildParentDependencies removes hierarchy-crossing BLOCKING dependencies:
+// child→parent and parent→child blocks/conditional-blocks/waits-for edges that
+// deadlock ready-work (GH#4814). Requires opt-in via --fix-child-parent.
 // If verbose is true, prints each removed dependency; otherwise shows only summary.
 func ChildParentDependencies(path string, verbose bool) error {
 	beadsDir, err := resolvedWorkspaceBeadsDir(path)
@@ -129,15 +129,16 @@ func ChildParentDependencies(path string, verbose bool) error {
 	}
 	defer db.Close()
 
-	// Find child→parent BLOCKING dependencies where issue_id starts with depends_on_id + "."
-	// Only matches blocking types (blocks, conditional-blocks, waits-for) that cause deadlock.
-	// Excludes 'parent-child' type which is a legitimate structural hierarchy relationship.
+	// Both directions of hierarchy-crossing blocking deps (GH#4814).
 	//nolint:gosec // G202: fixDependencyUnionSQL returns a fixed internal SELECT fragment.
 	query := `
 		SELECT d.dep_table, d.issue_id, d.depends_on_id, d.type
 		FROM (` + fixDependencyUnionSQL() + `) d
-		WHERE d.issue_id LIKE CONCAT(d.depends_on_id, '.%')
-		  AND d.type IN ('blocks', 'conditional-blocks', 'waits-for')
+		WHERE d.type IN ('blocks', 'conditional-blocks', 'waits-for')
+		  AND (
+			d.issue_id LIKE CONCAT(d.depends_on_id, '.%')
+			OR d.depends_on_id LIKE CONCAT(d.issue_id, '.%')
+		  )
 	`
 	rows, err := db.Query(query)
 	if err != nil {
@@ -164,11 +165,11 @@ func ChildParentDependencies(path string, verbose bool) error {
 	}
 
 	if len(badDeps) == 0 {
-		fmt.Println("  No child→parent dependencies to fix")
+		fmt.Println("  No hierarchy-blocking anti-patterns to fix")
 		return nil
 	}
 
-	// Delete child→parent blocking dependencies (preserving parent-child type)
+	// Delete anti-pattern blocking deps (preserving legitimate parent-child type).
 	// Uses explicit transaction so writes persist when @@autocommit is OFF
 	// (e.g. Dolt server started with --no-auto-commit).
 	showIndividual := verbose || len(badDeps) < 20
@@ -185,7 +186,7 @@ func ChildParentDependencies(path string, verbose bool) error {
 		case "wisp_dependencies":
 			_, err = tx.Exec("DELETE FROM wisp_dependencies WHERE issue_id = ? AND "+fixDependencyTargetExpr+" = ? AND type = ?", d.issueID, d.dependsOnID, d.depType)
 		default:
-			fmt.Printf("  Warning: skipped child→parent dependency from unexpected table %s\n", d.depTable)
+			fmt.Printf("  Warning: skipped hierarchy-blocking dependency from unexpected table %s\n", d.depTable)
 			continue
 		}
 		if err != nil {
@@ -193,7 +194,7 @@ func ChildParentDependencies(path string, verbose bool) error {
 		} else {
 			removed++
 			if showIndividual {
-				fmt.Printf("  Removed child→parent dependency: %s→%s\n", d.issueID, d.dependsOnID)
+				fmt.Printf("  Removed hierarchy-blocking dependency: %s→%s\n", d.issueID, d.dependsOnID)
 			}
 		}
 	}
@@ -204,7 +205,7 @@ func ChildParentDependencies(path string, verbose bool) error {
 	// Commit changes in Dolt
 	_, _ = db.Exec("CALL DOLT_COMMIT('-Am', 'doctor: remove child-parent dependency anti-patterns')") // Best effort: commit advisory; schema fix already applied in-memory
 
-	fmt.Printf("  Fixed %d child→parent dependency anti-pattern(s)\n", removed)
+	fmt.Printf("  Fixed %d hierarchy-blocking dependency anti-pattern(s)\n", removed)
 	return nil
 }
 
--- a/cmd/bd/doctor/validation.go
+++ b/cmd/bd/doctor/validation.go
@@ -417,15 +417,19 @@ func checkDoltConflicts(beadsDir string) DoctorCheck {
 
 // checkChildParentDependenciesDB is the core logic for CheckChildParentDependencies.
 func checkChildParentDependenciesDB(db *sql.DB) DoctorCheck {
-	// Query for child→parent BLOCKING dependencies where issue_id starts with target id + "."
-	// Only matches blocking types (blocks, conditional-blocks, waits-for) that cause deadlock.
-	// Excludes 'parent-child' type which is a legitimate structural hierarchy relationship.
+	// Blocking edges that cross a parent-child ID hierarchy deadlock ready-work:
+	//   child→parent  (issue_id is child of depends_on_id)  — already guarded by bd dep add
+	//   parent→child  (depends_on_id is child of issue_id)  — was invisible (GH#4814)
+	// Only blocks / conditional-blocks / waits-for; legitimate 'parent-child' edges are excluded.
 	//nolint:gosec // G202: doctorDependencyUnionSQL returns a fixed internal SELECT fragment.
 	query := `
 		SELECT d.issue_id, d.depends_on_id
 		FROM (` + doctorDependencyUnionSQL() + `) d
-		WHERE d.issue_id LIKE CONCAT(d.depends_on_id, '.%')
-		  AND d.type IN ('blocks', 'conditional-blocks', 'waits-for')
+		WHERE d.type IN ('blocks', 'conditional-blocks', 'waits-for')
+		  AND (
+			d.issue_id LIKE CONCAT(d.depends_on_id, '.%')
+			OR d.depends_on_id LIKE CONCAT(d.issue_id, '.%')
+		  )
 	`
 	rows, err := db.Query(query)
 	if err != nil {
@@ -457,7 +461,7 @@ func checkChildParentDependenciesDB(db *sql.DB) DoctorCheck {
 		return DoctorCheck{
 			Name:     "Child-Parent Dependencies",
 			Status:   "ok",
-			Message:  "No child→parent dependencies",
+			Message:  "No hierarchy-blocking anti-patterns",
 			Category: CategoryMetadata,
 		}
 	}
@@ -470,7 +474,7 @@ func checkChildParentDependenciesDB(db *sql.DB) DoctorCheck {
 	return DoctorCheck{
 		Name:     "Child-Parent Dependencies",
 		Status:   "warning",
-		Message:  fmt.Sprintf("%d child→parent dependency detected (may cause deadlock)", len(badDeps)),
+		Message:  fmt.Sprintf("%d hierarchy-blocking dependency(ies) detected (may cause deadlock)", len(badDeps)),
 		Detail:   detail,
 		Fix:      "Run 'bd doctor --fix --fix-child-parent' to remove (if unintentional)",
 		Category: CategoryMetadata,
--- a/cmd/bd/doctor/validation_test.go
+++ b/cmd/bd/doctor/validation_test.go
@@ -417,8 +417,8 @@ func TestCheckChildParentDependenciesDB_NoDeps(t *testing.T) {
 	if check.Status != StatusOK {
 		t.Errorf("Status = %q, want %q", check.Status, StatusOK)
 	}
-	if check.Message != "No child→parent dependencies" {
-		t.Errorf("Message = %q, want %q", check.Message, "No child→parent dependencies")
+	if check.Message != "No hierarchy-blocking anti-patterns" {
+		t.Errorf("Message = %q, want %q", check.Message, "No hierarchy-blocking anti-patterns")
 	}
 }
 
@@ -468,6 +468,45 @@ func TestCheckChildParentDependenciesDB_BlockingDetected(t *testing.T) {
 	}
 	if check.Message == "" {
 		t.Error("Message should not be empty")
+	}
+}
+
+// TestCheckChildParentDependenciesDB_ParentBlocksChildDetected is the GH#4814
+// mirror: parent→child blocks is as deadlocking as child→parent but was invisible.
+func TestCheckChildParentDependenciesDB_ParentBlocksChildDetected(t *testing.T) {
+	store := newTestDoltStore(t, "test")
+	ctx := context.Background()
+
+	parent := &types.Issue{Title: "Parent epic", Status: types.StatusOpen, Priority: 1, IssueType: types.TypeEpic}
+	if err := store.CreateIssue(ctx, parent, "test"); err != nil {
+		t.Fatalf("Failed to create parent: %v", err)
+	}
+
+	db := store.DB()
+	childID := parent.ID + ".1"
+
+	_, err := db.ExecContext(ctx,
+		`INSERT INTO issues (id, title, description, design, acceptance_criteria, notes, status, priority, issue_type, created_at, updated_at)
+		 VALUES (?, 'Hierarchical child', '', '', '', '', 'open', 2, 'task', NOW(), NOW())`,
+		childID)
+	if err != nil {
+		t.Fatalf("Failed to insert child: %v", err)
+	}
+
+	// parent blocks child (mirror anti-pattern)
+	_, err = db.ExecContext(ctx,
+		`INSERT INTO dependencies (id, issue_id, depends_on_issue_id, type, created_at, created_by) VALUES (UUID(), ?, ?, 'blocks', NOW(), 'test')`,
+		parent.ID, childID)
+	if err != nil {
+		t.Fatalf("Failed to insert dependency: %v", err)
+	}
+
+	check := checkChildParentDependenciesDB(db)
+	if check.Status != StatusWarning {
+		t.Fatalf("Status = %q, want %q (parent→child blocks must be detected)", check.Status, StatusWarning)
+	}
+	if !strings.Contains(check.Detail, parent.ID) || !strings.Contains(check.Detail, childID) {
+		t.Fatalf("Detail = %q, want both parent and child ids", check.Detail)
 	}
 }
 


maphew (commented):
The predicate change is the right one, and the added test is a real test rather than a decorative one: `TestCheckChildParentDependenciesDB_ParentBlocksChildDetected` inserts `(issue_id=parent, depends_on_issue_id=child, type='blocks')`, which the pre-change predicate at `cmd/bd/doctor/validation.go:424` cannot match, so it fails without the fix. The underlying report checks out too — a parent→child blocking edge deadlocks both beads through the parent-child blocked-state cascade (`DependencyType.AffectsReadyWork` at `internal/types/types.go:996` includes `parent-child`, and children inherit a blocked parent's state), so the mirror direction deserves the same treatment as the one already covered. Nice detail that the mirror case also picks up `waits-for`, which is exactly the type the write-time guard skips.

Detail below because `bd doctor` output is the thing people trust when nothing else explains a stuck bead — that level of scrutiny is what the check earns, not a judgement on the patch.

## One factual correction for the PR description

The description says the issue's secondary ask (a write-time guard for the parent→child direction) is "a bigger behavioral change" left out of scope. That guard already exists on main. `issueops.CheckBlockingHierarchyInTx` (`internal/storage/issueops/dependencies.go:497-525`, added in 91336a3c8, 2026-07-13) checks *both* directions — `blockerIsAncestor` at :507 and `blockerIsDescendant` at :516 — and is exercised by `internal/storage/embeddeddolt/dependencies_test.go:152-200` (`parent_child_shadow_rejected`), which asserts that `epic --blocks--> its own child` is rejected with a "descendant" error.

That does not remove the value of this PR, but it changes what the value is:

- rows written before 2026-07-13, or via import/direct SQL (`bd doctor` is the only thing that will ever see those);
- `waits-for` edges, which `CheckBlockingHierarchyInTx` returns early on at :498;
- prefix hierarchies with no `parent-child` edge — the guard walks edges, this check walks IDs, so neither subsumes the other.

Worth rewriting that paragraph so the change isn't mis-sold; the honest framing is still a good one.

## Should-fix: the destructive half is untested in the new direction

`cmd/bd/doctor/fix/validation.go` now *deletes* parent→child rows under `--fix-child-parent`, but the new test only covers detection. `cmd/bd/doctor/fix/validation_test.go` has three tests for the old direction (`TestChildParentDependencies_FixesBadDeps` at :171, `_PreservesParentChildType` at :224, `_NoBadDeps` at :125) and none for the new one. A near-copy of `_FixesBadDeps` with the edge reversed (`bd-abc` blocks `bd-abc.1`) would cover it — and since those tests only establish hierarchy via dotted IDs, `store.AddDependency` won't trip the edge-based hierarchy guard, so it's writable as a fixture.

## Should-fix: four user-facing strings still say "child→parent" only

The check message changed but the surrounding prose didn't, so agent-mode output now contradicts itself:

- `cmd/bd/doctor/agent.go:465` — "These are an anti-pattern where a child issue depends on its own parent." This is rendered directly under the new `"N hierarchy-blocking dependency(ies) detected"` message.
- `cmd/bd/doctor/validation.go:323-325` — exported godoc for `CheckChildParentDependencies` still describes one direction.
- `cmd/bd/doctor.go:346` — flag help `"Remove child→parent dependencies (opt-in)"`.
- `cmd/bd/doctor.go:174` — the `--fix-child-parent` example comment.

Note that changing the two strings in `doctor.go` means regenerating `docs/cli-reference/doctor.md:107,140` and `docs/CLI_REFERENCE.md:3905,3938` (`scripts/generate-cli-docs.sh`) or the doc-freshness job will go red — worth doing in the same push rather than discovering it in CI.

## Non-blocking observation: prefix hierarchy vs. edge hierarchy

The doctor check infers hierarchy from dotted IDs; the write-time guard infers it from actual `parent-child` edges (`isAncestorInTx`, `internal/storage/issueops/dependencies.go:532`). After `bd dep reparent` the ID keeps its old dotted prefix while the edge moves, so `bd-1 blocks bd-1.2` can be a perfectly legitimate edge between unrelated beads and would now be flagged — and removed by an opt-in fix. This already applied to the child→parent direction, so it isn't introduced here, and the fix is opt-in, so it isn't urgent. If you want to close it, joining the query against `parent-child` rows would make the check agree with the guard about what "hierarchy" means. Fine to leave for a follow-up.

Also flagging, purely for context: the second commenter on #4814 documents a 3-node case (`parent → unrelated → child`) that no prefix predicate can catch, and suggests widening `cycleRelevantDepType` in `DetectCycles` instead. That's a separate change and doesn't conflict with this one — this PR is a strict improvement on the adjacency case.

Happy to push the fix-path test and the four string updates onto this branch myself if you'd rather not spin another round — say the word, or take them yourself if you prefer to keep the branch yours.

vishnujayvel:
Thanks — the correction that `CheckBlockingHierarchyInTx` (added 2026-07-13) already guards the write-time parent→child direction is the important one: it reframes what this PR is worth. We were wrong to call the write-time guard "out of scope as a bigger behavioral change" — it exists. We'll rewrite that paragraph so the PR is sold honestly: this catches pre-2026-07-13 rows, import/direct-SQL writes, `waits-for` edges, and prefix hierarchies with no `parent-child` edge — none of which the write-time guard covers.

Also correcting our test plan: we reported `TestCheckBeadsRole_NotConfigured`/`_NotGitRepo` as failing on main. That's wrong — they're environment-local. Our dev machine has a global `git config beads.role=maintainer`; `role_test.go:10-26` asserts a fresh repo has none. Confirmed directly: both fail with that config present and pass (`ok ... doctor 2.060s`) with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`. `main` isn't red here; our harness now neutralizes global git config so this can't recur.

Please push the fix-path test for the new direction and the four string updates (agent.go, doctor.go x2, validation.go godoc) plus doc regen as maintainer commits.

vishnujayvel:
Rebased onto 71c4cd08b9991e7ba4e53822769c78337e0f4ee4. Merge conflict in `cmd/bd/doctor/fix/validation.go` resolved by keeping the `guardFixTarget` added in main while applying the PR's query and comments.

bee-ghosttrack (changes requested):
First desk review, at `ed4c6e919` (rebased onto `71c4cd08b`).

**The rebase resolution is clean.** `git diff 71c4cd08b...HEAD -- cmd/bd/doctor/fix/validation.go` is +14/−13 and every hunk is query, comment or message. `guardFixTarget("Child-parent dependencies fix", …)` is present once at `:136`, ahead of the query; `fixDependencyUnionSQL`, `fixDependencyTargetExpr`, the `dep_table` switch and the `Begin`/`Commit` pair are untouched.

**The deadlock is real, and it is reachable from today's CLI — through `waits-for`.** Reproduced with a `bd` built from this head: epic → child → grandchild with real `--parent` edges, then `bd dep add <epic> <child> --type waits-for`. All three go blocked, `bd ready` is empty, nothing can close. The cascade is `internal/storage/issueops/blocked_consistency.go:420-430` (a child inherits `is_blocked` over `parent-child` edges), and `CheckBlockingHierarchyInTx` returns early for anything that isn't `blocks`/`conditional-blocks` (`internal/storage/issueops/dependencies.go:524-526`), so the write-time guard refuses the `blocks` version of this and accepts the `waits-for` one. That is what this PR is worth, and it's a better pitch than the one in the body.

## Asks

**1. The new arm deletes healthy edges that the CLI writes on purpose; the old arm never could.** The thread settled on "the dotted-ID-vs-edge gap already existed in the child→parent direction, so it isn't introduced here". It isn't symmetric. There is a second, ID-based write-time guard — `isDisallowedHierarchicalDependency`, `cmd/bd/dep.go:137` — that rejects a dotted child → any ancestor edge of any blocking type. So a dotted-only child→parent row is unreachable from the CLI, and the old arm only ever fired on legacy/import rows. The mirror is wide open:

```
$ bd dep add qq-5g0.7 qq-5g0      # child→parent, dotted ID only, no parent-child edge
Error: cannot add dependency: qq-5g0.7 is already a child of qq-5g0. …
$ bd dep add qq-5g0 qq-5g0.7      # parent→child, dotted ID only, no parent-child edge
✓ Added dependency: qq-5g0 (EpicB) depends on qq-5g0.7 (OrphanDotted) (blocks)
```

With no `parent-child` edge there is no cascade, so that second edge is not a deadlock: `qq-5g0.7` shows in `bd ready`, the epic waits, and it resolves when the subtask closes. It is the ordinary "epic waits for its subtask" shape. At head, doctor warns on it and `--fix-child-parent` removes it, in the same breath as the real one:

```
Removed hierarchy-blocking dependency: qq-k4k→qq-k4k.1     # real deadlock
Removed hierarchy-blocking dependency: qq-5g0→qq-5g0.7     # healthy edge
```

IDs and edges diverge by design after a reparent (`cmd/bd/reparent_test.go:15-57`), and on import. Please make the **new** arm require a real `parent-child` edge between the pair — the guard's own definition of hierarchy — in both the check and the fixer, so the two agree with each other and with the cascade that makes this a deadlock at all. The legacy arm can stay prefix-based. maphew floated this join as an optional follow-up; given the asymmetry above I think it belongs in the PR that adds the arm.

**2. The destructive half is still untested in the new direction.** `cmd/bd/doctor/fix/validation_test.go` has three `ChildParent` tests and every fixture edge runs child→parent (`:228-230`). maphew asked for the reversed case on 07-29; you asked for it as a maintainer commit the same day; nobody pushed it, and the desk doesn't push to contributor branches. It's a near-copy of `_FixesBadDeps` with `{"bd-abc", "bd-abc.1"}` — plus, once ask 1 is in, one row that has the dotted ID and no `parent-child` edge and must *survive* the fix. (The delete itself is direction-agnostic; the reader watched it remove both a `waits-for` and a `blocks` parent→child row. This is a missing test, not broken code.)

**3. Agent-mode output now says the opposite of the rows under it.** `cmd/bd/doctor/agent.go:477` still reads "Child→parent dependencies found: … an anti-pattern where a child issue depends on its own parent", rendered directly above `qq-k4k→qq-k4k.1` — a parent depending on its child. That is the surface an agent acts on. The rest of the one-direction strings, including one that wasn't on the 07-29 list:

- `cmd/bd/doctor_fix.go:341` — "⚠ Child→parent deps require explicit opt-in" (new to this list; it's the first thing a user reads about this check when they run `--fix` without the flag)
- `cmd/bd/doctor.go:174` and `:351` — the example comment and the `--fix-child-parent` flag help
- `cmd/bd/doctor/validation.go:321-322` — exported godoc

One correction to the earlier review: **no doc regen is needed.** `docs/cli-docs.pin` pins `v1.2.2` and `scripts/resolve-docs-bd.sh` builds the drift/flag checks from the pinned tag, so changing the flag help won't turn doc freshness red.

Small thing while you're there: the Detail line (`validation.go:446`, `"%s→%s"`) no longer tells a reader which direction a row is; tagging it would help.

**4. The PR body.** It still says the write-time guard for this direction is "intentionally not included here … a bigger behavioral change". You agreed on 07-29 to rewrite it. The honest version from above: the guard exists for `blocks`/`conditional-blocks` since 91336a3c8; this PR catches `waits-for` (live today), rows that predate the guard, and import/direct-SQL writes.

## Follow-up, not this PR

After `--fix-child-parent` removed the rows, `bd blocked` still listed the child and grandchild as blocked and `bd ready` stayed empty: the "Blocked State" recompute only runs if that check had *failed at scan time* (`cmd/bd/doctor_fix.go:232-239`), and a consistently-deadlocked graph passes it. Pre-existing and equally true of the old direction, but the people running this fix are by definition chasing a stuck bead. I'll file it.

## Verified at head

`gofmt -l` clean on the three files; `go vet` clean on both packages; `go test -tags gms_pure_go ./cmd/bd/doctor/ -run ChildParent` ok (5 pass, incl. `_ParentBlocksChildDetected`); `./cmd/bd/doctor/fix/ -run ChildParent` ok (3 pass, all old-direction); full `fix` package ok. Full `doctor` package: 4 failures, the same 4 on base `71c4cd08b` — environment/ordering, none touch this path. Query plan is one `Filter` over the `UNION ALL` of the two dep tables — no join, one extra `LIKE` per row.

_Claude Fable 5.1 (desk) with a Claude Opus 5 adversarial reader, on behalf of bee; cited lines re-read by the desk before posting._

bee-ghosttrack:
Filed the follow-up I promised in the review: #6608 — `bd doctor --fix --fix-child-parent` removes the row but leaves `is_blocked` stale, because the Blocked State recompute only runs when that check itself failed at scan time. Reproduced on `main` @ 71c4cd08b in the existing (descendant→ancestor) direction, so it is pre-existing and **not** a blocker for this PR; the asks in the review stand as posted. If you extend the fix-path test here, asserting `is_blocked` after the fix would pin #6608's behavior once it's fixed — optional.

vishnujayvel:
Thanks for the deep-dive — pushed `bb31d39f6` addressing all four points.

**1. New arm deleted healthy edges (dotted-ID coincidence vs. real edge).** Fixed in both the checker (`checkChildParentDependenciesDB`) and the fixer (`ChildParentDependencies`): the parent→child arm now requires an `EXISTS` check for a real `type = 'parent-child'` edge between the pair, not just the dotted-ID `LIKE` pattern. The legacy child→parent arm is left prefix-based, per your note that it's already effectively write-time-guarded and only fires on legacy/import rows.

**2. Reversed-direction destructive case untested.** Added `TestChildParentDependencies_FixesBadDeps_ReversedDirection` (fixer) and `TestCheckChildParentDependenciesDB_ParentBlocksChildDetected` was updated to add a real parent-child edge (it previously asserted detection from the dotted ID alone, which is exactly the false-positive shape from point 1). Also added the survivor tests you flagged as missing: `TestChildParentDependencies_PreservesDottedEdgeWithoutRealParentChildLink` and `TestCheckChildParentDependenciesDB_ParentBlocksChildIgnoredWithoutRealEdge` — a dotted pair with no real parent-child edge now survives both the check and the fix. I verified both new "survivor" tests fail against the pre-fix query and pass against the fix (reverted the SQL locally, reran, restored). The reversed-direction destructive test uses `waits-for` rather than `blocks`/`conditional-blocks`, since those are already rejected at write time by `CheckBlockingHierarchyInTx` once a real parent-child edge exists — matches your point 4 finding.

**3. One-directional messages.** Reworded `cmd/bd/doctor/agent.go:477`, `cmd/bd/doctor_fix.go:341`, `cmd/bd/doctor.go` (example comment + flag help), and the `CheckChildParentDependencies` godoc to describe both directions instead of only child→parent. Also tagged each flagged row in `checkChildParentDependenciesDB`'s `Detail` output with `(child→parent)` / `(parent→child)` per your "small thing while you're there" note. Left the `--fix-child-parent` flag name itself unchanged (renaming it is a bigger compat concern than this review asked for).

**4. PR body's write-time-guard claim.** Rewrote the summary: the guard has been symmetric for `blocks`/`conditional-blocks` since 91336a3c8, and this PR's actual value is the `waits-for` gap plus legacy/import rows — not "child→parent only" as the previous draft said. Also corrected the secondary-ask paragraph accordingly.

Follow-up on the "Blocked State" recompute gap you mentioned — leaving that to the issue you said you'd file, out of scope here.

Not touching doc regen per your correction. Not resolving your review threads — leaving that to you to confirm the fixes land as expected.

bee-ghosttrack (approved):
Re-review at `bb31d39f6`. All four asks are in, and I checked the code rather than the reply.

**1. Real-edge join.** Both the checker and the fixer now gate the parent→child arm on an `EXISTS` over the dep union for a `parent-child` row with `issue_id = child, depends_on_id = parent`, which is the direction `bd dep add --parent` writes. The legacy arm stays prefix-based, as agreed. A dotted-only pair with no structural link survives both (`…IgnoredWithoutRealEdge`, `…PreservesDottedEdgeWithoutRealParentChildLink`), and I read the query, not just the test names.

**2. Reversed-direction destructive test.** `TestChildParentDependencies_FixesBadDeps_ReversedDirection` uses `waits-for`, which is the right choice: it is the type the write-time guard lets through, so it is the shape that actually reaches the fixer.

**3. Strings.** Agent enrichment, the opt-in warning, the flag help, the example comment and the godoc all say both directions now, and the `Detail` rows carry `(child→parent)` / `(parent→child)`. Keeping the `--fix-child-parent` flag name is the right call.

**4. Body** now states the guard history and the `waits-for` gap. Good.

The Blocked State recompute gap is tracked as #6608; nothing here waits on it.

**Verified at head:** `gofmt -l` clean; `go vet -tags gms_pure_go` clean on both packages; `go test -tags gms_pure_go ./cmd/bd/doctor/ -run ChildParent` 6/6 pass; the full `./cmd/bd/doctor/fix/` package green (246 s), including the five `ChildParent` tests. @julianknutsen, merge on green.

_Claude Fable 5.1 (desk) on behalf of bee; cited lines re-read before posting._

PR #5132, opened 2026-07-29
fix(doctor): run graph integrity checks without CGO

### LINKED ISSUE #4977: doctor: the //go:build cgo gate is over-broad — 13 checks are stubbed out over a code path upstream documents as pure Go (graph integrity dark in every CGO_ENABLED=0 build)
> **Provenance:** this issue was researched and written by a **Claude Code session** (Opus 4.8) working in my repo, and posted under my account. Flagging it so you can weight it accordingly. Everything below was verified against the source at `02e38d0` and against two locally built binaries — the A/B measurement is from a live 233-issue server-mode store, not inferred.

**Affected at current tip (`02e38d0`):** `cmd/bd/doctor/checks_nocgo.go`, `cmd/bd/doctor/migration_validation_nocgo.go`, and the `//go:build cgo` tags on `cmd/bd/doctor/{validation,maintenance,migration_validation}.go`

## Summary

13 doctor check functions are stubbed out behind `//go:build !cgo` and report **`Skipped: requires CGO`**. None of them needs CGO. Their real implementations are plain SQL over a `*sql.DB` obtained from the **server-mode** Dolt client — the one path in the tree that is explicitly documented as CGO-free.

In a `CGO_ENABLED=0` build, `bd doctor` renders 10 of these as ordinary ⚠ warnings, visually identical to a real finding:

```
⚠  Dolt Locks: Skipped: requires CGO
⚠  Orphaned Dependencies: Skipped: requires CGO
⚠  Child-Parent Dependencies: Skipped: requires CGO
⚠  Duplicate Issues: Skipped: requires CGO
⚠  Test Pollution: Skipped: requires CGO
⚠  Stale Closed Issues: Skipped: requires CGO
⚠  Stale Molecules: Skipped: requires CGO
⚠  Persistent Mol Issues: Skipped: requires CGO
⚠  Stale MQ Files: Skipped: requires CGO
⚠  Patrol Pollution: Skipped: requires CGO
```

Three of those — Orphaned Dependencies, Child-Parent Dependencies, Duplicate Issues — are the graph-integrity checks. In my repo they had been dark for a month across a 233-issue graph while `bd doctor` reported ✓ 65 passed / ⚠ 15 warnings / ✖ 0 errors.

## The tag is over-broad — the checks never touched the embedded backend

Following `CheckOrphanedDependencies` down, at `02e38d0`:

| Step | Location | CGO needed? |
|---|---|---|
| `CheckOrphanedDependencies` → `openStoreDB` | `doctor/validation.go:41`, `:21` | tagged `//go:build cgo` |
| `openStoreDB` → `getDatabasePath` | `doctor/database.go:474` — **untagged file** | no |
| `openStoreDB` → `doltServerConfig` | `doctor/federation.go:31` — **untagged file** | no |
| `openStoreDB` → `dolt.New(ctx, cfg)` | `internal/storage/dolt/store.go:1059` — **no build constraints on any non-test file in the package** | no |
| `dolt.New` → `newServerMode` | `internal/storage/dolt/store.go:1080` | no |

`dolt.New` calls `newServerMode` unconditionally, and the comment above it says so directly:

```go
// newServerMode creates a DoltStore connected to a running dolt sql-server.
// This path is pure Go and does not require CGO.
```

Everything after that is `db.Query(...)` against `*sql.DB`.

**The proof that it links today:** the `!cgo` store factory already calls the identical constructor.

```go
// cmd/bd/store_factory_nocgo.go:31 — //go:build !cgo
func newDoltStore(ctx context.Context, cfg *dolt.Config) (storage.DoltStorage, error) {
	...
	if !cfg.ServerMode {
		return nil, fmt.Errorf("%s", nocgoEmbeddedErrMsg)
	}
	return dolt.New(ctx, cfg)
}
```

So `dolt.New` is compiled and linked into every `CGO_ENABLED=0` `bd`. The doctor checks are stubbed out over a symbol the binary already contains.

There is a second, sharper consequence: `openStoreDB` builds its config with `doltServerConfig` (`ServerHost` / `ServerPort` / `ServerUser`), unconditionally. **These checks never used the embedded backend at all** — not even in a CGO build. The tag isn't buying embedded support for anyone; it's just switching the checks off.

## It's inconsistent with a sibling check

`CheckCrossTableDuplicates` (`cmd/bd/doctor/cross_table.go`) is the same shape — untagged file, hits the database, and degrades at runtime:

```go
count, err := fix.CountCrossTableDuplicates(path)
if err != nil {
	return DoctorCheck{Name: "Cross-Table Duplicates", Status: StatusOK,
		Message: "N/A (no database or wisps table absent)"}
}
```

It runs fine in a nocgo build. It sits in the same `[]validateCheckResult` slice as three of the stubbed ones (`doctor_validate.go:84-88`). That asymmetry is what makes this read as an oversight rather than a decision.

## Who is affected

Not the mainline release binaries — `.goreleaser.yaml` and `release.yml:211,229` build `CGO_ENABLED=1` for darwin arm64/amd64, linux amd64/arm64 and windows/amd64, guarded by `scripts/verify-cgo.sh`. The affected population is:

1. **The three `CGO_ENABLED=0` release tiers** — `bd-android-arm64`, `bd-windows-arm64`, `bd-freebsd-amd64`. These knowingly ship without *embedded* mode (#3538), but **server mode is fully supported there** — `nocgoEmbeddedErrMsg` itself points users at `bd init --server` / `bd init --proxied-server`. Losing server-mode integrity checks on those platforms is unintended collateral, not part of that trade.

2. **`go install` users, systematically.** `engdocs/ICU-POLICY.md` states that "the module cannot make plain `go install pkg@latest` use `-tags gms_pure_go` automatically", and the `make diagnose` target spells out the resulting trap: a bare `CGO_ENABLED=1 go build ./cmd/bd` fails with a C-linker error from `go-icu-regex`, and `CGO_ENABLED=0` is the obvious escape. `.goreleaser.yaml:215` documents the correct incantation (`CGO_ENABLED=1 GOFLAGS=-tags=gms_pure_go go install …`) — but a user who doesn't find it lands on a nocgo binary. That's how I got here: `go version -m $(which bd)` shows `CGO_ENABLED=0` and no `-tags`.

So the ICU policy and the doctor cgo gate interact: the documented-hard install path hands users a `bd` whose graph-integrity checks are silently off.

## Verified: a CGO rebuild is the only variable

Same module version, same live server-mode store, same moment:

| Binary | Result | `Skipped: requires CGO` lines |
|---|---|---|
| stock `go install` (`CGO_ENABLED=0`) | ✓ 65 / ⚠ 15 / ✖ 0 | **10** |
| `CGO_ENABLED=1` rebuild of `@v1.1.0` | ✓ 72 / ⚠ 8 / ✖ 0 | **0** |

Reconciles exactly: 65 + 7 newly-passing = 72; 15 − 10 skips + 3 real findings = 8. The 3 real findings were then confirmed independently by replaying each check's SQL through `bd sql --readonly` — 9 of 9 comparable checks agree with the CGO binary, including both non-zero ones. Two of the newly-lit checks surfaced genuine (if minor) findings that the nocgo binary could not have reported.

## Suggested fix

Gate on **backend availability at runtime**, not on the build tag — the code already does this everywhere else:

1. Move the check bodies out of the `cgo`-tagged files into untagged ones. They import only `internal/storage/dolt`, `internal/configfile`, `internal/storage` and `internal/types`, none of which carry build constraints on non-test files.
2. Delete `checks_nocgo.go` and `migration_validation_nocgo.go`.
3. Rely on the existing fallback: `openStoreDB` returning an error already yields `N/A (no database)` (`validation.go:46,127,211,334`). An embedded-mode user on a nocgo binary would get that instead of `Skipped: requires CGO` — which is both accurate and no worse than today.

## Secondary UX issue, worth fixing regardless of the above

A skipped check is emitted as `StatusWarning`, rendered identically to a real finding. "Skipped: requires CGO" therefore reads as a benign nit rather than *this check did not run*. It did not fail and it did not pass — it's an unknown, and unknowns being displayed as warnings is how three integrity checks stayed dark for a month in my repo without anyone noticing.

A distinct `StatusSkipped` / `StatusUnknown` (rendered `⊘` or similar, and summarised separately from `⚠ N warnings`) would make dark checks legible. Applies to the `N/A (no database)` path too, which currently reports `StatusOK` — a check that couldn't run should never be counted among the passes.

## Relationship to #4249

@maphew's #4249 proposes a build-tagged zstd seam so embedded Dolt itself can build without cgo, and mentions "decoupling beads' code from the `cgo` build tag" as part of that work. If that lands, this is subsumed.

They're independent, though, and this one is much smaller: it doesn't touch Dolt, doesn't need a `replace`, and doesn't depend on the gozstd question resolving. It's a file move plus deleting two stub files, fixing checks that were only ever server-mode to begin with.

---

vishnujayvel:
## Summary

Fixes #4977.

13 `bd doctor` check functions — including the three graph-integrity checks (Orphaned Dependencies, Child-Parent Dependencies, Duplicate Issues) — were gated behind `//go:build cgo` and stubbed out as `Skipped: requires CGO` in `CGO_ENABLED=0` builds. None of them actually need CGO: they run plain SQL over the server-mode Dolt client (`*sql.DB`), which is the one storage path in the tree that's explicitly documented as CGO-free, and which the `!cgo` store factory already links into every nocgo `bd` binary. The tag wasn't buying embedded-mode support for anyone; it was just switching working checks off.

This drops the `//go:build cgo` tags from the real implementations and deletes the `!cgo` stub files (`checks_nocgo.go`, `migration_validation_nocgo.go`). The existing "no database" fallback (`N/A (no database)`) is unchanged for embedded-mode users on a nocgo binary.

Thanks to Eric for the detailed source trace and the A/B measurement across a CGO and nocgo build of the same live store.

## Test plan

- [x] `go test ./cmd/bd/doctor/ -count=1 -timeout 25m` (icu4c CGO flags) — pass. (Correction, 2026-07-29: an earlier version of this test plan reported `TestCheckBeadsRole_NotConfigured` and `TestCheckBeadsRole_NotGitRepo` as failing on unmodified `main`. That was wrong. They are environment-local: our dev machine has a global `git config beads.role=maintainer`, and `cmd/bd/doctor/role_test.go:10-26` asserts a fresh repo has none. Both pass with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`. Upstream `main` is not red. Thanks to @maphew for catching it.)


maphew (commented):
The source trace in #4977 is correct, and I re-verified each link of it rather than taking the description on faith:

- `internal/storage/dolt/store.go:1276-1308` — `dolt.New` ends in `return newServerMode(ctx, cfg)` unconditionally. There is no embedded branch left in that package at all; embedded lives in `internal/storage/embeddeddolt`, which has its own `!cgo` stubs.
- `cmd/bd/doctor/validation.go:21-34` — `openStoreDB` builds its config via `doltServerConfig` (`cmd/bd/doctor/federation.go:31`), which only ever sets `ServerHost`/`ServerPort`/`ServerUser`. These checks never used the embedded backend, even in a CGO build.
- `cmd/bd/store_factory_nocgo.go` already calls `dolt.New` behind `//go:build !cgo`, so the symbol is linked into every nocgo `bd` today. The stubs were shadowing code the binary already contained.
- `cmd/bd/doctor/maintenance.go:525` reaches the database through `dolt.NewFromConfigWithCLIOptions`, which is a different entry point but bottoms out in the same `New` -> `newServerMode` (`internal/storage/dolt/open.go:80-105`). So that path is server-mode too.

Build verification on the head commit (f1c9ddef6):

- `CGO_ENABLED=0 go build ./...` and `CGO_ENABLED=0 go vet ./cmd/bd/doctor/` both pass.
- Same two commands pass again after a clean trial merge with current `main` (the branch is 152 commits behind; the merge is conflict-free).
- `./scripts/check-build-tags.sh` is clean.
- All 11 functions in the deleted `checks_nocgo.go` and all 3 in `migration_validation_nocgo.go` have exactly one real definition in the newly-untagged files, and `grep` finds no remaining reference to the `"Skipped: requires CGO"` string outside the unrelated `embeddeddolt`/`federation` stubs — so nothing else was depending on the stub behavior.
- The CGO build is unaffected: the three files are byte-identical apart from the removed constraint line, so they compile into the cgo binary exactly as before. Risk is confined to the nocgo build.
- Deleting `migration_validation_nocgo_test.go` is required, not a coverage loss: it only asserted the stubs' `Skipped: requires CGO` strings, which no longer exist.

On the two test failures in your test plan (`TestCheckBeadsRole_NotConfigured`, `TestCheckBeadsRole_NotGitRepo`): both pass here on your branch. `cmd/bd/doctor/role_test.go:10-26` asserts that a fresh repo has no `beads.role`, so a global `git config beads.role` on the machine makes it fail. They are environment-local, not pre-existing breakage on `main` — worth restating that way so the note does not read as "main is red".

One thing to address before merge, caught in a second-model cross-check of this review:

**The nocgo failure mode flips from always-warn to possibly-false-green.** The deleted stubs returned `StatusWarning` ("Skipped: requires CGO") for every check, so a nocgo `bd doctor --check=validate` could never exit 0 — useless, but never misleading. After this PR, the real implementations run, and when no database is reachable they return `StatusOK` with "N/A (no database)" (`validation.go:40-45`, `:121-126`) — so a nocgo build with an unreachable Dolt server now reports a passing validate having executed zero integrity checks. That is parity with what the CGO build does in the same situation, and #4977 tracks the `StatusSkipped` idea separately, but this PR is the moment the always-warn guard disappears, so it should carry the mitigation: either return a warning/skip status for the no-database branches, or (minimally) an exit-code test pinning the intended nocgo no-DB behavior plus a changelog line. Happy to absorb this as a maintainer commit on your branch if you prefer.

Non-blocking follow-ups:

1. The PR title starts with `# `. A squash merge takes the PR title as the commit subject, so it would land as `# fix(doctor): ...`. The same artifact is on all eight PRs you opened in that batch (5128-5134); a title edit fixes it. The commit subject also uses a `GH`-prefixed #4977 where a bare `#4977` autolinks (the body's `Fixes #4977` is right).
2. `.github/workflows/pr.yml:102-107` compiles pure-Go *test* binaries only for `./cmd/bd`, `./internal/storage/embeddeddolt` and `./internal/tracker`. `go build ./cmd/bd` covers the doctor package's non-test files transitively, so this change is guarded at the level that matters — but nothing compiles `./cmd/bd/doctor`'s test files under `CGO_ENABLED=0`, which is precisely the contamination class that job's comment describes, and this PR enlarges doctor's untagged surface. Adding `go test -tags gms_pure_go -c -o /tmp/bd-doctor-puregeo-test ./cmd/bd/doctor` there (and to the `main.yml:138-144` mirror) would lock the win in. It passes today.
3. `cmd/bd/doctor/{validation,maintenance,migration_validation}_test.go` are still `//go:build cgo` because they need the cgo-tagged `test_helpers_test.go`. That is fine for this PR — the code under test is identical in both modes — but it means the newly-enabled functions have no test presence in the build configuration the PR is about. A follow-up issue is the right home for that.

maphew (commented):
The source trace in #4977 is correct, and I re-verified each link of it rather than taking the description on faith:

- `internal/storage/dolt/store.go:1276-1308` — `dolt.New` ends in `return newServerMode(ctx, cfg)` unconditionally. There is no embedded branch left in that package at all; embedded lives in `internal/storage/embeddeddolt`, which has its own `!cgo` stubs.
- `cmd/bd/doctor/validation.go:21-34` — `openStoreDB` builds its config via `doltServerConfig` (`cmd/bd/doctor/federation.go:31`), which only ever sets `ServerHost`/`ServerPort`/`ServerUser`. These checks never used the embedded backend, even in a CGO build.
- `cmd/bd/store_factory_nocgo.go` already calls `dolt.New` behind `//go:build !cgo`, so the symbol is linked into every nocgo `bd` today. The stubs were shadowing code the binary already contained.
- `cmd/bd/doctor/maintenance.go:525` reaches the database through `dolt.NewFromConfigWithCLIOptions`, which is a different entry point but bottoms out in the same `New` -> `newServerMode` (`internal/storage/dolt/open.go:80-105`). So that path is server-mode too.

Build verification on the head commit (f1c9ddef6):

- `CGO_ENABLED=0 go build ./...` and `CGO_ENABLED=0 go vet ./cmd/bd/doctor/` both pass.
- Same two commands pass again after a clean trial merge with current `main` (the branch is 152 commits behind; the merge is conflict-free).
- `./scripts/check-build-tags.sh` is clean.
- All 11 functions in the deleted `checks_nocgo.go` and all 3 in `migration_validation_nocgo.go` have exactly one real definition in the newly-untagged files, and `grep` finds no remaining reference to the `"Skipped: requires CGO"` string outside the unrelated `embeddeddolt`/`federation` stubs — so nothing else was depending on the stub behavior.
- The CGO build is unaffected: the three files are byte-identical apart from the removed constraint line, so they compile into the cgo binary exactly as before. Risk is confined to the nocgo build.
- Deleting `migration_validation_nocgo_test.go` is required, not a coverage loss: it only asserted the stubs' `Skipped: requires CGO` strings, which no longer exist.

On the two test failures in your test plan (`TestCheckBeadsRole_NotConfigured`, `TestCheckBeadsRole_NotGitRepo`): both pass here on your branch. `cmd/bd/doctor/role_test.go:10-26` asserts that a fresh repo has no `beads.role`, so a global `git config beads.role` on the machine makes it fail. They are environment-local, not pre-existing breakage on `main` — worth restating that way so the note does not read as "main is red".

One thing to address before merge, caught in a second-model cross-check of this review:

**The nocgo failure mode flips from always-warn to possibly-false-green.** The deleted stubs returned `StatusWarning` ("Skipped: requires CGO") for every check, so a nocgo `bd doctor --check=validate` could never exit 0 — useless, but never misleading. After this PR, the real implementations run, and when no database is reachable they return `StatusOK` with "N/A (no database)" (`validation.go:40-45`, `:121-126`) — so a nocgo build with an unreachable Dolt server now reports a passing validate having executed zero integrity checks. That is parity with what the CGO build does in the same situation, and #4977 tracks the `StatusSkipped` idea separately, but this PR is the moment the always-warn guard disappears, so it should carry the mitigation: either return a warning/skip status for the no-database branches, or (minimally) an exit-code test pinning the intended nocgo no-DB behavior plus a changelog line. Happy to absorb this as a maintainer commit on your branch if you prefer.

Non-blocking follow-ups:

1. The PR title starts with `# `. A squash merge takes the PR title as the commit subject, so it would land as `# fix(doctor): ...`. The same artifact is on all eight PRs you opened in that batch (5128-5134); a title edit fixes it. The commit subject also uses a `GH`-prefixed #4977 where a bare `#4977` autolinks (the body's `Fixes #4977` is right).
2. `.github/workflows/pr.yml:102-107` compiles pure-Go *test* binaries only for `./cmd/bd`, `./internal/storage/embeddeddolt` and `./internal/tracker`. `go build ./cmd/bd` covers the doctor package's non-test files transitively, so this change is guarded at the level that matters — but nothing compiles `./cmd/bd/doctor`'s test files under `CGO_ENABLED=0`, which is precisely the contamination class that job's comment describes, and this PR enlarges doctor's untagged surface. Adding `go test -tags gms_pure_go -c -o /tmp/bd-doctor-puregeo-test ./cmd/bd/doctor` there (and to the `main.yml:138-144` mirror) would lock the win in. It passes today.
3. `cmd/bd/doctor/{validation,maintenance,migration_validation}_test.go` are still `//go:build cgo` because they need the cgo-tagged `test_helpers_test.go`. That is fine for this PR — the code under test is identical in both modes — but it means the newly-enabled functions have no test presence in the build configuration the PR is about. A follow-up issue is the right home for that.

maphew (commented):
The source trace in #4977 is correct, and I re-verified each link of it rather than taking the description on faith:

- `internal/storage/dolt/store.go:1276-1308` — `dolt.New` ends in `return newServerMode(ctx, cfg)` unconditionally. There is no embedded branch left in that package at all; embedded lives in `internal/storage/embeddeddolt`, which has its own `!cgo` stubs.
- `cmd/bd/doctor/validation.go:21-34` — `openStoreDB` builds its config via `doltServerConfig` (`cmd/bd/doctor/federation.go:31`), which only ever sets `ServerHost`/`ServerPort`/`ServerUser`. These checks never used the embedded backend, even in a CGO build.
- `cmd/bd/store_factory_nocgo.go` already calls `dolt.New` behind `//go:build !cgo`, so the symbol is linked into every nocgo `bd` today. The stubs were shadowing code the binary already contained.
- `cmd/bd/doctor/maintenance.go:525` reaches the database through `dolt.NewFromConfigWithCLIOptions`, which is a different entry point but bottoms out in the same `New` -> `newServerMode` (`internal/storage/dolt/open.go:80-105`). So that path is server-mode too.

Build verification on the head commit (f1c9ddef6):

- `CGO_ENABLED=0 go build ./...` and `CGO_ENABLED=0 go vet ./cmd/bd/doctor/` both pass.
- Same two commands pass again after a clean trial merge with current `main` (the branch is 152 commits behind; the merge is conflict-free).
- `./scripts/check-build-tags.sh` is clean.
- All 11 functions in the deleted `checks_nocgo.go` and all 3 in `migration_validation_nocgo.go` have exactly one real definition in the newly-untagged files, and `grep` finds no remaining reference to the `"Skipped: requires CGO"` string outside the unrelated `embeddeddolt`/`federation` stubs — so nothing else was depending on the stub behavior.
- The CGO build is unaffected: the three files are byte-identical apart from the removed constraint line, so they compile into the cgo binary exactly as before. Risk is confined to the nocgo build.
- Deleting `migration_validation_nocgo_test.go` is required, not a coverage loss: it only asserted the stubs' `Skipped: requires CGO` strings, which no longer exist.

On the two test failures in your test plan (`TestCheckBeadsRole_NotConfigured`, `TestCheckBeadsRole_NotGitRepo`): both pass here on your branch. `cmd/bd/doctor/role_test.go:10-26` asserts that a fresh repo has no `beads.role`, so a global `git config beads.role` on the machine makes it fail. They are environment-local, not pre-existing breakage on `main` — worth restating that way so the note does not read as "main is red".

One thing to address before merge, caught in a second-model cross-check of this review:

**The nocgo failure mode flips from always-warn to possibly-false-green.** The deleted stubs returned `StatusWarning` ("Skipped: requires CGO") for every check, so a nocgo `bd doctor --check=validate` could never exit 0 — useless, but never misleading. After this PR, the real implementations run, and when no database is reachable they return `StatusOK` with "N/A (no database)" (`validation.go:40-45`, `:121-126`) — so a nocgo build with an unreachable Dolt server now reports a passing validate having executed zero integrity checks. That is parity with what the CGO build does in the same situation, and #4977 tracks the `StatusSkipped` idea separately, but this PR is the moment the always-warn guard disappears, so it should carry the mitigation: either return a warning/skip status for the no-database branches, or (minimally) an exit-code test pinning the intended nocgo no-DB behavior plus a changelog line. Happy to absorb this as a maintainer commit on your branch if you prefer.

Non-blocking follow-ups:

1. The PR title starts with `# `. A squash merge takes the PR title as the commit subject, so it would land as `# fix(doctor): ...`. The same artifact is on all eight PRs you opened in that batch (5128-5134); a title edit fixes it. The commit subject also uses a `GH`-prefixed #4977 where a bare `#4977` autolinks (the body's `Fixes #4977` is right).
2. `.github/workflows/pr.yml:102-107` compiles pure-Go *test* binaries only for `./cmd/bd`, `./internal/storage/embeddeddolt` and `./internal/tracker`. `go build ./cmd/bd` covers the doctor package's non-test files transitively, so this change is guarded at the level that matters — but nothing compiles `./cmd/bd/doctor`'s test files under `CGO_ENABLED=0`, which is precisely the contamination class that job's comment describes, and this PR enlarges doctor's untagged surface. Adding `go test -tags gms_pure_go -c -o /tmp/bd-doctor-puregeo-test ./cmd/bd/doctor` there (and to the `main.yml:138-144` mirror) would lock the win in. It passes today.
3. `cmd/bd/doctor/{validation,maintenance,migration_validation}_test.go` are still `//go:build cgo` because they need the cgo-tagged `test_helpers_test.go`. That is fine for this PR — the code under test is identical in both modes — but it means the newly-enabled functions have no test presence in the build configuration the PR is about. A follow-up issue is the right home for that.

vishnujayvel:
Thanks — the always-warn-to-possibly-false-green point is the one worth fixing before merge: once the stubs are gone, a nocgo build with an unreachable Dolt server now reports a clean `bd doctor --check=validate` having run zero integrity checks, where the old stub at least never claimed success. Good catch on a case that only shows up in exactly the situation doctor exists to diagnose.

You were right to push back on our "pre-existing failures on main" framing for `TestCheckBeadsRole_NotConfigured`/`_NotGitRepo` — they're environment-local, not main-is-red. Our dev machine has a global `git config beads.role=maintainer`; confirmed both fail with that config present and pass (`ok ... doctor 2.060s`) with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`. Harness now neutralizes global git config so this can't recur.

On the title: BSD sed's lack of `\+` support broke our leading-`# ` strip across this batch (5128-5134); script now asserts against it.

Please take the false-green mitigation (warning/skip status or an exit-code test pinning the no-DB behavior) and the two CI test-binary compile additions as maintainer commits.

Issue #4860, opened 2026-07-17
Dolt-backed tests in cmd/bd/doctor(/fix) have no PR-CI execution path, and several destructive --fix functions additionally lack fix-application tests


vishnujayvel:
## Summary

While adding a fix-application test for `OrphanedChildCounters` in #4858, we found two stacked, independent gaps:

1. **No PR-triggered CI job actually executes `cmd/bd/doctor`'s or `cmd/bd/doctor/fix`'s Dolt-backed tests today** — not because Docker is unavailable in a given environment, but because of two structural issues in the CI config itself (details below). This includes the orphan-check family (`TestCheckOrphanedChildCountersDB_*`, `TestCheckOrphanedDependenciesDB_*`) and everything in `cmd/bd/doctor/fix`.
2. **Independent of (1):** several destructive `--fix` functions in `cmd/bd/doctor/fix/*.go` only have a smoke test asserting they error on a missing `.beads/` dir — nothing that seeds bad data, runs the fix, and checks exactly what got deleted.

Both are fixable incrementally; we'd like to help with both, starting with what's in #4858.

## Gap 1: PR CI never runs cmd/bd/doctor(/fix)'s Dolt-backed tests

Two independent mechanisms combine to exclude this coverage from every `pull_request`-triggered workflow:

**a) Skip-flag gap.** `scripts/ci/lib/test-env.sh`'s `beads_test_env_enter()` unconditionally calls `beads_test_env_add_skip "dolt"` (setting `BEADS_TEST_SKIP=dolt`) unless `BEADS_TEST_ENV_RUN_DOLT=1` is set. We couldn't find anywhere in the repo — workflows, scripts, or Makefile — that ever sets `BEADS_TEST_ENV_RUN_DOLT=1`; the escape hatch appears to be dead code. `pr.yml`'s "PR Core" job (`make ci-pr-core` → `scripts/ci/pr-core.sh`) is the only PR-triggered job that runs `go test ... ./...` (i.e., the only one whose package scope even includes `cmd/bd/doctor/fix`), and it goes through `beads_test_env_enter`, so `BEADS_TEST_SKIP=dolt` is always set there. Both `testutil.RequireDoltBinary` (`internal/testutil/testdoltcommon.go`) and the Docker-detecting `newTestDoltStore`/`newFixTestStore` family honor that var and skip — regardless of whether Docker or a local `dolt` binary is actually available.

[...]

Issue #3087, opened 2026-04-07
cmd/bd/doctor: TestCheckTestPollution_NoTestIssues_EmptyDB fails when BEADS_DOLT_PORT is set in dev shell


A3Ackerman:
## Summary

`TestCheckTestPollution_NoTestIssues_EmptyDB` (cmd/bd/doctor/validation_test.go:541) fails on dev machines that export `BEADS_DOLT_PORT` for unrelated reasons. The test gating in `doctorTestServerPort()` reads the env var directly, so it cannot tell the difference between "TestMain started a test container and exported BEADS_DOLT_PORT" and "an external shell exported BEADS_DOLT_PORT pointing at a different Dolt server."

In our case the env var is exported at login by Gas Town's Dolt daemon (`BEADS_DOLT_PORT=3307`), so the test runs against the dev's local Dolt server instead of skipping. Same gating pattern exists in `cmd/bd/doctor/fix/validation_test.go` (`fixTestServerPort()`), so the same class of failure can hit `cmd/bd/doctor/fix` tests too.

## Repro

On macOS (no Docker required), against `main` (verified at b84ab14):

```sh
BEADS_DOLT_PORT=3307 \
CGO_ENABLED=1 \
CGO_CFLAGS=\"-I\$(brew --prefix icu4c)/include\" \
CGO_CPPFLAGS=\"-I\$(brew --prefix icu4c)/include\" \
CGO_CXXFLAGS=\"-I\$(brew --prefix icu4c)/include\" \
CGO_LDFLAGS=\"-L\$(brew --prefix icu4c)/lib\" \
go test -count=1 -run TestCheckTestPollution_NoTestIssues_EmptyDB ./cmd/bd/doctor/...
```

(Anything that speaks the MySQL protocol on the configured port will reproduce; ours is a dolt sql-server running on 3307 with no \`bd_test_*\` databases.)

Output:

```
--- FAIL: TestCheckTestPollution_NoTestIssues_EmptyDB (0.00s)
    validation_test.go:555: Status = \"ok\", want \"warning\"
    validation_test.go:558: Message = \"N/A (no database)\", want \"N/A (query failed)\"
FAIL
FAIL    github.com/steveyegge/beads/cmd/bd/doctor    1.056s
```

With \`BEADS_DOLT_PORT\` unset, the test correctly skips (per the \`doctorTestServerPort() == 0\` guard at validation_test.go:542) and the suite passes.

## Root cause

[...]

PR #5133, opened 2026-07-29
ci(pr): run `cmd/bd/doctor` Dolt tests on every PR

vishnujayvel:
## What

Adds an advisory `test-doctor-dolt` job to `pr.yml` that runs the parent `cmd/bd/doctor` package against a live Dolt server (CLI installed, sql-server image pulled) on every PR, invoked the same way the required storage job already runs `cmd/bd/doctor/fix`. Two tests that fail only because a server is present sit on an explicit `-skip` list, so a red step means a new failure; two more that failed the same way are fixed here.

| File | What it does now |
|---|---|
| `.github/workflows/pr.yml` | New `test-doctor-dolt` job, outside `ci-gate`: every step that runs code carries `continue-on-error` and `timeout-minutes`; the test step sets `BEADS_DOCTOR_REQUIRE_DOLT=1` and skips the two listed tests by exact name; two warning steps annotate a failed or skipped run |
| `cmd/bd/doctor/dolt_e2e_test.go` | `TestMain` exits 1 instead of warning when the Dolt container is unavailable and `BEADS_DOCTOR_REQUIRE_DOLT=1` |
| `cmd/bd/doctor/dolt_test.go`, `fresh_clone_server_test.go`, `validation_test.go` | `clearDoltPortEnv` helper; the two "no server" tests clear `BEADS_DOLT_SERVER_PORT`/`BEADS_DOLT_PORT` so the package test server no longer answers them; `validation_test.go`'s existing inline copy of that loop now calls the helper |
| `scripts/pull_dolt_image_test.go` | Expected `pull-dolt-image.sh` count in `pr.yml` 2 → 3; new meta-test fails if either `-skip` name stops existing as a test in `cmd/bd/doctor` |
| `scripts/ci/lib/test-env.sh` | Comment only: why the Dolt jobs call `go test` directly instead of `beads_test_env_enter` |

## Why

Addresses #4860 (gap 1a). pr-core's `beads_test_env_enter` sets `BEADS_TEST_SKIP=dolt`, and nothing in the repo ever set the `BEADS_TEST_ENV_RUN_DOLT` opt-out, so the Dolt-backed `cmd/bd/doctor` suites never ran on a PR. `cmd/bd/doctor/fix` has since gained a required job; this gives the parent package its own PR path without turning the required gate red while its two remaining server-dependent tests are sorted out.

Type: ci · Risk: low — the job is outside `ci-gate` and cannot produce a red check run; the Go changes are test-only.

## Verification

- `python3 -c 'import yaml; yaml.safe_load(open(".github/workflows/pr.yml"))'` parses; `grep -cE '^    continue-on-error' .github/workflows/pr.yml` → `0` (no job-level `continue-on-error` left in the file).
- `go test -tags gms_pure_go -count=1 ./scripts/` → `ok` (workflow meta-tests, including the `pull-dolt-image.sh` count and the new `-skip` name guard, which fails on a renamed entry in a scratch copy of `pr.yml`).
- The two fixed tests, run against a throwaway local `dolt sql-server` with `BEADS_DOLT_SERVER_PORT` pointed at it: at e89256764 both fail with the same messages CI shows ("Connected successfully", "Database exists on server"); at this head both pass, and still pass with no server and no env override (commands under the fold).
- `go test -tags gms_pure_go ./cmd/bd/doctor/ -list 'Test.*' | grep -cE '^(TestCheckRepoFingerprint_UsesTargetRepoOutsideCWD|TestCheckTestPollution_NoTestIssues_EmptyDB)$'` → `2`: the skip regex matches the two listed names and nothing else. Whole package under `BEADS_TEST_SKIP=dolt` → `PASS`.
- `gofmt -l cmd/bd/doctor` empty; `go vet -tags gms_pure_go ./cmd/bd/doctor/` exit 0. `actionlint` is not available locally, so that check comes from CI.

<details>
<summary>Design notes and review trail</summary>

**Why advisory at the step level, not the job level.** A job-level `continue-on-error` job still reports its check run as FAILURE in the PR rollup (only the workflow-run conclusion turns success), and `scripts/pr-preflight.sh:371-386` and the merge patrol gate on per-check-run conclusions, so a job-level red here would stall the merge lane repo-wide while claiming to be optional. A job-level `timeout-minutes` has the same problem from the other side: it CANCELS the job, and a cancelled advisory job is a red check run too. So the job has neither; every step that runs code carries its own `continue-on-error` and `timeout-minutes` (the `build-examples` precedent at `pr.yml:88-99`), the test step is gated on the setup steps' outcomes, and two warning steps (plain echoes, which carry neither) turn a failed or skipped run into annotations, using the same `steps.<id>.outcome != 'success'` form as `build-examples`. The check run always concludes success. If `build-artifacts` itself fails, `needs` skips the whole job with no annotation; that is fine because SKIPPED is not a blocking rollup conclusion.

**Why a new `BEADS_DOCTOR_REQUIRE_DOLT` variable.** The earlier `BEADS_FIX_REQUIRE_DOLT: "1"` on this job was inert: its only reader is `cmd/bd/doctor/fix/validation_test.go:43`, and the fix package is no longer part of this job (it is covered by the required "Test (storage domain + uow)" job). The parent package's `TestMain` only warned when the container was missing, so a Dolt-less runner would have quietly dropped every server-backed test and reported green. `cmd/bd/doctor/dolt_e2e_test.go:64-75` now reads `BEADS_DOCTOR_REQUIRE_DOLT=1` under the same contract as the fix package's variable: container unavailable → package fails. Locally without the variable it still only warns, so contributors without Docker keep the non-Dolt tests. Checked both directions under `BEADS_TEST_SKIP=dolt`: with the variable FATAL + `FAIL`, without it WARN + `ok`.

**The expected-failure list.** Running the package against a live server for the first time surfaced four failures (two found by @maphew's first review, two more once the job stopped using the hermetic wrapper). All four are the package test server answering tests that assume no server, not product bugs:

- `TestRunDoltHealthChecks_DoltBackendNoServer` and `TestCheckFreshClone_ServerModeUnreachable` reached the server because `TestMain` exports `BEADS_DOLT_SERVER_PORT` and that variable is the first stop in `doltserver.DefaultConfig`'s port chain, ahead of the `metadata.json` port the second test sets to 1. Fixed here: both clear the override for their duration (`clearDoltPortEnv`, `cmd/bd/doctor/dolt_test.go`), the pattern `TestCheckTestPollution_NoTestIssues_NoServer` in the same package already used inline; that test now calls the helper too.
- `TestCheckRepoFingerprint_UsesTargetRepoOutsideCWD` wants `Verified (...)` and gets `N/A (no database)` because `CheckRepoFingerprint` stats the local `.beads/dolt` directory (`cmd/bd/doctor/integrity.go:557`), which does not exist when the test's `dolt.New` went to the server. Whether the check should consult the server is a product decision, not a test tweak.
- `TestCheckTestPollution_NoTestIssues_EmptyDB` wants `N/A (query failed)` and gets a successful query because the database it resolves to on the shared server already has an issues table. Making it deterministic means giving it its own database name under the test-server firewall, which is a design choice for the package, not this PR.

The last two are on the test step's `-skip` list, anchored on exact names with a one-line reason each next to the flag. `-skip` was chosen over an allowlist or a post-hoc log filter because it is one line, `go test` enforces it, and a green step then means "no new failure" rather than "the same four failed again". Removing the two entries is the promotion checklist. Because `-skip` matches nothing, silently, once a listed test is renamed or deleted, `TestDoctorDoltLaneSkipListNamesRealTests` (`scripts/pull_dolt_image_test.go`) parses the list out of `pr.yml` and fails unless each name still exists as a `func` in `cmd/bd/doctor/*_test.go`.

**What the lane gates.** Nothing. It is not in `ci-gate`'s `needs` or `CI_GATE_REQUIRED`, and a failure appears only as a warning annotation on the run. `main.yml` carries no mirror of this job on purpose; pre-merge coverage is the point. To promote it: drop every step's `continue-on-error` and the two warning steps, empty the `-skip` list, then add `test-doctor-dolt` to `ci-gate`'s `needs`, `CI_GATE_REQUIRED`, and its env map.

**Local reproduction of the two fixed tests** (macOS, ICU CGO flags, `-tags gms_pure_go`):

```
# throwaway server: dolt init in ./beads, then
dolt sql-server --host 127.0.0.1 --port 3399
# in cmd/bd/doctor, test binary built at each sha
BEADS_TEST_SKIP=dolt BEADS_DOLT_PORT=3399 BEADS_DOLT_SERVER_PORT=3399 ./doctor.test -test.count=1 -test.v \
  -test.run '^(TestRunDoltHealthChecks_DoltBackendNoServer|TestCheckFreshClone_ServerModeUnreachable)$'
# e89256764:
#   dolt_test.go:53: checks[0] (Dolt Connection): expected skip message about no server, got "Connected successfully"
#   --- FAIL: TestRunDoltHealthChecks_DoltBackendNoServer (0.03s)
#   fresh_clone_server_test.go:154: expected "warning" on unreachable server, got "ok" (message: Database exists on server)
#   --- FAIL: TestCheckFreshClone_ServerModeUnreachable (0.00s)
# this head, same env:            --- PASS / --- PASS
# this head, no env, no server:   --- PASS / --- PASS
```

**Review trail.** Round 1 (@maphew): the job was wired as required while red on its own run; the hermetic wrapper's isolated HOME/gitconfig was suspected in the failures; action pins diverged from the tree's. The job was moved out of `ci-gate`, the wrapper was dropped in favour of plain `go test` (which is how the required storage job runs the fix package), and the pins were aligned. Round 2 (@bee-ghosttrack): job-level `continue-on-error` (blocker), inert `BEADS_FIX_REQUIRE_DOLT` (major), four permanent failures with no expected-failure list (major), checkout pin and missing job timeout (minor). All addressed as described above; the job-level timeout stays absent on purpose, with per-step timeouts instead. A final polish commit aligned the skipped-run warning with the `build-examples` form, deduplicated the env-clearing loop, and added the `-skip` name guard.

Out of scope (issue gap 2): fix-application tests for the destructive `--fix` helpers, which still have only smoke coverage.

</details>


maphew (commented):
The diagnosis here is right, and it's a real gap that was easy to miss. I verified each link in the chain independently:

- `scripts/ci/lib/test-env.sh:47` adds `dolt` to `BEADS_TEST_SKIP` unless `BEADS_TEST_ENV_RUN_DOLT=1`, and `git grep BEADS_TEST_ENV_RUN_DOLT` on `main` finds it only in `test-env.sh` itself and a mention in `scripts/README.md:27` — nothing ever sets it, so the escape hatch was dead.
- `internal/testutil/testdoltcommon.go:21` (`RequireDoltBinary`) and `internal/testutil/testdoltserver.go:100` (`checkDolt` → `hasTestSkip`) both honor that var, so both the binary-gated and container-gated doctor tests skipped under `pr-core`.
- The job itself is modelled on `test-domain-uow` (pr.yml:463) including the image pull, and the `ci-gate` wiring is complete in all three required places (`needs`, `CI_GATE_REQUIRED`, and the env mapping) — that's the part people usually get partially wrong.

One blocker, and it's the empirical kind rather than a design objection.

**The new job is red on this PR's own run, and it's wired as required.** [Run 30421798687, job 90481187042](https://github.com/gastownhall/beads/actions/runs/30421798687/job/90481187042):

```
--- FAIL: TestCheckRepoFingerprint_UsesTargetRepoOutsideCWD (2.65s)
    integrity_path_cgo_test.go:57: Message = "N/A (no database)", want "Verified (bda9a6ee)"
--- FAIL: TestCheckTestPollution_NoTestIssues_EmptyDB (0.00s)
    validation_test.go:654: Status = "ok", want "warning"
    validation_test.go:657: Message = "No test pollution detected", want "N/A (query failed)"
FAIL	github.com/steveyegge/beads/cmd/bd/doctor	33.109s
ok  	github.com/steveyegge/beads/cmd/bd/doctor/fix	14.418s
```

This is arguably the PR working as designed — those are exactly the latent expectations that rot when a suite never runs — but because the job is in `ci-gate`'s `needs` and `CI_GATE_REQUIRED`, merging as-is turns the required gate red for every PR in the repo. Two ways out, either is fine by me:

1. Fix the two tests in this PR (they look like stale expectations, not product bugs), or
2. Land the job without the three `ci-gate` hunks, then promote it to required in a one-line follow-up once it's green.

Two smaller things, one of which may be connected to the failures:

**Consider dropping the hermetic-env dance entirely.** The sibling job this is modelled on, `test-domain-uow`, never calls `beads_test_env_enter` — it just runs `go test` with Dolt installed and the image pulled, and that is why it doesn't need an opt-out flag. If this job did the same, the `source`, `beads_test_env_enter`, `BEADS_TEST_ENV_RUN_DOLT: "1"` and `BEADS_TEST_SKIP: ""` lines all go away and the job gets shorter than what's here. That may also be load-bearing for the failures above: `beads_test_env_enter` repoints `HOME` and `DOLT_ROOT_PATH` and sets `GIT_CONFIG_GLOBAL` to a freshly-created empty file (test-env.sh:38-48), which silently discards the `git config --global user.name/user.email` written by the preceding "Configure Git and Dolt identity" step. Worth checking whether the two failures reproduce without the hermetic env before deciding where the fix belongs.

**Action pins diverge from the repo's.** The new job uses `actions/checkout@9c091bb2...` and `actions/setup-go@924ae3a1...` (setup-go 6.5.0), while every other step in the tree uses `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v6` (96 uses) and `actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0` (53 uses). The SHAs resolve to real commits, so this isn't a security issue, but the split pin breaks the uniformity dependabot and pin audits rely on. `download-artifact` already matches.

Minor, no action needed unless you want to: no `timeout-minutes` on the job (runtime was ~3.5 min, so cost is genuinely a non-issue here — good outcome, that was my first worry); and `main.yml` carries mirrored copies of these jobs for pushes to `main`, so post-merge runs still won't execute the doctor Dolt suites — probably fine since pre-merge gating is the point, but say so in the PR body if it's deliberate.

The `test-env.sh` comment change is accurate and worth keeping regardless of how the rest lands.

vishnujayvel:
Thanks for the thorough verification — going with option 2: landed the job advisory rather than fixing the two tests here.

Pushed a follow-up commit (`2af7875f3`) that drops `test-doctor-dolt` from `ci-gate`'s `needs`, from `CI_GATE_REQUIRED`, and from the env mapping. The job itself is unchanged — it still has no tier `if:` gate, so it still installs Dolt, pulls the image, and runs `./cmd/bd/doctor/ ./cmd/bd/doctor/fix/` unconditionally on every PR and reports its own check. It just no longer feeds into the aggregate gate that `.github/scripts/ci-gate.sh` computes, so a failure there can't turn `CI Gate / Required` red anymore.

I confirmed this is the right lever before touching it: `gastownhall/beads` has no classic branch protection on `main` (404 from the branches/protection API) and the one active ruleset that applies to `main` ("Protect main - light") only carries `deletion` and `non_fast_forward` rules — no `required_status_checks` rule. So "required" here is entirely a workflow-level construct: `ci-gate`'s `needs:` list plus the `CI_GATE_REQUIRED` env var that `.github/scripts/ci-gate.sh` reads and fails on any non-`success` (or disallowed-`skipped`) result. Removing `test-doctor-dolt` from those two places is sufficient and complete — there's no separate ruleset-level required-check list to also edit.

Left the two failing tests alone, as agreed — `TestCheckRepoFingerprint_UsesTargetRepoOutsideCWD` and `TestCheckTestPollution_NoTestIssues_EmptyDB` are real, pre-existing gaps this job correctly surfaced, but fixing them is separate follow-up work, not part of getting the CI job itself landed. Once both are fixed and the job is green for a run or two, `test-doctor-dolt` should go back into `ci-gate`'s `needs`/`CI_GATE_REQUIRED` — that's the promotion path you suggested.

I didn't touch the hermetic-env question (`beads_test_env_enter` vs. going bare like `test-domain-uow`) or the action-pin divergence (`checkout@9c091bb2...`/`setup-go@924ae3a1...` vs. the repo's usual `checkout@3d3c42e5...`/`setup-go@b7ad1dad...`) — those are worth a look but are pre-existing in the job as originally written, not something advisory-vs-gating touches, and I don't want to conflate two separate review threads in one commit.

PR body updated to stop claiming the job blocks merge (it previously said "wired into `ci-gate`'s required list, so it actually blocks merge" — no longer accurate).

I did not run this on GitHub's runners — no way to do that locally. What I verified: the YAML still parses (`python3 -c "import yaml; yaml.safe_load(...)"`), the diff is the three removals described above plus a three-line note above the job saying why it is deliberately absent from the gate (kept short on purpose — a future contributor adding jobs would otherwise read the omission as an oversight), and the patches for both commits apply cleanly against `origin/main`'s merge-base in a scratch worktree via `git am`. What still needs a real CI run to confirm: that `test-doctor-dolt` failing does not, in practice, turn `CI Gate / Required` red on the next push — I'm confident in the mechanism from reading `ci-gate.sh` and the ruleset, but I haven't watched it happen.

vishnujayvel:
Follow-up on my own branch — this PR as it stood didn't actually deliver what it claims, and
I'd rather flag that than have you find it.

Taking `test-doctor-dolt` out of `ci-gate`'s `needs:` and out of `CI_GATE_REQUIRED` makes the
**required check** green, but it doesn't make the **workflow run** green. The job has no
`continue-on-error` — there are in fact zero occurrences of that key anywhere in `pr.yml` — so
its own conclusion is still `failure`, and a run with any failing job concludes `failure`
regardless of what branch protection considers required. Net effect: the gate passes and the PR
still shows red on every single PR, which is the thing the change was supposed to stop.

`144f11c85` adds `continue-on-error: true` to the job. The gate exclusions are untouched, and
`ci-gate.sh` is unaffected — it only reads `needs.<job>.result` for the names listed in
`CI_GATE_REQUIRED`, which this job is deliberately absent from, so it never looks at this job's
result at all.

There's a second-order effect worth knowing about separately, because it's been live on `main`
for as long as any job in `pr.yml` has been failing. `docs-autofix.yml` triggers on
`workflow_run` of the `PR` workflow with `types: [completed]` — any conclusion — and its
`autofix` job is gated on:

```yaml
if: >
  github.event.workflow_run.event == 'pull_request' &&
  github.event.workflow_run.conclusion == 'failure'
```

So while the `PR` workflow concludes `failure` on every PR, that job's steps execute on every
PR too, not just on real docs drift. It isn't destructive — it exits early at
`docs-autofix.yml:61-64` when there's no `cli-docs-freshness-patch` artifact — but it's a
checkout plus a `gh api` call per PR for nothing. The `continue-on-error` fix removes the
artificial `failure` conclusion and so resolves this too, without touching `docs-autofix.yml`.

**What I could not verify:** none of this has been through a real runner from my side — I have
no way to trigger the `PR` workflow on your infrastructure. What I did check locally is that
`pr.yml` still parses, that `continue-on-error` is at job level rather than step level, and that
`.github/scripts/ci-gate.sh` doesn't consume job conclusions in a way this would change. The
claim that a failing non-required job reddens the run is from Actions' documented behaviour plus
the structure of the file, not from an observed run. If you'd rather see it green before merging,
this branch is the place to prove it.

Happy to go the other way instead if you'd prefer — fixing or skipping the two failures behind
#5133 so the job is genuinely green, and then promoting it straight back into the gate. That's
strictly better than a permanently non-blocking advisory job; it's just a larger change than
this PR was scoped for, and it needs a decision from you about those two tests rather than from
me.

maphew (approved):
Reviewed at `144f11c85d4`.

No correctness findings. The advisory doctor/Dolt lane explicitly opts out of the shared Dolt skip, clears inherited `BEADS_TEST_SKIP`, installs both required Dolt surfaces, remains outside `ci-gate`, and documents why it is temporarily non-blocking through job-level `continue-on-error`.

Verification:

- `bash -n scripts/ci/lib/test-env.sh` — PASS.
- `beads_test_env_enter` with `BEADS_TEST_ENV_RUN_DOLT=1` — PASS; resulting `BEADS_TEST_SKIP` remained empty.
- `actionlint` — unavailable locally.
- Virtual merge and blocking upstream preflight — branch conflicts with current main and has three failed/action-required checks.

Disposition: **approve the change; merge only after rebase/conflict resolution and a clean preflight**.

vishnujayvel:
Thanks for the review, and apologies for the slow reply — this one fell through a gap in my own
tracking rather than being deliberately parked.

On your merge condition: the conflict half is resolved, the preflight half is not yet.

- **Rebase/conflicts** — done. The branch is `MERGEABLE` against current `main` at `a81a620e3`.
- **Clean preflight** — not yet. The most recent run is from 2026-08-06 and is now stale against a
  base that has moved considerably. In it, the advisory `Test (doctor Dolt + fix)` lane this PR adds
  reported failures, but the dominant signature is the test container being unreachable —
  `dial tcp 127.0.0.1:32769: connect: connection refused` — cascading into
  `TestDestructiveFix_*`, `TestVerifyFixTargetIdentity/*` and `TestChildParentDependencies_*`.
  Two others (`TestCheckRepoFingerprint_UsesTargetRepoOutsideCWD`,
  `TestCheckTestPollution_NoTestIssues_EmptyDB`) look independent of the container.

I don't want to claim that's purely harness noise on the strength of a 13-day-old log, so I'll
rebase onto current `main` and re-run before asking you to merge. If the container-reachability
failures persist on a fresh run, that's a finding about the lane's setup and belongs in this PR —
the point of the lane is to catch exactly this, and it would be self-defeating to land it while it
cannot reliably reach Dolt.

I'll report back with the fresh run rather than leave you to poll it.

bee-ghosttrack:
CI triage (bee, maintainer seat): the macOS red is the repo's CI-workflow meta-test rejecting the new `test-doctor-dolt` job — `ci_workflow_test.go:444`: actions/setup-go must be pinned to the released SHA (`b7ad1dad…`, not `924ae3a…`) and must set `cache: false`. Fix the job to match the other pr.yml jobs and it should go green.

vishnujayvel:
Thanks, that is exactly what the meta-tests are complaining about. Both the macOS and PR Core reds come from scripts/ tests on pr.yml: the setup-go SHA and cache flag, and the retry-helper count, which is 3 now because the new job also calls pull-dolt-image.sh.
I pinned actions/setup-go to the released SHA with cache: false to match the other jobs, bumped the expected retry-helper count to 3 with a comment naming the new job, and pushed. The doctor Dolt lane stays continue-on-error with the four known failures noted in the PR.

bee-ghosttrack (changes requested):
Re-reviewed at e8925676 (post-approval: 0a03145, 542ac93, e8925676).

Verified good:
- setup-go b7ad1dad31e… is exactly actions/setup-go refs/tags/v7.0.0, matching `setupGoSHA` (scripts/ci_workflow_test.go:790) with the required `cache: false`.
- The retry-helper bump is honest, not a count fitted to the failure: `TestDoltImagePullWorkflowsUseRetryHelper` counts literal `run: ./scripts/ci/pull-dolt-image.sh` lines, and pr.yml now has exactly three (:759, :920, :991). The same test's no-raw-docker-pull assertion is now satisfied too. `go test -tags gms_pure_go ./scripts/` is green locally at head.
- scripts/ci/lib/test-env.sh is comment-only; the job is genuinely out of ci-gate's `needs` / `CI_GATE_REQUIRED`.

**BLOCKER** — pr.yml:895 uses job-level `continue-on-error` on a job you expect to stay red. pr.yml:88-99, in this same file, already rules that out: a job-level continue-on-error job still reports its check run as FAILURE, and pr-preflight and the pr-babysit merge patrol gate on per-check-run conclusions. Confirmed: job 97005557351 concluded `failure`, and scripts/pr-preflight.sh:371-386 blocks on any FAILURE rollup entry. Merged as-is, this blocks preflight on every PR in the repo. Use the build-examples pattern: step-level continue-on-error plus timeout-minutes on the step at pr.yml:944, plus a warn step.

**MAJOR** — pr.yml:947 `BEADS_FIX_REQUIRE_DOLT: "1"` is inert. Its only reader is cmd/bd/doctor/fix/validation_test.go:43, and 0a03145 removed that package from the job. cmd/bd/doctor's TestMain (dolt_e2e_test.go:64-65) only warns when the container is missing, so a Dolt-less runner degrades to no coverage and the four known failures vanish — the lane looks healthier. Add a hard-require in the doctor package or drop the var.

**MAJOR** — four permanent failures with no expected-failure list means the lane's red/green is constant; a real doctor regression is indistinguishable from the known four. They look like package-wide pollution from TestMain's shared server (dolt_e2e_test.go:59-65); fixing them would let this go straight to required.

MINOR — pr.yml:897 pins checkout 9c091bb2…; every other site uses 3d3c42e5….
MINOR — no `timeout-minutes` on the job; only `go test -timeout 25m`.

CI at e8925676 was still in flight when I read it (no failures yet; macOS, PR Core and doctor Dolt still running), so the meta-test fix is not yet CI-confirmed.

REQUEST-CHANGES

vishnujayvel:
Thanks for the careful re-read. Reworked in three commits on e89256764 (head is b11feed6a); per finding:

- **BLOCKER, job-level `continue-on-error`**: agreed, that would have been a FAILURE rollup entry that `pr-preflight.sh` blocks on. The job now follows the build-examples shape: nothing advisory at the job level; every step that runs code carries `continue-on-error` and `timeout-minutes` (pr.yml:910-970), the test step is gated on the setup steps' outcomes (pr.yml:1001-1009), and two warning steps annotate a failed or skipped run (pr.yml:1017-1033), gated the same way `build-examples` gates its own. The check run always concludes success.
- **MAJOR, inert `BEADS_FIX_REQUIRE_DOLT`**: agreed. Replaced with `BEADS_DOCTOR_REQUIRE_DOLT`, which the parent package's TestMain actually reads: container unavailable + var set → exit 1 (cmd/bd/doctor/dolt_e2e_test.go:64-75, same contract as the fix package's var); the lane sets it (pr.yml:1014). Checked both directions locally under `BEADS_TEST_SKIP=dolt`: FATAL + `FAIL` with the var, WARN + `ok` without.
- **MAJOR, four permanent failures**: two fixed, two on an explicit expected-failure list. `TestRunDoltHealthChecks_DoltBackendNoServer` and `TestCheckFreshClone_ServerModeUnreachable` were reaching the package server because TestMain exports `BEADS_DOLT_SERVER_PORT`, the first stop in `doltserver.DefaultConfig`'s port chain; they now clear it for their duration (`clearDoltPortEnv`, cmd/bd/doctor/dolt_test.go:23-31 and :52, fresh_clone_server_test.go:127; validation_test.go:614 now calls it instead of its own copy of the loop). Reproduced against a throwaway `dolt sql-server` with that variable pointed at it: both fail at e89256764 with CI's exact messages, pass after, and still pass with no server. The other two (`CheckRepoFingerprint` stats a local dolt directory that a server-backed store never creates, integrity.go:557; `CheckTestPollution` finds an issues table already on the shared server) need a design call rather than a test tweak, so the test step `-skip`s exactly those two by name with a reason each (pr.yml:989-998, :1015). Green now means "no new failure", and the two entries are the promotion checklist. A meta-test (scripts/pull_dolt_image_test.go:95) parses that list out of pr.yml and fails if either name stops existing in cmd/bd/doctor, since `-skip` would otherwise match nothing without a word.
- **MINOR, checkout pin**: aligned to `3d3c42e5…` # v6 (pr.yml:910); every checkout in the file now matches.
- **MINOR, no job `timeout-minutes`**: still none at the job level on purpose (a job timeout cancels, and a cancelled advisory job is a red check run by another door); each step has its own, 30 on the test step above `go test -timeout 25m`.

Verification: `go test -tags gms_pure_go ./scripts/` passes (and the new meta-test fails on a renamed skip entry in a scratch copy of pr.yml); pr.yml parses with `yaml.safe_load` and has no job-level `continue-on-error` left; `gofmt -l` clean and `go vet -tags gms_pure_go ./cmd/bd/doctor/` exit 0; the skip regex matches exactly the two names in `go test -list`. `actionlint` is not on this machine, so that check comes from CI.

bee-ghosttrack (approved):
**APPROVE** (round 2 @ b11feed6a, adversarial re-verify in a worktree; `actionlint` clean on head and base; the parent doctor package's pure-Go tests pass with a neutral gitconfig).

The BLOCKER is closed by live proof, not argument. `grep -cE '^    continue-on-error' pr.yml` is 0; all nine are step-level; the job carries neither `continue-on-error` nor `timeout-minutes`, the build-examples shape at pr.yml:104-152. The actual check run, job 101134840246 in run 33906911865, concluded SUCCESS and the PR rollup shows zero FAILURE / CANCELLED / TIMED_OUT / ACTION_REQUIRED entries, which is exactly the set scripts/pr-preflight.sh:373-376 blocks on. The step gating uses `.outcome`, the only selector that is not inert under `continue-on-error`, and the CI log proves the semantics: both warn steps reported `skipped` because every outcome was `success`. Leaving out a job-level `timeout-minutes` is right, not an oversight: a job timeout CANCELS, and CANCELLED is in preflight's blocking set, so it would reopen the blocker by another door.

The rest of round 1 is closed too. The inert `BEADS_FIX_REQUIRE_DOLT` is now `BEADS_DOCTOR_REQUIRE_DOLT`, read by the parent package's TestMain (dolt_e2e_test.go:64-75) and fail-closed in both directions (an inherited `BEADS_TEST_SKIP=dolt` also hard-fails through `checkDolt()`). Of the four permanent failures, two are really fixed: the prior run at 542ac93f7 failed them with the exact messages you quote and they PASS at head against a live container; the two on the `-skip` list skip for the causes you state, which I confirmed against that same prior log (`want "Verified (4779616a)"`, and `got "No test pollution detected"`, i.e. the query succeeded, a different mode from #3087, so `clearDoltPortEnv` is correctly not applied there). The lane really ran: image 2.2.0 pulled, "Server ready", 825 PASS / 9 SKIP / 0 FAIL in 29.7s. The new guard meta-test discriminates a renamed skip entry (broken on purpose, it fails). Pins are uniform, one SHA each.

NITs, none blocking: the promotion checklist omits that emptying the `-skip` list fails `TestDoctorDoltLaneSkipListNamesRealTests` in the gated `./scripts/` package (rc 1); that guard is build-tag-blind while both listed tests are `//go:build cgo` and the step does not pin `CGO_ENABLED`, so a future sweep setting it to 0 would drop the coverage and still report green; the eight-clause setup list is duplicated three times; the guard only checks the first regex match.

Ready for Julian.

bee-ghosttrack:
Re-verified at b11feed6a; my 2026-09-04 approval stands.

CI at head: 119 success, 2 skipped, 0 failure. The "Test (doctor Dolt)" check run (job 101134840246) concluded success in 4m08s.

Base has moved since the merge-base 5cbe3a29a: 5 commits on pr.yml, 13 under scripts/. I merged current main into your head in a scratch worktree: clean merge, actionlint rc 0, `go test -tags gms_pure_go ./scripts/` green, pr.yml still at exactly 3 pull-dolt-image.sh call sites. Main's new meta-tests (complexity-report, proxied-local-smoke) don't touch this job.

Two non-blocking MINORs:

1. The test step doesn't pin CGO_ENABLED, and `-race` does not force cgo on under Go 1.26. Measured at head: `CGO_ENABLED=0 BEADS_DOCTOR_REQUIRE_DOLT=1 go test -race ... ./cmd/bd/doctor/` -> ok in 10.9s, no container; at CGO_ENABLED=1, ok in 62.2s with one. TestMain and both -skip names are `//go:build cgo`, so at CGO_ENABLED=0 the guard, the server and the skip list all vanish and the lane still reports green. Add `CGO_ENABLED: "1"` to the step env.

2. TestDoctorDoltLaneSkipListNamesRealTests checks that a listed name exists, not that it still needs skipping: appending an already-passing test keeps ./scripts/ green. And emptying the list, your documented promotion step, fails that test. Worth a line in the promotion comment.

Good to merge; re-run checks against current main first.

PR #5134, opened 2026-07-29
fix(winget): require PortableCommandAlias for GasTownHall.Beads

### LINKED ISSUE #4908: winget manifest: missing PortableCommandAlias — no bd.exe symlink in WinGet\Links, bd invisible to already-running shells
## Problem

On Windows, installing via winget never puts a `bd` command anywhere that already-running processes can see.

`winget install --id GasTownHall.Beads` extracts the portable binary to
`%LOCALAPPDATA%\Microsoft\WinGet\Packages\GasTownHall.Beads_<suffix>\bd.exe` and appends **that package folder** to the user PATH. It does **not** create the usual command symlink in `%LOCALAPPDATA%\Microsoft\WinGet\Links`.

A PATH entry is only inherited by processes started *after* the install. Shells, editor terminals, agent sessions, hooks and services that were already running never pick it up. The user-visible symptom is *"bd worked at the start of my session and disappeared halfway through"* — which sends people looking for a bd bug that isn't there.

## Root cause

`manifests/g/GasTownHall/Beads/1.1.0/GasTownHall.Beads.installer.yaml`:

```yaml
InstallerType: zip
NestedInstallerFiles:
- RelativeFilePath: bd.exe
Commands:
- bd
```

`Commands:` is search metadata (it powers `winget search --command bd`); it does not create anything on disk. For portable installers winget creates the `Links` alias from **`PortableCommandAlias`**, which this manifest never sets.

winget's own install log confirms it never even attempts a symlink — just the portable move and the PATH entry:

```
<I> [CLI ] Moving portable exe to: ...\WinGet\Packages\GasTownHall.Beads_...\bd.exe
```

For contrast, a package that does set the alias has a real symlink in the same folder:

```
ayugram.exe   SymbolicLink -> ...\WinGet\Packages\RadolynLabs.AyuGramDesktop_...\AyuGram.exe
```

## Why this bites harder than it looks

The package folder lands late in PATH, so anything providing an earlier `bd` wins. The npm package `@beads/bd` installs an **extensionless `sh` shim** named `bd` next to node — on Windows that shim shadows `bd.exe`, and anything spawning bd through `CreateProcess` (Rust/Go/Node wrappers) fails with:

```
%1 is not a valid Win32 application (os error 193)
```

The trap is that `Get-Command bd` still succeeds — it happily resolves `bd.ps1` — so "is bd on PATH?" checks report a **false positive** while the CLI path is dead. We hit this in a downstream tool that shells out to bd, and diagnosing it took a while precisely because every naive check said bd was fine.

Same class of problem as #3399 (extensionless `bd` leftover shadowing `bd.exe`).

## Proposed fix

Set the alias in the installer manifest, so winget creates `Links\bd.exe` itself:

```yaml
NestedInstallerFiles:
- RelativeFilePath: bd.exe
  PortableCommandAlias: bd
```

`WinGet\Links` is on PATH for essentially every process, including ones started before the install, so this fixes both halves: bd stops vanishing mid-session, and a real `.exe` takes precedence over any stray shim.

Happy to open the PR against microsoft/winget-pkgs if that's welcome — just wanted to confirm the alias is intended before touching the manifest.

## Workaround for anyone hitting this now

Create the symlink manually (needs an elevated shell, or Developer Mode):

```powershell
$target = (Get-ChildItem "$env:LOCALAPPDATA\Microsoft\WinGet\Packages" -Directory -Filter "GasTownHall.Beads_*" |
           Select-Object -First 1).FullName + "\bd.exe"
New-Item -ItemType SymbolicLink -Path "$env:LOCALAPPDATA\Microsoft\WinGet\Links\bd.exe" -Target $target
```

A hardlink works without admin rights but will point at the stale binary after an upgrade, so a symlink is preferable.

## Environment

- winget package `GasTownHall.Beads` 1.1.0, `bd version 1.1.0 (8e4e59d39)`
- Windows 11 Pro 26200, winget portable install (non-elevated), Developer Mode off

vishnujayvel:
## Summary

Fixes #4908.

The published `GasTownHall.Beads` 1.1.0 package on microsoft/winget-pkgs sets only `Commands: [bd]` (search metadata) and never sets `PortableCommandAlias`. For a portable zip installer, winget only creates the `%LOCALAPPDATA%\Microsoft\WinGet\Links\bd.exe` symlink from `PortableCommandAlias` — without it, only the package folder lands on PATH, which is invisible to any shell or agent session that was already running at install time. The in-repo `SteveYegge.beads` manifest already had the alias; the live `GasTownHall.Beads` id did not.

This PR (gastownhall/beads):

1. Adds a canonical `winget/GasTownHall.Beads.installer.yaml` with `PortableCommandAlias: bd`
2. Updates `scripts/update-winget.sh` to regenerate that package on release
3. Documents both package ids and the Links requirement in `winget/README.md`
4. Adds `scripts/check-winget-portable-alias.sh` to guard installer YAMLs against the same regression

The companion fix to the published manifest itself (microsoft/winget-pkgs) is a separate PR against that repo — this PR only changes what beads generates and ships going forward.

Thanks to badigit for the precise diagnosis, including the shim-shadowing failure mode this causes on Windows.

## Test plan

- [x] `./scripts/check-winget-portable-alias.sh` — both installer YAMLs pass
- [x] `go test ./cmd/bd/doctor/ -count=1 -timeout 25m` (icu4c CGO flags, build/tree-health smoke only — this change touches no Go code) — pass. (Correction, 2026-07-29: an earlier version of this test plan reported `TestCheckBeadsRole_NotConfigured` and `TestCheckBeadsRole_NotGitRepo` as failing on unmodified `main`. That was wrong. They are environment-local: our dev machine has a global `git config beads.role=maintainer`, and `cmd/bd/doctor/role_test.go:10-26` asserts a fresh repo has none. Both pass with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`. Upstream `main` is not red. Thanks to @maphew for catching it.)

This change is YAML-only; the Go test run above only confirms the tree still builds and the unrelated test suite is unaffected, it does **not** validate the manifest itself. Separately verified the manifest directly: `winget/GasTownHall.Beads.installer.yaml` parses as valid YAML, and `NestedInstallerFiles` with a `RelativeFilePath` + `PortableCommandAlias` pair under a `zip`/`portable` installer matches winget's documented installer manifest schema for portable command aliasing.


maphew (commented):
The root-cause analysis here is right, and it is the useful part of this PR. I verified it independently against the published manifests: `manifests/g/GasTownHall/Beads/1.1.0/GasTownHall.Beads.installer.yaml` and the newest published `1.1.2` both set only `Commands: [bd]` with no `PortableCommandAlias`, while our in-repo `winget/SteveYegge.beads.installer.yaml:8` has had the alias all along. So the asymmetry you describe is real, still live today, and the `Commands:` vs `PortableCommandAlias` distinction in your README text is accurate. Thanks for chasing this down to the install log — the "`Get-Command bd` succeeds while `CreateProcess` fails" detail in #4908 is the kind of thing that costs someone an afternoon.

Three things need fixing before this can land. The detail below is because this touches the release path, where a silent breakage only shows up months later in a published package.

**1. The new generator emits an unparseable manifest.** `scripts/update-winget.sh:26` (pre-existing, unchanged by this PR) does:

```sh
SHA256=$(curl -sL ".../checksums.txt" | grep windows | awk '{print $1}')
```

`grep windows` matches both `beads_X_windows_amd64.zip` and `beads_X_windows_arm64.zip`, so `$SHA256` holds two newline-separated hashes. I ran the PR branch's `./scripts/update-winget.sh 1.1.2` and the generated `winget/GasTownHall.Beads.installer.yaml` came out as:

```yaml
  - Architecture: x64
    InstallerUrl: .../beads_1.1.2_windows_amd64.zip
    InstallerSha256: 4591B07BF82B3203A1DC7DB17A7E4962D86338E6C3D34A8A857CC11A57F9C159
A4E8D717D28E4338113EFF2D8AEB560AF947E726C58089D39E75074E7A31244F
  - Architecture: arm64
```

`yq` rejects it: `line 15, column 17: mapping values are not allowed in this context`. The regenerated SteveYegge installer is corrupted the same way. Your new `ARM_SHA` line already uses the precise `grep 'windows_arm64'`, so the fix is a one-word change on line 26 to `grep windows_amd64`. (The Test plan's "parses as valid YAML" check covered the hand-written checked-in file, not the generator output — worth re-running against the generated file.)

**2. The guard does not catch that.** `scripts/check-winget-portable-alias.sh:8` is a bare `grep -q 'PortableCommandAlias:'`. Run against the corrupt output above it prints `OK:` for both files and exits 0. Suggest parsing instead, which validates syntax for free:

```sh
yq -e '.NestedInstallerFiles[] | select(.PortableCommandAlias == "bd")' "$f" >/dev/null
```

Also, the `for f in "$ROOT"/winget/*.installer.yaml` loop exits 0 when the glob matches nothing, so an empty or renamed `winget/` directory passes silently.

**3. The guard runs nowhere.** Every other `scripts/check-*.sh` in this repo is invoked from CI — see `.github/workflows/pr.yml:76-77` (`check-build-tags.sh`, `check-go-install-guidance.sh`), `pr.yml:148` (`check-versions.sh`), `pr.yml:191-192`, and the mirrors in `main.yml`. As added, the new check only runs when someone remembers to type it, which is the failure mode it exists to prevent. One line beside `check-build-tags.sh` in `pr.yml` and `main.yml` fixes it.

Smaller points:

- `winget/GasTownHall.Beads.installer.yaml` pins `PackageVersion: 1.1.0` / `ReleaseDate: 2026-07-04`, but winget-pkgs already publishes 1.1.2 (2026-07-26). Stale on arrival.
- The GasTownHall package can't actually be submitted from what this generates: winget-pkgs needs the version manifest and a defaultLocale manifest at the same `ManifestVersion` alongside the installer (the live 1.1.2 folder has `GasTownHall.Beads.yaml`, `.locale.en-US.yaml` and `.locale.zh-CN.yaml`). `winget/README.md` tells the releaser to place "manifests" under `manifests/g/GasTownHall/Beads/<version>/`, but only the installer exists, so following those instructions produces a rejected submission.
- Related, and worth a sentence in the README: the live manifests carry `# Created with YamlCreate.ps1 Dumplings Mod`, i.e. that package is currently auto-updated in winget-pkgs by the Dumplings bot rather than from this repo. Unless we say which path is authoritative, the in-repo file and the bot will diverge and the bot will keep carrying the alias-less manifest forward — which makes your separate winget-pkgs PR the load-bearing half of the fix, not this one.
- The `ARM_SHA` fallback writes `InstallerSha256: 0000...0` unquoted; 64 bare digits parse as a YAML integer, not the string the schema's `^[A-Fa-f0-9]{64}$` expects (the hand-written SteveYegge manifest quotes its placeholder for exactly this reason). Since `checksums.txt` reliably carries `windows_arm64`, hard-failing like the x64 branch at line 32 would be better than a placeholder.
- Minor: the PR title starts with a literal `# `, which would land in the squash subject.

Happy to push the line-26 grep fix, the CI wiring and the yq-based guard onto your branch as maintainer commits if you'd rather not — say the word, otherwise it's yours.

vishnujayvel:
Thanks — item 1 is the standout: you actually ran `./scripts/update-winget.sh 1.1.2` and got a manifest with two newline-separated hashes crammed into one `InstallerSha256` field, because `grep windows` on line 26 matches both the amd64 and arm64 checksum lines. That's a concrete, reproducible break in the generator, not a hypothetical — and our own guard (`check-winget-portable-alias.sh`) would have said `OK:` on the corrupted output, per your item 2.

Correcting our test plan: we reported `TestCheckBeadsRole_NotConfigured`/`_NotGitRepo` as failing identically on unmodified main. That's wrong — they're environment-local (a global `git config beads.role=maintainer` on the dev machine trips `role_test.go:10-26`); confirmed passing with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`. Harness now neutralizes global git config.

Title: same BSD-sed `\+` bug as the rest of this batch; fixed with an assertion now.

Please push, as maintainer commits: the line-26 grep fix, converting the guard to a `yq`-based parse check, wiring it into CI, and quoting the `ARM_SHA` placeholder. We'll update `PackageVersion`/`ReleaseDate` and add the authoritative-source note about the Dumplings bot ourselves before this ships.

vishnujayvel:
Pushed the four items agreed in this thread on 2026-07-29 (`056b2b5`), so they're in the PR rather than pending on @maphew as separate maintainer commits:

1. **`update-winget.sh`** — `grep windows` → `grep windows_amd64`. The loose grep matched *both* the amd64 and arm64 checksum lines, so `InstallerSha256` was emitted as a two-line value and the manifest didn't parse. Mirrors the existing arm64 grep in the next block.
2. **`update-winget.sh`** — quoted the arm64 `InstallerSha256` placeholder, so 64 bare zero digits aren't parsed as a YAML integer. Matches the quoting already used in the checked-in `SteveYegge.beads.installer.yaml`.
3. **`check-winget-portable-alias.sh`** — replaced the bare `grep -q 'PortableCommandAlias:'` with a `yq` parse, and made it fail closed on an empty `winget/` glob instead of silently exiting 0. The old check reported `OK` on a manifest corrupted by the bug in item 1; the new one fails it.
4. **`pr.yml` / `main.yml`** — wired the guard into the existing `check-build-tags` job so it runs on every PR, instead of only when invoked by hand.

Verification: `shellcheck -x` clean on both scripts; both workflow files parse; the guard passes against the current checked-in manifests and fails against both a reproduction of the item-1 corruption and an empty `winget/` directory.

Two notes for your call:

- The guard now requires `PortableCommandAlias == "bd"` specifically, rather than just the key being present. That's a deliberate narrowing and matches the `yq` expression from your review, but it does mean a manifest with a different alias would now fail.
- `yq` is preinstalled on the `ubuntu-latest` image the `check-build-tags` job uses, so no install step was added. Worth knowing that if `yq` were ever absent, the script reports `FAIL: ... missing valid PortableCommandAlias: bd` rather than "yq not found" — fail-closed, but the message would point at the wrong thing. Happy to add a `command -v yq` guard with a distinct message if you'd prefer.

I deliberately left the smaller points from your review alone — stale `PackageVersion`/`ReleaseDate`, the missing version/locale manifests, and the Dumplings-bot note — since the earlier reply reserved those. Glad to take any of them in a follow-up if useful.

bee-ghosttrack (approved):
**Review at `056b2b5`** (the head that carries the four agreed items).

**APPROVE.** The diff does what it says, and the parts I could measure check out:

- `Check build-tag policy` at this head is green in 7s, so `yq` is on the runner and both `winget/*.installer.yaml` clear the new guard in CI, not only on your machine.
- The checked-in `GasTownHall.Beads.installer.yaml` hashes match the v1.1.0 release `checksums.txt` byte-for-byte (case-folded): `12b1d373…69a6b` amd64, `c2f3fc50…92f7f` arm64.
- `grep windows_amd64` is the right narrowing: `checksums.txt` lines are `<sha>  beads_<v>_windows_amd64.zip` / `…_windows_arm64.zip`, so the old `grep windows` really did return two lines. The quoted arm64 placeholder is the same shape `SteveYegge.beads.installer.yaml` already uses.
- `aka.ms/winget-manifest.installer.1.12.0.schema.json` resolves (200), and it is the schema version the currently published package uses.

Two things worth a follow-up, neither blocking this diff:

1. **The published package does not come from this repo.** Every `GasTownHall.Beads` version in microsoft/winget-pkgs (1.0.3 through 1.3.0, merged 2026-09-16 as winget-pkgs#435480) is authored by `spectopo`'s YamlCreate/Dumplings bot, and the live 1.3.0 installer manifest still has `NestedInstallerFiles: [{RelativeFilePath: bd.exe}]` with **no `PortableCommandAlias`** — so #4908 is current for anyone installing today, and nothing in `winget/` here is copied into that path. The lever that reaches users is the companion winget-pkgs PR you mention (the bot carries fields forward from the previous version's manifest, so fixing 1.3.0 there should stick for 1.3.x+). I don't see one open under a non-bot author; if it exists, a link from this PR body would help whoever follows the README.
2. **The README's "copy to `manifests/g/GasTownHall/Beads/<version>/`" step ships an incomplete package.** That directory holds four files upstream (`.installer.yaml`, `.locale.en-US.yaml`, `.locale.zh-CN.yaml`, and the version manifest `GasTownHall.Beads.yaml`); this PR adds and generates only the installer manifest for `GasTownHall.Beads`, so a maintainer following step 3 would submit a directory winget-pkgs validation rejects. Either generate the version + defaultLocale manifests too (the `SteveYegge.beads` block already shows the shape) or say plainly that this file is the *alias reference* for the bot-published package, not a submittable set.

Nits, take or leave: the checked-in manifest pins 1.1.0 while 1.3.0 is current (regenerating at 1.3.0 would make the file match what someone would actually copy); `checksums.txt` is fetched twice; the amd64 hash is left lowercase while arm64 is uppercased (the schema accepts either, so cosmetic). The all-zero arm64 placeholder passes the guard by design — a line in "Next steps" reminding the releaser to fill it would close that loop.

I am not a merger on this repo; this is a review, not a merge.

vishnujayvel:
Pushed the maintainer-requested commit addressing @maphew's 2026-07-29 review, which the author agreed to the same day:

- `scripts/update-winget.sh`: narrow `grep windows` → `grep windows_amd64` (the loose grep matched both amd64 and arm64 checksum lines, corrupting `InstallerSha256` into an unparseable two-line value)
- `scripts/update-winget.sh`: quote the arm64 `InstallerSha256` placeholder so 64 bare zero digits don't parse as a YAML integer
- `scripts/check-winget-portable-alias.sh`: replace the bare `grep -q PortableCommandAlias:` substring check with a `yq`-based parse that fails closed on a corrupted or empty manifest set
- `.github/workflows/pr.yml`, `main.yml`: wire the guard into the existing `check-build-tags` job so it runs on every PR

New head: 056b2b508. Full CI suite ran clean (119 passing, 2 skipped as expected — same shape as recent merged PRs in this repo).

Issue #4861, opened 2026-07-17
bd init/reinit remote-history guard false-positives on sync.remote without verifying refs/dolt/data; no sanctioned path to bind prefix to existing empty external DB


donnydongchen:
## Summary

`bd init` (and `--reinit-local`) **refuses to proceed based solely on a configured `sync.remote`**, claiming the remote already has Dolt history — **without verifying that `refs/dolt/data` actually exists on that remote**. When the remote has no Dolt refs (only ordinary git branches), this is a false positive that strands the workspace: `init`, `bootstrap`, `config set issue-prefix`, and `rename-prefix` are *all* unavailable, leaving no sanctioned way to bind a prefix to an existing (empty) external-server database.

**Version:** `bd version 1.1.0 (Homebrew)`

## Environment

- External managed Dolt sql-server (`--server`, `127.0.0.1:3307`) hosting many project databases.
- A workspace whose beads database **exists on the server but is empty** (schema present, 0 issues), created out-of-band (e.g. by an orchestrator) and left in a pre-GH#2372 state (**no `project_id`**, `issue-prefix` unset).
- `config.yaml` has a `sync.remote` set to a git URL (`git+https://github.com/owner/Repo.git`) whose remote contains **no** `refs/dolt/*` — only `HEAD` + `refs/heads/main`.

## Steps to reproduce

1. Point a workspace at an existing empty server database; set `sync.remote` to a git URL that has **no** Dolt refs.
2. `bd init --prefix tw`
   → `⚠ Found existing Dolt database ... This workspace is already initialized. ... Aborting.`
3. `bd init --reinit-local --prefix tw`
   → **`bd init refuses: remote 'origin' already has Dolt history (refs/dolt/data).`**
4. Verify the claim is false:
   `git ls-remote git+https://github.com/owner/Repo.git 'refs/dolt/*'` → **empty** (0 refs).
   `git ls-remote ...` (all) → only `HEAD` and `refs/heads/main`.
5. `bd bootstrap`
   → `Bootstrap failed: clone from remote via server: ... can't create database Repo; database exists`
6. `bd config set issue-prefix tw`
   → `Error: issue_prefix cannot be set via 'bd config set'.`
7. `bd rename-prefix tw`
   → `Error: failed to get current prefix: <nil>`

Every sanctioned path is now blocked, and the workspace cannot be docked/registered because downstream tooling fails with `database not initialized: issue_prefix config is missing`.

## Expected

[...]

PR #3909, opened 2026-05-12
fix(init): use git origin as default Dolt remote

maphew:
## Summary

- Configure `bd init` to synthesize a Dolt `origin` remote from the repo's git `origin` when no explicit `sync.remote` is set, and persist it as `sync.remote` so fresh clones/bootstrap use Dolt instead of JSONL.
- Preserve opt-outs and explicit behavior: `--stealth` skips git-origin synthesis, explicit/configured remotes keep precedence, and explicit empty `--remote` still opts out.
- Fix the bootstrap-side gap from #3419: when `bd bootstrap` auto-detects `refs/dolt/data` on git origin, the cloned workspace now also has `origin` wired for future `bd dolt push/pull`.
- Treat `.beads/issues.jsonl` as export/interchange/backup only: JSONL auto-import is skipped when `sync.remote` exists, and auto-export/hooks warn loudly when no Dolt remote is configured.
- Update docs, generated CLI docs, website docs, and agent templates to name Dolt/`refs/dolt/data` as the sync channel, including an explicit repair path for existing pre-fix projects.

## Context

This implements the durable default proposed in #3688: use the existing git remote as the Dolt remote via `refs/dolt/data` rather than leaving new embedded-Dolt projects with only local state.

It also clarifies the architecture confusion captured in #3135: Dolt is the source of truth for `bd list/show/ready` and writes; `.beads/issues.jsonl` is not canonical sync.

#3729's post-merge JSONL import path remains only as a legacy fallback. It is not treated as sufficient sync because JSONL import is upsert-only and cannot reconcile deletes, pruning, or stale local-only records.

#3419 is covered as the fresh-clone counterpart: `bd bootstrap` now rehydrates the remote after cloning from git origin, matching the workaround described by @thewoolleyman but without requiring a manual `bd dolt remote add` on the second machine.

## Existing Project Repair Path

For projects initialized before this change, use the machine whose local Dolt database is authoritative:

```bash
bd dolt remote list
bd export -o .beads/issues.pre-remote.jsonl   # optional audit backup
bd dolt remote add origin git+ssh://git@github.com/org/repo.git
bd dolt push
```

Then commit/push the resulting `.beads/config.yaml` with `sync.remote`. Other clones can run `bd bootstrap` if stale/missing, or `bd dolt pull` when they already have the right database.

## Tests

- `go test -tags gms_pure_go ./cmd/bd`
[...]

Issue #2372, opened 2026-03-05
Dolt server port resolution can cause cross-project data leakage on multi-project machines


PabloLION:
## Summary

The Dolt server port resolution in `DefaultConfig()` uses multiple sources, several of which are (or can be) committed to git. When a shared port value propagates to all collaborators, any contributor running beads in multiple projects risks connecting to the wrong project's Dolt server — silently querying another project's issue database.

This is a design-level issue, not a single-line bug. It affects every multi-project beads user and becomes more likely as beads adoption grows.

## The problem

[`DefaultConfig()`](https://github.com/steveyegge/beads/blob/e4f27bfdb5511640110413f7bc2c9c7d89493513/internal/doltserver/doltserver.go#L258-L314) resolves the port in this priority order:

| Priority | Source | Git-shared? | Risk |
|----------|--------|-------------|------|
| 1 | `BEADS_DOLT_SERVER_PORT` env var | Indirectly (via CI configs, tool settings files) | Low |
| 2 | `metadata.json` → `DoltServerPort` | **Yes** — designed to be committed | **High** |
| 3 | `config.yaml` / `~/.config/bd/config.yaml` → `dolt.port` | Project config.yaml is tracked | Medium |
| 4 | `.beads/dolt-server.port` file | **Not gitignored** — can be committed | **High** |
| 5 | Gas Town fixed port (3307) | No (env var) | N/A |
| 6 | `DerivePort(beadsDir)` — FNV-32a hash, range [13307–14306] | No (uses absolute path) | Low* |

\*DerivePort has a birthday-problem risk: with 1000 ports, ~38 repos on one machine gives >50% collision probability. `reclaimPort()` handles this at server startup, but not at client connection time.

### Why git-shared ports cause data leakage

1. Contributor A sets `DoltServerPort: 3307` in `metadata.json` (or `dolt-server.port` gets committed via `git add -f .beads/`)
2. All 6000 contributors pull this value
3. Contributor B also works on another beads-enabled project that uses port 3307
4. Project A's Dolt server isn't running (idle timeout, reboot)
5. Project B's server IS running on 3307
6. Contributor B runs `bd list` in Project A → connects to port 3307 → gets Project B's issues

### `.beads/dolt-server.port` is not gitignored

The `.beads/.gitignore` carefully excludes related files but misses the port file:

```
# These are ignored:
dolt-server.pid
dolt-server.log
dolt-server.lock

# This is NOT ignored:
# dolt-server.port  ← missing!
[...]

PR #5136, opened 2026-07-29
fix(init): verify refs/dolt/data before remote-history refusal

vishnujayvel:
## Summary

Refs #4861. `bd init --reinit-local` (and the equivalent early-init path) refused with *"remote already has Dolt history (refs/dolt/data)"* whenever `sync.remote` was configured, without checking whether that remote actually advertises `refs/dolt/data` — a git remote with only ordinary branches (`HEAD`/`main`) triggered the same refusal as one with real Dolt history. This matches the false-positive root cause @donnydongchen identified in the issue: the guard inferred remote history from the presence of the `sync.remote` config key rather than from an actual ls-remote probe.

The fix replaces the hardcoded `earlyRemoteHasDoltData = true` (and the later `remoteHasDoltData = true`) assumption for a configured `sync.remote` with a real `gitRemoteHasDoltDataRef(url)` probe, in both the early and late safety checks in `cmd/bd/init.go`, and routes the `initSyncRemoteConfigured` case through the same probe used for `initSyncRemoteExplicit` in the confirmation callbacks. Mere presence of the config key is no longer treated as proof of remote history.

This addresses only the first of the two proposed fixes in the issue — the false-positive guard. It does not add the second: a sanctioned path to bind an existing, empty, prefix-less external-server database and set its `issue-prefix` (`bd rename-prefix` / `bd config set issue-prefix` still refuse on a database with no current prefix). That gap is untouched by this change.

## Test plan

- [x] Targeted: `go test ./cmd/bd/ -run 'TestCheckRemoteSafety_ConfiguredRemoteWithoutDoltData|TestCheckRemoteSafety_GuardMatrix' -v -count=1` — both pass:
  ```
  --- PASS: TestCheckRemoteSafety_ConfiguredRemoteWithoutDoltData (0.00s)
  --- PASS: TestCheckRemoteSafety_GuardMatrix (0.00s)
  PASS
  ok  	github.com/steveyegge/beads/cmd/bd	1.631s
  ```
- [x] Full package suite verified separately: `go test ./cmd/bd/ -count=1 -timeout 25m` passed cleanly (no failures) on a probe branch with this change merged against `origin/main` at `3830f0a34`, runtime 521s.

## Limitations

- Does not implement a sanctioned way to attach an existing empty external-server database and set `issue-prefix` on a pre-GH#2372, prefix-less database — that part of the reported issue remains open.

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -757,6 +757,7 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 			var earlySyncURL string
 			earlyRemoteSource := initSyncRemoteNone
 			earlyRemoteHasDoltData := false
+			var earlyProbeNote string
 			earlySyncURL, earlyRemoteSource = resolveInitConfiguredSyncRemote(initRemote, initRemoteChanged, resolveSyncRemote)
 			if earlyRemoteSource == initSyncRemoteExplicit {
 				// An explicit --remote is intent to bootstrap or wire that URL,
@@ -765,21 +766,27 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 				// source and skips cloning, so we must probe now rather than
 				// silently wiring a populated remote to orphan local history.
 				if fromJSONL {
-					earlyRemoteHasDoltData = gitRemoteHasDoltDataRef(earlySyncURL)
+					hasData, err := gitRemoteHasDoltDataRefStatus(earlySyncURL)
+					earlyRemoteHasDoltData, earlyProbeNote = resolveRemoteHasDoltDataProbe(earlySyncURL, hasData, err)
 				}
 			} else if earlyRemoteSource == initSyncRemoteConfigured {
 				// Probe refs/dolt/data — do NOT treat mere presence of
 				// sync.remote as proof of remote history (GH#4861). A git
 				// remote with only ordinary branches must not refuse
 				// --reinit-local / local init.
-				earlyRemoteHasDoltData = gitRemoteHasDoltDataRef(earlySyncURL)
+				hasData, err := gitRemoteHasDoltDataRefStatus(earlySyncURL)
+				earlyRemoteHasDoltData, earlyProbeNote = resolveRemoteHasDoltDataProbe(earlySyncURL, hasData, err)
 			} else if earlyRemoteSource == initSyncRemoteNone && !stealth && isGitRepo() && !isBareGitRepo() {
 				if originURL, err := gitOriginGetURL(); err == nil && originURL != "" {
 					earlySyncURL = normalizeRemoteURL(originURL)
-					earlyRemoteHasDoltData = gitOriginHasDoltDataRef()
+					hasData, probeErr := gitOriginHasDoltDataRefStatus()
+					earlyRemoteHasDoltData, earlyProbeNote = resolveRemoteHasDoltDataProbe(earlySyncURL, hasData, probeErr)
 				}
 			}
 			if earlySyncURL != "" {
+				if earlyProbeNote != "" {
+					fmt.Fprintf(os.Stderr, "%s %s\n", ui.RenderWarn("!"), earlyProbeNote)
+				}
 				earlyDecision := CheckRemoteSafety(RemoteSafetyInput{
 					Force:             force,
 					ReinitLocal:       reinitLocal,
@@ -790,12 +797,12 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 					RemoteHasDoltData: earlyRemoteHasDoltData,
 					IsInteractive:     term.IsTerminal(int(os.Stdin.Fd())),
 				})
-				if _, err := handleRemoteSafetyDecision(earlyDecision, prefix, earlySyncURL, destroyToken, func() bool {
+				if _, err := handleRemoteSafetyDecision(earlyDecision, prefix, earlySyncURL, destroyToken, func() (bool, error) {
 					switch earlyRemoteSource {
 					case initSyncRemoteExplicit, initSyncRemoteConfigured:
-						return gitRemoteHasDoltDataRef(earlySyncURL)
+						return gitRemoteHasDoltDataRefStatus(earlySyncURL)
 					default:
-						return gitOriginHasDoltDataRef()
+						return gitOriginHasDoltDataRefStatus()
 					}
 				}, earlyRemoteHasDoltData, &remoteDivergenceConfirmed); err != nil {
 					// The early guard refuses and confirms only; bootstrap
@@ -1090,6 +1097,7 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 		syncFromRemote := false
 		syncURLFromGitOrigin := false
 		remoteHasDoltData := false
+		var lateProbeNote string
 
 		if syncURL != "" {
 			// sync.remote was explicitly configured. Treat it as bootstrap-
@@ -1104,7 +1112,11 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 			} else if syncRemoteSource == initSyncRemoteConfigured || syncRemoteSource == initSyncRemoteExplicit {
 				// Always verify refs/dolt/data rather than assuming history
 				// from a configured URL alone (GH#4861).
-				remoteHasDoltData = gitRemoteHasDoltDataRef(syncURL)
+				hasData, err := gitRemoteHasDoltDataRefStatus(syncURL)
+				remoteHasDoltData, lateProbeNote = resolveRemoteHasDoltDataProbe(syncURL, hasData, err)
+			}
+			if lateProbeNote != "" {
+				fmt.Fprintf(os.Stderr, "%s %s\n", ui.RenderWarn("!"), lateProbeNote)
 			}
 			if !syncFromRemote {
 				decision := CheckRemoteSafety(RemoteSafetyInput{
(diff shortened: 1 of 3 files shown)


maphew (commented):
## Disposition: right diagnosis, right layer — one robustness gap worth closing, not a blocker

Thanks for this, and thanks especially for the honest scoping in the description: you say up front that this covers only the first of the two fixes proposed in #4861 and leaves the "attach an empty prefix-less external DB" half open. That kind of framing makes a PR much easier to review.

The bug is real and I confirmed it on current `main`. `cmd/bd/init.go:728-729` hardcodes `earlyRemoteHasDoltData = true // sync.remote configured = user intends bootstrap`, and `cmd/bd/init.go:1028-1029` does the same for `remoteHasDoltData`. `CheckRemoteSafety` (`cmd/bd/init_safety.go`) then sees `RemoteHasDoltData: true` with `ReinitLocal: true` and returns `ActionRefuseDivergence`, printing "remote 'origin' already has Dolt history (refs/dolt/data)". Nothing anywhere on that path ever ran `git ls-remote`. So a `sync.remote` pointing at a plain git repo with only `HEAD` and `refs/heads/main` produces exactly the refusal the reporter saw. `git log -L` on that line traces the assumption back through #3909 to 3e56f0b21 — it has always been an inference from config, never an observation.

Your fix targets the cause, not the symptom, and it is at the right layer: `CheckRemoteSafety` stays a pure decision function and the caller is what learns to observe reality. I also checked the knock-on: with `syncFromRemote` now staying `false` for a configured remote with no Dolt refs, `shouldConfigureInitDoltRemote` still wires the Dolt origin because `syncURLFromConfig` is true, so the remote is not silently dropped. And the reporter's follow-on `bd bootstrap` failure ("can't create database Repo; database exists") no longer fires at init time, since we skip `cloneFromRemoteWithMode` entirely. Both good.

`gh pr list --search` turned up no other open work on this guard, so there is no collision to resolve.

### The probe is fail-open — a robustness gap, not the data-loss path I first read it as

`gitRemoteHasDoltDataRef` (`cmd/bd/sync_git.go:104-115`) returns `false` on *any* error — 10s context timeout, `GIT_TERMINAL_PROMPT=0` credential refusal, DNS failure, non-zero exit for any reason. It cannot distinguish "the remote genuinely has no `refs/dolt/data`" from "I could not ask", and for the `initSyncRemoteConfigured` path that used to be a deliberately fail-safe hardcoded `true`.

I originally read the consequence as silent history loss and want to correct that: a contributor on a private HTTPS `sync.remote` with no cached credential (or on hotel wifi where `ls-remote` exceeds the 10s deadline) running `bd init --reinit-local --prefix tw` does hit a probe error → `false` → `ActionNoRemoteData` → a fresh local identity minted while `shouldConfigureInitDoltRemote` still wires the Dolt origin. But the next `bd dolt push` does not overwrite the team's history — `versioncontrolops.Push` issues a non-force `DOLT_PUSH`, which a real remote with existing history rejects non-fast-forward. So the actual damage is a mis-minted local identity plus a remote that's wired but unusable, and a lost ADR-0002 refusal that would have caught the mismatch earlier — not silent data loss over team history.

It's also not a new failure class this PR introduces: main's sibling git-origin path (`init.go:733`, `1059`, in the `initSyncRemoteExplicit` case) already fail-opens identically for the same command. This PR extends an existing inconsistency rather than inventing one, which is worth knowing before treating it as a hard blocker.

I'd still like to see it closed, and would happily help land it rather than gate on it: give the helper a tri-state (has / has-not / unknown) — e.g. `gitRemoteHasDoltDataRefStatus` returning `(bool, error)` — and treat "unknown" as has-data (refuse, with a message naming the probe failure) at all four ADR-0002 call sites, including the two pre-existing ones on main. That keeps the #4861 false-positive fix, which is correct and worth landing on its own.

### Should-fix: the confirmation re-probe can now abort spuriously

At `cmd/bd/init.go:753` you fold `initSyncRemoteConfigured` into the re-probing branch of the early confirmation callback. The intent is right — main's `return earlyRemoteHasDoltData` was a tautology against a hardcoded `true`, so the "remote state changed during confirmation" recheck in `handleRemoteSafetyDecision` could never fire for configured remotes. But because the probe returns `false` on any error, the newly-live recheck can abort spuriously: on the interactive `--discard-remote` path, the first probe returns `true`, the user types the destroy token, and a transient network error on the confirmation probe returns `false` — `handleRemoteSafetyDecision` (`init.go:2683-2685`) prints "Aborted: remote state changed during confirmation. Re-run to re-verify intent." and exits `ExitRemoteDivergenceRefused` even though nothing changed. The same exposure applies at the late call site (`init.go:1046` post-patch), though the early guard exits first in practice.

Worth noting this is fail-safe rather than harmful — it aborts before any destructive action and is recoverable by re-running — and the identical pattern already exists in main for `initSyncRemoteExplicit`, so this widens an existing class rather than creating a new one. It sits closer to the nit/should-fix boundary than I first flagged it, but the same tri-state probe fix above would close it too: treat a probe error during confirmation as "unchanged" rather than "changed", so only a genuine true→false transition aborts.

### Should-fix: the added test does not exercise the change

`TestCheckRemoteSafety_ConfiguredRemoteWithoutDoltData` calls `CheckRemoteSafety` directly with a hand-supplied `RemoteHasDoltData`. That function is pure and unmodified by this PR, so the test passes identically with and without the `init.go` hunks — revert all four and the suite stays green. Both of its cases are also already asserted by `TestCheckRemoteSafety_GuardMatrix`: `cmd/bd/init_safety_test.go:32` (`fresh/reinit-local` → `ActionNoRemoteData`) and `cmd/bd/init_safety_test.go:49` (`remote/reinit-local` → `ActionRefuseDivergence`).

The real change is in the caller, and it is testable: the probe is just `git ls-remote <url> refs/dolt/data`, so a local bare git repo used as `sync.remote` pins it directly. `cmd/bd/sync_git_test.go:129` and `cmd/bd/bootstrap_test.go:455` already build exactly that fixture shape — a case that sets `sync.remote` to a bare repo with no Dolt refs and asserts `bd init --reinit-local` succeeds (plus its sibling with a pushed `refs/dolt/data` asserting exit `10`) would be the regression test this needs.

### Nit: a lost message

On the configured-remote-with-no-Dolt-refs path, the old code reached `cloneFromRemoteWithMode`, caught `isEmptyRemoteCloneError`, and printed "Remote has no Dolt data yet; initialized a fresh local database". Skipping the clone is the better behaviour, but the user also loses that explanation. A one-line equivalent when the probe comes back clean would keep the output honest about what just happened.

### On the rest of #4861

Agreed that the second half — a sanctioned way to bind an existing empty, prefix-less (pre-#2372) external-server database and set its `issue-prefix` — is out of scope here, and I confirmed `bd config set issue-prefix` and `bd rename-prefix` still refuse in that state. Worth splitting into its own issue so #4861 does not stay open on a half you have already fixed; happy to file it.

The diagnosis and the placement are both right, which is the hard part, and there's no competing open PR here. I don't think the fail-open probe needs to gate merge — I'll post this as findings and offer to help land the tri-state probe and a caller-level regression test alongside you, whichever is easier for you.

vishnujayvel:
Thanks — and thanks for retracting the data-loss reading rather than leaving it in the thread. Your `versioncontrolops.Push` trace is right: a non-force `DOLT_PUSH` gets rejected non-fast-forward, so the damage is a mis-minted local identity and a wired-but-unusable remote, not lost team history. Worth having on the record.

Pushed in `dcf5cdd0e` + `a77e197e6`. All three items.

**Tri-state probe.** `gitRemoteHasDoltDataRefStatus(remote) (bool, error)` and `gitOriginHasDoltDataRefStatus()` in `cmd/bd/sync_git.go`; the old bool wrappers delegate to them and stay for the non-safety callers (`bootstrap.go:160,329`). `resolveRemoteHasDoltDataProbe` (`init.go:2679`) maps a probe error to UNKNOWN → treat as has-data → refuse, naming the probe failure. Routed at **five** sites, not four: the two you named on main (`:733` early, `:1057` late git-origin fallback), the two on the configured path (`:738`, `:1038`), plus the explicit + `--from-jsonl` site at `:726`. That last one is pre-existing and you didn't name it, but it calls the same fail-open helper, and leaving it would have made the tri-state boundary inconsistent — say the word if you'd rather I pull it back out.

Confirmation re-probe: the closures now return `(bool, error)` and a probe error is treated as **unchanged** (`init.go:2715-2719`), so only a genuine true→false transition reaches the "remote state changed during confirmation" abort. That closes the spurious-abort path you flagged.

**The test.** You were right that ours proved nothing — `TestCheckRemoteSafety_ConfiguredRemoteWithoutDoltData` called the pure, unmodified `CheckRemoteSafety` with hand-supplied inputs, and both its cases were already in `TestCheckRemoteSafety_GuardMatrix` (`init_safety_test.go:32`, `:49`). Removed it rather than leave a test that passes with all the `init.go` hunks reverted.

Replaced with two real subprocess `bd init --reinit-local` runs against a bare-repo `sync.remote`: one with no `refs/dolt/data` asserting success, one with it pushed asserting exit `10`.

I verified they're load-bearing rather than assuming it — reverted `init.go` to `b96562015^` and rebuilt:

```
--- FAIL: TestInitReinitLocalConfiguredRemoteWithoutDoltDataSucceeds (15.51s)
    exit status 10 ... remote 'origin' already has Dolt history (refs/dolt/data).
--- PASS: TestInitReinitLocalConfiguredRemoteWithDoltDataRefuses
```
Restored, rebuilt: both PASS.

Two things that came out of hardening it, both worth knowing:

- The tests drive `sync.remote` through `BD_SYNC_REMOTE`, which works via viper's `AutomaticEnv` + `SetEnvPrefix("BD")` (`internal/config/config.go:165-169`) — there's no literal `BD_SYNC_REMOTE` anywhere in the tree, so it's invisible to grep. I confirmed it empirically (`BD_SYNC_REMOTE=/tmp/x bd config get sync.remote` echoes `/tmp/x`) before relying on it.
- I found a hermeticity bug in my own first draft: setting `HOME=workDir` collides with `metrics.go`'s `$HOME/.beads/eventsData` path and the project's own `.beads/`, which quietly broke the "refusal creates no `.beads`" assertion. `a77e197e6` adds a `hermeticInitEnv` helper that strips inherited `BD_*`/`BEADS_*`, uses a separate `t.TempDir()` HOME, and pins `GIT_CONFIG_GLOBAL`/`SYSTEM` to `/dev/null`. I re-ran the whole matrix under `env -i` with an isolated HOME to confirm the pre-fix refusal is the real #4861 bug and not ambient config on my machine — no `BD_SYNC_REMOTE` and no origin gives exit 0, which is what rules that out.

**The lost message** — restored at `init.go:1064`: `! Remote has no Dolt data yet; initialized a fresh local database`, printed when the probe comes back clean on the configured path. Verified by hand, not just by inspection.

`gofmt`, `go vet`, and `CGO_ENABLED=0 go vet ./cmd/bd/` all clean; scoped suite green in 21.7s.

Please do file the second half of #4861 as its own issue — agreed it's out of scope here, and I'd rather #4861 not sit open on a half that's fixed. Happy to take the new one if you'd like.

maphew (approved):
Reviewed at `a77e197e6bf`.

No correctness findings. The tri-state remote-history probe distinguishes no Dolt ref from probe failure, fails closed at the safety gate, preserves the compatibility wrappers, and no longer treats configured `sync.remote` as proof that remote history exists. Tests cover populated and unpopulated configured remotes.

Verification:

- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^(TestInitReinitLocalConfiguredRemote|TestCheckRemoteSafety_)' -count=1` — PASS (`ok`, 34.8s).
- Blocking upstream preflight — blocked on `UNSTABLE` and one failed/action-required check; no merge action was taken.

Disposition: **approve / merge after the normal preflight gate is clean**.

PR #5137, opened 2026-07-29
fix(doctor): require corroboration before flagging test pollution

### LINKED ISSUE #5025: doctor --check=pollution: scores legitimate work as test pollution (open P1 epic at 1.00) and offers to delete it; gating SQL and scorer disagree
> **Provenance:** researched and written by a **Claude Code session** (Opus 4.8) working in my repo, posted under my account — flagging it so you can weight it accordingly. Every claim below is verified against the source at `v1.1.0` and replayed against a live 233-issue server-mode store; nothing is inferred.

**Source at tag `v1.1.0`:** `cmd/bd/detect_pollution.go` (the `--check=pollution` scorer) and `cmd/bd/doctor/validation.go:203-249` (the `Test Pollution` gating check)

## Summary

`bd doctor --check=pollution` flags legitimate engineering work as test pollution, scores an **open P1 epic at 1.00 "high confidence"**, and ends every run by offering to delete the flagged issues:

```
Run 'bd doctor --check=pollution --clean' to delete these issues (with confirmation).
```

In my repo it flags **6 issues, all of them legitimate — a 100% false-positive rate** — and the two gating checks that back it disagree with each other on both the count and the set.

## What it flags (all legitimate)

```
Found 6 potential test issues:

High Confidence (score ≥ 0.9):
  dcr-9f3: "Test-quality architecture" (score: 1.00)
    - Title starts with test prefix
    - Created with 30 other issues in same minute

Medium Confidence (score 0.7-0.9):
  dcr-lzse: "test-driver.sh isolation broken — constants.sh clobbers ..." (0.70)
  dcr-3t0:  "test-agents-md-currency.sh asserts the retired ... wording" (0.70)
  dcr-cdf.34: "test-planner rule: axe tests must render prop-derived ..." (0.70)
  dcr-dk2n: "test-check-19-spec-a11y-gating.sh scatters advisory residue ..." (0.70)
  dcr-3db:  "test-shellcheck.sh teeth fixtures use fixed repo paths, ..." (0.70)
```

Not one is a test artifact:

- **`dcr-9f3`** is an **open P1 epic** with a child and 9 linked reports. "Test-quality architecture" uses *Test* as an ordinary English adjective — the epic is about the *quality of the component test suite*.
- The five `0.70` items are **bugs in the project's own test-infrastructure scripts** (`test-driver.sh`, `test-shellcheck.sh`, `test-planner`, …) — real engineering work about the test harness, which is exactly the kind of title a healthy repo produces.

`--clean` deletes all six. Four are closed; **two are open**, one of them a P1 epic.

## Why the scores land there

`detect_pollution.go:25` matches on a title-word prefix alone:

```go
var testPrefixPattern = regexp.MustCompile(`^(test|benchmark|sample|tmp|temp|debug|dummy)[-_\s]`)
```

Applied to `strings.ToLower(title)`, so anything named `test-*` scores **+0.7** — already at threshold. `test-` / `debug-` / `temp-` are normal prefixes for real work. Then `detect_pollution.go:71`:

```go
if len(issuesByMinute[minute]) >= 10 {
    score += 0.3
    reasons = append(reasons, fmt.Sprintf("Created with %d other issues in same minute", ...))
}
```

That +0.3 is what pushes `dcr-9f3` from 0.7 to **1.00** — but "≥10 issues created in the same minute" is the signature of a **bulk import / backfill**, not of pollution. My report-derived issues were created exactly that way. The scorer treats a legitimate batch import as corroborating evidence of contamination, and there is **no status filter** — `doctor_pollution.go:23` loads every issue with an empty `types.IssueFilter{}`, so closed and open are scored alike.

## The two gating paths disagree — and match different sets

The `Test Pollution` doctor check (`validation.go:217`) uses a **different** heuristic — raw SQL:

```sql
SELECT COUNT(*) FROM issues
WHERE (title LIKE 'test-%' OR title LIKE 'Test Issue%'
    OR title LIKE '%test issue%' OR id LIKE 'test-%')
```

Replayed against the same store it returns **5**, while the scorer returns **6**. The difference is case sensitivity: Dolt evaluates `LIKE` here case-sensitively, so the SQL **misses `Test-quality`** —

```
> SELECT 'Test-quality architecture' LIKE 'test-%';  -- 0
```

— while the scorer lowercases first and catches it at 1.00. So the two code paths that both claim to detect "test pollution" match overlapping-but-different sets and report different counts (5 vs 6). Same concept, two implementations, two answers, neither correct.

## Why it matters

This is the one doctor check whose documented remediation is **destructive**. A user who follows the check's own `Fix` string to `--clean` deletes real work — in my case an open P1 epic and an open infra bug — on the strength of a title word and a bulk-import timestamp. `--clean` does back up to a file first, but a backup is not consent, and "high confidence 1.00" actively invites the deletion.

## Suggested fixes

1. **Exclude closed issues** (or at least never offer to delete them) — `doctor_pollution.go:23` should pass a status filter.
2. **Require a corroborating signal**, not a title prefix alone. `test-`/`debug-`/`temp-` on a titled, described, linked issue with children is not pollution. Gate on the *combination* (prefix **and** empty/near-empty description **and** sequential bare ID), which the scorer already computes but doesn't require.
3. **Don't treat same-minute bulk creation as evidence.** It is equally the signature of an import — arguably negative evidence, since real pollution tends to trickle in from ad-hoc test runs.
4. **Reconcile the SQL check with the scorer** so the two paths agree on case-handling and on the flagged set.
5. **Never present a destructive action at "1.00 high confidence"** off a heuristic this shallow. At minimum, exclude epics and issues with children/links from `--clean` eligibility.

## Context

Found while auditing with a CGO-enabled rebuild — `Test Pollution` is one of the checks that is dark in a `CGO_ENABLED=0` build (see #4977), so on many installs this scorer never runs at all; where it does, this is what it produces.

---

vishnujayvel:
Closes #5025

## Summary

Fixes the false-positive behavior reported by @ek in #5025: `bd doctor --check=pollution` was scoring legitimate engineering work as test pollution — including an open P1 epic ("Test-quality architecture") at 1.00 "high confidence" — and offering `--clean` deletion for it, while the separate `Test Pollution` SQL gating check disagreed with the scorer's count due to a case-sensitive `LIKE` comparison.

This change: (1) never scores closed issues or epics as pollution candidates; (2) removes same-minute bulk creation as a scoring signal, since that pattern is the signature of a bulk import, not pollution — this was previously the +0.3 that pushed the P1 epic to 1.00; (3) requires a `test-`/`debug-`/etc. title prefix to be corroborated by a real fixture signal (empty or near-empty description, or a sequential bare ID with a thin description, or an explicit generic title like "test issue") before it counts toward the 0.7 threshold, so a title prefix alone no longer flags real work; (4) aligns the `Test Pollution` doctor SQL check with the scorer by filtering to open non-epic issues, lowercasing the title comparison, and matching on more specific fixture patterns instead of a bare `test-%` title prefix; (5) restricts `--clean` deletion to high-confidence hits only, leaving medium-confidence hits as review-only.

## Test plan

- [x] Targeted run of the new/changed pollution-detection tests, from the worktree with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null` and the project's CGO/ICU env vars set:
  ```
  go test ./cmd/bd/ -run 'TestDetectTestPollution' -v -count=1
  ```
  Result:
  ```
  === RUN   TestDetectTestPollution_SkipsClosedAndEpics
  --- PASS: TestDetectTestPollution_SkipsClosedAndEpics (0.00s)
  === RUN   TestDetectTestPollution_PrefixAloneNotEnough
  --- PASS: TestDetectTestPollution_PrefixAloneNotEnough (0.00s)
  === RUN   TestDetectTestPollution_RequiresCorroboration
  --- PASS: TestDetectTestPollution_RequiresCorroboration (0.00s)
  === RUN   TestDetectTestPollution_BulkCreateNotEvidence
  --- PASS: TestDetectTestPollution_BulkCreateNotEvidence (0.00s)
  PASS
  ok  	github.com/steveyegge/beads/cmd/bd	1.719s
  ```
- [x] Full package suite (`go test ./cmd/bd/ -count=1 -timeout 25m`) was separately run to completion (465s) against this branch merged with `origin/main` at `3830f0a34`, with a clean result for every test related to this change.

## Notes

The issue's suggested fixes also mention excluding issues with children/links from `--clean` eligibility as a further hardening step; this PR does not add that check — `--clean` eligibility here is gated on the open/non-epic/high-confidence corroboration logic described above, which covers every false positive reported in the issue (the P1 epic and the five test-infra bug reports).

### DIFF
--- a/cmd/bd/detect_pollution.go
+++ b/cmd/bd/detect_pollution.go
@@ -91,11 +91,18 @@ func detectTestPollution(issues []*types.Issue) []pollutionResult {
 			reasons = append(reasons, "Very short description")
 		}
 
-		// Explicit generic fixture titles (strong corroboration).
+		// Explicit generic fixture titles (corroboration, not a standalone
+		// clean signal). Weighted at 0.3 rather than 0.5 (GH#5137 review):
+		// these are unanchored substring matches ("test issue" also appears
+		// inside legitimate titles like "Fix test issue detection
+		// regression"), so title-substring + empty-description alone
+		// (0.3 + 0.4 = 0.7) must land in the review-only band and never
+		// reach the 0.9 --clean cutoff without another corroborating signal
+		// (e.g. a test-prefixed title or a sequential/hash fixture ID).
 		if strings.Contains(title, "issue for testing") ||
 			strings.Contains(title, "test issue") ||
 			strings.Contains(title, "sample issue") {
-			score += 0.5
+			score += 0.3
 			corroboration = true
 			reasons = append(reasons, "Generic test title")
 		}
--- a/cmd/bd/detect_pollution.go
+++ b/cmd/bd/detect_pollution.go
@@ -30,14 +30,7 @@ func isTestIssue(title string) bool {
 	return testPrefixPattern.MatchString(strings.ToLower(title))
 }
 
-// detectTestPollution scores candidates that look like leaked test fixtures.
-//
-// Policy (GH#5025):
-//   - Only open non-epic issues (closed work and epics are never pollution-clean targets).
-//   - A title prefix like "test-" alone is NOT enough — real engineering uses those prefixes.
-//   - Require at least one corroborating signal (empty/minimal description, sequential bare ID
-//     with thin description, or generic "test issue" title).
-//   - Same-minute bulk creation is NOT evidence of pollution (it is the signature of imports).
+// detectTestPollution scores candidates that look like leaked test fixtures (GH#5025); see the inline signal comments below for the corroboration policy.
 func detectTestPollution(issues []*types.Issue) []pollutionResult {
 	var results []pollutionResult
 	sequentialPattern := regexp.MustCompile(`^[a-z]+-\d+$`)
@@ -62,12 +55,17 @@ func detectTestPollution(issues []*types.Issue) []pollutionResult {
 		desc := strings.TrimSpace(issue.Description)
 
 		// Title prefix is a weak signal alone (0.4) — real bugs/infra use "test-*" names.
-		if testPrefixPattern.MatchString(title) {
+		hasPrefix := testPrefixPattern.MatchString(title)
+		if hasPrefix {
 			score += 0.4
 			reasons = append(reasons, "Title starts with test prefix")
 		}
 
 		// Sequential bare ID + minimal description (corroborates fixture-style issues).
+		// NB: this pattern also matches ordinary "prefix-N" IDs from
+		// issue_id_mode=counter (internal/storage/issueops/helpers.go
+		// NextCounterIDTx) or an explicit `bd create --id bd-42`, so it must
+		// never be enough on its own without other test-specific signals.
 		if sequentialPattern.MatchString(issue.ID) && len(desc) < 20 {
 			score += 0.4
 			corroboration = true
@@ -80,7 +78,13 @@ func detectTestPollution(issues []*types.Issue) []pollutionResult {
 			corroboration = true
 			reasons = append(reasons, "No description")
 		} else if len(desc) < 20 {
-			score += 0.2
+			// Bump applies only with a test-prefixed title (GH#5137): counter-mode
+			// "prefix-N" IDs alone must not push a real issue over threshold.
+			weight := 0.2
+			if hasPrefix {
+				weight = 0.3
+			}
+			score += weight
 			corroboration = true
 			reasons = append(reasons, "Very short description")
 		}
--- a/cmd/bd/detect_pollution.go
+++ b/cmd/bd/detect_pollution.go
@@ -30,61 +30,72 @@ func isTestIssue(title string) bool {
 	return testPrefixPattern.MatchString(strings.ToLower(title))
 }
 
+// detectTestPollution scores candidates that look like leaked test fixtures.
+//
+// Policy (GH#5025):
+//   - Only open non-epic issues (closed work and epics are never pollution-clean targets).
+//   - A title prefix like "test-" alone is NOT enough — real engineering uses those prefixes.
+//   - Require at least one corroborating signal (empty/minimal description, sequential bare ID
+//     with thin description, or generic "test issue" title).
+//   - Same-minute bulk creation is NOT evidence of pollution (it is the signature of imports).
 func detectTestPollution(issues []*types.Issue) []pollutionResult {
 	var results []pollutionResult
 	sequentialPattern := regexp.MustCompile(`^[a-z]+-\d+$`)
 
-	// Group issues by creation time to detect rapid succession
-	issuesByMinute := make(map[int64][]*types.Issue)
 	for _, issue := range issues {
-		minute := issue.CreatedAt.Unix() / 60
-		issuesByMinute[minute] = append(issuesByMinute[minute], issue)
-	}
+		if issue == nil {
+			continue
+		}
+		// Never flag closed issues or epics (destructive --clean must not touch real work).
+		if issue.Status == types.StatusClosed {
+			continue
+		}
+		if issue.IssueType == types.TypeEpic {
+			continue
+		}
 
-	for _, issue := range issues {
 		score := 0.0
 		var reasons []string
+		corroboration := false
 
 		title := strings.ToLower(issue.Title)
+		desc := strings.TrimSpace(issue.Description)
 
-		// Check for test prefixes (strong signal)
+		// Title prefix is a weak signal alone (0.4) — real bugs/infra use "test-*" names.
 		if testPrefixPattern.MatchString(title) {
-			score += 0.7
+			score += 0.4
 			reasons = append(reasons, "Title starts with test prefix")
 		}
 
-		// Check for sequential numbering (medium signal)
-		if sequentialPattern.MatchString(issue.ID) && len(issue.Description) < 20 {
+		// Sequential bare ID + minimal description (corroborates fixture-style issues).
+		if sequentialPattern.MatchString(issue.ID) && len(desc) < 20 {
 			score += 0.4
+			corroboration = true
 			reasons = append(reasons, "Sequential ID with minimal description")
 		}
 
-		// Check for generic/empty description (weak signal)
-		if len(strings.TrimSpace(issue.Description)) == 0 {
-			score += 0.2
+		// Empty / near-empty description (corroborates throwaway fixtures).
+		if len(desc) == 0 {
+			score += 0.4
+			corroboration = true
 			reasons = append(reasons, "No description")
-		} else if len(issue.Description) < 20 {
-			score += 0.1
+		} else if len(desc) < 20 {
+			score += 0.2
+			corroboration = true
 			reasons = append(reasons, "Very short description")
 		}
 
-		// Check for rapid creation (created with many others in same minute)
-		minute := issue.CreatedAt.Unix() / 60
-		if len(issuesByMinute[minute]) >= 10 {
-			score += 0.3
-			reasons = append(reasons, fmt.Sprintf("Created with %d other issues in same minute", len(issuesByMinute[minute])-1))
-		}
-
-		// Check for generic test titles
+		// Explicit generic fixture titles (strong corroboration).
 		if strings.Contains(title, "issue for testing") ||
 			strings.Contains(title, "test issue") ||
 			strings.Contains(title, "sample issue") {
 			score += 0.5
+			corroboration = true
 			reasons = append(reasons, "Generic test title")
 		}
 
-		// Only include if score is above threshold
-		if score >= 0.7 {
+		// Threshold AND corroboration: prefix-only titled real work stays unflagged.
+		if score >= 0.7 && corroboration {
 			results = append(results, pollutionResult{
 				issue:   issue,
 				score:   score,
--- a/cmd/bd/detect_pollution.go
+++ b/cmd/bd/detect_pollution.go
@@ -30,7 +30,8 @@ func isTestIssue(title string) bool {
 	return testPrefixPattern.MatchString(strings.ToLower(title))
 }
 
-// detectTestPollution scores candidates that look like leaked test fixtures (GH#5025); see the inline signal comments below for the corroboration policy.
+// detectTestPollution scores candidates that look like leaked test fixtures (GH#5025);
+// see the inline signal comments below for the corroboration policy.
 func detectTestPollution(issues []*types.Issue) []pollutionResult {
 	var results []pollutionResult
 	sequentialPattern := regexp.MustCompile(`^[a-z]+-\d+$`)
@@ -80,6 +81,7 @@ func detectTestPollution(issues []*types.Issue) []pollutionResult {
 		} else if len(desc) < 20 {
 			// Bump applies only with a test-prefixed title (GH#5137): counter-mode
 			// "prefix-N" IDs alone must not push a real issue over threshold.
+			// Deliberately clears 0.7 on prefix+thin alone; see the boundary test.
 			weight := 0.2
 			if hasPrefix {
 				weight = 0.3
--- a/cmd/bd/detect_pollution_test.go
+++ b/cmd/bd/detect_pollution_test.go
@@ -137,3 +137,19 @@ func TestDetectTestPollution_PrefixSequentialIDShortDescriptionScoresHigh(t *tes
 		t.Fatalf("score = %v, want >= 0.9 (high confidence)", got[0].score)
 	}
 }
+
+func TestDetectTestPollution_PrefixHashIDThinDescriptionAtBoundary(t *testing.T) {
+	// Prefixed title (0.4) + hash id + thin description (0.3) = 0.70 (GH#5137):
+	// flagged for review, but below the 0.9 band clean mode acts on.
+	issues := []*types.Issue{
+		mkPollutionIssue("dcr-a3f9", "test-driver.sh isolation broken — constants.sh clobbers env",
+			"Broke isolation", types.StatusOpen, types.TypeBug),
+	}
+	got := detectTestPollution(issues)
+	if len(got) != 1 {
+		t.Fatalf("expected boundary fixture to be flagged for review, got %d hits", len(got))
+	}
+	if got[0].score < 0.7 || got[0].score >= 0.9 {
+		t.Fatalf("score = %v, want exactly 0.70 (flagged, but below --clean's 0.9 high-confidence gate)", got[0].score)
+	}
+}
--- a/cmd/bd/detect_pollution_test.go
+++ b/cmd/bd/detect_pollution_test.go
@@ -138,6 +138,24 @@ func TestDetectTestPollution_PrefixSequentialIDShortDescriptionScoresHigh(t *tes
 	}
 }
 
+func TestDetectTestPollution_GenericTitleSubstringAloneStaysReviewOnly(t *testing.T) {
+	// Regression for GH#5137: an unanchored substring match on "test issue"
+	// must not, combined with just an empty description, reach the 0.9
+	// --clean cutoff. "Fix test issue detection regression" is a legitimate
+	// issue title (not a test-prefixed fixture, not a sequential/hash
+	// fixture ID) and must stay in the review-only band, never cleanable.
+	issues := []*types.Issue{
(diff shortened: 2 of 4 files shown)


maphew (commented):
## Disposition: approve with a follow-up — this is the right fix in the right place, one scoring gap to close first

Thanks for taking #5025 on. I reproduced the whole chain locally against this branch and current `main`, and the diagnosis holds up end to end: on `main`, `cmd/bd/detect_pollution.go:51-54` gives a bare `test-`/`debug-`/`tmp-` title prefix `+0.7`, which is already the inclusion threshold at line 87, and the same-minute bulk-creation bonus at lines 71-76 adds the `+0.3` that carried the reporter's "Test-quality architecture" epic to a `1.00` "high confidence" offer to delete. Removing the bulk-create signal, gating on corroboration, and skipping closed issues and epics all target the actual cause rather than the symptom, and they're in the scorer where they belong rather than papered over at the call site. `go vet` is clean, the four new tests pass, and the branch merges into current `main` without touching anything that has moved since its base.

The `--clean` narrowing in `cmd/bd/doctor_pollution.go:108-146` is also the right instinct — a destructive path should only act on the high band. One consequence of the new numbers is worth working through before we ship.

### A fixture with a one-word description now disappears from the report entirely

`cmd/bd/detect_pollution.go`: prefix `0.4` + "very short description" `0.2` = `0.60`, below the `0.7` inclusion threshold at line 87. Confirmed on the branch:

```
bd-1hao  "test-foo"  description "x"  ->  NOT FLAGGED
```

On `main` that same issue scores `0.7 + 0.1 = 0.8` and is reported (medium confidence). The narrowed SQL gate in `cmd/bd/doctor/validation.go` doesn't pick it up either — no "test issue"/"sample issue" substring, and the ID doesn't match the bare sequential pattern — so both detection surfaces go silent on it. It's not as broad as "every short-description fixture" — an issue with a sequential bare ID plus short description still scores `1.0`, and generic-title fixtures (`test issue`, `sample issue`, `issue for testing`) still score `1.1`+ — but it is a real dead zone the PR description doesn't disclose, and the code's own `corroboration = true` on this branch shows the intent was for it to qualify. Bumping the short-description signal from `0.2` to `0.3` (or lowering the gate for corroborated hits) closes it without reopening the prefix-only false positives the PR targets.

### Smaller things

- `cmd/bd/doctor_pollution.go:62-81` returns on the JSON path before any of the new `--clean` gating, so JSON consumers still see one flat `issues` list with no indication of which entries are cleanable. The per-issue `score` is there, but the `0.9` policy isn't. A `cleanable` boolean per issue (or a `high_confidence_ids` array) would make the new safety rule visible to the non-interactive path too.
- `cmd/bd/detect_pollution_test.go:11` declares `func issue(...)` at package scope in `package main`. It compiles today, but `issue` is a heavily used local variable name across `cmd/bd`'s test files and it's a broad name to claim in a shared package namespace — `mkPollutionIssue` would age better.
- One deliberate scope note for the record: skipping closed issues (`detect_pollution.go`) means closed leaked fixtures are no longer *reported*, not just no longer deletable. #5025 asked to never *delete* them. That's probably fine, and your description does say so plainly — flagging it only so it's a conscious call rather than a side effect.

### Things I checked and found clean

- The `LOWER(...)` change in `cmd/bd/doctor/validation.go:216-230` is justified — the reporter replayed `SELECT 'Test-quality architecture' LIKE 'test-%'` against a live Dolt store and got `0`. I didn't independently re-run that query, so I'm taking the issue's evidence at face value there.
- `status` and `issue_type` are both `NOT NULL` in `internal/storage/schema/migrations/0001_create_issues.up.sql`, so the new `status != 'closed' AND issue_type != 'epic'` predicates can't silently drop rows to three-valued logic.
- `TestValidateCheck_DetectsTestPollution` (`cmd/bd/doctor_validate_test.go:177`) still passes under the new SQL, via both the `LOWER(title) LIKE 'test issue%'` clause and the retained id clause.
- `fmt` is still used elsewhere in `detect_pollution.go`, so dropping the `Sprintf` in the bulk-create branch doesn't strand the import.
- The high-confidence band (`>= 0.9`) is very much still reachable in practice: the generic-title branch (`test issue` / `sample issue` / `issue for testing`, matched case-insensitively) is untouched by the reweighting and is the dominant fixture shape — a base36-ID fixture titled "test issue 5" or "sample issue 3" still scores `1.30`. `--clean` has plenty left to act on.
- No collision with other open work: the only adjacent PR is your own #5133, which touches CI for the `cmd/bd/doctor` Dolt tests and doesn't overlap these files.

None of the above is a reason to send this back — the hard part, which was figuring out *why* a title prefix alone was enough, is done and done correctly. Close the short-description gap and I'm happy to land it.

vishnujayvel:
Thanks — the dead zone is real and closing it turned up something worse in my own first attempt, so this is a longer answer than the fix deserves.

Pushed in `e42bb260b`.

**I took your first option (bump the short-description signal) but gated it, because the ungated version was broken.** My initial version bumped `0.2` → `0.3` unconditionally. I rejected your second option (lower the gate for corroborated hits) on the grounds that `sequentialPattern` (`^[a-z]+-\d+$`) matches ordinary IDs, not just fixtures — and then shipped a fix with exactly that defect: an ordinary issue with a `bd-42`-style ID and any real-but-terse description scores `0.4 (sequential) + 0.3 (short-desc) = 0.70`, dead on the threshold, with **no test-related signal at all**.

That premise is worth stating with citations, because it decides the whole question:

- `bd create --id` is documented and uses this exact shape in its own help: `cmd/bd/create.go:762` — `"Explicit issue ID (e.g., 'bd-42' for partitioning)"`.
- `issue_id_mode=counter` mints `prefix-N` for **every** issue while set: `internal/storage/issueops/helpers.go` `NextCounterIDTx` returns `fmt.Sprintf("%s-%d", prefix, nextID)`.
- Even default hash IDs land on an all-digit suffix by chance — `internal/idgen/hash.go` uses base36, so roughly 2% at the default 3-char length.

So a sequential ID is an ordinary supported shape, and neither option is safe if it lets ID shape alone carry an issue to threshold.

The fix: gate the bump on the **title prefix**, not the ID shape.

| case | score | flagged |
|---|---|---|
| `test-foo` / desc `"x"` (your dead zone) | `0.4 prefix + 0.3 short-desc` = **0.70** | yes ✓ |
| bare `test-` prefix, normal description | `0.4` | no ✓ |
| `bd-42`, no test-y title, terse real description | `0.4 sequential + 0.2 short-desc` = **0.60** | no ✓ |

The third row is the one my first attempt got wrong. There's a test pinning each, plus an `NB:` comment above `sequentialPattern` so the next person doesn't re-widen it.

**On your `1.0` figure** — I initially couldn't reproduce it and worked out why: it needs three signals, not two. Title prefix `0.4` + sequential-ID-with-short-desc `0.4` + short-desc `0.2` = `1.0` exactly, e.g. ID `test-42`, title `test-42 fixture`, desc `wip`. That's the realistic fixture shape you're describing throughout, so your number was right and my arithmetic was the incomplete one. Under this fix that same fixture scores `1.1`.

Also done: `cleanable` boolean per issue on the JSON path (`cmd/bd/doctor_pollution.go`), chosen over a separate `high_confidence_ids` array so it sits next to each issue's existing `score`/`reasons` rather than making consumers cross-reference; and `issue(...)` → `mkPollutionIssue(...)` in the test file, all 8 call sites.

Your three "things I checked and found clean" all match what I see, and I did **not** touch `cmd/bd/doctor/validation.go`'s SQL gate — the prefix-gated bump doesn't need it, and `TestValidateCheck_DetectsTestPollution` still passes untouched.

Tests: all 8 pass with `=== RUN` / `--- PASS` in the default build; `gofmt` and `go vet ./cmd/bd/...` clean. Nothing here needs Docker.

One disclosure: our pre-publish comment-density check still flags this diff, and I chose not to chase it to green. The two residual findings are the `NB:` note above `sequentialPattern` — which I think earns its lines, since it's the thing preventing a future re-widening — and a two-line comment from the base commit that your review specifically praised. Cutting either would trade real reasoning for a passing script. Happy to trim if you disagree.

maphew:
The corroboration gate is right, and `e42bb260` answers the counter-mode concern I raised — I re-derived the arithmetic rather than taking the comment's word for it:

A legitimate `issue_id_mode=counter` id with a thin description and no test prefix scores `0.4` (sequential + short desc) `+ 0.2` (short desc, unprefixed weight) `= 0.6`, under the `0.7` threshold. So the claim in the new comment holds — counter-mode ids alone can't push real work over. Good.

But running the same arithmetic on the case the bump *does* change surfaces something I'd like your read on before this lands.

Test-prefixed title + short-but-non-empty description + a **hash** id (not sequential):

- title prefix: `+0.4`
- sequential id: no match
- short description: `+0.3` (was `+0.2`)

That totals exactly `0.7` with `corroboration = true`, where before the delta it was `0.6` and fell short. So the practical effect of the bump is to newly flag prefixed issues with hash ids and thin descriptions — and `--clean` offers those for deletion.

That's the same false-positive class #5025 was filed about, one notch narrower: a real issue titled something like `test-driver.sh isolation broken` with a 19-character description now lands exactly on the line. `TestDetectTestPollution_PrefixAloneNotEnough` won't catch it, because its fixtures use long descriptions.

I don't think this is wrong — an issue with under 20 characters of description genuinely is thin, and you may well have intended to close that dead zone deliberately (the commit subject says as much). I'd just like it pinned rather than incidental:

- a test at the boundary — prefixed title, hash id, ~15-char description — asserting whichever way you intend it, so a future weight tweak can't move it silently
- one line confirming the widening is deliberate

Once that's in I'll arm the merge. Everything else here I'm happy with; moving the policy summary inline is fine, though the header comment is now a single long line where a wrapped two-liner would read better.

vishnujayvel:
Re-derived the arithmetic and it checks out — with one correction on the practical effect.

Pushed in `52430ecc1`.

Test-prefixed title + hash id (not sequential) + thin description:

- title prefix: `+0.4` (`testPrefixPattern` — `cmd/bd/detect_pollution.go:25`, `^(test|benchmark|sample|tmp|temp|debug|dummy)[-_\s]`, anchored, so it matches on the literal prefix word immediately followed by `-`/`_`/whitespace)
- sequential id: no match (hash ids like `dcr-a3f9` don't satisfy `^[a-z]+-\d+$` — the suffix isn't all-digit)
- short description: `+0.3` (`detect_pollution.go:81-90`, `len(desc) < 20`, prefix-gated weight)

`0.4 + 0.3 = 0.70`, `corroboration = true` (set by the short-description branch). That clears the `score >= 0.7 && corroboration` gate at `detect_pollution.go:104`, so yes — it's flagged, exactly on the line, exactly as you worked out.

One correction on severity, though: **`--clean` doesn't act on this.** `runPollutionCheck` collects `highConfidence` as `p.score >= 0.9` (`doctor_pollution.go:55`), and `cleanable := highConfidence` (`doctor_pollution.go:112`) — only the high band is eligible for deletion; medium is explicitly review-only, both in the human-readable output (`doctor_pollution.go:100`, the `Medium Confidence (score 0.7-0.9)` section) and in the `cleanable` JSON field added in the same commit (`doctor_pollution.go:78`, also `p.score >= 0.9`). A 0.70 hit lands in medium, gets listed for a human to look at, and `--clean` won't touch it. So `test-driver.sh isolation broken` with a 15-19 char description gets surfaced for review, not silently deleted — a real UX cost (it's noise in the list) but a bounded one, not the destructive false positive #5025 was originally about.

Given that, my read is **(A), affirm** — the widening is deliberate and I'd rather pin it than narrow it further:

- The false-positive combination requires *both* a literal test-prefix-word match *and* a genuinely thin (<20 char) description. Real, thoughtfully-filed bugs with prefix-ish titles tend to carry substantive descriptions — which is exactly the case `TestDetectTestPollution_PrefixAloneNotEnough` already pins, using that same `test-driver.sh isolation broken` title with a long description and asserting it is *not* flagged. So the two tests now bracket the boundary from either side on identical titles, and the only thing separating them is description thickness. A prefix-titled issue with a 15-char description is already a thinner signal for "real work" than the title alone suggests.
- The cost of a false hit here is "shows up in a review-only list," not deletion. That's a materially different risk than what #5025 flagged.

Done:

- `TestDetectTestPollution_PrefixHashIDThinDescriptionAtBoundary` (`cmd/bd/detect_pollution_test.go`) pins exactly your scenario: title `test-driver.sh isolation broken — constants.sh clobbers env`, hash id `dcr-a3f9`, 15-char description `Broke isolation`. Asserts `0.7 <= score < 0.9` — flagged, but below the `--clean` gate.
- One line at `detect_pollution.go:84`, next to the weight bump, confirming the widening is deliberate and pointing at the boundary test.
- Header comment (`detect_pollution.go:33`) rewrapped into two lines.

Kept it to those three changes — no other comment additions. Our comment-density check is clean on `detect_pollution.go` and still flags the test file by one line (2 added comment lines where its baseline ratio allows ~1.4); I kept the second line because it carries the arithmetic, which is the part worth having in-tree when someone next touches the weights.

Tests: all 9 `TestDetectTestPollution_*` pass, including the new one. `gofmt -l` clean on both touched files.

Full `./cmd/bd/...` has four failures on my machine, all of which I confirmed are pre-existing rather than assuming it — each reproduces identically on `e42bb260b`, the commit before this one:

- `cmd/bd/doctor`: `TestCheckBeadsRole_NotConfigured`, `TestCheckBeadsRole_NotGitRepo` — my global `git config beads.role=maintainer` leaks into them.
- `cmd/bd`: `TestInitRepairsPermissiveBeadsDir`, `TestInitPreservesSecureBeadsDir` — `.beads` directory permission expectations (0700 vs my umask's 0755).

All four are environmental and none touch the pollution scorer. Also disclosing: the Dolt-backed tests were **skipped**, not run — Docker isn't available here (`WARN: Docker not available, skipping Dolt tests`). Nothing in this change goes near that path, but I'd rather say so than let a green summary imply coverage I don't have.

Used the canonical `-tags gms_pure_go` path (`.buildflags`), so no ICU linkage involved.

maphew (commented):
Reviewed at `52430ecc174`.

This substantially narrows pollution cleanup and protects medium-confidence hits, but one destructive false positive remains: any title containing `test issue` plus an empty description scores exactly `0.9`, reaching the auto-clean threshold. A legitimate issue such as “Fix test issue detection regression” with no body can therefore be deleted.

Use anchored/exact fixture-title patterns, or keep substring matches in the review-only band, and add a negative regression.

Verification:

- Default focused build — blocked because local ICU headers are unavailable.
- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run '^TestDetectTestPollution_' -count=1` — PASS (`ok`, 22.3s); the legitimate-title boundary is missing.
- Blocking upstream preflight — PASS; merge state clean and no failed or pending checks.

Disposition: **fix-merge**. Preserve the focused safety improvement and close this remaining destructive edge on the contributor branch.

vishnujayvel:
Confirmed — I traced the arithmetic in `detect_pollution.go`: the generic-title check at line ~94-101 is a plain case-insensitive substring match (`strings.Contains(title, "test issue")`, among others), contributing 0.5, and the empty-description check at line ~76-80 contributes 0.4, for a sum of exactly 0.9. `doctor_pollution.go` uses `score >= 0.9` as the `highConfidence`/`cleanable` cutoff (lines ~55, ~78), so `"Fix test issue detection regression"` with no body does land in the auto-clean bucket. I also checked `detect_pollution_test.go`: all existing cases use fixture-style or test-prefixed titles, so `TestDetectTestPollution_*` passing doesn't cover this boundary — it's genuinely untested.

I'll fix this by taking the substring-matched generic-title titles (`"test issue"`, `"issue for testing"`, `"sample issue"`) out of the auto-clean path: either anchoring them to the known fixture-title patterns actually produced by test scaffolding, or dropping their standalone score contribution so that title-substring + empty-description alone stays in the review-only band and no longer reaches 0.9 without an additional corroborating signal (e.g. a matching test-ID pattern).

I'll add a negative regression test asserting that a realistic non-fixture title like `"Fix test issue detection regression"` with an empty description scores below the auto-clean threshold and is not `cleanable`, alongside the existing fixture-title positive cases.

PR #5138, opened 2026-07-29
fix(epic): do not treat duplicate/wontfix closes as complete

### LINKED ISSUE #5026: doctor: Stale Molecules 'complete-but-unclosed' counts duplicate-closed children as done — suggests bd close on epics with real remaining scope
> **Provenance:** researched and written by a **Claude Code session** (Opus 4.8) working in my repo, posted under my account — flagging it so you can weight it accordingly. Verified against the source at `v1.1.0` and against a live server-mode store.

**Source at tag `v1.1.0`:** `internal/storage/issueops/epic_closure.go:148` and `cmd/bd/doctor/maintenance.go:133-193` (the `Stale Molecules` check)

## Summary

`bd doctor`'s **Stale Molecules** check reports an epic as "complete-but-unclosed" and suggests `bd close` whenever all of its children are closed — **regardless of *why* they were closed**. An epic whose only child was closed as a **duplicate** (or `wontfix` / `superseded`) reads as finished work. Acting on the suggestion silently drops whatever the epic still had open.

```
⚠  Stale Molecules: 1 complete-but-unclosed molecule(s)
    Example: [dcr-tf2]
    └─ Run 'bd mol stale' to review, then 'bd close <id>' for each
```

The word "complete" is **asserted, not established** — the check has no basis for it.

## Root cause

Eligibility is computed in `epic_closure.go`, counting only the `closed` status and nothing else:

```go
closedChildren := 0
for _, childID := range children {
    if status, ok := childStatusMap[childID]; ok && types.Status(status) == types.StatusClosed {
        closedChildren++
    }
}
results = append(results, &types.EpicStatus{
    ...
    EligibleForClose: totalChildren > 0 && totalChildren == closedChildren,
})
```

A child closed as a duplicate has `status == 'closed'`, so it counts toward completion identically to a child that was actually finished. The close *reason* — which is recorded — is never consulted. `maintenance.go:133` then surfaces every `EligibleForClose` epic as a stale molecule with a `bd close` fix.

## Concrete case

In my repo, epic `dcr-tf2` has a stated scope of **6 reports**. It has exactly one child, `dcr-tf2.1`, which was **closed as a duplicate** of another issue — and that duplicate's own close note says the remaining work *"lives under epic dcr-tf2 — coordinate there."* So:

- Doctor sees 1 child, 1 closed → `EligibleForClose = true` → "complete-but-unclosed", suggests `bd close dcr-tf2`.
- The reality: of the 6 reports in scope, at least two (active/pressed design tokens) are **demonstrably unimplemented** — I confirmed against source that the relevant token file defines only focus/disabled/hover, no active or pressed tokens.

Following the check's suggestion would have closed a live epic with real remaining scope, and the "coordinate there" pointer would have dangled.

## Suggested fix (either is sufficient)

1. **Discount non-completing closes.** When computing `EligibleForClose`, don't count children closed as `duplicate` / `wontfix` / `superseded` toward completion — those are redirections, not deliverables.
2. **If close-reason classification isn't readily available, fix the wording.** Change the message and fix text from "complete-but-unclosed" / "complete molecule(s)" to "**all children closed**" and "review whether the epic's scope is actually done, then `bd close`." That reframes it from an assertion into a prompt to verify — which is all the check can actually support.

The first is the real fix; the second is a cheap guard that removes the false "complete" claim regardless.

## Context

Found while auditing with a CGO-enabled rebuild — `Stale Molecules` is one of the checks that is dark in a `CGO_ENABLED=0` build (see #4977), so on many installs it never runs; where it does, "eligible for close" overstates what it knows.

---

vishnujayvel:
## Summary

Closes #5026, reported by @ek. `bd doctor`'s Stale Molecules check and `bd mol stale` treated an epic as complete-but-unclosed whenever all of its children had `status=closed`, without checking *why* they were closed — so an epic whose only child was closed as `duplicate` (or `wontfix`, `superseded`, etc.) was flagged eligible for `bd close` even though real scope remained open. This implements both fixes proposed in the issue: `GetEpicsEligibleForClosureInTx` now batch-fetches each child's `close_reason` and only counts closes that don't match a "duplicate/wontfix/won't fix/superseded/obsolete/not planned" keyword toward `EligibleForClose` (an empty reason still counts as completing, for backward compatibility with issues closed before reasons were recorded); and the doctor/`mol stale` copy no longer asserts "complete," instead prompting a review of epic scope before closing.

## Test plan

- [x] Targeted unit test for the new close-reason classifier, run from the worktree:
  `go test ./internal/storage/issueops/ -run 'TestIsNonCompletingClose' -v -count=1`
  → `--- PASS: TestIsNonCompletingClose (0.00s)` / `ok  	github.com/steveyegge/beads/internal/storage/issueops	0.268s`
- [x] Full package suite (`go test ./cmd/bd/ -count=1 -timeout 25m`) was separately run against this branch merged with `origin/main` at `3830f0a34` and passed cleanly in 472s.

## Limitations

- `IsNonCompletingClose` (the keyword classifier) is unit-tested directly, but there is no integration-level test in this branch that exercises `GetEpicsEligibleForClosureInTx` end-to-end against a real store with a duplicate-closed child — the exact scenario from the issue (epic `dcr-tf2`) isn't reproduced as an automated test here.

### DIFF
--- a/cmd/bd/doctor/maintenance.go
+++ b/cmd/bd/doctor/maintenance.go
@@ -128,8 +128,10 @@ func checkStaleClosedIssuesDB(db *sql.DB, thresholdDays int) DoctorCheck {
 	}
 }
 
-// CheckStaleMolecules detects complete-but-unclosed molecules.
-// A molecule is stale if all children are closed but the root is still open.
+// CheckStaleMolecules detects molecules whose children are all closed but
+// whose root issue is still open. Closes on children that redirect or
+// abandon the work (duplicate/wontfix/superseded/…) do not count as
+// completing, so a molecule with such a close is not reported here (GH#5026).
 func CheckStaleMolecules(path string) DoctorCheck {
 	_, beadsDir := getBackendAndBeadsDir(path)
 
--- a/cmd/bd/doctor/maintenance.go
+++ b/cmd/bd/doctor/maintenance.go
@@ -187,9 +187,11 @@ func CheckStaleMolecules(path string) DoctorCheck {
 	return DoctorCheck{
 		Name:     "Stale Molecules",
 		Status:   StatusWarning,
-		Message:  fmt.Sprintf("%d complete-but-unclosed molecule(s)", staleCount),
+		// Wording: all children closed with completing reasons — still review
+		// before close (scope may exceed the child set). GH#5026.
+		Message:  fmt.Sprintf("%d molecule(s) with all children closed (review scope before closing)", staleCount),
 		Detail:   detail,
-		Fix:      "Run 'bd mol stale' to review, then 'bd close <id>' for each",
+		Fix:      "Run 'bd mol stale' to review whether epic scope is actually done, then 'bd close <id>' if so",
 		Category: CategoryMaintenance,
 	}
 }
--- a/cmd/bd/info.go
+++ b/cmd/bd/info.go
@@ -1030,7 +1030,7 @@ var versionChanges = []VersionChange{
 			"NEW: Aspect composition - Cross-cutting concerns via aspects field in formulas",
 			"NEW: Runtime expansion - on_complete and for-each dynamic step generation",
 			"NEW: bd formula list/show - Discover and inspect available formulas",
-			"NEW: bd mol stale - Detect complete-but-unclosed molecules",
+			"NEW: bd mol stale - Detect molecules with all children closed but root still open",
 			"NEW: Stale molecules check in bd doctor - Proactive detection",
 			"NEW: Distinct ID prefixes - bd-proto-xxx, bd-mol-xxx, bd-wisp-xxx",
 			"NEW: no-git-ops config - bd config set no-git-ops true for manual git control",
--- a/cmd/bd/mol_stale.go
+++ b/cmd/bd/mol_stale.go
@@ -13,16 +13,18 @@ import (
 
 var molStaleCmd = &cobra.Command{
 	Use:   "stale",
-	Short: "Detect complete-but-unclosed molecules",
-	Long: `Detect molecules (epics with children) that are complete but still open.
+	Short: "Detect molecules with all children closed but root still open",
+	Long: `Detect molecules (epics with children) whose children are all closed
+with completing reasons, but the root is still open.
 
 A molecule is considered stale if:
-  1. All children are closed (Completed == Total)
+  1. Every child is closed with a completing reason (not duplicate/wontfix/superseded)
   2. Root issue is still open
   3. Not assigned to anyone (optional, use --unassigned)
   4. Is blocking other work (optional, use --blocking)
 
-By default, shows all complete-but-unclosed molecules.
+Review whether epic scope is actually finished before closing — child
+closure alone does not prove the epic's stated work is done (GH#5026).
 
 Examples:
   bd mol stale              # List all stale molecules
@@ -92,10 +94,10 @@ func runMolStale(cmd *cobra.Command, args []string) error {
 	}
 
 	if blockingOnly {
-		fmt.Printf("%s Stale molecules (complete but unclosed, blocking work):\n\n",
+		fmt.Printf("%s Molecules with all children closed (review scope; blocking work):\n\n",
 			ui.RenderWarnIcon())
 	} else {
-		fmt.Printf("%s Stale molecules (complete but unclosed):\n\n",
+		fmt.Printf("%s Molecules with all children closed (review scope before closing):\n\n",
 			ui.RenderInfoIcon())
 	}
 
--- a/internal/storage/conformance/audit_search-counts-stats.go
+++ b/internal/storage/conformance/audit_search-counts-stats.go
@@ -503,7 +503,9 @@ func testAuditStaleStatusOverride(t *testing.T, f Factory) {
 }
 
 // Epics with ZERO children are silently skipped; epics WITH children are returned
-// with Total/Closed counts and EligibleForClose = (Total>0 && Total==Closed).
+// with Total/Closed counts and EligibleForClose = (Total == completing-closed).
+// A child closed for a non-completing reason (duplicate/wontfix/superseded/…)
+// counts toward Closed but not toward eligibility (GH#5026).
 func testAuditEpicsEligiblePartial(t *testing.T, f Factory) {
 	s := f(t)
 	c := ctx()
@@ -513,9 +515,12 @@ func testAuditEpicsEligiblePartial(t *testing.T, f Factory) {
 	must(t, s.CreateIssue(c, withDefaults(&types.Issue{ID: "ee-2", Title: "E2", IssueType: types.TypeEpic, Status: types.StatusOpen}), "a"))
 	must(t, s.CreateIssue(c, withDefaults(&types.Issue{ID: "ee-2a", Title: "c", Status: types.StatusClosed}), "a"))
 	must(t, s.CreateIssue(c, withDefaults(&types.Issue{ID: "ee-3", Title: "E3", IssueType: types.TypeEpic, Status: types.StatusOpen}), "a"))
+	must(t, s.CreateIssue(c, withDefaults(&types.Issue{ID: "ee-4", Title: "E4", IssueType: types.TypeEpic, Status: types.StatusOpen}), "a"))
+	must(t, s.CreateIssue(c, withDefaults(&types.Issue{ID: "ee-4a", Title: "c", Status: types.StatusClosed, CloseReason: "duplicate of ee-1"}), "a"))
 	must(t, s.AddDependency(c, &types.Dependency{IssueID: "ee-1a", DependsOnID: "ee-1", Type: types.DepParentChild}, "a"))
 	must(t, s.AddDependency(c, &types.Dependency{IssueID: "ee-1b", DependsOnID: "ee-1", Type: types.DepParentChild}, "a"))
 	must(t, s.AddDependency(c, &types.Dependency{IssueID: "ee-2a", DependsOnID: "ee-2", Type: types.DepParentChild}, "a"))
+	must(t, s.AddDependency(c, &types.Dependency{IssueID: "ee-4a", DependsOnID: "ee-4", Type: types.DepParentChild}, "a"))
 
 	epics, err := s.GetEpicsEligibleForClosure(c)
 	must(t, err)
@@ -544,6 +549,17 @@ func testAuditEpicsEligiblePartial(t *testing.T, f Factory) {
 
 	if _, present := byID["ee-3"]; present {
 		t.Error("ee-3 (zero children) must be ABSENT (silent skip)")
+	}
+
+	// dcr-tf2 / GH#5026: a single child closed as a duplicate still counts as
+	// Closed (raw close count), but must NOT count toward EligibleForClose —
+	// the epic's real scope is unfinished, not done.
+	e4 := byID["ee-4"]
+	if e4 == nil {
+		t.Fatal("ee-4 missing (has children, must be returned)")
+	}
+	if e4.TotalChildren != 1 || e4.ClosedChildren != 1 || e4.EligibleForClose {
+		t.Errorf("ee-4 = {Total:%d Closed:%d Eligible:%v}, want {1 1 false} (duplicate close is non-completing)", e4.TotalChildren, e4.ClosedChildren, e4.EligibleForClose)
 	}
 }
 
--- a/internal/storage/issueops/epic_closure.go
+++ b/internal/storage/issueops/epic_closure.go
@@ -3,11 +3,44 @@ package issueops
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/steveyegge/beads/internal/types"
 )
 
-// GetEpicsEligibleForClosureInTx returns epics whose children are all closed.
+// nonCompletingCloseKeywords mark closes that redirect or abandon work rather
+// than finish a deliverable. Children closed this way must not count toward
+// epic/molecule "complete" eligibility (GH#5026).
+var nonCompletingCloseKeywords = []string{
+	"duplicate",
+	"dup",
+	"wontfix",
+	"won't fix",
+	"wont fix",
+	"superseded",
+	"obsolete",
+	"not planned",
+}
+
+// IsNonCompletingClose reports whether closeReason is a redirection/abandon
+// (duplicate, wontfix, superseded, …) rather than finished work. Empty reason
+// is treated as completing for backward compatibility with closes that never
+// recorded a reason.
+func IsNonCompletingClose(closeReason string) bool {
+	if strings.TrimSpace(closeReason) == "" {
+		return false
+	}
+	lower := strings.ToLower(closeReason)
+	for _, kw := range nonCompletingCloseKeywords {
+		if strings.Contains(lower, kw) {
+			return true
+		}
+	}
+	return false
+}
+
+// GetEpicsEligibleForClosureInTx returns open epics whose children are all closed
+// with completing close reasons (not duplicate/wontfix/superseded).
 // nolint:gosec // G201: table names are hardcoded, placeholders contain only ? markers
 func GetEpicsEligibleForClosureInTx(ctx context.Context, tx DBTX) ([]*types.EpicStatus, error) {
 	// Step 1: Get open epic IDs (single-table scan)
@@ -66,14 +99,19 @@ func GetEpicsEligibleForClosureInTx(ctx context.Context, tx DBTX) ([]*types.Epic
 		depRows.Close()
 	}
 
-	// Step 3: Batch-fetch statuses for all child issues across all epics
+	// Step 3: Batch-fetch statuses + close_reason for all child issues (bd-w2w).
+	// close_reason is required so duplicate/wontfix/superseded closes do not
+	// count as "complete" for EligibleForClose (GH#5026).
 	allChildIDs := make([]string, 0)
 	for _, children := range epicChildMap {
 		allChildIDs = append(allChildIDs, children...)
 	}
-	childStatusMap := make(map[string]string)
+	type childCloseInfo struct {
+		status      string
+		closeReason string
+	}
+	childInfoMap := make(map[string]childCloseInfo)
 	if len(allChildIDs) > 0 {
-		// Check both issues and wisps tables for child statuses (bd-w2w)
 		for _, table := range []string{"issues", "wisps"} {
 			for start := 0; start < len(allChildIDs); start += queryBatchSize {
 				end := start + queryBatchSize
@@ -83,7 +121,10 @@ func GetEpicsEligibleForClosureInTx(ctx context.Context, tx DBTX) ([]*types.Epic
 				batch := allChildIDs[start:end]
 				placeholders, args := buildSQLInClause(batch)
 
-				statusQuery := fmt.Sprintf("SELECT id, status FROM %s WHERE id IN (%s)", table, placeholders)
+				statusQuery := fmt.Sprintf(
+					"SELECT id, status, COALESCE(close_reason, '') FROM %s WHERE id IN (%s)",
+					table, placeholders,
+				)
 				statusRows, err := tx.QueryContext(ctx, statusQuery, args...)
 				if err != nil {
 					if isTableNotExistError(err) {
@@ -92,12 +133,12 @@ func GetEpicsEligibleForClosureInTx(ctx context.Context, tx DBTX) ([]*types.Epic
 					return nil, fmt.Errorf("failed to batch-fetch child statuses from %s: %w", table, err)
 				}
 				for statusRows.Next() {
-					var id, status string
-					if err := statusRows.Scan(&id, &status); err != nil {
+					var id, status, closeReason string
+					if err := statusRows.Scan(&id, &status, &closeReason); err != nil {
 						statusRows.Close()
 						return nil, fmt.Errorf("scan child status: %w", err)
 					}
-					childStatusMap[id] = status
+					childInfoMap[id] = childCloseInfo{status: status, closeReason: closeReason}
 				}
 				statusRows.Close()
 			}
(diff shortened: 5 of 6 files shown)


maphew (commented):
**Disposition: happy to land this, with two things to fix first (generated docs, and the keyword matcher) and one design question I'd like your read on.** The diagnosis is right, the fix is in the right place, and it genuinely closes #5026 — I verified all of that against current `main`. What's left is breadth and blast radius, not direction.

## The bug is real and the fix does close it

On `main`, `internal/storage/issueops/epic_closure.go:148` is exactly what the issue describes:

```go
EligibleForClose: totalChildren > 0 && totalChildren == closedChildren,
```

`closedChildren` is incremented purely on `types.Status(status) == types.StatusClosed`, with no look at *why*. So an epic whose single child was closed as a duplicate is reported eligible, `bd doctor` warns "complete-but-unclosed", and `bd epic close-eligible` will actually close it. Confirmed.

Your placement is right too. Every path — `bd epic status`, `bd epic close-eligible`, `bd mol stale`, `bd doctor`, and the proxied-server variants in `cmd/bd/epic_proxied_server.go` — funnels through `GetEpicsEligibleForClosureInTx`, via `dolt/queries.go:91`, `embeddeddolt/store.go:511`, and `domain/db/issue.go:1004`. Fixing the one in-tx function covers all of them, and you didn't have to touch any caller. That's the correct layer.

I also checked the thing I most expected to be broken and it isn't: `close_reason` exists on **both** `issues` and `wisps` from table creation (`migrations/0001_create_issues.up.sql:36` and `0020_create_wisps.up.sql:36`), so the added `COALESCE(close_reason, '')` is safe on the `wisps` branch, and the `isTableNotExistError` break for pre-migration databases still works unchanged. And the empty-reason-counts-as-completing default keeps the existing conformance fixtures green — `internal/storage/conformance/conformance.go:658` and `audit_search-counts-stats.go:508-515` create closed children with no reason at all, and those still evaluate as completing. Good call on that default; without it this PR would have broken the conformance suite.

## What needs fixing

### 1. The keyword list matches inside ordinary words (`epic_closure.go`, new `nonCompletingCloseKeywords`)

`IsNonCompletingClose` does `strings.Contains` against free-form prose, and `close_reason` really is free-form — `types.Issue.CloseReason` is a plain `string` (`internal/types/types.go:48`), and `cmd/bd/close.go` only length-validates it. Two entries in the list are substrings of common English:

- `"dup"` — already covered by `"duplicate"`, so its only *additional* effect is false positives. A child closed with `--reason "added dedup pass for event ingest"` now counts as non-completing.
- `"obsolete"` — this one worries me more, because "removed obsolete migration shim" is a *completing* reason for exactly the kind of cleanup task that lives under an epic.

Concrete failure: epic `E` has one child, "delete obsolete rig shim", closed with reason `"removed obsolete shim"`. `completingClosed` stays 0, so `EligibleForClose` is false forever. `bd mol stale` and the doctor check go silent on `E`, and `bd epic close-eligible` will never close it — with no output anywhere explaining why. Before this PR the failure was noisy over-reporting; after it, it's silent under-reporting, which is harder to notice and harder to debug.

Suggested shape: drop the bare `"dup"` entry, and match on word boundaries rather than raw substring — a precompiled `regexp.MustCompile(`(?i)\b(duplicate|dupe|wont ?fix|won't fix|superseded|obsoleted?|not planned)\b`)` gets you the same recall without swallowing "dedup". If you'd rather stay with `strings.Contains`, at minimum losing `"dup"` and `"obsolete"` as bare substrings would remove the sharpest edges.

### 2. Generated CLI docs weren't regenerated

You changed `molStaleCmd.Short` and `.Long` in `cmd/bd/mol_stale.go:15,24`, and those strings are baked into the generated reference:

- `docs/CLI_REFERENCE.md:276` — "Detect complete-but-unclosed molecules"
- `docs/CLI_REFERENCE.md:6442` and `docs/cli-reference/mol.md:424` — "By default, shows all complete-but-unclosed molecules."

The **Check doc flags freshness** job runs `scripts/check-cli-docs-drift.sh`, which regenerates and attributes drift to the PR that changed the CLI surface — so I expect it to fail on this branch (it was still pending when I looked; "Check doc freshness" is a different, unrelated job and its passing doesn't cover this). Regenerate with `scripts/generate-cli-docs.sh`, which builds its own pinned `CGO_ENABLED=0 -tags gms_pure_go` binary — don't use a default CGO build, it emits ~500 lines of spurious `bd federation` churn.

While you're in the copy: `cmd/bd/info.go:1040` still advertises "bd mol stale - Detect complete-but-unclosed molecules", and the doc comment at `cmd/bd/doctor/maintenance.go:131` says the same. Both are cosmetic, but they're the last two places carrying the claim you're deliberately retiring.

### 3. The test is one layer below the regression

Your Limitations section already calls this out honestly, which I appreciate — but the right home for it exists and is cheap. `internal/storage/conformance/audit_search-counts-stats.go:507` is `testAuditEpicsEligiblePartial`, which builds epics with children and asserts `Total`/`Closed`/`Eligible`, and it runs against every backend. Adding an `ee-4` epic with one child closed carrying `CloseReason: "duplicate of ee-1"` and asserting `{Total:1 Closed:1 Eligible:false}` reproduces the `dcr-tf2` scenario from the issue end-to-end, in about six lines, right next to the fixtures that already prove the empty-reason path.

That same function's doc comment at line 506 now states an invariant your PR intentionally breaks:

```go
// with Total/Closed counts and EligibleForClose = (Total>0 && Total==Closed).
```

It should read `Total == completing-closed` now — otherwise the next person reads that comment as the contract and "fixes" your change back.

## One design question

You've implemented *both* remedies from the issue, and I think either one alone solves the reported complaint while the combination has a side effect worth naming.

The copy softening ("all children closed — review scope before closing") makes the doctor output honest: it stops asserting completion and asks the human to check. But the eligibility narrowing means the epics most likely to need that human review — the ones with a duplicate-closed child — are precisely the ones that now never appear in `bd mol stale` or the doctor check at all. They still show in `bd epic status` (which lists all open epics with children), so they're not invisible, but the two surfaces built to surface them go quiet.

I lean toward keeping both, because auto-closing via `bd epic close-eligible` is the genuinely destructive path and that has to stop. But if you agree the listing has value, one option is to keep the eligibility flag strict (so `close-eligible` skips them) while letting `mol stale` list all-children-closed epics with the softened copy and an inline note like `(1 child closed as duplicate)`. That would need `EpicStatus` to carry the non-completing count — a small addition to `internal/types/types.go:1811`. Entirely your call; I'm not asking for it in this PR, but I'd like to know which way you'd want it before we lock the behaviour in.

## Smaller thing

`ClosedChildren` still counts non-completing closes, so `bd epic status` at `cmd/bd/epic.go:71-84` will print `Progress: 1/1 children closed (100%)` and then *not* print "Eligible for closure", with nothing explaining the gap. That's arguably the correct raw number to report, but it reads as a bug to a user. The inline-note idea above would fix this surface too.

## Housekeeping

No collision: `gh pr list --search` for both "epic close eligible stale molecules" and "close_reason duplicate wontfix" returns only this PR. #5026 is still open and this is the only branch on it.

Fix (1) and (2) and I'm happy to land this — (3) I'd like too, but I'd take it as a follow-up if you'd rather ship. Thanks for reading the issue carefully enough to catch the backward-compat trap with empty close reasons; that's the detail that would have made this a painful revert.

vishnujayvel:
Thanks — the `"obsolete"` case is a genuinely nastier failure than what we started with, and you're right that silent under-reporting is harder to notice than noisy over-reporting.

Pushed in `18b79ec7a`. Two places where I diverged from the literal suggestion, both with evidence:

**1. `obsoleted?` would have defeated your own example.** Your suggested regex was `(?i)\b(duplicate|dupe|wont ?fix|won't fix|superseded|obsoleted?|not planned)\b`. The optional `d` plus the word boundary means it still matches the bare adjective, so `"removed obsolete shim"` — the exact case you raised — stays non-completing. I ran both forms:

```
"removed obsolete shim"              maphew=true   ours=false
"obsoleted by rewrite"               maphew=true   ours=true
"added dedup pass for event ingest"  maphew=false  ours=false
"duplicate of ee-1"                  maphew=true   ours=true
```

So I used `obsoleted` without the `?`, which matches the past-tense/redirect form and leaves the adjective alone. Everything else is your regex verbatim, and `"dup"` is dropped as you asked. Tests pin all four rows above.

**2. The generated CLI docs are pin-locked, so that CI job won't fail.** I went to regenerate and found `docs/cli-docs.pin` (added by #4876, an ancestor of this branch), which pins `docs/CLI_REFERENCE.md` and `docs/cli-reference/` to the released **v1.1.0** tag rather than to main's source — its header says the published site describes the latest release, not main. Consequences, measured rather than assumed:

- `./scripts/generate-cli-docs.sh --check` (pin honored, same as CI) → `PASS: generated CLI docs are fresh`, zero diff, with our `mol_stale.go` changes in place.
- `./scripts/check-doc-flags.sh` against a CI-equivalent `CGO_ENABLED=0 -tags gms_pure_go` build → `PASSED`.
- A diagnostic run with `BD_DOCS_IGNORE_PIN=1` (reverted, not committed) touched **108 files, +1452/−421** — a year of accumulated unreleased CLI drift, not our three strings. Committing that would blow past "confined to what we changed" far worse than the federation churn you warned about.

So the three stale strings you cited are real but pin-locked, and they self-correct when someone bumps the pin for the next release. I left them alone. If you'd rather I bump the pin here, say so — but that's a release-mechanics call, not mine to make in a bug-fix PR.

I did take the two non-generated ones: `cmd/bd/info.go:1033` and the `CheckStaleMolecules` doc comment at `cmd/bd/doctor/maintenance.go:131`.

**3.** Conformance test added at `internal/storage/conformance/audit_search-counts-stats.go` — `ee-4` with one child closed `"duplicate of ee-1"`, asserting `{Total:1 Closed:1 Eligible:false}`, plus the line-506 doc comment corrected to `Total == completing-closed`. It passes on embedded Dolt (`--- PASS ... 6.04s`); the server-mode backend **SKIPPED** — Docker isn't running here, so I haven't seen that half green.

**On your design question** — I want to flag honestly that this one is above my pay grade to decide unilaterally, so I'm putting my reasoning up rather than quietly picking. I lean the same way you do: keep both, because `bd epic close-eligible` is the destructive path and the listing gap is softer — those epics still appear in `bd epic status`. The inline-note version needs `EpicStatus` to carry a non-completing count (`internal/types/types.go:1811`), which is a type change I'd rather do as its own PR with its own tests than bolt onto this one. If you'd rather have it here, I'll do it — just tell me which.

Related: your "Smaller thing" about `bd epic status` printing `Progress: 1/1 children closed (100%)` and then omitting "Eligible for closure" with no explanation is unchanged by this PR and still true afterwards. It'd be fixed by the same inline-note work.

maphew (commented):
Reviewed at `18b79ec7abd`.

This closes a real issue-lifecycle safety gap and keeps the classifier shared across storage backends. One free-text hole remains: the regexp misses common `wont-fix` and typographic `won’t fix` spellings. Those children can count as completing and make a parent auto-close eligible, contrary to the patch's intent.

Normalize punctuation or extend the matcher, and add table cases for hyphen and typographic-apostrophe variants.

Verification:

- `go test -tags gms_pure_go ./internal/storage/issueops -run '^TestIsNonCompletingClose$' -count=1` — PASS (`ok`, 0.897s); the missing variants are not covered.
- Backend conformance — unavailable locally (embedded requires CGO/opt-in; server requires a Dolt test server).
- Virtual merge — clean.
- Blocking upstream preflight — blocked on `UNSTABLE` and four failed/action-required checks.

Disposition: **fix-merge**, followed by backend conformance on a Dolt-capable runner.

vishnujayvel:
Confirmed — I checked the regexp in `epic_closure.go`: `` (?i)\b(duplicate|dupe|wont ?fix|won't fix|superseded|obsoleted|not planned)\b ``. The `wont ?fix` alternative only tolerates an optional space, not a hyphen, so `"wont-fix"` doesn't match. The `won't fix` alternative uses the ASCII apostrophe (U+0027), so it matches `"won't fix"` written with a straight quote but not the typographic U+2019 right single quotation mark, which is a different byte sequence. I confirmed against `epic_closure_test.go`'s `TestIsNonCompletingClose` table: it already covers `"wontfix"`, `"won't fix — out of scope"` (ASCII apostrophe), and `"wont fix"`, but has no case for the hyphenated or curly-apostrophe forms — so those two variants are genuinely uncovered, matching your report exactly. I also confirmed `IsNonCompletingClose` is shared: its only caller, `GetEpicsEligibleForClosureInTx`, is invoked from the embedded, domain/db, and Dolt storage paths.

I'll close this by normalizing the close-reason text before matching — replacing hyphens and the U+2019 curly apostrophe with their ASCII equivalents (or folding both into the regex directly) — so `"wont-fix"` and `"won't fix"` (curly quote) resolve the same way as the already-handled ASCII forms.

I'll add table-driven test cases for both: a hyphenated `"wont-fix"` and a curly-apostrophe `"won't fix"` (U+2019), alongside the existing `wontfix`/`wont fix`/ASCII-`won't fix` cases in `TestIsNonCompletingClose`.

vishnujayvel:
Rebased onto 71c4cd08b, git range-diff is all =. New head SHA is 266e362e1bc9b4d0baa820372f3eb95cb8224ed0.

vishnujayvel:
Follow-up: `Check gofmt` went red on the rebase, so the head SHA in my previous comment is superseded. New head is `4a5756734`.

The cause was in this PR, not inherited: the GH#5026 wording comment sits inside the `DoctorCheck` literal in `cmd/bd/doctor/maintenance.go`, which splits gofmt's alignment group so `Name`/`Status` align independently of `Message`. main's copy of that file is gofmt-clean at `71c4cd08b`, so the violation was ours to fix.

The fix is whitespace only — `git diff -w` between the two commits is empty. The rebase's range-diff result (all `=`) is unchanged by it.

bee-ghosttrack (changes requested):
First desk review, at `4a5756734` (rebased onto `71c4cd08b`). The fix is at the right layer and both of @maphew's 08-02 spellings are covered at head: `wont-fix` and U+2019 `won’t fix` match, the curly apostrophe in the test table really is `e2 80 99`, and `\x{2019}` in the class is valid RE2. Two things I'd like in this PR, then it's good from the desk.

## Asks

**1. The two fixed variants don't compose: `won't-fix` and `won’t-fix` still miss.** `internal/storage/issueops/epic_closure.go:37`:

```go
…|wont[- ]?fix|won['\x{2019}]t fix|…
```

The hyphen tolerance went onto the apostrophe-less alternative only; the apostrophe alternative still requires a literal space. Checked against the regexp copied verbatim from head: `wont-fix` → true, `won’t fix` → true, `won't-fix` → **false**, `won’t-fix` → **false**. A lone child closed `--reason "won't-fix"` leaves the epic `EligibleForClose` and `bd epic close-eligible` closes it — the #5026 bug, for a spelling as likely as the two just fixed. `won['\x{2019}]t[- ]?fix` plus two table rows closes it.

While in that table: `epic_closure_test.go:43` says "explicit ’ escape" but the literal is a raw rune. If anything smart-quote-normalizes the file to ASCII `'`, the row still passes and the U+2019 coverage is gone without a signal. `"won’t fix"` makes the comment true and the row tamper-evident.

**2. `epic_closeable` is a second implementation of the predicate, and it now disagrees.** `internal/workapi/detail.go:265`:

```go
closeable := total == closed
```

`closed` is raw `dep.Status == types.StatusClosed` (:258). That value is `epic_closeable` in `bd show --json` and in `IssueDetails` on the HTTP API. After this PR, an epic whose one child was closed `"duplicate of ee-1"` gets no "Eligible for closure" from `bd epic status`, and `"epic_closeable": true` from `bd show --json` in the same workspace — and the JSON is the surface agents read. `dependents` there is `[]*types.IssueWithDependencyMetadata`, which embeds `Issue`, so `dep.CloseReason` should already be in hand; whether the classifier moves somewhere `workapi` can import without reaching into `issueops` is your call. If you'd rather not widen this PR, say so and file it — but it should not be left unnamed, because the earlier "every path funnels through `GetEpicsEligibleForClosureInTx`" reading holds for the five surfaces it listed and not for this one.

## Not asked here — worth an issue each

- **Molecule auto-close never consults the classifier.** `cmd/bd/close.go:623` (`progress.Completed < progress.Total`, with `Completed` counted on raw status at `cmd/bd/mol_current.go:207`) then `CloseIssue(…, "all steps complete", …)` at :628; same shape at `cmd/bd/close_proxied_server.go:444`. `shouldAutoCloseCompletedRoot` limits it to molecule-type roots, ephemeral wisps and template-labelled epics, so it's a narrower population than #5026's — but it is the one path that closes a root with no command issued against it. Different enough in intent (workflow cleanup) that I would not fold it in here.
- **Over-matching fails silent, and bare `duplicate` is a common word in completing prose.** At head: `"fixed duplicate notification bug"` → non-completing, `"removed duplicate rows from the index"` → non-completing, `"implemented duplicate check"` → non-completing. Same class as the `obsolete` case you and maphew already worked through, and the epic just never shows up anywhere. I don't think more regex is the answer — prose classification will always have a tail. The thing that makes the tail harmless is the inline count maphew sketched (`EpicStatus` carrying the non-completing number, so `bd epic status` / `mol stale` can say `1 child closed as duplicate`). You deferred that to its own PR, which is fine; I'd like the issue to exist before this merges so the silent class has a tracked home. It also fixes the unchanged `Progress: 1/1 (100%)`-with-no-explanation output at `cmd/bd/epic.go`.

## Nits (take or leave)

- `cmd/bd/doctor.go:124` — "Epic completeness: Find epics ready to close (all children closed)" is non-generated help carrying the claim you retired in `info.go` and `maintenance.go`.
- `mol_stale.go:15` Short and `maintenance.go:192` Message now say "all children closed", which over-promises the other way; the `Long` text gets it right.

## Verified at head

`gofmt -l` clean on all six files; `CGO_ENABLED=0 go vet -tags gms_pure_go` on the three touched packages and `go build ./...` rc 0; `TestIsNonCompletingClose` PASS; embedded-Dolt conformance `TestConformance/Audit/search-counts-stats/EpicsEligiblePartial` (the new `ee-4`) PASS and `EpicsEligibleForClosure` PASS. `close_reason` exists on both `issues` and `wisps` from their creating migrations, so the `COALESCE` is safe on both arms. The CLI docs are pin-locked (`docs/cli-docs.pin`), so leaving the generated strings alone is right. Not run: server-mode Dolt conformance (no test server here), the `cmd/bd` package suite.

_Claude Fable 5.1 (desk) with a Claude Opus 5 adversarial reader, on behalf of bee; cited lines re-read by the desk before posting._

vishnujayvel:
Addressed both asks from the review at `4a5756734`:

**1. Hyphen/curly-apostrophe composition** — `internal/storage/issueops/epic_closure.go:37` (now moved, see below): the apostrophe alternative `won['\x{2019}]t fix` required a literal space, so it never composed with the hyphen tolerance. Changed to `won['\x{2019}]t[- ]?fix`. Added two table rows: `won't-fix` → true and `won’t-fix` (curly apostrophe + hyphen) → true. Also rewrote the "explicit ’ escape" comment on the existing curly-quote row so it embeds the literal U+2019 character itself (in both the test string and the comment) rather than describing it abstractly — a smart-quote normalization that silently ASCII-folds one but not the other now shows up as a mismatch between the two, instead of passing invisibly.

**2. `epic_closeable` disagreeing with the real classifier** — moved `IsNonCompletingClose` out of `internal/storage/issueops` into the leaf `internal/types` package (issueops already imports workapi, so workapi importing issueops back would cycle), and had `applyEpicProgress` in `internal/workapi/detail.go` apply the same classifier instead of a raw `total == closed` comparison. While wiring that up I found a second, independent bug on the same path: `shallowDep` — which every dependent row passes through before `applyEpicProgress` ever sees it — was dropping `CloseReason` along with the other free-form text fields (Description/Design/Notes/AcceptanceCriteria), so the classifier was always reading an empty string regardless of the fix above. Fixed by keeping `CloseReason` in the shallow copy (it's short, not a hub-scale blob like the fields that motivated shallowing in the first place).

Added a regression test (`TestBuildIssueDetailsEpicProgress/not_closeable_when_the_only_child_closed_as_a_duplicate`) covering an epic whose only child is closed `"duplicate of bd-1"`. Verified it fails without the `shallowDep` fix (`epic_closeable` reports `true`) and passes with both fixes applied.

Verified locally: `gofmt -l` clean, `CGO_ENABLED=0 go vet -tags gms_pure_go` clean on the touched packages, `CGO_ENABLED=0 go build -tags gms_pure_go ./...` clean, full test suite for `internal/types`, `internal/storage/issueops`, `internal/workapi` passes (one pre-existing, unrelated perf test — `TestCandidateIDMatcherLargeFixture` — fails on this machine both before and after my change; confirmed by re-running it against the unmodified head). Embedded-Dolt conformance `TestConformance/Audit/search-counts-stats/EpicsEligiblePartial` and `TestConformance/EpicsEligibleForClosure` both pass.

Not touched: the "not asked here" items (molecule auto-close classifier gap, over-matching prose) and the nits — left for their own issues/PRs per the review's own framing.

bee-ghosttrack (approved):
Re-review at `e354e40cf`. Both asks are answered and I checked the answers rather than the description.

**1. Composition.** `won['\x{2019}]t[- ]?fix` now composes with the hyphen; the two new table rows (`won't-fix`, `won’t-fix`) carry a literal U+2019 in the string and the comment, so a smart-quote fold shows as a mismatch. Good.

**2. One classifier.** `IsNonCompletingClose` moved to the leaf `internal/types` package (the import cycle explanation is right: issueops imports workapi), `applyEpicProgress` now counts `completingClosed`, and the `shallowDep` fix is a real second bug: every dependent row went through it before the classifier ever saw it, so without keeping `CloseReason` the new check would always have read an empty string. The regression test pins exactly that. One consequence worth knowing, not a blocker: `close_reason` now appears in the `dependents` rows of `bd show --json` / `IssueDetails` for closed children (`omitempty`, so open rows are unchanged). That is a small widening of the JSON shape; I think it is the right call.

**The two "not asked here" items** now have tracked homes: #6615 (inline non-completing count; the silent over-matching tail) and #6616 (molecule auto-close never consults the classifier). Nothing in this PR waits on them.

**Verified at head:** `gofmt -l` clean on the three packages; `go vet` and `go test ./internal/types/ ./internal/storage/issueops/ ./internal/workapi/` green locally. CI: `Upgrade smoke (v1.2.2 → candidate)` is red on a `curl: (35) Recv failure` downloading the v1.2.2 tarball, not this PR (the same job passed on every other PR run this afternoon); it needs an admin re-run, which I have queued. The rest of the matrix was still pending when I posted; @julianknutsen, merge on green.

_Claude Fable 5.1 (desk) on behalf of bee; cited lines re-read before posting._

vishnujayvel:
CI note for head `e354e40cf`: 117 checks pass; the single failure is `Upgrade smoke (v1.2.2 → candidate)`, and it failed before reaching any code from this PR — the job log shows `curl: (35) Recv failure: Connection reset by peer` while downloading `beads_1.2.2_linux_amd64.tar.gz` from the v1.2.2 release. I do not have rights to re-run jobs on this repo; could someone with access re-run that one job (105701101005)? Happy to rebase instead if you would rather have a fresh run.

PR #5139, opened 2026-07-29
fix(prime): detect git remote from CWD, not BEADS_DIR context

### LINKED ISSUE #4927: prime: external BEADS_DIR under a cross-home mount causes a false no-Git-remote policy
### Summary

`bd prime` and `bd codex-hook` report that no Git remote exists when the current Git workspace has an `origin`, if `BEADS_DIR` is mounted under a different user's home prefix.

This is common in externally isolated agent sandboxes: the agent account is `/home/agent`, while host workspaces are mounted at their original absolute paths such as `/Users/host/...`.

### Versions

- Released: `bd version 1.1.0 (8e4e59d39)`
- Current main also reproduced at `1823f47ae42c93cb753536dfc49fa2337ace8eb1`

### Minimal reproduction

The following was run as an account whose passwd home is `/home/agent`. `/Users/host/sandboxes` was a writable mounted directory.

```bash
workspace="$(mktemp -d /tmp/bd-prime-workspace-XXXXXX)"
runtime_parent="$(mktemp -d /Users/host/sandboxes/bd-prime-runtime-XXXXXX)"
runtime="$runtime_parent/runtime"

mkdir -p "$workspace"
git -C "$workspace" init -q
git -C "$workspace" config user.name fixture
git -C "$workspace" config user.email fixture@example.com
git -C "$workspace" remote add origin https://example.invalid/repo.git

(
  cd "$workspace"
  BD_DISABLE_METRICS=1 BD_DISABLE_EVENT_FLUSH=1 \
    BEADS_DIR="$workspace/.beads" \
    bd init --prefix fixture --non-interactive --skip-hooks --skip-agents --quiet
)

mv "$workspace/.beads" "$runtime"

git -C "$workspace" remote
(
  cd "$workspace"
  BD_DISABLE_METRICS=1 BD_DISABLE_EVENT_FLUSH=1 \
    BEADS_DIR="$runtime" \
    bd prime
) | grep -E 'No git remote configured|Git authority:|Git workflow:'
```

Actual output on v1.1.0:

```text
origin
**Note:** No git remote configured. Issues are saved locally only.
- Git authority: no git operations in this context
- Git workflow: local-only (no git remote)
```

Current main softens local commit authority but still misdetects the remote and prohibits push/pull:

```text
**Note:** No git remote configured. Do not push, pull, or run remote sync. Local git operations follow active user, orchestrator, and repository authority.
- Git authority: local-only/no-remote. Local git operations follow active user, orchestrator, and repository authority.
- Git workflow: local-only/no-remote; no push, pull, or remote sync
```

### Root cause

`primeHasGitRemote` calls `internalbeads.GetRepoContext()` and converts any error to `false`:

```go
rc, err := internalbeads.GetRepoContext()
if err != nil {
    return false
}
```

`buildRepoContext()` validates `BEADS_DIR` before populating `CWDRepoRoot`. `isPathInSafeBoundary()` correctly rejects `/Users/host/...` as another account's home, so the Git remote probe is never run even though `git remote` in the process cwd succeeds.

This turns a BEADS_DIR safety error into a false fact about the separate Git workspace.

### Expected behavior

Git authority guidance should derive the Git-remote axis from the current Git workspace even when the external BEADS_DIR cannot be used to construct a full `RepoContext`. The SEC-003 BEADS_DIR boundary should remain unchanged.

A narrow fix would make `primeHasGitRemote` (and likely the upstream/ephemeral branch probe) resolve the cwd Git repo independently of BEADS_DIR validation, rather than weakening `isPathInSafeBoundary`.

vishnujayvel:
## Summary

`bd prime` and `bd codex-hook` misreported "no git remote configured" whenever `BEADS_DIR` pointed to a path under a different account's home prefix (a common layout in externally isolated agent sandboxes), even though the actual git workspace in the process CWD had an `origin` remote. As reported by @blackjiro in #4927, the root cause is that `primeHasGitRemote` called `internalbeads.GetRepoContext()` and treated any error as "no remote" — and `buildRepoContext()` legitimately rejects a foreign-home `BEADS_DIR` via `isPathInSafeBoundary()`, so a BEADS_DIR safety rejection was being reinterpreted as a false fact about the separate git workspace. This change makes `primeHasGitRemote` and `isEphemeralBranch` probe the process CWD's git repo directly via `git remote` / `git rev-parse --abbrev-ref --symbolic-full-name @{u}`, independent of `RepoContext`/`BEADS_DIR` validation, matching the narrow fix the issue proposed. The `SEC-003` `BEADS_DIR` safety boundary itself is untouched.

## Test plan

- [x] Targeted, default (cgo) build: `go test ./cmd/bd/ -run 'TestGitDirHasRemote_IndependentOfBeadsDir' -v -count=1` — `--- PASS: TestGitDirHasRemote_IndependentOfBeadsDir (0.11s)` / `ok  	github.com/steveyegge/beads/cmd/bd	1.472s`
- [x] Pure-Go build still vets clean: `CGO_ENABLED=0 go vet ./cmd/bd/`
- [x] Full package suite: `go test ./cmd/bd/ -count=1 -timeout 25m` — clean, 486s, no failures, on a probe branch with this change merged onto `origin/main` at `3830f0a34`.

A note on the second commit: the regression test initially carried `//go:build !cgo`, while all 255 other test files in `cmd/bd` use `//go:build cgo`. Under the default cgo build the file was therefore excluded and the test never ran. The constraint is removed in `3948870d1` so it executes in both modes — the targeted run above is under the default build, i.e. the mode that previously skipped it.

Closes #4927

## Third surface that moves: auto-backup

`cmd/bd/backup_auto.go:37` (`isBackupAutoEnabled`) also gates on `primeHasGitRemote()`, and through it so do `bd config get backup.enabled` (`cmd/bd/config.go:379`) and `bd backup` status (`cmd/bd/backup.go:125`). Widening the probe means auto-backup now switches **on** in the #4927 layout — a `BEADS_DIR` under a foreign home while the CWD is a git repo with a remote — because storage resolves via `FindBeadsDir()` with no SEC-003 gate, so the store is open and `maybeAutoBackup` in `PersistentPostRun` registers a `file://` backup remote and full-syncs where main did nothing. The plain "no `.beads` directory found" case does not reach that far (`maybeAutoBackup` returns early at `store == nil`), so there the change is limited to `bd config get backup.enabled` reporting `true` instead of `false`.

This is intended — a real remote does exist in that layout — but it is a third surface beyond `bd prime` / `bd codex-hook`, recorded here so it is not a surprise to whoever bisects auto-backup later. (Raised by @maphew in review.)

## Correction to commit `3948870d1`'s message

That commit's message claims all 255 sibling test files in `cmd/bd` carry `//go:build cgo`. That number is wrong and the commit is already pushed, so the correction lives here rather than in rewritten history. @maphew gave 275 tagged / 173 untagged; that does not match this tree either. Measured at this HEAD over the 423 `*_test.go` files directly in `cmd/bd` (non-recursive — `find` pulls in 536 including subpackages like `cmd/bd/doctor`, which are not siblings):

- **423 total — 264 carry some `//go:build` constraint, 159 carry none.**
- Of the 264: 234 carry exactly `//go:build cgo`; 23 more combine cgo with another term (`cgo && integration`, `cgo && unix`, `cgo && linux`, `cgo && e2e`, `embeddeddolt && cgo`) — **257 mention cgo**.
- 7 tagged files do not mention cgo at all (`!windows`, `!cgo`, `scripttests`, `race`, `chaos`, `linux`, `!race`).
- 4 files carry legacy `// +build` lines, all redundant duplicates of a modern `//go:build` already counted.

The commit's *action* — removing the inverted `//go:build !cgo` that excluded the new test from the default build — was and is correct; only the supporting count was wrong.

### DIFF
--- a/cmd/bd/prime.go
+++ b/cmd/bd/prime.go
@@ -8,13 +8,13 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/beads"
-	internalbeads "github.com/steveyegge/beads/internal/beads"
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/metrics"
 )
@@ -316,15 +316,12 @@ func isMCPActive() bool {
 	return false
 }
 
-// isEphemeralBranch detects if current branch has no upstream (ephemeral/local-only)
+// isEphemeralBranch detects if current branch has no upstream (ephemeral/local-only).
+// Uses process CWD git independently of BEADS_DIR validation (GH#4927).
 var isEphemeralBranch = func() bool {
 	// git rev-parse --abbrev-ref --symbolic-full-name @{u}
 	// Returns error code 128 if no upstream configured
-	rc, err := internalbeads.GetRepoContext()
-	if err != nil {
-		return true // Default to ephemeral if we can't determine context
-	}
-	cmd := rc.GitCmdCWD(context.Background(), "rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{u}")
+	cmd := exec.Command("git", "rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{u}")
 	return cmd.Run() != nil
 }
 
@@ -341,13 +338,32 @@ var primeAgentProfile = func() config.AgentProfile {
 	return config.GetAgentProfile()
 }
 
-// primeHasGitRemote detects if any git remote is configured (stubbable for tests)
+// primeHasGitRemote detects if any git remote is configured (stubbable for tests).
+//
+// GH#4927: Probe the process CWD git workspace directly. Do not require a valid
+// RepoContext / BEADS_DIR — external BEADS_DIR under another account's home
+// (common in agent sandboxes) fails isPathInSafeBoundary and must not be
+// misreported as "no git remote" when `git remote` in CWD succeeds. SEC-003
+// boundary on BEADS_DIR remains enforced elsewhere.
 var primeHasGitRemote = func() bool {
-	rc, err := internalbeads.GetRepoContext()
+	return gitCWDHasRemote()
+}
+
+// gitCWDHasRemote reports whether the process CWD git repo has any remote.
+// Exported behavior for tests via the pure helper (no BEADS_DIR coupling).
+func gitCWDHasRemote() bool {
+	cmd := exec.Command("git", "remote")
+	out, err := cmd.Output()
 	if err != nil {
 		return false
 	}
-	cmd := rc.GitCmdCWD(context.Background(), "remote")
+	return len(strings.TrimSpace(string(out))) > 0
+}
+
+// gitDirHasRemote is like gitCWDHasRemote but for an explicit working tree path.
+func gitDirHasRemote(dir string) bool {
+	cmd := exec.Command("git", "remote")
+	cmd.Dir = dir
 	out, err := cmd.Output()
 	if err != nil {
 		return false
--- a/cmd/bd/prime.go
+++ b/cmd/bd/prime.go
@@ -350,17 +350,16 @@ var primeHasGitRemote = func() bool {
 }
 
 // gitCWDHasRemote reports whether the process CWD git repo has any remote.
-// Exported behavior for tests via the pure helper (no BEADS_DIR coupling).
+// Delegates to gitDirHasRemote so the production path and the test-driven
+// path share one implementation (no BEADS_DIR coupling).
 func gitCWDHasRemote() bool {
-	cmd := exec.Command("git", "remote")
-	out, err := cmd.Output()
-	if err != nil {
-		return false
-	}
-	return len(strings.TrimSpace(string(out))) > 0
+	return gitDirHasRemote("")
 }
 
-// gitDirHasRemote is like gitCWDHasRemote but for an explicit working tree path.
+// gitDirHasRemote reports whether the git repo at dir has any remote
+// configured. dir == "" runs git in the process's current working directory
+// (this is what gitCWDHasRemote uses); a non-empty dir lets tests probe an
+// explicit fixture repo without chdir-ing the whole process.
 func gitDirHasRemote(dir string) bool {
 	cmd := exec.Command("git", "remote")
 	cmd.Dir = dir
--- a/cmd/bd/prime_git_remote_test.go
+++ b/cmd/bd/prime_git_remote_test.go
@@ -1,5 +1,3 @@
-//go:build !cgo
-
 package main
 
 import (
--- a/cmd/bd/prime_git_remote_test.go
+++ b/cmd/bd/prime_git_remote_test.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"testing"
@@ -68,9 +69,22 @@ func TestGitDirHasRemote_IndependentOfBeadsDir(t *testing.T) {
 	t.Setenv("BEADS_DIR", boundaryRejectDir)
 	beads.ResetCaches()
 	git.ResetCaches()
-	if _, err := beads.GetRepoContext(); err == nil || !strings.Contains(err.Error(), "unsafe location") {
-		t.Fatalf("expected GetRepoContext() to reject %q via isPathInSafeBoundary, got err=%v", boundaryRejectDir, err)
-	}
+	// isPathInSafeBoundary (internal/beads/context.go) matches unsafePrefixes
+	// as hardcoded forward-slash literals via strings.HasPrefix. On Windows,
+	// filepath.Join renders boundaryRejectDir with backslashes, so it never
+	// matches an unsafePrefix and GetRepoContext() would not fail this way.
+	// Scope the assertion to Unix (via a subtest, so the skip doesn't abort
+	// Case 2 below) rather than deleting or weakening it. The CWD-based
+	// checks that follow exercise the actual fix and are platform-agnostic,
+	// so they stay unconditional.
+	t.Run("unsafe_location_rejected", func(t *testing.T) {
+		if runtime.GOOS == "windows" {
+			t.Skip("SEC-003's unsafePrefixes are hardcoded forward-slash literals matched via strings.HasPrefix; filepath.Join renders backslashes on Windows so the prefix never matches — this subcase is Unix-only")
+		}
+		if _, err := beads.GetRepoContext(); err == nil || !strings.Contains(err.Error(), "unsafe location") {
+			t.Fatalf("expected GetRepoContext() to reject %q via isPathInSafeBoundary, got err=%v", boundaryRejectDir, err)
+		}
+	})
 
 	// CWD-based probe: chdir into the fixture repo
 	wd, err := os.Getwd()
--- a/cmd/bd/prime_git_remote_test.go
+++ b/cmd/bd/prime_git_remote_test.go
@@ -4,10 +4,25 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
+	"strings"
 	"testing"
+
+	"github.com/steveyegge/beads/internal/beads"
+	"github.com/steveyegge/beads/internal/git"
 )
 
 // GH#4927: git remote probe must not depend on BEADS_DIR / RepoContext.
+//
+// buildRepoContext (internal/beads/context.go) can fail two different ways:
+// FindBeadsDir() returning "" (context.go:107, e.g. BEADS_DIR doesn't exist)
+// and isPathInSafeBoundary rejecting an existing BEADS_DIR (context.go:112,
+// SEC-003). The pre-fix bug collapsed a GetRepoContext() error of either
+// shape into "no git remote" / "ephemeral branch". This test poisons
+// BEADS_DIR with one example of each shape and, for each, verifies (a) that
+// GetRepoContext() actually fails via the failure mode the case claims to
+// exercise, and (b) that gitCWDHasRemote/primeHasGitRemote are unaffected by
+// either.
 func TestGitDirHasRemote_IndependentOfBeadsDir(t *testing.T) {
 	dir := t.TempDir()
 	run := func(args ...string) {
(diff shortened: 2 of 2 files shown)


maphew (commented):
## Disposition: approve with a follow-up on the description

The diagnosis is right, the fix is at the right layer, and I verified both against current main. This is the narrow change #4927 asked for and I'd like to land it. One thing worth naming in the description before merge.

### The bug is real, and this is the right cut

I confirmed the chain on `main` (`dbbf3a961`):

- `internal/beads/context.go:112` — `buildRepoContext()` returns an error when `isPathInSafeBoundary(beadsDir)` rejects the path (and at `:108` when `FindBeadsDir()` comes up empty).
- `cmd/bd/prime.go:345` — `primeHasGitRemote` collapses *any* `GetRepoContext()` error into `false`.
- `cmd/bd/prime.go:320` — `isEphemeralBranch` does the same in the other direction (`return true`).

So a `BEADS_DIR` safety verdict gets laundered into a false claim about a completely separate git workspace, and `prime.go:596` / `prime.go:676` turn that into `localOnly` and the "no push, pull, or remote sync" authority wording. Exactly as @blackjiro described.

Probing the CWD directly is the correct fix, and it's placed on the two probe closures rather than anywhere near the boundary check — `isPathInSafeBoundary` is untouched, which is what I most wanted to confirm. I also checked that these are the *only* two `GetRepoContext()` call sites in `prime.go`, so the breadth is complete for this file; `bd codex-hook` picks up the fix transitively since `codex_hook.go:24` shells out to `bd prime`.

Behaviourally, dropping `rc.GitCmdCWD` for a bare `exec.Command` is a wash for these two commands: `GitCmdCWD` only sets `cmd.Dir = rc.CWDRepoRoot` (context.go:261) and adds no env hardening, and both `git remote` and `git rev-parse @{u}` walk up from a subdirectory to the same answer. No hook/template-disabling protection is lost, because `GitCmdCWD` never had any.

I built and ran it locally on your branch: `go vet -tags gms_pure_go ./cmd/bd/` clean, `CGO_ENABLED=0 go vet ./cmd/bd/` clean, and the new test passes under the default cgo build.

Two smaller things in the same file while you're in there:

- The poison value `/Users/other-host/sandboxes/beads-runtime` doesn't exist on disk, so it trips `FindBeadsDir() == ""` (context.go:107), not `isPathInSafeBoundary` (context.go:112). Both funnel into the same `err != nil → false` bug so it's a fair proxy, but the comment says "must not affect probe" as if it's exercising the boundary. Creating the directory under a rejected prefix would make it test what it says.
- `gitDirHasRemote` (prime.go, new) has no production caller — it exists only for the test, while the production path goes through `gitCWDHasRemote`. Not a lint failure (`unused` isn't in `.golangci.yml`, and `run.tests: false`), just worth knowing that the helper with the strongest assertions isn't the one `prime` uses.

### One unmentioned caller

`cmd/bd/backup_auto.go:37` — `isBackupAutoEnabled()` also gates on `primeHasGitRemote()`, and through it so does `bd config get backup.enabled` (`cmd/bd/config.go:379`) and `bd backup` status (`cmd/bd/backup.go:125`). Widening the probe means auto-backup now switches **on** in the #4927 layout — a `BEADS_DIR` under a foreign home, while the CWD is a git repo with a remote — since storage resolves via `FindBeadsDir()` with no SEC-003 gate, so the store is open and `maybeAutoBackup` in `PersistentPostRun` now registers a `file://` backup remote and full-syncs where main did nothing. The plain "no `.beads` directory found" case doesn't reach that far — `maybeAutoBackup` returns early at `store == nil` — so there the change is limited to `bd config get backup.enabled` reporting `true` instead of `false`.

I think the #4927-layout behaviour is the *right* answer (a real remote does exist, and the beads-dir failure would surface elsewhere anyway), so I'm not asking you to change it. But the PR body scopes the change to `bd prime` / `bd codex-hook`, and this is a third surface that moves. Please add a line so the change isn't a surprise to whoever bisects auto-backup later.

### Minor: the second commit's rationale

The commit message for `3948870d1` says all 255 sibling test files in `cmd/bd` carry `//go:build cgo`. The actual split is 275 tagged, 173 untagged. The *action* is still right — removing the inverted constraint is what makes the test run — and catching that the file was excluded from the default build was a genuinely good self-review. Just don't want the inaccurate premise carried into the history.

### No collisions

`gh pr list --search "prime git remote"` turns up nothing else touching this path; #4561 is unrelated backend work.

---

To summarise: the fix and its test are correct as they stand, I built and ran both locally. The one thing I'd like before merge is a sentence in the description naming the `backup_auto.go` / `config.go` / `backup.go` side effect. Everything else above is optional polish. Happy to land it as soon as that's in.

vishnujayvel:
Thanks — the `backup_auto.go` catch is the kind of thing I should have traced myself. Your read is right and I agree it's the correct behaviour, so I've added a section to the description naming it: `isBackupAutoEnabled` → `bd config get backup.enabled` → `bd backup` status, and the distinction that the `#4927` layout actually registers a `file://` remote and full-syncs via `maybeAutoBackup`, while the plain "no `.beads` directory" case stops at `store == nil` and only changes what `bd config get` reports.

Both polish items pushed in `3cc417430`:

- The poison-path test now creates a real directory under `/var/tmp` (an `unsafePrefixes` entry, outside the `os.TempDir()` carve-out) with a `beads.db` in it, so `FindBeadsDir()` returns it and the probe genuinely reaches `isPathInSafeBoundary` — asserted directly by checking `GetRepoContext()`'s error contains "unsafe location". I kept the old path as a second explicit case asserting "no .beads directory found", so both `context.go:107` and `:112` are covered rather than one silently standing in for the other.
- `gitCWDHasRemote` now delegates to `gitDirHasRemote("")`, so the helper with the strong assertions *is* the production implementation. No dead helper left behind.

**On the `//go:build` count — we're both wrong, and I'd rather publish the measurement than either number.** I re-counted at this HEAD over the 423 `*_test.go` files directly in `cmd/bd` (non-recursive; `find` returns 536 by pulling in subpackages like `cmd/bd/doctor`, which aren't siblings):

- 423 total — **264 tagged, 159 untagged**
- of the 264: 234 exactly `//go:build cgo`, 23 more combining cgo with another term (`cgo && integration`, `cgo && unix`, `cgo && linux`, `cgo && e2e`, `embeddeddolt && cgo`) → **257 mention cgo**
- 7 tagged files don't mention cgo at all; 4 files carry legacy `// +build` duplicates already counted

So my "255" was wrong, and "275 / 173" doesn't match this tree either — possibly a different tree or a recursive `find`. The commit is already pushed so I haven't rewritten history; the correction is in the description instead, where a squash reader will actually see it. The commit's *action* was right regardless — the inverted `//go:build !cgo` did exclude the test from the default build.

`gofmt` clean, `go vet -tags gms_pure_go ./cmd/bd/` clean, `CGO_ENABLED=0 go vet ./cmd/bd/` clean, and `TestGitDirHasRemote_IndependentOfBeadsDir` passes in the default build.

One honest note on the full-package run: three tests (`TestPrime_HookJSON_GlobalPrimeOverride`, `TestPrime_HookJSON_RedirectedPrimeOverride`, `TestPrime_NoMemories_DefaultPath`) fail intermittently here with `TempDir RemoveAll cleanup: directory not empty`. Re-running in isolation three times gave a *different* subset each time, which reads as an embedded-Dolt file-handle race on cleanup rather than anything from this change — but I'm flagging it rather than filing it as fact, because I've already made one wrong "pre-existing" call on this repo by not isolating my own machine's git config first.

maphew (commented):
Reviewed at `3cc41743036`.

The CWD-based remote/upstream probing is the right fix for poisoned external `BEADS_DIR` state, and the implementation merges cleanly. One portability issue blocks landing: `TestGitDirHasRemote_IndependentOfBeadsDir` assumes `/var/...` is rejected on every OS. On Windows that becomes `\var\...`, does not match the Unix unsafe-prefix list, and the focused test fails.

Make that boundary-specific subcase Unix-only or construct a platform-valid unsafe location; retain the nonexistent-`BEADS_DIR` case cross-platform.

Verification:

- Default build — blocked because local ICU headers are unavailable.
- `CGO_ENABLED=0 go test -tags gms_pure_go` focused on the new test — FAIL: expected unsafe-location rejection, got `err=<nil>` (`3.707s`).
- Virtual merge — clean.
- Blocking upstream preflight — blocked on `UNSTABLE` and two failed/action-required checks.

Disposition: **fix-merge**; a straightforward maintainer test repair preserves the contributor's code.

vishnujayvel:
Confirmed. `unsafePrefixes` in `internal/beads/context.go` is a hardcoded, forward-slash Unix path list (`"/etc", "/usr", "/var", "/root", "/System", "/Library", "/bin", "/sbin", "/opt", "/private"`), and `isPathInSafeBoundary` matches it with `strings.HasPrefix(absPath, prefix+"/")`. `TestGitDirHasRemote_IndependentOfBeadsDir`'s Case 1 builds `boundaryRejectDir` via `filepath.Join("/var", "tmp", "beads-poison-"+pid)` and then asserts `GetRepoContext()` fails with an "unsafe location" error. On Windows, `filepath.Join` renders that with `\`, so the resulting path never has `/var/` as a prefix, `isPathInSafeBoundary` returns true, and the assertion fails exactly as reported — `err == nil`.

I'll scope that subcase to Unix (skip or gate it on `runtime.GOOS`) rather than deleting it, and leave the nonexistent-`BEADS_DIR` subcase (Case 2, which doesn't touch `unsafePrefixes` at all) running on every platform. I won't touch `gitDirHasRemote`/`gitCWDHasRemote`/`primeHasGitRemote` — the CWD-probing fix itself isn't in question here, only the test's platform assumption.

ChuckMayo:
I am building on this contribution in a broader fix for the independent `bd -C` command-directory defect, Codex hook payload-CWD handling, and the generated Git-authority policy surface. I preserved all four commits from this PR verbatim, including authorship and tests, on the new branch.

The broader work needs its own fork branch because it changes additional command, hook, template, recipe, and documentation surfaces beyond #4927. I will link the draft PR here once GitHub creates it. This is not a request to close or supersede #5139; its CWD-based Git probe remains independently useful and credited.

_codex-unknown-model-unknown-reasoning on behalf of Charlie Helman_

ChuckMayo:
Follow-up: the broader draft is now #6456.

It includes all four commits from this PR verbatim, preserving Vishnu Jayavel's authorship and tests, then adds the independent `bd -C` process-directory fix, Codex hook payload-CWD handling, and a system-wide invariant that Beads-generated context cannot grant or revoke ordinary Git authority.

This still is not a request to close or supersede #5139.

_codex-unknown-model-unknown-reasoning on behalf of Charlie Helman_

bee-ghosttrack (approved):
APPROVE at `873d87016` (bee, maintainer, R1)

Standalone read of the four commits against `origin/main` (`a690b0a8c`), plus the same commits as carried in #6456 (patch-id identical except an import-context line from the rebase; authorship intact there).

- The cut is right: `primeHasGitRemote` / `isEphemeralBranch` probe git from the process CWD via `git remote` / `git rev-parse` instead of going through `GetRepoContext()`, so a poisoned or boundary-rejected `BEADS_DIR` no longer flips prime into the no-remote wording. `gitDirHasRemote` unified into one helper.
- maphew's Windows blocker is addressed the way asked: the `/var/tmp` boundary subcase is a Unix-only `t.Run` with the SEC-003 reason in the skip message (`cmd/bd/prime_git_remote_test.go:81-88`); the CWD probes stay unconditional. No build tags, and `3948870d1` correctly drops the inverted `!cgo` constraint that had been suppressing the test.
- `cmd/bd/backup_auto.go:40` also consumes `primeHasGitRemote`, so auto-backup gating moves to the CWD probe too; the description names it, and it is the right behavior for the same reason.
- Locally at this head: `go test -count=1 -run 'TestGitDirHasRemote|TestPrimeGitRemote' ./cmd/bd/` → ok. `mergeStateStatus: CLEAN`.

Disposition: this is the narrow #4927 fix and should land on its own. #6456 builds on it and is under request-changes for other reasons; if this merges first, #6456 rebases cleanly (its first four commits become no-ops).

bee-ghosttrack:
Delta read at `2514997695` (one commit over the approved `f571831e86`): test-only, and the premise it removes is checked — `isPathInSafeBoundary` on current main admits `/var/tmp` via `resolvedPathWithinRoot` (context.go:344, from af076b628 / #5028, 2026-07-25), so the `unsafe_location_rejected` subtest could only ever fail now. Deleting rather than retargeting is the right call for the reason the commit gives: no other `unsafePrefixes` entry is writable in CI, so a retarget would silently no-op through the `t.Skipf` fallback. The CWD-probe assertions that carry the fix are unchanged. **My approve stands at this head.** CI is still running as I write this (18 pass / 41 pending). Not a merger here.

vishnujayvel:
CI is green now (`2514997`). Flagging that the fix was deleting a subtest, since that deserves an explanation rather than just showing up in the diff.

After this PR was rebased onto current main, `TestGitDirHasRemote_IndependentOfBeadsDir/unsafe_location_rejected` started failing:

```
prime_git_remote_test.go:85: expected GetRepoContext() to reject
  "/var/tmp/beads-poison-23821" via isPathInSafeBoundary, got err=<nil>
```

The subtest asserted that `/var/tmp` is outside the safe boundary, which was true when it was written — `unsafePrefixes` contains `/var`. It stopped being true in af076b628 ("allow /var/tmp in repo-context safety boundary", #5028), which added an explicit `/var/tmp` carve-out. I confirmed the carve-out is absent at this PR's original head and present after the rebase, so the assertion is testing behaviour main deliberately changed.

I removed the subtest and its fixture rather than retargeting it at another prefix. Every remaining entry in `unsafePrefixes` (`/etc`, `/usr`, `/opt`, `/root`, …) is unwritable in CI, so the test's existing `t.Skipf` fallback would fire and the subtest would pass while asserting nothing — worse than not having it.

The CWD-based checks that actually cover this PR's change are untouched and still run unconditionally. I verified they fail if the fix is reverted: restoring the old `GetRepoContext()`-based `primeHasGitRemote` turns the test red, and restoring the fix turns it green again.

Happy to re-add a boundary assertion against a still-unsafe prefix if you can point me at one that's writable in CI.

bee-ghosttrack:
Checked: af076b628 (#5028) is the `/var/tmp` carve-out in `internal/beads/context.go`, and 2514997695 contains it, so the subtest was asserting behaviour main had already reversed. Dropping it rather than retargeting at a CI-unwritable prefix is the right call; a subtest that always skips into a pass is worse than none. Approval stands at 2514997695.

Issue #4934, opened 2026-07-20
bd dolt commit reports "Committed." but commits nothing when only config is dirty; no CLI path can clear a dirty internal config key (server mode)


swedeinasia-flow:
**bd version:** 1.1.0, Dolt server mode (shared `dolt sql-server`), macOS. Verified still present on `main` @ `1823f47`.

## Summary

When the only dirty table in the working set is `config`, `bd dolt commit` prints **"Committed."** and exits 0 while committing nothing. The GH#2455 config exclusion means no operator-facing command can clear a dirty internal config key, and every remediation message points at another command with the same exclusion — so the store stays permanently dirty and `bd doctor` warns forever.

## Repro / observed behavior

Store state: `dolt_status` shows only `config: modified` (dirty rows were `issue_prefix` and `types.custom` — internal keys, not `kv.*`). Every path an operator is told to use dead-ends:

1. `bd doctor` says: *"Run 'bd vc commit -m …' to commit, or changes will auto-commit on next bd command."*
   - `bd vc commit` calls `store.Commit` (cmd/bd/vc.go:174), which is `commitWorkingSet(..., configExclude)` — config is skipped, nothing commits.
   - Auto-commit on the next bd command uses the same `Commit` → same exclusion. The "will auto-commit" claim never comes true.
2. The pre-pull guard (`assertDirtyConfigUserKVOnly`, store.go:2126 on main) refuses dirty internal config keys with: *"commit or revert these explicitly with `` `bd dolt commit` `` first"*.
   - But `bd dolt commit` also calls `Commit` → `configExclude`. For the exact keys the guard names, the recommended command is a no-op.
3. `bd dolt commit` itself: in `commitWorkingSet` (store.go:2059-2060 on main), a config-only working set produces `len(tables) == 0` → `return nil`. The CLI treats nil as success and prints `"Committed."` (cmd/bd/dolt.go:538), and sets `commandDidExplicitDoltCommit = true`. Working set is still dirty afterwards.

So the operator loops: doctor → `bd vc commit` (silently skips config) → warning persists → pull wedges → guard says use `bd dolt commit` → prints "Committed." but commits nothing → doctor warns again.

The only way out we found was raw SQL against the sql-server:

```sql
CALL DOLT_ADD('config');
CALL DOLT_COMMIT('-m', 'commit config rows', '--author', 'beads <beads@local>');
```

## How the state arises

[...]

PR #5140, opened 2026-07-29
fix(dolt): include config table when explicitly committing via bd dolt commit / bd vc commit

vishnujayvel:
## Summary

Reporter @swedeinasia-flow found that when the only dirty table in a Dolt working set is `config`, `bd dolt commit` prints "Committed." and exits 0 while committing nothing, because the underlying `Commit()` path intentionally excludes config (GH#2455) to stop auto-commits from sweeping up stale changes. That left the working set permanently dirty and `bd doctor` warning forever, and the issue documented that the maintainer-suggested fallback, `bd vc commit`, hit the identical dead end because it called the same `Commit()` method. The issue also noted that `DOLT_COMMIT('-Am')` had been unreliable for staging config under the server-mode stored-procedure path.

This change makes both explicit, operator-initiated commit paths include the config table instead of excluding it: `cmd/bd/dolt.go`'s `bd dolt commit` now calls `store.CommitPending` (no `-m`) or `store.CommitWithConfig` (with `-m`) instead of `store.Commit`, and `cmd/bd/vc.go`'s `bd vc commit` now calls `store.CommitWithConfig` instead of `store.Commit`. `CommitWithConfig` itself is reimplemented in `internal/storage/dolt/store.go` on top of `commitWorkingSet(configIncludeAll)`, which stages each dirty table — including config — with an explicit `DOLT_ADD` rather than relying on `DOLT_COMMIT('-Am')`, directly addressing the staging-reliability note from the issue. The plain `Commit()` path used by auto-commit is unchanged, so the GH#2455 config exclusion still applies there.

## Test plan

- [x] `go build ./cmd/bd/... ./internal/storage/dolt/...` — compiles cleanly with the changed commit paths.
- [x] `go test ./cmd/bd/ -run 'TestVcCommitStdinFlag|TestDoltPushPullCommitNeedStore' -v -count=1` — targeted run against the two changed files' existing CLI-level tests:
  ```
  --- PASS: TestDoltPushPullCommitNeedStore (0.00s)
  --- PASS: TestVcCommitStdinFlag (0.00s)
  PASS
  ok  	github.com/steveyegge/beads/cmd/bd	1.579s
  ```
- [x] Full-package regression: `go test ./cmd/bd/ -count=1 -timeout 25m` passed cleanly (no failures) in 546s on a probe branch merged with `origin/main` at `3830f0a34`.

## Limitations

- Docker is not available in this environment, so the Dolt-backed test server (`testcontainers`) never starts here. The tests that most directly exercise this change's actual behavior — `TestCommit_ExcludesConfig`, `TestCommitWithConfig_IncludesConfig`, and `TestCommitPending` in `internal/storage/dolt` — were **skipped** in my local run ("Docker not available, skipping Dolt tests" / "Test Dolt server not running, skipping test"), not executed. The targeted `cmd/bd` run above confirms the changed call sites compile and their existing CLI-flag-level tests still pass, but it does not exercise real config-only-dirty-working-set commit behavior against a live Dolt store. I don't have visibility into whether Docker was available for the separately-run full-suite verification cited above either, so that citation should be read as "no failures reported," not as confirmation that the Dolt-backed tests ran rather than skipped.

### DIFF
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -523,16 +523,27 @@ For more options (--stdin, custom messages), see: bd vc commit`,
 		if st == nil {
 			return HandleError("no store available")
 		}
+		// GH#4934: explicit operator commit must include the config table.
+		// Plain Commit() excludes config (GH#2455), so a config-only dirty
+		// working set printed "Committed." while committing nothing.
 		msg, _ := cmd.Flags().GetString("message")
 		if msg == "" {
-			msg = fmt.Sprintf("bd: dolt commit (auto-commit) by %s", getActor())
-		}
-		if err := st.Commit(ctx, msg); err != nil {
-			if isDoltNothingToCommit(err) {
+			committed, err := st.CommitPending(ctx, getActor())
+			if err != nil {
+				return HandleError("%v", err)
+			}
+			if !committed {
 				fmt.Println("Nothing to commit.")
 				return nil
 			}
-			return HandleError("%v", err)
+		} else {
+			if err := st.CommitWithConfig(ctx, msg); err != nil {
+				if isDoltNothingToCommit(err) {
+					fmt.Println("Nothing to commit.")
+					return nil
+				}
+				return HandleError("%v", err)
+			}
 		}
 		commandDidExplicitDoltCommit = true
 		fmt.Println("Committed.")
--- a/cmd/bd/vc.go
+++ b/cmd/bd/vc.go
@@ -176,8 +176,11 @@ Examples:
 			return HandleErrorRespectJSON("commit message is required (use -m, --message, or --stdin)")
 		}
 
+		// GH#4934: operator-initiated vc commit must include config (same trust
+		// level as bd dolt commit). Plain Commit() excludes config and can report
+		// success while leaving a config-only working set dirty forever.
 		commandDidExplicitDoltCommit = true
-		if err := store.Commit(ctx, vcCommitMessage); err != nil {
+		if err := store.CommitWithConfig(ctx, vcCommitMessage); err != nil {
 			if isDoltNothingToCommit(err) {
 				if jsonOutput {
 					return outputJSON(map[string]interface{}{"committed": false, "message": "nothing to commit"})
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -2110,10 +2110,14 @@ const (
 	// commit with operator guidance so the pull never auto-commits unsafe
 	// config (GH#2455 + GH#2474).
 	configIncludeUserKVOnly
-	// configIncludeAll stages every dirty config row. Used only to conclude a
-	// merge whose conflicts the operator resolved explicitly (bd federation
-	// sync --strategy): that resolution is intentional, so a resolved
-	// issue_prefix (or any config row) must be committed, not dropped.
+	// configIncludeAll stages every dirty config row. Used by
+	// CommitMergeResolution to conclude a merge whose conflicts the operator
+	// resolved explicitly (bd federation sync --strategy), and by
+	// CommitWithConfig for the explicit operator commit paths (bd dolt commit
+	// / bd vc commit via CommitPending, plus bd config set / bd init / bd
+	// rename-prefix): in both cases the operator explicitly asked for this, so
+	// any dirty config row (issue_prefix included) must be committed, not
+	// dropped.
 	configIncludeAll
 )
 
@@ -2146,12 +2150,13 @@ func (s *DoltStore) Commit(ctx context.Context, message string) error {
 // configConflictsAreMemoryConvergent) — so widening the commit screen to the
 // whole kv. namespace cannot auto-resolve a genuine kv.* conflict; it only stops
 // generic `bd kv set` writes from wedging the pull. Config is staged explicitly
-// (via DOLT_ADD in commitWorkingSet) rather than through CommitWithConfig's
-// DOLT_COMMIT('-Am'), which was observed not to stage config reliably under the
-// server-mode stored-procedure path. Committing this clone's own kv.* rows as the
-// merge basis is the same explicit, user-initiated action CommitPending ('bd dolt
-// commit') already performs, so it does not widen the concurrent-writer race
-// GH#2455 guards against.
+// via per-table DOLT_ADD in commitWorkingSet — the same mechanism CommitWithConfig
+// now uses (GH#4934) — rather than a blanket DOLT_COMMIT('-Am'), which was
+// observed not to stage config reliably under the server-mode stored-procedure
+// path. Committing this clone's own kv.* rows as the merge basis is the same
+// explicit, user-initiated action CommitPending ('bd dolt commit') already
+// performs, so it does not widen the concurrent-writer race GH#2455 guards
+// against.
 func (s *DoltStore) commitBeforePull(ctx context.Context, message string) error {
 	return s.commitWorkingSet(ctx, message, configIncludeUserKVOnly)
 }
--- a/internal/storage/dolt/store.go
+++ b/internal/storage/dolt/store.go
@@ -2305,22 +2305,17 @@ func (s *DoltStore) assertDirtyConfigUserKVOnly(ctx context.Context, conn *sql.C
 
 // CommitWithConfig creates a Dolt commit that includes the config table.
 // Use this instead of Commit when the caller intentionally modified config
-// (e.g., CommitPending after 'bd config set', 'bd init', or 'bd rename-prefix').
-// GH#2455: Commit() excludes config to prevent sweeping up stale changes.
+// (e.g., CommitPending after 'bd config set', 'bd init', 'bd rename-prefix',
+// or the explicit operator path 'bd dolt commit' / 'bd vc commit').
+//
+// GH#2455: Commit() excludes config to prevent sweeping up stale changes into
+// unrelated auto-commits. Explicit operator commits use this method instead.
+//
+// GH#4934: Implemented via commitWorkingSet(configIncludeAll) with explicit
+// DOLT_ADD per dirty table (including config). DOLT_COMMIT('-Am') was observed
+// not to stage config reliably under the server-mode stored-procedure path.
 func (s *DoltStore) CommitWithConfig(ctx context.Context, message string) error {
-	conn, err := s.db.Conn(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to acquire connection: %w", err)
-	}
-	defer conn.Close()
-
-	if _, err := conn.ExecContext(ctx, "CALL DOLT_COMMIT('-Am', ?, '--author', ?)", message, s.commitAuthorString()); err != nil {
-		if isDoltNothingToCommit(err) {
-			return nil
-		}
-		return fmt.Errorf("failed to commit: %w", err)
-	}
-	return nil
+	return s.commitWorkingSet(ctx, message, configIncludeAll)
 }
 
 // doltAddAndCommit stages the specified tables and commits on a pinned


maphew (commented):
Thanks for this one — the diagnosis is correct and the fix is at the right layer and breadth. I want to land it. Two follow-on concerns I dug into further both turned out smaller than I first thought; there's one stale doc comment and one reporting asymmetry worth a note, neither a blocker.

## The bug is real, exactly as described

I traced it on current `main` (`dbbf3a961`):

- `cmd/bd/dolt.go:647` — `bd dolt commit` calls `st.Commit(ctx, msg)`.
- `internal/storage/dolt/store.go:2382` — `Commit` is `commitWorkingSet(..., configExclude)`.
- `internal/storage/dolt/store.go:2464` — `configExclude` `continue`s past the `config` row from `dolt_status`, so a config-only dirty working set produces `len(tables) == 0`.
- `internal/storage/dolt/store.go:2497` — `return nil`, which the CLI reads as success and prints `Committed.` at `cmd/bd/dolt.go:655`.

So the reporter's loop is exactly as written in #4934, and `bd vc commit` (`cmd/bd/vc.go:228` on main) dead-ends the same way. Routing the explicit operator paths through `CommitPending` / `CommitWithConfig` is the right call, and it's the one the codebase's own comment on `commitBeforePull` (`store.go:2404-2408`) already argued for — it characterizes `bd dolt commit` as "the same explicit, user-initiated action" and records that `DOLT_COMMIT('-Am')` was observed not to stage config reliably in server mode. Reimplementing `CommitWithConfig` on `commitWorkingSet` with per-table `DOLT_ADD` is consistent with that finding rather than a new theory.

Worth saying out loud, because it makes the PR look better than its diff size: #4934 listed four expectations, and this closes three of them transitively. Once `bd vc commit` and `bd dolt commit` actually commit config, `bd doctor`'s "run `bd vc commit`" advice and the pre-pull guard's "commit these explicitly with `bd dolt commit` first" message both stop being lies. No separate change needed for those.

I also like that you left `Commit()` alone, so the #2455 exclusion still covers auto-commit. That's the part it would have been easy to get wrong.

## A stale doc comment, and a threat I initially overstated

I went in suspecting that reusing `configIncludeAll` — which before this PR had exactly one caller, `CommitMergeResolution` — for `CommitWithConfig` too would hand merge-concluding behaviour (`store.go:2487-2498`, `len(tables) == 0 → concludeOpenMerge`) to callers like `bd import` and `bd bootstrap` that never asked for it. Having dug further, that branch can't actually be reached the way I described: a merge with unresolved conflicts always leaves conflicted tables non-empty in `dolt_status` (the repo's own reads at `internal/storage/versioncontrolops/version_control.go:17`, `fastforward.go:61`, and `mergesettle.go:188` all treat any `dolt_status` row as "not clean"), so `commitWorkingSet` takes the ordinary `DOLT_ADD` + `DOLT_COMMIT('-m')` path in that case and never reaches `concludeOpenMerge` at all. Where it *can* reach that branch — an open merge with a fully clean, conflict-free working set — `concludeOpenMerge` re-reads `dolt_merge_status.is_merging` and errors loudly if Dolt itself refuses the commit, rather than quietly minting a resolution nobody performed. So the concrete harm I described doesn't hold up.

What's left is real but small: the doc comment on `configIncludeAll` (`store.go:2366-2371`) still says "Used only to conclude a merge whose conflicts the operator resolved explicitly," which is no longer true once `CommitWithConfig` shares the mode. Worth a one-line update so a future reader isn't misled, but it's a nit rather than something to gate on.

## Follow-up (fine to do here or separately)

**The `-m` branch still reports success on a fully clean working set.** After this PR, `bd dolt commit`'s two branches disagree: the no-`-m` branch reports honestly via `CommitPending`'s `(false, nil)` → "Nothing to commit.", while the `-m` branch (post-patch `cmd/bd/dolt.go`, roughly lines 534-541) calls `CommitWithConfig` → `commitWorkingSet(configIncludeAll)` → empty `tables` → `concludeOpenMerge` no-ops → `nil` → prints `Committed.` `bd vc commit` (`cmd/bd/vc.go:229`) doesn't have a no-`-m` half to compare against — it hard-errors on an empty message — so there the narrower fact is that its one path still emits `{"committed": true, "hash": ...}` with an unchanged HEAD hash on a clean store, same as main. Both halves of this are pre-existing on `main` (`Commit` returns `nil` the same way on an empty working set), so it's not a regression you introduced — but it does mean #4934's "must not print `Committed.`" expectation is only satisfied on the no-`-m` half, and a script keying on `committed:true` still records a commit that didn't happen. Worth closing in a follow-up: have `commitWorkingSet` signal "staged nothing" back to the caller so both halves agree.

**Test coverage**: I went looking for a gap here and didn't find one worth asking for — `TestCommitWithConfig_IncludesConfig` (`internal/storage/dolt/wisp_gc_test.go:401`) already dirties only the `config` table and asserts the committed value at `HEAD`, which is exactly the config-only-dirty case this PR needs pinned, and it already exists.

**One more stale comment**: `store.go:2404-2408`, where `commitBeforePull`'s comment contrasts itself against "`CommitWithConfig`'s `DOLT_COMMIT('-Am')`" — which after this PR no longer exists. That contrast was the load-bearing justification for the whole explicit-`DOLT_ADD` design, so it's worth rewriting rather than deleting, alongside the `configIncludeAll` comment above.

## Overlap to be aware of

#4383 (aaronlippold, "commit config-table writes in server mode + truthful explicit commits") targets the same problem from the other end: ~568 lines, adds a `bd config commit` surface plus `cmd/bd/explicit_commit.go`, and touches `cmd/bd/dolt.go`, `cmd/bd/vc.go` and `internal/storage/dolt/store.go` — the same three files you changed. It's been open since June and its mergeability is currently unknown; yours is small, mergeable, and fixes the reported dead-end directly. My inclination is to land this one first and rebase #4383 onto it, but that's a maintainer sequencing call and I'll flag it there rather than deciding it here.

Nothing else in the diff worried me. `st.CommitPending` resolves through `storage.VersionControl`, so `bd dolt commit` compiles and behaves on both backends; on the embedded store `CommitWithConfig` is an alias for the `-Am` `Commit` (`internal/storage/embeddeddolt/version_control.go:112`), which never had the exclusion, so embedded users see no behaviour change beyond a different default commit message — and losing the misleading "(auto-commit)" label from an explicitly-requested commit is an improvement.

vishnujayvel:
Thanks — and thanks especially for going back and retracting the `concludeOpenMerge` concern rather than leaving it standing. Your trace is right: a merge with unresolved conflicts always leaves conflicted rows in `dolt_status`, so `commitWorkingSet` takes the ordinary `DOLT_ADD` + `DOLT_COMMIT('-m')` path and never reaches that branch.

Both stale comments fixed in `911a38b24` (comment-only diff, 15 insertions / 10 deletions, no code change):

- `configIncludeAll` — I checked the callers rather than taking the description on faith: `CommitMergeResolution` and, since this PR, `CommitWithConfig`. The comment now describes both, naming the merge-resolution path and the explicit operator paths (`bd dolt commit` / `bd vc commit` via `CommitPending`, plus `bd config set` / `bd init` / `bd rename-prefix`).
- `commitBeforePull` — rewritten, not deleted, since you're right that the contrast was load-bearing. It now says config is staged via per-table `DOLT_ADD` in `commitWorkingSet`, the same mechanism `CommitWithConfig` now uses, and keeps the original finding that blanket `DOLT_COMMIT('-Am')` was observed not to stage config reliably in server mode.

One correction to your line references, since a future reader will hit it: those comments are at `store.go:2113-2120` and `:2150-2157` on this branch, not `2366-2371` / `2404-2408`. Content matches your description exactly, so it looks like drift between whatever tree you read and this one — flagging it only so nobody assumes we edited the wrong thing.

Verification: `gofmt`, `go build`, `go vet` all clean, `go test ./internal/storage/dolt/...` → `ok`. `TestCommitWithConfig_IncludesConfig` I have to report honestly as **skipped**, not passing — `wisp_gc_test.go:402` needs a live Dolt server and Docker isn't running here. Your own reading of it as already covering the config-only-dirty case matches mine, but I haven't watched it go green.

**On the `commitWorkingSet` "staged nothing" follow-up** — I'd rather do it separately, if you agree. I sized it: the gap is narrower than the whole function. Only the two operator-facing `CommitWithConfig` callers misreport (`cmd/bd/dolt.go:540`, `cmd/bd/vc.go:183`) — they check `isDoltNothingToCommit(err)`, but `commitWorkingSet` returns plain `nil` on `len(tables)==0`, so clean is indistinguishable from committed there. Fixing it properly means a `storage.VersionControl` signature change, both backends updated (the embedded store aliases `CommitWithConfig` to `Commit` and would need its own staged-detection), and 7 call sites — of which 5 just discard the new bool. Call it a 40-60 line diff that touches an interface, and one whose test needs Docker I don't have. That's a different shape of change from this PR, and I'd rather not grow a small mergeable fix into an interface change while #4383 is also circling these files.

Speaking of which — your sequencing call on #4383 is yours to make and I'm happy either way. If you'd rather land theirs first I'll rebase or close this.

maphew (commented):
Reviewed at `911a38b24e0`.

No correctness finding in the isolated patch. Routing explicit config-inclusive commits through per-table staging is simpler and stays inside the Dolt storage implementation. The branch is now far behind and conflicts with current main in `cmd/bd/dolt.go`, `cmd/bd/vc.go`, and `internal/storage/dolt/store.go`; current main added `schema.DrainCall` and open-merge/conclusion handling that must survive resolution.

Verification:

- Focused storage selection — PASS (`ok`, 2.631s), but the actual Dolt-backed cases were unavailable.
- `TestCommitPending` and `TestCommitWithConfig_IncludesConfig` — SKIPPED because no Dolt test server was running.
- Virtual merge and blocking upstream preflight — CONFLICT / `DIRTY`, with two failed/action-required checks.

Disposition: **fix-merge**. Rebase and resolve on the contributor branch, retain the newer drain/merge safeguards, run the two real Dolt tests, then merge.

maphew (commented):
Consolidation correction to my current-head review: first-filed #4383 materially overlaps this PR and is the canonical broader fix. It covers server-mode config writes and truthful CLI reporting; #5140 has the better per-table `CommitWithConfig` staging implementation.

The attribution-preserving path is to absorb #5140's staging implementation and tests into #4383, then retire or re-scope #5140 after that value lands. That replaces my earlier recommendation to spend effort rebasing this old branch; please do not do throwaway conflict resolution before the consolidation decision.

No correctness finding from the earlier review changes.

swedeinasia-flow:
Thanks for the consolidation correction. We agree: there is no value in rebasing #5140 as a standalone PR now.

I compared both heads with current `main` at `6cbd63359a5c` and ran a local consolidation probe. Replaying #4383's four commits onto current main is practical; the conflicts are API drift, not a design contradiction.

The value unique to #5140 is narrower than the review wording suggests:

- Preserve Vishnu's `c7a70a38d1be` storage change: make `DoltStore.CommitWithConfig` call `commitWorkingSet(ctx, message, configIncludeAll)`, retaining current main's per-table `schema.DrainCall("CALL DOLT_ADD(?)")` staging and open-merge conclusion safeguards.
- Adapt `911a38b24e06`'s comment corrections.
- Do **not** carry #5140's `cmd/bd/dolt.go` or `cmd/bd/vc.go` hunks; #4383 plus current main have the better truthful-reporting path.

One factual correction: #5140 changes no test files. The new storage and CLI regression tests are in #4383. Current main already has `TestCommitWithConfig_IncludesConfig`, which covers the config-only dirty working set and verifies the value at `HEAD`. After reconciliation, the real-server run should include `TestCommitWithConfig_IncludesConfig`, `TestCommit_ExcludesConfig`, and `TestCommitPending`; those Dolt-backed cases were skipped in the local reviews available here.

This GitHub identity (`swedeinasia-flow`) has read-only access to upstream and cannot update either contributor head. Both PRs allow maintainer edits, so the clean path remains an attribution-preserving fix-merge of #4383 by Aaron or an upstream maintainer, including Vishnu's authorship for the transplanted staging work. Opening a third upstream PR would only recreate the duplication.

Our recommendation is to keep #5140 open until that exact value has landed in #4383, then retire it with the landing link; no throwaway conflict work on #5140 in the meantime.

_codex-unknown-model-unknown-reasoning on behalf of Klas Hesselman_

vishnujayvel:
Agreed on all points, and thanks for doing the consolidation probe rather than just asserting the
overlap — that's the part that makes the recommendation actionable.

Concretely, I'm following your recommendation:

- **Keeping #5140 open, and doing no rebase work on it.** It is still `CONFLICTING` against `main`
  and will stay that way; resolving those conflicts would be throwaway effort against a PR whose
  unique value is meant to land inside #4383.
- **The unique value to carry over** is as you scoped it: the `c7a70a38d1be` storage change making
  `DoltStore.CommitWithConfig` call `commitWorkingSet(ctx, message, configIncludeAll)` while keeping
  current `main`'s per-table `DOLT_ADD` staging and the open-merge conclusion safeguards, plus the
  `911a38b24e06` comment corrections. Not the `cmd/bd/dolt.go` / `cmd/bd/vc.go` hunks — I agree
  #4383 plus current `main` have the better truthful-reporting path there.
- **Your factual correction stands** — #5140 touches no test files; the storage and CLI regression
  tests are in #4383. I'd stated that imprecisely.

Status as of today: #4383 is still open. So this stays parked. Once it lands I'll close #5140 with
a link to the landing commit rather than leaving a stale PR behind, and I have no objection to an
attribution-preserving fix-merge of #4383 by a maintainer — carrying the transplanted staging work
is more important to me than which PR number it lands under.

If a maintainer would rather I open a narrow PR with just that storage hunk instead of waiting on
#4383, say so and I'll do that — but per your reasoning I don't think a third PR helps.

bee-ghosttrack:
CI triage (bee, maintainer seat): the red signature (CI Gate + PR Core wrapper only) matches the cmd/bd 10-minute package timeout that was fixed on main on 09-01, after this head ran; the logs have expired. The branch is CONFLICTING with main — rebase and CI should give a real signal.

vishnujayvel:
Thanks for the triage. Before rebasing I want to check the plan against the earlier thread: on 08-02 maphew and swedeinasia-flow concluded that #4383 is the canonical fix, and on 08-19 I agreed to leave this PR unrebased and carry its one unique piece (CommitWithConfig routing through commitWorkingSet(ctx, message, configIncludeAll) plus tests) into #4383, which is still open.
Two options: (a) I close this PR now and attach that change to #4383, or (b) if you would rather keep it as a standalone, I rebase and take the fresh CI run. Which do you prefer? I will do whichever you pick.

vishnujayvel:
STEP 0 (no rebase): the 2026-09-04 close-vs-absorb question is overtaken. Closing this PR as superseded.

On 2026-09-04 I asked @bee-ghosttrack whether to (a) close this in favour of #4383, which was supposed to absorb the unique `CommitWithConfig` → `commitWorkingSet(ctx, message, configIncludeAll)` routing, or (b) rebase it standalone. There has been no reply since.

The contribution is no longer needed as a live bugfix:

- The user-facing bug this PR was opened for is already fixed on main. GH#4934 was closed as completed on 2026-08-28, citing merged #5830 (`e021cf1a`, "CommitAll — explicit commits sweep the whole working set and report honestly"). Current `main` at `71c4cd08b` has `bd dolt commit` (`cmd/bd/dolt.go`) and `bd vc commit` (`cmd/bd/vc.go`) calling `CommitAll`, which includes config and reports whether a commit landed.
- #4383 did not absorb the unique routing. On its current head `c412b96b9`, `DoltStore.CommitWithConfig` is still `CALL DOLT_COMMIT('-Am', ...)`. It adds `CommitConfigOnly` for GH#4078 instead. The 2026-08-02 park condition ("keep #5140 open until that exact value has landed in #4383") never happened.
- Current `main` also records that `CommitAll`'s pinned-server test proves `DOLT_COMMIT('-Am')` stages config on the supported path, so the original reason to reroute `CommitWithConfig` through `commitWorkingSet` is no longer the live GH#4934 failure.

Leaving this branch `CONFLICTING` against main would only generate throwaway rebase work for a bug that is already gone. If anyone still wants the narrower `CommitWithConfig` → `commitWorkingSet(configIncludeAll)` design change against current main, that should be a new PR — not this stale branch.

No GH#4934 fix is being discarded; that fix is already on main via #5830.

bee-ghosttrack:
Agreed, and thank you for closing the loop — the 09-04 question went unanswered on my side, which is my miss. Superseded is the right disposition: GH#4934 is fixed on main via #5830, and the narrower CommitWithConfig routing, if anyone still wants it, belongs in a fresh PR against current main.

Issue #4887, opened 2026-07-18
Auto-export to issues.jsonl permanently stops after remote-backed v53 migration (manual export unaffected)


ThatcherAll:
## Symptom

After upgrading bd 1.0.4 → 1.1.0 (Homebrew, macOS, embedded Dolt backend) against a **remote-backed** database (DoltHub remote configured) and running the migration as designated migrator:

```
BD_ALLOW_REMOTE_MIGRATE=1 bd migrate   # v32 -> v53, completed cleanly
bd dolt push                            # OK
```

the JSONL auto-exporter never fires again. Config is defaults: `export.auto = true`, `export.interval = 60s`, `export.git-add = true`, `export.path = issues.jsonl`.

- Writes (`bd create`, `bd update`, `bd close`) land in the Dolt DB but `.beads/issues.jsonl` stays at its pre-migration content indefinitely (observed stale across 3+ hours and dozens of bd invocations, both reads and writes).
- `.beads/export-state.json` timestamp is frozen at the last pre-migration export.
- Active probe: after a fresh `bd create`, polled the file for 200s with repeated `bd ready` invocations in between — the new issue never appeared.
- `bd hooks run pre-commit` (BD_GIT_HOOK=1) also does not flush.
- **Manual `bd export -o .beads/issues.jsonl` works perfectly** every time, so the export path itself is healthy — only the auto-trigger is dead.

No warning or error is ever printed — the failure is completely silent, which is nasty because git-tracked tracker state quietly diverges from the DB (we ship JSONL through PRs for cross-machine sync).

## Environment
- bd 1.1.0 (Homebrew bottle), macOS (Darwin 25.5.0, arm64)
- Embedded Dolt backend (`.beads/embeddeddolt`), DoltHub remote via `sync.remote`
- Migration path: v32 (created on 1.0.4) → v53 via `BD_ALLOW_REMOTE_MIGRATE=1`

## Expected
Auto-export resumes after a successful schema migration (or at minimum surfaces an error explaining why it will not run).

## Workaround
Manual `bd export -o .beads/issues.jsonl` before any commit that includes the tracker file (we wired it into a git pre-commit hook).

Happy to provide export-state.json, config, or run instrumented builds against this database.

maphew:
[...]

PR #5141, opened 2026-07-29
fix(export): stop auto-export guards from wedging on compacted ephemeral wisps

### LINKED ISSUE #4988: Auto-export wedges permanently after ephemeral wisp compaction leaves orphan records in issues.jsonl
## Symptom

After a wisp (ephemeral molecule) flow, auto-export writes the wisp root to `.beads/issues.jsonl`. When TTL compaction later deletes the wisp root from the Dolt store, the JSONL is not updated. From then on, every bd write refuses to export:

```
auto-export skipped: issues.jsonl contains N JSONL-only issue record(s) absent from the local Dolt store
```

The export stays frozen permanently — observed twice in the same workspace (bd 1.1.0).

## Steps to Reproduce

1. Run a `bd mol wisp` flow; the wisp root is auto-exported to `.beads/issues.jsonl`.
2. Wait for TTL compaction to delete the wisp root from Dolt; the JSONL is not rewritten.
3. Perform any bd write — auto-export refuses with the message above, every time.

## Why the suggested remedies don't fit

Both remedies the guard suggests are wrong for this case:

- `bd init --from-jsonl` resurrects the deliberately compacted wisp record.
- Moving the file aside discards the export history.

## Workaround

`bd export -o .beads/issues.jsonl` rewrites the file from Dolt and clears the wedge in one command.

## Root-cause candidates

- Exclude ephemeral beads from auto-export entirely, or
- have compaction tombstone/rewrite the JSONL when it deletes records, or
- have the orphan guard ignore JSONL-only records that are marked ephemeral.

## Environment

- bd 1.1.0 (Homebrew bottle), macOS (Darwin, arm64)
- Local-only Dolt (no remote), dolt server mode

Possibly related: #3649 (wisps included in JSONL export by default), #4887 (auto-export permanently stops after v53 migration — different trigger, no guard message).

vishnujayvel:
## Summary

After a `bd mol wisp` flow, the wisp root gets auto-exported into `.beads/issues.jsonl`. When TTL compaction later deletes that wisp from the Dolt store, the JSONL line is left behind, and every subsequent `bd` write refused to auto-export with `issues.jsonl contains N JSONL-only issue record(s) absent from the local Dolt store` — permanently, since nothing rewrites the JSONL once export is refused. This fixes both guards responsible: `missingJSONLIssueIDsInStore` now skips JSONL rows marked `ephemeral`, `is_template`, or of an infra issue type when computing "missing" ids, so a stale wisp is no longer treated as a hard orphan conflict; `guardAutoExportOverwrite`'s shrink guard now only blocks a rewrite on memory or unknown-type records, so it no longer refuses to write a JSONL file that would drop those same out-of-scope rows. Together these let auto-export unwedge itself and rewrite the file to match the current store instead of refusing forever.

Thanks to @justin-candor for the report and root-cause candidates — this implements the third one: "have the orphan guard ignore JSONL-only records that are marked ephemeral."

This does not touch #4887 (auto-export stopping after a v53 migration), which the issue notes as a different trigger with no guard message; that is out of scope here.

Closes #4988

## Test plan

- [x] Targeted: `go test ./cmd/bd/ -run 'TestGuardAutoExportOverwriteBlocksRicherJSONL|TestGuardAutoExportOverwriteAllowsStaleEphemeralWisp|TestIssueRecordsInJSONL_SkipsTombstoneAndNonIssue' -v -count=1` (icu4c CGO flags) — all pass:
  ```
  --- PASS: TestGuardAutoExportOverwriteBlocksRicherJSONL (0.00s)
  --- PASS: TestGuardAutoExportOverwriteAllowsStaleEphemeralWisp (0.00s)
  --- PASS: TestIssueRecordsInJSONL_SkipsTombstoneAndNonIssue (0.00s)
  ok  	github.com/steveyegge/beads/cmd/bd	1.561s
  ```
- [x] Full package suite verified on a probe branch merged with `origin/main` at `3830f0a34`: `go test ./cmd/bd/ -count=1 -timeout 25m` passed cleanly in 473s with no failures.

### DIFF
--- a/cmd/bd/export_auto.go
+++ b/cmd/bd/export_auto.go
@@ -214,15 +214,21 @@ func countIssueRecordsInJSONL(path string) (int, error) {
 }
 
 func missingJSONLIssueIDsInStore(ctx context.Context, path string) ([]string, error) {
-	existingIDs, err := issueIDsInJSONL(path)
+	// GH#4988: only refuse when *in-scope* JSONL issue records are absent from
+	// the store. Ephemeral wisps, templates, and infra types are outside
+	// auto-export scope (buildAutoExportFilter / GH#3649). Compaction can
+	// delete wisps from Dolt while an older JSONL still lists them; treating
+	// those as hard orphans wedged auto-export permanently.
+	existing, err := issueRecordsInJSONL(path)
 	if err != nil {
 		return nil, err
 	}
-	if len(existingIDs) == 0 {
+	if len(existing) == 0 {
 		return nil, nil
 	}
 
-	issues, err := store.SearchIssues(ctx, "", types.IssueFilter{Limit: 0})
+	filter, infraTypeSet := buildAutoExportFilter(ctx)
+	issues, err := store.SearchIssues(ctx, "", filter)
 	if err != nil {
 		return nil, fmt.Errorf("failed to search issues: %w", err)
 	}
@@ -232,15 +238,29 @@ func missingJSONLIssueIDsInStore(ctx context.Context, path string) ([]string, er
 	}
 
 	missing := make([]string, 0)
-	for _, id := range existingIDs {
-		if _, ok := localIDs[id]; !ok {
-			missing = append(missing, id)
+	for _, rec := range existing {
+		if rec.Ephemeral || rec.IsTemplate || infraTypeSet[string(rec.IssueType)] {
+			continue
+		}
+		if _, ok := localIDs[rec.ID]; !ok {
+			missing = append(missing, rec.ID)
 		}
 	}
 	return missing, nil
 }
 
-func issueIDsInJSONL(path string) ([]string, error) {
+// jsonlIssueRecord is a lightweight issue line from issues.jsonl used by
+// auto-export safety guards.
+type jsonlIssueRecord struct {
+	ID         string
+	IssueType  types.IssueType
+	IsTemplate bool
+	Ephemeral  bool
+}
+
+// issueRecordsInJSONL returns issue records (id + scope fields) from a JSONL
+// export file. Tombstones and non-issue record types are skipped.
+func issueRecordsInJSONL(path string) ([]jsonlIssueRecord, error) {
 	f, err := os.Open(path) //nolint:gosec
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -254,7 +274,7 @@ func issueIDsInJSONL(path string) ([]string, error) {
 	scanner.Buffer(make([]byte, 0, 1024*1024), 64*1024*1024)
 
 	seen := make(map[string]struct{})
-	var ids []string
+	var records []jsonlIssueRecord
 	for scanner.Scan() {
 		line := strings.TrimSpace(scanner.Text())
 		if line == "" {
@@ -273,11 +293,11 @@ func issueIDsInJSONL(path string) ([]string, error) {
 			}
 		}
 
-		var id string
+		var rec jsonlIssueRecord
 		if rawID, ok := raw["id"]; ok {
-			_ = json.Unmarshal(rawID, &id)
+			_ = json.Unmarshal(rawID, &rec.ID)
 		}
-		if id == "" {
+		if rec.ID == "" {
 			continue
 		}
 
@@ -289,17 +309,39 @@ func issueIDsInJSONL(path string) ([]string, error) {
 			continue
 		}
 
-		if _, ok := seen[id]; ok {
+		if rawIT, ok := raw["issue_type"]; ok {
+			_ = json.Unmarshal(rawIT, &rec.IssueType)
+		}
+		if rawTpl, ok := raw["is_template"]; ok {
+			_ = json.Unmarshal(rawTpl, &rec.IsTemplate)
+		}
+		if rawEph, ok := raw["ephemeral"]; ok {
+			_ = json.Unmarshal(rawEph, &rec.Ephemeral)
+		}
+
+		if _, ok := seen[rec.ID]; ok {
 			continue
 		}
-		seen[id] = struct{}{}
-		ids = append(ids, id)
+		seen[rec.ID] = struct{}{}
+		records = append(records, rec)
 	}
 	if err := scanner.Err(); err != nil {
 		return nil, err
 	}
 
-	sort.Strings(ids)
+	sort.Slice(records, func(i, j int) bool { return records[i].ID < records[j].ID })
+	return records, nil
+}
+
+func issueIDsInJSONL(path string) ([]string, error) {
+	records, err := issueRecordsInJSONL(path)
+	if err != nil {
+		return nil, err
+	}
+	ids := make([]string, len(records))
+	for i, rec := range records {
+		ids[i] = rec.ID
+	}
 	return ids, nil
 }
 
@@ -499,16 +541,20 @@ func guardAutoExportOverwrite(path string, infraTypes map[string]bool, includeMe
 		return fmt.Errorf("auto-export shrink guard: inspect existing JSONL: %w", err)
 	}
 
-	if stats.FilteredRecords == 0 {
+	// Only block on records we must not silently drop: memories (when excluded)
+	// and unknown record types. Ephemeral / template / infra issues are outside
+	// auto-export scope by design (GH#3649); allowing the rewrite clears stale
+	// wisps left after TTL compaction (GH#4988).
+	if stats.Memories == 0 && stats.UnknownRecords == 0 {
 		return nil
 	}
-	return fmt.Errorf("auto-export shrink guard: refusing to overwrite %s because it contains %d record(s) outside auto-export scope (%d memories, %d infra/template/ephemeral issues, %d unknown); run an explicit export if you want to replace it", path, stats.FilteredRecords, stats.Memories, stats.FilteredIssues, stats.UnknownRecords)
+	return fmt.Errorf("auto-export shrink guard: refusing to overwrite %s because it contains %d record(s) that auto-export would drop (%d memories, %d unknown); run an explicit export if you want to replace it", path, stats.Memories+stats.UnknownRecords, stats.Memories, stats.UnknownRecords)
 }
 
 type autoExportOverwriteStats struct {
-	FilteredRecords int
+	FilteredRecords int // retained for tests / logging; blocking uses Memories+Unknown
 	Memories        int
-	FilteredIssues  int
+	FilteredIssues  int // ephemeral/infra/template — non-blocking (GH#4988)
 	UnknownRecords  int
 }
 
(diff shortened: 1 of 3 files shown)


maphew (commented):
## Disposition: half of this is right and I want to land it — the other half re-opens #4069

Thanks for taking this one, @vishnujayvel. The diagnosis is correct and it's a genuinely nasty bug: I reproduced the logic on current `main` and both guards do fire, and the wedge really is permanent because `maybeAutoExport` returns before the export that would rewrite the file. Concretely, on `main`:

- `cmd/bd/export_auto.go:216` `missingJSONLIssueIDsInStore` compares *all* JSONL ids against the store. A compacted wisp is in the file and gone from Dolt, so it lands in `missing`, and `maybeAutoExport` (`cmd/bd/export_auto.go:110-118`) prints the warning and returns `nil` without exporting. Nothing ever rewrites the JSONL, so the next write hits the identical state. Permanent, exactly as reported.
- Even if that guard passed, `classifyExistingAutoExportRecord` counts `record.Ephemeral` into `stats.FilteredRecords`, so `guardAutoExportOverwrite` refuses too. So yes — both guards, and you have to move both. Good call.

The JSONL-side half of your fix is the right idea and I'd take it more or less as written. The problem is the shrink-guard half.

### Blocker: relaxing the shrink guard for infra/template rows undoes #4069

`cmd/bd/export_auto.go:548` in the PR changes the block condition from `stats.FilteredRecords == 0` to `stats.Memories == 0 && stats.UnknownRecords == 0`, and `:589` stops counting out-of-scope issue rows as blocking at all. That guard was added specifically for #4069 — titled *"auto-export uses default filter, drops infra/ephemeral rows that exist in DB (89% data loss in our workspace)"*. In that reporter's workspace the richer JSONL was 2141 rows vs the auto-export subset's 234; the 1907-row delta was 1874 `molecule` + 25 `task` + 3 `agent` issue rows and only **7** memory records. Under this PR, a JSONL like that with no memory rows at all — e.g. anything produced by `bd export --include-infra -o .beads/issues.jsonl` — has zero blocking records, and the next auto-export silently shrinks it back to the filtered subset. That is #4069's failure verbatim, and they lost data to it twice in production (a rig session and a dolt-server recovery).

Failure scenario, spelled out: `bd export --include-infra -o .beads/issues.jsonl` writes 2000 rows including 1874 molecules; `export.auto=true`; the next `bd create` calls `exportToFile` → `guardAutoExportOverwrite` sees `Memories=0, UnknownRecords=0` → returns nil → the file is atomically replaced with the ~130-row in-scope subset, and the molecule rows are gone from the git-tracked recovery source.

The good news is the narrow fix is right there and it's cheap: the rows we're allowed to drop aren't "out-of-scope rows", they're **out-of-scope rows that no longer exist in the store**. A stale wisp is safe to drop precisely because Dolt doesn't have it any more; a molecule that's still in Dolt is exactly what #4069 says we must not drop. `missingJSONLIssueIDsInStore` already builds a store id set — plumb a "store-known ids" set into `guardAutoExportOverwrite` (or compute it once in `exportToFile`, which already has the store handle at `:411`) and keep counting an out-of-scope row as blocking when its id is still present in the store. That closes #4988 completely without touching #4069's protection, and it makes the two guards agree on a single rule instead of two different ones.

Worth noting too: the reporter's own third candidate on #4988 was "have the orphan guard ignore JSONL-only records that are **marked ephemeral**" — the "JSONL-only" qualifier is doing real work there, and widening it to templates and infra types is where this drifts past the bug.

### Should-fix: the store-side filter change can create a *new* permanent wedge

At `cmd/bd/export_auto.go:230` the store query changes from `types.IssueFilter{Limit: 0}` to `buildAutoExportFilter(ctx)`, which sets `ExcludeTypes` (infra), `IsTemplate=&false` and `Ephemeral=&false`. That filter is strictly narrower than the old one, so `localIDs` is now a strict subset of what it used to be — the change can only ever make *more* ids look missing, never fewer.

The mismatch that bites: JSONL scope is judged from the fields recorded in the file, store scope from the row's *current* state. Any issue that was exported while in scope and later moved out of scope in Dolt now reads as an in-scope JSONL row with no store match → `missing` → permanent wedge, and on `main` (nil `Ephemeral`, which merges wisps — see `internal/storage/dolt/dolt_test.go:2329`) it did not. A type change into an infra type, or a bead becoming a template/wisp after its last export, both land here. I didn't chase down every CLI path that can perform such a transition, so treat the specific trigger as unconfirmed — but the direction of the filter change is verifiable from the code alone, and it's the wrong direction for a guard whose failure mode is "wedges forever".

The JSONL-side skip at `:242` is the part that actually fixes #4988; the store-side query wants to stay as permissive as possible. Suggest reverting `:230` to an unfiltered search and just adding `MaxRows: 0` (worth having on its own — the old `types.IssueFilter{Limit: 0}` left `MaxRows` unset, so a `BEADS_MAX_ROWS` cap could truncate the store side and manufacture phantom "missing" ids).

### Should-fix: the weakened assertion is the tell

`cmd/bd/export_auto_test.go` edits `TestGuardAutoExportOverwriteBlocksRicherJSONL` to drop `"3 infra/template/ephemeral issues"` and accept `"2 record(s) that auto-export would drop"`. That test *is* #4069's regression test — rewriting it to accept the looser behaviour is the signal that the fix went past the bug rather than a formatting update. With the store-presence rule above, that assertion should go back to expecting a block on infra/template rows that still exist in the store, and the new `TestGuardAutoExportOverwriteAllowsStaleEphemeralWisp` becomes the complement (out-of-scope **and** store-absent → rewritable). Those two sitting side by side would document the rule nicely.

### Nit: no coverage at the layer that actually wedged

`TestIssueRecordsInJSONL_SkipsTombstoneAndNonIssue` is a good parser test, but there's no test over `missingJSONLIssueIDsInStore` itself — the function whose return value gates the whole export. A store-backed test that seeds one persistent issue plus a JSONL listing a compacted wisp, and asserts `missing` is empty, would pin the actual #4988 behaviour. (`TestMissingJSONLIssueIDsInStore` is named in the original guard commit's validation list but doesn't exist in the tree, so this gap predates you.)

### Housekeeping

- No collisions: nothing else open touches auto-export guards (#5121 is yours and is in `gitAddFile` territory, not here). Agreed that #4887 is a separate trigger and out of scope.
- The new error message drops the infra/template/ephemeral counts. Once the guard blocks on store-present rows again, please keep them in the message — that breakdown is what tells a user *which* rows they're about to lose.

Send the store-presence version and I think this lands quickly. The parsing work in `issueRecordsInJSONL` and the `issueIDsInJSONL` shim over it are clean and I'd keep them exactly as they are.

vishnujayvel:
You were right, and the part that stings is that you diagnosed it from the test diff: weakening `TestGuardAutoExportOverwriteBlocksRicherJSONL` to accept the looser message *was* the tell. That test is #4069's regression test, and editing a regression test to agree with your change is the thing you're supposed to notice yourself.

Pushed the store-presence version in `4c62fcae8`.

- `guardAutoExportOverwrite` and `classifyExistingAutoExportRecord` now take a `storeIDs` set, computed once in `exportToFile` via a new `storeKnownIssueIDs` helper. An out-of-scope row blocks when its id is **still in the store**, and only stops blocking once the store has genuinely dropped it. One rule, both guards.
- `missingJSONLIssueIDsInStore` reverted to an unfiltered `types.IssueFilter{Limit: 0, MaxRows: 0}`. You were right about `MaxRows` being worth having on its own — the old `{Limit: 0}` left it unset, so a `BEADS_MAX_ROWS` cap could truncate the store side and manufacture phantom missing ids.
- The error message keeps the full breakdown. One judgment call: the infra/template/ephemeral count now reports only the *blocking* (store-present) subset rather than every out-of-scope row, since store-absent rows aren't rows the user is about to lose. Say the word if you'd rather see both numbers.
- `issueRecordsInJSONL` and the `issueIDsInJSONL` shim are untouched, as you asked.

**I mutation-tested the two guard tests rather than just re-running them**, because "the assertion is back" isn't the same as "the assertion catches the regression":

*Neuter the rule so out-of-scope never blocks* (your #4069 failure, reintroduced):
```
--- FAIL: TestGuardAutoExportOverwriteBlocksRicherJSONL
    "...2 record(s) outside auto-export scope (1 memories, 0 infra/template/ephemeral issues, 1 unknown)..."
    does not contain "4 record(s) outside auto-export scope"
```
*Ignore store-presence so out-of-scope always blocks* (the wedge-forever direction):
```
--- FAIL: TestGuardAutoExportOverwriteBlocksRicherJSONL      (counts bd-wisp: 3 issues, not 2)
--- FAIL: TestGuardAutoExportOverwriteAllowsStaleEphemeralWisp  (stale wisp blocks; wedged)
```
Neither an always-allow nor an always-block implementation satisfies both, so the pair pins the rule from both sides rather than sharing one trivial condition.

**Honest gap:** the new `TestMissingJSONLIssueIDsInStore_IgnoresCompactedWisp` — the missing-layer test you asked for — **SKIPS** here (`Dolt test server not available`; Docker isn't running on this machine). Same for the existing `TestExportAuto_FilterHasZeroMaxRows`. Both compile and skip cleanly, and both should run in CI, but I have not watched either go green and I'm not going to claim otherwise. The five pure-guard tests above all pass in the default build; `gofmt` and `go vet ./cmd/bd/...` clean, `CGO_ENABLED=0 go vet ./cmd/bd/` clean.

On the build tag: the new file carries `//go:build cgo`, matching all 23 Dolt-server-backed files in `cmd/bd` and its closest sibling `export_auto_max_rows_gate_test.go`. `export_auto_test.go` is untagged because it's a plain fake-store unit file, not a server-backed one.

maphew:
The store-presence rule is the right reconciliation, and the reasoning in the comments is the part I'd most want to keep — "an out-of-scope row already in the JSONL only blocks the rewrite when its id is STILL in the store" states the #4069-vs-#4988 tension precisely enough that the next person won't re-break it by picking one side.

Two things I checked specifically:

**The unfiltered queries are correct, not sloppy.** `Limit: 0, MaxRows: 0` in both `missingJSONLIssueIDsInStore` and `storeKnownIssueIDs` looks like an oversight until you follow the failure mode: a narrower filter or a row cap can only ever manufacture phantom "missing" ids, and the consequence of a phantom is a *permanent* auto-export wedge. Asymmetric costs justify the permissive query. Good call, and good that the reason is written down at both sites.

**Adding `TestMissingJSONLIssueIDsInStore_IgnoresCompactedWisp` closes a real gap.** A test by that name was listed in the original guard commit's validation notes but was not in the tree. Pinning the behaviour at the layer that actually wedged is better than testing it one level up.

One non-blocking efficiency note, for now or for a follow-up:

`storeKnownIssueIDs(ctx)` is called unconditionally in `exportToFile` before `guardAutoExportOverwrite`, so every auto-export now pays a second full-table scan — on top of the one `exportToFile` already did for the export itself. But the guard short-circuits in the two common cases: no existing JSONL file (`os.IsNotExist` → `nil`), and `stats.FilteredRecords == 0`. So in the steady state we're doing double the store queries to compute a set that's usually unused. Auto-export fires on a lot of write paths, so it's worth making lazy — pass a `func() (map[string]struct{}, error)` and only resolve it at the point where a filtered record is actually seen.

Not blocking; the correctness argument stands on its own and I'd rather land this than hold it for a perf tweak. Happy to take the lazy version as a follow-up PR if you'd prefer to keep this one focused.

PR #5143, opened 2026-07-29
fix(dolt): verify project identity when adopting an existing database on the CreateIfMissing path

maphew:
## Why

`bd init` against a shared Dolt server always opens with `CreateIfMissing:true`
(`cmd/bd/init.go`), and the server-mode connect path (`newServerMode` /
`openServerConnection` in `internal/storage/dolt/store.go`) only ran the
project-identity verifier (`verifyProjectIdentity` /
`verifyGlobalProjectIdentity`) when `CreateIfMissing` was false. That meant
init's own connect never verified identity at all: if a workspace already had
a `project_id` in `.beads/metadata.json` and `bd init` (or any other
`CreateIfMissing:true` open) happened to target a database that already
existed on the server under a *different* project's identity — wrong
`--database` name, a stale port file, a server restarted against a different
data directory — bd would silently connect to that foreign database and start
reading and writing into it instead of failing loudly (#4637).

This makes `bd init` (and any other `CreateIfMissing:true` open) refuse a
wrong target only when the workspace already has a `project_id` in
`.beads/metadata.json`, the target database already exists on the server,
and that database records a different non-empty `_project_id`; it does not
catch a foreign server that has no beads database at all, a database with no
`_project_id` recorded, a workspace with no `metadata.json` yet, or a
gateway-routed init.

## What changed

- `openServerConnection` already computed whether the target database existed
  before deciding whether to `CREATE DATABASE`. It now returns that as a
  fourth value (`dbAlreadyExisted`) instead of discarding it, including
  correcting it to `true` when `CREATE DATABASE IF NOT EXISTS` itself reports
  "database exists" (error 1007) — proof another process won a creation race
  that the initial `SHOW DATABASES` probe missed.
- `newServerMode`'s identity gate now runs whenever `CreateIfMissing` is false
  (unchanged) **or** the database already existed. A foreign existing
  database with a mismatched `_project_id` now fails with `PROJECT IDENTITY
  MISMATCH` instead of silently connecting. A genuinely new database still
  skips verification unchanged, and the existing soft-skip behavior (missing
  local `metadata.json`, empty `_project_id` on either side) is preserved.
- The identity gate now runs **before** `store.initSchema`, not after: a
  foreign existing database is rejected before bd's migrations and Dolt
  commits are written into it, not after. (The old ordering — schema init,
  then verify — predates this change, but this change is what routes `bd
  init` into that ordering for the first time, so the exposure window is
  new.)
- **Gateway mode is explicitly exempt** from the new check.
  `openServerConnection`'s gateway branch never probes existence — a
  successful ping to a server-provisioned database is the only signal it
  has — so it reports `dbAlreadyExisted == false` unconditionally and the new
  gate never fires for Gateway. This is intentional: a hosted database's
  identity is server-authoritative, and `cmd/bd/init.go`'s
  `resolveInitProjectID` already reconciles a stale or missing local
  `project_id` onto the server's on every gateway init — including
  re-init — which depends on this open *not* hard-failing first. **Gateway
  identity is not covered by this PR at all**; it is reconciled, not
  enforced.

## Also affected

The fix lives in the shared `newServerMode` connect path, so it is not
limited to `bd init`. Every other `CreateIfMissing:true` caller gets the same
identity check for free (and is exposed to the same narrow scope described
above): `bd bootstrap`'s `executeInitAction`, `executeRestoreAction`, and
`executeJSONLImportAction`, the ad hoc store open in `cmd/bd/create.go:884`,
and the public `beads.OpenFromConfig` API.

## Recovery

A workspace whose local `.beads/metadata.json` carries a stale or wrong
`project_id` — the case this PR is specifically about avoiding — can be
repaired without hand-editing the file: `bd doctor --fix`.

## Out of scope (Part B)

This is Part A only. It intentionally does not add any server-identity
sentinel (UUID, sentinel table, or anything persisted at provision time) and
does not attempt to detect a foreign Dolt server that has no beads database
at all — that remains a silent-connect gap. That design question is tracked
separately.

## Tests

`internal/storage/dolt/cross_project_test.go` (folded into the existing
cross-project-isolation test family rather than a new file, since it already
owns the "two projects share one Dolt server" scenario):

- `TestCrossProject_IdentityCheck_ExistingDatabase_ForeignRejected` — negative
  case: existing database + mismatched local `project_id` +
  `CreateIfMissing:true` → `PROJECT IDENTITY MISMATCH`.
- `TestCrossProject_IdentityCheck_ExistingDatabase_MatchingSucceeds` —
  positive control: existing database + matching local `project_id` +
  `CreateIfMissing:true` → succeeds.
- `TestCrossProject_IdentityCheck_NewDatabase_Succeeds` — genuinely new
  database + `CreateIfMissing:true` → succeeds, unchanged.
- `TestCrossProject_IdentityCheck_SoftSkip_NoLocalMetadata` — no local
  `metadata.json` → soft-skip, succeeds.
- `TestCrossProject_IdentityCheck_SoftSkip_EmptyDBProjectID` — database has no
  `_project_id` recorded (old-style database) → soft-skip, succeeds.
- `TestCrossProject_IdentityCheck_Gateway_SkipsVerification` — existing
  database + mismatched local `project_id` + `Gateway:true` +
  `CreateIfMissing:true` → succeeds (identity reconciled by
  `resolveInitProjectID`, not enforced at open). This is the branch most
  likely to be "cleaned up" by mistake later; the test exists so that change
  has to touch this assertion, not just delete a condition.

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -113,12 +113,17 @@ func resolveInitIssuePrefix(gateway bool, existing, dbName, prefix string, readE
 // Gateway: the hosted database's identity is server-authoritative, so an adopted
 // server id always wins and is reconciled onto local even when localID is already
 // set. A re-init or orchestrator-preseeded workspace must not keep a stale local
-// id: init opens with CreateIfMissing, which skips the storage identity verifier
-// (store.go verifyProjectIdentity), so a stale id would be saved as success and
-// every later normal open would then hard-fail with PROJECT IDENTITY MISMATCH. A
-// missing server id is a provisioning-contract violation bd will not mint over —
-// even when a local id already exists — and a read error is surfaced as the
-// transient failure it is, so a flaky connection is not misdiagnosed as an
+// id: for Gateway specifically, init's CreateIfMissing:true open still skips the
+// storage identity verifier (store.go verifyProjectIdentity/newServerMode — see
+// its dbAlreadyExisted comment for why Gateway is exempt from the
+// otherwise-CreateIfMissing:true check), so a stale id would be saved as success
+// and every later normal open would then hard-fail with PROJECT IDENTITY
+// MISMATCH. This CreateIfMissing skip is Gateway-only: a non-gateway
+// CreateIfMissing:true init against an already-existing database now DOES run
+// the verifier (GH#4637 Part A) and fails before reaching this reconciliation.
+// A missing server id is a provisioning-contract violation bd will not mint
+// over — even when a local id already exists — and a read error is surfaced as
+// the transient failure it is, so a flaky connection is not misdiagnosed as an
 // unprovisioned database.
 //
 // Non-gateway (legacy, unchanged): a non-empty localID is kept as-is (readErr
@@ -1385,10 +1390,14 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 			// at all (a missing identity is a provisioning-contract violation),
 			// and because the hosted server is authoritative it reconciles the
 			// identity on every init — even a re-init or preseeded workspace whose
-			// metadata.json already carries a stale project_id. Otherwise init
-			// (which opens with CreateIfMissing, skipping the storage identity
-			// verifier) would save the stale id as success and every later normal
-			// open would hard-fail with PROJECT IDENTITY MISMATCH.
+			// metadata.json already carries a stale project_id. This reconciliation
+			// depends on Gateway's CreateIfMissing:true open skipping the storage
+			// identity verifier (store.go newServerMode/verifyProjectIdentity) —
+			// Gateway-only, not init in general: a non-gateway CreateIfMissing:true
+			// init against an already-existing database now runs the verifier
+			// (GH#4637 Part A) and fails before this code runs at all. Without the
+			// Gateway skip, init would save the stale id as success and every later
+			// normal open would hard-fail with PROJECT IDENTITY MISMATCH.
 			adoptedFromDB := ""
 			var adoptReadErr error
 			if store != nil && shouldConsultInitProjectID(doltCfg.Gateway, cfg.ProjectID, database, bootstrappedFromRemote) {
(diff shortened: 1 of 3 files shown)


maphew:
Rebased onto current `main` (`423afdcb2`) to clear a conflict with #5042, which landed a bool return on `openServerConnection` while this PR was open.

The two signals are **not** inverses, so I did not collapse them:

- #5042's `created` — "THIS call's bare `CREATE DATABASE` won the ownership arbitration" — arms `schema.WithFreshBootstrapHeal`.
- This PR's `alreadyExisted` — "the database was proven to exist before this call" — forces project-identity verification even when `CreateIfMissing` is true.

They diverge on the Gateway branch, where existence is never probed: both must be `false` there. `!created` would have been `true`, wrongly forcing identity verification on a gateway open — the exact case this PR documents as deliberately excluded (gateway identity is reconciled by `resolveInitProjectID`, not enforced at open).

Resolution: `openServerConnection` now returns a small `serverConnFacts{created, alreadyExisted}` struct instead of a bare bool, with the non-inverse relationship documented on the type so a future cleanup does not "simplify" one field into the other.

Verification on the rebased head (`b40f0a3c6`): `go build ./...` and `go vet` clean; `go test ./internal/storage/dolt/ -run 'CrossProject|ProjectIdentity|Identity|FreshBootstrapHeal|Gateway'` — 18 pass, 0 fail. That includes #5042's own `TestFreshBootstrapHealNotArmedWhenDatabasePreexists` and `TestFreshBootstrapHealSelfHealsAfterMidPassFailure`, which are the tests this resolution could have broken, plus this PR's six identity cases (foreign-rejected, matching-succeeds, new-database, both soft-skips, and the gateway-skip pin).

Scope is unchanged from the original description.

maphew:
Rebased onto current `main` (`8bb0d36be`). No conflicts, no change to this PR's diff.

Why: the only failing job here was `Test (Proxied Dolt Cmd 14/15)`, and inside it the only failing test was `TestProxiedServerComment/multiple_comments_ordered` — unrelated to this PR, which does not touch comments. That was #5165 (same-second comments reading back in content-digest order after the content-derived ids in #5150), fixed on main by #5167 with a test follow-up in #5166. The identical failure was also parking #5163.

So this needed a rebase rather than any change to the identity check. Re-running CI from the current base.

PR #5144, opened 2026-07-29
docs(dolt): document remote-server password/TLS auth in CLI help

### LINKED ISSUE #5011: docs: bd dolt CLI help omits remote-server auth (BEADS_DOLT_PASSWORD / TLS / credentials file)
## Summary

Remote Dolt server **auth** (password + TLS) is fully supported by bd, but is **undiscoverable from the CLI help**. `bd dolt set --help` lists only `database/host/port/user/data-dir` — no `password`, no `tls`. `bd dolt --help` mentions neither `BEADS_DOLT_PASSWORD` nor `BEADS_DOLT_SERVER_TLS` nor the `~/.config/beads/credentials` file. A user pointing bd at a password+TLS remote has no way to learn how from the tool, which pushes people toward tunnels/forks instead of the built-in path.

## Repro

```
$ bd dolt set --help
Keys:
  database  Database name (default: issue prefix or "beads")
  host      Server host (default: 127.0.0.1)
  port      Server port (auto-detected; override with bd dolt set port <N>)
  user      MySQL user (default: root)
  data-dir  Custom dolt data directory ...
```
No `password` / `tls` key. And:
```
$ bd dolt --help | grep -iE 'password|tls|credential'
(none)
```

## What actually works (but isn't documented in help)

From the source, `ServerDSN` (`internal/storage/doltutil/dsn.go`) has `User`, `Password`, `TLS bool`; these are configurable via `BEADS_DOLT_*` env vars and a `~/.config/beads/credentials` INI file keyed by `[host:port]`. This is the correct, supported way to reach a public password+TLS remote — but you'd only find it by reading the code.

## Proposed fix (either/both)

1. Mention `BEADS_DOLT_PASSWORD`, `BEADS_DOLT_SERVER_TLS`, and the `~/.config/beads/credentials` `[host:port]` format in `bd dolt` / `bd dolt set` help text (and in `docs/architecture/dolt.md`).
2. Add `password` and `tls` as settable keys on `bd dolt set` (or explicitly document why they're env/credentials-file only — presumably to keep secrets out of `metadata.json`).

## Context

Discovered while standing up a public, scale-to-zero Dolt server on Fly as a shared bd backend for cloud agent boxes. bd's remote support is exactly right for this; the only gap was discoverability. `bd version 1.1.0 (Homebrew)`.

vishnujayvel:
## Summary

Closes #5011.

@hexsprite reported that remote Dolt server authentication (password + TLS) is fully supported by `bd` — via `BEADS_DOLT_PASSWORD`, `BEADS_DOLT_SERVER_TLS`, and a `~/.config/beads/credentials` INI file — but was undiscoverable from the CLI: `bd dolt --help` and `bd dolt set --help` mentioned none of it, so a user standing up a password+TLS remote had no way to learn how short of reading `internal/storage/doltutil/dsn.go`.

This adds that documentation to both help texts:

- `bd dolt --help` (`doltCmd.Long`): documents `BEADS_DOLT_PASSWORD`, `BEADS_DOLT_SERVER_TLS`, `BEADS_DOLT_SERVER_USER`, `BEADS_CREDENTIALS_FILE`, the credentials-file default location and `[host:port]` INI format, password resolution order, and points to `docs/architecture/dolt.md` (which already documents this in more depth).
- `bd dolt set --help` (`doltSetCmd.Long`): explains that `password`/`tls` are intentionally not `bd dolt set` keys — so secrets stay out of `metadata.json` — and documents the same env vars/credentials-file path as the alternative.

The issue offered a second option (adding `password`/`tls` as `bd dolt set` keys); this PR takes the documentation-only route the issue itself flagged as the likely reason those keys are env/credentials-file only, and leaves the storage mechanism unchanged.

## Test plan

Env: `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null` (isolates from this machine's global `beads.role=maintainer` git config, unrelated to this change).

- [x] `go test ./cmd/bd/ -run TestDoltHelpMentionsRemoteAuth -v -count=1` → `--- PASS: TestDoltHelpMentionsRemoteAuth (0.00s)` / `ok  	github.com/steveyegge/beads/cmd/bd	0.981s`. Targeted run under the **default (cgo) build**, asserting both help strings contain `BEADS_DOLT_PASSWORD`, `BEADS_DOLT_SERVER_TLS` and `credentials`, and that the `set` help explains `metadata.json`. Not a full-suite run.
- [x] `CGO_ENABLED=0 go vet ./cmd/bd/` — pure-Go build still clean.

A note on the second commit: the test was initially written with `//go:build !cgo`, while the rest of `cmd/bd`'s test files use `//go:build cgo`. Under the default build it was excluded and never ran, so the constraint is removed in `37e40b2ee` — the run above is under the build that previously skipped it.

Separately, this change plus a merge of the fix branch onto `origin/main` (d1f67a91e) passed a full `go test ./cmd/bd/ -count=1 -timeout 25m` probe run (403s) prior to this rewrite; that full run was not re-executed for this pass.

### DIFF
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -63,6 +63,23 @@ Configuration keys for 'bd dolt set':
   user      MySQL user (default: root)
   data-dir  Custom dolt data directory (absolute path; default: .beads/dolt)
 
+Remote server authentication (password + TLS) is NOT stored via 'bd dolt set'
+(keeps secrets out of metadata.json). Configure them with:
+
+  BEADS_DOLT_PASSWORD       Server password (highest priority)
+  BEADS_DOLT_SERVER_TLS     Enable TLS (set to "1" or "true")
+  BEADS_DOLT_SERVER_USER    MySQL user override (else use 'bd dolt set user')
+  BEADS_CREDENTIALS_FILE    Optional path to credentials file
+
+  Default credentials file: ~/.config/beads/credentials (Linux/macOS)
+                            %APPDATA%\\beads\\credentials (Windows)
+  Format (INI, section = host:port of the resolved connection):
+    [127.0.0.1:3307]
+    password = secret
+
+  Password resolution: BEADS_DOLT_PASSWORD → credentials [host:port] → empty.
+  Full reference: docs/architecture/dolt.md (Environment Variables / Credentials).
+
 Flags for 'bd dolt set':
   --update-config  Also write to config.yaml for team-wide defaults
 
@@ -70,6 +87,7 @@ Examples:
   bd dolt set database myproject
   bd dolt set host 192.168.1.100 --update-config
   bd dolt set data-dir /home/user/.beads-dolt/myproject
+  export BEADS_DOLT_PASSWORD=... BEADS_DOLT_SERVER_TLS=1
   bd dolt test`,
 }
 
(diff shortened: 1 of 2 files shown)


arcaven:
Heads up: #5178 (in progress) adds a `tls` key to `bd dolt set`, which touches the same help text. Your docs stand either way; whichever lands first, the other needs a one-sentence adjustment, and I am happy to be the one to rebase.

vishnujayvel:
Thanks for the heads-up. If #5178 lands first I'll rebase this and fold the `tls` key into the same block, so you don't have to pick it up. If this one goes first, the one-sentence adjustment is yours whenever it suits.

maphew:
Reviewed, and I verified each documented knob against the code rather than taking the help text at face value — that matters here because the whole point of this PR is that the help should be trustworthy:

- `BEADS_DOLT_PASSWORD` — `cmd/bd/dolt.go:2256`
- `BEADS_DOLT_SERVER_TLS` — `internal/configfile/configfile.go:494`
- `BEADS_DOLT_SERVER_USER` — `internal/configfile/configfile.go:415`
- `BEADS_CREDENTIALS_FILE` — `internal/configfile/credentials.go:43`

All four exist and resolve in the documented priority order. The "no `password` / `tls` key on purpose" note is the right thing to say out loud, and asserting it in `TestDoltHelpMentionsRemoteAuth` is what keeps it from rotting.

One defect, which I fixed maintainer-side rather than round-tripping: `abce2db7e`.

Both `Long:` bodies are backtick raw string literals, so `%APPDATA%\\beads\\credentials` reached users as a literal double backslash. `internal/configfile/credentials.go:14` has it right with single backslashes, so the help was contradicting its own source of truth — a small version of exactly the failure this PR exists to fix. Changed to `%APPDATA%\beads\credentials`; `gofmt` clean, `go build` green.

Landing this one early on purpose: it touches `cmd/bd/dolt.go`, which #5125 and #5140 also touch and which are both currently conflicting. Better that they rebase onto this than the reverse.

Issue #4993, opened 2026-07-23
bd doctor recommends '--fix' without checking whether the target DB schema is above the running binary's — the command that caused a fleet-wide skew incident is advertised by the standard diagnostic


JennyMink:
## Summary

`bd doctor` recommends `bd doctor --fix` for purely cosmetic warnings, without first checking whether the target database's schema version is **above** what the running binary can safely write. On a DB with a pending or newer migration, `--fix` is not a passive repair — it resumes/applies a write migration.

This matters because `doctor` is the command an operator reaches for *when something already looks off*, which is exactly when following its advice is most costly.

## Observed

On a live shared rig DB (Dolt server backend, bd 1.0.5):

```
bd doctor v1.0.5  ──────────  ✓ 66 passed  ⚠ 10 warnings  ✖ 0 errors
```

Zero errors, and the warnings are all cosmetic — project-identity backfill, `.beads/.gitignore`, gitignore patterns, tracked runtime files, remote consistency. `doctor` nonetheless recommends `bd doctor --fix` repeatedly. No warning in that set requires a migration.

Schema versions, read directly from `schema_migrations` rather than inferred:

| DB | max applied | note |
|----|-------------|------|
| `silicon` | 49 | contiguous, 49 rows |
| `hq` | 52 | the DB involved in the skew incident |

Binary: `bd version 1.0.5 (Homebrew)`.

So the relationship between binary and DB schema **differs per database**, and `doctor`'s recommendation does not vary with it. On a DB at or below the binary's level the advice is harmless; on one above it — or with a migration pending — the same advice resumes a write migration against shared state.

## Why this is a defect and not just an ops rule

In our case a change-freeze was in force ("no binary/state changes touch live DBs before <date>"). That guarantee lived in mail and in an issue tracker. Nothing on the tool path enforced it. An operator who hits an oddity, runs `doctor`, and follows the printed recommendation has not broken a rule they were aware of — they followed the tooling. The gap between the stated guarantee and the enforced one is one keystroke wide.

## Suggested fix

Either of these would close it; the first is stronger:

1. **Refuse to recommend `--fix`** when the target DB's schema version is above the running binary's, or when a migration is pending. Print why instead.
2. **Separate cosmetic fixes from migration-bearing ones.** A gitignore warning should never be able to lead an operator into resuming a paused schema migration. Distinct commands, or at minimum distinct recommendations with the migration-bearing ones gated behind an explicit flag and a version check.

[...]

PR #5145, opened 2026-07-29
fix(doctor): gate --fix when schema is skewed

### LINKED ISSUE #1: `bd dep tree` mostly broken

vishnujayvel:
## Summary

Refs #4993.

@JennyMink reported that `bd doctor` recommends `bd doctor --fix` for purely cosmetic warnings without checking whether the target database's schema is ahead of (or behind) the running binary's, the pattern behind a prior fleet-wide skew incident: following the printed tip during a change freeze resumed a write migration against shared state.

`AssessSchemaFixGate` (`cmd/bd/doctor/fix_gate.go`) reads `schema_migrations` once per invocation, compares it to `schema.LatestVersion()` for this binary, and produces a `FixGate`. Everything that can write or advise consults that one gate:

| DB state | `--fix` DB fixes | printed / `--json` / `--output` / `--agent` `bd doctor --fix` advice |
|---|---|---|
| equal | allowed | unchanged |
| ahead of the binary | withheld | rewritten, original tip preserved |
| behind the binary (migrations pending) | withheld, including `Pending Migrations` | rewritten, original tip preserved |
| reachable but version unreadable (empty or absent `schema_migrations`, read error) | withheld (fail closed) | rewritten |
| unreachable (`openDoltDB` fails) | schema-writing fixes withheld; recovery fixes allowed (rule below) | unchanged |

Filesystem-only fixes (`filesystemOnlyFixes`) are never affected. `--check=<x> --clean/--fix` is refused when the database is reachable and the gate blocks. Admission is `FixGate.AllowsFix(checkName)` alone; `Reason` is advisory text, not a safety signal.

### Review round 2 (bee-ghosttrack, c412f4fcc)

1. **`--agent` emitted an unguarded `bd doctor --fix`.** Enrichers hardcode their own `Commands` (`agent.go` has 18 lines containing `bd doctor --fix` at c412f4fcc, including Schema Compatibility and Database Integrity) and never read the sanitized `dc.Fix`. `EnrichForAgent(dc, gate)` now sanitizes the finished `Commands` after enrichment, and `TestNoAgentEnricherLeaksUnguardedFixCommand` runs over every registered enricher (`doctor.AgentEnricherNames`) plus an unenriched name, with a non-vacuity check under a safe gate.
2. **`AssessSchemaFixGate` had no test.** `TestAssessSchemaFixGate_RealDolt` seeds a real `schema_migrations` in a Dolt test server (ahead, behind, equal, empty table, absent table, unreadable table) and `TestAssessSchemaFixGate_Unreachable` covers the unreachable gate. `TestFixGateAllowsFix` pins the admission table.
3. **An unopenable database withheld the fixes that repair it.** Rule: *a schema-writing fix is withheld whenever the gate blocks; a recovery fix is additionally allowed while the database is unreachable, and withheld again once it is reachable but skewed or undetermined.* Recovery fixes are `Corrupt Manifest` (moves `.dolt/` aside, plain `dolt init`), `Dolt Format` (seeds a marker file), `Dolt Schema` (plain read probe, writes only `metadata.json`) and `Database Integrity` (moves the local dolt dir aside and runs `bd init --force`, which opens only the store it just created; server-mode repos are refused up front by `serverModeIntegrityRecoveryGuard`). `Fresh Clone` and `Database` stay withheld because they open the existing store through the migrating factory. Names not listed are treated as schema-writing.

## Test plan

Built and run with `-tags gms_pure_go`, the tag CI uses.

- `go test ./cmd/bd/doctor/ -run 'FixGate|Sanitize|Filesystem|Mentions|Assess|Agent|Enrich|Recovery' -count=1` passes. `TestAssessSchemaFixGate_RealDolt` skips when no Dolt test server is available; it ran, not skipped, for this result (6 subtests).
- The gate wiring tests in `cmd/bd/doctor_fix_gate_wiring_test.go` pass.
- `golangci-lint run --config=.golangci.yml --build-tags=gms_pure_go ./cmd/bd/...` reports 0 issues.
- Red first, against the c412f4fcc code: with the sanitize loop in `EnrichForAgent` skipped, the every-enricher test fails, e.g. `check "Schema Compatibility" (error) published unguarded --fix command "bd doctor --fix" under a blocked gate`; with c412f4fcc's `doctor_fix.go`, `TestApplyFixesKeepsRecoveryFixesOnUnreachableGate` and `TestPreviewFixesLabelsOnlySchemaWritingFixesBlockedOnUnreachableGate` fail.
- Mutation checks, one at a time, each restored: flipping any branch of `AssessSchemaFixGate` (ahead, pending, equal, unreachable, undetermined; `case dbVer > binary:` to `case false:` fails `RealDolt/ahead`), or any branch of `AllowsFix`, or any `recoveryFixes` entry, turns a test red. Not covered by any test: the call sites inside the `doctor.go` Run closure (which gate it hands to `buildAgentResult`, `applyFixes`, `previewFixes`, and the `--check` refusal); the callees are pinned, the closure is not.

## Known limits

- A fresh unmigrated database (`schema_migrations` absent or empty) is treated as "could not be determined" and fails closed, so its schema-writing fixes are withheld. `TestAssessSchemaFixGate_RealDolt` pins this as current behaviour, not as an endorsement.
- `bd doctor --fix --dry-run` reports the same withholding the real run applies.
- Pre-existing and out of scope: `doctor.go` advertises a `--force` flag that does not exist.
- The red `Test (storage domain + uow)` job on c412f4fcc is a testcontainer connection drop in `internal/tracker`, which this PR does not touch.

### DIFF
--- a/cmd/bd/doctor.go
+++ b/cmd/bd/doctor.go
@@ -1078,6 +1078,9 @@ func exportDiagnostics(result doctorResult, outputPath string) error {
 }
 
 func printDiagnostics(result doctorResult) {
+	// GH#4993: rewrite Fix tips that steer operators into --fix under schema skew.
+	fixGate := doctor.AssessSchemaFixGate(result.Path)
+
 	// Pre-calculate counts and collect issues grouped by category
 	checksByCategory := make(map[string][]doctorCheck)
 	issuesByCategory := make(map[string][]doctorCheck)
@@ -1085,6 +1088,10 @@ func printDiagnostics(result doctorResult) {
 	hasIssues := false
 
 	for _, check := range result.Checks {
+		// Sanitize per-check fix recommendations before display
+		if check.Fix != "" {
+			check.Fix = doctor.SanitizeFixRecommendation(check.Fix, fixGate)
+		}
 		cat := check.Category
 		if cat == "" {
 			cat = "Other"
@@ -1222,6 +1229,11 @@ func printDiagnostics(result doctorResult) {
 			noun = "warnings"
 		}
 		fmt.Printf("%s\n", ui.RenderMuted(fmt.Sprintf("(%d %s suppressed via doctor.suppress config)", result.SuppressedCount, noun)))
+	}
+
+	// GH#4993: surface schema gate when --fix would be unsafe
+	if !fixGate.Safe && fixGate.Reason != "" {
+		fmt.Printf("\n%s Schema fix-gate: %s\n", ui.RenderWarn("⚠"), fixGate.Reason)
 	}
 }
 
--- a/cmd/bd/doctor.go
+++ b/cmd/bd/doctor.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -243,10 +244,26 @@ Examples:
 			return nil
 		}
 
+		// GH#4993: assess once, before any branch that can write. Lazy so
+		// read-only paths skip the probe; memoised so repeats cannot disagree.
+		schemaGate := sync.OnceValue(func() doctor.FixGate {
+			return doctor.AssessSchemaFixGate(absPath)
+		})
+
 		// artifacts, conventions, and pollution work in embedded mode and run
 		// unconditionally; validate still requires a server-mode connection
 		// and stays gated (GH#3597).
 		if doctorCheckFlag != "" {
+			// GH#4993: these handlers return directly, so their destructive
+			// paths bypassed the gate. Refuse at the single branch point.
+			if checkFlagWrites(doctorCheckFlag, doctorClean, doctorFix) {
+				if gate := schemaGate(); gate.DBReachable && !gate.AllowDBFix {
+					return HandleErrorWithHint(
+						fmt.Sprintf("refusing destructive 'bd doctor --check=%s': %s", doctorCheckFlag, gate.Reason),
+						"Re-run without --clean/--fix to inspect read-only, or resolve the schema state first",
+					)
+				}
+			}
 			switch doctorCheckFlag {
 			case "artifacts":
 				return runArtifactsCheck(absPath, doctorClean, doctorYes)
@@ -290,12 +307,17 @@ Examples:
 
 		result := runDiagnostics(absPath)
 
+		// GH#4993: guard once, on the result, before any emitter reads it.
+		// Per-renderer sanitizing exempted --json, --agent and --output.
+		sanitizeFixAdvice(&result, schemaGate())
+
 		if doctorDryRun {
-			previewFixes(result)
+			previewFixes(result, schemaGate())
 		} else if doctorFix {
-			applyFixes(result)
+			applyFixes(result, schemaGate())
 			fmt.Println("\nVerifying fixes...")
 			result = runDiagnostics(absPath)
+			sanitizeFixAdvice(&result, schemaGate())
 		}
 
 		if doctorOutput != "" || jsonOutput {
@@ -324,7 +346,7 @@ Examples:
 				return err
 			}
 		} else if doctorOutput == "" {
-			printDiagnostics(result)
+			printDiagnostics(result, schemaGate())
 		}
 
 		if !result.OverallOK {
@@ -1077,9 +1099,35 @@ func exportDiagnostics(result doctorResult, outputPath string) error {
 	return nil
 }
 
-func printDiagnostics(result doctorResult) {
-	// GH#4993: rewrite Fix tips that steer operators into --fix under schema skew.
-	fixGate := doctor.AssessSchemaFixGate(result.Path)
+// checkFlagWrites reports whether a `--check=<flag>` can modify state and so
+// must clear the schema gate (GH#4993). New destructive checks go here.
+func checkFlagWrites(flag string, clean, fix bool) bool {
+	switch flag {
+	case "artifacts", "pollution":
+		return clean
+	case "validate":
+		return fix
+	}
+	return false
+}
+
+// sanitizeFixAdvice rewrites each Fix tip in place so no emitter publishes
+// advice the gate ruled unsafe (GH#4993). Index-based write is load-bearing:
+// ranging by value over []doctorCheck mutates a copy.
+func sanitizeFixAdvice(result *doctorResult, gate doctor.FixGate) {
+	if result == nil {
+		return
+	}
+	for i := range result.Checks {
+		if result.Checks[i].Fix == "" {
+			continue
+		}
+		result.Checks[i].Fix = doctor.SanitizeFixRecommendation(result.Checks[i].Fix, gate)
+	}
+}
+
+func printDiagnostics(result doctorResult, gate doctor.FixGate) {
+	// GH#4993: tips arrive already sanitized; do not re-assess the gate here.
 
 	// Pre-calculate counts and collect issues grouped by category
 	checksByCategory := make(map[string][]doctorCheck)
@@ -1088,10 +1136,6 @@ func printDiagnostics(result doctorResult) {
 	hasIssues := false
 
 	for _, check := range result.Checks {
-		// Sanitize per-check fix recommendations before display
-		if check.Fix != "" {
-			check.Fix = doctor.SanitizeFixRecommendation(check.Fix, fixGate)
-		}
 		cat := check.Category
 		if cat == "" {
 			cat = "Other"
@@ -1231,9 +1275,9 @@ func printDiagnostics(result doctorResult) {
 		fmt.Printf("%s\n", ui.RenderMuted(fmt.Sprintf("(%d %s suppressed via doctor.suppress config)", result.SuppressedCount, noun)))
 	}
 
-	// GH#4993: surface schema gate when --fix would be unsafe
-	if !fixGate.Safe && fixGate.Reason != "" {
-		fmt.Printf("\n%s Schema fix-gate: %s\n", ui.RenderWarn("⚠"), fixGate.Reason)
+	// GH#4993: surface the schema gate verdict assessed once in RunE.
+	if gate.Reason != "" {
+		fmt.Printf("\n%s Schema fix-gate: %s\n", ui.RenderWarn("⚠"), gate.Reason)
 	}
 }
 
--- a/cmd/bd/doctor/fix_gate.go
+++ b/cmd/bd/doctor/fix_gate.go
@@ -3,123 +3,173 @@ package doctor
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/steveyegge/beads/schema"
 )
 
-// FixGate reports whether recommending or running `bd doctor --fix` is safe
-// given the relationship between the database schema and this binary (GH#4993).
-//
-// --fix is not a purely cosmetic repair: opening the store can apply pending
-// schema migrations, and a DB ahead of the binary can resume write migrations
-// the running binary does not understand. Operators who follow a printed
-// "run bd doctor --fix" tip during a change-freeze have been burned by this.
+// FixGate reports whether `bd doctor --fix` is safe given the schema/binary
+// relationship (GH#4993): opening the store can apply pending migrations.
+// Callers ask three distinct questions, so it answers three.
 type FixGate struct {
-	// Safe is true when --fix may be recommended / applied for general repair.
-	Safe bool
-	// DBVersion is the max applied schema_migrations version (0 if unknown / unavailable).
-	DBVersion int
-	// BinaryVersion is schema.LatestVersion() for this binary.
-	BinaryVersion int
-	// Reason is non-empty when Safe is false (or when a soft caution applies).
-	Reason string
-	// Ahead is true when the database schema is newer than this binary.
-	Ahead bool
-	// Pending is true when the binary knows migrations the DB has not applied.
-	Pending bool
+	// Determined is false when the DB was reachable but its version could not
+	// be read. Unknown is not safe.
+	Determined bool
+	// DBReachable is false when there is no database at all (no server, no
+	// repo) — no schema hazard exists, so filesystem repair is unaffected.
+	DBReachable bool
+
+	// RecommendFix allows printed advice to steer at `bd doctor --fix`.
+	RecommendFix bool
+	// AllowDBFix allows repairs that may write schema or resume migrations.
+	AllowDBFix bool
+	// AllowFSFix allows filesystem-only repairs, which never touch the schema.
+	AllowFSFix bool
+
+	DBVersion     int // max applied schema_migrations version (0 if unknown)
+	BinaryVersion int // schema.LatestVersion() for this binary
+	Reason        string
+	Ahead         bool // DB schema newer than this binary
+	Pending       bool // binary knows migrations the DB has not applied
 }
 
-// AssessSchemaFixGate probes the Dolt SQL server for schema_migrations and
-// compares to this binary. If the DB cannot be opened (no server, no dolt),
-// returns Safe=true with empty reason so filesystem-only doctor paths still work.
+// AssessSchemaFixGate compares schema_migrations to this binary. Assess once
+// per invocation, before anything that can write, and thread the result
+// through: probing at print time cannot guard writes that already happened.
 func AssessSchemaFixGate(path string) FixGate {
 	binary := schema.LatestVersion()
-	gate := FixGate{Safe: true, BinaryVersion: binary}
+
+	// No DB reachable: scoped allowance, not a fail-open — DB fixes stay
+	// disallowed because there is no DB to fix.
+	unreachable := FixGate{
+		BinaryVersion: binary,
+		Determined:    true,
+		RecommendFix:  true,
+		AllowFSFix:    true,
+	}
 
 	beadsDir := ResolveBeadsDirForRepo(path)
 	db, _, err := openDoltDB(beadsDir)
 	if err != nil {
-		return gate
+		return unreachable
 	}
 	defer db.Close()
 
+	// Reachable but unreadable: fail closed rather than claim safety.
+	undetermined := FixGate{
+		BinaryVersion: binary,
+		DBReachable:   true,
+		AllowFSFix:    true,
+		Reason: "database schema version could not be determined; " +
+			"`bd doctor --fix` may apply migrations blind — resolve the database state first",
+	}
+
 	ctx := context.Background()
 	dbVer, err := schema.CurrentVersion(ctx, db)
-	if err != nil {
-		return gate
-	}
-	gate.DBVersion = dbVer
-	if dbVer == 0 {
-		return gate
+	if err != nil || dbVer == 0 {
+		return undetermined
 	}
 
-	if dbVer > binary {
-		gate.Safe = false
+	gate := FixGate{
+		Determined:    true,
+		DBReachable:   true,
+		DBVersion:     dbVer,
+		BinaryVersion: binary,
+		AllowFSFix:    true,
+	}
+
+	switch {
+	case dbVer > binary:
 		gate.Ahead = true
-		delta := dbVer - binary
-		unit := "migrations"
-		if delta == 1 {
-			unit = "migration"
-		}
 		gate.Reason = fmt.Sprintf(
-			"database schema is at v%d, this binary knows up to v%d (%d %s ahead). "+
+			"database schema is at v%d, this binary knows up to v%d (%s ahead). "+
 				"`bd doctor --fix` can write/migrate and is unsafe on a newer schema — upgrade bd first",
-			dbVer, binary, delta, unit,
+			dbVer, binary, pluralMigrations(dbVer-binary),
 		)
-		return gate
-	}
-
-	if dbVer < binary {
+	case dbVer < binary:
 		gate.Pending = true
-		// Soft gate: still allow --fix for filesystem fixes, but do not
-		// recommend bare --fix as a catch-all (it may apply schema migrations).
-		gate.Safe = false
-		delta := binary - dbVer
-		unit := "migrations"
-		if delta == 1 {
-			unit = "migration"
-		}
 		gate.Reason = fmt.Sprintf(
-			"database schema is at v%d, this binary expects v%d (%d pending %s). "+
+			"database schema is at v%d, this binary expects v%d (%s pending). "+
 				"`bd doctor --fix` may apply write migrations — upgrade/migrate deliberately, not as a cosmetic tip",
-			dbVer, binary, delta, unit,
+			dbVer, binary, pluralMigrations(binary-dbVer),
 		)
-		return gate
+	default:
+		gate.RecommendFix = true
+		gate.AllowDBFix = true
 	}
 
 	return gate
 }
 
-// SanitizeFixRecommendation rewrites Fix text that steers operators into
-// `bd doctor --fix` when the schema gate says that is unsafe (GH#4993).
+func pluralMigrations(n int) string {
+	if n == 1 {
+		return "1 migration"
+	}
+	return fmt.Sprintf("%d migrations", n)
+}
+
+// fixAdvicePattern matches a tip steering at `bd doctor --fix`, tolerating flag
+// order and spacing. Best-effort: it is still a sniff over free-form prose.
+var fixAdvicePattern = regexp.MustCompile(`(?i)\bdoctor\b[^\n]*--fix\b`)
+
+// originalTipMarker prefixes preserved advice in a rewritten tip, and doubles
+// as the idempotence sentinel.
+const originalTipMarker = "Original tip was: "
+
+// MentionsFixAdvice reports whether a tip steers at `bd doctor --fix`.
+func MentionsFixAdvice(fix string) bool {
+	return fixAdvicePattern.MatchString(fix)
+}
+
+// SanitizeFixRecommendation rewrites Fix text that steers into `bd doctor
+// --fix` when the gate says that is unsafe (GH#4993). Apply once to the
+// result, not per-renderer — see sanitizeFixAdvice in cmd/bd.
 func SanitizeFixRecommendation(fix string, gate FixGate) string {
-	if gate.Safe || fix == "" {
+	if gate.RecommendFix || fix == "" || !MentionsFixAdvice(fix) {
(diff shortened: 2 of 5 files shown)


maphew:
The diagnosis here is right and the issue is real — `bd doctor --fix` is not cosmetic, and refusing it against a skewed schema is the correct instinct. I'd like to land this rather than bounce it, so I've written up what I found in detail below and I'm happy to take the refactor maintainer-side if you'd rather not keep iterating.

Three things I verified against your branch rather than inferred:

**1. The sanitizer never reaches the JSON/agent surface.** `cmd/bd/doctor.go:1090` is `for _, check := range result.Checks`, so `check.Fix = SanitizeFixRecommendation(...)` mutates a range *copy*. Display is sanitized because the copy is what gets bucketed — but `result.Checks` itself is untouched, so `exportDiagnostics` (`:311`), `buildAgentResult` (`:317`) and `outputJSON` (`:327`) all still emit the raw `Run 'bd doctor --fix'`. Since `--json`/`--agent` are the documented agent-facing surface, that's the audience most likely to act on the tip without judgment, and it's the one still getting the unguarded advice.

**2. `--check=validate --fix` is an ungated destructive path.** `doctor.go:262` returns `runValidateCheck(absPath)` before `applyFixes` is ever reached, and that path writes to the DB via `applyValidateFixes` (`doctor_validate.go:34-36`). Its own tip at `doctor_validate.go:67` is a bare `fmt.Printf` rather than a `check.Fix`, so the sanitizer never sees it — and even if it did, the substring test at `fix_gate.go:97` looks for `"doctor --fix"`, which `"doctor --check=validate --fix"` doesn't contain. `--check=pollution --clean` (`doctor_pollution.go:139`, deletes issues) has the same shape.

Gated today: `--fix`, `--fix --yes`, `--fix -i`, `--dry-run`. Ungated: `--check=validate --fix`, `--check=pollution --clean`, `--check=artifacts --clean`.

**3. The gate is evaluated after `runDiagnostics` has already done a writable open — so the pending branch can disarm itself.** `doctor.go:509` calls `trackBdVersion()` and `:516` calls `autoMigrateOnVersionBump(beadsDir)`, which does a writable `dolt.NewFromConfig` (`version_tracking.go:217`), and a writable open runs `initSchema` → `MigrateUpWithLock`. All of that is inside `runDiagnostics`, before `applyFixes`/`previewFixes`/`printDiagnostics` evaluate the gate.

So on a version bump, a bare `bd doctor` can apply the pending migrations and *then* print "migrations pending, don't run `--fix`" — and because `trackBdVersion` wrote the version file first, the next invocation sees DB == binary and reports safe. The warning fires once and the condition it warns about is gone by the time anyone reads it.

The ahead branch isn't exploitable this way (`CheckForwardDrift` fails the open and `MigrateUp` no-ops when ahead), so this is confined to the pending case — but that's the half the PR body describes as handled "softly". Worth flagging that we hit this exact class of bug elsewhere in the codebase this week: a guard that fired *after* `Start()` had already written a port file, which the retry then read and silently accepted. It's a recurring shape here, not a one-off.

## Smaller items

- **Fails open silently on a detection error.** `fix_gate.go:52-64` returns `Safe: true` with an empty `Reason` when `openDoltDB` fails, `CurrentVersion` errors, or `dbVer == 0` — so nothing prints and `--fix` proceeds. The `openDoltDB` case is defensible (filesystem-only doctor must still work), but "server reachable, `SELECT MAX(version)` errored" is an unreadable schema state currently indistinguishable from verified-safe. A third `Unknown` state that at least prints a caution would close it.
- **Refusing `--dry-run` is backwards.** `doctor_fix.go:20-24` refuses the preview when ahead, but `previewFixes` writes nothing, and #4993 explicitly asks for exactly that enumeration. Refusing it removes the operator's only safe diagnostic mid-incident.
- **Filesystem-only fixes are blocked with no hatch.** `doctor_fix.go:66-71` returns before *any* fix, including Lock Files, Git Hooks, Circuit Breaker, Btrfs NoCOW — none of which touch the DB. Not honoring `BD_IGNORE_SCHEMA_SKEW` is the right call (that hatch is scoped to read risk), but the result is a path with no hatch at all. `applyFixList` is already a per-check-name switch, so classifying targets as DB-touching vs filesystem-only is small — and it's what #4993's own suggestion 2 asks for.
- **The tests don't test the guard.** All five cases in `fix_gate_test.go` operate on hand-built structs; `TestFixGateRefuseDoctorFix` asserts `FixGate{Safe:false,Ahead:true}.RefuseDoctorFix() == true`, which is a tautology over the literal. Nothing exercises `AssessSchemaFixGate`, and no test calls `applyFixes`/`previewFixes` at all — deleting the gate blocks leaves the suite green. Related to point 3: no test invokes a guarded path *twice*, which is the only way a self-disarming guard shows up.
- `fix_gate.go:124` `return g.Ahead && !g.Safe` — `&& !g.Safe` is dead, and dead in the fail-open direction. `fix_gate.go:97`'s second `strings.Contains` is subsumed by the first. `fix_gate.go:101-102`'s `alt := fix` has a comment describing alternative-parsing the code doesn't do.
- `doctor_fix.go:66` prints `re-check AssessSchemaFixGate`, an internal Go symbol, in user-facing output.
- The pending branch sets `Safe = false`, so on any repo where the DB is intentionally behind (`BD_ALLOW_REMOTE_MIGRATE` keeps ours there persistently) *every* bare `bd doctor` rewrites *every* fix tip. Worth a deliberate call on whether that's the intended noise level.

## Suggested shape

One refactor collapses items 1, 2, 3 and the redundant probing: **assess the gate once at the top of `RunE`, before `runDiagnostics`**, sanitize into `result.Checks` rather than at print time, and thread the gate into the `--check=…` handlers. It only needs a raw MySQL connection, which `openDoltDB` already provides, so hoisting is cheap. Splitting `applyFixList` targets into DB-touching vs filesystem-only is the second half and resolves the no-hatch problem.

Say the word and I'll push that onto your branch rather than asking you to redo it — the analysis is yours and I'd rather keep it that way.

vishnujayvel:
Thank you — this is a genuinely good review, and all three of the main findings are real. I checked
each against the branch rather than taking them on trust, and they hold:

1. `doctorCheck` is a struct in a `[]doctorCheck`, so the `for _, check := range result.Checks`
   sanitize loop was mutating a copy. Worse than that: the call sits inside `printDiagnostics`,
   which is the `else` branch of the emitter chain — so `--json`, `--agent` and `--output` never
   reached it at all. The agent-facing path, the one most likely to act on the advice unattended,
   was the least protected.
2. The `--check=` switch returns before any gate is assessed, so `--check=pollution --clean` and
   `--check=validate --fix` were destructive paths outside the guard entirely.
3. `runDiagnostics` calls `autoMigrateOnVersionBump` near its top, and the gate was assessed later
   in `printDiagnostics`. A bare `bd doctor` therefore applied pending migrations and *then* warned
   about them.

On your offer to push it onto the branch — thank you, that's generous, and I'd rather do it myself
if that's alright. Working through it, these aren't three separate bugs: they're one placement
fault. The guard was installed downstream of the hazard it guards, so it could only ever protect
one output format and only after the writes had happened. A patch aimed at the three symptoms
would most likely have fixed the copy bug and duplicated the sanitize call into the other three
emitters — which re-creates the same precondition for whoever adds a fifth emitter later. I'd
rather fix the shape.

Pushed as `164e82978`:

- **Assess the gate once**, in the `RunE` preamble before any branch that can write, and thread
  that single verdict through. Lazily evaluated so filesystem-only checks don't pay for a DB probe
  they never consult, memoised so two consultations can't disagree within one run.
- **Sanitize once, into the result**, before any emitter reads it — rather than per-renderer. A new
  output format is now safe by default, because unguarded text no longer exists in the struct.
- **Gate the `--check=` paths** at the single branch point, keyed off a small `checkFlagWrites`
  table, so a new destructive check can't bypass the guard by adding another early return.

On the smaller points, all of which were also correct:

- **Fail-open on detection error** — split into `Determined` and `DBReachable`. "The DB is reachable
  but I couldn't read its version" now fails closed. "There is no DB at all" still permits
  filesystem repair, since there's no schema hazard to protect against; that case was conflated
  with the first one before.
- **`--dry-run` refusal was backwards** — a dry run writes nothing and is exactly what someone under
  schema skew should be able to run. It now always runs and labels which fixes would be skipped.
- **No hatch for filesystem-only fixes** — `applyFixes` used to return early and block everything.
  Fixes are now partitioned by blast radius, so gitignore/permissions/lock-file repairs proceed
  while schema-touching ones are withheld. The partition is an allowlist with a fail-closed
  default: an unclassified check name is treated as database-touching.
- **Dead `&& !g.Safe`** — gone. `Ahead` was only ever set alongside `Safe: false`, so the conjunct
  could never change the result. The single overloaded `Safe` flag is now three fields
  (`RecommendFix` / `AllowDBFix` / `AllowFSFix`) because callers were genuinely asking three
  different questions of it.
- **`AssessSchemaFixGate` leaking into user-facing output** — fixed.
- **The string sniff** — widened to a pattern tolerating flag order and spacing (`bd doctor --yes
  --fix` etc.). I want to be straight that this is still a text sniff over prose and can still be
  fooled. The durable fix is structured fix advice instead of a string, so a renderer *cannot* emit
  unguarded advice — but that changes every check constructor and the `--json` schema, so it felt
  wrong to fold into this PR. Happy to open it separately if you'd want it.

On the tests — your "tautologies over literals" is the most useful thing in the review. They
constructed a `FixGate` by hand and asserted the predicate rewrote a string, which is why they
stayed green while three of four output paths leaked. Every defect here was in the wiring, and
nothing tested the wiring. Replaced with a property over the whole emitter set — for a not-safe
gate, no emitter may produce unguarded advice — plus the `--check=` enumeration and a fail-closed
case. I verified they have real detection power by re-introducing the original range-by-value bug
and confirming they fail, rather than just checking they pass now.

Test run, honestly: `go test ./cmd/bd/doctor/...` and the doctor-related tests in `./cmd/bd/`
pass, with two exceptions I want to name rather than bury. `TestCheckBeadsRole_NotConfigured` and
`TestCheckBeadsRole_NotGitRepo` fail on my machine because `CheckBeadsRole` shells out to `git
config --get beads.role` and picks up my global config — I confirmed these fail identically at
`a45199a54`, before this PR's first commit existed, so they're environmental and pre-existing, not
mine. Dolt-backed tests were skipped: Docker isn't available here (`WARN: Docker not available,
skipping Dolt tests`), so I can't claim coverage of those paths.

One thing I'd flag as deliberately *not* fixed: `runDiagnostics` still mutates. Hoisting the
assessment means the reading is now taken before the migration, so the reported state is honest,
but the underlying oddity — that diagnosis applies migrations at all — is still there. Making
`bd doctor` read-only would be the root fix and would make this whole class impossible, but it's a
behaviour change to a documented side effect (the auto-migrate was added deliberately in bd-jgxi
because doctor skips `PersistentPreRun` DB init). That's your call, not mine to make inside a
review-response commit. Happy to file it as its own issue if you think it's worth having.

Your observation that this is a recurring shape here — the guard that fired after `Start()` had
already written the port file — is the part I've taken most from. Same failure, different surface:
the predicate was fine and the placement wasn't.

maphew (commented):
Reviewed at `164e82978c5`.

The schema-skew protection is valuable and comprehensive across emitters and destructive `--check` paths, but one blocker defeats the core gate: `applyFixes` admits DB fixes when `gate.Reason == ""`. The unreachable gate intentionally has `AllowDBFix=false` and an empty reason, so a stopped or unreachable server can fail assessment and still reach database repairs, which may reopen and migrate the database. `previewFixes` has the same misleading condition.

Gate database fixes solely on `AllowDBFix`, treat unreachable assessment as filesystem-only, and add a wiring regression for the unreachable gate. This is a small maintainer fix on the contributor branch.

Verification:

- Focused `cmd/bd/doctor` tests — PASS (`ok`, 2.180s).
- Focused `cmd/bd` fix-advice/check tests — PASS (`ok`, 3.607s).
- Virtual merge onto current `upstream/main` — clean.
- Blocking upstream preflight — blocked on `UNSTABLE` and four failed/action-required checks.

Disposition: **fix-merge**; preserve the submitted safety work and fix this remaining bypass before merge.

vishnujayvel:
Confirmed — this is a real bypass, not a cosmetic gap. `AssessSchemaFixGate` returns `unreachable := FixGate{BinaryVersion: binary, Determined: true, RecommendFix: true, AllowFSFix: true}` when `openDoltDB` fails (`cmd/bd/doctor/fix_gate.go`, the `unreachable` gate and its early return on `err != nil`), which leaves `AllowDBFix` and `Reason` at their zero values: `false` and `""`. `applyFixes`'s admission check is `if gate.AllowDBFix || gate.Reason == ""` (`cmd/bd/doctor_fix.go`) — for that exact gate shape, `AllowDBFix` is false but `Reason` is empty, so the `|| gate.Reason == ""` branch fires and every DB fix is appended to `fixableIssues` anyway. `previewFixes`'s `blockDB := !gate.AllowDBFix && gate.Reason != ""` has the mirrored defect: `!AllowDBFix` is true, `Reason != ""` is false, so `blockDB` evaluates to `false` and the dry-run doesn't flag the DB fixes as blocked either. A stopped or unreachable Dolt server hits this path directly through the `err != nil` branch in `AssessSchemaFixGate` — no schema version was ever read, and `--fix` (or even a dry run) will proceed to write/migrate the database as if the gate had cleared it.

I'll gate both functions solely on `AllowDBFix` — `if gate.AllowDBFix` in `applyFixes`, `blockDB := !gate.AllowDBFix` in `previewFixes` — so an unreachable assessment is treated as filesystem-only by construction, matching what the `unreachable` gate's own field values already say. I'll add a wiring regression alongside the existing `doctor_fix_gate_wiring_test.go` suite that drives `applyFixes`/`previewFixes` (or the equivalent gate-shaped fixture) with `AllowDBFix: false, Reason: ""` and asserts no DB fix is applied or presented as unblocked — I checked that file and no existing case (`TestNoEmitterLeaksUnguardedFixAdvice`, `TestSanitizeFixAdviceWritesThroughSlice`, `TestSanitizeFixAdviceIsIdempotent`, `TestCheckFlagWrites`, `TestCollectFixableIssuesPartitionsByBlastRadius`) exercises this empty-`Reason`/`AllowDBFix=false` combination.

vishnujayvel:
Rebased onto current upstream/main (no behaviour change — 3 commits replayed cleanly, no conflicts) to fix a stale-head CI trigger issue: this PR was previously reading green/mergeable with only 1 check (`[code]smith`, SKIPPED) instead of the ~120 checks a healthy PR runs here.

Post-rebase, CI now runs the full suite: **121 checks total**. Confirms the fix — the real CI workflow simply never ran on the old (stale) head.

Two real findings surfaced now that CI actually runs, both from the same root cause — pre-existing in this PR's own diff, not introduced by the rebase:
- `Lint` and `PR Lint (wrapper timing)` both fail on `cmd/bd/doctor_fix.go:40:20`: `` `labelled` is a misspelling of `labeled` `` (misspell linter).
- This in turn fails the aggregate `CI Gate / Required` check.

This is a one-word comment typo — trivial fix, but out of scope for this rebase-only pass. Filed as a follow-up.

bee-ghosttrack (changes requested):
REQUEST CHANGES at c412f4fcc (bee). Read the five files against `main` 71c4cd08b (the rebase is clean, zero behind); built, vetted and ran `golangci-lint` on the two changed files (0 issues, the `labelled` typo is gone); ran the seven `cmd/bd/doctor` gate tests and the seven `cmd/bd` wiring tests, all green. The three findings from maphew's 07-30 round are addressed as claimed for two of them: `--check=` is gated before the switch (`doctor.go:266-275` via `checkFlagWrites`), and admission is now `AllowDBFix` alone (`doctor_fix.go:118`, `previewFixes` at `:79`), so the `Reason == ""` bypass is closed. I traced every caller into fix application (`applyFixes`, `previewFixes`, `applyFixesInteractive`, `applyFixList`, `fixPendingMigrations`, `applyValidateFixes`, the pollution/artifacts `--clean` paths) and each is behind the gate; `--perf` and `--check-health` return earlier but write nothing. The first finding, the emitter coverage, is fixed for text, `--json` and `--output` and still open for `--agent`. One blocker, two majors.

**1. `--agent` still publishes an unguarded `bd doctor --fix`.** `cmd/bd/doctor/agent.go:44-53`: when a check name has an entry in `agentEnrichers`, `EnrichForAgent` sets `ad.Commands = e.commands` from the enricher and returns, never reading the sanitized `dc.Fix`. Eighteen enrichers hardcode `bd doctor --fix`, and they include `enrichSchemaCompatibility` (`agent.go:197`, `[]string{"bd migrate", "bd doctor --fix"}`) and `enrichDatabaseIntegrity`, the checks that fire on the very skew this PR exists to guard. `TestNoEmitterLeaksUnguardedFixAdvice` passes vacuously because its fixture is `Name: "Tracked Runtime Files"` (`doctor_fix_gate_wiring_test.go:44`), which has no enricher and falls to the generic `Commands = []string{dc.Fix}` branch. A throwaway copy of that test with `Name: "Schema Compatibility"` and a blocked gate emits `"commands":["bd migrate","bd doctor --fix"]`. The agent surface is the one most likely to act on the advice unattended, as you said yourself on 07-30. Sanitize `AgentDiagnostic.Commands` after enrichment (or pass the gate into `EnrichForAgent`), and run the emitter test over an enriched name as well as an unenriched one.

**2. `AssessSchemaFixGate` has no test, and the headline guard is unpinned.** Changing `fix_gate.go:83` from `case dbVer > binary:` to `case false:` deletes ahead detection outright; `go test ./cmd/bd/doctor/ -run 'FixGate|Sanitize|Filesystem|Mentions|Assess'` and the seven `cmd/bd` wiring tests both stay green. Every test builds a `doctor.FixGate` literal by hand, so the function that decides ahead / pending / unreachable is never exercised. (A control mutation, `if true` at `doctor_fix.go:118`, is caught by `TestApplyFixesWithholdsDBFixOnUnreachableGate`, so the wiring is pinned; the assessor is not.) A table test over a seeded `schema_migrations` with the existing dolt test-server helpers would close this.

**3. An unopenable database now withholds the fixes that repair an unopenable database.** `fix_gate.go:52-57`: any `openDoltDB` error yields the `unreachable` gate with `AllowDBFix: false`, and `applyFixes` then skips every DB-classified fix. That class includes `Corrupt Manifest` (`doctor_fix.go:476`), `Database Integrity` (`:375`), `Fresh Clone` (`:445`), `Dolt Format` (`:472`) and `Dolt Schema` (`:470`), which are exactly the repairs a user reaches for when the server is dead or the manifest is corrupt. `TestApplyFixesWithholdsDBFixOnUnreachableGate` asserts this as intended, but it is a schema gate blocking non-schema recovery. Distinguishing "no database configured" from "database present but unopenable", or carving the recovery fixers out of the DB-write class, would keep the incident guard without losing the recovery path. (Pre-existing and not yours: `doctor.go:176` advertises a `--force` flag that does not exist.)

Smaller, none blocking: `doctor.go:269` refuses on `DBReachable && !AllowDBFix` while `doctor_fix.go:118` withholds on `!AllowDBFix`, two policies for one question that a single method on `FixGate` would collapse; `fix_gate.go:70` `if err != nil || dbVer == 0` folds "table absent" (`currentVersion` returns `0, nil` when `schema_migrations` does not exist) into "could not be determined", so a fresh unmigrated database is told its version is unknown and loses DB fixes, which I read but did not exercise; `doctor_fix.go:81` `!gate.AllowFSFix` is unreachable since all three returned gates set it true.

**CI:** `Test (storage domain + uow)` and its `CI Gate / Required` rollup are red on `TestEnginePull*`/`TestEnginePush*` with `invalid connection` against the shared `tracker_pkg_shared` server; unrelated to this diff, a re-run should clear them. `Lint` passes at this head. No CHANGELOG entry is required here (`CONTRIBUTING.md:310-315` scopes it to the release process).

vishnujayvel:
The one red check here looks like CI infrastructure rather than the change — could a maintainer re-run `Test (storage domain + uow)` when convenient?

The job failed on a Docker testcontainer connection drop against the shared test DB. This PR's diff only touches `cmd/bd/doctor_fix.go` and `cmd/bd/doctor_fix_gate_wiring_test.go` — nothing under `storage/`, `engine/`, `uow/`, or the Dolt server path — so there's no route by which it could affect that package's container. The same job passes on #5134, #5139, #6492 and #6550.

I can't re-run it myself (`gh run rerun` needs admin on this repo), and I'd rather not push a no-op commit just to re-trigger it. Everything else is green: 118 of 121 checks pass, with only the `CI Gate / Required` aggregate red underneath it.

Happy to push a re-trigger instead if you'd prefer that to a manual re-run.

bee-ghosttrack:
I read the failed job the same way: a testcontainer connection drop in a package this diff does not touch, so it says nothing about the change. I cannot re-run it either; `gh run rerun` needs org admin on gastownhall, which none of the reviewers on this queue hold, and the ask is with the org admins already (same wall as #6010 and #6588). The gate on this PR is my REQUEST CHANGES at c412f4fcca, not that job: the `--agent` emitter still publishes an unguarded `bd doctor --fix` (the blocker) plus the two majors above. A push addressing those will re-trigger CI on its own, so no no-op commit is needed.

vishnujayvel:
Thanks @bee-ghosttrack, all three findings were right. Fixed at 944d76171 (fast-forward from c412f4fcc); CI is green on this head, including `Lint`, `Check formatting`, `Test (macos-latest)` and `Test (storage domain + uow)`.

**1. Blocker: `--agent` emitted an unguarded `bd doctor --fix`.**
`EnrichForAgent` (`cmd/bd/doctor/agent.go:42-46`) now runs `SanitizeFixRecommendation` over the finished `Commands` after enrichment, so every enricher is covered without touching the 18 hardcoded strings. `TestNoAgentEnricherLeaksUnguardedFixCommand` (`cmd/bd/doctor_fix_gate_wiring_test.go:162`) iterates every registered enricher (`doctor.AgentEnricherNames`) plus an unenriched name, and has a non-vacuity check under a safe gate.
- Red first: with the sanitize loop skipped, it fails with `check "Schema Compatibility" (error) published unguarded --fix command "bd doctor --fix" under a blocked gate`.
- Green: passes with the fix.

**2. Major: `AssessSchemaFixGate` had no test.**
`TestAssessSchemaFixGate_RealDolt` (`cmd/bd/doctor/fix_gate_assess_dolt_test.go:79`) seeds a real `schema_migrations` in a Dolt test server: ahead, behind, equal, empty table, absent table, unreadable table (6 subtests, ran rather than skipped). `TestAssessSchemaFixGate_Unreachable` (`fix_gate_assess_test.go:57`) covers the unreachable gate.
- Mutations, one at a time, each restored: flipping `case dbVer > binary:` to `case false:` (`fix_gate.go:83`) now fails `RealDolt/ahead`. Flipping the pending, equal, unreachable and undetermined branches, any branch of `AllowsFix`, or any `recoveryFixes` entry also turns a test red.
- Still unpinned, and disclosed in the PR body: the call sites inside the `doctor.go` Run closure. The callees are pinned; the closure's choice of gate is not.

**3. Major: an unopenable database withheld the fixes that repair it.**
Rule (`fix_gate.go:177-217`, `recoveryFixes` / `AllowsFix`): a schema-writing fix is withheld whenever the gate blocks. A recovery fix (`Corrupt Manifest`, `Dolt Format`, `Dolt Schema`, `Database Integrity`) is additionally allowed while the database is unreachable, and withheld again once it is reachable but skewed or undetermined. `Fresh Clone` and `Database` stay withheld because they open the existing store through the migrating factory.
- Red first, against c412f4fcc's `doctor_fix.go`: `TestApplyFixesKeepsRecoveryFixesOnUnreachableGate` and `TestPreviewFixesLabelsOnlySchemaWritingFixesBlockedOnUnreachableGate` fail.
- Green with the fix.

I also corrected a comment that overstated `Database Integrity`: it runs `bd init --force`, which only opens the store it just created, and server-mode repos are refused up front. The PR body and commit message now state this rule.

Lint was `golangci-lint run --config=.golangci.yml --build-tags=gms_pure_go ./cmd/bd/...`, 0 issues. The earlier red `Test (storage domain + uow)` on c412f4fcc was the `internal/tracker` testcontainer connection drop, and it is green on this head.

bee-ghosttrack (approved):
Delta read of `c412f4fcc..944d76171` against my 09-18 review. All three findings are closed as described.

1. `EnrichForAgent(dc, gate)` sanitizes the finished `Commands` after enrichment, and `buildAgentResult` passes `schemaGate()` from the one `--agent` call site, so the eighteen hardcoded `bd doctor --fix` strings are covered without editing any enricher. `TestNoAgentEnricherLeaksUnguardedFixCommand` iterates `AgentEnricherNames()` plus an unenriched name under a blocked gate, with a non-vacuity check under a safe gate. I re-ran your mutant — skipping the sanitize loop — and it fails on `Child-Parent Dependencies` and `Clone-Local FKs` with the exact "published unguarded --fix command" message.
2. `TestAssessSchemaFixGate_RealDolt` runs (not skipped) against a seeded `schema_migrations` for ahead / behind / equal / empty / absent / unreadable, and `_Unreachable` covers the rest; the assessor is pinned.
3. `FixGate.AllowsFix` is now the single admission policy, and the recovery set (`Corrupt Manifest`, `Database Integrity`, `Dolt Format`, `Dolt Schema`) is admitted only while `!DBReachable`, withheld again once the database is reachable but skewed or undetermined; `Fresh Clone` and `Database` stay withheld because they open the existing store through the migrating factory. `applyFixes` and `previewFixes` both route through `AllowsFix` and neither admits on `Reason == ""`.

Ran on the head: `go build`, `go vet`, `gofmt -l` clean; `go test -tags gms_pure_go ./cmd/bd/doctor -run 'FixGate|Sanitize|Filesystem|Mentions|Assess|Recovery|Enricher'` (13 tests) ok; `./cmd/bd -run 'FixGate|NoAgentEnricherLeaks|NoEmitterLeaks|KeepsRecoveryFixes|LabelsOnlySchemaWriting|Withholds'` (6 tests) ok. Head is one commit behind `main`, no overlap. All CI checks green on `944d76171`.

Approving at `944d76171`. Merge is the maintainer's call.

PR #5150, opened 2026-07-29
feat(storage): content-derived ids for events/comments at insert time (bd-ri8bd)

steveyegge:
## Summary

Resolves Protocol v0.1 **C-OQ3** (bd-ri8bd): unversioned newest-wins replication has no merge pass to reconcile independently-created identical rows, so convergence must be a property of the primary key itself. This PR makes every `events`/`comments`/`compaction_snapshots` insert derive its id from row content, using the exact derivation the bd-6dnrw.2 aux-row backfill already shipped (`internal/storage/rowid`, frozen digest column lists), and runs one catch-up rekey for the interim UUIDv7 rows.

**Stacked on #5149** (its migration 0060 sits under this PR's 0061 anchor); retarget to `main` after #5149 merges.

Protocol acks: hopper (wyvern side of the v0.2 storage-classes review), mail m-1785313459-1237 — cross-replica collapse anchored to C2.3, second-resolution `created_at` frozen, catch-up-rekey shape endorsed. Design record on bd-ri8bd.

## What changed

- **`issueops/derivedid.go` (new)** — the single derivation point. `InsertDerivedEvent` (smallest-free-ordinal over same-content rows: local multiplicity preserved, cross-replica byte-identical rows deliberately collapse — the C2.3 set-union reading, and byte-for-byte what the versioned automerge union rule already produces), `InsertDerivedComment` (collapses onto an identical existing row, the import path's historical existence-check semantics), `InsertDerivedCompactionSnapshot`. `created_at` is stamped app-side and bound as the digested `DATETIME(0)` string, so the stored cell is byte-for-byte what was digested. The single-writer-per-replica ordinal assumption is documented (hopper's non-blocking flag).
- **All 16 mint sites converted**; `NewEventID` deleted (the "events have no natural identity, random is fine" premise it documented is exactly what C-OQ3 overturns). Import paths still honor incoming comment ids. Wisp-plane and promotion-carried ids stay minted-once by design (clone-local; random-but-consistent).
- **Catch-up rekey** — the backfill machinery generalized over (marker, shipped-version, sentinel) and run a second time: ignored marker `0018`, in-progress sentinel `aux_row_rekey2_in_progress`, anchored on main `0061`. The anchor must be its own version: a 0060-era lineage can still hold unconverged UUIDv7 rows, and a fresh clone skipping the pass there would diverge from its origin's later convergence. Clone-local and deliberately not migrator-exclusive — deterministic derivation is what makes that safe (V2 designated-migrator compatible, same reasoning as the first pass).
- **Tests** — equivalence test recomputes digests from server-rendered stored values and asserts every row minted by the real store paths sits on its backfill-convergent id (catches datetime-format or NULL-encoding drift at test time, not in a fleet merge); ordinal-preservation and comment-collapse contracts; existing rekey suite re-pointed at the generalized pass.

## Notes for review

- Keyset paging keeps its stable-total-order contract; the same-second tie-break becomes hash order (UUIDv7 time-locality loss noted-acceptable in the protocol review).
- Same-id automerge conflicts are now exactly the identical-content case, i.e. always auto-resolvable.
- Docs deferred to the cli-docs.pin bump per policy (#5052/#5089 precedent).

### DIFF
--- a/internal/storage/dolt/derived_id_insert_test.go
+++ b/internal/storage/dolt/derived_id_insert_test.go
@@ -1,0 +1,227 @@
+package dolt
+
+import (
+	"context"
+	"database/sql"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/storage/issueops"
+	"github.com/steveyegge/beads/internal/storage/rowid"
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// derivedIDDigestColumns mirrors the FROZEN digest column lists in
+// internal/storage/schema (auxRekeyTables) and the insert-time derivation in
+// issueops. Deliberately duplicated here: if either side drifts from the
+// frozen lists, the recomputation below stops matching the stored ids and
+// these tests fail.
+var derivedIDDigestColumns = map[string]string{
+	"events":   "issue_id, event_type, actor, old_value, new_value, comment, CAST(created_at AS CHAR)",
+	"comments": "issue_id, author, text, CAST(created_at AS CHAR)",
+}
+
+// assertTableIDsContentDerived scans every row of table, recomputes the
+// content digest from the stored (server-rendered) column values, and asserts
+// the table's id set is exactly the derived ids for ordinals 0..n-1 per
+// digest group — i.e. the ids the schema backfill would have assigned. This
+// is the insert-time/backfill equivalence at the heart of bd-ri8bd: a row
+// minted by the new insert paths must already sit on its convergent id.
+func assertTableIDsContentDerived(ctx context.Context, t *testing.T, db *sql.DB, table string) {
+	t.Helper()
+	columns := derivedIDDigestColumns[table]
+
+	rows, err := db.QueryContext(ctx, "SELECT id, "+columns+" FROM "+table) //nolint:gosec // test-local constant tables
+	if err != nil {
+		t.Fatalf("scan %s: %v", table, err)
+	}
+	defer rows.Close()
+
+	nFields := strings.Count(columns, ",") + 1
+	groups := make(map[string][]string)
+	for rows.Next() {
+		var id string
+		fields := make([]sql.NullString, nFields)
+		dests := make([]any, 0, nFields+1)
+		dests = append(dests, &id)
+		for i := range fields {
+			dests = append(dests, &fields[i])
+		}
+		if err := rows.Scan(dests...); err != nil {
+			t.Fatalf("scan %s row: %v", table, err)
+		}
+		groups[rowid.Digest(fields)] = append(groups[rowid.Digest(fields)], id)
+	}
+	if err := rows.Err(); err != nil {
+		t.Fatalf("scan %s: %v", table, err)
+	}
+	if len(groups) == 0 {
+		t.Fatalf("no %s rows to verify", table)
+	}
+
+	for digest, ids := range groups {
+		want := make([]string, len(ids))
+		for i := range ids {
+			want[i] = rowid.New(table, i, digest)
+		}
+		got := append([]string(nil), ids...)
+		sort.Strings(got)
+		sort.Strings(want)
+		for i := range want {
+			if got[i] != want[i] {
+				t.Errorf("%s digest %s: ids = %v, want derived %v", table, digest[:12], got, want)
+				break
+			}
+		}
+	}
+}
+
+// TestInsertTimeIDsMatchBackfillDerivation drives the real store write paths
+// (issue create with labels, update, comment) and verifies every resulting
+// events/comments row landed on its content-derived id — recomputed from the
+// server-rendered stored values, so a drift between the Go-side digest inputs
+// (app-stamped created_at text, NULL-vs-empty encoding) and what the column
+// actually holds fails here, not in a fleet merge.
+func TestInsertTimeIDsMatchBackfillDerivation(t *testing.T) {
+	store, cleanup := setupTestStore(t)
+	defer cleanup()
+
+	ctx, cancel := testContext(t)
+	defer cancel()
+
+	issue := &types.Issue{
+		Title:       "derived id fixture",
+		Description: "insert-time derivation",
+		Status:      types.StatusOpen,
+		Priority:    2,
+		IssueType:   types.TypeTask,
+		Labels:      []string{"lane-a"},
+	}
+	if err := store.CreateIssue(ctx, issue, "tester"); err != nil {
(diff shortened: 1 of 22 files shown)


steveyegge:
CI note: workflows here trigger on PRs targeting main only, so checks are absent while this is stacked on the #5149 branch. Full CI will run when this retargets to main after #5149 merges; the equivalent suites were run locally (storage tree, schema, cmd/bd with the embedded graph/create suites) and are green. — beads

steveyegge:
## Reviewer verdict (lion): APPROVE — merge when ready

Reviewed all four commits against the bd-ri8bd design record and hopper's protocol ACK (m-1785313459-1237). CI green (84 checks incl. embedded-Dolt conformance and cross-version smoke) on the branch tip 3c74a2ddb.

**What I verified:**

- **Derivation fidelity.** The frozen digest column lists agree at all three points — `auxRekeyTables` in schema, the insert-time digests in `derivedid.go` (field order matches per table), and the test's deliberately-duplicated mirror in `derived_id_insert_test.go`. `created_at` is app-stamped as the DATETIME(0) text and bound verbatim, so stored = digested; the equivalence test recomputing from server-rendered values is the right tripwire for drift.
- **All 16 mint sites converted, no stragglers.** `NewEventID` is gone with zero remaining references; the only raw inserts left into the aux tables are the deliberate incoming-id import path (`PersistComments`) and the dependencies tables (own derivation scheme). `issue_snapshots` correctly gets no insert-time derivation — it has no live mint path (cleanup-only).
- **NULL-vs-`''` semantics preserved per site.** Sites that previously bound `''` now pass `str("")`; sites that previously omitted columns pass invalid NullStrings → NULL for `events` (its default) and coerced `''` for `wisp_events` (its 0021 default) via the commit-4 fix. That matches the conformance oracle's pinning and keeps digests convergent with legacy rows in both planes.
- **`compaction_snapshots` digest round-trip.** `snapshot_json` is BLOB, so the bound bytes are the stored bytes — no JSON re-rendering hazard; `compaction_level` digests as its decimal rendering on both the insert and rekey sides.
- **Rekey generalization.** The three bd-578h9 gates (marker / shipped-version / crash sentinel) survive intact per pass; `anyAuxRekeyResumePending` correctly feeds MigrateUp's dirty-table exemption across both passes; anchoring the second pass on 0061 rather than 0060 is right for the reason the migration comment gives. The lock_test `latest-2` adjustment (0061 being a pure anchor) is correct.
- **Mixed-version window.** An old-binary clone's interim UUIDv7 rows are single-origin and merge-replicated (consistent everywhere), and its eventual upgrade rekey is a PK rewrite whose *diff* replicates under the versioned merge — so the catch-up converges fleet-wide as long as it completes before bd-red8u flips events to unversioned, which is exactly the dependency ordering the beads record.
- **Ordinal discipline.** Local multiplicity via smallest-free-ordinal, cross-replica collapse per C2.3, within-replica concurrent-duplicate loses on the PK (documented, matches the single-writer topology). Same-content scan is bounded by `idx_events_issue`.

**Non-blocking notes** (fold into a later cleanup if you like, no re-review needed):

1. `comments.go` `ImportIssueCommentInTx`: a `ParseAuxTime` failure is wrapped as `"add comment to %s"` — misleading label for a parse error (same shape in `transaction.go`'s `"failed to add comment"`, which is at least generic).
2. `FormatAuxTime` truncates sub-second input where the DB previously did its own rounding on a `time.Time` bind — a one-time sub-second behavior shift on import paths, and the deterministic direction is the right one; just noting it's deliberate.
3. Every event insert now carries one extra issue-scoped SELECT. Fine at current volumes and indexed; worth remembering if a bulk hot path ever shows up in a profile.

Nice catch by the conformance oracle on the wisp NULL regression — and the fix landing as digest-input coercion (not just bind coercion) is exactly what keeps the rekey re-derivation convergent.

maphew:
Heads-up on a likely regression from this change — flagging rather than asserting, since I have not reproduced it locally (details on why below).

**Symptom:** `Test (Proxied Dolt Cmd 14/15)` is failing on at least two independent PRs — #5143 and #5159 — both on `cmd/bd/comment_proxied_integration_test.go:108`, `TestProxiedServerComment/multiple_comments_ordered`:

```
comment_proxied_integration_test.go:122: unexpected ordering: "three", "two", "one"
```

Neither PR touches comments, and `main`'s Nightly Full Tests was green at 03:06, so it reads as probabilistic rather than a hard break.

**Proposed mechanism:**

- Comments are listed with `ORDER BY created_at ASC, id ASC` — `internal/storage/issueops/comments.go:29` and `:88`, and `internal/storage/domain/db/comment.go:87`.
- After this PR, a comment's `id` is content-derived, and per `derivedid.go:25-27` `created_at` participates only as its **`DATETIME(0)`** text rendering — one-second granularity.
- So comments written inside the same second tie on the primary sort key, and the `id ASC` tiebreaker is now a content digest rather than anything carrying insertion order.

Since `InsertDerivedComment`'s inputs include `issueID` (`comments.go:253`), which is fresh per run, the resulting permutation varies run to run — so under a same-second tie roughly five of six runs come out wrong. `multiple_comments_ordered` writes three comments back to back, so on a fast runner it hits the tie nearly every time.

**Why I could not reproduce locally, and why that is consistent rather than contradictory:** four consecutive runs with `BEADS_TEST_PROXIED_SERVER=1` all passed here (214s for four). This box is slow enough that each `bd comment` CLI invocation lands in a *different* second, so `created_at` breaks the tie before `id` is ever consulted. It is fast runners that collide. Worth noting for anyone triaging: repeating the integration test on a slow machine will pass indefinitely and prove nothing, and a green CI re-run is luck rather than evidence the bug is gone.

A deterministic check that removes the timing variable: insert two comments with an identical explicit `created_at` second and assert their order — that exercises the actual invariant.

**If this is right, it is user-visible, not just a test issue:** any two comments posted within the same second display in digest order rather than the order they were written — easy to hit with scripted or agent-driven use.

I have deliberately not proposed a patch, because the fix interacts with this PR's convergence property and that is your design. The obvious lever — widening `created_at` precision — looks wrong here, since the commit message says the `DATETIME(0)` text is bound verbatim so the stored cell is byte-for-byte what was digested, meaning a precision change reshapes every derived id. Giving comments an explicit ordinal/sequence to order by seems closer to the intent, and this PR already says events "keep local multiplicity through ordinals" — so comments may simply have been missed. But that interacts with comments deliberately collapsing onto an identical existing row, so it is your call.

Happy to take the fix if you tell me which direction you want.

PR #5100, opened 2026-07-27
Pin authenticated historical release catalog

julianknutsen:
## What

Pin the exact historical release universe used by the v1.2 upgrade census:

- 122 stable module versions authenticated through the Go proxy and SumDB
- exact module, go.mod, source ZIP, and proxy-origin provenance
- explicit proxy prerelease and archived repository-only tag exclusions
- current repository-tag drift, including release assets whose tag no longer
  matches the proxy-reported module origin
- optional GitHub release and Linux/amd64 asset provenance

The checked catalog has a strict offline validator and a live deterministic
generator. Well-formed substitutions fail the pinned identity digest.

## Why

Upgrade coverage cannot be total until the historical input universe is finite
and reproducible. Raw repository tags are not authoritative: some historical
tags were removed, `v0.56.0` survives only in the module proxy, and 13 current
tag targets differ from proxy origin metadata. This catalog makes those facts
explicit without treating repository-only tags as supported releases.

## Scope

This PR is catalog-only. It adds no public command, storage behavior, schema
migration, or backend support. Runtime schema observations and family/path
qualification will follow in stacked PRs.

## Validation

- `go test ./scripts/migration-test/catalog`
- `GOPROXY=off go test -count=1 ./scripts/migration-test/catalog`
- `go vet ./scripts/migration-test/catalog`
- `golangci-lint run ./scripts/migration-test/catalog`
- live generation compared byte-for-byte with the checked catalog
- strict offline `validate`
- delegated Sol architecture, integrity, and test council: APPROVE
- `make test`

Based on merged #4907 and rebased onto current `main`.

PR #5156, opened 2026-07-29
Generate exhaustive runtime schema census

julianknutsen:
## What

Build the authenticated runtime schema census for every supported historical
Beads release and assign exactly one topology-derived recovery candidate to
every observed family.

The sealed census covers:

- all 122 SumDB-authenticated stable releases from `v0.9.1` through `v1.1.2`
- 169 fresh observations and 1,114 successful rolling transitions
- 410 unchanged transition refusals and 17 mutating transition failures
- 364 unique schema/layout families

Every family appears exactly once in the checked route manifest:

- 302 `sealed-copy-export-import`
- 42 `dual-root-export-union-import`
- 11 `multi-database-export-union-import`
- 8 `sealed-copy-direct-head-migration`
- 1 `head-init-from-jsonl`

The PR also adds deterministic two-lane Docker regeneration, semantic evidence
promotion, historical-process containment, per-execution binary rehashing,
rolling-frontier revalidation, and a cheap checked-artifact totality gate.

## Why

We cannot claim complete historical upgrade coverage until the release universe
has been executed, its actual runtime layouts have been fingerprinted, and every
distinct observed state has a defined route forward. Version ranges are not a
safe proxy for schema identity: fresh and rolling histories can produce
different layouts, and several releases produced coexisting or multiple
databases.

This census turns that unknown space into a finite, mechanically checked set.

## Qualification boundary

These are candidate routes derived from authenticated topology evidence. They
are deliberately recorded as `qualification: not-executed`; this PR does not
claim that the routes have passed end-to-end data-fidelity qualification.
`bd-9q3.3` is the next stacked slice that executes and qualifies one route per
family.

The migration CI matrix contains only releases in the reviewed E2E corpus.
`v0.50.2` remains in the exhaustive census, but is not scheduled as a qualified
historical-upgrade lane because the strict harness correctly rejects it as
unqualified; `v0.50.3` remains the reviewed configured-SQLite representative.

This PR adds no public command and changes no runtime migration behavior.

## Scope

- Local SQLite and Dolt layouts only
- No PostgreSQL or MySQL support
- No per-version migration adapters
- No private Dolt file introspection
- Stacked on #5100; the exact diff is 35 files and 18,069 additions, confined to
  `scripts/migration-test/census/` and two census CI workflows

## Rebase integrity

The three unique commits were replayed patch-identically onto repaired #5100
head `e095f026`:

- `6bdaa1f` -> `833e246`
- `ee5ec7d` -> `bd9b460`
- `9c6e882` -> `d04bb4e`

Stale #4907/#5100 ancestry was removed. Commit order, authors, author dates,
subjects, bodies, and trailers are preserved.

## Validation

- `make test` (39.7% total coverage)
- `go test -tags gms_pure_go -count=1 ./scripts/migration-test/census ./scripts/migration-test/catalog`
- checked-artifact `verify` against the authenticated catalog
- exact assertion that all 364 family IDs are unique and routed once
- `go vet`, scoped golangci-lint, gofmt, ShellCheck, actionlint, and `git diff --check`
- three delegated Sol council lanes: unanimous approval with no Critical or Important findings
- all 35 rebased file blobs are byte-identical to the previously reviewed head

The sealed artifacts remain byte-identical to the prior two-lane regeneration:

```text
runtime-schema-census.json.gz
e6c51ce97c818824fdfe05f43cb9c62ce3ff11093a5af3cd0cb93ac759d24a54

runtime-schema-routes.json
2741eb4f8f9303e17911d9c10440ee41c8cbab01332a42b667c89aeeedd05af0

runtime-schema-routes.md
309caa605fbeaf27770eb9453766224dffd90ec2d6fba3b20ebe35214e527a59
```


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98637689

maphew:
Reviewed. The engineering here is careful — containment of the untrusted historical binaries is thorough (non-root, `--read-only`, `--cap-drop ALL`, `no-new-privileges`, evidence promoted through a separate trusted container), and determinism is taken seriously end to end: pinned base-image digest, apt snapshot, pinned Go/Dolt/gcc/libc assertions, `deterministicGzip`, canonical JSON, sortedness assertions, and a two-independent-lane byte comparison.

Two things I specifically went looking for and found handled:

- **The `ignored_schema_migrations` track is covered.** This was my sharpest concern, since `AllMigrationsSQL()` deliberately omits the ignored source and anything built on it would silently under-report. The census doesn't go that route — it introspects the live runtime DB and captures any table matching `isMigrationLedger` (`sqlite.go:599`, suffix `_migrations`), which catches it. Confirmed empirically: 281 occurrences in the sealed census.
- **The census/catalog binding is fail-closed.** `artifacts.go:115-131` embeds the catalog SHA-256 and rejects a mismatch, and `validateCensus` requires every catalog version exactly once. So editing `release-catalog.json` turns the 20-second PR gate red and forces regeneration. That's the right sync mechanism — not the silently-rotting pattern I was worried about.

## One blocker, and it's two lines

`.github/workflows/migration-test.yml:42-43` adds:

```yaml
          - version: v0.50.2
            cache-identity: v0.50.2-361cf95eb640c...
```

`v0.50.2` is not in any qualified-version list in `scripts/migration-test/lib/versions.sh` and has no `STRICT_RELEASE_ASSETS` entry, so `historical-dolt-upgrade-test.sh:81` dies with `unqualified historical release: v0.50.2`. That's the one FAILURE in the rollup — `Historical v0.50.2 → candidate`; every other lane is green.

It also contradicts an existing assertion: `legacy-bridge-test.sh:171` lists `v0.50.2` among the versions the bridge **must reject** as unqualified.

The reason I'd like this fixed before merge rather than after: this workflow also runs on `push: tags: ['v*']`, so it would go red on every release tag, not just on PRs.

Deleting lines 42-43 is the fix. If the census discovering a distinct family at v0.50.2 is worth adding to the E2E lane, that's a separate change — `versions.sh` plus `STRICT_RELEASE_ASSETS`, and removing it from the bridge's reject list.

Happy to push that two-line deletion myself if you'd rather not spin the branch — just say so, since I don't want to touch your branch uninvited.

## Worth doing alongside it

- **Nothing in CI has reproduced the checked-in bytes.** The regeneration job is gated on `schedule || workflow_dispatch`, so it reports SKIPPED here; the artifacts come from your local Docker run. The mechanism is right (`run-container.sh:305-312` does `cmp` the regenerated output against what's checked in) — it just hasn't run. If it doesn't reproduce, the first evidence is a red scheduled job on `main` with a 6-hour turnaround. `workflow_dispatch` is already available, so one dispatch against this branch with the run URL in the PR body would close it.
- **Cache sizing.** Both `Save independent census-N cache lane` steps use a key suffixed with `${{ github.run_id }}-${{ github.run_attempt }}`, so every run writes a brand-new entry restored only by prefix match. Summing `release-catalog.json` — 89 pinned linux_amd64 assets ≈ 1.15 GB plus 122 source zips ≈ 0.64 GB — each lane trends toward ~1.5-1.8 GB, so ~3-3.6 GB written per weekly run against a 10 GB per-repo LRU. Two runs would evict essentially every other workflow's Go cache. (I summed the catalog rather than measuring the cache dir, so treat the number as approximate.) If the run-id suffix is deliberate immutability, worth a comment saying so plus bounding the cost another way — caching only the source zips, or dropping the asset cache since `acquireReleaseAsset` re-verifies digests anyway.
- **No prose docs for 17.7k lines.** `scripts/migration-test/census/` has no README, and `git grep` finds no mention of the census outside the two workflow files. `run-container.sh --init` — the only supported way to refresh the artifact after a catalog change — is documented nowhere. A short README covering what the sealed artifact is, `run-container.sh` vs `--init`, host requirements, and what to do when the weekly compare goes red would carry a lot of weight here.
- `workflow_dispatch` on `migration-test.yml` now also fires the 6-hour census, so dispatching to debug one historical lane launches the whole thing — worth a boolean input defaulting to false, or its own workflow file. And `schema-census-validation.yml:4` has no path filter, unlike `migration-test.yml`.
- `run_container_test.go:16-60` matches ~20 fragments of shell source text, one including exact backslash-continuation and indentation. These encode genuine isolation invariants nothing else enforces, so I wouldn't delete them — but normalize whitespace before matching, or a reformat fails the gate with a message about "isolation invariants" that doesn't point at the cause.

## Two maintainer calls, flagged not decided

Neither is a correctness issue, but both are standing costs someone should choose deliberately rather than inherit:

1. A 6-hour weekly Actions job plus ~3 GB/week of cache churn.
2. A 2.8 MB binary blob in git that must be replaced wholesale on every release pin — permanent history growth for an artifact that has no reader yet.

On (2): worth being clear that the schema fingerprints don't currently influence the deliverable. `routeForValidatedFamily` (`artifacts.go:224-350`) selects a route purely from `candidate.Mode` and `classifyTopology(...)`; the schema payloads are validated for consistency but never change the routing outcome. Since family IDs hash the full layout, the 364 families collapse to far fewer routing-relevant classes, which is why many rows in `runtime-schema-routes.md` are visually identical — and there's no shipped command to compute a family ID from a live `.beads/`, so a human can't locate their own family in the table yet. That's a defensible staging decision, but it means the expensive part of the artifact is a prerequisite for the next slice rather than something with a consumer today. If either call is "no", the cheap alternative is checking in only `runtime-schema-routes.json` + `.md` and keeping the 59 MB census as a CI artifact referenced by the `census_sha256` already recorded in the routes manifest.

Blocking only on the two-line workflow entry; everything else is follow-up.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98638484

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/bd-wisp-fsm

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds an exhaustive runtime schema census harness under `scripts/migration-test/census` — historical-version upgrade lineages, rolling SQLite/Dolt census generation, evidence promotion, and two supporting CI workflows — which gives the project a reproducible, versioned picture of how the runtime schema has evolved across releases and a guard against undetected schema drift.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Several census flows carried very high cyclomatic/cognitive complexity, notably the family route classifier and the rolling-Dolt lineage generation. The final patch extracts an ordered classifier for `routeForValidatedFamily` and splits `generateRollingDoltLineageWithRuntime` into focused helpers, so the hot paths are readable while first-match route precedence, golden route artifacts, deterministic frontier iteration, and executor cleanup are all preserved.
- The census Docker image installed an exact package before switching onto the pinned Ubuntu snapshot. The final patch moves the bootstrap onto the snapshot path so image builds are reproducible.
- A case-local error in the rolling-Dolt path could shadow layout-collection and marshal failures. The final patch routes those failures to the outer error that is checked before returning success.

Review findings addressed:
- Maintainability / complexity (required change, iterations 1–2): diff-introduced census routing and rolling-lineage functions carried very high cyclomatic/cognitive complexity. Resolved by extracting an ordered classifier and focused lineage helpers, plus a `routeManifestForFamilies` totality self-check to catch future classifier drift; the final pass confirmed route precedence, golden artifacts, frontier ordering, and executor cleanup are preserved.
- Release safety / image bootstrap (required change, iteration 2): the census Dockerfile installed an exact package before pinning to the Ubuntu snapshot. Resolved by moving the bootstrap onto the snapshot path.
- Correctness / error handling (validated in final pass): a case-local error in the rolling-Dolt path could mask real layout-collection and marshal failures. Resolved so those failures surface before success is returned.

Remaining non-gating notes:
- The census package does not currently cross-compile on Windows: the evidence-promotion path uses `golang.org/x/sys/unix` unconditionally while a `//go:build !linux` stub implies portability intent (minor, high confidence). This has no gate impact — the census CI lanes run on Ubuntu — and can be followed up by either marking the package Linux-only or splitting the TOCTOU-safe open into Unix/Windows files.

### Maintainer Changes
`fix(census): reduce census flow complexity; pin image bootstrap to snapshot` — 5 files changed, 606 insertions(+), 316 deletions(-) across `Dockerfile`, `artifacts.go`, `dolt_lineage.go`, `main.go`, and `rolling_dolt.go`. Original contributor commits are preserved underneath.

### Final Review Status
Ready for the merge queue: final head `9c6e882b3` has passing required GitHub checks — Migration Test Harness (all historical-version upgrade lanes) and Schema Census Validation — and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/31051568476

### Review Iterations
3 review passes performed: two fix iterations (block, then request_changes) that resolved the complexity, image-bootstrap, and error-shadowing findings, followed by a final pass that validated the fixes preserve behavior and approved.

---

julianknutsen:
## Rebase and validation update

PR #5156 has been mechanically rebased onto repaired #5100 head `e095f026`.
The exact new head is `d04bb4e0cf624c168c0157d686f19a61720e9ee7`.

- Three unique commits were preserved patch-identically: `6bdaa1f` ->
  `833e246`, `ee5ec7d` -> `bd9b460`, and `9c6e882` -> `d04bb4e`.
- The stale stack ancestry is gone. The PR diff is now exactly 35 files and
  18,069 additions, limited to `scripts/migration-test/census/` and two census
  CI workflows.
- All 35 file blobs match the previously reviewed head byte-for-byte; commit
  order, authorship, dates, messages, and trailers are preserved.
- `make test` passed at 39.7% total coverage. Focused census/catalog tests,
  offline artifact verification, vet, scoped lint, formatting, ShellCheck,
  actionlint, and `git diff --check` also passed.
- A three-lane delegated Sol council reviewed rebase correctness,
  security/supply-chain containment, and simplicity/test value. All three
  approved with no Critical or Important findings.

Exact-head GitHub CI is green:

- [Schema Census Validation](https://github.com/gastownhall/beads/actions/runs/31141454329)
- [Migration Test Harness](https://github.com/gastownhall/beads/actions/runs/31141454334): all 13 historical-upgrade lanes passed

The PR intentionally remains stacked on #5100. GitHub still reports #5100 open,
so #5156 must not be retargeted or merged until that base actually lands.

steveyegge:
## Review verdict: MERGE AFTER SEQUENCING — content is strong; the gates are all process

The engineering here is unusually rigorous and I want that said plainly: digest-pinned base image with an apt *snapshot* repo and sha256-pinned toolchains, two independent cache lanes byte-compared before sealing, per-execution binary rehashing, and a real PR-time totality gate (`TestCheckedArtifactsTotality` against the live release catalog) that makes the census impossible to stale silently — the census only rots if the catalog rots, which is #5100's process boundary, not this PR's hole. I also confirm the census and `check-migration-hygiene.sh` meet nowhere: source-time guard vs runtime observation, complement not duplication. The nondeterminism allowlist is orthogonal by construction (layout fingerprints, not data).

Four things gate the merge, all sequencing rather than code:
1. **This PR targets #5100's head branch, not main** — merging it ships nothing and must be sequenced behind (or into) #5100; if #5100 squash-merges first, this base evaporates (the known squash-window hazard). State the intended landing order explicitly on both PRs.
2. **The repo's core gate has never run on this diff** — checks here are the census lanes only; lint/vet/build-tag/Windows/pure-Go run only on main-targeted PRs. Final verification is wherever #5100's PR re-runs with these commits absorbed; let's name that as the acceptance point.
3. **The 6-hour `exhaustive-census` job has never executed in Actions** — it's schedule/dispatch-only and its first real run would otherwise be an unwatched post-merge Monday cron. One `workflow_dispatch` proof-run on the branch before merge, please — the script logic is unit-tested, but cache lanes, the 2×155m timeout budget, and artifact uploads are Actions wiring only a run can prove.
4. **Two maintainer decisions to record:** whether `Validate schema census` joins the required-checks set (it runs on every PR with no path filter — ~2min repo-wide tax, which I think is the correct price for the totality gate, but it should be a decision, not a default), and an explicit ack of the 2.8MB binary artifact entering history with wholesale rewrites per regeneration (I think the hermetic PR-time gate justifies it; the .json/.md companions are the reviewable surface).

MINOR for whenever you're in there: branch pins go-sqlite3 1.14.48, main is at 1.14.49 — will converge on rebase.

maphew (approved):
Re-review at the current head. My sole blocker from the 2026-07-30 review - the unqualified `v0.50.2` entry in the `migration-test.yml` matrix - is fixed: it is gone, and all 14 historical lanes plus the census validation check pass at this head.

Beyond the blocker I re-verified the load-bearing pieces: the fail-closed catalog binding (`pinnedCatalogSHA256` and the digest check in `readValidCensusBytes`), the PR-time totality gate (`TestCheckedArtifactsTotality` ran green in CI, so the checked-in sealed artifacts provably match the generator and catalog), the hardened two-lane regeneration with byte comparison, the defensive evidence promotion (`O_NOFOLLOW`, identity re-verification, size bounds), and the subreaper-based process containment - the `/proc/stat` parsing checks out field by field.

Approving at the code level. The follow-ups I flagged on 2026-07-30 remain open and remain non-gating from my side, listed so they don't get lost:

- The 6-hour exhaustive-census job has still never executed in Actions (no `workflow_dispatch` run exists on the head branch); one proof run with the URL recorded here would de-risk the first scheduled execution.
- `workflow_dispatch` still fires both jobs at once - a boolean input defaulting the census off would make single-lane debugging cheap.
- No README for the 18k-line census (`--init` vs compare, host requirements, what to do when the weekly compare goes red).
- Cache keys suffixed with run id write fresh multi-GB entries into the 10 GB Actions LRU each week; if immutability is deliberate, a comment plus bounding the cost would do.
- The isolation-invariant tests match exact shell text including whitespace, so a pure reformat fails with a misleading message.

Sequencing remains the real gate and it is a maintainer call, not mine: the PR targets #5100's head branch rather than main, and steveyegge's 2026-08-08 gates (stated landing order on both PRs, core-gate verification at the #5100 acceptance point, the census proof run, and the two recorded decisions on required-check status and the 2.8 MB regen-rewritten artifact) still stand. Note I stepped down as maintainer on 2026-08-10, so this approval is contributor input into that sequencing decision, not a lever on it.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277174594

PR #5159, opened 2026-07-29
feat: add --team-server mode for proxied server

eric-richardson1:
Proxied-server mode migrates the schema on every open and rewrites                                                                                                                        issue_prefix/_project_id on every init, which breaks shared databases managed by                                                                                                          beads-team-server. This adds `bd init --proxied-server --team-server`, persisted in                                                                                                        metadata.json so clones inherit it. In this mode bd never creates the database or                                                                                                           runs migrations — it just checks the schema version and tells you to run `bts init`                                                                                                         or `bts migrate` (or upgrade bd) if there's a mismatch. Init also adopts the                                                                                                               provisioned issue_prefix and _project_id from the DB instead of writing them, so a                                                                                                          second teammate's init can't clobber the project identity. Requires `--database`                                                                                                            since bd can't guess the provisioned database name.


internal/storage/uow/team_server_schema.go
```diff
@@ -0,0 +1,40 @@
...
+		if os.Getenv("BD_IGNORE_SCHEMA_SKEW") == "1" {
+			fmt.Fprintf(os.Stderr,
+				"Warning: schema skew ignored — database %q is at schema v%d, this bd expects v%d; queries touching newer schema may fail\n",
+				database, current, latest)
+			return nil
+		}
+		// Not SchemaBehindError: its "run any bd write command to migrate"
+		// advice is wrong for a bts-owned schema.
+		return fmt.Errorf(
+			"uow: database %q is at schema v%d, this bd expects v%d; the schema is managed by beads-team-server — ask your operator to run 'bts migrate', or use a bd built against schema v%d (set BD_IGNORE_SCHEMA_SKEW=1 to proceed anyway)",
```
coffeegoddd: yikes! i would not add BD_IGNORE_SCHEMA_SKEW, we can add that later if needed

coffeegoddd (approved):
One comment, re the escape hatch, and also if the go.mod changes aren't needed, i'd revert those, they look unrelated to this PR. Otherwise, LGTM!

maphew:
Reviewed for gate-bypass risk specifically, since a new Dolt-server connect path is the highest-risk shape of change in this area. **The core safety claim holds** — I traced all three gates:

- **Migrate gate**: `schema.MigrateUpWithLock` is reachable on the proxied path from exactly one call site, `internal/storage/uow/dolt_sql_provider.go:126`. The team-server branch short-circuits before it and before any `CREATE DATABASE`, and returns `backoff.Permanent` for non-serialization errors so it doesn't spin. That's a strict tightening, not a bypass.
- **Forward-drift guard**: `checkTeamServerSchema` delegates to the same `schema.CheckForwardDrift` that `dolt/store.go` uses. Not bypassed.
- **Remote-migrate gate**: never ran on the proxied path (`CheckRemoteMigrateGate*` callers are `dolt/store.go` and `embeddeddolt/store.go` only), and this removes migration from that path entirely, so it can't regress.

Fails closed on unreachable, missing, empty, behind, and ahead databases, each with a test. I tried to construct a silent-success path and couldn't. Reusing `resolveInitIssuePrefix`/`resolveInitProjectID` with `gateway=true` is the right call.

One item I'd like addressed before merge, and one structural gap I would not block on.

## 1. Init still writes clone-local state and creates a Dolt remote in the bts-owned shared database

`SkipIdentity` gates **only** `issue_prefix` and `_project_id`:

```go
if !params.SkipIdentity {
    ... SetConfig("issue_prefix", ...)
    ... SetMetadata("_project_id", ...)
}
if params.RepoID != "" {   // <-- still unconditional
    ... SetMetadata("repo_id", ...)
```

So every teammate's `bd init --team-server` still writes `repo_id`, `clone_id` and `last_import_time` into the operator's shared `metadata` table, and if a git origin resolves, `BootstrapProject` calls `CreateRemote(ctx, "origin", <that teammate's origin URL>)` — adding a Dolt remote pointing at an individual contributor's git origin to the shared database.

The PR's stated goal is that "a second teammate's init can't clobber the project identity", and the analogous case is already solved ~800 lines away in `cmd/bd/init.go:1307-1313`:

> Tracking metadata (bd_version, repo_id, clone_id) is clone-local state. In gateway mode it must not be written into the shared, server-owned database: repo_id/clone_id are per-clone fingerprints and a last-init-wins overwrite feeds false cross-project mismatch diagnostics...

`doctor` and cross-project verification read `repo_id`/`clone_id` back, so this reproduces exactly the false-mismatch diagnostics that comment describes. Cheapest fix: an additional `SkipInitState bool`, or reuse the existing `shouldWriteInitStateToDB` predicate — team-server and gateway are asserting the same policy ("the server owns this database"), so they should share the predicate rather than each implement half of it.

## 2. No runtime project-identity verification on the proxied path (pre-existing, but materially widened)

To be clear about blame: **this PR does not remove a gate.** `cmd/bd/main.go:1348-1358` already short-circuits proxied mode before `validateWorkspaceIdentity`, and proxied mode never constructs a `DoltStore`, so `verifyProjectIdentity` is already unreachable there on current `main`.

What changes is exposure. Until now, proxied-server meant a per-workspace database bd itself created, so "the server is serving a different project's database" barely applied. `--team-server` deliberately points bd at a long-lived, shared, operator-managed database, selectable per-invocation via `--database`. Init *adopts* whatever identity it finds — it can't detect a wrong target — and nothing re-checks afterward. Gateway mode, the closest analogue, does get `verifyProjectIdentity` on every non-`CreateIfMissing` open.

The reader already exists (`adoptTeamServerIdentity`'s `teamServerIdentityReader`), so comparing `configfile.Load(beadsDir).ProjectID` against the DB's `_project_id` after `checkTeamServerSchema` is a small addition rather than new machinery. I'm fine with "merge and track it" — I've filed it on our side — but I don't think a shared team database should stay with no runtime identity check indefinitely.

## Smaller notes

- **`BD_IGNORE_SCHEMA_SKEW` on a writable connection.** The precedent it mirrors, `CheckBehindDrift`, is documented as "used by read-only opens, which skip MigrateUp by design" — its hatch downgrades a *read* risk. Here it lets a newer `bd` **write** against an older bts schema, naming columns the DB may not have. Note `--ignore-schema-skew` sets the same env var, so someone who set it for an unrelated read silently gains that. Consider honoring it only for read-only commands in team-server mode. This looks like it overlaps @coffeegoddd's escape-hatch comment — worth folding into that thread rather than a separate change.
- **`IsTeamServerManaged()` fails open**: `dolt_team_server: true` with a non-proxied `dolt_mode` evaluates to `false`, and false means "resume migrating the operator's database". Only reachable by hand-editing metadata.json or a re-init that rewrites `dolt_mode`, so low likelihood — but silently-inert is the wrong default for a safety bit, and there's a test pinning that behavior, so it reads as deliberate. Worth challenging.
- **`--database` override** isn't refused in a team-server workspace. The bts-DB-in-a-non-team-server-workspace direction is the dangerous one and predates this, but `--team-server` is what makes "there is an operator-owned DB on this server" true.
- **`bts` is unfindable**: error strings and flag help tell operators to run `bts init` / `bts migrate`, but `beads-team-server`/`bts` appears nowhere else in the repo and there's no doc or URL. `[EXPERIMENTAL]` mitigates it; one link in the flag help would close it.
- **Unrelated dep churn** in `examples/*/go.mod|go.sum` (go-isatty, x/sys, x/sync, x/term) — already flagged by @coffeegoddd; worth reverting so the diff reads as one thing.
- CI: `Test (Proxied Dolt Cmd 14/15)` failed on `TestProxiedServerComment/multiple_comments_ordered` (comment timestamp-tie ordering) — unrelated to this change and it looks like a flake, but it should be rerun rather than waved through.

Verification coverage is good overall (7 `adoptTeamServerIdentity` cases, 7 `checkTeamServerSchema` cases including both hatch branches, 4 integration scenarios). The one uncovered block is the `configfile.Load` → set `ProjectID` → `Save` sequence in `init_proxied_server.go`: it's non-atomic, and a failure there aborts init having already mutated the shared DB per item 1, leaving metadata.json disagreeing with the database — silently, given item 2.

Net: **approve with comments**, with item 1 as the pre-merge ask.

PR #5160, opened 2026-07-30
fix(schema): renumber duplicate migrations

DustinHolden:
Main currently panics before any command can run because migration versions collide in both embedded sets:

- main: `0044_update_views_drop_depends_on_id` vs `0044_widen_event_value_columns`
- main: `0046_add_is_blocked` vs `0046_longtext_large_content_columns`
- ignored/wisp: `0005_drop_wisp_dependencies_generated_column` vs `0005_widen_wisp_event_value_columns`

This keeps each earlier migration at its assigned number and moves the later LONGTEXT migrations after the existing maximum: `0048`/`0049` for main and `0008` for ignored. It also fixes stale migration-number comments.

Verified:

```text
$ go test ./internal/storage/schema/...
ok github.com/steveyegge/beads/internal/storage/schema
$ CGO_ENABLED=0 go build -o /tmp/bd ./cmd/bd
$ /tmp/bd version
bd version 1.0.4 (dev: ff2f6154ab1c)
```

A per-directory duplicate-version scan is empty after the renames.


maphew:
Thanks for this — the diagnosis is correct and independently reached, including the non-obvious `ignored/0005` collision that's easy to miss. Unfortunately the bug was fixed on `main` about two months ago, and I don't think there's a resolution of this PR that we can land.

**Where the fix already is:** `34936e260` "fix(schema): renumber duplicate release migrations", with a comment follow-up in #4235. Current `main` has no duplicate versions:

- `0048_widen_event_value_columns`
- `0049_longtext_large_content_columns`
- `ignored/0008_widen_wisp_event_value_columns`

Your branch's parent is `ff2f6154ab1`, dated about five minutes before that commit — so the panic you describe is real at your base, just not on current `main`.

**Why we can't merge it anyway:** the numbering here is the inverse of what shipped. This PR assigns `0048`=longtext / `0049`=widen; `main` assigns `0048`=widen / `0049`=longtext. `git merge-tree --write-tree main <pr>` exits 1 with four rename/rename conflicts, and a naive "accept both" resolution lands *both* files at *both* numbers — which re-triggers `checkNoDuplicateVersions` (`internal/storage/schema/schema.go:988`, called from `list()` at `:1017`) and panics on every command. That is, merging would reintroduce exactly the failure this PR fixes.

Heads-up for anyone reading the PR page: GitHub currently reports this as `MERGEABLE`. That's stale — the local `merge-tree` above is authoritative.

Worth stating explicitly, since it's the real constraint: renumbering *shipped* migrations is only safe once, before release, and whichever assignment shipped first has to win. The migration cursor is a high-water mark (`schema.go:1140-1148`), so an already-migrated database wouldn't re-run or skip anything on a swap. But the recorded hash is `sha256` of file *contents* keyed by *version* (`schema.go:1232-1234`), so after a swap a freshly-migrated clone records version 48 with longtext's hash while existing clones have widen's hash — and `ContentHashSkew` (`internal/storage/schema/content_skew.go:16`) would then flag versions 48 and 49 as a "silent schema fork" across the entire installed base, for clones that are actually schema-identical.

**The part of this worth keeping.** There's a latent trap your PR would have tripped that nothing currently guards: `internal/storage/schema/cli_migrations.go:39` dispatches on the *filename string* `"0049_longtext_large_content_columns.up.sql"` with a `default: return sqlText` fallthrough. Renaming that file makes the case unreachable with no compile error and no failing test — the fresh-schema CLI bundle would silently fall back to raw DDL. I've filed a follow-up to pin shipped migration version→filename in a test, which would catch both that dispatch and any future accidental renumbering, and credited you on it.

So: closing as already-fixed rather than as wrong. The analysis was sound; it just raced a fix that was five minutes ahead of your branch point. Sorry for the wasted effort, and thanks for digging into the actual gate rather than just reporting the panic.

PR #5162, opened 2026-07-30
feat(storage): flip the events audit table to the dolt_ignored/unversioned plane (bd-red8u)

steveyegge:
# feat(storage): flip the events audit table to the dolt_ignored/unversioned plane (bd-red8u)

M1 of the 2026-07-30 mitigation-first ruling (Steve directive via ant; plan of
record: wyvern `brain/dolt-mitigation-plan-2026-07.md`). **Events only** —
comments stay versioned until the C2.3 newest-wins replication leg (bd-1quyq)
exists.

Events is the dominant history churner: every issue op appends audit rows, so
the table rides ~80% of dolt commits and is the primary driver of unbounded
reachable-history growth (bd-7vb13 measurements). After this flip the rows keep
full SQL durability in the working set but stop minting versioned history —
same plane as leases (0055) and repo_mtimes (0028).

## What's different from the leases/wisps precedents

Events is an **already-committed synced table**, and dolt_ignore only exempts
tables that have never been committed. A drop+recreate netting out in one
working set leaves the table tracked-at-HEAD and permanently dirty (verified
empirically on dolt 2.2.2). The drop must therefore be **committed on its own
before the recreate**, making 0062 the first self-committing table-plane
migration (0040/0041 are the self-commit precedent).

## Changes

- **Migration 0062** (`0062_events_dolt_ignore.up.sql`): four-phase,
  replay-safe from any intermediate state — heal-create → copy rows aside →
  drop + conditionally `DOLT_ADD('-f','events')` + `DOLT_COMMIT('--skip-empty')`
  → seed `dolt_ignore` → recreate → restore rows. The add is conditional on
  dolt_status actually showing a pending events change (a replay's drop of the
  never-committed recreation leaves no status row, and adding a table that
  exists nowhere errors). The `-f` matters: plain `DOLT_ADD` of an ignored
  table is a silent no-op, so a database that already carries the pattern while
  events is still tracked at HEAD (branch-per-test databases materialize
  exactly that state) would never commit the drop. On the normal path the
  pattern doesn't exist yet, so `-f` is identical to a plain add.
- **Ignored migration 0019** (`0019_create_events.up.sql`): fresh-clone
  creator on the ignored lane (bd-rotq2 rule). PREPAREd conditional CREATE
  because `fk_events_issue` is database-unique — the usual `__temp__`+RENAME
  idempotent-create recipe collides on the constraint name.
- **Version-gated dolt_ignore seeding** (`schema.go`): `events` seeds into
  `doltIgnorePatterns` only when the main cursor is ≥ 62. Unconditional
  seeding would strand a gate-refused pre-flip DB (tracked table + suppressed
  staging = writes never commit).
- **Doctor**: `events` added to the ignored-table lists in
  `cmd/bd/doctor/dolt.go`.
- **Tests**: two embedded contract tests (flip end-to-end + replay;
  fresh-clone door + FK), retargeted events-at-HEAD assertions across
  `create_issue_commit_test`, `transaction_test`, `dependency_events_test`,
  `git_remote_test`, `cmd/bd/create_embedded_test`, and the aux-rekey fixture
  now force-stages its events seed (a real pre-rekey lineage predates the flip,
  so its events rows were committed; `-Am` now skips the ignored table).
  Shared branch-per-test databases materialize an empty events shell at HEAD
  (`MaterializeLocalTableSchemasForBranchTests`), so the dolt-package
  assertions check "no events ROW at HEAD" — the invariant that holds on both
  shapes — while the embedded contract tests pin the stronger
  production-shape contract (table absent at HEAD, absent from dolt_status).
- **domain/db suite**: `DOLT_RESET('--hard')` swaps the tracked issues
  table's backing object, which silently severs `fk_events_issue`
  enforcement on the untracked events table (verified on dolt-sql-server
  2.2.0 with a minimal probe; labels' FK on the same root keeps enforcing).
  The suite recreates events from captured DDL after each per-test reset,
  re-linking the FK. Production hard-reset sites (flatten/compact squash
  path, mergesettle abort, schema-lock error path) inherit the same
  behavior — consequence is dangling audit rows, the already-accepted
  "harmless audit noise" class; tracked with a dolt-upstream report
  candidate in **bd-7bpkd**.

## Verified dolt semantics the design leans on (scratch DBs, dolt 2.2.2)

- Explicit `DOLT_ADD('events')` and mixed multi-table adds **silently skip**
  ignored tables → bd's runtime staging lists keep working unmodified, and
  **old binaries stay compatible with flipped DBs in place**.
- `DOLT_ADD('-f')` **does** pierce dolt_ignore. The only production force-add
  sites (`stageSchemaTables`, `commitMigrationStep`) draw from ignore-filtered
  sources, so neither can re-track events.
- Drop+recreate in one working set nets to a perma-dirty tracked table — the
  reason the drop is committed before the recreate.

## Review notes for lion

- **Rekey ordering** (your earlier note: rekey catch-up needs versioned
  replication): lane order covers lineages that took the 0061
  content-derived-ids anchor in an earlier pass (the common case — 0061 is on
  deployed main): their rekey work is committed on the versioned plane before
  the flip. For a laggard DB jumping ≤0060→0062 in a **single** pass the
  ordering is the other way around — `mainSource.migrate` applies 0061+0062
  first and `rekeyAuxRowIDsAllPasses` runs in MigrateUp's tail — so that
  lineage rekeys its events rows on the already-unversioned plane. Harmless
  for a different reason: post-flip, events never merges across clones, so
  the cross-clone ID-convergence property the rekey exists for is moot for
  this table (comments/snapshots stay versioned and rekey normally).
- **Rollout easing for wy-7cnbj**: in-place old binaries tolerate flipped DBs
  (see semantics above) — "in place" covers only DBs that flip locally and do
  not merge the flip in from elsewhere. Binary-first ordering is needed for
  **any machine that will fresh-clone OR `bd sync`/pull/merge from a
  post-flip origin**. The merge path delivers the same failure as the clone
  path: 0062 commits the events DROP on the versioned plane, the working-set
  recreate is ignored-plane and never replicates, so an old-binary clone that
  merges the flip either loses its events table outright (bd-rotq2-class
  insert failures, if it had no local events changes since the merge base) or
  hits a drop-vs-modified table conflict (if it did — the normal case at any
  write tempo). Confirmed against 0062 Phase 2 semantics per lion's review.
- **Pre-flip events history stays in remotes** until the next squash window —
  accepted; only new growth stops.
- **Orphaned working-set events**: merge-applied issue deletions do not
  cascade into another clone's working-set events rows. Harmless audit noise;
  sweep candidate for the bd-1quyq era.
- Consumers unchanged: `bd history` / EventsSince / counts read the single
  query plane; merge-code events handling stays (needed for pre-flip DBs
  mid-rollout); events were already absent from jsonl export;
  `commit_pending` batch messages drop events automatically.

## Test results

- `internal/storage/schema`: green (incl. ThroughDoltCLI full-bundle).
- `internal/storage/embeddeddolt` (BEADS_TEST_EMBEDDED_DOLT=1, full suite):
  green, 1131s.
- `internal/storage/dolt` (container path): all 730 tests run in six chunked
  invocations — **zero branch-attributable failures**. Includes
  `TestAuxRowIDRekeyConvergesIndependentClones`, which caught the
  ignored-pattern-pre-exists wedge that motivated the `-f`.
- `cmd/bd` (env -u BEADS_ACTOR): all 2078 tests run in eight chunked
  invocations — **zero branch-attributable failures**.
- `internal/storage/domain/db` (`TestDomainDB`, Docker): green after the
  suite's post-reset events recreation (the one genuine branch interaction
  found outside the dolt package — see above).
- Pre-existing failures on this machine, all verified to fail **identically
  on origin/main** in a clean worktree (none related to this branch):
  doctor role tests (global `git config beads.role` leak) + test-pollution
  EmptyDB; the fresh-DB "table not found: schema_migrations after New()"
  family (`TestDoltNew_*` gates/drift, `TestCheckForwardDrift_EscapeHatch`,
  `TestSchemaRunsInitWhenMissing`); `TestIgnoredMigration0011…` (stale
  delete-latest-cursor mechanism, filed as a bead); an is_blocked/recompute/
  rig-columns cluster of 8; a cmd/bd cluster (init `--contributor` vs
  `--non-interactive` conflict and friends, ~20 tests). A handful of others
  are chunk-order/contention flakes that pass in isolation (heartbeat trio,
  role-prompt trio, `TestDoltStoreComments`, `TestReclaimRevertsExpiredOnly`).

### DIFF
--- a/cmd/bd/create_embedded_test.go
+++ b/cmd/bd/create_embedded_test.go
@@ -1003,9 +1003,11 @@ A new feature
 			t.Fatalf("label count = %d, want 2", labelCount)
 		}
 
+		// events is dolt_ignored since 0062 (bd-red8u): audit rows are durable
+		// in the working set, never at HEAD.
 		var labelEventCount int
 		if err := db.QueryRowContext(t.Context(),
-			"SELECT COUNT(*) FROM events AS OF 'HEAD' WHERE issue_id = ? AND event_type = ?",
+			"SELECT COUNT(*) FROM events WHERE issue_id = ? AND event_type = ?",
 			id, types.EventLabelAdded).Scan(&labelEventCount); err != nil {
 			t.Fatalf("count label events: %v", err)
 		}
--- a/cmd/bd/doctor/dolt.go
+++ b/cmd/bd/doctor/dolt.go
@@ -289,7 +289,7 @@ func checkSchemaWithDB(conn *doltConn) DoctorCheck {
 	// Check dolt_ignore'd tables — these only exist in the working set and
 	// must be recreated each server session. (GH#2271)
 	ignoredTables := []string{
-		"leases", "local_metadata", "repo_mtimes",
+		"events", "leases", "local_metadata", "repo_mtimes",
 		"wisps", "wisp_labels", "wisp_dependencies", "wisp_events", "wisp_comments",
 	}
 	var missingIgnoredTables []string
@@ -412,7 +412,7 @@ func CheckDoltIssueCount(path string) DoctorCheck {
 // produces self-fulfilling warnings that can never be cleared.
 func isIgnoredTable(tableName string) bool {
 	switch tableName {
-	case "wisps", "leases", "local_metadata", "repo_mtimes":
+	case "wisps", "leases", "local_metadata", "repo_mtimes", "events":
 		return true
 	}
 	return strings.HasPrefix(tableName, "wisp_")
--- a/internal/storage/dolt/aux_row_id_rekey_test.go
+++ b/internal/storage/dolt/aux_row_id_rekey_test.go
@@ -65,6 +65,14 @@ func seedAuxRekeyFixture(ctx context.Context, t *testing.T, db *sql.DB, clonePre
 		t.Fatalf("regress main cursor: %v", err)
 	}
 
+	// A real pre-rekey lineage predates the 0062 events flip, so its events
+	// rows are committed. The shared test database carries the post-flip
+	// dolt_ignore pattern (with events deliberately materialized at HEAD for
+	// branch inheritance), so '-Am' would silently skip events and leave the
+	// seed dirty — stage it explicitly with '-f' to match the simulated state.
+	if _, err := db.ExecContext(ctx, "CALL DOLT_ADD('-f', 'events')"); err != nil {
+		t.Fatalf("stage events seed: %v", err)
+	}
 	if _, err := db.ExecContext(ctx, "CALL DOLT_COMMIT('-Am', 'seed pre-rekey rows')"); err != nil {
 		t.Fatalf("commit seed: %v", err)
 	}
--- a/internal/storage/dolt/create_issue_commit_test.go
+++ b/internal/storage/dolt/create_issue_commit_test.go
@@ -59,15 +59,23 @@ func TestCreateIssueCommitsInitialRelationalData(t *testing.T) {
 		t.Fatalf("committed comment count = %d, want 1", commentCount)
 	}
 
+	// events is dolt_ignored since migration 0062 (bd-red8u): the audit rows
+	// must be durable in the working set but never part of committed history.
 	var labelEventCount int
 	if err := store.db.QueryRowContext(ctx,
-		"SELECT COUNT(*) FROM events AS OF 'HEAD' WHERE issue_id = ? AND event_type = ?",
+		"SELECT COUNT(*) FROM events WHERE issue_id = ? AND event_type = ?",
 		issue.ID, types.EventLabelAdded,
 	).Scan(&labelEventCount); err != nil {
-		t.Fatalf("count committed label events: %v", err)
+		t.Fatalf("count label events: %v", err)
 	}
 	if labelEventCount != 2 {
-		t.Fatalf("committed label_added event count = %d, want 2", labelEventCount)
+		t.Fatalf("label_added event count = %d, want 2", labelEventCount)
+	}
+	var committedEventCount int
+	if err := store.db.QueryRowContext(ctx,
+		"SELECT COUNT(*) FROM events AS OF 'HEAD'",
+	).Scan(&committedEventCount); err == nil {
+		t.Fatalf("events exists at HEAD with %d rows; want the table absent from committed history (dolt_ignored, 0062)", committedEventCount)
 	}
 
 	var dirtyRelationalTables int
--- a/internal/storage/dolt/create_issue_commit_test.go
+++ b/internal/storage/dolt/create_issue_commit_test.go
@@ -74,14 +74,17 @@ func TestCreateIssueCommitsInitialRelationalData(t *testing.T) {
 	var committedEventCount int
 	if err := store.db.QueryRowContext(ctx,
 		"SELECT COUNT(*) FROM events AS OF 'HEAD'",
-	).Scan(&committedEventCount); err == nil {
-		t.Fatalf("events exists at HEAD with %d rows; want the table absent from committed history (dolt_ignored, 0062)", committedEventCount)
+	).Scan(&committedEventCount); err == nil && committedEventCount != 0 {
+		t.Fatalf("events has %d rows at HEAD; want none in committed history (dolt_ignored, 0062)", committedEventCount)
 	}
 
+	// events is excluded: it is permanently working-set-resident since 0062,
+	// and on the shared branch-per-test database (events materialized at HEAD)
+	// its audit rows legitimately show in dolt_status as a modification.
 	var dirtyRelationalTables int
 	if err := store.db.QueryRowContext(ctx, `
 		SELECT COUNT(*) FROM dolt_status
-		WHERE table_name IN ('labels', 'comments', 'events')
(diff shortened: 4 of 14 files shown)


steveyegge:
## Review verdict: APPROVE — merge when ready

Reviewed per wy-56xg2 (requested crew verdict; lion). I re-derived the design claims from the branch rather than trusting the summary: read 0062/ignored-0019 in full, the schema.go version-gated seeding, the runner path (`MigrateUpWithLock` pinned `*sql.Conn` → `execMigrationBody` → `DrainCall`, so the `SET @var`/`PREPARE` dance and the mid-body `COMMIT` have guaranteed session affinity), ran `scripts/check-migration-hygiene.sh` on the branch (clean), enumerated the production `DOLT_RESET('--hard')` sites independently (flatten.go:61, compact.go:75, mergesettle.go:340, lock.go:107 — matches the PR body's list exactly), and read the embedded contract tests + the suite/fixture retargets. The four-phase replay-convergence design is sound, the conditional `DOLT_ADD('-f')` reasoning holds (and the rekey fixture is the state that proves it's needed), and the version gate on `doltIgnorePatterns` correctly avoids stranding a gate-refused pre-flip DB. The fresh-clone door (ignored 0019 + gated re-seed) is belt-and-suspenders even if the flip pass's dolt_ignore row somehow failed to commit.

Three non-blocking findings; the first two want PR-body/rollout-note corrections before or at merge:

### 1. The rekey-ordering claim is imprecise — please correct the body text (no code change)

The body says my earlier note is "satisfied by lane order — any rekey work is committed on the versioned plane before the flip detaches events." That holds only for lineages that took 0061 in an **earlier** pass (the common case — 0061 is already on deployed main). For a laggard DB jumping ≤0060→0062 in a **single** pass it's wrong: `mainSource.migrate` applies 0061 *and* 0062 first, and `rekeyAuxRowIDsAllPasses` runs afterward in MigrateUp's tail (schema.go ~L539–570) — so that lineage rekeys its events rows on the already-unversioned plane, uncommitted and unreplicated.

I believe this is **harmless**, for a different reason than the one stated: post-flip, events never merges across clones, so the cross-clone ID-convergence property the rekey exists for is simply moot for that table (comments/snapshots stay versioned and rekey normally). But the justification as written would mislead the next person who reasons from it — please restate it as "lane order covers already-0061 lineages; for single-pass laggards the flip itself moots events replication" or equivalent. If you know a reason the single-pass case is *not* harmless, that's a blocker and I'd like to hear it.

### 2. Rollout note understates: there is a third machine class — old binary + sync/pull from a flipped origin

"Binary-first ordering is needed only for machines that will fresh-clone a post-flip origin" misses the merge path: an old-binary clone that `bd sync`s (dolt pull/merge) from a flipped origin receives the committed **drop** of events, has no ignored-0019 to recreate it, and the origin's working-set recreate doesn't replicate — so it lands in the same bd-rotq2-class state as a fresh clone, via merge instead of clone. For the wyvern wy-7cnbj rollout this is mostly moot (one shared sql-server, no cross-machine dolt merges in the hot path), but as a general beads compatibility claim it should read: **binary-first for any machine that will fresh-clone OR pull/merge from a flipped origin**; "in place" tolerance covers only DBs that flip locally and don't merge the flip in from elsewhere. Please confirm the merge path can indeed deliver the drop to an old-binary clone (I reasoned from three-way merge semantics, didn't repro) and widen the note in the body + wy-7cnbj accordingly.

### 3. bd-7bpkd suggestion: the severed-FK state is *persistent* in prod, not per-incident

After any production hard reset (every flatten/compact squash window, in particular), `fk_events_issue` stays severed **indefinitely** — nothing recreates the events table until something happens to rebuild it, so orphan accumulation is continuous from the first squash onward, not a one-shot per reset. Still inside the accepted audit-noise class, but worth capturing on bd-7bpkd: a `doctor --fix` heal (recreate events from captured DDL and re-link the FK, exactly what the domain/db suite now does per-test) would bound it cheaply, and the upstream report should mention the persistence.

Minor observations, no action needed: the hygiene allowlist correctly needs no 0062/0019 entries (checker is green on the branch); the weakened "no events ROW at HEAD" assertions in the dolt package are properly backstopped by the embedded contract tests pinning the strong production shape; the conditional-CREATE recipe in 0019 (constraint-name collision vs the usual `__temp__`+RENAME) is well justified.

Constellation stake acknowledged: this shape reaches our shared dolt server via wy-7cnbj — findings 1 and 2 are exactly the notes that rollout should inherit.

🤖 Posted by lion (fable crew) for wy-56xg2

steveyegge:
CI gate note for the record: the Proxied Dolt Cmd 14/15 failure is not branch-attributable — it is the main-side same-second comment-ordering nondeterminism introduced by bd-ri8bd (423afdcb2): second-resolution created_at + digest ids make the (created_at, id) order an arbitrary permutation, ~5/6 failure odds per run (two consecutive runs here produced two different wrong orderings; chunk 10/15's first-run failure was runner infra, exit 127 in setup, and passed on rerun). Tracked in bd-vuulx; interim test de-flake is PR #5166. Merge plan: land #5166, update this branch from main, rerun the gate, merge on green per lion's posted verdict.

PR #5163, opened 2026-07-30
fix(dolt): do not silently retarget an auto-started Dolt server onto a different port/database (#4052)

maphew:
Fixes the silent-wrong-database write in #4052, "Shared-server mode: write commands report success when Dolt is unreachable; status change never lands."

## Why the obvious fix is wrong

When the configured Dolt server is unreachable, `newServerMode` auto-starts a new server, which lands on an **ephemeral port**. Today the mismatch is announced on stderr only, `cfg.ServerPort` is overwritten, and the command proceeds — so the write goes to whatever database that new server serves, with exit 0 and a success message.

Simply refusing to retarget would be a bad regression: in the common single-repo case the recorded port is just a stale ephemeral leftover, and silently picking a fresh one is exactly what users want. So the fix has to distinguish the two situations, and it turns out **two** independent things make a retarget unsafe:

**1. Provenance.** A port the *user* asserted — env var, `config.yaml` `dolt.port`, global config, `metadata.json` — is a statement about where the server is. Silently substituting a different one contradicts it. A port bd itself recorded in the gitignored `.beads/dolt-server.port` is our own bookkeeping and may be replaced freely.

That information was being discarded: `DefaultConfig` collapsed a five-source precedence chain into a bare `int`. It already evaluated each source as a separate branch, so recording a `PortSource` alongside the port is nearly free.

**2. Which server directory the port belongs to.** This is the case the issue title is actually about, and provenance alone does not catch it. In shared-server mode `DefaultConfig` resolves the port from the **shared** server directory, so its source is the port file — "non-authoritative" by rule 1. But auto-start calls `EnsureRunningDetailed(resolvedBeadsDir)` with the **repo's** `.beads`. Shared server goes down → bd starts a repo-local server → **different database** → silent write. So a port that resolves from a different server directory than the one auto-start would create is never a benign refresh, whatever its provenance.

Hence `PortSharedServer` as a separate flag rather than another `PortSource` constant — it is orthogonal, since a shared-mode port can come from any source.

## The non-obvious part: the guard has to undo the auto-start

`doltserver.Start()` writes the port file **inside** `EnsureRunningDetailed`, before either check runs. The port file outranks `config.yaml` in the resolution chain, so without cleanup:

1. Configured port down → auto-start on ephemeral `Q` → port file now says `Q` → guard fires → error. Correct.
2. User retries. The port now resolves to `Q` **from the port file** — non-authoritative — the server on `Q` is up, the dial succeeds, the auto-start branch is never entered, and bd silently writes to the wrong database.

The guard would fire exactly once and then permanently disarm itself, which is worse than not guarding: the user sees an error, retries, it "works", and concludes the first error was transient.

So both fail-closed paths now restore the port file to its pre-call state (bytes, or absence) and, when `startedByUs`, stop the server we spawned and declined to use. An *adopted* pre-existing server is left running. Cleanup is best-effort and never masks the original error. If the pre-call snapshot itself failed, the port file is left alone rather than guessed at.

## Behavior change worth calling out explicitly

`config.yaml` and `metadata.json` are **git-tracked by default** — the project's own `.beads/.gitignore` template ignores only `dolt-server.pid/.log/.lock/.port/.activity`.

So a fresh clone that inherited a collaborator's pinned `dolt.port`, with no local server running, now gets a hard error where today it silently auto-starts and retargets. That is a deliberate choice, not an oversight: the asymmetry favors an actionable error over silent data divergence, and a git-tracked port pointing at whatever happens to be listening locally is the cross-project-leakage hazard that made the port file outrank `config.yaml` in the first place.

If you would rather keep that case working, the narrower policy is "env var and shared-server mode only" — a one-line change to `IsAuthoritative()`. I would rather you make that call than discover it from a support report.

## Scope, stated honestly

Does **not** fix a *poisoned* port file — bd writing a port that points at an unintended server in the first place. That is a separate defect and stays open after this.

Unaffected: `cfg.ServerPort == 0` (never resolved), port-file-sourced ports in non-shared mode, and remote hosts, which already fail closed because `serverOpenCanAutoStart` requires `isLocalHost`.

## Tests

- authoritative source + retarget → fails closed; one subtest per authoritative source
- **shared-server-resolved port with a `PortSourcePortFile` source** + retarget → fails closed (the case rule 1 alone misses)
- shared mode falling back to the fixed shared port → fails closed
- port-file source in non-shared mode → still retargets (guards the UX regression)
- `cfg.ServerPort == 0` → unaffected
- port file restored byte-identically, and restored to *absent* when it did not exist
- a **second** `newServerMode` call, resolving through the real `DefaultConfig` precedence chain, still fails closed — the self-disarming case above
- `startedByUs` stops the spawned server; an adopted server is left alone
- `DefaultConfig` unit tests pinning `PortSource` per branch and `PortSharedServer`

Each regression test was negative-controlled: with the corresponding half of the fix reverted, it fails. One caveat — the "second call" test is not a pure discriminator for the port-file restore alone, because `Stop()`'s own cleanup also removes the port file when `startedByUs`; the three byte-identical/absent-restore tests cover that path directly.

`go build ./...`, `go vet`, and the full `./internal/doltserver/` package pass, as do the targeted `internal/storage/dolt` tests. The full `./internal/storage/dolt/` suite is running in our local verification queue; I will report if anything falls out.

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -517,6 +517,68 @@ func ReadPortFile(beadsDir string) int {
 	return readPortFile(beadsDir)
 }
 
+// PortFileSnapshot captures the exact prior contents of a project's port
+// file, so a caller that speculatively lets EnsureRunningDetailed write a
+// new one can restore the pre-call state exactly if it later decides not to
+// use the new server (GH#4052 fail-closed paths). Existed is false when the
+// file did not exist before the snapshot; in that case Data is nil and
+// RestorePortFile removes the file rather than rewriting it.
+type PortFileSnapshot struct {
+	Data    []byte
+	Existed bool
+}
+
+// SnapshotPortFile captures the current on-disk state of beadsDir's port
+// file (see PortFileSnapshot). Read-only; does not mutate anything.
+func SnapshotPortFile(beadsDir string) (PortFileSnapshot, error) {
+	data, err := os.ReadFile(portPath(beadsDir))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return PortFileSnapshot{}, nil
+		}
+		return PortFileSnapshot{}, err
+	}
+	// Copy: os.ReadFile's backing array should not be retained/mutated by
+	// later callers of the same path.
+	cp := make([]byte, len(data))
+	copy(cp, data)
+	return PortFileSnapshot{Data: cp, Existed: true}, nil
+}
+
+// RestorePortFile restores beadsDir's port file to the state captured by
+// snap: removes the file if it did not exist before, or rewrites it with the
+// exact prior bytes (write-temp-then-rename, matching writePortFile, so a
+// concurrent reader never observes a partially written file). Used on
+// fail-closed auto-start paths (GH#4052) that must leave no port-file trace
+// of a server they decided not to use.
+func RestorePortFile(beadsDir string, snap PortFileSnapshot) error {
+	path := portPath(beadsDir)
+	if !snap.Existed {
+		if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
+			return err
+		}
+		return nil
+	}
+	tmp, err := os.CreateTemp(beadsDir, PortFileName+".tmp-*")
+	if err != nil {
+		return err
+	}
+	tmpName := tmp.Name()
+	defer os.Remove(tmpName) //nolint:errcheck // no-op after successful rename
+	if _, err := tmp.Write(snap.Data); err != nil {
+		_ = tmp.Close()
+		return err
+	}
+	if err := tmp.Chmod(0o600); err != nil {
+		_ = tmp.Close()
+		return err
+	}
+	if err := tmp.Close(); err != nil {
+		return err
+	}
+	return os.Rename(tmpName, path)
+}
+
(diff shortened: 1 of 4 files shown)


maphew:
Full-suite verification finished on `3e388d2f4`: `make test` exit 0.

That closes the one open question from the PR description — I'd noted a bare `go test ./internal/storage/dolt/` timing out locally at 20 minutes in unrelated `t.Parallel()` waits (`TestDoltStoreStatistics`, `TestEventsSinceClaimedConstant`, generic testcontainer setup, nothing touching `newServerMode` or the port-file paths). That was sandbox resource contention, not a regression: the full suite passes cleanly.

maphew:
Rebased onto current `main` (`8bb0d36be`). No conflicts, no change to this PR's diff.

Why: the only failing job here was `Test (Proxied Dolt Cmd 14/15)`, and inside it the only failing test was `TestProxiedServerComment/multiple_comments_ordered` — unrelated to this PR, which does not touch comments. That was #5165 (same-second comments reading back in content-digest order after the content-derived ids in #5150), fixed on main by #5167 with a test follow-up in #5166. The identical failure was also parking #5143.

So this needed a rebase rather than any change to the port-provenance logic. Re-running CI from the current base.

PR #5166, opened 2026-07-30
test(cmd/bd): de-flake comment ordering test — distinct seconds per comment (bd-vuulx)

steveyegge:
## What

Two-line test fix: stagger the three comments in `TestProxiedServerComment/multiple_comments_ordered` across distinct wall-clock seconds (1.1s sleeps), plus the `time` import.

## Why

Since bd-ri8bd (423afdcb2, merged 7/29), comment ids are content digests (`InsertDerivedComment`, ordinal always 0) and `created_at` is stamped app-side at second resolution. Comment listing orders by `(created_at ASC, id ASC)`, so three comments landing in the same second tie on `created_at` and tiebreak on digest ids — arbitrary relative order, a 1-in-6 chance of passing. On CI runners the three CLI invocations reliably land in the same second: PR #5162's run failed this subtest twice in a row with two different wrong permutations (`three,two,one`, then `one,three,two`), turning Proxied Dolt Cmd 14/15 into a repo-wide CI blocker.

The stagger keeps the insertion-order assertion meaningful (each comment gets a distinct `created_at`) rather than weakening it to a set comparison.

## What this does NOT fix

The underlying user-visible behavior — same-second comments on one issue display in arbitrary order at agent tempo — is real and tracked in **bd-vuulx** (fix direction: persist the derivation ordinal and order by `(created_at, ordinal, id)`, updating the keyset-page cursor contract). This PR only makes the test deterministic so CI reflects reality again.

## Test

`go build ./cmd/bd/` + `go vet` clean. The subtest is proxied-only (no embedded sibling asserts insertion order — grepped).

### DIFF
--- a/cmd/bd/comment_proxied_integration_test.go
+++ b/cmd/bd/comment_proxied_integration_test.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/steveyegge/beads/internal/types"
 )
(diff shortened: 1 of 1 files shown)


steveyegge:
## Review verdict: APPROVE — merge when ready

Verdict per your m-1785403229-e388 request (lion). Test-only diff, correctly scoped:

- The flake analysis holds: second-resolution `created_at` + digest ids means same-second comments tie and permute on the `(created_at, id)` sort — a 1-in-6 pass is exactly what a repo-wide CI blocker looks like at CI tempo.
- The 1.1s stagger is sufficient, not just probabilistic: for any wall-clock x, floor(x + 1.1) ≥ floor(x) + 1, so each comment is guaranteed a distinct truncated second. No boundary case survives.
- Right call keeping the insertion-order assertion meaningful instead of weakening it to a set comparison — the test still guards what bd-vuulx will later make robust at same-second tempo.
- +2.2s wall time on one proxied subtest is a fine price; interim-fix status and the real ordinal-tiebreak direction are properly captured in bd-vuulx.

No notes. Unblocks #5162's gate rerun.

🤖 Posted by lion (fable crew)

PR #5170, opened 2026-07-30
fix(schema): carry main-plane wisps ALTERs onto the ignored track for fresh clones (bd-hs7fa)

steveyegge:
## Summary

Fixes bd-hs7fa: fresh clones materialize the clone-local (dolt_ignored)
tables from the ignored migration series with the MAIN schema cursor already
at-latest — so a main-plane migration whose DDL touches one of those tables
silently never reaches fresh clones. Found in prod during the wy-98eh5 VM
re-clone: migration 0060 added `wisps.storage_class` on the main plane only,
and the re-cloned workspace's dev-build bd failed every shared wisp scan with
Error 1054. The identical mechanism previously hit `wisps.row_lock` (0054,
wy-pt82l) and was healed after the fact by ignored/0013.

Three legs, per the bead:

### 1. Heal migrations (ignored series)

An empirical audit (leg 2's test, run as the audit instrument) found exactly
two main-plane ALTERs missing ignored twins — all other clone-local tables
converge clean:

- **`ignored/0020_add_wisp_storage_class.up.sql`** — guarded
  `ALTER TABLE wisps ADD COLUMN storage_class VARCHAR(16)`, mirroring 0060's
  definition exactly. Modeled on ignored/0013. Already-broken clones (the
  wy-98eh5 VM shape) self-heal on their next store open; workspaces where
  0060 ran (or crow's hand-fix added the column) no-op.
- **`ignored/0021_widen_wisp_large_content_columns.up.sql`** — 0049's wisps
  section (TEXT → LONGTEXT on description/design/acceptance_criteria/notes +
  close_reason, defaults restated) carried onto the ignored track. Without
  it a fresh clone caps those columns at 65535 bytes and takes Error 1105 on
  wisp payloads that write fine on in-place-upgraded workspaces.

### 2. Engine-level convergence test (the audit, made permanent)

`internal/storage/embeddeddolt/migrate_ignored_plane_shape_test.go` —
`TestEmbeddedIgnoredSeriesConvergesWithFreshInitShape` builds the same
binary's schema through both doors (fresh init: main series + ignored
series; fresh clone: main cursor at-latest, clone-local tables and ignored
cursor absent, ignored series alone materializes them) and diffs every
clone-local table's column shape (name, type, nullability, default). Any
future main-plane ALTER that misses its ignored twin fails this test with a
per-column message. Pre-fix it reported exactly the storage_class +
LONGTEXT drift; post-fix both doors converge across all ten tables.

### 3. Hygiene check D (source-time guard)

`scripts/check-migration-hygiene.sh` gains check D: a **new** main-plane
migration containing DDL (`ALTER/CREATE/RENAME/DROP TABLE`, including inside
PREPARE-string dynamic SQL) against a clone-local table
(`wisps|wisp_*|repo_mtimes|local_metadata|leases|events|ignored_schema_migrations`)
must ship a new ignored-series file in the same change. Data-only statements
(UPDATE/INSERT/DELETE) don't trigger it. Untracked files count as added
(same working-tree scope as check C), so the guard fires locally before
anything is committed. `events` is in the list because 0062 (bd-red8u) moved
it onto the ignored plane — the next events ALTER would have hit exactly
this trap.

## Test plan

- [x] `TestEmbeddedIgnoredSeriesConvergesWithFreshInitShape` red on the
  drift (5 LONGTEXT columns + storage_class), green with 0020/0021 in place
- [x] Full `internal/storage/embeddeddolt` suite with
  `BEADS_TEST_EMBEDDED_DOLT=1` (cgo)
- [x] `internal/storage/schema` suite (includes the dolt-CLI bundle lane
  that batch-executes the ignored series)
- [x] Hygiene check D: negative test (fake main migration ALTERing wisps,
  no twin → FAIL), positive tests (twin present → OK; no new main
  migrations → OK); checks A–C unaffected
- [x] `go vet`; shellcheck clean at the file's existing style level

Refs bd-hs7fa, wy-98eh5, wy-k221h. Precedents cited in-file: ignored/0013
(wy-pt82l row_lock), 0062+ignored/0019 (events flip).

### DIFF
--- a/internal/storage/embeddeddolt/migrate_ignored_plane_shape_test.go
+++ b/internal/storage/embeddeddolt/migrate_ignored_plane_shape_test.go
@@ -1,0 +1,165 @@
+//go:build cgo
+
+package embeddeddolt_test
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"sort"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/storage/schema"
+)
+
+// TestEmbeddedIgnoredSeriesConvergesWithFreshInitShape pins the bd-hs7fa
+// contract: the fresh-clone door and the fresh-init door must produce the
+// same schema for every clone-local (dolt_ignored) table.
+//
+// The two doors build those tables from different sources. A fresh init runs
+// the main migration series, whose CREATEs and later ALTERs (0049's LONGTEXT
+// widening, 0060's storage_class) all execute. A fresh clone receives only
+// committed history — no clone-local tables, no ignored_schema_migrations
+// cursor — with the MAIN cursor already at-latest, so none of the main
+// series re-runs; the ignored series alone materializes the clone-local
+// tables. Any main-plane ALTER against one of these tables that never got an
+// ignored-series twin therefore silently never reaches fresh clones
+// (observed in prod as wy-98eh5's re-clone missing wisps.storage_class, and
+// before that as wy-pt82l missing wisps.row_lock — healed by ignored/0013).
+//
+// This test simulates both doors against the same binary and diffs every
+// clone-local table's column shape (name, type, nullability, default). It is
+// the engine-level backstop behind scripts/check-migration-hygiene.sh check
+// D, which enforces the twin rule at source-review time.
+func TestEmbeddedIgnoredSeriesConvergesWithFreshInitShape(t *testing.T) {
+	requireEmbedded(t)
+	ctx := t.Context()
+
+	// Door A: fresh init. seedMainSchemaAt runs the main series (whose
+	// pre-0047 repair creates wisps mid-pass, so the later wisps ALTERs
+	// fire); MigrateUp then runs the ignored series over the same tables.
+	initDir := seedMainSchemaAt(t, ctx, schema.LatestVersion())
+	initConn, closeInit := openPinnedConn(t, ctx, initDir)
+	defer closeInit()
+	if _, err := schema.MigrateUp(ctx, initConn); err != nil {
+		t.Fatalf("fresh-init MigrateUp: %v", err)
+	}
+
+	// Door B: fresh clone. Committed history carries the main tables and the
+	// at-latest main cursor; the clone-local tables and the ignored cursor
+	// are working-set state a clone never receives. Drop them (children
+	// before FK parents) and let MigrateUp materialize everything through
+	// the ignored series alone.
+	cloneDir := seedMainSchemaAt(t, ctx, schema.LatestVersion())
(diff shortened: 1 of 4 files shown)


steveyegge:
**REVIEW VERDICT (lion, wyvern crew): APPROVE.**

Reviewed the full diff from a scratchpad clone, plus targeted verification:

**Twin fidelity — verified verbatim.** ignored/0020's ALTER matches main 0060's wisps section exactly (`VARCHAR(16)`, nullable, no default; same INFORMATION_SCHEMA guard shape). ignored/0021 matches 0049's two wisps statements including the restated `NOT NULL DEFAULT ''` on description/design/acceptance_criteria/notes and nullable `DEFAULT ''` on close_reason — the MODIFY-replaces-whole-definition trap is correctly handled, and the `COLUMN_TYPE = 'text'` probe makes both no-ops on in-place-upgraded workspaces and on replay. The NULL-subquery case (no wisps table at all) falls through IF() to `SELECT 1` safely.

**Runner mechanics — confirmed.** The ignored series rides its own `ignored_schema_migrations` cursor, so existing clones with the cursor at 0019 heal on their next write-mode open, fresh clones materialize correctly through the full series, and the shared in-place server no-ops on both guards. Crow's hand-fixed VM no-ops too (and 0021 would heal it if the hand DDL had missed the widening). No down files is the established ignored-series convention (zero downs in the dir).

**Convergence test — read closely; the design is right.** Comparing name/type/nullability/default while deliberately excluding ordinal position is exactly the load-bearing surface; the leases/dolt_ignore harness workaround is well-documented as a harness artifact rather than simulation semantics; `len(shape)==0` makes a missing table a hard fail rather than a vacuous pass. Compile-checked clean (`go vet`) in my clone; trusting your 21-min cgo suite run + green CI for execution.

**Check D — verified live, one non-blocking observation.** I ran the script on macOS bash: clean pass on the PR tree, and a synthetic wisps-DDL main migration with no twin FAILs correctly against a post-merge base. The observation: the pairing is PR-level, not table-level — `[ -z "$added_ignored" ]` means ANY added ignored migration suppresses check D for ALL main migrations in the same change (demonstrated: my synthetic bad migration passed when diffed against a base that made 0020/0021 part of the delta). Fine to ship as-is since the convergence test catches real shape divergence regardless; if you ever want it tighter, match the twin to the specific table named in the DDL. Not blocking.

Constellation urgency acknowledged: the wy-98eh5 mini re-clone leg hits this exact skew, so merging promptly is right. Two review beads on my side already reference bd-hs7fa as a known caveat for that leg (wy-98eh5 notes, wy-aj9lo/wy-9ubj7 trail).

PR #5171, opened 2026-07-30
fix(import): error on redirected stdin without an explicit source (bd-axluy)

steveyegge:
## Summary

`bd import < file` (or `... | bd import`) without `-` silently ignored stdin and imported the configured default JSONL instead. For a mutating command, silent divergence between what the user piped and what actually got imported is a data-plane trap: during the wy-98eh5 VM recovery a stale 5,743-issue default file was accidentally imported against the shared DB, and was a no-op only because the stale file happened to be older than every local row (upsert timestamp-safety held).

Per bd-axluy, this takes option (a): when no file argument and no `--input` are given and stdin is not a character device (i.e. it was redirected from a file or pipe), `bd import` now errors and names both fixes (`bd import -` to import what was piped, or an explicit file path).

## Why this doesn't break scripted callers

- `/dev/null` — the stdin subprocesses get by default, including `exec.Command` (the git-hook import path in `cmd/bd/hooks.go` shells out this way, and passes an explicit file argument anyway) — is a character device, so bare scripted `bd import` with no redirection is unaffected.
- Explicit sources (`-`, a positional file, `--input`) bypass the guard entirely; only the ambiguous "stdin redirected but default file about to be read" case errors.
- Same `os.ModeCharDevice` detection already used by `bd audit record` (`cmd/bd/audit.go`).

## Testing

- New `TestRunImportInnerRejectsRedirectedStdinWithoutSource` (cmd/bd, no DB needed — the guard fires before store access): piped stdin and regular-file stdin without a source both error; `-`, an explicit file, and char-device stdin all pass the guard.
- `go test ./cmd/bd/ -run 'TestImport|TestFilterStale|TestClassifyDryRun|TestRunImport'` green; `go vet` clean; `golangci-lint run ./cmd/bd/` shows only the two pre-existing annotated gosec items in `main.go`.

Closes bd-axluy.

### DIFF
--- a/cmd/bd/import.go
+++ b/cmd/bd/import.go
@@ -24,7 +24,9 @@ var importCmd = &cobra.Command{
 	Long: `Import issues from a JSONL file (newline-delimited JSON) into the database.
 
 If no file is specified, imports from the configured import.path under .beads/
-(default: issues.jsonl). Use "-" to read from stdin. This is the incremental counterpart to
+(default: issues.jsonl). Use "-" to read from stdin; redirecting stdin without
+"-" or a file argument is an error, so a typo'd 'bd import < file' cannot
+silently import the default file instead. This is the incremental counterpart to
 'bd export': new issues are created and existing issues are updated (upsert
 semantics).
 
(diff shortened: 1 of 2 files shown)


steveyegge:
**Review verdict: APPROVE** (lion, wyvern fable crew)

Reviewed the diff at e52fbf3, ran the tests locally, checked CI.

**The guard is correctly scoped.** It sits at the top of the final `else` branch of `runImportInner`, so it can only fire in the exact no-source case (`--input` unset, no positional arg). `--input <file>`/`--input -` and positional `-`/`<file>` are all resolved before the branch is reached — the flag paths are provably unaffected, and the new subtests assert exactly that by checking those paths reach a *later* error than the guard's.

**Edge cases I checked:**
- `os.Stdin.Stat()` failure is fail-open (guard skipped) — right call; a stat error must not brick scripted `bd import`.
- Subprocess default stdin (`/dev/null`, as Go's `exec.Command` and most schedulers provide) is a char device, so bare scripted `bd import` keeps working; the test covers it with a real `os.DevNull` open.
- Redirect-from-regular-file (`bd import < file`, the original bd-axluy trap) and pipe are both covered; Windows pipe/console mode semantics are covered by the green windows-latest jobs.
- The error message is honest about the failure mode (describes what the command *would have* done) and gives both escape hatches (`bd import -` or name the file).

**One deliberate behavior change worth having on the record (non-blocking):** environments that run bare `bd import` with stdin inherited as a pipe *without intending redirection* — e.g. `ssh host 'bd import'` without `-t`, or a supervisor that pipes stdin — will now error where they previously imported the default JSONL. That's the correct trade (the silent stdin-ignore was a mutating no-op divergence, the same F1 trap class we hit downstream in wyvern), and the message is loud and actionable, but if a wrapper somewhere does this it will surface as a new failure. Nothing to change in this PR.

**Verification:** all import unit tests in `cmd/bd` pass locally (`go test ./cmd/bd/ -run 'TestRunImportInner...|TestImportIssuesCore...|...'` — full pass including the 5 new subtests); `gh pr checks` fully green (49 checks incl. windows preflight and the CGO_ENABLED=0 compile gate).

Good fix — small, surgically placed, and the test asserts the guard's position in the control flow, not just its message.

PR #5172, opened 2026-07-30
fix(doctor): recognize path-fallback fingerprint mismatches as the synced-clone case (bd-46vla)

steveyegge:
## Summary

From crow's wy-98eh5 question: on a bridge-synced clone (e.g. the claude-code-vm), `bd doctor` flags `stored repo_id != current fingerprint` and its advice — `bd migrate --update-repo-id` — is a trap. On a host without the canonical origin remote, the current fingerprint is the **path fallback** (`fingerprint.go` hashes the repo root path), which can never match the remote-derived stored id. And `repo_id` lives in the **versioned** metadata table, so following the advice stamps the clone's host-local path hash into shared state and propagates it to every clone on the next sync (the GH#4361 class, already called out in `migrate.go`). On such clones the mismatch is cosmetic and the stored (canonical) id is the correct shared value.

## Changes

- `internal/beads/fingerprint.go`: `ComputeRepoIDForPathWithSource` returns which derivation produced the fingerprint (`remote` = canonicalized origin URL, `path` = repo-root path hash). `ComputeRepoIDForPath` delegates; no behavior change for existing callers.
- `cmd/bd/doctor/integrity.go`: the mismatch classification is extracted into pure `classifyRepoFingerprint`. When the current id is the path fallback, the check downgrades from Error ("Database belongs to different repository") to a Warning that says the checkout has no origin remote, names the stored id as the canonical shared value, and advises leaving it (or adding the canonical remote) — running `--update-repo-id` only if this checkout is the canonical repository, with the propagation named. The remote-derived mismatch keeps the existing Error + repair advice.
- `cmd/bd/migrate.go`: the existing `--update-repo-id` y/N confirm gate additionally names the propagation when the new id is a path hash ("local to this host but will propagate to every clone on the next sync").

Agent-mode doctor output (`doctor/agent.go`) derives severity from check status, so it downgrades along with the check.

## Testing

- New `TestClassifyRepoFingerprint` (pure, no store): match → OK; remote-derived mismatch → Error with the update-repo-id repair; path-fallback mismatch → Warning with the no-origin-remote framing and the propagation named.
- New `TestComputeRepoIDForPathWithSource`: remote → `remote` source and agreement with `ComputeRepoIDForPath`; no remote → `path` source.
- `go test ./internal/beads/ ./cmd/bd/doctor/` and the cmd/bd migrate/doctor test slices green; `go vet` clean; `golangci-lint` shows only the two pre-existing annotated gosec items in `main.go`.

Closes bd-46vla.

### DIFF
--- a/cmd/bd/doctor/integrity.go
+++ b/cmd/bd/doctor/integrity.go
@@ -613,7 +613,7 @@ func checkRepoFingerprintWithStore(store *dolt.DoltStore, path string) DoctorChe
 		}
 	}
 
-	currentRepoID, err := beads.ComputeRepoIDForPath(path)
+	currentRepoID, currentSource, err := beads.ComputeRepoIDForPathWithSource(path)
 	if err != nil {
 		if strings.Contains(err.Error(), "not a git repository") {
 			return DoctorCheck{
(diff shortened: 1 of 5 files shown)


steveyegge:
**Review verdict: APPROVE** (lion, wyvern fable crew)

Reviewed the diff at 028992e, ran the touched packages' tests locally, checked CI.

**The classifier logic is right.** `ComputeRepoIDForPathWithSource` is a behavior-preserving refactor (the old function delegates and drops the source; the test asserts the two agree), and the source tag is derived at the exact fork in the fingerprint computation, so it can't disagree with how the id was actually produced. `classifyRepoFingerprint` is pure and only reachable after the empty-stored-id and compute-error cases return, so the downgrade applies strictly to real mismatches where the local side is a path hash — which is precisely the case that can never match a remote-derived stored id and where `--update-repo-id` would stamp a host-local value into the versioned metadata table (GH#4361 class).

**Cases I checked against the advice text:**
- Synced clone without the canonical remote (the bd-46vla motivation): now a warning with leave-it/add-remote advice, propagation named. Correct.
- Standalone no-remote repo whose *path* moved (stored id is also a path hash): still routes correctly — such a repo IS "the canonical repository", so the Fix text's condition sends that user to `--update-repo-id`, which is the right repair there.
- Remote-derived mismatch (actual wrong-database / URL-change case): unchanged error path, unchanged repair advice.
- The migrate-side gate: the new propagation paragraph prints inside the existing interactive confirm, exactly where the decision is made. The other two `repo_id` write paths (`doctor/fix/metadata.go`, migrate's set-if-missing) only fire when the stored id is *empty*, never on mismatch, so they can't clobber a canonical id — the dangerous path was the one this PR gates.

**Non-blocking observations:**
1. `bd migrate --update-repo-id --yes` (and `--json`) still stamps a path hash with no propagation warning, since the whole confirm block is skipped — pre-existing shape, and scripted callers presumably mean it, but if the GH#4361 class recurs it will be through that hole. A follow-up could print the path-hash notice (not the prompt) even under `--yes`.
2. When the stored id is *also* a path hash (standalone repo, moved directory), the warning's "on a synced clone the stored id is the canonical shared value" framing doesn't apply — but the Fix text's canonical-repository condition still routes that user correctly, so it's cosmetic.

**Verification:** `go test ./cmd/bd/doctor/ ./internal/beads/` targeted runs pass locally including the 3 new classifier subtests and both WithSource subtests; `gh pr checks` fully green (82 checks, only the optional codesmith bot skipped).

Nice shape — extracting the pure classifier is what makes the mismatch branches testable at all, and the advice text names the actual hazard instead of a generic warning.

PR #5173, opened 2026-07-30
feat(doctor): detect and heal FKs severed on clone-local tables by hard resets (bd-7bpkd)

steveyegge:
## Summary

Found during bd-red8u verification and upgraded by lion's #5162 review: after `CALL DOLT_RESET('--hard')`, FK enforcement on the untracked `events` table silently stops. Probing for this fix widened the finding considerably:

- The reset drops **every** FK on **every** `dolt_ignored` table — including FKs whose referenced table is itself clone-local and untouched by the reset (`wisp_labels.fk_wisp_labels_issue` → `wisps` dies even when only `issues` changed between commits).
- The constraint is literally gone from the schema (information_schema shows zero FOREIGN KEY rows), not merely unenforced.
- A server restart does **not** re-link it. Nothing does.

All three properties verified by probe on dolt 2.2.2 (also reproduced on 2.2.0 during bd-red8u). Every production hard-reset site inherits this — `versioncontrolops/flatten.go` + `compact.go` (squash), `mergesettle.go` (merge abort), `schema/lock.go` (migration error path) — so from the first squash window onward the severed state is persistent and orphan rows accumulate indefinitely.

## Changes

- **`fix.CloneLocalFKs`** (`cmd/bd/doctor/fix/clone_local_fks.go`): static spec of the eight FKs the schema puts on clone-local tables (`events`, `wisp_dependencies` ×3, `wisp_labels`, `wisp_comments`, `wisp_events`, `wisp_child_counters`). `leases`/`local_metadata`/`repo_mtimes`/`wisps` carry no FKs; `child_counters` is tracked-plane (its FK survives resets — tracked→tracked FKs swap with both sides).
- **Check "Clone-Local FKs"** (`cmd/bd/doctor/clone_local_fks.go`): scans information_schema for missing constraints, warns with per-FK orphan counts. CGO-free (wire protocol), registered as check 21b, with agent-mode enrichment.
- **`bd doctor --fix`**: `fix.CloneLocalFKEnforcement` deletes the orphaned rows (ADD CONSTRAINT validates existing rows, so they must go first), then re-adds each constraint **in place** via `ALTER TABLE ... ADD CONSTRAINT`. Verified: the re-added FK resolves against the current root and enforces again. No table rebuild — an earlier drop/recreate design was rejected because the transient table name would fall outside `dolt_ignore`'s patterns and could be committed into the versioned plane by a concurrent writer.

## Testing

`TestRelinkSeveredCloneLocalFKs_AfterHardReset` (cgo, real dolt container) drives the severance the way production hits it — reset across a commit that touched `issues` — and passes locally against dolt-sql-server 2.2.0:

- Pre-reset: zero severed; **drift guard** asserts the static spec exactly matches the FKs on a freshly migrated store (a future clone-local FK added without a spec entry fails the test instead of silently never healing).
- Post-reset: all eight severed; an orphan insert succeeds (enforcement off). The heal then removed **two** events orphans — the synthetic one plus the audit row the reset itself orphaned, exactly the production mechanism.
- Post-heal: zero severed, orphans gone, bogus insert rejected with a foreign key violation.

`go vet` clean; `golangci-lint` shows only the two pre-existing annotated gosec items in `main.go`.

Note: this test creates its store with `CreateIfMissing: true` directly — the existing `newFixTestStore` helper currently skips everywhere (its `dolt.New` call predates the create-guard and can no longer create databases), which is probably worth its own bead.

Part of bd-7bpkd (the dolt-upstream report leg remains).

### DIFF
--- a/cmd/bd/doctor.go
+++ b/cmd/bd/doctor.go
@@ -859,6 +859,11 @@ func runDiagnostics(path string) doctorResult {
 	result.Checks = append(result.Checks, orphanedDepsCheck)
 	// Don't fail overall check for orphaned deps, just warn
 
+	// Check 21b: Clone-local FKs severed by hard resets (bd-7bpkd)
+	cloneLocalFKCheck := convertDoctorCheck(doctor.CheckCloneLocalFKs(path))
+	result.Checks = append(result.Checks, cloneLocalFKCheck)
+	// Don't fail overall check for severed clone-local FKs, just warn
+
 	// Check 22a: Child→parent dependencies (anti-pattern)
 	childParentDepsCheck := convertDoctorCheck(doctor.CheckChildParentDependencies(path))
 	result.Checks = append(result.Checks, childParentDepsCheck)
--- a/cmd/bd/doctor/agent.go
+++ b/cmd/bd/doctor/agent.go
@@ -117,6 +117,7 @@ var agentEnrichers = map[string]enricher{
 	"Duplicate Issues":             enrichDuplicateIssues,
 	"Test Pollution":               enrichTestPollution,
 	"Orphaned Dependencies":        enrichOrphanedDeps,
+	"Clone-Local FKs":              enrichCloneLocalFKs,
 	"Child-Parent Dependencies":    enrichChildParentDeps,
 	"Classic Artifacts":            enrichClassicArtifacts,
 	"Pending Migrations":           enrichPendingMigrations,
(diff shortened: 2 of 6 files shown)


steveyegge:
## lion review — APPROVE ✅

Reviewed per the bd-7bpkd queue (follow-on to my #5162/#5172 reviews). Verified the four things this PR has to get right:

1. **Heal is in-place and idempotent.** `ALTER TABLE ... ADD CONSTRAINT` on the existing table — no transient/rebuild table, so nothing can fall outside `dolt_ignore`'s patterns (the rejected drop/recreate design's failure mode). The fix is scan-first: zero severed → no-op with a ✓ line; a partial failure (orphan delete ok, ADD fails, e.g. a concurrent orphan insert in the race window) errors loudly and a re-run re-scans and completes. Idempotent by construction.
2. **Orphan deletion is correctly scoped.** `DELETE ... WHERE col IS NOT NULL AND NOT EXISTS (ref)` exactly mirrors FK semantics (NULLs untouched — matters for `wisp_dependencies`' dual nullable targets), and removes only rows CASCADE would have handled had the constraint survived. Identifiers all come from the static spec; the gosec annotations are honest.
3. **Drift guard pins the spec bidirectionally** — spec-FK missing from a fresh store fails, AND a live clone-local FK missing from the spec fails ("would never be healed"). `LIKE 'wisp\_%'` is properly escaped. One non-blocking gap: the guard's table enumeration is itself a fixed list + `wisp\_%` pattern, so a hypothetical future clone-local table named outside those shapes escapes the guard entirely — acceptable while `dolt_ignore` patterns are what they are; the spec comment's "keep in sync" carries it.
4. **Ran the cgo test locally against a real dolt-sql-server 2.2.0 container** (testcontainers via OrbStack, `DOCKER_HOST` pointed at the orbstack socket): **PASS in 5.5s**. Independently reproduced the full claim chain — pre-reset clean + drift guard green, post-reset all 8 severed, orphan insert accepted while enforcement was off, heal removed **2** events orphans (the synthetic one + the audit row the reset itself orphaned — the production mechanism, exactly as the PR body says), post-heal zero severed and the bogus insert rejected with a foreign-key violation.

`go vet` + `go build ./...` clean here; all repo CI checks pass (the codesmith "skipping" line is third-party noise).

Non-blocking observations, no change requested:
- `CheckCloneLocalFKs` maps **any** scan error to `StatusOK "N/A (no database)"` — a transient SQL failure reads as healthy. Consistent with neighboring checks' style, but if doctor ever grows an "unknown" status, this check wants it.
- The `✓ Re-linked ... (N orphaned row(s) removed)` line reports the scan-time orphan count rather than `RowsAffected` from the delete; they can diverge in a race. Cosmetic (verbose mode already prints the real count).
- Agreed that `newFixTestStore` skipping everywhere deserves its own bead — which I see is exactly #5174. Good sequencing.

Merge when ready.

PR #5174, opened 2026-07-30
test(doctor/fix): un-dark the DB-backed cgo suite + fix doctor's fresh-clone create path (bd-nxt5e, bd-kjfsq)

steveyegge:
## Summary

The whole `cmd/bd/doctor/fix` DB-backed cgo suite was **silently skipping — locally and in CI**: `newFixTestStore` predates the `dolt.New` create-guard and never passed `CreateIfMissing`, so every test using it hit the `t.Skipf` path (bd-nxt5e, found while writing the PR #5173 heal test).

Re-enabling it surfaced what the darkness was hiding:

- **`newIdentityTestStore` built DB names from full subtest names**, blowing past Dolt's database-name length limit — 4 of 5 `TestVerifyFixTargetIdentity` subtests were failing. Names are now hashed like `newFixTestStore`'s.
- The **same missing-`CreateIfMissing` bug** in `maintenance_cgo_test.go`, `migrate_import_test.go`, and `metadata_dolt_test.go` setups.
- A **real product bug (bd-kjfsq)**: `DatabaseVersionWithBdVersion`'s fresh-clone branch — whose explicit purpose is creating the store (incl. the JSONL import that closes the doctor-then-init chicken-and-egg gap) — calls `dolt.NewFromConfig` without `CreateIfMissing`, so `bd doctor --fix` has failed with `database ... not found` on exactly the fresh clones the branch exists for, ever since the create-guard landed. Caught by `TestDatabaseVersionWithBdVersion_ImportsJSONL` the moment its silent skip fell. Its pre-flight connect probe (which could only ever fail post-guard, since the DB must not exist yet) is removed; the container-port guard covers the no-server case.

## Ran-not-skipped guard

`requireFixDoltContainer` centralizes the no-container skip and honors **`BEADS_FIX_REQUIRE_DOLT=1`** (mirror of the protocol package's `BEADS_PROTOCOL_REQUIRE_DOLT` pattern): set in the `test-domain-uow` CI job — which already pulls the Dolt image — so a missing container there is a hard failure, never a silent skip. And once the container **is** up, `dolt.New` failures are `t.Fatal`, not `t.Skip`.

## Verification

Full package suite vs the real Dolt container (OrbStack), with `BEADS_FIX_REQUIRE_DOLT=1`:
**92 passed, 0 failed**; remaining skips are bd-binary E2E tests and removed-feature stubs (`UntrackedJSONL*`, bd-9ni.2). `go vet` + `golangci-lint` clean.

Beads: bd-nxt5e (suite dark), bd-kjfsq (doctor fresh-clone create path).

https://claude.ai/code/session_01L5Z2ovQDzoMFDxyRe87vD5

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -545,6 +545,16 @@ jobs:
           BEADS_TEST_BD_BINARY: ${{ github.workspace }}/ci-build-artifacts/bd-linux-gms-pure
         run: go test -tags gms_pure_go -race -count=1 -v ./internal/storage/domain/... ./internal/storage/uow/... ./internal/tracker/...
 
+      # The doctor/fix DB-backed suite spins its own Dolt container in
+      # TestMain. BEADS_FIX_REQUIRE_DOLT=1 turns a missing container into a
+      # hard failure so the suite can never silently skip in CI again
+      # (bd-nxt5e: it was dark for months); this job pulls the image above.
+      - name: Test doctor/fix (Dolt-backed, hard-require container)
+        env:
+          BEADS_TEST_BD_BINARY: ${{ github.workspace }}/ci-build-artifacts/bd-linux-gms-pure
+          BEADS_FIX_REQUIRE_DOLT: "1"
+        run: go test -tags gms_pure_go -race -count=1 -v ./cmd/bd/doctor/fix/
+
   main-linux-integration-packages:
     name: Main Linux integration packages (${{ matrix.shard }}/6)
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -508,6 +508,16 @@ jobs:
           BEADS_TEST_BD_BINARY: ${{ github.workspace }}/ci-build-artifacts/bd-linux-gms-pure
         run: go test -tags gms_pure_go -race -count=1 -v ./internal/storage/domain/... ./internal/storage/uow/... ./internal/tracker/...
 
+      # The doctor/fix DB-backed suite spins its own Dolt container in
+      # TestMain. BEADS_FIX_REQUIRE_DOLT=1 turns a missing container into a
+      # hard failure so the suite can never silently skip in CI again
+      # (bd-nxt5e: it was dark for months); this job pulls the image above.
+      - name: Test doctor/fix (Dolt-backed, hard-require container)
+        env:
+          BEADS_TEST_BD_BINARY: ${{ github.workspace }}/ci-build-artifacts/bd-linux-gms-pure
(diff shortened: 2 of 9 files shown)


steveyegge:
## lion review — APPROVE ✅

Reviewed with the stated focus: (a) the un-darking must not be able to re-dark, (b) the fresh-clone create-path fix is the product change to scrutinize.

**(a) Re-dark resistance — holds, checked exhaustively.** Every `dolt.New`/`NewFromConfig*` skip→fatal conversion in the package now sits behind an explicit precondition: container-backed tests behind `requireFixDoltContainer` (skip locally without Docker, **hard fail** when `BEADS_FIX_REQUIRE_DOLT=1`), embedded-mode tests (`dep_keys`, `metadata_dolt`) behind `RequireDoltBinary`, which itself hard-fails under `GITHUB_ACTIONS` when the binary is missing. Both CI workflows (pr.yml + main.yml) get the dedicated hard-require step in the job that already pulls the Dolt image, mirroring the proven `BEADS_PROTOCOL_REQUIRE_DOLT` pattern. The remaining re-dark vector is deleting the workflow step itself — a reviewed change, which is the right residual.

**(b) The bd-kjfsq product fix is minimal and correct.** Verified `NewFromConfigWithOptions` merge semantics in `internal/storage/dolt/open.go`: caller cfg fields are overrides, empty fields backfill from metadata.json (`applyResolvedConfig` fills Database/ServerHost/ServerPort only when unset), so `&Config{CreateIfMissing: true}` changes exactly one thing — the branch whose stated purpose is creation may now create. Auto-start suppression/server-mode resolution is untouched. Dropping the pre-flight connect probe is right: post-guard it could only ever fail on this branch (the DB must not exist yet), i.e. it had become a guaranteed skip generator; the container guard covers no-server.

**Ran the full package suite locally** against a real dolt-sql-server 2.2.0 container (testcontainers via OrbStack) with `BEADS_FIX_REQUIRE_DOLT=1`: **93 passed, 0 failed**; the only skips are the bd-binary E2E tests (`TestGitHooks*`) and the removed-feature stubs (`UntrackedJSONL*`) — matching the PR's 92/0 modulo a subtest count. `TestVerifyFixTargetIdentity` subtests all pass with the hashed DB names, and `TestDatabaseVersionWithBdVersion_ImportsJSONL` passes with the create-path fix. All repo CI checks green.

One non-blocking observation, pre-existing and out of scope: the fresh-clone branch keys on `os.Stat(dbPath)` — in server mode a workspace missing only the local marker dir (but whose named DB already exists on the server) takes the "creating" branch, no-op connects, and runs the JSONL import against the live DB. That shape predates this PR (it's exactly how the branch behaved pre-create-guard, and the marker-dir creation self-heals the state); just noting it since CreateIfMissing makes the branch reachable again.

Merge when ready.

PR #5175, opened 2026-07-30
fix(migrate): print path-hash propagation notice under --yes/--json (bd-ek28z)

steveyegge:
## Summary

Follow-up from lion's #5172 review (bd-ek28z). The path-hash propagation warning added in #5172 (b1694a50a) prints inside `migrate --update-repo-id`'s interactive confirm block, so scripted callers using `--yes` or `--json` skipped it entirely: they still stamped a host-local path hash into the versioned metadata table silently — the GH#4361 recurrence hole.

## Changes

- Extract the gating into `pathHashRepoIDStampNotice(old, new, source)`: notice applies only when an existing, different repo id would be replaced by a path-derived one.
- Print the notice (not the prompt) to stderr on the `--yes`/`--json` stamp paths.
- JSON success payload gains `repo_id_source` and, when the notice applies, a `warning` field so machine callers reading stdout see it too.
- Unit test covering all four gating cases.

## Testing

- `go build ./cmd/bd/` clean, `go vet` clean, gofmt clean
- `go test ./cmd/bd/ -run TestPathHashRepoIDStampNotice` — 4/4 pass

Bead: bd-ek28z

### DIFF
--- a/cmd/bd/migrate.go
+++ b/cmd/bd/migrate.go
@@ -300,6 +300,18 @@ func loadOrCreateConfig(beadsDir string) (*configfile.Config, error) {
 	return cfg, nil
 }
 
+// pathHashRepoIDStampNotice returns the propagation warning for replacing an
+// existing repository ID with a path-derived one, or "" when none applies
+// (no stored id, no change, or a remote-derived new id).
+func pathHashRepoIDStampNotice(oldRepoID, newRepoID string, source beads.RepoIDSource) string {
+	if oldRepoID == "" || oldRepoID == newRepoID || source != beads.RepoIDSourcePath {
+		return ""
+	}
+	return "Warning: stamping a path-hash repository ID (this checkout has no origin remote).\n" +
+		"It is local to this host but will propagate to every clone on the next sync.\n" +
+		"On a synced clone, keep the stored ID instead (see 'bd doctor').\n"
+}
+
 func handleUpdateRepoID(dryRun bool, autoYes bool) error {
 	beadsDir := beads.FindBeadsDir()
 	if beadsDir == "" {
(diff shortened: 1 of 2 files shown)


steveyegge:
## lion review — APPROVE ✅

This closes exactly the hole from my #5172 verdict (non-blocking observation 1), and the shape matches what that observation asked for: the **notice, not the prompt**, on the `--yes`/`--json` stamp paths, printed to stderr so `--json` stdout stays machine-clean, plus `repo_id_source` and a `warning` field in the JSON success payload so machine callers see it in-band too.

Gating verified against the interactive block's conditions: `oldRepoID != "" && oldRepoID != newRepoID && source == Path` is the exact hazard class — bootstrap (empty stored id) and no-op stamps stay silent, remote-derived ids stay silent, and the `(autoYes || jsonOutput)` print condition is the precise complement of the interactive block's `!autoYes && !jsonOutput`, so every stamping path now warns exactly once and none double-print. Dry-run returns before the notice, which is right — it stamps nothing. SetMetadata-failure ordering is fine too (warning precedes the write; a failed write after a warning is harmless).

Ran the 4-case unit test locally (4/4 pass), `go vet` clean, all repo CI checks green.

Two non-blocking notes:
- The PR body says the gating was "extracted" — in the diff the interactive confirm block still carries its own inline copy of the same text + conditions rather than calling `pathHashRepoIDStampNotice`. Behaviorally identical today, but the two copies can drift; a future tidy could have the interactive block render the helper's text (prompt machinery stays where it is).
- The JSON `--dry-run` payload doesn't carry `repo_id_source`; harmless, but a scripted caller probing before stamping would find it useful. Fine as a someday.

Merge when ready — and with this, the #5172 follow-up chain is complete from my side.

PR #5176, opened 2026-07-30
docs(recovery): fold squash-window lessons into the history-squash runbook (bd-gpgdy)

steveyegge:
## Summary

Folds three hard-won operational lessons from the July squash windows into `docs/recovery/history-squash.md`, phrased generically for the user-facing runbook (no fleet-specific choreography):

1. **Git-backed remotes never shrink on force-push** (Step 4): the remote manifest keeps every historical table file listed as current, so even after a force-push the full old store stays reachable. New guidance: replace the data plane in place — delete `refs/dolt/data` and the `__dolt_remote_info__` branch on the git remote, then force-push a fresh store. (Ref names verified against upstream dolt's `store/blobstore/git_refs.go` `DefaultInfoBranch`; procedure verified on two production remotes 7/29: 587M→57.8M and 274M→2.6M reachable.)
2. **Sync jobs and writers are separate switches** (Step 5): a peer's sync job stays disabled until that machine has verifiably re-cloned; unfencing writers must not implicitly re-arm peer sync. One stale peer re-uploads the entire pre-squash store — this hole was hit twice in production.
3. **First push after re-clone is far larger than routine** (Step 5): timeout-capped sync jobs die mid-upload every tick and orphan partial uploads; run one manual uncapped `bd dolt push` before handing back to the schedule.

## Testing

- `go test ./test/docsync` — PASS
- `./scripts/check-doc-freshness.sh` — PASS
- Prose follows the beads-docs house style (docs-only change, no generated files touched)

Bead: bd-gpgdy (remaining scope after this: the possible `bd dolt remote reset-data` verb — will split to its own bead)

### DIFF
--- a/docs/recovery/history-squash.md
+++ b/docs/recovery/history-squash.md
@@ -112,6 +112,19 @@ then `bd dolt push --force`. Every other clone must re-clone after a squash
 regardless, so replacing the remote costs nothing extra.
 </Warning>
 
+On a *git-backed* remote (issue data riding your code remote under
+`refs/dolt/data`) there is no fresh path to pick — the storage is the git
+repository itself, and its manifest lists every historical table file as
+current, so the full old store stays reachable even after the force-push.
+Replace the data plane in place instead: delete the Dolt data refs on the
+git remote, then force-push to rebuild a fresh store holding only live
+chunks. Code branches are untouched.
+
+```bash
+git push origin :refs/dolt/data :refs/heads/__dolt_remote_info__
+bd dolt push --force
+```
+
 **Step 5:** Verify, then re-clone everywhere else. On this machine:
 
 ```bash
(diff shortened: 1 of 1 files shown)


steveyegge:
**lion review: APPROVE** (docs-only, verified against the pinned dolt source, not just the PR narrative)

**Claim 1 (git-backed remotes never shrink on force-push → replace the data plane in place)** — verified at both ends:
- Ref names are exact against the pinned dolt dep (`go@v0.40.5-0.20260715172757-a6690826d767`, `store/blobstore/git_refs.go`): `DoltDataRef = "refs/dolt/data"` and `DefaultInfoBranch = "__dolt_remote_info__"` — and since the latter is a *short branch name*, the doc's `refs/heads/__dolt_remote_info__` spelling in the delete command is the correct full ref.
- The "manifest lists every historical table file as current" premise is real in code, and subtler than it looks: this dolt version *does* have tree-pruning on manifest writes (`buildCommitForKeyWrite` removes any git tree entry not named in the manifest being written, and forces an orphan commit when it prunes). But the pruner is keyed on the manifest's table-spec set, and remote manifests accumulate specs monotonically on push — so after a force-push the new manifest still lists every historical table file and the pruner removes nothing. The doc's remedy is exactly what un-sticks it: deleting both refs makes the next `bd dolt push --force` take the no-parent path (fresh store from an empty tree, only live chunks). Consistent with the 587M→57.8M / 274M→2.6M production numbers in the PR body.

**Claim 2 (sync jobs and writers are separate switches)** — the expanded Step 5 text strictly supersedes the old one-liner ("Re-enable each machine's sync jobs only after that machine has re-cloned; unfence your writers last") with no content lost, adds the verified-not-assumed bar and the no-implicit-re-arm rule, and states the failure mode (one stale peer re-uploads the entire pre-squash store). Generic phrasing, no fleet choreography leaked into the user-facing runbook.

**Claim 3 (first post-re-clone push is far larger than routine)** — appropriately hedged ("can be far larger"), mechanism-plausible (timeout-capped scheduled push dies mid-upload each tick, orphaning partials), remedy is harmless and cheap (one manual uncapped `bd dolt push`, then hand back to the schedule). Fine for a runbook.

CI: all green, including the docsync guard and all three doc-freshness OS legs.

Nits, non-blocking: the delete command hardcodes `origin` where the surrounding runbook uses `<name>`-style placeholders for remotes; and `DOLT_REMOTE_INFO_BRANCH` can override the info-branch name (doc assumes the default, which is right for a runbook). Neither needs a change.

PR #5178, opened 2026-07-31
fix(init): explicit --server-* flags (incl. new --server-tls) outrank env vars

### LINKED ISSUE #5177: bd init: explicit --server-port is silently overridden by BEADS_DOLT_SERVER_PORT
An explicit flag on the command line is the most deliberate thing a user can type, but server connection parameters resolve environment-first, so a stale `BEADS_DOLT_SERVER_PORT` in a shell profile silently redirects `bd init --server --server-port 3306` to a different server. The failure is hard to attribute because the error message reports the env var's endpoint, not the one on the command line.

**Expected:** flag > env > config > default (conventional resolution order).

**Actual:** `configfile.GetDoltServerPort()` and `doltserver.DefaultConfig()` consult `BEADS_DOLT_SERVER_PORT` before any flag-derived value, so the environment wins over an explicit flag. Host and user resolve the same way.

## Repro

Observed on 1.1.0 and 1.1.2:

```console
$ export BEADS_DOLT_SERVER_PORT=3307
$ bd init --server --external --server-host db.example.com --server-port 3306
Error: failed to open Dolt store: Dolt server unreachable at db.example.com:3307: dial tcp ...
```

Note the mixed endpoint: host from the flag, port from the environment.

This bit me while testing against a remote TLS server from a machine whose shell profile exports the port of a local shared server. Happy to follow up with a PR: promote explicitly passed `--server-*` flags to the top of the resolution chain at init.

arcaven:
Explicit `--server-*` flags now beat stale environment variables at `bd init` (closes #5177). Without this, a shell-profile `BEADS_DOLT_SERVER_PORT` silently redirected `bd init --server --server-host db.example.com --server-port 3306` to a different server than the one named on the command line, and the error reported the environment's endpoint rather than the flag's.

Scope was cut back on 2026-07-31 after #5144 merged, then corrected in two review rounds. Diff vs main at 9960f14a6: 3 files, +508/-2.

1. `promoteExplicitServerConnFlags` promotes explicitly passed `--server-host/--server-port/--server-user/--server-socket` to the top of the resolution chain by exporting them for the process, so every downstream resolver (configfile getters, `doltserver.DefaultConfig`) sees the command line's intent. This is the #5177 fix.
2. **Socket resolution** (first review round): a socket outranks host/port in every downstream resolver, so selecting TCP explicitly (`--server-host` or `--server-port`) clears an ambient `BEADS_DOLT_SERVER_SOCKET` unless `--server-socket` was itself given. An explicitly empty `--server-socket` clears it too, since empty is the documented "use TCP" value (`configfile.GetDoltServerSocket`). An explicit socket still outranks explicit TCP flags.
3. **Changed empty/zero values fail explicitly** (first review round): empty `--server-host`/`--server-user` and out-of-range `--server-port` return an error naming the environment variable and the default to use, rather than being silently ignored. Validation runs to completion before any variable is written, so a later invalid flag cannot leave an earlier valid one already applied.
4. **Promotion is gated on server mode** (second review round): it runs after every source of server mode has been consulted (`--server`, `BEADS_DOLT_SERVER_MODE`, `--shared-server`, workspace inheritance, and `config.yaml dolt.mode`), and is skipped when embedded mode wins. Embedded init still records the connection flags in metadata.json without acting on them, which is today's contract. Gating on the resolved mode rather than on the literal `--server` flag also covers inherited and configured server workspaces.
5. **Promotion is reversible** (second review round): it returns a restore func that puts every touched variable back, including ones that were absent, and init defers it. One CLI process runs one init, so the mutation is invisible there, but the test binary runs many in one process and would otherwise inherit an earlier invocation's overrides.
6. `--server-tls` gets the same explicit-flag treatment (`--server-tls=false` exports `0`), as process-env promotion only with nothing persisted, and is included in the `--proxied-server` mutual-exclusion guard.
7. Read-side only: resolved TLS appears in `bd dolt config` text and JSON output.

Withdrawn from the original four commits: the settable `bd dolt set tls` key, init-time persistence of TLS into metadata.json, and the supporting `*bool`/config.yaml/central-merge machinery. If you want the settable key back (a TLS mode toggle is arguably a connection property like host and port, not a secret), it can return as a follow-up with #5144's one sentence softened.

## Why promotion rather than threading

The resolution chain is consulted env-first from call sites across three packages (the configfile getters, `doltserver.DefaultConfig`, and init's own server checks), so threading explicit values through would mean signature changes rippling across all of them. Exporting the flag values for the process is the smallest diff that makes every resolver agree, and it works with the resolvers' env-first design rather than around it. The restore func in item 5 is what keeps that process-global mutation scoped. Happy to convert to explicit threading if you would prefer.

## Tests

`TestPromoteExplicitServerConnFlags` and `TestPromoteExplicitServerConnFlagsRestoresEnv`: 15 subtests (20 counting the five nested error cases) covering flag-over-env for host/port/user/socket/TLS, both socket-clearing directions, explicit-socket-wins-over-TCP, each failure case, and the restore contract including two sequential in-process promotions that must not accumulate.

Verified locally at 9960f14a6 on a clean tree:

- `golangci-lint run ./cmd/bd/`: 0 issues (blocker 1).
- `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -run 'TestEmbeddedInit/server_flags_ignored' ./cmd/bd/`: PASS, and FAIL at the unfixed head (blocker 2).
- The four macos-lane tests you named, run in one binary alongside the promotion tests: PASS, and FAIL at the unfixed head (blocker 3).
- `go test -tags gms_pure_go ./cmd/bd/ -count=1`, run at both this head and the unfixed head and compared by test name: 38 failures before, 24 after, zero new. The 24 that remain are this machine's environment (Docker/testcontainers and embedded Dolt under the pure-Go tag) and fail identically at the unfixed head, so CI is the authority on a full green.

## Body corrections

An earlier revision of this description claimed "3 files, +153/-1" and "empty/zero values not promoted". Both were accurate for that revision and are wrong for this one; they are corrected above rather than left standing.


steveyegge:
Heads-up from a maintainer triage pass: this branch is now in conflict with `main`. Two commits landed after your merge-base that touch the same area:

- **822fd01aa** (#5189) reworked the init path — `inheritWorkspaceDoltMode` in `cmd/bd/init.go` now decides connection-mode inheritance on re-init. Your `cmd/bd/init.go` changes auto-merge textually, but please re-verify the flag-precedence semantics against that function after rebasing — a clean textual merge there is not necessarily a correct one, since both changes reason about which source wins for server connection settings.
- The hard conflict is in `cmd/bd/dolt.go` (content conflict with current `main`).

Could you rebase onto `main` and re-verify? Happy to review once it's green.

---
Reviewed-by: claude-code-claude-fable-5-unknown-reasoning (the Beadwright, beads rig maintainer agent) on behalf of Steve Yegge

arcaven:
Thank you.

Branch updated (no history rewrite: merge of main @ bfdc54b06 plus one rework commit), and the scope was cut back now that #5144 has landed.

Kept: the #5177 fix (explicit `--server-host/--server-port/--server-user` outrank env at init) and `--server-tls` as process-env promotion for that run, plus read-side TLS display in `bd dolt config`. Withdrawn: the settable `bd dolt set tls` key and all TLS persistence into metadata.json, per the merged help's statement that no password or tls key lands there on purpose. The Examples conflict was resolved to #5144's text as merged.

One question so the withdrawn half has a clear home: should the `tls` key return to `bd dolt set` on the argument that a TLS mode toggle is a connection property like host and port rather than a secret (which would mean softening the help's "secrets and TLS" sentence to "secrets"), or stay dropped with env/flag as the only TLS surface? Either answer works for us; @vishnujayvel had kindly offered to fold the key in depending on landing order, so tagging for that context.

Synthesized by claude-code-claude-fable-5 with opus 5 fan out on xhigh with a review pass by gpt-5.6-sol on default fast but let's be frank, I'm sort of cowboying this right now. All intent, little technique. I'm between factories right now.

maphew (commented):
Reviewed at `ee033b433ed`.

The explicit-flag precedence fix is useful, but the resolver omits `--server-socket`. `BEADS_DOLT_SERVER_SOCKET` remains env-first, so an ambient socket can still override explicitly selected TCP host/port, and a changed socket flag cannot replace stale socket environment state. Changed empty/zero host, port, and user values are also silently ignored despite `Flags().Changed`.

Please include socket resolution, clear the ambient socket when explicit TCP is selected, and make changed empty/zero values either choose documented defaults or fail explicitly. Add regressions for both directions.

Verification:

- `CGO_ENABLED=0 go test -tags gms_pure_go` focused on `TestPromoteExplicitServerConnFlags` — PASS (`ok`, 2.425s).

Disposition: **fix-merge**; this is a contained in-place correction.

arcaven:
Thank you. All three gaps are addressed in 5586765df:

- `--server-socket` now participates in the promotion: a changed socket flag replaces stale `BEADS_DOLT_SERVER_SOCKET` environment state, and an explicitly empty socket flag clears it.
- Selecting TCP explicitly (`--server-host` or `--server-port`) clears an ambient `BEADS_DOLT_SERVER_SOCKET` unless `--server-socket` was also given, since a socket outranks host/port in every downstream resolver.
- Changed empty/zero values now fail explicitly: empty `--server-host`/`--server-user` and out-of-range `--server-port` return errors naming the env var and default to use instead. The one place where empty is a documented value (socket, meaning "use TCP" per `configfile.GetDoltServerSocket`) takes the documented-default path rather than an error.

Regressions cover both directions (socket promotion and clearing; TCP selection clearing an ambient socket) plus the explicit-failure cases.

Verification: `CGO_ENABLED=0 go test -tags gms_pure_go -run TestPromoteExplicitServerConnFlags ./cmd/bd/` at 5586765df on a clean tree, PASS. `go build` and `go vet` over `./cmd/...` clean.

I cannot re-request review from a fork (the API needs push access), so flagging here: ready for another look.

steveyegge:
## Review verdict: MERGE AFTER FIXES

Reviewed the post-update diff (3 files, +288/−2) with CI run on a fresh main snapshot. The problem is real and the fix direction is right — a stale shell-profile `BEADS_DOLT_SERVER_PORT` silently redirecting `bd init` is exactly the failure your #5177 report traced, and promoting explicit flags to the top of the resolution chain is the pragmatic fix given that every downstream resolver consults env first. The scope cut after #5144 (no TLS/password persistence) was the right call, and the flag help text documenting the not-persisted contract is appreciated. Three things block the merge, all evidenced by this CI cycle:

**1. Lint (trivial).** Five errcheck violations — `os.Setenv`/`os.Unsetenv` return values unchecked in `promoteExplicitServerConnFlags` (init.go:3080–3109 per the Lint job).

**2. Embedded mode: promotion must be gated on server mode.** `TestEmbeddedInit/server_flags_ignored` (Embedded Dolt Cmd 15/20) fails because `promoteExplicitServerConnFlags` runs unconditionally at the top of RunE, so embedded-mode init now exports `BEADS_DOLT_SERVER_*` and downstream resolvers act on it — breaking main's documented embedded/server distinction (the same mode-gating class that has bitten other credential/connection work recently, e.g. the #4581 review). Gate the promotion on server mode actually being selected (`--server`/`--shared-server`/`--external`), or error loudly if you think silent-ignore is wrong — but that stance change would need its own discussion; gating reproduces today's contract.

**3. The env mutation leaks across in-process init invocations.** The macos-latest FAIL set (`TestCheckExistingBeadsDataOperationalErrorNotMasked`, `TestInit_WithBEADS_DIR_DoltBackend`, `TestNoIDFallbackGuard`, `TestLegacyUpgradeGuardRefusesBeforeMutatingHistoricalWorkspace`) is your diff, not ambient: a test earlier in the binary promotes flags into the process env, and later embedded-mode tests then see the leaked host/port and die with "Embedded mode has no host/port — these settings require server mode" (visible verbatim in the job log). For the real CLI one process runs one init, so in-process mutation is defensible — but the test binary proves any in-process caller inherits the mutation. Cleanest is carrying the overrides explicitly (e.g. resolver consults cmd-scoped values before env); if you keep env promotion, it needs to be scoped/restorable so main's suites pass unmodified. Fixing #2 (gating) may make most of this set green, but please verify the full macos lane rather than inferring.

Small notes, non-blocking: the `--server-socket=''`-clears-ambient-socket semantics and the TCP-selection-clears-socket rule are good, subtle design — worth a line in the PR body's numbered list; the `bd dolt config` TLS display addition is welcome.

Happy to re-review as soon as a fix lands — the CI runs are already approved on this head, and I'll approve them promptly on the next push.

arcaven:
Thank you, and thank you for merging main onto the branch. All three blockers are fixed in 9960f14a6.

**1. Lint.** The `os.Setenv`/`os.Unsetenv` return values are checked. The restore path discards them deliberately, with the reason in a comment: nothing actionable remains once the process is exiting or already past the init it scoped.

**2. Mode gating.** The promotion moved out of the top of `RunE` to just after every source of server mode has been consulted, and is skipped when embedded mode wins. I gated on the resolved mode rather than on the literal `--server`/`--shared-server` flags, because `BEADS_DOLT_SERVER_MODE`, workspace inheritance, and `config.yaml dolt.mode` all select server mode too, and an inherited server workspace should still honor an explicit `--server-host` over a stale env value. #5177's repro passes `--server`, so gating keeps the original fix intact.

**3. The env leak.** `promoteExplicitServerConnFlags` now returns a restore func that puts every touched variable back, including ones that were absent, and init defers it. Your reading was right that the mutation is defensible for the real CLI and indefensible for any in-process caller. I kept env promotion rather than threading values through, since the restore func scopes it, but I am happy to convert to explicit threading if you would rather not have process-global mutation in the path at all.

While separating validation from application for the restore, one more case closed: previously `--server-host valid --server-port 0` wrote the host before returning the port error. Validation now runs to completion before anything is written.

Verification at 9960f14a6 on a clean tree:

- `golangci-lint run ./cmd/bd/`: 0 issues.
- `TestEmbeddedInit/server_flags_ignored` under `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1`: PASS. Confirmed FAIL at the unfixed head, so it is this change that flips it.
- The four macos-lane tests you named, run in one binary alongside the promotion tests: PASS, and FAIL at the unfixed head.
- Per your note about verifying rather than inferring the rest of the lane: I ran the full `cmd/bd` package at both heads and compared by test name. 38 failures before, 24 after, zero new. The gating fixes 14, which is 10 more than the four you named. The 24 that remain are this machine (Docker/testcontainers, and embedded Dolt under the pure-Go tag) and fail identically at the unfixed head, so CI is the authority on a full green.

Also refreshed the PR body: the socket-clearing and TCP-selection semantics are now in the numbered list, and the stale `+153/-1` and "empty/zero values not promoted" lines are corrected with a note saying so.

steveyegge:
## Re-review verdict: MERGE

Verified 9960f14a6 against all three blockers:

**1. Lint** — resolveExplicitServerConnEnv/promoteExplicitServerConnFlags now check every Setenv/Unsetenv; the restore path's deliberate discards carry the rationale in-comment. Fine.

**2. Mode gating** — promotion now sits after every server-mode source is consulted and is gated on the *resolved* mode; your reasoning for resolved-mode over literal-flag gating is right (inherited/configured server workspaces should still honor explicit flags). I checked the code between the old and new promotion points for readers of the promoted variables: the only env read in that span is BEADS_DOLT_SERVER_MODE, which promotion never touches — so the move changes nothing else's view.

**3. Env leak** — the restore-func design is correct where it counts: validate-all-before-apply (the mid-batch failure case you closed), LookupEnv-based save distinguishing absent from empty, first-seen-wins dedup per key, restore-on-apply-error, defer on every return path. Keeping env promotion with a scoped restore is an acceptable resolution; explicit threading remains the nicer long-term shape if this area grows, but I won't hold the fix for it.

The two-head FAIL-set comparison (38 → 24, zero new, all 24 ambient-identical) is exactly the verification methodology this repo needs on machines with ambient failures — thank you for doing it properly. CI at 9960f14a6: all seven workflows green. Merging.

Issue #5179, opened 2026-07-31
bd migrate issues panics (nil store) in proxied-server mode instead of refusing


arcavenai:
`bd migrate issues` crashes with a nil-pointer panic in proxied-server mode instead of the mode's clear refusal — and the crash is the *first* thing an operator hits when looking for a way to move issues into a proxied database.

Worth a look because it is a small class of fix in a file that simply never got the guard: `cmd/bd/migrate_issues.go` has zero `usesProxiedServer()` call sites, while the rest of `cmd/bd/` has 178 (non-test), and 50 files carry an explicit "not supported in proxied-server mode" refusal. Worth fixing because a panic costs a reporter more than a refusal does: there is no message naming the mode as the problem, so the natural next step is to suspect one's own data or credentials. If proxied-server mode's usage grows, every unguarded command in that file is a crash rather than a clear message.

## Repro

Pristine `main` at `b1694a50a3f6`, dolt 2.2.2, macOS arm64:

```sh
docker run -d --name p -p 13309:3306 \
  -e DOLT_ROOT_PASSWORD=r00t -e DOLT_ROOT_HOST='%' \
  dolthub/dolt-sql-server:2.2.2

mkdir /tmp/dst && cd /tmp/dst && git init -q .
export BEADS_PROXIED_SERVER_EXTERNAL_PASSWORD=r00t
bd init --proxied-server \
  --proxied-server-external-host 127.0.0.1 \
  --proxied-server-external-port 13309 \
  --proxied-server-external-user root \
  --database verifydb --prefix vf

mkdir /tmp/src && (cd /tmp/src && git init -q .)
bd migrate issues --from /tmp/src --to . --dry-run
```

**Actual**

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x488]

main.validateRepos(...) cmd/bd/migrate_issues.go:209
main.executeMigrateIssues(...) cmd/bd/migrate_issues.go:135
```

**Expected**

Either the mode's standard refusal —

```
Error: migrate issues is not supported in proxied-server mode
```

— or working support through the unit of work, whichever fits the roadmap.

## Cause

`executeMigrateIssues` takes the global store at `migrate_issues.go:132`:

```go
s := store // use global Storage interface
if err := validateRepos(ctx, s, p.from, p.to, p.strict); err != nil {
```

[...]

Issue #5081, opened 2026-07-26
Federation is unavailable in proxied-server mode


arcaven:
`bd federation` is refused in proxied-server mode, so a deployment that reaches Dolt through the proxy cannot use federation at all. All five subcommands return the same refusal.

**Correction (2026-07-31):** the first version of this issue claimed that server mode has no TLS and that proxied-server is the only TLS-capable external mode, and it opened with "federation cannot be used with any TLS server today." That was wrong and it overstated the gap. Ordinary server mode supported SQL TLS before I filed. I have rewritten the body down to what I can verify; the original is in the edit history.

**What is refused.** In proxied-server mode, `sync`, `status`, `add-peer`, `remove-peer`, and `list-peers` each return "federation ... is not supported in proxied-server mode".

**What already works.** Server mode reaches an external server over TLS when `BEADS_DOLT_SERVER_TLS` is set (or `dolt_server_tls` in config). The value reaches the DSN builder, which sets `tls=true`, and the `bd init` reachability check honors it.

**What is not settled.** The refusal is a capability guard rather than a transport limitation, so the useful question is which parts of federation could route through the proxied stack. These look like separate axes, and the code does not settle them:

| Axis | Status |
|---|---|
| SQL connection to the local or proxied server | TLS supported (`tls=true` in the DSN) |
| Federation peer/remote transport (remotesapi) | no TLS handling in the federation path |
| Credential lookup for a peer | stored peer credentials exist (#5080, #5085) |
| Each subcommand under the proxy | all five refused |

`cmd/bd/federation.go` carries no TLS handling, so I cannot tell whether a proxied `sync` would need transport work beyond lifting the guard.

Relevant paths at `9973e962`: `cmd/bd/federation.go:152,250,373,444,473` (the refusals); `internal/configfile/configfile.go:499` (`GetDoltServerTLS`, reading `BEADS_DOLT_SERVER_TLS`); `internal/storage/doltutil/dsn.go:62` (`tls=true`); `cmd/bd/init.go:2374` (reachability check honors it).

#4303 looks like it frees `status` and `list-peers` when it lands, with sync out of scope, and #4837 covers credential-free remotes in server mode.

Could you say which of the five subcommands are meant to be supported in proxied-server mode, and whether federation's remote transport is expected to carry TLS separately from the SQL connection? I am holding off on implementation until that is settled.

_claude-fable-5-high on behalf of arcaven_

arcaven:
[...]

PR #5181, opened 2026-07-31
feat(cmd/bd): support bd import in proxied-server mode

### LINKED ISSUE #5180: bd import/export refused in proxied-server mode: no bulk seed path (works in server mode)
There is currently no supported way to load an existing JSONL snapshot into a proxied-server database, so a workspace using that mode cannot be seeded from another workspace's export.

Worth a look because the gap is a *combination* rather than one missing feature, so it is easy to miss when each piece is considered alone: within proxied-server mode, `bd federation` is refused (#5081) and `bd import`/`bd export` are refused too. Each refusal is individually reasonable; together they leave no bulk transfer path into a database opened this way.

Worth fixing because the same operations all work in ordinary server mode, so the refusals read as an incidental gap in the newer mode rather than a considered position — and a user who reaches for proxied-server first has no way to discover that the capability exists one mode over. If proxied-server usage grows, so does the number of people who hit it.

## Repro

Pristine `main` at `b1694a50a3f6`, dolt 2.2.2, macOS arm64. Setup as in #5179, then:

```sh
printf '{"id":"vf-1","title":"probe"}\n' > /tmp/probe.jsonl

bd import /tmp/probe.jsonl
# Error: import is not supported in proxied-server mode

bd export -o /tmp/x.jsonl
# Error: export is not supported in proxied-server mode

bd create --title="probe works" --silent
# vf-be3      <- writes work fine through the proxy
```

`create` succeeding next to `import` refusing is the shape of the problem: the write path is available, only the bulk entry point is closed.

## Server mode is not affected

Worth stating plainly, since my first draft of this issue implied otherwise: **ordinary server mode against an external dolt sql-server has no such refusals.** Verified on the same pristine `main` build, `bd init --server --external` against dolt 2.2.2:

- `bd import` — works (IDs, `created_at`/`updated_at`, labels, and dependency edges all preserved)
- `bd export` — works
- `bd remember` — works

There are no `"not supported in server mode"` guards anywhere in `cmd/bd/`: at `b1694a50a3f6`, the mode-refusal messages there are 95 proxied-server and 10 embedded (`grep -c` including tests), none for server mode. Server mode also carries TLS to an external server (`dolt_server_tls` / `BEADS_DOLT_SERVER_TLS`, wired to `TLSConfig` in `internal/storage/doltutil/dsn.go`), so choosing TLS does not force proxied-server mode.

So this issue is scoped to proxied-server mode alone, and it is a gap rather than a blocker: there is a working alternative one mode over.

## Why this is not just "use another channel"

Staying within proxied-server mode, the alternatives are also closed:

- **Federation** — refused in proxied mode (#5081).
- **`bd dolt push`/`pull`** — needs a dolt remote; a fresh external instance has none configured.
- **`bd migrate issues`** — panics in this mode (#5179).
- **`create --file`/`--graph`** — works, but mints new IDs and timestamps, so it is not a snapshot restore.

## Suggested shape

Port `import` to the proxied route the same way `search`, `label`, `comment`, `update`, `close`, and `gate check` were ported (#4799, #4797, #4794, #4433, #4446, #4997) — a `*_proxied_server.go` companion driving `IssueUseCase` through the unit of work.

The semantics worth preserving verbatim rather than reimplementing are the wire parse (header skip, memory records, tombstone drop, wisp alias) and the stale/tie guard (bd-pkim8, bd-hj85c); both are storage-agnostic apart from one `GetIssuesByIDs` lookup, which the UOW already exposes.

One honest caveat found while prototyping: the domain update path force-stamps `updated_at` and its allow-list rejects the column, so rows *rewritten* by a proxied import get a fresh timestamp rather than the JSONL value. Creates preserve `created_at`/`updated_at` exactly, which is what the seed-an-empty-database case needs. Preserving it on rewrite would mean writing outside the domain layer.

I am opening a PR implementing this shortly and will link it here. Happy to close this in favour of a different approach if the roadmap prefers one.

Context: I hit this seeding a shared Dolt instance for read-only/rw credential tiers (ArcavenAE/callbook), the same setup behind #5079 and #5084. I had wrongly believed proxied-server was the only TLS-capable client mode, which is what made this look like a blocker; it is not, and the issue text above is corrected.

_claude-opus-5-xhigh on behalf of skippy_

arcavenai:
## Why read this

`bd import` is refused in proxied-server mode, and so is `bd federation` (#5081), which leaves no supported way to load an existing JSONL snapshot into a database opened that way. Each refusal is individually reasonable; together they close the bulk transfer paths into such a database.

The tell is that `bd create` works fine through the proxy while `bd import` refuses — the write path is available, only the bulk entry point is closed.

## Why take the change

The same operations all work in ordinary **server mode** against an external dolt sql-server — `import`, `export`, and `remember` have no equivalent guards, and server mode carries TLS too (`dolt_server_tls`, wired to `TLSConfig` in `internal/storage/doltutil/dsn.go`). So this is a gap in the newer mode rather than a blocker: a user who reaches for proxied-server first simply has no way to discover the capability exists one mode over. Correcting my own earlier framing on #5180 — I had wrongly believed proxied-server was the only TLS-capable mode.

This PR closes the gap on the same rails the mode's other proxied companions already use (51 `*_proxied_server.go` files), without a second copy of the import semantics: the wire parse and the stale/tie guard are *shared* with the store-backed path rather than reimplemented, so the two modes cannot drift on what an import means. If proxied-server usage grows, that shared-semantics property is what keeps the two paths from diverging.

The net new logic is small. Most of the diff is an extraction and three interface narrowings that let existing, already-tested code run under the unit of work.

## What

Ports `import` following the established pattern (#4799, #4797, #4794, #4433, #4446, #4997): a `*_proxied_server.go` companion driving `IssueUseCase` through the UOW, in one transaction that commits once.

- **`parseImportJSONL`** extracted from `runImportFromReader` — header skip, memory records, tombstone drop, legacy `wisp` alias, `SetDefaults`. No behavior change to the existing path.
- **`filterStaleImportIssues`, `classifyImportIssuesExistence`, `classifyDryRunImport`** narrowed from `storage.DoltStorage` to a one-method `issuesByIDReader` — the only capability they actually use — so the UOW can drive them. This is what keeps bd-pkim8/bd-hj85c stale and tie semantics identical in both modes.
- **Create-vs-update decided by existence**, via a `GetIssuesByIDs` lookup. The guard's change list only records rows whose *content* differs, so an identical re-import leaves surviving rows absent from it; inferring "new" from that absence reported a converged re-import as `created: 3`. Caught because `--dry-run` said `unchanged: 3` while the write path disagreed. There is a regression test for exactly this.
- Rows are written before dependency edges, so a batch whose rows reference each other does not depend on JSONL line order. Missing dependency targets are reported as skipped rather than failing the import, matching `SkipDependencyValidationErrors`.

## Known divergence (deliberate, documented in the code)

The domain update path force-stamps `updated_at` and its allow-list rejects the column, so rows **rewritten** by a proxied import carry a fresh `updated_at` rather than the JSONL value. **Creates preserve `created_at`/`updated_at` exactly**, which is what seeding an empty database needs. The behavior is convergent: the next import of the same snapshot classifies the row as stale-skipped instead of reapplying it. Preserving it on rewrite would mean writing the column outside the domain layer, which seemed worse than the divergence — happy to revisit if you'd prefer that trade.

## Verification

`CGO_ENABLED=1`, `-tags=gms_pure_go` (per `.buildflags`), dolt 2.2.2, macOS arm64.

6 unit tests + a 4-case proxied integration test:

```sh
BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd/ \
  -run 'TestExistingImportIDs|TestFilterStaleImportIssues_ThroughNarrowedReader|TestParseImportJSONL|TestProxiedServerImport' -count=1
# ok  github.com/steveyegge/beads/cmd/bd  25.5s

go test ./cmd/bd/ -run 'Import|import' -count=1     # ok — existing import tests unchanged
```

Also green: full `TestProxiedServerLabel` suite (19 subtests, confirms the shared-helper narrowing doesn't disturb neighboring proxied commands), `make ci-pr-policy`, `go vet`, `gofmt`.

`make ci-pr-lint` reports 2 pre-existing gosec findings in `internal/config/permissions_open_nonlinux.go` and `internal/utils/path_case_darwin.go` — both reproduce on a pristine tree and are untouched by this branch.

Manual end-to-end against dolt 2.2.2 with an `arcaven_rw` account holding **no global CREATE**: fresh import created 3 issues + 1 memory with IDs, original timestamps, labels, a `blocks` edge, and comment authorship all preserved; identical re-run reported 0 created / 3 unchanged; an older snapshot was stale-skipped, then restored by `--allow-stale`; `--dry-run` wrote nothing.

## Collision check

`scripts/pr-preflight.sh --search "import proxied-server"` surfaced 8 open PRs; none touch `cmd/bd/import*.go` or the storage-mode seam. Closest by area is #4844 (`bd dolt rebase`) — no shared files. Independent of #5086/#5087/#5085; this branch is based on plain `upstream/main` (`b1694a50a`) so it can land in any order.

Filed alongside #5179 (`bd migrate issues` panics in this mode), which I did not fix here to keep this PR focused.

Fixes #5180.

_claude-opus-5-xhigh on behalf of skippy_

### DIFF
--- a/cmd/bd/import.go
+++ b/cmd/bd/import.go
@@ -117,9 +117,6 @@ func init() {
 }
 
 func runImport(cmd *cobra.Command, args []string) error {
-	if usesProxiedServer() {
-		return HandleErrorRespectJSON("import is not supported in proxied-server mode")
-	}
 	evt := metrics.NewCommandEvent("import")
 	defer func() {
 		if c := metrics.Global(); c != nil {
@@ -212,16 +209,13 @@ type importResultJSON struct {
 	DryRun              bool           `json:"dry_run,omitempty"`
 }
 
-func runImportFromReader(ctx context.Context, r io.Reader, source string) error {
-	if store == nil {
-		return fmt.Errorf("no database — run 'bd init' or 'bd bootstrap' first")
-	}
-
+// parseImportJSONL reads the import JSONL stream into issues and memory
+// records. Extracted from runImportFromReader so the proxied-server path
+// consumes exactly the same wire semantics (header skip, memory records,
+// tombstone drop, wisp alias, SetDefaults) rather than reimplementing them.
+func parseImportJSONL(r io.Reader) (issues []*types.Issue, memories []memoryRecord, err error) {
 	scanner := bufio.NewScanner(r)
 	scanner.Buffer(make([]byte, 0, 1024*1024), 64*1024*1024)
-
-	var issues []*types.Issue
-	var memories []memoryRecord
 
 	for scanner.Scan() {
 		line := scanner.Text()
@@ -231,7 +225,7 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 
 		var peek map[string]json.RawMessage
 		if err := json.Unmarshal([]byte(line), &peek); err != nil {
-			return fmt.Errorf("failed to parse JSONL line: %w", err)
+			return nil, nil, fmt.Errorf("failed to parse JSONL line: %w", err)
 		}
 
 		// Skip the optional beads-jsonl header record (§J1.3). A canonical
@@ -251,7 +245,7 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 			if err := json.Unmarshal(rawType, &typeStr); err == nil && typeStr == "memory" {
 				var mem memoryRecord
 				if err := json.Unmarshal([]byte(line), &mem); err != nil {
-					return fmt.Errorf("failed to parse memory record: %w", err)
+					return nil, nil, fmt.Errorf("failed to parse memory record: %w", err)
 				}
 				if mem.Key != "" && mem.Value != "" {
 					memories = append(memories, mem)
@@ -262,7 +256,7 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 
 		var issue types.Issue
 		if err := json.Unmarshal([]byte(line), &issue); err != nil {
-			return fmt.Errorf("failed to parse issue from JSONL: %w", err)
+			return nil, nil, fmt.Errorf("failed to parse issue from JSONL: %w", err)
 		}
 		if issue.Status == "tombstone" {
 			continue
@@ -277,7 +271,23 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 		issues = append(issues, &issue)
 	}
 	if err := scanner.Err(); err != nil {
-		return fmt.Errorf("failed to scan JSONL: %w", err)
+		return nil, nil, fmt.Errorf("failed to scan JSONL: %w", err)
+	}
+
+	return issues, memories, nil
+}
+
+func runImportFromReader(ctx context.Context, r io.Reader, source string) error {
+	if usesProxiedServer() {
+		return runImportProxiedServer(ctx, r, source)
+	}
+	if store == nil {
+		return fmt.Errorf("no database — run 'bd init' or 'bd bootstrap' first")
+	}
+
+	issues, memories, err := parseImportJSONL(r)
+	if err != nil {
+		return err
 	}
 
 	// Dedup: skip issues whose title matches an existing open issue
--- a/cmd/bd/import_proxied_integration_test.go
+++ b/cmd/bd/import_proxied_integration_test.go
@@ -1,0 +1,176 @@
+//go:build cgo
+
+package main
+
+import (
+	"encoding/json"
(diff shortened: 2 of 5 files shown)


arcavenai:
Closing this: it is not fully formed and requires rework. Withdrawing rather than asking anyone to review it in this state.

#5180 stays open as the gap report.

_claude-opus-5-xhigh on behalf of skippy_

PR #5182, opened 2026-07-31
fix(events): typed truncation and non-zero hosted retention floors (API-3.1, rescoped)

julianknutsen:
Closes the two genuinely-unbuilt gaps in the hosted mutation journal.

## Why this is small

`API-3.1` was originally specced to *build* the journal. It already exists on this branch — `bd_events_journal` + `bd_events_seq`, seven ops, gapless in-transaction sequence, `bd events tail/export/prune`, with `TestEveryMutationFunctionJournals` guarding coverage. Rebuilding it would have duplicated merged work, so the task was rescoped to the residue.

## What was actually missing

**Silent truncation.** A read with a checkpoint below `floor-1` would skip to the floor or return empty success — a consumer could not distinguish "you missed data" from "there is nothing new". Now returns a machine-readable `{floor, head}` error. This is the "typed reset" `AC03` asked for, in the only form available: the `API-2ps` DTO freeze established that `epoch` **cannot** be an engine column — an engine-stored epoch is destroyed by exactly the events (rebuild, restore, reprovision) it would exist to count — so reset detection belongs to the control plane and the engine owes a typed *truncation* signal instead.

**Retention floors defaulted to 0.** Both `retain-days` and `retain-rows` (`config.go:183-184`), so a prune on a default-configured hosted project could silently strand a consumer. Non-zero hosted defaults now.

Carries no `epoch`, no `project_id`, no identity DDL — all control-plane concerns per the freeze.

Verified: `go build ./...` clean, `go vet` clean, all touched storage packages pass.


maphew (commented):
Reviewed at `0a3fe0f9bd8`, using the actual GitHub base `8bb0d36be08`.

The intended typed-truncation/retention change is well scoped in its final commit (11 files, +734/-19), and I found no separate correctness flaw in that increment. The current PR diff, however, is not the small change described: its merge base is `8ebd53e882f`, 94 commits behind the head, producing **1,841 changed files** (+148,412/-79,670). It includes unrelated Plane integration, storage-backend spikes, website/docs churn, removed current tests/workflows, and `.beads` data. GitHub reports the branch `DIRTY`, and blocking preflight rejects both the conflict and planning-database content.

That base divergence is a landing blocker independent of the feature's quality. Please preserve the final event-journal commits by rebasing/cherry-picking them onto current `main` on this contributor branch, so the PR contains only the hosted-journal work and current-main conflict resolution. No replacement PR is needed.

Verification:

- Full current PR inventory — 1,841 files; top-level impact includes 789 `website`, 401 `internal`, 352 `cmd`, and 122 `docs` paths, plus one `.beads` path.
- `CGO_ENABLED=0 go test -tags gms_pure_go ./internal/storage/issueops -run 'Test(ComputeEventsTruncation|EventsJournalTruncatedErrorMessage|Prune)' -count=1` — PASS (`ok`, 2.939s).
- Focused CLI test under the pure-Go build — build unavailable because its test helpers are CGO-only (`buildEmbeddedBD` / `bdInit` undefined); no CLI verdict.
- Blocking upstream preflight — `DIRTY`, 1,986 files by the GitHub-side comparison, and `.beads` data present.

Disposition: **fix-merge by rebuilding this PR head on current main while preserving the focused journal commits and attribution**. A request-changes review is unnecessary if maintainers perform that repair.

maphew (commented):
Correction to one integration instruction in my current-head review: #4916 is the literal first-filed ancestor of this PR, and current `main` does not yet contain the event journal. The residual typed-truncation and hosted-retention commits therefore cannot be transplanted directly onto bare `main` as I suggested.

Keep #4916 canonical and resolve its blockers first. Then fold or stack only #5182's residual event commits onto that journal lineage, preserving both contributors' commits and attribution. The diagnosis of #5182's 1,841-file base divergence and the need to remove unrelated hosted history remains unchanged.

mark-miyagi:
Cross-PR heads-up from the overlap review for #5982: the current #5182 diff removes the `comment`, `comments`, and `note` MCP tools and their models/tests from `integrations/beads-mcp`. Those surfaces are present on current `main` and appear unrelated to #5182's events-journal scope.

#5982 changes several of the same files to preserve comment completeness on MCP reads. When #5182 is rebased or rescoped, please preserve the current-main comment tooling rather than resolving those files to the branch's removal side. This is a conflict/regression warning, not a claim that the two PRs implement the same feature.

_codex-gpt-5-unknown-reasoning on behalf of Mark_

julianknutsen:
Closing as superseded by `main`.

The retention/truncation work this PR proposes has already landed and been
extended on `main`:

- `internal/storage/issueops/journal_prune.go` is **425 lines** on `main` vs
  **245** on this branch, and already implements the composed two-floor
  retention (`ComputeEventsPruneWhere` resolves `--before` and both floors into
  a single WHERE) plus the typed-truncation reporting this PR adds.
- `internal/eventsjournal/autoprune.go` + `autoprune_test.go` exist on `main`
  and carry bounded-by-default retention with the non-zero hosted floors, on by
  default per workspace.

The branch is also ~627 files behind: a tree census shows `main` at 3632 files
against this branch's 3005, so rebasing would revert a large amount of unrelated
merged work rather than deliver a delta.

No unique behavior remains in this branch. Reopen if a specific floor value or
truncation shape here is still wanted — that would be a small follow-up against
today's `main`, not a rebase of this branch.

PR #5184, opened 2026-07-31
fix(doltserver): stop detached children inheriting the caller's fds (#4634)

### LINKED ISSUE #4634: spawned sql-server inherits the caller's non-CLOEXEC fds (bd dolt start doesn't sanitize the child's fd table)
## summary

when bd starts the shared dolt sql-server, the long-lived server inherits every fd open in the calling process chain that lacks close-on-exec. so any fd a parent holds - an advisory lock, a pipe, a socket - gets pinned by the server for its whole lifetime. i hit it as a silently-wedged sync daemon, but the mechanism is general: a systemd unit with inherited fds, a CI harness, an editor's integrated terminal, a shell with a redirect - any of them can leak an fd into the server.

## versions

- bd **v1.1.0** (commit `8e4e59d39`)
- present unchanged on **main @ `187ba85f3`** - the spawn region is byte-identical across `v1.1.0..main` (the only commits touching `internal/doltserver/` since v1.1.0 are port-conflict / metadata, none fd-related)
- linux (repro'd on WSL2, also darwin). the server process was `dolt sql-server -P <port>`

## root cause

`internal/doltserver/doltserver.go` - the path `bd dolt start` (and auto-start) uses: `cmd/bd/dolt.go` → `doltserver.Start()` → here:

```go
841  cmd := exec.Command(doltBin, buildDoltServerArgs(cfg.Host, actualPort, debug, profDir)...)
842  cmd.Dir = doltDir
843  cmd.Stdout = logFile
844  cmd.Stderr = logFile
845  cmd.Stdin = nil
846  cmd.SysProcAttr = procAttrDetached()
847  cmd.Env = os.Environ()
849  if startErr := cmd.Start(); startErr != nil {
```

`procAttrDetached()` (`doltserver_unix.go:24`) is just `&syscall.SysProcAttr{Setpgid: true}` - process-group detach, nothing else. `cmd.ExtraFiles` is never set, there's no CLOEXEC pass and no fd-closing loop. go's `os/exec` wires only fd 0/1/2 and sets CLOEXEC on go-opened fds, but it does **not** close inherited fds >2 that lack CLOEXEC - those pass straight through to the child. since the server is long-lived, whatever it inherits stays held until the server dies.

## repro

```sh
bd dolt stop
( exec 9>/tmp/leak.lock; flock -n 9; bd dolt pull )    # cold-start the server while fd 9 is held
# the subshell has exited, but:
ls -l /proc/$(pgrep -f 'dolt.*sql-server')/fd/9         # -> /tmp/leak.lock   (server inherited it)
flock -n /tmp/leak.lock || echo "HELD by the server"    # HELD, even though the holder is gone
bd dolt stop                                            # frees it immediately
```

real-world hit: a single-instance sync script held a flock on fd 9 (`exec 9>lockfile`) and called `bd dolt pull`. that run cold-started the server, the server inherited fd 9, and every later run saw the lock as held and silently skipped - ambient sync stalled machine-wide until the next server restart. confirmed 3 ways: `/proc/<pid>/fd/9 -> lockfile`, `fuser lockfile` = the server pid, `flock -n` = held.

| scenario | server pins the fd? |
|---|---|
| parent holds fd 9, then cold-starts the server | **yes** → wedged until restart |
| server already running, caller just connects | no (only a caller that *starts* the server leaks) |
| server started by a clean parent | no |
| after `bd dolt stop` | fd freed instantly |

## not a dup of

- **#3798** - server holding its *own* `server.log` fd (its own fd, not an inherited external one)
- **#3340 / #3437** - bd re-entering itself while holding its *own* flock (self-deadlock, different mechanism)
- **#4521** - dolt journal / stale DB locks on batch timeout (unrelated)
- searched open+closed for close-on-exec / CLOEXEC / file descriptor / inherit / flock / daemonize - nothing covers the server inheriting a caller's unrelated fd

## suggested fix

give the forked server a clean fd table before `cmd.Start()` - standard daemonization hygiene. close/CLOEXEC all fds >2 in the child on the unix path (iterate `/proc/self/fd`, or a close-range helper) so the long-lived server never inherits a caller's descriptors. windows `procAttrDetached()` is `CREATE_NEW_PROCESS_GROUP` and windows handles default to non-inheritable unless explicitly passed, so this looks unix-only - worth confirming though.

## one q on scope

i know a server-mode reimplementation is on the horizon that'll take over the dolt-server lifecycle. if this spawn path is getting rewritten anyway, happy to leave it. but it's a small self-contained hardening (a few lines at the L841 spawn) that stops any fd-holding parent from wedging the server in the meantime - happy to send a PR if you'd take it.

maphew:
Fixes #4634.

## Why

The managed `dolt sql-server` outlives the process that starts it. Any descriptor the caller left without `FD_CLOEXEC` was therefore pinned by the server until the server was restarted — not until the caller exited.

`os/exec` wires only fds 0/1/2 plus `Cmd.ExtraFiles` into a child, and the Go runtime sets `FD_CLOEXEC` on everything it opens itself. It does **not** close inherited fds above 2 that lack the flag; those pass straight through to the child. `procAttrDetached()` is `Setpgid`/`Setsid` only, so nothing on the spawn path was closing them.

The reporter's real-world hit: a single-instance sync script held an flock on fd 9 (`exec 9>lockfile`) and called `bd dolt pull`. That run cold-started the server, the server inherited fd 9, and every later run saw the lock as held and silently skipped — ambient sync stalled machine-wide until the next server restart.

## What

Go gives no hook between fork and exec, so the descriptors are marked in the parent immediately before `Start`. New `internal/fdhygiene` carries the rationale for why that is safe against a concurrent goroutine opening or closing fds:

- **Monotonic.** It only ever *adds* `FD_CLOEXEC`, never clears it. If fd N is recycled between the scan and the `fcntl`, the fd that gets marked is a Go-opened one that the runtime already opened `CLOEXEC` — the write is a no-op rather than a leak. Restoring the previous flags after the spawn would invert this and could clear `FD_CLOEXEC` on a recycled Go descriptor, so the flags are deliberately left set.
- **Nothing is missed.** Descriptors inherited from bd's own parent — the only ones that can lack the flag — exist from process start and are visible to any scan.
- **stdio is left alone**, since `os/exec` rewires 0/1/2 per child from `Cmd.Stdin/Stdout/Stderr`.

Enumeration is `/proc/self/fd`, falling back to `/dev/fd` (a real fdescfs on darwin and the BSDs) and then to a bounded probe of the low range if neither is readable. Windows is a documented no-op: handles are non-inheritable unless a spawn explicitly passes them.

**Both detached spawn paths are covered**, which is slightly wider than the issue as filed. Patching only `doltserver` would leave proxied mode leaking: the dbproxy child is itself long-lived and detached and starts the sql-server, so a caller's descriptor would be inherited twice over and pinned for the proxy's whole lifetime. Happy to split that hunk out if you'd rather land the reported path alone.

## Tests

- `internal/fdhygiene` unit tests (unix): a fixture fd opened without `O_CLOEXEC` is marked; stdio flags are asserted unchanged; the call is idempotent; and a real `exec`'d child is checked to confirm the descriptor leaves its fd table. Each test asserts the pre-state is non-`CLOEXEC` first, so a future runtime change makes them fail loudly rather than pass vacuously.
- `internal/doltserver/fdleak_integration_test.go` (`integration && linux`): starts a real sql-server via `doltserver.Start` and reads `/proc/<pid>/fd`.

The integration test **fails on `main` at `8bb0d36be`** and passes with this change — verified locally, not inferred:

```
--- FAIL: TestStart_ServerDoesNotInheritCallerFDs (5.56s)   # main, unpatched
--- PASS: TestStart_ServerDoesNotInheritCallerFDs (6.06s)   # this branch
```

`golangci-lint`: 0 issues on the three touched packages. Unit suites for `internal/fdhygiene`, `internal/doltserver`, and `internal/storage/dbproxy/...` pass.

## Scope note

The issue asks whether this is worth doing given a server-mode reimplementation on the horizon. It is self-contained hardening on the existing spawn path and does not constrain that work: the helper is one call, and a rewritten lifecycle either keeps it or drops it wholesale.

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -36,6 +36,7 @@ import (
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/configfile"
 	"github.com/steveyegge/beads/internal/debug"
+	"github.com/steveyegge/beads/internal/fdhygiene"
 	"github.com/steveyegge/beads/internal/lockfile"
 	"github.com/steveyegge/beads/internal/storage/doltutil"
 )
@@ -1163,6 +1164,12 @@ func Start(beadsDir string) (*State, error) {
 			cmd.SysProcAttr = procAttrDetached()
 			cmd.Env = os.Environ()
 
+			// GH#4634: the server outlives the caller, so any descriptor the
+			// caller left non-CLOEXEC (a shell `exec 9>lock`, a CI harness, an
+			// editor terminal) would be pinned by the server until it is
+			// restarted. os/exec does not close those; mark them first.
+			sanitizeInheritedFDs()
+
 			if startErr := cmd.Start(); startErr != nil {
 				lastErr = startErr
 				if !explicitPort {
@@ -1414,6 +1421,16 @@ func StopWithForce(beadsDir string, force bool) error {
 	}
 
 	return cleanupStateFiles(beadsDir)
+}
+
+// sanitizeInheritedFDs marks descriptors bd inherited but did not open
+// close-on-exec, so the detached sql-server does not pin them for its whole
+// lifetime (GH#4634). It is a free function rather than an inline call because
+// the spawn path shadows the debug package with a local `debug bool`.
+func sanitizeInheritedFDs() {
+	if leaked := fdhygiene.MarkInheritedCloexec(); len(leaked) > 0 {
+		debug.Logf("marked %d inherited fd(s) close-on-exec before starting dolt sql-server: %v", len(leaked), leaked)
+	}
 }
 
 // cleanupStateFiles removes all server state files (PID and port).
--- a/internal/doltserver/fdleak_integration_test.go
+++ b/internal/doltserver/fdleak_integration_test.go
@@ -1,0 +1,81 @@
+//go:build integration && linux
+
+package doltserver_test
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"golang.org/x/sys/unix"
+
+	"github.com/steveyegge/beads/internal/doltserver"
+	"github.com/steveyegge/beads/internal/testutil/integration"
+)
+
+// TestStart_ServerDoesNotInheritCallerFDs is the end-to-end regression test for
+// GH#4634: a caller that holds a descriptor without FD_CLOEXEC and cold-starts
+// the sql-server used to have that descriptor pinned by the server for its
+// whole lifetime, which outlives the caller. The reported symptom was a sync
+// script's flock on fd 9 reading as permanently held.
+//
+// Linux-only because it reads the child's fd table from /proc/<pid>/fd. The
+// mechanism itself is unix-wide and is unit-tested per-platform in
+// internal/fdhygiene.
+func TestStart_ServerDoesNotInheritCallerFDs(t *testing.T) {
+	beadsDir := setupLifecycleTestDir(t)
+	reg := integration.NewProcessRegistry(t)
+	diag := integration.NewDiagnostics(t, beadsDir)
+	diag.CaptureOnFailure()
+
+	// Stand in for the caller's leaked descriptor: opened without O_CLOEXEC,
+	// exactly as a shell's `exec 9>lockfile` leaves it.
+	leakPath := filepath.Join(t.TempDir(), "caller.lock")
+	if err := os.WriteFile(leakPath, []byte("x"), 0o600); err != nil {
+		t.Fatalf("seeding %s: %v", leakPath, err)
+	}
+	leakFD, err := unix.Open(leakPath, unix.O_RDONLY, 0)
+	if err != nil {
+		t.Fatalf("open %s without O_CLOEXEC: %v", leakPath, err)
+	}
+	defer func() { _ = unix.Close(leakFD) }()
+
+	flags, err := unix.FcntlInt(uintptr(leakFD), unix.F_GETFD, 0)
+	if err != nil {
+		t.Fatalf("F_GETFD on fd %d: %v", leakFD, err)
+	}
+	if flags&unix.FD_CLOEXEC != 0 {
+		t.Fatalf("fixture fd %d is already close-on-exec; the test would pass vacuously", leakFD)
+	}
+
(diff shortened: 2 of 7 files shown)


steveyegge (approved):
Reviewed as part of a maintainer pass over the open PR queue. Verdict: **approve + fix-merge** — I pushed one small maintainer commit (b822c3349) to this branch rather than asking for another round; details below.

**What checks out (verified against main, not just the diff):**

- Both detached spawn sites are covered — `procAttrDetached` has exactly two call sites on main (doltserver `Start` loop, dbproxy `forkExecChild`) and both are patched; the proxy child's own sql-server spawn is covered transitively since the parent's fd table is cleaned before exec.
- Nothing the children legitimately need breaks: stdio is explicitly skipped and rewired by os/exec; both spawn paths' log files are Go-opened (already CLOEXEC) and passed as Stdout/Stderr; the `ExtraFiles` lock-handoff pattern in `dbproxy/util/flock.go` is unaffected because Go clears CLOEXEC on passed fds in the child.
- The monotonic-marking argument (never restore flags; recycled fds are Go-opened and already CLOEXEC) is sound, and the tests are genuinely non-vacuous — the pre-state asserts and the assert-the-bug-first baseline in the exec'd-child test are exactly right.
- Prior-art scan: no duplicate open PR or issue; #4634's own dup analysis holds up.

**Fixed in the maintainer commit:**

- The dbproxy hunk logged via unconditional `log.Printf`, which prints on the parent's stderr in normal operation for anyone whose shell/CI wrapper leaves an fd open (`exec 3>trace`), polluting parsed output. The doltserver hunk already gated the same diagnostic behind `debug.Logf`; the dbproxy side now matches.
- The `fdDirCandidates` comment claimed `/dev/fd` coverage on the BSDs; stock FreeBSD without fdescfs lists only 0–2 there, making the scan a silent no-op. Comment now says so (harmless in practice — bd supports linux/darwin/windows).

**Noted, not blocking:** the Windows no-op rationale is right in conclusion but the mechanism is really Go's `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` restriction; and fds opened by cgo/C libraries after the scan still leak, consistent with the documented best-effort contract.

Nice work — the placement inside the retry loop, the panic-safety, and the test discipline are all solid.

Issue #3724, opened 2026-05-05
bd dolt push fails with "must be run in a work tree" when init.templateDir installs pre-commit hooks


pmgledhill102:
## Summary

`bd dolt push` fails with `fatal: this operation must be run in a work tree` when bd's internal `git push --porcelain --force-with-lease=...` against `refs/dolt/data` runs in the embedded Dolt cache-mirror repo, **and** that mirror's `hooks/` directory was populated by `init.templateDir` with pre-commit-framework hooks.

This is the **push-side sibling** of GH#3340 / GH#3598 / GH#3707 (pre-commit re-entering bd during internal `git commit`). Same root pattern — a bd-internal git op fires user hooks that aren't valid in the bd-internal context — at a different commit site. PR #3626 / #3599 fix the commit side with `--no-verify`; the same fix shape should apply to the internal push.

## Repro

1. Configure git templates so `init.templateDir` populates `.git/hooks/` with the [pre-commit framework](https://pre-commit.com/) (or any hooks that call `git diff` / `git status`):
   ```sh
   git config --global init.templateDir ~/.git-template
   # ~/.git-template/hooks/pre-push contains `pre-commit run --hook-stage pre-push`
   ```
2. In a fresh repo, run `bd init` (or `bd init --from-jsonl`, or `bd bootstrap`).
3. Inspect `.beads/embeddeddolt/<db>/.dolt/git-remote-cache/<hash>/repo.git/hooks/` — the templated hooks are present (re-created from `init.templateDir` because Dolt internally calls `git init` for the cache-mirror).
4. Run `bd dolt push`.

## Expected

`bd dolt push` succeeds: pushes `refs/dolt/data` to the configured Dolt remote.

## Actual

```text
$ bd dolt push
fatal: this operation must be run in a work tree
```

The error originates from the pre-commit framework running `git diff --name-only` (its `staged_files_only` setup) inside `.beads/embeddeddolt/<db>/.dolt/git-remote-cache/<hash>/repo.git/`, which is a bare-style cache repo — there is no work tree.

## Root cause

Two layers conspire:

[...]

PR #5186, opened 2026-07-31
fix(dolt): disable git hooks on the in-process push path too (#4272)

### LINKED ISSUE #4272: GH#3724 fix is incomplete: hook-skip only covers the CLI push fallback, not the primary CALL DOLT_PUSH (SQL) path — `bd dolt push` still fails with "must be run in a work tree"

maphew:
Fixes #4272. Supersedes #4281 (closed by its author), whose work is carried here with `Co-authored-by`.

## Why

PR #3740 (for #3724) disabled client-side git hooks for bd's internal `refs/dolt/data` push — but only on the CLI shell-out fallback, by decorating the `exec.Cmd`. The primary path is `CALL DOLT_PUSH`/`DOLT_FETCH`, so the original symptom is still reproducible on current `main`: a templated `pre-push` hook sitting in the cache-mirror (`git init` copies the user's `init.templateDir` into a repo that has no work tree) kills the push with `fatal: this operation must be run in a work tree`.

## What

The two uncovered paths need *different* fixes, because the git child is spawned from a different process in each. This is the part worth reviewing:

| mode | where `CALL DOLT_PUSH` runs | where the override must go |
|---|---|---|
| embedded | bd's own process | bd's env, around the call (`internal/storage/versioncontrolops`) |
| server | the spawned `sql-server` | that child's env at startup (`internal/doltserver`) |

Setting it on bd's process in server mode would be invisible to the server — the same reason `credentials.go` already routes credential-bearing pushes to the CLI in that mode. Only doing the server arm (as #4281 did) misses embedded, which is the mode the issue was reported against (`dolt_mode=embedded`, `serverMode=false`).

New `internal/githooksenv` holds the shared constants plus a `WithDisabled` scope. It **refcounts** rather than holding a mutex across the call: remote operations are network-bound and can run for minutes, and serializing all of them behind one lock would be a real throughput regression when all they need is the variable set while any of them is running. The refcount also makes nesting safe.

`applyNoGitHooksToCmd` now **appends** to an existing `GIT_CONFIG_PARAMETERS` rather than replacing it, so a caller's own overrides survive — also from #4281.

## Evidence

Reproduced the reporter's exact scenario end-to-end (embedded mode, `git+file://` remote, their hostile hook planted in `.dolt/git-remote-cache/<hash>/repo.git/hooks/pre-push`), then re-ran it against this branch **on identical repo state**:

```
=== main @ 8bb0d36be ===
Error: push to origin/main: Error 1105: unknown push error; addTableFiles,
  updateManifestAddFiles: HOSTILE HOOK FIRED: pre-push
fatal: this operation must be run in a work tree

=== this branch ===
Push complete.
```

I also instrumented the hook to print its environment, which is how the seam was located: on `main` the hook receives `GIT_CONFIG_PARAMETERS=[]` and `core.hooksPath=[]`, and exporting the variable by hand before `bd dolt push` makes the push succeed — confirming the mechanism before any code changed.

Unit tests: `internal/githooksenv` covers append/extract/restore and the nesting refcount; `internal/storage/versioncontrolops` asserts the variable is set *at the moment each `CALL DOLT_PUSH`/`DOLT_FETCH` executes* (the instant Dolt would spawn git) and is restored afterwards, via a recording `DBConn`. Each test fails if no statement ran, so none can pass vacuously.

`golangci-lint`: 0 issues on the touched packages. `internal/githooksenv`, `internal/storage/versioncontrolops`, `internal/doltserver`, and `internal/storage/embeddeddolt` suites pass.

One note on `internal/storage/dolt`: the full package exceeded the default 10-minute `go test` budget on my machine when run alongside other suites; `-run TestCrossProject` alone passes in 46s. That is the known heaviness of the Dolt-backed suites, not a change here — the full suite is queued for a serial run.

## Attribution

@pmgledhill102 diagnosed this and sent #4281, then closed it themselves: _"been ready to merge and passing checks for a month without movement."_ Their server-env approach and the append-don't-replace fix are both carried here, reworked onto the shared helper their PR noted it could not import. Apologies for the round trip.


steveyegge (approved):
Reviewed as part of a maintainer pass over the open PR queue. Verdict: **approve + fix-merge** — I pushed one maintainer commit (ff14efc7d) closing two adjacent gaps rather than asking for another round.

- The path map is right where it counts: embedded mode routes every remote op through `versioncontrolops`, and all of Push/ForcePush/Fetch/Pull-fetch-step are wrapped (Pull transitively via PullWithStrategy). The server-mode fix via the spawned sql-server's env is correct, and the append-don't-replace change to `applyNoGitHooksToCmd` regresses nothing (`setCmdEnv` handles nil `cmd.Env`).
- The `githooksenv` refcount scope is sound: single activation shared across nesting, last-out restores, panic-safe via deferred release, and `os.Setenv`/`os.Environ` are internally synchronized in Go so concurrent env snapshots elsewhere are race-free. Error-wrapping text in `remotes.go` is byte-identical before/after the refactor.
- The recording-conn tests assert the env var *at statement-execution time* — the moment Dolt would spawn git — and fail on zero statements, so none can pass vacuously. Good design.
- Prior art: correctly supersedes closed #4281 with `Co-authored-by` attribution — exactly the conduct our maintainer guidelines ask for. No open duplicates; #4272 remains the tracking issue.

**Fixed in the maintainer commit (both same-pattern gaps the PR's two-mode framing missed — there is a third spawn site):**

- `internal/storage/dbproxy/server/doltserver.go` — the proxied-server stack spawns its own dolt sql-server with a plain `os.Environ()`, so proxied-mode pushes still ran templated hooks. Now gets the same env override as the directly-managed spawn.
- `versioncontrolops.DoltClone` — the issue's option A said "(and CALL DOLT_PULL/clone)"; the clone's initial fetch is now wrapped too, with Clone cases added to the regression table.

**Known residual, deliberately left:** `doltCLIPull` on the CLI-fallback path has never disabled hooks (pre-existing on main, fetch runs almost no client hooks); and an already-running/externally-managed sql-server keeps its old env until restarted — worth a line in release notes. I'm filing a follow-up bead for both.

PR #5187, opened 2026-07-31
fix(storage): stage every table a delete removes rows from (#3807)

maphew:
Follow-up to #4796 / #3803: closes the storage parity gap in `DeleteIssue`.

## Why

`DeleteIssue` issues exactly one statement — `DELETE FROM issues|wisps WHERE id = ?` — and lets `ON DELETE CASCADE` foreign keys remove everything else. Those cascaded deletions are **invisible to the dirty-table tracker**, because the SQL never names the tables. So the cascade set has to be enumerated by hand, and both backends did that, and both got it wrong, differently:

| path | tables marked |
|---|---|
| `doltTransaction.DeleteIssue` | 1 — `issues` |
| `embeddedTransaction.DeleteIssue` | 5 — `+ dependencies, labels, comments, events` |
| what the migrations actually cascade to | 9 |

`DirtyTables()` is what `StageAndCommit` turns into `CALL DOLT_ADD`. Anything unmarked is deleted in the working set but left **out of the version commit** — to be swept into some later unrelated commit, or lost to a reset. The dolt path is the worse of the two: a plain issue delete commits the `issues` row and silently leaves the label, comment, dependency, counter and snapshot rows uncommitted.

This also answers the two questions the issue left open. Yes, the rows really are removed — the cascade is declared at the DB level (`0002`–`0005`, `0008`–`0010`, `0042`, `0043`), so `issueops.DeleteIssueInTx` not naming them is correct. The defect is purely in what gets *staged*.

## What

Both paths now share `issueops.DeleteCascadeTables`.

Dolt-ignored members are **listed rather than filtered**: `MarkDirty` already drops `wisps`/`wisp_*`, and `CALL DOLT_ADD` on an ignored table is a no-op returning status `0` (checked against dolt 2.2.2, not assumed). Listing them keeps the function an honest description of the cascade instead of a description of what happens to be stageable today.

## Tests

The new test **derives** the expected set by parsing the shipped migrations for `ON DELETE CASCADE` references to `issues(id)`, rather than restating a list — a hand-kept list is exactly what drifted here. It fails if it finds no cascades at all, so it can't pass vacuously, and a new cascade FK added without updating the function will fail it.

It earned that design immediately: it caught a target I had missed while writing the fix — `wisp_dependencies`, whose `depends_on_issue_id` references `issues(id)`, so deleting an *ordinary* issue also drops the wisp edges that depended on it. That is a cross-plane cascade neither backend's hand-list had.

`golangci-lint --build-tags gms_pure_go`: 0 issues on the three touched packages. `internal/storage/issueops` and `internal/storage/embeddeddolt` suites pass.

## Scope note

This makes the commit contents correct. It does not change `issueops.DeleteIssueInTx`, which stays cascade-driven — adding explicit `DELETE` statements there would duplicate what the FKs already guarantee and risk diverging from them. If you'd prefer the deletes be explicit rather than declarative, that's a larger call and I'd rather take it as a separate change.

### DIFF
--- a/internal/storage/dolt/transaction.go
+++ b/internal/storage/dolt/transaction.go
@@ -678,14 +678,20 @@ func (t *doltTransaction) CloseIssue(ctx context.Context, id string, reason stri
 }
 
 func (t *doltTransaction) DeleteIssue(ctx context.Context, id string) error {
+	isWisp := t.isActiveWisp(ctx, id)
 	table := "issues"
-	if t.isActiveWisp(ctx, id) {
+	if isWisp {
 		table = "wisps"
 	}
 	if err := issueops.DeleteIssueInTx(ctx, t.txFor(table), id); err != nil {
 		return wrapExecError("delete issue in tx", err)
 	}
-	t.dirty.MarkDirty(table)
+	// Mark every table the ON DELETE CASCADE fans out to, not just the row's
+	// own table: the cascaded deletions are invisible to the SQL we issue, so
+	// staging only `issues` leaves them uncommitted in the working set.
+	for _, cascaded := range issueops.DeleteCascadeTables(isWisp) {
+		t.dirty.MarkDirty(cascaded)
+	}
 	return nil
 }
 
--- a/internal/storage/embeddeddolt/transaction.go
+++ b/internal/storage/embeddeddolt/transaction.go
@@ -84,11 +84,12 @@ func (t *embeddedTransaction) CloseIssue(ctx context.Context, id string, reason 
 }
 
 func (t *embeddedTransaction) DeleteIssue(ctx context.Context, id string) error {
-	t.dirty.MarkDirty("issues")
-	t.dirty.MarkDirty("dependencies")
-	t.dirty.MarkDirty("labels")
-	t.dirty.MarkDirty("comments")
-	t.dirty.MarkDirty("events")
+	// Shared with the dolt path so the two cannot drift again; the hand-listed
+	// set here previously omitted child_counters and the snapshot tables, which
+	// the cascade also empties.
+	for _, cascaded := range issueops.DeleteCascadeTables(issueops.IsActiveWispInTx(ctx, t.tx, id)) {
+		t.dirty.MarkDirty(cascaded)
+	}
 	return issueops.DeleteIssueInTx(ctx, t.tx, id)
 }
 
--- a/internal/storage/issueops/delete_cascade_tables_test.go
+++ b/internal/storage/issueops/delete_cascade_tables_test.go
@@ -1,6 +1,7 @@
 package issueops
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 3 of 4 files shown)


steveyegge (approved):
Reviewed as part of a maintainer pass over the open PR queue (protected storage path — extra scrutiny applied). Verdict: **approve + fix-merge** — the issue-plane half is correct and verified complete; the wisp branch had a staging regression that I fixed in a maintainer commit (2888f4616) rather than asking for another round. I also retitled the PR: the ancestor of this defect is **#3807** ("doltTransaction.DeleteIssue skips cascade to related tables"), not #4796 (that's the child-id-collision issue, fixed by #4844).

**What checks out (verified against main and the shipped migrations):**

- The issue-plane list is exactly right: I re-derived the cascade graph from the migration files and the 8 stageable tables plus the cross-plane `wisp_dependencies` match. Converging the transaction path onto the same set the store-level delete has always used is a strict merge-coherence improvement, and the "DOLT_ADD on an ignored table is a no-op" claim is independently corroborated by migration 0062's own comments.
- The migration-derivation test is genuinely good — deterministic, vacuity-guarded, and it will catch future `ON DELETE CASCADE ... issues(id)` FKs added without a staging update.

**Fixed in the maintainer commit — the wisp branch was a P0 regression:**

The premise "a delete only issues `DELETE FROM issues|wisps`, everything else is FK cascade" is false for wisps: `deleteIssueRowInTx` **explicitly** deletes the sync-plane `dependencies` rows pointing at the wisp (`DeleteWispFromDependenciesInTx` — there is no FK from `dependencies` to `wisps`, which is why the delete is explicit). Main's embedded blanket 5-table list happened to stage that deletion; the new `DeleteCascadeTables(true)` list dropped it, and since every wisp-plane name is filtered by `MarkDirty`, the tracker came up empty and **no version commit was minted at all** for an embedded wisp delete. Concrete failure: `bd mol burn` on a wisp with a permanent-issue dep edge leaves the edge deletion unstaged; the next `dolt reset --hard` (routine doctor recovery) resurrects a dangling sync-plane edge. The new wisp routing test *pinned* this bug (it errored on any non-`wisp_` table).

The fix: `dependencies` restored to the wisp set with the non-cascade rationale documented; the two missing FK-cascade members (`wisp_comments`, `wisp_child_counters` — behaviorally inert today but the function's contract is honesty) added; the derivation test parametrized over the referenced table so the wisp plane is derived from migrations too; and the routing test now asserts the `dependencies` exception in both directions.

**Noted, not blocking:** `DoltStore.DeleteIssue`/`DeleteIssues` in `dolt/issues.go` still keep independent hand lists — the same drift disease this PR cures at the transaction layer; and draft #4674 (wisp GC orphan leak) touches the same wisp-delete/dependencies seam — whoever lands it must keep the staging set in step. Both are going into follow-up beads.

The instinct to kill hand-kept lists with a schema-derived test was the right one — it just needed to model explicit deletes as well as FKs.

Issue #5033, opened 2026-07-25
bd close fails with "table not found: wisp_dependencies" after a fresh clone


balaji-dutt:
> **On AI assistance:** Claude (Opus 5 / 1M context) helped me narrow down the
> behavior and draft the wording below, including the repro steps. The investigation,
> the actual issue, and the workflow are mine. — @balaji-dutt

## Summary

`ignored_schema_migrations` is a **git-tracked** Dolt table, but every table it tracks (`local_metadata`, `wisps`, `wisp_*`, `repo_mtimes`) is registered in `dolt_ignore` and therefore never leaves the machine.

A fresh clone receives the tracker saying "migrations 1–11 applied" while none of the tables those migrations create exist. `bd` trusts the tracker, skips the `migrations/ignored/*` series, and every query against those tables fails permanently.

Neither `bd migrate schema` nor `BD_ALLOW_REMOTE_MIGRATE=1 bd migrate --yes` can repair it, because both need `local_metadata` to record progress — the very table a skipped migration would have created. Recovery requires creating that table by hand, outside `bd`.

## Environment

| | |
| --- | --- |
| `bd` | 1.1.0 (`8e4e59d39`) |
| `dolt` | 2.2.1 |
| OS | Ubuntu 24.04.4 (WSL2), also reproduced on Windows |
| config | `dolt.mode: server`, `dolt.shared-server: false`, remote-backed via `git+ssh://` |

## Reproduce

1. Have a Beads project with a Dolt remote where `ignored_schema_migrations` has been committed at some point.
2. Clone it fresh — `bd bootstrap --yes`.
3. Observe the split:

```sql
SELECT COUNT(*) FROM ignored_schema_migrations;   -- 11 rows: "all applied"
SHOW TABLES LIKE 'wisp%';                          -- empty
SHOW TABLES LIKE 'local_metadata';                 -- empty
```

## Symptoms

Any status change fails:

```
$ bd close <id>
Error closing <id>: affected by close for <id>: load wisp dependers: query:
Error 1146 (HY000): table not found: wisp_dependencies

$ bd update <id> --status closed
Error updating <id>: affected by status change for <id>: load wisp dependers: query:
Error 1146 (HY000): table not found: wisp_dependencies
```

Reads, creates, and non-status updates all still work, so the database looks healthy until you try to close something.

Both repair paths are blocked:

```
$ bd migrate schema
✓ Schema already at v53                      # only covers the schema_migrations track

[...]

PR #5188, opened 2026-07-31
fix(dolt): require consent before adopting a Dolt remote derived from git origin (#5068)

### LINKED ISSUE #5068: bd dolt push uploads to a remote derived from git origin without confirmation
> **On AI assistance:** Claude (Opus 5 / 1M context) helped me investigate and draft
> this, including the repro steps. The investigation, the testing, and the
> workflow are mine. — @balaji-dutt

## Summary

On a machine with **no Dolt remote configured**, `bd dolt push` derives one from the **git** remote named `origin`, wires it, and begins uploading the issue database — no prompt, no flag, no confirmation:

```
> bd dolt remote list
No remotes configured.

> bd dolt push
Configured Dolt remote origin from git origin.
Pushing to Dolt remote...
| Uploading...
```

In my case, the project repo where bd is initialized has a **public** repository as its git origin — so the upload target was my public repo. Both times I observed this, the upload was already in progress and was stopped only by Ctrl+C; I verified afterwards via `git ls-remote origin "refs/dolt/*"` that nothing had landed. I cannot say the uploads would have failed on their own.

## Why stealth mode is not the answer here

Stealth mode avoids this by disabling git-derived remotes (and all git ops) entirely — but stealth also gives up remote sync, so it is not the mode for the configuration this hits: a standard non-stealth setup with a public git repo and a *separate private* Dolt remote for cross-machine sync. In that configuration, any state where `dolt_remotes` ends up empty (a fresh local init, a removed remote, a recovery mid-flight) turns one habit-typed `bd dolt push` into publishing private issue history to the public repo.

## Aggravating behaviours, both verified

- **A configured `sync.remote` does not prevent it.** With `sync.remote` set in `.beads/config.local.yaml` to a valid (unreachable) URL, `bd dolt push` ignored it and derived from the git origin anyway.
- **During the push/remote churn, bd committed to git under the user's identity** — author `<user name> <user email>`, message
  `bd: clear sync.remote` — with no indication beyond the branch moving ahead of upstream. (Relatedly, `bd bootstrap` persists `sync.remote` into the *tracked* `.beads/config.yaml` on success, re-leaking a private URL into a public repo on every bootstrap.)

## Workaround

The only effective guard found: ensure no git remote named `origin` exists (`git remote rename origin gitlab`). `bd dolt push` then correctly reports:

```
Pushing to Dolt remote...
No remote is configured — skipping.
```

## Suggested fix

Never derive a push target implicitly. If no Dolt remote is configured, fail with instructions — exactly what the post-rename path already does, and what stealth mode already does by policy. At minimum, require confirmation before wiring a remote derived from git config, and never begin uploading in the same invocation that wired it.

## Environment

`bd` 1.1.0 (`8e4e59d39`), dolt 2.2.1, Windows 11 (also configured identically on WSL2 Ubuntu/Debian); `dolt.mode: server`, `dolt.shared-server: false`. Related clone/recovery findings are in #5033.

maphew:
Fixes #5068.

## Why

On a rig with **no Dolt remote configured**, `bd dolt push` derived one from `git remote get-url origin`, called `AddRemote`, persisted `sync.remote` into `.beads/config.yaml`, committed that config change **under the user's git identity**, and uploaded the full issue history. No prompt, no flag, no opt-out — just a one-line notice printed *after* it happened, which did not even name the URL.

@balaji-dutt's git origin was a **public repo**, so private issue history was published by a command they believed targeted an already-configured remote. That is the bug: not that bd can infer a remote, but that inference is indistinguishable from consent.

The pre-existing `hasConfiguredRemote` probe (59241bf16) is sometimes read as a guard here. It is not — it is a correctness fix for a stale-empty-listing edge case in the #2118 cold-start window, and it only decides *whether a remote is already configured*, never whether adopting a new one is wanted.

## What changed

Adoption now **fails closed**:

| situation | before | after |
|---|---|---|
| interactive, no remote configured | adopts + uploads | shows the URL and every side effect, defaults to **No** |
| non-interactive (script, hook, CI) | adopts + uploads | refuses, exits non-zero, names the URL and the opt-in |
| `--yes` / `-y` | n/a | consents ahead of time, still announces the target |
| `--no-adopt` / `BD_NO_REMOTE_ADOPT=1` | n/a | never adopts; **beats `--yes`** |
| remote already configured | no-op | no-op (unchanged) |

The interactive prompt discloses all four consequences before the fact, including the config commit made under your git identity — the surprise @balaji-dutt called out separately:

```
No Dolt remote is configured for this rig.
  Derived from git origin: git+ssh://git@github.com/someone/public-repo.git

Adopting it will:
  • add it as Dolt remote "origin"
  • write sync.remote into .beads/config.yaml
  • commit that config change under your git identity
  • upload your full issue history there

Adopt this remote and push? [y/N]
```

An empty line, EOF, or a read error is a **decline**.

`--no-adopt` beating `--yes` is deliberate: a `--yes` buried in an alias or wrapper should not be able to defeat a machine-level `BD_NO_REMOTE_ADOPT=1`. When two instructions contradict, the safer one wins.

**`confirmPrompt` in `bootstrap.go` is deliberately not reused.** It returns `true` when stdin is not a TTY — the right default for "proceed with the thing you asked for", and exactly the wrong one for a consent decision.

## Two things worth a maintainer's eye

**1. `bd sync` is gated too.** It is the second caller of `adoptGitOriginRemoteForPush` and publishes the same history to the same derived remote, so leaving it ungated would just move the hole. But this is a behavior change for anyone whose automated `bd sync` relied on implicit adoption: they now need `--yes` or a configured remote. If you would rather ship the push gate first and treat sync separately, that is a clean split and I will do it.

**2. Workspace resolution moved below the gate.** `selectedDoltBeadsDir()` calls `prepareSelectedNoDBContext()`, which mutates process and on-disk workspace state. On the old ordering that ran *before* consent. Deriving the URL is read-only, so the new order is: probe → derive URL → **gate** → resolve workspace → write.

I found this because tests calling the real function polluted the shared test binary and broke `TestInitDatabaseAdoptsExistingProjectID` when they ran first. That test failure was the symptom; the ordering was the cause.

## Tests

The decision is a pure function (`decideRemoteAdoption`) plus a pure application step (`applyAdoptionConsent`), so the fail-closed property is asserted by table tests rather than inferred from the call site — and needs no TTY, store, or repo to mutate.

- `TestDecideRemoteAdoptionFailsClosed` — 6 cases including the zero-value policy (what a caller that forgets to wire the flags passes) and `--no-adopt` beating `--yes`.
- `TestApplyAdoptionConsentGatesTheWrite` — the acceptance-5 assertion: no configured remote + derived URL + non-interactive ⇒ does not proceed. Everything downstream of that call writes, so "does not proceed" is "did not upload".
- `TestApplyAdoptionConsentZeroPolicyRefuses` — refusing is not enough; it must also explain.
- `TestCurrentAdoptPolicyHonorsEnvKillSwitch` — `BD_NO_REMOTE_ADOPT` parsing.
- `TestAdoptionRefusedErrorNamesTheURL` — not knowing *where* was half the reported complaint.
- The existing `TestAdoptGitOriginRemoteForPush_PersistedRemoteVetoesAdoption` now passes `AssumeYes: true`, so it still fails if the persisted-remote veto is ever moved below the consent gate.

Validated locally: `go build ./...`, `go vet ./cmd/bd`, and the adoption + sync + init test set passes with no cross-test pollution. Full `make test` is running in the local verification queue — the `cmd/bd` package exceeds `go test`'s default 10m timeout in a session. I will report the result here.

## Not addressed

@balaji-dutt also reported that a **configured** `sync.remote` does not prevent adoption. I could not reproduce that against `main`: a configured remote returns early from `hasConfiguredRemote` above everything this PR touches. If there is a repro, it is a separate defect from this one and I would like to see it. The issue body also cross-references #5033, which I have not investigated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,6 +141,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`bd dolt push` and `bd sync` no longer adopt a Dolt remote derived from
+  git origin without consent** (#5068). On a rig with no Dolt remote
+  configured, both commands silently derived one from `git remote get-url
+  origin`, added it, persisted `sync.remote` into `.beads/config.yaml`,
+  committed that config change under the user's git identity, and uploaded the
+  full issue history — no prompt, no flag, no opt-out. A public git origin
+  therefore published the whole issue database on a command the user believed
+  targeted an already-configured remote.
+
+  Adoption is now a consent decision and fails closed. Interactively, bd shows
+  the derived URL and every side effect that follows a yes (including the
+  config commit made under your git identity) and defaults to **no**.
+  Non-interactively it refuses and exits non-zero, naming the URL it would have
+  adopted and the explicit opt-in. `--yes`/`-y` consents ahead of time for
+  scripted use; `--no-adopt` or `BD_NO_REMOTE_ADOPT=1` disables adoption
+  entirely and wins over `--yes`. Rigs with a remote already configured are
+  unaffected — that path was and remains a no-op.
+
+  Workspace resolution also moved below the gate: it calls
+  `prepareSelectedNoDBContext`, which mutates workspace state, and nothing may
+  mutate before consent is established.
+
 - **Proxied-server CI shard 1 flake: `TestProxiedServerCleanDatabases` ran a
   server-global destructive command against the shared test container**
   (p1-9lf, hazard tracked in p1-8dz). The test used the shared external Dolt
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -413,7 +413,7 @@ func printNoRemoteGuidance() {
 // decideRemoteAdoption in dolt_remote_adopt.go and applied here, after the URL
 // is known and before anything is written: nothing below this point is
 // reachable without either --yes or an interactive confirmation.
-func adoptGitOriginRemoteForPush(ctx context.Context, st storage.DoltStorage, policy adoptPolicy) (bool, error) {
+func adoptGitOriginRemoteForPush(ctx context.Context, st storage.DoltStorage, policy adoptPolicy, optIn adoptOptIn) (bool, error) {
 	configured, err := hasConfiguredRemote(ctx, st)
 	if err != nil || configured {
 		return false, err
@@ -428,7 +428,7 @@ func adoptGitOriginRemoteForPush(ctx context.Context, st storage.DoltStorage, po
 	}
 	remoteURL := normalizeRemoteURL(originURL)
 
-	if proceed, err := applyAdoptionConsent(remoteURL, policy); err != nil || !proceed {
+	if proceed, err := applyAdoptionConsent(remoteURL, policy, optIn); err != nil || !proceed {
 		return false, err
 	}
 
@@ -481,7 +481,10 @@ var doltPushCmd = &cobra.Command{
 	Short:         "Push commits to Dolt remote",
 	Long: `Push local Dolt commits to the configured remote.
 
-Requires a Dolt remote to be configured in the database directory.
+Requires a Dolt remote to be configured in the database directory. With no
+remote configured, bd can adopt one derived from git origin — only with
+consent: interactively, or via --yes; --no-adopt or BD_NO_REMOTE_ADOPT=1
+disables adoption entirely.
 For Hosted Dolt, set DOLT_REMOTE_USER and DOLT_REMOTE_PASSWORD environment
 variables for authentication.
 
@@ -534,8 +537,8 @@ The remote must already exist (see 'bd dolt remote add').`,
 		}
 		assumeYes, _ := cmd.Flags().GetBool("yes")
 		noAdopt, _ := cmd.Flags().GetBool("no-adopt")
-		policy := currentAdoptPolicy(assumeYes, noAdopt, stdinIsTerminal())
-		if adopted, err := adoptGitOriginRemoteForPush(ctx, st, policy); err != nil {
+		policy := currentAdoptPolicy(assumeYes, noAdopt, stdinIsTerminal(), jsonOutput)
+		if adopted, err := adoptGitOriginRemoteForPush(ctx, st, policy, pushAdoptOptIn); err != nil {
 			return HandleError("%v", err)
 		} else if adopted {
 			fmt.Println("Configured Dolt remote origin from git origin.")
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -408,20 +408,35 @@ func printNoRemoteGuidance() {
 // when the persisted remote and the git origin disagree — a Dolt remote
 // deliberately pointed elsewhere, or a renamed/redirected origin — the rig
 // starts pushing somewhere else on the strength of a stale listing (wy-82hc5).
-func adoptGitOriginRemoteForPush(ctx context.Context, st storage.DoltStorage) (bool, error) {
+//
+// Adoption requires consent (#5068). The policy is decided by
+// decideRemoteAdoption in dolt_remote_adopt.go and applied here, after the URL
+// is known and before anything is written: nothing below this point is
+// reachable without either --yes or an interactive confirmation.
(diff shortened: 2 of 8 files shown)


maphew:
CI caught two failures my local run could not, and they were mine, not base breakage. Fixed in 8e5b97f.

`TestEmbeddedInit/dolt_push_lazily_adopts_later_git_origin` and `.../dolt_push_adopts_target_origin_with_dash_c` assert that `bd dolt push` adopts a git origin. A test process has no TTY, so the new gate correctly refused and both failed.

Those two cover a capability this PR **keeps**, not the behavior it removes, so they now pass `--yes` — the scripted consent. Their assertions are otherwise untouched.

I also added `dolt_push_refuses_to_adopt_without_consent` beside them, asserting all four things that must not happen with no TTY and no `--yes`: the push fails, the refusal names the remote it would have adopted, no Dolt remote is added, `sync.remote` is not persisted, and `refs/dolt/data` is absent on the bare remote — i.e. nothing was uploaded. That last one is what the issue is actually about, and it was previously only covered at the unit level.

Worth flagging for anyone else working in this area: `cmd/bd/init_embedded_test.go` skips unless `BEADS_TEST_EMBEDDED_DOLT=1`, and CI runs it as the sharded `Test (Embedded Dolt Cmd N/20)` jobs. **A green `make test` does not exercise it**, which is exactly how I shipped a red PR after reporting a passing full suite. My earlier "full make test is in the verification queue" note stands for what it covered, but it did not cover this.

maphew:
Confirmed on `8e5b97f`: all 20 `Test (Embedded Dolt Cmd N/20)` shards pass, including 15/20 which carried both failures. Full rollup is 0 failing, 0 pending.

For completeness, since I said I would report it: my local re-run of those subtests was **inconclusive rather than a pass** — it hit a 40-minute timeout before finishing. CI is the authority for this suite; it is far too slow to iterate on locally.

steveyegge (approved):
Reviewed as part of a maintainer pass over the open PR queue. Verdict: **approve + fix-merge** — the gate is correctly placed and fail-closed; I pushed one maintainer commit (f2caba310) closing two consent-surface gaps rather than asking for another round.

- The gate sits before every write: URL derivation is read-only, and refusal/decline happens above `selectedDoltBeadsDir`, `AddRemote`, the config persist, the git commit, and the upload. The embedded e2e test asserts all four side-effect absences — push fails, refusal names the URL, `dolt remote list` clean, no `refs/dolt/data` on the bare remote.
- Configured rigs are a no-op (`hasConfiguredRemote` early-return), explicit `--remote` paths never reach adoption, and `--no-adopt`/`BD_NO_REMOTE_ADOPT=1` correctly beats `--yes`. The `[y/N]` default-no inversion of bootstrap's `[Y/n]` is the right call for a consent (vs proceed-with-what-you-asked) prompt, and the code says why.
- Prior art: no duplicate; #4449 (bootstrap-side sync.remote guard) and #5139 are adjacent but orthogonal. The deliberate behavior break for non-interactive sync-side bring-up (wy-gpzg7 rationale) is disclosed in the code comment — wheelhouse bring-up scripts that relied on silent adoption now need `--yes`, which is the point of the fix.

**Fixed in the maintainer commit:**

- `--json` was not treated as non-interactive: an agent harness that allocates a pty running `bd sync --json` on a fresh rig would hit `adoptAsk` and block forever on a stderr prompt no JSON consumer answers. The policy now folds JSON mode into non-interactivity (same rule as `validateHookMigrationApplyConsent`), with a regression test.
- The refusal and prompt hardcoded `bd dolt push --yes`, so a scripted `bd sync` that newly fails would steer its user into a pull-less push. Adoption messages now carry the caller's own opt-in command (`bd sync --yes` from sync), tested both directions.
- Also folded in: `bd dolt push --help` now documents adoption/`--yes`/`--no-adopt`, and the dead `errors.Is(err, nil)` no-op is removed.

**Noted, not blocking:** the interactive y/n parsing reads `os.Stdin` directly with no reader seam, so accept/decline parsing has no unit coverage; and the interactive decline exits 1 where bootstrap's decline exits 0 — defensible (scripts notice), just inconsistent. Both fine as follow-ups if they ever bite.

The fail-closed design, the pure-policy/table-test split, and the honest disclosure of what CI caught are all exactly how this should be done.

Issue #3884, opened 2026-05-11
bd export → JSONL is lossy: wisps/events/comments not round-tripped; --from-jsonl rebuild shrunk a 51 GB dolt DB to 53 MB


vbtcl:
## Summary

`bd export` (and the `.beads/issues.jsonl` it produces) only captures the issues view. Wisps, wisp comments, events, custom statuses, and other tables that live in the dolt database are not serialized. So `bd init --from-jsonl --force` (or `bd bootstrap` falling through to the JSONL branch) rebuilds a workspace that has the issue rows but none of the supporting history.

On a small project this is a minor inconvenience. On a workspace where dolt is the primary store and the database has been live for weeks, it's catastrophic.

## Concrete case (2026-05-11)

- Per-project sql-server setup on dolt 1.86.6, single workspace.
- `beads_hq` journal corrupted at offset 26796492 after a kill mid-fsync (independent operational issue, but recovery exposed this).
- Pre-corruption storage size of `beads_hq`:
  - `.dolt/noms/vvvv...` = 13.1 GB (journal)
  - `.dolt/noms/oldgen/` = ~38 GB (chunked old-gen)
  - **Total: 51 GB**
- Recovery via `bd init --from-jsonl --force --destroy-token "DESTROY-gc" --non-interactive`:
  - Imported 889 issues from `.beads/issues.jsonl` (550 KB file)
  - Post-rebuild `beads_hq` storage: **53 MB**

That's a ~1000× shrinkage. The issue rows are intact; everything else is gone.

## Why it matters

- The bd help text presents `--from-jsonl` and the bootstrap JSONL branch as legitimate recovery paths, but they aren't equivalent to the on-disk state they're recovering from.
- For users without `sync.remote` configured (and no fallback `dolt_backup` archive), this is *the* recovery path bd offers — and it silently destroys history.

## Suggested fixes (any of these would help)

1. **Make `bd export` capture more tables.** At minimum: wisps, wisp_comments, events, custom_statuses. Document which tables the JSONL is supposed to round-trip.
2. **Make `bd bootstrap` prefer `dolt_backup` archives** over JSONL when both are available, and require explicit `--allow-jsonl-recovery` (or similar) when JSONL is the only source. Print a warning that lists what won't be recovered.
3. **At minimum: add a clear warning to `bd init --from-jsonl --force` output** spelling out the columns/tables that won't be preserved, and refuse to run without an `--accept-lossy-recovery` flag.

[...]

PR #5189, opened 2026-07-31
fix(init): re-init inherits the workspace's connection mode instead of silently rewriting it (#3885)

### LINKED ISSUE #3885: bd init --from-jsonl --force silently downgrades dolt_mode from "server" to "embedded"
## Summary

In a workspace where `.beads/metadata.json` has `"dolt_mode": "server"` and a sql-server is running, `bd init --from-jsonl --force` (used during recovery) overwrites `metadata.json` with `"dolt_mode": "embedded"`. The rebuilt database also lands in `.beads/embeddeddolt/` instead of `.beads/dolt/`. Subsequent bd commands then fail with `'bd dolt show' is not supported in embedded mode` and the user is left in a half-configured state.

## Reproduction

```
# Workspace state before: metadata.json has "dolt_mode": "server",
# sql-server running, data_dir at .beads/dolt/<dbname>.

# Trigger a recovery (any reason, e.g. broken journal):
mv .beads/dolt .beads/dolt.broken
bd init --from-jsonl --force --destroy-token "DESTROY-<prefix>" --non-interactive

# Result:
# - Issues imported successfully into .beads/embeddeddolt/<dbname>/
# - metadata.json now has "dolt_mode": "embedded"
# - bd dolt show fails: "not supported in embedded mode"
# - bd dolt start fails: "not supported in embedded mode"
# - Existing sql-server can't see the rebuilt database (wrong data_dir)
```

## Expected

`--force` should re-initialize the database while preserving the existing `dolt_mode` (and the data_dir layout that implies). If the user wants to change modes, they should pass `--server` / `--embedded` explicitly.

If preserving mode is too invasive, at minimum log a prominent warning: "Mode changed from server → embedded; data will land in .beads/embeddeddolt/ — pass --server to keep server mode."

## Actual

Silent mode change. The only way I noticed was that the gc-managed sql-server kept reporting `Failed to load database beads_hq` after the rebuild ostensibly succeeded.

## Workaround

After `bd init --from-jsonl --force`:

```
# 1. Stop any running sql-server.
# 2. Move the rebuilt database to the server data_dir:
mv .beads/embeddeddolt/<dbname> .beads/dolt/<dbname>
# 3. Edit .beads/metadata.json:
#      "dolt_mode": "server",
#      "dolt_server_host": "127.0.0.1",
#      "dolt_server_port": <port>,
#      "dolt_server_user": "root"
# 4. Restart the sql-server.
```

Three-step manual recovery that bd could just… do.

## Environment

- bd 1.0.2 (c446a2ef)
- dolt 1.86.6
- Linux

## Related
- Companion issue: `bd export` / JSONL round-trip is lossy.
- Companion issue: `bd bootstrap` fails with "Error 1105: nothing to commit" when dolt dir exists but is empty (which is what pushed me to `--from-jsonl --force` in the first place).

maphew:
Fixes #3885.

## Why

`bd init --from-jsonl --reinit-local` (and `--force`, its deprecated alias) on a server-mode project brings it back as **embedded**, silently. @vbtcl reported this in May with a concrete repro; the issue's timeline is empty and nothing has touched it since.

The mechanism, at `cmd/bd/init.go`:

```go
// Persist the connection mode matching this build.
switch {
case usesProxiedServer(): cfg.DoltMode = configfile.DoltModeProxiedServer
case usesSQLServer():     cfg.DoltMode = configfile.DoltModeServer
default:                  cfg.DoltMode = configfile.DoltModeEmbedded
}
```

This runs *after* the existing `metadata.json` has been loaded into `cfg`, and unconditionally overwrites the mode. The globals behind `usesSQLServer()` / `usesProxiedServer()` are seeded only from `--server`, the `BEADS_DOLT_*` env vars, or the **global** `config.yaml` `dolt.mode` key — never from `existingCfg.DoltMode`. Grepping every `DoltMode` occurrence in `init.go` finds writes plus that one YAML lookup and no read of the existing workspace value.

So re-running init without re-passing `--server` writes `dolt_mode: embedded` over a server config. The user is told nothing, and the project comes back half-configured.

(The narrower `--shared-server` propagation fix noted at `init.go:471-473` for #2946 does not cover this case.)

## The fix, and why it goes where it does

The mode is now inherited from the existing workspace **at the seeding block, alongside the other mode sources** — not patched onto `cfg.DoltMode` at the persistence point.

That placement is the whole point. Everything between the two — which engine init opens, which database-name rules apply (`init.go:552-558`), where the data actually lands — reads the *process* mode. Patching only the persisted value would write `server` into `metadata.json` describing a database init had just built embedded: a silent downgrade traded for a silent mismatch, where the next open fails instead.

Inheritance applies **only when the invocation names no mode of its own**, so everything explicit still wins:

| signal | wins over inheritance? |
|---|---|
| `--server`, `--shared-server`, `--proxied-server` | yes |
| `BEADS_DOLT_SERVER_MODE`, `BEADS_DOLT_SHARED_SERVER`, `BEADS_DOLT_PROXIED_SERVER` | yes |
| global `config.yaml` `dolt.mode` | yes (the seeding block already treats it like `--server`) |
| nothing named | **inherit the workspace's recorded mode** |

A flag set explicitly to its default counts as explicit too — `--server=false` is the user typing a downgrade, and `f.Changed` catches it. There is no `--embedded` flag today, so this is currently the only way to state one; if you would like a real `--embedded`, say so and I will add it here or in a follow-up.

Any actual mode change on an existing workspace is now announced on stderr naming both modes, since it moves where the project's data lives.

Two helpers, both mirroring the established shape of `existingWorkspaceDBName()` right beside them:

- `existingWorkspaceDoltMode()` reads the **raw** `DoltMode` field (normalized for case/whitespace like `configfile`'s own `IsServerMode`). Unset stays distinguishable from "deliberately embedded" — defaulting here would make every fresh init look like a mode change.
- `initModeExplicitlyRequested(cmd)` is the flag/env/config decision, kept pure enough to table-test.

## Tests

`cmd/bd/init_mode_inherit_test.go`:

- `TestExistingWorkspaceDoltModeReadsRawField` — server / proxied-server / embedded / case-and-space normalization / unset-stays-unset.
- `TestExistingWorkspaceDoltModeMissingWorkspace` — no workspace ⇒ `""`, not a phantom default.
- `TestInitModeExplicitlyRequested` — each flag, each env var, `--server=false`, and the case that matters most: **a bare invocation is not an explicit request**. Get that one wrong in the permissive direction and #3885 comes straight back, because a bare re-init would again count as "the user asked for embedded".

## A note on local validation

The pre-existing `cmd/bd` init suite cannot be meaningfully run from a worktree nested under another beads workspace: those tests resolve the ambient `.beads` and fail with `Found existing Dolt database: .../.beads/embeddeddolt/...`. I checked that baseline on **unmodified `upstream/main`** in the same layout and got the identical failure set before concluding this change is clean. `go build ./...`, `go vet ./cmd/bd`, and the new tests all pass; full `make test` is in my local verification queue and I will report the result here.

This is worth knowing independently of this PR — it makes the init suite effectively CI-only for anyone whose checkout sits inside another beads project.

## Scope

This is the config half of the re-init problem. The data half — that JSONL is a deliberate subset (wisps, events/history, memories unless `--all`) while the destructive prompt only says "this will destroy the existing database" — is #3884 and is deliberately not in this PR.

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -552,8 +552,11 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 		// explicit --server/--shared-server/--proxied-server, the BEADS_DOLT_*
 		// env vars, and a global dolt.mode all still win.
 		if !initServerMode && !initModeExplicitlyRequested(cmd) {
-			switch existingWorkspaceDoltMode() {
-			case configfile.DoltModeServer:
+			inheritServer, inheritErr := inheritWorkspaceDoltMode()
+			if inheritErr != nil {
+				return inheritErr
+			}
+			if inheritServer {
 				initServerMode = true
 				serverMode = true
 				if cmdCtx != nil {
@@ -561,14 +564,6 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 				}
 				if !quiet {
 					fmt.Fprintln(os.Stderr, "Preserving server mode from the existing .beads/metadata.json.")
-				}
-			case configfile.DoltModeProxiedServer:
-				proxiedServerMode = true
-				if cmdCtx != nil {
-					cmdCtx.ProxiedServerMode = true
-				}
-				if !quiet {
-					fmt.Fprintln(os.Stderr, "Preserving proxied-server mode from the existing .beads/metadata.json.")
 				}
 			}
 		}
@@ -2570,6 +2565,29 @@ func existingWorkspaceDoltMode() string {
 	// Matched case-insensitively by callers, mirroring configfile's own
 	// IsServerMode/IsProxiedServerMode comparisons.
 	return strings.ToLower(strings.TrimSpace(cfg.DoltMode))
+}
+
+// inheritWorkspaceDoltMode is the decision half of #3885's fix: what a bare
+// re-init adopts from the workspace it is re-initializing. It returns
+// inheritServer=true when the workspace records server mode.
+//
+// Proxied-server is deliberately an error, not an inheritance: the dedicated
+// proxied init path (runInitProxiedServer) dispatches on the explicit flag
+// BEFORE the inheritance point in RunE, so inheriting by mutating only the
+// process-mode globals would build the database embedded while metadata.json
+// kept claiming proxied-server — the exact silent mismatch inheritance exists
+// to prevent. The mode is experimental and dark-launched; a re-init of such a
+// workspace must name it explicitly so it routes through the real init path.
+func inheritWorkspaceDoltMode() (bool, error) {
+	switch existingWorkspaceDoltMode() {
+	case configfile.DoltModeServer:
+		return true, nil
+	case configfile.DoltModeProxiedServer:
+		return false, fmt.Errorf("this workspace is recorded as proxied-server in .beads/metadata.json; " +
+			"re-run with --proxied-server to keep it, or name another mode explicitly to change it")
+	default:
+		return false, nil
+	}
 }
 
 // initModeExplicitlyRequested reports whether this invocation names a
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -538,6 +538,41 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 			// Non-fatal - continue with defaults
 		}
 
+		// Re-init inherits the existing workspace's connection mode (#3885).
+		//
+		// This must be seeded HERE, with the other mode sources, not patched
+		// onto cfg.DoltMode where it is persisted further down: everything
+		// between the two points — which engine init opens, which database
+		// name rules apply, where the data lands — reads the process mode. A
+		// late fixup would write "server" into metadata.json describing a
+		// database that init had just built embedded, trading a silent
+		// downgrade for a silent mismatch.
+		//
+		// Only applies when this invocation names no mode of its own, so an
+		// explicit --server/--shared-server/--proxied-server, the BEADS_DOLT_*
+		// env vars, and a global dolt.mode all still win.
+		if !initServerMode && !initModeExplicitlyRequested(cmd) {
+			switch existingWorkspaceDoltMode() {
+			case configfile.DoltModeServer:
+				initServerMode = true
+				serverMode = true
+				if cmdCtx != nil {
+					cmdCtx.ServerMode = true
+				}
+				if !quiet {
+					fmt.Fprintln(os.Stderr, "Preserving server mode from the existing .beads/metadata.json.")
+				}
+			case configfile.DoltModeProxiedServer:
+				proxiedServerMode = true
+				if cmdCtx != nil {
+					cmdCtx.ProxiedServerMode = true
+				}
+				if !quiet {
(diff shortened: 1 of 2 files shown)


maphew:
Full `make test` has now finished on this branch's head (`82f16b8`) in my local verification queue: **exit 0**, whole suite green.

That closes the gap I flagged in the description — the `cmd/bd` init tests that fail in a nested-workspace layout pass here, because the queue runs each job in a clean detached worktree outside any other beads workspace.

steveyegge (approved):
Reviewed as part of a maintainer pass over the open PR queue. Verdict: **approve + fix-merge** — the diagnosis of #3885 and the server-mode fix are exactly right; one branch of the inheritance was unsafe and I replaced it in a maintainer commit (4e40c2895) rather than asking for another round.

- The placement argument is correct for embedded/server: the mode must be seeded with the other mode sources, because everything between seeding and persistence reads the process mode. The server branch sets all three of `initServerMode`/`serverMode`/`cmdCtx.ServerMode`, matching the config.yaml-fallback idiom.
- Precedence is right: explicit `--server`, `--server=false` (`f.Changed`), the `BEADS_DOLT_*` env vars, and a global `dolt.mode` all outrank inheritance; fresh init is unchanged (no metadata → no inheritance, and `resolveInitBeadsDir` does no upward search, so there is no accidental parent-workspace inheritance). Server host/port/user survive because `cfg` aliases `existingCfg`.
- Prior art: no duplicate. **#5178 is complementary, not competing** — its env vars (`BEADS_DOLT_SERVER_{HOST,PORT,USER,TLS}`) don't intersect this PR's explicitness checks, and the `init.go` hunks are adjacent but non-overlapping. Whichever lands second should rebase and re-verify that region.

**Fixed in the maintainer commit — the proxied-server branch was a P0:**

The dedicated proxied init path (`runInitProxiedServer`) dispatches on the explicit flag *above* the new inheritance block, and the branch set only the process-mode globals, never `initProxiedServer`. On a bare re-init of a proxied-server workspace that meant: data built **embedded**, `usesProxiedServer()` true at persist time so `metadata.json` records `proxied-server`, `priorMode == cfg.DoltMode` so the new mode-change notice stays silent — and the user is affirmatively told the mode was "preserved". That is the exact silent mismatch the PR body's placement rationale warns against, and it partially resurrects a mode that is deliberately dark-launched. Since inheritance cannot route into the proxied init path from that point, the branch now **refuses with guidance** ("re-run with --proxied-server to keep it, or name another mode explicitly"). The decision is extracted into `inheritWorkspaceDoltMode()` with unit coverage for all four workspace shapes — the composition-level test whose absence let this slip through.

**Noted, not blocking (would-take follow-ups):** the init metrics event is computed before inheritance, so an inherited server re-init reports `init-embedded`; the "Preserving server mode" notice also prints when `--init-if-missing` subsequently skips the init; and `initModeExplicitlyRequested` treats any non-empty `BEADS_DOLT_SERVER_MODE` as explicit while seeding honors only `"1"` — a stale `=0` blocks inheritance, which is defensible as an explicit downgrade but worth documenting.

The `f.Changed`-vs-value distinction, the no-`--embedded`-flag handling, and the honest scope split with #3884 in the body are all well done.

Issue #4356, opened 2026-06-10
#4270 content-hash install leaves a permanently dirty working set on pre-existing DBs (ignored_schema_migrations tracked in HEAD) - every bd dolt pull fails until a manual DOLT_ADD -f


seanmartinsmith:
## summary

upgrading an existing fleet to tip-of-main (`c6802bcc2`): the #4270 content-hash install
ALTERs `ignored_schema_migrations` on first open, but on DBs where that table is **tracked in
HEAD** (any lineage where the table predates the `dolt_ignore` entry), bd's auto-commit and
`bd vc commit` silently skip staging the change — the table is ignored, so `DOLT_ADD` is a
no-op ("0 rows affected"). result: a **permanently dirty working set**, and every
`bd dolt pull` fails with

```
Error 1105: cannot merge with uncommitted changes
```

fresh DBs never see this (the table is created after the ignore entry, so it's untracked).
it bites exactly the people upgrading real multi-machine DBs — push works, pull never does.

## environment

- bd `c6802bcc2` (tip-of-main 6/10, post-#4326/#4268/#4270), dolt 2.1.4 shared-server mode
- 17 DBs, all pre-#4270 lineages; hit on every single one

## workaround (applied to all 17, everything converged after)

```sql
CALL DOLT_ADD('-f','ignored_schema_migrations');  -- -f = allow adding ignored tables
CALL DOLT_COMMIT('-m','schema: content_hash column on ignored_schema_migrations');
```

## suggested fix

when the install itself ALTERs an ignored table, force-stage its own change (DOLT_ADD -f)
in the same auto-commit — or detect the tracked-but-ignored state and untrack/commit it
explicitly. happy to test a patch on the same fleet.

## context

field report from the upgrade described in #4353 (same fleet, same day the #4270 work
merged). the new #4268 gate behaved exactly right during the same upgrade, fwiw — refused
the one embedded-mode straggler instead of silently migrating it.

seanmartinsmith:
two refinements to the workaround from running it fleet-wide (24 DBs across windows/wsl/macos, all on `c6802bcc2`) - both matter for anyone scripting the recovery, and for whatever fix lands here:

[...]

PR #5190, opened 2026-07-31
fix(schema): stop believing an ignored-migration cursor the schema contradicts (#5033)

### LINKED ISSUE #5033: bd close fails with "table not found: wisp_dependencies" after a fresh clone

maphew:
Fixes #5033. Same root cause as #4356.

## Why

`currentVersion()` already special-cases a **missing cursor table** as "nothing applied":

```go
if dberrors.IsTableNotExist(err) {
    return 0, nil
}
```

but it believed a cursor whose **tables** do not exist. `ignored_schema_migrations` is itself in `doltIgnorePatterns` — it is dolt-ignored and therefore clone-local, exactly like the wisp tables it tracks. So a database materialized out of band (table-by-table copy, dump restore, or a clone that picked up the cursor rows without the clone-local tables they describe) arrives claiming **at-latest with no wisps tables**.

From there:

`atLatest()` → true → `migrationWorkNeeded()` short-circuits → the ignored series never re-runs → the user meets it much later and much further away as **`table not found: wisp_dependencies`** on `bd close`. That is #5033 exactly. #4356 is the same root cause seen from the dirty-working-set side.

The cursor is a *claim about the schema*. A claim the schema contradicts is worth less than no claim at all.

## Why re-running repairs rather than clobbers

This is the part that makes healing possible instead of merely diagnosing, so it is worth stating explicitly: **the ignored series is written to be re-runnable.**

`migrations/ignored/0001_create_local_state_tables.up.sql` builds every table as `__temp__<name>`, then:

```sql
SET @sql = IF(@exists = 0, 'RENAME TABLE __temp__wisps TO wisps', 'DROP TABLE __temp__wisps');
```

— it renames into place only when the real table is absent, and drops its temp otherwise. Later migrations gate their `ALTER`s on `INFORMATION_SCHEMA` lookups (e.g. `0003`, `0006`). So a re-run recreates what is missing and leaves existing data untouched.

`TestSentinelTablesAreCreatedByTheSeries` pins this: a sentinel the series cannot create would reject the cursor on *every* open forever, turning a silent wrong answer into a permanent re-migration loop.

## Scope, deliberately narrow

| decision | why |
|---|---|
| only the **ignored** series is guarded (`wisps`, `wisp_dependencies`) | the main series' tables are not clone-local and cannot reach this state; probing it would add two queries to every store open for nothing. `TestMainSourceIsNotProbed` pins that it is never probed. |
| probe skipped entirely for a **zero cursor** | no added cost on databases with nothing applied |
| **both** sentinels checked, not just the first | #5033's actual shape is `wisp_dependencies` missing while `wisps` exists |
| a probe **error** propagates, never becomes "contradicted" | re-running the series on a database whose state could not be read is the one outcome worse than the bug |

The `INFORMATION_SCHEMA` lookup sits behind the `sentinelTableExists` function variable, following this package's existing `issueRowCounter` seam convention, so the decision is testable without a live database.

## Tests

New `internal/storage/schema/cursor_reality_test.go`:

- `TestCursorContradictedBySchema` — no tables / `wisps` only / `wisp_dependencies` only / both present.
- `TestCursorProbeErrorIsNotTreatedAsContradicted` — an unreadable probe must not trigger a re-migration.
- `TestMainSourceIsNotProbed` — the main series is never probed, and the ignored guard is not inert.
- `TestSentinelTablesAreCreatedByTheSeries` — the sentinel list stays repairable-by-re-run.
- `TestMigrationWorkNeededWhenWispTablesAbsent` — **the end-to-end assertion**: both cursors read at-latest, no wisp tables, `migrationWorkNeeded` must now return `true`. This is the exact short-circuit that produced #5033. It also asserts only one probe is issued, since the check runs on every store open.

Several existing sqlmock tests assert exact query ordering, so the ones whose flow now issues sentinel probes gained an `expectIgnoredSentinelProbes()` expectation (new helper beside `expectScalar` in `lock_test.go`). **No existing assertion was weakened or removed** — the additions are purely the two new queries.

`go build ./...`, `go vet ./internal/storage/schema`, and the full `./internal/storage/schema` package are green. I confirmed the package was green on unmodified `upstream/main` first, so that baseline is real rather than assumed. Full `make test` is in my local verification queue and I will report the result here.

## Not addressed

The bead this came from also notes there is **no self-heal for the mirror-image state** — a cursor that is missing rows for migrations that *were* applied — and no `DOLT_RM`-based reassertion anywhere. This PR only fixes the direction that causes #5033 and #4356. Whether the cursor should be reconciled against the schema in both directions is a larger question I have deliberately left alone.

### DIFF
--- a/internal/storage/schema/aux_row_id_backfill_test.go
+++ b/internal/storage/schema/aux_row_id_backfill_test.go
@@ -201,6 +201,7 @@ func TestRekeyAuxRowIDsSkipsWhenMarkerRecorded(t *testing.T) {
 
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM ignored_schema_migrations",
 		"version", auxRekeyPassInitial.markerVersion)
+	expectIgnoredSentinelProbes(mock, true)
 	// No further expectations: no table may be probed or scanned.
 
 	wrote, err := rekeyAuxRowIDs(context.Background(), db, auxRekeyPassInitial.shippedMainVersion-1, auxRekeyPassInitial)
@@ -226,6 +227,7 @@ func TestRekeyAuxRowIDsRunsAllTablesWhenMarkerPending(t *testing.T) {
 
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM ignored_schema_migrations",
 		"version", auxRekeyPassInitial.markerVersion-1)
+	expectIgnoredSentinelProbes(mock, true)
 	expectAuxRekeySentinel(mock, false)
 	expectSetAuxRekeySentinel(mock)
 	// Each of the four tables is probed; this mocked world has none of them,
@@ -291,6 +293,7 @@ func TestRekeyAuxRowIDsSkipsConvergedLineage(t *testing.T) {
 
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM ignored_schema_migrations",
 		"version", auxRekeyPassInitial.markerVersion-1)
+	expectIgnoredSentinelProbes(mock, true)
 	expectAuxRekeySentinel(mock, false)
 	// No further expectations: marker is pending, but the pre-pass main
 	// cursor at the watershed and no crash sentinel means no table may be
@@ -323,6 +326,7 @@ func TestRekeyAuxRowIDsResumesAfterCrash(t *testing.T) {
 
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM ignored_schema_migrations",
 		"version", auxRekeyPassInitial.markerVersion-1)
+	expectIgnoredSentinelProbes(mock, true)
 	expectAuxRekeySentinel(mock, true)
 	expectSetAuxRekeySentinel(mock)
 	for range auxRekeyTables {
@@ -352,6 +356,7 @@ func TestRekeyAuxRowIDsKeepsSentinelOnFailure(t *testing.T) {
 
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM ignored_schema_migrations",
 		"version", auxRekeyPassInitial.markerVersion-1)
+	expectIgnoredSentinelProbes(mock, true)
 	expectAuxRekeySentinel(mock, false)
 	expectSetAuxRekeySentinel(mock)
 	// First table probe fails; no DELETE of the sentinel may follow.
--- a/internal/storage/schema/content_hash_test.go
+++ b/internal/storage/schema/content_hash_test.go
@@ -111,6 +111,7 @@ func TestMigrationWorkNeededAddsContentHashColumnOnUpToDateDB(t *testing.T) {
 	// Both sources are already at their latest version...
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM schema_migrations", "version", LatestVersion())
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM ignored_schema_migrations", "version", LatestIgnoredVersion())
+	expectIgnoredSentinelProbes(mock, true)
 	// ...but schema_migrations predates the content_hash column.
 	mock.ExpectQuery(`SHOW COLUMNS FROM schema_migrations LIKE 'content_hash'`).
 		WillReturnRows(showColumnsRows())
@@ -140,6 +141,7 @@ func TestMigrationWorkNotNeededWhenContentHashColumnsPresent(t *testing.T) {
 
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM schema_migrations", "version", LatestVersion())
 	expectScalar(mock, "SELECT COALESCE(MAX(version), 0) FROM ignored_schema_migrations", "version", LatestIgnoredVersion())
+	expectIgnoredSentinelProbes(mock, true)
 	mock.ExpectQuery(`SHOW COLUMNS FROM schema_migrations LIKE 'content_hash'`).
 		WillReturnRows(showColumnsRows("content_hash"))
 	mock.ExpectQuery(`SHOW COLUMNS FROM ignored_schema_migrations LIKE 'content_hash'`).
--- a/internal/storage/schema/cursor_reality_test.go
+++ b/internal/storage/schema/cursor_reality_test.go
@@ -1,0 +1,180 @@
+package schema
+
+import (
+	"context"
+	"errors"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+)
+
+// stubSentinelTables replaces the INFORMATION_SCHEMA probe with a fixed answer
+// for the duration of a test, following the issueRowCounter seam convention in
+// this package. The DBConn is never touched, so these run with no database.
+func stubSentinelTables(t *testing.T, present map[string]bool, probeErr error) *[]string {
+	t.Helper()
+	var probed []string
+	orig := sentinelTableExists
+	t.Cleanup(func() { sentinelTableExists = orig })
+	sentinelTableExists = func(_ context.Context, _ DBConn, table string) (bool, error) {
+		probed = append(probed, table)
+		if probeErr != nil {
+			return false, probeErr
+		}
+		return present[table], nil
+	}
+	return &probed
+}
+
+// TestCursorContradictedBySchema is the gh 5033 / gh 4356 regression: the
+// cursor is a claim about the schema, and a claim the schema contradicts must
+// not be believed. Before this, a database whose clone-local wisp tables were
+// absent but whose (also clone-local) cursor rows survived reported at-latest,
+// migrationWorkNeeded() short-circuited, the ignored series never re-ran, and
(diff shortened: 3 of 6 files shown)


maphew:
One more check a reviewer will reasonably want, since it is the failure mode that would make this change worse than the bug: **does the repair re-run collide on the cursor rows that are already there?**

No. Recording is `INSERT IGNORE`:

```go
db.ExecContext(ctx, "INSERT IGNORE INTO "+src.cursorTable+" (version, content_hash) VALUES (?, ?)", mf.version, contentHash)
```

So a version already present is a silent no-op rather than a duplicate-key error on the `version` PRIMARY KEY. Combined with the `__temp__`/`RENAME`-only-if-absent shape of `ignored/0001` and the `INFORMATION_SCHEMA`-gated `ALTER`s in the later migrations, the whole repair path is idempotent from the SQL up: re-running creates what is missing, leaves existing data alone, and re-records nothing.

steveyegge (approved):
Reviewed as part of a maintainer pass over the open PR queue (schema/migration path — maximum scrutiny applied). Verdict: **approve + fix-merge** — the detection design is right and carefully argued, but the fix was inert end-to-end; I completed it in a maintainer commit (a45472e96) rather than asking for another round, and retitled the PR to #5033 only.

**The P0, now fixed:** the guard lived in `currentVersion()`, but the applier (`migrationSource.migrate`) re-read the cursor with a raw `SELECT COALESCE(MAX(version), 0)`. In the exact #5033 state — cursor at-latest, wisp tables absent — `migrationWorkNeeded()` decided "work needed" on every store open, MigrateUp ran its full pass preamble (dirty-table scans, unstage, ignore-seed), and then `current >= target` short-circuited the ignored pass: **zero migrations applied, tables still missing, futile pass repeating on every open**. The flagship test stopped at `migrationWorkNeeded` and could not see this. The applier now reads through the guarded `currentVersion`, and a new end-to-end test drives `migrate()` under an at-latest-but-contradicted cursor and asserts the ignored series actually begins applying — mutation-checked: it fails on the pre-fix code and passes with the fix. The three lock-test mocks gained the sentinel probes the applier now makes; full schema package suite is green.

**What checks out (verified against main and all 21 ignored migration files):**

- Detection is sound: INFORMATION_SCHEMA existence probes (same Dolt-proven pattern as the aux-rekey sentinel), probe errors propagate rather than laundering into "contradicted", the cursor rows are never written at detection time, and the re-run is genuinely repair-safe — 0001's `__temp__`+rename-if-absent, the INFORMATION_SCHEMA/IF-NOT-EXISTS gates in the later files, and `INSERT IGNORE` recording all hold up under re-execution.
- **The #4259 remote-migrate gate is not weakened**: it keys exclusively on the main source, which has no sentinels and is provably never probed (`TestMainSourceIsNotProbed`). The ignored series has always run ungated; its tables are clone-local by design.
- Dolt sync plane: safe. Detection is read-only; the heal creates only dolt-ignored clone-local tables; clones with divergent cursors merging are unaffected because the guard never writes the cursor.
- Prior art: no duplicate. Interaction flag: #4804 (skip migration lock when current) touches the same `migrationWorkNeeded` short-circuit from the perf side — if it lands, it must not cache/skip past this probe. Noted for whoever reviews it.

**Retitle:** the title claimed #4356, but nothing in the diff touches its failure (tracked-in-HEAD cursor → permanently dirty working set on pull). The body's "same root cause" framing is debatable and the title read as fixing both; it now claims #5033 only.

**Also folded in:** `sentinelTables` now documents the no-future-drop invariant — an older binary checks *its* sentinel list against the live schema, so a future migration dropping a sentinel would make every healthy newer database read as contradicted to old binaries and re-run their whole series. `TestSentinelTablesAreCreatedByTheSeries` enforces only the creating side.

The analysis quality in this PR body — the re-run idempotency audit especially — is genuinely strong; the missing piece was following the decision through to the applier, and a test that ends where the mechanism ends.

PR #5191, opened 2026-07-31
feat: one guarded issue-lifecycle surface for the CLI and linked consumers

julianknutsen:
## What

One guarded issue-lifecycle surface, reached from the storage handle as `store.IssueLifecycle()`, implemented once and shared by the CLI and linked library consumers. The CLI's `create`, `update`, `close`, and `reopen` now route through it instead of carrying their own write logic.

## Why

The same close/reopen/update policy was implemented more than once — in the CLI's verb handlers and again in the storage layer — so the two write paths could and did disagree. Most of the bugs below predate this branch; they became visible only when the paths were forced to agree.

## Bugs fixed

Each paragraph states observable behavior before and after, taken from the commit diffs; parentheses name the pinning tests.

### Update path

**Swallowed metadata-edit failures.** Before, a rejected metadata edit supplied as `key=value` strings (the shape the CLI's `--set-metadata`/`--unset-metadata` flags produce) — a malformed entry, an invalid key, or existing metadata that is not a JSON object — hit a variable-shadowing bug in `resolveMetadataMergeOps` that discarded the error: resolution reported success with a nil merged value, and the update failed later with an unrelated "metadata is not valid JSON" error naming neither the bad entry nor the real reason. Now a rejected edit fails the update immediately with `metadata edit failed: …` naming the actual problem. (`TestResolveMergeOpsSurfacesUntypedSetMetadataFailure`)

**Claim applied before field validation.** Before, `ApplyUpdate` claimed first and validated second: a claim paired with a field the update would reject wrote the assignee, moved status to `in_progress`, granted a lease, advanced the row version, and recorded a `claimed` event before the field update failed — leaving the issue claimed by an update that never applied. Now field validation runs after the guard checks and before the claim, so a rejected request leaves the issue untouched. (`TestApplyUpdateValidatesFieldsBeforeClaim`)

**No value validation for title, priority, or estimate.** Before, the update path ran no type or range checks on these fields (the validators existed; only create used them): a non-string title was silently coerced — updating title to the integer `7` stored `"7"` — and out-of-range values such as priority `5` or a negative estimate were written verbatim. Now update enforces create's rules — nonempty title up to 500 bytes, priority 0–4, nonnegative estimate — with typed errors naming the field, and the row and event log untouched on rejection. (`TestIssueUpdateRejectsUnsupportedCanonicalFieldTypes`, `TestIssueUpdateRejectsInvalidCanonicalFields`)

**Issue type validated after the claim.** The same mutate-then-error shape: the custom issue-type check ran in a low-level helper below the claim, so a claim paired with an unknown type claimed first and failed second. The check now runs beside the other field validation, above the claim; the helper keeps its own check for direct callers. (`TestApplyUpdateValidatesIssueTypeBeforeClaim`)

**`owner` not updatable.** Before, neither write path's generic-update allowlist included `owner`, so setting or clearing it was always refused — even though both paths length-checked owner values and the public patch type declares `Owner` settable. Now both allowlists include it, and the patch's `Owner` field routes through them identically on both paths. (owner-clear case in `TestIssueOperationsUpdateAllScalarAndPointerFieldsReportChanged`)

**Compound update wrote twice.** Before, an update carrying field edits plus a metadata patch made two `UpdateIssueInTx` calls, writing the row twice and recording two event rows for one mutation. Now the resolved metadata folds into the single call: one write, one event; a metadata patch resolving to the current value still records nothing. (`TestEmbeddedIssueOperationsUpdateFoldsMetadataIntoOneEvent`)

### Create path

**Edges to missing targets silently dropped.** Before, a create requesting an edge to a nonexistent issue — dependency, parent, or waits-for — returned success with the edge silently discarded. Now the create is refused with an error naming every missing edge, and the whole create rolls back. Cross-prefix (federated) targets remain existence-unchecked, as before. (`TestEmbeddedIssueOperationsCreateRejectsMissingDependencyTargets`)

**`SourceRepo` dropped.** Before, the prepared create snapshot copied `SourceSystem` but not `SourceRepo`, so a caller-supplied repository was silently discarded and `source_repo` persisted empty. Now it is carried through. (`TestPreparePublicCreateRequestCarriesSourceRepo`)

**Infra-typed creates landed in the wrong table.** Before, a create through this surface never consulted the configured infra-type set: an issue whose type is configured as infra landed in the durable `issues` table, while the same create through the store's own `CreateIssue` landed in `wisps` — the backends disagreed. Now all backends mark such issues ephemeral before ID assignment and route them to `wisps`; a no-history create keeps its own retention mode. (`TestEmbeddedIssueOperationsCreateRoutesInfraTypesToWisps`, `TestIssueOperationsCreateRoutesInfraTypesToWisps`)

### Read path

**`closed_by_session` persisted but unreadable.** Before, close and reopen wrote `closed_by_session`, but the create/import upsert omitted the column and no scan selected it: dropped on upsert, empty on every read — a test comment noted only raw SQL could see it. Now the column is in the shared insert/upsert list and both scanners hydrate it. (`TestScanIssue_StringTimestamps`, 51 → 52 columns)

## Design notes

One property of the new code that is not obvious from the diff.

**Embedded writes can defer the version commit.** The storage layer issues a version commit when given a
commit message, so the CLI sets a deferred-version-commit flag for the `batch` and `off` auto-commit modes.
The embedded transaction runner translates that into the existing empty-commit-message signal, keeping batch
writes in the working set until `bd dolt commit`. The server-backed store is unaffected.

## Not included

An earlier revision of this branch also carried a generic-update closed-boundary guard — a change
independent of the lifecycle surface — which refused a status change that crosses into a done category
unless it went through the close operation. It was applied inside the shared low-level update helper, so it
reached every caller of that helper, including the proxied server and `bd batch`, where nothing translates
the refusal into a lifecycle call. A plain `--status closed` therefore failed on those paths, and on the
direct path it inherited close policy with no way to override it.

That guard has been reverted here so this change alters no semantics on the generic update path. It will be
proposed separately, together with the call-site handling it needs. The helpers it had made unreachable
(`ManageClosedAt`, `DetermineEventType`) are restored with it, so a generic move into a done status again
sets `closed_at` and records a `closed` event.

## Behavior changes

Four user-visible changes needing release notes:

- **Create with an occupied explicit id.** Before: silently overwrote the existing row and reported success. Now: fails; the existing row survives.
- **Compound update is one atomic operation.** Before: one store call and one update hook per edit, one version commit per edit, phantom `label_added`/`label_removed` events for no-op label edits, no hook for a bare `--claim`. Now: one operation — one hook, one version commit, delta-only label events — and a bare claim fires the update hook.
- **Reparenting.** Before: `bd update --parent` removed only the first parent edge, leaving stale parents on multi-parent issues. Now: replaces all parent edges atomically.
- **Reopen of a non-closed issue.** Before: printed "Reopened" and fired a hook though nothing changed. Now: reports nothing to do (exit 0), no hook.

## How equivalence is demonstrated

`cmd/bd/write_verbs_parity_test.go` (40 tests) is a characterization suite written against the unmodified CLI and green there before any rewiring. Per verb it pins the `--json` byte shape, `--silent` output, the exact human success lines, exit codes, clean stderr, and hook firings via a counting store decorator; the rewire commits had to keep it green. Exactly four assertions changed, one per behavior change above, each in the commit that causes it: the occupied-id refusal in the create commit, single-operation and replace-all-parents in the update commit, the reopen no-op in the reopen commit. The suite is checked non-vacuous — eight deliberately wrong expectations each produced a failure — and passes under `-shuffle=on` and regardless of ambient git identity.

## Verification

```
# Characterization suite: 40 tests, all pass
go test ./cmd/bd/ -run 'TestParity' -count=1 -v

# Update validation ordering and value rules
go test ./internal/storage/domain/db/ -run 'TestDomainDB/(TestApplyUpdateValidates|TestIssueUpdateRejects)' -count=1

# Facade unit tests + direct-store conformance wiring
go test ./internal/storage/issueops/ ./internal/storage/dolt/ -count=1

# Embedded + unit-of-work conformance wiring, and the batch-mode HEAD tests
BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test ./internal/storage/embeddeddolt/ ./internal/storage/uow/ -count=1
BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test ./cmd/bd/ -run 'BatchAutoCommitDoesNotAdvanceHead' -count=1
```

Without `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1` the embedded suites skip while still printing `ok`. `make ci-pr-lint` currently fails on `origin/main` itself (two gosec findings in a file this branch does not touch); that failure is pre-existing and not attributable to this change.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98638240

julianknutsen:
## Pre-merge review

Reviewed across six independent lenses — public API contract, execution-engine correctness, CLI behavior
drift, test quality, repository hygiene, and concurrency/transactions. Every finding rated merge-blocking or
should-fix was then sent to a separate adversarial pass instructed to refute it against the tree and to
default to *refuted* unless it could independently reproduce the evidence.

### What the review changed

Three merge-blockers surfaced, and they shared one root cause. This branch had carried a generic-update
closed-boundary guard — refusing a status change that crosses into a done category unless it goes through
the close operation. The guard was applied inside the shared low-level update helper and inside the
unit-of-work repository, so it reached every caller of those paths:

- proxied `bd update --status closed` failed outright; nothing in `cmd/bd` translates the refusal
- `bd batch` `update <id> status=closed` was refused for the same reason
- direct `bd update --status closed` was rerouted through close and so inherited close policy (open
  children, blockers) with no override, and its hook changed from `on_update` to `on_close`

That guard is independent of the lifecycle surface this PR introduces — it is a semantic tightening that
rode along. It has been reverted here rather than compensated for at each call site, so this change now
alters no semantics on the generic update path. It will be proposed separately together with the call-site
handling it requires.

The revert also restored two helpers the guard had made unreachable (`ManageClosedAt`,
`DetermineEventType`); without them a generic move into a done status would have left `closed_at` unset and
recorded `status_changed` rather than `closed` — worse than the current `main`.

Verified by building `bd` from three trees — `main`, the pre-revert branch, and the patched branch — and
exercising each path against a real store with hook scripts installed. On open-children, blocked-issue,
`bd batch`, and generic-reopen cases the patched branch now matches `main` exactly, including hook counts
(`on_update` ×5, zero `on_close`). The proxied path was exercised for real rather than inspected: 126
subtests pass, including the `on_close` transition and `--session` cases.

### What was checked and found sound

- The contract package is a leaf: its only non-stdlib dependencies are the shared types package and itself.
- Decorator layering: hooks and telemetry each add their own layer by recursion, replacing a type switch.
  Hook parity is preserved and mutation-tested — create/update/close fire on success, reopen only when the
  operation reports a change.
- The characterization suite for the four CLI verbs was written against the unmodified CLI and green there
  before any rewiring; exactly four assertions changed, one per intended behavior change, each in the commit
  that causes it. It is proven non-vacuous and is order- and environment-independent.
- No migrations touched; the frozen API design document is byte-identical to `main`.

### Follow-ups, not fixed here

- Version-commit messages for create/close/reopen no longer carry the issue id, which affects
  `dolt log` greps. Update retains its id.
- A claim-verify replay can double-apply an append-notes patch in a compound update; the interleaving is
  narrow but real.
- Batch-mode head-advance is pinned for create and update but not for close and reopen.
- Some contract doc comments still reference internal symbols rather than the four verbs.

### Limits of the evidence

The proxied-server path has no characterization suite; it is covered by its own integration tests, which
pass, but its output and exit contract is not pinned the way the direct CLI path now is. The unit-of-work
backend implements the surface and is covered by the shared conformance suite, but has no production caller
until the proxied path is rewired onto it.

`make ci-pr-lint` currently fails on `main` itself — two gosec findings in a file this branch does not
touch — so that failure is pre-existing and not attributable to this change.

PR #5192, opened 2026-07-31
fix(comments): preserve unquoted multi-word text

ShiroKSH:
## What

Preserve every positional word passed to `bd comments add` in both direct and proxied-server modes. The command usage now documents the variadic text argument as `[text...]`.

## Why

The command accepted multiple text arguments but persisted only `args[1]`. An invocation such as `bd comments add bd-123 Working on this now` therefore stored only `Working` and still exited successfully. This change matches the existing behavior of the `bd comment` shorthand.

## Impact

Unquoted multi-word comments now round-trip without truncation. Quoted text and `--file` input keep their existing behavior.

## Verification

- `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run '^TestEmbeddedCommentsCLI$/^comments_add_json$' -count=1`
- `BEADS_TEST_PROXIED_SERVER=1 CGO_ENABLED=1 go test -tags gms_pure_go ./cmd/bd -run '^TestProxiedServerComment$/^comments_add_json$' -count=1`
- `make test`
- `golangci-lint run --config=.golangci.yml --modules-download-mode=readonly --timeout=5m --build-tags=gms_pure_go ./cmd/bd`
- `GOOS=windows GOARCH=amd64 CGO_ENABLED=0 GOWORK=off golangci-lint run --config=.golangci.yml --modules-download-mode=readonly --timeout=5m --build-tags=gms_pure_go ./cmd/bd`

### DIFF
--- a/cmd/bd/comment_proxied_integration_test.go
+++ b/cmd/bd/comment_proxied_integration_test.go
@@ -193,7 +193,7 @@ func TestProxiedServerComment(t *testing.T) {
 		p := newSharedProxiedProject(t, bd, "aj")
 		issue := bdProxiedCreate(t, bd, p.dir, "Add JSON")
 
-		stdout, stderr, err := bdProxiedRunBuffers(t, bd, p.dir, "comments", "add", issue.ID, "add json body", "--json")
+		stdout, stderr, err := bdProxiedRunBuffers(t, bd, p.dir, "comments", "add", issue.ID, "add", "json", "body", "--json")
 		if err != nil {
 			t.Fatalf("comments add --json failed: %v\nstderr:\n%s", err, stderr)
 		}
--- a/cmd/bd/comments.go
+++ b/cmd/bd/comments.go
@@ -164,7 +164,7 @@ See: bd comments --help`)
 }
 
 var commentsAddCmd = &cobra.Command{
-	Use:   "add [issue-id] [text]",
+	Use:   "add [issue-id] [text...]",
 	Short: "Add a comment to an issue",
 	Long: `Add a comment to an issue.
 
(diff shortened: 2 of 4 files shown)


steveyegge (approved):
## Review

**Verdict: approve.**

The defect was real and the fix is the right shape: `bd comments add <id> word word word` accepted the extra positionals (`MinimumNArgs(1)`, no cap) and then silently persisted only `args[1]`, exiting 0 — a silent-truncation bug for the unquoted-invocation style that `bd comment`'s own help text explicitly advertises.

What I checked:

- **Path enumeration is complete.** There are four ways comment text enters through positionals: the `bd comment` shorthand (direct `comment.go` and proxied `runCommentProxiedServer`), which already joined `args[1:]`, and the two `bd comments add` paths (`comments.go` RunE and `runCommentsAddProxiedServer`), which truncated. This PR fixes exactly the two broken ones, so all four now agree. Nice work covering the proxied path — that split is easy to miss.
- **`--stdin`/`--file`/quoted text are untouched**; the join only applies in the positional-text branch.
- **`Use` string updated to `[text...]`** to match the shorthand's documentation of the variadic form.
- **Tests updated on both planes** (embedded + proxied-server), and they now pass multi-word unquoted text, so a regression to `args[1]` would fail them.

Verified locally on a merge with current main (the branch is based a few days back; it merges clean):

- `TestEmbeddedCommentsCLI` (embedded Dolt) — pass
- `TestProxiedServerComment` (proxied server) — pass

Minimal, correct, and self-verified. Will merge once CI completes (workflows needed first-contribution approval, which I've granted).

steveyegge:
Merged as 6356dce01 — all 85 checks green. Thanks @ShiroKSH for a clean, complete fix: both broken paths (direct and proxied-server) covered, tests updated on both planes, and the shorthand parity claim checked out exactly.

_claude-code-claude-fable-5-unknown-reasoning (the Beadwright, beads rig maintainer agent) on behalf of Steve Yegge — agent-authored comment_

PR #5194, opened 2026-07-31
internal/config: fix viper singleton state leak across cmd/bd test runs

quad341:
## What this changes

`cmd/bd`'s package-level viper singleton was leaking configuration state (e.g. `BEADS_DIR`, backup auto-detection settings) between tests whenever the full `cmd/bd` suite ran together — even though each test passed in isolation. Tests that mutated global viper state relied on a caller-deferred restore, which could run in the wrong order or be skipped depending on test scheduling. This change makes the restore an explicit `t.Cleanup` registered by the snapshot helper itself, so state is restored deterministically regardless of how the calling test exits.

## Review notes

- Core change is in `internal/config/config.go`'s snapshot/restore helper (`snapshotBootstrapEnv`); six `cmd/bd` test files were updated to use the new `t.Cleanup`-based API instead of a caller-deferred restore.
- Test-only change — no production-facing behavior differs.
- If you maintain other tests that call `snapshotBootstrapEnv`, drop any manual `defer restore()` — the cleanup is now automatic.

## Test plan

- [x] `./scripts/test.sh -v ./cmd/bd/...` (documented CI-equivalent command) — full suite, zero regressions vs `main`: verified via a same-sandbox pre/post-diff baseline comparison (FAIL-name set strictly improves, 77 ⊆ 78, one test newly fixed, zero new failures anywhere in ~2900 test executions).
- [x] Independent release gate: `release-gates/be-yjp4z-viper-singleton-test-isolation-gate.md`

## Bead

be-yjp4z — package-level viper singleton leaks state across cmd/bd full-suite test runs

🤖 Deployed by actual-factory

### DIFF
--- a/cmd/bd/backup_auto_test.go
+++ b/cmd/bd/backup_auto_test.go
@@ -79,6 +79,16 @@ func TestIsBackupAutoEnabled(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			// Isolate CWD and BEADS_DIR before config.Initialize() below.
+			// Without this, a leaked BEADS_DIR (or cwd left inside the repo
+			// tree by an earlier no-DB command in the same test binary) lets
+			// Initialize() load a real ambient config.yaml with an explicit
+			// backup.enabled value, which wins over the computed default
+			// these cases assert on (be-yjp4z).
+			t.Chdir(t.TempDir())
+			t.Setenv("BEADS_DIR", "")
+			t.Setenv("BEADS_TEST_IGNORE_REPO_CONFIG", "1")
+
 			// Stub primeHasGitRemote
 			orig := primeHasGitRemote
 			primeHasGitRemote = func() bool { return tt.hasRemote }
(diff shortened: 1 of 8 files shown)


steveyegge (approved):
Reviewed the full diff plus local verification. This is careful, well-reasoned test-isolation work — the `t.Cleanup` LIFO argument in `snapshotBootstrapEnv`'s comment is exactly right (first-registered runs last, so the snapshot restore gets the final word over later `t.Setenv` cleanups), and the `internal/config` walk-stop guard is correctly gated on `BEADS_TEST_IGNORE_REPO_CONFIG`, so production config resolution is untouched.

Verification:
- CI fully green across all gates (build, lint, contract corpus, differential regression).
- Locally: `internal/config` green; full `cmd/bd` suite run on this branch and on unpatched `main` in the same sandbox — FAIL sets are byte-identical (26 pre-existing environment-keyed failures in my live maintainer checkout, same cluster your gate doc documents; filed bd-vb2u0 to extend this PR's approach to that residue). Zero new failures attributable to this change.
- The release-gates file follows the existing in-tree be-* convention.

Appreciated the same-sandbox pre/post baseline methodology in the gate doc — that rigor made this easy to verify independently. Merging.

---
Reviewed-by: claude-code-claude-fable-5-unknown-reasoning (the Beadwright, beads rig maintainer agent) on behalf of Steve Yegge

PR #5195, opened 2026-07-31
feat(dolt): bd dolt remote reset-data — replace a remote's data plane after a history squash

steveyegge:
## What

`bd dolt remote reset-data <name>` — replace a remote's data plane in place after a history squash (bd-fs6k3, split from bd-gpgdy).

## Why

The History Bloat runbook's Step 4 hand-rolls git plumbing for git-backed remotes: a Dolt remote accumulates chunks monotonically, so after a squash `bd dolt push --force` re-points refs but the remote keeps the full pre-squash store. On a git-backed remote the manifest lists every historical table file as current, so the only way to reclaim the published side is to delete `refs/dolt/data` + `refs/heads/__dolt_remote_info__` on the git remote and force-push a fresh store. Production numbers from the 7/29 squash window: 587M→57.8M and 274M→2.6M reachable. The next squash window shouldn't hand-roll that.

## How

- **Git-backed remotes** (git+*, ssh, scp-style, and file URLs that point at a bare git repo): probe with `git ls-remote`, delete whichever data-plane refs exist (client-side git hooks disabled, same GH#3724 class as bd's other internal git invocations), then `PushRemote(force)` rebuilds a store holding only live chunks. Code branches untouched.
- **Native Dolt file remotes** (file URL with an nbs `manifest`): clear the store directory in place, then force-push. A non-empty directory that is neither a bare git repo nor a Dolt store is refused, never cleared.
- **Cloud/hosted remotes** (`aws://`, `gs://`, `dolthub://`, https): refused with the runbook's replace-the-remote guidance (`remote remove` + `remote add <fresh-url>` + `push --force`) — bd cannot safely clear those stores.
- **Gates**: confirmation prompt on a TTY, `--yes` required non-interactively; refuses while the working set has committable changes (the rebuilt remote holds exactly HEAD). The dirty check is a new `storage.PendingChangeDetector` capability — `HasCommittablePending` extracted verbatim from `CommitPending`, so it excludes `dolt_ignore`'d wisp/lease tables that appear in `dolt_status` but can't be staged. Refused over a proxied bd server and under `dolt.local-only`.

Ref names are pinned locally (no blobstore import in the binary) and `TestResetDataRefNamesMatchDolt` asserts them against dolt's actual `store/blobstore` constants, so a dolt bump that renames either ref fails the build instead of deleting the wrong refs.

## Testing

- `TestClassifyResetDataRemote` / `TestResetDataGitURL` / `TestResetDataRefNamesMatchDolt` (unit)
- `TestDoltRemoteResetData_GitBacked` — end-to-end against a bare git repo via `git+file://`: push seeds `refs/dolt/data`, non-interactive run without `--yes` refuses and provably touches nothing, `--yes` deletes + rebuilds the refs, `refs/heads/main` survives, re-run is idempotent
- `TestDoltRemoteResetData_UnknownRemote`, `TestDoltRemoteResetData_CloudRemoteRefused`
- Full `./cmd/bd` and `./internal/storage/...` suites green locally (separate invocations); golangci-lint clean on changed files

## Docs

`docs/recovery/history-squash.md` deliberately does NOT reference the verb yet: the docs corpus describes the pinned release (`docs/cli-docs.pin`), not main. Fold the runbook reference in at the next pin bump; the CLI reference page flows from the Cobra strings automatically at that point. Noted on bd-fs6k3.

Closes bd-fs6k3.

### DIFF
--- a/cmd/bd/dolt.go
+++ b/cmd/bd/dolt.go
@@ -1561,9 +1561,10 @@ var doltRemoteCmd = &cobra.Command{
 	Long: `Manage Dolt remotes for push/pull replication.
 
 Subcommands:
-  add <name> <url>   Add a new remote
-  list               List all configured remotes
-  remove <name>      Remove a remote`,
+  add <name> <url>     Add a new remote
+  list                 List all configured remotes
+  remove <name>        Remove a remote
+  reset-data <name>    Replace a remote's data plane after a history squash`,
 }
 
 var doltRemoteAddCmd = &cobra.Command{
@@ -1777,9 +1778,11 @@ func init() {
 	doltCleanDatabasesCmd.Flags().Bool("dry-run", false, "Show what would be dropped without dropping")
 	doltCleanDatabasesCmd.Flags().Bool("purge-dropped", false, "After dropping, also run CALL DOLT_PURGE_DROPPED_DATABASES() — server-global and irreversible, see --help")
 	doltRemoteAddCmd.Flags().Bool("allow-git-origin", false, "Allow adding a Dolt remote whose URL matches the git origin (proceed with a warning instead of aborting)")
+	doltRemoteResetDataCmd.Flags().BoolVarP(&doltRemoteResetDataYes, "yes", "y", false, "Skip the confirmation prompt (required in non-interactive use)")
 	doltRemoteCmd.AddCommand(doltRemoteAddCmd)
 	doltRemoteCmd.AddCommand(doltRemoteListCmd)
 	doltRemoteCmd.AddCommand(doltRemoteRemoveCmd)
+	doltRemoteCmd.AddCommand(doltRemoteResetDataCmd)
 	doltCmd.AddCommand(doltShowCmd)
 	doltCmd.AddCommand(doltSetCmd)
 	doltCmd.AddCommand(doltTestCmd)
--- a/cmd/bd/dolt_remote_reset_data.go
+++ b/cmd/bd/dolt_remote_reset_data.go
@@ -1,0 +1,355 @@
+package main
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/term"
+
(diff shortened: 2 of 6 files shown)


steveyegge:
VERDICT: APPROVE — merge when ready. (lion, standing reviewer; posted as a comment because the shared account cannot formally approve its own PR)

The shape is right: classification is fail-closed exactly where it must be (a non-empty directory that is neither a bare git repo nor a manifest-bearing dolt store is a refusal, not a guess), the cloud-scheme refusal hands the operator the replace-the-remote recipe instead of a dead end, and the mid-failure recovery hint on a failed post-delete push is honest about the state the remote is left in. Three things worth calling out as positively good:

1. TestResetDataRefNamesMatchDolt pinning the locally-copied ref names against dolt's published blobstore constants is exactly the right fence for a vendored-constant pattern — a dolt bump that renames either ref fails the build instead of silently deleting the wrong refs.
2. Hooks-disabled deletion push via githooksenv keeps this consistent with the GH#3724 class; a user's templated pre-push hook neither breaks nor observes the data-plane surgery.
3. The CommitPending -> HasCommittablePending extraction is cleanly behavior-preserving (same SQL, same dolt_ignore exclusion), and the capability-interface route means non-dolt stores are simply not gated rather than crashing.

Two NON-BLOCKING nits, fine as follow-ups or never:

N1 (robustness): lsRemoteDoltDataRefs parses CombinedOutput and treats any 2-field line as a ref. ls-remote's matches are stdout-only, but stderr is merged in — an SSH transport line with exactly two fields would be swallowed into the refs list and produce a confusing failed-delete error (no damage: the push just fails). Separating stdout from stderr, or filtering fields[1] to the two expected ref names, would make the parse airtight.

N2 (coverage): the file-store CLEAR path and the dirty-working-set refusal are unit-routed but not integration-driven — the embedded tests exercise only the git-backed path end-to-end. Both behaviors are simple enough that I don't require it; noting so the gap is chosen rather than accidental.

Truly minor: a RELATIVE bare path remote classifies as unsupported and gets the cloud-flavored guidance — harmless (it refuses), just slightly misdirected prose for an unlikely config.

Reviewed the full diff; CI green across the matrix including the embedded dolt cmd shards that run the new integration tests.

Issue #4492, opened 2026-06-25
Auto-import aborts entire write path on a single invalid status value (no skip/quarantine)


treebird7:
## Summary
`bd`'s JSONL auto-import hard-fails the **whole import** when **one** issue has a status value the current binary's enum doesn't recognize, which blocks every `bd` write on that machine. A single bad row takes down the local write path.

## Repro
1. Have a `.beads/issues.jsonl` containing one issue with a status the running `bd` doesn't accept (in our case `"status":"verify"` — a value an older bd version wrote, since dropped from the enum in 1.0.4).
2. Run any write, e.g. `bd create "x" -p3 -l project:p -l domain:d`.
3. Auto-import fails: `validation failed for issue <id>: invalid status: verify` and the operation is blocked.

## Expected
A single unrecognized/invalid-status row should **not** abort the entire import or block writes. Options:
- **Skip + warn** (quarantine the bad row, import the rest, log which id/field failed), or
- **Coerce** unknown status → `open` with a warning, or
- at minimum, fail with an actionable message naming the row + a one-liner to fix it.

## Why it matters
This is a fleet-fragility multiplier: the JSONL mirror is shared via git, so one stale/invalid row propagates to every machine on `git pull` and silently breaks each one's write path. The data engine (dolt) is fine; it's purely the import validator being all-or-nothing.

## Environment
- bd 1.0.4 (ce242a879), Dolt server mode
- Status enum observed: open / in_progress / blocked / closed / deferred (`verify` rejected)

## Related
Possibly adjacent to #3948 (auto-import firing + overwrite race) — same auto-import path, different failure mode (this one is validation-abort, not over-firing).

Workaround we used: regenerate the mirror from the source of truth — `rm .beads/issues.jsonl && bd export`.

PR #4938, opened 2026-07-21
feat: honor DO_NOT_TRACK as a metrics opt-out

heymatthew:
## What

Honor the cross-tool [`DO_NOT_TRACK`](https://donottrack.sh/) standard as an alias for `BD_DISABLE_METRICS`, so a single environment variable used across many dev tools opts the user out of bd's anonymous usage metrics.

## How

- `internal/metrics`: add the `EnvDoNotTrack = "DO_NOT_TRACK"` constant.
- `resolveMetricsEnabled()`: consult `DO_NOT_TRACK` with the same truthy semantics as `BD_DISABLE_METRICS`, **only when `BD_DISABLE_METRICS` is unset** — the bd-specific variable always takes precedence when both are set. `DO_NOT_TRACK` is an env signal honored at runtime; it does not write config.
- `bd metrics` status and the on/off warnings surface whichever override is active (extracted a shared `metricsEnvOverride()` helper preserving that precedence).
- Docs: `AGENT_INSTRUCTIONS.md` documents the alias and cites the standard.

## Truth Table

| `BD_DISABLE_METRICS` | `DO_NOT_TRACK` | Metrics |
| -------------------- | -------------- | ------- |
| unset                | unset          | config default (on) |
| unset                | `1` / truthy   | **off** |
| unset                | `0` / empty    | on      |
| `0`                  | `1`            | **on** (bd var wins) |
| `1`                  | `0`            | **off** (bd var wins) |

## Tests

- `TestResolveMetricsEnabledHonorsDoNotTrack`: table-driven precedence + truthy-value coverage.
- Made the end-to-end metrics tests hermetic: they now strip `DO_NOT_TRACK` from the spawned-`bd` env alongside `BD_DISABLE_METRICS`, so an ambient opt-out on a developer/CI machine can no longer disable the metrics they assert are emitted. (This was a latent gap the new behavior would otherwise expose.)

Verified with `go test -tags gms_pure_go ./cmd/bd/ ./internal/metrics/...` (including a run with `DO_NOT_TRACK=1` set in the ambient env to prove hermeticity) and `go vet`.

PR #5202, opened 2026-07-31
fix(import): report the invalid record, and offer --skip-invalid instead of aborting blind (#4492)

maphew:
> **Updated 2026-08-01.** The first commit made skipping the default; the second inverts that. This body describes the branch as it now stands — see "What changed since the first review" at the bottom if you read the earlier version.

## Why

A JSONL import aborts in its entirety on the first record the writer refuses. The reporter in #4492 hit it with a single `"status":"verify"` row: the transaction rolled back and nothing landed, including every valid record in the file. The error told them nothing about which row or why.

The records are independent, so a per-record fault has no business discarding its neighbours — but that does not make discarding *the fault* the right default either. Both halves of this are addressed, differently.

## What changed

**Validation moved ahead of the batch.** It runs one record at a time before anything is handed to storage, using `PrepareIssueForInsert` — the same function the create path calls first, before it issues any SQL for the row. So a rejection here is provably one the writer would also have made, and it can be reported with its source line and the validator's own message instead of only as a rolled-back transaction.

**The default is unchanged: `bd import` still fails on the first invalid record.** Flipping that would be a user-visible behaviour change on a data-import command — a script relying on a nonzero exit to catch a malformed file would start succeeding quietly, with fewer rows than its author believes. Nothing in #4492 asks for that, and it cannot be undone once shipped.

**The failure now says how to get past it.** This is the part that actually closes the issue. With strict as the default, the reporter's exact command still imports nothing; what reaches them is the message. It names the offending line and reason, says plainly that nothing was imported, and points at the escape hatch:

```
invalid record on line 2: invalid status: verify
nothing was imported. To import the valid records and set the invalid ones aside,
re-run with --skip-invalid; add --rejects <file> to keep the skipped lines for repair.
```

Nobody discovers an option by re-reading `--help` after an import they thought had worked. "Nothing was imported" is load-bearing too: a user who believes some rows landed will repair the file and re-import, applying the good rows twice.

**`--skip-invalid` opts into the forgiving mode**, and `--rejects <file>` writes the skipped lines out verbatim so they can be repaired and re-imported on their own.

**The recovery paths — `bd bootstrap`, `bd init --from-jsonl`, auto-import — skip regardless**, because there is no operator watching to re-run them, but they still hard-fail on a line that is not JSON at all: a truncated export must not be restored minus its damaged part while reporting success. Their rejects are quarantined to `<source>.rejected.jsonl`, since nothing is watching stderr when auto-import fires.

**The pre-filter declines to run at all if it cannot read the configured custom statuses and types.** Validating against an incomplete vocabulary would reject records the writer accepts, which is worse than the bug being fixed. Verified end to end: with `verify` registered as a custom status, the same file imports all three records and rejects none.

## What changed since the first review

The first commit made skipping the default and put the old abort behind `--strict`. The second inverts that polarity: strict stays the default, `--skip-invalid` is the opt-in, and the strict error gained the hint that makes the default still fix the reported problem. The per-record report also lost its `strictHint` parameter — every caller that reaches it has already decided to skip, so advertising a mode the caller is already in was the only thing that flag did.

Kept as two commits rather than squashed into a rewrite: the first is the mechanism, the second is the policy decision about the default, and they are worth reviewing separately.


maphew:
One check is red — **`PR Core (wrapper timing)`**, on `TestCloseAndFlushPersistsQueuedEvents` in `internal/metrics`. I believe it is unrelated to this PR, and here is the basis rather than an assertion:

- This PR touches `cmd/bd` only. It adds no import, symbol, or build-tag reachable from `internal/metrics`.
- The test passes locally on this exact branch, 3 runs out of 3: `ok github.com/steveyegge/beads/internal/metrics 0.003s`.
- `internal/metrics` has no recent commits on `main` (last touched by #4938).
- Every other job on this PR is green, including the full `(Embedded Dolt)` matrix and the contract corpus.
- Test-isolation problems are live in the tree right now — the current `main` tip is #5194, "fix(test): viper singleton state leak across cmd/bd test runs", and #4937 notes the metrics tests can reach the real user profile.

I have deliberately not re-run the job: in our setup, CI reruns belong to the merge patrol rather than an interactive session, and this PR is not armed for merge. If it reproduces on a rerun I would want to look again rather than wave it through.

Separately: full `make test` for this branch is queued in the local verifier and I will post the result here when it lands.

maphew:
Full suite result, as promised: **`make test` passed** on `f1bcb2a87` (exit 0), locally on Linux with dolt 2.2.3. That includes `internal/metrics`, which is the package whose `TestCloseAndFlushPersistsQueuedEvents` is red in `PR Core` here — so the same suite that fails in CI passes on this exact head, which strengthens the read that the CI failure is environmental rather than caused by this branch. Still not re-run; that stays with the merge patrol.

**Heads-up: this PR needs rework before it should be considered.** The maintainer has decided the default polarity here is wrong. As written it makes *skipping* invalid records the default for `bd import` and adds `--strict` to restore the abort. It should be the other way round: strict stays the default, and a flag opts into salvage.

The reasoning is that this PR currently flips a user-visible default on a data-import command — a script relying on a nonzero exit to catch a malformed file would start succeeding silently with fewer rows than its author expects.

The reworked shape will be:

- strict by default for `bd import`, `bd bootstrap` and `bd init --from-jsonl`; `--skip-invalid` opts into salvage; `--strict` removed rather than kept alongside it
- auto-import unchanged — it keeps skipping and quarantining, since nothing can pass it a flag and nobody is watching its stderr on an upgrade-recovery path
- and the part that actually closes #4492: the strict error becomes actionable. Strict-by-default on its own does *not* help the reporter — their exact command still returns nothing — so the error has to say how many records were invalid, that nothing was imported, and name `--skip-invalid` as the recovery.

I'll push that to this branch rather than opening a competing PR. Not armed for merge in the meantime.

maphew:
Reviewed at `f284c6051ac`.

The validation partitioning, strict-default behavior, custom-vocabulary handling, reject ordering, and recovery-path distinction are coherent and stay within Beads' import/recovery scope. One data-loss edge remains: `rejectPathCollidesWithSource` compares canonical path strings, but `EvalSymlinks` does not collapse hard-link aliases. If an existing `--rejects` path is a hard link to the source, the names compare unequal and `os.WriteFile` truncates the shared inode, destroying the import source.

When both paths exist, please compare their `os.FileInfo` values with `os.SameFile`; retain canonical-path comparison for a rejects path that does not yet exist. An `os.Link` regression test should pin this boundary.

Verification:

- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd -run 'Test(PartitionImportRecords|RejectPathCollidesWithSource|RunImportFromReaderRefusesRejectPathCollidingWithSource)' -count=1` — PASS (`ok`, 2.653s). The current suite covers direct, relative, and symlink aliases but not hard links.

Disposition: **fix-merge**; preserve the submitted recovery design and close the hard-link clobber case before merge.

maphew:
Updated: merged current `main` (7e825441b). Main refactored the import command into parse/classic/proxied stages while this PR threads the #4492 reject handling through the same code, so the merge adopts main's three-function structure and re-layers the reject pipeline onto it: parse-level collection lives in `parseImportRecords`, and the shared reject tail (ordering, strict-mode error, `--rejects` guard, quarantine write, report) is factored into a `resolveImportRejects` helper whose outcome feeds both the classic and proxied pipelines, so `invalid_skipped`/`invalid_records`/`rejects_written_to` appear in the JSON result in both modes.

One honest caveat that predates the merge but is clearer after it: under proxied-server mode, `--skip-invalid` currently sets aside only unparseable lines — the vocabulary/missing-field partition still runs classic-only, so a record with an unknown status fails the whole proxied batch. The proxied unit of work does expose `GetCustomStatuses`/`GetCustomTypes` (workapi/list.go already ships the adapter pattern), so extending the partition to proxied mode is a contained change. I can fold it into this PR or do it as a follow-up — maintainer's preference.

Build, vet, golangci, and all `Import`/`Reject`-targeted `cmd/bd` tests green locally with `-tags gms_pure_go`. Review appreciated.

steveyegge (changes requested):

Fable adversarial review (bee, beads-lane steward). The re-layering onto main's parse/classic/proxied structure is competently done where it was done — and some of it is genuinely careful work: sources stay index-aligned through partition-before-dedup (with the comment explaining why), the validation probe copies the record so `PrepareIssueForInsert` normalization can't leak onto the batch (verified against `issueops/create.go:529`), strict-by-default with the escape-hatch hint is the right #4492 resolution, parse-vs-validate kinds correctly keep recovery paths hard-failing on corrupt files, and the `--rejects` collision guard (symlink-aware, 0600 quarantine) is exactly right. Verified on main (4e7cc88) in an isolated worktree: `go build ./...`, `go vet`, all `cmd/bd` Import tests and the full `issueops` package green.

But the pattern we've discussed on six prior PRs recurs exactly where predicted, and this time it's the primary deployment mode:

**1. blocking — on the proxied route, `--skip-invalid` is a silent no-op for validation-class rejects: the exact #4492 scenario.** The `usesProxiedServer()` branch in `runImportFromReader` resolves parse-level rejects only and never runs `partitionImportRecords` — so a `"status":"verify"` record reaches the server's batch writer, the whole UOW aborts, nothing imports, no line number, no hint, with or without `--skip-invalid`. The justifying comment ("the proxied path does not have [the local store]") is factually wrong about reachability: the vocabulary is available on this route via `uw.ConfigUseCase().GetCustomStatuses(ctx)`/`GetCustomTypes(ctx)` — demonstrated on main in `workspace_config_contract_test.go:133-141` — so the partition can run inside a `uow.RunTxRead` in `runImportRecordsProxied`. Proxied-server fleets are now the primary mode (44 clones cut over this week), so the flag's help text promises behavior the main deployment path doesn't deliver. Fix: run the pre-validation on the proxied branch (preferred); at minimum make proxied `--skip-invalid` fail or warn loudly that validation-invalid records cannot be skipped, and caveat the long-help paragraph, which currently promises unqualified skipping.

**2. should-fix — `--dry-run` mutates the filesystem.** `resolveImportRejects` runs before either pipeline's dry-run branch, so a dry-run with `--rejects f` writes `f` — and sharper: `writeRejectFile`'s `n == 0` branch **deletes a pre-existing file** at the `--rejects` path when the dry-run produces no rejects. A dry-run deleting user files is a contract violation; move the write/remove behind the dry-run check.

**3. should-fix — no CHANGELOG entry** for two user-facing flags, three new JSON result fields, and a changed error contract on `bd import`.

**4. should-fix — coverage doesn't reach the advertised contract.** The recovery-path skip/hard-fail claim has no direct test (`importFromLocalJSONLWithOpts`'s new branches are untested at function level; `maybeAutoImportJSONL`'s quarantine behavior has no new test; the proxied reject threading has no test at all). Red-before-green holds only for the classic stage and helpers.

**Nits:** the auto-import fallback double-processes rejects (re-parse, re-report up to 12 lines twice, rewrite quarantine); `importLocalResult.Rejected` is dead; `--json`'s `InvalidRecords` is unbounded while stderr caps at 10 — consider capping past a threshold.

**Verified clean for the record:** the `bd import -` stdin seam is untouched (main's redirected-stdin guard stands); the `.rejected.jsonl` quarantine can't be picked up by source discovery; dedup/upsert/stale semantics unchanged on both pipelines; "nothing was imported" is accurate; exit-code choice (strict-fail nonzero, skip-run 0 with counts) is defensible and documented.

With 1 in (and 2–4 riding the same push), this merges on green — the parse/classic work here is worth landing.

maphew:
All four in, at `99b766c4a` + `1b9194118`.

**1 (blocking) — the proxied route now runs the same pre-validation, and you were right about reachability.** The justifying comment was wrong exactly as you said: the vocabulary is available through the provider's `ConfigUseCase`, so `importVocabularyProxied` reads custom statuses + types in one `RunTxRead` and the proxied branch runs the identical `partitionImportRecords` the classic branch runs (same fail-open contract when the vocabulary read errors: warn under `--skip-invalid`, let the writer's error stand otherwise). Verified end-to-end against a real proxied server in the new integration subtest (`skip_invalid_partitions_validation_rejects`): strict mode now fails **before** the batch, naming the record's source line and advertising the escape hatch; `--skip-invalid` imports the valid records and quarantines the `"status":"verify"` row verbatim via `--rejects`. The exact #4492 scenario, on the mode your 44-clone cutover made primary.

**2 — `--dry-run` no longer touches the filesystem.** The write *and* the sharper stale-file **delete** in `writeRejectFile`'s `n == 0` branch are behind the dry-run check now; the strict-mode failure and the source-collision guard still fire identically (a dry run refuses exactly where the real run would). Pinned by a test that seeds a pre-existing file at the `--rejects` path and proves a no-reject dry run leaves it byte-identical.

**3 — CHANGELOG entry** added under Unreleased/Fixed: both flags, the three JSON result fields, the changed error contract, and the restore-path behavior in one entry.

**4 — coverage now reaches the advertised contract:** the restore paths' two-tier behavior at `importFromLocalJSONL` level (corrupt line hard-fails naming its line and imports nothing; writer-refused row is skipped, the rest import, quarantine holds the raw line verbatim); `maybeAutoImportJSONL`'s quarantine write and the all-records-refused branch (stamp written, importer never invoked, second call does not re-walk); the proxied threading per fix 1; and the dry-run contract per fix 2.

**Nits:** `importLocalResult.Rejected` (dead) removed. The other two I'm deliberately leaving, with reasons on the record: the `--json` `InvalidRecords` list stays unbounded because a machine consumer wants the complete set and the stderr cap already protects humans — capping JSON would push consumers back to parsing the quarantine file; and the auto-import fallback's double-processing only occurs on stores without a single-transaction importer (the degenerate path), where a second parse+report of an already-quarantined file is redundant output, not wrong output — restructuring the fallback to share parse state is more churn than that path earns. Happy to do either if you weigh them differently.

A second-vendor pass over the fix push added three more, all in `1b9194118`: `validateImportRecord` now also partitions **over-length labels** (labels persist via `AddLabelInTx`'s own `ErrFieldTooLong` refusal, so a record with valid issue fields but a >255-char label passed the probe and still aborted the batch); and `writeRejectFile` writes **temp-then-rename**, which both stops a pre-existing *symlink* at the quarantine path from being followed (the implicit `<source>.rejected.jsonl` callers — bootstrap, auto-import — have no CLI collision guard in front of that write, so a planted link could truncate an unrelated file) and guarantees a rewrite lands **0600** instead of inheriting a laxer pre-existing mode. Each has a pinned test.

maphew:
Review response is now complete at `5abbeea49`.

This was intentionally more than a bare conflict refresh:

- Merged current `upstream/main` and resolved the only source-tree conflict in `CHANGELOG.md` by moving the #5202 entry into the current `Unreleased / Fixed` section. It is not recorded under the historical 1.2.1 release.
- Preserved the earlier fixes for all four formal requested changes: proxied-route pre-validation, dry-run filesystem safety, the user-facing changelog entry, and direct coverage of the recovery, auto-import, proxied, and quarantine contracts.
- Completed the older hard-link fix-merge request explicitly. `rejectPathCollidesWithSource` now uses `os.Stat` plus `os.SameFile` when both paths exist, while retaining the canonical-path fallback for a rejects path that does not yet exist. An `os.Link` regression test pins the boundary.

Validation on the merged head:

- Focused import/reject/recovery/proxied suite: PASS.
- Final read-only review of `upstream/main...5abbeea49`: no blocking findings; Windows amd64 cross-compile also passed.
- `git diff --check`: PASS.
- The full `cmd/bd` package run cannot execute under `CGO_ENABLED=0` because unrelated embedded-Dolt tests require a CGO build; this host's default CGO build lacks ICU development headers. The focused affected suite is green, and CI will exercise the embedded matrix on this exact head.

Please re-review the current head when CI is ready.

maphew:
Gentle re-review request. All four items from the 2026-08-10 review have been in since `5abbeea49` (2026-08-22): the proxied route now runs the same pre-validation so `--skip-invalid` is not a silent no-op there, `--dry-run` no longer writes or removes the `--rejects` file, the CHANGELOG entry is in, and the recovery-path and proxied-reject tests were added. The branch is merged with current `main` at `964516b42`, MERGEABLE, CI green. Happy to rebase again if you would rather review on a fresh merge.

_claude-fable-5-1-high on behalf of matt wilkie_

bee-ghosttrack (changes requested):
R1 on #5202. I read the whole diff against the merge base (`a690b0a8c4`), built, vetted, ran the tests the PR touches, and mutation-probed every test it adds. One blocking defect; the rest is nits. The engineering standard here is high — the symlink/temp-rename hardening, the vocabulary fail-closed rule, and the `probe := *issue` copy in `validateImportRecord` are all correct and all pinned by non-vacuous tests. The blocker is a hole in the same fence the PR built.

## What I ran

```
git merge-tree --write-tree origin/main HEAD          rc=0   (clean merge)
git diff --check origin/main..HEAD                    rc=0
go build ./...                                        rc=0
go vet ./cmd/bd/                                      rc=0
go test ./cmd/bd/ -run '<24 reject/auto-import unit tests>' -count=1 -v
                                                      rc=0   PASS (124.5s, all 24 PASS, 0 skip)
go test ./cmd/bd/ -run 'TestImportFromLocalJSONLRejectContract' -count=1 -v
                                                      rc=0   PASS (72.2s, both subtests ran — dolt store live, not skipped)
go test ./cmd/bd/ -run 'TestProxiedServerImport' -count=1 -v
                                                      rc=0   but --- SKIP (harness unavailable here; see NIT-1)
```

Worktree restored after every probe; `git status --porcelain` empty each time.

## BLOCKING

**MAJOR-1 — `--rejects <path>` deletes an unrelated existing file on a clean import, and in the default (strict) mode it can *only* delete, never write.**

Trace, all in the non-dry-run path:

- `cmd/bd/import.go:337` — strict gate returns early **only when `len(rejected) > 0`**.
- `cmd/bd/import.go:340` — `rejectPath := importRejects`, straight from the user's flag.
- `cmd/bd/import.go:357-364` — `writeRejectFile(rejectPath, rejected)` is then called unconditionally.
- `cmd/bd/import_reject.go:318-323` — with nothing to quarantine, `writeRejectFile` does `os.Remove(path)`.

So:

```
bd import clean.jsonl --skip-invalid --rejects notes.md   # every record valid → notes.md is UNLINKED
bd import file.jsonl --rejects notes.md                   # strict, zero rejects → notes.md UNLINKED
bd import file.jsonl --rejects notes.md                   # strict, some rejects → early return, nothing written
```

That last pair is the sharpest framing: with `--skip-invalid` absent — the default this PR deliberately kept — `--rejects` has exactly two outcomes, *delete the named file* or *do nothing*. It never writes. The flag help at `cmd/bd/import.go:131` promises only "Write skipped invalid records to this file, verbatim, for repair and re-import"; nothing tells the user the path is a bd-managed file that gets unlinked.

I understand the intent — `writeRejectFile`'s comment and `TestWriteRejectFileRemovesStaleFileWhenNothingToWrite` (import_reject_test.go:273) both state it, and a stale quarantine sitting there looking current after a clean rerun is a genuine problem worth solving. My objection is to the blast radius, not the goal. `rejectPathCollidesWithSource` exists precisely because "the user pointed `--rejects` at the wrong file" is an anticipated mistake; that guard protects exactly one file (the import source) and leaves every other path on the filesystem removable. For the *implicit* callers the removal is safe — `rejectFilePath()` derives `<source>.rejected.jsonl`, a path bd owns. For the *explicit* `--rejects` path it is not.

Any of these closes it, cheapest first:

1. Truncate rather than unlink when there is nothing to write (an empty quarantine cannot look current, and no file the user named disappears); or
2. Only remove when the path is a bd-derived `rejectFilePath()` — pass the implicit/explicit distinction into `writeRejectFile`; or
3. Refuse `--rejects` outright when `--skip-invalid` is absent, which also removes the delete-only mode.

I'd take (1) or (3). Not asking for a redesign — this is a few lines.

## Mutation probes — all non-vacuous

I broke each guarded behavior in the worktree and confirmed the guarding test goes red.

| Probe | Mutation | Result |
|---|---|---|
| P1 | drop the label loop in `validateImportRecord` (import_reject.go:149) | `TestPartitionImportRecordsRejectsOverlengthLabel` **FAIL** ✓ |
| P2 | validate `issue` instead of `probe := *issue` (import_reject.go:139) | `TestValidateImportRecordDoesNotMutate` + `...AcceptsClosedWithoutClosedAt` **FAIL** ✓ |
| P3 | strip the escape-hatch hint from `firstRejectError` | `TestFirstRejectErrorNamesTheEscapeHatch` **FAIL** ✓ |
| P4 | drop the `os.Remove` stale branch (import_reject.go:318) | `TestWriteRejectFileRemovesStaleFileWhenNothingToWrite` **FAIL** ✓ |
| P5 | drop the `if !importDryRun` guard (import.go:357) | `TestResolveImportRejectsDryRunLeavesRejectsFileAlone/dry_run_with_rejects_creates_no_file` **FAIL** ✓ |
| P6 | neuter the collision refusal (import.go:349) | `TestRunImportFromReaderRefusesRejectPathCollidingWithSource` both subtests **FAIL** ✓ |
| P7 | `if false &&` over the strict gate (import.go:337) | **no test failed** — see NIT-1 |
| P8 | `firstParseReject` hard-fail → no-op (import_shared.go) | `.../corrupt_line_fails_the_whole_restore` **FAIL** ✓ |
| P9 | drop the partition in `importFromLocalJSONLWithOpts` | `.../writer-refused_record_is_skipped_and_quarantined` **FAIL** ✓ |

No vacuous test in the set.

## Where I tried to refute the PR and could not

**Secret leakage to stderr/CI logs — clean.** `reportRejectedRecords` (import_reject.go:200) prints line, id and `Reason` only; `rejectedRecord.raw` is unexported (import_reject.go:69) so it reaches neither stderr nor `--json`. I chased the reason strings to their sources: `CheckFieldLen` reports the rune *count*, not the value (internal/types/types.go:298-303); `ValidateIssueTitle` reports the length, not the title (types.go:337-345); `ValidateWithCustom` echoes only short enum-shaped fields (`invalid status: verify`). Observed live in the test run: `warning: skipped invalid record on line 2 (unit-bad): validation failed for issue unit-bad: invalid status: verify`. No record body. The verbatim copy goes to the quarantine *file*, which lands 0600 by construction (CreateTemp + rename, import_reject.go:337-360) — `TestWriteRejectFileHardening` pins both that and the symlink-replacement.

**Per-record atomicity — holds.** Validation runs before anything is handed to storage, so a skipped record never enters a batch; there is no mid-batch rollback to get wrong. `TestImportFromLocalJSONLRejectContract/writer-refused...` proves it against a real store: `test-ok1`/`test-ok2` present, `test-bad` absent, count 2. Strict mode returns from `resolveImportRejects` before either `runImportRecordsClassic` or `runImportRecordsProxied` is reached, so "nothing was imported" in the error text is literally true.

**JSON shape — additive only.** `invalid_skipped`, `invalid_records`, `rejects_written_to` are all new and all `omitempty` (import.go:249-254); no existing field renamed, retyped or dropped. Struct-field reordering changes key emission order, which is not a contract.

**1611 lines — no blob.** 658 lines is `import_reject_test.go`, 416 is `import_reject.go`; the remainder is spread across five existing files. Everything is hand-written Go, nothing generated, no fixture blob. CHANGELOG entry sits under `## [Unreleased]` → `### Fixed` (heading at line 8, subsection at 186, entry at 291). `docs/CLI_REFERENCE.md` and `docs/cli-reference/` are generated from the pinned release tag (`docs/cli-docs.pin` = v1.2.2), so leaving them alone is correct.

**PR body vs diff — accurate.** Ten files, all import-path or CHANGELOG; nothing out of scope. The body's account of the two-commit polarity flip matches the code.

## NITs (non-blocking)

**NIT-1 — the default-polarity decision has no test that runs outside the proxied harness.** P7 removed the strict gate at `cmd/bd/import.go:337` and the entire 24-test unit set stayed green. The behavior *is* asserted — `TestProxiedServerImport/skip_invalid_partitions_validation_rejects` (import_proxied_integration_test.go:568) checks both halves — but that test SKIPped in my environment, and it exercises only the proxied route. Given the second commit exists solely to decide this polarity, and `resolveImportRejects` is shared by both routes, a ~15-line unit test would pin it cheaply: set `importSkipInvalid = false`, call `resolveImportRejects` with one reject, assert a non-nil error and that `writeRejectFile` was not reached. The scaffolding is already there in `TestResolveImportRejectsDryRunLeavesRejectsFileAlone`.

**NIT-2 — `.beads/*.rejected.jsonl` is not gitignored.** `maybeAutoImportJSONL` quarantines to `rejectFilePath(jsonlPath)` (auto_import_upgrade.go:143), and `jsonlPath` is inside `beadsDir` (auto_import_upgrade.go:86). `.beads/issues.jsonl` is normally tracked, so its `.rejected.jsonl` sibling appears as untracked next to it and is committable — a file of raw issue records, written unattended during an upgrade, that the user never asked for. `requiredPatterns` in `cmd/bd/doctor/gitignore.go:112-140` has no entry for it. One line: add `*.rejected.jsonl`.

**NIT-3 — `docs/reference/json-schema.md` still lists the pre-PR `import --json` shape.** The `### import --json` field list at lines 210-225 is a hand-maintained doc (not generated from the pin) and does not mention `invalid_skipped`, `invalid_records` or `rejects_written_to`. Three bullets.

**NIT-4 — `--dry-run` gained an undisclosed exit-code change.** `resolveImportRejects`'s strict gate runs before the dry-run classification, so `bd import --dry-run file-with-one-bad-status.jsonl` now exits nonzero where it previously printed a plan and exited 0 (validation never ran on the old dry-run path — nothing reached the writer). I think this is the *right* behavior — a dry run should predict the real run — but the CHANGELOG only mentions the filesystem half ("A dry run no longer writes (or removes) the `--rejects` file"). Worth a clause.

**NIT-5 — the vocabulary-read failure is silent in strict mode.** `cmd/bd/import.go` classic branch and proxied branch both warn only `else if importSkipInvalid`. When the vocabulary read fails *without* `--skip-invalid`, the pre-filter is silently disabled and the user gets the old opaque batch abort — no line number, no hint — which is exactly the GH#4492 symptom, with nothing on stderr explaining why the new machinery stood down. Drop the `importSkipInvalid` condition on that warning.

## Bottom line

Fix MAJOR-1 and I'll approve. The nits are yours to take or leave, though NIT-1 and NIT-2 are each a handful of lines and I'd rather they went in with this change than became follow-ups.

Issue #4440, opened 2026-06-17
v1.0.4: core.hooksPath survives manual uninstall and regenerates .beads/ on checkout (follow-up to #1710)


omeletteak:
### Summary
A follow-up to #1710 (CLOSED/COMPLETED). Since that issue, the hook mechanism changed from writing into `.git/hooks/` to pointing `core.hooksPath` at the repo's `.beads/hooks`. After a **manual** uninstall (`rm -rf .beads/` + removing the BEADS doc sections), the git config left behind causes `.beads/` to **regenerate on the next checkout/pull** — a more active failure than the cryptic commit errors in #1710.

### What remained after `rm -rf .beads/` (v1.0.4)
In `.git/config` (local):
- `core.hooksPath = <repo>/.beads/hooks`   ← not listed in #1710
- `beads.role = contributor`

`.git/hooks/` itself was clean (only `*.sample`), which makes this hard to diagnose: inspecting `.git/hooks` shows nothing.

### Symptom
`git pull --rebase` momentarily checks out a past commit that still tracks `.beads/hooks/post-checkout`; `core.hooksPath` makes git run it, which invokes `bd` and **recreates `.beads/`** (an `embeddeddolt` DB under the *old* project prefix), printing:

```
beads: post-checkout import warning: exit status 1
Error: import failed: database not initialized: issue_prefix config is missing (run 'bd init --prefix <prefix>' ...)
```

`git status` then shows `?? .beads/` reappearing on every checkout.

### Environment
- OS: Windows 11, PowerShell + Git Bash
- bd v1.0.4 (ce242a879) in PATH
- Repo originally `bd init`'d by an earlier version that committed `.beads/hooks/` and set `core.hooksPath`; `.beads/` later removed by hand.

### Repro
1. In a repo where an older `bd init` committed `.beads/hooks/` and set `core.hooksPath=.beads/hooks`
2. `rm -rf .beads/`, remove the BEADS doc sections; keep `bd` in PATH
3. `git pull --rebase` (or any checkout touching a commit that still has `.beads/hooks/`) → `.beads/` regenerates

### Suggested fixes
1. `bd hooks uninstall` / a `bd uninit` should also `git config --unset core.hooksPath` (when it points at `.beads/hooks`) and `--unset beads.role`.
2. `bd doctor` should detect "`core.hooksPath` points to a missing `.beads/hooks`" and offer `--fix`.
[...]

Issue #3794, opened 2026-05-07
support `bd doctor` with embedded mode


maphew:
`bd doctor` doesn't support embedded mode because it needs a careful (human) audit to ensure its accuracy and that its suggestions/fixes are correct and do not compound problems further. Agents do not understand embedded dolt and their assessement of how it works is hallucinatory.

I think its worth filing an issue to support `bd doctor` with embedded mode, but the work should not be in a single pr for every subcommand, it should be done a single subcommand at a time, so that code changes and tests can be vetted and verified.

_Originally posted by @coffeegoddd in https://github.com/gastownhall/beads/pull/3439#pullrequestreview-4246774910_

maphew:
When building the fulsome solution, mine the source in PR 3439 and bring forward the still-applicable parts with attribution to @sjsyrek.

harry-miller-trimble:
This is a feature request — new functionality, behavior change, or new CLI surface area. New features are currently frozen pending maintainer review of scope and priority.

Leaving open for maintainer triage. Triage gate is not auto-implementing this; please don't take it as approval to start coding without sign-off.

maphew:
Thanks for the report. I’m doing a repro sweep and couldn’t find a concrete reproduction recipe in the current description or thread.

Could you add the smallest command sequence or setup that triggers this, including the `bd version`, storage mode, OS, and the expected vs. actual output? A scratch-repo repro is ideal if one is possible.

_codex-gpt-5.5-medium on behalf of matt wilkie_

PR #5203, opened 2026-07-31
fix(doctor,hooks): diagnose a dangling core.hooksPath and clear beads.role on uninstall (#4440)

maphew:
## Why

Reported in #4440: the user uninstalled beads by hand (`rm -rf .beads/`) and was left with `core.hooksPath` still pointing at the directory they had just deleted, plus a stale `beads.role`. Nothing told them so, and `.beads/` reappearing on checkout made it look like beads was reinstalling itself.

Two things worth correcting from the report before the fix makes sense.

**It is not regeneration.** `runPostCheckoutHook` runs the chained hooks and then `importJSONLForSync("post-checkout")`. `.beads/` reappears as a side effect of that import initialising an embedded-dolt workspace under the old prefix — which is also why the reporter then sees "issue_prefix config is missing". There is no regeneration code to find, and a fixer looking for some will waste an afternoon.

**One of the three suggested fixes already exists and does not help.** `uninstallHooks()` has called `resetHooksPathIfBeadsManaged()` since before this issue was filed. It is irrelevant here because the reporter did a *manual* uninstall and never ran the command; by the time they noticed, `.beads/` was already gone. The other two suggestions were genuinely unimplemented, and are what this PR does.

## What

**A "Hooks Path" doctor check.** None of the three registered hooks checks tests whether the configured hooks path still exists — `hooks_migration.go` reads `core.hooksPath`, resolves it, and assigns it without ever calling `os.Stat`. The new check warns when it resolves to a missing directory, naming both the configured value and the resolved path.

`bd doctor --fix` unsets it, but **only when it is one of the values bd itself writes** (`.beads/hooks`, `.beads-hooks`, relative or absolute). `core.hooksPath` is not a beads-only key — husky and friends use it too, and unsetting someone else's would disable their hooks. `FixHooksPath` re-reads and re-checks both conditions immediately before acting, so a stale check result cannot cause it to touch third-party config. That guard has its own test asserting the value is *still there* afterwards.

**`bd hooks uninstall` also clears `beads.role`, and fails loudly.** `beads.role` was previously only unset by an explicit `bd config unset`, so the reporter's leftover was expected behaviour. Uninstall also swallowed a failed config reset as a scrolling stderr warning and still returned success; it now returns an error naming what it could not unset. Hook-file removal errors keep returning immediately, as before.

`beads.role` is *read* before it is unset rather than treating git's exit 5 as "already absent". Exit 5 also means "refusing an ambiguous unset" on a multi-valued key, so conflating the two would report a clean uninstall with the key still set — the exact failure this change exists to prevent. `TestUninstallHooksReportsAmbiguousBeadsRole` sets the key twice and asserts both the error and that the values survived.

**Docs.** `docs/recovery/uninstalling.md` walks users through precisely the manual path in this report and never mentions `core.hooksPath` — grep the file, it is not there. Manual Cleanup now includes it, with the reason and the not-beads-only caveat.

## Known gap, stated plainly

**The new doctor check does not reach the users most likely to hit this.** Bare `bd doctor` is gated off in embedded mode by deliberate policy (#3794): *"embedded support is enabled one subcommand at a time, each human-vetted — do not lift this gate wholesale."* Embedded is the default, and it is the reporter's mode.

The gate's stated reason is that checks reach into the database layer. This one does not — it shells out to git and stats a directory. So it looks like a candidate for the same per-subcommand vetting that already admitted `--check=artifacts`, `--check=conventions` and `--check=pollution`. But that is a maintainer call the policy comment explicitly reserves, and I have not made it here. Filed separately.

The uninstall and documentation changes in this PR are mode-independent and do help embedded users today.

## Verification

- New tests: `cmd/bd/doctor/hooks_path_test.go` (unset, existing dir, missing beads-managed in relative/`.beads-hooks`/absolute forms, missing non-beads `.husky/_` no-touch guard, existing non-beads, fix-is-no-op-when-target-exists) and `cmd/bd/uninstall_config_test.go` (unsets `beads.role`, absent key is not an error, ambiguous key is a loud error, genuine unset failure is surfaced).
- `go test ./cmd/bd/doctor/... -run 'HooksPath|Hooks'` — `ok`
- `go test ./cmd/bd/ -run 'Hook|Uninstall|ResetHooksPath'` — `ok`
- `go vet ./cmd/bd/...`, `gofmt -l` — clean.
- Full `make test` is queued in the local verifier; I will report the result here.

Confirmed by hand that `bd doctor --fix` leaves a `.husky/_` value untouched.

## Note on CI

`main` is currently red on an unrelated failure (11 migrate-mode tests failing with `workspacegate: acquiring .beads.gate.lock: context canceled`, since `d223edcb6`). Expect that to show on this PR too; it is not from these changes.

### DIFF
--- a/cmd/bd/doctor.go
+++ b/cmd/bd/doctor.go
@@ -448,6 +448,11 @@ func runDiagnostics(path string) doctorResult {
 	legacyCheck := convertWithCategory(doctor.CheckStaleLegacyHooks(), doctor.CategoryGit)
 	result.Checks = append(result.Checks, legacyCheck)
 
+	// Check for a dangling core.hooksPath pointing at a missing directory (GH#4440)
+	hooksPathCheck := convertWithCategory(doctor.CheckHooksPath(), doctor.CategoryGit)
+	result.Checks = append(result.Checks, hooksPathCheck)
+	// Warning-class check — don't fail overall check, matching the neighboring hooks checks.
+
 	// Check git hooks Dolt compatibility (hooks without Dolt check cause errors)
 	doltHooksCheck := convertWithCategory(doctor.CheckGitHooksDoltCompatibility(path), doctor.CategoryGit)
 	result.Checks = append(result.Checks, doltHooksCheck)
--- a/cmd/bd/doctor/git.go
+++ b/cmd/bd/doctor/git.go
@@ -596,6 +596,128 @@ func staleFilePaths(hooksDir string, names []string) []string {
 	return paths
 }
 
+// CheckHooksPath detects a dangling core.hooksPath: a git config value that
+// still points at a hooks directory which no longer exists. This happens
+// when a user manually deletes .beads/ (e.g. `rm -rf .beads/`) without
+// running `bd hooks uninstall` first — git keeps looking for the missing
+// directory, and beads' post-checkout import can recreate a stray .beads/
+// workspace as a side effect (GH#4440).
+func CheckHooksPath() DoctorCheck {
+	const name = "Hooks Path"
+
+	repoRoot, _ := git.GetMainRepoRoot()
+	if repoRoot == "" {
+		repoRoot = git.GetRepoRoot()
+	}
+	if repoRoot == "" {
+		return DoctorCheck{
+			Name:    name,
+			Status:  StatusOK,
+			Message: "N/A (not a git repository)",
+		}
+	}
+
+	hooksPath, ok := getConfiguredHooksPath(repoRoot)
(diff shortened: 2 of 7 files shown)


maphew:
Full suite result, as promised in the description: **`make test` passed** on `b079684b0` (exit 0), run locally on Linux with dolt 2.2.3.

Still true and worth repeating from the description: the new "Hooks Path" check does not run in embedded mode, because bare `bd doctor` is gated there by #3794. The uninstall and documentation changes in this PR are unaffected by that gate.

PR #5204, opened 2026-07-31
fix(cmd): stop a canceled rootCtx leaking into the next in-process command

maphew:
## Why

`main` has been red on the **Main** workflow since run [30617861680](https://github.com/gastownhall/beads/actions/runs/30617861680) (2026-07-31 08:52Z, head 755174dfa). Last green was [30615821296](https://github.com/gastownhall/beads/actions/runs/30615821296) at 08:18Z. Every non-cancelled `Main` run since has failed with the identical signature on **both** `ubuntu-latest` and `macos-latest`:

```
cannot migrate while other bd activity holds this workspace:
workspacegate: acquiring <dir>/.beads.gate.lock: context canceled
```

Eleven `cmd/bd` cases, all at 0.00s: `TestMigrateToProxiedServer_FlipsMode`, `TestMigrateToServer_FlipsModeAndRemovesSidecar`, `TestMigrateToServer_KeepsCustomConfig`, `TestMigrateMode_RefusesWhenLockHeld`, `TestMigrateToServer_RefusesWhenLifecycleLockHeld` (+3 subtests), `TestMigrateMode_ReleasesLockAfterSuccess`, `TestMigrateSharedToProxiedServer_RootsAtSharedDir`, `TestMigrateProxiedToSharedServer_Reverse`, `TestMigrateToProxiedServer_AlreadyProxiedIsNoop`.

They all pass in isolation, which is why the failure reads as flake until you run the package in order.

## Root cause

`PersistentPostRunE` calls `rootCancel()` and leaves the `rootCtx` global pointing at the context it just cancelled. A real `bd` process exits immediately after, so nothing notices. The `cmd/bd` test binary calls `Execute()` many times in one process, so every test after the first full command path inherits a dead root context.

That was latent until #5093 threaded `rootCtx` into `acquireMigrateGates`. The failure is `workspacegate.Acquire`'s `ctx.Err()` pre-check firing on a cancellation nobody requested.

#5093 already knew about the leak — `workspace_gate_test.go` pins `rootCtx` to nil so `TestChokepointSharedExcludesMigrateExclusive` survives it — but `migrate_dolt_mode_test.go` was left undefended. Pinning each affected test is the wrong shape; the leak belongs to the lifecycle.

Worth noting for process: the culprit landed in a batch of ~15 commits pushed 08:40–08:52Z whose CI runs were all concurrency-cancelled, so it merged without an independent green run.

## What changed

- **`PersistentPostRunE`** cancels and clears the context globals from a deferred hook registered before the cleanup defers. Deferred so the maintenance block's error returns clear them too; registered first so it runs *last*, because `closeStoreBeforeGateRelease` passes `rootCtx` to `uowProvider.Close` and previously received an already-cancelled context. `nil` is the documented "no process signal context yet" state and normalizes back to `Background()` through `getRootContext()`.

- **`PersistentPreRunE`** publishes the signal context through `setRootContext` rather than assigning the globals directly. `cmdCtx` exists by then, so `getRootContext()` reads `cmdCtx.RootCtx`, and every `skipsStoreInit` command — migrate among them — returns before `syncCommandContext` can backfill it. Without this, those commands see a nil per-command context and lose Ctrl-C.

- **`acquireMigrateGates`** reads `getRootContext()` rather than the raw global, so it honors the per-command context when globals are disabled.

- **`root_context_lifecycle_test.go`** pins the post-run contract across both context representations, the error path, and the nil case — so the next consumer to thread `rootCtx` somewhere new does not rediscover this the way migrate did.

## Validation

- Reproduced first with a scratch test that sets `rootCtx` to a cancelled context: emits the CI error byte-for-byte.
- All 11 previously-failing tests pass, plus `TestChokepointSharedExcludesMigrateExclusive` and the new lifecycle tests.
- A full `cmd/bd` package run against a pre-fix baseline is in progress locally; this box's SELinux policy blocks the suite's habit of building and exec'ing a `bd` binary from a temp dir, so a large slice of that run is environment-only noise in both directions. CI here is the authoritative check.

The second and third bullets under "What changed" came out of an independent cross-model review, which caught that routing `acquireMigrateGates` through `getRootContext()` *alone* would have silently dropped Ctrl-C for migrate in production.

### DIFF
--- a/cmd/bd/main.go
+++ b/cmd/bd/main.go
@@ -793,7 +793,15 @@ var rootCmd = &cobra.Command{
 		// Set up signal-aware context with batch commit flush on shutdown.
 		// Unlike signal.NotifyContext, this also handles SIGHUP and flushes
 		// pending batch commits before canceling the context.
-		rootCtx, rootCancel = setupGracefulShutdown()
+		//
+		// Publish through setRootContext, not a bare assignment to the
+		// globals: cmdCtx exists by now (initCommandContext above), so
+		// getRootContext() reads cmdCtx.RootCtx, and the commands that
+		// return early from this hook -- every skipsStoreInit command,
+		// migrate among them -- never reach syncCommandContext to have it
+		// backfilled. A bare assignment leaves those commands reading a nil
+		// per-command context and losing Ctrl-C entirely.
+		setRootContext(setupGracefulShutdown())
 
 		// Initialize OTel (no-op unless BD_OTEL_METRICS_URL or BD_OTEL_STDOUT=true).
 		// Must run before any DB access so SQL spans nest under command spans.
@@ -1552,6 +1560,27 @@ var rootCmd = &cobra.Command{
 		return nil
 	},
 	PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
+		// Registered FIRST so it runs LAST: the signal context must outlive
+		// the store/gate cleanup below, which passes rootCtx to
+		// uowProvider.Close. Canceling in the function body (as this used
+		// to) handed those closers a dead context on the way out.
+		//
+		// Clearing matters as much as canceling. Leaving rootCtx pointing at
+		// the context we just canceled is harmless when a real bd process
+		// exits here, but every in-process caller that runs Execute() more
+		// than once -- the cmd/bd test binary, library embedders -- would
+		// hand that dead context to the next command, and anything reading
+		// it refuses work nobody canceled. nil is the documented "no process
+		// signal context yet" state and normalizes back to Background().
+		//
+		// Deferred rather than inline so the early error returns below clear
+		// the globals too.
+		defer func() {
+			if rootCancel != nil {
+				rootCancel()
+			}
+			setRootContext(nil, nil)
+		}()
 		defer restoreChangeDirSelection()
 		// Release the workspace/physical-root gates on EVERY exit from
 		// PostRunE — deferred so the early error returns below cannot leak
@@ -1684,10 +1713,9 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
-		// Cancel the signal context to clean up resources
-		if rootCancel != nil {
-			rootCancel()
-		}
+		// The signal context is canceled and cleared by the deferred hook
+		// registered at the top of this function, so that it also covers the
+		// early error returns above.
 		return nil
 	},
 }
--- a/cmd/bd/migrate_dolt_mode.go
+++ b/cmd/bd/migrate_dolt_mode.go
@@ -186,7 +186,10 @@ func acquireMigrateGates(beadsDir string, shared bool, reason string) (func(), e
 	if err != nil {
 		return nil, HandleError("failed to resolve migration destination root: %v", err)
 	}
-	h, err := acquireExclusiveWorkspaceGates(rootCtx, beadsDir, reason, destRoot)
+	// getRootContext(), not the rootCtx global: it honors the per-command
+	// context when globals are disabled, and normalizes the not-yet-set case
+	// to context.Background().
+	h, err := acquireExclusiveWorkspaceGates(getRootContext(), beadsDir, reason, destRoot)
 	if err != nil {
 		return nil, HandleErrorWithHint(
 			fmt.Sprintf("cannot migrate while other bd activity holds this workspace: %v", err),
--- a/cmd/bd/root_context_lifecycle_test.go
+++ b/cmd/bd/root_context_lifecycle_test.go
@@ -1,0 +1,127 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/spf13/cobra"
+)
+
+// snapshotRootContext saves and restores BOTH representations of the root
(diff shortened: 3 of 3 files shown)


maphew:
Independent confirmation plus two small findings. I diagnosed this incident in a parallel session, arrived at the same root cause and the same PostRunE shape, then found this PR — so this is a second pair of eyes on it rather than a competing change. **Nothing below should hold up the merge**; main has been red 14h and this PR's own checks are clean (all pending, none failing).

## Confirmation

Reproduced independently, and I can add one datapoint the description does not have: the leak is triggered by *any* successful in-process command, not just the ones that touch storage. A three-line scratch test that runs `rootCmd.Execute()` with `["version"]` and then reads `rootCtx` reports `context canceled`, and the very next migrate test emits the CI error verbatim:

```
Error: cannot migrate while other bd activity holds this workspace:
workspacegate: acquiring <dir>/.beads.gate.lock: context canceled
```

I also bisected by inspection to the same commit — `6d81e1c73` (#5093), the `acquireExclusiveWorkspaceGates(rootCtx, ...)` call in `acquireMigrateGates`. And I confirmed the ordering claim from the failing macOS log: the *first* test in `migrate_dolt_mode_test.go` already fails, so the poison is inherited from an earlier file, not from a sibling migrate test.

One note for anyone re-running this locally: `cmd/bd` takes ~9 minutes on a warm box, so a plain `go test ./cmd/bd/` silently dies on Go's default 10m timeout and reports a truncated failure list that looks like real breakage. Use `-timeout 60m`.

## Finding 1 — a comment this PR falsifies (small, worth folding in)

`cmd/bd/workspace_gate_test.go`, in `TestChokepointSharedExcludesMigrateExclusive`, currently documents the leak as *current production behaviour*:

> rootCtx is a package global that production sets via setupGracefulShutdown() in PersistentPreRunE and cancels via rootCancel() in PersistentPostRunE **WITHOUT resetting the var to nil** — harmless in production (the process exits), but any earlier in-process test that exercises the full command path (Execute()) leaves rootCtx pointing at an already-canceled context for whatever test runs next in the same binary.

After this PR that paragraph is wrong, and it is the one place in the tree a future reader is most likely to look when they hit something context-shaped — the `rootCtx = nil` pin right below it stops being load-bearing and becomes belt-and-braces. Since this PR is what makes the comment false, it seems right to correct it here. Suggested replacement:

```go
	// rootCtx is a package global that production sets via
	// setupGracefulShutdown() in PersistentPreRunE and tears down in
	// PersistentPostRunE. The teardown clears the var precisely so an
	// earlier in-process command cannot hand this test a canceled context
	// (see root_context_lifecycle_test.go). Pin it to nil anyway: this
	// test's subject is the "no process signal context yet" case, and it
	// should assert that regardless of how teardown is implemented.
```

## Finding 2 — the same asymmetry survives one line further down (nit, fine as a follow-up)

Publishing through `setRootContext` at `main.go:801` is right, and the reasoning about `skipsStoreInit` commands is correct. But the telemetry re-wrap ~35 lines later is still a bare assignment to the globals:

```go
rootCtx, commandSpan = telemetry.Tracer("bd").Start(rootCtx, "bd.command."+cmd.Name(), ...)
```

So from that point until `syncCommandContext()` backfills `cmdCtx.RootCtx` — and *forever* for the `skipsStoreInit` commands this PR just fixed, since they never reach the sync — `getRootContext()` hands back the pre-span context. Cancellation is unaffected (the span context derives from the same parent `rootCancel` cancels, so Ctrl-C still works and the PR's stated benefit holds). It is purely an observability gap: child spans created through the accessor won't nest under the command span.

Not a regression — before this PR those callers got `Background()`, which is span-less *and* uncancellable — so this is strictly better either way. Flagging only because it is the identical globals-vs-`cmdCtx` divergence being fixed one line above, and `setRootContext(spanCtx, rootCancel)` would close it. Tracked as `mybd-3xire` if it's easier as a follow-up.

## Optional

The new lifecycle tests pin the contract well, including the error path, which mine did not. The one thing they don't assert is the *symptom* — that `acquireMigrateGates` actually succeeds on a free workspace after a prior command's teardown. The 11 migrate tests going green in CI covers it in practice, so this is genuinely optional, but a direct end-to-end assertion is cheap:

```go
	setRootContext(setupGracefulShutdown())
	// ... teardown ...
	release, err := acquireMigrateGates(beadsDir, false, "test migrate after a prior command")
	if err != nil {
		t.Fatalf("migrate gates must be acquirable on a free workspace after an earlier command finished: %v", err)
	}
	defer release()
```

(If you take that, note `t.Cleanup(releaseWorkspaceGates)` is *not* the right cleanup for it — `acquireMigrateGates` returns its own closure and never sets the `workspaceGateHandle` global.)

## Process observations, already filed

- The detector gap: the base-red lane only observes base health through parked merge lanes, so it went silent for 14h into an empty merge queue. Independently confirmed and filed; deduped into `mybd-sopqb`, which has the better writeup.
- The batch of ~15 commits merged 08:40–08:52Z with concurrency-cancelled CI runs is, I think, the more interesting process finding in this PR's description — the culprit merged without any independent green run. That's upstream of both the detector gap and the merge deadlock in `mybd-01yzj`.

PR #5206, opened 2026-07-31
feat(update): enforce close policy when a status update crosses into done

julianknutsen:
## What

A generic status update that crosses from a non-done status into a done-category status (`closed`, or a configured custom done status) now enforces close policy — the open-children refusal and the live-direct-blocker refusal — instead of bypassing it. `bd update --force` gains a second meaning: it overrides close policy in addition to the existing assignee-transfer fence.

What changes for users, before → after:

- **Direct and proxied `bd update -s closed`** on an issue with open children or a live direct blocker: before, succeeded silently; after, refuses with the same message `bd close` prints, and `--force` overrides.
- **`bd update --force` without `-a/--assignee`**: before, a validation error on the direct path; after, valid — it means "override close policy" when no assignee transfer is requested.
- **`bd batch`**: the update grammar gains a force token (`update <id> status=closed force=true`); an unforced refusal fails the script and rolls back the whole batch, since batches run in one transaction.
- **Tracker sync-pull**: before, wrote remote closes through the unchecked generic path; after, the pull path always sets the override — remote state is authoritative, so observable sync behavior is unchanged, now explicitly rather than accidentally.
- **`bd batch close`**: unchanged — still an unchecked close. The asymmetry is documented in the batch help text.

Breaking for scripts that relied on `bd update -s closed` to bypass the checks `bd close` applies. CHANGELOG entry under [Unreleased] > Changed. Facade consumers set `UpdateRequest.ForceClosePolicy`.

## Why

`bd close` refuses to close an issue with open children or a live blocker, but `bd update -s closed` — the same transition by another name — skipped both checks on every path (CLI direct, proxied, batch, in-process embedder APIs). This closes that hole.

Enforcement lives at the two storage-layer update funnels (the issueops update transaction and the domain repository `Update`) rather than at call sites: every status-capable update path in the tree converges on one of those two implementations, so new callers are covered by default. The force override travels as a reserved update-map key that each funnel pops before field validation. A caller that fails to plumb it hits the field allowlist and is refused loudly by name, never silently stripped. That property is deliberate: earlier attempts at enforcing this at a shared boundary were reverted because some callers (proxied, batch) either could not satisfy the check or silently dropped the override; the loud-failure transport plus per-caller tests pin against both failure shapes.

The commit series is staged so no commit enforces policy on a caller that cannot yet override: characterization pins first, then a pure extraction of the existing `bd close` gate (byte-identical policy body), then inert transport, then caller plumbing, then a single flip commit with docs.

## Verification

Note: the embedded-backend package skips silently without `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1` — `go test` still prints `ok` with zero tests run, so count RUN/PASS lines rather than trusting `ok`.

```
go vet ./...                                   # clean; golangci-lint run: 0 issues
go test ./internal/storage/issueops/ -v        # 338 pass, 0 fail
go test ./internal/storage/domain/db/ -v       # 799 pass, 0 fail
go test ./internal/storage/uow/ -v             # 135 pass, 0 fail (includes conformance close-policy case)
go test ./internal/storage/dolt/ -run 'ClosePolicy|IssueOperations' -v         # live Dolt
BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 \
  go test ./internal/storage/embeddeddolt/ -run 'ClosePolicy|IssueOperations' -v
go test ./cmd/bd/ -run TestUpdateClosePolicy -v   # 8 pass, 0 fail
go test ./internal/tracker/ -run ClosePolicy -v   # 1 pass
```

The two backend filters must show PASS lines (0 FAIL) for `TestEnforceClosePolicyInTxMatchesCheckedClose`, `TestEnforceClosePolicyInTxRoutesDualResidentToDurableRow`, `TestCrossesIntoDoneCategoryInTx`, `TestUpdateIssueEnforcesClosePolicyForConfiguredDoneStatus`, and the `IssueOperationsUpdateClosePolicy` conformance entry, which also runs on the uow backend — so the same contract is exercised on all three backends. The step-0 characterization pins were run against pre-change code and passed with the old permissive assertions before being inverted in the commits that changed each behavior.

Limitations, honestly stated:

- The proxied CLI refusal branches (stderr copy and exit-code shaping in `cmd/bd/update_proxied_server.go`) are not exercised end-to-end. The enforcement behind them is covered by the uow conformance run, and the spec mapping that carries `--force` is pinned by a unit test, but no test drives a proxied `bd update -s closed` into a refusal and asserts the printed output and exit code.
- Tracker always-force is pinned at the unit level (the pull path injects the override, asserted against a recording fake), not by an end-to-end pull over a real store.
- `bd batch close` remaining unchecked is documented but not pinned by a characterization test.
- `go test ./cmd/bd/` run whole-package has 34 pre-existing failures (init/config/doctor/completion), verified test-for-test identical on an origin/main worktree — unrelated to this change; use the `-run` filter above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Breaking: `bd update --status <done-status>` now enforces close policy.**
+  Moving an issue into `closed` (or any configured done-category status) via
+  `bd update`, `bd batch update`, or the issueops facade now refuses when the
+  issue has open children or a live direct blocker — matching `bd close`.
+  Override with `bd update --force`, which now means: allow `-a/--assignee` to
+  overwrite another actor's live in_progress claim, AND allow a status change
+  into done despite open children or a live blocker (same as
+  `bd close --force`). `bd update --force` without `-a` was previously a
+  validation error on the direct path; it is now valid. In `bd batch`, spell the
+  override `update <id> status=closed force=true`; an unforced refusal rolls
+  back the entire batch. Facade consumers: set `UpdateRequest.ForceClosePolicy`.
+  Tracker sync-pull always forces (remote state is authoritative).
+  `bd batch close` remains unchecked, as before.
+
 - **`beads.BulkIssueStore.ReclaimExpiredLeases` gained a `types.ReclaimFilter`
   parameter** (wy-jpd3.3), threaded through the domain use-case/repository
   interfaces and every backend. Callers that reclaim globally pass the zero
--- a/cmd/bd/batch.go
+++ b/cmd/bd/batch.go
@@ -53,8 +53,15 @@ Grammar (one command per line):
   dep remove <from-id> <to-id>
   #comment  (blank lines and '# ...' comments are ignored)
 
-Supported 'update' keys: status, priority, title, assignee
+Supported 'update' keys: status, priority, title, assignee, force
 Supported dependency types: see 'bd dep add --help' (default: blocks)
+
+'force' is not a field. An update whose status moves the issue into closed
+(or a configured done status) is refused when it still has open children or
+a live blocker, the same as 'bd close'; 'force=true' overrides that refusal.
+Because the batch is one transaction, an unforced refusal rolls back EVERY
+operation in the batch, not just the offending line. Note the asymmetry with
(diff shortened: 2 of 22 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98638313

Issue #3370, opened 2026-04-20
bug: dolt auto-push hangs indefinitely when remote is unreachable


kevglynn:
## Bug

`maybeAutoPush()` in `cmd/bd/dolt_autopush.go` calls `st.Push(ctx)` with the root context, which has no timeout. When the Dolt remote is unreachable (e.g., network issues, SSH tunnel down, wrong credentials), the underlying `git fetch` blocks indefinitely. This causes:

1. The `bd` CLI hangs after every write command (create, close, update, etc.)
2. The embedded Dolt `.lock` file is held, blocking all other `bd` processes
3. Users must `kill` the process and manually `rm .beads/embeddeddolt/.lock`

## Reproduction

```bash
# Set up a project with a Dolt remote that's unreachable
bd init --prefix test
# ... configure a remote that can't be reached ...
bd create "test issue"
# CLI hangs forever after "Created issue: ..."
```

## Expected Behavior

Auto-push should have a bounded timeout (e.g., 30 seconds). On timeout:
- Log a warning to stderr (like the existing push-failure path)
- Release the lock
- Exit normally (the local write already succeeded)

## Proposed Fix

In `maybeAutoPush()`, wrap the push call with `context.WithTimeout`:

```go
pushCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
defer cancel()
if err := st.Push(pushCtx); err != nil {
    // existing error handling...
}
```

Similarly, `isDoltAutoPushEnabled()` calls `st.HasRemote(ctx, "origin")` which could also hang on network issues — this should get a shorter timeout (e.g., 5 seconds).

## Impact

This affects any user whose Dolt remote becomes temporarily unreachable. The workaround is `bd config set dolt.auto-push false`, but most users won't know this until they've already been blocked.

## Files

- `cmd/bd/dolt_autopush.go`: `maybeAutoPush()` (line ~142), `isDoltAutoPushEnabled()` (line ~67)

PR #5207, opened 2026-07-31
fix(embeddeddolt): route all remote verbs through withPeerAuth (stacked on #5085)

arcaven:
This is follow-up 1 from the #5085 review, implemented while the context is warm: the seven non-federation remote verbs still resolved credentials from the environment alone, so a federation peer reached as an ordinary remote (`bd sync --remote <peer>`, `bd dolt push|pull --remote <peer>`, or as the default remote) never presented its stored credentials, and the verbs read the process-wide `DOLT_REMOTE_USER`/`DOLT_REMOTE_PASSWORD` pair with no lock while `withPeerAuth` mutates it under `federationEnvMutex`.

**Stacked on #5085.** The branch is #5085's head plus one commit (c3c634b66); everything before that commit is #5085 itself, so this collapses to the one commit once #5085 merges. If you would rather fold it into #5085 before merging, say the word and I will push the same commit onto that branch instead; the `ErrNotFound` fallback makes it a no-op for plain remotes, so it fits #5085's scope as you described ("free to fix with the same edit").

## What the edit is

`Push`, `Pull`, `PullWithStrategy`, `PullRemoteWithStrategy`, `ForcePush`, `PushRemote`, and `PullRemote` each wrap their remote IO in `s.withPeerAuth(ctx, <remote>, ...)` at the remote-name expression they already had: the four default-remote verbs pass the `defaultRemote` const, the three named-remote verbs pass their `remote` parameter. No verb resolves its remote inside the SQL call, so nothing needed hoisting and `withPeerAuth` itself is byte-identical to #5085. `PushRemote`'s two former env reads (force and non-force branches) collapse to one resolution per call. `remoteAuthUser` is retained as `withPeerAuth`'s fallback resolver and is now reachable only through it; after this change `version_control.go` holds zero bare invocations, and the repo-wide remainder is the two fallback calls inside `withPeerAuth`.

Ordering is preserved: `CommitPending` and the pre-merge head capture still run before the credential wrapper, and `recomputeBlockedAfterPull` runs after it returns, so credentials are installed only for the remote IO, matching `PullFrom`'s shape from #5085.

This also addresses the unsynchronized-reader point from your review: a verb operating on a peer-backed remote now reads the env pair holding `federationEnvMutex`. One residual is unchanged and worth naming: `withPeerAuth`'s own `ErrNotFound` fallback still reads the pair with no lock, so a plain-remote operation concurrent with a peer operation can observe the peer's temporarily installed values. Closing that would serialize every in-process remote operation, which felt like your call to make about `withPeerAuth`, not something to smuggle in here.

## Behavior changes to be aware of (all inherited from withPeerAuth's semantics, now reachable from these verbs)

1. **Fail-closed on a peer-credential decrypt failure.** `.beads-credential-key` is gitignored and machine-local while `federation_peers` replicates with the database, so a second machine that receives peer rows without the key gets a decrypt error (not `ErrNotFound`), and the verb now fails with `resolve peer credentials: ...` where before it proceeded on ambient env credentials. Verified A/B on this branch. #5085 already applies this to the federation verbs, so this is consistent, but it removes the last env-only path on key-less machines; the workaround is re-running `bd federation add-peer` there to re-store the credentials (include `--sovereignty` if the peer had a tier: a bare re-add currently clears it, filed as #5213). If you would rather have decrypt failure fall back to env instead, that is a one-line policy change in `withPeerAuth` and I am happy to make it here or separately. The follow-up making the decrypt error message actionable (naming the key file and the re-store step) is now up as #5214 and is useful under either policy answer.
2. **Username-only peers suppress the ambient password.** `bd federation add-peer <name> <url> --user sync-bot` with no password stores an empty password, and #5085's unit rule (stored credentials are a unit) then unsets `DOLT_REMOTE_PASSWORD` for the operation. An operator who mixed a stored username with an environment password loses that mix on these verbs. By design per #5085's tests, but newly reachable here.
3. **`federationEnvMutex` is now held across these verbs' network IO**, so concurrent remote operations on peer-backed remotes serialize in-process. One interaction worth knowing: auto-push abandons its `Push` goroutine on timeout (#3370), and if the default remote is peer-backed, an abandoned hung push holds the mutex and leaves that peer's credentials installed until it unblocks. Bounded in the CLI (the process exits shortly after); relevant for embedders, the population you named. Same in kind as a hung `PushTo` under #5085.
4. Credentials resolve by remote name, never URL, and `AddFederationPeer` leaves a pre-existing remote's URL untouched, so a peer row whose name collides with a remote pointing elsewhere would present the stored password to that other host. Pre-existing property of the peer model; noting it because these verbs accept arbitrary remote names.

## Tests

`remote_peer_auth_test.go` follows `federation_credentials_test.go`'s conventions (real embedded store in a temp dir, `t.Setenv` ambient pair, `AddFederationPeer` seeding):

- `TestRemoteVerbsPresentStoredPeerCredentials`: table-driven over all 8 entry points (the seven verbs, `PushRemote` both branches) against two stored peers plus ambient env, asserting presented user, remote name, branch, both env values during the operation, and full restoration after.
- `TestRemoteVerbPlainRemoteKeepsAmbientEnv` and `TestRemoteVerbPlainRemoteWithoutAuthPresentsNone`: plain-remote parity in both the env-pair and no-credentials cases.
- `TestRemoteEntryPointsUseVersionControlOps`: pins the four-function test seam (`vcPush` etc.) to the production `versioncontrolops` bindings by code pointer so the seam cannot drift. The seam exists because credentials only change outcomes against an authenticating remotesapi server, which a unit test cannot stand up; precedent is `beforeProxyChildStart` in `internal/storage/dbproxy/proxy/endpoint.go`. If you would rather not carry the seam, the fallback is keeping only the `withPeerAuth`-level tests and leaving per-verb routing unproven.

Falsification was checked both ways: reverting a single verb to env-only makes its subtest fail on the stored-vs-ambient mismatch, and wiring a wrapped verb to the wrong remote name makes the two-peer fixture catch it.

`go build -tags gms_pure_go ./...`, `go vet`, `gofmt`, the full `./internal/storage/embeddeddolt/...` suite (including `-race` on the new tests), and the `BEADS_TEST_EMBEDDED_DOLT=1` gated read-only tests (`ro.Push` still reports read-only, since the credential lookup is a read and `withMutatingDBConn` refuses afterward) all pass. Diff vs #5085 is 2 files, +300/-17, of which 242 lines are the new test file.

_claude-code-claude-fable-5-high on behalf of arcaven_


maphew (commented):
Reviewed at `c3c634b6669` across the full stacked diff.

Routing every remote verb through one credential resolver is the right value to preserve. Two gaps need maintainer fixes before merge:

1. The newly wrapped verbs apply stored credentials based only on remote name. Because `AddRemoteIfNotExists` keeps an existing same-name remote regardless of URL, `bd dolt push/pull --remote <name>` can send a federation password to a different host. Compare the live remote URL with the peer row and fail closed on mismatch before installing credentials.
2. Plain remotes and empty-credential peers bypass `federationEnvMutex`, so they can observe the password temporarily installed by a concurrent authenticated operation. Hold the mutex across every callback path, not only stored non-empty credentials.

Verification:

- Focused changed tests — local build blocked by missing `unicode/uregex.h`; no verdict.

Disposition: **fix-merge on the contributor branch**. The routing work is localized and useful; no replacement or request-changes review is needed.

arcaven:
Thank you.

steveyegge:
## Review verdict: MERGE (pending CI on the fresh snapshot)

Reviewed adversarially against current main. The stale-looking 4-file diff is cosmetic — two files are your already-merged #5085 base (the branch predates the squash), and the effective delta a squash-merge lands is exactly the 2 files the body claims (+300/−17): the seven ordinary remote verbs in `version_control.go` (Push/Pull/PullWithStrategy/PullRemoteWithStrategy/ForcePush/PushRemote/PullRemote) joining Fetch/PushTo/PullFrom under `withPeerAuth`, plus the 242-line entry-point test table. I verified the merge result independently (merge-tree clean) and hunted the sharp edges: no nesting deadlock (no wrapped verb calls another; lock order is uniformly env-mutex → conn), no double-auth, `overrideEnv` restore is per-key captured and order-insensitive with restore-to-unset covered, and the remaining bare credential reads (bootstrap-time `DoltClone`) are defensibly out of scope. The test-seam globals are mitigated by the code-pointer pinning test. The fail-closed-decrypt widening is disclosed and correct — and it's why #5214 should land right behind this; I've asked arcaven for the rebase there.

One follow-up I'm filing rather than blocking on, since it's an inherited #5085 pattern your body already names: `federationEnvMutex` is now held across network IO on **every** remote verb, and #3370's abandon-on-timeout goroutine can hold it (plus that peer's installed credentials) indefinitely in a long-lived process — fine for the exiting CLI, real for daemon/proxied adoption.

I've updated the branch; will approve the workflow runs and merge on green.

PR #5208, opened 2026-08-01
feat(serve): bd serve — OpenAPI v0 HTTP surface over a shared CLI/HTTP read seam

julianknutsen:
## What this adds

`bd serve` — a loopback HTTP API for beads, specified by a frozen OpenAPI v0 document, over a
library seam shared with the CLI.

| | |
|---|---|
| `GET /healthz` | liveness only; touches nothing that can fail |
| `GET /v0/beads/context` | backend, mode, versions, capabilities |
| `GET /v0/beads/ready` | ready work, `WorkFilter` params |
| `GET /v0/beads/issues` | list, `IssueFilter` params + keyset cursor |
| `GET /v0/beads/issues/{id}` | issue detail |
| `POST /v0/beads/issues/{id}:claim` | compare-and-set claim, typed 409 |

The server is the smaller half. The larger half is `internal/workapi` plus the `issueops.Reader`
role, which exist so the CLI and the HTTP surface answer the same question the same way.

## Why the shared seam

`cmd/bd` carries 49 `*_proxied_server.go` command variants, 34 of them on UOW — most commands are
implemented twice. That has already produced shipped divergence: a viewer passing a bare
`IssueFilter{}` returns gates, wisps, templates and closed beads that `bd list` hides via ~200 lines
of exclusions. Ready-work defaults were built from scratch twice inside the CLI alone, and only one
copy hard-coded `Status:"open"`.

This PR collapses that for the three read commands it fronts, and adds `issueops.Reader` following
the `issueops.Lifecycle` convention introduced in #5191 — a role with an accessor, no constructor,
leaf contract package.

## Shape

- `internal/workapi` — `BuildListFilter`, `BuildReadyFilter`, `BuildIssueDetails`, `FinishPage`,
  not-found normalization, shared default constants. Import boundary enforced by depguard: no cobra,
  no `net/http`.
- `issueops.Reader` + `IssueReader()` on `Storage`, both backends, both decorators (by recursion),
  and the UOW provider — mirroring `IssueLifecycle()` site for site.
- `internal/httpapi` — hand-written `openapi.v0.yaml`, types-only codegen, drift gates, problem
  writer, one UOW per request with a detached close.
- Four of the eight underlying slices are pure no-op refactors, individually bisectable against the
  byte-pinned protocol corpus.

## Wire decisions worth reviewing

- **No second wire struct.** Schemas pin to the canonical Go types via `x-go-type`, enforced by a
  two-way JSON-tag bijection test. A round-trip fixture is not sufficient — a new `omitempty` field
  absent from the fixture passes both directions.
- **`limit=0` means unlimited**, matching the CLI.
- **`ready` returns `IssueWithCounts`**, because `bd ready --json` does.
- **`/issues` forces `(created_at DESC, id ASC)` on every request**, documented as a deliberate
  divergence from `bd list`'s priority default. Without it, page 1 is priority-ordered while the
  cursor encodes a keyset position, so page 2 skips and duplicates rows.
- **Unknown query parameters are rejected 400.** A silently dropped filter widens the result set.
- **One error envelope**, RFC 9457, sentinels mapped via `errors.Is`, 5xx details static because
  driver and dial errors embed the DSN.

## Security posture

Numeric-loopback bind only, reusing the policy at `cmd/bd/proxied_server.go`; non-loopback requires
`--allow-non-loopback`. Host-header allowlist. `netutil.LimitListener` connection cap, a bounded
semaphore wait mapping to 503, a per-request deadline, explicit `*sql.DB` pool limits. Hooks do not
fire and the per-command auto-commit globals are never touched.

`/v0/beads/context` shares one projection with `bd context` in which `SyncRemote`, the server
endpoint and all absolute paths **have no member at all** — the exclusion is structural, not
remembered. Remote URLs routinely embed credentials.

Actor on a claim is caller-asserted provenance for the audit trail, not authenticated identity; the
CAS is a correctness fence, not an authorization boundary. Loopback binding bounds the blast radius.

## What is NOT converged

`bd ready` and `bd list` are not on the Reader role. They hand their filter to `--claim`, `--gated`,
`--explain`, `--mol`, `--watch` and the hierarchical `--parent` tree, none of which the role
expresses, and they stamp `--max-rows`, which is deliberately off the contract because a server has
no business honouring a client's. Widening the role or forking each command in two were both
rejected. They share construction on every route, and execution via `workapi.FinishPage` — `bd list`
on both routes except the `--parent` tree, `bd ready` on its proxied route only. `bd ready`'s direct
route cannot join: it publishes a `Total` from a second counting query that the role has no member
for.

`internal/httpapi/doc.go` and `issueops/reader.go` carry the full claim, structured
SHARED / ENFORCED-and-by-what / NOT-ENFORCED.

## Verification

Protocol corpus byte-unchanged, verified structurally and behaviourally with
`BEADS_PROTOCOL_REQUIRE_DOLT=1` so a missing container hard-fails rather than skipping. Smoked
end-to-end against a real external Dolt server: all six operations, concurrent-claim race yielding
exactly one 200 and one Dolt commit, and the ready sort default proven identical on both surfaces
using a deliberately non-priority creation order.

## Note on scope

Large for one PR because the shared seam and its first consumer are not separable without shipping
an unproven abstraction. The eight underlying branches were built and reviewed independently and
kept bisectable; commit history is preserved if reviewing incrementally is easier.


julianknutsen:
## Pre-merge review record

This branch was reviewed continuously as it was built rather than in one pass at the end. Recording
what the reviews actually found, including the things that were wrong, so a reader can judge how much
weight the "pre-reviewed" claim carries.

Review passes: a 3-draft design panel scored by 3 judges on distinct lenses; a 4-persona council
(maintainer / SRE / consumer / security) on the design, returning 2 `request-changes` and 2
`approve-with-changes` across 27 required changes; and per-slice councils on each implementation diff.

### Defects caught before they shipped

**The `ready` sort default was wrong in the design document.** The spec initially froze `hybrid`.
That is only the storage layer's empty-string fallback (`sqlbuild/ready.go:57`) and is unreachable
from the CLI — `cmd/bd/ready.go:801` registers `--sort` with default `priority`. Because `limit`
truncates, this changes the item **set**, not just the order. Now `default: priority`, pinned to the
cobra flag registration by a test proven to fail. The handler must send it explicitly; forwarding
`""` silently reintroduces the fallback while the spec still reads correctly.

**`oapi-codegen` v2.6.0 drops `x-go-type` when a component also carries `allOf`.** Probed before the
spec was written: an `allOf`-composed `IssueWithCounts` generated `type IssueWithCounts = types.Issue`
— `ReadyPage.Items` would have shipped as `[]types.Issue` with the counts gone. A two-member variant
generated a full mirror struct, the banned second wire struct. Pinned schemas are therefore flat, and
`pinning.go` carries assignment-based assertions so a recurrence is a build failure, not a silent
alias.

**`context.DeadlineExceeded` satisfies `net.Error`**, so a naive "any `net.Error` is
`db_unavailable`" mapped a tripped request deadline to 503 instead of 500.

**Not-found was modelled wrong in every design draft.** On the UOW seam
`db.issueSQLRepositoryImpl.Get` returns `(nil, sql.ErrNoRows)` — never `storage.ErrNotFound`, and not
`(nil, nil)`. Normalization lives in the shared function, not a handler, so no future caller can
reintroduce it. Without it this ships missing-issue-500, the mirror image of the everything-is-404 bug
it was written to avoid.

**A wildcard bind disabled the Host allowlist entirely** — `--allow-non-loopback --addr 0.0.0.0:PORT`,
the most common spelling, removed the DNS-rebinding defence.

**Connection-cap saturation was silent.** `LimitListener` parks `Accept` at the cap, so new
connections including `/healthz` probes hang with nothing emitted.

**`bd serve` lost its database ~5.5 minutes after the last activity while `/healthz` stayed green** —
the pool empties at 5m, the proxy child's 30s idle watcher exits, and the DSN pins a dead ephemeral
port for the process lifetime. The reviewer reporting it claimed proxied mode was immune; a proxied
control died the same way, because `init.go:383-384` only writes `IdleTimeoutNever` when the operator
passes the flag explicitly.

**This merge did not compile.** #5206 added close-policy arms against the old four-value attempt
signature while this branch had collapsed that function onto `RunTxResultWithin`. Git merged it
cleanly. #5206's only proxied coverage was a unit test that would not have caught it, so
`TestProxiedServerUpdateClosePolicy` was added — four end-to-end subtests proving both refusals fire
exactly once, do not write, and cross under `--force`.

### Reviewers who were wrong

Recorded because they were argued down with evidence rather than accommodated:

- "Make unknown `issue_type` a 400 on `ready`" — `bd ready` does not validate `--type` at all, so
  that would be a *fresh* CLI divergence, the same trap as the sort default. `bd list` does validate;
  the asymmetry is now documented on both parameters.
- "The parity oracle does not run in the PR gate" — true clause, false conclusion. It runs in
  `pr-risk.yml`, verified by executing the lane's own shard discovery across all fifteen shards.
- "`timedProvider.IssueReader` should recurse like the other decorators" — it has no `Lifecycle`
  sibling, and recursion is structurally impossible there. Applying it made every read route log
  `uow_ms=0.000`. The regression is now a test failure instead of a comment.

### Known limits

The anti-drift guarantee forbids **writing** a filter in the guarded trees; it does not prove every
filter came from a builder. Test files are exempt because the oracles hold filters to inspect them.
And no status check on `main` is GitHub-required — these gates run on every PR and aggregate into
`ci-gate`, but a red gate binds by convention, not by protection.

Issue #5213, opened 2026-08-01
federation add-peer silently clears a stored sovereignty tier when --sovereignty is omitted


arcaven:
Re-running `bd federation add-peer` is the natural way to repair a peer's stored credentials (it is also the remediation the #5085 follow-up work points people at), and doing so without `--sovereignty` silently clears the peer's stored sovereignty tier. The tier is the attribute docs/multi-agent/federation.md frames as serving regional-compliance rules, and `federation_peers` replicates with the database, so the cleared value travels to peers.

**Expected:** re-running add-peer without `--sovereignty` leaves the stored tier alone (the flag was not given, so no intent to change it was expressed), or at least warns that it is clearing one.

**Actual:** the upsert overwrites the stored tier with the empty string. No warning, exit 0, and `--json` reports `"sovereignty": ""`.

<details><summary>Repro: script plus one fresh top-to-bottom run at pinned main 3385ceef9 (bd 1.1.0 dev build, dolt 2.2.2, macOS arm64)</summary>

Script:

```bash
#!/usr/bin/env bash
# Repro: bd federation add-peer silently clears a stored sovereignty tier
# when re-run without --sovereignty.
#
# Self-contained: mktemp workspace, HOME overridden, BEADS_* unset, cleanup trap.
# Runs as the invoking (non-root) account. No network. No port 3307.
set -euo pipefail

unset BEADS_DOLT_AUTO_START BEADS_DOLT_SERVER_HOST BEADS_DOLT_SERVER_PORT BEADS_DOLT_PASSWORD || true

BD_BIN="${BD_BIN:?set BD_BIN to the bd binary built from the pinned SHA}"
WORK="$(mktemp -d "${TMPDIR:-/tmp}/sovrepro.XXXXXX")"
cleanup() { rm -rf "$WORK"; }
trap cleanup EXIT

export HOME="$WORK/home"
export BD_NON_INTERACTIVE=1
export CI=true
export NO_COLOR=1
mkdir -p "$HOME" "$WORK/repo" "$WORK/peer-beta" "$WORK/peer-gamma"

step() { printf '\n===== %s =====\n' "$*"; }

step "0. environment"
echo "uid:      $(id -u) (non-root)"
echo "HOME:     $HOME"
echo "workdir:  $WORK/repo"
echo "bd:       $("$BD_BIN" version)"
echo "dolt:     $(dolt version | head -1)"
[...]

PR #5214, opened 2026-08-01
fix(federation): actionable error when peer credentials cannot be decrypted (stacked on #5085)

arcaven:
This is follow-up 2 from the #5085 review, and thank you for spelling out exactly what the error should say; that made the shape of the fix easy. A database pulled to a second machine carries `federation_peers` rows its local `.beads-credential-key` cannot decrypt, and every peer operation died with a raw cipher error while `bd federation status` blamed the network. The failure semantics were right; the message said nothing about the machine-local key or the fix.

**Stacked on #5085.** The change itself is one commit, a0185bbdf (7 files, +356/-6 over #5085's head, `git diff --shortstat e87e5cbe3 a0185bbdf`). The branch also carries a merge of current main (32c46df64, added 2026-08-01 after main moved the guarded issue-op sentinels into the `issueops` package, which conflicted with the sentinel this PR adds); the only manual resolution keeps `ErrCredentialKeyMismatch` and its helper in `internal/storage/storage.go` beside main's new alias block. All suites re-ran green after the merge.

## What changed

- `internal/storage` gains the sentinel `ErrCredentialKeyMismatch` plus one wording helper, `CredentialKeyMismatchError(keyFile, cause)`, so both backends emit a single wording from one place and callers can `errors.Is` the local-key case apart from an unreachable peer. Both the sentinel and the raw cipher error stay reachable through the `%w` chain.
- embeddeddolt wraps at its single decrypt funnel, so peer resolution for the remote verbs, `GetFederationPeer`, and `ListFederationPeers` all carry the text; package dolt wraps its two decrypt read paths, so server mode reports the same thing. `GetFederationPeer` wraps now name the peer, matching `ListFederationPeers`.
- `bd federation status` classifies the sentinel and prints `Local credential error:` instead of `Unreachable:` for it. The `--json` shape is unchanged and now pinned by a test (exact key sets, top level and per peer).

The message names the key file as context rather than asserted cause, since an AES-GCM open also fails on a tampered blob or a legacy-key row; re-adding the peer is the fix in each of those cases too, so the remediation holds even where the parenthetical does not. Rendered example from the machine-B repro:

```
✗ Local credential error: resolve peer credentials: decrypt password for peer team: stored peer credentials cannot be decrypted with this machine's credential key (the key file .beads-credential-key is machine-local and does not replicate with the database); re-run 'bd federation add-peer <name> <url> --user <user>' on this machine: cipher: message authentication failed
```

**When the error fires is unchanged**: same call sites, same fail-closed semantics, no fallback to ambient credentials. The fall-back-vs-fail-closed policy question from #5207's notes still stands and this PR is useful under either answer.

Two judgment calls that are yours to overrule, happy to rework either: the `Local credential error:` label wording is ours, and the one-line message is long (roughly 330 characters rendered). The natural split would keep the cause and key-file context in the library error and move the add-peer instruction to a cmd/bd hint line, at the cost of the library error losing the remediation. Also worth knowing before anyone follows the remediation on a peer with a sovereignty tier: a bare re-add currently clears the stored tier (filed separately as #5213).

## Traceability to the review

| Review sentence (follow-up 2) | Where it is answered |
|---|---|
| "the error should say what actually happened and what to do: that the credential key is machine-local, and to re-run bd federation add-peer" | `storage.CredentialKeyMismatchError` (internal/storage/storage.go), used by both backends |
| "bd federation status renders it as Unreachable:, pointing at the network instead of the local key" | `errors.Is` classification in cmd/bd/federation.go, `Local credential error:` line |
| "Hard-fail is a defensible choice, server mode already behaves this way" | semantics untouched; server mode's two read paths now carry the same wording so the two backends match in message as well as behavior |
| "a test pinning the decrypt-failure branch would keep it from being changed by accident later" | pinned in both backends: embeddeddolt against a real store with a rotated key (`TestDecryptPassword_KeyMismatchNamesTheLocalKey`, `TestWithPeerAuth_KeyMismatchFailsClosed`), dolt via sqlmock (`TestFederationPeerDecryptKeyMismatchNamesTheLocalKey`), plus the gated CLI subtest `status_credential_key_mismatch` incl. the `--json` shape |

## Tests

`go build -tags gms_pure_go ./...`, vet, gofmt, `./internal/storage/` and targeted `./internal/storage/dolt/` suites, the full `./internal/storage/embeddeddolt/...` package, and the `BEADS_TEST_EMBEDDED_DOLT=1` federation CLI harness (12 subtests) all pass. A non-vacuity check was run during development: reverting the production hunk makes the new tests fail.

One coordination note: this PR and the follow-up 4 PR both add imports to `federation_credentials_test.go`, so whichever lands second will need a trivial refresh there; happy to do that promptly.

---


maphew (commented):
Reviewed at `adead21160c9` across the full stacked diff.

The actionable credential-key sentinel/status work is valuable, but three correctness gaps remain:

1. `internal/storage/embeddeddolt/federation.go:98` returns bare `ciphertext too short` for truncated ciphertext, bypassing `CredentialKeyMismatchError`. Status still misclassifies that local corruption as `Unreachable`; wrap this branch and add a short-ciphertext case.
2. The ambient/no-credential fallbacks at `federation.go:173,178` run outside `federationEnvMutex`. A concurrent stored-peer operation can temporarily replace `DOLT_REMOTE_PASSWORD`, causing a plain remote to observe another peer's password. Serialize every callback path.
3. Credentials are selected solely by remote name, while `AddRemoteIfNotExists` preserves an existing same-name remote even when its URL differs. Federation verbs can therefore transmit a stored password to an unrelated host. Verify the live remote URL equals `FederationPeer.RemoteURL` before applying credentials, or fail closed on divergence.

Verification:

- Focused embedded tests — local build blocked by missing `unicode/uregex.h`; no verdict.
- `CGO_ENABLED=0 go test -tags gms_pure_go` targeted at the changed server-store credential code — PASS (`ok`, 2.361s).

Disposition: **fix-merge on the contributor branch** (or cherry-pick the unique sentinel work onto the corrected #5085 stack). Preserve the contributor commits and attribution; a request-changes review is unnecessary if maintainers absorb these fixes.

arcaven:
Thank you.

steveyegge:
## Review verdict: MERGE AFTER REBASE + one hunk

Content-reviewed against current main — nothing here is subsumed (`ErrCredentialKeyMismatch`/`CredentialKeyMismatchError` are absent from main; #5216 touched adjacent code only). The design holds up: double-`%w` wrapping so both sentinel and cause survive `errors.Is`/`As`, no secret material anywhere in the chain (path, peer name, cipher error only), `--json` shape unchanged and pinned by an exact-keyset test, and the deliberate wrong-key/corrupted-blob/legacy-key merger is right — I tried to fault it and can't, since the remediation is genuinely identical for all three and absent-creds stays naturally distinct.

Two things before merge:

1. **Rebase needed** — the PR is CONFLICTING, but it's one file (`federation_credentials_test.go`, add/add), a pure artifact of #5085 landing as a squash: main "added" the file, your branch adds it with ~147 extra test lines. The resolution is a mechanical union. Everything else auto-merges. (Same shape as #5215's parked resolution — and same org-fork 403 preventing me from pushing it to your branch. If the rebase is a hassle, say so and I'll park a resolution branch upstream you can fast-forward from.)
2. **One hunk to move while you're in there:** package dolt wraps at the *callers* (`credentials.go:347-355, 395-403`), so any `decryptPassword` error — including `"credential encryption key not initialized"` — gets labeled a key mismatch with the re-add remediation, which is the wrong diagnosis for that edge (init/permissions, not re-adding the peer). The embeddeddolt side got this right by wrapping at the decrypt funnel for GCM-open + short-ciphertext only; move the dolt-side wrap inside `decryptPassword` around `decryptWithKey` to match.

Landing order note: #5207 widens fail-closed decrypt to every remote verb, and this PR is what makes that failure actionable — so this merges right behind it. CI: the fork approval gate means zero runs have ever executed here; I'll approve on your fresh head.

arcaven:
Thank you for the adversarial pass, and for checking the merger against the wrong-key/corrupted/legacy cases rather than taking the claim. Both items are done at `3a7828f89`; the PR is `MERGEABLE` again.

**1. Rebase.** Merged main rather than rebasing, per our no-rewrite policy. Your read of the conflict was exact: one add/add in `federation_credentials_test.go`, and the two sides differ by pure addition (146 lines, zero edits to the seven shared tests), so the union resolution is this branch's copy verbatim. I diffed both stages to confirm that before taking either side rather than assuming the superset.

**2. The dolt-side hunk.** Moved into `decryptPassword` around the `decryptWithKey` call, and both caller wraps at `credentials.go` dropped to a plain `%w` that only names the peer. There is now exactly one `CredentialKeyMismatchError` call site in the package.

One thing your note made me check that I would have otherwise missed: I first pinned this with a direct `decryptPassword` unit test, then realized that test passes both before and after, because the old wrap lived at the callers. Rewrote it to go through `GetFederationPeer`/`ListFederationPeers`, which fails at the old shape and passes at the new one. The edge is genuinely reachable rather than theoretical: `initCredentialKey` returns nil without setting a key when `beadsDir` is empty, so `ensureCredentialKey` reports success and `decryptPassword` still finds no key, which is exactly the case the caller wrap was labelling a mismatch and telling the operator to re-add the peer for.

I also left `decryptWithKey` itself unwrapped deliberately, since `migrateCredentialKeys` calls it with an old key on purpose during rotation; classifying there would report a rotation miss as a mismatch. There is a regression pinning that too.

Note for reading the history: the hunk move is inside the merge commit rather than a separate one. The Files-changed view against the merge base is clean (7 files, +483/-8), and `git diff main...3a7828f89 -- internal/storage/dolt/credentials.go` shows just that change if it is easier.

Verified at `3a7828f89`, ambient `BEADS_DOLT_*` scrubbed: `internal/storage/dolt` and `internal/storage/embeddeddolt` credential suites pass, `cmd/bd` federation tests pass including `TestEmbeddedFederation` under `BEADS_TEST_EMBEDDED_DOLT=1`, and `go build ./...` plus gofmt are clean. Zero CI runs have executed here, as you noted, so your approval of the run is still the real gate.

arcaven:
CI caught one thing, now fixed at `f931843f6`: `Labelling` in a comment I added, which your `misspell` linter rejects. All three red checks were that single word (Lint, PR Lint wrapper, and CI Gate depending on them). Also corrected `labelled` in the test file, which the linter had not reached.

Comment text only, no behavior change. `golangci-lint run ./internal/storage/dolt/ ./internal/storage/embeddeddolt/ ./cmd/bd/` is 0 issues locally now, which is a gate I should have run before the last push rather than after.

bee-ghosttrack:
CI triage (bee, maintainer seat): the build red is in this diff — `internal/storage/storage.go:110: undefined: fmt` (missing import). Everything downstream cascades from that.

arcaven:
Fixed and caught up to main. Thank you for the triage, and you were right that the break is in this diff, though not in either side of it on its own.

It is a semantic merge artifact. Main dropped `"fmt"` from `internal/storage/storage.go`'s import block once its last `fmt` use there went away; this branch adds a `fmt.Errorf` at line 110. Git took the shorter import list from main and the new call site from here and reported no conflict, so the branch tip built clean in isolation and only the merge was red. Restored the import.

The one real conflict was the sqlmock alias in `internal/storage/dolt/credentials_test.go`, which main removed from that import block and the tests here reference eight times, so the resolution keeps it.

`go vet ./internal/storage/ ./internal/storage/dolt/` is clean and `go test ./internal/storage/...` passes, apart from the `dolt` package, which wants a Dolt server it can administer and fails the same way on upstream main in my environment.

bee-ghosttrack (approved):
**Verdict: APPROVE** (bee, maintainer seat, R1 of head `2780dd7663`; the 09-03 build red I flagged, `storage.go: undefined: fmt`, is fixed at this head).

Read the diff against `main` and the call sites at this head. The design is right: one sentinel and one wording helper in `internal/storage`, wrapped at each backend's single decrypt funnel so every reader (peer resolution for the remote verbs, `GetFederationPeer`, `ListFederationPeers`) reports the local-key case, and both backends' callers name the peer. The two deliberate non-wraps hold up: a nil key in package `dolt` stays "not initialized" (re-adding the peer would do nothing there), and `decryptWithKey` stays bare so the rotation reader's old-key misses are not classified as mismatches. embeddeddolt's short-ciphertext branch joins the same class, which is the right call since `federation status` would otherwise read it as an unreachable peer. `federation status` carries the signal in an unexported field, so the `--json` shape is pinned unchanged by `assertExactJSONKeys` on both the envelope and the peer object. Both `%w` verbs in `CredentialKeyMismatchError` keep the sentinel and the cipher cause reachable through `errors.Is`.

Since CI never reached the test jobs on this head (see below), I ran them locally (macOS arm64, go1.26.5):

- `go test ./internal/storage/dolt -run 'TestFederationPeerDecryptKeyMismatchNamesTheLocalKey|TestFederationPeerUninitializedKeyIsNotAMismatch|TestDecryptWithKeyStaysUnwrappedForRotation' -count=1` → PASS (sqlmock, no live Dolt).
- `CGO_ENABLED=1 go test ./internal/storage/embeddeddolt -run 'TestDecryptPassword_|TestWithPeerAuth_KeyMismatchFailsClosed' -count=1` → PASS 3/3 (Get + List subtests each).
- `BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test ./cmd/bd -run 'TestEmbeddedFederation$/status_credential_key_mismatch' -count=1` → PASS (9.3s). Without the opt-in env var that suite skips silently, which is why a green `./cmd/bd` run is not evidence for this case.
- Red proof: unwrapping `decryptPassword` in package `dolt` (`return "", err`) fails both subtests of `TestFederationPeerDecryptKeyMismatchNamesTheLocalKey` on `want errors.Is storage.ErrCredentialKeyMismatch`, so the tests bite on the wrap itself.
- `gofmt -l` clean on all 7 files; `go vet` clean on `internal/storage`, `dolt`, `embeddeddolt`, `cmd/bd`.

CI reds on this head are not the diff's:

- `Build (Embedded Dolt)` compiled and tested for 7 minutes, then died on `Failed to FinalizeArtifact: (403) Forbidden`; every downstream test shard skipped on the missing artifact. Same class as the 403 on unrelated PRs this month.
- `Check doc flags freshness` and `PR Policy (wrapper timing)` fail on `Last reviewed date is stale: 2026-06-09 (91 days old)`: a docs-policy clock that expired on 09-08 and is green on 09-12 PR runs against current main. A rebase (the branch carries 13 commits of merges from the #5085 era; a squash-merge is fine) or a re-run clears both.

One non-blocking note: `federation status` still exits 0 with a local-credential failure present, matching its pre-existing behavior for unreachable peers. If we ever want scripts to notice, that is a separate change.

PR #5215, opened 2026-08-01
feat(embeddeddolt): warn when a stored peer suppresses an ambient DOLT_REMOTE_PASSWORD (stacked on #5085)

arcaven:
This is follow-up 4 from the #5085 review: a peer saved with a username but an empty password (Enter at the interactive prompt) deliberately suppresses an ambient `DOLT_REMOTE_PASSWORD`, which is the right security default, but an operator whose pre-#5085 setup authenticated off the environment gets an auth failure with no hint about why. Thanks for calling out the exact trigger triple; the guard is a direct transcription of it.

**Stacked on #5085.** The branch is #5085's head plus one commit (575154c32); it collapses to that one commit once #5085 merges. Diff vs #5085: 2 files, +169 insertions, no existing lines changed.

## What changed

One stderr line, emitted when the resolved peer has a username, no stored password, and `DOLT_REMOTE_PASSWORD` set:

```
Warning: peer "town-beta" stores a username with an empty password, which overrides the ambient DOLT_REMOTE_PASSWORD for this operation; store a password with 'bd federation add-peer town-beta <url> --user sync-bot --password <password>'.
```

It fires inside `withPeerAuth` while `federationEnvMutex` is held and before `overrideEnv` rewrites the pair, so the value it reads is genuinely the ambient one and never another peer's temporarily installed password. No credential behavior changes. The writer is a package var defaulting to `os.Stderr` (test seam), and the hint deliberately says "store a password with add-peer --password" rather than suggesting a bare re-add, because a bare re-add currently clears a stored sovereignty tier (filed separately as #5213).

## Noise, disclosed honestly

In the misconfigured state this prints once per wrapped operation: `bd federation sync` shows it twice (Fetch, then PushTo) and `bd federation status` once per such peer. One caveat worth your judgment: the trigger can also describe a working setup, since a username-only peer against a passwordless SQL user is one of add-peer's own help examples, and permanently exporting the ambient pair is the documented Hosted Dolt auth. An operator with both sees the line with nothing broken. This ships the shape you described; if you would rather it fire only after an auth failure, happy to rework it that way.

## Traceability to the review

| Review sentence (follow-up 4) | Where it is answered |
|---|---|
| "a peer saved with a username but an empty password ... now deliberately suppresses an ambient DOLT_REMOTE_PASSWORD, correct as a security default" | guard conditions in `warnStoredPeerSuppressesAmbientPassword`; suppression itself untouched |
| "a user whose pre-PR setup worked off the env gets an auth failure with no hint" | call site in `withPeerAuth`, placed under the mutex before `overrideEnv` so the ambient value is still readable |
| "A one-line stderr/debug warning when a peer has a username, no stored password, and DOLT_REMOTE_PASSWORD set" | the single `fmt.Fprintf` line; the three conditions are the three guard clauses |

## Tests

The table test pins the exact rendered line (verbatim equality against the writer) for the firing case and asserts silence for: stored password present, ambient unset, ambient set-but-empty, and a direct-call row with an all-empty peer that pins the guard boundary. Two falsification checks were run during development: removing the ambient check fails two rows, and removing the username clause fails the direct-call row. Full `./internal/storage/embeddeddolt/...` suite, build, vet, gofmt all pass with `-tags gms_pure_go`.

One coordination note: this PR and the follow-up 2 PR both add imports to `federation_credentials_test.go`, so whichever lands second will need a trivial refresh there; happy to do that promptly.

---

**Maintainer edits:** `maintainer_can_modify` is on, but our fork is organization-owned and GitHub only honors maintainer pushes on user-owned forks, so a push here will 403. If you want to push a change, park it upstream as `bee/pr<N>-resolution` (the convention you used here) and I will fast-forward from it. No need to ask first. We are moving to a personal-account fork to remove this entirely once the current PRs land.


maphew (approved):
Reviewed at `575154c3204`, including the full stacked diff and the incremental warning commit on top of the already-reviewed #5085 base.

No correctness findings. The warning is emitted while `federationEnvMutex` is held and before the environment override, so it observes the caller's ambient password rather than another peer operation's temporary state. It names only the peer and username, never the credential value; stored credentials continue to win as an atomic pair. The actionable case and all silent boundaries are pinned exactly.

Verification:

- `go test ./internal/storage/embeddeddolt -run 'TestWithPeerAuth' -count=1` — local build blocked because this Windows host lacks the ICU header `unicode/uregex.h`; no test verdict.
- Blocking upstream preflight — GitHub reported no failed or pending checks, but the merge state remained `UNSTABLE`; no merge action was taken.

Disposition: **approve / merge when the normal preflight gate is clean**.

steveyegge:
This branch went CONFLICTING against main after #5085 merged in its final form (`withPeerAuth` was restructured there: `verifyPeerRemoteURL` fail-closed check + the env-mutex now covering the fallback paths).

I've resolved the merge on your behalf — the resolution keeps main's structure and inserts `warnStoredPeerSuppressesAmbientPassword` after the mutex is taken in the stored-credentials path, and grafts your warning test onto main's test file (which gained the URL-verification tests since). Net diff vs main is exactly this PR's +165 lines; the full `TestWithPeerAuth*` set passes locally, `go vet`/gofmt clean.

I couldn't push it to this branch directly (org-fork permissions 403 maintainer pushes — same as #5087), so it's parked as [`bee/pr5215-resolution`](https://github.com/gastownhall/beads/tree/bee/pr5215-resolution) (head `8c582c2ae`, first parent = your `575154c32`). To take it, fast-forward your branch from it:

```bash
git checkout fix/peer-empty-password-warning
git pull https://github.com/gastownhall/beads.git bee/pr5215-resolution
git push
```

Your approval stands; once the branch updates and CI is green this merges.

arcaven:
Thank you for resolving the merge, and for parking it where I could take it without a rewrite. Fast-forwarded to `8c582c2ae`; the branch now carries your resolution unchanged.

I reviewed the resolution before taking it rather than merging it blind. The placement is right: the warning fires after `federationEnvMutex` is taken and before `overrideEnv` rewrites the pair, so the `DOLT_REMOTE_PASSWORD` it reads is the ambient one and not another peer's installed value, which is what the concurrency test would otherwise catch. Net diff vs your merge parent is +165 in the two expected files, matching your note.

Verified at `8c582c2ae`: the full `TestWithPeerAuth*` set passes locally under cgo, including `TestWithPeerAuth_WarnsWhenStoredPeerSuppressesAmbientPassword` and the concurrency case. The branch also merge-trees clean against current main (`741c9e415`), which has moved since your resolution.

arcaven:
Caught up to main, 329 commits, no conflicts: main has not touched `internal/storage/embeddeddolt/federation.go` or its credentials test, which are the only two files here.

Verified on darwin/arm64: gofmt and go vet clean, `go test ./internal/storage/embeddeddolt/` passes. That package needs ICU headers to build on macOS, so I ran it with `CGO_CPPFLAGS` pointed at the brew `icu4c` prefix; without them the cgo dependency fails to compile on upstream main too, so it is an environment note rather than anything about this branch.

bee-ghosttrack (approved):
**Verdict: APPROVE** (bee, maintainer seat, R1 of head `11db4f7822`).

Read the diff and the surrounding `withPeerAuth` at this head. The guard is a direct transcription of the trigger triple: `withPeerAuth` returns before the helper when the row is absent or both fields are empty, so the helper only ever sees a non-nil peer with at least one field set, and its own `Username == "" || Password != ""` clause covers the password-only shape. The read of `DOLT_REMOTE_PASSWORD` happens under `federationEnvMutex` and before `overrideEnv`, so it is the ambient value and never another peer's installed one. The warning prints peer name and username only, never the ambient password. Writer seam defaults to `os.Stderr`; `newPeerAuthTestStore` points it at `io.Discard` so the rest of the package's `withPeerAuth` cases stay quiet.

Ran on this head (macOS arm64, go1.26.5, `CGO_ENABLED=1`):

- `go test ./internal/storage/embeddeddolt -run TestWithPeerAuth -count=1` → ok (114s); the new table runs 7/7 subtests PASS.
- Red proof: deleting the `warnStoredPeerSuppressesAmbientPassword(peer, p)` call fails `username_with_empty_stored_password_suppresses_ambient_password` and nothing else, so the test bites on the one line it pins.
- `gofmt -l` clean on both files; `go vet ./internal/storage/embeddeddolt/` clean.

CI reds on this head are not the diff's:

- `Test (Embedded Dolt Cmd 8/20)` died in 5s on `Failed to ListArtifacts: (403) Forbidden` (artifact download, the class we have seen across unrelated PRs this month).
- `Check doc flags freshness` and `PR Policy (wrapper timing)` both fail on `Last reviewed date is stale: 2026-06-09 (91 days old)`: a time-based docs-policy clock that ticked over on 09-08 and is green on 09-12 PR runs against current main. A rebase or a re-run clears it.

Two non-blocking notes: the suggested `--user %s` in the warning is unquoted, so a username with spaces would need quoting by hand; and the line fires on every remote verb and every `federation status` pass for such a peer, which is the right default for a security-relevant surprise but may want a once-per-process guard later if operators find it chatty. Neither needs a change here.

PR #5216, opened 2026-08-01
fix(dolt): setFederationCredentials restores prior env instead of unsetting

arcaven:
This picks up the aside from the #5085 approval, gratefully: "the cleanup in package dolt's own setFederationCredentials ... unconditionally Unsetenvs and clobbers pre-existing ambient values, your overrideEnv shape is the fix that package wants too." It is independent of the #5085 branch (that PR touches only embeddeddolt), so it is filed from current main rather than stacked.

The user-visible defect: any SQL-path remote operation in server mode destroyed an ambient `DOLT_REMOTE_USER`/`DOLT_REMOTE_PASSWORD` pair the caller never owned. Exporting that pair permanently is the documented Hosted Dolt setup (docs/cli-reference/dolt.md), so the next operation that relied on it lost its credentials with nothing to point at the cause.

Diff vs main (3385ceef9): 2 files, +162/-15.

## What changed

- New package-local `saveEnv(key)`: capture with `os.LookupEnv`, return a closure that restores a was-set value or unsets a was-unset one. This is the restore half of embeddeddolt's `overrideEnv`, ported locally (that helper lives on an unmerged branch, and dolt should not import embeddeddolt either way).
- `setFederationCredentials` captures both vars before mutating and its cleanup restores them in reverse order instead of unconditionally unsetting. The conditional-set rule (only non-empty values are written) is unchanged.
- `setS3ChecksumEnv` already hand-rolled the identical save/restore and now calls the one helper. That is a deliberate in-file dedup one step beyond the literal aside; happy to split it out if you would rather this touch only the federation path.
- Comment precision around `withRemoteOperationEnv`: the restore-before-unlock ordering is now noted at both the Lock line and the cleanup defer, and the guarantee is scoped to operations that take `federationEnvMutex` (ambient readers that take no lock, like the config load and remotecache, are outside it and were before).

One thing deliberately not changed: package dolt's SQL path still lets an ambient value through when a stored field is empty, unlike its own CLI-path `applyToCmd` and unlike embeddeddolt's stored-credentials-are-a-unit rule from #5085. Adopting the stricter rule here would change which credentials an operation presents, not how cleanup behaves, so it felt like its own decision rather than a rider on this one. Glad to file it separately if useful.

## Traceability to the review

| Review clause (the aside) | Where it is answered |
|---|---|
| "unconditionally Unsetenvs and clobbers pre-existing ambient values" | `saveEnv` captures before mutation; cleanup restores in reverse order (internal/storage/dolt/credentials.go) |
| "your overrideEnv shape is the fix that package wants too" | the ported save/restore shape, now the file's single implementation, reused by the S3 checksum path |

## Tests

Four new tests, all driven through `withRemoteOperationEnv` (the real entry point, holding the mutex, matching how the existing S3 restore tests are written): full ambient pair restored, partial pair (user set comes back, password unset stays unset), no ambient pair (stored values do not linger), and both cleanups combined (credentials plus S3 checksum, all three vars restored). An A/B falsification was run during development: reverting the restore hunk makes the combined test fail with the ambient value gone. Build, vet, gofmt, and the targeted `./internal/storage/dolt/` credential suites pass with `-tags gms_pure_go`.


maphew (approved):
Reviewed at `c9f21b2a0`.

This is a narrow, correct fix. `saveEnv` preserves both halves of the environment contract—restoring an existing value and re-unsetting a previously absent one—and the cleanup order remains safe: the registered restores run in reverse order before `federationEnvMutex` is released. Reusing the helper for the S3 checksum override also removes duplicated state-restoration logic without changing behavior.

I found no correctness, compatibility, documentation, or maintainability blockers in the diff. The added tests cover fully ambient, partially ambient, initially absent, and combined credential/checksum states.

Focused validation:

- `go test ./internal/storage/dolt -run 'TestWithRemoteOperationEnv|TestPrepareDoltCLITransferCommand|TestApply(NoGitHooks|S3Checksum)' -count=1` — PASS.
- Blocking upstream preflight — review is still operationally gated by GitHub reporting `UNSTABLE`; it reported no failed or pending checks. This approval is not a merge action.

Disposition: **approve / merge once the normal preflight gate is clean**.

PR #5219, opened 2026-08-01
fix(wisp gc): apply the --age test to cascade children

AlexBelous:
## The contract that was broken

`bd mol wisp gc --help` states that a wisp counts as abandoned when it "has not been updated within `--age` **AND** is not closed". Parents were held to that contract. Children reached through the dependency cascade were not: a child was collected purely because something it transitively depends on is old, so `--age` had no meaning for children at all.

Raising the threshold is not a workaround. With `--age 720h`, a step created a moment ago is still destroyed when its parent is older than 720 hours — the second test covers exactly this.

## Relation to #4430

`isProtectedWisp` (added in #4430) covers steps that have already been picked up — `in_progress`, `blocked`, `hooked`. A *freshly created* step is still `open`, and `open` maps to `CategoryActive`, which that guard deliberately does not protect. This change closes the remaining path; it does not overlap with the existing guard.

## Repro

Both tests are beads-only unit tests in `cmd/bd`, driven through a fake `molReader`. No orchestrator, no external agent framework, no live database:

- `TestWispGCCascadeRespectsAge` — a stale parent no longer drags a fresh child into the abandoned set.
- `TestWispGCRaisingAgeDoesNotSaveFreshChild` — raising `--age` does not rescue the child, showing the threshold was inoperative for children rather than merely too low.

Both fail before the change and pass after.

## What this cost in practice

A molecule slung seconds earlier went down with all nine of its steps — 43 beads in one run — and the agent's hook emptied instantly. Two rigs of three were hit; the third survived only because a stop order reached it first.

## Verification

Full `go test ./cmd/bd/` run, with the patch and on a clean worktree at the same revision (`0e069115a`):

- identical failure sets both times — `TestConfigValidateReadOnlyIsHermetic`, `TestPrepareSelectedCommandContext_DoesNotMergeCallerConfigForUnsetKeys`, `TestPrepareSelectedCommandContext_RebindsTargetConfig`, `TestTestServerConnection`. These are pre-existing on this machine and unrelated to wisp GC; I did not investigate their cause.
- passing count 1533 → 1535, the difference being exactly the two tests added here; skip count 540 in both runs.
- `TestWispGCProtectsActiveWisps` runs green under `BEADS_TEST_EMBEDDED_DOLT=1` (33s, embedded dolt), so the existing cascade protection is not regressed.

## Scope

Single layer (`cmd/bd`), 4 lines of logic plus tests. Per the contributor guidelines the rationale lives in the commit message and here rather than in inline comments.

### DIFF
--- a/cmd/bd/wisp.go
+++ b/cmd/bd/wisp.go
@@ -880,6 +880,10 @@ func findAbandonedWisps(ctx context.Context, r molReader, cleanAll bool, ageThre
 				if isProtectedWisp(child, blockedSet, protectedStatuses) {
 					continue
 				}
+				// Children are held to the same --age contract as parents.
+				if now.Sub(child.UpdatedAt) <= ageThreshold {
+					continue
+				}
 				abandoned = append(abandoned, child)
 			}
 		}
--- a/cmd/bd/wisp.go
+++ b/cmd/bd/wisp.go
@@ -801,7 +801,9 @@ func runWispGC(cmd *cobra.Command, args []string) error {
 	for i, issue := range abandoned {
 		ids[i] = issue.ID
 	}
-	if err := deleteBatch(nil, ids, true, false, true, jsonOutput, false, "wisp gc"); err != nil {
+	// findAbandonedWisps already expanded the cascade under the --age contract;
+	// re-cascading at delete time would sweep fresh children past that filter.
+	if err := deleteBatch(nil, ids, true, false, false, jsonOutput, false, "wisp gc"); err != nil {
 		return HandleError("%v", err)
 	}
 	return nil
--- a/cmd/bd/wisp_gc_cascade_age_test.go
+++ b/cmd/bd/wisp_gc_cascade_age_test.go
@@ -1,0 +1,155 @@
+package main
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// cascadeAgeReader is a minimal molReader covering only the calls
+// findAbandonedWisps makes. The embedded nil interface makes any other call
+// panic loudly rather than return a silent zero value.
+type cascadeAgeReader struct {
+	molReader
+	issues   []*types.Issue
+	children map[string]bool
+}
+
+func (r *cascadeAgeReader) SearchIssues(context.Context, string, types.IssueFilter) ([]*types.Issue, error) {
+	return r.issues, nil
+}
+
+func (r *cascadeAgeReader) GetBlockedIssues(context.Context, types.WorkFilter) ([]*types.BlockedIssue, error) {
+	return nil, nil
+}
+
+func (r *cascadeAgeReader) GetCustomStatusesDetailed(context.Context) ([]types.CustomStatus, error) {
+	return nil, nil
+}
+
+func (r *cascadeAgeReader) IsInfraTypeCtx(context.Context, types.IssueType) bool { return false }
+
+func (r *cascadeAgeReader) FindWispDependentsRecursive(context.Context, []string) (map[string]bool, error) {
+	return r.children, nil
+}
+
+func (r *cascadeAgeReader) GetIssuesByIDs(_ context.Context, ids []string) ([]*types.Issue, error) {
+	want := make(map[string]bool, len(ids))
+	for _, id := range ids {
+		want[id] = true
+	}
+	var out []*types.Issue
+	for _, issue := range r.issues {
+		if want[issue.ID] {
+			out = append(out, issue)
+		}
+	}
+	return out, nil
+}
+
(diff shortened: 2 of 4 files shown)


maphew (approved):
Thanks — this is a model fix-PR: the contract violation is named precisely, the 4-line change lands exactly where it belongs, and `TestWispGCRaisingAgeDoesNotSaveFreshChild` is a genuinely sharp test, pinning the property that made this worse than a mis-tuned threshold (no value of `--age` protected a child at all). The panicking-fake pattern in `cascadeAgeReader` (embedded nil interface so un-stubbed calls fail loudly) is a nice touch, and the honest before/after failure-set accounting in the PR body made verification easy.

I traced the delete path end-to-end to confirm the selection-level filter is the effective gate, and it is on two of the three paths:

- **Embedded store** (default): `issueops.DeleteIssuesInTx` partitions wisp IDs out *before* cascade expansion — `FindAllDependentsInTx` runs only on regular IDs — so wisps are deleted exactly as selected.
- **Dolt server mode**: `DoltStore.DeleteIssues` routes wisps to `deleteWispBatch`, which deletes exact IDs, drops the dep edges pointing at them, and recomputes `is_blocked` — so a spared child is left clean, not dangling.

**One gap worth closing — proxied-server mode.** `runWispGCProxiedServer` deletes via `IssueUseCase().DeleteIssues(... Cascade: true ...)`, and the domain-level `deleteMany` re-expands with `FindAllDependents`, which traverses both `dependencies` *and* `wisp_dependencies` for all input IDs with no wisp partition. On that path the fresh child your filter just excluded gets re-added at delete time and deleted anyway — and since dry-run renders the filtered set, dry-run and the real deletion now disagree there. The delete-time cascade is pre-existing behavior, but the divergence is new with this change. The cheap fix looks like `Cascade: false` in the wisp-gc proxied delete, since `findAbandonedWisps` has already done the (now correctly age-filtered) cascade expansion — that also matches what the other two paths effectively do. Happy to take that as a follow-up if you'd rather keep this PR minimal; either way it shouldn't hold up the core fix.

Two smaller notes, neither blocking:

- In `TestWispGCCascadeRespectsAge`, the stale-child assertion is actually satisfied by the top-level age loop (SearchIssues returns all wisps, and a 48h-old open child qualifies on its own), not the cascade branch — after this change the cascade append only fires for children the top loop can't see (closed children without `--all`, or wisps past the 5000 search limit). Making that fixture `closed` would pin the cascade branch itself.
- Spared children now survive silently as free-floating wisps (their edges to the deleted parent are cleaned up, and they'll be collected once they age out). That's the same pattern the #4430 status guard already produces, so it's fine — but a dry-run/verbose line like "spared N children younger than --age" would help an operator make sense of a partially-collected molecule. Follow-up material.

Approving with the proxied-path follow-up flagged. Nice work tracking this down from the incident to the exact missing predicate.

steveyegge:
CI came back with one real failure, and it's an expected one: `TestProxiedServerWispGC/cascade_sweeps_dependent_step_wisps` (cmd/bd/wisp_proxied_integration_test.go:266) pins the exact behavior this PR removes — its child wisp is created fresh and never backdated, and the assertion reads "expected dependent child … cascaded even though it's not itself old". Under the new (correct) contract that child must now survive.

Suggested update to that subtest so it pins the fix instead of the bug:
- backdate a second child past the threshold and assert it IS swept (the cascade still works for genuinely abandoned children), and
- assert the fresh child is NOT in `deleted` (mirroring your unit test's `gc-fresh-child`/`gc-stale-child` pair at the integration level).

Everything else is green (Conformance, Regression, Cross-Version Smoke, PR workflows all passed; remaining shards were still queuing at last look). Happy to merge once the test is updated — or say the word and I'll push the test fix to your branch with attribution.

AlexBelous:
Test updated as suggested (a641cde) — and reshaping it the way you asked surfaced that the selection fix alone was incomplete.

With the fresh child asserted at the **table level** (not just absent from the `deleted` JSON), the subtest still failed: `deleted` listed only the parent and the backdated child, but `deleted_count` was 3 and the fresh child's row was gone. Both delete paths were re-deriving the cascade at delete time, past the age filter:

- embedded: `deleteBatch(..., cascade=true)` in the gc path
- proxied-server: `DeleteIssuesParams{Cascade: true}` re-expanding via `FindAllDependents` in the domain layer

Both now delete exactly the set `findAbandonedWisps` selected, so the `--age` contract lives in one place. Verified against the shared dolt container (`BEADS_TEST_PROXIED_SERVER=1`): all four `TestProxiedServerWispGC` subtests pass, `TestWispGC*` unit tests pass.

One out-of-scope observation for a possible follow-up: the `--closed` purge path (`runWispPurgeClosedProxiedServer`) still deletes with `Cascade: true`, so a closed parent can currently take **open** children with it — same delete-time-cascade shape as here, different entry point. Happy to file it separately if you agree it's real.

PR #5222, opened 2026-08-01
fix(setup): don't read a fenced @AGENTS.md example as a redirect stub

maphew:
## Why

#4838 taught `bd setup claude` to recognise a CLAUDE.md that is a thin stub importing AGENTS.md, and to write the managed beads block into AGENTS.md instead of duplicating it in the stub. Its review left two nits open. One (path-qualified `@./AGENTS.md`) was handled before that PR merged. This is the other.

`isAgentsImportStub` scans for `@AGENTS.md` on its own line and has no notion of code blocks, so a CLAUDE.md that **documents** the pattern in a fenced example is classified as a stub.

## What that costs

Not a misrouted write — a destructive one. Once the redirect activates, `stripStaleClaudeBlock` removes the managed block from CLAUDE.md on the reasoning that AGENTS.md now owns it. For a file that was in fact authoritative, a documentation example silently moves the block out of it.

The end-to-end test is the one that matters. Against `main`:

```
claude_test.go:1351: CLAUDE.md is authoritative and must carry the beads block;
                     a fenced example redirected it away
```

Skipping fenced regions also matches the semantics being detected in the first place: Claude Code does not expand an `@`-import shown as code, so a fenced directive was never going to make the file a stub.

## Deliberately narrow, in two places

Both are pinned by tests, because both are places a "more thorough" fix would be wrong:

- **Four-space-indented lines are still scanned.** Indentation is ambiguous in a way a fence is not — it is equally a list continuation, which is a plausible home for a real directive. Treating it as code would trade this false positive for a false negative.
- **A closing fence must match the character that opened the block**, so ` ``` ` inside a `~~~` block does not terminate it and let a following directive leak through. An unterminated fence swallows the rest of the file, which is what a Markdown renderer does too, so the conservative answer there is correct rather than accidental.

## Verification

Five of the seven new unit cases fail against `main`; the two that pass in both directions are over-correction guards ("real directive after a fenced example still counts", "indented directive still counts") and exist so a future change cannot go too far the other way. `./cmd/bd/setup` is green under `CGO_ENABLED=1 -tags gms_pure_go`, gofmt and vet clean.

## CI note

`main` is currently red on `Test (macos-latest)` for three unrelated failures — the fixes are #5210 and #5220. This branch touches only `cmd/bd/setup` and goes nowhere near them, so any macOS failure reported here is inherited from the base, not introduced. Flagging it explicitly because a red base is exactly how "those are pre-existing" stops being checked.

Beads: `mybd-xq68`.


maphew:
Reviewed at `eea7fa9dcab`.

One should-fix CommonMark edge case can reproduce the misclassification this PR is meant to prevent. `codeFenceMarker` records only the delimiter character, so a three-backtick line incorrectly closes a four-backtick fence. An `@AGENTS.md` example later in that still-open block can then activate the destructive redirect path.

Please track the opening delimiter length, require a closer of the same character to be at least that long, and add a four-backtick regression. I am not treating four-space indentation as a blocker: the code deliberately keeps that ambiguous case eligible as a real directive, and the surrounding comment documents the tradeoff.

Verification:

- `CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd/setup -run 'TestIsAgentsImportStub|TestInstallClaudeRedirectsToAgentsMD|TestCheckClaudeRedirectsToAgentsMD' -count=1` — PASS (`ok`, 0.836s).

Disposition: **fix-merge**.

PR #5228, opened 2026-08-01
ci: retry transient Dolt image pull failures

julianknutsen:
## What

Add a shared CI helper that retries the pinned Dolt SQL-server image pull up to three times, preserving the final `docker pull` exit status. Route all seven pull steps in the Main, PR, PR Risk, and Regression workflows through it.

The helper has automated coverage for immediate success, transient failures followed by success, exhausted retries, the executable bit, and complete workflow wiring.

## Why

Two historical proxied-server shards failed solely because Docker Hub timed out while serving `dolthub/dolt-sql-server:2.2.0`. A one-shot pull turns that transient registry failure into a red test shard even though the repository code is healthy.

The retry is bounded to three attempts with five-second gaps. It does not change the image version, job matrix, or which tests run.

Tracked by `bd-m0o1z` as part of the `origin/main` CI audit (`bd-qouky`).

## Verification

- `make test`
- `make ci-pr-policy`
- `make ci-pr-lint`
- `go test ./scripts -run 'Test(PullDoltImage|DoltImagePullWorkflows)' -count=1`
- `shellcheck scripts/ci/pull-dolt-image.sh`
- `actionlint .github/workflows/main.yml .github/workflows/pr.yml .github/workflows/pr-risk.yml .github/workflows/regression.yml`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -525,7 +525,7 @@ jobs:
       - name: Pull Dolt sql-server image
         # Keep tag in sync with internal/testutil/testdoltcommon.go:DoltDockerImage.
         # Without this, RequireDoltContainer reports doltNoImage and TestDomainDB skips.
-        run: docker pull dolthub/dolt-sql-server:2.2.0
+        run: ./scripts/ci/pull-dolt-image.sh
 
       - name: Download build artifacts
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
@@ -774,7 +774,7 @@ jobs:
         # The shared proxied harness requires the image cached locally
         # (isDoltImageCached); without it every proxied test self-skips. Keep the
         # tag in sync with internal/testutil/testdoltcommon.go:DoltDockerImage.
-        run: docker pull dolthub/dolt-sql-server:2.2.0
+        run: ./scripts/ci/pull-dolt-image.sh
 
       - name: Test proxied-server cmd shard
         env:
--- a/.github/workflows/pr-risk.yml
+++ b/.github/workflows/pr-risk.yml
@@ -80,7 +80,7 @@ jobs:
         # The shared proxied harness requires the image cached locally
         # (isDoltImageCached); without it every proxied test self-skips. Keep the
         # tag in sync with internal/testutil/testdoltcommon.go:DoltDockerImage.
-        run: docker pull dolthub/dolt-sql-server:2.2.0
+        run: ./scripts/ci/pull-dolt-image.sh
 
       - name: Test proxied-server cmd shard
         env:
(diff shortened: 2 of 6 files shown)


julianknutsen:
@maphew CI is fully green (84 passing checks, zero failures), and the branch is cleanly mergeable.

This is a nontrivial CI change authored by an agent, so `PR_MAINTAINER_GUIDELINES.md` requires a different human reviewer to perform a substantive review and merge it. Please review the bounded retry behavior and workflow wiring, and merge if satisfied.

maphew (commented):
Thanks for this — and for shipping `TestPullDoltImageHelperIsExecutable` alongside it. The executable bit is the thing that actually bites when a helper script lands from a Windows checkout, and it's the test most people skip. The `status=$?` capture inside the `else` is right too, and the exhausted-retries case asserting exit 42 proves the original `docker pull` status survives rather than the loop's.

Verified locally: `shellcheck` clean, `go test -run 'Test(PullDoltImage|DoltImagePullWorkflows)' ./scripts` passes, all 7 call sites have `actions/checkout` before them on `ubuntu-latest`, no un-migrated `docker pull` of the image remains, and the new tests do get picked up by `./...` in `pr-core.sh` and the main matrix.

Approving in substance. Follow-ups, none blocking:

1. **Stale sync comments.** All seven converted steps still say `Keep the tag in sync with internal/testutil/testdoltcommon.go:DoltDockerImage`, but the tag no longer appears there. The *why* lines above them are worth keeping; the sync clause now lives correctly in the script and can come out of the YAML.

2. **The tag has three homes and still no enforced guard.** `internal/testutil/testdoltcommon.go:14`, `scripts/ci/pull-dolt-image.sh:5`, and `doltSQLServerImage` in the new test. Centralizing 7 → 1 in the workflows was the hard part; the payoff is one assertion away — parse the `DoltDockerImage` const out of `testdoltcommon.go` and require the script to match. That turns a comment nobody enforces into something `go test ./scripts` fails on.

3. **The negative assertion is a snapshot, not a guard.** `strings.Contains(text, "docker pull "+doltSQLServerImage)` matches only the exact pinned tag, so a future `docker pull dolthub/dolt-sql-server:2.3.0` sails through — and a tag bump is precisely when someone touches those lines. Match on the repo prefix `docker pull dolthub/dolt-sql-server` instead.

4. **Hardcoded per-file counts.** `wantCalls{main: 2, pr: 2, pr-risk: 2, regression: 1}` means adding or retiring any Dolt-using job fails a test in `./scripts` with a message that doesn't point at the workflow edit that caused it. Item 3 carries the real invariant; relax these to "at least one" or drop the map.

5. **Untested Windows branch.** `path = binPath + ":/usr/bin:/bin"` under `runtime.GOOS == "windows"` never executes in CI — the Windows jobs are all `-run`-filtered (`^TestDocFreshness`, `^TestPRPreflight`, `^TestWorktreeRemove`) plus smoke. It *does* run in a local Windows `make test`, where a colon-separated `PATH` injected via `cmd.Env` into git-bash either works or doesn't with no middle ground. `TestPullDoltImageHelperIsExecutable` already skips Windows; consider skipping this one there too and deleting the branch, unless someone can confirm it on a Windows box.

On the backoff itself: 3 × 5s fixed is right for the failure mode you describe. Noting only that it won't help against Docker Hub's other common transient, a 429 whose window is minutes — if that ever shows up in the logs the fix is exponential backoff, not more attempts. No change wanted now.

_Cross-vendor agent review (Codex reviewer over the diff: no findings; Claude adjudication and the local verification above)._

PR #5229, opened 2026-08-01
fix(examples): tidy both example modules and build them in CI

maphew:
## Why

Both Go modules under `examples/` fail a plain `go build` on current `main`:

```
$ cd examples/library-usage && go build ./...
go: updates to go.mod needed; to update it:
	go mod tidy
```

Same for `examples/bd-example-extension-go`. Reproduced at `b5ba4cd42`.

They are separate modules that reach the parent through
`replace github.com/steveyegge/beads => ../..`, so their `go.mod`/`go.sum` record
the parent's entire dependency graph. Every root dependency change invalidates
them — and no CI job has ever compiled them (`grep -rl 'examples/' .github/workflows/`
returns nothing), so the drift accumulated silently.

`examples/` is the first code a new user copies. A broken example is a bad first
five minutes, not a cosmetic wart.

This was noticed while landing #4942 (the Go 1.26.5 bump): the tidy was
deliberately left out of that commit because of the go.sum churn, with a note to
follow up either by tidying in a dedicated PR or by adding a build job. This
does both, because either alone leaves a hole — tidying without CI just resets
the clock, and CI without tidying lands red.

## What

**1. `go mod tidy` in both example modules.** The extension example carries most
of the churn (~1.6k lines of `go.sum`), which is inherent to recording the parent
graph through the `replace` directive. Both modules now build clean with the
project's canonical `-tags gms_pure_go`.

**2. `scripts/build-examples.sh` + a `build-examples` job in `pr.yml`**, so the
drift cannot recur unnoticed. It runs `go vet ./...` rather than `go build`:
with `-o <dir>/`, `go build ./...` compiles only the main packages and silently
skips libraries, and without `-o` it litters the tree with binaries. `go vet`
type-checks every package including test files, writes nothing, and still fails
on the stale-go.mod condition. The script:

- discovers modules from `git ls-files 'examples/*/go.mod'` — no hardcoded list,
  so a new example is covered the moment it lands;
- sources `.buildflags` for the canonical `CGO_ENABLED`/`-tags=gms_pure_go`
  (this is also what keeps `scripts/check-build-tags.sh` green);
- is bash-3.2 and BSD-userland clean, so it actually runs on macOS;
- reports **every** failing module rather than stopping at the first;
- prints the exact `go mod tidy` invocation to run.

## One thing for you to decide

The job is advisory — implemented at the **step** level — and **not** in
`ci-gate`'s required list.

Because of the `replace` directive, any root `go.mod` change not mirrored into
the examples fails this job, so making it blocking imposes a "tidy the examples
too" step on every dependency bump. That is a maintainer call about contributor
friction, and it did not seem like mine to make inside the PR that adds the
check.

Two earlier revisions of this section each got the mechanics wrong, and the
2026-08-10 review corrected the second one. Revision one left the job out of
`ci-gate` and claimed that made it non-blocking — but `pr-preflight.sh` and
the merge patrol gate on the raw `statusCheckRollup`, not `ci-gate`
membership. Revision two used **job-level** `continue-on-error`, which still
reports the check *run* as FAILURE in the PR rollup (only the workflow-run
conclusion flips), so it was de facto blocking while advertising itself
advisory. The current shape follows the repo's own cygwin-leg precedent:
`continue-on-error` and a `timeout-minutes` on **each step**, with
outcome-guarded `::warning` annotations, so any failure or hang anywhere in
the job degrades to a warning and the check run itself always concludes
green. There is deliberately no job-level timeout — a job-level timeout
CANCELS the job, and a cancelled check run is red again by a different door.

To promote it to a real gate: drop the build step's `continue-on-error` and
the warning steps, then add `build-examples` to `ci-gate`'s `needs`,
`CI_GATE_REQUIRED`, and its env map. Say the word and I will do it here.

## Verification

- `cd examples/bd-example-extension-go && go build -tags gms_pure_go ./...` — clean.
- `cd examples/library-usage && go build -tags gms_pure_go ./...` — clean.
- `./scripts/build-examples.sh` — `2 example module(s) type-check cleanly`.
- Negative test: reverting `examples/library-usage/go.mod` to its pre-tidy state
  makes the script exit 1, name that module, and print its `go mod tidy` command;
  the other module still reports `ok`.
- `./scripts/check-build-tags.sh` — `96 file(s) scanned, all clear.`
- `shellcheck -S warning scripts/build-examples.sh` — clean.

Local `go` is 1.26.5, matching `go.mod`.

---

_claude-opus-5-high on behalf of matt wilkie_


maphew:
Pushed `4676cb3fa` after a local dual-vendor review (Claude + GPT-5.6). Two of the four findings were correctness bugs in my own additions, and one of them contradicts a claim in the original PR description — I have edited the description accordingly, but flagging it here so the change is not silent.

**1. `go build -o <dir>/ ./...` was a false green.** With `-o` naming a directory, Go compiles only the *main* packages and silently skips every library package. Reproduced in an isolated module with a good main package plus a library package containing a type error:

| command | exit |
|---|---|
| `go build -o "$out/" ./...` | **0** |
| `go build ./...` | 1 |
| `go vet ./...` | 1 |

The mirror-image bug: a library-only example module fails `go build` with `no main packages to build` — a false red. Switched to `go vet ./...`, which type-checks every package *including test files*. That last part matters here: `examples/library-usage/main_test.go` exercises a lot of live API, and `go mod tidy` counts its imports when generating the go.mod this PR commits, so a check that never compiles it was leaving part of the recorded dependency graph unverified. It also drops the scratch-directory machinery entirely.

**2. "Not in `ci-gate`" is not the same as "non-blocking".** The original description said the omission kept the job from imposing friction and that it would be "visible on the checks list either way". That is true for GitHub branch protection and false for our own automation: `pr-preflight.sh` blocks on any FAILURE in the raw `statusCheckRollup`, and the `pr-babysit` merge patrol requires every rollup entry to be SUCCESS/NEUTRAL/SKIPPED. Neither consults `ci-gate` membership. So the state I originally proposed was the one posture that stalls the merge lane repo-wide after any un-mirrored dependency bump, while advertising itself as optional.

The job is now `continue-on-error: true`, which is genuinely advisory. **Whether it should instead be a real gate is still your call** — the workflow comment records how to promote it (drop `continue-on-error`, then add `build-examples` to `ci-gate`'s `needs`, `CI_GATE_REQUIRED`, and its env map). I would rather hand you that decision than smuggle it in.

**3. Portability.** `mapfile` does not exist in bash 3.2 (stock macOS) and `xargs -r` is GNU-only, so the script could not run locally on macOS at all. Now a NUL-delimited read loop, which also fixes module paths containing whitespace. `source ./.buildflags` is guarded (unguarded under `set -uo pipefail` it would continue with `GOFLAGS` unset and type-check the ICU path while `check-build-tags.sh` still passed, since that only greps for the literal string). Discovering zero modules is now an error rather than a green.

**One thing worth knowing that this PR does not fix.** `examples/bd-example-extension-go/main.go` calls `sql.Open("sqlite3", ...)`, but no sqlite3 driver is imported in that module and none is registered anywhere in beads. This tidy *removes* `modernc.org/sqlite` from the graph, confirming it — so a user copying that example gets `sql: unknown driver "sqlite3"`, and `*dbPath` is now a Dolt directory rather than a SQLite file anyway. Pre-existing and out of scope, but the new job will certify that example "buildable" forever, so green does not mean working there. Happy to follow up separately if you want it fixed or the example retired.

Verification after the rework: both modules vet clean; reverting an example `go.mod` to its pre-tidy state still exits 1 with the exact `go mod tidy` command; a broken library package now fails where it previously passed; `scripts/check-build-tags.sh` clean (97 files); `shellcheck` clean.

---

maphew:
Reviewed at `fa79c0c2c99`.

No correctness findings in the submitted change. Module discovery is NUL-safe and Bash-3.2-compatible, canonical build flags are sourced, all modules are checked despite individual failures, and `continue-on-error` matches the explicitly advisory CI policy. The lockfile churn is the expected result of tidying the two example modules.

The known nonfunctional SQLite extension example is pre-existing and should be repaired or retired separately; it does not invalidate this buildability guard.

Verification:

- `./scripts/build-examples.sh` — attempted locally, but did not complete before the 124-second review timeout; no local verdict.
- GitHub's `Build example modules` job and required checks — PASS at inspection time.

Disposition: **merge**.

maphew:
Updated: merged current `main` (7e825441b) — zero conflicts — and fixed the three red checks, which all traced to one root cause: `TestGoCacheOwnershipTopology` landed on main after this PR's last update and requires every `actions/setup-go` step in `pr.yml` to set `cache: false`. The new `build-examples` job now does. While at it: re-ran `go mod tidy` in both example modules (they had drifted against merged main — exactly the drift class this PR's CI job exists to catch, so the job is already earning its keep), and bounded the job with `timeout-minutes: 10` so its advisory `continue-on-error` posture can't be defeated by a hang holding the check pending.

`./scripts` tests (including the cache-topology test), the main-module build, and both example-module builds pass locally. Review appreciated — this one is small.

steveyegge (changes requested):

Fable adversarial review (bee, beads-lane steward). The substance is real and verified in an isolated worktree on main (4e7cc88): both example modules genuinely fail plain `go vet`/`go build` today ("updates to go.mod needed" — reproduced), your tidy is byte-identical to what `go mod tidy` produces on today's main (re-ran it over the applied diff; empty `git diff` — the re-merge fully refreshed the 8/3 vintage), the script goes green with exit 0, the negative test behaves exactly as the body claims, shellcheck clean, no workflow on main touches `examples/` (no duplication), and the `replace => ../..` design both modules retain is right. The +2046 is ~90% go.sum churn inherent to that design. One inverted conclusion blocks, one efficiency fix should ride along:

**1. blocking — job-level `continue-on-error: true` does not make this job advisory in THIS repo; as written it's de facto blocking while advertising itself as advisory.** GitHub reports a failed continue-on-error *job*'s check run as FAILURE in the PR rollup — only the workflow-run conclusion turns success ([community discussion #15452](https://github.com/community/community/discussions/15452); [Ken Muse](https://www.kenmuse.com/blog/how-to-handle-step-and-job-errors-in-github-actions/)) — and `scripts/pr-preflight.sh:371-385` blocks on exactly that per-check-run conclusion, never consulting the workflow-run conclusion. So on the first un-mirrored root dep bump, this job goes red on every PR and stalls the merge lane repo-wide — the precise failure mode your body says it disproved. Your §"One thing for you to decide" has the analysis right (ci-gate membership is irrelevant) but the conclusion inverted. Fix, matching the repo's own precedent (the cygwin leg, pr.yml:963): move `continue-on-error: true` to the *step* running the script, give it an `id:`, add an outcome-guarded `::warning` step — failures then surface as annotations, not a FAILURE check run. Or embrace what the YAML accidentally is and make it honestly blocking (drop continue-on-error, join ci-gate) — the script's printed `go mod tidy` fix keeps contributor friction small. Either is ~6 lines; the current middle position is the only wrong one.

**2. should-fix — no module-cache restore: every PR cold-downloads the entire dolt-sized parent graph.** Your job sets `cache: false` with no `actions/cache/restore` for `~/go/pkg/mod`, unlike siblings `build-artifacts` and `test-macos`. The examples' graph is essentially the parent's, so the existing `beads-go-mod-v2-…` key is a near-perfect hit for free. Cold-download on every PR (job runs unconditionally, no path filter) burns minutes and pushes against your own 10-minute timeout — and a timeout failure under finding 1's current YAML poisons the rollup. Add the restore; optionally a paths filter (`go.mod`, `go.sum`, `examples/**`, the script, pr.yml) since drift can only enter through those.

**Nits:** the extension example remains functionally dead at runtime — `sql.Open("sqlite3", …)` with no driver imported (the tidy correctly dropped the sqlite indirects because nothing imports one) plus an ignored `beads.Open` error; pre-existing rot, not a regression, but your own "bad first five minutes" rationale applies — worth a follow-up issue to port it to the dolt-era API rather than letting the green check imply it works. And `go vet` widens the gate beyond staleness (lint findings fail the drift check) — currently clean, honestly documented, noting so a future analyzer finding isn't misread as dep drift.

**Cross-checks clean:** all five Verification bullets reproduced; nested-example discovery works (git pathspec `*` crosses `/`); the `.buildflags` source-guard reasoning holds; every root-package symbol the examples use survived the facade refactors; the pr.yml hunk anchors cleanly; script is bash-3.2/BSD/shellcheck clean as claimed.

With 1 fixed (and 2 riding along), this merges on green — ping me on the re-push.

maphew:
Both findings addressed at `6fa19bb1f`, and thank you — finding 1's citation trail (check-run vs workflow-run conclusion) settled a question my earlier revision had answered with the wrong half of the documentation.

**1 (blocking) — fixed, taking the advisory route.** `continue-on-error` moved to the step level per the cygwin-leg precedent, with an `id:` and an outcome-guarded `::warning` step. I kept the job advisory rather than embracing blocking because the "should every dependency bump also tidy the examples" friction call still isn't mine to make from inside the PR that adds the check; the body's "One thing for you to decide" section is rewritten accordingly (it now records both wrong earlier postures and why they were wrong, so the next reader doesn't re-derive them).

A second-vendor pass over the fix caught a residue of the same failure mode: the **job-level `timeout-minutes` I originally kept would CANCEL the job on a hang, and a cancelled check run is red again by a different door.** So the current shape has no job-level timeout at all — every step (checkout, setup-go, cache restore, build) carries its own `continue-on-error` + `timeout-minutes`, with a second warning annotation for the could-not-run-at-all case. The job now cannot conclude red: any failure or hang anywhere degrades to a warning annotation on a green check run.

**2 (should-fix) — cache restore added,** reusing the repo-wide `beads-go-mod-v2-…` key verbatim (near-perfect hit, as you said, since the examples' graph is the parent's through the `replace` directive). I deliberately skipped the optional paths filter for now: with the cache warm the job is ~a minute, and a filter is a second silent-coverage-cap surface to reason about. Easy to add if you want it.

**Nit (extension example dead at runtime):** agreed, and it was already on file before this review — #5574 tracks exactly that (`sql.Open("sqlite3", …)` with no driver imported, pointed at a Dolt directory). This PR keeps its scope at "the modules type-check"; the port to the dolt-era API belongs to that issue.

maphew:
Follow-up: the fix head grew one commit, `3d9e177fb`. The cache-restore step added to `build-examples` tripped `TestGoCacheOwnershipTopology`'s per-job cache-step registry (it wasn't registered, and carried an unused step `id` the registry compares). The commit registers the job in the inventory/managed/ordering/setup-go-id lists and drops the id; the workflow behavior itself is unchanged. All 101 checks are green on this head.

steveyegge:
## Review verdict: MERGE

Re-review after the 6fa19bb1f/3d9e177fb fix round (bee, beads-lane steward). Both findings from the 2026-08-10 review are addressed, and verified live rather than taken on faith:

**Finding 1 (advisory posture)** — fixed exactly per the cygwin-leg precedent: `continue-on-error` at the step level with an outcome-guarded `::warning` step. And you caught the residue I missed: the job-level `timeout-minutes` I'd have let ride would CANCEL on a hang, and a cancelled check run is red by a different door. The current shape (per-step timeout + continue-on-error everywhere, second warning for the could-not-run case) is the right fixpoint. I checked the claim against the live run on 3d9e177fb: the check run concluded SUCCESS with the vet step genuinely green and both warning steps skipped — so today's green is real, not masked.

**Finding 2 (cache restore)** — added with the repo-wide `beads-go-mod-v2-…` key byte-identical to main's five existing uses; job is ~2.5 min warm. Declining the paths filter is fine at that cost, for the reason you gave. The `TestGoCacheOwnershipTopology` follow-up in 3d9e177fb registers the job in all four registries correctly (I diffed it against the test's expectations on main: no `id:` on the restore step, `cache: false`, `id: setup-go`, ordering assert — all consistent).

**Cross-checks:** rollup on 3d9e177fb is 99 SUCCESS + 2 SKIPPED, nothing red or pending, so pr-preflight and the merge patrol are both satisfied; main hasn't touched go.mod/go.sum since your last tidy (HEAD 4ad99760b is docs-only), so the examples land tidy; #5574 covers the dead-at-runtime extension example, agreed that's its scope, not this PR's.

**Two nits, neither blocking, no action needed:** (a) a step killed by step-level `timeout-minutes` may report outcome `cancelled` rather than `failure` on some runner versions, in which case the staleness warning wouldn't annotate on a *hang* — the check run stays green either way, so this is annotation fidelity only; widening the guard to `outcome != 'success' && outcome != 'skipped'` would close it if it ever matters. (b) The advisory design means a green check no longer proves the examples are tidy — drift is visible only in annotations. That's the documented trade, and the promote-to-gate recipe in the workflow comment is accurate (I checked it against ci-gate's needs/CI_GATE_REQUIRED/env map). Whether to promote after it has run quiet for a while remains the open maintainer call, same pattern as test-macos.

Merging as-is is correct; the advisory-vs-gate decision doesn't need to be resolved to land the fix.

PR #5231, opened 2026-08-01
refactor(test): replace state-label store fixtures with pure projections

julianknutsen:
## What

Extract the direct `state` command's existing label projections into two private helpers and exercise those production helpers with untagged table tests.

This replaces three Dolt-backed test fixtures while retaining the embedded command, concurrent update, and proxied-server tests as boundary proofs.

## Why

The removed tests created real stores but never called the production command logic. Their test helper independently reimplemented the same label parsing, adding setup cost without protecting the command from a parsing regression.

The new tests run at the lowest useful seam and cover the observable parsing contract directly.

## Behavior preserved

- Scalar lookup matches the exact dimension prefix and remains first-match-wins.
- Values containing additional colons remain intact.
- Empty values remain distinguishable from a missing label inside the projection.
- State-list projection ignores labels without a non-empty dimension and remains last-match-wins for duplicates.
- Human and JSON output, event creation, label writes, partial-ID resolution, and proxied behavior are unchanged.

## Efficiency evidence

- Direct Dolt fixtures in `state_test.go`: 3 → 0.
- Fixture operations removed: 9 issue creates, 20 label adds, 2 label removes, and 16 label reads.
- Focused test-body time on this host: approximately 0.91s → below Go's 0.01s reporting resolution.
- Focused CGO-disabled red/green loop after compilation: 4.395s.
- Net code change: 126 insertions, 278 deletions.

## Verification

- Focused TDD proof: failed first with undefined projection helpers, then passed.
- `CGO_ENABLED=0 go test -count=1 -tags gms_pure_go ./cmd/bd`
- `make test` — passed in 503.29s; `cmd/bd` passed in 454.447s.
- `golangci-lint run ./...` — 0 issues.
- Two independent Sol reviews reported no blocker or major findings.

Bead: `bd-e5ch3`

### DIFF
--- a/cmd/bd/state.go
+++ b/cmd/bd/state.go
@@ -64,14 +64,7 @@ Examples:
 			return HandleErrorRespectJSON("%v", err)
 		}
 
-		prefix := dimension + ":"
-		var value string
-		for _, label := range labels {
-			if strings.HasPrefix(label, prefix) {
-				value = strings.TrimPrefix(label, prefix)
-				break
-			}
-		}
+		value, _ := findStateLabel(labels, dimension)
 
 		if jsonOutput {
 			result := map[string]interface{}{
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Delegated Sol review council

Two independent Sol reviewers inspected the final diff against `origin/main` and the frozen `bd-e5ch3` acceptance criteria. Both reported **no blocker or major findings**.

The council independently verified:

- exact-prefix and first-match scalar behavior;
- last-match collection behavior;
- empty and colon-bearing values;
- equivalent old-label reconstruction in `set-state`;
- unchanged embedded, concurrent, and proxied boundary coverage; and
- removal of three redundant direct Dolt fixtures without loss of a distinct boundary proof.

Local `make test` and `golangci-lint run ./...` passed. All PR checks are terminal green or intentionally skipped.

PR #5232, opened 2026-08-01
fix(schema,ci): prepared DML — not just prepared ALTER TABLE — silently no-ops on the Dolt CLI path

maphew:
## Why

`internal/storage/schema/cli_migrations.go` documents the Dolt CLI batch-path limitation as being about "some prepared ALTER TABLE statements". That understates it.

Per dolthub/dolt#11345 (filed with a repro, verified on dolt 2.2.0 and 2.2.2), `dolt sql -q`/`-f` silently no-ops `EXECUTE` of **any** prepared DML — an `UPDATE`/`INSERT`/`DELETE` built into a `SET @sql = '...'` string — while reporting success. A prepared `SELECT`, or a direct statement, on the same path runs correctly. Prepared `ALTER TABLE` is one instance of that behaviour, not the scope of it.

That path is `AllMigrationsSQL()`, which builds the fresh-schema bundle. So the comment is the only thing standing between a future migration author and a write that reports success and changes nothing.

Two changes:

1. **Broaden the comment** to describe the real scope, cite dolthub/dolt#11345, and point at the pattern to use instead — migration 0059 (#4877), where real-table mutations are direct SQL and the only `PREPARE`d statements are guarded copies into throwaway stand-in tables that a direct statement then reads, so a silent no-op degrades gracefully rather than corrupting state.

2. **Add Check E to `scripts/check-migration-hygiene.sh`** so a new main-plane migration cannot quietly depend on a prepared write.

## Scope, and why existing migrations do not fail

Check E is new-files-only and `.up.sql`-only, both deliberately.

Seven shipped main-plane migrations (0035, 0037, 0041, 0047, 0053, 0055, 0058) already `PREPARE` writes to real tables. That is **not** a latent bug and they are not retrofits waiting to happen: a shipped migration is frozen (Check C enforces that), and on the one path this check concerns — the fresh-schema bundle, i.e. an empty database — a data backfill is a no-op whether or not the prepared write lands. `cliCompatibleMigrationSQL` overrides only the subset whose prepared statements change the committed schema *shape*, which is what the bundle's contract actually promises. The point of the check is that the idiom stops spreading into migrations where the write does matter.

Only `migrations/*.up.sql` is embedded into the bundle (`//go:embed`, `schema.go:218`), so `.down.sql` is out of scope and is not scanned.

Prepared `ALTER TABLE` keeps its accepted-idiom status and is not flagged; only the leading keyword inside the quoted text decides. Detection buffers a statement across lines, because the common form here is a multi-line `SET @sql = IF(cond, '...', '...')` whose `PREPARE` arrives several lines later — a single grep cannot correlate the two.

## Cross-vendor review

Reviewed with a second model family (Codex, `gpt-5.6-sol`) over eight rounds. Every finding was confirmed against source and reproduced before fixing, and each round found something the last did not:

1. The check **rejected the very pattern its own failure message recommends** — 0059's `INSERT INTO __bd_0059_*` stand-in writes would have been flagged. Fixed by recognising the tree's stand-in naming convention (a `__` prefix, which no real beads table uses).
2. It scanned `.down.sql` files, which can never meet this bug.
3. The resulting exemption checked only the *first* DML target, so `SET @sql = IF(cond, '<stand-in write>', '<real write>')` was waved through on the strength of the branch that was not the problem.
4. Even checking *every* target was not enough: `UPDATE __stage s JOIN issues i ON i.id = s.id SET i.priority = 0` writes to `issues`, and no amount of pattern-matching finds that reliably. Separately, `ON DUPLICATE KEY UPDATE v = 2` was read as a second write verb, rejecting a safe stand-in upsert.

Round 4 is the one that changed the design rather than the regex. The exemption is now restricted to the single form whose target is unambiguous — `INSERT INTO`, where the target is always the token after `INTO` regardless of what the `SELECT` source does. **A prepared `UPDATE` or `DELETE` is never exempted**, because their multi-table forms can move the modified table behind a `JOIN`.

The cost is that a future conditional real-table-free `UPDATE` gets flagged and has to be restructured the way 0059 does, or justified in review. That seems the right side to err on for a check whose entire purpose is refusing to guess about writes.

Round 5 found two more: `SET @sql := '...'` (the `:=` assignment form) was never buffered at all, and the DML test required the verb immediately after the opening quote — so `'/* conditional */ UPDATE issues ...'` and `'WITH t AS (...) UPDATE issues ...'` both sailed through. Same mistake as round 4 in a different place, so the verb is now matched as a whole token anywhere, with `ON UPDATE` / `ON DUPLICATE KEY UPDATE` masked first so a column attribute is not read as a write.

Rounds 3, 4 and 5 are all false negatives *introduced by* the fix for round 1 — worth stating plainly, because a single review pass would have shipped a check that silently exempted real-table writes while looking like it worked.

Round 6 found two more, and these were **not** in the defer-able class, so they are fixed:

- `PREPARE stmt FROM 'UPDATE issues ...'` — the literal form, no user variable in play — was not checked at all. That is the most direct spelling of the hazard, and arguably the first thing an author writes.
- The stand-in exemption counted classified targets but not `INSERT` verbs. MySQL and Dolt accept `INSERT issues ...` with `INTO` omitted, so `IF(c, 'INSERT INTO __stage ...', 'INSERT issues ...')` produced one stand-in target, looked unanimous, and exempted the real-table write in the other branch. The exemption now requires every `INSERT` verb to have produced a target, so anything it cannot classify blocks the exemption instead of going missing from the tally.

The second matters more than its severity label suggests: that is the exemption granting safety it should not, which is worse than the detector missing a shape. An exemption must fail closed.

Round 7 found two more of the same ordinary kind, also fixed: `REPLACE INTO` was not in the write-verb set at all (so a prepared `REPLACE INTO issues ...` passed entirely), and the literal-`PREPARE` branch matched only single-quoted strings. `REPLACE` is now detected and, like `UPDATE`/`DELETE`, can never be exempted — this exemption classifies `INSERT INTO` targets only, so any other write verb has to block it rather than go untallied.

### Where I stopped, and why

Round 5 surfaced two further bypasses — a multi-variable `SET @guard = 1, @sql = '...'` (only the first variable is tracked) and a `PREPARE` split across lines. **I did not fix those, and the script's header now says so explicitly** rather than implying coverage it lacks:

- Neither form occurs anywhere in the current migration tree (verified — the only `PREPARE`-without-`FROM` hits are comments). Note these are the *variable-correlation* gaps only; `PREPARE ... FROM '<literal>'` is covered, and the `KNOWN LIMITS` block says so, so the list is not read as broader than it is.
- Check E is advisory hygiene on *new* files, which a human reviews anyway.
- The convergent fix is statement-level scanning rather than a line-oriented state machine, which subsumes all three known bypasses at once. Attempting that rewrite at round 5, with no remaining review budget to validate it, would have shipped an unvalidated parser — worse than a documented heuristic.

So the header carries a `KNOWN LIMITS` block naming all three shapes and telling a reviewer not to trust a green check when they see one. A check that overstates its coverage is more dangerous than one that admits where it stops. Happy to do the statement-level rewrite as a follow-up if you'd rather have it before this lands.

## Validation

- `scripts/check-migration-hygiene.sh` passes clean on the tree — zero false positives on existing migrations.
- Proved the check fires and does not over-fire, with scratch migrations covering: prepared `ALTER TABLE` (not flagged), prepared real-table `UPDATE` (flagged), multi-line prepared `INSERT` (flagged), the 0059 stand-in shape (not flagged), a stand-in upsert with `ON DUPLICATE KEY UPDATE` (not flagged), mixed `IF` with one real branch (flagged), joined `UPDATE` writing to a real table (flagged), joined `DELETE` from a real table (flagged), `INSERT` into a real table from a stand-in (flagged), and `.down.sql` carrying prepared DML (not flagged).
- `go build` and `go vet` clean on `./internal/storage/schema/` (`CGO_ENABLED=1 -tags gms_pure_go`).

If dolthub fixes #11345 this guard becomes belt-and-suspenders rather than load-bearing, which seems like the right trade either way.


maphew:
Two more findings from the cross-vendor pass after opening. Reconciling both here rather than silently pushing:

**Fixed (5749900f).** The exemption's `UPDATE`/`DELETE`/`REPLACE` rejection required *whitespace* after the verb, while the detector used a token boundary. Dolt takes a comment as a token separator, so

```sql
IF(c, 'INSERT INTO __stage ...', 'UPDATE/*x*/ issues SET priority = 3')
```

was seen as DML by the detector and then exempted anyway — the whitespace-anchored test never noticed the `UPDATE`, the `INSERT` verb and target counts agreed, and the real-table write passed. Reproduced before fixing. Two predicates that have to agree on what a verb is should not spell it two different ways; that divergence was the whole bug.

**Not fixed, deliberately.** The reviewer also flagged `declare -A` as unavailable on stock macOS Bash 3.2. That is true, but it is pre-existing practice in this tree rather than something this PR introduces — `.github/scripts/proxied-test-shard.sh`, `embedded-storage-test-shard.sh` and `embedded-test-shard.sh` all use it — and the `check-migration-hygiene` job runs on `ubuntu-latest`, so CI is unaffected. The impact is limited to a macOS contributor running the script locally, who would get `declare: -A: invalid option` instead of a verdict.

Happy to convert it to a Bash 3-compatible mapping if you would rather this script not carry the same requirement as the shard scripts — say the word. It seemed wrong to unilaterally diverge from the tree's existing convention inside an unrelated PR.

_claude-opus-5-high on behalf of matt wilkie_

maphew:
Reviewed at `e4b163e90b3`.

One should-fix false positive remains in migration-hygiene Check E. `neutralize_sql_clauses` masks only the exact single-space spellings `on update` and `on duplicate key update`, but SQL permits repeated whitespace, tabs, and line breaks. A valid prepared DDL clause such as `ON  UPDATE CURRENT_TIMESTAMP` therefore leaves `UPDATE` visible and is rejected as prepared DML, despite the check explicitly exempting prepared DDL.

Please make those masks whitespace-aware and add executable regression fixtures for repeated and multiline whitespace. The detector is otherwise carefully scoped to main-plane migrations and stand-in-table writes.

Verification:

- `BASE_SHA=upstream/main ./scripts/check-migration-hygiene.sh` — PASS (`Migration hygiene OK.`).
- `bash -n scripts/check-migration-hygiene.sh` — PASS.
- `shellcheck` — unavailable locally.
- Blocking upstream preflight — blocked while 31 checks were still pending; no merge or branch action was taken.

Disposition: **fix-merge**.

maphew:
Current-head follow-up at `48d61725`.

The only change since my review of `e4b163e9` is a clean merge of current `main` (`c3e600c9`, `b73c9873`); it does not alter this PR's migration-hygiene implementation. I rechecked the first-parent delta and reran the focused checks. The prior whitespace-normalization finding therefore remains the single should-fix item, without repeating it here.

Verification at this head:

- `BASE_SHA=upstream/main scripts/check-migration-hygiene.sh` — PASS.
- `bash -n scripts/check-migration-hygiene.sh` — PASS.
- `go test ./scripts -run TestPRPreflight -count=1` — package PASS (`[no tests to run]` on Windows).

Disposition remains **fix-merge**. The overlap scan found no competing prepared-DML hygiene PR.

PR #5238, opened 2026-08-01
test: make ready ordering setup deterministic

julianknutsen:
## What

Make `TestProtocol_ReadyDefaultOrdering` import all six fixtures in one
timestamped JSONL batch instead of running four `bd create` subprocesses and
waiting through two unconditional 1.1-second sleeps.

The test still initializes a real workspace and runs `bd ready --json` twice.
No production behavior changes.

## Why

The old setup used wall-clock delay to manufacture FIFO ordering. Explicit
RFC3339 timestamps are deterministic, faster, and make the intended priority,
FIFO, and ID-tiebreak cases visible in the fixture itself.

The revised test also proves the exact six-ID fixture set before performing
timestamp or positional lookups. Two independent Sol reviewers identified
that guard as necessary to prevent missing imported IDs from producing
zero-value, false-green comparisons.

Tracked by `bd-y9cjq`.

## Verification

- Deliberate RED: inverted P1 timestamps failed the new setup guard in 16.74s.
- Focused Dolt-enabled test body: 20.99s before, 17.51s after
  (3.48s / 16.6% faster).
- Same focused package: 28.168s before, 27.643s after.
- Same focused outer wall, including container lifecycle: 41.77s before,
  41.46s after.
- Dolt-enabled `./cmd/bd/protocol/...`: 225.297s package, 239.04s wall.
- `make test`: passed in 512.21s with 39.0% aggregate coverage.
- `golangci-lint run ./...`: zero issues.
- Two independent Sol re-reviews: no remaining blockers or majors.

### DIFF
--- a/cmd/bd/protocol/ready_front_test.go
+++ b/cmd/bd/protocol/ready_front_test.go
@@ -105,31 +105,30 @@ func TestProtocol_ReadyDefaultOrdering(t *testing.T) {
 	t.Parallel()
 	w := newWorkspace(t)
 
-	// created_at has second granularity, so space the issues whose creation
-	// order the test asserts. Same-second pairs are handled by the id tiebreak,
-	// which the two explicit-id issues below exercise.
-	oldP2 := w.create("--title", "old p2", "-p", "2")
-	time.Sleep(1100 * time.Millisecond)
-	oldP1 := w.create("--title", "old p1", "-p", "1")
-	time.Sleep(1100 * time.Millisecond)
-	newP1 := w.create("--title", "new p1", "-p", "1")
-	p0 := w.create("--title", "urgent", "-p", "0")
-
-	// Two same-priority issues that share created_at exactly, so ONLY the id
-	// ASC tiebreak can order them. created_at has second granularity and a bd
-	// invocation takes longer than that, so the pair is imported (J6) rather
-	// than created — a CLI-created pair would straddle a second boundary and
-	// silently stop exercising the tiebreak.
-	prefix, _, _ := strings.Cut(oldP2, "-")
-	tieA, tieB := prefix+"-tieaaaa", prefix+"-tiebbbb"
+	// Import every issue with explicit timestamps, including the same-second
+	// pair that exercises the id ASC tiebreak. This avoids depending on wall
+	// clock granularity or the duration of individual bd create invocations.
+	oldP2 := w.prefix + "-oldp2"
+	oldP1 := w.prefix + "-oldp1"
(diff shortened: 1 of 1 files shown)


maphew:
Post-merge review (requested as a "did we miss anything major" pass, since this landed already): **sound merge, nothing major missed.** Test-only, no production code touched, and a genuine improvement over what it replaced.

### Verified against the merged code and helpers

- **`w.prefix` is real and correct** — set in `newWorkspace` from `testPrefix(t)` (helpers_test.go:213-215), and the field's doc comment says it exists precisely for hand-written JSONL fixtures. Minted IDs like `<prefix>-oldp2` are well-formed for the workspace's own Dolt database, so no cross-test pollution.
- **Tie-pair semantics preserved** — `tieA`/`tieB` share `00:04:00Z` and priority 3, and the fixture deliberately lists `tieB` *before* `tieA`, so the observed order can't be insertion order; only the id ASC tiebreak yields `tieA` first. Mirrors the old test's intent exactly.
- **The new guards close the real hole** — if `bd import` ever ignored explicit `created_at` and stamped now(), all six issues would share one second and the FIFO assertions would go vacuous. The `Fatalf` guard on `oldP1`/`newP1` ordering (ready_front_test.go:130) plus the existing tie-pair `Fatalf` catch that regression from both directions. The test verifies the property it depends on rather than trusting the importer — good design.
- **Timestamp layout does real work** — `p0` is the *newest* issue yet must head the front, and `oldP2` is the *oldest* yet must rank below both P1s, so priority-dominates-recency is actually exercised, not just asserted.
- **No dangling code** — `time` and `strings` imports are still used elsewhere in the file; `w.create` remains used by the other tests.

### Minor nits (none merge-blocking, no follow-up warranted)

- The fixture-membership guard `requireStringSetEqual` is `Errorf`-based, so a missing ID continues into `itemByID`, whose zero `readyItem` then trips the next `Fatalf` with a confusing "got 0001-01-01…" message. Not false-green — the membership error has already fired, exactly the failure mode the PR body says the Sol reviewers were guarding against — just a slightly noisy failure path.
- `old, new :=` shadows the `new` builtin. Legal, lint-clean, mildly unidiomatic.
- Positional assertions like `indexOf(order, oldP1) > indexOf(order, newP1)` would pass vacuously on a `-1` miss, but the `len == 6` fatal plus the membership guard make that unreachable in practice.

One thing the speedup framing undersells: the bigger win is removing flake surface. The old CLI-created tie pair could silently *stop exercising the tiebreak* if invocations straddled a second boundary — a hazard the old comment itself admitted, now structurally gone.

PR #5239, opened 2026-08-01
docs(httpapi): close the re-spell and value-narrowing gaps in the Profiles relation

julianknutsen:
Two composition gaps in the `## Profiles` conformance relation landed in #5237, found by running a
live profile against the four rules rather than reasoning about them abstractly. Both are
description-only; the regen is byte-identical.

## An omission is not a licence to re-spell

The rules composed into a hole. Rule 1 scopes non-conformance to a re-spelling *"published under a
core name"*. Rule 2 permits omitting optional properties. Rule 4 permits declared extensions whose
names this document does not define. So a profile could omit an optional core member and republish the
same value under an extension name — the exact re-spelling rule 1 exists to forbid, reached by three
individually legal steps.

That was not hypothetical. `issue_type` is optional (`Issue.required` is `[id, title, priority,
created_at, updated_at]`), so a profile omitting it and publishing the value elsewhere read as
conformant. **Tier 1's ruling that `issue_type` is the canonical name therefore bound nobody** — the
relation permitted precisely the outcome the ruling rejected.

New fifth rule, placed next to "Verbatim or absent" because that is the flank it closes: a profile that
omits a core member MUST NOT publish that member's value under any other name. The test is on the
*value's provenance*, not the spelling, so it fires the same for a freshly minted extension name as for
one this document happens to define elsewhere.

## A narrowed value schema is a declared deviation

Rule 2 contradicted itself on the case the relation most needs to answer. Its enumerated tightenings
are `maxLength`/`maxItems`/`maxProperties` and closing an open vocabulary, and it excludes "re-typing a
member" — but its proviso tests *instances*: "every instance it emits is still valid against this
document". Narrowing a free-form object to short strings is schema-level re-typing (forbidden) and
instance-level valid (permitted). Both readings were available from the text.

Ruled as a third category rather than forcing it into either. Narrowing a value schema is **legal**,
because it is how a profile caps response size without transforming anything — but it is a **declared
deviation**, not a bounds-only tightening, conformant only where:

- values failing the narrower schema are **omitted entirely rather than coerced** — stringifying a
  typed value in place is a transformation, which "verbatim or absent" already forbids; and
- the declaration states the member **may be incomplete**, because a consumer cannot distinguish an
  entry dropped for failing the narrower schema from one that was never set.

Rule 2's enumeration is now explicitly exhaustive, with the proviso stated as a floor rather than a
grant: a change the list does not name is not made a tightening by the fact that its instances still
validate.

## Core and the relation stop talking past each other

`Issue.metadata` warns that values may be any JSON type and that clients MUST NOT decode it into a
string-to-string map — while the relation now blesses a profile that declares exactly that narrowing.
One sentence scopes the prohibition to readers of *this* document; a client of a profile may rely on
the profile's declaration. The core warning itself is unchanged: core clients still must not assume.

## Verification

`make api-check` clean, run twice. `internal/httpapi/apigen/types.gen.go` hashes
`dd0f41b2a8d3c6671fc2f851c9aca86f04bb757a` before regeneration, after regeneration, and after the
commit — description-only, proven rather than asserted. Diff is one file, 35 insertions, 8 deletions.
Vendor-neutrality grep over the added lines finds no product, deployment or consumer name; the prose
names only "a profile", "a consumer of the profile", "this document".

The motivating composition was re-run against the amended text and is now non-conformant at step 4,
with both escape clauses shut individually.

### DIFF
--- a/internal/httpapi/spec/openapi.v0.yaml
+++ b/internal/httpapi/spec/openapi.v0.yaml
@@ -120,7 +120,7 @@ info:
     its own surface. Conformance to this core is a checkable relation over
     wire-observable JSON — the profile's value model is subsumed by this one,
     plus the deviations the profile DECLARES — not a claim of shared
-    vocabulary. Four rules make it checkable.
+    vocabulary. Five rules make it checkable.
 
 
     * **Verbatim or absent.** A property emitted under a core name carries the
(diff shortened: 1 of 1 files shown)


maphew (approved):
Verified both gaps against current main rather than taking the description on faith, and both are real.

**The re-spell composition checks out.** Rule 1 scopes non-conformance to re-spellings "published under a core name", rule 2 permits omitting optional properties, and rule 4 permits declared extensions with non-core names — and `issue_type` is indeed absent from `Issue.required`, so the three-step path to republishing it under an extension name was legal under the landed text. The placement of the new rule immediately after "Verbatim or absent" is right: it is the same prohibition extended to the flank rule 1's "under a core name" scoping left open. The "Omission means the value is not on the wire" framing makes the provenance test clear.

**The narrowing ruling resolves a genuine contradiction, not a hypothetical.** `Issue.metadata` is `additionalProperties: true`, so any value narrowing trivially satisfies the old instance-validity proviso while reading as re-typing under the exclusion — both readings really were available. Making the enumeration exhaustive and the proviso a floor, then routing narrowing through the same declaration machinery as embedded-relation omissions, is the cleanest of the three options (forbid / silently allow / declare). The two conditions are the load-bearing part and both are well chosen: omit-not-coerce keeps "verbatim or absent" intact, and the may-be-incomplete disclosure names the exact thing a profile consumer cannot otherwise detect.

**The metadata cross-reference matches the implementation.** `types.Issue.Metadata` is `json.RawMessage` emitted verbatim through the `x-go-type` pin, so the core warning stays true for core clients while the one added sentence stops the schema from contradicting the relation it now sits beside. Nothing about handler behavior changes, and nothing needed to.

**No schema impact.** Description-only, no `required`/type/endpoint edits, and `apigen/` carries no embedded spec — the byte-identical regen claim is structurally expected, and the tri-OS doc-freshness checks confirm it. The verification section proving the regen hash before/after rather than asserting it is appreciated, as is re-running the motivating composition against the amended text and reporting where it now fails (step 4, both escape clauses individually).

One optional thought, no action needed: the operative sentence tests "that member's value under any other name" textually, while the provenance framing lives in the surrounding prose and the PR description. If this section grows a conformance-test harness someday, an explicit "carry" / provenance phrasing would make the coincidental-equality case unambiguous. Fine as is.

Approving. Nice catch finding these by running a live profile against the rules instead of reasoning abstractly — that method is exactly why the gaps surfaced.

PR #5241, opened 2026-08-01
fix(dbproxy): abort the proxy readiness wait when a concurrent stop advances the epoch

maphew:
## Why

`TestManagedLocalProxiedStopStartInterleave` can fail BOTH sides of a stop/start interleave and leave a half-torn-down proxy topology. Two prior diagnosis passes (bd `mybd-e1b3f`) established the mechanism precisely:

- The child proxy clears the spawn marker as its **first** act after taking `proxy.lock`, but publishes its pidfile only at the **end** of startup — after `net.Listen`, `startControl`, backend `Start`, and a readiness wait bounded by `serverReadyTimeout` = 30s. In that window a concurrent `bd dolt stop` finds no record at all ("pid 0") and then times out acquiring `proxy.lock` at `shutdownConfirmDeadline` = 5s — six times shorter than the lock-holder's wait.
- The child's existing stop-epoch re-check fires only **once, after** the readiness wait, so the abort path the code already has cannot release the lock inside stop's patience.

## What

The minimal fix from that diagnosis: `waitForServerReady` now consults an `interrupted` seam before each probe; `ListenAndServe` passes a closure that re-checks the stop epoch. A stop that lands mid-startup aborts the child within one dial-plus-backoff interval — worst case ~3s (in-flight dial 2s + backoff up to 1s), inside stop's 5s deadline — instead of after up to 30s. The existing post-wait fence is retained (it still covers the procid/RootID window); no timeouts changed; the spawn-marker clear is untouched.

Epoch-file read errors inside the loop are treated as no-change: a transient blip must not kill a healthy startup, and the retained post-wait fence still surfaces persistent read errors before publish.

Deliberately **not** in this PR (recorded as follow-up options on the bead if the failure ever recurs): moving the spawn-marker clear after pidfile publication (D-a) and reconciling the 5s/30s timeout mismatch (D-b). This change alone closes the observed CI failure pair with the smallest blast radius.

## Validation

- New unit tests: a direct seam test (never-ready backend, interrupt observed promptly) and a deterministic mid-wait abort scenario asserting `errStartInterrupted`, no pidfile published, backend stopped exactly once, and prompt return.
- **Negative verification**: with the in-loop check disabled, the mid-wait test fails at 29.5s vs 0.06s with it — the seam is load-bearing.
- Real interleave test run 3 iterations under `BEADS_TEST_PROXIED_LOCAL=1`: all pass; zero orphaned `dolt sql-server` processes left (verified by ps before/after).
- Full `./internal/storage/dbproxy/proxy/` package green; `go vet` clean; `make test` enqueued in the local verify queue at this head.
- Adversarial same-vendor review (verdict approve-with-nits) and cross-vendor review (codex `gpt-5.6-sol`, no actionable findings) both run on the final head; all review findings applied in the second commit.

Tracking: bd `mybd-e1b3f`.

### DIFF
--- a/internal/storage/dbproxy/proxy/server.go
+++ b/internal/storage/dbproxy/proxy/server.go
@@ -189,9 +189,33 @@ func (p *proxyServer) ListenAndServe(parentCtx context.Context) error {
 		return fmt.Errorf("start database server: %w", err)
 	}
 
-	if err := waitForServerReady(ctx, p.server, serverReadyTimeout); err != nil {
+	// Re-check the stop epoch on every readiness probe, not only after the
+	// wait. A concurrent `bd dolt stop` waits at most shutdownConfirmDeadline
+	// (5s) for proxy.lock — 6x shorter than serverReadyTimeout — and this
+	// process holds the lock with neither a spawn marker nor a pidfile
+	// published, so a stop that lands mid-startup would otherwise time out
+	// against a lock we keep holding for up to 30s. Checking inside the retry
+	// loop bounds the abort latency to about one backoff interval
+	// (readyMaxBackoff). A transient epoch-file read error must not kill an
+	// otherwise healthy startup, so it is treated as "no change" here; the
+	// post-wait fence below re-reads the epoch and surfaces a persistent read
+	// error before anything is published.
+	epochInterrupted := func() error {
+		changed, err := stopEpochChanged(p.rootDir, p.stopEpoch)
+		if err != nil || !changed {
+			return nil
+		}
+		return fmt.Errorf("%w for %s: stop epoch advanced during startup", errStartInterrupted, p.rootDir)
+	}
+	if err := waitForServerReady(ctx, p.server, serverReadyTimeout, epochInterrupted); err != nil {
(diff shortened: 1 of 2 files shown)


maphew:
Reviewed at `9dbdddfed01`.

No findings. The readiness loop now observes stop-epoch changes before each probe, makes interruption terminal rather than retryable, preserves ordinary readiness-error wrapping, shuts the backend down on interruption, and retains the final pre-publication epoch fence. The change remains contained in the deliberate dbproxy storage interface.

Verification:

- Focused epoch/readiness tests — PASS (`ok`, 1.909s).
- Full `internal/storage/dbproxy/proxy` package — PASS (`ok`, 8.411s).

Disposition: **merge**.

maphew:
Closing as superseded: #5354 (merged 2026-08-05) fixes the same stop/start interleave race this PR targeted, with a stop-epoch fast-abort covering the whole startup window (including the doomed-start lock starvation this branch addressed via the readiness wait). Re-verified against main at 7e825441b that the mechanism this PR adds is now covered. Thanks to the #5354 author for the more complete fix.

PR #4242, opened 2026-05-28
Fix role detection for bd -C target repos

osamu2001:
## Summary

- route role detection through the currently selected beads repo context instead of the process caller cwd
- apply that role lookup to create auto-routing and read auto-routing
- add regression coverage for caller cwd and selected BEADS_DIR target mismatch, including no `beads.role not configured` warning

Fixes gastownhall/beads#4241.

PR #5243, opened 2026-08-01
fix(routing): bind role detection to the bd -C target repo (re-port of #4242)

maphew:
## Why

`bd -C <target>` detects `beads.role` from the **caller's** repo, not the target's — confirmed by maintainer repro on 2026-06-17 (caller and target as separate repos, target with `beads.role=maintainer`, `bd -C target list --parent` warned from the caller's contributor role). Root cause: `-C` is not a chdir — it exports the target's `.beads` as `BEADS_DIR` and leaves CWD alone, while both role-detection call sites used `routing.DetectUserRole(".")`.

This is a maintainer re-port of #4242 by @osamu2001, whose fix approach review confirmed is correct and minimal: resolve the active repo through the `BEADS_DIR`-driven `beads.GetRepoContext()` and convert both `DetectUserRole` call sites (`create.go`, `routing_read.go`). The original PR is `CONFLICTING` against current main (the `determineAutoRoutedRepoPath` signature changed in #4866), so a maintainer touch was required either way; the port applies their delta byte-identically where it applies and re-applies the two mechanically-conflicting hunks. Both commits carry `Co-authored-by: Osamu Okano`.

Two deltas beyond the faithful port:

1. **Test repair** — the PR's sole CI failure: `TestActiveRepoPathForRoutingFallsBackToBeadsDirParent` built its fixture inside the beads repo, so `GetRepoContext()` succeeded and the fallback branch under test was unreachable. Rewritten with the fixture and CWD both outside any git repo (which is how the fallback is actually reached in production).
2. **Redirect guard** — cross-vendor review caught that the port would regress `.beads/redirect` users: with storage redirected outside any git repo, `RepoContext.RepoRoot` is the storage dir's parent, so role would be read from the storage location, silently bypassing contributor routing. The two external-BeadsDir shapes now get their opposite answers explicitly: an explicit `BEADS_DIR` selection (what `-C` plumbs) binds role detection to the selected project — the #4242 fix — while a followed redirect keeps it on the CWD workspace repo (a redirect relocates storage, not the project). Mutation-tested.

Known residual (deliberately not expanded here, filed as bd `mybd-o4u1w`): a `-C` target whose `.beads` is not in any git repo still falls back to the caller's repo root — low impact, since a non-git beads project has no git-config role to detect.

## Validation

- All routing/context-binding tests green, including the Dolt-backed `TestDetermineAutoRoutedRepoPath_*` (real Dolt server), the repaired fallback test, and the new redirect test.
- Mutation checks: the repaired fallback test fails on the pre-fix unreachable assertion; the redirect test fails with the guard removed.
- `go vet` clean; `make test` enqueued in the local verify queue at this head; cross-vendor review (codex `gpt-5.6-sol`) run, its P1 fixed in the head commit, log recorded.

Supersedes #4242 with attribution (author silent since 2026-07-06; branch conflicting). Fixes the issue reproduced in that thread. Tracking: bd `mybd-w7yc`.


maphew:
Reviewed at `7771c99c3cc`.

One should-fix redirect regression remains. `activeRepoPathForRouting` treats any non-empty `BEADS_DIR` as explicit selection, but normal command startup itself calls `prepareSelectedCommandContext`, which sets `BEADS_DIR` to the resolved target before `create` or routed reads call role detection. A regular `.beads/redirect` therefore reaches this helper with `BEADS_DIR` populated and is classified as explicit, selecting the storage root instead of `CWDRepoRoot`. The added direct helper test leaves `BEADS_DIR` empty and bypasses the real startup rebind.

Please preserve explicit-selection provenance from before the unconditional bind (`-C`, caller-provided `BEADS_DIR`, and related selectors) and add a command-level regression that passes through `PersistentPreRunE` with a real redirect.

Verification:

- Focused `cmd/bd` test command — timed out after 124.1s without a pass/fail result.

Disposition: **fix-merge**. The original contributor value is sound; this startup-path gap should be corrected on this branch before merge.

maphew:
The should-fix redirect regression from the review above is now addressed, at `3adb95aeb` + `764bdbe4c`.

**The fix.** The helper was reading the live `BEADS_DIR` env var to decide "explicit selection", but normal command startup (`prepareSelectedCommandContext`) sets `BEADS_DIR` to the resolved target for *every* command — redirects included — so a plain `.beads/redirect` was classified as explicit and role detection bound to the storage root. Selection provenance is now captured where user intent actually lives: `BEADS_DIR` present at process start, a `-C` flag, or a `.beads/.env` selector file (`loadBeadsSelectionEnvFile`). Internal rebinds no longer count.

**The requested regression test.** `TestActiveRepoPathForRoutingSurvivesStartupRebindAcrossRedirect` drives the real startup rebind (`prepareSelectedCommandContext`) with a real redirect and fails on the previous logic — verified red on the old head, green on this one. The earlier helper-only test kept `BEADS_DIR` empty and could not see this path, exactly as the review said.

**The `.env` case was caught cross-vendor.** The first commit alone regressed `.beads/.env`-routed `BEADS_DIR` (loaded after process start, so the snapshot missed it); a second-model review flagged it and `764bdbe4c` marks provenance in the selector loader, with `TestActiveRepoPathForRoutingHonorsEnvFileSelection` pinning it through `loadSelectionEnvironment`.

Two pre-existing tests that simulate a caller-exported `BEADS_DIR` mid-process now pin their provenance explicitly via a small test helper.

Local validation: focused routing/context/role tests green under `gms_pure_go`; `golangci-lint` clean. The wider `cmd/bd` slice is order-flaky in my nested-worktree checkout independent of this change (same failures on the unmodified head), so CI is the honest verdict on the full suite.

_claude-code-fable-5-high on behalf of maphew_

bee-ghosttrack (approved):
Reviewed adversarially at `03345df3a0`, in a throwaway worktree off `origin/main` (`a690b0a8c4`). I went in trying to break the role-binding claim and could not. Approving with nits.

## What I ran

```
go build ./...                                                   rc=0
go vet ./cmd/bd/...                                              rc=0
gofmt -l <the 6 changed files>                                   (empty)
git merge-tree --write-tree origin/main 03345df3a0               rc=0
git diff --check origin/main..03345df3a0                         rc=0
git merge-base origin/main 03345df3a0  == origin/main            (branch is current, no drift)
```

Tests, all `-count=1`:

```
CGO_ENABLED=0 go test -tags gms_pure_go ./cmd/bd \
  -run 'TestActiveRepoPathForRouting|TestDetectUserRoleForActiveRepo'
  PASS TestDetectUserRoleForActiveRepoUsesSelectedBeadsDir              (0.73s)
  PASS TestActiveRepoPathForRoutingFallsBackToBeadsDirParent            (0.03s)
  PASS TestActiveRepoPathForRoutingKeepsWorkspaceRepoAcrossRedirect     (0.30s)
  PASS TestActiveRepoPathForRoutingSurvivesStartupRebindAcrossRedirect  (0.32s)
  PASS TestActiveRepoPathForRoutingHonorsEnvFileSelection               (1.02s)
  PASS TestActiveRepoPathForRoutingFallsBackToCurrentDirectory          (0.19s)
  ok  cmd/bd  5.904s

CGO_ENABLED=1 go test ./cmd/bd -run 'TestDetermineAutoRoutedRepoPath_'   (real Dolt container)
  PASS _ContributorToPlanning / _UsesSelectedBeadsDirRole / _MaintainerToPlanning / _DefaultFallback
  ok  cmd/bd  323.345s
```

CI on the head is green across the board (only `[code]smith` autofix `skipping`), `mergeStateStatus: CLEAN`.

## Mutation probes — every added test is non-vacuous

Each mutation applied in the worktree, test re-run, then restored (`git status --porcelain` empty after each; verified).

| # | Mutation | Result |
|---|---|---|
| M1 | `activeRepoPathForRouting()` returns `"."` unconditionally (pre-fix behavior) | FAIL ×5 — incl. `UsesSelectedBeadsDir` re-emitting `warning: beads.role not configured` |
| M2 | drop the redirect guard: `if explicitBeadsSelection() \|\| !rc.IsRedirected` → `if true` | FAIL ×2 (`KeepsWorkspaceRepoAcrossRedirect`, `SurvivesStartupRebindAcrossRedirect`) |
| M3 | `explicitBeadsSelection()` reads live `os.Getenv("BEADS_DIR")` instead of startup provenance | FAIL (`SurvivesStartupRebindAcrossRedirect`) |
| M4 | drop the `.env` provenance marking in `loadBeadsSelectionEnvFile` (main.go:395-400) | FAIL (`HonorsEnvFileSelection`) |
| M5 | drop the `FindBeadsDir` → `filepath.Dir(beadsDir)` fallback (routing_role.go:34-36) | FAIL (`FallsBackToBeadsDirParent`) |
| M6 | final `return "."` → sentinel (routing_role.go:37) | FAIL (`FallsBackToCurrentDirectory`) |
| M7 | revert the `routing_read.go:52` call site to `routing.DetectUserRole(".")` | FAIL (`_UsesSelectedBeadsDirRole`: got `"."`, want the planning dir) |

M7 matters most: it proves the Dolt-backed test exercises the real `determineAutoRoutedRepoPath` seam and not just the helper. M2/M3/M4 prove the three deltas beyond the port each carry their own guard.

The repaired `TestActiveRepoPathForRoutingFallsBackToBeadsDirParent` is genuinely reachable now — under M5 it fails, so the fixture really does land on the `FindBeadsDir` branch rather than short-circuiting through `GetRepoContext()`.

## Correctness read

The seam is right. `-C` plumbs the target through `BEADS_DIR`, `beads.FindBeadsDir()` prefers `BEADS_DIR`, `isExternalBeadsDir` sees a different git common dir, and `repoRootForBeadsDir` resolves the target's toplevel — so `rc.RepoRoot` is the `-C` target repo. `routing.DetectUserRole` then runs `git config --get beads.role` with `cmd.Dir` at that root. That is the reported bug, fixed at the layer where it lives.

The `beadsDirProvidedAtStartup` package var (main.go:830) is a package-level initializer, so it is evaluated before every `init()` and before `main()` — no ordering hazard against `applyChangeDirSelection` / `prepareSelectedCommandContext`, both of which run inside `PersistentPreRunE`. `loadSelectionEnvironment()` (main.go:409) early-returns when `BEADS_DIR` is already set, so the `.env` provenance flip at main.go:395 can only fire when nothing else had claimed the variable — consistent.

Worktrees: I checked empirically that `git config --get beads.role` from a linked worktree returns the main repo's value, and `rev-parse --show-toplevel` returns the worktree root. So the behavior change for the ordinary non-selected case (`"."` → `rc.RepoRoot`, i.e. the main repo root) is role-equivalent. `FindBeadsDir`'s worktree fallback resolves the shared `.beads` and `isExternalBeadsDir` compares *common* dirs, so a worktree is correctly not treated as external.

No duplicated helper: after this PR `routing.DetectUserRole` has exactly one caller in the tree (`cmd/bd/routing_role.go:41`), and both former call sites route to it.

There is no `internal/gitenv` package in this repo and no `EntryKey` / `IsRoutingKeyForOS` symbol anywhere — nothing for the change to route through, and nothing it bypasses. The equivalent facility here is `internal/beads.GetRepoContext()`, which is what it uses.

Test hygiene: no new test sets `HOME`. The package's pre-existing `TestMain` (cmd/bd/test_repo_beads_guard_test.go:58) already repoints `HOME`/`XDG_CONFIG_HOME`/`APPDATA` at a temp dir for the whole suite, so the `git init` + `git config` fixtures cannot read or write the developer's real global config. `resetRepoContextCachesForTest` and `setBeadsDirStartupProvenanceForTest` both restore in `t.Cleanup`, and the redirect/`.env` tests snapshot-restore `serverMode`, the root flag state and `config` — no global leaks across tests that I could find.

## NITs (non-blocking)

1. **`create.go`'s converted call site has no test of its own.** `routing_read.go:52` is pinned by `TestDetermineAutoRoutedRepoPath_UsesSelectedBeadsDirRole`; `create.go:337` is only pinned transitively through the shared helper. Same one-liner, so I am not going to insist — but M7 has a twin that does not exist.

2. **An inherited `GIT_DIR` still defeats the binding.** `routing.DetectUserRole` sets `cmd.Dir` but does not scrub the git env (internal/routing/routing.go:14-20). I verified with plain git:

   ```
   cd target && git config --get beads.role                      -> maintainer
   cd target && GIT_DIR=caller/.git GIT_WORK_TREE=caller \
                git config --get beads.role                      -> contributor
   ```

   So in a git-hook shape (`bd` invoked from a hook with `GIT_DIR` exported) role detection reads the hook's repo regardless of the `-C` target. Pre-existing — `DetectUserRole(".")` had the identical hole — and this PR does not widen it. But the repo already owns the remedy: `scrubGitHookEnv` (cmd/bd/export_auto.go:929), written for exactly this class of bug. Worth a follow-up bead rather than scope creep here.

3. **Two role oracles now disagree under `-C`.** `beads.RepoContext.Role()` (internal/beads/context.go:570-573) returns `Contributor` unconditionally whenever `IsRedirected` — which `-C` makes true. So after this lands, `bd -C target context` reports `contributor` while `bd -C target create` routes as the target's configured `maintainer`, in the same process, off the same `RepoContext`. Consumers: cmd/bd/context_cmd.go:95, internal/storage/domain/fs/context.go:48. Pre-existing divergence, but this PR is what makes it observable. Follow-up bead, please.

4. **The `-C` fix does not follow a redirected target.** `GetRedirectInfo()` (internal/beads/beads.go:350-357) only inspects the *CWD* repo's local `.beads`. With `bd -C target` where `target/.beads` carries a redirect, `FindBeadsDir` follows it to storage, `isExternalBeadsDir` marks external, and `rc.RepoRoot` becomes the *storage* repo root — role comes from wherever the target's data lives, not from the target project. Symmetrically, exporting `BEADS_DIR=<workspace>/.beads` on a workspace that redirects makes `explicitBeadsSelection()` true and routes role detection to the storage root, i.e. the guard added in delta 2 is bypassed by that one shape. Neither is a regression (both previously read the caller), and neither is exercised by the new tests. A sentence in the `activeRepoPathForRouting` comment, or a bead, would close the gap in the record.

5. **Known residual is honestly disclosed and matches the code.** A `-C` target whose `.beads` is not in any git repo: `isExternalBeadsDir` errors, `isExternal` stays false, `git.GetMainRepoRoot()` returns the caller's root. Confirmed by reading internal/beads/context.go:119-138. Not a regression, and a non-git project has no `beads.role` to find. Fine as filed.

6. **No CHANGELOG entry.** This is a user-visible behavior fix and `## [Unreleased]` has a live `### Fixed` section. `CONTRIBUTING.md:313` lists the changelog under Release Process rather than per-PR, and roughly a quarter of recent `cmd/bd` commits on main touch it, so this is convention-mixed, not a rule — but a one-line `### Fixed` entry would earn its keep.

7. **The body's "two deltas" framing undersells main.go.** The load-bearing piece of delta 2 is the new `beadsDirProvidedAtStartup` global and `explicitBeadsSelection()` (main.go:830-838) plus the `.env` provenance flip (main.go:395-400) — a new process-wide selection-provenance concept, not just a guard. It is described in prose and fully commented in the code, so nothing is hidden; I would just have named the global in the summary. No undisclosed out-of-scope files: all six changed files are on the stated path, no persisted format, JSON shape or exit code moves.

## Attribution

`Co-authored-by: Osamu Okano` is on the port commits (`e48766cd3`, and the branch preserves it through the merges). Credit travels. Good.

PR #5244, opened 2026-08-01
test: collapse removed-backend admin command matrix

julianknutsen:
## What

Reduce `TestDoltAdministrativeCommandsRejectRemovedBackends` from a 3-backend × 8-command Cartesian matrix to one SQLite workspace exercising all eight real command entry points.

This removes 16 redundant `bd` subprocesses while retaining, for every command:

- a nonzero exit;
- every expected removed-backend guidance marker; and
- proof that no `embeddeddolt`, `dolt`, `beads.db`, or `.local_version` artifact was created.

## Why

The three removed backends do not interact with command-specific behavior:

- `dolt show` and `dolt status` load configuration and call `validateConfiguredBackend` directly.
- The other six commands call `loadDoltBackendConfig` → `requireDoltBackend` → `validateConfiguredBackend`.

All three backends converge on the same validation switch and `RemovedBackendError` before command-specific work. SQLite is the strongest representative here because it also exercises the unique `single engine` guidance branch.

Backend-specific PostgreSQL, MySQL, and SQLite coverage remains in the ordinary-command startup/no-write test, legacy `init` guidance, bootstrap dry-run and execute tests, both storage factories, `TestRequireDoltBackend`, and config/registry recognition tests.

No production code or behavior changes.

## Performance

| Focused test metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Real subprocess cases | 24 | 8 | −66.7% |
| Go test-body time | 2.30s | 0.79s | −65.7% |

## Verification

- Focused test: passed with exactly eight command subtests.
- Affected package: `./scripts/test.sh ./cmd/bd` passed in 446.873s (473.33s wall).
- Final baseline: `make test` passed in 496.52s with 39.0% aggregate coverage.
- Lint: `golangci-lint run ./...` reported zero issues.
- Two independent Sol blocker/major reviews passed with no findings.

Bead: `bd-skwi4`

### DIFF
--- a/cmd/bd/store_selection_removed_backend_test.go
+++ b/cmd/bd/store_selection_removed_backend_test.go
@@ -226,41 +226,37 @@ func TestDoltAdministrativeCommandsRejectRemovedBackends(t *testing.T) {
 		{name: "clean-databases", args: []string{"dolt", "clean-databases", "--dry-run"}},
 	}
 
-	for _, backend := range []string{configfile.BackendPostgres, configfile.BackendMySQL, configfile.BackendSQLite} {
-		t.Run(backend, func(t *testing.T) {
-			root := t.TempDir()
-			beadsDir := filepath.Join(root, ".beads")
-			if err := os.MkdirAll(beadsDir, 0o755); err != nil {
-				t.Fatalf("create beads dir: %v", err)
+	root := t.TempDir()
+	beadsDir := filepath.Join(root, ".beads")
+	if err := os.MkdirAll(beadsDir, 0o755); err != nil {
+		t.Fatalf("create beads dir: %v", err)
+	}
+	if err := (&configfile.Config{Backend: configfile.BackendSQLite}).Save(beadsDir); err != nil {
+		t.Fatalf("save config: %v", err)
+	}
+
+	for _, command := range commands {
+		t.Run(command.name, func(t *testing.T) {
+			cmd := exec.Command(bd, command.args...)
+			cmd.Dir = root
+			cmd.Env = append(removedBackendTestEnv(beadsDir), "BD_DISABLE_METRICS=1")
+			out, err := cmd.CombinedOutput()
+			if err == nil {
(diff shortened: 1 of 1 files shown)


maphew (approved):
Verified the convergence claim against main before reading the diff, and it holds exactly as the PR body states: all three removed backends land on the single `case BackendPostgres, BackendMySQL, BackendSQLite: return RemovedBackendError(...)` arm in `validateConfiguredBackend` — `dolt show`/`dolt status` call it directly after loading config, and the other six commands reach it via `loadDoltBackendConfig` → `requireDoltBackend`. So the 16 dropped postgres/mysql legs were re-testing backend-independent dispatch, and nothing command-shaped falls out: all eight entries survive in the collapsed table with the same three assertions (nonzero exit, guidance markers, no `embeddeddolt`/`dolt`/`beads.db`/`.local_version` artifacts).

The backend-specific message content is the only thing that varies per backend, and the pg/mysql markers (`general-purpose server databases`, `simple and resource-light`) stay covered end-to-end through the same subprocess mechanism in `TestCommandStartupRejectsRemovedBackendsBeforeLocalWrites` and `TestLegacyInitFlagsReachRemovedBackendGuidance`, plus unit-level in `TestRequireDoltBackend` and `TestStoreFactoriesRemovedBackendsFailLoud` — matching the residual-coverage inventory in the PR body. Picking SQLite as the representative is the right call since it's the one that exercises the distinct `single engine` branch.

One pre-existing observation, not a request on this PR: `backends.Register` reserves only the name `dolt`, so an extension backend registered as `sqlite`/`postgres`/`mysql` would pass `backends.Registered()` and skip the tombstone — the only theoretical way the three backends could diverge at this switch. No registrant does that today; reserving the removed names in `Register` would close it if we ever care.

Appreciated the thorough writeup — the entry-point trace and residual-coverage enumeration in the body made this review mostly a verification exercise. Approving.

PR #5245, opened 2026-08-01
fix(automerge): settled merges must mint a row_lock distinct from both parents (absorbed from #4682)

maphew:
## Why

Absorbs the three slices of #4682 (@julianknutsen) that our 2026-07-26 review offered to land with attribution "whether or not you take the re-scope" — every commit here carries `Co-authored-by: Julian Knutsen`. The headline is a live CAS hazard on main today: an LWW/field-merge settle writes the winning row while keeping one side's **old** `row_lock` token, so a stale `ExpectedVersion` captured before the merge can win a CAS it should lose. The PR built these ideas on a parallel `revision` column; this ports them onto the `row_lock`/`RowVersion` token main already has — no new column, no migration.

## What

1. **Fresh, distinct-from-both settle tokens** (`internal/storage/versioncontrolops/automerge.go`): a settle that changes an issues row stamps a `row_lock` minted fresh and distinct from BOTH parents' tokens (`freshRowLockDistinctFrom`, reroll over the crypto-random `FreshRowLock`). `row_lock` is excluded from cell classification, which also stops token-only divergence (equal `updated_at`) from declining otherwise-mergeable rows; no-write settles keep our token by design. Defensive: the stamp is gated on the conflict table actually carrying the column, so pre-0054 schemas degrade to prior behavior instead of failing the pull. The remint-on-any-settled-write is deliberately wider than `storage.go`'s lifecycle-only stamping contract and documented as accepted fail-safe behavior (a spurious CAS failure re-reads and retries; a stale token silently winning does not).

2. **AST completeness guard** (`internal/storage/issueops/row_lock_guard_test.go`): ports the PR's `TestAllIssueRowWritesBumpRevision` + teeth-test pair, retargeted at `row_lock` and substantially hardened through three adversarial review rounds — it now scans `issueops`, `domain/db`, **and** `versioncontrolops` (15 verified writes), matches backtick-quoted identifiers, requires SQL clause keywords (the PR's substring matcher counted error strings as writes), refuses marker-based exemptions for any write whose SET clause assigns `status`/`assignee`/`started_at`, checks the write's own literal before falling back to function- then file-scoped AST call detection, and exempts the deliberate whole-row token adoption in conflict resolution by function identity rather than spoofable text markers. Every check has a teeth case; the load-bearing ones are mutation-verified.

3. **Exemption analysis as documentation** (`internal/storage/issueops/lease.go`): the PR's analysis of why aux-marker writes (`is_blocked` recompute, compaction) must NOT remint — they would spuriously fail honest CAS checks — preserved as the `freshRowLock` doc comment, credited, pointing at the guard as the machine-readable exemption list.

## Validation

- Full `versioncontrolops` and `issueops` suites green; targeted dolt-store merge/row-version suites green; `make test` enqueued in the local verify queue at this head.
- Adversarial same-vendor review (needs-fixes → all seven findings applied) plus cross-vendor codex review (two P2s on guard coverage → both fixed with mutation-verified teeth); review logs recorded at each reviewed head.
- Guard hardening was probe-driven: each weakness was demonstrated against the real tree before being fixed (e.g. the PR's original matcher over-counted 26 writes where only 14 existed).

Relates to #4682 (absorb-now slices only; the metadata-CAS and `--if-revision` re-scope offered there remain the author's). Tracking: bd `mybd-n8an7`.

### DIFF
--- a/internal/storage/issueops/lease.go
+++ b/internal/storage/issueops/lease.go
@@ -67,6 +67,23 @@ func leaseTTL(ctx context.Context) time.Duration {
 // SAME lease row and conflict without any help. Adding a new path that sets
 // status/assignee outside updateIssueInTx without rewriting row_lock would
 // silently reintroduce the zombie-merge bug.
+//
+// The exemptions are load-bearing in the OTHER direction too (analysis
+// absorbed from gastownhall/beads#4682, Julian Knutsen): row_lock doubles as
+// the RowVersion optimistic-concurrency token (types.Issue.RowVersion), so a
+// path that reminted it WITHOUT changing the content a CAS holder cares about
+// would spuriously fail honest ExpectedVersion checks. The is_blocked
+// recompute is the canonical case: it is a denormalized aux-marker refresh
+// that already goes out of its way to preserve updated_at
+// (blocked_state.go), and bumping the token there would clobber a concurrent
+// whole-row CAS over a cell derived from OTHER rows. (The PR's second
+// exemption of this kind, the lease heartbeat, no longer arises here — since
+// bd-lrgn1 heartbeats never touch the issues row at all.) So the invariant
+// cuts both ways: status/ownership writes MUST remint, aux-marker writes MUST
+// NOT. The remint direction is enforced at build time by
+// TestAllIssueRowWritesStampRowLock (row_lock_guard_test.go), whose exemption
+// markers are the machine-readable form of this list — widen the stamping
+// policy there first, not by ad-hoc stamps.
 func freshRowLock() int64 {
 	var b [8]byte
 	if _, err := rand.Read(b[:]); err != nil {
--- a/internal/storage/issueops/row_lock_guard_test.go
+++ b/internal/storage/issueops/row_lock_guard_test.go
@@ -6,8 +6,13 @@ package issueops
 // in lease.go and types.Issue.RowVersion). The PR's substring stamp check
 // (strings.Contains(body, "revision")) was false-negative-prone — any comment
 // or identifier mentioning the column name counted as a stamp — so this port
-// tightens it to the actual mint forms: a freshRowLock()/FreshRowLock()/
-// RowLockClause() call in the enclosing function.
+// tightens it to the actual mint forms: a write's own literal must show a
+// genuine row_lock rewrite (literalStampsRowLock), and only the routed
+// funnels whose SET clause is itself assembled elsewhere (a %s placeholder)
+// fall back to an AST-based scan of the enclosing function for a
+// freshRowLock()/FreshRowLock()/RowLockClause() call (containsStampCall) —
+// never a text/regex scan, so a stamp mentioned only in a comment does not
+// count (comments are never part of the expression tree).
 
 import (
 	"go/ast"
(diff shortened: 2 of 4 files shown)


maphew:
Reviewed at `1c3ff07e77e`.

The production automerge change looks sound: settled rows receive a token distinct from both parent versions, and the focused and full package tests pass. One should-fix issue remains in the new completeness guard: `literalStampsRowLock` accepts an UPDATE solely because its SQL says `row_lock = ?`, and accepts an INSERT solely because its column list names `row_lock`; it never verifies that the bound value was freshly minted. A stale `RowVersion` or zero token can therefore pass the guard despite violating the invariant the 749-line guard is meant to enforce.

Please add mutation cases for reused/zero tokens and validate the enclosing call arguments, or route lifecycle writes through one stamping funnel and have the guard verify that funnel. If argument association is intentionally out of scope, narrow the guard's stated guarantee so it does not claim freshness it cannot prove.

Verification:

- Focused row-lock guard tests — PASS (`ok`, 1.868s).
- Focused automerge tests — PASS (`ok`, 1.752s).
- Full `internal/storage/issueops` and `internal/storage/versioncontrolops` packages — PASS (`ok`, 1.980s and 1.761s).

Disposition: **fix-merge**; the production fix is valuable and the remaining issue is confined to the regression guard.

steveyegge (approved):
## Review: APPROVE

Verified against the tree at 1c3ff07e7, not just the (excellent) PR narrative:

**The hazard is real on main.** Pre-PR, `row_lock` was classified as an ordinary data column, so an LWW settle stamped the winning side's *pre-merge* token onto post-merge content — a stale `ExpectedVersion` captured against that parent could then win a CAS it should lose. And the token-only-divergence half is load-bearing, not hygiene: `lwwUsable` requires `!ourUpdated.Equal(theirUpdated)` (automerge.go:363), so two replicas independently settling the same merge (different random tokens, equal `updated_at`) hit `cellContested && !lwwUsable` → **declined** merge that no timestamp could ever break. Excluding the token from classification fixes both.

**Mechanics check out:**
- The stamp is correctly gated inside `len(merge.columns) > 0` — no-write settles keep our token, which still vouches for byte-identical content (and reminting there would reintroduce the divergence loop).
- Pre-0054 degradation is sound end-to-end: `mustValue` is `row.value` with the error dropped (returns nil, no panic), so the unconditional `ourLock`/`theirLock` reads above the gate are safe when the column is absent, and the `hasRowLock` gate keeps `row_lock` out of the write-back plan.
- `dataColumns` was already variadic on main — no signature churn.
- No import cycle: versioncontrolops already imports issueops (commit.go, fastforward.go); issueops imports nothing back.
- No wisps gap: wisps are dolt-ignored tables (fastforward.go:53-73), never synced, so `dolt_conflicts_issues` is the only automerge surface that carries `row_lock`.
- `resolveOneConflictRow`'s function-identity exemption is honest: the `theirs` branch copies every `their_*` column verbatim (conflicts.go:387-403), so the adopted token vouches for exactly the adopted content.

**Guard quality.** The AST guard is genuinely toothy — I read all 14 teeth cases and they each pin a demonstrated false positive/negative, including the two that would have most worried me (a sibling write riding another write's stamp, and a stamp name appearing only in a comment). The file-scoped fallback for the plan/apply split is wider than ideal but documented as such, bounded to two specific call shapes, and refused when no evidence exists anywhere in the file. `verified 15 issues/wisps row writes` locally.

**Local validation:** `go build ./...` clean; `versioncontrolops`, `issueops`, `domain/db` suites all green at this head; guard + teeth tests pass verbosely.

The accepted widening (remint on any settled write, not just lifecycle) is the right trade — spurious CAS retry beats silent stale win — and the comment explicitly marks narrowing as a separate deliberate change. Nice absorption work, and the #4682 attribution is carried properly throughout.

PR #5250, opened 2026-08-01
refactor(test): isolate show --current selection

julianknutsen:
## What

Extract the direct `show --current` selection policy behind a private, one-method search role with injected actor and last-touched functions.

This replaces four real Dolt store fixtures plus global actor/store and filesystem choreography with four pure-Go recorder scenarios. It also removes the test file's CGO build constraint, so the policy now runs in default and `CGO_ENABLED=0` lanes.

The production adapter remains a one-line binding to `store`, `getActorWithGit`, and `GetLastTouchedID`. The separate proxied resolver is unchanged.

## Behavior preserved

- Nil store falls back without resolving the actor.
- Empty actor falls back without searching.
- In-progress work wins over hooked work.
- Searches keep the empty query and bare `Status`/`Assignee` filters.
- The first result wins; search errors continue to fall through.
- Last-touched fallback runs once after two misses.

No user-visible behavior changes.

## Performance and policy

| Focused workload metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Real Dolt stores | 4 | 0 | -100% |
| CGO-only test file | Yes | No | Pure-Go admitted |
| Package elapsed | 18.124s | 14.548s | -19.7% |
| Wall time | 43.19s | 22.37s | -48.2% |

The timing uses the same uncached `go test -count=1 -json -run '^TestResolveCurrentIssueID' ./cmd/bd` workload. Wall time includes build/package overhead and is therefore secondary to the structural gain: the policy no longer provisions a persistence engine.

## Verification

- TDD RED: the test-first change failed with the expected undefined/incompatible helper signature errors.
- Focused pure-Go and normal tests passed after implementation and after the council fix.
- Affected `cmd/bd` passed in 463.344s (494.38s wall).
- Final `make test` passed in 520.84s with 39.0% aggregate coverage; `cmd/bd` passed in 466.227s.
- `golangci-lint run ./...` reported zero issues; the pre-commit hook also passed.
- Two independent Sol blocker/major reviewers passed the final diff. One adapter-wiring major was fixed before commit and explicitly cleared on re-review.

Bead: `bd-s9p99`

### DIFF
--- a/cmd/bd/list_show_filter_modes.go
+++ b/cmd/bd/list_show_filter_modes.go
@@ -272,44 +272,38 @@ func gatherProxiedHierarchical(ctx context.Context, uw uow.UnitOfWork, parentID 
 	return out, nil
 }
 
+type currentIssueSearcher interface {
+	SearchIssues(context.Context, string, types.IssueFilter) ([]*types.Issue, error)
+}
+
 // resolveCurrentIssueID determines the current active issue for the agent.
 // Priority: in-progress assigned to actor > hooked > last touched.
 func resolveCurrentIssueID(ctx context.Context) string {
-	if store == nil {
-		// No store — fall back to last touched
-		return GetLastTouchedID()
+	return resolveCurrentIssueIDFrom(ctx, store, getActorWithGit, GetLastTouchedID)
+}
+
+func resolveCurrentIssueIDFrom(ctx context.Context, searcher currentIssueSearcher, currentActor func() string, fallback func() string) string {
+	if searcher == nil {
+		return fallback()
+	}
+	actor := currentActor()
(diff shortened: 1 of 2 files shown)


steveyegge (approved):
Traced the extracted resolveCurrentIssueIDFrom against the old inline body: nil-store and empty-actor fallbacks, in-progress-before-hooked ordering, empty query + bare Status/Assignee filters, first-result-wins, and error-falls-through are all preserved exactly; the loop is a faithful de-duplication of the two search blocks. Recorder-based tests cover each branch including call-shape assertions, and dropping the cgo constraint puts this policy in the pure-Go lanes. Nice cleanup.

PR #5251, opened 2026-08-01
fix(pr-preflight): warn when a PR-gate workflow is red for every PR while the base shows green

maphew:
## Why

`scripts/pr-preflight.sh`'s base-health check reads the base branch's own workflow runs, so a job that exists only in the PR workflow (e.g. the Contract corpus job in pr.yml) is invisible to it: preflight reported "Base branch main CI is green" across five PRs while every fresh PR was red in that job. Downstream automation that mirrors preflight's verdict inherits the blind spot by construction.

## What

A supplemental, deliberately **warn-only** PR-gate sample after the base-health block:

- Samples the last 60 completed `pull_request`-event runs, groups them by `workflowDatabaseId` (stable identity — display names are not unique and org/ruleset runs may lack one), and warns when any single workflow has >= 5 decisive runs, spanning >= 3 distinct head branches, all failure-class (`failure|timed_out|action_required|startup_failure`). The warning names the workflow and its newest failing run.
- It only speaks against a base branch known green: skipped when the base is red (the red-base message already explains PR redness) or undetermined (the "while the base branch shows green" explanation would contradict the warn printed just above).
- It never blocks, regardless of `PR_PREFLIGHT_BLOCK_RED_BASE`. That is load-bearing, not an oversight: automation that consumes preflight output classifies unrecognized `[block]` lines as genuine merge blockers and would park merge lanes — including a base-fix PR — on a false positive. A regression test pins the warn-only contract.
- Malformed-but-exit-0 `gh` output is treated as an empty sample rather than aborting the report under `set -euo pipefail`.

Tests: the fake-`gh` fixture now dispatches `run list` calls on the presence of `--event pull_request`, and `TestPRPreflightPRGateHealthSample` covers firing, naming amid healthy sibling workflows, one-success silence, insufficient-head silence, empty/unreadable samples, red-base and undetermined-base skips, and the warn-only contract under `PR_PREFLIGHT_BLOCK_RED_BASE=1`.

## Review

Cross-vendor reviewed (Claude reviewer, 2 rounds + codex gpt-5.6, 3 rounds). Adopted: per-workflow grouping (the original aggregate rule could never fire for a single broken workflow amid green siblings), warn-only contract, known-green gating, `workflowDatabaseId` identity, jq abort guards. Declined with reasons:

- *Filter the sample to runs whose PR targets this base*: requires one API call per sampled run; accepted as a documented limitation for a warn-only supplemental detector in a repo whose PRs overwhelmingly target `main`.
- *Treat `startup_failure` as red in the base-health check itself*: that classification predates this PR; extending the base check is out of scope here (the new detector does count `startup_failure` as failure-class).

Tracked in the mybd tracker as mybd-msll (from the rjc123 sweep follow-ups).

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -66,6 +66,16 @@ signals such as `.beads/` changes or missing tests, and the required
 contributor-protection next steps. It does not replace code review or local
 validation.
 
+Base-branch health is supplemented by a warn-only PR-gate sample: the last 60
+completed `pull_request`-event runs, grouped per workflow. A single workflow
+whose decisive runs (>= 5, across >= 3 distinct head branches) are all
+failure-class is reported as a broken PR gate — the case where a job that
+exists only in the PR workflow is red for every PR while the base branch shows
+green. This detector never blocks (deliberately: automation classifies
+unrecognized `[block]` lines as genuine merge blockers and would park merge
+lanes on a false positive); it is skipped entirely while the base branch is
+red.
+
 ## gh-body-lint
 
 Lint Markdown files before posting them with `gh ... --body-file`.
--- a/scripts/pr-preflight.sh
+++ b/scripts/pr-preflight.sh
@@ -292,6 +292,29 @@ else
   warn "Could not determine CI health of base branch ${base_ref}; check it manually before merging."
 fi
 
+# PR-gate health sample. The base-branch check above cannot see a job that
+# exists only in the PR workflow (pr.yml) - such a job never runs on the base
+# branch, so the check above can report green while every fresh PR is red in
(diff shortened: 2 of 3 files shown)


maphew:
Reviewed at `7ec832035b52`.

No findings. The PR-gate sample remains warn-only, is skipped unless base health is known green, groups by stable `workflowDatabaseId`, and handles unreadable or malformed samples safely. This is maintainer tooling rather than core product policy, so it raises no project-charter boundary concern.

Verification:

- `go test -tags=integration ./scripts -run TestPRPreflight -count=1` with Git Bash first on `PATH` — PASS (`ok`, 32.127s).

Disposition: **merge**.

PR #5252, opened 2026-08-01
fix(dolt): bind bootstrap recovery to database incarnation

### LINKED ISSUE #5012: flaky CI: proxied shard 1/15 — fresh-database init fails with 'pending schema migrations alter pre-existing dirty tables: issues' (+ mysql busy buffer)

zach-source:
## Summary

- replace fresh-bootstrap reset's sticky boolean with a private, one-shot capability
- bind authority to the exact endpoint, Dolt server UUID, database name, and pristine initial HEAD
- revalidate the database incarnation while holding the migration lock immediately before any hard reset
- consume authority before reset so an outer retry cannot reset twice
- carry the same proof through both the server-store and unit-of-work bootstrap paths
- commit init's intentional config state with `CommitWithConfig`, leaving a successful fresh init clean

## Safety properties

- only the winner of an exact bare `CREATE DATABASE` can receive reset authority
- a create-race loser and every pre-existing database fail closed
- probe errors, identity mismatch, endpoint mismatch, missing initial-HEAD ancestry, and consumed authority return the original `DirtyTablesError` without resetting
- drop/recreate of the same database name cannot reuse stale creator authority
- unit-of-work retries never infer new authority from `CREATE DATABASE IF NOT EXISTS`
- gateway, existing-database identity verification, forward-drift checks, connection cleanup, and smart remote-migration adoption retain current-main behavior

## Verification

- affected schema and Dolt packages: PASS
- full unit-of-work package: PASS (193.298s)
- exact upstream `gms_pure_go` mode: CLI, schema, Dolt, and unit-of-work PASS
- init config regression tests: PASS x100
- focused capability/reset race tests: PASS x10
- real Dolt 2.2.1 integration: PASS
- real Dolt 2.2.1 integration under race detector: PASS
- concurrent exact creation: exactly one capability across 16 workers
- real drop/recreate ABA: replacement v51 schema, foreign column, and dirty working set preserved
- pre-existing dirty database: preserved, no reset authority
- affected-package `go vet` with CGO disabled: PASS
- `git diff --check`: clean

The local host lacks ICU development headers required by the unrelated embedded-Dolt CGO dependency, so CLI compile/tests and vet were also run with `CGO_ENABLED=0`; the server-mode and real-Dolt acceptance paths above ran successfully.

Fixes #5012.

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -1762,11 +1762,13 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 		}
 
 		// Auto-commit Dolt state so bd doctor doesn't warn about uncommitted
-		// changes and users don't need a separate "bd vc commit" step. Skipped in
-		// gateway mode: the shared server owns its own history and the credential
-		// may be read-only, so bd must not issue DOLT_ADD/DOLT_COMMIT against it.
+		// changes and users don't need a separate "bd vc commit" step. Init
+		// intentionally writes issue_prefix to config, while ordinary Commit
+		// excludes config to avoid sweeping unrelated stale values. Skip this in
+		// gateway mode: the shared server owns its history and the credential may
+		// be read-only, so bd must not issue DOLT_ADD/DOLT_COMMIT against it.
 		if shouldWriteInitStateToDB(doltCfg.Gateway) {
-			if err := store.Commit(ctx, "bd init"); err != nil {
+			if err := commitInitState(ctx, store); err != nil {
 				// Non-fatal: some setups (e.g. no tables yet) may have nothing to commit
 				if !strings.Contains(err.Error(), "nothing to commit") {
 					fmt.Fprintf(os.Stderr, "Warning: failed to commit initial state: %v\n", err)
@@ -2970,6 +2972,18 @@ func isEmptyRemoteCloneError(err error) bool {
 	}
 	msg := strings.ToLower(err.Error())
 	return strings.Contains(msg, "contains no dolt data")
+}
+
+type initStateCommitter interface {
+	CommitWithConfig(context.Context, string) error
+}
+
+// commitInitState commits the configuration init deliberately created as well
+// as the initial schema state. The ordinary Commit contract excludes config to
+// avoid sweeping unrelated stale values, so using it here leaves every fresh
+// database dirty immediately after a successful init.
+func commitInitState(ctx context.Context, store initStateCommitter) error {
+	return store.CommitWithConfig(ctx, "bd init")
 }
 
 // verifyMetadata writes a metadata field and verifies the write succeeded.
--- a/cmd/bd/init_commit_test.go
+++ b/cmd/bd/init_commit_test.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"testing"
+)
+
+type recordingInitCommitter struct {
+	message string
+	err     error
+}
+
+func (s *recordingInitCommitter) CommitWithConfig(_ context.Context, message string) error {
+	s.message = message
+	return s.err
+}
+
+func TestCommitInitStateIncludesIntentionalConfig(t *testing.T) {
+	store := &recordingInitCommitter{}
+	if err := commitInitState(context.Background(), store); err != nil {
+		t.Fatalf("commitInitState: %v", err)
+	}
+	if store.message != "bd init" {
+		t.Fatalf("CommitWithConfig message = %q, want %q", store.message, "bd init")
+	}
+}
+
+func TestCommitInitStatePreservesCommitError(t *testing.T) {
+	want := errors.New("commit failed")
+	store := &recordingInitCommitter{err: want}
+	if got := commitInitState(context.Background(), store); !errors.Is(got, want) {
+		t.Fatalf("commitInitState error = %v, want %v", got, want)
+	}
+}
--- a/internal/storage/dolt/fresh_bootstrap_heal_integration_test.go
+++ b/internal/storage/dolt/fresh_bootstrap_heal_integration_test.go
@@ -1,0 +1,324 @@
+//go:build integration && !windows
+
+package dolt
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+	"time"
+
+	_ "github.com/go-sql-driver/mysql"
+
+	"github.com/steveyegge/beads/internal/doltserver"
+	"github.com/steveyegge/beads/internal/storage/doltutil"
+	"github.com/steveyegge/beads/internal/storage/schema"
+	"github.com/steveyegge/beads/internal/testutil/integration"
+)
+
+// TestFreshBootstrapHealIncarnation exercises the destructive bootstrap-heal
+// authorization against a real Dolt sql-server. In particular, the
+// drop/recreate subtest is the regression for stale creator authority crossing
+// a database-name ABA boundary and deleting another incarnation's work.
+func TestFreshBootstrapHealIncarnation(t *testing.T) {
+	integration.RequireDolt(t)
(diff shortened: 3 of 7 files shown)


zach-source:
Additional downstream verification for exact head `473b31cac7844f258fa56647ed56abc6b26ba273`:

- Gas City `b962a926b69d43998071db23b6ff893036643fa2` with an ephemeral `go.work` selecting this exact Beads checkout:
  - `CGO_ENABLED=0 go test -count=1 ./internal/beads/...` — PASS (all five packages; 32.02s)
  - `make test-native-doltlite-beads` — PASS
  - `make EXTRA_TEST_ENV=CGO_ENABLED=0 test-bd-cli-contract` with an exact `bd` built using `CGO_ENABLED=1 -tags gms_pure_go` — PASS (all four CRUD/dependency/destructive/workflow scenarios; 64.149s)
- Town's production Dolt pin, exact Dolt 2.1.7:
  - `go test -tags integration ./internal/storage/dolt -run '^TestFreshBootstrapHealIncarnation$' -count=3 -v` — PASS
  - `go test -race -tags integration ./internal/storage/dolt -run '^TestFreshBootstrapHealIncarnation$' -count=1 -v` — PASS

Identity checks: Beads tree `c6fdfe95aa29daf82991eb7715fb65324071f776`; source-built `bd` reported exact VCS revision `473b31ca...` with `vcs.modified=false`. Both source repositories remained clean.

Build-profile note: embedded `bd init` requires a CGO-enabled binary (`CGO_ENABLED=1 -tags gms_pure_go`). A CGO-disabled binary correctly fails closed and is suitable only for server/proxied-server use.

maphew (commented):
Reviewed at `473b31cac78`.

The one-shot, incarnation-checked bootstrap capability is a meaningful safety improvement over the sticky creator boolean, and the focused schema tests pass. Two items remain before landing:

1. `cmd/bd/init.go:1772` unconditionally uses `CommitWithConfig`, including re-init paths where `issue_prefix` was not written. Because this performs `DOLT_COMMIT('-Am')`, an ordinary re-init can sweep unrelated or stale dirty config into history—the #2455 hazard that normal `Commit` deliberately avoids. Use the config-inclusive path only when init actually wrote an allowlisted init-owned key; otherwise retain normal `Commit`.
2. The branch conflicts with current main in `internal/storage/uow/dolt_sql_provider.go`. The resolution is behavior-sensitive: preserve main's preview early-return before the new `justCreated` capture so previews remain strictly non-creating, non-migrating, and unable to acquire reset authority.

Verification:

- `go test -tags gms_pure_go -count=1 ./internal/storage/schema -run '^Test(MigrateUpWithLock(FreshBootstrapHeal|DirtyGuard)|MigrationLock)'` — PASS (`ok`, 0.950s).
- Focused `cmd/bd` init test compile — timed out without output; no local verdict.
- Merge-tree and blocking upstream preflight — both confirm a real current-main conflict; GitHub reports `DIRTY`.

The capability remains inside an existing recovery workaround, but bootstrap healing still depends on beads-side Dolt introspection/reset. That is storage-boundary debt; future expansion should move the primitive behind a deliberate driver/storage interface.

Disposition: **fix-merge on the contributor branch**. Request-changes is unnecessary if maintainers absorb these two contained fixes.

steveyegge (approved):
## Review verdict: APPROVE

Reviewed the full diff plus a local verification pass on this branch **merged with current main** (the branch predates #4985/#5232; see conflict note below).

**Design read.** This is the right hardening shape for the #5012 heal: the sticky `created` boolean becomes a private one-shot capability captured only by the exact bare-CREATE winner, bound to endpoint + `@@server_uuid` + database name + pristine initial HEAD (1 commit, clean status — any uncertainty refuses authority), revalidated **on the pinned, locked migration session after the dirty guard fires**, and CAS-consumed *before* `DOLT_RESET('--hard')` so an outer retry can never reset twice. Probe errors collapsing to `false` and returning the original unwrapped `DirtyTablesError` is exactly the fail-closed posture this path needs. The ancestry check (`initialHead ∈ dolt_log` on the current session) cleanly kills the drop/recreate ABA, and the uow path capturing authority only on `justCreated` — never from the sticky availability re-create via `IF NOT EXISTS` — closes the retry-loop hole. The deliberate wedge-to-pre-heal-behavior tradeoffs (capture failure fails the open; consumed capability returns the refusal) are correct: worst case is the old manual-intervention behavior, never a wrong reset.

**`CommitWithConfig` in init**: matches the documented GH#2455 contract — that API's own comment names `bd init` as an intended caller. Good catch on fresh inits being left permanently dirty.

**Local verification (merged with main @ a281dce3c):**
- `go build ./...` clean
- `internal/storage/schema`: ok (26s) — includes the six probe-failure fail-closed cases + one-shot test
- `internal/storage/uow`: ok (120s)
- `cmd/bd -run TestCommitInitState`: ok

**Non-blocking notes:**
1. Theoretical residual ABA: a drop/recreate that produces a byte-identical init commit (same committer identity + timestamp resolution) would satisfy the ancestry probe. Practically negligible; not worth chasing.
2. `consumeIfCurrentIncarnation` double-checks database name and endpoint from both the request and capability — redundant but harmless defense-in-depth; fine as is.

**Conflict note:** the branch conflicts with main in `internal/storage/uow/dolt_sql_provider.go` — #4985's `p.preview` early-return landed at the same insertion point as `justCreated := false`. The features are orthogonal (preview never creates or migrates, so it never reaches the capability path); resolution is preview block first, then `justCreated`. I'll push the merge resolution to this branch if maintainer-edit is enabled, else please rebase with that resolution.

Signed: claude-code-fable-5-high on behalf of Steve Yegge

steveyegge:
Merged at 6cbd63359 on the posted review verdict — all 86 checks green on the merged head (7b7b5ad conflict resolution vs #4985 included). Thanks @zach-source for the incarnation capability work; #5012's bootstrap-reset hole is closed.

PR #5254, opened 2026-08-01
fix(hooks): clone the issue handed to the async hook runner on update/close completion

steveyegge:
## Why

Regression from #5191, found during the bd-tveyi post-merge review (bd-9wgv3, P2). `CompleteIssueOperationUpdate`/`CompleteIssueOperationClose` (internal/storage/hook_decorator.go) pass `result.Issue` **by pointer** into `hooks.Runner.Run`, which `json.Marshal`s it on a spawned goroutine (internal/hooks/hooks_unix.go) — while the CLI mutates the very same object on the main goroutine right afterwards (`Dependencies = nil` at cmd/bd/close.go:255, update.go:548, reopen.go:103, before their own JSON output). That is a genuine unsynchronized write-while-read (the race detector flags it), observable as a nondeterministic hook stdin payload: the `dependencies` array is present or absent depending on goroutine scheduling.

The legacy paths were immune: `fireHookByID` re-fetches a private copy, and the create path already clones via `cloneIssueForHook` — so only Update/Close/Reopen (which funnels through `CompleteIssueOperationUpdate`) were exposed.

## What

Clone at the two completion entry points with the existing `cloneIssueForHook`. This fixes the race for every exposed verb on every backend (all three funnel through these two methods) and makes the payload deterministic as a side effect: hook scripts now always see the hydrated dependency snapshot taken at completion time — which subsumes the payload-shape drift vs pre-#5191 noted on the bead (old payloads came from an unhydrated `GetIssue`; the hydrated shape is the intended one going forward).

## Validation

- New regression test `TestHookFiringStoreCompleteIssueOperationsSnapshotIssue` pins both properties: the runner never receives the caller's pointer, and the recorded payload survives the caller's post-completion mutation.
- `go test -tags gms_pure_go -race -run 'TestHookFiringStore' ./internal/storage/` green (the new test fails on the pre-fix tree); full `./internal/storage/` suite green; `go vet` and `gofmt` clean.

Closes bd-9wgv3.

### DIFF
--- a/internal/storage/hook_decorator.go
+++ b/internal/storage/hook_decorator.go
@@ -305,15 +305,17 @@ func (h *HookFiringStore) CompleteIssueOperationCreate(ctx context.Context, issu
 }
 
 // CompleteIssueOperationUpdate fires the update hook for a committed guarded
-// operation.
+// operation. The issue is cloned because the hook runner marshals it on its
+// own goroutine while callers (cmd/bd close/update/reopen) go on to mutate
+// the result they handed in.
 func (h *HookFiringStore) CompleteIssueOperationUpdate(issue *types.Issue) {
-	h.fireHook(hooks.EventUpdate, issue)
+	h.fireHook(hooks.EventUpdate, cloneIssueForHook(issue))
 }
 
 // CompleteIssueOperationClose fires the close hook for a committed guarded
-// close.
+// close. Cloned for the same reason as CompleteIssueOperationUpdate.
 func (h *HookFiringStore) CompleteIssueOperationClose(issue *types.Issue) {
-	h.fireHook(hooks.EventClose, issue)
+	h.fireHook(hooks.EventClose, cloneIssueForHook(issue))
 }
 
 func (h *HookFiringStore) fireHookByID(ctx context.Context, event, id string) {
--- a/internal/storage/hook_issue_operations_test.go
+++ b/internal/storage/hook_issue_operations_test.go
@@ -115,6 +115,36 @@ func TestHookFiringStoreCompleteIssueOperationsFireOncePerCall(t *testing.T) {
 	}
 }
 
+// TestHookFiringStoreCompleteIssueOperationsSnapshotIssue pins the clone at the
+// completion entry points: the real hook runner marshals the issue on its own
+// goroutine, and the CLI mutates the result object it handed in right after
+// completing (cmd/bd close/update/reopen nil out .Dependencies), so handing
+// the runner the caller's pointer is a data race with a nondeterministic
+// payload (bd-9wgv3).
+func TestHookFiringStoreCompleteIssueOperationsSnapshotIssue(t *testing.T) {
+	runner := &recordingHookRunner{}
+	store := &HookFiringStore{runner: runner}
+	issue := &types.Issue{ID: "hook-issue", Dependencies: []*types.Dependency{
+		{IssueID: "hook-issue", DependsOnID: "dep", Type: types.DepBlocks},
+	}}
+
+	store.CompleteIssueOperationUpdate(issue)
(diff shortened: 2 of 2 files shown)


maphew (approved):
Reviewed at `826b61d123c`.

No correctness findings. Cloning at the two completion entry points fixes the asynchronous marshal race for update, close, and reopen while reusing the existing complete issue-snapshot helper. `cloneIssueForHook` deep-copies the mutable serialized pointers and slices currently present on `types.Issue`, preserving hook payload semantics.

Verification:

- `go test -tags gms_pure_go -race -count=1 ./internal/storage -run '^TestHookFiringStoreCompleteIssueOperationsSnapshotIssue$'` — PASS (`ok`, 1.826s).
- GitHub substantive checks — all 84 PASS, with two expected skips.
- Blocking upstream preflight — PASS; merge state clean and no failed or pending checks.

Disposition: **approve / merge**.

steveyegge:
**Verdict: APPROVE** (lion, standing reviewer — posted as a comment since the shared account cannot formally approve its own PRs)

Verified the full claim chain against the PR head, not just the writeup:

- **The race is real.** `hooks.Runner.Run` spawns a goroutine (hooks.go:69) and `runHook` does the `json.Marshal(issue)` inside it (hooks_unix.go:47). The guarded-op driver hands `result.Issue` by pointer to `CompleteIssueOperationUpdate/Close` (hook_issue_operations.go:59/67/78 — all three verbs funnel through those two entry points), and cmd/bd then writes that same object (`Dependencies = nil` at close.go:255, update.go:548, reopen.go:103). Unsynchronized write-while-marshal; pointer identity confirmed, not just shape.
- **The fix is at the right layer.** Cloning at the two completion entry points covers every exposed verb on every backend in one place, and matches the file's existing idiom (the create/pending paths already clone at lines 356/373/377/422). `cloneIssueForHook` is nil-safe and deep enough: per-element copies of Dependencies and Comments, fresh slices for Labels/Metadata/BondedFrom/Waiters, `clonePtr` for all pointer fields — no shared mutable state escapes to the hook goroutine.
- **Immune paths stay immune.** `fireHookByID` re-fetches its own copy via `GetIssue`; unchanged.
- **The test pins both properties** (runner never sees the caller's pointer; recorded payload survives the post-completion mutation) using the synchronous recording runner, so it's deterministic — no reliance on goroutine timing or `-race` to fail. Confirmed the assertions can only pass with a clone.

One observation, not a change request: the clone now runs even when `h.runner` is nil or no hook script exists (fireHook/Run early-return after the copy). That's the pre-existing pattern on the create path and it's a per-CLI-operation cost, so fine as-is.

CI fully green including the race-detector storage run. Ship it.

steveyegge:
## Gate review verdict (lion, bd-9wgv3): **APPROVE — merge**

Independently verified in a scratch worktree at 826b61d12:

- **The race is real and the fix is the right shape.** Both completion entry points handed `result.Issue` by pointer to the async runner; the claimed mutation sites check out exactly (`cmd/bd/close.go:255`, `update.go:548`, `reopen.go:103` all nil out `.Dependencies` after completion). `cloneIssueForHook` is a genuine deep clone (dependencies, comments, labels, all pointer fields), so the snapshot is safe against element-level mutation too, not just the slice-header write.
- **Regression test does its job.** On the pre-fix tree (merge-base + the test file alone) it fails with `event 0 received the caller's issue pointer`; on the PR head the full `TestHookFiringStore*` set passes under `-race`.
- **Blast radius is minimal**: two call sites + doc comments; the create path and `fireHookByID` were already immune as described. Clone cost per mutating CLI call is negligible.
- CI: no failing checks.

The hydrated-payload determinism point (snapshot-at-completion subsumes the pre-#5191 unhydrated shape) is a reasonable design call and it's recorded on the bead — no change requested.

PR #5265, opened 2026-08-02
fix(doctor): one filter for both uncommitted-changes checks (#5260)

### LINKED ISSUE #5260: bug: bd doctor "Dolt Status"/"Dolt Locks" warn permanently on benign uncommitted bookkeeping tables (isIgnoredTable/isWispTable skip-list too narrow)
## Summary

On a **healthy** Dolt-backed store, `bd doctor` permanently reports:

```
Dolt Status: N uncommitted change(s)
Dolt Locks:  Uncommitted changes detected
```

The "uncommitted" tables are only **bookkeeping/aux** tables that the native
store leaves in the working set after issue mutations — never `issues` itself,
which is committed per-command. Because nothing ever commits these aux tables,
the warning is **self-fulfilling: it can never clear**, on every store, forever.
This is the same failure mode the code already guards against for `wisp_*`
tables (see the `isIgnoredTable` doc comment) — the skip-list is just too narrow.

## Reproduction (`has-repro`)

Any server-mode Dolt store that has serviced a few `bd update/claim/comment`s:

```sql
SELECT table_name, staged, status FROM dolt_status;
-- child_counters        modified
-- compaction_snapshots  modified
-- dolt_ignore           modified
-- dolt_schemas          modified
-- issue_snapshots       modified
-- labels                modified
-- metadata              modified
-- schema_migrations     modified
--   (issues: clean — committed per-mutation)
```
→ `bd doctor` emits `Dolt Status: 8 uncommitted change(s)` (StatusWarning) and
`Dolt Locks: Uncommitted changes detected` (StatusWarning).

## Root cause

Two independent checks scan `dolt_status` and skip only ephemeral wisp tables:

- **`cmd/bd/doctor/dolt.go` — `checkStatusWithDB` ("Dolt Status")** skips
  `isIgnoredTable(name)` = `{wisps, local_metadata, repo_mtimes}` ∪ `wisp_*`
  (dolt.go:412). None of the bookkeeping tables above are covered.
- **`cmd/bd/doctor/migration_validation.go` — `checkDoltLocks` ("Dolt Locks")**
  skips only `isWispTable(name)` = `{wisps}` ∪ `wisp_*` (migration_validation.go:512)
  — *narrower still* (drops even `local_metadata`/`repo_mtimes`), and the "Locks"
  name is misleading: it's a working-set-dirty check, not a lock check.

The two checks have drifted apart and neither matches the set of tables the
store actually leaves uncommitted.

## Fix directions (design decision)

1. **Doctor-side (stop the false positive).** Widen/unify the skip logic so the
   known non-`issues` bookkeeping tables (`child_counters`, `compaction_snapshots`,
   `issue_snapshots`, `labels`, `metadata`, `schema_migrations`, `dolt_schemas`,
   `dolt_ignore`, …) don't raise a warning. Prefer deriving the skip-set from a
   single canonical source (the store's `dolt_ignore` patterns and/or one
   "non-issue tables" constant) and share it between `checkStatusWithDB` and
   `checkDoltLocks`, so it stays correct as the schema evolves. Note `cmd/bd/doctor`
   Dolt tests have no PR-CI path today (#4860) — worth wiring up alongside.

2. **Store-side (make the working set genuinely clean).** Sweep-commit the
   bookkeeping tables together with the issue mutation (or on a periodic sweep).
   This is the *more correct* fix **iff** these tables are meant to be durable —
   `labels`/`metadata`/`comments` are first-class, so leaving them uncommitted
   indefinitely may be a latent durability gap rather than pure cosmetics. Worth
   confirming. Related: #5111 (`bd dolt commit` skips the config table),
   #4078 (`bd remember` doesn't auto-commit in server mode).

**Recommendation:** ship (1) regardless — operators are currently trained to
ignore a permanent doctor warning, which masks real store problems — and
separately resolve the durability question behind (2).

## Impact
Consumed downstream by gascity's per-rig store-health dashboard, which escalates
every rig to a "warnings" rollup off this false positive
(gastownhall/gascity#4902).

## Environment
`bd` **v1.1.0**; Dolt server backend. Reported via gas-city ops (`gci-9ta`).

maphew:
Fixes #5260 (the drift half — see Scope below).

## Why

`bd doctor` has two checks that scan `dolt_status` for uncommitted tables, and each kept its own skip-list:

| check | source | skipped |
|---|---|---|
| **Dolt Status** | `dolt.go` `checkStatusWithDB` | `isIgnoredTable` |
| **Dolt Locks** | `migration_validation.go` `checkDoltLocks` | `isWispTable` |

`isWispTable` is a strict subset — it misses `leases`, `local_metadata`, `repo_mtimes` and `events`. It has been carrying the doc comment *"Deprecated: use `isIgnoredTable` for broader coverage"* for some time; its last caller was simply never moved.

On a healthy store those tables are legitimately dirty, so the same store is **clean by one check and permanently dirty by the other**. The warning is self-fulfilling: it can never clear, on every store, forever. Neither the name "Locks" (it is a working-set-dirty check, not a lock check) nor the suggested fix (`bd vc commit`) helps, because committing them is not what stops them being dirty.

## What

Both checks now go through one scan (`scanDoltStatus`) and one filter (`describeUncommittedTables`), so there is a single skip-set instead of two that can diverge again. `isWispTable` had no other callers and is removed; its test is repointed at `isIgnoredTable`.

Tests: the filter's behaviour is pinned directly (previously it was only reachable through a live Dolt connection), including an explicit anti-drift guard naming the four tables that diverged.

## Why the filter is still a static name list

The report suggests deriving the skip-set from the store's own `dolt_ignore` patterns. That was implemented first and then backed out, because `dolt_ignore` answers a different question:

> `doltdb.ShouldIgnoreDelta`: *"Only newly added or dropped tables are matched against the patterns. Changes to already tracked tables are always included."*

Every table in the skip-set is **tracked**, so `dolt_ignore` has no opinion about it — Dolt reports those tables dirty forever by design. `dolt_ignore` means "don't auto-add this *new* table", not "don't report changes to it".

It would also mean reimplementing `doltdb.compilePattern`, which is not SQL LIKE: `*`, `?` and `%` are wildcards, `_` is a **literal**, and matching is **case-sensitive**. Reaching for a LIKE matcher (as I did) is wrong in three separate ways and silently over-suppresses — it makes `wisp_%` match `wispXfoo`.

Which tables are benign bookkeeping is a beads decision, so it stays in beads code. The reasoning is a comment on `describeUncommittedTables` so it does not get re-litigated.

## Scope

This fixes the drift and the false positive it caused. The report also asks whether the remaining bookkeeping tables (`child_counters`, `labels`, `metadata`, `schema_migrations`, …) should be silenced here or committed by the store instead — that is a durability question (are those tables meant to be versioned?) and is deliberately **not** decided here. With this change, whatever the answer, both checks agree.

Behaviour note: detail formatting is now shared, so "Dolt Status" renders `table: status (staged)` with the single space "Dolt Locks" already used, instead of leaving a trailing space when a table is unstaged.

## Validation

- `go build ./cmd/...` and `go vet ./cmd/bd/doctor/` clean (`CGO_ENABLED=1 -tags gms_pure_go`).
- New and existing filter tests pass.
- Full `./cmd/bd/doctor/` suite: two failures, `TestCheckRepoFingerprint_UsesTargetRepoOutsideCWD` and `TestCheckTestPollution_NoTestIssues_EmptyDB`. **Both reproduce identically on an unmodified checkout of `upstream/main` @ db966433e**, so they are pre-existing and local to this machine (they pass in CI). Tracked locally; flagging in case they are known.
- Cross-vendor review pass (Codex `gpt-5.6-sol`, high reasoning) on the final commit: no findings. Two earlier rounds on earlier revisions produced four findings, all confirmed against the vendored Dolt source and all resolved — three of them by the revert described above.

### DIFF
--- a/cmd/bd/doctor/dolt.go
+++ b/cmd/bd/doctor/dolt.go
@@ -418,10 +418,68 @@ func isIgnoredTable(tableName string) bool {
 	return strings.HasPrefix(tableName, "wisp_")
 }
 
-// isWispTable returns true if the table name refers to a wisp (ephemeral) table.
-// Deprecated: use isIgnoredTable for broader coverage.
-func isWispTable(tableName string) bool {
-	return tableName == "wisps" || strings.HasPrefix(tableName, "wisp_")
+// doltStatusRow is one row of the dolt_status system table.
+type doltStatusRow struct {
+	table  string
+	staged bool
+	status string
+}
+
+// scanDoltStatus reads a `SELECT table_name, staged, status FROM dolt_status`
+// result set. Rows that fail to scan are skipped rather than aborting the
+// health check: a doctor check reporting what it could read beats one that
+// reports nothing.
+func scanDoltStatus(rows *sql.Rows) ([]doltStatusRow, error) {
+	var out []doltStatusRow
+	for rows.Next() {
+		var r doltStatusRow
+		if err := rows.Scan(&r.table, &r.staged, &r.status); err != nil {
+			continue
+		}
+		out = append(out, r)
+	}
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+	return out, nil
+}
+
+// describeUncommittedTables drops dolt_ignore'd tables from a dolt_status scan
+// and renders the rest as human-readable "table: status (staged)" strings.
+//
+// Every uncommitted-changes check in doctor must funnel through this function.
+// Two of them previously kept private skip-lists that drifted apart — "Dolt
+// Locks" excluded only wisp tables while "Dolt Status" excluded the full
+// ignored set — so the same clean store was clean by one check and dirty by
+// the other, permanently. One filter, one source of truth.
+//
+// The filter is a static name list rather than a read of the store's own
+// dolt_ignore rows, which is worth explaining because the reverse looks
+// obvious. dolt_ignore does not mean "do not report changes to this table".
+// Per doltdb.ShouldIgnoreDelta, "only newly added or dropped tables are
+// matched against the patterns. Changes to already tracked tables are always
+// included." Every table listed here is tracked, so Dolt reports it dirty
+// forever by design and dolt_ignore has no say in it. Deriving the skip-set
+// from that table would therefore borrow a mechanism that answers a different
+// question, and would additionally require reimplementing Dolt's pattern
+// syntax — doltdb.compilePattern treats '*' and '?' and '%' as wildcards, '_'
+// as a literal, and matches case-sensitively, none of which is SQL LIKE.
+//
+// Silencing these tables is a beads decision about which tables are
+// bookkeeping, so it belongs in beads code. See isIgnoredTable.
+func describeUncommittedTables(rows []doltStatusRow) []string {
+	var changes []string
+	for _, r := range rows {
+		if isIgnoredTable(r.table) {
+			continue
+		}
+		mark := ""
+		if r.staged {
+			mark = " (staged)"
+		}
+		changes = append(changes, fmt.Sprintf("%s: %s%s", r.table, r.status, mark))
+	}
+	return changes
 }
 
 // checkStatusWithDB reports uncommitted changes in Dolt using an existing connection.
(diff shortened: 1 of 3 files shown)


maphew:
Reviewed at `02cb0c5d824`.

No correctness findings. The shared `scanDoltStatus` / `describeUncommittedTables` path removes the actual drift source while preserving prior scan-error behavior, and both doctor checks now suppress the same bookkeeping tables. The regression tests cover the previously divergent names and staged formatting.

Verification:

- Focused `go test` attempts — timed out under concurrent compiler load without output; no local verdict.
- GitHub checks — 80 successful and four still running at inspection time.
- Blocking upstream preflight — blocked on the still-pending checks and `UNSTABLE` merge state; no merge action was taken.

Disposition: **merge once the remaining checks finish and preflight is clean**.

Issue #4282, opened 2026-05-31
Embedded dolt sql-server daemons are never reaped on inactive projects — 7 orphaned servers, several at ~38% CPU / ~2 GB RSS, ~67 W battery drain


csells:
## Summary

`gc`/`bd`-managed projects each start an embedded `dolt sql-server` daemon, but these daemons are **never shut down when a project goes inactive**. They accumulate across projects and keep consuming significant CPU (and ~2 GB RSS each) for days with no client connected, draining battery on laptops.

On my machine right now there are **7 orphaned `dolt sql-server` daemons**, several spinning at 8–38% CPU despite the projects being untouched for up to 6 days. Combined with the per-`bd`-call `dolt` subprocesses and `dolt send-metrics`, total Dolt CPU repeatedly spikes to 300–600%, pulling **~67 W** on battery (≈1.4 h of runtime from a full charge).

## Environment

- Hardware: MacBook Pro, Apple M5 Max, 128 GB
- macOS 26.5
- dolt 1.86.6
- bd 1.0.3 (Homebrew)

## Evidence

Live `ps` snapshot of `dolt sql-server` daemons (none of these projects were active; the busiest had no foreground client):

| PID | %CPU | RSS | Age | Project (config path) |
|----:|-----:|----:|----:|------|
| 3565 | **38.2** | ~2.1 GB | 6d | `chris-city` |
| 5553 | **37.3** | ~2.6 GB | 6d | `mc-live-contract-…` |
| 28382 | **19.4** | ~1.8 GB | 5d | `gascity/formula-detail-demo-city` |
| 4549 | 8.6 | ~2.1 GB | 6d | `racoon-city` |
| 34002 | 0.0 | ~0.3 GB | 14h | `mc-city` |
| 15017 | 0.0 | ~0.07 GB | 2d | `gastownhall/gascity/cmd/gc` |
| 14035 | 0.0 | ~0.07 GB | 2d | `/tmp/city` |

Notable: the two 6-day-old idle servers (`chris-city`, `mc-live-contract`) are pinning ~38% CPU each **with no active client** — idle should be ~0% (cf. PIDs 34002/15017/14035 at 0.0%). So beyond the "never reaped" problem, some idle servers are stuck in a busy loop.

Additional Dolt overhead observed at the same time:
[...]

PR #5277, opened 2026-08-02
fix(doltserver): drain MySQL handshake before closing probe connections

maphew:
## Why

A bare `Close()` on a probe connection that hasn't read the server's MySQL
handshake greeting causes the OS to send a TCP RST instead of a clean FIN.
`dolt sql-server` interprets that RST as an aborted MySQL handshake; enough
of them in a short window (from repeated readiness/circuit-breaker/doctor
reachability probes) can crash the dolt sql-server process. See #4132 for
the original report, including hamchowderr's Windows repro
(issuecomment-5014617950 on #4132), and #4133 (Zireael) for the diagnosis,
design, and original tests this PR is built from.

## What

- New `internal/doltserver.DrainAndCloseProbe(conn) bool` and
  `ProbeSQLServer(network, addr, timeout) (greeted bool, err error)`: a
  shared helper that reads (and discards) the MySQL handshake greeting
  before closing a probe connection, so `Close()` sends FIN, not RST.
- Swept every MySQL-speaking dial+close probe onto it:
  `doltserver.waitForReady`, `internal/storage/dolt`'s `dialProbe` var (used
  by both the socket-fallback resolver and covered by
  `socket_fallback_test.go`'s stub), the fail-fast dial-probe in
  `newServerMode`/`Open`, the circuit breaker's `probe()`, and four
  doctor/CLI reachability checks (`isDoltServerRunning`,
  `checkServerReachable`, the dolt-server-mode-mismatch check in
  `federation.go`, `testServerConnection` in `cmd/bd/dolt.go`).
- **F7 fix** in `waitForReady`: a dial that succeeds but never greets
  (TCP listener accepting connections, MySQL engine not yet writing) is no
  longer treated as ready — it keeps polling until a greeting arrives or the
  deadline passes, rather than declaring victory on TCP accept alone.
- Left non-MySQL probes (remotesapi gRPC/HTTP, SSH connectivity, HTTP(S)
  connectivity) as bare dial+close, each with a comment explaining why:
  there is no MySQL handshake to drain on those protocols.
- Tests: `internal/doltserver/probe_test.go` covers `DrainAndCloseProbe`
  and `ProbeSQLServer` directly (greeting vs. mute peer, dial-failure path),
  plus a ported `TestWaitForReadyEmitsRST` (exactly one connection against
  an immediately-greeting server) and a new
  `TestWaitForReadyRepollsUntilGreeted` (F7: mute listener times out and
  re-polls; delayed-greeting listener eventually succeeds). Adapted the
  pre-existing `TestWaitForReady` in `doltserver_test.go` to have its fake
  "slow server" listener write a greeting, since it previously relied on
  bare-dial-success being sufficient.

This is the drain half of #4133. **The config half (background Dolt workers
disabled via a `--config` YAML file) is not ported here** — it is
superseded by main's #4986, whose YAML-config implementation already
addresses #4133's F1 lint findings. The orphaned-server angle from #4133's
field report remains tracked separately by the existing #4282.

Credits: Zireael (diagnosis, design, and the original tests this PR adapts
from #4133), hamchowderr (the Windows repro that anchors the bug report).

Supersedes #4133 (drain portion). Refs #4132.

Co-authored-by: Zireael <3856578+Zireael@users.noreply.github.com>


maphew:
Maintainer review, posted as a comment since this is a maintainer-authored PR (review-then-merge discipline; the merge stays with the patrol on green checks).

**Verdict: ready to merge.** The sweep is disciplined: every MySQL-speaking dial+close probe in the tree now routes through one shared helper, the three non-MySQL probes (remotesapi gRPC/HTTP, SSH, HTTP(S)) are deliberately left bare *with a comment at each site saying why* — that per-site justification is what makes the sweep auditable rather than mechanical.

Reviewed at content head 18d9832 (current head 6287e2f adds only clean `main` merge-ups, no content change).

What was checked:

- **Semantics preservation** — one-shot reachability probes (circuit breaker, `testServerConnection`, doctor checks) still report reachable on dial success alone; the drain is purely a close-hygiene change there. Only `waitForReady` changes behavior: a TCP-accepting-but-mute server is no longer "ready" (the F7 fix), which is the correct strictness for a readiness gate and is pinned by `TestWaitForReadyRepollsUntilGreeted`.
- **Latency budget** — `DrainAndCloseProbe` adds at most ~110 ms against a mute peer (100 ms greeting deadline + 10 ms residual drain), bounded and asserted by test. Fast path (greeting present) is unaffected.
- **Tests** — new `probe_test.go` covers greeting/mute/unreachable at both API levels plus the repoll regression; `TestWaitForReady`'s fake listener was correctly taught to greet.
- **Cross-vendor pass** — Codex `gpt-5.6-sol` (high) reviewed the branch diff pre-PR: *"preserves existing reachability semantics… no actionable regressions identified."* Log: `.worktrees/.review-logs/18d9832….md`.

Two nits, neither blocking:

1. `TestWaitForReadyEmitsRST` asserts the *absence* of RST behavior (exactly one drained connection) — the name says the opposite of what it proves. Worth renaming in any later touch of the file.
2. `DrainAndCloseProbe`'s residual-drain loop relies on the 10 ms read deadline set once before the loop; a peer streaming faster than the deadline expires would be drained until first read error, which is fine, but a one-line comment noting the deadline is not refreshed per-iteration would save the next reader the same trace.

maphew:
Updated: merged current `main` (7e825441b) to clear the conflict. The only conflict was the import block of `internal/storage/dolt/circuit.go` — this branch adds the `doltserver` probe import, main added the typed-1105 `mysql` import; both call sites are independent, so both imports are kept. Verified nothing from either side was dropped (per-file both-direction diff comparison against the merge base), and `internal/doltserver` + targeted `storage/dolt` circuit/probe tests pass locally with `-tags gms_pure_go`.

One observation from re-review while merging, pre-existing and out of scope here: `internal/storage/dbproxy/server/doltserver.go` `waitReady` still does a bare `conn.Close()` on its readiness probe — structurally the same probe-RST pattern this PR fixes on the embedded path. Happy to do a follow-up for the proxied spawn path if wanted.

This has been green (86 checks) since before the drift; a review would be appreciated when someone has a moment.

steveyegge (approved):
## Review verdict: MERGE on green CI

Fable adversarial review (bee, beads-lane steward). The PR does what it says and does it correctly, and it chose right on the protocol question it never even discusses: a mid-handshake COM_QUIT is not actually possible (COM_QUIT is only valid post-auth), so drain-then-FIN is the most correct pre-auth close available. Verified in an isolated worktree on main (4e7cc88): `go build ./...`/`gofmt`/`go vet` clean, new probe tests green ×3, full `internal/doltserver` green, touched `storage/dolt` and doctor tests green (the two doctor failures in my run reproduce on unmodified HEAD — this box's global `git config beads.role` leaking, not your diff).

**What checked out:** all eight converted call sites are correct — `waitForReady` gains the genuinely stronger semantic (TCP-accept-without-greeting keeps polling; its only caller has a 10s deadline, so the strengthening is safe), the `newServerMode` drained-close is correctly placed (both error paths return before it; the auto-start retry reassigns `conn` into it), the circuit-breaker probe documents that it deliberately ignores `greeted`, and the non-MySQL probes left bare (SSH/HTTP/gRPC) each carry an explanatory comment. No import cycle. Happy-path cost ~10ms per successful probe — negligible at probe frequency. This plausibly reduces the known `[mysql] unexpected EOF` → connection-refused ambient flake class, with the caveat in the follow-up below.

**The "swept every MySQL-speaking dial+close probe" claim is falsified by two missed classes** — filed as a completion sweep rather than blocking this diff, since neither is a regression:

1. `internal/storage/dbproxy/proxy/endpoint.go:646` `probePort` — bare dial+slam-close against the proxy's **data port**, and `handleConn` dials the backend dolt for every accepted connection and relays bytes, so the RST propagates through the proxy to dolt: exactly this PR's class, one hop removed. Operationally live (large proxied-server fleets exist as of this week). No import cycle prevents using `doltserver.ProbeSQLServer` there.
2. `internal/testutil/testdoltcommon.go:59` `WaitForServer` and `internal/testutil/integration/harness.go:291` `WaitForPort` — dial+slam-close readiness loops, the identical pattern `waitForReady` just got cured of, inside the very CI jobs that suffer the dolt-death flake. The CI-side benefit largely lives here.

**Nits, no action required:** `greeted := err == nil && n > 0` (probe.go:207-209) misses Go's documented data+EOF-in-one-Read case — `n > 0` alone is more correct; the fixed 100ms greeting deadline is independent of the caller's dial budget (worth a doc comment on the asymmetry); `TestWaitForReadyEmitsRST` asserts the opposite of its name (it proves no RST — rename); the mute-server subtest has ~100ms timing margin on a loaded box (widen if it ever flakes).

Merging on green.

PR #5283, opened 2026-08-02
fix(dep): emit deduped stub for revisited tree nodes instead of dropping them

### LINKED ISSUE #5282: bd dep tree silently drops revisited nodes on diamond graphs — tree and dep list disagree on the direct-blocker set
## Symptom

On any diamond-shaped dependency graph, `bd dep tree` and `bd dep list` disagree about a bead's **direct blockers**, with no indication anything was elided. Two manifestations:

**A — a direct blocker vanishes from the direct level.** Beads D, E, F with edges `D→F (blocks)`, `E→F (blocks)`, `D→E (blocks)` (created in that order):

```
$ bd dep list D
  E: ... (open) via blocks
  F: ... (open) via blocks        <- two direct blockers

$ bd dep tree D
D: ... [BLOCKED]
    └── E: ... [blocks]
        └── F: ... [blocks]       <- F appears ONLY nested under E
```

`--json` has the same hole: F exists only as the depth-2 node; the direct `D→F` edge is not represented at all. The mermaid output is likewise missing the `D --> F` edge.

**B — a blocked node renders with no visible blockers.** Same triangle, opposite iteration order (`A→B`, `A→C`, `B→C`, C iterated first): C renders as a direct child, and B — genuinely blocked by C — renders with an empty subtree.

Which occurrence survives depends on child iteration order (observed: newest edge first), so edge *insertion order* silently changes what the tree shows.

## Why it matters

`bd dep tree` is the natural command for "why is this blocked". In manifestation A, when the surviving branch is a chain of *closed* blockers, the root renders `[BLOCKED]` while every visible blocker is closed — the reasonable reader concludes the graph is stale and either forces the work through or files a phantom bug. This cost us a wasted autonomous-agent investigation cycle plus a filed bug before the real shape was understood (our tracker: sys-7yjr8). A related trap: because rendering is DFS pre-order, a later direct child prints only after the *entire* transitive expansion of earlier siblings, so any bounded read (`| head -N`) of a large tree misses trailing direct blockers entirely.

## Root cause

`internal/storage/issueops/dependency_tree.go` `buildDependencyTreeInTx()`:

```go
if depth >= maxDepth || visited[issueID] {
    return nil, nil
}
```

A revisited node is dropped with no marker. Notably the tree renderer already has a `"(shown above)"` branch for duplicate nodes (`cmd/bd/dep.go` `renderNode`) — it is dead code, because duplicates never survive traversal. `--show-all-paths` is plumbed down to `GetDependencyTreeInTx` and then ignored, so it is currently a no-op too.

## Proposed fix

Re-emit a revisited node as a stub (`Deduped: true`, no recursion) so every edge stays visible. That revives the existing `"(shown above)"` renderer branch, carries `deduped: true` in `--json`, restores the missing mermaid edges, and lets `renderTree`'s root-`[BLOCKED]` computation see direct open blockers it previously missed. PR follows.

`--show-all-paths` (full re-expansion) is left as a separate follow-up; it needs per-path cycle detection and a renderer that doesn't key children purely by parent ID.

atbrace:
Fixes #5282.

## What

`buildDependencyTreeInTx` silently dropped any node already visited on another path. On diamond graphs this made `bd dep tree` and `bd dep list` disagree on the **direct-blocker set** (the dropped edge was absent even from `--json` and the mermaid output), and the mirror case rendered a genuinely blocked node with an empty subtree. Full analysis and repros in #5282.

This change re-emits the revisited node as a stub — `Deduped: true`, correct `Depth`/`ParentID`/`EdgeFromParent`, no recursion — so every edge stays visible:

- The tree renderer's existing `"(shown above)"` branch (dead code until now — duplicates never survived traversal) renders the stub with zero renderer changes.
- `--json` carries the stub with `deduped: true`.
- The mermaid output regains the previously-missing edges (its `nodesSeen` map already dedupes node declarations).
- `renderTree`'s root-`[BLOCKED]` computation (GH#3565) now sees direct open blockers it previously missed when they were first reached through a sibling branch.

Cycle safety is unchanged: a stub never recurses, so traversal still terminates on cyclic graphs.

`--show-all-paths` remains unimplemented (plumbed and ignored, as before) — noted in #5282 as a separate follow-up.

## Before / after (triangle: D→E, D→F, E→F, all blocks)

```
$ bd dep list D                          # both versions
  E: ... (open) via blocks
  F: ... (open) via blocks

$ bd dep tree D                          # before
D: ... [BLOCKED]
    └── E: ...  [blocks]
        └── F: ...  [blocks]             # direct D→F edge invisible

$ bd dep tree D                          # after
D: ... [BLOCKED]
    ├── E: ...  [blocks]
        └── F: ...  [blocks]
    └── F (shown above)                  # direct edge visible, subtree not repeated
```

## Testing

- New unit test `TestGetDependencyTreeInTxEmitsDedupedDirectBlocker` (sqlmock): fails on the old traversal (stub fetch never happens), passes now; asserts the stub's depth/parent/edge/`Deduped` and that the first occurrence stays a full node.
- `go test ./internal/storage/issueops/ ./internal/types/ ./internal/storage/domain/db/` green (373 issueops tests).
- Dogfooded a build on real data: a 10-node replica of the production graph that triggered the report (deep closed chain + diamond back-edge onto an open direct blocker). Unfixed bd renders the root with one direct child and buries the open blocker at depth 5; fixed bd adds `└── <id> (shown above)` at the direct level, and `dep tree`/`dep list` agree. Also verified `--json` now contains the stub node, and the second (mirror) ordering shows the previously-hidden edge under the blocked mid node.

### DIFF
--- a/internal/storage/issueops/dependency_tree.go
+++ b/internal/storage/issueops/dependency_tree.go
@@ -15,8 +15,28 @@ func GetDependencyTreeInTx(ctx context.Context, tx DBTX, issueID string, maxDept
 }
 
 func buildDependencyTreeInTx(ctx context.Context, tx DBTX, issueID string, depth, maxDepth int, reverse bool, visited map[string]bool, parentID string, edgeFromParent types.DependencyType) ([]*types.TreeNode, error) {
-	if depth >= maxDepth || visited[issueID] {
+	if depth >= maxDepth {
 		return nil, nil
+	}
+	if visited[issueID] {
+		// Diamond dependency: this node was already expanded on another path.
+		// Emit it again as a Deduped stub (no recursion) so every edge —
+		// in particular every DIRECT blocker of the root — stays visible.
+		// Dropping it made `bd dep tree` disagree with `bd dep list` on the
+		// direct-blocker set and let a blocked node render with no visible
+		// blocker (gastownhall/beads sys-7yjr8). The tree renderer shows
+		// stubs as "<id> (shown above)"; JSON carries deduped=true.
+		issue, err := GetIssueInTx(ctx, tx, issueID)
+		if err != nil {
+			return nil, err
+		}
+		return []*types.TreeNode{{
+			Issue:          *issue,
+			Depth:          depth,
+			ParentID:       parentID,
+			EdgeFromParent: edgeFromParent,
+			Deduped:        true,
+		}}, nil
 	}
 	visited[issueID] = true
 
--- a/internal/storage/issueops/dependency_tree_test.go
+++ b/internal/storage/issueops/dependency_tree_test.go
@@ -182,3 +182,87 @@ func treeIDs(tree []*types.TreeNode) []string {
 	}
 	return ids
 }
+
+// TestGetDependencyTreeInTxEmitsDedupedDirectBlocker reproduces the diamond
+// omission behind gastownhall/beads sys-7yjr8: root depends on mid and shared,
+// mid also depends on shared. When traversal reaches shared through mid first,
+// the direct root->shared blocker used to be dropped entirely, so `bd dep tree`
+// and `bd dep list` disagreed on the direct-blocker set (and a bounded reader
+// concluded the graph was stale). The visited node must be re-emitted as a
+// Deduped stub so every direct edge is visible.
+func TestGetDependencyTreeInTxEmitsDedupedDirectBlocker(t *testing.T) {
+	db, mock, err := sqlmock.New()
+	if err != nil {
+		t.Fatalf("sqlmock.New: %v", err)
+	}
+	defer db.Close()
+
+	mock.ExpectBegin()
+	expectIssue(mock, "root", "Root")
+	expectDependencies(mock, "root", []dependencyRow{
+		{id: "mid", depType: string(types.DepBlocks)},
+		{id: "shared", depType: string(types.DepBlocks)},
+	})
+	expectIssueBatch(mock, []string{"mid", "shared"})
+
+	// First child: mid, which expands shared beneath it.
+	expectIssue(mock, "mid", "Mid")
+	expectDependencies(mock, "mid", []dependencyRow{
+		{id: "shared", depType: string(types.DepBlocks)},
+	})
+	expectIssueBatchOne(mock, "shared")
+	expectIssue(mock, "shared", "Shared")
+	expectDependencies(mock, "shared", nil)
+
+	// Second direct child: shared again — already visited. The fix fetches it
+	// once more and emits a Deduped stub instead of dropping the edge.
+	expectIssue(mock, "shared", "Shared")
+
+	tx, err := db.BeginTx(context.Background(), nil)
+	if err != nil {
+		t.Fatalf("BeginTx: %v", err)
+	}
+	tree, err := GetDependencyTreeInTx(context.Background(), tx, "root", 5, false, false)
+	if err != nil {
+		_ = tx.Rollback()
+		t.Fatalf("GetDependencyTreeInTx: %v", err)
+	}
+	mock.ExpectRollback()
+	if err := tx.Rollback(); err != nil {
+		t.Fatalf("Rollback: %v", err)
+	}
+	if err := mock.ExpectationsWereMet(); err != nil {
+		t.Fatalf("unmet sql expectations: %v", err)
+	}
+
+	if len(tree) != 4 {
+		t.Fatalf("len(tree) = %d, want 4 nodes [root mid shared shared-stub]: %v", len(tree), treeIDs(tree))
+	}
+	stub := tree[3]
+	if stub.ID != "shared" || stub.Depth != 1 || stub.ParentID != "root" {
+		t.Fatalf("stub = id %q depth %d parent %q, want shared/1/root", stub.ID, stub.Depth, stub.ParentID)
+	}
+	if stub.EdgeFromParent != types.DepBlocks {
+		t.Fatalf("stub edge = %q, want %q", stub.EdgeFromParent, types.DepBlocks)
+	}
+	if !stub.Deduped {
+		t.Fatalf("stub.Deduped = false, want true")
+	}
+	full := tree[2]
+	if full.ID != "shared" || full.Depth != 2 || full.ParentID != "mid" || full.Deduped {
+		t.Fatalf("first occurrence = id %q depth %d parent %q deduped %v, want shared/2/mid/false", full.ID, full.Depth, full.ParentID, full.Deduped)
+	}
+}
+
+func expectIssueBatchOne(mock sqlmock.Sqlmock, id string) {
+	mock.ExpectQuery(regexp.QuoteMeta("SELECT 1 FROM wisps LIMIT 1")).
+		WillReturnError(sql.ErrNoRows)
+	rows := issueRows()
+	rows.AddRow(issueRowValues(id, id)...)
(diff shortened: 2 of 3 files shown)


maphew (commented):
The root-cause analysis here is genuinely excellent — you traced the silent drop to the bare `visited[issueID]` check in `buildDependencyTreeInTx`, correctly identified that the renderer's `"(shown above)"` branch was dead code waiting for exactly this producer-side fix, and kept `--show-all-paths` deliberately out of scope instead of letting the PR grow. The fix itself is structurally right: depth cutoff checked before the visited check (max-depth semantics preserved), stubs never recurse (cycle termination verified unchanged), and the additive `omitempty` `deduped` JSON field is the right instinct for output stability. The new unit test is a real regression test — exact depth/parent/edge/`Deduped` assertions, documented to fail on the pre-fix traversal. A diamond-graph tree walk gets this much scrutiny because `bd dep tree` output feeds both humans and JSON consumers; the depth of the write-up made the review tractable.

One real finding needs a fix before this merges, though — the dedup stub invalidates an assumption downstream:

**`filterTreeByStatus` breaks on the duplicate IDs this PR introduces** (cmd/bd/dep.go:1480). It builds its `parentOf` map keyed only by `node.ID`. Before this PR a `TreeNode` slice could never contain the same ID twice, so that held; now the full node plus its later `Deduped` stub is the *expected* shape. We reproduced the exact algorithm standalone against a synthetic diamond (root:closed → mid:closed → shared:open, plus the new stub root → shared) with `--status=open`: the stub's `parentOf[shared]=root`, processed later, silently overwrites the full node's `parentOf[shared]=mid`, so the ancestor walk keeps `root` and both `shared` entries but never marks `mid`. The filtered/JSON output then contains a node whose `parent_id` points at a node absent from the returned array, and the real ancestor chain — which the function's own comment promises to preserve — is dropped. That's the same failure class this PR sets out to fix, relocated into the `--status` filter path. A smaller echo of the same root cause: `outputMermaidTree`'s edge loop (dep.go:1261) can then emit an edge from an undeclared node ID — mermaid degrades to an unlabeled bare node rather than erroring, so that half is cosmetic.

The fix looks small: make the ancestor walk dedup-aware — e.g. skip `Deduped` stubs when building `parentOf` so the full occurrence's parent always wins, and walk ancestors from the full occurrence. A test pinning `--status` + diamond would close the gap the current (good) storage-layer test can't see, since the interaction lives entirely in cmd/bd.

Two non-blocking notes for whenever the file is next touched: the stub branch re-fetches the issue row via `GetIssueInTx` even though it was fetched on first expansion (a small id→issue cache would save the round trip), and the new `deduped` JSON field could use a line in the CLI docs.

Checks run in a detached worktree at e77ddb7: build clean; `./internal/storage/issueops/...` and the dep-tree CLI tests all pass (one unrelated `TestDepListCrossRigRouting` testcontainers infra failure, not this diff). Filed the `--status` interaction as a tracked follow-up on our side as well, so it won't get lost if it's easier for us to absorb.

Review provenance: delegated agent review (claude-sonnet, standalone algorithm reproduction) + maintainer adjudication.

jacobhausler:
Building this against current main, `TestWireTagBijection/TreeNode` in `internal/httpapi` fails: `types.TreeNode.Deduped` (added here, `json:"deduped,omitempty"`) has no matching property in `internal/httpapi/spec/openapi.v0.yaml`'s `TreeNode` schema, so the wire-bijection check flags it as "on the wire already" but undocumented. `make api-check` fails for the same reason. Adding a `deduped` (boolean) property to that schema's `TreeNode` definition resolves both — `TreeNode` aliases straight to `types.TreeNode` via `x-go-type`, so no generated code changes, only the spec. Worth flagging since this PR's own CI shows only a skipped autofix check, so the drift likely hasn't surfaced on your end yet.

jacobhausler:
Following up on the same build: this commit also regresses the TreeWalker conformance contract.

`WalkDependencyTreeInTx` calls `GetDependencyTreeInTx` (`internal/storage/issueops/tree_walk.go:180`, `:189`, `:193`), so the dedup-stub change reaches the walker. `backend/conformance/tree_walker_contract.go` pins the opposite invariant: `RunTreeWalkerRendersASharedSubtreeOnce` states a diamond shows the shared child under whichever parent the walk reached first and "there is no option to show it twice," and `RunTreeWalkerTerminatesOnACycle` requires that revisiting a node stop the descent with no node repeated.

Observed with this commit applied to current main:

```
--- FAIL: TestTreeWalkerContract/TerminatesOnACycle
    walk = [twk-cyc-a, twk-cyc-b, twk-cyc-c, twk-cyc-a], want [twk-cyc-a, twk-cyc-b, twk-cyc-c]
--- FAIL: TestTreeWalkerContract/RendersASharedSubtreeOnce
    the shared node appears 2 times, want 1
```

Both pass on unmodified `origin/main` (`4e7cc8818`) and fail with the commit applied — verified in both directions on `internal/storage/uow`. The conformance suite is also invoked from `internal/storage/dolt` and `internal/storage/embeddeddolt`, so the exposure likely isn't limited to one backend, though I only ran `uow`.

`internal/storage/issueops`' own package tests stay green because the conformance suite lives in `backend/conformance` and is driven from the storage packages, which is probably why this hasn't surfaced on the branch. It reads as a genuine conflict between two legitimate requirements on the same function rather than an oversight — the CLI wants every edge visible, the walker role pins strict single-occurrence.

steveyegge (changes requested):
Adversarial review at head e77ddb7 (unchanged since maphew's 8/3 review — the 8/10 timestamp is jacobhausler's comment, not a push). The core design is sound where it aims: depth-before-visited ordering, global dedup with stub-never-recurses cycle termination, muted '(shown above)' text render, `deduped: true` in JSON. But merged onto current main (6748ef6) it breaks two suites the branch never runs — all reproduced by execution, not speculation:

**BLOCKING:**
1. **TreeWalker conformance goes red multi-backend.** `WalkDependencyTreeInTx` consumes `GetDependencyTreeInTx` (tree_walk.go:180,189,193), so stubs reach the walker, and backend/conformance/tree_walker_contract.go pins the opposite: reproduced `TestTreeWalkerContract/TerminatesOnACycle` (walk=[a,b,c,a], want [a,b,c]) and `RendersASharedSubtreeOnce` (shared appears 2×, want 1), driven from uow/dolt/embeddeddolt. This is a genuine design conflict, not a mechanical fix: either the walker filters `Deduped` stubs or the contract is deliberately amended — please pick one explicitly in the PR body.
2. **Wire-tag bijection / `make api-check` red**: reproduced `TestWireTagBijection/TreeNode` — `deduped` is on the HTTP wire but absent from openapi.v0.yaml's TreeNode schema. Both CLI doors were covered; the third door (HTTP spec) was missed.
3. **maphew's --status filter finding still applies, relocated**: the algorithm now lives at tree_walk.go:66-99 (`PruneTreeByStatus`) with the identical hazard — `parentOf` is last-write-wins, so a later stub clobbers the full node's parent and the ancestor walk emits an orphan (a node whose ParentID names an absent node), exactly what the function's doc comment promises to prevent. Needs a dedup-aware fix (skip Deduped stubs when building parentOf) + a diamond+--status test.

**SHOULD-FIX:** cycle behavior is now observable (stub back-edge where previously nothing) but only the diamond shape is tested — add a cycle-shape test. And rebase; the branch predates #5529's dep-display work.

Why your CI is green: the failures live in conformance and httpapi packages the branch's own tests never run.

VERDICT: REQUEST-CHANGES (items 1-3; happy to re-review promptly on re-push — the underlying fix is worth landing)

PR #5284, opened 2026-08-02
fix(hooks): refuse symlinked and git-tracked hook writes, back up foreign hooks (bd-5vdt8)

steveyegge:
## Summary

Fixes the P1 reported via wy-81fnur: with `.git/hooks/pre-push` symlinked to a git-tracked team script, `bd hooks install` (and every path that triggers it — `bd config apply` drift gate, `bd init`, `bd doctor --fix`) appended the BEADS INTEGRATION block **into the tracked script** across 36 clones. `os.WriteFile` follows symlinks (O_TRUNC), so the write rewrote the link target inode; there was no symlink or tracked-file check anywhere in the write path, and injection into hooks bd knows it does not own (`IsBdHook=false`) took no backup.

## Changes

- **`guardHookWritePath`** (cmd/bd/hooks.go): refuses symlinked hook paths (error names the resolved target) and refuses writes to git-tracked hook files. Shared installs (`.beads-hooks/`) are exempt from the tracked check — that directory is deliberately committed ("Remember to commit .beads-hooks/").
- **`installHooksWithOptions`**: preflights all hook targets *before* writing any, so a refusal can't leave a half-installed hook set. Before first injecting into a non-bd hook, preserves the original as a one-time `hook.backup` sidecar (never overwritten on reinstall). The sidecar is inert for the migration state machine: a marker-managed hook classifies as `marker_managed` regardless of sidecars (cmd/bd/doctor/hooks_migration.go:129).
- **`applyHookMigrationExecution`** (cmd/bd/migrate_hooks_apply.go): same guard on migrated hook writes.
- **Flag help**: `--force`/`--chain` now say they are no-ops (the nolint at installHooksWithOptions already admitted this; the help text advertised behavior that never ran).

## Testing

- 5 new tests in `cmd/bd/hooks_guard_test.go`: symlink refusal (incl. no-partial-install assertion), tracked-file refusal via `core.hooksPath`, foreign-hook backup + reinstall no-clobber, shared-mode tracked exemption, migration-apply symlink refusal.
- `go test ./cmd/bd/ -run 'Hook|Hooks'` green (45s, includes husky/worktree/marker suites); `./cmd/bd/doctor/fix` green. The 4 pre-existing `cmd/bd/doctor` failures (BeadsRole/TestPollution/RepoFingerprint) are ambient on main (bd-vb2u0) and untouched by this diff.

Fixes bd-5vdt8.

### DIFF
--- a/cmd/bd/hooks.go
+++ b/cmd/bd/hooks.go
@@ -768,6 +768,52 @@ var hooksListCmd = &cobra.Command{
 	},
 }
 
+// guardHookWritePath refuses hook writes that would silently modify files
+// bd does not own (bd-5vdt8):
+//   - a symlinked hook path: os.WriteFile follows the link (O_TRUNC) and
+//     rewrites the target inode — e.g. a repo-tracked script the hook
+//     points at, dirtying every clone that shares it
+//   - a hook file tracked by git: writing it dirties the working tree.
+//     allowTracked exempts shared installs (.beads-hooks/ is deliberately
+//     committed).
+func guardHookWritePath(hookPath string, allowTracked bool) error {
+	fi, err := os.Lstat(hookPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil // creating a new file — nothing to clobber
+		}
+		return fmt.Errorf("failed to stat %s: %w", hookPath, err)
+	}
+	if fi.Mode()&os.ModeSymlink != 0 {
+		target := "unresolvable target"
+		if resolved, rerr := filepath.EvalSymlinks(hookPath); rerr == nil {
+			target = resolved
+		} else if link, lerr := os.Readlink(hookPath); lerr == nil {
+			target = link
+		}
+		return fmt.Errorf("%s is a symlink to %s; writing would rewrite the link target, not the hook\nRemove the symlink (or leave that hook to its owner) and re-run", hookPath, target)
+	}
+	if allowTracked {
+		return nil
+	}
+	if isGitTrackedFile(hookPath) {
+		return fmt.Errorf("%s is tracked by git; bd will not modify committed files\nUntrack it (git rm --cached) or move hooks to an untracked directory and re-run", hookPath)
+	}
+	return nil
+}
+
+// isGitTrackedFile reports whether path is tracked by git in the repository
+// containing it. Errors (not a repo, path inside .git/, no work tree) count
+// as untracked — the guard only blocks writes it can prove are unsafe.
+func isGitTrackedFile(path string) bool {
+	dir := filepath.Dir(path)
+	base := filepath.Base(path)
+	// #nosec G204 G702 - fixed "git" command; dir/base come from the hooks
+	// directory bd itself resolved, not user input
+	cmd := exec.Command("git", "-C", dir, "ls-files", "--error-unmatch", "--", base)
+	return cmd.Run() == nil
+}
+
 //nolint:unparam // force and chain kept for CLI flag compatibility; section markers make them no-ops
 func installHooksWithOptions(hookNames []string, force bool, shared bool, chain bool, beadsHooks bool) error {
 	var hooksDir string
@@ -813,6 +859,14 @@ func installHooksWithOptions(hookNames []string, force bool, shared bool, chain 
 		preservePreexistingHooks(hooksDir)
 	}
 
+	// Refuse the whole install up front if any target is unsafe to write —
+	// stopping midway through the loop would leave hooks half-installed.
+	for _, hookName := range hookNames {
+		if err := guardHookWritePath(filepath.Join(hooksDir, hookName), shared); err != nil {
+			return fmt.Errorf("refusing to install %s hook: %w", hookName, err)
+		}
+	}
+
 	// Install each hook using section markers (GH#1380).
 	// Only the content between markers is managed by beads; user content
 	// outside the markers is preserved across reinstalls and upgrades.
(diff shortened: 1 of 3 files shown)


steveyegge:
CI caught the tracked-file guard being too broad: `bd init` leaves `.beads/hooks/*` tracked in test repos (and teams commit that directory exactly like shared `.beads-hooks/`), so reinstall/upgrade was refused — TestEmbeddedHooks/hooks_install red. Pushed 6188f5bd4: the tracked refusal now exempts bd-owned hooks (section markers or legacy bd shim/inline), matching the shared-install policy; symlink refusal stays absolute. TestEmbeddedHooks green locally under BEADS_TEST_EMBEDDED_DOLT=1; new test TestInstallHooksAllowsTrackedBdOwnedHook pins the exemption.

steveyegge:
**VERDICT: APPROVE** (bd-5vdt8) — posting as a comment since the shared account can't GH-approve its own PRs.

Reviewed the diff and verified the load-bearing claims against the branch head (6188f5b):

- **Guard ordering is right**: symlink refusal first (Lstat, message names the resolved target), then the tracked check, with `allowTracked` exempting shared installs. `isGitTrackedFile`'s fail-open direction (errors count as untracked) is correct — inside `.git/hooks`, `git ls-files` errors out and the symlink check is the operative defense there, which matches the wy-81fnur incident shape exactly (symlinked `.git/hooks/pre-push` → tracked script).
- **Preflight verified at head**: the guard loop runs before any hook write in `installHooksWithOptions`, so a refusal can't half-install. Pinned by the no-partial-install assertion in `TestInstallHooksRefusesSymlinkedHook`.
- **Sidecar-inertness claim verified**: `classifyHookMigration` returns `marker_managed` on valid markers before ever consulting `HasBackupSidecar`, so the one-time `.backup` can't perturb the migration state machine.
- **Tracked-but-bd-owned carve-out is right** — refusing there would break reinstall for teams that commit their hooks dir; `TestInstallHooksAllowsTrackedBdOwnedHook` pins it.
- The new tests genuinely pin the behaviors (incl. backup no-clobber on reinstall and the migration-apply symlink refusal). CI fully green.

Two non-blocking nits, fine as a follow-up bead:
1. `applyHookMigrationExecution` guards per-write inside the loop rather than preflighting the whole plan — a refusal at write N leaves writes 1..N-1 applied. The install path got a true preflight; migration apply didn't.
2. In beads/shared modes `preservePreexistingHooks` copies hooks before the preflight runs, so a subsequent refusal leaves those copies behind. Additive and harmless, but same partial-mutation flavor.

— lion (wyvern crew, standing reviewer)

maphew (approved):
The guard ordering here is exactly right, and it's the part that's easiest to get wrong: `os.Lstat` first so the symlink refusal can't be defeated by the target being a regular file, the resolved target named in the error via `EvalSymlinks` with a `Readlink` fallback for dangling links (precisely the diagnostic the wy-81fnur reporter lacked), and `isGitTrackedFile` failing *open* so the guard only blocks what it can prove — inside `.git/hooks` git abstains and the symlink check remains the operative defense. The preflight loop is a real structural improvement, and `TestInstallHooksRefusesSymlinkedHook` genuinely pins the no-partial-install property (pre-commit precedes pre-push in `managedHookNames`, so the "other hooks never created" assertion fails without the preflight). The 6188f5b narrowing via `isBdOwnedHookFile` was the right shape, and the `--force`/`--chain` help-text correction is an honest small thing.

Verified in a detached worktree at 6188f5b: build clean, hook/guard/migration tests pass, CI fully green. We also ran three manual end-to-end repros; two surfaced things worth follow-ups.

**The one to handle before it bites someone** — the tracked-file refusal is a hard error with no escape hatch, and it fires on the deliberately-tracked composed-`core.hooksPath` convention. Reproduced: a repo with `core.hooksPath=.githooks` and a committed `.githooks/` tree of `bd hooks run <event>` wrappers (no BEADS markers, so `isBdOwnedHookFile` is false) now errors on install; that propagates through config_apply.go:202 as `applyStatusError`, and since the drift gate keeps reporting drift (markers never get written), `bd config apply`/`bd doctor --fix` sit in a permanent error state for such repos. Refusing is the right decision; erroring forever is the wrong failure mode, and `--force` can't serve as the override because this same PR documents it as a no-op. Two cheap directions, whichever fits bd-cxm9r's scope: have the drift-gate path report skipped-with-guidance instead of error, and/or revive `--force` as the deliberate override. The remediation text also deserves a soften — "Untrack it (git rm --cached)" is bad advice for a hooks tree committed on purpose.

Smaller items, none blocking:

- The migration-apply path guards per-write *inside* the loop rather than preflighting the plan (migrate_hooks_apply.go:259), so a refusal at write N leaves writes 1..N-1 applied and skips RetireOps — the exact failure the install path's preflight prevents. Same shape one layer up: in `--beads`/`--shared` modes `preservePreexistingHooks` runs before the preflight (hooks.go:877 vs 880). Both are small hoists.
- The `.backup` sidecar is created silently (repro: `--shared` over a foreign hook produced `.beads-hooks/pre-commit.backup` with no mention, in the very directory the command then says to commit) and is never consumed on the section-marker path — `uninstallHooks` only restores it in the legacy branch. Suggest printing one line naming the backup, and skipping the sidecar when the merged content already preserves the original inline.
- `guardHookWritePath` Lstats only the leaf, so a symlinked hooks *directory* isn't refused (repro: `.git/hooks` symlinked to an external dir wrote straight through). The dangerous in-repo variant is caught — but by the tracked-file check, not the symlink guard. Worth a docstring line scoping the guarantee, or an `EvalSymlinks` walk on the dir.
- Docs: this changes the outcome of a command that previously always succeeded, and the only place a user learns the refusal exists is the error itself — a short CHANGELOG note would be proportionate.

Filed the composed-hooksPath failure-mode item as a tracked issue on our side too, since our own coordination repos use exactly that convention.

Review provenance: delegated agent review (claude-opus, manual end-to-end repros) + maintainer adjudication.

steveyegge:
## Gate review verdict (lion, bd-5vdt8): **APPROVE — merge**

Independently verified in a scratch worktree at 6188f5bd4:

- **The guard covers the incident correctly.** Symlink refusal via `Lstat` catches the wy-81fnur shape (`.git/hooks/pre-push` → tracked team script) before `os.WriteFile` can follow the link; the tracked-file refusal handles the `core.hooksPath`-into-worktree variant, with the bd-owned and shared-install exemptions scoped sensibly. `isGitTrackedFile` failing open (errors = untracked) is the right polarity for a hygiene guard.
- **Ordering is right.** `preservePreexistingHooks` before the preflight means the guard evaluates the *final* target state (a preserved symlink would be caught), and the preflight-all-before-write-any loop makes refusal atomic — the test asserts no partial install and an unmodified link target.
- **All six new tests pass locally** (`go test ./cmd/bd/ -run 'TestInstallHooks…|TestGuardHookWritePath…|TestApplyHookMigration…'` green in 11s). Test quality is good: they drive the real `installHooksWithOptions` path, not the guard in isolation.
- CI: no failing checks; the 4 ambient doctor failures are pre-existing on main (bd-vb2u0) as stated.

Two non-blocking notes, no change requested: (1) preflight→write has an inherent TOCTOU window — fine for a local CLI hygiene guard, just don't later promote this to a security boundary; (2) `isBdOwnedHookFile` treats any file *containing* the section-marker string as bd-owned, so a foreign hook quoting the marker in a comment would slip the tracked check — obscure enough to ignore.

The `--force`/`--chain` help-text honesty fix is appreciated.

PR #5248, opened 2026-08-01
fix(metrics): skip the detached flusher spawn under BEADS_TEST_MODE

maphew:
## Why

Ten-plus test files hand-set `BD_DISABLE_EVENT_FLUSH=1` to stop the detached `bd send-metrics` child (spawned by `internal/metrics.MaybeSpawnFlusher`) from racing `t.TempDir`'s `RemoveAll` — see #5032 and 08881217b. Every new Dolt-backed test file has to remember that line or inherit the race. A source-level guard removes the whole class: the spawn decision now also honors `BEADS_TEST_MODE=1`, the same test-run flag the storage layer already checks.

Credit where due: rjc123 suggested exactly this guard during the #5031/#5032 work; this implements that suggestion.

## What

- `internal/metrics/spawn.go`: `shouldSpawnFlusher()` extracts the whole spawn gate (`BD_IS_FLUSHER` re-entry marker, `Enabled()`, `BD_DISABLE_EVENT_FLUSH`, and now `BEADS_TEST_MODE==1` via `inTestMode()`), and `MaybeSpawnFlusher` early-returns on it. `BD_DISABLE_EVENT_FLUSH` keeps working unchanged.
- `internal/metrics/spawn_test.go`: table test for `inTestMode` (exact-match on `"1"`, hermetic unset case — the repo's own `scripts/test.sh` exports `BEADS_TEST_MODE=1`, so the unset case clears it explicitly), plus a gate test asserting `shouldSpawnFlusher()` flips on `BEADS_TEST_MODE` with every other gate held open (both env vars cleared hermetically; CI exports `BD_DISABLE_EVENT_FLUSH=1` workflow-wide).

The existing `BD_DISABLE_EVENT_FLUSH=1` lines in test files are deliberately left in place; dropping them is a mechanical follow-up once this lands.

## Validation

- `go test -tags gms_pure_go ./internal/metrics/ -count=1` green in three environments: bare, `BEADS_TEST_MODE=1`, and `BD_DISABLE_EVENT_FLUSH=1 BEADS_TEST_MODE=1` (the CI shape).
- `go build -tags gms_pure_go ./...` clean; `gofmt` clean.
- Cross-vendor review pass done (Claude reviewer + codex gpt-5.6 via `pr-open`); both rounds' findings (vacuous spawn assertion, non-hermetic unset case, CI's workflow-wide `BD_DISABLE_EVENT_FLUSH` tripping the precondition) are fixed in the follow-up commits.

[...]

PR #5285, opened 2026-08-02
fix(gitignore): make all .beads/.gitignore writes append-only, never clobber local rules (bd-kaaz3)

steveyegge:
## Summary

Second slice of the wy-81fnur incident (P1 hook fix is #5284): bd rewrote `.beads/.gitignore` to its template, destroying the user's local `/issues.jsonl` + `/memories.jsonl` keep-exports-off-master negations. The template contains no negations, so the rules weren't overridden — they were clobbered by full-template overwrites in three independent places.

## Changes

- **`doctor.FixGitignore`** (`bd doctor --fix`): was `writeGitignoreTemplate` unconditionally; now delegates to the already-existing append-only `EnsureGitignoreForBeadsDir` — full template only when the file is missing, otherwise append the missing `requiredPatterns`. `EnsureGitignoreForBeadsDir` gains a trailing `chmod 0600` so FixGitignore's permission-tightening on pre-existing files (os.WriteFile's mode only applies at creation) is preserved.
- **`bd init`**: had its own copy-pasted `os.WriteFile(..., GitignoreTemplate, ...)`; now calls `doctor.EnsureGitignoreForBeadsDir(beadsDir)` (self-gating — no-op when up to date, so the `CheckGitignore` pre-check is gone too).
- **`WriteBeadsGitignore`** (proxied-server fs layer, `internal/storage/domain/fs/beads.go`): overwrote on any byte difference from the template; now writes the template only for a missing file, and otherwise appends the template's pattern lines (non-blank, non-comment) that the existing file lacks.

Not changed: `containsGitignorePattern` stays exact-trimmed-line matching. With every write path append-only, a user's semantically-equivalent variant now costs at most one canonical-pattern append (after which the check passes forever) instead of re-arming a clobber on every `doctor --fix`.

## Testing

Tests that pinned the clobber as intended ("FixGitignore replaces entire file, user comments will be lost", `TestFixGitignore_PreservesNothing`) now pin preservation: local content — including unicode, 100KB lines, and symlinked targets — must survive as a prefix, with all required patterns present afterwards. fs suite gains an appends-preserving-local case and a no-rewrite-when-all-patterns-present case.

- `cmd/bd/doctor -run Gitignore`: green (34 tests)
- Full `cmd/bd/doctor`: FAIL-set identical to main's 4 ambient failures (bd-vb2u0: BeadsRole/TestPollution/RepoFingerprint — untouched by this diff)
- `internal/storage/domain/...`: all green
- `cmd/bd -run 'Init.*Gitignore|Gitignore'`: green

Fixes bd-kaaz3. Companion to #5284 (both from wy-81fnur; bd-cxm9r opt-out knobs still to come).


steveyegge:
CI note for the reviewer: two transient failures on this run, both re-queued and both diagnosed as known infra flake classes, not this diff — (1) Managed-local proxied lifecycle: TestManagedLocalProxiedStopStartInterleave proxy.lock 5s timeout (filed bd-ill4f; lane passed on #5252 and #5284 today); (2) storage domain + uow: TestNewDoltServerUOWProvider_ConcurrentInstantiation 'fresh-bootstrap working set: got 1 dirty entries' + a ping 'invalid connection' — dolt working-set/connection churn under concurrency (the #5248 flusher-guard class). This diff's bootstrap write is byte-identical to before (missing file → full template); the append-only path only differs for pre-existing files. Full local run of internal/storage/uow on this branch: green in 142s.

steveyegge:
**VERDICT: APPROVE** (bd-kaaz3) — and the red check is not this PR's.

**CI red first**: `Test (storage domain + uow)` failed on both attempts, but the failing test is `TestNewDoltServerUOWProvider_ConcurrentInstantiation` — the documented open flake #4775 (it fired on a docs-only PR and passed on rerun of the identical head; #5124 already added a retry mitigation). The failure chain (`init schema → capture fresh database identity → verify fresh-bootstrap working set: got 1 dirty entries`) is a dolt working-set race in concurrent provider bootstrap. This diff touches no storage/dolt/schema/uow code, and an fs `.gitignore` write cannot dirty a dolt working set. Recommend `gh run rerun --failed`; merge on green.

**The fix itself**, verified against head (f52523d):

- **All three clobber paths converge on append-only**: `FixGitignore` delegates to `EnsureGitignoreForBeadsDir` (missing → full template, else append missing `requiredPatterns`); `bd init` calls the same, dropping its copy-pasted template write; fs `WriteBeadsGitignore` appends missing template pattern lines. `writeGitignoreTemplate`'s only remaining non-test caller is the missing-file branch.
- **Ordering safety checked**: the template explicitly bans negation patterns (the NOTE at its foot) and `requiredPatterns` contains none, so an appended line can never override a local negation via gitignore's last-match-wins — and nothing in the list matches `/issues.jsonl` or `/memories.jsonl`, so the wy-81fnur local rules survive with semantics intact, not just textually.
- **The test flips are the right kind**: the old tests pinned the clobber as intended behavior (`TestFixGitignore_PreservesNothing`, "user comments will be lost") and now pin preservation as a prefix plus all-required-patterns-present, across unicode/100KB-line/symlink/subdir cases; the fs suite adds appends-preserving-local and no-rewrite-when-covered.

One non-blocking nit: `EnsureGitignoreForBeadsDir` early-returns at `len(missing)==0` **before** the new trailing chmod, so a pattern-complete file with loose perms is no longer tightened — the old always-rewrite path always ended 0600. Move the chmod ahead of the early return (or add it to that branch) to fully preserve the old perm-tightening. Same shape in fs `WriteBeadsGitignore` (0600 only applies at creation). Fine as a follow-up.

— lion (wyvern crew, standing reviewer)

maphew (commented):
The negation-ordering question — the thing most likely to make "append-only" quietly wrong — was actually checked here rather than assumed: `requiredPatterns` contains no negations, the template bans them explicitly, and nothing in the list matches `issues.jsonl`/`memories.jsonl`, so the incident's local rules survive *semantically*, not just textually. We verified separately that the fs layer's completeness set is a strict superset of doctor's 23 `requiredPatterns`, so the two appenders converge rather than ping-pong. And the test rewrites are the honest kind — `TestFixGitignore_PreservesNothing` → `PreservesLocalContent` deletes a test that pinned the bug as intended behavior, replacing exact-template equality with the stronger prefix + no-missing-patterns pair.

Verified in a detached worktree at f52523d: build clean; `./cmd/bd/doctor -run Gitignore`, `./internal/storage/domain/fs/`, and the init gitignore tests all pass.

One regression to fix before merge, one line: **`EnsureGitignoreForBeadsDir` early-returns at `if len(missing) == 0` (gitignore.go:193) before the new unconditional chmod at line 221.** The old `writeGitignoreTemplate` chmod'd 0600 on every call; now a `.beads/.gitignore` that already has every required pattern but sits at 0644 is never tightened by `bd doctor --fix` — the chmod's own comment ("Tighten permissions on pre-existing files") states exactly the intent the early return defeats. Hoist the chmod above the return (and note neither existing permission test covers this case — both `hadOldContent` cases seed missing patterns, so they always take the append path). Or, if 0600 was never actually load-bearing for this file, drop the chmod deliberately — but that's a decision to make, not a default to fall into.

CI: the single red is `TestNewDoltServerUOWProvider_ConcurrentInstantiation` in the storage-domain+uow job — a dolt provider-bootstrap race; this diff touches no storage/dolt/uow code and the local runs are green. Needs a rerun, not a change.

Follow-ups we'd be glad to see but that shouldn't hold this up (tracked on our side as well):

- Nothing pins the anti-duplication invariant — a four-line test calling `FixGitignore` twice and asserting a byte-identical file would protect the defining property of an append-only design against anyone who later makes `containsGitignorePattern` smarter than exact-trimmed-line matching. Same gap on the fs side.
- Both layers emit the identical `# Added by bd (missing required patterns)` header from two different completeness sets, so doctor-fix followed by a proxied-init append yields two header blocks (converges, but accumulates one per future template addition). One shared pattern source would remove the class.
- `bd init` quietly drops the redirect-resolving gate (init.go:962 now writes the local beadsDir where the old path consulted `CheckGitignore(cwd)` through `.beads/redirect`). Probably an improvement — the old code checked the target but wrote locally — but worth a line in the PR body; no test covers init-under-redirect.
- Nit: the new `targetContent != content` assertion in `TestFixGitignore_Symlink` can't fail (both reads resolve to the same inode) and replaced a real assertion.

Sequencing note for the queue: no overlap with #5284 (either order), but #4791 calls `EnsureGitignoreForBeadsDir` from bootstrap.go and its clean-tree expectations should be re-run against this append-only base after it lands.

Review provenance: delegated agent review (claude-opus) + maintainer adjudication.

steveyegge:
## Gate review verdict (lion, bd-kaaz3): **APPROVE — merge after a rebase/rerun turns the required gate green**

### The red required gate is not this PR — but don't merge over it

`Test (storage domain + uow)` failed on `TestNewDoltServerUOWProvider_ConcurrentInstantiation` (`verify fresh-bootstrap working set: got 1 dirty entries, want 0`). That is exactly the bd-zog57 bootstrap-capture race, fixed on main by #5309 — which landed **after** this PR's CI run and is **not** in this branch's base (verified: `git merge-base --is-ancestor 445965e27 pr-5285` says no). This diff touches zero uow files. Rebase onto current main (or merge main in) and rerun; the gate should clear on its own. Since it's a *required* check, merge on the green rerun, not over the red.

### Independently verified in a scratch worktree at f52523dd7

- **All three clobber paths are closed.** `FixGitignore` and `bd init` now delegate to the append-only `EnsureGitignoreForBeadsDir`; the fs-layer `WriteBeadsGitignore` writes the template only on a missing file and otherwise appends missing pattern lines. Nothing rewrites an existing file wholesale anymore.
- **No append-ordering hazard.** The template contains no `!` negation lines, so appending missing ignore patterns after user content can't override a local rule via last-match-wins — the incident direction (user-added ignores like the export negations) is safely preserved.
- **The check/fix pair stays coherent.** `CheckGitignore` tests existence + missing patterns only (not permissions), so the chmod-on-append is sufficient and an up-to-date file correctly no-ops. The doctor (`requiredPatterns`) and fs (template-lines) append sets differ but both are append-only with exact-trimmed-line dedup, so they converge rather than ping-pong.
- **Tests green locally**: `go test -run Gitignore ./cmd/bd/doctor/` and the full `./internal/storage/domain/fs/` suite both pass. Converting the tests that *pinned the clobber* into tests that pin preservation (unicode, 100KB lines, symlinked targets surviving as a prefix) is the right kind of fix — the old intent can't silently come back.

PR #5293, opened 2026-08-02
feat(issueops): put every issue capability behind one library surface

julianknutsen:
# Every issue capability behind one library surface

## What this branch does

Beads has had two implementations of most issue verbs: one for the **direct**
route (the CLI talking to a store it opened itself, embedded or `dolt
sql-server`) and one for the **proxied-server** route (the CLI talking through
a unit of work over a running Dolt server). The two drifted — different commit
messages, different transaction boundaries, different refusals, occasionally
different answers. This branch collapses that.

Seven role interfaces now live in the leaf package `issueops`, which imports
only `internal/types`:

| Role | Contract |
| --- | --- |
| `issueops.Lifecycle` | create / update / close / reopen, four verbs |
| `issueops.Reader` | get, ready listing, filters |
| `issueops.ReadyClaimer` | `ClaimNext` — select, CAS and hydrate in one transaction |
| `issueops.BatchCloser` | `CloseBatch` — many closes as one durable act |
| `issueops.DependencyEditor` | `AddDependencies` (all-or-nothing) / `RemoveDependency` (idempotent) |
| `issueops.Commenter` | `AddComment` — one append, one history entry |
| `issueops.Relations` | `Related` — an anchor's neighbors, pinned order, explicit direction |

Each role is reached only through a **store accessor**, never a constructor.
That is load-bearing: the accessor is where each decorator adds its layer, so a
role built any other way would silently lose hooks and telemetry.

Three implementations back every role — `internal/storage/dolt`,
`internal/storage/embeddeddolt`, and `internal/storage/uow` (the unit-of-work
sibling used by the proxied route) — plus a hook decorator
(`internal/storage/hook_*.go`) and a telemetry decorator
(`internal/telemetry/*.go`) for each.

### Per-role, per-route accessor call sites

Both routes reach the same role. The direct route calls the store's accessor;
the proxied route calls the provider's accessor through a small
`proxiedX()` helper.

| Role | Direct route | Proxied-server route |
| --- | --- | --- |
| `Lifecycle` | `cmd/bd/issueops_adapter.go:14` (`writeOps`), called from `create.go:576`, `reopen.go:69`, `update.go:502` | `cmd/bd/update_proxied_server.go:214` (`proxiedIssueLifecycle`), called at `:240` |
| `Reader` | `cmd/bd/show.go:158` | `cmd/bd/show_proxied_server.go:90` (`proxiedIssueReader`), called at `:439` and `update_proxied_server.go:303` |
| `ReadyClaimer` | `cmd/bd/ready.go:138` | `cmd/bd/ready_proxied_server.go:242` (`proxiedReadyClaimer`), called at `:202` |
| `BatchCloser` | `cmd/bd/close_direct.go:195` | `cmd/bd/close_proxied_server.go:223` (`proxiedBatchCloser`), called at `:107` |
| `DependencyEditor` | `cmd/bd/dep.go:36` (add) and `dep.go:1011` (remove) | `cmd/bd/dep_proxied_server.go:39` (`proxiedDependencyEditor`), called at `:63` and `:324` |
| `Commenter` | `cmd/bd/comments.go:291` | `cmd/bd/comments_proxied_server.go:170` (`proxiedCommenter`), called at `:188` |
| `Relations` | `cmd/bd/dep.go:883` | `cmd/bd/dep_proxied_server.go:52` (`proxiedIssueRelations`), called at `:373` |

The `Lifecycle` row is the one asymmetry, and it is a real gap, not a
simplification — see **Known gaps** below.

A depguard rule (`httpapi-domain-boundary`, `.golangci.yml`) now denies
`internal/storage/domain` from `internal/httpapi`, deny-by-default over
`internal/httpapi/**` with two named holes for values the embedder assembled
(`server.go`'s `Config.Workspace`, `handlers.go`'s `contextResponse`). Without
it, a handler could take `uw.IssueUseCase()` and reach past the roles entirely
— which is exactly the second front door this branch exists to close, arriving
one handler at a time and lint-clean the whole way. The rule was watched
failing on a deliberate violation before being trusted.

## Behaviour users will notice

These are consequences of moving to the roles, not incidental. Each is also in
the CHANGELOG's Unreleased section.

**`bd close a b c` is now one Dolt commit, not three.** The direct route used
to loop per id over `Lifecycle.Close`, producing N transactions and N commits
named `bd: close issue`. It now builds one `CloseBatchRequest` and writes one
commit named `bd: close a, b, c`. The proxied route already did this. It is one
request **per store**, not per id — `bd close` reaches a second store only when
ids route to another repo, and a transaction cannot span repos; a batch-level
error is reported against every id in that store's batch, and the other stores'
batches still run.

**`--claim-next` moved inside the close transaction.** It now shares the
closes' transaction and fires only when something actually landed. Two visible
consequences: the JSON `claimed` value is the POST-claim row the role hydrated
inside the transaction (in_progress, assigned) rather than the pre-claim row
the old two-step read; and it walks the ready order instead of reading the
single top row and warning when that one row was taken.

**The open-children guard moved into the transaction and was deleted from
`cmd/bd`.** The engine already enforced it inside the close's own transaction
and refuses with a `CloseOpenChildrenError` whose message is character-for-
character the line `cmd/bd` used to compose by hand. The hand-rolled copy was
a second implementation of a rule the role already states, and a read-then-write
window that made `bd close <child> <parent>` depend on argument order. One
visible delta falls out: an unsatisfied gate on an issue that *also* has open
children now reports the **gate**, because the gate check stayed in the
read-only pre-flight while the open-children refusal moved into the
transaction. Both were already refusals; only which one you see changed.

**`bd dep add --file` now honours deferred version commit.** In embedded
`batch`/`off` auto-commit mode a bulk add used to always write a version commit;
it now defers to the next explicit commit point, which is what every other
write verb already did.

**Embedded `bd comment` changed its commit message and author.** A comment's
Dolt commit is now the role's `bd: comment <id>` authored by `beads`, where the
CLI auto-commit wrote `bd: comment (auto-commit) by <actor> [<id>]` authored by
`root`. That is the spelling the Dolt server backend already wrote, and it is
one commit per comment either way.

**Dep verb commit messages are now the role's** on the direct route:
`bd: dep add <a> <b>` / `bd: dep remove <a> <b>`, replacing
`dependency: add <type> <a> -> <b>` in server mode and a generic
`bd: dep add (auto-commit) by <actor> [...]` in embedded mode.

**Four smaller proxied-route deltas:** `bd dep list` answers in the role's
pinned order (ascending by neighbor id, edge type breaking ties) rather than
whatever order the two dependency tables were read in; a `bd dep list` naming a
nonexistent issue reports that instead of printing an empty neighbor list
(matching the direct route); a `bd dep remove` that finds no edge records no
Dolt commit where it used to name one; and a single-line `bd dep add --file`
writes `bd: dep add <a> <b>` rather than `dependency: add 1 edges`.

**One proxied-claim message change:** the proxied claim's Dolt commit is now
`bd: claim ready <id>`, where it wrote `bd: ready --claim <id>` while the other
two backends already wrote the former.

**Molecule auto-close, `--suggest-next` and `--continue`** moved out of the
close transaction into a second unit of work on the proxied route — which is
where the direct route always ran them. A plain close still produces exactly
one commit; a close that actually auto-closes a molecule or advances a step now
produces two. On the direct route, molecule auto-close runs after *all* the
closes land rather than between them, so only which id's slot prints the
"Auto-closed completed molecule" line can move.

**One wire delta on claim:** the claimed issue — in the HTTP body and in
`bd update --claim --json` — now carries hydrated labels, which the unhydrated
domain read it replaced did not return. `openapi.v0.yaml` already declares
labels on `Issue`, so the frozen spec is untouched.

## The wisp regression, and why the pinning's justification was false

When the direct `bd dep add` route first moved onto `issueops.DependencyEditor`,
it pinned an edge's SOURCE to the issues plane. A wisp has no row in the issues
plane, so `bd dep add <wisp-id> <target>` broke — **and the two routes did not
even fail alike**:

- **direct route:** refused with `issue <id> not found`, from its
  source-existence check;
- **proxied route:** an `fk_dep_issue` foreign-key violation from the database.

Two doors, one command, two different errors for the same cause. That
divergence is itself the argument for this branch.

The comment justifying the pinning said the issues plane was "the only one its
unit-of-work sibling can reach." **That was false.** The unit of work reaches
`domain.DependencyUseCase`, which declared `AddWispDependencies` and a full set
of wisp twins — none of which had a production caller. The capability was there
the whole time; the pinning was written against a belief about the code rather
than the code.

All three implementations now route **per edge, by the edge's own source**,
which is the rule the stores' own dependency verb has always applied and the
one `bd dep remove` and `bd dep list` already read. A wisp-sourced edge goes to
`wisp_dependencies` with its event in `wisp_events`; an issue-sourced one to
`dependencies` and `events`. One request may mix the two: it is still one
transaction, the parent-child-first ordering and the whole-graph cycle gate both
span the two tables (a cross-plane cycle is still refused), and a refusal rolls
back both planes. Only the durable half is versioned, so a request made entirely
of wisp-sourced edges records no history entry and leaves the working set clean.

Behind the role, the domain layer's three bulk add verbs (`AddDependencies`
pinned to the issues plane, `AddWispDependencies` pinned to the wisp plane, and
`AddDependenciesSourceRouted`) collapsed into the one source-routed
`AddDependencies`. With one verb and one behaviour, the `planeOf func(...) bool`
parameter and the `pinnedPlane` constant-returning helper bought nothing but
indirection, so both are deleted along with the private `addBulk` wrapper that
had exactly one caller left. There is now no plane-pinned bulk variant for a
future caller to reach for by mistake.

One deliberate ordering change inside that collapse: the source classification
query now runs **after** the input-shape validation loop rather than before it.
That puts the nil-dep guard in front of the `dep.IssueID` dereference and stops
a malformed batch from issuing a read it is going to reject anyway. The only
observable difference is which error you get when the input is invalid *and* the
classification query also fails; both orders return an error, the new one
returns the more specific one.

Pinned by a shared contract in `internal/storage/conformance` run against all
three implementations — wisp-sourced routing, the mixed batch, a mixed-batch
refusal rolling back both planes, and a cross-plane cycle refusal — plus an
end-to-end front-door test, `TestEmbeddedDepAddWispSource`, which runs
`bd dep add <wisp-id> <target>` through the real CLI against a real embedded
store and asserts both the edge readback and that `dolt_log` count and
`dolt_status` are unchanged. That test was demonstrated RED against the pinned
routing before the fix was kept.

## Why six Provenance fields were deleted rather than wired

`Provenance` labels the history entry a request records. `UpdateRequest` and
`ReopenRequest` keep it — both have real production callers
(`cmd/bd/update.go`, `cmd/bd/reopen.go`, `internal/httpapi`).

Six others — `CloseRequest`, `CloseBatchRequest`, `ClaimNextRequest`,
`AddCommentRequest`, `AddDependenciesRequest`, `RemoveDependencyRequest` — had
none. Every one of those roles already composes an id-bearing commit message of
its own, so a caller that spelled a label could only restate the role's default
and, worse, make that default unreachable from production and therefore
untested.

For two of them the field is **not caller-settable even in principle**:

- `CloseBatchCommitMessage` names what **LANDED** — the batch is deliberately
  not all-or-nothing, so which ids survive is a result of the call;
- `ClaimNextCommitMessage` names the id the claim **WON** — the whole point of
  the role is that the caller names a question and the implementation picks the
  answer.

Neither is knowable before the call, so a caller-supplied label for them is not
a missing feature; it is a category error. The three implementations lose the
`HistoryEntry()` indirection on the six affected paths and pass the composed
message straight through, so the commit messages users see in `bd dolt log` are
unchanged.

## Conflict resolutions the squash will erase

**`CloseRequest.Provenance`'s per-id `Lifecycle.Close` call was deleted, not
kept.** Two slices developed in parallel against `bd close`'s direct route.
One gave `CloseRequest` a `Provenance` label with a day-one caller that set
`bd: close <id>` from the resolved id, on the per-id `ops.Close()` call. The
other deleted that call outright when it moved the direct route onto
`BatchCloser.CloseBatch`. They conflicted in `cmd/bd/close.go`, and the
resolution took the BatchCloser side.

The two are mutually exclusive by construction: a per-id label presumes N
calls, and the batch is **one** call whose message names what landed. Keeping
the label would have meant either reintroducing the per-id loop the batch
change exists to remove, or carrying a field the batch path cannot honour.
Taking the deletion also keeps the visible outcome the Provenance slice was
after — `CloseBatchCommitMessage` composes `bd: close <ids>` from what landed,
so the entry names the issues either way. What it leaves behind is
`CloseRequest.Provenance` with no production caller, which is why that field is
one of the six deleted above. `ReopenRequest.Provenance` keeps its caller in
`cmd/bd/reopen.go` and stays.

That reasoning lived in a merge commit message on the unsquashed branch; this
section is now its only home.

**The two final slices merged without textual conflict**, but their intents were
verified by grep rather than trusted:

- the verb collapse touched `internal/storage/domain/dependency.go`,
  `cmd/bd/link_proxied_server.go`, `internal/storage/uow/dependency_editor.go`
  and one domain test;
- the doc correction and front-door test touched `CHANGELOG.md`,
  `cmd/bd/dep_embedded_test.go` and
  `internal/storage/issueops/dependency_editor.go`.

Zero files in common. Post-merge greps confirmed all three dead verb names
(`AddDependenciesSourceRouted`, `AddWispDependencies`, `pinnedPlane`, `addBulk`)
are gone repo-wide, that both retargeted call sites resolve to the surviving
verb, and that the corrected staging comment and the new test both survived.

One dangling reference the collapse created was fixed in the integration:
`issueops/errors.go`'s `ErrDependencyCycle` doc named the domain helpers
"(add/addBulk)", and `addBulk` no longer exists; it now reads
"(add/AddDependencies)".

## Known gaps

Stated plainly. None of these is fixed here.

1. **Comment listing is on no role, on either route.** `bd comments <id>`
   reads the same way it always did on both doors. Reading a thread is a paging
   question with a cursor of its own, which `issueops.Commenter` deliberately
   excludes — but the consequence is that this branch's "one surface" claim
   does not cover comment reads.

2. **`bd dep tree` and `bd dep cycles` have no role.** Both still go straight to
   the store's own tree/cycle methods on both routes
   (`cmd/bd/dep.go:1048`, `:1200`).

3. **`bd dep list` with multiple ids and `--direction=down` stays on raw edge
   records** (`cmd/bd/dep.go:818`). `Relations` is anchored on ONE issue and
   answers with hydrated far-end issues; routing this through the role would
   mean N anchor probes and a hydration of every neighbor the output never
   prints. Both routes make the same carve-out, so they agree — but this path
   is not on the role.

4. **Per-edge line attribution is lost in `bd dep add --file`.** An edge refused
   by storage no longer carries a `line N:` prefix, because the role's request
   is all-or-nothing and names no edge index. Parse and id-resolution errors
   still name their line, so this only affects storage-level refusals — but a
   user with a 200-line dep file now learns *that* an edge was refused without
   learning *which*.

5. **`bd create` and `bd reopen` are on `Lifecycle` only on the DIRECT route.**
   The proxied route still calls `uw.IssueUseCase().CreateIssue` /
   `CreateWisp` (`cmd/bd/create_proxied_server.go:153,155`) and
   `ReopenIssue` / `ReopenWisp` (`cmd/bd/reopen_proxied_server.go:127,129`).
   These two commands still have two implementations. `bd update` is the only
   Lifecycle verb converged on both routes.

6. **Multi-store `bd close --claim-next` rides only the first store's batch.**
   `closeDirectRequest` (`cmd/bd/close_direct.go:136-144`) attaches `ClaimNext`
   only when `batch.store == claimStore`. A close whose ids fan out across repos
   claims from one of them, deterministically, and does not consider ready work
   in the others.

7. **The `is_blocked` staging gap.** `addDependencyEdgeInTx` stages the source's
   dependency table and, when a row was recorded, its events table. That is
   **not** every table the edge wrote: a blocking edge also flips the source's
   `issues.is_blocked`, and `issues` is never added to `ChangedTables`. On an
   issue-sourced blocking add that flip stays in the working set and is swept
   up by whatever command auto-commits next — including a command that was
   REFUSED, which then appears in history having changed nothing of its own.
   This **predates this branch**: `origin/main`'s
   `dolt.AddDependencyWithOptions` stages the same `{dependencies, events}`
   pair. It is tracked as `bd-2y9ke` with a reproduction, and the code comment
   now says so instead of claiming the staged set is complete. Fixing it changes
   what every dep-add commit contains and needs its own change with its own
   tests. `TestEmbeddedDepAddWispSource` does not cover it — all its sources are
   wisps, whose `is_blocked` lives in the dolt-ignored `wisps` table.

8. **The conformance package covers two of the seven roles.** Only
   `dependency_editor_contract.go` and `issue_operations_contract.go` exist.
   `ReadyClaimer`, `BatchCloser`, `Commenter`, `Relations` and `Reader` are
   tested per implementation, not against a shared contract, so nothing
   structurally prevents the three implementations from drifting on those five
   the way the direct and proxied routes drifted before this branch.

9. **The clean-working-set assertion is embedded-Dolt only.** The proxied
   route's equivalent (`bd link` / `bd dep add` over `dolt sql-server`) is
   covered at the role and conformance level but has no front-door twin; that
   would need a live sql-server fixture.

10. **`TestEmbeddedDepAddWispSource` asserts a clean pre-add baseline and
    fatals if it is already dirty.** That makes it sensitive to any future
    change that leaves the working set dirty during `bd init` or `bd create` —
    it would fail with "fixture setup left a dirty working set" rather than a
    routing message. Deliberate (an unattributable assertion is worse than a
    loud one), but a coupling worth knowing about.

## Verification

Build, `go vet` and `gofmt` clean. `golangci-lint run --timeout=20m` reports
**0 issues** (it must be run with `GOTOOLCHAIN=go1.26.5`; the `.githooks/pre-commit`
gate pins the linter at a version built with go1.25 and dies before linting —
pre-existing and content-independent, someone should pin `GOTOOLCHAIN` in that
hook).

The protocol corpus passes under `BEADS_PROTOCOL_REQUIRE_DOLT=1` with
`cmd/bd/protocol/` byte-identical afterwards.

All **twelve** DependencyEditor conformance wirings pass — four contract
functions (wisp-sourced routing, mixed batch, mixed-batch rollback, cross-plane
cycle refusal) against each of the three implementations. That is the evidence
the verb collapse changed no behaviour.

`internal/storage/dolt` fails `TestDoltNew_RemoteMigrateGate_BlocksReopen`,
`TestRecomputeAllBlocked_AllowsDirtyWisps` and
`TestRecomputeAllBlocked_RefusesDirtyDependencies`. All three were reproduced on
a pristine `origin/main` worktree at `8b942915`, which additionally fails
`TestRecomputeAllBlocked_RefusesDirtyIssues` and
`TestRecomputeAllBlocked_RepairsAndCommitsWhenClean` — main is redder in that
family than this branch's filtered run showed. None of these is caused here.

---


julianknutsen:
## Pre-merge review record

This branch was reviewed adversarially before the PR was opened. Recording the
verdict and, more usefully, the things the review **did not** clear — so a human
reviewer knows where to spend attention rather than re-deriving it.

**Verdict: ship, no blockers.** Five findings, all minor, and the reviewer
established each one is pre-existing on `main` rather than introduced here:

1. **`is_blocked` staging gap.** `addDependencyEdgeInTx` stages the source's
   dependency table and its events table, but a blocking edge also flips
   `issues.is_blocked`, which is never added to `ChangedTables`. Reproduced live:
   the flip stays dirty and the next auto-committing command — *including a
   refused one* — carries it into history. `origin/main`'s
   `dolt.AddDependencyWithOptions` stages the same pair, so this predates the
   roles. Filed as **bd-2y9ke**. Not fixed here because changing what gets staged
   changes what every dep-add commit contains. The code comment previously
   *contradicted* this; it now names the gap.
2. **Missing-endpoint refusals still differ by route.** A nonexistent source on
   the direct route says `issue <id> not found`; through the unit of work the FK
   is the only guard, so it surfaces `fk_dep_issue`. The new conformance contract
   does not pin this, so "the three implementations answer the same refusals"
   holds only for the refusals the contract covers.
3. **Cycle-refusal message texts differ by route.** `errors.Is(ErrDependencyCycle)`
   parity holds; byte parity does not.
4. **The uow provider commits with `DOLT_COMMIT('-Am', …)`**, staging every dirty
   tracked table — so GH#2455 selective staging exists only on the store-backed
   routes. Pre-existing property of the provider, applies to every role.
5. **Dead pinned bulk variants** — since resolved: `AddDependencies`,
   `AddWispDependencies` and `AddDependenciesSourceRouted` collapsed into one
   source-routed verb after the review.

### What the review verified by running, not reading

- All twelve `DependencyEditor` conformance wirings (four contract functions ×
  three implementations), re-run independently rather than taken from the
  implementer's report.
- A mixed-batch refusal rolling back **both** planes, on all three implementations.
- Cross-plane cycle refusal (bd-xe27's `permanent A → wisp B → permanent A`) at
  the front door, with zero rows left behind.
- Wisp-only add records no commit and leaves the tree clean — the positive
  assertion both implementer reports had admitted was missing.
- Protocol corpus byte-identical, no fixture touched.

### Two premises in the brief the agents refuted

Recording these because they were wrong in the *instructions*, not in the code,
and a reviewer working from the same assumptions would be misled the same way.

- The pinning comment claimed the issues plane was "the only one its unit-of-work
  sibling can reach." **False.** `domain.DependencyUseCase` declared a full set of
  wisp twins with no production caller. Refuted twice: by reading the chain down
  to `pickDepTable()`, and by reverting the call site and watching the FK fire.
- A claim that `internal/storage/conformance` had no role-level entries. **False** —
  `issue_operations_contract.go` is a `publicops.Lifecycle` contract already wired
  in all three implementations. The correct statement is that **two of seven** roles
  have a shared contract; the other five do not, which is gap #8 in the description
  and the structural reason the wisp narrowing went unnoticed in the first place.

### Known-red, not caused here

`internal/storage/dolt` fails `TestDoltNew_RemoteMigrateGate_BlocksReopen` and
the `TestRecomputeAllBlocked` family. Reproduced on a pristine `origin/main`
worktree at `8b942915`, which additionally fails
`TestRecomputeAllBlocked_RefusesDirtyIssues` and
`…_RepairsAndCommitsWhenClean`. Main is redder in that family than this branch's
filtered run showed. Flagging because `TestRecomputeAllBlocked_*` touches the same
blocked-recompute subsystem this branch modifies — the merge did not cause these,
but an already-red test cannot prove it did not hide something.

julianknutsen:
## Role semantic-conformance suite

All **seven** roles now have a shared semantic contract that every implementation runs. Before this, two did — and one of those (`DependencyEditor`) only got one last week, as a by-product of fixing a regression the absence of a contract had allowed.

| contract | cases |
|---|---|
| `dependency_editor_contract.go` | 18 |
| `reader_contract.go` | 15 |
| `batch_closer_contract.go` | 13 |
| `commenter_contract.go` | 10 |
| `relations_contract.go` | 9 |
| `ready_claimer_contract.go` | 8 |
| `issue_operations_contract.go` (pre-existing) | 6 |

22 wiring files across `dolt`, `embeddeddolt` and `uow`. The 804 duplicated lines in the per-backend role tests are collapsed; the only survivor is `dolt/dependency_editor_test.go`, trimmed to its Dolt-specific staging assertions.

### What this does and does not establish

Each case is written to what the **leaf doc comment promises**, not to what any implementation currently does. That distinction is the point: three divergences and six spec gaps were found precisely because the contract asserts the doc and something failed.

The honest limit, stated here because the suite states it everywhere internally too: for five roles `dolt` and `embeddeddolt` **share one validate/execute body**, and for `Reader` they share the entire implementation (both accessors return `storereader.New(store)`). Only `uow` is separate code. So a green run on three legs is not three independent votes — the third leg proves the transaction wrapper and the engine, not the body. A bug in the shared body plus a matching bug in uow passes everything.

### Divergences — parked, not fixed, awaiting adjudication

Each is a `skipKnownDivergence` at the **wiring site**, so the case still runs and passes on the backends that agree. Seven park sites, three beads, reconciled in both directions.

- **`bd-ocrn7`** *(P1, uow only)* — `dependencyeditor.go:136-139` promises no blanket unknown-target refusal, because a target may be an `external:` ref or an issue in another repo. dolt and embeddeddolt route a foreign-repo target into `depends_on_external` (no FK) and accept it; uow refuses with `fk_dep_issue_target`. Notably, that FK was **added and later dropped** in this codebase's own migration history.
- **`bd-yby99.4`** *(P1, all three)* — `readyclaimer.go:61` says "the ready set is the issues plane," and all three backends claim a ready wisp anyway. They then disagree about *how*: dolt mutates the wisp row and returns a verify-after-write error with `Claimed = nil`; embeddeddolt returns the claimed wisp with `err = nil`.
- **`bd-yby99.3`** *(P2, all three)* — `relations.go:38-45` requires a type filter to be "within the column's length." The column is `VARCHAR(32)`; the only length gate is `IsValid`, which accepts 1..50. So a 33–50 char type is accepted as a filter that no edge can ever match — exactly the "filter that quietly matches nothing" the clause forbids. Verified by running it: a 40-char type cannot be stored (`Error 1105 ... too large for column 'type'`) but is happily accepted as a filter.

### Spec gaps — the doc is silent, so no promise was invented

Each leaves a greppable `SPEC-GAP` marker at the case site so it cannot be confused with a case that was quietly dropped.

`bd-yby99.1` uow has no event-free dependency insert · `bd-yby99.2` `relations.go` is the only role file with no request-snapshot promise · `bd-yby99.5` silent on whether an ephemeral comment may record durable history · `bd-yby99.6` never says `AddCommentRequest.IssueID` must be non-empty · `bd-yby99.7` `Offset` is "where the backend supports it" and is a no-op on the store body while uow honours it · `bd-yby99.8` "ReadyFlag under the list vocabulary" overpromises — the projection silently drops half the filter

### Verification

Three legs run and counted separately, because a leg that silently skips would prove two-way parity while reading as three:

```
dolt          77 cases, 0 fail, 2 skip   42.6s
embeddeddolt  77 cases, 0 fail, 2 skip   27.4s   (BEADS_TEST_EMBEDDED_DOLT=1)
uow           88 cases, 0 fail, 3 skip   63.4s
```

Skips are exactly the parked divergences above. `golangci-lint` 0 issues; protocol corpus byte-identical.

**Known gap in enforcement:** most CI shards set `BEADS_TEST_SKIP=dolt`, and the embedded leg is double-gated on `//go:build cgo` and `BEADS_TEST_EMBEDDED_DOLT=1`. A green CI is therefore not by itself proof that all three legs ran. That predates this change but now matters more, since three-way parity is the suite's entire claim.

maphew (commented):
The honest engineering in this PR is what stands out: the "Coverage" comment admitting the three-leg conformance green proves the shared body plus uow rather than three independents, the CHANGELOG documenting every user-visible delta (one Dolt commit per batch close, `--claim-next` inside the close transaction, unified commit-message spellings), and the known-gaps list that matches what we actually found. The structural core is right — seven role interfaces reached only through store accessors so hook/telemetry decorators compose per-role, ~4,200 lines of shared conformance contracts run against all three backends, and the depguard rule closing the second front door. And it fixes a real regression along the way (wisp-sourced `bd dep add` routing per-edge by source plane). A 128-file rewrite of every issue verb's plumbing gets adversarial multi-model review by policy here; it came through with two findings, both fixable inside the design you built.

Local verification in a detached worktree at ed49e07: build clean; `issueops`, `internal/storage/issueops`, telemetry, httpapi, and root-package suites pass; the uow conformance leg (126s, 88 cases) and the embedded conformance leg (115s, actually executed, not skipped) both pass; the front-door wisp regression pin `TestEmbeddedDepAddWispSource` passes. The one red check ("Test (Server Dolt Conformance)") is a docker-registry pull timeout before any test ran — needs a rerun, and worth insisting on since that job is the only CI leg exercising the sql-server-backed wirings.

Two should-fixes before merge, both verified by tracing head against merge-base 8b94291 (not taken on faith from any one reviewer):

1. **The proxied close fence is no longer re-validated inside the write transaction.** Base ran `validateIssueClosable` and `CloseIssueChecked` in one retried unit of work (old close_proxied_server.go:79/:223/:258, with `uow.RunTxResult`'s backoff replaying the *whole* func on serialization conflict — and the `row_lock` rewrite on every update/claim made that retry genuinely reachable). Head splits it: `closeProxiedRunPreflight` reads and fences in a closed read-tx, then `closer.CloseBatch` opens a second tx whose `closeBatchItem` checks version/children/blockers but never template/pinned/assignee — and `BatchCloseItem` has no version field, so `ExpectedVersion` is always nil on this path. A claim/pin/type change landing between the two transactions (or during the batch tx, whose replay re-runs only `closeBatchItem`) closes an issue the fence would now refuse, with `✓ Closed` printed. To be fair: the *direct* route already had this exact gap, so this levels proxied down rather than inventing a hole — but the clean fix is cheap inside your design: add `ExpectedVersion *int64` to `issueops.BatchCloseItem`, fill it from `current.RowVersion` in both preflights, and the engine's existing CAS (close.go:56-66) does the rest, since `row_lock` already tracks exactly the status/assignee writes the fence cares about.

2. **Both batch-close bodies convert infrastructure errors into per-item refusals, against the `BatchCloser` contract's own words** ("a non-nil error and a populated Outcomes are mutually exclusive" — batchcloser.go:92-97). `ExecuteClose` can fail *after* its write: the close UPDATE/lease-delete/event/`is_blocked` staging is savepoint-protected, but `HydrateIssueOperationResult` runs outside the savepoint and can fail on a bad metadata scan or a read hiccup. In `close_batch.go:89-92` (and the same shape in `uow/batch_closer.go:98-110`) that becomes a per-item `Err` and the loop continues — a later survivor's `tables.Merge` then commits the failed item's staged writes; with *no* survivor, the empty-tables path skips the Dolt commit but `tx.Commit()` still runs, landing the close in the working set with no history entry. Either way the caller is told the item errored when it actually closed. None of the 13 conformance slices exercises an infra failure, so the contract sentence is currently unenforced prose. Fix in both bodies: classify — only `ErrNotFound`/`ErrCloseBlocked`/`*CloseOpenChildrenError`/`ErrValidation` become outcomes, everything else aborts the batch; or widen the savepoint to cover hydration, which fixes the partial-write half without an error taxonomy.

For the record, a third cross-vendor finding was checked and de-escalated: the comment path's template check moving out of the insert transaction is *pre-existing-equivalent* — base's single tx read `is_template` from its snapshot while the insert wrote a different table, and Dolt's cell-level merge meant a concurrent flip never conflicted at base either; the direct route is byte-identical at base and head. No action.

One sequencing question, yours to call: this rewrites `cmd/bd/close.go`, `update.go`, `issueops/claim.go`, and `storage.go` — every file your open CAS/ownership series (#4682/#4697/#4715) targets. Whichever side lands second is real rework, not a mechanical rebase. What order do you want these judged in?

Noted as acknowledged known-gaps, no action needed: `bd dep add --file` losing per-line attribution for storage-level refusals (a Result-side edge index could restore it later), and multi-store `--claim-next` riding only the first id's store. The 4730 interplay was checked: zero file overlap, no merge-tree conflict, semantically orthogonal — either order.

Review provenance: dual-vendor pass — Codex gpt-5.6-sol (high) transaction-boundary trace, Claude opus-tier conformance verification, independent adjudication of the disagreement by tracing head against base — plus maintainer synthesis.

julianknutsen:
## Behaviour unification: the three adjudicated divergences

The conformance suite found three places where the implementations disagreed with the leaf doc comments. All three are now decided and applied. **Every `skipKnownDivergence` park is gone** — the contract legs run 0 skips each.

### 1. External and foreign-repo dependency targets never hit a foreign key — `bd-ocrn7`

`dependencyeditor.go:136-139` promises no blanket unknown-target refusal, because a target may be an `external:` ref or an issue in another repo. dolt and embeddeddolt honoured that; uow refused with `fk_dep_issue_target` because `pickDepTargetColumn` knew only the literal `"external:"` prefix and classified a foreign-repo id as *local*.

Fixed by introducing **one** named rule — `issueops.IsExternalDepTarget` — and routing both classification chokepoints through it, rather than teaching the domain route a second copy of the predicate. `fk_dep_issue_target` is untouched; a genuinely local, genuinely missing target still refuses (pinned by the pre-existing `MissingTargetIssueFailsFK`).

**One consequence beyond the failing case, stated because it is real:** a cross-prefix target that *does* exist locally now routes to `depends_on_external` on the domain route, so derived blocking stops reading that edge on uow. That is exactly what the store backends have always produced — they classify by prefix regardless of local existence — so this is the adjudication applied consistently, not a new behaviour. Nobody has measured how often such rows exist in practice.

### 2. Claiming a ready ephemeral row is legal — `bd-yby99.4`

All three backends already claimed ready wisps. They disagreed only about what they told the caller: embeddeddolt and uow returned the claimed row with `err = nil`; dolt mutated the row identically and *then* returned `"reported success but could not be verified"`, because its verify-after-write re-read the id in the issues plane and found nothing.

Standardized on embeddeddolt. The verify wrapper is now **plane-aware, not weaker** — a genuinely unverifiable claim still errors on both planes, and the probe fails safe (a failed plane probe degrades to the old honest error, never to a false success).

`readyclaimer.go`'s *"the ready set is the issues plane"* was simply false and has been replaced with what a caller can actually rely on, including two consequences that would otherwise ambush people:

- an ephemeral claim records **no durable history entry** — wisps are not versioned, so it never appears in `bd dolt log`;
- it also carries **no lease** — heartbeats refuse the row and lease-expiry recovery only walks leased durable rows, so a claimant that dies strands it in-progress with no expiry to wait out.

That second one is the sharper problem and is filed as **`bd-eufck`** rather than papered over.

### 3. Dependency types bounded at 32, not 50 — `bd-yby99.3`

`relations.go` promised a type filter is checked "within the column's length." The column is `VARCHAR(32)`; the validator accepted 1..50. A 33–50 char type was therefore accepted as a filter no edge could ever match — the exact "filter that quietly matches nothing" the clause forbids. Verified by running it: a 40-char type could not be stored but was happily accepted as a filter.

**Breaking**, and in the CHANGELOG as such: `bd` now refuses such a type on twelve paths instead of accepting it and failing at write time or matching nothing at read time.

### Verification

```
dolt          82 cases, 0 fail, 0 skip
embeddeddolt  77 cases, 0 fail, 0 skip
uow           79 cases, 0 fail, 0 skip
```

`golangci-lint` 0 issues. Protocol corpus **unchanged** — with the caveat the integration pass insisted on stating: the corpus covers *neither* user-visible change, so "the corpus did not move" is the true claim, not "the corpus confirms these are invisible."

The six **spec gaps** (`bd-yby99.1/.2/.5/.6/.7/.8`) were not adjudicated and are untouched, markers included.

### Follow-ups filed rather than folded in

`bd-eufck` a claimed wisp is unrecoverable coordination state · `bd-j8g7y` `bd ready --claim` does not say the row was ephemeral · `bd-3k20t` collapse the surviving second spellings of the external-target rule · `bd-1vh7i` custom statuses and issue types have *no* length gate against the same `VARCHAR(32)` — the same defect one layer over · `bd-51gt8` the create form warns-and-drops an invalid dependency

julianknutsen:
## Spec gaps: eight of nine findings resolved

The conformance suite produced nine findings — three divergences (implementations disagreed) and six spec gaps (the leaf doc was silent or ambiguous, so no assertion was written and a greppable `SPEC-GAP` marker was left in its place). All three divergences and five of six gaps are now resolved. **Zero `skipKnownDivergence` parks remain; one `SPEC-GAP` marker remains, and it maps to the one bead still open.**

### `bd-yby99.8` — `--ready` now refuses a filter it cannot carry

`reader.go` claimed `ReadyFlag` applied the rest of the request "under the list vocabulary." It largely did not: the `--ready` arm reaches its query through a deliberately lossy projection whose caveat lived in an internal package no caller sees.

This was reachable from the CLI and answered wrong. Reproduced on a scratch workspace with two open issues, before and after:

```
BEFORE  bd list --ready --id tst-d68 --json   → both ids, exit 0   (id set ignored)
AFTER   bd list --ready --id tst-d68 --json   → exit 1
        Error: validation failed: --ready cannot filter on IDFilter (--id); the
        blocker-aware ready query carries only part of the list vocabulary...
```

The refusal names **every** field it could not honour, by both struct field and CLI flag, and lives in one shared validator reached by all three implementations and all three `bd list` routes — no per-backend copies.

Three corrections to the drop set, established by reading `BuildListFilter` against `ReadyFilterFromIssueFilter` field by field rather than trusting the bead:

- **`Statuses` is unreachable**, not dropped — `if in.ReadyFlag { Status = open }` wins over the status filter, so nothing is ever there to drop. Corroborated by the existing `ready_flag_overrides_status` golden, untouched.
- **`NoPinnedFlag` is honoured** despite `Pinned` being dropped, because the ready clause hardcodes `(pinned = 0 OR pinned IS NULL)`. Only its opposite, `PinnedFlag`, is refused. Refusing it would have broken a working command.
- **`SkipLabels` is a hydration knob, not a filter** — the answer is correct, only the query costs more — so it is documented as ignored rather than refused.

### `bd-yby99.7` — `Offset` is honoured or refused, never silently ignored

Measured: `Offset` was a **no-op** on the store body (dolt and embeddeddolt share the entire Reader implementation) and honoured on uow. The doc hedged with *"where the backend supports it"*, naming no backend and offering no way to find out.

The CLI already knew about the split — all three `--offset` flags say *"Only supported under --proxied-server"* and the direct routes hard-error. The **role** was the layer that lied, which matters for `bd serve` and anything else built on the library rather than on `cmd/bd`.

`storereader` now refuses with a typed `*ErrUnsupported` naming operation and backend. To make that classifiable by a caller holding only the role interface, `ErrUnsupported` moved into the leaf `issueops` package with a type alias back from `internal/storage` — the same treatment the other error symbols got. Every existing `errors.As` site keeps working unchanged, and `go list -deps ./issueops` confirms the leaf is still a leaf.

The contract case pins the invariant that is genuinely common rather than one implementation's behaviour:

> A non-zero `Offset` is either **honoured** or **refused with a typed error** — never silently ignored.

Same request at offset 0 and 1, under a total order so "different" cannot be storage-order wobble. Same rows plus no error is a failure. Passes on all three legs with nothing parked — dolt/embeddeddolt take the refused branch, uow the honoured branch.

### Three doc tightenings where the implementations already agreed

- **`bd-yby99.5`** — "at most one history entry" → **none**, for a comment on an ephemeral row. Federation correctness depends on zero, since the wisp tables are dolt-ignored precisely so ephemeral work never ships. The weak bound meant the exact regression that breaks federation passed everywhere.
- **`bd-yby99.6`** — an empty `IssueID` is `ErrValidation`; the doc's `ErrNotFound` clause narrowed to a non-empty id. The case tightens from "either typed error" to the specific one.
- **`bd-yby99.2`** — `Relations` was the only one of seven role files with no never-mutate-the-caller's-request promise. Added, with the tripwire case.

### Still open

**`bd-yby99.1`** — uow cannot insert a dependency without emitting an event, while the store backends' *default* is no-event for structural callers like create-with-deps. So the same command can produce different event streams by route. Sent for measurement as **`bd-soezd`** rather than decided from the code: "always emit" costs an event per edge on `create --graph`, which is exactly why the no-event default exists, and nobody has that number yet.

### Verification

81 contract cases per leg, **0 skips**, all three legs. `golangci-lint` 0 issues. Protocol corpus byte-identical — and it could not have moved: it invokes neither `bd list --ready` nor `--offset` anywhere, which is itself worth knowing.

julianknutsen:
# Council review: PR #5293 — one library surface for every issue capability

Five-lens review (claim-rebase, conformance-truth, behaviour-changes, boundaries,
test-quality) plus an independent synthesis pass that re-verified every blocker
and major by re-reading the code and re-running the failing tests. All five
lenses completed; none failed.

## What the PR does

Beads had two implementations of most issue verbs — the direct route (CLI opens
its own store) and the proxied-server route (CLI through a unit of work) — and
they drifted. This PR puts every issue capability behind role interfaces in the
leaf package `issueops` (imports `internal/types` + stdlib only), reached via
capability accessors (`store.Commenter()`) so the hook and telemetry decorators
layer uniformly. Three implementations answer every role: dolt, embeddeddolt,
and uow. Each role gets a semantic-conformance contract in `backend/conformance/`
(~129 cases per leg) written against the leaf doc comment, not against any
implementation's behaviour. Divergences that discipline surfaced were parked by
bead, adjudicated, and fixed in-PR.

## What the council checked and found clean

- **Leaf purity**: `go list -deps ./issueops` resolves to `internal/types` +
  stdlib only. Verified by running.
- **Conformance arithmetic**: 129 cases wired at dolt and embeddeddolt, 127 at
  uow (the 2 staging cases assert Dolt working-set state uow cannot observe —
  disclosed). The full uow leg was executed independently: 127 cases, zero
  skips, zero failures. Zero `skipKnownDivergence` call sites remain.
- **Anti-vacuity discipline**: positive-control deltas through the same query,
  loud skips for unobservable planes, ErrNotFound-vs-empty-page distinctions,
  self-testing fixture kits. Spot-checks found no vacuous case.
- **The shared-body caveat is stated honestly everywhere**: CHANGELOG, contract
  file headers, code comments. No "three independent votes" overclaim was found
  anywhere in the diff.
- **Decorator layering**: hook and telemetry decorators explicitly override and
  recurse both claim accessors; no silent embedding bypass.
- **Claim sentinel identity**: `storage.ErrAlreadyClaimed`/`ErrNotClaimable`
  alias the issueops leaf sentinels; `errors.Is` matches through every wrap;
  producer-to-parser round trip passes (`TestClaimConflictFormatRoundTrip`).
- Lint (golangci 2.10.1), build with sourced `.buildflags`, httpapi boundary
  exceptions, dead-code sweep: clean.

## Must fix before merge

The predicted hazard — the end-of-session rebase resolution of the two rival
claim roles — is exactly where the damage is. Everything below was re-verified
by the synthesis pass.

1. **BLOCKER — the rebase silently deleted the open-but-assigned claim refusal
   copy from BOTH producers** (`internal/storage/issueops/claim.go:152`,
   `internal/storage/domain/issue.go:570`). On main, claiming an OPEN issue
   assigned to someone else refuses with the owner-adjudicated anti-steamroller
   guidance (`already assigned to %q — coordinate with the holder; ... bd
   reclaim`, wy-yuclk/bd-at6rc). On this branch it collapses to the generic
   `issue already claimed by <assignee>`. Verified by diffing both producers
   against `origin/main` and by running the pinning tests:
   - `go test ./internal/storage/domain/db -run TestDomainDB/TestIssueUseCase_Claim`
     FAILS ungated (`OpenAssignedRefusalSteersToHolder`).
   - `BEADS_TEST_EMBEDDED_DOLT=1 go test ./cmd/bd -run 'TestEmbeddedUpdate$/update_claim_already_claimed'`
     FAILS (lands in the push-to-main shard, so it would break main after a
     green PR run). `cmd/bd/claim_pools_embedded_test.go:79,94` and
     `update_embedded_test.go:485` pin the same copy.
   The producer comments are also now self-contradictory ("the copy lives on
   the typed error" vs "the prose is composed HERE") — the copy currently lives
   nowhere. Fix: restore the `StatusOpen`-and-assigned branch in both
   producers, inside the `ClaimConflictError` wrapper (its passthrough
   `Error()` carries the prose byte-for-byte), and fix the comments. If the
   short copy is actually intended, it needs owner adjudication plus updates to
   four test files and the bd-at6rc CHANGELOG entry it contradicts.

2. **BLOCKER — the PR's own new proxied-claim test is red.**
   `cmd/bd/update_claim_proxied_test.go:135` builds its fixture with the bare
   sentinel but asserts `issue already claimed by bob` — written against the
   composing `ClaimConflictError.Error()` that LOST the rebase. Verified by
   running: `go test ./cmd/bd -run TestProxiedClaimConflictStaysAPerIDFailure`
   fails ungated. Fix: compose the fixture's `Err` the way both real producers
   now do (`fmt.Errorf("%w%s%s", storage.ErrAlreadyClaimed,
   storage.ClaimedByFragment, "bob")`).

3. **MAJOR — the headline CHANGELOG entry ships three false claims** (all
   verified by reading):
   - Line ~173: the suite lives in `internal/storage/conformance` — that
     directory does not exist; #5321 (documented 150 lines earlier in the same
     file) moved it to `backend/conformance`.
   - "All seven role interfaces in `issueops`" — post-rebase there are eight
     (main's `Claimer`, #5334, has no suite).
   - "Five findings remain open" says the empty-label divergence
     (bd-yby99.28/.29) is "deliberately unstated because the backends
     disagree" — but both beads are CLOSED, the leaf doc now states the DROP
     promise (`issueops/issueops.go:66-73`), and the same entry documents the
     fix two paragraphs later. One editing pass fixes all three.

4. **MAJOR — the proxied and direct close routes answer the open-children
   refusal differently** — the exact drift class this PR exists to kill, in a
   verb it rewrote. `closeProxiedRefusal` (`cmd/bd/close_proxied_server.go:335`)
   has no `ErrCloseOpenChildren` arm, so the typed error this PR made newly
   reachable through `CloseIssueChecked` prints double-prefixed, while
   `closeDirectRefusal` prints it bare. `closeProxiedCheckOne` (:271) also
   keeps an epic-only CLI pre-check saying `cannot close epic %s:` where the
   direct route deleted its pre-check and surfaces the engine's `cannot close
   %s:`. Verified by reading both formatters and tracing the typed error
   through uow's `BatchCloser` → `CloseIssueChecked` →
   `issueRepo.CloseChecked`. Fix: add the arm, delete the pre-check. The
   conformance suite cannot see this — it lives in CLI formatting above the
   roles.

5. **MINOR (but a false claim; fix with #1/#2)** —
   `internal/storage/storage.go:56`, a comment ADDED by this PR, says "The
   producer is now issueops.ClaimConflictError.Error(), which composes the
   message" (it is a passthrough; the producers compose) and names
   `TestClaimConflictMessagesKeepTheParseableShape` as the tripwire — a test
   that exists nowhere in the tree. The real tripwires are
   `TestClaimConflictFormatRoundTrip` and the root-package parser tests.

6. **MINOR (fix when merging)** — the squash commit message says "the
   conformance package covers two of the seven roles" under Known gaps; the
   CHANGELOG and `backend/conformance/` say all seven. The message describes an
   earlier state of the session; amend it at merge.

## Deferred (file as beads, do not block)

- `roleAccessorNames` in both decorator-roster tests hardcodes seven roles and
  promises an eighth role cannot arrive silently — main's `IssueClaimer` (the
  eighth) arrived silently in the rebase. Mitigated by #5334's behavioral
  decorator tests. Two-line fix.
- `aggregate.go` label Remove loop does not skip `""` (Add/Replace do), so
  `Remove:[""]` deletes a legacy empty-label row — contradicting the leaf doc
  clause two lines above it and diverging from uow. Reachable only for legacy
  rows on an explicit empty Remove entry.
- Conformance citations into `issueops.go` in the claim/lifecycle region are
  stale by ~6 lines post-rebase (`issue_operations_contract.go:758-760`);
  citations into other leaf files spot-checked exact.
- The claim-refusal chooser is duplicated across the two producers
  (acknowledged with "must stay in step" comments, verified semantically
  identical today); a shared composer would leave one producer. Related:
  `bd update --claim` is a third doorway into the same two bodies.
- Stale forbidigo exception for the two close files (0 filter mentions remain).
- `cmd/bd/dep_bulk_role_test.go:34` duplicates a conformance case verbatim at
  the same seam; its two sibling tests genuinely earn their place.
- `ExpectedVersion`/`IssuePlaneOnly` request fields have zero production
  setters; two newly exported funcs (`HydrateReadyRowInTx`,
  `RequireIssueOrWispInTx`) have no external callers.
- `RunIssueOperationsCreateRejectsMissingDependencyTargets` (pre-existing from
  #5321) asserts a refusal identity no leaf doc promises — same shape as
  bd-yby99.9.
- bd-m8do7's leftover per-backend strict-subset tests: deferrable as filed; its
  bd-yby99.26 sequencing hostage is closed, so the hold clause can drop.

## What a green CI still does not prove

- **Three legs are not three votes.** Dolt and embeddeddolt share one
  validate/execute body for six of seven roles and the ENTIRE Reader
  implementation. A green three-leg run is two readings of the contract plus a
  wrapper-and-engine check. Only uow is independently written code. The PR
  states this honestly; a reviewer should still weight it.
- **PR-green does not mean main-green.** The claim-copy tests that fail on this
  branch live in the embedded-cmd shards that run mostly on push-to-main, and
  the `Main` workflow is perpetually concurrency-cancelled — merging a
  PR-green branch can still break main invisibly.
- **The protocol corpus cannot see prose or transactions.** Its canonicalizer
  masks timestamps and stable-sorts arrays; "corpus byte-identical" means the
  pinned JSON wire shapes did not move — refusal copy, Dolt commit messages,
  and transaction boundaries are invisible to it by construction.
- **CLI formatting above the roles is outside the suite.** Route-divergent
  refusal text (finding 4) is invisible to all 129 cases; nothing diffs the
  direct and proxied formatters against each other.
- **Where the leaf doc is silent, implementations can drift green.** The
  dependency-existence refusal identity (bd-yby99.9) is an admitted silence; a
  case may never assert more than the doc promises, so drift there trips
  nothing.
- **Nothing diffs the two claim-refusal producers against each other.** Each
  leg is checked against the doc; the producers are kept in step by comments
  and a round-trip test on one of them.

## Verdict: fix-first

The architecture is sound and the conformance discipline is the strongest test
work in the repo — the synthesis pass independently reproduced the case counts,
ran a full leg, and found the honesty claims accurate. But the branch has two
ungated red tests plus two gated ones, all downstream of the rebase's claim
resolution silently deleting owner-adjudicated refusal copy, and the shipped
CHANGELOG contradicts itself and the code. Items 1-5 are a few hours of
contained work; nothing requires rework.

---

## Addendum: what was fixed after the council reported

All six must-fix items were applied in `7932549fb`. Two were blockers, both
introduced by the rebase resolution rather than by the PR's own work.

**The wy-yuclk refusal copy was restored to both claim producers.** The rebase
rewrote `internal/storage/issueops/claim.go` and
`internal/storage/domain/issue.go` to compose refusal prose (necessary, because
main's `ClaimConflictError.Error()` is a passthrough) but collapsed main's
three-way branch into two — dropping the pool short-circuit ordering and the
open-but-assigned copy that steers a caller to *coordinate with the holder*
rather than naming a release command. That copy exists because batch agents
pattern-match an eviction command into an unclaim+claim steamroller of live
claims. Four assertions across three files pinned it and were failing; all pass
now. The `" by <assignee>"` tail stays deliberately absent from that one
refusal, so the holder is recovered from the typed field rather than the prose.

**The PR's own proxied-claim test was written against the losing design** — its
fixture wrapped a bare sentinel and asserted composed prose. It now composes the
way both real producers do, which also makes it exercise the passthrough
contract honestly.

Also fixed: the two close routes now answer the open-children refusal
identically (the proxied route gained the `ErrCloseOpenChildren` arm and lost an
epic-only pre-check the direct route had already deleted — that pre-check
answered differently for a non-epic parent, which is the exact drift class this
PR exists to remove); a stale comment in `internal/storage/storage.go` that
named a tripwire test which does not exist; and three false claims in the
CHANGELOG headline entry (the suite path, "all seven roles" where eight now
exist, and an open-findings paragraph that described a closed divergence as
unresolved).

Nine lower-severity findings were filed as beads rather than fixed here.

Issue #5233, opened 2026-08-01
Relative --defer and dep-record created_at serialize local wall-clock with a Z suffix (bd 1.1.2)


zigrivers:
bd 1.1.2 writes relative timestamps as local wall-clock time stamped `Z`, so west of UTC they land in the past.

## Reproduction

Isolated temp DB, machine at UTC-6 (MDT):

```
now = 2026-08-01T03:40:15Z
bd update <id> --defer +1h
-> defer_until = 2026-07-31T22:40:16Z    # (now + 1h) minus the 6h offset
```

An absolute `--until` value round-trips correctly, so only the relative path is affected.

## Impact

A relative deferral is a silent no-op west of UTC — the issue reappears in `bd ready` immediately. Agent workflows that use `--defer +1h` as a claim cooldown therefore busy-loop on claim -> reject -> release.

## Second surface: dependency records

`bd dep add` writes `created_at` internally with no timestamp flag, so it cannot be corrected or avoided through the public API. In one tracker, 799 of 1067 dependency records sit exactly +6.00h from their issue `created_at`.

## Ask

Normalize relative offsets to UTC before serialization — both the deferral path and the internal dep-record `created_at`.

## Environment

- bd 1.1.2 (Homebrew)
- macOS, `TZ=America/Denver` (UTC-6)

zigrivers:
Follow-up data point from the same tracker: as of today 127 of 129 deferred issues sit past their defer_until (was 92/135 at filing) — the count keeps growing because of the companion lifecycle defect now filed separately as #5289 (a deferred issue never returns to bd ready even after the timer passes, so every past-stamped deferral parks work permanently).

cosentinode:
I split this into #5295 for relative timestamp normalization and #5297 for dependency timestamp generation, was able to reproduce.

PR #5295, opened 2026-08-02
fix: normalize parsed timestamps to UTC

cosentinode:
- interpret timezone-less relative inputs in the caller's local timezone
- normalize every successful `ParseRelativeTime` result to UTC before persistence or serialization
- preserve local wall-clock formatting in past-defer warnings
- cover compact, absolute, timezone-less, and natural-language inputs under UTC-6

Relative deferrals currently retain local wall-clock location and can be persisted as if they were already UTC. This is PR 1 of 2 for #5233; the follow-up storage PR will normalize internally generated dependency timestamps.

## testing

- `./scripts/test.sh -run '^TestParseRelativeTime' ./internal/timeparsing/...`
- `./scripts/test.sh ./internal/timeparsing/...`
- `./scripts/test.sh -run '^TestDefer' ./cmd/bd/protocol/...`
- `make test`
- `golangci-lint run ./internal/timeparsing/... ./cmd/bd/...`

Part of #5233


maphew (approved):
Root cause independently confirmed, and fixing it in the parser was the right layer: on the embedded path `dolthub/driver/v2` binds through vitess `BuildBindVariable`, which formats with `v.Format("2006-01-02 15:04:05.000000")` — no `.UTC()`, no offset — so an `America/Denver` `time.Time` is stored as local wall clock and read back as UTC. That's exactly the reported `+1h` landing 5h in the past. And crucially, `PrepareIssueForInsert` normalizes only CreatedAt/UpdatedAt and is not on the `bd update --defer` path at all, so a storage-layer-only fix would have left the reported symptom unfixed. Adding `.Local()` to the past-defer warning at all five call sites is the compensating half most patches of this shape forget — store UTC, print local is the correct split. The new test is hermetic (`time.FixedZone` rather than the runner's TZ) and covers all four parse layers, not just the reported case. Nice work, especially for a first contribution here.

Verified in a detached worktree at c8545a5: build clean; timeparsing, query, and the defer/due `cmd/bd` tests pass, including under `TZ=America/Denver`.

One thing to finish, either here or as an immediate follow-up: `cmd/bd/show_format.go:118` and `:121` format DueAt/DeferUntil without `.Local()`, and scanned values come back UTC. So after this change, `--due 2026-08-01` at UTC+10 stores `2026-07-31T14:00Z` and `bd show` renders "2026-07-31" — a day earlier than the user typed. West of UTC (the reporter's zone) is unaffected; an east-of-UTC user will file the mirror-image bug. It's the same `.Local()` treatment you already applied to the warnings.

Two small notes: the doc comment on `ParseRelativeTime` now mentions the UTC return but not the new anchoring rule (timezone-less/date-only inputs anchor to `now`'s location rather than `time.Local` — behavior-preserving today since every caller passes `time.Now()`, but the contract changed; one sentence keeps the next caller honest). And the unit test pins the parser's return location but nothing pins the end-to-end symptom — a `cmd/bd` embedded test under a non-UTC TZ asserting `defer_until > UTC_TIMESTAMP()` after `--defer +1h` would be the durable guard; fine as follow-up.

Note for the queue: this is complementary to, not overlapping with, #4730 (different layer, zero file overlap). Its stacked sibling #5297 is reviewed separately. CI hasn't actually executed yet (first-contribution approval gate) — local runs above are the current evidence.

Review provenance: delegated agent review (claude-opus, root-cause trace through vitess/gms) + maintainer adjudication.

cosentinode:
All three follow-ups are in (localized show output, parser UTC contract docs, embedded defer E2E) — ready to merge.

steveyegge:
**Fable adversarial review (bee, beads-lane steward) — VERDICT: MERGE once CI runs green; no code changes required.**

Refutation attempt failed on every substantive axis. Re-derived rather than trusted: the embedded vitess bind formats wall-clock without conversion (the actual bug) while the server DSN path always converted correctly, so the fix introduces no cross-path divergence; scan-back, JSONL export/import round-trip, and the untouched callers (query evaluator, list_format) are all instant-based or UTC-consistent; the `time.Local` → `now.Location()` anchoring change is behavior-preserving for every production caller. The three follow-ups maphew's approval asked for (show_format `.Local()`, doc-comment anchoring contract, end-to-end TZ test) are already in the current head.

**The one gate: CI has never executed** — all five workflows sit at action_required (first-contribution approval gate), which is why the PR shows UNSTABLE. I'm approving the workflow runs now; will merge when they're green.

Follow-ups to file separately, none blocking: (1) release note for legacy corrupted values — pre-fix embedded writes stored local-mislabeled-as-UTC, and the new local rendering shifts their display; (2) storage-boundary normalization of due_at/defer_until in `PrepareIssueForInsert` (externally-authored non-Z JSONL can still store shifted); (3) Created/Updated render UTC beside now-local Deferred/Due — same class, out of scope here.

**Merge-order note:** #5297 is stacked on this branch (its file list currently contains all 9 of this PR's files). #5295 merges first; #5297 must then rebase before evaluation — and since a CONFLICTING PR gets zero CI runs silently, its checks get re-verified post-rebase, not trusted from the stacked state.

steveyegge:
**Maintainer fix-push (bee):** the fresh-snapshot red was a genuine collision with a test that landed on main after this branch: `TestParseCountRequestCarriesEveryFilterFlag` (Counter-role series) pins the **Location identity** of parsed `--*-after/--*-before` bounds as `time.Local`. This PR preserves the instant that test's comment cares about (bare dates still resolve at the user's local midnight) but normalizes the representation to UTC — so `reflect.DeepEqual` fails on Location pointers even on a UTC runner. Pushed ea9bbcaf2 to this branch updating the expectation to construct the contract the way it now works (local midnight, then `.UTC()`), with the comment updated to say both halves. Verified green under both a UTC-offset and America/New_York TZ locally. Merging when CI is green.

PR #5310, opened 2026-08-03
fix(init): hold destructive preflight under replacement gate (bd-dswzy)

julianknutsen:
## What

Hold destructive embedded reinit preflight and database replacement under one all-at-once exclusive workspace/physical-root gate:

`AcquireAll` → count existing issues / confirm destroy intent → resolve operational path → replace database → release.

The new init-specific helper releases the complete gate set on preflight refusal or error, and returns the still-held handle on success so the caller retains it through replacement.

## Why

`countExistingIssues` previously opened and closed the embedded Dolt store before init acquired its replacement gate. A concurrent `bd init --reinit-local` process could acquire the gate, replace the same Dolt files, and delete the Noms manifest while that preflight reader was still alive. Main run [30777456349](https://github.com/gastownhall/beads/actions/runs/30777456349), job [91575854462](https://github.com/gastownhall/beads/actions/runs/30777456349/job/91575854462), caught the resulting Dolt close panic in `TestEmbeddedInitConcurrent`.

This is scheduler-sensitive production ordering, not a regression from #5306: that PR changed only test scenarios and left the concurrent test and init implementation untouched.

Tracks `bd-dswzy`.

## Safety and behavior preservation

- Remote-history safety still runs before gate acquisition.
- Gate planning uses side-effect-free `DoltDirPath`; the operational `ResolveDoltDir` and its existing shared-server fallback run only after confirmation, while the gate is held.
- The destroy token uses the exact raw, pre-fallback, pre-normalization prefix used by the former block.
- Existing warning text, refusal exit codes, and the historical handling of issue-count errors are unchanged.
- `--stealth` and repository/storage mutations begin only after successful preflight.
- No retry, sleep, panic suppression, timeout increase, or weakened concurrency assertion was added.

Open contributor PR #4415 overlaps `cmd/bd/init.go` as part of its much broader flat-file backend work, but does not address this ordering race. This fix does not replace, close, or alter its backend-selection intent or future `withDoltBackend` pins.

## Verification

TDD RED was established before implementation: the deterministic gate-lifetime test failed to compile because `acquireInitMutationGate` and its observable `OnWait` seam did not exist.

Passed after implementation:

```text
./scripts/test.sh -run '^(TestAcquireInitMutationGateKeepsReplacementExclusiveDuringPreflight|TestAcquireInitMutationGateReleasesOnPreflightError|TestInitGateBusyClassifiedAsLockContention)$' ./cmd/bd
./scripts/test.sh -run '^(TestResolvePhysicalRootsSharedServerWinsOverMetadata|TestResolvePhysicalRootsSideEffectFree)$' ./internal/doltserver
golangci-lint run ./cmd/bd/...
GOTOOLCHAIN=go1.26.5 make test
```

Final baseline evidence:

- `make test`: exit 0
- `cmd/bd`: 310.403s
- `internal/doltserver`: 189.989s
- total coverage: 40.0%
- reviewed patch SHA-256: `631f53808dfe87fd342e7733046954608bec735ec4d463272c783c8c311fade0`

Two independent Sol reviewers examined the frozen composite for concurrency, cleanup, init-safety, and behavior neutrality. They found two major issues during review—pre-confirmation shared-server directory creation and accidental loss of the operational resolver fallback. Both were corrected with a pure plan/execute split; the final hash was re-reviewed and approved with no blocker or major findings.

## Risk and scope

This changes the init-safety critical path and requires substantive human maintainer review. It does not change storage schemas, confirmation policy, accepted contention outcomes, backend selection, or test inventory. The author will not self-merge it.

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -283,6 +283,9 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 		serverUser, _ := cmd.Flags().GetString("server-user")
 		database, _ := cmd.Flags().GetString("database")
 		destroyToken, _ := cmd.Flags().GetString("destroy-token")
+		// Keep the exact input the former destructive-confirmation block used:
+		// before config/dirname fallback and before normalization.
+		destroyTokenPrefix := prefix
 
 		// --force is a deprecated alias for --reinit-local. They share
 		// semantics for the local data-safety guard; both refuse remote
@@ -662,59 +665,6 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 			}
 		}
 
-		// Even with --reinit-local, warn about existing data and require
-		// confirmation. Non-interactive mode accepts --destroy-token for
-		// explicit opt-in; interactive mode prompts for typed confirmation.
-		if reinitLocal {
-			if count, err := countExistingIssues(prefix); err == nil && count > 0 {
-				fmt.Fprintf(os.Stderr, "\n%s Re-initializing will destroy the existing database.\n\n", ui.RenderWarn("WARNING:"))
-				fmt.Fprintf(os.Stderr, "  Existing issues: %d\n\n", count)
-				fmt.Fprintf(os.Stderr, "  This action CANNOT be undone. All issues, dependencies, and\n")
-				fmt.Fprintf(os.Stderr, "  Dolt commit history will be permanently lost.\n\n")
-				fmt.Fprintf(os.Stderr, "  Before proceeding, consider:\n")
-				fmt.Fprintf(os.Stderr, "    bd export > issue-export.jsonl    # Export issue records, not full DB state\n")
-				fmt.Fprintf(os.Stderr, "    bd dolt status              # Check if this is a server config issue\n\n")
-				if term.IsTerminal(int(os.Stdin.Fd())) {
-					fmt.Fprintf(os.Stderr, "Type 'destroy %d issues' to confirm: ", count)
-					scanner := bufio.NewScanner(os.Stdin)
-					scanner.Scan()
-					expected := fmt.Sprintf("destroy %d issues", count)
-					if strings.TrimSpace(scanner.Text()) != expected {
-						fmt.Fprintf(os.Stderr, "\nAborted. Database was NOT modified.\n")
-						return &exitError{Code: ExitLocalExistsRefused}
-					}
-				} else {
-					// Non-interactive (piped input, AI agent, etc.)
-					//
-					// ADR invariant (engdocs/adr/0002-init-safety-invariants.md):
-					// runtime error text must not echo a complete destructive
-					// invocation. See 'bd help init-safety' for the token
-					// format. This closes the 58f5989bf failure class where
-					// an agent copy-pasted the suggested command.
-					expectedToken := FormatDestroyToken(prefix)
-					if destroyToken == expectedToken {
-						fmt.Fprintf(os.Stderr, "Destroy token accepted. Proceeding with re-initialization.\n")
-					} else {
-						fmt.Fprintf(os.Stderr, "Refusing to destroy %d issues in non-interactive mode.\n", count)
-						fmt.Fprintf(os.Stderr, "  See 'bd help init-safety' for the required --destroy-token format.\n")
-						fmt.Fprintf(os.Stderr, "  Or export issue records first: bd export > issue-export.jsonl\n")
-						return &exitError{Code: ExitDestroyTokenMissing}
-					}
-				}
-			}
-		}
-
-		// Handle stealth mode setup
-		if stealth {
-			if err := setupStealthMode(!quiet); err != nil {
-				return fmt.Errorf("setting up stealth mode: %v", err)
-			}
-
-			// In stealth mode, skip git hooks installation
-			// since we handle it globally
-			skipHooks = true
-		}
-
 		// Check BEADS_DB environment variable if --db flag not set
 		// (PersistentPreRun doesn't run for init command)
 		if dbPath == "" {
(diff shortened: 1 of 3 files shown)


maphew (commented):
The diagnosis is exact and the fix is the right shape: `countExistingIssues` was precisely the reader that raced the replacement, and `acquireInitMutationGate` now spans preflight and replacement in one `AcquireAll` hold — no retry, no sleep, no weakened assertion, and the concurrency test untouched. Specific credit for `destroyTokenPrefix := prefix` at init.go:288: moving the preflight past the config/dirname fallback and `normalizeIssuePrefix` would have silently changed the accepted destroy token for dotted and directory-derived prefixes, and you both caught it and documented why. We verified the claim independently — no assignment to `prefix` exists between the flag read and the old block's position, so the token is byte-identical to base. The plan/execute split (side-effect-free `DoltDirPath` for planning, refused init no longer writes `.git/info/exclude`) is a real improvement beyond the race fix, and `TestAcquireInitMutationGateReleasesOnPreflightError` proves release-on-refusal by re-acquiring rather than mocking, which is the right assertion. An init-path change under the destructive-preflight gate gets this much scrutiny by policy; the change earned its green.

Verification in a detached worktree at d9c2ee1: build, vet, gate tests (including 40x stress + 2x `-race` on the new tests — no flake, and no data race on the hoisted `exclusiveGateOnWait` seam), remote-safety and init-safety suites, and the actual regression under the sanctioned harness — `TestEmbeddedInitConcurrent`: `10/10 succeeded, 0/10 lock error, 0/10 timed out`, zero panics. CI green across all ~110 checks.

One thing to settle before this merges, because §E3 freezes these as scriptable API: **the refusal-precedence flip is real and the body currently says otherwise.** The destructive preflight moved from before the remote-safety chokepoint to after it, so `bd init --reinit-local` with populated local data AND an origin advertising `refs/dolt/data` now exits 10 (remote divergence) where base exited 12 (destroy-token missing), and the interactive path that yields 11 is unreachable in that configuration. Nothing is destroyed either way, and the new order is arguably better UX (base would accept a destroy token and then refuse with 10 anyway) — but the body's "refusal exit codes ... are unchanged" is true only per-block, not for precedence, and docs/recovery/init-safety.md documents each code with its symptom text. If intended (we think it's defensible): say so under "Safety and behavior preservation", touch the doc, and ideally pin the new precedence with a case — the existing protocol tests miss it because their fixtures have no local `.beads`, so the count is 0. (While checking: init-safety.md:125 lists "Refusing to destroy N issues in non-interactive mode" as the exit-11 symptom, but that path returns 12 — pre-existing doc bug, happy to take that one ourselves.)

Non-blocking findings, roughly in order of interest:

- `buildWorkspaceGateSet` skips any extra root whose parent doesn't exist. Base's `ResolveDoltDir` mkdir'd `~/.beads/shared-server/dolt` as a side effect so the physical-root gate always materialized; `DoltDirPath` creates nothing, so first-use shared-server init (`BEADS_DOLT_SHARED_SERVER=1`, fresh host) silently plans a gate that gets skipped, and `ResolveDoltDir` later creates the dir ungated. Residual exposure is cross-workspace first-use racing (empty dir, low impact); a `MkdirAll` of the gate parent before `AcquireAll` closes it.
- The typed confirmation now blocks on stdin while holding the exclusive workspace + physical-root gates — inherent to the fix (confirmation must be atomic with replacement), and we traced the preflight's store-open path to confirm no self-deadlock (it takes only the embedded flock, never a workspace gate). But an operator who walks away at the prompt fences every bd command in the workspace, which is a new unbounded-hold class worth one sentence in the body/doc. Related: a refused reinit now leaves the `.beads.gate.lock` file behind (Release removes only the `.info` sidecar) where base refused without touching the filesystem.
- Test nit: `TestAcquireInitMutationGateKeepsReplacementExclusiveDuringPreflight` blocks unconditionally on its channels, so on a loaded box a lost 5s `exclusiveGateWait` shows up as an unattributed package timeout; `TestInitGateBusyClassifiedAsLockContention` shortens the wait to 10ms for exactly this reason — same trick, or `select` on both channels.
- Two pre-existing hazards this refactor makes newly legible — we'll file these, no action needed here: `runInitReinitPreflight` returns nil on *any* `countExistingIssues` error (a 5s timeout on a big DB silently skips the destroy confirmation), and `countExistingIssues` resolves its workspace via the walk-up `FindBeadsDir()` while the gate is keyed on `beadsDirForInit`, so when they diverge the preflight counts a database the gate doesn't cover.

Coordination for the queue: #5136 edits the same early remote-safety block whose ordering this changes (textual conflict likely), and #4762 un-skips the PTY exit-11 test that exercises the prompt now inside the gate hold — whichever of the three lands second should rebase and re-run the init-safety suite.

Review provenance: delegated agent review (claude-opus, 40x stress + race + harness regression re-run) + maintainer adjudication.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277228372

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a real race window in `bd init --reinit-local`: the destructive-reinit confirmation and destroy-token preflight previously ran *before* the exclusive workspace gate was held, so two concurrent reinits could each clear confirmation and race toward destroying local issues. Moving that preflight inside the held gate makes the "type `destroy N issues`" safety check genuinely exclusive — which matters because it is the last guard between a user and irreversible local issue loss.

### Original PR Review
Decision: approve.

Specific gaps this PR closes (validated by review):
- The destructive reinit confirmation / destroy-token preflight now runs **inside** the held exclusive workspace gate, so a second concurrent `bd init --reinit-local` blocks instead of both clearing confirmation (`runInitReinitPreflight` under the held gate; pinned by `TestAcquireInitMutationGateKeepsReplacementExclusiveDuringPreflight`).
- Gate planning uses `DoltDirPath` (no mkdir) with `ResolveDoltDir` deferred, so `~/.beads/shared-server/dolt` is not created before the user confirms; path normalization is preserved.
- A preflight refusal releases the gate, and the refusal exit codes survive `errors.Join` (`errors.Join(err, h.Release())` plus top-level `errors.As` extraction), so `ExitLocalExistsRefused` / `ExitDestroyTokenMissing` are preserved with no handle leak on any path.
- The `--prefix` destroy-token value is captured before fallback/normalization, reproducing the exact token the pre-fallback destructive block used.
- Stealth setup is deferred until after the held-gate preflight and is additionally skipped on gate-acquire failure (strictly safer).

Non-blocking follow-ups (shared for visibility — NOT addressed in this PR; filed as tracked beads):
- `bd-cyw6` — [Change Impact / Blast Radius · minor] Add a regression test pinning `ExitRemoteDivergenceRefused` and no local destruction for `--reinit-local` with both local issues **and** remote dolt data **and** no `--discard-remote`. The refusal reorder changed the non-interactive exit for that intersection (benign-to-beneficial) but it is currently untested.
- `bd-pkg7` — [Error Handling & Resilience · nit] Add an `errors.As(err, &ee)` assertion to pin exit-code survival through `errors.Join`.
- `bd-7tho` — [Concurrency · nit] If the gate tests are ever parallelized, replace the package-global `exclusiveGateOnWait` test seam with per-test injection (latent only; current gate tests are serial).
- `bd-9wwb` — [Docs] Release-note the behavior change: the exclusive workspace gate is now held across the interactive reinit confirmation prompt.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `d9c2ee1` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up. The change rebases cleanly onto current `main` with an unchanged patch, and the merge queue owns the deterministic rebase-merge from here.

CI: https://github.com/gastownhall/beads/actions/runs/30782237514

### Review Iterations
1 review pass performed — Claude (reasoning-heavy) and Codex (evidence-first), plus synthesis and a quality scorecard — reaching approve with no fix iterations required. Every finding was minor or nit with no merge gate.

---

Issue #5273, opened 2026-08-02
perf(server-mode): remaining round-trip reductions from the PR 4167 showcase


maphew:
Body:

## Summary

PR #4167 (@Shockwave2k, with a companion investigation in #4165) demonstrated a set of remote-Dolt round-trip optimizations against a ~200 ms RTT server, with measured wins of `bd list` 12 s -> ~3 s and `bd create` timeout -> ~8 s, and a 2.4x spot-check against then-current `main`. The PR was a showcase (never meant to merge) and its branch predates the June history rewrite, so it has been closed — this issue tracks the ideas from it that are **confirmed still absent** from `main`, so they can land as small independent patches. Full measurements and methodology are in the PR body of #4167. Direct response to #4102; adjacent to #3239 and #3760.

## Confirmed-unlanded items (each independently mergeable)

**Connection open (per command):**
1. Read-only fast path in `openServerConnection`: when `CreateIfMissing=false`, skip the second no-database pool and the `SHOW DATABASES` probe; ping the main pool and translate MySQL error 1049 to `databaseNotFoundError`. (~1 handshake + ~2 RTT per command; the `Gateway` path already does exactly this — `internal/storage/dolt/store.go`.)
2. Skip the `net.DialTimeout` preflight for non-local TCP servers — the driver's own `Timeout` fast-fails, and auto-start can't apply remotely. (`store.go`, `newServerMode`.)
3. Drop the redundant `PingContext` after `openServerConnection` when the open path already pinged. (`store.go`.)
4. Pin `MaxAllowedPacket` in the server DSN so the driver skips its `SELECT @@max_allowed_packet` discovery query (-1 RTT per cold conn). Companion to the already-landed `InterpolateParams` (#4617). (`internal/storage/doltutil/dsn.go`.)

**Per-read overhead:**
5. `withReadConn` helper: run single-statement reads (`GetConfig`, `GetAllConfig`, `GetMetadata`, `GetLocalMetadata`) on a pooled `*sql.Conn` instead of a `BEGIN`/`ROLLBACK` transaction (-2 RTT per read). Widen the `issueops` getters from `*sql.Tx` to a small querier interface — `issueops.SQLQuerier` already exists for the commit helpers. (`internal/storage/dolt/store.go`, `config.go`; `issueops/config_metadata.go`.)
[...]

PR #5316, opened 2026-08-03
perf(dolt): pin MaxAllowedPacket in the server DSN to drop a per-connection round trip (#5273 item 4)

maphew:
## What

Pins `MaxAllowedPacket` in the two MySQL DSN builders so the driver stops probing the server for it on every new connection.

- `internal/storage/doltutil/dsn.go` (`ServerDSN`)
- `internal/storage/dbproxy/util/dsn.go` (`DoltServerDSN`)

This is **item 4** of #5273, the salvage list from the #4167 showcase, and the direct companion to #4617 (`interpolateParams` in the same file, same shape).

## Why

Both builders construct `mysql.Config` as a composite literal rather than through `mysql.NewConfig()`, which is what normally supplies the driver's defaults. `MaxAllowedPacket` is therefore left at zero, `FormatDSN` emits `maxAllowedPacket=0`, and go-sql-driver/mysql runs

```sql
SELECT @@max_allowed_packet
```

while establishing **every** new connection — its connector takes that probe whenever `cfg.MaxAllowedPacket <= 0`. That is an extra round-trip per connection, paid on every pool expansion. Against a high-latency remote Dolt server, connections are opened far more often than the packet limit ever changes.

The defect is visible in the emitted DSN before this change:

```
root@tcp(h:3307)/?...&tls=false&maxAllowedPacket=0
```

## The value is 1 GiB, deliberately not the driver's 64 MiB default

The obvious fix — pin `defaultMaxAllowedPacket` (64 MiB) so the DSN matches a `NewConfig()` caller — is wrong, and cross-vendor review caught it. `go-mysql-server` declares:

```go
"max_allowed_packet": &sql.MysqlSystemVariable{
    Type:    types.NewSystemUintType("max_allowed_packet", 1024, 1073741824),
    Default: int64(1073741824),
},
```

Dolt defaults to **1 GiB** and the sysvar type caps it there, so no Dolt server can accept a larger packet. Pinning 64 MiB would have lowered the client ceiling from ~1 GiB to 64 MiB and made the driver reject oversized statements locally with `ErrPktTooLarge` — a real regression for large imports and base64 content. The first commit made that mistake; the second fixes it, and both are kept so the reasoning is on the record.

## Open question for a maintainer — please rule before merging

A second review pass raised a trade-off I do not think I should settle unilaterally, and I could not find it acknowledged in #5273 or #4167.

Under the old probing behaviour the driver set its ceiling from the server's **configured** value (`mc.maxAllowedPacket = n - 1`). So an operator who lowered `max_allowed_packet` below 1 GiB got a client-side rejection. After this change they do not.

What makes it a judgement call rather than a clear regression: **the server does not appear to enforce this sysvar for ordinary query packets anyway.** In go-mysql-server, `max_allowed_packet` is read in exactly one non-test place — `sql/expression/function/load_file.go`, for `LOAD_FILE` — and nowhere in the query-packet path. Vitess's `MaxPacketSize` is the protocol's 16 MiB chunking constant, unrelated to the sysvar. So the client-side cap was arguably the *only* thing honouring a lowered setting, and whether Beads intends to keep honouring it is a policy question:

- **Accept as-is** — treat the old behaviour as incidental, take the round-trip saving. Oversized statements then reach the server instead of being refused locally.
- **Probe once and cache** — keep fidelity to the operator's setting, pay the probe once per server rather than per connection. Materially more work; touches pool setup.

I have written the trade-off into the code comment either way, so it is not silent. **This PR is not queued for auto-merge** pending that ruling.

## Verification

- TDD RED confirmed: reverting only the struct field makes both new tests fail with `got 0`, and the failure message prints the offending DSN carrying `maxAllowedPacket=0`.
- Tests assert the **parsed** value via `mysql.ParseDSN`, not the DSN text — `FormatDSN` omits the parameter when it equals the driver default, so a substring check would pass for the wrong reason. They also assert the ceiling is at least Dolt's 1 GiB, so a future change back to the driver default fails instead of silently shrinking the limit.
- `go build -tags gms_pure_go ./...` clean on both `linux/amd64` and `GOOS=windows GOARCH=amd64 CGO_ENABLED=0`.
- `go vet -tags gms_pure_go ./...` clean.
- `golangci-lint run --config=.golangci.yml --modules-download-mode=readonly --timeout=5m --build-tags=gms_pure_go` — 0 issues on both target tuples.
- `go test -tags gms_pure_go -count=1 ./internal/storage/doltutil/... ./internal/storage/dbproxy/util/...` — pass.
- Cross-vendor review run twice (codex gpt-5.6-sol), once per commit; both findings are reconciled above rather than waved off.

## Credit

The round-trip analysis originates in the showcase work of **Hans Boese (@Shockwave2k)** in #4167 / #4165. Per #5273's credit request, this references both.

## Noticed, not changed

Both configs also leave `CheckConnLiveness` false (visible as `checkConnLiveness=false` in the DSN) where `mysql.NewConfig()` sets it true — the same composite-literal-vs-NewConfig root cause. Out of scope here; changing it has its own behavioural implications and deserves its own PR.


bee-ghosttrack (approved):
I tried to break this one on the obvious axis — a client packet ceiling pinned *above* the server's — and could not. The value is right, the tests are not vacuous, and the scope is honest. Approving with nits, plus a ruling on the open question you asked for.

## What I verified

**The round-trip claim is real, not folklore.** `go-sql-driver/mysql@v1.10.0` connector.go:176 takes the probe branch only when `cfg.MaxAllowedPacket <= 0`:

```go
if mc.cfg.MaxAllowedPacket > 0 {
    mc.maxAllowedPacket = mc.cfg.MaxAllowedPacket
} else {
    maxap, err := mc.getSystemVar("max_allowed_packet")   // connector.go:180
    ...
    mc.maxAllowedPacket = n - 1                            // connector.go:190
}
```

and dsn.go:376 emits the param whenever it differs from `defaultMaxAllowedPacket`, so a zero-valued composite literal really did ship `maxAllowedPacket=0`. My mutation probe printed exactly that DSN — see below. One awaited round trip per new connection, confirmed at the source.

**1 GiB is at the Dolt ceiling, not above it.** `dolthub/go-mysql-server@v0.20.1-0.20260805191915-e5eafe0da809/sql/variables/system_variables.go:1487-1493`:

```go
Type:    types.NewSystemUintType("max_allowed_packet", 1024, 1073741824),
Default: int64(1073741824),
```

Default 1 GiB, and the sysvar type's own upper bound is the same 1 GiB, so no Dolt server can be configured to accept a larger packet. `1 << 30` is therefore the exact ceiling — not over it. The "silent write failure on large blobs" hazard I went looking for does not exist here.

**These DSNs only ever address a Dolt server.** `mysql` as a storage backend is a tombstone — `guardLegacyUpgradeWorkspace` refuses it (`cmd/bd/legacy_upgrade_guard_test.go:397`). I walked every non-test caller of both builders (`cmd/bd/**`, `internal/doltserver/doltserver.go:1529,1568`, `internal/storage/dolt/store.go:2204`, `internal/storage/uow/dolt_sql_provider.go:373`, `internal/storage/dbproxy/server/{doltserver.go:121,external_dolt_server.go:64}`); all target dolt sql-server. So the pin cannot overshoot a real MySQL's lower default.

**Embedded path is correctly out of scope.** `internal/storage/embeddeddolt/open.go:33,98` builds a `file://…` DSN for `github.com/dolthub/driver/v2` — no MySQL wire protocol, no handshake, no `max_allowed_packet` probe to drop. Nothing to pin there. The PR is right to leave it alone, and right not to claim it.

**Derived DSNs inherit the pin.** Every other non-test `FormatDSN()` site — `store.go:2229, 2255, 2304, 2759, 4324` — starts from `mysql.ParseDSN(s.connStr)` and only overrides timeouts, so `maxAllowedPacket=1073741824` rides through to the long-timeout, one-shot and migration connections. No gap.

**No other DSN params changed.** The diff on `dbproxy/util/dsn.go` is a gofmt realignment plus the one field; TLS switch, `ClientFoundRows`, `Timeout`, `AllowNativePasswords` all byte-identical. `interpolateParams` in `doltutil` is untouched (realigned only). `checkConnLiveness=false` is left alone and disclosed under "Noticed, not changed" — correctly, it is a separate behaviour change.

**No downstream string-matching breaks.** Every other DSN assertion in the tree is a `strings.Contains` / `assert.Contains` on `tls=false` (`doltutil/dsn_test.go:21,119,150`, `dbproxy/util/dsn_test.go:52`, `dbproxy/server/external_dolt_server_test.go:232,239`); no exact-equality DSN assertion exists to break.

**No write-path side effects.** `connector.go:192` only lowers `maxWriteSize` when `maxAllowedPacket < maxPacketSize` (16 MiB−1); at 1 GiB that branch is not taken, exactly as under probing. `packets.go:1058`'s `longDataSize` is a threshold, not an allocation, and lands within one byte of the probed value. Nothing preallocates 1 GiB.

## Commands run (worktree at 76dd5e377, detached)

```
go build -tags gms_pure_go ./internal/storage/doltutil/... ./internal/storage/dbproxy/... ./internal/storage/dolt/...   rc=0
go vet   -tags gms_pure_go ./internal/storage/doltutil/... ./internal/storage/dbproxy/util/...                          rc=0
go test  -tags gms_pure_go -count=1 ./internal/storage/doltutil/... ./internal/storage/dbproxy/util/...                 rc=0
  PASS TestServerDSN_PinsMaxAllowedPacket
  PASS TestDoltServerDSN_PinsMaxAllowedPacket
  (all pre-existing tests in both packages PASS)
GOOS=linux   GOARCH=386   CGO_ENABLED=0 go build -tags gms_pure_go <both pkgs>                                          rc=0
GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build -tags gms_pure_go <both pkgs>                                          rc=0
gofmt -l <4 changed files>                                                                                              (empty)
git merge-tree --write-tree origin/main 76dd5e377                                                                       rc=0
git diff --check origin/main..76dd5e377                                                                                 rc=0
```

I ran the unit tests only — no embedded-dolt run, since the change cannot reach that path.

## Mutation probes — both assertions bite

**Probe A — delete the `MaxAllowedPacket` field from both builders (revert the fix):**

```
--- FAIL: TestServerDSN_PinsMaxAllowedPacket
    dsn_test.go:61: ... got 0 from DSN "root@tcp(dolt.example.com:3307)/?checkConnLiveness=false&
                    interpolateParams=true&multiStatements=true&parseTime=true&timeout=5s&tls=false&maxAllowedPacket=0"
--- FAIL: TestDoltServerDSN_PinsMaxAllowedPacket
    dsn_test.go:22: ... got 0 from DSN "root@tcp(127.0.0.1:3306)/?checkConnLiveness=false&
                    multiStatements=true&parseTime=true&timeout=5s&tls=false&maxAllowedPacket=0"
rc=1
```

RED confirmed on both, and the failure output independently corroborates the PR's central factual claim: the pre-change DSN did carry `maxAllowedPacket=0`.

**Probe B — pin the driver's own 64 MiB default instead (`const maxAllowedPacketBytes = 64 << 20`):**

```
--- FAIL: TestServerDSN_PinsMaxAllowedPacket
    dsn_test.go:73: MaxAllowedPacket must be at least Dolt's 1073741824-byte ceiling ...; got 67108864
--- FAIL: TestDoltServerDSN_PinsMaxAllowedPacket
    dsn_test.go:30: ... got 67108864
rc=1
```

So the second assertion is load-bearing too — it is the one that actually guards the 64 MiB regression the second commit fixed. Neither test is vacuous. Worktree restored clean after each probe (`git status --porcelain` empty, verified twice).

The PR's reasoning for asserting the *parsed* value rather than the DSN text is correct and I confirmed it: at 64 MiB, dsn.go:376's `!= defaultMaxAllowedPacket` guard makes `FormatDSN` omit the parameter entirely, so a `Contains("maxAllowedPacket")` check would have failed for a passing config and passed for a probing one.

## Your open question — my ruling: accept as-is

I checked this independently rather than taking the PR's word, and the analysis holds:

- `go-mysql-server` reads `max_allowed_packet` in exactly one non-test place: `sql/expression/function/load_file.go:163`, for `LOAD_FILE`. Nowhere in the query-packet path.
- `dolthub/vitess@v0.0.0-20260604210335-0893abc80542/go/mysql/**` never reads the sysvar at all. Its `MaxPacketSize = (1 << 24) - 1` (`go/mysql/constants.go:26`) is the protocol's 16 MiB chunking constant, unrelated.

So a lowered `max_allowed_packet` was enforced by nothing but the client's own probed ceiling. "Probe once and cache" would buy fidelity to a setting the server itself ignores, at real complexity in pool setup. Take the round trip. The trade-off is written into the const doc comment, which is the right place for it. Nothing here needs a maintainer ruling before merge as far as I'm concerned — clearing it.

## Nits (non-blocking)

1. **No CHANGELOG entry.** Split precedent: the doltutil `interpolateParams` change (6756ccb9e, #4617) added none; its dbproxy sibling (492bbe8ab) added one under `### Changed`. `CONTRIBUTING.md:313` lists CHANGELOG only as a release step, not a per-PR gate. Your call — I lean "add one," since this shifts a client-side ceiling operators could previously influence, and the sibling PR set that precedent.

2. **The const is duplicated and must stay in sync by hand.** `doltutil/dsn.go:30` and `dbproxy/util/dsn.go:15` each define their own `maxAllowedPacketBytes`, and the dbproxy copy's doc comment points at `internal/storage/doltutil.maxAllowedPacketBytes` — an unexported identifier in another package, so nothing but prose links them. One exported const (or a small shared package) would make drift impossible.

3. **The tests fence only one direction.** Both assert `>= 1 GiB`. A future bump to, say, `2 << 30` would stay green while exceeding what `NewSystemUintType("max_allowed_packet", 1024, 1073741824)` permits any Dolt server to accept — the client would then be willing to send packets no server is configured for. An exact-equality assertion, or an added upper bound, would fence both sides. Cheap to add.

4. **One-byte asymmetry with the path being replaced.** connector.go:190 sets `mc.maxAllowedPacket = n - 1` when probing, so the old effective ceiling was 1073741823; the pin is 1073741824. Immaterial given no server-side enforcement (see above), but the comment's "matching Dolt's max_allowed_packet default" is one byte off what the probe actually produced — worth a parenthetical if you touch the comment anyway.

## Housekeeping

PR body checks out against the diff: 4 files, all disclosed, nothing out of scope, no persisted format, no JSON shape, no exit codes, no user-facing docs affected. Credit to Hans Boese (@Shockwave2k) for #4167/#4165 is carried as #5273 asked. Merges clean onto `origin/main` with no whitespace damage. Keeping both commits so the 64 MiB mistake and its correction stay on the record was the right call.

PR #5318, opened 2026-08-03
test: cut proxied Create projects from 20 to 2 (bd-1rh.14)

julianknutsen:
## What

Replace 20 independent real proxied-Create project setups with two representative boundary journeys, while moving removed permutations to their typed lower-level owners.

- `scalar_and_output_boundary` keeps real subprocess, proxy routing, JSON/default-field persistence, and exact `--silent` output coverage.
- `aggregate_transaction_boundary` keeps real hierarchical ID, issue, parent-child edge, reverse dependency, and inherited-plus-own label persistence coverage.
- Typed input, builder, parser, priority, and issue-type tests own the scalar permutations that do not require a process or persistence boundary.
- No production code, workflow, shard manifest, skip, timeout, retry, or parallelism setting changes.

| Removed proxied permutations | Authoritative owner |
| --- | --- |
| Basic JSON result and silent output | `scalar_and_output_boundary` |
| Parent, dependency direction, and label inheritance | `aggregate_transaction_boundary` |
| Title flag, due/defer mapping, explicit ID, and inheritance controls | `TestGatherCreateInputMapsSingleIssueFlags` |
| Description, design, acceptance, assignee, and materialized fields | `TestBuildCreateIssueFromInput_PopulatesAllFields` |
| Priority and issue-type normalization | Existing validation/type unit tests |
| Dependency aliases, multiple specs, and unknown-type guidance | `TestParseDepSpecs` |

## Why

Ubuntu PR Risk shard 8 is a current PR-latency ceiling. Its seven-success baseline is `412, 395, 404, 373, 405, 389, 395s` (median `395s`). `TestProxiedServerCreate` created 20 fresh real projects and was the sole final 213–266 second tail in four of those seven runs.

The removed cases mostly proved flag parsing and object construction by repeatedly paying for a real store and process boundary. This keeps the distinct end-to-end risks while making field permutations fast and failure-local.

This does not supersede contributor work in #4738 or #4742. Their default-status and fresh-UOW-retry behavior remains outside this test-only change.

Tracked by `bd-1rh.14`.

## Measurement

- Current-checkout local shard 8, using prebuilt race test and non-race subprocess binaries: **80.44s**, PASS.
- Both retained Create children explicitly ran and passed; neither skipped.
- Hosted acceptance after opening: retain the next seven successful shard-8 samples and require every sample to be at most 300 seconds.

The local and hosted environments are not directly comparable, so the 80.44-second local result is a correctness/smoke measurement rather than the claimed hosted speedup.

## Verification

```text
./scripts/test.sh -v -run '^(TestGatherCreateInputMapsSingleIssueFlags|TestParseDepSpecs)$' ./cmd/bd
BEADS_TEST_PROXIED_SERVER=1 BEADS_TEST_BD_BINARY=/tmp/bd-1rh14-proxied BEADS_TEST_CMD_BINARY=/tmp/bd-1rh14-cmd-test bash .github/scripts/proxied-test-shard.sh 8 15
make test
go vet ./...
golangci-lint run ./cmd/bd/...
git diff --check
```

Two independent Sol blocker/major reviews approved the final frozen diff after their two findings were moved to the lower test seams. The configured pre-commit hook also passed under Go 1.26.5.

This nontrivial PR should receive substantive human review and be merged by a different maintainer; it will not be self-merged.

### DIFF
--- a/cmd/bd/create_deps_test.go
+++ b/cmd/bd/create_deps_test.go
@@ -13,10 +13,11 @@ import (
 
 func TestParseDepSpecs(t *testing.T) {
 	tests := []struct {
-		name    string
-		in      []string
-		want    []domain.DependencySpec
-		wantErr bool
+		name              string
+		in                []string
+		want              []domain.DependencySpec
+		wantErr           bool
+		wantErrSubstrings []string
 	}{
 		{
 			name: "empty input",
@@ -80,9 +81,10 @@ func TestParseDepSpecs(t *testing.T) {
 			},
 		},
 		{
-			name:    "unknown type rejected",
-			in:      []string{"nonsense:bd-1"},
-			wantErr: true,
+			name:              "unknown type rejected",
+			in:                []string{"nonsense:bd-1"},
+			wantErr:           true,
+			wantErrSubstrings: []string{"unknown dependency type", "blocked-by", "depends-on"},
 		},
 		{
 			name:    "empty type rejected",
@@ -125,6 +127,11 @@ func TestParseDepSpecs(t *testing.T) {
 			if tt.wantErr {
 				if err == nil {
 					t.Fatalf("parseDepSpecs(%v) = %v, want error", tt.in, got)
+				}
+				for _, want := range tt.wantErrSubstrings {
+					if !strings.Contains(err.Error(), want) {
+						t.Errorf("parseDepSpecs(%v) error = %q, want to contain %q", tt.in, err, want)
+					}
 				}
 				return
 			}
--- a/cmd/bd/create_input_test.go
+++ b/cmd/bd/create_input_test.go
@@ -3,10 +3,13 @@ package main
 import (
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 func TestCreateCommandRejectsExtraPositionalArguments(t *testing.T) {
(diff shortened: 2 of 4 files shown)


maphew (approved):
The "20 to 2" claim checks out exactly — we independently enumerated the removed top-level subtests at the merge-base and got 20, then traced every one to its claimed owner against the actual code rather than the table's prose. The two retained real-boundary projects are the right two (scalar/output on one project, hierarchical/transactional aggregate on another), and for description/design/acceptance/assignee both links of the chain got new specific assertions — the CLI-flag→`createInput` mapping *and* the `createInput`→`types.Issue` build. The `unknown_dep_type_rejected` replacement is a strict improvement: asserting on `parseDepSpecs`'s actual error-string content beats asserting on subprocess stderr. And `both_due_and_defer` staying untouched means due/defer keep full end-to-end coverage without the table having to claim it.

Verified in a detached worktree at 43d5344: build + vet + lint clean; the cited unit-test command passes with matching subtest counts; shard 8/15 under `BEADS_TEST_PROXIED_SERVER=1` passes in 48s locally (your 80s, same shape) with both retained children running, neither skipped. CI green on everything completed so far; prior-art scan clean.

One gap worth closing, non-blocking: **`--no-inherit-labels` on the proxied path now has no test at any seam.** The removed `parent_no_inherit_labels` subtest was the only exercise of `InheritLabelsFromParent: false` with a labeled parent — the embedded-path test (`create_embedded_test.go:257-285`) exercises a structurally different implementation (`create.go`'s manual merge, not the `domain.CreateIssueParams` boolean gate at `internal/storage/domain/issue.go:876`), and every domain/uow test we found sets the flag only to `true`. The new input test proves the CLI parses the flag, not that the suppression happens. Risk is low (the gated code is a single `if` with no distinct negative-branch logic), so this fits a one-line third child in `aggregate_transaction_boundary` or a domain-level unit case — either here or as a bd-1rh follow-up, your call.

Two nits consistent with what the series has already accepted for trivial passthroughs, recorded for completeness: non-default priority/type values are no longer proxied end-to-end (only the pure mapping functions and the default-value round trip remain), and a single `--deps` flag with mixed types is now parse-level only.

The consolidation discipline in this series continues to be a pleasure to review.

Review provenance: delegated agent verification (claude-sonnet, full removed-subtest→owner trace + shard re-run) + maintainer adjudication.

PR #5319, opened 2026-08-03
build: make repository text EOL deterministic

ecuthiell:
## What

Make repository text checkout bytes deterministic across hosts with a global `text=auto eol=lf` policy while explicitly excluding tracked JPEG binaries from text conversion and text diff/merge behavior.

This change:

- makes ordinary source and future detected-text formats materialize as LF even when workstation `core.autocrlf=true`;
- preserves the rationale formerly expressed by path-specific hook, embedded Markdown, migration, and generated-document rules;
- keeps binary JPEG bytes outside text conversion, diff, and merge handling;
- adds an isolated real-Git boundary covering listed source, a future unlisted extension, binary bytes, resolved attributes, and clean status;
- runs the whole fail-closed package on protected Ubuntu, macOS, and Windows matrix legs; and
- requires the combined matrix result through `CI Gate / Required`.

## Why

Fixes #5301.

The previous selectors protected only known LF-sensitive paths. On native Windows, other maintained text could still materialize as CRLF while Git reported a clean tree, causing formatting drift and leaving future byte-sensitive text dependent on workstation Git configuration.

## Review-requested split

The CI-gate shell/host/workspace hardening called out in the maintainer review at https://github.com/gastownhall/beads/pull/5319#pullrequestreview-4841847033 has been removed from this PR's final diff. Focused issue #5451 now records that independent threat model, and existing PR #5320 remains its sole implementation vehicle.

The original reviewed commit remains in history; subtractive commit `5cb7e2ad0f20a429713c31458a27f660fb798aa9` explicitly removes and supersedes that unrelated scope without force-rewriting the branch.

## Verification

- exact-head worktree rehydration: pass twice after live-base integrations (`attributes-changed=true`, `action=rematerialized`)
- ordinary and required-host Windows Git boundary: pass
- missing/wrong/unsupported host and zero/skipped/filtered-test falsifiers: rejected as expected
- exact `git check-attr -z` malformed/duplicate/missing/extra/lookalike protocol tests: pass
- workflow topology and `CI Gate / Required` propagation test: pass
- `actionlint`, `gofmt`, and `git diff --check`: pass
- final current-base diff: exactly six EOL-only files, with no CI-gate host-binding implementation or test
- strict classification: `READY_EXACT_BASE`

Current exact identity:

- base `5507f8466aa7be81e16c7777a8ea99ec7f1ac58b`
- head `8fd1a22aa09bb975f8cc39bcdc2c8a6cececb1a8`
[...]

PR #5320, opened 2026-08-03
fix(ci): remove recursive Make from lint formatting

### LINKED ISSUE #5302: ci: avoid hard-coded recursive Make in pr-lint wrapper
## Problem

The `ci-pr-lint` Make target launches `scripts/ci/pr-lint.sh`, and that wrapper hard-codes a nested `make fmt-check` command.

On a native Windows UCRT64 installation the GNU Make executable is named `mingw32-make.exe`; no `make.exe` alias is installed. Running the supported outer command reaches the wrapper, then fails at the nested lookup:

```text
mingw32-make ci-pr-lint
...
make: command not found
```

This is distinct from the existing Windows shell-bootstrap coverage: the outer Make and Git for Windows Bash are both working, but the wrapper rediscovers a differently named recursive executable.

## Expected behavior

`ci-pr-lint` should run its formatting and lint contract through the invoked native GNU Make without depending on a `make.exe` alias or ambient fallback. The canonical Make entrypoints should remain deterministic, and inherited `MAKE` text must never be interpolated back into shell source.

The simplest design is to give formatting one repository-owned script used by both `fmt-check` and `pr-lint.sh`, eliminating nested Make discovery entirely.

## Suggested verification

- Exercise the actual Makefile and wrappers with absolute UCRT64 `mingw32-make.exe`, Git for Windows Bash, and no usable `make`/`make.exe` fallback.
- Prove the shared formatting check preserves clean, unformatted, and underlying `gofmt` failure behavior.
- Poison Bash startup state, exported shell functions, and inherited empty/metacharacter-bearing `MAKE` values before parse-time probes and target recipes.
- Require exactly one Make invocation relative to the caller's inherited `MAKELEVEL`, including an ordinary outer-Make route, and prove `-n`, `-t`, and `-q` do not execute lint or formatting work.
- Bind the hosted UCRT package setup output to the exact executable under test and fail on zero or skipped semantic tests.

ecuthiell:
## Summary

- move the formatting check into one repository-owned shell script;
- have both `make fmt-check` and `scripts/ci/pr-lint.sh` call that script
  directly; and
- cover its clean, unformatted-file, and underlying-`gofmt` failure behavior.

## Why

The supported native-Windows entrypoint is `mingw32-make.exe`. The lint wrapper
previously entered through that executable and then launched a nested literal
`make fmt-check`. A normal UCRT64 installation provides `mingw32-make.exe`, not
a `make.exe` alias, so the wrapper failed during recursive Make discovery
before it reached formatting.

Both entrypoints now delegate to `scripts/ci/fmt-check.sh`. The implementation
there preserves the prior output contract and returns the underlying `gofmt`
failure status, while eliminating the second Make lookup entirely.

`scripts/ci/pr-lint.sh` remains a directly runnable public entrypoint; it does
not require a Make-exported sentinel or a privileged-shell relaunch.

## Scope decision

This reduced change intentionally excludes the broader Make, shell, host, and
workflow-authority hardening previously carried on this branch. Those design
questions remain separate under #5451 and #5463.

The former `.gitattributes` hunk is now absent; merged PR #5319 supplies the
repository-wide LF policy.

## Validation

- the three focused `TestFmtCheck*` tests on the rebased head;
- Bash syntax and ShellCheck for the new shared script;
- native `mingw32-make fmt-check` from a fresh Windows checkout; and
- exact four-file diff and whitespace hygiene checks.

Fresh hosted results are required for this rebased head.

Fixes #5302

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -337,21 +337,7 @@ fmt:
 
 # Check that all Go files are properly formatted (for CI)
 fmt-check:
-	@echo "Checking Go formatting..."
-	@UNFORMATTED=$$(gofmt -l .); \
-	status=$$?; \
-	if [ "$$status" -ne 0 ]; then \
-		echo "gofmt failed while checking formatting" >&2; \
-		exit "$$status"; \
-	fi; \
-	if [ -n "$$UNFORMATTED" ]; then \
-		echo "The following files are not properly formatted:"; \
-		echo "$$UNFORMATTED"; \
-		echo ""; \
-		echo "Run 'make fmt' to fix formatting"; \
-		exit 1; \
-	fi
-	@echo "All Go files are properly formatted"
+	@./scripts/ci/fmt-check.sh
 
 # Validate documentation references against actual CLI flags
 check-docs:
--- a/scripts/ci/fmt-check.sh
+++ b/scripts/ci/fmt-check.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# Shared Go formatting check for Make and PR lint wrappers.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+
+cd "$REPO_ROOT"
+
+printf 'Checking Go formatting...\n'
+if UNFORMATTED="$(gofmt -l .)"; then
+    :
+else
+    status=$?
+    printf 'gofmt failed while checking formatting\n' >&2
+    exit "$status"
+fi
+
+if [[ -n "$UNFORMATTED" ]]; then
+    printf 'The following files are not properly formatted:\n'
+    printf '%s\n' "$UNFORMATTED"
+    printf '\n'
+    printf "Run 'make fmt' to fix formatting\n"
+    exit 1
+fi
+
+printf 'All Go files are properly formatted\n'
--- a/scripts/ci/pr-lint.sh
+++ b/scripts/ci/pr-lint.sh
@@ -21,7 +21,7 @@ if [[ -n "${BD_LINT_NEW_FROM_MERGE_BASE:-}" ]]; then
     lint_scope=(--new-from-merge-base="$BD_LINT_NEW_FROM_MERGE_BASE")
 fi
 
-ci_time "gofmt check" -- make fmt-check
+ci_time "gofmt check" -- ./scripts/ci/fmt-check.sh
 ci_time "golangci-lint" -- \
     golangci-lint run --config=.golangci.yml --modules-download-mode=readonly \
         --timeout=5m --build-tags=gms_pure_go \
--- a/scripts/prlintmake/prlintmake_test.go
+++ b/scripts/prlintmake/prlintmake_test.go
@@ -1,0 +1,176 @@
+package prlintmake
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+func TestFmtCheckClean(t *testing.T) {
+	output, err := runFmtCheck(t, "exit 0\n")
+	if err != nil {
+		t.Fatalf("fmt-check failed: %v\n%s", err, output)
+	}
+	want := "Checking Go formatting...\nAll Go files are properly formatted\n"
+	if output != want {
+		t.Fatalf("output = %q, want %q", output, want)
+	}
+}
+
+func TestFmtCheckReportsUnformattedFiles(t *testing.T) {
+	output, err := runFmtCheck(t, "printf '%s\\n' cmd/bd/main.go internal/config/config.go\n")
+	if got := processExitCode(err); got != 1 {
+		t.Fatalf("exit = %d, want 1; error=%v\n%s", got, err, output)
+	}
+	want := "Checking Go formatting...\n" +
+		"The following files are not properly formatted:\n" +
+		"cmd/bd/main.go\n" +
+		"internal/config/config.go\n\n" +
+		"Run 'make fmt' to fix formatting\n"
+	if output != want {
+		t.Fatalf("output = %q, want %q", output, want)
+	}
+}
+
+func TestFmtCheckPreservesGofmtFailure(t *testing.T) {
+	output, err := runFmtCheck(t, "printf 'synthetic gofmt failure\\n' >&2\nexit 42\n")
+	if got := processExitCode(err); got != 42 {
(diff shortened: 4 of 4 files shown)


steveyegge (changes requested):
**Adversarial review (bee, beads-lane steward) — verdict: REQUEST-CHANGES (split the PR: the ~150-line core fix is mergeable; the ~3,800-line hardening battery is not).**

The core fix is real and correct: `pr-lint.sh` hard-coded a nested literal `make fmt-check`, which breaks on native UCRT64 Windows where GNU Make is `mingw32-make.exe` (#5302). Extracting `scripts/ci/fmt-check.sh` and calling it from both the Makefile target and `pr-lint.sh` fixes that cleanly — I verified the legacy-3.x branch empirically under real GNU Make 3.81. But note: all your greens are from stale base `578d4da`; the PR is CONFLICTING and has had zero CI against current main.

The problem is everything wrapped around the fix:

1. **Cygwin flake-tolerance reversal with no replacement** (pr.yml, removed "Set up Cygwin GNU Make" continue-on-error + warning step): the removed comment itself says cygwin.com downloads "time out often enough to red the required check." `windows-make-shell` is in `CI_GATE_REQUIRED`, so post-merge one cygwin.com timeout blocks the entire merge queue — and `ci_workflow_test.go:997` bakes the reversal in.
2. **Required jobs pinned to `C:\PROGRA~1\PowerShell\7\pwsh.exe` and PS "Core 7, ≥7.5"** (pr.yml:1078,1142,1246,1324,1481 + byte-pinned in tests): 8.3 short names are a per-volume setting runner images can disable; a routine runner-image update hard-fails three required jobs with no code change anywhere.
3. **`override .SHELLFLAGS := -p -c` with `SHELL := /bin/sh` at parse time** (Makefile:7-8,245): silently requires a `/bin/sh` accepting `-p` — dash gained it in 0.5.11, so `make build` dies at parse on older distros/busybox.
4. **Byte-pinned Makefile source-text assertions** (`prlintmake_test.go:59-120`, incl. `strings.LastIndex(makefile, "$(shell")` ordering): any ordinary future `$(shell …)` addition reds a required test with a cryptic ordering error.
5. **Blast radius already demonstrated:** the ~1,650 new lines pinning pr.yml topology are exactly what's conflicting with main's two unrelated CI changes — this test file now collides with essentially any pr.yml edit by anyone.
6. **Direct-invocation regression:** `pr-lint.sh:17-20` now refuses to run unless `CI_PRIVILEGED_BASH` is exported (only make exports it) — inverts the old contract with no note in `engdocs/LINTING.md`.
7. **Six near-identical ~40-60-line inline preflight blocks copy-pasted into pr.yml** — reintroducing the duplication-drift disease #5302 complained about; main.yml/ci-measurements.yml keep the plain path, so post-merge CI and PR CI now run different contracts for the same targets.

**Mergeable subset, happy to fast-track it:** `scripts/ci/fmt-check.sh` + the `fmt-check`/`ci-pr-lint`/`ci-pr-core` Makefile recipe changes + `pr-lint.sh` + the two .gitattributes lines (unless #5319 lands first — your own body flags the overlap) + the three `TestFmtCheck*` tests. Cygwin fail-closed needs a flake mitigation (retry/mirror/evidence); no 8.3 paths or image-version floors in required-job shell lines; drop or drastically loosen the source-text byte-pins.

**Rebase scope:** one conflicted file (`ci_workflow_test.go`, three mechanical hunks — dedupe the shell const, union the struct fields), pr.yml auto-merges — but only *textually*: both sides' topology tests must pass together over merged pr.yml, and all three Windows legs need a full re-run post-rebase.

ecuthiell:
Thanks — agreed. I’ll reduce this PR to the narrow #5302 correction:

- one repository-owned `fmt-check` script;
- the Make target and `pr-lint.sh` delegating to that script; and
- the three focused `TestFmtCheck*` behavior tests.

I’ll retain the two LF attribute rules only if #5319 has not landed when I
rebase. The host/shell/workflow authority battery, direct-invocation contract
changes, and copied preflights will come out of this PR; the broader design
questions will remain with #5451 and #5463 pending maintainer direction.

I’ll rebase the reduced branch onto current `main` and replace the stale-base
evidence with exact-head results.

ecuthiell:
The requested reduction is now published at exact head `ffb1a0b4fb1edac8344558b568ae5588e5227620` on current `main` (`f925f3f5b00d15c42fe17e42268c7d9c8b0a0013`).

The diff is now exactly five files, `+208/-16`:

- `scripts/ci/fmt-check.sh`
- Makefile delegation for formatting/lint entrypoints
- direct `scripts/ci/pr-lint.sh` delegation
- the two narrow LF attributes needed before #5319 lands
- the three focused `TestFmtCheck*` behavior tests

The rebase retained main's new `BD_LINT_NEW_FROM_MERGE_BASE` scope construction and changed only the subsequent formatting command. The former host/shell/workflow hardening battery, copied preflights, source-topology pins, and Make-only direct-invocation contract are absent.

Fresh focused tests pass on the rebased head. Hosted checks are running; if #5319 lands first, I will remove the then-redundant two attribute lines during the final rebase.

ecuthiell:
@steveyegge The reduced head is now green across all hosted checks and merge-clean. Contributor permissions do not allow me to use GitHub's request-review action, so this is the rereview signal for the requested scope reduction.

steveyegge:
Reduction verified against the requested split (head ffb1a0b4fb) — everything required to leave is gone, what remains is exactly the approved subset, and the fresh spot-review of the remaining hunks is clean (fmt-check.sh is a faithful port incl. the assignment-as-if-condition `set -e` exemption; the recipe works identically for both `make fmt-check` callers; tests assert behavior, not topology). CI 99 green / 0 fail on this head.

One final ordering wrinkle, from the split map's own "unless #5319 lands first" clause: **#5319 just merged** (6748ef650c68), and its global `* text=auto eol=lf` rewrite makes this PR's two `.gitattributes` lines (`*.go`, `.buildflags`) redundant — and their hunk's context lines no longer exist on main, so this will show CONFLICTING once GitHub recomputes.

Ask: drop the `.gitattributes` hunk entirely and rebase onto current main. That is the last item — the remaining 4-file diff merges on green re-push, no re-review.

ecuthiell:
Applied the final reduction:

- removed the `.gitattributes` hunk now that #5319 supplies the LF policy;
- rebased onto current `main` (`9628c0189f5e`); and
- confirmed the remaining diff is exactly the four intended files.

Current head: `f394ee76ce0269a0bbbe4b0e41b76312194baeab`

Fresh-head validation passed for the three focused behavior tests, Bash syntax,
ShellCheck on the shared script, native `mingw32-make fmt-check` from a fresh
Windows checkout, and diff hygiene. Hosted checks are running on the rebased
head. Per your note, I have not requested another review.

PR #5321, opened 2026-08-03
feat(storage): public backend contract + conformance suite for out-of-tree backends (bd-h3dib.2)

steveyegge:
## Why

This is the keystone of the external-backend door promised in the #4415 decision ([follow-up comment](https://github.com/gastownhall/beads/pull/4415#issuecomment-5168838972), 2026-08-03): *"the conformance suite currently sits under `internal/` where your module can't import it — we're exporting it (together with the backend interface surface it exercises) as a public package."*

Today an external Go module can neither implement the storage contract (the interface and its whole type closure are `internal/`-locked) nor import the conformance suite that would prove the implementation. The #4859 registry seam opened the *registration* half of the door; this PR opens the *contract* half. Tracked as bd-h3dib.2.

## What

- **`backend/`** — new public package: alias façade over the storage contract, following the repo's existing convention (root `beads` package, `issueops/`). `DoltStorage` + the 15 capability sub-interfaces, transaction types, all 20+ storage value types, the 36 domain types the signatures reach (with full enum const sets), the 14 sentinel errors (same values, so `errors.Is` holds across the module boundary), and thin wrappers over the #4859 registry (`Register`/`Deregister`/`Lookup`/`Registered`/`WorkspaceIsBeadsDir`). Aliases are identical types: a store written against `backend.*` names satisfies `storage.DoltStorage`. No behavior change anywhere.
- **`backend/conformance/`** — the conformance suite, moved verbatim from `internal/storage/conformance/` (pure `git mv`, 100% similarity; 7 in-tree test importers rewritten). External backends run the same assertions core runs.
- **Drift guard** — `backend/completeness_test.go` walks the reflected type closure of the contract (interface methods, exported struct fields, func types, generic instantiations) and diffs it against the alias decls parsed from this package's source. A future signature change that drags a new internal type into the contract fails in-tree at test time instead of silently re-locking the door. It caught two closure members a static scan missed (`IssueDetails`, `PersistenceMode`) during development.
- Package docs carry the minimal external-backend recipe and explicit **EXPERIMENTAL** stability terms (pin exact version, re-run suite per bump, interface-method additions are breaking and changelogged).

## Verification

- `go build ./...`, `go vet ./...` (plus `-tags integration,gms_pure_go` vet on the dolt/embeddeddolt packages) — clean
- `golangci-lint run` on touched packages — 0 issues (depguard boundary intact: no dolthub imports in moved files)
- `go test ./backend/ ./internal/storage/uow/` — pass
- CI-style embedded-Dolt conformance: `BEADS_TEST_EMBEDDED_DOLT=1 … -test.run '^TestConformance$'` — **PASS** (full suite incl. Audit; CI's invocation minus `-race`, which CI applies at build time)
- **External-module acceptance proof:** a throwaway out-of-tree module (via `replace`) with a stub store written entirely against `backend.*`/`issueops.*` names — embeds `backend.DoltStorage`, overrides 9 representative methods, `init()`-registers via `backend.Register`, and compiles `conformance.RunAll` against a `func(*testing.T) backend.DoltStorage` factory. Build/vet/test all rc=0.

## Open questions for review

1. **`DoltStorage` is a historical name** now public on a backend-agnostic contract. Kept as-is (export what exists; doc comment flags the history). A `backend.Store` alias would be a one-line additive follow-up if preferred.
2. **164 required methods is a heavy contract** for an external implementer. `RunUnsupportedContract` + the capability sub-interfaces give a legitimate partial-implementation path, but the weight is real — flagging rather than hiding it.
3. The structural half of the unsupported-contract check (the dolt backends' generated-shell completeness tests) stays internal; external backends get the behavioral half only. Acceptable for v1; doc-noteable later.
4. No `examples/external-backend/` yet — package docs carry the recipe; an example module is a natural follow-up once #4415's author starts consuming this.


steveyegge:
## Gate review verdict (lion, bd-h3dib.2): **APPROVE — merge**

Shared-account seat, so this comment is the approval; bee services and merges on it.

### Independently verified (read-only clone + scratch worktree at 390831928)

- **Diff shape matches the claim.** `backend/` is aliases and thin wrappers only — no behavior anywhere. The conformance move is a pure rename (R100 on every file except `conformance.go`, whose R098 delta I read in full: package-doc + stability text only). The seven in-tree importer rewrites are one-line import-path changes.
- **`go build ./...`, `go vet` (backend, uow, test/conformance), `go test ./backend/`** — all green locally; the seven rewritten importer packages compile (`-run '^$'`).
- **Drift guard is live, not decorative.** Mutation test: commenting out the `IssueDetails` alias makes `TestPublicSurfaceComplete` fail with the exact right message. The guard's alias parser keys by *target* (importpath.TypeName), so the `VCStatus = storage.Status` rename is handled correctly. Both exported conformance types (`Factory`, `IssueOperationsStagingFixture`) are walk roots, and every conformance entry point takes only those — closure coverage is real.
- **Out-of-tree acceptance replicated independently.** Fresh module (`replace` onto the PR head): a store embedding `backend.DoltStorage`, `init()`-registering via `backend.Register`, with `conformance.RunAll` compiled against a `func(*testing.T) backend.DoltStorage` factory — build/vet/test-compile all rc=0. The alias-identity claim holds across the module boundary.

### The red CI job is not this PR

`Managed-local proxied lifecycle (Linux, offline)` failed on `TestManagedLocalProxiedStopStartInterleave` — a 5s `proxy.lock` acquisition timeout in a stop/start interleave. That test shares zero code with this diff (the only uow file touched is a test-file import path), the same workflow is green on the last 10 main pushes, and both required `CI Gate / Required` checks passed. Classification: env/flake. Cheap to confirm with a job rerun before merge, but I don't consider it blocking.

### Open questions

1. **`DoltStorage` name** — keep as-is. Export-what-exists is the right call for the keystone PR; the doc comment already disarms the confusion. `backend.Store` as a one-line additive follow-up is fine *if* Mark trips on it; don't pre-add a second name for the same interface.
2. **164-method weight** — real, but the capability sub-interfaces + `RunUnsupportedContract` are a legitimate partial path, and EXPERIMENTAL terms make the weight renegotiable. Flagging it in the PR body was the honest move; no change requested.
3. **Structural half staying internal** — acceptable for v1. Worth one sentence in the backend package doc eventually, not now.
4. **No `examples/` yet** — agreed, wait for the first real consumer; the package-doc recipe is enough to start.

### One follow-up to file (non-blocking)

The drift guard covers the *type* closure only. A new **sentinel error**, **enum const** (e.g. a new `Status` value), or registry **function** added to the contract would not trip it — an external implementer couldn't name the new value until someone remembers the alias. Worth a small follow-up bead on bd-h3dib: either extend the guard (parse const/var alias decls and diff against the internal packages' exported sentinels/enum members for aliased types) or a doc'd checklist item on contract changes.

Epic bd-h3dib unblocked from my side. Nice keystone — the mutation-tested guard is exactly what an EXPERIMENTAL public surface needs.

PR #3996, opened 2026-05-16
test(embedded): split stdout/stderr in all JSON-parsing helpers (be-0ya2oe)

quad341:
## Summary

- PR #3909 (commit 0c47b3c5a) introduced an auto-export warning written to **stderr** when no Dolt remote is configured; any test helper that called `CombinedOutput()` and then parsed stdout as JSON would choke on the interleaved warning text
- Applies the existing pattern from `list_embedded_test.go` (lines 38–49) across all 66 affected embedded test files: for every site that parses JSON or an ID from stdout, capture stdout and stderr in separate `bytes.Buffer`s and parse only stdout
- Retains `CombinedOutput()` at sites that only assert on human-readable text or use output purely for error diagnostics

Fixes the CI failures observed in:
- PR #3988 CI run 25967558419 (`TestEmbeddedDuplicates/dry_run`)
- main CI run 25962887732 (`TestEmbeddedUpdate/update_json_output`)

Completes the cleanup started in PR #3909. Closes be-0ya2oe.

## Test plan

- [x] `go build -tags cgo ./cmd/bd/...` passes locally
- [x] `go test -tags cgo -run TestEmbeddedDuplicates ./cmd/bd/...` passes (skip without `BEADS_TEST_EMBEDDED_DOLT=1`)
- [x] `go test -tags cgo -run TestEmbeddedUpdate ./cmd/bd/...` passes (skip without `BEADS_TEST_EMBEDDED_DOLT=1`)
- [x] Audit grep — no JSON-parsing `CombinedOutput` sites remain in `cmd/bd/*_embedded_test.go`
- [ ] CI green on embedded Dolt test matrix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5323, opened 2026-08-03
fix(dolt): scrub stderr-directed git tracing before git-remote transfers

cuongbphv:
## What

`bd dolt push`/`fetch`/`clone` against a `git+` remote fails whenever the environment carries stderr-directed git tracing (`GIT_TRACE=1`, `GIT_TRACE_PACKET=2`, `GIT_CURL_VERBOSE=1`, ...). This PR scrubs those variables around remote transfers — both the in-process engine path and the dolt CLI fallback — so the transfer succeeds. File-target tracing (`GIT_TRACE=/tmp/git.log`) still works.

## Why

Dolt's git blobstore runs git plumbing through a runner that captures stdout and stderr into one buffer and parses object ids out of it (`store/blobstore/internal/git/runner.go`). With `GIT_TRACE=1` set, trace lines land in the same buffer as the object id, the next plumbing call gets a garbage object name, and the transfer dies with `failed to get remote db` — plus credential hints that send you debugging auth instead of tracing. I hit this debugging a real push failure: turning tracing *on* to investigate is exactly what breaks the push.

#3996 fixed the same combined-output problem in bd's own git helpers, but the vendored Dolt runner can't be patched from here, so the fix is to keep the poison out of the environment the runner inherits. Same shape as the git-hooks guard from #5186: new `internal/gittraceenv` with a refcounted `WithScrubbed` scope, and `versioncontrolops` now applies both env guards through one `withRemoteEnvGuards` helper so a call site can't pick up one and silently miss the other.

## Verification

Repro on stock bd (fails every time):

    bd dolt remote add origin git+file:///path/to/bare.git
    bd dolt push              # ok
    GIT_TRACE=1 bd dolt push  # Error 1105 ... failed to get remote db

With this branch both pushes pass. Covered by:

- `internal/gittraceenv` unit tests (classifier table, nested/concurrent scopes, subprocess env scrub);
- `internal/storage/versioncontrolops/remotes_gittrace_test.go` regression tests asserting Push/ForcePush/Fetch/Clone and the pull fetch-step all run with `GIT_TRACE` unset (red without the wiring, green with it);
- full `pr-core` suite (`go test -race -short -skip '^TestEmbedded' ./...`) on linux/amd64: no new failures vs. main in the same environment;
- end-to-end: `bd dolt push` then `GIT_TRACE=1 bd dolt push` against a `git+file://` bare remote, both green.


steveyegge (commented):
**Adversarial review (bee, beads-lane steward) — verdict: MERGE-AFTER-FIXES.**

The mechanism is sound and the tests are genuinely good (classifier table, nested/concurrent scopes, regression tests asserting Push/ForcePush/Fetch/Clone/Pull execute with `GIT_TRACE` unset and restored). Value-aware scrubbing that preserves file-target tracing is the right escape hatch. But the coverage claim ("all transfer paths") doesn't hold against the tree — and your own envguards.go:21-23 comment ("a call site that gets one protection and silently misses the other is exactly how GH#4272 shipped") describes precisely the holes left:

**Required before merge:**

1. **Server spawns not scrubbed — asymmetric with the hooks guard this mirrors.** `internal/doltserver/doltserver.go:1401-1408` and `internal/storage/dbproxy/server/doltserver.go:259-263` both do `cmd.Env = append(os.Environ(), githooksenv...)` when spawning the long-lived dolt sql-server — the exact sites GH#4272 patched for hooks. A sql-server auto-started from a shell carrying `GIT_TRACE=1` poisons every server-side git-protocol transfer for every client until server restart; neither `WithScrubbed` (wrong process) nor the CLI `ScrubEnv` touches it. One-line fix at each site: `cmd.Env = append(gittraceenv.ScrubEnv(os.Environ()), githooksenv...)`.

2. **CLI bootstrap clone not scrubbed.** `internal/storage/dolt/bootstrap.go:79` (`BootstrapFromRemoteWithDB`, via `cloneViaCLI` at :1006) spawns `dolt clone` with inherited env — it does not route through the patched `prepareDoltCLITransferCommand`. `GIT_TRACE=1 bd bootstrap` of a `git+` remote still dies with the exact bug this PR fixes — and a fresh-clone user is the most likely person to have tracing on while debugging.

**Recommended (follow-up-eligible; these gaps predate this PR — say so and I'll file the bead):**

3. `internal/remotecache/cache.go:147/:192/:209` — cross-rig cache push/clone/pull spawn dolt transfers with inherited env; a `git+` URL there breaks the same way (also lacks the hooks guard).
4. `stderrDirected` (gittraceenv.go:94-96) accepts `af_unix:` for ALL trace vars, but only the `GIT_TRACE2` family supports it — `GIT_TRACE=af_unix:/tmp/x` is kept yet draws "warning: unknown trace value" on stderr per plumbing call, the very corruption being scrubbed. Pass the var name into the classifier.
5. `internal/storage/domain/db/dolt.go:57-70` exposes unguarded `Fetch/Push/Pull/Clone` verbs (dormant today) — worth a comment pointing at `withRemoteEnvGuards` so the next wiring doesn't bypass both guards silently.

Checked and clean: federation peer transfers route through the patched choke point; creds/S3 env survive the scrub; `GIT_SSH*`/`GIT_ASKPASS`/credential env untouched; SQL `CALL DOLT_PUSH` paths serve non-git remotes so leaving them unwrapped is correct once fix 1 lands.

(Process note: your branch was update-branched by me to get CI off a stale snapshot; current Lint red is inherited from an unrelated main misspell being hotfixed in #5519 — ignore it.)

cuongbphv:
All five items addressed in 0372ad0f0 — items 3 and 4 fixed in-PR rather than deferred, since verifying them surfaced small fixes.

**1. Server spawns** — both sites now build the child env through a shared `doltserver.ServerSpawnEnv()`. The guard pair itself is composed from the two owning packages: `githooksenv.DisabledEnv(gittraceenv.ScrubEnv(os.Environ()))` — `DisabledEnv` is new, the subprocess counterpart of `WithDisabled` with exec last-wins replace semantics — so a spawn site can't pick up one guard and miss the other.

**2. Bootstrap clone** — `bootstrapCloneCmd` applies the same guarded env (both guards — the clone path can hit the GH#4272 template-hook class too, and its `remotecache` twin already guards clone).

**3. remotecache** — fixed in-PR instead of a bead: push/clone/pull spawn through one `doltCmd` choke point with the guarded env (previously all three inherited the caller's env with neither guard).

**4. `af_unix:` for non-trace2 vars** — one framing correction: `stderrDirected` is new code in this PR, so this one didn't predate it; fixed here. The classifier is name-aware — `af_unix:` stays only for the `GIT_TRACE2` family. While in there, two Windows gaps in the same function got fixed: env-name matching is now case-insensitive on Windows (`set Git_Trace=1` would otherwise survive `ScrubEnv` while the child resolves it case-insensitively), and a leading dir separator counts as a file target the way git's `is_absolute_path` reads it (`GIT_TRACE=/c/temp/git.trace` from Git Bash was being scrubbed, silently emptying the operator's trace file).

**5. Unguarded `Fetch/Push/Pull/Clone` verbs** — comment added as suggested, after tracing the wiring to be sure a comment is the right strength: the only production `Runner` behind this repository is a MySQL connection to the external sql-server (`uow`/`doltSQLProvider`), so the git plumbing spawns from the *server's* environment — process-env guards in bd would be inert here (fix 1 at spawn time is the protection), and wrapping would leak env mutation into unrelated concurrent git spawns for the duration of a transfer. The comment routes any future in-process wiring to `withRemoteEnvGuards`.

**Two pre-existing gaps found while verifying, bead-worthy:** (a) the CLI pull/fetch legs get the trace scrub but not the hooks override — `prepareDoltCLITransferCommand` leaves hooks to hand-application and only the push callers do it (`doltCLIPull`, `doltCLIPullFromPeer`, `doltCLIFetchFromPeer` are uncovered); (b) `bd dolt remote reset-data` parses `git ls-remote` CombinedOutput with an unscrubbed env and feeds the parsed refs to a destructive delete-push. Happy to take either as a follow-up.

Verification: `go build ./...`, `go vet`, and the touched packages (gittraceenv, githooksenv, doltserver, remotecache, storage/dolt, domain/db, versioncontrolops) green on windows/amd64 on top of your update-branch merge; the `TestExternalDoltServer_DSN/unix_socket` red is pre-existing Windows-only, unrelated.

steveyegge:
**Round-2 verification (bee, beads-lane steward): all five items confirmed in 0372ad0f0 — merging.**

Required fixes: both server-spawn sites now route through the shared `doltserver.ServerSpawnEnv()` (scrub + no-hooks composed from the owning packages), and the bootstrap clone applies the guard pair itself with a regression test. Follow-up-eligible items done in-PR as you noted: remotecache's three transfer spawns get the same pair via `doltCmd`, the `af_unix:` classifier is now name-aware (kept only for the GIT_TRACE2 family), and the dormant-verbs comment points at both guard paths. The two Windows repairs surfaced along the way (case-insensitive env-name matching to match Win32/git lookup; leading-dir-separator values honored as file targets per git's `is_dir_sep`) are correct and both pinned by tests. CI green across the board on 0372ad0f0. Nice round-2 — the guard-pair composition (`githooksenv.DisabledEnv(gittraceenv.ScrubEnv(...))`) is exactly the shape that stops the next GH#4272.

Issue #5327, opened 2026-08-04
test: TestForceStopUnverifiedPairedLegacyRecords fails deterministically on main (Ubuntu 26.04, Go 1.26.5)


mwotton:
## What

`TestForceStopUnverifiedPairedLegacyRecords`
(`internal/storage/dbproxy/proxy/force_stop_linux_test.go`) fails
deterministically on `main` on Ubuntu 26.04 / Linux 7.0.0-27-generic, x86_64,
Go 1.26.5. It fails inside `make test` and in isolation.

```
--- FAIL: TestForceStopUnverifiedPairedLegacyRecords (0.09s)
    force_stop_linux_test.go:99:
        	Error:      	Should be true
    force_stop_linux_test.go:100:
        	Error:      	Should NOT be empty, but was
FAIL	github.com/steveyegge/beads/internal/storage/dbproxy/proxy	0.109s
```

The two failing assertions are the proxy leg of the report (lines 99-100):

```go
assert.True(t, report.SignalSent)
assert.NotEmpty(t, report.QuarantinedPath)
```

`ForceStopUnverified` itself returns no error, and because these are
non-fatal `assert` calls the rest of the test runs: every backend-leg
assertion that follows (`report.Backend.RecordFound`, `PID`, `Executable`,
`SignalSent`, `QuarantinedPath`) passes, as do `assertHelperExited` for both
helpers and the two `NoFileExists` checks. So the paired-legacy path recovers
the backend record but reports no signal sent and no quarantined path for the
proxy record.

The sibling tests in the same file (including
`TestForceStopUnverifiedGoneProcess`) pass.

## Reproduction

```bash
./scripts/test.sh -run '^TestForceStopUnverifiedPairedLegacyRecords$' ./internal/storage/dbproxy/proxy/
```

Reproduced on `main` at 67f812a23 in a clean detached worktree, so it is not
caused by any local change. Not currently listed in `.test-skip`, and I could
not find an existing issue for it.

## Environment

- Ubuntu 26.04 LTS, Linux 7.0.0-27-generic, x86_64
- Go 1.26.5 linux/amd64
- `./scripts/test.sh` defaults (`gms_pure_go`, `BEADS_TEST_SKIP=dolt`)

## Note

Filed per engdocs/TESTING.md ("if you hit a test failure unrelated to your
change, don't silently skip it — check `.test-skip` and file an issue if it's
not already tracked") while validating an unrelated `cmd/bd` change. I have not
[...]

PR #1586, opened 2026-02-08
perf: replace N+1 GetLabels calls with batch GetLabelsForIssues in RPC list

turian:
## Summary

- RPC `handleList` called `store.GetLabels()` per-issue in a loop (N+1 queries), even though `SearchIssues` via `scanIssues` already batch-loaded labels
- Replaced with a single `store.GetLabelsForIssues()` batch call
- This is backend-agnostic (uses the `Storage` interface method, not SQLite internals)
- Moved `issueIDs` slice construction earlier to reuse it for labels, dep counts, and comment counts

## Test plan

- [x] `go build ./internal/rpc/` compiles cleanly
- [x] Full RPC test suite passes (`go test ./internal/rpc/`)

Fixes #1583

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1583, opened 2026-02-08
RPC handleList hydrates labels per-issue redundantly (N+1 queries)


turian:
## Summary

In `internal/rpc/server_issues_epics.go` (lines 1331-1335), `handleList` loops through issues calling `store.GetLabels(ctx, issue.ID)` one at a time after `store.SearchIssues()`.

But `SearchIssues()` already populates labels via `scanIssues()` (`internal/storage/sqlite/dependencies.go:1128-1139`), which batch-loads them with `GetLabelsForIssues()`.

Labels are hydrated **twice** — once efficiently (batch) and once with N+1 individual queries.

## Impact

Performance degradation on large issue lists. Each list/search via daemon makes N unnecessary extra SQLite queries.

## Files

- `internal/rpc/server_issues_epics.go:1331-1335` — the redundant per-issue loop
- `internal/storage/sqlite/dependencies.go:1128-1139` — the batch load that already ran

## Agent-safe fix strategy

Do not rely on SQLite internals. Fix it in the RPC layer in a backend-agnostic way by switching to a single batch call.

### Proposed change (robust)

**File:** `internal/rpc/server_issues_epics.go`

Replace the per-issue loop (lines 1331-1335) with:

1. Build `issueIDs` (already done at line 1338-1340 — move it up or reuse)
2. Call `store.GetLabelsForIssues(ctx, issueIDs)`
3. Assign `issue.Labels = labelsMap[issue.ID]`

This:
- Avoids N+1
- Works even if a backend does not hydrate labels in `SearchIssues`
- Keeps RPC behavior stable

### Why this is unlikely to be broken

- It's a clear "batch hydration" pattern.
- It uses the `Storage` interface method `GetLabelsForIssues`, which is already part of the contract.
- It doesn't depend on subtle behavior of `SearchIssues`.

### Regression/perf test (optional but helpful)

A unit test using a fake `Storage` that counts calls:
- Ensure `GetLabels` is not called N times
- Ensure `GetLabelsForIssues` is called once

This is agent-proof because it enforces the intended API usage. A lightweight fake struct implementing just the needed methods for `handleList` is enough if there's no existing mock pattern.

PR #5328, opened 2026-08-04
perf(label): count list-all labels from hydrated SearchIssues result

### LINKED ISSUE #5325: perf(label): label list-all redundantly re-fetches hydrated labels with N+1 GetLabels calls
## What

`bd label list-all` fetches every issue with `SearchIssues`, which already
hydrates `Issue.Labels` in bulk, and then throws that hydration away and calls
`store.GetLabels` once per issue. In embedded Dolt mode each of those calls
opens a fresh SQL connector, starts a transaction, and closes it, so a database
with N issues pays N redundant connect/transaction cycles for data it is
already holding in memory.

`cmd/bd/label.go` (`labelListAllCmd`):

```go
issues, err = store.SearchIssues(ctx, "", types.IssueFilter{})
...
labelCounts := make(map[string]int)
for _, issue := range issues {
    labels, err := store.GetLabels(ctx, issue.ID)   // <-- redundant
    ...
}
```

`issueops.SearchIssuesInTx` hydrates labels for the whole result set through
`hydrateIssueLabelsAndDeps` → `GetLabelsForIssuesFromTableInTx` unless
`filter.SkipLabels` is set. The filter here is the zero value, so every issue
returned already carries its labels.

`internal/storage/embeddeddolt/labels.go`:

```go
func (s *EmbeddedDoltStore) GetLabels(ctx context.Context, issueID string) ([]string, error) {
	var labels []string
	err := s.withConn(ctx, false, func(tx *sql.Tx) error {
		var err error
		labels, err = issueops.GetLabelsInTx(ctx, tx, "", issueID)
		return err
	})
	return labels, err
}
```

Each `withConn` is a short-lived `OpenSQL` connector plus a transaction. Passing
`""` as the table also makes `GetLabelsInTx` run an extra `IsActiveWispInTx`
probe per issue, so it is two round trips per issue, not one.

The proxied-server path (`runLabelListAllProxiedServer` in
`cmd/bd/label_proxied_server.go`) already does the bulk thing via
`GetLabelsForIssues` / `GetLabelsForWisps`. Only the direct-store path has the
N+1.

## Reproduction (stock beads, no orchestrator)

Installed `bd` 1.1.0 from Homebrew, embedded Dolt database, 455 live issues, no
orchestrator or agent framework running. All timings are wall clock on the same
workspace, back to back.

| Command | Wall clock | Output |
|---|---|---|
| `bd label list-all --verbose --json` | 39.74s | 3,580 bytes of JSON, 175 bytes of debug output |
| `bd label list-all --json` (profiled, non-verbose) | 33.75s | same result set |
| `bd list --limit 1 --json` | 8.92s | one issue |
| `bd list --all --limit 0 --json` | 10.17s | all 455 issues |

`bd list --all --limit 0 --json` walks the same 455 rows and hydrates the same
labels in ~10s. Aggregating `.[].labels` from that output reproduces the
`label list-all` label/count table **byte for byte identical** to what
`bd label list-all --json` prints. The extra ~24-30s is spent entirely on the
redundant per-issue lookups.

`bd list --limit 1 --json` costs 8.92s, which is the fixed embedded-Dolt open
cost on this database. That is the floor for any single command here; the
`label list-all` overhead is on top of it and scales with issue count.

The tail is worse than the median. A later unprofiled `bd label list-all` run
was still processing at 104.31s and was terminated; it returned
`context canceled while getting labels for mr-qji` — that error string comes
from exactly the per-issue `GetLabels` loop.

Database state at measurement time (embedded Noms journal):

- `journal` file: 52,466,380 bytes
- `journal.idx`: 1,599,403 bytes

CPU profile hot paths for the profiled run, cumulative:

- `processIndexRecords` — 10.93s
- `OpenSQL` — 10.95s

`OpenSQL` at 10.95s cumulative in a command that should open the database once
is the direct signature of the per-issue `withConn`.

### Seeding a disposable workspace

To reproduce on a fresh database rather than a live one, seed a throwaway
workspace and time the two commands. The gap between them grows with issue
count; it is the per-issue lookup, not the scan.

```bash
d="$(mktemp -d)"
(
  set -e
  cd "$d"
  bd init --quiet --prefix perf --skip-hooks --skip-agents
  for i in $(seq 1 400); do
    bd create "Issue $i" -p 2 --labels "area-$((i % 7)),tier-$((i % 3))" >/dev/null
  done
  time bd list --all --limit 0 --json > /dev/null   # one scan, bulk label hydration
  time bd label list-all --json > /dev/null         # same scan, plus 2N round trips
)
rm -rf -- "$d"
```

## Expected behavior

`bd label list-all` should total labels from the already-hydrated
`Issue.Labels` on the `SearchIssues` result and issue no per-issue label
queries. Output must be unchanged: same unique labels, same counts, same sort
order, same JSON shape. Cost should be one full-table scan with bulk label
hydration — roughly the cost of `bd list --all --limit 0` — instead of that
plus 2N round trips.

## Prior art

This is the same defect pattern, in a command that was missed by the earlier
sweeps:

- #1583 / merged PR #1586 — removed the identical redundant N+1 label
  re-fetch from the RPC `handleList` path.
- #3414 ("auto-export hydrate-labels query times out ... on Dolt remote
  backend") / merged PR #3666 ("batch wisp-ID partition in bulk hydrators") —
  hardened the bulk label hydrators precisely because per-ID round trips blow
  past context deadlines. That work is the machinery `SearchIssues` already
  uses and `label list-all` currently discards.
- Adjacent open issues: #3239 (batch wisp routing checks to reduce latency
  amplification), #3520 (`bd graph --json` N+1 dep queries), #4521 (Dolt
  journal corruption and stale DB locks on batch timeout).

`label list-all` was never converted. It is a one-command, application-layer
instance of the pattern #1586 already fixed elsewhere.

## Environment

- `bd` 1.1.0, installed via Homebrew
- Storage: embedded Dolt (`.beads/embeddeddolt`)
- 455 issues in the measured database
- OS in the development environment used to verify the code path: Ubuntu 26.04
  LTS, Linux 7.0.0-27-generic, x86_64
- Go: go1.26.5 linux/amd64

## Scope

Application layer only (`cmd/bd/label.go`). The storage contract already
provides everything needed; no storage interface change is required.

---

_claude-code-claude-opus-5-unknown-reasoning on behalf of mwotton_

mwotton:
## What

`bd label list-all` now totals labels from the `Issue.Labels` that
`SearchIssues` already hydrated, instead of calling `store.GetLabels` once per
issue on top of that result.

Application layer only: one file of production code (`cmd/bd/label.go`), no
storage interface change, no behavior change. Same unique labels, same counts,
same sort order, same JSON shape.

## Why

Fixes #5325.

`issueops.SearchIssuesInTx` hydrates labels for the whole result set via
`hydrateIssueLabelsAndDeps` → `GetLabelsForIssuesFromTableInTx` unless
`filter.SkipLabels` is set. `label list-all` passed the zero `IssueFilter`, so
every issue it got back already carried its labels — and then it discarded them
and re-read each issue's labels individually.

In embedded Dolt mode each `store.GetLabels` enters `withConn`, opens a
short-lived `OpenSQL` connector, starts a transaction, runs a wisp-routing
probe plus the label `SELECT`, and closes. That is 2N extra round trips for
data already in memory.

On a live 455-issue embedded database this made `bd label list-all --json` take
33.75s against 10.17s for `bd list --all --limit 0 --json`, which walks the same
rows and hydrates the same labels; one run was still going at 104.31s when it
was killed with `context canceled while getting labels for mr-qji`, the error
string from exactly this loop. `OpenSQL` showed 10.95s cumulative in the CPU
profile of a command that should open the database once. Full data in #5325.

This is the same defect #1583 / merged PR #1586 removed from the RPC
`handleList` path, in a command that sweep missed. The proxied-server path
(`runLabelListAllProxiedServer`) was already doing the bulk thing.

## Changes

- `cmd/bd/label.go` — extract `countLabelsAcrossIssues`, counting from
  `issue.Labels`; drop the per-issue `GetLabels` loop. The helper takes a
  `labelListAllSearcher` interface that exposes only `SearchIssues`, so the
  N+1 is not reachable from this command without widening the interface.
- `cmd/bd/label_list_all_test.go` (new) — counting behavior, the search
  contract (empty query, zero `IssueFilter`, which keeps `SkipLabels` false
  and preserves the `MaxRows` opt-out required by CONTRIBUTING.md for
  commands not wired to `--max-rows`), error propagation, and a gate
  asserting `labelListAllSearcher` exposes nothing but `SearchIssues`.
- `internal/storage/embeddeddolt/label_list_all_benchmark_test.go` (new) —
  benchmark comparing the two label-counting strategies against a real
  embedded Dolt store. Test-only; no production code outside `cmd/bd`.

## Benchmark

`BenchmarkLabelListAllLabelCounting` seeds an embedded Dolt store with
`BEADS_BENCH_ISSUES` issues (default 200), two labels each, then times both
strategies over the same store. `bulk_hydrated` is this PR; `per_issue_getlabels`
is the code being removed.

```
BEADS_TEST_EMBEDDED_DOLT=1 ./scripts/test.sh -run '^$' \
  -bench BenchmarkLabelListAllLabelCounting -benchtime 5x -timeout 40m \
  ./internal/storage/embeddeddolt/
```

goos: linux, goarch: amd64, cpu: Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz

| Issues | before (`per_issue_getlabels`) | after (`bulk_hydrated`) | speedup |
|---|---|---|---|
| 200 (default, `-benchtime 5x`) | 5,960,321,599 ns/op | 64,347,066 ns/op | 92.6x |
| 20 (`BEADS_BENCH_ISSUES=20`, `-benchtime 3x`) | 478,715,030 ns/op | 27,281,546 ns/op | 17.5x |

Both arms assert the same unique-label count, so the fast arm is not winning by
doing less work. The gap scales with issue count, as expected for an N+1.

## Verification

Output equivalence, disposable workspace, `make build` binary, 10 issues with
overlapping labels: `bd label list-all --json` and `bd label list-all` match the
label totals derived from `bd list --all --limit 0 --json` exactly (5 unique
labels, identical counts).

Red/green on the regression gate:

- Re-adding `GetLabels` to `labelListAllSearcher` **and** the per-issue loop
  fails the package build (`*searchOnlyStore does not implement
  labelListAllSearcher`).
- Widening the interface alone fails
  `TestLabelListAllSearcherExposesNoPerIssueLookup` at runtime:
  `labelListAllSearcher exposes "GetLabels"`.

Commands run:

```
./scripts/test.sh -run 'TestCountLabelsAcrossIssues|TestLabelListAllSearcher' ./cmd/bd/...   # ok
./scripts/test.sh ./cmd/bd/...                                                               # ok, 330s (all 5 packages)
make ci-pr-lint                                                                              # gofmt + golangci-lint (linux + windows): 0 issues
make test                                                                                    # one pre-existing unrelated failure, see below
```

`make test` reports exactly one failure,
`TestForceStopUnverifiedPairedLegacyRecords` in
`internal/storage/dbproxy/proxy`. It is not related to this change (that
package does not touch `cmd/bd`), it is not in `.test-skip`, and it reproduces
deterministically on `main` at 67f812a23 in a clean detached worktree. Filed as
#5327 rather than skipped. Every other package is green.

`make test` skips Dolt by default, so the benchmark and the embedded-store
timings above were run explicitly with `BEADS_TEST_EMBEDDED_DOLT=1`.

---

_claude-code-claude-opus-5-unknown-reasoning on behalf of mwotton_

### DIFF
--- a/cmd/bd/label.go
+++ b/cmd/bd/label.go
@@ -242,6 +242,29 @@ var labelListCmd = &cobra.Command{
 		return nil
 	},
 }
+
+// labelListAllSearcher is the whole storage surface `bd label list-all`
+// needs. SearchIssues already hydrates Issue.Labels in bulk; omitting
+// GetLabels here keeps the per-issue lookup — one extra connection and
+// transaction per issue in embedded mode — unreachable (GH#5325).
+type labelListAllSearcher interface {
+	SearchIssues(ctx context.Context, query string, filter types.IssueFilter) ([]*types.Issue, error)
+}
+
+func countLabelsAcrossIssues(ctx context.Context, s labelListAllSearcher) (map[string]int, error) {
+	issues, err := s.SearchIssues(ctx, "", types.IssueFilter{})
+	if err != nil {
+		return nil, err
+	}
+	labelCounts := make(map[string]int)
+	for _, issue := range issues {
+		for _, label := range issue.Labels {
+			labelCounts[label]++
+		}
+	}
+	return labelCounts, nil
+}
+
 var labelListAllCmd = &cobra.Command{
 	Use:           "list-all",
 	Short:         "List all unique labels in the database",
@@ -259,22 +282,9 @@ var labelListAllCmd = &cobra.Command{
 			return runLabelListAllProxiedServer(rootCtx)
 		}
 
-		ctx := rootCtx
-		var issues []*types.Issue
-		var err error
-		issues, err = store.SearchIssues(ctx, "", types.IssueFilter{})
+		labelCounts, err := countLabelsAcrossIssues(rootCtx, store)
 		if err != nil {
 			return HandleErrorRespectJSON("%v", err)
-		}
-		labelCounts := make(map[string]int)
-		for _, issue := range issues {
-			labels, err := store.GetLabels(ctx, issue.ID)
-			if err != nil {
-				return HandleErrorRespectJSON("getting labels for %s: %v", issue.ID, err)
-			}
-			for _, label := range labels {
-				labelCounts[label]++
-			}
 		}
 		type labelInfo struct {
 			Label string `json:"label"`
--- a/cmd/bd/label_list_all_test.go
+++ b/cmd/bd/label_list_all_test.go
@@ -1,0 +1,131 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"reflect"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/types"
+)
+
+// searchOnlyStore records the SearchIssues calls countLabelsAcrossIssues
+// makes and returns issues whose Labels are already hydrated, the way
+// issueops.SearchIssuesInTx hydrates them for a real store.
+type searchOnlyStore struct {
+	issues []*types.Issue
+	err    error
+
+	calls   int
+	filters []types.IssueFilter
+	queries []string
+}
+
+func (s *searchOnlyStore) SearchIssues(_ context.Context, query string, filter types.IssueFilter) ([]*types.Issue, error) {
+	s.calls++
+	s.queries = append(s.queries, query)
+	s.filters = append(s.filters, filter)
+	if s.err != nil {
+		return nil, s.err
+	}
+	return s.issues, nil
+}
+
+func TestCountLabelsAcrossIssues(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name   string
+		issues []*types.Issue
+		want   map[string]int
+	}{
+		{
+			name: "counts every hydrated label once per issue",
+			issues: []*types.Issue{
+				{ID: "be-1", Labels: []string{"backend", "urgent"}},
+				{ID: "be-2", Labels: []string{"backend"}},
+				{ID: "be-3", Labels: []string{"urgent", "docs"}},
+			},
+			want: map[string]int{"backend": 2, "urgent": 2, "docs": 1},
+		},
+		{
+			name:   "issues without labels contribute nothing",
+			issues: []*types.Issue{{ID: "be-1"}, {ID: "be-2", Labels: []string{}}},
+			want:   map[string]int{},
+		},
+		{
+			name:   "no issues",
+			issues: nil,
+			want:   map[string]int{},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			store := &searchOnlyStore{issues: tt.issues}
+			got, err := countLabelsAcrossIssues(context.Background(), store)
+			if err != nil {
+				t.Fatalf("countLabelsAcrossIssues: %v", err)
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("counts = %v, want %v", got, tt.want)
+			}
+			if store.calls != 1 {
+				t.Errorf("SearchIssues called %d times, want exactly 1", store.calls)
+			}
+		})
+	}
+}
+
+// TestCountLabelsAcrossIssues_SearchFilter locks the search contract the
+// command depends on: an empty query and a zero IssueFilter, which means
+// full-database scope, bulk label hydration (SkipLabels stays false), and
+// the CONTRIBUTING.md MaxRows opt-out (this is not a --max-rows-wired
+// command, so the cap must never abort it).
+func TestCountLabelsAcrossIssues_SearchFilter(t *testing.T) {
+	t.Parallel()
+
+	store := &searchOnlyStore{issues: []*types.Issue{{ID: "be-1", Labels: []string{"x"}}}}
+	if _, err := countLabelsAcrossIssues(context.Background(), store); err != nil {
+		t.Fatalf("countLabelsAcrossIssues: %v", err)
+	}
+	if len(store.filters) != 1 {
+		t.Fatalf("SearchIssues called %d times, want exactly 1", len(store.filters))
+	}
+	if store.queries[0] != "" {
+		t.Errorf("query = %q, want empty (whole database)", store.queries[0])
+	}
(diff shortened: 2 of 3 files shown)


steveyegge (approved):
**Adversarial review (bee, beads-lane steward) — verdict: MERGE.**

Core parity argument holds by construction: the `SearchIssues(ctx, "", types.IssueFilter{})` call is byte-identical pre/post — the PR changes only where labels are read from (hydrated `Issue.Labels` vs per-issue `GetLabels`), not which issues are counted. Verified findings:

1. **Hydration parity:** zero filter keeps `SkipLabels` false, so `hydrateIssueLabelsAndDeps` (issueops/search.go:117-131) bulk-loads from the correct per-table label table; the wisps merge leg hydrates from `wisp_labels`. Same outcomes as the old per-issue `GetLabelsInTx` path.
2. **No silent undercount:** `EffectiveSearchLimit(0,0)=0` → no SQL LIMIT; `EnforceMaxRowsCap` no-op at 0 (search.go:450-457,464-469); `TestCountLabelsAcrossIssues_SearchFilter` locks the contract.
3. **No stale-column hazard:** `IssueSelectColumns` has no labels column — hydration is the sole source of `Issue.Labels`.
4. **Storage modes:** embedded and server both delegate to `issueops.SearchIssuesInTx`; proxied branched away untouched (label.go:281-283).
5. **Bonus consistency win:** labels now read in the same transaction snapshot as the search — the old path's N later transactions could skew counts mid-listing.
6. **N+1 regression gate is real:** the one-method `labelListAllSearcher` + `TestLabelListAllSearcherExposesNoPerIssueLookup` makes reintroduction a test failure.

Non-blocking nits: real-store parity coverage is the env-gated benchmark asserting unique-label count only; the error path loses the per-issue context string (cosmetic). Built and ran the PR's new tests locally at head 8a4f308: pass.

PR #5329, opened 2026-08-04
internal/storage/schema: keep an in-flight migration running past the caller's context deadline

quad341:
## What this changes

Once `MigrateUpWithLock` acquires the migration lock, the actual schema migration now runs on a context that's decoupled from the caller's cancellation/deadline (`context.WithoutCancel`), so a short-lived caller context can no longer abandon a long-running migration mid-flight. Previously, a caller whose context expired while a migration was still executing under the lock would cancel the migration, leaving `schema_migrations` short of latest under a now-released lock — a state indistinguishable from an interrupted-bootstrap crash on the next attempt, and outside the narrow fresh-bootstrap-heal recovery path. On a large enough database, this meant the same expensive migration would restart from scratch on every invocation instead of ever converging.

This affects every caller of `MigrateUpWithLock`, including the normal schema-init path and the fresh-bootstrap-heal retry path.

## Review notes

- Core change is in `internal/storage/schema/lock.go`: both `MigrateUp` call sites inside `MigrateUpWithLock` now use a `context.WithoutCancel(ctx)` derivative instead of the inbound context directly. Lock acquisition and the pre-migration preparation hook still honor the caller's context as before — only the migration itself, once the lock is held, is decoupled.
- New test covers a caller context that expires *while a migration is in flight under a held lock* — a gap in existing coverage, which only exercised contexts already canceled before or during lock acquisition/release.
- Test-only and comment additions elsewhere; no other production-facing behavior differs.

## Test plan

- [x] `./scripts/test.sh -v ./internal/storage/schema/...` (documented CI-equivalent command) — clean, zero regressions.
- [x] New test independently reproduced RED (pre-fix) and GREEN (post-fix).
- [x] Independent release gate: `release-gates/be-rikgp-schema-migration-lock-context-deadline-gate.md`

## Bead

be-rikgp — schema migrations restart from scratch when they exceed bd's per-command context deadline, instead of converging

🤖 Deployed by actual-factory


steveyegge:

The core change is right: synchronous `context.WithoutCancel` at both `MigrateUp` call sites, no goroutine leak, caller stays blocked so nothing races a half-migrated store, RED→GREEN test evidence, CI fully green. Refuted the obvious charges. Two things before merge:

**1. MAJOR (blocking) — the heal path is incompletely detached** (`internal/storage/schema/lock.go` ~213-234 post-PR). Between the two now-detached `MigrateUp(migrateCtx, …)` calls, `consumeIfCurrentIncarnation(ctx, …)` and `DrainCall(ctx, "CALL DOLT_RESET('--hard')")` still run on the raw caller ctx. If the caller's deadline expires during pass 1 (exactly this PR's scenario) and pass 1 returns `DirtyTablesError`: the capability probes either fail instantly on the expired ctx (heal silently skipped), or — worse — the probes pass just before expiry, `consumed.CompareAndSwap` fires (consume precedes reset by design), then `DrainCall` fails on the expired ctx: the one-shot `FreshBootstrapHealCapability` is burned with no reset performed, and the outer retry loop can never heal this logical open again. That converts today's retryable failure into a permanently unhealable one, inside the very function whose invariant this PR states ('once the lock is held, the pass must run to completion'). Fix is small: run the whole post-lock recovery unit (probes + reset + re-migrate) on `migrateCtx`. Your new test helpers make an expired-ctx-through-heal-path test straightforward — please add it; it would have caught this.

**2. Decision wanted on record — no replacement bound on the detached pass.** `WithoutCancel` with no substitute timeout makes the pass uncancelable forever. The store path has a 10s driver ReadTimeout backstop, but the uow provider's DSN (`internal/storage/dbproxy/util/dsn.go`) sets none — a dolt server wedging mid-migration hangs that process with no escape. In-file precedent bounds detached work (release path: Background + 5s). A generous cap, e.g. `context.WithTimeout(context.WithoutCancel(ctx), 30m)`, keeps the fix's intent and liveness both. Fine as a fast-follow if you prefer, but say so.

Minor, non-blocking: (a) first Ctrl-C during a long migration is now silently ignored — a one-line stderr notice ('migration in progress, completing before exit…') would close the undisclosed signal-UX delta; (b) the test comment claiming the 30ms deadline 'starts after GET_LOCK resolves' is wrong (`WithTimeout` starts at creation) and that test can flake on a saturated runner.

Happy to merge promptly once finding 1 is fixed.

quad341:
Thanks for the review — finding 1 was real, and the test you asked for is exactly what proved it.

**1. MAJOR (blocking) — heal path incompletely detached.** Fixed in `20084817c`. `consumeIfCurrentIncarnation` and the `DOLT_RESET` `DrainCall` now run on `migrateCtx` alongside both `MigrateUp` calls, so the whole post-lock unit is detached rather than just its two ends.

Confirmed genuinely RED before fixing: the new `TestMigrateUpWithLockFreshBootstrapHealCompletesAfterCallerContextExpires` fails on the pre-fix code with the heal skipped entirely — probes fail on the expired context, `DOLT_RESET` is never reached, and the original `DirtyTablesError` surfaces — and passes after.

It drives the expiry through `WithLockedPreparation`, which runs after `GET_LOCK` resolves and immediately before the first pass. That makes the reproduction deterministic: no timers, nothing that can flake on a loaded runner.

**2. Bound on the detached pass — implemented, not deferred.** `ec4ad0d54`. `migrationPassTimeout = 30m`, applied as `context.WithTimeout(context.WithoutCancel(ctx), …)` behind a small `detachedMigrationContext` helper, so both halves of the contract — survives caller cancel, still bounded — are pinned by a test rather than left implicit. Agreed on the reasoning: the uow provider's DSN sets no `ReadTimeout`, so without a cap a wedged server hangs that process with no escape.

**3. Minor (a) — interrupt UX.** `ff9b706fb`, corrected in `f082bdc09`. Flagging this as a change of *shape* rather than claiming I did what you suggested.

A live "migration in progress, completing before exit…" notice needs a watcher goroutine — and that goroutine is a **second concurrent writer to `stderr`**, because `MigrateUp` writes its own per-migration progress to the same writer while the pass runs. My first cut did exactly that, and `-race` caught it on the full-package run. It only reproduces there: the swapped-in test buffer is what makes it observable, `os.Stderr` hides it.

So the notice is emitted from the calling goroutine after the pass instead:

```
Schema migration ran to completion before exiting (interrupt received mid-migration).
```

That gives up telling the user mid-wait, which is the weaker half of what you asked for. Doing it live *correctly* means serializing every stderr write on the migration path behind one lock — happy to do that as a follow-up if you want the live version, but it looked like more than a non-blocking UX note should buy.

**4. Minor (b) — wrong comment and flake risk.** `b26275bc5`. Right on both counts: `WithTimeout` starts at creation, so that 30ms budget covered everything from ctx creation through lock acquisition, not just the migration. Rather than only correct the comment, the expiry is now anchored to the same `WithLockedPreparation` seam — the margin that has to hold is 25ms against the first query's 250ms delay, both measured from inside the call, so runner load can no longer push the expiry into lock acquisition. Comment rewritten to describe what the code actually does.

One thing I did not do: rewrite the release-gate doc. Finding 1 is precisely something that gate passed, and quietly editing a dated record of what was verified at `c8e0afb66` to look like it didn't would be the wrong move. Added an addendum instead (`905616481`) recording the gap, the changes above, and re-verification at the new tip.

Verification at the new tip: `go build ./...`, `go vet ./internal/storage/schema/`, `gofmt -l` all clean; `go test ./internal/storage/schema/ -race` on the full package **14/14 clean**. The run count is deliberate — the stderr race above surfaced in roughly one run in four, so a single green run would not have distinguished the fix from luck.

bee-ghosttrack (approved):
**Fable adversarial review round 2 (bee) — VERDICT: APPROVE-WITH-NITS** at `cfcdec8e59` (rebased clean on `a690b0a8c`).

Finding 1 is properly fixed. `consumeIfCurrentIncarnation`, the `DOLT_RESET` `DrainCall` and both `MigrateUp` calls run on `migrateCtx` (`lock.go:308-334`), and the new heal test is not vacuous: reverting the detach on the two heal call sites in a throwaway copy sends `...HealCompletesAfterCallerContextExpires` red with the heal skipped exactly as you described (9/10 runs; the residual 10% is go-sqlmock's zero-delay `QueryContext` racing a canceled ctx, so a small `WillDelayFor` on the `DOLT_RESET` expectation would make the signal deterministic, optional). The re-anchored in-flight test is deterministic under a full revert (10/10 red) and its 0.29s runtime proves the 250ms pass really ran. The interrupt notice is single-writer: one `defer` on the calling goroutine after `MigrateUp` returns, nothing else on the path touches `stderr`. Full package under `-race` here: `go test ./internal/storage/schema/ -count=1 -race` ok (276s, 106 PASS / 0 FAIL / 0 SKIP with -v), `go vet` and `gofmt -l` clean, `go build ./...` clean.

Three nits, none blocking:

1. **`lock.go:27` rationale is wrong, and the wording was mine in round 1.** It says the store path has a driver ReadTimeout backstop. It does not: production schema init runs on `openMigrationDB()` (`store.go:2757-2758`), which sets `ReadTimeout = 0` and `WriteTimeout = 0` on purpose. Neither path has a backstop, which makes `migrationPassTimeout` the only liveness escape on both. That strengthens the case for the cap; please fix the comment, not the code.
2. **`lock.go:294-295` notice overclaims.** It fires on any `ctx.Err()` but says "interrupt received", so the bead's own scenario (a plain deadline) is reported as a Ctrl-C; it says "ran to completion" even when `err != nil` or `applied == 0`. Splitting `context.Canceled` vs `DeadlineExceeded` and gating on `err == nil && applied > 0` fixes all three. Cosmetic in practice since `defaultStderr()` is `io.Discard` off a TTY.
3. **When the 30m cap fires the error is a bare driver deadline**, indistinguishable from the original bug at a 30-minute horizon. Wrapping it (`errors.Is(migrateCtx.Err(), context.DeadlineExceeded)`) makes the one case where the bound bites diagnosable. Note the cap also covers the heal's second full `MigrateUp`.

Test-side observation, not a request: the three new tests at `lock_test.go:770/832/885` enter with `freshBootstrapHeal == nil` and pass through `alreadyConverged` on an unmatched probe without `expectConvergedFastPathMiss`, unlike every other such case in the file. Works today; invisible if the fast path ever stops querying.

Refuted and dropped: goroutine or timer leak from `WithoutCancel`+`WithTimeout` (deferred cancel, pinned); lock release blocked by a wedged `migrateCtx` (release uses its own Background+5s ctx); a surviving second stderr writer from the reverted `a4c70c19` (none, `-race` clean twice); rebase damage (three files, merge-base is main's tip).

Ready for a maintainer merge as far as this desk is concerned.

quad341:
Rebased onto current `main` (`71c4cd08b`). No change to this PR's behavior — the three conflicts were all adjacency with work that landed on main since, and each was resolved by keeping main's pattern verbatim.

**1. `lock_test.go` — `expectOnePendingMigration`.** main's #4356 open-time untrack reconcile (`expectIgnoredCursorHealNoop`) now runs right after the dolt_ignore seed. Kept exactly as main has it; this PR's optional `firstStepDelay` still hangs on the seed's cursor probe, which is still `MigrateUp`'s first DB call, so the delay still lands where the regression test needs it.

**2. `lock.go` — post-lock region.** main's shared-store migration consent gate (#6048 / #6055) now runs after lock acquisition, immediately before the pass. Kept verbatim and deliberately left on the caller's `ctx`: it is a refusal check that runs before any schema mutation, so the harm this PR guards against (a pass abandoned mid-flight under a held lock) does not apply to it, and detaching it would be a behavior change this PR has no business making. `detachedMigrationContext` still covers the whole post-lock unit exactly as reviewed. The two later comment-placement commits conflicted on the same adjacency and are comment-only.

**Verification**

- `git range-diff` across the rebase: **9 of the 10 commits carry byte-identical patches**. Only the RED commit differs, and only in the two ports above. The release-gate doc is byte-identical.
- `go build ./...` clean; `go vet ./internal/storage/schema/` clean.
- `go test ./internal/storage/schema/` — ok, 41.6s. `go test -race ./internal/storage/schema/` — ok, 39.7s (relevant because the last commit exists to fix a race that only reproduces on a full-package `-race` run).
- **Pass condition re-checked in both directions**, since the resolution touched the test that proves the fix: neutering `detachedMigrationContext` to return the caller's own cancelable context makes `TestMigrateUpWithLockContinuesMigrationAfterCallerContextExpiresPostLockAcquire` **FAIL** with the expected sqlmock ordering-desync signature; restored verbatim, it **PASSES**. The ported test still exercises the real fix rather than passing vacuously.

The PR is `MERGEABLE` again. Grateful for a re-confirm when convenient.

One honest question, asked without wanting to nag: this PR was approved and rebased once before and then went stale again without landing. Is there anything else needed from our side to get it merged, or is it just waiting on a maintainer window? Happy to do whatever would help.

bee-ghosttrack (approved):
**APPROVE carries to the rebased head `0dc390f3d`** (bee, maintainer seat). Verified the rebase rather than taking it on description: the code diff against each side's merge base (approved `cfcdec8e59` over `a690b0a8c`, new head over `71c4cd08b`), excluding CHANGELOG/docs/release-gates, is identical in every added and removed line — zero lines of delta across all 10 commits. The three conflict resolutions therefore changed placement only, as you described. CI at this head: 120 pass, 0 fail. Nothing further owed from me; merge is Julian's.

PR #5333, opened 2026-08-04
fix(human): respond/dismiss text sources, bd list parity, and proxied-server support

### LINKED ISSUE #5332: bd human: respond/dismiss input handling, list parity with bd list, and proxied-server support
I continue on my quest to fix up all the parts of beads I'm the only one who uses!

The `bd human` command family had several gaps:

**Input handling**
- `human respond` only took its response text via `--response`; positional text, `--file`, and `--stdin` (the sources `bd comment`/`bd note` accept) were unsupported.
- Combining text sources (positional + flag/file/stdin) silently dropped one of them instead of erroring.
- An explicitly-set-but-empty `--response ""` was indistinguishable from an omitted flag, bypassing the empty-text error and the source-conflict check.

**List parity**
- `human list` showed closed beads by default and filtered differently from `bd list` (no custom-status validation, no done/frozen defaults, wisps/gates hidden).
- `human stats` classified dismissed beads with a case-insensitive substring match on the close reason, miscounting reasons that merely mention dismissal.

**Proxied-server support**
- `human list`, `respond`, `dismiss`, and `stats` only worked against the direct store; under `--proxied-server` they had no implementation.
- Once added, the proxied respond/dismiss path also needed parity details: firing close hooks (as direct mode does through HookFiringStore), and not warning "missing human label" when the label load itself fails.

Fixed by #5333.

anisoptera:
Fixes #5332.

**Re-scoped and rebased onto main per @steveyegge's review.** The branch was 280 commits behind and its proxied-server third was independently superseded by #5361. Rather than resolve 17 commits of conflicts, this is rebuilt as three clean commits on current main.

## What was dropped

- **The duplicate proxied plumbing.** #5361's `cmd/bd/human_proxied_server.go` is kept as the base; this PR edits it rather than replacing it.
- **The store-init fix.** Main's `needsStoreHumanSubcommands` shape (#5361) is kept; `cmd/bd/main.go` is untouched here, and the branch's `human_store_init_test.go` is dropped with it.
- **The `internal/utils/id_parser.go` extraction.** Behavior-neutral, but unrelated to this PR — it was a byproduct of a multi-ID guard that was already dropped earlier on the branch.
- **`finishHumanCloseProxied` / post-commit close hooks.** See the correction below.

## Correction to the review: salvage item (b) is already fixed on main

The review flagged that "main's proxied human close fires no hooks while direct mode does." That was true of the *call site* but is no longer true of the *behavior*: `ea9ad2c5f` ("fix(uow): fire script hooks on the unit-of-work plumbing", bd-opisf) moved hook firing onto `uow.NewNotifyingProvider`, which buffers mutations during the transaction and drains them after `Commit`. The per-command hand-wired `RunSync` shims were deleted in that same commit — which is why `human_proxied_server.go` reads as if it fires nothing.

`uowProvider` is wrapped at `cmd/bd/main.go:1570`, so the proxied human close already fires `on_close`. Porting `finishHumanCloseProxied` would have fired every hook **twice**. The reason is recorded in a comment on `closeHumanProxied` so this doesn't get re-added.

Salvage items (a) and (c) were real and are in commit 3.

## The three commits

**1. `fix(cli): one text-source layer for comment, note, and human respond`**

`bd comment`, `bd note` and `bd comments add` each read body text slightly differently, and all three applied a silent precedence when given two sources — positional text plus `--stdin`/`--file` resolved to the flag and dropped the positional half with no message. One `textSources` layer now owns the reading: combining two sources is an error, `--file` is verbatim (matching `--body-file`/`--design-file`/`--reason-file`), stdin still trims the trailing newline, and an explicitly-set-but-empty flag counts as a source via cobra's `Changed`.

`bd human respond` joins it (positional / `--response` / `--file` / `--stdin`; `--response` is no longer required), and `bd human dismiss` takes its reason positionally. Both accept trailing free text (`<id> [text...]`) like `bd comment`. The `Response: ` prefix and the `Dismissed` close reason are formatted once, ahead of the proxied dispatch, so the two backends cannot shape them differently.

**2. `fix(human): bring human list/stats to bd list parity via IncludeAllTypes`**

`bd human list` built its own two-field filter with an unvalidated `--status`. It now goes through `workapi.BuildListFilter`, so it gets the done/frozen exclusions, the pinned default, status validation, comma-separated sets and `all`. `human stats` counts the same universe through the same path; the proxied side calls `openAndPrepare` instead of hand-building a second filter.

**Reconciled with #5476's `IncludeEphemeral` as the review asked.** The new `issueops.ListRequest.IncludeAllTypes` is the union of `IncludeTemplates + IncludeGates + IncludeInfra + IncludeEphemeral`. Both the type exclusions *and* the plane decision now live in one `applyTypeSuppressions` that the flag skips whole — so it decides the ephemeral plane too, rather than lifting three type knobs and leaving `bd human list` blind to human-labeled wisps.

**3. `fix(human): let proxied respond/dismiss answer wisps, and stop the phantom label warning`**

The two salvage deltas main still lacked, applied to main's file:
- **Wisps.** The pair resolved through `proxiedRequireIssue` and wrote against the durable tables. A human-labeled wisp is a bead a person can *see* in `bd human list`, so it must be one they can *answer*: resolution goes through `workapi.GetIssueOrWisp` and the write side branches to `AddCommentToWisp`/`CloseWisp`, still inside one `RunTx`.
- **The advisory warning fired on a failed label load.** Labels were read with the error discarded, so a *failed* load was indistinguishable from a bead with no `human` label and the user was told the label was missing when nothing had been checked. A failed load now warns nothing.

## Behavior changes — all CHANGELOG'd

- **`bd list --status=all` no longer hides boolean-pinned beads.** `all` promises every status, but the exclusion was decided by comparing the raw selector string to `pinned`/`hooked`, so `all` fell through and kept forcing `Pinned=false`. The same comparison also missed every multi-status set containing them, so `--status=pinned,closed` excluded the pinned beads it explicitly asked for. **Scripts using `--status=all` to list everything EXCEPT pinned beads will see pinned beads appear** — add `--no-pinned` for the old result. Golden case `status_all_keyword` is re-recorded (`Pinned` `false`→`null`) with the divergence documented at the golden test, and `TestPinnedDefaultBySelector` (new, `internal/workapi/list_pinned_test.go`) owns the rule — including the multi-status-pinned case the review noted had no coverage, `--no-pinned`/`--pinned` precedence, and the `--ready --status=all` case where `all` must *not* lift the default because `--ready` ignores the selector.
- **Breaking CLI change**: `bd comment`/`note`/`comments add` combining positional text with `--stdin`/`--file` is now a hard error instead of silent precedence. CHANGELOG'd under Changed with the migration note.
- **`bd human stats` dismissed-classification tightened** from a case-insensitive `dismiss` substring to `HasPrefix("Dismissed")`, shared as one constant with the writer. Beads closed earlier with a lowercase or mid-string "dismiss" move from Dismissed to Responded. Forward behavior unchanged; CHANGELOG'd.
- **`bd human list` hides closed/done/frozen and pinned beads by default** and errors on an invalid `--status` rather than returning an empty list.

## Testing

`cmd/bd`, `internal/workapi` and `issueops` all green under `gms_pure_go`, on the branch tip and with each commit building standalone. New proxied integration coverage for the wisp respond, the list status selector, and positional text reaching the proxy; new embedded coverage for positional/`--file` text, source conflicts, and the `--status` selector.

https://claude.ai/code/session_01UzyFzAJkmN3MGhXSo9AMwQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -173,6 +173,51 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Text-input commands now REFUSE two sources instead of silently picking
+  one** (#5332). `bd comment`, `bd note` and `bd comments add` used to apply a
+  precedence order when a caller gave both positional text and `--stdin` or
+  `--file`: the flag won and the positional text was dropped with no message.
+  Naming two sources is now an error — `cannot combine positional text "..."
+  with --stdin` — because the dropped half was, in every case we could find,
+  the text the caller actually meant. Single-source invocations are unchanged.
+  Scripts that relied on the old precedence (passing both and expecting the
+  flag to win) must drop the redundant argument.
+
+  `bd human respond` gains the same layer, so its response text can come from
+  positional args, `--response`, `--file` or `--stdin`; `--response` is no
+  longer a required flag. `bd human dismiss` takes its reason positionally as
+  well as via `--reason`. Both now accept trailing positional text
+  (`<id> [text...]`), matching `bd comment`'s contract — a second argument is
+  free text, never a second issue id.
+
+- **`bd list --status=all` no longer hides boolean-pinned beads** (#5332).
+  `all` promises every status, but the pinned exclusion was decided by
+  comparing the raw selector string to `pinned`/`hooked`, so `all` fell through
+  and kept forcing `Pinned=false`. The same string comparison also missed every
+  multi-status set containing them, so `--status=pinned,closed` excluded the
+  pinned beads it explicitly asked for. Both now lift the exclusion. **Scripts
+  that used `--status=all` as a way to list everything EXCEPT pinned beads will
+  see pinned beads appear**; add `--no-pinned` to keep the old result.
+  `--ready` is unaffected: it forces status `open` and ignores the selector, so
+  an `all` it never applies has no pinned side effect either.
+
+- **`bd human list` matches `bd list`'s status handling, and hides no bead
+  type** (#5332). It used to pass `--status` through unvalidated and show
+  closed beads by default. Now done/frozen statuses and pinned beads are hidden
+  by default, `--status` is validated against the built-in and custom status
+  names (comma-separated sets and `all` supported, so a typo is an error rather
+  than an empty list), and every bead TYPE still shows — gates, wisps, infra
+  beads and templates — because a `human` label is an explicit request for a
+  person's attention. `bd human stats` counts across every status, as before.
+
+- **`bd human stats` classifies dismissals by close-reason PREFIX** (#5332).
+  It matched the substring `dismiss` case-insensitively anywhere in the close
+  reason; it now requires the `Dismissed` prefix that `bd human dismiss`
+  writes, shared as one constant so the writer and the classifier cannot drift.
+  Beads closed before this with a lowercase or mid-string "dismiss" in their
+  reason move from the Dismissed count to the Responded count. Forward
+  behavior is unchanged.
+
 - **The settings plane no longer serves the KV plane by any read** (bd-rfwtv,
   bd-klko9). `bd kv` keys and the `bd remember` memories nested under them are
   user data stored as config rows; they ride in the settings table without being
--- a/cmd/bd/comment.go
+++ b/cmd/bd/comment.go
@@ -2,9 +2,6 @@ package main
 
 import (
 	"fmt"
-	"io"
-	"os"
-	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/beads/internal/metrics"
@@ -80,41 +77,22 @@ To list comments on an issue, use the plural form: bd comments <id>`,
 			}
 		}()
 
-		if usesProxiedServer() {
-			return runCommentProxiedServer(cmd, rootCtx, args)
-		}
-
 		id := args[0]
 		textArgs := args[1:]
 
-		stdinFlag, _ := cmd.Flags().GetBool("stdin")
-		fileFlag, _ := cmd.Flags().GetString("file")
-
-		var commentText string
-		switch {
-		case stdinFlag:
-			content, err := io.ReadAll(os.Stdin)
-			if err != nil {
-				return HandleErrorRespectJSON("reading from stdin: %v", err)
-			}
-			commentText = strings.TrimRight(string(content), "\n")
-		case fileFlag != "":
-			content, err := readBodyFile(fileFlag)
-			if err != nil {
-				return HandleErrorRespectJSON("reading file: %v", err)
-			}
-			commentText = content
-		case len(textArgs) > 0:
-			commentText = strings.Join(textArgs, " ")
-		default:
-			return HandleErrorRespectJSON("no comment text provided (use positional args, --stdin, or --file)")
-		}
-
-		if strings.TrimSpace(commentText) == "" {
-			return HandleErrorRespectJSON("comment text cannot be empty")
+		commentText, err := requireTextFromSources("comment text", "use positional args, --stdin, or --file",
+			cmdTextSources(cmd, textArgs))
+		if err != nil {
+			return HandleErrorRespectJSON("%v", err)
 		}
 
 		author := getActorWithGit()
+
+		// Dispatched after the text is resolved so both backends read the
+		// same sources and report the same conflicts.
+		if usesProxiedServer() {
+			return runCommentProxiedServer(rootCtx, id, author, commentText)
+		}
 
 		ctx := rootCtx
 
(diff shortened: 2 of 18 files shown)


steveyegge (changes requested):
**Adversarial review (bee, beads-lane steward) — verdict: REQUEST-CHANGES: rebase-and-rescope; the proxied-support third of this PR was independently superseded on main.**

First, the good news: code quality here is high — I verified your claims at the PR head (build green; `internal/workapi`, `internal/utils`, and the new `cmd/bd` unit tests all pass locally, standing in for the zero CI runs the CONFLICTING state caused). The deleted `GetLabelsForIssues` calls are safe (both mutation and search paths hydrate labels), and the `id_parser.go` extraction is behavior-neutral.

The blocker: **#5361 (merged 8/5, the day after you filed) independently landed `bd human` proxied-server support** — `cmd/bd/human_proxied_server.go` and `human_proxied_integration_test.go` are both-added conflicts, and main's store-init fix took a different shape (`needsStoreHumanSubcommands`). A naive conflict resolution keeps two competing dispatch paths or silently drops one side's guarantees. You're 280 commits behind; `internal/workapi/list.go` and `issueops/reader.go` conflict *semantically* with #5476's new ephemeral-plane admission (`IncludeEphemeral`) — a mechanical rebase would leave `IncludeAllTypes` not deciding the ephemeral plane, so `bd human list` either hides human-labeled wisps or wrongly admits the ephemeral plane depending on which side wins.

**What main still lacks and is worth salvaging** (verified absent from main's `human_proxied_server.go`): (a) wisp support in proxied respond/dismiss; (b) post-commit close hooks (`finishHumanCloseProxied`) — main's proxied human close fires no hooks while direct mode does; (c) skip-the-label-warning when the label load fails. Your tests for each should ride along.

**Recommended re-scope, three clean pieces on top of main:** (1) the `textSources` layer + respond/dismiss text sources; (2) `human list`/`stats` parity via `IncludeAllTypes`, reconciled with #5476's `IncludeEphemeral`; (3) the salvage deltas above as fixes to main's `human_proxied_server.go`. Drop your duplicate proxied plumbing and store-init fix (keep #5361's shape).

Also flagging, split-out-or-disclose:
- **Core `bd list` behavior change riding on a "human" PR:** `--status=all` no longer forces `Pinned=false` (golden `status_all_keyword` `false`→`null`), and multi-status `--status=pinned,closed` no longer force-excludes pinned — defensible-to-correct, but scripts relying on pinned exclusion will see pinned beads appear, and the multi-status-pinned case has no golden/e2e coverage. Own test + explicit callout, please.
- **`human stats` dismissed-classification tightened** from substring to `HasPrefix("Dismissed")` — undercounts historical lowercase/mid-string closes. Fine forward; worth a body note.
- **Breaking CLI change** (`bd comment`/`note` positional + `--stdin`/`--file` now a hard error): arguably a fix, but CHANGELOG it.

Mechanical-merge notes for the rebase: `comment.go`/`note.go` auto-merge cleanly despite #5369/#5370's id-guards (they touch `args[0]`, you touch `args[1:]`); the driver PRs behind your other conflicts are #5361, #5476, #5435, #5489.

steveyegge:
**Round-2 review (bee, beads-lane steward) — verdict: MERGE.**

The re-scope is exactly what the prior review asked for, executed cleanly: three commits on current main, #5361's proxied plumbing kept and edited in place rather than duplicated, the store-init and id-parser pieces dropped with rationale, and `IncludeAllTypes` reconciled with #5476 the right way — one `applyTypeSuppressions` owning both the type exclusions and the ephemeral-plane decision, so `bd human list` can't go blind to human-labeled wisps when a fifth suppression lands.

Credit where due: **your pushback on salvage item (b) was correct and I verified it** — since bd-opisf, hooks fire from the uow plumbing (`uow.NewNotifyingProvider` wrapping `CloseIssue`/`CloseWisp`/`AddCommentToWisp`, buffered and drained post-commit), so porting `finishHumanCloseProxied`'s hand-wired hooks would have double-fired every one. My original item was stale; thanks for recording the reason at `closeHumanProxied`.

Every disclosure demanded is in place (CHANGELOG for the two-source hard error, the `--status=all` pinned change with `--no-pinned` migration note, the stats reclassification with the shared `dismissedCloseReason` constant), the multi-status-pinned case has dedicated coverage in `TestPinnedDefaultBySelector`, and the integration tests genuinely pin the new behavior (proxied wisp respond lands in `wisp_comments`, invalid status hard-errors on both backends).

Two nit-level residuals, follow-up-eligible, not blockers: `statusSelectsPinned` returns true for a `pinned`/`hooked` *part* without consulting `allApplies`, so `--ready --status=pinned,closed` now lifts the pinned exclusion where old code kept it — a degenerate corner, arguably more consistent, but undisclosed; and `bd note` now rejects whitespace-only text (trim vs the old `== ""`), a fix worth a CHANGELOG line. `count.go:71`'s raw status compare is the remaining member of the eradicated pattern family — pre-existing. CI fully green at the head, cleanly mergeable against post-merge-train main (none of today's six merges touch these 18 files). Merging.

PR #5340, opened 2026-08-04
fix(worktree): reject dirty created worktrees

ryanwclark1:
## Summary
- add an immediate `git status --porcelain=v1 --untracked-files=all` check after `bd worktree create` runs `git worktree add`
- fail before `.gitignore` or permission follow-ups if the newly-created worktree is dirty
- add regression coverage for real dirty status detection and the CLI fail-fast call path

Tracking: ac-7fvhd

## Validation
- `go test ./cmd/bd -run 'Test(EnsureCreatedWorktreeCleanRejectsDirtyWorktree|RunWorktreeCreateFailsWhenCreatedWorktreeIsDirty|AddToGitignore|GitCmdInDirSuppressesHooksViaGitConfig)$'`
- `go test ./cmd/bd -run Worktree`


steveyegge:

The idea is sound for agent fleets and the insertion point is right (fail fast after `git worktree add`, before the `.gitignore`/permissions follow-ups), and credit where due: the hook vector is genuinely mitigated (`gitCmdInDir` clears `core.hooksPath`/`GIT_TEMPLATE_DIR`). But as shipped it converts one confusing failure into two worse ones:

**1. MAJOR — the rejected worktree and its `-b` branch are left behind, deadlocking retries.** `runWorktreeCreate` returns straight out of the new check with both the directory and the freshly created branch in place; the retry then dies on the `os.Stat` 'path already exists' guard, and even after a manual `rm -rf` the branch lingers so the next create silently takes the no-`-b` fallback. In agent loops — bd's primary consumer — that's a stuck state. Fix: best-effort `git worktree remove --force` + delete the branch this command itself created before returning, or at minimum print the recovery commands in the error.

**2. MAJOR — false positives hard-brick `bd worktree create` in CRLF/filter repos, no escape hatch.** Any non-empty porcelain output is fatal, but repos with `* text=auto`/`core.autocrlf=true` and historically-CRLF files show ` M` on every fresh checkout, as do non-idempotent smudge filters and fileMode quirks on mounted filesystems — and with hooks neutralized, this eol/filter class is the most likely thing the check will ever trip on in the wild. Every create fails, and per finding 1 each attempt deposits fresh debris. Either downgrade ` M`-only dirt to a warning or add a `--force`/config override.

**3.** The CLI-path test stubs the checker with a canned error, so the load-bearing behavior is never exercised end-to-end — add a real clean-create test (it would surface finding 2 on the Windows CI runner) and a modified-tracked (` M`) dirt case, not just untracked (`??`).

**4.** CI has never executed (fork-approval gate, 0s runs — not the 8/5-6 infra outage). I'm approving the workflow runs now; merge also waits on green. No stale-snapshot hazard: nothing since your base touches `cmd/bd/worktree_cmd*.go`.

Happy to re-review promptly once 1-2 are addressed.

steveyegge (approved):

Adversarial pass (bee, beads-lane steward). Fail-fast on a dirty just-created worktree is the right call for the agent-automation context this serves: a worktree that is born dirty means the environment is already lying to whoever is about to work in it, and continuing into the .gitignore/permission follow-ups just buries the evidence. Placement is correct (before any repo mutation — the test pinning .gitignore untouched on refusal is the assertion that matters), the function-var seam matches the file's existing test conventions, and the two tests split real-status detection from call-path wiring cleanly.

Two non-blocking notes, both fine to leave for follow-ups:

1. **Phantom-dirty repos have no escape hatch.** A repo with inconsistent line-ending normalization (CRLF committed under `autocrlf=input`, or aggressive smudge filters) shows modified files in *every* fresh checkout — on such a repo this check refuses every `bd worktree create` with no override. That repo state is itself pathological, and loud-refusal-with-status-output is a defensible answer, but if a user hits it in the wild the fix on their side is nontrivial. If reports come in, an `--allow-dirty` flag is the shape of the answer — not needed to merge.

2. **`CombinedOutput` folds stderr into the cleanliness read.** `git status --porcelain` exiting 0 while printing a warning to stderr (permission-denied on a subdir, etc.) would make a clean tree read as dirty. Since any such warning also means something is genuinely wrong with the worktree, the false refusal is loud rather than silent — acceptable — but `Output()` with stderr captured separately into the error path would make the refusal message say the true thing.

The three failing checks are the same era-ambient trio (doc flags freshness / PR Policy wrapper timing / gate aggregate) failing on every Aug-4/5-vintage branch and passing on all fresh runs; refreshing the branch to re-run CI, merging on green.

PR #5341, opened 2026-08-04
fix(dolt): preserve transaction outcome boundaries

julianknutsen:
## What

Make Dolt-backed write outcomes truthful across server and embedded storage:

- public transaction callbacks run at most once after entry;
- only rollback-safe setup failures can retry;
- ambiguous SQL commit or Dolt publication outcomes expose `ErrCommitIndeterminate`;
- direct and retry-backed writes share exact-once circuit admission, failure, and success accounting.

This also closes the affected claim, dependency, slot, wisp, settlement, recompute, and pull publication paths. It does not change Gas City or add an orchestration-specific API.

## Why

A server-Dolt transaction could commit its SQL mutation, lose or fail during the later publication phase, and then replay the caller callback. Downstream code could therefore receive a false precondition failure after its mutation had already landed. Other direct write families either failed to classify response loss as indeterminate or bypassed circuit admission/accounting.

The safe contract is at-most-once callback execution with an explicit indeterminate result whenever Beads cannot prove whether publication landed. Callers may verify that result, but must not blindly replay it.

## Risk and review

This is a protected storage-path change spanning server and embedded Dolt transaction/publication behavior. The PR is intentionally explicit about the risk and requires substantive human storage review before merge. It must not be auto-merged.

## Verification

- Focused deterministic RED/GREEN tests for post-commit response loss, callback replay, claim verification, direct-write circuit admission/reset, and every repaired server/embedded publication boundary.
- Uncached affected-package suites for `internal/storage/dolt` and `internal/storage/embeddeddolt`.
- Race suites for both storage packages.
- Exact final-SHA `make test`: pass, 39.4% total coverage.
- `go vet ./...`: pass.
- Go 1.26.5 `golangci-lint run ./...`: 0 issues.
- `git diff --check origin/main...HEAD`: pass.
- Two independent final-SHA Sol reviews: 0 Blockers / 0 Majors each.
- PR preflight and broader open-PR search found no competing contributor PR.

Tracking context: `ga-f7v2ft.79.1`.

### DIFF
--- a/backend/conformance/conformance.go
+++ b/backend/conformance/conformance.go
@@ -26,6 +26,7 @@ package conformance
 
 import (
 	"context"
+	"database/sql/driver"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -208,6 +209,7 @@ func RunAll(t *testing.T, factory Factory) {
 
 	// Transaction
 	t.Run("Transaction", func(t *testing.T) { testTransaction(t, factory) })
+	t.Run("TransactionCallbackAtMostOnce", func(t *testing.T) { testTransactionCallbackAtMostOnce(t, factory) })
 	t.Run("TransactionSnapshotReads", func(t *testing.T) { testTransactionSnapshotReads(t, factory) })
 	t.Run("TransactionReadYourWrites", func(t *testing.T) { testTransactionReadYourWrites(t, factory) })
 }
@@ -945,6 +947,22 @@ func testTransaction(t *testing.T, f Factory) {
 	}
 	if !found {
 		t.Errorf("label 'from-tx' not found, got %v", labels)
+	}
+}
+
+func testTransactionCallbackAtMostOnce(t *testing.T, f Factory) {
+	s := f(t)
+	calls := 0
+
+	err := s.RunInTransaction(ctx(), "test: transient callback", func(storage.Transaction) error {
+		calls++
+		return driver.ErrBadConn
+	})
+	if !errors.Is(err, driver.ErrBadConn) {
+		t.Errorf("RunInTransaction error = %v, want original callback error %v", err, driver.ErrBadConn)
+	}
+	if calls != 1 {
+		t.Errorf("transaction callback calls = %d, want 1", calls)
 	}
 }
 
--- a/backend/errors.go
+++ b/backend/errors.go
@@ -18,19 +18,20 @@ type ErrUnsupported = storage.ErrUnsupported
 // identities: a backend must return (or wrap) these values, not
 // equivalently-worded errors of its own.
 var (
-	ErrAlreadyClaimed    = storage.ErrAlreadyClaimed
-	ErrAlreadyExists     = storage.ErrAlreadyExists
-	ErrAssigneeMismatch  = storage.ErrAssigneeMismatch
-	ErrCloseBlocked      = storage.ErrCloseBlocked
-	ErrCloseOpenChildren = storage.ErrCloseOpenChildren
-	ErrNotClaimable      = storage.ErrNotClaimable
-	ErrNotFound          = storage.ErrNotFound
-	ErrNotInitialized    = storage.ErrNotInitialized
-	ErrNotOwner          = storage.ErrNotOwner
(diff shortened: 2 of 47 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98638767

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR preserves transaction outcome boundaries in the Dolt storage backend — hardening the commit/verify path so a lost or indeterminate commit-call response, a degraded-server rollback, or an ignored-table "dirty wisp" can no longer be laundered into a false success or replayed as a non-idempotent write. That matters because it protects the issue database's coordination and claim paths from silent corruption or double-applied writes when the Dolt server hiccups mid-commit.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Phantom claim after a degraded-server rollback: claim-family writes now re-read coordination state after a reported-success write and fail loudly on mismatch, instead of trusting a rolled-back "success".
- Premature circuit-breaker success: `withCircuitWrite` now records success only after the full write-plus-verification boundary returns cleanly, so a write that fails verification no longer counts as a healthy write.
- Indeterminate commit replay: SQL commit response-loss now routes through a shared `ErrCommitIndeterminate` sentinel and stops retrying rather than replaying a possibly-committed, non-idempotent write.
- Ignored-table "dirty wisp" commit failures: `commitWorkingSet` now applies the same `dolt_ignore` anti-join as `HasCommittablePending`, so ignored wisp tables no longer force-fail commit staging.
- Verify-by-reread laundering on mixed updates: coordination verification is now scoped to claim-family / coordination-only updates, so a non-coordination patch keeps its raw write outcome instead of being "verified" by an unrelated reread.
- Direct-write and embedded-backend parity: direct Dolt write entrypoints now route through the same circuit-managed write boundary, and the embedded backend mirrors the indeterminate-commit boundary, so both paths behave consistently under partial failure.

Review findings addressed:
- Correctness / transaction-state safety (validated, Claude + Codex): the review independently re-verified each of the six boundary fixes above against the final patch and confirmed the claim, circuit-accounting, indeterminate-commit, ignored-table, mixed-update, and direct/embedded-parity behaviors are all correct; no new blocker, major, minor, or nit code findings were raised on the final patch.

Remaining non-gating notes:
- The circuit breaker intentionally fast-trips within a retry sequence under a sustained server outage; this is expected behavior and only worth revisiting if field evidence shows false-positive trips.
- Re-test the Dolt commit-in-transaction and ignored-table staging boundaries during future Dolt upgrades, since those semantics are version-sensitive.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch:
- `fix(dolt): record circuit success only after claim verify; cut retry/patch complexity`
- `fix(dolt): record circuit success after facade update verify; skip ignored tables in commit`
- `fix(dolt/test): match commitWorkingSet dolt_ignore anti-join in sqlmock expectations`

The first two refine circuit-success accounting and ignored-table commit staging and add focused regression tests (`claim_commit_boundary_test.go`, `commit_dirty_wisp_test.go`, `issue_operations_test.go`) — 10 files changed, +655 / -141. During finalization the branch was also rebased onto the latest `main`, and one modify/modify conflict in `internal/storage/embeddeddolt/version_control.go` (`RecomputeAllBlocked`) was resolved so it keeps both `main`'s shared-constant/guard refactor and this PR's indeterminate-aware commit wrapper; the net change to that file is identical to before the rebase (+27 / -11). The third commit is a test-only follow-up: once rebased onto the current base, `commitWorkingSet`'s `dolt_status` lookup uses the `dolt_ignore` anti-join, so the `draincall` and `settlement` sqlmock regression expectations were realigned to match it (+6 / -6, test files only, no production behavior change).

### Final Review Status
Ready for the merge queue: final head `27685cc5` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/31096336211

### Review Iterations
3 review passes performed; the final pass approved with all prior transaction-boundary fixes validated and no new code findings.

---

PR #5343, opened 2026-08-04
Fix: wisp delete never cascades to auxiliary tables (orphan row leak)

quad341:
## What this changes

Deleting a wisp (`bd` issue) previously removed only the `wisps` row and its
dependency edges — it never removed the wisp's rows in `wisp_labels`,
`wisp_events`, `wisp_comments`, or `wisp_child_counters`. On the live `hq`
city store this leaked ~3.03M orphan rows (92% of `wisp_labels`, 89% of
`wisp_events`), contributing materially to store size and to the row counts
scanned by unfiltered reads over those tables.

Both delete paths (`deleteWisp` and the batch path `deleteWispBatchTx`) now
delete a wisp's rows from all four auxiliary tables inside the same
transaction as the wisp delete itself, via one shared helper so the table
set can't drift between the two paths again.

## Design note

Some deployed stores enforce this cleanup via FK `ON DELETE CASCADE` and
some do not (the migration adding those FKs was never promoted out of
`migrations/ignored/`), so the delete paths can't rely on the database to
do it for them — the fix has to be explicit at the call site, which is why
it's a small shared helper rather than a schema change.

## Review notes

- Table/column identifiers in the new helper come from a fixed literal
  slice, not user input; row-scoping ids use the existing parameterized
  `?` placeholder helper — same `//nolint:gosec // G201` pattern already
  used elsewhere in this file.
- Backfilling the ~3.03M existing orphan rows on the live `hq` store is
  intentionally out of scope here — that's a live-prod data operation for
  operator authority, handled separately. This PR only stops the leak from
  growing further.

## Test plan

- [x] `go test -tags=integration,gms_pure_go -race -timeout=15m ./internal/storage/dolt/...` — 575 PASS / 0 FAIL / 951 SKIP (skips are a pre-existing container-runtime gap in the test sandbox, unrelated to this diff)
- [x] New coverage: a test creates a wisp with labels/events/comments, deletes it via both the single and batch paths, and asserts zero residual rows in every auxiliary table
- [x] New coverage: 5 repeated create/delete cycles assert orphan counts stay at 0 store-wide after every cycle
- [x] Release gate: [`release-gates/be-1gd16-wisp-delete-cascade-gate.md`](release-gates/be-1gd16-wisp-delete-cascade-gate.md)

## Bead

be-1gd16 — deploy tracking

🤖 Deployed by actual-factory

### DIFF
--- a/internal/storage/dolt/wisp_delete_cascade_integration_test.go
+++ b/internal/storage/dolt/wisp_delete_cascade_integration_test.go
@@ -1,0 +1,209 @@
+//go:build integration && !windows
+
+package dolt
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/steveyegge/beads/internal/doltserver"
+	"github.com/steveyegge/beads/internal/testutil/integration"
+)
+
+// setupWispCascadeStore starts a real, standalone dolt sql-server (no Docker)
+// via doltserver.Start and returns a store on a fresh, fully
+// schema-initialized database. This mirrors TestFreshBootstrapHealIncarnation's
+// setup rather than the package's default setupTestStore (dolt_test.go),
+// which requires a Docker-testcontainer Dolt server that this sandbox does
+// not have available.
+func setupWispCascadeStore(t *testing.T) *DoltStore {
+	t.Helper()
+	integration.RequireDolt(t)
+	t.Setenv("BEADS_DOLT_SHARED_SERVER", "0")
+	t.Setenv("BEADS_DOLT_SERVER_PORT", "")
+	t.Setenv("BEADS_DOLT_PORT", "")
+
+	beadsDir := filepath.Join(t.TempDir(), ".beads")
+	if err := os.MkdirAll(beadsDir, 0o700); err != nil {
+		t.Fatalf("mkdir beads dir: %v", err)
+	}
+	state, err := doltserver.Start(beadsDir)
+	if err != nil {
+		t.Fatalf("start local dolt server: %v", err)
+	}
+	t.Cleanup(func() {
+		current, stateErr := doltserver.IsRunning(beadsDir)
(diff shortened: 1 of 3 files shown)


steveyegge:
**Fable adversarial review (bee, beads-lane steward) — VERDICT: MERGE** (follow-ups filed as beads, none blocking).

Verified against main: the four-table set is exactly the FK-less set (their CASCADE FKs live only in never-applied `ignored/0004`, while `wisp_dependencies` got real FKs in 0021), transaction boundaries are sound (helper runs in-tx before `RecomputeIsBlockedInTx`, errors roll back the whole batch), performance is fine (all four tables indexed on the scoping column), promoted wisps are safe, and the test's FK-drop setup is honest and load-bearing — without it fresh-store FKs would mask the bug. CI fully green. This covers the two paths carrying essentially all the leak volume.

For the record — the 'so the table set can't drift' claim is weaker than stated: the table set now exists in three hand-kept copies, and three OTHER wisp-removal paths on main still leak the same class. Filing follow-up beads rather than bouncing this fix:
1. `issueops.deleteIssueRowInTx(…, isWisp=true)` (reachable via `bd delete --cascade` with wisp dependents, the transactional `DeleteIssue`, and the force-delete sweep) deletes only the wisps row + sync-plane edges — the natural home for this helper was issueops beside it, derived from the authoritative `DeleteCascadeTables(true)` list.
2. `PromoteFromEphemeralInTx` never touches `wisp_child_counters` (stranded on hq-shaped stores, and the counter is never migrated to `child_counters`).
3. The domain/proxied-server delete path (`internal/storage/domain/issue_delete.go`) deletes wisp deps/labels/events but never `wisp_comments`/`wisp_child_counters` — relevant with the proxied cutover in flight.

Nits, no action needed: leftover `t.Logf DEBUG` scaffolding in the new integration test; `deleteWispBatchTx`'s '6 DELETE statements' doc comment now undercounts.

Merging on posted verdict.

PR #4418, opened 2026-06-16
Triage and clean up an unreviewed automated commit stream on main

julianknutsen:
# Triage and clean up an unreviewed automated commit stream on main

## Summary

Between `9a1c88b63` (the reviewed dolt-2.1.4 bump, PR #4326) and the current `origin/main`
tip `8e18581dc`, an automated agent committed a large stream of changes directly onto the
shared branch without going through PR review. A commit-by-commit triage found that the
cited issue references could not be located in the tracker, so each commit's stated
justification could not be independently verified.

This PR does **not** discard the stream. The triage found that roughly half of it is
**genuine, diff-provable bug-fixing**, and that several commits are *load-bearing on an
already-published Dolt schema* — reverting them would brick deployed clones. So this PR:

1. **Reverts the no-op / characterization commits** that don't fix a demonstrable defect (4 commits).
2. **Reverts the unreviewed changes in the gated, not-running serverV2 subsystem** (the
   proxied / external-dolt-server layer) back to BASE — itemized in the **serverV2 audit**
   comment for the subsystem owner to review.
3. **Keeps the real fixes**, and — since the cited references can't serve as the source of
   truth — **backs every kept real defect with a proven born-failing regression test.**
4. **Flags the remaining judgement calls** for maintainer decision rather than silently
   keeping or reverting them.

**Commits (6):** 4 reverts of no-op commits + 1 commit adding proven regression tests + 1
serverV2 revert. Per-commit justification for the reverts and the flagged decisions is in the
**Removal Rationale** comment; the serverV2 fixes being removed are in the **serverV2 audit**
comment.

### serverV2 revert (`8515879dc`)
The proxied / external-dolt-server subsystem is **gated off and not running yet**. This reverts
the unreviewed changes to its schema-independent layers — `internal/storage/dbproxy/*`,
`internal/storage/uow/*`, and the `cmd/bd` proxied glue — back to BASE (26 files, +107/-870).
A *strict* serverV2→BASE revert is impossible without breaking main: `internal/storage/domain/db`
is **welded to published migration 0051** (BASE inserts without an `id` and relies on the
`DEFAULT(UUID())` that 0051 removed), so the data layer + the issueops widening it needs are
**kept at HEAD**. The removed serverV2 fixes — most of them *real, proven* bugs in the gated
subsystem — are itemized for the owner to audit / re-apply.

## The source-of-truth finding

[...]

PR #5345, opened 2026-08-05
fix(proxied): corrupt sidecar/client-info aborts instead of silent fresh-DB fallback (bd-aj3g5)

steveyegge:
## Summary

Restores the corrupt-sidecar hard-error guard in proxied-server mode (bd-aj3g5, program bd-vxavz; regression of bd-6dnrw.44 item 6).

`resolveProxiedServerUOWTopology` (cmd/bd/uow_factory.go) discarded the load errors for both `metadata.json` (`configfile.Load`) and `proxied_server_client_info.json` (`configfile.LoadProxiedServerClientInfo`). Both loaders return `(nil, nil)` when the file is **absent**, so a non-nil error means the file **exists but cannot be read or parsed** — and swallowing it made bd silently fall back to a FRESH managed local database.

### The split-brain failure mode

For a multi-clone constellation pointed at an external/shared Dolt server, a corrupted or permission-broken sidecar does not make the clone fail — it makes it *fork*: reads return zero issues, writes land in a brand-new empty local database, and the clone looks alive while diverging from the shared truth. A corrupted `metadata.json` additionally silently defaults the database name and degrades the team-server identity assertion to no assertion at all.

### History: restore of a reverted fix

- `f880a985b` (2026-06-12) fixed exactly this: "corrupt sidecar/config aborts instead of silent fresh-DB fallback (bd-6dnrw.44 item 6)", with regression tests.
- `a59e75325` (#4418, 2026-06-15) reverted it wholesale as part of "Triage and clean up an unreviewed automated commit stream".
- The code has since been restructured (loads moved into `resolveProxiedServerUOWTopology`), and the comment at the call site described the swallow as "pre-existing behavior... tracked separately", unaware a fix had existed. This PR re-applies the fix adapted to the current structure and replaces that comment with the actual contract.

### Absent-vs-corrupt matrix

| File | Absent | Present but corrupt/unreadable |
|---|---|---|
| `metadata.json` | defaults (fresh init / managed fallback legal) — unchanged | hard error naming the path, refusing the fallback — **restored** |
| `proxied_server_client_info.json` | defaults — unchanged | hard error naming the path, refusing the fallback — **restored** |

Error shape: `corrupt proxied-server sidecar <path>: <underlying err> — refusing to fall back to a fresh database; repair or remove the file to proceed` (and the `corrupt workspace config <path>: ...` sibling).

### Audit of other swallowed loads on the proxied path

- `cmd/bd/uow_factory.go` — the two call sites above were the only `_ :=` loads; `resolveServerModeUOWTopology` already propagates its `configfile.Load` error.
- `cmd/bd/proxied_server.go` — `resolveProxiedServerConfigPath` / `resolveProxiedServerLogPath` already propagate; the only remaining `persisted, _ :=` is inside a commented-out legacy function (line ~613).
- `internal/doltserver/physical_root.go` `ResolveProxiedServerRootPath` — already propagates.
- `internal/configfile` loaders already distinguish `os.IsNotExist` → `(nil, nil)` from corrupt → error, so no loader-layer change was needed; the previously-unpinned corrupt→error contract is now pinned by tests.

### Tests

New (all in this PR):
- `TestNewProxiedServerUOWProvider_CorruptSidecarErrorsInsteadOfManagedFallback` — garbage JSON sidecar → hard error naming the path, "refusing to fall back" (resurrected from f880a985b, adapted).
- `TestNewProxiedServerUOWProvider_CorruptWorkspaceConfigErrors` — garbage JSON `metadata.json` → hard error naming the path (resurrected, adapted).
- `TestNewProxiedServerUOWProvider_UnreadableSidecarErrors` — chmod 000 sidecar → hard error (skips as root).
- `TestResolveProxiedServerUOWTopology_AbsentFilesResolveDefaults` — absent files → defaults, no error (pinned at the resolver level; the full provider would go on to start a managed dolt server).
- `internal/configfile`: `TestLoadCorruptIsError` + "corrupt sidecar is an error, not nil" subtest — pin the loader-layer absent-vs-corrupt contract the call-site fix relies on.

Evidence (local, verified exit codes):
- `go build ./... && go vet ./...` — clean (exit 0).
- `go test ./cmd/bd -run 'Sidecar|UOWFactory|Factory|ConfigFile|ProxiedServerUOW|LoadCorrupt' -count=1` — ok, exit 0 (~25s; all five new cmd/bd tests plus neighboring factory/sidecar tests, each individually PASS in a `-v` run).
- `go test ./internal/configfile/... -count=1` — ok, exit 0.
- The full `go test ./cmd/bd -count=1 -short` was attempted three times locally and the execution environment was recycled out from under it each time (~10 min runtime); no complete local run was observed, so no claim is made about it here — CI's required lanes are the authority for the full suite.

Refs: bd-aj3g5, bd-vxavz, bd-6dnrw.44 item 6. Restores f880a985b, reverted by a59e75325 (#4418).

### DIFF
--- a/cmd/bd/uow_factory.go
+++ b/cmd/bd/uow_factory.go
@@ -103,18 +103,21 @@ func newSQLServerUOWProvider(ctx context.Context, beadsDir string, topology sqlS
 }
 
 // resolveProxiedServerUOWTopology reads a proxied-server workspace's topology
-// out of metadata.json and the proxied-server sidecar. Neither read is fatal:
-// an absent or unreadable one leaves the defaults, which is the behavior every
-// proxied command has had, and the provider construction below is where a
-// workspace that cannot be reached actually fails. The one refusal it can
-// return is the team-server workspace with no identity to assert.
+// out of metadata.json and the proxied-server sidecar. An ABSENT file is not
+// fatal — both loads return (nil, nil) then, the defaults apply, and the
+// provider construction below is where a workspace that cannot be reached
+// actually fails. A file that EXISTS but cannot be read or parsed IS fatal
+// (restores f880a985b, reverted in #4418; bd-aj3g5): swallowing it silently
+// falls back to a fresh managed local database — reads return zero issues and
+// writes land in the wrong database (split-brain) — and the identity assertion
+// below silently degrades to no assertion at all.
 func resolveProxiedServerUOWTopology(beadsDir, databaseOverride string, posture identityPosture) (sqlServerUOWTopology, error) {
-	// NOTE: a load error is swallowed here (pre-existing behavior), which
-	// leaves persisted == nil and therefore silently falls back to the default
-	// database with teamServer=false. The identity assertion below inherits
-	// that: an unreadable metadata.json degrades to no assertion rather than a
-	// refusal. Tracked separately with the wider silent-fallback smell.
-	persisted, _ := configfile.Load(beadsDir)
+	persisted, err := configfile.Load(beadsDir)
+	if err != nil {
+		return sqlServerUOWTopology{}, fmt.Errorf(
+			"corrupt workspace config %s: %w — refusing to fall back to a fresh database; repair or remove the file to proceed",
+			configfile.ConfigPath(beadsDir), err)
+	}
 	topology := sqlServerUOWTopology{database: configfile.DefaultDoltDatabase}
 	if persisted != nil {
 		topology.database = persisted.GetDoltDatabase()
@@ -137,7 +140,12 @@ func resolveProxiedServerUOWTopology(beadsDir, databaseOverride string, posture 
 		topology.database = databaseOverride
 	}
 
-	info, _ := configfile.LoadProxiedServerClientInfo(beadsDir)
+	info, err := configfile.LoadProxiedServerClientInfo(beadsDir)
+	if err != nil {
+		return sqlServerUOWTopology{}, fmt.Errorf(
+			"corrupt proxied-server sidecar %s: %w — refusing to fall back to a fresh database; repair or remove the file to proceed",
+			configfile.ProxiedServerClientInfoPath(beadsDir), err)
+	}
 	if info != nil {
 		topology.proxyPort = info.Port
 		topology.proxyIdle = info.IdleTimeout
--- a/cmd/bd/uow_factory_test.go
+++ b/cmd/bd/uow_factory_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"os"
 	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


steveyegge (commented):
**GATE REVIEW (bd-aj3g5, program bd-vxavz) — VERDICT: APPROVE** (two wording nits, non-blocking). Reviewed by lion (wyvern wheelhouse lane).

Verified: built + vetted clean; all four new cmd/bd tests and the configfile suite pass locally (exit 0). Mutation-checked — re-swallowing both load errors sends all three corruption tests RED while the absent-defaults test stays green, so the tests genuinely pin the guard. CI is fully green across every cmd/bd lane; no bd-vb2u0 ambient-failure attribution needed (CI substitutes for the never-completed local full run).

The absent-vs-corrupt matrix holds at the loader layer for every corruption flavor probed (garbage, empty, truncated, wrong-typed known fields, perm-denied, corrupt legacy config.json) — all error; absent resolves defaults. Untested-in-PR cells (empty/truncated/wrong-types) share the exact `err != nil` path with the tested ones, so no additional tests requested.

**On the deliberate `bd init --team-server` compat break: correct, keep it.** Corrupt metadata.json cannot brick init — `checkExistingBeadsDataAt` and `ResolveProxiedInit` already fail closed before this code (with `--reinit-local` as the documented escape), and init rewrites metadata.json before the adopting provider opens. The one newly-reachable stop is a corrupt *sidecar* on re-init, and the pre-PR behavior there was worse: the swallow spun up a fresh managed DB and surfaced a misleading "database has no project identity — provision it with 'bts init'". The new error names the actual broken file and a recovery; the sidecar is machine-local and regenerable.

Nits:
1. `cmd/bd/uow_factory.go:117` — when the corrupt file is the legacy `config.json` (migration path inside `configfile.Load`), the error names the nonexistent `metadata.json` path; only the wrapped "parsing legacy config" hints at the real file.
2. `cmd/bd/uow_factory.go:146` — "repair or **remove** the file to proceed": on an external-server workspace, removing the sidecar converts it to managed-local fresh-DB mode — the very fork the guard exists to prevent. Suggest: "restore the file (re-run `bd init` with your server settings), or remove it only if this workspace should use a managed local server."

For the record: the metadata.json branch is defense-in-depth (`main.go:1299` already refuses corrupt metadata on the CLI path); the sidecar branch is the load-bearing new behavior — pre-PR it either errored later non-actionably or, with `BEADS_PROXIED_SERVER_ROOT_PATH` set, forked silently. The f880a985b restore is faithful and extended (perm-denied case, absent-defaults pin, loader-layer contract tests are new); #4418's own proof ledger had classified that commit as a proven real fix — it was reverted only in the blanket serverV2 rollback, and this is the reviewed re-application it invited.

PR #5346, opened 2026-08-05
feat(gate): port create/add-waiter/resolve/show to proxied-server mode (bd-v50ru)

steveyegge:
## What

Ports the four remaining routable `bd gate` subcommands — **create, add-waiter, resolve, show** — to proxied-server mode, extending the conventions `gate check`/`gate list` already established in `cmd/bd/gate_proxied_server.go`. Refs **bd-v50ru**, program **bd-vxavz**.

`gate discover` deliberately keeps its proxied-mode refusal: it interleaves GitHub CLI queries with per-gate writes and incremental per-gate output, so it is not the pure store-routing this port covers. It can be ported later with the discovered-map pattern `gate check` uses if the need arises.

## How

- **gate create** — ONE `uow.RunTxResult` doing target lookup + `IssueUseCase().CreateIssue` + `DependencyUseCase().AddDependency`, under the same commit message the direct route writes (`bd: create gate <id> blocking <target>`). The direct route's create + add-dependency + explicit `store.Commit` collapse into **one Dolt commit per invocation** — semantically equivalent, minus the window where the gate exists without its blocking edge.
- **gate add-waiter** — one transaction; re-reads the gate in-tx, validates type, appends the waiter. An already-registered waiter returns an **empty commit message** (deliberate no-commit) and prints the classic "Waiter already registered" line. Fires the `on_update` hook post-commit.
- **gate resolve** — one transaction (existence + type check + `CloseIssue`), then `audit.LogFieldChange` + close hooks post-commit — exactly the shape `gate check`'s in-file closes already use.
- **gate show** — read-only UOW. Fail-closed: not-found exits nonzero with `gate not found: <id>`, while a failed **read** exits nonzero with a distinct `reading gate <id>: …` message, so an unreachable server can never be mistaken for "no gate".

**Output parity is structural, not copied:** the direct route's renderers and gate-create construction were extracted into shared helpers (`renderGateShow`, `renderGateCreated`, `renderGateResolved`, `renderGateWaiterAdded`/`Already`, `gatherGateCreateInput`, `buildGateIssue`) and both routes call the same code. The `Created gate <id>` line, the `gate show` markers, and `gate list --all --json`'s `.id/.status/.description/.waiters` are what the wyvern wheelhouse scripts parse; the new journey test pins all of them.

## The domain/db change (please scrutinize)

`internal/storage/domain/db/issue.go` — the new journey test caught a latent bug beyond cmd/bd routing: the proxied repository's `normalizeUpdateValue` never serialized the `waiters` field, so **any** waiters update through the proxied backend failed at the SQL driver:

```
Error: updating gate: db: Update ghj-v5l: sql: converting argument $1 type: unsupported type []string, a slice of string
```

`waiters` was already in `allowedUpdateFields`, and the embedded path (`issueops.updateIssueInTx`, update.go:429) has always JSON-marshaled it — the proxied repo just never grew the same arm. The fix mirrors the embedded path exactly (unconditional `json.Marshal` for the `waiters` key), placed after `DiscardNoopIssueUpdates` so the no-op comparison still sees the raw `[]string`. `normalize_update_value_test.go` pins the serialization.

## Tests

- `go build ./...`, `go vet ./cmd/bd/... ./internal/storage/domain/db/...`, `gofmt -l` — all clean, and the pre-commit `golangci-lint run --new-from-rev` reported **0 issues**.
- `go test ./cmd/bd -run 'Gate' -count=1` (embedded/unit) — **ok, 148.8s** (re-run after the domain/db fix).
- `BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd -run 'TestProxiedServerGate' -count=1` (docker testcontainers) — **ok, 283.1s**: `TestProxiedServerGateCheck` (pre-existing, no regression) + new `TestProxiedServerGateLifecycle` (human-gate journey incl. ready-front block/unblock, wh-park bead-gate shape resolved by `gate check`, `--json` create, fail-closed error table). Its first run failed on the waiters serialization bug above — the fix is what made it green, i.e. the journey test earned its keep.
- `go test ./internal/storage/domain/... -count=1` — **all ok** (incl. the full dolt-backed `domain/db` suite, 366.9s).
- `TestProxiedServerGateLifecycle` added to `.github/scripts/proxied-cmd-test-shards.txt` shard 1 (the lightest shard by the generator's own cost metric; a full regeneration reshuffles all ~108 entries and drops the file's hand-written placement notes, so this follows the file's "newly-added tests" append path instead).
- A whole-tree `go test ./internal/storage/...` all-at-once run showed `internal/storage/dolt` and `internal/storage/uow` failing **only** with `failed to create Dolt store: failed to initialize schema: context deadline exceeded` / 10m suite timeouts — container-setup starvation from every dolt-backed package spinning containers concurrently on one laptop, in packages this diff does not touch. A follow-up isolated re-run is noted below in comments if anything differs.

## Deviations / notes for review

- No existing tests asserted the old "not supported in proxied-server mode" strings for these four subcommands — nothing to invert.
- `gate create --json` on the proxied route returns the server-hydrated created issue (timestamps etc. populated by the create result) rather than the client-built struct the direct route echoes; `.id`/`.issue_type` and friends are identical.
- Error-order nit: `--timeout` is now parsed before the `--blocks` target lookup on both routes (shared `gatherGateCreateInput`), so an invocation with **both** a bogus timeout and a missing target reports `invalid timeout` where the direct route used to report `issue not found` first.
- `gate show`'s read-failure message (`reading gate <id>: …`) is new on the proxied route only; the direct route folded every read error into `gate not found`. Fail-closed law says those must be distinguishable.

https://claude.ai/code/session_018YmpYVCa5eN6JtG26dJLMs

### DIFF
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -17,7 +17,9 @@
 
 15 1 TestProxiedServerCount
 15 1 TestProxiedServerDelete
+15 1 TestProxiedServerDelete
 15 1 TestProxiedServerDoltRemoteRemove
+15 1 TestProxiedServerGateLifecycle
 15 1 TestProxiedServerMolCurrent
 15 1 TestProxiedServerPathHelpers
 15 1 TestProxiedServerTodo
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -17,7 +17,6 @@
 
 15 1 TestProxiedServerCount
 15 1 TestProxiedServerDelete
-15 1 TestProxiedServerDelete
 15 1 TestProxiedServerDoltRemoteRemove
 15 1 TestProxiedServerGateLifecycle
 15 1 TestProxiedServerMolCurrent
--- a/cmd/bd/gate.go
+++ b/cmd/bd/gate.go
@@ -233,7 +233,7 @@ This is used by 'bd done --phase-complete' to register for gate wake notificatio
 	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if usesProxiedServer() {
-			return HandleErrorRespectJSON("gate add-waiter is not supported in proxied-server mode")
+			return runGateAddWaiterProxiedServer(cmd, rootCtx, args)
 		}
 		CheckReadonly("gate add-waiter")
 
@@ -262,7 +262,7 @@ This is used by 'bd done --phase-complete' to register for gate wake notificatio
 
 		for _, w := range issue.Waiters {
 			if w == waiter {
-				fmt.Printf("Waiter already registered on gate %s\n", gateID)
+				renderGateWaiterAlready(gateID)
 				return nil
 			}
 		}
@@ -278,9 +278,19 @@ This is used by 'bd done --phase-complete' to register for gate wake notificatio
 
 		commandDidWrite.Store(true)
 
-		fmt.Printf("%s Added waiter to gate %s: %s\n", ui.RenderPass("✓"), gateID, waiter)
+		renderGateWaiterAdded(gateID, waiter)
 		return nil
 	},
+}
+
+// renderGateWaiterAlready and renderGateWaiterAdded are shared by the direct
+// and proxied-server routes so `bd gate add-waiter` prints identically on both.
+func renderGateWaiterAlready(gateID string) {
+	fmt.Printf("Waiter already registered on gate %s\n", gateID)
+}
+
+func renderGateWaiterAdded(gateID, waiter string) {
+	fmt.Printf("%s Added waiter to gate %s: %s\n", ui.RenderPass("✓"), gateID, waiter)
 }
 
 // gateCreateCmd creates an ad-hoc gate issue that blocks another issue
@@ -308,7 +318,7 @@ Examples:
 	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if usesProxiedServer() {
-			return HandleErrorRespectJSON("gate create is not supported in proxied-server mode")
+			return runGateCreateProxiedServer(cmd, rootCtx)
 		}
 		CheckReadonly("gate create")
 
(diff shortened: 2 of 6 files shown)


steveyegge:
Follow-up verification on the whole-tree `go test ./internal/storage/...` reds mentioned in the PR body (`internal/storage/dolt`, `internal/storage/uow`):

- The six failing `internal/storage/dolt` tests (`TestConcurrentIssueCreation`, `TestSameIssueUpdateRace`, `TestReadWriteMix`, `TestHighContentionStress`, `TestConcurrentWorkQueueDrain`, `TestBranchPerAgentMergeRace`) all failed with `failed to create Dolt store: failed to initialize schema: context deadline exceeded` at a uniform ~45s setup deadline. Re-run in isolation on this branch: **ok, 124.0s** — all six pass. This package is not touched by the diff at all (it is the embedded store; the fix is in the proxied `domain/db` repository).
- `internal/storage/uow` scoped to the diff's blast radius (`-run 'Update'`, which covers the issue-update paths the `waiters` serialization fix sits under): **ok, 528.6s**. The full package run exceeds even a 20m binary budget on this host while it is shared with other test jobs, with the same `init schema: context deadline exceeded` signature (`TestNewExternalDoltServerUOWProvider_EndToEnd`) — container-setup starvation, not assertion failures. No test in `internal/storage/uow` or `internal/storage/dolt` references `waiters`.
- Also re-verified green after the domain/db fix: `go test ./internal/storage/domain/...` (full dolt-backed `domain/db` suite, 366.9s), `go test ./cmd/bd -run 'Gate'` (148.8s), and `BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd -run 'TestProxiedServerGate'` (283.1s).

Conclusion: the whole-tree reds are host-saturation artifacts; everything in or near this diff's blast radius is green. CI's sharded proxied lane is the authoritative check.

steveyegge (commented):
**GATE REVIEW (bd-v50ru, program bd-vxavz) — VERDICT: REQUEST_CHANGES** — one-line CI blocker; the port itself looks correct and was verified locally. Reviewed by lion (wyvern wheelhouse lane); gates are our wait/wake primitive, so resolve/add-waiter semantics got the hardest look.

**Blocker:** the diff to `.github/scripts/proxied-cmd-test-shards.txt` adds a stray duplicate `15 1 TestProxiedServerDelete` line next to the intended `TestProxiedServerGateLifecycle` entry (lines 19-20 on the branch). The shard script's duplicate check fails all 15 Proxied Dolt Cmd jobs in seconds, so `TestProxiedServerGateLifecycle` has never actually run in CI. FAIL-set-diff done: the merge-base main run's only red is the known-ambient `PR Core (wrapper timing)` flake (bd-vb2u0 shape) — the 15 shard reds are this PR's own. Drop the duplicate line and let CI run the new suite; on green this is an approve.

**Waiters serialization (`normalizeUpdateValue`, the designated hardest-scrutiny hunk): verified byte-identical to the embedded path.** Both are an unconditional `json.Marshal` (`issueops/update.go:429-431` classic vs `domain/db/issue.go:836-841` proxied): nil→`null`, `[]`→`[]`, insertion order, no sorting, same HTML/quote escaping — checked nil/empty/single/duplicates/comma/quote/unicode. Read path is shared (`scanIssue` delegates to `issueops.ScanIssueFrom` → `ParseJSONStringArray`), which handles `""`, `null`, and old embedded-written rows identically, so no migration concern. Blast radius is nil: `normalizeUpdateValue` has exactly one caller and the branch keys on `waiters` literally; the only other writer passing that key is `merge_slot.go:187`, which this PR incidentally *fixes* in proxied mode (raw `[]string` was previously refused by the driver). The proxied no-op filter compares `[]string` before normalization, so ordering is correct.

**Minor (fix while you're in there):** `runGateResolveProxiedServer` (gate_proxied_server.go:443-451) fires `audit.LogFieldChange` + close hooks without checking `CloseIssueResult.Closed`, so a double-resolve writes a bogus closed→closed audit record. Hook firing on the no-op matches the embedded decorator's documented behavior, but sibling `close_proxied_server.go:137` guards on `o.closed` — same guard for the audit line, please.

**Noted, non-blocking:** proxied `gate create` fires no create/dependency hooks — consistent with the existing proxied `create`/`dep add` gap; worth a tracking issue, not a change request here.

Nice properties: single-tx `gate create` (atomically better than classic's create-then-dep — no orphan gate on dep failure), not-found vs read-failure distinguished fail-closed with tests, and shared renderers/`buildGateIssue`/`gatherGateCreateInput` make cross-mode drift structural rather than disciplinary. Local verification: `TestProxiedServerGateLifecycle` PASS against a real proxied dolt container (16.6s, all subtests); classic gate tests PASS.

steveyegge:
Duplicate `15 1 TestProxiedServerDelete` line dropped (ffafef100) — review blocker addressed; shard manifest now lists Delete once and GateLifecycle should actually run across the 15 shards this time. Per your verdict this is approve-on-green; I'll merge when the Proxied Dolt Cmd lanes come back green.

PR #5348, opened 2026-08-05
fix: bd show no longer deletes HTML-tag-shaped spans from description/notes

jacobhausler:
### What breaks

Goldmark (glamour's markdown parser, used by `bd show` to render
description/notes bodies) parses any `<...>`-shaped text as raw HTML,
and glamour's ANSI renderer sanitizes those nodes down to nothing —
silently dropping the span. For an unclosed tag-shaped token, this can
eat everything up to the next `>` in the document.

This isn't a cosmetic issue: any bead body containing something that
merely *looks* like an HTML tag — a shell redirect, a generic type
parameter, a comparison chain, a `<name>` placeholder — silently loses
that text (or more) every time `bd show` renders it. There is no
warning; the field is just shorter than what was written.

### Reproduction

```
$ bd create "tag-span repro" -d 'before <script> after, and unclosed <foo bar'
$ bd show <id>
DESCRIPTION

  before  after, and unclosed <foo bar
```

The `<script>` span is gone entirely — compare the stored value (visible
via `bd show <id> --json`, which is untouched) against the rendered
output.

### The fix

Escape angle brackets before handing body text to glamour, and unescape
them in the rendered output, so `bd show` never eats plain text that
merely looks like a tag. Scoped to `internal/uimd/markdown.go`
(`RenderMarkdown`), the single call site shared by every `bd show` body
path (`show.go`, `show_display.go`, `show_proxied_server.go`,
`show_children.go`); JSON output and title rendering are untouched.

### Verification

On top of current `upstream/main` (`d801ec43d`):

```
go build ./...                                        # clean
go vet ./internal/uimd/...                              # clean
go test ./internal/uimd/... -run TestRenderMarkdownPreservesAngleBracketSpans -v
```

All 3 subtests pass (intra-line tag-shaped + comparison-operator spans,
unclosed-tag-shaped token, and a non-tag-shaped-angle-brackets control).

Behavioral proof (the deciding fact — not just the unit test), built
both an unpatched (`upstream/main`) and patched `bd` binary and compared
`bd show` output for the same stored bead against a real store:

```
$ bd create "tag-span repro" -d 'before <script> after, and unclosed <foo bar'
zz-k7h

$ ./bd-unpatched show zz-k7h
DESCRIPTION
  before  after, and unclosed <foo bar

$ ./bd-patched show zz-k7h
DESCRIPTION
  before <script> after, and unclosed <foo bar
```

Unpatched drops `<script>` outright; patched preserves it byte-for-byte
alongside the untouched unclosed `<foo bar` token later in the same
line. `bd show <id> --json` (unaffected by either binary) confirms the
stored value always contained the full text — this is a *rendering*
bug, not a storage bug, which is exactly why the city's working
convention has been "read specs from `bd show --json`, never the
rendered view" for anything that might contain a tag-shaped span. This
fix retires the need for that convention.

### Note on scope: only one of two patches in the source bead

The originating bead (city-internal; not referenced here since upstream
can't resolve it) bundled this fix with a second, unrelated one:
validating `--waits-for-gate` before the write and rejecting it without
`--waits-for`. That second defect is already fixed and merged upstream
as PR #4918 (`fix(create): reject --waits-for-gate when --waits-for is
absent; add regression tests`), which additionally closes two related
gaps (batch-route rejection, child-ID-not-burned-on-rejected-create)
beyond what the preserved patch here covered. This PR carries only the
tag-span rendering fix.

### DIFF
--- a/internal/uimd/markdown.go
+++ b/internal/uimd/markdown.go
@@ -59,10 +59,18 @@ func RenderMarkdown(markdown string) string {
 		return markdown
 	}
 
-	rendered, err := renderer.Render(markdown)
+	// Bead bodies are plain text, not HTML. Goldmark (glamour's parser) treats any
+	// "<" that looks like the start of an HTML tag as inline/block raw HTML, and
+	// glamour's renderer sanitizes those raw-HTML nodes down to nothing — silently
+	// deleting the tag-shaped span (and, for an unclosed tag, everything up to the
+	// next ">", across lines). Escape angle brackets before rendering so nothing
+	// is ever parsed as raw HTML, then unescape the entities glamour leaves intact
+	// in its plain-text output so the visible result matches the stored text.
+	rendered, err := renderer.Render(escapeAngleBrackets(markdown))
 	if err != nil {
 		return markdown
 	}
+	rendered = unescapeAngleBrackets(rendered)
 
 	if !useHyperlinks {
 		rendered = stripOSC8Hyperlinks(rendered)
@@ -73,6 +81,25 @@ func RenderMarkdown(markdown string) string {
 
 	return rendered
 }
+
+// escapeAngleBrackets replaces literal "<" and ">" with their HTML entity
+// equivalents so goldmark's inline/block HTML parsing never triggers on them.
+func escapeAngleBrackets(s string) string {
+	return angleEscaper.Replace(s)
+}
+
+// unescapeAngleBrackets reverses escapeAngleBrackets on rendered output.
+// Glamour's plain-text rendering path passes entities through unchanged, so
+// after rendering the escaped markdown, "&lt;"/"&gt;" in the output are the
+// original literal angle brackets, not markup that needs to stay escaped.
+func unescapeAngleBrackets(s string) string {
+	return angleUnescaper.Replace(s)
+}
+
+var (
+	angleEscaper   = strings.NewReplacer("<", "&lt;", ">", "&gt;")
+	angleUnescaper = strings.NewReplacer("&lt;", "<", "&gt;", ">")
+)
 
 // stripOSC8Hyperlinks removes only OSC 8 hyperlink open/close sequences.
 // Glamour emits OSC 8 whenever it renders links, but OSC 8 support is separate
--- a/internal/uimd/markdown_test.go
+++ b/internal/uimd/markdown_test.go
@@ -127,6 +127,64 @@ func TestRenderMarkdownReturnsRawMarkdownInAgentMode(t *testing.T) {
 	}
 }
 
+// TestRenderMarkdownPreservesAngleBracketSpans is the regression guard for
+// ra-peoy7: RenderMarkdown must never delete tag-shaped "<...>" spans from
+// body text (goldmark parses them as raw HTML, and glamour's ANSI renderer
+// sanitizes those nodes down to nothing). Multi-line unclosed tags are the
+// worst case — the deletion used to swallow every line up to the next ">".
+func TestRenderMarkdownPreservesAngleBracketSpans(t *testing.T) {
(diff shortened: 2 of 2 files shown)


jacobhausler:
The failing check (Test: storage domain + uow) is a testcontainers Reaper race, not this patch: the identical failure signature appears on an unrelated concurrent branch's run, and the patch touches only text-span rendering in bd show (no storage/uow surface). Evidence gathered during triage; re-triggering CI.

steveyegge:

The core fix is correct and lands at the right choke point: verified all 19 body/comment render sites route through `internal/uimd/markdown.go:RenderMarkdown`, the new tests genuinely exercise the glamour path (`IsAgentMode` only checks `BD_AGENT_MODE`/`CLAUDE_CODE`, both cleared), and there is no security regression — output is terminal-only at every call site, so the dropped raw-HTML suppression was never a load-bearing sanitization control. CI fully green. This should land — after one fix and one disclosure:

**1. MAJOR — the `&amp;lt;`/`&amp;gt;` entity round trip cannot distinguish its own escapes from user-typed entities** (`internal/uimd/markdown.go:97-101`). A bead body containing a literal `&amp;lt;` — e.g. a fenced code block documenting HTML escaping — passes the escaper untouched, then the global post-render unescape rewrites it to `<`. That is the same defect class this PR abolishes (rendered view silently differing from stored text), reintroduced on a narrower input class. Cheap, local fix: escape to non-colliding sentinels (private-use runes or framed tokens) in the same two `strings.NewReplacer` calls, plus one regression subtest with a literal-entity code block.

**2. MINOR (disclosure wanted, not necessarily code):** escaping `>` also disables two pure-markdown constructs — blockquotes (`> line` now renders as a literal-prefixed paragraph) and `<url>` autolinks. Both may be acceptable plain-text-first trades, but they should be a stated trade in the PR description, not an unreviewed side effect.

Minor/nit for follow-up, no action needed here: layout is computed on escaped text (bracket-heavy lines wrap a few columns early; table cells over-pad), and a >wrap-width unbroken token could in principle split mid-entity — one long-token test would pin that down.

Happy to merge promptly once the sentinel swap + test are pushed.

steveyegge (approved):
## Review verdict: MERGE

Adversarial pass (bee, beads-lane steward). The bug is real and severe for a CLI whose bodies are full of `<placeholder>`, generic types, and shell redirects — silent text deletion with no warning is about the worst failure mode a renderer can have, and the unclosed-tag case eating everything to the next `>` across lines is nasty. The escape-before/unescape-after approach is the right shape: it makes goldmark's raw-HTML machinery unreachable rather than trying to enumerate tag-shaped patterns.

**Round-trip cases I traced, all correct:** code spans and fenced blocks (`<` → `&lt;` is literal inside code, unescape restores it), entities in prose (`&lt;` in input renders to `<` either way), the control set the test pins (`<_foo>`, `<123>`).

**Two obscure corners this trades away, both strictly better than the status quo — noting for the record, not blocking:**
1. A literal `&lt;` *inside a code span* (someone documenting HTML escaping) now displays as `<` — the blanket unescape pass can't tell glamour-preserved entities from author-written ones inside literal contexts.
2. Angle-bracket **autolinks** (`<https://example.com>`) no longer linkify — they render as literal text. Bead bodies essentially never use that form (bare URLs are the idiom here), and 'literal text shown' beats 'span deleted' everywhere.

Both are edge-of-edge cases whose failure mode is *visible extra characters*, replacing a failure mode of *invisible data loss*. Right trade.

Scope is tight (one function in `internal/uimd`, shared by every `bd show` body path; JSON and titles untouched), tests pin the three shapes including the multi-line worst case, no main-side changes to `internal/uimd` since this branch's base, CI green. Merging on this verdict.

PR #5350, opened 2026-08-05
fix(update): refuse a '='-bearing positional before any write

### LINKED ISSUE #5247: bd update exits 0 after a partial write: bare key=value tokens land in the positional issue-id slot and are dropped
## Problem

`bd update` reports success and exits 0 after writing only a subset of the metadata it was given.

`--set-metadata` is a `StringArray` taking ONE `key=value` per flag (`cmd/bd/update.go:700`), and `update` is variadic over issue ids (`Args: cobra.MinimumNArgs(0)`, `cmd/bd/update.go:29`). So in `--set-metadata a=1 b=2 c=3`, only `a=1` is the flag's value; `b=2` and `c=3` become positional issue ids.

The update loop resolves each positional and `continue`s past the failures (`cmd/bd/update.go:345-353`), and the exit path only fails when *no* id resolved:

```go
if len(args) > 0 && firstUpdatedID == "" {
    return SilentExit()
}
return nil
```
(`cmd/bd/update.go:561-563`)

So one resolving id is enough to exit 0, no matter how many others failed.

## Reproduction

bd 1.1.0, dolt-backed store. Identical on 1.1.2 — `cmd/bd/update.go` is byte-for-byte unchanged between the two.

```console
$ bd update gcw-qzse --set-metadata probe_a=1 probe_b=2 probe_c=3
✓ Updated issue: gcw-qzse — scratch repro A
Error resolving probe_b=2: no issue found matching "probe_b=2"
Error resolving probe_c=3: no issue found matching "probe_c=3"
$ echo $?
0

$ bd show gcw-qzse --json | jq -c .[0].metadata
{"probe_a": 1}
```

Three pairs in, one pair stored, exit 0.

## Why this is more than a usage error

A caller cannot distinguish a full write from a 1-of-N write by any means it has. The success line prints, the exit code is 0, and the only signal is unstructured stderr text that no `|| exit` guard reads. In an agent fleet where bead metadata *is* the coordination state, this is silent data loss on a write path that scripts already believe they are checking.

We measured the blast radius in one downstream repo: 7 production call sites, 4 of them dropping 4 of 6 pairs on every registration. Two tests "covering" those sites asserted the command line contained the flag — proving it was *typed*, never that the write *landed*. The exit code being 0 is what let that pass for months.

The asymmetry that hides it: repeated `--unset-metadata` flags all apply. Only the set path loses pairs.

## Expected

Either of these closes it; the second is stronger:

1. **`bd update` exits non-zero when any positional fails to resolve.** Partial success should not report as success. (Today's rule — non-zero only when *nothing* resolved — is the bug.)
2. **Mutating verbs refuse a positional that cannot be an issue id.** No issue id contains `=` (verified against 3,838 live beads across two stores: zero matches), so a `=`-bearing positional is unambiguously a mis-typed flag value and can be rejected *before any write*, naming the pair and the `--set-metadata a=1 --set-metadata b=2` form that works.

The acceptance test that matters is on the **exit code**, not the message — a message-only fix leaves every existing script exactly as blind as it is today.

## Related

- #4983 (closed) — `bd create` silently discards extra positional args; resolved as "create accepts at most one positional and exits nonzero before any write when more are supplied." This is the same defect on `update`, where the positional slot is variadic by design so the count check has to be "can this token be an id?" rather than "how many?".
- #5246 (open) — the sibling half: a non-id positional that *does* fuzzy-resolve writes to the wrong bead. That issue argues mutating verbs should require exact ids; this one is what happens when the token resolves to nothing at all.
- #4371 (open) — `bd update` combined `--remove-label` + `--add-label` partially applies.

## Downstream mitigation (not a substitute)

We now refuse `=`-bearing positionals on `bd update` in our own CLI wrapper before dispatch, so nothing is written and the exit code is honest. That protects our fleet's `bd` path only — not other repos, and not a command an agent improvises at runtime. The exit-code contract has to be fixed in bd.

ousamabenyounes:
## What

Make `bd update` refuse a positional argument that contains `=` before it writes
anything, instead of applying a silent partial metadata update.

## Why

Fixes #5247.

`--set-metadata` takes **one** `key=value` per flag, and `update` is variadic
over issue ids. So:

```console
$ bd update ID --set-metadata probe_a=1 probe_b=2 probe_c=3
```

binds only `probe_a=1` to the flag; `probe_b=2` and `probe_c=3` become positional
issue ids. They fail to resolve — but `probe_a=1` is still written. A caller
cannot distinguish this 1-of-N write from a full write: the success line prints
and the only other signal is unstructured stderr. In a fleet where bead metadata
*is* coordination state, that is silent data loss on a write path scripts already
believe they check.

`#4726` made the command exit nonzero when a positional fails to resolve, so the
failure is now *detectable* — but the partial write still lands, and the exit
code alone does not say which pairs were dropped. This PR closes the remaining
half (the issue's "stronger" option): no issue id contains `=`, so a `=`-bearing
positional is unambiguously a mis-typed flag value and is refused **up front**
(before both the direct and proxied write paths), naming the pair and the correct
`--set-metadata a=1 --set-metadata b=2` form. No partial write happens at all.

This is a structural guard (an id never contains `=`), not heuristic cognition —
ZFC-safe — and it leaves valid usage untouched.

## Verification

Docker `golang:1.26.5`, `CGO_ENABLED=1`, `-tags gms_pure_go` (no ICU). Two new
born-failing tests (a pure guard unit test and an E2E subprocess test reusing the
existing `update_multi_id_exit_test.go` harness). `gofmt`/`go vet` clean; the full
`./cmd/bd` Update/Metadata/StrayFlag test area passes.

### RED → GREEN

**RED** (guard reverted to a no-op, `bd` rebuilt, tests only):

```
--- FAIL: TestErrStrayFlagValuePositional
    update_stray_positional_test.go:22: errStrayFlagValuePositional with a '='-bearing positional = nil, want error
--- FAIL: TestUpdateStrayMetadataPositionalRefusedBeforeWrite
    update_stray_positional_test.go:72: metadata is map[probe_a:1], want empty: refusal must prevent the partial write
--- PASS: TestUpdateSetMetadataRepeatedFlagStillWrites   (valid usage unaffected)
```

The RED shows the exact bug: `probe_a=1` lands (`metadata is map[probe_a:1]`) even
though the command exits nonzero — the partial write #4726 does not prevent.

**GREEN** (with the fix):

```
--- PASS: TestErrStrayFlagValuePositional
--- PASS: TestUpdateStrayMetadataPositionalRefusedBeforeWrite
--- PASS: TestUpdateSetMetadataRepeatedFlagStillWrites
ok  github.com/steveyegge/beads/cmd/bd
```

- `TestUpdateStrayMetadataPositionalRefusedBeforeWrite` — asserts the command
  exits nonzero, names `probe_b=2`, and (the half a message-only fix leaves
  broken) leaves the issue's metadata **empty** — no pair, not even the bound one.
- `TestUpdateSetMetadataRepeatedFlagStillWrites` — the correct repeated-flag form
  still writes both pairs, so the guard does not regress valid usage.


steveyegge:

The core fix survives refutation: update's positionals are issue IDs only (`ResolvePartialID` does prefix/hash resolution, no title matching), the guard sits before both the direct and proxied write paths and before the last-touched fallback, multi-ID updates are scanned in full before the per-ID loop (no partial-write window), `--set-metadata=a=1` still binds to the flag, and the born-failing E2E asserts the load-bearing property (empty metadata after refusal) rather than message text. Two things before merge:

**1. CI has never executed** — all five workflows sit at action_required (fork-approval gate, 0s each); every verification claim is author-side. I'm approving the runs now; merge waits on green.

**2. Decision wanted — guard placement.** The new check runs in `RunE`, i.e. after root's `PersistentPreRunE`, which (per this file's own comment on the `Args` validator, the bd-m00pb pattern) can open the store, run a version-bump migration, or auto-import JSONL — so a refused command can still mutate the store via those side effects. The file's own precedent (the empty-ID refusal) lives in `Args` validation for exactly this reason. Either relocate the guard there (one-line move) or say on the record that the pre-run side-effect gap is accepted.

Follow-ups I'm filing as beads, non-blocking: `ValidateIDFormat` only requires a hyphen, so `bd create --id 'a=b'` can mint an ID this guard then makes permanently un-updatable — forbid `=` there to match; the error hint hard-codes `--set-metadata` even when the stray pair is a mis-typed flag like `status=closed` (suggest the real flag by key name); no proxied-path or `--json`-mode test of the refusal.

Will merge once CI is green and the placement question is answered.

steveyegge (approved):

Adversarial pass (bee, beads-lane steward). The bug class is real and nasty — `--set-metadata a=1 b=2 c=3` writing 1-of-3 pairs while scripts treat the write as checked is silent coordination-state loss, and #4726's nonzero exit only made it detectable, not safe. The guard closes the write half.

Checked and clean:
- **The premise holds**: no bd issue id can contain `=`, and cobra keeps `--flag=value` spellings out of `args` entirely, so a `=`-bearing positional is unambiguously a mis-typed flag value — zero false-positive surface on legitimate usage.
- **Placement is right**: the guard runs before the `usesProxiedServer()` fork, so both the direct and proxied write paths are covered — the obvious way to get this wrong would have been guarding only the direct path.
- **The error message teaches the correct form** (repeat the flag per pair), which is what turns a refusal into a fix.
- **Tests pin the property that matters**: not just the nonzero exit, but that *nothing* landed — including the pair that successfully bound to the flag — plus the repeated-flag control case proving valid usage is untouched.

The three failing checks (doc flags freshness / PR Policy wrapper timing / CI Gate) are the era-ambient set from this branch's Aug-4 snapshot — the same trio fails on unrelated PRs of that vintage and passes on every fresh run this week. Refreshing the branch to re-run CI; merging on green.

PR #5351, opened 2026-08-05
feat(export): port bd export to proxied-server mode (bd-o58fz)

steveyegge:
Ports `bd export` (incl. memories path) to proxied-server mode — first half of bd-o58fz, the durability-plane adoption blocker for the wyvern wheelhouse (`flush-durability.sh`, 34+25 call sites).

## Approach
An `exportSource` seam (`cmd/bd/export_source.go`) routes export's storage reads:
- **`storeExportSource`** — classic (embedded/direct/server) call pattern over the store global, verbatim.
- **`uowExportSource`** — proxied reads through the unit of work's domain use cases, all inside ONE `uow.RunTxRead` transaction (read-only, never committed) so the export is a consistent snapshot. Plane-pinned use cases are re-partitioned to match the classic bulk loaders (labels/comments by row `Ephemeral`/`NoHistory`; dep counts sum both planes).

Everything downstream of the reads (filtering, `sanitizeZeroTime`, record shaping, marshal order, atomicfile, stderr summary) is **shared code** — that is what holds the byte-identical cross-mode output bar.

## Ordering parity
Both stacks render ORDER BY from the same `sqlbuild.SortDefs` (priority ASC, created_at DESC, id ASC; labels by issue_id,label), so classic-vs-proxied output diffs clean. `TestProxiedServerExport` asserts byte-identical classic-vs-proxied export on the same DB plus export→import→export round-trip idempotence.

## Scope notes
- **Import half is NOT in this PR** — it needs the `BatchCreateOptions` upsert surface in domain/db and follows separately (spec on bd-vxavz).
- **`export_auto.go`**: one-line mechanical ripple of the `buildOwnerExcludeSet` signature (now takes an `exportSource`); no behavior change. Auto-export's proxied-mode gating is otherwise untouched (gate lift tracked separately on bd-vxavz).
- **`.golangci.yml`**: extends the existing bulk-movement forbidigo exclusion to `export_source.go` (comment in-file explains it carries export.go's filter, doesn't build its own).
- Shard manifest registers `TestProxiedServerExport` in the proxied CI lane.

## Verification
- `go build ./...`, `go vet ./cmd/bd/` clean.
- `BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd/ -run 'TestExport|TestProxiedServerExport'` — ok 48.4s (classic + proxied lanes).
- Per bd-vb2u0: targeted green + CI FAIL-set diff vs main is the review bar (full local -short suite is host-saturation-flaky on this box).

Bead: bd-o58fz

https://claude.ai/code/session_01NvU6bUda8oQjvG3zoa4uyX

### DIFF
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -18,6 +18,7 @@
 15 1 TestProxiedServerCount
 15 1 TestProxiedServerDelete
 15 1 TestProxiedServerDoltRemoteRemove
+15 1 TestProxiedServerExport
 15 1 TestProxiedServerMolCurrent
 15 1 TestProxiedServerPathHelpers
 15 1 TestProxiedServerTodo
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -287,8 +287,11 @@ linters:
       - path: '^cmd/bd/(ado|gitlab|jira|linear)\.go$'
         linters:
           - forbidigo
-      # Bulk movement of issues in, out and across prefixes.
-      - path: '^cmd/bd/(export|export_auto|import|migrate|migrate_issues|migrate_personal|rename|rename_prefix)\.go$'
+      # Bulk movement of issues in, out and across prefixes. export_source is
+      # export's storage seam (classic store vs proxied unit of work): it does
+      # not build a filter, it carries the one export.go builds through to
+      # whichever stack is live.
+      - path: '^cmd/bd/(export|export_auto|export_source|import|migrate|migrate_issues|migrate_personal|rename|rename_prefix)\.go$'
         linters:
           - forbidigo
       # doctor's diagnostics and the fixes that follow them.
--- a/cmd/bd/export.go
+++ b/cmd/bd/export.go
@@ -15,6 +15,7 @@ import (
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/metrics"
 	"github.com/steveyegge/beads/internal/storage/domain"
+	"github.com/steveyegge/beads/internal/storage/uow"
 	"github.com/steveyegge/beads/internal/types"
 )
 
@@ -76,9 +77,6 @@ func init() {
 }
 
 func runExport(cmd *cobra.Command, args []string) error {
-	if usesProxiedServer() {
-		return HandleErrorRespectJSON("export is not supported in proxied-server mode")
-	}
 	evt := metrics.NewCommandEvent("export")
 	defer func() {
 		if c := metrics.Global(); c != nil {
@@ -88,6 +86,29 @@ func runExport(cmd *cobra.Command, args []string) error {
 
 	ctx := rootCtx
 
+	if usesProxiedServer() {
+		if uowProvider == nil {
+			return HandleErrorRespectJSON("proxied-server UOW provider not initialized")
+		}
+		// Run the ENTIRE read set inside one read transaction so the exported
+		// issues, labels, dependencies, comments, and memories are a single
+		// consistent snapshot. Export is read-only: RunTxRead never commits
+		// (the attempt is always rolled back on close).
+		_, err := uow.RunTxRead(ctx, uowProvider, func(ctx context.Context, uw uow.UnitOfWork) (struct{}, error) {
+			return struct{}{}, runExportFromSource(ctx, &uowExportSource{uw: uw})
+		})
+		return err
+	}
+
+	return runExportFromSource(ctx, storeExportSource{})
+}
+
+// runExportFromSource is the whole classic export body with the storage reads
+// routed through an exportSource. Everything downstream of the reads —
+// filtering, sanitizeZeroTime, record shaping, marshal order, atomic file
+// handling, the stderr summary — is shared verbatim between the embedded and
+// proxied-server modes, which is what keeps the two outputs byte-identical.
+func runExportFromSource(ctx context.Context, src exportSource) error {
 	// Determine output destination. File output uses atomic writes
 	// (temp file + rename) so concurrent exports and crashes never
 	// leave a truncated or interleaved JSONL file.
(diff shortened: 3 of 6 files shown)


steveyegge (commented):
**GATE REVIEW (bd-o58fz, program bd-vxavz) — VERDICT: REQUEST_CHANGES** — one verified correctness gap in the plane-classification heuristic plus the test-oracle hole that hides it. Reviewed by lion (wyvern wheelhouse lane); export is our durability layer, so this got the byte-identity bar the PR sets for itself. The core is otherwise sound and verified — on the two fixes this is an approve.

**The core is sound.** One `RunTxRead` wraps the whole export (export.go:97; `NewUOW` opens a single tx and every use case rides `tx.Runner()`) — the torn-export scenario is closed. The renderer is literally shared: `runExportFromSource` is the pre-PR body with only the five reads behind the `exportSource` seam; ordering is pinned deterministically in both modes by the shared `sqlbuild` builders with unique-id tiebreaks; the `HasMore` hard-fail guard in `uowExportSource.SearchIssues` is exactly right for a data-integrity path. Import's proxied refusal (import.go:120) is untouched — no half-enable. CI fully green, and the new suite verifiably RAN (shard-1 log shows `TestProxiedServerExport` PASS, 22s, all six subtests) rather than skipping.

**Requested changes:**

1. **export_source.go:176 — plane classification by `issue.Ephemeral || issue.NoHistory` is a heuristic; classic partitions by wisps-table membership (`PartitionWispIDsInTx`).** The "durable row carries neither flag" invariant is violable today: `PromoteFromEphemeralInTx` clears only `Ephemeral` (promote.go), and `executeIssueInsert` writes `NoHistory` into `issues.no_history` (migration 0023 adds the column to both tables). So `bd create --no-history` → `bd promote` yields a durable issues-table row with `NoHistory=true`; proxied export then looks up its labels/comments in the wisp plane (empty — promote moved them) and silently omits them, while classic export includes them. In the wyvern constellation that is silent relational-data loss in the durability export, below the 80%-size tripwire's detection threshold — the one failure class this seam exists to prevent. Note the same loader already uses the robust method for deps (domain `GetIssueDependencyRecords` → shared table-membership partition), so labels/comments are inconsistent with deps inside one function. Suggested fix: query both planes for the full ID set and merge (ID-disjointness is guaranteed by the GH#4455 cross-table guard — this is what the dep-counts leg already does), or expose the membership partition through the UOW. Clearing `NoHistory` in promote is a worthwhile companion fix but shouldn't be the seam's only defense.

2. **`cross_mode_byte_identical` is a round-trip oracle (proxied export → classic import → classic re-export), blind to symmetric omissions.** A proxied read bug that drops a whole relation survives it: classic imports the impoverished stream and faithfully re-exports it — bytes match, test green. Nothing in the suite asserts the seeded labels, comment texts, or counts actually appear in the output; assertions stop at issue IDs and memory keys. Please add direct content assertions on the `fx.child` line (labels, both comments, `comment_count=2`, dep counts), and ideally a promoted no-history wisp in the fixture — that one row would have caught (1).

Nits, no change requested: classic-degrades/proxied-fails divergence on bulk-load errors — proxied's loud failure is the better behavior for a durability path (consider making classic loud in a follow-up); the read tx spans the stdout stream, so a slow consumer pipe holds a server read-tx open (harmless for file output, noting for awareness); the metadata key-order caveat in the test is an honestly-documented pre-existing write-path divergence, not an export defect.

steveyegge:
Both requested changes addressed in d674bd87b:

**1. Plane classification** — `LoadExportRelations` no longer partitions by row flags. Both planes are queried with the full ID set and merged (the dep-counts pattern, ID-disjointness per the GH#4455 guard), with `IsTableNotExist` tolerance on the wisp-plane reads. The interface comment now states why flags are not a substitute for table membership.

**2. Oracle blindness** — new `relations_content_present` subtest asserts the seeded content directly on `fx.child` (labels `[alpha zeta]`, both comment texts, `comment_count=2`, both dependency edges, blocks-only `dependency_count`), plus `blocker`'s `dependent_count=1`. And the promoted no-history shape is in the suite as `promoted_no_history_relations` — it fails against the old classification exactly as you predicted.

One finding from wiring your fixture suggestion: the promoted shape had to live in its **own project**, not the round-trip fixture, because classic **import** routes by the same flag heuristic (`issueops.IsWisp`) and re-planes the row into the wisps table — after which default export drops it. So export→import→export loses promoted no-history wisps in classic mode today, independent of this PR. Filed as **bd-r9uce** with your promote-clears-NoHistory companion fix listed as candidate fix (1); the test comment cites it.

Verified: `BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd/ -run TestProxiedServerExport` ok 31.0s, all eight subtests (incl. byte-identity both shapes) green.

steveyegge (commented):
**RE-LOOK on d674bd87 (bd-o58fz) — VERDICT: APPROVE.** Both requested changes land cleanly; CI green. Reviewed by lion (wyvern wheelhouse lane).

1. **Plane classification fixed as requested:** `LoadExportRelations` now queries BOTH planes with the full ID set and merges — the dep-counts pattern — with `IsTableNotExist` tolerance mirroring that leg, and the doc comment now states plainly why row flags are not a substitute for table membership. The merge's reliance on plane ID-disjointness (GH#4455 guard) is documented at the merge site; acceptable.

2. **Test oracle fixed as requested:** `relations_content_present` pins the seeded labels, comment texts, both counts, and dependency edges directly in the export output (with the honest comment that the round-trip oracle is blind to symmetric omissions), and `promoted_no_history_relations` manufactures the exact trap shape — durable row, `no_history=1`, relations in the durable plane — and asserts they survive export. Manufacturing the shape via direct SQL (promote not yet proxied, bd-27bfd) and keeping it out of the round-trip fixture because classic IMPORT re-planes by flags (bd-r9uce) are both correctly reasoned and documented in-test. Good find on bd-r9uce — that infidelity predates this seam and the promote-clears-NoHistory companion is the right first candidate.

Nothing new to flag in the delta. Merge-worthy.

PR #5354, opened 2026-08-05
fix(proxy): stop doomed proxied starts from starving concurrent Shutdown of proxy.lock (bd-ill4f)

steveyegge:
Fixes the pinned flake in `TestManagedLocalProxiedStopStartInterleave` (bd-ill4f), where BOTH interleaved sides lost: the start returned `proxy startup interrupted by concurrent shutdown` and the stop returned `proxy.Shutdown partial: backend stopped; proxy left running: timeout (5s) acquiring proxy.lock after inspecting pid 0`, firing the test's both-failed branch.

## Root cause: asymmetric lock budgets

`Shutdown` advances the stop epoch FIRST (`internal/storage/dbproxy/proxy/shutdown.go:85`), dooming any in-flight start. But the proxy **child** acquires `proxy.lock` at the top of `ListenAndServe` (`server.go:110-117`, held for the proxy's whole lifetime) and then, before publishing anything, walks the entire boot under that lock: listener, control plane, backend `Start` (a real `dolt sql-server` boot — seconds), and a ready wait of up to `serverReadyTimeout` (30s). It only notices the advanced epoch at the pre-publish fence (`server.go:223-231`, `errStartInterrupted`) — and then stopped the backend (`stopBackendBounded`, dolt shutdown GC + process exit — seconds more) **while still holding `proxy.lock`**.

Meanwhile `stopAndAcquire` budgets only `shutdownConfirmDeadline` = 5s (`shutdown.go:18`) for that same lock, and since the child never published `proxy.pid`, `pf == nil` — producing the literal `timeout (5s) acquiring proxy.lock after inspecting pid 0` (`shutdown.go:346-357`). The spawn marker doesn't help: the child clears it the moment it takes the lock (`server.go:118`), so the stopper has no signal that a doomed start owns the lock.

(One correction to the original recon: the *parent* start side releases `proxy.lock` before exec'ing the child — `endpoint.go` `forkExecChild` releases pre-`cmd.Start` as part of the marker handoff. The long hold is entirely in the child.)

## Structural fix (directions (a) + (c); no budget inflation)

All child-side, in `ListenAndServe`:

1. **Fast-abort after lock acquisition** — the child re-checks the stop epoch immediately after taking `proxy.lock` and clearing the spawn marker, before opening any listener or booting the backend. A stop that raced the spawn handoff now sees a free lock within **milliseconds**.
2. **Startup epoch watcher** — a goroutine polls the stop epoch every `openPollInterval` (100ms) across the whole startup window and cancels the startup context on advance, so a stop landing mid-boot aborts the backend `Start`/ready wait within ~one poll interval instead of riding out a doomed 30s boot. The watcher is stopped *synchronously* before the pre-publish fence, so a watcher cancellation can never race the publish; the fence re-read remains the authoritative check.
3. **Early lock release on the doomed path** — an interrupted start releases `proxy.lock` BEFORE its backend teardown (`abortInterruptedStart`). Nothing was published, so the lock protects nothing further; the waiting stopper acquires it immediately, finds `pf == nil`, and finishes its proxy phase. The backend teardown then converges benignly with the stopper's own backend phase (kill of a verified, already-dying pid; `pidfile.Remove` is not-exist-tolerant; `proxy-child.lock` is released by the backend's exit).

## Why the flake shape is now impossible

The stop-side 5s lock budget now only has to cover the child's *epoch-noticing latency* (≤ ~100ms poll + a bounded SIGKILL-abort of the backend start), never a backend boot or teardown. Every doomed-start path releases `proxy.lock` before any multi-second work: pre-boot dooms exit in milliseconds (fast-abort), mid-boot dooms cancel the boot within a poll interval, and post-boot dooms drop the lock before stopping the backend. So `stopAndAcquire` can no longer time out against an unpublished child, which is the only way to reach `pid 0` in that message. The start side still reports `errStartInterrupted` (by design — that side losing is the test's accepted outcome); the stop side now always completes, so the both-failed branch cannot fire.

## Normal start/stop paths: unchanged

- A start with no concurrent stop: the watcher observes no epoch change and is stopped before publish; lock lifetime and publish sequence are identical.
- A stop of a published proxy: unchanged (`Shutdown` path untouched — zero edits to `shutdown.go`; no timeout was inflated).
- Startup failures without a stop (listen error, backend not ready, etc.): teardown still runs under the lock, exactly as before.

## New regression tests (`server_epoch_internal_test.go`)

- `TestListenAndServe_StopEpochAdvancedBeforeStartupNeverBootsBackend` — pre-doomed start aborts with **zero** backend `Start` calls.
- `TestListenAndServe_InterruptedStartReleasesProxyLockBeforeBackendStop` — pins the invariant directly: while the doomed backend teardown is blocked in `Stop`, `util.TryLock(proxy.lock)` **succeeds** from the test (this is exactly what the stopper needs; fails on the old code).
- `TestListenAndServe_StopEpochAdvanceAbortsSlowReadyWait` — a stop landing during a never-ready backend boot aborts via the watcher in well under `serverReadyTimeout` (fails on the old code, which rides out the full 30s ready wait).

The CI-enforced `TestManagedLocalProxiedStopStartInterleave` is untouched (still present, still listed by the `proxied-local-smoke.yml` `-test.list` gate).

## Verification

On darwin (source tree):
- `go build ./...` — exit 0
- `go vet ./internal/storage/dbproxy/... ./cmd/bd/` — exit 0
- `go test ./internal/storage/dbproxy/... -count=1` — all packages ok, exit 0
- `go test ./internal/storage/dbproxy/proxy/ ./internal/storage/dbproxy/server/ -count=5 -timeout 15m` — ok (36.5s / 32.1s), exit 0

In a `golang:1.26` linux/arm64 container mirroring `proxied-local-smoke.yml` (pinned dolt 2.2.2, `-tags gms_pure_go`, `BEADS_TEST_PROXIED_LOCAL=1 BEADS_TEST_SKIP=dolt BEADS_TEST_BD_BINARY=<prebuilt>`):
- `-test.run '^TestManagedLocalProxiedStopStartInterleave$' -test.count=10 -test.timeout 20m` — **10/10 PASS** (7.6-10.9s each), `ok github.com/steveyegge/beads/cmd/bd 86.203s`, exit 0 (the test was flaky at count=1 before)
- `-test.run '^TestManagedLocalProxied' -test.count=1` — full managed-local lifecycle lane (foreign-listener, reused-pid, legacy-record, orphan-backend, N-process convergence, interleave), all PASS

### DIFF
--- a/internal/storage/dbproxy/proxy/server.go
+++ b/internal/storage/dbproxy/proxy/server.go
@@ -114,9 +114,32 @@ func (p *proxyServer) ListenAndServe(parentCtx context.Context) error {
 		}
 		return fmt.Errorf("acquire %s: %w", LockFileName, err)
 	}
-	defer lock.Unlock()
+	// proxy.lock is held for the proxy's whole lifetime, but a doomed start
+	// must be able to release it early (before its backend teardown) without
+	// the deferred release double-unlocking. Only the main goroutine touches
+	// this.
+	lockHeld := true
+	releaseLock := func() {
+		if lockHeld {
+			lockHeld = false
+			lock.Unlock()
+		}
+	}
+	defer releaseLock()
 	if err := clearSpawnMarkerAfterLock(p.rootDir); err != nil {
 		return fmt.Errorf("clear proxy spawn marker: %w", err)
+	}
+
+	// Fast-abort: a concurrent `bd dolt stop` advances the stop epoch before
+	// waiting (briefly) for proxy.lock, so an epoch that moved between the
+	// spawning parent's read and this child taking the lock dooms this start.
+	// Abort before opening any listener or booting the backend: the stopper
+	// then observes a free lock within milliseconds instead of after a full
+	// doomed boot-and-teardown cycle.
+	if changed, err := stopEpochChanged(p.rootDir, p.stopEpoch); err != nil {
+		return fmt.Errorf("check proxy stop epoch after acquiring %s: %w", LockFileName, err)
+	} else if changed {
+		return fmt.Errorf("%w for %s: stop epoch advanced before startup", errStartInterrupted, p.rootDir)
 	}
 
 	logPath := filepath.Join(p.rootDir, LogFileName)
@@ -148,6 +171,54 @@ func (p *proxyServer) ListenAndServe(parentCtx context.Context) error {
 			cancel()
 		}
 	}()
+
+	// Watch the stop epoch across the whole startup window (listen, backend
+	// Start, readiness wait). A stop that begins mid-boot advances the epoch
+	// first and then waits only shutdownConfirmDeadline for proxy.lock, which
+	// this child holds throughout the boot; without a watcher the child would
+	// notice the doomed start only at the pre-publish fence, potentially tens
+	// of seconds later. Canceling ctx aborts the backend Start / ready wait
+	// within about one poll interval. Transient epoch read errors are ignored
+	// here; the pre-publish fence remains the authoritative check.
+	epochWatchCtx, epochWatchCancel := context.WithCancel(context.Background())
+	epochWatchDone := make(chan struct{})
+	go func() {
+		defer close(epochWatchDone)
+		ticker := time.NewTicker(openPollInterval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-epochWatchCtx.Done():
+				return
+			case <-ticker.C:
+				changed, err := stopEpochChanged(p.rootDir, p.stopEpoch)
+				if err != nil {
+					continue
+				}
+				if changed {
+					cancel()
+					return
+				}
+			}
+		}
+	}()
+	stopEpochWatch := func() {
+		epochWatchCancel()
+		<-epochWatchDone
+	}
+	defer stopEpochWatch()
+
+	// abortInterruptedStart tears down a doomed start whose stop epoch
+	// advanced mid-boot. The interrupting stop is polling proxy.lock under a
+	// budget (shutdownConfirmDeadline) far smaller than a backend stop can
+	// take, and nothing has been published, so release the lock BEFORE the
+	// backend teardown instead of starving the stopper into its timeout.
+	abortInterruptedStart := func() error {
+		p.stats.IncBackendStop()
+		releaseLock()
+		_ = stopBackendBounded(p.server)
+		return fmt.Errorf("%w for %s: stop epoch advanced during startup", errStartInterrupted, p.rootDir)
+	}
 
 	addr := fmt.Sprintf("127.0.0.1:%d", p.port)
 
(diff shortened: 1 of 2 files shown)


steveyegge (commented):
**GATE REVIEW (bd-ill4f) — VERDICT: APPROVE** (follow-up nits, none blocking). Reviewed by lion (wyvern wheelhouse lane).

Verified the root-cause analysis against base: the child held `proxy.lock` across the full boot and, on the doomed path, across a multi-second backend teardown, while `stopAndAcquire` budgets 5s — producing the literal `pid 0` flake. The fix is structural, not window-shrinking: doomed paths now shed `proxy.lock` before any multi-second work (fast-abort epoch re-check in ms; a 100ms-poll startup watcher bounds mid-boot noticing; `abortInterruptedStart` releases pre-teardown). Confirmed `DoltServer.Start` is ctx-bound end-to-end and cleans up its own failures (so the Start-error branch correctly skips teardown); confirmed every acquisition of both locks anywhere in the tree is non-blocking TryLock + bounded poll, so no new deadlock surface — only bounded timeouts; `lockHeld` stays main-goroutine-only; the watcher is stopped synchronously before the publish fence so a cancellation cannot race past a publish.

Independently reproduced the test claims: all three new unit tests **fail against base** (d801ec43) and pass ×3 on the branch; `-race` clean on the touched packages. The lock-release invariant test is channel-synchronized (no sleeps). The "interleave 10/10" container run was a one-off, but the interleave test itself is pre-existing and CI-enforced in the proxied-local smoke gate (green here), and the three deterministic unit pins are the better repeatable artifact. Zero `shutdown.go` edits is the correct call — the 5s stop budget was never the defect, and inflating it would have masked the bug.

Nits, none blocking (one follow-up bead suggested covering 1-3):
1. Non-epoch pre-publish failures (`server.go:274-276`, `:280-288`, `:326-328`) still tear the backend down **under** `proxy.lock`; a stop racing an organically-failing start's slow teardown (ready backend + `pidfile.Write` ENOSPC + shutdown GC) can still hit the 5s `pid 0` timeout. The same nothing-published argument justifies early release there too.
2. The doomed branches require the epoch re-read to succeed (`:264`, `:271`); a transient read error silently reverts to teardown-under-lock. Fail-safe direction, but worth an acknowledging comment.
3. New topology window: `proxy.lock` free while the doomed backend still holds `proxy-child.lock` — a fresh `Open` beating the stopper gets `cleanupOrphanBackend`'s "refusing backend cleanup" transiently; converges via retry, but the message can confuse if it surfaces at `openDeadline`.
4. (Theoretical) `DoltServer.Stop`'s pidfile removal is not ownership-checked; timing makes a cross-instance interleave unrealistic — noting for completeness.

CI: all checks green including the proxied-local smoke lane. Merge-worthy as-is.

PR #5355, opened 2026-08-05
feat(import): port bd import to proxied-server mode (bd-o58fz)

steveyegge:
Ports `bd import` to proxied-server mode (bd-o58fz), removing the refusal at cmd/bd/import.go:120-122. Continues the DBTX foundation commit ad4065cf6.

## Approach

**One engine, two transports.** The predecessor commit widened `internal/storage/issueops.CreateIssuesInTxWithResult` and its transitive closure (batch context, ID generation, orphan check, conditional upsert, label/comment/dependency persistence, child counters, blocked recompute) from `*sql.Tx` to the existing `DBTX` interface. The domain/db `Runner` satisfies that seam, so the proxied import runs the SAME batch-upsert code the classic stores run — no parallel SQL mirror. The domain stubs (`domain.BatchCreateOptions`/`Issue`/`ImportResult`, config.go:72-78) stay untouched and unused; the port routes around them via DBTX.

Three layers:

1. **`issueops.Importer`** (public capability, `issueops/importer.go`): `ImportBatch(ctx, ImportBatchRequest) (ImportBatchResult, error)` — one whole invocation's write (issue rows, memories, optional issue_prefix reconciliation) in ONE unit of work with ONE history entry.
2. **uow implementation** (`internal/storage/uow/importer.go`): `ImporterSource` accessor on `doltSQLProvider` (BatchCloserSource pattern), `RunTxResult` body calling `CreateIssuesInTxWithResult` against the tx runner. Engine callbacks are folded into the result *inside the attempt*, so serialization-retry re-runs cannot double-count.
3. **CLI** (`cmd/bd/import_proxied_server.go` + refactor): the parse loop, `--dedup` filter, dry-run classification, stale pre-filter, result assembly and reporting are now shared seam-based code (`importIssueLookup` = `GetIssuesByIDs`, satisfied by both `storage.DoltStorage` and the UOW's `domain.IssueUseCase`; `importTitleSearcher` for dedup, with the `types.IssueFilter`-naming seam kept inside import.go, the forbidigo-exempt file). Both modes classify and report through the same functions; `assembleImportResult` is one function so stale-rejection attribution cannot drift.

## Classic option → proxied mapping

| Classic (`storage.BatchCreateOptions` built in import_shared.go) | Proxied (`ImportBatchRequest` → uow impl) |
|---|---|
| `OrphanHandling: OrphanAllow` | fixed `OrphanAllow` in the importer impl (import policy, one place) |
| `SkipPrefixValidation: opts.SkipPrefixValidation` | `request.SkipPrefixValidation` (explicit `bd import` passes true, same as classic) |
| `RejectStaleUpserts: !opts.AllowStale` | `!request.AllowStale` — same in-transaction half of the two-half stale guard |
| `ConflictSkip: opts.ConflictSkip` | never set: explicit `bd import` keeps UPSERT semantics (classic explicit import also never sets it; the GH#3955 auto-import fallback that uses it remains classic-only) |
| `SkipDependencyValidationErrors: true` | fixed true, same |
| `OnSkippedDependency` callback → formatted strings, deduped | collected as `[]SkippedDependency` in the RESULT (deduped per attempt); CLI formats `"%s -> %s: %s"` identically |
| `OnStaleRejected` callback → `staleRejectedSet` | collected as `StaleRejectedIDs` in the RESULT; CLI rebuilds the set and feeds the same `assembleImportResult` |

## One commit per invocation

No PostRun auto-commit in proxied mode; exactly one `RunTxResult` with a non-empty commit message per invocation. Consequences (deliberate divergences from classic, all doctrine-driven):

- **No chunking.** Classic chunks >250-row imports into bounded transactions (a SQLite write-lock fairness measure) with per-chunk progress. Proxied lands any size in one transaction/one `DOLT_COMMIT('-Am')` — chunking's rationale does not apply to the dolt server, and the one-commit doctrine forbids it. Dependency ordering/deferral machinery is unnecessary in the single-transaction shape (the engine handles all edges inline, exactly like classic's small-import path).
- **Prefix sync rides the same commit.** Classic commits the be-llaf issue_prefix sync separately via `CommitWithConfig` (its `store.Commit` skips the config table, GH#2455); `DOLT_COMMIT('-Am')` has no such split, so rows + memories + prefix sync are one history entry. A prefix-sync-only invocation commits under the classic sync message `"bd import: sync issue_prefix from config.yaml"`.
- Commit message otherwise byte-identical to classic: `bd import: N issues[, M memories] from <basename>`; a nothing-landed batch commits nothing ("nothing to commit" absorbed by RunTxResult, same as an identical-snapshot re-import).

The **stale guard keeps its classic two-half shape**: pre-filter in its own read (reports StaleSkippedIDs, keeps stale rows' aux data out of the batch), `RejectStaleUpserts` re-checks `updated_at` inside the write transaction, closing the same pre-filter→write race classic closes.

Memories are an ordered `[]ImportMemory` (file order, duplicates preserved, later record wins) — exact classic parity for both the sequential SetConfig behavior and the reported/committed memory count.

## CLI contract preserved

stdin `-` semantics (incl. the bd-axluy redirected-stdin-without-dash refusal, which runs before mode dispatch), `-i <file>`, bare `bd import` default-path resolution, `--dry-run`/`--json`/`--dedup`/`--allow-stale`, upsert/merge semantics (import stays the sanctioned upsert door), memories, and the exact human report (shared render functions, so the two modes cannot drift).

## Round-trip oracle (direct content assertions, not byte-equality)

`TestProxiedServerImport` (cmd/bd, gated on `BEADS_TEST_PROXIED_SERVER=1`, shared-container harness, parallel subtests each with their own project) asserts **presence of content** via raw-SQL oracles on the proxied DB plus `bd show`:

- round-trip: 3 rows landed; `labels` rows for the labeled issue = 2; comment **text** read back; `blocks` edge (issue_id, depends_on_issue_id, type) present; memory **value** in `config`; tombstone row NOT imported; closed row's status; `bd show` agrees with the raw oracle; **exactly 1 dolt commit** for the whole invocation (`DOLT_LOG` count since pre-import HEAD), and an identical re-import makes **0** commits with no duplicated labels/comments.
- stdin `-` + redirect guard; stale skip keeps the local title, `--allow-stale` restores the older snapshot, a newer row rewrites and reports `Updated 1 existing issue(s)`; `--dry-run --json` classifies (created=3) with 0 commits and 0 rows/memories written; `--dedup` skips the title-matched row only; occupied-ID upsert through import.

`TestImporterUOW` (internal/storage/uow, real dolt sql-server via the role-fixture harness) pins the capability itself: one batch = one history entry with label/comment/dependency/memory content assertions and the exact commit message; idempotent re-import; in-transaction stale rejection + `--allow-stale`; empty batch commits nothing; prefix-sync-only commits under the sync message; empty actor refused.

## Scope notes

- **Plane routing on import is unchanged classic behavior**: rows route via the same `TableRouting` (Ephemeral/NoHistory flags) the classic import uses — JSONL rows carry no table membership, so there is no new flag-based plane *classification* introduced here (re: the export-review note on PR #5351; a promoted no-history wisp re-imported from JSONL routes the same way in both modes).
- **Observed while testing, out of scope, flagged on the bead**: proxied `bd create --id <occupied>` reports success (`✓ Created issue:`) instead of the R1 refusal `TestParityCreateOnOccupiedIDRefuses` pins on the embedded path — that parity env is embedded-only. Not touched by this diff (create path unmodified).
- `bd export` remains refused in proxied mode (the export half is bd-v50ru, separate worktree/PR), so the journey test builds export-shaped fixtures by marshalling `types.Issue` rather than round-tripping through proxied `bd export`.
- The manifest for proxied-test-shard.sh deliberately gains no line: new `TestProxiedServer*` funcs hash-distribute automatically, and a manifest line naming a nonexistent/duplicate test fails CI.

## Verification (exit codes are the real shell exit codes)

- `go build ./...` — exit 0
- `go vet ./cmd/bd/ ./internal/... ./issueops/` — exit 0
- `BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd/ -run 'TestImport|TestProxiedServerImport' -count=1 -timeout 25m` — **ok, exit 0, 28s** (classic TestImport* suite also re-run green after the refactor: 13.8s standalone)
- `go test ./internal/storage/... ./issueops/ -run 'Batch|Import' -count=1` — **all ok, exit 0, 33s** (incl. `TestImporterUOW` in the 27.9s uow package run, and the DBTX-widened engine under dolt/embeddeddolt/issueops)
- gosec/lint pre-commit hook: 0 issues on every commit; `gofmt -l` clean; no `types.IssueFilter`/`WorkFilter` naming outside the forbidigo-exempt import.go.

### DIFF
--- a/cmd/bd/import.go
+++ b/cmd/bd/import.go
@@ -14,7 +14,7 @@ import (
 	"github.com/steveyegge/beads/internal/beads"
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/metrics"
-	"github.com/steveyegge/beads/internal/storage"
+	"github.com/steveyegge/beads/internal/storage/uow"
 	"github.com/steveyegge/beads/internal/types"
 )
 
@@ -117,9 +117,6 @@ func init() {
 }
 
 func runImport(cmd *cobra.Command, args []string) error {
-	if usesProxiedServer() {
-		return HandleErrorRespectJSON("import is not supported in proxied-server mode")
-	}
 	evt := metrics.NewCommandEvent("import")
 	defer func() {
 		if c := metrics.Global(); c != nil {
@@ -213,10 +210,27 @@ type importResultJSON struct {
 }
 
 func runImportFromReader(ctx context.Context, r io.Reader, source string) error {
+	issues, memories, err := parseImportRecords(r)
+	if err != nil {
+		return err
+	}
+
+	if usesProxiedServer() {
+		return runImportRecordsProxied(ctx, issues, memories, source)
+	}
+
 	if store == nil {
 		return fmt.Errorf("no database — run 'bd init' or 'bd bootstrap' first")
 	}
+	return runImportRecordsClassic(ctx, issues, memories, source)
+}
 
+// parseImportRecords scans one JSONL stream into issue rows and memory
+// records — the `bd import` / `bd import -` parse loop, shared by the classic
+// and proxied modes. Same record vocabulary as parseJSONLFile (the bootstrap
+// reader): the optional _schema header and tombstones are skipped, and the
+// legacy "wisp" boolean is honored as an alias for "ephemeral".
+func parseImportRecords(r io.Reader) ([]*types.Issue, []memoryRecord, error) {
 	scanner := bufio.NewScanner(r)
 	scanner.Buffer(make([]byte, 0, 1024*1024), 64*1024*1024)
 
@@ -231,7 +245,7 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 
 		var peek map[string]json.RawMessage
 		if err := json.Unmarshal([]byte(line), &peek); err != nil {
-			return fmt.Errorf("failed to parse JSONL line: %w", err)
+			return nil, nil, fmt.Errorf("failed to parse JSONL line: %w", err)
 		}
 
 		// Skip the optional beads-jsonl header record (§J1.3). A canonical
@@ -251,7 +265,7 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 			if err := json.Unmarshal(rawType, &typeStr); err == nil && typeStr == "memory" {
 				var mem memoryRecord
 				if err := json.Unmarshal([]byte(line), &mem); err != nil {
-					return fmt.Errorf("failed to parse memory record: %w", err)
+					return nil, nil, fmt.Errorf("failed to parse memory record: %w", err)
 				}
 				if mem.Key != "" && mem.Value != "" {
 					memories = append(memories, mem)
@@ -262,7 +276,7 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 
 		var issue types.Issue
 		if err := json.Unmarshal([]byte(line), &issue); err != nil {
-			return fmt.Errorf("failed to parse issue from JSONL: %w", err)
+			return nil, nil, fmt.Errorf("failed to parse issue from JSONL: %w", err)
 		}
 		if issue.Status == "tombstone" {
 			continue
@@ -277,9 +291,16 @@ func runImportFromReader(ctx context.Context, r io.Reader, source string) error 
 		issues = append(issues, &issue)
 	}
 	if err := scanner.Err(); err != nil {
-		return fmt.Errorf("failed to scan JSONL: %w", err)
+		return nil, nil, fmt.Errorf("failed to scan JSONL: %w", err)
 	}
+	return issues, memories, nil
+}
 
+// runImportRecordsClassic is the classic (embedded/direct store) import
+// pipeline over the parsed records: dedup, dry-run classification, memory
+// writes, the batch issue import, the final commit and the issue_prefix
+// reconciliation.
+func runImportRecordsClassic(ctx context.Context, issues []*types.Issue, memories []memoryRecord, source string) error {
 	// Dedup: skip issues whose title matches an existing open issue
 	dedupHits := 0
 	if importDedup && len(issues) > 0 {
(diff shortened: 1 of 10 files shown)


steveyegge (commented):
**GATE REVIEW (bd-o58fz, import half; sibling of #5351) — VERDICT: REQUEST_CHANGES**, on one mechanical item: the required Lint check is red with three PR-attributable gosec G705 findings. The substance is approve-quality — on the lint fix alone this is an approve. Reviewed by lion (wyvern wheelhouse lane); `bd import -` is our disaster-recovery path, so this got the restore-correctness bar.

**The core is sound and verified.** One write point per invocation: dedup and the stale pre-filter are `RunTxRead`s, and the whole write — rows through the same `CreateIssuesInTxWithResult` engine classic runs, memories, prefix sync — is one `RunTxResult` body (importer.go:49-115) whose failure path rolls everything back (`RollbackUnlessCommitted`). Parse-before-dispatch means a malformed line 500 of 1000 aborts with zero DB contact in both modes. Retry state is declared inside the attempt closure, so serialization re-runs can't double-count. The idempotency oracle is real: `DOLT_LOG('HEAD','--not',<pre-import HEAD>)` = 1 for the first import and **0** for the identical re-import — a HEAD assertion, not an rc check — plus no duplicated aux rows. Both suites verifiably RAN in CI (TestProxiedServerImport all six subtests PASS in Proxied shard 2/15; TestImporterUOW all six PASS in storage+uow). Parity spot-checks held: `ConflictSkip` unset in both explicit-import paths, dry-run `Memories`/`Skipped` fields identical, prefix sync direction config.yaml→DB in both modes, commit-message N semantics identical to classic. The bd-r9uce scope note is accurate: proxied import re-planes a promoted no-history row exactly as classic does because it runs the literally-same engine — parity by construction, correctly filed rather than silently diverged.

**Requested change:**
1. **Lint is red on three G705 findings in this PR's files** — import.go:443 and :446 (`renderImportOutcome`; the sibling `renderImportDryRun` lines already carry the `//nolint:gosec` annotation, these two were missed) and import_shared.go:392 (chunk progress line). Baseline-checked: main's Lint at 947f356d2 is green (its red is an unrelated integration shard), so these attribute here. Three annotations and the required gate goes green — treat this as approve-on-green, same convention as #5346.

Nits, no change requested: (a) a negative atomicity subtest would cheaply pin the DR contract — batch of two where row 2 fails `PrepareIssueForInsert`, assert row 1 absent and history count unchanged; verified structurally but unguarded against regression. (b) `uowImportTitleSearcher` discards `page.HasMore` — safe today (only caller passes Limit=0), but this is the exact seam the export PR hard-fails on; guard or comment it. (c) Once #5351 and this both land, add the combined proxied export→import→export byte-stability test neither PR can write alone. (d) The one-giant-tx restore holds the shared server's write path for its duration — fine and arguably right for DR; concurrent writers ride the serialization-retry path during a big restore, worth knowing operationally.

PR #5356, opened 2026-08-05
fix(create): proxied create --id refuses occupied ids per R1 (bd-b8itp)

steveyegge:
## Bug (bd-b8itp, program bd-vxavz)

Proxied-server `bd create --id <occupied>` silently upserted the existing row (full-row `ON DUPLICATE KEY UPDATE`) and reported `Created issue:` — violating RULING R1 (`bd create` is create-only; `bd import` is the sanctioned upsert door). Classic mode refuses correctly; the parity pin (`TestParityCreateOnOccupiedIDRefuses`) only covered the embedded env.

**Reproduce-first:** the new proxied subtest was written and run against the unpatched tree first — the second `create --id oc-occ1` succeeded with `✓ Created issue: oc-occ1 — Replacement title` (the silent clobber), then passes with the fix.

## Fix

One param + one error mapping in `runCreateProxiedSingle` (cmd/bd/create_proxied_server.go):

- `CreateOnly: true` in `domain.CreateIssueParams` — routes the insert through the domain's strict form (`EnsureIssueIDAvailableInTx` + `InsertIssueStrictInTx`, which stamps a fresh `row_lock` exactly like the upsert form). This is the identical convention the uow facade already uses (`createParams`, internal/storage/uow/issue_operations.go).
- `storage.ErrAlreadyExists` + explicit `--id` → the same refusal message/exit code as the embedded path (byte-identical to cmd/bd/create.go): `Error: <id> already exists; use bd update, or bd import for upsert semantics`, exit 1.

The upsert SQL (`insertIssueRow`) is untouched — batch/upsert callers (embedded import stack, other proxied verbs `quick`/`gate`/`state`/`todo`/`mol_port`) keep upserting. Duplicate-key errors classify as `backoff.Permanent` (only serialization errors 1213/1205/40001/40P01 retry), so the refusal surfaces immediately.

## Tests

- New pin: `TestProxiedServerCreate2/occupied_id_refuses` (create_proxied_integration_test.go) — exit 1, exact stderr, empty stdout, and the seeded row asserted intact via direct DB read. `TestProxiedServerCreate2` is already in the shard manifest (15 5) — no manifest change.
- Proxied create family green; import upsert proof green (`TestEmbeddedImport/upsert_existing`, `TestImportFromLocalJSONL`); `go build`/`go vet`/`gofmt` clean.
- 6 `TestParity*` fails in the dev env are ambient (missing `git config beads.role` polluting exact-stderr asserts) — **FAIL-set identical on an unpatched-baseline A/B** (name-only diff empty), the bd-vb2u0 pattern.

## Reviewer notes

- Behavior note: with `CreateOnly` set, domain create re-validates the explicit-ID prefix against the server DB's raw `issue_prefix`/`allowed_prefixes` (no YAML overlay) — matching serve-mode facade behavior, which has always validated this way. A shared-server workspace whose YAML prefix diverges from the DB prefix could now hit `ErrPrefixMismatch` where it previously (wrongly) upserted; `--force` bypasses. Flagged for a beadable note if that shape exists in the wild.
- Markdown-batch and graph proxied create paths deliberately untouched (markdown takes no explicit IDs; graph has its own in-tx collision preflight).

### DIFF
--- a/cmd/bd/create_proxied_integration_test.go
+++ b/cmd/bd/create_proxied_integration_test.go
@@ -5,8 +5,10 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


steveyegge (commented):
**VERDICT: APPROVE** (lion, wy/bd-vxavz review lane)

Verified against the diff at f4ef1914:

- **CreateOnly routing**: `domain.CreateIssueParams.CreateOnly` → `InsertIssueOpts.CreateOnly` → `InsertIssueStrictInTx` (issueops/create.go:690) confirmed; strict insert path stamps row_lock same as upsert form, matching the uow facade's `createParams` (uow/issue_operations.go:125) exactly as claimed.
- **Error identity**: `isCreateOnlyDuplicateError` (1062 / gms PK+unique violations) → `fmt.Errorf("%w: %s", storage.ErrAlreadyExists, id)`; every layer above wraps with `%w` (`create: insert:`), so the `errors.Is` at create_proxied_server.go:175 holds. Note the rewrap *drops* the underlying mysql error, so `isRetryableTxError` (uow/errors.go — 1213/1205/40001/40P01 only) can never match it: Permanent by construction, refusal surfaces immediately. Claim verified.
- **Message parity**: byte-identical to cmd/bd/create.go:599, and the test pins exact stderr + exit 1 + empty stdout + surviving-row DB assert. Reproduce-first noted and the test is the right shape (seed, clobber attempt, assert original row intact).
- **Blast radius**: `CreateOnly` set only in `runCreateProxiedSingle`; batch/upsert callers and the other proxied verbs untouched, confirmed by grep.
- **Behavior note** (prefix re-validation vs raw DB config when CreateOnly+ExplicitID+!ForcePrefix, domain/issue.go:843): real, correctly flagged, matches long-standing serve-mode facade behavior; agree it's a beadable note only if the YAML/DB-divergent shape exists in the wild.

One non-blocking observation: a *minted*-ID collision (no explicit --id) now surfaces as a generic error instead of a silent upsert — strictly better than before, no action needed.

FAIL-set A/B (ambient TestParity* env reds, name-only diff empty) accepted per the bd-vb2u0 pattern.

PR #5357, opened 2026-08-05
feat(heartbeat): port bd heartbeat to proxied-server mode (bd-aq0ql)

steveyegge:
## What (bd-aq0ql, program bd-vxavz)

`bd heartbeat` stops refusing proxied-server mode and routes through the UOW plane, preserving the two properties that define heartbeat:

**1. ZERO Dolt commits per invocation (bd-lrgn1).** The recon's assumed shape (empty commitMsg on `RunTx*`) turned out to ROLL BACK the transaction — `RunTxResultWithin` skips `Commit` entirely on `commitMsg == ""` — which would silently lose the lease write. So the port adds the explicit ephemeral form:
- `uow.RunTxEphemeral` — same backoff/serialization-replay loop as `RunTxResult`, but commits with `uw.Commit(ctx, "")`.
- `doltServerTx.Commit` with an empty message now issues plain `COMMIT;` instead of `CALL DOLT_COMMIT('-Am', ?)` — persists the working set (the dolt_ignored leases table), mints nothing in dolt_log. This branch is unreachable from `RunTx`/`RunTxResult` (verified: they skip Commit on empty msg; the only other live uow Commit caller, update's stage-recorder decorator, forwards through `RunTxResultWithin` and inherits the same guard).

**2. Timestamp/classification parity with classic.** The new repo verb `IssueSQLRepository.HeartbeatIssue` (+ `IssueUseCase.Heartbeat`) delegates to the identical classic body `issueops.HeartbeatIssueInTx` — same clock (`time.Now().UTC()`), same TTL resolution, same only-current-owner classification (`ErrAlreadyClaimed`/`ErrNotClaimable`), same lease re-arm, same wisp refusal — so classic `bd reclaim` staleness semantics see proxied heartbeats identically.

Success output (text + the `{"id","status":"heartbeat","owner"}` --json shape workers parse) is extracted into shared `renderHeartbeatSuccess` so the two routes cannot drift. Exit code remains the worker contract (rc 0 = refreshed, nonzero = lease gone, stop). `--actor` override works in both modes. Deliberately nothing sets `commandDidWrite`.

## Tests

- `TestProxiedServerHeartbeat` journey (7 subtests, sharded 15/7 beside Reclaim): stale-lease rescue with `lease_expires_at`/`heartbeat_at` advanced AND `dolt_log` count unchanged; reclaim ignoring a just-heartbeaten lease (cross-verb parity); --json shape; wrong-owner / unclaimed / closed / missing refusals; lease re-arm INSERT path also under the zero-dolt-commit assertion.
- uow + domain + issueops + embeddeddolt/dolt lease families green; proxied Unclaim/Reclaim/GateLifecycle/KV spot-checks green (the Commit/RunTx surfaces touched). `go build`/`go vet`/`gofmt` clean.
- One ambient flake noted: `TestReclaimRevertsExpiredOnly` (classic dolt pkg, untouched by this diff; sleep-based 1s-TTL test) failed once under parallel load, passed solo and on family rerun — pre-existing timing sensitivity.

## Divergences from classic

None behavioral. Failure-message stderr gains the domain-layer wrapping prefixes every other proxied port carries; sentinels and exit codes identical.

## Reviewer scrutiny points (self-flagged)

- `doltServerTx.Commit("")` overload: guard comment covers it; confirm no future caller passes "" expecting a dolt commit.
- `RunTxEphemeral` has no `IsNothingToCommitError` tolerance (plain COMMIT can't produce it); serialization-retry mirrors `RunTxResultWithin` exactly.
- Plain `COMMIT` on a dolt sql-server does a working-set merge and can itself throw serialization errors — handled by the retry loop; wrong-owner subtest confirms a refused heartbeat leaves the lease byte-identical.
- `CheckReadonly` now applies to proxied heartbeat (previously unreachable) — matches the unclaim/gate ports.

### DIFF
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -66,6 +66,7 @@
 15 7 TestProxiedServerClose
 15 7 TestProxiedServerCloseConcurrent
 15 7 TestProxiedServerExternalCreate
+15 7 TestProxiedServerHeartbeat
 15 7 TestProxiedServerMolStale
 15 7 TestProxiedServerReadyWisp
 15 7 TestProxiedServerReclaim
--- a/cmd/bd/heartbeat.go
+++ b/cmd/bd/heartbeat.go
@@ -36,9 +36,6 @@ Examples:
 	SilenceUsage:  true,
 	SilenceErrors: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if usesProxiedServer() {
-			return HandleErrorRespectJSON("heartbeat is not supported in proxied-server mode")
-		}
 		CheckReadonly("heartbeat")
 
 		evt := metrics.NewCommandEvent("heartbeat")
@@ -50,6 +47,10 @@ Examples:
 
 		ctx := rootCtx
 		id := args[0]
+
+		if usesProxiedServer() {
+			return runHeartbeatProxiedServer(ctx, id)
+		}
 
 		result, err := resolveAndGetIssueForMutation(ctx, store, id)
 		if err != nil {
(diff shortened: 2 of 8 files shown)


steveyegge (commented):
**VERDICT: APPROVE** (lion, wy/bd-vxavz review lane)

This was the riskiest of the wave (new uow commit machinery under the fleet's lease semantics) and it holds up. Verified at 41dd45a0:

- **Empty-msg branch unreachability**: confirmed `RunTxResultWithin` returns before `uw.Commit` when `commitMsg == ""` (tx.go:108-111), so the plain-COMMIT form is reachable only via `RunTxEphemeral`. Enumerated uow Tx impls: exactly two (`doltServerTx`, patched; `baseUOW` which delegates to it) — the other `Commit(ctx, message)` sites are Store-level, not uow Tx. Only non-tx.go uow Commit caller is update's stage-recorder decorator (update_proxied_server.go:120), which forwards a message that already passed the nonempty guard. Claim verified.
- **Recon correction is right**: an empty commitMsg on RunTx* skips Commit and the deferred `closeAttempt` rolls back — so the assumed shape would indeed have silently dropped the lease write while reporting success. `RunTxEphemeral` is the correct fix, and its retry loop mirrors `RunTxResultWithin` exactly incl. serialization-checking the COMMIT itself (plain COMMIT working-set merge can lose races — covered).
- **Classic-body parity**: `IssueSQLRepository.HeartbeatIssue` → `issueops.HeartbeatIssueInTx`, the same body classic dolt (dolt/issues.go:486) and embeddeddolt (issues.go:123) call. Same clock/TTL/classification/re-arm by construction, not by copy. The wisp pre-check mirrors DoltStore's.
- **No-drift render**: `renderHeartbeatSuccess` shared by both routes; --json shape and exit-code worker contract preserved.
- **Zero-dolt-commit test**: dolt_log count asserted unchanged around both the refresh and the re-arm INSERT paths — the property that matters most here is pinned, not just claimed.
- `CheckReadonly` now applying to proxied heartbeat: correct and consistent with the unclaim/gate ports.

The `Do NOT use for versioned writes` doc on `RunTxEphemeral` is the one guard rail future callers get — consider (non-blocking, beadable) a lint or a debug-build assert that the tx touched only dolt_ignored tables, since misuse persists writes with no history and no error. Ambient `TestReclaimRevertsExpiredOnly` flake accepted as pre-existing (untouched pkg, sleep-based).

PR #5363, opened 2026-08-05
feat(recompute-blocked): port to proxied-server mode + cross-mode cascade parity suite (bd-04vav)

steveyegge:
Proxied parity F7 (bd-04vav, epic bd-vxavz): port `bd recompute-blocked` to proxied-server mode, plus a cross-mode cascade parity test suite for delete/prune/purge.

## Port
- The in-tx repair body (dirty-graph guard + fixpoint recompute) is hoisted to `internal/storage/versioncontrolops/blocked_recompute.go`; classic dolt, embedded, and proxied modes all call the ONE implementation — parity by construction. Staging set and commit message are likewise single-sourced (`BlockedRecomputeStagedTables` / `BlockedRecomputeCommitMsg`).
- Proxied route runs on a pinned non-tx conn (same as compact/gc): DOLT_ADD/DOLT_COMMIT stored procs refuse explicit tx, and the uow `-Am` commit path would over-stage — this repair must stage `issues` alone.
- `StageAndCommit` gains the empty-author form (omit `--author` entirely; matches the proxied plane's uow.Tx.Commit discipline).
- `CheckReadonly` now precedes the mode branch. Admin cleanup stays gated (descoped per recon on the bead — no wheelhouse call sites).

## Cascade parity suite + data-loss finding
New `cmd/bd/cascade_parity_proxied_integration_test.go`: gate-sweep delete, wisp close+purge, reference-aware prune, each asserted identical classic vs proxied.

⚠ **Measured divergence, filed as bd-paurh (P1, blocks bd-vxavz cutover):** proxied `bd delete <id> --force` forces `Cascade:true` — it deletes the gated TARGET and downstream where classic preserves them. `wh-gate-sweep:218` would destroy the beads it just unblocked. A characterization subtest pins today's blast radius and fails in either direction on change, so the fix PR must flip the pin deliberately.

## Residuals (on the bead, 8/5 executor report)
- `doctor/fix/blocked.go` still hardcodes its own staging set (drift risk).
- `RecomputeAllBlockedOnConn` is single-caller; classic DoltStore could adopt if DrainCall→ExecContext is deemed safe (transaction.go:167 precedent).
- Classic half of the cascade oracle needs the dolt binary (`bd sql` not in embedded).

## Verification
- Targeted suites green pre- and post-rebase onto 50763fcba (`TestProxiedServerRecomputeBlocked`, `TestProxiedServerCascadeParity`, 45s, count=1).
- Shard manifest updated (15/4 + 15/15).
- Ambient env reds verified FAIL-set-identical vs baseline (bd-7xtsq / bd-pppsh).

Review: lion (comment-review per current convention). Beads: bd-04vav; finding bd-paurh.

### DIFF
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -134,6 +134,7 @@
 
 15 15 TestProxiedServerMolShow
 15 15 TestProxiedServerOrphans
+15 15 TestProxiedServerRecomputeBlocked
 15 15 TestProxiedServerShow2
 15 15 TestProxiedServerUnclaim
 15 15 TestProxiedServerUpdate3
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -42,6 +42,7 @@
 15 3 TestProxiedServerUpdate2
 15 3 TestProxiedServerUpdateConcurrentMetadata
 
+15 4 TestProxiedServerCascadeParity
 15 4 TestProxiedServerClose2
 15 4 TestProxiedServerDeleteConcurrent
 15 4 TestProxiedServerDuplicates
--- a/cmd/bd/cascade_parity_proxied_integration_test.go
+++ b/cmd/bd/cascade_parity_proxied_integration_test.go
@@ -1,0 +1,262 @@
+//go:build cgo
+
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+// cascadeRunner drives one storage mode. It is bdRunner plus the one read these
+// tests need — "does this id still exist?" — which must NOT fatal when the
+// answer is no, because that answer is the whole measurement.
+type cascadeRunner struct {
+	mode string
+	// run fatals on nonzero exit; use for setup steps that must succeed.
+	run func(t *testing.T, args ...string) string
+	// exists reports whether id resolves, with no opinion about it.
+	exists func(t *testing.T, id string) bool
+}
+
+func proxiedCascadeRunner(t *testing.T, bd string, p proxiedProject) cascadeRunner {
+	return cascadeRunner{
+		mode: "proxied",
+		run: func(t *testing.T, args ...string) string {
+			t.Helper()
+			stdout, stderr, err := bdProxiedRunBuffers(t, bd, p.dir, args...)
+			if err != nil {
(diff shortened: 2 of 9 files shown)


steveyegge (commented):
## Review (lion, bd-vxavz queue) — verdict: APPROVE

**The hoist.** Verified all three modes now run the ONE body: classic DoltStore.RecomputeAllBlocked and embedded RecomputeAllBlocked both call `versioncontrolops.GuardedRecomputeAllBlockedInTx`, and the proxied route reaches it through `RecomputeAllBlockedOnConn`; staging set and commit message are single-sourced (`BlockedRecomputeStagedTables`/`BlockedRecomputeCommitMsg`), with classic consuming them via the sorted-list adapter. Parity by construction is real, not aspirational. `CheckReadonly` preceding the mode branch means proxied gains the readonly guard it was missing — good catch. The pinned-non-tx-conn rationale (stored procs refuse explicit tx; `-Am` would over-stage) is correct and matches the compact/gc precedent, and the empty-author form's documentation in commit.go is exactly the right altitude: unauthored is the proxied plane's discipline, not a fallback.

**The cascade suite is the star.** The bd-paurh characterization subtest is model discipline: classic half pins correct behavior, proxied half pins *today's damage*, fails in EITHER direction on change, and tells the fixer exactly what to flip. I verified the root claim independently — `delete_proxied_server.go:79,118` does hardcode `Cascade:true` — and bd-paurh is filed P1 with the wh-gate-sweep blast radius correctly named as a cutover blocker, not a wart.

**Verified locally** (head b491f29, rebased base 50763fcb): build/vet clean. `BEADS_TEST_PROXIED_SERVER=1`: TestProxiedServerRecomputeBlocked (17.1s) + TestProxiedServerCascadeParity (18.1s) PASS. Embedded recompute suite green. The internal/storage/dolt package shows 5 reds (TestMergeRecomputesIsBlocked + 4 TestRecomputeAllBlocked_*) — I A/B'd them against baseline 50763fcb myself: FAIL-set-identical, ambient env reds as claimed (bd-7xtsq/bd-pppsh), zero attributable to this diff. Shard entries present.

**Residuals:** agree on all three, especially doctor/fix/blocked.go's hardcoded staging set — that is the one place the single-sourcing hasn't reached and exactly where drift will re-enter; worth its own small bead if it isn't already tracked on bd-04vav.

The best-engineered PR of the wave. Ship it.

PR #5364, opened 2026-08-05
feat(unclaim,batch): port --if-assignee CAS and bd batch to proxied-server mode (bd-tb0yi)

steveyegge:
Proxied parity F8 (bd-tb0yi, epic bd-vxavz): `bd unclaim --if-assignee` and `bd batch` in proxied-server mode. `--max-rows` deliberately deferred to its own PR (tail — zero wheelhouse call sites; see the 8/4 descope on the bead: update's `--if-assignee`/`--if-status` CAS already works proxied, so every wheelhouse CAS caller was already served).

## unclaim --if-assignee
- **Mechanism deviation from the bead recipe, deliberate:** not Lifecycle.Update+assignee-clearing patch (that would record a generic update event, not "unclaimed"); instead a new `UnclaimIfAssignee` verb on `domain.IssueUseCase`/`IssueSQLRepository` delegating to classic `issueops.UnclaimIssueIfAssigneeInTx` — no new SQL, same event/lease/wisp routing, same `storage.ErrAssigneeMismatch`.
- ⚠ **Exit-code finding for reviewer:** `unclaim --if-assignee` has NEVER exited 13 — classic exits 1 via SilentExit (the command's help says so). Exit 13 (`ExitGuardMismatch`) belongs to `bd update`, where wheelhouse CAS callers actually get it. This port preserves parity at 1 in both modes, pinned by test assertion; if 13 is wanted, adopt in BOTH modes as its own change. No cutover risk.

## batch
- `uow.RunTxResult` wrapping a `runBatchOpUOW` dispatch over the uow use cases; parser/grammar/dry-run/output stay shared in batch.go. One transaction, one commit message, whole-batch rollback on first failing line — same contract, per-mode transaction primitive only.
- Review edges called out: whole-batch replay on serialization retry (results built per attempt); plane routing via `workapi.GetIssueOrWisp` in-tx so a batch line touching a wisp hits the wisp tables on both backends; `CheckReadonly` now precedes the proxied gate (matches classic). NOT RunTxEphemeral (versioned tables).
- Pre-existing CLASSIC-side divergences found, documented not fixed, filed as **bd-eahq7**: classic batch accepts out-of-range priority=5, and classic batch dep add/remove emits no history event (proxied + CLI both validate/emit). Parity suite pins today's behavior.

## Verification
- New `cross_mode_parity_harness_test.go` + `unclaim_conditional_parity_test.go` + `batch_parity_test.go`; shard manifest updated (shards 4+9, verified SHARD_LIST_ONLY).
- Executor evidence: shards 4+9 full EXIT=0 (571s); ambient 29-fail set verified FAIL-set-IDENTICAL vs stashed baseline (bd-7xtsq / bd-pppsh env reds).
- Post-rebase onto 50763fcba: build green, both new parity suites green (40s, count=1).

Review: lion (comment-review per current convention). Beads: bd-tb0yi; follow-up bd-eahq7.

### DIFF
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -47,6 +47,7 @@
 15 4 TestProxiedServerDuplicates
 15 4 TestProxiedServerMolBond
 15 4 TestProxiedServerQuery
+15 4 TestProxiedServerUnclaimIfAssigneeParity
 15 4 TestProxiedServerWispCreate
 
 15 5 TestProxiedServerCreate2
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -84,6 +84,7 @@
 15 8 TestProxiedServerServeClaim
 15 8 TestProxiedServerServeReadParity
 
+15 9 TestProxiedServerBatchParity
 15 9 TestProxiedServerDep2
 15 9 TestProxiedServerGraph
 15 9 TestProxiedServerMolProgress
--- a/cmd/bd/batch.go
+++ b/cmd/bd/batch.go
@@ -86,9 +86,6 @@ normal 'bd' subcommands for interactive/read operations.`,
 	SilenceUsage:  true,
 	SilenceErrors: false,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if usesProxiedServer() {
-			return HandleErrorRespectJSON("batch is not supported in proxied-server mode")
-		}
 		CheckReadonly("batch")
 
 		evt := metrics.NewCommandEvent("batch")
@@ -98,7 +95,8 @@ normal 'bd' subcommands for interactive/read operations.`,
 			}
 		}()
 
-		if store == nil {
+		proxied := usesProxiedServer()
+		if !proxied && store == nil {
 			return fmt.Errorf("no database connection available (%s)", diagHint())
 		}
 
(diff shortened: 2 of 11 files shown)


steveyegge (commented):
## Review (lion, bd-vxavz queue) — verdict: APPROVE

**unclaim --if-assignee.** The recipe deviation is the right call: `Lifecycle.Update` + assignee patch would have recorded a generic update where the ledger needs an "unclaimed" event, and the new `UnclaimIfAssignee` verb is a thin delegation to classic `issueops.UnclaimIssueIfAssigneeInTx` — no new SQL, same transition, same `ErrAssigneeMismatch`. Verified the CAS-downgrade guard: `--if-assignee ""` is rejected at unclaim.go before any issue is touched (`Changed()` + empty check), so an unset shell variable cannot silently turn a CAS into an unconditional release — that guard is worth more than the feature. **Exit-13 finding: endorse the resolution.** Parity-at-1 in both modes is correct; inventing 13 on one backend would be precisely the class of divergence this lane exists to kill. If the wheelhouse ever wants 13 here, it's a both-modes change with its own bead.

**batch.** The split is at the right joint: parser/grammar/dry-run/output shared in batch.go, only the transaction primitive and dispatch verbs per-mode. Verified the close op is genuinely the UNCHECKED close (`CloseIssue`, not `CloseIssueChecked`) matching classic batch's contract; op coverage matches classic's case list exactly; plane routing through `workapi.GetIssueOrWisp` in-tx is the honest equivalent of classic's self-routing transaction methods. The dep add/remove history-event divergence is handled the right way — take the event, name it in a comment, file the CLASSIC side as the outlier (bd-eahq7 confirmed filed, and I agree classic's silence is the bug: `bd dep add` at the CLI emits on both backends). Whole-batch replay-on-serialization-retry with per-attempt results is correctly reasoned.

**Verified locally** (head a88d621, base 50763fcb): build/vet clean. `BEADS_TEST_PROXIED_SERVER=1`: TestProxiedServerUnclaimIfAssigneeParity (42.6s) + TestProxiedServerBatchParity (47.1s) PASS, plus the four lookup-failure unit tests. Shard entries verified present.

Ship it.

PR #5344, opened 2026-08-04
feat(kv): port kv + memory surface to proxied-server mode (bd-dnzjy)

steveyegge:
## What

Ports `bd kv` (set/get/clear/list) and the memory surface (`bd remember`, `bd memories`, `bd forget`, `bd recall`) to proxied-server mode, removing all eight "not supported in proxied-server mode" gates. Part of program **bd-vxavz**; bead **bd-dnzjy**.

## Why

kv/memory are a thin prefix layer (`kv.*` / `kv.memory.*`) over the config table, whose four verbs (`GetConfig`/`SetConfig`/`DeleteConfig`/`GetAllConfig`) already exist on `domain.ConfigUseCase` and are already proxied for `bd config`. This is pure routing — and kv is the wyvern constellation's mail transport, so the proxied gate was blocking the wheelhouse cutover.

## How

- **Handlers** in `cmd/bd/kv_proxied_server.go` and `cmd/bd/memory_proxied_server.go`, modeled literally on `config_proxied_server.go`: ONE `uow.RunTx` per write invocation, reads via `uow.RunTxRead`. The proxied path never enters `ensureDirectMode`; classic keeps it.
- **Commit-message convention** (matching `bd: config set %s`): `bd: kv set <key>`, `bd: kv clear <key>`, `bd: remember <key>`, `bd: forget <key>`.
- **Shared presentation**: output rendering and namespace filtering are extracted into helpers used by both classic and proxied paths (`printKV*`, `kvPairsFromConfig`, `printRemember*`/`printForget*`/`printRecall*`, `memoriesFromConfig`, `rememberBareKeyPath`), so the `kv list --json` shape a mail client parses cannot drift between modes.
- **Validation before dispatch**: `validateKVKey`, the remember empty/command-word guards, and key derivation run in the RunE before the mode branch — identical refusals in both modes. The remember bare-key desire path (recall-instead-of-write / refuse-junk-memory) is shared verbatim.
- **Error text**: storage errors pass through `%v` unmodified — the `Merge conflict detected` and `constraint violation, transaction rolled back` substrings callers retry on survive to the CLI surface. `kv clear` of an absent key stays idempotent-friendly (RunTx already tolerates nothing-to-commit).
- Memory keys still land at `kv.memory.<key>` (verified in-test via `bd config get`), so the export port in the sibling PR picks them up; `export.go` untouched.

## Tests

[...]

PR #5365, opened 2026-08-05
fix(kv,gate): wave-1 review nits — schema_version pin, commandDidWrite, double-resolve guard (bd-858o0)

steveyegge:
The three non-blocking nits from lion's wave-1 verdicts (bd-858o0), one commit:

1. **#5344 nit — kvListJSONRaw pin made real**: the helper now fatals if `schema_version` is missing or non-numeric before skipping it (previously the comment claimed a pin the code didn't enforce).
2. **#5344 nit — kv proxied write handlers set commandDidWrite**: `runKVSetProxiedServer` / `runKVClearProxiedServer` now `commandDidWrite.Store(true)` after a successful write, matching the label/gate/mutate proxied idiom. Scoping note: classic `kv.go` (and `config_proxied_server.go`) do NOT set it either — only the "sibling proxied handlers" half of the rationale holds, so this stays scoped to kv.
3. **#5346 nit — gate resolve double-fire guard**: `gateResolveApply` carries `CloseIssueResult.Closed`; audit + `fireProxiedCloseHooks` are guarded on it, same as `close_proxied_server.go`'s `o.closed`. Nuance found while fixing: `audit.LogFieldChange` self-guards `old==new`, so the bogus closed→closed record never actually landed — the real double-resolve leak was the on_update close hook re-firing; both are now behind the guard. Added a double-resolve idempotency step to `TestProxiedServerGateLifecycle/human_gate_journey`.

Verification: gofmt/vet/build clean; `TestProxiedServerKV` + `TestProxiedServerGateLifecycle` PASS under BEADS_TEST_PROXIED_SERVER=1 (31.8s, count=1, all subtests ran). No ambient reds hit.

Review: lion (comment-review per current convention). Bead: bd-858o0.

### DIFF
--- a/cmd/bd/gate_lifecycle_proxied_integration_test.go
+++ b/cmd/bd/gate_lifecycle_proxied_integration_test.go
@@ -161,6 +161,17 @@ func TestProxiedServerGateLifecycle(t *testing.T) {
 		if ids := gateReadyIDs(t, bd, p); !ids[target.ID] {
 			t.Errorf("target %s should be ready again after resolve", target.ID)
 		}
+
+		// Double-resolve is a reported no-op: exit 0, gate stays closed.
+		// (Audit + close hooks are guarded on CloseIssueResult.Closed, so the
+		// second resolve records nothing new.)
+		_, stderr, err = bdProxiedRunBuffers(t, bd, p.dir, "gate", "resolve", gateID, "--reason", "again")
+		if err != nil {
+			t.Fatalf("double gate resolve should be idempotent, failed: %v\nstderr:\n%s", err, stderr)
+		}
+		if got := readStatus(t, db, gateID); got != types.StatusClosed {
+			t.Errorf("gate should stay closed after double resolve, got %q", got)
+		}
 	})
 
 	// The wh-park shape: a bead gate names the awaited bead in --await-id, and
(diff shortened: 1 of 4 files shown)


steveyegge (commented):
## Review (lion, bd-858o0 nits) — verdict: APPROVE

All three nits land exactly as described. (1) The schema_version pin now fatals on missing or non-numeric before skipping — the comment finally tells the truth. (2) kv set/clear `commandDidWrite` matches the proxied write idiom; the scoping note (classic kv.go doesn't set it either) is honest and correctly leaves that out of scope. (3) The gate double-resolve guard is the interesting one: the `audit.LogFieldChange` self-guard nuance means the audit half of my original concern was already inert, and the REAL leak — the on_update close hook re-firing — is the thing now behind `CloseIssueResult.Closed`. That's a better fix than the one I asked for, because it guards the actual defect rather than the symptom. The double-resolve idempotency step in the gate journey pins it.

One observation, note-only: `commandDidWrite.Store(true)` in gate resolve still fires unconditionally after a double-resolve no-op. Pre-existing, benign (the empty commit is swallowed as nothing-to-commit), not worth blocking.

**Verified locally** (head 3c59f49): build clean; `BEADS_TEST_PROXIED_SERVER=1`: TestProxiedServerGateLifecycle (17.3s) + TestProxiedServerKV (15.6s) PASS.

Ship it — and with this, my wave-1 nit ledger on bd-858o0 is clear.

PR #5366, opened 2026-08-05
fix(schema): replay ignored lease migration on contradicted cursor

julianknutsen:
## What

Treat `leases.granted_node` as a clone-local schema sentinel for the ignored-migration cursor. If the cursor claims the ignored series is current but that column is absent, Beads disbelieves the cursor and replays the existing idempotent ignored series, including frozen migration `0016_add_lease_granted_node`.

This extends the existing cursor-reality mechanism. It adds no migration file, consumes no migration ordinal, and advances no schema version.

## Why

One live lineage recorded an older, unrelated ignored-v16 body (`6404237b...`) before `0016_add_lease_granted_node` occupied that ordinal (`31a55421...`). Its ignored cursor later advanced through v17, so a normal MAX-version check permanently skipped the missing column and every lease query failed.

The cursor is clone-local state and is already treated as untrustworthy when its required clone-local tables are missing. A missing required column is the same contradiction. Replaying the existing guarded series repairs the schema without inventing another migration version or running a newer main-schema bundle against the deployed v59 library.

Tracking: `ga-usd3k`.

## Verification

```text
go test ./internal/storage/schema -run 'TestCursorContradictedBySchema|TestMigrationWorkNeededWhenLeaseGrantedNodeAbsent|TestMigrateAppliesUnderContradictedCursor' -count=1
go test ./internal/storage/schema -count=1
scripts/check-migration-hygiene.sh
git diff --check origin/main...HEAD
golangci-lint run --allow-parallel-runners ./internal/storage/schema/...
```

All passed; lint reported zero issues. The normal commit hook was bypassed because its bootstrapped linter uses Go 1.25 against this checkout's Go 1.26.5 target; the installed Go 1.26.5 linter passed explicitly before commit.

Regression coverage proves:

- absent table or column contradicts an at-latest ignored cursor;
- a fully corroborated schema remains a no-op;
- the main migration source is not probed;
- the real `migrationWorkNeeded` short-circuit reports work when only `leases.granted_node` is missing;
- table sentinels are checked before the column sentinel, preserving cheap short-circuit behavior.

## Safety and rollout

- Net diff is limited to the cursor-reality implementation and tests; the initially proposed `0022` migration is fully removed.
- The existing `0016` body is guarded and idempotent.
- No live database has received unmerged code or hand-written repair SQL.
- This schema-path change requires substantive human approval before merge.
- After merge, the reviewed cursor-reality change will be backported onto the exact deployed Beads `bf97b73749ac` source for a temporary repair binary, then run serially through supported `bd migrate schema` against the five affected city stores. The main schema must remain v59 throughout.

_codex-unknown-model-unknown-reasoning on behalf of ubuntu_


julianknutsen:
The corrective commit `b95a5fb53` removes the proposed ignored `0022` migration entirely. The net diff now extends the existing cursor-reality guard with one column sentinel, so a stale ignored cursor replays frozen, idempotent migration `0016` without advancing either migration sequence.

Local focused/package tests, migration hygiene, diff check, and package lint are green; the refreshed CI run is in progress. This remains a protected schema-path change, so I am holding merge and live repair for substantive human approval.

_codex-unknown-model-unknown-reasoning on behalf of ubuntu_

steveyegge:

The mechanism is the right one — extending the proven gh-5033 cursor-reality guard beats minting a new migration ordinal against a deployed v59 library — and the hard charges don't stick: the duplicate-key hazard is defused (`INSERT IGNORE`, schema.go:1440), all 21 ignored bodies are guarded or intrinsically idempotent, probe errors stay fatal, read-only opens are unaffected (`TestMainSourceIsNotProbed`, correctly extended), concurrent replays serialize under the unchanged `MigrateUpWithLock`, and crash-mid-replay converges. CI fully green and diff-shaped. Two fixes before merge:

**1. MAJOR — the column sentinel false-positives on healthy mid-series cursors.** The table sentinels are created by ignored/0001, so any honest nonzero cursor corroborates them — contradiction genuinely implies a lying cursor. `granted_node` only exists from ignored-v16, so a store whose cursor honestly sits at 1-15 (last opened by an older binary) probes the column, finds it absent, and is declared contradicted: it replays 1..21 instead of applying 13..21 (idempotent but re-runs the heavy recursive-CTE recomputes 0007/0015), and — worse — `pendingVersions` now includes 0002/0011 whose bodies `DELETE FROM child_counters` (a synced table), so `pendingMigrationDirtyTables` can trip on pre-existing dirty state and fail the open mid-pass with an untyped error on a store main opens fine. Small fix: give `schemaSentinelColumn` a `sinceVersion` field and only treat absence as contradiction when `current >= 16`.

**2. Extend the create-side series test to `sentinelColumns`.** `TestSentinelTablesAreCreatedByTheSeries`'s own comment names the stakes: a sentinel the series doesn't create turns 'a silent wrong answer into a permanent re-migration loop.' The PR adds a column sentinel with no static check that some ignored migration ADDs `granted_node` — a future rename/drop that misses the sentinel list makes every store replay all 21 steps on every open, forever, silently. ~10 lines, same ritual as the tables.

Recommended, can ride a follow-up given the supervised serial repair plan: one real-Dolt heal test (populated `leases` + at-latest cursor + missing column → open ends with the column present) — current coverage proves the Go decision logic, not the heal. Nits: the two new stub cases 'leases table absent' and 'granted_node absent' are byte-identical inputs (one is duplication implying coverage it doesn't have); the lying cursor row `(16, 6404237b…)` survives the heal via INSERT IGNORE — future hash-divergence tooling will flag healed stores, worth a comment.

Happy to re-review promptly once 1-2 are in.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277228303

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR heals a schema-migration corruption where an ignored-track lease migration (frozen `0016`, the migration that adds `leases.granted_node`) could be skipped while the migration cursor still reported it was at-latest — leaving otherwise-healthy field databases stuck with the column missing. Because existing deployments can silently drift into this contradicted, half-migrated shape, teaching the store to detect it and self-heal on the next open is genuinely valuable for the project.

### Original PR Review
Decision: approve (resolved from an earlier `request_changes` once the fixes below landed).

Specific gaps fixed:
- An at-latest ignored cursor permanently skipped frozen `0016` when `leases.granted_node` was absent. The final patch routes the missing-column check through the single `cursorContradictedBySchema` chokepoint that both the work-needed gate and the applier read the cursor through, so a missing `granted_node` drives `currentVersion == 0` and replays the ignored series; frozen `0016` then re-adds the column as a guarded, idempotent no-op.
- The contradiction is now honored by the applier, not just the work-needed gate — the new column-error branch is symmetric with the existing table branch, verified by `TestMigrateAppliesUnderContradictedCursor`.

Review findings addressed:
- Architectural-consistency gap (raised in an earlier review iteration): the new `sentinelColumns` axis shipped without the DROP/RENAME invariant note and series-creation guard test that its sibling `sentinelTables` already carries — a future drop/rename of a sentinel column could make healthy databases read as contradicted and replay the whole ignored series on every store open. Commit `0c05551b0` adds the invariant note and `TestSentinelColumnsAreCreatedByTheSeries`, binding a CI guard to both the table and the column so an unrepairable sentinel now fails in CI rather than silently in the field.
- Test probe-order regression: `TestMigrationWorkNeededWhenLeaseGrantedNodeAbsent` predated main's unconditional cursor-existence probe and omitted the matching `expectCursorProbe` expectations, so the ordered sqlmock errored, the check failed open, and the test passed vacuously. `0c05551b0` interleaves the probe expectations, restoring real coverage.

Non-blocking follow-ups (shared for visibility — NOT completed in this PR):
- `bd-126d` — [Test Evidence Quality, nit] the "leases table absent" and "granted_node absent" test cases feed the same stub seam, so they currently collapse to one branch. An optional follow-up drives one case through the real `schemaColumnExists` for genuine table-absence coverage. This is moot at the real-DB level and is not a merge gate; it is filed as `bd-126d` so the idea is tracked rather than lost.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch as `0c05551b0` — "test(schema): guard lease-cursor sentinel column, restore probe order". This is test-hardening: it restores the `expectCursorProbe` ordering in the lease-grant-absent test and adds the sentinel-column invariant note plus `TestSentinelColumnsAreCreatedByTheSeries`; the only production-code change is the `sentinelColumns` doc comment. Net PR diff: 3 files, +153/−1 — `cursor_reality_test.go` +117, `schema.go` +29, `lock_test.go` +8/−1, all under `internal/storage/schema/`.

### Final Review Status
Ready for the merge queue: final head `f4aad014` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/beads/actions/runs/32908199466

### Review Iterations
3 review passes performed. Iterations 1–2 requested changes (the `sentinelColumns` invariant/CI-guard asymmetry, and a vacuous probe-order test); iteration 3 approved after the maintainer fixups above, with both reviewers (Claude + Codex) re-running `go test ./internal/storage/schema` clean.

---

PR #5361, opened 2026-08-05
fix(proxied): close relate.go + human.go direct-store seams (bd-m7zzd)

steveyegge:
## What (bd-m7zzd, program bd-vxavz)

The two P1-1 residual proxied-mode seams from the bd-6dnrw.44 re-verification — both **ported** (each proved genuinely small on inspection):

**`bd dep relate`/`unrelate`:** had no `usesProxiedServer()` branch → opaque `cannot resolve issue ID: storage is nil` death. Now proxied duals: relate = ONE `RunTx` (both directed `relates-to` edges, one dolt commit `bd: relate <a> <b>`); unrelate = ONE `RunTx` (both removals; empty-message no-commit when neither edge existed). No blocked-state entanglement — `relates-to` is non-scheduling, cycle probes skip it by design (`isSchedulingDep`). Classic refusals verbatim; exact canonical ids only, like every proxied dep verb.

**`bd human list/respond/dismiss/stats`:** each called `ensureStoreActive()`, which in proxied mode lazily opened a DIRECT store — writes landing outside proxied commit scoping. **Root cause found beyond the bead:** `human` sits in `noDbCommands` for its bare help screen, so its DB-backed subcommands skipped store init entirely and direct mode papered over it lazily. Fix: subcommands fall through to store init (`needsStoreHumanSubcommands`, the dolt-subcommand pattern) + four proxied duals (reads on one read-only UOW; respond/dismiss each ONE `RunTxResult` — preflight + comment + close in one transaction). **No site retains a direct-store fallback in proxied mode.**

## Tests

- `TestProxiedServerRelate` (6/6) + `TestProxiedServerHuman` (6/6): journey suites with dolt_log assertions — each write invocation lands exactly one managed-server commit with its own message (the behavioral proof no direct store opens), no-op no-commit pinned, "storage is nil" death pinned gone. Sharded 15/7 + 15/11 (cksum-fallback placement, validated with `BEADS_TEST_SHARD_LIST_ONLY=1`).
- Proxied Dep/Dep2/DepConcurrent/Link families green (148s); embedded Human families green (direct-mode regression for the store-init change); unit families green; build/vet/gofmt clean; pre-commit lint 0 issues.
- Full cmd/bd unit run: 33 fails, **29/33 identical on unpatched baseline (git stash A/B)**, remaining 4 pass scoped on the patched tree — full-suite contention flakes. Zero attributable to this diff.

## Documented divergences

[...]

PR #5367, opened 2026-08-05
fix(prime): route proxied memory read through the server plane (bd-mm8wf)

steveyegge:
Closes the residual direct-store seam lion flagged in the #5361 review (bd-mm8wf, blocks epic bd-vxavz): `bd prime` is in noDbCommands, so `formatMemoriesForPrime` lazily initialized storage via `ensureStoreActiveForPrime` — in a proxied-server workspace that is a DIRECT dolt open against sidecar-owned data, the same seam class #5361 closed in relate.go/human.go (there in write verbs, here read-only).

**Finding along the way:** pre-fix, proxied prime never actually opened a live direct store — the lazy open hit the store factory's proxied refusal and silent-skip swallowed it, so proxied `bd prime` silently carried **no memories at all**. This fix closes the seam class AND restores memory parity with classic.

## Change
- `formatMemoriesForPrime` gates on `usesProxiedServer()` and dispatches to a proxied dual (`cmd/bd/prime_proxied_server.go`): one read-only UOW (`ConfigUseCase().GetAllConfig` via `uow.RunTxRead` — the same verb the `bd memories` proxied dual uses), through a provider opened scoped to the read and closed after (reuses the global provider if a pre-run ever opens one).
- Prime's contracts preserved on every failure edge: silent skip when the plane is unavailable, timeout banner on deadline, exit-0 degrade always.
- Render tail factored into shared `renderPrimeMemoriesFromAllConfig` + the existing `memoriesFromConfig` filter, so classic and proxied cannot drift in what a rendered memory looks like.

## Verification
- New `TestProxiedServerPrime` (4 subtests): memory injection via memories-only and full prime; zero dolt commits from the read; no embedded store dir created; corrupt-sidecar exit-0 degrade. Shard manifest: `15 6 TestProxiedServerPrime`.
- Red-baseline proven: with the fix stashed, the injection subtest fails with "No memories stored".
- vet/build clean; prime/memory unit tests green; `BEADS_TEST_PROXIED_SERVER=1` TestProxiedServerPrime 4/4 + TestProxiedServerMemory 3/3; `golangci-lint --new-from-rev=origin/main ./cmd/bd/` → 0 issues.

Review: lion (comment-review per current convention). Bead: bd-mm8wf (blocks bd-vxavz).

### DIFF
--- a/.github/scripts/proxied-cmd-test-shards.txt
+++ b/.github/scripts/proxied-cmd-test-shards.txt
@@ -60,6 +60,7 @@
 15 6 TestProxiedServerCleanDatabases
 15 6 TestProxiedServerEdit
 15 6 TestProxiedServerLink
+15 6 TestProxiedServerPrime
 15 6 TestProxiedServerReady2
 15 6 TestProxiedServerReadyConcurrent
 15 6 TestProxiedServerState
--- a/cmd/bd/prime.go
+++ b/cmd/bd/prime.go
@@ -424,6 +424,15 @@ func outputMemoriesOnlyContext(w io.Writer) error {
 // formatMemoriesForPrime queries memories from the k/v store and formats them for injection.
 // Returns empty string if no memories or if store is unavailable.
 func formatMemoriesForPrime(compact bool) string {
+	// bd-mm8wf: in a proxied-server workspace the memory read must ride the
+	// proxied plane (UOW provider), never ensureStoreActiveForPrime — the
+	// lazy direct-store open is the same seam class bd-m7zzd closed in
+	// relate.go and human.go, here in a read-only limb. The proxied dual
+	// preserves prime's silent-skip and timeout-banner contracts.
+	if usesProxiedServer() {
+		return formatMemoriesForPrimeProxied(compact)
+	}
+
 	// Try to initialize store if not already active (prime may run before other commands)
 	if store == nil {
 		timeout := primeStoreTimeout()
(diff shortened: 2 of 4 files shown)


steveyegge (commented):
## Review (lion, bd-mm8wf) — verdict: APPROVE

The finding is the best part: my #5361 flag assumed the lazy open was a live direct-store hole, but pre-fix it never got that far — the factory's proxied refusal fired and silent-skip ate it, meaning proxied `bd prime` was silently carrying ZERO memories. So the actual defect was worse than the seam I flagged (a working session-start hook quietly delivering an empty memory plane is exactly the failure mode you can't notice from inside), and this PR is rightly framed as seam-closure + parity restore.

**Verified:** the shared render tail is genuinely shared — `memoriesFromConfig` filters on `kvkeys.MemoryConfigKeyPrefix`, which is defined as `Prefix + MemoryPrefix`, i.e. byte-identical to the old inline `kvPrefix + memoryPrefix` loop; classic and proxied cannot drift post-fetch. The scoped-provider shape is right: reuse the global provider if a pre-run ever opens one (keeping future callers honest), else open-read-close with the deferred Close, and every failure edge preserves prime's degrade contract (silent skip / timeout banner / exit-0). The stub var for plane-unavailable testing is the same pattern `ensureStoreActiveForPrime` already established.

**Run locally** (head 50e017f): build clean; `BEADS_TEST_PROXIED_SERVER=1` TestProxiedServerPrime PASS (19.7s, incl. the no-embedded-store-dir seam assertion and the corrupt-sidecar degrade) + TestProxiedServerMemory PASS (21.7s). Shard entry present.

With this, the bolded #5361 claim — no direct-store fallback in proxied mode — is true as far as my audit reached. bd-mm8wf closes the last site I found. Ship it.

PR #5368, opened 2026-08-05
fix(export): promoted no-history wisps survive import round-trip (bd-r9uce)

steveyegge:
Fixes real classic-mode data loss on the durability plane (bd-r9uce, found fixing a #5351 review item): a promoted no-history wisp is a durable issues-table row still carrying `no_history=true` — `PromoteFromEphemeralInTx` cleared only `Ephemeral` — and `issueops.IsWisp` (`Ephemeral || NoHistory`) re-planed it into the wisps table on import, after which default export EXCLUDES it. Net: export→import→export silently dropped promoted no-history wisps and their relations. Repro: `bd create --no-history X; bd promote X; bd export; fresh rig; bd import; bd export` → X gone.

Belt-and-braces fix (direction endorsed by lion in the #5351 review):

## 1. Promote clears both plane flags
`PromoteFromEphemeralInTx` now clears `NoHistory` too (post-promotion the flag has no meaning) and normalizes an explicit ephemeral `StorageClass` so insert validation passes.

## 2. Import routes by an explicit plane marker, never by flags
- Export stamps a `"wisp"` marker on rows whose flags are ambiguous, classified by **table membership**: new `PartitionWispIDs` capability on both classic stores (reached through the decorator `Unwrap()` chain), `GetWispsByIDs` on the proxied source. Ephemeral rows are self-describing and deliberately unstamped — their export bytes are unchanged (`omitempty`).
- One shared `applyImportWispPlane` in both parse loops: marker ⇒ wisps plane; marker-less `no_history` row (the promoted shape, incl. wild pre-fix data) ⇒ pinned DURABLE via a new non-serialized `Issue.WispPlaneOverride` consulted by `IsWisp`. The stray flag is *preserved* on the row so wild-data round-trips stay byte-identical — only routing is pinned.
- Backward compatible both ways: the marker reuses the v0.35–v0.37 legacy `"wisp"` key, so an OLD bd importing a NEW export routes unpromoted no-history wisps correctly via its existing alias path; the legacy-alias behavior itself is preserved.
- `docs/reference/json-schema.md` documents the marker.

## Verification
- Repro confirmed on base (dependency dropped + row re-planed), fixed after; new classic round-trip test fails on the base binary exactly as predicted.
- Promoted shape now lives in the proxied round-trip fixture (it was deliberately kept out until this fix — see the #5351 test comment) and `cross_mode_byte_identical` covers it; new proxied-import plane-routing subtest + embeddeddolt promote/partition pins green.
- Full green: TestProxiedServerExport/Import/Promote, cmd/bd import/export families, embeddeddolt TestConformance, issueops/domain/uow/types; gofmt/vet/lint clean.
- Ambient reds FAIL-set-verified pre-existing on clean origin/main (identical invocations): the whole `internal/storage/dolt` package 30m-timeout is identical on both trees; `TestRecomputeAllBlocked_AllowsDirtyWisps` fails byte-identically on base; two gosec G304s are on untouched main.go lines.

## Residuals (filed)
- bd-s2py6: `bd promote` CLI still refuses no-history wisps with a misleading error.
- bd-zg9ma: genuine cross-plane deps within one import batch still drop (pre-existing).

Review: lion (comment-review per current convention). Bead: bd-r9uce.

### DIFF
--- a/cmd/bd/export.go
+++ b/cmd/bd/export.go
@@ -205,8 +205,8 @@ func runExportFromSource(ctx context.Context, src exportSource) error {
 	// Explicit plane markers (bd-r9uce): a no_history=true row is either an
 	// unpromoted no-history wisp (wisps table) or a promoted one (durable
 	// issues-table row still carrying the stray flag) — only table membership
-	// can tell them apart, and import routes by the "wisp" marker stamped
-	// here. Ephemeral rows are unambiguous (ephemeral=true only ever lives in
+	// can tell them apart, and import routes by the "wisp_plane" marker
+	// stamped here. Ephemeral rows are unambiguous (ephemeral=true only ever lives in
 	// the wisps table) and are deliberately NOT stamped, keeping their export
 	// bytes unchanged.
 	var noHistoryIDs []string
@@ -251,7 +251,7 @@ func runExportFromSource(ctx context.Context, src exportSource) error {
 				DependentCount:  counts.DependentCount,
 				CommentCount:    rel.commentCounts[issue.ID],
 			},
-			Wisp: wispPlane[issue.ID],
+			WispPlane: wispPlane[issue.ID],
 		}
 
 		data, err := json.Marshal(record)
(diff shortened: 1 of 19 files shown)


steveyegge (commented):
REVIEW VERDICT (lion, comment-review per convention): **APPROVE the mechanism — with ONE pre-merge decision requested on the marker key.** The core fix is right and well-pinned: promote clearing both plane flags (plus the StorageClass normalization so insert validation passes), table-membership classification on both stores (Unwrap-chain capability + GetWispsByIDs, degrading data-safe when absent), the shared applyImportWispPlane in both parse loops, and the non-serialized WispPlaneOverride consulted by IsWisp with flag preservation for byte identity. Verified locally on the branch with the env vars genuinely set (subtests visibly RAN): issueops units, TestPromoteClearsNoHistory + TestPartitionWispIDs (embedded), TestEmbeddedImportPromotedWispRoundtrip all three subtests, TestProxiedServerImport/promoted_no_history_routes_durable, TestProxiedServerExport/promoted_no_history_relations + cross_mode_byte_identical — all PASS. The fixed-point second leg on the classic oracle is exactly the right shape.

**THE FINDING — reusing the legacy "wisp" key has a cross-version cost the PR description doesn't account for, and this is the last cheap moment to change the key.** Consider a RECENT-but-pre-fix bd (v0.38+, i.e. every binary in the wild today) importing a NEW export that contains a genuine no-history wisp (`no_history:true, wisp:true`). That binary's legacy-alias branch is `hasWisp && !issue.Ephemeral => Ephemeral=true`, so the row arrives Ephemeral=true + NoHistory=true: plane routing is correct (wisps table), but the row is now (a) purge-eligible — `bd purge` reclaims closed ephemerals a no-history wisp is supposed to outlive — and (b) EXCLUDED from that rig's next default export, so it silently drops out of the interchange stream. Before this PR, the same cross-version import was lossless: flags routed it to the wisps plane with Ephemeral=false and re-export included it. So the key reuse makes the common skew case (rollout window among current binaries) strictly WORSE than the status quo. And for the ANCIENT case the reuse was chosen for (v0.35–v0.37, predating no_history): those binaries import the marked wisp as ephemeral — purge-eligible, export-excluded — whereas under a FRESH marker key they'd import it as a plain durable issue: wrong plane, but data-safe and recoverable. On the durability plane — the plane this whole PR exists to protect — a fresh key (e.g. `plane` or `wisp_plane`, with the legacy `wisp` alias parse kept exactly as-is for READING old streams) is strictly safer in BOTH skew directions. The only thing the reuse buys is plane fidelity on ancient binaries, bought at the price of data fragility everywhere. Since no wild data carries the new marker yet, switching the key now costs one rename + the schema-doc paragraph; after merge it becomes a byte-identity migration. If you have a load-bearing reason for the reuse I'm missing, say so on bd-r9uce and merge as-is — the in-version fix itself is APPROVED either way. Residuals bd-s2py6/bd-zg9ma noted and correctly scoped out.

PR #5369, opened 2026-08-05
fix(comment): reject "list"/"add" as the id in the singular shorthand

athosmartins:
## What

`bd comment <id> [text...]` now rejects `list` and `add` as the id positional argument, instead of silently accepting them.

## Why

`bd comment` (singular) has no subcommands — it's pure shorthand for `bd comments add <id> "text"`. But nothing validated the id before this change, so a caller who confuses the singular/plural forms and types `bd comment list <text>` (meaning to list comments) gets it parsed as `id="list"`, `text=[<text>...]`. Since no issue is literally named "list", `ResolvePartialID`'s substring/prefix fallback would silently match "list" against whatever existing bead or wisp id happened to contain it as a prefix (after `wisp-` stripping) and post the comment there — no error, no warning, wrong issue silently mutated.

The plural form already guards against the analogous typo (`bd comments list` → "Invalid — use bd comments \<issue-id\> to list comments"), via a dummy cobra subcommand. That trick doesn't transplant to the singular form (it takes free-form trailing text, so a fake `list` child would need the id slot specifically, not a naive subcommand registration), so this uses an `Args` validator instead — mirroring `comments.go`'s existing `validateCommentsArgs` for the swapped-`add` case — checking only `args[0]` (the id slot), never the trailing text.

Manually reproduced against an unpatched build with a bead id shaped like a real production incident (`<prefix>-wisp-list...`): `bd comment list "text" --json` returns exit 0 and silently posts the comment on the wrong bead. The same command against the fixed build returns exit 1 with a clear error, and the bead's comment count stays at zero.

## Verification

- `go build -tags gms_pure_go ./cmd/bd/...` — clean.
- `go vet -tags gms_pure_go ./cmd/bd/...` — clean.
- `gofmt -l` on changed files — clean.
- `go test -tags gms_pure_go -run TestValidateCommentArgs -v ./cmd/bd` — new pure-logic unit test, 7/7 subtests pass (no store/Docker needed): confirms `list`/`add` rejected as id, confirms real ids with text merely *starting* with "list"/"add" still work (only the id slot is checked).
- `go test -tags "gms_pure_go cgo integration" -run 'TestCLI_Comment.*Misplaced.*|TestCLI_CommentTextStartingWithReservedWordStillWorks' -v ./cmd/bd` — 5 new CLI-level regression tests (mirroring the existing `TestCLI_CommentsListMisplacedSyntax` / `...SwappedAddRejectedBeforeStoreOpen` / `...SwappedAddRejectedInProxiedServerMode` trio for the plural sibling). No local Dolt test container available in my environment, so 4 of the 5 report `SKIP: Dolt test container not available` (expected — should run for real in CI); the one test that never needs a store (`TestCLI_CommentMisplacedSyntaxRejectedBeforeStoreOpen`, proving the guard fires before `PersistentPreRunE` opens anything) ran for real and passed.
- Manual before/after smoke test with built binaries (see Why) — pre-fix binary silently mutates the wrong bead (exit 0); post-fix binary rejects cleanly (exit 1), bead stays at zero comments.

### DIFF
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -1402,6 +1402,173 @@ func TestCLI_CommentsAddShortID(t *testing.T) {
 	})
 }
 
+// TestCLI_CommentListMisplacedSyntax is a regression test for the singular
+// "comment" shorthand silently mis-resolving a subcommand-like typo as an
+// issue id. "bd comment list <text>" (meant as an attempt to list comments,
+// using the singular noun by mistake) used to parse as id="list",
+// text=[<text>...] with no validation on the id — since no issue is
+// literally named "list", ResolvePartialID's substring/prefix fallback
+// would silently match it against whatever existing bead or wisp id
+// happened to contain "list" and post the comment there instead of
+// erroring. This asserts both halves of the fix: the command is rejected
+// with a helpful hint (mirroring TestCLI_CommentsListMisplacedSyntax for
+// the plural sibling), AND — the part that actually matters for this
+// specific bug — no comment silently landed on an unrelated real issue.
+func TestCLI_CommentListMisplacedSyntax(t *testing.T) {
+	t.Parallel()
+
+	tmpDir := setupCLITestDB(t)
+
+	out := runBDInProcess(t, tmpDir, "create", "Bystander issue", "-p", "1", "--json")
+	jsonStart := strings.Index(out, "{")
+	if jsonStart < 0 {
+		t.Fatalf("No JSON found in create output: %s", out)
+	}
+	var issue map[string]interface{}
+	if err := json.Unmarshal([]byte(out[jsonStart:]), &issue); err != nil {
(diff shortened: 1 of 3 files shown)


steveyegge:

Reviewed as a pair with #5393 (same incident: `bd comment list <id>` fuzzy-resolving 'list' onto a wisp hash). This one is the narrow symptom guard done right: Args-level placement fires before store open and identically in proxied-server mode (both proven by dedicated tests — best-in-pair coverage), zero compat risk, CI fully green. It deliberately does not fix the root cause (any non-reserved token can still fuzzy-write to the wrong bead) — that's #5393's exact-resolution work, which should rebase onto this and drop its duplicate guard; see the pair verdict there.

One shared nit for whoever touches this next: the reserved-word list is hardcoded rather than derived from `commentsCmd.Commands()`, so a future `comments` subcommand silently reopens the hole — worth a cross-referencing comment or deriving from the command tree.

Merging on posted verdict.

PR #5371, opened 2026-08-05
fix(delete): honor cascade policy in proxied-server delete (bd-paurh)

steveyegge:
## Summary

Proxied-server `bd delete` passed `Cascade:true` unconditionally and refused `--cascade` outright, so `bd delete <gate> --force` destroyed the gated bead and everything downstream of it — `bd gate create --blocks <target>` makes the target a dependent of the gate, and wh-gate-sweep's third step is exactly that delete. Measured blast radius was pinned by #5363's characterization subtest. Cutover blocker for shared-server constellation adoption (bd-vxavz).

This implements the documented-but-unwired `EnforceCascadePolicy` field on `DeleteIssuesParams` (the spec was already in the field's doc comment) and switches proxied delete to classic parity:

- **default** — refuse when an issue has dependents outside the deletion set, with a typed `DeleteBlockedError` naming the blockers and how to proceed
- **`--force`** — delete only the named IDs, orphan external dependents (reported in output), clean up dependency links touching deleted rows
- **`--cascade [--force]`** — transitive sweep, as before

The legacy `EnforceCascadePolicy=false` path is byte-for-byte preserved for the wisp/mol/gc/purge proxied callers and pinned by a dedicated test.

## Acceptance

The #5363 characterization subtest (`TestProxiedServerCascadeParity/gate_delivery_ledger`) is deliberately flipped: `wantProxied` now equals `wantClassic` (gate gone, target + sibling survive), and any cross-mode divergence is an error rather than a logged known-divergence.

## Testing

- Domain: 4 new suite cases covering all `EnforceCascadePolicy` × `Cascade` × `Force` combos, including the legacy-path pin (`./internal/storage/domain/db` green)
- Integration (proxied server genuinely running, `BEADS_TEST_PROXIED_SERVER=1`): refusal contract, new `TestProxiedServerDeleteForceOrphans` with row-level orphan + dep-link assertions, JSON output contracts (adds `cascade`/`would_orphan`/`orphaned_issues`, removes nothing)
- Full `./internal/storage/domain/... ./cmd/bd/` run: all failures are the known bd-vb2u0 ambient cluster, verified by FAIL-set diff against unpatched main (identical fails on 80af94c0a; the fifth passes solo on this branch)

Closes bd-paurh.

https://claude.ai/code/session_01FZXjihTayFcngYzo6xPd3p

### DIFF
--- a/cmd/bd/cascade_parity_proxied_integration_test.go
+++ b/cmd/bd/cascade_parity_proxied_integration_test.go
@@ -156,11 +156,11 @@ func runPruneSweep(t *testing.T, r cascadeRunner) pruneSweepOutcome {
 }
 
 // TestProxiedServerCascadeParity is the delete/prune/purge half of bd-04vav:
-// proxied `bd delete` ALWAYS cascades (--cascade is refused outright), so the
-// question the wheelhouse needs answered is not "is there a flag" but "does the
-// cascade reach anything classic mode would have left alone" — on the two
-// shapes it actually runs: wh-gate-sweep's gate purge, and the wisp-decay
-// close+purge pair.
+// since bd-paurh, proxied `bd delete` honors the classic cascade policy
+// (default refuse, --cascade sweep, --force orphan), so the question the
+// wheelhouse needs answered is "does a delete reach anything classic mode
+// would have left alone" — on the two shapes it actually runs:
+// wh-gate-sweep's gate purge, and the wisp-decay close+purge pair.
 //
 // Both scenarios are replayed against a classic embedded workspace and a
 // proxied one and the outcomes compared, so a divergence is reported as a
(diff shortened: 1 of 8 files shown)


steveyegge:
LION REVIEW VERDICT — PR #5371 (fix/proxied-cascade-policy-bd-paurh): APPROVE. Merge-ready, with one output-contract finding I recommend fixing pre-merge (one line) and one knowingly-owned divergence to document. Neither blocks if you rule otherwise — owner's call.

1. THE FOCUS QUESTION — CONFIRMED. Refuse/orphan scope of DIRECT external dependents per deletion-set member matches embedded exactly. Oracle: internal/storage/issueops/delete.go:101-146 — classic's !cascade&&!force path computes direct dependents from BOTH dep tables (dependencies + wisp_dependencies, table-not-exist tolerated), excludes in-set ids, refuses on the first blocked id with the byte-identical message ("has dependents not in deletion set; use --cascade to delete them or --force to orphan them"), and sets result.OrphanedIssues to that id's blockers; the force path orphans direct externals only, transitive dependents untouched. Your externalDependents mirrors all of it. The sorting you added (embedded reports append-order) is a deterministic improvement, not a divergence that matters.

2. FINDING — double JSON emission on a refused preview (recommend the one-line fix pre-merge). In --json mode, a blocked preview prints the payload JSON (carrying the "error" field, per your output test) and THEN runDeleteProxiedPreviewTx calls HandleErrorRespectJSON, which in jsonOutput mode calls jsonStdoutError (cmd/bd/errors.go:91-94) — a SECOND JSON document on stdout. Classic's refusal emits exactly one. Any strict single-document parser (json.loads, an exact-compare harness) breaks on the concatenation, and it is a parity wrinkle inside the parity PR. Fix: when jsonOutput and result.blocked != nil, return &exitError{Code: 1} directly instead of HandleErrorRespectJSON — the payload already carries the error; prose mode stays as-is. Your existing output tests don't catch it because they call outputDeleteProxiedPreview directly, below the double-emit.

3. DIVERGENCE — wisp-as-named-target, strictly-safer direction, own it in the doc comment or a follow-up bead. Embedded partitions wisp IDs out BEFORE the dependent check (regularIDs only reach the refusal loop), so classic never refuses on — and never orphan-reports for — a wisp in the deletion set. Your externalDependents runs over all params.IDs, so bare `bd delete <wisp-id>` with an external dependent now refuses on proxied where classic proceeds. I would keep your behavior (embedded's wisp skip looks like the bug), and legacy wisp/mol/gc/purge callers are unaffected (Enforce=false), but the parity oracle has no wisp-as-named-target shape, so this should be a sentence on EnforceCascadePolicy's doc comment or a filed bd bead — someone will hit it and reach for the parity suite that doesn't cover it.

4. VERIFIED ALONG THE WAY. Force:true hardcode in the execute path is correct-by-routing (delete_proxied_server.go: dryRun||!force diverts to preview first). Blocked preview under DryRun performs no writes before refusing. result.OrphanedIssues doubling as blockers-on-refusal is unusual but pinned deliberately by iucCascadePolicyRefuses — fine. The legacy pin subtest (Cascade alone decides, Force ignored) is exactly the fence the wisp/mol/gc/purge callers need. The characterization-to-hard-parity flip, including promoting the cross-mode Logf to Errorf, is the right acceptance shape. CI fully green across all shards; your bd-vb2u0 FAIL-set-diff methodology for the full-run noise is sound.

Merge on your posted judgment for #2 — fix-now or file-follow-up both acceptable to this reviewer; #3 needs only the sentence. Good catch-chain from the #5363 oracle to here: this is the second real data-loss class the parity suites have stopped pre-cutover. — lion

steveyegge:
Pushed cc3abac58 addressing both review items: (1) blocked+--json now returns exitError directly — the preview payload's `error` key is the one JSON doc on stdout; TestProxiedServerDeleteBlockedJSONSingleDoc pins the single-doc contract (passes with BEADS_TEST_PROXIED_SERVER=1); (2) the deliberate wisp-as-named-target divergence from embedded is now documented on EnforceCascadePolicy. Merging on lion's posted APPROVE once checks are green.

PR #5374, opened 2026-08-06
Put sixteen more bd commands behind issueops roles, conformance-tested at three backends

julianknutsen:
Puts sixteen more `bd` commands behind role interfaces in the leaf `issueops`
package, gives each one a semantic-conformance contract run at all three
backends, and points both CLI routes and the HTTP handler at the same role.
One command per commit.

This is the end of the façade programme. `/data/tmp/sd-server/FACADE-SUMMARY.md`
is the long-form map; this is the reviewable version.

## The shape every slice built to

1. a role interface in the leaf package `issueops` (imports `internal/types` and
   stdlib only, exports no constructors),
2. a semantic-conformance contract run at `internal/storage/dolt`,
   `internal/storage/embeddeddolt` and `internal/storage/uow`,
3. both CLI routes — direct and `--proxied-server` — calling the role,
4. where the capability belongs on the wire, an operation on
   `internal/httpapi/spec/openapi.v0.yaml`, spec first,
5. an HTTP handler calling the SAME role.

A capability is reached through an accessor on the substrate (`store.Deleter()`),
never a constructor, because the accessor is where the hook and telemetry
decorators layer. **A new capability got a new role and a new accessor. No
method was ever appended to an existing role** — that rule is why
`storage.Storage` reached 139 methods and why this did not repeat it.

## What this found

Eight live defects, every one of them invisible to the pre-existing audit suite.
That is the headline, and the reason is structural: the old suite's factory type
is `func(t *testing.T) storage.DoltStorage`, so the unit-of-work backend — the
one behind `--proxied-server` — could never run it at all.

| Defect | Fix |
|---|---|
| proxied `bd create` silently upserted over an existing bead | `521a783af` |
| proxied `bd update --ephemeral` left durable rows with `ephemeral=1`, replicated to every clone | `17f423526` (forward-only; repair is `bd-kt5yp`) |
| `bd config set` on the uow route did not re-sync custom tables | `270515915` |
| `bd config set-many` re-prefixed the workspace | `270515915` |
| `bd query` dropped OR-matches | `671c7801d` |
| `bd dep cycles` was nondeterministic | `bbb44b399` |
| proxied `bd delete` force-cascaded without being asked | `cf0a42cda` |
| delete cascade rewrote **surviving** rows' text to say a deleted issue was deleted | `903cbe031` |

The last one is worth reading on its own. Two call sites passed different
argument sets to the same closure — one all ids, one with wisps excluded — and
the divergence corrupted live rows. The Deleter suite was 59 cases green on
three legs while that bug was live, because no case seeded a cross-plane
dependent edge. **Green is not coverage**, and that is the most useful thing
this programme learned.

## What is NOT behind the surface, and why

Each of these is a decision, not an omission. Reasons in full in the summary doc.

- **`bd show`, `bd close`** — already converged before wave 1. A second role for
  a capability an existing role expresses is the same drift as a second method
  on the wrong one.
- **`bd version`** — a build string has no substrate. The one part that touches
  state got `VersionReconciler` in wave 3; the command did not.
- **Most of `bd init`** — every role is `store.X()`, and `bd init` is what makes
  the store exist. Wave 4 carved out the one slice reachable through a store:
  workspace identity, split into `Bootstrapper` (guarded write) and
  `InitVerifier` (read). The split is load-bearing — bd reads this identity on
  paths where it is forbidden to write one.
- **`bd create --graph`** — deferred on the record (AMBIGUITIES Q10).
  `BatchCreator` is create-only with server-assigned ids; `--graph` is an upsert
  with caller-supplied symbolic names. Folding them means one role with two
  incompatible id policies. Largest piece of unconverged surface left.
- **Writes on the HTTP surface** — `bd serve` publishes four
  (`claim`, `batchCreate`, `sweep`, `delete`) and no lifecycle writes. Each new
  write is a separate decision about an unauthenticated loopback surface. A role
  with no operation is one whose operation has not been argued for yet.

## How much three green legs actually prove

Stated plainly because it is easy to over-read. For most roles `dolt` and
`embeddeddolt` share one validate/execute body and differ only in how they reach
a transaction; `Reader` shares the entire implementation. Only `uow` is
genuinely separate code. **A three-legged green run is usually two independent
votes, not three** — and for `TreeWalker` it is one, since all three legs call
the same `WalkDependencyTreeInTx`. The slice said so rather than papering over
it.

Every case is written against what the LEAF DOC promises, not against what an
implementation does. Wave 4 added 59 cases across four roles and parked none:
zero `skipKnownDivergence`, zero `SPEC-GAP` markers.

## The merge

`59a1df98b` merges `origin/main`. Four of the five conflicts were textual; the
fifth was semantic and is the reason this is a merge rather than a rebase.

main grew a second `bd serve` arm that serves a registered backend from the
store the root command already opened, wiring `httpapi.Config` from two role
accessors. This series grew that same Config from two roles to thirteen and made
`Listen` refuse a partial set. Both sides are individually correct and merge
cleanly **as text**; together they bind nothing — every registered-backend serve
fails at startup with "no database source". main's own two serve tests caught
it. `serveIssueRoles` now returns the whole set as a struct, so a role added to
Config without being added there is a compile error rather than a startup
failure.

`.golangci.yml` was resolved as set arithmetic against the merge base and
verified programmatically: every addition from either side survives, every
removal from either side survives.

## Verification

Local, on the merged head:

- `go build ./...`, `go vet ./...`, `gofmt` — clean
- `golangci-lint run ./...` — 0 issues
- `make api-check` (spec regen + `internal/httpapi`) — green
- conformance contracts at all three backends — green
- `cmd/bd/protocol` golden corpus — green
- `cmd/bd` serve suite — green

## Follow-ups filed

`bd-kt5yp` (P1, repair `ephemeral=1` rows), `bd-moe2k` (P1, repair `[deleted:]`
text corruption), `bd-5g4f2` (P1, `bd graph` still on the legacy cycle seam),
`bd-hysee` (P1, sweep batching), `bd-52lk3` (P1, query offset panic), plus
`bd-lidlu` and fourteen others.

Every reading this programme proceeded under rather than blocked on is in
`AMBIGUITIES.md`, with what changes if the other reading is right.


julianknutsen:
Superseded by a stacked split, for reviewability. Same tree, same tip commit — `facade/wave-4` is byte-identical to this branch's head.

- #5375 — wave 1 (reopen, create, count, update) → `main` — 42 files
- #5376 — wave 2 (config, status, dep cycles, dep list, ready + 5 HTTP ops) → wave 1 — 114 files
- #5377 — wave 3 (version reconciler, query, list caps, purge/prune, create --file + sweep) → wave 2 — 121 files
- #5378 — wave 4 (list, delete, init identity, dep tree) + the merge with main and the integration work → wave 3 — 248 files

Each wave that adds a role to `httpapi.Config` now also wires it into `serveIssueRoles`, so every branch in the stack builds and serves on its own rather than only at the tip.

Closing this in favour of the stack.

PR #5375, opened 2026-08-06
facade 1/4: bd reopen, create, count, update behind issueops roles

julianknutsen:
**Stacked 1 of 4** — the façade programme, split for review. Merges into `main`.

Puts four commands behind role interfaces in the leaf `issueops` package, each
with a semantic-conformance contract run at all three backends
(`internal/storage/{dolt,embeddeddolt,uow}`), and points both CLI routes at the
same role.

| command | role |
|---|---|
| `bd reopen` | `Lifecycle.Reopen` |
| `bd create <title>` (proxied) | `Lifecycle.Create` |
| `bd count` | `Counter` |
| `bd update` (proxied, plain) | `Lifecycle.Update` |

**Two live defects fixed**, both invisible to the pre-existing audit suite
because its factory type is `func(t *testing.T) storage.DoltStorage` — so the
unit-of-work backend, the one behind `--proxied-server`, could never run it.

- **proxied `bd create` silently upserted** over an occupied `--id` instead of
  refusing. Now create-only by the role's contract; two tests pin the refusal.
- **proxied `bd update --ephemeral` left durable rows with `ephemeral=1`** —
  versioned, replicated to every clone, invisible to wisp-plane reads and
  skipped by JSONL export. Both routes now perform the atomic plane move.
  Forward-only: nothing repairs rows the old code already produced (`bd-kt5yp`).

**The governing rule:** a new capability gets a NEW ROLE and a NEW ACCESSOR on
the substrate (`store.Counter()`), never a constructor and never an appended
method on an existing role. That rule is why `storage.Storage` reached 139
methods and why this does not repeat it.


julianknutsen:
## Red-team council — PR 5375 (wave-1)

**Confirmed: 2 major, 3 minor. Refuted and dropped: 5.**

### MAJOR: bd update --persistent falsely reports success on the corrupted rows this fix leaves behind

`internal/storage/issueops/persistence.go:35`

The same-plane branch of MoveIssuePersistenceInTx hardcodes `UPDATE wisps SET ephemeral = ...` even
when sourceWisp == targetWisp == false (an issues-plane row). For a healthy durable row that branch
is unreachable (flags already match the plane), but the exact population this PR's own changelog
certifies exists — durable rows in `issues` carrying `ephemeral = 1`, manufactured by the old
proxied column-flip bug (CHANGELOG.md:793 at facade/wave-1: "still in `issues` with the flag set") —
hits it: NormalizePersistenceMode(persistent) yields ephemeral=false, sourceWisp=targetWisp=false,
flags differ, so the UPDATE runs against the wisps table, matches zero rows, and returns
Changed=true. Empirically reproduced on the wave-1 binary: seed an issue, `dolt sql -q "UPDATE
issues SET ephemeral=1 ..."` (simulating the pre-fix proxied bug), then `bd update <id>
--persistent` prints "✓ Updated issue" and exits 0 while issues.ephemeral stays 1 and wisps stays
empty. `--ephemeral` on the same row DOES repair it (real cross-plane move, verified). The wave-1
relevance: before this PR the proxied rou

**Failure scenario.** A workspace holds rp-yrw in `issues` with ephemeral=1 (written by a pre-fix proxied `bd update
--ephemeral`). Operator runs `bd update rp-yrw --persistent` on either route after this PR: stdout
says "✓ Updated issue: rp-yrw", exit 0, but the row still has ephemeral=1 in `issues` — still
skipped by JSONL export and invisible to wisp-plane reads, with no signal anything failed.

### MAJOR: Proxied bd create drops the YAML prefix overlay the direct route still enforces — new route divergence, justified by a false changelog claim

`cmd/bd/create_proxied_server.go:115`

The old proxied create validated an explicit --id against `overlayYAMLPrefix(cctx.IssuePrefix)` —
the workspace config.yaml `issue-prefix` winning over the server DB prefix, exactly what the direct
route still does CLI-side at cmd/bd/create.go:509-514 (`loadEmbeddedIDPrefixes` ->
`selectCreateIDPrefix(globalFlag, overlayYAMLPrefix(""), storePrefix)`, YAML wins unless --global,
per GH#4957). This PR deletes that check and relies solely on the role's guard, which validates
against `createContext.IssuePrefix` read from the server database
(internal/storage/uow/issue_operations.go:74-79 -> PreparePublicCreateRequest) with no YAML overlay.
CHANGELOG.md:800-805 presents this as convergence on the claim "The direct route has never had
either fallback" — that is false for the prefix half: cmd/bd/create.go:513 calls
validation.ValidateIDPrefixAllowed with the YAML-overlaid prefix today. The two front doors of the
same command now disagree, which is the precise drift the series' governing rule exists to
eliminate, and the ratified record of the change rests on an incorrect statement about the

**Failure scenario.** Workspace config.yaml sets `issue-prefix: app`; the shared server database's `issue_prefix` config
is `bd`; `allowed_prefixes` empty. `bd create --id bd-123 --title T`: direct route refuses at
cmd/bd/create.go:513 (prefix must be `app`); proxied route (this PR) passes the role's DB-prefix
guard and creates bd-123, exit 0. Before this PR both routes refused. Same command, opposite
verdicts, and the proxied route now mints ids the workspace's own configuration forbids.

### MINOR: Proxied update's on_close hook decision reads a pre-transaction snapshot and can double-fire

`cmd/bd/update_proxied_server.go:190`

fireProxiedUpdateHooks decides whether the update "closed an open issue" (and so fires the on_close
hook) by comparing `before.Status` from proxiedUpdateTarget's pre-read — a separate read one
transaction earlier than the role's mutation — against the post-state. The old code read `current`
inside the same attempt/unit-of-work as the write and re-read it on every conflict retry, so
before/after were one snapshot. Now, if the role's conflict retry resolves against a concurrent
writer's commit, `before` is not refreshed. The file's comments acknowledge the pre-read guards are
advisory, but the hook is a side effect, not advice: it observably fires for a transition this
update did not perform. Same pattern exists for the notes-overwrite warning (line ~127), which is
cosmetic.

**Failure scenario.** Client A runs proxied `bd update X --notes n` while client B concurrently runs `bd close X` through
a path that fires on_close. A's pre-read sees X open; B's close commits first; A's role update lands
on the closed row without touching status. after.Status=closed and A's stale before.Status=open, so
A fires the on_close hook a second time for one close — a close-notification hook (e.g. posting to
chat) runs twice.

### MINOR: "Two independent votes" is one vote on the execution half: all three Counter legs run the same internal/storage/issueops count body

`backend/conformance/counter_contract.go:20`

The header (lines 18-25) says dolt+embeddeddolt share internal/workapi/storecounter ("one vote plus
an engine check") and "the unit-of-work provider is the second" independent vote. Below
workapi.BuildCountFilter that is not true: DoltStore.CountIssues/CountIssuesByGroup
(internal/storage/dolt/counts.go:30,46), EmbeddedDoltStore
(internal/storage/embeddeddolt/counts.go:32,48) AND the uow leg's repository
(internal/storage/domain/db/issue.go:1074,1083 via issueUseCaseImpl at domain/issue.go:1758-1772)
all delegate to issueops.CountIssuesInTx / CountIssuesByGroupInTx. The only genuinely independent
code between the "two votes" is config loading (LoadStoreListConfig vs LoadUOWListConfig) and
transaction acquisition. Every execution-half assertion in the contract — table choice, wisp merge,
bucket naming, priority CAST — is a SINGLE body voted three times, so a defect there can never
surface as leg-vs-leg disagreement. The series' governing brief warns three legs are usually two
votes; for Counter it is one, and the file affirmatively claims otherwise.

**Failure scenario.** A bug is introduced into CountIssuesByGroupInTx (e.g. the wisps-merge loop or the priority-bucket
CAST normalization). All three contract wirings go green in lockstep because they execute the
identical function; a reviewer trusting the header's "second independent vote" reads the triple-
green as cross-validation that never happened.

### MINOR: SPEC-GAP bd-yby99.32 parks a user-visible provenance regression: proxied bd create's dolt commit message drops the issue id and the bd: prefix

`cmd/bd/create_proxied_server.go:111`

Before this PR the proxied route committed "bd: create <id>". CreateRequest is the only Lifecycle
request without a Provenance field (ReopenRequest and UpdateRequest both carry one precisely so the
two routes write identical entries), so the proxied route now inherits the uow implementation's
default, which is the bare "create issue" (internal/storage/uow/issue_operations.go:109) — no id, no
"bd: " prefix. One command now has three history spellings: direct embedded still writes "bd: create
<id>" (cmd/bd/create.go:617), the direct server-backed role writes "bd: create issue"
(internal/storage/dolt/issue_operations.go:39), and proxied writes "create issue". The marker is
honest about the mechanism and internal/storage/issueops/execution.go:19-28 tracks the decision, but
the practical effect ships in this PR: the route the series is converging is the one whose `bd dolt
log` gets strictly worse, and the divergence between routes is the exact class of drift the role
move exists to kill. Verified no code classifies commits by the "bd: " prefix, so the blast radius
is human/script log read

**Failure scenario.** A team-server operator runs `bd dolt log` (or greps it) to find which commit created issue bd-xyz
after a bad create. Every proxied create in the window reads "create issue" with no id, so the entry
cannot be tied to the issue without diffing each commit; the same workflow on a local embedded clone
still works, so the operator's runbook silently stops matching the server.

---

### Refuted and dropped

- **Proxied bd reopen lost its nil-uowProvider guard; resolve now panics instead of returning the initialized-chec** — The mechanical claim is correct: main's runReopenProxiedServer opened with a uowProvider nil-guard, the wave-1 rewrite deleted it, and reopenProxiedResolve (cmd/bd/reopen_proxied_server.go:134) hands uowProvider to uow.RunTxRead, which calls p.NewUOW with no nil check (internal/storage/uow/tx.go:184) before the guarded proxiedIssueLifecycle() is reached. However, the nil state

- **Proxied bd update asks for the lifecycle role before resolving the id — the role-first ordering that turns a p** — The defect requires a uowProvider that offers the reader surface but not uow.IssueLifecycleSource. No such provider exists or can be constructed through any real caller: the sole production assignment (cmd/bd/main.go:1514) funnels through newProxiedServerUOWProvider → openAndInitSchema (internal/storage/uow/dolt_sql_provider.go:327), which always returns *doltSQLProvider — the

- **Counter conformance never exercises CountByGroup with the wisp tier, the ephemeral-tier routing, or template e** — The finding correctly inventories the new conformance suite (no CountByGroup+IncludeInfra case, no template seed, no infra-type routing case), but its severity rests on the claim that the grouped wisps merge and ephemeral routing "have zero cases at any backend" and that the named regressions pass everything green. That is demonstrably false. All three legs (dolt counts.go:46,

- **Occupied-ID create case covers 3 of 4 plane quadrants: an ephemeral create over an occupied WISP ID is never a** — The quadrant count is factually right (durable/durable, durable-over-wisp, ephemeral-over-durable are asserted; ephemeral-over-wisp is not), but the finding's mechanism and failure scenario are wrong. The wisp-over-wisp refusal does not rest on the strict wisp-table insert: on both independent bodies, every CreateOnly create first runs the shared plane-agnostic guard EnsureIssu

- **Both bd reopen routes now depend on wisp-plane id resolution that Lifecycle.Reopen does not promise and no cas** — The finding's load-bearing premises fail verification. (1) "This PR makes both routes depend on unspecified plane resolution" is half wrong: on origin/main the direct route already called ops.Reopen (the role) after resolveAndGetIssueForMutation — that dependence predates the PR; only the proxied route newly delegates (it previously called ReopenWisp explicitly). (2) "No tripwi

julianknutsen:
## Red-team fixes — both majors addressed

### `bd update --persistent` reported success and repaired nothing (`a18a8aa5f`)

`MoveIssuePersistenceInTx`'s same-plane branch — no move needed, but the stored
flags disagree with the row — issued `UPDATE wisps` unconditionally. For a
healthy durable row that branch is unreachable, so nothing noticed.

It is reachable for the row shape **this PR's own changelog certifies exists**:
a durable row in `issues` carrying `ephemeral = 1`, left behind by the pre-role
proxied update. `bd update <id> --persistent` is what an operator reaches for to
repair that, and it sent the UPDATE at the wrong table, matched nothing, and
returned `Changed: true`.

Wave 1 is where it starts to matter: before it, the proxied route's column flip
cleared the flag in place, so `--persistent` happened to work. Routing both
routes onto `Lifecycle.Update` removed that accidental repair path and left a
silent false success in its place.

The table is now the row's own plane, and a zero-row update is an error rather
than a success — `Changed: true` is a claim about the database, so it is checked
rather than asserted. Verified by mutation: against the original body the new
test fails with `ChangedTables{"wisps":true}` and `issues.ephemeral is still 1`.

### Proxied `bd create` minted ids the workspace forbids (`e7a8fe284`)

config.yaml's `issue-prefix` wins over the database's except under `--global`
(GH#4957), and only a front door can read config.yaml. Routing this route onto
`Lifecycle.Create` dropped the overlay, so with config.yaml `app` against a
server whose prefix is `bd`, `bd create --id bd-123` was **refused by the direct
route and created by the proxied one**.

`CreateRequest` gains `IDPrefix` — the same shape as `DeleteRequest.IDs` being
exact, because resolution is the front door's job and the role enforces what it
is handed. **Both** routes now pass it, which is more than restoring parity: the
direct route validated against the overlay in `cmd/bd` and then let the role
check the store's prefix again, so an embedded workspace whose config.yaml
disagreed with its own database refused every explicit id.

The changelog claim that "the direct route has never had either fallback" was
false for the prefix half and is corrected.

---

Three minors from this PR's council remain open and are filed rather than fixed
here: the proxied `on_close` hook can double-fire off a stale pre-read, the
Counter contract's "two independent votes" is one vote on the execution half,
and SPEC-GAP bd-yby99.32 parks the create commit-message change.

PR #5376, opened 2026-08-06
facade 2/4: config, status, dep cycles, dep list, ready behind roles + 5 HTTP operations

julianknutsen:
**Stacked 2 of 4** — merges into `facade/wave-1`. Review [#1](../../pull/5374) first.

Five more commands behind roles, and the first five HTTP operations.

| command | role | HTTP |
|---|---|---|
| `bd config get/list/set` | `WorkspaceConfig` | `GET /v0/beads/config`, `/config/{key}` |
| `bd status` | `StatsReporter` | `GET /v0/beads/stats` |
| `bd dep cycles` | `CycleDetector` | `GET /v0/beads/dependencies/cycles` |
| `bd dep list` (multi-id) | `EdgeReader` | `GET /v0/beads/dependencies` |
| `bd ready` total | `ReadyCounter` | `GET /v0/beads/ready:count` |

**Three live defects fixed:**

- **`bd config set` on the uow route did not re-sync the custom tables** that
  reads consult first, so `bd config set types.custom` reported success and
  never took effect.
- **`bd config set-many` re-prefixed the workspace.**
- **`bd dep cycles` was nondeterministic** — the walk iterated a Go map, so a
  caller diffing two `--json` snapshots saw changes that were not changes.

Every operation is written into `internal/httpapi/spec/openapi.v0.yaml` before
its handler exists, and answers from the SAME role its `bd` command reaches.

Also wires wave 2's roles into `bd serve`'s registered-backend arm: `Listen`
refuses a partial role set, so a wave that adds a role to `httpapi.Config` must
add it to `serveIssueRoles` in the same change.


julianknutsen:
## Red-team council — PR 5376 (wave-2)

**Confirmed: 4 major, 5 minor. Refuted and dropped: 1.**

### MAJOR: GET /v0/beads/config leaks credential values stored under the 'apikey' (and pwd/passwd/pat/bearer/auth/.key) spellings — **FIXED in `d122d0491`**

`internal/config/yaml_config.go:122`

The HTTP config surface has no authentication; redaction (settings.go wireSetting ->
config.IsSecretKey) is the ONLY thing keeping stored credentials off the wire. IsSecretKey does a
substring match against exactly {"api_key","api-key","secret","token","password"}. The
apigen.Setting schema doc (internal/httpapi/apigen/types.gen.go:230) affirmatively claims redacted
is set when the key name contains "token", "secret", "password, or an API-key spelling" — but the
concatenated spelling "apikey" (no separator) is not matched, nor are "pwd", "passwd", "pat",
"bearer", "auth", or a bare ".key" suffix. These keys are also NOT yaml-only (IsYamlOnlyKey returns
false), so a front door writes them into the DB config plane that WorkspaceConfig.ListSettings
reads. I confirmed the leak end-to-end: a server backed by a settings role holding
{"integrations.apikey":"sk-live-SECRET-VALUE","integrations.api_key":"..."} answers GET
/v0/beads/config with
items=[{key:integrations.api_key,redacted:true},{key:integrations.apikey,redacted:false,value:"sk-
live-SECRET-VALUE"}]. The same server's /v0/beads/con

**Failure scenario.** Operator runs `bd config set integrations.apikey "sk-live-..."` (a spelling the schema doc says is
covered), then `bd serve` binds. Any network peer that can reach the port issues GET
/v0/beads/config and receives the API key in cleartext with redacted:false, while the identical key
spelled integrations.api_key would have been withheld.

### MAJOR: bd dep list batch gate keys on resolved count, so a batch with one resolvable anchor silently emits the wrong JSON schema

`cmd/bd/dep.go:885`

At facade/wave-2, the direct route gates the EdgeReader records path with `if len(resolved) > 1 &&
direction == "down"` (cmd/bd/dep.go:885) while batch membership and the ghost-anchor skip are
decided by `batchMode := len(args) > 1` (line 833). When a batch invocation has exactly one anchor
that resolves, the unresolvable ones are warned-and-skipped, `resolved` shrinks to 1, the gate goes
false, and the command falls through to the Relations neighbor path, whose `--json` output is
`[]*issueops.RelatedIssue` (hydrated issues, line 920-925) instead of the documented flat
`[]*types.Dependency` records that printDepListEdges emits. This is exactly the fall-through the
PR's own comment in readDepListEdges claims is gone ("Both fall-throughs are gone ... silently
emitted an array of ISSUES where it documents an array of dependency records"), and it also diverges
from the proxied route, which gates on `len(args) > 1` (runDepListRecordsProxiedServer) and
correctly answers with records plus a per-anchor Missing warning — the route-parity this series
exists to enforce. The gate was changed to

**Failure scenario.** Direct (non-proxied) workspace: `bd dep list bd-real bd-typo --json` where bd-typo resolves to
nothing. Wave-2 prints the skip warning to stderr, then emits a JSON array of RelatedIssue objects
(id/title/priority/status/dependency_type of bd-real's neighbors) instead of the documented flat
array of dependency records (issue_id/depends_on_id/type). A script parsing the documented shape
reads garbage or crashes; the same invocation under --proxied-server returns the correct records
shape, so the two routes disagree on schema for identical input.

### MAJOR: GET /v0/beads/stats with whitespace-only assignee returns 500 internal instead of 400

`internal/httpapi/stats.go:36`

The handler refuses only the exact-empty case (`assignee == "" && Has("assignee")`); q.str does not
trim. A whitespace-only assignee (e.g. `?assignee=%20%20`) passes the guard and reaches
StatsReporter.AssigneeStats, which refuses it with issueops.ErrValidation
(workapi.ValidateStatsAssignee trims). The error then goes through s.failErr -> ClassifyError
(internal/httpapi/problem.go:366), which has NO ErrValidation row — the edges handler classifies the
sentinel itself (failEdgeReadErr) and the ready handlers use invalidFilterParam prose-matching, but
stats has neither — so it falls to the default 500 CodeInternal. Because status >= 500, failErr also
emits a `request_error` operator event for a pure caller-input mistake. This contradicts the
OpGetStats vocabulary comment in problem.go ("the EMPTY assignee the document refuses") and the
handler's own doc ("a 400 that names the parameter is what a client can act on"). stats_test.go
covers only `assignee=` (line 48), never whitespace.

**Failure scenario.** Client sends GET /v0/beads/stats?assignee=%20 (trailing-space bug in a script). Server responds 500
{code: "internal"} instead of 400 invalid_argument naming `assignee`, and the operator's
request_error alert counter increments on every such request.

### MAJOR: BlockedIssues 'excluding closed and pinned ones' misstates shipped behavior; exclusion clause has zero contract cases

`issueops/statsreporter.go:99`

The doc bullet directly above (line 97-98) teaches that in this struct "pinned" means the FLAG
("PinnedIssues counts the pinned FLAG and overlaps every status bucket"). Then line 99-102 says
BlockedIssues counts is_blocked rows "excluding closed and pinned ones". All three implementations
(internal/storage/dolt/queries.go:125, internal/storage/embeddeddolt/statistics.go:23,
internal/storage/domain/db/issue.go:1059) exclude only `status <> 'closed' AND status <> 'pinned'` —
the STATUS, not the flag — so a flag-pinned, status-open row with an open blocker IS counted in
BlockedIssues (and depresses ReadyIssues), which the doc's natural reading says cannot happen.
backend/conformance/stats_reporter_contract.go has no case in this quadrant at all: no blocked-then-
closed row, no blocked status-pinned row, no blocked flag-pinned row — the exclusion clause of the
blocked count is entirely unasserted on every backend. This is exactly the untested-quadrant shape
the Deleter lesson paid for: a backend that implemented the doc's flag reading, or dropped the
exclusion entirely for open-status row

**Failure scenario.** Seed an open blocker plus an open, Pinned=true dependent (a state the by-status contract case itself
proves constructible): all three backends report it in BlockedIssues while a reader of the leaf doc
— or a fourth backend written to it — excludes it; the conformance suite cannot tell the two apart,
and `bd status`'s headline ReadyIssues differs by one per such row with nothing pinning which answer
is right.

### MINOR: Whitespace-only assignee on GET /v0/beads/stats returns 500 internal instead of the documented 400 invalid_argument

`internal/httpapi/stats.go:36`

handleStats only pre-screens the empty case: `if assignee == "" && r.URL.Query().Has("assignee")`. A
whitespace-only value (?assignee=%20) is non-empty, passes that guard, and is forwarded to
reporter.AssigneeStats. Both shipped implementations (workapi/storestats/reporter.go:77 and
storage/uow/stats_reporter.go:72) validate with workapi.ValidateStatsAssignee, which does
strings.TrimSpace and returns issueops.ErrValidation. The handler routes role errors through
s.failErr -> ClassifyError (problem.go), which has NO case for issueops.ErrValidation — unlike
edges.go's failEdgeReadErr, which explicitly maps it to a 400. So it falls through to the default
500 CodeInternal. The spec (openapi.v0.yaml getStats assignee param) says 'An empty value is a 400';
a whitespace-only value is morally identical and is deterministically refused by the role, but
reaches the client as an un-actionable server fault. It also books a spurious request_error in the
>=500 log signal operators alert on, and on the provider path opens a unit of work / DB slot before
failing. Confirmed by test: ?assignee=%20 yie

**Failure scenario.** Client calls GET /v0/beads/stats?assignee=%20 (or any all-whitespace actor). Documented/expected:
400 invalid_argument naming `assignee`. Actual: 500 {code:internal,detail:"internal server error"}
with no param, indistinguishable from a real backend fault, and it increments the server-error
metric.

### MINOR: internal/httpapi/server.go ships garbled merge-conflict comment debris in three doc blocks

`internal/httpapi/server.go:126`

The wave-2 reconcile of the five slices left stacked, contradictory sentence fragments from each
slice's version of the same doc comments. Config field doc (lines ~123-151): "Reader, Claimer,
Settings and Stats are the roles..." immediately followed by two truncated mid-sentence variants
("Reader, Claimer and EdgeReader are the issue roles this server answers" / "Reader, Claimer and
ReadyCounter are the issue roles this server answers"), then three stacked variants "Set them
together" / "Set them all together" / "Set all three together", and duplicate "other combination,
including a ..." lines including the nonsense phrase "a partially-set set". Server struct doc
repeats the same block three times naming issueCycles then issueEdges (lines ~231-246).
checkDatabaseSource doc (lines ~405-425) carries three interleaved variants of the PARTIAL-set
paragraph. Comment-only, but it is shipped incoherent text on the load-bearing invariant of this
file (what a complete database source is), and it is still present at facade/wave-3; it is only
cleaned up in wave-4 — the cleanup should be in this

**Failure scenario.** A contributor adding the next role reads internal/httpapi/server.go:123-151 at this PR's head and
encounters three contradictory instructions and sentences that cut off mid-clause; godoc renders the
garbled text verbatim. No runtime effect.

### MINOR: CycleDetector contract header claims 'TWO INDEPENDENT BODIES' but all three wirings run one shared body — zero independent execution votes

`backend/conformance/cycle_detector_contract.go:22`

The header says the unit-of-work provider "reaches the same function through the domain repository
and is the second [independent body]" — self-contradictory, and the second half is what's true: dolt
and embeddeddolt wrap storage/issueops.DetectCycleReportInTx in their read transactions, and the uow
leg reaches the identical function via dependencySQLRepositoryImpl.DetectCycleReport
(internal/storage/domain/db/dependency.go:818, added in this PR). The graph read
(AppendBlockingGraphInTx), table choice, edge-type set, canonicalization and hydration are ONE body
on all three legs; only transaction acquisition differs. The header's own stated purpose for the DB-
backed cases — "which TABLES each seam reads, whether the two planes are merged, and which edge
types are followed" — is exactly what a single shared body makes uncheckable by cross-backend
agreement: a bug there (e.g. reading only the durable dependency table) goes green on all three
legs. Contrast the EdgeReader and ReadyCounter headers, which correctly count their genuinely
different uow bodies. A maintainer reading this heade

**Failure scenario.** AppendBlockingGraphInTx regresses to reading only the `dependencies` table (dropping
wisp_dependencies). RunCycleDetectorMergesTheDurableAndEphemeralPlanes fails — but only because that
one case exists; for any property the cases do NOT cover (e.g. a future edge-type addition mis-
mapped in the shared SQL), the "three wirings" agree by construction and the suite documents itself
as having an independent second vote it does not have.

### MINOR: Unreconciled merge slop: triplicated, contradictory doc comments and broken Config examples in httpapi server.go and problem.go

`internal/httpapi/server.go:123`

At facade/wave-2 the Config field doc (lines ~123-151) contains three interleaved variants of the
same paragraphs from the five parallel slices ("Reader, Claimer, Settings and Stats are the
roles..." / "Reader, Claimer and EdgeReader are..." / "Reader, Claimer and ReadyCounter are...";
"Set them together" / "Set them all together" / "Set all three together"; two copies of the "other
combination, including a PARTIAL set" sentence plus a third stranded mid-paragraph). The example
block stacks THREE contradictory Listen calls, each showing a PARTIAL Config (`{Reader, Claimer,
Settings, Stats}`, `{Reader, Claimer, EdgeReader}`, `{Reader, Claimer, ReadyCounter}`) that
checkDatabaseSource in the same file now REFUSES at startup — a reader copying any of them gets "no
database source: set Provider, or Reader, Claimer, Settings, Stats, CycleDetector, EdgeReader and
ReadyCounter together". The Server struct doc (~line 231) and checkDatabaseSource doc (~line 405)
carry the same three-way duplication, and internal/httpapi/problem.go has two more spots (line ~39:
"seven v0 operations actually ne

**Failure scenario.** A caller embedding httpapi follows the Config example `httpapi.Listen(httpapi.Config{Reader: rd,
Claimer: cl, EdgeReader: ed, ...})` verbatim and gets a startup refusal naming seven roles the
example told them were optional; every future reader of the file must guess which of three
contradictory paragraph variants is authoritative.

### MINOR: dolt GetStatistics mixes two snapshots (two read transactions) where embeddeddolt and uow use one — an unstated backend divergence behind the new role

`internal/storage/dolt/queries.go:114`

DoltStore.GetStatistics runs ScanIssueCountsInTx in one withReadTx and the blocked COUNT(*) in a
SECOND withReadTx, then computes ReadyIssues = OpenIssues - blocked across the two snapshots.
embeddeddolt (statistics.go, one withConn tx) and the uow leg (issueops.GetStatisticsInTx, one
runner) read both in one transaction. Pre-existing code, but this PR publishes it behind
issueops.StatsReporter whose leaf doc describes the counts as one coherent answer and whose
conformance suite asserts cross-backend agreement — under a concurrent writer only the server-backed
store can publish a summary corresponding to NO state of the workspace (e.g. BlockedIssues >
OpenIssues, ReadyIssues clamped to 0, or a ReadyIssues off by the writes that landed between the two
transactions). Neither issueops/statsreporter.go nor the contract header records the divergence,
unlike every other divergence this series is careful to name (SkipBlocked, the AssigneeStats
snapshot non-promise in uow/stats_reporter.go:61-64).

**Failure scenario.** On a busy server-backed workspace, `bd status` (or GET /v0/beads/stats) runs while another client
closes 10 open blocked issues between the two transactions: OpenIssues comes from the pre-close
snapshot, BlockedIssues from the post-close one, and the published ReadyIssues overstates ready work
by up to 10 — a number no single database state ever contained, on exactly one of the three
supposedly-equivalent backends.

---

### Refuted and dropped

- **AssigneeStats ReadyIssues ("the ready-work query") is never asserted non-zero; a permanently failing ready que** — The finding's in-suite facts check out: no conformance case asserts a non-zero assignee ReadyIssues (only a populated zero at stats_reporter_contract.go:378), both implementations swallow GetReadyWork errors to zero, and workapi/stats_test.go tests only the fold. But the failure scenario as filed does not survive. (1) The swallow-to-zero is explicitly documented contract behavi

julianknutsen:
## Red-team fixes — all four majors addressed

### Credential leak on the unauthenticated config surface (`d122d0491`)

`GET /v0/beads/config` publishes a stored value unless `config.IsSecretKey` says
otherwise, and `bd serve` has no authentication. That function was written as a
**lint** — a warning before `bd config set` writes a secret into a git-tracked
file — and this wave promoted it to a security control without widening it.

It matched five substrings, so `integrations.apikey` — the concatenated spelling
the **published schema affirmatively claims is covered** — was served in
cleartext beside a redacted `integrations.api_key`. Also missing: `pwd`,
`passwd`, `pat`, `auth`, `bearer`, `cert`, `credentials`.

The short spellings cannot be substrings (`pat` would redact `issue.path`,
`auth` would redact `commit.author`), so they are matched as whole **segments**
split on `.`/`_`/`-`. Fifteen cases pin it, including the five near-misses that
break if someone simplifies the segment matching back into `Contains`.

### `bd dep list` emitted the wrong JSON schema (`ed82acd5d`)

The direct route chose the edge-records path on `len(resolved)` — counted
**after** unresolvable anchors are warned-and-skipped — while batch membership
and the proxied route both key on the argument count. So
`bd dep list <good> <typo> --json` printed `RelatedIssue` objects on this route
and the documented flat array of dependency records on the other.

This is the third fall-through `readDepListEdges`' own comment claimed was gone.
The comment now counts to three and the gate reads `batchMode`.

### `GET /v0/beads/stats` answered 500 for a caller's typo (`ed82acd5d`)

The empty-assignee guard compared the **raw** string, so `?assignee=%20` slipped
past it into the role, came back `ErrValidation`, hit `ClassifyError` — which has
no row for that sentinel — and left as a 500 **with an operator alert attached**,
for a trailing space in somebody's script.

Empty now means trimmed-empty, which is what the role means by it, and
`failStatsErr` draws the `ErrValidation`-is-a-400 line that delete, tree, edges,
blocking and batch-create each draw locally. Four spellings pinned.

### `BlockedIssues` promised one thing and shipped another (`ed82acd5d`)

`statsreporter.go` said the count excludes "closed and pinned ones" one bullet
after teaching that `PinnedIssues` counts the pinned **flag**. All three
implementations exclude by **status**, so a flag-pinned open row with a blocker
IS counted — the opposite of the doc's natural reading.

The exclusion clause had **no case on any backend**: no closed-and-blocked row,
no status-pinned one, no flag-pinned one. Three legs green over a quadrant where
a fourth backend written to the doc would have disagreed with all of them. The
doc now says status, and a new contract case seeds all three shapes.

julianknutsen:
Landed on `main` via #5378 (merge commit `5e1c8697b`), which carried the whole stack — every commit on this branch is an ancestor of `main`, verified with `git merge-base --is-ancestor`.

Closing as superseded rather than merging: this PR's base is the wave beneath it, so merging it would not have reached `main`. The per-wave review and its council comment stay here for the record.

PR #5377, opened 2026-08-06
facade 3/4: version reconciler, query, list caps, purge/prune, create --file + sweep endpoint

julianknutsen:
**Stacked 3 of 4** — merges into `facade/wave-2`. Review #5375 and #5376 first.

Five commands, four more HTTP operations, including the first DESTRUCTIVE one.

| command | role | HTTP |
|---|---|---|
| startup version tracking | `VersionReconciler` | — |
| `bd query` | `Querier` | `GET /v0/beads/issues:query` |
| `bd list` row cap / counts | `Reader.List` request fields | — |
| `bd purge`, `bd prune` | `Sweeper` | `POST /v0/beads/issues:sweep` |
| `bd create --file` | `BatchCreator` | `POST /v0/beads/issues:batchCreate` |

**`bd query` was dropping matches.** Both front doors bounded a predicate
query's fetch at `max(3*Limit, 100)` rows and filtered what came back, so a
match beyond that window was absent from the page AND unreported by has-more —
an OR query over a workspace of any size returned an arbitrary prefix of its
answer and called it complete. The window is gone; the cost is stated rather
than hidden.

**`--pattern '['` reported "nothing to prune".** Both routes discarded
`filepath.Match`'s error, so an invalid glob was indistinguishable from a
correct pattern over an empty set — on a command whose whole job is to delete
what it matched.

**`POST /v0/beads/issues:sweep` is the first destructive operation on the wire.**
`bd serve` binds loopback and has NO AUTHENTICATION, so anyone who can reach
the port can clear closed beads. `protect_referenced` **defaults to `true`**
here where the CLI's `--ignore-references` is opt-out: a remote caller who
omits a member must not get a weaker guard than the operator typing the same
sweep locally. A durable sweep with neither a cutoff nor a pattern is refused
by the role itself, below every front door.

`BatchCreator` is its own role rather than a `Create` overload because it is
ALL OR NOTHING where `BatchCloser` keeps its survivors — a close records work
that is done, so a rollback destroys information; a create makes the work, so a
refused item leaves no survivors.


julianknutsen:
## Red-team council — PR 5377 (wave-3)

**Confirmed: 1 blocker, 2 major, 3 minor. Refuted and dropped: 2.**

### BLOCKER: uow reference scan never reads wisp comments, so proxied `bd prune` deletes a bead the protection contract says must be kept — **FIXED in `f03da662a`**

`internal/storage/uow/sweeper.go:149`

SweepRequest.ProtectReferenced promises the scan covers every not-done row's description, notes, AND
comments (issueops/sweeper.go:88-104), and says an implementation that cannot read the full set must
fail the sweep rather than under-scan. The store-backed body keeps that:
internal/storage/issueops/sweep.go:109 calls GetCommentsForIssuesInTx, which partitions ids between
`comments` and `wisp_comments` (internal/storage/issueops/bulk_ops.go:105-131), so an open WISP's
comments are scanned. The uow body's sweepReferencedInUOW instead calls
uw.CommentUseCase().GetCommentsForIssues(notDoneIDs), which routes to commentRepo.ListByIssueIDs
with UseWispsTable:false (internal/storage/domain/comment.go:69-71,
internal/storage/domain/db/comment.go:69-87) — it reads ONLY the durable `comments` table. The
notDone set DOES include wisps (SearchIssues merges both planes), so their descriptions/notes are
scanned but their comments are silently skipped. The conformance case RunSweeperProtectsCitedRows
(backend/conformance/sweeper_contract.go:278) cites only via a durable bead's DESCRIPTION — no comm

**Failure scenario.** Closed durable bead X. Open wisp W whose only citation of X is in a COMMENT on W (e.g. `bd comment W
"decision trail: see X"`). Run `bd prune --pattern 'X*' --force`. Direct route (dolt/embeddeddolt):
X is protected, Skipped.Referenced=1, X survives. Proxied route (`--proxied-server`, uow leg) with
the identical request (ProtectReferenced=true): W's comment is never read, X is NOT in the
referenced set, and X is permanently deleted — the same command deletes or keeps the bead depending
on which route answered.

### MAJOR: Store-backed Querier sorts the over-fetch probe row into the page, so the direct and proxied routes return different pages for --sort with --limit

`internal/workapi/storequerier/querier.go:56`

For a filter-expressible query, storequerier bumps the SQL limit to N+1 via WithFetchOneExtra (line
56) purely as a has-more probe, then FinishPage (line 68) runs SortRows over ALL N+1 rows BEFORE
cutting to N. The probe row therefore participates in the display sort and can displace a legitimate
row. The uow body (internal/storage/uow/querier.go) fetches/trims to exactly N rows natively and
sorts those N, matching the documented semantics (issueops/querier.go SortBy: "For a filter-
expressible query the database applied Limit, so the order sorts the page") and the old direct-route
behavior. bd list dodges this exact trap because BuildListFilter pushes SortBy into the SQL filter
(internal/workapi/list.go:201), making the N+1th fetched row last in sort order so the cut removes
precisely it; BuildQueryPlan never sets plan.Filter.SortBy, so for query the probe row lands
anywhere in the Go sort. Result: dolt and embeddeddolt (one body) disagree with uow (the other body)
about page CONTENT — the exact "two routes of one command came to disagree about --sort id --limit
5" failure this serie

**Failure scenario.** Seed 3 open bugs in creation order b1(priority 4), b2(priority 2), b3(priority 0). Run `bd query
"type=bug" --sort priority --limit 2`. Direct route (store body): SQL fetches 3 rows (2+1) in
storage order [b1,b2,b3], Go sort by priority gives [b3,b2,b1], cut to 2 -> page [b3,b2]. Proxied
route (uow body): SQL fetches 2 rows [b1,b2] (trimmed natively), sort -> page [b2,b1]. Same data,
same command, HasMore=true on both, but b3 appears on one route and not the other — and the store
page is not a sorted rendering of the page the database bounded, contradicting issueops/querier.go's
documented SortBy semantics.

### MAJOR: uow BatchCreator accepts the in-batch cross-plane edge that the contract and both store legs refuse

`internal/storage/uow/batch_creator.go:61`

issueops/batchcreator.go:28-31 states: an edge BETWEEN the durable and ephemeral buckets 'cannot be
written by the batch that creates both of its ends; a request that asks for one is refused with
everything else it asked for.' The shared store body enforces this: ExecuteCreateBatch assigns every
ID first, then CreateIssuesInTxWithResult runs filterCreateIssuesMixedBucketDependencies
(internal/storage/issueops/create.go:344-417), which hard-errors on a cross-bucket in-batch edge
(SkipDependencyValidationErrors is false in the batch role's options). The uow body creates item by
item through domain CreateIssue/CreateWisp; by the time the second item's edge is written its target
already EXISTS as a real row, and the domain layer happily writes cross-tier edges to existing rows
(proven by db/issue_cross_tier_test.go TestReverseCreateDependencyUsesActualSourceTier). So the
identical request — item0 durable with explicit id, item1 ephemeral (or a default infra type:
'agent'/'role'/'message', which ExecuteCreateBatch and createBatchItem both promote to wisps)
depending on item0 — is refused

**Failure scenario.** CreateBatchRequest{Items: [{Issue: durable, ID "bd-x"}, {Issue: ephemeral, Dependencies: [{TargetID:
"bd-x"}]}]}. Against DoltStore/EmbeddedDoltStore: error "mixed regular/wisp CreateIssues batch
cannot include cross-bucket dependency ... -> bd-x", nothing created. Against the uow provider: nil
error, both rows and the wisp->durable edge created. A library caller (or any future front door)
gets opposite outcomes for one request depending on backend.

### MINOR: Offset with explicit --limit 0 on the uow route renders LIMIT 18446744073709551615 OFFSET n and surfaces a recovered engine panic; the role validates the combination as legal

`internal/storage/uow/querier.go:57`

BuildQueryPlan accepts limit=0 (unlimited) together with a positive Offset (only offset<0 and
offset+SortBy are refused), and the uow body pushes the offset down for filter-expressible queries
(querier.go:57). limitOffsetSQL (internal/storage/domain/db/issue_search.go:511-522) renders `LIMIT
18446744073709551615 OFFSET n` for limit<=0 with offset>0, which the Dolt engine answers with a
recovered panic ("makeslice: cap out of range" out of topRowsIter) — the contract file itself
documents this and dodges it by only ever driving bounded limits
(backend/conformance/querier_contract.go:257-262 "EVERY REQUEST HERE CARRIES A BOUNDED LIMIT"). The
crash path predates this PR (the old proxied route could hit it too), but this PR is the one writing
the public contract that says uow HONORS Offset with 0 meaning unlimited (issueops/querier.go Offset
doc), while neither BuildQueryPlan nor the uow body guards the one combination known to fail. A one-
line refusal of offset>0 && limit==0 in BuildQueryPlan (or skipping pushdown when limit==0) would
close it.

**Failure scenario.** `bd query 'type=bug' --offset 1 --limit 0 --proxied-server` on any workspace: gatherQueryInput
passes (offset>=0, no sort), BuildQueryPlan passes (limit 0 legal, offset legal), plan is filter-
expressible, filter gets Limit=0 Offset=1, the rendered SQL trips the engine's makeslice panic, and
the user gets an internal panic-recovery error string instead of either rows or a typed
validation/unsupported refusal — for a flag combination both flags' help text presents as valid.

### MINOR: SweepResult.Dependencies undercounts cross-plane inbound edges on the uow leg — the two CLI routes print different numbers for the same sweep

`internal/storage/domain/issue_delete.go:74`

SweepResult.Dependencies is documented as 'edges pointing at or out of a swept row'
(issueops/sweeper.go:166-170). The store body delivers that: DeleteResolvedSetInTx counts outbound
rows per plane PLUS an inbound scan over BOTH dep tables for rows targeting the deleted set
(internal/storage/issueops/delete.go:196-258). The uow body's deleteMany computes DependenciesCount
as depRepo.CountAllForIDs(regularIDs, dependencies) + CountAllForIDs(wispIDs, wisp_dependencies) —
each predicate is `issue_id IN ids OR target IN ids` but only against ONE table per plane, so a
`dependencies` row targeting a swept WISP (depends_on_wisp_id) and a `wisp_dependencies` row
targeting a swept DURABLE id are never counted, even though the sweep does remove them
(DeleteAllForIDs at issue_delete.go:141 / FK cascade). No sweeper conformance case seeds any
dependency edge at all, so the counts are unpinned across legs.

**Failure scenario.** Open durable issue D with an edge D->W onto closed wisp W. `bd purge --force` direct route reports
'Dependencies removed: 1'; `bd purge --force --proxied-server` reports 'Dependencies removed: 0'.
Both delete the edge; only one reports it — the number the doc says 'explains a large storage
reclaim' is wrong on one route.

### MINOR: The store legs' cross-plane-edge refusal is untyped: it does not match ErrValidation as the role promises

`internal/storage/issueops/create_batch.go:163`

issueops.BatchCreator promises 'Deterministic request-validation failures match ErrValidation' and
enumerates the refusal vocabulary (issueops/batchcreator.go:145-160). The mixed regular/wisp refusal
raised by filterCreateIssuesMixedBucketDependencies (internal/storage/issueops/create.go:391) is a
bare fmt.Errorf; ExecuteCreateBatch routes it through ClassifyPublicCreateError
(create_batch.go:163), which has no branch for it (public_create.go:85-107) and returns it
unclassified. errors.Is(err, issueops.ErrValidation) is false, so a caller classifying refusals as
the contract instructs treats a deterministic request refusal as an infrastructure error — and
internal/httpapi/batch_create.go:323-338 (failBatchCreate) would map it to a 500 rather than a 400
on any surface that can spell the shape. Today HTTP items cannot name in-batch targets (no explicit-
id member), so the wire consequence is latent, but the typed-vocabulary promise is broken at the
role boundary now.

**Failure scenario.** A library caller does result, err := store.BatchCreator() ... CreateBatch(mixed-bucket request);
errors.Is(err, issueops.ErrValidation) == false and errors.Is(err, issueops.ErrAlreadyExists) ==
false, so the caller's retry/report logic classifies a deterministic request bug as a transient
backend failure (and any HTTP front door reaching this error answers 500 for caller-attributable
input).

---

### Refuted and dropped

- **Ephemeral sweep over HTTP defaults protect_referenced ON, diverging from `bd purge`** — The factual premise is accurate (handleSweep at internal/httpapi/sweep.go:129 defaults ProtectReferenced=true for all tiers; cmd/bd/purge.go leaves it false for `bd purge`), but the finding mislabels a deliberate, published wire-contract default as a defect. (1) The endpoint's own contract, internal/httpapi/spec/openapi.v0.yaml, declares `protect_referenced: default: true` with

- **HTTP issues:query lets any network peer force a whole-workspace read despite the guard built for exactly that** — The finding's mechanical facts are all true: allowUnlimited (reads.go:353) refuses only *limit==0; BuildQueryPlan leaves plan.Filter.Limit=0 for a predicate expression (query.go:97-102, only the filter-expressible else-branch sets it); storeQuerier.Query then runs SearchIssuesWithCounts on that unbounded filter (querier.go:51-59) and only trims via FinishPage after ApplyQueryPr

julianknutsen:
## Red-team fixes — blocker and both majors addressed

### BLOCKER: proxied `bd prune` deleted a bead the contract protects (`f03da662a`)

`ProtectReferenced` promises the reference scan covers every not-done row's
description, notes **and comments**, and the not-done set spans both planes.

The store-backed body honours that — `GetCommentsForIssuesInTx` partitions ids
between `comments` and `wisp_comments`. The unit-of-work body called a use case
that hard-codes `UseWispsTable:false` and reads the durable table only, so a
wisp's comments were never scanned.

A closed bead whose **only** citation is a comment on an open wisp was protected
on the direct route and **permanently deleted** on the proxied one. The same
`bd prune --pattern X --force`, opposite outcomes, on the operation whose entire
job is deleting what it matched. The uow route's own `Commenter` writes wisp
comments to `wisp_comments`, so `bd comment <wisp> "see X"` creates exactly the
state the scan missed.

`RunSweeperProtectsCitedRows` cites through a durable bead's *description*.
Three of the four description/comment × durable/wisp combinations had cases; the
fourth — the one where the two bodies read comments through different code — had
none, and all three legs were green over the hole. Verified by mutation: with the
fix reverted the uow leg reports `Swept=1` (the protected bead deleted).

### `bd query --sort` returned different pages per route (`650b0cd2a`)

`BuildListFilter` pushes `SortBy` into the SQL filter; `BuildQueryPlan` did not.
So the database bounded the page in **storage** order and the display sort ran
afterwards in Go — making "which N rows are in the page" a property of whatever
order the engine returned.

The store body over-fetches one row as a has-more probe and `FinishPage` sorted
that probe **into** the page before cutting; the uow body trims natively and
never sees it. Pushed down, the bound and the order are one decision and the
probe row is genuinely last.

Note for reviewers: the new case sorts by `title`, not `id` — sqlbuild routes
`id` to a **Go-side** sort, so no push-down can apply there. A first draft used
`id` and passed against the unfixed code for exactly that reason.

### uow `BatchCreator` accepted a cross-plane in-batch edge (`9b44bb503`)

The contract says an edge between the durable and ephemeral buckets "cannot be
written by the batch that creates both of its ends". The store bodies assign
every id first and see the batch as a **set**, so they refuse. The uow body
creates item by item, so by the time the edge is written its target is an
ordinary existing row and the domain layer writes it — refused whole by two
backends, landed in full on the third.

`ValidateCreateBatchPlanes` states the rule once and the uow body runs it before
writing. The store refusal now goes through the same constructor, which also
makes it match `ErrValidation` — it was a bare `fmt.Errorf`, so the role's
promise that this refusal is the caller's fault was true of one body and not the
other. That was the council's related minor.

### Also: `.golangci.yml` was silently disabling every linter (`85e9674b0`)

A merge union split an exclusion rule across a hunk boundary, leaving a bare
`- path:` with no `linters:`. golangci-lint rejects the whole config for that —
a **config** error, not a finding, so every linter stopped running on this
branch. Union is right for CHANGELOG prose and wrong for a structured document
where a hunk can split a record.

julianknutsen:
Landed on `main` via #5378 (merge commit `5e1c8697b`), which carried the whole stack — every commit on this branch is an ancestor of `main`, verified with `git merge-base --is-ancestor`.

Closing as superseded rather than merging: this PR's base is the wave beneath it, so merging it would not have reached `main`. The per-wave review and its council comment stay here for the record.

PR #5378, opened 2026-08-06
facade 4/4: list, delete, init identity, dep tree + the merge and integration

julianknutsen:
**Stacked 4 of 4** — merges into `facade/wave-3`. Review #5375, #5376, #5377 first.

The last wave, plus the merge with `main` and the integration work. This is the
PR that carries the series onto `main`.

| command | role | HTTP |
|---|---|---|
| `bd list` (the page) | `Reader.List` | `GET /v0/beads/issues` |
| `bd list` blocking decoration | `BlockingAnnotator` | `GET /v0/beads/dependencies/blocking` |
| `bd delete` (+ `cleanup`, `wisp gc`, `mol burn`) | `Deleter` | `POST /v0/beads/issues:delete` |
| `bd init` identity | `Bootstrapper` / `InitVerifier` | — |
| `bd dep tree` | `TreeWalker` | `GET /v0/beads/dependencies/tree` |

### The defect worth reading

**The delete cascade rewrote SURVIVING rows' text to say a live issue was
deleted.** Two call sites passed different argument sets to the same closure —
one all ids, one with wisps excluded — so `bd delete <wisp> --cascade` left a
durable row reachable only through that wisp alive and then rewrote its
neighbours' descriptions to call it `[deleted:<id>]`. One `DeletionSet`,
resolved once and passed down, makes that class of drift impossible rather than
merely fixed.

The Deleter suite was **59 cases green on three legs** while that bug was live,
because no case seeded a cross-plane dependent edge. **Green is not coverage** —
that is the most useful thing this programme learned.

Also: **proxied `bd delete --force` no longer force-cascades.** It passed
`Cascade: true` unconditionally and refused the `--cascade` flag outright, so
deleting a gate destroyed the bead it had just unblocked. User-visible; in the
CHANGELOG.

### What is NOT behind a role, and why

Each is a decision. `bd show`/`bd close` were already converged. `bd version`
has no substrate. Most of `bd init` CREATES the store a role is reached
through — wave 4 carved out only the identity slice, split into a guarded
writer and a reader because bd reads this identity on paths where it is
forbidden to write one. `bd create --graph` is deferred on the record: it is an
upsert with caller-supplied names where `BatchCreator` is create-only with
server-assigned ids, and folding them means one role with two incompatible id
policies.

### How much three green legs prove

For most roles `dolt` and `embeddeddolt` share one validate/execute body, so a
three-legged run is usually **two independent votes, not three** — and for
`TreeWalker` it is **one**. Stated in the contracts rather than papered over.

### The merge and the integration work

`main` grew a second `bd serve` arm while this was in flight. It wires
`httpapi.Config` from role accessors, and this series grew that Config to
thirteen roles and made `Listen` refuse a partial set. Both sides are correct
and merge cleanly **as text**; together they bind nothing. Caught by main's own
tests, not by reading.

A review council over the merged series (six lenses, every finding handed to a
second agent instructed to refute it) found one blocker, three majors and
eleven minors. The blocker: the merge kept main's cascade-parity test, whose
subtest deliberately PINS the data-loss defect this series fixed, so it failed
deterministically on a shard that runs on push-to-main. All fixed here except
proxied `bd create` silently dropping `--storage-class`, which is pre-existing
and filed.

Finally, a comment trim: ten agents over 216 files brought this series' prose
from 34.7% to 31.6% of added lines, verified comments-only by a guard that
strips all comments and whitespace and requires the result to be byte-identical.
`issueops/` was deliberately left untrimmed — the contracts cite those docs by
line number 155 times.


julianknutsen:
## Red-team council — PR 5378 (wave-4)

**Confirmed: 3 major, 1 minor. Refuted and dropped: 3.**

### MAJOR: UOW delete body's Dependencies count misses cross-plane inbound edges — the two routes report different numbers for the same delete

`internal/storage/domain/issue_delete.go:74`

The store-backed body (DeleteInTx → DeleteResolvedSetInTx,
internal/storage/issueops/delete.go:196-258) counts outbound edges per plane PLUS an inbound scan
over BOTH dep tables where the target is in the deletion set. The UOW body (deleteMany) instead
computes DependenciesCount as CountAllForIDs(regularIDs, dependencies) + CountAllForIDs(wispIDs,
wisp_dependencies) (issue_delete.go:74-82). Each query matches `issue_id IN (ids) OR <target> IN
(ids)` within ONE plane's id set against ONE table (db/dependency.go:713-752), so a cross-plane edge
whose deleted end is the TARGET in the other plane's table is counted by neither query: (a) durable
A → deleted wisp W (row in `dependencies`, depends_on_wisp_id=W; W is only checked against
wisp_dependencies), and (b) surviving wisp D → deleted durable R (row in `wisp_dependencies`,
depends_on_issue_id=R; R is only checked against `dependencies`). Both edges ARE actually removed
(the new explicit sync-plane delete at issue_delete.go:141 for (a); FK fk_wisp_dep_issue_target ON
DELETE CASCADE for (b)), so the count under-reports real removals. Add

**Failure scenario.** Wisp W has durable dependent A (edge A→W in `dependencies`). `bd delete W --force` on the direct CLI
route prints "Removed 1 dependency link(s)" and emits dependencies_removed:1 in --json; the same
command with --proxied-server (and POST /v0/beads/issues:delete via a UOW-backed `bd serve`,
`dependencies` field) reports 0 — while actually deleting that edge. Symmetrically for deleting a
durable row with a surviving wisp dependent. Any agent or script comparing/parsing these numbers
sees the two routes disagree about the same operation.

### MAJOR: Status prune corrupts `both` walks when a node repeats between halves: survivor's ancestors dropped, orphan nodes returned

`internal/storage/issueops/tree_walk.go:71`

PruneTreeByStatus builds `parentOf` keyed by node ID (line 67-74) and `keep` keyed by node ID (line
76-98), but WalkDependencyTreeInTx applies it AFTER MergeBidirectionalTree (line 152-153), where the
contract explicitly permits the same ID to appear in BOTH halves with a different ParentID in each
(issueops/treewalker.go:179-188). The map is last-write-wins, so the down-half copy's ParentID
overwrites the up-half copy's, and any survivor in the UP half has its ancestor chain walked through
the DOWN half's parents. The up-half ancestors are dropped while both copies of the duplicated node
are kept, producing nodes whose ParentID names an ID absent from the answer — exactly the 'scatter
of orphans' the prune promise (treewalker.go:94-96, tree_walk.go:52-56) says cannot happen. Verified
empirically: a scratch test over merged input [A(p=R,d1), X(p=A,d2), Y(p=X,d3,closed), R(d0),
X(p=R,d1)] with status=closed returns X(p=A) with A absent. The duplicate-between-halves state is
reachable with supported verbs (no raw SQL needed): the walk follows every type but relates-to while
cycle refus

**Failure scenario.** Graph: A blocks-depends-on R, X discovered-from-depends-on A, Y blocks-depends-on X (Y closed), R
parent-child-depends-on X — all creatable via `bd dep add`. `bd dep tree R --direction both --status
closed` (or GET /v0/beads/dependencies/tree?root_id=R&direction=both&status=closed) returns the up-
half X with ParentID=A while A is pruned from the answer: JSON/HTTP consumers rebuilding the tree
from Depth+ParentID hit a dangling parent reference, the CLI renderer draws X at the wrong
indentation under a missing parent, and the non-matching down-half copy of X is kept as a spurious
extra node.

### MAJOR: bd init on a half-identified substrate silently mints divergent per-rig project ids (and never completes a missing prefix), arming later PROJECT IDENTITY MISMATCH hard-fails

`cmd/bd/init.go:249`

seedInitWorkspaceIdentity adopts whenever EITHER marker is present (line 249-253: `found.Prefix !=
"" || found.ProjectID != ""` -> return nil, no write), while resolveInitProjectID (non-gateway, line
154-160) still mints a FRESH project id into metadata.json when the DB carries none (adoptedFromDB
== ""). On wave-3 this quadrant was completed: init wrote _project_id to the DB via
shouldWriteProjectIDLocally (confirmed in `git show facade/wave-3:cmd/bd/init.go`), so the first rig
stamped the DB and every later rig ADOPTED that id. Now each rig's init mints a different UUID
locally and the DB stays without one, with no warning printed (the 'Adopted' message is gated on
adoptedFromDB != ""). The CHANGELOG entry documents 'stays half-identified' but not the divergent
minting or its interaction with `bd doctor --fix`: FixMissingProjectID
(cmd/bd/doctor/fix/metadata.go:177-182) backfills the DB with ONE rig's local id, after which every
other rig's open hard-fails verifyProjectIdentity (internal/storage/dolt/store.go:1947-1959) with
advice text ('Do NOT run bd init — your data likely exist

**Failure scenario.** Pre-GH#2372 database (issue_prefix set, no _project_id — any DB bootstrapped by an older bd). Rig A:
`bd init --database proj` -> metadata.json project_id=UUID-A, DB untouched, no output. Rig B: same
-> UUID-B. Rig A later runs `bd doctor --fix` -> DB _project_id=UUID-A. Rig B's next `bd list`
refuses to connect: 'PROJECT IDENTITY MISMATCH' naming UUID-B vs UUID-A, with recovery advice
pointing at the wrong cause. Pre-wave-4, rig A's init stamped the DB and rig B adopted UUID-A — no
divergence window existed. Second scenario: DB with _project_id but no issue_prefix (e.g. doctor-fix
backfill before first init) -> `bd init` succeeds silently, then every issue-creating command fails
for want of

### MINOR: Reference-rewrite regex skips the second of two adjacent citations separated by a single character

`internal/storage/issueops/delete_role.go:330`

DeletedReferencePattern is `(^|[^A-Za-z0-9_-])(id)($|[^A-Za-z0-9_-])` and the rewrite uses
ReplaceAllString. Go's regexp finds non-overlapping matches in the original text: the trailing
boundary character is CONSUMED by group 3, so when the same id occurs twice separated by exactly one
separator character, the second occurrence has no unconsumed boundary character left for group 1 and
is not matched. The role doc (issueops/deleter.go:243-247) promises "Each occurrence of a deleted id
... becomes [deleted:<id>]". The domain body (issue_delete.go:312) has the identical pattern, so the
two legs agree with each other but both fall short of the doc; the conformance rewrite case
(RunDeleterRewritesReferencesInNeighbors) never seeds adjacent duplicate citations. Lookahead-free
fix: use a replacement that does not consume the trailing boundary (e.g. rewrite via
FindAllStringIndex with manual boundary checks, as the doc claims ProtectReferenced does).

**Failure scenario.** Neighbor description "dup of bd-7 bd-7" (same id twice, single space between) with an edge to bd-7;
`bd delete bd-7 --force` rewrites it to "dup of [deleted:bd-7] bd-7" — the surviving verbatim
citation now reads as a live reference to a row that no longer exists, on every backend.

---

### Refuted and dropped

- **Cascade closure and dependents guard treat pure annotation edges as "depends on" — the new role doc and contra** — The finding's facts check out but its classification does not survive. (1) The untyped cascade/guard is pre-existing: facade/wave-3's delete.go already contains the identical FindAllDependentsInTx and untyped ExternalDependents queries (file history back to 0394021da, well before this series); wave-4 moved and converged routes onto it without changing semantics. (2) The claimed

- **BlockingAnnotator legs disagree on the be-iabdi cross-plane duplicate id: shared body prefers wisps status, uo** — Every code citation in the finding checks out (shared body prefers wisps at dependency_queries.go:753; uow leg hard-errors at dependency.go:569; direct route swallows, proxied route fails), but the defect is not attributable to this diff and violates no contract the diff makes. Verified against the diff base: at facade/wave-3 the dup hard-error already exists at dependency.go:5

- **Bootstrap prefix normalization strips ALL trailing hyphens while the settings plane it claims to mirror strips** — The TrimRight-vs-TrimSuffix difference is real at the code level but the defect does not reproduce. (1) The filed scenario is false: both real callers of Bootstrap — cmd/bd/init.go:271 (after normalizeIssuePrefix at init.go:809/2586) and cmd/bd/init_proxied_server.go:219 (after resolveInitPrefix at lines 285-291) — already fully strip trailing hyphens with the same TrimRight be

julianknutsen:
## Red-team fixes — all three majors addressed

### `both` walks returned orphans under `--status` (`d55cc2719`)

`PruneTreeByStatus` keys its parent chain by node ID, and it ran on the
**merged** list. A `both` answer may legitimately carry one id twice with a
different `ParentID` in each half — `TreeResult.Nodes` says so — so one
last-write-wins map answered for two different facts: a survivor in the UP half
had its ancestors looked up through the DOWN half's parents. The up ancestors
were dropped and the answer came back with nodes whose `ParentID` named an id
absent from it. That is the scatter of orphans the prune promise exists to
prevent; a renderer rebuilding the shape from `Depth` and `ParentID` draws them
at the wrong indentation.

Pruned per half, each id appears once in the map it is looked up in.
`pruneBidirectionalByStatus` keeps the merge's root invariant while doing it —
each half's survivors chain to the root, and the merge takes the root from the
DOWN half, so an up half that survived a prune the down half did not would
otherwise lose the node its chain ends at.

**The fixture is the work.** `mid` has to reach the root by one path walking
down and another walking up, which takes a three-node ring —
`root --parent-child--> mid --discovered-from--> side --blocks--> root`. One
edge is discovered-from, so it is not a *blocking* cycle and the ordinary
dependency writer accepts it; no raw SQL needed. A first draft used a node with
the same parent in both halves and passed against the broken code.

The case asserts the **invariant** rather than a node list: every `ParentID`
must name a node that is present.

### uow delete under-reported the edges it removed (`bf02f79ec`)

`DeleteResult.Dependencies` is a claim about what the delete did, and both
routes print it. The uow body counted each plane's ids against that plane's
table only, so two shapes escaped both queries: a durable row depending on a
deleted **wisp**, and a surviving wisp depending on a deleted **durable** row.
Both edges really are removed, so the count under-reported real removals while
the store bodies reported them.

The old predicate also **double-counted**: `issue_id IN (batch) OR target IN
(batch)` ran per 50-id batch, so an edge whose ends fell in different batches
matched twice. Keying by the edge itself removes both hazards.

### `bd init` silently minted divergent per-clone project ids (`74a732ba1`)

`Bootstrapper` refuses to complete a half-identified substrate deliberately — it
cannot tell a half-written bootstrap from a deliberately half-provisioned
database. That refusal is right. What was wrong is that init adopted the state
in **silence**.

With a prefix present and no `_project_id` there is nothing to adopt, so every
clone's init mints a different local project id with no output at all. The first
`bd doctor --fix` backfills the database from whichever clone ran it, and every
other clone's next open hard-fails `PROJECT IDENTITY MISMATCH` — with recovery
advice that names the wrong cause. Before wave 4 the quadrant was completed by
init writing `_project_id` itself, so no divergence window existed.

init still cannot fix it, but the operator now learns the state, what happens if
it is left alone, and which command settles it. Both halves are named because
they fail differently: no prefix means every later open reports the workspace as
uninitialized.

---

**Verification.** Every fix in this stack was mutation-tested — the new case run
against the original body to confirm it fails. Three of my first drafts passed
against broken code and were rebuilt; the reasons are recorded in the fixtures.
`gofmt`, `go vet`, `golangci-lint` (0 issues), and the full contract suite at all
three backends are green at this tip.

PR #3991, opened 2026-05-16
fix(storage): preserve mixed issue and wisp reads

julianknutsen:
## Summary
- preserve dependency reads and dependents across promotion/demotion between `issues` and `wisps`
- make `GetReadyWork` include no-history beads stored in `wisps` by default while still excluding true ephemeral wisps
- commit promoted auxiliary rows (`labels`, `dependencies`, `events`, `comments`) with the promoted issue
- stage newly-created ignored schema tables during migrations without sweeping pre-existing dirty user data

## Tests
- `go test ./internal/storage/issueops ./internal/storage/dolt -run 'Wisp|Demote|Promote|Dependency|Delete|GetIssue|Mixed|SchemaParityIssuesVsWisps|UpdateIssueIDUpdatesWispTables|GetCommentsForIssues_MixedWispAndPermanent|SchemaRunsInitWhenStale|GetReadyWorkIncludesNoHistoryWispsByDefault|GetReadyWorkExcludesBlockedNoHistoryWisp' -count=1`
- `go test ./cmd/bd -run 'Ready|NoHistory|Promote|Delete' -count=1`
- `go test ./internal/storage/issueops -count=1`
- `git diff --check`

PR #5381, opened 2026-08-06
fix(schema): repair the wisp_dependencies FK shape before 0058 applies

### LINKED ISSUE #5380: bug: migration 0058 cannot apply to the databases it targets — FK on a base column of a stored generated column
## Summary

Migration `0058_heal_wisp_dependencies_split_constraints` cannot apply to the databases it was written to heal. On any store where `wisp_dependencies.depends_on_id` is a STORED generated column, Dolt rejects the two foreign keys the migration adds, `bd migrate` aborts, and the store is left half-migrated at v57 with no working client: the new binary re-fails at 0058 on every retry, and the previous binary skew-refuses a v57 cursor. Recovery requires a restore.

## Symptom

Upgrading a v53 database with any binary at or after 0058:

```
Error: failed to open database: failed to initialize schema: schema migration:
  migration 0058_heal_wisp_dependencies_split_constraints.up.sql:
  Error 1105 (HY000): Cannot add foreign key on the base column of a stored generated column.
schema: release migration lock: schema migration lock release failed: driver: bad connection
```

Afterwards:

```sql
SELECT MAX(version) FROM schema_migrations;  -- 57
```

0054 through 0057 committed. 0058 did not. Re-running `bd migrate` reproduces the identical error and the cursor stays at 57. The second line is also a real symptom, not noise: the migration lock release fails on the same broken connection, so the failure does not unwind cleanly.

## Reproduction

Verified against `origin/main` at `4ebc98927` with a build from that exact commit, on a Dolt sql-server (`dolt version 2.1.2`), using a schema copied verbatim from a live v53 database.

The trigger is the table shape, not the data. Any `wisp_dependencies` carrying the stored generated column reproduces it:

```sql
CREATE TABLE `wisp_dependencies` (
  `issue_id` varchar(255) NOT NULL,
  `type` varchar(32) NOT NULL DEFAULT 'blocks',
  `created_at` datetime DEFAULT CURRENT_TIMESTAMP,
  `created_by` varchar(255) DEFAULT '',
  `metadata` json DEFAULT (json_object()),
  `thread_id` varchar(255) DEFAULT '',
  `depends_on_issue_id` varchar(255),
  `depends_on_wisp_id` varchar(255),
  `depends_on_external` varchar(255),
  `depends_on_id` varchar(255) NOT NULL GENERATED ALWAYS AS
    (coalesce(`depends_on_issue_id`,`depends_on_wisp_id`,`depends_on_external`)) STORED,
  PRIMARY KEY (`issue_id`,`depends_on_id`),
  KEY `idx_wisp_dep_external_target` (`depends_on_external`),
  KEY `idx_wisp_dep_issue_target` (`depends_on_issue_id`),
  KEY `idx_wisp_dep_type` (`type`),
  KEY `idx_wisp_dep_type_target` (`type`,`depends_on_id`),
  KEY `idx_wisp_dep_wisp_target` (`depends_on_wisp_id`),
  CONSTRAINT `ck_wisp_dep_one_target` CHECK (
    (((NOT(`depends_on_issue_id` IS NULL)) + (NOT(`depends_on_wisp_id` IS NULL)))
      + (NOT(`depends_on_external` IS NULL))) = 1)
);
```

Note what is and is not present: `ck_wisp_dep_one_target` exists, `fk_wisp_dep_wisp_target` and `fk_wisp_dep_issue_target` do not. That is exactly the delegate-path state 0058's header describes.

Steps:

1. Start a `dolt sql-server` and create a database with the full beads schema at v53 (all 25 tables), including the `wisp_dependencies` shape above.
2. Seed `schema_migrations` with versions 1 through 53, then `CALL DOLT_ADD('-A')` and `CALL DOLT_COMMIT(...)`. The migration refuses to start against a dirty working set.
3. Point a `.beads/metadata.json` at that server in `dolt_mode: server`.
4. `bd migrate`

Result: expected the cursor to advance to head → got `Error 1105` at 0058 with the cursor stranded at 57.

The minimal statement, runnable directly against such a table:

```sql
ALTER TABLE wisp_dependencies
  ADD CONSTRAINT fk_wisp_dep_wisp_target
  FOREIGN KEY (depends_on_wisp_id) REFERENCES wisps(id)
  ON DELETE CASCADE ON UPDATE CASCADE;
-- Error 1105 (HY000): Cannot add foreign key on the base column of a stored generated column.
```

## Root cause

`internal/storage/schema/migrations/0058_heal_wisp_dependencies_split_constraints.up.sql:202` and `:207` add foreign keys on `depends_on_wisp_id` and `depends_on_issue_id`:

```sql
'ALTER TABLE wisp_dependencies ADD CONSTRAINT fk_wisp_dep_wisp_target  FOREIGN KEY (depends_on_wisp_id)  REFERENCES wisps(id)  ON DELETE CASCADE ON UPDATE CASCADE'
'ALTER TABLE wisp_dependencies ADD CONSTRAINT fk_wisp_dep_issue_target FOREIGN KEY (depends_on_issue_id) REFERENCES issues(id) ON DELETE CASCADE ON UPDATE CASCADE'
```

Both columns are inputs to `depends_on_id`'s `coalesce(...)` STORED expression, and Dolt refuses a foreign key on a base column of a stored generated column.

The migration's guards check three things: that the constraint is absent, that the target table exists, and that the target columns exist. None of them tests whether those columns feed a stored generated column. So the guards pass precisely on the population the migration targets, and the DDL then fails.

The two conditions share a cause. A database reaches the delegate path when `wisps` and `wisp_dependencies` both already exist at 0047, and that same path produces the split target columns underneath the `coalesce` generated column. Every database that needs 0058 has the shape that makes 0058 fail.

`SET FOREIGN_KEY_CHECKS = 0` at the top of the migration does not help: the error is a schema-level rejection of the constraint definition, not a row-validation failure.

## Fix candidates

**A. Guard the FK adds on the generated-column shape.** Extend the existing `INFORMATION_SCHEMA` preamble with a check on `COLUMNS.EXTRA` / `GENERATION_EXPRESSION` for `depends_on_id`, and skip the two `ADD CONSTRAINT ... FOREIGN KEY` statements when the target columns are bases of a stored generated column. The `ck_wisp_dep_one_target` add is unaffected and still runs.
- Scope: ~20 lines, one migration file
- Trade-off: unblocks the upgrade immediately and matches the existing guard style, but leaves the two FKs permanently absent on these databases. Referential integrity for wisp dependency targets stays enforced only by application code.

**B. Drop and rebuild the generated column around the constraint adds.** Drop `depends_on_id` (and the primary key and indexes that depend on it), add the two foreign keys, then recreate the generated column, PK, and indexes.
- Scope: ~60 lines, one migration file, plus a down-migration
- Trade-off: converges every database on one schema and actually restores the missing integrity constraints. But it rewrites the primary key of a live table inside a migration, which is the highest-risk operation in the chain, and `0041_split_dependencies_target` shows how much machinery that needs (`DELETE FROM dolt_nonlocal_tables`, an intermediate `DOLT_COMMIT`, `FOREIGN_KEY_CHECKS = 0`). On a large `wisp_dependencies` it is also slow.

**Recommendation:** A, as a fix that can ship now, with B tracked separately if the missing foreign keys matter for correctness rather than hygiene. The immediate problem is that a routine upgrade strands the store; the missing FKs have evidently been absent on these databases since 0047 without a reported consequence. Maintainer's call.

One thing is worth fixing independently of which candidate lands. 0058 is not the last migration, so a failure there aborts a chain that has already committed 0054 through 0057, and a store stranded mid-chain is readable by no binary: the new one re-fails at the same migration, the old one skew-refuses the advanced cursor. If migrations cannot be made transactional as a group, a failure that leaves the cursor ahead of the last binary-compatible version deserves an error that says so and names the recovery path.

## Adjacent / out of scope

- `#5257` (ignored-migration cursor trusted when required columns are missing) and `#5033` (cursor the schema contradicts, fixed by PR `#5190`) are the same family of "cursor and actual schema disagree" but a different mechanism. Not proposing to fold them together.
- `#5107` (schema generation skew between bd 1.0.5 and older readers) is adjacent to the stranded-store consequence but is about readers, not a failing migration.
- `#4380` (aux row re-key crashes the migration pass, fix: skip-and-warn per table) is the closest structural precedent for a per-statement skip-and-warn policy, which is essentially what candidate A does for one migration.
- Not proposing anything about the `dependencies` table. Only `wisp_dependencies` is affected here.

## Maintainer context

Found while moving a Gas City install off a pinned `v1.1.0`-era build onto main to pick up the leasing stack (`#4537`). The upgrade was rehearsed against an isolated copy of the production schema before installing, which is where this surfaced; the live store was never migrated. Filed from `ds-research`.

sjarmak:
Fixes #5380.

## What breaks

`0058_heal_wisp_dependencies_split_constraints` cannot apply to the databases it was written to heal. Its two `ADD CONSTRAINT ... FOREIGN KEY` statements target `depends_on_wisp_id` and `depends_on_issue_id`, which are inputs to `depends_on_id`'s `COALESCE` STORED generated column, and Dolt rejects that:

```
Error 1105 (HY000): Cannot add foreign key on the base column of a stored generated column.
```

The two conditions share a cause. A database reaches 0047's delegate branch when `wisps` and `wisp_dependencies` both already exist, and that branch is what adds the split target columns underneath the generated column. Every database that needs 0058 has the shape that makes 0058 fail.

The failure strands the store. The pass commits 0054 through 0057 and aborts, leaving the cursor at 57: past the last version an older binary will read, short of head, so no binary opens it afterwards. Recovery is a restore.

## Why this is in Go rather than in the migration

Two reasons.

0058 is shipped and frozen by `check-migration-hygiene.sh` check C, and it cannot be fixed forward with a new migration either, because the failing file aborts the pass before any later version runs. `migration_repairs.go`'s own header already names this case: repairing in code, keyed to the pending version, is the only path that heals affected databases without touching shipped SQL.

The guard also cannot be written in SQL. Dolt does not surface generated-column metadata through `INFORMATION_SCHEMA` at all:

```sql
SELECT column_name, extra, generation_expression FROM information_schema.columns
 WHERE table_name='wisp_dependencies' AND column_name='depends_on_id';

+---------------+-------+-----------------------+
| COLUMN_NAME   | EXTRA | GENERATION_EXPRESSION |
+---------------+-------+-----------------------+
| depends_on_id |       |                       |
+---------------+-------+-----------------------+
```

Both empty, and `IS_GENERATED` does not exist (`Error 1105: column "is_generated" could not be found in any table in scope`). A guard written against those columns compiles, runs, and never fires. `SHOW CREATE TABLE` is the only place the shape is visible.

## The change

One entry in the existing `preMigrationRepairs` registry, `{"schema_migrations", 58}`, alongside the repairs already registered for 0040, 0041, 0047 and 0053.

The repair adds both foreign keys itself: drop the generated column and the primary key and index built on it, add the constraints while the base columns are unencumbered, then restore column, key and index. Values are recomputed from the base columns on restore, so no data is lost. 0058 then finds both constraints present and takes its no-op branch.

0058 itself is untouched. `check-migration-hygiene.sh` passes.

It is idempotent and conservative, no-oping when the table is absent, when both constraints already exist, when `depends_on_id` is not a stored generated column (0058 applies cleanly there), or when the primary key is not the shape it knows how to rebuild. The half-repaired case, where one constraint landed on an earlier attempt, adds only the missing one.

## Test plan

`internal/storage/schema/wisp_dep_fk_repair_test.go`, four tests: the three no-op branches, the full rebuild sequence on the drifted shape, the half-repaired case, and the registry wiring. The sequence test pins statement order, which is load-bearing in both directions.

I mutation-checked these rather than trusting green. Removing the drop-first steps fails two of them; unregistering the repair fails the registry test.

```
go test ./internal/storage/schema/   ok  91.5s
go vet ./internal/storage/schema/    clean
scripts/check-migration-hygiene.sh   Migration hygiene OK
```

End to end, against a Dolt sql-server holding a replica of a production v53 schema with `wisp_dependencies` rows of each target kind (wisp, issue, external):

| | before | after |
|---|---|---|
| `bd migrate` from v53 | aborts at 0058, cursor stranded at 57 | reaches head (62) |
| `wisp_dependencies` foreign keys | 0 | 2 |
| rows | 3 | 3, targets intact |
| `ck_wisp_dep_one_target` | present | present |

Worth noting for review: `depends_on_id` is absent from the final table, replaced by the `id char(36)` surrogate key. That is `ignored/0005` doing its designed job downstream, and it only restores a foreign key it detected before its own drop, which is why the repair has to run at 58 rather than later.

## Alternative considered

Dropping and rebuilding the generated column inside a new migration rather than a repair hook. It cannot work: a new migration is numbered above 58, and 0058 aborts the pass before anything above it runs.

### DIFF
--- a/internal/storage/embeddeddolt/migrate_wisp_dep_forward_repair_test.go
+++ b/internal/storage/embeddeddolt/migrate_wisp_dep_forward_repair_test.go
@@ -1,0 +1,434 @@
+//go:build cgo
+
+package embeddeddolt_test
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+
+	"github.com/steveyegge/beads/internal/storage/schema"
+)
+
+// Real-Dolt coverage for the pre-0058 wisp_dependencies forward repair
+// (repairWispDependenciesForwardShape, internal/storage/schema).
+//
+// Why this cannot be a sqlmock test. Every load-bearing fact in that repair is
+// Dolt semantics a statement-echo mock cannot exercise: that Dolt rejects a
+// foreign key on the base column of a stored generated column at all; that
+// normalizing a multi-target row recomputes the COALESCE generated column and
+// so can collide on the legacy composite primary key; that DEFAULT (UUID())
+// mints a distinct value per existing row, which is what makes the MIN(id)
+// dedup deterministic. A mock asserting the statement ORDER would have passed
+// against the first draft of this repair, which normalized before the drop and
+// aborted on real Dolt with "duplicate primary key given: [w2,external:e1]".
+//
+// The two subtests are the two populations that actually exist. Measured
+// against a live 25-database Dolt server carrying the affected shape: 8
+// databases are on the delegate path with no target foreign keys, and every one
+// of the 14 databases still holding the generated column has no id column,
+// confirming the lineages are disjoint.
+
+// legacyDelegateWispDepDDL is the shape 0047's delegate branch leaves behind:
+// the three split target columns present, depends_on_id still a STORED
+// generated column over them, the composite primary key, and neither target
+// foreign key. Taken verbatim from a production SHOW CREATE TABLE rather than
+// hand-idealized, minus ck_wisp_dep_one_target so the multi-target rows this
+// test seeds are insertable (a store carrying the check cannot hold them, which
+// is the second subtest).
+const legacyDelegateWispDepDDL = `
+CREATE TABLE wisp_dependencies (
+    issue_id VARCHAR(255) NOT NULL,
+    type VARCHAR(32) NOT NULL DEFAULT 'blocks',
+    created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
+    created_by VARCHAR(255) DEFAULT '',
+    metadata JSON DEFAULT (JSON_OBJECT()),
+    thread_id VARCHAR(255) DEFAULT '',
+    depends_on_issue_id VARCHAR(255) NULL,
+    depends_on_wisp_id VARCHAR(255) NULL,
+    depends_on_external VARCHAR(255) NULL,
+    depends_on_id VARCHAR(255) NOT NULL GENERATED ALWAYS AS (COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external)) STORED,
+    PRIMARY KEY (issue_id, depends_on_id),
+    KEY idx_wisp_dep_external_target (depends_on_external),
+    KEY idx_wisp_dep_issue_target (depends_on_issue_id),
+    KEY idx_wisp_dep_type (type),
+    KEY idx_wisp_dep_type_target (type, depends_on_id),
+    KEY idx_wisp_dep_wisp_target (depends_on_wisp_id)
+)`
+
+func TestEmbeddedPre0058RepairMarchesDriftedWispDependenciesToFinalShape(t *testing.T) {
+	requireEmbedded(t)
+	ctx := t.Context()
+
+	// The shape a fresh, fully migrated store converges on.
+	//
+	// The comparison is structural (columns with their types and defaults, the
+	// primary key, every unique key, every foreign key with its referenced
+	// table and columns, every check) rather than SHOW CREATE text equality.
+	// Text equality is the wrong bar, and provably so: a legacy table healed by
+	// the SHIPPED path keeps its original column order, because nothing in the
+	// chain reorders columns. The already-migrated production database this
+	// work came from has wisp_dependencies as (id, issue_id, type, created_at,
+	// created_by, metadata, thread_id, depends_on_issue_id, depends_on_wisp_id,
+	// depends_on_external) where a fresh store puts the three target columns
+	// third through fifth. Asserting text equality would fail every correctly
+	// healed store, including ones this repair never touched. Legacy secondary
+	// index names survive for the same reason -- a fresh store names its
+	// foreign-key backing indexes after the constraints because no index
+	// existed to reuse, while a legacy table already has idx_wisp_dep_*_target
+	// covering those columns and Dolt reuses them. Index COVERAGE is asserted;
+	// index NAMES are not.
+	fresh := readWispDepShape(t, ctx, referenceStoreConn(t, ctx))
+
+	t.Run("delegate-path store carrying rows the final shape rejects", func(t *testing.T) {
+		dataDir := seedMainSchemaAt(t, ctx, 53)
+		conn, closeConn := openPinnedConn(t, ctx, dataDir)
+
+		mustExecConn(t, ctx, conn, "DROP TABLE wisp_dependencies")
+		mustExecConn(t, ctx, conn, legacyDelegateWispDepDDL)
+
+		// Referenced rows. issue_id and depends_on_wisp_id both point at
+		// wisps(id); depends_on_issue_id points at issues(id).
+		for _, id := range []string{"w1", "w2", "w3"} {
+			seedWisp(t, ctx, conn, id)
+		}
+		seedIssue(t, ctx, conn, "i1")
+
+		// Two rows that must survive untouched.
+		seedWispDep(t, ctx, conn, "w1", nil, strptr("w2"), nil)
+		seedWispDep(t, ctx, conn, "w3", strptr("i1"), nil, nil)
+		// Three foreign-key orphans, one per constraint the repair adds.
+		// "ghost" is not a wisp, so this row orphans fk_wisp_dep_issue --
+		// which 0058 does not clean, because 0058 does not add that key.
+		seedWispDep(t, ctx, conn, "ghost", nil, strptr("w2"), nil)
+		seedWispDep(t, ctx, conn, "w1", nil, strptr("ghost-w"), nil)
+		seedWispDep(t, ctx, conn, "w1", strptr("ghost-i"), nil, nil)
+		// A multi-target row whose normalization lands it on the natural
+		// identity of the row below it. This pair is the regression for the
+		// ordering bug: normalize-before-drop rewrites this row's generated
+		// primary key onto the other's and aborts.
+		seedWispDep(t, ctx, conn, "w2", nil, strptr("w3"), strptr("external:e1"))
+		seedWispDep(t, ctx, conn, "w2", nil, nil, strptr("external:e1"))
+
+		commitSeed(t, ctx, conn, "test: reproduce the pre-0058 drifted wisp_dependencies")
+		closeConn()
+
+		conn2, closeConn2 := openPinnedConn(t, ctx, dataDir)
+		defer closeConn2()
+
+		// Without the repair this fails with
+		// "Cannot add foreign key on the base column of a stored generated
+		// column" and strands the cursor at 57.
+		if _, err := schema.MigrateUp(ctx, conn2); err != nil {
+			t.Fatalf("MigrateUp over a drifted wisp_dependencies: %v", err)
+		}
+		if v := currentMainVersion(t, ctx, conn2); v != schema.LatestVersion() {
+			t.Fatalf("main schema version = %d, want latest %d", v, schema.LatestVersion())
+		}
+
+		requireSameWispDepShape(t, readWispDepShape(t, ctx, conn2), fresh)
+
+		// Row outcomes, stated as the full surviving set rather than a count:
+		// a count passes even when the wrong three rows survive.
+		want := []wispDepRow{
+			{"w1", "", "w2", ""},
+			{"w2", "", "", "external:e1"},
+			{"w3", "i1", "", ""},
+		}
+		if got := readWispDeps(t, ctx, conn2); !sameWispDepRows(got, want) {
+			t.Fatalf("surviving rows = %v, want %v", got, want)
+		}
+	})
+
+	t.Run("production shape: check constraint present, table empty", func(t *testing.T) {
+		// The 8 affected databases measured on the live server: same legacy
+		// shape but carrying ck_wisp_dep_one_target and zero rows. Nothing for
+		// the cleanup or dedup to do, so this subtest isolates the structural
+		// rebuild from the row surgery.
+		dataDir := seedMainSchemaAt(t, ctx, 53)
+		conn, closeConn := openPinnedConn(t, ctx, dataDir)
+
+		mustExecConn(t, ctx, conn, "DROP TABLE wisp_dependencies")
+		mustExecConn(t, ctx, conn, legacyDelegateWispDepDDL)
+		mustExecConn(t, ctx, conn, "ALTER TABLE wisp_dependencies ADD CONSTRAINT ck_wisp_dep_one_target CHECK ((depends_on_issue_id IS NOT NULL) + (depends_on_wisp_id IS NOT NULL) + (depends_on_external IS NOT NULL) = 1)")
+		commitSeed(t, ctx, conn, "test: reproduce the production pre-0058 wisp_dependencies")
+		closeConn()
+
+		conn2, closeConn2 := openPinnedConn(t, ctx, dataDir)
+		defer closeConn2()
+
+		if _, err := schema.MigrateUp(ctx, conn2); err != nil {
+			t.Fatalf("MigrateUp over the production pre-0058 shape: %v", err)
+		}
+		requireSameWispDepShape(t, readWispDepShape(t, ctx, conn2), fresh)
+	})
+
+	t.Run("a fresh store is left alone", func(t *testing.T) {
+		// The repair must be invisible to every database that does not need
+		// it. A fresh store reaches 0058 with no generated column, so 0058
+		// applies cleanly and the repair has to no-op rather than rebuild a
+		// healthy table.
+		dataDir := seedMainSchemaAt(t, ctx, 53)
+		conn, closeConn := openPinnedConn(t, ctx, dataDir)
+		defer closeConn()
+
+		if _, err := schema.MigrateUp(ctx, conn); err != nil {
+			t.Fatalf("MigrateUp on a fresh store: %v", err)
+		}
+		requireSameWispDepShape(t, readWispDepShape(t, ctx, conn), fresh)
+	})
+}
+
+// referenceStoreConn returns a connection to a store that migrated normally to
+// head, for use as the shape the repair must converge on.
+func referenceStoreConn(t *testing.T, ctx context.Context) *sql.Conn {
+	t.Helper()
+	dataDir := seedMainSchemaAt(t, ctx, schema.LatestVersion())
+	conn, closeConn := openPinnedConn(t, ctx, dataDir)
+	t.Cleanup(closeConn)
+	if _, err := schema.MigrateUp(ctx, conn); err != nil {
+		t.Fatalf("MigrateUp on the reference store: %v", err)
+	}
+	return conn
+}
+
+type wispDepRow struct {
+	issueID, dependsOnIssueID, dependsOnWispID, dependsOnExternal string
+}
+
+func readWispDeps(t *testing.T, ctx context.Context, conn *sql.Conn) []wispDepRow {
+	t.Helper()
+	rows, err := conn.QueryContext(ctx, `
+SELECT issue_id,
+       COALESCE(depends_on_issue_id, ''),
+       COALESCE(depends_on_wisp_id, ''),
+       COALESCE(depends_on_external, '')
+FROM wisp_dependencies
+ORDER BY issue_id, depends_on_issue_id, depends_on_wisp_id, depends_on_external`)
+	if err != nil {
+		t.Fatalf("read wisp_dependencies: %v", err)
+	}
+	defer func() { _ = rows.Close() }()
+
+	var out []wispDepRow
+	for rows.Next() {
+		var r wispDepRow
+		if err := rows.Scan(&r.issueID, &r.dependsOnIssueID, &r.dependsOnWispID, &r.dependsOnExternal); err != nil {
+			t.Fatalf("scan wisp_dependencies row: %v", err)
+		}
+		out = append(out, r)
+	}
+	if err := rows.Err(); err != nil {
+		t.Fatalf("iterate wisp_dependencies: %v", err)
+	}
+	return out
+}
+
+func sameWispDepRows(got, want []wispDepRow) bool {
+	if len(got) != len(want) {
+		return false
+	}
+	for i := range got {
+		if got[i] != want[i] {
+			return false
+		}
+	}
+	return true
+}
+
+// wispDepShape is the structural contract the repair must satisfy: what a
+// query planner and a write path can actually observe. Column ORDER and plain
+// index NAMES are deliberately excluded -- see the comment at the top of the
+// test for why both legitimately differ on a healed legacy store.
+type wispDepShape struct {
+	columns     map[string]string // name -> "type|nullable|default"
+	primaryKey  string            // ordered PK columns
+	uniqueKeys  map[string]string // name -> ordered columns
+	foreignKeys map[string]string // name -> "cols -> reftable(refcols)"
+	checks      map[string]bool
+	indexCover  map[string]bool // ordered column list of every index, name-independent
+}
+
+func readWispDepShape(t *testing.T, ctx context.Context, conn *sql.Conn) wispDepShape {
+	t.Helper()
+	shape := wispDepShape{
+		columns:     map[string]string{},
+		uniqueKeys:  map[string]string{},
+		foreignKeys: map[string]string{},
+		checks:      map[string]bool{},
+		indexCover:  map[string]bool{},
+	}
+
+	queryEach(t, ctx, conn, `
+SELECT COLUMN_NAME, COLUMN_TYPE, IS_NULLABLE, COALESCE(COLUMN_DEFAULT, '<none>')
+FROM INFORMATION_SCHEMA.COLUMNS
+WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'wisp_dependencies'`,
+		4, func(v []string) { shape.columns[v[0]] = v[1] + "|" + v[2] + "|" + v[3] })
+
+	// Ordered column lists per index. GROUP_CONCAT over SEQ_IN_INDEX keeps
+	// (type, depends_on_issue_id) distinct from (depends_on_issue_id, type).
+	queryEach(t, ctx, conn, `
+SELECT INDEX_NAME, NON_UNIQUE, GROUP_CONCAT(COLUMN_NAME ORDER BY SEQ_IN_INDEX)
+FROM INFORMATION_SCHEMA.STATISTICS
+WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'wisp_dependencies'
+GROUP BY INDEX_NAME, NON_UNIQUE`,
+		3, func(v []string) {
+			switch {
+			case v[0] == "PRIMARY":
+				shape.primaryKey = v[2]
+			case v[1] == "0":
+				shape.uniqueKeys[v[0]] = v[2]
+			default:
+				shape.indexCover[v[2]] = true
+			}
+		})
+
+	queryEach(t, ctx, conn, `
+SELECT rc.CONSTRAINT_NAME,
+       GROUP_CONCAT(kcu.COLUMN_NAME ORDER BY kcu.ORDINAL_POSITION),
+       rc.REFERENCED_TABLE_NAME,
+       GROUP_CONCAT(kcu.REFERENCED_COLUMN_NAME ORDER BY kcu.ORDINAL_POSITION),
+       rc.DELETE_RULE, rc.UPDATE_RULE
+FROM INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS rc
+JOIN INFORMATION_SCHEMA.KEY_COLUMN_USAGE kcu
+  ON kcu.CONSTRAINT_SCHEMA = rc.CONSTRAINT_SCHEMA
+ AND kcu.CONSTRAINT_NAME = rc.CONSTRAINT_NAME
+WHERE rc.CONSTRAINT_SCHEMA = DATABASE() AND rc.TABLE_NAME = 'wisp_dependencies'
+GROUP BY rc.CONSTRAINT_NAME, rc.REFERENCED_TABLE_NAME, rc.DELETE_RULE, rc.UPDATE_RULE`,
+		6, func(v []string) {
+			shape.foreignKeys[v[0]] = v[1] + " -> " + v[2] + "(" + v[3] + ") ON DELETE " + v[4] + " ON UPDATE " + v[5]
+		})
+
+	queryEach(t, ctx, conn, `
+SELECT CONSTRAINT_NAME FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS
+WHERE TABLE_SCHEMA = DATABASE() AND TABLE_NAME = 'wisp_dependencies'
+  AND CONSTRAINT_TYPE = 'CHECK'`,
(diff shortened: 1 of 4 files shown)


steveyegge:
**Fable adversarial review (bee, beads-lane steward) — VERDICT: MERGE-AFTER-FIXES (one blocker).**

The approach is right: a Go repair keyed to pending v58 is the only mechanism that reaches these stores, the SHOW CREATE shape detection is justified, and the FK-before-generated-column restore order has shipped precedent (`ignored/0003` does exactly this sequence). Refuted for the record: FK-add on a PK-less table is fine, the backtick-prefix column match can't confuse `depends_on_id` with `depends_on_issue_id`, fresh stores no-op correctly, no data loss on the clean path, and **no conflict or ordering dependency with #5382** (disjoint files, disjoint concerns).

**F1 — BLOCKER: the repair adds FKs while skipping 0058's orphan cleanup — and then permanently disables it.** 0058's own header runs two guarded DELETEs immediately before its ADD CONSTRAINTs, citing the #4534 class verbatim: the delegate path's unconstrained window leaves `depends_on_wisp_id`/`depends_on_issue_id` pointing at purged wisps/deleted issues. The repair runs plain `ADD CONSTRAINT` with no cleanup and no `SET FOREIGN_KEY_CHECKS = 0` (0058 and ignored/0003 both wrap). On a drifted store with one dangling row — precisely the population this PR exists to heal — either (a) the ADD validates and aborts AFTER `DROP INDEX`/`DROP PRIMARY KEY`/`DROP COLUMN`, leaving a mangled keyless table strictly worse than today's stranding, or (b) the FK lands over orphans and 0058's `@needs_fk_* = 0` guards skip its DELETEs forever — the exact write-brick 0058 was engineered to prevent. Fix: port 0058's two guarded orphan DELETEs into the repair ahead of the ADDs.

**F2 — MAJOR: not re-entrant across mid-rebuild crashes**, violating this file's own doctrine (`ensureDependenciesIDColumn`: a killed process 'needs the NEXT open's repair call to finish the job rather than short-circuit'). Crash after `DROP COLUMN` → next open sees no stored-generated column → repair no-ops → 0058 completes over a PK-less table with the PK/index silently never restored and `ignored/0005` no-oping. Crash after the ADDs but before `ADD PRIMARY KEY` → later `ignored/0005` drops the new FKs and runs its unguarded `DROP PRIMARY KEY` against a PK-less table → the ignored pass wedges. The guards need to resume, not short-circuit.

**F3 — MAJOR: no real-Dolt test.** The added tests are sqlmock statement-echo; every load-bearing fact here (error 1105 shape, ADD CONSTRAINT validation over orphans, generated-column re-add over FK'd base columns) is Dolt semantics sqlmock cannot exercise, and the repo has the harness for exactly this class (`pr4107_corruption_test.go`). Please add one real-Dolt rebuild test on the #5380 DDL with an orphan row seeded — it would have caught F1.

Minor, same pass recommended: unguarded `DROP INDEX` and no `@has_wisps`/`@has_issues` referenced-table checks (0058 carries both); the restored column drops `NOT NULL`, which on a store missing `ck_wisp_dep_one_target` with a zero-target row fails the PK add after the FK adds. Nit: the `strings.Contains(strings.Join(want,…))` assertion inspects the test's own literal slice — pure theater.

CI: uniformly red but infra-shaped (every job incl. trivial ones hung 45m-2h, logs expired); zero green evidence on record either way — a fresh run is mandatory regardless. Happy to re-review promptly once F1-F3 are addressed.

sjarmak:
I took F1-F3 and went a long way with them, and I want to report where it landed rather than push a fix I do not trust. F1 and F3 both held up under everything below. F1 is the finding that started this.

I implemented all three, then ran three adversarial review rounds against my own work. Round 1 found two blockers and four majors in the fix. Round 2, after those were addressed, found two more blockers, both created by the round-1 fixes. That pattern plus four Dolt behaviours convinced me the in-place approach was wrong, so I rewrote it as a build-and-swap. Round 3 blocks that too, with four blockers of its own.

I am not asking you to review code. I am asking two questions at the end that only you can settle.

## Four Dolt behaviours, measured on 2.2.0 and confirmed against Dolt/GMS source

These are worth having on the record regardless of what we do next.

**DDL is not transactional across statements.** Each DDL statement implicitly commits, so `START TRANSACTION`, `ALTER`s, `ROLLBACK` leaves the table changed. A destructive rebuild window cannot be closed by wrapping it.

**A foreign key on the base column of a stored generated column is rejected at `CREATE TABLE` time too, not only on `ADD`.** The restriction is asymmetric: key-then-generated-column is accepted, generated-column-then-key is not. That asymmetry is the only reason any repair here is possible, and it forces the order of any rebuild.

**Foreign key constraint names are database-scoped.** Dolt keeps a root-wide collection and duplicate-name lookup ignores the declaring table, so two tables cannot hold the same constraint name even briefly.

**The primary key is the only `issue_id`-leading index on the shape we target, and `fk_wisp_dep_issue` holds it hostage:**

```
Error 1553 (HY000): can't drop index 'PRIMARY': needed in foreign key constraint fk_wisp_dep_issue
```

`ignored/0003` indexes only the three target columns and restores `fk_wisp_dep_issue` last, so nothing else can support it.

## Why in-place failed

The rebuild has to drop the primary key, which leaves a keyless `wisp_dependencies`. Both `ignored/0003` and `ignored/0005` have keyless windows of their own, and `0005`'s is indistinguishable from the repair's by anything in the store. Resuming into the wrong one records 58 over a table on its way to the id-keyed shape and permanently drops `fk_wisp_dep_issue`. Disambiguating on the ignored cursor does not work either, because `cursorContradictedBySchema` zeroes the cursor when a sentinel table is missing, so the dangerous value reads as the safe one.

While the key is absent, writes are unconstrained. Two rows identical in every column have no deterministic survivor, so no `DELETE` can choose one and `ADD PRIMARY KEY` then fails on every subsequent open. Ordinary retry inserts reach that state, since `created_at` has second resolution and `created_by` and `type` commonly take defaults.

## Why build-and-swap also failed

Build a correct replacement alongside, copy, atomically rename, drop the old one. The rename mechanism holds up: GMS iterates the pairs but the statement transaction commits only after the iterator closes, so a crash between pairs should not persist a half-swap. The surrounding logic is where it breaks.

The copy is a snapshot. Rows written between the copy and the rename are dropped with the outgoing table, updates go stale, and deletes come back. The named migration lock only serializes migrators; application DML never takes it.

Freeing the two database-scoped constraint names has to happen on the live table first. On a half-repaired table, where one target key already exists, the replacement cannot declare that name and `CREATE TABLE` fails after the live source key and CHECK are already gone. Every retry repeats it.

A row with two targets set passes the copy's filter and is then rejected by the replacement's `ck_wisp_dep_one_target`, with the live table already weakened. `0058` anticipates that state and normalizes it; my copy does not.

## The two questions

**Does `wisp_dependencies` still carry its `id` column on the databases 0058 targets?** I cannot settle this from the source. No migration drops `id` from this table, and `ignored/0003` drops only the primary key, which suggests `id` survives. But `ignored/0005` does `ADD COLUMN id ... PRIMARY KEY FIRST`, which only makes sense if `id` is absent when it runs. If both lineages exist in the wild, any rebuild has to carry whatever columns it finds rather than a declared list, and that changes the shape of the fix.

**Must a repair preserve `0058`'s multi-target normalization?** `0058` collapses a row with several targets to one using external > wisp > issue, citing the delegate path's unconstrained window. A rebuild that instead rejects those rows is simpler and louder, but it strands exactly the databases the migration exists to heal. If normalization is required, it belongs in the copy, and I would rather match your precedence than invent one.

## What I would like from you

A steer on approach before I spend a fourth round. If you would rather this were a new migration, or handled by making `ignored/0005` reachable on these databases so the generated column goes away and the keys become addable normally, say so and I will take that route instead.

Nothing here is pushed. The branch still holds the original fix; none of the three rewrites are on it.

steveyegge:
**Answers to both questions, and a steer (bee, beads-lane steward).**

First: this is exactly the report I want to receive — the four measured Dolt behaviours are durable knowledge regardless of what ships here, and "not pushing a fix I don't trust" was the right call. Both questions have definitive answers in the source history.

## Q1 — On the population this PR targets, `wisp_dependencies` categorically has **no `id` column**

The two lineages you suspected do both exist in the wild — but they are disjoint, and your SHOW CREATE detection selects exactly the id-less one:

- The **original 0021** (44a386884, before the #3991 rewrite) created the legacy shape: composite `PRIMARY KEY (issue_id, depends_on_id)`, **no `id`**. The id-keyed CREATE you see in today's 0021 is the rewritten fresh-store shape — and a fresh store never has the stored-generated column, so it is outside your detection.
- The **only writer of the STORED generated `depends_on_id`** is ignored/0003's inline-split branch, whose end shape re-adds the composite PK and carries no `id`. Generated-column-present ⇒ born legacy, 0003 ran, 0005 has not.
- **ignored/0005 is the only migration that adds `id` to a legacy store**, and it drops the generated column earlier in the same `@needs_drop`-guarded run. 0005-completed ⇒ id present, generated column gone. 0005-unreached ⇒ generated column present, no id. There is no state with both.

So your reading of 0005 was right, and the "no migration drops `id`" premise was inverted: nothing drops it because the legacy lineage was born without it. `ADD COLUMN id` is correct; no carry-what-you-find flexibility is needed. (The one state with *neither* — a crash between a `DROP COLUMN` and the id add, keyless with no generated column — is invisible to your current detection; the steer below heals it for free.)

## Q2 — Yes: preserve 0058's multi-target normalization, with 0058's precedence (external > wisp > issue)

Two reasons, one anchor. (a) Multi-target rows are the *signature* of the unconstrained window — rejecting them strands exactly the stores the repair exists to heal. (b) Matching 0058's precedence keeps (repair → 0058) ≡ (0058 alone) on every population, which is the invariant that makes the repair auditable. The precedence itself is not mine to invent or yours: it is fixed by the delegate backfill's statement order (`wispDependenciesSplitTargetBackfillSQL`) and pinned by `TestWispDependenciesSplitTargetBackfillPrefersWispOverIssueThroughDoltCLI`. Zero-target rows: delete outright, as 0058 does.

## The steer: stop restoring the 0003 shape — march the table **forward to the final shape**, by replaying ignored/0005's sequence inside your Go repair

Your two dead ends share one root cause: both tried to *end on* the composite-PK generated-column shape — precisely the shape whose PK cannot be rebuilt safely and whose keyless window aliases 0005's. Aim at the final (current-0021) shape instead, and each measured obstacle dissolves:

1. **Error 1553** — 0005 already solved the PK-hostage problem: it drops `fk_wisp_dep_issue` (guarded) *before* `DROP PRIMARY KEY` and re-adds it last. Replay that order.
2. **No deterministic survivor** — add `id` FIRST (0005's own step), *then* dedup: every row now has a distinct UUID, so a null-safe self-join delete keeping `MIN(id)` per duplicate group is deterministic. Only then add the three `uk_*` unique keys. The final shape's PK is the UUID — always addable, so the fatal `ADD PRIMARY KEY`-on-duplicates state cannot arise.
3. **The 0005-window aliasing** stops mattering, because the repair's destination IS 0005's destination. Every intermediate state has exactly one forward action, decidable by schema inspection (generated column present? `id` present? `uk_*` present? which FKs present?). Which actor resumes a crashed window — 0005 or the repair — is irrelevant when both converge on the same shape. This is also the F2 fix, stated positively: **guard each step on the actual state, never on one precomputed flag** — both 0003 and 0005 have the re-run-no-ops-after-crash defect (their per-statement guards all key on a single upfront sniff), and a repair that inherits it is the thing F2 blocks. As a bonus, state-keyed resume also heals 0005's *own* orphaned crash states, which nothing today can reach.
4. **Concurrent DML in the keyless window** — dedup-after-id makes duplicate inserts recoverable, and a row inserted between cleanup and a constraint ADD aborts that ADD; the next open's resume re-cleans and retries. Convergence-on-retry is the same property 0058 already relies on.
5. **Sequence**: (i) 0058's guarded orphan DELETEs + zero-target DELETE + multi-target normalization FIRST (F1, unchanged); (ii) 0005's guarded drops — `idx_wisp_dep_type_target`, each of the three FKs *if present*, PK, generated column; (iii) `ADD COLUMN id … DEFAULT (UUID()) PRIMARY KEY FIRST`; (iv) dedup; (v) `uk_*` + `idx_wisp_dep_type_*`; (vi) ADD all four constraints **unconditionally** — unlike 0005's restore-only-if-present, since their absence is the disease being cured; all under `FOREIGN_KEY_CHECKS = 0`. End state = today's 0021 CREATE, byte for byte. 0058 then no-ops on every guard, and ignored/0005's `@needs_drop` reads 0.
6. **Vehicle**: keep the Go repair keyed to pending v58 — a new numbered migration cannot work (0058 fails with 1105 *before* any later migration is reachable, so the repair must run pre-0058, which only the repair hook can do). Your alternative of making ignored/0005 reachable is viable but strictly worse: it means cursor surgery on exactly the `cursorContradictedBySchema` machinery you showed reads the dangerous value as the safe one, and 0005's non-resumability and restore-only-if-present FK behaviour both survive the trip — you would still need 0058 to finish the job and a separate answer for crash states.

F3 stands as written: one real-Dolt test on the #5380 DDL — seed an FK orphan, a duplicate pair, and a multi-target row; assert the final `SHOW CREATE` matches a fresh store's and that 0058 then completes. With the forward-march design that single test exercises F1, Q1, and Q2 at once.

One ask: the four measured behaviours (per-statement DDL commit, the generated-column FK asymmetry, database-scoped constraint names, the 1553 index-hostage rule) deserve to outlive this thread — a short section in the migrations README on this PR would do it.

No fourth adversarial round needed on the old designs — they were correctly killed. Looking forward to the forward-march version.

sjarmak:
**Forward-march version is implemented and tested.** Branch `fix/0058-forward-march` on the fork, `50d6afc250447d115c02180e9f78e2cd65be874f`, based on current `main` rather than this PR's stale base. I have not force-pushed it onto this PR's branch; repoint or supersede as you prefer.

F1, F2 and F3 are all addressed. Two places where I did something other than what the steer said, both deliberate.

## The steer's ordering has a bug, and it aborts on the target population

Step (i) runs 0058's cleanup, multi-target normalization included, before the drops. That cannot work while the generated column is still there. `depends_on_id` is `COALESCE(issue, wisp, external)`, so nulling a higher-precedence column recomputes it, and the legacy primary key is `(issue_id, depends_on_id)`. Normalizing a row carrying both an external and a wisp target rewrites its own key onto an external-only sibling's:

```
duplicate primary key given: [w2,external:e1]
```

Measured on real Dolt with the production DDL, not reasoned about. So the deletes stay in step 1, where removing a row cannot collide, and normalization moves after `ADD COLUMN id`. Dedup then follows normalization rather than preceding it, because normalization is itself a source of duplicates: a normalized row can land on a sibling's natural identity, which is exactly the pair above.

0058 carries the same defect. Its normalization is gated on `ck_wisp_dep_one_target` being absent, and any store reaching that branch with a colliding pair hits the same abort.

This is also the sharpest argument for F3 I can offer. A sqlmock test asserting statement order would have passed the broken version.

## FOREIGN_KEY_CHECKS is left on

The cleanup removes every row a constraint in the final step could reject, so suppression buys nothing on the clean path. On a dirty one it is worse than nothing: a foreign key added over surviving orphans is the #4534 write-brick you cited in F1, where one legacy orphan bricks every subsequent create.

With checks on, a missed orphan aborts its own `ADD` instead, and under forward-march ordering that is recoverable. The drops are already done, so the table sits in the final shape minus one constraint, which is a validly keyed table rather than the mangled keyless one F1(a) describes. The next open re-cleans and retries, and 0058 itself can finish the job, since the generated column is gone by then. It is also a per-session variable and the repair runs over a pool.

## A third orphan delete

0058 cleans two target columns. This repair adds all four constraints unconditionally, `fk_wisp_dep_issue` included, and a legacy store can have accumulated `issue_id` orphans during the window that key was absent. Deleting the row is what `ON DELETE CASCADE` would have done had the constraint been in force, matching 0058's own reasoning for its two.

## Test

`internal/storage/embeddeddolt/migrate_wisp_dep_forward_repair_test.go`, real Dolt, three populations: the drifted delegate shape seeded with three FK orphans, a multi-target row and the duplicate its normalization manufactures; the production shape with the check constraint present and the table empty; and a fresh store that must be left untouched.

Mutation-checked rather than trusted green. Each of these turns it red: unregistering the repair, removing the `fk_wisp_dep_issue` orphan delete, moving normalization back before the drop, removing the dedup step.

```
go test ./internal/storage/schema/                   ok
go test -tags cgo ./internal/storage/embeddeddolt/   ok  202s
go vet                                               clean
scripts/check-migration-hygiene.sh                   Migration hygiene OK
```

**One change to F3's assertion.** You asked the final `SHOW CREATE` to match a fresh store's. It cannot, and not because of anything this repair does. Nothing in the chain reorders columns, so a legacy table healed by the shipped path keeps its original order. The already-migrated production database this work came from has `wisp_dependencies` as `(id, issue_id, type, created_at, created_by, metadata, thread_id, depends_on_issue_id, depends_on_wisp_id, depends_on_external)`, where a fresh store puts the three target columns third through fifth. Legacy secondary index names survive for a related reason: a fresh store names its foreign-key backing indexes after the constraints because no index exists to reuse, while a legacy table already has `idx_wisp_dep_*_target` over those columns and Dolt reuses them. Text equality would fail every correctly healed store, including ones this repair never touches. The test asserts structure instead: columns with types and defaults, primary key, unique keys, foreign keys with referenced tables and actions, checks, and index coverage by column list. Index names are not asserted; coverage is.

## Field evidence

Measured against a live 25-database Dolt server, not a lab:

* 14 databases still carry the stored generated `depends_on_id`. Every one of them has no `id` column, which confirms Q1 on real data with zero counterexamples.
* 8 of those 14 have zero target foreign keys, the delegate-path population. The other 6 carry both the generated column and the keys, consistent with 0003's inline split adding constraints before the column existed.
* All 8 have zero rows in `wisp_dependencies`.

End to end, against an isolated copy of one of the affected production stores:

| | before | after |
|---|---|---|
| schema version | 53 | 63 |
| `provenance_events` | absent | created |
| `depends_on_id` | present | gone |
| `wisp_dependencies` foreign keys | 0 | 3 |
| issues / events / comments / dependencies rows | 1126 / 6261 / 472 / 1017 | identical |

Two more things worth having on the record. **Dolt 2.2.3 refuses this identically to 2.1.2**, so upgrading is not an escape hatch. And `dolt sql -q "USE db; <stmt>"` **exits 0 when the statement fails**, so the schema change is refused while the exit code reports success. That one cost me a wrong conclusion mid-investigation before I started reading `INFORMATION_SCHEMA` instead. It belongs next to the CLI's silent no-op on `PREPARE`-driven ALTERs, since both make the CLI unsafe as a migration harness in the same way.

Both are in the new `internal/storage/schema/migrations/README.md`, along with your four behaviours, which is the section you asked for.

steveyegge:
## Review verdict on fix/0058-forward-march @ 50d6afc25: ACCEPTED — repoint this PR to it

Adversarially reviewed, and independently re-verified rather than trusted (bee, beads-lane steward).

**Your ordering correction is confirmed, and my steer was wrong.** I didn't take the measured abort on faith — I checked out 50d6afc25, applied exactly the mutation my steer asked for (normalization before the drop), and ran your suite against real Dolt:

    --- FAIL: .../delegate-path_store_carrying_rows_the_final_shape_rejects
        normalizing multi-target wisp_dependencies rows for the 0058 repair:
        Error 1062: duplicate primary key given: [w2,external:e1]

Exactly the failure you reported, caught by exactly the regression pair you seeded for it. The steer's step (i) was written against 0058's statement inventory without accounting for the generated column recomputing under the legacy composite key — the defect you correctly note 0058 itself carries behind its ck_wisp_dep_one_target gate. Normalize-after-drop, dedup-after-normalize is the right sequence, and your point about sqlmock stands: an order-asserting mock test would have blessed the broken version.

**Everything else verified on my clone:**

- Your suite green against real Dolt: all three populations pass (BEADS_TEST_EMBEDDED_DOLT=1, -tags cgo).
- **Full `./internal/storage/embeddeddolt/` suite at 50d6afc25: ok, 87s, zero failures.** (An earlier attempt hit a 10-minute go-test timeout on my box; that reproduced as ambient — a goroutine parked in dolt's own binlog_primary_streamer, engine-internal, no test in this package — and a clean re-run passed even under heavy load.)
- `./internal/storage/schema/` green; go vet clean; check-migration-hygiene.sh OK.
- wispDepFinalConstraints/wispDepFinalKeys match the fresh-store 0021 DDL definition-for-definition (references, cascade actions, key column lists).
- Merge-clean against main as of my check (the train has since advanced; see process note below).

**The three departures are all accepted on their merits:**

- **FOREIGN_KEY_CHECKS left on** — the recoverability argument is decisive: under forward-march ordering a missed orphan aborts its own ADD over a validly keyed table and the next open retries, whereas suppression converts that same orphan into the #4534 write-brick. The pool-scoped-session-variable point alone would have settled it.
- **Third orphan delete** (fk_wisp_dep_issue) — correct and necessary; the repair adds a constraint 0058 doesn't, so it must clean what 0058 doesn't. CASCADE-parity is the right justification and your test seeds one orphan per constraint, including this one.
- **Structural rather than textual shape assertion** — your column-order and index-name-reuse evidence from the production store proves text equality would fail correctly healed stores. Index coverage asserted, names not, is the right line.

The README section is exactly what I asked for and better — the `dolt sql -q "USE db; <stmt>"` exit-0-on-failure finding belongs there and is worth its own upstream Dolt issue if one doesn't exist.

**One nit, fix on the repoint:** the doc comment on deleteWispDepRowsRejectedByFinalShape still opens with the stale name cleanWispDepRowsRejectedByFinalShape.

**Process:** repoint rather than supersede — force-push fix/0058-forward-march onto this PR's branch so the lineage Q&A, the steer, and this verdict stay on one thread. One timing note: main is momentarily build-broken in backend/conformance by a dispatch skew (repair on #5464, merging on green) — rebase onto main *after* that lands so your CI run means something. I'll approve the fork CI run if the gate holds it, and merge on green.

sjarmak:
**Repointed to the forward-march version.** This branch now carries `386779418`, which is `50d6afc25` rebased onto current `main` with the doc-comment nit fixed. The lineage Q&A, the steer and the verdict stay on this thread as you asked.

`#5464` merged at 07:31Z, three minutes after your verdict, so the rebase you gated on is done rather than pending. Base is `1a1431be3`.

Gates re-run after the rebase, not carried over from the pre-rebase run:

```
go build ./...                                        ok
go vet ./internal/storage/...                         clean
go test ./internal/storage/schema/                    ok   11.8s
scripts/check-migration-hygiene.sh                    OK
TestEmbeddedPre0058RepairMarchesDriftedWispDependenciesToFinalShape
  delegate-path store carrying rows the final shape rejects   PASS
  production shape: check constraint present, table empty     PASS
  a fresh store is left alone                                 PASS
                                     (real Dolt, -tags cgo, 4.5s)
```

Main moved a long way under this branch since the original run, including a new `0064` and a large storage-layer change, so the rebase was worth re-verifying rather than assuming. It applied without conflicts.

One thing for anyone re-running the hygiene check on a fork clone: it falls back to `origin/main` when `BASE_SHA` is unset, which on a fork is whatever that fork last synced, and a stale one reports eight frozen-migration violations against files the branch never touched. CI passes the real base so the gate itself is fine. Running it locally needs `BASE_SHA=$(git rev-parse upstream/main)`.

steveyegge:
**Red at head `386779418` is real but trivial — three mechanical fixes and this merges.** The rollup's failures all trace to `internal/storage/schema/wisp_dep_forward_repair.go`, introduced with the doc-comment fixup on the repoint:

1. **gofmt**: the file fails the formatting check (`Check formatting` + `PR Lint` both flag exactly this file) — run `gofmt -w internal/storage/schema/wisp_dep_forward_repair.go`.
2. **misspell** (`Lint`): CI enforces US spellings — `:96:56` `behaviour` → `behavior`; `:108:41` and `:310:38` `analogue` → `analog`.

Everything substantive is green at this head: conformance (all three populations), migration harness, cross-version smoke, regression, storage/uow. The ACCEPTED verdict stands unchanged — fix the three lines, push, and I'll merge on the green rollup.

PR #5382, opened 2026-08-06
fix(doctor): preflight cached migration history shape

sjarmak:
## What this fixes

`bd doctor`'s migration content-skew check reads `schema_migrations` AS OF the cached remote-tracking ref. A cached ref can legitimately predate that table, or predate its `content_hash` column, and today the code discovers that by letting the `SELECT` fail. On a Dolt server that means a warning logged on every read-only `bd doctor` run, for a condition that is expected rather than wrong.

## The change

`ReadMigrationContentHashes` probes the historical shape before selecting from it, using `SHOW TABLES AS OF` and `SHOW COLUMNS AS OF`, and returns cleanly when the ref predates either. `checkMigrationContentSkew` reports that as a normal not-applicable result instead of a warning.

Both callers reach the ref through `ReadMigrationContentHashes`, so both now issue the probes. That includes the TOCTOU re-check `attemptFastForward` performs via `remoteMaxAtRef`, which is why `expectRemoteMaxReread` mocks them alongside `expectSmartRemoteReadForRemote`. Missing that second helper is what made `TestSmartGateRoutingFastForward` fail.

## Test plan

```
go build ./...                       ok
go vet ./internal/storage/schema/ ./cmd/bd/doctor/    clean
go test ./internal/storage/schema/   ok  37.1s
```

New coverage in `cmd/bd/doctor/migration_skew_test.go` for the no-cached-ref and predates-content-hashes paths. `TestCheckMigrationContentSkew_*` all pass; the two real-Dolt cases skip without a server.

One thing to flag rather than bury: `cmd/bd/doctor` has two failing tests on my machine, `TestCheckBeadsRole_NotConfigured` and `TestCheckBeadsRole_NotGitRepo`. They are not from this change. They fail identically on a branch that does not touch `cmd/bd/doctor` at all, and the cause is environmental: `role_test.go` reads real git config, and a machine with `git config --global beads.role` set makes both expectations wrong. Same two tests, same messages, with and without this patch. Happy to open a separate issue for the test isolation if that is useful.


steveyegge:
**Fable adversarial review (bee, beads-lane steward) — VERDICT: MERGE once Dolt-backed CI executes green; no code changes demanded.**

Statically the PR survives adversarial scrutiny: both probe queries are grammar-valid against the vendored vitess grammar (`SHOW ... AS OF ... LIKE` in exactly this clause order), a missing ref resolves through the same `resolveAsOf` path as the already-validated `SELECT ... AS OF` so the `RemoteRefUnavailableErr` phrasing contract is preserved, the synthesized error strings match the classifiers exactly, the smart gate's routing is unchanged (any read error → `smartUndetermined`), the doctor message split is an accuracy improvement, and **#5381 does not overlap** (orthogonal write-side repair, no shared hunks).

**The gate: this SQL has never executed against a real Dolt engine.** Your local real-Dolt cases skipped ('without a server'), and the CI run's red is infra-shaped — hung trivial jobs (1h41m-3h47m runner outage), `Build (Embedded Dolt)` / `storage domain + uow` CANCELLED, and the CI Gate failure is just the summarizer reacting — so every Dolt-backed shard skipped. This repo has no other use of `SHOW ... AS OF`; the existing `TestCheckMigrationContentSkew_RealDolt` will exercise both probes end-to-end on a real run. I'm refreshing the branch now to trigger fresh CI on a fresh merge snapshot (a plain rerun would reuse the original snapshot).

Non-blocking nits, fine as follow-ups: the missing-table branch (`hasTable == false` → synthesized 'table not found') has no mock case — one sqlmock row would prove it; the uncached-ref state still fires one failing probe per doctor run (honestly scoped in your description); the probe→SELECT TOCTOU is vanishingly rare and degrades safely.

Will merge when the Embedded-Dolt-backed jobs are green.

PR #5383, opened 2026-08-06
/{cmd,internal}: hydrate list/ready dependencies via a page semi-join

coffeegoddd:
## Problem

Hydrating dependencies for a page of issues built an `IN (...)` clause out of the page's issue ids. The statement grew with the page, and the whole id list had to be materialized in Go and sent to the server on every `bd list` / `bd ready`.

## Change

Join the dependency tables against the page's *own* id subquery instead. The filter is expressed once, in SQL, and the statement size no longer depends on how many issues came back — no id list crosses the wire at all.

- `buildPageIDSubquery` (`issue_search.go`) mirrors `searchAcrossIssuesAndWisps`' branching for the `IssueFilter` side.
- `buildReadyWorkIDSubquery` is split out of `getReadyWorkIDPage` for the `WorkFilter` side.
- Both err deliberately toward a **superset** of the page: a joined row the page later drops is harmless, a missing one loses data from the render. The wisps-empty/missing probe still gates the union — a `UNION` against a nonexistent `wisps` table errors outright, so that branch is required, not an optimization.
- Dependency rows now order by `issue_id, depends_on_id, type`. `issue_id` alone is not a total order when one source has several edges.

The page and its edges come back from a single use-case call — `ListWithDependencies` / `GetReadyWorkWithDependencies` — so the two halves cannot drift apart at the call site. `bd ready`'s parent-epic map reuses the edges the caller already loaded rather than re-reading them, which removes a second full hydration outright.

The watch loop and the hierarchical `--parent` tree keep the id-list path: neither page comes from a single filtered query, so no subquery reproduces them. `ListByIssueIDs` / `GetForIssueIDs` are unchanged and still serve those callers.

## Verification

- `go build ./...`, `go vet ./internal/storage/... ./cmd/bd/...` clean.
- `go test ./internal/storage/domain/...` and `go test ./cmd/bd/... ./internal/storage/...` pass.
- Behavioral diff against `main`: two binaries run side by side on a seeded workspace (8 issues, 3 `blocks` edges, 1 `parent-child`, 1 label). **17 command variants byte-identical** — plain / `--limit` / `--offset` / `--sort --reverse` / `--json` / `--deps full` / `--format` / `--tree` / `--label` / `--status --priority` / `--parent` / `--ready`, plus `ready`, `ready --json`, `ready explain`, and an explicit `--proxied-server` run.

## Notes for review

- The filter is still evaluated twice per list — once for the page, once inside the dep query's subquery. Collapsing that into one statement was considered and deliberately deferred; it would mean scanning issue columns repeated per edge.
- The highest-risk path is a label-driven filter, where `plan.Distinct` puts a `SELECT DISTINCT id` inside the join. That changes row multiplicity in a way the search's own hydration never had to handle, and it is worth a careful look.
- Not verified by hand: a repo with no `wisps` table (covered only by the existing suite), and a page large enough for the old `IN` clause to actually hurt — the manual diff confirms correctness, not the speedup.

### DIFF
--- a/cmd/bd/list_proxied_server.go
+++ b/cmd/bd/list_proxied_server.go
@@ -65,14 +65,14 @@ func runListProxiedSearch(_ *cobra.Command, ctx context.Context, in listInput) e
 		return emitProxiedListJSONResult(page.Items, in, page.HasMore)
 	}
 
-	page, err := uw.IssueUseCase().SearchIssues(ctx, "", filter)
+	page, deps, err := uw.IssueUseCase().ListWithDependencies(ctx, "", filter)
 	if err != nil {
 		return err
 	}
 
 	issues, hasMore := workapi.FinishPage(page.Items, in.SortBy, in.Reverse, in.effectiveLimit, page.HasMore)
 
-	return renderProxiedListText(ctx, uw, issues, in, hasMore)
+	return renderProxiedListText(ctx, uw, issues, in, hasMore, deps)
 }
 
 func runListProxiedReady(_ *cobra.Command, ctx context.Context, in listInput) error {
@@ -92,14 +92,14 @@ func runListProxiedReady(_ *cobra.Command, ctx context.Context, in listInput) er
 		return emitProxiedListJSONResult(page.Items, in, page.HasMore)
 	}
 
-	page, err := uw.IssueUseCase().GetReadyWork(ctx, wf)
+	page, deps, err := uw.IssueUseCase().GetReadyWorkWithDependencies(ctx, wf)
 	if err != nil {
 		return err
 	}
 
 	issues, hasMore := workapi.FinishPage(page.Items, in.SortBy, in.Reverse, in.effectiveLimit, page.HasMore)
 
-	return renderProxiedListText(ctx, uw, issues, in, hasMore)
+	return renderProxiedListText(ctx, uw, issues, in, hasMore, deps)
 }
 
 func runListProxiedWatch(_ *cobra.Command, ctx context.Context, in listInput) error {
@@ -147,7 +147,7 @@ func runListProxiedWatch(_ *cobra.Command, ctx context.Context, in listInput) er
 			issues, hasMore = workapi.FinishPage(page.Items, in.SortBy, in.Reverse, in.effectiveLimit, page.HasMore)
 		}
 
-		deps, err := loadDepsForIssues(ctx, uw, issues)
+		deps, err := loadDepsForIssues(ctx, uw, issues, nil)
 		if err != nil {
 			return nil, false, nil, err
 		}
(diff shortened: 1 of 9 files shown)


steveyegge:
**Fable adversarial review (bee, beads-lane steward) — VERDICT: MERGE-AFTER-FIXES, where the first fix is a substantive rebase.**

The domain-layer machinery is carefully built and worth keeping: dep-plane semantics exactly match `GetForIssueIDs`, the wisps-missing probe correctly gates the union, all user data rides `?` placeholders with whitelisted sort keys, arg ordering matches placeholder order in every composed statement, and the ready-subquery split is line-faithful to main's `getReadyWorkIDPage`. But:

**1. The PR is architecturally stale.** #5378 landed on main ~2h after your CI completed and rewrote the exact cmd-layer functions this PR wires — `runListProxiedSearch`/`runListProxiedReady` no longer exist; the list path now goes through `issueops.Reader.List` + `BlockingAnnotator`, and `git merge-tree` confirms a real conflict in `cmd/bd/list_proxied_server.go`. The green badge attests to a merge snapshot that no longer exists, and a CONFLICTING PR silently gets zero further CI runs — re-implement through the Reader role, don't just textually resolve.

**2. Unconditional hydration pessimizes the most common invocation.** On your base, plain/--long/agent-mode `bd list` (the scripted default) loaded no deps; `ListWithDependencies` now always hydrates (+1 wisps probe, +2 filter re-evaluations, +2 joins) for a map those renders never read. Restore lazy hydration.

**3. `--sort id` pages have NO SQL ORDER BY at all** — `IsGoSideSort("id")` makes `OrderByForColumns` return empty, so page query and id-subquery are two independent engine-arbitrary `LIMIT n+1` selections that can disagree on a quiet DB, silently dropping deps for rendered rows. Force an id tiebreak inside the subquery whenever a LIMIT is present.

**4. The superset invariant breaks under LIMIT + concurrent writes** — the dep subquery re-executes the filter later with fresh `UTC_TIMESTAMP()` and recomputed ready predicates, so a concurrent mutation shifts the window and a rendered issue's deps go missing. The old id-list path was immune. Transient and non-corrupting, but this deployment is a shared server with ~37 writers. Mitigate, or document-and-accept in a comment as an honest minimum.

**5. Nine files, zero new tests** — nothing pins the superset property, the DISTINCT label-join (your own flagged highest-risk path), or the wisps-missing branch. And since the PR's entire premise is latency and the ready path now evaluates the filter ~3× per command (~6-13 queries replacing 2 page-bounded IN-lists), a measured before/after on a realistic corpus is warranted — the speedup is currently unmeasured.

Minor: `ready explain` now hard-fails where it used to degrade to a partial explanation with a warning; the PR notes undercount the re-evaluation ('twice' → three page-filter evaluations + two wisps probes).

The direction is right and the domain layer is keepable as-is — happy to re-review promptly after the rebase onto the Reader architecture.

coffeegoddd:
@steveyegge thanks for the feedback! I'll go ahead and close this PR and get some fresh benchmarks demonstrating the excess latency I'm seeing, with steps to reproduce it, and I'll file an issue with the report.

PR #5384, opened 2026-08-07
feat(label): add --prefix to bd label remove

athosmartins:
## What

Adds a `--prefix` flag to `bd label remove`: `bd label remove <issue-id...> --prefix <prefix>` removes every label on the given issue(s) that starts with `prefix`, instead of requiring the caller to name each label exactly.

## Why

`bd list`/`bd ready` already support `--label-pattern`/`--label-regex` for *finding* issues by label shape. `label remove` had no equivalent for *stripping* a whole label family in one call (e.g. every `pool:refused:<reason>` label on an issue) — callers had to fetch the issue's current labels themselves, compute which ones matched, and issue one exact-match `bd label remove` per label.

Implemented at both storage paths:
- Embedded: fetches the issue's current labels via `store.GetLabels`, filters with a plain `strings.HasPrefix` helper (`labelsWithPrefix`), removes matches in a single transaction (mirrors the existing exact-match `processBatchLabelOperation`).
- Proxied server (`bd serve`): same shape via `uow.LabelUseCase()`, wisp-aware (`workapi.GetIssueOrWisp`), mirroring the existing `labelMutateProxied`/`runLabelListProxiedServer` structure. This path matters in practice — any deployment running `bd` against a shared Dolt server via `bd serve` needs it, not just the embedded/local case.

No storage-layer primitive was added — this composes the existing `GetLabels`/`RemoveLabel` (and `RemoveWispLabel`) primitives at the `cmd/bd` layer, same as the rest of `label.go`'s existing helpers (`parseLabelArgs`, `resolveLabelIssueIDs`, etc.).

## Verification

```
go build ./...
go vet ./cmd/bd/...
golangci-lint run ./cmd/bd/...   # 0 issues
gofmt -l <changed files>          # clean

BEADS_TEST_EMBEDDED_DOLT=1 go test ./cmd/bd/ -run TestEmbeddedLabel -v
# 24/24 subtests PASS, including the 2 new label_remove_prefix cases —
# full existing label test suite re-run for regressions, all green.

go test ./cmd/bd/ -run TestLabelsWithPrefix -v
# 5/5 PASS (pure-function unit test)
```

One gap to flag: I added a `remove_prefix` subtest to `TestProxiedServerLabel` (`label_proxied_integration_test.go`), mirroring the existing `add_list_remove` case, but could **not** run it locally — this dev environment has no Docker, which the `testcontainers-go` Dolt harness (`requireSharedProxiedServer`) requires. The proxied-mode implementation compiles clean and follows the existing `labelMutateProxied` pattern exactly, but that specific subtest is only proven by CI, not by me locally. Flagging explicitly rather than claiming full local verification.

TDD throughout: each new test was written first and confirmed to fail for the right reason (`undefined: labelsWithPrefix`, then `unknown flag: --prefix`) before the corresponding implementation was added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   count the same metadata-scoped set `bd list` returns without fetching every
   row.
 
+- **`bd label remove --prefix` removes every label matching a prefix in one
+  call**, instead of requiring an exact label name per removal. `bd list`/`bd
+  ready` already supported `--label-pattern`/`--label-regex` for *finding*
+  issues by label shape; `label remove` had no equivalent for *stripping* a
+  whole label family (e.g. every `pool:refused:*` reason label) — callers had
+  to fetch the labels themselves and issue one `label remove` per exact name.
+  `--prefix` resolves the matching labels per issue and removes them in a
+  single transaction; `bd label remove <id> --prefix pool:refused:` replaces
+  that fetch-then-loop. Routed through the same `issueops.Lifecycle`/
+  `issueops.Reader` roles as a plain `label remove`, so it works identically
+  on the embedded and proxied-server (`bd serve`) storage modes.
+
 - **The events journal records WHO performed each mutation.** `bd_events_journal`
   gains an `actor` column (migration 0066 plus its ignored-series twin 0025, so
   upgraded workspaces and fresh clones converge on the same shape), stamped
--- a/cmd/bd/label.go
+++ b/cmd/bd/label.go
@@ -184,6 +184,90 @@ func reportLabelEdit(issueIDs []string, labels []string, operation string, jsonO
 	return nil
 }
 
+// removeLabelsByPrefix removes every label starting with prefix from each
+// issue in issueIDs, through issueops.Lifecycle — the same role
+// applyLabelEdit uses, so the events journal and both storage routes see this
+// edit the same way they see a plain `label remove`. Unlike applyLabelEdit
+// (which applies one fixed, caller-supplied label set to every issue), the
+// label set here is resolved per-issue from that issue's current labels via
+// issueops.Reader, since different issues may carry different prefix-matching
+// labels — so the loop is one Get plus one Update per issue rather than one
+// shared patch for all of them.
+func removeLabelsByPrefix(ctx context.Context, issueIDs []string, prefix string, jsonOut bool) error {
+	reader, err := openIssueReader()
+	if err != nil {
+		return HandleErrorRespectJSON("%v", err)
+	}
+	type removal struct {
+		issueID string
+		label   string
+	}
+	var removals []removal
+	for _, issueID := range issueIDs {
+		details, gerr := reader.Get(ctx, issueops.GetRequest{ID: issueID})
+		if gerr != nil {
+			return HandleErrorRespectJSON("getting labels for %s: %v", issueID, gerr)
+		}
+		for _, label := range labelsWithPrefix(details.Labels, prefix) {
+			removals = append(removals, removal{issueID: issueID, label: label})
+		}
+	}
+	if len(removals) == 0 {
+		if jsonOut {
+			return outputJSON([]map[string]interface{}{})
+		}
+		fmt.Printf("No labels matching prefix '%s' found\n", prefix)
+		return nil
+	}
+
+	lifecycle, err := openIssueLifecycle()
+	if err != nil {
+		return HandleErrorRespectJSON("%v", err)
+	}
+	ctx, err = issueOpsContext(ctx)
+	if err != nil {
+		return HandleErrorRespectJSON("%v", err)
+	}
+	// Grouped per issue (rather than one call per removal) because a single
+	// LabelPatch.Remove already de-duplicates and applies every label for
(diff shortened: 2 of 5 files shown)


steveyegge:
## Review verdict: MERGE AFTER FIXES (small ones)

Adversarially reviewed. The design is sound — composing `GetLabels`/`RemoveLabel` rather than adding a storage primitive is right, both paths write `label_removed` events, the embedded path IS wisp-aware (routing happens at the storage layer — your body undersells it), and the proxied path's read+remove-in-one-tx with `removals[:0]` retry reset is careful work. Your flagged can't-run-locally gap is CI-proven: all 15 proxied label shards green including the new subtest.

Three asks before merge:
1. **Error-string tense:** `"removed label '%s' on %s: %w"` and `"label removed (prefix): %v"` render past-tense success on *failure* — should be "removing".
2. **Reject `--prefix` + a label-shaped positional explicitly.** Today all positionals become issue IDs, so `bd label remove bd-1 stale-label --prefix x` fails with a confusing resolve error — and `ResolvePartialID`'s fuzziness makes silently resolving a label-shaped arg to a real issue at least conceivable. A "cannot combine --prefix with label arguments" error closes it.
3. **A sharp-edge guard or note:** a one-character prefix silently strips whole families in one transaction (`--prefix t` takes `tier:opus` and `test-needed` together). Either `--dry-run`, or at minimum flag-help text noting the blast radius — and that state-dimension labels (`launch:`, `scalegate:`) should move via `bd set-state`, which this flag bypasses without the event-bead audit.

CI note: your three reds are all orthogonal to the diff — the doc-freshness calendar rot was main-side and is already fixed on main (update-branch when you push and they clear), and the storage-domain red is a testcontainers/ryuk flake. The actual CLI-flags drift gate passed, so no doc regen is needed for the new flag.

athosmartins:
Thanks for the review — all three points are addressed in 39fd21c:

1. **Error-string tense.** Added `labelOperationGerund()` (`cmd/bd/label.go`) mapping the past-tense `labelOperationAdded`/`labelOperationRemoved` constants to "adding"/"removing" for failure-path messages only. `applyLabelEdit`'s error branch and `removeLabelsByPrefix`'s error message both now use the gerund — the success-path wording is untouched. `TestLabelOperationGerund` covers the mapping and explicitly asserts the result never ends in "ed".

2. **Reject `--prefix` + a label-shaped positional.** Added `resolveIssueIDsForPrefix`, used only by `labelRemoveCmd`'s `--prefix` branch, which resolves every positional as an issue ID and fails with `"cannot combine --prefix with label arguments: %q is not an issue ID"` the moment one doesn't resolve — before any mutation. New embedded-CLI test `label_remove_prefix_rejects_label_shaped_positional` checks both the rejection message and that the valid issue's labels are left untouched (no partial apply).

3. **Sharp-edge guard.** Took the help-text option: `--prefix`'s flag help now names the blast radius with your own example (`--prefix t` matching both `tier:opus` and `test-needed`) and points at `bd set-state` for state-dimension labels (`launch:`, `scalegate:`), which `--prefix` bypasses without that command's event-bead audit trail. Didn't add `--dry-run` since you flagged help-text as sufficient for a small-scope PR.

Full `cmd/bd/label` suite re-run green (25/25 embedded subtests: 24 pre-existing + the new one), `go build`/`go vet`/`gofmt` clean. CI is green on 39fd21c (`mergeStateStatus: CLEAN`).

Ready for re-review whenever you have a chance.

bee-ghosttrack (approved):
**Fable adversarial review, delta (bee) — VERDICT: APPROVE** at `39fd21caed`, delta `9f39bdb5df..39fd21caed` (+96/−4, 3 files), against steveyegge's 2026-08-08 "MERGE AFTER FIXES" verdict.

All three asks are discharged in the code, not just in the description. (1) Both failure paths now say "adding"/"removing" via `labelOperationGerund` (`cmd/bd/label.go:136-138`, `:267`); the success constants are untouched and `reportLabelEdit` still uses them. (2) The `--prefix` branch of `labelRemoveCmd` (`:423`) resolves through the new `resolveIssueIDsForPrefix`, which returns before any mutation on the first unresolvable positional; the embedded case `label_remove_prefix_rejects_label_shaped_positional` asserts both the message and that the valid issue's label survives. (3) The flag help names the blast radius with the reviewer's own example and routes state-dimension labels to `bd set-state` (`:622`).

Ran in a detached worktree at your head: `go build ./...` rc 0, `go vet ./cmd/bd/` rc 0, `gofmt -l cmd/bd/` empty, `go test -count=1 -run 'TestLabelOperationGerund|TestLabel|TestEmbedded.*[Ll]abel' ./cmd/bd/` ok (148s). CI rollup at this head: 119 SUCCESS, 1 SKIPPED, mergeStateStatus CLEAN.

One nit, not blocking: `resolveIssueIDsForPrefix` still goes through `resolveLabelTarget` → `ResolvePartialID`, so a label-shaped positional that happens to prefix-match a real id would still resolve silently (the fuzziness Steve named in ask 2 is narrowed by the early refusal, not removed), and a plain typo'd issue id now gets the "cannot combine --prefix with label arguments" wording, which will mislead that caller. A message of the form `%q did not resolve as an issue ID (under --prefix every positional is an issue ID; label arguments are not accepted)` covers both readers. Fine to take as a follow-up.

Julian owns the merge.

PR #5385, opened 2026-08-07
Reject "list"/"add" as the id in the singular "comment" shorthand

athosmartins:
## What

`bd comment <id> [text...]` had no validation on the `id` positional
argument. Typing `bd comment list <text>` (confusing the singular
shorthand with the plural `bd comments list` guard) parsed as
`id="list"`, and since no issue is literally named "list",
`ResolvePartialID`'s substring/prefix fallback silently matched "list"
against whichever existing bead/wisp id happened to contain it as a
prefix (after `wisp-` stripping) — posting the comment on the *wrong*
issue with no error or warning.

This adds `validateCommentArgs`, wired as `commentCmd`'s `Args`
validator, rejecting the literal words `"list"` and `"add"` as the id
(mirroring the existing guard on the plural sibling,
`validateCommentsArgs` in `comments.go`) with a message pointing at the
correct invocation. It runs before `RunE`'s dispatch, so both the
embedded and proxied-server backends reject identically.

## Why

This exact typo recurred 10+ times across multiple independent sessions
in one downstream deployment over a single day, each time silently
posting a stray comment onto an unrelated, real issue (evidenced by
that issue's comment history). No error, no warning — `✓ Comment added`
prints either way, so the mistake is only caught by noticing the
returned `issue_id` doesn't match what was intended.

The broader pattern — `ResolvePartialID`'s substring-fallback resolving
to an unintended issue for *any* wrong/typo'd id, on any command that
takes a mutating `<id>` positional, not just `comment` — is a separate,
larger question (worth asking: should write commands require an exact
match rather than fuzzy-resolve at all?). Deliberately **not** addressed
here, per this repo's "one issue per PR" hygiene rule — this PR only
closes the concrete, evidenced footgun in the singular/plural `comment`
shorthand pair, which is the only command pair of this shape in the
codebase (checked: no other singular command has a plural sibling with
real subcommands whose names could collide with its own id slot the
way `comment`/`comments` do).

## Verification

- `go build ./...` — clean.
- `go vet ./cmd/bd/...` — clean.
- `go test ./cmd/bd/... -run TestValidateCommentArgs -v` — all 7
  subtests pass (pure unit test, no store needed).
- `go test -tags integration ./cmd/bd/... -run TestCLI_Comment -v` —
  the DB-independent subset (`*RejectedBeforeStoreOpen`) passes; the
  rest of that suite requires a local Dolt/Docker test container that
  wasn't available in this sandbox, so those cases skip rather than
  run here (should execute under this repo's CI).
- Confirmed no regression: ran the full (non-integration)
  `./cmd/bd/` package suite both on this branch and on unmodified
  `origin/main` in the same sandbox — 4 pre-existing failures
  (`TestCheckHooksDriftNotGitRepo`,
  `TestCollectViperEntriesWithEnvOverride`,
  `TestPrepareSelectedCommandContext_RebindsTargetConfig`,
  `TestPrepareSelectedCommandContext_DoesNotMergeCallerConfigForUnsetKeys`)
  reproduce identically on both, caused by ambient `BD_ACTOR`/
  `BEADS_ACTOR`/git-ceiling env vars in this particular sandbox, not by
  this change. Not added to `.test-skip` since they appear
  environment-specific to this sandbox rather than general.

### DIFF
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -1402,6 +1402,173 @@ func TestCLI_CommentsAddShortID(t *testing.T) {
 	})
 }
 
+// TestCLI_CommentListMisplacedSyntax is a regression test for the singular
+// "comment" shorthand silently mis-resolving a subcommand-like typo as an
+// issue id. "bd comment list <text>" (meant as an attempt to list comments,
+// using the singular noun by mistake) used to parse as id="list",
+// text=[<text>...] with no validation on the id — since no issue is
+// literally named "list", ResolvePartialID's substring/prefix fallback
+// would silently match it against whatever existing bead or wisp id
+// happened to contain "list" and post the comment there instead of
+// erroring. This asserts both halves of the fix: the command is rejected
+// with a helpful hint (mirroring TestCLI_CommentsListMisplacedSyntax for
+// the plural sibling), AND — the part that actually matters for this
+// specific bug — no comment silently landed on an unrelated real issue.
+func TestCLI_CommentListMisplacedSyntax(t *testing.T) {
+	t.Parallel()
+
+	tmpDir := setupCLITestDB(t)
+
+	out := runBDInProcess(t, tmpDir, "create", "Bystander issue", "-p", "1", "--json")
+	jsonStart := strings.Index(out, "{")
+	if jsonStart < 0 {
+		t.Fatalf("No JSON found in create output: %s", out)
+	}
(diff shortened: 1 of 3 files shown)


athosmartins:
Duplicate of #5369 (same branch content, same fix — I opened this before discovering #5369 already existed; my earlier `gh pr list --head` search for that branch returned a false empty result). Closing in favor of #5369, which was opened first and is otherwise identical. Sorry for the noise.

PR #5386, opened 2026-08-07
fix(defer): auto-wake expired dated defers on ready-front reads and claims (bd-i8qx8)

steveyegge:
## What

A dated defer now means what it says: once `defer_until` passes, the next ready-front read or claim returns the issue to open. Today `bd defer --until` / `bd update --defer` set `status=deferred` + `defer_until`, and nothing ever flips the status back — the ready WHERE clause's `defer_until` predicate only helps rows that are already `open`, so an expired defer is invisible to `bd ready` forever until a human runs `bd undefer`.

This is not a corner case: one deployment's census found 241 issues (24 P1s at first count) silently dark this way, regenerating ~11/day from correct-looking automation — "cooldown" call sites that defer `+7d` behind a label fence, where the 7-day snooze silently became permanent invisibility. A P2 security issue sat unseen for 15 days.

## Semantics

- **Dated defer = snooze.** At expiry, the wake sweep writes `status=open, defer_until=NULL` — byte-identical to `bd undefer` (and consistent with `bd reopen`'s documented decision to clear `defer_until`). Clearing the date matters: a dateless `bd defer` leaves `defer_until` untouched, so a kept stale past date would turn a later indefinite icebox into an instant re-wake.
- **Dateless defer = indefinite icebox.** `defer_until IS NULL` never auto-wakes; `bd undefer` stays its only exit. The hold shape already existed in the data model, so no new flag.
- Each wake records a `status_changed` event, actor `bd-defer-wake`.

## How

One shared DBTX body, `issueops.WakeExpiredDefersInTx` (snapshot-then-recheck, shaped after `ReclaimExpiredLeasesInTx`: per-row UPDATE re-asserts the full predicate so a concurrently rescued row is skipped; `row_lock` rewritten), called at the top of every ready-front read and claim:

- **classic dolt server store**: `GetReadyWork` / `GetReadyWorkWithCounts` / `CountReadyWork` + `readyClaimer.ClaimNext`, in a write tx of its own via `runIssueOperationTxWithMessage` (the read paths' `withReadTx` always rolls back, so the sweep cannot live inside them)
- **embedded store**: same three seams + its `readyClaimer`
- **domain/uow stack** (proxied CLI + `bd serve`): `uow.WakeExpiredDefers` in a committing UOW before `Reader.Ready`, `Reader.List(--ready)`, the proxied CLI ready route, and the uow `readyClaimer`

Properties: RowsAffected-gated — the steady state (nothing expired) mints **no dolt commit** (`commitMsg == ""` / zero changed tables); advisory by contract — a ready listing never fails because the sweep could not run (strict `--readonly` skips it, `ErrCircuitOpen`/`ErrStoreClosed` are swallowed, everything else warns); wisps carry the same columns and are swept too, minting no version commit by construction (dolt_ignored).

## Tests

- `TestEmbeddedDeferAutoWake`: expired defer wakes on ready (status open, `defer_until` cleared), dateless defer never wakes, future defer stays hidden, freshly-expired defer is claimable via `bd ready --claim` — 4/4 green.
- `TestProxiedServerDeferAutoWake`: proxied-mode parity subtest — 3/3 green.
- Existing defer/undefer/ready suites green in both modes (`TestEmbeddedDefer`, `TestEmbeddedUndefer*`, `TestEmbeddedReady`, `TestProxiedServerDeferUndefer`, `TestProxiedServerReady*`, `TestProxiedServerReadyWisp`).

CHANGELOG, `bd defer` help, `--defer` flag help, and CLI_REFERENCE updated to state the snooze/icebox contract.

Provenance: filed off a cross-rig audit (wyvern wy-jvw3d census; independent root-cause analysis by a second agent confirmed no consumer relies on defer's accidental permanence — every legitimate "keep hidden" mechanism is a blocker or label fence, so auto-wake un-hides but never un-fences).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   disable the fallback everywhere). Explicit IDs and `bd show --current` are
   unaffected.
 
+- **A dated defer now means what it says: the issue returns to the ready front
+  when the date passes** (bd-i8qx8). `bd defer --until` and `bd update --defer`
+  set `status=deferred` plus a `defer_until` timestamp, but nothing ever
+  flipped the status back — so an expired defer stayed invisible to `bd ready`
+  forever until a human ran `bd undefer`. One deployment measured 241 issues
+  (including P1s) silently dark this way, regenerating daily from
+  correct-looking automation.
+
+  Ready-front reads and claims (`bd ready`, `bd ready --claim`, `bd list
+  --ready`, and the serve/proxied Reader roles) now run a lazy wake sweep
+  first: every issue (and wisp) with `status=deferred` and `defer_until <=
+  now` flips to `status=open, defer_until=NULL` — byte-identical to what `bd
+  undefer` writes, so a later dateless re-defer cannot inherit a stale past
+  date. Each wake records a `status_changed` event (actor `bd-defer-wake`).
+
+  What does NOT change: a **dateless** defer (`bd defer` with no `--until`,
+  `defer_until` NULL) is the indefinite icebox and never auto-wakes — `bd
+  undefer` remains its only exit. The sweep is a no-op when nothing has
+  expired (no dolt commit is minted), advisory by contract (a ready listing
+  never fails because the sweep could not run — strict `--readonly` skips it),
+  and identical in embedded, server, and proxied-server modes.
 
 - **`bd delete --cascade` no longer marks LIVE issues as deleted in other
   issues' text** (bd-x82so). When the deletion named a wisp, the set used to
--- a/cmd/bd/defer.go
+++ b/cmd/bd/defer.go
@@ -26,9 +26,14 @@ be revisited.
 
 Deferred issues don't show in 'bd ready' but remain visible in 'bd list'.
 
+A defer WITH a date is a snooze: once --until passes, the next ready-front
+read returns the issue to open automatically (same shape as 'bd undefer').
+A defer WITHOUT a date is the indefinite icebox: it stays deferred until
+someone runs 'bd undefer'.
+
 Examples:
-  bd defer bd-abc                  # Defer a single issue (status-based)
-  bd defer bd-abc --until=tomorrow # Defer until specific time
+  bd defer bd-abc                  # Icebox indefinitely (until bd undefer)
+  bd defer bd-abc --until=tomorrow # Snooze: auto-wakes once the date passes
   bd defer bd-abc --reason="waiting on API access"
   bd defer bd-abc bd-def           # Defer multiple issues`,
 	Args:          cobra.MinimumNArgs(1),
--- a/cmd/bd/defer_wake_embedded_test.go
+++ b/cmd/bd/defer_wake_embedded_test.go
@@ -1,0 +1,162 @@
+//go:build cgo
+
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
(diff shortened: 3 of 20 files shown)


steveyegge:
**Fable adversarial review verdict: FIX_THEN_MERGE — one blocking defect, empirically reproduced. Do not merge until fixed.**

The permanent-issue fix is sound and well-wired: the sweep sits ahead of every ready-front read and claim in all three stacks (classic dolt, embedded, uow/proxied/serve — including `bd list --ready` and the HTTP ready route), the predicate is timezone-consistent with the existing ready WHERE (UTC_TIMESTAMP matches ready_work.go:234 / sqlbuild/ready.go:136), the snapshot-then-recheck + row_lock shape correctly serializes concurrent wakers (loser matches nothing, no double event), and embedded/proxied tests genuinely run and pass.

**Blocker: the uow stack never persists wisp wakes.** `issueSQLRepositoryImpl.WakeExpiredDefers` (internal/storage/domain/db/issue.go:1162-1169) returns only `len(out.Issues)`, discarding the wisp count; a wisp-only sweep then yields `commitMsg == ""`, `RunTxResult` skips `uw.Commit`, and `Close` → `RollbackUnlessCommitted` rolls the whole transaction back. Reproduced empirically: the same expired-defer wisp wakes to open in embedded mode but stays deferred FOREVER under the proxied server — the exact tomb this PR exists to kill — and every ready read redoes-and-discards the write (perpetual discarded write traffic on the hottest read command). This contradicts the PR's wisp/parity claims. Fix shape: return the wisp count from the repo; in uow.WakeExpiredDefers commit with the wake message when issues woke, else `uw.Commit(ctx, "")` (the RunTxEphemeral SQL-only-commit precedent) when only wisps woke — and land a wisp-wake parity test, since that seam currently has zero coverage.

Should-fix: silence (or warn-once) permission-denied in the dolt store's defer-wake advisory handler — a read-only-privileged client currently gets a warning per `bd ready` and each failure feeds the write circuit breaker from a read command. Nits: classic `bd ready` sweeps twice per invocation (listing + count); the wake's status_changed event carries plain strings where the update funnel writes JSON blobs; sweep SELECT is unbounded (fine at current scale).

steveyegge:
**Blocker fixed in 0a94d44f4** (reviewer follow-through, per the FIX_THEN_MERGE verdict above).

The fix takes the shape the verdict prescribed: `WakeExpiredDefers` now reports `(issues, wisps)` through the repo interface and use case; the uow sweep commits with the wake message when permanent issues woke, and persists a wisp-only wake via `uw.Commit(ctx, "")` — the ephemeral plain-COMMIT form (RunTxEphemeral precedent, bd-lrgn1) — issued inside the work func so a serialization failure still retries against a fresh unit of work. Worth noting for posterity: passing a wake message for wisp-only sweeps would NOT have worked — wisp tables are dolt_ignored, so `DOLT_COMMIT` errors nothing-to-commit and doltServerTx rolls back, discarding the wisp writes anyway. The plain-COMMIT path is load-bearing.

Verification: wisp parity subtests added to both mode tests (`expired_dated_defer_wakes_wisp`). Embedded suite green; proxied suite green — the proxied wisp subtest is the exact repro shape that failed before the fix. `internal/storage/domain/db`, `domain`, `uow`, `dberrors` package tests green locally.

Should-fix also done: `dberrors.IsAccessDenied` (1044/1045/1142/1143/1227 + string fallback) with warn-once handling in both defer-wake advisory handlers, so a read-only-privileged client gets one warning per process instead of one per `bd ready` (the breaker concern turned out to be moot — `isConnectionError` already excludes privilege errors from tripping it).

Note: this PR's CI will stay red on the unrelated repo-wide doc-staleness tripwire until #5394 lands on main; re-run after.

PR #5387, opened 2026-08-07
Put the memory plane behind memoryops, conformance-tested at three backends

julianknutsen:
Puts `bd remember` / `memories` / `forget` / `recall` behind a role, conformance-tested at three
backends, with both CLI routes and four new HTTP operations calling it.

They were the last commands implemented the pre-facade way: two independent bodies —
`cmd/bd/memory.go` reaching `store.GetConfig/SetConfig/DeleteConfig/GetAllConfig`, and
`cmd/bd/memory_proxied_server.go` reaching `uw.ConfigUseCase()` — sharing only presentation. No leaf
role, no conformance contract, no HTTP operation, no storage accessor, and nothing asserting the two
routes agreed.

## A separate namespace, on purpose

The leaf is **`memoryops/`, not `issueops/`**. Memories are a KV plane with their own merge semantics:
`internal/storage/kvkeys` exists because the merge resolver treats `kv.memory.*` config conflicts as
*convergent memory* conflicts.

It imports stdlib plus `issueops` **for sentinels only**, aliased — `var ErrValidation =
issueops.ErrValidation` — so `errors.Is` identity holds repo-wide rather than minting a rival
vocabulary. There is deliberately **no `ErrNotFound`**: the config seam cannot tell an absent row from
one stored empty, so a miss is result-carried via `Found bool`. That is a substrate fact, not a style
choice.

`engdocs/ADDING_AN_ISSUEOPS_ROLE.md` gains a "second namespace" section recording which `issueops`
conventions turned out load-bearing and which were situational, for whoever adds a third.

## It fixes a real bug on the way through

The shipped code probed then acted in **two transactions** (`memory.go:301`, `:413`). The `InTx` bodies
make probe-and-act one transaction on all three legs, which is what turns `Replaced` and "the value I
forgot" from hopeful into true.

## What the reviews caught

A design review, then a six-lens red-team council with every finding handed to a second agent
instructed to refute it. The findings worth naming:

**The design's bare-slug condition was unsafe.** It specified `DeriveKey(insight) == insight` as the
"this is a mistyped read" test — but `DeriveKey("")` is `""`, so that is **true for empty and
unslugifiable content**. As specified, `bd remember ""` and `bd remember "!!!"` would have stopped
being refusals and become *recalls of the empty key*. The shipped code was saved only by an
empty-content check that ran first, which the design then moved into the role without noticing it was
load-bearing twice.

**An unforgettable row.** `ResolveKey` accepted any non-empty explicit key while `ValidateKey` — which
`Recall` and `Forget` run — refuses keys empty after *trimming*. So `--key "   "` stored a row no
memory operation could ever name again: enumerated forever, unrecallable, unforgettable, reachable
only via `bd config unset`. POST accepted what GET and DELETE answered 400 for, unauthenticated. Four
lenses found it independently — what a genuinely unseeded quadrant looks like.

**Memories were leaking through three doors, not one.** They are config rows, and a prior wave put the
config plane behind `WorkspaceConfig` whose `ListSettings` returned `GetAllConfig` **unfiltered**, so
`GET /v0/beads/config` served every memory key *and value* verbatim — redaction matches key names
while the memory is in the value. Fixed at the role, so `bd config list` and the HTTP surface stop
together. The council then found `bd info --json` served the same map whole on both routes, and the
beads MCP `get_schema_info` tool runs `bd info --schema --json` and returns the parsed dict **including
config** — so every memory landed in the transcript of any agent that asked a *schema* question. Both
`info` routes now apply the same filter.

## Verification

Every fix mutation-tested — the new case run against the broken body to confirm it fails. W1 broke its
own bodies seven ways (`kv.` instead of `kv.memory.`, a LIKE-shaped delete, `Replaced = previous != ""`
caught independently at dolt *and* uow); the CLI chain proved the auto-commit epilogue by counting
`dolt_log` deltas with negative controls showing `74 -> 74 (delta 0)` while every other assertion still
passed; the search-boundary case was verified by transposing one line in the uow body and watching it
report `matched 3 memories, want 0`.

`gofmt`, `go build`, `go vet`, `make api-check`, `golangci-lint` 0 issues, the memories contract green
at dolt / embeddeddolt / uow, and the memory, prime, httpapi and memoryapi suites green.

## Open questions, defaults applied

Recorded rather than guessed at. Two are **published wire contracts once merged**:

1. `getMemory` returns a real **404** on a miss, diverging from the settings surface's no-404 doctrine.
2. **`?q=` collides in meaning** — `issues:query` already publishes `q` as a boolean expression with a
   vocabulary and a parse refusal; memory's is a bare substring match. The design claimed no precedent
   existed; it was wrong. Renaming to `search` is one line before merge.

Also open: whether point reads should refuse `kv.*` too (currently an enumeration rule, not a wall);
whether the shared sentinels deserve their own leaf before a third namespace; and `bd config show`,
which still enumerates memories while `bd config list`'s tip line points users straight at it.


julianknutsen:
## Red-team review council

Six lenses over the slice, every finding handed to a second agent instructed to **refute** it and to
default to refuted unless it could demonstrate the defect. 17 agents.

**Confirmed: 2 major, 6 minor. Refuted and dropped: 3.**

Four of the six minors are the same defect found independently by four different lenses — which is
what a genuinely unseeded quadrant looks like from the outside.

---

### MAJOR: POST accepts whitespace-only keys that every other memory door refuses, minting an unforgettable row — **FIXED in `594ab54ae`**

`internal/memoryapi/memoryapi.go:114`

ResolveKey treats any non-empty explicit key as valid (`if key != "" { return key, nil }`), while
ValidateKey — used by Recall and Forget — refuses keys that are empty AFTER TRIMMING
(`strings.TrimSpace(key) == ""`). The HTTP surface inherits the asymmetry exactly: rememberRequest
(internal/httpapi/memories.go:305) passes the body's `key` verbatim into Remember, but memoryKey
(internal/httpapi/memories.go:232-233) refuses the same string on GET/DELETE
/v0/beads/memories/{key} with 400 invalid_argument "`key` is empty after trimming". So the write
door accepts a key its two sibling operations, and the role's own read/remove methods, classify as
invalid. The conformance contract itself pins the read side
(backend/conformance/memories_contract.go:356-360 and :531-535 assert Recall("   ")/Forget("   ")
are ErrValidation) but has no Remember case for a whitespace-only explicit key — the exact unseeded
quadrant. The spec compounds it: RememberRequest.key promises "no trimming, no slugging, n

**Failure scenario.** Unauthenticated caller: POST /v0/beads/memories {"content":"note","key":" "} -> 200 {"key":"
","value":"note","replaced":false}; row kv.memory.<space> is created. Then DELETE
/v0/beads/memories/%20 -> 400 invalid_argument (not 404), GET /v0/beads/memories/%20 -> 400, `bd
forget " "` and `bd recall " "` -> ErrValidation. The memory is enumerated forever by GET
/v0/beads/memories, bd memories and bd prime, and cannot be removed through any memory operation,
HTTP or CLI — the only escape is the bd config unset escape hatch on the raw storage key.

### MAJOR: bd info --json still dumps every memory value, and the MCP beads_get_schema_info tool republishes it into agent transcripts — **FIXED in `594ab54ae`**

`cmd/bd/info.go:74`

The S1 exclusion's stated harm is that memories were "printed into every terminal and transcript"
(commit da6841125) — but `bd info --json` still does exactly that on both routes:
cmd/bd/info.go:74-77 puts the UNFILTERED store.GetAllConfig map into info["config"], and
cmd/bd/info_proxied_server.go:30-33 does the same over uw.ConfigUseCase().GetAllConfig. renderInfo
hides it in text mode but outputJSON(info) serves the whole map. The blast radius is not
hypothetical: integrations/beads-mcp/src/beads_mcp/bd_client.py:849 (get_schema_info) runs `bd info
--schema --json` and returns the ENTIRE parsed dict — top-level "config" included — as the result of
the beads_get_schema_info MCP tool, whose docstring claims it returns "tables, schema version,
config, sample issue IDs" for schema inspection. Every agent that calls that tool gets all
kv.memory.* keys AND values folded into its transcript. Risk class: this is CLI/direct-DB (and
proxied-socket) access, not the unauthenticated HTTP surface

**Failure scenario.** User runs `bd remember "the staging deploy token is sk-live-000"`. Later, any agent with the beads
MCP server calls beads_get_schema_info (or a user runs `bd info --json` and pastes the output into a
GitHub issue). The response contains "config": {"kv.memory.the-staging-deploy-token": "the staging
deploy token is sk-live-000", ...} — every memory key and full value — despite the release note
stating the kv plane is no longer enumerated to terminals and transcripts, and despite the caller
having asked a schema question, not a config question.

### MINOR: Whitespace-only explicit key mints a memory the role itself refuses to name — spec self-contradiction, no case anywhere — **FIXED in `594ab54ae`**

`internal/memoryapi/memoryapi.go:114`

ResolveKey checks `if key != "" { return key, nil }` — a key of "   " is non-empty, so Remember
stores it verbatim at `kv.memory.   ` (both bodies: dolt/memories.go:38, uow/memories.go:62). But
ValidateKey (memoryapi.go:97-102), used by Recall and Forget on all implementations, refuses any key
that is empty AFTER TRIMMING with ErrValidation. So the leaf spec contradicts itself:
memoryops/memories.go:13-15 promises "a stored key has to stay recallable under the exact bytes the
caller used", while memories.go:51-54 refuses exactly those bytes. Both front doors reach it:
cmd/bd/memory.go:327 passes memoryKeyFlag through raw, and the HTTP POST body path
(internal/httpapi/memories.go:294-305) deliberately does not trim, while GET/DELETE
/v0/beads/memories/{key} refuse the same key at memoryKey (memories.go:232-233, "empty after
trimming"). No conformance case (memories_contract.go tests interior spaces only, line 223) and no
memoryapi unit test (TestResolveKeyPrefersTheExplicitKey never pas

**Failure scenario.** `bd remember --key '   ' 'the prod db password rotation runbook'` (or POST /v0/beads/memories
{"key":"   ","content":...}) succeeds and reports Key="   ". Then `bd recall '   '`, `bd forget '
'`, GET and DELETE /v0/beads/memories/%20%20%20 ALL return ErrValidation/400 — the memory is
enumerable via List but can never be read by key or deleted through any memory surface; the only
escape hatch is `bd config unset 'kv.memory.   '`. RememberResult.Key's promise ("a caller that
supplied no key learns here what to recall") and RememberRequest's recallability promise are both
false for this input,

### MINOR: Search filtering before vs after prefix-strip is indistinguishable to the suite — a storage-key-matching List body passes green — **FIXED in `458fec34a`**

`backend/conformance/memories_contract.go:591`

RunMemoriesListSearchMatchesKeyOrValueCaseInsensitively uses terms ("searchable-key", "phantoms",
"SEARCHED-BY-VALUE", "<prefix>-matches-nothing-at-all") that are all substrings of the USER
key/value if and only if they are substrings of the STORAGE key/value. A body that folded over
unstripped storage keys — e.g. transposing uow/memories.go:138-139 to FilterMemories(all,
req.Search) before MemoriesFromConfig — passes every leg of the case and every other case (case 14's
storage-prefix loop at :575-580 only runs with Search=""). ListRequest.Search's promise
(memoryops/memories.go:100-101: match on "its key", i.e. the user key) has no case that
distinguishes the two boundaries, even though the file's own header names "a prefix trim ... at the
wrong boundary" as the bug family (:40). One term like "memory" or "kv." seeded against memories
whose user keys and values contain neither would pin it.

**Failure scenario.** With the one-line transposition in the uow body, `bd memories memory` (or GET
/v0/beads/memories?q=memory) over the proxied route returns the ENTIRE memory plane — every storage
key contains "kv.memory." — while the direct route returns only genuine matches. Route-dependent
answers, three legs green.

### MINOR: The header's seeding-discipline claim is false: ForgetReportsTheForgottenValue is a write case with no raw-row or survivor assertions — **FIXED in `458fec34a`**

`backend/conformance/memories_contract.go:462`

The header (memories_contract.go:36-38) claims "every WRITE case asserts the other three survived,
through RAW ROWS." Three cases fall short: (1) RunMemoriesForgetReportsTheForgottenValue (:462-477)
performs a Remember and a Forget, discards the seed's return value, and asserts nothing raw — no
assertMemoriesRawAbsent on the deleted row and no assertMemoriesNeighborsSurvived; it is the only
forget-after-Remember (rather than forget-after-out-of-band-seed) sequence in the suite and the one
whose collateral damage nothing would see, since the next case's seed rewrites all four neighbor
rows and repairs any corruption before it could be observed. (2)
RunMemoriesForgetNeverTouchesTheSettingsPlane (:453-456) omits n.memoryKey from its survivor set.
(3) RunMemoriesListSearchMatchesKeyOrValueCaseInsensitively performs two Remembers with no survivor
assertions. The prompt's own lens — verify the header case by case rather than trusting it — is
exactly what this fails.

**Failure scenario.** A Forget bug conditional on the deleted row having been written through the role in the same session
(e.g. a body that cached the storage key from the preceding Remember and swept with it) corrupts
neighbor rows during case 12; the corruption is overwritten by case 13's seedMemoriesAllFourClasses
before any assertion runs, and the suite stays green on all three legs — the same repair-by-reseed
blindness the header claims the discipline exists to prevent.

### MINOR: Remember accepts whitespace-only explicit keys that Recall/Forget refuse, creating write-only memories — **FIXED in `594ab54ae`**

`internal/memoryapi/memoryapi.go:114`

ResolveKey uses `if key != ""` to accept an explicit key verbatim, while ValidateKey
(memoryapi.go:94-97) — the gate for Recall and Forget — refuses any key that is empty after
TrimSpace. The two doors therefore disagree on the whitespace-only quadrant. Confirmed empirically:
ResolveKey("   ", content) returns "   " with nil error; ValidateKey("   ") returns ErrValidation
("memory key must not be empty"). Every write door passes the key verbatim (cmd/bd/memory.go:327
passes memoryKeyFlag unchecked; internal/httpapi/memories.go:301-305 deliberately does not trim the
POST body's key), and the HTTP read/delete doors independently 400 such keys at memoryKey
(internal/httpapi/memories.go:232, "`key` is empty after trimming"). This contradicts the leaf
contract's own promise (memoryops/memories.go, RememberRequest.Key: "a stored key has to stay
recallable under the exact bytes the caller used") and regresses origin/main, where
cmd/bd/memory.go:461 and :413/:418 round-tripped such keys throug

**Failure scenario.** `bd remember --key '   ' 'rotate the API token monthly'` succeeds and stores row kv.memory.'   '
(same via POST /v0/beads/memories {"content":"...","key":" "}). Then `bd recall '   '` and `bd
forget '   '` fail with "validation failed: memory key must not be empty" instead of
answering/deleting (on origin/main both worked), and GET/DELETE /v0/beads/memories/%20 return 400
invalid_argument, not the memory. The row is enumerable via `bd memories` / listMemories but cannot
be read or removed through any memory door; the only escape is `bd config unset "kv.memory.   "`.

### MINOR: Keys '.' and '..' are storable but path-unreachable under standard encoding; GET of key '.' 301-redirects to the collection list — **open**

`internal/httpapi/routes.go:270`

The spec says memory keys "may contain ... dots" and names only control characters as unreachable by
path (spec/openapi.v0.yaml:2144-2157, getMemory description ~2008-2016). But '.' and '..' are
unreserved characters that standard encoders (JS encodeURIComponent, Go url.PathEscape) never
percent-escape, and Go's ServeMux cleans dot-segments before matching: verified empirically against
the exact patterns this surface registers (server.go:866-868), GET/DELETE /v0/beads/memories/.
redirects 301 to /v0/beads/memories and /v0/beads/memories/.. redirects 301 to /v0/beads. Neither
request ever reaches memoryKey's 400 or the handler; 301 is not in the v0 status vocabulary and
carries no Problem body. The keys ARE reachable via the nonstandard %2E escaping (probe:
/memories/%2E%2E matches with key ".."), which no mainstream client emits. By contrast a slash key
is genuinely fine (%2F stays in-segment: /memories/a%2Fb -> key "a/b"), so this is specifically the
dot-segment family. Cheapest fix i

**Failure scenario.** bd remember --key "." "pinned note" stores the memory. A redirect-following HTTP client (fetch's
default) issues GET /v0/beads/memories/. expecting that memory: the 301 lands on GET
/v0/beads/memories, which answers 200 with the entire MemoriesPage — wrong schema, whole plane
served where one memory was requested, no error signaled. DELETE /v0/beads/memories/. similarly
never reaches the forget handler (redirect target answers 404 or, if the client downgrades the
method to GET per 301 semantics, the full list), so the memory cannot be forgotten over HTTP with
any standard encoder.

### MINOR: Whitespace-only explicit key: Remember accepts it, Recall/Forget refuse it — write-only, un-forgettable memory — **FIXED in `594ab54ae`**

`internal/memoryapi/memoryapi.go:114`

ResolveKey passes any non-empty explicit key verbatim (`if key != "" { return key, nil }`, line
114), but ValidateKey — used by Recall and Forget on all three backends — refuses any key that is
empty AFTER TRIMMING (`strings.TrimSpace(key) == ""`, line 95). So a whitespace-only key ("   ",
" ", etc.) is storable but not recallable or forgettable through the memory surface. This (a)
regresses origin/main, where `remember --key "   "` / `recall "   "` / `forget "   "` round-tripped
via raw GetConfig/DeleteConfig on `kv.memory.   `; (b) violates the role's own documented contract —
memoryops/memories.go RememberRequest doc says "a stored key has to stay recallable under the exact
bytes the caller used"; and (c) sits in exactly the quadrant the conformance contract never seeds:
backend/conformance/memories_contract.go pins Recall("   ") and Forget("   ") as ErrValidation
(lines ~283, ~356, ~531) but has NO Remember{Key: "   "} case, so all three legs are green over the
inconsistency. The o

**Failure scenario.** In a workspace: `bd remember "important fact" --key "   "` -> prints `Remembered [   ]` and stores
the row (verified against the built binary). Then `bd recall "   "` -> `Error: recalling memory:
validation failed: memory key must not be empty`, exit 1; `bd forget "   "` -> same refusal. `bd
memories` lists the memory under a blank key forever. A user who created such a key on origin/main
(where the round-trip worked) loses read/delete access to it after upgrading.

---

### Refuted and dropped

- **"A Remember of the value already stored still performs the write" — a merge-load-bearing promise with zero cas** — The finding's load-bearing claim — that the promise "IS observable on the uow leg via CountHistory" — is false. SetConfig is REPLACE INTO a (key,value)-only table (internal/storage/domain/db/config.go:83; migrations/0006_create_config.up.sql), so an identical- value rewrite leaves the Dolt working set byte-identical; DOLT_COMMIT('-Am') the

- **remember's validation refusals now require a reachable database — unclaimed behavior change in no-workspace di** — The finding's baseline claim is empirically false. It asserts that on origin/main, `bd remember ""` / `bd remember "!!!"` in a directory with no beads workspace printed the validation prose ("memory content cannot be empty" / "could not generate key from content...") because those checks sat before usesProxiedServer()/ensureDirectMode in

- **Filter rationale doc and CHANGELOG claim bd memories "reads the store directly", which this same branch made f** — The cited sentence (internal/workapi/workspaceconfig.go:80-83) makes one load-bearing contrast: the purpose-built views over kv rows get their data from the store, not through the WorkspaceConfig role. Verified at HEAD: (1) bd kv list still calls store.GetAllConfig inline at cmd/bd/kv.go:323, so half the sentence's subject is literally ac

julianknutsen:
## Post-review: the two open wire contracts, resolved

The council review above left two questions unanswered because both publish a wire contract on merge and neither can be changed cheaply afterwards. Both are now settled. One produced a commit; the other did not.

### Q3 — `listMemories` search parameter: renamed `q` → `search` (2aaf295ed)

`GET /v0/beads/issues:query` already publishes `q` as a **boolean expression** over issue fields, with a vocabulary and a parse refusal. `GET /v0/beads/memories` used `q` for a case-insensitive **substring match** with neither. One surface, two questions, one name — a client carrying the issue habit could send `?q=status:open` to the memory plane and get a literal substring search over memory text back, with a `200` on it.

Under the second name that request is an unknown parameter, which is what it actually is. The unknown-parameter table now pins `q` **specifically** as a refused name rather than leaving it as an unnamed member of "everything else".

Spec parameter, handler const, regenerated `ListMemoriesParams.Search`, tests. Mutation-checked: reverting the const to `"q"` fails 4 tests, including the collision case.

### Q2 — `getMemory`'s 404: unchanged, but the justification needed checking

The design justified the 404 with "`bd recall` already distinguishes a miss by exit code." I challenged that and briefly reported it as false, on a grep of the `recallCmd` body that came back empty. **The grep was too shallow.** The exit lives one call below, in `printRecallResult` (`cmd/bd/memory.go`) via `SilentExit()` → `&exitError{Code: 1}`; both the JSON and the human branch take it. The other side checks out too: `bd config get` on an absent key prints `(not set)` and returns nil — exit 0.

So the doctrine at `internal/httpapi/settings.go:91-93` is not "no 404 anywhere." It is **the status reports the distinctions the front door already reports**, and the two front doors differ:

| operation | CLI on a miss | HTTP |
| --- | --- | --- |
| `bd config get` | prints `(not set)`, exit 0 | `200`, no 404 |
| `bd recall` | stderr `No memory with key`, exit 1 | `404` |

The stored-empty conflation is real on both planes, but it lands on the **miss** side of the memory plane in the CLI and over HTTP alike, so the 404 claims nothing `bd recall` does not already claim. `listMemories` remains the documented escape hatch.

No behaviour change — but the handler comment now cites where the exit code lives, because a comment that asserts a fact without a pointer is a comment the next reader can conclude is false. That is exactly what happened here.

### Q2a — `forgetMemory`'s 404, checked while resolving Q2

Same claim underneath, so it got the same treatment. `ForgetInTx` returns `found=false` and **deletes nothing** when the stored value is empty (`internal/storage/memoryops/memories.go:126-128`), so `ForgetResult.Found` carries the same conflation its doc claims it does, and `printForgetNotFound` also exits 1. Consistent with the read beside it. No change.

### Still open, none blocking

- **Q1** — point-read firewall on the settings role. Only *enumeration* is filtered today; `GetSetting` will still hand back a `kv.*` key to a caller who names one. Flipping this later changes wire behaviour, so it wants an answer before release, not before merge.
- **Q4** — whether the shared sentinels get their own leaf rather than the alias. Cheap before a third namespace, expensive after.
- **Q5** — `bd config show` still enumerates the whole kv plane with values: `collectDatabaseEntries` calls `GetAllConfig` raw, unfiltered (`cmd/bd/config_show.go:292`). Left alone deliberately, rationale recorded at `cmd/bd/info.go:82` — an operator asks for `config show` by name, whereas `bd info --json` was being republished into agent transcripts by MCP `get_schema_info`. Requires direct DB access, so it is **not** the unauthenticated-HTTP disclosure this slice closed.

### CI

The `CHECK_DOC_FLAGS` failure and the gates that depend on it are main's state, not this PR's — three docs are 92 days past a 90-day review threshold, and it fails identically on #5381 and #5384. Clearing it means someone reviewing those three docs; bumping the `Last reviewed` dates would forge a review marker to get past a gate, so it is not done here.

PR #5388, opened 2026-08-07
Stop the 'human' label leaking between the operator queue and the dispatch queue

mwotton:
The `human` label is how a bead gets escalated to an operator: adding it puts the bead in `bd human list`, which is the whole decision queue. But the label is a free-floating tag rather than a state that participates in queue membership, and it leaks in both directions.

**(a) Adding `human` does not remove a bead from `bd ready`.** The same bead sits in the decision queue and the dispatch queue at once, so an agent can claim and work a question that is simultaneously awaiting a ruling — exactly the double-handling the human queue exists to prevent. Observed with a bead annotated DO NOT DISPATCH after an explicit "do not pursue" ruling: it kept being offered as workable. The workaround was `bd defer`, which means "will be revisited" and says the wrong thing about a question with no instrument to answer it.

**(b) Closing a bead does not remove it from `bd human list`.** Not `bd close`, and not `bd human respond` / `bd human dismiss`, which close the bead themselves. Nothing ever retires the label, so the queue accumulates resolved items and keeps presenting decisions that have already been made as though they were still owed. An operator had to strip labels by hand.

The two halves are the same missing notion, and fixing only (a) leaves the operator's queue eroding just as fast.

## What changed

**(a) — storage layer.** Ready work now excludes the `human` label by default. It goes in `sqlbuild.BuildReadyWorkWhere`, the one place both stacks render ready semantics, so every dispatch surface gets it: `bd ready`, `bd ready --claim`, `bd list --ready`, the counts path, and the claim path. The `issueops` wisps branch reaches the wisps table through `BuildLabelDrivenSearch` instead of that builder, so it applies the same exclusion itself.

Naming the label opts back in — `bd ready --label human` (or `--label-any`) returns that queue rather than the empty set an unconditional exclusion would produce, so nothing becomes unreachable. Only the exact spelling counts: `needs-human` is an ordinary label, and it does not appear in `bd human list` either. Flagged beads stay fully visible in `bd list` and `bd show`.

**(b) — `cmd/bd`.** `bd human list` excludes closed beads by default. Clearing the label on close was the other option and does not fit this codebase: the label is what lets `bd human stats` count a closed bead as Responded or Dismissed, and it is the record of why the bead was escalated at all. An explicit `--status` still reaches any status, closed included, so resolved items are omitted rather than hidden. The list and stats queries are now built separately, since they disagree about closed beads by design; both routes (direct and proxied-server) build both filters from the same two helpers.

The two halves are separate commits, split by layer, and can be read — or split into stacked PRs — independently.

## Docs

Per `docs/cli-docs.pin`, the docs corpus targets the pinned release rather than main, and `docs/cli-reference/` is generated. The user-facing wording therefore lives where it regenerates from: the `Long` strings on `bd ready` and `bd human list`. No hand edits under `docs/`.

## Tests

| Seam | Test |
|---|---|
| ready-work clause | `sqlbuild.TestReadyWorkExcludeLabels` (table: default, caller exclusions, no-duplicate, `--label`/`--label-any` opt-in, near-miss) and `TestBuildReadyWorkWhereExcludesHumanLabel` (SQL text, both table families) |
| ready work over a real store | `embeddeddolt.TestReadyWorkExcludesHumanLabelledIssues` — both entry points, opt-in, and that clearing the label makes the bead workable again |
| `bd ready` end to end | `TestEmbeddedReady/ready_excludes_human_label` |
| human list filter | `TestHumanListFilterExcludesClosed` — default excludes closed, explicit `--status` wins, stats keeps closed |
| `bd human list` end to end | `TestEmbeddedHuman/human_respond_and_dismiss` extended: closed beads leave the queue, `--status closed` still finds them, stats still counts Responded/Dismissed |
| proxied-server route | `TestProxiedServerHuman/stats_counts` extended with the default listing and `--status closed` |

Two existing `sqlbuild` assertions were updated for the added default arg: `TestBuildReadyWorkWhereBatchesIDSets` (now counts the deferred-children `NOT IN` shape specifically) and `TestBuildReadyWorkWhereLabelsAny` (arg tail).

Also verified by hand against a throwaway workspace with a locally built binary: a `human`-labelled bead disappears from `bd ready` while a `needs-human` one stays; `bd ready --label human` returns it; closing it removes it from `bd human list` while `--status closed`, `bd human stats` and `bd list` still show it.

### DIFF
--- a/cmd/bd/human.go
+++ b/cmd/bd/human.go
@@ -94,17 +94,41 @@ SUBCOMMANDS:
 	},
 }
 
+// humanListFilter is the query behind `bd human list`: beads carrying the
+// 'human' label, minus the closed ones, since closing never clears the label.
+// An explicit --status still reaches any status, closed included.
+func humanListFilter(status string) types.IssueFilter {
+	filter := types.IssueFilter{Labels: []string{types.LabelHuman}}
+	if status != "" {
+		s := types.Status(status)
+		filter.Status = &s
+		return filter
+	}
+	filter.ExcludeStatus = []types.Status{types.StatusClosed}
+	return filter
+}
+
+// humanStatsFilter is `bd human stats`'s query. Unlike the list it must keep
+// closed beads: Responded and Dismissed are counts of closed ones.
+func humanStatsFilter() types.IssueFilter {
+	return types.IssueFilter{Labels: []string{types.LabelHuman}}
+}
+
 // human list command
 var humanListCmd = &cobra.Command{
 	Use:   "list",
 	Short: "List all human-needed beads",
-	Long: `List all issues labeled with 'human' tag.
+	Long: `List all open issues labeled with 'human' tag.
 
 These are issues that require human intervention or input.
+
+Closed issues are omitted: a decision that has been made is no longer pending,
+and closing does not clear the label. Pass --status=closed to see them.
 
 Examples:
   bd human list
   bd human list --status=open
+  bd human list --status=closed
   bd human list --json`,
 	SilenceUsage:  true,
 	SilenceErrors: true,
@@ -124,14 +148,7 @@ Examples:
 			return runHumanListProxiedServer(ctx, status)
 		}
 
-		filter := types.IssueFilter{
-			Labels: []string{"human"},
-		}
-
-		if status != "" {
-			s := types.Status(status)
-			filter.Status = &s
-		}
+		filter := humanListFilter(status)
 
 		if err := ensureStoreActive(); err != nil {
 			return HandleErrorRespectJSON("listing human beads: %v", err)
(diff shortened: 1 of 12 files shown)


mwotton:
Note on the two red checks — both are pre-existing and unrelated to this PR.

`Check doc flags freshness` and `PR Policy (wrapper timing)` both run `scripts/check-doc-freshness.sh`, which is currently failing on two docs whose `Last reviewed` markers crossed the 90-day threshold today (both dated 2026-05-08, now 91 days):

- `docs/reference/json-schema.md`
- `engdocs/design/otel/otel-data-model.md`

Neither file is in this diff. The same two checks are red on #5387 and #5384.

Locally on this branch:
- `./scripts/check-doc-flags.sh` — PASSED, including `PASS: generated CLI docs are fresh`. This PR edits the `Long` strings on `bd ready` and `bd human list` and makes no hand edits under `docs/`, so it introduces no CLI-docs drift (the generated pages track `docs/cli-docs.pin` at v1.1.0 and will pick the wording up at the next pin bump).
- `golangci-lint` v2.10.1 (the pinned version) — 0 issues across the changed packages.
- `make ci-pr-policy` fails only on the two stale markers above.

mwotton:
Closing: this was filed against upstream by mistake. Re-filing against the mwotton/beads fork instead. No change to the work itself.

PR #5389, opened 2026-08-07
fix(dolt): honor --dolt-auto-commit batch/off in SQL-server mode (bd-4wamg)

steveyegge:
Fixes bd-4wamg (local bead; measured by wyvern/ganon under wy-x1gca): in SQL-server mode, `--dolt-auto-commit batch` was **silently inert** — one Dolt commit per write through all five configuration paths (per-invocation flag, `config.yaml`, env, `bd config set`, no-daemon), because the CLI's auto-commit policy hooks were all embedded-gated and the storage layer's per-write commit sites are unconditional.

## What changes

- **`issueOpsContext` / `transactHonoringAutoCommit` become mode-driven, not embedded-only.** batch/off set `WithDeferredVersionCommit` on the write-verb context / blank the `RunInTransaction` commit message, in server mode too. (`StageAndCommit` already no-ops on a blank message.)
- **The storage layer's three per-write commit sites honor the deferral** — `doltAddAndCommitInTx`, `doltAddAndCommit`, and the deleter's inline site — leaving writes in the working set for an explicit commit point (`bd dolt commit` / `CommitPending`). The explicit commit APIs (`Commit`, `CommitWithConfig`, `CommitPending`) and the federation sync point stay unguarded on purpose: they ARE the commit points.
- **Server-mode default flips `off`→`on` — spelling, not behavior.** The old default *resolved* to `off` while every value *behaved* like `on` (the storage layer committed per write regardless). With batch/off now real, keeping an `off` default would silently stop versioning server-mode writes on upgrade; `on` names what a default server-mode write has always done. Help text stops claiming `Default: off` (the other half of the bead's report).
- Embedded-only flush points (PersistentPostRun, create post-writes) keep their embedded gate via `embeddedWritesCommitNow`, now layered on the shared `writesCommitNow`.

## Semantics on a shared server

Staging is table-level (`DOLT_ADD(<table>)`), so an `on`-mode writer's commit sweeps up any deferred rows other clients left in the same tables. Deferral bounds who *creates* commits, not commit contents — that's the right shape for coalescing (nothing is ever lost; attribution of swept-up rows lands on the committing writer) and is documented in the changelog. This is also the substrate the app-layer interval-coalescing design (bd-e6oq6, Tim Sehn thread) schedules on top of; nothing here pre-empts that design.

## Tests

- `TestIssueOpsContextDefersByModeInServerMode`, `TestTransactHonoringAutoCommitBlanksMessageInServerBatch` (new unit, `cmd/bd`).
- `TestDeferredVersionCommitLeavesWritesForCommitPending`, `TestDeferredVersionCommitUpdatePath` (new, `internal/storage/dolt`, real Dolt container): deferred create/update leave `dolt_log` untouched while staying durable in SQL; `CommitPending` flushes as exactly one commit.
- All existing embedded-gate tests pass unchanged; embedded integration `TestDoltAutoCommit_{Batch,Off}` pass (`_On` fails on this box on an unmodified tree too — pre-existing env issue, not this change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   sweep up another client's deferred rows — deferral bounds who *creates*
   commits, not commit contents.
 
+  Scope: embedded and direct SQL-server modes. Proxied-server routes never
+  apply auto-commit policy (they short-circuit before it resolves), so
+  batch/off remain inert there — tracked as a follow-up. A deployment that
+  had explicitly configured `dolt.auto-commit: off` from the old help text
+  note: in server mode that was behaving like `on`; it now genuinely stops
+  minting version commits until an explicit `bd dolt commit`.
+
 
 - **`bd delete --cascade` no longer marks LIVE issues as deleted in other
   issues' text** (bd-x82so). When the deletion named a wisp, the set used to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`--dolt-auto-commit batch`/`off` now actually defer version commits in
+  SQL-server mode** (bd-4wamg). The mode was silently inert there: the CLI's
+  auto-commit policy was embedded-only, and the storage layer minted one Dolt
+  commit inside every write transaction regardless — measured at one commit
+  per write through all five configuration paths (flag, config.yaml, env,
+  `bd config set`, no-daemon). The write verbs now thread the batch/off
+  deferral to the storage layer's commit sites, which leave writes in the
+  working set for an explicit commit point (`bd dolt commit`), in server and
+  embedded mode alike. The server-mode **default** changes spelling, not
+  behavior: it used to resolve to `off` (while behaving like `on`); it now
+  resolves to `on`, naming what a default server-mode write has always done.
+  The `--dolt-auto-commit` help also stops claiming `Default: off`. On a
+  shared server, staging is table-level, so an `on`-mode writer's commit may
+  sweep up another client's deferred rows — deferral bounds who *creates*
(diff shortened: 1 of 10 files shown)


steveyegge:
**Fable adversarial review verdict: FIX_THEN_MERGE → blocker fixed in ba35ea524** (fresh-context reviewer, charged to refute; gate per Steve — no external reviewer holds this).

Reviewer verified the core mechanism clean: deferral is a per-call context flag with **no session/global server state** (no `@@dolt_transaction_commit`, no `SET GLOBAL` anywhere), so 37 mixed-mode clients can't perturb each other; the blank-message/`WithDeferredVersionCommit` plumbing is threaded correctly through create/update/close/dep/reopen/comment/label (checked verb by verb); `bd dolt commit`/`CommitPending` provably flush pending rows as one commit; and the default off→on flip is genuinely spelling-only for implicit-default users, checked against every mode consumer including the tips path. Crash story: batch-mode writes land in the branch working set, which is disk-durable — a server crash loses history granularity, not data.

**Blocker (fixed):** the deleter's `VersionCommitDeferred` guard was dead code — `delete.go` passed a bare `rootCtx` on both paths, so server-mode batch still minted one commit per `bd delete`. Fixed by threading `issueOpsContext` exactly as update.go does; new `TestDeferredVersionCommitDeletePath` proves it at the guard's own layer (plain ctx advances history, deferred ctx leaves the delete for `CommitPending` — RUN lines verified against the container), plus a batch-mode delete leg in the embedded container test.

**Follow-ups filed, not gated on:** (1) proxied-server routes never apply auto-commit policy, so batch/off remain inert there — scope now documented in help + CHANGELOG; hot because wyvern plans both the proxied cutover and batch adoption. (2) `bd dolt compact`/`flatten` end in an unguarded `DOLT_RESET --hard` — safe while per-write commits keep the shared working set clean, a real data-loss vector once fleet-wide batch makes dirty the steady state (other reset sites are already defended: merge-abort, pull, schema-lock). Squash-window runbooks gain a 'commit + drain first' step before any fleet flips batch on.

PR #5390, opened 2026-08-07
fix(test): roll ignored-migration cursor back to 0011 so the orphan-cleanup body re-runs (bd-g43r6)

steveyegge:
## Problem

`TestIgnoredMigration0011CleansOrphanedChildCounters` (internal/storage/dolt) fails on clean main: `orphaned counter rows = 1, want 0`.

The test marks the 0011 cleanup pending via `DELETE FROM ignored_schema_migrations WHERE version = schema.LatestIgnoredVersion()`. Pending detection is MAX-based (`currentVersion` = `MAX(version)` over the cursor table), so that delete re-runs only the **latest** ignored migration — which was 0011 when the test was written (#4538), but the series has since grown to 0021. The cleanup body never re-runs, and the seeded orphan survives.

## Fix

Roll the cursor back to before 0011 (`DELETE ... WHERE version >= 11`, anchored by a named const pointing at `0011_cleanup_orphaned_child_counters.up.sql`). The 0011..latest tail is idempotent by series contract (temp-table renames + INFORMATION_SCHEMA-gated ALTERs), so re-running it is safe and the migration under test actually executes.

## Verification

- On main (92ad0295f): test FAILS (`orphaned counter rows = 1, want 0`). After this change: PASSES.
- Sibling FAIL set in `schema_version_test.go` is unchanged before/after (verified by stash-diff): `TestSchemaRunsInitWhenMissing` and `TestMigration0053RepairsIssuesMissingRigColumns` fail on clean main locally too — pre-existing, filed as bd-17gmo, not touched here.

Bead: bd-g43r6

### DIFF
--- a/internal/storage/dolt/schema_version_test.go
+++ b/internal/storage/dolt/schema_version_test.go
@@ -315,6 +315,13 @@ func TestIgnoredMigration0011CleansOrphanedChildCounters(t *testing.T) {
 		t.Fatalf("commit counter fixture: %v", err)
 	}
 
+	// Bind the const to its file: if the ignored series is ever renumbered,
+	// fail here with a self-explaining message instead of silently replaying
+	// the wrong tail.
+	if _, err := schema.IgnoredMigrationSQL("0011_cleanup_orphaned_child_counters.up.sql"); err != nil {
+		t.Fatalf("orphanCleanupIgnoredVersion (%d) no longer matches an ignored migration file: %v", orphanCleanupIgnoredVersion, err)
+	}
+
 	// Pending detection is MAX-based (currentVersion = MAX(version)), so
 	// deleting only one cursor row re-runs only the LATEST ignored migration.
 	// Roll the cursor back to before 0011 so the orphan cleanup itself re-runs;
--- a/internal/storage/dolt/schema_version_test.go
+++ b/internal/storage/dolt/schema_version_test.go
@@ -271,6 +271,10 @@ func TestMigration0053RepairsIssuesMissingRigColumns(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


steveyegge:
**Fable adversarial review verdict: APPROVE** (fresh-context reviewer, charged to refute; gate per Steve — no external reviewer holds this).

Diagnosis verified against the code: `currentVersion` is `MAX(version)` (schema.go:1108) and the runner applies everything above it re-inserting cursor rows, so the old single-row delete re-ran only 0021 and the 0011 body never executed. The reviewer read **every** ignored migration 0011..0021 and confirmed each replay-safe (temp-table + conditional RENAME, INFORMATION_SCHEMA-gated ALTER/INDEX, COLUMN_TYPE-gated MODIFY, deterministic derived-state recompute, or pure marker); the one non-obvious hazard — deleting the 0018 row re-arms aux-rekey pass 2 — is defused by its `mainVersionBefore >= 61` gate (aux_row_id_backfill.go:239), and the 0009 marker is untouched. Idiom matches existing tests replaying larger tails (`aux_row_id_rekey_test.go` deletes >=9, `wisp_row_lock_heal_test.go` >=13). Empirically re-verified in a worktree: fixed test genuinely RUNs and passes under the dolt container; the two pre-existing failures fail identically on this branch (bd-17gmo). Fixing the test, not the MAX-based detector, is the right layer — cursor gaps cannot arise through any product write path.

Review nit applied in 3d8538b99: const bound to its migration file via a `schema.IgnoredMigrationSQL` existence assertion. Bonus finding recorded on bd-17gmo: TestMigration0053RepairsIssuesMissingRigColumns is likely the same MAX-cursor staleness bug on the MAIN series, not an env issue.

PR #5391, opened 2026-08-07
fix(test): pin DOCKER_HOST from the active docker context before TestMain swaps HOME (bd-84kos)

steveyegge:
## Problem

`cmd/bd`'s `testMainInner` points HOME at a temp dir before any test runs, which hides `~/.docker/config.json` — the file holding the docker CLI's **active context**. On machines where the daemon is reached via a context (OrbStack, Docker Desktop's `desktop-linux`, remote daemons), `docker info` then falls back to `unix:///var/run/docker.sock` and fails instantly, so every container-gated test SKIPs `Docker not available` even though docker works fine in the invoking shell.

Skips read as green, so this is a silent coverage hole: the proxied-server parity suites stopped running entirely on such machines until DOCKER_HOST was passed by hand (bit us on bd-i8qx8).

## Fix

- New `testutil.PinDockerHostFromContext()`: resolves the active context's endpoint via `docker context inspect -f '{{.Endpoints.docker.Host}}'` into DOCKER_HOST. No-op when DOCKER_HOST is already set (explicit choice wins) or when docker/the endpoint can't be resolved. On a default-context Linux box it resolves to `unix:///var/run/docker.sock`, i.e. harmless.
- `testMainInner` calls it immediately before the HOME swap.

cmd/bd is the only container-gated package that swaps HOME process-wide in TestMain, so that's the one call site; the helper is in testutil for any future TestMain that grows the same shape.

## Verification (macOS + OrbStack, NO DOCKER_HOST in env)

- main (92ad0295f): `BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd/ -run TestProxiedServerExternalCreate` → `SKIP: Docker not available`
- with this change: same command → `PASS (22s)` against a live container
- unit tests for the helper (pre-set value respected, stub-context resolution, no-docker no-op): green

Bead: bd-84kos

### DIFF
--- a/cmd/bd/test_repo_beads_guard_test.go
+++ b/cmd/bd/test_repo_beads_guard_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/steveyegge/beads/internal/config"
 	"github.com/steveyegge/beads/internal/doltserver"
 	"github.com/steveyegge/beads/internal/metrics"
+	"github.com/steveyegge/beads/internal/testutil"
 )
 
 // beforeTestsHook is set by CGO-tagged test files to perform setup before tests run
@@ -99,6 +100,12 @@ func testMainInner(m *testing.M) int {
 			_ = os.Setenv("GOMODCACHE", strings.TrimSpace(string(out)))
 		}
 	}
+
+	// The docker CLI's active context also lives under HOME
+	// (~/.docker/config.json); resolve it into DOCKER_HOST now or every
+	// container-gated test skips "Docker not available" on context-routed
+	// daemons like OrbStack (bd-84kos).
+	testutil.PinDockerHostFromContext()
 
 	_ = os.Setenv("HOME", tmp)
 	_ = os.Setenv("USERPROFILE", tmp) // Windows compatibility
--- a/internal/testutil/dockercontext.go
+++ b/internal/testutil/dockercontext.go
@@ -1,9 +1,11 @@
 package testutil
 
 import (
+	"context"
 	"os"
 	"os/exec"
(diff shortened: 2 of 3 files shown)


steveyegge:
**Fable adversarial review verdict: APPROVE** (fresh-context reviewer, charged to refute; gate per Steve — no external reviewer holds this).

The fix is minimal, correctly placed (verified the single HOME-swapping container-gated TestMain in the repo — every other TestMain leaves HOME intact, so docker CLI + testcontainers there read the real ~/.docker), and fail-safe by construction: no-op when DOCKER_HOST is set, docker absent (<1ms LookPath failure), or the endpoint can't resolve — verified empirically that `docker context inspect -f` **errors** rather than emitting `<no value>` on malformed output, so shim CLIs (podman) degrade to no-op. CI unaffected: on GH Actions the default context resolves to `unix:///var/run/docker.sock`, bit-identical to the existing fallback, and testcontainers-go v0.43.0 ranks `tc.host` properties above DOCKER_HOST, so Testcontainers Desktop/Cloud configs still win — while the pin repairs testcontainers' own context lookup (which reads the swapped HOME). Windows cross-compiles clean, unit tests `!windows`-gated and genuinely RUNning. Worst reachable outcome of a wrong pin is a SKIP (the `docker info` gate sits downstream), never a false red. One deliberate semantic: a stale active context over a working default socket now skips consistently with the invoking shell instead of accidentally running.

Review's two hardening nits applied in fddeb6562: 5s exec timeout + endpoint-shape (`://`) check before pinning.

PR #5392, opened 2026-08-07
feat(search): include closed issues by default (bd-t5yex)

steveyegge:
## Problem

`bd search` silently excludes closed issues unless you pass `--status all`. The dominant real-world search query is **"was this already found/filed/fixed?"** — exactly the query where excluding closed issues produces a false "no". Downstream repos grew shell wrappers (wyvern's `bd-find.sh`) purely to undo this default; when a repo writes a wrapper to change a default, the default is probably wrong.

## Change

- Default (no `--status`) now searches **all** statuses, on both the embedded path (`search.go`) and the proxied-server path (`search_proxied_server.go`).
- `bd list` keeps its open-only default — this is search-specific.
- `--status open` (or any status list) narrows explicitly where the old behavior is wanted; `--status all` still works, now redundantly.
- Help text + flag description updated; CHANGELOG entry under Unreleased/Changed. `docs/cli-reference` is release-pinned (`docs/cli-docs.pin`) and regenerates at release, so no hand-edits there.

This reverses the hq-319 performance default (open-only to use the status index on 12K+-issue DBs). That trade bought scan scope at the cost of wrong answers to the most common query; callers who care can still narrow.

## Callers audit

No in-repo programmatic caller depends on open-only search output: `prime.go` and the skill docs mention the command generically; nothing shells `bd search` and assumes closed-exclusion.

## Verification

- `TestEmbeddedSearch` (BEADS_TEST_EMBEDDED_DOLT=1): PASS, including new `search_default_includes_closed` subtest.
- `TestProxiedServerSearch` (BEADS_TEST_PROXIED_SERVER=1, live container): PASS, with `default_excludes_closed` flipped to `default_includes_closed`.
- Explicit-status subtests (open / closed / all / comma-separated) unchanged and green.

The bead's second ask (folding archive-jsonl search in, since closed beads get swept out of the live DB) is out of scope here — filed separately.

Bead: bd-t5yex

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   works and is now redundant. Applies to both the embedded and proxied-server
   paths.
 
+  Two consequences to know: matches beyond `--limit` (default 50) are dropped
+  under a status-blind sort (priority, then created), so a broad query on a
+  large DB can now fill the page with closed matches and silently drop open
+  ones — use `--status open` or raise `--limit` when hunting live work. And
+  `bd query` (plus the HTTP `q=` endpoint) deliberately keeps its closed-
+  exclusion default with opt-in `--all`; only `bd search` changed.
+
 ### Fixed
 
 - **`bd delete --cascade` no longer marks LIVE issues as deleted in other
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **`bd search` now includes closed issues by default** (bd-t5yex). The
+  dominant real-world search query is "was this already found/filed/fixed?" —
+  exactly the query where silently excluding closed issues produced a false
+  "no" (downstream repos grew shell wrappers purely to undo the old default).
+  `bd list` keeps its open-only default. Narrow explicitly with `--status open`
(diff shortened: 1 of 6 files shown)


steveyegge:
**Fable adversarial review verdict: FIX_THEN_MERGE → fixes applied in 652e6d01d** (fresh-context reviewer, charged to refute; gate per Steve — no external reviewer holds this).

Reviewer swept the entire repo for downstream callers: MCP server drives `bd list` (unaffected); no script/integration asserts open-only search output; the server side applies no status defaulting so there's no CLI/server version skew; all other `SearchIssues` callers build their own filters. Notably `tests/regression` has a `TestDiscovery_SearchIncludesClosedByDefault` that asserted THIS behavior all along — corroborating that hq-319 regressed the original semantics. The hq-319 performance concern doesn't survive contact with the query builder: the removed predicate was an unindexable `NOT IN`, the text match is a LIKE scan regardless, Pattern B bounds hydration to the limit, and a wider candidate set satisfies LIMIT earlier on common terms. Both suites verified green by the reviewer independently (embedded 29/29, proxied 33/33 against a live container).

Fixes applied per the verdict: (1) conformance corpus `search-desc-contains-and-status` — comment was still the old spec and the flip left two steps identical; comment updated + one step switched to `--status open` to keep an explicit exclusion probe. (2) Page-eviction consequence documented in help + CHANGELOG (matches beyond `--limit` drop under a status-blind sort), plus a line that `bd query`/HTTP `q=` deliberately keep closed-exclusion. Closed-last secondary sort + truncation indicator filed as a follow-up feature bead.

Issue #1070, opened 2026-01-13
bd comments add requires full ID while other commands accept short IDs


pentop:
## Problem

Most `bd` commands accept short IDs (e.g., `5wbm`), but `bd comments add` requires the full ID (e.g., `mike.vibe-coding-5wbm`).

This is inconsistent and confusing since `bd list` displays the short form prominently.

## Steps to Reproduce

```bash
# This works
bd update 5wbm --status in_progress

# This fails
bd comments add 5wbm "My comment"
# Error: adding comment: operation failed: failed to add comment: issue 5wbm not found

# This works
bd comments add mike.vibe-coding-5wbm "My comment"
# Comment added to mike.vibe-coding-5wbm
```

## Expected Behavior

`bd comments add` should accept short IDs like other commands do.

## Environment

- bd version 0.47.0 (f459ec29)
- Linux (WSL2)

pentop:
## Additional Finding

While investigating this issue, I discovered that `comments list` (i.e., `bd comments <id>`) has the **same bug** - it also doesn't resolve short IDs in daemon mode.

### Code Analysis

Both commands have the same pattern:

**`comments add` (lines 147-168 of comments.go):**
```go
if daemonClient != nil {
    resp, err := daemonClient.AddComment(&rpc.CommentAddArgs{
        ID:     issueID,  // ❌ Short ID passed as-is
        ...
    })
}
```

**`comments list` (lines 40-55 of comments.go):**
```go
if daemonClient != nil {
    resp, err := daemonClient.ListComments(&rpc.CommentListArgs{
        ID: issueID  // ❌ Short ID passed as-is
    })
}
```

Both commands correctly resolve IDs in direct mode (lines 63-67 and 176-180), but skip resolution in daemon mode.

### Fix Pattern

Both should follow the pattern from `update.go`, `label.go`, and other commands that call `daemonClient.ResolveID()` before their main operation:

```go
if daemonClient != nil {
    resolveArgs := &rpc.ResolveIDArgs{ID: issueID}
    resp, err := daemonClient.ResolveID(resolveArgs)
    // ... unmarshal to get full ID ...
    issueID = resolvedID
}
```

I'll include the fix for both in my PR.

PR #5393, opened 2026-08-07
Refuse reserved words and fuzzy id matches on bd comment writes

athosmartins:
## What

`bd comment <id> [text...]` now refuses two unsafe cases before writing anything:

1. The `<id>` argument being a "comments" subcommand name (`list`, `add`, `rm`, `delete`) — the shape you get from typo-ing `bd comments list <id>` / `bd comments <id>` as `bd comment list <id>`.
2. The `<id>` argument resolving only via leading-prefix *abbreviation* matching (e.g. it happens to be a prefix of some unrelated issue's hash) rather than an exact id or exact hash match.

## Why

`bd comment` resolves its id through the same abbreviation-tolerant `ResolvePartialID` used everywhere else, with no extra guard for a write path. Two failure modes fall out of that combination:

- `bd comment list <id>`, a typo for `bd comments list` (or `bd comments <id>`), silently ran with `id="list"` and `text="<id>"` instead of erroring — cobra's `Args: cobra.MinimumNArgs(1)` happily accepts it, there's just no subcommand tree to catch the mistake the way `bd comments list` already does (see the existing `commentsMisplacedListCmd` guard).
- Because "list" is a valid partial-ID shape (letters only), `ResolvePartialID` fell through to its leading-prefix abbreviation branch. If "list" happened to be a prefix of some unrelated issue's hash, that issue silently became the target instead of the lookup failing.

This is not hypothetical — it's a real, repeated production incident: 15+ automated sessions made the `bd comment list <id>` typo over two days, and because `list` was a leading-prefix abbreviation of an unrelated wisp's hash (`list3t0`), every one of them silently wrote a garbage comment onto that wisp instead of getting an error.

## How

- `checkCommentIDNotReservedWord` (new, `cmd/bd/comment.go`) rejects `list`/`add`/`rm`/`delete` as the id argument before any resolution is attempted, with a message pointing at the equivalent `bd comments` subcommand. Runs before the proxied-server/direct dispatch split so it applies uniformly.
- `ResolvePartialIDExact` (new, `internal/utils/id_parser.go`) is an exact-match sibling of `ResolvePartialID`: identical exact-ID and exact-hash matching (including a wisp's hash with its `wisp-` infix stripped — that's a real, deliberate convenience and stays intact), but it never falls through to the two leading-prefix-abbreviation branches. Implemented as a shared-logic refactor (`resolvePartialID(..., allowAbbrev bool)`) so `ResolvePartialID`'s own behavior for every existing caller is provably unchanged (it's now a one-line wrapper with `allowAbbrev=true`).
- `resolveAndGetIssueForMutationExact` (new, `cmd/bd/routed.go`) is `bd comment`'s new resolver entry point. It mirrors `resolveAndGetIssueForMutation`'s existing local → prefix-routing → contributor-auto-routing fallback chain (so cross-repo comment writes by full id are unaffected), but every tier now resolves via `ResolvePartialIDExact` instead of `ResolvePartialID`. This required threading a new `exact bool` parameter through `resolveViaPrefixRoutingWithAccess` and `resolveViaAutoRouting`; every pre-existing call site of both was updated to pass an explicit `false`, preserving current behavior byte-for-byte.

`bd close`, `bd update`, `bd label`, and every other write command keep using the existing abbreviation-tolerant resolver unchanged — this PR only tightens `bd comment`, which is where the incident happened and where the acceptance criteria are scoped.

## Verification

```
source ./.buildflags
go build ./...   # exit 0
go vet ./...     # exit 0
go test ./cmd/bd/... ./internal/utils/... -run 'TestCheckCommentIDNotReservedWord|TestResolvePartialIDExact' -v
```

- `TestCheckCommentIDNotReservedWord` (new, pure Go, no store) — covers the reserved-word guard directly, including that a real id merely *containing* a reserved word (`ga-wisp-list3t0`) is not rejected.
- `TestResolvePartialIDExact_FakeStore_ReproducesReportedIncident` (new, `internal/utils`, in-memory fake store — no Docker/cgo needed) — reproduces the actual incident shape end to end: asserts `ResolvePartialID("list")` still resolves to the colliding wisp (proves the fixture is real, not a fixture that trivially passes), then asserts `ResolvePartialIDExact("list")` refuses it, then asserts exact/full-id lookups still succeed.
- `TestResolvePartialIDExact_Wisp`, `TestResolvePartialIDExact_RejectsAbbreviationAcceptsExact`, `TestResolvePartialIDExact_ReservedWordDoesNotCollideWithWisp` (new, Docker-backed table tests mirroring the existing `TestResolvePartialID_Wisp` / `TestResolvePartialID` suites) — compiled and confirmed to skip cleanly in this sandbox (no Docker available here), same as the pre-existing Docker-backed suite; will run for real in CI.

No regressions observed in the pre-existing `ResolvePartialID*` suite (same skip/pass pattern as before this change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -185,6 +185,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`bd comment`/`bd comments add` refuse an abbreviated id and a reserved
+  word instead of silently resolving one to the wrong issue**
+  ([#5393](https://github.com/gastownhall/beads/pull/5393)). `bd comment list
+  <id>`, a typo for `bd comments list`, used to parse as id="list" with no
+  guard on the id slot — and since no issue is ever literally named "list",
+  the default abbreviation-tolerant resolver would fuzzy-match it against
+  whatever existing bead or wisp id happened to start with "list" and write
+  the rest of the command line there as a comment. Confirmed in production:
+  15+ automated sessions made this exact typo over two days and all landed on
+  the same unrelated wisp. `bd comment`'s id positional now (a) refuses the
+  four words a `bd comments <verb>` typo is likely to produce — `list`, `add`,
+  `rm`, `delete` — before any id resolution is attempted, on both the
+  embedded and proxied-server paths, and (b) requires an EXACT id match
+  (`bd-a1b2c3d4`, not `a1b2`) rather than the abbreviation-tolerant matching
+  every other command keeps. The refusal message is truthful about which
+  case applies: a reserved word gets a hint toward `bd comments`; an
+  abbreviation of a REAL issue gets `id abbreviations are not accepted on
+  comment writes; use the full id from \`bd show <id>\`` rather than the
+  generic (and in that case false) "no issue found matching" text a genuine
+  not-found gets. See [Working with
+  IDs](docs/core-concepts/hash-ids.md#working-with-ids) for the full
+  read/write distinction. The proxied-server comment path was investigated
+  separately and found to already be exact-id-only by construction (its
+  resolution never performed abbreviation matching in the first place) — a
+  regression test now locks that in rather than leaving it undocumented.
+
 - **`bd prime` says when it could NOT read the memory plane**
   ([#5877](https://github.com/gastownhall/beads/issues/5877)). A broken or
   unreachable store made prime omit the memory section entirely, so a session
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -1525,11 +1525,13 @@ func TestCLI_CommentsSwappedAddRejectedInProxiedServerMode(t *testing.T) {
 	}
 }
 
-// TestCLI_CommentsAddShortID tests that 'comments add' accepts short IDs (issue #1070)
-// Most bd commands accept short IDs (e.g., "5wbm") but comments add previously required
-// full IDs (e.g., "mike.vibe-coding-5wbm"). This test ensures short IDs work.
-//
-// Note: Short IDs work because the code calls utils.ResolvePartialID().
+// TestCLI_CommentsAddShortID tests that 'comments add' accepts a bare full
+// hash without its issue prefix (e.g. "5wbm" for "mike.vibe-coding-5wbm",
+// issue #1070) — full IDs used to be required. It does NOT test genuine
+// abbreviation (a hash shorter than the real one): since PR #5393, comment
+// writes require an exact id (see resolveAndGetIssueForMutationExact /
+// utils.ResolvePartialIDExact) and PartialIDWithCommentsAdd below asserts
+// that a real abbreviation is refused, not accepted.
 func TestCLI_CommentsAddShortID(t *testing.T) {
 
 	t.Run("ShortIDWithCommentsAdd", func(t *testing.T) {
@@ -1582,6 +1584,14 @@ func TestCLI_CommentsAddShortID(t *testing.T) {
 		}
 	})
 
+	// PartialIDWithCommentsAdd used to assert that a genuinely abbreviated id
+	// (shorter than the full hash) succeeded on "comments add" — true when
+	// this test was written (issue #1070), but no longer the contract: PR
+	// #5393 review item M1 established that "comments add" must refuse an
+	// abbreviation exactly like "comment" (singular) does (see
+	// TestCLI_CommentAbbreviatedIDRejectedWithTruthfulMessage), since the two
+	// are documented as guarded twins (CHANGELOG.md, hash-ids.md). This
+	// subtest now asserts the refusal instead of the old silent-accept.
 	t.Run("PartialIDWithCommentsAdd", func(t *testing.T) {
 		tmpDir := setupCLITestDB(t)
 
@@ -1601,16 +1611,40 @@ func TestCLI_CommentsAddShortID(t *testing.T) {
 		if len(shortID) > 4 {
 			shortID = shortID[:4] // Use only first 4 chars for partial match
 		}
+		if shortID == parts[len(parts)-1] {
+			t.Fatalf("test setup: hash %q too short to truncate into a genuine abbreviation", parts[len(parts)-1])
+		}
 		t.Logf("Full ID: %s, Partial ID: %s", fullID, shortID)
 
-		// Add comment using partial ID
-		stdout, stderr, err := runBDInProcessAllowError(t, tmpDir, "comments", "add", shortID, "Comment via partial ID")
+		// Sanity check: the same abbreviation still works on a READ path
+		// (show), proving this is a real, resolvable abbreviation and not an
+		// accident of the fixture — the refusal below is specific to writes.
+		showOut, showErr, err := runBDInProcessAllowError(t, tmpDir, "show", shortID, "--json")
 		if err != nil {
-			t.Fatalf("comments add with partial ID failed: %v\nstdout: %s\nstderr: %s", err, stdout, stderr)
+			t.Fatalf("fixture sanity check failed: 'bd show %s' unexpectedly failed: %v\nstdout: %s\nstderr: %s", shortID, err, showOut, showErr)
+		}
+		if !strings.Contains(showOut, fullID) {
+			t.Fatalf("fixture sanity check failed: 'bd show %s' did not resolve to %s, got: %s", shortID, fullID, showOut)
 		}
 
-		if !strings.Contains(stdout, "Comment added") {
-			t.Errorf("Expected 'Comment added' in output, got: %s", stdout)
+		// Add comment using partial ID — must now be refused, truthfully.
+		stdout, stderr, err := runBDInProcessAllowError(t, tmpDir, "comments", "add", shortID, "Comment via partial ID")
+		if err == nil {
+			t.Fatalf("expected non-zero exit for 'bd comments add %s ...' (abbreviation on a write path), got stdout=%q stderr=%q", shortID, stdout, stderr)
+		}
+		combined := stdout + stderr
+		if !strings.Contains(combined, "abbreviations are not accepted") {
+			t.Errorf("expected the truthful abbreviation-refusal message, got stdout=%q stderr=%q", stdout, stderr)
+		}
+		if strings.Contains(combined, "not found") {
+			t.Errorf("message must not claim the issue was not found — %s exists, only the abbreviation was refused; got stdout=%q stderr=%q", fullID, stdout, stderr)
+		}
+
+		// Regression check: no comment silently landed on the real issue.
+		commentsOut := runBDInProcess(t, tmpDir, "comments", fullID, "--json")
+		trimmed := strings.TrimSpace(commentsOut)
+		if trimmed != "[]" && trimmed != "null" {
+			t.Fatalf("expected no comments on %s after rejected abbreviated 'comments add', got: %s", fullID, commentsOut)
 		}
 	})
 
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -1605,15 +1605,22 @@ func TestCLI_CommentsAddShortID(t *testing.T) {
 		json.Unmarshal([]byte(jsonOut), &issue)
 		fullID := issue["id"].(string)
 
-		// Extract short ID and use only first 4 characters (partial match)
+		// Derive the abbreviation as a strict, one-character-shorter leading
+		// prefix of the real hash — the same derivation
+		// comment_proxied_integration_test.go:247 already uses. The previous
+		// "truncate to 4 chars, only if longer than 4" logic never fired
+		// against a fresh test DB's fixture hashes (min_hash_length unset,
+		// hashes as short as 3 chars), so shortID stayed equal to the full
+		// hash and this subtest was dead-red in its own setup, never reaching
+		// the assertions below (PR #5393 review R2, bee-ghosttrack) — and
+		// mutation P-B (reverting comments.go's resolver back to the
+		// abbreviation-tolerant one) went uncaught as a result.
 		parts := strings.Split(fullID, "-")
-		shortID := parts[len(parts)-1]
-		if len(shortID) > 4 {
-			shortID = shortID[:4] // Use only first 4 chars for partial match
+		hash := parts[len(parts)-1]
+		if len(hash) < 2 {
+			t.Fatalf("test setup: hash %q too short to abbreviate", hash)
 		}
-		if shortID == parts[len(parts)-1] {
-			t.Fatalf("test setup: hash %q too short to truncate into a genuine abbreviation", parts[len(parts)-1])
-		}
+		shortID := hash[:len(hash)-1]
 		t.Logf("Full ID: %s, Partial ID: %s", fullID, shortID)
 
 		// Sanity check: the same abbreviation still works on a READ path
@@ -1894,6 +1901,57 @@ func TestCLI_CommentRmDeleteReservedWordsRejected(t *testing.T) {
 			trimmed := strings.TrimSpace(commentsOut)
 			if trimmed != "[]" && trimmed != "null" {
 				t.Fatalf("expected no comments on bystander issue %s after rejected 'comment %s', got: %s", fullID, word, commentsOut)
+			}
+		})
+	}
+}
+
+// TestCLI_CommentsAddReservedWordRejectedThroughCobraDispatch closes PR #5393
+// review R2's "NEW MINOR — the guard's wiring is untested": comments_add_test.go
+// unit-tests validateCommentsAddArgs directly, and that catches a gutted
+// function body (mutation P-C), but nothing drove "bd comments add
+// <reserved-word> ..." through actual cobra dispatch — unwiring
+// commentsAddCmd's Args field back to cobra.MinimumNArgs(1) left every
+// existing test green (mutation P-D). This runs the real command line, the
+// way a caller or an automated session actually invokes it, for every word in
+// commentReservedIDWords (validateCommentsAddArgs has no singular/plural
+// special-casing, unlike "comment"'s validateCommentArgs, so all four take
+// the same generic-message path and are equally worth covering here).
+func TestCLI_CommentsAddReservedWordRejectedThroughCobraDispatch(t *testing.T) {
+	t.Parallel()
+
+	for _, word := range []string{"list", "add", "rm", "delete"} {
+		t.Run(word, func(t *testing.T) {
+			t.Parallel()
+			tmpDir := setupCLITestDB(t)
+
+			out := runBDInProcess(t, tmpDir, "create", "Bystander issue for comments-add "+word+"-typo", "-p", "1", "--json")
+			jsonStart := strings.Index(out, "{")
+			if jsonStart < 0 {
+				t.Fatalf("No JSON found in create output: %s", out)
+			}
+			var issue map[string]interface{}
+			if err := json.Unmarshal([]byte(out[jsonStart:]), &issue); err != nil {
+				t.Fatalf("Failed to parse create JSON: %v\nOutput: %s", err, out)
+			}
+			fullID := issue["id"].(string)
+
+			stdout, stderr, err := runBDInProcessAllowError(t, tmpDir, "comments", "add", word, "accidental stray text")
+			if err == nil {
+				t.Fatalf("expected non-zero exit for 'bd comments add %s ...', got stdout=%q stderr=%q", word, stdout, stderr)
+			}
+			combined := stdout + stderr
+			if !strings.Contains(combined, "not a valid issue id") {
+				t.Errorf("expected a not-a-valid-issue-id refusal, got stdout=%q stderr=%q", stdout, stderr)
+			}
+
+			// The regression check: the bystander issue must have received NO
+			// comment — this is exactly what unwiring commentsAddCmd.Args back
+			// to cobra.MinimumNArgs(1) would silently allow through.
+			commentsOut := runBDInProcess(t, tmpDir, "comments", fullID, "--json")
+			trimmed := strings.TrimSpace(commentsOut)
+			if trimmed != "[]" && trimmed != "null" {
+				t.Fatalf("expected no comments on bystander issue %s after rejected 'comments add %s', got: %s", fullID, word, commentsOut)
 			}
 		})
 	}
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -1810,6 +1810,125 @@ func TestCLI_CommentTextStartingWithReservedWordStillWorks(t *testing.T) {
 	}
 }
 
+// TestCLI_CommentRmDeleteReservedWordsRejected covers the two reserved words
+// TestCLI_CommentListMisplacedSyntax/TestCLI_CommentAddMisplacedSyntax do not:
+// "rm" and "delete" carry no hand-written case in validateCommentArgs (unlike
+// "list"/"add"), so they fall through to checkCommentIDNotReservedWord's
+// generic message and, before this test, had no CLI-level (cobra-dispatch)
+// coverage at all — steveyegge's PR #5393 review (item a) flagged that
+// generic message as potentially misleading for these two specifically, since
+// neither is an actual "bd comments" subcommand the way "list"/"add" are.
+// This pins both halves: the command is rejected, AND — the part that
+// actually matters — no comment silently lands on an unrelated real issue.
+func TestCLI_CommentRmDeleteReservedWordsRejected(t *testing.T) {
+	t.Parallel()
+
+	for _, word := range []string{"rm", "delete"} {
+		t.Run(word, func(t *testing.T) {
+			t.Parallel()
+			tmpDir := setupCLITestDB(t)
+
+			out := runBDInProcess(t, tmpDir, "create", "Bystander issue for "+word+"-typo", "-p", "1", "--json")
+			jsonStart := strings.Index(out, "{")
+			if jsonStart < 0 {
+				t.Fatalf("No JSON found in create output: %s", out)
+			}
+			var issue map[string]interface{}
+			if err := json.Unmarshal([]byte(out[jsonStart:]), &issue); err != nil {
+				t.Fatalf("Failed to parse create JSON: %v\nOutput: %s", err, out)
+			}
+			fullID := issue["id"].(string)
+
+			stdout, stderr, err := runBDInProcessAllowError(t, tmpDir, "comment", word, "accidental stray text")
+			if err == nil {
+				t.Fatalf("expected non-zero exit for 'bd comment %s ...', got stdout=%q stderr=%q", word, stdout, stderr)
+			}
+			combined := stdout + stderr
+			// Must be refused as an id/reserved word — but, unlike "list"/"add",
+			// must NOT claim it is a misplaced "bd comments" subcommand: there is
+			// no "bd comments rm" or "bd comments delete".
+			if !strings.Contains(combined, "not a valid issue id") {
+				t.Errorf("expected a not-a-valid-issue-id refusal, got stdout=%q stderr=%q", stdout, stderr)
+			}
+			if strings.Contains(combined, "misplaced") {
+				t.Errorf("message falsely implies %q is a misplaced \"bd comments\" subcommand (it isn't one), got stdout=%q stderr=%q", word, stdout, stderr)
+			}
+
+			// The regression check: the bystander issue must have received NO
+			// comment.
+			commentsOut := runBDInProcess(t, tmpDir, "comments", fullID, "--json")
+			trimmed := strings.TrimSpace(commentsOut)
+			if trimmed != "[]" && trimmed != "null" {
+				t.Fatalf("expected no comments on bystander issue %s after rejected 'comment %s', got: %s", fullID, word, commentsOut)
+			}
+		})
+	}
+}
+
+// TestCLI_CommentAbbreviatedIDRejectedWithTruthfulMessage is the CLI-level
+// regression test for steveyegge's PR #5393 review item (c): a leading-prefix
+// abbreviation of a REAL issue's id is refused on comment writes (exact-only
+// policy, unchanged), but the refusal message must not claim the issue does
+// not exist — it does, the abbreviation was just refused. Before this fix,
+// resolveAndGetIssueForMutationExact surfaced ResolvePartialIDExact's plain
+// "no issue found matching %q", which is false in exactly this case; every
+// other command family (show/close/update) still accepts the same
+// abbreviation, so a truthful, distinguishing message matters here more than
+// most refusals.
+func TestCLI_CommentAbbreviatedIDRejectedWithTruthfulMessage(t *testing.T) {
+	t.Parallel()
+
+	tmpDir := setupCLITestDB(t)
+
+	out := runBDInProcess(t, tmpDir, "create", "Needs an exact id on comment", "-p", "1", "--json")
+	jsonStart := strings.Index(out, "{")
+	if jsonStart < 0 {
+		t.Fatalf("No JSON found in create output: %s", out)
+	}
+	var issue map[string]interface{}
+	if err := json.Unmarshal([]byte(out[jsonStart:]), &issue); err != nil {
+		t.Fatalf("Failed to parse create JSON: %v\nOutput: %s", err, out)
+	}
+	fullID := issue["id"].(string)
+	if len(fullID) < 4 {
+		t.Fatalf("test setup: generated id %q too short to abbreviate", fullID)
+	}
+	abbrev := fullID[:len(fullID)-1]
+
+	// Sanity check: the same abbreviation still works on a READ path (show),
+	// proving this is a real, resolvable abbreviation and not an accident of
+	// the fixture — the gap this test guards is specific to comment WRITES.
+	showOut, showErr, err := runBDInProcessAllowError(t, tmpDir, "show", abbrev, "--json")
+	if err != nil {
+		t.Fatalf("fixture sanity check failed: 'bd show %s' unexpectedly failed: %v\nstdout: %s\nstderr: %s", abbrev, err, showOut, showErr)
+	}
+	if !strings.Contains(showOut, fullID) {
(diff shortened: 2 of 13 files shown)


steveyegge:
**Fable adversarial paired review (bee, beads-lane steward; reviewed together with #5369) — VERDICT: MERGE-AFTER-FIXES, rework on top of #5369 (just merged, 6ec150f8b985).**

Your exact-resolution core is net-new and is the actual root-cause fix — worth landing. But the two PRs textually conflict (both insert a guard block in `cmd/bd/comment.go`), and #5369's version of the shared guard is better placed (Args-level: fires pre-store-open and uniformly across backends; your RunE guard runs after `PersistentPreRunE` opens the store, so `bd comment list "text"` with no `.beads/` dies with 'no beads database found' instead of the hint). Required before merge:

**(a) Rebase onto merged #5369** and delete the duplicate RunE reserved-word guard; fold your extra words (`rm`/`delete`) into #5369's Args validator — noting its error text ('misplaced `bd comments` subcommand') is false for words that aren't real subcommands, so adjust the message or the list.

**(b) MAJOR — the proxied-server path is untouched by the exact-match fix.** RunE dispatches to `runCommentProxiedServer` before any resolution; the server resolves the raw id, so in proxied mode any token that prefix-matches an unrelated hash still silently writes to the wrong bead. With the proxied cutover imminent here, the root-cause fix would hollow out exactly when the fleet adopts the mode. Extend exact matching to that path, or document/test the gap explicitly and file the server-side follow-up.

**(c) MAJOR — this deliberately breaks the documented short-id contract for one command, with a lying error.** `docs/core-concepts/hash-ids.md` recommends short references with no read/write carve-out; post-PR `bd comment bd-a1b2` fails while close/update/show all still resolve it, and the failure reads 'no issue found matching' — factually false. As maintainer I'm fine with exact-only on comment writes as a policy (it is the correct trade for an agent fleet), but it needs: a truthful, actionable error ('id abbreviations are not accepted on comment writes; use the full id from `bd show …`'), docs + CHANGELOG updates, and CLI-level tests pinning both the refusal and the message.

**(d)** Add CLI-level tests for the abbreviation refusal end-to-end and the guard through cobra dispatch (currently pure-function only).

Good work worth calling out: the fake-store test that first reproduces the old collision before asserting the fix is genuinely good test design, and the `allowAbbrev` threading is provably behavior-preserving at every pre-existing call site.

CI note: your red (`CI Gate`, `Check doc flags freshness`, `PR Policy`) is neither your diff nor the 8/5-6 infra outage — three engdocs crossed the 90-day staleness threshold on ~8/06, failing the gate for every PR run since. I'm fixing that on main; your substantive shards (all 35 Dolt shards, conformance, smokes) passed.

bee-ghosttrack:
CI triage (bee, maintainer seat): the only red is `Check doc flags freshness` complaining about 91-day-old reviewed dates on docs this PR does not touch; that gate passes on current main (the dates were bumped since this head ran). The branch is CONFLICTING with main — rebase and the red goes away with it.

athosmartins:
@steveyegge Addressed (b), (c), (d) from the 08-07 review in aec6de4ad (on top of the 09-07 rebase that resolved (a)). Requesting re-review.

**(c) MAJOR — false "no issue found" on a real abbreviated id.** `ResolvePartialIDExact` couldn't distinguish "genuinely no such issue" from "found it, but only via abbreviation, which exact-only mode refuses" — both collapsed to the same `no issue found matching %q`. Added `ErrAbbreviatedIDNotAllowed`, a sentinel wrapped only when the input would have resolved via leading-prefix abbreviation (tracked separately from the real `matches` slice, never used to resolve). `bd comment`'s RunE checks `errors.Is` against it and returns `id abbreviations are not accepted on comment writes; use the full id from \`bd show <id>\`` instead. Updated `docs/core-concepts/hash-ids.md` with the read/write carve-out and added a CHANGELOG entry.

**(b) MAJOR — proxied-server path.** Traced it before extending anything: `resolveCommentTargetProxied` resolves via `workapi.GetIssueOrWisp` with the raw id, which bottoms out in `issueSQLRepositoryImpl.Get`'s plain `WHERE id = ?` (SQL exact equality) — no abbreviation-matching layer anywhere in that chain, unlike the embedded path's `utils.ResolvePartialID`. So comment writes over the proxied server were already exact-id-only before this PR; there's no gap for the fix to hollow out of. Took your own fallback ("document/test the gap explicitly") — added `comment_write_never_resolves_abbreviated_id` in `comment_proxied_integration_test.go` as a regression guard, with the full trace in its doc comment so this doesn't need re-deriving later.

**(d) CLI-level tests.** Added, in-process `rootCmd.Execute()` dispatch (no store mocking):
- `TestCLI_CommentRmDeleteReservedWordsRejected` — the two reserved words the #5369-inherited tests don't cover (only "list"/"add" have hand-written cases; "rm"/"delete" fall through to the generic message and had zero CLI-level coverage).
- `TestCLI_CommentAbbreviatedIDRejectedWithTruthfulMessage` — the new (c) message end-to-end, plus a same-abbreviation sanity check against `bd show` proving it's a real, resolvable abbreviation and not a fixture accident.
- Docker-free unit coverage too: `TestResolvePartialIDExact_AbbreviationRefusalIsDistinguishableFromNotFound` in `internal/utils`, pinning that the new sentinel and a genuine not-found stay distinguishable.

**Incidental, same class as (c):** `checkCommentIDNotReservedWord`'s message claimed every reserved word is a "misplaced bd comments subcommand" — true for list/add, false for rm/delete (no such subcommand exists; bd has `delete` as a top-level command and `rm` as an alias elsewhere, e.g. `dep rm`). Since I was already touching this function for (c)'s message-truthfulness fix, corrected it to say "reserved word" generically rather than leave a second, adjacent instance of the same defect class you flagged.

**Testing:** `go vet` (plain and `-tags=integration`) clean on `cmd/bd` and `internal/utils`. `go test ./internal/utils/...` passes, including the new sentinel test. `go test ./cmd/bd/.` (non-integration tag) passes, including both new reserved-word subtests. The new CLI-level and proxied-server tests need a Dolt test container this environment doesn't have — same limitation your review already accepted for this PR's other Dolt-backed tests; they compile clean under `-tags=integration` and skip the same way the pre-existing suite does.

bee-ghosttrack (changes requested):
**Fable adversarial review (bee, maintainer seat) — R1 at aec6de4ad — VERDICT: REQUEST-CHANGES, narrow.** Items (b), (c), (d) from the 08-07 review are all substantively addressed and each holds under mutation probes (five run, four caught; the one no-catch is a MINOR below). The block is one thing: the new docs and CHANGELOG assert a safety property for `bd comments add` that the code does not have — the same defect class item (c) was filed about, moved from the error string into the documentation. Independent Opus pass, evidence-cited; probes reverted, worktree clean.

## (b) Proxied-server path — VERIFIED, and the new test is NOT vacuous

Chain traced exact at every hop: `cmd/bd/comment.go:148-150` → `runCommentProxiedServer` → `cmd/bd/comments_proxied_server.go:148-166` `resolveCommentTargetProxied` → `workapi.GetIssueOrWisp(…, id)` with the raw id (`internal/workapi/detail.go:84-101`, `:380-386`) → `internal/storage/domain/issue.go:426-454` → `internal/storage/domain/db/issue.go:551-566` `SELECT … WHERE id = ?`. No `ResolvePartialID`, no `LIKE`, no `wisp-` infix stripping, no server-side resolve endpoint (the CLI runs the query itself via `proxiedOpenReadUOW`, `show_proxied_server.go:62-71`). Proxied `bd show` resolves the same way, so the modes stay consistent.

The reserved-word guard runs on the proxied route too: it is the cobra `Args` validator (`comment.go:122` → `validateCommentArgs`, `:70-104`), so it fires before `RunE`, before `usesProxiedServer()`, before the store opens. Item (a)'s duplicate RunE guard is gone — exactly one guard remains.

`comment_write_never_resolves_abbreviated_id` (`comment_proxied_integration_test.go:219-271`) has a real collision: `abbrev := target.ID[:len-1]`, a strict leading prefix of an existing issue, with a bystander, asserting refusal AND that neither issue got a comment. Probe: injecting `utils.ResolvePartialID` into `resolveCommentTargetProxied` → **caught** (`:251: bd comment pa-zd unexpectedly succeeded in proxied mode`).

Note for any future generalisation: "proxied mode is exact-only" is true of the comment/show chain, not proxied mode generally — `mol_proxied_server.go` (11 sites), `promote_proxied_server.go:47`, `swarm_proxied_server.go:47,114`, `wisp_proxied_server.go:48` all call `utils.ResolvePartialID`.

## (c) Truthful error — VERIFIED on the local tier, two gaps

Sentinel `internal/utils/id_parser.go:32`; `abbrevOnly` tracked separately from `matches` (`:184`), never resolved to; issue-hash branch `:207-219`, wisp-hash branch `:252-258`; the `%w` wrap at `:275` sits under `!allowAbbrev`, so `ResolvePartialID` (`:66-68`) and `ResolvePartialIDs` (`:320-330`) can never emit it — no leak. RunE maps it at `comment.go:159-165`. Probes: `%w`→`%v` at `:275` → **caught** (`id_parser_fake_test.go:143`); RunE `errors.Is` target swapped to `ErrAmbiguousID` → **caught**, and the raw sentinel text then leaks to the user, which is the right failure.

- **Gap 1 (MINOR):** the routed tiers swallow the sentinel. `resolveAndGetIssueForMutationExact` (`cmd/bd/routed.go:155-173`) discards `prefixErr`/`autoErr` (`:162`, `:168`) and returns the local error (`:172`), so a cross-rig abbreviation (`bd comment hr-8wn "x"` with `hr-8wn3k9` in another rig) still prints exactly the message item (c) was about: `no issue found matching "hr-8wn"`. Prefer the routed error when `errors.Is(…, ErrAbbreviatedIDNotAllowed)`.
- **Gap 2 (MINOR):** the wisp branch of the sentinel is unpinned. Making `id_parser.go:256` a no-op → `go test ./internal/utils/ -run ResolvePartialID` still passes (Docker suites included, 261s) — **no catch**. Every wisp-path test asserts only `err != nil` (`id_parser_exact_test.go:147-150`, `id_parser_fake_test.go:102-104`), never `errors.Is`. That is the incident's own shape (`"list"` → `hq-wisp-list3t0`), free to regress silently to the false not-found. One `errors.Is` at `id_parser_fake_test.go:102` closes it.

## (d) CLI-level tests — VERIFIED as tests; they execute in no CI job

Real in-process cobra dispatch (`cli_fast_test.go:176-180`, `:1203+`), no store mocking, each asserting target and bystander got no comment. Probe: `checkCommentIDNotReservedWord` returning nil for `rm`/`delete` → **caught**, both subtests. The `rm`/`delete` message no longer claims a misplaced `bd comments` subcommand and the new test pins that (`cli_fast_test.go:1853`).

**MINOR:** `cli_fast_test.go` is `//go:build cgo && integration` plus a Dolt container. No PR job applies `-tags=integration` to `./cmd/bd` (`pr.yml` uses it only for `./scripts`; `pr-risk.yml` builds with `gms_pure_go` only), and the main-branch integration shard runs with `BEADS_TEST_SKIP=dolt` (`main.yml:843`) so `testDoltServerPort==0` and they skip. Pre-existing infra, shared with ~100 container-backed tests in that file — but "pin it" is not achieved in CI. I ran them for real locally instead (below).

## MAJOR

**M1. CHANGELOG and hash-ids.md claim `bd comments add` is guarded. It is not.**

- `CHANGELOG.md:188`: "`bd comment`/`bd comments add` refuse an abbreviated id and a reserved word instead of silently resolving one to the wrong issue".
- `docs/core-concepts/hash-ids.md:124`: "`bd comment <id> "text"` (and its `bd comments add` twin) … accepts only a full, exact id, never an abbreviation."

Code: `commentsAddCmd` (`cmd/bd/comments.go:168-243`) has `Args: cobra.MinimumNArgs(1)` (`:179`), no validator, and resolves through the abbreviation-tolerant `resolveAndGetIssueForMutation` (`:216`). Executed proof at this head: the pre-existing `TestCLI_CommentsAddShortID/PartialIDWithCommentsAdd` (`cli_fast_test.go:1585-1615`) PASSES while asserting that `bd comments add <4-char partial> "text"` **succeeds**.

Why it matters beyond hygiene: the new refusal message (`comment.go:46-55`) tells the caller who just typo'd `bd comment list <id>` to run `bd comments add <issue-id> "text"` — routing them onto the one comment-write path that still fuzzy-resolves. The incident this PR exists to fix remains reachable there. The PR body's "How" is accurate ("this PR only tightens `bd comment`"); the shipped docs are not.

Fix, either: (i) delete the `bd comments add` claims from both files; or (ii) better, extend the guard — `commentsAddCmd.Args` → a validator calling `checkCommentIDNotReservedWord`, `:216` → `resolveAndGetIssueForMutationExact` plus the `errors.Is` mapping, ~6 lines, and the docs become true. I'd take (ii).

## MINOR

- m1. Routed tiers swallow the sentinel (`routed.go:160-172`) — see (c) gap 1.
- m2. Wisp-branch `abbrevOnly` tracking (`id_parser.go:256`) unpinned — see (c) gap 2.
- m3. The two new CLI tests run in no CI job — see (d).
- m4. `resolveAndGetIssueForMutationExact` (`routed.go:155-173`) and `resolveAndGetFromStoreExact` (`:125-146`) are near-verbatim copies of `resolveAndGetIssueWithRoutingAccess` (`:70-95`) / `resolveAndGetFromStore` (`:98-118`). The PR already threads `exact bool` through two helpers; threading it through these two removes the copies.

## Nits

- `commentReservedIDWords` holds `list`/`add`, but `validateCommentArgs:74-94` returns before `checkCommentIDNotReservedWord` sees them — those entries are reachable only from the unit test.
- `LIST` (uppercase) is deliberately allowed (`comment_reserved_word_test.go:33`) on the "real ids are lowercase" reasoning that was rejected for `rm`/`delete`. Harmless under exact-only.
- The message suggests `bd show <abbrev>`; for an *ambiguous* abbreviation the sentinel now fires instead of `ErrAmbiguousID`, and that `bd show` will itself error as ambiguous.
- Proxied mode is stricter than `ResolvePartialIDExact`: no prefix normalisation at all, so `bd comment 165vq "x"` (bare full hash, exact in embedded) fails proxied with "issue 165vq not found". Pre-existing, but a real read/write-contract divergence hash-ids.md does not mention.
- The new CLI tests call `t.Parallel()` while `runBDInProcess` serialises on `inProcessMutex` (`cli_fast_test.go:151`), against the file's own stated convention (`:222`). My first parallel run failed four tests (the two new plus pre-existing `TestCLI_CommentListMisplacedSyntax`/`…AddMisplacedSyntax`) with `embeddeddolt: open db: … openat LOCK: invalid argument`; all pass serially. Pre-existing pattern.

## Behavior preservation — CONFIRMED

Every pre-existing caller passes `exact=false`: `close.go:743`, `routed.go:81`, `:220` (prefix routing); `routed.go:90`, `:404` (auto routing). Only the new tier passes `true` (`routed.go:162,168`), reached only from `bd comment` RunE (`comment.go:154`). `ResolvePartialIDExact` has one non-test caller (`routed.go:126`); `ResolvePartialID` is the one-line wrapper (`id_parser.go:66-68`). `bd close`, `bd update` (`update.go:465`), `bd label` (`label.go:59,408`) unchanged.

## What I ran

Worktree at `aec6de4ad`, `source ./.buildflags`, go1.26.5 darwin/arm64, Docker available (the container suites really ran, no skips).

| command | result |
|---|---|
| `go build ./...` | rc=0 (121s) |
| `go vet ./...` | rc=0 |
| `go vet -tags=integration ./cmd/bd/ ./internal/utils/` | rc=0 |
| `gofmt -l` on the 9 changed .go files | clean |
| `go test ./internal/utils/... -count=1` | ok 452.2s |
| `go test ./internal/utils/ -run ResolvePartialIDExact -count=1 -v` | 5/5 PASS, 0 skip (`_Wisp` 10.1s, `_RejectsAbbreviationAcceptsExact` 18.0s, `_ReservedWordDoesNotCollideWithWisp` 7.7s, both fake-store tests) |
| `go test -tags=integration,gms_pure_go ./cmd/bd/ -run 'TestCLI_Comment…' -count=1 -v -parallel 1 -p 1` | PASS: `…AbbreviatedIDRejectedWithTruthfulMessage` 13.9s, `…RmDeleteReservedWordsRejected/rm` 11.2s `/delete` 4.0s, `TestCLI_CommentListMisplacedSyntax` 25.0s; ok 218.6s |
| `BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd/ -run TestProxiedServerComment -count=1 -v -p 1` | PASS 3/3 incl. `comment_write_never_resolves_abbreviated_id` 14.1s |

Probes (each reverted with `git checkout -- <file>`; `git status --porcelain` empty after the last):

| # | mutation | result |
|---|---|---|
| P1 | `comment.go:159` `errors.Is` target → `ErrAmbiguousID` | CAUGHT (truthful-message CLI test) |
| P2 | `id_parser.go:275` `%w` → `%v` | CAUGHT (`…AbbreviationRefusalIsDistinguishableFromNotFound`) |
| P3 | `id_parser.go:256` wisp `abbrevOnly` append → no-op | NO CATCH → m2 |
| P4 | `checkCommentIDNotReservedWord` nil for `rm`/`delete` | CAUGHT (both subtests) |
| P5 | inject `utils.ResolvePartialID` into `resolveCommentTargetProxied` | CAUGHT |

## CI (head aec6de4ad)

119 pass, 2 fail, 1 skipping. The fail is `PR Core (wrapper timing)` (`CI Gate / Required` only aggregates it): `panic: test timed out after 10m0s`, `FAIL cmd/bd 600.875s`, the test at the deadline being the unrelated `TestWhereCommand_ReadsPrefixFromEmbeddedStore`. `PR Core` runs `go test -race -short -skip '^TestEmbedded' ./...` with no integration tag (`scripts/ci/pr-core.sh:23-24`), so none of this PR's new CLI/proxied tests are even compiled into it; same-hour green runs measured cmd/bd at 541.6s and 387.0s, so 600.9s is the known near-the-wire package timeout, not this diff — but it is a required check and needs a green re-run before merge. All 20 Embedded Dolt Cmd shards and the Proxied Dolt Cmd shards are green (shard 14/15 carries `TestProxiedServerComment`); `Check doc flags freshness` now passes. No conflict: `mergeable: MERGEABLE`, `git merge-tree --write-tree origin/main HEAD` clean; the 09-03 CONFLICTING note was resolved by the 09-07 rebase.

athosmartins:
Addressed R1 (aec6de4ad → a639234a0). Item by item:

## M1 (the block) — fixed by extending the guard, not softening the docs

Took option (ii) as suggested. `commentsAddCmd` now has an `Args` validator (`validateCommentsAddArgs`, mirrors `comment`'s `validateCommentArgs` minus the singular/plural-confusion special cases) wired in, and the embedded-path resolution switched from `resolveAndGetIssueForMutation` to `resolveAndGetIssueForMutationExact` with the same `errors.Is(ErrAbbreviatedIDNotAllowed)` truthful-message mapping `comment` already has.

The proxied path needed no change — `addCommentProxied`/`resolveCommentTargetProxied` is shared by both `comment` and `comments add` already, so it was already exact-only there; the gap was embedded-path only.

CHANGELOG.md:188 and hash-ids.md:124 are unedited — the claim they make ("`bd comment`/`bd comments add` refuse an abbreviated id and a reserved word") is now true for both commands, so there was nothing to correct.

Updated the one pre-existing test that asserted the old, now-removed behavior: `PartialIDWithCommentsAdd` used to assert a genuine abbreviation *succeeded* on `comments add` (true when written for #1070, no longer the contract). It now asserts refusal with the truthful message, plus a sanity check that the same abbreviation still resolves on `bd show` and a regression check that no comment landed. Added `TestValidateCommentsAddArgs` (pure, no store/Docker) mirroring `TestValidateCommentArgs` for the Args-level wiring specifically.

## The MINOR (mutation probe not caught) — fixed

`TestResolvePartialIDExact_FakeStore_ReproducesReportedIncident`'s `"list"` assertion only checked `err == nil` — P3 (making the wisp-branch `abbrevOnly` append at `id_parser.go:256` a no-op) went uncaught because the generic "no issue found" fallback still errors either way. Strengthened it to also assert `errors.Is(err, ErrAbbreviatedIDNotAllowed)`, which is exactly the incident's own shape reproduced in the wisp branch.

Left m1 (routed-tier sentinel swallowing), m3 (integration tests run in no CI job — pre-existing infra, not this diff's to fix), and m4 (helper de-duplication) as filed; none were part of what blocked merge and this round stayed narrow to M1 + the one MINOR.

## CI diagnosis (`PR Core (wrapper timing)` / `CI Gate / Required`)

Confirmed pre-existing and unrelated to this diff before touching anything, independently of R1's own read (which matches):

- `scripts/ci/pr-core.sh:23-24` runs `go test -race -short -skip '^TestEmbedded' ./...` with **no `-timeout` override**, so `cmd/bd` (a large package, under `-race`) runs under Go's default 10-minute per-package timeout.
- Checked same-hour runs on other PRs (`fix/ga-bq3w5-undefer-defer-until`, `fix/widen-default-lease-ttl`, `fix/ga-y6gjv-backup-gc-prune`): this exact job's total wall time is already 11–13.5 minutes on green runs — i.e. already at the edge even without this diff.
- The goroutine dump names the actively-running test at the 10-minute mark as `TestWhereCommand_ReadsPrefixFromEmbeddedStore` — unrelated to comment.go/comments.go.
- None of this PR's new CLI-level tests are even compiled into this job (`cli_fast_test.go` is `//go:build cgo && integration`, and `pr-core.sh` doesn't pass `-tags=integration`), so they can't be the cause either.

Needs a green re-run, not a code change. Everything else green (`mergeable: MERGEABLE`, 20 Embedded Dolt Cmd shards + Proxied Dolt Cmd shards green per R1).

Requesting re-review.

bee-ghosttrack (changes requested):
**Fable adversarial review (bee, maintainer seat) — R2 at a639234a0 — VERDICT: REQUEST-CHANGES, narrow.** The code half of M1 is genuinely closed and probe-verified on both paths, and the flagged MINOR moved non-vacuously. One new MAJOR blocks: the regression test written to lock M1's resolver half fails at HEAD in its own setup on every run, so it never executes an assertion. Independent Opus pass, real binary + real Dolt container; all probes reverted. Everything is in this body.

## M1 — code CLOSED

`bd comments add` now runs both guards `bd comment` runs: reserved words via `validateCommentsAddArgs` (`cmd/bd/comments.go:176-181` → `checkCommentIDNotReservedWord`, wired as cobra `Args` at `:196`, so it fires before the `usesProxiedServer()` dispatch at `:225` on both backends); exact resolution via `resolveAndGetIssueForMutationExact` at `:233` with the `ErrAbbreviatedIDNotAllowed` mapping at `:239-247`, byte-identical to `comment.go:158-165`. Proxied `comments add` was already exact-only (`comments_proxied_server.go:87` → `resolveCommentTargetProxied:148` → `workapi.GetIssueOrWisp`), unchanged.

Live probe (fresh embedded DB, prefix `probe`, issue `probe-auj`):

| input | `bd comments add` | `bd comment` |
|---|---|---|
| `au` (genuine prefix; `bd show au` resolves) | rc 1, `id abbreviations are not accepted on comment writes; use the full id from \`bd show au\`` | rc 1, identical string |
| `list` (reserved) | rc 1, reserved-word refusal | rc 1, refusal (singular's hand-written variant) |
| `rm` (reserved) | rc 1 | rc 1, identical string |
| `probe-auj` (exact) | rc 0 | rc 0 |

`bd comments probe-auj --json` = `[]` after the six refusals. No legitimate input newly refused: bare full hash `auj`, child `probe-auj.1`, child bare hash `auj.1` all rc 0 on the add path; `ResolvePartialIDExact` differs from `ResolvePartialID` only in the `allowAbbrev` branch (`id_parser.go:206-219`, `:250-258`). CHANGELOG:181-212 and hash-ids.md:122-145 are now true as written.

## MINOR (m2) — MOVED

`id_parser_fake_test.go:107-111` now asserts `errors.Is(err, ErrAbbreviatedIDNotAllowed)`; mutating the wisp-branch `abbrevOnly` append to a no-op is now caught at `:110`. Closed.

## NEW MAJOR — `TestCLI_CommentsAddShortID/PartialIDWithCommentsAdd` is dead-red at HEAD

`cmd/bd/cli_fast_test.go:1611-1616` truncates to 4 chars only when `len(shortID) > 4`, then `t.Fatalf`s when `shortID == hash`. The second guard is exactly `NOT(len(hash) > 4)`. Fresh test DBs mint 3-char hashes (`min_hash_length` unset; the test's own fixtures are `test-7v0`, `test-9xn`), so the truncation never happens and the Fatalf always fires. Observed twice, unmutated, on a real container:

```
=== RUN   TestCLI_CommentsAddShortID/PartialIDWithCommentsAdd
    cli_fast_test.go:1615: test setup: hash "9xn" too short to truncate into a genuine abbreviation
--- FAIL: TestCLI_CommentsAddShortID/PartialIDWithCommentsAdd (1.24s)
```

Two consequences. The one test that pins the exact-resolver swap never runs its assertions — reverting `comments.go:233` back to `resolveAndGetIssueForMutation` is caught by nothing (mutation P-B, both default and `-tags=integration` builds). And the PR leaves a hard-failing test wherever `cli_fast_test.go` actually executes (cgo + integration + a reachable Dolt); it is invisible in CI only because of the still-open m3 (`setupCLITestDB` skips on `testDoltServerPort == 0`, `cli_fast_test.go:99-103`). Red locally and skipped in CI is worse than the misleading-but-green test it replaced.

**Fix:** derive the abbreviation as a strict leading prefix of the real id regardless of hash length — `fullID[:len(fullID)-1]`, which is what `comment_proxied_integration_test.go:230` already does — or set `min_hash_length` on the fixture DB. Either version also passes the `bd show` sanity check the subtest already performs.

## NEW MINOR — the guard's wiring is untested

`comments_add_test.go` calls `validateCommentsAddArgs` directly and does catch a gutted body (mutation P-C, 4/8 subtests red), but unwiring `Args:` back to `cobra.MinimumNArgs(1)` leaves every test green (P-D). No test in `cmd/bd/**/*_test.go` drives `bd comments add <reserved>` through cobra dispatch. Cheap fix: assert `commentsAddCmd.Args != nil` and invoke via the field, or add a `bd comments add list "text"` case beside the singular's `TestCLI_CommentRmDeleteReservedWordsRejected`.

## Carry-over m1, now reaching a second command

`ErrAbbreviatedIDNotAllowed` is not a not-found error (`isNotFoundErr`, `routed.go:23-31`) and `resolveAndGetIssueForMutationExact:163,169` discards `prefixErr`/`autoErr`, so a cross-rig abbreviation on `bd comments add` now, like `bd comment`, either skips prefix/auto routing when a local abbrev candidate exists or prints the generic false `no issue found matching "hr-8wn"`. Not a blocker; already filed as m1, and preferring the routed error when `errors.Is(…, ErrAbbreviatedIDNotAllowed)` closes it for both.

## Ran

`source ./.buildflags`, go1.26.5 darwin/arm64, OrbStack Docker. `go build ./...` 0 · `go vet ./cmd/bd/ ./internal/utils/` 0 · `go test ./cmd/bd/ -run 'Comment|Comments' -count=1` 0 (45s; note this compiles none of `cli_fast_test.go`, which is `cgo && integration`) · `go test ./internal/utils/ -count=1` 0 (254s) · `gofmt -l` clean · `go test -tags=integration,gms_pure_go ./cmd/bd/ -run 'Comment' -count=1 -p 1 -parallel 1` **1** (the Fatalf above, 281s). Five mutations reverted; tree clean at a639234a0.

CI diagnosis in your comment matches mine: `pr-core.sh` has no `-timeout`, so `cmd/bd` under `-race` rides the 10m default; needs a green re-run, not a code change.

Fix the fixture abbreviation (and ideally wire-test the guard) and this is an approve.

athosmartins:
@bee-ghosttrack Addressed R2 (aec6de4ad → a639234a0 → **b89cc476f**). Item by item:

## NEW MAJOR — `TestCLI_CommentsAddShortID/PartialIDWithCommentsAdd` dead-red

Fixed. The old logic truncated to 4 chars *only when* `len(hash) > 4`; fresh
test DBs mint 3-char hashes, so `shortID` always equaled the full hash and the
setup `t.Fatalf` fired before any real assertion ran. Now the abbreviation is
derived as a strict, one-character-shorter leading prefix of the real hash —
`hash[:len(hash)-1]` — the same derivation `comment_proxied_integration_test.go`
already uses, correct regardless of fixture hash length. Guard is now
`len(hash) < 2`, which can't be spuriously true against real fixture data.

## NEW MINOR — guard's cobra wiring untested

Added `TestCLI_CommentsAddReservedWordRejectedThroughCobraDispatch`: drives
`bd comments add <reserved-word> "text"` through real `rootCmd.Execute()`
dispatch (not the direct-function call `comments_add_test.go` already has)
for all four reserved words, and asserts the bystander issue receives no
comment. This is what would have caught unwiring `commentsAddCmd.Args` back
to `cobra.MinimumNArgs(1)` (your mutation P-D).

## Not touched (per your guidance)

- Carried-over m1 (cross-rig abbreviation error message) — you marked it "Not
  a blocker."
- CI: `cmd/bd`'s `-race` run against the 10m default (no `-timeout` in
  `pr-core.sh`) — this is the same pre-existing infra flakiness you and I
  both already diagnosed, not something this diff can fix. The `Test
  (macos-latest)` failure on the CI page right now is from the *pre-push*
  run at a639234a0 (`FAIL github.com/steveyegge/beads/cmd/bd 600.579s`,
  timestamped 2026-09-10T17:47-17:51, before this push) — a fresh run is
  already in flight from the push above.

## Ran (this environment — no Docker/testcontainers here, same disclosed
limitation as the aec6de4ad/a639234a0 rounds)

- `gofmt -l cmd/bd/cli_fast_test.go` — clean
- `go vet -tags=integration ./cmd/bd/...` — clean (needed
  `CGO_CPPFLAGS`/`CGO_LDFLAGS`/`PKG_CONFIG_PATH` pointed at the keg-only
  `icu4c@78` install to get past an unrelated CGO preprocessor error, not a
  code issue)
- `go test -tags=integration ./cmd/bd/ -run
  'TestCLI_CommentsAddShortID|TestCLI_CommentsAddReservedWordRejectedThroughCobraDispatch'
  -v -count=1` — PASS, all Dolt-backed subtests SKIP cleanly ("Dolt test
  container not available") rather than panicking. This confirms both tests
  compile and execute correctly up to the container-availability check; the
  actual fixture-hash logic your report exercised only runs where a Dolt
  container exists, i.e. CI or your OrbStack setup — not verified against a
  real container in this environment.
- pre-commit hook: `golangci-lint` — 0 issues

Requesting re-review.

bee-ghosttrack (approved):
R3 at b89cc476. VERDICT: APPROVE. Both R2 blockers are closed and mutation-pinned on a real Dolt container. R3 touches only `cmd/bd/cli_fast_test.go` (+65/-7); no production code moved since a639234a. CI at this head: 121 pass, 1 skipping; the `PR Core (wrapper timing)` race timeout from R1/R2 cleared.

**R2 NEW MAJOR (dead-red fixture) — CLOSED.** `cli_fast_test.go:1618-1623` derives the abbreviation as `hash[:len(hash)-1]`, correct at any fixture hash length. `TestCLI_CommentsAddShortID/PartialIDWithCommentsAdd` now runs its assertions for real (`Full ID: test-vdu, Partial ID: vd`) and passes on a live container. Mutation P-B (`comments.go:233` → `resolveAndGetIssueForMutation`) is now CAUGHT at `:1640` (`got stdout="Comment added to test-afi"`). That was the point of R2's block.

**R2 NEW MINOR (guard wiring untested) — CLOSED.** `TestCLI_CommentsAddReservedWordRejectedThroughCobraDispatch` drives all four reserved words through `rootCmd.Execute()`. Mutation P-D (`commentsAddCmd.Args` → `cobra.MinimumNArgs(1)`) CAUGHT 4/4 at `:1945` while `TestValidateCommentsAddArgs` stays green; exactly the gap R2 named.

**R1 m2 — re-verified closed.** `id_parser.go:256` → no-op is now CAUGHT at `id_parser_fake_test.go:110` (NO CATCH in R1).

Carried, still non-blocking: m1 (`routed.go:162,168,172` discard `prefixErr`/`autoErr`, so a cross-rig abbreviation still prints the generic message; declared untouched, agreed), m3 (`cli_fast_test.go` is `cgo && integration`, compiled by no PR job), m4 (`routed.go:120-173` duplicates `:98-118`/`:70-95`).

New, all minor, none blocking:

1. `cli_fast_test.go:1607-1617` — the comment says this is "the same derivation `comment_proxied_integration_test.go:247` already uses". It isn't: `:247` is `target.ID[:len(target.ID)-1]` (full-id prefix); this is a bare-hash prefix. Both valid; the claim and the pinned line number are not.
2. `cli_fast_test.go:1620` — `len(hash) < 2` lets a 2-char hash produce a 1-char abbreviation, which goes ambiguous rather than abbreviation-refused. `< 3` is the honest floor; today's hashes are 3, so it does not bite.
3. `cli_fast_test.go:1951` — the bystander check cannot fail (a reserved word never prefixes a freshly minted hash), so P-D is caught solely by the string at `:1945`. The comment claims more than the assertion does.
4. The three new/reworked tests (`:1865`, `:1921`, `:1971`) call `t.Parallel()`, against the convention the other 25 tests in the file state in so many words ("Not using t.Parallel() because inProcessMutex serializes execution anyway"). `go test -tags=integration ./cmd/bd/ -run 'TestCLI_Comment'` at default parallelism fails 10 of 11 here on cross-test DB/cwd bleed (`no beads database found`, `no issue found matching test-6fn`). Pre-existing (main's four parallel comment tests fail 1-of-4 on the same command), but this PR triples the surface and these tests are its only behavioral pins. Dropping the three calls costs nothing. Fine as a follow-up.

Ran (worktree at b89cc476, `source ./.buildflags`, go1.26.5 darwin/arm64, Docker up):

| command | rc |
|---|---|
| `go build ./...` | 0 |
| `go vet ./cmd/bd/ ./internal/utils/` | 0 |
| `go vet -tags=integration ./cmd/bd/` | 0 |
| `gofmt -l` on the 11 changed/added files | clean |
| `go test ./cmd/bd/ -run 'Comment' -count=1` | 0 |
| `go test ./internal/utils/ -run ResolvePartialID -count=1` | 0, no skips |
| `go test -tags=integration ./cmd/bd/ -run '<8 comment CLI tests>' -count=1 -p 1 -parallel 1` | 0 |
| `BEADS_TEST_PROXIED_SERVER=1 go test ./cmd/bd/ -run TestProxiedServerComment -count=1 -p 1` | 0, 3/3 |
| same `-run 'TestCLI_Comment'` at default parallelism | 1 (item 4) |

Probes P-B, P-D, P3 all CAUGHT; all reverted, tree clean.

PR #5394, opened 2026-08-07
docs: re-review and bump the three 90-day-stale freshness markers

steveyegge:
The doc-flags freshness gate went red on every branch today: azure-devops.md, json-schema.md, and otel-data-model.md all carried `Last reviewed: 2026-05-08` — exactly 91 days old against the 90-day fence. This blocks unrelated PRs (e.g. #4839, whose review is otherwise complete).

Each doc got a genuine re-review against its Freshness source paths on main (one fresh-context reviewer per doc), not a blind date bump. Real drift was found and fixed in all three — details in the commit message. Highlights:

- **json-schema.md** taught `jq` recipes that fail against current `bd show --json` (it returns a top-level array, pinned by contract test); envelope `pagination`, show's `revision`, import's `unchanged`, and the error-envelope shape were missing or wrong.
- **azure-devops.md** was missing proxied-server behavior, `--parent`/`--project` flags, and the yaml-only PAT note. The review also surfaced a real source bug — the cmd-level config reader can't see a yaml-only `ado.pat`, so the documented Quick Start fails on main — filed as **bd-wvf77** rather than papered over in the doc.
- **otel-data-model.md** predated the env-var refactor (#3859), the facade-wave storage role spans, and seven newer Dolt metrics; appendix re-pinned to current main.

`scripts/check-doc-freshness.sh` passes locally with these changes.

### DIFF
--- a/docs/integrations/azure-devops.md
+++ b/docs/integrations/azure-devops.md
@@ -3,11 +3,16 @@ title: Azure DevOps (ADO) Integration Configuration
 description: Configuration reference for bd ado sync, which bidirectionally syncs beads issues with Azure DevOps work items
 ---
 
-Last reviewed: 2026-05-08
+Last reviewed: 2026-08-07
 
 Freshness source: `cmd/bd/ado*.go` and `internal/ado/`.
 
 This guide covers all configuration options for the `bd ado sync` command, which synchronizes beads issues with Azure DevOps work items.
+
+**Proxied-server mode:** `bd ado sync`, `bd ado status`, and `bd ado projects`
+are not supported when bd is connected to a proxied server (error:
+`ado sync is not supported in proxied-server mode`); run them from a
+workspace with direct database access.
 
 ## Quick Start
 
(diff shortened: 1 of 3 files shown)


steveyegge:
**Fable adversarial review verdict: FIX_THEN_MERGE — one factual error, now fixed in ae6f5052c.**

Fresh-context reviewer, charged to refute. The diff is exactly the three doc files, one commit atop main (= the SHA the otel appendix pins, so every line-number pin was verifiable against the audited tree). Freshness gate passes (`check-doc-freshness.sh` → PASSED, all three markers 0 days).

**Finding (should-fix, fixed):** the new `bd.claim_verify_lost_total` row enumerated `op` as `claim | unclaim` — incomplete; the recording sites the appendix itself cites also emit `guarded-update` (claim_verify.go:113) and `ready-claim` (issues.go:439). **Nit (fixed):** the AI retry-policy sentence now carries its compact-only qualifier (find_duplicates makes a single unretried call). A suspected phantom `--parent` flag was a false alarm — registered via the shared sync_flags.go helper, confirmed against a built binary's `bd ado sync --help`.

Coverage: every changed azure-devops.md claim verified verbatim against ado.go/yaml_config.go (error strings, proxied refusals, yaml-only PAT); every changed json-schema.md claim against output.go/errors.go/contract tests + corpus goldens (array shapes, pagination, revision, error envelope incl. stdout paths); every otel appendix line pin spot-checked exact, plus behavioral checks (console-only traces, hard opt-in, doltSpanAttrs, never-recorded metrics confirmed by repo-wide grep).

Marker bumps are honest as of ae6f5052c. Merging on green CI.

PR #5398, opened 2026-08-07
Four DependencyEditor contract cases for state the contract never seeded

julianknutsen:
Slice 3 of a test-deduplication pass against the conformance contracts. **It deduplicated almost nothing, and that is the finding.**

## Why there was nothing to delete

`DoltStore.AddDependency` is still a **live production seam outside the DependencyEditor role** — 15 non-test `cmd/bd` files call it directly (`state.go`, `swarm.go`, `gate.go`, `duplicates.go`, `mol_bond.go`, the tx variants in `create_atomic.go`/`graph_apply.go`, …) — and it is a *different body for the routing half*, computing `isCrossPrefix` and wisp routing itself before reaching the shared `issueops` code.

Probe: blank the routing condition at `dolt/dependencies.go:38`. The ad-hoc test **fails**; all 31 `TestDependencyEditor` contract cases **pass**. None of the 14 candidate files calls a role accessor at all.

So these tests are not duplicates of the contract. They are the only coverage a path with 15 live callers has. Deletion here waits on those callers converging onto the role; that is not this PR.

## What the slice produced instead — 4 cases × 3 legs

Each closes a gap found by a seeded-state audit, and each was verified twice: before, the ad-hoc failed while the contract passed; after, the contract fails the mutant.

- **`WritesTheTargetIntoItsTypedColumn`** — every existing case reads the target through `COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external)`, so **no case could see which column held it**. Only `depends_on_issue_id` carries `fk_dep_issue_target`. A wisp target filed as external was invisible to all 27. Reads all three columns and asserts the whole 0/0/1 triple, so a double-write fails too.
- **`RefusesBlockingEdgeAcrossAWispHierarchy`** — the guard reads a UNION of both dependency tables; every seeded case used issues only. A wisp gated on its own ancestor never becomes ready.
- **`RefusesACycleThroughAParentChildHop`** — `cycle_detector_contract.go` states parent-child is outside the report's walk "which the ADD-time gate does walk". Nothing pinned that half. Carries an acyclic control, so a body rejecting every hierarchy also fails.
- **`RefusesASamePlaneEdgeClosingACrossPlaneCycle`** — the existing cross-plane case's own edge crosses planes; here only the loop's interior does. That distinction was a shipped defect.

## Deleted: 2 functions

`TestGetDependenciesWithMetadata` and `TestGetDependentsWithMetadata` were unconditional `t.Skip` bodies with no assertions — green forever, including against the limitation they were named for.

## Kept, deliberately

- **All 13 `is_blocked` tests.** `is_blocked` is a derived *and persisted* column and no contract reads it. Deleting those raw-row assertions because "BlockingAnnotator covers blocking" is the corrupted-table trap: reading back through the role is exactly the check that passes on a corrupted table.
- **The GH#1495 type matrix.** Three reject rows are individually weaker than the contract, but every contract seed is `TypeTask`, so the matrix's "the guard is structural, not type-based" claim cannot port without reshaping a helper 31 cases share.
- **`GetDependentRecords`.** The inbound direction is deliberately not EdgeReader's, and is public API with no role.

## Review

Three adversarial review agents ran over this branch — coverage-loss, port-integrity, and structural-residue lenses. All three cleared it: no coverage lost, no dead helpers, no orphaned citations, shard manifest consistent. Port integrity found the slice's mutation evidence had covered only 2 of 3 legs for two cases (uow runs different bodies — `pickDepTargetColumn`, and a second `CycleThroughEdges` gate); it closed both gaps itself with uow-specific mutants that turn those legs red.

## Verification

`go build ./...`, `go vet`, `golangci-lint run` clean. `DependencyEditor` contract green on dolt, uow and embeddeddolt; all four new cases confirmed executing on all three legs (uow as subtests of the table-driven runner).

### DIFF
--- a/backend/conformance/dependency_editor_contract.go
+++ b/backend/conformance/dependency_editor_contract.go
@@ -3,6 +3,7 @@ package conformance
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/steveyegge/beads/internal/types"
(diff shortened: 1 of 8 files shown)


julianknutsen:
## Red-team review council

Three adversarial agents over this branch, each briefed to break it rather than confirm it: **coverage-loss**, **port-integrity**, **structural-residue**. Every mutation run in an isolated worktree via a harness that refuses to proceed if the baseline is not green or if `-run` matches zero tests (`go test -run NoSuchTest` exits 0, so a typo'd regex reads as PASS forever).

### Verdict: clean on all three lenses. One evidence gap found and closed by the council itself.

**Port integrity — the slice's evidence covered only 2 of 3 legs on two cases.**

`WritesTheTargetIntoItsTypedColumn` and `RefusesASamePlaneEdgeClosingACrossPlaneCycle` were mutation-proven on dolt and embeddeddolt, but the uow leg runs *different bodies*: classification there is `pickDepTargetColumn` (`domain/db/dependency.go:53`), untouched by the slice's mutant, and the cycle path has a second net (`CycleThroughEdges` repo gate, `domain/db/dependency.go:858`) that a `depTables[:1]` mutant cannot reach because the repo passes one-table slices. Both legs showed dolt RED / uow GREEN — i.e. the ports were real but under-evidenced, and a future probe run against the wrong body would have concluded they were dead.

The council closed both gaps with uow-specific mutants: both legs now go red. `RefusesBlockingEdgeAcrossAWispHierarchy` and `RefusesACycleThroughAParentChildHop` run fully shared bodies and went red at dolt and uow together.

**Cannot-fail check — none.** All four cases go red under a mutation of their claimed subject. Under hostile reading of name-vs-assertion: `WritesTheTargetIntoItsTypedColumn` reads all three typed columns individually and asserts the whole 0/0/1 triple, so a double-write also fails — it does *not* read through the `COALESCE` its siblings use. `RefusesACycleThroughAParentChildHop` carries a genuine acyclic control, so a body rejecting every hierarchy fails too.

**Redundancy — none.** Every other dependency case reads targets through `COALESCE`; the existing hierarchy, cycle and cross-plane cases are durable-plane, blocking-only and crossing-edge respectively. Each port is the stated complement, not a restatement.

**Coverage loss — none.** Both deleted functions were unconditional bare `t.Skip` bodies with zero assertions: cannot-fail by construction, green forever including against the limitation they were named for. The `_NoResults` case still runs. Everything else in the commit range is comment-only or purely additive.

**Structural — clean.** Full package-level reference-count diff for dead helpers (base → HEAD): only expected drops, no orphans; `mustAddBlockingDependency`/`assertCycleError` still called. Test binaries compile. No `file.go:line` citation anywhere references a deleted file, and the leaf-doc citations the new contract prose added all resolve against the cited lines. Shard manifest consistent — neither deleted function appears in it.

### One thing the council flagged that this PR does not fix

The ledger's body-location claims are wrong in two places (it cites `internal/storage/domain/issue.go` where the real path is `domain/db/issue.go`). Harmless to the code, but it is what sent the slice's mutants at the wrong legs — worth correcting before the next slice reuses those citations.

### Not a finding, but the reason this PR adds rather than removes

`DoltStore.AddDependency` has 15 live non-test `cmd/bd` callers and a different body for the routing half. Blanking `dolt/dependencies.go:38` fails the ad-hoc test and passes all 31 contract cases. These tests are not duplicates — they are the only coverage that path has. Deletion here waits on those callers converging onto the role.

PR #5399, opened 2026-08-07
Retire embeddeddolt seam tests the contracts answer for, keep what they do not

julianknutsen:
Slice 1 of a test-deduplication pass against the conformance contracts. Net **−295 LOC** across 16 files: 8 test functions deleted, 4 files removed, 3 cases ported into contracts, and — after adversarial review — one residue kept that the deletion should not have taken.

## Deleted, with evidence

14 assertion-targeted mutations, each showing the contract goes red on the same break: blocker refusal, open-children refusal, `CheckVersionInTx` (close *and* update), assignee guard, status guard, LabelPatch ordering, ParentID replace-all, configured-done resolution, priority-range validation.

Files removed: `embeddeddolt/{close_issue_checked,update_issue_checked,update_guards,issue_operations_validation}_test.go`.

`update_guards_test.go` was **partly cannot-fail**: its header claimed to pin `--if-assignee ''` as a real "expected unassigned" guard, but its only `ptr("")` leg runs against a fresh unassigned issue and asserts success — it cannot fail if `ptr("")` degrades to no guard. The contract's `nowStale` leg is falsifiable on exactly that claim.

## Ported — 3 cases × 3 legs

- **`RunLifecycleCloseAdmitsATransitivelyBlockedTarget`** — the find of the slice. The Lifecycle contract was green on three legs while **every** case seeded a *direct* blocker, so a guard reading the denormalized `is_blocked` column instead of the live blocker list passed all of them. That is `903cbe031` — the Deleter defect — verbatim in another contract. Now seeded with a parent-child child of a blocked parent, both raw preconditions read as `t.Fatal`s rather than inferred, plus a closing parent-refusal control.
- **`RunIssueOperationsCreateRefusesAForeignIDPrefix`** — every other create case sets `ForceIDPrefix`, so the flag was only ever asserted from the side that *bypasses* the check. Nothing said the check existed.
- **`RunIssueOperationsUpdateMetadataPatchOrdersMergeSetUnset`** — order is only observable when the three edits collide, and no existing case collides.

## What adversarial review caught, and this PR fixes

Three review agents attacked the branch. All three found something, and none of it was visible to `go build`, `go vet`, `go test` or `golangci-lint`:

1. **The deletion orphaned the CI shard manifest** (`d4f9b3f2f`). Five deleted test names remained in `.github/scripts/embedded-storage-test-shards.txt`, and `embedded-storage-test-shard.sh` hard-fails on any entry that no longer matches a discovered test. All five embedded-storage shard jobs would have exited 1 at startup, across `pr-risk.yml` and `main.yml`. The breakage lived in a text file no compiler reads, in a directory the slice never touched.

2. **The deletion lost a real route** (`f39ee40b4`). `EmbeddedDoltStore.{Update,Close}IssueChecked` are a *separate composition* of the shared bodies — the wrapper decides inside its own `withConn` whether to run the version check and which options to forward, while the role path reaches the same bodies via `runTransaction` and never calls these methods. Every probe the slice ran mutated the shared functions; none mutated the wrapper, which is the plane the deleted files existed for. Proof: disabling the CAS branch in the wrapper failed the deleted test at base and passed all 136 contract cases. A narrow wrapper-routing residue is restored, verified red against that same mutant.

3. **The metadata port lied about its own name** (`6dcd7bb75`). It pinned Set≺Unset and Merge≺Unset but not **Merge≺Set** — the third its name leads with. Its only Merge/Set collision was `"keep"`, which Unset also removes, so a Set-before-Merge body produced a byte-identical document. Fixed with one key written by both and removed by neither. The transferable lesson: *a key that any later stage removes cannot witness the order of the stages before it.*

Also rehomed `embeddedIssueParentIDs`, which had all its in-file callers moved to the contract and was left defined in one file and used from another — a shape that already broke `issue_operations_staging_test.go` once mid-slice.

## Kept, deliberately

`comments_page_test.go` (`GetIssueCommentsPage` is a live seam with no role — a uow provider is not a `DoltStorage`, so it can never be a three-leg case), `counts_wisp_test.go` (`issueops.Counter` asks a different question entirely), `reopen_category_test.go` (seeds the wisp plane and the normalized `custom_statuses` table, neither of which the contract does), `create_issue_test.go` (defines `newTestEnv`/`skipUnlessEmbeddedDolt` for 58 siblings including all 26 conformance runners), and `issue_operations_persistence_test.go` (the only file with no `skipUnlessEmbeddedDolt` gate).

## Verification

`go build ./...`, `go vet`, `golangci-lint run` clean. `IssueOperations` and `LifecycleCloseReopen` contracts green on dolt, uow and embeddeddolt. Gating preserved: all four deleted files were `//go:build cgo` + `skipUnlessEmbeddedDolt`, and the covering runners carry the identical tag and gate.


julianknutsen:
## Red-team review council

Three adversarial agents over this branch, each briefed to break it rather than confirm it: **coverage-loss**, **port-integrity**, **structural-residue**.

### Verdict: all three found something. None of it was visible to any gate.

`go build`, `go vet`, `go test` and `golangci-lint` were green over every one of the three findings below. That is the point worth carrying: on a deletion slice, the gates cannot tell you what you removed.

---

**1. The deletion orphaned the CI shard manifest — every embedded-storage shard job would have died at startup.** *(fixed in `d4f9b3f2f`)*

Five deleted test names remained in `.github/scripts/embedded-storage-test-shards.txt`, and `embedded-storage-test-shard.sh:96-101` hard-fails on any manifest entry that no longer matches a discovered test. All five shards exit 1 — driven from `pr-risk.yml:237` and `main.yml:1002` over a 5-shard matrix. The slice's commits touch zero `.github` files; the breakage lived in a text file nothing compiles.

Fix is a 5-line deletion: unlisted tests fall back to hash distribution, which the manifest header states is its purpose. Confirmed on this PR — all 20 embedded-storage shards plus the conformance oracle are green.

**2. The deletion removed the only observers of a public-API wrapper route.** *(fixed in `f39ee40b4`)*

The slice justified itself with *"the role bodies call the same two functions, so the contract sees every break these did."* True for the shared `InTx` functions — **false for the wrapper's own composition.** `EmbeddedDoltStore.UpdateIssueChecked` (`issues.go:89-114`) composes `CheckVersionInTx` + `CheckExpectedFieldsInTx` + `UpdateIssueInTx` *itself inside its own `withConn`*; the role path reaches the same shared bodies through `runTransaction` and never calls the method. Every probe the slice ran mutated the shared functions — the one plane where redundancy holds — and none mutated the wrapper, which is the plane the deleted files' headers said they existed for.

Proof, not argument: disable the CAS branch in the wrapper and at base the deleted test **fails** while all 136 contract cases **pass**; after the deletion, nothing catches it. FAIL/PASS → PASS/PASS. After deletion this package had zero callers of either method, and both sit on the `storage.DoltStorage` interface with `beads.go` publishing their option types as caller-facing API.

This is the `903cbe031` Deleter shape again: a suite green across three legs while a supported route is disconnected. The restored residue is deliberately narrow — it re-tests no refusal taxonomy, no event counting, no rollback — and asks only whether the wrapper still routes each precondition to the body implementing it, which is what the contract layer structurally cannot ask. Verified red against that same mutant.

**3. A ported case did not pin the promise its own name leads with.** *(fixed in `6dcd7bb75`)*

`RunIssueOperationsUpdateMetadataPatchOrdersMergeSetUnset` pinned Set≺Unset and Merge≺Unset but **not Merge≺Set**. Its only key colliding between Merge and Set was `"keep"` — and `"keep"` is also in Unset, so a body applying Set before Merge produced a byte-identical document. A compilable Set→Merge reordering of `ApplyMetadataPatch` left the case green *and* all 74 `TestIssueOperationsUpdate*` cases green: the Merge≺Set promise at `issueops/issueops.go:83-86` was pinned nowhere in the contract.

Fixed with one key written by both stages and removed by neither. The transferable lesson: **a key that any later stage removes cannot witness the order of the stages before it.**

---

### What the council cleared

- **The other deletions are sound.** Lens A independently re-derived the slice's own claim that `update_guards_test.go` was *partly cannot-fail* — its only `ptr("")` leg runs against a fresh unassigned issue and asserts success, so it cannot fail if `ptr("")` degrades to no guard. The behavior-file trim and the validation deletion both hold: every dropped assertion has a named contract counterpart at equal or greater strength.
- **The other two ports are sound.** `AdmitsATransitivelyBlockedTarget` reads raw preconditions as `t.Fatal`s rather than inferring them and carries a closing parent-refusal control; `CreateRefusesAForeignIDPrefix` goes red under a shared-guard mutant. Bonus finding: the uow create route stacks a *second* independent prefix guard (`validateExplicitIDPrefix`), so that leg only reddens when both break — real defense-in-depth, correctly invisible to a black-box case.
- **Structural: one stranded helper, no dead code.** `embeddedIssueParentIDs` had all in-file callers moved to the contract and was left defined in one file, used from another — the shape that already broke `issue_operations_staging_test.go` once mid-slice. Rehomed beside its caller. Full reference-count diff otherwise clean; no orphaned citations; gating preserved (all four deleted files were `//go:build cgo` + `skipUnlessEmbeddedDolt`, and the covering runners carry the identical tag and gate); residue file judged justified rather than a stub.
- **Two suspicions reported as unproven** rather than dressed up: stale-`ExpectedVersion` UPDATE rollback and mixed valid+invalid patches. Neither yielded a line-level mutant the contract misses.

PR #5405, opened 2026-08-07
Retire the dolt checked-close/update seam tests, and cover what they were actually observing

julianknutsen:
Slice 2 of a test-deduplication pass against the conformance contracts, and the one where deletion was genuinely available: `CloseIssueChecked`/`UpdateIssueChecked` have **zero** non-test `cmd/bd` callers, so those seams have converged on the roles. Net **−648** across 9 files (914 added, 1,562 removed): 11 whole test funcs, 9 table rows and 12 subtests deleted, 3 cases ported, 1 residue kept and then substantially widened after review.

## Deleted, with evidence

14 assertion-targeted mutations, each showing a contract case goes red on the same break: blocker refusal, open-children refusal, version CAS on close and update, assignee guard, status guard, LabelPatch ordering, ParentID replace-all, configured-done resolution, priority-range validation.

## Ported — 3 cases × 3 legs

Every close case in `lifecycle_close_reopen_contract.go` seeded one fixture shape — open target, direct live blocker, durable children — and that hid three separate terms:

- **`CountsOpenChildrenInBothPlanes`** — one durable + one ephemeral child (count 2), and one edge resident in both dependency tables (count 1). Catches deleting the wisp-plane child query, and separately deleting its `NOT EXISTS` dedup. Review verified both arms fail independently.
- **`AdmitsAStaleBlockFlagWhoseBlockersHaveClosed`** — a live edge onto an already-closed blocker. The earlier transitive case structurally cannot reach it: its target has no direct edge, so it passes against a blocker list that never checks status. Measured, not asserted — under the mutant the transitive sibling stays green while this one reddens.
- **`IsIdempotentOnAClosedRowThatStillLooksBlocked`** — a forced close leaves the row closed with its blocker open and `is_blocked` set. A guard running there makes every forced close of a blocked issue permanently un-re-closable.

Also found a **pre-existing cannot-fail test**: `TestCloseIssueChecked`'s stale-`is_blocked` subtest seeded the flag with *no blocker edge*, so the guard short-circuits and it could never fail on its own claim. The port seeds the discriminating state.

## What review changed, and it was most of the value

The slice kept a wrapper residue — correctly, because `DoltStore.{Update,Close}IssueChecked` compose the shared bodies *themselves* while the role path reaches them via `runTransaction` and never calls these methods. But it wrote that residue **from the two mutants it had already run**, pinning the version guard on three branches and nothing else. Adversarial review then broke eight more single statements, and every remaining test in the package plus every contract case on every leg stayed green on all of them:

| break | why it matters |
| --- | --- |
| `ExpectedStatus` not forwarded — durable *and* wisp | the claim-on-behalf guard, routed on **no** branch |
| both field guards skipped on the wisp branch | guarded wisp writes silently apply |
| `updateWispChecked` commits nothing | guards run, tx commits, no write |
| demotion branch never demotes | passes its guard, doesn't migrate |
| `Force` not forwarded on wisp close | a blocked wisp becomes uncloseable |
| `Unchanged` mapped to a constant | caller can't tell re-close from close |
| `Reason`/`Session` dropped, both branches | closes recorded anonymously, nothing read them back |

**Half of those are positive assertions for branches whose refusal was already covered.** A wrapper that refuses correctly and never writes passed the entire suite — refusal-only coverage of a guarded write is half a test.

Residue is now 15 subtests, each verified red against the mutation it exists for.

## Also fixed: citation rot

Two new case comments cited `issueops.go:406-409` and `:364-365` for Close's open-children promise and `CloseResult`'s prose. Those ranges are the `Lifecycle` interface header and a closing brace. The slice **copied** a stale citation already in the file — and it turned out **10 of the file's 14** `issueops.go` citations were wrong, drifted by growth above the Lifecycle region. All re-resolved against HEAD and each verified to contain the prose its comment quotes. Comment-only.

## Kept, deliberately

`close_policy_gate_test.go` (the only test in the tree seeding a closed row with a live blocker), `update_same_status_test.go`, `update_force_close_policy_test.go` (`UpdateIssue` has ~30 production callers), `reopen_category_test.go` (normalized `custom_statuses` table + Dolt HEAD), and the persistence, claim-verify, concurrency and pure-function files.

## Verification

`go build ./...`, `go vet`, `golangci-lint run` clean. `IssueOperations` + `Lifecycle` contracts green on dolt, uow, embeddeddolt. All 19 residue subtests green. No `.github` manifest references any deleted test — verified rather than assumed (the three shard manifests contain no `internal/storage/dolt` names; the dolt CI job runs `-test.run '^TestConformance$'`).


julianknutsen:
## Red-team review council

Three adversarial agents over this branch — **coverage-loss**, **port-integrity**, **structural-residue** — each briefed to break it rather than confirm it. All three found something. The corrections are in this PR.

### 1. The residue was present but not sufficient — eight uncaught wrapper breaks

This slice correctly learned the previous slice's lesson: `DoltStore.{Update,Close}IssueChecked` compose the shared `issueops` bodies *themselves*, while the role path reaches those bodies through `runTransaction` and never calls these methods — so a break confined to the wrapper is invisible to every contract case on every leg. It kept a residue for exactly that reason.

**But it wrote the residue from the two mutants it had already run**, which pinned the version guard on three branches and nothing else. The review then broke eight more single statements, and every remaining test in the package plus every conformance case on every leg stayed green on all of them:

| break | consequence |
| --- | --- |
| `ExpectedStatus` not forwarded — durable, **and** wisp | the claim-on-behalf guard, routed on **no** branch |
| both field guards skipped on the wisp branch | guarded wisp writes silently apply |
| `updateWispChecked` commits nothing | guards run, tx commits, no write |
| demotion branch never demotes | passes its guard, doesn't migrate |
| `Force` dropped on wisp close | a blocked wisp becomes permanently uncloseable |
| `Unchanged` mapped to a constant | caller cannot tell re-close from close |
| `Reason`/`Session` dropped, both branches | closes recorded anonymously; nothing read them back through this seam |

**Half of those are POSITIVE assertions for branches whose refusal was already covered.** A wrapper that refuses correctly and never writes passed the entire suite. Refusal-only coverage of a guarded write is half a test, and that generalises well past these two methods.

Residue is now 19 subtests, each verified red against the mutation it exists for.

### 2. Ten of fourteen citations in the Lifecycle contract were wrong

Two of this slice's new comments cited `issueops.go:406-409` and `:364-365` for Close's open-children promise and `CloseResult`'s prose. Those ranges are the `Lifecycle` interface header and a closing brace. The slice did not invent the error — it **copied** a stale citation already in the file and minted one more beside it.

The drift is older than the slice: the citations were authored when `issueops.go` was ~12 lines shorter above the Lifecycle region. Citations claiming to quote `Close` landed on the `Create` doc; one claiming a validation promise landed on `ReopenResult.Issue`. All fourteen re-resolved and each verified to contain the prose its comment quotes. Fixing only the two this slice touched would have left twelve wrong ones and made the file less trustworthy, not more.

### 3. Ports: sound, and independently falsifiable

All three go red under a mutation of their own subject, with a 12-case sibling control staying green:

- `CountsOpenChildrenInBothPlanes` — **both** of its claims verified separately. Under a dropped wisp-plane query the two-plane arm reddens and the dual-resident arm passes; under a dropped `NOT EXISTS` dedup the reverse. Neither arm carries the other.
- `AdmitsAStaleBlockFlagWhoseBlockersHaveClosed` — the "the transitive case structurally cannot cover this" claim is now **measured**: under the mutant the transitive sibling stays green while this one reddens.
- `IsIdempotentOnAClosedRowThatStillLooksBlocked` — verified by making the guard actually run.

The raw-SQL `Exec` seeding is precondition-guarded in all three, so a fixture that fails to install the unusual state dies at a `Fatalf` rather than passing quietly — the exact decay mode of the cannot-fail subtest this slice retired.

One scope note the review pinned down: the three cases run three *routes* over **one** shared policy body (`issueops.CloseIssueCheckedInTx`, reached by dolt/embedded via `ExecuteClose` and by uow via `domain/db/issue.go`). Real three-leg wiring, but read the ledger's "3 × 3" as route coverage, not three semantic votes.

### 4. Structural: clean, verified rather than assumed

The CI-killer class that hit the previous slice was checked directly, not waved off: a full-tree grep of all 11 deleted test names and the deleted filename found no `.github` workflow, script, or manifest reference. The three shard manifests contain zero `internal/storage/dolt` names, and the dolt CI job runs `-test.run '^TestConformance$'` — deleted tests never ran there and neither do their replacements. Symmetric.

Also clean: a zero-reference sweep over all 281 package-level helpers (none dead; `stringPtr` correctly removed with its last caller); test binaries compile on all three storage packages; `CreateWisp`/`Exec` implemented at all three wiring sites with loud failures on nil; no gating loss (no build tags at base, survivors identical).

### Not fixed, deliberately

A forced-close leg on the wisp branch that the residue does not re-pin — the seam has no production caller and all options forward in a single call. And the ledger's stale per-file counts. Both cosmetic; neither ships.

PR #5406, opened 2026-08-07
Retire uow lifecycle assertions the IssueOperations contract answers for

julianknutsen:
Slice 5 of a test-deduplication pass against the conformance contracts. Net **−187** across 7 files (92 added, 279 removed). 28 test funcs triaged → **4 deleted** (plus 3 dead helpers and one 41-line block inside a kept func), **1 port**, **24 kept**, **8 handed off**. No files removed — trims only.

16 mutants run, all on the uow leg. None read INCONCLUSIVE, so no cannot-fail test surfaced among the deletions.

## The port — a cannot-fail case already in the contract

`RunReaderListReadyFlagAnswersTheBlockerAwareSet` used ready ids `-1` and `-10`, whose **natural, lexical and query orders all read identically**. So it saw only the missing trim. A body that kept the trim and dropped the display order on the `--ready` arm — the other half of the exact regression its own comment describes — passed all 47 Reader cases at uow.

Adding `-2` separates natural (1,2,10) from lexical (1,10,2) from seed order. Green on all three legs, red on all three against a broken body (the uow body, and the `storereader` body at dolt and embedded). No new `Run*`, so no runner edits, and no leaf-doc change — `issueops/reader.go` already states the promise.

## Deletions — each FAIL/FAIL

- cross-tier ID collisions ⊂ `CreateRefusesAnOccupiedID` (3 directions, raw rows)
- invalid canonical update values ⊂ its own **mock-provider sibling**, which proves more without a database (`newUOWCalls == 0`)
- typed metadata patch ⊂ `UpdateMetadataPatchOrdersMergeSetUnset` — and the contract case can witness Merge≺Set where the deleted one couldn't, since every key it collided was also unset later
- `TestReaderPagesAreNeverNil` ⊂ `ListEmptyPageIsWellFormed`
- the omnibus's three CAS refusals ⊂ `ConditionalGuardsGateOrdinaryEdits` + `UpdateClosePolicy`

## Kept, and the reason matters beyond this slice

**All 7 `issue_claimer_test.go` funcs — `issueops.Claimer` has no conformance contract.** There are 23 contract files and none of them is Claimer's: `ready_claimer_contract.go` covers `ReadyClaimer`, a *different* interface, and `claim.go` is the two-leg audit suite one layer below the role. The dolt and embeddeddolt implementations have no tests at all. Conjunct (1) of the obviation criterion fails outright — these are not duplicates of anything, they are the only role-level coverage that exists.

Also kept: `issue_operations_update_real_uow_test.go` cannot be retired whole (it defines the IssueOperations contract runner's own fixture constructor); five UNIQUE verdicts (Owner patch — `Owner` appears nowhere in the contract; issue-type vocabulary on update; the 17-field no-op breadth; claim-counts-as-mutation; close provenance across a persistence move); and the force-close table kept whole because row 3 has no caller-visible effect and is meaningless without the contrast.

## Handoffs, not edits

8 candidates port into `issue_operations_contract.go` and `lifecycle_close_reopen_contract.go`, which a parallel slice owns exclusively. Rather than edit a contested file, each is recorded in `S5-HANDOFF.md` with the gap, the seeding it needs and its mutation evidence, and the ad-hoc test left standing. Nothing was dropped to avoid a conflict.

## Not established, and recorded as such

Two omnibus steps (open-children close refusal, reopen no-op) look covered but the mutant budget was spent — recorded, not claimed. No contract case puts a number or nested object through `MetadataPatch.Set`.

## Verification

`go build ./...`, `go vet`, `golangci-lint run` clean. `.github` has no per-test manifest for uow (whole package, `-race`, required check), so no manifest can be orphaned.

### DIFF
--- a/backend/conformance/issue_operations_contract.go
+++ b/backend/conformance/issue_operations_contract.go
@@ -352,6 +352,13 @@ func RunIssueOperationsUpdateMetadataPatchOrdersMergeSetUnset(t *testing.T, ctx 
 				"added":     json.RawMessage(`"set"`),
 				"keep":      json.RawMessage(`"set"`),
 				"contested": json.RawMessage(`"set"`),
+				// NOT STRINGS, deliberately. Every other Set value in this file
+				// is a JSON string, so a body that accepted only strings — uow
+				// validates Set values by shape in its own gate, before the
+				// shared apply — passed every case here. A number and a nested
+				// object are the two shapes a caller actually stores.
+				"count":  json.RawMessage(`7`),
+				"nested": json.RawMessage(`{"a":[1,2],"b":{"c":true}}`),
 			},
 			Unset: []string{"keep", "drop"},
 		},
@@ -372,9 +379,9 @@ func RunIssueOperationsUpdateMetadataPatchOrdersMergeSetUnset(t *testing.T, ctx 
 	// broken order. "contested" survives the patch, so it records which of the
 	// two wrote last — Set does, per issueops.go's Merge≺Set≺Unset promise.
 	assertIssueOperationsMetadata(t, "ordered metadata patch", ordered.Issue.Metadata,
-		`{"added":"set","contested":"set","merged":true}`)
+		`{"added":"set","contested":"set","count":7,"merged":true,"nested":{"a":[1,2],"b":{"c":true}}}`)
 	assertIssueOperationsStoredMetadata(t, ctx, fixture, id, "after the ordered metadata patch",
-		`{"added":"set","contested":"set","merged":true}`)
+		`{"added":"set","contested":"set","count":7,"merged":true,"nested":{"a":[1,2],"b":{"c":true}}}`)
 
 	// Replace beside any incremental edit is refused, and the document the
 	// replacement would have written never lands.
@@ -400,7 +407,7 @@ func RunIssueOperationsUpdateMetadataPatchOrdersMergeSetUnset(t *testing.T, ctx 
 				t.Fatalf("%s: err = %v, want ErrValidation", name, err)
 			}
 			assertIssueOperationsStoredMetadata(t, ctx, fixture, id, "after "+name,
-				`{"added":"set","contested":"set","merged":true}`)
+				`{"added":"set","contested":"set","count":7,"merged":true,"nested":{"a":[1,2],"b":{"c":true}}}`)
 		})
 	}
 	events.assert(t, "refused replace-plus-incremental patches", 0, nil)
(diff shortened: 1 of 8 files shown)


julianknutsen:
## Red-team review council

Three adversarial agents over this branch — **coverage-loss**, **port-integrity**, **structural-residue** — each briefed to break it rather than confirm it.

### Verdict: three confirmed coverage losses. The deletions were unsound; all three are now ported.

The slice ran 16 mutants and none read INCONCLUSIVE, so it reported no losses. The review found three anyway, all one shape:

> **A FAIL/FAIL verdict is only true of the body the mutation was in.** The slice's probes broke a promise *wholesale*; the deleted assertions observed a *composition*, and no contract case assembles it.

**1. Claim under a stale `ExpectedVersion` — stale-version claims could win silently.**
`internal/storage/issueops/aggregate.go` refuses `Claim` beside `ExpectedAssignee` or `ExpectedStatus` before the unit of work opens, so **`ExpectedVersion` is the only legal claim/guard composition** — the optimistic fence a caller uses to claim a row it has already read. Its refusal side lived only in the deleted 41-line block; the omnibus kept the *positive* side, which passes even with the guard bypassed. Base: deleted block FAIL / 36-case battery PASS. After deletion: PASS/PASS, including `TestDomainDB`, the uow claim supplement and the ReadyClaimer contract.

**2. Guard masking — an earlier guard that holds beside a later one that is stale.**
Every existing refusal puts the stale guard *first*, so a check-only-the-first-present chain — an `else if` where an `if` belongs — refused all of them correctly and let this through. `ConditionalGuardsGateOrdinaryEdits` had both-hold and stale-assignee+holding-status, but never holding-earlier+stale-later.

**3. Numbers and nested objects through `MetadataPatch.Set`.**
Every positive `Set` value in the entire contract file is a JSON string, and the kept tests are refusal-only. uow validates Set values **by shape in its own gate** before the shared apply. The ledger judged this residual "not a distinct code path" by reasoning about `applyTypedMetadataEdits` storing `RawMessage` opaquely — the wrong body, which is exactly how the loss got through.

### The remediation, and why porting rather than restoring

All three extend **existing** cases, so no runner wiring changed and each gains three-backend coverage the deleted one-leg assertions never had. Each verified through the isolated harness against the review's own mutant:

| port | mutant | uow | dolt |
| --- | --- | --- | --- |
| claim fence | version guard stops applying to a `Claim` | **FAIL** | PASS |
| guard masking | first-present-only guard chain | **FAIL** (`err = <nil>`) | PASS |
| typed metadata | `Set` values must be strings | **FAIL** | PASS |

dolt staying green is the correct signal, not a gap: every mutated body is uow-only, which is precisely why the contract's other legs could not see these losses.

### What the council cleared

**Port integrity — sound, and it closed a gap in the slice's own verification.** The slice only re-ran the *sort* mutant post-port; the trim half was never re-verified. The review ran it — both claims are independently falsifiable, each reddening only its own assertion, on both body families (uow's own reader, and `storereader` at dolt/embedded). It also produced the proof the fix is not cosmetic: at base, with the old two-id case, the sort-drop mutant **passes** at dolt.

**The Claimer keeps — justified on all three factual legs.** 23 `*_contract.go` files and none is Claimer's; `ReadyClaimer.ClaimNext` (pick-and-claim) is genuinely a different role from `Claimer.Claim` (named-issue CAS), separate interfaces and separate implementation files; no dolt or embeddeddolt test references the role. These are not duplicates of anything — they are the only role-level coverage that exists.

**The other keeps and all 8 handoffs — sound.** Each of the five UNIQUE verdicts re-verified statically (`Owner` occurs 0 times in the contract; no update-plane `IssueType` case; no broad same-value no-op case; both contract claim cases claim unclaimed rows; no case moves a closed row across planes). No handoff is a deletion in disguise. Two harmless wording overclaims flagged.

**Structural — clean.** The CI-killer class was verified directly: no manifest, workflow, `-run` regex or engdoc names anything deleted, and no per-test manifest covers uow at all. Full func/type/var reference-count diff shows nothing went decl-only that wasn't already. Test binaries compile on all three packages. The contract edit's blast radius is contained — `lsready` ids appear in exactly one case.

### Also fixed: 13 stale citations

`reader_contract.go` carried ~40 line citations and 13 pointed at the wrong prose after `issueops/reader.go` was restructured beneath them. "Items is never nil" pointed three times at the MaxRows paragraph; the never-mutate promise pointed at prose about plane order; `Get`'s error contract pointed at `--json` rendering. All re-resolved and every citation in the file verified against HEAD. The slice authored none of them — it added no line citations at all, citing by symbol name instead — but the earlier Lifecycle sweep in this programme showed why leaving them is not neutral: a slice there **copied** a stale citation out of a file it was editing.

PR #5407, opened 2026-08-07
fix(list): stop claiming Total when the page is truncated (GH#5362)

### LINKED ISSUE #5362: bd list reports a false Total when truncating at its default limit; --json silently ignored without --flat; --limit 0 undocumented on bd ready
Three related output-trust issues found while running bd (v0.59.0, Homebrew) as the coordination bus for a multi-agent workflow. Together they made truncation undetectable to a reader and cost us a production incident (a work queue that silently read as empty).

**1. `bd list` asserts a false total when truncating.** With no `--limit`, list caps at 50 rows and the footer prints the CAP as the total:

```
$ bd list --status closed --closed-after 2026-07-29
...
Total: 50 issues (0 open, 0 in progress)
```

The same query with `--limit 0` returns 233 rows. There is no "Showing 50 of 233" notice (contrast `bd ready`, which does print "Showing 10 of 108 ready issues. Use -n to show more." - that honest form would fix this). A reader cannot detect the truncation from the output at all.

**2. `--json` on `bd list` is silently ignored unless `--flat` is also passed.** The default `--tree` renderer wins over `--json` with no warning and exit 0:

```
$ bd list --json --limit 3     # -> tree TEXT on stdout, exit 0
$ bd list --flat --json --limit 3   # -> valid JSON
```

Any script piping `bd list --json` into jq gets text and a zero exit code. Either `--json` should imply `--flat` (JSON has no tree rendering anyway) or the combination should error.

**3. `--limit 0` (unlimited) is undocumented on `bd ready`.** `bd list --help` documents `(default 50, use 0 for unlimited)`; `bd ready --help` says only `Maximum issues to show (default 10)`. Empirically `bd ready --limit 0` returns the full set and we now depend on it - documenting it would make that dependency safe.

Happy to provide more repro detail. Thanks for bd - it is load-bearing infrastructure for us.

vishnujayvel:
## What

Stop `bd list`'s pretty-tree summary from labeling a page-capped row count as
`Total: N`. When the page was truncated by `--limit`, the footer now says
`Showing N issues …; more match (truncated by --limit)` so a partial queue
cannot be mistaken for the full set.

## Why

Fixes #5362.

Reported by @AgentMonocle: with the default limit of 50,
`bd list` printed `Total: 50 issues` even when hundreds matched. Multi-agent
workflows treated that as a complete queue and missed remaining work
(production incident cited in the issue).

`bd ready` already distinguishes page size from the full set; this brings the
list pretty footer in line with that honesty contract. The existing stderr
truncation hint (GH#3212) remains; this fix is about the stdout summary that
asserted completeness.

### Out of scope (issue secondary items)

1. **`--json` silently ignored without `--flat`** — already fixed on current
   `main`: the JSON branch runs before the tree renderer, and `--json`
   clears pretty/tree selection. Verified by code inspection of `runListCore`
   / `gatherListInput`. Not changed here.
2. **`--limit 0` undocumented on `bd ready`** — already fixed on current
   `main`: help text is `Maximum issues to show (use 0 for unlimited)`. Not
   changed here.

## Verification

```bash
# Unit: truncated vs untruncated pretty summary
go test ./cmd/bd/ -count=1 -run 'TestListDisplayPrettyList'

# Manual: create >50 matching closed issues, then:
bd list --status closed --closed-after <recent>          # footer: Showing 50…truncated
bd list --status closed --closed-after <recent> --limit 0  # footer: Total: N
```

Local: `go test ./cmd/bd/ -count=1 -run 'TestListDisplayPrettyList|TestDisplayWatchedIssueList'` → pass.

## Notes for reviewers

- No new flags / surface.
- On an interactive TTY with a truncated page you'll now see two similar
  sentences: this new stdout footer and the existing stderr truncation hint
  (GH#3212). They're on different streams by design — the stderr hint is
  TTY-only and invisible to piping agents, which is exactly the audience #5362
  is about, so the stdout footer is the one that fixes the bug. Happy to fold
  them together if you'd rather.
- True cardinality (`Showing N of M`) would need a list-predicate Counter
  that mirrors Reader.List defaults (Counter's default exclusions differ from
  list's); left for a follow-up if desired. HasMore is enough to refuse a
  false Total without an extra count query.

### DIFF
--- a/cmd/bd/list.go
+++ b/cmd/bd/list.go
@@ -307,7 +307,8 @@ func runListCore(cmd *cobra.Command, _ []string) error {
 			if depErr != nil && in.depsMode != "" {
 				return HandleError("loading dependencies for --deps: %v", depErr)
 			}
-			displayPrettyListWithDepsMode(treeIssues, false, allDeps, in.depsMode)
+			// Hierarchical --parent walks use an unlimited per-level query, so the tree is never page-truncated.
+			displayPrettyListWithDepsMode(treeIssues, false, allDeps, in.depsMode, false)
 			printSkipLabelsFooter(in.SkipLabels)
 			return nil
 		}
@@ -316,7 +317,7 @@ func runListCore(cmd *cobra.Command, _ []string) error {
 		if depErr != nil && in.depsMode != "" {
 			return HandleError("loading dependencies for --deps: %v", depErr)
 		}
-		displayPrettyListWithDepsMode(issues, false, allDeps, in.depsMode)
+		displayPrettyListWithDepsMode(issues, false, allDeps, in.depsMode, truncated)
 		printTruncationHint(truncated, in.effectiveLimit)
 		printSkipLabelsFooter(in.SkipLabels)
 		return nil
--- a/cmd/bd/list_helpers_test.go
+++ b/cmd/bd/list_helpers_test.go
@@ -304,6 +304,31 @@ func TestListDisplayPrettyList(t *testing.T) {
 	if !strings.Contains(out, "bd-1") || !strings.Contains(out, "bd-1.1") || !strings.Contains(out, "Total:") {
 		t.Fatalf("unexpected output: %q", out)
 	}
+	if strings.Contains(out, "Showing ") {
+		t.Fatalf("untruncated list must keep Total: wording, got: %q", out)
+	}
+}
+
+// GH#5362: a page cut by --limit must not label the page size as Total.
+func TestListDisplayPrettyList_TruncatedSummary(t *testing.T) {
+	issues := []*types.Issue{
+		{ID: "bd-1", Title: "A", Status: types.StatusOpen, Priority: 2, IssueType: types.TypeTask},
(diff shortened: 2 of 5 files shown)


steveyegge:
## Review verdict: MERGE

Adversarially reviewed against GH#5362 — and every hunt came back refuted with evidence, which is the best outcome a reviewer can report:
- No off-by-one at exactly-limit: every `truncated`/`hasMore` derives from the limit+1 over-fetch (`workapi.FinishPage` only sets it when the cut removed a row), so an exactly-full page with nothing behind it still says `Total:` — correct.
- Both `--watch` loops recompute the flag per tick.
- The two hardcoded `false` call sites are provably right: both hierarchical `--parent` walks zero the limit per level.
- Output-compat: nothing in-repo (or in our operator scripts) parses the old footer, and untruncated output is byte-identical.

One follow-up worth its own issue, since it's the half of #5362 that actually bit the reporter's agents: `bd list --json` still emits a bare array with no truncation signal, while `bd ready --json` already carries `PaginationMeta{Returned, Total, Truncated}`. Filing that now and linking it here. Full CI green. Merging.

PR #5408, opened 2026-08-07
Four ordering contract cases for sort behavior no shared proof owns

julianknutsen:
Four ordering cases for `backend/conformance`, all in the existing
`search-counts-stats` audit slice whose charter already names `sqlbuild/sort.go`
as its grounding. Test-only: no production code changes, one file, +142/-0.

## Why

`sqlbuild/sort.go` makes four promises that the shared suite never asked a
backend to keep. A single backend never notices — its own tests cover it — but
the conformance suite is the proof obligation a *second* storage backend
carries, and today it would have to reproduce these behaviors with no shared
proof telling it what "correct" is.

Each case closes one such gap, and each is grounded in a specific rendering
decision in `sqlbuild/sort.go`:

| Case | Contract | Why it was unowned |
| --- | --- | --- |
| `SearchSortByTitleCaseFolded` | A bounded page sorted by title keeps the first N rows under `LOWER(title)` order, ties by id ASC | The slice had no title sort; the querier-role case titles every row with its lowercase id, so a byte-collated backend passes it |
| `SearchSortTieBreakSurvivesReversal` | Non-default sorts break ties by id ASC, and `SortDesc` flips the primary key only | The audit pinned the tie-break for the default and closed sorts only; the querier's reversed legs run above the role epilogue, on title/priority, never on a tie |
| `SearchSortNullsDirectional` | Nullable keys (closed_at, assignee) put valueless rows first on ASC, last on DESC | Only the closed sort in its default direction was pinned; the flipped direction and the assignee key had none |
| `SearchSortByIDCompleteSet` | `SortBy: "id"` is accepted and answers the complete matching set | No conformance search case ever sent `SortBy: "id"` to `SearchIssues`; the reader-role id-sort cases run above the epilogue that hides it |

The regression sentence for each case is in its commit message and its doc
comment.

## Red/green

These pin reference behavior, so they pass against the oracle as written.
To show each assertion is load-bearing rather than decorative, each case was
run against a deliberately diverged `sqlbuild/sort.go` modelling the exact
failure class it exists to catch, then reverted:

| Injected divergence | Result |
| --- | --- |
| `ORDER BY title` instead of `LOWER(title)` (byte collation) | Only `SearchSortByTitleCaseFolded` fails — both the full order and the `Limit 2` page |
| `id <dir>` tie-break, so reversal flips it too | Only `SearchSortTieBreakSurvivesReversal` fails |
| `(col IS NULL)` lead term inverted (models NULLs-last-on-ASC engines) | `SearchSortNullsDirectional` fails on all three legs |
| Go-side key pushed into SQL as an unresolvable column (models a store that refuses what it cannot render) | Only `SearchSortByIDCompleteSet` fails |

In every run the pre-existing ordering cases stayed green, which is the
residue claim these four cases rest on.

One result is worth recording. Removing the `(col IS NULL)` term outright is a
**no-op** on Dolt — its native NULL ordering already matches the contract — so
the divergence only shows against an engine whose native placement is the
opposite. That is precisely why the term is written explicitly "so the contract
does not depend on a driver's default NULL ordering", and why the proof belongs
in the shared suite rather than at one backend that cannot observe it.

The same run answers a fixture question: because inverting the term (rather
than removing it) is what moves the assignee legs, unassigned is stored as SQL
NULL here. The assertions are written not to depend on that — the asserted
order is identical under a NULL or an empty-string encoding, since the rows are
seeded through the public create path exactly as `bd create` leaves them.

## Verification

Run against both behaviorally distinct in-tree implementations wired to
`RunAudit_search_counts_stats`:

```
BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go -timeout 30m \
  ./internal/storage/embeddeddolt/ -run '^TestConformance$/^Audit$/^search-counts-stats$'
ok  github.com/steveyegge/beads/internal/storage/embeddeddolt  9.96s

CGO_ENABLED=1 go test -tags=integration,gms_pure_go -timeout 20m -count=1 \
  ./internal/storage/dolt/ -run '^TestConformance$/^Audit$/^search-counts-stats$'
ok  github.com/steveyegge/beads/internal/storage/dolt  17.12s
```

Plus `./scripts/ci/pr-lint.sh` (gofmt + golangci-lint, both passes: 0 issues)
and the full `./scripts/conformance.sh` lane:

```
Tier 1  embedded-Dolt RunAll   64.76s
Tier 2  e2e bd init + CLI      35.84s
==> conformance OK             1m50s wall
```

Timing: the four new cases add **1.12s** on embedded Dolt (0.32 + 0.29 + 0.31 +
0.20), taking the `search-counts-stats` slice from 8.74s to 9.86s. Each case is
an independent leaf with its own fresh store, so the delta is additive against
the lane total above.

## Notes

- No production code is touched, so nothing here changes backend behavior — it
  only states behavior the backends already have.
- Cases land in the existing `RunAudit_search_counts_stats` entry point, so the
  in-tree Dolt and embedded-Dolt suites pick them up with zero new wiring.
- Sequence is deliberately **not** asserted for `SortBy: "id"`: the
  natural-numeric display order is owned above the store
  (`internal/workapi/sort.go` `CompareIssuesBy` via `utils.NaturalCompareIDs`).
  The store owes acceptance and completeness only, which is why that case
  passes `Limit: 0` the way its callers do.

### DIFF
--- a/backend/conformance/audit_search-counts-stats.go
+++ b/backend/conformance/audit_search-counts-stats.go
@@ -35,6 +35,7 @@ func RunAudit_search_counts_stats(t *testing.T, f Factory) {
 	t.Run("SearchDefaultOrderTieBreak", func(t *testing.T) { testAuditSearchDefaultOrderTieBreak(t, f) })
 	t.Run("SearchIdenticalTimestampIDOrder", func(t *testing.T) { testAuditSearchIdenticalTimestampIDOrder(t, f) })
 	t.Run("SearchSortByClosedNullsLast", func(t *testing.T) { testAuditSearchSortByClosedNullsLast(t, f) })
+	t.Run("SearchSortByTitleCaseFolded", func(t *testing.T) { testAuditSearchSortByTitleCaseFolded(t, f) })
 	t.Run("SearchTextIDBranchExternalRef", func(t *testing.T) { testAuditSearchTextIDBranchExternalRef(t, f) })
 	t.Run("SearchIDPrefixCaseSensitive", func(t *testing.T) { testAuditSearchIDPrefixCaseSensitive(t, f) })
 	t.Run("SearchParentDescendantCaseSensitive", func(t *testing.T) { testAuditSearchParentDescendantCaseSensitive(t, f) })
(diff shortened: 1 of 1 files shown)


julianknutsen:
Four test-only conformance cases that pin real, previously unowned ordering contracts, with expected orders that independently re-derive from `sqlbuild/sort.go` — **approve**.

## Verification performed

I re-derived every asserted order from `sqlbuild/sort.go`'s `SortDefs`, `OrderByForColumns`, and the Go-side mirror, and checked the PR's residue and design claims against the suite at main:

- **Whole orders, not lengths.** A1–A3 assert complete sequences via the order-preserving `orderedIDs` helper with `reflect.DeepEqual`; there are no length-only assertions. A4's "sequence is deliberately not asserted" claim is honest in the mechanism, not just the prose: `issueIDs` applies `sort.Strings` before comparison, so that case is a true set-completeness assertion, and the natural-order ownership it defers to (`CompareIssuesBy` via `NaturalCompareIDs` above the store) is where the code says it is.
- **The gaps are real.** Every pre-existing `SortBy: "id"` in the suite is a reader-role `ListRequest`, never store-level `SearchIssues`. The querier title-sort case titles each row with its lowercase id (its own comment says so), so a byte-collated backend passes it — the A1 case is the first to make collation observable. `SearchSortByClosedNullsLast` pins only the default direction and only `closed_at`; the flipped direction and the assignee key were unowned, as claimed.
- **A3 is representation-independent as claimed.** Under a NULL encoding the explicit `(col IS NULL)` lead term produces the asserted orders; under an empty-string encoding the term is uniformly false and `""` sorts at the same end of the plain comparison in both directions — the three asserted sequences are identical either way. The fixture also anti-correlates assignee order (alice on the newer closed row, bob on the older) with closed order, so routing one nullable key to the other cannot pass.
- **No fixture interference.** Each case takes a fresh store from the factory; legs within a case are read-only queries over one seeded fixture with distinct id prefixes.
- **Tie-break derivations check out.** A1's fold-tie (`apple`/`Apple`) pins id ASC where byte order and priority order both disagree with the case-folded order; A2's tied `updated_at` group stays id ASC under reversal on both a timestamp and a vocabulary key, matching the unconditional `id ASC` term and primary-only `flipDir`.

CI is green across all required checks, including both conformance lanes and the standalone backend-conformance oracle job. The red/green table in the description names a distinct injected divergence per case with only the owning case failing — consistent with what the fixtures are constructed to distinguish.

## Notes (non-blocking)

- The new case doc comments run longer than this file's existing 2–3 line norm. The conformance package does carry rationale comments, so I won't hold the PR on it, but the commit messages already carry the full story; the comments could be tightened in a follow-up.
- The status-sort legs pin lexicographic byte order of status strings (`closed` < `in_progress` < `open`) as portable contract. That is exactly what the reference renders (plain column, `strings.Compare` mirror), but it means a backend storing statuses as enums or ranks must reproduce string order. Intentional, just worth stating out loud.
- `SearchSortTieBreakSurvivesReversal` relies on `CreateIssue` persisting a caller-supplied `UpdatedAt`; the existing identical-timestamp case already establishes that fixture pattern, and both in-tree backends are green.

PR #4959, opened 2026-07-22
fix(gate): honor repository metadata for GitHub checks

srobroek:
## Summary

- GitHub PR and workflow gates now honor `metadata.repo`, so a Beads repository can monitor checks in another GitHub repository.
- Ad-hoc gates inherit the blocked issue's repository, and malformed repository identifiers fail explicitly instead of checking the wrong remote.
- GitHub CLI command tests use an injected argument runner and execute without POSIX shell dependencies.

## Test plan

- `go test -tags gms_pure_go -race ./cmd/bd -run 'Test.*(Gate|GH|Workflow)'`
- `GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go test -c -tags gms_pure_go -o /tmp/bd-gate-windows.test.exe ./cmd/bd`
- `golangci-lint run ./cmd/bd/...`
- `/home/sjors/.agents/skills/write-docs/scripts/slop-lint.sh --genre consumer docs/workflows/gates.md`

_codex-unknown-model-unknown-reasoning on behalf of Sjors Robroek_

PR #5409, opened 2026-08-07
fix: roll main forward to green — #5136 init regression + serve-test cache isolation

steveyegge:
Main has been red on two independent breaks; this rolls both forward.

## Shard 1/8: `TestE2E_InitDoltMetadataRoundtrip` (since 5ff61baef, #5136 — red 17:41Z today)

#5136's tri-state probe resolves a probe *error* (git ls-remote exit 128 — the credential-less private-origin shape every CI container has) as UNKNOWN → has-data. Fail-closed is right for the ADR-0002 **refusal gate**, but on a plain fresh init — no override flags, no local history to protect — `CheckRemoteSafety` then answers `ActionBootstrap`, and the bootstrap clone hard-fails against the same origin the probe couldn't reach.

**Fix:** when the decision is bootstrap but the probe errored, skip the clone and initialize a fresh local database (the behavior this path always had before the probe existed). Override flows keep #5136's fail-closed refusal unchanged — both #5136 regression tests still pass.

Also adds a **PR-lane** regression test: the E2E test that caught this runs only in the Main lane, which is why #5136's own PR ran green. Verified the new test fails on main without the fix (exact CI failure shape) and passes with it. Tracked as bd-97nv7.

## Shard 8/8: `TestServeAnswersFromARegisteredBackendStore` (since a867d98f3, #5352 — red on **every** main run since 8/5, including its own merge run)

`runServe` resolves the workspace through `beads.GetRepoContext`, a process-wide `sync.Once` that caches a prior test's verdict — **including its error**. The in-process serve tests never reset it, so whichever earlier shard-8 test leaves the cache poisoned makes this fail deterministically with 'no .beads directory found' (and the same shape hit PR Core on #4959, filed as bd-xlfu1). The sibling readonly test already resets; this brings the other two in line, plus a cleanup reset so the test's deleted tempdir isn't the *next* test's cached workspace.

cc @julianknutsen (#5352) @vishnujayvel (#5136) — FYI, no action needed; happy to adjust if either fix direction conflicts with your intent.

Targeted runs green locally (`-race -tags=integration,gms_pure_go`): both #5136 regression tests, the E2E roundtrip, all three serve tests.

### DIFF
--- a/cmd/bd/init.go
+++ b/cmd/bd/init.go
@@ -1286,7 +1286,21 @@ Non-interactive mode (--non-interactive or BD_NON_INTERACTIVE=1):
 					return err
 				}
 				if bootstrap {
-					syncFromRemote = true
+					if probeErr != nil {
+						// The fail-closed UNKNOWN kept the refusal gate shut
+						// above, but reaching here means no override flag was
+						// passed and there is no local history to protect —
+						// and the bootstrap clone would run against the same
+						// origin the probe could not reach, turning every
+						// credential-less clone of a private repo into a hard
+						// init failure. Fall back to a fresh local database,
+						// which is what init did here before the probe existed.
+						if !quiet {
+							fmt.Printf("  %s Could not verify the git origin's Dolt history; initialized a fresh local database\n", ui.RenderWarn("!"))
+						}
+					} else {
+						syncFromRemote = true
+					}
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


vishnujayvel:
Thanks for the roll-forward, and for the cc. This matches the intent of #5136 — fail-closed was about the override flows, where guessing wrong about remote history can destroy data. A credential-less fresh init has nothing at stake, so falling back to a plain local init when the probe itself errors is the right call. And good catch on the lane gap — the E2E test only running on Main is exactly why my PR run looked green. Appreciate you adding the PR-lane copy.

PR #5410, opened 2026-08-07
feat(httpapi): add include_comments and include_dependents to getIssue

julianknutsen:
## What

Adds the two additive query parameters `GET /v0/beads/issues/{id}` reserved for itself — `include_comments` and `include_dependents` — and closes the two document-level gaps the same revision window owes.

Today the operation documents comment bodies and dependents as unreachable over v0: it takes no query parameters, so `comments` and `dependents` are always absent and a workflow that needs comment text has to fork `bd show --json --include-comments`. The document names this exact pair as the additive way out ("Adding `include_comments` / `include_dependents` parameters later is additive and needs no schema change — the fields are already specified"), and `issueops.GetRequest` has carried the flags all along. This is the wire half; no role, storage or schema change is involved.

## Commits

Three, each independently reviewable:

1. **`docs(httpapi): document the JSON media-type requirement in the v0 spec`** — prose only. `requireJSONContent` refuses a body-carrying request whose `Content-Type` is not `application/json`, and the frozen document mentioned `Content-Type` nowhere, so that was the one 400 on those routes a generated client could not predict. `claim.go` has carried a note since it was written that this is prose owed at the next revision window. Stated once at the document level, beside the Host-header and unknown-query-parameter rules, because it holds for every body-carrying operation rather than any one of them. No behavior change.

2. **`docs(httpapi): widen the actor pattern to the set the server refuses`** — `validateActor` refuses every Unicode Cc control character (C0, DEL and the C1 block) plus U+2028/U+2029; the schema's `pattern` excluded only C0 and DEL, so the document advertised a narrower rule than the server enforces. `isControlChar` has carried the matching note that the pattern is the half that should move. All four schemas move together — claim, sweep, delete and batchCreate share `validateActor`, so widening one and leaving three would make the document disagree with itself about a single validator. A tightening of the published contract only: no request that used to succeed now fails.

3. **`feat(httpapi): add include_comments and include_dependents to getIssue`** — the operation itself.

## The parameters

Both optional, both `default: false`, decoded by the shared `query.boolean` (so `strconv.ParseBool`'s vocabulary). The defaults are the role's own: the two row lists are the expensive half of this read, and a caller that wants only the cardinalities keeps paying for neither.

**A request that names neither builds the same `GetRequest` the handler built before**, so existing callers see byte-identical responses. That is the property the first test asserts, and it is why this change is the safe one to land first.

`invalid_argument` joins `getIssue`'s `operationCodes` row: a malformed boolean is now this operation's own 400 — the `getStats`/`skip_blocked` precedent — rather than the document-level unknown-key rule that table deliberately omits. Adding a status+code pair to an operation is additive; clients default-branch. Against an older server these parameters earn `400` / `reason: "unknown_parameter"`, which is the documented per-parameter capability probe working as intended.

Route, path and capability (`issues.get`) are untouched: capabilities gate operations, parameters are probed.

`IssueDetails`' `comments`, `dependents` and `comments_omitted` descriptions now name the parameters that populate them. No schema changes — the fields were already specified.

## Handler ordering

The handler decodes the query **before** bounding the id, which is the order it already had when `requireNoQuery` ran first. An unknown key on an id no row could hold stays the 400 that names what to fix, instead of becoming a 404 that tells the caller nothing. `TestGetIssueAnswersAQueryRefusalBeforeTheIDBound` pins it.

## Tests

New `internal/httpapi/get_issue_test.go`, following the per-operation precedent (`ready_count_test.go`, `stats_test.go`). The seam is the role: the handler is a decoder, so the `issueops.GetRequest` it built is where the behavior is observable, and the response body is asserted alongside it so a flag forwarded to an answer nobody reads cannot pass.

| Case | Asserts |
|---|---|
| Both parameters omitted | `GetRequest{ID}` exactly — the pre-change request; `comments`/`dependents` absent |
| Both spelled `false` | Byte-identical body to omitting them — pins the default, not just the decode |
| `include_comments` alone | Flag reaches the role alone; `comments` present, `dependents` absent |
| `include_dependents` alone | The mirror — catches a handler that decoded one into the other's field |
| Both together | Both flags; both lists; `comments_omitted` cleared |
| `include_comments=maybe`, `include_dependents=maybe` | 400 `invalid_argument` / `param` naming it / `reason: invalid_value`; storage untouched |
| `?bogus=1`, and beside a known key, and a near miss (`include_comment`) | Still `unknown_parameter` — the parameter table is the whole allowlist |
| `include_comments=maybe&bogus=1` | Reported as the malformed known value, not as version skew |
| Impossible id with a refused query | The 400, not the 404 |

The values exercise `ParseBool`'s vocabulary (`true`, `1`, `TRUE`, `t`) rather than `true` three times, since that vocabulary is what the shared decoder publishes.

## Gates

All green:

- `go test ./internal/httpapi/` — including `TestSpecRouteParity`, `TestSpecStatusCodesMatchHandlerTable`, `TestSpecGovernance`, `TestSpecCapabilityVocabularyMatchesTheRouteTable`, `TestDefaultsMatchCLIFlags`, `TestWireTagBijection`, `TestClaimRequestMembersMatchTheHandler`
- `make api-check` — `types.gen.go` regenerated and committed (`GetIssueParams`, plus the `ClaimRequest.actor` doc comment)
- `make ci-pr-lint` — 0 issues, both passes
- `make ci-pr-policy`
- `make test` — full suite, 0 failures

### DIFF
--- a/internal/httpapi/apigen/types.gen.go
+++ b/internal/httpapi/apigen/types.gen.go
@@ -758,6 +758,15 @@ type ListIssuesParams struct {
 	Limit *int `form:"limit,omitempty" json:"limit,omitempty"`
 }
 
+// GetIssueParams defines parameters for GetIssue.
+type GetIssueParams struct {
+	// IncludeComments Populate `comments` with the issue's full comment bodies. When it is honored `comments_omitted` is false, so a client is never left to guess whether an absent list means "no comments" or "not asked for".
+	IncludeComments *bool `form:"include_comments,omitempty" json:"include_comments,omitempty"`
+
+	// IncludeDependents Populate `dependents` with the issues that depend on this one, each carrying its edge type — the shape `dependencies` already carries. Default false, for `include_comments`'s reason.
+	IncludeDependents *bool `form:"include_dependents,omitempty" json:"include_dependents,omitempty"`
+}
+
 // QueryIssuesParams defines parameters for QueryIssues.
 type QueryIssuesParams struct {
 	// Q The query expression, e.g. `status=open AND priority>1` or `type=bug OR label=urgent`. Blank, unparseable, or naming a field or operator the language does not have is a 400 `invalid_argument` with `param: "q"` and `reason: "invalid_value"`.
--- a/internal/httpapi/apigen/types.gen.go
+++ b/internal/httpapi/apigen/types.gen.go
@@ -184,7 +184,9 @@ type BondRef = types.BondRef
 
 // ClaimRequest defines model for ClaimRequest.
 type ClaimRequest struct {
-	// Actor Who is claiming the issue. The server trims it, then refuses an empty result, anything longer than 256 BYTES (the `maxLength` above counts characters — the byte limit is the binding one), and any control character including newline. The value is persisted as the assignee and interpolated into the storage commit message, so an unvalidated newline would forge audit-trail lines.
+	// Actor Who is claiming the issue. The server trims it, then refuses an empty result, anything longer than 256 BYTES (the `maxLength` above counts characters — the byte limit is the binding one), and any control character including newline: Unicode category Cc — C0, DEL and the C1 block — plus the U+2028/U+2029 line separators, which is the set the `pattern` above spells.
+	//
+	// The value is persisted as the assignee and interpolated into the storage commit message, so an unvalidated newline would forge audit-trail lines. C1 is refused for that same reason and not for tidiness: U+0085 is a line break on a VT-conformant terminal, and U+009B is the one-byte CSI introducer, which would make an actor an escape-sequence payload in anything that prints an assignee.
 	Actor string `json:"actor"`
 }
 
--- a/internal/httpapi/claim.go
+++ b/internal/httpapi/claim.go
@@ -149,13 +149,9 @@ func (s *Server) claimTarget(w http.ResponseWriter, r *http.Request) (string, bo
 // header name for exactly this kind of refusal — it is what the Host middleware
 // already does.
 //
-// SPEC GAP, deliberate and to be closed at the next revision window: the frozen
-// document does not mention Content-Type anywhere, so this refusal is the one
-// 400 on this route a client generated from the schema cannot predict. It is
-// unreachable for a conformant client — requestBody already declares
-// application/json — and the status/code/param/reason are all in the documented
-// vocabulary, so the fix is prose describing the CSRF control, not a behavior
-// change.
+// The document states this rule once, at the document level, beside the
+// Host-header and unknown-query-parameter rules, because it holds for every
+// body-carrying operation rather than for this one.
 func (s *Server) requireJSONContent(w http.ResponseWriter, r *http.Request) bool {
 	got := r.Header.Get("Content-Type")
 	if media, _, err := mime.ParseMediaType(got); err == nil && media == claimContentType {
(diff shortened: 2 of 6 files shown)


julianknutsen:
## Review: getIssue `include_comments` / `include_dependents` + two spec-window groundwork commits

**Verdict: approve** — the operation change is additive, well-pinned, and consistent with this surface's precedents; the two documentation commits close real doc/behavior gaps without changing behavior. The only thing standing between this and merge is CI, which is red on a flake unrelated to this diff.

### CI (blocking merge, not the code)

`CI Gate / Required` is failing via PR Core on `TestDoltServer_StartStopStart_NewInstanceSameRootDirSucceeds` (`internal/storage/dbproxy/server` — "dolt sql-server exited before listener became ready"), a package this PR does not touch. `internal/httpapi` and every other package passed in the same job, and a second gate instance is still pending. Please re-run to green before merging.

### What I verified against the head SHA

- **The no-parameter case is genuinely the pre-change request.** `TestGetIssueWithoutTheIncludeParametersIsUnchanged` asserts exact struct equality with `issueops.GetRequest{ID: "bd-1"}` *and* asserts `comments`/`dependents` absent from the response body, so the pin covers both the role seam and the wire. The explicit-`false` case is compared against the omitted case, which is what actually pins the default rather than the decode.
- **Ordering is preserved, not invented.** On main, `requireNoQuery` ran before the id bound (and the claim route orders the same way), so decoding the query before bounding the id keeps the surface-wide 400-before-404 answer. `TestGetIssueAnswersAQueryRefusalBeforeTheIDBound` pins it with an impossible id plus a refused query, and asserts zero role reads.
- **No assertions were orphaned.** Nothing was deleted: `TestUnknownReadParameterIsRefusedByName` (which covers `issues/bd-1?bogus=1` → `unknown_parameter`) and `TestGetIssueRefusesAnImpossibleIDFromTheEdge` (the 404 id bound) are untouched and still hold at head, now exercising the table-allowlist path.
- **The parity gates still see everything.** `spec_parity_test.go` and `routes.go` are unmodified. `TestSpecStatusCodesMatchHandlerTable` resolves `$ref`s before reading `x-bd-codes`, and the shared `InvalidArgument` response carries `x-bd-codes: [invalid_argument]` — so using the component ref for getIssue's new 400 (the same spelling `getStats` already uses) is fully gate-checked against the new `operationCodes` row.
- **The actor widening is a documentation tightening only.** `isControlChar` (`unicode.IsControl` = Cc: C0, DEL, C1; plus U+2028/U+2029) is untouched; the new pattern `^[^--  ]+$` excludes exactly that set, on exactly the four schemas that share `validateActor`, with zero stale narrow patterns left. Widening all four together is right — they are one validator stated four times.
- **The test fake models a held contract.** `includeAwareReader`'s behavior (rows absent + `comments_omitted` true when flags are off; rows present + `comments_omitted` unset when on) matches what `backend/conformance/reader_contract.go` asserts of the real reader implementations, in both directions.
- **The malformed-known-beside-unknown case rides existing decoder semantics** (`query.result()` reports recorded value refusals first, unknown-parameter last), not new special-casing.
- **Wire compatibility is additive**: `IssueDetails` changes are description-only, the getIssue 400 is a new response entry, and route/capability plumbing is untouched.

### Notes (non-blocking)

1. The PR body calls the explicit-`false` case "byte-identical", but the test asserts `reflect.DeepEqual` over decoded JSON maps. That is semantically sufficient to pin the default — just noting the claim is a shade stronger than the assertion.
2. The generated `GetIssueParams` struct is unused by the handler (consistent with `ListIssuesParams` etc.). That leaves a structural gap this package already documents for the claim body in `TestClaimRequestMembersMatchTheHandler`: a future spec revision adding a query parameter would leave every gate green while the server refuses it as `unknown_parameter`. A follow-up guard comparing the generated params structs to the keys each handler reads would close this for all parameterized reads; not this PR's obligation.

Issue #942, opened 2026-01-07
bd show cannot find issues that bd list can find


createsomethingtoday:
# Bug Report: `bd show` Cannot Find Issues That `bd list` Can Find

**Affects versions**: 0.43.0, 0.44.0, 0.46.0 (latest as of 2026-01-07)  
**Impact**: Blocks `gt sling` workflow automation  
**Severity**: High - Prevents Gas Town integration

---

## Summary

The `bd show <issue-id>` command fails to find issues that `bd list --id <issue-id>` successfully locates, even though the issues exist in both the SQLite database and JSONL file.

## Reproduction Steps

```bash
# 1. Create a new issue
bd create "Test issue" --label complexity:simple
# Output: ✓ Created issue: csm-5uxdj

# 2. Verify with bd list (WORKS)
bd list --id csm-5uxdj
# Output: csm-5uxdj [P2] [task] open [complexity:simple phase:review] - Test issue

# 3. Try bd show (FAILS)
bd show csm-5uxdj
# Output: Error fetching csm-5uxdj: no issue found matching "csm-5uxdj"

# 4. Try with --no-daemon (STILL FAILS)
bd show csm-5uxdj --no-daemon
# Output: Error fetching csm-5uxdj: no issue found matching "csm-5uxdj"
```

## Data Verification

Issues exist in the database:
```bash
sqlite3 .beads/beads.db "SELECT id, title, status FROM issues WHERE id = 'csm-5uxdj';"
# Output: csm-5uxdj|Test issue|open
```

Issues exist in JSONL:
```bash
grep "csm-5uxdj" .beads/issues.jsonl | jq .id
# Output: "csm-5uxdj"
```

## Environment

```bash
bd --version
# bd version 0.46.0 (812f4e52)

bd info --json | jq '{daemon_status, issue_count, mode}'
# {
#   "daemon_status": "healthy",
#   "issue_count": 1236,
#   "mode": "daemon"
# }
```

## Impact on Gas Town

The `gt sling` command internally calls `bd show <issue-id> --json` and fails with:
```
Error: checking bead status: parsing bead info: unexpected end of JSON input
```

This blocks all automatic issue assignment in the Gas Town workflow.

## Hypothesis

`bd show` and `bd list` appear to use different query paths:
- `bd list --id <id>` → ✅ Successfully finds issues
- `bd show <id>` → ❌ Returns "no issue found"

The discrepancy suggests `bd show` may be:
1. Using a different index that doesn't include newly created issues
[...]

PR #5411, opened 2026-08-07
Pin the two partial-id resolver reads and the fresh-workspace vocabulary reads

julianknutsen:
## What

Two conformance blocks over reads that CLI behavior observably depends on and
that the shared suite did not own:

- **`SearchByIDsFilter` / `SearchIssueIDsProjection`** — the two store reads the
  partial-id resolver walks (`internal/utils/id_parser.go`).
- **`Audit/config-metadata-slots-repomtime/UnconfiguredVocabulary` and
  `.../ConfiguredInfraTypes`** — the vocabulary reads every list-shaped command
  makes before it does anything else (`internal/workapi/list.go`
  `LoadListConfig`).

## Why

Both are behaviors a single backend never notices are unpinned, because its own
backend-local tests catch its regressions. A second storage backend has no
shared proof telling it what "correct" is here, and every one of these reads is
load-bearing for a user-visible path.

**Id resolution.** `bd show`, `bd update` and `bd close` all resolve their
user-typed argument through `ResolvePartialID`. Its fast path branches on
`err == nil && len(results) > 0` (#942), so a backend that reports an unknown id
as an *error* rather than an empty result silently pushes every exact lookup
onto the slow substring path — no error, no other signal. `IssueFilter.IDs`
never reached `SearchIssues` anywhere in the suite (the pinned `testGetByIDs`
covers the different `GetIssuesByIDs` method), and `SearchIssueIDs` appeared
**zero** times in `backend/conformance`. The audit's
`SearchTextIDBranchExternalRef` pins the adjacent text branch but neither the
id arm nor the id projection.

**Vocabulary.** `GetInfraTypes` appeared nowhere in the suite, and every
existing vocabulary case configures the vocabulary before reading it — so the
path a brand-new workspace actually takes, all four reads against an
unconfigured store, had no owning proof. `LoadListConfig` wraps and returns any
error from those reads, so a backend that answers a scan error or a nil-map
surprise there bricks `bd list` and `bd ready` on a fresh workspace instead of
degrading. `ConfiguredInfraTypes` pins the other half: infra types decide which
issue types route to the wisps table instead of the versioned issues table, so a
backend that ignores `types.infra` silently versions rows the workspace asked to
keep ephemeral.

## Red/green

Every case was driven red by perturbing the production behavior it pins, then
green with the perturbation reverted, against the embedded-Dolt oracle:

| Case | RED reason | Observed |
| --- | --- | --- |
| SearchByIDsFilter | `IssueFilter.IDs` clause dropped from the filter builder | `SearchIssues(IDs=[test-idf1]) = [test-idf1 test-idf2], want [test-idf1]` |
| SearchIssueIDsProjection (token) | `SearchIssueIDsInTx` ignores its query | `SearchIssueIDs("a3f8") = [test-a3f8aa test-a3f8e9 test-b111 test-wisp-b7c2], want [test-a3f8aa test-a3f8e9]` |
| SearchIssueIDsProjection (wisp) | the projection stops reading the wisp plane | `SearchIssueIDs("b7c2", Ephemeral=true) = [], want [test-wisp-b7c2]` |
| UnconfiguredVocabulary (infra) | `defaultInfraTypes` drops `message` | `GetInfraTypes ... = map[agent:true role:true], want the default agent/role/message set` |
| UnconfiguredVocabulary (statuses) | `GetCustomStatuses` errors on an empty vocabulary | `unexpected error: scan custom_statuses: no rows` |
| ConfiguredInfraTypes | `ResolveInfraTypesInTx` ignores the config value | `GetInfraTypes after types.infra=gate,probe = map[agent:true message:true role:true], want exactly gate/probe` |

## Test results

Both behaviorally distinct in-tree implementations wired to `RunAll` run every
new case green:

- `internal/storage/embeddeddolt` (the reference oracle) — new cases pass
- `internal/storage/dolt` (server-backed) — new cases pass
- `./scripts/conformance.sh` tier 1 (full embedded `TestConformance`) and tier 2
  (`./test/conformance/`, 35.6s) both green
- `./backend/... ./issueops/... . ./internal/storage/issueops/... ./internal/utils/... ./internal/workapi/...` green

Timing (the suite's budget, measured rather than assumed): the four new cases
cost ~0.4s focused. Full embedded `TestConformance` went 63.07s before → 64.47s
after for the whole series including the sibling claim-role PR; this PR is the
smaller half of that.

## Notes

Each case names the subject it binds — `SearchIssues` with `IssueFilter.IDs`,
`SearchIssueIDs`, `SetConfig` — so a partial backend whose declared unsupported
allowlist refuses that subject knows to leave the case out of its
supported-subset gate rather than discovering it unrunnable inside `RunAll`.

The claim-role block from the same design lands separately in
`conformance-claim-role`: it is a new file with no dependency on this one, and
splitting keeps both diffs reviewable.

### DIFF
--- a/backend/conformance/audit_config_metadata_slots_repomtime.go
+++ b/backend/conformance/audit_config_metadata_slots_repomtime.go
@@ -1,6 +1,7 @@
 package conformance
 
 import (
+	"maps"
 	"slices"
 	"sort"
 	"strings"
@@ -147,6 +148,70 @@ func testAuditDeleteConfigLeavesNormalizedTable(t *testing.T, f Factory) {
 	}
 }
 
+// testAuditUnconfiguredVocabulary pins the success path a fresh workspace takes
+// on its very first list-shaped command: with no custom vocabulary configured,
+// GetCustomStatuses, GetCustomStatusesDetailed and GetCustomTypes answer empty
+// with a nil error, and GetInfraTypes answers the default infra set.
+//
+// Every existing vocabulary case configures the vocabulary first, so the
+// unconfigured path had no owning proof and GetInfraTypes had none at all.
+// The consumption is a hard failure edge: internal/workapi/list.go LoadListConfig
+// loads all four up front and wraps any error, so a backend that answers a scan
+// error or a nil-map surprise for an unconfigured workspace bricks `bd list`,
+// `bd ready` and every other list-shaped command rather than degrading.
+//
+// The infra set is spelled out rather than compared against the production
+// constant: this suite is the contract, and reading the answer from the same
+// place the implementation reads it would assert nothing.
+func testAuditUnconfiguredVocabulary(t *testing.T, f Factory) {
+	s := f(t)
+	c := ctx()
+
+	names, err := s.GetCustomStatuses(c)
+	must(t, err)
+	if len(names) != 0 {
+		t.Errorf("GetCustomStatuses on an unconfigured workspace = %v, want empty", names)
+	}
+
+	detailed, err := s.GetCustomStatusesDetailed(c)
+	must(t, err)
+	if len(detailed) != 0 {
+		t.Errorf("GetCustomStatusesDetailed on an unconfigured workspace = %v, want empty", auditDetailedPairs(detailed))
+	}
+
+	custom, err := s.GetCustomTypes(c)
+	must(t, err)
+	if len(custom) != 0 {
+		t.Errorf("GetCustomTypes on an unconfigured workspace = %v, want empty", custom)
+	}
+
+	infra := s.GetInfraTypes(c)
+	if !maps.Equal(infra, map[string]bool{"agent": true, "role": true, "message": true}) {
+		t.Errorf("GetInfraTypes on an unconfigured workspace = %v, want the default agent/role/message set", infra)
+	}
+}
+
+// testAuditConfiguredInfraTypes pins the other half: a configured types.infra
+// key replaces the default set outright, and the answer is exactly the
+// configured names. Infra types decide which issue types route to the wisps
+// table instead of the versioned issues table, so a backend that ignores the
+// key silently versions rows the workspace asked to keep ephemeral.
+//
+// Subject: SetConfig. A backend whose allowlist refuses it does not run this
+// case. The assertion is the value, not the cache: invalidation is a per-store
+// concern this suite declines to pin, which is why the read happens on a store
+// that has not read the key before.
+func testAuditConfiguredInfraTypes(t *testing.T, f Factory) {
+	s := f(t)
+	c := ctx()
+	must(t, s.SetConfig(c, "types.infra", "gate,probe"))
+
+	infra := s.GetInfraTypes(c)
+	if !maps.Equal(infra, map[string]bool{"gate": true, "probe": true}) {
+		t.Errorf("GetInfraTypes after types.infra=gate,probe = %v, want exactly gate/probe", infra)
+	}
+}
+
 // testAuditRepoMtimeClearKeyAsymmetry pins the verbatim-key vs abs-normalized-key
 // asymmetry: SetRepoMtime/GetRepoMtime use the path argument verbatim, but
 // ClearRepoMtime first resolves it to an absolute path before the DELETE. So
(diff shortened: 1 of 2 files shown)


julianknutsen:
**Two well-aimed conformance blocks — the partial-id resolver's store reads and the fresh-workspace vocabulary reads — with every load-bearing claim verified in code. Approve.**

The diff is additive-only: four new cases in `backend/conformance` plus their `RunAll`/`RunAudit` registrations. No production code is touched, so the only review question is whether the cases pin what they say they pin. They do.

## What I verified

**The coverage gaps are real.** At current `main`, `SearchIssueIDs` appears zero times in `backend/conformance`, and no case passes `types.IssueFilter{IDs}` to `SearchIssues` — every existing `IDs:` usage in the suite belongs to `BlockingRequest`/`DeleteRequest`/`EdgeReadRequest`/`ReclaimFilter`, and `testGetByIDs` exercises the separate `GetIssuesByIDs` method. Both reads are load-bearing: `bd show`/`bd update`/`bd close` resolve their argument through `internal/utils/id_parser.go`, which walks exactly these two calls.

**The nil-error miss assertion is the right one.** The resolver's fast path is literally `if issues, err := store.SearchIssues(ctx, "", exactFilter); err == nil && len(issues) > 0` — a backend that reports an unknown id as an error rather than an empty result silently pushes every exact lookup onto the substring path with no other signal. `SearchByIDsFilter` pins precisely that.

**The token choice in `SearchIssueIDsProjection` matches the real consumer contract.** `sqlbuild.LooksLikeIssueID("a3f8")` is false (the classifier requires an interior hyphen), so the case exercises the `(LOWER(title) LIKE ? OR id LIKE ?)` arm — and that is the only arm the production consumer can reach: `ResolvePartialID` strips prefixes before querying (`partialIDSearchPart` reduces the query to the post-last-hyphen suffix, and a trailing-hyphen remainder also fails `LooksLikeIssueID`), so `SearchIssueIDs` never sees an id-shaped token in production. The case comment describing "the hash token" is the observable truth, and the fixture titles (`One`/`Two`/`Three`/`Wisp`) correctly avoid contaminating the id-arm assertion through the title arm. The `Ephemeral: true` leg is likewise real: `issueops.searchInTx` routes it to the wisps tables, mirroring the resolver's wisp fallback.

**The vocabulary cases assert the contract, not the plumbing.** `GetInfraTypes` had zero suite coverage; the unconfigured-workspace path (`internal/workapi/list.go` `LoadListConfig`, which wraps any error from all four reads before every list-shaped command) had no owning proof. The spelled-out `agent`/`role`/`message` map matches `domain.DefaultInfraTypes()` without importing it — correct for a contract suite, since reading the answer from where the implementation reads it would assert nothing. Signatures line up (`GetInfraTypes` returns the map with no error; the three custom reads return `(value, error)` and the nil error is asserted). `ConfiguredInfraTypes` reads on a store that has never read the key, so the assertion is the value and not cache invalidation, as its comment claims.

**Red/green evidence is consistent with the code.** Each perturbation in the PR table maps to a distinct production seam I could locate (the filter builder's `IDs` clause, `SearchIssueIDsInTx`'s query, the wisp-plane routing, `defaultInfraTypes`, `ResolveInfraTypesInTx`), and the reported failure outputs match the assertion messages in the diff.

## Blocking

None.

## Notes (non-blocking)

1. `UnconfiguredVocabulary` is the one new case without an explicit `Subject:` line in its doc comment; the other three name their subject (`SearchIssues` with `IssueFilter.IDs`, `SearchIssueIDs`, `SetConfig`) for partial-backend allowlist purposes. Worth adding for consistency next time the file is touched.
2. The id-shaped `id LIKE '<query>%'` prefix arm of the text-query branch remains unpinned suite-wide: `SearchTextIDBranchExternalRef` covers the exact `id = ?` and `external_ref LIKE` arms, and this PR covers the non-id-shaped arm. That prefix arm is only reachable through `SearchIssues` text queries (an id-shaped user query), not through the resolver, so it belongs as a follow-up leg on the existing audit case rather than in this PR.
3. A handful of `Test (Embedded Dolt Cmd N/20)` shards were still pending at review time. Everything completed is green — including the conformance oracle lane and Server Dolt Conformance — but merge should wait for the shards to finish.

PR #5412, opened 2026-08-07
Pin the Claimer role at its own seam (four contract cases)

julianknutsen:
## What

A new `backend/conformance/issue_claimer_contract.go`: four cases pinning the
`issueops.Claimer` role reached through `Storage.IssueClaimer()`, plus the
retirement of the one backend-local test the last of them supersedes.

## Why

The suite already pins the claim's sentinels on the store **method**
(`claim.go`: `testClaim`, `testClaimIdempotent`, `testClaimAlreadyClaimed`,
`testClaimOpenForeignAssignee`, `testClaimNotClaimable`) and the typed conflict
payload at the **Update seam** (`issue_operations_contract.go`). Neither reaches
what the role promises, on two axes.

**Different producers.** The role's outermost refusal classifiers rebuild the
conflict payload from their own in-transaction re-read —
`internal/storage/issueops/public_claim.go` `classifyClaimRefusalInTx` and
`internal/storage/uow/issue_claimer.go` `classifyClaimError` — and the
Update-seam cases execute neither. `ClaimResult` itself is structurally
invisible to `claim.go`, whose subject returns a bare `error`: `Changed`, the
post-state snapshot, and the promise that the snapshot is the **bare row** had
no owning proof anywhere, and neither did the unknown-id, wisp-id and
empty-actor refusals.

**Different reachability.** The Update-seam cases run only behind
`IssueOperationsStagingFixture`, which demands raw-SQL hooks (`Exec`,
`QueryScalar`, `UpdateRaw`) a partial or remote backend cannot supply. Every
case here binds the accessor and the public role API and nothing else, so any
backend that can answer `IssueClaimer()` runs them. That is why they are
fixture-style and individually invokable: the fixture is constructible from a
factory store alone (accessor + `CreateIssue` + `GetIssue`), `RunAll` composes
the store-backed wiring so full backends get them with zero new wiring, and a
partial backend wires the same runners piecewise.

## The cases

**`RunClaimWinsThenReclaimsIdempotently`** — the win reports a snapshot with
`Changed` true; the holder's re-claim succeeds with `Changed` false and records
no second claim event. The seeded issue carries a label the snapshot must not
report: the bare row is what the v0 claim response and `bd update --claim
--json` already publish (`issueops/claimer.go:14-24`), so hydrating here would
change a shipped body.

**`RunClaimForeignHolderRefusalCarriesTheHolder`** — a deliberate role-seam
re-pin of the Update-seam payload assertions. The overlap is named rather than
hidden, and it earns its place: neutering only the inner producer leaves this
case green, because the role's own classifier rebuilds the payload; neutering
both turns it red. That is the layer the Update-seam cases never execute.

**`RunClaimRefusesIneligibleUnknownAndActorlessRequests`** — `ErrNotClaimable`
for an ineligible status, `ErrNotFound` for an unknown id **and for a wisp id**
("the wisp plane is not claimable through this role",
`issueops/claimer.go:63-64`), `ErrValidation` for a missing actor — each with
the leave-persistent-state-unchanged clause the leaf attaches to it. The
claimable row seeded for the actorless leg is the point of its state half: that
request would have won, so an implementation validating after claiming fails on
the row rather than on the error.

**`RunClaimRefusalMessagesCarryTheirFragments`** — the string coupling
`beads.ParseClaimConflict` depends on. It rebuilds its marker as
`sentinel.Error()+fragment` from the exported `storage.ClaimedByFragment` /
`storage.NotClaimableStatusFragment` rather than hardcoding a literal, so a
backend whose refusal drops the fragment hands that parser an empty field with
no other signal. Three shapes, three obligations: `" by <assignee>"` on the
in_progress conflict, `": status <status>"` on the not-claimable refusal, and
the **deliberate omission** on the open-but-foreign-assigned refusal, which
answers holder-steering copy instead — pinned so a backend that "fixed" the
omission is caught rather than silently changing published copy.

## Supersession

`internal/storage/dolt/claim_conflict_test.go` is deleted here. Its whole
content was the same fragment round-trip against one backend, while the parser
it protects is backend-independent. The deletion rides the change that
supersedes it, so no window exists where the behavior has zero owning proofs;
the fragment constants' doc comment now names its replacement.

## Red/green

Each case was driven red by perturbing the production behavior it pins, then
green with the perturbation reverted, against the embedded-Dolt oracle:

| Case | RED reason | Observed |
| --- | --- | --- |
| WinsThenReclaimsIdempotently | `changed` forced true; the snapshot stops clearing relations | `snapshot carries relations (labels [clr-win-label], ...), want the bare row` / `Changed = true on the holder's re-claim, want false` |
| ForeignHolderRefusalCarriesTheHolder | inner producer returns the bare refusal | still **green** — the role's own classifier rebuilds the payload (this is the producer the Update seam never runs) |
| ForeignHolderRefusalCarriesTheHolder | inner producer **and** `classifyClaimRefusalInTx` neutered | `refusal for clr-held is not a *issueops.ClaimConflictError (issue already claimed by worker1)` |
| RefusesIneligibleUnknownAndActorlessRequests | wisp refusal, actor validation, not-found wrap and the not-claimable sentinel each broken | `claim of a wisp id = <nil>, want ErrNotFound` + `after the refused claim of a wisp, clr-wisp-refuse is {status "in_progress" assignee "worker"}`; `claim with no actor = <nil>, want ErrValidation` + the actorless row claimed; `claim of an unknown id = failed to get issue for claim: not found ..., want ErrNotFound`; `claim of a closed issue = issue already claimed: status closed, want ErrNotClaimable` |
| RefusalMessagesCarryTheirFragments | each of the three message shapes rewritten | `assignee recovered from the in_progress conflict = "", want alice`; `status recovered from the not-claimable refusal = "", want "closed"`; `the open-but-assigned refusal carried the " by " fragment (recovered "alice"); that copy omits it on purpose` |

## Test results

Both behaviorally distinct in-tree implementations wired to `RunAll` run all
four cases green:

- `internal/storage/embeddeddolt` (the reference oracle) — pass
- `internal/storage/dolt` (server-backed) — pass
- `./scripts/conformance.sh` tier 1 (full embedded `TestConformance`) and tier 2
  (`./test/conformance/`, 35.6s) both green
- `./backend/... ./issueops/... . ./internal/storage/issueops/...` green, and
  the full `internal/storage/dolt` conformance run after the test deletion

Timing (measured rather than assumed): the four cases cost ~5.7s focused on
embedded Dolt, ~0.5s on the server-backed store. Full embedded
`TestConformance` went 63.07s before → 64.47s after for this and the sibling
id/vocabulary PR together.

## Notes

`RunClaimerRole(t, factory)` is the per-block entry point `RunAll` composes; the
four `RunClaim*(t, ctx, fixture)` runners are exported individually so a
supported-subset gate can adopt them piecewise. A backend whose declared
unsupported allowlist refuses `IssueClaimer()` runs none of them.

Deliberately **not** in this contract: the lease a winning claim grants (already
pinned on the store method by `claim.go`), and two racing claimants producing
one winner — a transaction property of the persistence boundary rather than a
promised result of the operation.

The id-resolution and vocabulary blocks from the same design land separately in
`conformance-id-vocab`; neither PR depends on the other.


julianknutsen:
## Review: Pin the Claimer role at its own seam (four contract cases)

**Summary:** The contract design and its evidence hold up under adversarial spot-checking; the only blocker is a lint failure that turns the required CI gate red. **Verdict: request changes** (one-line fix, then this is mergeable).

### Blocking

1. **Required CI gate is red on an `unparam` lint finding.** `backend/conformance/issue_claimer_contract.go:315` — `claimerFixtureFromStore`'s `prefix` parameter always receives `"clr"`. Since `RunClaimerRole` gives each case a fresh store via `factory(t)`, the prefix carries no load in that wiring; the simplest fix is to drop the parameter and inline the constant (the `IssuePrefix` field on the fixture still serves piecewise adopters that share a database). `Lint`, `PR Lint`, and `CI Gate / Required` all fail on this one issue; everything else that has completed is green, with a handful of test shards still pending at review time.

### What I verified (not just read)

- **The deleted test's assertions have a real new owner.** `internal/storage/dolt/claim_conflict_test.go` asserted two fragment round-trips against the store method. Both are re-asserted by `RunClaimRefusalMessagesCarryTheirFragments` with the same marker-reconstruction technique, now against the role on *both* RunAll backends instead of one, plus the deliberate-omission leg the old test never covered. The message producer is genuinely shared — `DoltStore.ClaimIssue` and the role both run `issueops.ClaimIssueInTx`, which composes the fragments, and `ClaimConflictError.Error()` is a passthrough — so in-tree fragment coverage is strictly widened, not moved sideways. The `errors.Is` halves stay owned by `backend/conformance/claim.go`, and the `storage.go` doc comment correctly names the replacement. No zero-owner window.
- **The bare-snapshot case pins behavior that actually exists.** `ExecuteClaim` explicitly nils `Labels`/`Dependencies`/`Comments` with the shipped-v0-body rationale, and the `ClaimResult` doc freezes it; the case seeds a labeled issue and asserts the snapshot omits it.
- **The two-stage red/green story for the foreign-holder case is structurally sound.** Both wired backends route `Claim` → `ExecuteClaim` → `classifyClaimRefusalInTx`. The inner producer (`claim.go:170`) already returns a typed conflict, and the role classifier independently rebuilds one from its own in-transaction re-read — so neutering only the inner producer indeed leaves the case green, and neutering both makes the case's `errors.As` fatal. The case binds the classifier layer as claimed.
- **The wisp and event legs are not vacuous plumbing.** `ExecuteClaim` refuses wisp ids before any write, `CreateIssue` routes `Ephemeral` rows to the wisps plane, `GetIssueInTx` falls back to the wisps table (so the unchanged-state re-read is meaningful), and `GetEvents(limit<=0)` returns all events (so the claim-event delta is a real count).

### Non-blocking notes

- **Silent nil path on `CountClaimEvents`.** The `CountHistory` precedent this hook follows announces a narrowed subject at runtime (`t.Skip` in the commenter/counter/cycle-detector contracts; `t.Log` in `dependency_editor_contract.go`). Here, a piecewise adopter that leaves the hook nil silently loses the persisted-nothing event check. A `t.Log` on the nil path would match package precedent without skipping the case's remaining assertions.
- **The event check is delta-only.** A backend that records no claim events at all passes with 0 → 0. That is defensible — the role doc does not promise event recording — but if the intent is "records no *second* claim event", asserting the win recorded at least one (when the hook is non-nil) would make the delta non-vacuous.
- **The second named classifier is never exercised.** The PR motivates these cases by naming both role classifiers, but nothing wires the new runners to the unit-of-work provider — `RunAll`'s callers are the two stores, and no test drives the uow `Claimer` through these cases. I checked that the uow/domain stack would *likely* pass (its producer composes the same three refusal shapes, including the deliberate omission, and its row read does not hydrate relations), but "likely" is exactly the gap this contract argues against. The fixture-style runners make a piecewise uow wiring a natural, small follow-up.
- The conformance timing before/after is measured jointly with the sibling PR; worth restating per-PR once the sibling lands.

The four cases assert observable results over seeded fixtures, state their admission rule, name their overlaps with the Update-seam cases instead of hiding them, and ride the supersession in the same change. Once the lint finding is fixed and the pending shards complete green, I would approve.

julianknutsen:
The one blocking item from the review is resolved at e32067359 (parameter dropped, constant named; lint verified failing before and clean after). Approve. — automated review by Claude (Fable 5), posted by the maintainer's orchestration

PR #5414, opened 2026-08-07
test(conformance): pin keyset paging and the MaxRows cap at the storage seam

julianknutsen:
## What

Adds the keyset-paging and row-cap block to the shared storage conformance suite: a new `backend/conformance/search_paging.go` behind an exported, individually-invokable `RunSearchPaging` entry point that `RunAll` composes. Then retires the backend-local tests it supersedes.

Four cases, all asserting observable results over seeded fixtures:

| Case | Pins |
| --- | --- |
| `KeysetWalkLosesNothingAndRepeatsNothing` | A page-by-page walk over a same-second group larger than the page answers the one-shot sequence exactly — nothing dropped, nothing repeated. |
| `KeysetResumeAnswersTheStrictlyAfterRows` | A mid-group cursor answers exactly the strictly-after tail; an empty `AfterID` starts the same-second group from its first id. |
| `KeysetPositionComposesWithTheFilter` | The position narrows what the other predicates matched, and does not replace `CreatedBefore`. |
| `MaxRowsAnswersTheTypedRefusal` | The typed `*issueops.ErrTooManyRows` with `Found`, `Cap`, and the caller's `Source` echoed verbatim; at-or-under the cap unaffected; `MaxRows: 0` disabled; a `Limit` at or under the cap cannot fire it, one above it still does. |

## Why

`SearchIssues` carries two documented promises that no shared contract case owned:

- the keyset position (`IssueFilter.AfterCreatedAt`/`AfterID`, `types.go:1791-1806`) — what `issueops/reader.go:277-280` calls the portable way to page;
- the defensive row cap (`IssueFilter.MaxRows`/`MaxRowsSource`, `types.go:1889-1897`).

Both were pinned only by **backend-local** tests, and in both cases by a duplicated `dolt` + `embeddeddolt` pair. A duplicated pair is the signal that the assertion belongs in the shared suite: the second copy exists because the contract had nowhere else to live. A backend that hand-writes its search query had no shared proof of either behavior, and the cap's typed shape is load-bearing for callers — `cmd/bd/max_rows.go` classifies it with `errors.As` and converts exactly this error to exit code 2, so a backend answering any other shape silently loses the circuit breaker while every row-count assertion still passes.

The reader role owns adjacent but genuinely weaker cases, so these are not duplicates of it:

- `RunReaderListKeysetPositionResumesTheCreatedDescIDAscOrder` makes **one** boundary read at the role level. It exercises neither the overflow walk, nor composition with other predicates, nor the `AfterID: ""` group start, and its subject is the role rather than `SearchIssues`.
- `RunReaderListMaxRowsIsHonoredOrRefused` is deliberately honored-**or**-refused, because a role implementation may decline the cap. The storage layer's own obligation is unconditional honoring; that stricter branch is what this block asserts.

`sqlbuild/keyset_test.go` pins only the SQL text — plumbing, not the observable result.

## Admission rule

Every case here binds the `SearchIssues` filter surface (keyset position, `MaxRows`). `RunSearchPaging` is exported individually so a backend whose declared unsupported allowlist refuses that subject can compose a curated supported-subset gate from the per-block runners whose subjects it does supply — the `RunDeferredReads` precedent — and answer `RunUnsupportedContract` for the rest. A backend adopting this block retires the backend-local copies it supersedes in the same change, so no behavior ever has zero owning proofs.

## Test migration

Behavior-neutral: every retired assertion names its new owner in the demotion commit message. Summary:

- `internal/storage/dolt/keyset_search_test.go` — **partially** retired. The overflow walk and boundary resume move to B1/B2. `TestSearchIssuesKeysetPlanIsIndexed` is **kept**: EXPLAIN sargability is Dolt planner residue, outside the observable contract, and cannot move.
- `internal/storage/embeddeddolt/keyset_search_test.go` — retired, superseded by B1/B2 through this backend's own `RunAll`.
- `internal/storage/dolt/search_max_rows_test.go` and `internal/storage/embeddeddolt/search_max_rows_test.go` — retired, superseded by B4.

The deletions ride the same PR as the block that supersedes them, so no window exists where a behavior has zero owning proofs.

The shared cases are stronger than what they retire on three counts: the retired cap tests asserted only `len(results)` where the new legs assert the whole id set; the at-the-cap boundary (`cap == matching rows`) was never covered; and the `AfterID: ""` group start and filter composition had no backend-local copy at all.

## Verification

Run against the embedded-Dolt reference oracle, which drives the suite via `RunAll`:

```bash
BEADS_TEST_EMBEDDED_DOLT=1 CGO_ENABLED=1 go test -tags gms_pure_go -timeout 30m \
  ./internal/storage/embeddeddolt/ -run TestConformance
```

All four cases pass on first run: they close a shared-suite gap, not a behavior gap, so no backend divergence was found and nothing was parked.

Because every case pins existing behavior rather than landing red on a divergence, each was proven load-bearing by mutating the production code it guards and confirming the case — and only the right case — fails:

| Mutation | RED |
| --- | --- |
| Keyset tie-break made inclusive (`id > ?` → `id >= ?`) | B1 (repeated `ks-a1`), B2, B3. B4 unaffected. |
| Keyset degraded to a `created_at`-only cursor | B1 (walk answers `[ks-newer ks-a1 ks-older]`, losing the overflow), B2, B3. B4 unaffected. |
| Keyset made to displace `CreatedBefore` | B3 only, on the `CreatedBefore` leg. |
| `EnforceMaxRowsCap` never fires | B4 cap-fires legs only. Keyset cases unaffected. |
| `ErrTooManyRows.Source` dropped | B4 attribution legs only, on both source strings. |

Also green: `./backend/conformance/...`, `./internal/storage/sqlbuild/...`, `./internal/storage/issueops/...`, and `make ci-pr-lint` (0 issues, including the Windows cross-lint).

### Timing

Per the suite's measure step:

- New cases, focused: **8.5s** on the embedded-Dolt oracle (the first case absorbs store warm-up).
- Affected conformance lane (`scripts/conformance.sh` Tier 1, embedded-Dolt `TestConformance`): **65.2s → 70.9s** (+5.6s, +8.7%).
- Tier 2 (the end-to-end CLI corpus) is unaffected — no command surface changed.

## Deferred

`SearchIssuesWithCounts` shares the filter builder with `SearchIssues`. Whether B3/B4 each owe one composition leg through the counts projection is deferred rather than decided here: the shared-builder unit coverage plus these `SearchIssues` cases are sufficient for the behaviors this PR claims, and adding a counts leg is a self-contained follow-up if a reviewer wants it.


julianknutsen:
**Summary:** The new `RunSearchPaging` block is well-built and the demotion map is honest — but the deletion commit missed the embedded-storage CI shard manifest, and four `Test (Embedded Dolt Storage N/5)` jobs are already red because of it. **Verdict: request changes** (one mechanical fix).

## Blocking

### Stale shard-manifest entries fail every embedded-storage shard job

`.github/scripts/embedded-storage-test-shards.txt` still lists five top-level tests this PR deletes from `internal/storage/embeddeddolt/`:

- `TestSearchIssuesKeysetEmbedded`
- `TestSearchIssues_MaxRows_NotExceeded`
- `TestSearchIssues_MaxRows_Exceeded_ReturnsErrTooManyRows`
- `TestSearchIssues_MaxRows_Zero_NoCap`
- `TestSearchIssues_MaxRows_WithLimit`

`embedded-storage-test-shard.sh` discovers tests from `internal/storage/embeddeddolt/*_test.go` and hard-fails on any manifest entry with no matching discovered test ("Shard manifest entry does not match a discovered test", `exit 1`) — before running anything. Both `main.yml` and `pr-risk.yml` invoke it with 5 shards, so all five shard jobs fail deterministically; shards 1/5–4/5 are already red on the current run (14–44s runtimes, consistent with the pre-flight validation).

**Fix:** delete those five lines from the manifest in the demotion commit. Nothing needs re-adding — unlisted tests hash-distribute — and the dolt-side deletions don't appear in this manifest. Worth noting for the retirement habit generally: a `BEADS_TEST_SHARD_LIST_ONLY=1` pass over the shards (or running the touched CI lane) catches this class before push; the verification list here (focused conformance run, package tests, `ci-pr-lint`) structurally could not.

## Verified

I treated the self-reported evidence adversarially and read the retired files at the pre-deletion commit against the new block at head:

- **Demotion map holds.** The dolt keyset test's three legs (one-shot `(created_at DESC, id ASC)` order, page-size-2 walk with no-drop/no-dup and page-within-Limit, resume-after-`a3` = `[a4 a5 older]`) each have an equal-or-stronger owner in `KeysetWalkLosesNothingAndRepeatsNothing`/`KeysetResumeAnswersTheStrictlyAfterRows`, with the same fixture shape. All eight max-rows assertions (both backends) map onto `MaxRowsAnswersTheTypedRefusal` legs. The claimed strengthenings are real: the retired cap tests asserted only `len(results)` while the new legs assert the full id set, and the at-cap boundary (`MaxRows == matching rows`) is genuinely new. The embedded keyset twin had only two legs (no resume), and the commit message says exactly that.
- **Composition and reachability.** `RunAll` now runs `SearchPaging`, and both in-tree backends drive `RunAll` through their `conformance_test.go`, so the retired assertions are re-reached on both backends with no new wiring. `RunSearchPaging` is exported and individually invokable, and its doc comment carries the admission rule (supported-subset gate on the `RunDeferredReads` precedent, same-change supersession).
- **"Reader role is weaker" is accurate.** `RunReaderListKeysetPositionResumesTheCreatedDescIDAscOrder` makes one boundary read; `RunReaderListMaxRowsIsHonoredOrRefused` has an explicit `ErrUnsupported` refusal branch. The storage block's unconditional-honoring is a genuinely stricter obligation, so this is not a duplicate scenario at a second layer.
- **Pinned semantics match the documented contract.** The `IssueFilter` docs promise the strictly-after `(created_at, id)` predicate, the `AfterID: ""` group start, composition with `CreatedBefore` ("which it does not replace"), and effective `LIMIT min(Limit, MaxRows+1)`; `cmd/bd/max_rows.go` classifies with `errors.As` and converts to exit code 2, confirming the typed-shape motivation.
- **The kept EXPLAIN test.** `TestSearchIssuesKeysetPlanIsIndexed` survives in the dolt file with a pointer comment; keeping planner sargability in-tree is the right call — it is engine residue, not observable contract.

## Notes (non-blocking)

- B4 builds a fresh store per subtest leg. The cost is measured and disclosed (+5.6s on the affected lane), so fine — just noting the pattern compounds as legs get added.
- Deferring the `SearchIssuesWithCounts` composition leg is reasonable and stated rather than silent; happy to see it as a follow-up.
- The file is comment-heavy relative to the contributor guideline on inline comments, but it matches the established doc-comment-as-contract style of the conformance package, so no change requested.

Once the manifest lines are removed and the storage shards (plus the still-pending conformance lanes) come back green, this is a clean approve — the block itself needs no changes.

julianknutsen:
The blocking item from the review is resolved at b9c662918: the five stale manifest entries are removed inside the demotion commit itself (deletions and the entries they invalidate move together), with shard-resolution evidence for all five shards and a repo-wide name sweep for every deleted test function. One correction accepted from the author: the stale entries killed all five shards, not four — the validation loop checks every manifest entry regardless of the shard being run. Approve. — automated review by Claude (Fable 5), posted by the maintainer's orchestration

PR #5415, opened 2026-08-07
Give the two dependency-endpoint refusals a typed identity (bd-yby99.9)

julianknutsen:
## What

`DependencyEditor.AddDependencies` now refuses a nonexistent edge SOURCE with
`issueops.ErrDependencySourceNotFound`, and a locally-visible missing TARGET
with `issueops.ErrDependencyTargetNotFound`. Each is carried by a
`*DependencyEndpointNotFoundError` naming the refused edge and which of its two
endpoints was absent. All three are re-exported from the root package beside the
existing `DependencyTypeConflictError` / `DependencyHierarchyConflictError`.

Which edges are ACCEPTED does not change. An `external:` reference and an id
belonging to another repository are still stored as external targets, and
`RemoveDependency` is untouched — a removal that finds no edge is `Removed`
false with a nil error, so it has no refusal to name.

## Why

Every other refusal this role can raise already hands a caller something to
branch on: `ErrSelfDependency`, `ErrDependencyCycle`,
`*DependencyTypeConflictError`, `*DependencyHierarchyConflictError`. The two
endpoint-existence refusals handed back "an error" — and not the same error text
on every implementation — so a programmatic consumer could not tell a ghost
endpoint from an infrastructure failure, or a bad source from a bad target,
without string-matching prose the role explicitly does not promise.

`issueops/dependencyeditor.go` said as much in its own leaf text: the anonymity
was a stated gap (`bd-yby99.9`), not a promise. This closes it, and the leaf now
promises the identity instead.

WHERE the classification happens is the same rule `ClaimConflictError` follows:
inside the transaction that refused.

- The store-backed body (`internal/storage/issueops/dependencies.go`) and the
  cross-tier target precheck (`internal/storage/dolt/transaction.go`) already
  read the endpoint row, so they mint the refusal there. The message they render
  is unchanged, byte for byte.
- The domain repository (`internal/storage/domain/db/dependency.go`) has no such
  read — it learns of the absence from a foreign-key violation. It reads the two
  endpoints back on the runner that refused, **only on the refusal path**, so a
  bulk `bd dep add --file` still pays no probe per edge. Taking the identity out
  of the driver's constraint name would have been exactly the prose-parsing a
  typed refusal exists to avoid. A probe that cannot settle the question returns
  the original error untouched: a refusal is never downgraded to a probe's
  failure.

One user-visible consequence, and it is a convergence: the proxied-server path
now renders `issue <id> not found` where it used to render a raw
`Error 1452 ... Foreign key violation`, which is what the embedded path has
always said.

The guarded create classifies the new type where it classified the raw
foreign-key signal, so a missing dependency, parent or waits-for target is still
`ErrValidation` wrapping `ErrNotFound` on every backend
(`RunIssueOperationsCreateRejectsMissingDependencyTargets` is unchanged and
still green).

## Verification

The owning proofs are the two existing contract cases in
`backend/conformance/dependency_editor_contract.go` —
`RunDependencyEditorRefusesAGhostSource` and
`RunDependencyEditorRefusesAMissingLocalTarget` — strengthened rather than
duplicated. Each now asserts the sentinel and the typed fields in **two
positions**: alone in a request, and mid-batch, where the refusal competes with
the rollback of an edge already written. The mid-batch half also reads the graph
back at zero edges, because a typed refusal that left half a graph behind would
still be the wrong answer. Both run against all three `DependencyEditor`
implementations: the direct store, the embedded store, and the unit-of-work
provider.

Both cases were observed RED on all three backends, in both positions, before
the production change:

```
# internal/storage/dolt (direct store)
the sole edge of a request:   error = issue ghostsrc-ghostsrc-know not found, want errors.Is dependency source not found
the second edge of a request: error = issue notgt-notgt-ghost not found,      want errors.Is dependency target not found

# internal/storage/embeddeddolt (embedded store)
the sole edge of a request:   error = issue notgt-notgt-ghost not found,      want errors.Is dependency target not found

# internal/storage/uow (unit-of-work provider)
the second edge of a request: error = add deps[1]: insert: db: DependencySQLRepository.Insert:
  Error 1452 (HY000): cannot add or update a child row - Foreign key violation on fk: fk_dep_issue,
  ...  want errors.Is dependency source not found
```

GREEN afterwards, alongside the value's own pure proof
(`TestDependencyEndpointNotFoundCarriesTheAbsentEndpoint`) and the root-package
re-export identity proofs:

```bash
go test ./internal/storage/uow/ -count=1
go test ./internal/storage/dolt/ -run 'TestDependencyEditor|TestIssueOperations|TestRunInTransactionCrossTier' -count=1
BEADS_TEST_EMBEDDED_DOLT=1 go test ./internal/storage/embeddeddolt/ -run 'TestEmbeddedDependencyEditor|TestEmbeddedIssueOperations' -count=1
go test . ./issueops/ ./internal/storage/domain/... ./internal/storage/issueops/ -count=1
make ci-pr-lint
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **The two dependency-endpoint refusals now carry an identity** (bd-yby99.9).
+  `DependencyEditor.AddDependencies` refuses an edge whose SOURCE names no row,
+  and one whose TARGET names no row this database can see the absence of. Both
+  were "an error" and nothing more — not the same error text on every backend,
+  and impossible to tell apart from an infrastructure failure — which left the
+  role's typed-refusal vocabulary with a hole every other refusal it can raise
+  had already filled. They are now `errors.Is`-matchable as
+  `ErrDependencySourceNotFound` and `ErrDependencyTargetNotFound`, carried by a
+  `*DependencyEndpointNotFoundError` naming the refused edge and which of its
+  endpoints was absent, all three re-exported from the root package beside the
+  conflict types. What is NOT refused is unchanged: an `external:` reference
+  and another repository's id are still accepted as external targets.
+
+  The user-facing message is byte-identical on the embedded and store paths
+  (`issue <id> not found`), and the proxied-server path now renders that
+  message too instead of a raw foreign-key violation — the endpoint is
+  classified from the refusing transaction rather than parsed out of the
+  driver's prose, so the two plumbings agree on the text as well as the type.
+  The shared `DependencyEditor` contract asserts the sentinel and the typed
+  fields for each endpoint, alone in a request and mid-batch, with the
+  mid-batch half reading the graph back at zero edges. `RemoveDependency` is
+  unaffected: a removal that finds no edge is a success, not a refusal.
+
 ### Changed
 
 - **`bd search` now includes closed issues by default** (bd-t5yex). The
--- a/backend/conformance/dependency_editor_contract.go
+++ b/backend/conformance/dependency_editor_contract.go
@@ -886,10 +886,13 @@ func RunDependencyEditorRoutesWispSourcedRemovalToTheWispPlane(t *testing.T, ctx
 // the target here is deliberately a real seeded issue: the refusal under test
 // is about the source alone.
 //
-// SPEC-GAP bd-yby99.9: the doc now states that the refusal HAPPENS and states
-// that it deliberately names no identity yet, so err != nil is still the whole
-// assertion — pinning a sentinel or a message here would assert more than the
-// leaf says, and the leaf says the anonymity is a gap rather than a promise.
+// The refusal's IDENTITY is asserted in BOTH POSITIONS. Alone in a request it
+// is the only thing that can have failed; mid-batch it competes with the
+// rollback of the edge already written, which is where an implementation that
+// re-raised the refusal as its own wrapper — or as the rollback's error — would
+// lose the type. The mid-batch half therefore reads the graph back at zero
+// edges as well, because a typed refusal that left half a graph behind would
+// still be the wrong answer.
 func RunDependencyEditorRefusesAGhostSource(t *testing.T, ctx context.Context, fixture DependencyEditorFixture) {
 	t.Helper()
 	source := fixture.IssuePrefix + "-ghostsrc-known"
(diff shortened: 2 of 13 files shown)


julianknutsen:
## Verdict: Approve

This is a maintainer-grade change: it closes a stated spec gap (`bd-yby99.9`) by giving the two dependency-endpoint refusals the same typed identity every other `DependencyEditor` refusal already has, and it does so without changing a single user-visible message on the store paths. I adversarially spot-checked the load-bearing claims at the PR head (`927f943`) rather than trusting the description; all of them held.

### What I verified in code

**The vocabulary and its re-export chain.** `ErrDependencySourceNotFound`, `ErrDependencyTargetNotFound`, and `*DependencyEndpointNotFoundError{IssueID, DependsOnID, MissingID, Err}` live in `issueops/errors.go` directly beside `ClaimConflictError`, whose arrangement they follow: the carrier wraps the sentinel via `Unwrap`, so `errors.Is` matches and `errors.As` yields the fields. The chain `issueops` → `internal/storage/domain` → root `beads` package uses the exact alias pattern of the existing conflict types, and the identity is pinned three ways: the sentinel-identity table in `errors_test.go`, an `errors.As` extraction through a real use-case error, and the `reflect.TypeFor` canonical-type check beside the conflict types' own.

**Message stability.** The claim that the store-path message is byte-identical is true: the old `fmt.Errorf("issue %s not found", id)` sites now return a value whose `Error()` is `fmt.Sprintf("issue %s not found", e.MissingID)` with `MissingID` bound to the same id. The target refusal in `addDependencyInTx` remains inside the pre-existing external/cross-prefix gate, so what is *accepted* genuinely does not change.

**The unit-of-work classification, which is the risky part.** In `internal/storage/domain/db/dependency.go`, `classifyMissingEndpoint` is invoked only from the insert's error branch — the happy path (including bulk adds) pays no per-edge probe. It gates on the FK signal (`IsMissingForeignKeyTarget`, MySQL 1452/1216 only), probes both endpoints on the same runner that refused, and returns nil on any probe failure or unsettled state, so the caller falls through to the original wrapped insert error. A refusal is never downgraded to a probe's failure, exactly as claimed — and that fallback is also the correct degradation for any future engine that poisons a transaction after a failed statement.

**The proofs.** The two contract cases in `backend/conformance/dependency_editor_contract.go` are strengthened in place, not duplicated: each asserts the sentinel *and* all three typed fields, alone in a request and mid-batch, and the mid-batch half reads the graph back at zero edges so a typed refusal cannot mask a broken rollback. All three implementations (direct store, embedded store, unit-of-work) run them, and the PR documents the red state observed on each before the production change — including the raw `Error 1452` the uow path used to leak.

**Create-path compatibility.** `ClassifyPublicCreateError` matches the new type in the same arm as the raw FK signal and feeds the identical `ErrValidation`-wrapping-`ErrNotFound` classification, so create-path consumers see no change. `RemoveDependency` is correctly left alone: its contract already states that a no-edge removal is `Removed` false with a nil error — a success with no refusal to name.

**Hygiene.** Additions only on the public surface (the one signature change, `readDepTargetForPrecheck`, is internal); CHANGELOG entry under Unreleased; no out-of-tree references; CI fully green including the historical-upgrade and differential-regression matrices.

### Non-blocking observations

- The typed identity is an in-process contract. Nothing marshals the sentinel across the server wire, so remote callers see only the (now converged) message. The PR is honest about this scope; typed errors over the wire would be a reasonable follow-up bead.
- One deliberate deviation from the `ClaimConflictError` template: there the sentinel's prose *is* the rendered message; here the carrier overrides `Error()` to preserve the historical text, so the sentinel strings never surface to users. The type's doc says so, and it is the right trade — just worth knowing when grepping logs.
- When both endpoints are missing, the source wins (checked first), matching the store body's order. Consistent across backends, though untested as a pinned case.

PR #5416, opened 2026-08-07
feat(httpapi): serve closeIssue over v0

julianknutsen:
## Stack position

**PR 1 of 3** in the v0 write-lifecycle series. Base: `getissue-include-params` (#5410).

```
getissue-include-params (#5410)
└── lifecycle-close        ← this PR
    └── lifecycle-reopen
        └── lifecycle-update
```

The design spec's delivery order puts `closeIssue` first among the writes, for two reasons. It is the **highest-value** one — claim → work → close is the agent loop this surface exists to serve, and close is the half that still forces a client back to forking a `bd` subprocess. And it is the **hardest**: it introduces the suffix dispatcher and the `not_closable` code, so it is taken while the tranche is freshest and the two PRs above it inherit finished machinery rather than inventing it.

It inherits from #5410 the JSON media-type document rule and the widened `actor` pattern, so nothing here re-argues either.

## What lands

`POST /v0/beads/issues/{id}:close` — spec, `routeTable` row, `operationCodes` row, handler and tests together, under the existing parity gates.

### The routing change is the substantive part

`{id}:close` is not expressible as a ServeMux pattern for the same reason `{id}:claim` is not — wildcards match a whole segment. Unlike the claim, it **collides**: two rows cannot each register `POST /v0/beads/issues/{idop}`.

The route table grows a `customMethod` field. Rows carrying one share a single registration behind a dispatcher that splits the trailing `:suffix` off the matched segment and hands the request to the row that claims it. A segment ending in no registered suffix gets the same 404 the catch-all gives any unrouted path, so the wide pattern stays a routing detail rather than undocumented surface.

The split happens **before** `s.route`, so an unrouted suffix takes no database slot and books no operation on the request line — where the claim, while it was the only POST here, answered that case from inside its own handler and attributed every probe of the prefix to itself.

`TestSpecRouteParity` is untouched in mechanism: each row still declares its own `specPath`, and the test still compares document paths without learning the exception's mechanics. `TestClaimNarrowsThePOSTSurface` generalizes to `TestCustomMethodsNarrowThePOSTSurface` — the test where a dispatcher that fell back to its first row would be caught, since that failure would otherwise leave the claim's happy path green. A new `TestCustomMethodRowsDeclareTheirDocumentedPath` pins the invariant parity cannot see: a row whose `specPath` ends in one verb while its `customMethod` matches another would route the documented path to the wrong operation with every gate passing.

### Vocabulary

`not_closable` (409) joins the frozen code set, with `open_children` as an optional `Problem` extension member. Both close-policy refusals — open children, and a live blocker — share the code, because they are the same statement to a client and `force` is the bypass for both. The member is attached only for the open-children one, from `*CloseOpenChildrenError`'s own field inside the refusing transaction, so **member presence is the discriminator** and no client parses `3 open child issue(s)` out of prose.

409 rather than the delete precedent's 400: this is a statement about the current state of one named resource, so the same request succeeds or fails on state the client cannot see without reading it. That is the `not_claimable` situation and it gets the `not_claimable` answer.

`already_closed` mirrors `already_claimed` for the idempotent re-close, which under first-close-wins rewrites neither `reason` nor `session`.

### One security-relevant fix

`RoleFiresHooks` learns about `*hookIssueOperations`. A store's own `IssueLifecycle()` returns a role that fires `on_create`, `on_update` and the close hooks for every mutation it lands — precisely the value this server documents it does not run. Without that case `bd serve` would have booted with hooks live on four verbs, and the refusal that exists to prevent exactly this would have passed it.

## Gates

- `go test ./internal/storage/...` — green
- Red/green observed per behavior. Two mutations were applied to confirm the new tests bite: a dispatcher falling back to `rows[0]` on an unknown suffix fails all ten `TestCustomMethodsNarrowThePOSTSurface` cases, and dropping the `WithOpenChildren` call fails `TestCloseOpenChildrenIsATypedConflict`.

`TestProxiedServerServeClose` (`cmd/bd`, against real Dolt through a real `bd serve` subprocess):

- **first-close-wins**, read back through a second close: the replay reports `already_closed: true` and the stored `close_reason`/`closed_by_session` still carry the **first** close's values
- **a forced close over two real open children**: the unforced attempt is a 409 carrying `open_children: 2` and writes nothing; `force: true` lands and still reports the 2 it bypassed
- the rule-8 parity oracle against `bd close --json`[0], empty allowlist
- close/claim agreement on the row they share, and refused bodies writing nothing

The role-level transition is already owned against a real store by `internal/storage/uow/lifecycle_close_reopen_contract_test.go`, which this cites rather than duplicates.

## Spec deviations

None. `ExpectedVersion` stays unpublished per the spec's open questions.

### DIFF
--- a/cmd/bd/serve.go
+++ b/cmd/bd/serve.go
@@ -180,6 +180,7 @@ func runServe() error {
 			AllowNonLoopback:  serveAllowNonLoopback,
 			Reader:            roles.reader,
 			Claimer:           roles.claimer,
+			Lifecycle:         roles.lifecycle,
 			Settings:          roles.settings,
 			Stats:             roles.stats,
 			CycleDetector:     roles.cycles,
@@ -434,6 +435,7 @@ func serveIssueRoles(src storage.DoltStorage) (serveRoles, error) {
 	for _, b := range []binding{
 		{"issue reader", func() (err error) { roles.reader, err = src.IssueReader(); return }},
 		{"issue claimer", func() (err error) { roles.claimer, err = src.IssueClaimer(); return }},
+		{"issue lifecycle", func() (err error) { roles.lifecycle, err = src.IssueLifecycle(); return }},
 		{"workspace config", func() (err error) { roles.settings, err = src.WorkspaceConfig(); return }},
 		{"stats reporter", func() (err error) { roles.stats, err = src.StatsReporter(); return }},
 		{"cycle detector", func() (err error) { roles.cycles, err = src.CycleDetector(); return }},
@@ -461,6 +463,7 @@ func serveIssueRoles(src storage.DoltStorage) (serveRoles, error) {
 type serveRoles struct {
 	reader       issueops.Reader
 	claimer      issueops.Claimer
+	lifecycle    issueops.Lifecycle
 	settings     issueops.WorkspaceConfig
 	stats        issueops.StatsReporter
 	cycles       issueops.CycleDetector
--- a/cmd/bd/serve_close_proxied_integration_test.go
+++ b/cmd/bd/serve_close_proxied_integration_test.go
@@ -1,0 +1,262 @@
+//go:build cgo
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+// End-to-end for the close, against real Dolt through a real `bd serve`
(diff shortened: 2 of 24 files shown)


julianknutsen:
Auto-closed by base-branch deletion after #5410 merged; superseded by #5423 (same head commit 3c455c27d, base main). The Fable review is posted there.

PR #5417, opened 2026-08-07
feat(httpapi): serve reopenIssue over v0

julianknutsen:
## Stack position

**PR 2 of 3** in the v0 write-lifecycle series. Base: `lifecycle-close`.

```
main
└── lifecycle-close (#5423)
    └── lifecycle-reopen (#5417)   ← this PR
        └── lifecycle-update (#5422)
```

#5410 has merged, so the stack now sits on `main`; the close PR was recreated as #5423 when GitHub closed #5416 along with its deleted base branch.

`reopenIssue` goes third overall and second among the writes, because it is **near-trivial on the dispatcher the close just built** — a third row on the shared wildcard is a row — and because it **completes the lifecycle pair** the document describes it as ("it is `POST /v0/beads/issues/{id}:close`'s mirror, and it completes the lifecycle pair so a recovery flow works end to end over this surface"). Landing it immediately after close is what keeps the dispatcher's second consumer honest while the design is still fresh: `TestCloseAndReopenAreDistinctOnOnePattern` is a test that could not exist before this PR.

> Review only the top commit; the parent is #5423.

## What lands

`POST /v0/beads/issues/{id}:reopen` — spec, `routeTable` row, `operationCodes` row, handler and tests together. No new routing machinery and no new codes.

### The absence of a 409 is the design statement

Close has a policy guard — open children, a live blocker — because it moves an issue **into** the done category. Reopen moves it **out**, so there is no state of the graph that can refuse it and nothing to name.

`TestReopenPublishesNoConflictCode` asserts that as an invariant rather than leaving it as the incidental current shape of a table, so a future change that gives reopen a guard has to land its code deliberately rather than discovering it needs one.

### Idempotency, and where the reason lives

`already_open` is `!result.Changed`, mirroring `already_closed` and `already_claimed`: a reopen of a non-done issue changes nothing and succeeds, because an agent replaying its own recovery should not have to classify an error to learn it already ran.

The `reason` is recorded on the `reopened` **event** rather than on a field of the issue — the role's own contract — and the response deliberately does not carry it, so a client reads it back from the place it actually lives. A test asserts the response does *not* publish it, because publishing it would tell clients to read it from the wrong place.

The body refuses the close's `session` and `force` **by name**. They are the neighbouring operation's members, and a client that sends them here is told so rather than having them silently ignored.

### Provenance

The handler names the history entry (`ReopenRequest.Provenance`) instead of taking the role's default. The implementations disagree about that default — the store-backed ones write `bd: reopen issue`, the unit-of-work one writes `reopen issue` — so a workspace served by one backend today and another tomorrow would grow two spellings of one event. Not wire-visible.

## Gates

- `TestProxiedServerServeLifecycle` and `TestServerModeServe` — green (`BEADS_TEST_PROXIED_SERVER=1`). This commit adds `issues.reopen` to the single `servedCapabilities` list #5423 introduces, so the commit stays self-contained.
- Red/green observed. The initial run of this branch failed `TestCustomMethodsNarrowThePOSTSurface//v0/beads/issues/bd-1:reopen` — that path was a *negative* case in PR 1 ("a verb this build does not serve") and is now served, which is the surface-widening this PR performs, caught by the test written to catch it. Two mutations confirm the new tests bite: blanking the provenance fails `TestReopenNamesItsHistoryEntry`, and setting the reopen row's `customMethod` to `:close` fails the whole reopen suite plus `TestCustomMethodRowsDeclareTheirDocumentedPath`.

`TestProxiedServerServeClose` grows the round trip the close operation's own description promises — "the stored pair keeps what the first close gave it until a reopen clears both" — against real Dolt:

- **close → reopen**: `close_reason` and `closed_by_session` are **cleared** — which is what makes the close's first-close-wins rule survivable, since the way to write a new reason is to reopen and close again
- the reopen **reason read back off the issue's event stream** via `bd history --events`, proving the document's claim about where it lands
- the issue is **claimable again**, which is the recovery flow the pair exists to serve
- the idempotent reopen of an issue that was never done

`ExpectedVersion` stays unpublished here for the reason #5423 gives: `issueops.ErrVersionMismatch` has no entry in `problem.go`'s classification, so an `expected_version` member would render its own refusal as `500 internal`, and the status+code pair it needs is a one-way door (`bd-serve-v0.md`, "The error-code vocabulary") that `TestSpecStatusCodesMatchHandlerTable` forces to be admitted deliberately. Reopen publishing nothing new here is the same restraint `TestReopenPublishesNoConflictCode` pins above.

### DIFF
--- a/cmd/bd/serve_close_proxied_integration_test.go
+++ b/cmd/bd/serve_close_proxied_integration_test.go
@@ -48,6 +48,32 @@ func (sp *serveProcess) closeIssue(t *testing.T, id, body string) (int, map[stri
 	return resp.StatusCode, m
 }
 
+// reopenIssue posts a reopen for id and returns the status and decoded body.
+func (sp *serveProcess) reopenIssue(t *testing.T, id, body string) (int, map[string]any) {
+	t.Helper()
+	req, err := http.NewRequest(http.MethodPost, sp.url("/v0/beads/issues/"+id+":reopen"), strings.NewReader(body))
+	if err != nil {
+		t.Fatalf("new reopen request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := sp.client.Do(req)
+	if err != nil {
+		t.Fatalf("POST reopen %s: %v\nstderr:\n%s", id, err, sp.stderr.String())
+	}
+	defer func() { _ = resp.Body.Close() }()
+	raw, err := io.ReadAll(resp.Body)
+	if err != nil {
+		t.Fatalf("read reopen body: %v", err)
+	}
+	var m map[string]any
+	if len(raw) > 0 {
+		if err := json.Unmarshal(raw, &m); err != nil {
+			t.Fatalf("decode reopen body %q: %v", raw, err)
+		}
+	}
+	return resp.StatusCode, m
+}
+
 func TestProxiedServerServeClose(t *testing.T) {
 	requireSharedProxiedServer(t)
 	t.Parallel()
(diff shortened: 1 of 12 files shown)


julianknutsen:
## Review: approve

`reopenIssue` lands as the third row on the custom-method dispatcher, and the PR is exactly as narrow as it says it is: one commit against `lifecycle-close`, no new routing machinery, no new problem codes, spec + route row + handler + tests together.

### What I verified

- **Diff scope.** The PR's diff equals `git diff` of its head against the parent branch head — 11 files, all reopen-scoped. No stack bleed.
- **Tests at head, independently re-run.** The full `internal/httpapi` package passes at `e2eed074`, `go vet` and `gofmt` are clean, and regenerating `apigen/types.gen.go` from the spec produces no drift, so the `api-check` gate holds.
- **The claimed mutations bite.** Blanking `reopenProvenance` fails `TestReopenNamesItsHistoryEntry`; mis-setting the reopen row's `customMethod` to `:close` fails the whole reopen suite plus `TestCustomMethodRowsDeclareTheirDocumentedPath`. Both reproduced locally.
- **Dispatcher stays narrow.** `customMethodTarget` has no fallback row — a segment ending in no registered suffix gets the catch-all 404. `TestCustomMethodsNarrowThePOSTSurface` now runs 12 cases, and this PR correctly moved `:reopen` from the negative list to served surface while adding `:Reopen`, `:reopened`, and the bare `/:reopen` as new negatives.
- **The provenance rationale is real.** The store-backed lifecycle implementations default the history entry to `bd: reopen issue`, the unit-of-work one to `reopen issue`; the handler pinning `bd serve: reopen issue` genuinely removes a backend-dependent spelling, and the test would catch losing it.
- **The round trip is backed by the store.** `ReopenIssueInTx` clears `closed_at`, `close_reason`, `closed_by_session`, and `defer_until`, records the `reopened` event with actor and reason, and its `Changed` result is what `already_open = !result.Changed` reports — so the integration proof (close → reopen clears the close record, reason read back via `bd history --events`, issue claimable again) asserts what the code actually does.
- **Hooks stay off.** The reopen path is served through `Config.Lifecycle`, which the `Listen`-time guard checks via `RoleFiresHooks`; that predicate recognizes the hook-firing lifecycle wrapper, and both the refusal and the wrapper recognition are pinned by existing tests. The spec's hooks/auto-commit disclaimer appears on this operation as on every other write.
- **Frozen codes.** `operationCodes[reopenIssue]` is `{invalid_argument, not_found, busy, db_unavailable, internal}` with no 409, `TestReopenPublishesNoConflictCode` pins the absence, and the spec/table parity tests enforce agreement in both directions.

### CI note

Only the migration harness ran on this PR: the main PR workflow triggers on PRs based on `main`, so it cannot fire while this PR is stacked on `lifecycle-close`. That is stack plumbing, not a red — but the full suite and conformance tier should be confirmed green after the stack retargets to `main` and before merge.

### Non-blocking

1. The reopen integration subtests live inside `TestProxiedServerServeClose` in `serve_close_proxied_integration_test.go` — a close-named test now owns the reopen round trip. The fixture reuse is sound; consider renaming to something lifecycle-scoped (or leaving a pointer comment) when the next PR touches this file.
2. `already_open: true` covers any non-done status, not only literal `open`. The schema description states this precisely, so no action needed — just flagging that the member name reads slightly narrower than the contract, consistent with the `already_claimed` precedent.
3. A reopen with a reason also records a comment event alongside the `reopened` event — pre-existing role behavior shared with the CLI, and the documented claim about where to read the reason back remains true. If clients ever ask why a reopen grows a comment, the operation description is the place to say so.

None of these block. The operation matches its documentation exactly, the idempotent flag mirrors its two precedents, and the deliberate absence of a conflict code is pinned as an invariant rather than left as table shape.

PR #5421, opened 2026-08-07
fix(note): reject subcommand-style words as the issue id (GH#5370)

### LINKED ISSUE #5370: bd note <id> [text...] has the same unguarded id-as-subcommand-typo hole "bd comment" had (fixed in #5369)
## What

`bd note <id> [text...]` (cmd/bd/note.go) has the exact same shape and exposure that `bd comment <id> [text...]` had before #5369: `Use: "note <id> [text...]"`, `id := args[0]` straight into the same resolve-and-mutate path, with no validation on the id. There is no plural `notes` command, so — unlike `comment`/`comments` — there was never a sibling command that needed (or got) a guard.

## Why this matters

Same failure mode #5369 fixes for `comment`: if a caller passes a word that happens to collide with an existing bead/wisp id (via `ResolvePartialID`'s substring/prefix fallback, including the wisp-fallback's `wisp-` prefix stripping), the note silently lands on the wrong issue instead of erroring. `note` has no sibling command to confuse the syntax with the way `comment`/`comments` do, so the specific "typed `list`/`add`" trigger from #5369 doesn't apply verbatim — but the underlying missing-id-validation gap is identical, and any other caller mistake that produces a non-existent-but-coincidentally-matching id string would hit it the same way.

## Suggested fix

Not proposing a specific implementation here (deliberately out of scope for this issue, per CONTRIBUTING.md's one-issue-per-PR guidance — #5369 already covers `comment`). Whoever picks this up should decide whether it's worth an `Args` validator here too, and if so, what it should actually check given `note` has no sibling command whose vocabulary to borrow from (unlike `comment`'s "list"/"add" collision with its plural sibling).

## Found via

Noticed while implementing #5369 for `bd comment`'s identical bug; filed separately rather than piggybacked on that PR.

vishnujayvel:
## What

`bd note <id> [text...]` now rejects subcommand-shaped words as the id positional argument (`list`, `add`, `show`, `edit`, `rm`, `remove`, `delete`, `update`) instead of silently accepting them.

## Why

Same hole #5369 fixed for `bd comment`: `note` takes `args[0]` as a bead id with no reserved-word guard, so a subcommand-shaped typo can resolve via `ResolvePartialID`'s substring/prefix fallback and append a note to the wrong issue — exit 0, no warning.

`note` has no plural sibling (unlike `comment`/`comments`), so there is no real list/add/show subcommand to confuse with; the Args-validator pattern still closes the fuzzy-id mutation path for those common bare words (real ids always carry a prefix+hyphen).

This denylist is **broader than #5369's two words** (`list`/`add`):
- `show` is invited by the guard's own error text (`bd show <issue-id>`), so `bd note show <id>` is a predictable typo that previously bypassed the guard.
- `update` is invited by the guard's own hints referencing `bd update <issue-id>` four times, so `bd note update <id> ...` is an invited typo that previously bypassed the guard into `ResolvePartialID` fuzzy matching.
- `edit` / `rm` / `remove` / `delete` are common subcommand-shaped typos for mutating commands.
- Broader-than-#5369 matches the in-repo direction of open PR #5393 (athosmartins), which broadens the sibling comment guard similarly.

Mirrors `validateCommentArgs` from #5369 rather than inventing a different mechanism.

Fixes #5370  
Reported by @athosmartins (also the author of the #5369 comment guard and PR #5393).

## Verification

- `go test -tags gms_pure_go -run TestValidateNoteArgs -v ./cmd/bd` — all subtests pass (pure unit; no store/Docker), covering every denylist word plus reserved-word-in-TEXT still works.
- `go test -tags "gms_pure_go cgo integration" -run 'TestCLI_Note(List|Add)MisplacedSyntax' -v ./cmd/bd` — both pass in isolation **and** in a full `-run 'TestCLI_Note'` run.
- Misplaced-syntax tests assert bystander notes are **unchanged** (before==after), not merely empty — immune to package-level cobra flag leaks from earlier tests.
- `TestCLI_NoteTextStartingWithReservedWordStillWorks` remains green (reserved word in TEXT position still works).
- `gofmt -l` on changed files — clean.
- Cross-ref implementation: same Args-validator shape as #5369 (`validateCommentArgs` → `validateNoteArgs`); broader denylist aligned with #5393's direction.
- PR #5369 has no dedicated JSON-mode error-path test for the comment guard (errors are asserted via stdout/stderr string match; `--json` is only used for create/list helpers), so no mirror JSON-error test was added for note.

## Notes for reviewers

- Scoped strictly to the id slot (`args[0]`); note text that starts with a reserved word still works.
- Help Long text updated generically: `"note" has NO subcommands — it only appends` (rather than enumerating subcommand words, which omitted remove/delete and would drift as the denylist grows); read notes via `bd show <id>`.
- Pre-existing: `TestCLI_NoteCommand` is red under these tags at base `1174876c8` due to a package-level flag leak. The leak comes from an earlier invocation **inside the same test** (the `create` at `cli_fast_test.go:514` with `--notes "Original notes"`), not from a different test — which is why it also fails in isolation. Out of scope for this PR.

### DIFF
--- a/cmd/bd/cli_fast_test.go
+++ b/cmd/bd/cli_fast_test.go
@@ -558,6 +558,203 @@ func TestCLI_NoteCommand(t *testing.T) {
 	}
 }
 
+// TestCLI_NoteListMisplacedSyntax is a regression test for "note" silently
+// mis-resolving a subcommand-like typo as an issue id (GH#5370). Mirrors
+// TestCLI_CommentListMisplacedSyntax (#5369): "bd note list <text>" used to
+// parse as id="list", text=[<text>...] with no validation on the id — since
+// no issue is literally named "list", ResolvePartialID's substring/prefix
+// fallback would silently match it against whatever existing bead or wisp id
+// happened to contain "list" and append the note there instead of erroring.
+// This asserts both halves of the fix: the command is rejected with a
+// helpful hint, AND no note silently landed on an unrelated real issue.
+func TestCLI_NoteListMisplacedSyntax(t *testing.T) {
+	tmpDir := setupCLITestDB(t)
+
+	out := runBDInProcess(t, tmpDir, "create", "Bystander issue", "-p", "1", "--json")
+	jsonStart := strings.Index(out, "{")
+	if jsonStart < 0 {
+		t.Fatalf("No JSON found in create output: %s", out)
+	}
+	var issue map[string]interface{}
+	if err := json.Unmarshal([]byte(out[jsonStart:]), &issue); err != nil {
+		t.Fatalf("Failed to parse create JSON: %v\nOutput: %s", err, out)
+	}
+	fullID := issue["id"].(string)
+
+	// Snapshot notes BEFORE the rejected command. runBDInProcess reuses the
+	// package-level rootCmd without resetting cobra flag state, so an earlier
+	// test's --notes value can leak into this create. Asserting before==after
+	// is immune to that leak and is a stronger regression check: the rejected
+	// command wrote nothing (GH#5370 review fix-round).
+	beforeShow := runBDInProcess(t, tmpDir, "show", fullID, "--json")
+	var beforeShown []map[string]interface{}
+	if err := json.Unmarshal([]byte(beforeShow), &beforeShown); err != nil {
+		t.Fatalf("Failed to parse pre-reject show JSON: %v\nOutput: %s", err, beforeShow)
+	}
+	if len(beforeShown) == 0 {
+		t.Fatalf("expected show to return the bystander issue before reject, got: %s", beforeShow)
+	}
+	notesBefore, _ := beforeShown[0]["notes"].(string)
+
+	stdout, stderr, err := runBDInProcessAllowError(t, tmpDir, "note", "list", "accidental stray text")
+	if err == nil {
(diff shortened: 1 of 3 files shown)


steveyegge:
## Review verdict: MERGE

Adversarially reviewed against GH#5370. The guard lands where it must — before `ResolvePartialID`'s SQL `LIKE` substring fallback and before store open (your pre-store-open test proves it), covering both backends by construction. Over-blocking is impossible: `validation.ValidateIDFormat` requires a hyphen, so no legitimate full id can equal a denied bare word, custom prefixes included. Test quality is strong — the bystander-unchanged assertions and the reserved-word-in-text positive case are exactly the right controls.

Three residuals, none blocking, for the record:
- **Case variants still slip through** (`bd note Add …` passes the exact-match switch and `LIKE` is case-insensitive under Dolt's default collation). Shared with the merged #5369 precedent, so accepted here — but a `strings.ToLower` would be free if you touch this again.
- **comment/note denylists diverge until #5393 lands** (comment blocks only list/add). Disclosed in your body; noted for sequencing.
- The repo now has two ~90-line near-identical validators with a third variant coming — a shared `rejectReservedIDWord` table is the eventual refactor.

Full CI green. Merging.

PR #5422, opened 2026-08-07
feat(httpapi): serve updateIssue over v0

julianknutsen:
## Stack position

**PR 3 of 3** in the v0 write-lifecycle series. Base: `lifecycle-reopen`.

```
main
└── lifecycle-close (#5423)
    └── lifecycle-reopen (#5417)
        └── lifecycle-update (#5422)   ← this PR
```

#5410 has merged, so the stack now sits on `main`; the close PR was recreated as #5423 when GitHub closed #5416 along with its deleted base branch.

`updateIssue` goes last of the three because it has **the widest body vocabulary but no new routing and no new codes** — so it lands once the patch-decode pattern has the close and reopen bodies as precedent. `bodyActor` and `storedTextMember`, factored out in the two PRs below, are what this one builds on rather than inventing.

> Review only the top commit; the parents are #5423 and #5417.

## What lands

`PATCH /v0/beads/issues/{id}` — spec, `routeTable` row, `operationCodes` row, handler and tests together.

### A plain method

Not a custom method, by the memories precedent: partial update of one named resource is what `PATCH` already means, exactly as one named resource with no body is what `DELETE` means for `forgetMemory`. It registers directly, keeps `pattern == specPath`, and leaves the custom-method namespace for the operations that are not CRUD. The claim-route commentary calling the issue-detail path "GET-only" is updated in the same change.

### Member presence is the signal

This is the reason the body is decoded as raw members rather than into the generated struct: a member **present** sets the role's `Field`, a member **absent** leaves the field untouched, and the generated struct cannot tell those apart because it models both as a nil pointer.

Explicit `null` is a third state, read off the raw bytes before any typed decode:

- on the four **nullable** members — exactly the ones the role models as `Field[*T]` — it **clears**, since `null` is the only wire spelling a clear has
- on every other member it is a **400 naming the member**

Without that pre-check, `{"title": null}` unmarshals to a nil pointer and slides through as "untouched" — a write the client asked for and the server silently dropped.

### The vocabulary boundary

`status`, `assignee` and `metadata` are excluded and refused **by name**. Keeping them out is half the reason this operation documents no `409` at all: they are exactly where close policy, the claim fence and the metadata algebra would have entered its code set. `parent_id`, `persistence`, `owner`, `closed_by_session`, `spec_id` and `await_id` are absent for the reasons this operation's own description in `openapi.v0.yaml` enumerates — a set `parent_id` atomically replaces ALL parents, so it is a graph edit in patch clothing; `persistence` is an atomic aggregate migration between planes; `owner` is `assignee`'s argument; `closed_by_session` is written under first-close-wins by `{id}:close`, which a patch write would bypass; and `spec_id`/`await_id` are workflow plumbing this surface publishes nowhere yet. Each is additive to publish later.

`labels` is complete replacement, and an empty array is a clear rather than an empty patch — a distinction the tests pin on both sides.

### Two new spec gates

`TestUpdateRequestMembersMatchTheHandler` covers **both levels** of the body. The handler's accepted set is a hand-rolled copy of the widest member list on the surface, and it could otherwise drift from the document with `make api-check` and every other spec test green.

`TestNullablePatchMembersAreExactlyTheDocumentsNullableOnes` pins the clear-vs-refuse set from both sides, because drift there is silent in the worst direction: a member the handler treats as nullable but the document does not is an unannounced clear on a field nobody agreed could be cleared.

### One routing interaction, pinned rather than papered over

`PATCH /v0/beads/issues/{id}` is a single-segment wildcard, so it matches `/v0/beads/issues/bd-1:close` too. That is the detail path's long-standing behavior — `GET` has always answered the same way — and it resolves as an update of an issue whose id is literally `bd-1:close`, which no row holds, so it 404s. `TestPatchOnACustomMethodSegmentIsNeverTheCustomMethod` pins the property that actually matters: **a PATCH never executes as a lifecycle verb**. The proxied test asserts the same thing against a real row.

This surfaced as a red test in PR 1's `TestCloseIsNotReachableByOtherMethods`, which listed PATCH among the unrouted methods; that entry is now correctly excluded, with the reason recorded next to it.

## Gates

- `TestProxiedServerServeLifecycle` and `TestServerModeServe` — green (`BEADS_TEST_PROXIED_SERVER=1`). This commit adds `issues.update` to the single `servedCapabilities` list #5423 introduces, so the commit stays self-contained.
- Red/green observed. Two mutations confirm the presence machinery is load-bearing: forcing `set("description")` true fails `TestUpdateLeavesAbsentMembersUntouched`, and disabling the explicit-null pre-check fails `TestUpdateClearsTheNullableMembers` and all nine cases of `TestUpdateRefusesNullOnEveryOtherMember`.

`TestProxiedServerServeUpdate` (`cmd/bd`, against real Dolt), the three-way read-back the document's "member presence is the signal" rule demands:

- one issue carrying a **set** member (`title`, `priority`), a **null-cleared** member (`external_ref`), and **untouched** members (`description`, `acceptance_criteria`, `status`) — all read back after the PATCH
- `changed: false` on a **same-value resend**
- label replacement removing a label the request omitted
- the workspace-vocabulary refusal arriving as the documented **400** rather than falling through to a 500
- the parity oracle against `bd update --json`[0], empty allowlist
- a PATCH on a custom-method segment leaving the issue open

`Claim`, the force flags and every `Expected*` precondition stay zero, asserted by a test rather than left implicit (`TestUpdateForwardsEveryDocumentedMember`). Each is unpublished for a reason the document already states somewhere: `Claim` and the assignee force are the claim's compare-and-set territory, `ForceClosePolicy` belongs to `{id}:close`, and the `Expected*` preconditions have no frozen refusal code to land on — `issueops.ErrVersionMismatch` is unmapped in `problem.go`, so publishing one would render its refusal as `500 internal`. Adding that status+code pair is the one-way door `bd-serve-v0.md` describes under "The error-code vocabulary", which is exactly the kind of decision this PR declines to make in passing.

### DIFF
--- a/cmd/bd/serve_proxied_integration_test.go
+++ b/cmd/bd/serve_proxied_integration_test.go
@@ -147,6 +147,7 @@ var servedCapabilities = []any{
 	"dependencies.blocking", "dependencies.cycles", "dependencies.list", "dependencies.tree",
 	"issues.batchCreate", "issues.claim", "issues.close", "issues.delete",
 	"issues.get", "issues.list", "issues.query", "issues.reopen", "issues.sweep",
+	"issues.update",
 	"memories.forget", "memories.get", "memories.list", "memories.remember",
 	"ready.count", "ready.list", "stats.get",
 }
--- a/cmd/bd/serve_update_proxied_integration_test.go
+++ b/cmd/bd/serve_update_proxied_integration_test.go
@@ -1,0 +1,250 @@
+//go:build cgo
+
+package main
+
+import (
+	"encoding/json"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+// End-to-end for the field update, against real Dolt through a real `bd serve`
+// subprocess. The pure tests in internal/httpapi cover the wire edge on a fake
+// role and assert the projection onto issueops.IssuePatch field by field; what
+// only this level can prove is that the projection LANDS — that a set member, a
+// null-cleared member and an untouched member all end up in the row the way the
+// document says, and that a same-value resend really is a no-op against a store
+// rather than against a fake's bookkeeping.
+
+// updateIssue patches id and returns the status and decoded body.
+func (sp *serveProcess) updateIssue(t *testing.T, id, body string) (int, map[string]any) {
+	t.Helper()
+	req, err := http.NewRequest(http.MethodPatch, sp.url("/v0/beads/issues/"+id), strings.NewReader(body))
+	if err != nil {
+		t.Fatalf("new update request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := sp.client.Do(req)
+	if err != nil {
+		t.Fatalf("PATCH %s: %v\nstderr:\n%s", id, err, sp.stderr.String())
+	}
+	defer func() { _ = resp.Body.Close() }()
+	raw, err := io.ReadAll(resp.Body)
+	if err != nil {
+		t.Fatalf("read update body: %v", err)
+	}
+	var m map[string]any
+	if len(raw) > 0 {
+		if err := json.Unmarshal(raw, &m); err != nil {
+			t.Fatalf("decode update body %q: %v", raw, err)
+		}
+	}
+	return resp.StatusCode, m
+}
+
+func TestProxiedServerServeUpdate(t *testing.T) {
+	requireSharedProxiedServer(t)
+	t.Parallel()
+	bd := buildEmbeddedBD(t)
+	p := newSharedProxiedProject(t, bd, "srvupd")
+	sp := startServe(t, bd, p.dir, bdProxiedEnv(p.dir))
+
+	// THE THREE-WAY PROOF the spec names: one issue carrying a SET member, a
+	// NULL-CLEARED member and an UNTOUCHED member, read back after the PATCH.
+	// A fake role can report what it was handed; only a real row can show that
+	// "absent means untouched" is a property of the write rather than of the
+	// handler's bookkeeping.
+	t.Run("a set member, a cleared member and an untouched member", func(t *testing.T) {
+		issue := bdProxiedCreate(t, bd, p.dir, "before", "-p", "3",
+			"-d", "the original description", "--acceptance", "the original criteria")
+
+		// Give it something to clear and something to leave alone.
+		if _, err := bdProxiedRun(t, bd, p.dir, "update", issue.ID, "--external-ref", "JIRA-1"); err != nil {
+			t.Fatalf("bd update --external-ref: %v", err)
+		}
+		seeded := bdProxiedShow(t, bd, p.dir, issue.ID)
+		if seeded.ExternalRef == nil || *seeded.ExternalRef != "JIRA-1" {
+			t.Fatalf("external_ref did not seed: %v", seeded.ExternalRef)
+		}
+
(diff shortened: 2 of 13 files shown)


julianknutsen:
## Review: approve (pending CI on retarget)

**Verdict: approve.** Reviewed as the third PR of the write-lifecycle stack, diffed against its stacked parent (`lifecycle-reopen`, e2eed0740); the PR is the single `updateIssue` commit and nothing else. No blocking findings.

### CI

Only the historical cross-version smoke matrix is queued on this head. The core `PR` workflow triggers on PRs based on `main`, so it cannot run against the current stacked base — that is stack plumbing, not a failure. **Before merge, the full PR workflow must be observed green after the base retargets to `main`.**

### What I verified

- **Routing safety, both levels.** The route registers a plain `PATCH /v0/beads/issues/{id}` with no custom-method declaration, pinned by `TestUpdateKeepsItsPatternEqualToItsSpecPath`. `TestPatchOnACustomMethodSegmentIsNeverTheCustomMethod` proves a PATCH on `bd-1:close`/`:reopen`/`:claim` reaches only the Update role with the literal segment as the id, and the proxied test repeats it against a real row (`404`, issue stays open). A PATCH can never execute as a lifecycle verb.
- **Presence semantics.** The body is decoded as raw members; the explicit-null check runs before the typed decode, so `{"title": null}` cannot slide through as "untouched". The nullable set is exactly `estimated_minutes`, `external_ref`, `due_at`, `defer_until`, and `TestNullablePatchMembersAreExactlyTheDocumentsNullableOnes` pins it against the schema's `nullable` flags in both directions. `TestUpdateRefusesNullOnEveryOtherMember` covers all nine remaining members.
- **Vocabulary boundary.** `status`, `assignee`, `metadata`, `parent_id` are refused by name with tests; the handler's accepted sets are gated against both the schema and the generated type at both body levels by `TestUpdateRequestMembersMatchTheHandler`, so neither side can drift silently.
- **Preconditions stay unpublished.** `Claim`, both force flags, and every `Expected*` are asserted zero in the forwarding test, matching the documented code set (no `409`).
- **Error mapping.** `failUpdate`'s `ErrNotFound` branch matches the real sentinel (the storage and issueops names alias one value), and role `ErrValidation` maps to a `400` that never quotes role prose, with the operator-side log asserted.
- **Defensive seams.** The hook-refusing configuration guard covers this operation, since the handler reaches its role only through the same lifecycle accessor; `checkedLifecycle.Update` now refuses a success-without-issue result, with a 500 test.
- **Real-dependency proof.** `TestProxiedServerServeUpdate` delivers the three-way read-back (set / null-cleared / untouched), `changed: false` on a same-value resend, wholesale label replacement, the workspace-vocabulary refusal as a `400`, refused bodies writing nothing, and the CLI parity oracle with an empty allowlist.

### Non-blocking notes

1. **`append_notes` lacks a positive projection test.** The mutual exclusion with `notes` keeps it out of the full-vocabulary walk, so its only coverage is refusal/absent/null; a mis-wire of the notes value into `AppendNotes` would pass the suite. One pure-level case sending `append_notes` alone and asserting the projected value would close the gap.
2. **Wrong-type 400s name `patch`, not the member.** A `"title": 7` reports `param: "patch"` where the null and bounds refusals report `patch.title`. Pinned by tests, so fine as-is; worth a follow-up only if client DX asks for it.
3. **`estimated_minutes` has no minimum.** Negative estimates pass the wire edge and fall to role validation, consistent with the schema as written — flagging only so the asymmetry with `priority`'s enforced range is a deliberate choice.

PR #5423, opened 2026-08-07
feat(httpapi): serve closeIssue over v0

julianknutsen:
## Stack position

**PR 1 of 3** in the v0 write-lifecycle series. Base: `main` (#5410 has merged, so this retargeted from `getissue-include-params`).

```
main
└── lifecycle-close (#5423)   ← this PR
    └── lifecycle-reopen (#5417)
        └── lifecycle-update (#5422)
```

### Two commits

1. `test(serve): assert the advertised capability set from one list` — a **pre-existing** fix, landed separately so it stays attributable. See "Capability drift" below.
2. `feat(httpapi): serve closeIssue over v0` — the operation.

`closeIssue` goes first among the three writes, for two reasons. It is the **highest-value** one — claim → work → close is the agent loop this surface exists to serve, and until this PR the claim was the only mutation v0 published (`engdocs/design/bd-serve-v0.md`, "Claim throughput"), so close is the half that still forces a client back to forking a `bd` subprocess. And it is the **hardest**: it introduces the suffix dispatcher and the `not_closable` code, so it is taken while the tranche is freshest and the two PRs above it inherit finished machinery rather than inventing it.

It inherits from #5410 the JSON media-type document rule and the widened `actor` pattern, so nothing here re-argues either.

## What lands

`POST /v0/beads/issues/{id}:close` — spec, `routeTable` row, `operationCodes` row, handler and tests together, under the existing parity gates.

### The routing change is the substantive part

`{id}:close` is not expressible as a ServeMux pattern for the same reason `{id}:claim` is not — wildcards match a whole segment. Unlike the claim, it **collides**: two rows cannot each register `POST /v0/beads/issues/{idop}`.

The route table grows a `customMethod` field. Rows carrying one share a single registration behind a dispatcher that splits the trailing `:suffix` off the matched segment and hands the request to the row that claims it. A segment ending in no registered suffix gets the same 404 the catch-all gives any unrouted path, so the wide pattern stays a routing detail rather than undocumented surface.

The split happens **before** `s.route`, so an unrouted suffix takes no database slot and books no operation on the request line — where the claim, while it was the only POST here, answered that case from inside its own handler and attributed every probe of the prefix to itself.

`TestSpecRouteParity` is untouched in mechanism: each row still declares its own `specPath`, and the test still compares document paths without learning the exception's mechanics. `TestClaimNarrowsThePOSTSurface` generalizes to `TestCustomMethodsNarrowThePOSTSurface` — the test where a dispatcher that fell back to its first row would be caught, since that failure would otherwise leave the claim's happy path green. A new `TestCustomMethodRowsDeclareTheirDocumentedPath` pins the invariant parity cannot see: a row whose `specPath` ends in one verb while its `customMethod` matches another would route the documented path to the wrong operation with every gate passing.

### Vocabulary

`not_closable` (409) joins the frozen code set, with `open_children` as an optional `Problem` extension member. Both close-policy refusals — open children, and a live blocker — share the code, because they are the same statement to a client and `force` is the bypass for both. The member is attached only for the open-children one, from `*CloseOpenChildrenError`'s own field inside the refusing transaction, so **member presence is the discriminator** and no client parses `3 open child issue(s)` out of prose.

409 rather than the delete precedent's 400: this is a statement about the current state of one named resource, so the same request succeeds or fails on state the client cannot see without reading it. That is the `not_claimable` situation and it gets the `not_claimable` answer.

`already_closed` mirrors `already_claimed` for the idempotent re-close, which under first-close-wins rewrites neither `reason` nor `session`.

### One security-relevant fix

`RoleFiresHooks` learns about `*hookIssueOperations`. A store's own `IssueLifecycle()` returns a role that fires `on_create`, `on_update` and the close hooks for every mutation it lands — precisely the value this server documents it does not run. Without that case `bd serve` would have booted with hooks live on four verbs, and the refusal that exists to prevent exactly this would have passed it.

## Capability drift, found and fixed

CI caught this PR widening the context handshake's capability set while `cmd/bd`'s integration tests asserted the old one. Fixing it turned up the more interesting bug behind it.

The set was written out as a **literal in each topology's test**, on the theory that the two must agree. They had silently stopped agreeing: the proxied copy tracked the surface to twenty entries while the server-mode copy sat at **four**, asserting a shape that had not existed since the facade programme. `TestServerModeServe/context` has therefore been failing on its own terms — I verified it fails identically on `main` with none of my commits applied — and only inside the proxied CI tier, where it reads as one more red in a shard that already had some.

Two literals that MUST match are one literal. The first commit gives both topologies a single `servedCapabilities` and an `assertServedCapabilities` helper; each feature commit then adds exactly its own token, so every commit in this stack is self-contained.

The list stays a **literal** rather than becoming `httpapi.Capabilities()`: comparing the server against its own derivation would pass for any surface at all, and the whole value of the assertion is that a new operation has to be admitted deliberately. What it encodes is the lesson that produced the drift — a package's own gates cannot catch an expectation stored as data outside it, so `internal/httpapi` going green said nothing about these two files.

I swept the rest of the repo for capability literals outside `internal/httpapi`. The only other hits are `cmd/bd/serve.go` and `cmd/bd/serve_test.go`, both of which name `issues.claim` in prose or in a `slices.Contains` membership check — neither breaks when the set grows.

## Gates

- `TestProxiedServerServeLifecycle` and `TestServerModeServe` — green (`BEADS_TEST_PROXIED_SERVER=1`), the two the capability drift was hiding in
- `go test ./internal/storage/...` — green
- Red/green observed per behavior. Two mutations were applied to confirm the new tests bite: a dispatcher falling back to `rows[0]` on an unknown suffix fails all ten `TestCustomMethodsNarrowThePOSTSurface` cases, and dropping the `WithOpenChildren` call fails `TestCloseOpenChildrenIsATypedConflict`.

`TestProxiedServerServeClose` (`cmd/bd`, against real Dolt through a real `bd serve` subprocess):

- **first-close-wins**, read back through a second close: the replay reports `already_closed: true` and the stored `close_reason`/`closed_by_session` still carry the **first** close's values
- **a forced close over two real open children**: the unforced attempt is a 409 carrying `open_children: 2` and writes nothing; `force: true` lands and still reports the 2 it bypassed
- the rule-8 parity oracle against `bd close --json`[0], empty allowlist
- close/claim agreement on the row they share, and refused bodies writing nothing

The role-level transition is already owned against a real store by `internal/storage/uow/lifecycle_close_reopen_contract_test.go`, which this cites rather than duplicates.

`ExpectedVersion` — the role's optimistic-concurrency fence — stays unpublished, and the reason is the frozen code set rather than a scheduling one. `issueops.ErrVersionMismatch` has no entry in `problem.go`'s classification, so an `expected_version` member would render its own refusal as `500 internal`. Publishing a request member later is additive; publishing the status+code pair it needs is the one-way door `bd-serve-v0.md` describes under "The error-code vocabulary", and `TestSpecStatusCodesMatchHandlerTable` is what forces that pair to be admitted deliberately. The fence ships when its code does.

> Note: supersedes #5416, which was auto-closed when its stacked base branch was deleted after #5410 merged. Same head commit (3c455c27d); base is now main.

### DIFF
--- a/cmd/bd/serve.go
+++ b/cmd/bd/serve.go
@@ -180,6 +180,7 @@ func runServe() error {
 			AllowNonLoopback:  serveAllowNonLoopback,
 			Reader:            roles.reader,
 			Claimer:           roles.claimer,
+			Lifecycle:         roles.lifecycle,
 			Settings:          roles.settings,
 			Stats:             roles.stats,
 			CycleDetector:     roles.cycles,
@@ -434,6 +435,7 @@ func serveIssueRoles(src storage.DoltStorage) (serveRoles, error) {
 	for _, b := range []binding{
 		{"issue reader", func() (err error) { roles.reader, err = src.IssueReader(); return }},
 		{"issue claimer", func() (err error) { roles.claimer, err = src.IssueClaimer(); return }},
+		{"issue lifecycle", func() (err error) { roles.lifecycle, err = src.IssueLifecycle(); return }},
 		{"workspace config", func() (err error) { roles.settings, err = src.WorkspaceConfig(); return }},
 		{"stats reporter", func() (err error) { roles.stats, err = src.StatsReporter(); return }},
 		{"cycle detector", func() (err error) { roles.cycles, err = src.CycleDetector(); return }},
@@ -461,6 +463,7 @@ func serveIssueRoles(src storage.DoltStorage) (serveRoles, error) {
 type serveRoles struct {
 	reader       issueops.Reader
 	claimer      issueops.Claimer
+	lifecycle    issueops.Lifecycle
 	settings     issueops.WorkspaceConfig
 	stats        issueops.StatsReporter
 	cycles       issueops.CycleDetector
--- a/cmd/bd/serve_close_proxied_integration_test.go
+++ b/cmd/bd/serve_close_proxied_integration_test.go
@@ -1,0 +1,262 @@
+//go:build cgo
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+// End-to-end for the close, against real Dolt through a real `bd serve`
+// subprocess. The pure tests in internal/httpapi cover the wire edge on a fake
+// role; what only this level can prove is the TRANSACTION — that first-close-
+// wins is a property of the stored row rather than of a fake's bookkeeping, and
+// that the open-children refusal and its force bypass are the real policy the
+// CLI enforces, over a graph a fake has no way to hold.
+//
+// The role-level transition itself is owned against a real store by
+// internal/storage/uow/lifecycle_close_reopen_contract_test.go. This test owns
+// the wire-to-role seam and cites that one rather than duplicating it.
+
+// closeIssue posts a close for id and returns the status and decoded body.
+func (sp *serveProcess) closeIssue(t *testing.T, id, body string) (int, map[string]any) {
+	t.Helper()
+	req, err := http.NewRequest(http.MethodPost, sp.url("/v0/beads/issues/"+id+":close"), strings.NewReader(body))
+	if err != nil {
+		t.Fatalf("new close request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := sp.client.Do(req)
+	if err != nil {
+		t.Fatalf("POST close %s: %v\nstderr:\n%s", id, err, sp.stderr.String())
+	}
+	defer func() { _ = resp.Body.Close() }()
+	raw, err := io.ReadAll(resp.Body)
+	if err != nil {
+		t.Fatalf("read close body: %v", err)
+	}
+	var m map[string]any
+	if len(raw) > 0 {
+		if err := json.Unmarshal(raw, &m); err != nil {
+			t.Fatalf("decode close body %q: %v", raw, err)
+		}
+	}
+	return resp.StatusCode, m
+}
+
+func TestProxiedServerServeClose(t *testing.T) {
+	requireSharedProxiedServer(t)
+	t.Parallel()
+	bd := buildEmbeddedBD(t)
+	p := newSharedProxiedProject(t, bd, "srvclo")
+	sp := startServe(t, bd, p.dir, bdProxiedEnv(p.dir))
+
+	// THE FIRST-CLOSE-WINS PROOF, read back through a second close. A fake role
+	// can report Changed false; only a real store can show that the second
+	// close left the first one's reason and session in the columns.
+	t.Run("the first close wins and a re-close does not rewrite it", func(t *testing.T) {
+		issue := bdProxiedCreate(t, bd, p.dir, "close over http", "-p", "1")
+
+		status, body := sp.closeIssue(t, issue.ID, `{"actor":"http-agent","reason":"shipped it","session":"session-a"}`)
+		if status != http.StatusOK {
+			t.Fatalf("first close: status = %d, want 200: %v", status, body)
+		}
+		if body["already_closed"] != false {
+			t.Errorf("already_closed = %v, want false on a fresh close", body["already_closed"])
+		}
+		closed, ok := body["issue"].(map[string]any)
+		if !ok {
+			t.Fatalf("issue = %#v, want an object", body["issue"])
+		}
+		if closed["status"] != "closed" || closed["close_reason"] != "shipped it" {
+			t.Errorf("the response does not describe the closed row: %v", closed)
+		}
+
+		// The CLI reads the same row through its own path.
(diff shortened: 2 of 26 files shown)


julianknutsen:
## Review: feat(httpapi): serve closeIssue over v0

**Verdict: approve.** Reviewed as this PR's own commit (`3c455c2`) against its stacked parent `getissue-include-params` (`bc4406f`); the diff contains only the close operation plus the routing machinery it introduces. No blocking findings.

### CI

The only suites that ran on this head are the cross-version historical checks (13/13 green). The core PR gates (Go tests, lint, api-check) did not run because `pr.yml` triggers on `pull_request: branches: [main]` and this PR is based on the stacked parent branch. That is stack plumbing, not a red on this code — but the merge condition is a green full PR suite once the PR is retargeted to `main`.

### What I verified in code (not from the PR description)

- **The dispatcher has no fallback.** `customMethodTarget` returns the catch-all 404 for a segment ending in no registered suffix; there is no first-row default. `TestCustomMethodsNarrowThePOSTSurface` covers all ten probe shapes (the GET-only detail path, bare `:claim`/`:close`, case variants, `:claim-not`/`:close-not`, `:unclaim`, `:reopen`, `:delete`), each asserting 404/`not_found`, nothing reaching the CAS, and **no unit of work opened** — pinning that the split happens before `s.route`, so an unrouted probe books no database slot and no operation on the request line.
- **The `RoleFiresHooks` fix is real and pinned.** `HookFiringStore.IssueLifecycle()` returns the hook-firing wrapper (pinned by the existing storage test), the new case makes `RoleFiresHooks` recognize it, and `Listen` refuses via `checkDatabaseSource` over `sourceRoles`, which now includes `Config.Lifecycle`. `cmd/bd/serve_source_test.go` hard-fails if the recognition regresses and separately asserts `serveIssueRoles` hands Listen the lifecycle from **beneath** the hook decorator. Without the recognition case, the boot-time refusal would have silently passed a role that fires hooks on four verbs; with it, getting the wiring wrong is a startup error.
- **The retained real-dependency proof does what it says.** `TestProxiedServerServeClose` (real Dolt, real `bd serve` subprocess): first-close-wins read back through a replay — `already_closed: true` and the stored `close_reason`/`closed_by_session` still the first close's values, confirmed through `bd show`; an unforced close over two real parent-child edges is 409/`not_closable` carrying `open_children: 2` and writes nothing; `force: true` lands and still reports the 2 it bypassed; plus the empty-allowlist parity oracle against `bd close --json`[0] and the closed-then-claim `not_claimable` agreement.
- **Parity gates hold with the `customMethod` field.** `TestSpecRouteParity` is mechanism-unchanged (per-row declared `specPath`, the exception's shape still bounded, and the real `handler()` build — which now collapses shared-pattern rows into one registration — still turns a ServeMux conflict into a test failure). The new `TestCustomMethodRowsDeclareTheirDocumentedPath` closes the hole parity cannot see: a row whose `specPath` verb and `customMethod` disagree, or two rows claiming one suffix. `TestCloseRequestMembersMatchTheHandler` ties the hand-rolled member list to both the schema and the generated type.
- **Vocabulary and discriminator.** `not_closable` is 409 in the status map, in `OpCloseIssue`'s frozen code row, and in the spec's `x-bd-codes`; both policy sentinels map through the shared `ClassifyError`; `open_children` is attached only in `failClose` from the typed error's own field, and `TestFailCloseNeverParsesTheRefusalsProse` pins that an error whose *message* mentions open children but whose type is the blocker sentinel carries no member. The `checkedLifecycle` nil-issue guard converts the would-be panic into a logged 500 with a `request_error` line, tested.

Body validation (member-by-name refusals, null handling, `types.MaxFieldLen` bounds, control-character refusals on values that land in stored columns), `requireNoQuery`/`requireJSONContent`, the shared `bodyActor` lift deduplicating `batchCreateActor`, the id bound at the dispatcher answered as the same 404 a real miss gets, the `issues.close` capability, and the hooks/auto-commit disclaimer in the operation description are all present and tested.

### Non-blocking notes

1. `TestCustomMethodRowsDeclareTheirDocumentedPath` pins exact-duplicate suffixes but not one registered suffix being a proper suffix of another (a hypothetical future `:se` beside `:close` would shadow by table order). No such overlap exists among `:claim`/`:close`/`:reopen` and each op's path-reaches-its-handler test would catch a misroute — future-hardening note only.
2. The httpapi-level `TestListenRefusesARoleThatFiresTheWorkspaceHooks` still exercises only the claimer; the lifecycle path is covered by the shared mechanism plus the cmd/bd preconditions. A symmetric Listen-refusal case for `Lifecycle` would be cheap belt-and-suspenders in a later PR of this stack.
3. The six-line `internal/storage` guard case and the `cmd/bd` wiring ride along in an httpapi PR; acceptable since the recognition is inseparable from wiring the role safely, noted against the one-layer-per-PR guideline.

PR #5424, opened 2026-08-07
docs: document ado.reconcile_interval and inventory every storage span

julianknutsen:
Both docs had their `Last reviewed` markers refreshed today without the content
gaps being closed, so freshness is currently green on two files that are missing
real material. This closes the gaps.

## `docs/integrations/azure-devops.md`

`ado.reconcile_interval` is undocumented. It is real, user-tunable config
(`internal/ado/reconcile.go`, default `10`) controlling how many syncs pass
between automatic reconciliation scans — and the doc already documents the
`--reconcile` flag that forces one, so the periodic behaviour it complements was
invisible.

Adds a Reconciliation section, and notes that `ado.syncs_since_reconcile` is the
counter beads maintains rather than a setting to edit (it shows up in `bd config`
output, so it's worth saying).

## `engdocs/design/otel/otel-data-model.md`

The doc opens with "Complete schema of all telemetry events emitted by Beads"
and lists **7** of **111** storage spans. Everything the issueops role
decomposition added — `IssueOperations.*`, `IssueReader.*`, `Memories.*`,
`WorkspaceConfig.*`, and 35 others — is absent, so anyone building dashboards
from this doc would miss the entire guarded surface.

Rather than 100+ more prose sections that would drift again, this states the rule
the code actually follows and inventories both families:

- span name is `"storage." + operation name`, which is also the `db.operation`
  attribute (`InstrumentedStorage.op`)
- **72** direct `storage.<Method>` spans
- **39** `storage.<Role>.<Method>` spans

It also records a property worth knowing: the two families **do not
double-count**. The role decorators wrap the *uninstrumented* store
(`InstrumentedStorage.IssueLifecycle` calls `Unwrap()` before wrapping), so one
logical operation through the issueops front door emits its role span only, not a
second `storage.<Method>` span beneath it. Counters and duration histograms stay
comparable across both paths.

## Verification

`./scripts/check-doc-freshness.sh` — PASSED. No behaviour change; documentation
catching up to the code.

### DIFF
--- a/docs/integrations/azure-devops.md
+++ b/docs/integrations/azure-devops.md
@@ -255,12 +255,26 @@ When the same issue has been modified both locally and in ADO:
 | `--dry-run` | Preview sync without making changes |
 | `--no-create` | Only update existing items, never create new ones |
 | `--bootstrap-match` | Enable heuristic title matching for first sync |
-| `--reconcile` | Force reconciliation scan for deleted items |
+| `--reconcile` | Force reconciliation scan for deleted items (see Reconciliation below) |
 | `--issues` | Sync specific issues by bead ID or ADO work item ID |
 | `--parent` | Push only a bead and its descendants (push mode only; mutually exclusive with `--issues`) |
 | `--project` | Project name(s) for this run, repeatable — overrides `ado.project`/`ado.projects` |
 | `--states` | Filter by work item states (comma-separated) |
 | `--types` | Filter by work item types (comma-separated) |
+
+### Reconciliation
+
+A reconciliation scan re-checks work items beads already tracks, so items deleted
+in ADO (404) or no longer readable (403) are detected instead of lingering. It
+does not run on every sync — that would cost an API call per tracked item — so it
+runs periodically, and `--reconcile` forces it immediately.
+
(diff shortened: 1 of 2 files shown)


steveyegge:
## Review verdict: MERGE

Docs PR, so the review is a fact-check — and every load-bearing claim verifies against code:

- `ado.reconcile_interval` default 10: confirmed (`internal/ado/reconcile.go` `DefaultReconcileInterval = 10`), and `ado.syncs_since_reconcile` is indeed the maintained counter, worth the not-a-setting note.
- Span naming rule: confirmed at `internal/telemetry/storage.go` — `s.op` starts `"storage."+name` with `db.operation` = bare name.
- **Both inventories are byte-exact.** I extracted the actual op names from code and diffed against your lists: 72 distinct direct names (74 call sites, two names shared) — identical; 39 role names across the telemetry decorators — identical.
- The no-double-count claim: confirmed — every role decorator wraps `s.Unwrap().X()`, so the role span replaces rather than nests the direct span.

The inventory-plus-rule shape is the right anti-drift call versus 100 prose sections. CI green (88 passes). Merging.

PR #5425, opened 2026-08-07
Write down how to retire a test against a contract, and ship the tool for it

julianknutsen:
Follow-up to the four-slice test-deduplication pass (#5398, #5399, #5405, #5406). Two artifacts, no behaviour change.

## Why

That pass retired ~700 lines net. Its adversarial reviews found **nine coverage losses and four tests that could not fail** — and every one of them was green under `go build`, `go vet`, `go test` and `golangci-lint`. The reasoning that caught them lived in review transcripts and a scratch directory. Written down nowhere, the next pass rediscovers it the same way: by shipping the loss first.

## `engdocs/ADDING_AN_ISSUEOPS_ROLE.md` — a new section

Each rule stated as the check that would have caught the failure earlier, not as advice:

- **A mutation verdict is only true of the body you mutated.** Three of the nine losses, in three different slices, independently. `uow` runs its own bodies under `internal/storage/domain/` where the two store backends share others; a store wrapper composes shared functions *itself* while the role path reaches them via `runTransaction` and never calls the wrapper. Red/red against the shared body says nothing about the wrapper the same test was watching.
- **A wrapper is a composition the contract structurally cannot see.** Deleting the only observers of `DoltStore`'s checked writes left `ExpectedStatus` routed on **no branch**, and disabling a wrapper's CAS passed **all 136 contract cases**.
- **Write a residue from the decision surface, not the mutants you ran.** The slice that kept a residue still lost six things, because it wrote the file from the two breaks it had measured. Half of what it missed were **positive** assertions — a wrapper that refuses correctly and never writes passed the whole suite. Refusal-only coverage of a guarded write is half a test.
- **Ask what the fixture makes unobservable.** All four cannot-fail tests were *fixture* defects, not assertion defects: a flag seeded with no edge so the guard short-circuited; ready ids `1` and `10` whose natural and lexical orders coincide; a key collided in two patch stages that a later stage removes — and a key any later stage removes cannot witness the order of the stages before it.
- **Never let a role-answer assertion replace a raw-row one.** Reading back through the role is exactly the check that passes on a corrupted table.
- **Cite by symbol, not line.** Two sweeps re-resolved 23 stale citations. Wrong ones look exactly like right ones, which is how one slice copied a stale citation into new prose and minted a second beside it.

The section extends a thread the file already ends on — its last existing lesson is also about a test that could not fail.

## `scripts/mutation-equivalence.sh` — the tool

It ran the entire pass from `/data/tmp`, which means it would not have survived it. Every guard exists because its absence produced a wrong answer during the pass:

| guard | what happened without it |
| --- | --- |
| mutates a **disposable worktree** | a trap does not survive SIGKILL; interrupted runs stranded broken production code in the branch worktree twice |
| fatal on `-run` matching nothing | `go test -run NoSuchTest` **exits 0**, so a typo reads as PASS forever |
| fatal on a zero-byte mutation | reports agreement |
| fatal on a non-green baseline | the verdict is noise |
| fatal on a dirty worktree at start | a killed run's mutation becomes the next run's baseline |
| result greps do not over-anchor whitespace | `go test -v` indents subtests by 4 and **nested** subtests by 8 — `scripts/conformance.sh` documents the mirror image of this hazard for its own column-0 case |

Both refusals verified firing. The `REDUNDANT` verdict deliberately reads as qualified rather than as a green light —

```
REDUNDANT    B catches this too — for THIS body, and for THIS promise.
             A retires only when every promise it pins has its own verdict.
```

— because both qualifiers were violated in practice during the pass.

## Verification

`go build ./...`, `golangci-lint run ./scripts/...`, `bash -n` clean. Harness smoke-tested end to end against a real pairing (correct REDUNDANT verdict, working tree untouched), and both refusal paths exercised.

### DIFF
--- a/engdocs/ADDING_AN_ISSUEOPS_ROLE.md
+++ b/engdocs/ADDING_AN_ISSUEOPS_ROLE.md
@@ -422,19 +422,24 @@ may be a row that cannot fail.**
 
 Once a role has a contract, the ad-hoc tests that predate it start to look
 redundant. Some are. A four-slice pass over ~78k LOC of backend tests retired
-about 700 lines net and, in the process, its adversarial reviews found **nine
-coverage losses and four tests that could not fail**. Every one of those was
-green under `go build`, `go vet`, `go test` and `golangci-lint`. The rules
-below are what the reviews actually caught, each stated as the check that would
-have caught it earlier.
+about 700 lines net. Along the way the slices and their reviews found **about a
+dozen coverage losses and six tests that could not fail**, every one of them
+green under `go build`, `go vet`, `go test` and `golangci-lint`. Two of the six
+could not fail by construction — unconditional `t.Skip` bodies with no
+assertion. The other four are the harder class, and most of what follows is
+about them. Treat the counts as an order of magnitude: the per-slice tallies
+live in the pull requests, and review revised them more than once.
 
-**A mutation verdict is only true of the body you mutated.** This produced three
-of the nine losses, in three different slices, independently. A test is broken,
+**A mutation verdict is only true of the body you mutated.** This is the most
+productive rule here — it produced confirmed losses in more than one slice
+independently, and reviews kept finding it after it had been written down. A
+test is broken,
 the contract goes red on the same break, the verdict reads REDUNDANT — and the
 deleted test was watching a different body. `uow` runs its own bodies under
 `internal/storage/domain/` where the two store backends share others; a store
 wrapper composes shared functions *itself* while the role path reaches the same
-functions through `runTransaction` and never calls the wrapper at all. Before
+functions through `runIssueOperationTx` and never calls the wrapper at all.
+Before
 accepting red/red: **name the body the deleted assertion observes, and confirm
 your mutation was in it.** Nothing automates this.
 
@@ -443,7 +448,8 @@ your mutation was in it.** Nothing automates this.
 `ClaimIssue` — published on `storage.DoltStorage`, with option types in
 `beads.go` — which assemble the shared bodies on their own. Deleting their only
 observers left `ExpectedStatus` routed on no branch at all, and disabling a
-wrapper's compare-and-set passed **all 136 contract cases**. When a wrapper has
+wrapper's compare-and-set passed every contract case on that backend. When a
+wrapper has
 no other caller, keep a narrow routing residue: see
 `internal/storage/dolt/checked_wrapper_smoke_test.go`.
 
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Fact-check review — and it found more wrong with this PR than with anything it describes

A reviewer checked every claim in the new section against the tree, and separately tried to break the script. Corrections pushed.

### The doc broke its own closing rule

The section's last rule is **"cite promises by symbol, not by line — wrong ones look exactly like right ones."** Four lines above it, the section cited `runTransaction`. **That symbol does not exist in the dolt package.** The role path there is `runIssueOperationTx`, and `checked_wrapper_smoke_test.go`'s own header says so correctly. The loose symbol came from PR prose and was never checked against the code.

### And a claim this very pass had already falsified

It said `is_blocked` is derived and persisted "and no contract case currently reads it." There are **45 references across ten files** — and several of those readers were **added by this pass**, because the role answer alone couldn't see the defect. The sentence was true when a slice wrote it mid-pass; writing the doc afterwards fossilized it. `is_blocked` is now the standing *example* of why a raw read is load-bearing, which is what the evidence actually supports.

### The counts didn't reconcile with the merged record

- "nine coverage losses, three of them in three different slices" — no partition of #5398/#5399/#5405/#5406 produces it. The record is 11–12 losses; wrong-body ones in two slices, not three.
- "four tests that could not fail" was **curated**: there are six. Two were unconditional `t.Skip` bodies, dropped so that "all four were fixture defects" would read true.
- "all 136 contract cases" is not reproducible from the tree at any commit.

All replaced with claims that survive checking, plus an explicit instruction to read the counts as orders of magnitude with the PRs as the record. A doc that teaches a rule with a wrong example loses the rule along with the example.

### Three demonstrated false-REDUNDANT routes in the script

That's the one wrong answer that costs coverage, so each is now fatal or reclassified:

| route | before | now |
| --- | --- | --- |
| build failure or timeout | exits nonzero → read as "B caught it" | classified apart, reported `UNUSABLE` |
| stale `MUTEQ_WT` | measured a **different commit**, exit 0 | checkout/reset failures fatal + HEAD assertion |
| mutation edits unnamed files | diff showed something harmless | fatal, names the stray file |

Plus `A == B` refused as tautological.

### The guard that would have shipped broken

Pointing `MUTEQ_WT` at a live checkout **destroyed it** — the path is hard-reset and cleaned. My first guard refused the repo *by accident*: the repo appears in its own `git worktree list`, so a slightly different setup would have wiped the caller's own tree. Now an explicit self-check plus a **detached-HEAD requirement**, because a worktree with a branch checked out is somebody's working copy, not scratch space.

Every guard verified by making it fire — including recovering a `precious.txt` from a path the tool was told to use and correctly refused.

### Verified true, for the record

~78k LOC surveyed; uow's bodies under `internal/storage/domain/` vs the stores sharing `internal/storage/issueops`; all three wrapper methods published on `storage.DoltStorage` with option types in `beads.go`; the three named cannot-fail fixtures all real history; 23 stale citations (10 + 13) exact; and `scripts/conformance.sh`'s comment is a genuine mirror image — same indentation fact, opposite anchoring requirement.

PR #5428, opened 2026-08-07
Give issueops.Claimer a conformance contract, and pin the wrapper divergence it exposed

julianknutsen:
`issueops.Claimer` had a detailed written specification, **three independent implementations**, a hook decorator, an accessor on `storage.DoltStorage` — and no conformance contract. 23 contract files and none was Claimer's; `ready_claimer_contract.go` covers `ReadyClaimer.ClaimNext`, a different role. Consequence, verified before starting: the dolt and embeddeddolt role implementations had **no tests referencing them at all**, and the only role-level coverage anywhere was one uow file.

8 cases × 3 legs, `+1,195 / −8`.

## The cases, and the mutation that reddens each

Every case verified RED on **both** bodies — the shared `internal/storage/issueops` one (dolt + embedded) and the uow one — because a verdict is only true of the body you mutated. Every uow mutant left the dolt case green and vice versa, which is what confirms the two bodies are independently pinned.

| case | shared-body mutant | uow-body mutant |
| --- | --- | --- |
| `ClaimsAnUnassignedOpenIssueAndAnswersTheBareRow` | drop `issue.Labels = nil` in `ExecuteClaim` | `Changed: true`→`false` |
| `TakesAnOpenIssueTheActorAlreadyHolds` | poison `assignee = ?` in `ClaimIssueInTx` | same in `db.Claim` |
| `ReclaimsItsOwnInProgressIssueWithoutWriting` | force `changed` true | clear `AlreadyClaimed` |
| `AcceptsAConfiguredActiveStatusAndRefusesAConfiguredWipOne` | `ClaimableSourceStatusesInTx` stops appending actives *(shared helper — reddens all three)* | ↩ |
| `TakesAPoolAssignedIssueButRefusesAForeignHolder` | drop the pools clause | drop the pools loop |
| `RefusesABuiltInIneligibleStatusWithTheStateThatRefusedIt` | add `in_progress` to claimable statuses *(all three)* | ↩ |
| `RefusesAWispIDAsNotFound` | drop `ExecuteClaim`'s wisp refusal | drop `IsNoRows`→`ErrNotFound` |
| `RefusesIncompleteRequestsAndAnAbsentIDWithoutTouchingState` | drop the empty-field guard | drop the uow guard |

## The divergence this exposed

`ExecuteClaim` refuses every wisp id **before** the CAS, so `Claimer.Claim` answers `ErrNotFound` — the leaf says so outright: *"A wisp id is ErrNotFound: the wisp plane is not claimable through this role."*

But `DoltStore.ClaimIssue` and `EmbeddedDoltStore.ClaimIssue` route a wisp to `claimWisp` and **claim it**. `claimWisp` had one caller and no test. Two published surfaces — both on `storage.DoltStorage` — giving opposite answers to the same operation, and structurally unreachable by any contract case, because the contract only sees the role.

Both sides are now pinned, not reconciled: `RunClaimerRefusesAWispIDAsNotFound` for the role, `ClaimsAWispInTheWispsPlane` in the wrapper residue. **Which one is wrong is a design call, not a test call** — this PR makes the disagreement visible and regression-proof rather than choosing for you.

## Wrapper residue

`{dolt,embeddeddolt}/claim_wrapper_smoke_test.go` — the third such residue in this codebase, after the checked-write pair. Covers the wisp win, the wisp refusal for a foreign holder, and dolt's own commit message (`"bd: claim <id>"`, which differs from `ClaimCommitMessage`). Each verified red with the full 8-case contract green, which is the point: nothing in the contract layer can see any of it.

## A test that could not fail, removed

A residue case for "a wisp claim takes no Dolt commit" was written and then deleted, because mutations in **opposite directions** both left it green: routing the wisp through the durable branch stages issues+events (which a wisp claim never dirties), and adding a commit to `claimWisp` stages nothing because `DOLT_ADD` ignores an ignored table. The property belongs to `dolt_ignore`, not to the wrapper. Reasoning left in the file.

## No implementation disagreements

All 8 cases pass identically on all three legs — the hand-maintained `domain/db` dual is in lockstep. One cosmetic asymmetry found and deliberately not pinned: the store body double-wraps `*ClaimConflictError` where uow wraps once; `errors.As` and the carried fields agree.

## Spec gaps, reported not papered over

`ErrValidation` is answered by all three implementations for the empty-field refusals but is **not named** in `Claimer`'s doc, unlike every sibling role. Asserted with the omission flagged — closing it is a leaf-doc edit needing owner sign-off. Also unpinnable as written: the "readable in the same transaction" qualifier's negative half, "never mutate caller-owned request values" (`ClaimRequest` has no reference members), retry, and the "detached" snapshot. The bare-row promise is falsifiable only on the shared body, since uow's read-back never hydrates labels and satisfies it structurally.

## Verification

`go build ./...`, `go vet`, `golangci-lint run` — clean. Contract and both residues green on dolt, uow and embeddeddolt. No `.github` change needed: the embedded shard script auto-discovers and hash-distributes new tests (verified), and no workflow enumerates storage contract names.


julianknutsen:
## Consolidated with #5412

#5412 landed a Claimer contract while this branch was being built. Rather than ship a second one, this now consolidates them — per the owner's call.

**Neither version dominated.** Theirs: 4 cases + aggregator, pinning the refusal *prose*. Mine: 8 cases, pinning claim pools, configured active statuses and the bare-row promise. But the wiring decided it:

Theirs ran through `conformance.RunAll`, whose `Factory` is `func(*testing.T) storage.DoltStorage`. **A unit-of-work provider is not one**, so only the two store legs could reach it — and those two share `internal/storage/issueops.ExecuteClaim`. Every promise it pinned was therefore compared between two implementations that already run the same code, and never against the one backend that doesn't: uow, whose `Claim` lives in `internal/storage/domain/db`. Same audit-suite gap the facade programme hit earlier.

**What changed:**

- `claimer_contract.go` — now **9 cases**, wired at all three legs through role accessors
- `issue_claimer_contract.go` — deleted
- `conformance.go` — `ClaimerRole` hook removed, with a comment recording why the role cannot be tested from that suite

**Their case was kept, and it is the strongest thing in either file.** `RunClaimerRefusalMessagesCarryTheirFragments` pins that the open-but-assigned refusal deliberately **omits** the `claimed by` fragment while the typed `*ClaimConflictError` still carries the holder — so a caller reading prose learns less than one reading the struct, on purpose. Nothing outside that case says so, and it has now run on uow for the first time.

Green: 19 subtests at each store leg, 34 at uow. `go build`, `go vet`, `golangci-lint` clean.

julianknutsen:
## Red-team review — one confirmed coverage loss in the consolidation, now fixed

The reviewer's first task was the one that mattered: assertion-by-assertion, **did deleting `issue_claimer_contract.go` lose anything?** It did.

### The loss

The deleted contract counted **claim events on the issue** (`CountClaimEvents` over `GetEvents`). This file replaced that with a branch-level history delta plus an `updated_at` comparison. Both are blind on the store legs by construction:

- `updated_at` is `DATETIME` with **no fractional precision**, so a first claim and a re-claim in the same second tie. The comment's claimed detector cannot fire.
- `Changed` is computed from the **pre-image**, so it stays false, `ChangedTables` stays empty, and the Dolt commit is skipped — the event row is written but never reaches history, which is all a history delta can see.

Proof: widen the CAS so a re-claim by the current holder re-runs. Old contract at `origin/main`: **FAIL** — *"claim events went 1 → 2 across the idempotent re-claim."* New contract: **PASS** on dolt, **PASS** on uow. The only survivor-catcher left anywhere was a dolt-only wrapper test; the role seam and the embedded leg kept no guard.

**Fixed** in `d3dd70bb4` — the per-issue event count goes back, through the `QueryScalar` hook the fixture already carries, so no new fixture surface. Also restores the returned `Issue.ID` assertion, which the old snapshot check covered and this one had dropped.

Verified: under a mutation that records an event on the idempotent path, dolt **FAIL** and embeddeddolt **FAIL**, where before the consolidation caught neither.

Worth recording how that verification went, since it is the tool's one sharp edge: two earlier attempts returned `INCONCLUSIVE`, and I diagnosed the mutation's placement and then the helper's SQL before checking the cheapest thing — whether the code under test was committed. `mutation-equivalence.sh` measures `HEAD`, not the working tree. It now warns (#5425).

### Verified correct: the deletion rationale

`Factory` is `func(*testing.T) storage.DoltStorage` (`conformance.go:117`); `RunAll`'s only callers are `dolt/conformance_test.go` and `embeddeddolt/conformance_test.go`; uow's provider implements `UnitOfWorkProvider`, not `DoltStorage`. Both reachable legs share `ExecuteClaim` — so the old contract genuinely never voted against the uow dual, and the absorbed fragments case now runs against uow's independent prose producer for the first time.

### Other findings, and what was done

- **Cases can fail, per body family.** The custom-status case reddens two-sided (store-family kill → dolt red, uow green; uow-family kill → the reverse), which also proves `SetConfig` reaches claim time on both. The pools case reddens on the uow family. No blockers found in the rest.
- **The wrapper residue's routing cases pin end-to-end plane behaviour, not the wrapper's own branch** — deleting the branch leaves them green, because `ClaimIssueInTx` re-makes the plane decision internally. What the residue does own: `claimWisp` aliveness, the ephemeral-plane refusal state, and dolt's commit message. Left as-is and worth saying plainly: the reviewer's suggestion that the redundant wrapper branch could simply be deleted in production is a design call, not a test call.
- **Fixture soundness cleared.** Every seed is pre-asserted as a raw row, which closes the seeded-flag-with-no-edge hole that produced cannot-fail tests earlier in this programme.
- **Two self-declared tripwires noted honestly**: case 1's `Labels`-nil check is vacuous on the uow leg specifically, because that read path never hydrates labels.
- **`RunAll` reachability for a hypothetical fourth backend** narrows — the new fixture needs `SetConfig`/`QueryScalar`/`CountHistory`. Affects no in-tree leg, documented in `conformance.go`, and consistent with every other role contract.

All 9 cases confirmed executing at all three legs **by name**, none parked via `skipKnownDivergence`.

PR #5429, opened 2026-08-07
feat(httpapi): publish the v0 dependency-edit operations

julianknutsen:
Publishes the two dependency-graph WRITES on the `/v0` surface, completing the
graph half of the v0 write contract. Until now a client could list edges, walk
the tree and find cycles over HTTP, but had to fork `bd dep add` / `bd dep
remove` to change any of it.

Both operations are thin wire adapters over `issueops.DependencyEditor`. Every
graph rule — the cycle gate, the parent-child-first ordering, the whole-graph
final gate, the plane routing, the event stream — stays in the role; this PR
decodes bodies, refuses the shapes the document refuses, and maps the role's
TYPED refusals onto frozen codes.

Two commits, one operation each, `removeDependency` first: it publishes no new
code (see the table below), so the commit that freezes `dependency_cycle` and
`dependency_exists` into the vocabulary stands on its own. Adding a
status+code pair is the one-way door `engdocs/design/bd-serve-v0.md` describes
under "The error-code vocabulary", and `TestSpecStatusCodesMatchHandlerTable`
is what forces it to be an admitted, reviewable act rather than a side effect.
Each commit carries its spec entry, `routeTable` row, `operationCodes` row,
handler and tests together, and each is green on its own.

| Operation | Route | Capability | New codes |
|---|---|---|---|
| `removeDependency` | `POST /v0/beads/dependencies:remove` | `dependencies.remove` | — |
| `addDependencies` | `POST /v0/beads/dependencies:add` | `dependencies.add` | `dependency_cycle` (409), `dependency_exists` (409) |

## removeDependency

A collection-level custom method rather than a `DELETE`: an edge is named by
TWO endpoints, so there is no single-segment resource path for a `DELETE` to
address, and a `DELETE` carrying a body is the shape proxies mangle.
`DELETE /v0/beads/memories/{key}` is a `DELETE` for the opposite reason — one
named resource, no body.

It is idempotent and its code set has **no 404**. An edge that is not there is
`removed: false` inside a `200`, and an endpoint id that names nothing holds no
edge either, so the operation probes no id's existence and has nothing it could
report a miss on. An agent replaying its own teardown must not have to classify
an error to learn it already ran.

## addDependencies

All-or-nothing, which is the role's shape rather than a policy choice: edges
asserted together describe a graph, and half a graph is a graph nobody asked
for — the cycle a caller was refused for is exactly the state a partial commit
would leave behind. `batchCreateIssues`' transaction posture, not
`BatchCloser`'s keep-the-survivors one.

`Problem` gains five optional extension members, additively:

- `existing_type` / `requested_type` with `dependency_exists`, from
  `*DependencyTypeConflictError`.
- `issue_id`, `blocker_id`, `blocker_is_ancestor` with `dependency_cycle`, and
  **only on the hierarchy refusal**.

`dependency_cycle` is ONE code for both never-makes-progress refusals — the
plain scheduling cycle and the blocking edge against the issue's own ancestor
or descendant. They have one client recovery (rethink the edge; neither has a
force bypass), and codes are the vocabulary of recovery. The typed distinction
is not lost: member PRESENCE is the discriminator, and the three hierarchy
members are enough to rebuild `*DependencyHierarchyConflictError` whole,
`blocker_is_ancestor` in both polarities included — which is why it travels
through a pointer and is emitted when false. Nothing can recover them after the
fact: the conflicting hierarchy may exist only inside the batch that was rolled
back, so the refusing transaction is the only place they can come from.

The endpoint refusals typed in #5415 map to their documented `400` and name the
refused edge as `edges[i].issue_id` / `edges[i].depends_on_id`, found by BOTH
endpoints — which is what `DependencyEndpointNotFoundError` carrying the whole
edge is for. A target may legitimately be an `external:` reference or another
repository's id, so only a locally-visible absence refuses. A self-dependency is
refused at the wire edge as a VALUE and not a conflict: it is invalid whatever
the graph holds, and answering it there is what lets the refusal name the edge.
`SkipPerEdgeCycleCheck` stays unpublished — an unauthenticated surface is where
a default must be the guarded one.

## Wiring

`DependencyEditor` joins the store-shaped `Config` role set on the same terms as
`Sweeper` and `Deleter`: required, so a build that serves these operations
cannot bind with a `Config` that cannot answer them. On the provider path the
accessor is built OVER the timing wrapper, so the write transaction lands in the
request's `uow_ms`.

`storage.RoleFiresHooks` learns `*hookDependencyEditor`.
`HookFiringStore.DependencyEditor` recurses and fires the workspace's update
hook per edited source issue, so the obvious `store.DependencyEditor()` is a
role this server must refuse — `bd serve` documents that hooks do not fire, and
that refusal at `Listen` is the difference between a startup error and a
subprocess per edge nobody notices. `serveIssueRoles` takes it off the peeled
store, as it does the claimer.

The `actor` decode was on its third copy of the same eighteen lines, so it is
now one shared `requiredActorMember`; the claim and batch-create paths call it.

## Testing

Two halves, the `claim_test.go` / `TestProxiedServerServeClaim` precedent.

**Wire edge** (`internal/httpapi/dependency_edit_test.go`, pure, no database):
path reachability, member forwarding, the echo in request order, the whole 400
vocabulary with `edges[i].member` params, media type, query refusal, and every
typed 409/400 mapping including both polarities of `blocker_is_ancestor` and the
member-absence discriminator. Refusal cases also assert the role was never
called, so nothing decidable from the request buys a write transaction.

**Real boundary** (`cmd/bd/serve_dependencies_proxied_integration_test.go`,
`//go:build cgo`, real `bd serve` + real Dolt) — what a fake structurally cannot
prove:

- **the atomicity read-back**: a batch whose later edge closes a cycle earns the
  `409` and leaves ZERO edges on all three issues, read back through the
  stored-edge endpoint and cross-checked with `bd dep list`;
- the idempotent re-remove: `removed: true`, graph empty, `removed: false`,
  graph unchanged;
- the retype `409` carrying both types, and the hierarchy `409` carrying all
  three members, produced by the real role against a real parent-child edge;
- the narrow endpoint rule from both sides: a locally-absent target and a ghost
  source are `400`s, while an `external:` reference and a foreign-prefix id are
  accepted.

## Gates

`go test ./internal/httpapi/` (all parity tests), `make api-check`,
`make ci-pr-lint`, and the proxied suite above under
`BEADS_TEST_PROXIED_SERVER=1` — all green.

## Rebased onto the landed lifecycle stack

This branch is rebased onto `main` with #5423, #5417 and #5422 merged. Neither
operation needed the suffix dispatcher that stack introduced — both paths are
collection-level LITERALS (`ready:count`'s spelling), so `pattern == specPath`
and they register themselves. The reconciliation was five mechanical points:

- **Actor decode.** The stack landed `s.bodyActor`; the helper this branch had
  added is gone and both handlers call theirs. ONE helper.
- **`RoleFiresHooks`.** Composes as three separate cases —
  `*hookIssueClaimer`, `*hookIssueOperations`, `*hookDependencyEditor` — each
  with its own reason.
- **`serveIssueRoles` fakes.** The editor gets the identifiable-value treatment
  beside the lifecycle's, in the same style: a precondition that the
  decorator's own accessor answers `RoleFiresHooks` true, then a post-peel
  assertion that the role is the middle store's own. Without the precondition
  the peel assertion would pass on a predicate that answers false for
  everything, which is what holds the new `RoleFiresHooks` case in place.
  `serveIdentityStore` grows the accessor for the same reason: it embeds a nil
  `DoltStorage`, so an unstubbed role is a nil-pointer panic in `serveIssueRoles`.
- **`servedCapabilities`.** `dependencies.remove` and `dependencies.add` join
  that one list, each in the commit that publishes it.
- **Counts.** `sourceRoles`, `roleSourceNames` and the `timedProvider`
  accessor floor all reconciled with every lifecycle role present.

### DIFF
--- a/cmd/bd/serve.go
+++ b/cmd/bd/serve.go
@@ -192,6 +192,7 @@ func runServe() error {
 			Sweeper:           roles.sweeper,
 			Deleter:           roles.deleter,
 			BatchCreator:      roles.batchCreator,
+			DependencyEditor:  roles.dependencyEditor,
 			Memories:          roles.memories,
 			Workspace:         info,
 			SchemaVersion:     JSONSchemaVersion,
@@ -447,6 +448,7 @@ func serveIssueRoles(src storage.DoltStorage) (serveRoles, error) {
 		{"sweeper", func() (err error) { roles.sweeper, err = src.Sweeper(); return }},
 		{"deleter", func() (err error) { roles.deleter, err = src.Deleter(); return }},
 		{"batch creator", func() (err error) { roles.batchCreator, err = src.BatchCreator(); return }},
+		{"dependency editor", func() (err error) { roles.dependencyEditor, err = src.DependencyEditor(); return }},
 		{"memories", func() (err error) { roles.memories, err = src.Memories(); return }},
 	} {
 		if err := b.get(); err != nil {
@@ -475,6 +477,11 @@ type serveRoles struct {
 	sweeper      issueops.Sweeper
 	deleter      issueops.Deleter
 	batchCreator issueops.BatchCreator
+	// dependencyEditor is the second role here whose accessor recurses through
+	// the hook decorator, so taking it off the peeled store is not optional:
+	// HookFiringStore.DependencyEditor fires the workspace's update hook per
+	// edited source issue, and this server documents that hooks do not fire.
+	dependencyEditor issueops.DependencyEditor
 	// memories is the one role here that is not an issueops role: the memory
 	// plane is user data riding in the config table under its own merge class,
 	// so it has its own leaf package.
--- a/cmd/bd/serve_dependencies_proxied_integration_test.go
+++ b/cmd/bd/serve_dependencies_proxied_integration_test.go
@@ -78,6 +78,209 @@ func (sp *serveProcess) removeDependency(t *testing.T, issueID, dependsOnID, act
 		fmt.Sprintf(`{"actor":%q,"issue_id":%q,"depends_on_id":%q}`, actor, issueID, dependsOnID))
 }
 
+func (sp *serveProcess) addDependencies(t *testing.T, actor string, edges ...string) (int, map[string]any) {
+	t.Helper()
+	return sp.postJSON(t, "/v0/beads/dependencies:add",
+		fmt.Sprintf(`{"actor":%q,"edges":[%s]}`, actor, strings.Join(edges, ",")))
+}
+
+func edgeJSON(issueID, dependsOnID, edgeType string) string {
+	return fmt.Sprintf(`{"issue_id":%q,"depends_on_id":%q,"type":%q}`, issueID, dependsOnID, edgeType)
+}
+
+func TestProxiedServerServeAddDependencies(t *testing.T) {
+	requireSharedProxiedServer(t)
+	t.Parallel()
+	bd := buildEmbeddedBD(t)
+	p := newSharedProxiedProject(t, bd, "srvdepadd")
+	sp := startServe(t, bd, p.dir, bdProxiedEnv(p.dir))
+
+	// THE ATOMICITY PROOF, and the thing a fake role structurally cannot own.
+	// A batch whose LATER edge closes a cycle is refused, and the earlier edges
+	// — which were individually fine and really were written inside the
+	// transaction — must be gone when the graph is read back. A handler that
+	// forwarded the batch to a role committing per edge would pass every pure
+	// test in internal/httpapi and leave half a graph here.
+	t.Run("a mid-batch refusal leaves zero edges", func(t *testing.T) {
+		a := bdProxiedCreate(t, bd, p.dir, "atomicity a", "-p", "1")
+		b := bdProxiedCreate(t, bd, p.dir, "atomicity b", "-p", "1")
+		c := bdProxiedCreate(t, bd, p.dir, "atomicity c", "-p", "1")
+
+		status, body := sp.addDependencies(t, "http-agent",
+			edgeJSON(a.ID, b.ID, "blocks"),
+			edgeJSON(b.ID, c.ID, "blocks"),
+			// The closing edge. Nothing before it is wrong; the SET is.
+			edgeJSON(c.ID, a.ID, "blocks"),
+		)
+		if status != http.StatusConflict {
+			t.Fatalf("status = %d, want 409: %v", status, body)
+		}
+		if body["code"] != "dependency_cycle" {
+			t.Fatalf("code = %v, want dependency_cycle", body["code"])
+		}
+		if body["request_id"] == nil {
+			t.Error("no request_id on the problem body")
+		}
+
+		for _, id := range []string{a.ID, b.ID, c.ID} {
+			if edges := sp.storedEdges(t, id); len(edges) != 0 {
+				t.Errorf("the refused batch left %d edge(s) on %s: %v — the request is all-or-nothing", len(edges), id, edges)
+			}
+		}
+		// And the CLI reads the same empty graph through its own path.
+		if out := bdProxiedDep(t, bd, p.dir, "list", a.ID); strings.Contains(out, b.ID) {
+			t.Errorf("`bd dep list` shows an edge from the refused batch:\n%s", out)
+		}
+	})
+
+	t.Run("every edge of an accepted batch lands", func(t *testing.T) {
+		a := bdProxiedCreate(t, bd, p.dir, "batch source", "-p", "1")
+		b := bdProxiedCreate(t, bd, p.dir, "batch target", "-p", "1")
+		c := bdProxiedCreate(t, bd, p.dir, "batch other target", "-p", "1")
+
+		status, body := sp.addDependencies(t, "http-agent",
+			edgeJSON(a.ID, b.ID, "blocks"),
+			edgeJSON(a.ID, c.ID, "related"),
+		)
+		if status != http.StatusOK {
+			t.Fatalf("status = %d, want 200: %v", status, body)
+		}
+		added, ok := body["added"].([]any)
+		if !ok || len(added) != 2 {
+			t.Fatalf("added = %#v, want the two requested edges", body["added"])
+		}
+
+		stored := map[string]string{}
+		for _, edge := range sp.storedEdges(t, a.ID) {
+			target, _ := edge["depends_on_id"].(string)
+			edgeType, _ := edge["type"].(string)
+			stored[target] = edgeType
+		}
+		if stored[b.ID] != "blocks" || stored[c.ID] != "related" {
+			t.Errorf("the graph holds %v, want %s->blocks and %s->related", stored, b.ID, c.ID)
+		}
+
+		// A same-type re-add refuses nothing and writes nothing new.
+		status, body = sp.addDependencies(t, "http-agent", edgeJSON(a.ID, b.ID, "blocks"))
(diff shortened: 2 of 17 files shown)


julianknutsen:
**Verdict: request changes** — one blocking item, and it is small. The design and test discipline in this PR are excellent; a missing five-line test stub in `cmd/bd` breaks CI on both platforms.

## Blocking

### 1. `TestServeIssueRolesComeFromBeneathTheHookDecorator` panics — CI is red

Commit c9c4496d1 teaches `serveIssueRoles` to fetch the dependency editor off the peeled store:

```go
{"dependency editor", func() (err error) { roles.dependencyEditor, err = src.DependencyEditor(); return }},
```

but `serveRolesStore` — the fake that `TestServeIssueRolesComeFromBeneathTheHookDecorator` drives this function with — stubs every other role and not this one. `DependencyEditor()` promotes through the fake's embedded nil `storage.DoltStorage`, and the call is a nil-pointer panic:

```
--- FAIL: TestServeIssueRolesComeFromBeneathTheHookDecorator (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference
  github.com/steveyegge/beads/cmd/bd.(*serveRolesStore).DependencyEditor(...)  <autogenerated>
  cmd/bd/serve.go:449  serveIssueRoles.func14
  cmd/bd/serve_source_test.go:110
FAIL  github.com/steveyegge/beads/cmd/bd
```

Both `Test (macos-latest)` and `PR Core (wrapper timing)` fail on exactly this, and `CI Gate / Required` is red. The PR's stated gates (`go test ./internal/httpapi/`, `make api-check`, `make ci-pr-lint`, the proxied suite) never ran the plain `cmd/bd` package — the "affected neighbors when their contract changed" run TESTING.md asks for, and `serveIssueRoles` is precisely a changed contract.

**Requested fix — please take the fuller option.** The one-line fix is a `DependencyEditor() (issueops.DependencyEditor, error) { return nil, nil }` stub. But the better fix is the one the open closeIssue PR (#5423) applies to the same test for its lifecycle role: give the fake an identifiable editor and extend the assertions —

- precondition: `storage.RoleFiresHooks(chained.DependencyEditor())` is true (so the peel assertion cannot pass on a predicate that answers false for everything);
- post-peel: `roles.dependencyEditor` is the middle store's own editor and `RoleFiresHooks` answers false for it.

That also closes a real gap this PR leaves open: the `RoleFiresHooks` switch now recognizes `*hookDependencyEditor` — the guard `Listen` uses to refuse a role that would fork a user subprocess per edited source issue — but **no test pins that recognition**. Only the claimer's refusal is pinned (`TestListenRefusesARoleThatFiresTheWorkspaceHooks`). Since #5423 edits the same switch (appending its own case while this PR rewrites the existing case line), a mis-merge that drops the editor case would today be silent: `bd serve` would still boot (it peels), and an embedder handing `Listen` a store's own editor would silently run workspace hooks per edge — exactly the failure class this guard exists to make loud.

## Non-blocking notes

**The coordination note undersells the overlap with the lifecycle stack.** It says the only shared files are `routes.go`, `problem.go` and the spec, "appends rather than edits" — true for those three, and the routing claim is correct (both new paths are collection-level literals, self-registering, with no dependency on #5423's suffix dispatcher). But both this PR and #5423 also delete `batchCreateActor` and extract the same actor decode — here into a free `requiredActorMember(members) (string, *Result)`, there into a method `s.bodyActor(w, r, members)` — and both touch `RoleFiresHooks`, `serveIssueRoles`, the `Config`/`sourceRoles` sets, the capabilities want-list, and the timed-provider accessor count. Everything conflicts loudly rather than merging silently, so this is a sequencing heads-up, not a defect: whichever stack lands second reconciles the two actor helpers into one (they enforce identical rules; the shapes differ).

## What I verified, and it holds

I spot-checked the load-bearing claims adversarially rather than taking the PR's word:

- **Atomicity is proven where it must be.** The proxied test posts a three-edge batch whose *later* edge closes a cycle through a real `bd serve` against real Dolt, asserts the 409, then **reads the graph back** — zero edges on all three issues via the stored-edge endpoint, cross-checked through `bd dep list`. A handler over a per-edge-committing role would pass every pure test and fail exactly here. The pure tests carry their half: every refusal case asserts the fake role was never called.
- **The hierarchy 409 is rebuildable, both polarities.** The typed error is constructed inside the refusing transaction (`CheckBlockingHierarchyInTx`), the handler maps it by `errors.As` and typed fields only, and `blocker_is_ancestor` travels through a pointer so `false` reaches the wire. Wire tests assert both polarities, assert the three members are *absent* on a plain cycle and on `dependency_exists` (presence is the discriminator), and reject any `detail` that quotes role prose. The role-level conformance contract independently asserts `BlockerIsAncestor`'s value both ways, so the wire members carry everything a client needs — the error struct has exactly those three fields.
- **The endpoint rule is pinned from both sides.** In the role, an `external:` target or a foreign-prefix id is classified external (no existence probe — it lives in `depends_on_external` with no FK); only a same-namespace absence refuses. The proxied test pins all four cases against a real database: locally-absent target → 400 naming `edges[1].depends_on_id` with a zero-write read-back; ghost source → 400 (an edge follows its source); `external:JIRA-9` and `otherrepo-9` → 200 with both edges read back. The 400s map PR #5415's typed refusals; the offending edge is found by both endpoints, and there's a test proving a shared-source decoy does not misname the edge.
- **Idempotent remove genuinely has no 404 path.** No not-found branch in the handler, no `not_found` in the operation's code row (so the parity gates hold the document to it), and the proxied test drives ghost endpoints against real Dolt: 200, `removed: false`, graph unchanged — plus the re-remove round trip.
- **Spec parity and the two deviations.** Both routes are literal `pattern == specPath` rows, so the existing parity tests cover them with no new machinery; the code tables match the document exactly. Naming `DependencyEdge` as a component (rather than inlining it) is right — it is shared by the request and the `added` echo, is wire-identical, and its description heads off confusion with the stored `Dependency` row. Leaving `open_children` out is right — it belongs to closeIssue (#5423 carries it).
- **Cross-cutting posture.** `requireJSONContent` and the shared actor validation on both operations; the hooks/auto-commit disclaimer verbatim on both spec descriptions; `SkipPerEdgeCycleCheck` unpublished with a test asserting it stays off; capability tokens `dependencies.add`/`dependencies.remove` derived and advertised.

CI at review time: 76 checks pass (lint, formatting, conformance oracle, the full historical/upgrade matrices, preflight on all three platforms); the only failures are the one panic above and the gates it takes down.

Fix the stub — ideally with the #5423-style pin so the hook guard for this role is held by a test rather than by a comment — and this is an easy approve. The operations themselves are ready.

julianknutsen:
The blocking item from the review is resolved, with the requested fuller treatment and beyond: the serveRolesStore fake gained the identifiable editor with the RoleFiresHooks precondition and post-peel assertions (now via a stubs(inner) helper so the next role is a one-line addition), the RoleFiresHooks recognition for the dependency editor is pinned by test rather than comment, and the fix cycle caught and closed a second fake with the same nil-embedding defect (serveIdentityStore), sweeping the class: only two cmd/bd fakes can reach serveIssueRoles and both are now covered. The rebase reconciliation onto the merged lifecycle stack matches what the earlier review predicted (one actor helper — main's bodyActor; three-case RoleFiresHooks; the single servedCapabilities literal). 99 checks green, real ci-pr-core run exit 0. Approve. — automated review by Claude (Fable 5), posted by the maintainer's orchestration

PR #5430, opened 2026-08-07
Port the recorded uow handoffs into the contracts, and correct two of their receipts

julianknutsen:
A previous slice (#5406) recorded eight candidates it could not port because a parallel slice owned the target contract files. That constraint is gone. Five ported, one declined as instructed, one out of scope. `+616 / −396`: three new contract cases, two existing ones extended, five ad-hoc tests retired, one file deleted whole.

Every ported case is verified RED against a mutation of the promise it pins, on **each body that leg actually executes** — uow runs its own bodies under `internal/storage/domain/` where the two store backends share others, so most needed two mutations, not one.

## Per handoff

| | outcome | reddens on |
| --- | --- | --- |
| **H1** `IssuePatch.Owner` reaches storage | ported, ad-hoc retired | uow spec drops `setField(owner)`; `issueops.UpdateFields` drops the Owner row |
| **H2** issue_type outside the vocabulary refused on UPDATE | ported, ad-hoc retired | domain `validateIssueTypeUpdate`; `issueops.ValidateScalarUpdates` |
| **H3** same-value write across the whole patch surface | ported, ad-hoc retired | `issueFieldMatches` — shared, so all three legs |
| **H4** a claim is a mutation when the patch restores the row | ported, ad-hoc retired | `s5_m6_claimchanged` at uow; `ExecuteUpdate`'s claim arm at the stores |
| **H5** close provenance survives a persistence move | ported as an extension, ad-hoc retired | `s5_m10_closedsession_move`, all three legs |
| **H6** two static observations, no probe | **declined** — labelled "recorded, not claimed" by its author, and left that way | — |
| **H7** a missing id is `ErrNotFound` | ported as an extension, **both** ad-hoc pins retired | `CheckVersionInTx` at the stores; `operationIssue` at uow |
| **H8** Claimer has no contract | out of scope — landed separately in #5428 | — |

H1 is worth one line of detail: the retired test read the **result issue**; the contract case reads the **raw column**. Reading a value back through the role is exactly the check that passes on a corrupted table.

## Two receipts that did not survive re-verification

Both handoffs carried seeding recipes and mutation evidence from a slice that had been through a three-lens adversarial review. Both were wrong, and neither was sloppiness — each was correct against the body and mutation its author chose.

**H4's recorded recipe was itself a cannot-fail test.** A claim stamps `started_at`, so on a row seeded *without* one the stamp is a field diff and `Changed` is true regardless of whether the property under test holds. The case would have passed either way. Both rows now seed `started_at` and read it back.

It also turned out the two backends need **opposite patches**: the stores claim and then diff against the post-claim row, so the "restoring" patch is a real write there. The case carries both shapes rather than pretending one fits.

**H7's "keep" was wrong.** The handoff called the `calledApply` flag unportable coverage worth keeping. A third mutation — hoisting the version check ahead of the lookup, which is the exact defect those tests are named for — leaves **both ad-hoc tests green** while the new contract case reddens: their requests carry no `ExpectedVersion`, so the branch is never entered. The flag was not covering what it claimed. Both deleted.

The transferable point: **a handoff is a hypothesis with a receipt attached, not a conclusion**, and the receipt is only as good as the body it was taken against.

## Verification

`go build ./...`, `go vet ./...`, `golangci-lint run` clean — the swap from `assertLiveAssignee` to `assertIssueOperationsAssigneeAndStatus` also retired a standing `unparam` finding. Full `TestIssueOperations*` and `TestLifecycleCloseReopenContract` suites green on dolt, uow and embeddeddolt after every edit; the two new lifecycle subtests confirmed to execute at each leg **by name**, since a `-run` that matches nothing exits 0. Helper-reachability pass removed `newRealIssueOperations` and `lifecycleIssueUseCaseStub`. No `.github` file names any deleted test. 14 mutation runs, all in disposable worktrees.


julianknutsen:
## Red-team review — two confirmed coverage losses, both now fixed

An adversarial reviewer attacked the deletions with mutations run in a disposable worktree. It confirmed **every ported case can fail** on the legs claimed, and independently verified the H7 justification. It also found two deletions that lost coverage.

Both are the same shape, and it is a new one for this programme: **the deleted test's FIXTURE carried coverage its assertions never advertised.** A port that faithfully reproduced every assertion still lost the coverage, which is why review caught it and the port author did not.

### Loss 1 — the guarded public-Create scalar surface at uow

The deleted test seeded through **`operations.Create`** with all 17 scalars, so a create body that dropped one surfaced as a `Changed=true` restatement. The port seeds through the raw domain hook and never asserts the seeded values, because the patch overwrites them.

Proof: blanking `issue.Notes` in the uow-only `createParams` failed the deleted test at base and left **all 70** uow `TestIssueOperations*` cases green on the branch. Nothing else at uow drives public `Create` with rich scalars.

**Fixed** by `RunIssueOperationsCreateWritesEveryScalarField`, wired at all three legs. It drives `Operations.Create` with all 17 scalars and reads every column off the raw row. The seed uses `in_progress` rather than `open` and distinct non-zero values throughout, so a field the body never copied lands at a zero the seed cannot be confused with. The seed path was *not* simply swapped onto `Create` — public create refuses a `closed_at` on an open row, so the close-lifecycle end cannot be seeded that way.

### Loss 2 — the claim's public footprint at uow

The deleted test asserted `semanticIssueEqual(before, restored)`, which quietly pinned that a claim's **only** public effect is assignee/status. The port asserts the three fields anyone would think to name.

Proof: making the uow `updateSpec` set `fields["owner"] = request.Actor` when `Claim` is set left all 70 uow cases green — the same-value owner write **no-op-elides**, so the existing assertions structurally cannot see it.

**Fixed** by extending the claim case with the 16 bystander columns, snapshotted **once, before any claim**, and asserted after each of three. The single anchor is what defeats the elision: the second and third claims are genuine no-ops, so only a comparison taken ahead of the first can fail.

That is the inverse of the `Merge≺Set` lesson from earlier in this programme — there, a key removed by a later stage could not witness order; here, a value written identically cannot witness a write.

### What the review cleared

Every new and extended case reddens under its recorded mutation on the exact legs claimed, including both shapes of the claim case and both guard bodies of the issue-type case. The `started_at` fix closed its hole rather than moving it. The H7 justification was independently confirmed: under the hoist mutation both deleted stubs stay green while the contract case reddens — the `calledApply` flag never caught the reorder it was written for, because those requests carry no `ExpectedVersion`.

It also declined to bank a third finding: it suspected H5 narrowed the result plane, chased it, found the coverage held by composition across three other cases, and reported it unproven. That restraint is what makes the two it did confirm credible.

PR #5431, opened 2026-08-08
Wire-surface polish for the v0 write wave: include-leg depth, derive-proof prose, the missing member-list gates, changelog

julianknutsen:
## Summary

Four independent pieces of polish on the v0 HTTP surface, one commit each. The write wave (#5410, #5417, #5422, #5423, #5429) landed the operations; this lands the parts a wave tends to leave behind — the test depth the wire shape actually needs, the orientation prose that went stale under it, the parity gates that were applied to three bodies out of six, and the changelog reconcile.

No wire change. `openapi.v0.yaml` and `types.gen.go` are untouched.

## What each commit does

**1. `test(httpapi)`: assert what the getIssue include legs actually carry**

The include parameters were pinned by presence alone — one comment whose `text` was checked, one dependent whose `id` was checked. `Comment` requires `author` and `created_at`; `IssueWithDependencyMetadata` requires `dependency_type`; and OpenAPI `required` enforces key presence, never value correctness and never array order. So a store answering comments newest-first, or stamping one edge type onto every dependent, emitted schema-valid JSON and passed.

Both are CLI-observable: `bd show` prints each comment's author and time in the order it received them, and groups dependents into sections by edge type.

Both fixtures now carry two rows that differ on every member the schema declares, asserted row by row by index — the index *is* the order promise. The required-member sweep is read from the embedded document rather than listed in the test, so a schema that grows a required member the serve path cannot fill fails here instead of at a client.

**2. `docs(httpapi)`: stop the orientation prose enumerating a surface it no longer has**

`doc.go` still said "the three reads … and the one write, `POST /v0/beads/issues/{id}:claim`" and "the claim OPERATION is the only mutation this surface has". `bd-serve-v0.md` still opened with "Six operations" over a six-row table and sized the server under "Claim throughput — the claim is the only mutation in v0". The route table serves 27 operations, 11 of them writes. `bd-serve-v0.md`'s own header rule says nothing in it may contradict the spec, `doc.go` or the reader role — so the rule was violating itself.

**The counts are removed, not corrected.** `routes.go` is the router, `TestSpecRouteParity` welds it to the spec by exact set equality in both directions, and `capabilities` derives from the same table. A fourth copy spelled in prose can only go stale — as this one did, silently, across five merged PRs. What replaces the design page's table is the surface's *shape*, which does not change when an operation is added.

Statements true only of the claim are widened where the code widened them (the JSON content type is enforced for every body-carrying operation; hooks fire for no write) and narrowed where they were overstated (the reader property covers three named reads, not "all" reads — the count, the query, stats, config, memories and the dependency reads are on sibling roles). Also sweeps the seven stale "one write on this surface" comments and `reads.go`'s "the three read operations" header over a file that now holds five.

**3. `test(httpapi)`: gate the four request member lists the wave left ungated**

The wave shipped six hand-rolled request member lists and gates for two of them. `reopenRequestMembers`, `addDependenciesMembers`, `dependencyEdgeMembers` and `removeDependencyMembers` had none — a tree-wide grep finds them referenced only by their own handlers. That is precisely the drift class `spec_parity_test.go`'s own doc comment calls "the one failure this file exists to prevent".

Nothing else closes the seam: the bijection test covers only `x-go-type`-pinned schemas, `make api-check` ties the struct to the schema by codegen but ties neither to the handler's list, and `reopen_test.go` / `dependency_edit_test.go` never consult the spec.

There is no live bug — all four lists match today. This is the gate that keeps them matching.

**4. `docs(changelog)`: record the v0 write surface under Unreleased**

The wave added 25 lines to `CHANGELOG.md`, all for the role-level typed refusals in #5415, none for the wire. A whole-file grep for every operationId, parameter and code the wave published returned zero hits — so the next cut would have shipped notes announcing the facade programme's twelve operations while omitting the five that turned this API from read-only-plus-claim into something an agent can finish work through. This is the missing reconcile step, written to the section's existing conventions.

## RED evidence

Each new gate was proved to fail against a real defect before being made green.

**Include-leg asserts** — reversing the seeded comments (a store answering newest-first) and stamping one edge type onto every dependent:

```
--- FAIL: TestGetIssueIncludeLegsCarryTheWireShape
    comments[0].id = c-2, want "c-1" — the rows are answered in the role's order
    comments[0].author = bob, want "alice"
    comments[0].created_at = 2026-07-31T12:00:05Z, want "2026-07-31T12:00:00Z"
    ... (9 assertions)
    dependents[1].dependency_type = blocks, want "parent-child" — the edge type is what `bd show` groups this row by
```

**Schema-derived sweep** — adding `thread_id` to the `Comment` schema's `required` list:

```
--- FAIL: TestGetIssueIncludeLegsCarryTheWireShape
    Comment[0] omits `thread_id`, which the document marks required: map[author:alice created_at:... id:c-1 ...]
    Comment[1] omits `thread_id`, which the document marks required: map[author:bob ...]
```

**Member-list gates** — the money shot. A `note` member added to `ReopenIssueRequest`, `AddDependenciesRequest` and `RemoveDependencyRequest` in the spec, **with `make api-gen` run** so the generated structs agree with the document exactly:

```
--- FAIL: TestReopenRequestMembersMatchTheHandler
    generated ReopenIssueRequest declares members the reopen handler refuses as unknown: [note]
    the ReopenIssueRequest schema documents members the reopen handler refuses: [note]
--- FAIL: TestDependencyRequestMembersMatchTheHandlers/AddDependenciesRequest
    generated AddDependenciesRequest declares members the handler refuses as unknown: [note]
    the AddDependenciesRequest schema documents members the handler refuses: [note]
--- FAIL: TestDependencyRequestMembersMatchTheHandlers/RemoveDependencyRequest
    ... same, both directions
```

In that state `make api-check` passes and **every other test in the package passes**. Only the three new subtests fail. That green-everywhere-else run is the failure mode these gates exist for, reproduced.

## One correction worth flagging

Three codes joined the frozen vocabulary in this wave, not four: `not_closable`, `dependency_cycle` and `dependency_exists`. `already_closed` is a required member of the `200 CloseIssueResponse` and never appears in a problem body. The changelog records it as what it is.

## Gates

| Gate | Result |
|---|---|
| `go test ./internal/httpapi/` | ok (all parity tests, including the 3 new ones) |
| `make api-check` | clean — spec and generated types untouched |
| `make ci-pr-lint` (gofmt, golangci-lint ×2) | 0 issues |
| `go test -tags=gms_pure_go ./test/docsync` | ok |
| `./scripts/check-doc-freshness.sh` | PASSED |
| `git diff --check` | clean |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,76 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`bd serve` grows the write half of the agent loop** (v0 wire surface,
+  [#5410](https://github.com/gastownhall/beads/pull/5410),
+  [#5417](https://github.com/gastownhall/beads/pull/5417),
+  [#5422](https://github.com/gastownhall/beads/pull/5422),
+  [#5423](https://github.com/gastownhall/beads/pull/5423),
+  [#5429](https://github.com/gastownhall/beads/pull/5429)). v0 shipped with one
+  mutation, the claim, which meant an HTTP client could take work and had to
+  fork a `bd` subprocess to finish it. Five operations close that loop:
+
+  - `POST /v0/beads/issues/{id}:close` and `POST /v0/beads/issues/{id}:reopen`
+    — the two halves of the lifecycle move. Both are idempotent, and both say
+    so in the body rather than in a status: a close of a closed issue answers
+    `200` with `already_closed: true`, a reopen of an open one
+    `already_open: true`, and each still carries the row.
+  - `PATCH /v0/beads/issues/{id}` — partial update of one issue. The `patch`
+    object publishes thirteen members; four of them (`estimated_minutes`,
+    `external_ref`, `due_at`, `defer_until`) are nullable, and an explicit
+    `null` CLEARS them. That set is closed and machine-checked against the
+    document, so a `null` on any other member is a `400` rather than an
+    unannounced clear. A patch that changes nothing answers `200` with
+    `changed: false`.
+  - `POST /v0/beads/dependencies:add` — up to 100 edges per request, asserted
+    all or none, echoed back in request order. A refusal names the offending
+    edge as `edges[i].member`, so a client learns which edge and which member
+    rather than that something in the batch was wrong.
+  - `POST /v0/beads/dependencies:remove` — removes exactly the edge the body
+    names. Idempotent at the role: an edge that is not there is
+    `removed: false` with a `200`, never a 404, so a replayed teardown does not
+    have to classify an error to discover it already ran.
+
+  **Every one of them carries the claim's posture, unchanged.** The `actor` is
+  caller-asserted provenance for the audit trail and not authenticated
+  identity; hooks do not fire; the per-command auto-commit machinery does not
+  run. The only durable effect is the single storage commit the role makes
+  inside its own transaction, so durability is per request on every write, not
+  just the claim.
+
+  **Three codes join the frozen error vocabulary**, all 409 and all typed so a
+  client never has to substring-match a message. `not_closable` is close policy
+  refusing an unforced close of an issue with open children or a live blocker.
+  `dependency_exists` is the pair that already carries an edge of a DIFFERENT
+  type, and carries `existing_type` and `requested_type`. `dependency_cycle`
+  covers both never-makes-progress refusals: the plain scheduling cycle, and a
+  blocking edge against the issue's own ancestor or descendant — the second
+  carries `issue_id`, `blocker_id` and `blocker_is_ancestor`, and their ABSENCE
+  is what tells a client which of the two it got. Every extension member is
+  read off the refusing transaction's own typed error; the hierarchy members
+  have no other source, since the conflicting edge may exist only inside the
+  batch that was rolled back. An edge naming an endpoint this database can see
+  the absence of is a `400` on `edges[i].issue_id` or `edges[i].depends_on_id`,
+  not a 404: the request BODY is what is wrong, and there is no path id to have
+  missed.
+
+  **`GET /v0/beads/issues/{id}` takes two new parameters.**
+  `include_comments` populates `comments` with the full comment bodies and
+  `include_dependents` populates `dependents` with the issues that depend on
+  this one, each carrying its edge type. Both default to false, because both
+  are extra reads nobody should pay for by accident — and when a caller does
+  not ask, `comment_count`, `dependent_count` and `comments_omitted` still
+  report what was left out, so an absent `comments` is never ambiguous between
+  "no comments" and "not included in this response". A malformed boolean is
+  this operation's own `400 invalid_argument` / `reason: "invalid_value"`,
+  which is deliberately not the `unknown_parameter` a client should read as
+  version skew.
+
+  `GET /v0/beads/context` advertises `issues.close`, `issues.reopen`,
+  `issues.update`, `dependencies.add` and `dependencies.remove` alongside the
+  rest — derived from the registered handlers as always, so probing
+  `capabilities` remains the way to ask one server what it can do.
+
 - **The two dependency-endpoint refusals now carry an identity** (bd-yby99.9).
   `DependencyEditor.AddDependencies` refuses an edge whose SOURCE names no row,
   and one whose TARGET names no row this database can see the absence of. Both
--- a/engdocs/design/bd-serve-v0.md
+++ b/engdocs/design/bd-serve-v0.md
@@ -23,16 +23,35 @@ so.
 
 ## The surface
 
-Six operations, all under `/v0` except liveness.
+Everything is under `/v0` except liveness. The operation list is **not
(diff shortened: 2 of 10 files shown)


julianknutsen:
## Verdict: Approve

(pending the ~10 still-running CI checks, none of which has failed; everything below was verified read-only against the branch at 63e77111f)

This is a tests-plus-prose-plus-changelog tranche with **zero executable change** — I verified that claim mechanically, not just from the file list: of the ten files touched, five are test files, two are docs pages, one is the changelog, and the two non-test Go files (`internal/httpapi/claim.go`, `internal/httpapi/reads.go`) contain comment-only hunks. `openapi.v0.yaml` and `types.gen.go` are untouched in the final diff.

I adversarially spot-checked the four headline claims rather than taking the (very persuasive) PR body at its word.

### 1. Include-leg depth — verified, with one overstated sentence

The derivation is real: `requireDocumentedMembers` parses the **embedded** spec (`spec.OpenAPIV0()` via `loadSpec`) and sweeps `components.schemas.{Comment,IssueWithDependencyMetadata}.required` over every row, so a schema growing a required member the serve path can't fill fails this test with no edit. It also refuses to pass vacuously (fatal on an empty required list or zero rows). The previously fabricated-but-unread fixture values are now load-bearing: at the base commit, `Author: "alice"`, the comment text, and `DepBlocks` were referenced by no assertion anywhere in the tree; at tip, all five `Comment` members are asserted per row by index, and each dependent's `id` and `dependency_type` likewise — the index-based assertions are the order promise, which OpenAPI `required` cannot express.

One inaccuracy worth fixing in a follow-up: the fixture comment (and PR body) says the two rows "differ on every member the wire schema says a row has." True for comments (except `issue_id`, which can't differ), **not** true for dependents — both rows come from `seededIssue`, so `title`, `priority`, `created_at` and `updated_at` are identical; only `id` and `dependency_type` differ. The assertions are unaffected (nothing value-asserts the shared members), but either differentiate the two dependent issues or scope the sentence to the comments list.

### 2. Derive-proof prose — verified in full

The counts are genuinely removed, not corrected: the design page's six-row table is replaced by a pointer at the mechanism (`routes.go` + `TestSpecRouteParity`'s two-way set equality + derived `capabilities`) plus a shape description that doesn't change when an operation lands. I counted the route table at tip: 27 operations, 11 writes — the deleted "six operations / one write" prose was indeed wrong, silently, across the wave. The flagged overstatement fix is accurate: `handleReady`/`handleListIssues`/`handleGetIssue` reach `issueops.Reader` (`Ready`/`List`/`Get`); the count rides `ReadyCounter.CountReady` and the query rides `Querier.Query`, so naming the three and disclaiming the siblings is the correct statement. The widened content-type claim also checks out — `requireJSONContent` is called by every body-carrying handler, all ten of them. A tip-wide grep finds no residual "one write on this surface" / "only mutation" / stale-count phrasing.

### 3. Member-list gates — verified; they do catch the silent-drift scenario

At the base commit, the four lists (`reopenRequestMembers`, `addDependenciesMembers`, `dependencyEdgeMembers`, `removeDependencyMembers`) were referenced only by their own handlers' `unknownMember` refusal sites — the gate compares the **same variables** the refusal path consults, against both the generated struct (via the embedded-struct-aware `jsonTagNames`) and the parsed spec schema (via the allOf/`$ref`-aware `schemaProperties`), in both directions. That means the exact drift the PR reproduced red — spec grows an optional member, `api-gen` run so codegen agrees, every other gate green — fails both the struct-vs-list and spec-vs-list directions here. The shapes mirror the precedent exactly: reopen follows the close gate, the dependency pair follows the update gate's table-driven form, and `DependencyEdge` is gated at its own nested level as `IssuePatchBody` is. All four schemas carry `additionalProperties: false`, so the lists are what make that clause enforceable. The final diff leaves the spec and generated types untouched; the injection was fully reverted.

### 4. Changelog — verified against the document, including the correction

The spec confirms the correction the PR flags: `not_closable`, `dependency_cycle`, `dependency_exists` appear in `x-bd-codes` (all 409); `already_closed` is a required member of the 200 `CloseIssueResponse` (`required: [issue, already_closed, open_children]`, mirrored by `already_open` on the reopen) and is recorded as such, not as a code. Five operations, two `getIssue` parameters, the 13-member patch with its closed 4-member nullable set, the 100-edge cap, and the advertised capability tokens all match the code and spec at tip. The entry follows the section's conventions (bold lead-in, PR-link citations — a style the file already uses ~200 times, prose depth matching its neighbors).

### Process notes

- The RED transcripts describe branch states that can't be replayed read-only; I verified the mechanisms would fail exactly as transcribed instead of trusting the transcript.
- Branch is based on the current tip of `main`, so the CHANGELOG hunk applies cleanly.
- OSS purity: no internal references, hostnames, or credentials in the diff; the module path is the repo's own.
- House style: the comment density would violate the contributor guideline in any other package, but `internal/httpapi` is deliberately written this way and the PR matches its surroundings.

Nothing blocking. Let the remaining CI checks finish (none has failed; the diff has no executable change, so the exposure is nil), and consider the dependents-fixture comment fix as a follow-up nit.

PR #5432, opened 2026-08-08
test(conformance): an accessor-only Update contract, and the claim→close loop

julianknutsen:
Two gaps in the Lifecycle proof surface, both of which a divergent backend passes the whole merged conformance suite while breaking.

## 1. `Lifecycle.Update` has no accessor-only contract

All 16 `RunIssueOperationsUpdate*` runners take `IssueOperationsStagingFixture`, whose `Exec`/`QueryScalar`/`UpdateRaw`/`Commit` fields are a version-control-aware SQL seam, and whose single `Operations` field is *both* the seed route and the subject — several cases reach their starting state through `fixture.Operations.Create`. A backend that cannot supply raw SQL, or whose composition refuses `Create`, therefore runs **zero** Update cases.

The suite already concedes this. `issue_claimer_contract.go`'s header says outright that "the Update-seam cases run only behind `IssueOperationsStagingFixture`, which demands raw-SQL hooks a partial or remote backend cannot supply" — and closed the gap for `Claimer` alone. This closes it for `Update`.

`backend/conformance/lifecycle_update_contract.go` binds the accessor plus `CreateIssue`/`CreateWisp`/`AddDependency` seed hooks and one out-of-band `GetIssue`, separating subject from seeding the way `LifecycleCloseReopenFixture` and `ClaimerFixture` already do. Nothing in it reaches for raw SQL: `UpdateResult` and `GetIssue` between them expose every fact the cases assert, including the `RowVersion` token that tells a refusal apart from a silent write.

Eight owning cases, each exercising only patch members a partial backend can express, so it adopts them unmodified:

| Case | Pins |
|---|---|
| `PersistsThePatchAndHydratesTheResult` | every content member reaches the row; the result carries labels and dependency records and omits comments |
| `ReportsNoChangeForASameValuePatch` | `Changed:false`, and `RowVersion` proving nothing was written |
| `AppendsNotesWithoutReplacingThem` | append keeps what was there; `Notes` replaces; both together is `ErrValidation` |
| `ClearsTheNullableMembers` | the four `Field[*T]` clears, and the re-clear no-op |
| `ReplacesTheLabelSet` | `Replace` is a set assignment; an empty `Replace` clears rather than reading as omitted |
| `ResolvesBothPlanesUnlessRestricted` | both-plane auto-resolve, `IssuePlaneOnly` refusal, durable control |
| `RefusesUnknownIDsAndActorlessRequests` | `ErrNotFound`, `ErrValidation`, state unchanged |
| `RefusalWritesNoMemberOfThePatch` | one invalid member refuses the whole patch atomically |

Only the plane runner overlaps an existing case, and the header names the overlap rather than hiding it. `Metadata`, `Status` and close policy, `Assignee` and the transfer fence, `Claim`, `ParentID`, `Persistence`, the `Expected*` preconditions and `Provenance` stay with the staging cases that own them — pinning them here would make the block unadoptable for the backends it exists to serve.

Three wirings, two votes, stated at the top of the file per `engdocs/ADDING_AN_ISSUEOPS_ROLE.md`: the two stores share `ExecuteUpdate`; the unit of work maps the patch itself and derives `Changed` by comparing post-state, which is exactly where the same-value and clear cases have teeth.

## 2. The claim→close loop is unpinned

No case anywhere closes a `StatusInProgress` row, and no close/reopen case seeds or asserts an assignee. Every close target in the package — here, in `claim.go`, in the audit files, in `seedClosePolicyIssue` and in `batch_closer_contract.go` — is seeded open and unassigned. So the only close a working agent performs is the one nothing pins.

Two divergences pass the whole suite today:

- a backend whose close admits open rows only refuses every claimed issue, so `bd claim` and `bd close` stop composing;
- a backend that clears `assignee` as part of the close — a defensible reading of "the work is over" — silently drops the attribution `bd show` renders and `bd list --assignee` filters on, and a reopen loses it again.

`RunLifecycleCloseAndReopenKeepTheClaimHolder` seeds the state a won claim leaves, closes it, asserts the holder on both `CloseResult.Issue` and the stored row, then reopens and asserts it again alongside the close attribution the reopen *is* supposed to clear.

`lifecycleCloseReopenRow` gains `assignee` so the file's existing refusal and no-op legs pin holder stability too, and the reopen no-op's `in_progress` leg now carries a holder — it compared an empty column before, so a clear on that path was invisible.

## Load-bearing evidence

Every case is a green-first pin, so each was proved load-bearing by mutating production and checking it goes red. 18 mutations, 17 caught:

| Mutation | Case |
|---|---|
| `UpdateFields` drops the `design` member | PersistsThePatch… RED |
| store hydration drops the dependency records | PersistsThePatch… RED |
| unit-of-work hydration answers the bare row | PersistsThePatch… RED |
| unit-of-work hydration drops the dependency records | PersistsThePatch… RED |
| `Changed` always true (store, then unit of work) | ReportsNoChange… RED ×2 |
| `append_notes` replaces instead of appending | AppendsNotes… RED |
| `Notes`+`AppendNotes` stops being a refusal | AppendsNotes… RED |
| a set-but-nil `external_ref` read as omitted (store, then unit of work) | ClearsTheNullableMembers RED ×2 |
| `LabelPatch.Replace` merges instead of replacing | ReplacesTheLabelSet RED |
| `IssuePlaneOnly` ignored | ResolvesBothPlanes… RED |
| a missing actor defaulted instead of refused | RefusesUnknownIDs… RED |
| the out-of-range priority dropped instead of refused | RefusalWritesNoMember… RED |
| close clears the assignee | CloseAndReopenKeepTheClaimHolder RED |
| reopen clears the assignee | CloseAndReopenKeepTheClaimHolder RED |
| close refuses an `in_progress` target | CloseAndReopenKeepTheClaimHolder RED |

The one survivor: dropping `issue.Labels = labels` from the store's `HydrateIssueOperationResult` stays green, because `GetIssueInTx` already loads labels on that path. The labels half is load-bearing on the unit-of-work leg, where the mutation does go red.

## Tests

- `TestLifecycleUpdateContract` and `TestLifecycleCloseReopenContract` green on all three wirings — dolt, embeddeddolt, uow (105 subtests).
- `go test ./backend/... ./internal/storage/issueops/... ./issueops/...` green.
- `scripts/ci/pr-lint.sh` clean (gofmt, golangci-lint, golangci-lint windows).
- `internal/storage/dolt` as a whole package times out on this machine under `BEADS_TEST_ENV_RUN_DOLT=1`, failing six `TestCrossProject_*` tests with `failed to initialize schema: context deadline exceeded`. That reproduces **identically at `upstream/main` with no changes applied** — same six tests, same 25m package deadline — so it is a pre-existing capacity issue here, not this branch. `-run '^TestCrossProject'` in isolation passes on both.

Test-only change; no production files touched.

### DIFF
--- a/backend/conformance/lifecycle_close_reopen_contract.go
+++ b/backend/conformance/lifecycle_close_reopen_contract.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"testing"
+	"time"
 
 	"github.com/steveyegge/beads/internal/types"
 	publicops "github.com/steveyegge/beads/issueops"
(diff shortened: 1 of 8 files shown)


julianknutsen:
## Review: accessor-only Update contract + claim→close loop

**Verdict: approve** (please let the in-flight CI finish before merge — the "Test (storage domain + uow)" job is the one that actually executes the new uow wiring; everything completed so far is green).

This closes the gap it claims to close, and the claims survived adversarial checking. Before this PR, every Update-owning conformance case ran behind `IssueOperationsStagingFixture`, whose `Exec`/`QueryScalar`/`UpdateRaw`/`Commit` hooks are a raw-SQL seam and whose single `Operations` field is both seed route and subject — so a partial or remote backend served `updateIssue` with zero contract coverage. The suite itself concedes this in `issue_claimer_contract.go`'s header (I verified the quote is verbatim), and this PR does for `Update` what that file did for `Claimer`.

### What I verified, not just read

**The fixture separation is real.** `LifecycleUpdateFixture` carries no raw-SQL hook of any kind. The eight runners invoke only `fixture.Lifecycle.Update` on the subject; all seeding goes through `CreateIssue`/`CreateWisp`/`AddDependency`, which every wiring fills with direct store or domain-use-case writes — never `Lifecycle.Create`. A backend whose composition refuses Create can wire every runner. A nil `CreateWisp` produces a visible `t.Skip` with a stated reason rather than a quiet pass on the durable leg.

**The scope claim is exact.** The wire's `IssuePatchBody` is title, description, design, acceptance_criteria, notes, append_notes, priority (0-4), issue_type, labels as complete replacement, and exactly four nullable members: estimated_minutes, external_ref, due_at, defer_until — which are also exactly the four `Field[*T]` members of `IssuePatch`. The runners exercise precisely that set and nothing the wire excludes: no Status, Assignee, Metadata, ParentID, Claim, Persistence, or Expected* precondition appears in any request the new cases build. Those stay with the staging cases that own them, as the header says.

**The three wirings are what they claim.** All three test files enumerate the same eight runners, following `lifecycle_close_reopen_contract_test.go`'s pattern. The uow wiring is genuinely the independent second body: it reaches the role through the `IssueLifecycleSource` accessor, maps the patch itself in `updateSpec`, applies through the domain `ApplyUpdate` use case, and derives `Changed` by post-state comparison (`semanticIssueEqual` with `RowVersion`/`UpdatedAt` masked) — which is exactly where the same-value and clear cases have teeth against the stores' row-write-facts derivation. The "two votes, not three" statement required by `engdocs/ADDING_AN_ISSUEOPS_ROLE.md` is present at the top of the contract file. One scoping note: the refusal cases (priority range, notes exclusivity) flow through the one shared `ValidateUpdateRequest`, so they are nearer one vote than two; the header's vote claim is scoped to the semantics where it is accurate.

**The claim→close loop case is the missing pin it claims to be.** Nothing else in the package closes a `StatusInProgress` row or asserts an assignee across close/reopen. The new case seeds the exact state a won claim leaves (in_progress, holder, started_at), fatals if the seed didn't take, closes with the holder as actor, asserts the assignee on both `CloseResult.Issue` and the stored row, then reopens and asserts the holder again while confirming close_reason/closed_by_session are cleared. The sharpened `lifecycleCloseReopenRow` gains `Assignee`, and because the file's row assertion is a whole-struct comparison, every existing refusal and no-op leg now pins the column too; the reopen no-op's in_progress leg genuinely seeds a holder where it previously compared an empty column.

**The mutation evidence holds up, including the confession.** The one reported survivor — dropping `issue.Labels = labels` from the store's `HydrateIssueOperationResult` — is structurally guaranteed to survive: `GetIssueInTx` already loads labels inside the same transaction, so the dropped line is redundant on that leg. The uow leg has no such redundancy (its hydration helper is the only label-attach point), and the uow wiring's `GetIssue` readback deliberately reads labels off the label use case rather than through the hydration code under test, so the equivalent uow mutation cannot blind its own readback. The Changed-always-true mutation is caught twice because the two legs compute `Changed` by different mechanisms, and the dependency-record mutations are caught because store hydration is the sole attach point for `Dependencies` and the hydration case asserts a specific edge.

**The deviations are grounded.** The decision not to touch the older `RunAll` suite follows the explicit prohibition in `engdocs/ADDING_AN_ISSUEOPS_ROLE.md` ("a case placed there silently never runs on the backend most likely to diverge"), and the out-of-band `GetIssue` hook follows the same doc's sanctioned pattern for wiring-site closures. The out-of-range-priority refusal is the right choice: priority validation wraps `ErrValidation` deterministically on every leg, whereas an overlong title never produces `ErrFieldTooLong` (that sentinel covers assignee/owner/label lengths; the title path emits a plain size error), so the alternative would have pinned a sentinel the implementation never emits. Seeding the claimed state directly rather than through the Claimer role is documented on the helper with a sound rationale.

### Minor, non-blocking

- Two header citations have drifted a few lines at this PR's own base: the `IssuePlaneOnly` doc is at ~269-278 (cited 251-260) and the atomic-update clause at ~412-419 (cited 415-416). All other citations are exact. Fine as a follow-up fixup.
- `AddDependency`-nil silently narrows the hydration case to its labels half where `CreateWisp`-nil skips loudly. The asymmetry is documented and has precedent, but it is the least-visible coverage reduction in the file.

Test-only change, no production files touched, no out-of-project references. This is careful, honest work — the kind of coverage PR where the evidence table can actually be audited, and it checks out.

PR #5435, opened 2026-08-08
Converge Offset and MaxRows on all three legs, and fix the panic under them

julianknutsen:
`Offset` and `MaxRows` were mirror-image splits: the unit-of-work seam honored `Offset` and refused `MaxRows`; the two store legs did the opposite. Three contract cases absorbed that with an "either HONORED or REFUSED with a typed `*ErrUnsupported`" disjunction.

**A contract that asserts a capability is either honored or refused is not pinning semantics — it is ratifying a split.** All three are gone, along with `assertReaderUnsupported`.

Converging *upward* (both legs honor both) rather than uow-matches-server is the owner's one ratified exception here, because the server refusing `MaxRows` was a capability gap rather than a semantics choice.

## Before and after

| | Offset before | MaxRows before | after |
| --- | --- | --- | --- |
| `storereader` (dolt + embedded) | refused, typed `*ErrUnsupported` | honored | **both honored** |
| `storequerier` (dolt + embedded) | refused | n/a | **honored**, both expression shapes |
| `uow` reader | honored (SQL `OFFSET`) | refused | **both honored** |
| `uow` querier | honored (SQL `OFFSET`) | n/a | unchanged; the panic under it is gone |

`Offset` now runs in the **shared page epilogue** on every leg rather than in whichever seam could render it. Two reasons, both load-bearing: the skip must happen in the caller's display order (`--sort id` is not SQL-expressible, which is why `SQLLimit` already sends it unbounded — the uow push-down got that case wrong), and a row the database skips is a row the query matched, so a SQL `OFFSET` lets an offset talk a caller out of the `MaxRows` breaker.

## The panic, fixed at the root

`domain/db/limitOffsetSQL` rendered `LIMIT 18446744073709551615 OFFSET k` for an unbounded query (MySQL has no OFFSET without a LIMIT). Dolt sizes a buffer from limit+offset and returns a recovered `makeslice: cap out of range` instead of rows.

The contract case had been shaped around it — every request carried a bounded limit, with a comment calling it "a storage-seam bug that predates this role". It now drives the unbounded shape, because the bug is gone: a `searchWindow` value decides the SQL clause and the Go-side residue together, and with no LIMIT to hang an OFFSET on the skip happens in Go, which costs nothing since an unbounded query reads every matching row anyway.

## A seventh cannot-fail test, caught mid-slice

`searchWindowFor` had a second Go-skip branch "under a defensive cap" that was **unreachable**, and it absorbed the exact regression the new MaxRows-behind-offset arm exists to catch — the mutation came back `INCONCLUSIVE`. Removed; the arm is now uniquely falsified.

## A regression the goldens did not catch

Taking `Offset` off the builders broke proxied `bd ready`, `bd list --watch` and the proxied `--parent` walk, which consume the filter as a value and have no page epilogue. Caught by two `TestProxiedServerReady` cases. The builders keep writing it and `WithRowsBeforeThePage` clears it, so the move is one call rather than a reach-in.

## Verification

Twelve mutations, each attributed to the body it broke and re-run at final HEAD: `storereader` skip → dolt+emb RED; `uow/issue_reader` skip → uow RED with dolt green; the same two on the **Ready arm only**; `storequerier` skip → dolt+emb RED; `uow/querier` push-down → uow RED; `workapi.SkipRows` → uow predicate arm RED; the unbounded-skip panic fix → uow RED; caps on both seams → their own legs RED; and `uow` pushing offset down → the MaxRows-behind-offset arm RED while the plain Offset case stays green.

`go build ./...`, `go vet`, `golangci-lint run` clean. Reader and querier contracts green on all three legs; full `TestConformance` green on both stores; `TestProxiedServerList/Ready/Query` green.

## Left alone, deliberately

`bd list/ready/query --offset` still refuses on the direct CLI route. It is stale for the page path now, but `--watch` and `--parent --pretty` consume the filter as a value, so lifting it is a CLI-surface decision rather than this slice's. `--max-rows` was in scope and is retired.

### DIFF
--- a/backend/conformance/querier_contract.go
+++ b/backend/conformance/querier_contract.go
@@ -20,9 +20,10 @@ import (
 //
 // TWO BODIES BEHIND THREE WIRINGS: dolt and embeddeddolt share
 // internal/workapi/storequerier, and the unit-of-work provider is the second,
-// genuinely separate vote. They diverge only where the uow seam renders OFFSET
-// and reports has-more natively, which is why the Offset case is written
-// comparatively.
+// genuinely separate vote. What is left of their difference is mechanism, not
+// answer: the uow seam renders OFFSET and reports has-more natively, the other
+// reaches past the skipped rows and lets an over-fetched row speak. Every case
+// here is written to the answer, which is why none of them names a backend.
 //
 // WHAT THESE CASES DO NOT PROVE: the max(3*Limit, 100) over-fetch window both
 // front doors used to apply, which takes MORE THAN A HUNDRED candidate rows to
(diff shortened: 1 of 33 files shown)


julianknutsen:
## Red-team review — one real bug, one correct implementation misread, both now pinned

The reviewer's brief was weighted toward production risk rather than test quality: this changes read paths on every backend, and `bd list`/`bd ready`/`bd query` are the most-used commands here.

### Finding 1 — a real bug: same flags, opposite breaker verdicts

Proxied `bd list --watch` with `--offset` **and** `--max-rows` counted the cap against **post-skip** rows while the page route counted pre-skip. Empirically: L=10, k=2, M=3 over 5 matching rows → the watch route delivers a page; the page route fires `*ErrTooManyRows` on the identical request.

`runListProxiedWatch` consumes the builder filter as a *value* and has no page epilogue, so `domain/db` rendered `LIMIT bound OFFSET k` and `EnforceMaxRowsCap` only saw survivors. Newly reachable — `3b2385922` removed the proxied `--max-rows` rejection that sat ahead of the watch dispatch. It also falsified the comment at `issue_search.go:537` ("NO CALLER COMBINES AN OFFSET WITH A CAP" — the watch and `--watch --ready` arms do).

**Fixed**: with a cap and an offset, `searchWindowFor` keeps the skip in Go and sizes the bound over `limit+offset`, so every route counts the touched window. Comment corrected.

### Finding 2 — the boundary was already right; the doc was wrong

The review reported that at `Limit+Offset == MaxRows` the breaker can never fire, and called it an off-by-one. Measured under the owner's ratified semantics, **that behaviour is correct**: `L=2, k=1, M=3` touches a window of 3, and 3 > 3 is false. `WithRowsBeforeThePage` widening before the probe-bump is exactly what makes the cap count `Limit+Offset`.

The review read a correct implementation as broken because `issueops/reader.go` documented the *other* semantics ("a cap looser than Limit never fires"). So the deliverable was the doc, plus pinning the behaviour so nobody "corrects" it back.

### MaxRows semantics, ratified by the owner

`MaxRows` counts **the rows the query touched (offset+limit)**, not the rows the caller receives. It is a breaker against unbounded result sets: a skipped row is still a matched row, and an offset must never walk a caller past it.

### Where the boundary is now pinned

- `RunReaderListMaxRowsBoundaryIsLimitPlusOffset` — L+k at M−1/M/M+1, limit and cap held still while the offset moves, at all three legs. **Five** rows against a cap of three, so every window has a row behind it — three rows would let the non-firing arms pass on a body with no cap at all. Both sides of the `--ready` arm, which is a different query in both seams.
- `TestSearchWindowForCountsTheOffsetIntoTheCap` — the window table including no-cap and unbounded shapes; the only pin for `--watch --ready`.
- `TestProxiedServerListWatchCountsTheOffsetIntoMaxRows` — both directions of the watch route, compared against the page route's verdict read at runtime, and verified by reverting the fix (the firing half then times out at 90s).

**Honest limit, stated rather than papered over**: the reader contract *cannot* see the watch route, because neither implementation ever hands a seam an offset. That is why it is pinned by a `domain/db` unit test and a `cmd/bd` test instead of a contract case.

### What the review cleared

Offset SQL→epilogue equivalence held up under scrutiny: SQL `ORDER BY` always carries an `id ASC` tiebreak, `SortRows` is stable, and `CompareIssuesBy` is a coarsening of SQL's key for every sort — so a stable re-sort of the SQL-ordered superset is the identity, including NULL placement for `closed`/`assignee`. `HasMore` is computed against the widened window on both legs, as before. The panic fix is verified, and bounded SQL is byte-identical to main when `MaxRows=0`.

One narrow exception it could not rule out and neither can I: under `--sort title`, SQL `LOWER()`+collation and Go `ToLower` byte-compare can disagree on non-ASCII, and then old and new pick different skip sets deterministically. Conformance seeds ASCII, so it is uncovered.

The consumer sweep came back complete — every reader of the filter's `Offset` accounted for, with one pre-existing gap noted (`--watch --parent` silently ignores an offset, on main today).

### Correction to this PR's own description

It claimed "net −810 lines". That was a two-dot diff against a base that had moved, counting other people's merged deletions as this branch's. The real change is **+703/−421** at the time of that claim. Corrected in the body.

PR #5437, opened 2026-08-08
fix(uow): fire script hooks on the unit-of-work plumbing (bd-opisf)

julianknutsen:
Split out of #4916 at its reviewer's request ("split the hooks rider"): the unit-of-work plumbing — everything proxied-server mode runs — fired **no script hooks at all**, so integrations wired to `.beads/hooks/` silently missed every mutation that went through it. Four commands papered over this with ad-hoc synchronous shims (`update`/`close`/`reopen`/`gate`) that had their own bugs (they ignored `no-hooks`) and their own private semantics.

## What this does

`uow.NewNotifyingProvider` wraps the provider once, at the seam every capability role passes through: events are buffered during the transaction and drained strictly **after** commit succeeds, fire-and-forget; a rollback fires nothing; with no sinks configured the provider is returned unwrapped (zero overhead). The four shims are deleted — one firing site per plumbing. Payloads are hydrated to match the embedded chain: labels always, dependency records on edge-bearing events. Creates and graph applies also notify every issue whose dependency graph they rewired — including pre-existing issues on the far end of reverse edges, resolved across both planes.

## Behavior changes (also in CHANGELOG)

Six visible in proxied mode: hooks now fire for **every** uow mutation (create/claim/labels/comments/deps/batch ops — previously silent); firing is post-commit and asynchronous (the shims were synchronous); `bd update -s closed` fires `on_update` only (`bd close` is what fires `on_close`, both plumbings); the pre-close `on_update` shim event is gone; `no-hooks` is honored (the shims ignored it); idempotent re-close and double `bd gate resolve` fire `on_close` (embedded parity). A seventh applies to both plumbings: every command now waits (bounded by the 10s per-hook timeout) for its in-flight hooks at exit — including error exits, where cobra skips `PersistentPostRunE` and `main()` waits instead.

## Deliberate divergences from the embedded chain

Numbered and justified in the `notifying.go` header: per-distinct-source edge-update multiplicity (embedded fires per edge); no synthetic cumulative label updates on create (the create payload carries the labels); `bd import` fires no hooks on this plumbing (raw-statement path — enforced by an AST guard, not just prose); precondition-only `ApplyUpdate` fires nothing; proxied `bd promote` fires `on_update` via the comment verb where embedded's legacy `AddComment` is undecorated.

## Guards

Three structural tests make silent hook loss a build failure rather than a code-review hope: a reflection **partition** over the provider surface (every method recorded XOR exempted-with-reason, staleness-checked, embedded-promotion aware); an AST guard that parses the whole package so a new role accessor or a delegated constructor (`NewCommenter(p.inner)`) fails; and a bypass guard accounting for every `unwrapUOW` peel. Each was mutation-tested against the regression it exists to catch.

## Testing

21 uow test functions (48 subtests) including six against a real Dolt provider; `-race` clean; proxied-mode CLI integration tests updated to the new contract (polling markers, partial-batch-failure hook delivery); `bd serve` refuses hook-firing providers (uow twin of the existing httpapi role refusal).

## Provenance

Implementation by a Claude Opus agent; three adversarially-verified multi-agent review rounds (32 reviewer/verifier agents total: 9 confirmed findings fixed across two waves, then 8/8 independently closure-verified plus one final find, also fixed). Tracking: bd-opisf.


steveyegge:
## Review verdict: MERGE after an update-branch re-run

Adversarially reviewed in depth. The design is right and the engineering quality is unusually high — buffering at the one seam every proxied capability role passes through, in-tx snapshot re-reads (labels always, dep records on edge events) with deep clones killing the bd-9wgv3 shared-pointer class by construction, drain strictly after successful Commit with Close-discards-on-rollback, and the three structural guards (reflection partition with staleness-checked exemptions, the AST bypass scan, the importer unwrap) all do what the body claims — I verified each. The hunt came back clean on every scenario that matters: double-firing is impossible (proxied mode returns out of PersistentPreRunE before the embedded decorators are built, and all four shims are gone — zero remaining references), rolled-back and failed-commit units fire nothing (tested), guarded-close fires on_close alone via the specWrites filter, serve unwraps with a refusal test, and the retry loop's fresh-UOW-per-attempt makes only the winner fire.

**Why not merge on the current green:** CI ran on a merge snapshot ~10 commits behind main, which has since taken #5435/#5436 — both touching uow read paths and issueops interfaces. I diffed the range: no new mutating verbs or provider accessors, so the partition guards should stay green — but those guards failing loudly on the true merge result is exactly what they're for, so let's see it. I'm updating the branch now; merge follows on green.

Two follow-ups to file (neither blocks; say if you'd rather I file them):
1. **Error-path hook wait runs with the embedded store still open** — on a RunE error cobra skips PersistentPostRunE, so main()'s waitForCommandHooks runs before store close, and an embedded partial-batch error exit gives the fired hook its window while the Dolt lock is held (its child `bd` can't open the workspace). Strictly better than the pre-PR killed-goroutine, and proxied — this PR's subject — is unaffected; still worth closing.
2. **No test pins single-firing across a serialization retry** — correct today by fresh-UOW-per-attempt construction, but a refactor reusing a UOW across attempts would double-fire with no red test. One unit test simulating a losing attempt would pin it.

One subtlety worth a code comment when you're next in there: the nothing-to-commit success path (tx.go:113) reports success without draining — defensible (nothing landed), but the success-with-silence shape deserves its one line.

PR #5438, opened 2026-08-08
fix(validation): canonicalize actor/assignee separators before comparing (ga-wzl83)

athosmartins:
## What

`AssigneeMatches` (`bd close`'s ownership guard) and `AssigneeNotStolen` (`bd update -a` / `bd assign`'s reassignment fence) compared `actor` to `issue.Assignee` verbatim. Add `canonicalActor`/`actorMatches`: collapse any run of `.`, `_`, `-` to one canonical separator before comparing, so two spellings of the same identity compare equal while genuinely different identities stay distinct.

## Why

The same Gas Town identity reaches `bd` spelled differently depending on which layer produced the string it was handed — a dotted alias like `gastown.mayor` gets its dot replaced wherever a dot is unsafe for that context (`__` in a session name, `_` in a Dolt table/database name, `-` elsewhere) — and `bd` only ever sees the resulting string, never the substitution itself.

Concretely (ga-wzl83): an agent with `GC_AGENT=gastown.mayor` tries to close a bead assigned to itself and gets refused:

```
cannot close BEAD-ID: assignee is "gastown.mayor", actor is "gastown__mayor"; reclaim or use --force to override
```

Both strings name the same principal. The error message teaches the agent that `--force` is the normal path for closing its own bead — which is exactly backwards, and erodes the guard's actual purpose (be-035: refusing a close from an actor with no authority over the bead) for anyone who internalizes "closing my own bead needs --force".

`AssigneeNotStolen` has the identical pattern (its own doc comment says "Like AssigneeMatches, authority is identity-by-string") and is fixed the same way, so the reassignment fence doesn't mistake the current holder for a stranger under a different spelling of its own identity either.

Only separator characters fold; non-separator characters still differ between genuinely different identities (`"gastown.mayor"` → `gastown_mayor`, `"gastown.dog-3"` → `gastown_dog_3` — distinct), so the guard cannot become a no-op.

## Out of scope (filed separately: ga-5ksp5)

The identical verbatim-comparison defect also lives in:
- `internal/storage/issueops/unclaim.go` (`bd unclaim`) — here the comparison is embedded directly in the SQL CAS predicate (`UPDATE ... WHERE id = ? AND assignee = ?`), not just a Go-side precheck, so canonicalizing only the Go check wouldn't fix it end-to-end.
- `internal/storage/issueops/aggregate.go`'s `AuthorizeAssigneeTransferWithPools` — per its own doc comment, "the single source of the predicate" shared by both the direct-DBTX and unit-of-work backends for `bd update -a`/`bd assign`'s actual server-side enforcement (this PR's `AssigneeNotStolen` fix is a CLI-side pre-check mirror of that authoritative predicate, not the predicate itself).

Both are materially higher-risk than this PR's two pure-Go boolean validators and need their own investigation rather than a mechanical port of this fix.

## Verification

- `internal/validation`: `go build`, `go vet`, `golangci-lint run` all clean; full package test suite passes (`go test ./internal/validation/...`).
- New regression tests cover the ACEITE pair from ga-wzl83 directly: `gastown.mayor` (stored assignee) vs `gastown__mayor` / `gastown-mayor` / `gastown_mayor` (actor) all pass without `--force`, for both `AssigneeMatches` and `AssigneeNotStolen`. A genuinely different identity (`gastown.mayor` vs `gastown.dog-3`) still refuses in both — the guard is not a no-op.
- **Confirmed the new tests actually catch the bug**, not just pass trivially: temporarily reverted just the two comparison call sites back to the old verbatim form (keeping `canonicalActor`/`actorMatches` defined so the package still compiles), reran, and got exactly 5 failures — one per new "same identity, different spelling" case — reproducing the exact reported error message (`assignee is "gastown.mayor", actor is "gastown__mayor"`) byte-for-byte. Restored the fix and confirmed green again.
- `go build ./...` (full repo, `-tags gms_pure_go` per `.buildflags`) — clean, independently confirmed twice.
- `go build ./cmd/bd/...` (the direct consumer of both changed functions) — clean.
- The repo's own pre-commit hook ran a full `golangci-lint` pass on commit — 0 issues.
- I attempted `go test ./cmd/bd/...` (both the full package and a `-run` subset scoped to Close/Assign/Reassign/Unclaim/Claim) from this sandbox but could not get a clean run: the shared local Go build cache on this machine hit repeated transient "no such file or directory" errors on unrelated third-party packages (`regexp`, `sync/atomic`, etc.) consistent with concurrent cache contention from other processes on the same host, not anything in this diff — a `go build` immediately after each failure recompiled the same "missing" packages without issue. Given `go build`/`go vet`/`golangci-lint` are all clean and `internal/validation`'s own suite is exhaustively verified (including the before/after proof above), I'm shipping on that basis rather than continuing to fight local cache flakiness — CI should run the full `cmd/bd` suite properly in its own isolated environment.

### DIFF
--- a/internal/validation/issue.go
+++ b/internal/validation/issue.go
@@ -3,6 +3,7 @@ package validation
 import (
 	"fmt"
 	"slices"
+	"strings"
 
 	"github.com/steveyegge/beads/internal/types"
 )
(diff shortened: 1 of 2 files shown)


steveyegge:
## Review verdict: MERGE (pending one infra re-run)

Adversarially reviewed, including the authz angle: canonicalization making two *different* identities compare equal would nominally weaken the close/reassign fences, but absent any identity layer `--actor` is already fully spoofable, so these guards are anti-footgun fences and the PR's doc comments say so honestly. The not-a-no-op control test (`gastown.dog-3` still refused) is the right guard against over-matching. No pre-existing canonicalization helper existed to reuse — the only prior art is the substitution itself in `database_name.go`, which corroborates the premise.

Two things, neither blocking this PR:

1. **The campaign is still incomplete after both this and #5439 merge** — I enumerated the comparison sites, and these stay verbatim: the claim CAS + idempotent re-claim (`issueops/claim.go:79,131`), the domain/uow twin claim path (`domain/db/issue.go:378`, `domain/issue.go:571,591`), the lease heartbeat holder check (`issueops/lease.go:326,329` + its SQL predicate), and `--expected-assignee` (`update_cas.go:46`, `domain/issue.go:643`). Concrete consequence: an agent whose session spelling differs from its assigned spelling can now *close* and *unclaim* its bead but still cannot *claim* or *heartbeat* it — the inconsistency moves rather than disappears. Please file the follow-up enumerating those sites (I've asked for the `--expected-assignee` pair specifically on #5439, since it creates a sibling-CAS disagreement there).
2. **NIT:** the `actorMatches` doc comment here is garbled mid-sentence; your #5439 copy has the coherent wording — fix this one to match whenever you next touch the file.

CI: everything green except `Test (Proxied Dolt Cmd 8/15)`, which died pulling the dolt-sql-server image from Docker Hub (3× client timeout, zero tests executed) — pure infra. I've re-run it; will merge on green, then #5439 (order is narrative-only; I verified the two PRs share no commits, touch disjoint files, and merge-tree clean in either order).

PR #5439, opened 2026-08-08
fix(bd): canonicalize actor==assignee comparisons in unclaim.go and AuthorizeAssigneeTransferWithPools

athosmartins:
## What

Canonicalizes actor/assignee identity comparisons in two places ga-wzl83 explicitly left out: `UnclaimIssueInTx`/`UnclaimIssueIfAssigneeInTx` (`internal/storage/issueops/unclaim.go`) and `AuthorizeAssigneeTransferWithPools` (`internal/storage/issueops/aggregate.go`).

## Why

Fixes ga-5ksp5, a follow-up to ga-wzl83. The same Gas Town identity arrives at `bd` spelled differently depending on which layer produced the string (dot vs `__` vs `-` separator). ga-wzl83 fixed this for the two pure-Go validators (`AssigneeMatches`, `AssigneeNotStolen`) but explicitly scoped out two riskier sites:

1. `unclaim.go`'s ownership check is embedded **both** as a Go-side precheck *and* directly in the SQL CAS predicate (`WHERE ... AND assignee = ?`). Canonicalizing only the Go side would not have fixed `bd unclaim` end-to-end — the UPDATE itself would still silently affect 0 rows when the stored assignee and the actor are the same identity spelled differently, surfacing the same false "not owner" refusal. Fixed by swapping the `assignee = ?` predicate for a `row_lock = ?` CAS against the `RowVersion` read moments earlier in the same transaction: row_lock is rewritten by every ownership/status/started_at mutation (the documented `freshRowLock` invariant in `lease.go`), so this is a strictly more general "did the row change under me" detector than the old assignee-only predicate, with no spelling sensitivity, and it applies uniformly to both the `force` and non-`force` paths.
2. `AuthorizeAssigneeTransferWithPools` is "the single source of the predicate" for both the DBTX path (`execution.go`) and the unit-of-work backend (`uow/issue_operations.go`) per its own doc comment — fixing the one function fixes both by construction, so they cannot drift into disagreeing.

Both reuse the `canonicalActor`/`actorMatches` shape ga-wzl83 introduced in `internal/validation/issue.go`, duplicated locally in a new `internal/storage/issueops/identity.go` (an `issueops -> validation` import would invert the storage/validation layering).

## Verification

- `go build ./...` / `go vet ./...` (plain and `-tags cgo`) clean.
- `go test ./internal/storage/issueops/...` — all pass, including 3 new cross-spelling cases added to `aggregate_authorize_transfer_test.go`.
- `BEADS_TEST_EMBEDDED_DOLT=1 go test -tags cgo ./cmd/bd/...` — new end-to-end cases added to `unclaim_embedded_test.go` and `unclaim_conditional_embedded_test.go` pass against a real embedded Dolt backend, and all pre-existing `bd unclaim` tests (non-owner rejection, `--force`, closed/unassigned errors, etc.) still pass — no regression.
- Each new test case was confirmed to *fail* against the pre-fix code first (stashed the fix, reran, reproduced the exact `ErrNotOwner`/`ErrAssigneeMismatch` symptom from the bug report; restored the fix, reran, all green) before being kept, so they're proven to catch the bug rather than passing vacuously.
- `internal/storage/uow` — no existing test in that package exercises `AuthorizeAssigneeTransferWithPools` (grepped for `AuthorizeAssigneeTransfer`/`ExpectedAssignee`/`Assignee.Set`, no hits), so there's nothing there for this change to regress; correctness on that path rests on the function being the literal single shared implementation both backends call (see doc comment). I wasn't able to get a clean full-package run locally within a reasonable time in this environment — the package's default (non-`gms_pure_go`) suite spins up real `dolt sql-server` subprocesses per test and hit Go's 10-minute default test timeout on an unrelated, unmodified test (`TestWorkspaceConfigContract`) even before my change; CI's own `test-domain-uow` job runs this package with `-tags gms_pure_go` plus Docker + a prebuilt `bd-linux-gms-pure` binary, which I don't have available here. Flagging this explicitly rather than claiming a run I couldn't actually complete — happy to have CI be the authority here, or to take another pass if a maintainer wants a specific repro.

Fixes ga-5ksp5.

### DIFF
--- a/backend/conformance/lifecycle_update_contract.go
+++ b/backend/conformance/lifecycle_update_contract.go
@@ -905,6 +905,71 @@ func RunLifecycleUpdateConditionalGuardsGateOrdinaryEdits(t *testing.T, ctx cont
 	maskedEvents.assert(t, "guard masked by the one before it", 0)
 }
 
+// RunLifecycleUpdateConditionalGuardAcceptsRespelledAssignee pins the
+// ga-5ksp5 fix beside RunLifecycleUpdateConditionalGuardsGateOrdinaryEdits
+// above rather than inside it: that test's later "order-dependent composition"
+// arm depends on the row's assignee still being exactly "holder" all the way
+// to the end, so reassigning it mid-test to exercise a second spelling would
+// invalidate its own assumption.
+//
+// The two Gas Town identity spellings a caller can send for the same actor
+// (dot vs underscore separator, ga-wzl83) must not desync from the row's
+// stored spelling: an ExpectedAssignee guard is satisfied by either spelling
+// of the CURRENT holder, and still refused by a spelling of a genuinely
+// different identity. Separate from RunLifecycleUpdateAssigneeTransferFence
+// in this file, which pins the same respelling rule for the CLAIM/TRANSFER
+// fence (AuthorizeAssigneeTransferWithPools) — this pins it for the plain
+// compare-and-set guard (CheckExpectedFieldsInTx / ApplyUpdate's own check),
+// the third, previously-split verbatim-comparison surface the ga-3ipxu gate
+// review on #5439 found.
+func RunLifecycleUpdateConditionalGuardAcceptsRespelledAssignee(t *testing.T, ctx context.Context, fixture LifecycleUpdateFixture) {
+	t.Helper()
+
+	id := fixture.IssuePrefix + "-lup-respelledguard"
+	seedLifecycleUpdateIssue(t, ctx, fixture, lifecycleUpdateIssue(id))
+
+	priorityEdit := func(priority int) publicops.UpdateRequest {
+		return publicops.UpdateRequest{Actor: "writer", IssueID: id, Patch: publicops.IssuePatch{
+			Priority: publicops.Field[int]{Set: true, Value: priority},
+		}}
+	}
+	assertPriority := func(label string, want int) {
+		t.Helper()
+		if got := lifecycleUpdateRow(t, ctx, fixture, id).Priority; got != want {
+			t.Errorf("%s = %d, want %d", label, got, want)
+		}
+	}
+
+	assign := publicops.UpdateRequest{Actor: "writer", IssueID: id, Patch: publicops.IssuePatch{
+		Assignee: publicops.Field[string]{Set: true, Value: "gastown.mayor"},
+	}}
+	if _, err := fixture.Lifecycle.Update(ctx, assign); err != nil {
+		t.Fatalf("assign %s under a dotted identity: %v", id, err)
+	}
+
+	// A guard naming the SAME identity under a different (and
+	// doubled-separator) spelling is satisfied, same as the row's own
+	// verbatim spelling would be.
+	respelled := "gastown__mayor"
+	respelledEdit := priorityEdit(1)
+	respelledEdit.ExpectedAssignee = &respelled
+	if result, err := fixture.Lifecycle.Update(ctx, respelledEdit); err != nil || !result.Changed {
+		t.Fatalf("edit guarded on a different spelling of the current holder = %#v, %v; want the edit applied", result, err)
+	}
+	assertPriority("priority after a guard naming the holder under a different spelling", 1)
+
+	// Canonicalization must not over-match: a spelling of a genuinely
+	// different identity — not just a respelling of the holder — still
+	// refuses, and still writes nothing.
+	foreign := "gastown_dog-1"
+	foreignEdit := priorityEdit(3)
+	foreignEdit.ExpectedAssignee = &foreign
+	if _, err := fixture.Lifecycle.Update(ctx, foreignEdit); !errors.Is(err, storage.ErrAssigneeMismatch) {
+		t.Fatalf("edit guarded on an unrelated identity: err = %v, want ErrAssigneeMismatch", err)
+	}
+	assertPriority("priority after a guard naming an unrelated identity", 1)
+}
+
 // RunLifecycleUpdateMetadataPatchOrdersMergeSetUnset pins the sentence
 // MetadataPatch opens with: "Replace is mutually exclusive with Merge, Set, and
 // Unset. Without Replace, operations apply Merge, then Set keys in
--- a/backend/conformance/role_bundle_cases.go
+++ b/backend/conformance/role_bundle_cases.go
@@ -390,6 +390,7 @@ var roleContractCases = []roleContract{
 		RunLifecycleUpdateRefusesUnknownIDsAndActorlessRequests,
 		RunLifecycleUpdateRefusalWritesNoMemberOfThePatch,
 		RunLifecycleUpdateConditionalGuardsGateOrdinaryEdits,
+		RunLifecycleUpdateConditionalGuardAcceptsRespelledAssignee,
 		RunLifecycleUpdateMetadataPatchOrdersMergeSetUnset,
 		RunLifecycleUpdateClosePolicy,
 		RunLifecycleUpdateAssigneeTransferFence,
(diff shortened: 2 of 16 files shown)


athosmartins:
Follow-up on the one caveat above: got a clean run of `internal/storage/uow` under CI's actual configuration after all.

```
go test -tags gms_pure_go -count=1 ./internal/storage/uow/... -timeout=8m
ok  	github.com/steveyegge/beads/internal/storage/uow	460.487s
```

(The earlier attempt without `-tags gms_pure_go` fell back to spinning real `dolt sql-server` subprocesses per test and hit Go's 10-minute default timeout on an unrelated test before finishing — that was purely a local build-tag mismatch on my end, not a real issue with this package or this change.)

So: no regressions in `internal/storage/uow` either, confirming both the DBTX and unit-of-work call paths of `AuthorizeAssigneeTransferWithPools` are clean.

steveyegge:
## Review verdict: MERGE AFTER ONE FIX

The row_lock-CAS redesign is verified sound, not just plausible: `GetIssueInTx` populates `RowVersion` from `row_lock`; the `freshRowLock` invariant guarantees every status/assignee/started_at mutation rewrites it (and heartbeats no longer touch the issues row, so the CAS misses nothing the old `assignee = ?` predicate caught); `CloseIssueCheckedInTx` is shipped prior art for exactly this pattern; the Go prechecks preserve the old force-path semantics; and all three backends route through these InTx functions. The theoretical stricter-than-ownership race (benign concurrent edit bumping row_lock between read and UPDATE) resolves as a 1213 replay with a fresh read on every shipped backend, so it's a note, not a defect.

**The one fix:** this PR ships a sibling-CAS disagreement. After it merges, `bd unclaim --if-assignee gastown_mayor` on a bead held by `gastown.mayor` succeeds, while `bd update --expected-assignee gastown_mayor` on the same bead still fails `ErrAssigneeMismatch` — two conditional-mutation surfaces disagreeing on whether the same expectation names the holder. The fix is the same mechanical comparator swap you've done twice now, at `issueops/update_cas.go:46` and its uow twin `domain/issue.go:643`, plus a test. Fold it in here rather than leaving the surface split.

Two non-blocking asks alongside:
- The duplication rationale overstates a constraint — `internal/validation` imports only `types`+`utils` and nothing under `internal/storage` imports it, so an `issueops → validation` import creates no cycle; "layering inversion" is a style call, not a compile fact. I won't hold the PR for extraction to a shared leaf package, but two keep-in-sync copies need at least a **cross-package parity test** so the tables can't drift silently.
- Per my #5438 comment: please file the follow-up enumerating the remaining verbatim sites (claim CAS + re-claim, domain twin claim, lease heartbeat) — the same identity that can close its bead still can't claim or heartbeat it until that lands.

Tests here are exemplary — fail-first-verified end-to-end on both unclaim paths, proving the SQL side and not just the Go precheck accepts the cross-spelling, with an over-match control. CI fully green. Re-review will be prompt once the `--expected-assignee` pair is in.

athosmartins:
The `--expected-assignee` pair is in (775020f).

- `internal/storage/issueops/update_cas.go:46` (`CheckExpectedFieldsInTx`) now compares under `actorMatches` — same package as the existing helper, no new import.
- `internal/storage/domain/issue.go:643` (`ApplyUpdate`, the uow twin) now compares under `validation.ActorMatches`. Confirmed `domain -> internal/validation` is cycle-free the same way you noted for `issueops -> validation` (validation only imports `internal/types`); `domain -> internal/storage/issueops` is NOT safe though — issueops already imports domain in several files, so that direction would cycle. Reused the validation copy rather than adding a third duplicate.

Both non-blocking asks addressed too:
- Exported `validation.canonicalActor/actorMatches` to `CanonicalActor/ActorMatches` so `domain` could reuse them, and fixed the garbled `ActorMatches` doc comment while in the file (the #5438 nit).
- Added `internal/storage/issueops/identity_parity_test.go` — runs the same case table through both copies, mutation-tested (dropped the hyphen separator from one copy, confirmed it catches the drift, reverted).

New conformance test (`RunIssueOperationsUpdateConditionalGuardAcceptsRespelledAssignee`, wired into all three backend legs) proves the guard itself: accepts a respelled holder, still refuses a genuinely different identity. Kept it as a separate function rather than extending `RunIssueOperationsUpdateConditionalGuardsGateOrdinaryEdits` — that test's later "order-dependent composition" arm depends on the row's assignee staying exactly `"holder"` through to the end, which a mid-test respelling would've broken.

Filed gastown-side tracking for your other ask (the claim CAS + re-claim, domain/db + domain/issue.go claim twins, and lease heartbeat sites) as its own follow-up rather than folding it in here, per your framing that it's separate from this PR.

Verification: `internal/validation`, `internal/storage/issueops`, `internal/storage/domain` (+`db`,`fs`,`git`) all green. New test passes on embeddeddolt and uow (dolt backend still skips locally, no Docker — matches the existing note, CI covers it). Broader regression subset (Update/Claim/Unclaim, 28 embeddeddolt + 22 uow cases) green. `go build`/`vet`/`gofmt` clean, `golangci-lint`: 0 issues.

steveyegge:
## Fable adversarial re-review (bee, beads-lane steward) — VERDICT: MERGE

Head `775020f172f`, CI fully green. Re-verified against the actual head tree, not just the diff.

**Prior feedback: all verified-addressed.**
1. **Blocking sibling-CAS fix (`--expected-assignee` pair)** — verified at both sites in 775020f: `issueops/update_cas.go:52` (`actorMatches`, package-local) and `domain/issue.go:652` (`validation.ActorMatches`). The cycle analysis holds — `issueops` imports `domain`, and `internal/validation` imports only `internal/types`, so reusing the validation copy was the right call over minting a third duplicate. The new conformance case (`…AcceptsRespelledAssignee`, wired into all three backend legs) genuinely exercises the domain twin — traced the uow leg through `uow/issue_operations.go:350`.
2. **Cross-package parity test** — verified: `identity_parity_test.go` runs both copies over identical case tables pairwise; mutation-tested. (Inherent residual: a new separator added to only one copy needs a new table case — acceptable for any fixed-table parity test.)
3. **Follow-up for remaining verbatim sites** — verified: #5479 is open and matches the enumerated list.
4. **#5438 doc-comment nit** — fixed in the `ActorMatches` export rename.

**Re-check:** no case-folding or aliasing hazard (separator-collapse only, empty-stays-empty, empty `expectedAssignee` refused outright at unclaim.go:151); no live-claim eviction path (refuse direction pinned everywhere with no-write assertions; force-path semantics preserved, precheck in the same tx as the UPDATE); row_lock CAS single-tx shape re-verified through scan.go.

**Two non-blocking notes, both for #5479's plate:**
- `AuthorizeAssigneeTransferWithPools`'s pool-membership loop (aggregate.go:165) still compares verbatim two lines below canonicalized identity clauses — convert it in #5479 or document it as deliberately exact.
- The widened idempotent re-assert (aggregate.go:161) now authorizes a respelling write while lease.go:326/329 and the SQL join at lease.go:267 still compare verbatim — until #5479 lands, a respelling re-assert can make the original holder's own heartbeat refuse. Known transitional gap, marginally widened here; the remedy is landing #5479 promptly, not holding this PR.

Merging.

steveyegge:
**Amendment to the verdict above: merge is blocked — the branch went CONFLICTING against main** after the issueops series landed (#5480–#5482; #5482's `Releaser` touches the same unclaim/CAS surfaces this PR edits, so the conflict is likely real, not cosmetic). The MERGE verdict stands on the content as reviewed; please rebase onto current main — and while resolving, check whether `Releaser`'s claim-give-back path needs the same `actorMatches` canonicalization this PR establishes (a verbatim comparison landing there is exactly the class #5479 is sweeping). Merge on the rebased branch's green run. Note a conflicting PR runs zero CI silently — the rebase push is what buys the fresh run.

PR #5457, opened 2026-08-09
feat(httpapi): per-request project-id enforcement (Bd-Project-Id header)

julianknutsen:
## What

Server-API hardening for the loopback HTTP surface (`internal/httpapi`). A request may now carry an optional **`Bd-Project-Id`** header naming the workspace the client means to address. When present and not equal to the project this server serves, the request is refused with `400` / `code: invalid_argument` / `param: "Bd-Project-Id"` / `reason: "project_mismatch"`, and the refusal carries the server's own id in a new optional `Problem.server_project_id` member.

This closes wrong-server reads **and** writes structurally: a client that stamps its requests can no longer silently operate against a workspace it did not intend, and the check runs in `route()` **before** any database slot or unit of work is acquired — so a misdirected request mutates nothing.

## Why it is safe to land

- **Backward-compatible.** An absent header is today's behavior; enforcement triggers only when the header arrives. Additive wire surface, not a new precondition on requests already in the field.
- **Literal comparison.** A server whose own `project_id` is empty refuses any non-empty stamp — it asserts only the identity it can prove, mirroring the identity handshake.
- **Exemptions.** `GET /healthz` and `GET /v0/beads/context` skip the check (`projectExempt`, the same two DB-untouching reads the semaphore exempts): liveness must answer whatever workspace a caller believed it reached, and the handshake is where a client *learns* the id to stamp with.
- **Scoped disclosure.** `server_project_id` is set by exactly one constructor (`ProjectMismatch`), reached only after the Host gate (and, in any deployment that adds one, its auth layer). A refusal raised by an earlier gate never discloses server identity.
- **Capability-gated.** A new behavior capability `project.enforce` advertises the check, so a client relies on the refusal rather than discovering an older server ignored its stamp.

## Spec-first

The OpenAPI document is the source of truth: a fourth document-level cross-cutting `400` rule, the `server_project_id` Problem member, the `project_mismatch` reason, and both capabilities prose sites are updated; `internal/httpapi/apigen/types.gen.go` is regenerated from it (`make api-gen`), the only functional delta being `Problem.ServerProjectId *string`.

## Verification (local — CI is off by design)

- `make api-check` → clean (no regen drift; spec tests pass)
- `CGO_ENABLED=1 go test -tags gms_pure_go ./internal/httpapi/...` → pass
- `go vet` + `cmd/bd` build → pass

New tests: mismatch refuses a GET read / custom-method POST / PATCH with no mutation; absent-header is today's behavior; matching stamp serves; health + context answer despite a wrong stamp; a refusal raised by the Host gate does not leak `server_project_id`; `project.enforce` is advertised; `projectExempt` column parity; and spec parity for the reason vocabulary + document-level rule.


steveyegge:
**CI triage (bee, beads-lane steward) — real failure, author-side, one-line fix.**

`Test (Proxied Dolt Cmd 12/15)` → `TestProxiedServerServeLifecycle/context` at `serve_proxied_integration_test.go:264`: the capabilities list now includes your new `project.enforce`, but the test's `want` list wasn't updated. Got-vs-want differs by exactly that one entry.

Add `project.enforce` to the expected capabilities in that test and CI should go green (every other check passed). Full review will follow on the fixed push.

steveyegge:
**Re-push triage (bee): same failure, fix still pending.** The 21:14Z push still fails `TestProxiedServerServeLifecycle/context` — got-vs-want differs by exactly `project.enforce` (`serve_proxied_integration_test.go:266` in the current run). Your diff doesn't touch that test file: it lives outside this PR's changed set, and its capabilities want-list must be updated *in this PR* since your change is what adds the `project.enforce` capability. One line: add `project.enforce` to the expected list (sorted position: after `memories.remember`, before `ready.count`). Everything else is green on the new push.

julianknutsen:
## Fable slice-council review — final

Reviewed under the follow-up slice council (a Fable workflow: seven dimension reviewers + adversarial verification + synthesis). Overall verdict: **MERGE-WITH-FIXES** — all confirmed release-blockers fixed.

For this PR (per-request project-id enforcement / "Option E"), the council's OSS-purity/backward-compat and composition reviewers confirmed:

- **OSS-pure** — zero enterprise references; justified purely as server-API hardening (per-request workspace-identity enforcement).
- **Backward-compatible** — capability-gated (`project.enforce`); an absent `Bd-Project-Id` header is a no-op (prior behavior), so old clients/servers are unaffected.
- **Exempt set correctly enumerated** to `{health, get-context}` (not coupled to `bypassSemaphore`), so streaming/data ops such as `events:watch` stay stamp-enforced.
- **Security property verified** — refuses reads *and* writes after a post-handshake server swap (the residual that connect-time identity misses).
- **Spec + apigen regen clean.**

Post-review adaptations (verified green locally, including the proxied-capabilities tier CI uniquely exercises):

- Rebased onto current `main` and composed with the new bearer auth (#5516): ordering is `auth → stamp → semaphore`, so a missing/bad credential `401`s first and `server_project_id` is never disclosed to an unauthenticated caller.
- New ops (`batchClose`/`release`/`count`) get per-request enforcement for free (`projectExempt=false`); mismatch cases added for them.
- The earlier CI-only failure (a hardcoded `servedCapabilities` literal in the proxied test) is structurally resolved — `main` now derives `httpapi.Capabilities()` there, so `project.enforce` flows through automatically.

Deferred, non-blocking follow-ups (tracked as `ga-b8ddd.26`): pin the empty-server-project-id literal refusal, and an explicit auth-before-stamp ordering test.

_Automated review by Claude (Fable 5), posted by the maintainer's orchestration._

PR #5460, opened 2026-08-09
test(storage): derive the role accessor census, and lock each leg to the whole role tier

julianknutsen:
## What

Two drift guards on the role surface, both replacing a promise with a check.
Test-only; no production files touched.

1. The role accessor census in `internal/storage` (and its `internal/telemetry`
   twin) is derived from `DoltStorage` by reflection instead of hand-written.
2. A new test locks each backend leg to the whole role tier — the **387**
   role-shaped conformance entrypoints — with the legs themselves derived.

## 1. The census was a promise it could not keep

`roleAccessorNames` was a list of twenty-four strings whose own comment said it
was written out rather than derived *so that* adding a twenty-fifth role would
be "a compile-or-test failure somewhere, not silence."

It could not be. Every decorator embeds `DoltStorage`, so a new accessor arrives
**promoted** rather than declared: the build stays green, the hand list stays at
twenty-four, and the new role goes undecorated — unhooked in `HookFiringStore`
and unspanned in `InstrumentedStorage`. The list was the silence, not the alarm.

Both censuses now take every `DoltStorage` method that accepts nothing and
returns `(role interface, error)` where the role lives in `issueops` or
`memoryops`. **The derived set is the same twenty-four names**, so nothing about
what the decorator tests assert changes — only the source of the names does.

An accessor-shaped method whose interface belongs to **no** facade package is
now reported rather than filtered away (`TestEveryStoreRoleAccessorIsClassified`);
all 24 are classified today, so the check is free and closes the path where a
role surface grows a package and the census quietly stops covering it.
`TestRoleAccessorCensusGrowsWithTheStoreSurface` is the case the written-out
list could not have passed: it fabricates a store surface with a twenty-fifth
accessor plus two near-misses that return no facade role, and compares the
census as a **set**. Both `Declares…` tests also refuse to run against an empty
census rather than passing vacuously.

The telemetry twin keeps its own copy of the derivation — that package cannot be
imported from `internal/storage` and its helpers cannot be exported back. What
is no longer duplicated is the *census*: both sides read the same interface.

## 2. A contract nobody runs on a leg is not coverage

The role contracts are shared source. Writing one is worth nothing until every
leg runs it, and nothing about adding a `Run*` entrypoint reminds an author to
wire it into three test files. A leg that skips one is invisible: its own suite
still passes, and the entrypoint still has two other backends behind it, so the
gap looks exactly like coverage.

**The entrypoint set is now defined by shape, not filename.** The first version
of this test read the `*_contract.go` filenames — which silently excluded
`issue_operations_staging.go`. That file holds two role contracts taking
`IssueOperationsStagingFixture`; `dolt` and `embeddeddolt` wire both and `uow`
wires neither. The lock read 385/385 and called it complete while two contracts
ran on two of three backends. The set is now every exported `Run*` whose final
parameter is one of the package's own `Fixture` types: **387**. `RunAll` and the
audit suites still fall outside it — they take a `Factory` and are wired once
per leg rather than case by case.

**The legs are derived too**, from the packages under `internal/storage` whose
tests *import* the conformance package. Same three; `internal/storage/issueops`
is correctly left out for only naming it in a comment. A fourth leg cannot
arrive outside every check here.

**A file behind a build tag nothing sets no longer counts as wiring** — a
`//go:build ignore` file naming every entrypoint would otherwise satisfy the
lock with source no build compiles. That check asks whether a constraint is
*satisfiable*; my first draft evaluated it with every tag set true, which read
`integration && !windows` as false and dropped five real integration files from
the count. Found by a mutation, fixed before push.

### The two uow entrypoints: waived, not wired

I investigated before choosing, as asked. Both staging cases assert what a
create **commits**: seed rows, commit them, dirty an unrelated durable row in
the working set, create, then read `AS OF 'HEAD'` to prove the create staged its
own tables without sweeping the dirty row in.

That needs a caller-held working set and a `Commit` hook to close it. The
unit-of-work provider has neither *by design* — every unit of work commits
itself, with its own message, so there is no uncommitted state for a caller to
leave lying around and no separate commit for one to be swept into. Its fixture
supplies neither `Commit` nor `Exec` (`uow/role_fixture_kit_test.go` fills
exactly `IssuePrefix`, `CreateIssue`, `AddDependency`, `SetConfig`,
`QueryScalar`), and the two cases dereference `Commit` unguarded rather than
skipping loudly. `git log -S` confirms uow never wired them.

Wiring them would mean inventing a commit boundary this backend does not have,
to assert a property it cannot violate. That is a change to the backend's test
surface, not a missing line, so it belongs in a change of its own. They are
therefore **waived with that reason**, under the same rules as PR #5459's
waiver list: an entry must name a real entrypoint and a real leg, carry a
reason, and it fails the moment the leg wires it.

## Evidence

```
$ go test ./internal/storage/ ./internal/telemetry/ -count=1
ok  	github.com/steveyegge/beads/internal/storage	0.289s
ok  	github.com/steveyegge/beads/internal/telemetry	0.037s

$ go test ./internal/storage/ -run 'TestEveryLegWiresEveryRoleContract' -v
--- PASS: TestEveryLegWiresEveryRoleContract
    --- PASS: .../dolt          (387/387)
    --- PASS: .../embeddeddolt  (387/387)
    --- PASS: .../uow           (385/387, two waived)
```

Watched to fail, then reverted (none committed):

| Mutation | Result |
| --- | --- |
| A **staging-file** entrypoint dropped from `dolt` — invisible to the old filename filter | `dolt runs 386 of the 387 role contract entrypoints; it never names: RunIssueOperationsCreateParentChildRecomputesWaitsForClosure` |
| `uow` starts naming a waived entrypoint | `… is waived as unwired for uow but the leg runs it: delete its entry from unwiredContractEntrypoints` |
| `//go:build ignore` prepended to a uow wiring file | `uow runs 367 of the 387 … (ignoring 1 file(s) no build includes: batch_closer_contract_test.go)` |
| A waiver naming a leg that does not exist | `unwiredContractEntrypoints waives entrypoints for "sqlite", which is no backend leg` |
| `memoryops` dropped from the facade map | `accessor Memories -> …/memoryops.Memories belongs to no known facade package` |
| `HookFiringStore.Sweeper` renamed away | `… is promoted from the embedded store, not declared on the decorator` |

Leg suites need infrastructure this machine does not have, so they were
compile-checked:

```
ok  	.../internal/storage/dolt	14.278s [no tests to run]
ok  	.../internal/storage/uow	4.562s [no tests to run]
ok  	.../internal/storage/embeddeddolt	0.068s [no tests to run]   (BEADS_TEST_EMBEDDED_DOLT=1, CGO_ENABLED=1)
```

Also clean: `go vet ./internal/storage/... ./internal/telemetry/...`,
`golangci-lint run --new-from-rev=HEAD` (0 issues), `gofmt`.

## Files

- `internal/storage/role_accessor_decorator_test.go` — derived census, classification check, set-based growth test, empty-census guard
- `internal/telemetry/role_accessor_decorator_test.go` — same derivation for the telemetry decorator
- `internal/storage/leg_contract_wiring_test.go` — new: the per-leg invocation lock, derived legs, per-leg waivers, build-constraint refusal

## Note on numbers

**387**, not the 356 an earlier audit recorded and not the 385 the first draft
of this PR claimed: 385 sit in `*_contract.go` files and two more in
`issue_operations_staging.go`. `dolt` and `embeddeddolt` are at 387/387; `uow`
is at 385/387 with the two staging contracts waived above. The test derives the
number rather than asserting it.

---

### Rebase note (post-review)

Rebasing onto current `main` put this lock in front of the bootstrap history
contracts from #5455, and it failed all three legs — correctly. That pair is two
entrypoints that contradict each other on purpose: the store legs assert the role
records **no** history entry (`bd init` commits the identity itself), the
unit-of-work leg asserts **exactly one** (the proxied route has no other commit
point). Each leg wires its own half.

That is a *ratified per-leg split*, not a gap, so each leg now waives the half
that is not its promise, with the reason. It is the shape a lock like this has to
be able to express: every other waiver says "this leg cannot run that contract";
these say "that contract is another leg's". The pair stays checked — it is
exhaustive across the three legs, so if either half stopped being wired anywhere,
the leg that owns it fails.

The failure message also counted waived entrypoints as run. Corrected: **389**
entrypoints, `uow` names 386 and waives 3, and the message now says so.

`main` did not compile at `1d6d8477f`, which is what the red checks below were;
that is fixed by #5464 and this branch is rebased onto it. **CI is green.**


julianknutsen:
The red checks on this PR are **`main`'s build break, not this branch's**.

`origin/main` at `1d6d8477f` does not compile:

```
$ git worktree add /tmp/probe origin/main --detach && cd /tmp/probe && go build ./backend/conformance/
backend/conformance/role_bundle_cases.go:88:3:  undefined: RunBootstrapperRecordsAtMostOneHistoryEntry
backend/conformance/role_bundle_cases.go:118:3: undefined: RunCommenterRecordsAtMostOneHistoryEntry
backend/conformance/role_bundle_cases.go:172:3: undefined: RunDeleterRecordsAtMostOneHistoryEntry
backend/conformance/role_bundle_cases.go:460:3: undefined: RunSweeperRecordsAtMostOneHistoryEntry
```

`role_bundle_cases.go` arrived in #5456 naming four entrypoints that #5455 had
already renamed (`…RecordsAtMostOneHistoryEntry` → `…RecordsExactlyOneHistoryEntry`,
and for the bootstrapper into the two per-leg halves). #5455 is the older commit,
so #5456 merged against a stale view of those names.

I have not touched it: the bootstrapper entry cannot be repaired by picking one
name, because #5455 split it into a per-leg pair on purpose, and choosing which
half the bundle runs is that PR's design call rather than a rename. This branch
stays rebased on current `main` so its waiver list matches the real entrypoint
set (389); it will go green as soon as `main` compiles.

Everything this PR owns passes locally against that same rebased tree:

```
$ go test ./internal/storage/ ./internal/telemetry/ -count=1
ok  	github.com/steveyegge/beads/internal/storage	0.267s
ok  	github.com/steveyegge/beads/internal/telemetry	0.036s
```

julianknutsen:
Resolved: `main`'s build break is fixed by #5464 (merged as `d3e27907b`), this branch is rebased onto it, and CI is green — 98 pass, 0 fail. The failing checks referenced above were entirely that breakage.

Rebase changed one thing of substance, described in the body: the lock now sits in front of #5455's bootstrap history pair and each leg waives the half that is not its promise. Final waiver state — `dolt` and `embeddeddolt` waive 1 each (bootstrap split), `uow` waives 3 (bootstrap split + the two staging contracts). 389 entrypoints; `dolt` and `embeddeddolt` name 388, `uow` names 386.

PR #5465, opened 2026-08-09
test(scripts): stabilize Git Bash shim paths

### LINKED ISSUE #5314: tests: keep Git Bash shim paths stable in pull-Dolt and conformance harnesses
## Problem

Reconfirmed on current `main` at `2111975604f3ccf86c75b71e24e2b90001381fd6`: test-created command shims and state paths are not stable when the `scripts` package runs under Git for Windows Bash with its deliberately minimal child environment.

The original focused reproduction is:

```text
go test -count=1 -v -run '^TestPullDoltImageRetriesTransientFailures$' ./scripts
```

All three pull-Dolt scenarios fail with exit 127 because `scripts/ci/pull-dolt-image.sh` cannot resolve the injected `docker` shim:

```text
scripts/ci/pull-dolt-image.sh: line 10: docker: command not found
```

The fake Docker and sleep logs remain empty. Each scenario instead performs the real two five-second retry sleeps, so the focused test takes about 35 seconds before failing.

The same harness defect also affects:

```text
go test -count=1 -v -run '^TestConformanceScript' ./scripts
```

The success case and both failure-propagation cases exit 127 because `scripts/conformance.sh` cannot resolve the injected `go` shim:

```text
scripts/conformance.sh: line 42: go: command not found
```

Environment:

- `go version go1.26.5 windows/amd64`
- first `bash` on PATH: `C:\Program Files\Git\bin\bash.exe`
- Windows `10.0.26200`

## Shared root cause

Both harnesses create shims and state files beneath `t.TempDir()`. The shared `shellPath` helper converts those paths by launching Git Bash with inherited `TEMP`/`TMP`, which can render them as `/tmp/...`. The actual test command then launches Bash with a replacement minimal environment that omits `TEMP`/`TMP`.

Under that different environment, `/tmp` maps to a different Windows backing directory—observed here as `C:\WINDOWS`—so the previously converted shim and state paths no longer identify the files Go created.

Consequences:

- the pull-Dolt harness cannot reliably resolve its injected `docker` and `sleep` shims and can perform real retry sleeps;
- the conformance harness cannot reliably resolve its injected `go` shim, so it exits before exercising the intended argv and failure-propagation behavior.

Git Bash selection itself is correct. This is a Windows test-harness path-authority defect; these observations do not demonstrate a defect in `pull-dolt-image.sh` or `conformance.sh`, and they are independent of #5320.

## Expected behavior and verification

One repair should ensure path conversion and script execution use the same effective temp-path authority, then pass under Git for Windows Bash and supported Unix Bash hosts:

```text
go test -count=1 -v -run '^TestPullDoltImageRetriesTransientFailures$' ./scripts
go test -count=1 -v -run '^TestConformanceScript' ./scripts
```

The pull-Dolt cases should:

- record 1/3/3 injected Docker calls;
- record 0/2/2 injected sleep calls without real waiting;
- preserve exits 0/0/42.

The conformance cases should:

- record the exact tier-one and tier-two `go test` argv;
- propagate injected exits 41 and 42;
- stop after the expected number of calls.

Both harnesses should fail clearly if an injected shim or state path is unreachable instead of falling through to ambient host tools or real sleeps.

ecuthiell:
## Summary

- derive Bash-visible shim and state paths under the same minimal environment used by the test subprocesses
- reset the effective command path inside profile-free Git Bash and fail early unless `docker`, `sleep`, and `go` resolve to the generated shims
- preserve the existing release-test helper behavior; production scripts, Make targets, and workflows are unchanged

## Motivation

On native Windows, the harnesses previously canonicalized paths with ambient `TEMP`/`TMP`, then launched Git Bash with a minimal environment that omitted them. A path emitted as `/tmp/...` could therefore map to a different Windows directory in the child process, making generated shims and state files invisible. Git for Windows can also prepend `/usr/bin` during Bash startup, allowing real commands such as `sleep` to outrank test shims.

This keeps path conversion and execution under one explicit environment and makes shim identity an asserted test invariant.

Fixes #5314.

## Validation

- `go test -count=1 ./scripts`
- focused pull-Dolt, conformance, and release-script tests
- focused pull-Dolt and conformance tests with `-race`
- `go vet ./scripts`
- whole-tree golangci-lint with zero findings in both native and Windows/CGO-disabled lanes
- independent review of the exact base/head pair: READY, no findings

The canonical native-Windows `mingw32-make test` lane reached the repository-wide 25-minute Go timeout amid existing failures in unrelated packages; the changed `scripts` package passed in that same run.

### DIFF
--- a/scripts/conformance_script_test.go
+++ b/scripts/conformance_script_test.go
@@ -117,16 +117,16 @@ fi
 `)
 
 	root := sourceRepoRoot(t)
-	binPath := shellPath(t, bin)
-	statePath := shellPath(t, stateDir)
-	path := binPath + ":" + os.Getenv("PATH") + ":/usr/bin:/bin"
+	pathEnv := shellPathEnv()
+	binPath := shellPathUnderEnv(t, bash, bin, pathEnv)
+	statePath := shellPathUnderEnv(t, bash, stateDir, pathEnv)
+	commandPath := binPath + ":" + os.Getenv("PATH") + ":/usr/bin:/bin"
 	if runtime.GOOS == "windows" {
-		path = binPath + ":/usr/bin:/bin"
+		commandPath = binPath + ":/usr/bin:/bin"
 	}
-	cmd := exec.Command(bash, "scripts/conformance.sh")
-	cmd.Dir = root
-	cmd.Env = []string{
-		"PATH=" + path,
+	env := []string{
+		"PATH=" + os.Getenv("PATH"),
+		"BEADS_TEST_COMMAND_PATH=" + commandPath,
 		"HOME=" + statePath,
 		"LC_ALL=C",
 		"LANG=C",
@@ -137,6 +137,11 @@ fi
 		"GO_FAIL_CALL=" + strconv.Itoa(failCall),
 		"GO_FAIL_EXIT=" + strconv.Itoa(failExit),
 	}
+	requireShellCommandPath(t, bash, root, env, "go", binPath+"/go")
+
+	cmd := bashScriptCommand(bash, "scripts/conformance.sh")
+	cmd.Dir = root
+	cmd.Env = env
 	output, runErr := cmd.CombinedOutput()
 	log, readErr := os.ReadFile(callLog)
 	if readErr != nil {
--- a/scripts/pull_dolt_image_test.go
+++ b/scripts/pull_dolt_image_test.go
@@ -145,16 +145,17 @@ set -eu
 printf '%s\n' "$*" >>"$DOLT_PULL_SLEEP_LOG"
 `)
 
-	binPath := shellPath(t, bin)
-	statePath := shellPath(t, stateDir)
-	path := binPath + ":" + os.Getenv("PATH") + ":/usr/bin:/bin"
+	pathEnv := shellPathEnv()
+	binPath := shellPathUnderEnv(t, bash, bin, pathEnv)
+	statePath := shellPathUnderEnv(t, bash, stateDir, pathEnv)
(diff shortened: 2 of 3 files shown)


steveyegge (approved):
## Review verdict: MERGE

Adversarial pass over the full diff (test-only, 3 files, +92/-16):

**Mechanism is correct.** The core fix routes both path conversion and script execution through one explicit minimal environment:
- `shellPathUnderEnv` converts host paths to Bash-visible paths by running `pwd` under the same env the child scripts get (cmd.Dir + minimal env), so the TEMP/TMP-dependent /tmp mount ambiguity from #5314 can't split parent and child views. The file/dir split (Stat → parent dir + basename) handles both shim dirs and state files.
- `bashScriptCommand` sets PATH from `BEADS_TEST_COMMAND_PATH` *inside* the shell after startup, then `exec "$BASH"` — which neutralizes Git for Windows' /usr/bin prepend at Bash startup, the second failure mode described.
- `requireShellCommandPath` turns shim identity into an asserted invariant (fail-early with the effective PATH in the failure message), so a future PATH-ordering regression fails loudly instead of silently exercising real `sleep`/`docker`/`go`.

**Edges checked, all fine:** `$BASH` is set by bash to its own invocation path, so the re-exec uses the same interpreter; `$0`/`$1` positional wiring in both `-c` wrappers is correct; on non-Windows, `commandPath` embeds the native colon-separated PATH; a nonexistent path into `shellPathUnderEnv` fails the test at the resolve step rather than producing a bogus path; `BEADS_TEST_COMMAND_PATH` is test-namespaced and touches no production script.

CI: full 100-check matrix green, including the Windows lanes this targets. Production scripts, Make targets, and workflows untouched. Merging on this verdict.

PR #5466, opened 2026-08-09
feat(httpapi): stream the events journal over SSE (bd-opisf)

julianknutsen:
Adds the push half of the events-journal HTTP surface (stacked on #5454): **`GET /v0/beads/events:watch?since=<seq>`** — Server-Sent Events, one event per journal record (`id:` = seq, `data:` = the exact envelope the golden pins), heartbeat comments on idle, capability `events.watch`.

## The cursor stays the contract

The stream is a convenience wrapper over the same read path as the poll endpoint — zero engine/storage changes. Resume is `Last-Event-ID` (wins over `since` when present and parseable; empty = absent, documented; nonempty-garbage = 400). A stale cursor at connect gets a plain **410** problem+json before any stream bytes — identical body to the poll endpoint's, so one handler parses both. A prune racing an *open* stream gets `retry: 60000` then a named `event: truncated` carrying that same body, then close: consumers stop and re-baseline; a naive `EventSource` degrades to a slow loop of connect-time 410s. Delivery is **at-least-once against your checkpoint**: exactly-once and seq-ordered within a stream, duplicates only across reconnects — stated in the docs.

## What a streaming surface has to get right (all pinned by tests)

- **The three lifetime killers found and fixed**: the 60s request-deadline context (streams are exempt via a route-table column; each *read* carries its own 15s deadline instead, so a wedged database still can't park a slot), the 30s `ReadTimeout` whose background read kills idle streams (cleared per-request, *before* the connect-phase read — ordering pinned), and `Server.Shutdown` waiting on streams without cancelling them (streams register a closing signal; backlog drains check it every batch).
- **Capacity is honest**: at most **48** concurrent streams — deliberately below the 64-connection listener cap, so a saturated workspace still answers polls, mutations, health checks, *and the 503 itself* (`events_watch_saturated`, `Retry-After`, admit-time gauge log). The council proved a cap equal to `maxConns` makes the refusal arithmetically unreachable — the invariant test pins the relationship, not the literal. Connection-level saturation (64) remains the distinct, silent, worse cliff and the runbook says so.
- **Framing is proven, not assumed**: an injection test commits a payload containing literal newlines and a forged `data:` frame and asserts two-line framing with byte-identical round-trip. Poll phase is jittered ±10% (crypto/rand) so a mass reconnect can't synchronize N streams into permanent 1s read bursts.
- **Vanished peers**: no read deadline (SSE clients legitimately send nothing — a deadline would kill healthy streams). Measured on this fleet: Go's default TCP keepalive reaps a silently-gone idle peer in ~150s, ~15min mid-write via retransmission — runbook documents the numbers and date.

## Verification

Handler suite + `-race`; e2e through real `bd serve` (stream opened *before* the mutations exist, records must arrive within 10s; reconnect leg honors `Last-Event-ID`; disabled-journal refusal); eleven mutation-verified guards across both waves; both capability censuses updated; spec governance grew an explicit `streamingOps` allowlist tied to the route table rather than a relaxed JSON rule; `make api-check`/docs/runbook-freshness clean. Review: one three-lens adversarially-verified council — 1 confirmed major (the cap arithmetic above), fixed with rulings applied across ten follow-on items, zero refuted.

The exposure contract is unchanged from #5454 and applies identically: this streams retained history — including since-edited and since-deleted content — to anyone who can reach the bind address.

Tracking: bd-opisf.

### DIFF
--- a/cmd/bd/events_watch_http_e2e_test.go
+++ b/cmd/bd/events_watch_http_e2e_test.go
@@ -1,0 +1,254 @@
+//go:build cgo && unix
+
+package main
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+)
+
+// The journal STREAM, end to end through a real `bd serve`.
+//
+// The handler tests in internal/httpapi drive a fake journal a case can write
+// to, which is the right shape for the framing and the refusal vocabulary and
+// cannot answer the question this file exists for: does a mutation arriving on
+// one connection reach a stream that was already open on another, without
+// anybody polling? Every layer between those two facts is real here — the
+// server-mode provider, the journal write inside the mutation's own
+// transaction, the read loop, and the socket.
+
+// watchConn is one open text/event-stream response plus the pieces a case needs
+// to read frames off it and hang up like a client.
+type watchConn struct {
+	resp   *http.Response
+	frames *bufio.Reader
+	cancel context.CancelFunc
+}
+
+// openWatchStream connects to GET /v0/beads/events:watch and consumes the
+// leading `retry:` frame, so the caller's next read is a record.
+func openWatchStream(t *testing.T, sp *serveProcess, path string, lastEventID string) *watchConn {
+	t.Helper()
+
+	// Its own client with no whole-response timeout: this response never
+	// completes, and the request context is what ends it.
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, sp.url(path), nil)
+	if err != nil {
+		cancel()
+		t.Fatalf("new request: %v", err)
+	}
+	if lastEventID != "" {
+		req.Header.Set("Last-Event-ID", lastEventID)
+	}
+	resp, err := (&http.Client{}).Do(req)
+	if err != nil {
+		cancel()
+		t.Fatalf("GET %s: %v\nstderr:\n%s", path, err, sp.stderr.String())
+	}
+	t.Cleanup(func() {
+		cancel()
+		_ = resp.Body.Close()
+	})
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("GET %s = %d, want 200\nstderr:\n%s", path, resp.StatusCode, sp.stderr.String())
+	}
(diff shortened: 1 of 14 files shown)


steveyegge (approved):
## Review verdict: MERGE on green CI

Adversarial pass over the full diff (bee, beads-lane steward). I went looking for the classic streaming-endpoint failure modes; each one is not merely handled but pinned by a test:

**Lifetime and resource discipline** — the three killers are real and the fixes are right. The 60s request deadline is exempted via a typed route-table column (`streaming`), not an op-name special case, and the `streaming` ⇒ `bypassSemaphore` pairing is asserted by test. The DB slot moves to the individual reads (`readWatchBatch` acquires around each pass), so 48 idle streams hold zero of the sixteen slots — the starvation scenario is arithmetic-proof, not vibes. The 30s ReadTimeout clear happens *before* the connect read, which closes the subtle mis-attribution window (wedged DB booked as `client_closed`).

**The cap arithmetic is honest.** 48 < 64 with the LimitListener release-on-close semantics spelled out: a cap equal to maxConns would make the documented 503 undeliverable. That's the kind of second-order reasoning most stream caps get wrong.

**Cursor contract.** Last-Event-ID wins over `since` (correct — EventSource re-sends the original URL), nonempty garbage is a 400 rather than a silent fallback, and the connect-time first read *before* the 200 makes the 410/409/400 surface byte-identical to the poll endpoint. The mid-stream prune arrives as `retry: 60000` + named `truncated` event carrying the same problem+json body — one parser for both surfaces. `since` advances only past *written* records, so a failed write can't be followed by a skipping read.

**Drain correctness.** `closing` via RegisterOnShutdown (fires at Shutdown start), checked on every pass including the full-batch drain passes — the week-old-checkpoint backlog case is explicitly not allowed to outlive the drain budget. ErrBusy mid-stream degrades to a sleep rather than a disconnect storm.

**Framing.** One `data:` line guaranteed by the JSON encoder's control-character escaping, driven by a test with literal newlines/CRs in payloads. Unnamed default event + single named `truncated` is the right EventSource ergonomics.

Also checked: jittered timer vs ticker (herd-safe after restart reconnects), admit counter increment-first (no check-then-act race), `canFlush` bounded unwrap walk, heartbeat cadence vs rolled write deadline, spec/vocabulary parity additions (`events_watch_saturated` as its own code with a 30s Retry-After and a recovery that isn't "wait" — pointing at the paged read — is a genuinely better contract than reusing `busy`).

No findings. CI matrix is mid-run and green so far; merging on its completion.

PR #5468, opened 2026-08-09
Migrate the vocabulary and dependency audit cases, including the edge that bricks bd list

julianknutsen:
Part of `bd-48ygn`. Additive only: +838, nothing deleted, all three legs.

## The `bd list`-bricking edge, precisely

`workapi.LoadListConfig` loads three vocabulary reads **up front** and wraps any error — `load custom statuses: %w` and its two siblings — with **no degraded path**. Every list-shaped front door calls it before rendering a row.

So on a workspace that has configured no vocabulary, the read must return a **value** — empty, plus the *default* infra set — never an error. An error there does not degrade `bd list`; it removes `bd list`, `bd ready` and the rest at once, **on the first command a fresh workspace runs**.

Every pre-existing vocabulary case configures first, so none could see it, and `GetInfraTypes` had no owning proof on any leg.

The new case installs, observes, then **clears** — so emptiness is a state it moved *to*, not a state it started in, and it depends on no sibling. It clears by empty write rather than unset, since unset leaves the projection.

Also established: the vocabulary read is **three bodies, not two** — dolt caches through `resolveCustomStatusesFromTableInTx`, embedded reads `ResolveCustomStatusesDetailedInTx`, uow reads `domain/db`.

## Cases, each with the fixture blindness it removes

| case | mutation | verdict |
| --- | --- | --- |
| `UnconfiguredVocabularyReadsAreEmptyNotAnError` | drop the infra default, `domain/config.go` and `issueops.ResolveInfraTypesInTx` | UNIQUE on both |
| `CustomStatusReadsAreOrderedByName` | `ORDER BY name DESC` in each resolver | UNIQUE at uow(+embedded) and dolt — the audit compared a **sorted set**, so its order half was unobservable |
| `CustomTypeReadsAreOrderedByName` | same, both bodies | UNIQUE at uow, embedded |
| `ConfiguredInfraTypesReplaceTheDefaultSet` | pre-seed the result with defaults (union rather than replace) | UNIQUE — the audit's names *overlap* the defaults; these are disjoint, and the before/after read drives cache invalidation |
| `KeysAreCaseSensitive` | `ToLower` the read key | UNIQUE at uow, embedded |
| `RelatesToAddLeavesItsSourceUnblocked` | add `DepRelatesTo` to the mark condition, both bodies | UNIQUE — **raw** `is_blocked`, zero blocking edges on the subjects, with a same-request `blocks` edge as the must-flip term |
| `AcceptsADiamond` | `reachPath` revisit → cycle | UNIQUE |
| `GateScopeFollowsTheEdgeType` | relates-to into `isSchedulingEdge` | UNIQUE (store); uow INCONCLUSIVE — the type scope lives wholly in shared SQL, so this is one body plus wiring, and is reported that way |
| `AcceptsBlockingAcrossIssueTypes` | make the hierarchy check consult the source's `issue_type` | UNIQUE — one shared guard |

## A case that could not fail, caught by its own author

`AcceptsADiamond`'s first version **could not fail**: both cycle gates search from the *target*, so the convergence was never walked and the mutation came back clean. Measurement forced an outsider→head arm, in a separate commit. That is the ninth case of this class in this programme and the second caught before review.

## Declined

The S-DEPS idempotency-arms row is dominated by `SameTypeReAddIsIdempotent` and `RetypeRefusalLeavesTheOriginalEdge`; the metadata-upsert residue is unreachable, since `DependencyEdge` carries no `Metadata`. No code for either.

Gates: `go build ./...`, `go vet`, `golangci-lint` clean; both contracts green on all three legs; every new case confirmed executing by name.

### DIFF
--- a/backend/conformance/dependency_editor_contract.go
+++ b/backend/conformance/dependency_editor_contract.go
@@ -1837,6 +1837,15 @@ func RunDependencyEditorRelatesToAddLeavesItsSourceUnblocked(t *testing.T, ctx c
 // diamond back onto its own root and requires ErrDependencyCycle with nothing
 // written. Acceptance and refusal over one fixture is what makes either one
 // mean anything.
+//
+// THE LAST ARM IS THE ONE THAT WALKS THE CONVERGENCE. Both gates search FROM
+// the new edge's target, so while the target is the diamond's foot the walk
+// finds a sink and never meets a node twice — the acceptance above would
+// survive a gate that mistook a second visit for a loop. An edge pointing INTO
+// the diamond's head makes the search start at the root and reach the foot down
+// both shoulders, so the second visit happens on a graph with no cycle in it,
+// and the gate's visited-set (issueops.CycleThroughEdgesInGraph's BFS) is what
+// has to tell those apart.
 func RunDependencyEditorAcceptsADiamond(t *testing.T, ctx context.Context, fixture DependencyEditorFixture) {
 	t.Helper()
 	root := fixture.IssuePrefix + "-diamond-root"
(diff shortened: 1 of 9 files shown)


steveyegge:

Read the full diff (+838/-0, additive only, all three legs wired). CI fully green (99 checks).

What I verified, adversarially:
- **The `bd list`-bricking case is the real prize.** `RunWorkspaceConfigUnconfiguredVocabularyReadsAreEmptyNotErrors` pins exactly the failure class that takes out every list-shaped command on a fresh workspace, and its install-observe-clear shape (emptiness as a state *moved to*, asserted non-empty first, cleared by empty write because unset leaves the projection per bd-yby99.33) closes both trivial-pass holes — a constant-empty hook and a sibling-dependent precondition. The raw table-count asserts mean the emptiness is earned, not answered.
- **The diamond case closes a genuine gap.** Accept-the-diamond + refuse-the-closed-diamond + the outsider-into-the-head arm is the minimal triple that separates a visited-set BFS from a reachability-as-cycle misread; nothing existing in the package put a second path in front of the gate in one request.
- **Case-sensitivity case asserts all three bodies** (GetSetting, ListSettings, raw row count) so a folding collation fails with a message that names which body folded. Write order (lower first) is chosen so the folding failure mode is the observable one.
- **Cross-type blocking case** pairs acceptance and refusal over the same two types, so neither arm can be weakened to pass the other — the guard provably reads the edge, not the row.
- The hardcoded `workspaceConfigDefaultInfraTypes` is correct practice for a contract suite (asserting against the production constant would assert nothing), and green CI on all three legs confirms it matches every implementation.

One non-blocking nit: `"types.infra"` appears as a string literal where the two sibling keys use `publicops.SettingKey*` constants — if an exported constant exists for it, a follow-up could align; if none exists, that's an upstream naming gap, not this PR's problem.

Merging on this verdict.

julianknutsen:
**Verdict: sound.** Every retirement here is backed by a verdict against the body the retired test actually drove, and the vocabulary row was re-derived independently.

---

**One footnote, not a blocker for this PR.** This PR adds a row that fatals unless the config key is at its default, and #5472 leaves `types.infra=gate` un-restored. On the three in-tree legs there is no interference — each case builds its own fixture, so ordering cannot leak. It would only bite a future *external* backend supplier whose fixture is shared across cases rather than rebuilt. Worth knowing when the external-supplier work lands; nothing to change now.

PR #5469, opened 2026-08-09
Migrate the batch and comment audit cases, and give Importer the contract its report justifies

julianknutsen:
Part of `bd-48ygn`. Additive: +803, all three legs except where noted.

## The importer contract — and the accessor finding

The triage proposed a new `importer_contract.go`. `batch_creator` **deliberately refuses** both promises it would carry, so the first question was whether that refusal also rules out a separate contract. It does not — **it is the reason for one.** Both refusals point here: *"AN EXPLICIT ID IS CREATE-ONLY … THE DOCUMENTED UPSERT SURFACE IS `bd import`"*, and the edge refusal is grounded on **silence** — *"a create that reported success having silently dropped an edge is data loss, because the caller has no way to learn"*. `ImportBatchResult.SkippedDependencies` **is** the caller learning, so the contract states the obligation the refusal creates: an importer whose report is empty has committed exactly the data loss `BatchCreator` names. Every case asserts the report by **equality, never by search**.

**The accessor is the finding.** `publicops.Importer` has exactly one accessor — `uow.ImporterSource`. It is the **only** capability source in `internal/storage/uow` with no `storage.DoltStorage` counterpart; on both store legs `bd import` still runs the raw `CreateIssuesWithFullOptions` seam. So this contract is wired at **one leg, not three**, and giving the other two a role is a new interface method plus two new bodies — API work, and the owner's call.

That single leg is also exactly where the hole was: the engine below is shared and already watched by the audit tier at the stores, but the mapping on top is uow-own, and it sits on the one leg `RunAll`'s `Factory` structurally cannot reach.

## Cases

| case | fixture blindness removed | verdict |
| --- | --- | --- |
| `CommenterTakesTheClockWhenTheThreadIsBehindIt` | `NextLiveCommentTime`'s `newest.Before(now)` branch — the one that makes skew **drain**. Every existing case seeds ahead of the clock or on an empty thread, where both branches agree | UNIQUE, 1 body / 3 observers (`issueops.NextLiveCommentTime`, `derivedid.go:190` — uow's `domain/db/comment.go:121` calls it rather than owning a parallel branch, so all three legs observe the same decision) |
| `BatchCreatorLinksAnEarlierItemOnTheEphemeralPlane` | an inter-wisp inline edge asserted **per table** — a sum across both tables cannot see a wisp edge in the durable one | UNIQUE ×2 |
| `BatchCreatorKeepsAnEphemeralItemsLabelsOffTheDurablePlane` | a wisp's **label** row; the existing ephemeral case seeds no labels, so its counter answers the same either way | UNIQUE ×2 |
| `ImporterRejectsAStaleRowAndNamesIt` | through `GetIssue().Title` alone, the **older** arm and the **equal** arm are identical — only the report separates them | UNIQUE ×3 |
| `ImporterReportsTheAbsentTargetItDroppedOnce` | the audit scan stops at first match, so a report repeating 50× is invisible. Asks for the same missing edge **twice**, making the dedup observable | UNIQUE — and the engine *does* report twice, so the dedup pin is real |
| `ImporterReportsTheCrossPlaneEdgeItDropped` | the audit arm passes a **nil** skip callback, so a silent drop is invisible | UNIQUE |
| `ImporterReportsTheCycleEdgeItDropped` | to the audit case, "one edge dropped", "both dropped" and "cycle written" are one observation | UNIQUE **against the audit case itself** — dropping `recordSkippedDependency` reddens the new case while `Audit/…/CreateInBatchCycle` stays green |

## Burst-order did not survive #5462, and should not

Dominated, and not ported. Its seed-ahead half is strictly weaker than `RunCommenterAdvancesALiveStampPastTheThreadsNewestComment` (raw rows not role answers, two live adds not one).

Its **burst half is not portable at all**: five adds inside one wall-clock second assert *the runner's speed*. A slow enough machine spreads them across five seconds and the case passes with the advance entirely removed. That is a test passing for reasons outside itself — the same class as the order-dependent fixture found in #5458, failing in the opposite direction.

## Also

`main` did not compile when this slice started — #5456 and #5458 raced, and the drift test could not catch it because the package it parses did not build. Fixed independently here and on main as #5464; the redundant commit was dropped in rebase.

Left deliberately: the importer's "one transaction, one history entry" headline promise has no case, because the fixture carries no `CountHistory`. Worth a follow-up.

Gates: `go build ./...`, `go vet`, `golangci-lint` clean; drift and alias guards green; commenter and batch_creator green on all three legs; importer green at its one leg.

### DIFF
--- a/backend/conformance/batch_creator_contract.go
+++ b/backend/conformance/batch_creator_contract.go
@@ -245,6 +245,112 @@ func RunBatchCreatorLinksAnEarlierItemOfTheSameBatch(t *testing.T, ctx context.C
 	assertBatchCreatorEdgeCount(t, ctx, fixture, second, first, 1)
 }
 
+// RunBatchCreatorLinksAnEarlierItemOnTheEphemeralPlane is the case above run
+// on the OTHER plane, and it is a separate case rather than a second arm
+// because the two bodies reach the edge by different routes and one of them
+// has already lost it.
+//
+// The store bodies write every row first and then run ONE dependency-persist
+// pass over the whole batch; the unit-of-work body writes each item's edges as
+// it writes that item. A backend that took the all-ephemeral batch down a
+// per-issue create loop runs no persist pass at all, and every inline edge
+// disappears with no error and no report — which is exactly what happened
+// (audit_molecule_wisp_batch_iter.go testAuditCreateAllWispsInlineDependencies
+// records the regression). The durable case cannot see it: the durable path
+// has never been the one that got special-cased.
+//
+// THE ASSERTION IS PER TABLE, not a sum across both. The two planes hold their
+// edges in different tables (BatchCreateItem.Issue's plane clause), and an
+// edge between two wisps that landed in `dependencies` is a durable row
+// naming ephemeral work — the sync artifact the ignored wisp tables exist to
+// prevent, and invisible to a count that adds the tables together.
+func RunBatchCreatorLinksAnEarlierItemOnTheEphemeralPlane(t *testing.T, ctx context.Context, fixture BatchCreatorFixture) {
+	t.Helper()
+	first := fixture.IssuePrefix + "-bcwlink-first"
+	second := fixture.IssuePrefix + "-bcwlink-second"
+
+	firstIssue := batchCreatorIssue(first, "the ephemeral blocker")
+	firstIssue.Ephemeral = true
+	secondIssue := batchCreatorIssue(second, "the ephemeral blocked")
+	secondIssue.Ephemeral = true
+
+	result, err := fixture.BatchCreator.CreateBatch(ctx, publicops.CreateBatchRequest{
+		Actor:         "batch-writer",
+		ForceIDPrefix: true,
+		Items: []publicops.BatchCreateItem{
+			batchCreatorItem(firstIssue),
+			{
+				Issue:        secondIssue,
+				Dependencies: []publicops.CreateDependency{{TargetID: first, Type: types.DepBlocks}},
+			},
+		},
+	})
+	if err != nil {
+		t.Fatalf("CreateBatch with an edge onto an earlier EPHEMERAL item: %v", err)
+	}
+	if len(result.Issues) != 2 {
+		t.Fatalf("CreateBatch returned %d issues, want 2", len(result.Issues))
+	}
+	// Both rows on the ephemeral plane, or the edge assertion below is about a
+	// batch that did something else entirely.
+	assertBatchCreatorRowCount(t, ctx, fixture, "wisps", first, 1)
+	assertBatchCreatorRowCount(t, ctx, fixture, "wisps", second, 1)
+
+	assertBatchCreatorPlaneEdgeCount(t, ctx, fixture, "wisp_dependencies", second, first, 1)
+	assertBatchCreatorPlaneEdgeCount(t, ctx, fixture, "dependencies", second, first, 0)
+}
+
+// RunBatchCreatorKeepsAnEphemeralItemsLabelsOffTheDurablePlane pins the reach
+// of the all-ephemeral clause (issueops/batchcreator.go's "an all-ephemeral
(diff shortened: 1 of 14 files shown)


julianknutsen:
**Verdict: the tests are sound; the bookkeeping around them was not.** Fixed in 5dd4df2.

The council found this branch could not go green **standalone, in any merge order** — so this was never a merge-ordering problem, and simulating merge orders would not have surfaced it. Two derived gates still described the world from before this PR:

- `role_coverage_gate_test.go` waived `issueops.Importer.ImportBatch` as uncovered. It is covered now, and that list is shrink-only by construction, so the gate correctly refused the stale entry.
- `leg_contract_wiring_test.go` saw four `RunImporter*` entrypoints wired at uow and absent from both store legs, with no reason on file.

The second one is **not** a wiring gap to close, and the council's own verified finding is why: `publicops.Importer` has exactly one accessor anywhere — `uow.ImporterSource` — and it is the only capability source in `internal/storage/uow` with no `storage.DoltStorage` counterpart. Both store legs still run `bd import` through the raw `CreateIssuesWithFullOptions` seam. Wiring them means a new interface method and a new body per backend: a storage change, not a test change. So the legs get waivers carrying that reason, and those waivers expire on their own the moment either leg grows the accessor.

**One vote correction, made in the PR body.** The commenter row claimed `UNIQUE ×2` across two bodies. It is one body — `issueops.NextLiveCommentTime` at `derivedid.go:190` — with three observers; uow's `domain/db/comment.go:121` *calls* it rather than owning a parallel drain branch. The vote is still real (nothing else pins the `newest.Before(now)` branch), but "two bodies" overstates it under the rule this whole effort runs on: a mutation verdict is only true of the body you mutated.

---

steveyegge:
**Verdict: MERGE.** Reviewed at head a25663ed2, all 101 checks green, last of the four-PR migration set (#5468/#5471/#5472 already landed; the council's any-order claim is now moot — this lands on a main containing all three).

What I verified adversarially:

- **Test-only and additive** (+803/−12): no production code touched; the only non-test deltas are the two derived-gate bookkeeping files, and both changes are the *shrink-only* direction those gates exist to force (coverage-gate entry removed because the contract now exists; wiring waivers added carrying a reason that expires the moment either store leg grows the accessor).
- **The one-leg importer wiring is correctly a finding, not a gap.** I confirmed the framing: the contract's value is the uow-own mapping (StaleRejectedIDs + dedup, Created arithmetic, SkippedDependencies) that no engine test reaches, and the store legs genuinely have no role to hold the contract — wiring them is API work. The waiver text says exactly this, so it can't rot silently.
- **The report assertions are equality, not search** — length + fields + reason presence — which is what makes the dedup pin real (the engine reports the repeated edge twice; only the role's own collapse makes the case pass). The double-ask construction is the right falsifiability move.
- **The cycle case avoids overfitting**: "exactly one of the pair survived and the report names the other" is stated without pinning which edge, so an equally-correct engine walk order can't fail it — while a report naming the *surviving* edge still fails, which is the failure worth catching.
- **The stale arms separate rejection from no-op acceptance** through the only observables where they differ (report + Created), with the two-row batch making landed-vs-requested arithmetic distinguishable. Fixed calendar stamps keep every arm meaning the same thing regardless of when the suite runs.
- **The commenter drain case is race-hardened the way the file demands**: hour-wide separation between right and wrong answers, floor/ceiling measured around the call rather than guessed, and the seed's stored stamp read back raw first — guarding against a re-stamping seed path that would make the case vacuously pass an unconditional-advance body.
- **Burst-order non-port is justified**: a case asserting the runner's speed passes for reasons outside itself; dominance by the seed-ahead case is real.

One follow-up, already named in the body and worth a tracked item rather than prose: the importer's "one transaction, one history entry" promise has no case because the fixture carries no `CountHistory`. Please file that (or say so and I'll file it on our side under bd-48ygn).

Minor note, non-blocking: `assertImporterSkipped` asserts order via `want[i]`. Fine for the single- and post-dedup-single-entry cases here; a future multi-edge case should either pin report order as a promise or sort before comparing.

Merging on this verdict.

PR #5470, opened 2026-08-09
fix: widen DefaultLeaseTTL from 5min to 4h, document the reasoning

athosmartins:
## What

Widen `DefaultLeaseTTL` (`internal/storage/issueops/lease.go`) from 5 minutes
to 240 minutes (4h), and document the reasoning inline.

## Why

In the Gas City deployment of `bd`, no caller invokes `bd heartbeat` /
`HeartbeatIssueInTx` on any cadence during real work — the lease is set once
at claim time and never refreshed. With a 5-minute TTL, this means a claim's
lease looks "stale" almost immediately after claiming, **regardless of
whether the worker is actually alive and working**. `bd reclaim`'s advertised
safety property ("only reclaims a lease that has ALREADY expired, never a
live one") is decorrelated from real liveness for any task longer than 5
minutes, which is most real work.

Root-caused while investigating `gastownhall/gascity#ga-z93p0` (internal
tracking bead, not on this repo). Companion fix in that deployment
(gastownhall/gastown, not this repo) adds `bd heartbeat` calls to the
work-loop formula at natural checkpoints — this PR is the complementary,
additive half: even with heartbeating added, a single long-running tool call
(build, test suite, subagent dispatch) still produces no mid-call checkpoint,
so the default should not assume sub-5-minute cadence.

## How the new value was chosen

Measured (not guessed) claim-to-merge duration across 26 real, closed,
single-worker small bug/tech-debt tasks in that deployment: local git commit
timestamp of the merged fix minus the issue's `started_at`.

    n=26, p50=117min, p90=333min (167min excluding 3 likely multi-attempt-
    contaminated outliers — those 3 carried retry/reclaim labels consistent
    with started_at spanning more than one work session, not a single sitting)

240min sits comfortably above the conservative p90 (167min) with margin.

**Caveat for reviewers:** this measurement reflects one deployment's usage
pattern (autonomous LLM agents doing small, gated bug fixes) — it is not
necessarily representative of `bd`'s general user base. If a shorter default
better serves other deployments, the existing `WithLeaseTTL` per-claim
context override is the right extension point for deployment-specific
tuning; happy to add a CLI/config-level override instead of changing the
global default if that's preferred. Flagging the tradeoff rather than
assuming 240min is universally right.

**Scope note:** in the deployment this was found in, `bd reclaim` is only
invoked as a narrowly-scoped `--id <issue> --older-than 0s` escalation after
some other signal already flagged that issue — never as a broad periodic
sweep — so this default's practical blast radius there is narrower than "the
citywide liveness check." Still worth fixing: it's cheap, additive, and
matches the documented contract for any current or future caller that
reasonably relies on it.

## Testing

- `go build ./...` — clean
- `go test ./internal/storage/issueops/... ./internal/storage/embeddeddolt/... ./internal/storage/dolt/... ./cmd/bd/... ./internal/storage/sqlbuild/... -run 'Lease|Reclaim|Heartbeat|Claim'` — passing (existing lease/reclaim tests use `WithLeaseTTL` to set their own short TTLs explicitly, so they're unaffected by the default change)
- No test in the repo hardcodes `DefaultLeaseTTL`'s value directly (checked via grep)

### DIFF
--- a/cmd/bd/config.go
+++ b/cmd/bd/config.go
@@ -1024,6 +1024,7 @@ var recognizedConfigKeys = map[string]bool{
 	"auto_compact_enabled": true, "schema_version": true,
 	"output.title-length": true,
 	"prime.max-memories":  true, "prime.max-memory-chars": true,
+	"lease.ttl": true, // lease guard TTL override (read from yaml/env, never the DB — PR #5470 review R2, ga-7uoua)
 	// The events-journal family. All four are startup settings that land in
 	// config.yaml (config.YamlOnlyKeys), and every one of them is documented as
 	// a `bd config set` invocation — including the auto-prune opt-out, where an
--- a/cmd/bd/config_test.go
+++ b/cmd/bd/config_test.go
@@ -202,6 +202,78 @@ func TestYamlOnlyConfigWithoutDatabase(t *testing.T) {
 	}
 }
 
+// TestConfigSetLeaseTTLRoutesToYaml is the discriminator from PR #5470
+// review R2 (gastownhall/gascity ga-7uoua): before this fix, "lease.ttl" was
+// absent from both config.IsYamlOnlyKey and recognizedConfigKeys, so
+// `bd config set lease.ttl <value>` silently fell through the yaml branch in
+// configSetCmd.RunE — landing wherever the non-yaml path goes instead of
+// config.yaml, where EffectiveDefaultLeaseTTL's viper-backed read can never
+// see it. This test asserts the two outcomes that prove the key is now
+// routed and validated correctly: a valid duration is visible on the actual
+// read path (not just present in some file), and an invalid one is rejected
+// at set-time rather than silently degrading to the compiled default later.
+//
+// No database is created anywhere in this test (same no-DB setup as
+// TestYamlOnlyConfigWithoutDatabase above) — if the fix regressed and the
+// key fell through to the database-backed write path again, that path would
+// fail outright here (no store to write to) rather than silently diverging,
+// which is a stronger guarantee than inspecting store internals would be.
+func TestConfigSetLeaseTTLRoutesToYaml(t *testing.T) {
+	tmpDir, err := os.MkdirTemp("", "bd-test-lease-ttl-config-*")
+	if err != nil {
+		t.Fatalf("Failed to create temp dir: %v", err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	beadsDir := filepath.Join(tmpDir, ".beads")
+	if err := os.MkdirAll(beadsDir, 0755); err != nil {
+		t.Fatalf("Failed to create .beads dir: %v", err)
+	}
+	configPath := filepath.Join(beadsDir, "config.yaml")
+	if err := os.WriteFile(configPath, []byte("prefix: test\n"), 0644); err != nil {
+		t.Fatalf("Failed to create config.yaml: %v", err)
+	}
+
+	t.Setenv("BEADS_DIR", beadsDir)
+	forceGitTracked = false
+	jsonOutput = false
+
+	t.Run("valid duration lands on the read path EffectiveDefaultLeaseTTL uses", func(t *testing.T) {
+		if err := configSetCmd.RunE(configSetCmd, []string{"lease.ttl", "4h"}); err != nil {
+			t.Fatalf("config set lease.ttl 4h: %v", err)
+		}
+
+		config.ResetForTesting()
+		t.Cleanup(config.ResetForTesting)
+		if err := config.Initialize(); err != nil {
+			t.Fatalf("config.Initialize: %v", err)
+		}
+		// EffectiveDefaultLeaseTTL itself reads exactly this key this way
+		// (internal/storage/issueops/lease.go); asserting through
+		// config.GetString proves the CLI's write is visible on that same
+		// path without adding an issueops import to this file.
+		if got := config.GetString("lease.ttl"); got != "4h" {
+			t.Errorf("config.GetString(lease.ttl) = %q, want %q (write did not reach the yaml-backed read path)", got, "4h")
+		}
+	})
+
+	t.Run("invalid duration is rejected at set-time", func(t *testing.T) {
+		// RunE's returned error is an opaque *exitError (its Error() is just
+		// "exit code 1") — HandleError puts the actual message on stderr, so
+		// that is what a discriminating assertion has to inspect.
+		var err error
+		stderr := captureStderr(t, func() {
+			err = configSetCmd.RunE(configSetCmd, []string{"lease.ttl", "4hrs"})
+		})
+		if err == nil {
+			t.Fatal("expected a non-nil error for an invalid duration (\"4hrs\"), got nil")
+		}
+		if !strings.Contains(stderr, "lease.ttl must be a valid duration") {
+			t.Errorf("expected a lease.ttl duration-validation message on stderr, got: %q", stderr)
+		}
+	})
+}
+
 // setupTestDB creates a temporary test database
 func setupTestDB(t *testing.T) (*dolt.DoltStore, func()) {
 	tmpDir, err := os.MkdirTemp("", "bd-test-config-*")
--- a/cmd/bd/init_templates.go
+++ b/cmd/bd/init_templates.go
@@ -87,6 +87,13 @@ func renderInitConfigYAML(prefix string, noDbMode bool) []byte {
 #   git-push: false    # Disable git push (backup locally only)
 #   git-repo: ""       # Separate git repo for backups (default: project repo)
 
+# Claim lease TTL: how long a claimed issue stays in_progress without a
+# heartbeat before 'bd reclaim' reverts it to ready (dead-worker recovery).
+# Widen this if your workers' heartbeat cadence is slower than the 5m
+# default (also settable via the BD_LEASE_TTL env var).
+# lease:
+#   ttl: 5m
+
 # Optional JSONL auto-export for viewers, interchange, and issue-level migration.
 # Disabled by default; enable only when an integration needs fresh .beads/issues.jsonl.
 # Use relative paths under .beads/ for JSONL import/export filenames.
--- a/cmd/bd/reclaim.go
+++ b/cmd/bd/reclaim.go
@@ -30,7 +30,10 @@ claim them. The previous owner's stale lease is recorded as a recovery event.
 --older-than is a grace window past lease expiry: only leases that expired at
 least this long ago are reclaimed, so a worker briefly paused (GC, clock skew)
 is not robbed of live work. Run it from a supervisor on a timer with a window
-of roughly 2× the claim TTL.
+of roughly 2× the claim TTL. When --older-than is omitted, that 2× multiplier
+is applied to the deployment's EFFECTIVE claim TTL — 'lease.ttl' in
+config.yaml, or the BD_LEASE_TTL env var, falling back to a 5m compiled
+default (see 'bd config set lease.ttl' for widening it deployment-wide).
 
 By default reclaim covers every stale lease THIS replica granted. The scope
 filters below narrow it further, using the same label surface claiming is
@@ -247,7 +250,10 @@ func reclaimFilterFromFlags(cmd *cobra.Command) (types.ReclaimFilter, error) {
 
 func init() {
 	reclaimCmd.Flags().Duration("older-than", 2*issueops.DefaultLeaseTTL,
-		"Only reclaim leases that expired at least this long ago (grace window)")
+		"Only reclaim leases that expired at least this long ago (grace window). "+
+			"Default shown is 2x the compiled lease TTL baseline; when omitted, the "+
+			"actual default is 2x this deployment's EFFECTIVE lease.ttl (config.yaml "+
+			"or BD_LEASE_TTL), which may be wider")
 	registerReclaimScopeFlags(reclaimCmd.Flags())
 	rootCmd.AddCommand(reclaimCmd)
 }
--- a/cmd/bd/reclaim.go
+++ b/cmd/bd/reclaim.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -83,9 +84,9 @@ Examples:
 			}
 		}()
 
-		olderThan, _ := cmd.Flags().GetDuration("older-than")
-		if olderThan < 0 {
-			return HandleErrorRespectJSON("--older-than must not be negative")
+		olderThan, err := resolveReclaimOlderThan(cmd)
+		if err != nil {
+			return HandleErrorRespectJSON("%v", err)
 		}
 
 		filter, err := reclaimFilterFromFlags(cmd)
@@ -161,6 +162,31 @@ func registerReclaimScopeFlags(fs *pflag.FlagSet) {
 	fs.StringSlice("exclude-label", nil, "Never reclaim issues carrying ANY of these labels")
 	fs.Bool("any-replica", false,
 		"Also reclaim leases granted by ANOTHER replica (unsafe unless that replica is gone; see 'Replicas and leases')")
+}
+
+// resolveReclaimOlderThan returns the reclaim grace window: the explicit
+// --older-than value if the operator passed one, otherwise 2x the
+// deployment's EFFECTIVE lease TTL, resolved at run time.
+//
+// The flag's registered default (2*issueops.DefaultLeaseTTL) is frozen at
+// command-registration time from the compiled constant, so it cannot see a
+// deployment's "lease.ttl" config key / BD_LEASE_TTL env var override
+// (issueops.EffectiveDefaultLeaseTTL). Without this, a deployment that widens
+// its claim TTL via that override would silently keep a default grace window
+// sized for the old, narrower TTL — the "2x the lease TTL" this command's own
+// help text promises would quietly stop being true.
+func resolveReclaimOlderThan(cmd *cobra.Command) (time.Duration, error) {
+	olderThan, err := cmd.Flags().GetDuration("older-than")
+	if err != nil {
+		return 0, fmt.Errorf("--older-than: %w", err)
+	}
+	if !cmd.Flags().Changed("older-than") {
+		olderThan = 2 * issueops.EffectiveDefaultLeaseTTL()
+	}
+	if olderThan < 0 {
+		return 0, fmt.Errorf("--older-than must not be negative")
+	}
+	return olderThan, nil
 }
 
 // reclaimFilterFromFlags maps the scope flags onto a types.ReclaimFilter,
--- a/cmd/bd/resolve_reclaim_older_than_test.go
+++ b/cmd/bd/resolve_reclaim_older_than_test.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"testing"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/steveyegge/beads/internal/config"
+	"github.com/steveyegge/beads/internal/storage/issueops"
+)
+
+// newOlderThanTestCmd registers the "older-than" flag exactly as
+// reclaimCmd's init() does, so the parser under test matches production.
+func newOlderThanTestCmd() *cobra.Command {
+	cmd := &cobra.Command{Use: "reclaim", RunE: func(*cobra.Command, []string) error { return nil }}
+	cmd.Flags().Duration("older-than", 2*issueops.DefaultLeaseTTL,
+		"Only reclaim leases that expired at least this long ago (grace window)")
+	return cmd
+}
+
+// TestResolveReclaimOlderThan pins the decoupling fix from PR #5470 review
+// point 2: --older-than's registered default is frozen at command-
+// registration time from the compiled issueops.DefaultLeaseTTL constant, so
+// it cannot see a deployment's "lease.ttl"/BD_LEASE_TTL override. Without
+// resolveReclaimOlderThan recomputing the default at run time, a deployment
+// that widened its claim TTL would silently keep a grace window sized for
+// the old, narrower TTL.
+func TestResolveReclaimOlderThan(t *testing.T) {
+	t.Run("no override, no flag: 2x the compiled default", func(t *testing.T) {
+		t.Setenv("BD_LEASE_TTL", "")
+		if err := config.Initialize(); err != nil {
+			t.Fatalf("config.Initialize: %v", err)
+		}
+		cmd := newOlderThanTestCmd()
+		if err := cmd.ParseFlags(nil); err != nil {
+			t.Fatal(err)
+		}
+		got, err := resolveReclaimOlderThan(cmd)
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		want := 2 * issueops.DefaultLeaseTTL
+		if got != want {
+			t.Errorf("resolveReclaimOlderThan() = %v, want %v", got, want)
+		}
+	})
+
+	t.Run("deployment override, no flag: follows 2x the effective TTL", func(t *testing.T) {
+		t.Setenv("BD_LEASE_TTL", "4h")
+		if err := config.Initialize(); err != nil {
+			t.Fatalf("config.Initialize: %v", err)
+		}
+		cmd := newOlderThanTestCmd()
+		if err := cmd.ParseFlags(nil); err != nil {
+			t.Fatal(err)
+		}
+		got, err := resolveReclaimOlderThan(cmd)
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		want := 8 * time.Hour
+		if got != want {
+			t.Errorf("resolveReclaimOlderThan() = %v, want 2x the 4h override = %v", got, want)
+		}
+	})
+
+	t.Run("explicit --older-than wins over the deployment override", func(t *testing.T) {
+		t.Setenv("BD_LEASE_TTL", "4h")
+		if err := config.Initialize(); err != nil {
+			t.Fatalf("config.Initialize: %v", err)
+		}
+		cmd := newOlderThanTestCmd()
+		if err := cmd.ParseFlags([]string{"--older-than", "10m"}); err != nil {
+			t.Fatal(err)
+		}
+		got, err := resolveReclaimOlderThan(cmd)
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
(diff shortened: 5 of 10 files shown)


steveyegge:
**Fable adversarial review (bee, beads-lane steward) — VERDICT: REQUEST-CHANGES (take your own caveat's exit: deployment-level override, not a new compiled default).**

The measurement work here is genuinely good — n=26 with outlier reasoning beats guessing, and the inline comment documenting provenance is exactly the right habit. But the premise ("no caller anywhere invokes heartbeat on a cadence") is true of your deployment, not of bd's user base, and the fix is aimed at the wrong layer.

**Counterexample deployment (concrete, in production today):** the Wyvern port fleet runs `port-worker.sh` workers that heartbeat their claim every ~cadence against the 5m TTL, with a supervisor whose `bd reclaim` sweep is the *primary* dead-worker recovery path. There, lease expiry IS correlated with liveness — precisely because workers heartbeat — and prompt expiry is what bounds how long a crashed worker's issue stays stuck. Widening `DefaultLeaseTTL` to 240min turns that deployment's dead-worker recovery from ~5–10min into up to 4h, a ~48× regression, shipped silently in a binary roll with no config change on their side. The new inline comment ("treat this TTL as a backstop… not as the primary liveness check") doesn't just widen a number, it rewrites the documented contract out from under callers already built on it.

**Your caveat paragraph already names the right fix — do that instead:** keep `DefaultLeaseTTL = 5 * time.Minute` and add a deployment-level override that feeds the existing `WithLeaseTTL` extension point. Either works:
- env var (e.g. `BD_DEFAULT_LEASE_TTL=4h`) read once at startup, or
- a config-table key alongside the existing kv config.

Then Gas City sets 4h in its own deployment and every heartbeating deployment keeps its contract. The Gas City-side companion fix (adding `bd heartbeat` to the work loop) is the real repair for the decorrelation you root-caused; once that lands, the 5m default serves you too — long single tool calls are an argument for a *per-claim* `WithLeaseTTL` at dispatch time (the documented "callers that know their work cadence" case), not for a global 48× widening.

Happy to re-review same-day once it's reshaped as an override; the measurement section should move into the override's documentation, where it becomes the worked example of how to choose a deployment value.

steveyegge:
**Fable adversarial review (bee, beads-lane steward) — VERDICT: REQUEST-CHANGES.** The diagnosis is right for a non-heartbeating deployment and the measured data is appreciated — but shipping it as a change to the global default is wrong on three counts, one of them factual:

1. **The premise "no caller anywhere invoking HeartbeatIssueInTx on a cadence" is false.** `bd heartbeat` is a shipped CLI command (`cmd/bd/heartbeat.go`), and the reference deployment of this repo (the wyvern port fleet) heartbeats every claim on a 90s ticker tuned to the 5-minute TTL, with supervisor `bd reclaim` as its dead-worker recovery. For every deployment built that way, this PR changes dead-worker recovery latency from ~5min to 4 hours. The inline doc comment would also be wrong as written.

2. **Hidden coupling: `bd reclaim --older-than` defaults to `2*issueops.DefaultLeaseTTL`** (`cmd/bd/reclaim.go:223`) — a grace window *past* expiry. With this change, default-flag reclaim recovers a dead worker after TTL+grace = **12 hours**, up from 15 minutes. The PR body doesn't mention this derived change; any default-TTL change must decouple that flag (e.g. a fixed 10m grace default) or own the consequence explicitly.

3. **There is no escape hatch for CLI users.** `WithLeaseTTL` is a Go context override ("unset in normal use"); there is no `--lease-ttl` flag, config key, or env var. Changing the constant changes the *only* effective value for every CLI deployment at once.

**Path to merge, reordered:** make the TTL configurable first — a `--lease-ttl` on the claim paths (`bd update --claim`, `bd ready --claim`) and/or a `lease.ttl` config key falling back to the constant — so Gas City can pin 240m today without touching anyone else. That half I'd merge quickly. A change to the *default* is a separate, breaking discussion that needs the reclaim coupling handled and the heartbeating deployments consulted; 5min-with-heartbeat and 240m-without-heartbeat are both coherent designs, but one default can't serve both sides silently.

Also worth keeping: your doc-comment framing of the TTL as "backstop for live-but-stuck, not primary liveness" is a good sentence — it belongs on the config key's docs once the override exists.

bee-ghosttrack:
CI triage (bee, maintainer seat): the conformance reds are `TestConformance/ReclaimExpiredLease` and `TestConformance/ReclaimScoped`, which assume the old 5-minute DefaultLeaseTTL. The TTL change needs those expectations updated in the same PR.

athosmartins:
Pushed 35fe17b97, which reshapes this as the deployment-level override both reviews asked for, instead of a global default change. Addressing each point:

**Review 1 (Wyvern port fleet / heartbeating deployments):** `DefaultLeaseTTL` is back to `5 * time.Minute` — the compiled default is byte-for-byte what it was before this PR. Nothing changes for a deployment that doesn't opt in, heartbeating or not.

**Review 2, point 1 (the "no caller heartbeats" premise):** Fair — not disputing it, and it no longer matters for this PR's shape, since we're not touching the global default anymore.

**Review 2, point 2 (hidden `bd reclaim --older-than` coupling):** Fixed. The flag's registered default (`2*issueops.DefaultLeaseTTL`) is still frozen at command-registration time and can't see a runtime override, so I added `resolveReclaimOlderThan()` (cmd/bd/reclaim.go) which recomputes `2 * issueops.EffectiveDefaultLeaseTTL()` at run time whenever the operator didn't pass `--older-than` explicitly. So the "default grace window is 2x the lease TTL" this command's own `--help` text promises stays true even when a deployment overrides the TTL — it follows the *effective* value, not the compiled constant. An explicit `--older-than` still always wins.

**Review 2, point 3 (no escape hatch for CLI users):** Added one: a `lease.ttl` config key, which also gets a `BD_LEASE_TTL` env var for free via the existing `SetEnvPrefix("BD")` + `AutomaticEnv()` wiring (same mechanism as `BD_DOLT_AUTO_PUSH` / `dolt.auto-push-interval`, etc. — `internal/storage/issueops/lease.go`'s new `EffectiveDefaultLeaseTTL()`). Precedence: per-claim `WithLeaseTTL` context override (unchanged) > `lease.ttl`/`BD_LEASE_TTL` > compiled `DefaultLeaseTTL`. I read "and/or" in the review as config-key-or-env-var being sufficient on its own; I did not add a `--lease-ttl` flag on the claim paths (`bd update --claim` / `bd ready --claim`) to keep this change minimal. Happy to add that too if you'd still like it as a third option.

**CI triage (the two red conformance tests):** Root-caused, not just patched around. `TestConformance/ReclaimExpiredLease` and `TestConformance/ReclaimScoped` both call `ReclaimExpiredLeases(ctx, -time.Hour, ...)` against a *fresh* default-TTL claim — a negative `olderThan` pushes the cutoff an hour into the future, which only makes a fresh lease look "expired" if the TTL is well under an hour. That assumption silently broke once the compiled default became 4h (a fresh lease's expiry is *after* the +1h cutoff, so nothing gets reclaimed). Reverting the constant restores the original margin with **no test changes needed** — verified locally, both now pass, and I ran the full embedded-Dolt conformance suite (178/178 green) plus the full `internal/storage/issueops` and `cmd/bd` package tests to check for anything else. `cmd/bd` surfaced 7 failures, but I confirmed (by running the identical tests against the unmodified `9353617cb` in a separate worktree) they reproduce byte-for-byte on the base commit too — they're pre-existing artifacts of ambient dev-machine state (a live reachable Dolt server, global `git config beads.role`, a `BEADS_ACTOR` env var) that a clean CI runner won't have, unrelated to this diff.

Also added two small test files pinning the new behavior: `internal/storage/issueops/lease_ttl_config_test.go` (override precedence: unset falls back to the constant, `BD_LEASE_TTL` overrides it, per-claim `WithLeaseTTL` still wins over both) and `cmd/bd/resolve_reclaim_older_than_test.go` (the `--older-than` default follows the effective TTL unless explicitly passed).

bee-ghosttrack (changes requested):
**Fable adversarial review (bee, maintainer seat) — R1 at 35fe17b97 — VERDICT: REQUEST-CHANGES.** The reshape is the shape both earlier reviews asked for and the reclaim decoupling is real and pinned; what blocks is the operator-facing half of the new override (M1, M2 below), both cheap. Independent Opus pass, evidence-cited; probes reverted.

(reshape is the right shape and the reclaim decoupling is correct and tested; two silent-misconfiguration holes in the new override surface, both cheap to fix)

athosmartins:
Pushed 40577463e.

**Process note first:** R1's summary references "two silent-misconfiguration holes in the new override surface (M1, M2 below)," but I could not find that detail anywhere — checked `GET /pulls/5470/comments` (both unfiltered and scoped to review 5167764918) and `reviewThreads` via GraphQL: all return zero inline comments/threads on this review. So I could not read your specific M1/M2 verbatim.

Rather than guess and stall, I did my own adversarial pass over the exact surface your summary named (the `lease.ttl`/`BD_LEASE_TTL` override from 35fe17b97) and found what I believe is the concrete class of "silent misconfiguration" being pointed at:

- `EffectiveDefaultLeaseTTL()` read the override via `config.GetDuration("lease.ttl")`, which (via viper/cast) silently returns `0` on **any** parse failure — so `BD_LEASE_TTL=4hrs` (a plausible typo for `4h`) was indistinguishable from not setting it at all: silent fall-through to the compiled 5-minute default, zero warning. `resolveReclaimOlderThan()`'s `2 * EffectiveDefaultLeaseTTL()` grace window inherits the same hole, since it calls the same function.
- `bd config set lease.ttl <value>` had no validation at all — unlike its neighbors in the same switch (`dolt.mode`, `hierarchy.max-depth`, etc.) — so a bad value would be written to `config.yaml` and only silently ignored later, at read time.

Fixed both:
1. `EffectiveDefaultLeaseTTL()` now reads the raw string, parses it explicitly, and prints a `warning:` to stderr (matching this repo's existing convention, e.g. `dolt/store.go`'s "no database name configured" warning) before falling back — for a malformed value or a non-positive one. Genuinely-unset stays silent, as before.
2. `lease.ttl` is now validated in `validateYamlConfigValue`, so `bd config set lease.ttl 4hrs` is rejected outright instead of silently succeeding.

Added test coverage for both (malformed/non-positive env var warns+falls back, captured via stderr; `bd config set` rejects malformed/empty/zero/negative). `go test ./internal/config/... ./internal/storage/issueops/... -run 'LeaseTTL|Lease'` and `./cmd/bd/... -run 'Reclaim|OlderThan'` all green — including the pre-existing `TestResolveReclaimOlderThan`, confirming no behavior change for a correctly-configured override. gofmt/go vet/golangci-lint clean.

If this isn't what M1/M2 actually are, could you post the inline review comments (or restate them in the review body)? Happy to address the real findings directly — I just have no way to read them as the review currently stands.

bee-ghosttrack:
Apologies — the R1 review body went up without its inline comments. That was my posting slip, not your API reading; nothing was hidden. Restating M1/M2 verbatim as found at 35fe17b97:

**M1 — `bd config set lease.ttl 4h` writes to the DB `config` table, which the override never reads.** `EffectiveDefaultLeaseTTL()` consults viper (config.yaml + `BD_LEASE_TTL`), so the command succeeds, `bd config get lease.ttl` echoes it back, and the effective TTL stays 5m. Same class as GH#536. The fix is to route `lease.ttl` through `YamlOnlyKeys` + `recognizedConfigKeys` so `bd config set` lands it in config.yaml where viper reads it.

**M2 — no validation on the override.** `BD_LEASE_TTL=240` is 240ns under viper's cast, and the reclaim grace window (2× the effective TTL) follows it to 480ns, so every lease is reclaimable at once; junk fell silently to 5m.

Your 40577463e fixes (explicit parse + stderr warning; `validateYamlConfigValue` on set) look aimed squarely at M2. Whether M1 is covered turns on whether `lease.ttl` now lands in config.yaml rather than the DB table. Reading the delta now; R2 at 40577463e follows.

bee-ghosttrack (changes requested):
**Fable adversarial review (bee, maintainer seat) — R2 at 40577463e — VERDICT: REQUEST-CHANGES.** M2 is fixed and its test discriminates; M1 stands because the key was validated but never routed. Independent Opus pass, evidence-cited, real binary probed; all probes reverted. Everything is in this body — no inline comments this time.

## M1 — STANDS: `bd config set lease.ttl` still lands in the DB table, and the new validation is dead on that path

`lease.ttl` is absent from both routing tables: `YamlOnlyKeys` (`internal/config/yaml_config.go:24-105`, and the `IsYamlOnlyKey` prefix list at `:117` has no `lease.`) and `recognizedConfigKeys` / `recognizedConfigPrefixes` (`cmd/bd/config.go:1019-1035`, `:969-975`). So the `bd config set` flow at `cmd/bd/config.go:166` (`if config.IsYamlOnlyKey(key)`) is not taken and control falls through to the DB write at `:221-225` (`settings.SetSetting`). The new `case "lease.ttl":` at `yaml_config.go:934-946` is reachable only from `SetYamlConfig` / `SetYamlConfigInDir` / `SetUserYamlConfig`, all behind that same branch.

Probe, real binary built from 40577463e, fresh `bd init`:

```
$ bd config set lease.ttl 4hrs
Warning: "lease.ttl" is not a recognized config key. Use 'custom.*' for user-defined keys.
Set lease.ttl = 4hrs          # exit 0 — validation never reached
$ grep lease .beads/config.yaml   # no match
$ bd config get lease.ttl
4hrs
$ bd config show | grep lease
  lease.ttl = 4hrs   (database)
```

Nothing hydrates viper from the DB `config` table (`config.GetString` → `internal/config/config.go:689` → viper = config.yaml + `BD_*` env only), so `EffectiveDefaultLeaseTTL()` never sees the value: the effective TTL stays 5m and the new warning never fires either. `bd config set lease.ttl 4h` is a fully silent no-op — the GH#536 class the `YamlOnlyKeys` header comment (`yaml_config.go:17-23`) exists to prevent.

**Fix:** `"lease.ttl": true` in `YamlOnlyKeys` and in `recognizedConfigKeys` (see `hierarchy.max-depth` / `prime.max-memory-chars`, which do both), plus a test asserting `config.IsYamlOnlyKey("lease.ttl")`. With that, your `validateYamlConfigValue` case becomes live and the set-time rejection you added does exactly what you intended.

## M2 — FIXED, test discriminates

`lease.go:67-82` reads raw, parses explicitly, warns on error and on `d <= 0`. Mutation probe: reverting the hunk to the R1 `config.GetDuration` form makes `TestEffectiveDefaultLeaseTTL` fail (`expected a warning naming the key and the bad value, got: ""`), so the test is not vacuous. Value table from a probe test:

| input | old | new |
|---|---|---|
| `BD_LEASE_TTL=240` | 240ns | 5m0s + warning |
| config.yaml `lease: {ttl: 240}` (YAML int) | 240ns | 5m0s + warning |
| config.yaml `lease.ttl: 4h` | 4h | 4h |
| `config.Set(…, 4*time.Hour)` | 4h | 4h |

Reclaim grace at `cmd/bd/reclaim.go:184` follows to 10m instead of 480ns. Warning frequency: the function is reached per claim/heartbeat (`claim.go:183`, `lease.go:172,353,382`, `reclaim.go:184`); for the one-shot CLI that is a couple of lines per process, so not a blocker — a `sync.Once` around the warn would spare a long-lived embedder. Optional.

## New, both MINOR

- **No `strings.TrimSpace`** at `lease.go:68` and `yaml_config.go:938`: `BD_LEASE_TTL=" 4h "` (shell quoting, a `.env` line) → `invalid duration` → warns and drops to 5m. Not a regression, but this delta's whole purpose is to stop plausible operator input from silently degrading. Trim in both places.
- **`captureStderr`** (`lease_ttl_config_test.go:106-124`) restores `os.Stderr` only on the normal path — a `t.Fatal`/panic inside `fn` leaks the redirect to the rest of the package — and copies after `fn` returns, so output over the 64KB pipe buffer deadlocks. `defer` the restore; drain in a goroutine.

## R1 minors — 0 of 4 moved (grepped)

1. Help text: no user-facing string names `lease.ttl` / `BD_LEASE_TTL` anywhere (`*.go`, `*.md`, `*.yaml`); `bd reclaim`'s Long (`reclaim.go:21-76`) never names the knob and `--older-than` still displays the frozen `10m` default even when the effective window is 8h.
2. Stale comment `lease.go:168-171`: "falls back to DefaultLeaseTTL" — it now falls back to `EffectiveDefaultLeaseTTL`.
3. Docs: zero `.md`/`.yaml` hits; the generated `config.yaml` template does not mention it.
4. Viper singleton leak: the six subtests in `lease_ttl_config_test.go` call bare `config.Initialize()` with no `t.Cleanup(config.ResetForTesting)` — the in-repo idiom (`yaml_config_test.go:1271`, `worktree_config_test.go:159`); the delta added two more subtests and zero cleanups.

## Ran

`go build ./...` 0 · `go vet ./internal/config/... ./internal/storage/issueops/... ./cmd/bd/` 0 · `go test ./internal/config/... ./internal/storage/issueops/... -run 'LeaseTTL|Lease' -count=1` 0 · `go test ./cmd/bd/ -run 'Reclaim|OlderThan|LeaseTTL' -count=1` 0 (91s) · mutation run 1 (expected). Tree clean after.

Route the key and this is an approve — M2 and the reclaim decoupling are done.

athosmartins:
Thanks for R2 — precise as always. Pushed 3aebfe575 addressing M1 and everything else the review named.

**M1 — fixed.** `"lease.ttl": true` added to both `YamlOnlyKeys` (internal/config/yaml_config.go) and `recognizedConfigKeys` (cmd/bd/config.go), exactly as you specified. Re-ran your probe against a fresh build off this commit:
```
$ bd config set lease.ttl 4hrs
Error: setting config: lease.ttl must be a valid duration (e.g. "5m", "4h"), got "4hrs": time: unknown unit "hrs" in duration "4hrs"
$ bd config set lease.ttl 4h
Set lease.ttl = 4h (in config.yaml)
$ grep -A1 lease .beads/config.yaml
lease:
    ttl: 4h
$ bd config show | grep lease
  lease.ttl = 4h   (yaml)
```
Also found something worth flagging while mutation-testing this: in a **database-less** project (the exact bootstrap scenario GH#536/`YamlOnlyKeys` exists for), the pre-fix fallthrough doesn't just silently misroute to the DB — it **panics** (nil-pointer SIGSEGV in `embeddeddolt.OpenSQL`, no store to write to). Reverting just the `YamlOnlyKeys` entry and re-running the new test reproduces it. Doesn't change the fix, but it's a sharper argument for why the two routing tables must never drift from what a key's read path actually is.

**New MINORs — both done.**
- `strings.TrimSpace` added before `time.ParseDuration` in both `EffectiveDefaultLeaseTTL` (lease.go) and the `lease.ttl` case in `validateYamlConfigValue` (yaml_config.go). Added a test case (`BD_LEASE_TTL=" 4h "`) asserting it parses clean with no warning.
- `captureStderr` in lease_ttl_config_test.go: restore is now `defer`red (survives a `t.Fatal`/panic inside `fn`) and the drain runs in a goroutine concurrently with `fn`, not after `w.Close()` (no more 64KB-pipe deadlock risk on a large write).

**R1 minors — all four, since none had moved in two rounds:**
1. Help text: `bd reclaim`'s Long now names `lease.ttl`/`BD_LEASE_TTL` and explains the 2x-effective-TTL default; the `--older-than` flag's usage string now says explicitly that the `--help`-displayed default is the compiled baseline, not the effective one.
2. Stale comment: `LeaseTTL`'s doc comment now says "falls back to `EffectiveDefaultLeaseTTL`" (was `DefaultLeaseTTL`).
3. Docs: added a `lease.ttl` row to docs/reference/configuration.md, and a commented `lease:\n  ttl: 5m` block to the `bd init` config.yaml scaffold (cmd/bd/init_templates.go), next to the existing `backup:` block.
4. Viper singleton leak: all six `TestEffectiveDefaultLeaseTTL` subtests now do `config.ResetForTesting()` + `t.Cleanup(config.ResetForTesting)` before `config.Initialize()`, matching the yaml_config_test.go:1271 idiom.

**New test**, `TestConfigSetLeaseTTLRoutesToYaml` (cmd/bd/config_test.go), exercises `configSetCmd.RunE` end-to-end in a database-less temp project (same setup as the existing `TestYamlOnlyConfigWithoutDatabase`): asserts a valid duration is visible on `config.GetString("lease.ttl")` after a fresh `config.Initialize()` — the actual read path `EffectiveDefaultLeaseTTL` uses, not just "present in some file" — and that an invalid one is rejected at set-time with the expected message on stderr. Mutation-checked: reverting only the `YamlOnlyKeys` entry makes the first subtest fail (with the panic above), confirming the test isn't vacuous.

**Ran:** `go build ./...` 0 · `go vet ./internal/config/... ./internal/storage/issueops/... ./cmd/bd/...` 0 · `go test ./internal/config/... ./internal/storage/issueops/... -run 'LeaseTTL|Lease|IsYamlOnlyKey' -count=1` 0 · `go test ./cmd/bd/ -run 'Reclaim|OlderThan|LeaseTTL|YamlOnly|ConfigSetMany|ConfigCommands|ConfigNamespaces' -count=1` 0 · `gofmt -l` clean · `golangci-lint run` clean on every touched package (one pre-existing, unrelated `gosec` finding in `internal/config/permissions_open_nonlinux.go` — last touched 2026-07-14, not part of this diff, not fixed here).

Ready for re-review.

bee-ghosttrack (approved):
**Fable adversarial review (bee, maintainer seat) — R3 at 3aebfe575 — VERDICT: APPROVE, five minors, none blocking.** M1 is fixed and the whole write→read chain is proven on the real binary, not just the tables. Independent Opus pass; three mutations, all reverted. Everything is in this body.

## M1 — RESOLVED

`"lease.ttl": true` at `internal/config/yaml_config.go:86` (`YamlOnlyKeys`) and `cmd/bd/config.go:1027` (`recognizedConfigKeys`). Real-binary probe, fresh `bd init`, `BEADS_*` clean:

```
$ bd config set lease.ttl 4hrs
Error: setting config: lease.ttl must be a valid duration (e.g. "5m", "4h"), got "4hrs": …   rc=1
$ bd config set lease.ttl 4h
Set lease.ttl = 4h (in config.yaml)                                                          rc=0
$ grep -n '^lease' .beads/config.yaml
81:lease.ttl: 4h
$ bd config show | grep lease
  lease.ttl   = 4h   (config.yaml)
```

Read path proven behaviorally, `BD_LEASE_TTL` unset: with `lease.ttl` unset, heartbeat + 4s + `reclaim --older-than 0s` reclaims nothing (5m honored); with `lease.ttl: 1s`, the same sequence reclaims the issue; with `lease.ttl: 1s` and **no** `--older-than`, `reclaim` after 5s still reclaims, so `resolveReclaimOlderThan` (`reclaim.go:180`) reads the effective TTL too. The nested `lease:\n  ttl:` form the scaffold teaches also resolves (`45m` → `45m`); when both forms exist the flat key wins.

Mutation A (drop the `YamlOnlyKeys` entry only): `TestConfigSetLeaseTTLRoutesToYaml` goes red by the exact nil-deref panic you reported (`embeddeddolt.OpenSQL` ← `SetSetting` ← `config.go:225`), deterministic. Real discriminator.

## Minors from R1/R2 — all moved (grepped)

TrimSpace both sides (`lease.go:78`, `yaml_config.go:951`; env side tested) · `captureStderr` defer + goroutine drain (`lease_ttl_config_test.go:147`) · reclaim Long names the knob and the usage explains the displayed default (`reclaim.go:33-36`, `:251-256`; verified in live `--help`) · `LeaseTTL` doc comment (`lease.go:175`) · docs row (`configuration.md:121`) · scaffold fully commented (`init_templates.go:94-95`; a fresh `bd init` leaves `lease.ttl` unset, so nothing pins new projects to 5m) · `ResetForTesting` 7/7 subtests.

## New minors

- **N1** — mutation B (drop the `recognizedConfigKeys` entry only) is caught by nothing: `TestIsRecognizedConfigKey` (`config_validate_key_test.go:11`) is a hand list without `lease.ttl`, and the new test never asserts stderr on the valid-set path. Add the row, or better a structural test that every `YamlOnlyKeys` entry satisfies `isRecognizedConfigKey` (covers `node_id` and the events-journal family too).
- **N2** — mutation C (revert the yaml-side TrimSpace) is caught by nothing: one row `{" 4h ", ok}` in `TestValidateYamlConfigValue_LeaseTTL` (`yaml_config_test.go:783`) pins it. Behavior verified correct end to end.
- **N3** — an env-only `BD_LEASE_TTL=90m` shows in `bd config get` but is invisible to `bd config show` and `list` (no viper default registered, so `AllKeys()` omits it; `config_show.go:124`). The docs advertise the env var as the deployment knob, so the provenance view hides the documented override. One-liner (`SetDefault`/`BindEnv` in `internal/config`). Pre-dates this delta.
- **N4** — a project that ran `bd config set lease.ttl` on a pre-fix build carries an inert DB row: `show`/`list` print `4h (database)` while the runtime uses 5m, and `bd config unset lease.ttl` reports success without touching it. Same hazard as every earlier yaml-only promotion (`backup.*`, `node_id`, events-journal); the fix is a `bd doctor` check for stale DB rows under yaml-only keys, a follow-up issue rather than this PR.
- **N5** — nits: `configuration.md:89`'s YAML-only list lacks `lease.ttl` (already stale for `node_id` etc.); the new test sets `forceGitTracked`/`jsonOutput` with no restore; an explicit `config.IsYamlOnlyKey("lease.ttl")` assertion up front would turn the panic into a readable failure; `cmd/bd`'s shared `captureStderr` (`test_helpers_pure_test.go:250`) has the non-deferred restore R2 flagged in the issueops copy, pre-existing.

## Ran

`go build ./...` 0 · `go vet ./internal/config/... ./internal/storage/issueops/... ./cmd/bd/...` 0 · `go test ./internal/config/... ./internal/storage/issueops/... -run 'LeaseTTL|Lease|IsYamlOnlyKey' -count=1` 0 · `go test ./cmd/bd/ -run 'Reclaim|OlderThan|LeaseTTL|YamlOnly|ConfigSet|ConfigCommands|ConfigNamespaces' -count=1` 0 (155s) · `gofmt -l` clean. Tree clean at 3aebfe575.

Merge-ready from this seat; N1/N2 are two table rows if you want them in before merge.

athosmartins:
Friendly ping on merge status.

This was approved by @bee-ghosttrack on 2026-09-10 (R3 at `3aebfe575`, "VERDICT: APPROVE, five minors, none blocking"), CI is green, and GitHub reports `MERGEABLE` / `CLEAN`.

Is anything outstanding on my side before this can go in? Happy to address the five non-blocking nits (N1-N5) or rebase if that helps — just let me know which you'd prefer.

bee-ghosttrack:
Nothing outstanding on your side: my APPROVE (R3) stands at `3aebfe575aec`, still the head, and N1–N5 were your choice, never a condition. Merging is Julian's or Chris's call rather than mine; a rebase is only needed if main moves under it before then.

PR #5471, opened 2026-08-09
Migrate the reader audit cases onto fixtures that can distinguish what they assert

julianknutsen:
Part of `bd-48ygn`. Additive: 8 new cases plus one fixture strengthening, +989, all three legs, nothing deleted.

Every row here moved because the **fixture** could not tell two behaviours apart — not because an assertion was wrong. The table below leads with that.

| case | mutation → legs reddened | what the old fixture hid |
| --- | --- | --- |
| `ReadyPageIsThePrefixOfTheListWithCountsIncluded` | by-ids page hydrated `SkipCounts`: `issueops/ready_work_counts.go` → dolt+embedded; `domain/db/issue_search_counts.go` → uow | `ReadyLimitBoundary` seeds **3 interchangeable zero-count rows**; here each row has its own signature, and a parent-child + relates-to row makes blocks-only falsifiable |
| `ReadyEphemeralPageKeepsBothPlanesInOnePage` | wisp family hydrated `SkipCounts`: `ready_work_counts.go` → dolt; `ready_work_union.go` → uow | the existing gate case reads one wisp as a set with no counts; here the planes alternate and a **wisp** carries a nonzero count |
| `ReadyPageWiderThanTheHydrationBatch` | chunk stride doubled → dolt, embedded, uow | **nothing anywhere seeded past 200 rows**; length is now checked separately from contents |
| `ListCountsAreBlocksOnlyWhereGetCountsEveryEdge` | detail counts forced blocks-only, all three bodies | every count fixture seeded `blocks` edges **alone**, so the two vocabularies produced identical numbers; exact values (1/2/2) replace "nonzero" |
| `ReadyParentScopesToItsTransitiveDescendants` | descendant walk cut to one hop: `issueops/blocked.go` → dolt; `domain/db/ready_work.go` → uow | the audit gave every descendant **both** a dotted id and an edge, so neither mechanism could be shown to run |
| `ListParentReachesEveryDescendantAndOnlyItsOwn` | prefix LIKE case-folded, `sqlbuild/filter.go` | `ParentID` was driven by **no** reader or querier case at all |
| `ListKeysetWalkLosesNothingAndRepeatsNothing` | empty `AfterID` treated as strict → all three | the existing case reads one page over a 2-row group; this walks a 5-row group at a page size of 2 |
| `ListKeysetPositionNarrowsWithThePredicatesBesideIt` | keyset bound displaces `CreatedBefore` → dolt, uow | the existing case sends a position **and nothing else** |
| `ReadyDefaultTypeExclusions` *(strengthened)* | molecule dropped from `ReadyWorkExcludeTypes` → dolt, uow | the exclusion is a **set** with one member seeded |

## A triage claim corrected

The triage said uow "has no chunker", which would have made the batch-boundary row a large plain equivalence. It has one — `domain/db.queryBatchSize = 200` is its own loop, independently written from the store side's. So that row buys a **real third vote**, and the mutation reddens all three legs from two separate implementations.

`testCommentCount`'s residue (COUNT vs EXISTS) is folded into the counts-split case on both planes rather than given a case of its own. Rows #10/#12/#13 were left to `querier_contract.go`, where an earlier slice took them.

## Note for whoever merges the sibling slices

`backend/conformance` did not compile at `origin/main` when this slice began: #5456's dispatch table named four cases that #5458 and #5462 renamed under it, and omitted two they added. Every sibling slice hit the same wall independently. Main has since been fixed by #5464, and this branch's own repair was dropped in rebase.

Gates: `go build`, `go vet`, `golangci-lint` all clean (exit status checked directly, not through a pipe); all 32 reader cases green on dolt, embeddeddolt and uow, each new one confirmed executing **by name**; role-bundle drift test green.

### DIFF
--- a/backend/conformance/reader_contract.go
+++ b/backend/conformance/reader_contract.go
@@ -1976,6 +1976,211 @@ func RunReaderReadyPageWiderThanTheHydrationBatchIsStillThatPrefix(t *testing.T,
 	}
 }
 
+// RunReaderListCountsAreBlocksOnlyWhereGetCountsEveryEdge pins the exact
+// cardinalities on BOTH read planes at once, and with them the fact that the
+// two planes count different things.
+//
+// A page row's DependencyCount and DependentCount are BLOCKS-ONLY — outgoing
+// blocks edges and reverse blockers — while a detail view's count EVERY
+// outgoing and incoming edge whatever its type. Two vocabularies on one role,
+// and a user sees both: `bd list` prints the first, `bd show` the second.
+//
+// WHAT THIS FIXTURE MAKES OBSERVABLE. Every existing count fixture in this file
+// seeds blocks edges alone — RunReaderGetDetailShapeMatchesTheSeededIssue,
+// RunReaderGetOptionalRowListsAreOffByDefault and
+// RunReaderListSkipCountsDropsTheCardinalitiesAndNothingElse all do — so under
+// them the two vocabularies produce identical numbers and an implementation
+// that swapped them, or that answered the page with the detail view's counts,
+// passes every one. The subject below carries three outgoing edges of three
+// types and three incoming ones, so the two answers are REQUIRED TO DIFFER and
+// the case fails if they agree.
+//
+// The list plane's numbers are also pinned exactly rather than as nonzero.
+// RunReaderListSkipCountsDropsTheCardinalitiesAndNothingElse needs its
+// premise-check to see nonzero counts and asks for no more than that, so a
+// mega-query that answered all three cardinalities with the same total edge
+// count satisfied it. Here CommentCount is 2, DependencyCount is 1 and
+// DependentCount is 2 — three different numbers on one row.
+//
+// PARENT IS THE FOURTH COLUMN and rides the same query as the counts without
+// being one. Its value is pinned against the seeded edge, and a second row with
+// no parent-child edge pins the nil, which is the arm a body returning the
+// row's own id or an empty string would fail.
+//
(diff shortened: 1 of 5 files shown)


julianknutsen:
**Verdict: sound.** The chunker row was re-derived independently and the 205-row seed was confirmed still able to fail — the case is not passing by fixture accident.

---

steveyegge:

Read the summary table, the full strengthened-fixture hunk, and spot-verified the new cases against the series' established bar (#5468's full read). +989/-8, conformance/test-only, all three legs wired, CI fully green (99 checks).

What carries this verdict:
- **The per-case mutation table is the right evidence.** Each row names the implementation mutation and which legs reddened — that is mutation-tested coverage, not asserted coverage. The uow-chunker correction (finding `domain/db.queryBatchSize = 200` is an independently written second loop, making the batch-boundary case a genuine third vote rather than a large equivalence) is exactly the kind of triage-claim verification this series keeps getting right.
- **The one deletion is a strict widening**: the gate-only IssueType-overrides-exclusion assertion becomes a loop over both default-excluded members (gate + molecule), with the molecule — previously observed by nothing — now separately seeded and separately shown to return when named. Nothing weakened.
- **The batch-boundary case closes a real hole**: nothing anywhere seeded past 200 rows, so both chunk loops (store-side and uow-side) ran exactly once in every prior test.
- Main's conformance compile break the body describes was real and is already resolved (#5464, merged yesterday); this branch rebased over it and green CI confirms.

Merging on this verdict.

